PR #5984, opened 2016-11-04
Address issue with integer overflow in RoundRobinInetAddressResolver counter

Spikhalskiy:
Motivation:

Address possible counter overflow issue in RoundRobinInetAddressResolver.

Modification:

Make index calculation safe in case of integer overflow.

Thanks @guidomedina for pointing it https://github.com/netty/netty/commit/5eebe9a06c6873e797b5f8c847086c35ae990ef5#commitcomment-19698899


resolver/src/main/java/io/netty/resolver/RoundRobinInetAddressResolver.java
```diff
@@ -97,4 +97,17 @@ public void operationComplete(Future<List<InetAddress>> future) throws Exception
...
+
+    /**
+     * @param anyInt any integer value to covert
+     * @param len max limit of range
+     * @return integer in (0, len] range, using formula {@code (anyInt % len + anyInt) % len}
+     */
+    protected static int toIndex(int anyInt, int len) {
+        int index = anyInt % len;
+        if (index < 0) {
+            index = index + len;
```
guidomedina: Shouldn't it be from 0 (inclusive) to lenght (exclusive) to mimic a Collection index? `0..(length-1)`
If that's the case then your method should resumed to:
```
  final int index = anyInt % len;
  return index < 0 ? len + index -1 : index;
```

Spikhalskiy: Nope, code that I wrote do specifically ```0..(length-1)```. Would clarify comment, but leave current implementation.

guidomedina: I thought I was crazy but execute the following and see the base of your formula is always 1 and reaches length:
```
    for (int i = -10; i < 0; i++) {
      final int index = i % 5 + 5; // +5 = length
      System.out.println("Without -1 for i = " + i + ", index = " + index);
      System.out.println("With -1 for i = " + i + ", index = " + index);
    }
```
Output:
```
Without -1 for i = -10, index = 5
With -1 for i = -10, index = 4
Without -1 for i = -9, index = 1
With -1 for i = -9, index = 0
Without -1 for i = -8, index = 2
With -1 for i = -8, index = 1
Without -1 for i = -7, index = 3
With -1 for i = -7, index = 2
Without -1 for i = -6, index = 4
With -1 for i = -6, index = 3
Without -1 for i = -5, index = 5
With -1 for i = -5, index = 4
Without -1 for i = -4, index = 1
With -1 for i = -4, index = 0
Without -1 for i = -3, index = 2
With -1 for i = -3, index = 1
Without -1 for i = -2, index = 3
With -1 for i = -2, index = 2
Without -1 for i = -1, index = 4
With -1 for i = -1, index = 3
```

guidomedina: Previous comment edited.

Spikhalskiy: Yes, because your code != mine

Spikhalskiy: ```int index = i % 5 + 5```
and
```
i = i % 5
if (i < 0) {
    i += 5;
}
```
are not same

guidomedina: i % 5 + 5 = (i % 5) + 5 as % takes precedence, if that wasn't true then my output would be equivalent to i % 10 which would show on my example and it doesn't.

```
for (int i = -10; i < 0; i++) {
      final int index = i % 5 + 5;
      System.out.println("Without -1 for i = " + i + ", index = " + index);
      System.out.println("With -1 for i = " + i + ", index = " + (index - 1));
      System.out.println("With index function for = " + i + ", index = " + toIndex(i, 5));
    }
```
Result:
```
Without -1 for i = -10, index = 5
With -1 for i = -10, index = 4
With index function for = -10, index = 0
Without -1 for i = -9, index = 1
With -1 for i = -9, index = 0
With index function for = -9, index = 1
Without -1 for i = -8, index = 2
With -1 for i = -8, index = 1
With index function for = -8, index = 2
Without -1 for i = -7, index = 3
With -1 for i = -7, index = 2
With index function for = -7, index = 3
Without -1 for i = -6, index = 4
With -1 for i = -6, index = 3
With index function for = -6, index = 4
Without -1 for i = -5, index = 5
With -1 for i = -5, index = 4
With index function for = -5, index = 0
Without -1 for i = -4, index = 1
With -1 for i = -4, index = 0
With index function for = -4, index = 1
Without -1 for i = -3, index = 2
With -1 for i = -3, index = 1
With index function for = -3, index = 2
Without -1 for i = -2, index = 3
With -1 for i = -2, index = 2
With index function for = -2, index = 3
Without -1 for i = -1, index = 4
With -1 for i = -1, index = 3
With index function for = -1, index = 4
```

Spikhalskiy: mathematical accident... ok.

You code
```
    @Test
    public void test() {
        for (int i = -10; i <= 0; i++) {
            System.out.print(toIndex(i, 5) + " ");
        }
    }

    protected static int toIndex(int anyInt, int len) {
        int index = anyInt % len;
        if (anyInt < 0) { // anyInt instead of index
            index = index + len - 1;
        }
        return index;
    }
```
4 0 1 2 3 4 0 1 2 **3 0**

Mine:

```
    @Test
    public void test() {
        for (int i = -10; i < 0; i++) {
            System.out.print(toIndex(i, 5) + " ");
        }
    }

    protected static int toIndex(int anyInt, int len) {
        int index = anyInt % len;
        if (index < 0) {
            index = index + len;
        }
        return index;
    }
```
0 1 2 3 4 0 1 2 3 4

If we are talking about "continuity". I think, we should close this discussion, submitted code produces expected output.

guidomedina: After some experiences by overflowing:

```
  protected static int toIndex1(int anyInt, int len) {
    int index = anyInt % len;
    if (anyInt < 0) {
      index = index + len - 1;
    }
    return index;
  }

  protected static int toIndex2(int anyInt, int len) {
    int index = anyInt % len;
    if (index < 0) {
      index = index + len;
    }
    return index;
  }

    int index = Integer.MAX_VALUE - 5;
    for (int i = 0; i < 10; i++) {
      System.out.println("With index1 function for index = " + index + ", index = " + toIndex1(index, 3));
      System.out.println("With index2 function for index = " + index + ", index = " + toIndex2(index, 3));
      System.out.println();
      index++;
    }
}
```

I have come to the conclusion that both formulas are right, but yours will do less computation in cases where it is divisible `when anyInt % len == 0` so I guess I prefer yours ;-)

guidomedina: Your continuity will break from `Integer.MAX_VALUE` to `Integer.MIN_VALUE` (see my latest code), don't jump too fast to conclusions, one divisor might preserve continuity but not the other.

The only way to guarantee continuity is when the length is a power of two, either way to preserve continuity you would have to amortize.

Either way, it will work even with the jump, but it will jump...I'm not trying to have an argument here, I was just trying to simply understand, but I will stop now for good.

Spikhalskiy: > Your continuity will break from Integer.MAX_VALUE to Integer.MIN_VALUE

It's fine. You are trying to handle a gap which is a result of the absence of -Integer.MIN_VALUE. I think it's not a good idea, especially because it introduces a gap around 0. I'm pretty sure there is no easy way to do it without a gap (if we don't write non-obvious tricks that take into account turn number over Integer scale). Not worth any effort.

guidomedina: That's exactly what I meant.
I was able to demostratewhy your function works, negative % always return between -(len + 1) and -1 so adding len to it would make from [1..len - 1], index = 0 is introduced by a divisible counter or zero hence filling the gap and making it always continous and harmonic.

And yes, an amortization for a dumb round robin is not worth the effort.
Thanks for helping fix the other function, would had never found the error as it is unlikely to reach `Long.MAX_VALUE` with few years.

Not even my processing 30m increments per second, but at least it is now theoretically correct.

guidomedina: There is another way but it is 5 times slower on my benchmark on a 4 CPU PC:
`AtomicInteger.compareAndSet()` and not worth for a dumb round robin.

resolver/src/main/java/io/netty/resolver/RoundRobinInetAddressResolver.java
```diff
@@ -64,7 +64,7 @@ public void operationComplete(Future<List<InetAddress>> future) throws Exception
                     if (numAddresses > 0) {
                         // if there are multiple addresses: we shall pick one by one
                         // to support the round robin distribution
-                        promise.setSuccess(inetAddresses.get(index.getAndIncrement() % numAddresses));
+                        promise.setSuccess(inetAddresses.get(toIndex(counter.getAndIncrement(), numAddresses)));
```
normanmaurer: @Spikhalskiy @guidomedina maybe I am missing something but why can't we just use `Math.abs(index.getAndIncrement() % numAddresses)` ?

Spikhalskiy: @normanmaurer This method has one problem. It will change a direction of selecting servers after changing sign. It's not awfully bad, but it's changing behavior from one stable behavior to another one in an unpredictable moment of time (overflow, crossing 0). If this cause any problems, which is not very probable, but still probably - would be hard to find and debug. So, should be good to avoid, even before current code is not harder to read and maintain.
One more - this code is just the same what's doing in Collections.rotate() with parameter in the JDK code.

guidomedina: `Math.abs` changes direction as stated, that's the problem for neg counter and `%` operation, the current proposed implementation is the best you can get for non-power of two sizes.

For power of two which is not the case you can always do it faster by shifting bits but again, not the case here, and for a perfect rotation you would have to use a `CAS` operation which is at least 5 times slower the last time I benchmark it using JDK 8u40+

PR #5914, opened 2016-10-12
HTTP/2 Child Channel and FrameCodec Feature Parity.

buchgr:
First of all, I would like to say that I am really sorry for the large PR. It's really 2 PRs in one, but unfortunately it's been some time after the first PR #5747, performance evaluation #5772 and moving gRPC to this API, in which things have changed so much that it was impossible for me to keep the changes separate and in sync :(.

This PR firstly adds outbound flow control to the `Http2MultiplexCodec` and secondly aims to make the `Http2FrameCodec` ready to be used standalone and without the `Http2MultiplexCodec`. I have already (experimentally) ported gRPC to this `Http2FrameCodec` API, with good results.

Classes like `Http2Stream2`, `Http2Stream2Exception`, `Http2Stream2Visitor`, ... are planned to eventually replace their counterparts without the 2 suffix. That is, once we have a plan to make the Http2Connection API package private.

---

**Motivation:**

This PR (unfortunately) does 4 things:
1. Add outbound flow control to the Http2MultiplexCodec: 
   The HTTP/2 child channel API should interact with HTTP/2 outbound/remote flow control. That is, if a H2 stream used up all its flow control window, the corresponding child channel should be marked unwritable and a writability-changed event should be fired. Similarly, a unwritable child channel should be marked writable and a writability-event should be fired, once a WINDOW_UPDATE frame has been received. The changes are (mostly) contained in ChannelOutboundBuffer, AbstractHttp2StreamChannel and Http2MultiplexCodec.
2. Introduce a Http2Stream2 object, that is used instead of stream identifiers on stream frames. A Http2Stream2 object allows an application to attach state to it, and so a application handler no longer needs to maintain stream state (i.e. in a map(id -> state)) himself.
3. Remove stream state events, which are no longer necessary due to the introduction of Http2Stream2. Also those stream state events have been found hard and complex to work with, when porting gRPC to the Http2FrameCodec.
4. Add support for HTTP/2 frames that have not yet been implemented, like PING and SETTINGS. Also add a Http2FrameCodecBuilder that exposes options from the Http2ConnectionHandler API that couldn't else be used with the frame codec, like buffering outbound streams, window update ratio, frame logger, etc.

**Modifications:**
[...]

Issue #5874, opened 2016-09-29
SniHandlerTest.testServerNameParsing() fails 


radai-rosenblatt:
the test runs up to line 154:

``` java
try {
   // Push the handshake message.
   // Decode should fail because SNI error
   ch.writeInbound(Unpooled.wrappedBuffer(DatatypeConverter.parseHexBinary(tlsHandshakeMessageHex1)));
   ch.writeInbound(Unpooled.wrappedBuffer(DatatypeConverter.parseHexBinary(tlsHandshakeMessageHex)));
   fail();
} catch (DecoderException e) {
   // expected
}

assertThat(ch.finish(), is(true)); <--- line 154
```

but at that point, both `ch.inboundMessages` and `ch.outboundMessages` are null, so `ch.finish()` returns false and the test fails as it expects true.

this happens to me when building under windows 7 using the git shell:

```
$ uname -a
MINGW64_NT-6.1 latitude 2.5.0(0.295/5/3) 2016-03-31 18:47 x86_64 Msys
$ java -version
java version "1.8.0_92"
Java(TM) SE Runtime Environment (build 1.8.0_92-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.92-b14, mixed mode)
$ mvn -version
Apache Maven 3.3.9 (bb52d8502b132ec0a5a3f4c09453c07478323dc5; 2015-11-10T08:41:47-08:00)
Maven home: C:\dev\tools\maven
Java version: 1.8.0_92, vendor: Oracle Corporation
Java home: C:\dev\tools\jdk\jre
Default locale: en_US, platform encoding: Cp1252
OS name: "windows 7", version: "6.1", arch: "amd64", family: "dos"
```

radai-rosenblatt:
also happens on OSX at commit 1c588ce7e9573a6327f0a8a86cf1d754fd10e775:
```
$ uname -a
Darwin rrosenbl-mn1 15.6.0 Darwin Kernel Version 15.6.0: Mon Aug 29 20:21:34 PDT 2016; root:xnu-3248.60.11~1/RELEASE_X86_64 x86_64
$ java -version
java version "1.8.0_102"
Java(TM) SE Runtime Environment (build 1.8.0_102-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.102-b14, mixed mode)
$ mvn -version
[...]

PR #5991, opened 2016-11-08
Allow setting non-default Http/2 server initial settings

nhnFreespirit:
Allow setting non-default Http/2 server initial settings

Motivation:

Currently it is not possible to have an Http/2 server send non default initial settings to clients when doing the initial connection handshake

Modifications:

Add additional constructors to Http2Codec allowing users to specify the initial settings to send to the client and apply locally

Result:

You can now specify non default initial settings

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Codec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Codec.java
@@ -45,19 +45,44 @@ public final class Http2Codec extends ChannelDuplexHandler {
     }
 
     /**
+     * Construct a new handler whose child channels run in the same event loop as this handler.
+     *
+     * @param server {@code true} this is a server
+     * @param streamHandler the handler added to channels for remotely-created streams. It must be
+     *     {@link ChannelHandler.Sharable}. {@code null} if the event loop from the parent channel should be used.
+     * @param initialSettings non default initial settings to send to peer
+     */
+    public Http2Codec(boolean server, ChannelHandler streamHandler, Http2Settings initialSettings) {
+        this(server, new Http2StreamChannelBootstrap().handler(streamHandler), HTTP2_FRAME_LOGGER,
+            initialSettings);
+    }
+
+    /**
      * Construct a new handler whose child channels run in a different event loop.
      *
      * @param server {@code true} this is a server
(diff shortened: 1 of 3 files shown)


nhnFreespirit:
I seem to be affected by https://github.com/netty/netty/issues/5874 so that one tests fails (both on 4.1 and on my branch) when running mvn test

normanmaurer:
@nhnFreespirit interesting... can you add some more details to #5874 like which idk version / OS / hw etc ?

normanmaurer:
@nhnFreespirit LGTM...

@Scottmitch @nmittler @buchgr WDYT ?

normanmaurer:
@nhnFreespirit please squash and adjust the commit message to match our template (basically include there what you did here in the PR message ;))

nhnFreespirit:
That should do it, I think... ;)

buchgr:
LGTM. This is also in #5914.

buchgr:
Although, IIRC this doesn't work for clients. I forgot but there is more complexity involved (I think it was something with connection upgrade). I had to do it on `channelActive`: https://github.com/netty/netty/pull/5914/files#diff-004b9d3b44cadc4c6987434ba0e599beR263 ... maybe we should just get #5914 in fast? There have been a few issues recently trying to fix stuff that's fixed in #5914 :).

nhnFreespirit:
Right, this currently works differently on the client side. In finagle we could make this work as we have a RichHttpToHttp2ConnectionHandler and builder which exposes changing the initial settings, 

see https://github.com/twitter/finagle/commit/4b63e45b99e7afa6be1e86c2981eea3d683ffdf3

My goal with this Netty change is to be able to make the finagle server side config work in the same way (or at all even). This will then finally allow me to make my finagle based thingie (which is what I'm really interested in at the end of the day) work.

What is the time horizon for https://github.com/netty/netty/pull/5914 ? If you choose to discard this PR and wait for that, I might have to re-plan my work or opt to continue with a custom built netty package for the time being.

normanmaurer:
@buchgr I think I would prefer to merge this one first and check #5914 in more detail. Would this be ok for you ?

buchgr:
yeah sure :-).

normanmaurer:
@nhnFreespirit cherry-picked into 4.1 (5b5f6a60317eceff8726c4bb9802f8b1f00dea47) . Thanks!

Issue #5975, opened 2016-11-03
Openssl-based SSLEngine differs slightly in unwrap behavior


jrudolph:
We are experimenting with netty's openssl based SSLEngines as an alternative to the JDK one ([for our effort](https://github.com/akka/akka-http/tree/jr/http2-on-netty-openssl) to produce an HTTP/2 implementation for akka-http). We use 
netty-tcnative-boringssl-static, version 1.1.33.Fork23. Thanks for providing that! When we talked at reactive summit you mentioned that you put much effort into making the openssl-based SSLEngines compatible with the JDK ones. So, you might be interested in the difference we found (knowing that the SSLEngine API has enough degrees of freedom to warrant a lot of behavior...).

I found that our stream-based wrapper around SSLEngine doesn't read all incoming data from the SSLEngine when there is no new incoming data from the network, so the application gets stuck waiting for incoming data. After some debugging I found out that with netty's openssl SSLEngine unwrap seems to produce plain text data in smaller chunks (probably single TLS frames) and keeps buffering the rest of the data internally. The src buffer is fully consumed but calling unwrap again with an empty src buffer will still produce more data. This differs from what the JDK SSLEngine does in two points:
 * the JDK SSLEngine consumes and produces as much data as possible in one go while the openssl one produces less output in one call
 * the JDK SSLEngine doesn't buffer encrypted data internally between calls to `unwrap` but "puts them back" into the src buffer

This also somewhat documented with this comment from https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLEngine.html:

> The SSLEngine produces/consumes complete SSL/TLS packets only, and does not store application data internally between calls to wrap()/unwrap().

Consequently, we [implemented the `unwrap` loop](https://github.com/akka/akka/blob/master/akka-stream/src/main/scala/akka/stream/impl/io/TLSActor.scala#L402) as

```scala
private def doUnwrap(ignoreOutput: Boolean = false): Unit = {
    val result = engine.unwrap(transportInBuffer, userOutBuffer)
    // ...
    result.getStatus match {
      case OK ⇒
        result.getHandshakeStatus match {
          case NEED_WRAP ⇒ // ...
          case FINISHED ⇒ // ...
          case _ ⇒
            if (transportInBuffer.hasRemaining) doUnwrap() // loop here only if there's still src data available
            else flushToUser()
```

[...]

Issue #4238, opened 2015-09-18
OpenSslEngine: `unwrap` sets the position of the source buffer incorrectly if it follows more than one `wrap` call


ijuma:
_Note that the description below has been updated to describe the actual problem after more investigation._

Netty version: commit 94bf412edbea73c0b1dbac4ada8919708c4f369e

Context:

I was trying to use `OpenSslEngine` instead of JDK's `SSLEngine` in Kafka and I ran into an issue where `unwrap` sets the position of the source buffer incorrectly if the `unwrap` call happens after two or more `wrap` calls. I included a test below that shows the problem, but it basically boils down to:
- Each `unwrap` call will return the data for one record
- Multiple `wrap` calls means that the source buffer has more than one record
- In `writeEncryptedData`, there is the following code:

```
netWrote = SSL.writeToBIO(networkBIO, addr, len);
if (netWrote >= 0) {
    src.position(pos + netWrote);
}
```
- So, we set the position of the source buffer based on what is written to `BIO` without taking into account that `SSL.readFromSSL` will not return all of the data that we wrote to the `BIO`.

Steps to reproduce:
1. Checkout the openssl-unwrap branch in my fork https://github.com/ijuma/netty/tree/openssl-unwrap
2. Run OpenSslEngineTest
3. Run JdkSslEngineTest
4. Notice that `testWrapWrapUnwrap` fails with the former and passes with the latter

$ java -version
java version "1.8.0_60"
Java(TM) SE Runtime Environment (build 1.8.0_60-b27)
Java HotSpot(TM) 64-Bit Server VM (build 25.60-b23, mixed mode

Operating system: OS X 10.11

$ uname -a
Darwin ijuma.local 15.0.0 Darwin Kernel Version 15.0.0: Wed Aug 26 16:57:32 PDT 2015; root:xnu-3247.1.106~1/RELEASE_X86_64 x86_64

normanmaurer:
@ijuma thanks will check

normanmaurer:
@ijuma is it fine to include your test-code into netty ?

ijuma:
@normanmaurer of course!

normanmaurer:
@ijuma I just tried to run your test-code against the JDK SSLEngine end it fails here. Is this expected ? I thought you said the SSLEngineImpl does unwrap both records.

normanmaurer:
@ijuma actually fails is the wrong word here... it just runs forever :)

ijuma:
[...]

PR #5995, opened 2016-11-09
[#5975] [#4238] Let OpenSslEngine.wrap(...) / OpenSslEngine.unwrap(..…

normanmaurer:
….) behave like stated in the javadocs.

Motivation:

OpenSslEngine.wrap(...) and OpenSslEngie.unwrap(...) may consume bytes even if an BUFFER_OVERFLOW / BUFFER_UNDERFLOW is detected. This is not correct as it should only consume bytes if it can process them without storing data between unwrap(...) / wrap(...) calls. Beside this it also should only process one record at a time.

Modifications:

- Correctly detect BUFFER_OVERFLOW / BUFFER_UNDERFLOW and only consume bytes if non of them is detected.
- Only process one record per call.

Result:

OpenSslEngine behaves like stated in the javadocs of SSLEngine.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -67,6 +67,7 @@ import static javax.net.ssl.SSLEngineResult.HandshakeStatus.NEED_UNWRAP;
 import static javax.net.ssl.SSLEngineResult.HandshakeStatus.NEED_WRAP;
 import static javax.net.ssl.SSLEngineResult.HandshakeStatus.NOT_HANDSHAKING;
 import static javax.net.ssl.SSLEngineResult.Status.BUFFER_OVERFLOW;
+import static javax.net.ssl.SSLEngineResult.Status.BUFFER_UNDERFLOW;
 import static javax.net.ssl.SSLEngineResult.Status.CLOSED;
 import static javax.net.ssl.SSLEngineResult.Status.OK;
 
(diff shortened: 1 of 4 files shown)


normanmaurer:
@jrudolph this should fix it...

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -775,9 +784,29 @@ public final SSLEngineResult unwrap(
                 }
             }
 
-            // Write encrypted data to network BIO
+            if (len < SslUtils.SSL_RECORD_HEADER_LENGTH) {
+                return new SSLEngineResult(BUFFER_UNDERFLOW, getHandshakeStatus(), 0, 0);
+            }
+
+            int packetLength = SslUtils.getEncryptedPacketLength(srcs, srcsOffset);
```
nmittler: @normanmaurer Have you verified that doing this each time will not impact performance? SslHandler would cache the packetLength so that it wouldn't recalculate it each time. I suspect it's not an issue.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -601,6 +605,11 @@ public final SSLEngineResult wrap(
                 }
             }
 
+            if (dst.remaining() < MAX_ENCRYPTED_PACKET_LENGTH) {
```
nmittler: Is it correct to use a constant here or should we get this value from the session? Or does the session always return this constant?

normanmaurer: It always return the constants

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: I ran a benchmark and the perf was the same

Scottmitch:
lgtm

normanmaurer:
@nmittler so all good ?

nmittler:
@normanmaurer ship it!

normanmaurer:
ok will wait for @jrudolph to verify as well if possible :)

jrudolph:
LGTM! Thanks, again :)

normanmaurer:
Cherry-picked into 4.1 (fc3c9c9523150190760801dd0fbf014909519942) and 4.0 (f87d0698f37e5360af7c7cb9b4f1de7cbf66db64)

PR #5997, opened 2016-11-09
Flush LZ4FrameEncoder buffer when channel flush() is received.

jasobrown:
Motivation:

LZ4FrameEncoder maintains an internal buffer of incoming data compress, and only writes out compressed data when a size threshold is reached. LZ4FrameEncoder does not override the flush() method, and thus the only way to flush data down the pipeline is via more data or close the channel.

Modifications:

Override the flush() function to flush on demand. Also overrode the allocateBuffer() function so we can more accurately size the output buffer (instead of needing to potatntially realloc via buffer.ensureWritable()).

Result:

Implementation works as described.

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/MessageToByteEncoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/MessageToByteEncoder.java
@@ -153,4 +153,8 @@ public abstract class MessageToByteEncoder<I> extends ChannelOutboundHandlerAdap
      * @throws Exception    is thrown if an error accour
      */
     protected abstract void encode(ChannelHandlerContext ctx, I msg, ByteBuf out) throws Exception;
+
+    protected boolean isPreferDirect() {
+        return preferDirect;
+    }
 }
--- a/codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
@@ -24,8 +24,10 @@ import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelPipeline;
 import io.netty.channel.ChannelPromise;
 import io.netty.channel.ChannelPromiseNotifier;
+import io.netty.handler.codec.EncoderException;
 import io.netty.handler.codec.MessageToByteEncoder;
 import io.netty.util.concurrent.EventExecutor;
+import io.netty.util.internal.ObjectUtil;
 import net.jpountz.lz4.LZ4Compressor;
 import net.jpountz.lz4.LZ4Exception;
 import net.jpountz.lz4.LZ4Factory;
@@ -54,6 +56,8 @@ import static io.netty.handler.codec.compression.Lz4Constants.*;
  *  * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *     * * * * * * * * * *
  */
 public class Lz4FrameEncoder extends MessageToByteEncoder<ByteBuf> {
+    static final int DEFAULT_MAX_ENCODE_SIZE = Integer.MAX_VALUE;
+
     private final int blockSize;
 
     /**
@@ -64,27 +68,22 @@ public class Lz4FrameEncoder extends MessageToByteEncoder<ByteBuf> {
     /**
      * Underlying checksum calculator in use.
      */
-    private Checksum checksum;
+    private ByteBufChecksum checksum;
 
     /**
-     * Compression level of current LZ4 encoder (depends on {@link #compressedBlockSize}).
+     * Compression level of current LZ4 encoder (depends on {@link #blockSize}).
      */
     private final int compressionLevel;
 
     /**
-     * Inner byte buffer for outgoing data.
+     * Inner byte buffer for outgoing data. It's capacity will be {@link #blockSize}.
      */
     private ByteBuf buffer;
 
     /**
-     * Current length of buffered bytes in {@link #buffer}.
+     * Maximum size for any buffer to write encoded (compressed) data into.
      */
-    private int currentBlockLength;
-
-    /**
-     * Maximum size of compressed block with header.
-     */
-    private final int compressedBlockSize;
+    private final int maxEncodeSize;
 
     /**
      * Indicates if the compressed stream has been finished.
@@ -131,6 +130,24 @@ public class Lz4FrameEncoder extends MessageToByteEncoder<ByteBuf> {
      * @param checksum        the {@link Checksum} instance to use to check data for integrity
      */
     public Lz4FrameEncoder(LZ4Factory factory, boolean highCompressor, int blockSize, Checksum checksum) {
+        this(factory, highCompressor, blockSize, checksum, DEFAULT_MAX_ENCODE_SIZE);
+    }
+
+        /**
+         * Creates a new customizable LZ4 encoder.
+         *
+         * @param factory         user customizable {@link LZ4Factory} instance
+         *                        which may be JNI bindings to the original C implementation, a pure Java implementation
+         *                        or a Java implementation that uses the {@link sun.misc.Unsafe}
+         * @param highCompressor  if {@code true} codec will use compressor which requires more memory
+         *                        and is slower but compresses more efficiently
+         * @param blockSize       the maximum number of bytes to try to compress at once,
+         *                        must be >= 64 and <= 32 M
+         * @param checksum        the {@link Checksum} instance to use to check data for integrity
+         * @param maxEncodeSize   the maximum size for an encode (compressed) buffer
+         */
+    public Lz4FrameEncoder(LZ4Factory factory, boolean highCompressor, int blockSize,
+                           Checksum checksum, int maxEncodeSize) {
         if (factory == null) {
             throw new NullPointerException("factory");
         }
@@ -139,13 +156,11 @@ public class Lz4FrameEncoder extends MessageToByteEncoder<ByteBuf> {
         }
 
         compressor = highCompressor ? factory.highCompressor() : factory.fastCompressor();
-        this.checksum = checksum;
+        this.checksum = ByteBufChecksum.wrapChecksum(checksum);
 
         compressionLevel = compressionLevel(blockSize);
         this.blockSize = blockSize;
-        currentBlockLength = 0;
-        compressedBlockSize = HEADER_LENGTH + compressor.maxCompressedLength(blockSize);
-
+        this.maxEncodeSize = ObjectUtil.checkPositive(maxEncodeSize, "maxEncodeSize");
         finished = false;
     }
 
@@ -163,54 +178,99 @@ public class Lz4FrameEncoder extends MessageToByteEncoder<ByteBuf> {
     }
 
     @Override
+    protected ByteBuf allocateBuffer(ChannelHandlerContext ctx, ByteBuf msg, boolean preferDirect) {
+        return allocateBuffer(ctx, msg, preferDirect, true);
+    }
+
+    private ByteBuf allocateBuffer(ChannelHandlerContext ctx, ByteBuf msg, boolean preferDirect,
+                                   boolean allowEmptyReturn) {
+        int targetBufSize = 0;
+        int remaining = msg.readableBytes() + buffer.readableBytes();
+
+        // quick overflow check
+        if (remaining < 0) {
+            throw new EncoderException("too much data to allocate a buffer for compression");
+        }
+
+        while (remaining > 0) {
+            int curSize = Math.min(blockSize, remaining);
+            remaining -= curSize;
+            // calculate the total compressed size of the current block (including header) and add to the total
+            targetBufSize += compressor.maxCompressedLength(curSize) + HEADER_LENGTH;
+        }
+
+        // in addition to just the raw byte count, the headers (HEADER_LENGTH) per block (configured via
+        // #blockSize) will also add to the targetBufSize, and the combination of those would never wrap around
+        // again to be >= 0, this is a good check for the overflow case.
+        if (targetBufSize > maxEncodeSize || 0 > targetBufSize) {
+            throw new EncoderException(String.format("requested encode buffer size (%d bytes) exceeds the maximum " +
+                                                     "allowable size (%d bytes)", targetBufSize, maxEncodeSize));
+        }
+
+        if (allowEmptyReturn && targetBufSize < blockSize) {
+            return Unpooled.EMPTY_BUFFER;
+        }
+
+        if (preferDirect) {
+            return ctx.alloc().ioBuffer(targetBufSize, targetBufSize);
+        } else {
+            return ctx.alloc().heapBuffer(targetBufSize, targetBufSize);
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * Encodes the input buffer into {@link #blockSize} chunks in the output buffer. Data is only compressed and
+     * written once we hit the {@link #blockSize}; else, it is copied into the backing {@link #buffer} to await
+     * more data.
+     */
+    @Override
     protected void encode(ChannelHandlerContext ctx, ByteBuf in, ByteBuf out) throws Exception {
         if (finished) {
             out.writeBytes(in);
             return;
         }
 
-        int length = in.readableBytes();
-
         final ByteBuf buffer = this.buffer;
-        final int blockSize = buffer.capacity();
-        while (currentBlockLength + length >= blockSize) {
-            final int tail = blockSize - currentBlockLength;
-            in.getBytes(in.readerIndex(), buffer, currentBlockLength, tail);
-            currentBlockLength = blockSize;
-            flushBufferedData(out);
-            in.skipBytes(tail);
-            length -= tail;
+        int length;
+        while ((length = in.readableBytes()) > 0) {
+            final int nextChunkSize = Math.min(length, buffer.writableBytes());
+            in.readBytes(buffer, nextChunkSize);
+
+            if (!buffer.isWritable()) {
+                flushBufferedData(out);
+            }
         }
-        in.readBytes(buffer, currentBlockLength, length);
-        currentBlockLength += length;
     }
 
     private void flushBufferedData(ByteBuf out) {
-        int currentBlockLength = this.currentBlockLength;
-        if (currentBlockLength == 0) {
+        int flushableBytes = buffer.readableBytes();
+        if (flushableBytes == 0) {
             return;
         }
         checksum.reset();
-        checksum.update(buffer.array(), buffer.arrayOffset(), currentBlockLength);
+        checksum.update(buffer, buffer.readerIndex(), flushableBytes);
         final int check = (int) checksum.getValue();
 
-        out.ensureWritable(compressedBlockSize);
+        final int bufSize = compressor.maxCompressedLength(flushableBytes) + HEADER_LENGTH;
+        out.ensureWritable(bufSize);
         final int idx = out.writerIndex();
         int compressedLength;
         try {
             ByteBuffer outNioBuffer = out.internalNioBuffer(idx + HEADER_LENGTH, out.writableBytes() - HEADER_LENGTH);
             int pos = outNioBuffer.position();
             // We always want to start at position 0 as we take care of reusing the buffer in the encode(...) loop.
-            compressor.compress(buffer.internalNioBuffer(0, currentBlockLength), outNioBuffer);
+            compressor.compress(buffer.internalNioBuffer(0, flushableBytes), outNioBuffer);
             compressedLength = outNioBuffer.position() - pos;
         } catch (LZ4Exception e) {
             throw new CompressionException(e);
         }
         final int blockType;
-        if (compressedLength >= currentBlockLength) {
+        if (compressedLength >= flushableBytes) {
             blockType = BLOCK_TYPE_NON_COMPRESSED;
-            compressedLength = currentBlockLength;
-            out.setBytes(idx + HEADER_LENGTH, buffer, 0, currentBlockLength);
+            compressedLength = flushableBytes;
+            out.setBytes(idx + HEADER_LENGTH, buffer, 0, flushableBytes);
         } else {
             blockType = BLOCK_TYPE_COMPRESSED;
         }
@@ -218,12 +278,20 @@ public class Lz4FrameEncoder extends MessageToByteEncoder<ByteBuf> {
         out.setLong(idx, MAGIC_NUMBER);
         out.setByte(idx + TOKEN_OFFSET, (byte) (blockType | compressionLevel));
         out.setIntLE(idx + COMPRESSED_LENGTH_OFFSET, compressedLength);
-        out.setIntLE(idx + DECOMPRESSED_LENGTH_OFFSET, currentBlockLength);
+        out.setIntLE(idx + DECOMPRESSED_LENGTH_OFFSET, flushableBytes);
         out.setIntLE(idx + CHECKSUM_OFFSET, check);
         out.writerIndex(idx + HEADER_LENGTH + compressedLength);
-        currentBlockLength = 0;
+        buffer.clear();
+    }
 
-        this.currentBlockLength = currentBlockLength;
+    @Override
+    public void flush(final ChannelHandlerContext ctx) throws Exception {
+        if (buffer != null && buffer.isReadable()) {
+            final ByteBuf buf = allocateBuffer(ctx, Unpooled.EMPTY_BUFFER, isPreferDirect(), false);
+            flushBufferedData(buf);
+            ctx.write(buf);
+        }
+        ctx.flush();
     }
 
     private ChannelFuture finishEncode(final ChannelHandlerContext ctx, ChannelPromise promise) {
@@ -235,7 +303,7 @@ public class Lz4FrameEncoder extends MessageToByteEncoder<ByteBuf> {
 
         try {
             final ByteBuf footer = ctx.alloc().heapBuffer(
-                    compressor.maxCompressedLength(currentBlockLength) + HEADER_LENGTH);
+                    compressor.maxCompressedLength(buffer.readableBytes()) + HEADER_LENGTH);
             flushBufferedData(footer);
 
             final int idx = footer.writerIndex();
@@ -330,15 +398,20 @@ public class Lz4FrameEncoder extends MessageToByteEncoder<ByteBuf> {
     }
 
     @Override
-    public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
+    public void handlerAdded(ChannelHandlerContext ctx) {
         this.ctx = ctx;
         // Ensure we use a heap based ByteBuf.
         buffer = Unpooled.wrappedBuffer(new byte[blockSize]);
+        buffer.clear();
     }
 
     @Override
     public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
         super.handlerRemoved(ctx);
         cleanup();
+    }
+
+    final ByteBuf getBackingBuffer() {
+        return buffer;
     }
 }
--- a/codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
+++ b/codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
@@ -15,17 +15,52 @@
  */
 package io.netty.handler.codec.compression;
 
+import java.io.InputStream;
+import java.util.zip.Checksum;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
 import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.ByteBufInputStream;
 import io.netty.buffer.Unpooled;
+import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.embedded.EmbeddedChannel;
+import io.netty.handler.codec.EncoderException;
 import net.jpountz.lz4.LZ4BlockInputStream;
+import net.jpountz.lz4.LZ4Factory;
+import net.jpountz.xxhash.XXHashFactory;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
 
-import java.io.InputStream;
-
-import static org.junit.Assert.*;
+import static io.netty.handler.codec.compression.Lz4Constants.DEFAULT_SEED;
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.when;
 
 public class Lz4FrameEncoderTest extends AbstractEncoderTest {
+    /**
+     * For the purposes of this test, if we pass this (very small) size of buffer into
+     * {@link Lz4FrameEncoder#allocateBuffer(ChannelHandlerContext, ByteBuf, boolean)}, we should get back
+     * an empty buffer.
+     */
+    private static final int NONALLOCATABLE_SIZE = 1;
+
+    @Mock
+    private ChannelHandlerContext ctx;
+
+    /**
+     * A {@link ByteBuf} for mocking purposes, largely because it's difficult to allocate to huge buffers.
+     */
+    @Mock
+    private ByteBuf buffer;
+
+    @Before
+    public void setup() {
+        MockitoAnnotations.initMocks(this);
+        when(ctx.alloc()).thenReturn(ByteBufAllocator.DEFAULT);
+    }
 
     @Override
     public void initChannel() {
@@ -58,5 +93,126 @@ public class Lz4FrameEncoderTest extends AbstractEncoderTest {
         }
 
         return Unpooled.wrappedBuffer(decompressed);
+    }
+
+    @Test
+    public void testAllocateDirectBuffer() {
+        final int blockSize = 100;
+        testAllocateBuffer(blockSize, blockSize - 13, true);
+        testAllocateBuffer(blockSize, blockSize * 5, true);
+        testAllocateBuffer(blockSize, NONALLOCATABLE_SIZE, true);
+    }
+
+    @Test
+    public void testAllocateHeapBuffer() {
+        final int blockSize = 100;
+        testAllocateBuffer(blockSize, blockSize - 13, false);
+        testAllocateBuffer(blockSize, blockSize * 5, false);
+        testAllocateBuffer(blockSize, NONALLOCATABLE_SIZE, false);
+    }
+
+    private void testAllocateBuffer(int blockSize, int bufSize, boolean isDirect) {
+        // allocate the input buffer to an arbitrary size less than the blockSize
+        ByteBuf in = ByteBufAllocator.DEFAULT.buffer(bufSize, bufSize);
+        in.writerIndex(in.capacity());
+
+        ByteBuf out = null;
+        try {
+            Lz4FrameEncoder encoder = newEncoder(blockSize, Lz4FrameEncoder.DEFAULT_MAX_ENCODE_SIZE);
+            out = encoder.allocateBuffer(ctx, in, isDirect);
+            Assert.assertNotNull(out);
+            if (NONALLOCATABLE_SIZE == bufSize) {
+                Assert.assertFalse(out.isWritable());
+            } else {
+                Assert.assertTrue(out.writableBytes() > 0);
+                Assert.assertEquals(isDirect, out.isDirect());
+            }
+        } finally {
+            in.release();
+            if (out != null) {
+                out.release();
+            }
+        }
+    }
+
+    @Test (expected = EncoderException.class)
+    public void testAllocateDirectBufferExceedMaxEncodeSize() {
+        final int maxEncodeSize = 1024;
+        Lz4FrameEncoder encoder = newEncoder(Lz4Constants.DEFAULT_BLOCK_SIZE, maxEncodeSize);
+        int inputBufferSize = maxEncodeSize * 10;
+        ByteBuf buf = ByteBufAllocator.DEFAULT.buffer(inputBufferSize, inputBufferSize);
+        try {
+            buf.writerIndex(inputBufferSize);
+            encoder.allocateBuffer(ctx, buf, false);
+        } finally {
+            buf.release();
+        }
+    }
+
+    private Lz4FrameEncoder newEncoder(int blockSize, int maxEncodeSize) {
+        Checksum checksum = XXHashFactory.fastestInstance().newStreamingHash32(DEFAULT_SEED).asChecksum();
+        Lz4FrameEncoder encoder = new Lz4FrameEncoder(LZ4Factory.fastestInstance(), true,
+                                                      blockSize,
+                                                      checksum,
+                                                      maxEncodeSize);
+        encoder.handlerAdded(ctx);
+        return encoder;
+    }
+
+    /**
+     * This test might be a invasive in terms of knowing what happens inside
+     * {@link Lz4FrameEncoder#allocateBuffer(ChannelHandlerContext, ByteBuf, boolean)}, but this is safest way
+     * of testing the overflow conditions as allocating the huge buffers fails in many CI environments.
+     */
+    @Test (expected = EncoderException.class)
+    public void testAllocateOnHeapBufferOverflowsOutputSize() {
+        final int maxEncodeSize = Integer.MAX_VALUE;
+        Lz4FrameEncoder encoder = newEncoder(Lz4Constants.DEFAULT_BLOCK_SIZE, maxEncodeSize);
+        when(buffer.readableBytes()).thenReturn(maxEncodeSize);
+        buffer.writerIndex(maxEncodeSize);
+        encoder.allocateBuffer(ctx, buffer, false);
+    }
+
+    @Test
+    public void testFlush() {
+        Lz4FrameEncoder encoder = new Lz4FrameEncoder();
+        EmbeddedChannel channel = new EmbeddedChannel(encoder);
+        int size = 27;
+        ByteBuf buf = ByteBufAllocator.DEFAULT.buffer(size, size);
+        buf.writerIndex(size);
+        Assert.assertEquals(0, encoder.getBackingBuffer().readableBytes());
+        channel.write(buf);
+        Assert.assertTrue(channel.outboundMessages().isEmpty());
+        Assert.assertEquals(size, encoder.getBackingBuffer().readableBytes());
+        channel.flush();
+        Assert.assertTrue(channel.finish());
+        Assert.assertTrue(channel.releaseOutbound());
+        Assert.assertFalse(channel.releaseInbound());
+    }
+
+    @Test
+    public void testAllocatingAroundBlockSize() {
+        int blockSize = 100;
+        Lz4FrameEncoder encoder = newEncoder(blockSize, Lz4FrameEncoder.DEFAULT_MAX_ENCODE_SIZE);
+        EmbeddedChannel channel = new EmbeddedChannel(encoder);
+
+        int size = blockSize - 1;
+        ByteBuf buf = ByteBufAllocator.DEFAULT.buffer(size, size);
+        buf.writerIndex(size);
+        Assert.assertEquals(0, encoder.getBackingBuffer().readableBytes());
+        channel.write(buf);
+        Assert.assertEquals(size, encoder.getBackingBuffer().readableBytes());
+
+        int nextSize = size - 1;
+        buf = ByteBufAllocator.DEFAULT.buffer(nextSize, nextSize);
+        buf.writerIndex(nextSize);
+        channel.write(buf);
+        Assert.assertEquals(size + nextSize - blockSize, encoder.getBackingBuffer().readableBytes());
+
+        channel.flush();
+        Assert.assertEquals(0, encoder.getBackingBuffer().readableBytes());
+        Assert.assertTrue(channel.finish());
+        Assert.assertTrue(channel.releaseOutbound());
+        Assert.assertFalse(channel.releaseInbound());
     }
 }


normanmaurer:
@jasobrown would it be possible to have a unit test for this ? :)

jasobrown:
completely reasonable! Let me put something together

normanmaurer:
@jasobrown 😍

jasobrown:
Added unit tests, and amended into previous commit.

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
```diff
@@ -52,4 +77,51 @@ protected ByteBuf decompress(ByteBuf compressed, int originalLength) throws Exce
...
+    }
+
+    @Test
+    public void testAllocateHeapBuffer() {
+        final int blockSize = encoder.getBlockSize();
+        testAllocateBuffer(blockSize - 13, false);
+        testAllocateBuffer(blockSize * 5, false);
+    }
+
+    private void testAllocateBuffer(int bufSize, boolean isDirect) {
```
normanmaurer: static

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
```diff
@@ -52,4 +77,51 @@ protected ByteBuf decompress(ByteBuf compressed, int originalLength) throws Exce
...
+    public void testFlush() {
+        EmbeddedChannel channel = new EmbeddedChannel(encoder);
+        int size = 27;
+        ByteBuf buf = allocator.buffer(size, size);
+        buf.writerIndex(size);
+        Assert.assertEquals(0, encoder.currentBlockLength());
+        channel.write(buf);
+        Assert.assertTrue(channel.outboundMessages().isEmpty());
+        Assert.assertEquals(size, encoder.currentBlockLength());
+        channel.flush();
```
normanmaurer: @jasobrown call `assertTrue(channel.finish())`

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
```diff
@@ -52,4 +77,51 @@ protected ByteBuf decompress(ByteBuf compressed, int originalLength) throws Exce
...
+        Assert.assertFalse(channel.outboundMessages().isEmpty());
```
normanmaurer: Replace with `assertTrue(channel.releaseOutbound());`

normanmaurer: Call  `assertFalse(channel.releaseInbound());``

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
```diff
@@ -15,18 +15,43 @@
...
+    private Lz4FrameEncoder encoder;
+
+    @Mock
+    private ChannelHandlerContext ctx;
+
+    @Before
+    public void setup() {
+        MockitoAnnotations.initMocks(this);
+        when(ctx.alloc()).thenReturn(allocator);
+        encoder = new Lz4FrameEncoder();
```
normanmaurer: Why not create this one directly in the test itself ?

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
```diff
@@ -15,18 +15,43 @@
...
+
+    private Lz4FrameEncoder encoder;
+
+    @Mock
+    private ChannelHandlerContext ctx;
+
+    @Before
+    public void setup() {
+        MockitoAnnotations.initMocks(this);
+        when(ctx.alloc()).thenReturn(allocator);
```
normanmaurer: Just inline the allocator directly?

jasobrown:
@normanmaurer I've addressed your nits, except making {{testAllocateBuffer}} static - it references the mocked {{ChannelHandlerContext ctx}}

jasobrown:
really committed the updates this time

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -163,6 +163,24 @@ private static int compressionLevel(int blockSize) {
     }
 
     @Override
+    protected ByteBuf allocateBuffer(ChannelHandlerContext ctx, ByteBuf msg, boolean preferDirect) {
```
Scottmitch: consider just passing `ByteBufAllocator` instead of `ChannelHandlerContext`.

codec/src/main/java/io/netty/handler/codec/MessageToByteEncoder.java
```diff
@@ -46,7 +46,7 @@
 public abstract class MessageToByteEncoder<I> extends ChannelOutboundHandlerAdapter {
 
     private final TypeParameterMatcher matcher;
-    private final boolean preferDirect;
+    protected final boolean preferDirect;
```
Scottmitch: I wonder if instead of exposing this it would be preferable to just expose a method which can take it into account by default:

```
protected final ByteBuf allocateBuffer(ChannelHandlerContext ctx, ByteBuf msg) {
  return allocateBuffer(ctx, msg, preferDirect);
}
```

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -163,6 +163,24 @@ private static int compressionLevel(int blockSize) {
...
+        int bufSize = 0;
+        int remaining = msg.readableBytes();
+        while (remaining > 0) {
+            int curSize = Math.min(blockSize, remaining);
+            remaining -= curSize;
+            // calculate the total compressed size of the current block (including header) and add to the total
+            bufSize += compressor.maxCompressedLength(curSize) + HEADER_LENGTH;
```
Scottmitch: should we worry about overflow here?

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -341,4 +370,12 @@ public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
         super.handlerRemoved(ctx);
         cleanup();
     }
+
+    public int getBlockSize() {
```
Scottmitch: make these new methods `final`?

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
Scottmitch: What may be useful to protect against overflow and a "unexpected large input" is to have some "private final int maxAllocationSize" that can be configured via the constructor, and enforced here. @normanmaurer - WDYT?

codec/src/main/java/io/netty/handler/codec/MessageToByteEncoder.java
jasobrown: That won't help in the case of where I'm accessing that value from: `flush()`. We could either create a `protected boolean isPreferDirect()`, or ... ?

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
normanmaurer: @Scottmitch I am not sure if we really need a constructor in the base class or are you talking about `Lz4Encoder` here ?

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
normanmaurer: actually I think we should also remove the `get` prefix as we not have it in `currentBlockLenght` as well.

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
jasobrown: i suspect @Scottmitch meant for it to be in the `Lz4FrameEncoder`.

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
jasobrown: lol, not putting `get` as a prefix to the (new) `currentBlockLength` method was my fault! Let me know which convention you would prefer

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
Scottmitch: I was just thinking about `Lz4FrameEncoder`. This class adds new behavior to calculate the max buffer size ahead of time, but opens a new vector of allocating "too much". We should have a way to configure what "too much" means (default can potentially be very large, but IMO we should have some story here)

codec/src/main/java/io/netty/handler/codec/MessageToByteEncoder.java
Scottmitch: @jasobrown - if this method existing in the base class you would just call it and wouldn't need to directly access `preferDirect` (or know how it is determined).

```java
Lz4FrameEncoder.allocateBuffer(ctx, msg)
// which calls
Lz4FrameEncoder.allocateBuffer(ctx, msg,
        /*MessageToByteEncoder has the responsibility to calculate the preferDirect*/)
```

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
Scottmitch: @normanmaurer @jasobrown - Are we all on the same page as my last suggestion ... WDYT?

Scottmitch:
@normanmaurer @jasobrown - I have a few comments above that still need to be resolved. Ping when ready for me to take another look.

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
normanmaurer: yep..

normanmaurer:
@jasobrown ping

jasobrown:
@normanmaurer @Scottmitch sorry to be coming back to this so late.

wrt to buffer sizing, it seems this has opened a can of worms larger than this ticket, and my from the discussion it seems like @Scottmitch imagines something more broad across netty for ensuring buffer sizes don't allocate "too much". Should we just revert that part from the PR?

jasobrown:
also, looks like forgot to push the last set of changes to resolve some nits. (facepalm)

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
jasobrown: done

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -129,8 +132,10 @@ public Lz4FrameEncoder(boolean highCompressor) {
      * @param blockSize       the maximum number of bytes to try to compress at once,
      *                        must be >= 64 and <= 32 M
      * @param checksum        the {@link Checksum} instance to use to check data for integrity
+     * @param maxEncodeSize   the maximum size for an encode (compressed) buffer
      */
-    public Lz4FrameEncoder(LZ4Factory factory, boolean highCompressor, int blockSize, Checksum checksum) {
+    public Lz4FrameEncoder(LZ4Factory factory, boolean highCompressor, int blockSize,
+                           Checksum checksum, int maxEncodeSize) {
```
normanmaurer: @jasobrown we also need to preserve the old constructor to not break stuff.

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -173,6 +177,11 @@ protected ByteBuf allocateBuffer(ChannelHandlerContext ctx, ByteBuf msg, boolean
             bufSize += compressor.maxCompressedLength(curSize) + HEADER_LENGTH;
         }
 
+        if (bufSize > maxEncodeSize) {
+            String errorMsg = "requested encode buffer size (%d bytes) exceeds the maximum allowable size (%d bytes)";
+            throw new EncoderException(String.format(errorMsg, bufSize, maxEncodeSize));
```
normanmaurer: just inline the `errorMsg` ?

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
```diff
@@ -106,6 +112,20 @@ private void testAllocateBuffer(int bufSize, boolean isDirect) {
...
+    @Test (expected = EncoderException.class)
+    public void testAllocateDirectBuffer_ExceedMaxEncodeSize() {
+        final int maxEncodeSize = 1024;
+        Checksum checksum = XXHashFactory.fastestInstance().newStreamingHash32(DEFAULT_SEED).asChecksum();
+        Lz4FrameEncoder encoder = new Lz4FrameEncoder(LZ4Factory.fastestInstance(), true,
+                                                      Lz4Constants.DEFAULT_BLOCK_SIZE,
+                                                      checksum,
+                                                      maxEncodeSize);
+        int inputBufferSize = maxEncodeSize * 10;
+        ByteBuf buf = PooledByteBufAllocator.DEFAULT.buffer(inputBufferSize, inputBufferSize);
```
normanmaurer: `ByteBufAllocator.DEFAULT.`

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
```diff
@@ -106,6 +112,20 @@ private void testAllocateBuffer(int bufSize, boolean isDirect) {
...
+        buf.writerIndex(inputBufferSize);
+        encoder.allocateBuffer(null, buf, false);
```
normanmaurer: call `buf.release()` in a finally block?

jasobrown: I believe the `channel.finish()` and `channel.release*()` methods will already release the buffer.

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
jasobrown: done

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
jasobrown: the style checker kept bitching at me about line length, and this seemed the most readable version; splitting the message across lines looked ugly.

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
normanmaurer: @jasobrown maybe I am missing something but you are not calling any of these methods here and notes the `channel` at all so nope you will need to take care by yourself ;)

jasobrown: ahh, sorry - was looking at `testFlush()`. fixing this one now

jasobrown: pushed

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -115,7 +118,7 @@ public Lz4FrameEncoder() {
      */
     public Lz4FrameEncoder(boolean highCompressor) {
         this(LZ4Factory.fastestInstance(), highCompressor, DEFAULT_BLOCK_SIZE,
-                XXHashFactory.fastestInstance().newStreamingHash32(DEFAULT_SEED).asChecksum());
+                XXHashFactory.fastestInstance().newStreamingHash32(DEFAULT_SEED).asChecksum(), DEFAULT_MAX_ENCODE_SIZE);
```
Scottmitch: remove the new argument? Delegating to the other constructor should supply `DEFAULT_MAX_ENCODE_SIZE` for us in a central location.

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -131,6 +134,24 @@ public Lz4FrameEncoder(boolean highCompressor) {
...
+         *                        or a Java implementation that uses the {@link sun.misc.Unsafe}
+         * @param highCompressor  if {@code true} codec will use compressor which requires more memory
+         *                        and is slower but compresses more efficiently
+         * @param blockSize       the maximum number of bytes to try to compress at once,
+         *                        must be >= 64 and <= 32 M
+         * @param checksum        the {@link Checksum} instance to use to check data for integrity
+         * @param maxEncodeSize   the maximum size for an encode (compressed) buffer
+         */
+    public Lz4FrameEncoder(LZ4Factory factory, boolean highCompressor, int blockSize,
+        Checksum checksum, int maxEncodeSize) {
```
Scottmitch: please fix formatting for argument alignment and javadoc alignment

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -163,6 +183,29 @@ private static int compressionLevel(int blockSize) {
...
+        int bufSize = 0;
+        int remaining = msg.readableBytes();
+        while (remaining > 0) {
+            int curSize = Math.min(blockSize, remaining);
+            remaining -= curSize;
+            // calculate the total compressed size of the current block (including header) and add to the total
+            bufSize += compressor.maxCompressedLength(curSize) + HEADER_LENGTH;
+        }
+
+        if (bufSize > maxEncodeSize) {
```
Scottmitch: isn't it possible this could overflow in the while loop and we wouldn't detect it?

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -341,4 +395,12 @@ public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
         super.handlerRemoved(ctx);
         cleanup();
     }
+
+    public final int blockSize() {
```
Scottmitch: can we just make these new accessors package private for now, or are they generally useful outside the context of unit tests?

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -163,6 +183,29 @@ private static int compressionLevel(int blockSize) {
...
+            bufSize += compressor.maxCompressedLength(curSize) + HEADER_LENGTH;
+        }
+
+        if (bufSize > maxEncodeSize) {
+            String errorMsg = "requested encode buffer size (%d bytes) exceeds the maximum allowable size (%d bytes)";
+            throw new EncoderException(String.format(errorMsg, bufSize, maxEncodeSize));
+        }
+
+        if (preferDirect) {
+            return ctx.alloc().ioBuffer(bufSize);
```
Scottmitch: we should also pass `maxEncodeSize` as the second parameter (e.g. `maxCapacity`) to these allocations. This will ensure the buffer will not grow beyond this max capacity.

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -194,7 +237,8 @@ private void flushBufferedData(ByteBuf out) {
         checksum.update(buffer.array(), buffer.arrayOffset(), currentBlockLength);
         final int check = (int) checksum.getValue();
 
-        out.ensureWritable(compressedBlockSize);
+        final int bufSize = compressor.maxCompressedLength(buffer.readableBytes()) + HEADER_LENGTH;
+        out.ensureWritable(bufSize);
```
Scottmitch: is it possible the compression will take less than `bufSize`? If so we may fail the compression upfront if the `out` buffer has a `maxCapacity` set which would accommodate the real compression size, but can't accommodate the max `bufSize` size.

Scottmitch: also should `buffer` be `out` here?

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -341,4 +395,12 @@ public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
         super.handlerRemoved(ctx);
         cleanup();
     }
+
+    public final int blockSize() {
+        return blockSize;
+    }
+
+    public final int currentBlockLength() {
+        return currentBlockLength;
```
Scottmitch: can you explain what this is? Can we just use `buffer.readableBytes()` instead?

Also some places where it is used seem suspect. For example line 257...

```java
out.setBytes(idx + HEADER_LENGTH, buffer, 0, currentBlockLength);
```
Should the following changes be made here?
`buffer` -> `buf`
`0` -> `buf.readerIndex()`
`currentBlockLength` -> `buf.readableBytes()`

jasobrown: I've simply, perhaps naively, used an existing field to check the state of the class. I'll look at the use of that field, as well, but hadn't thought about it too much, tbh.

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
jasobrown: I think you are correct here ... I was just being wishy-washy about not doing it before :). done.

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
jasobrown: mathematically possible, largely improbable - but adding an additional `&& 0 > bufSize` check for ensuring safety.

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
jasobrown: If I put all the parameters on one line, it exceeds 120 chars and the style checker complains and fails compilation. I looked at other Netty functions that have a longer parameter list, and I didn't necessarily see a pattern of consistency as to how/where the new line is placed.

Please advise.

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
jasobrown: done

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
jasobrown: added test, as well

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
jasobrown: Hold off on reviewing ... I'm running into a problem with enforcing the max buffer size (from the buffer returned from `allocateBuffer()` and the reuse of the member field `buffer` across multiple `encode()` invocations.

jasobrown: UPDATE 2: ok, problem is only with current mock up of removing `currentBlockLength`. Please review away :)

jasobrown: OK, I've ditched `currentBlockLength` in favor of `buffer.readableBytes()`. This required a little clean up, which allowed me to make some of the code (especially the `while` loop in `encode()`) simpler and more understandable.

As we don't always write all the `msg`'s bytes that come into `encode()`, we only compress and write when the block size is reached (or, now on `flush()`). Thus, some data will remain buffered in the member field `buffer`. In `allocateBuffer()`, I now account for any of that buffered data when calculating the output buffer size. This was not a bug earlier because I did not set the max buffer size on the `allocateBuffer()`'s return value, and we are calling `out.ensureWritable()` later on (which would have incresed the buffer size).

Further, I've optimized `allocateBuffer()` to return an empty buffer if we know we do not have enough data between any previously not written data (still buffered in `buffer`) plus the incoming `msg`'s bytes. I would have preferred to return either `null` or `Unpooled.EMPY_BUFFER` but we call `release()` on the buffer after `encode()` is invoked from `MessageToByteEncoder#write()`.

jasobrown:
reposting this comment as it looks like the PR swallowed it in amongst everything else :-/

OK, I've ditched `currentBlockLength` in favor of `buffer.readableBytes()`. This required a little clean up, which allowed me to make some of the code (especially the `while` loop in `encode()`) simpler and more understandable.

As we don't always write all the data out that comes into `encode()`, we only compress and write when the block size is reached (or, now on `flush()`). Thus, some data will remain buffered in the member field `buffer`. In `allocateBuffer()`, I now account for any of that buffered data when calculating the output buffer size. This was not a bug earlier because I did not set the max buffer size on the `allocateBuffer()`'s return value, and we are calling `out.ensureWritable()` later on (which would have incresed the buffer size).

Further, I've optimized `allocateBuffer()` to return an empty buffer if we know we do not have enough data between any previously not written data (still buffered in `buffer`) plus the incoming `msg`'s bytes. I would have preferred to return either `null` or `Unpooled.EMPY_BUFFER` but we call `release()` on the buffer after `encode()` is invoked from `MessageToByteEncoder#write()`.

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -330,15 +391,24 @@ private ChannelHandlerContext ctx() {
     }
 
     @Override
-    public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
+    public void handlerAdded(ChannelHandlerContext ctx) {
         this.ctx = ctx;
         // Ensure we use a heap based ByteBuf.
         buffer = Unpooled.wrappedBuffer(new byte[blockSize]);
+        buffer.clear();
```
Scottmitch: is this necessary?

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -163,38 +177,77 @@ private static int compressionLevel(int blockSize) {
...
+                buffer.clear();
+            }
         }
-        in.readBytes(buffer, currentBlockLength, length);
-        currentBlockLength += length;
     }
 
     private void flushBufferedData(ByteBuf out) {
-        int currentBlockLength = this.currentBlockLength;
+        int currentBlockLength = buffer.readableBytes();
```
Scottmitch: `currentBlockLength` -> `readableBytes`?

the usage of `block length` here is a bit confusing. This doesn't necessary relate to a block

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -163,38 +177,77 @@ private static int compressionLevel(int blockSize) {
...
-            length -= tail;
+        int length = in.readableBytes();
+        while (length > 0) {
+            final int nextChunkSize = Math.min(length, buffer.writableBytes());
+            in.readBytes(buffer, nextChunkSize);
+            length -= nextChunkSize;
+
+            if (!buffer.isWritable()) {
+                flushBufferedData(out);
+                buffer.clear();
```
Scottmitch: should we instead just have `flushBufferedData` be responsible for setting the indexes on `buffer` according to how much was consumed?

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -330,15 +391,24 @@ private ChannelHandlerContext ctx() {
...
+        buffer.clear();
     }
 
     @Override
     public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
         super.handlerRemoved(ctx);
         cleanup();
     }
+
+    final int blockSize() {
```
Scottmitch: this is not used ... can we remove it?

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -163,38 +177,77 @@ private static int compressionLevel(int blockSize) {
...
 
     @Override
+    protected ByteBuf allocateBuffer(ChannelHandlerContext ctx, ByteBuf msg, boolean preferDirect) {
+        return  allocateBuffer(ctx, msg, preferDirect, true);
+    }
+
+    ByteBuf allocateBuffer(ChannelHandlerContext ctx, ByteBuf msg, boolean preferDirect,
+                                     boolean allowEmptyReturn) {
+        int targetBufSize = 0;
+        int remaining = msg.readableBytes() + buffer.readableBytes();
```
jasobrown: We will get a more accurate `targetBufSize` if we include all the bytes in the calculation upfront; this is because we would potentially oversize a little extra.

Imagine this contrived example:

- `buffer`: capacity = 100, writeIndex = 70
- msg.readableBytes()` = 5

As we would execute a separate iteration of the loop for both `buffer` and `msg`, we would add:
- `70` and `HEADER_LENGTH` (which is 21) = 91 
- `5` and `HEADER_LENGTH` = 26
for a total of 115, which would cause use to allocate a buffer even though it will not be populated in `allocate()` as per my last round of changes.

If we add the number of bytes of `buffer` and `msg`before the loop (like I am doing), we get:
- `70` + `5` + HEADER_LENGTH = 96
and we would not allocate a buffer.

It's a little more efficient this way.

Now, wrt to overflow of just adding together `msg.readableBytes() + buffer.readableBytes()`,  you are still screwed as we still can only return a buffer from this function that is `Integer.MAX_VALUE` in size. Further, given that `compressor.maxCompressedLength()` returns a value that is larger than the input number, the situation will not improve. Thus, if anything, we can add a check after `int remaining = msg.readableBytes() + buffer.readableBytes()` to see if it overflowed or is above some value that we know will overflow.

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -163,38 +177,77 @@ private static int compressionLevel(int blockSize) {
...
+            targetBufSize += compressor.maxCompressedLength(curSize) + HEADER_LENGTH;
+        }
+
+        if (targetBufSize > maxEncodeSize || 0 > targetBufSize) {
+            String errorMsg = "requested encode buffer size (%d bytes) exceeds the maximum allowable size (%d bytes)";
+            throw new EncoderException(String.format(errorMsg, targetBufSize, maxEncodeSize));
+        }
+
+        if (allowEmptyReturn && targetBufSize < blockSize) {
+            //would prefer to return null or Unpooled.EMPTY_BUFFER, but MessageToByteEncoder calls buffer.release()
```
jasobrown: Huh, I naively assumed that calling `Unpooled.EMPTY_BUFFER.release()` would have thrown an exception after the first release, but I guess that's not the case as you probably wouldn't have made it a constant :) . I ran a quick unit test and indeed it does not error out, so switching to use `Unpooled.EMPTY_BUFFER`.

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
jasobrown: `currentBlockLength`, in it's previous incarnation, was a loose counter of how much data had been written to the `buffer`. I'll go ahead and rename that variable to ... `flushableBytes`. lmk if that's a reasonable choice.

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
jasobrown: Yes, it is as `Unpooled.wrappedBuffer()` sets the `writeIndex` to the length of the array and we need the write index to be at 0 when we start using this class.

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
jasobrown: nuked

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -143,9 +159,7 @@ public Lz4FrameEncoder(LZ4Factory factory, boolean highCompressor, int blockSize
 
         compressionLevel = compressionLevel(blockSize);
         this.blockSize = blockSize;
-        currentBlockLength = 0;
-        compressedBlockSize = HEADER_LENGTH + compressor.maxCompressedLength(blockSize);
-
+        this.maxEncodeSize = maxEncodeSize;
```
jasobrown: done

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -163,38 +177,77 @@ private static int compressionLevel(int blockSize) {
     }
 
     @Override
+    protected ByteBuf allocateBuffer(ChannelHandlerContext ctx, ByteBuf msg, boolean preferDirect) {
+        return  allocateBuffer(ctx, msg, preferDirect, true);
```
jasobrown: done

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
Scottmitch: sgtm. thanks for clarifying.

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
```diff
@@ -52,4 +82,128 @@ protected ByteBuf decompress(ByteBuf compressed, int originalLength) throws Exce
...
+        return encoder;
+    }
+
+    @Test (expected = EncoderException.class)
+    public void testAllocateDirectBuffer_OverflowsOutputSize() {
+        final int maxEncodeSize = Integer.MAX_VALUE;
+        Lz4FrameEncoder encoder = newEncoder(Lz4Constants.DEFAULT_BLOCK_SIZE, maxEncodeSize);
+        ByteBuf buf = null;
+        try
+        {
```
Scottmitch: This is causing a build failure

```
[io.netty:netty-codec] /var/lib/teamcity-agent/work/a00bdf77e621d005/codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java:156:9: '{' should be on the previous line.
```

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -163,67 +178,120 @@ private static int compressionLevel(int blockSize) {
...
+        }
+
+        while (remaining > 0) {
+            int curSize = Math.min(blockSize, remaining);
+            remaining -= curSize;
+            // calculate the total compressed size of the current block (including header) and add to the total
+            targetBufSize += compressor.maxCompressedLength(curSize) + HEADER_LENGTH;
+        }
+
+        // as compressor.maxCompressedLength(Integer.MAX_VALUE) would never wrap around again to be >= 0,
```
Scottmitch: Can we elaborate on this comment because it depends upon a few different factors:

- MIN_BLOCK_SIZE is 64
- HEADER_LENGTH is 21
- compressor.maxCompressedLength upper bound is `length + length / 255 + 16`

So all of these constraints means that even if `remaining == Integer.MAX_INT` we cannot overflow back around to `>= 0`.

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -163,67 +178,120 @@ private static int compressionLevel(int blockSize) {
...
             out.writeBytes(in);
             return;
         }
 
+        final ByteBuf buffer = this.buffer;
         int length = in.readableBytes();
+        while (length > 0) {
+            final int nextChunkSize = Math.min(length, buffer.writableBytes());
+            in.readBytes(buffer, nextChunkSize);
+            length -= nextChunkSize;
```
Scottmitch: can we just use `in.readableBytes()` instead (see below)?

```java
int length;
while ((length = in.readableBytes()) > 0) {
  in.readBytes(buffer, Math.min(length, buffer.writableBytes()));

  if (!buffer.isWritable()) {
    flushBufferedData(out);
  }
}
```

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
```diff
@@ -52,4 +82,128 @@ protected ByteBuf decompress(ByteBuf compressed, int originalLength) throws Exce
...
+
+    private void testAllocateBuffer(int blockSize, int bufSize, boolean isDirect) {
+        // allocate the input buffer to an arbitrary size less than the blockSize
+        ByteBuf in = ByteBufAllocator.DEFAULT.buffer(bufSize, bufSize);
+        in.writerIndex(in.capacity());
+
+        ByteBuf out = null;
+        try {
+            Lz4FrameEncoder encoder = newEncoder(blockSize, Lz4FrameEncoder.DEFAULT_MAX_ENCODE_SIZE);
+            out = encoder.allocateBuffer(ctx, in, isDirect);
```
Scottmitch: I wonder if we can test this using the public interface of the class instead (that is what we ultimately care about). For example maybe calling encode with small buffers (smaller than block size) and testing that things don't get written until a `flush` occurs. maybe `testFlush` already tests this?

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
```diff
@@ -52,4 +82,128 @@ protected ByteBuf decompress(ByteBuf compressed, int originalLength) throws Exce
...
+        } finally {
+            in.release();
+            if (out != null) {
+                out.release();
+            }
+        }
+    }
+
+    @Test (expected = EncoderException.class)
+    public void testAllocateDirectBuffer_ExceedMaxEncodeSize() {
```
Scottmitch: remove the `_` in the method name? we typically don't include this in method names.

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
```diff
@@ -52,4 +82,128 @@ protected ByteBuf decompress(ByteBuf compressed, int originalLength) throws Exce
...
+        final int maxEncodeSize = Integer.MAX_VALUE;
+        Lz4FrameEncoder encoder = newEncoder(Lz4Constants.DEFAULT_BLOCK_SIZE, maxEncodeSize);
+        ByteBuf buf = null;
+        try
+        {
+            buf = ByteBufAllocator.DEFAULT.buffer(maxEncodeSize, maxEncodeSize);
+            buf.writerIndex(maxEncodeSize);
+            encoder.allocateBuffer(ctx, buf, false);
+        } catch (OutOfDirectMemoryError e) {
+            // this sometimes fails on the continuous integration server if we try to allocate Integer.MAX_VALUE.
```
Scottmitch: can we just allocate a smaller buffer and still satisfy the test to accommodate our under provisioned CI servers?

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
```diff
@@ -52,4 +82,128 @@ protected ByteBuf decompress(ByteBuf compressed, int originalLength) throws Exce
...
+
+        int nextSize = size - 1;
+        buf = ByteBufAllocator.DEFAULT.buffer(nextSize, nextSize);
+        buf.writerIndex(nextSize);
+        channel.write(buf);
+        Assert.assertEquals(size + nextSize - blockSize, encoder.getBuffer().readableBytes());
+
+        channel.flush();
+        Assert.assertEquals(0, encoder.getBuffer().readableBytes());
+    }
```
Scottmitch: Make sure you call `channel.finishAndReleaseAll();` or verify the channel has some output buffers like you do in the previous `testFlush` test.

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
jasobrown: Correct, I believe test `testFlush()` does do that, but it's goal can be seen as a bit higher level - testing `Lz4FrameEncoder` in the context of a pipeline, whereas `testAllocateBuffer` tests the specific method.

it's not often complain about too much testing ;) so let me know what you want to do here

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
jasobrown: I usually add those in test method names, to more easily delineate similar tests. I'll remove though

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
jasobrown: fixed

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
jasobrown: This is to test the integer overflow aspects of `allocateBuffer()`, and the only way to get into that situation is to have a buffer near `Integer.MAX_VALUE`. Anecdotally, looks like allocating on-heap buffers did not have a problem and only direct buffers failed. Since it doesn't matter which buffer type gets us into the overflow situation, I will remove the large allocation for the direct buffer and rely on the on-heap allocation to test that path.

jasobrown: UPDATE: Well, now the test fails locally with using on-heap buffers, but still works with the direct buffer. Instead of actually allocating a buffer, I just need to satisfy the `ByteBuf` interface, either with a mock or some kind of stub. Do you have any examples of that lying around I can rip off? Trying to keep it simple ;)

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
jasobrown: done

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
jasobrown: done

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
jasobrown: updated the comment, lmk

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
Scottmitch: Tests are good, but I generally prefer to validate the interface contracts as opposed to internal implementation details (which may be subject to change).

codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
Scottmitch: > This is to test the integer overflow aspects of allocateBuffer()

You can mock the `ByteBufAllocator` interface and return this from the mocked `ChannelHandlerContext`, and then mock the `ByteBuf` interface as necessary. Its generally a tradeoff in test robustness & mock completeness vs time to write tests. Here is an untested example:

```java
@Mock
private ChannelHandlerContext ctx;

@Mock
private ByteBufAllocator alloc;

@Mock
private ByteBuf buffer;

when(ctx.alloc()).thenReturn(alloc);
when(alloc.ioBuffer(anyInt(), anyInt())).thenReturn(buffer);
when(alloc.heapBuffer(anyInt(), anyInt())).thenReturn(buffer);
when(buffer.readableBytes()).thenReturn(Integer.MAX_VALUE);
```

Here is an example of mocking `ByteBuf` [1], but I don't think its something we do very frequently.

[1] https://github.com/netty/netty/blob/4.1/buffer/src/test/java/io/netty/buffer/ReadOnlyByteBufTest.java#L77

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -163,67 +178,120 @@ private static int compressionLevel(int blockSize) {
...
     private void flushBufferedData(ByteBuf out) {
-        int currentBlockLength = this.currentBlockLength;
-        if (currentBlockLength == 0) {
+        int flushableBytes = buffer.readableBytes();
+        if (flushableBytes == 0) {
             return;
         }
         checksum.reset();
-        checksum.update(buffer.array(), buffer.arrayOffset(), currentBlockLength);
+        checksum.update(buffer, buffer.arrayOffset(), flushableBytes);
```
Scottmitch: with a `ByteBufChecksum` I think this should be:

```java
checksum.update(buffer, buffer.readerIndex(), flushableBytes);
```

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -163,67 +178,120 @@ private static int compressionLevel(int blockSize) {
...
         final int check = (int) checksum.getValue();
 
-        out.ensureWritable(compressedBlockSize);
+        final int bufSize = compressor.maxCompressedLength(buffer.readableBytes()) + HEADER_LENGTH;
```
Scottmitch: `buffer.readableBytes()` -> `flushableBytes` (re-use since you have already calculated readable bytes).

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
jasobrown: looks like I missed that one - thx

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
jasobrown: done, and pushed up

Scottmitch:
I kicked off another build bcz it looks like we ran into https://github.com/netty/netty/issues/6161

jasobrown:
squashed and rebased the PR. tests should hopefully kick off asap...

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -330,15 +398,20 @@ private ChannelHandlerContext ctx() {
...
+        buffer.clear();
     }
 
     @Override
     public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
         super.handlerRemoved(ctx);
         cleanup();
     }
+
+    final ByteBuf getBuffer() {
```
jasobrown: renamed to `getBackingBuffer`

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -163,67 +178,120 @@ private static int compressionLevel(int blockSize) {
...
+            // calculate the total compressed size of the current block (including header) and add to the total
+            targetBufSize += compressor.maxCompressedLength(curSize) + HEADER_LENGTH;
+        }
+
+        // in addition to just the raw byte count, the headers (HEADER_LENGTH) per block (configured via
+        // #blockSize) will also add to the targetBufSize, and the combination of those would never wrap around
+        // again to be >= 0, this is a good check for the overflow case.
+        if (targetBufSize > maxEncodeSize || 0 > targetBufSize) {
+            String errorMsg = "requested encode buffer size (%d bytes) exceeds the maximum allowable size (%d bytes)";
+            throw new EncoderException(String.format(errorMsg, targetBufSize, maxEncodeSize));
```
jasobrown: I did it like this because it will line wrap in an odd way. either way, done

jasobrown:
addressed minor nits from @normanmaurer, amended to last (squashed) commit, and pushed

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -163,67 +178,120 @@ private static int compressionLevel(int blockSize) {
...
         int compressedLength;
         try {
             ByteBuffer outNioBuffer = out.internalNioBuffer(idx + HEADER_LENGTH, out.writableBytes() - HEADER_LENGTH);
             int pos = outNioBuffer.position();
             // We always want to start at position 0 as we take care of reusing the buffer in the encode(...) loop.
-            compressor.compress(buffer.internalNioBuffer(0, currentBlockLength), outNioBuffer);
+            compressor.compress(buffer.internalNioBuffer(0, flushableBytes), outNioBuffer);
             compressedLength = outNioBuffer.position() - pos;
         } catch (LZ4Exception e) {
             throw new CompressionException(e);
```
Scottmitch: What do we expect to happen with `buffer` at this point? Is it possible that appending more data will remedy the issue? Should we do `buffer.clear()` or just leave the state unchanged?

jasobrown: `buffer` does get cleared at the [end of the method](https://github.com/jasobrown/netty/blob/f7dd2534c742b18cc1e262c2739ab6151ab6f529/codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java#L284). We need to wait until after this `try`/`catch` block as might end up using the [uncompressed data](https://github.com/jasobrown/netty/blob/f7dd2534c742b18cc1e262c2739ab6151ab6f529/codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java#L273)

> Is it possible that appending more data will remedy the issue?

I"m not sure what you mean here.

Scottmitch: > buffer does get cleared at the end of the method. 

That is why I'm asking in this case. The control flow from this line will not reach the end of the method. If an exception is encountered during compression what are the expected next steps? Is it possible that we can recover with the existing data (for example by appending more data so just throwing is OK), should we give up and just put the "uncompressed" data into `out`, or is this unrecoverable and we should just dump the current data?

I know you didn't introduce this code, but I'm curious if the current code is correct.

Scottmitch: discussed with @jasobrown offline. I think we can handle this as a followup PR because this behavior hasn't changed.

Scottmitch:
4.1 (3ea807e)

Thanks @jasobrown !

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
Scottmitch: See https://github.com/netty/netty/issues/6248

PR #5998, opened 2016-11-09
Not use InternalThreadLocalMap where access may be done from outside …

normanmaurer:
…the EventLoop.

Motivation:

We should not use the InternalThreadLocalMap where access may be done from outside the EventLoop as this may create a lot of memory usage while not be reused anyway.

Modifications:

Not use InternalThreadLocalMap in places where the code-path will likely be executed from outside the EventLoop.

Result:

Less memory bloat.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -24,7 +24,6 @@ import io.netty.util.ResourceLeak;
 import io.netty.util.ResourceLeakDetector;
 import io.netty.util.ResourceLeakDetectorFactory;
 import io.netty.util.internal.EmptyArrays;
-import io.netty.util.internal.InternalThreadLocalMap;
 import io.netty.util.internal.PlatformDependent;
 import io.netty.util.internal.StringUtil;
 import io.netty.util.internal.ThrowableUtil;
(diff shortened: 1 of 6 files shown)


normanmaurer:
@nmittler @Scottmitch @trustin PTAL

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1147,7 +1147,7 @@ public final void setEnabledCipherSuites(String[] cipherSuites) {
 
     @Override
     public final String[] getEnabledProtocols() {
-        List<String> enabled = InternalThreadLocalMap.get().arrayList();
+        List<String> enabled = new ArrayList<String>(5);
```
Scottmitch: whoops missed a call to`add` ... 5 -> 6

Scottmitch:
1 comment and lgtm

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: doh!

normanmaurer:
Cherry-picked into 4.1 (705e3f629ab802539586c7d92f7888697a44c450) and 4.0 (de4692e6e0c64385606f9dfae729cc49d8ac1460)

PR #6001, opened 2016-11-09
WebSocketClientHandshaker.rawPath(URI) should use the raw query

adrianluisgonzalez:
Motivation:

If the wsURL contains an encoded query, it will be decoded when generating the raw path.  For example if the wsURL is http://test.org/path?a=1%3A5, the returned raw path would be /path?a=1:5

Modifications:

Use wsURL.getRawQuery() rather than wsURL.getQuery()

Result:

rawPath will now return /path?a=1%3A5

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
@@ -434,7 +434,7 @@ public abstract class WebSocketClientHandshaker {
      */
     static String rawPath(URI wsURL) {
         String path = wsURL.getRawPath();
-        String query = wsURL.getQuery();
+        String query = wsURL.getRawQuery();
         if (query != null && !query.isEmpty()) {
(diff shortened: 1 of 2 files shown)


normanmaurer:
@slandelle WDYT ?

normanmaurer:
@adrianluisgonzalez also can you add an unit test please

adrianluisgonzalez:
@normanmaurer added unit test.  Sorry, that was lazy of me ...

normanmaurer:
@slandelle @jroper WDYT ?

jroper:
Makes sense to me. I don't think we've encountered this issue in Play, but it probably does correspond to a bug in Play.

jroper:
Actually sorry, I just realised it's the client handshaker, Play doesn't offer a WebSocket client. But Lagom does.

normanmaurer:
Cherry-picked into 4.1 (baac352f74d7e5b2980ed9c1071515e6f85598eb) and 4.0 (68100cdd7f53077aa35fe656dfc4a2571a8eeb92).

@adrianluisgonzalez thanks a lot!

PR #6002, opened 2016-11-09
ByteBuf Input Stream Reference Count Ownership

Scottmitch:
Motivation:
Netty provides a adaptor from ByteBuf to Java's InputStream interface. The JDK Stream interfaces have an explicit lifetime because they implement the Closable interface. This lifetime may be differnt than the ByteBuf which is wrapped, and controlled by the interface which accepts the JDK Stream. However Netty's ByteBufInputStream currently does not take reference count ownership of the underlying ByteBuf. There may be no way for existing classes which only accept the InputStream interface to communicate when they are done with the stream, other than calling close(). This means that when the stream is closed it may be appropriate to release the underlying ByteBuf, as the ownership of the underlying ByteBuf resource may be transferred to the Java Stream.

Motivation:
- ByteBufInputStream.close() supports taking reference count ownership of the underyling ByteBuf

Result:
ByteBufInputStream can assume reference count ownership so the underlying ByteBuf can be cleaned up when the stream is closed.

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/serialization/ObjectDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/serialization/ObjectDecoder.java
@@ -71,14 +71,11 @@ public class ObjectDecoder extends LengthFieldBasedFrameDecoder {
             return null;
         }
 
-        ObjectInputStream is = new CompactObjectInputStream(new ByteBufInputStream(frame), classResolver);
-        Object result = is.readObject();
-        is.close();
-        return result;
-    }
-
-    @Override
-    protected ByteBuf extractFrame(ChannelHandlerContext ctx, ByteBuf buffer, int index, int length) {
-        return buffer.slice(index, length);
+        ObjectInputStream ois = new CompactObjectInputStream(new ByteBufInputStream(frame, true), classResolver);
+        try {
+            return ois.readObject();
+        } finally {
+            ois.close();
+        }
     }
 }
--- a/handler/src/main/java/io/netty/handler/ssl/SslContext.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslContext.java
@@ -1010,13 +1010,24 @@ public abstract class SslContext {
         CertificateFactory cf = CertificateFactory.getInstance("X.509");
         X509Certificate[] x509Certs = new X509Certificate[certs.length];
 
+        int i = 0;
         try {
-            for (int i = 0; i < certs.length; i++) {
-                x509Certs[i] = (X509Certificate) cf.generateCertificate(new ByteBufInputStream(certs[i]));
+            for (; i < certs.length; i++) {
+                InputStream is = new ByteBufInputStream(certs[i], true);
+                try {
+                    x509Certs[i] = (X509Certificate) cf.generateCertificate(is);
+                } finally {
+                    try {
+                        is.close();
+                    } catch (IOException e) {
+                        // This is not expected to happen, but re-throw in case it does.
+                        throw new RuntimeException(e);
+                    }
+                }
             }
         } finally {
-            for (ByteBuf buf: certs) {
-                buf.release();
+            for (; i < certs.length; i++) {
+                certs[i].release();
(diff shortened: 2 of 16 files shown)


codec/src/main/java/io/netty/handler/codec/serialization/ObjectDecoder.java
```diff
@@ -20,6 +20,7 @@
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.codec.LengthFieldBasedFrameDecoder;
 
+import java.io.InputStream;
```
netkins: MINOR Remove this unused import 'java.io.InputStream'. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

handler/src/main/java/io/netty/handler/ssl/SslContext.java
```diff
@@ -1010,13 +1010,24 @@ protected static TrustManagerFactory buildTrustManagerFactory(
         CertificateFactory cf = CertificateFactory.getInstance("X.509");
         X509Certificate[] x509Certs = new X509Certificate[certs.length];
 
+        int i = 0;
         try {
-            for (int i = 0; i < certs.length; i++) {
-                x509Certs[i] = (X509Certificate) cf.generateCertificate(new ByteBufInputStream(certs[i]));
+            for (; i < certs.length; i++) {
+                InputStream is = new ByteBufInputStream(certs[i], true);
```
netkins: BLOCKER Close this "ByteBufInputStream". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2095)

handler/src/main/java/io/netty/handler/ssl/SslContext.java
```diff
@@ -1010,13 +1010,24 @@ protected static TrustManagerFactory buildTrustManagerFactory(
...
+                try {
+                    x509Certs[i] = (X509Certificate) cf.generateCertificate(is);
+                } finally {
+                    try {
+                        is.close();
+                    } catch (IOException e) {
+                        // This is not expected to happen, but re-throw in case it does.
+                        throw new RuntimeException(e);
```
netkins: MAJOR Refactor this code to not throw exceptions in finally blocks. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1163)
BLOCKER Remove this throw statement from this finally block. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1143)

handler/src/main/java/io/netty/handler/ssl/SslContext.java
Scottmitch: its closed on line 1021

handler/src/main/java/io/netty/handler/ssl/SslContext.java
Scottmitch: I'm planning on ignoring this unless someone has an alternative suggestion (logging, rethrow unchecked, etc...)

normanmaurer: ignoring is fine.

normanmaurer:
@Scottmitch I wonder if we should not add the same constructor to `ByteBufOutputStream` for consistency. I know most of the times you not want to release but there may be situations where it is fine.

Scottmitch:
@normanmaurer - I had this previously, but this is really a special case. The `ByteBuf` in the `ByteBufOutputStream` is the sink for the data source. The user is expected to get the `ByteBuf` and have a lifetime independent the stream. However I can add it back if you think there is value (e.g. consistency)

normanmaurer:
@Scottmitch I see... ok lets just continue with it

Scottmitch:
buffer leak in build fixed by https://github.com/netty/netty/issues/6007

Scottmitch:
4.1 (c1932a8) 4.0 (2ec06e2)

PR #6003, opened 2016-11-09
CompatibleObjectEncoder cached ObjectOutputStream backed by release b…

Scottmitch:
…uffer bug

Motivation:
ObjectOutputStream uses a Channel Attribute to cache a ObjectOutputStream which is backed by a ByteBuf that may be released after an object is encoded and the underlying buffer is written to the channel. On subsequent encode operations the cached ObjectOutputStream will be invalid and lead to a reference count exception.

Modifications:
- CompatibleObjectEncoder should not cache a ObjectOutputStream.

Result:
CompatibleObjectEncoder doesn't use a cached object backed by a released ByteBuf.


Scottmitch:
@trustin @normanmaurer - PTAL

normanmaurer:
@Scottmitch ship it.

Scottmitch:
4.1 (e47da7b) 4.0 (7cf5630)

Issue #5980, opened 2016-11-04
IllegalArgumentException in io.netty.util.internal.PriorityQueue


rkapsi:
4.1.7-SNAPSHOT

It's always the same message: `e.priorityQueueIndex(): 0 (expected: -1)`. It's strange because I don't see `WeightedFairQueueByteDistributor$State#priorityQueueIndex(0)` ever being called (or the `priorityQueueIndex` field getting set). 

```java
2016-11-04 13:12:14,722 [WorkerThread-2] ExceptionHandler ERROR: name=Http2ExceptionHandler, channel=[id: 0x020ee146, L:/aaa.aaa.aaa.aaa:443 - R:/bbb.bbb.bbb.bbb:63000]
 java.lang.IllegalArgumentException: e.priorityQueueIndex(): 0 (expected: -1)
	at io.netty.util.internal.PriorityQueue.offer(PriorityQueue.java:83) ~[netty-all-4.1.7.Final-SNAPSHOT.jar:4.1.7.Final-SNAPSHOT]
	at io.netty.util.internal.PriorityQueue.offer(PriorityQueue.java:32) ~[netty-all-4.1.7.Final-SNAPSHOT.jar:4.1.7.Final-SNAPSHOT]
	at io.netty.handler.codec.http2.WeightedFairQueueByteDistributor$State.offer(WeightedFairQueueByteDistributor.java:299) ~[netty-all-4.1.7.Final-SNAPSHOT.jar:4.1.7.Final-SNAPSHOT]
	at io.netty.handler.codec.http2.WeightedFairQueueByteDistributor.distributeToChildren(WeightedFairQueueByteDistributor.java:198) ~[netty-all-4.1.7.Final-SNAPSHOT.jar:4.1.7.Final-SNAPSHOT]
	at io.netty.handler.codec.http2.WeightedFairQueueByteDistributor.distribute(WeightedFairQueueByteDistributor.java:130) ~[netty-all-4.1.7.Final-SNAPSHOT.jar:4.1.7.Final-SNAPSHOT]
	at io.netty.handler.codec.http2.DefaultHttp2RemoteFlowController$WritabilityMonitor.writePendingBytes(DefaultHttp2RemoteFlowController.java:631) ~[netty-all-4.1.7.Final-SNAPSHOT.jar:4.1.7.Final-SNAPSHOT]
[...]

PR #6004, opened 2016-11-09
HTTP/2 WeightedFairQueueByteDistributor Bug

Scottmitch:
Motivation:
If a stream is not able to send any data (flow control window for the stream is exhausted) but has descendants who can send data then WeightedFairQueueByteDistributor may incorrectly modify the pseudo time and also double add the associated state to the parent's priority queue. The pseudo time should only be modified if a node is moved in the priority tree, and not if there happens to be no active streams in its descendent tree and a descendent is moved (e.g. removed from the tree because it wrote all data and the last data frame was EOS). Also the state objects for WeightedFairQueueByteDistributor should only appear once in any queue. If this condition is violated the pseudo time accounting would be biased at and assumptions in WeightedFairQueueByteDistributor would be invalidated.

Modifications:
- WeightedFairQueueByteDistributor#isActiveCountChangeForTree should not allow re-adding to the priority queue if we are currently processing a node in the distribution algorithm. The distribution algorithm will re-evaluate if the node should be re-added on the tail end of the recursion.

Result:
Fixes https://github.com/netty/netty/issues/5980

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
@@ -25,7 +25,6 @@ import static io.netty.handler.codec.http2.Http2CodecUtil.CONNECTION_STREAM_ID;
 import static io.netty.handler.codec.http2.Http2CodecUtil.streamableBytes;
 import static io.netty.handler.codec.http2.Http2Error.INTERNAL_ERROR;
 import static io.netty.handler.codec.http2.Http2Exception.connectionError;
-import static io.netty.util.internal.ObjectUtil.checkNotNull;
 import static java.lang.Math.min;
 
 /**
@@ -85,7 +84,7 @@ public final class WeightedFairQueueByteDistributor implements StreamByteDistrib
                     if (state.activeCountForTree != 0) {
                         State pState = state(parent);
                         pState.offerAndInitializePseudoTime(state);
-                        pState.isActiveCountChangeForTree(state.activeCountForTree);
+                        pState.activeCountChangeForTree(state.activeCountForTree);
                     }
                 }
             }
@@ -98,7 +97,7 @@ public final class WeightedFairQueueByteDistributor implements StreamByteDistrib
                     if (state.activeCountForTree != 0) {
                         State pState = state(parent);
                         pState.remove(state);
-                        pState.isActiveCountChangeForTree(-state.activeCountForTree);
+                        pState.activeCountChangeForTree(-state.activeCountForTree);
                     }
                 }
             }
@@ -113,8 +112,6 @@ public final class WeightedFairQueueByteDistributor implements StreamByteDistrib
 
     @Override
     public boolean distribute(int maxBytes, Writer writer) throws Http2Exception {
-        checkNotNull(writer, "writer");
-
         // As long as there is some active frame we should write at least 1 time.
         if (connectionState.activeCountForTree == 0) {
             return false;
@@ -146,7 +143,7 @@ public final class WeightedFairQueueByteDistributor implements StreamByteDistrib
     }
 
     private int distribute(int maxBytes, Writer writer, State state) throws Http2Exception {
-        if (state.active) {
+        if (state.isActive()) {
             int nsent = min(maxBytes, state.streamableBytes);
             state.write(nsent, writer);
             if (nsent == 0 && maxBytes != 0) {
@@ -176,10 +173,11 @@ public final class WeightedFairQueueByteDistributor implements StreamByteDistrib
         long oldTotalQueuedWeights = state.totalQueuedWeights;
         State childState = state.poll();
         State nextChildState = state.peek();
+        childState.setDistributing();
         try {
             assert nextChildState == null || nextChildState.pseudoTimeToWrite >= childState.pseudoTimeToWrite :
-                "nextChildState.pseudoTime(" + nextChildState.pseudoTimeToWrite + ") < " + " childState.pseudoTime(" +
-                    childState.pseudoTimeToWrite + ")";
+                "nextChildState[" + nextChildState.stream.id() + "].pseudoTime(" + nextChildState.pseudoTimeToWrite +
+                ") < " + " childState[" + childState.stream.id() + "].pseudoTime(" + childState.pseudoTimeToWrite + ")";
             int nsent = distribute(nextChildState == null ? maxBytes :
                             min(maxBytes, (int) min((nextChildState.pseudoTimeToWrite - childState.pseudoTimeToWrite) *
                                                 childState.stream.weight() / oldTotalQueuedWeights + allocationQuantum,
@@ -191,7 +189,8 @@ public final class WeightedFairQueueByteDistributor implements StreamByteDistrib
             childState.updatePseudoTime(state, nsent, oldTotalQueuedWeights);
             return nsent;
         } finally {
-            // Do in finally to ensure the internal state is not corrupted if an exception is thrown.
+            childState.unsetDistributing();
+            // Do in finally to ensure the internal flags is not corrupted if an exception is thrown.
             // The offer operation is delayed until we unroll up the recursive stack, so we don't have to remove from
             // the priority queue due to a write operation.
             if (childState.activeCountForTree != 0) {
@@ -215,11 +214,13 @@ public final class WeightedFairQueueByteDistributor implements StreamByteDistrib
      * The remote flow control state for a single stream.
      */
     private final class State implements PriorityQueueNode<State> {
+        private static final int STATE_IS_ACTIVE = 0x1;
+        private static final int STATE_IS_DISTRIBUTING = 0x2;
         final Http2Stream stream;
         private final Queue<State> queue;
         int streamableBytes;
         /**
-         * Count of nodes rooted at this sub tree with {@link #active} equal to {@code true}.
+         * Count of nodes rooted at this sub tree with {@link #isActive()} equal to {@code true}.
          */
         int activeCountForTree;
         private int priorityQueueIndex = INDEX_NOT_IN_QUEUE;
@@ -228,11 +229,11 @@ public final class WeightedFairQueueByteDistributor implements StreamByteDistrib
          */
         long pseudoTimeToWrite;
         /**
-         * A pseudo time maintained for immediate children to base their {@link pseudoTimeToSend} off of.
+         * A pseudo time maintained for immediate children to base their {@link #pseudoTimeToWrite} off of.
          */
         long pseudoTime;
         long totalQueuedWeights;
-        boolean active;
+        private byte flags;
 
         State(Http2Stream stream) {
             this(stream, 0);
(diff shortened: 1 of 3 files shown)


Scottmitch:
@nmittler @rkapsi - PTAL

normanmaurer:
@Scottmitch please fix checkstyle error :)

codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
```diff
@@ -113,8 +114,6 @@ public void updateStreamableBytes(StreamState state) {
 
     @Override
     public boolean distribute(int maxBytes, Writer writer) throws Http2Exception {
-        checkNotNull(writer, "writer");
```
normanmaurer: why you removed this ?

codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
```diff
@@ -256,19 +257,37 @@ void isActiveCountChangeForTree(int increment) {
...
+                assert activeCountForTree != increment ||
+                       priorityQueueIndex == INDEX_NOT_IN_QUEUE ||
+                       pState.queue.contains(this) :
+                     "State[" + stream.id() + "].activeCountForTree changed from 0 to " + increment + " is in a queue" +
+                     ", but not in parent[ " + pState.stream.id() +"]'s queue";
                 if (activeCountForTree == 0) {
                     pState.remove(this);
-                } else if (activeCountForTree - increment == 0) { // if frame count was 0 but is now not, then queue.
+                } if (activeCountForTree == increment && !isInProcess()) {
+                    // if frame count was 0 but is now not, and this node is not already in a queue (assumed to be
```
normanmaurer: Just nit but start the sentence with an uppercase `I` ;)

codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
```diff
@@ -338,5 +381,33 @@ public int priorityQueueIndex() {
         public void priorityQueueIndex(int i) {
             priorityQueueIndex = i;
         }
+
+        @Override
+        public String toString() {
+            StringBuilder sb = InternalThreadLocalMap.get().stringBuilder();
```
normanmaurer: If we expect to call this from outside the EventLoop please just create a new StringBuilder.
See https://github.com/netty/netty/pull/5998

codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
Scottmitch: done

codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
normanmaurer: whatever you prefer... just wanted to understand why :) Its your call

codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
Scottmitch: +1. I was actually thinking about this this morning. Other places where we use this and don't provide any other option may have similar issues.

codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
normanmaurer: lol...

Scottmitch:
@normanmaurer - Build failure seems like a CI issue, java version incompatibility?

https://garage.netty.io/teamcity/viewLog.html?buildId=13590&buildTypeId=pulls_netty&tab=buildLog#_state=80&focus=193455

```
[23:37:46][io.netty:netty-parent] Caused by: java.lang.UnsupportedClassVersionError: org/sonarsource/scanner/maven/SonarQubeMojo : Unsupported major.minor version 52.0
[23:37:46][io.netty:netty-parent] 	at java.lang.ClassLoader.defineClass1(Native Method)
[23:37:46][io.netty:netty-parent] 	at java.lang.ClassLoader.defineClass(ClassLoader.java:800)
[23:37:46][io.netty:netty-parent] 	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
[23:37:46][io.netty:netty-parent] 	at java.net.URLClassLoader.defineClass(URLClassLoader.java:449)
[23:37:46][io.netty:netty-parent] 	at java.net.URLClassLoader.access$100(URLClassLoader.java:71)
[23:37:46][io.netty:netty-parent] 	at java.net.URLClassLoader$1.run(URLClassLoader.java:361)
[23:37:46][io.netty:netty-parent] 	at java.net.URLClassLoader$1.run(URLClassLoader.java:355)
[23:37:46][io.netty:netty-parent] 	at java.security.AccessController.doPrivileged(Native Method)
[23:37:46][io.netty:netty-parent] 	at java.net.URLClassLoader.findClass(URLClassLoader.java:354)
[23:37:46][io.netty:netty-parent] 	at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClassFromSelf(ClassRealm.java:401)
[23:37:46][io.netty:netty-parent] 	at org.codehaus.plexus.classworlds.strategy.SelfFirstStrategy.loadClass(SelfFirstStrategy.java:42)
[23:37:46][io.netty:netty-parent] 	at org.codehaus.plexus.classworlds.realm.ClassRealm.unsynchronizedLoadClass(ClassRealm.java:271)
[23:37:46][io.netty:netty-parent] 	at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClass(ClassRealm.java:247)
[23:37:46][io.netty:netty-parent] 	at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClass(ClassRealm.java:239)
[23:37:46][io.netty:netty-parent] 	at org.eclipse.sisu.space.URLClassSpace.loadClass(URLClassSpace.java:139)
```

Scottmitch:
@rkapsi - have you had a chance to try this out?

codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
nmittler: Staying "slim and dangerous" ... is this another meth joke? :)

codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
```diff
@@ -324,6 +340,30 @@ void close() {
             updateStreamableBytes(0, false);
         }
 
+        boolean isActive() {
+            return (state & 0x1) != 0;
```
nmittler: Let's make constants for these?

codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
```diff
@@ -256,19 +255,36 @@ void isActiveCountChangeForTree(int increment) {
             activeCountForTree += increment;
             if (!stream.isRoot()) {
                 State pState = state(stream.parent());
+                assert activeCountForTree != increment ||
```
nmittler: FYI, the name of this method (`isActiveCountChangeForTree`) is a bit confusing ... it sounds like it should be a `boolean`.

Maybe we should fix that in a follow-on PR?

codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
```diff
@@ -324,6 +340,30 @@ void close() {
...
+
+        private void setActive() {
+            state |= 0x1;
+        }
+
+        private void unsetActive() {
+            state &= ~0x1;
+        }
+
+        boolean isInProcess() {
```
nmittler: Maybe something like `isDistributing`?

codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
```diff
@@ -228,11 +227,11 @@ int streamableBytes0(Http2Stream stream) {
          */
         long pseudoTimeToWrite;
         /**
-         * A pseudo time maintained for immediate children to base their {@link pseudoTimeToSend} off of.
+         * A pseudo time maintained for immediate children to base their {@link #pseudoTimeToWrite} off of.
          */
         long pseudoTime;
         long totalQueuedWeights;
-        boolean active;
+        private byte state;
```
nmittler: Maybe something like `flags`, since the class name is already `State`?

codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
```diff
@@ -256,19 +255,36 @@ void isActiveCountChangeForTree(int increment) {
...
+                     "State[" + stream.id() + "].activeCountForTree changed from 0 to " + increment + " is in a queue" +
+                     ", but not in parent[ " + pState.stream.id() + "]'s queue";
                 if (activeCountForTree == 0) {
                     pState.remove(this);
-                } else if (activeCountForTree - increment == 0) { // if frame count was 0 but is now not, then queue.
+                } else if (activeCountForTree == increment && !isInProcess()) {
+                    // If frame count was 0 but is now not, and this node is not already in a queue (assumed to be
+                    // pState's queue) then enqueue it. If this State object is being processed the pseudoTime for this
+                    // node should not be adjusted, and the node will be added back to the queue/tree structure after it
+                    // is done being processed. This may happen if the activeCountForTree == 0 (a node which can't
```
nmittler: What about the case above where we do `pState.remove(this)`? ... will that happen automatically after the distribute finishes as well? Sorry, haven't followed the logic through :)

rkapsi:
@Scottmitch I'll give it a spin in ~30m-1h.

normanmaurer:
@Scottmitch will check the ci problem

codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
Scottmitch: we could rename this method `distribute_meth` if you prefer?

codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
Scottmitch: done

codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
Scottmitch: sure done

codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
Scottmitch: `flags` sgtm

codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
Scottmitch: I can just remove the `is` prefix. Would this make things more clear?

codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
Scottmitch: If `isDistributing()` (new name) is true that means the `State` object has been removed from its parent's queue to be processed, and will be added back (if necessary) after it has been processed. So I don't think it is a problem if this method is called while the `State` object is being processed, because it already won't be in the queue so `queue.remove(..)` should be a noop.

It is possible for a element to be remove in a transient manner (during a tree shift), but it will be added back after the tree shift operation is complete if it has children which are "active" ... this is what happens during the new unit test.

Scottmitch:
@nmittler - Comments addressed.

normanmaurer:
@Scottmitch fixed the CI problem...

normanmaurer:
@Scottmitch LGTM

rkapsi:
@Scottmitch fix looks good. Branch is running without errors on our end. Thanks!

codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
nmittler: Yep, I think that's better. Thanks.

Scottmitch:
4.1 (c4e96d0)

Thanks for reporting and providing debug info @rkapsi!

PR #6006, opened 2016-11-10
Revert counter optimization in RoundRobin until better implementation found

Spikhalskiy:
Motivation:
Revert counter based index selection from RoundRobinInetAddressResolver as it can cause stuck into static resolution in some setups.
Related discussion:
https://github.com/netty/netty/commit/5eebe9a06c6873e797b5f8c847086c35ae990ef5#commitcomment-19727669

### DIFF
--- a/resolver/src/main/java/io/netty/resolver/RoundRobinInetAddressResolver.java
+++ b/resolver/src/main/java/io/netty/resolver/RoundRobinInetAddressResolver.java
@@ -19,6 +19,7 @@ import io.netty.util.concurrent.EventExecutor;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.FutureListener;
 import io.netty.util.concurrent.Promise;
+import io.netty.util.internal.ThreadLocalRandom;
 import io.netty.util.internal.UnstableApi;
 
 import java.net.InetAddress;
@@ -27,7 +28,6 @@ import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * A {@link NameResolver} that resolves {@link InetAddress} and force Round Robin by choosing a single address
@@ -38,7 +38,6 @@ import java.util.concurrent.atomic.AtomicInteger;
 @UnstableApi
 public class RoundRobinInetAddressResolver extends InetNameResolver {
     private final NameResolver<InetAddress> nameResolver;
-    private final AtomicInteger index = new AtomicInteger();
 
     /**
      * @param executor the {@link EventExecutor} which is used to notify the listeners of the {@link Future} returned by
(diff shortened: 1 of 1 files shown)


resolver/src/main/java/io/netty/resolver/RoundRobinInetAddressResolver.java
```diff
@@ -64,7 +63,7 @@ public void operationComplete(Future<List<InetAddress>> future) throws Exception
                     if (numAddresses > 0) {
                         // if there are multiple addresses: we shall pick one by one
                         // to support the round robin distribution
-                        promise.setSuccess(inetAddresses.get(index.getAndIncrement() % numAddresses));
+                        promise.setSuccess(inetAddresses.get(shiftSize(numAddresses)));
```
normanmaurer: @Spikhalskiy don't we also need to make sure its positive here ?

Spikhalskiy: No, shiftSize now returns number in [0, numAddresses) interval.

guidomedina: The name is misleading, should be called something like `randomIndex`, I honestly don't understand why so much waste of CPU on generating a random number, the other solution is so much simple and better.

Spikhalskiy: @guidomedina another solution could "can cause stuck into static resolution in some setups." Scott even provided an example here https://github.com/netty/netty/commit/5eebe9a06c6873e797b5f8c847086c35ae990ef5#commitcomment-19776079 Better safe than sorry. And better ineffective solution than incorrect. Feel free to submit effective solution without this bug. I'm not ready right now.

Scottmitch: @Spikhalskiy - I agree with @guidomedina in that the method name is not very clear. can we rename `shiftSize` to `randomIndex` or something more representative of what it is doing?

guidomedina:
@Scottmitch
I think the following comment I made at the wrong place, WDTY of the following?

Another solution can be to have a set of sequencers and use the sequencer corresponding to the hashcode of the DNS address, the problem here is the size of set which could be a system property and default to 128 for example, the bigger the better to avoid hashes coalition:
```
public class Sequencer {
  final AtomicInteger sequencer = new AtomicInteger();

  public int nextIndex(int size) {
    final int index = sequencer.getAndIncrement() % size;
    return index < 0 ? index + size : index;
  }
}

static final Sequencer[] SEQUENCER = new Sequencer[128];

static {
  for (int i=0; i<SEQUENCER.length; i++) {
     SEQUENCER[i] = new Sequencer();
  }
}

// How to get the Sequencer corresponding to an address hashcode:
final int nextIndex = SEQUENCER[Math.abs(dnsAddress.hashCode() % SEQUENCER.length)].nextIndex(addresses.size());
```

Spikhalskiy:
@Scottmitch addressed, thanks.

normanmaurer:
@Spikhalskiy thanks looking good. Can you please update the commit message to match our template. After this I will pull in:

Scottmitch:
4.1 (4bce554)

Scottmitch:
Whoops sorry @normanmaurer ... I missed your comment and pulled in 😢 . I guess the change is relatively small so the comment should be sufficient.

Thanks @Spikhalskiy !

PR #6009, opened 2016-11-11
Restructuring SslHandler to support new engines

nmittler:
Motivation:

In preparation for support of Conscrypt, I'm consolidating all of the engine-specific details so that it's easier to add new engine types that affect the behavior of SslHandler.

Modifications:

Added an enum SslEngineType that provides SSL engine-specific details.

Result:

SslHandler is more extensible for other engine types.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -183,7 +183,7 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
         TCNATIVE(true, COMPOSITE_CUMULATOR) {
             @Override
             SSLEngineResult unwrap(SslHandler handler, ByteBuf in, int readerIndex, int len, ByteBuf out)
-                throws SSLException {
+                    throws SSLException {
                 int nioBufferCount = in.nioBufferCount();
                 int writerIndex = out.writerIndex();
                 final SSLEngineResult result;
@@ -197,8 +197,7 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
                     try {
                         handler.singleBuffer[0] = toByteBuffer(out, writerIndex,
                             out.writableBytes());
-                        result = opensslEngine
-                            .unwrap(in.nioBuffers(readerIndex, len), handler.singleBuffer);
+                        result = opensslEngine.unwrap(in.nioBuffers(readerIndex, len), handler.singleBuffer);
                         out.writerIndex(writerIndex + result.bytesProduced());
                     } finally {
                         handler.singleBuffer[0] = null;
@@ -213,9 +212,8 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
         },
         JDK(false, MERGE_CUMULATOR) {
             @Override
-            SSLEngineResult unwrap(SslHandler handler, ByteBuf in,
-                int readerIndex, int len,
-                ByteBuf out) throws SSLException {
+            SSLEngineResult unwrap(SslHandler handler, ByteBuf in, int readerIndex, int len, ByteBuf out)
+                    throws SSLException {
                 int writerIndex = out.writerIndex();
                 final SSLEngineResult result = handler.engine.unwrap(toByteBuffer(in, readerIndex, len),
                     toByteBuffer(out, writerIndex, out.writableBytes()));
@@ -236,8 +234,8 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
             this.cumulator = cumulator;
         }
 
-        abstract SSLEngineResult unwrap(SslHandler handler,
-            ByteBuf in, int readerIndex, int len, ByteBuf out) throws SSLException;
+        abstract SSLEngineResult unwrap(SslHandler handler, ByteBuf in, int readerIndex, int len, ByteBuf out)
+                throws SSLException;
 
         // BEGIN Platform-dependent flags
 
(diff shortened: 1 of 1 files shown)


nmittler:
@normanmaurer can you take a look?

normanmaurer:
Sure thing... still on my way back from Devoxx but will do

nmittler:
Ok great ... this is a pretty easy one.  Just moving code around a bit ... no functionality changes.

normanmaurer:
@nmittler actually I am not sure why we even need this ... We should just handle all the SSLEngine implementations the same way and only do optimisations  for "known" ones like the OpenSslEngine which is part of Netty itself. So I would prefer to not pull this in

nmittler:
@normanmaurer I'm not sure I understand your point. The point of this PR is to consolidate the code for "known" engines into one place rather than having little hacks (for OpenSSL or coming soon ... conscrypt) scattered throughout the code, as is done in these places:

https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/SslHandler.java#L277
https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/SslHandler.java#L287
https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/SslHandler.java#L1073

Consolidating the logic into the enum makes it very clear upfront as to how SslHandler handles the various known engines differently. And it also gets rid of extra conditionals at runtime.

normanmaurer:
@nmittler actually never mind after talking with you on chat I now understand what you are doing... Ship it!

normanmaurer:
@Scottmitch you want to check as well ?

nmittler:
@Scottmitch gentle ping :)

nmittler:
@normanmaurer I can't imagine the resource leak was caused by this PR ... can you confirm?

Scottmitch:
@nmittler - Leak looks related to https://github.com/netty/netty/issues/5850#issuecomment-261073798

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -179,8 +179,89 @@
...
+                        toByteBuffer(out, writerIndex, out.writableBytes()));
+                }
+                out.writerIndex(writerIndex + result.bytesProduced());
+                return result;
+            }
+        },
+        JDK(false, MERGE_CUMULATOR) {
+            @Override
+            SSLEngineResult unwrap(SslHandler handler, ByteBuf in,
+                int readerIndex, int len,
```
Scottmitch: nit: can you align the parameters ... was a bit hard to look at on first glance

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -179,8 +179,89 @@
...
+                     * If {@link OpenSslEngine} is in use,
+                     * we can use a special {@link OpenSslEngine#unwrap(ByteBuffer[], ByteBuffer[])} method
+                     * that accepts multiple {@link ByteBuffer}s without additional memory copies.
+                     */
+                    OpenSslEngine opensslEngine = (OpenSslEngine) handler.engine;
+                    try {
+                        handler.singleBuffer[0] = toByteBuffer(out, writerIndex,
+                            out.writableBytes());
+                        result = opensslEngine
+                            .unwrap(in.nioBuffers(readerIndex, len), handler.singleBuffer);
```
Scottmitch: nit: put this on the previous line?

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -179,8 +179,89 @@
...
+            return JDK;
+        }
+
+        SslEngineType(boolean wantsDirectBuffer, Cumulator cumulator) {
+            this.wantsDirectBuffer = wantsDirectBuffer;
+            this.cumulator = cumulator;
+        }
+
+        abstract SSLEngineResult unwrap(SslHandler handler,
+            ByteBuf in, int readerIndex, int len, ByteBuf out) throws SSLException;
```
Scottmitch: nit: fix formatting here too

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
nmittler: done

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
nmittler: done.

nmittler:
@Scottmitch PTAL

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -179,8 +179,87 @@
...
+            }
+            return JDK;
+        }
+
+        SslEngineType(boolean wantsDirectBuffer, Cumulator cumulator) {
+            this.wantsDirectBuffer = wantsDirectBuffer;
+            this.cumulator = cumulator;
+        }
+
+        abstract SSLEngineResult unwrap(SslHandler handler, ByteBuf in, int readerIndex, int len, ByteBuf out)
```
normanmaurer: @nmittler just one more comment... what you think about also moving the wrap(...) method in there just to make it consistent ?

nmittler: @normanmaurer I thought about it, but moving wrap into the enum isn't really as helpful. The only parameter to the process is `wantsDirectBuffer` and that's already in the enum. Let's leave it as-is for now. If we find a need to move wrap later, we can. Sound good?

normanmaurer: ok... please merge :)

normanmaurer:
@nmittler I squashed and cherry-picked for you into 4.1 (9725a4d004f10dc60d0406c197e1583f91461341) and 4.0 (5a8a365d1740234d6942a218edf5d79f1d8da800)

nmittler:
@normanmaurer great! Thanks, buddy!

Issue #5976, opened 2016-11-03
Openssl SSLEngine reports too little size in session.getPacketBufferSize


jrudolph:
It seems OpenSSLEngine needs more package buffer space than the JDK one. Both JDK and Openssl report for `session.getPacketBufferSize` 18713 bytes (16665 + 2048) but the OpenSSLEngine will in some cases only succeed with buffers of size 16665 + 2048 + 58 (I bisected the exact amount, 1 byte less will still fail) and will otherwise return a BUFFER_OVERFLOW. 

This isn't a fatal problem as you can just provide more buffer space and call again. However, 
we relied on the invariant that a call to `wrap` can never result with BUFFER_OVERFLOW if the destination buffer was at least of size `session.getPacketBufferSize`. This allows us not to care for BUFFER_OVERFLOWs at all using the JDK SSLEngine. Not sure if that invariant really exists, though, so, one could say that this is just another degree of freedom the SSLEngine might leave to implementors.

normanmaurer:
@jrudolph thats interesting... I wonder where the 58 bytes come from. Any idea ?

normanmaurer:
@jrudolph also do you have a reproducer for this as well ? Would be interested to see when this happens.

normanmaurer:
@jrudolph ping...

jrudolph:
@normanmaurer sorry for the delay... Here's a reproducer: https://github.com/netty/netty/compare/4.1...jrudolph:5976-test-case

I couldn't reproduce the number above exactly. Instead, I found that the number of output buffer space depends on the number of input bytes. I guess the reason is that the openssl engine tries to output all outstanding data for several frames in one go.

normanmaurer:
@jrudolph thanks a lot! I will open a PR with a fix for the other reported issue today. Maybe this will also fix this one :) Can you please confirm you grant me the right to include your test-code and so publish it under the ASL2 license as part of netty ?

jrudolph:
Yes, I confirm that you can take the code under ASL2 license.

normanmaurer:
[...]

PR #6010, opened 2016-11-11
[#5976] Ensure we only consume as much data on wrap(...) as we can ha…

normanmaurer:
…ndle.

Motiviation:

We need to ensure we only consume as much da as we can maximal put in one ssl record to not produce a BUFFER_OVERFLOW when calling wrap(...).

Modification:

- Limit the amount of data that we consume based on the maximal plain text size that can be put in one ssl record
- Add testcase to verify the fix
- Tighten up testcases to ensure the amount of produced and consumed data in SslEngineResult matches the buffers. If not the tests will fail now.

Result:

Correct and conform behavior of OpenSslEngine.wrap(...) and better test coverage during handshaking in general.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -379,10 +379,9 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
      *
      * Calling this function with src.remaining == 0 is undefined.
      */
-    private int writePlaintextData(final ByteBuffer src) {
+    private int writePlaintextData(final ByteBuffer src, int len) {
         final int pos = src.position();
         final int limit = src.limit();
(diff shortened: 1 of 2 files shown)


normanmaurer:
@Scottmitch @nmittler @jrudolph PTAL

normanmaurer:
Cherry-picked into 4.1 (c8575c42d0964812c0d2617f98efa7cb64c46be8) and 4.0 (595b2676be9fae5ead92263d0a8396929b3f88b5)

jrudolph:
:+1:

normanmaurer:
@jrudolph thanks! Let me know if you find any other issues.

Scottmitch:
@normanmaurer - did you run benchmarks for use cases this may impact so we can understand the impact?

normanmaurer:
@Scottmitch yes but I did not see any impact even if I wrote 32k buffers.

Scottmitch:
@normanmaurer - great thanks!

Issue #5971, opened 2016-11-02
Building 4.0 branch failures in SniHandlerTest.testServerNameParsing:113


garydgregory:
Building the 4.0 branch with:
```
Apache Maven 3.2.5 (12a6b3acb947671f09b81f49094c53f426d8cea1; 2014-12-14T09:29:23-08:00)
Maven home: E:\Java\apache-maven-3.2.5
Java version: 1.7.0_79, vendor: Oracle Corporation
Java home: C:\Program Files\Java\jdk1.7.0_79\jre
Default locale: en_US, platform encoding: Cp1252
OS name: "windows 7", version: "6.1", arch: "amd64", family: "windows"
```
I get:

```
Tests run: 1, Failures: 0, Errors: 0, Skipped: 1, Time elapsed: 0.003 sec - in io.netty.handler.ssl.OpenSslClientContextTest
Running io.netty.handler.ssl.JdkSslEngineTest
10:32:58.829 [nioEventLoopGroup-9-1] DEBUG i.n.u.i.JavassistTypeParameterMatcherGenerator - Generated: io.netty.util.internal.__matchers__.io.netty.buffer.ByteBufMatcher
10:32:58.847 [nioEventLoopGroup-9-1] DEBUG i.n.h.s.u.InsecureTrustManagerFactory - Accepting a server certificate: CN=example.com
10:32:58.855 [nioEventLoopGroup-8-1] DEBUG io.netty.handler.ssl.SslHandler - [id: 0xd6fffb30, L:/10.18.99.172:65375 - R:/10.18.99.172:65376] HANDSHAKEN: TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA
10:32:58.856 [nioEventLoopGroup-9-1] DEBUG io.netty.handler.ssl.SslHandler - [id: 0xcd1ed556, L:/10.18.99.172:65376 - R:/0:0:0:0:0:0:0:0:65375] HANDSHAKEN: TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA
10:32:59.316 [nioEventLoopGroup-12-1] DEBUG i.n.h.s.u.InsecureTrustManagerFactory - Accepting a server certificate: CN=example.com
[...]

PR #6011, opened 2016-11-11
[#5874] [#5971] Ensure SniHandlerTest.testServerNameParsing not fails…

normanmaurer:
… with SslProvider.JDK

Motivation:

The SniHandlerTest.testServerNameParsing did fail when SslProvider.JDK was used as it the JDK SSLEngineImpl does not send an alert.

Modifications:

Ensure tests pass with JDK and OPENSSL ssl implementations.

Result:

SniHandlerTest will run with all SslProvider and not fail when SslProvider.JDK is used.

### DIFF
--- a/handler/src/test/java/io/netty/handler/ssl/SniHandlerTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/SniHandlerTest.java
@@ -26,6 +26,8 @@ import static org.junit.Assume.assumeTrue;
 
 import java.io.File;
 import java.net.InetSocketAddress;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 
@@ -62,7 +64,10 @@ import io.netty.util.ReferenceCountUtil;
 import io.netty.util.ReferenceCounted;
 import io.netty.util.concurrent.Promise;
 import io.netty.util.internal.ObjectUtil;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
 
+@RunWith(Parameterized.class)
 public class SniHandlerTest {
 
     private static ApplicationProtocolConfig newApnConfig() {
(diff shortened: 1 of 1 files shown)


normanmaurer:
@Scottmitch @nmittler PTAL

Scottmitch:
Is it not a problem that the OpenSslEngine and JdkSslEngine have different behaviors? I looked at this briefly before but I didn't have time to determine if the way were were handling errors was different for the two engines, if the error should have originated from SSLEngineImpl and didn't, or if the spec allows flexibility here for an alert not to be generated.

handler/src/test/java/io/netty/handler/ssl/SniHandlerTest.java
```diff
@@ -75,292 +80,348 @@ private static ApplicationProtocolConfig newApnConfig() {
...
         File keyFile = new File(SniHandlerTest.class.getResource("test_encrypted.pem").getFile());
         File crtFile = new File(SniHandlerTest.class.getResource("test.crt").getFile());
 
-        return SslContextBuilder.forServer(crtFile, keyFile, "12345")
-                .sslProvider(provider)
-                .applicationProtocolConfig(newApnConfig()).build();
+        SslContextBuilder sslCtxBuilder = SslContextBuilder.forServer(crtFile, keyFile, "12345")
+                .sslProvider(provider);
+        if (apn) {
+            sslCtxBuilder.applicationProtocolConfig(newApnConfig());
```
Scottmitch: should we do the same `assumeTrue` checks as we do in `makeSslContext` here?

handler/src/test/java/io/netty/handler/ssl/SniHandlerTest.java
```diff
@@ -75,292 +80,348 @@ private static ApplicationProtocolConfig newApnConfig() {
...
+        SslContextBuilder sslCtxBuilder = SslContextBuilder.forClient().trustManager(crtFile).sslProvider(provider);
+        if (apn) {
+            sslCtxBuilder.applicationProtocolConfig(newApnConfig());
+        }
+        return sslCtxBuilder.build();
+    }
+
+    @Parameterized.Parameters(name = "{index}: sslProvider={0}")
+    public static Iterable<? extends Object> data() {
+        List<SslProvider> params = new ArrayList<SslProvider>();
```
Scottmitch: initial size of 3?

handler/src/test/java/io/netty/handler/ssl/SniHandlerTest.java
```diff
@@ -75,292 +80,348 @@ private static ApplicationProtocolConfig newApnConfig() {
...
+                        sc.close().syncUninterruptibly();
+                    }
+                    if (sslContext != null) {
+                        ReferenceCountUtil.release(sslContext);
+                    }
+                    group.shutdownGracefully();
 
-            cert.delete();
+                    cert.delete();
+                }
```
Scottmitch: add a `case JDK: return;` and `case default: throw new Error();`?

handler/src/test/java/io/netty/handler/ssl/SniHandlerTest.java
normanmaurer: sure I can do... but honestly I don't think it matters... Its a unit test.

handler/src/test/java/io/netty/handler/ssl/SniHandlerTest.java
normanmaurer: yep why not.

normanmaurer:
@Scottmitch I think what we do in openssl is correct. So I think its either an SSLEngineImpl bug or its optional. So I would just not put an assert there.

handler/src/test/java/io/netty/handler/ssl/SniHandlerTest.java
normanmaurer: This is part of the `makSslContext` method and so it already has the guards.

normanmaurer:
Cherry-picked into 4.1 (86bbf24) and 4.0 (d80bd8fb4aa04490171e398b1cd17fbf0237815b)

PR #6018, opened 2016-11-14
Fix cookie date parsing, close #6016

### LINKED ISSUE #6016: ClientCookieDecoder doesn't support some valid RFC6265 expires fields
`ClientCookieDecoder` currently uses `HttpHeaderDateFormat` underneath, which has a fallback strategy and tries the following patterns:
* `E, dd MMM yyyy HH:mm:ss z`
* `EEEE, d-MMM-y HH:mm:ss z`
* `EEE, MMM d HH:mm:ss yyyy`

The former is more or less RFC1123 (it accepts full days of week, while RFC1123 requires 3 letters), I don't know where the 2 others come from.
RFC1123 is [the spec for headers such as `Expires` or `Date`](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html).

The problem is that RFC6265 defines [its own spec](https://tools.ietf.org/html/rfc6265#section-5.1.1) (shaking head) that supports RFC1123 format.

While most people stick to RFC1123, some don't.
For example, Github uses a thing of beauty such as `Sun, 27 Nov 2016 19:37:15 -0000` (`-0000` instead of `GMT` (banging head on desk).
Basically, the spirit is that:
* the day of week is ignored (completely redundant/useless)
* the timezone is ignored as RFC6265 enforces UTC
* delimiters can be many things, so it's not problem to write something like `27-Nov-2016`

Working on a fix.

slandelle:
Motivation:
* RFC6265 defines its own parser which is different from RFC1123 (it accepts RFC1123 format but also other ones). Basically, it's very lax on delimiters, ignores day of week and timezone. Currently, ClientCookieDecoder uses HttpHeaderDateFormat underneath, and can't parse valid cookies such as Github ones whose expires attribute looks like "Sun, 27 Nov 2016 19:37:15 -0000"
* ServerSideCookieEncoder currently uses HttpHeaderDateFormat underneath for formatting expires field, and it's slow.

Modifications:
* Introduce CookieDateFormatter that correctly implement RFC6265
* Use CookieDateFormatter in ClientCookieDecoder and ServerCookieEncoder

Result:
* Proper RFC6265 dates support
* Faster ServerCookieEncoder and ClientCookieDecoder

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormat.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormat.java
@@ -32,7 +32,9 @@ import java.util.TimeZone;
  * <li>Sunday, 06-Nov-94 08:49:37 GMT: obsolete specification</li>
  * <li>Sun Nov  6 08:49:37 1994: obsolete specification</li>
  * </ul>
+ * @deprecated Use {@link HttpHeaderDateFormatter} instead
  */
+@Deprecated
 public final class HttpHeaderDateFormat extends SimpleDateFormat {
     private static final long serialVersionUID = -925286159755905325L;
 
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
@@ -1,0 +1,443 @@
+/*
+ * Copyright 2016 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.codec.http;
+
+import static io.netty.util.internal.ObjectUtil.checkNotNull;
+
+import io.netty.util.AsciiString;
+import io.netty.util.concurrent.FastThreadLocal;
+
+import java.util.BitSet;
+import java.util.Calendar;
+import java.util.Date;
+import java.util.GregorianCalendar;
+import java.util.TimeZone;
+
+/**
+ * A formatter for HTTP header dates, such as "Expires" and "Date" headers, or "expires" field in "Set-Cookie".
+ *
+ * On the parsing side, it honors RFC6265 (so it supports RFC1123).
+ * Note that:
+ * <ul>
+ *     <li>Day of week is ignored and not validated</li>
+ *     <li>Timezone is ignored, as RFC6265 assumes UTC</li>
+ * </ul>
+ * If you're looking for a date format that validates day of week, or supports other timezones, consider using
+ * java.util.DateTimeFormatter.RFC_1123_DATE_TIME.
+ *
+ * On the formatting side, it uses RFC1123 format.
+ *
+ * @see <a href="https://tools.ietf.org/html/rfc6265#section-5.1.1">RFC6265</a> for the parsing side
+ * @see <a href="https://tools.ietf.org/html/rfc1123#page-55">RFC1123</a> for the encoding side.
+ */
+public final class HttpHeaderDateFormatter {
+
+    private static final BitSet DELIMITERS = new BitSet();
+    static {
+        DELIMITERS.set(0x09);
+        for (char c = 0x20; c <= 0x2F; c++) {
+            DELIMITERS.set(c);
+        }
+        for (char c = 0x3B; c <= 0x40; c++) {
+            DELIMITERS.set(c);
+        }
+        for (char c = 0x5B; c <= 0x60; c++) {
+            DELIMITERS.set(c);
+        }
+        for (char c = 0x7B; c <= 0x7E; c++) {
+            DELIMITERS.set(c);
+        }
+    }
+
+    private static final String[] DAY_OF_WEEK_TO_SHORT_NAME =
+            new String[]{"Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"};
+
+    private static final String[] CALENDAR_MONTH_TO_SHORT_NAME =
+            new String[]{"Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"};
+
+    private static final FastThreadLocal<HttpHeaderDateFormatter> INSTANCES =
+            new FastThreadLocal<HttpHeaderDateFormatter>() {
+                @Override
+                protected HttpHeaderDateFormatter initialValue() {
+                    return new HttpHeaderDateFormatter();
+                }
+            };
+
+    /**
+     * Parse some text into a {@link Date}, according to RFC6265
+     * @param txt text to parse
+     * @return a {@link Date}, or null if text couldn't be parsed
+     */
+    public static Date parse(CharSequence txt) {
+        return parse(txt, 0, txt.length());
+    }
+
+    /**
+     * Parse some text into a {@link Date}, according to RFC6265
+     * @param txt text to parse
+     * @param start the start index inside <code>txt</code>
+     * @param end the end index inside <code>txt</code>
+     * @return a {@link Date}, or null if text couldn't be parsed
+     */
+    public static Date parse(CharSequence txt, int start, int end) {
+        int length = end - start;
+        if (length == 0) {
+            return null;
+        } else if (length < 0) {
+            throw new IllegalArgumentException("Can't have end < start");
+        } else if (length > 64) {
+            throw new IllegalArgumentException("Can't parse more than 64 chars," +
+                    "looks like a user error or a malformed header");
+        }
+        return formatter().parse0(checkNotNull(txt, "txt"), start, end);
+    }
+
+    /**
+     * Format a {@link Date} into RFC1123 format
+     * @param date the date to format
+     * @return a RFC1123 string
+     */
+    public static String format(Date date) {
+        return formatter().format0(checkNotNull(date, "date"));
+    }
+
+    /**
+     * Append a {@link Date} to a {@link StringBuilder} into RFC1123 format
+     * @param date the date to format
+     * @param sb the StringBuilder
+     * @return the same StringBuilder
+     */
+    public static StringBuilder append(Date date, StringBuilder sb) {
+        return formatter().append0(checkNotNull(date, "date"), checkNotNull(sb, "sb"));
+    }
+
+    private static HttpHeaderDateFormatter formatter() {
+        HttpHeaderDateFormatter formatter = INSTANCES.get();
+        formatter.reset();
+        return formatter;
+    }
+
+    // delimiter = %x09 / %x20-2F / %x3B-40 / %x5B-60 / %x7B-7E
+    private static boolean isDelim(char c) {
+        return DELIMITERS.get(c);
+    }
+
+    private static boolean isDigit(char c) {
+        return c >= 48 && c <= 57;
+    }
+
+    private static int getNumericalValue(char c) {
+        return c - 48;
+    }
+
+    private final GregorianCalendar cal = new GregorianCalendar(TimeZone.getTimeZone("UTC"));
+    private final StringBuilder sb = new StringBuilder(29); // Sun, 27 Nov 2016 19:37:15 GMT
+    private boolean timeFound;
+    private int hours;
+    private int minutes;
+    private int seconds;
+    private boolean dayOfMonthFound;
+    private int dayOfMonth;
+    private boolean monthFound;
+    private int month;
+    private boolean yearFound;
+    private int year;
+
+    private HttpHeaderDateFormatter() {
+        reset();
+    }
+
+    public void reset() {
+        timeFound = false;
+        hours = -1;
+        minutes = -1;
+        seconds = -1;
+        dayOfMonthFound = false;
+        dayOfMonth = -1;
+        monthFound = false;
+        month = -1;
+        yearFound = false;
+        year = -1;
+        cal.set(Calendar.MILLISECOND, 0);
+        sb.setLength(0);
+    }
+
+    private boolean tryParseTime(CharSequence txt, int tokenStart, int tokenEnd) {
+        int len = tokenEnd - tokenStart;
+
+        // h:m:s to hh:mm:ss
+        if (len < 5 || len > 8) {
+            return false;
+        }
+
+        int localHours = -1;
+        int localMinutes = -1;
+        int localSeconds = -1;
+        int currentPartNumber = 0;
+        int currentPartValue = 0;
+
+        for (int i = tokenStart; i < tokenEnd; i++) {
+            char c = txt.charAt(i);
+            if (isDigit(c)) {
+                currentPartValue = currentPartValue * 10 + getNumericalValue(c);
+            } else if (c == ':') {
+                if (currentPartValue == 0) {
+                    // invalid :: (nothing in between)
+                    return false;
+                }
+                switch (currentPartNumber) {
+                    case 0:
+                        // flushing hours
+                        localHours = currentPartValue;
+                        currentPartValue = 0;
+                        currentPartNumber++;
+                        break;
+                    case 1:
+                        // flushing minutes
+                        localMinutes = currentPartValue;
+                        currentPartValue = 0;
+                        currentPartNumber++;
+                        break;
+                    default:
+                        // invalid, too many :
+                        return false;
+                }
+            } else {
+                // invalid char
+                return false;
+            }
+        }
+
+        if (currentPartValue > 0) {
+            // pending seconds
+            localSeconds = currentPartValue;
+        }
+
+        if (localHours >= 0 && localMinutes >= 0 && localSeconds >= 0) {
+            hours = localHours;
+            minutes = localMinutes;
+            seconds = localSeconds;
+            return true;
+        }
+
+        return false;
+    }
+
+    private boolean tryParseDayOfMonth(CharSequence txt, int tokenStart, int tokenEnd) {
+        int len = tokenEnd - tokenStart;
+
+        if (len == 1) {
+            char c0 = txt.charAt(tokenStart);
+            if (isDigit(c0)) {
+                dayOfMonth = getNumericalValue(c0);
+                return true;
+            }
+
+        } else if (len == 2) {
+            char c0 = txt.charAt(tokenStart);
+            char c1 = txt.charAt(tokenStart + 1);
+            if (isDigit(c0) && isDigit(c1)) {
+                dayOfMonth = getNumericalValue(c0) * 10 + getNumericalValue(c1);
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    private static boolean matchMonth(String month, CharSequence txt, int tokenStart) {
+        return AsciiString.regionMatchesAscii(month, true, 0, txt, tokenStart, 3);
+    }
+
+    private boolean tryParseMonth(CharSequence txt, int tokenStart, int tokenEnd) {
+        int len = tokenEnd - tokenStart;
+
+        if (len != 3) {
+            return false;
+        }
+
+        if (matchMonth("Jan", txt, tokenStart)) {
+            month = Calendar.JANUARY;
+        } else if (matchMonth("Feb", txt, tokenStart)) {
+            month = Calendar.FEBRUARY;
+        } else if (matchMonth("Mar", txt, tokenStart)) {
+            month = Calendar.MARCH;
+        } else if (matchMonth("Apr", txt, tokenStart)) {
+            month = Calendar.APRIL;
+        } else if (matchMonth("May", txt, tokenStart)) {
+            month = Calendar.MAY;
+        } else if (matchMonth("Jun", txt, tokenStart)) {
+            month = Calendar.JUNE;
+        } else if (matchMonth("Jul", txt, tokenStart)) {
+            month = Calendar.JULY;
+        } else if (matchMonth("Aug", txt, tokenStart)) {
+            month = Calendar.AUGUST;
+        } else if (matchMonth("Sep", txt, tokenStart)) {
+            month = Calendar.SEPTEMBER;
+        } else if (matchMonth("Oct", txt, tokenStart)) {
+            month = Calendar.OCTOBER;
+        } else if (matchMonth("Nov", txt, tokenStart)) {
+            month = Calendar.NOVEMBER;
+        } else if (matchMonth("Dec", txt, tokenStart)) {
+            month = Calendar.DECEMBER;
+        } else {
+            return false;
+        }
+
+        return true;
+    }
+
+    private boolean tryParseYear(CharSequence txt, int tokenStart, int tokenEnd) {
+        int len = tokenEnd - tokenStart;
+
+        if (len == 2) {
+            char c0 = txt.charAt(tokenStart);
+            char c1 = txt.charAt(tokenStart + 1);
+            if (isDigit(c0) && isDigit(c1)) {
+                year = getNumericalValue(c0) * 10 + getNumericalValue(c1);
+                return true;
+            }
+
+        } else if (len == 4) {
+            char c0 = txt.charAt(tokenStart);
+            char c1 = txt.charAt(tokenStart + 1);
+            char c2 = txt.charAt(tokenStart + 2);
+            char c3 = txt.charAt(tokenStart + 3);
+            if (isDigit(c0) && isDigit(c1) && isDigit(c2) && isDigit(c3)) {
+                year = getNumericalValue(c0) * 1000 +
+                        getNumericalValue(c1) * 100 +
+                        getNumericalValue(c2) * 10 +
+                        getNumericalValue(c3);
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    private boolean parseToken(CharSequence txt, int tokenStart, int tokenEnd) {
+        // return true if all parts are found
+        if (!timeFound) {
+            timeFound = tryParseTime(txt, tokenStart, tokenEnd);
+            if (timeFound) {
+                return dayOfMonthFound && monthFound && yearFound;
+            }
+        }
+
+        if (!dayOfMonthFound) {
+            dayOfMonthFound = tryParseDayOfMonth(txt, tokenStart, tokenEnd);
+            if (dayOfMonthFound) {
+                return timeFound && monthFound && yearFound;
+            }
+        }
+
+        if (!monthFound) {
+            monthFound = tryParseMonth(txt, tokenStart, tokenEnd);
+            if (monthFound) {
+                return timeFound && dayOfMonthFound && yearFound;
+            }
+        }
+
+        if (!yearFound) {
+            yearFound = tryParseYear(txt, tokenStart, tokenEnd);
+        }
+        return timeFound && dayOfMonthFound && monthFound && yearFound;
+    }
+
+    private Date parse0(CharSequence txt, int start, int end) {
+        boolean allPartsFound = parse1(txt, start, end);
+        return allPartsFound && normalizeAndValidate() ? computeDate() : null;
+    }
+
+    private boolean parse1(CharSequence txt, int start, int end) {
+        // return true if all parts are found
+        int tokenStart = -1;
+
+        for (int i = start; i < end; i++) {
+            char c = txt.charAt(i);
+
+            if (isDelim(c)) {
+                if (tokenStart != -1) {
+                    // terminate token
+                    if (parseToken(txt, tokenStart, i)) {
+                        return true;
+                    }
+                    tokenStart = -1;
+                }
+            } else if (tokenStart == -1) {
+                // start new token
+                tokenStart = i;
+            }
+        }
+
+        // terminate trailing token
+        return tokenStart != -1 && parseToken(txt, tokenStart, txt.length());
+    }
+
+    private boolean normalizeAndValidate() {
+        if (dayOfMonth < 1
+                || dayOfMonth > 31
+                || hours > 23
+                || minutes > 59
+                || seconds > 59) {
+            return false;
+        }
+
+        if (year >= 70 && year <= 99) {
+            year += 1900;
+        } else if (year >= 0 && year < 70) {
+            year += 2000;
+        } else if (year < 1601) {
+            // invalid value
+            return false;
+        }
+        return true;
+    }
+
+    private Date computeDate() {
+        cal.set(Calendar.DAY_OF_MONTH, dayOfMonth);
+        cal.set(Calendar.MONTH, month);
+        cal.set(Calendar.YEAR, year);
+        cal.set(Calendar.HOUR_OF_DAY, hours);
+        cal.set(Calendar.MINUTE, minutes);
+        cal.set(Calendar.SECOND, seconds);
+        return cal.getTime();
+    }
+
+    private String format0(Date date) {
+        append0(date, sb);
+        return sb.toString();
+    }
+
+    private StringBuilder append0(Date date, StringBuilder sb) {
+        cal.setTime(date);
+
+        sb.append(DAY_OF_WEEK_TO_SHORT_NAME[cal.get(Calendar.DAY_OF_WEEK) - 1]).append(", ");
+        sb.append(cal.get(Calendar.DAY_OF_MONTH)).append(' ');
+        sb.append(CALENDAR_MONTH_TO_SHORT_NAME[cal.get(Calendar.MONTH)]).append(' ');
+        sb.append(cal.get(Calendar.YEAR)).append(' ');
+        appendZeroLeftPadded(cal.get(Calendar.HOUR_OF_DAY), sb).append(':');
+        appendZeroLeftPadded(cal.get(Calendar.MINUTE), sb).append(':');
+        return appendZeroLeftPadded(cal.get(Calendar.SECOND), sb).append(" GMT");
+    }
+
+    private static StringBuilder appendZeroLeftPadded(int value, StringBuilder sb) {
+        if (value < 10) {
+            sb.append('0');
+        }
+        return sb.append(value);
+    }
+}
--- a/codec-http/src/main/java/io/netty/handler/codec/http/cookie/ClientCookieDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/cookie/ClientCookieDecoder.java
@@ -15,12 +15,11 @@
  */
 package io.netty.handler.codec.http.cookie;
 
-import static io.netty.util.internal.ObjectUtil.checkNotNull;
+import io.netty.handler.codec.http.HttpHeaderDateFormatter;
 
-import io.netty.handler.codec.http.HttpHeaderDateFormat;
-
-import java.text.ParsePosition;
 import java.util.Date;
+
+import static io.netty.util.internal.ObjectUtil.checkNotNull;
 
 /**
  * A <a href="http://tools.ietf.org/html/rfc6265">RFC6265</a> compliant cookie decoder to be used client side.
@@ -168,14 +167,11 @@ public final class ClientCookieDecoder extends CookieDecoder {
             // max age has precedence over expires
             if (maxAge != Long.MIN_VALUE) {
                 return maxAge;
-            } else {
-                String expires = computeValue(expiresStart, expiresEnd);
-                if (expires != null) {
-                    Date expiresDate = HttpHeaderDateFormat.get().parse(expires, new ParsePosition(0));
-                    if (expiresDate != null) {
-                        long maxAgeMillis = expiresDate.getTime() - System.currentTimeMillis();
-                        return maxAgeMillis / 1000 + (maxAgeMillis % 1000 != 0 ? 1 : 0);
-                    }
+            } else if (isValueDefined(expiresStart, expiresEnd)) {
+                Date expiresDate = HttpHeaderDateFormatter.parse(header, expiresStart, expiresEnd);
+                if (expiresDate != null) {
+                    long maxAgeMillis = expiresDate.getTime() - System.currentTimeMillis();
+                    return maxAgeMillis / 1000 + (maxAgeMillis % 1000 != 0 ? 1 : 0);
                 }
             }
             return Long.MIN_VALUE;
@@ -254,8 +250,12 @@ public final class ClientCookieDecoder extends CookieDecoder {
             }
         }
 
+        private static boolean isValueDefined(int valueStart, int valueEnd) {
+            return valueStart != -1 && valueStart != valueEnd;
+        }
+
         private String computeValue(int valueStart, int valueEnd) {
-            return valueStart == -1 || valueStart == valueEnd ? null : header.substring(valueStart, valueEnd);
+            return isValueDefined(valueStart, valueEnd) ? header.substring(valueStart, valueEnd) : null;
         }
     }
 }
--- a/codec-http/src/test/java/io/netty/handler/codec/http/HttpHeaderDateFormatterTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/HttpHeaderDateFormatterTest.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2016 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.codec.http;
+
+import org.junit.Test;
+
+import java.util.Date;
+
+import static org.junit.Assert.*;
+import static io.netty.handler.codec.http.HttpHeaderDateFormatter.*;
+
+public class HttpHeaderDateFormatterTest {
+    /**
+     * This date is set at "06 Nov 1994 08:49:37 GMT", from
+     * <a href="http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html">examples in RFC documentation</a>
+     */
+    private static final long TIMESTAMP = 784111777000L;
+    private static final Date DATE = new Date(TIMESTAMP);
+
+    @Test
+    public void testParseWithSingleDigitDay() {
+        assertEquals(DATE, parse("Sun, 6 Nov 1994 08:49:37 GMT"));
+    }
+
+    @Test
+    public void testParseWithDoubleDigitDay() {
+        assertEquals(DATE, parse("Sun, 06 Nov 1994 08:49:37 GMT"));
+    }
+
+    @Test
+    public void testParseWithDashSeparatorSingleDigitDay() {
+        assertEquals(DATE, parse("Sunday, 06-Nov-94 08:49:37 GMT"));
+    }
+
+    @Test
+    public void testParseWithSingleDoubleDigitDay() {
+        assertEquals(DATE, parse("Sunday, 6-Nov-94 08:49:37 GMT"));
+    }
+
+    @Test
+    public void testParseWithoutGMT() {
+        assertEquals(DATE, parse("Sun Nov 6 08:49:37 1994"));
+    }
+
+    @Test
+    public void testParseWithFunkyTimezone() {
+        assertEquals(DATE, parse("Sun Nov 6 08:49:37 1994 -0000"));
+    }
+
+    @Test
+    public void testParseWithSingleDigitHourMinutesAndSecond() {
+        assertEquals(DATE, parse("Sunday, 6-Nov-94 8:49:37 GMT"));
+    }
+
+    @Test
+    public void testParseWithSingleDigitTime() {
+        assertEquals(DATE, parse("Sunday, 6 Nov 1994 8:49:37 GMT"));
+
+        Date _08_09_37 = new Date(TIMESTAMP - 40 * 60 * 1000);
+        assertEquals(_08_09_37, parse("Sunday, 6 Nov 1994 8:9:37 GMT"));
+        assertEquals(_08_09_37, parse("Sunday, 6 Nov 1994 8:09:37 GMT"));
+
+        Date _08_09_07 = new Date(TIMESTAMP - (40 * 60 + 30) * 1000);
+        assertEquals(_08_09_07, parse("Sunday, 6 Nov 1994 8:9:7 GMT"));
+        assertEquals(_08_09_07, parse("Sunday, 6 Nov 1994 8:9:07 GMT"));
+    }
+
+    @Test
+    public void testParseInvalidInput() {
+        // missing field
+        assertNull(parse("Sun, Nov 1994 08:49:37 GMT"));
+        assertNull(parse("Sun, 6 1994 08:49:37 GMT"));
+        assertNull(parse("Sun, 6 Nov 08:49:37 GMT"));
+        assertNull(parse("Sun, 6 Nov 1994 :49:37 GMT"));
+        assertNull(parse("Sun, 6 Nov 1994 49:37 GMT"));
+        assertNull(parse("Sun, 6 Nov 1994 08::37 GMT"));
+        assertNull(parse("Sun, 6 Nov 1994 08:37 GMT"));
+        assertNull(parse("Sun, 6 Nov 1994 08:49: GMT"));
+        assertNull(parse("Sun, 6 Nov 1994 08:49 GMT"));
+        //invalid value
+        assertNull(parse("Sun, 6 FOO 1994 08:49:37 GMT"));
+        assertNull(parse("Sun, 36 Nov 1994 08:49:37 GMT"));
+        assertNull(parse("Sun, 6 Nov 1994 28:49:37 GMT"));
+        assertNull(parse("Sun, 6 Nov 1994 08:69:37 GMT"));
+        assertNull(parse("Sun, 6 Nov 1994 08:49:67 GMT"));
+    }
+
+    @Test
+    public void testFormat() {
+        assertEquals("Sun, 6 Nov 1994 08:49:37 GMT", format(DATE));
+    }
+}
--- a/codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieEncoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieEncoder.java
@@ -20,7 +20,9 @@ import static io.netty.handler.codec.http.cookie.CookieUtil.addQuoted;
 import static io.netty.handler.codec.http.cookie.CookieUtil.stringBuilder;
 import static io.netty.handler.codec.http.cookie.CookieUtil.stripTrailingSeparator;
 import static io.netty.util.internal.ObjectUtil.checkNotNull;
-import io.netty.handler.codec.http.HttpHeaderDateFormat;
+
+import io.netty.handler.codec.http.HttpHeaderDateFormatter;
+import io.netty.handler.codec.http.HttpConstants;
 import io.netty.handler.codec.http.HttpRequest;
 
 import java.util.ArrayList;
@@ -102,7 +104,11 @@ public final class ServerCookieEncoder extends CookieEncoder {
         if (cookie.maxAge() != Long.MIN_VALUE) {
             add(buf, CookieHeaderNames.MAX_AGE, cookie.maxAge());
             Date expires = new Date(cookie.maxAge() * 1000 + System.currentTimeMillis());
-            add(buf, CookieHeaderNames.EXPIRES, HttpHeaderDateFormat.get().format(expires));
+            buf.append(CookieHeaderNames.EXPIRES);
+            buf.append((char) HttpConstants.EQUALS);
+            HttpHeaderDateFormatter.append(expires, buf);
+            buf.append((char) HttpConstants.SEMICOLON);
+            buf.append((char) HttpConstants.SP);
         }
 
         if (cookie.path() != null) {
--- a/codec-http/src/test/java/io/netty/handler/codec/http/cookie/ClientCookieDecoderTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/cookie/ClientCookieDecoderTest.java
@@ -15,16 +15,8 @@
  */
 package io.netty.handler.codec.http.cookie;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
-
+import io.netty.handler.codec.http.HttpHeaderDateFormatter;
 import org.junit.Test;
-
-import io.netty.handler.codec.http.HttpHeaderDateFormat;
 
 import java.util.ArrayList;
 import java.util.Calendar;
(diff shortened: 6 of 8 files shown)


codec-http/src/test/java/io/netty/handler/codec/http/cookie/CookieDateFormatterTest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.junit.Test;
+
+import java.util.Date;
+
+import static org.junit.Assert.assertEquals;
+
+public class CookieDateFormatterTest {
+    /**
+     * This date is set at "06 Nov 1994 08:49:37 GMT" (same used in example in
+     * RFC documentation)
```
normanmaurer: Just use a `<a href=...`

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDateFormatter.java
```diff
@@ -0,0 +1,452 @@
...
+        }
+        for (char c = 0x5B; c <= 0x60; c++) {
+            DELIMITERS.set(c);
+        }
+        for (char c = 0x7B; c <= 0x7E; c++) {
+            DELIMITERS.set(c);
+        }
+    }
+
+    private static final ThreadLocal<CookieDateFormatter> INSTANCES = new ThreadLocal<CookieDateFormatter>() {
```
normanmaurer: Using a `FastThreadLocal` ?

slandelle:
Side comment: I'd be in favor of deprecating `HttpHeaderDateFormat`, renaming `CookieDateFormatter` and promoting it as a replacement:
* `HttpHeaderDateFormat` is slow. For Java 8 users, java.time is faster than SimpleDateFormat. `CookieDateFormatter` is way faster, both on parsing and formatting sides.
* `CookieDateFormatter` supports parsing RFC2311 dates, such as Expires and Date headers and formats in RFC2311.

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDateFormatter.java
```diff
@@ -0,0 +1,452 @@
...
+        hours = -1;
+        minutes = -1;
+        seconds = -1;
+        dayOfMonthFound = false;
+        dayOfMonth = -1;
+        monthFound = false;
+        month = -1;
+        yearFound = false;
+        year = -1;
+        cal.set(Calendar.MILLISECOND, 0);
```
normanmaurer: @slandelle shouldnt you also reset sb ?

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDateFormatter.java
```diff
@@ -0,0 +1,452 @@
...
+    }
+
+    private boolean tryParseTime(String txt, int tokenStart, int tokenEnd) {
```
normanmaurer: `String` -> `CharSequence` ?

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDateFormatter.java
```diff
@@ -0,0 +1,452 @@
...
+            hours = localHours;
+            minutes = localMinutes;
+            seconds = localSeconds;
+            return true;
+        }
+
+        return false;
+    }
+
+    private boolean tryParseDayOfMonth(String txt, int tokenStart, int tokenEnd) {
```
normanmaurer: `String` -> `CharSequence` ?

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDateFormatter.java
```diff
@@ -0,0 +1,452 @@
...
+            } else if (txt.regionMatches(true, tokenStart, "Dec", 0, 3)) {
+                month = Calendar.DECEMBER;
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    private boolean tryParseYear(String txt, int tokenStart, int tokenEnd) {
```
normanmaurer: `String` -> `CharSequence` ?

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDateFormatter.java
```diff
@@ -0,0 +1,452 @@
...
+ *     <li>Timezone is ignored, as RFC6265 assumes UTC</li>
+ * </ul>
+ *
+ * Also note that this class can be used to encode and format HTTP headers such as "Expires" and "Date"
+ * as it's way more efficient than HttpHeaderDateFormat.
+ *
+ * @see <a href="https://tools.ietf.org/html/rfc6265#section-5.1.1">RFC6265</a> for the parsing side
+ * @see <a href="https://tools.ietf.org/html/rfc1123#page-55">RFC1123</a> for the encoding side.
+ */
+public final class CookieDateFormatter {
```
normanmaurer: @slandelle should this be public or better package private ?

codec-http/src/test/java/io/netty/handler/codec/http/cookie/CookieDateFormatterTest.java
slandelle: Doh, I'll have to dig. I forked existing `io.netty.handler.codec.http.HttpHeaderDateFormatTest`.

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDateFormatter.java
slandelle: Agree. I do reset it in `format0` but it would be more consistent to do it in reset?

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDateFormatter.java
slandelle: makes sense

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDateFormatter.java
slandelle: makes sense

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDateFormatter.java
slandelle: makes sense

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDateFormatter.java
slandelle: Note that as CharSequence doesn't support regionMatches, I will have to inspect type, cast, and copy if it's not a String.

normanmaurer: I see... then keep it as `String`

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDateFormatter.java
normanmaurer: yes it has a small penalty but a huge win if you are on the EventLoop

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDateFormatter.java
normanmaurer: I see... public sounds fine then

normanmaurer:
@slandelle deprecation sound fine to me.

slandelle:
@normanmaurer Should I rename `CookieDateFormatter` into `HttpHeaderDateFormatter` and move it to parent `io.netty.handler.codec.http` package so it sits next to deprecated `HttpHeaderDateFormat`?

normanmaurer:
Yes

slandelle:
All requested changes done

normanmaurer:
@slandelle as you said its faster can you "prove" this with a jmh benchmark ?

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,489 @@
...
+        } else {
+            int end = Math.min(offset + other.length(), txt.length());
+            return txt.subSequence(offset, end).toString().equalsIgnoreCase(other);
+        }
+    }
+
+    private boolean tryParseMonth(CharSequence txt, int tokenStart, int tokenEnd) {
+        int len = tokenEnd - tokenStart;
+
+        if (len == 3) {
```
normanmaurer: Just call `String txtStr = txt.toString()` here and make `regionMatchesIgnoreCase` take a `String`. This way we only need to "convert" to a String one time in all cases.

slandelle: Sorry, I don't get you. If the CharSequence is a String, I don't have to copy anything. Otherwise, I just have to copy 3 chars instead of the whole header.

slandelle:
> as you said its faster can you "prove" this with a jmh benchmark ?

On its way

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
normanmaurer: And string.toString() is a noop.

slandelle:
benchmarks added.
results on my computer:

```
Benchmark                                  Mode  Cnt         Score        Error  Units
formatHttpHeaderDateFormat                thrpt   20   4526329,072 ± 469456,711  ops/s
formatHttpHeaderDateFormatter             thrpt   20  16355066,044 ± 943954,410  ops/s
parseHttpHeaderDateFormat                 thrpt   20    386189,902 ±  15081,814  ops/s
parseHttpHeaderDateFormatter              thrpt   20   5328386,627 ± 135844,563  ops/s
decodeCookieWithRfc1123ExpiresField(old)  thrpt   20    315294,657 ±  14478,838  ops/s
decodeCookieWithRfc1123ExpiresField(new)  thrpt   20   2523366,444 ±  70711,820  ops/s
```

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,482 @@
...
+            String tokenString = txt.subSequence(tokenStart, tokenEnd).toString();
+
+            if (tokenString.equalsIgnoreCase("Jan")) {
+                month = Calendar.JANUARY;
+                return true;
+            } else if (tokenString.equalsIgnoreCase("Feb")) {
```
netkins: MAJOR The Cyclomatic Complexity of this method "tryParseMonth" is 26 which is greater than 25 authorized. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AMethodCyclomaticComplexity)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,482 @@
...
+            DELIMITERS.set(c);
+        }
+        for (char c = 0x7B; c <= 0x7E; c++) {
+            DELIMITERS.set(c);
+        }
+    }
+
+    private static final FastThreadLocal<HttpHeaderDateFormatter> INSTANCES =
+            new FastThreadLocal<HttpHeaderDateFormatter>() {
+                protected HttpHeaderDateFormatter initialValue() {
```
netkins: MAJOR Add the "@Override" annotation above this method signature [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1161)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,482 @@
...
+        cal.setTime(date);
+
+        switch (cal.get(Calendar.DAY_OF_WEEK)) {
+            case Calendar.MONDAY:
+                sb.append("Mon, ");
+                break;
+            case Calendar.TUESDAY:
+                sb.append("Tue, ");
+                break;
+            case Calendar.WEDNESDAY:
```
netkins: MAJOR Add a default case to this switch. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3ASwitchLastCaseIsDefaultCheck)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,419 @@
...
+            dayOfMonth = localDayOfMonth;
+            return true;
+        }
+
+        return false;
+    }
+
+    private boolean tryParseMonth(CharSequence txt, int tokenStart, int tokenEnd) {
+        int len = tokenEnd - tokenStart;
+
```
netkins: MAJOR Move the "Feb" string literal on the left side of this string comparison. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1132)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,419 @@
...
+            DELIMITERS.set(c);
+        }
+        for (char c = 0x7B; c <= 0x7E; c++) {
+            DELIMITERS.set(c);
+        }
+    }
+
+    private static final String[] DAY_OF_WEEK_TO_SHORT_NAME =
+            new String[]{"Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"};
+
```
netkins: MAJOR Add the "@Override" annotation above this method signature [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1161)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,419 @@
...
+        return false;
+    }
+
+    private boolean tryParseMonth(CharSequence txt, int tokenStart, int tokenEnd) {
+        int len = tokenEnd - tokenStart;
+
+        if (len == 3) {
+            String tokenString = txt.subSequence(tokenStart, tokenEnd).toString();
+
```
netkins: MAJOR Move the "Mar" string literal on the left side of this string comparison. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1132)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,419 @@
...
+            if (tokenString.equalsIgnoreCase("Jan")) {
+                month = Calendar.JANUARY;
+            } else if (tokenString.equalsIgnoreCase("Feb")) {
```
netkins: MAJOR Move the "Apr" string literal on the left side of this string comparison. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1132)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,419 @@
...
+                month = Calendar.FEBRUARY;
+            } else if (tokenString.equalsIgnoreCase("Mar")) {
+                month = Calendar.MARCH;
```
netkins: MAJOR Move the "May" string literal on the left side of this string comparison. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1132)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,419 @@
...
+            } else if (tokenString.equalsIgnoreCase("Apr")) {
+                month = Calendar.APRIL;
+            } else if (tokenString.equalsIgnoreCase("May")) {
```
netkins: MAJOR Move the "Jun" string literal on the left side of this string comparison. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1132)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,419 @@
...
+                month = Calendar.MAY;
+            } else if (tokenString.equalsIgnoreCase("Jun")) {
+                month = Calendar.JUNE;
```
netkins: MAJOR Move the "Jul" string literal on the left side of this string comparison. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1132)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,419 @@
...
+            } else if (tokenString.equalsIgnoreCase("Jul")) {
+                month = Calendar.JULY;
+            } else if (tokenString.equalsIgnoreCase("Aug")) {
```
netkins: MAJOR Move the "Aug" string literal on the left side of this string comparison. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1132)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,419 @@
...
+            if (isDigit(c)) {
+                localDayOfMonth = localDayOfMonth * 10 + getNumericalValue(c);
+            } else {
+                // invalid
+                return false;
+            }
+        }
+
+        if (localDayOfMonth > 0) {
+            dayOfMonth = localDayOfMonth;
```
netkins: MAJOR The Cyclomatic Complexity of this method "tryParseMonth" is 26 which is greater than 25 authorized. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AMethodCyclomaticComplexity)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,419 @@
...
+            } else if (tokenString.equalsIgnoreCase("Aug")) {
+                month = Calendar.AUGUST;
+            } else if (tokenString.equalsIgnoreCase("Sep")) {
+                month = Calendar.SEPTEMBER;
+            } else if (tokenString.equalsIgnoreCase("Oct")) {
+                month = Calendar.OCTOBER;
+            } else if (tokenString.equalsIgnoreCase("Nov")) {
+                month = Calendar.NOVEMBER;
+            } else if (tokenString.equalsIgnoreCase("Dec")) {
+                month = Calendar.DECEMBER;
```
netkins: MAJOR Move the "Nov" string literal on the left side of this string comparison. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1132)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,419 @@
...
+        }
+
+        if (localDayOfMonth > 0) {
+            dayOfMonth = localDayOfMonth;
+            return true;
+        }
+
+        return false;
+    }
+
```
netkins: MAJOR Move the "Jan" string literal on the left side of this string comparison. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1132)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,419 @@
...
+                month = Calendar.SEPTEMBER;
+            } else if (tokenString.equalsIgnoreCase("Oct")) {
+                month = Calendar.OCTOBER;
+            } else if (tokenString.equalsIgnoreCase("Nov")) {
+                month = Calendar.NOVEMBER;
+            } else if (tokenString.equalsIgnoreCase("Dec")) {
+                month = Calendar.DECEMBER;
+            }
+        }
+
```
netkins: MAJOR Move the "Dec" string literal on the left side of this string comparison. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1132)

codec-http/src/main/java/io/netty/handler/codec/http/cookie/ClientCookieDecoder.java
```diff
@@ -254,8 +250,12 @@ private void parse8(int nameStart, int valueStart, int valueEnd) {
             }
         }
 
+        private static boolean isValueDefined(int valueStart, int valueEnd) {
```
netkins: MINOR Make "isValueDefined" a "static" method. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2325)

slandelle:
@normanmaurer I've already addressed most Sonar comments by the time CI completed. There's still some left, like do you really want Yoda style comparison, eg `"Jan".equalsIgnoreCase(tokenString)`?

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,419 @@
...
+                month = Calendar.FEBRUARY;
+            } else if (tokenString.equalsIgnoreCase("Mar")) {
+                month = Calendar.MARCH;
+            } else if (tokenString.equalsIgnoreCase("Apr")) {
+                month = Calendar.APRIL;
+            } else if (tokenString.equalsIgnoreCase("May")) {
+                month = Calendar.MAY;
+            } else if (tokenString.equalsIgnoreCase("Jun")) {
+                month = Calendar.JUNE;
+            } else if (tokenString.equalsIgnoreCase("Jul")) {
```
netkins: MAJOR Move the "Oct" string literal on the left side of this string comparison. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1132)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,419 @@
...
+        }
+
+        return false;
+    }
+
+    private boolean tryParseMonth(CharSequence txt, int tokenStart, int tokenEnd) {
+        int len = tokenEnd - tokenStart;
+
+        if (len == 3) {
+            String tokenString = txt.subSequence(tokenStart, tokenEnd).toString();
```
netkins: MAJOR Move the "Mar" string literal on the left side of this string comparison. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1132)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,419 @@
...
+
+            if (tokenString.equalsIgnoreCase("Jan")) {
```
netkins: MAJOR Move the "Apr" string literal on the left side of this string comparison. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1132)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,419 @@
...
+                month = Calendar.APRIL;
+            } else if (tokenString.equalsIgnoreCase("May")) {
+                month = Calendar.MAY;
+            } else if (tokenString.equalsIgnoreCase("Jun")) {
+                month = Calendar.JUNE;
+            } else if (tokenString.equalsIgnoreCase("Jul")) {
+                month = Calendar.JULY;
+            } else if (tokenString.equalsIgnoreCase("Aug")) {
+                month = Calendar.AUGUST;
+            } else if (tokenString.equalsIgnoreCase("Sep")) {
```
netkins: MAJOR Move the "Dec" string literal on the left side of this string comparison. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1132)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,419 @@
...
+        int len = tokenEnd - tokenStart;
+
+        if (len == 3) {
+            String tokenString = txt.subSequence(tokenStart, tokenEnd).toString();
+
+            if (tokenString.equalsIgnoreCase("Jan")) {
+                month = Calendar.JANUARY;
+            } else if (tokenString.equalsIgnoreCase("Feb")) {
+                month = Calendar.FEBRUARY;
+            } else if (tokenString.equalsIgnoreCase("Mar")) {
```
netkins: MAJOR Move the "Jun" string literal on the left side of this string comparison. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1132)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,419 @@
...
+                month = Calendar.MARCH;
+            } else if (tokenString.equalsIgnoreCase("Apr")) {
+                month = Calendar.APRIL;
+            } else if (tokenString.equalsIgnoreCase("May")) {
+                month = Calendar.MAY;
+            } else if (tokenString.equalsIgnoreCase("Jun")) {
```
netkins: MAJOR Move the "Sep" string literal on the left side of this string comparison. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1132)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,419 @@
...
+                month = Calendar.JUNE;
+            } else if (tokenString.equalsIgnoreCase("Jul")) {
+                month = Calendar.JULY;
+            } else if (tokenString.equalsIgnoreCase("Aug")) {
+                month = Calendar.AUGUST;
+            } else if (tokenString.equalsIgnoreCase("Sep")) {
+                month = Calendar.SEPTEMBER;
+            } else if (tokenString.equalsIgnoreCase("Oct")) {
```
netkins: MAJOR Move the "Oct" string literal on the left side of this string comparison. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1132)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,419 @@
...
+                month = Calendar.OCTOBER;
+            } else if (tokenString.equalsIgnoreCase("Nov")) {
+                month = Calendar.NOVEMBER;
+            } else if (tokenString.equalsIgnoreCase("Dec")) {
```
netkins: MAJOR Move the "Dec" string literal on the left side of this string comparison. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1132)

normanmaurer:
@slandelle I don't care about the equals style..

normanmaurer:
@nmittler @Scottmitch PTAL as well

slandelle:
> I don't care about the equals style..

Argggg, just pushed after Sonar spammed me for the last 2 hrs...

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,419 @@
...
+        for (int i = tokenStart; i < tokenEnd; i++) {
+            char c = txt.charAt(i);
+            if (isDigit(c)) {
+                currentPartValue = currentPartValue * 10 + getNumericalValue(c);
+            } else if (c == ':') {
+                if (currentPartValue == 0) {
+                    // invalid :: (nothing in between)
+                    return false;
+
+                } else if (currentPartNumber == 0) {
```
normanmaurer: How about remove the else stuff and just use a switch statement ?

```java
switch(currentPartNumber) {
    case 0:
        ....
    case 1:
       ....
}
```

slandelle: I can. Just that from previous experience, switch tend to be slower.

normanmaurer:
LGTM... Waiting for @Scottmitch to take a pass as well.

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,421 @@
...
+        if (len != 1 && len != 2) {
+            return false;
+        }
+
+        int localDayOfMonth = 0;
+        for (int i = tokenStart; i < tokenEnd; i++) {
+            char c = txt.charAt(i);
+            if (isDigit(c)) {
+                localDayOfMonth = localDayOfMonth * 10 + getNumericalValue(c);
+            } else {
```
Scottmitch: nit: no need for `else`

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,421 @@
...
+            } else if ("Oct".equalsIgnoreCase(tokenString)) {
+                month = Calendar.OCTOBER;
+            } else if ("Nov".equalsIgnoreCase(tokenString)) {
+                month = Calendar.NOVEMBER;
+            } else if ("Dec".equalsIgnoreCase(tokenString)) {
+                month = Calendar.DECEMBER;
+            }
+        }
+
+        return month != -1;
```
Scottmitch: nit: you could just `return true;` here and add an `else` above which does `return false;` for one less conditional and not have to worry about `-1` being a special value. You could also have a `if (len != 3) { return false;` precondition check.

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
Scottmitch: you can also use `AsciiString.regionMatchesAscii`

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,421 @@
...
+    }
+
+    private boolean tryParseYear(CharSequence txt, int tokenStart, int tokenEnd) {
+        int len = tokenEnd - tokenStart;
+
+        if (len != 2 && len != 4) {
```
Scottmitch: the `len` can only be 2 values ... can we just unroll the loop below? This should remove some conditionals, a temp variable, and arithmetic.

```java
switch (len) {
  case 2:
   // do stuff
  case 4:
   // do stuff
  default:
    return false;
}
```

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,421 @@
...
+            char c = txt.charAt(i);
+
+            if (isDelim(c)) {
+                if (tokenStart != -1) {
+                    // terminate token
+                    parseToken(txt, tokenStart, i);
+                    tokenStart = -1;
+                }
+            } else {
+                if (tokenStart == -1) {
```
Scottmitch: nit: make this an `else if`

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,421 @@
...
+            // terminate trailing token
+            parseToken(txt, tokenStart, txt.length());
+        }
+
+        if (!timeFound || !dayOfMonthFound || !monthFound || !yearFound) {
+            // missing field
+            return null;
+        }
+
+         if (dayOfMonth < 1 || dayOfMonth > 31 || hours > 23 || minutes > 59 || seconds > 59) {
```
Scottmitch: can we make the above conditionals also represent "valid values"? For example can `timeFound` only be true if the value is valid?

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,421 @@
...
+            // invalid values
+            return null;
+        }
+
+        if (year >= 70 && year <= 99) {
+            year += 1900;
+        } else if (year >= 0 && year < 70) {
```
Scottmitch: is this a heuristic ... if so can we make this overridable?

codec-http/src/test/java/io/netty/handler/codec/http/HttpHeaderDateFormatterTest.java
```diff
@@ -0,0 +1,65 @@
...
+    @Test
+    public void testParseWithFunkyTimezone() {
+        assertEquals(DATE, HttpHeaderDateFormatter.parse("Sun Nov 6 08:49:37 1994 -0000"));
+    }
+
+    @Test
+    public void testFormat() {
+        assertEquals("Sun, 6 Nov 1994 08:49:37 GMT", HttpHeaderDateFormatter.format(DATE));
+    }
+}
```
Scottmitch: can we add some unit tests which validate the way we handle invalid input? For example testing the behavior in `tryParseTime` and related functions.

Scottmitch:
@slandelle - finished first pass. let me know if you have questions.

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,421 @@
...
+                }
+            };
+
+    /**
+     * Parse some text into a {@link Date}, according to RFC6265
+     * @param txt text to parse
+     * @return a {@link Date}, or null if text couldn't be parsed
+     */
+    public static Date parse(CharSequence txt) {
+        return parse(checkNotNull(txt, "txt"), 0, txt.length());
```
slandelle: ok

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,421 @@
...
+        return formatter;
+    }
+
+    // delimiter = %x09 / %x20-2F / %x3B-40 / %x5B-60 / %x7B-7E
+    private static boolean isDelim(char c) {
+        return DELIMITERS.get(c);
+    }
+
+    private static boolean isDigit(char c) {
+        return c >= 0x30 && c <= 0x39;
```
slandelle: ok

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,421 @@
...
+        yearFound = false;
+        year = -1;
+        cal.set(Calendar.MILLISECOND, 0);
+        sb.setLength(0);
+    }
+
+    private boolean tryParseTime(CharSequence txt, int tokenStart, int tokenEnd) {
+        int len = tokenEnd - tokenStart;
+
+        // d:m:yy to dd:mm:yyyy
```
slandelle: woops. Good catch!

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,421 @@
...
+
+        for (int i = tokenStart; i < tokenEnd; i++) {
+            char c = txt.charAt(i);
+            if (isDigit(c)) {
+                currentPartValue = currentPartValue * 10 + getNumericalValue(c);
+            } else if (c == ':') {
+                if (currentPartValue == 0) {
+                    // invalid :: (nothing in between)
+                    return false;
+                } else {
```
slandelle: ok

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,421 @@
...
+
+    private boolean tryParseDayOfMonth(CharSequence txt, int tokenStart, int tokenEnd) {
+        int len = tokenEnd - tokenStart;
+
+        if (len != 1 && len != 2) {
+            return false;
+        }
+
+        int localDayOfMonth = 0;
+        for (int i = tokenStart; i < tokenEnd; i++) {
```
slandelle: ok

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
slandelle: ok

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
slandelle: @Scottmitch Cool, wasn't aware of this one!

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
slandelle: ok

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
slandelle: ok. If you don't mind, I'd rather use `if-else` than `switch`, as variable names are not scoped in `case` branches (and I'd rather not do smart things with falling through cases as it would hurt readability).

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
slandelle: ok

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
slandelle: It's not a heuristic, it's the spec.

codec-http/src/test/java/io/netty/handler/codec/http/HttpHeaderDateFormatterTest.java
slandelle: omw

slandelle:
@Scottmitch Comments addressed. Please have a look at my comments.

slandelle:
Note: failing test is `Http2ConnectionRoundtripTest.flowControlProperlyChunksLargeMessage` which is not related to this PR.

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,439 @@
...
+                return;
+            }
+        }
+
+        if (!yearFound) {
+            yearFound = tryParseYear(txt, tokenStart, tokenEnd);
+        }
+    }
+
+    private Date parse0(CharSequence txt, int start, int end) {
```
netkins: MAJOR The Cyclomatic Complexity of this method "parse0" is 28 which is greater than 25 authorized. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AMethodCyclomaticComplexity)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,439 @@
...
+                tokenStart = i;
+            }
+        }
+
+        if (tokenStart != -1) {
+            // terminate trailing token
+            parseToken(txt, tokenStart, txt.length());
+        }
+
+        if (!timeFound
```
netkins: MAJOR Reduce the number of conditional operators (8) used in the expression (maximum allowed 5). [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1067)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormat.java
```diff
@@ -32,7 +32,9 @@
  * <li>Sunday, 06-Nov-94 08:49:37 GMT: obsolete specification</li>
  * <li>Sun Nov  6 08:49:37 1994: obsolete specification</li>
  * </ul>
+ * @deprecated Use {@link HttpHeaderDateFormatter} instead
  */
+@Deprecated
```
Scottmitch: Let just do in a followup to keep work scope smaller.

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,439 @@
...
+        if (len != 1 && len != 2) {
+            return false;
+        }
+
+        char c0 = txt.charAt(tokenStart);
+        if (isDigit(c0)) {
+            int localDayOfMonth = getNumericalValue(c0);
+
+            if (len == 2) {
+                char c1 = txt.charAt(tokenStart + 1);
```
Scottmitch: nit: `tokenStart + 1` -> `tokenEnd`?

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,439 @@
...
+            return true;
+        }
+
+        return false;
+    }
+
+    private boolean tryParseDayOfMonth(CharSequence txt, int tokenStart, int tokenEnd) {
+        int len = tokenEnd - tokenStart;
+
+        if (len != 1 && len != 2) {
```
Scottmitch: nit: consider using a similar approach as `tryParseYear` for consistency and less conditionals?

```java
if (len == 1) {
} else if (len == 2) {
} else {
  return false;
}
```

I know you mentioned you didn't like switch statements bcz of variables & scoping, but you can always introduce scope into the switch cases to get the same effect as above:

```java
switch (len) {
  case 1: {
    return ...;
  }
  case 2: {
    return ..;
  }
  default:
    return false;
}
```

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,439 @@
...
+    private Date parse0(CharSequence txt, int start, int end) {
+        int tokenStart = -1;
+
+        for (int i = start; i < end; i++) {
+            char c = txt.charAt(i);
+
+            if (isDelim(c)) {
+                if (tokenStart != -1) {
+                    // terminate token
+                    parseToken(txt, tokenStart, i);
```
Scottmitch: can we jump out early if we fail to parse a token, or we detect invalid input?

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
Scottmitch: I would consider at least combining the 2 conditionals. Currently it looks like there are duplicate checks between the 2.

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
slandelle: I guess 2 digit years will have been forbidden by then :)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormat.java
slandelle: ok

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
slandelle: That's not what I meant. I meant that something like this doesn't compile:
```java
switch(len) {
  case 1:
     char c0 = ???;
     return ???,
  case 2:
     char c0 = ???; // c0 is already defined
     char c1 = ???;
     return ???,
}

Comment (slandelle) [inline codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java]: going with if/else just like in `tryParseYear `

Comment (Scottmitch) [inline codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java]: @slandelle - Understood. You can scope each case as shown in the code above and that should  compile. That is roughly equivalent to the if/else blocks which define a new scope for each condition.

```java
switch(len) {
  case 1: {
     char c0 = ???;
     return ???;
  }
  case 2: {
     char c0 = ???; // c0 is already defined
     char c1 = ???;
     return ???;
  }
}
```

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
slandelle: tokenEnd == tokenStart + 2 here.

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
slandelle: No. the algorithm in the spec is intended to ignored non-matching tokens. You're supposed to keep on parsing until you've found all expected tokens.
For example, that's how day of week is completely ignored.

Scottmitch: does that mean infinite invalid input is spec compliant?

slandelle: Until you've found all expected tokens, or until you run out of tokens. I guess the only limit would be the HTTP header max length (if there's one).

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,425 @@
...
+        } else if (len == 2) {
+            char c0 = txt.charAt(tokenStart);
+            char c1 = txt.charAt(tokenStart + 1);
+            if (isDigit(c0) && isDigit(c1)) {
+                dayOfMonth = getNumericalValue(c0) * 10 + getNumericalValue(c1);
+                return true;
+            }
+        }
+
+        return false;
```
slandelle: @Scottmitch What do you think of this version? I'm not sure using switch would be better have as it would force me to have multiple returns per case statement.

Scottmitch: Yip this is fine. I just wanted to point out that what you were asking for (variables defined in case statements) is possible.

slandelle: I'm sorry, but I really fails to see how (without using different variables).
The example below definitively doesn't compile:
```java
switch(tokenEnd - tokenStart) {
    case 1:
        char c0 = txt.charAt(tokenStart);
        if (isDigit(c0)) {
            dayOfMonth = getNumericalValue(c0);
            return true;
        }
        return false;
    case 2:
        char c0 = txt.charAt(tokenStart);
        char c1 = txt.charAt(tokenStart + 1);
        if (isDigit(c0) && isDigit(c1)) {
            dayOfMonth = getNumericalValue(c0) * 10 + getNumericalValue(c1);
            return true;
        }
        return false;
    default:
        return false;
}
```

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.5.1:compile (default-compile) on project netty-codec-http: Compilation failure
[ERROR] /Users/slandelle/Documents/dev/IdeaProjects/netty-project/netty/codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java:[240,13] error: variable c0 is already defined in method tryParseDayOfMonth(CharSequence,int,int)
```

Am I missing something?

Scottmitch: You are missing some brackets which define a new scope for the case statement (just like an `if` has `{}` to define a scope).

```java
        switch(tokenEnd - tokenStart) {
            case 1: {
                char c0 = txt.charAt(tokenStart);
                if (isDigit(c0)) {
                    dayOfMonth = getNumericalValue(c0);
                    return true;
                }
                return false;
            }
            case 2: {
                char c0 = txt.charAt(tokenStart); // here, c0 is already defined above. AFAIK, switch doesn't scope per case statement so it can support falling through
                char c1 = txt.charAt(tokenStart + 1);
                if (isDigit(c0) && isDigit(c1)) {
                    dayOfMonth = getNumericalValue(c0) * 10 + getNumericalValue(c1);
                    return true;
                }
                return false;
            }
        }
        return false;
```

slandelle: Ahhh, right! Thanks!

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
Scottmitch: @slandelle - Even 64 or 128 is fine ... but no limit may be problematic.

slandelle: Instead of having a limit on the loop, wouldn't it be better to throw an IllegalArgumentException upstream directly in `parse`?

Scottmitch: @slandelle - As early as we can. However is it a problem if the input passed to `parse(CharSequence txt)` is large but the date is actually at the beginning of the string?

slandelle: There's some confusion here and I'll improve the Javadoc.
Input is supposed to have the exact length.

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,443 @@
...
+     * @param txt text to parse
+     * @param start the start index inside <code>txt</code>
+     * @param end the end index inside <code>txt</code>
+     * @return a {@link Date}, or null if text couldn't be parsed
+     */
+    public static Date parse(CharSequence txt, int start, int end) {
+        if (end - start > 64) {
+            throw new IllegalArgumentException("Can't parse more than 64 chars," +
+                    "looks like a user error or a malformed header");
+        }
```
slandelle: @Scottmitch see check here

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,443 @@
...
+
+        for (int i = start; i < end; i++) {
+            char c = txt.charAt(i);
+
+            if (isDelim(c)) {
+                if (tokenStart != -1) {
+                    // terminate token
+                    parseToken(txt, tokenStart, i);
+                    if (allPartsFound()) {
+                        return;
```
slandelle: @Scottmitch See fast exit here

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,443 @@
...
+                    }
+                    tokenStart = -1;
+                }
```
netkins: MAJOR Reduce the number of conditional operators (8) used in the expression (maximum allowed 5). [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1067)

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,443 @@
...
+        int tokenStart = -1;
+
+        for (int i = start; i < end; i++) {
+            char c = txt.charAt(i);
+
+            if (isDelim(c)) {
+                if (tokenStart != -1) {
+                    // terminate token
+                    parseToken(txt, tokenStart, i);
+                    if (allPartsFound()) {
```
Scottmitch: consider deriving this information from `parseToken` which already knows if everything is parsed to avoid re-evaluating all the conditionals. For example:

```java
private boolean parseToken(CharSequence txt, int tokenStart, int tokenEnd) {
// return true if you parsed all tokens, otherwise false.
}
```

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,443 @@
...
+     * @param start the start index inside <code>txt</code>
+     * @param end the end index inside <code>txt</code>
+     * @return a {@link Date}, or null if text couldn't be parsed
+     */
+    public static Date parse(CharSequence txt, int start, int end) {
+        int length = end - start;
+        if (length == 0) {
+            return null;
+        } else if (length < 0) {
+            throw new IllegalArgumentException("Can't have end < start");
```
slandelle: @Scottmitch 0 and negative length checks added

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,443 @@
...
+            monthFound = tryParseMonth(txt, tokenStart, tokenEnd);
+            if (monthFound) {
+                return timeFound && dayOfMonthFound && yearFound;
+            }
+        }
+
+        if (!yearFound) {
+            yearFound = tryParseYear(txt, tokenStart, tokenEnd);
+        }
+        return timeFound && dayOfMonthFound && monthFound && yearFound;
```
slandelle: allPartsFound logic moved here. Not sure if that saves that many evaluations but it's indeed better to have it here.

slandelle:
@Scottmitch last comments addressed. PTAL

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -0,0 +1,443 @@
...
+                    tokenStart = -1;
+                }
+            } else if (tokenStart == -1) {
+                // start new token
+                tokenStart = i;
+            }
+        }
+
+        // terminate trailing token
+        return tokenStart != -1 ? parseToken(txt, tokenStart, txt.length()) : false;
```
Scottmitch: `return tokenStart != -1 && parseToken(txt, tokenStart, txt.length());`

slandelle: feeling stupid

slandelle:
@Scottmitch Thanks a lot for the great review!

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
Scottmitch: nah I do this all the time ... helps to be reminded periodically ... thats what reviews are for :)

normanmaurer:
Cherry-picked this into 4.1 as edc4842309ad0491b7f2088c8116318b3d680db8. 

@slandelle if you also want to have this in 4.0 please create another PR against 4.0 :)

slandelle:
@normanmaurer No need to. My plan is to drop my own cookie stuff and upgrade to Netty 4.1. Thanks!

normanmaurer:
Ok Sounds good

PR #6022, opened 2016-11-15
Expose CharSequence version of HttpUtil#getMimeType and HttpUtil#getCharset

Spikhalskiy:
Motivation:

It would be more flexible to  make getCharset and getMimeType code usable not only for HttpMessage entity but just for any CharSequence. This will improve usability in general purpose code and will help to avoid multiple fetching of ContentType header from a message. It could be done in an external code once and CharSequence method versions could be applied.

Modification:
Expose HttpUtil#getMimeType, HttpUtil#getCharsetAsString, HttpUtil#getCharset versions which works with CharSequence. New methods are reused in the old ones which work with HttpMessage entity.

Result:

More flexible methods set with a good code reusing.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
@@ -372,7 +372,11 @@ public final class HttpUtil {
      * if charset is not presented or unparsable
      */
     public static Charset getCharset(CharSequence contentTypeValue) {
-        return getCharset(contentTypeValue, CharsetUtil.ISO_8859_1);
+        if (contentTypeValue != null) {
+            return getCharset(contentTypeValue, CharsetUtil.ISO_8859_1);
+        } else {
+            return CharsetUtil.ISO_8859_1;
+        }
     }
 
     /**
@@ -385,23 +389,31 @@ public final class HttpUtil {
      */
     public static Charset getCharset(HttpMessage message, Charset defaultCharset) {
         CharSequence contentTypeValue = message.headers().get(HttpHeaderNames.CONTENT_TYPE);
-        return getCharset(contentTypeValue, defaultCharset);
+        if (contentTypeValue != null) {
+            return getCharset(contentTypeValue, defaultCharset);
+        } else {
+            return defaultCharset;
+        }
     }
 
     /**
      * Fetch charset from Content-Type header value.
      *
      * @param contentTypeValue Content-Type header value to parse
-     * @param defaultCharset result to use in case of empty, incorrect or doesn't conain required part header value
+     * @param defaultCharset result to use in case of empty, incorrect or doesn't contain required part header value
      * @return the charset from message's Content-Type header or {@code defaultCharset}
      * if charset is not presented or unparsable
      */
     public static Charset getCharset(CharSequence contentTypeValue, Charset defaultCharset) {
-        CharSequence charsetCharSequence = getCharsetAsSequence(contentTypeValue);
-        if (charsetCharSequence != null) {
-            try {
-                return Charset.forName(charsetCharSequence.toString());
-            } catch (UnsupportedCharsetException unsupportedException) {
+        if (contentTypeValue != null) {
+            CharSequence charsetCharSequence = getCharsetAsSequence(contentTypeValue);
+            if (charsetCharSequence != null) {
+                try {
+                    return Charset.forName(charsetCharSequence.toString());
+                } catch (UnsupportedCharsetException unsupportedException) {
+                    return defaultCharset;
+                }
+            } else {
                 return defaultCharset;
             }
         } else {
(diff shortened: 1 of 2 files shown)


Spikhalskiy:
Build failed with "Build failure on specific text in build log: Resource leak detected", pretty sure it's unrelated to this PR, but could be worth to investigate.

normanmaurer:
@Spikhalskiy will check soon.

normanmaurer:
@Spikhalskiy just one thing... can you please name it `get*AsSequence` as its return type is `CharSequence`. I will cleanup the existing methods.

codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
```diff
@@ -346,13 +347,39 @@ public static Charset getCharset(HttpMessage message) {
...
      *
+     * @param message entity to fetch Content-Type header from
+     * @param defaultCharset result to use in case of empty, incorrect or doesn't conain required part header value
      * @return the charset from message's Content-Type header or {@code defaultCharset}
      * if charset is not presented or unparsable
      */
     public static Charset getCharset(HttpMessage message, Charset defaultCharset) {
-        CharSequence charsetCharSequence = getCharsetAsString(message);
+        CharSequence contentTypeValue = message.headers().get(HttpHeaderNames.CONTENT_TYPE);
+        return getCharset(contentTypeValue, defaultCharset);
```
normanmaurer: please only call when `contentTypeValue` != null

codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
```diff
@@ -370,11 +397,26 @@ public static Charset getCharset(HttpMessage message, Charset defaultCharset) {
...
+     * Fetch charset from Content-Type header value as a char sequence.
+     *
+     * A lot of sites/possibly clients have charset="CHARSET", for example charset="utf-8". Or "utf8" instead of "utf-8"
+     * This is not according to standard, but this method provide an ability to catch desired mistakes manually in code
+     *
+     * @param contentTypeValue Content-Type header value to parse
+     * @return the {@code CharSequence} with charset from message's Content-Type header
+     * or {@code null} if charset is not presented
+     */
+    public static CharSequence getCharsetAsString(CharSequence contentTypeValue) {
```
normanmaurer: rename to `getCharsetAsSequence`

codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
```diff
@@ -370,11 +397,26 @@ public static Charset getCharset(HttpMessage message, Charset defaultCharset) {
...
         if (contentTypeValue != null) {
```
normanmaurer: we should not allow `contentTypeValue` which is `null`

codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
```diff
@@ -399,6 +442,21 @@ public static CharSequence getCharsetAsString(HttpMessage message) {
      */
     public static CharSequence getMimeType(HttpMessage message) {
         CharSequence contentTypeValue = message.headers().get(HttpHeaderNames.CONTENT_TYPE);
+        return getMimeType(contentTypeValue);
```
normanmaurer: please only call when `contentTypeValue` != null

codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
```diff
@@ -399,6 +442,21 @@ public static CharSequence getCharsetAsString(HttpMessage message) {
...
+     * @param contentTypeValue Content-Type header value to parse
+     * @return the MIME type as a {@code CharSequence} from message's Content-Type header
+     * or {@code null} if content-type header or MIME type part of this header are not presented
+     * <p/>
+     * "content-type: text/html; charset=utf-8" - "text/html" will be returned <br/>
+     * "content-type: text/html" - "text/html" will be returned <br/>
+     * "content-type: " or no header - {@code null} we be returned
+     */
+    public static CharSequence getMimeType(CharSequence contentTypeValue) {
         if (contentTypeValue != null) {
```
normanmaurer: we should not allow `contentTypeValue` to be null.

Spikhalskiy:
@normanmaurer Addressed "we should not allow contentTypeValue to be null." comments is a separate commit. Let me know if it's fine, I will squash. I personally don't like the new interfaces.
CharSequence version works different from Charset. If a request has null Content-Type header - we return null, if we pass null to CharSequence method - we get NullPointerException. Unobviously, an ugly consumer code.

Maybe a better idea to make Sequence versions with default values too? And return always ```defaultCharset | defaultCharset``` in case of nulls or empties, or incorrect values? So, accept nulls as "", never return nulls in both CharSequence and Charset versions? WDYT?

normanmaurer:
Cherry-picked into 4.1 (464ae9fb7a1710c63646467fd0ca34c78bd1747a).

@Spikhalskiy sorry for the delay

PR #6025, opened 2016-11-15
Read Only Http2Headers

Scottmitch:
Motivation:
A read only implementation of Http2Headers can allow for a more efficient usage of memory and more performant combined construction and iteration during serialization.

Modifications:
- Add a new ReadOnlyHttp2Headers class

Result:
ReadOnlyHttp2Headers exists and can be used for performance reasons when appropriate.

```
Benchmark                                            (headerCount)  Mode  Cnt    Score   Error  Units
ReadOnlyHttp2HeadersBenchmark.defaultClientHeaders               1  avgt   20   96.156 ± 1.902  ns/op
ReadOnlyHttp2HeadersBenchmark.defaultClientHeaders               5  avgt   20  157.925 ± 3.847  ns/op
ReadOnlyHttp2HeadersBenchmark.defaultClientHeaders              10  avgt   20  236.257 ± 2.663  ns/op
ReadOnlyHttp2HeadersBenchmark.defaultClientHeaders              20  avgt   20  392.861 ± 3.932  ns/op
ReadOnlyHttp2HeadersBenchmark.defaultServerHeaders               1  avgt   20   48.759 ± 0.466  ns/op
ReadOnlyHttp2HeadersBenchmark.defaultServerHeaders               5  avgt   20  113.122 ± 0.948  ns/op
ReadOnlyHttp2HeadersBenchmark.defaultServerHeaders              10  avgt   20  192.698 ± 1.936  ns/op
ReadOnlyHttp2HeadersBenchmark.defaultServerHeaders              20  avgt   20  348.974 ± 3.111  ns/op
ReadOnlyHttp2HeadersBenchmark.defaultTrailers                    1  avgt   20   35.694 ± 0.271  ns/op
ReadOnlyHttp2HeadersBenchmark.defaultTrailers                    5  avgt   20   98.993 ± 2.933  ns/op
ReadOnlyHttp2HeadersBenchmark.defaultTrailers                   10  avgt   20  171.035 ± 5.068  ns/op
ReadOnlyHttp2HeadersBenchmark.defaultTrailers                   20  avgt   20  330.621 ± 3.381  ns/op
ReadOnlyHttp2HeadersBenchmark.readOnlyClientHeaders              1  avgt   20   40.573 ± 0.474  ns/op
ReadOnlyHttp2HeadersBenchmark.readOnlyClientHeaders              5  avgt   20   56.516 ± 0.660  ns/op
ReadOnlyHttp2HeadersBenchmark.readOnlyClientHeaders             10  avgt   20   76.890 ± 0.776  ns/op
ReadOnlyHttp2HeadersBenchmark.readOnlyClientHeaders             20  avgt   20  117.531 ± 1.393  ns/op
ReadOnlyHttp2HeadersBenchmark.readOnlyServerHeaders              1  avgt   20   29.206 ± 0.264  ns/op
ReadOnlyHttp2HeadersBenchmark.readOnlyServerHeaders              5  avgt   20   44.587 ± 0.312  ns/op
ReadOnlyHttp2HeadersBenchmark.readOnlyServerHeaders             10  avgt   20   64.458 ± 1.169  ns/op
ReadOnlyHttp2HeadersBenchmark.readOnlyServerHeaders             20  avgt   20  107.179 ± 0.881  ns/op
ReadOnlyHttp2HeadersBenchmark.readOnlyTrailers                   1  avgt   20   21.563 ± 0.202  ns/op
ReadOnlyHttp2HeadersBenchmark.readOnlyTrailers                   5  avgt   20   41.019 ± 0.440  ns/op
ReadOnlyHttp2HeadersBenchmark.readOnlyTrailers                  10  avgt   20   64.053 ± 0.785  ns/op
ReadOnlyHttp2HeadersBenchmark.readOnlyTrailers                  20  avgt   20  113.737 ± 4.433  ns/op
```

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
@@ -1,0 +1,831 @@
+/*
+ * Copyright 2016 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.codec.http2;
+
+import io.netty.handler.codec.Headers;
+import io.netty.util.AsciiString;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Set;
+
+import static io.netty.handler.codec.CharSequenceValueConverter.INSTANCE;
+import static io.netty.handler.codec.http2.DefaultHttp2Headers.HTTP2_NAME_VALIDATOR;
+import static io.netty.util.internal.EmptyArrays.EMPTY_ASCII_STRINGS;
+
+/**
+ * A variant of {@link Http2Headers} which only supports read-only methods.
+ * <p>
+ * Any array passed to this class may be used directly in the underlying data structures of this class. If these
+ * arrays may be modified it is the caller's responsibility to supply this class with a copy of the array.
+ * <p>
+ * This may be a good alternative to {@link DefaultHttp2Headers} if your have a fixed set of headers which will not
+ * change.
+ */
+public final class ReadOnlyHttp2Headers implements Http2Headers {
+    private static final byte PSEUDO_HEADER_TOKEN = (byte) ':';
+    private final AsciiString[] pseudoHeaders;
+    private final AsciiString[] otherHeaders;
+
+    /**
+     * Used to create read only object designed to represent trailers.
+     * <p>
+     * If this is used for a purpose other than trailers you may violate the header serialization ordering defined by
+     * <a href="https://tools.ietf.org/html/rfc7540#section-8.1.2.1">RFC 7540, 8.1.2.1</a>.
+     * @param validateHeaders {@code true} will run validation on each header name/value pair to ensure protocol
+     *                        compliance.
+     * @param otherHeaders A an array of key:value pairs. Must not contain any
+     *                     <a href="https://tools.ietf.org/html/rfc7540#section-8.1.2.1">pseudo headers</a>
+     *                     or {@code null} names/values.
+     *                     A copy will <strong>NOT</strong> be made of this array. If the contents of this array
+     *                     may be modified externally you are responsible for passing in a copy.
+     * @return A read only representation of the headers.
+     */
+    public static ReadOnlyHttp2Headers trailers(boolean validateHeaders, AsciiString... otherHeaders) {
+        return new ReadOnlyHttp2Headers(validateHeaders, EMPTY_ASCII_STRINGS, otherHeaders);
+    }
+
+    /**
+     * Create a new read only representation of headers used by clients.
+     * @param validateHeaders {@code true} will run validation on each header name/value pair to ensure protocol
+     *                        compliance.
+     * @param method The value for {@link PseudoHeaderName#METHOD}.
+     * @param path The value for {@link PseudoHeaderName#PATH}.
+     * @param scheme The value for {@link PseudoHeaderName#SCHEME}.
+     * @param authority The value for {@link PseudoHeaderName#AUTHORITY}.
+     * @param otherHeaders A an array of key:value pairs. Must not contain any
+     *                     <a href="https://tools.ietf.org/html/rfc7540#section-8.1.2.1">pseudo headers</a>
+     *                     or {@code null} names/values.
+     *                     A copy will <strong>NOT</strong> be made of this array. If the contents of this array
+     *                     may be modified externally you are responsible for passing in a copy.
+     * @return a new read only representation of headers used by clients.
+     */
+    public static ReadOnlyHttp2Headers clientHeaders(boolean validateHeaders,
+                                                     AsciiString method, AsciiString path,
+                                                     AsciiString scheme, AsciiString authority,
+                                                     AsciiString... otherHeaders) {
+        return new ReadOnlyHttp2Headers(validateHeaders,
+                new AsciiString[] {
+                  PseudoHeaderName.METHOD.value(), method, PseudoHeaderName.PATH.value(), path,
+                  PseudoHeaderName.SCHEME.value(), scheme, PseudoHeaderName.AUTHORITY.value(), authority
+                },
+                otherHeaders);
+    }
+
+    /**
+     * Create a new read only representation of headers used by servers.
+     * @param validateHeaders {@code true} will run validation on each header name/value pair to ensure protocol
+     *                        compliance.
+     * @param status The value for {@link PseudoHeaderName#STATUS}.
+     * @param otherHeaders A an array of key:value pairs. Must not contain any
+     *                     <a href="https://tools.ietf.org/html/rfc7540#section-8.1.2.1">pseudo headers</a>
+     *                     or {@code null} names/values.
+     *                     A copy will <strong>NOT</strong> be made of this array. If the contents of this array
+     *                     may be modified externally you are responsible for passing in a copy.
+     * @return a new read only representation of headers used by servers.
+     */
+    public static ReadOnlyHttp2Headers serverHeaders(boolean validateHeaders,
+                                                     AsciiString status,
+                                                     AsciiString... otherHeaders) {
+        return new ReadOnlyHttp2Headers(validateHeaders,
+                                        new AsciiString[] { PseudoHeaderName.STATUS.value(), status },
+                                        otherHeaders);
+    }
+
+    private ReadOnlyHttp2Headers(boolean validateHeaders, AsciiString[] pseudoHeaders, AsciiString... otherHeaders) {
+        assert (pseudoHeaders.length & 1) == 0; // pseudoHeaders are only set internally so assert should be enough.
+        if ((otherHeaders.length & 1) != 0) {
+            throw newInvalidArraySizeException();
+        }
+        if (validateHeaders) {
+            validateHeaders(pseudoHeaders, otherHeaders);
+        }
+        this.pseudoHeaders = pseudoHeaders;
+        this.otherHeaders = otherHeaders;
+    }
+
+    private static IllegalArgumentException newInvalidArraySizeException() {
+        return new IllegalArgumentException("pseudoHeaders and otherHeaders must be arrays of [name, value] pairs");
+    }
+
+    private static void validateHeaders(AsciiString[] pseudoHeaders, AsciiString... otherHeaders) {
+        // We are only validating values... so start at 1 and go until end.
+        for (int i = 1; i < pseudoHeaders.length; i += 2) {
+            // pseudoHeaders names are only set internally so they are assumed to be valid.
+            if (pseudoHeaders[i] == null) {
+                throw new IllegalArgumentException("pseudoHeaders value at index " + i + " is null");
+            }
+        }
+
+        boolean seenNonPseudoHeader = false;
+        final int otherHeadersEnd = otherHeaders.length - 1;
+        for (int i = 0; i < otherHeadersEnd; i += 2) {
+            AsciiString name = otherHeaders[i];
+            HTTP2_NAME_VALIDATOR.validateName(name);
+            if (!seenNonPseudoHeader && !name.isEmpty() && name.byteAt(0) != PSEUDO_HEADER_TOKEN) {
+                seenNonPseudoHeader = true;
+            } else if (seenNonPseudoHeader && !name.isEmpty() && name.byteAt(0) == PSEUDO_HEADER_TOKEN) {
+                throw new IllegalArgumentException(
+                     "otherHeaders name at index " + i + " is a pseudo header that appears after non-pseudo headers.");
+            }
+            if (otherHeaders[i + 1] == null) {
+                throw new IllegalArgumentException("otherHeaders value at index " + (i + 1) + " is null");
+            }
+        }
+    }
+
+    private AsciiString get0(CharSequence name) {
+        final int nameHash = AsciiString.hashCode(name);
+
+        final int pseudoHeadersEnd = pseudoHeaders.length - 1;
+        for (int i = 0; i < pseudoHeadersEnd; i += 2) {
+            AsciiString roName = pseudoHeaders[i];
+            if (roName.hashCode() == nameHash && roName.contentEqualsIgnoreCase(name)) {
+                return pseudoHeaders[i + 1];
+            }
+        }
+
+        final int otherHeadersEnd = otherHeaders.length - 1;
+        for (int i = 0; i < otherHeadersEnd; i += 2) {
+            AsciiString roName = otherHeaders[i];
+            if (roName.hashCode() == nameHash && roName.contentEqualsIgnoreCase(name)) {
+                return otherHeaders[i + 1];
+            }
+        }
+        return null;
+    }
+
+    @Override
+    public CharSequence get(CharSequence name) {
+        return get0(name);
+    }
+
+    @Override
+    public CharSequence get(CharSequence name, CharSequence defaultValue) {
+        CharSequence value = get(name);
+        return value != null ? value : defaultValue;
+    }
+
+    @Override
+    public CharSequence getAndRemove(CharSequence name) {
+        throw new UnsupportedOperationException("read only");
+    }
+
+    @Override
+    public CharSequence getAndRemove(CharSequence name, CharSequence defaultValue) {
+        throw new UnsupportedOperationException("read only");
+    }
+
+    @Override
+    public List<CharSequence> getAll(CharSequence name) {
+        final int nameHash = AsciiString.hashCode(name);
+        List<CharSequence> values = new ArrayList<CharSequence>();
+
+        final int pseudoHeadersEnd = pseudoHeaders.length - 1;
+        for (int i = 0; i < pseudoHeadersEnd; i += 2) {
+            AsciiString roName = pseudoHeaders[i];
+            if (roName.hashCode() == nameHash && roName.contentEqualsIgnoreCase(name)) {
+                values.add(pseudoHeaders[i + 1]);
+            }
+        }
+
+        final int otherHeadersEnd = otherHeaders.length - 1;
+        for (int i = 0; i < otherHeadersEnd; i += 2) {
+            AsciiString roName = otherHeaders[i];
+            if (roName.hashCode() == nameHash && roName.contentEqualsIgnoreCase(name)) {
+                values.add(otherHeaders[i + 1]);
+            }
+        }
+
+        return values;
+    }
+
+    @Override
+    public List<CharSequence> getAllAndRemove(CharSequence name) {
+        throw new UnsupportedOperationException("read only");
+    }
+
+    @Override
+    public Boolean getBoolean(CharSequence name) {
+        AsciiString value = get0(name);
+        return value != null ? INSTANCE.convertToBoolean(value) : null;
+    }
+
+    @Override
+    public boolean getBoolean(CharSequence name, boolean defaultValue) {
+        Boolean value = getBoolean(name);
+        return value != null ? value : defaultValue;
+    }
+
+    @Override
+    public Byte getByte(CharSequence name) {
+        AsciiString value = get0(name);
+        return value != null ? INSTANCE.convertToByte(value) : null;
+    }
+
+    @Override
+    public byte getByte(CharSequence name, byte defaultValue) {
+        Byte value = getByte(name);
+        return value != null ? value : defaultValue;
+    }
+
+    @Override
+    public Character getChar(CharSequence name) {
+        AsciiString value = get0(name);
+        return value != null ? INSTANCE.convertToChar(value) : null;
+    }
+
+    @Override
+    public char getChar(CharSequence name, char defaultValue) {
+        Character value = getChar(name);
+        return value != null ? value : defaultValue;
+    }
+
+    @Override
+    public Short getShort(CharSequence name) {
+        AsciiString value = get0(name);
+        return value != null ? INSTANCE.convertToShort(value) : null;
+    }
+
+    @Override
+    public short getShort(CharSequence name, short defaultValue) {
+        Short value = getShort(name);
+        return value != null ? value : defaultValue;
+    }
+
+    @Override
+    public Integer getInt(CharSequence name) {
+        AsciiString value = get0(name);
+        return value != null ? INSTANCE.convertToInt(value) : null;
+    }
+
+    @Override
+    public int getInt(CharSequence name, int defaultValue) {
+        Integer value = getInt(name);
+        return value != null ? value : defaultValue;
+    }
+
+    @Override
+    public Long getLong(CharSequence name) {
+        AsciiString value = get0(name);
+        return value != null ? INSTANCE.convertToLong(value) : null;
+    }
+
+    @Override
+    public long getLong(CharSequence name, long defaultValue) {
+        Long value = getLong(name);
+        return value != null ? value : defaultValue;
+    }
+
+    @Override
+    public Float getFloat(CharSequence name) {
+        AsciiString value = get0(name);
+        return value != null ? INSTANCE.convertToFloat(value) : null;
+    }
+
+    @Override
+    public float getFloat(CharSequence name, float defaultValue) {
+        Float value = getFloat(name);
+        return value != null ? value : defaultValue;
+    }
+
+    @Override
+    public Double getDouble(CharSequence name) {
+        AsciiString value = get0(name);
+        return value != null ? INSTANCE.convertToDouble(value) : null;
+    }
+
+    @Override
+    public double getDouble(CharSequence name, double defaultValue) {
+        Double value = getDouble(name);
+        return value != null ? value : defaultValue;
+    }
+
+    @Override
+    public Long getTimeMillis(CharSequence name) {
+        AsciiString value = get0(name);
+        return value != null ? INSTANCE.convertToTimeMillis(value) : null;
+    }
+
+    @Override
+    public long getTimeMillis(CharSequence name, long defaultValue) {
+        Long value = getTimeMillis(name);
+        return value != null ? value : defaultValue;
+    }
+
+    @Override
+    public Boolean getBooleanAndRemove(CharSequence name) {
+        throw new UnsupportedOperationException("read only");
+    }
+
+    @Override
+    public boolean getBooleanAndRemove(CharSequence name, boolean defaultValue) {
+        throw new UnsupportedOperationException("read only");
+    }
+
+    @Override
+    public Byte getByteAndRemove(CharSequence name) {
+        throw new UnsupportedOperationException("read only");
+    }
+
+    @Override
+    public byte getByteAndRemove(CharSequence name, byte defaultValue) {
+        throw new UnsupportedOperationException("read only");
+    }
+
+    @Override
+    public Character getCharAndRemove(CharSequence name) {
+        throw new UnsupportedOperationException("read only");
+    }
+
+    @Override
+    public char getCharAndRemove(CharSequence name, char defaultValue) {
+        throw new UnsupportedOperationException("read only");
+    }
+
+    @Override
+    public Short getShortAndRemove(CharSequence name) {
+        throw new UnsupportedOperationException("read only");
+    }
+
+    @Override
+    public short getShortAndRemove(CharSequence name, short defaultValue) {
+        throw new UnsupportedOperationException("read only");
+    }
+
+    @Override
+    public Integer getIntAndRemove(CharSequence name) {
+        throw new UnsupportedOperationException("read only");
+    }
+
+    @Override
+    public int getIntAndRemove(CharSequence name, int defaultValue) {
+        throw new UnsupportedOperationException("read only");
+    }
+
+    @Override
+    public Long getLongAndRemove(CharSequence name) {
+        throw new UnsupportedOperationException("read only");
+    }
+
+    @Override
+    public long getLongAndRemove(CharSequence name, long defaultValue) {
+        throw new UnsupportedOperationException("read only");
+    }
+
+    @Override
+    public Float getFloatAndRemove(CharSequence name) {
+        throw new UnsupportedOperationException("read only");
+    }
+
+    @Override
+    public float getFloatAndRemove(CharSequence name, float defaultValue) {
+        throw new UnsupportedOperationException("read only");
+    }
+
+    @Override
+    public Double getDoubleAndRemove(CharSequence name) {
+        throw new UnsupportedOperationException("read only");
+    }
+
+    @Override
+    public double getDoubleAndRemove(CharSequence name, double defaultValue) {
+        throw new UnsupportedOperationException("read only");
+    }
+
+    @Override
+    public Long getTimeMillisAndRemove(CharSequence name) {
+        throw new UnsupportedOperationException("read only");
+    }
+
+    @Override
+    public long getTimeMillisAndRemove(CharSequence name, long defaultValue) {
+        throw new UnsupportedOperationException("read only");
+    }
+
+    @Override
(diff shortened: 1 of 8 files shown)


Scottmitch:
@buchgr @carl-mastrangelo - FYI ... we discussed making a more general version of this in the past ... and I was motivated this weekend. Let me know if you will be able to digest this as the APIs are slightly different than [GrpcHttp2OutboundHeaders.java](https://github.com/grpc/grpc-java/blob/master/netty/src/main/java/io/grpc/netty/GrpcHttp2OutboundHeaders.java), but you should be able to fold the non-pseudo headers treated as independent parameters into `otherHeaders` (it looks like there is already an array copy [here](https://github.com/grpc/grpc-java/blob/master/netty/src/main/java/io/grpc/netty/GrpcHttp2OutboundHeaders.java#L80)), right?

common/src/main/java/io/netty/util/internal/EmptyArrays.java
```diff
@@ -27,6 +29,7 @@
     public static final Object[] EMPTY_OBJECTS = {};
     public static final Class<?>[] EMPTY_CLASSES = {};
     public static final String[] EMPTY_STRINGS = {};
+    public static final AsciiString[] EMPTY_ASCII_STRINGS = {};
```
netkins: CRITICAL Make this array "private". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1873)
CRITICAL Make this member "protected". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2386)

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
```diff
@@ -0,0 +1,824 @@
...
+        String separator = "";
+        for (Map.Entry<CharSequence, CharSequence> entry : this) {
+            builder.append(separator);
+            builder.append(entry.getKey()).append(": ").append(entry.getValue());
+            separator = ", ";
+        }
+        return builder.append(']').toString();
+    }
+
+    private final class ReadOnlyIterator implements Map.Entry<CharSequence, CharSequence>,
```
netkins: MAJOR io.netty.handler.codec.http2.ReadOnlyHttp2Headers$ReadOnlyIterator is both an Iterator and a Map.Entry [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=findbugs%3APZ_DONT_REUSE_ENTRY_OBJECTS_IN_ITERATORS)

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
```diff
@@ -0,0 +1,824 @@
...
+            throw new IllegalArgumentException("pseudoHeaders and otherHeaders must be arrays of [name, value] pairs");
+        }
+        if (validateHeaders) {
+            validateHeaders(pseudoHeaders, otherHeaders);
+        }
+        this.pseudoHeaders = pseudoHeaders;
+        this.otherHeaders = otherHeaders;
+    }
+
+    private static void validateHeaders(AsciiString[] pseudoHeaders, AsciiString[] otherHeaders) {
```
netkins: MINOR Consider using varargs for methods or constructors which take an array the last parameter. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUseVarargs)

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
```diff
@@ -0,0 +1,824 @@
...
+     * Create a new read only representation of headers used by servers.
+     * @param status The value for {@link PseudoHeaderName#STATUS}.
+     * @param otherHeaders A an array of key:value pairs. Must not contain any
+     *                     <a href="https://tools.ietf.org/html/rfc7540#section-8.1.2.1">pseudo headers</a>
+     *                     or {@code null} names/values.
+     * @param validateHeaders {@code true} will run validation on each header name/value pair to ensure protocol
+     *                        compliance.
+     * @return a new read only representation of headers used by servers.
+     */
+    public static ReadOnlyHttp2Headers serverHeaders(AsciiString status, AsciiString[] otherHeaders,
```
carl-mastrangelo: gRPC has a particular requirement that there are only two times during the stream when headers can be sent.  In theory, its possible (per the HTTP Spec) to send multiple headers blocks during the stream.  (for example, headers, then data, then headers, then data, .......)


This seems like a specialization that gRPC can safely do, but not sure about the general purpose one here.

Scottmitch: >  In theory, its possible (per the HTTP Spec) to send multiple headers blocks during the stream

Can you clarify where the spec permits this? https://tools.ietf.org/html/rfc7540#section-8.1 formally define the message flow and prevent the scenario you described.

Scottmitch:
also /cc @nmittler

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
carl-mastrangelo: I was looking at the diagram in https://tools.ietf.org/html/rfc7540#section-5.1 which seems to imply that only RST or EOS could indicate a life cycle change.  

I suppose the section you linked to is in addition to that one.

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
```diff
@@ -0,0 +1,824 @@
...
+     * <a href="https://tools.ietf.org/html/rfc7540#section-8.1.2.1">RFC 7540, 8.1.2.1</a>.
+     * @param otherHeaders A an array of key:value pairs. Must not contain any
+     *                     <a href="https://tools.ietf.org/html/rfc7540#section-8.1.2.1">pseudo headers</a>
+     *                     or {@code null} names/values.
+     * @param validateHeaders {@code true} will run validation on each header name/value pair to ensure protocol
+     *                        compliance.
+     * @return A read only representation of the headers.
+     */
+    public static ReadOnlyHttp2Headers trailers(AsciiString[] otherHeaders, boolean validateHeaders) {
+        return new ReadOnlyHttp2Headers(EMPTY_ASCII_STRINGS, otherHeaders, validateHeaders);
```
carl-mastrangelo: Something tricky with making a generic library is that you might have to enforce some safety that you may not want to.   For example, If someone passes an array here but then mutates the array after, the invariant that this is ReadOnly would be violated.  Also, what happens if there are null gaps in the array?

You might need to add some strong words to the javadoc.

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
```diff
@@ -0,0 +1,824 @@
...
+                                                     boolean validateHeaders) {
+        return new ReadOnlyHttp2Headers(new AsciiString[] { PseudoHeaderName.STATUS.value(), status }, otherHeaders,
+                                        validateHeaders);
+    }
+
+    private ReadOnlyHttp2Headers(AsciiString[] pseudoHeaders, AsciiString[] otherHeaders, boolean validateHeaders) {
+        if ((pseudoHeaders.length & 1) != 0 || (otherHeaders.length & 1) != 0) {
+            throw new IllegalArgumentException("pseudoHeaders and otherHeaders must be arrays of [name, value] pairs");
+        }
+        if (validateHeaders) {
```
carl-mastrangelo: Adding any sort of logic will impair the ability of Java to inline this constructor.  You might consider not reusing code and just make specialized overloads with smaller bytesizes.

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
```diff
@@ -0,0 +1,824 @@
...
+    }
+
+    private static void validateHeaders(AsciiString[] pseudoHeaders, AsciiString[] otherHeaders) {
+        boolean seenNonPseudoHeader = false;
+        int i;
+        AsciiString name;
+        for (i = 0; i < pseudoHeaders.length; ++i) {
+            name = pseudoHeaders[i];
+            HTTP2_NAME_VALIDATOR.validateName(name);
+            if (!seenNonPseudoHeader && !name.isEmpty() && name.byteAt(0) != ':') {
```
carl-mastrangelo: byte at is slow right?  it does 2 bounds checks correct?

Scottmitch: AsciiString could represent a subsequence so yes it checks if the index is within [lower, upper] bounds. Do you have an alternative suggestion? I'm not sure this is significant considering we are doing a full iteration of the name before this.

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
Scottmitch: If I abstract out the above `throw` statement into another method then the constructor is 42 bytes of byte code with this conditional vs 33 bytes without ... however depending on the scenario they can be both inlined. The performance relative to "non-readonly" as demonstrated by the benchmarks seems like good improvement, and having separate methods may make it more likely users will miss the validation. My preference is to consider this tradeoff in a followup PR if necessary.

as is:
```
@ 9   io.netty.handler.codec.http2.ReadOnlyHttp2Headers::<init> (42 bytes)   inline (hot)
```

remove this validate conditional:
```
@ 9   io.netty.handler.codec.http2.ReadOnlyHttp2Headers::<init> (33 bytes)   inline (hot)
```

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
Scottmitch: Agreed there is a performance/encapsulation tradeoff in leaking the internal data structure externally. The purpose of this class is primarily for a performance benefit in certain situations over the existing headers which do not leak internals. I think its OK to allow the caller to determine if copying is necessary or not and clarify in javadocs.

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
Scottmitch: can you clarify if you agree that 8.1 prevents the scenario you described, or do you think we need to take action to support this use case?

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
```diff
@@ -0,0 +1,843 @@
...
+        }
+        this.pseudoHeaders = pseudoHeaders;
+        this.otherHeaders = otherHeaders;
+    }
+
+    private static IllegalArgumentException newIllegalArgumentException() {
+        return new IllegalArgumentException("pseudoHeaders and otherHeaders must be arrays of [name, value] pairs");
+    }
+
+    private static void validateHeaders(AsciiString[] pseudoHeaders, AsciiString[] otherHeaders) {
```
netkins: MINOR Consider using varargs for methods or constructors which take an array the last parameter. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUseVarargs)

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
carl-mastrangelo: And the null gaps?  gRPC's Headers used nulls to indicate as a shortcut that there were no more headers in the array.  Jakob added this, but that's because we could be really flexible with it.  Here, I think you still need to call it out.

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
carl-mastrangelo: I *think* you are right.

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
carl-mastrangelo: iteration is not that expensive because it can skip the bounds check right?  Also, won't comparing ':' be an unnecessary upcast from byte to char?

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
```diff
@@ -0,0 +1,843 @@
...
+        boolean seenNonPseudoHeader = false;
+        int i;
+        AsciiString name;
+        for (i = 0; i < pseudoHeaders.length; ++i) {
+            name = pseudoHeaders[i];
+            HTTP2_NAME_VALIDATOR.validateName(name);
+            if (!seenNonPseudoHeader && !name.isEmpty() && name.byteAt(0) != ':') {
+                seenNonPseudoHeader = true;
+            } else if (seenNonPseudoHeader && !name.isEmpty() && name.byteAt(0) == ':') {
```
carl-mastrangelo: This seems like doing extra work.  In the case that the caller used one of the special constructors, the pseudo headers are always going to be valid right?  It would be a waste to check them here.

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
```diff
@@ -0,0 +1,843 @@
...
+                throw new IllegalArgumentException(
+                     "pseudoHeaders name at index " + i + " is a pseudo header that appears after non-pseudo headers.");
+            }
+            if (pseudoHeaders[++i] == null) {
```
carl-mastrangelo: Not sure how the rest of Netty is written, but I personally find the i++ tricky.  I also vaguely suspect it hurts ILP due to the dependence of i being incremented then read, incremented then read.

Why not use i+1 here, and i+=2 in the loop?

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
```diff
@@ -0,0 +1,843 @@
...
+    }
+
+    private AsciiString get0(CharSequence name) {
+        final int nameHash = name.hashCode();
+        AsciiString roName;
+        int roHash;
+        final int psudoHeadersEnd = pseudoHeaders.length - 1;
+        for (int i = 0; i < psudoHeadersEnd; i += 2) {
+            roName = pseudoHeaders[i];
+            roHash = roName.hashCode();
```
carl-mastrangelo: How do you ensure that different charsequences all have the same hash?  The AsciiString hash and a regular String hash will be different, despite being the same list of characters right?

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
```diff
@@ -0,0 +1,843 @@
...
+        }
+
+        @Override
+        public void remove() {
+            throw new UnsupportedOperationException("read only");
+        }
+
+        @Override
+        public String toString() {
+            return key.toString() + '=' + value.toString();
```
carl-mastrangelo: toString here doesn't make much since, since the iterator is going to be changing it often.

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
Scottmitch: > iteration is not that expensive because it can skip the bounds check right?

Each individual access is be comparably cheaper than `byteAt()` but it is an `O(n)` iteration.

> Also, won't comparing ':' be an unnecessary upcast from byte to char?

Good catch. Let me be careful to avoid this.

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
Scottmitch: > the pseudo headers are always going to be valid right

Good point. Currently I am not allowing the pseudo headers array to be specified from the outside. I can simplify this check.

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
Scottmitch: @carl-mastrangelo - I can change. That being said I don't know definitively which approach is superior. `++i` corresponds to its own byte code and it is feasible that less load immediate may be necessary. However I also see the potential reducing store operations and dependencies on `I`...

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
Scottmitch: yip this should be `AsciiString.hashCode()` ... This may incur additional costs if the hash code is not cached (e.g. `String`) so I may just pull this out...

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
Scottmitch: if we are concerned about the object state being mutated then perhaps we shouldn't have a single object represent  `Iterator` and `Entry`. However the `toString()` may still provide a useful snapshot into the object state for debugging etc. Are you concerned this may be misleading or miss-used in some way?

codec/src/main/java/io/netty/handler/codec/DefaultHeaders.java
```diff
@@ -978,7 +978,7 @@ protected HeaderDateFormat initialValue() {
                     }
                 };
 
-        static HeaderDateFormat get() {
+        public static HeaderDateFormat get() {
```
Scottmitch: just a reminder to my self ... instead of exposing this I may want to investigate using the date parsing from https://github.com/netty/netty/pull/6018.

Scottmitch:
@carl-mastrangelo - pushed new code PTAL

Scottmitch:
@carl-mastrangelo - gentle ping

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
```diff
@@ -0,0 +1,825 @@
...
+
+    private static IllegalArgumentException newIllegalArgumentException() {
+        return new IllegalArgumentException("pseudoHeaders and otherHeaders must be arrays of [name, value] pairs");
+    }
+
+    private static void validateHeaders(AsciiString[] pseudoHeaders, AsciiString[] otherHeaders) {
+        boolean seenNonPseudoHeader = false;
+        final int pseudoHeadersEnd = pseudoHeaders.length - 1;
+        for (int i = 0; i < pseudoHeadersEnd; i += 2) {
+            HTTP2_NAME_VALIDATOR.validateName(pseudoHeaders[i]);
```
carl-mastrangelo: Remind me: this is statically determinable that the names will always be correct right?  The only place these can get set is from this file, which can be verified to be right.

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
```diff
@@ -0,0 +1,825 @@
...
+            }
+        }
+        final int otherHeadersEnd = otherHeaders.length - 1;
+        for (int i = 0; i < otherHeadersEnd; i += 2) {
+            AsciiString name = otherHeaders[i];
+            HTTP2_NAME_VALIDATOR.validateName(name);
+            if (!seenNonPseudoHeader && !name.isEmpty() && name.byteAt(0) != PSEUDO_HEADER_TOKEN) {
+                seenNonPseudoHeader = true;
+            } else if (seenNonPseudoHeader && !name.isEmpty() && name.byteAt(0) == PSEUDO_HEADER_TOKEN) {
+                throw new IllegalArgumentException(
```
carl-mastrangelo: should this be `newIllegalArgumentException` ?

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
```diff
@@ -0,0 +1,825 @@
...
+    }
+
+    @Override
+    public boolean containsTimeMillis(CharSequence name, long value) {
+        return contains(name, String.valueOf(value));
+    }
+
+    @Override
+    public int size() {
+        return (pseudoHeaders.length + otherHeaders.length) >>> 1;
```
carl-mastrangelo: why the zero fill shift?  Below, names() uses this function to preallocate  a Hashset to store all names.  If size() were to overflow, then the array allocated in LinkedhashSet would overflow too.

Effectively, something is going to overflow and blow up incase this is too large, so there isn't any point trying to be careful here.

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
```diff
@@ -0,0 +1,832 @@
...
+            throw newIllegalArgumentException();
+        }
+        if (validateHeaders) {
+            validateHeaders(pseudoHeaders, otherHeaders);
+        }
+        this.pseudoHeaders = pseudoHeaders;
+        this.otherHeaders = otherHeaders;
+    }
+
+    private static IllegalArgumentException newIllegalArgumentException() {
```
Scottmitch: This code path will generally not be taken and it makes the byte code for the method smaller.

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
Scottmitch: @carl-mastrangelo - good catch ... shouldn't need to verify the pseudo headers. I was exposing them externally at one point but now I'm not.

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
Scottmitch: `newIllegalArgumentException` is if the length of the arrays I not appropriate. This check validates there is a `pseudo` header which appears after a non-pseudo header. The exception message here is different and provides more context for this specific case. Are you suggesting making `newIllegalArgumentException` more general?

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
Scottmitch: > Below, names() uses this function to preallocate a Hashset to store all names. If size() were to overflow, then the array allocated in LinkedhashSet would overflow too.

Not necessarily. I don't see why letting this method overflow and returning a negative value makes sense. I would rather let the available memory and the `LinkedHashSet` implementation impose its own limits than not account for the `size()` correctly. Also `size()` may be used for other reasons than allocating that amount of memory (e.g. `toString()` debugging, hash index bucketing, threshold checks, etc...).

```java
int x = Integer.MAX_VALUE / 2 + 1; // 1073741824
int y = x; // 1073741824
int a = (x + y) >>> 1; // 1073741824
int b = (x + y) >> 1; // -1073741824
// We wouldn't want to return a negative value for size()
```

Can you provide more context?

Scottmitch:
@carl-mastrangelo - Back to you :)

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
```diff
@@ -0,0 +1,819 @@
...
+        return new IllegalArgumentException("pseudoHeaders and otherHeaders must be arrays of [name, value] pairs");
+    }
+
+    private static void validateHeaders(AsciiString[] otherHeaders) {
```
netkins: MINOR Consider using varargs for methods or constructors which take an array the last parameter. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUseVarargs)

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
Scottmitch: I should still be validating the values for `pseudoHeaders` though

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
carl-mastrangelo: actually, why not do it the same way binary search implementations avoid overflows:
```
return pseudoHeaders.length /2 + otherHeaders.length / 2;
```

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
carl-mastrangelo: Good point.  That said, I think you can still lose the `.toString()` calls, since they are implicit.

codec-http2/src/test/java/io/netty/handler/codec/http2/ReadOnlyHttp2HeadersTest.java
```diff
@@ -0,0 +1,187 @@
...
+import org.junit.Test;
+
+import java.util.Iterator;
+import java.util.Map;
+
+import static io.netty.handler.codec.http2.DefaultHttp2HeadersTest.verifyPseudoHeadersFirst;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
```
carl-mastrangelo: Does Netty not add the `@RunWith(JUnit4.class)` annotation?

codec-http2/src/test/java/io/netty/handler/codec/http2/ReadOnlyHttp2HeadersTest.java
```diff
@@ -0,0 +1,187 @@
...
+public class ReadOnlyHttp2HeadersTest {
+    @Test(expected = IllegalArgumentException.class)
```
carl-mastrangelo: You might consider using the `@Rule ExpectedException` in Junit4, which lets you set an expected exception AND error message.  It's useful for declaring which exception you expect more clearly.

codec-http2/src/test/java/io/netty/handler/codec/http2/ReadOnlyHttp2HeadersTest.java
```diff
@@ -0,0 +1,187 @@
...
+        assertTrue(headers.contains(Http2Headers.PseudoHeaderName.PATH.value(), "/foo"));
+        assertFalse(headers.contains(Http2Headers.PseudoHeaderName.STATUS.value(), "200"));
+        assertFalse(headers.contains("a missing header", "a missing value"));
+    }
+
+    @Test
+    public void testGet() {
+        Http2Headers headers = newClientHeaders();
+        assertTrue(AsciiString.contentEqualsIgnoreCase("value1", headers.get("Name1")));
+        assertTrue(AsciiString.contentEqualsIgnoreCase("/foo",
```
carl-mastrangelo: Out of curiousity, what is Netty's policy on line wrapping?  Is there a style guide for how to wrap long lines?

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
Scottmitch: @carl-mastrangelo - Oh I see ... I can clarify the name. It was designed to reduce the byte code for the "hot" path. The though is with validation there will be move overhead incurred anyways so it is not as urgent.

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
Scottmitch: `(x+y) >>> 1` seems to be more correct w.r.t. overflow and involves less operations.

```java
@Test
public void testSizeAlgos() {
  final int halfway = Integer.MAX_VALUE / 2;
  for (int o = 1; o <= halfway; ++o) {
    for (int i = Integer.MAX_VALUE; i >= halfway; --i) {
        final long expected = ((long) o + i) / 2;
        assertEquals("zero shift o: " + o + " i: " + i,  expected, (o + i) >>> 1);
        assertEquals("double divide o: " + o + " i: " + i,  expected, o / 2 + i / 2);
    }
  }
  // run this where o is an odd number

  // java.lang.AssertionError: double divide o: 1 i: 2147483647 
  // Expected :1073741824
  // Actual   :1073741823

  // java.lang.AssertionError: double divide o: 3 and i: 2147483647
  // Expected: 1073741824
  // Actual  : 1073741825
}
```

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
Scottmitch: To get the implicit to compile I can't use a `char` for `'='`. I can do this if you prefer (not sure it is a big deal either way though).

codec-http2/src/test/java/io/netty/handler/codec/http2/ReadOnlyHttp2HeadersTest.java
Scottmitch: We don't use `@RunWith(JUnit4.class)` ... is that an issue? We do use `@RunWith(Theories.class)` and `@RunWith(Parameterized.class)` where necessary...

codec-http2/src/test/java/io/netty/handler/codec/http2/ReadOnlyHttp2HeadersTest.java
Scottmitch: I generally try to avoid verifying error messages unless there is no other way to verify the constraint I'm after. It can be fragile with string localization, different JDK implementations, etc... I think verifying the exception is good enough here.

codec-http2/src/test/java/io/netty/handler/codec/http2/ReadOnlyHttp2HeadersTest.java
Scottmitch: I don't think there is a strict style guide for line wrapping (other than 120 chars). Here is our check style file: https://github.com/netty/netty-build/blob/master/src/main/resources/io/netty/checkstyle.xml. Feel free to submit a PR if you have an idea for an improvement.

Scottmitch:
@carl-mastrangelo - ping

codec-http2/src/test/java/io/netty/handler/codec/http2/ReadOnlyHttp2HeadersTest.java
carl-mastrangelo: Not a big issue, it was more to sate my curiosity.   Google requires the annotation.

codec-http2/src/test/java/io/netty/handler/codec/http2/ReadOnlyHttp2HeadersTest.java
carl-mastrangelo: OKay.  In the future, you might consider using it when looking for NPE, since those are hard to ensure the NPE you were expecting is actually the one you get.

codec-http2/src/test/java/io/netty/handler/codec/http2/ReadOnlyHttp2HeadersTest.java
carl-mastrangelo: Ok.  Google's style guide is 100 chars, +4 for line wrap, +2 for new block.   I don't want to comment on your style, just curious what Netty's policy is.

carl-mastrangelo:
@Scottmitch LGTM.  You should probably get review from someone else though.  Sorry for the massive delays on this; Q4 goals beckon.

Scottmitch:
@carl-mastrangelo - Thanks for review ... no problem about delay. I'll just pull this in now:

- the HTTP/2 interface is not stable anyways ... so we can fix API issues later if necessary.
- change is relatively self contained.
- The PR has been opened for a while ... interested folks likely would have reviewed anyways.

Scottmitch:
4.1 (06e7627)

PR #6024, opened 2016-11-15
retained[Slice|Duplicate] buffer reference count bug

Scottmitch:
Motivation:
Currently the ByteBuf created as a result of retained[Slice|Duplicate] maintains its own reference count, and when this reference count is depleated it will release the ByteBuf returned from unwrap(). The unwrap() buffer is designed to be the 'root parent' and will skip all intermediate layers of buffers. If the intermediate layers of buffers contain a retained[Slice|Duplicate] then these reference counts will be ignored during deallocation. This may lead to deallocating the 'root parent' before all derived pooled buffers are actually released. This same issue holds if a retained[Slice|Duplicate] is in the heirachy and a 'regular' slice() or duplicate() buffer is created.

Modifications:
- AbstractPooledDerivedByteBuf must maintain a reference to the direct parent (the buffer which retained[Slice|Duplicate] was called on) and release on this buffer instead of the 'root parent' returned by unwrap()
- slice() and duplicate() buffers created from AbstractPooledDerivedByteBuf must also delegate reference count operations to their immediate parent (or first ancestor which maintains an independent reference count).

Result:
Fixes #5999

PR #6026, opened 2016-11-16
LzmaFrameEncoderTest leak due to LzmaInputStream close behavior

Scottmitch:
Motivation:
c1932a8537b742aaf15a7cfacf9f76ad8239f3c7 made an assumption that the LzmaInputStream which wraps a ByteBufInputStream would delegate the close operation to the wrapped stream. This assumption is not true and thus we still had a leak. An issue has been logged with our LZMA dependency https://github.com/jponge/lzma-java/issues/14.

Modifications:
- Force a close on the wrapped stream

Result:
No more leak.


Scottmitch:
saw another leak while building https://github.com/netty/netty/pull/6024 https://garage.netty.io/teamcity/viewLog.html?buildId=13716&buildTypeId=pulls_netty&tab=buildLog

Scottmitch:
4.1 (2c78902)

PR #6028, opened 2016-11-16
00fc239995ce7b683993c88bb82b30dbd5fe7982 HashedWheelTimer introduced …

Scottmitch:
…test failure

Motivation:
00fc239995ce7b683993c88bb82b30dbd5fe7982 introduced a change to HashedWheelTimerTest which attempted to wait for an explicit event notification until more timer events can be added. However HashedWheelTimer will execute the timer Runnable before removing it from the queue and decrementing the total count. This make it difficult for users to know when it is safe to add another timer task as the limit is approached.

Modifications:
- HashedWheelTimer should remove the timer Runnable before executing the task.

Result:
Users can more reliably add new timers when the limit is reached and HashedWheelTimerTest will no longer fail spuriously due to this race condition.

### DIFF
--- a/common/src/main/java/io/netty/util/HashedWheelTimer.java
+++ b/common/src/main/java/io/netty/util/HashedWheelTimer.java
@@ -706,8 +706,9 @@ public class HashedWheelTimer implements Timer {
 
             // process all timeouts
             while (timeout != null) {
-                boolean remove = false;
+                HashedWheelTimeout next = timeout.next;
                 if (timeout.remainingRounds <= 0) {
+                    next = remove(timeout);
                     if (timeout.deadline <= deadline) {
                         timeout.expire();
                     } else {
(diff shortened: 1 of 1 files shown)


Scottmitch:
@normanmaurer - PTAL

Scottmitch:
discovered in https://garage.netty.io/teamcity/viewLog.html?buildId=13723&buildTypeId=pulls_netty&tab=buildLog while building https://github.com/netty/netty/pull/6024

Scottmitch:
note that this may allow the race condition in the other direction. we haven't technically executed the timer task but we will decrement the total count ... so this change may potentially allow more timers to exist at any given time ... however it seems more useful to apply the limit on the queue and remove the task before executing. for example if the limit is hit, and as a result of a timer task firing a user wants to add another timer ... then without this change the timer would not be added.

normanmaurer:
@Scottmitch LGTM

Scottmitch:
another failure building https://github.com/netty/netty/pull/6025 https://garage.netty.io/teamcity/viewLog.html?buildId=13784&buildTypeId=pulls_netty&guest=1

Scottmitch:
4.1 (a0e375b) 4.0 (dbc0808)

PR #6039, opened 2016-11-18
Consistency between pooled/unpooled derived buffers

Scottmitch:
Motivation:
4bba7526e2f58018817972f38279cc232f519100 introduced changes which made pooled and unpooled derived buffers inconsistent in a few ways:
- Pooled derived buffers always generated a duplicate buffer when duplicate() was called and always generated a sliced buffer when slice() was called. Unpooled derived buffers some times generated a sliced buffer when duplicate() was called.
- The indexes that were set for duplicate buffers generated from slices were not always consistent.
There were also some various bugs in the derived pooled buffer implementation.

Modifications:
- Make pooled/unpooled consistently generate duplicate buffers when duplicate() is called and sliced buffers when slice() is called.
- Fix bugs in the derived pooled buffer

Result:
More consistent behavior from the derived pooled/unpooled buffers.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/AbstractPooledDerivedByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/AbstractPooledDerivedByteBuf.java
@@ -269,15 +269,13 @@ abstract class AbstractPooledDerivedByteBuf extends AbstractReferenceCountedByte
 
         @Override
         public ByteBuf duplicate() {
-            // Capacity is not allowed to change for a sliced ByteBuf, so length == capacity()
-            final ByteBuf duplicate = slice(0, capacity());
-            duplicate.setIndex(readerIndex(), writerIndex());
-            return duplicate;
+            return new PooledNonRetainedDuplicateByteBuf(referenceCountDelegate, unwrap())
+                    .setIndex(idx(readerIndex()), idx(writerIndex()));
         }
 
         @Override
         public ByteBuf retainedDuplicate() {
-            return PooledDuplicatedByteBuf.newInstance(unwrap(), this, readerIndex(), writerIndex());
+            return PooledDuplicatedByteBuf.newInstance(unwrap(), this, idx(readerIndex()), idx(writerIndex()));
         }
 
         @Override
--- a/buffer/src/main/java/io/netty/buffer/AbstractUnpooledSlicedByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/AbstractUnpooledSlicedByteBuf.java
@@ -216,9 +216,7 @@ abstract class AbstractUnpooledSlicedByteBuf extends AbstractDerivedByteBuf {
 
     @Override
     public ByteBuf duplicate() {
-        final ByteBuf duplicate = unwrap().slice(adjustment, length());
-        duplicate.setIndex(readerIndex(), writerIndex());
-        return duplicate;
+        return unwrap().duplicate().setIndex(idx(readerIndex()), idx(writerIndex()));
     }
 
     @Override
--- a/buffer/src/test/java/io/netty/buffer/SlicedByteBufTest.java
+++ b/buffer/src/test/java/io/netty/buffer/SlicedByteBufTest.java
@@ -16,6 +16,7 @@
 package io.netty.buffer;
 
 import io.netty.util.internal.ThreadLocalRandom;
+import org.junit.Ignore;
 import org.junit.Test;
 
 import java.io.IOException;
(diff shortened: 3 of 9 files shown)


Scottmitch:
@normanmaurer - PTAL. Parts of this will be back ported to 4.0.

buffer/src/main/java/io/netty/buffer/AbstractPooledDerivedByteBuf.java
```diff
@@ -269,15 +269,13 @@ boolean release0(int decrement) {
 
         @Override
         public ByteBuf duplicate() {
-            // Capacity is not allowed to change for a sliced ByteBuf, so length == capacity()
-            final ByteBuf duplicate = slice(0, capacity());
-            duplicate.setIndex(readerIndex(), writerIndex());
-            return duplicate;
+            return new PooledNonRetainedDuplicateByteBuf(referenceCountDelegate, unwrap())
```
netkins: MAJOR Prefix this call to "unwrap" with "super.". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2388)

buffer/src/main/java/io/netty/buffer/AbstractPooledDerivedByteBuf.java
```diff
@@ -269,15 +269,13 @@ boolean release0(int decrement) {
...
+                    .setIndex(idx(readerIndex()), idx(writerIndex()));
         }
 
         @Override
         public ByteBuf retainedDuplicate() {
-            return PooledDuplicatedByteBuf.newInstance(unwrap(), this, readerIndex(), writerIndex());
+            return PooledDuplicatedByteBuf.newInstance(unwrap(), this, idx(readerIndex()), idx(writerIndex()));
```
netkins: MAJOR Prefix this call to "unwrap" with "super.". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2388)

buffer/src/main/java/io/netty/buffer/AbstractPooledDerivedByteBuf.java
Scottmitch: this class is static

buffer/src/main/java/io/netty/buffer/AbstractUnpooledSlicedByteBuf.java
```diff
@@ -214,11 +214,13 @@ protected long _getLongLE(int index) {
...
+    @SuppressWarnings("deprecation")
     @Override
     public ByteBuf duplicate() {
-        final ByteBuf duplicate = unwrap().slice(adjustment, length());
-        duplicate.setIndex(readerIndex(), writerIndex());
-        return duplicate;
+        final ByteBuf unwrapped = unwrap();
+        return (unwrapped instanceof AbstractByteBuf ?
+                new UnpooledDuplicatedByteBuf((AbstractByteBuf) unwrapped) :
+                new DuplicatedByteBuf(unwrapped)).setIndex(idx(readerIndex()), idx(writerIndex()));
```
normanmaurer: @Scottmitch why we can't jus use `duplicate` as before ?

buffer/src/test/java/io/netty/buffer/SlicedByteBufTest.java
```diff
@@ -138,16 +138,12 @@ public void testForEachByte2() {
...
-    @Test(expected = UnsupportedOperationException.class)
     @Override
     public void testDuplicateCapacityChange() {
-        super.testDuplicateCapacityChange();
     }
 
-    @Test(expected = UnsupportedOperationException.class)
     @Override
     public void testRetainedDuplicateCapacityChange() {
-        super.testRetainedDuplicateCapacityChange();
```
normanmaurer: @Scottmitch Please add a comment in the empty body why this is done so we remember later as well (for all of these methods).

Scottmitch: > for all of these methods

I added comments for the 2 that I changed ... is that what you meant by `all`?

normanmaurer: yes

buffer/src/main/java/io/netty/buffer/AbstractUnpooledSlicedByteBuf.java
normanmaurer: I was using why we return a different `ByteBuf` implementation depending on the type and can't just call `unwrap.duplicate().setIndex(...)` here.

Scottmitch:
@normanmaurer - comments addressed. PTAL

buffer/src/test/java/io/netty/buffer/SlicedByteBufTest.java
normanmaurer: @Scottmitch or we could also use `@Ignore("Sliced ByteBuf objects don't allow the capacity to change. So this test would fail and shouldn't be run")` . Maybe this would be even better... WDYT ?

Scottmitch: sgtm

Scottmitch:
@normanmaurer - comments resolved again.

normanmaurer:
@Scottmitch w000t... ship it

Scottmitch:
4.1 (9306333) 4.0 (88c41d8)

Issue #775, opened 2012-11-28
AbstractNioWorker does not exit its event loop when the Executor is not an ExecutorService


drewhk:
If netty is used with an Executor that does not implement ExecutorService then the worker threads are never released and remain running. The problem is the exit condition in the run() method:

if (shutdown || executor instanceof ExecutorService && ((ExecutorService) executor).isShutdown()) {

btw, the local variable shutdown is never set to true

(Netty version 3.5.8)

PR #6041, opened 2016-11-19
Fix HTTP/2 test cleanup with LocalChannel

Scottmitch:
Motivation:
When a LocalChannel is closed it is responsible to ensure all queued objects are released. When a LocalChannel is closed it will also close its peer channel. However in HTTP/2 unit tests we may not wait until all channels have completed the shutdown process before destroying the threads and exiting the test. This may mean buffers are GCed before they are released and be reported as a leak.

Modifications:
- In HTTP/2 tests when we use LocalChannel we should wait for all channels to close before exiting the test and cleaning up the associated EventLoopGroups.

Result:
More correct usage of LocalChannel in HTTP/2 tests.


Scottmitch:
this should help with #775

Scottmitch:
@normanmaurer - FYI

normanmaurer:
LGTM

Scottmitch:
4.1 (782b7bc)

Issue #6038, opened 2016-11-18
4.0.42: java.lang.NoClassDefFoundError: Could not initialize class io.netty.util.internal.PlatformDependent0


jesperpedersen:
In WildFly (a JavaEE server) I get the following during init using Netty 4.0.42
```
Caused by: java.lang.NoClassDefFoundError: Could not initialize class io.netty.util.internal.PlatformDependent0
        at io.netty.util.internal.PlatformDependent.getSystemClassLoader(PlatformDependent.java:694)
        at io.netty.channel.nio.NioEventLoop$4.run(NioEventLoop.java:171)
        at java.security.AccessController.doPrivileged(Native Method) [rt.jar:1.8.0_91]
        at io.netty.channel.nio.NioEventLoop.openSelector(NioEventLoop.java:167)
        at io.netty.channel.nio.NioEventLoop.<init>(NioEventLoop.java:149)
        at io.netty.channel.nio.NioEventLoopGroup.newChild(NioEventLoopGroup.java:102)
        at io.netty.util.concurrent.MultithreadEventExecutorGroup.<init>(MultithreadEventExecutorGroup.java:64)
        at io.netty.channel.MultithreadEventLoopGroup.<init>(MultithreadEventLoopGroup.java:49)
        at io.netty.channel.nio.NioEventLoopGroup.<init>(NioEventLoopGroup.java:70)
        at io.netty.channel.nio.NioEventLoopGroup.<init>(NioEventLoopGroup.java:65)
        at io.netty.channel.nio.NioEventLoopGroup.<init>(NioEventLoopGroup.java:56)
```

The jar contains
```
io/netty/util/internal/PlatformDependent0.class
io/netty/util/internal/PlatformDependent0$4.class
io/netty/util/internal/PlatformDependent0$5.class
io/netty/util/internal/PlatformDependent0$7.class
io/netty/util/internal/PlatformDependent0$6.class
io/netty/util/internal/PlatformDependent0$3.class
io/netty/util/internal/PlatformDependent.class
io/netty/util/internal/PlatformDependent0$8.class
io/netty/util/internal/PlatformDependent0$1.class
io/netty/util/internal/PlatformDependent$Mpsc.class
io/netty/util/internal/PlatformDependent$1.class
io/netty/util/internal/PlatformDependent$AtomicLongCounter.class
io/netty/util/internal/PlatformDependent$Mpsc$1.class
io/netty/util/internal/PlatformDependent0$2.class
```

[...]

Issue #6034, opened 2016-11-17
PooledByteBufAllocatorTest#testConcurrentUsage ByteBuf leak


Scottmitch:
https://garage.netty.io/teamcity/viewLog.html?buildId=13755&buildTypeId=pulls_netty&guest=1
found while building https://github.com/netty/netty/pull/6024

```
[01:24:54] :             [io.netty:netty-buffer] 01:24:54.156 [Thread-223] ERROR io.netty.util.ResourceLeakDetector - LEAK: ByteBuf.release() was not called before it's garbage-collected. See http://netty.io/wiki/reference-counted-objects.html for more information.^M
[01:24:54] :             [io.netty:netty-buffer] Recent access records: 0^M
[01:24:54] :             [io.netty:netty-buffer] Created at:^M
[01:24:54] :             [io.netty:netty-buffer]        io.netty.buffer.PooledByteBufAllocator.newDirectBuffer(PooledByteBufAllocator.java:271)^M
[01:24:54] :             [io.netty:netty-buffer]        io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:179)^M
[01:24:54] :             [io.netty:netty-buffer]        io.netty.buffer.PooledByteBufAllocatorTest$AllocationThread.run(PooledByteBufAllocatorTest.java:352)^M
```

normanmaurer:
Will have a look

Scottmitch:
Happened again while building PR https://github.com/netty/netty/pull/6042

https://garage.netty.io/teamcity/viewLog.html?buildId=13825&buildTypeId=pulls_netty&tab=buildLog

```
[02:23:34] :             [io.netty:netty-buffer] 02:23:34.817 [Thread-2335] ERROR io.netty.util.ResourceLeakDetector - LEAK: ByteBuf.release() was not called before it's garbage-collected. See http://netty.io/wiki/reference-counted-objects.html for more information.^M
[02:23:34] :             [io.netty:netty-buffer] Recent access records: 0^M
[02:23:34] :             [io.netty:netty-buffer] Created at:^M
[02:23:34] :             [io.netty:netty-buffer]        io.netty.buffer.PooledByteBufAllocator.newDirectBuffer(PooledByteBufAllocator.java:271)^M
[02:23:34] :             [io.netty:netty-buffer]        io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:179)^M
[...]

PR #6042, opened 2016-11-19
Catch exceptions from PlatformDependent#getSystemClassLoader

Scottmitch:
Motivation:
PlatformDependent#getSystemClassLoader may throw a wide variety of exceptions based upon the environment. We should handle all exceptions and continue initializing the slow path if an exception occurs.

Modifications:
- Catch Throwable in cases where PlatformDependent#getSystemClassLoader is used

Result:
Fixes https://github.com/netty/netty/issues/6038

### DIFF
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent.java
@@ -1163,9 +1163,11 @@ public final class PlatformDependent {
 
     private static long maxDirectMemory0() {
         long maxDirectMemory = 0;
+        ClassLoader systemClassLoader = null;
         try {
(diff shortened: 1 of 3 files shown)


Scottmitch:
@jesperpedersen - can you verify this PR fixes the issue?

Scottmitch:
build failure due to #6034

normanmaurer:
LGTM

Scottmitch:
4.1 (a043cf4) 4.0 (e5ac9d3)

jesperpedersen:
Yes, this works - thanks !

Scottmitch:
np @jesperpedersen. thanks for verifying!

Issue #6023, opened 2016-11-15
HTTP/2 LastInboundHandler queue behavior


Scottmitch:
LastInboundHandler maintains 2 queues which may contain "inbound messages". 1 queue which just contains "inbound messages" (`inboundMessages` [1]) and another queue that contains "inbound messages" and "user events" (`inboundMessagesAndUserEvents` [2]). The `readInbound()` method attempts to de-dup from these 2 queues by assuming that if an event is on the  `inboundMessages` queue it will also be the next element on the `inboundMessagesAndUserEvents` queue. It seems like this assumption does not necessarily hold and can lead to unexpected interactions in unit tests (`Http2FrameCodecTest#entityRequestEntityResponse`) where an `Http2DataFrame` is read via `readInbound()`, popped from the `inboundMessages`, but still remains in the `inboundMessagesAndUserEvents` queue. The side effect of this is `Http2DataFrame` cannot be released by the caller of `readInbound()` because it may be released later by `finishAndReleaseAll` [4].

Originally discussed here:
https://github.com/netty/netty/pull/5996#discussion_r87312089

[1] https://github.com/netty/netty/blob/4.1/codec-http2/src/test/java/io/netty/handler/codec/http2/LastInboundHandler.java#L36
[2] https://github.com/netty/netty/blob/4.1/codec-http2/src/test/java/io/netty/handler/codec/http2/LastInboundHandler.java#L38
[3] https://github.com/netty/netty/blob/4.1/codec-http2/src/test/java/io/netty/handler/codec/http2/LastInboundHandler.java#L103
[4] https://github.com/netty/netty/blob/4.1/codec-http2/src/test/java/io/netty/handler/codec/http2/LastInboundHandler.java#L153

Scottmitch:
@ejona86 @buchgr - fyi

buchgr:
IIRC `LastInboundHandler` got more "popular" within the test code than it was designed to. I believe `inboundMessagesAndUserEvents` is not even used anymore? The idea was for one test case to use either `inboundMessages` or `inboundMessagesAndUserEvents` but never both. It was initially a private inner class, but was then moved to it's own file as several tests used it.

normanmaurer:
@Scottmitch @buchgr @nmittler let me investigate tomorrow

normanmaurer:
looking into now... stay tuned

normanmaurer:
Fixed by https://github.com/netty/netty/pull/6044

PR #6044, opened 2016-11-19
[#6023] Ensure LastInboundHandler correctly handle different events /…

normanmaurer:
… data

Motivation:

LastInboundHandler maintains 2 queues which may contain the same data and tries to match these up when you read elements out of it. Because of this it can happen that you remove an element only out of one queue and so double free stuff later.

Modifications:

Just use one "queue" to store things.

Result:

Not possible to only remove things from one queue and so get into trouble later when release everything that sits in the handler.


normanmaurer:
@Scottmitch @nmittler @buchgr PTAL

Scottmitch:
build failure due to https://github.com/netty/netty/issues/6048

normanmaurer:
Cherry-picked into 4.1 (e5070b726613cc6c1297ce2cc2dd5ab377cf0fbc)

PR #6045, opened 2016-11-19
Deprecate HttpUtil.getCharsetAsString(...) and introduce HttpUtil.get…

normanmaurer:
…CharsetAsSequence(...).

Motivation:

The method HttpUtil.getCharsetAsString(...) is missleading as its return type is CharSequence and not String.

Modifications:

Deprecate HttpUtil.getCharsetAsString(...) and introduce HttpUtil.getCharsetAsSe
quence(...).

Result:

Less confusing method name.


normanmaurer:
@Scottmitch PTAL

normanmaurer:
Cherry-picked as 0b3122d8ff632e409524b6f3d54435156e70ffa2 into 4.1

PR #6046, opened 2016-11-19
Use the correct alert depending on the CertificateException when usin…

normanmaurer:
…g OpenSslEngine

Motivation:

We tried to detect the correct alert to use depending on the CertificateException that is thrown by the TrustManager. This not worked all the time as depending on the TrustManager implementation it may also wrap a CertPathValidatorException.

Modification:

- Try to unwrap the CertificateException if needed and detect the right alert via the CertPathValidatorException.
- Add unit to verify

Result:

Send the correct alert depending on the CertificateException when using OpenSslEngine.

### DIFF
--- a/handler/src/test/java/io/netty/handler/ssl/SslErrorTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/SslErrorTest.java
@@ -1,0 +1,248 @@
+/*
+ * Copyright 2016 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.ssl;
+
+import io.netty.bootstrap.Bootstrap;
+import io.netty.bootstrap.ServerBootstrap;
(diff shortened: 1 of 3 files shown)


normanmaurer:
@Scottmitch PTAL

handler/src/test/java/io/netty/handler/ssl/SslErrorTest.java
```diff
@@ -0,0 +1,248 @@
...
+                serverChannel.close().syncUninterruptibly();
+            }
+            group.shutdownGracefully();
+
+            ReferenceCountUtil.release(sslServerCtx);
+            ReferenceCountUtil.release(sslClientCtx);
+        }
+    }
+
+    // Its a bit hacky to verify against the message that is part of the exception but there is no other way
```
Scottmitch: agreed it is hacky ... maybe we can just validate the exception type for now and enhance later if JDK provides us with more granular exceptions, or a more reliable error code indicator?

normanmaurer: Nope as the exception type was the same before, so let us keep it

Scottmitch: this test may be JDK dependent though right? Also what about exception test localization?

normanmaurer: I am on the fence here... as without the message check it passes also without the fix

Scottmitch: Ok your call. Unfortunate that we don't have a more reliable way to verify. Can we get the alert code from an OpenSSL API maybe?

normanmaurer: @Scottmitch let me just pull it in as it is then... there is no way atm.

normanmaurer:
Cherry-picked into 4.1 (9b7fb2f) and 4.0 (139af2aac4908171122f348b6cf39159e49fc607)

PR #6047, opened 2016-11-19
Ensure alert is send when SSLException happens during calling SslHand…

normanmaurer:
…ler.unwrap(...)

Motivation:

When the SslHandler.unwrap(...) (which is called via decode(...)) method did produce an SSLException it was possible that the produced alert was not send to the remote peer. This could lead to staling connections if the remote peer did wait for such an alert and the connection was not closed.

Modifications:

- Ensure we try to flush any pending data when a SSLException is thrown during unwrapping.
- Fix SniHandlerTest to correct test this
- Add explicit new test in SslHandlerTest to verify behaviour with all SslProviders.

Result:

The alert is correctly send to the remote peer in all cases.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -523,9 +523,19 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
         if (startTls && !sentFirstMessage) {
             sentFirstMessage = true;
             pendingUnencryptedWrites.removeAndWriteAll();
-            ctx.flush();
+            forceFlush(ctx);
             return;
         }
+
+        try {
+            wrapAndFlush(ctx);
+        } catch (Throwable cause) {
+            setHandshakeFailure(ctx, cause);
+            PlatformDependent.throwException(cause);
+        }
+    }
+
+    private void wrapAndFlush(ChannelHandlerContext ctx) throws SSLException {
         if (pendingUnencryptedWrites.isEmpty()) {
             // It's important to NOT use a voidPromise here as the user
             // may want to add a ChannelFutureListener to the ChannelPromise later.
@@ -538,18 +548,14 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
         }
         try {
             wrap(ctx, false);
-        } catch (Throwable cause) {
-            // Fail pending writes.
-            pendingUnencryptedWrites.removeAndFailAll(cause);
-
-            PlatformDependent.throwException(cause);
         } finally {
             // We may have written some parts of data before an exception was thrown so ensure we always flush.
             // See https://github.com/netty/netty/issues/3900#issuecomment-172481830
-            ctx.flush();
+            forceFlush(ctx);
         }
     }
 
+    // This method will not call setHandshakeFailure(...) !
     private void wrap(ChannelHandlerContext ctx, boolean inUnwrap) throws SSLException {
         ByteBuf out = null;
         ChannelPromise promise = null;
@@ -607,9 +613,6 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
                     }
                 }
             }
-        } catch (SSLException e) {
-            setHandshakeFailure(ctx, e);
-            throw e;
         } finally {
             finishWrap(ctx, out, promise, inUnwrap, needUnwrap);
         }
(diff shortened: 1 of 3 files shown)


normanmaurer:
@Scottmitch @trustin PTAL

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -538,18 +548,14 @@ public void flush(ChannelHandlerContext ctx) throws Exception {
         }
         try {
             wrap(ctx, false);
-        } catch (Throwable cause) {
-            // Fail pending writes.
-            pendingUnencryptedWrites.removeAndFailAll(cause);
```
Scottmitch: can you clarify why we no longer want to fail/release all pending writes in the event of an Throwable (other than `SSLException`)? Are we sure there will be no unchecked exceptions, and if there is will we potentially leak?

normanmaurer: Yes there should only be SSLException.

Scottmitch: I'm a bit more pessimistic I guess ... we are calling the JDK's SSL APIs which can be backed by a boat load of complexity ... can you make me more comfortable and explain why none of these APIs will throw an unchecked exception under any condition, or if they do that we will handle it gracefully in some other way.

normanmaurer: I am just not think catching throwable all the time is hacky and inconsistent. If we really do we should at least do it in a consistent way. So you would prefer this?

Scottmitch: We should make a best effort to keep our state consistent and let higher layers determine if the condition is recoverable or not. If we want to re-throw then that is fine, but if we let control flow escape then all bets are off, and the upper layers have even less chance of recovering (without knowing this is the case). I agree it is unlikely to recover in the event of OOME but since this is part of the control flow it seems strange to ignore it. Its not the same but I try to draw an analogy to a non-memory managed environment:

```c
int* x = malloc(10 * sizeof(int));
// assume this worked.
int* y = malloc(10 * sizeof(int));
if (y == NULL) {
  // what would you do here?
  // generally you would free `x` and return some sort of error condition.
  // isn't this similar to an OOME in a non-memory managed environment?
}
```

johnou: Exactly, but consider that part of the code might be in the application, if Netty swallows the OOME the application won't have the opportunity to handle it.

Scottmitch: As I said we can rethrow.

normanmaurer: @Scottmitch so you want me to catch Throwable and rethrow ? I am not strong here...

normanmaurer: @johnou just like it was before `PlatformDependent.throwException(cause);`.

johnou: totally missed that, was burried under all the comments,

normanmaurer:
@johnou @Scottmitch PTAL at the followup commit https://github.com/netty/netty/pull/6047/commits/3396ada9e098c9406b5b001625a9207ce86359a0 . Good to go now ?

normanmaurer:
Nope

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
Scottmitch: looking at your commit now 🍺

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1389,9 +1400,10 @@ public void operationComplete(Future<Channel> future) throws Exception {
         try {
             engine.beginHandshake();
             wrapNonAppData(ctx, false);
-            ctx.flush();
-        } catch (Exception e) {
-            notifyHandshakeFailure(e);
+        } catch (SSLException e) {
```
Scottmitch: either change this to `Throwable` or add another catch clause and re-throw to ensure we free our resources?

normanmaurer: @Scottmitch ok can do...

normanmaurer:
PTAL again @Scottmitch

Scottmitch:
thanks @normanmaurer ship it!

normanmaurer:
Cherry-picked into 4.1 (f289ebf7fab65c9a9bf491a5060ce75c294ca60c) and 4.0 (1bb4c90bd9b01198494d659f62a2db4238f4950b)

Issue #6049, opened 2016-11-19
Testfailure in ClientCookieDecoderTest


normanmaurer:
https://garage.netty.io/teamcity/viewLog.html?buildId=13846&buildTypeId=netty_build_oraclejdk7

java.lang.AssertionError: expected: 50, actual: -9223372036854775808
    at io.netty.handler.codec.http.cookie.ClientCookieDecoderTest.testDecodingSingleCookieV0(ClientCookieDecoderTest.java:50)

@slandelle can you have a look?

slandelle:
I haven't been able to reproduce so far.

I even turned the test into a JMH benchmark with no avail:
```java
@Benchmark
public Cookie test() {
    String cookieString = "myCookie=myValue;expires=XXX;path=/apathsomewhere;domain=.adomainsomewhere;secure;";
    cookieString = cookieString.replace("XXX",
            DateFormatter.format(new Date(System.currentTimeMillis() + 50000)));

    Cookie cookie = ClientCookieDecoder.STRICT.decode(cookieString);

    boolean fail = true;
    for (int i = 40; i <= 60; i++) {
        if (cookie.maxAge() == i) {
            fail = false;
            break;
        }
    }
    if (fail) {
        throw new IllegalArgumentException("expected: 50, actual: " + cookie.maxAge());
    }

    return cookie;
}
```

slandelle:
@normanmaurer Could you please merge my pending PR, before I send another one to improve the test and try to figure out what's going on?

slandelle:
@normanmaurer Actually, I wonder if this failure really happened on the version that got finally merged. This failure might have been caused by an intermediate version that got fixed before final cherry-pick.

<img width="539" alt="teamcity" src="https://cloud.githubusercontent.com/assets/849197/20465795/6e062ec8-af66-11e6-9386-820e49b5d945.png">

radai-rosenblatt:
happens to me, at 782b7bcf4a1db5f33078fe8f7ded8059770f7fc4
`cookieString="myCookie=myValue;expires=Mon, 21 Nov 2016 00:36:11 GMT;path=/apathsomewhere;domain=.adomainsomewhere;secure;"`

it eventually fails to find the time part out of that expired string.

radai-rosenblatt:
after some further investigation the issue is parsing time stamps around midnight - "00:**:**".
this triggers the following code in HttpHeaderDateFormatter:
```
            char c = txt.charAt(i);
            if (isDigit(c)) {
[...]

PR #6050, opened 2016-11-21
Fix timestamp parsing in HttpHeaderDateFormatter

radai-rosenblatt:
Motivation:
code assumes a numeric value of 0 means no digits were read between separators, which fails for timestamps like 00:00:00

Modifications:
explicitly check for no digits encountered instead of checking for numeric value 0

Result:
timestamps with 00 successfully parse

Signed-off-by: radai-rosenblatt <radai.rosenblatt@gmail.com>

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
@@ -188,40 +188,43 @@ public final class HttpHeaderDateFormatter {
         int localSeconds = -1;
         int currentPartNumber = 0;
         int currentPartValue = 0;
+        int numDigits = 0;
 
         for (int i = tokenStart; i < tokenEnd; i++) {
             char c = txt.charAt(i);
             if (isDigit(c)) {
                 currentPartValue = currentPartValue * 10 + getNumericalValue(c);
+                if (++numDigits > 2) {
+                  return false; // too many digits in this part
+                }
             } else if (c == ':') {
-                if (currentPartValue == 0) {
-                    // invalid :: (nothing in between)
+                if (numDigits == 0) {
+                    // no digits between separators
                     return false;
                 }
                 switch (currentPartNumber) {
                     case 0:
                         // flushing hours
                         localHours = currentPartValue;
-                        currentPartValue = 0;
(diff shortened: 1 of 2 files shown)


radai-rosenblatt:
this resolves #6049

Scottmitch:
thanks @radai-rosenblatt! Can you add a unit test for the failure case too please?

radai-rosenblatt:
@Scottmitch - I've added tests to account for both (legal) midnight timestamps and timestamps of the form 0:000:0 which i found out the current code accepted.

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
```diff
@@ -188,40 +188,43 @@ private boolean tryParseTime(CharSequence txt, int tokenStart, int tokenEnd) {
...
             char c = txt.charAt(i);
             if (isDigit(c)) {
                 currentPartValue = currentPartValue * 10 + getNumericalValue(c);
+                if (++numDigits > 2) {
+                  return false; // too many digits in this part
+                }
             } else if (c == ':') {
-                if (currentPartValue == 0) {
-                    // invalid :: (nothing in between)
+                if (numDigits <= 0) {
```
slandelle: `== 0` should be enough

slandelle:
@radai-rosenblatt Thanks!

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderDateFormatter.java
radai-rosenblatt: yes, but there's no such thing as too paranoid :-). will fix

normanmaurer:
@radai-rosenblatt please squash... will cherry-pick once CI build completes.

radai-rosenblatt:
@normanmaurer - squash? its a single commit on top of current upstream head (i force pushed all the updates)

normanmaurer:
@radai-rosenblatt doh! I need more coffee... all good :(

normanmaurer:
Cherry-picked into 4.1 (886a7aae465cceb5f90ede01d98de31e35ae404a).

@radai-rosenblatt thanks!

Issue #5152, opened 2016-04-15
Throughput is low even with batching


maseev:
Hi, guys.

I created a simple client-server [app](https://github.com/maseev/clientserver) that sole purpose is to transfer 200 bytes messages as fast as possible between the client and the server.

During the tests, it turned out that the throughput isn't that high as I expected (about 20-30 Mb/sec on read/write), even though I call ```flush``` method on every 100th message:

![selection_001](https://cloud.githubusercontent.com/assets/3434800/14573847/63c80114-0379-11e6-9ba5-57f52e0ca8a1.png)

The [server](https://github.com/maseev/clientserver/blob/master/src/main/java/io/github/maseev/server/nio/Server.java) logic is pretty straightforward and doesn't do a lot (just batching messages and flush them when the counter hits the batch threshold).

I didn't see much difference by changing ```WRITE_BUFFER_HIGH_WATER_MARK``` and ``WRITE_BUFFER_LOW_WATER_MARK``` values either.
I could implement batching on the client side (that would definitely improve the overall throughput dramatically), but the thing is I can't really do this because the real client will be a third-party closed-source application.

According to this [ticket](https://github.com/netty/netty/issues/1759), I'm not the only one who's trying to implement a smart batching algorithm.
Am I missing something? Maybe there's some general approach how you can increase throughput in a Netty application which is supposed to transfer a gigantic amount of small messages (less than 300 bytes each)?

I couldn't find any answers in the Netty in Action book, as well as on StackOverflow. Netty's documentation (user guide, to be precise, which by the way only describes basic topics without going into details) doesn't say much about this particular use case.

johnou:
Try context.write(msg, context.voidPromise());

maseev:
@johnou, no luck. Throughput is still very low. The funny this is, I've got another application which is using Java NIO directly and does the same thing - reads data from the client, batches it and then sends it back to the client. For some reasons, it performs much better:

![screen](https://cloud.githubusercontent.com/assets/3434800/14579316/37c5a98e-03c6-11e6-8302-f44d2a3c4b13.png)

ninja-:
@maseev I can give you some quick tips from my experience in similiar kind of netty apps...
1. use epoll (for example, there's a cool feature that writes all pending buffers in one syscall instead of N write(...) calls)
2. try not touching these if there's no clear improvement  

        .childOption(ChannelOption.WRITE_BUFFER_HIGH_WATER_MARK, HIGH_WATERMARK)
[...]

Issue #1759, opened 2013-08-19
Optimize writeAndFlush for lots of small messages (do smart autoflush)


stepancheg:
Typical server code looks like this:

```
void channelRead(ChannelHandlerContext ctx, Object msg) {
    ctx.writeAndFlush(doSomething(msg));
}
```

This code is very ineffective when size of response is small, and number of request/responses are huge. Because this code seems to issue `send` syscall after each `writeAndFlush` operation.

There is an easy way to implement `writeAndFlush` efficiently. I implemented it on top of Netty, but I think it should be implemented inside of Netty.

Algorithm in pseudocode is this:

```
class Channel {
    ...
    // with lock-free-queue we don't need locks
    LockFreeQueue<Object> queue = new LockFreeQueue();

    // 3 states of write-queue-and-flush tasks:
    // * not executing
    // * executing
    // * executing, and should rerun one more time
    Tasks tasks = new Tasks();

    public void writeAndFlush(Object msg) {
        // enqueue message
        queue.enqueue(msg);
        if (tasks.addTask()) {
            // schedule write-queue-and-flush task if we have no one now
            executor.submit(this::writeQueueAndFlush);
        }
    }

    private void writeQueueAndFlush() {
        // while clients submit us tasks
        while (tasks.fetchTask()) {
            // we write these tasks, but do not flush
            List<Object> messages = queue.dequeueAll();
            for (message : messages) {
                doWrite(message);
            }
        }
        // and finally flush
        doFlush();
    }
}
```

I have sample project:

https://github.com/stepancheg/netty-td/

It has an implementation of this algorithm on top of Netty: [BetterWriter.write method](https://github.com/stepancheg/netty-td/blob/master/src/com/github/stepancheg/nettytd/BetterWrite.java#L56).

[Tasks](https://github.com/stepancheg/netty-td/blob/master/src/com/github/stepancheg/nomutex/tasks/framework/Tasks.java) class is lock-free helper to schedule not more than one write-queue-and-schedule task.

[LockFreeQueue implementation](https://github.com/stepancheg/netty-td/blob/master/src/com/github/stepancheg/nomutex/tasks/framework/LockFreeStackWithSize.java) (it is actually lock-free-stack; lock-free-stack is faster, and difference is not important for this issue).

Edit: [simpler implementation using atomics](https://github.com/stepancheg/netty-td/blob/master/src/com/github/stepancheg/nettytd/BetterWriteWithAtomic.java).

[...]

PR #5716, opened 2016-08-17
Auto-flush for channels. (`ChannelHandler` implementation)

NiteshKant:
This is an alternate approach to #5395. As suggested in the other PR by @buchgr, this change implements auto-flush as a `ChannelHandler` as opposed to a `ChannelOption`.

#### Motivation

  Netty is extremely flexible in providing a way to batch writes on the physical socket by providing a means to `write` and `flush` writes separately. This works for the usecases when the application pattern is one that provides predictable finite streams of data. In cases, when an infinite stream of data is to be written on a channel which does not have any predictable length and duration of writes, a user has the following choices:

 - Flush every write. 
 - Flush batches based on time and count (eg: Flush every 10 message or 1 second)   

Neither of the above is ideal as the nature of stream is completely unpredictable. Flushing every write is costly as `flush` is a system call. Flushing batches is arbitrary and has to be deviced for each stream in an arbitrary way.  

#### Modifications

A new handler `AutoFlushHandler` is added which must be added as the first handler in the `ChannelPipeline` to achieve this batching technique.
This handler flushes writes at the end of an eventloop iteration and when the channel becomes unwritable.

#### Result

This results in a much better model for infinite streams written to a channel.

I wrote a micro-benchmark(`io.netty.microbench.channel.AutoFlushBenchmark`) to demonstrate the throughput variations between all possible current alternatives.
Here are the results for the same:

```
Benchmark                                       (autoFlush)  (writeCount)   Mode  Cnt      Score      Error  Units
AutoFlushBenchmark.compareWithFlushAtEnd               true             1  thrpt   10  47460.595 ±  884.360  ops/s
AutoFlushBenchmark.compareWithFlushAtEnd               true            10  thrpt   10  41769.552 ±  509.156  ops/s
AutoFlushBenchmark.compareWithFlushAtEnd               true           100  thrpt   10  11431.234 ±  124.230  ops/s
AutoFlushBenchmark.compareWithFlushAtEnd              false             1  thrpt   10  47029.181 ± 1324.622  ops/s
AutoFlushBenchmark.compareWithFlushAtEnd              false            10  thrpt   10  41443.218 ±  569.993  ops/s
AutoFlushBenchmark.compareWithFlushAtEnd              false           100  thrpt   10  11186.934 ±  151.784  ops/s
AutoFlushBenchmark.compareWithFlushEvery5              true             1  thrpt   10  46541.047 ±  779.492  ops/s
[...]

PR #6052, opened 2016-11-21
 Implemented batching versions of LengthFieldBasedFrameDecoder and Fixed…

maseev:
…LengthFrameDecoder

Motivation:

According to performance tests, LengthFieldBasedFrameDecoder and FixedLengthFrameDecoder generate a lot of garbage as well as perform quite poorly in cases where clients send data at a very high-speed rate.

Modifications:

- Introduce a base class which implements batching of all incoming ByteBuf messages as well as provides a convenient API for decoding the resulting chunk of data
- Introduce an optimised version of LengthFieldBasedFrameDecoder - BatchLengthFieldBasedFrameDecoder
- Introduce an optimised version of FixedLengthFrameDecoder - BatchFixedLengthFrameDecoder

Result:

BatchLengthFieldBasedFrameDecoder and BatchFixedLengthFrameDecoder will provide much better throughput as compared to LengthFieldBasedFrameDecoder and FixedLengthFrameDecoder

P.S I [created](https://github.com/maseev/echo) a simple echo application which tests three batching strategies. Here are some throughput as well as GC diagrams:

[Dummy Handler](https://github.com/maseev/echo/blob/master/server/src/main/java/io/github/maseev/echo/handler/DummyEchoServerHandler.java)
Dummy Handler Throughput
Dummy Handler GC

[Batching Handler](https://github.com/maseev/echo/blob/master/server/src/main/java/io/github/maseev/echo/handler/BatchingEchoServerHandler.java)
Batching Handler Throughput
Batching Handler GC

[Optimised Handler](https://github.com/maseev/echo/blob/master/server/src/main/java/io/github/maseev/echo/handler/HighPerformanceEchoServerHandler.java)
Optimised Throughput
Optimised GC

This particular handler mimics the semantic of the one in the [PR](https://github.com/maseev/netty/commit/e67ac28504952f989f7a926359a24fdeac03ad0d#diff-424f83af20faabeda9dd287cb533fbc8). As you can see, throughput in the last case is much higher.

This PR is probably related to the following issues: #1759, #5152

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/BatchedByteToMessageDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/BatchedByteToMessageDecoder.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2016 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.netty.handler.codec;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.channel.ChannelHandlerContext;
+
+import java.util.List;
+
+public abstract class BatchedByteToMessageDecoder extends ByteToMessageDecoder {
+
+    private boolean readInProgress;
+
+    protected BatchedByteToMessageDecoder() {
+        setDiscardAfterReads(1);
+    }
+
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+        readInProgress = true;
+        super.channelRead(ctx, msg);
+    }
+
+    @Override
+    protected final void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) {
+        if (!readInProgress) {
+            decodeBatched(ctx, in, out);
+        }
+    }
+
+    /**
+     * Decodes from one {@link ByteBuf} to another. This method will be called till either the input
+     * {@link ByteBuf} has nothing to read when return from this method or till nothing was read from the input
+     * {@link ByteBuf}. Compared to {@link ByteToMessageDecoder#decode(ChannelHandlerContext, ByteBuf, List)}, this
+     * method decodes messages by applying the batching strategy. Instead of having one {@link ByteBuf} which
+     * would always contain only one message, this method decodes a {@link ByteBuf} so that the resulting
+     * {@link ByteBuf} will contain N messages at once (where N can be greater or equal to 1).
+     *
+     * @param ctx the {@link ChannelHandlerContext} which this {@link BatchedByteToMessageDecoder} belongs to
+     * @param in the {@link ByteBuf} from which to read data
+     * @param out the {@link List} to which decoded messages should be added
+     */
+    protected abstract void decodeBatched(ChannelHandlerContext ctx, ByteBuf in, List<Object> out);
+
+    @Override
+    public final void channelReadComplete(ChannelHandlerContext ctx) throws Exception {
+        resetReadAndFlushIfNeeded(ctx);
+        super.channelReadComplete(ctx);
+    }
+
+    @Override
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
+        resetReadAndFlushIfNeeded(ctx);
+        super.exceptionCaught(ctx, cause);
+    }
+
+    private void resetReadAndFlushIfNeeded(ChannelHandlerContext ctx) {
+        readInProgress = false;
+        CodecOutputList out = CodecOutputList.newInstance();
+        try {
+            decodeBatched(ctx, internalBuffer(), out);
+        } finally {
+            afterDecode(ctx, out);
+        }
+    }
+}
--- a/codec/src/main/java/io/netty/handler/codec/BatchedFixedLengthFrameDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/BatchedFixedLengthFrameDecoder.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2016 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.netty.handler.codec;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.util.internal.ObjectUtil;
+
+import java.util.List;
+
+/**
+ * A decoder that mostly mimics {@link FixedLengthFrameDecoder} semantics. The only difference is this decoder batches
+ * incoming {@link ByteBuf} messages and then decodes them at once, passing a big {@link ByteBuf} which can possibly
+ * contain N messages (where N >= 1) through the pipeline at once.
+ */
+public class BatchedFixedLengthFrameDecoder extends BatchedByteToMessageDecoder {
+
+    final int frameLength;
+
+    /**
+     * Creates a new instance.
+     *
+     * @param frameLength the length of the frame
+     */
+    public BatchedFixedLengthFrameDecoder(int frameLength) {
+        this.frameLength = ObjectUtil.checkPositive(frameLength, "frameLength");
+    }
+
+    @Override
+    protected void decodeBatched(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) {
+        final int readableBytes = in.readableBytes();
+
+        if (readableBytes < frameLength) {
+            return;
+        }
+
+        final int numberOfMessages = readableBytes / frameLength;
+        final int bytesToCopy = numberOfMessages * frameLength;
+        final ByteBuf chunk = in.readRetainedSlice(bytesToCopy);
+
+        out.add(chunk);
+    }
+}
--- a/codec/src/main/java/io/netty/handler/codec/BatchedLengthFieldBasedFrameDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/BatchedLengthFieldBasedFrameDecoder.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2016 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.codec;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.util.internal.ObjectUtil;
+
+import java.util.List;
+
+/**
+ * A decoder that mostly mimics {@link LengthFieldBasedFrameDecoder} semantics. The only difference is this decoder
+ * batches incoming {@link ByteBuf} messages and then decodes them at once, passing a big {@link ByteBuf} which can
+ * possibly contain N messages (where N >= 1) through the pipeline at once.
+ */
+public class BatchedLengthFieldBasedFrameDecoder extends BatchedByteToMessageDecoder {
+
+    private final int maxFrameLength;
+    private final int lengthFieldOffset;
+    private final int lengthFieldLength;
+    private final int lengthFieldEndOffset;
+
+    /**
+     * Creates a new instance.
+     *
+     * @param maxFrameLength the maximum length of the frame.  If the length of the frame is greater than this value,
+     * {@link TooLongFrameException} will be thrown.
+     * @param lengthFieldOffset the offset of the length field
+     * @param lengthFieldLength the length of the length field
+     */
+    public BatchedLengthFieldBasedFrameDecoder(int maxFrameLength, int lengthFieldOffset, int lengthFieldLength) {
+        this.maxFrameLength = ObjectUtil.checkPositive(maxFrameLength, "maxFrameLength");
+        this.lengthFieldOffset = ObjectUtil.checkPositiveOrZero(lengthFieldOffset, "lengthFieldEndOffset");
+
+        if (lengthFieldOffset > maxFrameLength - lengthFieldLength) {
+            throw new IllegalArgumentException(
+                    "maxFrameLength (" + maxFrameLength + ") " +
+                    "must be equal to or greater than " +
+                    "lengthFieldOffset (" + lengthFieldOffset + ") + " +
+                    "lengthFieldLength (" + lengthFieldLength + ").");
+        }
+
+        this.lengthFieldLength = lengthFieldLength;
+        lengthFieldEndOffset = lengthFieldOffset + lengthFieldLength;
+    }
+
+    @Override
+    protected void decodeBatched(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) {
+        final int previousRdIdx = in.readerIndex();
+        int rdIdx = previousRdIdx;
+        int readableBytes = in.readableBytes();
+
+        while (readableBytes >= lengthFieldEndOffset) {
+            int actualLengthFieldOffset = rdIdx + lengthFieldOffset;
+            long frameLength =
+                    UnadjustedFrameLengthDecoderUtil.decode(in, actualLengthFieldOffset, lengthFieldLength);
+
+            if (frameLength < 0) {
+                throw new CorruptedFrameException(
+                        "negative pre-adjustment length field: " + frameLength);
+            }
+
+            frameLength += lengthFieldEndOffset;
+
+            if (frameLength > maxFrameLength) {
+                throw new TooLongFrameException(
+                        "Adjusted frame length exceeds " + maxFrameLength +
+                        ": " + frameLength + " - discarded");
+            }
+
+            int frameLengthInt = (int) frameLength;
+
+            if (readableBytes < frameLengthInt) {
+                break;
+            }
+
+            readableBytes -= frameLengthInt;
+            rdIdx += frameLengthInt;
+        }
+
+        if (rdIdx == previousRdIdx) {
+            return;
+        }
+
+        final int bytesToCopy = rdIdx - previousRdIdx;
+        final ByteBuf chunk = in.readRetainedSlice(bytesToCopy);
+
+        out.add(chunk);
+    }
+}
--- a/codec/src/main/java/io/netty/handler/codec/LengthFieldBasedFrameDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/LengthFieldBasedFrameDecoder.java
@@ -435,29 +435,7 @@ public class LengthFieldBasedFrameDecoder extends ByteToMessageDecoder {
      * @throws DecoderException if failed to decode the specified region
      */
     protected long getUnadjustedFrameLength(ByteBuf buf, int offset, int length, ByteOrder order) {
-        buf = buf.order(order);
-        long frameLength;
-        switch (length) {
-        case 1:
-            frameLength = buf.getUnsignedByte(offset);
-            break;
-        case 2:
-            frameLength = buf.getUnsignedShort(offset);
-            break;
-        case 3:
-            frameLength = buf.getUnsignedMedium(offset);
-            break;
-        case 4:
-            frameLength = buf.getUnsignedInt(offset);
-            break;
-        case 8:
-            frameLength = buf.getLong(offset);
-            break;
-        default:
-            throw new DecoderException(
-                    "unsupported lengthFieldLength: " + lengthFieldLength + " (expected: 1, 2, 3, 4, or 8)");
-        }
-        return frameLength;
+        return UnadjustedFrameLengthDecoderUtil.decode(buf.order(order), offset, length);
     }
 
     private void failIfNecessary(boolean firstDetectionOfTooLongFrame) {
--- a/codec/src/main/java/io/netty/handler/codec/UnadjustedFrameLengthDecoderUtil.java
+++ b/codec/src/main/java/io/netty/handler/codec/UnadjustedFrameLengthDecoderUtil.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2016 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.netty.handler.codec;
+
+import io.netty.buffer.ByteBuf;
+
+final class UnadjustedFrameLengthDecoderUtil {
+
+    private UnadjustedFrameLengthDecoderUtil() {
+    }
+
+    /**
+     * Decodes the specified region of the buffer into an unadjusted frame length. The default implementation is
+     * capable of decoding the specified region into an unsigned 8/16/24/32/64 bit integer. Override this method to
+     * decode the length field encoded differently. Note that this method must not modify the state of the specified
+     * buffer (e.g. {@code readerIndex}, {@code writerIndex}, and the content of the buffer.)
+     *
+     * @throws DecoderException if failed to decode the specified region
+     */
+    static long decode(ByteBuf buf, int offset, int length) {
+        switch (length) {
+        case 1:
+            return buf.getUnsignedByte(offset);
+        case 2:
+            return buf.getUnsignedShort(offset);
+        case 3:
+            return buf.getUnsignedMedium(offset);
+        case 4:
+            return buf.getUnsignedInt(offset);
+        case 8:
+            return buf.getLong(offset);
+        default:
+            throw new DecoderException(
+                    "unsupported lengthFieldLength: " + length + " (expected: 1, 2, 3, 4, or 8)");
+        }
+    }
+}
--- a/codec/src/test/java/io/netty/handler/codec/BatchedFixedLengthFrameDecoderTest.java
+++ b/codec/src/test/java/io/netty/handler/codec/BatchedFixedLengthFrameDecoderTest.java
@@ -1,0 +1,119 @@
+/*
+ * Copyright 2016 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.codec;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import io.netty.channel.embedded.EmbeddedChannel;
+import org.junit.Test;
+
+import static org.hamcrest.CoreMatchers.*;
+import static org.junit.Assert.*;
+
+public class BatchedFixedLengthFrameDecoderTest {
+
+    @Test(expected = IllegalArgumentException.class)
+    public void decoderWithZeroFrameLengthShouldThrowException() {
+        new BatchedFixedLengthFrameDecoder(0);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void decoderWithNegativeFrameLengthShouldThrowException() {
+        new BatchedFixedLengthFrameDecoder(-1);
+    }
+
+    @Test
+    public void writingNotAByteBufShouldPassThePipeline() {
+        final int frameLength = 10;
+        final EmbeddedChannel channel = new EmbeddedChannel(new BatchedFixedLengthFrameDecoder(frameLength));
+        final Object marker = new Object();
+
+        final boolean inboundBufferChanged = channel.writeInbound(marker);
+        final Object receivedObject = channel.readInbound();
+
+        assertThat(inboundBufferChanged, is(true));
+        assertThat(receivedObject, is(equalTo(marker)));
+
+        final boolean bufferReadable = channel.finishAndReleaseAll();
+
+        assertThat(bufferReadable, is(false));
+    }
+
+    @Test
+    public void writingNotEnoughBytesReadsNothing() {
+        final int frameLength = 10;
+        final EmbeddedChannel channel = new EmbeddedChannel(new BatchedFixedLengthFrameDecoder(frameLength));
+        final ByteBuf chunk = Unpooled.copiedBuffer(new byte[5]);
+
+        final boolean inboundBufferChanged = channel.writeInbound(chunk);
+        final Object receivedObject = channel.readInbound();
+
+        assertThat(inboundBufferChanged, is(false));
+        assertThat(receivedObject, is(equalTo(null)));
+
+        final boolean bufferReadable = channel.finishAndReleaseAll();
+
+        assertThat(bufferReadable, is(false));
+    }
+
+    @Test
+    public void writingOneAndAHalfMessageShouldRetrieveOnlyOneMessage() {
+        final int frameLength = 10;
+        final int oneAndAHalfMessageLength = frameLength + frameLength / 2;
+        final EmbeddedChannel channel =
+                new EmbeddedChannel(new BatchedFixedLengthFrameDecoder(frameLength));
+        final ByteBuf chunk = Unpooled.copiedBuffer(new byte[oneAndAHalfMessageLength]);
+
+        final boolean inboundBufferChanged = channel.writeInbound(chunk);
+        final ByteBuf receivedChunk = channel.readInbound();
+
+        assertThat(inboundBufferChanged, is(true));
+        assertThat(receivedChunk.readableBytes(), is(equalTo(frameLength)));
+        receivedChunk.release();
+
+        final boolean bufferReadable = channel.finishAndReleaseAll();
+
+        assertThat(bufferReadable, is(false));
+    }
+
+    @Test
+    public void leftBytesShouldAccumulate() {
+        final int frameLength = 10;
+        final EmbeddedChannel channel =
+                new EmbeddedChannel(new BatchedFixedLengthFrameDecoder(frameLength));
+
+        boolean inboundBufferChanged = channel.writeInbound(Unpooled.copiedBuffer(new byte[15]));
+        ByteBuf receivedChunk = channel.readInbound();
+
+        assertThat(inboundBufferChanged, is(true));
+        assertThat(receivedChunk.readableBytes(), is(equalTo(frameLength)));
+        receivedChunk.release();
+
+        inboundBufferChanged = channel.writeInbound(Unpooled.copiedBuffer(new byte[10]));
+        receivedChunk = channel.readInbound();
+
+        assertThat(inboundBufferChanged, is(true));
+        assertThat(receivedChunk.readableBytes(), is(equalTo(frameLength)));
+        receivedChunk.release();
+
+        receivedChunk = channel.readInbound();
+        assertThat(receivedChunk, is(equalTo(null)));
+
+        final boolean bufferReadable = channel.finishAndReleaseAll();
+
+        assertThat(bufferReadable, is(false));
+    }
+}
--- a/codec/src/test/java/io/netty/handler/codec/BatchedLengthFieldBasedFrameDecoderTest.java
+++ b/codec/src/test/java/io/netty/handler/codec/BatchedLengthFieldBasedFrameDecoderTest.java
@@ -1,0 +1,129 @@
+/*
+ * Copyright 2016 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.netty.handler.codec;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import io.netty.channel.embedded.EmbeddedChannel;
+import org.junit.Test;
+
+import static org.hamcrest.CoreMatchers.*;
+import static org.junit.Assert.*;
+
+public class BatchedLengthFieldBasedFrameDecoderTest {
+
+    @Test(expected = IllegalArgumentException.class)
+    public void decoderWithZeroMaxFrameLengthShouldThrowException() {
+        new BatchedLengthFieldBasedFrameDecoder(0, 1, 1);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void decoderWithNegativeMaxFrameLengthShouldThrowException() {
+        new BatchedLengthFieldBasedFrameDecoder(-1, 1, 1);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void decoderWithNegativeLengthFieldOffsetShouldThrowException() {
+        new BatchedLengthFieldBasedFrameDecoder(1, -1, 1);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void decoderWithTooBigLengthFieldOffsetShouldThrowException() {
+        new BatchedLengthFieldBasedFrameDecoder(10, 8, 8);
+    }
+
+    @Test
+    public void writingNotAByteBufShouldPassThePipeline() {
+        final EmbeddedChannel channel = new EmbeddedChannel(new BatchedLengthFieldBasedFrameDecoder(2, 0, 1));
+        final Object marker = new Object();
+
+        final boolean inboundBufferChanged = channel.writeInbound(marker);
+        final Object receivedObject = channel.readInbound();
+
+        assertThat(inboundBufferChanged, is(true));
+        assertThat(receivedObject, is(equalTo(marker)));
+
+        final boolean bufferReadable = channel.finishAndReleaseAll();
+
+        assertThat(bufferReadable, is(false));
+    }
+
+    @Test
+    public void writingNotEnoughBytesReadsNothing() {
+        final EmbeddedChannel channel = new EmbeddedChannel(new BatchedLengthFieldBasedFrameDecoder(6, 2, 1));
+        final ByteBuf chunk = Unpooled.copiedBuffer(new byte[] { 1, 1, 3, 0, 0 });
+
+        final boolean inboundBufferChanged = channel.writeInbound(chunk);
+        final Object receivedObject = channel.readInbound();
+
+        assertThat(inboundBufferChanged, is(false));
+        assertThat(receivedObject, is(equalTo(null)));
+
+        final boolean bufferReadable = channel.finishAndReleaseAll();
+
+        assertThat(bufferReadable, is(false));
+    }
+
+    @Test
+    public void writingOneAndAHalfMessageShouldRetrieveOnlyOneMessage() {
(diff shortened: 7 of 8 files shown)


codec/src/main/java/io/netty/handler/codec/UnadjustedFrameLengthDecoderUtil.java
```diff
@@ -0,0 +1,44 @@
+package io.netty.handler.codec;
+
+import io.netty.buffer.ByteBuf;
+
+public final class UnadjustedFrameLengthDecoderUtil {
```
johnou: Package protected?

codec/src/main/java/io/netty/handler/codec/BatchFixedLengthFrameDecoder.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    protected void decode(ChannelHandlerContext ctx, ByteBuf buffer, List<Object> out) {
+        final int readableBytes = buffer.readableBytes();
+
+        if (readableBytes < frameLength) {
+            return;
+        }
+
+        final int chunkLength = buffer.readableBytes() / frameLength * frameLength;
```
Scottmitch: also can you comment as to why this arithmetic (`/ frameLength * frameLength`) is necessary?

codec/src/main/java/io/netty/handler/codec/BatchFixedLengthFrameDecoder.java
```diff
@@ -0,0 +1,73 @@
...
+        final ByteBuf chunk = ctx.alloc().buffer(chunkLength, chunkLength).writeBytes(buffer, chunkLength);
```
Scottmitch: any reason why you aren't using `in.readRetainedSlice(frameLength)` like in `FixedLengthFrameDecoder`?

codec/src/main/java/io/netty/handler/codec/BatchLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,136 @@
...
+
+        if (rdIdx == 0) {
+            return;
+        }
+
+        final int previousWriterIndex = buffer.writerIndex();
+
+        buffer.writerIndex(rdIdx);
+
+        final ByteBuf chunk = ctx.alloc().buffer(rdIdx, rdIdx).writeBytes(buffer, rdIdx);
```
Scottmitch: `rdIdx` which is based upon `buffer.readerIndex` may not be 0 based so I don't think this should be used for `capacity`/`length`.

codec/src/main/java/io/netty/handler/codec/BatchLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,136 @@
...
+
+        buffer.writerIndex(previousWriterIndex);
+        buffer.discardReadBytes();
```
Scottmitch: the parent class will do this, correct?

codec/src/main/java/io/netty/handler/codec/BatchLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,136 @@
...
+            rdIdx += frameLengthInt;
+        }
+
+        if (rdIdx == 0) {
+            return;
+        }
+
+        final int previousWriterIndex = buffer.writerIndex();
+
+        buffer.writerIndex(rdIdx);
```
Scottmitch: will `readRetainedSlice` work here too?

codec/src/main/java/io/netty/handler/codec/UnadjustedFrameLengthDecoderUtil.java
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * Decodes the specified region of the buffer into an unadjusted frame length.  The default implementation is
+     * capable of decoding the specified region into an unsigned 8/16/24/32/64 bit integer.  Override this method to
+     * decode the length field encoded differently.  Note that this method must not modify the state of the specified
+     * buffer (e.g. {@code readerIndex}, {@code writerIndex}, and the content of the buffer.)
+     *
+     * @throws DecoderException if failed to decode the specified region
+     */
+    public static long decode(ByteBuf buf, int offset, int length) {
```
Scottmitch: make this method package private instead of public

codec/src/main/java/io/netty/handler/codec/BatchFixedLengthFrameDecoder.java
maseev: Let's suppose we have 1405 bytes in the `buffer`. If we divide the number of readable bytes by the frame length (100, for example), we'll get 14 as a result. This will be the number of messages in our `buffer` variable. 

Now we can multiply the result by the frame length again in order to get the number of bytes wee need to copy. By multiplying the result by 100 we'll get 1400 bytes. Now we can simply create a new `ByteBuff`, copy 1400 bytes into it and pass it further through the pipeline. 

This new `ByteBuff` will contain 14 messages. The remainder (5 bytes) can now be moved to the beginning of our `buffer` variable, so we can continue buffering incoming data.

codec/src/main/java/io/netty/handler/codec/BatchLengthFieldBasedFrameDecoder.java
maseev: Correct. I'll fix that

codec/src/main/java/io/netty/handler/codec/BatchLengthFieldBasedFrameDecoder.java
maseev: Why it may not be 0 based? [BatchFrameDecoder](https://github.com/netty/netty/pull/6052/files#diff-db29166cae2ff3be4fe271a0b1311ed1R88) sets the reader index to 0 after every `decode` method invocation.

codec/src/main/java/io/netty/handler/codec/BatchFixedLengthFrameDecoder.java
maseev: I just tested the `readRetainedSlice` method and this is actually not what I want. The whole point of my approach is to keep the initial buffer (the one from the BatchFrameDecoder) until the connection is closed. So, I can batch incoming data into this buffer (which is not likely to go bigger after the decoder performs a couple of decoding), then I can simply copy N bytes from it to the completely new buffer and pass it down through the pipeline. If I call the `readRetainedSlice` method, the new ByteBuff will share the memory region with the buffer in the `BatchFrameDecoder` class.

codec/src/main/java/io/netty/handler/codec/BatchLengthFieldBasedFrameDecoder.java
maseev: See my comment above.

codec/src/main/java/io/netty/handler/codec/BatchFixedLengthFrameDecoder.java
Scottmitch: You should be able to keep the underlying buffer and still avoid the copy operation you are doing here. This is what [ByteToMessageDecoder](https://github.com/netty/netty/blob/4.1/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java#L311) does. It is still not clear why these classes can't extend from `ByteToMessageDecoder` (maybe some package private adjustments are necessary as discussed in https://github.com/netty/netty/pull/6052#discussion_r88950398 and https://github.com/netty/netty/pull/6052#discussion_r88951172).

codec/src/main/java/io/netty/handler/codec/BatchFrameDecoder.java
```diff
@@ -0,0 +1,114 @@
...
+        releaseInternalBuffer();
+    }
+
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+        if (msg instanceof ByteBuf) {
+            final ByteBuf frame = (ByteBuf) msg;
+
+            try {
+                buffer.writeBytes(frame);
```
Scottmitch: `BatchingEchoServerHandler extends ChannelInboundHandlerAdapter` so it doesn't extend `ByteToMessageDecoder`. It also does not have a configurable accumulation strategy like `ByteToMessageDecoder ` so `BatchingEchoServerHandler` was not what I had in mind.

If you look at `ByteToMessageDecoder` you will notice it is doing very similar things to `BatchFrameDecoder`. The main difference is when it call's the user's `decode` method (in `channelRead` vs `channelReadComplete`). What I'm wondering is instead of having yet another aggregator implementation that also share the same method signatures for base classes (`decode(ChannelHandlerContext ctx, ByteBuf buffer, List<Object> out);`) if we can instead modify`ByteToMessageDecoder` such that it can optionally call `decode` in `channelReadComplete`.

codec/src/main/java/io/netty/handler/codec/BatchLengthFieldBasedFrameDecoder.java
Scottmitch: IIUC there is no guarantee the buffer will initially have `0` indexes, `BatchFrameDecoder` does call `discardReadBytes` until after this method is called, and I don't think is is necessary to make assumptions about the base class here (we should be able to use the interfaces generically with minimal overhead).

codec/src/main/java/io/netty/handler/codec/BatchFrameDecoder.java
maseev: Ok, I got it. Let me check this really quick.

codec/src/main/java/io/netty/handler/codec/BatchFixedLengthFrameDecoder.java
maseev: Oh, right. I'll add the comment.

codec/src/main/java/io/netty/handler/codec/BatchFrameDecoder.java
maseev: So, I looked at `ByteToMessageDecoder` more closely and I think it wouldn't be easy to modify it so that we could reuse some code and change its behavior (make the `decode` method invocation optional in `channelReadComplete`). Let's start with extracting the `Cumulator` classes and reusing it in the new batch frame decoders. I think it would be much easier. WDYT?

Scottmitch: I think we should strive to share/extend from `ByteToMessageDecoder` as it already also covers some boundary cases for use (channel input closed, auto read, etc..), and `ByteToMessageDecoder` is widely used so if more issues arise they will likely be fixed there.

One potential approach with minimal changes to `ByteToMessageDecoder` would be to have have a class that extends `ByteToMessageDecoder` (something like `BatchedByteToMessageDecoder`) like the following:

```java
// This is just to sketch the general ideas ... this is not necessarily production ready
public class BatchedByteToMessageDecoder extends ByteToMessageDecoder {
  private boolean readInProgress;

  @Override
  public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
    readInProgress = true;
    super.channelRead(ctx, msg);
  }

  // This method may be able to be final, but if the user is not using autoRead then they may want to
  // read more data manually here.
  @Override
  protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) {
    if (!readInProgress) {
      // Some exception situation occurred ... dispatch this read!
      decodeBatched(ctx, in, out);
    }
    // If we are in channel read then we don't want to call user code
    // until channelReadComplete so this is a NOOP
  }

  protected abstract void decodeBatched(ChannelHandlerContext ctx, ByteBuf in, List<Object> out);

  @Override
  public void channelReadComplete(ChannelHandlerContext ctx) throws Exception {
    readInProgress = false;
    CodecOutputList out = CodecOutputList.newInstance();
    try {
      decodeBatched(ctx, internalBuffer(), out);
    } finally {
       afterDecode(out);
    }
    super.channelReadComplete(ctx);
  }
}

// this Method would be added to `ByteToMessageDecoder` and is the extracted pieces in the
// finally block of `channelRead`.
  final void afterDecode(CodecOutputList out) {
                if (cumulation != null && !cumulation.isReadable()) {
                    numReads = 0;
                    cumulation.release();
                    cumulation = null;
                } else if (++ numReads >= discardAfterReads) {
                    // We did enough reads already try to discard some bytes so we not risk to see a OOME.
                    // See https://github.com/netty/netty/issues/4275
                    numReads = 0;
                    discardSomeReadBytes();
                }

                int size = out.size();
                decodeWasNull = !out.insertSinceRecycled();
                fireChannelRead(ctx, out, size);
                out.recycle();
  }
```

maseev: This approach doesn't seem to be working very well:

selection_002

And after two minutes or so:

```
io.netty.handler.codec.DecoderException: io.netty.util.internal.OutOfDirectMemoryError: failed to allocate 1853882368 byte(s) of direct memory (used: 1866465280, max: 3713531904)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:251)
	at io.netty.handler.codec.BatchedByteToMessageDecoder.channelRead(BatchedByteToMessageDecoder.java:31)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:372)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:358)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:350)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1334)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:372)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:358)
	at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:926)
	at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:129)
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:610)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:551)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:465)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:437)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:873)
	at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:144)
	at java.lang.Thread.run(Thread.java:745)
Caused by: io.netty.util.internal.OutOfDirectMemoryError: failed to allocate 1853882368 byte(s) of direct memory (used: 1866465280, max: 3713531904)
	at io.netty.util.internal.PlatformDependent.incrementMemoryCounter(PlatformDependent.java:624)
	at io.netty.util.internal.PlatformDependent.allocateDirectNoCleaner(PlatformDependent.java:578)
	at io.netty.buffer.PoolArena$DirectArena.allocateDirect(PoolArena.java:686)
	at io.netty.buffer.PoolArena$DirectArena.newUnpooledChunk(PoolArena.java:682)
	at io.netty.buffer.PoolArena.allocateHuge(PoolArena.java:246)
	at io.netty.buffer.PoolArena.allocate(PoolArena.java:224)
	at io.netty.buffer.PoolArena.reallocate(PoolArena.java:372)
	at io.netty.buffer.PooledByteBuf.capacity(PooledByteBuf.java:120)
	at io.netty.buffer.AbstractByteBuf.ensureWritable0(AbstractByteBuf.java:284)
	at io.netty.buffer.AbstractByteBuf.ensureWritable(AbstractByteBuf.java:265)
	at io.netty.buffer.AbstractByteBuf.writeBytes(AbstractByteBuf.java:1068)
	at io.netty.buffer.AbstractByteBuf.writeBytes(AbstractByteBuf.java:1060)
	at io.netty.buffer.AbstractByteBuf.writeBytes(AbstractByteBuf.java:1050)
	at io.netty.buffer.WrappedByteBuf.writeBytes(WrappedByteBuf.java:772)
	at io.netty.handler.codec.ByteToMessageDecoder$1.cumulate(ByteToMessageDecoder.java:91)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:245)
	... 16 more
```

Looks like it doesn't discard read bytes. Debugging... :disappointed:

Scottmitch: @maseev - Thanks for the update. The above code was just for demonstration purposes, was not tested, and doesn't surprise me that it is not complete. However until we determine this approach is not feasible I think this is way to go.

Scottmitch: @maseev - Oh something that may be helpful [FlushConsolidationHandler.java](https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/flush/FlushConsolidationHandler.java#L63) uses the same approach of tracking when we are in read (via `readInProgress`) this class should have very similar (if not the same) trigger points (`exceptionCaught`, `disconnect`, `close`, etc..).

maseev: @Scottmitch Thanks, I'll have a look. Quick question - how did you pick this [number](https://github.com/netty/netty/blob/4.1/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java#L137)?

Scottmitch: IIRC it was just a "reasonable guess" with the info we had at the time knowing that we could tweak it later if necessary. We added `setDiscardAfterReads` so this value can be configured, and we could add a package private constructor (possibly just `protected`) if you want to configure it at construction time.  @normanmaurer may recall more details too...

maseev: Aha, It turns out that the problem was in the high `discardAfterReads` value. After lowering it down to 1 I get the following results:

selection_001
selection_002

It's been almost 11 minutes and it's still working :smile:

Scottmitch: > It's been almost 11 minutes and it's still working 

Good stuff :)

> it never discarded any data because the buffer's refCnt was > 1.

This was the subject of another one of my comments (use of `readRetainedSlice`). Typically Netty handlers don't make copies of data because it is generally not expected to have the data persist for a long time, and if it does the user can always manually copy the data. In this scenario the underlying data should be released when your application releases all the derived buffers.

maseev: As far as I can tell, this approach is quite stable (at least BatchedFixedLengthFrameDecoder). Is there anything I can fix or improve in order to make this PR ready to be merged into upstream? :smile:

codec/src/main/java/io/netty/handler/codec/BatchedByteToMessageDecoder.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+        readInProgress = true;
+        super.channelRead(ctx, msg);
+    }
+
+    @Override
+    protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) {
```
normanmaurer: final

codec/src/main/java/io/netty/handler/codec/BatchedByteToMessageDecoder.java
```diff
@@ -0,0 +1,58 @@
...
+        if (!readInProgress) {
+            decodeBatched(ctx, in, out);
+        }
+    }
+
+    protected abstract void decodeBatched(ChannelHandlerContext ctx, ByteBuf in, List<Object> out);
```
normanmaurer: javadocs

codec/src/main/java/io/netty/handler/codec/BatchedByteToMessageDecoder.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Override
+    public void channelReadComplete(ChannelHandlerContext ctx) throws Exception {
```
normanmaurer: final ?

codec/src/main/java/io/netty/handler/codec/LengthFieldBasedFrameDecoder.java
```diff
@@ -32,9 +32,9 @@
  * that it can decode any message with a length field, which is often seen in
```
normanmaurer: revert changes to this file as its only reformatting ?

codec/src/main/java/io/netty/handler/codec/UnadjustedFrameLengthDecoderUtil.java
```diff
@@ -0,0 +1,60 @@
...
+
+import io.netty.buffer.ByteBuf;
+
+final class UnadjustedFrameLengthDecoderUtil {
+
+    private UnadjustedFrameLengthDecoderUtil() {
+    }
+
+    /**
+     * Decodes the specified region of the buffer into an unadjusted frame length.  The default implementation is
```
normanmaurer: remove double space after `.`.

codec/src/main/java/io/netty/handler/codec/UnadjustedFrameLengthDecoderUtil.java
```diff
@@ -0,0 +1,60 @@
...
+     * capable of decoding the specified region into an unsigned 8/16/24/32/64 bit integer.  Override this method to
```
normanmaurer: remove double space after `.`.

codec/src/main/java/io/netty/handler/codec/UnadjustedFrameLengthDecoderUtil.java
```diff
@@ -0,0 +1,60 @@
...
+     * decode the length field encoded differently.  Note that this method must not modify the state of the specified
```
normanmaurer: remove double space after `.`.

codec/src/test/java/io/netty/handler/codec/BatchedFixedLengthFrameDecoderTest.java
```diff
@@ -0,0 +1,101 @@
...
+        final int frameLength = 10;
+        final EmbeddedChannel channel = new EmbeddedChannel(new BatchedFixedLengthFrameDecoder(frameLength));
+        final Object marker = new Object();
+
+        channel.writeInbound(marker);
+        final Object receivedObject = channel.readInbound();
+
+        assertThat(receivedObject, is(equalTo(marker)));
+
+        channel.finish();
```
normanmaurer: assert return value

codec/src/test/java/io/netty/handler/codec/BatchedFixedLengthFrameDecoderTest.java
```diff
@@ -0,0 +1,101 @@
...
+    public void writingNotEnoughBytesReadsNothing() {
+        final int frameLength = 10;
+        final EmbeddedChannel channel = new EmbeddedChannel(new BatchedFixedLengthFrameDecoder(frameLength));
+        final ByteBuf chunk = Unpooled.copiedBuffer(new byte[5]);
+
+        channel.writeInbound(chunk);
+        final Object receivedObject = channel.readInbound();
+        assertThat(receivedObject, is(equalTo(null)));
+
+        channel.finish();
```
normanmaurer: assert return value

codec/src/test/java/io/netty/handler/codec/BatchedFixedLengthFrameDecoderTest.java
```diff
@@ -0,0 +1,101 @@
...
+        final EmbeddedChannel channel =
+                new EmbeddedChannel(new BatchedFixedLengthFrameDecoder(frameLength));
+        final ByteBuf chunk = Unpooled.copiedBuffer(new byte[oneAndAHalfMessageLength]);
+
+        channel.writeInbound(chunk);
+        final ByteBuf receivedChunk = channel.readInbound();
+        assertThat(receivedChunk.readableBytes(), is(equalTo(frameLength)));
+        receivedChunk.release();
+
+        channel.finish();
```
normanmaurer: assert return value

codec/src/test/java/io/netty/handler/codec/BatchedFixedLengthFrameDecoderTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Test
+    public void writingOneAndAHalfMessageShouldRetrieveOnlyOneMessage() {
+        final int frameLength = 10;
+        final int oneAndAHalfMessageLength = frameLength + frameLength / 2;
+        final EmbeddedChannel channel =
+                new EmbeddedChannel(new BatchedFixedLengthFrameDecoder(frameLength));
+        final ByteBuf chunk = Unpooled.copiedBuffer(new byte[oneAndAHalfMessageLength]);
+
+        channel.writeInbound(chunk);
```
normanmaurer: assert return value

codec/src/test/java/io/netty/handler/codec/BatchedFixedLengthFrameDecoderTest.java
```diff
@@ -0,0 +1,101 @@
...
+        channel.finish();
+    }
+
+    @Test
+    public void writingNotEnoughBytesReadsNothing() {
+        final int frameLength = 10;
+        final EmbeddedChannel channel = new EmbeddedChannel(new BatchedFixedLengthFrameDecoder(frameLength));
+        final ByteBuf chunk = Unpooled.copiedBuffer(new byte[5]);
+
+        channel.writeInbound(chunk);
```
normanmaurer: assert return value

codec/src/test/java/io/netty/handler/codec/BatchedFixedLengthFrameDecoderTest.java
```diff
@@ -0,0 +1,101 @@
...
+        new BatchedFixedLengthFrameDecoder(-1);
+    }
+
+    @Test
+    public void writingNotAByteBufShouldPassThePipeline() {
+        final int frameLength = 10;
+        final EmbeddedChannel channel = new EmbeddedChannel(new BatchedFixedLengthFrameDecoder(frameLength));
+        final Object marker = new Object();
+
+        channel.writeInbound(marker);
```
normanmaurer: assert return value

codec/src/test/java/io/netty/handler/codec/BatchedFixedLengthFrameDecoderTest.java
```diff
@@ -0,0 +1,101 @@
...
+        channel.finish();
+    }
+
+    @Test
+    public void leftBytesShouldAccumulate() {
+        final int frameLength = 10;
+        final EmbeddedChannel channel =
+                new EmbeddedChannel(new BatchedFixedLengthFrameDecoder(frameLength));
+
+        channel.writeInbound(Unpooled.copiedBuffer(new byte[15]));
```
normanmaurer: assert return value

codec/src/test/java/io/netty/handler/codec/BatchedFixedLengthFrameDecoderTest.java
```diff
@@ -0,0 +1,101 @@
...
+        ByteBuf receivedChunk = channel.readInbound();
+        assertThat(receivedChunk.readableBytes(), is(equalTo(frameLength)));
+        receivedChunk.release();
+
+        channel.writeInbound(Unpooled.copiedBuffer(new byte[10]));
```
normanmaurer: assert return value

codec/src/test/java/io/netty/handler/codec/BatchedFixedLengthFrameDecoderTest.java
```diff
@@ -0,0 +1,101 @@
...
+        receivedChunk = channel.readInbound();
+        assertThat(receivedChunk.readableBytes(), is(equalTo(frameLength)));
+        receivedChunk.release();
+
+        receivedChunk = channel.readInbound();
+        assertThat(receivedChunk, is(equalTo(null)));
+        channel.finish();
```
normanmaurer: assert return value

codec/src/test/java/io/netty/handler/codec/BatchedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,112 @@
...
+        assertThat(receivedChunk.readableBytes(), is(equalTo(completeFrame.length)));
+        receivedChunk.release();
+
+        chunk = Unpooled.copiedBuffer(anotherHalfFrame);
+        channel.writeInbound(chunk);
+        receivedChunk = channel.readInbound();
+        assertThat(receivedChunk.readableBytes(), is(equalTo(halfFrame.length + anotherHalfFrame.length)));
+        receivedChunk.release();
+
+        channel.finish();
```
normanmaurer: assert return value

codec/src/test/java/io/netty/handler/codec/BatchedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,112 @@
...
+        final byte[] anotherHalfFrame = { 0, 0, 0 };
+        ByteBuf chunk = Unpooled.copiedBuffer(completeFrame, halfFrame);
+
+        channel.writeInbound(chunk);
+        ByteBuf receivedChunk = channel.readInbound();
+        assertThat(receivedChunk.readableBytes(), is(equalTo(completeFrame.length)));
+        receivedChunk.release();
+
+        chunk = Unpooled.copiedBuffer(anotherHalfFrame);
+        channel.writeInbound(chunk);
```
normanmaurer: assert return value

codec/src/test/java/io/netty/handler/codec/BatchedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,112 @@
...
+    @Test
+    public void leftBytesShouldAccumulate() {
+        final EmbeddedChannel channel =
+                new EmbeddedChannel(new BatchedLengthFieldBasedFrameDecoder(10, 2, 1));
+        final byte[] completeFrame = { 1, 1, 3, 0, 0, 0 };
+        final byte[] halfFrame = { 1, 1, 4, 0 };
+        final byte[] anotherHalfFrame = { 0, 0, 0 };
+        ByteBuf chunk = Unpooled.copiedBuffer(completeFrame, halfFrame);
+
+        channel.writeInbound(chunk);
```
normanmaurer: assert return value

codec/src/test/java/io/netty/handler/codec/BatchedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,112 @@
...
+        final byte[] completeFrame = { 1, 1, 3, 0, 0, 0 };
+        final byte[] halfFrame = { 1, 1, 2, 0 };
+        final ByteBuf chunk = Unpooled.copiedBuffer(completeFrame, halfFrame);
+
+        channel.writeInbound(chunk);
+        final ByteBuf receivedChunk = channel.readInbound();
+        assertThat(receivedChunk.readableBytes(), is(equalTo(completeFrame.length)));
+        receivedChunk.release();
+
+        channel.finish();
```
normanmaurer: assert return value

codec/src/test/java/io/netty/handler/codec/BatchedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+    @Test
+    public void writingOneAndAHalfMessageShouldRetrieveOnlyOneMessage() {
+        final EmbeddedChannel channel =
+                new EmbeddedChannel(new BatchedLengthFieldBasedFrameDecoder(10, 2, 1));
+        final byte[] completeFrame = { 1, 1, 3, 0, 0, 0 };
+        final byte[] halfFrame = { 1, 1, 2, 0 };
+        final ByteBuf chunk = Unpooled.copiedBuffer(completeFrame, halfFrame);
+
+        channel.writeInbound(chunk);
```
normanmaurer: assert return value

codec/src/test/java/io/netty/handler/codec/BatchedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,112 @@
...
+    @Test
+    public void writingNotEnoughBytesReadsNothing() {
+        final EmbeddedChannel channel = new EmbeddedChannel(new BatchedLengthFieldBasedFrameDecoder(6, 2, 1));
+        final ByteBuf chunk = Unpooled.copiedBuffer(new byte[] { 1, 1, 3, 0, 0 });
+
+        channel.writeInbound(chunk);
+        final Object receivedObject = channel.readInbound();
+        assertThat(receivedObject, is(equalTo(null)));
+
+        channel.finish();
```
normanmaurer: assert return value

codec/src/test/java/io/netty/handler/codec/BatchedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    @Test
+    public void writingNotEnoughBytesReadsNothing() {
+        final EmbeddedChannel channel = new EmbeddedChannel(new BatchedLengthFieldBasedFrameDecoder(6, 2, 1));
+        final ByteBuf chunk = Unpooled.copiedBuffer(new byte[] { 1, 1, 3, 0, 0 });
+
+        channel.writeInbound(chunk);
```
normanmaurer: assert return value

codec/src/test/java/io/netty/handler/codec/BatchedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,112 @@
...
+    public void writingNotAByteBufShouldPassThePipeline() {
+        final EmbeddedChannel channel = new EmbeddedChannel(new BatchedLengthFieldBasedFrameDecoder(2, 0, 1));
+        final Object marker = new Object();
+
+        channel.writeInbound(marker);
+        final Object receivedObject = channel.readInbound();
+
+        assertThat(receivedObject, is(equalTo(marker)));
+
+        channel.finish();
```
normanmaurer: assert return value

codec/src/test/java/io/netty/handler/codec/BatchedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,112 @@
...
+    public void decoderWithTooBigLengthFieldOffsetShouldThrowException() {
+        new BatchedLengthFieldBasedFrameDecoder(10, 8, 8);
+    }
+
+    @Test
+    public void writingNotAByteBufShouldPassThePipeline() {
+        final EmbeddedChannel channel = new EmbeddedChannel(new BatchedLengthFieldBasedFrameDecoder(2, 0, 1));
+        final Object marker = new Object();
+
+        channel.writeInbound(marker);
```
normanmaurer: assert return value

maseev:
I wonder why [this](https://github.com/netty/netty/blob/4.1/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java#L473) method is not part of the [Cumulator](https://github.com/netty/netty/blob/4.1/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java#L484)? Of course, you have to convert `Cumulator` to an abstract class. Was it done intentionally? I think it would be much cleaner if `Cumulator` knew how to 'expand' itself. WDYT?

maseev:
PTAL

normanmaurer:
@maseev the problem is we can not change Cumulator now as it would be a breaking change in terms of API

maseev:
@normanmaurer Fair enough :smile:

maseev:
@normanmaurer, @Scottmitch Let me know if there's anything left I need to fix :smile: :desktop_computer:

codec/src/main/java/io/netty/handler/codec/BatchedLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,118 @@
...
+
+        if (rdIdx == previousRdIdx) {
+            return;
+        }
+
+        final int previousWriterIndex = in.writerIndex();
+
+        in.writerIndex(rdIdx);
+
+        final ByteBuf chunk = ctx.alloc().buffer(rdIdx, rdIdx).writeBytes(in, rdIdx);
```
maseev: Forgot to replace this allocation with using readRetainedSlice. Will fix this shortly.

Scottmitch:
build failure is attributed to https://github.com/netty/netty/issues/6057 which has been fixed. If you rebase this issue should go away.

codec/src/main/java/io/netty/handler/codec/BatchedFixedLengthFrameDecoder.java
```diff
@@ -0,0 +1,60 @@
...
+
+    final int frameLength;
+
+    /**
+     * Creates a new instance.
+     *
+     * @param frameLength the length of the frame
+     */
+    public BatchedFixedLengthFrameDecoder(int frameLength) {
+        if (frameLength <= 0) {
```
Scottmitch:
```java
this.frameLength = ObjectUtil.checkPositive(frameLength, "frameLength");
```

codec/src/main/java/io/netty/handler/codec/BatchedLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * Creates a new instance.
+     *
+     * @param maxFrameLength the maximum length of the frame.  If the length of the frame is greater than this value,
+     * {@link TooLongFrameException} will be thrown.
+     * @param lengthFieldOffset the offset of the length field
+     * @param lengthFieldLength the length of the length field
+     */
+    public BatchedLengthFieldBasedFrameDecoder(int maxFrameLength, int lengthFieldOffset, int lengthFieldLength) {
+        if (maxFrameLength <= 0) {
```
Scottmitch: use `ObjectUtil.checkPositive` for these.

codec/src/main/java/io/netty/handler/codec/UnadjustedFrameLengthDecoderUtil.java
```diff
@@ -0,0 +1,60 @@
...
+     * buffer (e.g. {@code readerIndex}, {@code writerIndex}, and the content of the buffer.)
+     *
+     * @throws DecoderException if failed to decode the specified region
+     */
+    static long decode(ByteBuf buf, int offset, int length) {
+        final long frameLength;
+
+        switch (length) {
+        case 1:
+            frameLength = buf.getUnsignedByte(offset);
```
Scottmitch: consider removing `frameLength` and just directly `return`.

codec/src/main/java/io/netty/handler/codec/BatchedByteToMessageDecoder.java
```diff
@@ -0,0 +1,70 @@
...
+     *
+     * @param ctx the {@link ChannelHandlerContext} which this {@link BatchedByteToMessageDecoder} belongs to
+     * @param in the {@link ByteBuf} from which to read data
+     * @param out the {@link List} to which decoded messages should be added
+     */
+    protected abstract void decodeBatched(ChannelHandlerContext ctx, ByteBuf in, List<Object> out);
+
+    @Override
+    public final void channelReadComplete(ChannelHandlerContext ctx) throws Exception {
+        readInProgress = false;
```
Scottmitch: please use the same approach as [FlushConsolidationHandler#resetReadAndFlushIfNeeded](https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/flush/FlushConsolidationHandler.java#L153) and make sure the same trigger points are accounted for.

codec/src/test/java/io/netty/handler/codec/BatchedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,129 @@
...
+        final EmbeddedChannel channel = new EmbeddedChannel(new BatchedLengthFieldBasedFrameDecoder(2, 0, 1));
+        final Object marker = new Object();
+
+        final boolean inboundBufferChanged = channel.writeInbound(marker);
+        final Object receivedObject = channel.readInbound();
+
+        assertThat(inboundBufferChanged, is(true));
+        assertThat(receivedObject, is(equalTo(marker)));
+
+        final boolean bufferReadable = channel.finish();
```
Scottmitch: `channel.finish` -> `channel.finishAndReleaseAll()`

codec/src/test/java/io/netty/handler/codec/BatchedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,129 @@
...
+        final EmbeddedChannel channel = new EmbeddedChannel(new BatchedLengthFieldBasedFrameDecoder(6, 2, 1));
+        final ByteBuf chunk = Unpooled.copiedBuffer(new byte[] { 1, 1, 3, 0, 0 });
+
+        final boolean inboundBufferChanged = channel.writeInbound(chunk);
+        final Object receivedObject = channel.readInbound();
+
+        assertThat(inboundBufferChanged, is(false));
+        assertThat(receivedObject, is(equalTo(null)));
+
+        final boolean bufferReadable = channel.finish();
```
Scottmitch: `channel.finish` -> `channel.finishAndReleaseAll()`

codec/src/test/java/io/netty/handler/codec/BatchedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,129 @@
...
+        final ByteBuf chunk = Unpooled.copiedBuffer(completeFrame, halfFrame);
+
+        final boolean inboundBufferChanged = channel.writeInbound(chunk);
+        final ByteBuf receivedChunk = channel.readInbound();
+
+        assertThat(inboundBufferChanged, is(true));
+        assertThat(receivedChunk.readableBytes(), is(equalTo(completeFrame.length)));
+        receivedChunk.release();
+
+        final boolean bufferReadable = channel.finish();
```
Scottmitch: `channel.finish` -> `channel.finishAndReleaseAll()`

codec/src/test/java/io/netty/handler/codec/BatchedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+        chunk = Unpooled.copiedBuffer(anotherHalfFrame);
+        inboundBufferChanged = channel.writeInbound(chunk);
+        receivedChunk = channel.readInbound();
+
+        assertThat(inboundBufferChanged, is(true));
+        assertThat(receivedChunk.readableBytes(), is(equalTo(halfFrame.length + anotherHalfFrame.length)));
+        receivedChunk.release();
+
+        final boolean bufferReadable = channel.finish();
```
Scottmitch: `channel.finish` -> `channel.finishAndReleaseAll()`

codec/src/test/java/io/netty/handler/codec/BatchedFixedLengthFrameDecoderTest.java
```diff
@@ -0,0 +1,119 @@
...
+        receivedChunk = channel.readInbound();
+
+        assertThat(inboundBufferChanged, is(true));
+        assertThat(receivedChunk.readableBytes(), is(equalTo(frameLength)));
+        receivedChunk.release();
+
+        receivedChunk = channel.readInbound();
+        assertThat(receivedChunk, is(equalTo(null)));
+
+        final boolean bufferReadable = channel.finish();
```
Scottmitch: `channel.finish` -> `channel.finishAndReleaseAll()`

codec/src/test/java/io/netty/handler/codec/BatchedFixedLengthFrameDecoderTest.java
```diff
@@ -0,0 +1,119 @@
...
+        final ByteBuf chunk = Unpooled.copiedBuffer(new byte[oneAndAHalfMessageLength]);
+
+        final boolean inboundBufferChanged = channel.writeInbound(chunk);
+        final ByteBuf receivedChunk = channel.readInbound();
+
+        assertThat(inboundBufferChanged, is(true));
+        assertThat(receivedChunk.readableBytes(), is(equalTo(frameLength)));
+        receivedChunk.release();
+
+        final boolean bufferReadable = channel.finish();
```
Scottmitch: `channel.finish` -> `channel.finishAndReleaseAll()`

codec/src/test/java/io/netty/handler/codec/BatchedFixedLengthFrameDecoderTest.java
```diff
@@ -0,0 +1,119 @@
...
+        final EmbeddedChannel channel = new EmbeddedChannel(new BatchedFixedLengthFrameDecoder(frameLength));
+        final ByteBuf chunk = Unpooled.copiedBuffer(new byte[5]);
+
+        final boolean inboundBufferChanged = channel.writeInbound(chunk);
+        final Object receivedObject = channel.readInbound();
+
+        assertThat(inboundBufferChanged, is(false));
+        assertThat(receivedObject, is(equalTo(null)));
+
+        final boolean bufferReadable = channel.finish();
```
Scottmitch: `channel.finish` -> `channel.finishAndReleaseAll()`

maseev:
@Scottmitch PTAL.

codec/src/main/java/io/netty/handler/codec/BatchedByteToMessageDecoder.java
```diff
@@ -0,0 +1,80 @@
...
+    @Override
+    public final void channelReadComplete(ChannelHandlerContext ctx) throws Exception {
+        resetReadAndFlushIfNeeded(ctx);
+        super.channelReadComplete(ctx);
+    }
+
+    @Override
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
+        resetReadAndFlushIfNeeded(ctx);
+        super.exceptionCaught(ctx, cause);
```
Scottmitch: `super.channelReadComplete` will take care of calling read if `autoRead` is not enabled, and also discarding some bytes, but `super.exceptionCaught` will not ... I wonder if we should try to account for the same thing here since we may be decoding. @normanmaurer - WDYT

maseev:
Is there anything else I can do to pass the code review and get this PR merged?

maseev:
This [PR](https://github.com/netty/netty/pull/5716) addresses the similar problem and implements the batching strategy in a more idiomatic way.

normanmaurer:
@maseev sorry but I am currently super busy and so had not much time thinking about this. I will do once I have some cycles tho

maseev:
@normanmaurer I'm not sure this is a good idea after all. The PR I mentioned above solves the very similar problem in a more idiomatic way. I can restore the branch and reopen this PR if you still think it's worth reviewing.

maseev:
Closing in favor of #5716

PR #6053, opened 2016-11-21
Now that LocalChannel#releaseInboundBuffers is only called from the E…

Scottmitch:
…ventLoop (eb4d317b9d64f3945a209804fec4c3fe695f4f9f) it should clear readInProgress and drain/release the queue. Otherwise if a read event is pending (doBeginRead) was called we may later call channelRead or channelReadComplete after we have closed the channel.

Modifications:
LocalChannel#releaseInboundBuffers should always clear/release the queue and set readInProgress to false

Result:
LocalChannel queue is more reliably cleaned up.

### DIFF
--- a/transport/src/main/java/io/netty/channel/local/LocalChannel.java
+++ b/transport/src/main/java/io/netty/channel/local/LocalChannel.java
@@ -422,9 +422,8 @@ public class LocalChannel extends AbstractChannel {
     }
 
     private void releaseInboundBuffers() {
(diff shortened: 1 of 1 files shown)


Scottmitch:
@normanmaurer - PTAL

normanmaurer:
@Scottmitch doh! Ship it buddy 🍺

Scottmitch:
4.1 (a25101d) 4.0 (52a572a)

PR #6055, opened 2016-11-22
Define a inner class to avoid classloading when fd is full

Hanson1129:
When "Too many open files" happens,the great URLClassLoader cannot do any classloading because URLClassLoader need a FD to findClass.
So the anonymous inner class , new Runnable(){} will cause a problem.
The stack is like this:

``` java 

16:17:51.483 [HSF-HTTP-Boss-15-thread-2] DEBUG i.n.c.AbstractChannelHandlerContext - An exception java.lang.IllegalAccessError: tried to access class io.netty.bootstrap.ServerBootstrap$ServerBootstrapAcceptor$2 from class io.netty.bootstrap.ServerBootstrap$ServerBootstrapAcceptor
        at io.netty.bootstrap.ServerBootstrap$ServerBootstrapAcceptor.exceptionCaught(ServerBootstrap.java:279)
        at io.netty.channel.AbstractChannelHandlerContext.invokeExceptionCaught(AbstractChannelHandlerContext.java:296)
        at io.netty.channel.AbstractChannelHandlerContext.invokeExceptionCaught(AbstractChannelHandlerContext.java:275)
        at io.netty.channel.AbstractChannelHandlerContext.fireExceptionCaught(AbstractChannelHandlerContext.java:267)
        at io.netty.channel.DefaultChannelPipeline$HeadContext.exceptionCaught(DefaultChannelPipeline.java:1301)
        at io.netty.channel.AbstractChannelHandlerContext.invokeExceptionCaught(AbstractChannelHandlerContext.java:296)
        at io.netty.channel.AbstractChannelHandlerContext.invokeExceptionCaught(AbstractChannelHandlerContext.java:275)
        at io.netty.channel.DefaultChannelPipeline.fireExceptionCaught(DefaultChannelPipeline.java:914)
        at io.netty.channel.nio.AbstractNioMessageChannel$NioMessageUnsafe.read(AbstractNioMessageChannel.java:102)
        at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:651)
        at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:574)
        at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:488)
        at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:450)
        at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:873)
[...]

Issue #1328, opened 2013-05-02
Too many open files causes tight loop, no recovery


dewjose:
Netty 3.5.\* handled this case more gracefully. Once the open file handle count retreated to below the limit the netty pipeline recovered and continued to operate.

With netty 4.0 the pipeline crashes, I can't catch this exception, or don't know how, and the JVM spins out of contro.

05/02/2013 15:12:20  WARN nioEventLoopGroup-10-2 io.netty.channel.DefaultChannelPipeline:151 - An exceptionCaught() event was fired, and it reached at the end of the pipeline. It usually means the last h\
andler in the pipeline did not handle the exception.
java.io.IOException: Too many open files
        at sun.nio.ch.ServerSocketChannelImpl.accept0(Native Method)
        at sun.nio.ch.ServerSocketChannelImpl.accept(ServerSocketChannelImpl.java:226)
        at io.netty.channel.socket.nio.NioServerSocketChannel.doReadMessages(NioServerSocketChannel.java:108)
        at io.netty.channel.nio.AbstractNioMessageChannel$NioMessageUnsafe.read(AbstractNioMessageChannel.java:61)
        at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:401)
        at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:365)
        at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:302)
        at io.netty.channel.SingleThreadEventExecutor$2.run(SingleThreadEventExecutor.java:110)
        at java.lang.Thread.run(Thread.java:722)

Hanson1129:
It seems not work with my test...
When "Too many open files" happens,the great URLClassLoader cannot do any classloading because URLClassLoader need a FD to findClass.
So the anonymous inner class , new Runnable(){} will cause a problem.
The stack is like this:

``` java 

16:17:51.483 [HSF-HTTP-Boss-15-thread-2] DEBUG i.n.c.AbstractChannelHandlerContext - An exception java.lang.IllegalAccessError: tried to access class io.netty.bootstrap.ServerBootstrap$ServerBootstrapAcceptor$2 from class io.netty.bootstrap.ServerBootstrap$ServerBootstrapAcceptor
        at io.netty.bootstrap.ServerBootstrap$ServerBootstrapAcceptor.exceptionCaught(ServerBootstrap.java:279)
[...]

Issue #5962, opened 2016-10-31
Don't log the exception object while FD is full


Hanson1129:
check this stack

``` java
Exception in thread "HSF-Boss-15-thread-1" java.lang.NoClassDefFoundError: ch/qos/logback/core/pattern/SpacePadder
        at ch.qos.logback.core.pattern.FormattingConverter.write(FormattingConverter.java:66)
        at ch.qos.logback.core.pattern.PatternLayoutBase.writeLoopOnConverters(PatternLayoutBase.java:119)
        at ch.qos.logback.classic.PatternLayout.doLayout(PatternLayout.java:168)
        at ch.qos.logback.classic.PatternLayout.doLayout(PatternLayout.java:59)
        at ch.qos.logback.core.encoder.LayoutWrappingEncoder.doEncode(LayoutWrappingEncoder.java:134)
        at ch.qos.logback.core.OutputStreamAppender.writeOut(OutputStreamAppender.java:188)
        at ch.qos.logback.core.OutputStreamAppender.subAppend(OutputStreamAppender.java:212)
        at ch.qos.logback.core.OutputStreamAppender.append(OutputStreamAppender.java:103)
        at ch.qos.logback.core.UnsynchronizedAppenderBase.doAppend(UnsynchronizedAppenderBase.java:88)
        at ch.qos.logback.core.spi.AppenderAttachableImpl.appendLoopOnAppenders(AppenderAttachableImpl.java:48)
        at ch.qos.logback.classic.Logger.appendLoopOnAppenders(Logger.java:272)
        at ch.qos.logback.classic.Logger.callAppenders(Logger.java:259)
        at ch.qos.logback.classic.Logger.buildLoggingEventAndAppend(Logger.java:441)
        at ch.qos.logback.classic.Logger.filterAndLog_0_Or3Plus(Logger.java:395)
        at ch.qos.logback.classic.Logger.warn(Logger.java:712)
        at io.netty.util.internal.logging.Slf4JLogger.warn(Slf4JLogger.java:151)
        at io.netty.util.concurrent.SingleThreadEventExecutor$2.run(SingleThreadEventExecutor.java:115)

```

while vm 's fd is full, if you try to log the object of throwable,the JVM may try to do some class loading.Because of fxxking "too many open files", my JVM she cannot load any class then.

see this:

![image](https://cloud.githubusercontent.com/assets/4310631/19845965/d3ddc5bc-9f76-11e6-9d24-2653fa6e0465.png)

[...]

PR #6056, opened 2016-11-22
[#6055] [#1328] Ensure trying to recover from exceptionCaught on the …

normanmaurer:
…ServerChannel works as expected

Motivation:

When "Too many open files" happens,the URLClassLoader cannot do any classloading because URLClassLoader need a FD  for findClass. Because of this the anonymous inner class that is created to re-enable auto read may cause a problem.

Modification:

Pre-create Runnable that is scheduled and so ensure it is not lazy loaded.

Result:

No more problems when try to recover.

### DIFF
--- a/transport/src/main/java/io/netty/bootstrap/ServerBootstrap.java
+++ b/transport/src/main/java/io/netty/bootstrap/ServerBootstrap.java
@@ -168,7 +168,7 @@ public class ServerBootstrap extends AbstractBootstrap<ServerBootstrap, ServerCh
 
         p.addLast(new ChannelInitializer<Channel>() {
             @Override
-            public void initChannel(Channel ch) throws Exception {
+            public void initChannel(final Channel ch) throws Exception {
                 final ChannelPipeline pipeline = ch.pipeline();
                 ChannelHandler handler = config.handler();
                 if (handler != null) {
@@ -183,7 +183,7 @@ public class ServerBootstrap extends AbstractBootstrap<ServerBootstrap, ServerCh
                     @Override
                     public void run() {
                         pipeline.addLast(new ServerBootstrapAcceptor(
-                                currentChildGroup, currentChildHandler, currentChildOptions, currentChildAttrs));
+                                ch, currentChildGroup, currentChildHandler, currentChildOptions, currentChildAttrs));
                     }
                 });
             }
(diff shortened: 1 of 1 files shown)


normanmaurer:
@Hanson1129 PTAL... I did the changes that I suggested and squashed it into your commit. So you are still listed as author (also altered the commit message slightly).

Please let me know if this works.

Hanson1129:
aha , cool!
Let me test it in our presure test system immediately~

normanmaurer:
@Hanson1129 Thanks a lot! Also would be cool if you could add yourself to the adopter page:

https://github.com/netty/netty/wiki/Adopters

Only if possible of course ;)LE

Hanson1129:
@normanmaurer  

It works !! 

We are able to stable reproduction in our pressure test environment , and I rewrite this code in 4.0.35.Final , put it in the same environment, then the boss-thread is peace and quiet , this problem never happens.

We will upgrade to 4.1 after 4.1.7 release.

Buy our framework in live environment isn't open source in Github.What I write is not Dubbo.

But Dubbo use netty as well,I can put it into adopter page.

normanmaurer:
@Hanson1129 the adopters page [1] list companies (with close source stuff and OSS stuff) and we also have a projects page [2] which list OSS things:

[1] http://netty.io/wiki/adopters.html
[2] http://netty.io/wiki/related-projects.html

Also thanks for verify, this will be merged to 4.1 and 4.0.

normanmaurer:
@Scottmitch PTAL as well before I merge.

transport/src/main/java/io/netty/bootstrap/ServerBootstrap.java
```diff
@@ -219,14 +219,27 @@ public ServerBootstrap validate() {
...
+                final Channel channel, EventLoopGroup childGroup, ChannelHandler childHandler,
                 Entry<ChannelOption<?>, Object>[] childOptions, Entry<AttributeKey<?>, Object>[] childAttrs) {
             this.childGroup = childGroup;
             this.childHandler = childHandler;
             this.childOptions = childOptions;
             this.childAttrs = childAttrs;
+
+            // Task which is scheduled to re-enable auto-read.
+            // It's important to create this Runnable before we try to submit it as otherwise the URLClassLoader may
+            // not be able to load the class because of the file limit it already reached.
```
Scottmitch: Isn't there a potentially for this issue to occur in general especially for lazily loaded classes? I wonder if this will not just pop up somewhere else (e.g. if one of these methods called in this class try to load create another class like `ctx.channel().config()` or something).

Hanson1129: Yap.See [another issue](https://github.com/netty/netty/issues/5962).

What I have done is is..

``` java 

private void initNettyLogger() {
        if(bossGroup != null)
            bossGroup.execute(new Runnable() {
                @Override
                public void run() {
                    throw new RuntimeException("Please ignore this.HSF do this just for init netty logger.");
                }
            });
        if (httpBossGroup != null)
            httpBossGroup.execute(new Runnable() {
                @Override
                public void run() {
                    throw new RuntimeException("Please ignore this.We do this just to init Netty's logger.");
                }
            });
    }

```

normanmaurer: @Scottmitch sure... that said we should try to at least guard against it here as we try to recover from it here. So I think this can't harm. Ok to merge?

Scottmitch: @normanmaurer - agreed this won't hurt anything and I'm not against merging but I'm not convinced this addresses the original issue. I wonder if the general issue (no more FDs when attempting to load a class) is better addressed by application configuration updates (limit FDs accepted and/or connections made so the JVM has enough FDs to use for book keeping).

normanmaurer:
For sure a higher ulimit or backpressure is the way to go... that said let me merge this one

normanmaurer:
@Scottmitch can I pull in ?

Scottmitch:
@normanmaurer sure .. that said @Hanson1129 can you report back about https://github.com/netty/netty/pull/6056#discussion_r89387209 ? I fear we may be "kicking the can down the road" in your environment.

normanmaurer:
ping @Hanson1129

uglycow:
@normanmaurer   @Scottmitch   i guess this #6162 is what we need to solve this problem.  the key is to limit the connections accepted in.

Hanson1129:
We still want this feature.We will use backpressure strategy by setting high-water-mark as well.

normanmaurer:
Ok let me pull this pr in

normanmaurer:
Cherry-picked into 4.1 (d0a3877535814153306f8f934e2120204cd22fd7) and 4.0 (fb61cc8)

Viyond:
just mark.

Issue #6037, opened 2016-11-18
Make FixedChannelPool non final


derbylock:
We use **FixedChannelPool**  in our project. But we have some problems adopting it to our needs. For example, we want to override methods **_pollChannel_** and **_offerChannel_** for monitoring purposes. We want to monitor count of channel objects in the pool. But it is difficult because **FixedChannelPool** is final. Could you consider removing final keyword from it, please?

normanmaurer:
Sounds ok to me... @Scottmitch ?

Scottmitch:
sgtm

normanmaurer:
@derbylock could you provide a PR ?

normanmaurer:
Fixed by https://github.com/netty/netty/pull/6059

PR #6059, opened 2016-11-23
Removing final keyword from FixedChannelPool

derbylock:
See #6037 additionally


normanmaurer:
Cherry-picked into 4.1 (f6b37a38bac6dbea5dd2282c68a0f41c456f67a2) and 4.0 (1b528de)

PR #6060, opened 2016-11-23
Return status 431 on header size errors

nhnFreespirit:
Motivation:

Currently clients attempting to send headers that are too large recieve
a RST frame. This makes it harder than needed for implementations on top
of netty to handle this in a graceful way.

Modifications:

When the Decoder throws a StreamError of type FRAME_SIZE_ERROR, the
Http2ConnectionHandler will now attempt to send an Http2Header with
status 431 and endOfStream=true

Result:

Implementations now do not have to subclass parts of netty to handle
431s

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
@@ -374,6 +374,7 @@ public class DefaultHttp2Connection implements Http2Connection {
         private DefaultStream parent;
         private IntObjectMap<DefaultStream> children = IntCollections.emptyMap();
         private boolean resetSent;
+        private boolean headersSent;
 
         DefaultStream(int id, State state) {
             this.id = id;
@@ -399,6 +400,17 @@ public class DefaultHttp2Connection implements Http2Connection {
         public Http2Stream resetSent() {
             resetSent = true;
             return this;
+        }
+
+        @Override
+        public Http2Stream headersSent() {
+            headersSent = true;
+            return this;
+        }
+
+        @Override
+        public boolean isHeadersSent() {
+            return headersSent;
         }
 
         @Override
@@ -802,6 +814,16 @@ public class DefaultHttp2Connection implements Http2Connection {
 
         @Override
         public Http2Stream closeRemoteSide() {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public Http2Stream headersSent() {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public boolean isHeadersSent() {
             throw new UnsupportedOperationException();
         }
     }
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
@@ -185,13 +185,19 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
                     };
                     promise = promise.unvoid().addListener(closeStreamLocalListener);
                 }
-                return frameWriter.writeHeaders(ctx, streamId, headers, streamDependency, weight,
-                                                exclusive, padding, endOfStream, promise);
+                ChannelFuture future = frameWriter.writeHeaders(ctx, streamId, headers, streamDependency,
+                                                                weight, exclusive, padding, endOfStream, promise);
+                // Synchronously set the headersSent flag to ensure that we do not subsequently write
+                // other headers containing pseudo-header fields.
+                stream.headersSent();
+
+                return future;
             } else {
                 // Pass headers to the flow-controller so it can maintain their sequence relative to DATA frames.
                 flowController.addFlowControlled(stream,
                         new FlowControlledHeaders(stream, headers, streamDependency, weight, exclusive, padding,
                                                  endOfStream, promise));
+                stream.headersSent();
                 return promise;
             }
         } catch (Http2NoMoreStreamIdsException e) {
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2FrameReader.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2FrameReader.java
@@ -673,7 +673,7 @@ public class DefaultHttp2FrameReader implements Http2FrameReader, Http2FrameSize
          */
         private void headerSizeExceeded() throws Http2Exception {
             close();
-            headerListSizeExceeded(streamId, headersDecoder.configuration().headerTable().maxHeaderListSize());
+            headerListSizeExceeded(streamId, headersDecoder.configuration().headerTable().maxHeaderListSize(), true);
         }
 
         /**
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecUtil.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecUtil.java
@@ -30,7 +30,7 @@ import io.netty.util.internal.UnstableApi;
 import static io.netty.buffer.Unpooled.directBuffer;
 import static io.netty.buffer.Unpooled.unreleasableBuffer;
 import static io.netty.handler.codec.http2.Http2Error.PROTOCOL_ERROR;
-import static io.netty.handler.codec.http2.Http2Exception.streamError;
+import static io.netty.handler.codec.http2.Http2Exception.headerListSizeError;
 import static io.netty.util.CharsetUtil.UTF_8;
 import static java.lang.Math.max;
 import static java.lang.Math.min;
@@ -223,8 +223,10 @@ public final class Http2CodecUtil {
         return max(0, min(state.pendingBytes(), state.windowSize()));
     }
 
-    public static void headerListSizeExceeded(int streamId, long maxHeaderListSize) throws Http2Exception {
-        throw streamError(streamId, PROTOCOL_ERROR, "Header size exceeded max allowed size (%d)", maxHeaderListSize);
+    public static void headerListSizeExceeded(int streamId, long maxHeaderListSize,
+                                              boolean onDecode) throws Http2Exception {
+        throw headerListSizeError(streamId, PROTOCOL_ERROR, onDecode, "Header size exceeded max " +
+                                  "allowed size (%d)", maxHeaderListSize);
     }
 
     static void writeFrameHeaderInternal(ByteBuf out, int payloadLength, byte type,
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
@@ -22,6 +22,7 @@ import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelOutboundHandler;
 import io.netty.channel.ChannelPromise;
 import io.netty.handler.codec.ByteToMessageDecoder;
+import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http2.Http2Exception.CompositeStreamException;
 import io.netty.handler.codec.http2.Http2Exception.StreamException;
 import io.netty.util.concurrent.ScheduledFuture;
@@ -64,6 +65,9 @@ public class Http2ConnectionHandler extends ByteToMessageDecoder implements Http
                                                                             ChannelOutboundHandler {
 
     private static final InternalLogger logger = InternalLoggerFactory.getInstance(Http2ConnectionHandler.class);
+
+    private static final Http2Headers headersTooLarge = new DefaultHttp2Headers().status(
+        HttpResponseStatus.REQUEST_HEADER_FIELDS_TOO_LARGE.codeAsText());
 
     private final Http2ConnectionDecoder decoder;
     private final Http2ConnectionEncoder encoder;
@@ -600,7 +604,51 @@ public class Http2ConnectionHandler extends ByteToMessageDecoder implements Http
      */
     protected void onStreamError(ChannelHandlerContext ctx, @SuppressWarnings("unused") Throwable cause,
                                  StreamException http2Ex) {
-        resetStream(ctx, http2Ex.streamId(), http2Ex.error().code(), ctx.newPromise());
+        final int streamId = http2Ex.streamId();
+        Http2Stream stream = connection().stream(streamId);
+
+        //if this is caused by reading headers that are too large, send a header with status 431
+        if (http2Ex instanceof Http2Exception.HeaderListSizeException &&
+            ((Http2Exception.HeaderListSizeException) http2Ex).duringDecode() &&
+            connection().isServer()) {
+
+            // NOTE We have to check to make sure that a stream exists before we send our reply.
+            // We likely always create the stream below as the stream isn't created until the
+            // header block is completely processed.
+
+            // The case of a streamId referring to a stream which was already closed is handled
+            // by createStream and will land us in the catch block below
+            if (stream == null) {
+                try {
+                    stream = encoder.connection().remote().createStream(streamId, true);
+                } catch (Http2Exception e) {
+                    resetUnknownStream(ctx, streamId, http2Ex.error().code(), ctx.newPromise());
+                    return;
+                }
+            }
+
+            // ensure that we have not already sent headers on this stream
+            if (stream != null && !stream.isHeadersSent()) {
+                handleServerHeaderDecodeSizeError(ctx, stream);
+            }
+        }
+
+        if (stream == null) {
+            resetUnknownStream(ctx, streamId, http2Ex.error().code(), ctx.newPromise());
+        } else {
+            resetStream(ctx, stream, http2Ex.error().code(), ctx.newPromise());
+        }
+    }
+
+    /**
+     * Notifies client that this server has received headers that are larger than what it is
+     * willing to accept. Override to change behavior.
+     *
+     * @param ctx the channel context
+     * @param stream the Http2Stream on which the header was received
+     */
+    protected void handleServerHeaderDecodeSizeError(ChannelHandlerContext ctx, Http2Stream stream) {
+        encoder().writeHeaders(ctx, stream.id(), headersTooLarge, 0, true, ctx.newPromise());
     }
 
     protected Http2FrameWriter frameWriter() {
@@ -631,12 +679,17 @@ public class Http2ConnectionHandler extends ByteToMessageDecoder implements Http
     @Override
     public ChannelFuture resetStream(final ChannelHandlerContext ctx, int streamId, long errorCode,
                                      ChannelPromise promise) {
-        promise = promise.unvoid();
         final Http2Stream stream = connection().stream(streamId);
         if (stream == null) {
-            return resetUnknownStream(ctx, streamId, errorCode, promise);
+            return resetUnknownStream(ctx, streamId, errorCode, promise.unvoid());
         }
 
+       return resetStream(ctx, stream, errorCode, promise);
+    }
+
+    private ChannelFuture resetStream(final ChannelHandlerContext ctx, final Http2Stream stream,
+                                      long errorCode, ChannelPromise promise) {
+        promise = promise.unvoid();
         if (stream.isResetSent()) {
             // Don't write a RST_STREAM frame if we have already written one.
             return promise.setSuccess();
@@ -646,7 +699,7 @@ public class Http2ConnectionHandler extends ByteToMessageDecoder implements Http
             // We cannot write RST_STREAM frames on IDLE streams https://tools.ietf.org/html/rfc7540#section-6.4.
             future = promise.setSuccess();
         } else {
-            future = frameWriter().writeRstStream(ctx, streamId, errorCode, promise);
+            future = frameWriter().writeRstStream(ctx, stream.id(), errorCode, promise);
         }
 
         // Synchronously set the resetSent flag to prevent any subsequent calls
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Exception.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Exception.java
@@ -150,6 +150,28 @@ public class Http2Exception extends Exception {
     }
 
     /**
+     * A specific stream error resulting from failing to decode headers that exceeds the max header size list.
+     * If the {@code id} is not {@link Http2CodecUtil#CONNECTION_STREAM_ID} then a
+     * {@link Http2Exception.StreamException} will be returned. Otherwise the error is considered a
+     * connection error and a {@link Http2Exception} is returned.
+     * @param id The stream id for which the error is isolated to.
+     * @param error The type of error as defined by the HTTP/2 specification.
+     * @param onDecode Whether this error was caught while decoding headers
+     * @param fmt String with the content and format for the additional debug data.
+     * @param args Objects which fit into the format defined by {@code fmt}.
+     * @return If the {@code id} is not
+     * {@link Http2CodecUtil#CONNECTION_STREAM_ID} then a {@link HeaderListSizeException}
+     * will be returned. Otherwise the error is considered a connection error and a {@link Http2Exception} is
+     * returned.
+     */
+    public static Http2Exception headerListSizeError(int id, Http2Error error, boolean onDecode,
+            String fmt, Object... args) {
+        return CONNECTION_STREAM_ID == id ?
+                Http2Exception.connectionError(error, fmt, args) :
+                    new HeaderListSizeException(id, error, String.format(fmt, args), onDecode);
+    }
+
+    /**
      * Check if an exception is isolated to a single stream or the entire connection.
      * @param e The exception to check.
      * @return {@code true} if {@code e} is an instance of {@link Http2Exception.StreamException}.
@@ -210,7 +232,7 @@ public class Http2Exception extends Exception {
     /**
      * Represents an exception that can be isolated to a single stream (as opposed to the entire connection).
      */
-    public static final class StreamException extends Http2Exception {
+    public static class StreamException extends Http2Exception {
         private static final long serialVersionUID = 602472544416984384L;
         private final int streamId;
 
@@ -226,6 +248,19 @@ public class Http2Exception extends Exception {
 
         public int streamId() {
             return streamId;
+        }
+    }
+
+    public static final class HeaderListSizeException extends StreamException {
+        private final boolean decode;
+
+        HeaderListSizeException(int streamId, Http2Error error, String message, boolean decode) {
+            super(streamId, error, message);
+            this.decode = decode;
+        }
+
+        public boolean duringDecode() {
+            return decode;
         }
     }
 
--- a/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2ConnectionHandlerTest.java
+++ b/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2ConnectionHandlerTest.java
@@ -25,6 +25,7 @@ import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelPipeline;
 import io.netty.channel.ChannelPromise;
 import io.netty.channel.DefaultChannelPromise;
+import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http2.Http2CodecUtil.SimpleChannelPromiseAggregator;
 import io.netty.util.ReferenceCountUtil;
 import io.netty.util.concurrent.EventExecutor;
@@ -47,6 +48,7 @@ import static io.netty.buffer.Unpooled.copiedBuffer;
 import static io.netty.handler.codec.http2.Http2CodecUtil.connectionPrefaceBuf;
 import static io.netty.handler.codec.http2.Http2Error.PROTOCOL_ERROR;
 import static io.netty.handler.codec.http2.Http2Error.STREAM_CLOSED;
+import static io.netty.handler.codec.http2.Http2Exception.connectionError;
 import static io.netty.handler.codec.http2.Http2Stream.State.CLOSED;
 import static io.netty.handler.codec.http2.Http2Stream.State.IDLE;
 import static io.netty.handler.codec.http2.Http2TestUtil.newVoidPromise;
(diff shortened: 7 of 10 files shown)


nhnFreespirit:
@mosesn It returns afters ending the headers, so should do either or

Scottmitch:
Can you clarify how this accounts for the following cases:
- Clients shouldn't respond with a 431
- if we have already sent the headers we should also not attempt to send this response
- if create stream fails for a reason which means we aren't allowed to send a RST (e.g. stream may have already existed)

nhnFreespirit:
@Scottmitch 

1. Good point, should be an easy enough fix

2. Will look into this. Not sure how this would happen though unless I misunderstand something. If we are a server (fixing that), we should never send out our own headers before we have read headers on this stream from the client?

3. Looking at the ways that createStream can fail, they all indicate that we cannot send anything on that stream id (unless it was because it already exist, but we just checked that). But any state we can be in where this would be the case, would the existing code not call resetStream anyway?

nhnFreespirit:
Updated to only send 431 headers if server + a bit more testing

Scottmitch:
>  If we are a server (fixing that), we should never send out our own headers before we have read headers on this stream from the client?

`FRAME_SIZE_ERROR` is not restricted to reading header frames at the beginning of a stream.

>  But any state we can be in where this would be the case, would the existing code not call resetStream anyway?

Good point ... looks like this is what `resetStream` does if there is no stream ... I thought we were being a bit more conservative there.

nhnFreespirit:
@Scottmitch 

> FRAME_SIZE_ERROR is not restricted to reading header frames at the beginning of a stream.

This one is a little more tricky...

What we have previously done as a nasty workaround in an implementation was look for "Header size exceeded" in the exception message but even that is not really enough, as Http2CodecUtil.headerListSizeExceeded is called by both Decoder and Encoder and FrameReader so this could be caused by trying to encode headers that are too large as well.

It seems like the nicest (but also somewhat intrusive) thing to to would be to add another enum to Http2Exception that would allow for specifying additional purely internal info that can be used when deciding how to handle an exception? 

Or a bit less intrusive but also less generic, I could create a few StreamError subclasses for the cases of exceeding header list size when encoding/decoding respectively.

I'm happy to do the work, just want to make sure I don't start off in a bad direction.

nhnFreespirit:
Added a more specific exception for trying to decode headers that are too large. 431s should only be sent now when a server is trying to decode headers that are too large.

normanmaurer:
@nmittler @louiscryan PTAL as well

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -600,7 +604,27 @@ protected void onConnectionError(ChannelHandlerContext ctx, Throwable cause, Htt
...
+            // We likely always create the stream below as the stream isn't created until the
+            // header block is completely processed.
+            if (stream == null) {
+                try {
+                    encoder.connection().remote().createStream(streamId, true);
+                } catch (Http2Exception e) {
+                    resetUnknownStream(ctx, streamId, http2Ex.error().code(), ctx.newPromise());
+                }
+            }
+            frameWriter().writeHeaders(ctx, streamId, headersTooLarge, 0, true, ctx.newPromise());
```
nmittler: Could we flush here and then still do a `resetStream` so that we properly clean up this stream?

nmittler:
cc @ejona86

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecUtil.java
```diff
@@ -227,6 +228,11 @@ public static void headerListSizeExceeded(int streamId, long maxHeaderListSize)
         throw streamError(streamId, PROTOCOL_ERROR, "Header size exceeded max allowed size (%d)", maxHeaderListSize);
     }
 
+    public static void headerListSizeExceededOnDecode(int streamId, long maxHeaderListSize) throws Http2Exception {
```
Scottmitch: do we have to differentiate `onDecode`? Maybe just rename `DecodeHeaderListSizeException` to `HeaderListSizeException` and add a `boolean duringDecode()` method if we need to distinguish the inbound/outbound characteristics.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -600,7 +604,29 @@ protected void onConnectionError(ChannelHandlerContext ctx, Throwable cause, Htt
...
+            if (stream == null) {
+                try {
+                    encoder.connection().remote().createStream(streamId, true);
+                } catch (Http2Exception e) {
+                    resetUnknownStream(ctx, streamId, http2Ex.error().code(), ctx.newPromise());
+                }
+            }
+
+            frameWriter().writeHeaders(ctx, streamId, headersTooLarge, 0, true, ctx.newPromise());
+            ctx.flush();
```
Scottmitch: I think this was added bcz of @nmittler's suggestion https://github.com/netty/netty/pull/6060#discussion_r89879340 - can you clarify why this is necessary? We will flush outside the scope of this method in [onError](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java#L555) ... is there a call path I'm missing?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -600,7 +604,29 @@ protected void onConnectionError(ChannelHandlerContext ctx, Throwable cause, Htt
...
+            // header block is completely processed.
+            if (stream == null) {
+                try {
+                    encoder.connection().remote().createStream(streamId, true);
+                } catch (Http2Exception e) {
+                    resetUnknownStream(ctx, streamId, http2Ex.error().code(), ctx.newPromise());
+                }
+            }
+
+            frameWriter().writeHeaders(ctx, streamId, headersTooLarge, 0, true, ctx.newPromise());
```
Scottmitch: The user is allowed to provide additional information (headers, body, etc..) in https://tools.ietf.org/html/rfc6585#section-5. We should provide some mechanism for the user to override the response. Perhaps put this in a protected overridable method which does this behavior by default ... but allows the user to override the behavior if they want.

Scottmitch:
> Or a bit less intrusive but also less generic, I could create a few StreamError subclasses for the cases of exceeding header list size when encoding/decoding respectively.

sgtm. We know when this condition occurs (you already found [Http2CodecUtil#headerListSizeExceeded](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecUtil.java#L226)) so we should can be explicit when this condition occurs so we have a clean/reliable way to trigger this behavior.

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2ConnectionHandlerTest.java
```diff
@@ -307,6 +308,46 @@ public void connectionErrorShouldStartShutdown() throws Exception {
     }
 
     @Test
+    public void serverShouldSend431WhenHeadersAreTooLarge() throws Exception {
+        handler = newHandler();
+        Http2Exception e = new Http2Exception.DecodeHeaderListSizeException(STREAM_ID, Http2Error.FRAME_SIZE_ERROR,
```
Scottmitch: Why `FRAME_SIZE_ERROR` here IIRC `PROTOCOL_ERROR` is more appropriate for this condition.

Scottmitch: same question for the client test

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2ConnectionHandlerTest.java
```diff
@@ -307,6 +308,46 @@ public void connectionErrorShouldStartShutdown() throws Exception {
...
+            "Header size exceeded max allowed size 8196");
+
+        when(connection.isServer()).thenReturn(true);
+        when(remote.lastStreamCreated()).thenReturn(STREAM_ID);
+        when(frameWriter.writeRstStream(eq(ctx), eq(STREAM_ID),
+            eq(Http2Error.FRAME_SIZE_ERROR.code()), eq(promise))).thenReturn(future);
+        handler.exceptionCaught(ctx, e);
+        ArgumentCaptor<Http2Headers> captor = ArgumentCaptor.forClass(Http2Headers.class);
+        verify(frameWriter).writeHeaders(eq(ctx), eq(STREAM_ID),
+            captor.capture(), eq(0), eq(true), eq(promise));
```
Scottmitch: `eq(0)` -> `eq(CONNECTION_STREAM_ID)`

Scottmitch: (same comment in the client test)

nhnFreespirit:
Tried to address all comments/suggestions. Let me know what I'm still missing

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecUtil.java
```diff
@@ -223,8 +224,10 @@ public static int streamableBytes(StreamByteDistributor.StreamState state) {
         return max(0, min(state.pendingBytes(), state.windowSize()));
     }
 
-    public static void headerListSizeExceeded(int streamId, long maxHeaderListSize) throws Http2Exception {
-        throw streamError(streamId, PROTOCOL_ERROR, "Header size exceeded max allowed size (%d)", maxHeaderListSize);
+    public static void headerListSizeExceeded(int streamId, long maxHeaderListSize,
+            boolean onDecode)  throws Http2Exception {
```
Scottmitch: nit: fix style (line up params and remove double space after `)`)

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2FrameReader.java
```diff
@@ -673,7 +673,8 @@ final void close() {
          */
         private void headerSizeExceeded() throws Http2Exception {
             close();
-            headerListSizeExceeded(streamId, headersDecoder.configuration().headerTable().maxHeaderListSize());
+            headerListSizeExceeded(streamId, headersDecoder.configuration().headerTable()
```
Scottmitch: nit: fix style (leave `.maxHeaderListSize(), ` on the previous line and put `true);` on the next line)

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Exception.java
```diff
@@ -229,6 +251,19 @@ public int streamId() {
         }
     }
 
+    public static final class HeaderListSizeException extends StreamException {
+        private boolean decode;
```
Scottmitch: `final`

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Exception.java
```diff
@@ -229,6 +251,19 @@ public int streamId() {
...
+
+        HeaderListSizeException(int streamId, Http2Error error, String message, boolean decode) {
+            super(streamId, error, message);
+            this.decode = decode;
+        }
+
+        boolean duringDecode() {
```
Scottmitch: this is not used anywhere. Do we think it will be useful and if so should we make it `public`?

nhnFreespirit: whoops. I added that for a reason but forgot to actually check it! Fixed

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Exception.java
```diff
@@ -229,6 +251,19 @@ public int streamId() {
         }
     }
 
+    public static final class HeaderListSizeException extends StreamException {
+        private final boolean decode;
```
netkins: MAJOR Make this "decode" field final. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1165)

nhnFreespirit:
Hmmm.... Seeing lots of test failures here now, but "mvn test" is all green locally

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Exception.java
nhnFreespirit: ?

It already is...

Scottmitch: ignore this

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -600,7 +604,41 @@ protected void onConnectionError(ChannelHandlerContext ctx, Throwable cause, Htt
...
                                  StreamException http2Ex) {
-        resetStream(ctx, http2Ex.streamId(), http2Ex.error().code(), ctx.newPromise());
+
+        int streamId = http2Ex.streamId();
+
+        //if this is caused by reading headers that are too large, send a header with status 431
+        if (http2Ex instanceof Http2Exception.HeaderListSizeException && connection().isServer()) {
+            final Http2Stream stream = connection().stream(streamId);
+
+            if (((Http2Exception.HeaderListSizeException) http2Ex).duringDecode()) {
```
Scottmitch: move this into the previous conditional for less branch statements and less indentation:

```java
if (http2Ex instanceof Http2Exception.HeaderListSizeException &&
    ((Http2Exception.HeaderListSizeException) http2Ex).duringDecode() &&
    connection().isServer()) {
..
}
```

ejona86:
Two cases don't seem to be handled well, but... meh?:
 1. If a client sends too-large trailers after the server has already responded with headers, the server sends trailers with a `:status` which is invalid (`Pseudo-header fields MUST NOT appear in trailers`).
 2. If the server half-closes the stream before the client (without a following `RST_STREAM` with `NO_ERROR`... or does that even matter for the Netty-state checking?), then the client could send too-large trailers. At that point it is invalid for the server to send `HEADERS` for the half-closed (local) stream.

In both cases, `RST_STREAM` is the only real option. Client-provided trailers are pretty rare, but thought I should at least mention it.

normanmaurer:
@ejona86 I guess we can handle these cases a followup... @Scottmitch ok to pull in ?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -600,7 +604,41 @@ protected void onConnectionError(ChannelHandlerContext ctx, Throwable cause, Htt
...
+
+    /**
+     * Notifies client that this server has received headers that are larger than what it is
+     * willing to accept. Override to change behavior.
+     *
+     * @param ctx the channel context
+     * @param streamId the id of the stream on which the header was received
+     */
+    protected void handleServerHeaderDecodeSizeError(ChannelHandlerContext ctx, int streamId) {
+        frameWriter().writeHeaders(ctx, streamId, headersTooLarge, 0, true, ctx.newPromise());
```
Scottmitch: Instead of directly calling the `frameWriter()` we should take care that the state of the stream will permit headers being sent. We should check that the state is appropriate for the initial headers, and if not then send a `RST`. It may be best to put this state checking in an internal method and only call the `send error response when it is ok to send initial headers` if the state allows it.

We should also consider using `encoder().writeHeaders` to make sure all the state is updated appropriately.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -600,7 +604,41 @@ protected void onConnectionError(ChannelHandlerContext ctx, Throwable cause, Htt
...
+
+            // NOTE We have to check to make sure that a stream exists before we send our reply.
+            // We likely always create the stream below as the stream isn't created until the
+            // header block is completely processed.
+            if (stream == null) {
+                try {
+                    encoder.connection().remote().createStream(streamId, true);
+                } catch (Http2Exception e) {
+                    resetUnknownStream(ctx, streamId, http2Ex.error().code(), ctx.newPromise());
+                }
```
Scottmitch: We should `return;` here so we don't send headers and another reset.

Scottmitch:
@ejona86 - I share the same concerns. I think your item 1 maps to my item 2 in https://github.com/netty/netty/pull/6060#issuecomment-262644087. AFAIK we don't have a reliable way to check "have headers been sent for stream X". In general we rely upon users to distinguish headers from trailers (and not push pseudo headers in the trailers). I think we should address these concerns before we pull the PR in ... otherwise we may violate the RFC without any explicit user interaction.

nhnFreespirit:
Ok... That is a bit of extra work. I'll have a look. I'm pretty loaded with other stuff for the next few days, so it might be a little while before I have any updates. I guess I can track this in a similar way to how we track whether a GoAway frame has already been sent (don't have the code right here but seem to remember that is a thing)

nhnFreespirit:
Updated with the best way I could find to ensure that it is safe to send headers with a status code and default to the reset if not.

normanmaurer:
@nmittler @ejona86 @Scottmitch PTAL again

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
```diff
@@ -277,6 +278,27 @@ public boolean visit(Http2Stream stream) {
         }
     }
 
+    @Override
+    public void headersSent(int streamId, boolean sent) {
+        Http2Stream stream = stream(streamId);
+        if (stream == null) {
+            // TODO: Throw something?
```
netkins: INFO Complete the task associated to this TODO comment. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1135)

nhnFreespirit: Anyone have a good suggestion on how I should handle this case? I believe this will never happen as the stream id has already been checked in the encoder before calling this mehod

normanmaurer: maybe `assert stream != null` ?

nhnFreespirit: Good idea, I'll take it!

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Connection.java
```diff
@@ -403,4 +403,14 @@
      * Indicates that a {@code GOAWAY} was sent to the remote endpoint and sets the last known stream.
      */
     void goAwaySent(int lastKnownStream, long errorCode, ByteBuf message);
+
+    /**
+     * Indicates that headers has been sent to the remote on this stream.
+     */
+    void headersSent(int streamId, boolean sent);
```
Scottmitch: we can just move this method onto the `Http2Stream` interface to avoid duplicate map lookups, and this way we can also avoid using the stream's property map.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
```diff
@@ -277,6 +278,23 @@ public boolean visit(Http2Stream stream) {
...
+        Http2Stream stream = stream(streamId);
+        assert stream != null : "Invalid stream id";
+        stream.setProperty(headersSentKey, sent);
+    }
+
+    @Override
+    public boolean headersSent(int streamId) {
+        Http2Stream stream = stream(streamId);
+        if (stream == null) {
+            return false;
```
Scottmitch: Not sure if we want the interface to account for this but just because we don't have a stream object doesn't mean we haven't already sent the headers. For example we could have already sent headers and closed the stream. One approach could be if the stream ID is `< nextStreamIdToCreate` for the particular endpoint we return `true`, and document this behavior in the interface.

nhnFreespirit: I see how this makes sense from a generic interface, but in this particular case, if we can't get a stream object, we are not going to send any headers anyway.

nhnFreespirit: Or rather, if we cannot get a stream and cannot create one (becuase streamId < nextStreamIdToCreate) . Updated the comment in the code to explain this.

Scottmitch: Good point. Then the methods on the Http2Steam should be good enough and we don't have to deal with this ambiguity for now (also reviewing on my phone so code access isn't easy)

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -600,7 +604,48 @@ protected void onConnectionError(ChannelHandlerContext ctx, Throwable cause, Htt
...
+                } catch (Http2Exception e) {
+                    resetUnknownStream(ctx, streamId, http2Ex.error().code(), ctx.newPromise());
+                    return;
+                }
+            }
+
+            // ensure that we have not already sent headers on this stream
+            if (!stream.isHeadersSent()) {
+                handleServerHeaderDecodeSizeError(ctx, streamId);
+            }
```
Scottmitch: since we already have a `Http2Stream` object consider extracting out the portions of `resetStream` into a function which takes the `Http2Stream` instead of the streamId to avoid the duplicate map lookup:

```java
private ChannelFuture resetStream(final ChannelHandlerContext ctx, Http2Stream stream, long errorCode,
                                     ChannelPromise promise) {
if (stream.isResetSent()) {
...
}
...
```

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -170,6 +170,7 @@ public ChannelFuture writeHeaders(final ChannelHandlerContext ctx, final int str
                                 "Stream %d in unexpected state: %s", stream.id(), stream.state()));
                 }
             }
+            stream.headersSent();
```
Scottmitch: it may be safer to call this after the write is actually issued to ensure the headers were actually sent (or at least attempted to be sent).

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecUtil.java
```diff
@@ -223,8 +224,11 @@ public static int streamableBytes(StreamByteDistributor.StreamState state) {
         return max(0, min(state.pendingBytes(), state.windowSize()));
     }
 
-    public static void headerListSizeExceeded(int streamId, long maxHeaderListSize) throws Http2Exception {
-        throw streamError(streamId, PROTOCOL_ERROR, "Header size exceeded max allowed size (%d)", maxHeaderListSize);
+    public static void headerListSizeExceeded(int streamId, long maxHeaderListSize,
+                                              boolean onDecode) throws Http2Exception {
+
```
Scottmitch: nit: kill line?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -600,7 +604,48 @@ protected void onConnectionError(ChannelHandlerContext ctx, Throwable cause, Htt
...
+                try {
+                    encoder.connection().remote().createStream(streamId, true);
+                } catch (Http2Exception e) {
+                    resetUnknownStream(ctx, streamId, http2Ex.error().code(), ctx.newPromise());
+                    return;
+                }
+            }
+
+            // ensure that we have not already sent headers on this stream
+            if (!stream.isHeadersSent()) {
```
netkins: BLOCKER NullPointerException might be thrown as 'stream' is nullable here [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2259)

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
nhnFreespirit: If I understand correctly, the case where we pass the headers to the flow controller, they will not get written immediately but enqueued. In this case it seems like we want to ensure that even though we have not technically _sent_ the headers, we still want to make sure that we do not enqueue headers containing a status code as well as they would be transmitted in order and the second header would be in violation of the spec.

Therefore it seems safest (and simplest) to mark headersSent as soon as we had declared out intention to send them. Please let me know if I am missing something here.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
```diff
@@ -402,6 +403,15 @@ public Http2Stream resetSent() {
         }
 
         @Override
+        public Http2Stream headersSent() {
+            headersSent = true;
+            return this;
+        }
+
+        @Override
+        public boolean isHeadersSent() { return headersSent; }
```
nhnFreespirit: Not sure I understand what you mean here. I think I follow the format of the other similar methods?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
Scottmitch: Your understanding is generally correct. However enqueueing headers in the flow controller is primarily to preserve the ordering of trailers relative to data, and for the initial headers we can typically avoid going through the flow controller. Either way if we enqueue or "directly write" it is still not necessarily an "immediate" operation (depends on networking conditions, flush calls, etc...). The motivation for delaying the call to `headersSent()` is if the call to write (or queue a write in flow control) fail for some reason (even if its unlikely), the peer will not receive headers, and so headers have not actually been sent, and we shouldn't necessarily prevent future attempts to send headers if these conditions are corrected. Either way the `headersSent` is a synchronous operation (marked immediately) and the write call is asynchronous.

We do the same thing with [resetSent](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java#L654).

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -600,7 +604,47 @@ protected void onConnectionError(ChannelHandlerContext ctx, Throwable cause, Htt
...
+                }
+            }
+
+            // ensure that we have not already sent headers on this stream
+            if (stream != null && !stream.isHeadersSent()) {
+                handleServerHeaderDecodeSizeError(ctx, streamId);
+            }
+        }
+
+        resetStream(ctx, stream, http2Ex.error().code(), ctx.newPromise());
```
netkins: CRITICAL Null passed for non-null parameter of resetStream(ChannelHandlerContext, Http2Stream, long, ChannelPromise) in io.netty.handler.codec.http2.Http2ConnectionHandler.onStreamError(ChannelHandlerContext, Throwable, Http2Exception$StreamException) [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=findbugs%3ANP_NULL_PARAM_DEREF)

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
nhnFreespirit: Ok! Was just following exactly the example set above by the resetSent methods... :)

`        @Override
        public boolean isResetSent() {
            return resetSent;
        }

        @Override
        public Http2Stream resetSent() {
            resetSent = true;
            return this;
        }`

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -600,7 +604,51 @@ protected void onConnectionError(ChannelHandlerContext ctx, Throwable cause, Htt
...
+
+            // NOTE We have to check to make sure that a stream exists before we send our reply.
+            // We likely always create the stream below as the stream isn't created until the
+            // header block is completely processed.
+
+            // The case of a streamId referring to a stream which was already closed is handled
+            // by createStream and will land us in the catch block below
+            if (stream == null) {
+                try {
+                    encoder.connection().remote().createStream(streamId, true);
```
Scottmitch: do we need to assign `stream` here? Please add a unit test which covers this line either way.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -600,7 +604,51 @@ protected void onConnectionError(ChannelHandlerContext ctx, Throwable cause, Htt
...
+    }
+
+    /**
+     * Notifies client that this server has received headers that are larger than what it is
+     * willing to accept. Override to change behavior.
+     *
+     * @param ctx the channel context
+     * @param streamId the id of the stream on which the header was received
+     */
+    protected void handleServerHeaderDecodeSizeError(ChannelHandlerContext ctx, int streamId) {
```
Scottmitch: should we just pass `Http2Stream` here (instead of `int streamId`) since we already have it, and it must exist?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -631,12 +679,19 @@ public void operationComplete(ChannelFuture future) throws Exception {
     @Override
     public ChannelFuture resetStream(final ChannelHandlerContext ctx, int streamId, long errorCode,
                                      ChannelPromise promise) {
-        promise = promise.unvoid();
+
```
Scottmitch: nit: kill line to be consistent with the function below.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -631,12 +679,19 @@ public void operationComplete(ChannelFuture future) throws Exception {
...
         final Http2Stream stream = connection().stream(streamId);
         if (stream == null) {
+            promise = promise.unvoid();
             return resetUnknownStream(ctx, streamId, errorCode, promise);
```
Scottmitch: nit: you could save an assignment by doing the following:

```java
return resetUnknownStream(ctx, streamId, errorCode, promise.unvoid());
```

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
nhnFreespirit: This was based on similar code we had written elsewhere in a custom Pipeline and we found this to be needed based on trial an error. I'll try to write a test to prove it

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
nhnFreespirit: We could, but since that function only uses the id...

On the other hand, since the stream is potentially more generically useful and we intend for people to override this method as needed... I'll bite! :)

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
nhnFreespirit: oh, I misunderstood. I see what your are saying, now that we pass around the stream rather than the streamId...

Scottmitch:
great work @nhnFreespirit and thanks for your patience!

4.1 (cd458f1)

nhnFreespirit:
Great: Thanks for sticking with this! :)

Status on this now says "Closed with unmerged commits". Is the anything I need to do?

normanmaurer:
Nothing to do...

Scottmitch:
We cherry-pick on the command line instead of use the merge button. I think now GitHub supports cherry-picking so this maybe just for historical reasons.

PR #6070, opened 2016-11-29
make normalize ticksPerWheel more stable and high-performance

mmaxiaolei:
make normalize ticksPerWheel more stable and high-performance

### DIFF
--- a/common/src/main/java/io/netty/util/HashedWheelTimer.java
+++ b/common/src/main/java/io/netty/util/HashedWheelTimer.java
@@ -294,11 +294,13 @@ public class HashedWheelTimer implements Timer {
     }
 
     private static int normalizeTicksPerWheel(int ticksPerWheel) {
-        int normalizedTicksPerWheel = 1;
-        while (normalizedTicksPerWheel < ticksPerWheel) {
-            normalizedTicksPerWheel <<= 1;
-        }
-        return normalizedTicksPerWheel;
+        int n = ticksPerWheel - 1;
+        n |= n >>> 1;
+        n |= n >>> 2;
+        n |= n >>> 4;
+        n |= n >>> 8;
+        n |= n >>> 16;
+        return (n < 0) ? 1 : (n >= 1073741824) ? 1073741824 : n + 1;
     }
 
     /**


common/src/main/java/io/netty/util/HashedWheelTimer.java
```diff
@@ -294,11 +294,13 @@ public HashedWheelTimer(
...
-            normalizedTicksPerWheel <<= 1;
-        }
-        return normalizedTicksPerWheel;
+        int n = ticksPerWheel - 1;
+        n |= n >>> 1;
+        n |= n >>> 2;
+        n |= n >>> 4;
+        n |= n >>> 8;
+        n |= n >>> 16;
+        return (n < 0) ? 1 : (n >= 1073741824) ? 1073741824 : n + 1;
```
normanmaurer: Can you make `1073741824` an static constant with a descriptive name ?

normanmaurer: Also share it with other places in the code.

normanmaurer:
@mmaxiaolei also please sign our icla and adjust commit message to match our template:

http://netty.io/s/ilcla

slandelle:
> high-performance

@mmaxiaolei Could you please share some numbers?

normanmaurer:
@mmaxiaolei @slandelle also I wonder if it really makes a lot of difference as this one is only called in the constructor anyway.

Scottmitch:
I agree with @slandelle and @normanmaurer ... please demonstrate the value this provides via JMH benchmark. Thanks.

fenik17:
IMO benchmark not necessary. This optimization used also in JDK, referring to book "Hacker's Delight" sec 3-2 "Rounding Up/Down to the Next Power of 2".
For example look at ConcurrentHashMap.tableSizeFor().

common/src/main/java/io/netty/util/HashedWheelTimer.java
normanmaurer: @mmaxiaolei could you please address my comment ?

normanmaurer:
I am closing this now as it seems like @mmaxiaolei is not interested in this anymore. Please reopen if still interested and address the comments

Issue #6057, opened 2016-11-22
SslHandlerTest#testTruncatedPacket ByteBuf leak


Scottmitch:
Leak uncovered while building PR https://github.com/netty/netty/pull/6054

https://garage.netty.io/teamcity/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=13958#_focus=69576
```
[22:03:49][io.netty:netty-handler] 22:03:49.358 [defaultEventLoopGroup-27-1] ERROR io.netty.util.ResourceLeakDetector - LEAK: ByteBuf.release() was not called before it's garbage-collected. See http://netty.io/wiki/reference-counted-objects.html for more information.
[22:03:49][io.netty:netty-handler] Recent access records: 4
[22:03:49][io.netty:netty-handler] #4:
[22:03:49][io.netty:netty-handler] 	io.netty.buffer.AdvancedLeakAwareByteBuf.release(AdvancedLeakAwareByteBuf.java:955)
[22:03:49][io.netty:netty-handler] 	io.netty.util.ReferenceCountUtil.release(ReferenceCountUtil.java:84)
[22:03:49][io.netty:netty-handler] 	io.netty.util.ReferenceCountUtil.safeRelease(ReferenceCountUtil.java:109)
[22:03:49][io.netty:netty-handler] 	io.netty.channel.ChannelOutboundBuffer.remove(ChannelOutboundBuffer.java:271)
[22:03:49][io.netty:netty-handler] 	io.netty.channel.embedded.EmbeddedChannel.doWrite(EmbeddedChannel.java:675)
[22:03:49][io.netty:netty-handler] 	io.netty.channel.AbstractChannel$AbstractUnsafe.flush0(AbstractChannel.java:856)
[22:03:49][io.netty:netty-handler] 	io.netty.channel.AbstractChannel$AbstractUnsafe.flush(AbstractChannel.java:823)
[22:03:49][io.netty:netty-handler] 	io.netty.channel.DefaultChannelPipeline$HeadContext.flush(DefaultChannelPipeline.java:1296)
[22:03:49][io.netty:netty-handler] 	io.netty.channel.AbstractChannelHandlerContext.invokeFlush0(AbstractChannelHandlerContext.java:787)
[22:03:49][io.netty:netty-handler] 	io.netty.channel.AbstractChannelHandlerContext.invokeFlush(AbstractChannelHandlerContext.java:779)
[...]

PR #6075, opened 2016-12-01
SslHandlerTest ByteBuf leak

Scottmitch:
Motivation:
We are now more careful to flush alerts that are generated when errors occur. We should also be more careful in unit tests to release any buffers that may be queued due to potential errors wich result in alerts.

Modifications:
- When SslHandlerTest uses EmbeddedChannel we should always call finishAndReleaseAll

Results:
Fixes https://github.com/netty/netty/issues/6057


Scottmitch:
@normanmaurer - FYI

normanmaurer:
@Scottmitch thanks a lot!

normanmaurer:
Cherry-picked into 4.1 (55c291ae5bdb7ac79464c4f838ca0aaaba6518fe) and 4.0 (24d0f3d45d290bb4e8ee5e8107584c6f05dbe28e)

Issue #6015, opened 2016-11-13
PooledByteBufAllocatorTest#testThreadCacheToArenaMappings random failure


slandelle:
Just happened on my machine: OSX, Hotspot 1.8.0_112

```
<<< FAILURE! - in io.netty.buffer.PooledByteBufAllocatorTest
testThreadCacheToArenaMappings(io.netty.buffer.PooledByteBufAllocatorTest)  Time elapsed: 0.215 sec  <<< FAILURE!
java.lang.AssertionError: expected:<1> but was:<0>
	at io.netty.buffer.PooledByteBufAllocatorTest.testThreadCacheToArenaMappings(PooledByteBufAllocatorTest.java:238)
```

normanmaurer:
interesting... never saw this. Can you reproduce it in a regular fashion @slandelle ?

slandelle:
I haven't investigated nor tried to reproduce yet. It happened while working on an unrelated PR on codec-http (so built downstream this failing test). I wanted to report before I might forget.

normanmaurer:
will check

normanmaurer:
Ok I think what is wrong... But no worries its just a race in the test. Looking into it now.

slandelle:
cool :)

normanmaurer:
@slandelle fixed by https://github.com/netty/netty/pull/6077

slandelle:
@normanmaurer Yeah, I saw that :) Great job!

PR #6077, opened 2016-12-01
[#6015] Fix racy PooledByteBufAllocatorTests

normanmaurer:
Motivation:

We had a few tests PooledByteBufAllocatorTests which used parkNanos(...) to give a resource enough time to get destroyed. This is race and may not be good enough.

Modifications:

Ensure the ThreadCache is really destroyed.

Result:

No more racy tests that depend on ThreadCaches.


normanmaurer:
Cherry-picked into 4.1 (f70757da2a71db93c3b2108cef86d8a45ef67c08) and 4.0 (129fa9a)

Issue #6061, opened 2016-11-25
EpollServerDomainSocketChannel.isActive() return false after bind operation succeed


byoengguk:
Netty version: 4.1.0.Final

Context: I found bug on EpollServerDomainSocketChannel class. Its isActive() method always return false. It is required to set isActive = true in doBind() operation like EpollServerSocketChannel.

Steps to reproduce:
1. Use EpollServerDomainSocketChannel class to server bootstrap's channel class
2. Bind any unix domain socket.
3. Server Channel's channelActive() is not called due to EpollServerDomainSocketChannel.isActive() return false.

$ java -version
java version "1.8.0_45"
Java(TM) SE Runtime Environment (build 1.8.0_45-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.45-b02, mixed mode)

Operating system: Cent OS 7.2 64-bit

$ uname -a
Linux  3.10.0-327.22.2.el7.x86_64 #1 SMP Thu Jun 23 17:05:11 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux

normanmaurer:
Will check later today

normanmaurer:
@byoengguk thanks for reporting... you are right. Let me fix this

normanmaurer:
Fixed by https://github.com/netty/netty/pull/6078

PR #6078, opened 2016-12-01
[#6061] Correctly mark EpollServerDomainSocketChannel.isActive() as t…

normanmaurer:
…rue after bind is complete.]

Motivation:

We missed to set active = true in EpollServerDomainSocketChannel.doBind(...) which also means that channelActive(...) was never triggered.

Modifications:

Correct set active = true in doBind(...)

Result:

EpollServerDomainSocketChannel is correctly set to active when bound.


normanmaurer:
Cherry-picked into 4.1 (2de644224cfa383d282c960107a88ebbd5d1409c) and 4.0 (38b26a6145692ff8254f9a45da75baf1a0922a33)

PR #6079, opened 2016-12-01
Correct release buffers in RedisEncoderTest

normanmaurer:
Motivation:

RedisEncoderTest did not release all buffers correctly which could lead to leaks.

Modifications:

Call ByteBuf.release() after buffer is read.

Result:

No more leaks in RedisEncoderTest.


normanmaurer:
Cherry-picked into 4.1 (d1c8b099a50b5960c91f3ac938f2dc202cd3cec7)

PR #6080, opened 2016-12-01
Allow to build netty when sun.misc.Unsafe is not avaible or -Dio.nett…

normanmaurer:
…y.noUnsafe=true is used.

Motivation:

We support using Netty without sun.misc.Unsafe, so we should also support building it without it. This way we can also run all tests without sun.misc.Unsafe and so see if it works as expected.

Modifications:

Correctly skip tests that depend on sun.misc.Unsafe if its not present or -Dio.netty.noUnsafe=true is used.

Result:

Be able to build netty without sun.misc.Unsafe


normanmaurer:
Cherry-picked into 4.1 (5bc447c539e3560391bd9df78f23055d69971a66) and 4.0 (5bc447c539e3560391bd9df78f23055d69971a66)

PR #6081, opened 2016-12-01
SwappedByteBuf.unwrap() should return wrapped buffer.

normanmaurer:
Motivation:

SwappedByteBuf.unwrap() not returned the wrapped buffer but the buffer that was wrapped by the original buffer. This is not correct.

Modifications:

Correctly return wrapped buffer and fix test.

Result:

SwappedByteBuf.unwrap() works as expected.


normanmaurer:
Cherry-picked into 4.0 (5acf8c9) and 4.1 (feae0435b5bc17d6e72a5370947564c891517387)

PR #6082, opened 2016-12-01
PooledByteBufAllocatorTest may has memory visiblity issues as it uses…

normanmaurer:
… non concurrent queue

Motivation:

PooledByteBufAllocatorTest uses an ArrayQueue but access it from multiple threads (not concurrently but still from different threads). This may leak to memory visibility issues.

Modifications:

- Use a concurrent queue
- Some cleanup

Result:

Non racy test code.


normanmaurer:
Cherry-picked into 4.1 (243b2b9f193c8459c56ad9196401c46f9584b7e4) and 4.0 (b5e40b2ddec9c5f516f1b045a5049a5f358f02d8)

Issue #6058, opened 2016-11-23
Question of ioRatio in NioEventLoop and two backed array in SelectedSelectionKeySet


lightningMan:
Hi,
I was reading netty source code these days, netty is really a great project and the source code is so beautiful！

Now I have two questions to ask for your help：
## purpose of ioRatio in [NioEventLoop](https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/nio/NioEventLoop.java#L452).

Does ioRatio mean the elapsed time of io operation?As the method`processSelectedKey` above   is cpu intennsive(copy bytes from tcp buffer to jvm direct buffer), is `ioRatio` simply used to control the proportion of copy bytes and business task?Is there any other purpose,like efficiency? ^^

## perhaps two backed array in [SelectedSelectionKeySet](https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/nio/SelectedSelectionKeySet.java) may be reduced to one.

I see this set is backed by two arrays, but I think use one array can also do the same thing as every time after `NioEventLoop`call `processSelectedKeysOptimized(SelectionKey[] selectedKeys)`, the two `SelectionKey[]` is filled with `Null` point 

> flip method in SelectedSelectionKeySet
```java
SelectionKey[] flip() {
        if (isA) {
            isA = false;
            keysA[keysASize] = null;
            keysBSize = 0;
            return keysA;
        } else {
            isA = true;
            keysB[keysBSize] = null;
            keysASize = 0;
            return keysB;
        }
    }
```
perhaps if the flip method is replaced with `end` method as follows can do the same and also avoid exec`if (isA)` ^^
> end method
```java
   SelectionKey[] end() {
        keys[keysSize] = null;
        keysSize = 0;
        return keys;
    }
```

Thank you for your help!!!

Scottmitch:
> purpose of ioRatio in NioEventLoop.

This provides a rough mechanism to control how much time is spent processing IO related tasks (socket reads, connects, closed, hangups, etc..) vs "non-IO" related tasks. The "non-IO" the Executor related events the IO thread has queued. The lower the number the more time will be spent on "non-IO" tasks, and currently 100 disables the timeout all together and runs all pending "non-IO" tasks.

> perhaps two backed array in SelectedSelectionKeySet may be reduced to one.

[...]

PR #6083, opened 2016-12-01
Use a single array in SelectedSelectionKeySet

Scottmitch:
Motivation:
SelectedSelectionKeySet currently uses 2 arrays internally and users are expected to call flip() to access the underlying array and switch the active array. However we do not concurrently use 2 arrays at the same time and we can get away with using a single array if we are careful about when we reset the elements of the array.

Modifications:
- Introduce SelectedSelectionKeySetSelector which wraps a Selector and ensures we reset the underlying SelectedSelectionKeySet data structures before we select
- The loop bounds in NioEventLoop#processSelectedKeysOptimized can be defined more precisely because we know the real size of the underlying array

Result:
Fixes https://github.com/netty/netty/issues/6058

### DIFF
--- a/transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
+++ b/transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
@@ -116,7 +116,8 @@ public final class NioEventLoop extends SingleThreadEventLoop {
     /**
      * The NIO {@link Selector}.
      */
-    Selector selector;
+    private Selector selector;
+    private Selector unwrappedSelector;
     private SelectedSelectionKeySet selectedKeys;
 
     private final SelectorProvider provider;
@@ -150,15 +151,14 @@ public final class NioEventLoop extends SingleThreadEventLoop {
     }
 
     private Selector openSelector() {
-        final Selector selector;
         try {
-            selector = provider.openSelector();
+            unwrappedSelector = provider.openSelector();
         } catch (IOException e) {
             throw new ChannelException("failed to open a new selector", e);
         }
 
         if (DISABLE_KEYSET_OPTIMIZATION) {
-            return selector;
+            return unwrappedSelector;
         }
 
         final SelectedSelectionKeySet selectedKeySet = new SelectedSelectionKeySet();
@@ -179,12 +179,12 @@ public final class NioEventLoop extends SingleThreadEventLoop {
 
         if (!(maybeSelectorImplClass instanceof Class) ||
                 // ensure the current selector implementation is what we can instrument.
-                !((Class<?>) maybeSelectorImplClass).isAssignableFrom(selector.getClass())) {
+                !((Class<?>) maybeSelectorImplClass).isAssignableFrom(unwrappedSelector.getClass())) {
             if (maybeSelectorImplClass instanceof Throwable) {
                 Throwable t = (Throwable) maybeSelectorImplClass;
-                logger.trace("failed to instrument a special java.util.Set into: {}", selector, t);
+                logger.trace("failed to instrument a special java.util.Set into: {}", unwrappedSelector, t);
             }
-            return selector;
+            return unwrappedSelector;
         }
 
         final Class<?> selectorImplClass = (Class<?>) maybeSelectorImplClass;
@@ -205,8 +205,8 @@ public final class NioEventLoop extends SingleThreadEventLoop {
                         return cause;
                     }
 
-                    selectedKeysField.set(selector, selectedKeySet);
-                    publicSelectedKeysField.set(selector, selectedKeySet);
+                    selectedKeysField.set(unwrappedSelector, selectedKeySet);
+                    publicSelectedKeysField.set(unwrappedSelector, selectedKeySet);
                     return null;
                 } catch (NoSuchFieldException e) {
                     return e;
@@ -219,13 +219,12 @@ public final class NioEventLoop extends SingleThreadEventLoop {
         if (maybeException instanceof Exception) {
             selectedKeys = null;
             Exception e = (Exception) maybeException;
-            logger.trace("failed to instrument a special java.util.Set into: {}", selector, e);
-        } else {
-            selectedKeys = selectedKeySet;
-            logger.trace("instrumented a special java.util.Set into: {}", selector);
+            logger.trace("failed to instrument a special java.util.Set into: {}", unwrappedSelector, e);
+            return unwrappedSelector;
         }
-
-        return selector;
+        selectedKeys = selectedKeySet;
+        logger.trace("instrumented a special java.util.Set into: {}", unwrappedSelector);
+        return new SelectedSelectionKeySetSelector(unwrappedSelector, selectedKeySet);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


Scottmitch:
@lightningMan @normanmaurer - FYI

lightningMan:
pong…

Scott Mitchell <notifications@github.com>于2016年12月1日 周四14:58写道：

johnou:
@lightningMan Scott is letting you know that he refactored SelectedSelectionKeySet, you can review the code here https://github.com/netty/netty/pull/6083/files

lightningMan:
@johnou thank you for your Friendly reminder, I have reviewed the code ^^

normanmaurer:
@Scottmitch I am still not 100 % convinced we need these change and I am a bit concerned about introduce a regression as we also have no tests for this class (as we had none before as well). 

So let me cut a release first without this and then we can pick this one up and see... Ok ?

Scottmitch:
Yip I'm fine with delaying this until 4.1.8.

"Need" is a loaded statement :) Things work today so I guess in that sense I agree, however this simplifies and removes complexity that doesn't seem useful. `SelectedSelectionKeySetSelector.java` introduced in this PR ensures that we reset the array backing the `Set` at the appropriate time, and so in theory this should help with correctness (unless there are bugs in my implementation). Bounding the iteration and removing duplicate `null` checks may also clarify functionality.

normanmaurer:
@Scottmitch haha fair enough :) So let us pick this up as first thing after the release 👍

normanmaurer:
@Scottmitch if you want to pull this in can you please at least add a testcase for the `Set` impl ?

Scottmitch:
@normanmaurer - Fair enough. I added a few unit tests to verify expected behavior. Let me know if you want more.

transport/src/main/java/io/netty/channel/nio/SelectedSelectionKeySet.java
```diff
@@ -13,24 +13,20 @@
  * License for the specific language governing permissions and limitations
  * under the License.
  */
-
 package io.netty.channel.nio;
 
 import java.nio.channels.SelectionKey;
 import java.util.AbstractSet;
+import java.util.Arrays;
```
netkins: MINOR Remove this unused import 'java.util.Arrays'. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
```diff
@@ -572,21 +568,16 @@ private void processSelectedKeysOptimized(SelectionKey[] selectedKeys) {
...
                 // null out entries in the array to allow to have it GC'ed once the Channel close
                 // See https://github.com/netty/netty/issues/2363
-                for (;;) {
-                    i++;
-                    if (selectedKeys[i] == null) {
-                        break;
-                    }
-                    selectedKeys[i] = null;
-                }
+                this.selectedKeys.reset(i + 1);
```
johnou: no null check and break while iterating in reset, intentional?

transport/src/main/java/io/netty/channel/nio/SelectedSelectionKeySet.java
```diff
@@ -107,4 +62,25 @@ public boolean contains(Object o) {
...
+
+    void reset() {
+        reset(0);
+    }
+
+    void reset(int start) {
+        for (; start < size; ++start) {
+            keys[start] = null;
+        }
+        size = 0;
```
johnou: does setting size to 0 when start is non 0 make sense?

transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
Scottmitch: Yes. We know the explicit size of the array and we check for null on insertion into the set.

transport/src/main/java/io/netty/channel/nio/SelectedSelectionKeySet.java
Scottmitch: I think so. As we iterate over the array in `NioEventLoop` we null entries out. This method is used if we decide we are done iterating over the array and want to make the set ready for re-use. Do you see an issue here?

transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
johnou: right but the previous impl stopped iterating on the first non null element found presumably to avoid iterating the full set and burn less cpu cycles?

transport/src/main/java/io/netty/channel/nio/SelectedSelectionKeySet.java
johnou: no issue comes to mind at the moment, just unexpected behaviour that size of the set could be 0 when there are non null entries in the null.

Scottmitch: general use of these methods may result in non `null` entries being in the set when the size is zero, but keep in mind this class is package private and so are these methods. This shouldn't happen in practice based upon the narrow use case of this class (as we iterate we `null` elements out, and if we decide to prematurely stop iteration we call this method to `null` out the remaining elements https://github.com/netty/netty/pull/6083#discussion_r100674659).

transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
Scottmitch: The current code and the previous code should be logically equivalent but have the following differences:
- the previous implementation didn't utilize the size of the set when iterating. We now have a fixed/known boundary to iterate to (`size`) instead of conditional based on the next element contents.
- This set doesn't allow `null` elements to be added, so checking each element for `null` is not necessary.

I pushed some updated code to make the relationship with `this.selectedKeys` more clear.

transport/src/test/java/io/netty/channel/nio/SelectedSelectionKeySetTest.java
```diff
@@ -0,0 +1,66 @@
+/*
+ * Copyright 2013 The Netty Project
```
normanmaurer: 2017

transport/src/main/java/io/netty/channel/nio/SelectedSelectionKeySet.java
```diff
@@ -107,4 +61,21 @@ public boolean contains(Object o) {
...
     }
+
+    void reset() {
+        reset(0);
+    }
+
+    void reset(int start) {
+        for (; start < size; ++start) {
+            keys[start] = null;
+        }
```
normanmaurer: Use `Arrays.fill(...)` ?

transport/src/main/java/io/netty/channel/nio/SelectedSelectionKeySetSelector.java
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright 2016 The Netty Project
```
normanmaurer: 2017 ?

Scottmitch:
@normanmaurer - comments addressed PTAL

Scottmitch:
4.1 (795f318) 4.0 (26124f6)

PR #6084, opened 2016-12-01
Small performance improvements in ResourceLeakDetector

normanmaurer:
Motivation:

42fba015ce82ab4ab30e547c888db82fe74094e9 changed the implemention of ResourceLeakDetector to improve performance. While this was done a branch was missed that can be removed. Beside this  using a Boolean as value for the ConcurrentMap is sub-optimal as when calling remove(key, value) an uncessary instanceof check and cast is needed on each removal.

Modifications:

- Remove branch which is not needed anymore
- Replace usage of Boolean as value type of the ConcurrentMap and use our own special type which only compute hash-code one time and use a == operation for equals(...) to reduce overhead present when using Boolean.

Result:

Faster and cleaner ResourceLeakDetector.

### DIFF
--- a/common/src/main/java/io/netty/util/ResourceLeakDetector.java
+++ b/common/src/main/java/io/netty/util/ResourceLeakDetector.java
@@ -144,7 +144,7 @@ public class ResourceLeakDetector<T> {
     }
 
     /** the collection of active resources */
-    private final ConcurrentMap<DefaultResourceLeak, Boolean> allLeaks = PlatformDependent.newConcurrentHashMap();
+    private final ConcurrentMap<DefaultResourceLeak, LeakEntry> allLeaks = PlatformDependent.newConcurrentHashMap();
 
     private final ReferenceQueue<Object> refQueue = new ReferenceQueue<Object>();
     private final ConcurrentMap<String, Boolean> reportedLeaks = PlatformDependent.newConcurrentHashMap();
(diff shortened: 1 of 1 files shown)


common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -108,6 +108,8 @@
         }
     }
 
+    private static final LeakEntry LEAK_ENTRY = new LeakEntry();
```
Scottmitch: what about the following change:

```diff
diff --git a/common/src/main/java/io/netty/util/ResourceLeakDetector.java b/common/src/main/java/io/netty/util/ResourceLeakDetector.java
index 0c94a5b..027c987 100644
--- a/common/src/main/java/io/netty/util/ResourceLeakDetector.java
+++ b/common/src/main/java/io/netty/util/ResourceLeakDetector.java
@@ -108,8 +108,6 @@ public class ResourceLeakDetector<T> {
         }
     }
 
-    private static final LeakEntry LEAK_ENTRY = new LeakEntry();
-
     // Should be power of two.
     static final int DEFAULT_SAMPLING_INTERVAL = 128;
 
@@ -319,7 +317,7 @@ public class ResourceLeakDetector<T> {
                 creationRecord = null;
             }
 
-            allLeaks.put(this, LEAK_ENTRY);
+            allLeaks.put(this, LeakEntry.INSTANCE);
         }
 
         @Override
@@ -352,7 +350,7 @@ public class ResourceLeakDetector<T> {
         @Override
         public boolean close() {
             // Use the ConcurrentMap remove method, which avoids allocating an iterator.
-            return allLeaks.remove(this, LEAK_ENTRY);
+            return allLeaks.remove(this, LeakEntry.INSTANCE);
         }
 
         @Override
@@ -453,11 +451,12 @@ public class ResourceLeakDetector<T> {
     }
 
     private static final class LeakEntry {
-        private final int hash = System.identityHashCode(this);
+        static final LeakEntry INSTANCE = new LeakEntry();
+        private static final int HASH = System.identityHashCode(INSTANCE);
 
         @Override
         public int hashCode() {
-            return hash;
+            return HASH;
         }
 
         @Override
```

normanmaurer: @Scottmitch sure let me do this... that said I don't think it makes any difference

normanmaurer:
@Scottmitch PTAL again... addressed your comment

normanmaurer:
Cherry-picked into 4.1 (fd0ae840b64120efd2a29ba96f0467fda0f21dad) and 4.0 (cfd016c)

Issue #5930, opened 2016-10-20
"OutOfMemoryError: GC overhead limit exceeded" occurs while using HTTP2 based on netty-4.1.5


liuhualh:
Hi, dears

Our backend service using HTTP2 to send Apple push notification to APNS, but we may load many different SSL certificated in one process.  While testing we found Java execute GC and fail (OutOfMemoryError) after the service instance reaches 2000~

Anaylzed using jvisualvm, we found there are about:

io.netty.util.internal.shaded.org.jctools.queues.MpscArrayQueue 4.8919438505775705  224,333 (4.9%)  147,162,448 (10.9%)
java.lang.Object[]  11.54954768714657   529,635 (11.5%) 963,235,296 (71.7%)

Majority of the java.lang.Object[] is referenced by MpscArrayQueue

I wonder what happens for this case, and if there is some specific notice I have ignored while using netty HTTP2

Thank you for your any advice

Scottmitch:
can you provide a heap dump while the memory leak is visible? Also are you able to isolate this issue into a minimal reproducer?

liuhualh:
This is overview of the heap dump,  java.lang.Object[] occupies 71.4% heap size, and MpsArrayQueue occupies 10.9% heap size

<img width="984" alt="netty-memory-leak-overview" src="https://cloud.githubusercontent.com/assets/3794741/19595961/e593f692-97be-11e6-99ad-633ac37a7d90.png">

From this picture,  manny java.lang.Object[] are referenced by MpsArrayQueue

<img width="992" alt="netty-memory-leak-objectarray" src="https://cloud.githubusercontent.com/assets/3794741/19596097/895a555a-97bf-11e6-9abf-bcf2679b5aca.png">

From this picture, MpsArrayQueue is reference by other object through ThreadLocal, one of is my application thread (RcApnsService.procThread)

<img width="979" alt="netty-memory-leak-objref" src="https://cloud.githubusercontent.com/assets/3794741/19596129/b9e1f912-97bf-11e6-910d-db04ea548b19.png">

However, my application thread does not have reference to Netty buffer explicitly.  The  application thread  waits for user notification and calls Netty channel.write(httpRequest) eventually, I don't know if this make the above reference established.

[...]

PR #6085, opened 2016-12-01
Allow to disable caching in PooledByteBufAllocator for non FastThread…

normanmaurer:
…LocalThreads

Motivation:

If a user allocates a lot from outside the EventLoop we may end up creating a lot of caches in the PooledByteBufAllocator. This may be wasteful and so it may be useful for an other to configure that caches should only be used from within EventLoops.

Modifications:

Add new constructor which allows to configure the caching behaviour.

Result:

More flexible configuration of PooledByteBufAllocator possible

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java
+++ b/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java
@@ -17,6 +17,7 @@
 package io.netty.buffer;
 
 import io.netty.util.concurrent.FastThreadLocal;
+import io.netty.util.concurrent.FastThreadLocalThread;
 import io.netty.util.internal.PlatformDependent;
 import io.netty.util.internal.StringUtil;
 import io.netty.util.internal.SystemPropertyUtil;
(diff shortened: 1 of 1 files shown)


normanmaurer:
@liuhualh FYI as well

buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java
```diff
@@ -141,23 +147,42 @@ public PooledByteBufAllocator() {
...
+     * {@link PooledByteBufAllocator#PooledByteBufAllocator(boolean, int, int, int, int, int, int, int, boolean)}
+     */
+    @Deprecated
     public PooledByteBufAllocator(boolean preferDirect, int nHeapArena, int nDirectArena, int pageSize, int maxOrder,
                                   int tinyCacheSize, int smallCacheSize, int normalCacheSize) {
+        this(preferDirect, nHeapArena, nDirectArena, pageSize, maxOrder, tinyCacheSize, smallCacheSize,
+                normalCacheSize, DEFAULT_USE_CACHE_FOR_ALL_THREADS);
+    }
+
+    public PooledByteBufAllocator(boolean preferDirect, int nHeapArena, int nDirectArena, int pageSize, int maxOrder,
```
netkins: MAJOR Constructor has 9 parameters, which is greater than 7 authorized. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS00107)

normanmaurer: its all good :)

normanmaurer:
Cherry-picked into 4.1 (2b8fd8d43be916e29836943fc9d57ec494251756) and 4.0 (c131ef9f6f0f2effdda9d98f87b3bd05375d7002)

liuhualh:
@normanmaurer from call stack on #5930 , the cache can not be GC is allocated from SupportedCipherSuiteFilter.filterCipherSuites,  is this feature have influence on this allocation path ?

normanmaurer:
@liuhualh this issue was fixed by 705e3f629ab802539586c7d92f7888697a44c450

liuhualh:
I see, Thank you very much

normanmaurer:
@liuhualh np... I will release a new netty version this week which will include it.

Issue #5931, opened 2016-10-20
OpenSslEngine.handshake() Failed to mark a promise as failure because it has succeeded already


johnou:
``` noformat
    2016-10-20 21:48:33,264 WARN  [epollEventLoopGroup-3-4] (io.netty.channel.DefaultChannelPipeline) Failed to mark a promise as failure because it has succeeded already: DefaultChannelPromise@59a67d5(success)
javax.net.ssl.SSLException: engine closed
    at io.netty.handler.ssl.ReferenceCountedOpenSslEngine.handshake()(Unknown Source) ~[netty-handler-4.0.42.Final.jar:4.0.42.Final]
```

Haven't had a chance to investigate further yet but in the meantime I wanted to report the stacktrace.

Scottmitch:
@johnou - can you provide a reproducer? does your use case involve renegotation? If you are not able to provide a reproducer easily ... can you change [ReferenceCountedOpenSslEngine#HANDSHAKE_ENGINE_CLOSED](https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java#L86) to not be statically defined so we can get a stack trace?

Scottmitch:
If you aren't required to support renegotiation I would be curious if the behavior changes when you explicitly disable it.

johnou:
@Scottmitch I would be hesitant to disable renegotiation as we need to support multiple platforms and I doubt they all use the same parameters. Here is the full stacktrace from the client side, I imagine the server side stacktrace is the same.

```noformat
javax.net.ssl.SSLException: engine closed
	at io.netty.handler.ssl.ReferenceCountedOpenSslEngine.handshake(ReferenceCountedOpenSslEngine.java:1294) 
	at io.netty.handler.ssl.ReferenceCountedOpenSslEngine.wrap(ReferenceCountedOpenSslEngine.java:571) 
	at javax.net.ssl.SSLEngine.wrap(SSLEngine.java:509) ~[?:1.8.0_40]
	at io.netty.handler.ssl.SslHandler.wrap(SslHandler.java:699) 
	at io.netty.handler.ssl.SslHandler.wrap(SslHandler.java:522) 
	at io.netty.handler.ssl.SslHandler.flush(SslHandler.java:490) 
	at io.netty.handler.ssl.SslHandler.flush(SslHandler.java:1267) 
[...]

PR #6086, opened 2016-12-01
[#5931] Ensure we not complete the same promise that may be failed be…

normanmaurer:
…cause of outbound handler exception.

Motivation:

It's important that we do not pass in the original ChannelPromise to safeClose(...) as when flush(...) will throw an Exception it will be propagated to the AbstractChannelHandlerContext which will try to fail the promise because of this. This will then fail as it was already completed by safeClose(...).

Modifications:

Create a new ChannelPromise and pass it to safeClose(...).

Result:

No more confusing logs because of failing to fail the promise.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -1292,7 +1292,16 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
         try {
             flush(ctx, closeNotifyPromise);
         } finally {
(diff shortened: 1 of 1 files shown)


normanmaurer:
@Scottmitch PTAL

normanmaurer:
Also @johnou FYI

johnou:
nice catch!

normanmaurer:
Cherry-picked into 4.1 (6dbec4181cabe2d836e4a46b253d2f400fa296ac) and 4.0 (5b457c8a22b5e6ab2e40b8a6eacb7b53458c70b6)

PR #6088, opened 2016-12-01
Support compiling netty with Java9

normanmaurer:
Motivation:

Java9 will be released soon so we should ensure we can compile netty with Java9 and run all our tests. This will help to make sure Netty will be usable with Java9.

Modification:

- Add some workarounds to be able to compile with Java9, note that the full profile is not supported with Java9 atm.
- Remove some usage of internal APIs to be able to compile on java9
- Not support Alpn / Npn and so not run the tests when using Java9 for now. We will do a follow up PR to add support.

Result:

Its possible to build netty and run its testsuite with Java9.

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -105,6 +105,30 @@
             </configuration>
           </execution>
         </executions>
+        <!-- Needed to support JDK9 -->
+        <!-- See https://github.com/netty/netty/issues/6100 -->
+        <dependencies>
+          <dependency>
+            <groupId>org.ow2.asm</groupId>
+            <artifactId>asm</artifactId>
+            <version>${asm.version}</version>
+          </dependency>
+          <dependency>
+            <groupId>org.ow2.asm</groupId>
+            <artifactId>asm-commons</artifactId>
+            <version>${asm.version}</version>
+          </dependency>
+          <dependency>
+            <groupId>org.ow2.asm</groupId>
+            <artifactId>asm-analysis</artifactId>
+            <version>${asm.version}</version>
+          </dependency>
+          <dependency>
+            <groupId>org.ow2.asm</groupId>
+            <artifactId>asm-util</artifactId>
+            <version>${asm.version}</version>
+          </dependency>
+        </dependencies>
       </plugin>
       <!-- Add generated collection sources. -->
       <plugin>
--- a/handler/src/main/java/io/netty/handler/ssl/JdkAlpnSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/JdkAlpnSslEngine.java
@@ -26,6 +26,7 @@ import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLException;
 import javax.net.ssl.SSLHandshakeException;
 
+import io.netty.util.internal.PlatformDependent;
 import org.eclipse.jetty.alpn.ALPN;
 import org.eclipse.jetty.alpn.ALPN.ClientProvider;
 import org.eclipse.jetty.alpn.ALPN.ServerProvider;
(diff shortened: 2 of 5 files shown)


handler/src/main/java/io/netty/handler/ssl/JdkAlpnSslEngine.java
```diff
@@ -39,7 +40,8 @@ static boolean isAvailable() {
     }
 
     private static void updateAvailability() {
-        if (available) {
+        // TODO: Add support for ALPN when using Java9 and still be able to configure it the Netty way.
```
netkins: INFO Complete the task associated to this TODO comment. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1135)

normanmaurer: Will do a follow up PR.

normanmaurer:
@Scottmitch @nmittler FYI

common/pom.xml
```diff
@@ -105,6 +105,29 @@
             </configuration>
           </execution>
         </executions>
+        <!-- Needed to support JDK9, https://issues.apache.org/jira/browse/MSHADE-242 -->
```
Scottmitch: create a followup issue to cleanup and update dependencies when this external issue is resolved?

sbordet:
@normanmaurer you may want to wait before working on JDK9's ALPN, since it is going to change a lot in a future JDK9 snapshot (we talked to Oracle - that's why I know).

normanmaurer:
@sbordet doh! Thanks for the heads-up. Anything on a public mailinglist yet ?

sbordet:
@normanmaurer nothing public, and I can't tell you (usual Oracle secrecy). I can keep you in the loop when the change lands in a JDK9 snapshot though - will be great to have an alternative implementation to validate that the JDK9 ALPN changes are sane (we spoke to Oracle because we thought the existing ALPN support was not "flying").

normanmaurer:
@sbordet yes please keep me in the loop. Thanks a lot buddy!

common/pom.xml
normanmaurer: done https://github.com/netty/netty/issues/6100

normanmaurer:
@Scottmitch PTAL again... Ready to pull in ?

normanmaurer:
Cherry-picked into 4.1 (f332a00e1a1b131d866dc2b86916d97f3a1bcf17) and 4.0 (aa89f37c2a825307245d62eba5e1bd9a245baae6)

md-5:
Hmm, the codegen doesn't work for me on 9:
```
Failed to execute goal org.codehaus.gmaven:groovy-maven-plugin:2.0:execute (generate-collections) on project netty-common: Execution generate-collections of goal org.codehaus.gmaven:groovy-maven-plugin:2.0:execute failed: No signature of method: java.lang.String.equals() is applicable for argument types: (java.lang.String) values: [long]
Possible solutions: expand(), expand(int), plus(java.lang.CharSequence), plus(java.lang.Object), each(groovy.lang.Closure)
-> [Help 1]
```

Java(TM) SE Runtime Environment (build 9-ea+154)

normanmaurer:
@md-5 yep this is since the last update... The broke it again :/

handler/src/main/java/io/netty/handler/ssl/JdkAlpnSslEngine.java
rschmitt: Any updates on this? I've opened https://github.com/netty/netty/issues/6933 to track ALPN support.

PR #6089, opened 2016-12-01
Correct guard against non SSL data in ReferenceCountedOpenSslEngine

normanmaurer:
Motivation:

When non SSL data is passed into SSLEngine.unwrap(...) we need to throw an SSLException. This was not done at the moment. Even worse we threw an IllegalArgumentException as we tried to allocate a direct buffer with capacity of -1.

Modifications:

- Guard against non SSL data and added an unit test.
- Make code more consistent

Result:

Correct behaving SSLEngine implementation.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -796,6 +796,11 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
             }
 
             int packetLength = SslUtils.getEncryptedPacketLength(srcs, srcsOffset);
+
+            if (packetLength == SslUtils.NOT_ENCRYPTED) {
+                throw new NotSslRecordException("not an SSL/TLS record");
+            }
+
             if (packetLength - SslUtils.SSL_RECORD_HEADER_LENGTH > capacity) {
                 // No enough space in the destination buffer so signal the caller
                 // that the buffer needs to be increased.
--- a/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
@@ -130,7 +130,7 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
                             final int len = SslUtils.getEncryptedPacketLength(in, readerIndex);
 
                             // Not an SSL/TLS packet
-                            if (len == -1) {
+                            if (len == SslUtils.NOT_ENCRYPTED) {
                                 handshakeFailed = true;
                                 NotSslRecordException e = new NotSslRecordException(
                                         "not an SSL/TLS record: " + ByteBufUtil.hexDump(in));
@@ -140,7 +140,8 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
                                 SslUtils.notifyHandshakeFailure(ctx, e);
                                 return;
                             }
-                            if (writerIndex - readerIndex - SslUtils.SSL_RECORD_HEADER_LENGTH < len) {
+                            if (len == SslUtils.NOT_ENOUGH_DATA ||
+                                    writerIndex - readerIndex - SslUtils.SSL_RECORD_HEADER_LENGTH < len) {
                                 // Not enough data
                                 return;
                             }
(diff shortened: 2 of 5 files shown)


handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -877,7 +878,7 @@ public static boolean isEncrypted(ByteBuf buffer) {
             throw new IllegalArgumentException(
                     "buffer must have at least " + SslUtils.SSL_RECORD_HEADER_LENGTH + " readable bytes");
         }
-        return getEncryptedPacketLength(buffer, buffer.readerIndex()) != -1;
+        return getEncryptedPacketLength(buffer, buffer.readerIndex()) != SslUtils.NOT_ENCRYPTED;
```
netkins: MAJOR Unnecessary use of fully qualified name 'SslUtils.NOT_ENCRYPTED' due to existing static import 'io.netty.handler.ssl.SslUtils.NOT_ENCRYPTED' [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUnnecessaryFullyQualifiedName)

normanmaurer:
@nmittler you also want this for conscrypt

normanmaurer:
@Scottmitch PTAL

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -796,6 +796,11 @@ public final SSLEngineResult unwrap(
             }
 
             int packetLength = SslUtils.getEncryptedPacketLength(srcs, srcsOffset);
+
+            if (packetLength == SslUtils.NOT_ENCRYPTED) {
```
Scottmitch: should this go above the `BUFFER_UNDERFLOW` above?

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -140,7 +140,8 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) t
                                 SslUtils.notifyHandshakeFailure(ctx, e);
                                 return;
                             }
-                            if (writerIndex - readerIndex - SslUtils.SSL_RECORD_HEADER_LENGTH < len) {
+                            if (writerIndex - readerIndex - SslUtils.SSL_RECORD_HEADER_LENGTH < len ||
+                                    len == SslUtils.NOT_ENOUGH_DATA) {
```
Scottmitch: consider switching the order of conditionals to short circuit if the arithmetic may not make sense because `len` is a "special" value

normanmaurer: @Scottmitch +1

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: @Scottmitch its the same order in the JDK implementation. So let me keep it like this

normanmaurer:
@Scottmitch PTAL again

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
Scottmitch: @normanmaurer - can you provide a link? From what I can tell it looks like the JDK does this check up front.

http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/8b04ee324a1a/src/share/classes/sun/security/ssl/SSLEngineImpl.java#l868 -> http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/8b04ee324a1a/src/share/classes/sun/security/ssl/EngineInputRecord.java#l156

Also it doesn't make sense to report underflow/overflow if the data isn't legitimate.

normanmaurer: Brain fart... let me change

Scottmitch: @normanmaurer - good point ... fine as is

normanmaurer:
Cherry-picked into 4.1 (0ca2c3016b1e8aa1fa8d789ce3eac68ffff3d912) and 4.0 (4d327463c7ab14aafdacb1c6cc3902a0271bfb28)

PR #6090, opened 2016-12-01
Correctly not try to call handshake() when engine is already closed.

normanmaurer:
Motivation:

We need to ensure we not call handshake() when the engine is already closed. Beside this our implementation of isOutboundDone() was not correct as it not took the pending data in the outbound buffer into acount (which may be also generated as part of an ssl alert). Beside this we also called SSL_shutdown(...) while we were still in init state which will produce an error and so noise in the log with openssl later versions.

This is also in some extend related to #5931 .

Modifications:

- Ensure we not call handshake() when already closed
- Correctly implement isOutboundDone()
- Not call SSL_shutdown(...) when still in init state
- Added test-cases

Result:

More correct behaviour of our openssl SSLEngine implementation.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -240,8 +240,7 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
 
     // SSL Engine status variables
     private boolean isInboundDone;
-    private boolean isOutboundDone;
-    private boolean engineClosed;
+    private boolean outboundClosed;
 
     private final boolean clientMode;
     private final ByteBufAllocator alloc;
@@ -274,7 +273,7 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
         this.alloc = checkNotNull(alloc, "alloc");
         apn = (OpenSslApplicationProtocolNegotiator) context.applicationProtocolNegotiator();
         ssl = SSL.newSSL(context.ctx, !context.isClient());
-        session = new ReferenceCountedOpenSslEngine.OpenSslSession(context.sessionContext());
+        session = new OpenSslSession(context.sessionContext());
         networkBIO = SSL.makeNetworkBIO(ssl);
         clientMode = context.isClient();
         engineMap = context.engineMap;
@@ -366,8 +365,7 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
             SSL.freeBIO(networkBIO);
             ssl = networkBIO = 0;
 
-            // internal errors can cause shutdown without marking the engine closed
-            isInboundDone = isOutboundDone = engineClosed = true;
+            isInboundDone = outboundClosed = true;
         }
 
         // On shutdown clear all errors
@@ -545,16 +543,22 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
                 bytesProduced += produced;
                 pendingNet -= produced;
             }
-            // If isOuboundDone is set, then the data from the network BIO
-            // was the close_notify message -- we are not required to wait
+
+            SSLEngineResult.HandshakeStatus hs = mayFinishHandshake(
+                    status != FINISHED ? getHandshakeStatus(pendingNet) : status);
+            final SSLEngineResult.Status rs;
+
+            // If isOutboundDone, then the data from the network BIO
+            // was the close_notify message and all was consumed we are not required to wait
             // for the receipt the peer's close_notify message -- shutdown.
-            if (isOutboundDone) {
+            if (isOutboundDone()) {
+                rs = CLOSED;
                 shutdown();
+            } else {
+                rs = OK;
             }
 
-            return new SSLEngineResult(getEngineStatus(),
-                    mayFinishHandshake(status != FINISHED ? getHandshakeStatus(pendingNet) : status),
-                    bytesConsumed, bytesProduced);
+            return new SSLEngineResult(rs, hs, bytesConsumed, bytesProduced);
         }
         return null;
     }
@@ -581,8 +585,9 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
         }
 
         synchronized (this) {
-            // Check to make sure the engine has not been closed
-            if (isDestroyed()) {
+            // Check to make sure the engine has not been closed and the outbound is not considered as done while the
+            // handshake was not started at all.
+            if (isDestroyed() || handshakeState == HandshakeState.NOT_STARTED && isOutboundDone()) {
                 return CLOSED_NOT_HANDSHAKING;
             }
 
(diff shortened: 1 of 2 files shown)


normanmaurer:
@Scottmitch @nmittler @trustin PTAL 

Also @johnou FYI

nmittler:
@normanmaurer changes LGTM 

@davidben @kruton FYI ... We'll want to port this change over to conscrypt

johnou:
Lgtm, many issues left before next 4.0?

normanmaurer:
@johnou nope... just wrapping up stuff

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -596,11 +602,13 @@ public final SSLEngineResult wrap(
...
                 status = handshake();
                 if (status == NEED_UNWRAP) {
-                    return NEED_UNWRAP_OK;
+                    // Signal if the outbound is done or not.
+                    return isOutboundDone() ? NEED_UNWRAP_CLOSED : NEED_UNWRAP_OK;
                 }
 
-                if (engineClosed) {
-                    return NEED_UNWRAP_CLOSED;
+                if (outboundClosed) {
```
Scottmitch: consider adding a comment here that we explicitly want to use `outboundClosed` instead of calling `outboundClosed()` because we want to drain any bytes if they are available.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -786,7 +791,8 @@ public final SSLEngineResult unwrap(
                 if (status == NEED_WRAP) {
                     return NEED_WRAP_OK;
                 }
-                if (engineClosed) {
+                // Check if the inbound is considered to be closed if so let us try to wrap again.
+                if (isInboundDone()) {
```
Scottmitch: use `isInboundDone` directly to be consistent with the `wrap` code?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1109,7 +1120,8 @@ public final synchronized void closeOutbound() {
 
     @Override
     public final synchronized boolean isOutboundDone() {
-        return isOutboundDone;
+        // check if there is anything left in the outbound buffer.
+        return outboundClosed && (isDestroyed() || SSL.pendingWrittenBytesInBIO(networkBIO) == 0);
```
Scottmitch: is checking `isDestroyed()` necessary? Setting the `destroyed` variable to "true" also means that `outboundClosed` will be set to `true`.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -950,11 +956,10 @@ private int pendingAppData() {
...
     }
 
-    private SSLEngineResult newResult(
+    private SSLEngineResult newResult(SSLEngineResult.Status resultStatus,
             int bytesConsumed, int bytesProduced, SSLEngineResult.HandshakeStatus status) throws SSLException {
-        return new SSLEngineResult(
-                getEngineStatus(), mayFinishHandshake(status != FINISHED ? getHandshakeStatus() : status)
-                , bytesConsumed, bytesProduced);
+        return new SSLEngineResult(resultStatus, mayFinishHandshake(status != FINISHED ? getHandshakeStatus() : status),
+                bytesConsumed, bytesProduced);
```
Scottmitch: nit: line up these args with the args above.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java

normanmaurer:
@Scottmitch PTAL again

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
Scottmitch: but `isDestroyed() == true` means that `outboundClosed == true`, and if this is not the case then we have a bug. Am I missing something?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -367,7 +366,7 @@ public final synchronized void shutdown() {
             ssl = networkBIO = 0;
 
             // internal errors can cause shutdown without marking the engine closed
```
Scottmitch: this comment is obsolete now right?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1109,7 +1122,9 @@ public final synchronized void closeOutbound() {
 
     @Override
     public final synchronized boolean isOutboundDone() {
-        return isOutboundDone;
+        // Check if there is anything left in the outbound buffer.
+        // We need to ensure we only call SSL.pendingWrittenBytesInBIO(...) if the engine was not destroyed yet.
+        return outboundClosed && (isDestroyed() || SSL.pendingWrittenBytesInBIO(networkBIO) == 0);
```
Scottmitch: https://github.com/netty/netty/pull/6090#discussion_r90548328 was collapsed but still has some discussion

me:

> is checking `isDestroyed()` necessary? Setting the `destroyed` variable to "true" also means that `outboundClosed` will be set to `true`.

@normanmaurer 

> yes as we only want to call `SSL.pendingWrittenBytesInBIO(...)` if the engine was not destroyed before. If we not check this it may result in a segfault as the SSL pointer was freed before.

me: 

> but `isDestroyed() == true` means that `outboundClosed == true`, and if this is not the case then we have a bug. Am I missing something?

Scottmitch: brain fart ... I see `isDestroyed()` is necessary now. Below is an alternative to avoid a volatile read:

```java
return outboundClosed && (networkBIO  == 0 || SSL.pendingWrittenBytesInBIO(networkBIO) == 0)
```

normanmaurer: Ok can do this ... anything else?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: yes let me remove it

normanmaurer:
@Scottmitch all addressed... PTAL again

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -760,16 +769,13 @@ public final SSLEngineResult unwrap(
         }
 
         synchronized (this) {
-            // Check to make sure the engine has not been closed
-            if (isDestroyed()) {
+            // Check to make sure the engine has not been closed and the inbound was not marked as done.
+            if (isDestroyed() || (handshakeState == HandshakeState.NOT_STARTED && isInboundDone())) {
```
Scottmitch: nit: you have parens here but not on line 590 ... can we be consistent? strictly speaking they aren't necessary but as long as we are consistent its all good.

normanmaurer: added them... easier to read.

normanmaurer:
Cherry-picked into 4.1 (7ce0f35) and 4.0 (78fb4ecc530362a5b903e4b6c077b3ce8d32a610)

johnou:
@normanmaurer might be that we need similar guards in place for wrap too? I've only seen this happen once in production so if I deploy a canary build we might be waiting awhile..

```
javax.net.ssl.SSLException: SSLEngine closed already
	at io.netty.handler.ssl.SslHandler.wrap(...)(Unknown Source) ~[netty-handler-4.1.8.Final.jar:?]
```

Scottmitch:
@johnou - Would you mind filing another issue and provide any more details that you may have?

johnou:
@Scottmitch sounds good.

johnou:
@Scottmitch https://github.com/netty/netty/issues/6419

PR #6091, opened 2016-12-01
Only schedule a ThreadDeathWatcher task if caches are used.

normanmaurer:
Motivation:

If caches are disabled it does not make sense to schedule a task that will free up memory consumed by the caches.

Modifications:

Do not schedule if caches are disabled.

Result:

Less overhead.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/PoolThreadCache.java
+++ b/buffer/src/main/java/io/netty/buffer/PoolThreadCache.java
@@ -55,15 +55,10 @@ final class PoolThreadCache {
     private final int numShiftsNormalHeap;
     private final int freeSweepAllocationThreshold;
 
-    private int allocations;
+    private final Thread deathWatchThread;
+    private final Runnable freeTask;
 
-    private final Thread thread = Thread.currentThread();
-    private final Runnable freeTask = new Runnable() {
-        @Override
-        public void run() {
-            free0();
-        }
-    };
+    private int allocations;
 
     // TODO: Test if adding padding helps under contention
     //private long pad0, pad1, pad2, pad3, pad4, pad5, pad6, pad7;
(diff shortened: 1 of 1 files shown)


normanmaurer:
@Scottmitch PTAL

buffer/src/main/java/io/netty/buffer/PoolThreadCache.java
```diff
@@ -120,9 +114,22 @@ public void run() {
...
+            freeTask = new Runnable() {
+                @Override
+                public void run() {
+                    free0(Thread.currentThread());
+                }
+            };
+
+            // The thread-local cache will keep a list of pooled buffers which must be returned to
+            // the pool when the thread is not alive anymore.
+            ThreadDeathWatcher.watch(Thread.currentThread(), freeTask);
```
Scottmitch: don't we still want to be sure we use the same thread when we call `watch` and `unwatch`? consider keeping the `Thread` local variable and re-name to `deathWatchThread` to clarify.

buffer/src/main/java/io/netty/buffer/PoolThreadCache.java
```diff
@@ -231,11 +238,14 @@ boolean add(PoolArena<?> area, PoolChunk chunk, long handle, int normCapacity, S
...
-        free0();
+        Thread thread = Thread.currentThread();
+        if (freeTask != null) {
+            ThreadDeathWatcher.unwatch(thread, freeTask);
+        }
+        free0(thread);
     }
 
-    private void free0() {
+    private void free0(Thread thread) {
```
Scottmitch: maybe just use `Thread.currentThread()` in the log statement, and also log the `deathWatchThread` too if they may be different.

buffer/src/main/java/io/netty/buffer/PoolThreadCache.java
normanmaurer: @Scottmitch can do if you like.

normanmaurer:
@Scottmitch PTAL again

buffer/src/main/java/io/netty/buffer/PoolThreadCache.java
```diff
@@ -231,11 +242,14 @@ boolean add(PoolArena<?> area, PoolChunk chunk, long handle, int normCapacity, S
...
-        free0();
+        if (freeTask != null) {
+            assert deathWatchThread != null;
+            ThreadDeathWatcher.unwatch(deathWatchThread, freeTask);
+        }
+        free0(Thread.currentThread());
     }
 
-    private void free0() {
+    private void free0(Thread thread) {
```
Scottmitch: do we need to take a Thread parameter here? It is a bit misleading that this may be necessary to do the free operation. Consider just using `Thread.currentThread()` inline in the log statement.

normanmaurer: Will do... anything else or ok to pull in after it?

Scottmitch: lgtm after this

normanmaurer:
Cherry-picked into 4.1 (a07d71ea1d64388b0d4686373809c4733504e915) and 4.0 (a07d71ea1d64388b0d4686373809c4733504e915)

PR #6094, opened 2016-12-02
Add codec-smtp to all jar.

normanmaurer:
Motivation:

562d8d220028fbb3d62028bc5879a121dff2fdbd added codec-smtp but we missed to add it to the all jar.

Modifications:

Include codec-smtp in the all jar.

Result:

Include all codecs in the all jar.


normanmaurer:
Cherry-picked into 4.1 (efeb4cd8cbb148bb252a07f8a2664cd77ea7b5ea)

Issue #6074, opened 2016-11-30
Consider using an issue template


maseev:
I've noticed that despite the fact there are the [guidelines for contributing](https://github.com/netty/netty/blob/4.1/CONTRIBUTING.md), people still keep creating issues that contain so little information that is sometimes absolutely impossible to do anything without asking additional questions like:

- What Netty's version is a person using?
- What steps do I have to perform in order to reproduce the problem?
- What's the context of the problem (i.e expected behavior - actual behavior)?
- Any additional information: network configuration, JVM version, OS type/version, heap dump, stack trace, etc.

You might have heard about these features that GitHub [provides](https://github.com/blog/2111-issue-and-pull-request-templates). By creating a simple *.md file in the repository, GitHub will automatically use it as a template for any new issues/PRs (different *.md file, though). So, instead of asking people all these questions you might consider using the issue template. Chances are that you'll save a hell of a lot of time by integrating this feature into the current workflow.

[Here's](https://github.com/orientechnologies/orientdb/issues) an example of a project which uses this technique and here are the examples how to create [the pull request template](https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/) and [the issue template](https://help.github.com/articles/creating-an-issue-template-for-your-repository/).

Scottmitch:
@maseev - I had not heard about this ... but this sounds great thanks!

PR #6097, opened 2016-12-02
Use GitHub Issue/PR Template Feature

Scottmitch:
Motivation:
GitHub recently added the ability to setup PR and Issue templates https://github.com/blog/2111-issue-and-pull-request-templates. We should take advantage of this feature to ensure Issues / PRs are properly formed.

Modifications:
- add a .github directory with a CONTRIBUTING.md, ISSUE_TEMPLATE.md, and PULL_REQUEST_TEMPLATE.md file

Result:
Fixes https://github.com/netty/netty/issues/6074.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,0 +1,1 @@
+Please review the [guidelines for contributing](http://netty.io/wiki/developer-guide.html) for this repository.
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,13 @@
+### Expected behavior
+
+### Actual behavior
+
+### Steps to reproduce
+
+### Minimal yet complete reproducer code (or URL to code)
+
+### Netty version
+
+### JVM version (e.g. `java -version`)
+
+### OS version (e.g. `uname -a`)
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,14 @@
+Motivation:
+
+Explain here the context, and why you're making that change.
+What is the problem you're trying to solve.
+
+Modification:
+
+Describe the modifications you've done.
+
+Result:
+
+Fixes #<GitHub issue number>. 
+
+If there is no issue then describe the changes introduced by this PR.


Scottmitch:
@maseev - thanks again for the heads up

@normanmaurer - FYI

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1 @@
+Please review the [guidelines for contributing](ihttp://netty.io/wiki/developer-guide.html) for this repository.
```
johnou: typo ihttp -> http

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,6 @@
+#### Motivation
```
johnou: Why not copy direct from http://netty.io/wiki/writing-a-commit-message.html?

Scottmitch: I'm currently experimenting with the format. Markdown looks nice in UIs that support it and helps visually separate different sections/titles and still looks reasonable as plain text.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,13 @@
+### Expected Behavior
+
+### Actual Behavior
+
+### Steps to Reproduce
+
+### Minimal yet Complete Reproducer Code (or URL to Code)
```
johnou: nit caps? Minimal yet complete reproducer code

.github/PULL_REQUEST_TEMPLATE.md
johnou: Maybe something like this?

```text
**Motivation:**

**Modifications:**

**Result:**

After your change, what will change.
```

.github/ISSUE_TEMPLATE.md
Scottmitch: not strong here but the intention is that this is a "section header/title" (not a sentence)

.github/PULL_REQUEST_TEMPLATE.md
Scottmitch: I'll include the descriptions for each section, but I prefer the Markdown section header. That way extra "hard coded" newlines aren't necessary ... the spacing is determined by the styles.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,6 @@
+### Motivation
```
normanmaurer: I would prefer to not use markdown in commit messages honestly.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,13 @@
...
+
+### Netty Version
```
maseev: For Netty version, you can use something very similar to [this](https://github.com/orientechnologies/orientdb/blob/master/.github/ISSUE_TEMPLATE.md#orientdb-version-operating-system-or-hardware). You can also declare that Netty 3.x is EOL and no longer supported and a person needs to upgrade to 4.0.x or 4.1.x. WDYT?

Scottmitch: I would prefer to not update this template with every new release. I'm fine with leaving the field free form for now and updating it later if the field is being abused.

We could add a note that 3.x is EOL but it may be useful for folks to still report the issue for knowledge sharing, and in-case someone is motivated enough they can fix it.

.github/PULL_REQUEST_TEMPLATE.md
Scottmitch: @normanmaurer - You don't like the way it looks in UIs, just think it is too noisy as plain text, or something else?

.github/ISSUE_TEMPLATE.md
maseev: Fair enough.

.github/PULL_REQUEST_TEMPLATE.md
normanmaurer: I don't like to include markdown as not all tools support it and it looks ugly if they not do. For example I often use `git log` on the cmdline etc.

.github/ISSUE_TEMPLATE.md
johnou: To keep it consistent with the Netty wiki I would remove caps from every word even if it's a section header / title.

.github/PULL_REQUEST_TEMPLATE.md
nmittler: @normanmaurer You're not a fan of hashtags? :)

Honestly, the basic formatting used by this template wouldn't be that ugly with pure text-based tools. The only ugly bit would be if the user added a bunch of markdown within the comments.

.github/ISSUE_TEMPLATE.md
Scottmitch: sgtm

Scottmitch:
build failure is related to setup timeout. I will increase the timeout in another PR.

.github/PULL_REQUEST_TEMPLATE.md
Scottmitch: @normanmaurer - I'll just stay consistent with the wiki ... we can update later if we want.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,14 @@
+Motivation:
+
```
Scottmitch: this extra vertical space hurts my eyes 😜

Scottmitch:
any followup comments?

normanmaurer:
@Scottmitch go for it.

Scottmitch:
4.1 (3539ba6) 4.0 (f342ab4)

Issue #6095, opened 2016-12-02
EventLoop.java shouldn't catch ConcurrentModificationException and just move on.


s1monw:
[EventLoop.java ](https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/nio/NioEventLoop.java#L368) catches CME and simply continues to do what it was doing. I am not sure if that's intentional I mean if you hit this exception something really bad is going on. I wonder if this can happen due to concurrent close of the selector etc. There is also some really questionable code in this class that should be cleaned up like [this](https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/nio/NioEventLoop.java#L192-L204) where some members of internal JDK classes are replaced with a very questionable implementation that according to it's [commit message](https://github.com/netty/netty/commit/3ce9ab2e72235bf2ae45f57de11803249419cb69) a not so sizeable performance gain. I am happy to open another issue for this but this class really needs some love given it's the core of the project.

normanmaurer:
@s1monw I am looking into the first issue ATM. 

The second thing you brought about is more about reduce garbage and unfortunately is needed. Other high perf libraries like for example Aeron are doing basically the same for the same reason (reducing garbage).

normanmaurer:
@s1monw also you can disable the optimization via a system property if it makes you feel uneasy

normanmaurer:
@s1monw https://github.com/netty/netty/pull/6101 PTAL

normanmaurer:
Fixed by https://github.com/netty/netty/pull/6101

s1monw:
> The second thing you brought about is more about reduce garbage and unfortunately is needed. Other high perf libraries like for example Aeron are doing basically the same for the same reason (reducing garbage).

I think there is way more to this than you considered in your reply. Lemme give you some opinions and questions. 

* I really disagree with making crazy optimizations like the default behavior the fact that X% of people need to opt out of something that only  Y% use (I assume this is not necessary for the majority of users) is concerning.

* You are replacing an internal datastrucutre with a highly optimized implementation that does not suffice the contract of the set interface is concerning and I think if you do that then please do it right.

* SelectedSelectionKeySet.java doesn't have a single unit-test in place and it's used inside your event loop, really?

[...]

PR #6101, opened 2016-12-03
[#6095] Remove catching of ConcurrentModificationException as this ca…

normanmaurer:
…n not happen.

Motivation:

We should not catch ConcurrentModificationException as this can never happen because things are executed on the EventLoop thread.

Modifications:

Remove try / catch

Result:

Cleaner code.

### DIFF
--- a/transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
+++ b/transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
@@ -314,12 +314,15 @@ public final class NioEventLoop extends SingleThreadEventLoop {
             execute(new Runnable() {
                 @Override
                 public void run() {
-                    rebuildSelector();
+                    rebuildSelector0();
                 }
             });
             return;
         }
+        rebuildSelector0();
+    }
 
+    private void rebuildSelector0() {
         final Selector oldSelector = selector;
         final Selector newSelector;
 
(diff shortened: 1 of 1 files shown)


normanmaurer:
@s1monw @Scottmitch PTAL

normanmaurer:
Cherry-picked into 4.1 (fc1cdc991e71f504cf0d9fdb5aa60034fda04a28) and 4.0 (afffe3a681e6104d03dff6aa2496c5b38ce7212a)

johnou:
@normanmaurer what if someone invokes rebuildSelector outside of the event loop?

normanmaurer:
@johnou it's always scheduled on the eventloop if it's not already on it

transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
```diff
@@ -337,39 +340,33 @@ public void run() {
         // Register all channels to the new Selector.
         int nChannels = 0;
         for (;;) {
```
s1monw: this outer for loop seem pretty useless to me now? can you if it's still of use at least document what it does? I looked briefly and it seems like we exit with a break right away? I think it was only there to overcome the CME that was caught here

transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
```diff
@@ -337,39 +340,33 @@ public void run() {
...
-                            invokeChannelUnregistered(task, key, e);
-                        }
+                    int interestOps = key.interestOps();
+                    key.cancel();
+                    SelectionKey newKey = key.channel().register(newSelector, interestOps, a);
+                    if (a instanceof AbstractNioChannel) {
+                        // Update SelectionKey
+                        ((AbstractNioChannel) a).selectionKey = newKey;
+                    }
+                    nChannels ++;
```
s1monw: there is an extra space?

transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
```diff
@@ -337,39 +340,33 @@ public void run() {
...
+                } catch (Exception e) {
```
s1monw: what kind of exceptions can be thrown here? would be good to have a doc string explaining what exceptions are recoverable here if possible

transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
```diff
@@ -337,39 +340,33 @@ public void run() {
...
         int nChannels = 0;
         for (;;) {
-            try {
-                for (SelectionKey key: oldSelector.keys()) {
-                    Object a = key.attachment();
-                    try {
-                        if (!key.isValid() || key.channel().keyFor(newSelector) != null) {
-                            continue;
-                        }
+            for (SelectionKey key: oldSelector.keys()) {
```
s1monw: I wonder if you should be sync-ing on the `keys()` set - that's what all the JDK classes do and it's iterated in multiple places here? That seems to be the only sane way to do that? I also see that we close Selectors potentially concurrently that causes a `set.clear()` call in the selector and therefore might trigger the CME

normanmaurer: @s1monw were we call set.clear() concurrently ? As far as I see all these operations are always done in the same thread.

transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
normanmaurer: will do this as a follow up.

transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
normanmaurer: will check

normanmaurer: sou are right... Doh.

transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
s1monw: Selector#close calls it afaik.

transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
s1monw: can you link the followup here

PR #6102, opened 2016-12-04
Double checked locking for thread safe initialization of InternalLoggerFactor…

johnou:
…y (default factory).

Motivation:

Current implementation of getDefaultFactory is not thread safe.

Modifications:

Implement holder pattern for lazy initialization of default factory.

Result:

Thread safe InternalLoggerFactory.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
+++ b/common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
@@ -15,6 +15,8 @@
  */
 package io.netty.util.internal.logging;
 
+import io.netty.util.internal.ObjectUtil;
+
 /**
  * Creates an {@link InternalLogger} or changes the default factory
  * implementation.  This factory allows you to choose what logging framework
@@ -57,7 +59,11 @@ public abstract class InternalLoggerFactory {
      */
     public static InternalLoggerFactory getDefaultFactory() {
         if (defaultFactory == null) {
-            defaultFactory = newDefaultFactory(InternalLoggerFactory.class.getName());
+            synchronized (InternalLoggerFactory.class) {
+                if (defaultFactory == null) {
+                    defaultFactory = newDefaultFactory(InternalLoggerFactory.class.getName());
+                }
+            }
         }
         return defaultFactory;
     }
@@ -65,11 +71,8 @@ public abstract class InternalLoggerFactory {
     /**
      * Changes the default factory.
      */
-    public static void setDefaultFactory(InternalLoggerFactory defaultFactory) {
-        if (defaultFactory == null) {
-            throw new NullPointerException("defaultFactory");
-        }
-        InternalLoggerFactory.defaultFactory = defaultFactory;
+    public static synchronized void setDefaultFactory(InternalLoggerFactory defaultFactory) {
+        InternalLoggerFactory.defaultFactory = ObjectUtil.checkNotNull(defaultFactory, "defaultFactory");
     }
 
     /**


normanmaurer:
@jasontedor ptal

johnou:
Might need to rework this a little bit as invoking io.netty.util.internal.logging.InternalLoggerFactory#setDefaultFactory will invoke io.netty.util.internal.logging.InternalLoggerFactory#newDefaultFactory.

common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
```diff
@@ -31,7 +33,11 @@
  * as possible and shouldn't be called more than once.
  */
 public abstract class InternalLoggerFactory {
-    private static volatile InternalLoggerFactory defaultFactory;
+
+    private static class InternalLoggerHolder {
```
netkins: MAJOR Add a private constructor to hide the implicit public one. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1118)

johnou: Ofc

johnou:
Double checked locking might be the way to go.

normanmaurer:
Sounds good

johnou:
Replaced holder with double checked locking, unfortunately it is mutable and this seemed like the simplest approach.

normanmaurer:
LGTM...

jasontedor:
Is there really a problem with two threads racing to initialize `defaultFactory`, they are just going to initialize it to the same instance? Also, this change does not protect against two threads racing via `getDefaultFactory/setDefaultFactory`, one initializing `defaultFactory` via `getDefaultFactory` when `defaultFactory` is `null`, and the other initializing the default factory via `setDefaultFactory`.

Taking a step back though, is it really necessary that `defaultFactory` be mutable after it is initialized? I wonder if it would be better to just rule that out:

```diff
diff --git a/common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java b/common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
index 9f85e3646..32b143d3d 100644
--- a/common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
+++ b/common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
@@ -16,6 +16,8 @@
 
 package io.netty.util.internal.logging;
 
+import java.util.concurrent.atomic.AtomicBoolean;
+
 /**
  * Creates an {@link InternalLogger} or changes the default factory
  * implementation.  This factory allows you to choose what logging framework
@@ -33,6 +35,7 @@ package io.netty.util.internal.logging;
  */
 public abstract class InternalLoggerFactory {
 
+    private static AtomicBoolean set = new AtomicBoolean();
     private static volatile InternalLoggerFactory defaultFactory;
 
     @SuppressWarnings("UnusedCatchParameter")
@@ -54,24 +57,31 @@ public abstract class InternalLoggerFactory {
     }
 
     /**
-     * Returns the default factory.  The initial default factory is
-     * {@link JdkLoggerFactory}.
+     * Get the default factory that was either initialized automatically based on logging implementations on the
+     * classpath, or set explicitly via {@link #setDefaultFactory(InternalLoggerFactory)}.
      */
     public static InternalLoggerFactory getDefaultFactory() {
-        if (defaultFactory == null) {
+        if (set.compareAndSet(false, true)) {
             defaultFactory = newDefaultFactory(InternalLoggerFactory.class.getName());
         }
         return defaultFactory;
     }
 
     /**
-     * Changes the default factory.
+     * Set the default factory. This method must be invoked before the default factory is initialized via
+     * {@link #getDefaultFactory()}, and can not be invoked multiple times.
+     *
+     * @param defaultFactory a non-null implementation of {@link InternalLoggerFactory}
      */
     public static void setDefaultFactory(InternalLoggerFactory defaultFactory) {
         if (defaultFactory == null) {
             throw new NullPointerException("defaultFactory");
         }
-        InternalLoggerFactory.defaultFactory = defaultFactory;
+        if (set.compareAndSet(false, true)) {
+            InternalLoggerFactory.defaultFactory = defaultFactory;
+        } else {
+            throw new IllegalStateException("default factory is already set");
+        }
     }
 
     /**
```

normanmaurer:
@jasontedor we can do this as it will be a breaking change.

johnou:
without the double checked locking wouldn't it be possible for getDefaultFactory to return null on one thread while the other is still constructing the factory? I did see the race between defaultFactory and getDefaultFactory but left that as there is a disclaimer [1] on the class.

[1] " * Please note that the new default factory is effective only for the classes
 * which were loaded after the default factory is changed.  Therefore,
 * {@link #setDefaultFactory(InternalLoggerFactory)} should be called as early
 * as possible and shouldn't be called more than once."

jasontedor:
@normanmaurer Did you mean to say "can not"?

jasontedor:
> without the double checked locking wouldn't it be possible for getDefaultFactory to return null on one thread while the other is still constructing the factory?

@johnou You can also protect against with a CAS on an atomic reference but you're absolutely right, without some form of protection this can happen.

normanmaurer:
@jasontedor yes can not... sorry for the confusion

jasontedor:
> yes can not... sorry for the confusion

That's fine, I don't want to distract from this pull request but when can breaking changes be made, especially breaking changes for questionable use-cases (i.e., applications that set the default factory multiple times)?

normanmaurer:
@jasontedor in the next major release aka 5.0

jasontedor:
> in the next major release aka 5.0

I'm assuming this is very far off since there is not a branch for it?

johnou:
Since the logger factories are essentially singletons I think using AtomicReference would be OK (doesn't matter if newDefaultFactory is invoked multiple times.

normanmaurer:
@jasontedor yes we have not start with 5.0 yet.

johnou:
I still prefer the double checked locking for this specific use case but I think of others where cas would be appropriate. As I mentioned in a comment update above I think it's safer like this (code might be refactored or copied). That said I think Jason's proposal would be nice for next major release.

normanmaurer:
@johnou yes I think double looking is "ok" for this use-case and good enough. Let me merge this one

jasontedor:
There is no risk of a dangerous race between two invocations of `getDefaultFactory`. If no protection is going to be added against races between 'getDefaultFactory` and `setDefaultFactory` (where there is risk of a dangerous race), nor my desire to just protect against multiple sets period, I don't think unnecessary locking should be added. If it's going to be added, I think a code comment remarking that it's unneeded but deliberate should be added.

jasontedor:
And, to be clear, I do think protection against the `getDefaultFactory`/`setDefaultFactory` race should be added. Then I'm okay with adding locking here.

johnou:
@jasontedor something like this?

jasontedor:
@johnou Yes, that works, unless I can convince @normanmaurer permitting setting the factory twice is a bug to begin with in which case I think something stronger (like I proposed) should be the path.

jasontedor:
> yes we have not start with 5.0 yet.

@normanmaurer The more I think about this, the more I think it's a *bug* that Netty allows the InternalLoggerFactory to be set twice. Today it means that some objects will be class-level loggers using loggers from one factory (and they never reset because these are static fields), and some other objects will be using loggers from another factory. This already arose on our side where `ThreadLocalRandom` had its logger initialized with one of the built-in factories instead of the factory that we want it to be using (because of that premature optimization). I think you should disallow this, and I think it's a bug that today you do allow it (and, I certainly can't think of a valid use case for it).

So, I'm going to try again. 😇

I think is a bug, it should be fixed sooner rather than later, and I'm happy to open a PR with my proposed change above.

normanmaurer:
@jasontedor I am on a fence here... I agree with you that there should be no use-cases. Its just that netty is used everywhere and I am a bit afraid of breaking some app that uses it ... Let me think about it.

johnou:
@normanmaurer fwiw it is mentioned in the javadoc.. "Therefore,
 * {@link #setDefaultFactory(InternalLoggerFactory)} should be called as early
 * as possible and shouldn't be called more than once."

jasontedor:
@johnou Thank you, that's great. :smile:

@normanmaurer Are you more convinced it's a bug now? The Javadocs say do not call it more than once, the code just does not enforce it. Enforcing would do two things:
 - help users that do not realize they are not calling it early enough call it early enough (so that *all* loggers are created with their desired logger factory)
 - enforce what the Javadocs say regarding not calling `setDefaultFactory` it more than once

normanmaurer:
@johnou @jasontedor ok you convinced me... go for it

fenik17:
Why netty should support many logger factories? It create problems and overheads..
Usually java libraries provide logs only by slf4j api, and the target application uses one logger that wants. slf4j have bridges to logback, jul, log4j and others.

normanmaurer:
@fenik17 netty aims for being zero dependency for its core... Thats why.

fenik17:
@normanmaurer Ok! I will know.

jasontedor:
Thanks @johnou for opening this PR as it triggered a conversation about just enforcing only allowing this to be set once. I opened #6106 to take that approach.

johnou:
👍 thanks for taking the ball!

Issue #6062, opened 2016-11-26
Memory leak with ThreadDeathWatcher


pfpmeijers:
We are suffering from a memory leak:
At certain point in time ~200 M instances of MemoryRegionCache$Entry objects, retained by ~5000 of ThreadDeathWatcher$Entry objects. These ThreadDeathWatcher$Entry objects are all connected into a linked list. 
The ThreadDeathWatcher$Watcher's watchees array only contains 96 of direct references to ThreadDeathWatcher$Entry objects. In normal cases the size of this watchees array equals the total items in the linked list.
Hence, it looks like the linked list is not updated properly.

Is this a known issue? Any info on root cause, solution, work-around, etc.

Thanks in advance, Patrick

normanmaurer:
Can you share a heapdump?

normanmaurer:
Also which Version of meets you are using @pfpmeijers

pfpmeijers:
Its very big (10 GB) and because of security rules I am not allowed to share a complete heap dump. I am really sorry. I still hope there is some help possible without.
I can share specific views if you indicate what you want to see in the dump.

We use following version:
        <version.netty>3.2.6.Final</version.netty>

normanmaurer:
Sorry but 3.x is EOL ... please Upgrade to 4.x

normanmaurer:
That said you could also try upgrading to latest 3.x version. Yours is super old

normanmaurer:
@pfpmeijers actually you must be using netty 4.x as 3.x has no `MemoryRegionCache`. So please share some screenshots which show the references. Also I wonder what linked-list you are talking about as we not use a linked list in `ThreadDeathWatcher `

pfpmeijers:
I mean the ThreadDeathWatcher$Entry, see following picture. The MpscLinkedQueue refers to the head and the tail.
![image](https://cloud.githubusercontent.com/assets/8824711/20642889/724fad7e-b41b-11e6-8a2c-ca33933b2125.png)
Path to PoolThreadCache and MemoryRegionCache entries:
![image](https://cloud.githubusercontent.com/assets/8824711/20642896/96a225a8-b41b-11e6-9bb2-b0d499b9b1f0.png)
![image](https://cloud.githubusercontent.com/assets/8824711/20642901/a0348ba6-b41b-11e6-979a-c29a7c99e880.png)
  
Links to ThreadDeathWatcher$Entry items:
[...]

PR #6104, opened 2016-12-04
Ensure there are no races in ThreadDeathWatcher, related to [#6062]

normanmaurer:
Motivation:

We used a MPSC queue in ThreadDeathWatcher and checked if it empty via isEmpty() from multiple threads if very unlucky. Depending on the implementation this is not safe and may even produce things like live-locks.

Modification:

- Change to use a MPSP queue
- Some simplifications

Result:

No more risk to run into issues when multiple threads call watch(...) / unwatch(...) concurrently.

### DIFF
--- a/common/src/main/java/io/netty/util/ThreadDeathWatcher.java
+++ b/common/src/main/java/io/netty/util/ThreadDeathWatcher.java
@@ -17,6 +17,7 @@
 package io.netty.util;
 
 import io.netty.util.concurrent.DefaultThreadFactory;
+import io.netty.util.internal.ObjectUtil;
 import io.netty.util.internal.StringUtil;
 import io.netty.util.internal.SystemPropertyUtil;
 import io.netty.util.internal.logging.InternalLogger;
@@ -28,7 +29,7 @@ import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * Checks if a thread is alive periodically and runs a task when a thread dies.
@@ -48,8 +49,7 @@ public final class ThreadDeathWatcher {
     // concurrently depending on the implemenation of it in a MPSC queue.
     private static final Queue<Entry> pendingEntries = new ConcurrentLinkedQueue<Entry>();
     private static final Watcher watcher = new Watcher();
-    private static final AtomicBoolean started = new AtomicBoolean();
-    private static volatile Thread watcherThread;
+    private static final AtomicReference<Thread> watcherThread = new AtomicReference<Thread>();
 
     static {
         String poolName = "threadDeathWatcher";
@@ -102,10 +102,13 @@ public final class ThreadDeathWatcher {
     private static void schedule(Thread thread, Runnable task, boolean isWatch) {
         pendingEntries.add(new Entry(thread, task, isWatch));
 
-        if (started.compareAndSet(false, true)) {
-            Thread watcherThread = threadFactory.newThread(watcher);
-            watcherThread.start();
-            ThreadDeathWatcher.watcherThread = watcherThread;
+        // First check if there is no thread active atm and only if this is the case create a new one and try
+        // set it.
+        if (watcherThread.get() == null) {
+            Thread t = threadFactory.newThread(watcher);
+            if (watcherThread.compareAndSet(null, t)) {
+                t.start();
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


normanmaurer:
@Scottmitch FYI... I was thinking how to improve the `awaitInactivity` to be less racy but could not come up with anything good.

common/src/main/java/io/netty/util/ThreadDeathWatcher.java
```diff
@@ -142,22 +146,19 @@ public void run() {
...
                 }
 
                 if (watchees.isEmpty() && pendingEntries.isEmpty()) {
+                    Thread currentThread = Thread.currentThread();
 
                     // Mark the current worker thread as stopped.
                     // The following CAS must always success and must be uncontended,
                     // because only one watcher thread should be running at the same time.
-                    boolean stopped = started.compareAndSet(true, false);
+                    boolean stopped = watcherThread.compareAndSet(currentThread, null);
```
netkins: MAJOR Remove this useless assignment to local variable "stopped". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1854)

common/src/main/java/io/netty/util/ThreadDeathWatcher.java
```diff
@@ -121,10 +125,10 @@ public static boolean awaitInactivity(long timeout, TimeUnit unit) throws Interr
             throw new NullPointerException("unit");
         }
 
-        Thread watcherThread = ThreadDeathWatcher.watcherThread;
-        if (watcherThread != null) {
-            watcherThread.join(unit.toMillis(timeout));
-            return !watcherThread.isAlive();
+        Thread t = watcherThread.get();
+        if (t != null) {
```
Scottmitch: The synchronization approach used by this class is racy:

1. Watcher#run detects there are no more pending entries, so sets watcher thread to `null`.
2. schedule puts something on the `pendingEntries` queue.
3. `awaitInactivity` is called.
4. schedule or Watcher#run sets the watcher thread.

`awaitInactivity` may exit prematurely.

The comments provide a loop hole that this method should only be called after all calls to `watch`. However it makes me a bit uneasy that the code responsible for watching all threads is itself racy. I wonder if we can have a `private static void Thread TERMINATED = ...;` or some token object that is respected to avoid scheduling new work after the thread death watcher is "terminated" (e.g. `awaitInactivity` is called).

normanmaurer: @Scottmitch I don't think so... This would be a breaking change. Maybe we should just deprecate this method and add a new one that does this. WDYT ?

Scottmitch: IMHO the existing behavior is a bug. The javadocs state the expected behavior of the interface, but the code fails to enforce the behavior. My vote is just to fix the code and remove the race condition.

normanmaurer: @Scottmitch where does the javadoc state that you are now allowed to add more stuff after this method returns ? Disallowing this is a breakage IMHO.

Scottmitch: The javadocs on `awaitInactivity`:

```
this operation is only useful when you want to ensure that the watcher thread is terminated
<strong>after</strong> your application is shut down and there's no chance of calling
{@link #watch(Thread, Runnable)} afterwards.
```

normanmaurer: Which says nothing about that it will have side effects on the future calls to add / remove methods.

normanmaurer: Thats why I think we should deprecate it and add another method which adds these side-effects.

normanmaurer: Legitime in the sense that it is not forbidden in terms of the API documentation. I don't think it makes a lot of sense tho.. I just would hate to break some app because of such a change and I think we could easily just add a new method that will exactly do this and add this behaviour.

Something like `awaitInactivityAndClose()`

Scottmitch: Are you suggesting leaving the existing code as is? I think that is what I'm struggling with. The existing implementation of `awaitInactivity` is racy and therefore unreliable. I don't yet see the value in preserving this unreliable behavior.

normanmaurer: Not sure I am suggestion to keep it like it is but at least not alter it to have side-effects.

Scottmitch: I am comfortable if we want to offer the following functionality:

- `awaitInactivity` - will block until all currently scheduled jobs are complete, with no race, also allows schedule to be called after this method is called.
- `awaitInactivityAndClose` - will block until all currently scheduled jobs are complete, with no race, all subsequent calls to watch will fail.

The "no race" part being the key :)

normanmaurer: @Scottmitch ok sounds good

common/src/main/java/io/netty/util/ThreadDeathWatcher.java
```diff
@@ -116,18 +121,28 @@ private static void schedule(Thread thread, Runnable task, boolean isWatch) {
...
-            throw new NullPointerException("unit");
-        }
-
-        Thread watcherThread = ThreadDeathWatcher.watcherThread;
-        if (watcherThread != null) {
-            watcherThread.join(unit.toMillis(timeout));
-            return !watcherThread.isAlive();
-        } else {
-            return true;
+    public static boolean awaitInactivity(long amount, TimeUnit unit) throws InterruptedException {
```
Scottmitch: doesn't this have the same race condition described in https://github.com/netty/netty/pull/6104#discussion_r91187728?

normanmaurer: @Scottmitch need to think about this a bit more... That said let me factor out the Queue replacement to another PR as I think we want to pull this in in all cases.

normanmaurer: @Scottmitch see https://github.com/netty/netty/pull/6149

normanmaurer:
let me close this for now and come-up with something better once I have time

PR #6106, opened 2016-12-05
Disallow setting logger factory twice

jasontedor:
Motivation:

InternalLoggerFactory either sets a default logger factory
implementation based on the logging implementations on the classpath, or
applications can set a logger factory explicitly. If applications wait
too long to set the logger factory, Netty will have already set a logger
factory leading to some objects using one logging implementation and
other objets using another logging implementation. This can happen too
if the application tries to set the logger factory twice, which is
likely a bug in the application. Yet, the Javadocs for
InternalLoggerFactory warn against this saying that
InternalLoggerFactory#setLoggerFactory "should be called as early as
possible and shouldn't be called more than once". Instead, Netty should
guard against this.

Modications:

We replace the logger factory field with an atomic reference on which we
can do CAS operations to safely guard against it being set twice. We
also add an internal holder class that captures the static interface of
InternalLoggerFactory that can aid in testing.

Result:

The logging factory can not be set twice, and applications that want to
set the logging factory must do it before any Netty classes are
initialized (or the default logger factory will be set).

### DIFF
--- a/common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
+++ b/common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
@@ -16,24 +16,64 @@
 
 package io.netty.util.internal.logging;
 
+import java.util.concurrent.atomic.AtomicReference;
+
 /**
- * Creates an {@link InternalLogger} or changes the default factory
- * implementation.  This factory allows you to choose what logging framework
- * Netty should use.  The default factory is {@link Slf4JLoggerFactory}.  If SLF4J
- * is not available, {@link Log4JLoggerFactory} is used.  If Log4J is not available,
- * {@link JdkLoggerFactory} is used.  You can change it to your preferred
- * logging framework before other Netty classes are loaded:
- * <pre>
- * {@link InternalLoggerFactory}.setDefaultFactory({@link Log4JLoggerFactory}.INSTANCE);
- * </pre>
- * Please note that the new default factory is effective only for the classes
- * which were loaded after the default factory is changed.  Therefore,
- * {@link #setDefaultFactory(InternalLoggerFactory)} should be called as early
- * as possible and shouldn't be called more than once.
+ * Creates {@link InternalLogger}s. This factory allows you to choose what logging framework Netty should use. The
+ * default factory is {@link Slf4JLoggerFactory}. If SLF4J is not available, {@link Log4JLoggerFactory} is used. If
+ * Log4J is not available, {@link JdkLoggerFactory} is used. You can change it to your preferred logging framework
+ * before other Netty classes are loaded via {@link #setDefaultFactory(InternalLoggerFactory)}. If you want to change
+ * the logger factory, {@link #setDefaultFactory(InternalLoggerFactory)} must be invoked before any other Netty classes
+ * are loaded. Note that {@link #setDefaultFactory(InternalLoggerFactory)}} can not be invoked more than once.
  */
 public abstract class InternalLoggerFactory {
 
-    private static volatile InternalLoggerFactory defaultFactory;
+    private static final InternalLoggerFactoryHolder HOLDER = new InternalLoggerFactoryHolder();
+
+    /**
+     * This class holds a reference to the {@link InternalLoggerFactory}. The raison d'être for this class is primarily
+     * to aid in testing.
+     */
+    static final class InternalLoggerFactoryHolder {
+        private final AtomicReference<InternalLoggerFactory> reference;
+
+        InternalLoggerFactoryHolder() {
+            this(null);
+        }
+
+        InternalLoggerFactoryHolder(final InternalLoggerFactory holder) {
+            this.reference = new AtomicReference<InternalLoggerFactory>(holder);
+        }
+
+        InternalLoggerFactory getFactory() {
+            if (reference.get() == null) {
+                reference.compareAndSet(null, newDefaultFactory(InternalLoggerFactory.class.getName()));
+            }
+            return reference.get();
+        }
+
+        void setFactory(final InternalLoggerFactory factory) {
+            if (factory == null) {
+                throw new NullPointerException("factory");
+            }
+            if (!reference.compareAndSet(null, factory)) {
+                throw new IllegalStateException(
+                        "factory is already set to [" + reference.get() + "], rejecting [" + factory + "]");
+            }
+        }
+
+        InternalLogger getInstance(final Class<?> clazz) {
+            return getInstance(clazz.getName());
+        }
+
+        InternalLogger getInstance(final String name) {
+            return newInstance(name);
+        }
+
+        InternalLogger newInstance(String name) {
+            return getFactory().newInstance(name);
+        }
+    }
 
     @SuppressWarnings("UnusedCatchParameter")
     private static InternalLoggerFactory newDefaultFactory(String name) {
@@ -54,38 +94,35 @@ public abstract class InternalLoggerFactory {
     }
 
     /**
-     * Returns the default factory.  The initial default factory is
-     * {@link JdkLoggerFactory}.
+     * Get the default factory that was either initialized automatically based on logging implementations on the
+     * classpath, or set explicitly via {@link #setDefaultFactory(InternalLoggerFactory)}.
      */
     public static InternalLoggerFactory getDefaultFactory() {
-        if (defaultFactory == null) {
-            defaultFactory = newDefaultFactory(InternalLoggerFactory.class.getName());
-        }
-        return defaultFactory;
+        return HOLDER.getFactory();
     }
 
     /**
-     * Changes the default factory.
+     * Set the default factory. This method must be invoked before the default factory is initialized via
+     * {@link #getDefaultFactory()}, and can not be invoked multiple times.
+     *
+     * @param defaultFactory a non-null implementation of {@link InternalLoggerFactory}
      */
     public static void setDefaultFactory(InternalLoggerFactory defaultFactory) {
-        if (defaultFactory == null) {
-            throw new NullPointerException("defaultFactory");
-        }
-        InternalLoggerFactory.defaultFactory = defaultFactory;
+        HOLDER.setFactory(defaultFactory);
     }
 
     /**
      * Creates a new logger instance with the name of the specified class.
      */
     public static InternalLogger getInstance(Class<?> clazz) {
-        return getInstance(clazz.getName());
+        return HOLDER.getInstance(clazz);
     }
 
     /**
      * Creates a new logger instance with the specified name.
      */
     public static InternalLogger getInstance(String name) {
-        return getDefaultFactory().newInstance(name);
+        return HOLDER.getInstance(name);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


common/src/test/java/io/netty/util/internal/logging/InternalLoggerFactoryTest.java
```diff
@@ -230,4 +246,142 @@ public void shouldNotAllowToSetFactoryTwice() {
...
+        });
+    }
+
+    private static void awaitUnchecked(CyclicBarrier barrier) {
+        try {
+            barrier.await();
+        } catch (InterruptedException exception) {
+            throw new RuntimeException(exception);
+        } catch (BrokenBarrierException exception) {
+            throw new RuntimeException(exception);
```
normanmaurer: `IllegalStateException`

common/src/test/java/io/netty/util/internal/logging/InternalLoggerFactoryTest.java
```diff
@@ -230,4 +246,142 @@ public void shouldNotAllowToSetFactoryTwice() {
...
+                }
+            }
+        });
+    }
+
+    private static void awaitUnchecked(CyclicBarrier barrier) {
+        try {
+            barrier.await();
+        } catch (InterruptedException exception) {
+            throw new RuntimeException(exception);
```
normanmaurer: `IllegalStateException`

common/src/test/java/io/netty/util/internal/logging/InternalLoggerFactoryTest.java
jasontedor: Okay. I'll address this when I force push a squash commit before integration, if that's okay with you @normanmaurer?

common/src/test/java/io/netty/util/internal/logging/InternalLoggerFactoryTest.java
jasontedor: Okay. I'll address this when I force push a squash commit before integration, if that's okay with you @normanmaurer?

johnou:
@jasontedor love your branch names 😛

common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
```diff
@@ -16,24 +16,63 @@
...
+ * are loaded. Note that {@link #setDefaultFactory(InternalLoggerFactory)}} can not be invoked more than once.
  */
 public abstract class InternalLoggerFactory {
 
-    private static volatile InternalLoggerFactory defaultFactory;
+    private static final InternalLoggerFactoryHolder HOLDER = new InternalLoggerFactoryHolder();
+
+    /**
+     * This class holds a reference to the {@link InternalLoggerFactory}. The raison d'être for this class is primarily
+     * to aid in testing.
```
johnou: static holder pattern also works well for lazy init, not sure if it's worth mentioning or not.

johnou:
lgtm

common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
jasontedor: Good point. I'll add a comment to this effect when I squash before integration.

johnou: Thread safe lazy init* pretty sure I read it in a Bloch book.

jasontedor:
@normanmaurer Should I squash?

normanmaurer:
@Scottmitch PTAL as well...

@jasontedor let @Scottmitch have a look as well.

jasontedor:
@normanmaurer Sure thing.

common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
```diff
@@ -16,24 +16,63 @@
...
+                reference.compareAndSet(null, newDefaultFactory(InternalLoggerFactory.class.getName()));
+            }
+            return reference.get();
+        }
+
+        void setFactory(final InternalLoggerFactory factory) {
+            if (factory == null) {
+                throw new NullPointerException("factory");
+            }
+            if (reference.get() != null || !reference.compareAndSet(null, factory)) {
```
Scottmitch: I'm assuming `reference.get() != null` is a performance optimization? If so, can we do with out it? `setFactory` is only allowed to be called 1 time and in the majority case it will actually add overhead.

common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
```diff
@@ -16,24 +16,63 @@
...
+                throw new IllegalStateException("factory is already set");
```
Scottmitch: consider putting the existing factory and the new factory in this message.

common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
```diff
@@ -16,24 +16,63 @@
...
 public abstract class InternalLoggerFactory {
 
-    private static volatile InternalLoggerFactory defaultFactory;
+    private static final InternalLoggerFactoryHolder HOLDER = new InternalLoggerFactoryHolder();
+
+    /**
+     * This class holds a reference to the {@link InternalLoggerFactory}. The raison d'être for this class is primarily
+     * to aid in testing.
+     */
+    static class InternalLoggerFactoryHolder {
```
Scottmitch: make this class `final`?

common/src/test/java/io/netty/util/internal/logging/InternalLoggerFactoryTest.java
```diff
@@ -13,49 +13,63 @@
...
+import static org.junit.Assert.assertNotSame;
+import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 public class InternalLoggerFactoryTest {
     private static final Exception e = new Exception();
-    private InternalLoggerFactory oldLoggerFactory;
+    private InternalLoggerFactory.InternalLoggerFactoryHolder holder;
```
Scottmitch: Seems strange to create a mock of the class under test, and using internal implementation details of the class to verify its behavior.  Seems like we should care about testing the `InternalLoggerFactory` interface and not worry about internal implementation details.

Can we have the tests use a real `InternalLoggerFactory` and respect the "only set once" restriction enforced by the interface?

common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
```diff
@@ -16,24 +16,63 @@
...
+        private final AtomicReference<InternalLoggerFactory> reference;
```
Scottmitch: can you explain the value of this class in this context ... maybe I'm missing something but seems like we could just as easily pull `AtomicReference<InternalLoggerFactory>` out of this class and remove a layer of allocation/complexity, no?

Was this the intention: https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom? I'm not sure if it is worth it in this case bcz the logger will almost certainly be used.

common/src/test/java/io/netty/util/internal/logging/InternalLoggerFactoryTest.java
jasontedor: The methods on `InternalLoggerFactory` being tested here are static, there is no real `InternalLoggerFactory` to use (that's a reason I went the direction I did with the holder).

common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
Scottmitch: gotcha ... testing and static state ... thanks for clarifying.

common/src/test/java/io/netty/util/internal/logging/InternalLoggerFactoryTest.java
Scottmitch: gotcha thanks. https://github.com/netty/netty/pull/6106#discussion_r91399057 helped provide more context too.

common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
jasontedor: I pushed 945f85739b3d348bc819b8bd88e222f39b790f67.

common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
jasontedor: Yup, you got it. You're welcome.

common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
jasontedor: Yeah, I pushed 945f85739b3d348bc819b8bd88e222f39b790f67.

common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
jasontedor: I pushed 945f85739b3d348bc819b8bd88e222f39b790f67.

Scottmitch:
@jasontedor - thanks for the PR! please squash and I can pull in after CI finishes.

jasontedor:
Thanks for reviewing @normanmaurer, @johnou, and @Scottmitch. I'll squash.

jasontedor:
@Scottmitch I squashed in 23542902860f1ec88faa9a15655151f73bd27b53.

Scottmitch:
Build failure is unrelated and captured in https://github.com/netty/netty/issues/6114.

normanmaurer:
Cherry-picked into 4.1 (3c92f2b64a1c6803860ab310348abc9f1cb47ff4) and 4.0 (0dacd52)

@jasontedor thanks again!

normanmaurer:
I reverted the commit in 4.1 and 4.0 as discussed in https://github.com/netty/netty/pull/6135 as we need to do some more changes to make this change "less risky" in terms of breaking users of netty. I will reapply the change to 4.1 and 4.0 once a new release was cut so we have some more time to find a good way to apply this. As stated in https://github.com/netty/netty/pull/6135 I think in general the change makes sense and is good, we just need some extra work ...

jasontedor:
> As stated in #6135 I think in general the change makes sense and is good, we just need some extra work ...

Thanks for the update @normanmaurer. I look forward to collaborating on a solution for this. :smile:

normanmaurer:
@jasontedor same here... sorry that I had to revert it :(

Scottmitch:
@jasontedor - Thanks for understanding. We will resolve ASAP after the release as discussed in https://github.com/netty/netty/pull/6135#issuecomment-268140716.

normanmaurer:
@chrisvest I think we might want to cherry-pick this one into main... WDYT ?

chrisvest:
Replaced by #11997

PR #6107, opened 2016-12-05
NIO ServerChannel shouldn't close because of Exception

Scottmitch:
Motivation:
e102a008b63d0063581ba242539f0a8d473cae00 changed a conditional where previously the NIO ServerChannel would not be closed in the event of an exception.

Modifications:
- Restore the logic prior to e102a008b63d0063581ba242539f0a8d473cae00 which does not automatically close ServerChannels for IOExceptions

Result:
NIO ServerChannel doesn't close automatically for an IOException.

### DIFF
--- a/transport/src/main/java/io/netty/channel/nio/AbstractNioMessageChannel.java
+++ b/transport/src/main/java/io/netty/channel/nio/AbstractNioMessageChannel.java
@@ -176,7 +176,7 @@ public abstract class AbstractNioMessageChannel extends AbstractNioChannel {
         // accepting incoming connections. (e.g. too many open files)
         return cause instanceof IOException &&
                 !(cause instanceof PortUnreachableException) &&
-                this instanceof ServerChannel;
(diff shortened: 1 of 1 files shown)


Scottmitch:
@normanmaurer - FYI

johnou:
wow, unit test?

normanmaurer:
@Scottmitch doh! Shame on me 😭😭😭

I wonder if we could somehow write a unit test for this (like @johnou said) .... any idea?

Scottmitch:
maximum FD limit is platform dependent, and there may be a few configuration items which need to be set to have a controlled environment. I could try to add a test utility to try to set the limit for each platform, but I'm not sure if maintaining this is worth the effort (it may be unreliable and only work for a subset of environments).

Either way let me pull this in now because I need it for testing and we can consider unit tests as a followup effort.

Scottmitch:
4.1 (002c99e) 4.0 (443b040)

PR #6110, opened 2016-12-06
Fix frame-pointer to improve epoll flamegraph tracing

tjake:
This patch allows linux perf to track the frame pointer from java to epoll native lib.  This cleans up flamegraphs which use epoll.

There is also an issue with the extracted native lib being deleted by the netty library loader util.  I'm not sure how you want to deal with that, perhaps a property to keep it around for perf testing?


normanmaurer:
@tjake I will take care of the deleting. Can you please adjust your commit message to match our template and sign the cla:

http://netty.io/s/icla

tjake:
done.

normanmaurer:
@tjake thanks a lot

Cherry-picked into 4.1 (8d664fa0fdf05339e52e195709838dc7cc290eca) and 4.0 (06c785c)

PR #6112, opened 2016-12-07
Ensure SSLErrorTest also works with boringssl

normanmaurer:
Motivation:

boringssl uses different messages for the ssl alerts which are all uppercase. As we try to match case as well this fails in SSLErrorTest as we expect lower-case.

This test was introduced by 9b7fb2f3620449c8cfafa62fd294cb82600f043a.

Modifications:

Ensure we first translate everything to lower-case before doing the assert.

Result:

SSLErrorTest also pass when boringssl is used.


normanmaurer:
Cherry-picked into 4.1 (41ea9fa3b6ea463d82f043f9d7c0f4c34f1c6376) and 4.0 (ba24690e86a873f69654b3d9c43d98519c20b8b2)

PR #6113, opened 2016-12-07
Correctly handle the case when BUFFER_OVERFLOW happens during unwrap …

normanmaurer:
…but the readable bytes are bigger then the expected applicationBufferSize

Motivation:

We need to ensure we handle the case when BUFFER_OVERFLOW happens during unwrap but the readable bytes are bigger then the expected applicationBufferSize. Otherwise we may produce an IllegalArgumentException as we will try to allocate a buffer with capacity < 0.

Modifications:

- Guard against this case.
- Ensure we not double release buffer on exception when doing unwrap.

Result:

No more exception when running testsuite with java 9.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -1042,8 +1042,20 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
                     if (readableBytes > 0) {
                         decoded = true;
                         ctx.fireChannelRead(decodeOut);
+
+                        // This buffer was handled, null it out.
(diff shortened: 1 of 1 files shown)


johnou:
@normanmaurer works with JRE8 and fails with JRE9?

normanmaurer:
Yes... that said I think it is a valid behavior of a SSLEngine implementation

johnou:
Nice catch!

normanmaurer:
@Scottmitch PTAL .... once this is in we will get successful builds with java9 on our CI servers 👍

normanmaurer:
Cherry-picked into 4.1 (2055f4c) and 4.0 (b1d6b30)

PR #6117, opened 2016-12-08
Ensure leak aware buffers correctly close the ResourceLeakTracker

normanmaurer:
Motivation:

We should assert that the leak aware buffers correctly close the ResourceLeakTracker in the unit tests.

Modifications:

- Keep track of NoopResourceLeakTrackers and check if these were closed once the test completes
- Fix bugs in tests so the buffers are all released.

Result:

Better tests for leak aware buffers


normanmaurer:
Cherry-picked into 4.1 (712c16a) and 4.0 (9d759d2530a32d879f5d165901cf8c70214fb439)

PR #6118, opened 2016-12-08
Allow to disable deleting of the native library file after it is loaded.

normanmaurer:
Motivation:

When profiling it is sometimes needed to still have the native library file avaible. We should allow to disable the explicit deletion and just delete it when the JVM stops.

This is related to #6110

Modifications:

Add io.netty.native.deleteLibAfterLoading system property which allows to disable the explicit delete after laoding

Result:

Possible to profile native libraries better.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
+++ b/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
@@ -43,6 +43,7 @@ public final class NativeLibraryLoader {
     private static final String NATIVE_RESOURCE_HOME = "META-INF/native/";
     private static final String OSNAME;
     private static final File WORKDIR;
+    private static final boolean DELETE_NATIVE_LIB_AFTER_LOADING;
 
     static {
         OSNAME = SystemPropertyUtil.get("os.name", "").toLowerCase(Locale.US).replaceAll("[^a-z0-9]+", "");
@@ -64,6 +65,9 @@ public final class NativeLibraryLoader {
             WORKDIR = tmpdir();
             logger.debug("-Dio.netty.native.workdir: " + WORKDIR + " (io.netty.tmpdir)");
         }
+
+        DELETE_NATIVE_LIB_AFTER_LOADING = SystemPropertyUtil.getBoolean(
+                "io.netty.native.deleteLibAfterLoading", true);
     }
 
     private static File tmpdir() {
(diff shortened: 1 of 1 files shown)


normanmaurer:
@tjake I think this is what you need, can you verify ?

@Scottmitch PTAL

tjake:
+1

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -64,6 +65,9 @@
             WORKDIR = tmpdir();
             logger.debug("-Dio.netty.native.workdir: " + WORKDIR + " (io.netty.tmpdir)");
         }
+
+        DELETE_NATIVE_LIB_AFTER_LOADING = SystemPropertyUtil.getBoolean(
+                "io.netty.native.deleteLibAfterLoading", true);
```
Scottmitch: should this be scoped to a particular library? Just wondering if we ever load multiple libraries using this loader if we want to have a global flag or scope it per load.

normanmaurer: I think it should be global

Scottmitch: couldn't we build the global flag at a higher level if the low level library loader allows individual libraries to be configured?

normanmaurer: What you mean? I am not sure I understand what you are suggesting

Scottmitch:
```java
// NativeLibraryLoader.java

// Users can call this method if they want to respect the default.
// This also allows users to apply their own logic ...
// e.g. maybe EPOLL, KQUEUE, TCNATIVE have their own properties at some point.
public static boolean defaultDeleteAfterLoading() {
  return DELETE_NATIVE_LIB_AFTER_LOADING;
}

public static void loadFirstAvailable(boolean deleteAfterLoading, ClassLoader loader, String... names) {
// load(deleteAfterLoading, loader, name);
}

public static void load(boolean deleteAfterLoading, String name, ClassLoader loader) {
}
```

normanmaurer: @Scottmitch I can do this but I think either you want it for all or for none. So I think it is overkill and makes not a lot of sense in a practical sense. So please tell me if you feel super strong about it and if so I will do, but honestly I think it makes more sense as it is.

normanmaurer:
Cherry-picked into 4.1 (fe2b55c) and 4.0 (6a0f30d)

PR #6119, opened 2016-12-08
Allow to run tests with extra commandline arguments

normanmaurer:
Motivation:

Often its useful to run the tests with different commandline arguments (like different system properties).

Modifications:

Introduce argLine.javaProperties which can be set from the commandline as well to add arguments that should be append when run the unit tests.

Result:

More flexible way to run the tests.


normanmaurer:
Cherry-picked into 4.1 (42fca7a2fb71835ee64f3dd1a55c881038f00b19) and 4.0 (10fcd65)

PR #6122, opened 2016-12-08
HttpToHttp2ConnectionHandlerTest increase setup timeout

Scottmitch:
Motivation:
The 2 second timeout to bootstrap the test can timeout on the build servers. We should increase the timeout so it is less likely under powered or over worked machines are less likely to generate false failures.

Modifications:
- HttpToHttp2ConnectionHandlerTest setup timeout changed from 2 to 5 seconds

Result:
Less false build failures.


normanmaurer:
@Scottmitch ship it

Scottmitch:
4.1 (4639d56)

PR #6123, opened 2016-12-09
Allow to set max BIO memory size

normanmaurer:
Motivation:

By default openssl uses 17k bytes as memory size for its BIO. This can become very wasteful if you need to handle a lot of concurrent connections. We should allow to adjust the used memory and so trade memory usage with cpu usage.

Modification:

- Allow to adjust the max memory used by the BIO
- Adjust SSLEngine implementation to not assume 17k buffers.

Result:

Its now possible to use smaller buffers when using openssl based SSLEngine.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
@@ -22,6 +22,7 @@ import io.netty.util.ReferenceCounted;
 import io.netty.util.ResourceLeakDetector;
 import io.netty.util.ResourceLeakDetectorFactory;
 import io.netty.util.ResourceLeakTracker;
+import io.netty.util.internal.ObjectUtil;
 import io.netty.util.internal.PlatformDependent;
 import io.netty.util.internal.StringUtil;
 import io.netty.util.internal.SystemPropertyUtil;
@@ -85,6 +86,19 @@ public abstract class ReferenceCountedOpenSslContext extends SslContext implemen
                     return SystemPropertyUtil.getBoolean("jdk.tls.rejectClientInitiatedRenegotiation", false);
                 }
             });
+
+    private static final int MAX_BIO_BYTES =
+            AccessController.doPrivileged(new PrivilegedAction<Integer>() {
+                @Override
+                public Integer run() {
+                    int bytes = SystemPropertyUtil.getInt("io.netty.handler.ssl.openssl.maxBioBytes", 0);
+                    if (bytes <= 0) {
+                        return bytes;
+                    }
+                    return Math.max(bytes, SslUtils.MIN_HANDSHAKE_BUFFER_SIZE);
+                }
+            });
+
     private static final List<String> DEFAULT_CIPHERS;
     private static final Integer DH_KEY_LENGTH;
     private static final ResourceLeakDetector<ReferenceCountedOpenSslContext> leakDetector =
@@ -130,7 +144,8 @@ public abstract class ReferenceCountedOpenSslContext extends SslContext implemen
     final Certificate[] keyCertChain;
     final ClientAuth clientAuth;
     final OpenSslEngineMap engineMap = new DefaultOpenSslEngineMap();
-    volatile boolean rejectRemoteInitiatedRenegotiation;
+    private volatile boolean rejectRemoteInitiatedRenegotiation;
+    private volatile int maxBioBytes = MAX_BIO_BYTES;
 
     static final OpenSslApplicationProtocolNegotiator NONE_PROTOCOL_NEGOTIATOR =
             new OpenSslApplicationProtocolNegotiator() {
@@ -266,7 +281,7 @@ public abstract class ReferenceCountedOpenSslContext extends SslContext implemen
                 SSLContext.setOptions(ctx, SSL.SSL_OP_SINGLE_DH_USE);
                 SSLContext.setOptions(ctx, SSL.SSL_OP_NO_SESSION_RESUMPTION_ON_RENEGOTIATION);
 
-                // We do not support compression as the moment so we should explicitly disable it.
+                // We do not support compression at the moment so we should explicitly disable it.
                 SSLContext.setOptions(ctx, SSL.SSL_OP_NO_COMPRESSION);
 
                 // Disable ticket support by default to be more inline with SSLEngineImpl of the JDK.
@@ -428,6 +443,33 @@ public abstract class ReferenceCountedOpenSslContext extends SslContext implemen
      */
     public void setRejectRemoteInitiatedRenegotiation(boolean rejectRemoteInitiatedRenegotiation) {
         this.rejectRemoteInitiatedRenegotiation = rejectRemoteInitiatedRenegotiation;
+    }
+
+    /**
+     * Returns if remote initiated renegotiation is supported or not.
+     */
+    public boolean getRejectRemoteInitiatedRenegotiation() {
+         return rejectRemoteInitiatedRenegotiation;
+    }
+
+    /**
+     * Set the maximum number of bytes that are used for each BIO instance used in the pair. As two BIO instances are
+     * needed each {@link ReferenceCountedOpenSslEngine} will consume 2 * maxBioSize of native memory. Using {@code 0}
+     * will use the default value of 16 kb. Using a lower number will trade memory usage with higher cpu-usage.
+     */
+    public void setMaxBioBytes(int maxBioBytes) {
+        if (maxBioBytes < 0 || maxBioBytes < SslUtils.MIN_HANDSHAKE_BUFFER_SIZE) {
+            throw new IllegalArgumentException(
+                    "maxBioBytes: " + maxBioBytes + " (expected: >= " + SslUtils.MIN_HANDSHAKE_BUFFER_SIZE + " | 0)");
+        }
+        this.maxBioBytes = ObjectUtil.checkPositiveOrZero(maxBioBytes, "maxBioBytes");
+    }
+
+    /**
+     * Returns the maximum number of bytes that are used for each BIO instance.
+     */
+    public int getMaxBioBytes() {
+        return maxBioBytes;
     }
 
     /**
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -17,7 +17,6 @@ package io.netty.handler.ssl;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
-import io.netty.buffer.Unpooled;
 import io.netty.util.AbstractReferenceCounted;
 import io.netty.util.ReferenceCounted;
 import io.netty.util.ResourceLeakDetector;
@@ -177,8 +176,6 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
 
     private static final String INVALID_CIPHER = "SSL_NULL_WITH_NULL_NULL";
 
-    private static final long EMPTY_ADDR = Buffer.address(Unpooled.EMPTY_BUFFER.nioBuffer());
-
     private static final SSLEngineResult NEED_UNWRAP_OK = new SSLEngineResult(OK, NEED_UNWRAP, 0, 0);
     private static final SSLEngineResult NEED_UNWRAP_CLOSED = new SSLEngineResult(CLOSED, NEED_UNWRAP, 0, 0);
     private static final SSLEngineResult NEED_WRAP_OK = new SSLEngineResult(OK, NEED_WRAP, 0, 0);
@@ -282,10 +279,10 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
         apn = (OpenSslApplicationProtocolNegotiator) context.applicationProtocolNegotiator();
         ssl = SSL.newSSL(context.ctx, !context.isClient());
         session = new OpenSslSession(context.sessionContext());
-        networkBIO = SSL.makeNetworkBIO(ssl);
+        networkBIO = SSL.makeNetworkBIO(ssl, context.getMaxBioBytes());
         clientMode = context.isClient();
         engineMap = context.engineMap;
-        rejectRemoteInitiatedRenegation = context.rejectRemoteInitiatedRenegotiation;
+        rejectRemoteInitiatedRenegation = context.getRejectRemoteInitiatedRenegotiation();
         localCerts = context.keyCertChain;
 
         // Set the client auth mode, this needs to be done via setClientAuth(...) method so we actually call the
@@ -422,15 +419,17 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
     /**
      * Write encrypted data to the OpenSSL network BIO.
      */
-    private int writeEncryptedData(final ByteBuffer src, int len) {
+    private int writeEncryptedData(final ByteBuffer src, int len) throws SSLException {
         final int pos = src.position();
         final int netWrote;
         if (src.isDirect()) {
             final long addr = Buffer.address(src) + pos;
             netWrote = SSL.writeToBIO(networkBIO, addr, len);
-            if (netWrote >= 0) {
+            if (netWrote > 0) {
                 src.position(pos + netWrote);
+                return netWrote;
             }
+            checkRetryBioOperation("BIO_write");
         } else {
             final ByteBuf buf = alloc.directBuffer(len);
             try {
@@ -445,11 +444,13 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
                 src.limit(limit);
 
                 netWrote = SSL.writeToBIO(networkBIO, addr, len);
-                if (netWrote >= 0) {
+                if (netWrote > 0) {
                     src.position(pos + netWrote);
+                    return netWrote;
                 } else {
                     src.position(pos);
                 }
+                checkRetryBioOperation("BIO_write");
             } finally {
                 buf.release();
             }
@@ -496,7 +497,7 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
     /**
      * Read encrypted data from the OpenSSL network BIO
      */
-    private int readEncryptedData(final ByteBuffer dst, final int pending) {
+    private int readEncryptedData(final ByteBuffer dst, final int pending) throws SSLException {
         final int bioRead;
 
         if (dst.isDirect() && dst.remaining() >= pending) {
@@ -507,6 +508,7 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
                 dst.position(pos + bioRead);
                 return bioRead;
             }
+            checkRetryBioOperation("BIO_read");
         } else {
             final ByteBuf buf = alloc.directBuffer(pending);
             try {
@@ -520,6 +522,7 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
                     dst.limit(oldLimit);
                     return bioRead;
                 }
+                checkRetryBioOperation("BIO_read");
             } finally {
                 buf.release();
             }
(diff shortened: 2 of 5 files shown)


normanmaurer:
@trustin @nmittler @Scottmitch this allows to constumize the max number of bytes that are buffered by the BIO. By default it uses 17k buffers which means the whole SSL record can be buffered. The problem is that this may result in problems as you want to handle a lot of concurrent connections. When you lower the number of bytes in the buffer you can safe memory but need more CPU as you will need more often do JNI calls. So as always there are tradeoffs ;)

This is related to https://github.com/netty/netty-tcnative/pull/205

Also @siyengar check this... The old code did not always handle the case correctly when the size was not the default one, with this change it does (or at least it seems ;)).

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -836,98 +869,76 @@ public final SSLEngineResult unwrap(
...
+                            switch (sslError) {
+                                case SSL.SSL_ERROR_WANT_WRITE:
+                                case SSL.SSL_ERROR_WANT_READ:
+                                    // break to the outer loop as we want to read more data which means we need to write
+                                    // more to the BIO.
+                                    break readLoop;
+
+                                case SSL.SSL_ERROR_ZERO_RETURN:
+                                    // This means the connection was shutdown correctly, close inbound and outbound
+                                    if (!receivedShutdown) {
```
netkins: MAJOR Refactor this code to not nest more than 5 if/for/while/switch/try statements. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS134)

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -836,98 +869,76 @@ public final SSLEngineResult unwrap(
...
+                        if (bytesRead > 0) {
+                            bytesProduced += bytesRead;
 
-                    if (bytesRead > 0) {
-                        bytesProduced += bytesRead;
+                            if (!dst.hasRemaining()) {
+                                // Move to the next dst buffer as this one is full.
+                                dstsOffset++;
+                            } else {
+                                if (packetLength == 0) {
```
netkins: MAJOR Refactor this code to not nest more than 5 if/for/while/switch/try statements. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS134)

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -432,6 +443,29 @@ public void setRejectRemoteInitiatedRenegotiation(boolean rejectRemoteInitiatedR
     }
 
     /**
+     * Returns if remote initiated renegotiation is supported or not.
+     */
+    public boolean getRejectRemoteInitiatedRenegotiation() {
```
netkins: MAJOR A getX() method which returns a boolean should be named isX() [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3ABooleanGetMethodName)

sunng87:
@normanmaurer is it possible to specify max bio size for read/write buffer individually? In our application, the max request size is 5KB while there is no response size limit. It would be better in control if we can set different size for read/write buffer.

normanmaurer:
Not at the moment but sounds like a nice idea... let me come up with a pr for tcnative

johnou:
@normanmaurer for next 4.0, right? 🎱

normanmaurer:
@johnou Not 100 % sure yet as I want to do some more testing just to be 100 % sure all works (which seems to be the case so far)

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -132,7 +142,8 @@ protected void deallocate() {
     final Certificate[] keyCertChain;
     final ClientAuth clientAuth;
     final OpenSslEngineMap engineMap = new DefaultOpenSslEngineMap();
-    volatile boolean rejectRemoteInitiatedRenegotiation;
+    private volatile boolean rejectRemoteInitiatedRenegotiation;
+    private volatile int maxBioSize = MAX_BIO_SIZE;
```
nmittler: @normanmaurer I must be missing something ... where is this property used?

normanmaurer: here... expand the diff:
https://github.com/netty/netty/pull/6123/files#diff-cfd417915faad637e4278b54adafd9baR275

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -836,98 +869,76 @@ public final SSLEngineResult unwrap(
...
-                // everything is flushed in the BIO pair and so we can detect it in the pendingAppData() call.
-                if (SSL.readFromSSL(ssl, EMPTY_ADDR, 0) <= 0) {
-                    // We do not check SSL_get_error as we are not interested in any error that is not fatal.
-                    int err = SSL.getLastErrorNumber();
-                    if (OpenSsl.isError(err)) {
-                        return sslReadErrorResult(err, bytesConsumed, bytesProduced);
-                    }
-                }
-            }
-            if (pendingAppData() > 0) {
```
Scottmitch: can you explain why this was removed? is it no longer possible for the encryption portion to generate more data than can fit in the destination buffers? the `capacity` vs `packetLength` we check above is against the encrypted length not the decrypted length, right? Is the extra iteration over `dsts` even necessary if we can just deal with the `BUFFER_OVERFLOW` at the end instead of trying to check upfront based off of encrypted length?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -86,6 +87,15 @@ public Boolean run() {
                     return SystemPropertyUtil.getBoolean("jdk.tls.rejectClientInitiatedRenegotiation", false);
                 }
             });
+
+    private static final int MAX_BIO_SIZE =
+            AccessController.doPrivileged(new PrivilegedAction<Integer>() {
+                @Override
+                public Integer run() {
+                    return Math.min(0, SystemPropertyUtil.getInt("io.netty.handler.ssl.openssl.maxBioSize", 0));
```
normanmaurer: Yes as it uses the default 17k here then. Let me add some Javadoc explaining it and also make it clear it's about bytes

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -658,16 +641,23 @@ public final SSLEngineResult wrap(
                                 if (!receivedShutdown) {
                                     closeAll();
                                 }
-                                pendingNetResult = readPendingBytesFromBIO(dst, bytesConsumed, bytesProduced, status);
-                                return pendingNetResult != null ? pendingNetResult : CLOSED_NOT_HANDSHAKING;
+                                // Check to see if the engine wrote data into the network BIO
+                                pendingNet = SSL.pendingWrittenBytesInBIO(networkBIO);
```
normanmaurer: Yes but only if closeAll() was called, let me move it into the if block FTW

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -438,9 +437,11 @@ private int writeEncryptedData(final ByteBuffer src, int len) {
                 netWrote = SSL.writeToBIO(networkBIO, addr, len);
                 if (netWrote >= 0) {
```
normanmaurer: Yes, let me fix

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -681,25 +671,61 @@ public final SSLEngineResult wrap(
...
             }
+
             // We need to check if pendingWrittenBytesInBIO was checked yet, as we may not checked if the srcs was
             // empty, or only contained empty buffers.
-            if (bytesConsumed == 0) {
-                SSLEngineResult pendingNetResult = readPendingBytesFromBIO(dst, 0, bytesProduced, status);
-                if (pendingNetResult != null) {
-                    return pendingNetResult;
-                }
+            if (bytesConsumed == 0 && (pendingNet = SSL.pendingWrittenBytesInBIO(networkBIO)) > 0) {
```
normanmaurer: Actually we can remove this as we handle the case for handshake data above

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -681,25 +671,61 @@ public final SSLEngineResult wrap(
...
 
-            return newResult(isOutboundDone() ? CLOSED : OK, bytesConsumed, bytesProduced, status);
+    private SSLEngineResult.Status wrapStatus() {
+        // If isOutboundDone, then the data from the network BIO
+        // was the close_notify message and all was consumed we are not required to wait
+        // for the receipt the peer's close_notify message -- shutdown.
+        if (isOutboundDone()) {
+            shutdown();
+            return CLOSED;
+        } else {
```

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: Yes it's not possible to have any leftover... also we need to check upfront as the SSLEngine should not keep data in internal buffers between unwrap / wrap calls

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
Scottmitch: ok gotcha ... 0 means take OpenSSL's default.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -726,7 +745,7 @@ private SSLException shutdownWithError(String operation, String err) {
 
     public final SSLEngineResult unwrap(
             final ByteBuffer[] srcs, int srcsOffset, final int srcsLength,
-            final ByteBuffer[] dsts, final int dstsOffset, final int dstsLength) throws SSLException {
+            final ByteBuffer[] dsts, int dstsOffset, final int dstsLength) throws SSLException {
```
netkins: MAJOR The Cyclomatic Complexity of this method "unwrap" is 64 which is greater than 25 authorized. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AMethodCyclomaticComplexity)

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -836,98 +867,74 @@ public final SSLEngineResult unwrap(
...
+                            switch (sslError) {
+                                case SSL.SSL_ERROR_WANT_WRITE:
+                                case SSL.SSL_ERROR_WANT_READ:
+                                    // break to the outer loop as we want to read more data which means we need to write
+                                    // more to the BIO.
+                                    break readLoop;
+
+                                case SSL.SSL_ERROR_ZERO_RETURN:
+                                    // This means the connection was shutdown correctly, close inbound and outbound
+                                    if (!receivedShutdown) {
```
netkins: MAJOR Refactor this code to not nest more than 5 if/for/while/switch/try statements. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS134)

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -270,6 +281,9 @@ public String run() {
                 SSLContext.setOptions(ctx, SSL.SSL_OP_SINGLE_ECDH_USE);
                 SSLContext.setOptions(ctx, SSL.SSL_OP_SINGLE_DH_USE);
                 SSLContext.setOptions(ctx, SSL.SSL_OP_NO_SESSION_RESUMPTION_ON_RENEGOTIATION);
+                // We not support compression as the moment so not use it.
+                SSLContext.setOptions(ctx, SSL.SSL_OP_NO_COMPRESSION);
```
johnou: Compression was never supported? I suppose it is only safe with up to date clients anyway (CRIME).

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -270,6 +281,9 @@ public String run() {
                 SSLContext.setOptions(ctx, SSL.SSL_OP_SINGLE_ECDH_USE);
                 SSLContext.setOptions(ctx, SSL.SSL_OP_SINGLE_DH_USE);
                 SSLContext.setOptions(ctx, SSL.SSL_OP_NO_SESSION_RESUMPTION_ON_RENEGOTIATION);
+                // We not support compression as the moment so not use it.
```
johnou: We do not support compression at the moment

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
normanmaurer: yeah it was never supported... We just not disabled it explicit before which we should do anyway.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -809,6 +828,11 @@ public final SSLEngineResult unwrap(
                 throw new NotSslRecordException("not an SSL/TLS record");
             }
 
+            // IMPORTANT:
+            //
+            // This calculation only works as we not support compression and set
```
johnou: as we do not support

siyengar:
I was reading the diff, and there are several changes in here including removing compression. @normanmaurer what exactly in the previous code assumed a default buffer size?

siyengar:
@normanmaurer to clarify the previous implementation was only affected when compression was used?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -622,52 +614,50 @@ public final SSLEngineResult wrap(
...
                             case SSL.SSL_ERROR_WANT_READ:
                                 // If there is no pending data to read from BIO we should go back to event loop and try
                                 // to read more data [1]. It is also possible that event loop will detect the socket
                                 // has been closed. [1] https://www.openssl.org/docs/manmaster/ssl/SSL_write.html
-                                pendingNetResult = readPendingBytesFromBIO(dst, bytesConsumed, bytesProduced, status);
-                                return pendingNetResult != null ? pendingNetResult :
-                                        new SSLEngineResult(isOutboundDone() ? CLOSED : OK,
-                                                NEED_UNWRAP, bytesConsumed, bytesProduced);
+
+                                if (pendingNet <= 0) {
```
siyengar: in this case src could potentially have more bytes to SSL_write. This would be a bit pessimistic to return to the loop.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -270,6 +281,9 @@ public String run() {
                 SSLContext.setOptions(ctx, SSL.SSL_OP_SINGLE_ECDH_USE);
                 SSLContext.setOptions(ctx, SSL.SSL_OP_SINGLE_DH_USE);
                 SSLContext.setOptions(ctx, SSL.SSL_OP_NO_SESSION_RESUMPTION_ON_RENEGOTIATION);
+                // We do not support compression as the moment so not use it.
```
Scottmitch: `as the moment so not use it.` -> `so we should explicitly disable it.`

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -158,20 +157,16 @@
     }
 
     private static final int MAX_PLAINTEXT_LENGTH = 16 * 1024; // 2^14
-    private static final int MAX_COMPRESSED_LENGTH = MAX_PLAINTEXT_LENGTH + 1024;
-    private static final int MAX_CIPHERTEXT_LENGTH = MAX_COMPRESSED_LENGTH + 1024;
+    private static final int MAX_CIPHERTEXT_LENGTH = MAX_PLAINTEXT_LENGTH + 1024;
 
-    // Header (5) + Data (2^14) + Compression (1024) + Encryption (1024) + MAC (20) + Padding (256)
+    // Header (5) + Data (2^14) + Encryption (1024) + MAC (20) + Padding (256)
+    // No compression supported!
```
Scottmitch: Compression is not supported!

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -809,6 +828,11 @@ public final SSLEngineResult unwrap(
                 throw new NotSslRecordException("not an SSL/TLS record");
             }
 
+            // IMPORTANT:
+            //
+            // This calculation only works as we do not support compression and set
```
Scottmitch: This calculation only works as we do not support compression. We explicitly disable compression by setting the SSL_OP_NO_COMPRESSION option. If compression were enabled it is possible that the plain text length is > the packet length.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -809,6 +828,11 @@ public final SSLEngineResult unwrap(
                 throw new NotSslRecordException("not an SSL/TLS record");
             }
 
+            // IMPORTANT:
```
Scottmitch: consider removing this comment

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
Scottmitch: now that we explicitly disable compression ... the length check up front should be good enough. I wonder if we can improve the length check to not always just require max length (not sure if that would be beneficial or not).

normanmaurer: Let us consider this as a follow up.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: @Scottmitch btw do you think I should maybe factor out the disabling of compression explicitly in another PR just to make it more "clear" ?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: but its important ;)

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -836,98 +864,74 @@ public final SSLEngineResult unwrap(
...
-                    int written = writeEncryptedData(src, Math.min(packetLengthRemaining, src.remaining()));
+                    int written = writeEncryptedData(src, Math.min(packetLength, remaining));
                     if (written > 0) {
-                        packetLengthRemaining -= written;
-                        if (packetLengthRemaining == 0) {
-                            // A whole packet has been consumed.
-                            break;
-                        }
+                        packetLength -= written;
+                        bytesConsumed += written;
```
normanmaurer: I think the current code is a bit cleaner and easier to understand. I can check if it makes any difference tho if you feel strong about it

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
siyengar: its not necessary, however maybe a check like this may help

```
if (pendingNet <= 0 && !src.hasRemaining()) {...
```

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
Scottmitch: This is not necessary ... we can consider this as a followup PR if necessary.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
Scottmitch: @normanmaurer - it may be *cleaner* this way but it doesn't matter to me ... as long as it is disabled I'm happy.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
Scottmitch: OpenSSL is saying it needs some data to be read/unwrapped in this condition. I wouldn't expect this condition to change even if `src` (or other elements in the `srcs` array) have more data.

Can you clarify under what condition you would expect subsequent writes to succeed without reading/unwrapping data? Also we just tried to write `src` and failed ... can you explain why you expect different results if we tried to write `src` again (assuming there is data left to write)?

siyengar: Ah nvm, I misunderstood. It's an SSL_write which is returning a WANT_READ. Ya in this case the logic is correct.

johnou:
@normanmaurer how can I verify whether or not our game client has negotiated compression with a Netty SSL server?

johnou:
@normanmaurer how can I verify whether or not our game client has negotiated compression with a Netty SSL server? We are actively trying to minimize network traffic as players are complaining about data usage, how much extra work would it be to support turning compression on and off as a configuration option?

johnou:
@Scottmitch better to get this out sooner than later as this mitigates CRIME security exploit for TLS 1.0 clients.

normanmaurer:
@johnou I don't think there is an easy way beside checking with wireshark. Let me factor out the "explicit" disabling of compression into another PR.

I am not sure how much extra work is needed. Would need to check in detail, but honestly its not a super high priority atm :(

johnou:
@normanmaurer we had a quick tech meeting and decided compression is not suitable for our application (most of the messages are small and it's a time sensitive protocol so the overhead is simply not worth it).

That said I am keen to know if our client is trying to use compression because I have observed some strange errors related to parsing. I will try debugging the handshake and report back here.

johnou:
@normanmaurer looks as though compression is not used in our application afterall so we welcome this change 👍 

Client hello:
client_hello

Server hello:
server_hello

Scottmitch:
> better to get this out sooner than later as this mitigates CRIME security exploit for TLS 1.0 clients.

@johnou - Compression can be disabled via other means (it can be disabled at configure/compile time for OpenSSL, the user can explicitly set this option, etc..). This PR is not necessary to mitigate CRIME.

Scottmitch:
> how can I verify whether or not our game client has negotiated compression with a Netty SSL server?

@johnou - I'm not sure what OpenSSL library you are using or how its built, but assuming you are building it your self just disable compression at configure/compile time via `no-comp` [1].

If your OpenSSL library does not have compression explicitly disabled clients may still be able to negotiate compression. ssllabs can test your server for compression support [2].

[1] https://wiki.openssl.org/index.php/Compilation_and_Installation
[2] https://www.ssllabs.com/ssltest/index.html

johnou:
> @johnou - Compression can be disabled via other means (it can be disabled at configure/compile time for OpenSSL). This PR is not necessary to mitigate CRIME.

But still helpful.

> @johnou - I'm not sure what OpenSSL library you are using or how its built, but assuming you are building it your self just disable compression at configure/compile time via no-comp [1].

Server side we are using dynamic OpenSSL artifacte with stock OS openssl package.

Client side is a little more out of our hands, we are using C# SslStream but with the ancient Mono shipped with Unity 👎 

> If your OpenSSL library does not have compression explicitly disabled clients may still be able to negotiate compression. ssllabs can test your server for compression support [2].

"Ports other than 443 not supported", I will look around for other services to probe our application.

Scottmitch:
> Server side we are using dynamic OpenSSL artifacte with stock OS openssl package.
> Client side is a little more out of our hands

Have you considered building your own. Depending on your OS/distribution the OpenSSL package may be quite old, and per this discussion you don't have control over what options are enabled.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: @siyengar thanks for the discussion tho :) This stuff is always tricky ...

normanmaurer:
@Scottmitch addressed everything... And yes I agree we should let this cook a bit before include in a release. Too risky otherwise ;)

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -881,98 +915,74 @@ public final SSLEngineResult unwrap(
                         srcsOffset++;
                         continue;
                     }
+
                     // Write more encrypted data into the BIO. Ensure we only read one packet at a time as
                     // stated in the SSLEngine javadocs.
-                    int written = writeEncryptedData(src, min(packetLengthRemaining, src.remaining()));
+                    int written = writeEncryptedData(src, Math.min(packetLength, remaining));
```
netkins: MAJOR Unnecessary use of fully qualified name 'Math.min' due to existing static import 'java.lang.Math.min' [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUnnecessaryFullyQualifiedName)

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -86,6 +87,15 @@ public Boolean run() {
                     return SystemPropertyUtil.getBoolean("jdk.tls.rejectClientInitiatedRenegotiation", false);
                 }
             });
+
+    private static final int MAX_BIO_BYTES =
+            AccessController.doPrivileged(new PrivilegedAction<Integer>() {
+                @Override
+                public Integer run() {
+                    return Math.min(0, SystemPropertyUtil.getInt("io.netty.handler.ssl.openssl.maxBioBytes", 0));
```
normanmaurer: Doh yes! Fixing

normanmaurer:
@Scottmitch FYI there is something wrong here... still investigate

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -881,98 +916,74 @@ public final SSLEngineResult unwrap(
...
+                            switch (sslError) {
+                                case SSL.SSL_ERROR_WANT_WRITE:
+                                case SSL.SSL_ERROR_WANT_READ:
+                                    // break to the outer loop as we want to read more data which means we need to write
+                                    // more to the BIO.
+                                    break readLoop;
+
+                                case SSL.SSL_ERROR_ZERO_RETURN:
+                                    // This means the connection was shutdown correctly, close inbound and outbound
+                                    if (!receivedShutdown) {
```
netkins: MAJOR Refactor this code to not nest more than 5 if/for/while/switch/try statements. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS134)

Scottmitch:
I propose we close this in favor of https://github.com/netty/netty/pull/6201

normanmaurer:
@Scottmitch sounds good... closing now

Issue #3548, opened 2015-03-28
Cannot subclass HttpHeaders and build with JDK 7 with maven build of 4.0.26.Final


timboudreau:
It seems the build of Netty 4.0.26.Final from Maven is built with JDK 8, which adds default methods to collections classes that reference classes that don't exist in JDK 7.  That's binary compatible, but not source compatible, as follows:

I am stuck with JDK 7 for a while longer (thanks @schue ) for Actuer.  As of Netty 4.0.26.Final my [build is broken](http://timboudreau.com/builds/job/mastfrog-parent/703/console) because javac sees `spliterator` methods and classes in `HttpHeaders`.  To fix it I'm staying with Netty 4.0.25 for now for Acteur, but that's not a long-term solution.

Since AFAIK, Netty 4 is buildable on JDK 7, is it possible for JARs released to Maven Central also to be built with JDK 7?

I imagine I am not the only one in this situation.

Scottmitch:
@timboudreau - Thanks for reporting!

@normanmaurer - Building the releases from jdk7 sounds reasonable...WDYT?

trustin:
Sounds good to me. See the `restricted-release` profile to enforce JDK version when a release is made.

trustin:
Should be fixed on 0222f412336f87909ff71494e5af8e2b221945da (4.0) and bdf0bddc8544666ce73b0e7e486ddccf214e79ef (4.1)

trustin:
Should be fixed on 0222f412336f87909ff71494e5af8e2b221945da (4.0) and bdf0bddc8544666ce73b0e7e486ddccf214e79ef (4.1)

normanmaurer:
@timboudreau @trustin I will revert this one as it seems like it was fixed in java some time ago

PR #6124, opened 2016-12-09
Compile with java8 and so remove reflection hacks.

normanmaurer:
Motivation:

We previously changed netty to always compile with java7 as otherwise source compatibility was broken.

This was reported in [#3548] but was fixed in the meantime:
- https://bugs.openjdk.java.net/browse/JDK-8029240
- https://bugs.openjdk.java.net/browse/JDK-8030855

Modifications:

Use again java8 and so remove all the reflection hacks.

Result:

Easier to maintain code.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -32,15 +32,12 @@ import io.netty.util.internal.ThrowableUtil;
 import io.netty.util.internal.logging.InternalLogger;
 import io.netty.util.internal.logging.InternalLoggerFactory;
 
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
 import java.nio.ByteBuffer;
 import java.nio.ReadOnlyBufferException;
 import java.security.Principal;
 import java.security.cert.Certificate;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


normanmaurer:
@Scottmitch @nmittler PTAL ...

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1614,40 +1546,17 @@ public final synchronized void setSSLParameters(SSLParameters sslParameters) {
...
-                            for (Object serverName : servernames) {
-                                if (SNI_HOSTNAME_CLASS.isInstance(serverName)) {
-                                    SSL.setTlsExtHostName(ssl, (String) GET_ASCII_NAME_METHOD.invoke(serverName));
-                                } else {
-                                    throw new IllegalArgumentException("Only " + SNI_HOSTNAME_CLASS.getName()
-                                            + " instances are supported, but found: " +
-                                            serverName);
-                                }
-                            }
+                if (!isDestroyed()) {
```
netkins: MAJOR Merge this if statement with the enclosing one. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1066)

Scottmitch:
`Compile with java8 and so remove reflection hacks.` -> `Compile with java8 and remove reflection hacks`

Scottmitch:
@normanmaurer - also the PR template was still present when you opened the PR ... I removed it though ... so just for future reference.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1614,40 +1546,17 @@ public final synchronized void setSSLParameters(SSLParameters sslParameters) {
...
+                    if (clientMode) {
+                        List<String> servernames = Java8SslParametersUtils.getSniHostNames(sslParameters);
```
Scottmitch: `sniHostNames` is no longer assigned ... should we assign it here, or after?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
Scottmitch: +1 ... promote this `if` up into the previous one with a `&&`?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: fixed

normanmaurer:
@Scottmitch PTAL again

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1579,26 +1523,14 @@ public final synchronized SSLParameters getSSLParameters() {
...
-                    try {
-                        SET_USE_CIPHER_SUITES_ORDER_METHOD.invoke(sslParameters,
-                                (SSL.getOptions(ssl) & SSL.SSL_OP_CIPHER_SERVER_PREFERENCE) != 0);
-                    } catch (IllegalAccessException e) {
-                        throw new Error(e);
-                    } catch (InvocationTargetException e) {
-                        throw new Error(e);
-                    }
+                if (!isDestroyed()) {
+                    Java8SslParametersUtils.setUseCipherSuitesOrder(
```
nmittler: Should/could this just be part of the platform class?

normanmaurer: @nmittler maybe... I think its fine to keep it here for now tho

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1613,42 +1545,18 @@ public final synchronized void setSSLParameters(SSLParameters sslParameters) {
...
-                            }
-                        }
-                        sniHostNames = servernames;
-                    } catch (IllegalAccessException e) {
-                        throw new Error(e);
-                    } catch (InvocationTargetException e) {
-                        throw new Error(e);
+            if (version >= 8 && !isDestroyed()) {
+                if (clientMode) {
+                    List<String> servernames = Java8SslParametersUtils.getSniHostNames(sslParameters);
```
Scottmitch: is the temp variable necessary here? can we just do:

```java
sniHostNames = Java8SslParametersUtils.getSniHostNames(sslParameters);
for (String name: sniHostNames) {
  SSL.setTlsExtHostName(ssl, name);
}
```

normanmaurer: sure...

normanmaurer:
@Scottmitch comment addressed... all good ? 

That said I want to pull this in after the next release.

normanmaurer:
@trustin WDYT ?

normanmaurer:
@trustin the downside is that with this we may not be able to run our test suite with java7 anymore. Not sure its a big deal tho.

normanmaurer:
Cherry-picked into 4.1 (591293bfb4f6a48a311d195303ce772fb801ec95) and 4.0 (ed27831)

PR #6130, opened 2016-12-13
Remove usage of own Atomic*FieldUpdater in favor of JDKs

normanmaurer:
Motivation:

In later Java8 versions our Atomic*FieldUpdater are slower then the JDK implementations so we should not use ours anymore. Even worse the JDK implementations provide for example an optimized version of addAndGet(...) using intrinsics which makes it a lot faster for this use-case.

Modifications:

- Remove methods that return our own Atomic*FieldUpdaters.
- Use the JDK implementations everywhere.

Result:

Faster code.


normanmaurer:
@Scottmitch @nmittler PTAL.

I did some benchmarks for example:

```
Benchmark                                             Mode  Cnt         Score        Error  Units
AtomicUpdaterBenchmark.customAtomicUpdaterBenchmark  thrpt   20  11817692.180 ± 510071.352  ops/s
AtomicUpdaterBenchmark.jdkAtomicUpdaterBenchmark     thrpt   20  42355102.913 ± 859397.660  ops/s
```

@nitsanw made me aware of it... So 🍻 for him

nmittler:
@normanmaurer LGTM ... great perf! woot!

normanmaurer:
Cherry-picked into 4.0 (9b9666a3fecd6b6f9cee5d31b71d3bad27f94921) and 4.1 (89e93968ac601cbe3d6ca105abbefc99adcb56be)

Issue #6096, opened 2016-12-02
Netty should stop swallowing


jasontedor:
I've raised this issue [before](https://github.com/netty/netty/pull/5552/files/03f70037c0652e8a24ec2edb721ea4d479ba8a6b#r71620816) and it has come up [subsequently](https://github.com/netty/netty/pull/6047/files#r88775483), but I think a serious conversation needs to be had about it.

The Netty code base is filled with catch throwable blocks and far too many of these blocks swallow the caught throwable. I think these blocks are almost always dangerous. It means catching and swallowing [`Error`s](https://docs.oracle.com/javase/8/docs/api/java/lang/Error.html) from which there is likely no reasonable chance of recovery. It means hiding bugs. It means hiding fatal errors from the application layer. Here's a stack trace we saw recently on a production cluster:

```
java.lang.OutOfMemoryError: Direct buffer memory
  at java.nio.Bits.reserveMemory(Bits.java:658)
  at java.nio.DirectByteBuffer.<init>(DirectByteBuffer.java:123)
  at java.nio.ByteBuffer.allocateDirect(ByteBuffer.java:311)
  at org.jboss.netty.channel.socket.nio.SocketReceiveBufferAllocator.newBuffer(SocketReceiveBufferAllocator.java:64)
  at org.jboss.netty.channel.socket.nio.SocketReceiveBufferAllocator.get(SocketReceiveBufferAllocator.java:41)
  at org.jboss.netty.channel.socket.nio.NioWorker.read(NioWorker.java:62)
  at org.jboss.netty.channel.socket.nio.AbstractNioWorker.process(AbstractNioWorker.java:108)
  at org.jboss.netty.channel.socket.nio.AbstractNioSelector.run(AbstractNioSelector.java:337)
  at org.jboss.netty.channel.socket.nio.AbstractNioWorker.run(AbstractNioWorker.java:89)
  at org.jboss.netty.channel.socket.nio.NioWorker.run(NioWorker.java:178)
  at org.jboss.netty.util.ThreadRenamingRunnable.run(ThreadRenamingRunnable.java:108)
  at org.jboss.netty.util.internal.DeadLockProofWorker$1.run(DeadLockProofWorker.java:42)
  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
[...]

PR #6135, opened 2016-12-15
Allow configuration of Logger via System property.

rkapsi:
Allow configuration of Logger via System property

Motivation

Even from the main() method it's sometimes difficult to control the exact initialization order of classes and matters get worse if a user doesn't necessarily have access to the main() method. There was a recent commit (hash below) which broke our application because something eagerly initialized the InternalLoggerFactory with SLF4J and we want to use Log4j 2.

This brings us to an another problem. Netty's default behavior is to use SLF4J and the mere presence of it in the classpath will make Netty choose it. In complex applications it's impossible to avoid to have various bindings for different logging frameworks in the classpath.

Breaking commit: 3c92f2b64a1c6803860ab310348abc9f1cb47ff4

Modifications

Adding a system property which allows the user to specify the desired Logger.

Result

It's possible to configure the Logger via a System property such as:

-Dio.netty.logger.type=io.netty.util.internal.logging.Log4J2LoggerFactory

### DIFF
--- a/common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
+++ b/common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
@@ -16,7 +16,10 @@
 
 package io.netty.util.internal.logging;
 
+import java.lang.reflect.Field;
 import java.util.concurrent.atomic.AtomicReference;
+
+import io.netty.util.internal.SystemPropertyUtil;
 
 /**
  * Creates {@link InternalLogger}s. This factory allows you to choose what logging framework Netty should use. The
@@ -77,7 +80,22 @@ public abstract class InternalLoggerFactory {
 
     @SuppressWarnings("UnusedCatchParameter")
     private static InternalLoggerFactory newDefaultFactory(String name) {
-        InternalLoggerFactory f;
+        InternalLoggerFactory f = null;
+
+        String loggerType = SystemPropertyUtil.get("io.netty.logger.type");
+        if (loggerType != null) {
+            try {
+                Class<?> clazz = Class.forName(loggerType);
+                Field field = clazz.getField("INSTANCE");
+                f = (InternalLoggerFactory) field.get(null);
+                f.newInstance(name).debug("Using {} as the default logging framework", loggerType);
+            } catch (Throwable t) {
+                throw new RuntimeException("loggerType=" + loggerType, t);
+            }
+
+            return f;
+        }
+
         try {
             f = new Slf4JLoggerFactory(true);
             f.newInstance(name).debug("Using SLF4J as the default logging framework");
@@ -90,6 +108,7 @@ public abstract class InternalLoggerFactory {
                 f.newInstance(name).debug("Using java.util.logging as the default logging framework");
             }
         }
+
         return f;
     }
 


normanmaurer:
@jasontedor @Scottmitch @johnou PTAL

jasontedor:
> There was a recent commit (hash below) which broke our application because something eagerly initialized the InternalLoggerFactory with SLF4J and we want to use Log4j 2.

That was the point of #6106; it's a bug that the logger factory could be initialized twice and you'd end up using SLF4J in some places, and Log4J2 in others (we had a similar problem in our application: we wanted our own logger factory everywhere, but Netty *itself* (see #6069) was eagerly initializing and using Log4j 1 in some places (because we have the Log4j 1 bridge on our classpath to as backwards compatibility for plugins that are still using the Log4j 1 API). So I would not say it so much as broke your application as helped you fix a broken situation. 😇

> Adding a system property which allows the user to specify the desired Logger.

Sorry, I'm confused. Why not:

```java
class Main {

    static {
        InternalLoggerFactory.setLoggerFactory(Log4J2LoggerFactory.INSTANCE);
    }

    }

}
```

So I don't yet see the need to add a system property and more complexity to this situation?

common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
```diff
@@ -77,7 +80,22 @@ InternalLogger newInstance(String name) {
...
+        InternalLoggerFactory f = null;
+
+        String loggerType = SystemPropertyUtil.get("io.netty.logger.type");
+        if (loggerType != null) {
+            try {
+                Class<?> clazz = Class.forName(loggerType);
+                Field field = clazz.getField("INSTANCE");
+                f = (InternalLoggerFactory) field.get(null);
+                f.newInstance(name).debug("Using {} as the default logging framework", loggerType);
+            } catch (Throwable t) {
```
jasontedor: Can we please not catch throwable here? Yes, it's done elsewhere in this class but I think those should be removed as part of #6096.

common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java
```diff
@@ -77,7 +80,22 @@ InternalLogger newInstance(String name) {
...
-        InternalLoggerFactory f;
+        InternalLoggerFactory f = null;
+
+        String loggerType = SystemPropertyUtil.get("io.netty.logger.type");
+        if (loggerType != null) {
+            try {
+                Class<?> clazz = Class.forName(loggerType);
+                Field field = clazz.getField("INSTANCE");
+                f = (InternalLoggerFactory) field.get(null);
+                f.newInstance(name).debug("Using {} as the default logging framework", loggerType);
```
jasontedor: This is making an assumption about the structure of the logger factory that might not be correct. If this is indeed accepted, it should at least be documented what the structure of an `InternalLoggerFactory` that can be set via a system property must be.

rkapsi:
@jasontedor your Main class example looks exactly how our code looks like. Yet... It blew up. Twice. I can only speculate but...

Our production application is running inside a Dropwizard container. So the main class extends some Dropwizard class. And that super class does some stuff which I don't control.

On the side I'm working on porting our application form Dropwizard to Spring Boot. Same idea, you have your main() method but there is all that other stuff that may touch your classes due to classpath scanning.

Lastly and as I said. Not all applications have a main() method available to them. Not in my case but maybe others (e.g. OSGI containers and what not).

I agree, the breaking change is good but I disagree with the current state of brokenness.

rkapsi:
OK, decided to dig and see who's calling the InternalLoggerFactory in the first place (see Thread dump below).

It's Spring Boot -> Log4j2 -> loading its config (we use Clojure for that) which touches one of our classes which in turn touches a Netty class which in turn needs a Logger and ends up initializing the InternalLoggerFactory.

In our attempt to get our logging configuration in order and do it eagerly before anything else is up we end up shooting ourslves in the foot. But more importantly, at this point in our application's lifecycle it'd be still safe to change the InternalLoggerFactory but now we can't and we're stuck with this damn SLF4J binding (I'm really not happy about that).

```java
java.lang.IllegalStateException: factory is already set to [io.netty.util.internal.logging.Slf4JLoggerFactory@16aed49f], rejecting [io.netty.util.internal.logging.Log4J2LoggerFactory@66de00f2]
```

Again, I think the braking change is good but it's also opinionated. You basically can't know how the user's application's lifecycle looks like. The only safe option is IMHO to give the user the opportunity to state upfront what their intention is and the easiest, safest and most straight forward thing seems to be System property.


```java
java.lang.Exception: Stack trace
	at java.lang.Thread.dumpStack(Thread.java:1333)
	at io.netty.util.internal.logging.InternalLoggerFactory.newDefaultFactory(InternalLoggerFactory.java:83)
	at io.netty.util.internal.logging.InternalLoggerFactory.access$000(InternalLoggerFactory.java:32)
	at io.netty.util.internal.logging.InternalLoggerFactory$InternalLoggerFactoryHolder.getFactory(InternalLoggerFactory.java:53)
	at io.netty.util.internal.logging.InternalLoggerFactory$InternalLoggerFactoryHolder.newInstance(InternalLoggerFactory.java:77)
	at io.netty.util.internal.logging.InternalLoggerFactory$InternalLoggerFactoryHolder.getInstance(InternalLoggerFactory.java:73)
	at io.netty.util.internal.logging.InternalLoggerFactory$InternalLoggerFactoryHolder.getInstance(InternalLoggerFactory.java:69)
	at io.netty.util.internal.logging.InternalLoggerFactory.getInstance(InternalLoggerFactory.java:138)
	at io.netty.util.internal.PlatformDependent.<clinit>(PlatformDependent.java:76)
	at io.netty.util.ConstantPool.<init>(ConstantPool.java:32)
	at io.netty.util.AttributeKey$1.<init>(AttributeKey.java:27)
	at io.netty.util.AttributeKey.<clinit>(AttributeKey.java:27)
	at com.squarespace.echo.aaa.Aaa.<clinit>(Aaa.java:23)
	at com.squarespace.echo.bbb.Bbb.<clinit>(Bbb.java:15)
	at com.squarespace.echo.ccc.log4j$eval167.invokeStatic(<path>/log4j.clj:104)
	at com.squarespace.echo.ccc.log4j$eval167.invoke(<path>/log4j.clj:9)
	at clojure.lang.Compiler.eval(Compiler.java:6927)
	at clojure.lang.Compiler.load(Compiler.java:7379)
	at com.squarespace.log4j.clojure.ClojureConfigurationFactory.getConfiguration(ClojureConfigurationFactory.java:72)
	at org.apache.logging.log4j.core.config.ConfigurationFactory$Factory.getConfiguration(ConfigurationFactory.java:540)
	at org.springframework.boot.logging.log4j2.Log4J2LoggingSystem.loadConfiguration(Log4J2LoggingSystem.java:176)
	at org.springframework.boot.logging.log4j2.Log4J2LoggingSystem.loadConfiguration(Log4J2LoggingSystem.java:167)
	at org.springframework.boot.logging.AbstractLoggingSystem.initializeWithSpecificConfig(AbstractLoggingSystem.java:66)
	at org.springframework.boot.logging.AbstractLoggingSystem.initialize(AbstractLoggingSystem.java:56)
	at org.springframework.boot.logging.log4j2.Log4J2LoggingSystem.initialize(Log4J2LoggingSystem.java:148)
	at org.springframework.boot.logging.LoggingApplicationListener.initializeSystem(LoggingApplicationListener.java:306)
	at org.springframework.boot.logging.LoggingApplicationListener.initialize(LoggingApplicationListener.java:274)
	at org.springframework.boot.logging.LoggingApplicationListener.onApplicationEnvironmentPreparedEvent(LoggingApplicationListener.java:237)
	at org.springframework.boot.logging.LoggingApplicationListener.onApplicationEvent(LoggingApplicationListener.java:210)
	at org.springframework.context.event.SimpleApplicationEventMulticaster.invokeListener(SimpleApplicationEventMulticaster.java:166)
	at org.springframework.context.event.SimpleApplicationEventMulticaster.multicastEvent(SimpleApplicationEventMulticaster.java:138)
	at org.springframework.context.event.SimpleApplicationEventMulticaster.multicastEvent(SimpleApplicationEventMulticaster.java:121)
	at org.springframework.boot.context.event.EventPublishingRunListener.environmentPrepared(EventPublishingRunListener.java:69)
	at org.springframework.boot.SpringApplicationRunListeners.environmentPrepared(SpringApplicationRunListeners.java:54)
	at org.springframework.boot.SpringApplication.prepareEnvironment(SpringApplication.java:335)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:306)
	at com.squarespace.echo.server.Echo.main(Echo.java:45)
```

jasontedor:
Yeah, ours is more similar to yours than what I outlined in my sketch (I was trying to keep it simple). Ours is:

```
 - io.netty.util.internal.logging.InternalLoggerFactory.setDefaultFactory(io.netty.util.internal.logging.InternalLoggerFactory) @bci=0, line=80 (Interpreted frame)
 - org.elasticsearch.transport.netty4.Netty4Utils.<clinit>() @bci=7, line=44 (Interpreted frame)
 - org.elasticsearch.transport.Netty4Plugin.<clinit>() @bci=0, line=46 (Interpreted frame)
 - sun.reflect.NativeConstructorAccessorImpl.newInstance0(java.lang.reflect.Constructor, java.lang.Object[]) @bci=0 (Compiled frame)
 - sun.reflect.NativeConstructorAccessorImpl.newInstance(java.lang.Object[]) @bci=85, line=62 (Compiled frame [deoptimized])
 - sun.reflect.DelegatingConstructorAccessorImpl.newInstance(java.lang.Object[]) @bci=5, line=45 (Compiled frame [deoptimized])
 - java.lang.reflect.Constructor.newInstance(java.lang.Object[]) @bci=79, line=423 (Compiled frame [deoptimized])
 - org.elasticsearch.plugins.PluginsService.loadPlugin(java.lang.Class, org.elasticsearch.common.settings.Settings) @bci=41, line=373 (Interpreted frame)
 - org.elasticsearch.plugins.PluginsService.loadBundles(java.util.List) @bci=219, line=333 (Interpreted frame)
 - org.elasticsearch.plugins.PluginsService.<init>(org.elasticsearch.common.settings.Settings, java.nio.file.Path, java.nio.file.Path, java.util.Collection) @bci=165, line=115 (Interpreted frame)
 - org.elasticsearch.node.Node.<init>(org.elasticsearch.env.Environment, java.util.Collection) @bci=409, line=288 (Interpreted frame)
 - org.elasticsearch.node.Node.<init>(org.elasticsearch.env.Environment) @bci=5, line=229 (Interpreted frame)
 - org.elasticsearch.bootstrap.Bootstrap$6.<init>(org.elasticsearch.bootstrap.Bootstrap, org.elasticsearch.env.Environment) @bci=7, line=215 (Interpreted frame)
 - org.elasticsearch.bootstrap.Bootstrap.setup(boolean, org.elasticsearch.env.Environment) @bci=166, line=215 (Interpreted frame)
 - org.elasticsearch.bootstrap.Bootstrap.init(boolean, java.nio.file.Path, boolean, java.util.Map) @bci=155, line=307 (Interpreted frame)
 - org.elasticsearch.bootstrap.Elasticsearch.init(boolean, java.nio.file.Path, boolean, java.util.Map) @bci=13, line=121 (Interpreted frame)
 - org.elasticsearch.bootstrap.Elasticsearch.execute(org.elasticsearch.cli.Terminal, joptsimple.OptionSet, java.util.Map) @bci=200, line=112 (Interpreted frame)
 - org.elasticsearch.cli.SettingCommand.execute(org.elasticsearch.cli.Terminal, joptsimple.OptionSet) @bci=152, line=54 (Interpreted frame)
 - org.elasticsearch.cli.Command.mainWithoutErrorHandling(java.lang.String[], org.elasticsearch.cli.Terminal) @bci=78, line=122 (Interpreted frame)
 - org.elasticsearch.cli.Command.main(java.lang.String[], org.elasticsearch.cli.Terminal) @bci=78, line=88 (Interpreted frame)
 - org.elasticsearch.bootstrap.Elasticsearch.main(java.lang.String[], org.elasticsearch.bootstrap.Elasticsearch, org.elasticsearch.cli.Terminal) @bci=3, line=89 (Interpreted frame)
 - org.elasticsearch.bootstrap.Elasticsearch.main(java.lang.String[]) @bci=23, line=82 (Interpreted frame)
```

The point is to initialize the logger factory at your entry point into Netty (that has to happen somewhere, you *definitely* have control over it).

```java
	at com.squarespace.echo.aaa.Aaa.<clinit>(Aaa.java:23)
```

If that's your entry point into Netty, isn't that where you should set the logger factory?

rkapsi:
To give you a more complete example.

```java
;; Log4j2 is loading my Clojure based configuration file. Notice that
;; `com.squarespace.echo.bbb.Bbb` is referring to the Class and the
;; ClassLoader will touch it.
;;
;;
;; log4j.clj:104 - from the above Thread dump
(config
  (logger com.squarespace.echo.bbb.Bbb Level/INFO))
```

```java
// We tried to configure the logging level of Bbb but it's also touching Aaa
class Bbb {
  private static final Logger LOG = LogManager.getLogger(Bbb.class);

  private static final Aaa AAA = new Aaa();
}

// This in turn touches AttributeKey which will eventually touch InternalLoggerFactory. Damn singletons.
class Aaa {
  private static final AttributeKey<T> ATTRIBUTE_KEY = AttributeKey.valueOf(Aaa.class, "ATTRIBUTE_KEY");

  void biz(Channel channel) {
    // uses ATTRIBUTE_KEY on Channel
  }
}
```

This is stuff is not an entry point to our Netty code. It's just random business code that happens to be referred in the log4j's configuration file and all I'm trying to do is to configure logging levels. The application's real initialization including doing things such as `new ServerBootstap().bind()` hasn't even happened yet...

Yes, if the CAS check wasn't there our Netty code would use along that codepath SLF4J and the change exposes the problem in our application as well. So that's good!

The "we're not changing anything" fix is to move the `InternalLoggerFactory.setLoggerFactory(Log4J2LoggerFactory.INSTANCE)` call before log4j's configuration stage. But that seems strange to me as well. It's arbitary to favor Netty's initialization over some other component and assumes it can be done in code and in a certain order.

The only agnostic/non-biased way to do it safely and 100% reliably is to configure it externally. A system property seems straight forward to me and aligns "nicely" with all the other properties that are required in a bigger application.

```java
java -Dlog4j.configurationFile=<path> \
  -Djava.util.logging.manager=org.apache.logging.log4j.jul.LogManager \
  -Dio.netty.logger.type=io.netty.util.internal.logging.Log4J2LoggerFactory
  ...
```

To summarize, I think 3c92f2b64a1c6803860ab310348abc9f1cb47ff4 is a good change but the underlying problem remains. Netty picks a Logger arbitarily simply because the ClassLoader is touching some Netty class and the user failed to land a `setLoggerFactory(...)` call prior to that. Worse, it's actually blowing up in the user's face and it doesn't really tell how it got into that state in the first place (I had to modify Netty to get that stack trace).

jasontedor:
Why is your logging configuration triggering class loads anyway (are you reflectively setting the logging level on the `LOG` field)? That seems like something you would want to avoid, no matter what the outcome of this discussion is (and to be clear, I'm interested in continuing it, I just think you might have other undesired initialization going on in your code and you might end up with a cleaner, and clearer initialization sequence).

johnou:
The fact the class loader is touching classes should not initialise the logging factory, it is using the static holder pattern which should be lazy.

rkapsi:
No reflection. We use Clojure for configuration because it compiles to bytecode. I happy to discuss programatic configuration but I feel this is not relevant to the underlying problem. It's just exposing it.

The ClassLoader touches our `Bbb` class, which touches our `Aaa` class which has a static field of type `AttributeKey` which uses internally a static field of type `ConstantPool` which uses `PlatformDependent` and that uses `InternalLoggerFactory`. Because no default is configured yet it's picking SLF4J. Game over because I couldn't sneak my `setLoggerFactory(...)` in there.

jasontedor:
> No reflection. We use Clojure for configuration because it compiles to bytecode.

Oh, I see, it's because you're loading the class through the configuration that it's triggering the class load. That strikes me as awful, I really don't think you want your configuration code doing that.

> I happy to discuss programatic configuration but I feel this is not relevant to the underlying problem. It's just exposing it.

Sorry, I think it's relevant. Like, if you're using a pattern that is doing something like this (configuration triggering class loads), why should Netty add a system property and complexity to support it?

jasontedor:
> The fact the class loader is touching classes should not initialise the logging factory, it is using the static holder pattern which should be lazy.

@johnou It's because his class is touching a Netty classes that tickles the InternalLoggerFactory forcing the default logger factory to initialize

rkapsi:
> Sorry, I think it's relevant. Like, if you're using a pattern that is doing something like this (logging configuration triggering class loads), why should Netty add a system property and complexity to support it?

I'm a bit at loss here. Isn't `InternalLoggerFactory` doing exactly what you said above? It's using code for configuration, triggers class loads, and forces the consumer to follow a certain initialization pattern. 

Why should Netty support it? I think Netty should get rid of the current implementation. Mutable singletons are among the worst patterns to maintain and reason about. Netty should strive for simplicity and immutability whenever possible. System properties (with sensible defaults such as SLF4J) are dead simple, well understood by Java developers, and if we get rid of the current code we can make it immutable. It'll be reliable, without side effects, agnostic in regards to the consumer's use-case and environment. It'll be rock solid like the rest of the Netty codebase.

jasontedor:
> I'm a bit at loss here. Isn't `InternalLoggerFactory` doing exactly what you said above? It's using code for configuration, triggers class loads, and forces the consumer to follow a certain initialization pattern.

I do not see it as this same (code for configuration), but to be clear: I think what `InternalLoggerFactory` is doing in this respect (the classpath fallback game) is an abomination.

> I think Netty should get rid of the current implementation.

We are in complete agreement here.

> Mutable singletons are among the worst patterns to maintain and reason about.

+1

> Netty should strive for simplicity and immutability whenever possible.

+1

>  System properties (with sensible defaults such as SLF4J) are dead simple, well understood by Java developers, and if we get rid of the current code we can make it immutable.

I have a strong distaste for system properties (they are basically global variables). Additionally, I still have the concern about the assumption that you're making in your PR about how the `InternalLoggerFactory` can be reflectively constructed.

johnou:
@rkapsi if you use another dependency in your application which only uses the slf4j logging facade what are you doing? surely you have log4j-slf4j-impl on your classpath too which will anyway route all slf4j logging to log4j2?

rkapsi:
@jasontedor if we're agreement then what's your proposal? I think `InternalLoggerFactory` is conceptually broken and I don't think Netty as a library should judge how a user configures their environment and more importantly not engage in and depend on class insitialization shanahanigans.

Even if you disagree with my approach for configuration, and I feel I've made a mistake in trying to provide context in that regard, there's still nothing that shows that it's the user's (my) fault and I don't think it's breaking Netty even if I do so. A user should be able to "touch" classes such as `AttributeKey` all they want without experiencing side effects such as `InternalLoggerFactory` starting to fail on them at a later point in time due to unexpected side effects.

My PR is trying to give the user the option to use an immutable value using the little bit of "API" that there is without breaking backwards compatibility or force consumers to use it. I have no problem with global variables as long as they're immutable (or there's a gentleman's agreement to consider them as such) or they're otherwise safe.

@johnou yes of course. But if Netty jumps through all the hoops and efforts to be logging framework agnostic then I'd like to use it. Otherwise it's just a sham and Netty should get rid of it.

Scottmitch:
For the short term solving the more immediate problems (providing a way to avoid class path fallback) what about the following:
- system property w/out reflection. A system property which supports a fixed set of input (e.g. `slf4j`, `log4j`, `jdk`) and we select/construct the "default classes" in the same way. This wouldn't allow the user flexibility to plugin a custom implementation ... but it does allow more control for the defaults.
- system properties and more reflection. In addition to a system property specifying the fqcn we could have another system property for a static factory method which returns a `InternalLoggerFactory` (and comply with this API in our internal implementations). This is a bit more complex, but allows for more flexibility.

For the longer term seems like folks are generally unhappy with `InternalLoggerFactory`. Lets discuss concrete proposals for alternative approaches. Keep in mind that loggers are typically initialized in a static fashion at the class granularity. If we want to preserve this pattern (which seems to be common in java) then we will need some sort of static (e.g. global) way of accessing a logger object.

To get the conversation started do you think taking a logging API dependency will improve the situation? I know an unofficial goal of Netty is to be "zero dependency" ... but is it worth considering a logging API dependency? There are existing APIs that are well known (slf4j, log4j2, etc...) which have API bridges to different popular implementations. A con is that this dependency may cause headaches for projects which have transitive dependencies on Netty and may use a different version of the same logging API (not sure how frequently the logging APIs change at this point). Another con is what logger implementation actually gets used may depend upon class path ordering.

normanmaurer:
Honestly what happened here makes me feel a bit uneasy and I would even consider revert the change for now. I fear a bit that we may break a lot of apps otherwise which is a “no-go” for a bug-fix release.

WDYT ?

rkapsi:
I'd implement either of Scott's suggestions (maybe the first one if you don't want to commit to any public API yet) and promote it as the _preferred_ way of picking your logging framework and revert the breaking commit and demote the setter method to testing purposes only and/or for people who exercise full control over their class initialization (with a clear warning in the Javadoc). In addition I'd maybe emit an ERROR statement whenever the factory is called 2nd, 3rd... time.

Scottmitch:
> I would even consider revert the change for now

@normanmaurer - You mean "only set the logger once" change? I'm not convinced this is necessary. I agree with @jasontedor's https://github.com/netty/netty/pull/6135#issuecomment-267360408 about this can expose issues where the "wrong" logger is unintentionally used. I would prefer having a solution to more reliably set the default logger, allowing re-setting the logger seems like a hack/work around for an un-coordinated initialization sequence.

normanmaurer:
@scottmitch yes the logger change. I agree it would be "good" to have a fix but with the final releases out any day I think it would be better to revert, cut releases and then work on it as part of the next releases.
IMHO it's just too risky to rush the change in.

Scottmitch:
discussed with @normanmaurer offline ... for the "short-short" term I would be OK with reverting the "only set once" change temporarily so we don't delay the current release (this week hopefully 🎅🎄 🎁). We would re-merge the change into the next snapshot after the release and also address the "more control over the default logger" issue too.

normanmaurer:
@Scottmitch @jasontedor @rkapsi +1

rkapsi:
@normanmaurer sounds good to me

normanmaurer:
This was fixed by @chrisvest in main... lets close this

Issue #3722, opened 2015-05-04
OpenSSL zlib compression (DEFLATE) broken in Netty 4.0.27.Final and netty-tcnative-1.1.32.Final


lkoszanski:
When using Netty with OpenSSL and zlib compression enabled I get the following error:

io.netty.handler.codec.DecoderException: javax.net.ssl.SSLException: error:1408F119:SSL routines:SSL3_GET_RECORD:decryption failed or bad record mac
    at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:346)
    at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:229)
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:339)
    at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:324)
    at io.netty.handler.logging.LoggingHandler.channelRead(LoggingHandler.java:283)
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:339)
    at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:324)
    at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:847)
    at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:131)
    at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:511)
    at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:468)
    at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:382)
    at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:354)
    at io.netty.util.concurrent.SingleThreadEventExecutor$2.run(SingleThreadEventExecutor.java:111)
    at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:137)
    at java.lang.Thread.run(Thread.java:745)
Caused by: javax.net.ssl.SSLException: error:1408F119:SSL routines:SSL3_GET_RECORD:decryption failed or bad record mac
[...]

PR #6136, opened 2016-12-15
Explicit disable support of SSL / TLS Compression

normanmaurer:
Motivation:

Our ReferenceCountedOpenSslEngine does not support compression so we should explicit disable it.
This is related to #3722.

Modifications:

Set SSL_OP_NO_COMPRESSION option.

Result:

Not use compression.


normanmaurer:
@Scottmitch @johnou PTAL

johnou:
lgtm!

normanmaurer:
Cherry-picked into 4.1 (89cb50aefaceac53dfafa7bf8d20637bcbb3bfa8) and 4.0 (49c55cc50996867f9b5e7240bd3c8d3138b470e4)

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -270,6 +270,10 @@ public String run() {
                 SSLContext.setOptions(ctx, SSL.SSL_OP_SINGLE_ECDH_USE);
                 SSLContext.setOptions(ctx, SSL.SSL_OP_SINGLE_DH_USE);
                 SSLContext.setOptions(ctx, SSL.SSL_OP_NO_SESSION_RESUMPTION_ON_RENEGOTIATION);
+
+                // We do not support compression as the moment so we should explicitly disable it.
```
Scottmitch: `as the moment` -> `at the moment`

PR #6137, opened 2016-12-16
Check startsWith instead of exact match to websocketPath in WebSocket…

maliqq:
This small change enables flexible routing based on request URI. E.g.

```scala
pipeline.addLast(new WebSocketServerProtocolHandler("/subscribe", null, true))
pipeline.addLast(new WSFrameHandler() {
  override def userEventTriggered(ctx: ChannelHandlerContext, evt: Any) {
    super.userEventTriggered(ctx, evt)
    evt match {
      case complete: WebSocketServerProtocolHandler.HandshakeComplete =>
        complete.requestUri // <- do something with requestUri
      case _ =>
    }
  }
})
```
Allows connecting on client-side in that way:

```js
new WebSocket("ws://::1:3000/subscribe/topic-1?token=....")
new WebSocket("ws://::1:3000/subscribe/topic-2?token=....")
```

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketServerProtocolHandshakeHandler.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketServerProtocolHandshakeHandler.java
@@ -42,20 +42,32 @@ class WebSocketServerProtocolHandshakeHandler extends ChannelInboundHandlerAdapt
     private final boolean allowExtensions;
     private final int maxFramePayloadSize;
     private final boolean allowMaskMismatch;
+    private final boolean checkStartsWith;
 
     WebSocketServerProtocolHandshakeHandler(String websocketPath, String subprotocols,
             boolean allowExtensions, int maxFrameSize, boolean allowMaskMismatch) {
+        this(websocketPath, subprotocols, allowExtensions, maxFrameSize, allowMaskMismatch, false);
+    }
+
+    WebSocketServerProtocolHandshakeHandler(String websocketPath, String subprotocols,
+            boolean allowExtensions, int maxFrameSize, boolean allowMaskMismatch, boolean checkStartsWith) {
         this.websocketPath = websocketPath;
         this.subprotocols = subprotocols;
         this.allowExtensions = allowExtensions;
         maxFramePayloadSize = maxFrameSize;
         this.allowMaskMismatch = allowMaskMismatch;
+        this.checkStartsWith = checkStartsWith;
     }
 
     @Override
     public void channelRead(final ChannelHandlerContext ctx, Object msg) throws Exception {
         final FullHttpRequest req = (FullHttpRequest) msg;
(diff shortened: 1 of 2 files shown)


Scottmitch:
This would be a change in behavior. At a minimum we would need to preserve the existing behavior as default and provide some way to plugin alternate behaviors (such as this one).

maliqq:
@Scottmitch Default behaviour is preserved "as is" here, since startsWith returns true when websocketPath is strictly equal to req.uri().

Scottmitch:
> Default behaviour is preserved "as is" here

I don't agree. URLs that previously failed this conditional may now pass this conditional. I'm not sure if this is part of the interface contract but this is a change in behavior that existing implementation may rely upon.

maliqq:
@Scottmitch My general assumption is that WebSocket is a general HTTP endpoint with some internal handshake mechanisms, but other options (requests, uri) should be available to parse and dispatch. One plus argument: JS WebSocket interface is not enabled to send custom HTTP headers, so, the only option is to use URI and query parameters. Of course it's possible to rewrite request before WebSocketServerProtocolHandler, but is this a goal to achieve flexibility?

>URLs that previously failed this conditional may now pass this conditional.

I'm not sure If someone could possibly build handler chains after WebSocket's base paths to have some handler failed this condition.

normanmaurer:
@maliqq I agree with @Scottmitch ... Please make it configurable via a constructor and restore the old behaviour as default

maliqq:
@Scottmitch @normanmaurer done, rewrote my commit, please review.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketServerProtocolHandshakeHandler.java
```diff
@@ -42,20 +42,32 @@
...
     @Override
     public void channelRead(final ChannelHandlerContext ctx, Object msg) throws Exception {
         final FullHttpRequest req = (FullHttpRequest) msg;
-        if (!websocketPath.equals(req.uri())) {
+        if (checkStartsWith) {
+            if (!req.uri().startsWith(websocketPath)) {
+                ctx.fireChannelRead(msg);
+                return;
+            }
+        } else if (!websocketPath.equals(req.uri())) {
```
normanmaurer: @maliqq just to make it consistent please use `req.uri().equals(websocketPath)`

maliqq:
@normanmaurer done

normanmaurer:
@maliqq LGTM... Please sign our icla [1] and adjust the commit message to match our template [2]

[1] http://netty.io/s/icla
[2] 

@Scottmitch PTAL again as well.

maliqq:
@normanmaurer signed, did git commit --amend and applied new according with template.

normanmaurer:
@maliqq thanks a lot!

Cherry-picked into 4.1 (16ddf460a62d9946f49d7169a077840bcdf6dc0e) and 4.0 (b91426f)

PR #6142, opened 2016-12-18
Document and generalize `PromiseCombiner`

jchambers:
Motivation:

`PromiseCombiner` is really handy, but it's not obvious how to use it from its [existing documentation](http://netty.io/4.1/api/io/netty/util/concurrent/PromiseCombiner.html)/method signatures.

Modification:

- Added javadoc comments to explain the theory of operation of `PromiseCombiner`.
- Generalized `PromiseCombiner` to work with `Futures` so it's clearer that the things for which it's listening won't be modified.

Result:

`PromiseCombiner` is easier to understand.

### DIFF
--- a/common/src/main/java/io/netty/util/concurrent/PromiseCombiner.java
+++ b/common/src/main/java/io/netty/util/concurrent/PromiseCombiner.java
@@ -17,6 +17,18 @@ package io.netty.util.concurrent;
 
 import io.netty.util.internal.ObjectUtil;
 
+/**
+ * <p>A promise combiner monitors the outcome of a number of discrete futures, then notifies a final, aggregate promise
+ * when all of the combined futures are finished. The aggregate promise will succeed if and only if all of the combined
+ * futures succeed. If any of the combined futures fail, the aggregate promise will fail. The cause failure for the
+ * aggregate promise will be the failure for one of the failed combined futures; if more than one of the combined
+ * futures fails, exactly which cause of failure will be assigned to the aggregate promise is undefined.</p>
+ *
+ * <p>Callers may populate a promise combiner with any number of futures to be combined via the
+ * {@link PromiseCombiner#add(Future)} and {@link PromiseCombiner#addAll(Future[])} methods. When all futures to be
+ * combined have been added, callers must provide an aggregate promise to be notified when all combined promises have
+ * finished via the {@link PromiseCombiner#finish(Promise)} method.</p>
+ */
 public final class PromiseCombiner {
     private int expectedCount;
     private int doneCount;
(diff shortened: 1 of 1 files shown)


common/src/main/java/io/netty/util/concurrent/PromiseCombiner.java
```diff
@@ -36,23 +50,44 @@ public void operationComplete(Future<?> future) throws Exception {
...
+     *
+     * @param futures the futures to add to this promise combiner
+     */
     @SuppressWarnings({ "unchecked", "rawtypes" })
-    public void addAll(Promise... promises) {
-        checkAddAllowed();
-        expectedCount += promises.length;
-        for (Promise promise : promises) {
-            promise.addListener(listener);
+    public void addAll(final Future... futures) {
```
normanmaurer: @jchambers we can not make this change as it would break this:

```
Promise[] promises = ...;
combiner.addAll(promises);
```

Please keep the old method and mark it as `@deprecated`

common/src/main/java/io/netty/util/concurrent/PromiseCombiner.java
```diff
@@ -36,23 +50,44 @@ public void operationComplete(Future<?> future) throws Exception {
...
+     * then the aggregate promise will succeed. If one or more combined futures fails, then the aggregate promise will
+     * fail with the cause of one of the failed futures. If more than one combined future fails, then exactly which
+     * failure will be assigned to the aggregate promise is undefined.</p>
+     *
+     * <p>After this method is called, no more futures may be added via the {@link PromiseCombiner#add(Future)} or
+     * {@link PromiseCombiner#addAll(Future[])} methods.</p>
+     *
+     * @param aggregatePromise the promise to notify when all combined futures have finished
+     */
+    public void finish(final Promise<Void> aggregatePromise) {
```
normanmaurer: remove "final" .

common/src/main/java/io/netty/util/concurrent/PromiseCombiner.java
```diff
@@ -17,6 +17,20 @@
...
+ * futures succeed. If any of the combined futures fail, the aggregate promise will fail. The cause failure for the
+ * aggregate promise will be the failure for one of the failed combined futures; if more than one of the combined
+ * futures fails, exactly which cause of failure will be assigned to the aggregate promise is undefined.</p>
+ *
+ * <p>Callers may populate a promise combiner with any number of futures to be combined via the
+ * {@link PromiseCombiner#add(Future)} and {@link PromiseCombiner#addAll(Future[])} methods. When all futures to be
+ * combined have been added, callers must provide an aggregate promise to be notified when all combined promises have
+ * finished via the {@link PromiseCombiner#finish(Promise)} method.</p>
+ *
+ * @since 4.1
```
normanmaurer: remove

common/src/main/java/io/netty/util/concurrent/PromiseCombiner.java
jchambers: I'm happy to make the change if that's your preference, but why would this break if `Promise extends Future`?

normanmaurer: I think it will stop to compile as you can't pass 'Promise[]' when the signature wants 'Future...'

common/src/main/java/io/netty/util/concurrent/PromiseCombiner.java
jchambers: Done.

Scottmitch:
It is useful ... but also keep in mind it lives in an internal package 😉

normanmaurer:
@Scottmitch it is not in an internal package :(

normanmaurer:
Cherry-picked into 4.1 (7c630feefc62b739afea123547165b9efb4281c0).

@jchambers thanks!

Scottmitch:
@normanmaurer - whoops you are right. I guess its useful for everyone then :)

jchambers:
Ha—okay. That's a relief. Thought I was losing my mind for a minute there ;)

normanmaurer:
LOL no worries 😂

Scottmitch:
multi-tasking fail for me ... sry for the confusion

Issue #6134, opened 2016-12-15
DefaultChannelID : valid ProcessID is not accepted


jo-kin:
Running Netty on a Java-VM on IBM z/OS leads to unnecessary warnings.

You have to know that Unix System Services of IBM z/OS offers PIDs greater than 2^22, as assumed in DefaultChannelID (see Issue 2706)

`   Welcome to  Unix System Services ZOSU of`      
          ` z / O S   Version 2 Release 2`            
`$ ps`
      `PID TTY       TIME CMD`
     `   41 ttyp0000  0:00 /bin/sh`
 ` 50332311 ttyp0000  0:00 /bin/ps`    <== !
`$`

### Expected behavior
Netty should accept valid PIDs coming from any OS. It is a little bit infelicitous to assume that the Linux64 limits are valid for all OS which supports JAVA.

### Actual behavior
Even DefaultChannelID's auto-mechanism succeeded in detecting the correct PID on z/OS the PID is discarded by later check and replaced by a random value.

`Dec 15, 2016 6:05:49 AM io.netty.channel.DefaultChannelId defaultProcessId                  
WARNING: Failed to find the current process ID from '50332208'; using a random value: 474712`

### Steps to reproduce
Start any Netty-using application (i.e. your sample io.netty.example.echo.EchoServer) on IBM Mainframe z/OS. With a little luck the arising PID violates the limit in Netty.

### Minimal yet complete reproducer code (or URL to code)
io.netty.example.echo.EchoServer

### Netty version
4.1.5

### JVM version (e.g. `java -version`)
java version "1.7.0"
Java(TM) SE Runtime Environment (build pmz3170_27sr3fp40-20160422_01(SR3 FP40))
IBM J9 VM (build 2.7, JRE 1.7.0 z/OS s390-31 20160406_298393 (JIT enabled, AOT enabled)
J9VM - R27_Java727_SR3_20160406_0942_B298393
JIT  - tr.r13.java_20160328_114186
GC   - R27_Java727_SR3_20160406_0942_B298393
J9CL - 20160406_298393)
JCL - 20160421_01 based on Oracle jdk7u101-b14

### OS version (e.g. `uname -a`)
IBM z/OS 2.2
$ uname -a
OS/390 ZOSU 25.00 04 2828

normanmaurer:
Sounds good... you want to do a PR?

normanmaurer:
@jo-kin actually let me take care

normanmaurer:
[...]

PR #6144, opened 2016-12-19
[#6134] Do not limit the PID to be <= 4194304

normanmaurer:
Motivation:

On some platforms the PID my be bigger then 4194304 so we should not limit it to 4194304.

Modifications:

Only check that the PID is a valid Integer

Result:

No more warnings on systems where the PID is bigger then 4194304.


normanmaurer:
Cherry-picked into 4.1 (cfd8fb10db1df77183ffe464a7f2c7228f834715)

Issue #6141, opened 2016-12-17
OpenSSLContext Mutual Auth Not Returning Acceptable CAs


bm3780:
### Expected behavior
Openssl provider should behave same as JDK provider when mutual authentication is required and a specific set of trusted Certificate Authorities are specified.  The SSL handshake should return back to the connected peer the same list of configured Certificate Authorities.

```java
SslContext ctx = SslContextBuilder.forServer(APP_CERT, APP_KEY)
    .trustManager(TRUST_CA).sslProvider(SslProvider.JDK)
    .clientAuth(ClientAuth.REQUIRE).build();
```

```bash
$> openssl s_client -connect localhost:8443 -cert app.crt -key app.key -CAfile rootCA.pem -tls1

CONNECTED(00000003)
depth=1 /C=US/ST=VA/O=MyCompany/CN=TestCA
verify return:1
depth=0 /C=US/ST=VA/O=MyCompany/CN=application.com
verify return:1
---
Certificate chain
 0 s:/C=US/ST=VA/O=MyCompany/CN=application.com
   i:/C=US/ST=VA/O=MyCompany/CN=TestCA
---
Server certificate
-----BEGIN CERTIFICATE-----
MIIDAzCCAesCCQCRNC2p+O2dkjANBgkqhkiG9w0BAQsFADA/MQswCQYDVQQGEwJV
...
-----END CERTIFICATE-----
subject=/C=US/ST=VA/O=MyCompany/CN=application.com
issuer=/C=US/ST=VA/O=MyCompany/CN=TestCA
---
Acceptable client certificate CA names
/C=US/ST=VA/O=MyCompany/CN=TestCA
---
SSL handshake has read 1011 bytes and written 2421 bytes
---
New, TLSv1/SSLv3, Cipher is AES128-SHA
Server public key is 2048 bit
Secure Renegotiation IS supported
Compression: NONE
Expansion: NONE
SSL-Session:
    Protocol  : TLSv1
    Cipher    : AES128-SHA
    Session-ID: 5855B84CFD30126A45C3E9E4D54AE560DDFE8248DE82EB163823BB80CC10400C
    Session-ID-ctx: 
    Master-Key: A5296BB5726865BC8350D7DC286ED5472B6ACE3C404462779BDAAD114A54AB64109D965E154076112E9279611F1390D6
    Key-Arg   : None
    Start Time: 1482012748
    Timeout   : 7200 (sec)
    Verify return code: 0 (ok)
---

```
 
[...]

PR #6145, opened 2016-12-19
[#6141] OpenSSLContext Mutual Auth does not announce acceptable CAs

normanmaurer:
Motivation:

Openssl provider should behave same as JDK provider when mutual authentication is required and a specific set of trusted Certificate Authorities are specified. The SSL handshake should return back to the connected peer the same list of configured Certificate Authorities.

Modifications:

Correctly set the CA list.

Result:

Correct and same behaviour as the JDK implementation.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslServerContext.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslServerContext.java
@@ -153,6 +153,21 @@ public final class ReferenceCountedOpenSslServerContext extends ReferenceCounted
                 } else {
                     SSLContext.setCertVerifyCallback(ctx, new TrustManagerVerifyCallback(engineMap, manager));
                 }
+
+                X509Certificate[] issuers = manager.getAcceptedIssuers();
+                if (issuers != null && issuers.length > 0) {
+                    long bio = 0;
+                    try {
+                        bio = toBIO(issuers);
+                        if (!SSLContext.setCACertificateBio(ctx, bio)) {
+                            throw new SSLException("unable to setup accepted issuers for trustmanager " + manager);
+                        }
+                    } finally {
+                        freeBio(bio);
+                    }
+                }
+            } catch (SSLException e) {
+                throw e;
             } catch (Exception e) {
                 throw new SSLException("unable to setup trustmanager", e);
             }
--- a/handler/src/test/java/io/netty/handler/ssl/RenegotiateTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/RenegotiateTest.java
@@ -38,7 +38,7 @@ import java.util.concurrent.atomic.AtomicReference;
 
 public abstract class RenegotiateTest {
 
-    @Test(timeout = 5000)
+    @Test(timeout = 30000)
     public void testRenegotiateServer() throws Throwable {
         final AtomicReference<Throwable> error = new AtomicReference<Throwable>();
         final CountDownLatch latch = new CountDownLatch(2);
--- a/handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
@@ -538,7 +538,7 @@ public abstract class SSLEngineTest {
         }
     }
 
-    @Test(timeout = 3000)
+    @Test(timeout = 30000)
     public void clientInitiatedRenegotiationWithFatalAlertDoesNotInfiniteLoopServer()
             throws CertificateException, SSLException, InterruptedException, ExecutionException {
         final SelfSignedCertificate ssc = new SelfSignedCertificate();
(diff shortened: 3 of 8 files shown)


normanmaurer:
This depends on https://github.com/netty/netty-tcnative/pull/211/files

normanmaurer:
@Scottmitch @nmittler unfortunately I did found no way to test this with the java only apis and so could not add a unit test. A way to test this with openssl s_client is listed in #6141

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslServerContext.java
```diff
@@ -153,6 +153,17 @@ static ServerContext newSessionContext(ReferenceCountedOpenSslContext thiz, long
                 } else {
                     SSLContext.setCertVerifyCallback(ctx, new TrustManagerVerifyCallback(engineMap, manager));
                 }
+
+                X509Certificate[] issuers = manager.getAcceptedIssuers();
+                if (issuers != null && issuers.length > 0) {
+                    long bio = 0;
+                    try {
+                        bio = toBIO(issuers);
+                        SSLContext.setCACertificateBio(ctx, bio);
```
Scottmitch: did we want to use the return value ... or have this method throw if something bad happens?

normanmaurer: yes we should check if it returns false and throw something... good catch.

normanmaurer:
@Scottmitch PTAL again

Scottmitch:
@normanmaurer - Any ideas about the build failure:

https://garage.netty.io/teamcity/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=14906#_focus=107068
```
[14:05:56][io.netty:netty-handler] Tests in error: 
[14:05:56][io.netty:netty-handler]   JdkOpenSslEngineInteroptTest>SSLEngineTest.clientInitiatedRenegotiationWithFatalAlertDoesNotInfiniteLoopServer:546 ? TestTimedOut
[14:05:56][io.netty:netty-handler]   OpenSslEngineTest>SSLEngineTest.clientInitiatedRenegotiationWithFatalAlertDoesNotInfiniteLoopServer:546 ? TestTimedOut
[14:05:56][io.netty:netty-handler]   OpenSslRenegotiateTest>RenegotiateTest.testRenegotiateServer:49 ? TestTimedOut
[14:05:56][io.netty:netty-handler]   ReferenceCountedOpenSslEngineTest>SSLEngineTest.clientInitiatedRenegotiationWithFatalAlertDoesNotInfiniteLoopServer:546 ? TestTimedOut
[14:05:56][io.netty:netty-handler]   SniClientTest.testSniClientJdkSslServerOpenSsl:52->testSniClient:70 ? TestTimedOut
[14:05:56][io.netty:netty-handler]   SniClientTest.testSniClientOpenSslServerJdkSsl:58->testSniClient:88 ? TestTimedOut
[14:05:56][io.netty:netty-handler]   SniClientTest.testSniClientOpenSslServerOpenSsl:46->testSniClient:70 ? TestTimedOut
[14:05:56][io.netty:netty-handler]   SniHandlerTest.testReplaceHandler:334 ? TestTimedOut test timed out after 1000...
[14:05:56][io.netty:netty-handler]   SslHandlerTest.testRemoval:207 ? TestTimedOut test timed out after 5000 millis...
```

normanmaurer:
Nope not yet... can't reproduce locally :/

normanmaurer:
@Scottmitch so it seems like the SSL tests take a lot more time with this change when leak detection is enabled... still investigate why tho.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslServerContext.java
```diff
@@ -153,6 +153,23 @@ static ServerContext newSessionContext(ReferenceCountedOpenSslContext thiz, long
                 } else {
                     SSLContext.setCertVerifyCallback(ctx, new TrustManagerVerifyCallback(engineMap, manager));
                 }
+
+                /* Disable to test if test timeouts are gone.
+                X509Certificate[] issuers = manager.getAcceptedIssuers();
```
netkins: MAJOR This block of commented-out lines of code should be removed. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3ACommentedOutCodeLine)

normanmaurer:
@Scottmitch ha... I think I know why... let me check

normanmaurer:
@Scottmitch I figured it out... basically I had to increase all timeouts in the ssl tests as the tests take a lot more time now when running with leak detection. This is because of https://github.com/netty/netty/pull/6145/files#diff-59f5f02c750d4f4de262453ac10f5e52R161 as this will need to write all the certs to a BIO. Which are most of the times 92 certs in our test. This is done via the `Base64` class that uses `getByte(...)` and `setByte(...)` a lot which is just slow with leak detection for very big certs.

So I think we can pull it in. Agree ?

normanmaurer:
@Scottmitch yep... let me create an issue.

normanmaurer:
@Scottmitch see https://github.com/netty/netty/issues/6205

normanmaurer:
Cherry-picked into 4.1 (be8c16c) and 4.0 (81fbf9a)

PR #6000, opened 2016-11-09
New native transport for kqueue

Scottmitch:
Motivation:
We currently don't have a native transport which supports kqueue https://www.freebsd.org/cgi/man.cgi?query=kqueue&sektion=2. This can be useful for BSD systems such as MacOS to take advantage of native features, and provide feature parity with the Linux native transport.

Modifications:
- Make a new transport-native-unix-common module with all the java classes and JNI code for generic unix items. This module will build a static library for each unix platform, and included in the dynamic libraries used for JNI (e.g. transport-native-epoll, and eventually kqueue).
- Make a new transport-native-unix-common-tests module where the tests for the transport-native-unix-common module will live. This is so each unix platform can inherit from these test and ensure they pass.
- Add a new transport-native-kqueue module which uses JNI to directly interact with kqueue

Result:
JNI support for kqueue.
Fixes https://github.com/netty/netty/issues/2448
Fixes https://github.com/netty/netty/issues/4231

PR #6146, opened 2016-12-19
Wrap operations that require SocketPermission from the SecurityManage…

tbrooks8:
…r in doPrivileged blocks

Motivation:

Currently Netty does not wrap socket connect, bind, or accept operations in doPrivileged blocks. Nor does it wrap cases where a dns lookup might happen.

This prevents an application utilizing the SecurityManager from only granting SocketPermissions to Netty.

Modifications:

I have introduced a class (SocketUtils) that wraps operations requiring SocketPermissions in doPrivileged blocks.

Result:

A user of Netty can grant SocketPermissions explicitly to the Netty jar, without granting it to the rest of their application.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/SocketUtils.java
+++ b/common/src/main/java/io/netty/util/internal/SocketUtils.java
@@ -1,0 +1,197 @@
+/*
+ * Copyright 2016 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.util.internal;
+
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.NetworkInterface;
+import java.net.ServerSocket;
+import java.net.Socket;
+import java.net.SocketAddress;
+import java.net.SocketException;
+import java.net.SocketPermission;
+import java.net.UnknownHostException;
+import java.nio.channels.DatagramChannel;
+import java.nio.channels.ServerSocketChannel;
+import java.nio.channels.SocketChannel;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.security.PrivilegedActionException;
+import java.security.PrivilegedExceptionAction;
+import java.util.Enumeration;
+
+/**
+ * Provides socket operations with privileges enabled. This is necessary for applications that use the
+ * {@link SecurityManager} to restrict {@link SocketPermission} to their application. By asserting that these
+ * operations are privileged, the operations can proceed even if some code in the calling chain lacks the appropriate
+ * {@link SocketPermission}.
+ */
+public final class SocketUtils {
+
+    private SocketUtils() {
+    }
+
+    public static void connect(final Socket socket, final SocketAddress remoteAddress, final int timeout)
+            throws IOException {
+        try {
+            AccessController.doPrivileged(new PrivilegedExceptionAction<Void>() {
+                @Override
+                public Void run() throws IOException {
+                    socket.connect(remoteAddress, timeout);
+                    return null;
+                }
+            });
+        } catch (PrivilegedActionException e) {
+            throw (IOException) e.getCause();
+        }
+    }
+
+    public static void bind(final Socket socket, final SocketAddress bindpoint) throws IOException {
+        try {
+            AccessController.doPrivileged(new PrivilegedExceptionAction<Void>() {
+                @Override
+                public Void run() throws IOException {
+                    socket.bind(bindpoint);
+                    return null;
+                }
+            });
+        } catch (PrivilegedActionException e) {
+            throw (IOException) e.getCause();
+        }
+    }
+
+    public static boolean connect(final SocketChannel socketChannel, final SocketAddress remoteAddress)
+            throws IOException {
+        try {
+            return AccessController.doPrivileged(new PrivilegedExceptionAction<Boolean>() {
+                @Override
+                public Boolean run() throws IOException {
+                    return socketChannel.connect(remoteAddress);
+                }
+            });
+        } catch (PrivilegedActionException e) {
+            throw (IOException) e.getCause();
+        }
+    }
+
+    public static void bind(final SocketChannel socketChannel, final SocketAddress address) throws IOException {
+        try {
+            AccessController.doPrivileged(new PrivilegedExceptionAction<Void>() {
+                @Override
+                public Void run() throws IOException {
+                    socketChannel.bind(address);
+                    return null;
+                }
+            });
+        } catch (PrivilegedActionException e) {
+            throw (IOException) e.getCause();
+        }
+    }
+
+    public static SocketChannel accept(final ServerSocketChannel serverSocketChannel) throws IOException {
+        try {
+            return AccessController.doPrivileged(new PrivilegedExceptionAction<SocketChannel>() {
+                @Override
+                public SocketChannel run() throws IOException {
+                    return serverSocketChannel.accept();
+                }
+            });
+        } catch (PrivilegedActionException e) {
+            throw (IOException) e.getCause();
+        }
+    }
+
+    public static void bind(final DatagramChannel networkChannel, final SocketAddress address) throws IOException {
+        try {
+            AccessController.doPrivileged(new PrivilegedExceptionAction<Void>() {
+                @Override
+                public Void run() throws IOException {
+                    networkChannel.bind(address);
+                    return null;
+                }
+            });
+        } catch (PrivilegedActionException e) {
+            throw (IOException) e.getCause();
+        }
+    }
+
+    public static SocketAddress localSocketAddress(final ServerSocket socket) {
+        return AccessController.doPrivileged(new PrivilegedAction<SocketAddress>() {
+            @Override
+            public SocketAddress run() {
+                return socket.getLocalSocketAddress();
+            }
+        });
+    }
+
+    public static InetAddress addressByName(final String hostname) throws UnknownHostException {
+        try {
+            return AccessController.doPrivileged(new PrivilegedExceptionAction<InetAddress>() {
+                @Override
+                public InetAddress run() throws UnknownHostException {
+                    return InetAddress.getByName(hostname);
+                }
+            });
+        } catch (PrivilegedActionException e) {
+            throw (UnknownHostException) e.getCause();
+        }
+    }
+
+    public static InetAddress[] allAddressesByName(final String hostname) throws UnknownHostException {
+        try {
+            return AccessController.doPrivileged(new PrivilegedExceptionAction<InetAddress[]>() {
+                @Override
+                public InetAddress[] run() throws UnknownHostException {
+                    return InetAddress.getAllByName(hostname);
+                }
+            });
+        } catch (PrivilegedActionException e) {
+            throw (UnknownHostException) e.getCause();
+        }
+    }
+
+    public static InetSocketAddress socketAddress(final String hostname, final int port) {
+        return AccessController.doPrivileged(new PrivilegedAction<InetSocketAddress>() {
+            @Override
+            public InetSocketAddress run() {
+                return new InetSocketAddress(hostname, port);
+            }
+        });
+    }
+
+    public static Enumeration<InetAddress> addressesFromNetworkInterface(final NetworkInterface intf) {
+        return AccessController.doPrivileged(new PrivilegedAction<Enumeration<InetAddress>>() {
+            @Override
+            public Enumeration<InetAddress> run() {
+                return intf.getInetAddresses();
+            }
+        });
+    }
+
+    public static byte[] hardwareAddressFromNetworkInterface(final NetworkInterface intf) throws SocketException {
+        try {
+            return AccessController.doPrivileged(new PrivilegedExceptionAction<byte[]>() {
+                @Override
+                public byte[] run() throws SocketException {
+                    return intf.getHardwareAddress();
+                }
+            });
+        } catch (PrivilegedActionException e) {
+            throw (SocketException) e.getCause();
+        }
+    }
+}
--- a/example/src/main/java/io/netty/example/qotm/QuoteOfTheMomentClient.java
+++ b/example/src/main/java/io/netty/example/qotm/QuoteOfTheMomentClient.java
@@ -24,8 +24,7 @@ import io.netty.channel.nio.NioEventLoopGroup;
 import io.netty.channel.socket.DatagramPacket;
 import io.netty.channel.socket.nio.NioDatagramChannel;
 import io.netty.util.CharsetUtil;
-
-import java.net.InetSocketAddress;
+import io.netty.util.internal.SocketUtils;
 
 /**
  * A UDP broadcast client that asks for a quote of the moment (QOTM) to {@link QuoteOfTheMomentServer}.
@@ -52,7 +51,7 @@ public final class QuoteOfTheMomentClient {
             // Broadcast the QOTM request to port 8080.
             ch.writeAndFlush(new DatagramPacket(
                     Unpooled.copiedBuffer("QOTM?", CharsetUtil.UTF_8),
-                    new InetSocketAddress("255.255.255.255", PORT))).sync();
+                    SocketUtils.socketAddress("255.255.255.255", PORT))).sync();
 
             // QuoteOfTheMomentClientHandler will close the DatagramChannel when a
             // response is received.  If the channel is not closed within 5 seconds,
--- a/microbench/src/main/java/io/netty/microbench/internal/PrivilegedSocketOperationsBenchmark.java
+++ b/microbench/src/main/java/io/netty/microbench/internal/PrivilegedSocketOperationsBenchmark.java
@@ -1,0 +1,193 @@
+/*
+ * Copyright 2016 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.microbench.internal;
+
+import io.netty.microbench.util.AbstractMicrobenchmark;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.nio.channels.ServerSocketChannel;
+import java.security.AccessController;
+import java.security.NoSuchAlgorithmException;
+import java.security.Policy;
+import java.security.PrivilegedActionException;
+import java.security.PrivilegedExceptionAction;
+import java.security.URIParameter;
+import java.util.concurrent.TimeUnit;
+
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.SECONDS)
+public class PrivilegedSocketOperationsBenchmark extends AbstractMicrobenchmark {
+
+    @State(Scope.Benchmark)
+    public static class SecurityManagerInstalled {
+
+        @Setup
+        public void setup() throws IOException, NoSuchAlgorithmException, URISyntaxException {
+            final URI policyFile = PrivilegedSocketOperationsBenchmark.class.getResource("/jmh-security.policy")
+                    .toURI();
+            Policy.setPolicy(Policy.getInstance("JavaPolicy", new URIParameter(policyFile)));
+            System.setSecurityManager(new SecurityManager());
+        }
+
+        @TearDown
+        public void tearDown() throws IOException {
+            System.setSecurityManager(null);
+        }
+    }
+
+    @State(Scope.Benchmark)
+    public static class SecurityManagerEmpty {
+
+        @Setup
+        public void setup() throws IOException, NoSuchAlgorithmException, URISyntaxException {
+            System.setSecurityManager(null);
+        }
+    }
+
+    @Benchmark
+    public ServerSocketChannel testWithSMNoPrivileged(final SecurityManagerInstalled sm) throws IOException {
+        final ServerSocketChannel ssc = ServerSocketChannel.open();
+        ssc.socket().bind(null);
+        ssc.configureBlocking(false);
+        ssc.accept();
+        ssc.close();
+        return ssc;
+    }
+
+    @Benchmark
+    public ServerSocketChannel testWithSM(final SecurityManagerInstalled sm) throws IOException {
+        try {
+            final ServerSocketChannel ssc = AccessController.doPrivileged(
+                    new PrivilegedExceptionAction<ServerSocketChannel>() {
+                        @Override
+                        public ServerSocketChannel run() throws Exception {
+                            final ServerSocketChannel ssc = ServerSocketChannel.open();
+                            ssc.socket().bind(null);
+                            ssc.configureBlocking(false);
+                            ssc.accept();
+                            return ssc;
+                        }
+                    });
+            ssc.close();
+            return ssc;
+        } catch (final PrivilegedActionException e) {
+            throw (IOException) e.getCause();
+        }
+    }
+
+    @Benchmark
+    public ServerSocketChannel testWithSMWithNullCheck(final SecurityManagerInstalled sm) throws IOException {
+        if (System.getSecurityManager() != null) {
+            try {
+                final ServerSocketChannel ssc = AccessController.doPrivileged(
+                        new PrivilegedExceptionAction<ServerSocketChannel>() {
+                            @Override
+                            public ServerSocketChannel run() throws Exception {
+                                final ServerSocketChannel ssc = ServerSocketChannel.open();
+                                ssc.socket().bind(null);
+                                ssc.configureBlocking(false);
+                                ssc.accept();
+                                return ssc;
+                            }
+                        });
+                ssc.close();
+                return ssc;
+            } catch (final PrivilegedActionException e) {
+                throw (IOException) e.getCause();
+            }
+        } else {
+            // this should never happen during benchmarking, but we write the correct code here
+            final ServerSocketChannel ssc = ServerSocketChannel.open();
+            ssc.socket().bind(null);
+            ssc.configureBlocking(false);
+            ssc.accept();
+            ssc.close();
+            return ssc;
+        }
+    }
+
+    @Benchmark
+    public ServerSocketChannel testWithoutSMNoPrivileged(final SecurityManagerEmpty sm) throws IOException {
+        final ServerSocketChannel ssc = ServerSocketChannel.open();
+        ssc.socket().bind(null);
+        ssc.configureBlocking(false);
+        ssc.accept();
+        ssc.close();
+        return ssc;
+    }
+
+    @Benchmark
+    public ServerSocketChannel testWithoutSM(final SecurityManagerEmpty sm) throws IOException {
+        try {
+            final ServerSocketChannel ssc = AccessController.doPrivileged(
+                    new PrivilegedExceptionAction<ServerSocketChannel>() {
+                        @Override
+                        public ServerSocketChannel run() throws Exception {
+                            final ServerSocketChannel ssc = ServerSocketChannel.open();
+                            ssc.socket().bind(null);
+                            ssc.configureBlocking(false);
+                            ssc.accept();
+                            return ssc;
+                        }
+                    });
+            ssc.close();
+            return ssc;
+        } catch (final PrivilegedActionException e) {
+            throw (IOException) e.getCause();
+        }
+    }
+
+    @Benchmark
+    public ServerSocketChannel testWithoutSMWithNullCheck(final SecurityManagerEmpty sm) throws IOException {
+        if (System.getSecurityManager() != null) {
+            // this should never happen during benchmarking, but we write the correct code here
+            try {
+                final ServerSocketChannel ssc = AccessController.doPrivileged(
+                        new PrivilegedExceptionAction<ServerSocketChannel>() {
+                            @Override
+                            public ServerSocketChannel run() throws Exception {
+                                final ServerSocketChannel ssc = ServerSocketChannel.open();
+                                ssc.socket().bind(null);
+                                ssc.configureBlocking(false);
+                                ssc.accept();
+                                return ssc;
+                            }
+                        });
+                ssc.close();
+                return ssc;
+            } catch (final PrivilegedActionException e) {
+                throw (IOException) e.getCause();
+            }
+        } else {
+            final ServerSocketChannel ssc = ServerSocketChannel.open();
+            ssc.socket().bind(null);
+            ssc.configureBlocking(false);
+            ssc.accept();
+            ssc.close();
+            return ssc;
+        }
+    }
+}
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddresses.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddresses.java
@@ -16,12 +16,12 @@
 
 package io.netty.resolver.dns;
 
+import io.netty.util.internal.SocketUtils;
 import io.netty.util.internal.UnstableApi;
 import io.netty.util.internal.logging.InternalLogger;
 import io.netty.util.internal.logging.InternalLoggerFactory;
 
 import java.lang.reflect.Method;
-import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -54,7 +54,7 @@ public abstract class DnsServerAddresses {
             final List<String> list = (List<String>) nameservers.invoke(instance);
             for (String a: list) {
                 if (a != null) {
-                    defaultNameServers.add(new InetSocketAddress(InetAddress.getByName(a), DNS_PORT));
+                    defaultNameServers.add(new InetSocketAddress(SocketUtils.addressByName(a), DNS_PORT));
                 }
             }
         } catch (Exception ignore) {
@@ -70,8 +70,8 @@ public abstract class DnsServerAddresses {
         } else {
             Collections.addAll(
                     defaultNameServers,
-                    new InetSocketAddress("8.8.8.8", DNS_PORT),
-                    new InetSocketAddress("8.8.4.4", DNS_PORT));
+                    SocketUtils.socketAddress("8.8.8.8", DNS_PORT),
+                    SocketUtils.socketAddress("8.8.4.4", DNS_PORT));
 
             if (logger.isWarnEnabled()) {
                 logger.warn(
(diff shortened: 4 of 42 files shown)


tbrooks8:
Background:

We are looking into [this for ElasticSearch](https://github.com/elastic/elasticsearch/issues/22116). We essentially would like to remove [SocketPermissions](https://docs.oracle.com/javase/7/docs/api/java/net/SocketPermission.html) from the core of elasticsearch and only grant them to our transport layer.

I did this by moving the common operations into SocketUtils in common.

This raises a few concerns:

1. Is this the correct place for this? Would you prefer it in the NetUtil class?

2. I had issues with the NioUdtByteConnectorChannel class in transport-udt. Even though it appears to me that SocketChannelUDT extends SocketChannel properly, when I ran the test suite using the SocketUtils class I introduced, I was seeing test failures due to the following exception:

`java.lang.AbstractMethodError: java.nio.channels.SocketChannel.bind(Ljava/net/SocketAddress;)Ljava/nio/channels/SocketChannel;
	at io.netty.util.SocketUtils$4.run(SocketUtils.java:86)
	at io.netty.util.SocketUtils$4.run(SocketUtils.java:82)
	at java.security.AccessController.doPrivileged(Native Method)
	at io.netty.util.SocketUtils.bind(SocketUtils.java:82)
	at io.netty.channel.udt.nio.NioUdtByteConnectorChannel.privilegedBind(NioUdtByteConnectorChannel.java:197)
	at io.netty.channel.udt.nio.NioUdtByteConnectorChannel.doBind(NioUdtByteConnectorChannel.java:97)
	at io.netty.channel.udt.nio.NioUdtByteConnectorChannel.doConnect(NioUdtByteConnectorChannel.java:108)
	at io.netty.channel.nio.AbstractNioChannel$AbstractNioUnsafe.connect(AbstractNioChannel.java:254)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.connect(DefaultChannelPipeline.java:1266)
	at io.netty.channel.AbstractChannelHandlerContext.invokeConnect(AbstractChannelHandlerContext.java:546)
	at io.netty.channel.AbstractChannelHandlerContext.connect(AbstractChannelHandlerContext.java:531)
	at io.netty.channel.DefaultChannelPipeline.connect(DefaultChannelPipeline.java:991)
	at io.netty.channel.AbstractChannel.connect(AbstractChannel.java:260)
	at io.netty.bootstrap.Bootstrap$3.run(Bootstrap.java:254)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:163)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:403)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:445)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:858)
	at java.lang.Thread.run(Thread.java:745)`

There may be some class path configuration issue that I do not understand. However, when I duplicated the doPrivilege blocks directly into the NioUdtByteConnectorChannel class everything worked properly.

3. Have I covered all the use cases? It’s entirely possible that I have missed some. These are the instances I captured:

Socket
bind(SocketAddress)
connect(SocketAddress)

SocketChannel
bind(SocketAddress)
connect(SocketAddress)

ServerSocket
accept()

DatagramChannel
bind(SocketAddress)

InetAddress
getByName(String)
getAllByName(String)

InetSocketAddress
new InetSocketAddress(String, int)

4. Are there changes that should be made to testing? Obviously all of the code is currently tested from the current test suite. But if this change is accepted, are there other configuration changes that should be made to running tests? Like running with the SecurityManger enabled or something?

normanmaurer:
@tbrooks8 honestly I am a bit concerned about the overhead that this change will introduce for all people that not care about `SecurityManager`s. I think this will add a lot of overhead ...

transport-udt/src/main/java/io/netty/channel/udt/nio/NioUdtByteConnectorChannel.java
```diff
@@ -30,10 +30,14 @@
...
 
+import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
+import java.security.AccessController;
+import java.security.PrivilegedActionException;
+import java.security.PrivilegedExceptionAction;
 
-import static java.nio.channels.SelectionKey.OP_CONNECT;
+import static java.nio.channels.SelectionKey.*;
```
normanmaurer: revert ?

normanmaurer:
@tbrooks8 also I wonder if it may be "cheaper" to first check if a `SecurityManager` is present and only in this case use the `AcessController.doPrivileged`. WDYT ?

tbrooks8:
@normanmaurer I modified the PR to fix the formatting issues. I'm sorry about that. I think I must have lost the netty formatting configs.

As to what you're suggesting, that seems to be what the JDK does. It first does a null check on the SecurityManager before specifically checking permissions. 

Inside of Socket's bind(SocketAddress):
```
SecurityManager security = System.getSecurityManager();
        if (security != null) {
            security.checkListen(port);
        }
```

Adding an additional null check in the SocketUtils class before using the privileged versions would still offer a "fast path" and a path that allows applications to control permissions.

Let me know if this is an approach that you would prefer and I can further modify the PR.

jasontedor:
> I wonder if it may be "cheaper" to first check if a `SecurityManager` is present and only in this case use the `AcessController.doPrivileged`. WDYT ?

@normanmaurer The JVM is going to check if a security manager is present before checking permissions either way?

normanmaurer:
@tbrooks8 yep I think that would be better...

normanmaurer:
@jasontedor yes but you still would create more objects and enter a native call which is more expensive ? At least that is what it seems to do when I check the `AccessController.doPrivileged(...)` signature. Or I am missing something ?

normanmaurer:
@jasontedor btw please let me know if I am wrong as I am not very "into" `SecurityManager` stuff. So it could be that I am completely wrong here.

transport/src/main/java/io/netty/channel/socket/nio/NioServerSocketChannel.java
```diff
@@ -137,7 +138,7 @@ protected void doClose() throws Exception {
 
     @Override
     protected int doReadMessages(List<Object> buf) throws Exception {
-        SocketChannel ch = javaChannel().accept();
+        SocketChannel ch = SocketUtils.accept(javaChannel());
```
netkins: BLOCKER Close this "SocketChannel". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2095)

resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddresses.java
```diff
@@ -70,8 +71,8 @@
         } else {
             Collections.addAll(
                     defaultNameServers,
-                    new InetSocketAddress("8.8.8.8", DNS_PORT),
-                    new InetSocketAddress("8.8.4.4", DNS_PORT));
+                    SocketUtils.socketAddress("8.8.8.8", DNS_PORT),
```
netkins: MAJOR Make this IP "8.8.8.8" address configurable. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1313)

resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddresses.java
```diff
@@ -70,8 +71,8 @@
...
+                    SocketUtils.socketAddress("8.8.4.4", DNS_PORT));
```
netkins: MAJOR Make this IP "8.8.4.4" address configurable. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1313)

example/src/main/java/io/netty/example/qotm/QuoteOfTheMomentClient.java
```diff
@@ -52,7 +53,7 @@ public static void main(String[] args) throws Exception {
             // Broadcast the QOTM request to port 8080.
             ch.writeAndFlush(new DatagramPacket(
                     Unpooled.copiedBuffer("QOTM?", CharsetUtil.UTF_8),
-                    new InetSocketAddress("255.255.255.255", PORT))).sync();
+                    SocketUtils.socketAddress("255.255.255.255", PORT))).sync();
```
netkins: MAJOR Make this IP "255.255.255.255" address configurable. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1313)

s1monw:
> @tbrooks8 honestly I am a bit concerned about the overhead that this change will introduce for all people that not care about SecurityManagers. I think this will add a lot of overhead ...

I don't understand what you mean by overhead. Can you elaborate what you are concerned about? I really don't see the overhead for things that not considered hot code? Are you concerned about object creation, GC, runtime cost? I have trouble to understand the reasoning here.

slandelle:
> things that not considered hot code

Opening connections and resolving hostnames might not be hot code for ES use case, but it can be for some other people (definitively is for me).
For those people who don't have a SecurityManager, instantiating all those `PrivilegedExceptionAction` is definitively an overhead.
Or am I missing something?

s1monw:
> For those people who don't have a SecurityManager, instantiating all those PrivilegedExceptionAction is definitively an overhead.
Or am I missing something?

I don't argue there is additional code executed but can you measure it? Is it worth the complexity to add invariants? Like there are tons of these calls in the JDK, everytime you open a file etc. or delete a file. When you resolve hosts etc. Does the big O change here? I mean what are we optimizing away? I really want to understand that since I think there are 0 numbers here from  all the concerns I see. There is only premature optimization. Don't get me wrong, I really wonder if we optimize the right thing here.

slandelle:
> since I think there are 0 numbers here from all the concerns I see

@s1monw Not my project here, so definitively not my call here, but IMHO, you guys are the ones who should provide those numbers as you're the ones who want to push those changes.

s1monw:
> @s1monw Not my project here, so definitively not my call here, but IMHO, you guys are the ones who should provide those numbers as you're the ones who want to push those changes.

I tend to agree with the need of benchmarks. The level we are optimizing here is really questionable and it raises the bar enormously high to contribute to a project. Especialy when the patch fixes correctness since netty is literally unusable with a security manager in place. at the end of the day I don't care too much if the project adds another null check but the attitude on what level code is optimized here is concerning. There are a gazillion other places in the codebase where someone might think they run code that is tested but it's not. The reasons are the same.
My problem here is that if you have invariants they need to be tested but netty does not have the testing in place for this hence my correctness comment. @tbrooks8 would you mind looking into the overhead of a doPriv call with and without the sec manager installed. I think for us it's very important to get these numbers to not argue about this over and over again.

tbrooks8:
@s1monw yes I can do that. 

Setting aside performance, please let me know if anyone has issues with the code design. Or any other adjustments. Or any insight into the AbstractMethodError.

jasontedor:
> yes but you still would create more objects and enter a native call which is more expensive ?

A single small object, sure, and there's already native calls here anyway (bind, connect, etc.) (not to mention the use of `synchronized` and additional allocations in the wrapped calls, etc.) but then to start worrying about things at this level means maybe worrying about branch misprediction on the null check, and the impact of the larger method size from the extra bytecodes.

Adding or not adding the null check is not what concerns me here. What concerns me is the premature optimization and statements like:

> I think this will add a lot of overhead

normanmaurer:
@tbrooks8 in general I think the "design" is good. I have no idea about the `AbstractMethodError` and why it happens tho :( In another issue @s1monw and myself already discussed the problem with not testing with a `SecurityManager` in place and its something we need to tackle. That said I don't  have much experience in this area....

@s1monw I am concerned about the extra native call and work done in the native code. Again if it is more or less non-existent then let us just move on and not care. But for this it would be nice to have a small JMH benchmark that just measure such an overhead. And calling accept () can be a "hot-code" path for a very busy server.

@jasontedor @s1monw I can understand why you want these changes and why you feel so strong about these. I agree correctly handle restrictions that should be enforced via the `SecurityManager` is the way to go. Still when doing changes that may effect performance in a negative way (and note that I said "may", as I really don't know if this is the case) for others I try to be super caution. Only because the JDK is doing something does not mean it is always the most efficient way ;) So all I ask for is a simple JMH benchmark calling some method and measure the overhead of the change.

@jasontedor I agree I should not have said `I think this will add a lot of overhead` but `I fear this will add a noticeable overhead`

 And don't get me started on the synchronized usage in the wrapped calls, but this is a different story and nothing we can do about it.

So to sum it up, can we just demonstrate its not an "big overhead" via a JMH benchmark and then move on ? Once this is shown I am happy to pull this in and other PRs as well related to this in the future. 

@tbrooks8 @jasontedor @s1monw thanks again helping fixing these things. When netty can be used with a `SecurityManager` at the end its because of you!

common/src/main/java/io/netty/util/SocketUtils.java
```diff
@@ -0,0 +1,156 @@
...
+import java.net.UnknownHostException;
+import java.nio.channels.DatagramChannel;
+import java.nio.channels.ServerSocketChannel;
+import java.nio.channels.SocketChannel;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.security.PrivilegedActionException;
+import java.security.PrivilegedExceptionAction;
+
+public final class SocketUtils {
```
normanmaurer: I wonder if we should move this in the `internal` package as this is nothing the user should use directly.

common/src/main/java/io/netty/util/SocketUtils.java
```diff
@@ -0,0 +1,156 @@
...
+        try {
+            AccessController.doPrivileged(new PrivilegedExceptionAction<Void>() {
+                @Override
+                public Void run() throws IOException {
+                    networkChannel.bind(address);
+                    return null;
+                }
+            });
+        } catch (PrivilegedActionException e) {
+            throw (IOException) e.getCause();
```
normanmaurer: @tbrooks8 I think we also need to handle the other exceptions and not blindly cast:

https://docs.oracle.com/javase/7/docs/api/java/nio/channels/DatagramChannel.html#bind(java.net.SocketAddress)

common/src/main/java/io/netty/util/SocketUtils.java
```diff
@@ -0,0 +1,156 @@
...
+    public static SocketChannel accept(final ServerSocketChannel serverSocketChannel) throws IOException {
+        try {
+            return AccessController.doPrivileged(new PrivilegedExceptionAction<SocketChannel>() {
+                @Override
+                public SocketChannel run() throws IOException {
+                    return serverSocketChannel.accept();
+                }
+            });
+        } catch (PrivilegedActionException e) {
+            throw (IOException) e.getCause();
```
normanmaurer: @tbrooks8 I think we also need to handle the other exceptions and not blindly cast:

https://docs.oracle.com/javase/7/docs/api/java/nio/channels/ServerSocketChannel.html#accept()

common/src/main/java/io/netty/util/SocketUtils.java
```diff
@@ -0,0 +1,156 @@
...
+            throws Exception {
+        try {
+            return AccessController.doPrivileged(new PrivilegedExceptionAction<Boolean>() {
+                @Override
+                public Boolean run() throws IOException {
+                    return socketChannel.connect(remoteAddress);
+                }
+            });
+        } catch (PrivilegedActionException e) {
+            throw (IOException) e.getCause();
```
normanmaurer: Correctly handle different exceptions ?
https://docs.oracle.com/javase/7/docs/api/java/nio/channels/SocketChannel.html#connect(java.net.SocketAddress)

common/src/main/java/io/netty/util/SocketUtils.java
```diff
@@ -0,0 +1,156 @@
...
+        try {
+            AccessController.doPrivileged(new PrivilegedExceptionAction<Void>() {
+                @Override
+                public Void run() throws IOException {
+                    socketChannel.bind(address);
+                    return null;
+                }
+            });
+        } catch (PrivilegedActionException e) {
+            throw (IOException) e.getCause();
```
normanmaurer: same as above... Actually please review all exception handling for all methods :)

transport-udt/src/main/java/io/netty/channel/udt/nio/NioUdtByteConnectorChannel.java
```diff
@@ -186,4 +190,33 @@ public InetSocketAddress localAddress() {
...
+        try {
+            return AccessController.doPrivileged(new PrivilegedExceptionAction<Boolean>() {
+                @Override
+                public Boolean run() throws IOException {
+                    return socketChannel.connect(remoteAddress);
+                }
+            });
+        } catch (PrivilegedActionException e) {
+            throw (IOException) e.getCause();
+        }
```
normanmaurer: same comment as in the util class. That said I think the "end-goal" would be to not these methods here and just use the util class, correct @tbrooks8  ?

transport/src/main/java/io/netty/channel/socket/nio/NioServerSocketChannel.java
normanmaurer: false positive

common/src/main/java/io/netty/util/SocketUtils.java
normanmaurer: @jasontedor ah thanks! Learned something new :)

transport-udt/src/main/java/io/netty/channel/udt/nio/NioUdtByteConnectorChannel.java
s1monw: @normanmaurer you should consider using something like https://github.com/policeman-tools/forbidden-apis but I am afraid your codebase might not be ready for this yet.

common/src/main/java/io/netty/util/SocketUtils.java
s1monw: I think this class deserves some javadocs why it's used etc.

s1monw:
> I had issues with the NioUdtByteConnectorChannel class in transport-udt. Even though it appears to me that SocketChannelUDT extends SocketChannel properly, when I ran the test suite using the SocketUtils class I introduced, I was seeing test failures due to the following exception:

I wonder if that is reproducible? maybe you just had not cleaned an old build or something?

jasontedor:
Since you mentioned `accept`, I [benchmarked](https://gist.github.com/jasontedor/0a6171b92d9cf422e7fd0a0b8d7a473e#file-acceptsmbenchmark-java) it (please, feel free to critique).

Here are the [results](https://gist.github.com/jasontedor/0a6171b92d9cf422e7fd0a0b8d7a473e#file-results):

```
# Run complete. Total time: 00:04:03

Benchmark                                    Mode  Cnt       Score      Error  Units
AcceptSMBenchmark.testWithSM                  thrpt   30   92432.541 ± 2268.763  ops/s
AcceptSMBenchmark.testWithSMWithNullCheck     thrpt   30   92470.430 ± 1443.234  ops/s
AcceptSMBenchmark.testWithoutSM               thrpt   30  112860.461 ± 1166.055  ops/s
AcceptSMBenchmark.testWithoutSMWithNullCheck  thrpt   30  116936.804 ± 3027.707  ops/s
```

Explanation of the methods:

`testWithSM`: `accept` with a security manager installed, wrapped in a do privileged call, no security manager guard
`testWithSMWithNullCheck`: `accept` with a security manager installed, wrapped in a do privileged call, with security manager guard
`testWithoutSM`: `accept` with no security manager installed, wrapped in a do privileged call, no security manager guard
`AcceptSMBenchmark.testWithoutSMWithNullCheck`: `accept` with no security manager installed, wrapped in a do privileged call, with security manager guard

In each of these methods, we open a channel, bind it, set it to non-blocking, accept, and then close the channel. The immediate argument is wait, you should lift the setup for the channel out of the benchmarked methods. The problem there is that each invocation needs a new channel and then too much time is spent setting up each invocation (see the [warning](http://hg.openjdk.java.net/code-tools/jmh/file/27c92e77d4fd/jmh-core/src/main/java/org/openjdk/jmh/annotations/Level.java#l52) in the JMH source).

The two rows to compare are `testWithoutSM` and `testWithoutSMWithNullCheck` as that shows the difference between adding the guard and not adding the guard for the case where there is no security manager installed. The difference here is 3.6%, but within the margin of error.

> So to sum it up, can we just demonstrate its not an "big overhead" via a JMH benchmark and then move on ?

I have no idea if that qualifies as "big overhead" or not (since it's within the margin of error, I would think not) but the code and results are there for you to judge.

transport-udt/src/main/java/io/netty/channel/udt/nio/NioUdtByteConnectorChannel.java
normanmaurer: @s1monw thanks for the pointer... And yes I am afraid as well :)

normanmaurer:
@jasontedor thanks a lot of putting this together. As the overhead seems to be very small I am ok with just add the extra privileged actions like done in the PR atm. @Scottmitch @nmittler fine by you guys as well ?

nmittler:
@normanmaurer sgtm

tbrooks8:
So at this point I will modify the patch with the following changes:
- move SocketUtils to internal package.
- Add documentation to SocketUtils
- Fix year on the license.

Let me know if I am forgetting anything. I'll also take another look at the AbstractMethodError.

Scottmitch:
@jasontedor - Thanks for the benchmark! Also the case where we don't consider the security manager at all may be interesting (what we currently do). This would avoid a volatile read on the security manager itself. Would you mind contributing the benchmarks in a PR, or including it in this PR so we can retain this context?

A concern that I have is that these JMH benchmarks may not paint the full picture of impact on GC for a long running application where these methods are on the hot path. However since the code is centralized it should be easy enough to add the `null` checks if this becomes an issue.

I guess you guys are not interested in the EPOLL JNI transport? Note I have a PR https://github.com/netty/netty/pull/6000 which makes a bunch of changes in this area that we plan to pull in after the 4.1.7 release. You may want to wait until after these changes land to tackle the JNI stuff.

@normanmaurer - Lets pull this in after the 4.1.7 release?

normanmaurer:
@Scottmitch yes after the release

common/src/main/java/io/netty/util/internal/SocketUtils.java
```diff
@@ -0,0 +1,163 @@
...
+                    return null;
+                }
+            });
+        } catch (PrivilegedActionException e) {
+            throw (IOException) e.getCause();
+        }
+    }
+
+    public static boolean connect(final SocketChannel socketChannel, final SocketAddress remoteAddress)
+            throws Exception {
```
normanmaurer: Good catch

transport-udt/src/main/java/io/netty/channel/udt/nio/NioUdtByteConnectorChannel.java
```diff
@@ -186,4 +190,33 @@ public InetSocketAddress localAddress() {
...
+                    return null;
+                }
+            });
+        } catch (PrivilegedActionException e) {
+            throw (IOException) e.getCause();
+        }
+    }
+
+    private static boolean privilegedConnect(final SocketChannelUDT socketChannel, final SocketAddress remoteAddress)
+            throws Exception {
```
fenik17: throws IOException

common/src/main/java/io/netty/util/internal/SocketUtils.java
```diff
@@ -0,0 +1,163 @@
...
+                    return InetAddress.getAllByName(hostname);
+                }
+            });
+        } catch (PrivilegedActionException e) {
+            throw (UnknownHostException) e.getCause();
+        }
+    }
+
+    public static InetSocketAddress socketAddress(final String hostname, final int port) {
+
```
normanmaurer: @tbrooks8 nit: remove empty line

transport/src/main/java/io/netty/channel/socket/oio/OioSocketChannel.java
```diff
@@ -267,19 +268,19 @@ protected SocketAddress remoteAddress0() {
...
     @Override
     protected void doBind(SocketAddress localAddress) throws Exception {
-        socket.bind(localAddress);
+        SocketUtils.bind(socket, localAddress);
     }
 
     @Override
     protected void doConnect(SocketAddress remoteAddress,
-            SocketAddress localAddress) throws Exception {
+                             SocketAddress localAddress) throws Exception {
```
normanmaurer: revert formatting change ?

transport-udt/src/main/java/io/netty/channel/udt/nio/NioUdtByteConnectorChannel.java
```diff
@@ -186,4 +190,33 @@ public InetSocketAddress localAddress() {
     public InetSocketAddress remoteAddress() {
         return (InetSocketAddress) super.remoteAddress();
     }
+
+    private static void privilegedBind(final SocketChannelUDT socketChannel, final SocketAddress localAddress)
```
normanmaurer: @tbrooks8 so you did not find out why this can't use `SocketUtil` ?

tbrooks8:
I updated the PR with the requested changes.

In regards to the AbstractMethodError:

I looks like com.barchart.udt.barchart-udt-bundle (2.3.0) was built for Java 6. The SocketChannel channel bind(SocketAddress) method was not added until Java 1.7. 

```
     *
     * @since 1.7
     */
    @Override
    public abstract SocketChannel bind(SocketAddress local) throws IOException;
```

So even though SocketChannelUDT has an identical bind(SocketAddress) method, that method does not "override" the abstract method. 

With that discovery, I was able to move the SocketChannelUDT [connect](https://github.com/netty/netty/pull/6146/commits/15093e6dd00b16936345cfa89d5864dba346a217#diff-ffc1de1886a40df6d1187198e1b64079R111) functionality to the SocketUtil class. I did not make a specific bind method for SocketChannelUDT in SocketUtils because the common package does not depend on com.barchart.udt.barchart-udt-bundle.

tbrooks8:
@Scottmitch I will update the benchmark (hopefully) later today.

transport-udt/src/main/java/io/netty/channel/udt/nio/NioUdtByteConnectorChannel.java
tbrooks8: As I mentioned in my comment below, I believe bind will have to stay here until com.barchart.udt.barchart-udt-bundle is released built on Java 7.

tbrooks8:
I updated the [benchmark](https://gist.github.com/tbrooks8/cdbec029fe0cb5d278f186aed488e397) with two scenarios (installed SM and no SM) without doPrivileged or null checks.

I ran them on my MacBook (MacOS) and saw no noticeable difference from the null check case. I do not have access to a linux instance or machine at the moment which would give better numbers for normal use cases.

@jasontedor Any change you might have some time at some point to rerun the benchmarks on Linux with the two additional scenarios?

Scottmitch:
@tbrooks8 - Can you include the benchmark as part of this PR? Put it in the micro bench submodule [1]. You can extend `AbstractMicrobenchmark` which is what our other benchmarks do.

[1] https://github.com/netty/netty/tree/4.1/microbench/src/main/java/io/netty/microbench

tbrooks8:
@Scottmitch I added the benchmark. Let me know if that looks alright.

Scottmitch:
@tbrooks8 - Thanks! Do you have any results now that `testWithoutSMNoPrivileged` has been added?

tbrooks8:
I wrapped three more methods that needed to be wrapped (getLocalSocketAddress, getHardwareAddress, and getInetAddresses). Additionally I fixed the check style issues.

When I ran the benchmarks locally on my Mac this is what I got:

```
Benchmark                                                        Mode  Cnt      Score      Error  Units
PrivilegedSocketOperationsBenchmark.testWithSM                  thrpt   20  35886.163 ± 4292.334  ops/s
PrivilegedSocketOperationsBenchmark.testWithSMNoPrivileged      thrpt   20  35363.158 ± 4030.359  ops/s
PrivilegedSocketOperationsBenchmark.testWithSMWithNullCheck     thrpt   20  36120.718 ± 4791.163  ops/s
PrivilegedSocketOperationsBenchmark.testWithoutSM               thrpt   20  39674.314 ± 7910.967  ops/s
PrivilegedSocketOperationsBenchmark.testWithoutSMNoPrivileged   thrpt   20  41494.852 ± 8808.090  ops/s
PrivilegedSocketOperationsBenchmark.testWithoutSMWithNullCheck  thrpt   20  40717.227 ± 8043.291  ops/s
```

It will take some more time for me to get Linux numbers.

jasontedor:
> Any chance you might have some time at some point to rerun the benchmarks on Linux with the two additional scenarios?

Sure, after the holidays. :smile:

transport-udt/src/main/java/io/netty/channel/udt/nio/NioUdtByteConnectorChannel.java
jasontedor: It's not ready for it, but I've introduced it in #6224 which is a first step towards getting ready for it. 😇

tbrooks8:
I think @jasontedor is going to get to benchmarking on linux pretty soon. Is that the only component that is blocking this PR? Or does something else need to be done?

normanmaurer:
@tbrooks8 I think thats it 👍

jasontedor:
Here are the requested benchmarks:

```
# Run complete. Total time: 00:06:05

Benchmark                                      Mode  Cnt       Score      Error  Units
AcceptSMBenchmark.testWithSM                  thrpt   30   82825.252 ± 3188.193  ops/s
AcceptSMBenchmark.testWithSMNoPrivileged      thrpt   30   83136.388 ± 1187.034  ops/s
AcceptSMBenchmark.testWithSMWithNullCheck     thrpt   30   82036.347 ± 2492.052  ops/s
AcceptSMBenchmark.testWithoutSM               thrpt   30  100239.590 ± 2012.836  ops/s
AcceptSMBenchmark.testWithoutSMNoPrivileged   thrpt   30  106986.437 ± 4307.714  ops/s
AcceptSMBenchmark.testWithoutSMWithNullCheck  thrpt   30  102874.428 ± 3743.800  ops/s
```

Scottmitch:
thanks @jasontedor!

@tbrooks8 - can you rebase/squash and we can pull this in?

tbrooks8:
@Scottmitch Done.

normanmaurer:
let me take care....

jasontedor:
Thanks @normanmaurer and @Scottmitch.

normanmaurer:
Cherry-picked into 4.1 (3344cd21acc61819c327976ed5281daa49b0c05e) and 4.0 (095be39).

Thanks again @tbrooks8, @s1monw  and @jasontedor for your patience

microbench/src/main/java/io/netty/microbench/internal/PrivilegedSocketOperationsBenchmark.java
```diff
@@ -0,0 +1,193 @@
...
+
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.SECONDS)
+public class PrivilegedSocketOperationsBenchmark extends AbstractMicrobenchmark {
+
+    @State(Scope.Benchmark)
+    public static class SecurityManagerInstalled {
+
+        @Setup
+        public void setup() throws IOException, NoSuchAlgorithmException, URISyntaxException {
```
netkins: MAJOR Refactor this method to throw at most one checked exception instead of: java.io.IOException, java.security.NoSuchAlgorithmException, java.net.URISyntaxException [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1160)

microbench/src/main/java/io/netty/microbench/internal/PrivilegedSocketOperationsBenchmark.java
```diff
@@ -0,0 +1,193 @@
...
+        final ServerSocketChannel ssc = ServerSocketChannel.open();
+        ssc.socket().bind(null);
+        ssc.configureBlocking(false);
+        ssc.accept();
+        ssc.close();
+        return ssc;
+    }
+
+    @Benchmark
+    public ServerSocketChannel testWithSM(final SecurityManagerInstalled sm) throws IOException {
```
netkins: MAJOR Remove this unused method parameter "sm". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

microbench/src/main/java/io/netty/microbench/internal/PrivilegedSocketOperationsBenchmark.java
```diff
@@ -0,0 +1,193 @@
...
+        public void tearDown() throws IOException {
+            System.setSecurityManager(null);
+        }
+    }
+
+    @State(Scope.Benchmark)
+    public static class SecurityManagerEmpty {
+
+        @Setup
+        public void setup() throws IOException, NoSuchAlgorithmException, URISyntaxException {
```
netkins: MAJOR Refactor this method to throw at most one checked exception instead of: java.io.IOException, java.security.NoSuchAlgorithmException, java.net.URISyntaxException [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1160)

microbench/src/main/java/io/netty/microbench/internal/PrivilegedSocketOperationsBenchmark.java
```diff
@@ -0,0 +1,193 @@
...
+                    });
+            ssc.close();
+            return ssc;
+        } catch (final PrivilegedActionException e) {
+            throw (IOException) e.getCause();
+        }
+    }
+
+    @Benchmark
+    public ServerSocketChannel testWithoutSMWithNullCheck(final SecurityManagerEmpty sm) throws IOException {
```
netkins: MAJOR Remove this unused method parameter "sm". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

microbench/src/main/java/io/netty/microbench/internal/PrivilegedSocketOperationsBenchmark.java
```diff
@@ -0,0 +1,193 @@
...
+            ssc.socket().bind(null);
+            ssc.configureBlocking(false);
+            ssc.accept();
+            ssc.close();
+            return ssc;
+        }
+    }
+
+    @Benchmark
+    public ServerSocketChannel testWithoutSMNoPrivileged(final SecurityManagerEmpty sm) throws IOException {
```
netkins: MAJOR Remove this unused method parameter "sm". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

microbench/src/main/java/io/netty/microbench/internal/PrivilegedSocketOperationsBenchmark.java
```diff
@@ -0,0 +1,193 @@
...
+                    });
+            ssc.close();
+            return ssc;
+        } catch (final PrivilegedActionException e) {
+            throw (IOException) e.getCause();
+        }
+    }
+
+    @Benchmark
+    public ServerSocketChannel testWithSMWithNullCheck(final SecurityManagerInstalled sm) throws IOException {
```
netkins: MAJOR Remove this unused method parameter "sm". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

microbench/src/main/java/io/netty/microbench/internal/PrivilegedSocketOperationsBenchmark.java
```diff
@@ -0,0 +1,193 @@
...
+    public static class SecurityManagerEmpty {
+
+        @Setup
+        public void setup() throws IOException, NoSuchAlgorithmException, URISyntaxException {
+            System.setSecurityManager(null);
+        }
+    }
+
+    @Benchmark
+    public ServerSocketChannel testWithSMNoPrivileged(final SecurityManagerInstalled sm) throws IOException {
```
netkins: MAJOR Remove this unused method parameter "sm". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

microbench/src/main/java/io/netty/microbench/internal/PrivilegedSocketOperationsBenchmark.java
```diff
@@ -0,0 +1,193 @@
...
+        final ServerSocketChannel ssc = ServerSocketChannel.open();
+        ssc.socket().bind(null);
+        ssc.configureBlocking(false);
+        ssc.accept();
+        ssc.close();
+        return ssc;
+    }
+
+    @Benchmark
+    public ServerSocketChannel testWithoutSM(final SecurityManagerEmpty sm) throws IOException {
```
netkins: MAJOR Remove this unused method parameter "sm". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

microbench/src/main/java/io/netty/microbench/internal/PrivilegedSocketOperationsBenchmark.java
jasontedor: This is terrible advice @netkins.

Issue #6132, opened 2016-12-14
Wrong regular expression to check MACAdresses in DefaultChannelId


jo-kin:
Using Netty on a Java-VM running on IBM z/OS it occurs, that creating a unique channel-id leads to problems.
It seems to me, that this process bases at least on the MAC-Address of networking hardware.  On IBM z/OS this information is not available with the usual JAVA-methods in NetworkInterface.  Netty logs always a warning to System.err running on z/OS that it fell back to a random MAC.
So I decided to use the VM-Option io.netty.machineId=xxxxxxxx to inject here a valid MAC-Address retrieved from non-java resources on z/OS.

**To my regrets it does not work. The problem is not platform-dependent, even with in a SUN-VM running on Windows it is not possible to use this option.** 

Probably the Regular Expression, used to check the input, is not correct.
I would recommend to change the used RegExp 
` private static final Pattern MACHINE_ID_PATTERN = Pattern.compile("^(?:[0-9a-fA-F][:-]?){6,8}$");`
to an expression more like 
`([0-9a-fA-F]{2}[:-]){5,7}[0-9a-fA-F]{2}`

### Expected behavior
DefaultChannelId allows to customize the used MACAddress via System.property.
Using -Dio.netty.machineId=00-14-5E-5A-AF-77 as Java-VM-option should be accepted as valid MACAddress.

### Actual behavior
Netty rejects this MACAddress as malformed:

`Dez 14, 2016 2:10:49 PM io.netty.channel.DefaultChannelId <clinit>
WARNING: -Dio.netty.machineId: 00-14-5E-5A-AF-77 (malformed)
Dez 14, 2016 2:10:49 PM io.netty.handler.logging.LoggingHandler channelRegistered
INFO: [id: 0xa9354b7b] REGISTERED
Dez 14, 2016 2:10:49 PM io.netty.handler.logging.LoggingHandler bind
INFO: [id: 0xa9354b7b] BIND: 0.0.0.0/0.0.0.0:8007
Dez 14, 2016 2:10:49 PM io.netty.handler.logging.LoggingHandler channelActive
INFO: [id: 0xa9354b7b, L:/0:0:0:0:0:0:0:0:8007] ACTIVE`

### Steps to reproduce
[...]

PR #6147, opened 2016-12-20
Fix DefaultChannelId MAC address parsing bug

Scottmitch:
Motivation:
DefaultChannelId provides a regular expression which validates if a user provided MAC address is valid. This regular expression may allow invalid MAC addresses and also not allow valid MAC addresses.

Modifications:
- Introduce a MacAddressUtil#parseMac method which can parse and validate the MAC address at the same time. The regular expression check before hand is additional overhead if we have to parse the MAC address.

Result:
Fixes https://github.com/netty/netty/issues/6132.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/MacAddressUtil.java
+++ b/common/src/main/java/io/netty/util/internal/MacAddressUtil.java
@@ -16,6 +16,10 @@
 
 package io.netty.util.internal;
 
+import io.netty.util.NetUtil;
+import io.netty.util.internal.logging.InternalLogger;
+import io.netty.util.internal.logging.InternalLoggerFactory;
+
 import java.net.InetAddress;
 import java.net.NetworkInterface;
 import java.net.SocketException;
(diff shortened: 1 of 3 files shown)


Scottmitch:
@jo-kin - Please verify this fixes the issue.

normanmaurer:
@Scottmitch there must be something wrong... Please check:

```
io.netty.channel.ChannelException: Unable to create Channel from class class io.netty.channel.local.LocalServerChannel
    at io.netty.bootstrap.BootstrapTest.testAsyncResolutionFailure(BootstrapTest.java:263)
Caused by: java.lang.NoClassDefFoundError: Could not initialize class io.netty.channel.DefaultChannelId
    at io.netty.bootstrap.BootstrapTest.testAsyncResolutionFailure(BootstrapTest.java:263)
```

Scottmitch:
@normanmaurer - issue should be fixed now.

common/src/main/java/io/netty/util/internal/MacAddressUtil.java
```diff
@@ -110,27 +109,90 @@
...
+                bestMacAddr = Arrays.copyOf(bestMacAddr, EUI64_MAC_ADDRESS_LENGTH);
         }
 
         return bestMacAddr;
     }
 
     /**
+     * Returns the result of {@link #bestAvailableMac()} if non-{@code null} otherwise returns a random EUI-64 MAC
+     * address.
+     * @return the result of {@link #bestAvailableMac()} if non-{@code null} otherwise returns a random EUI-64 MAC
```
normanmaurer: @Scottmitch just remove the `@return` as it is the same as what is stated already ?

common/src/main/java/io/netty/util/internal/MacAddressUtil.java
```diff
@@ -110,27 +109,90 @@
...
+        }
+
+        machineId[end] = (byte) Integer.parseInt(value.substring(j, value.length()), 16);
+
+        return machineId;
+    }
+
+    private static void validateMacSeparator(char separator) {
+        if (separator != ':' && separator != '-') {
+            throw new IllegalArgumentException("unsupported seperator: " + separator + " (expected: [:-])");
```
normanmaurer: @Scottmitch `:-]` looks nice and friendly ;)

normanmaurer:
@Scottmitch just one comment.. otherwise LGTM

common/src/main/java/io/netty/util/internal/MacAddressUtil.java
Scottmitch: sure

normanmaurer:
@Scottmitch ship it

Scottmitch:
4.1 (3d11334)

PR #6148, opened 2016-12-20
Add unit test for HttpObjectDecoder with message split on buffer boun…

Scottmitch:
…daries

Motivation:
We should have a unit test which explicitly tests a HTTP message being split between multiple ByteBuf objects.

Modifications:
- Add a unit test to HttpRequestDecoderTest which splits a request between 2 ByteBuf objects

Result:
More unit test coverage for HttpObjectDecoder.


Scottmitch:
4.1 (3f82b53) 4.0 (23e8ff7)

PR #6149, opened 2016-12-20
Ensure we use a MPMC queue in ThreadDeathWatcher as it may be used fr…

normanmaurer:
…om multiple threads at the same time.

Motivation:

We used a MPSC queue in ThreadDeathWatcher and checked if it empty via isEmpty() from multiple threads if very unlucky. Depending on the implementation this is not safe and may even produce things like live-locks.

Modifications:

Change to use a MPMC queue.

Result:

No more risk to run into issues when multiple threads call watch(...) / unwatch(...) concurrently.

### DIFF
--- a/common/src/main/java/io/netty/util/ThreadDeathWatcher.java
+++ b/common/src/main/java/io/netty/util/ThreadDeathWatcher.java
@@ -17,7 +17,6 @@
 package io.netty.util;
 
 import io.netty.util.concurrent.DefaultThreadFactory;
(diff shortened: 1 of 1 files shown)


normanmaurer:
@Scottmitch agree.. Followup PR FTW ;)

normanmaurer:
Cherry-picked into 4.1 (28c39a3) and 4.0 (c7dbb7b6b2f9475f9661d85af402d51a6c4df5c2)

Issue #5831, opened 2016-09-16
HttpServerCodec annot encode a respons e to HEAD request with a 'content-encoding: chunked' header


bryce-anderson:
Netty versions: 4.x
### Context

It is valid to send a response to a HEAD request that contains a `transfer-encoding: chunked` header, but there is no way to do this using the netty4 `HttpServerCodec`.
### Root of the problem

The root cause is that the netty4 `HttpObjectEncoder` will [transition to the state](https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java#L80) `ST_CONTENT_CHUNK` and the only way to transition back to `ST_INIT` is through the `encodeChunkedContent` method which [will write the terminating length](https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java#L163-L180) (`0\r\n\r\n\r\n`), a protocol error when responding to a HEAD request. I don't believe this is a problem with netty3: it looks like submitting a new `HttpMessage` [effectively resets the encoder](https://github.com/netty/netty/blob/3.10/src/main/java/org/jboss/netty/handler/codec/http/HttpMessageEncoder.java#L62-L101), while the same strategy with netty4 [results in an illegal state exception](https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java#L68-L70).
### Workaround

The `content-encoding` MAY be omitted ([RFC 7231](https://tools.ietf.org/html/rfc7231#section-4.3.2)), but it would be nice to have.

normanmaurer:
@bryce-anderson interested to provide a PR + unit test ?

bryce-anderson:
@normanmaurer Do you mean PR to fix or PR to demonstrate? I'm happy to provide a PR either way, I just want to be clear on what PR we want. 😄

If we are talking about a fix, the easy solution is to adopt the netty3 behavior of resetting the encoder on every new `HttpMessage`, but before making the change I think it's worth quickly discussing if that is an acceptable solution: in my minds eye someone introduced that `IllegalStateException` for a reason.

normanmaurer:
let me fix this...

normanmaurer:
@bryce-anderson actually I not understand how this happens. Can you provide a unit test that shows it ?

normanmaurer:
@bryce-anderson nevermind... I think I understand it now

[...]

PR #6151, opened 2016-12-21
[#5831] Fix handling of FullHttpResponse when respond to HEAD in Http…

normanmaurer:
…ServerCodec

Motivation:

cb139043f321836746915e7ada2d507ca45ed734 introduced special handling of response to HEAD requests. Due a bug we failed to handle FullHttpResponse correctly.

Modifications:

Correctly handle FullHttpResponse for HEAD requests.

Result:

Works as expected.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java
@@ -97,12 +97,6 @@ public abstract class HttpObjectEncoder<H extends HttpMessage> extends MessageTo
             switch (state) {
                 case ST_INIT:
                     throw new IllegalStateException("unexpected message type: " + StringUtil.simpleClassName(msg));
-                case ST_CONTENT_ALWAYS_EMPTY:
-                    out.add(EMPTY_BUFFER);
-                    if (msg instanceof LastHttpContent) {
-                        state = ST_INIT;
-                    }
-                    return;
                 case ST_CONTENT_NON_CHUNK:
                     final long contentLength = contentLength(msg);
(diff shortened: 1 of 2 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java
```diff
@@ -116,29 +110,41 @@ protected void encode(ChannelHandlerContext ctx, Object msg, List<Object> out) t
...
+
+                    break;
                 default:
                     throw new Error();
             }
+
+            if (msg instanceof LastHttpContent) {
+                state = ST_INIT;
+            }
         } else {
```
Scottmitch: make this an `else if`

codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java
```diff
@@ -184,8 +190,6 @@ private void encodeChunkedContent(ChannelHandlerContext ctx, Object msg, long co
                 buf.writeBytes(CRLF);
                 out.add(buf);
             }
-
-            state = ST_INIT;
         } else {
```
Scottmitch: make this an `else if`

normanmaurer:
Merged into 4.1 (0eeeb764391490585f690c8b424d9ce2d43e6a9b) and 4.0 (0f069ae3119ffe7741df29f7e431b24786e5ac64)

rkapsi:
@normanmaurer :+1: we just encountered it and can confirm this fixes it. ty!

normanmaurer:
@rkapsi woooot :) thx

PR #6155, opened 2016-12-22
Extends HttpPostRequestEncoder to support all methods except TRACE

### LINKED ISSUE #6138: Netty only support the following 4 http method
default

There are other types like DELETE, HEAD need support.

fredericBregier:
Motivation:

In Netty, currently, the HttpPostRequestEncoder only supports POST, PUT, PATCH and OPTIONS, while the RFC 7231 allows with a warning that GET, HEAD, DELETE and CONNECT use a body too (but not TRACE where it is explicitely not allowed).
The RFC in chapter 4.3 says:
"A payload within a XXX request message has no defined semantics;
sending a payload body on a XXX request might cause some existing
implementations to reject the request."
where XXX can be replaced by one of GET, HEAD, DELETE or CONNECT.

Current usages, on particular in REST mode, tend to use those extra HttpMethods for such queries.

So this PR proposes to remove the current restrictions, leaving only TRACE as explicitely not supported.

Modification:

In the constructor, where the test is done, replacing all by checking only against TRACE, and adding one test to check that all methods are supported or not.

Result:

Fixes #6138.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostRequestEncoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostRequestEncoder.java
@@ -53,6 +53,15 @@ import static io.netty.buffer.Unpooled.wrappedBuffer;
 
 /**
  * This encoder will help to encode Request for a FORM as POST.
+ *
+ * <P>According to RFC 7231, POST, PUT and OPTIONS allow to have a body.
+ * This encoder will support widely all methods except TRACE since the RFC notes
+ * for GET, DELETE, HEAD and CONNECT: (replaces XXX by one of these methods)</P>
+ * <P>"A payload within a XXX request message has no defined semantics;
+ * sending a payload body on a XXX request might cause some existing
+ * implementations to reject the request."</P>
+ * <P>On the contrary, for TRACE method, RFC says:</P>
+ * <P>"A client MUST NOT send a message body in a TRACE request."</P>
  */
 public class HttpPostRequestEncoder implements ChunkedInput<HttpContent> {
 
@@ -150,7 +159,7 @@ public class HttpPostRequestEncoder implements ChunkedInput<HttpContent> {
      * @throws NullPointerException
      *             for request
      * @throws ErrorDataEncoderException
-     *             if the request is not a POST
+     *             if the request is a TRACE
      */
     public HttpPostRequestEncoder(HttpRequest request, boolean multipart) throws ErrorDataEncoderException {
         this(new DefaultHttpDataFactory(DefaultHttpDataFactory.MINSIZE), request, multipart,
@@ -168,7 +177,7 @@ public class HttpPostRequestEncoder implements ChunkedInput<HttpContent> {
      * @throws NullPointerException
      *             for request and factory
      * @throws ErrorDataEncoderException
-     *             if the request is not a POST
+     *             if the request is a TRACE
      */
     public HttpPostRequestEncoder(HttpDataFactory factory, HttpRequest request, boolean multipart)
             throws ErrorDataEncoderException {
@@ -190,7 +199,7 @@ public class HttpPostRequestEncoder implements ChunkedInput<HttpContent> {
      * @throws NullPointerException
      *             for request or charset or factory
      * @throws ErrorDataEncoderException
-     *             if the request is not a POST
+     *             if the request is a TRACE
      */
     public HttpPostRequestEncoder(
             HttpDataFactory factory, HttpRequest request, boolean multipart, Charset charset,
@@ -206,9 +215,8 @@ public class HttpPostRequestEncoder implements ChunkedInput<HttpContent> {
             throw new NullPointerException("charset");
         }
         HttpMethod method = request.method();
-        if (!(method.equals(HttpMethod.POST) || method.equals(HttpMethod.PUT)
-                || method.equals(HttpMethod.PATCH) || method.equals(HttpMethod.OPTIONS))) {
-            throw new ErrorDataEncoderException("Cannot create a Encoder if not a POST");
+        if (method.equals(HttpMethod.TRACE)) {
+            throw new ErrorDataEncoderException("Cannot create a Encoder if request is a TRACE");
         }
         this.request = request;
         this.charset = charset;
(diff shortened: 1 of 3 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostRequestEncoder.java
```diff
@@ -206,9 +215,8 @@ public HttpPostRequestEncoder(
             throw new NullPointerException("charset");
         }
         HttpMethod method = request.method();
-        if (!(method.equals(HttpMethod.POST) || method.equals(HttpMethod.PUT)
-                || method.equals(HttpMethod.PATCH) || method.equals(HttpMethod.OPTIONS))) {
-            throw new ErrorDataEncoderException("Cannot create a Encoder if not a POST");
+        if (method.equals(HttpMethod.TRACE)) {
+            throw new ErrorDataEncoderException("Cannot create a Encoder if request is a TRACE");
```
Scottmitch: any reason why this is not `IllegalArgumentException`? We are not necessarily encoding data here.

codec-http/src/main/java/io/netty/handler/codec/http/multipart/InterfaceHttpPostRequestDecoder.java
```diff
@@ -20,7 +20,8 @@
 import java.util.List;
 
 /**
- * This decoder will decode Body and can handle POST BODY (or for PUT, PATCH or OPTIONS).
+ * This decoder will decode Body and can handle POST BODY (or for PUT, PATCH or OPTIONS or any allowed in RFC 7231
+ * except TRACE).
```
Scottmitch: Consider rewording this sentence. The way it is worded makes it sound like TRACE is allowed in RFC 7231 but we just don't support it.

codec-http/src/test/java/io/netty/handler/codec/http/multipart/HttpPostRequestEncoderTest.java
```diff
@@ -75,6 +77,56 @@ public void testSingleFileUpload() throws Exception {
...
+
+        HttpPostRequestEncoder encoder = new HttpPostRequestEncoder(request, true);
+        File file1 = new File(getClass().getResource("/file-01.txt").toURI());
+        encoder.addBodyAttribute("foo", "bar");
+        encoder.addBodyFileUpload("quux", file1, "text/plain", false);
+
+        String multipartDataBoundary = encoder.multipartDataBoundary;
+        String content = getRequestBody(encoder);
+
+        String expected = "--" + multipartDataBoundary + "\r\n" +
```
Scottmitch: can we share this code with `testSingleFileUpload`? Seems like there is some duplication here.

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostRequestEncoder.java
fredericBregier: It was the original exception. I'm not against changing it to illegal but does it change the contract?

codec-http/src/main/java/io/netty/handler/codec/http/multipart/InterfaceHttpPostRequestDecoder.java
fredericBregier: Ok will propose something more concise
 "(or any methods allowed in RFC 7231, therefore preventing TRACE)"

codec-http/src/test/java/io/netty/handler/codec/http/multipart/HttpPostRequestEncoderTest.java
fredericBregier: Yes, we can even removes the first one since included in the added test

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostRequestEncoder.java
fredericBregier: Wdyt?

codec-http/src/test/java/io/netty/handler/codec/http/multipart/HttpPostRequestEncoderTest.java
fredericBregier: Done

codec-http/src/main/java/io/netty/handler/codec/http/multipart/InterfaceHttpPostRequestDecoder.java
fredericBregier: Since Decoder has no restriction, I just remove the related comments.

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostRequestEncoder.java
fredericBregier: For the moment, I let it like that, since I feel, maybe wrongly, that changing the signature / behaviour here (IllegalArgumentException instead of ErrorDataEncoerException as originally) may lead to incorrect behavior in end-user code if they were depending on this exception, as noted in the Javadoc. It seems part of the contract. And so changing it will lead to change the API contract.

I'm ready to change it however, but @Scottmitch WDYT ?

fredericBregier:
Build issue seems unrelated

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostRequestEncoder.java
Scottmitch: Thanks for clarifying. Seems like it has been this way for a while. Fine to leave as is for this PR.

fredericBregier:
Strange , it seems sonar is not responding?
@Scottmitch by the way, thanks for the review...
I will wait for another build .  is it the only one build having this issue ?

fredericBregier:
I just take a look at other builds, they do have the same issue

Scottmitch:
@fredericBregier - I reported https://github.com/netty/netty/issues/6161 for the CI issue. Feel free to cherry-pick this ... or I can do it if you prefer.

fredericBregier:
@Scottmitch Yes, please, go ahead to cherry-pick. I'm not 100% aware on Netty's usages, even if I try to follow as much as possible. 
Moreover I do not know if this should be backported to 4.0.

fredericBregier:
I take a look at 4.0 , it seems it is even more restrictive (line 190) where only POST is allowed so IMHO no need to backport to 4.0

Scottmitch:
4.1 (56ddc47) 4.0 (bc4fbd6)

Scottmitch:
@fredericBregier - I think that POST limitation was just due to forgetting to back port other changes. I don't see a reason why there should be a difference in behavior between versions here.

Thanks for the PR!

PR #6158, opened 2016-12-27
OCSP stapling support for Netty using netty-tcnative.

rkapsi:
OCSP stapling support for Netty using netty-tcnative.

https://github.com/netty/netty-tcnative/pull/215

Motivation

OCSP stapling (formally known as TLS Certificate Status Request extension) is alternative approach for checking the revocation status of X.509 Certificates. Servers can preemptively fetch the OCSP response from the CA's responder, cache it for some period of time, and pass it along during (a.k.a. staple) the TLS handshake. The client no longer has to reach out on its own to the CA to check the validity of a cetitficate. Some of the key benefits are:

1) Speed. The client doesn't have to crosscheck the certificate.
2) Efficiency. The Internet is no longer DDoS'ing the CA's OCSP responder servers.
3) Safety. Less operational dependence on the CA. Certificate owners can sustain short CA outages.
4) Privacy. The CA can lo longer track the users of a certificate.

https://en.wikipedia.org/wiki/OCSP_stapling
https://letsencrypt.org/2016/10/24/squarespace-ocsp-impl.html

Modifications

https://www.openssl.org/docs/man1.0.2/ssl/SSL_set_tlsext_status_type.html

Result

High-level API to enable OCSP stapling


rkapsi:
Build fails because this change depends on a version of netty-tcnative that isn't released yet: https://github.com/netty/netty-tcnative/pull/215

I was planning to include also an example for how to pull out the OCSP responder URL from the X509 Certificate and retrieve the OCSP staple from the CA but it'll require Bouncycastle.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -15,22 +15,7 @@
...
-import io.netty.util.internal.PlatformDependent;
-import io.netty.util.internal.StringUtil;
-import io.netty.util.internal.SystemPropertyUtil;
-import io.netty.util.internal.logging.InternalLogger;
-import io.netty.util.internal.logging.InternalLoggerFactory;
-import org.apache.tomcat.jni.CertificateVerifier;
-import org.apache.tomcat.jni.Pool;
-import org.apache.tomcat.jni.SSL;
-import org.apache.tomcat.jni.SSLContext;
+import static io.netty.util.internal.ObjectUtil.checkNotNull;
```
Scottmitch: Maybe we should add this to the style guide but can you keep the statics at the bottom, and preserve the previous import order?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -803,4 +865,58 @@ private static long newBIO(ByteBuf buffer) throws Exception {
             buffer.release();
         }
     }
+
+    private interface OcspCallbackExt extends OcspCallback {
+        OcspContext context();
+    }
+
+    private static class ClientOcspCallback implements OcspCallbackExt {
```
Scottmitch: `final`

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -803,4 +865,58 @@ private static long newBIO(ByteBuf buffer) throws Exception {
...
+
+        @Override
+        public int callback(long ssl) throws Exception {
+            byte[] staple = SSL.getOcspStaple(ssl);
+            boolean good = context.verify(staple);
+            return good ? ACK : NACK;
+        }
+    }
+
+    private static class ServerOcspCallback implements OcspCallbackExt {
```
Scottmitch: `final`

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -287,6 +289,18 @@ protected void deallocate() {
         keyMaterialManager = context.keyMaterialManager();
     }
 
+    /**
+     * Enables (or disables) OCSP stapling on the {@link SSLEngine}.
+     */
+    @UnstableApi
+    public synchronized boolean enableOcsp(boolean enabled) {
+        if (ssl != 0L) {
+            return SSL.setStatusType(ssl, enabled ? SSL.TLSEXT_STATUSTYPE_ocsp : 0);
```
Scottmitch: should `0` be `TLSEXT_STATUSTYPE_host_name`?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -287,6 +289,18 @@ protected void deallocate() {
         keyMaterialManager = context.keyMaterialManager();
     }
 
+    /**
+     * Enables (or disables) OCSP stapling on the {@link SSLEngine}.
+     */
+    @UnstableApi
+    public synchronized boolean enableOcsp(boolean enabled) {
+        if (ssl != 0L) {
```
Scottmitch: `SSL.setStatusType` protects against null `SSL*` too ... can we just rely on the null there instead of doing it in both places?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -15,22 +15,18 @@
...
-import io.netty.util.ResourceLeakTracker;
-import io.netty.util.internal.EmptyArrays;
-import io.netty.util.internal.PlatformDependent;
-import io.netty.util.internal.StringUtil;
-import io.netty.util.internal.ThrowableUtil;
-import io.netty.util.internal.logging.InternalLogger;
-import io.netty.util.internal.logging.InternalLoggerFactory;
-import org.apache.tomcat.jni.Buffer;
-import org.apache.tomcat.jni.SSL;
+import static io.netty.handler.ssl.OpenSsl.memoryAddress;
```
Scottmitch: Maybe we should add this to the style guide but can you keep the statics at the bottom, and preserve the previous import order?

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,471 @@
...
+
+public class OcspTest {
+
+    /**
+     * The Server provides an OCSP staple and the Client accepts it.
+     */
+    @Test(timeout = 10000L)
+    public void testClientAcceptingOcspStaple() throws Exception {
+
+        assumeTrue(OpenSsl.isOcspSupported());
```
Scottmitch: you could also move this to a `@Before` or `@BeforeClass`

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,471 @@
...
+        TestClientOcspContext clientOcspContext = new TestClientOcspContext(true);
+
+        handshake(latch, serverOcspContext, clientOcspContext, serverHandler, clientHandler);
+
+        byte[] expected = serverOcspContext.staple();
+        byte[] actual = clientOcspContext.staple();
+
+        assertNotNull(actual);
+        assertNotSame(serverOcspContext.staple(), actual);
+        assertTrue(Arrays.equals(expected, actual));
```
Scottmitch: `assertArrayEquals`

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,471 @@
...
+
+    /**
+     * The Server provides an OCSP staple and the Client rejects it.
+     */
+    @Test(timeout = 10000L)
+    public void testClientRejectingOcspStaple() throws Exception {
+
+        assumeTrue(OpenSsl.isOcspSupported());
+
+        final Throwable[] clientCauseRef = new Throwable[] { null };
```
Scottmitch: consider just using AtomicReference because you access this across different threads.

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,471 @@
...
+        TestClientOcspContext clientOcspContext = new TestClientOcspContext(false);
+
+        handshake(latch, serverOcspContext, clientOcspContext, serverHandler, clientHandler);
+
+        byte[] expected = serverOcspContext.staple();
+        byte[] actual = clientOcspContext.staple();
+
+        assertNotNull(actual);
+        assertNotSame(serverOcspContext.staple(), actual);
+        assertTrue(Arrays.equals(expected, actual));
```
Scottmitch: `assertArrayEquals`

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,471 @@
...
+        ServerOcspContext serverOcspContext = newServerOcspContext();
+        TestClientOcspContext clientOcspContext = new TestClientOcspContext(true);
+
+        handshake(latch, serverOcspContext, clientOcspContext, serverHandler, clientHandler);
+
+        byte[] expected = serverOcspContext.staple();
+        byte[] actual = clientOcspContext.staple();
+
+        assertNotNull(actual);
+        assertNotSame(serverOcspContext.staple(), actual);
```
Scottmitch: `serverOcspContext.staple()` -> `expected`

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,471 @@
...
+        assertTrue(Arrays.equals(expected, actual));
+
+        // If the client rejects the staple then Client and Server will fail with
+        // Exceptions which contain these respective messages.
+        String clientMessage = clientCauseRef[0].getMessage();
+        assertTrue(clientMessage, clientMessage.contains("invalid status response"));
```
Scottmitch: can we avoid depending upon exception text in tests?

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,471 @@
...
+            public boolean verify(byte[] staple) throws Exception {
+                throw new IllegalStateException();
+            }
+        };
+
+        handshake(latch, serverOcspContext, clientOcspContext, null, clientHandler);
+
+        // ATTENTION: This one is strange! It's a DecoderException caused by a NPE.
+        Throwable cause = causeRef[0];
+        assertTrue(cause instanceof DecoderException);
```
Scottmitch: should we instead unwrap and verify the NPE?

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,471 @@
...
+    }
+
+    private static void handshake(CountDownLatch latch, ServerOcspContext serverOcspContext,
+            ClientOcspContext clientOcspContext, ChannelHandler serverHandler,
+            ChannelHandler clientHandler) throws Exception {
+
+        SelfSignedCertificate ssc = new SelfSignedCertificate();
+        try {
+            SslContext serverSslContext = SslContextBuilder.forServer(ssc.certificate(), ssc.privateKey())
+                    .sslProvider(SslProvider.OPENSSL)
```
Scottmitch: can we run this test with `OPENSSL_REFCNT` too? We need to be careful about releasing in this case of course.

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,471 @@
...
+                try {
+                    if (!latch.await(10L, TimeUnit.SECONDS)) {
+                        throw new IllegalStateException("Something went wrong.");
+                    }
+                } finally {
+                    client.close();
+                    server.close();
+                }
+            } finally {
+                group.shutdownGracefully();
```
Scottmitch: consider specifying the timeout values here so the unit tests don't have to wait for the quite period etc...

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,471 @@
...
+                    .build();
+            ocspfySslContext(clientSslContext, clientOcspContext);
+
+            EventLoopGroup group = new DefaultEventLoopGroup();
+            try {
+                LocalAddress address = new LocalAddress("handshake-" + Math.random());
+                Channel server = newServer(group, address, serverSslContext, serverHandler);
+                Channel client = newClient(group, address, clientSslContext, clientHandler);
+                try {
+                    if (!latch.await(10L, TimeUnit.SECONDS)) {
```
Scottmitch: can we just do `assertTrue`?

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,471 @@
...
+        private final boolean valid;
+
+        private volatile byte[] staple;
+
+        public TestClientOcspContext(boolean valid) {
+            this.valid = valid;
+        }
+
+        public byte[] staple() throws InterruptedException, TimeoutException {
+            if (!latch.await(10L, TimeUnit.SECONDS)) {
```
Scottmitch: this is expected to be called in the junit thread right, if so can we just do `assertTrue`?

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,550 @@
...
+    public void testClientExceptionOpenSslRefCnt() throws Exception {
+        testClientException(SslProvider.OPENSSL_REFCNT);
+    }
+
+    /**
+     * Testing what happens if the {@link ClientOcspContext} throws an {@link Exception}.
+     * 
+     * The exception should bubble up on the client side and the connection should get closed.
+     */
+    private void testClientException(SslProvider sslProvider) throws Exception {
```
rkapsi: This test and `testServerException` produce the NPE described in https://github.com/netty/netty/issues/6336.

OpenSSL calls the `verify()` method (via the OCSP callback - see PR in netty-tcnative) and this test throws purposely an IllegalStateException to see what happens.

I'm expecting a DecoderException to bubble up (with the IllegalStateException as cause) but instead I'm getting DecoderException that was caused by a NPE.

rkapsi:
@Scottmitch I've added some example code to the PR. The client side is pretty simple but likely not very useful. The server side is more interesting as it involves pulling out the CA's OCSP responder URL from the X509 Certificate and then constructing a special POST request to obtain the OCSP staple from the said URL. It's a somewhat involved process and there isn't much information about it on the Intertubes. I tried to keep the examples as simple as possible and they're not meant to be an example how do run this code in production.


I used Wikipedia's Webserver for the client example because the CDN that is fronting netty.io doesn't support OCSP. And I used netty.io's X509 Certificate (Comodo) for the server example but it's incomplete because I don't have the PrivateKey for it.

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,551 @@
...
+        final CountDownLatch latch = new CountDownLatch(2);
+
+        ChannelInboundHandlerAdapter serverHandler = new ChannelInboundHandlerAdapter() {
+            @Override
+            public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
+                // NOTE: Expecting this Throwable to be a DecoderException that
+                // was caused by an IllegalStateException from a few lines further
+                // down but it's instead a NPE.
+                //
+                // https://github.com/netty/netty/issues/6336
```
Scottmitch: can we remove this and any associated work around code now?

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,551 @@
...
+                    latch.countDown();
+                }
+            }
+        };
+
+        ChannelInboundHandlerAdapter clientHandler = new ChannelInboundHandlerAdapter() {
+            @Override
+            public void channelInactive(ChannelHandlerContext ctx)
+                    throws Exception {
+
```
Scottmitch: nit: kill line

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,551 @@
...
+                } finally {
+                    latch.countDown();
+                }
+            }
+        };
+
+        ChannelInboundHandlerAdapter clientHandler = new ChannelInboundHandlerAdapter() {
+            @Override
+            public void channelInactive(ChannelHandlerContext ctx)
+                    throws Exception {
```
Scottmitch: nit: put on the previous line?

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,551 @@
...
+                .channel();
+    }
+
+    private static ChannelHandler newHandler(final SslContext context, final ChannelHandler handler) {
+        return new ChannelInitializer<Channel>() {
+            @Override
+            protected void initChannel(Channel ch) throws Exception {
+                ChannelPipeline pipeline = ch.pipeline();
+
+                SSLEngine engine = context.newEngine(ch.alloc());
```
Scottmitch: consider using `context.newHandler(ch.alloc())` instead. This may produce some different behavior in the future (PR https://github.com/netty/netty/pull/6365) and I would like to make sure it still works at that time.

example/src/main/java/io/netty/example/ocsp/OcspClientExample.java
```diff
@@ -0,0 +1,224 @@
...
+                    if (response.getStatus() != OCSPResponseStatus.SUCCESSFUL) {
+                        return false;
+                    }
+
+                    BasicOCSPResp basicResponse = (BasicOCSPResp) response.getResponseObject();
+                    SingleResp first = basicResponse.getResponses()[0];
+
+                    CertificateStatus status = first.getCertStatus();
+                    System.out.println(host);
+                    System.out.println("  Status: " + (status == CertificateStatus.GOOD ? "Good" : status));
```
Scottmitch: consider putting this in a string building to increase the likelihood this string prints out in a contiguous manner. This would be more consistent with other examples and self contained.

example/src/main/java/io/netty/example/ocsp/OcspClientExample.java
```diff
@@ -0,0 +1,224 @@
...
+        }
+
+        // Using Wikipedia as an example. I'd rather use Netty's own website
+        // but the server doesn't support OCSP. A few other examples could
+        // be Microsoft or Squarespace. Use OpenSSL's CLI client to assess
+        // if a server supports OCSP stapling. E.g.:
+        //
+        // openssl s_client -tlsextdebug -status -connect www.squarespace.com:443
+        //
+        final String host = "www.wikipedia.org";
```
Scottmitch: should we just have the client example talk to the server example and provide w/e certs are necessary?

example/src/main/java/io/netty/example/ocsp/OcspClientExample.java
```diff
@@ -0,0 +1,224 @@
...
+        }
+    }
+
+    private static ChannelInitializer<Channel> newClientHandler(final ReferenceCountedOpenSslContext context,
+            final FullHttpRequest request, final Promise<FullHttpResponse> promise) {
+
+        return new ChannelInitializer<Channel>() {
+            @Override
+            protected void initChannel(Channel ch) throws Exception {
+                ReferenceCountedOpenSslEngine engine
```
Scottmitch: consider using `context.newHandler(ch.alloc())` instead (you can get the engine via `engine()`). This may produce some different behavior in the future (PR https://github.com/netty/netty/pull/6365) and I would like to make sure it still works at that time.

example/src/main/java/io/netty/example/ocsp/OcspClientExample.java
```diff
@@ -0,0 +1,224 @@
...
+                pipeline.addLast(new SslHandler(engine));
+                pipeline.addLast(new HttpClientCodec());
+                pipeline.addLast(new HttpObjectAggregator(1 * 1024 * 1024));
+                pipeline.addLast(new HttpClientHandler(request, promise));
+            }
+
+            @Override
+            public void channelInactive(ChannelHandlerContext ctx) throws Exception {
+                if (!promise.isDone()) {
+                    promise.tryFailure(new IllegalStateException());
```
Scottmitch: consider adding a descriptive string: e.g `socket was closed before response was received`.

example/src/main/java/io/netty/example/ocsp/OcspClientExample.java
```diff
@@ -0,0 +1,224 @@
...
+        public void channelActive(ChannelHandlerContext ctx) throws Exception {
+            ctx.writeAndFlush(request).addListener(ChannelFutureListener.FIRE_EXCEPTION_ON_FAILURE);
+
+            ctx.fireChannelActive();
+        }
+
+        @Override
+        public void channelInactive(ChannelHandlerContext ctx) throws Exception {
+            if (!promise.isDone()) {
+                promise.tryFailure(new IllegalStateException());
```
Scottmitch: consider adding a descriptive error string for the exception

example/src/main/java/io/netty/example/ocsp/OcspRequestBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+        return this;
+    }
+
+    public OcspRequestBuilder issuer(X509Certificate issuer) {
+        this.issuer = issuer;
+        return this;
+    }
+
+    /**
+     * ATTENTION: This object is not re-usable! It contains an one-time nonce!
```
Scottmitch: `an one-time` -> `a one-time`

Scottmitch: also aren't you randomly generating the nonce on each build?

example/src/main/java/io/netty/example/ocsp/OcspUtils.java
```diff
@@ -0,0 +1,183 @@
...
+
+        DLSequence aiaSequence = (DLSequence) authorityInfoAccess;
+        DERTaggedObject taggedObject = findObject(aiaSequence,
+                OCSP_RESPONDER_OID, DERTaggedObject.class);
+        if (taggedObject == null) {
+            return null;
+        }
+
+        // Same as BERTags#OBJECT_IDENTIFIER ?!
+        if (taggedObject.getTagNo() != 6) {
```
Scottmitch: make a define for this value?

example/src/main/java/io/netty/example/ocsp/OcspUtils.java
```diff
@@ -0,0 +1,183 @@
...
+
+    /**
+     * This is a very crude and non-scalable HTTP client to fetch the OCSP response from the
+     * CA's OCSP responder server.
+     */
+    public static OCSPResp request(URL url, OCSPReq request, long timeout, TimeUnit unit) throws IOException {
+
+        byte[] encoded = request.getEncoded();
+
+        HttpURLConnection connection = (HttpURLConnection) url.openConnection();
```
Scottmitch: 😮 

Consider adding a comment `// TODO: use Netty` 😉

example/src/main/java/io/netty/example/ocsp/OcspClientExample.java
rkapsi: Unfortunately that is not possible or rather... It's not possible for normal mortals like us to produce valid OCSP staple. You've to be a CA to do that (realistically speaking).

Our server would need a real (non-self signed) certificate that was issued by a real CA (or internal CA) who operates an OCSP Responder Server.

Our server then contacts the CA's OCSP Responder Server and obtains the OCSP staple (see server example) which is then passed to the clients which in turn then can use it to validate the certificate and don't need to reach out on their own to the CA to check the validity.

rkapsi: Maybe this screenshort of Wikipedia's cert simplifies/explains the flow. The certificate issued by a real CA contains an URL that consumers can use to check its revocation status. Clients such as FF do check it! The interesting thing about OCSP is that it's possible for the SSL server to pre-fetch it from the CA and pass it along to the client and remove the necessity for the client to perform this lookup. The OCSP staple is a cartographically signed string (i.e. we as a server can't temper with it) and it's valid for up to 10 days (i.e. highly cachable). 

screenshot from 2017-03-06 12-23-46

example/src/main/java/io/netty/example/ocsp/OcspRequestBuilder.java
rkapsi: Yes, a new nonce is created for each `build()`. The comment is referring to the built object itself because it's tempting to cache that `OCSPReq` object.

example/src/main/resources/io/netty/example/ocsp/netty_io_chain.pem
```diff
@@ -0,0 +1,63 @@
+-----BEGIN CERTIFICATE-----
```
rkapsi: That cert chain is actually not generated. It's the real cert chain for https://netty.io and it was simply pulled with Chrome's export feature.

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,526 @@
...
+
+        byte[] expected = serverOcspContext.response();
+        byte[] actual = clientOcspContext.response();
+
+        assertNotNull(actual);
+        assertNotSame(expected, actual);
+        assertArrayEquals(expected, actual);
+
+        // If the client rejects the staple then Client and Server will fail with SSLHandshakeExceptions
+        assertTrue("" + clientCauseRef.get(), isCausedBy(clientCauseRef.get(), SSLHandshakeException.class));
```
Scottmitch: consider saving a reference to this exception ... just to ensure the same value is used in the comparison and string printout.

(general comment for this test)

example/src/main/java/io/netty/example/ocsp/OcspClientExample.java
rkapsi: I wouldn't say it's prohibitive but you'll need to bootstrap a CA with a root certificate, an intermediate certificate and finally your certificate. You then use your certificate and the intermediate certificate to construct an OCSP request. You POST that request to the CA's OCSP responder server which in turn returns a cryptographically signed OCSP response (signed using the intermediary cert+private key).

With that said, if you look at the OcspTest you'll see that OpenSSL doesn't care about the actual `byte[]` that are being passed around and it's essentially showing what you're asking.

Scottmitch: The examples you have produced are awesome ... I would just prefer if would could make them "self contained".  We actually generate self signed root CA, intermediate signed by that CA, and then certs signed by the intermediate in [1]. We can assume the server has access to the intermediate's cert+private key for the sake of the example (and comment to explain what should be done in the real world).  This way we can get rid of the whole `OcspResponderClient` class, leave this as exercise to the user, and just reference to the spec.

[1] https://github.com/netty/netty/blob/4.1/handler/src/test/resources/io/netty/handler/ssl/generate-certs.sh

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -248,6 +249,14 @@ protected void deallocate() {
         }
     }
 
+    /**
+     * Enables (or disables) OCSP stapling on the {@link SSLEngine}.
+     */
+    @UnstableApi
+    public synchronized boolean enableOcsp(boolean enabled) {
+        return SSL.setTlsExtStatusType(ssl, enabled ? SSL.TLSEXT_STATUSTYPE_ocsp : 0);
```
rkapsi: Ack. Happy to change it into `enableOcsp()` as there is no use-case for the toggle functionality.

example/src/main/java/io/netty/example/ocsp/OcspResponderClient.java
```diff
@@ -0,0 +1,265 @@
...
+import io.netty.util.concurrent.Future;
+import io.netty.util.concurrent.FutureListener;
+import io.netty.util.concurrent.ImmediateEventExecutor;
+import io.netty.util.concurrent.Promise;
+
+/**
+ * This is a very simple client to interact with the CA's OCSP Responder Server. Please be mindful
+ * when using it and adjust it to your needs. This is not meant to be production grade code!
+ */
+class OcspResponderClient {
```
rkapsi: I think we should revert this back to a plain JDK implementation (i.e. HttpURLConnection) and provide it just a short and simple example how to interact with the CA's responder as there's not much information about it on the Internet (or rather it's difficult to formulate the query as OCSP is foreign to many ppl.)

Once you know that it's a HTTP POST, with special `content-type` and `accept` headers and a special payload it's "trivial" to deduct a Netty client from it.

In general, in order to operate OCSP you need to build some infrastructure that goes beyond the Netty project. You need caches, an OCSP responder client (i.e. something like this class) and a strategy to manage and refresh all these staples that are only valid for < 10 days.

Scottmitch: I am fine with that for the sake of an example, and assuming we can't or its too much hassle to implement a more self contained example https://github.com/netty/netty/pull/6158#discussion_r104572135. My previous comment on this was meant to literally add something like `// TODO: we use HttpURLConnection for the sake of demonstration, but you should consider using Netty's HTTP client instead`.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -484,12 +489,34 @@ public final long sslCtxPointer() {
         return ctx;
     }
 
+    /**
+     * Returns {@code true} if the {@link ReferenceCountedOpenSslContext} has OCSP stapling enabled.
```
Scottmitch: update this comment.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -248,6 +250,39 @@ protected void deallocate() {
         }
     }
 
+    /**
+     * Enables OCSP stapling on the {@link SSLEngine} and throws an {@link Exception} if OCSP stapling is not supported.
+     */
+    @UnstableApi
+    public void enableOcsp(ReferenceCountedOpenSslContext context) {
+        context.enableOcsp();
```
Scottmitch: Hum this relationship is unnatural. Typically the `SSlContext` provides the default configuration for creating an `SSLEngine` and the same relationship olds for `SSL_CTX` and `SSL`. Could we just instead have tcnative throw if the `SSL_CTX` has not been setup correctly? For example we could add a `bool oscp_enabled` to `tcn_ssl_ctxt_t` (in `ssl_private.h`). `SSLContext#enableOcsp` would set this state (even for `BoringSSL`) and then `SSL#enableOcsp` would check this state has been set. You can get the `tcn_ssl_ctxt_t` from the `SSL*` via `SSL_get_app_data2`.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -248,6 +250,39 @@ protected void deallocate() {
...
+            SSL.setOcspResponse(ssl, response);
+        }
+    }
+
+    /**
+     * Returns the OCSP response or {@code null} if the server didn't provide a stapled OCSP response.
+     */
+    @UnstableApi
+    public synchronized byte[] getOcspResponse() {
+        if (clientMode) {
```
Scottmitch: nit: make a tertiary statement.

handler/src/main/java/io/netty/handler/ssl/ocsp/OcspClientHandler.java
```diff
@@ -0,0 +1,67 @@
...
+        if (evt instanceof SslHandshakeCompletionEvent) {
+            try {
+                SslHandshakeCompletionEvent event = (SslHandshakeCompletionEvent) evt;
+                if (event.isSuccess()) {
+                    if (!verify(ctx, engine)) {
+                        throw OCSP_VERIFICATION_EXCEPTION;
+                    }
+                }
+            } finally {
+                ctx.pipeline().remove(this);
```
Scottmitch: can this just be removed before the verification?

normanmaurer:
Sorry @Scottmitch @rkapsi but I had no time yet to look at this... Will try to do over the next days

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
rkapsi: Agree. Next PR fixes it. I've reduced it to:

```
SSLContext#enableOcsp
SSL#setOcspResponse
SSL#getOcspResponse
```

The `SSL#enableOcsp` call doesn't need to be exposed to the user.

example/src/main/java/io/netty/example/ocsp/OcspRequestBuilder.java
```diff
@@ -0,0 +1,105 @@
...
+        this.issuer = issuer;
+        return this;
+    }
+
+    /**
+     * ATTENTION: The returned {@link OCSPReq} is not re-usable/cacheable! It contains a one-time nonce
+     * and CA's will (should) reject subsequent requests with the same nonce.
+     */
+    public OCSPReq build() throws OCSPException, IOException, CertificateEncodingException {
+
```
normanmaurer: nit: remove empty line

example/src/main/java/io/netty/example/ocsp/OcspUtils.java
```diff
@@ -0,0 +1,178 @@
...
+        return null;
+    }
+
+    /**
+     * TODO: This is a very crude and non-scalable HTTP client to fetch the OCSP response from the
+     * CA's OCSP responder server. It's meant to demonstrate the basic building blocks on how to
+     * interact with the responder server and you should consider using Netty's HTTP client instead.
+     */
+    public static OCSPResp request(URI uri, OCSPReq request, long timeout, TimeUnit unit) throws IOException {
+
```
normanmaurer: nit: remove empty line

example/src/main/java/io/netty/example/ocsp/OcspUtils.java
```diff
@@ -0,0 +1,178 @@
...
+                    int contentLength = connection.getContentLength();
+                    if (contentLength == -1) {
+                        // Probably a terrible idea!
+                        contentLength = Integer.MAX_VALUE;
+                    }
+
+                    ByteArrayOutputStream baos = new ByteArrayOutputStream();
+                    try {
+                        byte[] buffer = new byte[8 * 1024];
```
normanmaurer: 8192

handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
```diff
@@ -256,6 +256,13 @@ public static boolean isAlpnSupported() {
     }
 
     /**
+     * Returns {@code true} if the used version of OpenSSL supports OCSP stapling.
+     */
+    public static boolean isOcspSupported() {
+      return version() >= 0x10002000L;
```
normanmaurer: @rkapsi is this good enough and also hold true for Libressl and Boringssl ?

handler/src/main/java/io/netty/handler/ssl/ocsp/OcspClientHandler.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+@UnstableApi
+public abstract class OcspClientHandler extends ChannelInboundHandlerAdapter {
+
+    private static final SSLHandshakeException OCSP_VERIFICATION_EXCEPTION = ThrowableUtil.unknownStackTrace(
+            new SSLHandshakeException("Bad OCSP response"), OcspClientHandler.class, "verify(...)");
+
+    private final ReferenceCountedOpenSslEngine engine;
+
+    public OcspClientHandler(ReferenceCountedOpenSslEngine engine) {
```
normanmaurer: protected as the class is abstract

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,464 @@
...
+
+                if (handler != null) {
+                    pipeline.addLast(handler);
+                }
+            }
+        };
+    }
+
+    private static boolean isCausedBy(Throwable t, Class<? extends Throwable> clazz) {
+        while (t != null) {
```
normanmaurer: consider using a `do {} while(...)` loop as it should not be called with a null `Throwable`

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,464 @@
...
+    }
+
+    private static byte[] newOcspResponse() {
+        // Assume we got the OCSP staple from somewhere. Using a bogus byte[]
+        // in the test because getting a true staple from the CA is quite involved.
+        // It requires HttpCodec and Bouncycastle and the test may be very unreliable
+        // because the OCSP responder servers are basically being DDoS'd by the
+        // Internet.
+
+        byte[] bogus = "I am a bogus OCSP staple. OpenSSL does not care about the format of the byte[]!".getBytes();
```
normanmaurer: just return directly

pom.xml
```diff
@@ -242,7 +242,7 @@
     <!-- Fedora-"like" systems. This is currently only used for the netty-tcnative dependency -->
     <os.detection.classifierWithLikes>fedora</os.detection.classifierWithLikes>
     <tcnative.artifactId>netty-tcnative</tcnative.artifactId>
-    <tcnative.version>2.0.0.Beta7</tcnative.version>
+    <tcnative.version>2.0.0.Final-SNAPSHOT</tcnative.version>
```
normanmaurer: 2.0.1.Final-SNAPSHOT now :)

handler/src/main/java/io/netty/handler/ssl/ocsp/OcspClientHandler.java
rkapsi: Making it protected would prevent ad-hoc instantiation and force the user to declare a class which can be a nuisance for prototyping/quick hackery.

```java
pipeline.addLast(new OcspClientHandler(engine) {
  @Override
  public boolean verify(...) {
    return true;
  }
});

File: handler/src/main/java/io/netty/handler/ssl/OpenSsl.java (hunk 42 above)
Comment (rkapsi) [inline handler/src/main/java/io/netty/handler/ssl/OpenSsl.java]: TBH I don't know. I'm operating under the assumption they're forks of OpenSSL and OCSP should be present as of 1.0.0 and we have only tested 1.0.2+.

Comment (normanmaurer) [inline handler/src/main/java/io/netty/handler/ssl/OpenSsl.java]: maybe try to run the test while using "netty-tcnative-boringssl-static" and see what happens :)

Comment (rkapsi) [inline handler/src/main/java/io/netty/handler/ssl/OpenSsl.java]: OK, LibreSSL works. 

BoringSSL doesn't work (fails gracefully with my OCSP is not supported exception) but that is because... It's either getting built with `OPENSSL_NO_OCSP` or `OPENSSL_VERSION_NUMBER` is something very different than expected. Checking... :thinking:

Comment (rkapsi) [inline handler/src/main/java/io/netty/handler/ssl/OpenSsl.java]: Confirm, BoringSSL gets built with `OPENSSL_NO_OCSP`.

Comment (normanmaurer) [inline handler/src/main/java/io/netty/handler/ssl/OpenSsl.java]: @nmittler should we change this ?

Comment (normanmaurer) [inline handler/src/main/java/io/netty/handler/ssl/OpenSsl.java]: also @davidben

Comment (nmittler) [inline handler/src/main/java/io/netty/handler/ssl/OpenSsl.java]: IIRC, the version number for BoringSSL is consistent with OpenSSL.

@davidben can you confirm?

Comment (normanmaurer) [inline handler/src/main/java/io/netty/handler/ssl/OpenSsl.java]: @nmittler the question was more if we should not build with `OPENSSL_NO_OCSP ` ;

Comment (rkapsi) [inline handler/src/main/java/io/netty/handler/ssl/OpenSsl.java]: I believe the `#define` is coming from here. Not sure what the best course of action is to unset it.

https://github.com/google/boringssl/blob/master/include/openssl/opensslconf.h#L45

Comment (Scottmitch) [inline handler/src/main/java/io/netty/handler/ssl/OpenSsl.java]: maybe we should only consider `OPENSSL_NO_OCSP` if we are not BoringSSL?

File: handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java [hunk 47]
```diff
@@ -246,12 +250,42 @@ protected void deallocate() {
...
+    /**
+     * Sets the OCSP response.
+     */
+    @UnstableApi
+    public synchronized void setOcspResponse(byte[] response) {
+        if (!ocspEnabled) {
+            throw new IllegalStateException("OCSP stapling is not enabled");
+        }
+
+        if (!clientMode) {
```
Scottmitch: should we treat this as an error if called while in the wrong mode?

handler/src/main/java/io/netty/handler/ssl/ocsp/OcspClientHandler.java
Scottmitch: @rkapsi - are you sure? I don't see why this would cause an issue, can you explain? `OcspClientHandler` is an anonymous type that should have access to `protected` things in the super class.

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,463 @@
...
+
+    @BeforeClass
+    public static void checkOcspSupported() {
+        assumeTrue(OpenSsl.isOcspSupported());
+    }
+
+    @Test(expected = IllegalStateException.class)
+    public void testClientOcspNotEnabled() throws Exception {
+        SslContext context = SslContextBuilder.forClient()
+                .sslProvider(SslProvider.OPENSSL)
```
Scottmitch: consider testing with both providers (`OPENSSL_REFCNT`) like other tests.

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,463 @@
...
+            ReferenceCountUtil.release(context);
+        }
+    }
+
+    @Test(expected = IllegalStateException.class)
+    public void testServerOcspNotEnabled() throws Exception {
+        SelfSignedCertificate ssc = new SelfSignedCertificate();
+        try {
+            SslContext context = SslContextBuilder.forServer(ssc.certificate(), ssc.privateKey())
+                    .sslProvider(SslProvider.OPENSSL)
```
Scottmitch: consider testing with both providers (`OPENSSL_REFCNT`) like other tests.

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,463 @@
...
+                try {
+                    EventLoopGroup group = new DefaultEventLoopGroup();
+                    try {
+                        LocalAddress address = new LocalAddress("handshake-" + Math.random());
+                        Channel server = newServer(group, address, serverSslContext, response, serverHandler);
+                        Channel client = newClient(group, address, clientSslContext, callback, clientHandler);
+                        try {
+                            assertTrue("Something went wrong.", latch.await(10L, TimeUnit.SECONDS));
+                        } finally {
+                            client.close();
```
Scottmitch: consider blocking on the channel close futures. If the event loop is closed before the channels close we could leak buffers (close may execute a task on the event loop to cleanup internal queues).

handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
davidben: @Scottmitch is right. OpenSSL's OCSP APIs involve a lot of machinery which we don't support. (`d2i_OCSP_RESPONSE`, etc.) BoringSSL has a simpler API which just blindly passes byte strings along. Since it's BoringSSL-specific, consumers are going to need BoringSSL-specific code anyway.

Re the version number, yeah we try to set the version number to align with OpenSSL as best we can, so right now we pretend to be 1.0.2.

nmittler: @Scottmitch does it make sense to use the BoringSSL API rather than just not supporting it?

Scottmitch: Thanks for confirming @davidben.

@nmittler - We are working with @rkapsi to support both BoringSSL and OpenSSL (and LibreSSL should come for "free"). We are currently using both APIs to make this happen and it is just a matter of working out the kinks to make sure we can make the two different APIs (BoringSSL and OpenSSL) work consistently with one API (tcnative/netty) on top. I think we are very close to pulling this off.

handler/src/main/java/io/netty/handler/ssl/ocsp/OcspClientHandler.java
rkapsi: Operator error, disregard me.

handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
rkapsi: In regards to the machinery: That is totally cool that BoringSSL doesn't support the heavy machinery because we don't use any of OpenSSL's either. All we want to do is blindly pass along byte strings. It's literally just that and BoringSSL's API does just that. Again, totally cool. Unfortunately it's defaulting to `OPENSSL_NO_OCSP` and I guess we're confused how to proceed. :thinking: 

I'm personally cool not supporting BoringSSL but that means I'll never be able to use it either because our lovely non-profit CA will not appreciate it being DDoS'd if our servers do handshakes without staples.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
rkapsi: sgtm

handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
davidben: `OPENSSL_NO_OCSP` means we do not support OpenSSL's OCSP API. If we did not define it, loads of projects would fail to build because they, naturally, do not use BoringSSL's API. Like with any other project, your options are not to support OCSP stapling with BoringSSL (in which case `OPENSSL_NO_OCSP`, what you would have been checking anyway, Just Works and you need no additional work) or to add BoringSSL-specific code that does not look at `OPENSSL_NO_OCSP`. BoringSSL-specific code is obviously needed anyway for BoringSSL-specific API, so I don't see why having to ignore a preprocessor symbol which happens to have a similar name as the feature matters.

davidben: In case this is where the confusion comes from, BoringSSL's APIs has *no relation* to `OPENSSL_NO_OCSP`. They are available regardless.

We do not, in general, have individual toggles for all our features like OpenSSL does. Instead, we define a set of `OPENSSL_NO_*` values based on what makes portability simplest. These symbols do not affect what features BoringSSL exposes.

Scottmitch: Agreed I think we have the tools we need and everything is ok :) @rkapsi we just need to adjust the macros see https://github.com/netty/netty-tcnative/pull/215#discussion_r105228059 ... also also add more details here.

pom.xml
```diff
@@ -241,7 +241,7 @@
     <!-- Configure the os-maven-plugin extension to expand the classifier on                  -->
     <!-- Fedora-"like" systems. This is currently only used for the netty-tcnative dependency -->
     <os.detection.classifierWithLikes>fedora</os.detection.classifierWithLikes>
-    <tcnative.artifactId>netty-tcnative</tcnative.artifactId>
+    <tcnative.artifactId>netty-tcnative-boringssl-static</tcnative.artifactId>
```
Scottmitch: nit: revert this after you are done testing.

pom.xml
```diff
@@ -241,8 +241,8 @@
     <!-- Configure the os-maven-plugin extension to expand the classifier on                  -->
     <!-- Fedora-"like" systems. This is currently only used for the netty-tcnative dependency -->
     <os.detection.classifierWithLikes>fedora</os.detection.classifierWithLikes>
-    <tcnative.artifactId>netty-tcnative</tcnative.artifactId>
-    <tcnative.version>2.0.0.Final</tcnative.version>
+    <tcnative.artifactId>netty-tcnative-boringssl-static</tcnative.artifactId>
```
Scottmitch: revert this after you are done testing

example/src/main/java/io/netty/example/ocsp/OcspClientExample.java
```diff
@@ -0,0 +1,254 @@
...
+        ReferenceCountedOpenSslContext context
+            = (ReferenceCountedOpenSslContext) SslContextBuilder.forClient()
+                .sslProvider(SslProvider.OPENSSL)
+                .build();
+        context.enableOcsp();
+
+        try {
+            EventLoopGroup group = new NioEventLoopGroup();
+            try {
+                FullHttpRequest request = new DefaultFullHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.GET, "/");
```
Scottmitch: can we create this request when it is needed? Otherwise we risk leaking this reference counted object.

example/src/main/java/io/netty/example/ocsp/OcspClientExample.java
```diff
@@ -0,0 +1,254 @@
...
+                SslHandler sslHandler = context.newHandler(ch.alloc());
+                ReferenceCountedOpenSslEngine engine
+                    = (ReferenceCountedOpenSslEngine) sslHandler.engine();
+
+                ChannelPipeline pipeline = ch.pipeline();
+                pipeline.addLast(sslHandler);
+                pipeline.addLast(new ExampleOcspClientHandler(engine));
+
+                pipeline.addLast(new HttpClientCodec());
+                pipeline.addLast(new HttpObjectAggregator(1 * 1024 * 1024));
```
Scottmitch: `1 * 1024 * 1024` -> `1024 * 1024`?

example/src/main/java/io/netty/example/ocsp/OcspClientExample.java
```diff
@@ -0,0 +1,254 @@
...
+                .append("OCSP status of ").append(ctx.channel().remoteAddress())
+                .append("\n  Status: ").append(status == CertificateStatus.GOOD ? "Good" : status)
+                .append("\n  This Update: ").append(first.getThisUpdate())
+                .append("\n  Next Update: ").append(first.getNextUpdate())
+                .append("\n  Cert Serial: ").append(certSerial)
+                .append("\n  OCSP Serial: ").append(ocspSerial)
+                .toString();
+            System.out.println(message);
+
+            if (status != CertificateStatus.GOOD) {
```
Scottmitch: nit: consider simplifying these conditional statements

```java
return status == CertificateStatus.GOOD && certSerial.equals(ocspSerial);
```

example/src/main/java/io/netty/example/ocsp/OcspUtils.java
```diff
@@ -0,0 +1,178 @@
...
+import org.bouncycastle.asn1.DLSequence;
+import org.bouncycastle.asn1.x509.Extension;
+import org.bouncycastle.cert.ocsp.OCSPReq;
+import org.bouncycastle.cert.ocsp.OCSPResp;
+import org.bouncycastle.x509.extension.X509ExtensionUtil;
+
+import io.netty.util.CharsetUtil;
+
+public final class OcspUtils {
+    // http://www.alvestrand.no/objectid/1.3.6.1.5.5.7.48.1.html
```
Scottmitch: nit: make this a javadoc comment so the link shows up in IDEs

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -487,12 +491,36 @@ public final long sslCtxPointer() {
         return ctx;
     }
 
+    /**
+     * Enables OCSP stapling or throws an {@link Exception} if OCSP is not supported.
+     */
+    @UnstableApi
+    public final void enableOcsp() {
```
Scottmitch: is the intention to dynamically toggle this value? For example is it sufficient to just add a method to the `SslContextBuilder` and configure it in the constructor of this method.

For example isn't there some assumptions from the `SSL`'s perspective that the `SSL_CTX` has some callback set? We may be OK if we only ever allow a transition from disabled -> enabled, but it may introduce undefined behavior if we forget this relationship and add a enabled -> disabled transition.

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,471 @@
...
+                    latch.countDown();
+                }
+            }
+        };
+
+        byte[] response = newOcspResponse();
+        OcspClientCallback callback = new OcspClientCallback() {
+            @Override
+            public boolean verify(byte[] response) throws Exception {
+                throw new OcspTestException("testClientException");
```
Scottmitch: consider creating this exception outside the scope and check for `assertSame` below:

```java
final OcspTestException clientException = new OcspTestException("testClientException");

...
@Override
public boolean verify(byte[] response) throws Exception {
  throw clientException;
}
...

assertSame(clientException, causeRef.get());
```

pom.xml
```diff
@@ -242,7 +242,7 @@
     <!-- Fedora-"like" systems. This is currently only used for the netty-tcnative dependency -->
     <os.detection.classifierWithLikes>fedora</os.detection.classifierWithLikes>
     <tcnative.artifactId>netty-tcnative</tcnative.artifactId>
-    <tcnative.version>2.0.0.Final</tcnative.version>
+    <tcnative.version>2.0.1.Final-SNAPSHOT</tcnative.version>
```
Scottmitch: reminder: we need to update this at some point. (/cc @normanmaurer )

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
rkapsi: We can move it into SslContextBuilder and it's only allowed to transition from disabled to enabled.

rkapsi:
@Scottmitch moved the `enableOcsp()` method to `SslContextBuilder`, made your suggested changes and added two unit tests for JDK SSL impl. which don't support OCSP and should blow up.

example/src/main/java/io/netty/example/ocsp/OcspClientExample.java
```diff
@@ -0,0 +1,250 @@
...
+                .append("OCSP status of ").append(ctx.channel().remoteAddress())
+                .append("\n  Status: ").append(status == CertificateStatus.GOOD ? "Good" : status)
+                .append("\n  This Update: ").append(first.getThisUpdate())
+                .append("\n  Next Update: ").append(first.getNextUpdate())
+                .append("\n  Cert Serial: ").append(certSerial)
+                .append("\n  OCSP Serial: ").append(ocspSerial)
+                .toString();
+            System.out.println(message);
+
+            if (status != CertificateStatus.GOOD || !certSerial.equals(ocspSerial)) {
```
Scottmitch: nit: consider simplifying to the following:

```java
return status == CertificateStatus.GOOD && certSerial.equals(ocspSerial);
```

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,492 @@
...
+    }
+
+    private void testClientOcspNotEnabled(SslProvider sslProvider) throws Exception {
+        SslContext context = SslContextBuilder.forClient()
+                .sslProvider(sslProvider)
+                .build();
+        try {
+            SslHandler sslHandler = context.newHandler(ByteBufAllocator.DEFAULT);
+            ReferenceCountedOpenSslEngine engine = (ReferenceCountedOpenSslEngine) sslHandler.engine();
+            engine.getOcspResponse();
```
Scottmitch: the `engine` must be released at some point. `SslHandler` typically takes ownership, but it must be interred into a pipeline ignorer to do so.

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,492 @@
...
+    private void testServerOcspNotEnabled(SslProvider sslProvider) throws Exception {
+        SelfSignedCertificate ssc = new SelfSignedCertificate();
+        try {
+            SslContext context = SslContextBuilder.forServer(ssc.certificate(), ssc.privateKey())
+                    .sslProvider(sslProvider)
+                    .build();
+            try {
+                SslHandler sslHandler = context.newHandler(ByteBufAllocator.DEFAULT);
+                ReferenceCountedOpenSslEngine engine = (ReferenceCountedOpenSslEngine) sslHandler.engine();
+                engine.setOcspResponse(new byte[] { 1, 2, 3 });
```
Scottmitch: the `engine` must be released at some point. `SslHandler` typically takes ownership, but it must be interred into a pipeline ignorer to do so.

handler/src/test/java/io/netty/handler/ssl/ocsp/OcspTest.java
```diff
@@ -0,0 +1,492 @@
...
+        testClientException(SslProvider.OPENSSL);
+    }
+
+    @Test(timeout = 10000L)
+    public void testClientExceptionOpenSslRefCnt() throws Exception {
+        testClientException(SslProvider.OPENSSL_REFCNT);
+    }
+
+    /**
+     * Testing what happens if the {@link ClientOcspContext} throws an {@link Exception}.
```
Scottmitch: `ClientOcspContext` cannot be found. Update this comment plz.

handler/src/main/java/io/netty/handler/ssl/ocsp/OcspClientHandler.java
```diff
@@ -0,0 +1,67 @@
...
+    protected abstract boolean verify(ChannelHandlerContext ctx, ReferenceCountedOpenSslEngine engine) throws Exception;
+
+    @Override
+    public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
+        if (evt instanceof SslHandshakeCompletionEvent) {
+            ctx.pipeline().remove(this);
+
+            SslHandshakeCompletionEvent event = (SslHandshakeCompletionEvent) evt;
+            if (event.isSuccess()) {
+                if (!verify(ctx, engine)) {
```
normanmaurer: merge these two branches:

```java
if (((SslHandshakeCompletionEvent) evt).isSuccess() && !verify(ctx, engine)) {
    throw OCSP_VERIFICATION_EXCEPTION;
}
```

handler/src/main/java/io/netty/handler/ssl/SslContext.java
```diff
@@ -398,15 +398,18 @@ static SslContext newServerContextInternal(
...
+            boolean enableOcsp) throws SSLException {
 
         if (provider == null) {
             provider = defaultServerProvider();
         }
 
         switch (provider) {
         case JDK:
+            if (enableOcsp) {
+                throw new IllegalArgumentException("OCSP is not supported.");
```
normanmaurer: add the provider to the message

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -246,12 +249,48 @@ protected void deallocate() {
...
+    public synchronized void setOcspResponse(byte[] response) {
+        if (!enableOcsp) {
+            throw new IllegalStateException("OCSP stapling is not enabled");
+        }
+
+        if (clientMode) {
+            throw new IllegalStateException("Not a server SSLEngine");
+        }
+
+        SSL.setOcspResponse(ssl, response);
```
normanmaurer: Please reduce the scope of the synchronized block:

```java
public void setOcspResponse(byte[] response) {
     if (!enableOcsp) {
         throw new IllegalStateException("OCSP stapling is not enabled");
     }
 
     if (clientMode) {
         throw new IllegalStateException("Not a server SSLEngine");
     }

     synchronized(this) {
         SSL.setOcspResponse(ssl, response);
     }
}
```

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -246,12 +249,48 @@ protected void deallocate() {
...
+    public synchronized byte[] getOcspResponse() {
+        if (!enableOcsp) {
+            throw new IllegalStateException("OCSP stapling is not enabled");
+        }
+
+        if (!clientMode) {
+            throw new IllegalStateException("Not a client SSLEngine");
+        }
+
+        return SSL.getOcspResponse(ssl);
```
normanmaurer: same as in https://github.com/netty/netty/pull/6158/files#r109108102

rkapsi:
@Scottmitch @normanmaurer - done!

Scottmitch:
@normanmaurer - Note the necessary changes for tcnative have already been merged. So if we plan to update the version of tcnative for the next version of Netty we could merge this now.

@rkapsi - Please rebase to resolve recently introduced conflicts.

rkapsi:
@Scottmitch - done.

Scottmitch:
the build failed due to a recent API change in netty-tcnative. I merged the corresponding Netty PR https://github.com/netty/netty/pull/6581 which should resolve the issue. Can you rebase again?

rkapsi:
@Scottmitch - done.

Scottmitch:
4.1 (077a198) 4.0 (305e788)

Scottmitch:
Great work! Thanks for your patience @rkapsi!

Issue #6150, opened 2016-12-20
Detecting actual Channels idleness vs. slowness


rkapsi:
I don't know if you'd consider it a missing feature or a bug but there appears to be no effective way of measuring write idleness. The natural choice is to use Netty's own `IdleStateHandler` but it may not work as expected for `write()`. For the sake of simplicity I'm providing a HTTP example but we encountered it with H2.

Basically... The user's (our) intention is to close an idle connection. The `IdleStateHandler` uses the write's `ChannelFuture` to determine if a client is idle but it doesn't take into consideration that the client might be just slow and we've just thrown a large `ByteBuf` at it. I think we ran into this with H2 due to things like `CoalescingBufferQueue` which try to aggregate multiple writes into a single one.

The repro is very simple. You just need to throttle your client a little bit. I'm using Chrome's [built-in](http://imgur.com/a/WfcZQ) throtting feature for it. On the server side (the code below) we want to close the connection if the client appears to be idle for 30 seconds.

The problem is that `IdleStateHandler` will make a pure binary decision on the completeness of the `ChannelFuture` vs. taking actual progression into consideration and it appears there is no API for it.

I haven't looked at all `Channel` implementations but `NioSocketChannel` and `EpollSocketChannel` (via `AbstractEpollStreamChannel`) do have knowledge about the number of bytes written in each cycle. An easy fix could be that channel's sum up these values in a `volatile long` field and expose it to the user. Or it could be a simple `++` for each write cycle where more than 1 byte was written to the underlying socket. The user could then use it to assess whether or not any or some progress was made between two calls.
 

```java
public class IdleChannelTest {

  private static final int PORT = 8080;
  
  public static void main(String[] args) throws Exception {
    
    ChannelHandler handler = new ChannelInitializer<Channel>() {
      @Override
      protected void initChannel(Channel ch) throws Exception {
        ChannelPipeline pipeline = ch.pipeline();
        
        pipeline.addLast(new HttpServerCodec());
        pipeline.addLast(new IdleStateHandler(0L, 0L, 30L, TimeUnit.SECONDS));
        pipeline.addLast(new HttpRequestAndIdleHandler());
      }
    };
    
    EventLoopGroup group = new NioEventLoopGroup();
    
    try {
      ServerBootstrap bootstrap = new ServerBootstrap()
          .channel(NioServerSocketChannel.class)
          .group(group)
          .childHandler(handler);
      
      Channel channel = bootstrap.bind(PORT)
[...]

PR #6160, opened 2016-12-28
Detecting actual Channel write idleness vs. slowness

rkapsi:
Motivation

The IdleStateHandler tracks write() idleness on message granularity but does not take into consideration that the client may be just slow and has managed to consume a subset of the message's bytes in the configured period of time.

Modifications

Adding an optional configuration parameter to IdleStateHandler which tells it to observe ChannelOutboundBuffer's state.

Result

Fixes https://github.com/netty/netty/issues/6150

### DIFF
--- a/handler/src/main/java/io/netty/handler/timeout/IdleStateHandler.java
+++ b/handler/src/main/java/io/netty/handler/timeout/IdleStateHandler.java
@@ -17,11 +17,13 @@ package io.netty.handler.timeout;
 
 import io.netty.bootstrap.ServerBootstrap;
 import io.netty.channel.Channel;
+import io.netty.channel.Channel.Unsafe;
 import io.netty.channel.ChannelDuplexHandler;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInitializer;
+import io.netty.channel.ChannelOutboundBuffer;
 import io.netty.channel.ChannelPromise;
 import io.netty.util.concurrent.EventExecutor;
 
@@ -101,11 +103,12 @@ public class IdleStateHandler extends ChannelDuplexHandler {
     private final ChannelFutureListener writeListener = new ChannelFutureListener() {
         @Override
         public void operationComplete(ChannelFuture future) throws Exception {
-            lastWriteTime = System.nanoTime();
+            lastWriteTime = ticksInNanos();
             firstWriterIdleEvent = firstAllIdleEvent = true;
         }
     };
 
+    private final boolean observeOutput;
     private final long readerIdleTimeNanos;
     private final long writerIdleTimeNanos;
     private final long allIdleTimeNanos;
@@ -123,6 +126,10 @@ public class IdleStateHandler extends ChannelDuplexHandler {
 
     private byte state; // 0 - none, 1 - initialized, 2 - destroyed
     private boolean reading;
+
+    private long lastChangeCheckTimeStamp;
+    private int lastMessageHashCode;
+    private long lastPendingWriteBytes;
 
     /**
      * Creates a new instance firing {@link IdleStateEvent}s.
@@ -150,8 +157,20 @@ public class IdleStateHandler extends ChannelDuplexHandler {
     }
 
     /**
+     * @see #IdleStateHandler(boolean, long, long, long, TimeUnit)
+     */
+    public IdleStateHandler(
+            long readerIdleTime, long writerIdleTime, long allIdleTime,
+            TimeUnit unit) {
+        this(false, readerIdleTime, writerIdleTime, allIdleTime, unit);
+    }
+
+    /**
      * Creates a new instance firing {@link IdleStateEvent}s.
      *
+     * @param observeOutput
+     *        whether or not the consumption of {@code bytes} should be taken into
+     *        consideration when assessing write idleness. The default is {@code false}.
      * @param readerIdleTime
      *        an {@link IdleStateEvent} whose state is {@link IdleState#READER_IDLE}
      *        will be triggered when no read was performed for the specified
@@ -168,12 +187,14 @@ public class IdleStateHandler extends ChannelDuplexHandler {
      *        the {@link TimeUnit} of {@code readerIdleTime},
      *        {@code writeIdleTime}, and {@code allIdleTime}
      */
-    public IdleStateHandler(
+    public IdleStateHandler(boolean observeOutput,
             long readerIdleTime, long writerIdleTime, long allIdleTime,
             TimeUnit unit) {
         if (unit == null) {
             throw new NullPointerException("unit");
         }
+
+        this.observeOutput = observeOutput;
 
         if (readerIdleTime <= 0) {
             readerIdleTimeNanos = 0;
@@ -269,7 +290,7 @@ public class IdleStateHandler extends ChannelDuplexHandler {
     @Override
     public void channelReadComplete(ChannelHandlerContext ctx) throws Exception {
         if ((readerIdleTimeNanos > 0 || allIdleTimeNanos > 0) && reading) {
-            lastReadTime = System.nanoTime();
+            lastReadTime = ticksInNanos();
             reading = false;
         }
         ctx.fireChannelReadComplete();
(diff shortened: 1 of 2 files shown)


handler/src/main/java/io/netty/handler/timeout/IdleStateHandler.java
```diff
@@ -355,15 +377,72 @@ protected IdleStateEvent newIdleStateEvent(IdleState state, boolean first) {
...
+    /**
+     * @see #hasOutputChanged(ChannelHandlerContext, boolean)
+     */
+    private void initOutputChanged(ChannelHandlerContext ctx) {
+        if (observeOutput) {
+            Channel channel = ctx.channel();
+            Unsafe unsafe = channel.unsafe();
+            ChannelOutboundBuffer buf = unsafe.outboundBuffer();
+
+            lastMessageHashCode = System.identityHashCode(buf.current());
```
Scottmitch: maybe not when this method is currently called but it is possible the `ChannelOutboundBuffer` may be `null` (e.g. channel is closed) .. consider protecting against this or adding a comment as to why it isn't necessary.

handler/src/main/java/io/netty/handler/timeout/IdleStateHandler.java
```diff
@@ -355,15 +377,72 @@ protected IdleStateEvent newIdleStateEvent(IdleState state, boolean first) {
...
+                if (!first) {
+                    return true;
+                }
+            }
+
+            Channel channel = ctx.channel();
+            Unsafe unsafe = channel.unsafe();
+            ChannelOutboundBuffer buf = unsafe.outboundBuffer();
+
+            int messageHashCode = System.identityHashCode(buf.current());
```
Scottmitch: protect against `buf == null`.

handler/src/main/java/io/netty/handler/timeout/IdleStateHandler.java
```diff
@@ -355,15 +377,72 @@ protected IdleStateEvent newIdleStateEvent(IdleState state, boolean first) {
...
+            long pendingWriteBytes = buf.totalPendingWriteBytes();
+
+            if (messageHashCode != lastMessageHashCode
+                    || pendingWriteBytes != lastPendingWriteBytes) {
```
Scottmitch: nit: put on previous line?

handler/src/main/java/io/netty/handler/timeout/IdleStateHandler.java
```diff
@@ -399,32 +492,33 @@ public void run() {
...
 
             long lastWriteTime = IdleStateHandler.this.lastWriteTime;
             long nextDelay = writerIdleTimeNanos - (System.nanoTime() - lastWriteTime);
             if (nextDelay <= 0) {
                 // Writer is idle - set a new timeout and notify the callback.
                 writerIdleTimeout = ctx.executor().schedule(
                         this, writerIdleTimeNanos, TimeUnit.NANOSECONDS);
+
+                boolean first = firstWriterIdleEvent;
+                if (firstWriterIdleEvent) {
```
Scottmitch: consider just setting the variable unconditionally. conditionals can be expensive and this variable is not volatile.

rkapsi:
@Scottmitch done + added unit tests for new and old code (there were no unit tests for IdleStateHandler).

handler/src/main/java/io/netty/handler/timeout/IdleStateHandler.java
```diff
@@ -297,27 +318,38 @@ private void initialize(ChannelHandlerContext ctx) {
...
+            allIdleTimeout = schedule(ctx, new AllIdleTimeoutTask(ctx),
                     allIdleTimeNanos, TimeUnit.NANOSECONDS);
         }
     }
 
+    /**
+     * This method is visible for testing! It's purpose is to decouple us from
+     * a clock that we can't control.
+     */
+    long tick(TimeUnit unit) {
```
Scottmitch: can we make this method always return NS granularity and do the conversion in the unit tests?

handler/src/test/java/io/netty/handler/timeout/IdleStateHandlerTest.java
```diff
@@ -0,0 +1,382 @@
...
+            idleStateHandler.tickRun(delayInNanos / 2L, TimeUnit.NANOSECONDS);
+            assertEquals(0, events.size());
+
+            // Advance the ticker by the full amount and it should yield
+            // in an IdleStateEvent.
+            idleStateHandler.tickRun();
+            assertEquals(1, events.size());
+            assertSame(expected, events.get(0));
+        } finally {
+            channel.close();
```
Scottmitch: use `channel.finishAndReleaseAll();` instead for `EmbeddedChannel`.

handler/src/test/java/io/netty/handler/timeout/IdleStateHandlerTest.java
```diff
@@ -0,0 +1,382 @@
...
+
+            assertEquals(expected.length, events.size());
+
+            // Compare the expected with the actual IdleStateEvents
+            for (int i = 0; i < expected.length; i++) {
+                Object evt = events.get(i);
+                assertSame("Element " + i + " is not matching", expected[i], evt);
+            }
+        } finally {
+            channel.close();
```
Scottmitch: use `channel.finishAndReleaseAll();` instead for `EmbeddedChannel`.

handler/src/test/java/io/netty/handler/timeout/IdleStateHandlerTest.java
```diff
@@ -0,0 +1,382 @@
...
+            idleStateHandler.tick(3L, TimeUnit.SECONDS);
+            assertNull(channel.consume());
+
+            idleStateHandler.tickRun(2L, TimeUnit.SECONDS);
+            assertEquals(1, events.size());
+            assertEquals(31L, idleStateHandler.tick(TimeUnit.SECONDS)); // 26s + 3s + 2s
+
+            // q.e.d.
+        } finally {
+            channel.close();
```
Scottmitch: use `channel.finishAndReleaseAll();` instead for `EmbeddedChannel`.

handler/src/test/java/io/netty/handler/timeout/IdleStateHandlerTest.java
```diff
@@ -0,0 +1,382 @@
...
+
+            idleStateHandler.tickRun(2L, TimeUnit.SECONDS);
+            assertEquals(0, events.size());
+            assertEquals(11L, idleStateHandler.tick(TimeUnit.SECONDS)); // 5s + 4s + 2s
+
+            // Consume one message in 3 seconds, then be idle for 4 seconds,
+            // then run the task and we shouldn't get an IdleStateEvent because
+            // we haven't been idle for long enough!
+            idleStateHandler.tick(3L, TimeUnit.SECONDS);
+            assertNotNull(channel.consume());
```
Scottmitch: we should take care to release any object returned by `channel.consume()`

handler/src/test/java/io/netty/handler/timeout/IdleStateHandlerTest.java
```diff
@@ -0,0 +1,382 @@
...
+            // ticker by an another 2 seconds and we should get an IdleStateEvent
+            // because we've been idle for 5 seconds.
+            idleStateHandler.tick(3L, TimeUnit.SECONDS);
+            assertNull(channel.consume());
+
+            idleStateHandler.tickRun(2L, TimeUnit.SECONDS);
+            assertEquals(1, events.size());
+            assertEquals(31L, idleStateHandler.tick(TimeUnit.SECONDS)); // 26s + 3s + 2s
+
+            // q.e.d.
```
Scottmitch: hah mathematical unit test proof done I guess

handler/src/main/java/io/netty/handler/timeout/IdleStateHandler.java
rkapsi: Can be done but is there a downside with the current impl? `TimeUnit#convert(...)` will be a ~noop if source and destination units are the same (which is the case inside the IdleStateHandler class).

Scottmitch: My preference is to keep the production code minimal, and not have to think/worry about implementation details and/or what optimizations may or may not be made if possible. In this case it looks like the extra complexity can be isolated in the unit tests and thus simplify the production code, right?

rkapsi: Certainly and fair enough. I'm personally of the opinion that time without units is among some of the most useless pieces of information one can have in computer programs and I never write code without stating it explicitly (encoding the unit in the field/method name being my least favorite but still better than noting stated at all).

Scottmitch: Agreed that time (or values in general) without units can be error prone. However not accepting `TimeUnit` as an argument doesn't mean you cannot indicate the units in other ways. For example in the function name (e.g. `timeNanos()`, `nanoTime()`) and javadocs. This function is also "internal" to this class and it not intended to be part of the public API of this class so the risk of confusion due to units is lower in this case. This is effectively equivalent to `System.nanoTime()`.

rkapsi: Done.

Scottmitch:
4.1 (68a941c) 4.0 (bc26e73)

Scottmitch:
thanks @rkapsi !

PR #6162, opened 2016-12-29
maxconnections support for NioServerSocketChannel

uglycow:
Motivation:

for now, we can't put a limit on how many connections NioServerSocketChannel would accept. 
each accepted connection consume one FD of current process, so a  traffic burst could exhaust our fds.
running out of fd could lead to wield errors. 
a case could be "NoClassDefFoundError" . because URLClassLoader need to open the .class file, if he failed to open it (no fd available), we receive a NoClassDefFoundError and we cant recover from this error.

So we need to limit the number of connections.

Modification:
a util class LimitLatch is added, which we could countUp a limited number of times

NioServerSocketChannel have a LimitLatch and this latch is shared among all the NioSocketChannel accepted by this serversocket.

before accept, we try to countUp, if failed, we give up accepting new connection.
before close  the NioSocketChannel, we countDown the latch, so new connections could be accepted.

the limit is configured with a new ChannelOption<Integer> named SO_MAXCONNECTIONS.
the default value for SO_MAXCONNECTIONS is Integer.MAX_VALUE , which means no limit.

NioServerSocketChannelMaxConnectionsTest is the test. 

Result:

now we have the ability to limit the connections we accept...


common/src/main/java/io/netty/util/concurrent/LimitLatch.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.concurrent.locks.AbstractQueuedSynchronizer;
+
+/**
+ * Shared latch that allows the latch to be acquired a limited number of times
+ * after which all subsequent requests to acquire the latch will fail.
+ */
+public class LimitLatch {
+
+    private class Sync extends AbstractQueuedSynchronizer {
+        private static final long serialVersionUID = 1L;
```
fenik17: 1L - is not good serialVersionUID..

uglycow: yeah, truly is..  i may just remove it..

common/src/main/java/io/netty/util/concurrent/LimitLatch.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+
+    /**
+     * Acquires a shared latch if one is available or waits for one if no shared
+     * latch is current available.
+     */
+    public void countUpOrAwait() throws InterruptedException {
+        sync.acquireSharedInterruptibly(1);
+    }
```
fredericBregier: Is this not too dangerous ? (Blocking)

common/src/main/java/io/netty/util/concurrent/LimitLatch.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.AbstractQueuedSynchronizer;
+
+/**
+ * Shared latch that allows the latch to be acquired a limited number of times
+ * after which all subsequent requests to acquire the latch will fail.
+ */
+public class LimitLatch {
+
+    private class Sync extends AbstractQueuedSynchronizer {
```
Scottmitch: why is this necessary? Can we just use a `AtomicLong`?

common/src/main/java/io/netty/util/concurrent/LimitLatch.java
```diff
@@ -0,0 +1,111 @@
+package io.netty.util.concurrent;
+
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.AbstractQueuedSynchronizer;
+
+/**
+ * Shared latch that allows the latch to be acquired a limited number of times
+ * after which all subsequent requests to acquire the latch will fail.
+ */
+public class LimitLatch {
```
Scottmitch: Why does this class need to be public? If this class is necessary (not sure we need more than an atomic of some sort) to share between different transport types then it should live in an `internal` package.

transport/src/main/java/io/netty/channel/socket/nio/NioServerSocketChannel.java
```diff
@@ -198,5 +220,37 @@ private NioServerSocketChannelConfig(NioServerSocketChannel channel, ServerSocke
...
+
+        public int getMaxConnections() {
+            return maxConnections;
+        }
+
+        public ServerSocketChannelConfig setMaxConnections(int maxConnections) {
+            if (maxConnections <= 0) {
+                throw new IllegalArgumentException("maxConnections: " + maxConnections);
+            }
+            this.maxConnections = maxConnections;
```
Scottmitch: why is this limit different than the one `limitLatch` maintains?

transport/src/main/java/io/netty/channel/socket/nio/NioServerSocketChannel.java
```diff
@@ -198,5 +220,37 @@ private NioServerSocketChannelConfig(NioServerSocketChannel channel, ServerSocke
         protected void autoReadCleared() {
             clearReadPending();
         }
+
+        @SuppressWarnings("unchecked")
+        @Override
+        public <T> T getOption(ChannelOption<T> option) {
+            if (option == SO_MAXCONNECTIONS) {
```
Scottmitch: Seems like this code should live in `DefaultServerSocketChannelConfig`? Perhaps we can do this as a followup though...

common/src/main/java/io/netty/util/concurrent/LimitLatch.java
uglycow: it truly is dangerous to block the loop. the reason i add keep this method is 
1. because there are incoming in the backlog queue, select() would block, so the loop would keep spinning. this blocking method is meant to slove this "busy loop" problem.
2. i made the assumption that we won't run task in boss loop.
3. another way to solve "busy loop" problem would be make io.netty.channel.nio.AbstractNioChannel.NioUnsafe#read() return a special value to indicate that we have read too much, so the loop may choose to pause a little while if there is no task to run. java.lang.Integer#MAX_VALUE would be perfect for this.

uglycow: it truly is dangerous to block the loop. the reason i add keep this method is 
1. because there are incoming in the backlog queue, select() would block, so the loop would keep spinning. this blocking method is meant to slove this "busy loop" problem.
2. i made the assumption that we won't run task in boss loop.
3. another way to solve "busy loop" problem would be make io.netty.channel.nio.AbstractNioChannel.NioUnsafe#read() return a special value to indicate that we have read too much, so the loop may choose to pause a little while if there is no task to run. java.lang.Integer#MAX_VALUE would be perfect for this.

common/src/main/java/io/netty/util/concurrent/LimitLatch.java
uglycow: this is used to solve "busy loop" problem. 
1. because there are incoming in the backlog queue, select() would block, so the loop would keep spinning. this blocking method is meant to slove this "busy loop" problem.
2. i made the assumption that we won't run task in boss loop.
3. another way to solve "busy loop" problem is we return a special value from io.netty.channel.nio.AbstractNioChannel.NioUnsafe#read() to indicate that we have read too much, so the loop may choose to pause a little while if there is no task to run. java.lang.Integer#MAX_VALUE would be perfect for this.

transport/src/main/java/io/netty/channel/socket/nio/NioServerSocketChannel.java
uglycow: yeah, i consider only nioserversocketchannel when i wrote the code. let me rewrite it to make it more universal

common/src/main/java/io/netty/util/concurrent/LimitLatch.java
fredericBregier: Well, it is not used (commented) and I was thinking you did comment it because you think it was dangerous.
If not, maybe defining a timeout could be a possible guard, even if I believe this is not a good idea to wait here.

I'm a bit like Scottmitch, believing that maybe a simple Atomic is enough there. But I'm surely missing something...

uglycow:
@fredericBregier @fenik17 @Scottmitch 
hey guys， the code is modified basing on your suggestion.  would u please review it  ; )

common/src/main/java/io/netty/util/concurrent/LimitLatch.java
Scottmitch: > i made the assumption that we won't run task in boss loop.

This is not a good assumption. Tasks may be run on the boss loop, and the "boss loop" may be used to handle "worker channels" too.

transport/src/main/java/io/netty/channel/socket/nio/NioServerSocketChannel.java
```diff
@@ -137,18 +152,34 @@ protected void doClose() throws Exception {
 
     @Override
     protected int doReadMessages(List<Object> buf) throws Exception {
-        SocketChannel ch = javaChannel().accept();
+        if (connectionCounter.incrementAndGet() > maxConnections) {
+            connectionCounter.decrementAndGet();
+            try {
+                //sleep a while so that we won't loop busily
+                Thread.sleep(10);
```
Scottmitch: I think we will need to use a different strategy here. We should actively shutdown/close the peer socket so:

1. the peer knows the connection has been rejected
2. our local backlog is drained and we can accept more connections in the future if we drop below the "max connections" threshold.

This will also be necessary in the case of Edge Triggered mode for the native transports. In this case we will not be notified by the selector that there is activity on the FD and so we may never accept sockets once we get into this condition.

The risk here is that we exceed the "max limit" temporarily (to accept a FD and then close it). This should be documented as part of the socket option.

normanmaurer:
I think what you want to archive canbe done without any changes in Netty.

I am currently on vacation but will come back to you next week so please not pull this in for now

transport/src/main/java/io/netty/channel/socket/nio/NioServerSocketChannel.java
uglycow: "fail fast" do have bring us these benefits. but accept and then close may not suit for our case.
we just switched from tomcat to netty http and get troubled by the unlimited connections. 
we are expecting a feature same as "maxConnections" in tomcat, so that limited connections get processed simultaneously. 
each connection won't stay long(http request without keep-alive). so the backlog serve as a buffer, just that the connection in it need no fd.
accept and close usually means an error at client, and the client had no choice but retry and so cost extra performance.
accept and close is a way to solve the original problem. but in this way , as @normanmaurer  said below, no need to change netty, we may maintain the counter in our biz-logic ;)

transport/src/main/java/io/netty/channel/socket/nio/NioServerSocketChannel.java
```diff
@@ -137,18 +152,34 @@ protected void doClose() throws Exception {
...
+            return 0;
+        }
 
+        SocketChannel ch = null;
         try {
+            ch = javaChannel().accept();
             if (ch != null) {
-                buf.add(new NioSocketChannel(this, ch));
+                NioSocketChannel nioSocketChannel = new NioSocketChannel(this, ch);
+                nioSocketChannel.closeFuture().addListener(closeListener);
```
johnou: Could be considered unnecessary overhead for users who limit max connections by other means.

Scottmitch:
To add to @normanmaurer's https://github.com/netty/netty/pull/6162#issuecomment-270187976 ... I was thinking this could potentially be built into a ChannelHandler or ChannelInitializer (for example) so we can still share code but doesn't necessarily have to be built into the core.

transport/src/main/java/io/netty/channel/socket/nio/NioServerSocketChannel.java
uglycow: u are right.   we  may save the listener if the maxconnection feature is not turned on ;)

normanmaurer:
Let me come up with a PR

transport/src/main/java/io/netty/channel/socket/nio/NioServerSocketChannel.java
```diff
@@ -137,18 +152,34 @@ protected void doClose() throws Exception {
 
     @Override
     protected int doReadMessages(List<Object> buf) throws Exception {
-        SocketChannel ch = javaChannel().accept();
+        if (connectionCounter.incrementAndGet() > maxConnections) {
+            connectionCounter.decrementAndGet();
+            try {
+                //sleep a while so that we won't loop busily
+                Thread.sleep(10);
+            }catch (InterruptedException exception){
```
fenik17: missing spaces.

normanmaurer:
Let me close this and open a new pr which will do the same with a handler.

dvlato:
Hi, did you create a new PR with a handler for this?

PR #3146, opened 2014-11-16
DNS server support

timboudreau:
Support for writing DNS servers, and various API changes to accomplish that, specifically:

 - `DnsType` -> `DnsType<RecordType>` where RecordType is a subclass of `DnsEntry`, and has payload decoders for specific types
 - Introduced record-specific subclasses of `DnsEntry` for common types, with fallback to the ByteBuf-based `DnsResource`
 - `DnsEntry` gains public `writeTo()` and `writePayload()` to write itself correctly into a ByteBuf
 - Added `DnsQueryDecoder` and `DnsResponseEncoder`
 - Introduced `NameWriter` for writing names, so that it's possible to implement a stateful `NameWriter` to do DNS name compression
 - `DnsMessage` is threadsafe (needed for proxy DNS servers where multiple calls to remote DNS servers may contribute partial responses)
    - If someone really feels strongly that creating the `List` instances it uses internally should be lazy, that could be done, but it would be good to have some evidence that really accomplishes something - feels like premature optimization
 - `DnsQuestion` is no longer a subclass of `DnsEntry` (which now has a time-to-live, since that is common to all other record types).
 - Added the package io.netty.handler.codec.dns.server with `DnsServerHandler`, a `SimpleChannelInboundHandler<DatagramPacket>` that handles the common plumbing of writing a DNS server that is unlikely to differ between implementations
 - Corrected error codes in `DnsResponseCode` to match https://tools.ietf.org/html/rfc2929 (they were sequential, but in fact 11-15 are unused and numbering restarts at 16)
 - Added error codes that exist but were not included in `DnsResponseCode` but are part of the spec
 - Added `DnsDecoderException` which can specify a `DnsResponseCode`
 - `DnsQuery` implements `Iterable<DnsQuestion>`, simply because it's intuitive
 - Reasonable `toString()` implementations on entity classes for logging purposes
 - Added to netty-examples:  DNS server, DNS client, proxy DNS server

PR #6170, opened 2017-01-02
Dns codec cleanup

timboudreau:
Motivation:

On the way to trying to reconcile my dns-server patches from last year with the current code, I noticed the current structure of DnsMessage and its subclasses requires boilerplate overloads and casts in implementation classes, which can be solved more cleanly with generics.

Modification:

Replaced overloads in DnsMessage subtypes, which were needed to alter the return type in overloaded methods, to instead parameterize the type on itself and return the type parameter - i.e. DnsMessage<M extends ReferenceCounted & DnsMessage>.

Result:

Implementations don't need casts, and supertype implementations of self-type returning methods do the right thing without needing overloads that just call the super method.

### DIFF
--- a/codec-dns/src/main/java/io/netty/handler/codec/dns/AbstractDnsMessage.java
+++ b/codec-dns/src/main/java/io/netty/handler/codec/dns/AbstractDnsMessage.java
@@ -30,7 +30,8 @@ import static io.netty.util.internal.ObjectUtil.checkNotNull;
 /**
  * A skeletal implementation of {@link DnsMessage}.
  */
-public abstract class AbstractDnsMessage extends AbstractReferenceCounted implements DnsMessage {
+public abstract class AbstractDnsMessage<M extends ReferenceCounted & DnsMessage<M>>
+        extends AbstractReferenceCounted implements DnsMessage<M> {
 
     private static final ResourceLeakDetector<DnsMessage> leakDetector =
             new ResourceLeakDetector<DnsMessage>(DnsMessage.class);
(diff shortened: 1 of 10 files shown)


timboudreau:
Umm, if master is deprecated, what branch should this pull request actually be against?

slandelle:
@timboudreau 4.1 (which is the default branch on Github)

timboudreau:
Hmm, my patches branch seems to contain a lot of 5.0 alpha commiys from
others when diffed against that. Do I need to start over from the 4.1
branch?

Scottmitch:
> Do I need to start over from the 4.1 branch?

@timboudreau - Thanks for the PR! What ever is easier for you. We currently only support the 4.1 and 4.0 branches.

Also keep in mind generally the API on these branches are considered "stable" and public facing API changes are generally not permitted. However the DNS resolver code is annotated with `@UnstableApi` so it is OK to change the API in this case.

timboudreau:
Understood.  This patch is a precursor to revamping my DNS server support patches in #3146 - going to try to do it as a layer on top of what's here, with minimal modifications to the existing code (given I have a day job I make no predictions about when that will be complete).

FWIW, the approach taken here could be used with `ResourceCounted` and `AbstractResourceCounted` as well - parameterize it on a subtype of itself to eliminate useless overloads just to alter the return type, e.g.

```
abstract class BaseFoo {
   public BaseFoo doSomething() { ... }
}
class RealFoo {
   public RealFoo doSomething() { return (RealFoo) super.doSomething(); }
}
```

with

```
abstract class BaseFoo<T extends BaseFoo> {
   public T doSomething() { ... }
```

so `RealFoo` doesn't need an overload.  Such a change would be binary (but not source) compatible.  Perhaps something to consider, as long as concrete classes don't expose the type parameter so we're not forcing people who just want to use concrete implementations to drag around the generic signatures of them for our convenience.  @normanmaurer - thoughts?

I didn't do that here since I'm keeping this package restricted to the dns code.

normanmaurer:
Sorry but we never picked this one up and so I will just close this as it is 2 years old now :(

PR #6171, opened 2017-01-03
InternalThreadLocalMap.arrayList should create a reusable ArrayList only if arrayList field is NULL.

lowka:
Motivation:

InternalThreadLocalMap.arrayList returns a new ArrayList every time it's called that defeats the purpose of having a reusable ArrayList.

Modification:

Modified InternalThreadLocalMap.arrayList to create an ArrayList only if arrayList field is NULL.

Result:

InternalThreadLocalMap.arrayList now creates a reusable ArrayList only if arrayList field is NULL.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/InternalThreadLocalMap.java
+++ b/common/src/main/java/io/netty/util/internal/InternalThreadLocalMap.java
@@ -192,14 +192,15 @@ public final class InternalThreadLocalMap extends UnpaddedInternalThreadLocalMap
         return arrayList(DEFAULT_ARRAY_LIST_INITIAL_CAPACITY);
     }
 
+    @SuppressWarnings("unchecked")
     public <E> ArrayList<E> arrayList(int minCapacity) {
         ArrayList<E> list = (ArrayList<E>) arrayList;
         if (list == null) {
-            list = (ArrayList<E>) new ArrayList<Object>(minCapacity);
-        } else {
-            list.clear();
-            list.ensureCapacity(minCapacity);
+            arrayList = new ArrayList<Object>(minCapacity);
+            return (ArrayList<E>) arrayList;
         }
+        list.clear();
+        list.ensureCapacity(minCapacity);
         return list;
     }
 


common/src/main/java/io/netty/util/internal/InternalThreadLocalMap.java
```diff
@@ -192,15 +192,17 @@ public StringBuilder stringBuilder() {
...
     }
 
+    @SuppressWarnings("unchecked")
     public <E> ArrayList<E> arrayList(int minCapacity) {
         ArrayList<E> list = (ArrayList<E>) arrayList;
         if (list == null) {
-            list = (ArrayList<E>) new ArrayList<Object>(minCapacity);
+            arrayList = new ArrayList<Object>(minCapacity);
+            return (ArrayList<E>) arrayList;
         } else {
```
Scottmitch: nit: we no longer need the `else` because we return in the `if` above.

normanmaurer:
Doh! Good catch

lowka:
@Scottmitch 
By accident.
I stumbled upon FastThreadLocalThread which javadocs says "it provides to fast thread local variables" so I decided to look into netty's internals to see how that works... :)

Scottmitch:
@lowka - Can you please sign the [ICLA](https://docs.google.com/forms/d/e/1FAIpQLSd7Bzje39G__THDJLRgKpQZ4gODNE26x_hZW3ofQOkgL6RGCA/viewform?formkey=dHBjc1YzdWhsZERUQnhlSklsbG1KT1E6MQ) referenced from the [contributor guidelines](http://netty.io/wiki/developer-guide.html)?

lowka:
@Scottmitch Done.

johnou:
Do we have any benchmarks to run to see how much this helps?

Scottmitch:
@johnou - not that I am aware of. It is only used in a few spots and may not be invoked for most users.

Scottmitch:
4.1 (a8950df) 4.0 (a6637be)

Scottmitch:
thanks again @lowka !

Issue #6153, opened 2016-12-22
Potential endless loop in Recycler.Stack.scavengeSome


zsxwing:
### Expected behavior

Recycler.Stack.scavengeSome should never run forever.

### Actual behavior

Recycler.Stack.scavengeSome may not stop and it prevents the event loop from running other tasks.

### Steps to reproduce

I cannot reproduce it. But I found out a cycle linked list of `io.netty.util.Recycler.WeakOrderQueue` when I was investigating an issue that a netty client was hung forever.

![screen shot 2016-12-21 at 5 21 52 pm](https://cloud.githubusercontent.com/assets/1000778/21412114/8eca8fd8-c7a2-11e6-9e9a-b7dc62a3314b.png)

The stack trace of this thread is 

```
"shuffle-client-7-4" daemon prio=5 tid=97 RUNNABLE
        at io.netty.util.Recycler$Stack.scavengeSome(Recycler.java:504)
           Local Variable: io.netty.util.Recycler$WeakOrderQueue#111
           Local Variable: io.netty.util.Recycler$WeakOrderQueue#112
        at io.netty.util.Recycler$Stack.scavenge(Recycler.java:454)
        at io.netty.util.Recycler$Stack.pop(Recycler.java:435)
        at io.netty.util.Recycler.get(Recycler.java:144)
           Local Variable: io.netty.util.Recycler$Stack#26
           Local Variable: io.netty.buffer.PooledUnsafeDirectByteBuf$1#1
        at io.netty.buffer.PooledUnsafeDirectByteBuf.newInstance(PooledUnsafeDirectByteBuf.java:39)
        at io.netty.buffer.PoolArena$DirectArena.newByteBuf(PoolArena.java:727)
        at io.netty.buffer.PoolArena.allocate(PoolArena.java:140)
           Local Variable: io.netty.buffer.PoolArena$DirectArena#4
           Local Variable: io.netty.buffer.PoolThreadCache#7
        at io.netty.buffer.PooledByteBufAllocator.newDirectBuffer(PooledByteBufAllocator.java:271)
        at io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:177)
        at io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:168)
        at io.netty.buffer.AbstractByteBufAllocator.ioBuffer(AbstractByteBufAllocator.java:129)
[...]

PR #6187, opened 2017-01-10
Add assert to ensure we not create an endless loop and fix unsafe pub…

normanmaurer:
…lication

Motivation:

[#6153] reports an endless loop that existed in the Recycler, while this was fixed adding a few asserts to ensure this remains fixed is a good thing. Beside this we also should ensure this can not escape the constructor to avoid unsafe publication.

Modifications:

- Add asserts
- Fix unsafe publication

Result:

More correct code.

### DIFF
--- a/common/src/main/java/io/netty/util/Recycler.java
+++ b/common/src/main/java/io/netty/util/Recycler.java
@@ -234,15 +234,24 @@ public abstract class Recycler<T> {
         private WeakOrderQueue(Stack<?> stack, Thread thread) {
             head = tail = new Link();
             owner = new WeakReference<Thread>(thread);
-            synchronized (stack) {
-                next = stack.head;
-                stack.head = this;
-            }
 
             // Its important that we not store the Stack itself in the WeakOrderQueue as the Stack also is used in
             // the WeakHashMap as key. So just store the enclosed AtomicInteger which should allow to have the
             // Stack itself GCed.
             availableSharedCapacity = stack.availableSharedCapacity;
+        }
+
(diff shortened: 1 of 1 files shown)


normanmaurer:
@zsxwing PTAL as well.

normanmaurer:
Some more details about unsafe publication:
https://shipilev.net/blog/2014/safe-public-construction/
http://vlkan.com/blog/post/2014/02/14/java-safe-publication/

zsxwing:
LGTM.

common/src/main/java/io/netty/util/Recycler.java
```diff
@@ -234,24 +234,33 @@ private WeakOrderQueue() {
...
+        static WeakOrderQueue newQueue(Stack<?> stack, Thread thread) {
+            WeakOrderQueue queue = new WeakOrderQueue(stack, thread);
+            // Done outside of the constructor to ensure WeakOrderQueue.this does not escape the constructor and so
+            // may be accessed while its still constructed.
+            stack.setHead(queue);
+            return queue;
+        }
+
+        private void setNext(WeakOrderQueue next) {
+            assert next != this;
```
zsxwing: This cannot capture the case in #6153 since it's a cycle linked list with multiple list nodes. However, it's better than nothing.

normanmaurer: Yep that said it still adds some safety

normanmaurer:
Cherry-picked into 4.0 (1e33365) and 4.1 (3c5e677)

Issue #6167, opened 2016-12-30
OpenSslEngine doesn't respond to close_notify


markdascher:
### Expected behavior

When a Netty server receives a TLS close_notify, it should respond with a close_notify.

### Actual behavior

Netty's [OpenSslEngine](https://github.com/netty/netty/blob/netty-4.1.6.Final/handler/src/main/java/io/netty/handler/ssl/OpenSslEngine.java) (actually [ReferenceCountedOpenSslEngine](https://github.com/netty/netty/blob/netty-4.1.6.Final/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java)) never sends the close_notify response.  When a close_notify comes in, it's processed inside of [unwrap()](https://github.com/netty/netty/blob/netty-4.1.6.Final/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java#L699), eventually reaching this [closeAll()](https://github.com/netty/netty/blob/netty-4.1.6.Final/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java#L845) call on line 845.  Then, [closeInbound()](https://github.com/netty/netty/blob/netty-4.1.6.Final/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java#L1007) calls [shutdown()](https://github.com/netty/netty/blob/netty-4.1.6.Final/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java#L359), which deallocates the OpenSSL data structures, before we have a chance for netty to grab the data.

The native Java [SSLEngine](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLEngine.html) works fine.  This [unwrap()](https://github.com/netty/netty/blob/netty-4.1.6.Final/handler/src/main/java/io/netty/handler/ssl/SslHandler.java#L968)
 call returns an SSLEngineResult where `status == CLOSED`, `handshakeStatus == NEED_WRAP`. As a result, [wrapNonAppData(ctx, true)](https://github.com/netty/netty/blob/netty-4.1.6.Final/handler/src/main/java/io/netty/handler/ssl/SslHandler.java#L1005) sets [needsFlush = true](https://github.com/netty/netty/blob/netty-4.1.6.Final/handler/src/main/java/io/netty/handler/ssl/SslHandler.java#L609), which lets netty know to flush the close_notify out to the network.

[...]

PR #6189, opened 2017-01-10
Ensure ReferenceCountedOpenSslEngine not swallow the close_notify

normanmaurer:
Motivation:

We need to ensure we not swallow the close_notify that should be send back to the remote peer. See [#6167]

Modifications:

Only call shutdown() in closeInbound() if there is nothing pending that should be send back to the remote peer.

Result:

close_notify is correctly send back to the remote peer.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -544,11 +544,14 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
             final SSLEngineResult.Status rs;
 
             // If isOutboundDone, then the data from the network BIO
-            // was the close_notify message and all was consumed we are not required to wait
-            // for the receipt the peer's close_notify message -- shutdown.
+            // was the close_notify message, see if we also received the response yet.
             if (isOutboundDone()) {
                 rs = CLOSED;
-                shutdown();
+                if (isInboundDone()) {
+                    // If the inbound was done as well, we need to ensure we return NOT_HANDSHAKING to signal we are
+                    // done.
+                    hs = NOT_HANDSHAKING;
+                }
             } else {
                 rs = OK;
             }
@@ -1054,7 +1057,11 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
 
         isInboundDone = true;
 
-        shutdown();
+        if (isOutboundDone()) {
+            // Only call shutdown if there is no outbound data pending.
+            // See https://github.com/netty/netty/issues/6167
+            shutdown();
+        }
 
         if (handshakeState != HandshakeState.NOT_STARTED && !receivedShutdown) {
             throw new SSLException(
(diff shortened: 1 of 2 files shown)


normanmaurer:
Also @markdascher ....

Scottmitch:
Do we also need to modify `status` to return `SSLEngineResult.HandshakeStatus#WRAP` if there is a close_notify pending [1]?

[1] https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java#L903

normanmaurer:
@Scottmitch it already returns `NEED_WRAP` (verified with print statements)

Scottmitch:
@normanmaurer - is it feasible for to provide a unit test for this case?

normanmaurer:
@Scottmitch I hope yes... gimme a day and I see if I can make it happen

normanmaurer:
@Scottmitch the good news is I was able to write a test... that bad is that it showed another "edge-case"... Working on it as we speak ...

normanmaurer:
@Scottmitch @markdascher PTAL again.. with unit test and everything ;)

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -659,6 +662,7 @@ public final SSLEngineResult wrap(
                                 // to read more data [1]. It is also possible that event loop will detect the socket
                                 // has been closed. [1] https://www.openssl.org/docs/manmaster/ssl/SSL_write.html
                                 pendingNetResult = readPendingBytesFromBIO(dst, bytesConsumed, bytesProduced, status);
+
```
Scottmitch: nit: revert to be consistent with other switch cases?

handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
```diff
@@ -1214,4 +1214,113 @@ private static void testCloseInboundAfterBeginHandshake(SSLEngine engine) throws
...
+            assertEquals(0, encryptedClientToServer.remaining());
+            assertEquals(0, plainServerOut.remaining());
+
+            assertFalse(server.isOutboundDone());
+            assertTrue(server.isInboundDone());
+
+            result = server.wrap(empty, encryptedServerToClient);
+            encryptedServerToClient.flip();
+            assertEquals(SSLEngineResult.Status.CLOSED, result.getStatus());
+            // We are done no futher UNWRAP/WRAP expected
```
Scottmitch: futher -> further

or just `UNWRAP/WRAP are not expected after this point`

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: fixed

handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
normanmaurer: fixed

normanmaurer:
Cherry-picked into 4.1 (dd055c0) and 4.0 (bb6b251)

PR #6194, opened 2017-01-10
Correct fail write with NotYetConnectedException when OioDatagramChan…

normanmaurer:
…nel is not connected yet.

Motivation:

NioDatagramChannel fails a write with NotYetConnectedException when the DatagramChannel was not yet connected and a ByteBuf is written. The same should be done for OioDatagramChannel as well.

Modifications:

Make OioDatagramChannel consistent with NioDatagramChannel

Result:

Correct and consistent implementations of DatagramChannel

### DIFF
--- a/transport/src/main/java/io/netty/channel/socket/oio/OioDatagramChannel.java
+++ b/transport/src/main/java/io/netty/channel/socket/oio/OioDatagramChannel.java
@@ -44,6 +44,7 @@ import java.net.NetworkInterface;
 import java.net.SocketAddress;
 import java.net.SocketException;
 import java.net.SocketTimeoutException;
+import java.nio.channels.NotYetConnectedException;
 import java.util.List;
 import java.util.Locale;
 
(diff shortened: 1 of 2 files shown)


normanmaurer:
@Scottmitch @nmittler @trustin PTAL

transport/src/main/java/io/netty/channel/socket/oio/OioDatagramChannel.java
```diff
@@ -253,17 +254,23 @@ protected void doWrite(ChannelOutboundBuffer in) throws Exception {
...
-                tmpPacket.setData(data.array(), data.arrayOffset() + data.readerIndex(), length);
-            } else {
-                byte[] tmp = new byte[length];
-                data.getBytes(data.readerIndex(), tmp);
-                tmpPacket.setData(tmp);
-            }
             try {
+                if (remoteAddress != null) {
+                    tmpPacket.setSocketAddress(remoteAddress);
+                } else {
```
nmittler: nit: `else if`?

normanmaurer:
Cherry-picked into 4.1 (8a5e42ad2ea091ccc88e02993927f9dd9d4c8205) and 4.0 (fb21891630808f91612ce13deaa219f551b869fb)

transport/src/main/java/io/netty/channel/socket/oio/OioDatagramChannel.java
normanmaurer: addressed

Issue #6152, opened 2016-12-21
HttpProxyHandler incorrectly formats IPv6 host and port in CONNECT request


ZuluForce:
### Expected behavior
The HttpProxyHandler is expected to be capable of issuing a valid CONNECT request for a tunneled connection to an IPv6 host. In this case we are passing an IPv6 address (eg fd00:c0de:42::c:293a:5736) rather than a host name.

### Actual behavior
The HttpProxyHandler does not properly concatenate the IPv6 address and port. The resulting error after we fail to connect will show you the problem:

```
io.netty.handler.proxy.ProxyConnectException: http, none, /fd00:c0de:42:0:5:562d:d54:1:3128 => /fd00:c0de:42:0:50:5694:2fda:1:4287, status: 503 Service Unavailable
```

In both cases you can see the IPv6 address is formatted without brackets: `fd00:c0de:42:0:50:5694:2fda:1:4287` should be `[fd00:c0de:42:0:50:5694:2fda:1]:4287`. This is just an exception message so it doesn't prove it's formatting incorrectly. However, if you look at the request on the wire you can see it is certainly wrong:

```
    CONNECT fd00:c0de:42:0:50:5694:2fda:1:4287 HTTP/1.1\r\n
    host: fd00:c0de:42:0:50:5694:2fda:1:4287\r\n
    \r\n
```

Here is the problem method from HttpProxyHandler:

```
    @Override
    protected Object newInitialMessage(ChannelHandlerContext ctx) throws Exception {
        InetSocketAddress raddr = destinationAddress();
        String rhost;
        if (raddr.isUnresolved()) {
            rhost = raddr.getHostString();
        } else {
            rhost = raddr.getAddress().getHostAddress();
        }

        final String host = rhost + ':' + raddr.getPort();
        FullHttpRequest req = new DefaultFullHttpRequest(
                HttpVersion.HTTP_1_1, HttpMethod.CONNECT,
                host,
                Unpooled.EMPTY_BUFFER, false);

        req.headers().set(HttpHeaderNames.HOST, host);

        if (authorization != null) {
            req.headers().set(HttpHeaderNames.PROXY_AUTHORIZATION, authorization);
        }

        return req;
    }
```

Specifically: ` final String host = rhost + ':' + raddr.getPort();`

### Steps to reproduce
* Setup an HTTP Proxy with IPv6.
* Setup a target server with an IPv6 address.
[...]

PR #6195, opened 2017-01-10
Correctly handle IPV6 in HttpProxyHandler

normanmaurer:
Motivation:

The HttpProxyHandler is expected to be capable of issuing a valid CONNECT request for a tunneled connection to an IPv6 host.

Modifications:

- Correctly format the IPV6 address.
- Add unit tests

Result:

HttpProxyHandler works with IPV6 as well. Fixes [#6152].

### DIFF
--- a/common/src/main/java/io/netty/util/NetUtil.java
+++ b/common/src/main/java/io/netty/util/NetUtil.java
@@ -25,6 +25,7 @@ import java.io.FileReader;
 import java.net.Inet4Address;
 import java.net.Inet6Address;
 import java.net.InetAddress;
+import java.net.InetSocketAddress;
 import java.net.NetworkInterface;
 import java.net.SocketException;
 import java.net.UnknownHostException;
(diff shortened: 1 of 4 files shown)


normanmaurer:
@ZuluForce PTAL as well

ZuluForce:
@normanmaurer LGTM Thanks for taking this on.

common/src/main/java/io/netty/util/NetUtil.java
```diff
@@ -934,6 +935,25 @@ public static Inet6Address getByName(CharSequence ip, boolean ipv4Mapped) {
...
+    public static String toSocketAddressString(InetSocketAddress addr) {
+        final String rhost;
+        if (addr.isUnresolved()) {
+            String hostString = PlatformDependent.javaVersion() >= 7 ? addr.getHostString() : addr.getHostName();
+            rhost = isValidIpV6Address(hostString) ? '[' + hostString + ']' : hostString;
+        } else {
+            InetAddress address = addr.getAddress();
+            rhost = address instanceof Inet4Address ? toAddressString(address) : '[' + toAddressString(address) + ']';
+        }
+        return rhost + ':' + addr.getPort();
```
Scottmitch: I wonder if it would should be more explicit about the string concatenation operations used here. For example it seems unlikely JIT will be able to combine all of these conditional concatenation operations into a single `StringBuilder`. Can we create a single `StringBuilder` and manually do the `append` calls (e.g. including what is now building `rhost`) to be sure we don't create multiple `String`/`StringBuilder` operations?

common/src/main/java/io/netty/util/NetUtil.java
```diff
@@ -934,6 +935,39 @@ public static Inet6Address getByName(CharSequence ip, boolean ipv4Mapped) {
...
+            InetAddress address = addr.getAddress();
+            String hostString = toAddressString(address);
+            sb = newSocketAddressStringBuilder(hostString, port, address instanceof Inet4Address);
+        }
+        return sb.append(':').append(port).toString();
+    }
+
+    private static StringBuilder newSocketAddressStringBuilder(String hostString, String port, boolean ipv4) {
+        if (ipv4) {
+            // Need to include enought space for hostString:port .
```
Scottmitch: `enought` -> `enough`

Scottmitch: also remove the space before the period?

common/src/main/java/io/netty/util/NetUtil.java
```diff
@@ -934,6 +935,39 @@ public static Inet6Address getByName(CharSequence ip, boolean ipv4Mapped) {
...
+            return new StringBuilder(hostString.length() + 1 + port.length()).append(hostString);
+        } else {
+            // Need to include enought space for [hostString]:port .
```
Scottmitch: `enought` -> `enough`

also remove the space before the period?

common/src/main/java/io/netty/util/NetUtil.java
```diff
@@ -934,6 +935,39 @@ public static Inet6Address getByName(CharSequence ip, boolean ipv4Mapped) {
...
+            sb = newSocketAddressStringBuilder(hostString, port, address instanceof Inet4Address);
+        }
+        return sb.append(':').append(port).toString();
+    }
+
+    private static StringBuilder newSocketAddressStringBuilder(String hostString, String port, boolean ipv4) {
+        if (ipv4) {
+            // Need to include enought space for hostString:port .
+            return new StringBuilder(hostString.length() + 1 + port.length()).append(hostString);
+        } else {
```
Scottmitch: nit: no need for `else`

common/src/main/java/io/netty/util/NetUtil.java
normanmaurer: I added the space before to make it clear it is not part of it... But I guess its ok to remove it as well.

normanmaurer:
Cherry-picked into 4.1 (eb5dc4b)

Issue #6163, opened 2016-12-29
NPE in DefaultHttp2Connection$DefaultStream.takeChild(DefaultHttp2Connection.java:629)


rkapsi:
First and only time I've seen this.

```java
java.lang.NullPointerException: null
  at io.netty.handler.codec.http2.DefaultHttp2Connection$DefaultStream.takeChild(DefaultHttp2Connection.java:629) ~[netty-all-4.1.7.Final-SNAPSHOT.jar:4.1.7.Final-SNAPSHOT]
  at io.netty.handler.codec.http2.DefaultHttp2Connection$DefaultStream.removeChild(DefaultHttp2Connection.java:670) ~[netty-all-4.1.7.Final-SNAPSHOT.jar:4.1.7.Final-SNAPSHOT]
  at io.netty.handler.codec.http2.DefaultHttp2Connection.removeStream(DefaultHttp2Connection.java:294) ~[netty-all-4.1.7.Final-SNAPSHOT.jar:4.1.7.Final-SNAPSHOT]
  at io.netty.handler.codec.http2.DefaultHttp2Connection$ActiveStreams.removeFromActiveStreams(DefaultHttp2Connection.java:1192) ~[netty-all-4.1.7.Final-SNAPSHOT.jar:4.1.7.Final-SNAPSHOT]
  at io.netty.handler.codec.http2.DefaultHttp2Connection$ActiveStreams.deactivate(DefaultHttp2Connection.java:1138) ~[netty-all-4.1.7.Final-SNAPSHOT.jar:4.1.7.Final-SNAPSHOT]
  at io.netty.handler.codec.http2.DefaultHttp2Connection$DefaultStream.close(DefaultHttp2Connection.java:534) ~[netty-all-4.1.7.Final-SNAPSHOT.jar:4.1.7.Final-SNAPSHOT]
  at io.netty.handler.codec.http2.DefaultHttp2Connection.close(DefaultHttp2Connection.java:149) ~[netty-all-4.1.7.Final-SNAPSHOT.jar:4.1.7.Final-SNAPSHOT]
  at io.netty.handler.codec.http2.Http2ConnectionHandler$BaseDecoder.channelInactive(Http2ConnectionHandler.java:185) ~[netty-all-4.1.7.Final-SNAPSHOT.jar:4.1.7.Final-SNAPSHOT]
[...]

PR #6197, opened 2017-01-10
DefaultHttp2Connection modifying child map while iterating

Scottmitch:
Motivation:
When DefaultHttp2Connection removes a stream it iterates over all children and adds them as children to the parent of the stream being removed. This process may remove elements from the child map while iterating without using the iterator's remove() method. This is generally unsafe and may result in an undefined iteration.

Modifications:
- We should use the Iterator's remove() method while iterating over the child map

Result:
Fixes https://github.com/netty/netty/issues/6163

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
@@ -79,8 +79,10 @@ public class DefaultHttp2Connection implements Http2Connection {
      * the assumption that most streams will have a small number of children. This choice may be
      * sub-optimal if when children are present there are many children (i.e. a web page which has many
      * dependencies to load).
+     *
+     * Visible only for testing!
      */
-    private static final int INITIAL_CHILDREN_MAP_SIZE =
+    static final int INITIAL_CHILDREN_MAP_SIZE =
             max(1, SystemPropertyUtil.getInt("io.netty.http2.childrenMapSize", 4));
 
     /**
@@ -609,7 +611,7 @@ public class DefaultHttp2Connection implements Http2Connection {
          * This method is intended to be used to support an exclusive priority dependency operation.
          * @return The map of children prior to this operation, excluding {@code streamToRetain} if present.
          */
-        private IntObjectMap<DefaultStream> retain(DefaultStream streamToRetain) {
+        private IntObjectMap<DefaultStream> removeAllChildrenExcept(DefaultStream streamToRetain) {
             streamToRetain = children.remove(streamToRetain.id());
             IntObjectMap<DefaultStream> prevChildren = children;
             // This map should be re-initialized in anticipation for the 1 exclusive child which will be added.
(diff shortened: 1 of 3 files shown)


Scottmitch:
@rkapsi - FYI. Please verify if possible (maybe not because you haven't been able to reproduce).

common/src/main/templates/io/netty/util/collection/KObjectHashMap.template
```diff
@@ -608,7 +606,7 @@ public class @K@ObjectHashMap<V> implements @K@ObjectMap<V> {
             if (nextIndex == -1) {
                 scanNext();
             }
-            return nextIndex < keys.length;
+            return nextIndex != values.length;
```
normanmaurer: @Scottmitch so we are sure nextIndex can never be > values.length ?  Just want to make sure we are :)

common/src/main/templates/io/netty/util/collection/KObjectHashMap.template
```diff
@@ -627,7 +625,7 @@ public class @K@ObjectHashMap<V> implements @K@ObjectMap<V> {
 
         @Override
         public void remove() {
-            if (prevIndex < 0) {
+            if (prevIndex == -1) {
```
normanmaurer: @Scottmitch so we are sure prevIndex can never be < -1 ?  Just want to make sure we are :)

common/src/main/templates/io/netty/util/collection/KObjectHashMap.template
Scottmitch: I believe so (correct me if I'm missing something). we never reduce the size of `values` and increment this index in `scanNext()` until we hit `values.length`. We may set `nextIndex` to the `prevIndex` but that is a previously accessed index which must be `< values.length`

common/src/main/templates/io/netty/util/collection/KObjectHashMap.template
Scottmitch: I think so. its either set to `-1` at init time, an index in an array (non-negative), or `-1` if the previous index has been removed.

Am I missing something?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
```diff
@@ -80,7 +80,7 @@
      * sub-optimal if when children are present there are many children (i.e. a web page which has many
      * dependencies to load).
      */
-    private static final int INITIAL_CHILDREN_MAP_SIZE =
+    static final int INITIAL_CHILDREN_MAP_SIZE =
```
nmittler: nit: maybe add a comment that it's only visible for testing?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
```diff
@@ -625,17 +625,17 @@ final void weight(short weight) {
...
 
             if (oldParent != this) {
                 events.add(new ParentChangedEvent(child, oldParent));
                 notifyParentChanging(child, this);
                 child.parent = this;
-                // Note that the removal operation may not be successful and may return null. This is because when an
-                // exclusive dependency is processed the children are removed in a previous recursive call but the
-                // child's parent link is updated here.
-                if (oldParent != null) {
+                if (childItr != null) {
```
nmittler: Maybe add comments here to explain these 2 cases?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
```diff
@@ -649,12 +649,17 @@ final void takeChild(DefaultStream child, boolean exclusive, List<ParentChangedE
             if (exclusive && !children.isEmpty()) {
                 // If it was requested that this child be the exclusive dependency of this node,
                 // move any previous children to the child node, becoming grand children of this node.
-                for (DefaultStream grandchild : retain(child).values()) {
-                    child.takeChild(grandchild, false, events);
+                Iterator<PrimitiveEntry<DefaultStream>> itr = retain(child).entries().iterator();
```
nmittler: nit: we really should consider renaming the `retain` method.  I was scratching my head for a while trying to understand this code :).  Not something we need to do now, but maybe something like `removeAllChildrenExcept` or `removeAllOtherChildren`.

common/src/main/templates/io/netty/util/collection/KObjectHashMap.template
nmittler: I think this is fine.

common/src/main/templates/io/netty/util/collection/KObjectHashMap.template
nmittler: nope - it's fine.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
Scottmitch: done

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
Scottmitch: done

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
Scottmitch: done. `removeAllChildrenExcept` sgtm

Scottmitch:
4.1 (ec3d077)

Issue #6173, opened 2017-01-04
Native epoll transport should honor AUTO_READ flag


wazhar:
### Expected behavior
Data should only be read when requested, if AUTO_READ option is set to false  
### Actual behavior
When using Epoll transport, netty calls handler's channelRead0 without requesting read, even if AUTO_READ option is set to false. NIO Transport on the other hand works as expected.
This could be a problem if you are acting as a proxy. 
### Steps to reproduce
This is reproducible using small sized data I could reproduce it starting from 1 byte till 150k. Client must close connection as soon as done writing. Here are some settings I used.

1) SO_RCVBUF option was not set, although setting it didn't help.
2) AUTO_READ false
3) TCP_NODELAY false
4) SO_REUSEADDR true 
5) cat /proc/sys/net/ipv4/tcp_rmem
4096	87380	6291456
6) /proc/sys/net/ipv4/tcp_wmem
4096	16384	4194304

This could help generating test data quickly:
cp some_large_file 10k_file
truncate -s 10240 10k_file
nc host port < 10k_file

### Netty version
4.1.6.Final
### JVM version 
java version "1.8.0_65"
Java(TM) SE Runtime Environment (build 1.8.0_65-b17)
Java HotSpot(TM) 64-Bit Server VM (build 25.65-b01, mixed mode)
### OS version 
Linux 4.8.15-300.fc25.x86_64 
x86_64 x86_64 x86_64 GNU/Linux

blueswhisper:
may be you should use Epoll level trigger, default Epoll Mode is edge trigger,  [see this](https://github.com/netty/netty/blob/4.1/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollChannelConfig.java#L127)

wazhar:
Setting LEVEL_TRIGGERED bootstrap option did not help.

Scottmitch:
> Client must close connection as soon as done writing.

IIRC if we get a EPOLLRDHUP then we must read until we get a socket error for stream sockets [1] while in ET mode as we may not be notified of activity on that FD again. This may not be necessary in LT mode, and especially if NIO transport is still able to deliver all data. I will need to investigate more.

[1] https://github.com/netty/netty/blob/4.1/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollRecvByteAllocatorHandle.java#L57

PR #6199, opened 2017-01-10
EPOLL RDHUP processing

Scottmitch:
Motivation:
EpollRecvByteAllocatorHandle will read unconditionally if EPOLLRDHUP has been received. However we can just treat this the same was we do as data maybe pending in ET mode, and let LT mode notify us if we haven't read all data.

Modifications:
- EpollRecvByteAllocatorHandle should not always force a read just because EPOLLRDHUP has been received, but just treated as an indicator that there maybe more data to read in ET mode

Result:
Fixes https://github.com/netty/netty/issues/6173.

### DIFF
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollRecvByteAllocatorHandle.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollRecvByteAllocatorHandle.java
@@ -32,8 +32,13 @@ class EpollRecvByteAllocatorHandle extends RecvByteBufAllocator.DelegatingHandle
         receivedRdHup = true;
     }
 
+    final boolean isReceivedRdHup() {
+        return receivedRdHup;
+    }
+
     boolean maybeMoreDataToRead() {
-        return isEdgeTriggered && lastBytesRead() > 0;
+        // If EPOLLRDHUP has been received we must read until we get a read error.
+        return isEdgeTriggered && (lastBytesRead() > 0 || receivedRdHup);
     }
 
     final void edgeTriggered(boolean edgeTriggered) {
@@ -53,9 +58,7 @@ class EpollRecvByteAllocatorHandle extends RecvByteBufAllocator.DelegatingHandle
          * continue to avoid a {@link StackOverflowError} between channelReadComplete and reading from the
          * channel. It is expected that the {@link #EpollSocketChannel} implementations will track if we are in
          * edgeTriggered mode and all data was not read, and will force a EPOLLIN ready event.
-         *
-         * If EPOLLRDHUP has been received we must read until we get a read error.
          */
-        return receivedRdHup || maybeMoreDataToRead() && config.isAutoRead() || super.continueReading();
+        return maybeMoreDataToRead() && config.isAutoRead() || super.continueReading();
     }
 }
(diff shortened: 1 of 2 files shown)


Scottmitch:
@wazhar - FYI

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollRecvByteAllocatorStreamingHandle.java
```diff
@@ -28,7 +28,9 @@ boolean maybeMoreDataToRead() {
         /**
          * For stream oriented descriptors we can assume we are done reading if the last read attempt didn't produce
          * a full buffer (see Q9 in <a href="http://man7.org/linux/man-pages/man7/epoll.7.html">epoll man</a>).
+         *
+         * If EPOLLRDHUP has been received we must read until we get a read error.
```
normanmaurer: Just a nit: `{@code EPOLLRDHUP}`... Feel free to ignore.

normanmaurer:
@Scottmitch just one NIT... LGTM

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollRecvByteAllocatorStreamingHandle.java
Scottmitch: I think I will just leave it because it doesn't actually refer to any code.

Scottmitch:
4.1 (631077c)

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollRecvByteAllocatorHandle.java
```diff
@@ -32,8 +32,13 @@ final void receivedRdHup() {
...
     }
 
+    final boolean isReceivedRdHup() {
+        return receivedRdHup;
+    }
+
     boolean maybeMoreDataToRead() {
-        return isEdgeTriggered && lastBytesRead() > 0;
+        // If EPOLLRDHUP has been received we must read until we get a read error.
+        return isEdgeTriggered && (lastBytesRead() > 0 || receivedRdHup);
```
fenik17: Is it possible to swap conditions "lastBytesRead() > 0" and "receivedRdHup" (receivedRdHup has faster access)?

normanmaurer: @Scottmitch WDYT ? I think receivedRdHup is false more often then lastBytesRead() <= 0, so I think its fine as it is. /cc @fenik17

Scottmitch: `receivedRdHup` is typically only true as the socket is being closed so the `lastBytesRead() > 0` condition is more likely to be the determining factor in the conditional.

normanmaurer: +1... that is what I assumed as well. So just leave it

johnou:
@Scottmitch 4.0 pretty please?

Scottmitch:
@johnou - This code is actually quite a bit different in 4.0. Doing this in 4.0 would be a bit more involved as we don't support auto read in all scenarios (ET mode).

johnou:
@Scottmitch np let's leave it as is.

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollRecvByteAllocatorHandle.java
fenik17: @Scottmitch @normanmaurer 
Ok! Thanks for clarify!

Issue #6165, opened 2016-12-30
Netty HTTP2 Request is timeout


zmyer:
### Expected behavior
When I run Netty Http2 helloworld example, it happened timeout request, How to do that, the example all run in localhost, and timeout set 5 seconds. It seem Netty Http2 codec module not too stable.

### Actual behavior
Message received for unknown stream id 864103
Exception in thread "main" java.lang.IllegalStateException: Timed out waiting for response on stream id 864103
	at io.netty.example.http2.helloworld.client.HttpResponseHandler.awaitResponses(HttpResponseHandler.java:77)
	at io.netty.example.http2.helloworld.client.Http2Client.main(Http2Client.java:126)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:147)

### Netty version
4.1.6.Final

### JVM version (e.g. `java -version`)
jdk1.8.0_91-b14

### OS version (e.g. `uname -a`)
Mac OS X EI Capitan 10.11.1

Scottmitch:
@zmyer - Can you please list the steps to reproduce this issue? Are you running the `http2-server` and `http2-client` on the same host? If so how are you running them and with what arguments?

zmyer:
@Scottmitch, yes, in the same host and all in default.  I run http2 example repeatedly. My IDE is IDEA 2016.

Scottmitch:
I am unable to reproduce this issue. Please explicitly list the steps required to reproduce this issue.

zmyer:
en, Ok, now I show such example below, and you can run it much more time. 
at such example, I run in local, exception still happen,
Message received for unknown stream id 1857213
Exception in thread "main" java.lang.IllegalStateException: Timed out waiting for response on stream id 1857213
	at io.netty.example.http2.helloworld.client.HttpResponseHandler.awaitResponses(HttpResponseHandler.java:81)
	at io.netty.example.http2.helloworld.client.Http2Client.main(Http2Client.java:126)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #6200, opened 2017-01-10
HTTP/2 HelloWorld Client Example Bug

Scottmitch:
Motivation:
The HTTP/2 helloworld client example has 2 bugs:
1. HttpResponseHandler has a map which is accessed from multiple threads, but the map is not thread safe.
2. Requests are flushed and maybe completely written and the responses may be received/processed by Netty before an element is inserted into the HttpResponseHandler map. This may result in an 'unexpected message' error even though the message has actually been sent.

Modifications:
- HttpResponseHandler should use a thread safe map
- Http2Client shouldn't flush until entries are added to the HttpResponseHandler map

Result:
Fixes https://github.com/netty/netty/issues/6165.

### DIFF
--- a/example/src/main/java/io/netty/example/http2/helloworld/client/Http2Client.java
+++ b/example/src/main/java/io/netty/example/http2/helloworld/client/Http2Client.java
@@ -15,7 +15,6 @@
 package io.netty.example.http2.helloworld.client;
 
 import io.netty.bootstrap.Bootstrap;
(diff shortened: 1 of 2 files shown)


Scottmitch:
@zmyer @vongosling - FYI

vongosling:
Cool, we will have  a try :-)

Scottmitch:
4.1 (d771526)

PR #6201, opened 2017-01-10
OpenSSL ByteBuffer BIO

Scottmitch:
Motivation:
Currently Netty utilizes BIO_new_bio_pair so we can control all FD lifetime and event notification but delegates to OpenSSL for encryption/decryption. The current implementation sets up a pair of BIO buffers to read/write encrypted/plaintext data. This approach requires copying of data from Java ByteBuffers to native memory BIO buffers, and also requires both BIO buffers to be sufficiently large to hold application data. If direct ByteBuffers are used we can avoid coyping to/from the intermediate BIO buffer and just read/write directly from the direct ByteBuffer memory. We still need an internal buffer because OpenSSL may generate write data as a result of read calls (e.g. handshake, alerts, renegotiation, etc..), but this buffer doesn't have to be be large enough to hold application data.

Modifications:
- Take advantage of the new ByteBuffer based BIO provided by netty-tcnative instead of using BIO_read and BIO_write.

Result:
Less copying and lower memory footprint requirement per TLS connection.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
@@ -17,6 +17,9 @@ package io.netty.handler.ssl;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
+import io.netty.tcnative.jni.CertificateVerifier;
+import io.netty.tcnative.jni.SSL;
+import io.netty.tcnative.jni.SSLContext;
 import io.netty.util.AbstractReferenceCounted;
 import io.netty.util.ReferenceCounted;
 import io.netty.util.ResourceLeakDetector;
@@ -27,9 +30,6 @@ import io.netty.util.internal.StringUtil;
 import io.netty.util.internal.SystemPropertyUtil;
 import io.netty.util.internal.logging.InternalLogger;
 import io.netty.util.internal.logging.InternalLoggerFactory;
-import io.netty.tcnative.jni.CertificateVerifier;
-import io.netty.tcnative.jni.SSL;
-import io.netty.tcnative.jni.SSLContext;
 
 import java.security.AccessController;
 import java.security.PrivateKey;
@@ -45,7 +45,6 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-
 import javax.net.ssl.KeyManager;
 import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLException;
@@ -57,6 +56,7 @@ import javax.net.ssl.X509KeyManager;
 import javax.net.ssl.X509TrustManager;
 
 import static io.netty.util.internal.ObjectUtil.checkNotNull;
+import static io.netty.util.internal.ObjectUtil.checkPositiveOrZero;
 
 /**
  * An implementation of {@link SslContext} which works with libraries that support the
@@ -85,6 +85,17 @@ public abstract class ReferenceCountedOpenSslContext extends SslContext implemen
                     return SystemPropertyUtil.getBoolean("jdk.tls.rejectClientInitiatedRenegotiation", false);
                 }
             });
+
+    private static final int DEFAULT_BIO_NON_APPLICATION_BUFFER_SIZE =
+            AccessController.doPrivileged(new PrivilegedAction<Integer>() {
+                @Override
+                public Integer run() {
+                    return Math.max(1,
+                            SystemPropertyUtil.getInt("io.netty.handler.ssl.openssl.bioNonApplicationBufferSize",
+                                                      2048));
+                }
+            });
+
     private static final List<String> DEFAULT_CIPHERS;
     private static final Integer DH_KEY_LENGTH;
     private static final ResourceLeakDetector<ReferenceCountedOpenSslContext> leakDetector =
@@ -130,7 +141,8 @@ public abstract class ReferenceCountedOpenSslContext extends SslContext implemen
     final Certificate[] keyCertChain;
     final ClientAuth clientAuth;
     final OpenSslEngineMap engineMap = new DefaultOpenSslEngineMap();
-    volatile boolean rejectRemoteInitiatedRenegotiation;
+    private volatile boolean rejectRemoteInitiatedRenegotiation;
+    private volatile int bioNonApplicationBufferSize = DEFAULT_BIO_NON_APPLICATION_BUFFER_SIZE;
 
     static final OpenSslApplicationProtocolNegotiator NONE_PROTOCOL_NEGOTIATOR =
             new OpenSslApplicationProtocolNegotiator() {
@@ -266,7 +278,7 @@ public abstract class ReferenceCountedOpenSslContext extends SslContext implemen
                 SSLContext.setOptions(ctx, SSL.SSL_OP_SINGLE_DH_USE);
                 SSLContext.setOptions(ctx, SSL.SSL_OP_NO_SESSION_RESUMPTION_ON_RENEGOTIATION);
 
-                // We do not support compression as the moment so we should explicitly disable it.
+                // We do not support compression at the moment so we should explicitly disable it.
                 SSLContext.setOptions(ctx, SSL.SSL_OP_NO_COMPRESSION);
 
                 // Disable ticket support by default to be more inline with SSLEngineImpl of the JDK.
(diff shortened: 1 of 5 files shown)


Scottmitch:
- only the last commit is part of this PR
- this PR depends upon https://github.com/netty/netty-tcnative/pull/216

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -812,9 +763,18 @@ public final SSLEngineResult unwrap(
             }
 
             if (len < SslUtils.SSL_RECORD_HEADER_LENGTH) {
-                return new SSLEngineResult(BUFFER_UNDERFLOW, getHandshakeStatus(), 0, 0);
-            }
-
+                return newResult(BUFFER_UNDERFLOW, 0, 0, status);
+            }
+
+            // TODO(scott): limiting to a single packet can result in thrashing in the pipeline to decode application
```
Scottmitch: @normanmaurer - this is meant to be a discussion point.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -845,9 +820,18 @@ public final SSLEngineResult unwrap(
             }
 
             if (len < SslUtils.SSL_RECORD_HEADER_LENGTH) {
-                return new SSLEngineResult(BUFFER_UNDERFLOW, getHandshakeStatus(), 0, 0);
-            }
-
+                return newResultMayFinishHandshake(BUFFER_UNDERFLOW, status, 0, 0);
+            }
+
+            // TODO(scott): limiting to a single packet can result in thrashing in the pipeline to decode application
```
Scottmitch: @normanmaurer - this is meant to be a discussion point.

pom.xml
```diff
@@ -244,7 +244,7 @@
     <!-- Fedora-"like" systems. This is currently only used for the netty-tcnative dependency -->
     <os.detection.classifierWithLikes>fedora</os.detection.classifierWithLikes>
     <tcnative.artifactId>netty-tcnative</tcnative.artifactId>
-    <tcnative.version>2.0.0.Beta1</tcnative.version>
+    <tcnative.version>2.0.0.Final-SNAPSHOT</tcnative.version>
```
Scottmitch: this must change

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -727,26 +665,63 @@ public final SSLEngineResult wrap(
...
+            } finally {
+                SSL.bioClearByteBuffer(networkBIO);
+                if (bioReadCopyBuf == null) {
+                    dst.position(dst.position() + bytesProduced);
+                } else {
+                    assert bioReadCopyBuf.readableBytes() <= dst.remaining() : "The destination buffer " + dst +
+                            " didn't have enough remaining space to hold the encrypted content in " + bioReadCopyBuf;
+                    dst.put(bioReadCopyBuf.internalNioBuffer(bioReadCopyBuf.readerIndex(),
+                                                      bioReadCopyBuf.readerIndex() + bytesProduced));
+                    bioReadCopyBuf.release();
```
normanmaurer: @Scottmitch can we ensure we also release when the `assert` fails. Just in case :)

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -845,9 +820,18 @@ public final SSLEngineResult unwrap(
...
-
+                return newResultMayFinishHandshake(BUFFER_UNDERFLOW, status, 0, 0);
+            }
+
+            // TODO(scott): limiting to a single packet can result in thrashing in the pipeline to decode application
+            // data. The javadocs elude to attempting to decode one complete SSL/TLS network packet [1], but I don't
+            // think this is meant to enforce a strict maximum of 1. The SslHandler's decode method already loops
+            // through the data to only give "full packets" to the SSLEngine implementations, we duplicate that
+            // logic here, and OpenSSL will also know how to delimit data. Can we remove this restriction which causes
+            // unnecessary overhead?
```
normanmaurer: @Scottmitch I think we can remove this via a constructor option ? Like strict and non-strict mode ?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1387,14 +1349,27 @@ public final synchronized void beginHandshake() throws SSLException {
...
+                int status;
+                if ((status = SSL.renegotiate(ssl)) != 1 || (status = SSL.doHandshake(ssl)) != 1) {
+                    int err = SSL.getError(ssl, status);
+                    switch (err) {
+                        case SSL.SSL_ERROR_WANT_READ:
+                        case SSL.SSL_ERROR_WANT_WRITE:
+                            // If the internal SSL buffer is small it is possible that doHandshake may "fail" because
+                            // there is not enough room to write, so we should try to wrap.
+
+                            // TODO(scott): there may be a OpenSSL bug here. If the internal SSL buffer is < 33 bytes
```
normanmaurer: @Scottmitch can you open an issue and link it here ?

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -790,6 +796,7 @@ private void wrapNonAppData(ChannelHandlerContext ctx, boolean inUnwrap) throws
                 out.release();
             }
         }
+        return false;
```
normanmaurer: @Scottmitch like discussed before I think we should factor this out in a separate PR.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
Scottmitch: I don't think this is necessary. The assert signifies a programatic error and was added to aid development and unit tests. Do you have a suggestion which doesn't introduce runtime checks or additional try/catch?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
Scottmitch: @normanmaurer - great! I noticed this was thrashing a lot during my testing. let me break this out into another PR.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
Scottmitch: https://github.com/openssl/openssl/issues/2551

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
Scottmitch: https://github.com/netty/netty/pull/6311

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
Scottmitch: false alarm ... issue was pre-existing but has been fixed in this PR.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -612,108 +509,162 @@ public final SSLEngineResult wrap(
...
+                        if (handshakeException == null) {
+                            status = handshake();
+
+                            if (renegotiationPending && status == FINISHED) {
+                                // If renegotiationPending is true that means when we attempted to start renegotiation
+                                // the BIO buffer didn't have enough space to hold the HelloRequest which prompts the
+                                // client to initiate a renegotiation. At this point the HelloRequest has been written
+                                // so we can actually start the handshake process.
+                                renegotiationPending = false;
+                                SSL.setState(ssl, SSL.SSL_ST_ACCEPT);
```
Scottmitch: The OpenSSL team has advised against doing this. I have fixed the current approach in this PR and will investigate removal in of `SSL.SSL_ST_ACCEPT` in https://github.com/netty/netty/issues/6320.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: yeah... its ok

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -85,6 +85,17 @@ public Boolean run() {
                     return SystemPropertyUtil.getBoolean("jdk.tls.rejectClientInitiatedRenegotiation", false);
                 }
             });
+
+    private static final int DEFAULT_BIO_NON_APPLICATION_BUFFER_SIZE =
+            AccessController.doPrivileged(new PrivilegedAction<Integer>() {
+                @Override
+                public Integer run() {
+                    return Math.max(1,
```
normanmaurer: why 1 and not 0 (which will use the default then) ?

Scottmitch: I removed support for the default at the tcnative level, and disallowed setting 0. Users should explicitly set the size they want. Default of 16k is likely way to big anyways.

pom.xml
Scottmitch: @normanmaurer will update this before the next release

Scottmitch:
4.1 (d06990f) 4.0 (00cd6d2)

normanmaurer:
LGTM... Thanks!

PR #6216, opened 2017-01-15
Cleanup : Level enum initialization simplified.

doom369:
Motivation:

Make code easier to read, remove unnecessary `trim().toUpperCase()` operations.

Modification:

`Level` enum initialization logic simplified, removed unnecessary `trim().toUpperCase()` invocations and logic moved to 1 method.

Result:

Code is easier to read and simpler.

### DIFF
--- a/common/src/main/java/io/netty/util/ResourceLeakDetector.java
+++ b/common/src/main/java/io/netty/util/ResourceLeakDetector.java
@@ -26,7 +26,6 @@ import java.lang.ref.PhantomReference;
 import java.lang.ref.ReferenceQueue;
 import java.util.ArrayDeque;
 import java.util.Deque;
-import java.util.EnumSet;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 
@@ -66,7 +65,23 @@ public class ResourceLeakDetector<T> {
          * Enables paranoid resource leak detection which reports where the leaked object was accessed recently,
          * at the cost of the highest possible overhead (for testing purposes only).
          */
-        PARANOID
+        PARANOID;
+
+        /**
+         * Returns level based on string value. Accepts also string that represents ordinal number of enum.
+         *
+         * @param levelStr - level string : DISABLED, SIMPLE, ADVANCED, PARANOID. Ignores case.
+         * @return corresponding level or SIMPLE level in case of no match.
+         */
+        static Level parseLevel(String levelStr) {
+            String trimmedLevelStr = levelStr.trim();
+            for (Level l : values()) {
+                if (trimmedLevelStr.equalsIgnoreCase(l.name()) || trimmedLevelStr.equals(String.valueOf(l.ordinal()))) {
+                    return l;
+                }
+            }
+            return DEFAULT_LEVEL;
+        }
     }
 
     private static Level level;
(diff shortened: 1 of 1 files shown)


common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -66,7 +65,22 @@
...
+
+        /**
+         * Returns level based on string value. Accepts also string that represents ordinal number of enum.
+         *
+         * @param levelStr - level string : DISABLED, SIMPLE, ADVANCED, PARANOID. Ignores case.
+         * @return corresponding level or SIMPLE level in case of no match.
+         */
+        static Level getLevel(String levelStr) {
+            for (Level l : values()) {
+                if (levelStr.equalsIgnoreCase(l.name()) || levelStr.equals(String.valueOf(l.ordinal()))) {
```
johnou: missing trim on levelStr

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -88,16 +102,11 @@
...
         // If new property name is present, use it
-        levelStr = SystemPropertyUtil.get(PROP_LEVEL, levelStr).trim().toUpperCase();
-        Level level = DEFAULT_LEVEL;
-        for (Level l: EnumSet.allOf(Level.class)) {
-            if (levelStr.equals(l.name()) || levelStr.equals(String.valueOf(l.ordinal()))) {
-                level = l;
-            }
-        }
+        levelStr = SystemPropertyUtil.get(PROP_LEVEL, levelStr);
+        Level level = Level.getLevel(levelStr.trim());
```
johnou: h0h0 no you didn't here it is, I wonder if it would be cleaner inside the getLevel method though?

doom369: > would be cleaner inside the getLevel method though?

Yes. As with separate method you don't need to read what is going on in loop. You need just to read method name and if you need more details you can go in method.

johnou: what I am proposing is moving the actual trim call into the enum method like so

```java
static Level getLevel(String levelStr) {
   String trimmedLevelStr = levelStr.trim();
```

normanmaurer: @doom369 PTAL ^^

doom369:
@normanmaurer @johnou fixed.

normanmaurer:
@Scottmitch PTAL as well.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -66,7 +65,23 @@
...
-        PARANOID
+        PARANOID;
+
+        /**
+         * Returns level based on string value. Accepts also string that represents ordinal number of enum.
+         *
+         * @param levelStr - level string : DISABLED, SIMPLE, ADVANCED, PARANOID. Ignores case.
+         * @return corresponding level or SIMPLE level in case of no match.
+         */
+        static Level getLevel(String levelStr) {
```
Scottmitch: `getLevel` -> `parseLevel`

doom369:
@Scottmitch done

Scottmitch:
4.1 (0e654f7) 4.0 (b2bf3cf)

Scottmitch:
thanks @doom369 !

PR #6224, opened 2017-01-16
Enable configuring available processors

jasontedor:
Motivation:

In cases when an application is running in a container or is otherwise constrained to the number of processors that it is using, the JVM invocation Runtime#availableProcessors will not return the constrained value but rather the number of processors available to the virtual machine. Netty uses this number in sizing various resources. Additionally, some applications will constrain the number of threads that they are using independenly of the number of processors available on the system. Thus, applications should have a way to globally configure the number of processors.

Modifications:

Rather than invoking Runtime#availableProcessors, Netty should rely on a method that enables configuration when the JVM is started or by the application. This commit exposes a new class NettyRuntime for enabling such configuraiton. This value can only be set once. Its default value is Runtime#availableProcessors so that there is no visible change to existing applications, but enables configuring either a system property or configuring during application startup (e.g., based on settings used to configure the application).

Additionally, we introduce the usage of forbidden-apis to prevent future uses of Runtime#availableProcessors from creeping. Future work should enable the bundled signatures and clean up uses of deprecated and other forbidden methods.

Result:

Netty can be configured to not use the underlying number of processors, but rather the constrained number of processors.

### DIFF
--- a/common/src/main/java/io/netty/util/NettyRuntime.java
+++ b/common/src/main/java/io/netty/util/NettyRuntime.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.netty.util;
+
+import io.netty.util.internal.ObjectUtil;
+import io.netty.util.internal.SystemPropertyUtil;
+
+import java.util.Locale;
+
+/**
+ * A utility class for wrapping calls to {@link Runtime}.
+ */
+public final class NettyRuntime {
+
+    /**
+     * Holder class for available processors to enable testing.
+     */
+    static class AvailableProcessorsHolder {
+
+        private int availableProcessors;
+
+        /**
+         * Set the number of available processors.
+         *
+         * @param availableProcessors the number of available processors
+         * @throws IllegalArgumentException if the specified number of available processors is non-positive
+         * @throws IllegalStateException    if the number of available processors is already configured
+         */
+        synchronized void setAvailableProcessors(final int availableProcessors) {
+            ObjectUtil.checkPositive(availableProcessors, "availableProcessors");
+            if (this.availableProcessors != 0) {
+                final String message = String.format(
+                        Locale.ROOT,
+                        "availableProcessors is already set to [%d], rejecting [%d]",
+                        this.availableProcessors,
+                        availableProcessors);
+                throw new IllegalStateException(message);
+            }
+            this.availableProcessors = availableProcessors;
+        }
+
+        /**
+         * Get the configured number of available processors. The default is {@link Runtime#availableProcessors()}.
+         * This can be overridden by setting the system property "io.netty.availableProcessors" or by invoking
+         * {@link #setAvailableProcessors(int)} before any calls to this method.
+         *
+         * @return the configured number of available processors
+         */
+        @SuppressForbidden(reason = "to obtain default number of available processors")
+        synchronized int availableProcessors() {
+            if (this.availableProcessors == 0) {
+                final int availableProcessors =
+                        SystemPropertyUtil.getInt(
+                                "io.netty.availableProcessors",
+                                Runtime.getRuntime().availableProcessors());
+                setAvailableProcessors(availableProcessors);
+            }
+            return this.availableProcessors;
+        }
+    }
+
+    private static final AvailableProcessorsHolder holder = new AvailableProcessorsHolder();
+
+    /**
+     * Set the number of available processors.
+     *
+     * @param availableProcessors the number of available processors
+     * @throws IllegalArgumentException if the specified number of available processors is non-positive
+     * @throws IllegalStateException    if the number of available processors is already configured
+     */
+    @SuppressWarnings("unused,WeakerAccess") // this method is part of the public API
+    public static void setAvailableProcessors(final int availableProcessors) {
+        holder.setAvailableProcessors(availableProcessors);
+    }
+
+    /**
+     * Get the configured number of available processors. The default is {@link Runtime#availableProcessors()}. This
+     * can be overridden by setting the system property "io.netty.availableProcessors" or by invoking
+     * {@link #setAvailableProcessors(int)} before any calls to this method.
+     *
+     * @return the configured number of available processors
+     */
+    public static int availableProcessors() {
+        return holder.availableProcessors();
+    }
+
+    /**
+     * No public constructor to prevent instances from being created.
+     */
+    private NettyRuntime() {
+    }
+}
--- a/common/src/main/java/io/netty/util/SuppressForbidden.java
+++ b/common/src/main/java/io/netty/util/SuppressForbidden.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.netty.util;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation to suppress forbidden-apis errors inside a whole class, a method, or a field.
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target({ ElementType.CONSTRUCTOR, ElementType.FIELD, ElementType.METHOD, ElementType.TYPE })
+public @interface SuppressForbidden {
+
+    String reason();
+}
--- a/common/src/test/java/io/netty/util/NettyRuntimeTests.java
+++ b/common/src/test/java/io/netty/util/NettyRuntimeTests.java
@@ -1,0 +1,206 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.netty.util;
+
+import io.netty.util.internal.SystemPropertyUtil;
+import org.junit.Test;
+
+import java.util.concurrent.BrokenBarrierException;
+import java.util.concurrent.CyclicBarrier;
+import java.util.concurrent.atomic.AtomicReference;
+
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasToString;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
+
+public class NettyRuntimeTests {
+
+    @Test
+    public void testIllegalSet() {
+        final NettyRuntime.AvailableProcessorsHolder holder = new NettyRuntime.AvailableProcessorsHolder();
+        for (final int i : new int[] { -1, 0 }) {
+            try {
+                holder.setAvailableProcessors(i);
+                fail();
+            } catch (final IllegalArgumentException e) {
+                assertThat(e, hasToString(containsString("(expected: > 0)")));
+            }
+        }
+    }
+
+    @Test
+    public void testMultipleSets() {
+        final NettyRuntime.AvailableProcessorsHolder holder = new NettyRuntime.AvailableProcessorsHolder();
+        holder.setAvailableProcessors(1);
+        try {
+            holder.setAvailableProcessors(2);
+            fail();
+        } catch (final IllegalStateException e) {
+            assertThat(e, hasToString(containsString("availableProcessors is already set to [1], rejecting [2]")));
+        }
+    }
+
+    @Test
+    public void testSetAfterGet() {
+        final NettyRuntime.AvailableProcessorsHolder holder = new NettyRuntime.AvailableProcessorsHolder();
+        holder.availableProcessors();
+        try {
+            holder.setAvailableProcessors(1);
+            fail();
+        } catch (final IllegalStateException e) {
+            assertThat(e, hasToString(containsString("availableProcessors is already set")));
+        }
+    }
+
+    @Test
+    public void testRacingGetAndGet() throws InterruptedException {
+        final NettyRuntime.AvailableProcessorsHolder holder = new NettyRuntime.AvailableProcessorsHolder();
+        final CyclicBarrier barrier = new CyclicBarrier(3);
+
+        final AtomicReference<IllegalStateException> firstReference = new AtomicReference<IllegalStateException>();
+        final Runnable firstTarget = getRunnable(holder, barrier, firstReference);
+        final Thread firstGet = new Thread(firstTarget);
+        firstGet.start();
+
+        final AtomicReference<IllegalStateException> secondRefernce = new AtomicReference<IllegalStateException>();
+        final Runnable secondTarget = getRunnable(holder, barrier, secondRefernce);
+        final Thread secondGet = new Thread(secondTarget);
+        secondGet.start();
+
+        // release the hounds
+        await(barrier);
+
+        // wait for the hounds
+        await(barrier);
+
+        firstGet.join();
+        secondGet.join();
+
+        assertNull(firstReference.get());
+        assertNull(secondRefernce.get());
+    }
+
+    private Runnable getRunnable(
+            final NettyRuntime.AvailableProcessorsHolder holder,
+            final CyclicBarrier barrier,
+            final AtomicReference<IllegalStateException> reference) {
+        return new Runnable() {
+            @Override
+            public void run() {
+                await(barrier);
+                try {
+                    holder.availableProcessors();
+                } catch (final IllegalStateException e) {
+                    reference.set(e);
+                }
+                await(barrier);
+            }
+        };
+    }
+
+    @Test
+    public void testRacingGetAndSet() throws InterruptedException {
+        final NettyRuntime.AvailableProcessorsHolder holder = new NettyRuntime.AvailableProcessorsHolder();
+        final CyclicBarrier barrier = new CyclicBarrier(3);
+        final Thread get = new Thread(new Runnable() {
+            @Override
+            public void run() {
+                await(barrier);
+                holder.availableProcessors();
+                await(barrier);
+            }
+        });
+        get.start();
+
+        final AtomicReference<IllegalStateException> setException = new AtomicReference<IllegalStateException>();
+        final Thread set = new Thread(new Runnable() {
+            @Override
+            public void run() {
+                await(barrier);
+                try {
+                    holder.setAvailableProcessors(2048);
+                } catch (final IllegalStateException e) {
+                    setException.set(e);
+                }
+                await(barrier);
+            }
+        });
+        set.start();
+
+        // release the hounds
+        await(barrier);
+
+        // wait for the hounds
+        await(barrier);
+
+        get.join();
+        set.join();
+
+        if (setException.get() == null) {
+            assertThat(holder.availableProcessors(), equalTo(2048));
+        } else {
+            assertNotNull(setException.get());
+        }
+    }
+
+    @Test
+    public void testGetWithSystemProperty() {
+        final String availableProcessorsSystemProperty = SystemPropertyUtil.get("io.netty.availableProcessors");
+        try {
+            System.setProperty("io.netty.availableProcessors", "2048");
+            final NettyRuntime.AvailableProcessorsHolder holder = new NettyRuntime.AvailableProcessorsHolder();
+            assertThat(holder.availableProcessors(), equalTo(2048));
+        } finally {
+            if (availableProcessorsSystemProperty != null) {
+                System.setProperty("io.netty.availableProcessors", availableProcessorsSystemProperty);
+            } else {
+                System.clearProperty("io.netty.availableProcessors");
+            }
+        }
+    }
+
+    @Test
+    @SuppressForbidden(reason = "testing fallback to Runtime#availableProcessors")
+    public void testGet() {
+        final String availableProcessorsSystemProperty = SystemPropertyUtil.get("io.netty.availableProcessors");
+        try {
+            System.clearProperty("io.netty.availableProcessors");
+            final NettyRuntime.AvailableProcessorsHolder holder = new NettyRuntime.AvailableProcessorsHolder();
+            assertThat(holder.availableProcessors(), equalTo(Runtime.getRuntime().availableProcessors()));
+        } finally {
+            if (availableProcessorsSystemProperty != null) {
+                System.setProperty("io.netty.availableProcessors", availableProcessorsSystemProperty);
+            } else {
+                System.clearProperty("io.netty.availableProcessors");
+            }
+        }
+    }
+
+    private static void await(final CyclicBarrier barrier) {
+        try {
+            barrier.await();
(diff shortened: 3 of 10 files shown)


jasontedor:
The [build failure in the PR CI build](https://garage.netty.io/teamcity/viewLog.html?buildId=15394&buildTypeId=pulls_netty&guest=1) has nothing to do with this PR.

common/src/main/java/io/netty/util/NettyRuntime.java
```diff
@@ -0,0 +1,74 @@
...
+     * @return the configured number of available processors
+     */
+    @SuppressForbidden(reason = "to obtain default number of available processors")
+    public static int availableProcessors() {
+        if (NettyRuntime.availableProcessors.get() == 0) {
+            final int availableProcessors =
+                    SystemPropertyUtil.getInt(
+                            "io.netty.availableProcessors",
+                            Runtime.getRuntime().availableProcessors());
+            setAvailableProcessors(availableProcessors);
```
normanmaurer: @jasontedor couldn't this race and so have `setAvailableProcessors(...)` called twice and produce an `llegalStateException` ? 

I think what you need to do here is just use the `AtomicInteger` directly like:

```java
int availableProcessors = NettyRuntime.availableProcessors.get();
if (availableProcessors == 0) {
    availableProcessors = SystemPropertyUtil.getInt("io.netty.availableProcessors", Runtime.getRuntime().availableProcessors());
    if (!NettyRuntime.availableProcessors.compareAndSet(0, availableProcessors)) {
        // refetch as it was set in the meantime by calling availableProcessors() concurrently.
        availableProcessors = NettyRuntime.availableProcessors.get();
    }
}
return availableProcessors;
```

common/src/main/java/io/netty/util/NettyRuntime.java
```diff
@@ -0,0 +1,74 @@
...
+import io.netty.util.internal.SystemPropertyUtil;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * A utility class for wrapping calls to {@link Runtime}.
+ */
+public final class NettyRuntime {
+
+    private static final AtomicInteger availableProcessors = new AtomicInteger();
```
fenik17: For why lazy initialization instead simple static?

Scottmitch: I guess it is to allow `setAvailableProcessors` to override the `availableProcessors` exactly once (assuming `availableProcessors` has not yet been called bcz `0` is treated as a "special" value)?

Scottmitch:
@jasontedor - Agreed build failure seems unrelated. Seems like timeouts maybe too aggressive for build servers. I will submit a PR to relax these timeouts.

common/src/main/java/io/netty/util/NettyRuntime.java
jasontedor: @fenik17 The explanation given by @Scottmitch correctly explains my intentions here.

fenik17: @Scottmitch @jasontedor
Yes, I'm understand this. But I can't imagine of a scenario where may need change value in runtime..

common/src/main/java/io/netty/util/NettyRuntime.java
normanmaurer: @jasontedor agree ?

common/src/main/java/io/netty/util/NettyRuntime.java
jasontedor: We have a need for setting it at runtime at startup based on configuration.

common/src/main/java/io/netty/util/NettyRuntime.java
jasontedor: Yes, definitely, although it was deliberate to handle this way: to scream when a get races with a set but it also means screaming if a get races with a get. I'm not sure how problematic that is given current uses.

normanmaurer: @jasontedor but could such a race just happen during init with the code you provided ? Like when multiple classes call `availableProcessors()` during init. I think in this case we not want to fail or I am missing something ?

common/src/main/java/io/netty/util/NettyRuntime.java
fenik17: I'm sorry for worry.. But looks like something I don't understand.. If you need setting it at runtime at startup based on configuration, why not do simple:

```java
public final class NettyRuntime {
    public static final int AVAILABLE_PROCESSORS = SystemPropertyUtil.getInt(
        "io.netty.availableProcessors", Runtime.getRuntime().availableProcessors());
}
```

common/src/main/java/io/netty/util/NettyRuntime.java
jasontedor: Yes, that's what I mean by two gets racing. You're not missing anything, but I'm not sure if it's actually a problem.

normanmaurer: @jasontedor I think it is ... how could this be not a problem ? Isnt it expected that multiple "static" blocks could call the method concurrently for example ?

common/src/main/java/io/netty/util/NettyRuntime.java
jasontedor: That works in some situations (that's why I provided it) but it will not work for ours because we size our thread pools based on configuration provided by the user (in our application configuration file) and we can not expect them to set this system property too.

common/src/main/java/io/netty/util/NettyRuntime.java
jasontedor: Why would you expect that (to be clear, you have to mean multiple different static blocks)?

normanmaurer: @jasontedor yes different static blocks or it could even be just different methods that call it that run concurrently.  Why would you not expect something like this happen ? Shouldnt it be safe to call a method concurrently that does lazy init ?

common/src/main/java/io/netty/util/NettyRuntime.java
fenik17: If I understand correctly you want invoke NettyRuntime.setAvailableProcessors() from your startup configuration code? But if static initialization block in some other class calling NettyRuntime.availableProcessors() before your configuration class, will thrown IllegalStateException..
May be will be prefer creating mechanism for binding your properties to netty properties (at the same time for the any other settings)?

jasontedor: > But if static initialization block in some other class calling NettyRuntime.availableProcessors() before your configuration class, will thrown IllegalStateException.

I will ensure that this never happens. The configuration will occur immediately prior to entry point to Netty, before any Netty classes have been loaded.

I have already tested against a build of Netty containing this change.

fenik17: What do you use to ensure correct class loading order? 
Anyway other netty users may failing on this and will be forced delve into the details of the internal implementation. IMHO it's not clear and SOLID-like way...

jasontedor: > What do you use to ensure correct class loading order?

It's simple: do not touch any Netty classes before invoking `NettyRuntime#setAvailableProcessors`. We do the same thing [today](https://github.com/elastic/elasticsearch/blob/2766b08ff45a039d9d7b15435d4680f915f4b1e1/modules/transport-netty4/src/main/java/org/elasticsearch/transport/netty4/Netty4Utils.java#L43-L52) with setting the logger factory. This will not be able to be in a static block anymore since it will require input from the user configuration, but I've already tested an approach that works.

> Anyway other netty users may failing on this and will be forced delve into the details of the internal implementation.

Most users will not care, and for those that do, I think that the Javadocs included with this PR are clear on the expectations.

> IMHO it's not clear and SOLID-like way...

Sorry, I don't understand this at all.

common/src/main/java/io/netty/util/NettyRuntime.java
jasontedor: > yes different static blocks or it could even be just different methods that call it that run concurrently. Why would you not expect something like this happen ? 

No, I don't expect that. That's why I am acknowledging that potential for a race is there but I am saying that I think it's fine in practice.

For example, look at `NioEventLoopGroup` which extends `MultithreadEventLoopGroup`. There is a class initializer on `MultithreadEventLoopGroup` with this block of code:

```
  static {
        DEFAULT_EVENT_LOOP_THREADS = Math.max(1, SystemPropertyUtil.getInt(
                "io.netty.eventLoopThreads", NettyRuntime.availableProcessors() * 2));

            logger.debug("-Dio.netty.eventLoopThreads: {}", DEFAULT_EVENT_LOOP_THREADS);
        }
    }
```

This is guaranteed to only execute once. It can not be the case that constructing two event loop groups lead to a get race here, the class initializer guards against that.

Since this is used statically everywhere, the only way to get the race that you're concerned about it by being nefarious. 

> Shouldnt it be safe to call a method concurrently that does lazy init ?

The downside to your way is that now you allow a get racing with a set, all silently being fine for a thousand runs until one day an evil scheduler executes the get before the set.

This is why I favor the approach that I took here. I do not think a race arises in practice. My way disallows all races: get with get, and get with set so there will always be a loser that throws an `IllegalStateException` so an application programmer fixes their buggy code and avoids the race to begin with. With your way, you introduce the possibility of a get racing with a set, but only sometimes is there a loser that throws an `IllegalStateException` (if the set wins, everything is fine, if the get wins, the set loses); that's why this I'm opposed to this option.

common/src/main/java/io/netty/util/NettyRuntime.java
fenik17: @jasontedor Thanks for your answers! And sorry for my bad english..

jasontedor: There is nothing wrong with your English and I appreciate all your questions! It's only your last point I didn't understand, not because of a language barrier, but only because I'm not sure what you meant. If you want to elaborate, I'm happy to try to understand.

common/src/main/java/io/netty/util/NettyRuntime.java
jasontedor: Do you have any additional thoughts @normanmaurer?

normanmaurer: Yes... I still disagree and I think it should be valid to call the getter with lazy init from multiple threads concurrently without this exception.

normanmaurer: @jasontedor btw beside this I am happy to pull in ;)

common/src/main/java/io/netty/util/NettyRuntime.java
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Get the configured number of available processors. The default is {@link Runtime#availableProcessors()}. This
+     * can be overridden by setting the system property "io.netty.availableProcessors" or by invoking
+     * {@link #setAvailableProcessors(int)} before any calls to {@link #availableProcessors()}.
+     *
+     * @return the configured number of available processors
+     */
+    @SuppressForbidden(reason = "to obtain default number of available processors")
+    public static int availableProcessors() {
+        if (NettyRuntime.availableProcessors.get() == 0) {
```
Scottmitch: can we cache this value?

```java
int ap = NettyRuntime.availableProcessors.get();
if (ap == 0) {
  ap = SystemPropertyUtil(...);
...
}
return ap;
```

common/src/main/java/io/netty/util/NettyRuntime.java
```diff
@@ -0,0 +1,74 @@
...
+            final int availableProcessors =
+                    SystemPropertyUtil.getInt(
+                            "io.netty.availableProcessors",
+                            Runtime.getRuntime().availableProcessors());
+            setAvailableProcessors(availableProcessors);
+        }
+        return NettyRuntime.availableProcessors.get();
+    }
+
```
Scottmitch: nit: kill line

common/src/main/java/io/netty/util/NettyRuntime.java
fenik17: Ok. Will try expand opinion.
Your solution looks like hack for solving a particular problem. It violates the [dependency inversion principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle) ('D' from [SOLID](https://en.wikipedia.org/wiki/SOLID_(object-oriented_design))) because enforce deepen in the implementation details. If someone wants to do a similar thing (configuring property 'availableProcessors' in runtime) and call setAvailableProcessors(), it can step on the rake and will be forced to classloading problem. This possibility forced you to create constraints (forbiddenapis).

Also if you need to do such a thing for the other Netty setting, you can't use reuse current code, and will be forced to create same hack.

Instead it, would be possible to modify SystemPropertyUtil for change/override/extend the property source (for example, for getting some properties from xml config or database).

common/src/main/java/io/netty/util/NettyRuntime.java
jasontedor: I think that you're optimizing for the wrong thing. You're optimizing for a scenario that shouldn't happen, it should be considered a bug if it happens. By optimizing for that scenario, you're permitting bad racing scenarios to silently succeed (namely, a get racing with a set). My implementation prevents all races, it treats them as application bugs. I maintain that my implementation should be be preferred.

common/src/main/java/io/netty/util/NettyRuntime.java
jasontedor: > This possibility forced you to create constraints (forbiddenapis).

No, the reason for using forbidden-apis is to prevent a future change to Netty from calling `Runtime#availableProcessors` when they should now be calling `NettyRuntime#availableProcessors`. Using forbidden-apis prevents mistakes from creeping into the codebase by failing the build with very clear instructions: use `NettyRuntime#availableProcessors`.

fenik17: Yes.. Thanks for clarifying forbidden-api destination.

common/src/main/java/io/netty/util/NettyRuntime.java
jasontedor: @normanmaurer Can we try to come to resolution on this before the 4.1.8.Final release?

normanmaurer: @jasontedor I still don't understand why you think this can never happen. How can you guarantee this ? Couldn't multiple classes be instanced or loaded concurrently and so call this concurrently ? Sorry I missed your comment :(

Scottmitch: @jasontedor - have you considered use cases such as the following:

```java
public class Whatever {
    private final int procs;
    public Whatever() {
        procs = NettyRuntime.availableProcessors();
    }
}

public class Whatever2 {
    private final int procs;
    public Whatever2() {
        procs = NettyRuntime.availableProcessors();
    }
}

    executor.execute(() -> {
        Whatever w = new Whatever();
    });
    
    executor.execute(() -> {
        Whatever2 w = new Whatever2();
    });
// block some how to prevent exiting
}
```

It is contrived but just to demonstrate that calling the accessor may happen from two different threads and it seems unexpected for this to be considered and error if the SystemProperty initialization path races between the two threads and ultimately end up setting the same value. 

IIUC you want to differentiate the case where a `setAvailableProcessors` races with the `availableProcessors` default initialization? Maybe `setAvailableProcessors` can use something similar to `AtomicStampedReference` to differentiate from the case of `setAvailableProcessors` raced with `availableProcessors` and coincidentally set the same value as the system default ... and `availableProcessors` would only throw if it tried to set the system default but a different value existed?

jasontedor: > I still don't understand why you think this can never happen. How can you guarantee this ?

@normanmaurer I never said it can't happen (notice I said "shouldn't", not "can't"). I've acknowledged from the beginning it can happen and I coded this with the possibility in mind. To be clear, a get race is possible, we're just arguing about how to handle it. You want to allow it, I want to forbid it (in the sense that if an application developer causes a get race to exist, it should be treated as an application bug and dealt with there). I want it this way because it's simple. I don't want it your way because your way permits get and set races in the sense that if an application allows a race to exist between a get and set, their application can appear to run fine if the set executes before the get until one day an evil scheduler runs the get before the set and then the application explodes. My way has a simple model: do not permit a race in your application code. My way is not restrictive because there's no practical situation in which obtaining this value should be racy. Look at the uses within Netty. Again, look at the example of `NioEventLoopGroup` I gave above. Also, a get race is not as innocent as you make it sound because the system property can be set in code; so a get race like you want to allow could have one thread win before the system property is set on another thread that tries to get the value to set it lazily yet sometimes that thread wins and the application ends up with different values depending on the winner. Let's have none of this, with a very clear and easy to reason about design. Let's not design this for a situation that is easily avoided and has no practical advantage in allowing.

jasontedor: > have you considered use cases such as the following

@Scottmitch That's not a use case, as you admit, it's a contrived example. 😇

> It is contrived but just to demonstrate that calling the accessor may happen from two different threads

I've acknowledged since the beginning that a get race is possible, we're just arguing about how to handle it.

> and it seems unexpected for this to be considered and error if the SystemProperty initialization path races between the two threads and ultimately end up setting the same value.

As I mentioned in my reply to @normanmaurer, this might not be the case if one racing thread is getting before another racing thread programmatically sets the system property.

> Maybe `setAvailableProcessors` can use something similar to `AtomicStampedReference` to differentiate from the case of `setAvailableProcessors` raced with `availableProcessors` and coincidentally set the same value as the system default ... and `availableProcessors` would only throw if it tried to set the system default but a different value existed?

We do not need to come up with complicated solutions for a problem that should not (nb. not "can not") happen short of contrived examples.

Scottmitch: > I've acknowledged since the beginning that a get race is possible, we're just arguing about how to handle it.

The following seems like reasonable behavior to me:
- OK - Two threads race to initialize from `System.getProperty` and result in setting `availableProcessors` to the same value
- ERROR - Two threads race between initialize from `System.getProperty` and `setAvailableProcessors`
- ERROR - Two threads race between `setAvailableProcessors`
- ERROR - No race, but attempting to call `setAvailableProcessors` after `availableProcessors` variable has been set

Based upon the discussion it seems like you disagree and think that the first case should also be an ERROR with the rational that "this shouldn't happen". Is this correct?

> this might not be the case if one racing thread is getting before another racing thread programmatically sets the system property.

My suggestion acknowledged this and proposes and alternative approach to differentiate between the two scenarios. The point is this condition seems plausible. I have not seen a convincing argument that we should ignore this case.

jasontedor: @Scottmitch I think I've come up with a simple solution that should satisfy both of us. Instead, we can just synchronize these methods. Now, two threads can get and it won't be racy so no error, I have the protection that I want, and we didn't have to do anything complicated. Would you please take another look, I'd really like to get this in?

jasontedor:
The test failure here has nothing to do with my change.

common/src/main/java/io/netty/util/NettyRuntime.java
```diff
@@ -0,0 +1,114 @@
...
+        /**
+         * Set the number of available processors.
+         *
+         * @param availableProcessors the number of available processors
+         * @throws IllegalArgumentException if the specified number of available processors is non-positive
+         * @throws IllegalStateException    if the number of available processors is already configured
+         */
+        void setAvailableProcessors(final int availableProcessors) {
+            synchronized (lock) {
```
normanmaurer: @jasontedor why do you need to even have this lock and not just mark the whole methods `synchronized` ?

normanmaurer:
@jasontedor let me check for the ci error

normanmaurer:
@jasontedor I rebased your patch locally here and resolved the conflicts and it compiles without a problem. Please rebase as well

common/src/main/java/io/netty/util/NettyRuntime.java
jasontedor: Oops, artifact of an earlier iteration that I never refactored into the simpler form after what I had done was no longer needed. I pushed feea27a5735c12d1aa99ec96a8c9bc1f9ddfa86c.

jasontedor:
@normanmaurer I've rebased and squashed.

dev-tools/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+  <parent>
+    <groupId>org.sonatype.oss</groupId>
+    <artifactId>oss-parent</artifactId>
+    <version>7</version>
+    <relativePath></relativePath>
+  </parent>
+
+  <groupId>io.netty</groupId>
+  <artifactId>netty-dev-tools</artifactId>
+  <version>4.1.8.Final-SNAPSHOT</version>
```
fenik17: 4.1.10 at now.

jasontedor: That's annoying, that change got lost when I squashed.

normanmaurer:
@jasontedor the test-failure we saw here is fixed by https://github.com/netty/netty/pull/6651

jasontedor:
Thanks @normanmaurer.

normanmaurer:
@trustin fine with you ?

common/src/test/java/io/netty/util/NettyRuntimeTests.java
```diff
@@ -0,0 +1,200 @@
...
+        }
+    }
+
+    @Test
+    public void testGetWithSystemProperty() {
+        final String availableProcessorsSystemProperty = SystemPropertyUtil.get("io.netty.availableProcessors");
+        System.setProperty("io.netty.availableProcessors", "2048");
+        final NettyRuntime.AvailableProcessorsHolder holder = new NettyRuntime.AvailableProcessorsHolder();
+        assertThat(holder.availableProcessors(), equalTo(2048));
+        if (availableProcessorsSystemProperty != null) {
```
Scottmitch: consider doing the restore of state in a final block as this may impact other tests if a failure occurs.

jasontedor: I pushed a39b1df5678641c71512d45b88e509ea04c7288a.

common/src/test/java/io/netty/util/NettyRuntimeTests.java
```diff
@@ -0,0 +1,200 @@
...
+        final NettyRuntime.AvailableProcessorsHolder holder = new NettyRuntime.AvailableProcessorsHolder();
+        assertThat(holder.availableProcessors(), equalTo(Runtime.getRuntime().availableProcessors()));
+        if (availableProcessorsSystemProperty != null) {
+            System.setProperty("io.netty.availableProcessors", availableProcessorsSystemProperty);
+        } else {
+            System.clearProperty("io.netty.availableProcessors");
+        }
+    }
+
+    private void await(final CyclicBarrier barrier) {
```
jasontedor: I pushed 9ba182e6351babb9e9be1dbb1df6a8ac234eecf2.

common/src/main/java/io/netty/util/SuppressForbidden.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Annotation to suppress forbidden-apis errors inside a whole class, a method, or a field.
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target({ ElementType.CONSTRUCTOR, ElementType.FIELD, ElementType.METHOD, ElementType.TYPE })
+public @interface SuppressForbidden {
+
+    String reason();
+
```
jasontedor: I pushed 9c80d37e9f185ea113f5d1684991377b84a7516c.

common/src/main/java/io/netty/util/NettyRuntime.java
```diff
@@ -0,0 +1,108 @@
...
+
+        /**
+         * Set the number of available processors.
+         *
+         * @param availableProcessors the number of available processors
+         * @throws IllegalArgumentException if the specified number of available processors is non-positive
+         * @throws IllegalStateException    if the number of available processors is already configured
+         */
+        synchronized void setAvailableProcessors(final int availableProcessors) {
+            if (availableProcessors <= 0) {
```
jasontedor: I pushed ceaed935f69754ee5f78fb02aaf88a973a6b8d50.

common/src/main/java/io/netty/util/NettyRuntime.java
```diff
@@ -0,0 +1,108 @@
...
+        /**
+         * Get the configured number of available processors. The default is {@link Runtime#availableProcessors()}.
+         * This can be overridden by setting the system property "io.netty.availableProcessors" or by invoking
+         * {@link #setAvailableProcessors(int)} before any calls to this method.
+         *
+         * @return the configured number of available processors
+         */
+        @SuppressForbidden(reason = "to obtain default number of available processors")
+        synchronized int availableProcessors() {
+            if (this.availableProcessors == 0) {
```
jasontedor: I think the `this` should stay because in the inner scope I named a local variable `availableProcessors`.

pom.xml
```diff
@@ -1300,6 +1320,83 @@
...
+              <id>check-forbidden-apis</id>
+              <configuration>
+                <targetVersion>${maven.compiler.target}</targetVersion>
+                <!-- allow undocumented classes like sun.misc.Unsafe: -->
+                <internalRuntimeForbidden>false</internalRuntimeForbidden>
+                <!-- if the used Java version is too new, don't fail, just do nothing: -->
+                <failOnUnsupportedJava>false</failOnUnsupportedJava>
+                <bundledSignatures>
+                  <!-- This will automatically choose the right signatures based on 'targetVersion': -->
+                  <!-- enabling these should be done in the future -->
```
jasontedor: I opened #6653.

jasontedor:
Thanks @Scottmitch. I've responded to your feedback. Let me know when I should squash again.

jasontedor:
Thanks @Scottmitch, I've squashed.

normanmaurer:
Cherry-picked into 4.1 (98beb777f81f092aa0fddf49ed08b426b2c72f01) and 4.0 (f9309bdfa9ddf880b2bff38b4ad9f1d5a8ad7146).

@jasontedor thanks a lot and sorry it took so long ...

jasontedor:
Thank you @normanmaurer @Scottmitch, I'm glad we arrived at a solution that satisfies all of us. Thank for the discussion and helpful reviews.

> sorry it took so long ...

No apology necessary here, it's on me that it stalled as long as it did.

Issue #6210, opened 2017-01-13
Unexpected behavior for MAX_HEADER_TABLE_SIZE


exell-christopher:
### Issue Description
I ran into some unexpected behavior when playing with MAX_HEADER_TABLE_SIZE, and the HTTP/2 Encoder/Decoder, and I figure it's worth noting.

The constant MAX_HEADER_TABLE_SIZE indicates that the largest value for the table size is the largest unsigned 32 bit int. Because of this the values that could be passed in are larger than Java int. However, when the encoder encodes the value it first casts the value to be sent to a Java int, resulting in overflow, then encodes it. This results in the receiver receiving a different value than what the sender expected to send.

On the decoder side, if you use something slightly less than Integer.MAX_VALUE, decodeULE128 fails to decode the value stating that it is too large.

While this isn't what I would consider "normal" behavior or sizes for the header table, I would expect that the MAX_HEADER_TABLE_SIZE constant should be less than the max value supported by decodeULE128 in the decoder.

### Netty version
4.1.6/4.1.7

Scottmitch:
> This results in the receiver receiving a different value than what the sender expected to send.

Are you sure about this? I think the issue is more likely that the `decodeULE128` returns an `int`(inherited from Twitter's hpack [1]) and the integer encoding cannot completely be representing in an 31 bit unsigned value (max value documented as [2]).

I agree this is an issue though. There are some cases that can tolerate a 32 bit unsigned value and in general reporting a decompression error due to the limitations of java's signed types is misleading.

[1] https://github.com/netty/netty/commit/8accc52b0313ac4148a591881ff66a3c60ef4707#diff-265b66ff1f2ddb67210a8584c03c7e4aR543
[2] https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/internal/hpack/Decoder.java#L434

Scottmitch:
`MAX_HEADER_TABLE_SIZE` is meant to limit [SETTINGS_HEADER_TABLE_SIZE](https://tools.ietf.org/html/rfc7540#section-6.5.2) as defined by the HTTP/2 RFC and that is where the limit comes from. Just because java's int type can't represent the decoded representation doesn't mean we shouldn't support encoding it.

PR #6229, opened 2017-01-17
HTTP/2 HPACK Integer Encoding Bugs

Scottmitch:
Motivation:
- Decoder#decodeULE128 has a bounds bug and cannot decode Integer.MAX_VALUE
- Decoder#decodeULE128 doesn't support values greater than can be represented with Java's int data type. This is a problem because there are cases that require at least unsigned 32 bits (max header table size).
- Decoder#decodeULE128 treats overflowing the data type and invalid input the same. This can be misleading when inspecting the error that is thrown.
- Encoder#encodeInteger doesn't support values greater than can be represented with Java's int data type. This is a problem because there are cases that require at least unsigned 32 bits (max header table size).

Modifications:
- Correct the above issues and add unit tests.

Result:
Fixes https://github.com/netty/netty/issues/6210.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/internal/hpack/Decoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/internal/hpack/Decoder.java
@@ -51,10 +51,12 @@ import static io.netty.util.AsciiString.EMPTY_STRING;
 import static io.netty.util.internal.ThrowableUtil.unknownStackTrace;
 
 public final class Decoder {
-    private static final Http2Exception DECODE_DECOMPRESSION_EXCEPTION = unknownStackTrace(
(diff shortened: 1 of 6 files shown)


Scottmitch:
Just incase anyone is concerned about performance I ran some benchmarks for using `long` instead of `int`.

```
Benchmark                                     Mode  Cnt   Score   Error  Units
DecoderULE128Benchmark.decodeMaxInt           avgt   10  16.204 ± 2.156  ns/op
DecoderULE128Benchmark.decodeMaxIntUsingLong  avgt   10  14.128 ± 1.108  ns/op
DecoderULE128Benchmark.decodeMaxIntWithLong   avgt   10  13.694 ± 0.538  ns/op
DecoderULE128Benchmark.decodeMaxLong          avgt   10  23.577 ± 1.581  ns/op
```

Scottmitch:
@exell-christopher - FYI

Scottmitch:
4.1 (b701da8)

Issue #6209, opened 2017-01-12
HTTP/2 HPACK decoder oversized header handling causes client/server Dynamic Table inconsistencies


exell-christopher:
### Problem Description
When a client sends oversized headers that causes the decoder to exceed the MAX_HEADER_LIST_SIZE setting, the decoder stops processing headers, and throws a HeaderListSizeException which causes a 431, and a RST_STREAM. If the incoming request had additional headers after the header that caused the HeaderListSizeException to be thrown that would have been incrementally indexed, they are ignored. This results in subsequent requests from the same client using what the client thinks are indexed headers to be rejected with GO_AWAY COMPRESSION_ERROR as the server side decoder dynamic table does not contain the indexed headers.

### Expected behavior
When oversized headers are received, netty will continue processing headers, and updating the dynamic table until all headers are received. Once header processing is complete the HeaderListSizeException should be thrown. To prevent malicious users from sending an infinite number of header frames, and tying up resources, there should be a capped MAX_HEADER_LIST_SIZE overflow that netty will accept then it should issue a GO_AWAY to the client and close the connection.

### Netty version
4.1.6 and later

### Repro Case
See attached maven project zip

### Other things worth noting
While this case occurs when oversized headers are received, any exception that occurs during header processing that is not a connection error (GO_AWAY) can potentially cause dynamic table corruption, and should result in a GO_AWAY

I'm planning on working on a pull request to fix this, but figured this is worth some discussion, so I'm opening the issue first.
[netty-decoder-debug.zip](https://github.com/netty/netty/files/702509/netty-decoder-debug.zip)

normanmaurer:
@Scottmitch @nmittler can you take a look ?

nhnFreespirit:
I've been helping @exell-christopher with debugging this.

Our current understanding is that there really is no safe way to recover from an exception thrown in the Decoder except for tearing down the connection as the index tables in client and server can have gotten out of synch with no way to safely bring them back in sync.

Scottmitch:
This is even called out in the RFC. Makes sense since HPACK is stateful.

https://tools.ietf.org/html/rfc7540#section-10.5.1
> The header block MUST be processed to ensure a consistent
   connection state, unless the connection is closed.

[...]

Issue #5952, opened 2016-10-29
Http2Codec fails silently when the server tries to send too big headers


mosesn:
When the server tries to send a response with too big headers according to the negotiated SETTINGS_MAX_HEADER_LIST_SIZE, it only half-closes and fails silently.  I think it should instead fireExceptionCaught.

mosesn:
This is easy to reproduce.  Spin up an http/2 server, and have it respond to a request with a header of size 10K.

mosesn:
I'm having an awful time testing this, but I think the solution is simple:

```
diff --git a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
index 8b3ec62..24953c7 100644
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
@@ -388,6 +388,7 @@ public final class Http2MultiplexCodec extends ChannelDuplexHandler {
                 promise.addListener(this);
             } else {
                 frame.streamId(streamId());
+                promise.addListener(ChannelFutureListener.FIRE_EXCEPTION_ON_FAILURE);
             }
             writeFromStreamChannel(frame, promise, false);
         }
```

mosesn:
@normanmaurer do you have tips for how I should test this?

normanmaurer:
@mosesn writes usually not results in `exceptionCaught(...)` calls but should just fail the promise. Let me think a bit about this.

mosesn:
@normanmaurer the problem I'm running into is that the promise isn't accessible.

normanmaurer:
@mosesn yep which I think is the "core problem" and I think we need to make this happen ;)

mosesn:
@normanmaurer should I post the diff I wrote in a comment before as a PR, or do you have another idea?

normanmaurer:
@mosesn like I said I need to think a bit more about it. I think we need to refactor the whole child channel thing a bit to make it "better" and so allow to be notified

mosesn:
oof, OK.  I'll just document the bug in our codebase for now.

normanmaurer:
@buchgr @ejona86 btw if you have any "good" idea let me know

buchgr:
[...]

PR #6230, opened 2017-01-17
HTTP/2 Max Header List Size Bug

Scottmitch:
Motivation:
If the HPACK Decoder detects that SETTINGS_MAX_HEADER_LIST_SIZE has been violated it aborts immediately and sends a RST_STREAM frame for what ever stream caused the issue. Because HPACK is stateful this means that the HPACK state may become out of sync between peers, and the issue won't be detected until the next headers frame. We should make a best effort to keep processing to keep the HPACK state in sync with our peer, or completely close the connection.
If the HPACK Encoder is configured to verify SETTINGS_MAX_HEADER_LIST_SIZE it checks the limit and encodes at the same time. This may result in modifying the HPACK local state but not sending the headers to the peer if SETTINGS_MAX_HEADER_LIST_SIZE is violated. This will also lead to an inconsistency in HPACK state that will be flagged at some later time.

Modifications:
- HPACK Decoder now has 2 levels of limits related to SETTINGS_MAX_HEADER_LIST_SIZE. The first will attempt to keep processing data and send a RST_STREAM after all data is processed. The second will send a GO_AWAY and close the entire connection.
- When the HPACK Encoder enforces SETTINGS_MAX_HEADER_LIST_SIZE it should not modify the HPACK state until the size has been checked.
- https://tools.ietf.org/html/rfc7540#section-6.5.2 states that the initial value of SETTINGS_MAX_HEADER_LIST_SIZE is "unlimited". We currently use 8k as a limit. We should honor the specifications default value so we don't unintentionally close a connection before the remote peer is aware of the local settings.
- Remove unnecessary object allocation in DefaultHttp2HeadersDecoder and DefaultHttp2HeadersEncoder.

Result:
Fixes https://github.com/netty/netty/issues/6209.


codec-http2/src/main/java/io/netty/handler/codec/http2/internal/hpack/Encoder.java
```diff
@@ -116,27 +117,26 @@ private void encodeHeadersEnforceMaxHeaderListSize(int streamId, ByteBuf out, Ht
...
             throws Http2Exception {
         long headerSize = 0;
+        // To ensure we stay consistent with our peer check the size is valid before we potentially modify HPACK state.
         for (Map.Entry<CharSequence, CharSequence> header : headers) {
             CharSequence name = header.getKey();
             CharSequence value = header.getValue();
-            long currHeaderSize = HeaderField.sizeOf(name, value);
             // OK to increment now and check for bounds after because this value is limited to unsigned int and will not
             // overflow.
-            headerSize += currHeaderSize;
```
Scottmitch: @nmittler - An alternative approach would be to stick with a single iteration, but close the connection if the limit is exceeded. The current approach was chosen to preserve stream independence at the cost of an extra iteration.

Scottmitch:
@exell-christopher - FYI

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameWriter.java
```diff
@@ -56,9 +56,10 @@
      * @param endStream indicates if this is the last frame to be sent for the stream.
      * @param promise the promise for the write.
      * @return the future for the write.
+     * @throws Http2Exception if an exception occurs while encoding headers.
      */
     ChannelFuture writeHeaders(ChannelHandlerContext ctx, int streamId, Http2Headers headers,
-            int padding, boolean endStream, ChannelPromise promise);
+            int padding, boolean endStream, ChannelPromise promise) throws Http2Exception;
```
Scottmitch: @nmittler - I choose to modify the method signature so implementations are free to throw stream errors or connection errors. This may have impact on user applications, and makes the potential for an exception to be generated explicit. We could apply this rational consistently to other methods in this interface, but currently no other methods have a need to throw.

exell-christopher: My primary concern here is that the docs give users overriding this functionality some breadcrumbs to make sure they don't break their HTTP/2 stack.  Doing the wrong thing when overriding, as we have discovered, can lead to header table corruption, and potentially cross stream contamination.

Scottmitch: sgtm ... I will add some additional text. Let me know what you think once pushed.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecUtil.java
```diff
@@ -229,6 +233,11 @@ public static void headerListSizeExceeded(int streamId, long maxHeaderListSize,
                                   "allowed size (%d)", maxHeaderListSize);
     }
```
Scottmitch: sgtm ... I will add docs

Scottmitch:
> This appear to open up a potential threat as a client is allowed to immediately send frames as soon as they connect,

@exell-christopher - Thanks for bringing this up. I decided to follow the RFC literally during development, and this is not what we want to push out there :).  The RFC provides a loophole which we were previously exploiting:

https://tools.ietf.org/html/rfc7540#section-6.5.2
> a lower limit than what is advertised MAY be enforced.

I will bring this back down to 8k. We can increase/modify this as a followup PR if necessary.

normanmaurer:
@Scottmitch PTAL at the unit test failure: https://garage.netty.io/teamcity/viewLog.html?buildId=15417&buildTypeId=pulls_netty&guest=1

exell-christopher:
One other question/clarification, how do we set the maxHeaderListSize on the decoder so that it's set so that the maxHeaderListSize is enforced on any header frames that come before the client acks the server settings? Right now it appears that the desired way to set that value is to set it via the Http2Settings initialSettings parameter to the Http2Codec CTOR, however, the value of the maxHeaderListSize is not propagated to the decoder until the settings are ack'ed.

Maybe I'm missing something here.

nhnFreespirit:
Per the spec, over https, the client is allowed to send additional frames to the server once it has sent it's own preface but _before_ receiving the servers preface settings frame. Because of this it seems like there will always be a chance of the server receiving headers before the initial settings have been ack'ed.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecUtil.java
```diff
@@ -104,12 +104,13 @@
     public static final int DEFAULT_WINDOW_SIZE = 65535;
     public static final short DEFAULT_PRIORITY_WEIGHT = 16;
     public static final int DEFAULT_HEADER_TABLE_SIZE = 4096;
+    public static final long INITIAL_SETTINGS_HEADER_LIST_SIZE = 8192;
```
nhnFreespirit: For a server, having this always hard coded at 8k leaves us open to clients configured with a different initial value crashing each new connection with an initial headers frame (before seeing settings from the server) large enough to trigger a GO_AWAY, potentially triggering an infinite loop of connection churn.

It seems this should be configurable so it can match whatever the client is configured with?

nhnFreespirit: This is totally @exell-christopher's idea, but stealing it and putting it here: Just grabbing the setting from the initialHttp2 settings passed to the Http2Codec (if present) and using the max from there and defaulting to 8k if no custom settings were passed seems like it would nicely resolve this and maintain sane defaults for simpler cases.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -197,14 +197,12 @@ public void operationComplete(ChannelFuture future) throws Exception {
...
-                stream.headersSent();
                 return promise;
             }
-        } catch (Http2NoMoreStreamIdsException e) {
-            lifecycleManager.onError(ctx, e);
-            return promise.setFailure(e);
-        } catch (Throwable e) {
-            return promise.setFailure(e);
+        } catch (Throwable t) {
+            lifecycleManager.onError(ctx, t);
```
nmittler: good catch ... this looks like a bug.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecUtil.java
Scottmitch: I think this is a different comment and makes sense to allow this value to be configured in the constructors. Note this is the only settings that can be set to a different initial value than what is defined in the RFC.

Scottmitch: I see https://github.com/netty/netty/pull/6230#issuecomment-273542646 now ... which does seem like the same comment :)

Scottmitch:
> Right now it appears that the desired way to set that value is to set it via the Http2Settings initialSettings parameter to the Http2Codec CTOR

@exell-christopher - Good catch. I didn't update the code for the "new API" (code has been updated). However the "stable API" inherits the changes via the settings in the builder here https://github.com/netty/netty/pull/6230/files#diff-39eee1eb2ea60a08c7845b4df9ba33ccR80.

codec-http2/src/main/java/io/netty/handler/codec/http2/internal/hpack/Decoder.java
```diff
@@ -80,25 +83,30 @@
 
     private final DynamicTable dynamicTable;
     private final HuffmanDecoder huffmanDecoder;
+    private long maxHeaderListSizeGoAway;
     private long maxHeaderListSize;
     private long maxDynamicTableSize;
     private long encoderMaxDynamicTableSize;
     private boolean maxDynamicTableSizeChangeRequired;
 
-    public Decoder() {
```
Scottmitch: @nmittler - FYI I intended to move all the HPACK code into the `io.netty.handler.codec.http2` package and make it package private so I'm not too concerned about API compatibility here.

mosesn:
@Scottmitch would this be a good opportunity to tackle https://github.com/netty/netty/issues/5952 too?

Scottmitch:
@mosesn - Lets handle that as an independent PR. The scope of this PR is already bigger than I originally planned.

codec-http2/src/main/java/io/netty/handler/codec/http2/AbstractHttp2ConnectionHandlerBuilder.java
```diff
@@ -334,7 +335,10 @@ protected T build() {
     }
 
     private T buildFromConnection(Http2Connection connection) {
-        Http2FrameReader reader = new DefaultHttp2FrameReader(isValidateHeaders());
+        Long maxHeaderListSize = initialSettings.maxHeaderListSize();
+        Http2FrameReader reader = new DefaultHttp2FrameReader(maxHeaderListSize == null ?
```
netkins: BLOCKER Close this "DefaultHttp2FrameReader". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2095)

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
```diff
@@ -136,7 +136,10 @@ public Http2FrameCodec(boolean server, Http2FrameLogger frameLogger) {
         Http2Connection connection = new DefaultHttp2Connection(server);
         frameWriter = new Http2OutboundFrameLogger(frameWriter, frameLogger);
         Http2ConnectionEncoder encoder = new DefaultHttp2ConnectionEncoder(connection, frameWriter);
-        Http2FrameReader frameReader = new DefaultHttp2FrameReader();
+        Long maxHeaderListSize = initialSettings.maxHeaderListSize();
+        Http2FrameReader frameReader = new DefaultHttp2FrameReader(maxHeaderListSize == null ?
```
netkins: BLOCKER Close this "DefaultHttp2FrameReader". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2095)

codec-http2/src/main/java/io/netty/handler/codec/http2/internal/hpack/Decoder.java
```diff
@@ -36,6 +36,7 @@
 import io.netty.handler.codec.http2.Http2Headers;
 import io.netty.handler.codec.http2.internal.hpack.HpackUtil.IndexType;
 import io.netty.util.AsciiString;
+import io.netty.util.internal.ObjectUtil;
```
netkins: MINOR Remove this unused import 'io.netty.util.internal.ObjectUtil'. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

codec-http2/src/main/java/io/netty/handler/codec/http2/AbstractHttp2ConnectionHandlerBuilder.java
nmittler: I guess this doesn't matter?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -141,6 +141,17 @@ public void close() {
...
 
+    /**
+     * Calculate the threshold in bytes which should trigger a {@code GO_AWAY} if a set of headers exceeds this amount.
+     * @param maxHeaderListSize
+     *      <a href="https://tools.ietf.org/html/rfc7540#section-6.5.2">SETTINGS_MAX_HEADER_LIST_SIZE</a> for the local
+     *      endpoint.
+     * @return the threshold in bytes which should trigger a {@code GO_AWAY} if a set of headers exceeds this amount.
+     */
+    protected long maxHeaderListSizeGoAway(long maxHeaderListSize) {
+        return maxHeaderListSize + (maxHeaderListSize >>> 2);
```
nmittler: maybe just `1.5 * maxHeaderListSize` to make it more immediately clear? Also, do you need to take into account overflow? I guess it's a long so maybe not?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -141,6 +141,17 @@ public void close() {
...
     }
 
+    /**
+     * Calculate the threshold in bytes which should trigger a {@code GO_AWAY} if a set of headers exceeds this amount.
+     * @param maxHeaderListSize
+     *      <a href="https://tools.ietf.org/html/rfc7540#section-6.5.2">SETTINGS_MAX_HEADER_LIST_SIZE</a> for the local
+     *      endpoint.
+     * @return the threshold in bytes which should trigger a {@code GO_AWAY} if a set of headers exceeds this amount.
+     */
+    protected long maxHeaderListSizeGoAway(long maxHeaderListSize) {
```
nmittler: The method name seems to imply we're setting a property.  Consider renaming to something like `calcMaxHeaderListSizeGoAway`.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2FrameWriter.java
```diff
@@ -106,8 +107,8 @@ public Configuration configuration() {
     }
 
     @Override
-    public Http2HeaderTable headerTable() {
-        return headersEncoder.configuration().headerTable();
+    public Http2HeadersEncoder.Configuration headersConfiguration() {
```
nmittler: do we still need this method if it's just delegating to the encoder?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -387,7 +398,7 @@ public void onSettingsAckRead(ChannelHandlerContext ctx) throws Http2Exception {
         private void applyLocalSettings(Http2Settings settings) throws Http2Exception {
```
nmittler: Could we update the method comment to indicate that it is only called after the local settings have been acknowledged by the remote endpoint?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -404,12 +415,12 @@ private void applyLocalSettings(Http2Settings settings) throws Http2Exception {
...
             Long headerTableSize = settings.headerTableSize();
             if (headerTableSize != null) {
-                headerTable.maxHeaderTableSize(headerTableSize);
+                headerConfig.maxHeaderTableSize(headerTableSize);
             }
 
             Long maxHeaderListSize = settings.maxHeaderListSize();
             if (maxHeaderListSize != null) {
-                headerTable.maxHeaderListSize(maxHeaderListSize);
+                headerConfig.maxHeaderListSize(maxHeaderListSize, maxHeaderListSizeGoAway(maxHeaderListSize));
```
nmittler: Just so I understand ... after the remote peer acks our settings, we come here and set the goAway threshold to `1.5 * maxHeadersListSize`? After the ack is received, shouldn't it just be `maxHeadersListSize`?

codec-http2/src/main/java/io/netty/handler/codec/http2/AbstractHttp2ConnectionHandlerBuilder.java
Scottmitch: Nah. Unless an exception is thrown before we transfer close responsibility to `DefaultHttp2ConnectionDecoder`, or before we close on line 356. We can be more defensive about exceptions but practically I'm not sure it will make a difference.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
Scottmitch: renamed to `calculateMaxHeaderListSizeGoAway`

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
Scottmitch: You mean `1.25 * maxHeaderListSize`? I prefer to avoid floating point arithmetic if possible ... I can add a comment to clarify if that helps though? I don't necessarily want to make this behavior part of the interface though so we are free to adjust later.

> Also, do you need to take into account overflow? I guess it's a long so maybe not?

Correct. There will be no overflow because the maximum value is 2^32-1 and we have a 63 bits to work with.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
Scottmitch: Sure.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
Scottmitch: I don't think so, but maybe I don't understand what you are asking. We enforce 2 limits:

1. maxHeadersListSize - This directly correlates to SETTINGS_MAX_HEADER_LIST_SIZE. We choose to interpret this limit as "send a RST_STREAM if it is exceeded.". In order to send a RST_STREAM we have to still process the entire header block, or HPACK will be out of sync (this is why the second limit is necessary).
2. maxHeadersListSizeGoAway - Derived from maxHeadersListSize and provides an upper bound as to how much data we will tolerate before we throw our hands up and send a GO_AWAY.

Both of these limits are derived from the last value we communicated to our peer.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2FrameWriter.java
Scottmitch: The rational is we don't want to expose the `Http2HeadersEncoder` which may make it easier for the user to call `Http2HeadersEncoder#encodeHeaders` which will result in HPACK getting out of sync. Did you have an alternative approach in mind?

Scottmitch:
@nmittler - PTAL

Scottmitch:
4.1 (2fd42cf)

nhnFreespirit:
Thanks for all your work on this @Scottmitch. This has been quite the blocker for us!

exell-christopher:
Thanks for all the hard work on this!! @Scottmitch

Scottmitch:
@nhnFreespirit @exell-christopher - Thanks for the reviews and keep the issue reports coming :)

PR #6231, opened 2017-01-18
HTTP/2 relax test timeouts

Scottmitch:
Motivation:
Build failures have been observed with 2 second timeouts on the CI servers. We should make the timeouts longer to reduce false positive test failures due to tests timing out prematurely.

Modifications:
- Increase timeouts from 2 and 3 seconds to 5 seconds.

Result:
Less false positive test failures.


Scottmitch:
A test failed while building PR https://github.com/netty/netty/pull/6224 due to a 2 second timeout.

normanmaurer:
@Scottmitch LGTM... let me pull in.

normanmaurer:
Cherry-picked into 4.1 (c590e3bd63c43e9c1104ce2ab3a5307f14bee960)

PR #6232, opened 2017-01-18
HTTP/2 limit reserved streams

Scottmitch:
Motivation:
The HTTP/2 RFC provides conflicting direction on how reserved streams should be limited. [Seciton 5.1.2](https://tools.ietf.org/html/rfc7540#section-5.1.2) states `Streams in either of the "reserved" states do not count toward the stream limit.` and [Section 8.2.2](https://tools.ietf.org/html/rfc7540#section-8.2.2) states `A client can use the SETTINGS_MAX_CONCURRENT_STREAMS setting to limit the number of responses that can be concurrently pushed by a server.`. We currently honor section 5.1.2 but this means we have have an implementation specific heuristic to limit on the number of reserved streams.

Modifications:
- Follow section 8.2.2 and include reserved streams toward the SETTINGS_MAX_CONCURRENT_STREAMS count.

Result:
The number of streams in the reserved states are limited according to the RFC instead of an implementation specific heuristic.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
@@ -514,7 +514,7 @@ public class DefaultHttp2Connection implements Http2Connection {
         @Override
         public Http2Stream open(boolean halfClosed) throws Http2Exception {
             state = activeState(id, state, isLocal(), halfClosed);
-            if (!createdBy().canOpenStream()) {
+            if (createdBy().isMaxConcurrentStreamsExhausted()) {
                 throw connectionError(PROTOCOL_ERROR, "Maximum active streams violated for this endpoint.");
             }
             activate();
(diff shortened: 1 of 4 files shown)


Scottmitch:
Note there is a discussion on the httpbis working list attempting to get additional clarity on this issue: https://lists.w3.org/Archives/Public/ietf-http-wg/2017JanMar/0057.html

Scottmitch:
I will wait to merge this until I get clarity on the mailing list. Please feel free to chime in if you think this approach make sense.

Scottmitch:
IIRC there wasn't exactly consensus in the community, and because lack of a good solution interest was lost (https://lists.w3.org/Archives/Public/ietf-http-wg/2017JanMar/0094.html). I think we should take the approach which is less likely to be incompatible with existing implementations, which seems to be "don't count pushed streams toward SETTINGS_MAX_CONCURRENT_STREAMS". I think we should close this and consider the following approaches:

1. Do nothing. The client can enforce a limit on streams in the pushed state.
2. Support a limit on pushed streams in Netty that the user can configure.

Lets discuss in https://github.com/netty/netty/issues/7063

PR #6234, opened 2017-01-18
Check if DnsCache is null in DnsNameResolver constructor.

normanmaurer:
Motivation:

We miss checking if DnsCache is null in DnsNameResolver constructor which will later then lead to a NPE. Better fail fast here.

Modifications:

Check for null and if so throw a NPE.

Result:

Fail fast.


normanmaurer:
Cherry-picked into 4.1 (d7ff71a)

Issue #6192, opened 2017-01-10
Add some warn log about unknown channel option?


wangyuntao:
Should add some WARN log [at this point](https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/DefaultChannelConfig.java#L109) ? like [this one](https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/bootstrap/ServerBootstrap.java#L242) ?

normanmaurer:
@wangyuntao we warn in `ServerBootstrap` and `Bootstrap`. The rest is up to the user when the methods return false.

wangyuntao:
@normanmaurer , I checked the code and found that [Bootstrap](https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/bootstrap/Bootstrap.java#L272) has warn logs, but [ServerBootstrap](https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/bootstrap/ServerBootstrap.java#L144) does't.

normanmaurer:
@wangyuntao ups... let me fix this then.

normanmaurer:
Should be fixed by https://github.com/netty/netty/pull/6235

normanmaurer:
Fixed by https://github.com/netty/netty/pull/6235

PR #6235, opened 2017-01-18
Warn about not-supported ChannelOption when bootstrap Channels.

normanmaurer:
Motivation:

We not warned about not-supported ChannelOptions when set the options for the ServerChannel.

Modifications:

- Share code for setting ChannelOptions during bootstrap

Result:

Warning is logged when a ChannelOption is used that is not supported during bootstrap a Channel. See also [#6192]

### DIFF
--- a/transport/src/main/java/io/netty/bootstrap/AbstractBootstrap.java
+++ b/transport/src/main/java/io/netty/bootstrap/AbstractBootstrap.java
@@ -30,6 +30,7 @@ import io.netty.util.AttributeKey;
 import io.netty.util.concurrent.EventExecutor;
 import io.netty.util.concurrent.GlobalEventExecutor;
 import io.netty.util.internal.StringUtil;
+import io.netty.util.internal.logging.InternalLogger;
 
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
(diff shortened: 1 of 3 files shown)


transport/src/main/java/io/netty/bootstrap/AbstractBootstrap.java
```diff
@@ -436,6 +437,33 @@ final ChannelHandler handler() {
...
+    }
+
+    @SuppressWarnings("unchecked")
+    private static void setChannelOption(
+            Channel channel, ChannelOption<?> option, Object value, InternalLogger logger) {
+        try {
+            if (!channel.config().setOption((ChannelOption<Object>) option, value)) {
+                logger.warn("Unknown channel option '{}' for channel '{}'", option, channel);
+            }
+        } catch (Throwable t) {
```
normanmaurer: @fenik17 I just moved the code around that was there before tho... @Scottmitch @nmittler @trustin WDYT ?

Scottmitch: fine for this PR. We will look at `Throwable` in a more broad sense for https://github.com/netty/netty/issues/6096

normanmaurer: @Scottmitch  +1 :)

normanmaurer:
Cherry-picked into 4.1 (fac0ca8) and 4.0 (018c6ecd7a0ebe4aa9fcbb3b01b682d047c9bd19)

PR #6236, opened 2017-01-18
Fix possible IOOBE when calling ReferenceCountedSslEngine.unwrap(...)…

normanmaurer:
… with heap buffers.

Motivation:

fc3c9c9523150190760801dd0fbf014909519942 introduced a bug which will have ReferenceCountedSslEngine.unwrap(...) produce an IOOBE when be called with an BŷteBuffer as src that contains multiple SSLRecords and has a position != 0.

Modification:

- Correctly set the limit on the ByteBuffer and so fix the IOOBE.
- Add test-case to verify the fix

Result:

Correctly handle heap buffers as well.


normanmaurer:
Cherry-picked into 4.1 (821501717ba3215a81863ad1c329ab8cb130dec4) and 4.0 (f88073f08797e520964ff822fc5a0b4e07b03170)

Scottmitch:
(branch deleted)

PR #6237, opened 2017-01-18
Run all tests in SSLEngineTest with heap, direct and mixed buffers

normanmaurer:
Motivation:

As we use different execution path in our SSLEngine implementation depending on if heap, direct or mixed buffers are used we should run the tests with all of them.

Modification:

Ensure we run all tests with different buffer types.

Result:

Better test-coverage

### DIFF
--- a/handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
@@ -19,6 +19,7 @@ import io.netty.bootstrap.Bootstrap;
 import io.netty.bootstrap.ServerBootstrap;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
+import io.netty.buffer.CompositeByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.buffer.UnpooledByteBufAllocator;
 import io.netty.channel.Channel;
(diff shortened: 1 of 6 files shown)


Scottmitch:
@normanmaurer - Does it make sense to update `OpenSslEngineTest#testWrapHeapBuffersNoWritePendingError` too?

normanmaurer:
@Scottmitch yes... I also renamed the test to `testWrapBuffersNoWritePendingError`. PTAL again

handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
```diff
@@ -186,6 +193,205 @@ protected void channelRead0(ChannelHandlerContext ctx, ByteBuf msg) throws Excep
...
     }
 
+    enum BufferType {
+        Direct,
+        Heap,
+        Mixed
+    }
+
+    @Parameterized.Parameters(name = "{index}: bufferType = {0}")
+    public static Collection<Object[]> data() {
```
trustin: When there is only a single constructor parameter, you don't need to use `Object[]` but just Object.

normanmaurer: ah cool... didnt know that :)

normanmaurer:
Will do

normanmaurer:
Cherry-picked into 4.1 (cf8e07f62f91f1963d501f6f670a6eef8e548da0) and 4.0 (22b255e)

PR #6238, opened 2017-01-18
Ensure calling ReferenceCountedSslEngine.unwrap(...) / wrap(...) can …

normanmaurer:
…be called after it was closed

Motivation:

SSLEngine allows to have unwrap(...) / wrap(...) called even after closeInbound() and closeOutbound() were called. We need to support the same in ReferenceCountedSslEngine.

Modification:

- Allow calling ReferenceCountedSslEngine.unwrap(...) / wrap(...) after the engine was closed
- Modify unit test to ensure correct behaviour.

Result:

Implementation works as expected.


Scottmitch:
`SSLEngine allows` -> `The JDK implementation of SSLEngine allows`?

normanmaurer:
Cherry-picked into 4.1 (91951a5) and 4.0 (7864dd7)

PR #6239, opened 2017-01-18
Ensure SslHandler.sslCloseFuture() is notified in all cases.

normanmaurer:
Motivation:

The SslHandler.sslCloseFuture() may not be notified when the Channel is closed before a closify_notify is received.

Modifications:

Ensure we try to fail the sslCloseFuture() when the Channel is closed.

Result:

Correctly notify the ssl close future.


normanmaurer:
Cherry-picked into 4.1 (d37702a) and 4.0 (876818a)

Issue #6212, opened 2017-01-13
Slf4jLoggerFactory masks all output when using slf4j-simple


johnrengelman:
### Expected behavior
Normal logging occurs to the console when using slf4j-simple

### Actual behavior
Output is hidden.

### Steps to reproduce
Initialize the Slf4jLoggerFactory before any Slf4jLogger (.e.g SimpleLogger)

### Netty version
4.1.5.FINAL

### JVM version (e.g. `java -version`)
java version "1.8.0_112"
Java(TM) SE Runtime Environment (build 1.8.0_112-b16)
Java HotSpot(TM) 64-Bit Server VM (build 25.112-b16, mixed mode)

### OS version (e.g. `uname -a`)
Darwin <hostname> 14.5.0 Darwin Kernel Version 14.5.0: Mon Jan 11 18:48:35 PST 2016; root:xnu-2782.50.2~1/RELEASE_X86_64 x86_64

Found this when working Ratpack. We moved some code around such that `ResourceLeakDetector.setLevel(ResourceLeakDetector.Level.DISABLED);` was called prior to any Sl4fj logger being instantiated in a `static` context.

This cause `Slf4jLoggerFactory` to replace `System.err` with it's own implementation which caused slf4j (with the slf4j-simple binding) to pick it up and use it for it's loggers. The implementation in `Slf4jLoggerFactory` never outputs to the original `System.err`. It simply appends all data to a `StringBuffer`.

https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/internal/logging/Slf4JLoggerFactory.java#L52

Notice that the buffer is only flushed at instantiation here: https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/internal/logging/Slf4JLoggerFactory.java#L63

normanmaurer:
@johnrengelman let me check...

normanmaurer:
Fixed by https://github.com/netty/netty/pull/6240

PR #6240, opened 2017-01-18
Do not replace System.err during Slf4LoggerFactory construction

normanmaurer:
Motivation:

Replacing System.err during Slf4LoggerFactory construction is problematic as another class may optain the System.err reference before we set it back to the original value.

Modifications:

Remove code that temporary replaced System.err.

Result:

Fixes [#6212].

### DIFF
--- a/common/src/main/java/io/netty/util/internal/logging/Slf4JLoggerFactory.java
+++ b/common/src/main/java/io/netty/util/internal/logging/Slf4JLoggerFactory.java
@@ -19,16 +19,13 @@ package io.netty.util.internal.logging;
 import org.slf4j.LoggerFactory;
 import org.slf4j.helpers.NOPLoggerFactory;
 
-import java.io.OutputStream;
-import java.io.PrintStream;
-import java.io.UnsupportedEncodingException;
-
 /**
  * Logger factory which creates a <a href="http://www.slf4j.org/">SLF4J</a>
  * logger.
  */
 public class Slf4JLoggerFactory extends InternalLoggerFactory {
 
+    @SuppressWarnings("deprecation")
     public static final InternalLoggerFactory INSTANCE = new Slf4JLoggerFactory();
 
     /**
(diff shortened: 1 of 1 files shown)


Scottmitch:
`as another class may optain the ...` -> `as another class may also use System.err before we set it back to the original value.`

common/src/main/java/io/netty/util/internal/logging/Slf4JLoggerFactory.java
```diff
@@ -40,31 +37,8 @@ public Slf4JLoggerFactory() {
 
     Slf4JLoggerFactory(boolean failIfNOP) {
         assert failIfNOP; // Should be always called with true.
-
-        // SFL4J writes it error messages to System.err. Capture them so that the user does not see such a message on
```
Scottmitch: Does this provide any utility or was it was a mistake to do this in the first place? The original issue seems to focus more on the content is not flushed at the appropriate time. I wonder if the original intention was to capture the output and then flush it after the logger is setup ... not sure though.

normanmaurer: I think it does more harm then any good... So I think we should jus remove it. Thats why I added @trustin as reviewer as he wrote the code ;)

trustin: It was really long time ago I wrote this, so I'm not completely sure.

Once upon a time, SLF4J was not a dominant logging framework, lots of people were still using commons-logging or log4j while some others use SLF4J. As a result, our classpath had slf4j-api.jar, but not the implementation JAR (e.g. slf4j-log4j.jar). If Netty chose SLF4J just because slf4j-api.jar is in the classpath, nothing will be logged at all because NOPLoggerFactory would be used.

Therefore, when the current LoggerFactory is NOPLoggerFactory, Slf4JLoggerFactory raised NoClassDefException so that Netty tries other logging framework automatically. However, during this process, SLF4J will write some warning/error messages to System.err. I thought this information is not very useful to users who expects Netty to use other logging framework than SLF4J, so I decided to hide it.

But it alters the global resource (System.err) which is obviously not a good idea. Please feel free to remove it.

normanmaurer: Thanks... if @Scottmitch is fine with it I will just remove it. @Scottmitch WDYT ?

Scottmitch: thanks for clarifying @trustin ... lets kill it

normanmaurer:
Cherry-picked into 4.1 (71efb0b39e4410001abb9d96c6a87316a567c16f) and 4.0 (d3e99cd75e2156dd2e70b84a95f74569948d597c)

PR #6241, opened 2017-01-18
Respect resolvedAddressTypes when follow CNAME records.

normanmaurer:
Motivation:

When we follow CNAME records we should respect resolvedAddressTypes and only query A / AAAA depending on which address types are expected.

Modifications:

Check if we should query A / AAAA when follow CNAMEs depending on resolvedAddressTypes.

Result:

Correct behaviour when follow CNAMEs.

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
@@ -35,6 +35,7 @@ import io.netty.handler.codec.dns.DnsRawRecord;
 import io.netty.handler.codec.dns.DnsRecord;
 import io.netty.handler.codec.dns.DatagramDnsResponseDecoder;
 import io.netty.handler.codec.dns.DnsQuestion;
+import io.netty.handler.codec.dns.DnsRecordType;
 import io.netty.handler.codec.dns.DnsResponse;
 import io.netty.resolver.HostsFileEntriesResolver;
 import io.netty.resolver.InetNameResolver;
@@ -59,7 +60,9 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Set;
 
 import static io.netty.util.internal.ObjectUtil.*;
 
@@ -148,6 +151,10 @@ public class DnsNameResolver extends InetNameResolver {
     private final HostsFileEntriesResolver hostsFileEntriesResolver;
     private final String[] searchDomains;
     private final int ndots;
+    private final boolean cnameFollowARecords;
+    private final boolean cnameFollowAAAARecords;
+    private final InternetProtocolFamily preferredAddressType;
+    private final DnsRecordType[] resolveRecordTypes;
 
     /**
      * Creates a new DNS-based name resolver that communicates with the specified list of DNS servers.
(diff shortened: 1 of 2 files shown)


resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
```diff
@@ -518,9 +518,31 @@ private void followCname(InetSocketAddress nameServerAddr, String name, String c
             trace.append(cname);
         }
 
+        boolean queryARecord = false;
+        boolean queryAAAARecord = false;
+        for (InternetProtocolFamily family: resolveAddressTypes) {
```
Scottmitch: we iterate over `resolveAddressTypes` which is a `final` array and I don't think the intention is the contents will change. I wonder if we should predetermine these types of conditions and store them in `final` variables to avoid constantly iterating. For example having a member variables:

```java
private final boolean cnameFollowARecords;
private final boolean cnameFollowAAAARecords;
```

We can check the pre-condition that at least one must be set in the constructor too.

normanmaurer: @Scottmitch yes this makes sense... let me do this.

normanmaurer:
@Scottmitch PTAL again... also did move some other code to only need to match one time :)

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -200,6 +207,32 @@ public DnsNameResolver(
         this.searchDomains = checkNotNull(searchDomains, "searchDomains").clone();
         this.ndots = checkPositiveOrZero(ndots, "ndots");
 
+        boolean cnameFollowARecords = false;
+        boolean cnameFollowAAAARecords = false;
+        // Use LinkedHashSet to maintain correct ordering.
+        Set<DnsRecordType> recordTypes = new LinkedHashSet<DnsRecordType>(resolvedAddressTypes.length);
```
trustin: Could be more efficient by allocating `DnsRecordType[]` directly given that resolvedAddressTypes.length is very small?

normanmaurer: @trustin its only done when create a new resolver so honestly I think it makes no difference at all.

trustin: With proper validation on resolvedAddressTypes, resolvedAddressTypes.length will always be 1 or 2.

normanmaurer: Imho the code will be more complex otherwise as we may have duplicates.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
```diff
@@ -409,7 +397,7 @@ private boolean gotPreferredAddress() {
         }
 
         final int size = resolvedEntries.size();
-        switch (resolveAddressTypes[0]) {
+        switch (parent.preferredAddressType()) {
```
netkins: MAJOR Add a default case to this switch. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3ASwitchLastCaseIsDefaultCheck)

Scottmitch: I guess what we typically do is throw a `default: throw new Error();`?

normanmaurer: yes let me add.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
normanmaurer: Let me leave it as it is for now... If you really feel strong we can do a follow up.

normanmaurer:
Cherry-picked into 4.1 (ead87b7)

PR #6242, opened 2017-01-18
Add SslCloseCompletionEvent that is fired once a close_notify was rec…

normanmaurer:
…eived

Motivation:

For the completion of a handshake we already fire a SslHandshakeCompletionEvent which the user can intercept. We should do the same for the receiving of close_notify.

Modifications:

Add SslCloseCompletionEvent and test-case.

Result:

More consistent API.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslCloseCompletionEvent.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslCloseCompletionEvent.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013 The Netty Project
+ * Copyright 2017 The Netty Project
  *
  * The Netty Project licenses this file to you under the Apache License,
  * version 2.0 (the "License"); you may not use this file except in compliance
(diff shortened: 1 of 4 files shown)


handler/src/main/java/io/netty/handler/ssl/SslCloseCompletionEvent.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.ssl;
+
+
```
Scottmitch: nit: remove extra like

normanmaurer:
Cherry-picked into 4.1 (d55c321306dc6f3c01cbbd5f5c080fabe1128583) and 4.0 (0ccab2b55b9582814c6684b3d071973daf58fa75)

PR #6243, opened 2017-01-18
Deprecate methods on SslHandler that have other replacements

normanmaurer:
Motivation:

    SslHandler has multiple methods which have better replacements now or are obsolete. We should mark these as `@Deprecated`.

    Modifications:

    Mark methods as deprecated.

    Result:

    API cleanup preparation.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -432,14 +432,20 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
     /**
      * Sends an SSL {@code close_notify} message to the specified channel and
      * destroys the underlying {@link SSLEngine}.
+     *
+     * @deprecated use {@link Channel#close()} or {@link ChannelHandlerContext#close()}
      */
+    @Deprecated
     public ChannelFuture close() {
         return close(ctx.newPromise());
     }
 
     /**
      * See {@link #close()}
+     *
+     * @deprecated use {@link Channel#close()} or {@link ChannelHandlerContext#close()}
      */
+    @Deprecated
     public ChannelFuture close(final ChannelPromise promise) {
         final ChannelHandlerContext ctx = this.ctx;
         ctx.executor().execute(new Runnable() {


handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -412,22 +412,30 @@ public String applicationProtocol() {
      *
      * @return the {@link Future} for the iniital TLS handshake if {@link #renegotiate()} was not invoked.
      *         The {@link Future} for the most recent {@linkplain #renegotiate() TLS renegotiation} otherwise.
+     * @deprecated intercept {@link SslHandshakeCompletionEvent}.
      */
+    @Deprecated
     public Future<Channel> handshakeFuture() {
```
trustin: Hmm, really? I thought this is really useful, especially when writing a client.

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -454,7 +462,9 @@ public void run() {
      * This method will return the same {@link Future} all the time.
      *
      * @see SSLEngine
+     * @deprecated intercept {@link SslCloseCompletionEvent}.
      */
+    @Deprecated
     public Future<Channel> sslCloseFuture() {
```
trustin: Ditto

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
normanmaurer: Why not just use the events and userEventTriggered(...) ?

trustin: A user could use the events if he or she writes the sending part of the client in ChannelHandler but it's often more convenient to write from outside of ChannelHandler. e.g. a user writes a synchronous client which is part of a command line utility:

```java
    Channel ch = ...;
    sslHandler.handshakeFuture().sync();
    ch.write(firstRequest).sync();
    mainHandler.receivedMessages().take()...
```

normanmaurer: @trustin hmm true... @Scottmitch makes sense to you as well ? If so let us just close this one.

trustin: I think it's OK to deprecate close() though.

normanmaurer: @trustin +1... close makes no sense anymore.

Scottmitch: I'm on the fence ... I would prefer to not have multiple ways of doing the same thing. The user can always translate this to a future even if it is more convenient for them.

normanmaurer: @Scottmitch let us just only deprecate close for now and then we can re-evaluate if needed. Ok ?

Scottmitch: sgtm

normanmaurer: ok will take care.

normanmaurer:
Cherry-picked into 4.1 (1b31397) and 4.0 (a219519)

PR #6244, opened 2017-01-18
Check if Epoll is avaible when construct EpollEventLoopGroup

normanmaurer:
Motivation:

We should call Epoll.ensureAvailability() when init EpollEventLoopGroup to fail fast and with a proper exception.

Modifications:

Call Epoll.ensureAvailability() during EpollEventLoopGroup init.

Result:

Fail fast if epoll is not availability (for whatever reason).


Scottmitch:
I took a different approach in the kqueue PR: https://github.com/netty/netty/pull/6000/files#diff-db3e069239a403b954e3ebc024ba9507R53. Do you prefer the approach in this PR more?

normanmaurer:
@Scottmitch not sure if its better or not but its more inline in what we do in our openssl based impls as well

Scottmitch:
@normanmaurer - I guess we really only need to check at class load time though right ... these conditions are not expected to change over time right?

normanmaurer:
Cherry-picked into 4.1 (81735c2535591a036a534efd41c5f2b56fa04fbb) and 4.0 (8381e7b)

PR #6247, opened 2017-01-18
Improvement : Removed unnecessary pattern matching during number parsing and unneces…

doom369:
Motivation:

Pattern matching not necessary for number parsing.

Modification:

Removed pattern matching for number parsing and removed unnecessary `toLowerCase()` operation.

Result:

No static variable with pattern, removed unnecessary matching operation and `toLowerCase()` operation.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/SystemPropertyUtil.java
+++ b/common/src/main/java/io/netty/util/internal/SystemPropertyUtil.java
@@ -20,7 +20,6 @@ import io.netty.util.internal.logging.InternalLoggerFactory;
 
 import java.security.AccessController;
 import java.security.PrivilegedAction;
-import java.util.regex.Pattern;
 
 /**
  * A collection of utility methods to retrieve and parse the values of the Java system properties.
@@ -123,8 +122,6 @@ public final class SystemPropertyUtil {
         return def;
     }
 
-    private static final Pattern INTEGER_PATTERN = Pattern.compile("-?[0-9]+");
-
     /**
      * Returns the value of the Java system property with the specified
      * {@code key}, while falling back to the specified default value if
@@ -140,13 +137,11 @@ public final class SystemPropertyUtil {
             return def;
         }
 
-        value = value.trim().toLowerCase();
-        if (INTEGER_PATTERN.matcher(value).matches()) {
-            try {
-                return Integer.parseInt(value);
-            } catch (Exception e) {
-                // Ignore
-            }
+        value = value.trim();
+        try {
+            return Integer.parseInt(value);
+        } catch (Exception e) {
+            // Ignore
         }
 
         logger.warn(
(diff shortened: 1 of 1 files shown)


common/src/main/java/io/netty/util/internal/SystemPropertyUtil.java
```diff
@@ -172,13 +167,11 @@ public static long getLong(String key, long def) {
...
-        if (INTEGER_PATTERN.matcher(value).matches()) {
-            try {
-                return Long.parseLong(value);
-            } catch (Exception e) {
-                // Ignore
-            }
+        value = value.trim();
+        try {
+            return Long.parseLong(value);
+        } catch (Exception e) {
```
fenik17: Need catch only NumberFormatException.

common/src/main/java/io/netty/util/internal/SystemPropertyUtil.java
```diff
@@ -140,13 +137,11 @@ public static int getInt(String key, int def) {
...
-        if (INTEGER_PATTERN.matcher(value).matches()) {
-            try {
-                return Integer.parseInt(value);
-            } catch (Exception e) {
-                // Ignore
-            }
+        value = value.trim();
+        try {
+            return Integer.parseInt(value);
+        } catch (Exception e) {
```
fenik17: Need catch only NumberFormatException.

common/src/main/java/io/netty/util/internal/SystemPropertyUtil.java
```diff
@@ -140,13 +137,11 @@ public static int getInt(String key, int def) {
...
-        value = value.trim().toLowerCase();
-        if (INTEGER_PATTERN.matcher(value).matches()) {
-            try {
-                return Integer.parseInt(value);
-            } catch (Exception e) {
-                // Ignore
-            }
+        value = value.trim();
+        try {
+            return Integer.parseInt(value);
```
fenik17: Better use Integer.parseInt(value, 10) for call stack decrease.

common/src/main/java/io/netty/util/internal/SystemPropertyUtil.java
```diff
@@ -172,13 +167,11 @@ public static long getLong(String key, long def) {
...
-        value = value.trim().toLowerCase();
-        if (INTEGER_PATTERN.matcher(value).matches()) {
-            try {
-                return Long.parseLong(value);
-            } catch (Exception e) {
-                // Ignore
-            }
+        value = value.trim();
+        try {
+            return Long.parseLong(value);
```
fenik17: Better use Long.parseLong(value, 10) for call stack decrease.

common/src/main/java/io/netty/util/internal/SystemPropertyUtil.java
doom369: This is not hot method. Just properties parser. No need for that.

common/src/main/java/io/netty/util/internal/SystemPropertyUtil.java
doom369: Make no difference.

Scottmitch:
@doom369 - Can you also sign the [ICLA](https://docs.google.com/spreadsheet/viewform?formkey=dHBjc1YzdWhsZERUQnhlSklsbG1KT1E6MQ)?

doom369:
@Scottmitch I did it before. Submitted one more time.

normanmaurer:
Cherry-picked into 4.1 (141554f5d15d0014eccefa8de42388bda1bfa255) and 4.0 (afd493362dd59494a9c5e8e6e1326bbba580fcfd)

Issue #6246, opened 2017-01-18
4.1.7: java.lang.NoClassDefFoundError: Could not initialize class io.netty.util.internal.PlatformDependent0


jesperpedersen:
### Expected behavior

No error

### Actual behavior

```
Caused by: java.lang.NoClassDefFoundError: Could not initialize class io.netty.util.internal.PlatformDependent0
        at io.netty.util.internal.PlatformDependent.getClassLoader(PlatformDependent.java:881)
        at io.netty.channel.DefaultChannelId.defaultProcessId(DefaultChannelId.java:110)
        at io.netty.channel.DefaultChannelId.<clinit>(DefaultChannelId.java:78)
        at io.netty.channel.AbstractChannel.newId(AbstractChannel.java:107)
        at io.netty.channel.AbstractChannel.<init>(AbstractChannel.java:79)
        at io.netty.channel.nio.AbstractNioChannel.<init>(AbstractNioChannel.java:84)
        at io.netty.channel.nio.AbstractNioByteChannel.<init>(AbstractNioByteChannel.java:54)
        at io.netty.channel.socket.nio.NioSocketChannel.<init>(NioSocketChannel.java:97)
        at io.netty.channel.socket.nio.NioSocketChannel.<init>(NioSocketChannel.java:87)
        at io.netty.channel.socket.nio.NioSocketChannel.<init>(NioSocketChannel.java:80)
        at io.netty.channel.socket.nio.NioSocketChannel.<init>(NioSocketChannel.java:73)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
        at java.lang.Class.newInstance(Class.java:442)
        at io.netty.channel.ReflectiveChannelFactory.newChannel(ReflectiveChannelFactory.java:38)
        at io.netty.bootstrap.AbstractBootstrap.initAndRegister(AbstractBootstrap.java:319)
        at io.netty.bootstrap.Bootstrap.doResolveAndConnect(Bootstrap.java:163)
        at io.netty.bootstrap.Bootstrap.connect(Bootstrap.java:145)
```

### Minimal yet complete reproducer code (or URL to code)

[...]

PR #6251, opened 2017-01-19
PlatformDependent#getClassLoader fails in restrictive classloader env…

Scottmitch:
…ironment

Motivation:
https://github.com/netty/netty/pull/6042 only addressed PlatformDependent#getSystemClassLoader but getClassLoader is also called in an optional manner in some common code paths but fails to catch a general enough exception to continue working.

Modifications:
- Calls to getClassLoader which can continue if results fail should catch Throwable

Result:
More resilient code in the presense of restrictive class loaders.
Fixes https://github.com/netty/netty/issues/6246.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -854,8 +854,9 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
                             && "com.sun.nio.sctp.SctpChannel".equals(clazz.getSuperclass().getName())) {
                         return true;
                     }
-                } catch (ClassNotFoundException e) {
-                    // This should not happen just ignore
+                } catch (Throwable cause) {
+                    logger.debug("Unexpected exception while loading class {} classname {}",
+                                 getClass(), classname, cause);
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


Scottmitch:
@jesperpedersen - please verify this fix. Also I'm not sure why you get `NoClassDefFoundError` in the first place. You mentioned in https://github.com/netty/netty/issues/6038#issue-190356174 `WildFly is using an isolated classloader environment based on JBoss Modules.` ... any ideas why `PlatformDependent0` would no be loaded?

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -851,8 +851,8 @@ private boolean ignoreException(Throwable t) {
                             && "com.sun.nio.sctp.SctpChannel".equals(clazz.getSuperclass().getName())) {
                         return true;
                     }
-                } catch (ClassNotFoundException e) {
-                    // This should not happen just ignore
+                } catch (Throwable cause) {
+                    logger.debug("Unexpected exception", cause);
```
normanmaurer: @Scottmitch can you use some better logging message ?

Maybe something like `Unexpected exception while loading class {}`, class, cause)`

normanmaurer:
@Scottmitch one nit...

jesperpedersen:
This works, with the following log message:
```
08:57:10,909 WARNING [io.netty.channel.DefaultChannelId] (management-handler-thread - 3) Failed to find the current process ID from ''; using a random value: 970346034
```

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
Scottmitch: done

Scottmitch:
@jesperpedersen - Thanks for verifying. That warning makes sense ... as I said I'm still not sure why the class loading would fail in your environment.

Scottmitch:
4.1 (9a4aa61) 4.0 (7528f11)

PR #6252, opened 2017-01-19
Use less memory during writes when using SslHandler with SslProvider.…

normanmaurer:
…OpenSsl

Motivation:

In commit fc3c9c9523150190760801dd0fbf014909519942 I changes the way how we calculate the capacity of the needed ByteBuf for wrap operations that happen during writes when the SslHandler is used. This had the effect that the same capacity for ByteBufs is needed for the JDK implementation of SSLEngine but also for our SSLEngine implementation that uses OpenSSL / BoringSSL / LibreSSL. Unfortunally this had the side-effect that applications that used our SSLEngine implementation now need a lot more memory as bascially the JDK implementation always needs a 16kb buffer for each wrap while we can do a lot better for our SSLEngine implementation.

Modification:

- Resurrect code that calculate a better ByteBuf capacity when using our SSLEngine implementation and so be able to safe a lot of memory
- Add test-case to ensure it works as expected and is not removed again later on.

Result:

Memory footprint of applications that uses our SSLEngine implementation based on OpenSSL / BoringSSL / LibreSSL is back to the same amount of before commit fc3c9c9523150190760801dd0fbf014909519942.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -62,6 +62,8 @@ import static io.netty.handler.ssl.OpenSsl.memoryAddress;
 import static io.netty.util.internal.EmptyArrays.EMPTY_CERTIFICATES;
 import static io.netty.util.internal.EmptyArrays.EMPTY_JAVAX_X509_CERTIFICATES;
 import static io.netty.util.internal.ObjectUtil.checkNotNull;
+import static java.lang.Math.max;
+import static java.lang.Math.min;
 import static javax.net.ssl.SSLEngineResult.HandshakeStatus.FINISHED;
 import static javax.net.ssl.SSLEngineResult.HandshakeStatus.NEED_UNWRAP;
 import static javax.net.ssl.SSLEngineResult.HandshakeStatus.NEED_WRAP;
@@ -158,6 +160,8 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
     static final int MAX_ENCRYPTED_PACKET_LENGTH = MAX_CIPHERTEXT_LENGTH + 5 + 20 + 256;
 
     static final int MAX_ENCRYPTION_OVERHEAD_LENGTH = MAX_ENCRYPTED_PACKET_LENGTH - MAX_PLAINTEXT_LENGTH;
+
+    private static final int MAX_ENCRYPTION_OVERHEAD_DIFF = Integer.MAX_VALUE - MAX_ENCRYPTION_OVERHEAD_LENGTH;
 
     private static final AtomicIntegerFieldUpdater<ReferenceCountedOpenSslEngine> DESTROYED_UPDATER =
             AtomicIntegerFieldUpdater.newUpdater(ReferenceCountedOpenSslEngine.class, "destroyed");
@@ -461,7 +465,7 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
         } else {
             final int pos = dst.position();
             final int limit = dst.limit();
-            final int len = Math.min(MAX_ENCRYPTED_PACKET_LENGTH, limit - pos);
+            final int len = min(MAX_ENCRYPTED_PACKET_LENGTH, limit - pos);
             final ByteBuf buf = alloc.directBuffer(len);
             try {
                 final long addr = memoryAddress(buf);
@@ -614,7 +618,30 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
                 }
             }
 
-            if (dst.remaining() < MAX_ENCRYPTED_PACKET_LENGTH) {
+            int endOffset = offset + length;
+            int srcsLen = 0;
+
+            for (int i = offset; i < endOffset; ++i) {
+                final ByteBuffer src = srcs[i];
+                if (src == null) {
+                    throw new IllegalArgumentException("srcs[" + i + "] is null");
+                }
+                if (srcsLen == MAX_PLAINTEXT_LENGTH) {
+                    continue;
+                }
+
+                srcsLen += src.remaining();
+                if (srcsLen > MAX_PLAINTEXT_LENGTH || srcsLen < 0) {
+                    // If srcLen > MAX_PLAINTEXT_LENGTH or secLen < 0 just set it to MAX_PLAINTEXT_LENGTH.
+                    // This also help us to guard against overflow.
+                    // We not break out here as we still need to check for null entries in srcs[].
+                    srcsLen = MAX_PLAINTEXT_LENGTH;
+                }
+            }
+
+            int maxEncryptedLen = calculateOutNetBufSize(srcsLen);
+
+            if (dst.remaining() < maxEncryptedLen) {
                 // Can not hold the maximum packet so we need to tell the caller to use a bigger destination
                 // buffer.
                 return new SSLEngineResult(BUFFER_OVERFLOW, getHandshakeStatus(), 0, 0);
@@ -622,18 +649,14 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
             // There was no pending data in the network BIO -- encrypt any application data
             int bytesProduced = 0;
             int bytesConsumed = 0;
-            int endOffset = offset + length;
 
             loop: for (int i = offset; i < endOffset; ++i) {
                 final ByteBuffer src = srcs[i];
-                if (src == null) {
-                    throw new IllegalArgumentException("srcs[" + i + "] is null");
-                }
                 while (src.hasRemaining()) {
                     final SSLEngineResult pendingNetResult;
                     // Write plaintext application data to the SSL engine
                     int result = writePlaintextData(
-                            src, Math.min(src.remaining(), MAX_PLAINTEXT_LENGTH - bytesConsumed));
+                            src, min(src.remaining(), MAX_PLAINTEXT_LENGTH - bytesConsumed));
 
                     if (result > 0) {
                         bytesConsumed += result;
@@ -837,7 +860,7 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
                     }
                     // Write more encrypted data into the BIO. Ensure we only read one packet at a time as
                     // stated in the SSLEngine javadocs.
-                    int written = writeEncryptedData(src, Math.min(packetLengthRemaining, src.remaining()));
+                    int written = writeEncryptedData(src, min(packetLengthRemaining, src.remaining()));
                     if (written > 0) {
                         packetLengthRemaining -= written;
                         if (packetLengthRemaining == 0) {
@@ -1659,6 +1682,11 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
 
     private boolean isDestroyed() {
         return destroyed != 0;
+    }
+
+    static int calculateOutNetBufSize(int pendingBytes) {
+        return min(MAX_ENCRYPTED_PACKET_LENGTH, MAX_ENCRYPTION_OVERHEAD_LENGTH
+                + min(MAX_ENCRYPTION_OVERHEAD_DIFF, pendingBytes));
     }
 
     private final class OpenSslSession implements SSLSession, ApplicationProtocolAccessor {
--- a/handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
@@ -26,12 +26,16 @@ import org.junit.BeforeClass;
 import org.junit.Test;
 
 import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.List;
 
 import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLEngineResult;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertTrue;
 import static org.junit.Assume.assumeTrue;
 
 public class OpenSslEngineTest extends SSLEngineTest {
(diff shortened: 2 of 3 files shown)


handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -614,21 +614,33 @@ public final SSLEngineResult wrap(
...
-            if (dst.remaining() < MAX_ENCRYPTED_PACKET_LENGTH) {
+            int endOffset = offset + length;
+            long maxEncryptedLen = 0;
+
+            for (int i = offset; i < endOffset; ++i) {
+                final ByteBuffer src = srcs[i];
+                if (src == null) {
+                    throw new IllegalArgumentException("srcs[" + i + "] is null");
+                }
+                maxEncryptedLen = src.remaining();
```
fenik17: May be was meant += or Math.max()?

normanmaurer: ups yes... +=

normanmaurer: Let me add a test for this as well just to be sure.

normanmaurer: @fenik17 thanks for spotting it... https://github.com/netty/netty/pull/6252/commits/e9d9e59cb10a4258017188646f278fb7d4fba590 contains a unit test for it.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -614,21 +614,32 @@ public final SSLEngineResult wrap(
...
             }
 
-            if (dst.remaining() < MAX_ENCRYPTED_PACKET_LENGTH) {
+            int endOffset = offset + length;
+            long maxEncryptedLen = MAX_ENCRYPTION_OVERHEAD_LENGTH;
+
+            for (int i = offset; i < endOffset; ++i) {
+                final ByteBuffer src = srcs[i];
+                if (src == null) {
+                    throw new IllegalArgumentException("srcs[" + i + "] is null");
```
trustin: Not strong on this, but shouldn't this be NPE for consistency? i.e. `new NullPointerException("srcs[" + i + ']')`

normanmaurer: @trustin its an `IllegalArgumentException` as stated in the SSLEngine javadocs: http://docs.oracle.com/javase/7/docs/api/javax/net/ssl/SSLEngine.html#wrap(java.nio.ByteBuffer[],%20java.nio.ByteBuffer)

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -614,21 +614,34 @@ public final SSLEngineResult wrap(
...
+                }
+                srcsLen += src.remaining();
+            }
+
+            int maxEncryptedLen = calculateOutNetBufSize(srcsLen);
+
+            // If overflow happened just use MAX_PLAINTEXT_LENGTH.
+            if (maxEncryptedLen < 0 || MAX_ENCRYPTED_PACKET_LENGTH < maxEncryptedLen) {
```
normanmaurer: @Scottmitch @trustin @nmittler in theory this could fail to detect overflow if it overflows enough to again be >= 0. Not sure if we need to care enough about it to justify moving the check inside the for loop and so have more branches. WDYT ?

johnou: naïve question, but if that occurs what will happen?

normanmaurer: you mean the overflow ? We may not correctly return BUFFER_OVERFLOW. This will only happen tho if the sum of the remaining bytes in srcs is super big. This should not happen in practice.

normanmaurer: That said I think a few more branches is not the end of the world and be sure we never overflow. Let me fix it and add another test for it.

normanmaurer: nah scratch that... for this I would need 2 gb or memory for the test. Let me just fix it without a test.

johnou: Right but if we do not return BUFFER_OVERFLOW what is going to happen? SSL engine state messes up and the connection will drop?

normanmaurer: yep yep... Fixed and added test in latest commit eb4924531f9d34076d67d1e6b5a5794dd1ab89f0

johnou: would it be worth adding a test for BUFFER_UNDERFLOW too?

normanmaurer:
There is one in SSLEngineTest.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
johnou: I will take your word for it that it is tested, just that I didn't see any references to the BUFFER_UNDERFLOW enum value in test classes (4.0 branch).

normanmaurer:
Its tested by our integration tests. But let me do another PR which adds explicit test for it for the engine itself. Cant harm.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -614,21 +614,43 @@ public final SSLEngineResult wrap(
...
+                srcsLen += src.remaining();
+                if (srcsLen > MAX_PLAINTEXT_LENGTH) {
+                    // If srcLen > MAX_PLAINTEXT_LENGTH just set it to MAX_PLAINTEXT_LENGTH.
+                    // This also help us to guard against overflow.
+                    // We not break out here as we still need to check for null entries in srcs[].
+                    srcsLen = MAX_PLAINTEXT_LENGTH;
+                }
+            }
+
+            int maxEncryptedLen = calculateOutNetBufSize(srcsLen);
```
nmittler: nit: maybe just use min?

```java
final maxEncryptedLen = Math.min(MAX_ENCRYPTED_PACKET_LENGTH,
    calculateOutNetBufSize(srcLen))
```

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -614,21 +614,43 @@ public final SSLEngineResult wrap(
...
+
+            if (MAX_ENCRYPTED_PACKET_LENGTH < maxEncryptedLen) {
```
Scottmitch: `calculateOutNetBufSize` already takes `MAX_ENCRYPTED_PACKET_LENGTH` into account so this condition isn't necessary.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1661,6 +1683,10 @@ private boolean isDestroyed() {
         return destroyed != 0;
     }
 
+    static int calculateOutNetBufSize(int pendingBytes) {
+        return Math.min(MAX_ENCRYPTED_PACKET_LENGTH, MAX_ENCRYPTION_OVERHEAD_LENGTH + pendingBytes);
```
Scottmitch: nit: add a static import for `Math.min` it is used in multiple spots.

Scottmitch: Also are we subject to overflow here?

```java
min(MAX_ENCRYPTED_PACKET_LENGTH, max(MAX_ENCRYPTION_OVERHEAD_LENGTH + pendingBytes, Integer.MAX_VALUE));
```

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
```diff
@@ -109,6 +113,142 @@ public void testWrapHeapBuffersNoWritePendingError() throws Exception {
...
+            assertEquals(0, result.bytesConsumed());
+            assertEquals(0, result.bytesProduced());
+            assertEquals(src.remaining(), src.capacity());
+            assertEquals(dst.remaining(), dst.capacity());
+
+            // Check that we can wrap with a dst buffer that has the capacity of
+            // src.capacity() + ReferenceCountedOpenSslEngine.MAX_ENCRYPTION_OVERHEAD_LENGTH
+            result = clientEngine.wrap(src, dst);
+            assertEquals(SSLEngineResult.Status.OK, result.getStatus());
+            assertEquals(src.capacity(), result.bytesConsumed());
```
Scottmitch: `capacity()` doesn't change as a result of the `wrap` call. instead `pos` is advanced ... do you mean to check that `pos` was advanced by `result.bytesConsumed()` amount?

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
```diff
@@ -109,6 +113,142 @@ public void testWrapHeapBuffersNoWritePendingError() throws Exception {
...
+            assertTrue(result.bytesProduced() > src.capacity());
```
Scottmitch: same comment as above ... `capacity()` doesn't change.

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
```diff
@@ -109,6 +113,142 @@ public void testWrapHeapBuffersNoWritePendingError() throws Exception {
...
+                .sslProvider(sslServerProvider())
+                .build();
+        SSLEngine clientEngine = null;
+        SSLEngine serverEngine = null;
+        try {
+            clientEngine = clientSslCtx.newEngine(UnpooledByteBufAllocator.DEFAULT);
+            serverEngine = serverSslCtx.newEngine(UnpooledByteBufAllocator.DEFAULT);
+            handshake(clientEngine, serverEngine);
+
+            ByteBuffer src = ByteBuffer.allocate(1024); // 10mb
```
Scottmitch: `10mb` -> `1kb`? or just remove the comment?

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
```diff
@@ -109,6 +113,142 @@ public void testWrapHeapBuffersNoWritePendingError() throws Exception {
...
+            List<ByteBuffer> srcList = new ArrayList<ByteBuffer>();
+            long srcsLen = 0;
+            long maxLen = ((long) Integer.MAX_VALUE) * 2;
+
+            while (srcsLen < maxLen) {
+                ByteBuffer dup = src.duplicate();
+                dup.limit(dup.capacity());
```
Scottmitch: limit is already equal to capacity. Is this necessary?

Scottmitch:
reminder: update these new tests to incorporate https://github.com/netty/netty/pull/6237

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: Good point let me fix and add a test

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
normanmaurer: Actually I wanted to check that bytesConsumed() has the same amount as the capacity of the etc buffer ( which means all was consumed) so I think this is correct . @Scottmitch WDYT?

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
normanmaurer: Same as above

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
normanmaurer: Yep left-over after refactoring. Let me remove the comment

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
normanmaurer: Nope let me remove it

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
Scottmitch: I think it would be more clear to check the that src has nothing remaining to read. `capacity()` could conceivably be higher than you originally asked for.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: true!

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: Actually your suggestion not works. Let me do it like this:

```java

    private static final int MAX_ENCRYPTION_OVERHEAD_DIFF = Integer.MAX_VALUE - MAX_ENCRYPTION_OVERHEAD_LENGTH;
    
    static int calculateOutNetBufSize(int pendingBytes) {
        return min(MAX_ENCRYPTED_PACKET_LENGTH, MAX_ENCRYPTION_OVERHEAD_LENGTH
                + min(MAX_ENCRYPTION_OVERHEAD_DIFF, pendingBytes));
    }
```

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
normanmaurer: actually let me test both. This will also ensure that how we update the position is the same as the amount of consumed bytes.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
fenik17: Or alternatively without additional constant:
```java
    static int calculateOutNetBufSize(int pendingBytes) {
        pendingBytes += MAX_ENCRYPTION_OVERHEAD_LENGTH;
        if (pendingBytes < MAX_ENCRYPTED_PACKET_LENGTH && pendingBytes > 0) {
            return pendingBytes;
        }
        return MAX_ENCRYPTED_PACKET_LENGTH;
    }
```

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
```diff
@@ -115,6 +119,146 @@ public void testWrapBuffersNoWritePendingError() throws Exception {
...
+                .sslProvider(sslClientProvider())
+                .build();
+        SelfSignedCertificate ssc = new SelfSignedCertificate();
+        serverSslCtx = SslContextBuilder.forServer(ssc.certificate(), ssc.privateKey())
+                .sslProvider(sslServerProvider())
+                .build();
+        SSLEngine clientEngine = null;
+        SSLEngine serverEngine = null;
+        try {
+            clientEngine = clientSslCtx.newEngine(UnpooledByteBufAllocator.DEFAULT);
```
Scottmitch: rebase reminder: update the allocators here.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -614,26 +618,45 @@ public final SSLEngineResult wrap(
...
+                final ByteBuffer src = srcs[i];
+                if (src == null) {
+                    throw new IllegalArgumentException("srcs[" + i + "] is null");
+                }
+                if (srcsLen == MAX_PLAINTEXT_LENGTH) {
+                    continue;
+                }
+
+                srcsLen += src.remaining();
+                if (srcsLen > MAX_PLAINTEXT_LENGTH) {
```
Scottmitch: we could overflow here.

Scottmitch: immagine sizes of srcs are: {2, `Integer.MAX_VALUE`}. Then we will overflow `srcsLen`.

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
Scottmitch: Do we need to update these unit tests as a result of #6237?

normanmaurer: already done ;)

Scottmitch: feel free to ignore if not ... rushing out the door at the moment :)

normanmaurer:
Cherry-picked into 4.1 (ac5f701) and 4.0 (ed61ec0f83699f85ac39a2e7368cfbdbf070715c)

PR #6254, opened 2017-01-19
Remove synchronized (ReferenceCountedOpenSslContext.class) blocks

normanmaurer:
Motivation:

We had some useless synchronized (ReferenceCountedOpenSslContext.class) blocks in our code which could slow down concurrent collecting and creating of ReferenceCountedOpenSslContext instances. Beside this we missed a few guards.

Modifications:

Use ReadWriteLock to correctly guard. A ReadWriteLock was choosen as SSL.newSSL(...) will be called from multiple threads all the time so using synchronized would be worse and there would be no way for the JIT to optimize it away

Result:

Faster concurrent creating and collecting of ReferenceCountedOpenSslContext instances and correctly guard in all cases.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/OpenSslServerSessionContext.java
+++ b/handler/src/main/java/io/netty/handler/ssl/OpenSslServerSessionContext.java
@@ -18,6 +18,8 @@ package io.netty.handler.ssl;
 import io.netty.internal.tcnative.SSL;
 import io.netty.internal.tcnative.SSLContext;
 
+import java.util.concurrent.locks.Lock;
+
 
 /**
  * {@link OpenSslSessionContext} implementation which offers extra methods which are only useful for the server-side.
@@ -32,12 +34,24 @@ public final class OpenSslServerSessionContext extends OpenSslSessionContext {
         if (seconds < 0) {
             throw new IllegalArgumentException();
         }
-        SSLContext.setSessionCacheTimeout(context.ctx, seconds);
+        Lock writerLock = context.ctxLock.writeLock();
+        writerLock.lock();
+        try {
+            SSLContext.setSessionCacheTimeout(context.ctx, seconds);
+        } finally {
+            writerLock.unlock();
+        }
     }
 
     @Override
     public int getSessionTimeout() {
-        return (int) SSLContext.getSessionCacheTimeout(context.ctx);
+        Lock readerLock = context.ctxLock.readLock();
+        readerLock.lock();
+        try {
+            return (int) SSLContext.getSessionCacheTimeout(context.ctx);
+        } finally {
+            readerLock.unlock();
+        }
     }
 
     @Override
@@ -45,23 +59,48 @@ public final class OpenSslServerSessionContext extends OpenSslSessionContext {
         if (size < 0) {
             throw new IllegalArgumentException();
         }
-        SSLContext.setSessionCacheSize(context.ctx, size);
+        Lock writerLock = context.ctxLock.writeLock();
+        writerLock.lock();
+        try {
+            SSLContext.setSessionCacheSize(context.ctx, size);
+        } finally {
+            writerLock.unlock();
+        }
     }
 
     @Override
     public int getSessionCacheSize() {
-        return (int) SSLContext.getSessionCacheSize(context.ctx);
+        Lock readerLock = context.ctxLock.readLock();
+        readerLock.lock();
+        try {
+            return (int) SSLContext.getSessionCacheSize(context.ctx);
+        } finally {
+            readerLock.unlock();
+        }
     }
 
     @Override
     public void setSessionCacheEnabled(boolean enabled) {
         long mode = enabled ? SSL.SSL_SESS_CACHE_SERVER : SSL.SSL_SESS_CACHE_OFF;
-        SSLContext.setSessionCacheMode(context.ctx, mode);
+
+        Lock writerLock = context.ctxLock.writeLock();
+        writerLock.lock();
+        try {
+            SSLContext.setSessionCacheMode(context.ctx, mode);
+        } finally {
+            writerLock.unlock();
+        }
     }
 
     @Override
     public boolean isSessionCacheEnabled() {
-        return SSLContext.getSessionCacheMode(context.ctx) == SSL.SSL_SESS_CACHE_SERVER;
+        Lock readerLock = context.ctxLock.readLock();
+        readerLock.lock();
+        try {
+            return SSLContext.getSessionCacheMode(context.ctx) == SSL.SSL_SESS_CACHE_SERVER;
+        } finally {
+            readerLock.unlock();
+        }
     }
 
     /**
@@ -74,6 +113,12 @@ public final class OpenSslServerSessionContext extends OpenSslSessionContext {
      * @return {@code true} if success, {@code false} otherwise.
      */
     public boolean setSessionIdContext(byte[] sidCtx) {
-        return SSLContext.setSessionIdContext(context.ctx, sidCtx);
+        Lock writerLock = context.ctxLock.writeLock();
+        writerLock.lock();
+        try {
+            return SSLContext.setSessionIdContext(context.ctx, sidCtx);
+        } finally {
+            writerLock.unlock();
+        }
     }
 }
--- a/handler/src/main/java/io/netty/handler/ssl/OpenSslSessionContext.java
+++ b/handler/src/main/java/io/netty/handler/ssl/OpenSslSessionContext.java
@@ -25,6 +25,7 @@ import javax.net.ssl.SSLSessionContext;
 import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.NoSuchElementException;
+import java.util.concurrent.locks.Lock;
 
 /**
  * OpenSSL specific {@link SSLSessionContext} implementation.
@@ -76,8 +77,14 @@ public abstract class OpenSslSessionContext implements SSLSessionContext {
             a += SessionTicketKey.AES_KEY_SIZE;
             tickets[i] = new SessionTicketKey(name, hmacKey, aesKey);
         }
-        SSLContext.clearOptions(context.ctx, SSL.SSL_OP_NO_TICKET);
-        SSLContext.setSessionTicketKeys(context.ctx, tickets);
+        Lock writerLock = context.ctxLock.writeLock();
+        writerLock.lock();
+        try {
+            SSLContext.clearOptions(context.ctx, SSL.SSL_OP_NO_TICKET);
+            SSLContext.setSessionTicketKeys(context.ctx, tickets);
+        } finally {
+            writerLock.unlock();
+        }
     }
 
     /**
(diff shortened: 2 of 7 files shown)


normanmaurer:
@rkapsi maybe interesting to you as well.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -461,7 +464,7 @@ public final long sslCtxPointer() {
     //            get access to an OpenSslSessionContext after this method was called to prevent the user from
     //            producing a segfault.
     final void destroy() {
-        synchronized (ReferenceCountedOpenSslContext.class) {
+        if (DESTROYED_UPDATER.compareAndSet(this, 0, 1)) {
```
Scottmitch: does this even need to be atomic? shouldn't the `AbstractReferenceCounted` guarantee this is only called a single time? It also doesn't look like we use `destroyed` or `DESTROYED_UPDATER` anywhere else.

Scottmitch: There are methods like `OpenSslServerSessionContext#getSessionCacheSize()` which use `ctx` that may cause issues (e.g. core dump). I know there is a "warning" message above this method but I think we should consider one of the the following approaches:

1. use synchronized to protect access to `ctx`
Just like we protect access to `ssl` in the `OpenSslEngine` we can protect `ctx` so we avoid calling native methods when `ctx == NULL`.
2. ensure native methods all check for `ctx == NULL`
If native methods are "safe" (don't core dump ... throwing an exception is fine) then maybe there is no risk? As long as we set `ctx = 0` before we call `SSLContext.free(ctx)` this should be OK.

If we can get away with approach 2 we can avoid synchronized. Can we use the same approach in `OpenSslEngine` to avoid synchronization?

nmittler: +1 to 2 :)

Scottmitch: @normanmaurer - WDYT?

normanmaurer: @Scottmitch 2) sounds good to me. Let me ensure we do in all our methods in netty-tcnative.

Scottmitch: We should also be careful with this approach though ... I'm not sure if this will happen but we should also be careful of the following:

Thread1:
long ctx = this.ctx; // volatile read

Thread2:
long ctx = this.ctx; // volatile read
this.ctx = 0; // volatile write
free(ctx);

Thread1:
someTcNativeMethod(ctx); // the memory pointed to by ctx has been freed!

Even if we do another volatile read at the end here ... it is possible a thread will use `ctx` in JNI while another thread destroys the memory. So we may need to leverage atomic reference counting (if offered by OpenSSL) on the ctx object ... or use synchronized.

normanmaurer:
@nmittler @Scottmitch PTAL... also added the missing guards

handler/src/main/java/io/netty/handler/ssl/OpenSslServerSessionContext.java
```diff
@@ -32,36 +34,73 @@ public void setSessionTimeout(int seconds) {
         if (seconds < 0) {
             throw new IllegalArgumentException();
         }
-        SSLContext.setSessionCacheTimeout(context.ctx, seconds);
+        Lock writerLock = context.ctxLock.writeLock();
+        try {
+            writerLock.lock();
```
Scottmitch: move the lock statement outside the `try`. If `lock` fails we shouldn't try to `unlock`.

This is a general comment when the locks are used.

normanmaurer: doh! You are right :(

Scottmitch:
discussed with @normanmaurer offline. Things to investigate:

- Do the methods which look like accessors in OpenSSL modify any state (and thus need the writer lock)
- Can we just use the reader/write to protect setting freeing the `ctx` object? This is really what the lock is intended to protect against. We should make sure we use atomics in the C code where necessary.
- Can the JIT optimize away reader/writer locks as easily as `synchronized` statements? If most users will be accessing this method from the EventLoop thread maybe we should just stick with `synchronized` for now.
- We should use a similar strategy in `OpenSslEngine`

normanmaurer:
@Scottmitch PTAL again

handler/src/main/java/io/netty/handler/ssl/OpenSslServerSessionContext.java
```diff
@@ -32,36 +31,48 @@ public void setSessionTimeout(int seconds) {
         if (seconds < 0) {
             throw new IllegalArgumentException();
         }
-        SSLContext.setSessionCacheTimeout(context.ctx, seconds);
+        synchronized (context) {
+            SSLContext.setSessionCacheTimeout(context.ctx, seconds);
```
Scottmitch: `setSessionCacheTimeout` doesn't check for `NULL` and may core dump. We should have a consistent approach for checking for `NULL` between here and the `OpenSslEngine`. If we check for null in java we may avoid crossing the JNI boundary ... but then maybe we should grab the `NULL` value from JNI instead of just checking 0 (as a followup PR).

normanmaurer: @Scottmitch I think we should do it in JNI . Let me do a followup there, ok ?

Scottmitch: @normanmaurer - followup is fine in tcnative ... consider having the `free` methods (for context and ssl) return something we can assign to the long pointer in java (e.g. NULL) ... so we don't have to hard code `0` in java.

```c
TCN_IMPLEMENT_CALL(jint, SSLContext, free)(TCN_STDARGS, jlong ctx)
{
    tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
    UNREFERENCED_STDARGS;
    TCN_ASSERT(ctx != 0);
    /* Run and destroy the cleanup callback */
    int result = apr_pool_cleanup_run(c->pool, c, ssl_context_cleanup);
    apr_pool_destroy(c->pool);

    // MODIFIED
    if (result != <expected value) {
       // throw exception?
    }
    return NULL;
    // MODIFIED
}
```

```java
ctx = SSLContext.free(ctx);
```

Scottmitch:
discussed a few more spots that needs synchronization offline ... ping me when resolved. Also please add a comment on the `ReferenceCountedOpenSslContext#ctx` member variable that it is protected with synchronization on `this`.

normanmaurer:
After talking more with @Scottmitch we realised that it is the best to stick with the `Lock` and also add "null guards" to tcnative as done here: https://github.com/netty/netty-tcnative/pull/287

@trustin @Scottmitch @nmittler PTAL again

handler/src/main/java/io/netty/handler/ssl/OpenSslSessionContext.java
```diff
@@ -76,21 +77,33 @@ public void setTicketKeys(byte[] keys) {
             a += SessionTicketKey.AES_KEY_SIZE;
             tickets[i] = new SessionTicketKey(name, hmacKey, aesKey);
         }
-        SSLContext.clearOptions(context.ctx, SSL.SSL_OP_NO_TICKET);
-        SSLContext.setSessionTicketKeys(context.ctx, tickets);
+        Lock lock = context.ctxLock.writeLock();
```
Scottmitch: `lock` -> `writerLock` to be more consistent with other uses

handler/src/main/java/io/netty/handler/ssl/OpenSslSessionContext.java
```diff
@@ -76,21 +77,33 @@ public void setTicketKeys(byte[] keys) {
...
      */
     public void setTicketKeys(OpenSslSessionTicketKey... keys) {
         ObjectUtil.checkNotNull(keys, "keys");
-        SSLContext.clearOptions(context.ctx, SSL.SSL_OP_NO_TICKET);
         SessionTicketKey[] ticketKeys = new SessionTicketKey[keys.length];
         for (int i = 0; i < ticketKeys.length; i++) {
             ticketKeys[i] = keys[i].key;
         }
-        SSLContext.setSessionTicketKeys(context.ctx, ticketKeys);
+        Lock lock = context.ctxLock.writeLock();
```
Scottmitch: `lock` -> `writerLock` to be more consistent with other uses

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -432,7 +435,13 @@ public final SSLEngine newEngine(ByteBufAllocator alloc) {
      */
     @Deprecated
     public final long context() {
-        return ctx;
+        Lock lock = ctxLock.readLock();
```
Scottmitch: `lock` -> `readerLock` to be more consistent with other uses

Scottmitch: also consider adding a comment to the javadocs that care must be taken with this pointer as it may be freed.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -491,18 +500,29 @@ public final void setTicketKeys(byte[] keys) {
 
     /**
      * Returns the pointer to the {@code SSL_CTX} object for this {@link ReferenceCountedOpenSslContext}.
-     * Be aware that it is freed as soon as the {@link #release()}  method is called.
+     * Be aware that it is freed as soon as the {@link #release()} method is called.
      * At this point {@code 0} will be returned.
+     *
+     * @deprecated Dont use.
```
Scottmitch: `Done use.` -> `This method is considered unsafe. Other methods on this class should be used and feature requests should be made if capabilities are missing.`

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -491,18 +500,29 @@ public final void setTicketKeys(byte[] keys) {
...
      */
+    @Deprecated
     public final long sslCtxPointer() {
-        return ctx;
+        Lock lock = ctxLock.readLock();
```
Scottmitch: `lock` -> `readerLock` to be more consistent with other uses

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -491,18 +500,29 @@ public final void setTicketKeys(byte[] keys) {
...
+        }
     }
 
     // IMPORTANT: This method must only be called from either the constructor or the finalizer as a user MUST never
     //            get access to an OpenSslSessionContext after this method was called to prevent the user from
     //            producing a segfault.
-    final void destroy() {
-        synchronized (ReferenceCountedOpenSslContext.class) {
+    private void destroy() {
+        Lock lock = ctxLock.writeLock();
```
Scottmitch: `lock` -> `writerLock` to be more consistent with other uses

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -234,7 +237,14 @@ protected void deallocate() {
         localCerts = context.keyCertChain;
         keyMaterialManager = context.keyMaterialManager();
         enableOcsp = context.enableOcsp;
-        ssl = SSL.newSSL(context.ctx, !context.isClient());
+
+        Lock lock = context.ctxLock.readLock();
```
Scottmitch: `lock` -> `readerLock` to be more consistent with other uses

normanmaurer:
Cherry-picked into 4.1 (e597756a56) and 4.0 (a05955d337)

PR #6257, opened 2017-01-20
Allow to configure SslHandler to wait for close_notify response before closing the Channel and fix racy flush close_notify timeout scheduling.

normanmaurer:
Motivation:

    SslHandler closed the channel as soon as it was able to write out the close_notify message. This may not be what the user want as it may make sense to only close it after the actual response to the close_notify was received in order to guarantee a clean-shutdown of the connection in all cases.

    Beside this closeNotifyFlushTimeoutMillis is volatile so may change between two reads. We need to cache it in a local variable to ensure it not change int between. Beside this we also need to check if the flush promise was complete the schedule timeout as this may happened but we were not able to cancel the timeout yet. Otherwise we will produce an missleading log message.

    Modifications:

    - Add new setter / getter to SslHandler which allows to specify the behavior (old behavior is preserved as default)
    - Added unit test.
    - Cache volatile closeNotifyTimeoutMillis.
    - Correctly check if flush promise was complete before we try to forcibly close the Channel and log a warning.
    - Add missing javadocs.

    Result:

    More clean shutdown of connection possible when using SSL and fix racy way of schedule close_notify flush timeouts and javadocs.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -308,7 +308,8 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
     private boolean firedChannelRead;
 
     private volatile long handshakeTimeoutMillis = 10000;
-    private volatile long closeNotifyTimeoutMillis = 3000;
+    private volatile long closeNotifyFlushTimeoutMillis = 3000;
+    private volatile long closeNotifyReadTimeoutMillis;
 
     /**
      * Creates a new instance.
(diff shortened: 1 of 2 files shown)


handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -378,24 +379,73 @@ public void setHandshakeTimeoutMillis(long handshakeTimeoutMillis) {
...
+
+    /**
+     * @deprecated use {@link #setCloseNotifyFlushTimeoutMillis(long)}
+     */
+    @Deprecated
+    public void setCloseNotifyTimeoutMillis(long closeNotifyFlushTimeoutMillis) {
+        setCloseNotifyFlushTimeoutMillis(closeNotifyFlushTimeoutMillis);
+    }
+
+    public long getCloseNotifyFlushTimeoutMillis() {
```
Scottmitch: make these new methods final to start ... or do you think they need to be overridable?

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1540,9 +1591,43 @@ public void operationComplete(ChannelFuture f)
...
+                    addCloseListener(ctx.close(ctx.newPromise()), promise);
+                } else {
+                    final ScheduledFuture<?> closeNotifyReadTimeoutFuture;
+
+                    if (!sslClosePromise.isDone()) {
+                        closeNotifyReadTimeoutFuture = ctx.executor().schedule(new Runnable() {
+                            @Override
+                            public void run() {
+                                if (!sslClosePromise.isDone()) {
+                                    logger.warn("{} Not received close_notify response in the configured timeout" +
```
Scottmitch: `warn` -> `info`. The peer may be dead or misbehaving and that may not be a warning.

Scottmitch: `{} Not received ...` -> `{} did not receive close_notify in {}ms; force-closing the connection.`

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
normanmaurer: nope final is ok :)

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
trustin: Why not even DEBUG? I think this happens quite a lot in reality.

normanmaurer: @trustin Yeah debug is fine I guess..

normanmaurer:
@Scottmitch PTAL again buddy

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -378,24 +379,86 @@ public void setHandshakeTimeoutMillis(long handshakeTimeoutMillis) {
...
+        return closeNotifyFlushTimeoutMillis;
+    }
+
+    /**
+     * Sets the timeout for flushing the close_notify that was triggered by closing the
+     * {@link Channel}. If the close_notify was not flushed in the given timeout the {@link Channel} will be closed
+     * forcibily.
+     */
+    public final void setCloseNotifyFlushTimeout(long closeNotifyFlushTimeoutMillis, TimeUnit unit) {
+        setCloseNotifyFlushTimeoutMillis(unit.toMillis(closeNotifyFlushTimeoutMillis));
```
johnou: ObjectUtil#checkNotNull(unit, "unit")?

Scottmitch: we will get a NPE anyways ... more efficient to let the JVM enforce this if the condition is not expected regularly.

johnou: right but at least then you know that it was unit that was null instead of checking line numbers / source and hoping that there isn't multiple objects on the one line that could have caused the NPE.

Scottmitch: I agree we may be able to provide a bit more context by manually throwing the exception but I think you ultimately have to check source code when you get this type of exception. There is only one object referenced on this line, and 1 object parameter passed into this method, so there shouldn't be any ambiguity.

normanmaurer:
@Scottmitch @trustin I will merge this later today... thanks!

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
johnou: just seemed a little strange to me to have the null check and then remove it.

normanmaurer:
Cherry-picked into 4.1 (640ef615be1bbf48f8506151518c0c55065e7f99) and 4.0 (95f47a44ee33a761d13bfb5b06ecea8829b4d05d)

Scottmitch:
(branch deleted)

normanmaurer:
Thx

PR #6258, opened 2017-01-20
DnsNameResolver.resolve*(...) never notifies the Future when empty ho…

normanmaurer:
…stname is used.

Motivation:

When an empty hostname is used in DnsNameResolver.resolve*(...) it will never notify the future / promise. The root cause is that we not correctly guard against errors of IDN.toASCII(...) which will throw an IllegalArgumentException when it can not parse its input. That said we should also handle an empty hostname the same way as the JDK does and just use "localhost" when this happens.

Modifications:

- If the try to resolve an empty hostname we use localhost
- Correctly guard against errors raised by IDN.toASCII(...) so we will always noify the future / promise
- Add unit test.

Result:

DnsNameResolver.resolve*(...) will always notify the future.

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
@@ -411,8 +411,11 @@ public class DnsNameResolver extends InetNameResolver {
      */
     public final Future<InetAddress> resolve(String inetHost, Iterable<DnsRecord> additionals,
                                              Promise<InetAddress> promise) {
-        checkNotNull(inetHost, "inetHost");
         checkNotNull(promise, "promise");
+        if (inetHost == null || inetHost.isEmpty()) {
+            // If an empty hostname is used we should use "localhost", just like InetAddress.getByName(...) does.
+            return promise.setSuccess(loopbackAddress());
+        }
         DnsRecord[] additionalsArray = toArray(additionals, true);
         try {
             doResolve(inetHost, additionalsArray, promise, resolveCache);
@@ -445,8 +448,13 @@ public class DnsNameResolver extends InetNameResolver {
      */
     public final Future<List<InetAddress>> resolveAll(String inetHost, Iterable<DnsRecord> additionals,
                                                 Promise<List<InetAddress>> promise) {
-        checkNotNull(inetHost, "inetHost");
         checkNotNull(promise, "promise");
+
+        if (inetHost == null || inetHost.isEmpty()) {
+            // If an empty hostname is used we should use "localhost", just like InetAddress.getAllByName(...) does.
+            return promise.setSuccess(Collections.singletonList(loopbackAddress()));
+        }
+
         DnsRecord[] additionalsArray = toArray(additionals, true);
         try {
             doResolveAll(inetHost, additionalsArray, promise, resolveCache);
@@ -490,6 +498,12 @@ public class DnsNameResolver extends InetNameResolver {
         if (validateType && record instanceof DnsRawRecord) {
             throw new IllegalArgumentException("DnsRawRecord implementations not allowed: " + record);
         }
+    }
+
+    @Override
+    protected final InetAddress loopbackAddress() {
+        return preferredAddressType() == InternetProtocolFamily.IPv4 ?
+                    NetUtil.LOCALHOST4 : NetUtil.LOCALHOST6;
     }
 
     /**
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
@@ -146,7 +146,9 @@ abstract class DnsNameResolverContext<T> {
     private void internalResolve(Promise<T> promise) {
         InetSocketAddress nameServerAddrToTry = nameServerAddrs.next();
         for (DnsRecordType type: parent.resolveRecordTypes()) {
-            query(nameServerAddrToTry, new DefaultDnsQuestion(hostname, type), promise);
+            if (!query(hostname, type, nameServerAddrToTry, promise)) {
+                return;
+            }
         }
     }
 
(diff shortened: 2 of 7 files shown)


resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
```diff
@@ -510,11 +512,26 @@ private void followCname(InetSocketAddress nameServerAddr, String name, String c
 
         final InetSocketAddress nextAddr = nameServerAddrs.next();
         if (parent.isCnameFollowARecords()) {
-            query(nextAddr, new DefaultDnsQuestion(cname, DnsRecordType.A), promise);
+            if (!query(hostname, DnsRecordType.A, nextAddr, promise)) {
```
netkins: MAJOR Merge this if statement with the enclosing one. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1066)

normanmaurer: I actually think its easier to read the way it is... So I would love to just leave it as it is. @Scottmitch thoughts ?

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -507,6 +512,12 @@ protected void doResolve(String inetHost,
             return;
         }
 
+        if (inetHost.isEmpty()) {
```
Scottmitch: the JDK also resolves `null` to localhost too

Scottmitch: we should move this check above `createByteArrayFromIpAddressString` which may throw a NPE if `null` is passed in.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
Scottmitch: I don't think it impacts readability much either way ... less conditionals & indentation if they are combined. minor issue though your call.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
normanmaurer: @Scottmitch yes it does but not our DefaultNameResolver where we disallow null: https://github.com/netty/netty/blob/4.1/resolver/src/main/java/io/netty/resolver/SimpleNameResolver.java#L58

We could remove this restriction tho... WDYT ?

Scottmitch: I guess it is inline with this PR which is making behavior more consistent with JDK's resolution.

normanmaurer: @Scottmitch maybe we even should just move the empty and null handling to `SimpleNameResolver` ?

Scottmitch: @normanmaurer - sgtm. I'm not sure if @trustin would have any opinion on this though.

normanmaurer: @trustin WDYT ?

trustin: Not to `SimpleNameResolver` but to `InetNameResolver`? `SimpleNameResolver` can be used for non-INET scenarios.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
trustin: I think it reads better this way and one less indentation level is a bonus.

```java
if (parent.isCnameFollowARecords() &&
    !query(hostname, DnsRecordType.A, nextAddr, promise)) {
    ...
}
```

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
vietj: +1 for being consistent with JDK's resolution

normanmaurer:
@Scottmitch PTAL again... This was the only way I could find without introduce some sort of API breaking in non `@UnstableApi` marked stuff.

resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
```diff
@@ -496,6 +497,47 @@ public void testResolveIp() {
...
+
+    @Test
+    public void testResolveEmptyIpv6() {
+        testResolveEmpty(InternetProtocolFamily.IPv6, NetUtil.LOCALHOST6);
+    }
+
+    private static void testResolveEmpty(InternetProtocolFamily family, InetAddress expectedAddr) {
+        DnsNameResolver resolver = newResolver(family).build();
+        try {
+            InetAddress address = resolver.resolve("").syncUninterruptibly().getNow();
```
Scottmitch: also test with `null`?

resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
```diff
@@ -496,6 +497,47 @@ public void testResolveIp() {
...
+
+    @Test
+    public void testResolveAllEmptyIpv6() {
+        testResolveAllEmpty(InternetProtocolFamily.IPv6, NetUtil.LOCALHOST6);
+    }
+
+    private static void testResolveAllEmpty(InternetProtocolFamily family, InetAddress expectedAddr) {
+        DnsNameResolver resolver = newResolver(family).build();
+        try {
+            List<InetAddress> addresses = resolver.resolveAll("").syncUninterruptibly().getNow();
```
Scottmitch: also test with `null`?

resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
normanmaurer: yep will do

resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
normanmaurer: yep will do

normanmaurer:
Cherry-picked into 4.1 (a416b79d865fd6ce11ecce74d53ce2dfd8be007f)

Issue #6260, opened 2017-01-20
javax.net.ssl.SSLException: error:140D00CF:SSL routines:SSL_write:protocol is shutdown


normanmaurer:
```
 javax.net.ssl.SSLException: error:140D00CF:SSL routines:SSL_write:protocol is shutdown
	at io.netty.handler.ssl.ReferenceCountedOpenSslEngine.shutdownWithError(ReferenceCountedOpenSslEngine.java:745) ~[netty-all-4.1.8.Final-SNAPSHOT.jar:4.1.8.Final-SNAPSHOT]
	at io.netty.handler.ssl.ReferenceCountedOpenSslEngine.shutdownWithError(ReferenceCountedOpenSslEngine.java:734) ~[netty-all-4.1.8.Final-SNAPSHOT.jar:4.1.8.Final-SNAPSHOT]
	at io.netty.handler.ssl.ReferenceCountedOpenSslEngine.wrap(ReferenceCountedOpenSslEngine.java:711) ~[netty-all-4.1.8.Final-SNAPSHOT.jar:4.1.8.Final-SNAPSHOT]
	at javax.net.ssl.SSLEngine.wrap(SSLEngine.java:509) ~[?:1.8.0_102]
	at io.netty.handler.ssl.SslHandler.wrap(SslHandler.java:764) ~[netty-all-4.1.8.Final-SNAPSHOT.jar:4.1.8.Final-SNAPSHOT]
	at io.netty.handler.ssl.SslHandler.wrap(SslHandler.java:596) ~[netty-all-4.1.8.Final-SNAPSHOT.jar:4.1.8.Final-SNAPSHOT]
	at io.netty.handler.ssl.SslHandler.wrapAndFlush(SslHandler.java:568) ~[netty-all-4.1.8.Final-SNAPSHOT.jar:4.1.8.Final-SNAPSHOT]
	at io.netty.handler.ssl.SslHandler.flush(SslHandler.java:549) ~[netty-all-4.1.8.Final-SNAPSHOT.jar:4.1.8.Final-SNAPSHOT]
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush0(AbstractChannelHandlerContext.java:777) [netty-all-4.1.8.Final-SNAPSHOT.jar:4.1.8.Final-SNAPSHOT]
[...]

PR #6262, opened 2017-01-20
Ensure calling ReferenceCountedOpenSslEngine.wrap(...) after closeOut…

normanmaurer:
…bound() was called will not throw an SSLException

Motivation:

PR [#6238] added guards to be able to call wrap(...) / unwrap(...) after the engine was shutdown. Unfortunally one case was missed which is when closeOutbound() was called and produced some data while closeInbound() was not called yet.

Modifications:

Correctly guard against SSLException when closeOutbound() was called, produced data and someone calls wrap(...) after it.

Result:

No more SSLException. Fixes [#6260].

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -62,7 +62,6 @@ import static io.netty.handler.ssl.OpenSsl.memoryAddress;
 import static io.netty.util.internal.EmptyArrays.EMPTY_CERTIFICATES;
 import static io.netty.util.internal.EmptyArrays.EMPTY_JAVAX_X509_CERTIFICATES;
 import static io.netty.util.internal.ObjectUtil.checkNotNull;
-import static java.lang.Math.max;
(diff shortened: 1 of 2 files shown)


normanmaurer:
@Scottmitch @rkapsi added an unit-test FTW 🎉

rkapsi:
Deploying... Should get some results in a few minutes...

normanmaurer:
@Scottmitch double is better then single approved 👍

rkapsi:
Make it triple, exception is gone.

normanmaurer:
@rkapsi 😍 thanks for the quick turn-around

Scottmitch:
that was quick @rkapsi! I guess the exception reproduces quickly when the issue is present?

rkapsi:
Yea, a couple of times per minute (before the first fix it was every other second). All I needed to do is to deploy one server with the new artifact and watch the logs. It's something we can thankfully do very quickly (and Netty is thanks to the Unit Tests quite low risk).

Scottmitch:
sgtm @rkapsi ... ship it @normanmaurer

normanmaurer:
Cherry-picked into 4.1 (e9fa40d) and 4.0 (f1130e4)

PR #6263, opened 2017-01-20
Update hpack Decoder CTOR to allow for overflow in maxHeaderListSize

exell-christopher:
Motivation:
When testing the fixes from #6209 locally I noticed that the first request into the decoder still causes a GOAWAY even when it's below what I would expect to be in maxHeaderListSizeGoAway.

Turns out both of the limits are set to exactly the same thing in the decoder CTOR instead of using the same math that is used when the local settings are applied.

Modification:

Moved calculate calculateMaxHeaderListSizeGoAway into Http2CodecUtil and used that in the CTOR for the hpack Decoder.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
@@ -149,8 +149,7 @@ public class DefaultHttp2ConnectionDecoder implements Http2ConnectionDecoder {
      * @return the threshold in bytes which should trigger a {@code GO_AWAY} if a set of headers exceeds this amount.
      */
     protected long calculateMaxHeaderListSizeGoAway(long maxHeaderListSize) {
-        // This is equivalent to `maxHeaderListSize * 1.25` but we avoid floating point multiplication.
-        return maxHeaderListSize + (maxHeaderListSize >>> 2);
+        return Http2CodecUtil.calculateMaxHeaderListSizeGoAway(maxHeaderListSize);
     }
 
     private int unconsumedBytes(Http2Stream stream) {
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecUtil.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecUtil.java
@@ -113,6 +113,18 @@ public final class Http2CodecUtil {
     public static final int DEFAULT_MAX_FRAME_SIZE = MAX_FRAME_SIZE_LOWER_BOUND;
 
     /**
+     * Calculate the threshold in bytes which should trigger a {@code GO_AWAY} if a set of headers exceeds this amount.
+     * @param maxHeaderListSize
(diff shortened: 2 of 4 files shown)


exell-christopher:
@Scottmitch Let me know if you need anything on this.

nhnFreespirit:
Looks good to me FWIW

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -141,18 +141,6 @@ public void close() {
...
     }
 
-    /**
-     * Calculate the threshold in bytes which should trigger a {@code GO_AWAY} if a set of headers exceeds this amount.
-     * @param maxHeaderListSize
-     *      <a href="https://tools.ietf.org/html/rfc7540#section-6.5.2">SETTINGS_MAX_HEADER_LIST_SIZE</a> for the local
-     *      endpoint.
-     * @return the threshold in bytes which should trigger a {@code GO_AWAY} if a set of headers exceeds this amount.
-     */
-    protected long calculateMaxHeaderListSizeGoAway(long maxHeaderListSize) {
```
Scottmitch: the reason why this method is here is so the user can override and provide their own algorithm. It still seems useful for the user to have some ability to override our default algorithm.

One of the reasons I didn't consolidate the two was because this method and the CTOR for Decoder are disjoint. I debated using a `Function` but that seemed like overkill and we are java6 anyways. I'm open to better suggestions.

Scottmitch:
@exell-christopher - Also please sign the [ICLA](https://docs.google.com/forms/d/e/1FAIpQLSd7Bzje39G__THDJLRgKpQZ4gODNE26x_hZW3ofQOkgL6RGCA/viewform?formkey=dHBjc1YzdWhsZERUQnhlSklsbG1KT1E6MQ)

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
exell-christopher: Doing a quick rework, to keep the ability to override, but putting the actual logic in Http2CodecUtil

exell-christopher:
@Scottmitch 
Change committed. I'll squash the commits once you think this is ready to go.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -423,7 +422,8 @@ private void applyLocalSettings(Http2Settings settings) throws Http2Exception {
 
             Long maxHeaderListSize = settings.maxHeaderListSize();
             if (maxHeaderListSize != null) {
-                headerConfig.maxHeaderListSize(maxHeaderListSize, calculateMaxHeaderListSizeGoAway(maxHeaderListSize));
+                headerConfig.maxHeaderListSize(maxHeaderListSize,
```
Scottmitch: nit: revert this change

exell-christopher:
Nit fixed, commits squashed.

normanmaurer:
@Scottmitch @exell-christopher works for Twitter so its fine as Twitter has signed the CCLA

Scottmitch:
4.1 (9077269)

Thanks @exell-christopher !

PR #6267, opened 2017-01-24
More readable code about shutdown and shutdownGracefully

### LINKED ISSUE #6264: Code suggestion about shutdownGracefully
Here's the current [code](https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java#L557), and the following is a code snippet:

```java
boolean inEventLoop = inEventLoop();
boolean wakeup;
int oldState;
for (;;) {
    if (isShuttingDown()) {
        return terminationFuture();
    }
    int newState;
    wakeup = true;
    oldState = STATE_UPDATER.get(this);
    if (inEventLoop) {
        newState = ST_SHUTTING_DOWN;
    } else {
        switch (oldState) {
            case ST_NOT_STARTED:
            case ST_STARTED:
                newState = ST_SHUTTING_DOWN;
                break;
            default:
                newState = oldState;
                wakeup = false;
        }
    }
    if (STATE_UPDATER.compareAndSet(this, oldState, newState)) {
        break;
    }
}
gracefulShutdownQuietPeriod = unit.toNanos(quietPeriod);
gracefulShutdownTimeout = unit.toNanos(timeout);
```

Would it be better to code like this?

```java
int oldState;
for (;;) {
    oldState = STATE_UPDATER.get(this);
    if (oldState >= ST_SHUTTING_DOWN) {
        return terminationFuture();
    }

    if (STATE_UPDATER.compareAndSet(this, oldState, ST_SHUTTING_DOWN)) {
        break;
    }
}
gracefulShutdownQuietPeriod = unit.toNanos(quietPeriod);
gracefulShutdownTimeout = unit.toNanos(timeout);
```

wangyuntao:
Motivation:

To make the shutdown related code more readable

Result:

Fixes  #6264

### DIFF
--- a/common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
+++ b/common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
@@ -547,37 +547,18 @@ public abstract class SingleThreadEventExecutor extends AbstractScheduledEventEx
             throw new NullPointerException("unit");
         }
 
-        if (isShuttingDown()) {
-            return terminationFuture();
-        }
-
-        boolean inEventLoop = inEventLoop();
-        boolean wakeup;
         int oldState;
         for (;;) {
-            if (isShuttingDown()) {
-                return terminationFuture();
-            }
-            int newState;
-            wakeup = true;
             oldState = STATE_UPDATER.get(this);
-            if (inEventLoop) {
-                newState = ST_SHUTTING_DOWN;
-            } else {
-                switch (oldState) {
-                    case ST_NOT_STARTED:
-                    case ST_STARTED:
-                        newState = ST_SHUTTING_DOWN;
-                        break;
-                    default:
-                        newState = oldState;
-                        wakeup = false;
-                }
+            if (oldState >= ST_SHUTTING_DOWN) {
+              return terminationFuture();
             }
-            if (STATE_UPDATER.compareAndSet(this, oldState, newState)) {
+
+            if (STATE_UPDATER.compareAndSet(this, oldState, ST_SHUTTING_DOWN)) {
                 break;
             }
         }
+
         gracefulShutdownQuietPeriod = unit.toNanos(quietPeriod);
         gracefulShutdownTimeout = unit.toNanos(timeout);
 
@@ -585,10 +566,7 @@ public abstract class SingleThreadEventExecutor extends AbstractScheduledEventEx
             doStartThread();
         }
 
-        if (wakeup) {
-            wakeup(inEventLoop);
-        }
-
+        wakeup(inEventLoop());
         return terminationFuture();
     }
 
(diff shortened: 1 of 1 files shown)


common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
```diff
@@ -551,44 +551,23 @@ private boolean runShutdownHooks() {
             return terminationFuture();
         }
 
-        boolean inEventLoop = inEventLoop();
-        boolean wakeup;
-        int oldState;
+        startThread();
```
Scottmitch: This is a change in behavior, are you sure this is desired? Previously the state was changed to shutdown before the thread was started, now we unconditionally start the thread before attempting shutdown. Perhaps now the thread will be allowed to process jobs that it previously wouldn't have been allowed to.

wangyuntao: @Scottmitch , Please check it again.

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
```diff
@@ -600,35 +583,19 @@ private boolean runShutdownHooks() {
     @Override
     @Deprecated
     public void shutdown() {
-        if (isShutdown()) {
+        if (isShuttingDown()) {
```
Scottmitch: I think you just moved this check from line 611 ... however I wonder if the intention is that this method should be able to "force the shutdown" even if the state is `ST_SHUTTING_DOWN` (@normanmaurer - WDYT). If that is the case this method should only exit prematurely if `state >= ST_SHUTDOWN`.

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
```diff
@@ -551,44 +551,27 @@ private boolean runShutdownHooks() {
             return terminationFuture();
```
Scottmitch: (for the line above): minor optimization ... `isShuttingDown()` is called above which will use the state, and you are also getting the state below in a loop. you could get the state outside the loop, use it above, and then refresh it at the end of the loop.

```java
int oldState = STATE_UPDATER.get(this);
if (isShuttingDown(state)) {
  return terminationFuture();
}

for(;;) {
//try atomic set
  oldState = STATE_UPDATER.get(this);
}

@Override
public boolean isShuttingDown() {
  return isShuttingDown(STATE_UPDATER.get(this));
}

private static boolean isShuttingDown(int state) {
  return state >= ST_SHUTDOWN;
}
```

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
wangyuntao: Yes, maybe you are right!

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
wangyuntao: @Scottmitch , how about we just remove the shutdown check before the for loop?

wangyuntao:
@Scottmitch , PTAL, thanks!

wangyuntao:
@Scottmitch Could you please check it again?

normanmaurer:
@trustin PTAL as well

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
```diff
@@ -547,48 +547,26 @@ private boolean runShutdownHooks() {
...
-            if (inEventLoop) {
-                newState = ST_SHUTTING_DOWN;
-            } else {
-                switch (oldState) {
-                    case ST_NOT_STARTED:
-                    case ST_STARTED:
-                        newState = ST_SHUTTING_DOWN;
-                        break;
-                    default:
-                        newState = oldState;
```
Scottmitch: this is a change in behavior. the new code will always try to change the state but the old code, but the old code only changed the state in the event loop. @normanmaurer - WDYT (here is the original change [1])

[1] https://github.com/netty/netty/commit/16a85e6cca46cfdcfe07c9e76a3b935c72c5ec1d#diff-cde1b8aa1fe0c3a083412f79421bafcfR501

trustin:
Not strong on this but I guess the original one seems readable enough to me?

Scottmitch:
Agreed. IMHO this isn't too high a priority, and involves some time to review to make sure behavior is preserved. However it would be nice to have simpler code if possible.

normanmaurer:
Agree...

wangyuntao:
Okay, got it, so close it?

johnou:
If it's not broken and isn't causing performance issues I would opt for leaving it as is.

normanmaurer:
Yeah let me close this for now

PR #6275, opened 2017-01-25
Include Http 1 request in error message

carl-mastrangelo:
Motivation:

When An HTTP server is listening in plaintext mode, it doesn't have
a chance to negotiate "h2" in the tls handshake.  HTTP 1 clients
that are not expecting an HTTP2 server will accidentally a request
that isn't an upgrade, which the HTTP/2 decoder will not
understand.  The decoder treats the bytes as hex and adds them to
the error message.

These error messages are hard to understand by humans, and result
in extra, manual work to decode.

Modification:

If the first bytes of the request are not the preface, the decoder
will now see if they are an HTTP/1 request first.  If so, the error
message will include the method and path of the original request in
the error message.

In case the path is long, the decoder will check up to the first
1024 bytes to see if it matches.  This could be a DoS vector if
tons of bad requests or other garbage come in.  A future optimization
would be to treat the first few bytes as an AsciiString and not do
any Charset decoding.  ByteBuf.toCharSequence alludes to such an
optimization.

The code has been left simple for the time being.

Result:

Faster identification of errant HTTP requests.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
+++ b/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
@@ -159,6 +159,22 @@ public final class ByteBufUtil {
     }
 
     /**
+     * Returns the reader index of needle in haystack, or -1 if needle is not in haystack.
+     */
+    public static int indexOf(ByteBuf needle, ByteBuf haystack) {
+        // TODO: maybe use Boyer Moore for efficiency.
+        int attempts = haystack.readableBytes() - needle.readableBytes() + 1;
+        for (int i = 0; i < attempts; i++) {
+            if (equals(needle, needle.readerIndex(),
+                       haystack, haystack.readerIndex() + i,
+                       needle.readableBytes())) {
+                return haystack.readerIndex() + i;
+            }
+        }
+        return -1;
+    }
+
+    /**
      * Returns {@code true} if and only if the two specified buffers are
      * identical to each other for {@code length} bytes starting at {@code aStartIndex}
      * index for the {@code a} buffer and {@code bStartIndex} index for the {@code b} buffer.
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
@@ -16,6 +16,7 @@ package io.netty.handler.codec.http2;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufUtil;
+import io.netty.buffer.Unpooled;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandlerContext;
@@ -25,6 +26,7 @@ import io.netty.handler.codec.ByteToMessageDecoder;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http2.Http2Exception.CompositeStreamException;
 import io.netty.handler.codec.http2.Http2Exception.StreamException;
+import io.netty.util.CharsetUtil;
 import io.netty.util.concurrent.ScheduledFuture;
 import io.netty.util.internal.UnstableApi;
 import io.netty.util.internal.logging.InternalLogger;
@@ -68,6 +70,8 @@ public class Http2ConnectionHandler extends ByteToMessageDecoder implements Http
 
     private static final Http2Headers HEADERS_TOO_LARGE_HEADERS = ReadOnlyHttp2Headers.serverHeaders(false,
             HttpResponseStatus.REQUEST_HEADER_FIELDS_TOO_LARGE.codeAsText());
+    private static final ByteBuf HTTP_1_X_BUF = Unpooled.unreleasableBuffer(
+        Unpooled.wrappedBuffer(new byte[] {'H', 'T', 'T', 'P', '/', '1', '.'})).asReadOnly();
 
     private final Http2ConnectionDecoder decoder;
     private final Http2ConnectionEncoder encoder;
@@ -272,6 +276,13 @@ public class Http2ConnectionHandler extends ByteToMessageDecoder implements Http
             if (bytesRead == 0 || !ByteBufUtil.equals(in, in.readerIndex(),
                                                       clientPrefaceString, clientPrefaceString.readerIndex(),
                                                       bytesRead)) {
+                int maxSearch = 1024; // picked because 512 is too little, and 2048 too much
+                int http1Index =
+                    ByteBufUtil.indexOf(HTTP_1_X_BUF, in.slice(in.readerIndex(), min(in.readableBytes(), maxSearch)));
+                if (http1Index != -1) {
+                    String chunk = in.toString(in.readerIndex(), http1Index - in.readerIndex(), CharsetUtil.US_ASCII);
+                    throw connectionError(PROTOCOL_ERROR, "Unexpected HTTP/1.x request: %s", chunk);
+                }
                 String receivedBytes = hexDump(in, in.readerIndex(),
                                                min(in.readableBytes(), clientPrefaceString.readableBytes()));
                 throw connectionError(PROTOCOL_ERROR, "HTTP/2 client preface string missing or corrupt. " +
--- a/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2ConnectionHandlerTest.java
+++ b/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2ConnectionHandlerTest.java
@@ -27,6 +27,7 @@ import io.netty.channel.ChannelPromise;
 import io.netty.channel.DefaultChannelPromise;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http2.Http2CodecUtil.SimpleChannelPromiseAggregator;
+import io.netty.util.CharsetUtil;
 import io.netty.util.ReferenceCountUtil;
 import io.netty.util.concurrent.EventExecutor;
 import io.netty.util.concurrent.GenericFutureListener;
@@ -52,6 +53,7 @@ import static io.netty.handler.codec.http2.Http2Exception.connectionError;
 import static io.netty.handler.codec.http2.Http2Stream.State.CLOSED;
 import static io.netty.handler.codec.http2.Http2Stream.State.IDLE;
 import static io.netty.handler.codec.http2.Http2TestUtil.newVoidPromise;
+import static io.netty.util.CharsetUtil.US_ASCII;
 import static io.netty.util.CharsetUtil.UTF_8;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
@@ -127,6 +129,8 @@ public class Http2ConnectionHandlerTest {
     @Mock
     private Http2FrameWriter frameWriter;
 
+    private String goAwayDebugCap;
+
     @SuppressWarnings("unchecked")
     @Before
     public void setup() throws Exception {
@@ -144,6 +148,7 @@ public class Http2ConnectionHandlerTest {
             @Override
             public ChannelFuture answer(InvocationOnMock invocation) throws Throwable {
                 ByteBuf buf = invocation.getArgumentAt(3, ByteBuf.class);
+                goAwayDebugCap = buf.toString(UTF_8);
                 buf.release();
                 return future;
             }
@@ -240,6 +245,18 @@ public class Http2ConnectionHandlerTest {
         verify(frameWriter).writeGoAway(eq(ctx), eq(0), eq(PROTOCOL_ERROR.code()),
                 captor.capture(), eq(promise));
         assertEquals(0, captor.getValue().refCnt());
+    }
+
+    @Test
+    public void serverReceivingHttp1ClientPrefaceStringShouldIncludePreface() throws Exception {
+        when(connection.isServer()).thenReturn(true);
+        handler = newHandler();
+        handler.channelRead(ctx, copiedBuffer("GET /path HTTP/1.1", US_ASCII));
+        ArgumentCaptor<ByteBuf> captor = ArgumentCaptor.forClass(ByteBuf.class);
+        verify(frameWriter).writeGoAway(eq(ctx), eq(0), eq(PROTOCOL_ERROR.code()),
+            captor.capture(), eq(promise));
+        assertEquals(0, captor.getValue().refCnt());
+        assertTrue(goAwayDebugCap.contains("/path"));
     }
 
     @Test


normanmaurer:
@carl-mastrangelo please check the failed test case.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -272,6 +275,12 @@ private boolean readClientPrefaceString(ByteBuf in) throws Http2Exception {
             if (bytesRead == 0 || !ByteBufUtil.equals(in, in.readerIndex(),
                                                       clientPrefaceString, clientPrefaceString.readerIndex(),
                                                       bytesRead)) {
+                String http1Request = in.toString(0, 1024, CharsetUtil.US_ASCII);
```
normanmaurer: why 1024 ? Also is it possible that the buffer may have less readable bytes then 1024 ? If so should this be `in.toString(in.readerIndex(), Math.min(in.readableBytes(), 1024))` ?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -272,6 +275,12 @@ private boolean readClientPrefaceString(ByteBuf in) throws Http2Exception {
...
+                int http1Index;
+                if ((http1Index = http1Request.indexOf("HTTP/1.")) != -1) {
+                    throw connectionError(PROTOCOL_ERROR, "Unexpected HTTP/1 request: %s",
```
normanmaurer: `Unexpected HTTP/1.x` ?

normanmaurer: And I also wonder if we really should include these 1024 bytes string here as it may contain "sensitive" data ?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
carl-mastrangelo: Hmm, I assumed that the size was a maximum, rather than an actual length.  I picked 1024 because 512 is too short, and 2048 too long.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
carl-mastrangelo: This is careful to not include any of the header data (which would include cookies and auth tokens and such).  It should limit to the first line, "GET /secretGovernmentFiles HTTP/1.1\r\n".  The hexdump does reveal some of it too (albeit less).

carl-mastrangelo:
PTAL.  Not sure why they fail, mvn install passes locally.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -272,6 +275,13 @@ private boolean readClientPrefaceString(ByteBuf in) throws Http2Exception {
             if (bytesRead == 0 || !ByteBufUtil.equals(in, in.readerIndex(),
                                                       clientPrefaceString, clientPrefaceString.readerIndex(),
                                                       bytesRead)) {
+                String http1Request =
```
Scottmitch: I see in your issue description you elude to future optimizations however I think we can avoid the string conversion if the data is not HTTP/1.x:

```diff
diff --git a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
index dfe677d..2709c47 100644
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
@@ -16,6 +16,7 @@ package io.netty.handler.codec.http2;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufUtil;
+import io.netty.buffer.Unpooled;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandlerContext;
@@ -198,6 +199,9 @@ public class Http2ConnectionHandler extends ByteToMessageDecoder implements Http
         }
     }
 
+    private static final ByteBuf HTTP_1_X_PREFIX_BUF = Unpooled.unreleasableBuffer(Unpooled.wrappedBuffer(
+            new byte[] {'H', 'T', 'T', 'P', '/', '1', '.'})).asReadOnly();
+
     private final class PrefaceDecoder extends BaseDecoder {
         private ByteBuf clientPrefaceString;
         private boolean prefaceSent;
@@ -275,12 +279,10 @@ public class Http2ConnectionHandler extends ByteToMessageDecoder implements Http
             if (bytesRead == 0 || !ByteBufUtil.equals(in, in.readerIndex(),
                                                       clientPrefaceString, clientPrefaceString.readerIndex(),
                                                       bytesRead)) {
-                String http1Request =
-                    in.toString(in.readerIndex(), Math.min(in.readableBytes(), 1024), CharsetUtil.US_ASCII);
-                int http1Index;
-                if ((http1Index = http1Request.indexOf("HTTP/1.")) != -1) {
+                if (ByteBufUtil.equals(HTTP_1_X_PREFIX_BUF, HTTP_1_X_PREFIX_BUF.readerIndex(),
+                                   in, in.readerIndex(), HTTP_1_X_PREFIX_BUF.readableBytes())) {
                     throw connectionError(PROTOCOL_ERROR, "Unexpected HTTP/1.x request: %s",
-                                                          http1Request.substring(0, http1Index));
+                            in.toString(in.readerIndex(), Math.min(in.readableBytes(), 1024), CharsetUtil.US_ASCII));
                 }
                 String receivedBytes = hexDump(in, in.readerIndex(),
                                                min(in.readableBytes(), clientPrefaceString.readableBytes()));

```

Scottmitch: also you mention that this is careful not to include any header data ... how is that accomplished? Should we more explicitly look for the CRLF to ensure we only include the request-line? 

```diff
diff --git a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
index dfe677d..3b4dbf8 100644
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
@@ -16,6 +16,7 @@ package io.netty.handler.codec.http2;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufUtil;
+import io.netty.buffer.Unpooled;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandlerContext;
@@ -26,6 +27,7 @@ import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http2.Http2Exception.CompositeStreamException;
 import io.netty.handler.codec.http2.Http2Exception.StreamException;
 import io.netty.util.AsciiString;
+import io.netty.util.ByteProcessor;
 import io.netty.util.CharsetUtil;
 import io.netty.util.concurrent.ScheduledFuture;
 import io.netty.util.internal.UnstableApi;
@@ -198,6 +200,9 @@ public class Http2ConnectionHandler extends ByteToMessageDecoder implements Http
         }
     }
 
+    private static final ByteBuf HTTP_1_X_PREFIX_BUF = Unpooled.unreleasableBuffer(Unpooled.wrappedBuffer(
+            new byte[] {'H', 'T', 'T', 'P', '/', '1', '.'})).asReadOnly();
+
     private final class PrefaceDecoder extends BaseDecoder {
         private ByteBuf clientPrefaceString;
         private boolean prefaceSent;
@@ -275,12 +280,16 @@ public class Http2ConnectionHandler extends ByteToMessageDecoder implements Http
             if (bytesRead == 0 || !ByteBufUtil.equals(in, in.readerIndex(),
                                                       clientPrefaceString, clientPrefaceString.readerIndex(),
                                                       bytesRead)) {
-                String http1Request =
-                    in.toString(in.readerIndex(), Math.min(in.readableBytes(), 1024), CharsetUtil.US_ASCII);
-                int http1Index;
-                if ((http1Index = http1Request.indexOf("HTTP/1.")) != -1) {
+                if (ByteBufUtil.equals(HTTP_1_X_PREFIX_BUF, HTTP_1_X_PREFIX_BUF.readerIndex(),
+                                   in, in.readerIndex(), HTTP_1_X_PREFIX_BUF.readableBytes())) {
+                    final int maxLength = 1024;
+                    int endIndex = in.forEachByte(in.readerIndex(), Math.min(in.readableBytes(), maxLength),
+                                                  ByteProcessor.FIND_CRLF);
+                    if (endIndex < 0) {
+                        endIndex = in.readerIndex() + maxLength;
+                    }
                     throw connectionError(PROTOCOL_ERROR, "Unexpected HTTP/1.x request: %s",
-                                                          http1Request.substring(0, http1Index));
+                            in.toString(in.readerIndex(), endIndex - in.readerIndex(), CharsetUtil.US_ASCII));
                 }
                 String receivedBytes = hexDump(in, in.readerIndex(),
                                                min(in.readableBytes(), clientPrefaceString.readableBytes()));

```

carl-mastrangelo: I don't think your proposed change works.  The "HTTP/1.1" is the last part of the first line, not at the beginning.  ByteBuf lacks a "contains" method, which is why I am using String.

An example request looks like:

```
POST /my/path HTTP/1.1
Host: myserver.com
Authorization: Bearer 12345

```

The code needs to find the `/HTTP/1\.[01]/`in the first line (delimited by an `\r\n` usually, but possibly just `\n`).  If present, it extract all data up to that sequence, which exclude all the subsequent headers.

carl-mastrangelo:
One other question: what happens if the server sends the preface 1 byte at a time?  What will be the contents of the ByteBuf?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -25,12 +25,15 @@
...
 import io.netty.handler.codec.http2.Http2Exception.StreamException;
+import io.netty.util.AsciiString;
+import io.netty.util.CharsetUtil;
 import io.netty.util.concurrent.ScheduledFuture;
 import io.netty.util.internal.UnstableApi;
 import io.netty.util.internal.logging.InternalLogger;
 import io.netty.util.internal.logging.InternalLoggerFactory;
 
 import java.net.SocketAddress;
+import java.nio.charset.Charset;
```
netkins: MINOR Remove this unused import 'java.nio.charset.Charset'. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -272,6 +275,13 @@ private boolean readClientPrefaceString(ByteBuf in) throws Http2Exception {
             if (bytesRead == 0 || !ByteBufUtil.equals(in, in.readerIndex(),
                                                       clientPrefaceString, clientPrefaceString.readerIndex(),
                                                       bytesRead)) {
+                String http1Request =
+                    in.toString(in.readerIndex(), Math.min(in.readableBytes(), 1024), CharsetUtil.US_ASCII);
```
netkins: MAJOR Unnecessary use of fully qualified name 'Math.min' due to existing static import 'java.lang.Math.min' [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUnnecessaryFullyQualifiedName)

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -25,12 +25,15 @@
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http2.Http2Exception.CompositeStreamException;
 import io.netty.handler.codec.http2.Http2Exception.StreamException;
+import io.netty.util.AsciiString;
```
netkins: MINOR Remove this unused import 'io.netty.util.AsciiString'. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
Scottmitch: > "HTTP/1.1" is the last part of the first line

👍  ... 🤦‍♂️ 

> ByteBuf lacks a "contains" method

we could add a method to `ByteBufUtil` which borrows from [AsciiString#indexOf](https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/AsciiString.java#L675) (which is what `AsciiString#contains` uses).

Scottmitch:
> what happens if the server sends the preface 1 byte at a time

You mean if the client sends 1 bytes at a time ... or more importantly the server receives 1 byte at a time?

`readClientPrefaceString` throws an error as soon as what is read deviates from the expected preface ... so your debug output would just contain 1 byte in this case.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
carl-mastrangelo: I tried adding a method to ByteBufUtil.  This skips the parsing to ascii, though, I think it gives up any sort of smart string search the JVM would do.

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -159,6 +159,22 @@ public static int hashCode(ByteBuf buffer) {
     }
 
     /**
+     * Returns the reader index of needle in haystack, or -1 if needle is not in haystack.
+     */
+    public static int indexOf(ByteBuf needle, ByteBuf haystack) {
+        // TODO: maybe use Boyer Moore for efficiency.
```
netkins: INFO Complete the task associated to this TODO comment. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1135)

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -272,6 +277,13 @@ private boolean readClientPrefaceString(ByteBuf in) throws Http2Exception {
             if (bytesRead == 0 || !ByteBufUtil.equals(in, in.readerIndex(),
                                                       clientPrefaceString, clientPrefaceString.readerIndex(),
                                                       bytesRead)) {
+                int http1Index =
+                    ByteBufUtil.indexOf(HTTP_1_X_BUF, in.slice(in.readerIndex(), min(in.readableBytes(), 1024)));
+                if (http1Index != -1) {
+                    String chunk = in.slice(in.readerIndex(), http1Index - in.readerIndex())
+                        .toString(CharsetUtil.US_ASCII);
```
normanmaurer: nit: Not need to create a slice for it. Just use:

```java
String chunk = in.toString(CharsetUtil.US_ASCII, in.readerIndex(), http1Index - in.readerIndex());
```

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2ConnectionHandlerTest.java
```diff
@@ -243,6 +244,17 @@ public void serverReceivingInvalidClientPrefaceStringShouldHandleException() thr
     }
 
     @Test
+    public void serverReceivingHttp1ClientPrefaceStringShouldIncludePreface() throws Exception {
+        when(connection.isServer()).thenReturn(true);
+        handler = newHandler();
+        handler.channelRead(ctx, copiedBuffer("GET /path HTTP/1.1", UTF_8));
```
normanmaurer: UTF_8 -> US_ASCII ?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -272,6 +277,13 @@ private boolean readClientPrefaceString(ByteBuf in) throws Http2Exception {
             if (bytesRead == 0 || !ByteBufUtil.equals(in, in.readerIndex(),
                                                       clientPrefaceString, clientPrefaceString.readerIndex(),
                                                       bytesRead)) {
+                int http1Index =
+                    ByteBufUtil.indexOf(HTTP_1_X_BUF, in.slice(in.readerIndex(), min(in.readableBytes(), 1024)));
```
normanmaurer: I am still not sure where this 1024 comes from. Can you add a comment why 1024 was chosen ?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -195,6 +197,9 @@ public boolean prefaceSent() {
         }
     }
 
+    private static final ByteBuf HTTP_1_X_BUF = Unpooled.unreleasableBuffer(
```
Scottmitch: nit: move this to the top of the class?

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2ConnectionHandlerTest.java
```diff
@@ -243,6 +244,17 @@ public void serverReceivingInvalidClientPrefaceStringShouldHandleException() thr
...
+        ArgumentCaptor<ByteBuf> captor = ArgumentCaptor.forClass(ByteBuf.class);
+        verify(frameWriter).writeGoAway(eq(ctx), eq(0), eq(PROTOCOL_ERROR.code()),
+            captor.capture(), eq(promise));
```
Scottmitch: Wouldn't this test case pass even without the code changes in `Http2ConnectionHandler.java`? Should we verify the content to differentiate from `serverReceivingInvalidClientPrefaceStringShouldHandleException` above?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
carl-mastrangelo: Added a comment.  There isn't a strong reason.  A number is needed so I picked one.  I think IE is around 2000 max path length, but that seems too long.  I'm content for the other error message to be returned in that case.

The RFC (7230, 7231) doesn't seem to specify a limit and implies it is up to the implementation.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
carl-mastrangelo: Done.

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2ConnectionHandlerTest.java
carl-mastrangelo: Done.  Technically the other one should be changed too since HTTP doesn't allow UTF_8 in that location.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
carl-mastrangelo: Sure.  I put it here as it was in your suggested changes.

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2ConnectionHandlerTest.java
carl-mastrangelo: Yeah, but I just wanted to make sure I didn't cause an IOOBE.  I don't see any clean way to get access to the error message.  Suggestions welcome!

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2ConnectionHandlerTest.java
```diff
@@ -243,6 +245,17 @@ public void serverReceivingInvalidClientPrefaceStringShouldHandleException() thr
     }
 
     @Test
+    public void serverReceivingHttp1ClientPrefaceStringShouldIncludePreface() throws Exception {
+        when(connection.isServer()).thenReturn(true);
+        handler = newHandler();
+        handler.channelRead(ctx, copiedBuffer("GET /path HTTP/1.1", US_ASCII));
+        ArgumentCaptor<ByteBuf> captor = ArgumentCaptor.forClass(ByteBuf.class);
+        verify(frameWriter).writeGoAway(eq(ctx), eq(0), eq(PROTOCOL_ERROR.code()),
```
Scottmitch: The `frameWriter` mock is configured in `setup` to release the buffer so each test doesn't have to worry about this (see below). For this test since we want to verify the content we should install a new handler for `writeGoAway` so we can verify the content of the message.

```java
doAnswer(new Answer<ChannelFuture>() {
            @Override
            public ChannelFuture answer(InvocationOnMock invocation) throws Throwable {
                ByteBuf buf = invocation.getArgumentAt(3, ByteBuf.class);
                buf.release();
                return future;
            }
        }).when(frameWriter).writeGoAway(
                any(ChannelHandlerContext.class), anyInt(), anyInt(), any(ByteBuf.class), any(ChannelPromise.class));
```

carl-mastrangelo: The error message is logged, but not part of the response right?

Scottmitch: We send the exception message in the GO_AWAY [1]. If you put a breakpoint in the code I referenced above we send `Unexpected HTTP/1.x request: GET /path`.

[1] https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java#L805

carl-mastrangelo: Hmm, I was under the impression that it isn't kosher to change the methods on a mock, after they have been mocked out.  I captured the string version as part of the class

Scottmitch: you may not get the desired behavior by stacking `doAnswer` methods (not sure). What you have done lgtm and in the future is also a `reset` method which I believe dumps all mocked methods.

carl-mastrangelo:
Done!

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2ConnectionHandlerTest.java
carl-mastrangelo: Ack to `reset`.  Stylists at Google have mixed opinions on it.

Scottmitch:
4.1 (ead9938)

PR #6276, opened 2017-01-26
More precise calculate the maximum record size when using SslProvider…

normanmaurer:
….OPENSSL* and so decrease mem usage.

Motivation:

We used ca 2k as maximum overhead for encrypted packets which is a lot more then what is needed in reality by OpenSSL. This could lead to the need of more memory.

Modification:

- Use a lower overhead of 86 bytes as defined by the spec and openssl itself
- Fix unit test to use the correct session to calculate needed buffer size

Result:

Less memory usage.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -151,14 +151,24 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
         SET_SERVER_NAMES_METHOD = setServerNamesMethod;
     }
 
-    private static final int MAX_PLAINTEXT_LENGTH = 16 * 1024; // 2^14
-    private static final int MAX_COMPRESSED_LENGTH = MAX_PLAINTEXT_LENGTH + 1024;
-    private static final int MAX_CIPHERTEXT_LENGTH = MAX_COMPRESSED_LENGTH + 1024;
+    static final int MAX_PLAINTEXT_LENGTH = 16 * 1024; // 2^14
 
-    // Header (5) + Data (2^14) + Compression (1024) + Encryption (1024) + MAC (20) + Padding (256)
-    static final int MAX_ENCRYPTED_PACKET_LENGTH = MAX_CIPHERTEXT_LENGTH + 5 + 20 + 256;
+    /**
+     * This is the maximum overhead when encrypting plaintext as defined by
+     * <a href="https://www.ietf.org/rfc/rfc5246.txt">rfc5264</a>,
+     * <a href="https://www.ietf.org/rfc/rfc5289.txt">rfc5289</a> and openssl implementation itself.
+     *
+     * Please note that we use a padding of 16 here as openssl uses PKC#5 which uses 16 bytes while the spec itself
+     * allow up to 255 bytes. 16 bytes is the max for PKC#5 (which handles it the same way as PKC#7) as we use a block
+     * size of 16. See <a href="https://tools.ietf.org/html/rfc5652#section-6.3">rfc5652#section-6.3</a>.
+     *
+     * 16 (IV) + 48 (MAC) + 1 (Padding_length field) + 15 (Padding) + 1 (ContentType) + 2 (ProtocolVersion) + 2 (Length)
+     *
+     * TODO: We may need to review this calculation once TLS 1.3 becomes available.
+     */
+    static final int MAX_ENCRYPTION_OVERHEAD_LENGTH = 15 + 48 + 1 + 16 + 1 + 2 + 2;
 
-    static final int MAX_ENCRYPTION_OVERHEAD_LENGTH = MAX_ENCRYPTED_PACKET_LENGTH - MAX_PLAINTEXT_LENGTH;
+    static final int MAX_ENCRYPTED_PACKET_LENGTH = MAX_PLAINTEXT_LENGTH + MAX_ENCRYPTION_OVERHEAD_LENGTH;
 
     private static final int MAX_ENCRYPTION_OVERHEAD_DIFF = Integer.MAX_VALUE - MAX_ENCRYPTION_OVERHEAD_LENGTH;
 
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -734,7 +734,10 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
             // See https://github.com/netty/netty/issues/5860
             while (!ctx.isRemoved()) {
                 if (out == null) {
-                    out = allocateOutNetBuf(ctx, 0);
+                    // As this is called for the handshake we have no real idea how big the buffer needs to be.
+                    // That said 2048 should give us enough room to include everything like ALPN / NPN data.
+                    // If this is not enough we will increase the buffer in wrap(...).
+                    out = allocateOutNetBuf(ctx, 2048);
                 }
                 SSLEngineResult result = wrap(alloc, engine, Unpooled.EMPTY_BUFFER, out);
 
--- a/handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
@@ -27,12 +27,17 @@ import org.junit.Test;
 
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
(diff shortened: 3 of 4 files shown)


handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -151,14 +151,24 @@
...
+    /**
+     * This is the maximum overhead when encrypting plaintext as defined by
+     * <a href="https://www.ietf.org/rfc/rfc5246.txt">rfc5264</a>,
+     * <a href="https://www.ietf.org/rfc/rfc5289.txt">rfc5289</a> and openssl implementation itself.
+     *
+     * Please note that we use a padding of 16 here as openssl uses PKC#5 which uses 16 bytes while the spec itself
+     * allow up to 255 bytes. 16 bytes is the max for PKC#5 (which handles it the same way as PKC#7) as we use a block
+     * size of 16. See <a href="https://tools.ietf.org/html/rfc5652#section-6.3">rfc5652#section-6.3</a>.
+     *
+     * 16 (IV) + 48 (MAC) + 1 (Padding_length field) + 16 (Padding) + 1 (ContentType) + 2 (ProtocolVersion) + 2 (Length)
```
normanmaurer: @davidben could you please also verify this, just to double check I am not doing anything stupid 😄

normanmaurer:
@nmittler @Scottmitch FYI... this highly reduce the needed buffer size when encrypt small buffers and using `SslProvider.OPENSSL*`.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -151,14 +151,24 @@
...
+     *
+     * TODO: We may need to review this calculation once TLS 1.3 becomes available.
```
netkins: INFO Complete the task associated to this TODO comment. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1135)

Scottmitch: create an issue to investigate?

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -734,7 +734,10 @@ private void wrapNonAppData(ChannelHandlerContext ctx, boolean inUnwrap) throws
             // See https://github.com/netty/netty/issues/5860
             while (!ctx.isRemoved()) {
                 if (out == null) {
-                    out = allocateOutNetBuf(ctx, 0);
+                    // As this is called for the handshake we have no real idea how big the buffer needs to be.
+                    // That said 2048 should give us enough room to include everything like ALPN / NPN data.
+                    // If this is not enough we will increase the buffer in wrap(...).
+                    out = allocateOutNetBuf(ctx, 2048);
```
Scottmitch: mental note to my self ... we may be able to use something like [MAX_BIO_BYTES](https://github.com/netty/netty/pull/6201/files#diff-82f94494cc35616a8a19b6aa9ed033efR90) in the future or create another define for "non-app-data"

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
davidben: There's also AEAD-based ciphers which have a completely different derivation, but they all have much smaller overhead than this. I don't believe you need the +1. There is a length, but since the goal is to get the total size to a multiple of the block size, minimal padding will only take you up at most one block anyway.

normanmaurer: @davidben so just to double-check, in general we should be good with how I calculate it (it's not too small) ?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: Will do once this is merged

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
davidben: Looks good. You can even remove the "+ 1 (Padding_length field)".

(BoringSSL does implement TLS 1.0 1/n-1 record-splitting by pretending the extra record is a really inefficient record header, but the largest MAC in a TLS 1.0 CBC cipher is 20 bytes for SHA-1 and TLS 1.0 is missing the explicit IV, so it fits in there.)

But rather than trust me to get my math right, I'd suggest adding a suite of tests which runs over ever cipher + every version and make sure you can write records at various sizes.

davidben:
By the way, at some point, BoringSSL will probably have some kind of in-place/buffer-free/BIO-free/bring-your-own-buffer/whatever-you-call-it API that should align much better with the wrap/unwrap thing you're doing.

(Although the first iteration is probably not going to be require contiguous buffers, so you'll probably still need to do some copying.)

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: Can you clarify why I not need the +1 ? I was under the impressions it's needed from reading the spec

davidben: The goal when padding is to get up to a multiple of the block size (say 16 for AES). So if my input has length, say, 26, I want 6 total bytes of padding which is 5 byte of "padding" and one byte of "padding length".

If you only ever add the minimal padding needed, as OpenSSL does, you will add somewhere between 0 and 15 bytes of padding, or somewhere between 0 + 1 = 1 and 15 + 1 = 16 bytes total. You wouldn't add 16 + 1 = 17 bytes of padding because you instead would have added 0 + 1 = 1 bytes.

normanmaurer: Ah got it. Thanks again 👍 Will add tests now (good idea)

Scottmitch:
@davidben - Interesting ... would this be better than implanting our own BIO which directly writes/reads from ByteBuffer's direct memory (like in https://github.com/netty/netty/pull/6201)?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -151,14 +151,24 @@
...
+     * <a href="https://www.ietf.org/rfc/rfc5246.txt">rfc5264</a>,
+     * <a href="https://www.ietf.org/rfc/rfc5289.txt">rfc5289</a> and openssl implementation itself.
+     *
+     * Please note that we use a padding of 16 here as openssl uses PKC#5 which uses 16 bytes while the spec itself
+     * allow up to 255 bytes. 16 bytes is the max for PKC#5 (which handles it the same way as PKC#7) as we use a block
+     * size of 16. See <a href="https://tools.ietf.org/html/rfc5652#section-6.3">rfc5652#section-6.3</a>.
+     *
+     * 16 (IV) + 48 (MAC) + 1 (Padding_length field) + 15 (Padding) + 1 (ContentType) + 2 (ProtocolVersion) + 2 (Length)
+     *
+     * TODO: We may need to review this calculation once TLS 1.3 becomes available.
```
netkins: INFO Complete the task associated to this TODO comment. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1135)

davidben:
@Scottmitch You should be able to save the copies in and out of the SSL stack's internal buffer? But I haven't looked very closely at your current code or the PR. Trying to mimic wrap/unwrap over `SSL_read` also has the problem that you can't "unread" data on BUFFER_UNDERFLOW, which is probably part of why you have a reimplemented record layer? This would avoid that.

(To clarify, this is more a direction we're keeping in mind as we rework the stack thing rather than something you'll have available next week or anything.)

normanmaurer:
@davidben FYI... added unit-tests and it seems like it "just works" :) Thanks again for the suggestion

Scottmitch:
> you can't "unread" data on BUFFER_UNDERFLOW

@davidben - Currently we don't attempt to read if we detect there is not enough room to store the data.

> part of why you have a reimplemented record layer

By "record layer" do you mean the BIO implementation https://github.com/netty/netty-tcnative/pull/216?

normanmaurer:
Cherry-picked into 4.1 (91f050d2ef6b22b7aec187aa1cf4593955dcea82) and 4.0 (8066af386ae266ca0d5b10024c5434f9dc991c99)

PR #6277, opened 2017-01-26
Update jetty-alpn-agent to support latest JDK releases.

normanmaurer:
Motivation:

We need to update jetty-alpn-agent to support latest JDK releases.

Modifications:

Update jetty-alpn-agent to 2.0.6

Result:

Be able to run tests with latest JDK releases.


normanmaurer:
Cherry-picked into 4.1 (b69c772) and 4.0 (f3caef6bc5056b04b272b62b1db5518b902387d3)

PR #6278, opened 2017-01-26
Add unit test that shows we correctly return BUFFER_UNDERFLOW

normanmaurer:
Motivation:

We should test that we correctly return BUFFER_UNDERFLOW if the src buffer not contains enough data to unwrap it.

Modification:

Add unit test to verify behaviour.

Result:

Better test coverrage of SSLEngine implementations.


normanmaurer:
more tests FTW

normanmaurer:
Cherry-picked into 4.1 (5cd8133477aa87445fe4277550094237fd2585ed) and 4.1 (2f4298a)

normanmaurer:
Also thanks to @johnou for bringing up that we have no test for this :)

PR #6279, opened 2017-01-26
Null out references to tmpNioBuf and chunk to allow quicker collecting

normanmaurer:
Motivation:

In PooledByteBuf we missed to null out the chunk and tmpNioBuf fields before recycle it to the Recycler. This could lead to keep objects longer alive then necessary which may hold a lot of memory.

Modifications:

Null out tmpNioBuf and chunk before recycle.

Result:

Possible to earlier GC objects.


normanmaurer:
This is interesting as PoolChunk instances may be dropped when not used anymore.

normanmaurer:
Cherry-Picked into 4.1 (8dda984afe9477fa09fe166305381d9ed1afa7c4) and 4.0 (dcf4292647021e2f593a52fcfddd7fcb945693e4)

PR #6280, opened 2017-01-26
Automatically decode DNS domain name to unicode

normanmaurer:
Motivation:

DnsNameResolver will return the domain / host name as ascii code using punycode (https://tools.ietf.org/html/rfc3492). This is different to what the JDK does which always convert it to unicode. We should do the same by default but allow to also not do it.

Modifications:

- Add new builder method on DnsNameResolverBuilder which allow to disable / enable converting. Default is to convert just like the JDK does.
- Add unit tests for it.

Result:

DnsNameResolver and JDK impl behave the same way.

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
@@ -155,6 +155,7 @@ public class DnsNameResolver extends InetNameResolver {
     private final boolean cnameFollowAAAARecords;
     private final InternetProtocolFamily preferredAddressType;
     private final DnsRecordType[] resolveRecordTypes;
+    private final boolean decodeIdn;
 
     /**
      * Creates a new DNS-based name resolver that communicates with the specified list of DNS servers.
@@ -175,7 +176,11 @@ public class DnsNameResolver extends InetNameResolver {
      * @param hostsFileEntriesResolver the {@link HostsFileEntriesResolver} used to check for local aliases
      * @param searchDomains the list of search domain
      * @param ndots the ndots value
+     * @deprecated use {@link #DnsNameResolver(EventLoop, ChannelFactory, DnsServerAddresses, DnsCache, long,
+     *             InternetProtocolFamily[], boolean, int, boolean, int,
+     *             boolean, HostsFileEntriesResolver, String[], int, boolean)}
      */
+    @Deprecated
     public DnsNameResolver(
             EventLoop eventLoop,
             ChannelFactory<? extends DatagramChannel> channelFactory,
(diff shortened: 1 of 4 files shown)


resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverBuilder.java
```diff
@@ -331,6 +332,18 @@ public DnsNameResolverBuilder ndots(int ndots) {
     }
 
     /**
+     * Set if domain / host names should be decoded to unicode when received.
+     * See <a href="https://tools.ietf.org/html/rfc3492">rfc3492</a>.
+     *
+     * @param decodeToUnicode if should get decoded
+     * @return {@code this}
+     */
+    public DnsNameResolverBuilder decodeToUnicode(boolean decodeToUnicode) {
```
normanmaurer: @Scottmitch @trustin I am happy to take any suggestions for better naming :)

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -191,6 +196,49 @@ public DnsNameResolver(
...
+     * @param traceEnabled if trace is enabled
+     * @param maxPayloadSize the capacity of the datagram packet buffer
+     * @param optResourceEnabled if automatic inclusion of a optional records is enabled
+     * @param hostsFileEntriesResolver the {@link HostsFileEntriesResolver} used to check for local aliases
+     * @param searchDomains the list of search domain
+     * @param ndots the ndots value
+     * @param decodeToUnicode {@code true} if domain / host names should be decoded to unicode when received.
+     *                        See <a href="https://tools.ietf.org/html/rfc3492">rfc3492</a>.
+     */
+    public DnsNameResolver(
```
netkins: MAJOR Constructor has 15 parameters, which is greater than 7 authorized. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS00107)

nmittler: nit: we should probably either get rid of this restriction or change the method signature.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverBuilder.java
nmittler: hmm ... maybe something like `assumeUnicodeDomainName`?

Scottmitch: not sure ... `domainNameResponsesAreUnicode`?

normanmaurer: Don't like any of these... and also not like my name. Let me think a bit more about it :) Maybe we should just always convert for now and add a param later on ?

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
normanmaurer: @nmittler let me get rid.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverBuilder.java
nmittler: That SGTM

normanmaurer: @Scottmitch ok for you as well ?

Scottmitch: I'm thinking we should at least have an option to comply with the RFC (if not do this by default)

normanmaurer: Yeah you are probably right... Let me try to come up with a good name ...

trustin: `enableIdnaDecode` ?

trustin: `decodeIdn`? I'm sure IDN is a well known term in this area.

normanmaurer: @trustin good idea... let me change it to `decodeIdn` and merge.

normanmaurer:
Cherry-picked into 4.1 (0d5b665)

PR #6284, opened 2017-01-26
HTTP/2 revert Http2FrameWriter throws API change

Scottmitch:
Motivation:
2fd42cfc6b32dacb812c6f2ecef8522cd94718d4 fixed a bug related to encoding headers but it also introduced a throws statement onto the Http2FrameWriter methods which write headers. This throws statement makes the API more verbose and is not necessary because we can communicate the failure in the ChannelFuture that is returned by these methods.

Modifications:
- Remove throws from all Http2FrameWriter methods.

Result:
Http2FrameWriter APIs do not propagate checked exceptions.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
@@ -187,16 +187,22 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
                 }
                 ChannelFuture future = frameWriter.writeHeaders(ctx, streamId, headers, streamDependency,
                                                                 weight, exclusive, padding, endOfStream, promise);
-                // Synchronously set the headersSent flag to ensure that we do not subsequently write
-                // other headers containing pseudo-header fields.
-                stream.headersSent();
+                // Writing headers may fail during the encode state if they violate HPACK limits.
+                Throwable failureCause = future.cause();
+                if (failureCause == null) {
+                    // Synchronously set the headersSent flag to ensure that we do not subsequently write
+                    // other headers containing pseudo-header fields.
+                    stream.headersSent();
+                } else {
+                    lifecycleManager.onError(ctx, failureCause);
+                }
 
                 return future;
             } else {
                 // Pass headers to the flow-controller so it can maintain their sequence relative to DATA frames.
                 flowController.addFlowControlled(stream,
                         new FlowControlledHeaders(stream, headers, streamDependency, weight, exclusive, padding,
-                                                 endOfStream, promise));
+                                                  endOfStream, promise));
                 return promise;
             }
         } catch (Throwable t) {
@@ -276,7 +282,13 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
 
             ChannelFuture future = frameWriter.writePushPromise(ctx, streamId, promisedStreamId, headers, padding,
                                                                 promise);
-            stream.pushPromiseSent();
+            // Writing headers may fail during the encode state if they violate HPACK limits.
+            Throwable failureCause = future.cause();
+            if (failureCause == null) {
+                stream.pushPromiseSent();
+            } else {
+                lifecycleManager.onError(ctx, failureCause);
+            }
             return future;
         } catch (Throwable t) {
             lifecycleManager.onError(ctx, t);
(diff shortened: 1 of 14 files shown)


codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -187,16 +187,22 @@ public void operationComplete(ChannelFuture future) throws Exception {
                 }
                 ChannelFuture future = frameWriter.writeHeaders(ctx, streamId, headers, streamDependency,
                                                                 weight, exclusive, padding, endOfStream, promise);
-                // Synchronously set the headersSent flag to ensure that we do not subsequently write
-                // other headers containing pseudo-header fields.
-                stream.headersSent();
+                // Writing headers may fail during the encode state if they violate HPACK limits.
+                Throwable failureCause = future.cause();
```
normanmaurer: I wonder if we need to check isDone() first and if not do it in a listener?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -276,7 +282,13 @@ public ChannelFuture writePushPromise(ChannelHandlerContext ctx, int streamId, i
...
             ChannelFuture future = frameWriter.writePushPromise(ctx, streamId, promisedStreamId, headers, padding,
                                                                 promise);
-            stream.pushPromiseSent();
+            // Writing headers may fail during the encode state if they violate HPACK limits.
+            Throwable failureCause = future.cause();
+            if (failureCause == null) {
+                stream.pushPromiseSent();
+            } else {
+                lifecycleManager.onError(ctx, failureCause);
+            }
```
normanmaurer: Same as above

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -450,15 +462,21 @@ public void error(ChannelHandlerContext ctx, Throwable cause) {
...
-            stream.headersSent();
+            ChannelFuture f = frameWriter.writeHeaders(ctx, stream.id(), headers, streamDependency, weight, exclusive,
+                                                       padding, endOfStream, promise);
+            // Writing headers may fail during the encode state if they violate HPACK limits.
+            Throwable failureCause = f.cause();
+            if (failureCause == null) {
+                stream.headersSent();
+            } else {
+                lifecycleManager.onError(ctx, failureCause);
+            }
```
normanmaurer: Same as above

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2FrameWriter.java
```diff
@@ -337,8 +337,6 @@ public ChannelFuture writePushPromise(ChannelHandlerContext ctx, int streamId,
             }
         } catch (Http2Exception e) {
             promiseAggregator.setFailure(e);
-            promiseAggregator.doneAllocatingPromises();
```
normanmaurer: Why remove this? Also why we can't just have the other catch block?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
Scottmitch: We make assumptions about what is being done in `FrameWriter#writeHeaders` and properties of the underlying transport. In this context the method is assumed to encode the headers and call `ctx.write()` in a synchronous manner. Also if a write fails we assume the transport is reliable/ordered and therefore unusable so we generally don't care about promise failures. So I'm not sure we need to handle the async failure case, and if you think we do this needs to be done more generally throughout codec-http2. WDYT?

normanmaurer: Sounds a bit like we depend on an implementation detail here. I guess for now it's ok but definitely something we should revisit later on.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2FrameWriter.java
Scottmitch: it is done below on the return statement so no need to call it twice.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
Scottmitch: https://github.com/netty/netty/issues/6287 .. WDYT?

normanmaurer: +1

Scottmitch:
4.1 (e13da21)

Issue #6225, opened 2017-01-17
ResourceLeakDetector on HashWheeledTimer instance threshold too low?


jroper:
As I understand it, if you have 1 processor, the resource leak detector will complain if you create more than 4 instances of a HashedWheelTimer.

In the JVM world, it's becoming very easy to need more than 4 instances of this.  Consider an app that has a database driver that uses and manages its own Netty instance, a message broker client that uses and manages its own Netty instance, an HTTP client that uses and manages its own Netty instance, an HTTP server that uses and manages its own Netty instance, and metrics reporting implementation that uses and manages its own Netty instance.  These are all third party libraries, as the end user you're just the app that plumbs them together to implement your business logic, you don't have much control over how they use Netty (in particular you can't tell them to create a HashWheeledTimer with leak detection turned off), yet they each created their own HashWheeledTimer, totaling 5, and you deploy it to a single CPU server and you always get the leak detection warning.  You don't want to turn leak detection off though because that would hide legitimate leaks, but you do want to get rid of this warning.

So firstly, why is the limit of HashedWheelTimers dependent on the number of processors?  The javadocs for it say that there should only be one per app, it says nothing about needing one per processor, nothing about the number of processors would change whether there is a resource leak or not, and the practical consequence is that on developer machines, where you often have 8 or more hyperthreads in total, you'll never see the warning, but then deploy to a single CPU container in production, and you'll get it.  It would be better to have a consistent behaviour between dev and production.

Secondly, the number should be something that reflects an upper bound for the number of independent Netty instances that a single JVM classloader could conceivably have without having a misconfiguration or other problem.  I'd say this number should be at least 16, maybe 32.

Scottmitch:
Thanks for reporting @jroper! The following is not clear to me:

- why in general is the creation of leak records is related to the lifetime of the objets being tracked [1]
- why tripping this threshold is logged as an error
- why the initial choice of `Runtime.getRuntime().availableProcessors() * 4L` was made [2]

At one point the sample rate was governed by a counter and some modulo arithmetic [3] (ignore the volatile increment bug). Now sampling is determine by a RNG distribution [4] and even if there was a nice relationship between leak record creation and object lifetime which we are tracking it is possible we could trip this threshold due to RNG "unluckiness".  I'm not convinced this type of "limiting" makes much sense in `ResourceLeakTracker` ... @trustin @normanmaurer can you help me understand?

[...]

PR #6289, opened 2017-01-27
Deprecate and ignore ResourceLeakDetector's maxActive parameter

Scottmitch:
Motivation:
ResourceLeakDetector supports a parameter called maxActive. This parameter is used in attempt to limit the amount of objects which are being tracked for leaks at any given time, and generates an error log message if this limit is exceeded. This assumes that there is a relationship between leak sample rate and object lifetime for objects which are already being tracked. This relationship may appear to work in cases were there are a single leak record per object and those leak records live for the lifetime of the application but in general this relationship doesn't exist. The original motivation was to provide a limit for cases such as HashedWheelTimer to limit the number of instances which exist at any given time. This limit is not enforced in all circumstances in HashedWheelTimer (e.g. if the thread is a daemon) and can be implemented outside ResourceLeakDetector.

Modifications:
- Deprecate all methods which interact with maxActive in ResourceLeakDetectorFactory and ResourceLeakDetector
- Remove all logic related to maxActive in ResourceLeakDetector
- HashedWheelTimer implements its own logic to impose a limit and warn users if too many instances exists at any given time.

Result:
Fixes https://github.com/netty/netty/issues/6225.

### DIFF
--- a/common/src/main/java/io/netty/util/HashedWheelTimer.java
+++ b/common/src/main/java/io/netty/util/HashedWheelTimer.java
@@ -29,8 +29,12 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.concurrent.atomic.AtomicLong;
+
+import static io.netty.util.internal.StringUtil.simpleClassName;
 
 /**
  * A {@link Timer} optimized for approximated I/O timeout scheduling.
@@ -78,8 +82,11 @@ public class HashedWheelTimer implements Timer {
     static final InternalLogger logger =
             InternalLoggerFactory.getInstance(HashedWheelTimer.class);
 
+    private static final AtomicInteger INSTANCE_COUNTER = new AtomicInteger();
+    private static final AtomicBoolean WARNED_TOO_MANY_INSTANCES = new AtomicBoolean();
+    private static final int INSTANCE_COUNT_LIMIT = 64;
     private static final ResourceLeakDetector<HashedWheelTimer> leakDetector = ResourceLeakDetectorFactory.instance()
-            .newResourceLeakDetector(HashedWheelTimer.class, 1, Runtime.getRuntime().availableProcessors() * 4L);
+            .newResourceLeakDetector(HashedWheelTimer.class, 1);
 
     private static final AtomicIntegerFieldUpdater<HashedWheelTimer> WORKER_STATE_UPDATER =
             AtomicIntegerFieldUpdater.newUpdater(HashedWheelTimer.class, "workerState");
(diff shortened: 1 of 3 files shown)


Scottmitch:
@jroper - FYI

common/src/main/java/io/netty/util/HashedWheelTimer.java
```diff
@@ -266,6 +273,24 @@ public HashedWheelTimer(
...
         this.maxPendingTimeouts = maxPendingTimeouts;
+
+        if (INSTANCE_COUNTER.incrementAndGet() > INSTANCE_COUNT_LIMIT &&
+            WARNED_TOO_MANY_INSTANCES.compareAndSet(false, true)) {
+            reportTooManyInstances();
+        }
+    }
+
+    @Override
+    protected void finalize() throws Throwable {
```
normanmaurer: @Scottmitch finalizers make me 😢

common/src/main/java/io/netty/util/HashedWheelTimer.java
```diff
@@ -337,33 +362,37 @@ public void start() {
...
-        }
-
-        if (leak != null) {
-            boolean closed = leak.close(this);
-            assert closed;
+            if (interrupted) {
+                Thread.currentThread().interrupt();
+            }
+        } finally {
+            INSTANCE_COUNTER.decrementAndGet();
```
normanmaurer: maybe add any assert that this can never be negative ?

common/src/main/java/io/netty/util/ResourceLeakDetectorFactory.java
```diff
@@ -57,30 +57,44 @@ public static void setResourceLeakDetectorFactory(ResourceLeakDetectorFactory fa
...
 
     /**
+     * Returns a new instance of a {@link ResourceLeakDetector} with the given resource class.
+     *
+     * @param resource the resource class used to initialize the {@link ResourceLeakDetector}
+     * @param samplingInterval the interval on which sampling takes place
+     * @param <T> the type of the resource class
+     * @return a new instance of {@link ResourceLeakDetector}
+     */
+    public abstract <T> ResourceLeakDetector<T> newResourceLeakDetector(Class<T> resource, int samplingInterval);
```
normanmaurer: @Scottmitch this is a breaking change. We can not add a new abstract method as the class is public.

common/src/main/java/io/netty/util/HashedWheelTimer.java
Scottmitch: me too 😭 

I'm open for alternatives.

common/src/main/java/io/netty/util/ResourceLeakDetectorFactory.java
Scottmitch: true dat ... let me fix this.

common/src/main/java/io/netty/util/HashedWheelTimer.java
Scottmitch: it could actually go negative (e.g. overflow) as we don't strictly enforce the limit ... we just use it to warn for "too many instances". In practice I wouldn't expect this to happen but I'm not sure we should add an assert for it unless we strictly enforce it.

Scottmitch:
@normanmaurer - PTAL

common/src/main/java/io/netty/util/HashedWheelTimer.java
```diff
@@ -400,6 +429,13 @@ private boolean shouldLimitTimeouts() {
         return maxPendingTimeouts > 0;
     }
 
+    private void reportTooManyInstances() {
```
netkins: MINOR Make "reportTooManyInstances" a "static" method. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2325)

common/src/main/java/io/netty/util/HashedWheelTimer.java
netkins: CRITICAL Do not override the Object.finalize() method. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AObjectFinalizeOverridenCheck)

common/src/main/java/io/netty/util/ResourceLeakDetectorFactory.java
```diff
@@ -57,30 +57,47 @@ public static void setResourceLeakDetectorFactory(ResourceLeakDetectorFactory fa
...
     /**
+     * Returns a new instance of a {@link ResourceLeakDetector} with the given resource class.
+     *
+     * @param resource the resource class used to initialize the {@link ResourceLeakDetector}
+     * @param samplingInterval the interval on which sampling takes place
+     * @param <T> the type of the resource class
+     * @return a new instance of {@link ResourceLeakDetector}
+     */
+    @SuppressWarnings("deprecation")
+    public <T> ResourceLeakDetector<T> newResourceLeakDetector(Class<T> resource, int samplingInterval) {
```
netkins: MAJOR Remove this unused method parameter "samplingInterval". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

common/src/main/java/io/netty/util/HashedWheelTimer.java
normanmaurer: ah gotcha... good as it is

normanmaurer:
LGTM... @jroper @trustin PTAL as well

Scottmitch:
4.1 (14b902f) 4.0 (a7007ca)

PR #6290, opened 2017-01-27
Correctly detect which protocols are supported when using OpenSSL

normanmaurer:
Motivation:

We failed to properly test if a protocol is supported on an OpenSSL installation and just always returned all protocols.

Modifications:

- Detect which protocols are supported on a platform.
- Skip protocols in tests when not supported. This fixes a build error on some platforms introduced by [#6276].

Result:

Correctly return only the supported protocols

### DIFF
--- a/handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
@@ -22,6 +22,7 @@ import io.netty.handler.ssl.ApplicationProtocolConfig.SelectorFailureBehavior;
 import io.netty.handler.ssl.util.InsecureTrustManagerFactory;
 import io.netty.handler.ssl.util.SelfSignedCertificate;
 import io.netty.util.internal.ThreadLocalRandom;
+import org.junit.Assume;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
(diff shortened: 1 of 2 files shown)


handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
```diff
@@ -426,13 +426,17 @@ public void testWrapWithDifferentSizes() throws Exception {
...
         // SSLv2 is not supported on most openssl installations so skip it.
+
         testWrapWithDifferentSizes(OpenSsl.PROTOCOL_TLS_V1, TLS_V1_CIPHERS);
         testWrapWithDifferentSizes(OpenSsl.PROTOCOL_TLS_V1_1, TLS_V1_1_CIPHERS);
         testWrapWithDifferentSizes(OpenSsl.PROTOCOL_TLS_V1_2, TLS_V1_2_CIPHERS);
         testWrapWithDifferentSizes(OpenSsl.PROTOCOL_SSL_V3, SSL_V3_CIPHERS);
     }
 
     private void testWrapWithDifferentSizes(String protocol, Set<String> ciphers) throws Exception {
+        if (!OpenSsl.SUPPORTED_PROTOCOLS_SET.contains(protocol)) {
```
Scottmitch: assumeTrue(OpenSsl.SUPPORTED_PROTOCOLS_SET.contains(protocol))

normanmaurer:
cherry-picked into 4.1 (7a39afd031accea9ee38653afbd58eb1c466deda) and 4.0 (bb1440759fbcd8221c772404f8269b748e667b1c)

PR #6294, opened 2017-01-30
Ensure tests added in 91f050d2ef6b22b7aec187aa1cf4593955dcea82 work w…

normanmaurer:
…ith different openssl installations

Motivation:

Tests were added in 91f050d2ef6b22b7aec187aa1cf4593955dcea82 to run with different protocols / ciphers. These may fail currently when openssl was compiled without support for the protocol / ciphers.

Modifications:

- Refactor tests to easier understand for which protocol / cipher it failed
- Not fail the test if the protocol is not supported with the used openssl version.

Result:

More robust testing.


normanmaurer:
Cherry-picked into 4.1 (7736534) and 4.0 (0d2a0abaa2e8a598a9a50fc04a8ae13dea0d9c55)

PR #6297, opened 2017-01-31
Allow to configure a maximum select timeout / epoll wait timeout

normanmaurer:
Motivation:

To optimize throughput it can be useful to allow set a maximum select / epoll wait timeout.

Modifications:

Add system property which allows to set a max value.

Result:

More flexible configuration

### DIFF
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
@@ -27,6 +27,7 @@ import io.netty.util.collection.IntObjectMap;
 import io.netty.util.concurrent.RejectedExecutionHandler;
 import io.netty.util.internal.ObjectUtil;
 import io.netty.util.internal.PlatformDependent;
+import io.netty.util.internal.SystemPropertyUtil;
 import io.netty.util.internal.logging.InternalLogger;
 import io.netty.util.internal.logging.InternalLoggerFactory;
 
@@ -45,6 +46,13 @@ final class EpollEventLoop extends SingleThreadEventLoop {
     private static final InternalLogger logger = InternalLoggerFactory.getInstance(EpollEventLoop.class);
     private static final AtomicIntegerFieldUpdater<EpollEventLoop> WAKEN_UP_UPDATER =
             AtomicIntegerFieldUpdater.newUpdater(EpollEventLoop.class, "wakenUp");
+
+    // Just to make it easier to switch between NIO and EPOLL.
+    private static final long MAX_SELECT_TIMEOUT =
+            SystemPropertyUtil.getLong("io.netty.maxSelectTimeout", Long.MAX_VALUE);
+
+    private static final long MAX_EPOLL_WAIT_TIMEOUT =
+            SystemPropertyUtil.getLong("io.netty.maxEpollWaitTimeout", MAX_SELECT_TIMEOUT);
 
     private final FileDescriptor epollFd;
     private final FileDescriptor eventFd;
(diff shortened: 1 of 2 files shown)


transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
```diff
@@ -46,6 +47,13 @@
     private static final AtomicIntegerFieldUpdater<EpollEventLoop> WAKEN_UP_UPDATER =
             AtomicIntegerFieldUpdater.newUpdater(EpollEventLoop.class, "wakenUp");
 
+    // Just to make it easier to switch between NIO and EPOLL.
+    private static final long MAX_SELECT_TIMEOUT =
+            SystemPropertyUtil.getLong("io.netty.maxSelectTimeout", Long.MAX_VALUE);
+
+    private static final long MAX_EPOLL_WAIT_TIMEOUT =
+            SystemPropertyUtil.getLong("io.netty.maxEpollWaitTimeout", MAX_SELECT_TIMEOUT);
```
trustin: Is this ever necessary then?

normanmaurer: @trustin I guess we could just use `io.netty.maxSelectTimeout` all the time if you prefer. WdYT ?

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
```diff
@@ -201,7 +209,8 @@ private int epollWait(boolean oldWakenUp) throws IOException {
         long currentTimeNanos = System.nanoTime();
         long selectDeadLineNanos = currentTimeNanos + delayNanos(currentTimeNanos);
         for (;;) {
-            long timeoutMillis = (selectDeadLineNanos - currentTimeNanos + 500000L) / 1000000L;
+            long timeoutMillis = Math.min(MAX_EPOLL_WAIT_TIMEOUT,
```
Scottmitch: Can someone explain this sleep logic here? The native code also ensures it sleeps the full time if EINTR pops up.

I'm wondering if I need to do something similar in kqueue https://github.com/netty/netty/pull/6000/files#diff-54822e50acc8dd1c0bd8053332e65659R141

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
trustin: Yes, please.

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
trustin: I'm not sure I get your question. Are you asking about `selectDeadLineNanos`? If so, it's to execute a scheduled task.

Scottmitch: I'm trying to understand why we have a loop to call `Native.epollWait` in `EpollEventLoop` and we also loop in [netty_epoll_native_epollWait0](https://github.com/netty/netty/blob/4.1/transport-native-epoll/src/main/c/netty_epoll_native.c#L249) to ensure we wait for the desired time. `delayNanos` gets the amount of time until the "next schedule task to execute (max of 1 second) and scheduling a `Runnable` involves a queueing a wakeup task on the event loop [1]. So even if a new task is scheduled while we are waiting the event loop should be woken up to re-evaluate wait times. It seems like we could reduce thrashing on the event loop when there is a low workload and in general reduce code here. This was the approach I took in KQUEUE but I'm not sure if I'm missing something...

[1] https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java#L212

normanmaurer:
Let me close this one for now and I think we may not need it at all

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
Scottmitch: @normanmaurer @trustin - despite this PR being closed I would still appreciate if someone could provide clarity 👍

PR #6299, opened 2017-01-31
PooledByteBuf.capacity(...) not enforces maxCapacity()

normanmaurer:
Motivation:

PooledByteBuf.capacity(...) miss to enforce maxCapacity() and so its possible to increase the capacity of the buffer even if it will be bigger then maxCapacity().

Modifications:

- Correctly enforce maxCapacity()
- Add unit tests for capacity(...) calls.

Result:

Correctly enforce maxCapacity().

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
@@ -1383,6 +1383,13 @@ public abstract class AbstractByteBuf extends ByteBuf {
         checkReadableBytes0(minimumReadableBytes);
     }
 
+    protected final void checkNewCapacity(int newCapacity) {
+        ensureAccessible();
+        if (newCapacity < 0 || newCapacity > maxCapacity()) {
+            throw new IllegalArgumentException("newCapacity: " + newCapacity + " (expected: 0-" + maxCapacity() + ')');
+        }
+    }
+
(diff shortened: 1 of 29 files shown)


normanmaurer:
Fixing all the bugs FTW 👍

Scottmitch:
thanks for taking care of this @normanmaurer !

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -1383,6 +1383,13 @@ protected final void checkReadableBytes(int minimumReadableBytes) {
         checkReadableBytes0(minimumReadableBytes);
     }
 
+    protected final void checkNewCapacity(int newCapacity) {
+        ensureAccessible();
+        if (newCapacity < 0 || newCapacity > maxCapacity()) {
+            throw new IllegalArgumentException("newCapacity: " + newCapacity);
```
trustin: `+ " (expected: 0-" + maxCapacity() + ')'`?

normanmaurer: done

normanmaurer:
Cherry-picked into 4.1 (66b17310418230d12c21f1f91287c22022d8d847) and 4.0 (66b17310418230d12c21f1f91287c22022d8d847)

PR #6304, opened 2017-02-02
More strict testing of handshake behaviour

normanmaurer:
Motiviation:

We should ensure we not need any extra wrap / unwrap calls during handshake once the handshake was signaled as finished

Modifications:

More strict testing

Result:

Better testing of handshake behaviour


normanmaurer:
Cherry-picked into 4.1 (1a05463) and 4.0 (e090178)

PR #6308, opened 2017-02-03
Do not prefer empty MAC address

jasontedor:
Motivation:

When comparing MAC addresses searching for the best MAC address, if locally-administered address (e.g., from a Docker container) is compared against an empty MAC address, the empty MAC address will be marked as preferred. In cases this is the only available MAC address, this leaves Netty using a random machine ID instead of using a perfectly valid machine ID from the locally-adminstered address.

Modifications:

This commit modifies the MAC address logic so that the empty MAC address is not preferred over a locally-administered address. This commit also simplifies the comparison logic here.

Result:

Empty MAC addresses will not be preferred over locally-administered addresses thus permitting the default machine ID to be the locally-adminstered MAC address if it is the only available MAC address.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/MacAddressUtil.java
+++ b/common/src/main/java/io/netty/util/internal/MacAddressUtil.java
@@ -203,7 +203,8 @@ public final class MacAddressUtil {
     /**
      * @return positive - current is better, 0 - cannot tell from MAC addr, negative - candidate is better.
      */
-    private static int compareAddresses(byte[] current, byte[] candidate) {
+    // visible for testing
+    static int compareAddresses(byte[] current, byte[] candidate) {
         if (candidate == null || candidate.length < EUI48_MAC_ADDRESS_LENGTH) {
             return 1;
         }
@@ -227,20 +228,23 @@ public final class MacAddressUtil {
         }
 
         // Prefer globally unique address.
-        if (current.length == 0 || (current[0] & 2) == 0) {
-            if ((candidate[0] & 2) == 0) {
+        if ((candidate[0] & 2) == 0) {
+            if (current.length != 0 && (current[0] & 2) == 0) {
                 // Both current and candidate are globally unique addresses.
                 return 0;
             } else {
+                // Only candidate is globally unique.
+                return -1;
+            }
+        } else {
+            if (current.length != 0 && (current[0] & 2) == 0) {
                 // Only current is globally unique.
                 return 1;
+            } else {
+                // Both current and candidate are non-unique.
+                return 0;
             }
-        } else if ((candidate[0] & 2) == 0) {
-            // Only candidate is globally unique.
-            return -1;
         }
-        // Both current and candidate are non-unique.
-        return 0;
     }
 
     /**
--- a/common/src/test/java/io/netty/util/internal/MacAddressUtilTest.java
+++ b/common/src/test/java/io/netty/util/internal/MacAddressUtilTest.java
@@ -17,10 +17,57 @@ package io.netty.util.internal;
 
 import org.junit.Test;
 
+import static io.netty.util.internal.EmptyArrays.EMPTY_BYTES;
 import static io.netty.util.internal.MacAddressUtil.parseMAC;
 import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
 
 public class MacAddressUtilTest {
+    @Test
+    public void testCompareAddresses() {
+        // should not prefer empty address when candidate is not globally unique
+        assertEquals(
+                0,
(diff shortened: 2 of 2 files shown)


jasontedor:
This problem was introduced by #6147.

common/src/main/java/io/netty/util/internal/MacAddressUtil.java
```diff
@@ -203,7 +203,7 @@ public static String formatAddress(byte[] addr) {
     /**
      * @return positive - current is better, 0 - cannot tell from MAC addr, negative - candidate is better.
      */
-    private static int compareAddresses(byte[] current, byte[] candidate) {
+    static int compareAddresses(byte[] current, byte[] candidate) {
```
normanmaurer: @jasontedor maybe add a comment that this is package private for testing.

common/src/main/java/io/netty/util/internal/MacAddressUtil.java
```diff
@@ -227,20 +227,23 @@ private static int compareAddresses(byte[] current, byte[] candidate) {
...
 
         // Prefer globally unique address.
-        if (current.length == 0 || (current[0] & 2) == 0) {
-            if ((candidate[0] & 2) == 0) {
+        if ((candidate[0] & 2) == 0) {
+            if (current.length != 0 && (current[0] & 2) == 0) {
                 // Both current and candidate are globally unique addresses.
                 return 0;
             } else {
                 // Only current is globally unique.
```
louxiu: Could it should be? PTAL
```
// Only candidate is globally unique.
return -1;
```

common/src/main/java/io/netty/util/internal/MacAddressUtil.java
```diff
@@ -227,20 +227,23 @@ private static int compareAddresses(byte[] current, byte[] candidate) {
...
                 return 1;
             }
-        } else if ((candidate[0] & 2) == 0) {
-            // Only candidate is globally unique.
-            return -1;
+        } else {
+            if (current.length != 0 && (current[0] & 2) == 0) {
+                // Only candidate is globally unique.
+                return -1;
```
louxiu:
```
// Only current is globally unique.
return 0;
```

common/src/main/java/io/netty/util/internal/MacAddressUtil.java
jasontedor: @normanmaurer I pushed 6c961704764ee994eff9938a41b86a9e0aa35f5c.

common/src/main/java/io/netty/util/internal/MacAddressUtil.java
jasontedor: Yes, those got swapped when I reversed the conditions. I pushed 0993c6282b55bcd7aff3fed6b3b79cd0dea59c3e which added more tests too. Thanks.

common/src/main/java/io/netty/util/internal/MacAddressUtil.java
jasontedor: I pushed 0993c6282b55bcd7aff3fed6b3b79cd0dea59c3e.

jasontedor:
@normanmaurer Thanks for reviewing, I think this is ready for you to take another look.

normanmaurer:
@Scottmitch PTAL and merge once satisfied

jasontedor:
Thanks @normanmaurer. I've squashed in preparation for integration.

normanmaurer:
@jasontedor thanks... @Scottmitch will take care of the rest 👍

Scottmitch:
4.1 (42c0359)

@jasobrown - Sorry for the breakage ... thanks for the fix & unit tests!

jasontedor:
Thank you @Scottmich (and no worries, breakage happens, especially when there are no unit tests).

normanmaurer:
@jasontedor doh, that hurts ... but yes you are right, no unit test == bad 👎

jasontedor:
@normanmaurer Sorry, I was not trying to make a hurtful comment, but rather an empathetic one.

normanmaurer:
@jasontedor haha all good :) Truth hurts, no worries ;)

Scottmitch:
@jasontedor - No harm done ... unit tests FTW.

farrukh-obaid:
@normanmaurer @jasontedor 
Hi i am getting same error when deployed on virtual linux container I am using Vertx framework latest version which uses Netty 4.1.8 final.
I just wanted to know is this warning can create problems for me and will my code work with this warning?

jasontedor:
> I just wanted to know is this warning can create problems for me and will my code work with this warning?

It's mostly annoying but harmless.

PR #6310, opened 2017-02-03
Cleanup : 'for each' loops for arrays and removed unnecessary toLowerCase()

doom369:
Motivation:

1. Make code easier to read with `for each` loops instead of `for (int i )`.
2. Simplify old code.

Modification:

1. `for each` loops introduced instead of `for (int i )` for arrays only (no iterator allocation).
2. Removed unnecessary `valueOf` and `toLowerCase` in `SslHandler.ignoreException` as `Pattern IGNORABLE_ERROR_MESSAGE` is `Pattern.CASE_INSENSITIVE`.

Result:

Code easier to read. No unnecessary operations.


CodingFabian:
looks good to me.

normanmaurer:
Cherry-picked into 4.0 (d576fca0ddbf99d8d8ea928b8ca39efb8c604b68) and 4.1 (b9abd3c9fc90cda1781a6f2f181735460a13b75d).

Thanks!

PR #6311, opened 2017-02-04
SslHandler avoid calling wrap/unwrap when unnecessary

Scottmitch:
Motivation:
The SSLEngine wrap and unwrap methods can be called in a way that has no side effects, but this could involve costly validation and allocation. The SslHandler should avoid calling into these methods if possible.

Modifications:
- wrapNonAppData should provide additional status which can be used by wrap to breakout early if possible

Result:
SslHandler invokes the SSLEngine less.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -725,8 +725,13 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
         }
     }
 
-    // This method will not call setHandshakeFailure(...) !
-    private void wrapNonAppData(ChannelHandlerContext ctx, boolean inUnwrap) throws SSLException {
+    /**
+     * This method will not call
+     * {@link #setHandshakeFailure(io.netty.channel.ChannelHandlerContext, Throwable, boolean)} or
+     * {@link #setHandshakeFailure(io.netty.channel.ChannelHandlerContext, Throwable)}.
+     * @return {@code true} if this method ends on {@link SSLEngineResult.HandshakeStatus#NOT_HANDSHAKING}.
+     */
+    private boolean wrapNonAppData(ChannelHandlerContext ctx, boolean inUnwrap) throws SSLException {
         ByteBuf out = null;
         ByteBufAllocator alloc = ctx.alloc();
         try {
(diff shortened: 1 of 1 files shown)


normanmaurer:
@Scottmitch please check unit test failures.

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1113,6 +1117,7 @@ private boolean unwrap(
         try {
             // Only continue to loop if the handler was not removed in the meantime.
             // See https://github.com/netty/netty/issues/5860
+            unwrapLoop:
             while (!ctx.isRemoved()) {
```
normanmaurer: @Scottmitch can you put this in front of the while just to make it more clear:

```
unwrapLoop: while(....)
```

Scottmitch: sure done

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -725,8 +725,11 @@ private void finishWrap(ChannelHandlerContext ctx, ByteBuf out, ChannelPromise p
         }
     }
 
-    // This method will not call setHandshakeFailure(...) !
-    private void wrapNonAppData(ChannelHandlerContext ctx, boolean inUnwrap) throws SSLException {
+    /**
+     * This method will not call setHandshakeFailure(...) !
```
normanmaurer: @Scottmitch as you changed this to java docs please use `{@link #setHandshakeFailure(.....`

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1163,7 +1167,9 @@ private boolean unwrap(
                     case NEED_UNWRAP:
                         break;
                     case NEED_WRAP:
-                        wrapNonAppData(ctx, true);
+                        if (wrapNonAppData(ctx, true) && length == 0) {
```
normanmaurer: can you add some comments why its ok to break the loop in this case ?

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1196,7 +1202,9 @@ private boolean unwrap(
                             flushedBeforeHandshake = false;
                             wrapLater = true;
                         }
-
+                        if (length == 0) {
```
normanmaurer: can you add some comments why its ok to break the loop in this case ?

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
Scottmitch: done

normanmaurer:
@Scottmitch ship it

Scottmitch:
4.1 (6353c22) 4.0 (ab65b1e)

PR #6315, opened 2017-02-06
Try to detect the cacheline size automatically

normanmaurer:
Motivation:

For max performance it may be worthwhile to allocate buffers from PooledByteBufAllocator alligned by the cacheline size.  In previous code-change we allowed to do so but the users needs to manually specify the cacheline size. On linux and osx (macos) its possible to detect it automatically and so let the user use -1 to signal it should be automatically detected.

Modifications:

- Add new code in PlatformDependent which allows detecting the cacheline size on a system
- Allow -1 as value when specify the cacheline size in PooledByteBufAllocator which will try to detect it automatically then.

Result:

Easier configuration of PooledByteBufAllocator.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent.java
@@ -31,6 +31,7 @@ import org.jctools.util.UnsafeAccess;
 
 import java.io.BufferedReader;
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.lang.reflect.Field;
@@ -80,6 +81,9 @@ public final class PlatformDependent {
 
     private static final boolean IS_ANDROID = isAndroid0();
     private static final boolean IS_WINDOWS = isWindows0();
+    private static final boolean IS_LINUX = isLinux0();
+    private static final boolean IS_OSX = isOsx0();
+
     private static volatile Boolean IS_ROOT;
 
     private static final int JAVA_VERSION = javaVersion0();
@@ -106,6 +110,8 @@ public final class PlatformDependent {
     private static final int BIT_MODE = bitMode0();
 
     private static final int ADDRESS_SIZE = addressSize0();
+    private static final int CACHE_LINE_SIZE = cacheLineSize0();
+
     private static final boolean USE_DIRECT_BUFFER_NO_CLEANER;
     private static final AtomicLong DIRECT_MEMORY_COUNTER;
     private static final long DIRECT_MEMORY_LIMIT;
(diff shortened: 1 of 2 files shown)


normanmaurer:
@Scottmitch @kiril-me PTAL

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -899,6 +905,90 @@ public static ClassLoader getSystemClassLoader() {
...
+            public Integer run() {
+                if (IS_LINUX) {
+                    BufferedReader in = null;
+                    try {
+                        in = new BufferedReader(new InputStreamReader(new FileInputStream("/proc/cpuinfo")));
+                        String line;
+                        while ((line = in.readLine()) != null) {
+                            if (line.startsWith("cache_alignment")) {
+                                String[] parts = line.split(":");
+                                return Integer.parseInt(parts[1].trim());
```
netkins: MAJOR Method io.netty.util.internal.PlatformDependent$1.run() converts String to boxed primitive using excessive boxing [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=fb-contrib%3ANAB_NEEDLESS_BOXING_VALUEOF)

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -899,6 +905,90 @@ public static ClassLoader getSystemClassLoader() {
...
+                                logger.debug("Failed to retrieve cache line size," +
+                                        " 'sysctl hw.cachelinesize' did exit with code " + exitCode);
+                                return -1;
+                            }
+                        } catch (InterruptedException e) {
+                            Thread.currentThread().interrupt();
+                        }
+                        if (line != null) {
+                            String[] parts = line.split(":");
+                            return Integer.parseInt(parts[1].trim());
```
netkins: MAJOR Method io.netty.util.internal.PlatformDependent$1.run() converts String to boxed primitive using excessive boxing [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=fb-contrib%3ANAB_NEEDLESS_BOXING_VALUEOF)

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -899,6 +905,90 @@ public static ClassLoader getSystemClassLoader() {
...
+                    }
+                } else if (IS_OSX) {
+                    Process p = null;
+                    BufferedReader in = null;
+                    try {
+                        p = Runtime.getRuntime().exec(new String[] { "sysctl", "hw.cachelinesize" });
+                        in = new BufferedReader(new InputStreamReader(p.getInputStream(), CharsetUtil.US_ASCII));
+                        String line = in.readLine();
+                        in.close();
+                        try {
```
netkins: MAJOR Extract this nested try block into a separate method. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1141)

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -899,6 +905,90 @@ public static ClassLoader getSystemClassLoader() {
...
+    }
+
+    private static int cacheLineSize0() {
+        return AccessController.doPrivileged(new PrivilegedAction<Integer>() {
+            @Override
+            public Integer run() {
+                if (IS_LINUX) {
+                    BufferedReader in = null;
+                    try {
+                        in = new BufferedReader(new InputStreamReader(new FileInputStream("/proc/cpuinfo")));
```
netkins: BLOCKER Close this "FileInputStream". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2095)

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -899,6 +905,90 @@ public static ClassLoader getSystemClassLoader() {
...
+                        if (p != null) {
+                            try {
+                                p.destroy();
+                            } catch (Exception ignore) {
+                                // Just ignore
+                            }
+                        }
+                    }
+                }
+                logger.debug("Failed to retrieve cache line size");
```
fenik17: Failed or non linux / osx.

kiril-me: Which Linux distribution do you use? Could you print /proc/cpuinfo ?

fenik17: @kiril-me it's question for me?
I mean what it will be printed on the Windows or other than linux/osx systems. But it's not a failure.

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -899,6 +905,90 @@ public static ClassLoader getSystemClassLoader() {
...
+                            } catch (IOException ignore) {
+                                // ignore on close
+                            }
+                        }
+                    }
+                } else if (IS_OSX) {
+                    Process p = null;
+                    BufferedReader in = null;
+                    try {
+                        p = Runtime.getRuntime().exec(new String[] { "sysctl", "hw.cachelinesize" });
```
Scottmitch: reminder ... we should share code here https://github.com/netty/netty/pull/6000/files#diff-a94053a0586a20f4f013bcaae8c3942eR314

Scottmitch: I've been hesitant about executing another process in the `kqueue` PR ... but not sure if there is another/better way to get this info.

normanmaurer: @Scottmitch yeah I did not find any better way for OSX / macOS , we could also only support Linux for now. WDYT ?

Scottmitch: I'm almost waiting for someone to yell at us and say "don't do that". What if someone's path is manipulated and some malicious program is run, etc... I just wish there was a "better way" to get the data. Maybe we can make it optional ... thats the approach I took in kqueue https://github.com/netty/netty/pull/6000/files#diff-a94053a0586a20f4f013bcaae8c3942eR275

normanmaurer: @Scottmitch yeah... starting programs is super awful :(

johnou: well the Linux approach is "ok" as it just reads a file, but what I can say is that all of our applications outright deny checkExec in a security manager to combat possible remote execution vulnerabilities. What about introducing a small jni utility to provide that information? https://github.com/NickStrupat/CacheLineSize

johnou: @normanmaurer if I go ahead and create some utility library that provides lookup of cache line size and maybe some stuff for kqueue is that something you'd be willing to pull into Netty as a shaded dependency? Catch is you need to come up with a name for the library.

Scottmitch: @johnou - Thanks for the tip! I would prefer if we can just add something to transport-native-epoll and transport-native-kqueue as opposed to introducing a new JNI library. Not sure if this is possible though given that `common` cannot depend on these modules.

normanmaurer: IMHO common should not have this dependency... maybe its ok to just only do it for linux for now. Or just not do it at all.

normanmaurer:
Let me just close this for now... I think its not super important anyway

PR #6327, opened 2017-02-07
Add a constant for Cookie "undefined maxAge"

slandelle:
Motivation:

DefaultCookie currently used an undocumented magic value for undefined
maxAge.

Clients need to be able to identify such value so they can implement a
proper CookieJar.

Ideally, we should add a `Cookie::isMaxAgeDefined` method but I guess
we can’t add a new method without breaking API :(

Modifications:

Add a new constant on `Cookie` interface so clients can use it to
compare with value return by `Cookie.maxAge` and decide if `maxAge` was
actually defined.

Result:

Clients have a better documented way to check if the maxAge attribute
was defined.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/cookie/Cookie.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/cookie/Cookie.java
@@ -22,6 +22,11 @@ package io.netty.handler.codec.http.cookie;
 public interface Cookie extends Comparable<Cookie> {
 
     /**
+     * Constant for undefined MaxAge attribute value.
+     */
+    long UNDEFINED_MAX_AGE = Long.MIN_VALUE;
+
+    /**
      * Returns the name of this {@link Cookie}.
      *
      * @return The name of this {@link Cookie}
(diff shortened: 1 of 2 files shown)


normanmaurer:
LGTM... @Scottmitch WDYT ?

Scottmitch:
this would be another great case for `default` methods ... JDK8 some day ...

The current behavior is not strictly defined in code but it is documented as part of the interface [1]. So technically to be compliant with this interface the `MIN_VALUE` concept should be honored. So if we do feel it is necessary to have another define for this we should also update the API javadocs ... which is logically the same interface so I think that is fine (@normanmaurer ?)

[1] https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/cookie/Cookie.java#L90

slandelle:
> we should also update the API javadocs

I missed this one, thanks! I can update.

> this would be another great case for default methods ... JDK8 some day ...

👍 At some point, "some day" will have to be defined...

normanmaurer:
@slandelle @Scottmitch yes sounds fine... Please make the changes.

slandelle:
@normanmaurer @Scottmitch Javadoc updated

Scottmitch:
That javadocs on `setMaxAge` also need to be updated

slandelle:
@Scottmitch Fixed

normanmaurer:
Cherry-picked into 4.1 (9d45f514a47ee8ad5259ee782fcca240017fc3a3) and 4.0 (c01ceb0e75fa5cb22350bea825e038e3bc301830)

normanmaurer:
@slandelle thx <3

slandelle:
Thanks :)

PR #6331, opened 2017-02-07
DnsNameResolver empty/null hostname missed by a416b79

Scottmitch:
Motivation:
a416b79 introduced a check for null or empty host name to be compatible with the JDK resolution. However the doResolve(String, Promise) method, and if the doResolve(String, DnsRecord[], Promise, DnsCache) method was overridden the empty/null hostname would not be correctly resolved.

Modifications:
- Move the empty/null host name check into the lowest level doResolve method in DnsNameResolver
- Remove the duplicate logic in InetNameResolver.java which can be bypassed anyways

Result:
By default (unless behavior is overridden) DnsNameResolver resolves null/empty host names to local host just like the JDK.


Scottmitch:
followup for PR https://github.com/netty/netty/pull/6258

resolver/src/main/java/io/netty/resolver/InetNameResolver.java
```diff
@@ -40,32 +37,15 @@
...
-    }
-
-    @Override
-    public Future<List<InetAddress>> resolveAll(String inetHost, Promise<List<InetAddress>> promise) {
-        if (inetHost == null || inetHost.isEmpty()) {
-            // If an empty hostname is used we should use "localhost", just like InetAddress.getByName(...) does.
-            return promise.setSuccess(Collections.singletonList(loopbackAddress()));
-        }
-        return super.resolveAll(inetHost, promise);
+        return SocketUtils.loopbackAddress();
```
normanmaurer: Any reason why you removed the caching of it ?

resolver/src/main/java/io/netty/resolver/InetNameResolver.java
```diff
@@ -40,32 +37,15 @@
      */
     protected InetNameResolver(EventExecutor executor) {
         super(executor);
-        loopbackAddress = SocketUtils.loopbackAddress();
     }
 
     /**
+     * @deprecated Use {@link SocketUtils#loopbackAddress()}.
```
normanmaurer: @Scottmitch I think this is a bad advice as `SocketUtils` is in the internal package.

resolver/src/main/java/io/netty/resolver/InetNameResolver.java
```diff
@@ -40,32 +37,15 @@
...
      * Returns the {@link InetAddress} for loopback.
      */
+    @Deprecated
     protected InetAddress loopbackAddress() {
-        return loopbackAddress;
-    }
-
-    @Override
-    public Future<InetAddress> resolve(String inetHost, Promise<InetAddress> promise) {
```
normanmaurer: @Scottmitch didn't we say we want to have the logic in `InetNameResolver` to make it easier to have a consistent behaviour when write a new implementation ?

resolver/src/main/java/io/netty/resolver/InetNameResolver.java
Scottmitch: This method should be an internal detail of `DnsNameResolver` and shouldn't be publicly exposed at this level. I would like to remove it but my assumption is this would be an API breaking change ... it is not clear if I can make API changes to this package because we forgot to put `@UnsableApi` annotations. wdyt?

resolver/src/main/java/io/netty/resolver/InetNameResolver.java
Scottmitch: Yes this did come up https://github.com/netty/netty/pull/6258#discussion_r97187696 but I'm not convinced this adds value as we will have to check for it at the lower layer anyways. My new train of though is to keep the scope limited to where we actually need to enforce the constraint instead of having it spread out in different places. The spread out nature was one of the reasons why this bug exists in the first place.

resolver/src/main/java/io/netty/resolver/InetNameResolver.java
Scottmitch: I originally wanted to remove this method but wasn't sure if this would be OK (see https://github.com/netty/netty/pull/6331#discussion_r100219110). I don't think it should be used but if we can't remove the method and you think it may be used then I can bring the caching back.

resolver/src/main/java/io/netty/resolver/InetNameResolver.java
Scottmitch: https://github.com/netty/netty/pull/6334 is pushing in the direction of marking the packages as `@UnstableApi` ... I like this. Unless you have strong objections I will follow that approach here. We can also discuss a followup PR to put in the `@UnstableApi` on the remainder of the package.

resolver/src/main/java/io/netty/resolver/InetNameResolver.java
normanmaurer: ok fine... Just wanted to ensure we are aware of it :). Shiiiiiiiiip it!

Scottmitch:
4.1 (007048d)

PR #6332, opened 2017-02-08
Ensure we release the previous retained AddressedEnvelope when we fai…

normanmaurer:
…l to notify the promise.

Motivation:

We need to ensure we release the AddressedEnvelope if we fail to notify the future (as it may be notified before because of an timeout). Otherwise we may leak.

Modifications:

Call release() if we fail to notify the future.

Result:

No memory leaks possible.


normanmaurer:
Cherry-picked into 4.1 (3462a86)

PR #6334, opened 2017-02-08
HostsFileParser should allow both IPv4 and IPv6 for a given host

slandelle:
Motivation:

HostsFileParser only retains the first address for each given hostname.
This is wrong, and it’s allowed to have both an IPv4 and an IPv6.

Modifications:

* Have `HostsFileParser` now return a  `HostsFileEntries` that contains
IPv4 entries and IPv6 entries
* Introduce `ResolvedAddressTypes` to describe resolved address types
preferences, as `InternetProtocolFamily` is locked in the
« transport »  module
* Add a new `ResolvedAddressTypes` parameter to
`HostsFileEntriesResolver::address` to account for address types
preferences

Result:

We now support hosts files that contains IPv4 and IPv6 pairs for a same
hostname.

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
@@ -39,6 +39,7 @@ import io.netty.handler.codec.dns.DnsRecordType;
 import io.netty.handler.codec.dns.DnsResponse;
 import io.netty.resolver.HostsFileEntriesResolver;
 import io.netty.resolver.InetNameResolver;
+import io.netty.resolver.ResolvedAddressTypes;
 import io.netty.util.NetUtil;
 import io.netty.util.ReferenceCountUtil;
 import io.netty.util.concurrent.FastThreadLocal;
@@ -56,13 +57,10 @@ import java.net.IDN;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
-import java.util.LinkedHashSet;
 import java.util.List;
-import java.util.Set;
 
 import static io.netty.util.internal.ObjectUtil.*;
 
@@ -75,24 +73,37 @@ public class DnsNameResolver extends InetNameResolver {
     private static final InternalLogger logger = InternalLoggerFactory.getInstance(DnsNameResolver.class);
     private static final String LOCALHOST = "localhost";
     private static final InetAddress LOCALHOST_ADDRESS;
-    private static final DnsRecord[] EMTPY_ADDITIONALS = new DnsRecord[0];
+    private static final DnsRecord[] EMPTY_ADDITIONALS = new DnsRecord[0];
+    private static final DnsRecordType[] IPV4_ONLY_RESOLVED_RECORD_TYPES =
+            new DnsRecordType[] {DnsRecordType.A};
+    private static final InternetProtocolFamily[] IPV4_ONLY_RESOLVED_PROTOCOL_FAMILIES =
+            new InternetProtocolFamily[] {InternetProtocolFamily.IPv4};
+    private static final DnsRecordType[] IPV4_PREFERRED_RESOLVED_RECORD_TYPES =
+            new DnsRecordType[] {DnsRecordType.A, DnsRecordType.AAAA};
+    private static final InternetProtocolFamily[] IPV4_PREFERRED_RESOLVED_PROTOCOL_FAMILIES =
+            new InternetProtocolFamily[] {InternetProtocolFamily.IPv4, InternetProtocolFamily.IPv6};
+    private static final DnsRecordType[] IPV6_ONLY_RESOLVED_RECORD_TYPES =
+            new DnsRecordType[] {DnsRecordType.AAAA};
+    private static final InternetProtocolFamily[] IPV6_ONLY_RESOLVED_PROTOCOL_FAMILIES =
+            new InternetProtocolFamily[] {InternetProtocolFamily.IPv6};
+    private static final DnsRecordType[] IPV6_PREFERRED_RESOLVED_RECORD_TYPES =
+            new DnsRecordType[] {DnsRecordType.AAAA, DnsRecordType.A};
+    private static final InternetProtocolFamily[] IPV6_PREFERRED_RESOLVED_PROTOCOL_FAMILIES =
+            new InternetProtocolFamily[] {InternetProtocolFamily.IPv6, InternetProtocolFamily.IPv4};
 
-    static final InternetProtocolFamily[] DEFAULT_RESOLVE_ADDRESS_TYPES;
-    static final String[] DEFAULT_SEACH_DOMAINS;
+    static final ResolvedAddressTypes DEFAULT_RESOLVE_ADDRESS_TYPES;
+    static final String[] DEFAULT_SEARCH_DOMAINS;
 
     static {
         if (NetUtil.isIpV4StackPreferred()) {
-            DEFAULT_RESOLVE_ADDRESS_TYPES = new InternetProtocolFamily[] { InternetProtocolFamily.IPv4 };
+            DEFAULT_RESOLVE_ADDRESS_TYPES = ResolvedAddressTypes.IPV4_ONLY;
             LOCALHOST_ADDRESS = NetUtil.LOCALHOST4;
         } else {
-            DEFAULT_RESOLVE_ADDRESS_TYPES = new InternetProtocolFamily[2];
             if (NetUtil.isIpV6AddressesPreferred()) {
-                DEFAULT_RESOLVE_ADDRESS_TYPES[0] = InternetProtocolFamily.IPv6;
-                DEFAULT_RESOLVE_ADDRESS_TYPES[1] = InternetProtocolFamily.IPv4;
+                DEFAULT_RESOLVE_ADDRESS_TYPES = ResolvedAddressTypes.IPV6_PREFERRED;
                 LOCALHOST_ADDRESS = NetUtil.LOCALHOST6;
             } else {
-                DEFAULT_RESOLVE_ADDRESS_TYPES[0] = InternetProtocolFamily.IPv4;
-                DEFAULT_RESOLVE_ADDRESS_TYPES[1] = InternetProtocolFamily.IPv6;
+                DEFAULT_RESOLVE_ADDRESS_TYPES = ResolvedAddressTypes.IPV4_PREFERRED;
                 LOCALHOST_ADDRESS = NetUtil.LOCALHOST4;
             }
         }
@@ -113,7 +124,7 @@ public class DnsNameResolver extends InetNameResolver {
             // Failed to get the system name search domain list.
             searchDomains = EmptyArrays.EMPTY_STRINGS;
         }
-        DEFAULT_SEACH_DOMAINS = searchDomains;
+        DEFAULT_SEARCH_DOMAINS = searchDomains;
     }
 
     private static final DatagramDnsResponseDecoder DECODER = new DatagramDnsResponseDecoder();
@@ -145,7 +156,8 @@ public class DnsNameResolver extends InetNameResolver {
     private final long queryTimeoutMillis;
     private final int maxQueriesPerResolve;
     private final boolean traceEnabled;
-    private final InternetProtocolFamily[] resolvedAddressTypes;
+    private final ResolvedAddressTypes resolvedAddressTypes;
+    private final InternetProtocolFamily[] resolvedInternetProtocolFamilies;
     private final boolean recursionDesired;
     private final int maxPayloadSize;
     private final boolean optResourceEnabled;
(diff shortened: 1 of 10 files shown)


resolver/src/main/java/io/netty/resolver/HostsFileEntries.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * A container of hosts file entries
+ */
+@UnstableApi
+public class HostsFileEntries {
+
+    /**
+     * Empty entries
+     */
+    public static final HostsFileEntries EMPTY =
```
fenik17: Is really needed caching empty entries? How many HostsFileEntries can be created?

slandelle: I guess we could live without. But it doesn't hurt either. I can remove if there's a consensus in this direction.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -308,6 +311,22 @@ public void operationComplete(ChannelFuture future) throws Exception {
         });
     }
 
+    private static ResolvedAddressTypes resolvedAddressTypes(InternetProtocolFamily[] internetProtocolFamilies) {
```
netkins: MINOR Consider using varargs for methods or constructors which take an array the last parameter. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUseVarargs)

slandelle:
@normanmaurer SonarQube failed the build, but the issue it reported is nonsensical (use varargs for resolvedAddressTypes parameters in `DnsNameResolver`).

resolver/src/main/java/io/netty/resolver/HostsFileEntries.java
slandelle: @fenik17 Actually, it's used in many places in `HostsFileParser` so a constant is indeed helpful.

resolver/src/main/java/io/netty/resolver/ResolvedAddressTypes.java
```diff
@@ -0,0 +1,40 @@
...
+ * under the License.
+ */
+package io.netty.resolver;
+
+/**
+ * Defined resolved address types.
+ * Ideally, we should be using "InternetProtocolFamily"
+ * but this class sits in the "transport" module so it can't be used here.
+ */
+public enum ResolvedAddressTypes {
```
normanmaurer: Rename to `ResolvedAddressType` ?

resolver/src/main/java/io/netty/resolver/ResolvedAddressTypes.java
```diff
@@ -0,0 +1,40 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.resolver;
+
+/**
+ * Defined resolved address types.
+ * Ideally, we should be using "InternetProtocolFamily"
+ * but this class sits in the "transport" module so it can't be used here.
```
normanmaurer: remove the above sentence.

resolver/src/main/java/io/netty/resolver/HostsFileEntries.java
```diff
@@ -0,0 +1,62 @@
...
+import java.net.Inet4Address;
+import java.net.Inet6Address;
+import java.util.Collections;
+import java.util.Map;
+
+/**
+ * A container of hosts file entries
+ */
+@UnstableApi
+public class HostsFileEntries {
```
normanmaurer: final

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -246,6 +248,7 @@ public DnsNameResolver(
         this.nameServerAddresses = checkNotNull(nameServerAddresses, "nameServerAddresses");
         this.queryTimeoutMillis = checkPositive(queryTimeoutMillis, "queryTimeoutMillis");
         this.resolvedAddressTypes = checkNonEmpty(resolvedAddressTypes, "resolvedAddressTypes");
+        this.resolvedAddressTypesEnum = resolvedAddressTypes(resolvedAddressTypes);
```
normanmaurer: did you consider changing the constructor argument to ResolvedAddressTypes ?

resolver/src/main/java/io/netty/resolver/ResolvedAddressTypes.java
slandelle: The plural sounds better to me, as it's a kind of strategy. Maybe a native english speaker would know better? @Scottmitch ?

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
slandelle: Yes, but I first wanted to see if you agreed with my `ResolvedAddressTypes` enum.
This enum looks better to me than the current "no empty and of max size 2" array, so I'd be indeed tempted to change the constructor.
Then, there's already a deprecated constructor there. Is it worth keeping deprecated stuff around as the class is properly flagged as unstable?

normanmaurer: I kept it when possible just so we not break user code if we can easily not do... so if possible yes please keep it if not possible break it

slandelle:
@normanmaurer PTAL again. Note that I fixed a few typos in `DnsNameResolver` constants.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -241,11 +241,57 @@ public DnsNameResolver(
...
+     * @param traceEnabled if trace is enabled
+     * @param maxPayloadSize the capacity of the datagram packet buffer
+     * @param optResourceEnabled if automatic inclusion of a optional records is enabled
+     * @param hostsFileEntriesResolver the {@link HostsFileEntriesResolver} used to check for local aliases
+     * @param searchDomains the list of search domain
+     * @param ndots the ndots value
+     * @param decodeIdn {@code true} if domain / host names should be decoded to unicode when received.
+     *                        See <a href="https://tools.ietf.org/html/rfc3492">rfc3492</a>.
+     */
+    public DnsNameResolver(
```
netkins: MAJOR Constructor has 16 parameters, which is greater than 7 authorized. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS00107)

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -75,24 +76,21 @@
     private static final InternalLogger logger = InternalLoggerFactory.getInstance(DnsNameResolver.class);
     private static final String LOCALHOST = "localhost";
     private static final InetAddress LOCALHOST_ADDRESS;
-    private static final DnsRecord[] EMTPY_ADDITIONALS = new DnsRecord[0];
+    private static final DnsRecord[] EMPTY_ADDITIONALS = new DnsRecord[0];
 
-    static final InternetProtocolFamily[] DEFAULT_RESOLVE_ADDRESS_TYPES;
-    static final String[] DEFAULT_SEACH_DOMAINS;
+    static final ResolvedAddressTypes DEFAULT_RESOLVE_ADDRESS_TYPES;
+    static final String[] DEFAULT_SEARCH_DOMAINS;
```
netkins: CRITICAL Make this array "private". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1873)

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -308,6 +363,28 @@ public void operationComplete(ChannelFuture future) throws Exception {
         });
     }
 
+    static ResolvedAddressTypes computeResolvedAddressTypes(InternetProtocolFamily[] internetProtocolFamilies) {
```
netkins: MINOR Consider using varargs for methods or constructors which take an array the last parameter. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUseVarargs)

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -224,6 +223,7 @@ public DnsNameResolver(
      * @param decodeIdn {@code true} if domain / host names should be decoded to unicode when received.
      *                        See <a href="https://tools.ietf.org/html/rfc3492">rfc3492</a>.
      */
+    @Deprecated
```
Scottmitch: this class is marked as unstable. my preference would be to avoid adding deprecated methods if possible. also most users will likely use the builder to construct this object always.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -258,32 +304,41 @@ public DnsNameResolver(
...
-                    break;
-                default:
-                    throw new Error();
-            }
+        switch (resolvedAddressTypes) {
+            case IPV4_ONLY:
+                supportsAAAARecords = false;
+                supportsARecords = true;
+                resolveRecordTypes = new DnsRecordType[] {DnsRecordType.A};
+                resolvedInternetProtocolFamilies = new InternetProtocolFamily[] {InternetProtocolFamily.IPv4};
```
Scottmitch: can these arrays be static to avoid object allocation?

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverBuilder.java
```diff
@@ -206,7 +185,9 @@ public DnsNameResolverBuilder resolvedAddressTypes(InternetProtocolFamily... res
      *
      * @param resolvedAddressTypes the address types
      * @return {@code this}
+     * @deprecated use {@link #resolvedAddressTypes(ResolvedAddressTypes)} instead
      */
+    @Deprecated
```
Scottmitch: this class is marked as unstable ... should we make the conversion routine `computeResolvedAddressTypes` available somewhere instead and call it out in the comments for the new method?

Scottmitch: the duplicate entry removal seems unnecessary here too (could be considered an application error) so I'm not sure we loose much by removing this.

resolver/src/main/java/io/netty/resolver/HostsFileEntries.java
Scottmitch: can we make it package private for now?

resolver/src/main/java/io/netty/resolver/HostsFileParser.java
```diff
@@ -153,14 +157,20 @@ private static File locateHostsFile() {
                 for (int i = 1; i < lineParts.size(); i ++) {
                     String hostname = lineParts.get(i);
                     String hostnameLower = hostname.toLowerCase(Locale.ENGLISH);
-                    if (!entries.containsKey(hostnameLower)) {
-                        // trying to map a host to multiple IPs is wrong
-                        // only the first entry is honored
-                        entries.put(hostnameLower, InetAddress.getByAddress(hostname, ipBytes));
+                    InetAddress address = InetAddress.getByAddress(hostname, ipBytes);
+                    if (address instanceof Inet4Address) {
+                        // Map::putIfAbsent is Java8 :(
```
Scottmitch: it is sad but we can remove this comment :)

We could also optimistically put and restore ... although I don't think it is a big deal here either way.

resolver/src/main/java/io/netty/resolver/ResolvedAddressTypes.java
```diff
@@ -0,0 +1,38 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.resolver;
+
+/**
+ * Defined resolved address types.
+ */
+public enum ResolvedAddressTypes {
```
Scottmitch: `@UnstableApi` (for now at least)

slandelle:
@Scottmitch Requested changes done.

normanmaurer:
@trustin PTAL as well.

slandelle:
@normanmaurer @Scottmitch I just pushed a change, allowing for passing a null `ResolvedAddressTypes` value to `DnsNameResolver` and then use the default value. This is useful for people who don't use the builder and directly extend `DnsNameResolver` (like me :) ).

normanmaurer:
@slandelle +1 makes sense.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -258,32 +227,39 @@ public DnsNameResolver(
...
-                    recordTypes.add(DnsRecordType.A);
-                    break;
-                case IPv6:
-                    supportsAAAARecords = true;
-                    recordTypes.add(DnsRecordType.AAAA);
-                    break;
-                default:
-                    throw new Error();
-            }
+        switch (resolvedAddressTypes) {
```
netkins: BLOCKER NullPointerException might be thrown as 'resolvedAddressTypes' is nullable here [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2259)

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -75,24 +73,37 @@
...
+            new DnsRecordType[] {DnsRecordType.AAAA};
+    private static final InternetProtocolFamily[] IPV6_ONLY_RESOLVED_PROTOCOL_FAMILIES =
+            new InternetProtocolFamily[] {InternetProtocolFamily.IPv6};
+    private static final DnsRecordType[] IPV6_PREFERRED_RESOLVED_RECORD_TYPES =
+            new DnsRecordType[] {DnsRecordType.AAAA, DnsRecordType.A};
+    private static final InternetProtocolFamily[] IPV6_PREFERRED_RESOLVED_PROTOCOL_FAMILIES =
+            new InternetProtocolFamily[] {InternetProtocolFamily.IPv6, InternetProtocolFamily.IPv4};
+
+    static final ResolvedAddressTypes DEFAULT_RESOLVE_ADDRESS_TYPES;
+    static final String[] DEFAULT_SEARCH_DOMAINS;
```
netkins: CRITICAL Make this array "private". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1873)

resolver/src/main/java/io/netty/resolver/HostsFileEntries.java
```diff
@@ -0,0 +1,62 @@
...
+            new HostsFileEntries(
+                    Collections.<String, Inet4Address>emptyMap(),
+                    Collections.<String, Inet6Address>emptyMap());
+
+    private final Map<String, Inet4Address> inet4Entries;
+    private final Map<String, Inet6Address> inet6Entries;
+
+    public HostsFileEntries(Map<String, Inet4Address> inet4Entries, Map<String, Inet6Address> inet6Entries) {
+        this.inet4Entries = Collections.unmodifiableMap(inet4Entries);
+        this.inet6Entries = Collections.unmodifiableMap(inet6Entries);
```
slandelle: good point, will do

resolver/src/main/java/io/netty/resolver/HostsFileParser.java
```diff
@@ -65,56 +68,57 @@ private static File locateHostsFile() {
...
      * @throws IOException file could not be read
      */
-    public static Map<String, InetAddress> parse(Reader reader) throws IOException {
+    public static HostsFileEntries parse(Reader reader) throws IOException {
         checkNotNull(reader, "reader");
         BufferedReader buff = new BufferedReader(reader);
         try {
-            Map<String, InetAddress> entries = new HashMap<String, InetAddress>();
+            Map<String, Inet4Address> ipv4Entries = new HashMap<String, Inet4Address>();
+            Map<String, Inet6Address> ipv6Entries = new HashMap<String, Inet6Address>();
```
slandelle: What would be the benefits? Wouldn't it just slow lookup down?

normanmaurer: Was just a thought... but I guess it's fine as it is

slandelle:
@normanmaurer I just pushed the hashmap copy in `HostsFileEntries`

normanmaurer:
@Scottmitch PTAL

Scottmitch:
4.1 (81f9de4)

Thanks @slandelle !

slandelle:
Thanks for the review!

PR #6340, opened 2017-02-09
Add exposeHeaders and allowedRequestHeaders that accept CharSequence, close #6328

### LINKED ISSUE #6328: CorsConfigBuilder exposeHeaders and allowedRequestHeaders should accept CharSequence
### Expected behavior

In Netty 4.1, `HttpHeaderNames` constants are `AsciiString`s. It should be possible to directly pass those constants to `CorsConfigBuilder` methods such as `exposeHeaders`, `allowedRequestHeaders`.

### Actual behavior

Those methods take `String`, so one has to convert and call `toString` on the `AsciiString`s.

### Netty version

4.1.8.Final

slandelle:
Motivation:

Netty 4.1 introduced AsciiString and defines HttpHeaderNames constants
as such.

It would be convenient to be able to pass them to `exposeHeaders` and
`allowedRequestHeaders` directly without having to call `toString`.

Modifications:

Add `exposeHeaders` and `allowedRequestHeaders` overloads that take a
`CharSequence`.

Result:

More convenient API

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/cors/CorsConfigBuilder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/cors/CorsConfigBuilder.java
@@ -149,6 +149,38 @@ public final class CorsConfigBuilder {
     }
 
     /**
+     * Specifies the headers to be exposed to calling clients.
+     *
+     * During a simple CORS request, only certain response headers are made available by the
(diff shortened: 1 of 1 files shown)


normanmaurer:
Cherry-picked into 4.1 (64abef5f5b283551f85fab4a1b20627275cffc3f)

Issue #6335, opened 2017-02-08
Remove OpenSSLEngine unwrap restriction


nmittler:
Creating this as a point of discussion. A caller of an `SSLEngine` should be able to pass a large buffer into `unwrap`, which it received from the network. 

However, we currently have [this check](https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java#L824) to make sure that the number of input bytes is <= a maximum packet size. 

This doesn't seem valid anymore since the new package processing logic was added [here](https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java#L873). IIUC, this logic should protect against any attacks related to the input buffer size.

This was discovered in a [Conscrypt bug](https://github.com/google/conscrypt/issues/76)

@normanmaurer thoughts?

normanmaurer:
@nmittler +1 ... Let me fix it.

nmittler:
@normanmaurer awesome, thanks! I think you just have to remove the check.

normanmaurer:
Yes and adding a Unit Test ;)

normanmaurer:
Fixed by https://github.com/netty/netty/pull/6350

PR #6350, opened 2017-02-13
Allow to unwrap ByteBuffer > MAX_ENCRYPTED_PACKET_LENGTH

normanmaurer:
Motivation:

We should remove the restriction to only allow to call unwrap with a ByteBuffer[] whose cumulative length exceeds MAX_ENCRYPTED_PACKET_LENGTH.

Modifications:

Remove guard.

Result:

Fixes [#6335].


normanmaurer:
Cherry-picked into 4.1 (adcde84) and 4.0 (d873adf)

Issue #6333, opened 2017-02-08
@Sharable CombinedChannelDuplexHandler usage problems


errandir:
Netty version: 4.1.8.Final

@Sharable CombinedChannelDuplexHandler implementation shares ChannelHandlerContext among different connections. Problem could be reproduced with [this code](https://gist.github.com/errandir/af45f3559421e3d12ace556af3e4999b) _(compile and run with netty-all-4.1.8.Final.jar in classpath)_. Output will differ from time to time.

For example I get:
<pre>
io.netty.channel.CombinedChannelDuplexHandler$1@3e234509: [1-1]
io.netty.channel.CombinedChannelDuplexHandler$1@3e328346: [1-2]
io.netty.channel.CombinedChannelDuplexHandler$1@3e328346: [2-1][2-2][2-3]
io.netty.channel.CombinedChannelDuplexHandler$1@3e328346: [2-4]
io.netty.channel.CombinedChannelDuplexHandler$1@3e328346: [1-3][1-4]
</pre>
but expect that `[1-*]` strings will be handled by the first handler (`3e234509`).

normanmaurer:
@errandir `CombinedChannelDuplexHandler` implementation can't be `@Sharable` as it contains state. Let me ensure we throw an exception if someone tries to make a sub-class of `CombinedChannelDuplexHandler ` `@Sharable`.

Thanks for reporting

errandir:
@normanmaurer There is real lack of such exception and even `CombinedChannelDuplexHandler` comment.

normanmaurer:
@errandir yep... will have a PR soon.

normanmaurer:
Fixed by #6352

PR #6352, opened 2017-02-13
Ensure CombinedChannelDuplexHandler can not be shared.

normanmaurer:
Motivation:

CombinedChannelDuplexHandler must not be shared as it contains state.

Modifications:

Enforce that it is not shared.

Result:

Fixes [#6333]


normanmaurer:
Cherry-picked into 4.1 (78586a9) and 4.0 (0a7cce243cc21e92e76df4cdf6d18c9e9206a0f6)

PR #6353, opened 2017-02-13
Two tests are missing `@Test` annotations

normanmaurer:
Motivation:

ReadOnlyByteBufTest contains two tests which are missing the `@Test` annotation and so will never run.

Modifications:

Add missing annotation.

Result:

Tests run as expected.


normanmaurer:
Cherry-picked into 4.1 (d8596d2)

PR #6354, opened 2017-02-13
Expose the chunkSize used by PooledByteBufAllocator.

normanmaurer:
Motivation:

Sometimes it may be useful to know the used chunkSize.

Modifications:

Add method to expose chunkSize.

Result:

More exposed details.


buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java
```diff
@@ -476,6 +477,13 @@ public int normalCacheSize() {
         return normalCacheSize;
     }
 
+    /**
+     * Return the chunk size for an arena.
+     */
+    public int chunkSize() {
```
Scottmitch: nit: `final` ?

normanmaurer: +1

normanmaurer:
Cherry-picked into 4.1 (f09a721) and 4.0 (6e834de)

PR #6355, opened 2017-02-13
Eliminate unnessary wrapping when call ByteBuf.asReadOnly() in some c…

normanmaurer:
…ases

Motivation:

We can eliminate unnessary wrapping when call ByteBuf.asReadOnly() in some cases to reduce indirection.

Modifications:

- Check if asReadOnly() needs to create a new instance or not
- Add test cases

Result:

Less object creation / wrapping.


normanmaurer:
Cherry-picked into 4.1 (371c0ca)

Issue #6345, opened 2017-02-10
Initialization of PlatformDependent0 fails on Java 9


mp911de:
Initialization of `PlatformDependent0` fails on Java 9 in static initializer when calling `setAccessible(true)`.

```
Field field = Buffer.class.getDeclaredField("address");
field.setAccessible(true);
```

### Expected behavior

Initialization should not fail without `--add-opens=java.base/java.nio=ALL-UNNAMED`

### Actual behavior

Class initialization fails with:

```
java.lang.reflect.InaccessibleObjectException: Unable to make field long java.nio.Buffer.address accessible: module java.base does not "opens java.nio" to unnamed module @598067a5
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:207)
	at java.base/java.lang.reflect.Field.checkCanSetAccessible(Field.java:171)
	at java.base/java.lang.reflect.Field.setAccessible(Field.java:165)
	at io.netty.util.internal.PlatformDependent0$1.run(PlatformDependent0.java:72)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at io.netty.util.internal.PlatformDependent0.<clinit>(PlatformDependent0.java:67)
	at io.netty.util.internal.PlatformDependent.getSystemClassLoader(PlatformDependent.java:888)
	at io.netty.util.internal.PlatformDependent.isAndroid0(PlatformDependent.java:905)
	at io.netty.util.internal.PlatformDependent.<clinit>(PlatformDependent.java:81)
	at io.netty.buffer.PooledByteBufAllocator.<clinit>(PooledByteBufAllocator.java:87)

```
### Steps to reproduce

Trigger class initialization of `PlatformDependent0` or call to `PlatformDependent.getContextClassLoader()`.

### Minimal yet complete reproducer code (or URL to code)

Variant 1:
```
PlatformDependent.getContextClassLoader()
```

Variant 2:
```
Version nettyBufferVersion = Version.identify().get("netty-buffer");
```
### Netty version

4.1.8

### JVM version (e.g. `java -version`)

Java(TM) SE Runtime Environment (build 9-ea+156)
Java HotSpot(TM) 64-Bit Server VM (build 9-ea+156, mixed mode)

### OS version (e.g. `uname -a`)

[...]

PR #6356, opened 2017-02-13
Initialization of PlatformDependent0 fails on Java 9

normanmaurer:
Motivation:

Initialization of PlatformDependent0 fails on Java 9 in static initializer when calling setAccessible(true).

Modifications:

Add RefelectionUtil which can be used to safely try if setAccessible(true) can be used or not and if not fail back to non reflection.

Result:

Fixed [#6345]

### DIFF
--- a/common/src/main/java/io/netty/util/internal/ReflectionUtil.java
+++ b/common/src/main/java/io/netty/util/internal/ReflectionUtil.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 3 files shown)


common/src/main/java/io/netty/util/internal/ReflectionUtil.java
```diff
@@ -0,0 +1,49 @@
...
+        } catch (RuntimeException e) {
+            return handleInaccessibleObjectException(e);
+        }
+    }
+
+    private static RuntimeException handleInaccessibleObjectException(RuntimeException e) {
+        // JDK 9 can throw an inaccessible object exception here; since Netty compiles
+        // against JDK 7 and this exception was only added in JDK 9, we have to weakly
+        // check the type
+        if (e.getClass().getName().equals("java.lang.reflect.InaccessibleObjectException")) {
```
netkins: MAJOR Move the "java.lang.reflect.InaccessibleObjectException" string literal on the left side of this string comparison. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1132)

Scottmitch: @normanmaurer +1 ... can't hurt

normanmaurer:
Addressed and cherry-picked into 4.1 (90bc60547784877366a3294e6b945a5241b7099f) and 4.0 (f365770)

PR #6357, opened 2017-02-13
Only try to calculate direct memory offset when sun.misc.Unsafe is pr…

normanmaurer:
…esent

Motivation:

We should only try to calculate the direct memory offset when sun.misc.Unsafe is present as otherwise it will fail with an NPE as PlatformDependent.directBufferAddress(...) will throw it.
This problem was introduced by 66b9be3a469a2cdcc5d18a8b94c679940ce002a9.

Modifications:

Use offset of 0 if no sun.misc.Unsafe is present.

Result:

PooledByteBufAllocator also works again when no sun.misc.Unsafe is present.


normanmaurer:
Cherry-picked into 4.1 (54339c0) and 4.0 (2d558b3c2ba48ad085e9b0d49c5a414445d1430e)

PR #6358, opened 2017-02-13
Ensure tests pass when sun.misc.Unsafe is not present

normanmaurer:
Motivation:

We need to ensure we pass all tests when sun.misc.Unsafe is not present.

Modifications:

- Make *ByteBufAllocatorTest work whenever sun.misc.Unsafe is present or not
- Let Lz4FrameEncoderTest not depend on AbstractByteBufAllocator implementation details which take into account if sun.misc.Unsafe is present or not

Result:

Tests pass even without sun.misc.Unsafe.

### DIFF
--- a/buffer/src/test/java/io/netty/buffer/AbstractByteBufAllocatorTest.java
+++ b/buffer/src/test/java/io/netty/buffer/AbstractByteBufAllocatorTest.java
@@ -15,6 +15,7 @@
  */
 package io.netty.buffer;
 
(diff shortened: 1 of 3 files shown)


normanmaurer:
Cherry-picked into 4.1 (9b2b3e251263f101fb594651659cb6aed01a4a42) and 4.0 (73af7c92151d939cfebcbf78d0335651aa7c5513)

Issue #6336, opened 2017-02-08
NPE in CipherSuiteConverter.toJava(CipherSuiteConverter.java:281)


rkapsi:
Providing repro shortly but it's triggered by an Unit Test for the OCSP Stapling PR.

Basically... OpenSSL (native) is calling my OCSP callback (java) which throws for the sake of the test an Exception. Instead of bubbling up the Exceptions it's running into an unhandled (?) NPE and my purposely thrown Exception gets swallowed somewhere.

Netty 4.1.9.Final-SNAPSHOT

The NPE's stack looks like this:

```java
io.netty.handler.codec.DecoderException: java.lang.NullPointerException
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:442)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:248)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1334)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
	at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:926)
	at io.netty.channel.local.LocalChannel.finishPeerRead0(LocalChannel.java:443)
	at io.netty.channel.local.LocalChannel.access$11(LocalChannel.java:424)
	at io.netty.channel.local.LocalChannel$5.run(LocalChannel.java:397)
	at io.netty.channel.DefaultEventLoop.run(DefaultEventLoop.java:54)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:858)
	at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:144)
	at java.lang.Thread.run(Thread.java:745)
[...]

PR #6363, opened 2017-02-13
CipherSuiteConverter NPE

Scottmitch:
Motivation:
CipherSuiteConverter may throw a NPE if a cipher suite from OpenSSL does not match the precomputed regular expression for OpenSSL ciphers. This method shouldn't throw and instead just return null.

Modifications:
- if cacheFromOpenSsl(..) fails the conversion toJava should return null

Result:
Fixes https://github.com/netty/netty/issues/6336.


Scottmitch:
@rkapsi - PTAL

rkapsi:
LGTM

Scottmitch:
4.1 (6765e9f) 4.0 (f7434c5)

PR #6364, opened 2017-02-13
Resolver package is unstable

Scottmitch:
Motivation:
The resolver package had some changes late in the 4.1.CR phase and the intention was to mark this package as unstable until these interfaces solidify, but we forgot to mark the package and public classes with the unstable annotation.

Modifications:
- resolver package public interfaces and package-info should be annotated with @UnstableApi

Result:
The unstable nature of the resolver package is more clearly communicated.


Scottmitch:
4.1 (b91dd06)

PR #6360, opened 2017-02-13
Add benchmarks for SSLEngine implementations

normanmaurer:
Motivation:

As we provide our own SSLEngine implementation we should have benchmarks to compare it against JDK impl.

Modifications:

Add benchmarks for wrap / unwrap and handshake performance.

Result:

Benchmarks FTW.

Issue #6348, opened 2017-02-11
Accept zero length TLS/SSL records for


Scottmitch:
### Expected behavior
Netty should be able to process zero length TLS records for application data. The SSLEngine should decide if the zero length record is invalid or not.

The TLSv1.2 RFC does allow for 0 length application data frames.

https://tools.ietf.org/html/rfc5246#section-6.2
> Implementations MUST NOT send zero-length fragments of Handshake,
   Alert, or ChangeCipherSpec content types.  Zero-length fragments of
   Application data MAY be sent as they are potentially useful as a
   traffic analysis countermeasure.

SSLv3 [1] and SSLv2 [2] do not address this scenario. I'm not sure if this means they are allowed or not.

[1] https://tools.ietf.org/html/rfc6101#section-5.2
[2] http://www-archive.mozilla.org/projects/security/pki/nss/ssl/draft02.html

### Actual behavior
`SslUtils#getEncryptedPacketLength` assumes that zero length records are invalid [3][4].

[3] https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/SslUtils.java#L106
[4] https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/SslUtils.java#L127

### Minimal yet complete reproducer code (or URL to code)
Here is a unit test which simulates the scenario. The SSLEngine implementations (SSL and OpenSSL) don't reject this and the unit test fails. If the engine doesn't reject this then we may get confused at the `SslHandler` layer when we attempt to determine the next packet's length, but we could be reading invalid data.

```java
    @Test
    public void testZeroLengthRecordDuringHandshakeIsRejected() throws Exception {
        SelfSignedCertificate cert = new SelfSignedCertificate();

        serverSslCtx = SslContextBuilder
                .forServer(cert.certificate(), cert.privateKey())
                .sslProvider(sslServerProvider())
                .build();
        SSLEngine server = serverSslCtx.newHandler(UnpooledByteBufAllocator.DEFAULT).engine();

        try {
            ByteBuffer plainServer = allocateBuffer(server.getSession().getPacketBufferSize());
            ByteBuffer emptyClientHello = allocateBuffer(5);
            // We create a zero length record during the handshake and the engine is expected to refuse this.
            // See https://tools.ietf.org/html/rfc5246#section-6.2.
            emptyClientHello.put(new byte[]{22, 3, 1, 0, 0});
            emptyClientHello.flip();

[...]

PR #6365, opened 2017-02-13
OpenSslEngine option to wrap/unwrap multiple packets per call

Scottmitch:
Motivation:
The JDK SSLEngine documentation says that a call to wrap/unwrap "will attempt to consume one complete SSL/TLS network packet" [1]. This limitation can result in thrashing in the pipeline to decode and encode data that may be spread amongst multiple SSL/TLS network packets.
ReferenceCountedOpenSslEngine also does not correct account for the overhead introduced by each individual SSL_write call if there are multiple ByteBuffers passed to the wrap() method.

Modifications:
- OpenSslEngine and SslHandler supports a mode to not comply with the limitation to only deal with a single SSL/TLS network packet per call
- ReferenceCountedOpenSslEngine correctly accounts for the overhead of each call to SSL_write
- SslHandler shouldn't cache maxPacketBufferSize as aggressively because this value may change before/after the handshake.

Result:
OpenSslEngine and SslHanadler can handle multiple SSL/TLS network packet per call.

[1] https://docs.oracle.com/javase/7/docs/api/javax/net/ssl/SSLEngine.html

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
@@ -409,11 +409,21 @@ public abstract class ReferenceCountedOpenSslContext extends SslContext implemen
 
     @Override
     public final SSLEngine newEngine(ByteBufAllocator alloc, String peerHost, int peerPort) {
-        return newEngine0(alloc, peerHost, peerPort);
+        return newEngine0(alloc, peerHost, peerPort, true);
     }
 
-    SSLEngine newEngine0(ByteBufAllocator alloc, String peerHost, int peerPort) {
-        return new ReferenceCountedOpenSslEngine(this, alloc, peerHost, peerPort, true);
+    @Override
+    final SslHandler newHandler(ByteBufAllocator alloc, boolean startTls) {
+        return new SslHandler(newEngine0(alloc, null, -1, false), startTls, false);
+    }
+
+    @Override
+    final SslHandler newHandler(ByteBufAllocator alloc, String peerHost, int peerPort, boolean startTls) {
+        return new SslHandler(newEngine0(alloc, peerHost, peerPort, false), startTls, false);
+    }
+
+    SSLEngine newEngine0(ByteBufAllocator alloc, String peerHost, int peerPort, boolean jdkCompatibilityMode) {
+        return new ReferenceCountedOpenSslEngine(this, alloc, peerHost, peerPort, jdkCompatibilityMode, true);
     }
 
     abstract OpenSslKeyMaterialManager keyMaterialManager();
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -43,8 +43,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
-
 import java.util.concurrent.locks.Lock;
+
 import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLEngineResult;
 import javax.net.ssl.SSLException;
@@ -58,10 +58,10 @@ import javax.net.ssl.SSLSessionContext;
 import javax.security.cert.X509Certificate;
 
 import static io.netty.handler.ssl.OpenSsl.memoryAddress;
-import static io.netty.handler.ssl.SslUtils.SSL_RECORD_HEADER_LENGTH;
 import static io.netty.util.internal.EmptyArrays.EMPTY_CERTIFICATES;
 import static io.netty.util.internal.EmptyArrays.EMPTY_JAVAX_X509_CERTIFICATES;
 import static io.netty.util.internal.ObjectUtil.checkNotNull;
+import static java.lang.Integer.MAX_VALUE;
 import static java.lang.Math.min;
 import static javax.net.ssl.SSLEngineResult.HandshakeStatus.FINISHED;
 import static javax.net.ssl.SSLEngineResult.HandshakeStatus.NEED_UNWRAP;
@@ -98,25 +98,10 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
      */
     private static final int DEFAULT_HOSTNAME_VALIDATION_FLAGS = 0;
 
-    static final int MAX_PLAINTEXT_LENGTH = 16 * 1024; // 2^14
-
     /**
-     * This is the maximum overhead when encrypting plaintext as defined by
-     * <a href="https://www.ietf.org/rfc/rfc5246.txt">rfc5264</a>,
-     * <a href="https://www.ietf.org/rfc/rfc5289.txt">rfc5289</a> and openssl implementation itself.
-     *
-     * Please note that we use a padding of 16 here as openssl uses PKC#5 which uses 16 bytes while the spec itself
-     * allow up to 255 bytes. 16 bytes is the max for PKC#5 (which handles it the same way as PKC#7) as we use a block
-     * size of 16. See <a href="https://tools.ietf.org/html/rfc5652#section-6.3">rfc5652#section-6.3</a>.
-     *
-     * TLS Header (5) + 16 (IV) + 48 (MAC) + 1 (Padding_length field) + 15 (Padding) + 1 (ContentType) +
-     * 2 (ProtocolVersion) + 2 (Length)
-     *
-     * TODO: We may need to review this calculation once TLS 1.3 becomes available.
+     * Depends upon tcnative ... only use if tcnative is available!
      */
-    static final int MAX_TLS_RECORD_OVERHEAD_LENGTH = SSL_RECORD_HEADER_LENGTH + 16 + 48 + 1 + 15 + 1 + 2 + 2;
-
-    static final int MAX_ENCRYPTED_PACKET_LENGTH = MAX_PLAINTEXT_LENGTH + MAX_TLS_RECORD_OVERHEAD_LENGTH;
+    static final int MAX_PLAINTEXT_LENGTH = SSL.SSL_MAX_PLAINTEXT_LENGTH;
 
     private static final AtomicIntegerFieldUpdater<ReferenceCountedOpenSslEngine> DESTROYED_UPDATER =
             AtomicIntegerFieldUpdater.newUpdater(ReferenceCountedOpenSslEngine.class, "destroyed");
@@ -146,7 +131,6 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
          * Started via {@link #beginHandshake()}.
          */
         STARTED_EXPLICITLY,
-
         /**
          * Handshake is finished.
          */
@@ -198,6 +182,7 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
     private boolean isInboundDone;
     private boolean outboundClosed;
 
+    private final boolean jdkCompatibilityMode;
     private final boolean clientMode;
     private final ByteBufAllocator alloc;
     private final OpenSslEngineMap engineMap;
@@ -209,6 +194,8 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
     private final ByteBuffer[] singleDstBuffer = new ByteBuffer[1];
     private final OpenSslKeyMaterialManager keyMaterialManager;
     private final boolean enableOcsp;
+    private int maxWrapOverhead;
+    private int maxWrapBufferSize;
 
     // This is package-private as we set it from OpenSslContext if an exception is thrown during
     // the verification step.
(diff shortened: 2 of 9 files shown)


Scottmitch:
- This depends upon https://github.com/netty/netty-tcnative/pull/233
- Once https://github.com/netty/netty/pull/6360 is pulled in I will update to account for multiple packets in the buffer.
- Discovered https://github.com/netty/netty/issues/6348 while developing this. TODOs will be addressed as part of that issue.

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -206,7 +206,7 @@ protected ByteBuf internalBuffer() {
     }
 
     @Override
-    public final void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
+    public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
```
normanmaurer: @Scottmitch I would love to not do this as this will basically make the `handlerRemoved0` redundant.

handler/src/main/java/io/netty/handler/ssl/OpenSslClientContext.java
```diff
@@ -181,10 +179,9 @@ public OpenSslClientContext(File trustCertCollectionFile, TrustManagerFactory tr
...
-                                KeyManagerFactory keyManagerFactory, Iterable<String> ciphers,
-                                CipherSuiteFilter cipherFilter, ApplicationProtocolConfig apn, String[] protocols,
-                                long sessionCacheSize, long sessionTimeout, boolean enableOcsp)
-            throws SSLException {
+                         KeyManagerFactory keyManagerFactory, Iterable<String> ciphers,
+                         CipherSuiteFilter cipherFilter, ApplicationProtocolConfig apn, String[] protocols,
+                         long sessionCacheSize, long sessionTimeout, boolean enableOcsp) throws SSLException {
         super(ciphers, cipherFilter, apn, sessionCacheSize, sessionTimeout, SSL.SSL_MODE_CLIENT, keyCertChain,
                 ClientAuth.NONE, protocols, false, enableOcsp);
         boolean success = false;
```
normanmaurer: as this seems to be only formatting can you revert this ?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -275,7 +275,7 @@ public String run() {
         try {
             synchronized (ReferenceCountedOpenSslContext.class) {
                 try {
-                    ctx = SSLContext.make(SSL.SSL_PROTOCOL_ALL, mode);
+                    ctx = SSLContext.make(SSL.SSL_PROTOCOL_TLS, mode);
```
normanmaurer: @Scottmitch this seems to be not related, please revert

normanmaurer:
@Scottmitch in general I would prefer if we could have two different PRs. One which fixes the "ReferenceCountedOpenSslEngine correctly accounts for the overhead of each call to SSL_write" and the other which introduce the new "feature". Can you please split it up ?

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -988,6 +1013,30 @@ private boolean ignoreException(Throwable t) {
         return false;
     }
 
+    @Override
+    public final void handlerRemoved(final ChannelHandlerContext ctx) throws Exception {
```
normanmaurer: See above

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
```diff
@@ -566,7 +572,7 @@ public void operationComplete(Future<Channel> future) {
         }
     }
 
-    @Test(timeout = 30000)
+    @Test(timeout = 60000)
```
normanmaurer: Why was this needed ?

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
```diff
@@ -601,33 +615,37 @@ private static void compositeBufSizeEstimationGuaranteesSynchronousWrite(
         Channel cc = null;
         try {
             final Promise<Void> donePromise = group.next().newPromise();
-            final int expectedBytes = 469 + 1024 + 1024;
+            // The goal is to provide the SSLEngine with many ByteBuf components to ensure that the overhead for wrap
+            // is correctly accounted for on each component.
+            final int numComponents = 150;
+            final int desiredBytes = 16384;
```
normanmaurer: can you add a comment why this value is chosen (like SSL record size)

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
Scottmitch: The issue is the buffer may be released by this method prematurely due to re-entry type issues. We have a few options on how to approach this but child classes not having the opportunity to control when the buffer released is a problem. We could deprecate `handlerRemoved0`, or do you have a more preferred approach?

handler/src/main/java/io/netty/handler/ssl/OpenSslClientContext.java
Scottmitch: done

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
Scottmitch: done

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
Scottmitch: This test now does 4x more work than it does before and the time constraints have changed. The point of the timeout is to catch "hung" tests and not performance issues and so to avoid false positives the timeout was doubled.

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
Scottmitch: done

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
Scottmitch: duplicate comment ... lets discuss above

normanmaurer:
@Scottmitch also please update the pr title and description

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
normanmaurer: ok sounds good.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -853,24 +921,38 @@ public final SSLEngineResult unwrap(
...
-                    int pendingEncryptedBytes = min(packetLength, remaining);
-                    ByteBuf bioWriteCopyBuf = writeEncryptedData(src, pendingEncryptedBytes);
                     try {
-                        readLoop:
-                        for (; dstsOffset < dstsEndOffset; ++dstsOffset) {
+                        for (;;) {
                             ByteBuffer dst = dsts[dstsOffset];
                             if (!dst.hasRemaining()) {
                                 // No space left in the destination buffer, skip it.
+                                if (++dstsOffset >= dstsEndOffset) {
```
netkins: MAJOR Refactor this code to not nest more than 5 if/for/while/switch/try statements. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS134)

pom.xml
```diff
@@ -263,7 +263,7 @@
     <!-- Fedora-"like" systems. This is currently only used for the netty-tcnative dependency -->
     <os.detection.classifierWithLikes>fedora</os.detection.classifierWithLikes>
     <tcnative.artifactId>netty-tcnative</tcnative.artifactId>
-    <tcnative.version>2.0.1.Final</tcnative.version>
+    <tcnative.version>2.0.2.Final-SNAPSHOT</tcnative.version>
```
Scottmitch: reminder: this must be updated

Scottmitch:
@normanmaurer - I have addressed all of your comments. PTAL

handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
```diff
@@ -57,9 +57,9 @@
 import java.nio.ByteBuffer;
 import java.nio.channels.ClosedChannelException;
 import java.security.KeyStore;
+import java.security.Provider;
```
normanmaurer: revert ?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -255,10 +249,20 @@ protected void deallocate() {
...
+
+            // setMode may impact the overhead.
+            maxWrapOverhead = SSL.getMaxWrapOverhead(ssl);
         } catch (Throwable cause) {
             SSL.freeSSL(ssl);
             PlatformDependent.throwException(cause);
         }
+        this.maxWrapOverhead = maxWrapOverhead;
+        // maxWrapBufferSize must be set after maxWrapOverhead because there maybe a dependency on this value.
+        maxWrapBufferSize = jdkCompatibilityMode ? maxEncryptedPacketLength() : maxEncryptedPacketLength() << 4;
```
normanmaurer: can you just add a comment why `<< 4` ?

Scottmitch:
>  I would like to delay pulling this in until we did another netty release

sgtm

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
Scottmitch: done ... a bit of a magic number ... the comment says we can make this configurable in the future if necessary. for now I think its fine to start with enough space for 16 TLS packets per wrap.

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -466,7 +495,7 @@ protected void decodeLast(ChannelHandlerContext ctx, ByteBuf in, List<Object> ou
         if (in.isReadable()) {
             // Only call decode() if there is something left in the buffer to decode.
             // See https://github.com/netty/netty/issues/4386
-            decode(ctx, in, out);
+            decode0(ctx, in, out);
         }
     }
```
Scottmitch: sure ... this PR will depend upon it but let me abstract it out.

Scottmitch: see https://github.com/netty/netty/pull/6728

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -469,6 +469,49 @@ private int readPlaintextData(final ByteBuffer dst) {
...
+
+    /**
+     * Visible only for testing!
+     */
+    final int maxEncryptedPacketLength() {
+        return maxWrapOverhead + MAX_PLAINTEXT_LENGTH;
+    }
+
+    final int calculateMaxLengthForWrap(int plaintextLength, int numComponents) {
+        return (int) min(maxWrapBufferSize, plaintextLength + (long) maxWrapOverhead * numComponents);
```
netkins: CRITICAL Inconsistent synchronization of io.netty.handler.ssl.ReferenceCountedOpenSslEngine.maxWrapOverhead; locked 66% of time [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=findbugs%3AIS2_INCONSISTENT_SYNC)

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -469,6 +469,65 @@ private int readPlaintextData(final ByteBuffer dst) {
...
+        return maxWrapOverhead + MAX_PLAINTEXT_LENGTH;
+    }
+
+    /**
+     * This method is intentionally not synchronized, only use if you know you are in the EventLoop
+     * thread and visibility on {@link #maxWrapBufferSize} and {@link #maxWrapOverhead} is achieved
+     * via other synchronized blocks.
+     */
+    final int calculateMaxLengthForWrap(int plaintextLength, int numComponents) {
+        return (int) min(maxWrapBufferSize, plaintextLength + (long) maxWrapOverhead * numComponents);
```
netkins: CRITICAL Inconsistent synchronization of io.netty.handler.ssl.ReferenceCountedOpenSslEngine.maxWrapOverhead; locked 83% of time [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=findbugs%3AIS2_INCONSISTENT_SYNC)

Scottmitch:
@normanmaurer - any more comments? this currently depends upon `tcnative 2.0.2.Final-SNAPSHOT` ... do you want me to wait to merge until after a release?

normanmaurer:
@Scottmitch no more comments... let me release tcnative first . I will do this over the next days

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: @Scottmitch just wonder if this is something we need to investigate

Scottmitch: let me know if you think there is a problem. The intention is not to necessarily to protect this variable in the synchronization block because it is mostly accessed inside the EventLoop (except maybe for `OpenSslSession#getPacketBufferSize()`).

blucas:
@Scottmitch what is the latest on this PR?

Scottmitch:
@blucas - We were just waiting for a release until we pulled this in. I will pull in shortly.

Scottmitch:
4.1 (f7b3caedd) 4.0 (c68a6adce)

PR #6369, opened 2017-02-13
Use tcnative's new setVerify modes

Scottmitch:
Modifications:
tcnative made some fixes and API changes related to setVerify. We should absorb these changes in Netty.

Modifications:
- Use tcnatives updated APIs
- Add unit tests to demonstrate correct behavior

Result:
Updated to latest tcnative code and more unit tests to verify expected behavior.


Scottmitch:
@jchambers - FYI. Also let me know if you are aware of a way to create an intermediate CA with out a configuration file (e.g. `openssl.cnf`).

Scottmitch:
This is necessary for https://github.com/netty/netty-tcnative/pull/232

handler/src/test/java/io/netty/handler/ssl/JdkSslEngineTest.java
```diff
@@ -271,6 +273,22 @@ public void testEnablingAnAlreadyDisabledSslProtocol() throws Exception {
...
+    }
+
+    @Ignore /* Does the JDK support a "max certificate chain length"? */
+    @Override
+    public void testMutualAuthValidClientCertChainTooLongFailRequireClientAuth() throws Exception {
+    }
+
+    @Override
+    protected boolean mySetupMutualAuthServerIsValidException(Throwable cause) {
+        // TODO(scott): work around for a JDK issue. The exception should be SSLHandshakeException.
```
Scottmitch: @normanmaurer - IIRC you opened an issue against the JDK? Can we create a Netty issue with the JDK issue and a link to this PR to cleanup these TODOs

jchambers:
Thanks for the heads-up!

> Also let me know if you are aware of a way to create an intermediate CA with out a configuration file.

I strongly suspect I may be missing your point, but if you omit the `-extfile` argument, openssl should just use its default configuration, and I think pretty much every default configuration within reason knows about the `v3_ca` extension.

Scottmitch:
> I think pretty much every default configuration within reason knows about the v3_ca extension.

I tried this however at least on my machine `basicConstraints = CA:true` was not set, and so the Intermediate CA was generated in a way that doesn't permit it to sign other certificates. I added a unit test to check the two cases (invalid intermediate CA [1], and valid intermediate CA [2]).

[1] https://github.com/netty/netty/pull/6369/files#diff-56d12ae4c3938365ef518ae82a30591cR10
[2] https://github.com/netty/netty/pull/6369/files#diff-56d12ae4c3938365ef518ae82a30591cR18

Scottmitch:
@normanmaurer - You put LGTM on https://github.com/netty/netty-tcnative/pull/232 (which is the counterpart to this) ... is this good too?

normanmaurer:
@Scottmitch Yep LGTM ..

handler/src/test/java/io/netty/handler/ssl/JdkSslEngineTest.java
normanmaurer: Not yet but will do soon

Scottmitch:
4.1 (fdcad31) 4.0 (a5386c3)

PR #6370, opened 2017-02-13
Cleanup  : String.length() == 0 replaced with String.isEmpty, ...

doom369:
Motivation:

IDE warnings fixes.

Modification:

`String.length() == 0` replaced with `String.isEmpty()`, unnecessary assertion removed, unnecessary class cast and correct `StringBuilder` size, empty `else` block commented.

Result:

Code easier to read. No warnings.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectDecoder.java
@@ -574,10 +574,11 @@ public abstract class HttpObjectDecoder extends ByteToMessageDecoder {
             do {
                 char firstChar = line.charAt(0);
                 if (name != null && (firstChar == ' ' || firstChar == '\t')) {
-                    StringBuilder buf = new StringBuilder(value.length() + line.length() + 1);
+                    String trimmedLine = line.toString().trim();
+                    StringBuilder buf = new StringBuilder(value.length() + trimmedLine.length() + 1);
                     buf.append(value)
                        .append(' ')
-                       .append(line.toString().trim());
+                       .append(trimmedLine);
                     value = buf.toString();
                 } else {
                     if (name != null) {
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpRequestEncoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpRequestEncoder.java
@@ -45,7 +45,7 @@ public class HttpRequestEncoder extends HttpObjectEncoder<HttpRequest> {
         // See http://tools.ietf.org/html/rfc2616#section-5.1.2
         String uri = request.uri();
 
-        if (uri.length() == 0) {
+        if (uri.isEmpty()) {
             uri += SLASH;
         } else {
             int start = uri.indexOf("://");
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/extensions/compression/PerMessageDeflateDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/extensions/compression/PerMessageDeflateDecoder.java
@@ -49,7 +49,7 @@ class PerMessageDeflateDecoder extends DeflateDecoder {
 
     @Override
     protected int newRsv(WebSocketFrame msg) {
-        return (((WebSocketFrame) msg).rsv() & WebSocketExtension.RSV1) > 0 ?
+        return (msg.rsv() & WebSocketExtension.RSV1) > 0 ?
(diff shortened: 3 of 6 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectDecoder.java
```diff
@@ -646,9 +647,10 @@ private LastHttpContent readTrailingHeaders(ByteBuf buffer) {
                         b.append(currentLastPos)
                          .append(lineTrimmed);
                         current.set(lastPos, b.toString());
-                    } else {
-                        // Content-Length, Transfer-Encoding, or Trailer
                     }
+                    //else {
+                    // Content-Length, Transfer-Encoding, or Trailer
```
Scottmitch: I realize you didn't introduce this comment but is this comment necessary? Can we make the comment more meaningful or just remove it?

@trustin - WDYT [1]

[1] https://github.com/doom369/netty/commit/4ce65f36212e7a97b902958ad6c56611091c83f5#diff-bba2e269877a9cc5f6b8ad204b1c89e5R451

trustin: I don't remember why I added that comment anymore unfortunately. I usually add a comment to an empty `else` block to explain when the condition of `if` clause will not match.

trustin: That being said, +1 for @Scottmitch's suggestion.

jasontedor: Knowing full well what I'm about to say will likely not have my intended effect, backfire and lead to me being viewed as contrary or difficult, as a user, follower, and occasional contributor to Netty, I feel it's worth mentioning these distractions (which happen quite frequently) just increase the cost of contributing. This comment has nothing to do with this PR. This discussion would have been better started on a separate issue.

normanmaurer: I agree with @jasontedor here... So all good @jasontedor  :)

normanmaurer: (also I will cleanup the stuff)

codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectDecoder.java
```diff
@@ -646,9 +647,10 @@ private LastHttpContent readTrailingHeaders(ByteBuf buffer) {
                         b.append(currentLastPos)
                          .append(lineTrimmed);
                         current.set(lastPos, b.toString());
-                    } else {
-                        // Content-Length, Transfer-Encoding, or Trailer
                     }
+                    //else {
```
netkins: MAJOR This block of commented-out lines of code should be removed. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3ACommentedOutCodeLine)

codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectDecoder.java
```diff
@@ -646,9 +647,10 @@ private LastHttpContent readTrailingHeaders(ByteBuf buffer) {
...
+                    // Content-Length, Transfer-Encoding, or Trailer
+                    //}
```
netkins: MAJOR This block of commented-out lines of code should be removed. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3ACommentedOutCodeLine)

doom369:
I reverted this change so it could be fixed with another PR.

normanmaurer:
Cherry-picked into 4.1 (506f0d8f8c10e1b24924f7d992a726d7bdd2e486) and 4.0 (49d4c35)

codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectDecoder.java
Scottmitch: The intention is not to hold up the PR but as the code was modified it is a good opportunity to get input/justifications or things that smell fishy. Opening another issue sgtm (all good @jasontedor)

jasontedor: Thanks for listening @normanmaurer and @Scottmitch.

I understand where you're coming from @Scottmitch and I agree with your tacit point that this code should not have been changed in this PR in the first place as it has nothing to do with the stated purpose of the PR. My point is that for both contributors and maintainers a more optimal response would have been a request to revert and the creation of a new issue.

PR #6376, opened 2017-02-15
OpenSslEngine wrap may not consume all data

Scottmitch:
Motivation:
If the OpenSslEngine has bytes pending in the non-application buffer and also generates wrapped data during the handshake then the handshake data will be missed. This will lead to a handshake stall and eventually timeout. This can occur if the non-application buffer becomes full due to a large certificate/hello message.

Modification:
- ReferenceCountedOpenSslEngine should not assume if no data is flushed from the non-application buffer that no data will be produced by the handshake.

Result:
New unit tests with larger certificate chains don't fail.


Scottmitch:
@normanmaurer - The unit tests introduced by PR https://github.com/netty/netty/pull/6369 uncovered these issues. Sorry I didn't realize until after rebase on another PR 😢

normanmaurer:
@Scottmitch  no worries... let me pull this in for you so it builds again

normanmaurer:
Cherry-picked into 4.1 (d60e37cb3da83af9541709284bb3532f489e0318) and 4.0 (b4624214092f0f9a67b4f34c5b63951780190d81)

PR #6377, opened 2017-02-15
Cleanup ChannelId handling of basic methods

### LINKED ISSUE #6371: DefaultChannelId equals, hashCode, compareTo clarification
`DefaultChannelId` chooses a hash as a random value in the constructor [1] and uses the `data` array in the comparison for `equals` [2]. This means that two `DefaultChannelId` could be `equal()` but have different `hashCode()` values which violates the interface contract of `Object#hashCode()`. 

`DefaultChannelId#compareTo()` always returns `0` [3]. The violates a "strong recommendation" on the `Comparable` interface that if `(x.compareTo(y)==0) == x.equals(y)`. Why is this?

This code has been around for a while [4] so I'm not sure if it is intentional or just cruft.

[1] https://github.com/netty/netty/blame/4.1/transport/src/main/java/io/netty/channel/DefaultChannelId.java#L181
[2] https://github.com/netty/netty/blame/4.1/transport/src/main/java/io/netty/channel/DefaultChannelId.java#L264
[3] https://github.com/netty/netty/blame/4.1/transport/src/main/java/io/netty/channel/DefaultChannelId.java#L251
[4] https://github.com/netty/netty/commit/40003ed250a41286dd73683e62737e318ab003ce

### Netty version
4.1.8

jasontedor:
Motiviation:

Simplify implementation of compareTo/equals/hashCode for ChannelIds.

Modifications:

We simplfy the hashCode implementation for DefaultChannelId by not making it random, but making it based on the underlying data. We fix the compareTo implementation for DefaultChannelId by using lexicographic comparison of the underlying data array. We fix the compareTo implementation for CustomChannelId to avoid the possibility of overflow.

Result:

Cleaner code that is easier to maintain.

Closes #6371

### DIFF
--- a/common/src/main/java/io/netty/util/internal/MathUtil.java
+++ b/common/src/main/java/io/netty/util/internal/MathUtil.java
@@ -65,6 +65,20 @@ public final class MathUtil {
     }
 
     /**
+     * Compares two {@code int} values.
+     *
+     * @param  x the first {@code int} to compare
+     * @param  y the second {@code int} to compare
+     * @return the value {@code 0} if {@code x == y};
+     *         {@code -1} if {@code x < y}; and
+     *         {@code 1} if {@code x > y}
+     */
+    public static int compare(final int x, final int y) {
+        // do not subtract for comparison, it could overflow
+        return x < y ? -1 : (x > y ? 1 : 0);
+    }
+
+    /**
      * Compare two {@code long} values.
      * @param x the first {@code long} to compare.
      * @param y the second {@code long} to compare.
--- a/transport/src/main/java/io/netty/channel/DefaultChannelId.java
+++ b/transport/src/main/java/io/netty/channel/DefaultChannelId.java
@@ -178,10 +178,10 @@ public final class DefaultChannelId implements ChannelId {
 
         // random
         int random = ThreadLocalRandom.current().nextInt();
-        hashCode = random;
         i = writeInt(i, random);
-
         assert i == data.length;
+
+        hashCode = Arrays.hashCode(data);
     }
 
     private int writeInt(int i, int value) {
@@ -247,21 +247,35 @@ public final class DefaultChannelId implements ChannelId {
     }
 
     @Override
-    public int compareTo(ChannelId o) {
-        return 0;
+    public int compareTo(final ChannelId o) {
+        if (o instanceof DefaultChannelId) {
+            final byte[] otherData = ((DefaultChannelId)o).data;
+            if (this == o) {
+                // short circuit
+                return 0;
+            }
+            // lexicographic comparison
+            int len1 = data.length;
+            int len2 = otherData.length;
+            int len = Math.min(len1, len2);
+
+            for (int k = 0; k < len; k++) {
+                byte x = data[k];
+                byte y = otherData[k];
+                if (x != y) {
+                    // treat these as unsigned bytes for comparison
+                    return (x & 0xff) - (y & 0xff);
+                }
+            }
+            return len1 - len2;
+        }
+
+        return asLongText().compareTo(o.asLongText());
     }
 
     @Override
     public boolean equals(Object obj) {
-        if (obj == this) {
-            return true;
-        }
-
-        if (!(obj instanceof DefaultChannelId)) {
-            return false;
-        }
-
-        return Arrays.equals(data, ((DefaultChannelId) obj).data);
+        return this == obj || (obj instanceof DefaultChannelId && Arrays.equals(data, ((DefaultChannelId) obj).data));
     }
 
     @Override
--- a/transport/src/main/java/io/netty/channel/embedded/EmbeddedChannelId.java
+++ b/transport/src/main/java/io/netty/channel/embedded/EmbeddedChannelId.java
@@ -40,7 +40,7 @@ final class EmbeddedChannelId implements ChannelId {
     }
 
     @Override
-    public int compareTo(ChannelId o) {
+    public int compareTo(final ChannelId o) {
         if (o instanceof EmbeddedChannelId) {
             return 0;
         }
--- a/transport/src/test/java/io/netty/channel/embedded/CustomChannelId.java
+++ b/transport/src/test/java/io/netty/channel/embedded/CustomChannelId.java
@@ -16,22 +16,25 @@
 package io.netty.channel.embedded;
 
 import io.netty.channel.ChannelId;
+import io.netty.util.internal.MathUtil;
 
 public class CustomChannelId implements ChannelId {
+
     private static final long serialVersionUID = 1L;
 
     private int id;
 
-    public CustomChannelId(int id) {
+    CustomChannelId(int id) {
         this.id = id;
     }
 
     @Override
-    public int compareTo(ChannelId o) {
+    public int compareTo(final ChannelId o) {
         if (o instanceof CustomChannelId) {
-            return id - ((CustomChannelId) o).id;
+            return MathUtil.compare(id, ((CustomChannelId)o).id);
         }
-        return -1;
+
+        return asLongText().compareTo(o.asLongText());
     }
 
     @Override
@@ -41,10 +44,7 @@ public class CustomChannelId implements ChannelId {
 
     @Override
     public boolean equals(Object obj) {
-        if (obj instanceof CustomChannelId) {
-            return id == ((CustomChannelId) obj).id;
-        }
-        return false;
+        return obj instanceof CustomChannelId && id == ((CustomChannelId) obj).id;
     }
 
     @Override


transport/src/main/java/io/netty/channel/ChannelId.java
```diff
@@ -43,7 +43,7 @@
  * </ul>
  * </p>
  */
-public interface ChannelId extends Serializable, Comparable<ChannelId> {
+public interface ChannelId<T extends ChannelId> extends Serializable, Comparable<T> {
```
Scottmitch: this is an API change. I don't think we can do this in a minor release. Can you revert this change please?

transport/src/main/java/io/netty/channel/DefaultChannelId.java
```diff
@@ -247,25 +247,36 @@ public int hashCode() {
...
+        if (this == o) {
+            // short circuit
+            return 0;
+        }
+        // lexicographic comparison
+        int len1 = data.length;
+        int len2 = o.data.length;
+        int len = Math.min(len1, len2);
+
+        int k = 0;
```
Scottmitch: consider just using a for loop `for (int k = 0; k < len; ++k)`

transport/src/main/java/io/netty/channel/DefaultChannelId.java
```diff
@@ -247,25 +247,36 @@ public int hashCode() {
...
-
-        return Arrays.equals(data, ((DefaultChannelId) obj).data);
+        return this == obj || (obj instanceof DefaultChannelId && Arrays.equals(data, ((DefaultChannelId) obj).data));
     }
 
     @Override
     public String toString() {
         return asShortText();
     }
+
```
Scottmitch: nit: kill line?

transport/src/main/java/io/netty/channel/embedded/EmbeddedChannelId.java
```diff
@@ -62,4 +58,5 @@ public boolean equals(Object obj) {
     public String toString() {
         return "embedded";
     }
+
```
Scottmitch: nit: kill line?

transport/src/main/java/io/netty/channel/DefaultChannelId.java
```diff
@@ -247,25 +247,36 @@ public int hashCode() {
...
+        int len1 = data.length;
+        int len2 = o.data.length;
+        int len = Math.min(len1, len2);
+
+        int k = 0;
+        while (k < len) {
+            byte x = data[k];
+            byte y = o.data[k];
+            if (x != y) {
+                return x - y;
```
Scottmitch: consider adding a comment that overflow won't occur because the arithmetic is promoted to `int`.

transport/src/test/java/io/netty/channel/embedded/CustomChannelId.java
```diff
@@ -17,21 +17,20 @@
...
 
     @Override
-    public int compareTo(ChannelId o) {
-        if (o instanceof CustomChannelId) {
-            return id - ((CustomChannelId) o).id;
-        }
-        return -1;
+    public int compareTo(CustomChannelId o) {
+        // do not subtract for comparison, it could overflow
+        return id < o.id ? -1 : (id == o.id ? 0 : 1);
```
Scottmitch: consider adding a `compare` to MathUtil for `int`. That way when we upgrade our java support it will be easier to identify and replace with the JDK methods.

fenik17:
IMHO, for  lexicographic comparison more suitable compare bytes as unsigned.

transport/src/main/java/io/netty/channel/ChannelId.java
normanmaurer: +1 .. Please revert

jasontedor:
Thanks for the feedback @Scottmitch and @normanmaurer. I have pushed a commit addressing your comments. I also agree with the suggestion of @fenik17 to do the lexicographic comparison treating the bytes as unsigned, so I included that in that commit as well.

transport/src/main/java/io/netty/channel/embedded/EmbeddedChannelId.java
```diff
@@ -40,12 +40,12 @@ public String asLongText() {
     }
 
     @Override
-    public int compareTo(ChannelId o) {
+    public int compareTo(final ChannelId o) {
         if (o instanceof EmbeddedChannelId) {
             return 0;
+        } else {
+            return asLongText().compareTo(o.asLongText());
```
normanmaurer: please revert this as the `else`is not needed or make this:

```java
return o instanceof EmbeddedChannelId ? 0 : asLongText().compareTo(o.asLongText());
```

transport/src/test/java/io/netty/channel/embedded/CustomChannelId.java
```diff
@@ -16,22 +16,26 @@
...
     }
 
     @Override
-    public int compareTo(ChannelId o) {
+    public int compareTo(final ChannelId o) {
         if (o instanceof CustomChannelId) {
-            return id - ((CustomChannelId) o).id;
+            final int otherId = ((CustomChannelId)o).id;
+            return MathUtil.compare(id, otherId);
+        } else {
```
normanmaurer: `} else {` is not needed. Please remove

jasontedor:
Thanks again @normanmaurer, I pushed another commit.

transport/src/main/java/io/netty/channel/DefaultChannelId.java
```diff
@@ -247,21 +247,35 @@ public int hashCode() {
...
+            for (int k = 0; k < len; k++) {
+                byte x = data[k];
+                byte y = otherData[k];
+                if (x != y) {
+                    // treat these as unsigned bytes for comparison
+                    return (x & 0xff) - (y & 0xff);
+                }
+            }
+            return len1 - len2;
+        } else {
```
normanmaurer: remove else block

transport/src/main/java/io/netty/channel/embedded/EmbeddedChannelId.java
```diff
@@ -40,7 +40,7 @@ public String asLongText() {
     }
 
     @Override
-    public int compareTo(ChannelId o) {
+    public int compareTo(final ChannelId o) {
```
normanmaurer: revert ?

transport/src/test/java/io/netty/channel/embedded/CustomChannelId.java
```diff
@@ -16,22 +16,26 @@
...
         this.id = id;
     }
 
     @Override
-    public int compareTo(ChannelId o) {
+    public int compareTo(final ChannelId o) {
         if (o instanceof CustomChannelId) {
-            return id - ((CustomChannelId) o).id;
+            final int otherId = ((CustomChannelId)o).id;
+            return MathUtil.compare(id, otherId);
```
normanmaurer: just merge these two lines ?

jasontedor: I pushed a commit that does this.

transport/src/main/java/io/netty/channel/embedded/EmbeddedChannelId.java
jasontedor: Why? The `final` is fine here. This is an improvement that breaks nothing.

transport/src/main/java/io/netty/channel/DefaultChannelId.java
jasontedor: I pushed a commit that does this.

jasontedor:
Thanks @normanmaurer. I've responded to your latest round of feedback.

transport/src/main/java/io/netty/channel/embedded/EmbeddedChannelId.java
normanmaurer: Just because it not really adds anything to the patch IMHO... that said I don't care too much

jasontedor: The goal of the patch was to clean up equals/compareTo/hashCode methods on the `ChannelId` implementations. :smile:

common/src/main/java/io/netty/util/internal/MathUtil.java
```diff
@@ -65,6 +65,20 @@ public static boolean isOutOfBounds(int index, int length, int capacity) {
...
     /**
+     * Compares two {@code int} values.
+     *
+     * @param  x the first {@code int} to compare
+     * @param  y the second {@code int} to compare
+     * @return the value {@code 0} if {@code x == y};
+     *         {@code -1} if {@code x < y}; and
+     *         {@code 1} if {@code x > y}
+     */
+    public static int compare(int x, int y) {
```
johnou: do we need to introduce this? can't we use java.lang.Integer#compare?

normanmaurer: We can't as we need to still support java6

johnou: ow, for some reason I thought Netty 4.0.x was java6 and Netty 4.1.x was java7.

jasontedor: No, it's 1.6 in the [pom.xml](https://github.com/netty/netty/blob/c92565d5c7000d2bd74355fe18cbdf98bfc5ba0d/pom.xml#L625).

normanmaurer:
@Scottmitch PTAL and pull in when satisfied. 

@jasontedor thanks again. Great work!

jasontedor:
Thanks @normanmaurer. I squashed.

Scottmitch:
4.1 (d59b484)

Thanks @jasontedor !

Scottmitch:
Also 4.1 (544d771)

transport/src/main/java/io/netty/channel/embedded/EmbeddedChannelId.java
trustin: I think it's always good to clean the code up while making modifications as long as it doesn't drain too much cognitive load. :-)

PR #6378, opened 2017-02-15
Allow to compile and run tests with different java versions

normanmaurer:
Motivation:

As we now need to compile with java8 we should still allow to run the tests with a different java version to ensure everythin also works with java 7 and 6.

Modifications:

Allow to pass "-DtestJavaHome" to the build and so use a different java version during running the tests.

Result:

Be able to run tests with different java versions.


normanmaurer:
Cherry-picked into 4.1 (c12908d608114856900d0f1a37dafd4cdcd6eacf) and 4.0 (adf0da3)

Issue #5458, opened 2016-06-28
PlatformDependent doesn't prefer JDK8 LongAdder 


CodingFabian:
I am running Netty on Java 8.

```
    /**
     * Creates a new fastest {@link LongCounter} implementaion for the current platform.
     */
    public static LongCounter newLongCounter() {
        if (HAS_UNSAFE) {
            return new LongAdderV8();
        } else {
            return new AtomicLongCounter();
        }
    }
```

On a normal modern Java8, where HAS_UNSAFE is true, this will always use the c&p LongAdderV8, otherwise the JDK Shipped AtomicLongCounter. Assuming that the JDK8 contained LongAdder is "better" (e.g. it might use intrinsics) I was wondering why there is no Java >=8 check which will directly use the JDK8 LongAdder (wrapped to comply to netty interface)

(also note the c&p typo "implementaion" in many of the javadocs)

normanmaurer:
@CodingFabian I will fix this.

normanmaurer:
@CodingFabian seems like there is no way to fix this as we still need to support java6 :(

CodingFabian:
Huh? Cant you keep the current if else and just wrap with an if else that checks version and if >=8 delegates to the JDK  Version? You might need an indirection to avoid classloading in <8

doom369:
@CodingFabian there are no specific intrinsics for `LongAdder`. However I agree that it is better to use JDK version as netty version differs from latest JDK version.

> You might need an indirection to avoid classloading in <8

Could you please show code example what do you mean?

johnou:
@CodingFabian something like this? https://github.com/johnou/netty/commit/bffeee37ca9e415688f68f33a573d44cb71579dc

normanmaurer:
Fixed by https://github.com/netty/netty/pull/6379

PR #6379, opened 2017-02-15
Remove backports of JDK8 classes

normanmaurer:
Motivation:

Java8 is out now for some time and JDK7 is no longer supported officially. We should remove all our backports and just use what the JDK provides us. This also will allow us to use intrinsics that are offered by the JDK implementations.

Modifications:

Remove all backports of jdk8 classes.

Result:

Use what the JDK offers us. This also fixes [#5458]


normanmaurer:
@CodingFabian FYI...

normanmaurer:
Cherry-picked into 4.1 (84188395be36893b214fcf63cea1544e35b2a654) and 4.0 (945addc)

PR #6380, opened 2017-02-15
Use the correct arguments when run with jdk9

normanmaurer:
Motivation:

We need to pass special arguments to run with jdk9 as otherwise some tests will not be able to run.

Modifications:

Allow to define extra arguments when running with jdk9

Result:

Tests pass with jdk9

### DIFF
--- a/handler/pom.xml
+++ b/handler/pom.xml
@@ -26,6 +26,11 @@
   <artifactId>netty-handler</artifactId>
   <packaging>jar</packaging>
 
+  <properties> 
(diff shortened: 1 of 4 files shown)


normanmaurer:
Cherry-picked into 4.1 (5728e0eb2cf0d2fe267d25c9f85a0a061dc7ab5f) and 4.0 (13b4aba5550fab38273192343fc7079b0171889d). 

Reviewed by @Scottmitch offline

normanmaurer:
@Scottmitch lol

PR #6382, opened 2017-02-15
Skip SSLEngineTests that depend on KeyManagerFactory when this is not…

normanmaurer:
… supported by the openssl version.

Motivation:

Some version of openssl dont support the needed APIs to use a KeyManagerFactory. In this case we should skip the tests.

Modifications:

- Use assumeTrue(...) to skip tests that need a KeyManagerFactory and its not supported.

Result:

Tests pass on all openssl versions we support.

### DIFF
--- a/handler/src/test/java/io/netty/handler/ssl/JdkOpenSslEngineInteroptTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/JdkOpenSslEngineInteroptTest.java
@@ -16,6 +16,7 @@
 package io.netty.handler.ssl;
 
 import org.junit.BeforeClass;
+import org.junit.Test;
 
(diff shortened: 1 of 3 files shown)


normanmaurer:
@Scottmitch PTAL... we either need to do this or just use "normal files" to construct the `SslContextBuilder`. WDYT ?

normanmaurer:
@Scottmitch we run the build with `netty-tcnative-boringssl-static` and `netty-tcnative`. With `netty-tcnative` this is needed.

normanmaurer:
Cherry-picked into 4.1 (43a2315372a8d642ce4df14db2ebfd001b9538e0) and 4.0 (02d5739)

normanmaurer:
@Scottmitch thanks for the review buddy

Issue #6385, opened 2017-02-15
OpenSslEngine may fail to write alerts if BIO is full


Scottmitch:
If, when attempting to send an alert, the write buffer is busy or BIO_write does not complete synchronously, there is no good way to tell the system to try again.

Once OpenSSL/BoringSSL provide a means to flush out pending alerts we should use it.

### Related
https://bugs.chromium.org/p/boringssl/issues/detail?id=130
https://github.com/openssl/openssl/issues/2638

### Minimal yet complete reproducer code (or URL to code)
Set the BIO non-application buffer size to 1 and run SslErrorTest

### Netty version
4.1.8

normanmaurer:
Shoulder but a restriction into Netty for now which at  least disallow such values?

Scottmitch:
@normanmaurer - We could limit the size of the non-application buffer, but I don't think this will help if the BIO happens to be full.

Scottmitch:
Once this is fixed we should modify the OpenSslEngine unit tests to run with a size of 1 for the non-application buffer size. If we do that now tests will fail so we should wait.

johnou:
@Scottmitch or ```@Parameterized.Parameters``` with app buff size 1 / default to cover both code paths?

Scottmitch:
@johnou - I'm not quite sure on the specific implementation yet but that is the general idea.

normanmaurer:
I think we fixed it

PR #6386, opened 2017-02-15
OpenSslEngine may lose data if the non-application buffer is small/full

Scottmitch:
Motivation:
If an event occurs which generates non-application data (shutdown, handshake failure, alert generation, etc...) and the non-application buffer in the ByteBuffer BIO is full (or sufficiently small) we may not propagate all data to our peer before tearing down the socket.

Modifications:
- when wrap() detects the outbound is closed, but there is more data pending in the non-application buffers, we must also check if OpenSSL will generate more data from calling SSL_shutdown again
- when wrap() detects a handshakeExcpetion during failure we should check if OpenSSL has any pending data (in addition to the non-application buff) before throwing the handshake exception

Result:
OpenSslEngine more reliably transmits data to the peer before closing the socket.


handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -484,6 +492,11 @@ public final SSLEngineResult wrap(
                     bytesProduced = SSL.bioFlushByteBuffer(networkBIO);
 
                     if (bytesProduced > 0 && handshakeException != null) {
+                        // TODO(scott): It is possible that when the handshake failed there was not enough room in the
+                        // non-application buffers to hold the alert. We should get all the data before progressing on.
+                        // However I'm not aware of a way to do this with the OpenSSL APIs.
+                        // See https://github.com/netty/netty/issues/6385.
```
Scottmitch: https://github.com/netty/netty/issues/6385

Scottmitch:
@normanmaurer - We already have unit tests :) Its just a reducing the size of the non-application buffer to `1`. However there are other test failures when this is done https://github.com/netty/netty/issues/6385#issuecomment-280078650.

normanmaurer:
@Scottmitch ok got it.. ship it once ci build completes.

normanmaurer:
@Scottmitch rebase on 4.1 ?

Scottmitch:
@normanmaurer - rebased & waiting for CI.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -484,6 +492,11 @@ public final SSLEngineResult wrap(
                     bytesProduced = SSL.bioFlushByteBuffer(networkBIO);
 
                     if (bytesProduced > 0 && handshakeException != null) {
+                        // TODO(scott): It is possible that when the handshake failed there was not enough room in the
```
netkins: INFO Complete the task associated to this TODO comment. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1135)

Scottmitch:
4.1 (4431ad8) 4.0 (2b07949)

PR #6388, opened 2017-02-15
Ensure Unsafe buffer implementations are used when sun.misc.Unsafe is…

normanmaurer:
… present

Motivation:

When sun.misc.Unsafe is present we want to use *Unsafe*ByteBuf implementations. We missed to do so in PooledByteBufAllocator when the heapArena is null.

Modifications:

- Correctly use UnpooledUnsafeHeapByteBuf
- Add unit tests

Result:

Use most optimal ByteBuf implementation.


normanmaurer:
Cherry-picked into 4.1 (8a3a3245df64351283714773dc23cee8322b14d3) and 4.0 (2935878eeeace36b7ce2d21c95bfbfe14e4da543)

Issue #6384, opened 2017-02-15
NetworkInterface.getNetworkInterfaces() can return null


CodingFabian:
To my suprise NetworkInterface.getNetworkInterfaces() can return null:
```
    public static Enumeration<NetworkInterface> getNetworkInterfaces()
        throws SocketException {
        final NetworkInterface[] netifs = getAll();

        // specified to return null if no network interfaces
        if (netifs == null)
            return null;
```

(not sure where its "specified")

https://github.com/netty/netty/blob/42c035982093d6e08201af9572ea43c7cf269adf/common/src/main/java/io/netty/util/internal/MacAddressUtil.java#L55

https://github.com/netty/netty/blob/3344cd21acc61819c327976ed5281daa49b0c05e/common/src/main/java/io/netty/util/NetUtil.java#L166

CodingFabian:
(derived from a bug report by a customer, not sure how they ended up having no network interfaces, also not sure if its relevant to use netty if you do not have network interfaces :))

normanmaurer:
@CodingFabian let me fix this

normanmaurer:
Fixed by https://github.com/netty/netty/issues/6389

PR #6389, opened 2017-02-15
Guard against having NetworkInterface.getNetworkInterfaces() return null

normanmaurer:
Motivation:

NetworkInterface.getNetworkInterfaces() may return null if no network interfaces are found. We should guard against it.

Modifications:

Check for null return value.

Result:

Fixes [#6384]

### DIFF
--- a/common/src/main/java/io/netty/util/internal/MacAddressUtil.java
+++ b/common/src/main/java/io/netty/util/internal/MacAddressUtil.java
@@ -24,6 +24,7 @@ import java.net.InetAddress;
 import java.net.NetworkInterface;
 import java.net.SocketException;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Enumeration;
 import java.util.LinkedHashMap;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


normanmaurer:
@CodingFabian FYI

common/src/main/java/io/netty/util/internal/MacAddressUtil.java
```diff
@@ -24,6 +24,7 @@
 import java.net.NetworkInterface;
 import java.net.SocketException;
 import java.util.Arrays;
+import java.util.Collections;
```
Scottmitch: remove?

normanmaurer: +1

Scottmitch:
lgtm after comment addressed

common/src/main/java/io/netty/util/internal/MacAddressUtil.java
netkins: MINOR Remove this unused import 'java.util.Collections'. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

normanmaurer: fixed

normanmaurer:
Cherry-picked into 4.1 (1843b318850ac7a3990c13b1d275e0e37491301e) and 4.0 (2dc9d119b19db57ce1b6f652a2374acef46937a1)

PR #6390, opened 2017-02-15
Correct comment for allowMaskMismatch parameter

### LINKED ISSUE #6387: allowMaskMismatch documentation issue
Netty version 4.1.8.Final

The documentation for allowMaskMismatch on various websocketx classes reads as follows:
'Allows to loosen the masking requirement on received frames. When this is set to false then also frames which are not masked properly according to the standard will still be accepted.'

The actual usage of this flag is here: https://github.com/netty/netty/blob/eb7f751ba519cbcab47d640cd18757f09d077b55/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocket08FrameDecoder.java#L192

So in fact when the flag is set to *true* then frames which are not masked properly according to the standard will still be accepted.

Suggest new wording:
'When set to true, frames which are not masked properly according to the standard will still be accepted'

(This same comment is present in quite a few classes: the Handshaker, HandshakerFactory, and Decoder classes for each version of the WebSocketClient.)

TheCycoONE:
Motivation:

The allowMaskMismatch parameter used throughout websocketx allows frames
with noncompliant masks when set to true, not false.

Modification:

Changed the javadoc comment everywhere it appears.

Result:

Fixes #6387

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocket07FrameDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocket07FrameDecoder.java
@@ -86,8 +86,8 @@ public class WebSocket07FrameDecoder extends WebSocket08FrameDecoder {
      *            Maximum length of a frame's payload. Setting this to an appropriate value for you application
      *            helps check for denial of services attacks.
      * @param allowMaskMismatch
-     *            Allows to loosen the masking requirement on received frames. When this is set to false then also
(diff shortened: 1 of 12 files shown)


normanmaurer:
@TheCycoONE thanks will merge soon :)

normanmaurer:
Cherry-picked into 4.1 (9ee4cc0ada3d8e46b139671dabfbfd35c8be3308).

@TheCycoONE thanks a lot!

Issue #6303, opened 2017-02-02
Bytes are dropped when using epoll and auto read = false when peer closes connection


alkemist:
### Expected behavior

Received but not yet emitted bytes at time of peer connection close when using epoll edge triggered and `autoRead=false` should be emitted (i.e. behaviour as of 4.1.6).

### Actual behavior

As of 4.1.7, As soon as the remote peer disconnects, no further bytes will be emitted.

The problematic change is here: https://github.com/netty/netty/commit/631077c79391fde22afdb1e43d0f642ba530101e#diff-5d2d71235faad7440ee0020202cfa744L59

This is effectively a different take on #3709, but for the non auto read case and is a regression introduced to fix #6173.

`maybeMoreDataToRead() && config.isAutoRead() || super.continueReading()` is _always_ `false` when `autoRead=false`.  This means that queued up bytes are not drained https://github.com/netty/netty/blob/4.1/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollChannel.java#L378 when `EPOLLRDHUP` is read, and the subsequent `shutdownInput();` causes future `channel.read()` calls to effectively NOOP because of https://github.com/netty/netty/blob/4.1/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollStreamChannel.java#L972 (i.e. `isInputShutdown()` is `true`).

I don't know what the right fix is, but my best naive guess would be to change the start of `epollInReady()` to be:

```
if (!maybeMoreDataToRead && fd().isInputShutdown()) {
  clearEpollIn0();
  return;
}
```

alkemist:
I think this is likely to affect lots of people, and is also blocking a new release of Ratpack. Can you please also indicate when a resolution for this could be released (i.e. is there already a planned patch release? are there other open issues blocking a 4.1.9?), when it is actually resolved. 

Thanks!

normanmaurer:
@Scottmitch PTAL

Scottmitch:
I'm on it ... will keep you posted when I have a fix. This will be fixed for 4.1.9.

normanmaurer:
@alkemist as a workaround you could use NIO for now.

alkemist:
Thanks @normanmaurer & @Scottmitch.

alkemist:
@Scottmitch bad news, the problem persists. I built from the PR @ ae3365c4793cd8d3233875a8aa35fa20999c7c34.

[...]

PR #6391, opened 2017-02-15
EPOLLRDHUP prematurely marking input as shutdown

Scottmitch:
Motivation:
When the EPOLLRDHUP event is received we assume that the read side of the FD is no longer functional and force the input state to be shutdown. However if the channel is still active we should rely upon EPOLLIN and read to indicate there is no more data before we update the shutdown state. If we do not do this we may not read all pending data in the FD if the RecvByteBufAllocator doesn't want to consume it all in a single read operation.

Modifications:
- AbstractEpollChannel#epollRdHupReady() shouldn't force shutdown the input if the channel is active

Result:
All data can be read even if the RecvByteBufAllocator doesn't read it all in the current read loop.
Fixes https://github.com/netty/netty/issues/6303

### DIFF
--- a/testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketHalfClosedTest.java
+++ b/testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketHalfClosedTest.java
@@ -150,6 +150,120 @@ public class SocketHalfClosedTest extends AbstractSocketTest {
         }
     }
 
+    @Test
+    public void testAllDataReadClosure() throws Throwable {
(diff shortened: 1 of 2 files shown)


Scottmitch:
@alkemist - Can you verify this one?

@normanmaurer - I will add a unit test.

Scottmitch:
Verified by @alkemist https://github.com/netty/netty/issues/6303#issuecomment-280196441

Scottmitch:
4.1 (f3dd410)

PR #6392, opened 2017-02-16
Enforce java8 for compilation

normanmaurer:
Motivation:

Commit 591293bfb4f6a48a311d195303ce772fb801ec95 changed the build to need java8 but missed to adjust the enforce rule as well.

Modifications:

Enforce java8+

Result:

Quickly fail when user tries to compile with pre java8

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -596,9 +596,9 @@
             <configuration>
               <rules>
                 <requireJavaVersion>
-                  <!-- Enforce JDK 1.7+ for compilation. -->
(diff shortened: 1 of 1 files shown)


Scottmitch:
we should update the [developer guide](http://netty.io/wiki/setting-up-development-environment.html) too

normanmaurer:
@Scottmitch yep will do

normanmaurer:
Cherry-picked into 4.1 (ddd0e2f) and 4.0 (2d94c39)

PR #6394, opened 2017-02-16
Make netty build work on Java9

normanmaurer:
Motivation:

We missed some stuff in 5728e0eb2cf0d2fe267d25c9f85a0a061dc7ab5f and so the build failed on java9

Modifications:

- Add extra cmdline args when needed
- skip the autobahntestsuite as jython not works with java9
- skip the osgi testsuite as the maven plugin not works with java9

Result:

Build finally passed on java9

### DIFF
--- a/testsuite-osgi/pom.xml
+++ b/testsuite-osgi/pom.xml
@@ -26,11 +26,13 @@
   <artifactId>netty-testsuite-osgi</artifactId>
   <packaging>jar</packaging>
 
+
+
   <name>Netty/Testsuite/OSGI</name>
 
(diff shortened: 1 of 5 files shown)


testsuite-osgi/pom.xml
```diff
@@ -26,11 +26,13 @@
   <artifactId>netty-testsuite-osgi</artifactId>
   <packaging>jar</packaging>
 
+
```
Scottmitch: kill lines?

normanmaurer:
addressed and upstream

PR #6395, opened 2017-02-16
EpollDatagramChannel avoid getOption

Scottmitch:
Motivation:
EpollDatagramChannel uses getOption in the isActive method. getOption is backed by a relatively large conditional if/else if block and this conditional checking can be avoided in the epoll transport.

Modifications:
- Add EpollDatagramChannelConfig#getActiveOnOpen and use this in EpollDatagramChannel

Result:
Conditional checking due to getOption is removed from EpollDatagramChannel.

### DIFF
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollDatagramChannelConfig.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollDatagramChannelConfig.java
@@ -132,6 +132,10 @@ public final class EpollDatagramChannelConfig extends EpollChannelConfig impleme
         this.activeOnOpen = activeOnOpen;
     }
 
+    boolean getActiveOnOpen() {
+        return activeOnOpen;
+    }
+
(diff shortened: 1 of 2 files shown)


transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollDatagramChannelConfig.java
```diff
@@ -132,6 +132,10 @@ private void setActiveOnOpen(boolean activeOnOpen) {
         this.activeOnOpen = activeOnOpen;
     }
 
+    boolean getActiveOnOpen() {
```
netkins: MAJOR A getX() method which returns a boolean should be named isX() [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3ABooleanGetMethodName)

Scottmitch:
4.1 (0befcd8) 4.0 (3b65e7e)

PR #6396, opened 2017-02-16
Make netty build work on Java9

normanmaurer:
Motivation:

We missed some stuff in 5728e0eb2cf0d2fe267d25c9f85a0a061dc7ab5f and so the build failed on java9

Modifications:

- Add extra cmdline args when needed
- skip the autobahntestsuite as jython not works with java9
- skip the osgi testsuite as the maven plugin not works with java9

Result:

Build finally passed on java9

### DIFF
--- a/handler-proxy/pom.xml
+++ b/handler-proxy/pom.xml
@@ -26,6 +26,11 @@
   <artifactId>netty-handler-proxy</artifactId>
   <packaging>jar</packaging>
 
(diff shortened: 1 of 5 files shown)


normanmaurer:
Cherry-picked into 4.1 (c5324ea48c1f1e2d4bacbf84899a67034474c4e0) and 4.0 (3fa3709)

PR #6397, opened 2017-02-16
Prefer JDK ThreadLocalRandom implementation over ours.

normanmaurer:
Motivation:

We have our own ThreadLocalRandom implementation to support older JDKs . That said we should prefer the JDK provided when running on JDK >= 7

Modification:

Using ThreadLocalRandom implementation of the JDK when possible.

Result:

Make use of JDK implementations when possible.

### DIFF
--- a/common/src/main/java/io/netty/util/ResourceLeakDetector.java
+++ b/common/src/main/java/io/netty/util/ResourceLeakDetector.java
@@ -18,7 +18,6 @@ package io.netty.util;
 
 import io.netty.util.internal.PlatformDependent;
 import io.netty.util.internal.SystemPropertyUtil;
-import io.netty.util.internal.ThreadLocalRandom;
 import io.netty.util.internal.logging.InternalLogger;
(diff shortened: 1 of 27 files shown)


common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -246,7 +244,7 @@ private DefaultResourceLeak track0(T obj) {
         }
 
         if (level.ordinal() < Level.PARANOID.ordinal()) {
-            if ((ThreadLocalRandom.current().nextInt(0, samplingInterval)) == 0) {
```
Scottmitch: its too bad we lose this method, but I guess its not worth supporting a new wrapper API just to expose this.

normanmaurer: Yeah

Scottmitch:
4.1 (fbf0e5f) 4.0 ( see https://github.com/netty/netty/pull/6402 )

PR #6398, opened 2017-02-16
Use Unsafe to read ByteBuffer.address field to make it work on Java9 …

normanmaurer:
…as well.

Motivation:

Java9 does not allow changing access level via reflection by default. This lead to the situation that netty disabled Unsafe completely as ByteBuffer.address could not be read.

Modification:

Use Unsafe to read the address field as this works on all Java versions.

Result:

Again be able to use Unsafe optimisations when using Netty with Java9


normanmaurer:
Cherry-picked into 4.1 (6ac5f35077925ebe2d76e9e2e1eab51a8b5eb35b) and 4.0 (d57d041)

PR #6400, opened 2017-02-16
Log why it was not possible to use ByteBuffer.cleaner

normanmaurer:
Motivation:

We should log why we can not use ByteBuffer.cleaner and so maybe allow the user to fix it.

Modifications:

- Use Unsafe to access the field
- Log the exception when we can not use ByteBuffer.cleaner

Result:

Easier to debug why using cleaner is not possible.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/Cleaner0.java
+++ b/common/src/main/java/io/netty/util/internal/Cleaner0.java
@@ -38,16 +38,15 @@ final class Cleaner0 {
 
     static {
         ByteBuffer direct = ByteBuffer.allocateDirect(1);
-        Field cleanerField;
         long fieldOffset = -1;
         Method clean = null;
         boolean cleanerIsRunnable = false;
+        Throwable error = null;
         if (PlatformDependent0.hasUnsafe()) {
             try {
-                cleanerField = direct.getClass().getDeclaredField("cleaner");
-                cleanerField.setAccessible(true);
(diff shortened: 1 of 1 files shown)


common/src/main/java/io/netty/util/internal/Cleaner0.java
```diff
@@ -42,12 +42,12 @@
         long fieldOffset = -1;
         Method clean = null;
         boolean cleanerIsRunnable = false;
+        Throwable error = null;
         if (PlatformDependent0.hasUnsafe()) {
             try {
                 cleanerField = direct.getClass().getDeclaredField("cleaner");
```
Scottmitch: move the variable declaration `Field cleanerField` inside this scope?

normanmaurer:
@Scottmitch PTAL again

common/src/main/java/io/netty/util/internal/Cleaner0.java
Scottmitch: looks like this still wasn't addressed ... did you push the most recent code?

normanmaurer: I am an idiot... forgot to do "git add ...." PTAL again

normanmaurer:
Cherry-picked into 4.1 (67be7c5b9f5920d5f99d40e8b483df112e103ba8) and 4.0 (6a2dd3e)

PR #6402, opened 2017-02-16
Prefer JDK ThreadLocalRandom implementation over ours.

Scottmitch:
Motivation:

We have our own ThreadLocalRandom implementation to support older JDKs . That said we should prefer the JDK provided when running on JDK >= 7

Modification:

Using ThreadLocalRandom implementation of the JDK when possible.

Result:

Make use of JDK implementations when possible.


normanmaurer:
Cherry-picked into 4.0 (da94a908ad5b045ef5ae6d2bb8b1fcade654bf00)

PR #6403, opened 2017-02-17
OpenSslEngine should respect hostname verification

Scottmitch:
Motivation:
OpenSSL doesn't automatically verify hostnames and requires extract method calls to enable this feature [1]. We should allow this to be configured.

Modifications:
- SSLParamaters#getEndpointIdentificationAlgorithm() should be respected and configured via tcnative interfaces.

Result:
OpenSslEngine respects hostname verification.

[1] https://wiki.openssl.org/index.php/Hostname_validation


Scottmitch:
depends upon https://github.com/netty/netty-tcnative/pull/237

handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
```diff
@@ -331,6 +341,14 @@ public static boolean supportsKeyManagerFactory() {
         return SUPPORTS_KEYMANAGER_FACTORY;
     }
 
+    /**
+     * Returns {@code true} if <a href="https://wiki.openssl.org/index.php/Hostname_validation">Hostname Validation</a>
+     * is supported when using OpenSSL.
+     */
+    public static boolean supportsHostnameVerification() {
```
normanmaurer: Just a nit but openssl calls it "Hostname Validation" while we call "Hostname Verification". Maybe make it consistent ?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1557,14 +1559,26 @@ public final synchronized void setSSLParameters(SSLParameters sslParameters) {
 
         int version = PlatformDependent.javaVersion();
         if (version >= 7) {
-            endPointIdentificationAlgorithm = sslParameters.getEndpointIdentificationAlgorithm();
+            final String endPointIdentificationAlgorithm = sslParameters.getEndpointIdentificationAlgorithm();
+            final boolean endPointVerificationEnabled = endPointIdentificationAlgorithm != null &&
+                    !endPointIdentificationAlgorithm.isEmpty();
+            SSL.setHostNameValidation(ssl, DEFAULT_HOSTNAME_VERIFICATION_FLAGS,
+                    endPointVerificationEnabled ? getPeerHost() : null);
```
normanmaurer: @Scottmitch I wonder if we need to handle stuff differently depending on what is the content of `endPointIdentificationAlgorithm `

Scottmitch: https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLParameters.html#setEndpointIdentificationAlgorithm-java.lang.String-
> If the algorithm parameter is non-null or non-empty, the endpoint identification/verification procedures must be handled during SSL/TLS handshaking.

OpenSSL Engine support HTTPS and LDAPS as defined in https://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html. As you can see I added support for defines in tcnative, but I'm not yet sure we need to use them.

normanmaurer: @Scottmitch does this mean we need to check if HTTPS / LDAPS is used and if not throw an exception ?

Scottmitch: I don't think we need to be this strict. These values are not strictly defined and are just used as strings. The API (referenced above) says if it is null/non-empty we should verify the certs.

I need to investigate the RFCs referenced by the docs for LDAPS and HTTPS to see if OpenSSL's options can mimic the behavior of the JDK.

Scottmitch:
Based on the flags described in https://www.openssl.org/docs/man1.0.2/crypto/X509_check_host.html I don't think we can exactly mimic the JDK's support for `HTTPS` [1] or `LDAPS` [2] which are described as supported by the JDK in [3].

I suggest we handle modifications to the flags used for OpenSSL verification on a per-request basis and re-evaluate at that time.

[1] http://www.ietf.org/rfc/rfc2818.txt
[2] http://www.ietf.org/rfc/rfc2830.txt
[3] https://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html

normanmaurer:
+1

normanmaurer:
@Scottmitch Beta5 is out... I kicked off another build

handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
```diff
@@ -130,6 +132,12 @@
                             availableOpenSslCipherSuites.add(c);
                         }
                         try {
+                            SSL.setHostNameValidation(ssl, 0, "netty.io");
+                            supportsHostNameValidation = true;
+                        } catch (Throwable ignore) {
+                            logger.debug("Hostname Verification not supported.");
+                        }
+                        try {
```
netkins: MAJOR Extract this nested try block into a separate method. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1141)

Scottmitch:
4.1 (d8e6fbb) 4.0 (7cbc404)

PR #6407, opened 2017-02-17
Introduced MqttMessageBuilders to fluently create MQTT messages

andsel:
Motivation:
This PR would simplify the creation of MqttMessages in tests, to avoid long parameters list in constructors of variable, fixed and payload parts.

Modification:

Just introduced a builder class that permit to create messaged with chined notation.

Result:

Created MessageBuilder class with some factory methods (for connect, publish etc) and some inner classes used to carry temporary variables. At the end there is a build method that contains the logic to wire up the parts of the message.


normanmaurer:
Cherry-picked into 4.1 (ad51cda2cd96400ef07a3b0107baba3b13b473a4)

@andsel thanks!

PR #6408, opened 2017-02-18
Cleanup DefaultHttp2FrameReader about verifyUnknownFrame

chhsiao90:
Just a minor code cleanup! Please help review, thanks!

Motivation:

In previous PR about handling unknown frame in the middle of header
block, I didn't notice and re-use about checking is processing header
. And I added a redundant method for same functionality.
I think that the redundant method would lead to some misleading
situation.

Modifications:

Removed redundant code on DefaultHttp2FrameReader

Result:

The code is more readable

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2FrameReader.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2FrameReader.java
@@ -227,6 +227,7 @@ public class DefaultHttp2FrameReader implements Http2FrameReader, Http2FrameSize
                 break;
             default:
                 // Unknown frame type, could be an extension.
+                verifyUnknownFrame();
                 break;
         }
     }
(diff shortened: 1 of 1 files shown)


codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2FrameReader.java
```diff
@@ -496,6 +492,12 @@ public void processFragment(boolean endOfHeaders, ByteBuf fragment,
         resetHeadersContinuationIfEnd(flags.endOfHeaders());
     }
 
+    /**
+     * If received {@code true} from {@link Http2Flags#endOfHeaders()} ,
+     * close {@link #headersContinuation} to represent current state is not processing header block,
+     * And can start receiving frame other than CONTINUATION frame
+     * @param endOfHeaders
```
Scottmitch: I'm not sure these comments are necessary. The method name seems pretty straightforward, the comments could use some rework, and it the method is private. Can we just remove these comments?

normanmaurer: +1

chhsiao90:
No problem, no comment was removed!

Scottmitch:
4.1 (b1436e8)

Thanks @chhsiao90 !

Issue #1798, opened 2013-08-29
LengthFieldBasedFrameDecoder produces IllegalArgumentException when frame is too big


normanmaurer:
The problem arise if the max frame size exceed and the buffer hold the complete frame already. This is because the calculation of the value for skipBytes(...) will be negative atm.

okou19900722:
Netty3 also has this problem, not fixed

Scottmitch:
Netty 3.x has been EOL for a while [1]. Please upgrade to 4.1 and report a new issue if the problem still exists.

[1] http://netty.io/news/2016/06/29/3-10-6-Final.html

PR #6411, opened 2017-02-18
netty3 LengthFieldBasedFrameDecoder skip too many

okou19900722:
#1798
netty4 fixed, but netty3 not


normanmaurer:
Netty 3.x is EOL

normanmaurer:
@okou19900722 when you want to have this pulled in please also provide a unit test. That said netty 3 is EOL and so no new release is expected atm

PR #6413, opened 2017-02-19
Log used native library by netty-tcnative

normanmaurer:
Motivation:

As netty-tcnative can be build against different native libraries and versions we should log the used one.

Modifications:

Log the used native library after netty-tcnative was loaded.

Result:

Easier to understand what native SSL library was used.


normanmaurer:
This will log something like:
```
13:19:28.877 [main] DEBUG io.netty.handler.ssl.OpenSsl - netty-tcnative using native library: OpenSSL 1.0.2j  26 Sep 2016

```

normanmaurer:
Cherry-picked into 4.1 (c57a1bd) and 4.0 (9392e7e)

PR #6415, opened 2017-02-20
feature: add string blank check

xujianhai:
Motivation:

String blank should be checked.

Modification:

Add isNullOrBlank to StringUtil.

Result:

String blank checked


johnou:
@xujianhai this is an internal class, you shouldn't add your own utility methods or use it at all..

normanmaurer:
@xujianhai as @johnou said... Closing

PR #6420, opened 2017-02-20
SSLEngineTest issue introduced by d8e6fbb9c3656663f41797a15f38bb25fdc…

Scottmitch:
…ab834

Motivation:
d8e6fbb9c3656663f41797a15f38bb25fdcab834 attempted to account for the JDK not throwing the expected SSLHandshakeException by allowing a SSLException to also pass the test. However in some situations the SSLException will not be the top level exception and the Throwable must be unwrapped to see if the root cause is an SSLException.

Modifications:
- Unwrap exceptions thrown by the JDK's SSLEngine to check for SSLException.

Result:
SSLEngineTest (and derived classes) are more reliable.

### DIFF
--- a/handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
@@ -593,6 +593,17 @@ public abstract class SSLEngineTest {
                 mySetupMutualAuthServerIsValidServerException(serverException));
     }
 
+    protected static boolean causedBySSLException(Throwable cause) {
+        Throwable next = cause;
+        while (next != null) {
+            if (next instanceof SSLException) {
+                return true;
+            }
+            next = next.getCause();
+        }
+        return false;
+    }
+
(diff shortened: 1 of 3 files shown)


handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
```diff
@@ -593,6 +593,17 @@ private void testMutualAuthClientCertFail(ClientAuth auth, String clientCert, bo
                 mySetupMutualAuthServerIsValidServerException(serverException));
     }
 
+    protected static boolean causedBySSLException(Throwable cause) {
+        Throwable next = cause;
+        while (next != null) {
```
normanmaurer: Make this a 'do {} while ' loop as the method should never be called with null?

Scottmitch: done

Scottmitch:
4.1 (2dffc2f) 4.0 (bca1d9c)

PR #6427, opened 2017-02-21
Fix javadoc issues

fenik17:
Motivation:

Invalid javadoc in project

Modifications:

Fix it

Result:

Correct javadoc

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/ClientCookieEncoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/ClientCookieEncoder.java
@@ -15,6 +15,8 @@
  */
 package io.netty.handler.codec.http;
 
+import io.netty.handler.codec.http.cookie.ClientCookieDecoder;
+
 /**
  * A <a href="http://tools.ietf.org/html/rfc6265">RFC6265</a> compliant cookie encoder to be used client side,
  * so only name=value pairs are sent.
--- a/codec-http/src/main/java/io/netty/handler/codec/http/ServerCookieEncoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/ServerCookieEncoder.java
@@ -15,6 +15,8 @@
  */
 package io.netty.handler.codec.http;
 
+import io.netty.handler.codec.http.cookie.ServerCookieDecoder;
+
 import java.util.Collection;
 import java.util.List;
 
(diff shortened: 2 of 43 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/ClientCookieEncoder.java
```diff
@@ -15,6 +15,8 @@
  */
 package io.netty.handler.codec.http;
 
+import io.netty.handler.codec.http.cookie.ClientCookieDecoder;
+
```
slandelle: IMHO, it's not a good thing to add an import for the sheer sake of javadoc while the actual code doesn't use it. I would rather use the fully qualified class name in javadoc then.

codec-http/src/main/java/io/netty/handler/codec/http/ServerCookieEncoder.java
```diff
@@ -15,6 +15,8 @@
  */
 package io.netty.handler.codec.http;
 
+import io.netty.handler.codec.http.cookie.ServerCookieDecoder;
+
```
slandelle: Same as above, I would use the fully qualified class name in javadoc instead.

codec/src/main/java/io/netty/handler/codec/protobuf/ProtobufVarint32LengthFieldPrepender.java
```diff
@@ -15,16 +15,16 @@
  */
 package io.netty.handler.codec.protobuf;
 
-import java.io.IOException;
-
+import com.google.protobuf.CodedOutputStream;
+import com.google.protobuf.nano.CodedOutputByteBufferNano;
```
slandelle: Idem

codec-http/src/main/java/io/netty/handler/codec/http/ClientCookieEncoder.java
fenik17: Explicit import protects against javadoc breakage at compile time if a class renamed negligently.
And a simple class name is more comfortable in javadoc..

slandelle: Don't modern IDEs update the javadoc when performing such refactoring?

normanmaurer: We prefer to import in netty. So all good

normanmaurer:
Cherry-picked into 4.1 (0623c6c5334bf43299e835cfcf86bfda19e2d4ce) and 4.0 (1ff8c762cd5228aaec2da6d464b1beb5e7eee276)

fenik17:
@normanmaurer 4.0 has unresolved conflicts in HttpHeaders.java

normanmaurer:
@fenik17 ups... fixed by 5d28bd1110b03ceae4c21f878e04a108d8e903ef

Issue #5857, opened 2016-09-23
http/2 cleartext upgrade + prior knowledge


mosesn:
http/2 cleartext has two mechanisms, the http/1.1 upgrade header, and the connection preface.  Netty supports inspecting the connection preface in Http2ConnectionHandler, and supports upgrading via the http/1.1 upgrade header in HttpServerUpgradeHandler + Http2ServerUpgradeCodec.  However, it's not obvious how to combine the two of these into a single piece.  One difficulty is that there are several moving pieces–we need an HttpCodec, and an HttpServerUpgradeHandler, which comes with an Http2ServerUpgradeCodec and the Http2Codec we will want to replace the HttpCodec with.

This looks like:

`==Bytes==>[Early stuff in pipeline]==>[HttpCodec]==HttpRequest==>[HttpServerUpgradeHandler]===>[Rest of the pipeline]==END`

which should be transformed into

`==Bytes==>[Early stuff in pipeline]==>[Http2Codec]==END`
which has child handlers which look like
`==Http2Frame==>[Rest of the pipeline]`

When we want to add prior knowledge support, the prior knowledge support must be before the HttpCodec, but must do the same work as the HttpServerUpgradeHandler.

`==Bytes==>[Early stuff in pipeline]==>[Prior Knowledge Upgrader]==>[HttpCodec]==HttpRequest==>[HttpServerUpgradeHandler]===>[Rest of the pipeline]==END`

but it still transforms to the same thing as before.

One thing we were considering was how we might add it simply to the Http2ServerUpgradeCodec, so that it installs the prior knowledge upgrader before the SourceCodec in the pipeline when it's installed.  Another possible way was to have a separate PriorKnowledgeHandler which is installed separately from the HttpServerUpgradeHandler.

Please let us (me + @nhnFreespirit) know what you think would be the right way forward here.

cc @Scottmitch @buchgr @normanmaurer @ejona86

Scottmitch:
related to https://github.com/netty/netty/issues/4438

normanmaurer:
@mosesn @nhnFreespirit @Scottmitch is this somehow related to https://github.com/netty/netty/pull/6430 ?

nhnFreespirit:
@normanmaurer yeah, same thing. 

Because we never really heard back on this one, I went ahead and added an implementation to Finagle in https://github.com/twitter/finagle/commit/b46a3f96002da7a7d50d197b8907d968fc4640ca

[...]

PR #6430, opened 2017-02-22
Accept two ways to start HTTP/2 over clear text

chhsiao90:
Related to #5857
Just a proposal, not sure that the approach is good or not, but it works fine.

Motivation:

HTTP/2 support upgrade and prior knowledge methodology to start HTTP/2
over clear text. Currently, the http2-client from example only support
for http upgrade. I think we can do a simple dispatch by peak first
bytes of inbound that match to magic bytes or not.

Modifications:

Add ClearTextHttp2ServerDispatcher to support start HTTP/2 via clear
text with two approach. And update example/http2-client to support
this.

Result:

netty HTTP/2 and the example http2-client support for two ways to start
HTTP/2 over clear text.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.codec.http2;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufUtil;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandler;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.util.internal.UnstableApi;
+
+import static io.netty.buffer.Unpooled.*;
+import static io.netty.util.CharsetUtil.UTF_8;
+
+/**
+ * Provide how starting http2 protocol by h2c (Upgrade) or h2 (Prior Knowledge) by peaking the
+ * first bytes is prior knowledge or not, see {@link #PRI_KNOWLEDGE_FIRST_BYTES}
+ * The dispatcher should be the first {@link ChannelInboundHandler} in the
+ * {@link io.netty.channel.ChannelPipeline}
+ * The msg from {@link #channelRead(ChannelHandlerContext, Object)} must be {@link ByteBuf}
+ */
+@UnstableApi
+public abstract class ClearTextHttp2ServerDispatcher extends ChannelInboundHandlerAdapter {
+    private static final String PRI_KNOWLEDGE_MAGIC = "PRI";
+    private static final int PRI_KNOWLEDGE_PEAK_LENGTH = PRI_KNOWLEDGE_MAGIC.length();
+    private static final ByteBuf PRI_KNOWLEDGE_FIRST_BYTES = unreleasableBuffer(
+            directBuffer(PRI_KNOWLEDGE_PEAK_LENGTH).writeBytes(PRI_KNOWLEDGE_MAGIC.getBytes(UTF_8)));
+
+    /**
+     * Peak the first n bytes of msg, and verify that if it matches to {@link #PRI_KNOWLEDGE_FIRST_BYTES}
+     * then assume client want to start HTTP/2 connection with prior knowledge,
+     * call {@link #configurePriorKnowledge(ChannelHandlerContext)} to configure pipeline
+     * for handling HTTP/2 connection. If it didn't match, then call
+     * {@link #configureUpgrade(ChannelHandlerContext)} for HTTP/2 upgrade
+     *
+     * @param ctx {@link ChannelHandlerContext} of this handler
+     * @param msg the message from inbound
+     * @throws IllegalStateException
+     *     if {@code !(msg instanceof io.netty.ByteBuf}
+     *     or if {@code msg.readableBytes() < PRI_KNOWLEDGE_PEAK_LENGTH}
+     */
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+        if (!(msg instanceof ByteBuf)) {
+            throw new IllegalStateException("Only can handle ByteBuf");
+        }
+
+        ByteBuf inbound = (ByteBuf) msg;
+        if (inbound.readableBytes() < PRI_KNOWLEDGE_PEAK_LENGTH) {
+            throw new IllegalStateException(
+                    String.format("must have first %d bytes to determine the protocol",
+                                  PRI_KNOWLEDGE_PEAK_LENGTH));
+        }
+
+        if (ByteBufUtil.equals(PRI_KNOWLEDGE_FIRST_BYTES, 0,
+                               inbound, 0, PRI_KNOWLEDGE_PEAK_LENGTH)) {
+            configurePriorKnowledge(ctx);
+        } else {
+            configureUpgrade(ctx);
+        }
+
+        ctx.fireChannelRead(msg);
+
+        // If the ctx removed before {@code ctx.fireChannelRead(msg)}, then the
+        // message will disappear because {@code ctx.next} would be the tail of
+        // pipeline.
+        ctx.pipeline().remove(this);
+    }
+
+    public abstract void configureUpgrade(ChannelHandlerContext ctx);
+
+    public abstract void configurePriorKnowledge(ChannelHandlerContext ctx);
+}
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
@@ -41,8 +41,6 @@ public abstract class ClearTextHttp2ServerDispatcher extends ByteToMessageDecode
     private static final ByteBuf PRI_KNOWLEDGE_FIRST_BYTES = unreleasableBuffer(
             directBuffer(PRI_KNOWLEDGE_PEAK_LENGTH).writeBytes(PRI_KNOWLEDGE_MAGIC.array()));
 
-    private boolean pipelineConfigured;
-
     /**
      * Peak the first {@link #PRI_KNOWLEDGE_PEAK_LENGTH} bytes of msg, and verify that
      * if it matches to {@link #PRI_KNOWLEDGE_FIRST_BYTES} then assume client want to
@@ -56,12 +54,6 @@ public abstract class ClearTextHttp2ServerDispatcher extends ByteToMessageDecode
      */
     @Override
     public void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws Exception {
-        if (pipelineConfigured) {
-            // After pipeline configured, just forward data to next inbound handler
-            out.add(in.readBytes(in.readableBytes()));
-            return;
-        }
-
         if (in.readableBytes() < PRI_KNOWLEDGE_PEAK_LENGTH) {
             return;
         }
@@ -73,8 +65,9 @@ public abstract class ClearTextHttp2ServerDispatcher extends ByteToMessageDecode
             configureUpgrade(ctx);
         }
 
-        pipelineConfigured = true;
         out.add(in.readBytes(in.readableBytes()));
+
+        ctx.pipeline().remove(this);
     }
 
     /**
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
@@ -19,11 +19,13 @@ import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufUtil;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandler;
-import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.handler.codec.ByteToMessageDecoder;
+import io.netty.util.AsciiString;
 import io.netty.util.internal.UnstableApi;
 
+import java.util.List;
+
 import static io.netty.buffer.Unpooled.*;
-import static io.netty.util.CharsetUtil.UTF_8;
 
 /**
  * Provide how starting http2 protocol by h2c (Upgrade) or h2 (Prior Knowledge) by peaking the
@@ -33,54 +35,49 @@ import static io.netty.util.CharsetUtil.UTF_8;
  * The msg from {@link #channelRead(ChannelHandlerContext, Object)} must be {@link ByteBuf}
  */
 @UnstableApi
-public abstract class ClearTextHttp2ServerDispatcher extends ChannelInboundHandlerAdapter {
-    private static final String PRI_KNOWLEDGE_MAGIC = "PRI";
+public abstract class ClearTextHttp2ServerDispatcher extends ByteToMessageDecoder {
+    private static final AsciiString PRI_KNOWLEDGE_MAGIC = AsciiString.of("PRI");
     private static final int PRI_KNOWLEDGE_PEAK_LENGTH = PRI_KNOWLEDGE_MAGIC.length();
     private static final ByteBuf PRI_KNOWLEDGE_FIRST_BYTES = unreleasableBuffer(
-            directBuffer(PRI_KNOWLEDGE_PEAK_LENGTH).writeBytes(PRI_KNOWLEDGE_MAGIC.getBytes(UTF_8)));
+            directBuffer(PRI_KNOWLEDGE_PEAK_LENGTH).writeBytes(PRI_KNOWLEDGE_MAGIC.array()));
+
+    private boolean pipelineConfigured;
 
     /**
-     * Peak the first n bytes of msg, and verify that if it matches to {@link #PRI_KNOWLEDGE_FIRST_BYTES}
-     * then assume client want to start HTTP/2 connection with prior knowledge,
-     * call {@link #configurePriorKnowledge(ChannelHandlerContext)} to configure pipeline
-     * for handling HTTP/2 connection. If it didn't match, then call
-     * {@link #configureUpgrade(ChannelHandlerContext)} for HTTP/2 upgrade
+     * Peak the first {@link #PRI_KNOWLEDGE_PEAK_LENGTH} bytes of msg, and verify that
+     * if it matches to {@link #PRI_KNOWLEDGE_FIRST_BYTES} then assume client want to
+     * start HTTP/2 connection with prior knowledge, call {@link #configurePriorKnowledge(ChannelHandlerContext)}
+     * to configure pipeline for handling HTTP/2 connection. If it didn't match, then
+     * call {@link #configureUpgrade(ChannelHandlerContext)} for HTTP/2 upgrade
      *
-     * @param ctx {@link ChannelHandlerContext} of this handler
-     * @param msg the message from inbound
      * @throws IllegalStateException
      *     if {@code !(msg instanceof io.netty.ByteBuf}
      *     or if {@code msg.readableBytes() < PRI_KNOWLEDGE_PEAK_LENGTH}
      */
     @Override
-    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
-        if (!(msg instanceof ByteBuf)) {
-            throw new IllegalStateException("Only can handle ByteBuf");
+    public void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws Exception {
+        if (pipelineConfigured) {
+            // After pipeline configured, just forward data to next inbound handler
+            out.add(in.readBytes(in.readableBytes()));
+            return;
         }
 
-        ByteBuf inbound = (ByteBuf) msg;
-        if (inbound.readableBytes() < PRI_KNOWLEDGE_PEAK_LENGTH) {
-            throw new IllegalStateException(
-                    String.format("must have first %d bytes to determine the protocol",
-                                  PRI_KNOWLEDGE_PEAK_LENGTH));
+        if (in.readableBytes() < PRI_KNOWLEDGE_PEAK_LENGTH) {
+            return;
         }
 
         if (ByteBufUtil.equals(PRI_KNOWLEDGE_FIRST_BYTES, 0,
-                               inbound, 0, PRI_KNOWLEDGE_PEAK_LENGTH)) {
+                               in, 0, PRI_KNOWLEDGE_PEAK_LENGTH)) {
             configurePriorKnowledge(ctx);
         } else {
             configureUpgrade(ctx);
         }
 
-        ctx.fireChannelRead(msg);
-
-        // If the ctx removed before {@code ctx.fireChannelRead(msg)}, then the
-        // message will disappear because {@code ctx.next} would be the tail of
-        // pipeline.
-        ctx.pipeline().remove(this);
+        pipelineConfigured = true;
+        out.add(in.readBytes(in.readableBytes()));
     }
 
-    public abstract void configureUpgrade(ChannelHandlerContext ctx);
+    protected abstract void configureUpgrade(ChannelHandlerContext ctx);
 
-    public abstract void configurePriorKnowledge(ChannelHandlerContext ctx);
+    protected abstract void configurePriorKnowledge(ChannelHandlerContext ctx);
 }
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
@@ -77,7 +77,15 @@ public abstract class ClearTextHttp2ServerDispatcher extends ByteToMessageDecode
         out.add(in.readBytes(in.readableBytes()));
     }
 
+    /**
+     * Invoked when dispatcher knows that the client want to start HTTP/2 by upgrade.
+     * Implement this method to configure your pipeline for handling HTTP -> HTTP/2 upgrade.
+     */
     protected abstract void configureUpgrade(ChannelHandlerContext ctx);
 
+    /**
+     * Invoked when dispatcher knows that the client want to start HTTP/2 by sending prior knowledge.
+     * Implement this method to configure your pipeline for HTTP/2 protocol.
+     */
     protected abstract void configurePriorKnowledge(ChannelHandlerContext ctx);
 }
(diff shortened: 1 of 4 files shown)


codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Provide how starting http2 protocol by h2c (Upgrade) or h2 (Prior Knowledge) by peaking the
+ * first bytes is prior knowledge or not, see {@link #PRI_KNOWLEDGE_FIRST_BYTES}
+ * The dispatcher should be the first {@link ChannelInboundHandler} in the
+ * {@link io.netty.channel.ChannelPipeline}
+ * The msg from {@link #channelRead(ChannelHandlerContext, Object)} must be {@link ByteBuf}
+ */
+@UnstableApi
+public abstract class ClearTextHttp2ServerDispatcher extends ChannelInboundHandlerAdapter {
+    private static final String PRI_KNOWLEDGE_MAGIC = "PRI";
```
normanmaurer: Use `AsciiString.valueOf(...)`

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
```diff
@@ -0,0 +1,86 @@
...
+     * @param ctx {@link ChannelHandlerContext} of this handler
+     * @param msg the message from inbound
+     * @throws IllegalStateException
+     *     if {@code !(msg instanceof io.netty.ByteBuf}
+     *     or if {@code msg.readableBytes() < PRI_KNOWLEDGE_PEAK_LENGTH}
+     */
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+        if (!(msg instanceof ByteBuf)) {
+            throw new IllegalStateException("Only can handle ByteBuf");
```
normanmaurer: `UnsupportedMessageTypeException`

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Provide how starting http2 protocol by h2c (Upgrade) or h2 (Prior Knowledge) by peaking the
+ * first bytes is prior knowledge or not, see {@link #PRI_KNOWLEDGE_FIRST_BYTES}
+ * The dispatcher should be the first {@link ChannelInboundHandler} in the
+ * {@link io.netty.channel.ChannelPipeline}
+ * The msg from {@link #channelRead(ChannelHandlerContext, Object)} must be {@link ByteBuf}
+ */
+@UnstableApi
+public abstract class ClearTextHttp2ServerDispatcher extends ChannelInboundHandlerAdapter {
```
normanmaurer: Let this extend `ByteToMessageDecoder` so we can buffer until we have enough bytes ?

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
```diff
@@ -0,0 +1,86 @@
...
+     *     or if {@code msg.readableBytes() < PRI_KNOWLEDGE_PEAK_LENGTH}
+     */
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+        if (!(msg instanceof ByteBuf)) {
+            throw new IllegalStateException("Only can handle ByteBuf");
+        }
+
+        ByteBuf inbound = (ByteBuf) msg;
+        if (inbound.readableBytes() < PRI_KNOWLEDGE_PEAK_LENGTH) {
```
normanmaurer: @chhsiao90 Like said before I think it should extend `ByteToMessageDecoder` so we can handle this case.

example/src/main/java/io/netty/example/http2/helloworld/server/HelloWorldHttp2Handler.java
```diff
@@ -71,6 +70,13 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws E
         ctx.close();
     }
 
+    private Http2Headers http1HeadersToHttp2Headers(FullHttpRequest request) {
```
normanmaurer: static

example/src/main/java/io/netty/example/http2/helloworld/server/Http2ServerInitializer.java
```diff
@@ -86,25 +83,22 @@ private void configureSsl(SocketChannel ch) {
...
             @Override
-            protected void channelRead0(ChannelHandlerContext ctx, HttpMessage msg) throws Exception {
-                // If this handler is hit then no upgrade has been attempted and the client is just talking HTTP.
-                System.err.println("Directly talking: " + msg.protocolVersion() + " (no upgrade was attempted)");
-                ChannelPipeline pipeline = ctx.pipeline();
-                ChannelHandlerContext thisCtx = pipeline.context(this);
-                pipeline.addAfter(thisCtx.name(), null, new HelloWorldHttp1Handler("Direct. No Upgrade Attempted."));
-                pipeline.replace(this, null, new HttpObjectAggregator(maxHttpContentLength));
-                ctx.fireChannelRead(ReferenceCountUtil.retain(msg));
+            public void configurePriorKnowledge(ChannelHandlerContext ctx) {
```
normanmaurer: protected ?

example/src/main/java/io/netty/example/http2/helloworld/server/Http2ServerInitializer.java
```diff
@@ -86,25 +83,22 @@ private void configureSsl(SocketChannel ch) {
      * Configure the pipeline for a cleartext upgrade from HTTP to HTTP/2.0
      */
     private void configureClearText(SocketChannel ch) {
-        final ChannelPipeline p = ch.pipeline();
-        final HttpServerCodec sourceCodec = new HttpServerCodec();
+        ch.pipeline().addLast(new ClearTextHttp2ServerDispatcher() {
+            @Override
+            public void configureUpgrade(ChannelHandlerContext ctx) {
```
normanmaurer: protected ?

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
```diff
@@ -0,0 +1,86 @@
...
+            throw new IllegalStateException(
+                    String.format("must have first %d bytes to determine the protocol",
+                                  PRI_KNOWLEDGE_PEAK_LENGTH));
+        }
+
+        if (ByteBufUtil.equals(PRI_KNOWLEDGE_FIRST_BYTES, 0,
```
normanmaurer: Just use a `ByteProcessor` for this ?

example/src/main/java/io/netty/example/http2/helloworld/server/HelloWorldHttp2Handler.java
```diff
@@ -18,6 +18,8 @@
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufUtil;
 import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.http.FullHttpRequest;
+import io.netty.handler.codec.http.HttpHeaders;
```
netkins: MINOR Remove this unused import 'io.netty.handler.codec.http.HttpHeaders'. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

example/src/main/java/io/netty/example/http2/helloworld/server/HelloWorldHttp2Handler.java
netkins: MINOR Make "http1HeadersToHttp2Headers" a "static" method. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2325)

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
chhsiao90: OK, I will try used ByteToMessageDecoder instead of!

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
chhsiao90: OK!

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
chhsiao90: OK, this check will be removed if we used ```ByteToMessageDecoder``` !

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
chhsiao90: No problem!

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
chhsiao90: Not sure how to check two byte array equality by using ```ByteProcessor```
I only find almost code use ```ByteProcessor``` to find index of specific char or byte.

example/src/main/java/io/netty/example/http2/helloworld/server/HelloWorldHttp2Handler.java
chhsiao90: OK!

example/src/main/java/io/netty/example/http2/helloworld/server/Http2ServerInitializer.java
chhsiao90: OK!

example/src/main/java/io/netty/example/http2/helloworld/server/Http2ServerInitializer.java
chhsiao90: OK!

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
```diff
@@ -33,54 +35,49 @@
...
         } else {
             configureUpgrade(ctx);
         }
 
-        ctx.fireChannelRead(msg);
-
-        // If the ctx removed before {@code ctx.fireChannelRead(msg)}, then the
-        // message will disappear because {@code ctx.next} would be the tail of
-        // pipeline.
-        ctx.pipeline().remove(this);
```
normanmaurer: You still want to remove it

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
```diff
@@ -0,0 +1,83 @@
...
+            configureUpgrade(ctx);
+        }
+
+        pipelineConfigured = true;
+        out.add(in.readBytes(in.readableBytes()));
+    }
+
+    protected abstract void configureUpgrade(ChannelHandlerContext ctx);
+
+    protected abstract void configurePriorKnowledge(ChannelHandlerContext ctx);
```
normanmaurer: Add javadocs

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -43,7 +43,7 @@
  *     }
  * </pre>
  *
- * <h3>Frame detection</h3>
+ * <h3>Frame detection</h3>`
```
normanmaurer: Revert

chhsiao90: Sorry about that :(

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
chhsiao90: No problem!

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
chhsiao90: There is a problem if we try removed the handler from the pipeline here.
When program run to this line, the inbound message hadn't pass to next handler by calling ctx.fireChannelRead.
And if we removed the handler here, then the message will never go to next handler correctly.

Is there any way to remove current handler in the pipeline, but also pass the message to next handler?
Or maybe I can try  ChannelInboundHandlerAdapter implementation, but will buffer the inbound until readable bytes enough?

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
```diff
@@ -0,0 +1,83 @@
...
+
+        if (ByteBufUtil.equals(PRI_KNOWLEDGE_FIRST_BYTES, 0,
+                               in, 0, PRI_KNOWLEDGE_PEAK_LENGTH)) {
+            configurePriorKnowledge(ctx);
+        } else {
+            configureUpgrade(ctx);
+        }
+
+        pipelineConfigured = true;
+        out.add(in.readBytes(in.readableBytes()));
```
normanmaurer: Can you add a test-case which shows the problem when you remove it from the pipeline? It should be possible to remove so I need to see the issue

chhsiao90: Sure!

normanmaurer: let me know one you have it

chhsiao90: Wow, there is no problem when we removed handler from pipeline here..
Because I encounter similar issue with my first implementation with ChannelInboundHandlerAdapter.
So I just thought that ByteToMessageDecoder will have that issue too..

Thanks!
I add some test for that, and will add more test at following commit!

chhsiao90:
All comment addressed.
And thanks for the review, that was really helpful!

nhnFreespirit:
I added something very similar to finagle which is based on Netty: https://github.com/twitter/finagle/commit/b46a3f96002da7a7d50d197b8907d968fc4640ca

There are a number of potential corner cases you might want to deal with, such as the case where not all the bytes of the preface are received in one go (could have been fragmented by the network or silly clients or ...). This also means that making sure all consumed bytes are re-emitted down the pipeline after a match or non match can get a little trickier.

I did originally intend to push my Finagle change down into Netty, but never got around to it.

normanmaurer:
@nhnFreespirit let us know if this PR would work for you and you could just reuse it

nhnFreespirit:
I added comments above. 

Alternatively, feel free to grab and tweak my handler from Finagle as that handles (and has tests for) all the corner cases I could think of and has been battle tested in Finagle for a while (most of the Finagle HTTP/2 end to end test rely on it)

chhsiao90:
@nhnFreespirit - Thanks for the feedback!
I will have a look at finagle implementation.

But for the case "not all the bytes of the preface are received in one go"
I think that might be covered after using ```ByteToMessageDecoder``` from @normanmaurer suggestion and there is a unit test it.
Or maybe there is some edge case that it didn't cover?
```
    @Test
    public void priorKnowledgeWhenFirstInboundNotEnough() throws Exception {
        ByteBuf inbound = Unpooled.directBuffer().writeBytes(
                AsciiString.of("PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n").array());

        try {
            ByteBuf copied = inbound.copy();
            channel.writeInbound(copied.readBytes(2), copied);

            assertTrue(upgradeInbounds.isEmpty());

            assertEquals(1, priorKnowledgeInbounds.size());
            assertEquals(inbound, priorKnowledgeInbounds.get(0));
        } finally {
            inbound.release();
        }
    }
```

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Provide how starting http2 protocol by h2c (Upgrade) or h2 (Prior Knowledge) by peaking the
+ * first bytes is prior knowledge or not, see {@link #PRI_KNOWLEDGE_FIRST_BYTES}
+ * The dispatcher should be the first {@link ChannelInboundHandler} in the
+ * {@link io.netty.channel.ChannelPipeline}
+ * The msg from {@link #channelRead(ChannelHandlerContext, Object)} must be {@link ByteBuf}
+ */
+@UnstableApi
+public abstract class ClearTextHttp2ServerDispatcher extends ByteToMessageDecoder {
+    private static final AsciiString PRI_KNOWLEDGE_MAGIC = AsciiString.of("PRI");
```
mosesn: why not use the full prior knowledge string?  I think it should be "PRI *
   HTTP/2.0\r\n\r\nSM\r\n\r\n" from the http/2 spec.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
```diff
@@ -0,0 +1,84 @@
...
+    private static final int PRI_KNOWLEDGE_PEAK_LENGTH = PRI_KNOWLEDGE_MAGIC.length();
+    private static final ByteBuf PRI_KNOWLEDGE_FIRST_BYTES = unreleasableBuffer(
+            directBuffer(PRI_KNOWLEDGE_PEAK_LENGTH).writeBytes(PRI_KNOWLEDGE_MAGIC.array()));
+
+    /**
+     * Peak the first {@link #PRI_KNOWLEDGE_PEAK_LENGTH} bytes of msg, and verify that
```
mosesn: s/Peak/Peek

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
```diff
@@ -0,0 +1,84 @@
...
+    public void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws Exception {
+        if (in.readableBytes() < PRI_KNOWLEDGE_PEAK_LENGTH) {
+            return;
+        }
+
+        if (ByteBufUtil.equals(PRI_KNOWLEDGE_FIRST_BYTES, 0,
+                               in, 0, PRI_KNOWLEDGE_PEAK_LENGTH)) {
+            configurePriorKnowledge(ctx);
+        } else {
+            configureUpgrade(ctx);
```
mosesn: I don't think we know enough to upgrade here yet.  We need to inspect whether the remote peer followed the normal upgrade process before we can declare an upgrade happened.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Provide how starting http2 protocol by h2c (Upgrade) or h2 (Prior Knowledge) by peaking the
+ * first bytes is prior knowledge or not, see {@link #PRI_KNOWLEDGE_FIRST_BYTES}
+ * The dispatcher should be the first {@link ChannelInboundHandler} in the
+ * {@link io.netty.channel.ChannelPipeline}
+ * The msg from {@link #channelRead(ChannelHandlerContext, Object)} must be {@link ByteBuf}
+ */
+@UnstableApi
+public abstract class ClearTextHttp2ServerDispatcher extends ByteToMessageDecoder {
```
mosesn: Why is this called a dispatcher?  What is it dispatching?  Can we call it CleartextHttp2ServerUpgradeHandler instead?

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
```diff
@@ -0,0 +1,84 @@
...
+        out.add(in.readBytes(in.readableBytes()));
+
+        ctx.pipeline().remove(this);
+    }
+
+    /**
+     * Invoked when dispatcher knows that the client want to start HTTP/2 by upgrade.
+     * Implement this method to configure your pipeline for handling HTTP -> HTTP/2 upgrade.
+     */
+    protected abstract void configureUpgrade(ChannelHandlerContext ctx);
```
mosesn: Instead of its own upgrade mechanism, can we unify with [HttpServerUpgradeHandler][0]?  it already has pretty clear semantics that we should take advantage of.  If possible, we shouldn't have a unique way of upgrading for each mechanism.

[0]: https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/HttpServerUpgradeHandler.java

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
chhsiao90: I think there is no big difference to check all of prior knowledge string or first bytes of it here.
When we find the first 3 bytes is PRI, we assumed the client want to start HTTP/2 by prior knowledge, and let ```Http2ConnectionHandler``` to verify the prior knowledge is correct or not.

But if there is any security issue here, I think check all prior knowledge string here is fine!

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
chhsiao90: Thanks!

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
chhsiao90: I think maybe the function name ```configureUpgrade``` is a little confused.

The configureUpgrade is designed for the case when client is NOT starting HTTP/2 with prior knowledge.

So it could be implement just only support HTTP/1.1, or HTTP/1.1 -> HTTP/2(upgrade), or support both case.
It's depends on how user implement configureUpgrade.

I'm not sure it's a good design or not, but it keeps the flexibility.
I had these idea, because I found almost netty example need user implement their own ChannelInitializer to configure their own pipeline handlers no matter how the purpose is simple, so that netty can fulfill almost every use case.

Above is my idea, I'm agreed that current design is not perfect.
Or maybe let this handler to control the pipeline will be better?
WDYT

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
chhsiao90: Please see [comment](https://github.com/netty/netty/pull/6430#discussion_r103081954) above

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
mosesn: Inspecting something like fifteen more characters is not expensive, let's just do it.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
mosesn: I think it's a reasonable approach, but let's keep the API consistent with what the other server upgrade handlers do. Also you're right, the name is confusing.  We can make it clearer.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
mosesn: This doesn't address my concern—if you look at the API I linked to, it sends userEventFired messages, and provides an Upgradable interface that can be used automatically.

This handler should play nicely with the HttpServerUpgradeHandler, since it will be working with it in the common case.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
chhsiao90: Sure

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
chhsiao90: I'm not sure that I understand about your concern.
Here is my understand about your concern, correct me if I'm wrong.

The ```HttpServerUpgradeHandler``` should be added at the start point. Like 
```
channel.pipeline().addLast(
    new CleartextHttp2ServerUpgradeHandler(),
    httpCodec, httpServerUpgradeHandler);
```
So when the client start HTTP/2  by upgrade, CleartextHttp2ServerUpgradeHandler should just remove itself from pipeline.
When client start HTTP/2 by prior knowledge, we configure pipeline for prior knowledge.

I think that what you had mention at [5857#comment](https://github.com/netty/netty/issues/5857#issue-178968404) and how finagle([code](https://github.com/twitter/finagle/blob/b46a3f96002da7a7d50d197b8907d968fc4640ca/finagle-http2/src/main/scala/com/twitter/finagle/http2/Http2CleartextServerInitializer.scala#L88-L92)) do!

I will say that your approach and mine had different idea and I think both ways make sense,
your approach assume that client will start HTTP/2 upgrade or HTTP/1.1 by default, so it config its pipeline for that, and when found client start HTTP/2 by prior knowledge, it removed handlers configured before.

And my approach is I don't assume what client want to do, I peek the first bytes and configured the pipeline when we know what client wants to do.
So that's why I named the class as a Dispatcher (I agreed it's not a good name), to dispatch ```HTTP/2 via prior knowledge``` or ```HTTP/1 + HTTP/2 via upgrade```.

I think both ways are all good and had its reason.
If we choose your way after discussion, I think I will close this one, and create another one and steal some of your code from finagle :-) , or maybe you would like to have one by yourself.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
Scottmitch: We should also share this definition with what is in `Http2CodecUtil#CONNECTION_PREFACE`.

chhsiao90: +1
When decided to inspect for full preface, then reuse will be better!

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerDispatcher.java
Scottmitch: I agree with @mosesn. We should be able to play nicely with `HttpServerUpgradeHandler`. The finale code should provide a good start and we could make it more general (e.g. not forcing `Http2Codec` into the pipeline [1], no assumptions about handler names [2], etc...) to fit into Netty.

[1] https://github.com/twitter/finagle/blob/b46a3f96002da7a7d50d197b8907d968fc4640ca/finagle-http2/src/main/scala/com/twitter/finagle/http2/transport/PriorKnowledgeHandler.scala#L66
[2] https://github.com/twitter/finagle/blob/b46a3f96002da7a7d50d197b8907d968fc4640ca/finagle-http2/src/main/scala/com/twitter/finagle/http2/transport/PriorKnowledgeHandler.scala#L67

chhsiao90:
Thank for your kindly review, then I think I would close this PR and open another one.
In my plan, there will be one CleartextHttp2ServerUpgradeHandler and PriorKnowlegeHandler just like finagle do and will provide more generic API.

Scottmitch:
sounds good @chhsiao90 ... thanks for all the hard work ... its much appreciated!

PR #6432, opened 2017-02-22
Make methods 'static' where it missed

fenik17:
Motivation:

Calling a static method is faster then dynamic

Modifications:

Add 'static' keyword for methods where it missed

Result:

A bit faster method calls

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/CombinedHttpHeaders.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/CombinedHttpHeaders.java
@@ -227,7 +227,7 @@ public class CombinedHttpHeaders extends DefaultHttpHeaders {
             return sb;
         }
 
-        private CharSequence commaSeparateEscapedValues(CharSequence currentValue, CharSequence value) {
+        private static CharSequence commaSeparateEscapedValues(CharSequence currentValue, CharSequence value) {
             return new StringBuilder(currentValue.length() + 1 + value.length())
                     .append(currentValue)
                     .append(COMMA)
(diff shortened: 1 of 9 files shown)


doom369:
@Scottmitch I have :) - https://shipilev.net/blog/2015/black-magic-method-dispatch/

slandelle:
Emphasizing that method doesn't use instance's state lgtm.
But I wonder if @shipilev's findings hold for private methods (case wasn't tested in his experiment).

doom369:
@slandelle there is such phrase : 

> Static methods are also statically bound, which means we don’t need to resolve what method should we call in runtime. We know exactly what methods we are about to call statically, at compile time. The same thing applies to our own private methods, they are also statically resolvable.

And my tests (copied from above article) : 

```
Benchmark                                     (count)  Mode  Cnt   Score   Error  Units
StaticVsVirtual.dynamic_Abstract_Ref            10000  avgt   50  34.452 ± 0.297  us/op
StaticVsVirtual.dynamic_Abstract_Ref_private    10000  avgt   50  34.346 ± 0.646  us/op
StaticVsVirtual.dynamic_Interface_Ref           10000  avgt   50  37.281 ± 0.739  us/op
StaticVsVirtual.static_Ref                      10000  avgt   50  34.195 ± 0.720  us/op
```

Results are very close and it is hard to make any conclusions here. And I'm too lazy to go deeper :).

normanmaurer:
Cherry-picked into 4.1 (943f4ec7ff6c5753fb79c76cabc043f5e4cdcd32) and 4.0 (f7fa1d5)

PR #6433, opened 2017-02-22
Base64 reduce byte manipulation operations

Scottmitch:
Motivation:
Base64#decode4to3 generally calculates an int value where the contents of the decodabet straddle bytes, and then uses a byte shifting or a full byte swapping operation to get the resulting contents. We can directly calculate the contents and avoid any intermediate int values and full byte swap operations. This will reduce the number of operations required during the decode operation.

Modifications:
- remove the intermediate int in the Base64#decond4to3 method.
- manually do the byte shifting since we are already doing bit/byte manipulations here anyways.

Result:
Base64#decode4to3 requires less operations to compute the end result.

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/base64/Base64.java
+++ b/codec/src/main/java/io/netty/handler/codec/base64/Base64.java
@@ -21,7 +21,6 @@ package io.netty.handler.codec.base64;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
-import io.netty.buffer.ByteBufUtil;
 import io.netty.util.ByteProcessor;
 import io.netty.util.internal.PlatformDependent;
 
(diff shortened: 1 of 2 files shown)


codec/src/main/java/io/netty/handler/codec/base64/Base64.java
```diff
@@ -362,43 +362,63 @@ public boolean process(byte value) throws Exception {
...
+                                               (decodabet[src3] & 0xff) << 16);
+                }
             } catch (IndexOutOfBoundsException ignored) {
                 throw new IllegalArgumentException("not encoded in Base64");
             }
-            // Just directly set it as medium
-            if (dest.order() == ByteOrder.BIG_ENDIAN) {
-                dest.setMedium(destOffset, outBuff);
-            } else {
-                dest.setMedium(destOffset,  ByteBufUtil.swapMedium(outBuff));
```
Scottmitch: @normanmaurer - in addition to `swapMedium` introducing more arithmetic (and function call, temp variables, etc..) it actually includes a conditional ... hooray for not using this method :)

codec/src/main/java/io/netty/handler/codec/base64/Base64.java
```diff
@@ -362,43 +362,63 @@ public boolean process(byte value) throws Exception {
...
+                                               (b1 & 0xf) << 12 |
+                                            // The top half of b1 are the least significant bits after the swap.
+                                               (b1 & 0xf0) >>> 4 |
+                                            // The bottom 2 bits of b2 will be the most significant bits after the swap.
+                                               (b2 & 0x3) << 22 |
+                                            // The remaining 6 bits of b2 remain in the middle.
+                                               (b2 & 0xfc) << 6 |
+                                               (decodabet[src3] & 0xff) << 16);
+                }
             } catch (IndexOutOfBoundsException ignored) {
```
normanmaurer: @Scottmitch I wonder if we should only catch the IOOBE for the array access but not for `setMedium` as if `setMedium` throws its a bug. WDYT ?

codec/src/main/java/io/netty/handler/codec/base64/Base64.java
normanmaurer: +1

codec/src/main/java/io/netty/handler/codec/base64/Base64.java
```diff
@@ -362,43 +362,63 @@ public boolean process(byte value) throws Exception {
...
+                } else {
+                    final byte b1 = decodabet[src1];
+                    final byte b2 = decodabet[src2];
+                    // The goal is to byte swap the BIG_ENDIAN case above. There are 2 interesting things to consider:
+                    // 1. We are byte swapping a 3 byte data type. The top and the bottom byte switch, but the middle
+                    //    remains the same.
+                    // 2. The contents straddles byte boundaries. This means bytes will be pulled apart during the byte
+                    //    swapping process.
+                    dest.setMedium(destOffset, (decodabet[src0] & 0x3f) << 2 |
+                                            // The bottom half of b1 remains in the middle.
```
normanmaurer: Align the comments with the lines below ?

Scottmitch: then they will wrap onto the next line ... I think its just as clear this way and looks nicer IMHO ... wdyt ?

codec/src/main/java/io/netty/handler/codec/base64/Base64.java
Scottmitch: good point ... it looks like we don't even enforce this uniformly in this method ... let me make it more consistent

codec/src/main/java/io/netty/handler/codec/base64/Base64.java
Scottmitch: the wrapping became a non-issue when refactoring the code to protect for IOOBE. fix :)

Scottmitch:
4.1 (77e65fe) 4.0 (492ba62)

PR #6434, opened 2017-02-22
Remove optional dependency on javassist

normanmaurer:
Motivation:

We shipped a javassist based implementation for typematching and logged a confusing debug message about missing javassist. We never were able to prove it really gives any perf improvements so we should just remove it.

Modifications:

- Remove javassist dependency and impl
- Fix possible classloader deadlock as reported by intellij

Result:

Less code to maintain and less confusing log message.


trustin:
Thanks for the clean-up!

normanmaurer:
Cherry-picked into 4.1 (7d08b4fc357e12ee2487e87d8fdcbeee1152e5a0) and 4.0 (5a181d40ee820eb14bc27ade23ba6d364c05d410)

normanmaurer:
@trustin np buddy..

Issue #6428, opened 2017-02-21
SSL Handshake failure closes connection disallowing a response


dlmarion:
It appears that the resolution to #5860 / #5880 in release 4.0.42 changed how a ssl handshake failure behaves. Previously, the SslHandler would call fireExceptionCaught() before setHandshakeFailure(). The order of the execution of these calls was reversed in one of these issues.

I was using the fireExceptionCaught() to send an HTTP redirect response back to the caller. Is there any way to do this in Netty 4.0.42+ ? Thanks!

dlmarion:
FWIW, the purpose of the HTTP redirect was to send the caller to an HTTPS address that would then return a HSTS header. If there is a pattern for doing this differently in Netty, please advise.

Scottmitch:
There are currently at least 2 ways to be notified of a handshake failure:

1. [SslHandler#handshakeFuture](https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/SslHandler.java#L491).
2. [SslHandshakeCompletionEvent](https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/SslHandshakeCompletionEvent.java#L36) will be fired up the pipeline as a user event (use [userEventTriggered](https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/ChannelInboundHandler.java#L61) in your inbound handler to get this).

Both of these are triggered at the same time and I don't think there is any guarantee w.r.t to relationship or ordering between the `fireExceptionCaught` and these methods. Explain why the ordering matters in your case, and what is preventing you from sequencing things "correctly"?

dlmarion:
It appears that the client connection is closed before I have the ability to respond. Set HandshakeFailure closes the connection, then fireExceptionCaught gets called. I will look at the handshake future tomorrow when I get back in the office.

normanmaurer:
@dlmarion hmm that should Born be the case. Please keep us posted

dlmarion:
@normanmaurer Did you find what I was talking about, or do I need to build a test case?

normanmaurer:
@dlmarion yep... stay tuned.

dlmarion:
The code in  SslHandler looks like the following in version 4.0.41:

```
        if (nonSslRecord) {
            // Not an SSL/TLS packet
            NotSslRecordException e = new NotSslRecordException(
[...]

PR #6438, opened 2017-02-23
Make SslHandler configurable to either close or not close channel on …

normanmaurer:
…handshake failure

Motivation:

We should allow to configure the SslHandler / SniHandler to either close or not close the Channel when an error during Handshake happened.

Modifications:

- Add new setter / getter to SslHandler / SniHandler
- Add test-case

Result:

Fixes [#6428].

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
@@ -22,6 +22,7 @@ import java.util.Locale;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufUtil;
+import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelOutboundHandler;
 import io.netty.channel.ChannelPromise;
@@ -61,6 +62,7 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
     private boolean suppressRead;
     private boolean readPending;
     private volatile Selection selection = EMPTY_SELECTION;
+    private volatile boolean closeOnHandshakeFailure = true;
 
     /**
      * Creates a SNI detection handler with configured {@link SslContext}
@@ -91,6 +93,20 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
     @SuppressWarnings("unchecked")
     public SniHandler(AsyncMapping<? super String, ? extends SslContext> mapping) {
         this.mapping = (AsyncMapping<String, SslContext>) ObjectUtil.checkNotNull(mapping, "mapping");
+    }
+
+    /**
+     * Sets if the {@link Channel} should be closed on a handshake failure, default is {@code true}.
+     */
+    public final void setCloseOnHandshakeFailure(boolean closeOnHandshakeFailure) {
+        this.closeOnHandshakeFailure = closeOnHandshakeFailure;
+    }
+
+    /**
+     * Gets if the {@link Channel} should be closed on a handshake failure, default is {@code true}.
+     */
+    public final boolean getCloseOnHandshakeFailure() {
+        return closeOnHandshakeFailure;
     }
 
     /**
@@ -135,9 +151,10 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
                                 NotSslRecordException e = new NotSslRecordException(
                                         "not an SSL/TLS record: " + ByteBufUtil.hexDump(in));
                                 in.skipBytes(in.readableBytes());
-                                ctx.fireExceptionCaught(e);
 
-                                SslUtils.notifyHandshakeFailure(ctx, e);
+                                SslUtils.notifyHandshakeFailure(ctx, e, closeOnHandshakeFailure);
+
+                                ctx.fireExceptionCaught(e);
                                 return;
                             }
                             if (len == SslUtils.NOT_ENOUGH_DATA ||
@@ -346,6 +363,7 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
         SslHandler sslHandler = null;
         try {
             sslHandler = sslContext.newHandler(ctx.alloc());
+            sslHandler.setCloseOnHandshakeFailure(closeOnHandshakeFailure);
             ctx.pipeline().replace(this, SslHandler.class.getName(), sslHandler);
             sslHandler = null;
         } finally {
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -310,6 +310,7 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
     private volatile long handshakeTimeoutMillis = 10000;
     private volatile long closeNotifyFlushTimeoutMillis = 3000;
     private volatile long closeNotifyReadTimeoutMillis;
+    private volatile boolean closeOnHandshakeFailure = true;
 
     /**
      * Creates a new instance.
@@ -442,7 +443,7 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
     }
 
     /**
-     * Sets the timeout  for receiving the response for the close_notify that was triggered by closing the
+     * Sets the timeout for receiving the response for the close_notify that was triggered by closing the
      * {@link Channel}. This timeout starts after the close_notify message was successfully written to the
      * remote peer. Use {@code 0} to directly close the {@link Channel} and not wait for the response.
      */
@@ -459,6 +460,20 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
                     "closeNotifyReadTimeoutMillis: " + closeNotifyReadTimeoutMillis + " (expected: >= 0)");
         }
         this.closeNotifyReadTimeoutMillis = closeNotifyReadTimeoutMillis;
+    }
+
+    /**
+     * Sets if the {@link Channel} should be closed on a handshake failure, default is {@code true}.
+     */
+    public final void setCloseOnHandshakeFailure(boolean closeOnHandshakeFailure) {
+        this.closeOnHandshakeFailure = closeOnHandshakeFailure;
+    }
+
+    /**
+     * Gets if the {@link Channel} should be closed on a handshake failure, default is {@code true}.
+     */
+    public final boolean getCloseOnHandshakeFailure() {
+        return closeOnHandshakeFailure;
     }
 
     /**
@@ -1389,7 +1404,7 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
 
     private void notifyHandshakeFailure(Throwable cause) {
         if (handshakePromise.tryFailure(cause)) {
-            SslUtils.notifyHandshakeFailure(ctx, cause);
+            SslUtils.notifyHandshakeFailure(ctx, cause, closeOnHandshakeFailure);
         }
     }
 
--- a/handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
@@ -32,6 +32,7 @@ import javax.net.ssl.X509TrustManager;
 
 import io.netty.bootstrap.Bootstrap;
 import io.netty.bootstrap.ServerBootstrap;
+import io.netty.buffer.ByteBuf;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelInitializer;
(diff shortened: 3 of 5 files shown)


handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -135,9 +151,10 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) t
                                 NotSslRecordException e = new NotSslRecordException(
                                         "not an SSL/TLS record: " + ByteBufUtil.hexDump(in));
                                 in.skipBytes(in.readableBytes());
-                                ctx.fireExceptionCaught(e);
 
-                                SslUtils.notifyHandshakeFailure(ctx, e);
+                                SslUtils.notifyHandshakeFailure(ctx, e, closeOnHandshakeFailure);
+
+                                ctx.fireExceptionCaught(e);
```
normanmaurer: This was done to make it consistent in terms of ordering with `SslHandler`

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -94,6 +96,20 @@ public SniHandler(AsyncMapping<? super String, ? extends SslContext> mapping) {
...
+     * Sets if the {@link Channel} should be closed on a handshake failure, default is {@code true}.
+     */
+    public final void setCloseOnHandshakeFailure(boolean closeOnHandshakeFailure) {
+        this.closeOnHandshakeFailure = closeOnHandshakeFailure;
+    }
+
+    /**
+     * Gets if the {@link Channel} should be closed on a handshake failure, default is {@code true}.
+     */
+    public final boolean getCloseOnHandshakeFailure() {
```
trustin: How about: `setShouldCloseOnHandshakeFailure()` and `shouldCloseOnHandshakeFailure()`?

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -462,6 +463,20 @@ public final void setCloseNotifyReadTimeoutMillis(long closeNotifyReadTimeoutMil
...
+     * Sets if the {@link Channel} should be closed on a handshake failure, default is {@code true}.
+     */
+    public final void setCloseOnHandshakeFailure(boolean closeOnHandshakeFailure) {
+        this.closeOnHandshakeFailure = closeOnHandshakeFailure;
+    }
+
+    /**
+     * Gets if the {@link Channel} should be closed on a handshake failure, default is {@code true}.
+     */
+    public final boolean getCloseOnHandshakeFailure() {
```
trustin: Ditto

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
normanmaurer: sounds good :)

normanmaurer:
@trustin addressed your comments. PTAL again

trustin:
> @trustin addressed your comments. PTAL again

Seems like nothing has changed yet?

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
netkins: MAJOR A getX() method which returns a boolean should be named isX() [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3ABooleanGetMethodName)

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
netkins: MAJOR A getX() method which returns a boolean should be named isX() [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3ABooleanGetMethodName)

normanmaurer:
@trustin ups... forgot to add the files. PTAL

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -94,6 +96,20 @@ public SniHandler(AsyncMapping<? super String, ? extends SslContext> mapping) {
...
+     * Sets if the {@link Channel} should be closed on a handshake failure, default is {@code true}.
+     */
+    public final void setShouldCloseOnHandshakeFailure(boolean closeOnHandshakeFailure) {
+        this.closeOnHandshakeFailure = closeOnHandshakeFailure;
+    }
+
+    /**
+     * Gets if the {@link Channel} should be closed on a handshake failure, default is {@code true}.
+     */
+    public final boolean getShouldCloseOnHandshakeFailure() {
```
trustin: getShould -> should ? i.e. `if (handler.shouldCloseOnHandshakeFailure()) { ... }`

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -462,6 +463,20 @@ public final void setCloseNotifyReadTimeoutMillis(long closeNotifyReadTimeoutMil
...
+     * Sets if the {@link Channel} should be closed on a handshake failure, default is {@code true}.
+     */
+    public final void setShouldCloseOnHandshakeFailure(boolean closeOnHandshakeFailure) {
+        this.closeOnHandshakeFailure = closeOnHandshakeFailure;
+    }
+
+    /**
+     * Gets if the {@link Channel} should be closed on a handshake failure, default is {@code true}.
+     */
+    public final boolean getShouldCloseOnHandshakeFailure() {
```
trustin: Ditto

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
normanmaurer: actually I am not sure and maybe I even think we should just revert it to what I had first as we use set and get for all the others params on the class. WDYT ?

netkins: MAJOR A getX() method which returns a boolean should be named isX() [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3ABooleanGetMethodName)

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
netkins: MAJOR A getX() method which returns a boolean should be named isX() [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3ABooleanGetMethodName)

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
trustin: Ahh, then it's better being consistent. How about just using `closeOnHandshakeFailure` back again?

normanmaurer:
@trustin ok reverted... PTAL again

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
```diff
@@ -75,6 +77,58 @@
...
+                // was not closed and fail otherwise.
+                ctx.pipeline().remove(SslHandler.class);
+                ctx.writeAndFlush(Unpooled.wrappedBuffer(bytes));
+                ctx.fireExceptionCaught(cause);
+            }
+        });
+
+        try {
+            ch.writeInbound(Unpooled.wrappedBuffer(bytes));
+        } catch (Throwable e) {
```
Scottmitch: add a `fail();` here?

normanmaurer: Good catch

normanmaurer:
@Scottmitch will add the test, good idea

normanmaurer:
@Scottmitch PTAL again.

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
netkins: MAJOR A getX() method which returns a boolean should be named isX() [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3ABooleanGetMethodName)

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
netkins: MAJOR A getX() method which returns a boolean should be named isX() [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3ABooleanGetMethodName)

johnou:
Instead of pulling this in please see my proposed workaround here https://github.com/netty/netty/issues/6428#issuecomment-282250260

Scottmitch:
I agree that @johnou's proposal will also work ... and I recall recommending this approach in the past. A "negative" is that details/knowledge of TLS protocol leak into this handler ... although minor and a utility method is provided by `SslHandler`.

One thing to consider is how would user react to a handshake failure in general.  Just knowing the handshake has failed requires users to take extra action to determine why (client didn't attempt TLS, handshake legitimately failed, etc...) so I can see value in not supporting "optional close" as it seems unlikely in general if the TLS connection fails that a user would be OK with doing plaintext (why wouldn't they just do this to start with), and may get folks into more trouble if they ignore the reason why the handshake failed (or incorrectly interpret the exception etc...).

dlmarion:
FWIW, my code that uses this was trying to do something nice for the user by implementing the HTTP Strict Transport Security mechanism so that future http requests in the browser would automatically switch to https. If you are not familiar with HSTS, it's a header that is returned in a secure session that is recognized by the browser and the browser will remember to send all traffic to that address over https in the future. If connected via non-secure means (http), then I have to send a HTTP redirect to the HTTPS address so that I can return the header.

If you feel that closing the connection on a handshake failure with no further action is the right thing to do, I'm fine with that and will just remove the feature from my code. However, if you want to support it, I'd rather not implement the solution that @johnou suggested. I would rather use the change in this PR and close the connection myself, or extend the SslHandler and override a method that is called before the connection is closed.

I'm also open to suggestions on how to implement HSTS differently than I described.

dlmarion:
Maybe a compromise, maybe have the SslHandler optionally send a HTTP redirect before closing the connection. My `exceptionCaught()` looks like:
```
        if (cause instanceof NotSslRecordException) {
            FullHttpResponse response = new DefaultFullHttpResponse(HttpVersion.HTTP_1_1,
                    HttpResponseStatus.MOVED_PERMANENTLY, Unpooled.EMPTY_BUFFER);
            response.headers().set(Names.LOCATION, redirectAddress);
            ctx.writeAndFlush(response);
        }
```

Scottmitch:
@dlmarion - You could check for an instance of `NotSslRecordException` in the handshake failure but that seems a bit fragile to me. For example in PR https://github.com/netty/netty/pull/6365 this exception may no longer be thrown any more, and instead we rely upon the SSLEngine to determine what is valid. I may need to investigate this to preserve compatibility but relying on something that is not explicitly part of the interface contract of `SslHandler` and not generated by the SSLEngine implementation may lead to issues. Also you could get a `NotSslRecordException` at other times during the connection at which it may not be appropriate to send this response.

I think the safest approach to handle the "is my peer attempting to do SSL" question is to do what is suggested in https://github.com/netty/netty/issues/6428#issuecomment-282250260. Since this is reasonably general we could maintain this handler in Netty ... @johnou any interest in submitting a PR?

dlmarion:
Inserting a handler in the pipeline before the SSL handler, that I could use to detect if the connection is encrypted or not, seems like an OK approach. It would be nice if it was part of the Netty library and not something custom.

Scottmitch:
> It would be nice if it was part of the Netty library and not something custom.

As mentioned in https://github.com/netty/netty/pull/6438#issuecomment-282509944 a PR for a handler which does this would be welcome 👍

johnou:
@dlmarion can you provide me with a minimal Netty client that asserts the desired behaviour?

dlmarion:
@johnou - here you go: https://github.com/dlmarion/netty-hsts

johnou:
@dlmarion thanks! I'll try and push something later this week or sooner depending on how our release goes, crunch time right now.

johnou:
I had time for a quick hack at the first iteration, you can check it out here https://gist.github.com/johnou/7cdfd97041921aaa173f03ae68ee2342 feedback appreciated.

johnou:
@dlmarion how about something like this? https://gist.github.com/johnou/f5a88579daa439bf374f366b8791248b

dlmarion:
I will test it out and get back to you. Thx.

dlmarion:
@johnou I tested this and it appears to work. The only change I made was to add an alternate constructor to the OptionalSslHandler so that I could set the name. I have code later in the pipeline that is looking for a specific name, and this code uses `OptionalSslHandler.class.getName()`. My code looks like:
```
    private final SslContext sslContext;
    private final String handlerName;

    public OptionalSslHandler(SslContext sslContext) {
        this.sslContext = sslContext;
        this.handlerName = OptionalSslHandler.class.getName();
    }

    public OptionalSslHandler(SslContext sslContext, String handlerName) {
        this.sslContext = sslContext;
        this.handlerName = handlerName;
    }
```

normanmaurer:
@johnou would you be willing to provide a PR for your handler so we can merge it into netty ?

normanmaurer:
Let me close this then for now... @johnou will hopefully do a PR for his handler 👍

johnou:
@normanmaurer just writing a unit test 👌

PR #6439, opened 2017-02-23
Throw if SSLParameters contains settings that are not supported by Re…

normanmaurer:
…ferenceCountedOpenSslEngine

Motivation:

We not support all SSLParameters settings so we should better throw if a user try to use them.

Modifications:

- Check for unsupported parameters
- Add unit test

Result:

Less surprising behavior.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -34,14 +34,17 @@ import io.netty.util.internal.logging.InternalLoggerFactory;
 
 import java.nio.ByteBuffer;
 import java.nio.ReadOnlyBufferException;
+import java.security.AlgorithmConstraints;
 import java.security.Principal;
 import java.security.cert.Certificate;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
+import javax.net.ssl.SNIMatcher;
 import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLEngineResult;
 import javax.net.ssl.SSLException;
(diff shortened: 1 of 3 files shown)


normanmaurer:
@nmittler you may want to do the same in conscrypt

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1555,10 +1558,34 @@ public final synchronized SSLParameters getSSLParameters() {
 
     @Override
     public final synchronized void setSSLParameters(SSLParameters sslParameters) {
-        super.setSSLParameters(sslParameters);
-
         int version = PlatformDependent.javaVersion();
         if (version >= 7) {
+            if (sslParameters.getAlgorithmConstraints() != null) {
+                throw new IllegalArgumentException("AlgorithmConstraints are not supported");
```
trustin: How about append a dot(.) since this is a sentence?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1555,10 +1558,34 @@ public final synchronized SSLParameters getSSLParameters() {
...
+            }
+
+            if (version >= 8) {
+                Collection<SNIMatcher> matchers = sslParameters.getSNIMatchers();
+                if (matchers != null && !matchers.isEmpty()) {
+                    throw new IllegalArgumentException("SNIMatchers are not supported");
```
trustin: ditto

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
```diff
@@ -557,6 +566,67 @@ private void testWrapDstBigEnough(SSLEngine engine, int srcLen) throws SSLExcept
...
+            parameters.setAlgorithmConstraints(new AlgorithmConstraints() {
+                @Override
+                public boolean permits(
+                        Set<CryptoPrimitive> primitives, String algorithm, AlgorithmParameters parameters) {
+                    return false;
+                }
+
+                @Override
+                public boolean permits(
+                        Set<CryptoPrimitive> primitives, Key key) {
```
trustin: How about joining with the previous line?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1555,10 +1558,34 @@ public final synchronized SSLParameters getSSLParameters() {
...
         int version = PlatformDependent.javaVersion();
         if (version >= 7) {
+            if (sslParameters.getAlgorithmConstraints() != null) {
+                throw new IllegalArgumentException("AlgorithmConstraints are not supported.");
+            }
+
+            if (version >= 8) {
+                Collection<SNIMatcher> matchers = sslParameters.getSNIMatchers();
+                if (matchers != null && !matchers.isEmpty()) {
+                    throw new IllegalArgumentException("SNIMatchers are not supported.");
```
normanmaurer: @Scottmitch FYI we can easily support SNIMatchers tho with a few changes in netty-tcnative. But I would like to fix this after this is pulled in as this needs more testing etc.

normanmaurer:
Cherry-picked into 4.1 (325cc84) and 4.0 (127f6e9)

nmittler:
@normanmaurer added it to the list in https://github.com/google/conscrypt/issues/31.  Thanks!

Issue #6436, opened 2017-02-23
Unpooled buffer acquired from pooled allocator may have .alloc() returning null


nayato:
### Expected behavior
- .alloc() returns allocator that buffer originated from. Backed by following behavior:
 - `allocator` field is reset on `deallocate()`.
 - `allocator` field is set properly in `initUnpooled()`
### Actual behavior
`.alloc()` returns null if buffer is initialized through `initUnpooled()` and was just created. It also may return wrong allocator if it was previously backed by chunk from different pooled buffer allocator.

### Steps to reproduce
https://github.com/netty/netty/blob/66b9be3a469a2cdcc5d18a8b94c679940ce002a9/buffer/src/main/java/io/netty/buffer/PooledByteBuf.java#L128 -- `alloc()` is backed by `allocator` field. As object is pooled it must be initialized during `init()` or `initUnpooled()`

https://github.com/netty/netty/blob/66b9be3a469a2cdcc5d18a8b94c679940ce002a9/buffer/src/main/java/io/netty/buffer/PooledByteBuf.java#L52 -- `allocator` initialized (as expected)

https://github.com/netty/netty/blob/66b9be3a469a2cdcc5d18a8b94c679940ce002a9/buffer/src/main/java/io/netty/buffer/PooledByteBuf.java#L60 -- `allocator` is not initialized.

### Netty version
current 4.1 branch code.

normanmaurer:
Good catch let me fix it

normanmaurer:
@nayato PTAL https://github.com/netty/netty/pull/6440

normanmaurer:
Fixed by https://github.com/netty/netty/pull/6440

PR #6440, opened 2017-02-23
Ensure PooledByteBuf.initUnpooled(...) correctly set the allocator

normanmaurer:
Motivation:

Commit 8dda984afe9477fa09fe166305381d9ed1afa7c4 introduced a regression which lead to the situation that the allocator is not set when PooledByteBuf.initUnpooled(...) is called. Thus it was possible that PooledByteBuf.alloc() returns null or the wrong allocator if multiple PooledByteBufAllocator are used in an application.

Modifications:

- Correctly set the allocator
- Add test-case

Result:

Fixes [#6436].


normanmaurer:
Cherry-picked into 4.1 (deb9092) and 4.0 (4ba3cd8)

PR #4819, opened 2016-02-02
EventExecutorGroup unwrap

Scottmitch:
Motivation:
The current channel implementations are tied to their specific type of EventLoop. This limits the ability to wrap the EventLoop / EventLoopGroup to extend the functionality of these executors. Supporting a method like ByteBuf.unwrap will allow the functionality to be extended.

Modifications:
- Add an unwrap method to EventExecutorGroup, EventExecutor, EventLoopGroup, and EventLoop
- Remove EventExecutorGroup.children() as introduced by 7dc63cc. children() allows the call sites to make assumptions about the type of `EventExecutor` that is maintained by the EventExecutorGroup, which may be invalid if they are actually wrapped. Netty is currently not taking advantage of the implicit cast and is still using EventExcecutorGroup as an iterator anyways.
- Make sure the Channels and places which care about a specific type of EventLoop get unwrapped / checked correctly

Result:
EventLoops can be wrapped to extend functionality.

PR #6441, opened 2017-02-23
Make EpollEventLoopGroup non final

michaelandrepearce:
This is to make it inline with NioEventLoopGroup, which is non final, allowing extension where needed. 

As detailed in wiki it should be a straight swap but this isn't the case, as class's in user code extending NioEventLoopGroup cannot be changed to extend EpollEventLoopGroup

Motivation:

Modification:

Result:

Fixes #<GitHub issue number>. 

### DIFF
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoopGroup.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoopGroup.java
@@ -32,7 +32,7 @@ import java.util.concurrent.ThreadFactory;
  * {@link EventLoopGroup} which uses epoll under the covers. Because of this
  * it only works on linux.
  */
-public final class EpollEventLoopGroup extends MultithreadEventLoopGroup {
+public class EpollEventLoopGroup extends MultithreadEventLoopGroup {
 
     /**
      * Create a new instance using the default number of threads and the default {@link ThreadFactory}.


normanmaurer:
@michaelandrepearce honestly if I could change it I would make `NioEventLoopGroup`  final but I can't . Why you want to to extend it ?

michaelandrepearce:
Apache Artemis code base has a class that extends it. Currently it only uses NIO. We wanted to quickly add support for it to use the EPOLL and make a PR there.  As should be a straight swap we literally doing a find replace exercise. Alas we hit this small issue.

Why can you not make it non final?

Regards
Mike

normanmaurer:
I can I just want to understand the use case. Can you point me to The class that extend it?

michaelandrepearce:
Sure

https://github.com/apache/activemq-artemis/blob/master/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/SharedNioEventLoopGroup.java

normanmaurer:
@michaelandrepearce IMHO a better approach to your impl would be to create your own `EventLoopGroup` which wraps another `EventLoopGroup` and just delegate everything exception the methods that are overridden. This way you can also share the same code for epoll / nio etc.

Scottmitch:
@michaelandrepearce - Note that allowing extension to EventLoopGroup can be non-trivial and create issues. I previously explored this path to be able to wrap event loop groups and decided to abandon the approach (see https://github.com/netty/netty/pull/4819#issuecomment-179826173). It looks like your use case may not be subject to these issues but it is something to be aware of.

I agree with @normanmaurer and would like to avoid supporting extending the EventLoopGroup implementations.

rkapsi:
My 2c, looks to me like FooEventLoopGroup and FooEventLoop should be an interfaces. Not a big fan of `instanceof` dependencies (see `AbstractChannel#isCompatible()`) concrete Classes as it breaks composability. The need to extend is just a symptom.

normanmaurer:
So I would like to close this as won't fix... @Scottmitch agree ?

michaelandrepearce:
I'm happy to close.

As you've pointed out we can use a delegation pattern to get around this issue.

Thanks
Mike

Scottmitch:
@rkapsi - agreed the current approach has limitations. Perhaps we can consider this as an improvement for a release that allows API breakage.

Issue #6349, opened 2017-02-13
Change default value of io.netty.maxDirectMemory ?


normanmaurer:
We introduced `io.netty.maxDirectMemory` to allow creating a Direct ByteBuffer which not is allocated through `Bits.reserveMemory` as this method and the corresponding `Bits.unreserveMemory` were static synchronized. This reduced the overhead but also made it harder to monitor direct memory usage as these are not exported through JMX anymore by Java itself. 

Since Java version 8u72 these methods are not static synchronized anymore so we may want to consider change the default back to use the normal allocation / deallocate path:
https://bugs.openjdk.java.net/browse/JDK-6857566

@Scottmitch @trustin @nmittler @snazy WDYT ?

snazy:
Yes, the methods are not (or no longer) synchronized. But IMO the real issue is in the implementation in {{Bits}} (forcing GC) and how the allocations are managed (linked list).

However, having some MXBean that allows people to get metrics about direct memory usage would be great.

Scottmitch:
Sounds good in general. However didn't this allow use to use a `realloc` in a few places. Maybe worth benchmarking?

nmittler:
Agreed with @Scottmitch ... assuming benchmarks are good, SGTM.

normanmaurer:
@snazy I am more concerned about the static synchronized  which is gone now. I think it would "good enough" to just use the default JDK impl as default and if people really want to make use of this "optimisation" they are free to use it. This way the metrics which are exposed via JMX etc are at least correct. 

@Scottmitch yeah we may be able to `realloc` in some places. That said I think as long as we allow the user to enable it again (via the property) its fine.  WDYT ?

normanmaurer:
@trustin any comments ?

Scottmitch:
My only concern is if performance is impacted in a meaningful way. If there is impact and we still want to change the default then we should call out the change in the Release notes.

johnou:
@snazy don't forget that sleep.. :)

```java
    // These methods should be called whenever direct memory is allocated or
    // freed.  They allow the user to control the amount of direct memory
    // which a process may access.  All sizes are specified in bytes.
    static void reserveMemory(long size, int cap) {
        synchronized (Bits.class) {
            if (!memoryLimitSet && VM.isBooted()) {
                maxMemory = VM.maxDirectMemory();
                memoryLimitSet = true;
            }
            // -XX:MaxDirectMemorySize limits the total capacity rather than the
[...]

PR #6444, opened 2017-02-23
Add benchmarks for UnpooledUnsafeNoCleanerDirectByteBuf vs UnpooledUnsafeDirectByteBuf

normanmaurer:
Motivation:

Issue [#6349] brought up the idea to not use UnpooledUnsafeNoCleanerDirectByteBuf by default. To decide what to do a benchmark is needed.

Modifications:

Add benchmarks for UnpooledUnsafeNoCleanerDirectByteBuf vs UnpooledUnsafeDirectB
yteBuf

Result:

Better idea about impact of using UnpooledUnsafeNoCleanerDirectByteBuf.

### DIFF
--- a/microbench/src/main/java/io/netty/buffer/ByteBufNoCleanerAllocReleaseBenchmark.java
+++ b/microbench/src/main/java/io/netty/buffer/ByteBufNoCleanerAllocReleaseBenchmark.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.buffer;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.annotations.Warmup;
+
+@BenchmarkMode(Mode.Throughput)
+@Threads(16)
+@Warmup(iterations = 5)
+@Measurement(iterations = 10)
+public class ByteBufNoCleanerAllocReleaseBenchmark extends AbstractByteBufNoCleanerBenchmark {
+
+    @Param({ "64", "1024", "8192" })
+    public int initialCapacity;
+
+    @Benchmark
+    public boolean allocateRelease() {
+        ByteBuf buffer = bufferType.newBuffer(initialCapacity);
+        return buffer.release();
+    }
+}
--- a/microbench/src/main/java/io/netty/buffer/ByteBufNoCleanerChangeCapacityBenchmark.java
+++ b/microbench/src/main/java/io/netty/buffer/ByteBufNoCleanerChangeCapacityBenchmark.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.buffer;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.annotations.Warmup;
+
+@BenchmarkMode(Mode.Throughput)
+@Threads(16)
+@Warmup(iterations = 5)
+@Measurement(iterations = 10)
+public class ByteBufNoCleanerChangeCapacityBenchmark extends AbstractByteBufNoCleanerBenchmark {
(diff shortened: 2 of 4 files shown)


Scottmitch:
it would be good to understand the performance impacts of this (see https://github.com/netty/netty/issues/6349#issuecomment-282079611)

normanmaurer:
I will write a Benchmark tomorrow and add it here

normanmaurer:
So as @Scottmitch pushed me to write benchmarks I did it... And guess what, I owe him a few 🍺 as he was right and the impact is "huge". While my argument still stands that this is mostly for the `UnpooledByteBufAllocator` as the pooled one allocates in chunks of 16mb and so not does so many allocations anyway I think this shows we should most likely not change the default but expose the used  direct memory on the allocator directly (as I suggested at some point). 

```
Result "io.netty.buffer.ByteBufNoCleanerAllocReleaseBenchmark.allocateRelease":
  1706022.219 ±(99.9%) 43706.904 ops/s [Average]
  (min, avg, max) = (1647640.240, 1706022.219, 1819280.165), stdev = 50332.935
  CI (99.9%): [1662315.316, 1749729.123] (assumes normal distribution)


# Run complete. Total time: 00:03:13

Benchmark                                                     (bufferType)  (initialCapacity)   Mode  Cnt        Score        Error  Units
ByteBufNoCleanerAllocReleaseBenchmark.allocateRelease  UNPOOLED_NO_CLEANER                 64  thrpt   20  7872071.165 ± 250049.916  ops/s
ByteBufNoCleanerAllocReleaseBenchmark.allocateRelease  UNPOOLED_NO_CLEANER               1024  thrpt   20  7433399.305 ± 144013.542  ops/s
ByteBufNoCleanerAllocReleaseBenchmark.allocateRelease  UNPOOLED_NO_CLEANER               8192  thrpt   20  6377821.325 ± 272251.871  ops/s
ByteBufNoCleanerAllocReleaseBenchmark.allocateRelease             UNPOOLED                 64  thrpt   20  2084158.678 ±  59913.052  ops/s
ByteBufNoCleanerAllocReleaseBenchmark.allocateRelease             UNPOOLED               1024  thrpt   20  1962968.283 ±  65960.593  ops/s
ByteBufNoCleanerAllocReleaseBenchmark.allocateRelease             UNPOOLED               8192  thrpt   20  1706022.219 ±  43706.904  ops/s


Result "io.netty.buffer.ByteBufNoCleanerChangeCapacityBenchmark.capacityChange":
  21.521 ±(99.9%) 1.638 ops/s [Average]
  (min, avg, max) = (18.677, 21.521, 25.499), stdev = 1.886
  CI (99.9%): [19.883, 23.159] (assumes normal distribution)


# Run complete. Total time: 00:01:40

Benchmark                                                      (bufferType)  (initialCapacity)   Mode  Cnt     Score    Error  Units
ByteBufNoCleanerChangeCapacityBenchmark.capacityChange  UNPOOLED_NO_CLEANER               1024  thrpt   20  2128.061 ± 71.963  ops/s
ByteBufNoCleanerChangeCapacityBenchmark.capacityChange             UNPOOLED               1024  thrpt   20    21.521 ±  1.638  ops/s
```
The code for the benchmarks is in the second commit.

normanmaurer:
Also FYI @snazy @trustin @johnou @doom369

normanmaurer:
So I guess I would propose to close this PR and just merge the benchmark at the end ;)

johnou:
@normanmaurer out of interest, did you spot where the bottleneck is?

normanmaurer:
@johnou did not check but as the only difference is basically go through Bits.reserveMemory and Bits.unreserveMemory or not got through these I suspect these are guilty for the allocateRelease benchmark and for the changeCapacity its the stated thing + we can not use realloc when using normal ByteBuffer.

microbench/src/main/java/io/netty/buffer/ByteBufNoCleanerChangeCapacityBenchmark.java
```diff
@@ -0,0 +1,49 @@
...
+    public int initialCapacity;
+
+    @Param
+    public ByteBufType bufferType;
+
+    @Benchmark
+    public boolean capacityChange() {
+        ByteBuf buffer = bufferType.newBuffer(initialCapacity);
+        // Change capacity until we would exceed the 1mb per thread limit
+        for (int newCapacity = initialCapacity << 1; newCapacity <= MAX_DIRECT_MEMORY_PER_THREAD;
```
normanmaurer: I did this because I wanted to minimise the allocations of new `ByteBuf` for this benchmark

johnou:
fwiw https://blogs.oracle.com/alanb/entry/monitoring_direct_buffers suggests allocating a large buffer and slicing that for allocating smaller buffers.

normanmaurer:
@johnou which is more or less exactly what we do in our pooled allocator + other things ;)

johnou:
@normanmaurer right, that's why I'm wondering if it would be more realistic to benchmark this with the allocator.

normanmaurer:
@johnou no it would not at least not with 4.0 which uses the `UnpooledByteBufAllocator` by default and so does exactly what I am doing here. Like I said with the `PooledByteBufAllocator` I think there will no difference noticeable as the allocation / deallocation will happen in 16mb chunks and so not really often. That said what we could do is use the  "no cleaner" thing when `UnpooledByteBufAllocator` is used and the normal allocation path when `PooledByteBufAllocator` is used. Not sure this makes sense tho.  Also its already possible to get the number of used memory of `PooledByteBufAllocator `.

microbench/src/main/java/io/netty/buffer/ByteBufNoCleanerAllocReleaseBenchmark.java
```diff
@@ -0,0 +1,43 @@
...
+@Threads(16)
+@Warmup(iterations = 5)
+@Measurement(iterations = 10)
+public class ByteBufNoCleanerAllocReleaseBenchmark extends AbstractByteBufNoCleanerBenchmark {
+
+    @Param({ "64", "1024", "8192" })
+    public int initialCapacity;
+
+    @Param
+    public ByteBufType bufferType;
```
netkins: MAJOR "bufferType" is the name of a field in "AbstractByteBufNoCleanerBenchmark". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2387)

microbench/src/main/java/io/netty/buffer/ByteBufNoCleanerChangeCapacityBenchmark.java
```diff
@@ -0,0 +1,49 @@
...
+@Warmup(iterations = 5)
+@Measurement(iterations = 10)
+public class ByteBufNoCleanerChangeCapacityBenchmark extends AbstractByteBufNoCleanerBenchmark {
+    private static final int MAX_DIRECT_MEMORY_PER_THREAD = 1024 * 1024; // 1 mb per thread.
+
+    @Param("1024")
+    public int initialCapacity;
+
+    @Param
+    public ByteBufType bufferType;
```
netkins: MAJOR "bufferType" is the name of a field in "AbstractByteBufNoCleanerBenchmark". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2387)

Scottmitch:
thanks for the benchmarks @normanmaurer ... I agree lets ditch the property change and just take the benchmarks.

normanmaurer:
Cherry-picked into 4.1 (90a6104) and 4.0 (b372a2f19fafc37d58492a6783256fc79b935814)

PR #6445, opened 2017-02-23
Move all the microbenchmark code into one directory.

normanmaurer:
Motivation:

Allmost all our benchmarks are in src/main/java but a few are in src/test/java. We should make it consistent.

Modifications:

Move everything to src/main/java

Result:

Consistent code base.


normanmaurer:
Noticed this while running the ssl benchmarks. Also I think we should do the same in 4.0 which means means move all to src/main/java WDYT ?

Scottmitch:
They were moved from test to main a while ago [1] and moving them all to the same location sgtm.

[1] https://github.com/netty/netty/commit/f3fb77f4bce9d11d7b54878e91728674f6ee5342

normanmaurer:
Cherry-picked into 4.1 (a80d341) . Will do another PR for 4.0

PR #6446, opened 2017-02-23
Move all the microbenchmark code into src/main/java

normanmaurer:
Motivation:

All our benchmarks are in src/main/java in 4.1.x . We should make it consistent.

Modifications:

Move everything to src/main/java

Result:

Consistent code base.


normanmaurer:
Cherry-picked into 4.0 (00dde3224cf1de11d3a075746c5813e750caafa8)

PR #6447, opened 2017-02-23
Correctly have exceptions thrown from decode(...) method be wrapped w…

normanmaurer:
…ith DecodingException

Motivation:

Normally if a decoder produces an exception its wrapped with DecodingException. This is not the cause for NotSslRecordException in SslHandler and SniHandler.

Modifications:

Just throw the NotSslRecordException exception for decode(...) and so ensure its correctly wrapped in a DecodingException before its passed through the pipeline.

Result:

Consist behavior.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
@@ -36,6 +36,7 @@ import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.FutureListener;
 import io.netty.util.concurrent.Promise;
 import io.netty.util.internal.ObjectUtil;
+import io.netty.util.internal.PlatformDependent;
 import io.netty.util.internal.logging.InternalLogger;
 import io.netty.util.internal.logging.InternalLoggerFactory;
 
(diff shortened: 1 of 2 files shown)


normanmaurer:
@Scottmitch @trustin I noticed this while addressing comments of @Scottmitch  in https://github.com/netty/netty/pull/6438.  WDYT ?

Scottmitch:
`This is not the cause` -> `This is not the case`

Scottmitch:
what about the following:
- SniHandler:249
- SniHandler:331 -> we could call `fireExceptionCaught` or throw conditionally based upon if it is called from the future listener or not.

normanmaurer:
@Scottmitch good point... missed these so let me fix this.

That said I was wondering what you and @trustin think about the change in general. Its a change in behaviour but I would argue the old one was a bug. WDYT ?

Scottmitch:
@normanmaurer - I agree with this being considered a bug.

normanmaurer:
@trustin wdyt ?

normanmaurer:
@Scottmitch didnt I adjust it to your comments ? I am confused 😆

normanmaurer:
Cherry-picked into 4.1 (ca2c349a4a95f79e1f8b5315bf7e7617386f266e) and 4.0 (e39718bc8d6f092daf6abdf4a91c68c64e159c95)

Issue #6442, opened 2017-02-23
Http2ConnectionRoundtripTest.flowControlProperlyChunksLargeMessage failure


normanmaurer:
Happened during CI build:
```
java.lang.AssertionError
    at io.netty.handler.codec.http2.Http2ConnectionRoundtripTest.flowControlProperlyChunksLargeMessage(Http2ConnectionRoundtripTest.java:615)
------- Stdout: -------
11:37:46.289 [defaultEventLoopGroup-48-2] DEBUG io.netty.buffer.PoolThreadCache - Freed 5 thread-local buffer(s) from thread: defaultEventLoopGroup-48-2
11:37:46.291 [defaultEventLoopGroup-49-1] DEBUG io.netty.buffer.PoolThreadCache - Freed 4 thread-local buffer(s) from thread: defaultEventLoopGroup-49-1
```

https://garage.netty.io/teamcity/viewLog.html?buildId=17335&buildTypeId=pulls_netty&guest=1

Scottmitch:
This test may have been unusually slow on this particular build. Here are times from this build:

https://garage.netty.io/teamcity/viewLog.html?buildId=17335&buildTypeId=pulls_netty&tab=testsInfo
```
Status	Test	Duration	Order#
OK	Http2ConnectionRoundtripTest. stressTest  (io.netty.handler.codec.http2)	27s,344ms  View trend	10207
 Failure	Http2ConnectionRoundtripTest. flowControlProperlyChunksLargeMessage  (io.netty.handler.codec.http2)	5s,546ms  View trend	10211
```

Here are results from other builds:

https://garage.netty.io/teamcity/viewLog.html?buildId=16985&buildTypeId=pulls_netty&tab=testsInfo
```
Status	Test	Duration	Order#
OK	Http2ConnectionRoundtripTest. stressTest  (io.netty.handler.codec.http2)	12s,120ms  View trend	10219
OK	Http2ConnectionRoundtripTest. nonHttp2ExceptionInPipelineShouldNotCloseConnection  (io.netty.handler.codec.http2)	2s,918ms  View trend	10216
```

https://garage.netty.io/teamcity/viewLog.html?buildId=17138&buildTypeId=pulls_netty&tab=testsInfo
```

Status	Test	Duration	Order#
OK	Http2ConnectionRoundtripTest. stressTest  (io.netty.handler.codec.http2)	12s,506ms  View trend	10291
[...]

PR #6448, opened 2017-02-23
Increase timeouts in Http2ConnectionRoundtripTest

normanmaurer:
Motivation:

The timeouts used in the Http2ConnectionRoundtripTest seems to be too low when leak-detection is enabled so we sometimes get failed tests due timeout.

Modifications:

Increase timeouts.

Result:

Fixes [#6442].


normanmaurer:
Cherry-picked into 4.1 (7aff6b0)

PR #6449, opened 2017-02-24
Fix test-failures introduced 325cc84a2ebbf4f5e362ec9ca09202b4647bd100…

normanmaurer:
… on Java7

Motivation:

325cc84a2ebbf4f5e362ec9ca09202b4647bd100 introduced new tests which uses classes only provided by Java8+. We need to ensure we only try to load classes needed for these when we run the tests on Java8+ so we still can run the testsuite with Java7.

Modifications:

Add extra class which only gets loaded when Java8+ is used and move code there.

Result:

No more class-loader issue when running tests with Java7.

### DIFF
--- a/handler/src/test/java/io/netty/handler/ssl/Java8SslUtils.java
+++ b/handler/src/test/java/io/netty/handler/ssl/Java8SslUtils.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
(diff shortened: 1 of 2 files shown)


normanmaurer:
Cherry-picked into 4.1 (22ccf6c7b1d4e4bbd674be3bc79d5e0e545e72ba) and 4.0 (bb37e59)

PR #6451, opened 2017-02-24
Simplify Http2FrameCodec by directly extending HttpConnectionHandler

normanmaurer:
Motiviation:

Http2FrameCodec should just extend HttpConnectionHandler and so remove some uncessary levels of abstraction. This way we can make it faster and less hacky.

Modifications:

- Let HttpFrameCodec extend HttpConnectionHandler
- Adjust tests

Result:

Less indirection and cleaner implementation.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Codec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Codec.java
@@ -75,13 +75,13 @@ public final class Http2Codec extends ChannelDuplexHandler {
      * @param initialSettings non default initial settings to send to peer
      */
     public Http2Codec(boolean server, Http2StreamChannelBootstrap bootstrap, Http2FrameLogger frameLogger,
-                      Http2Settings initialSettings) {
+                Http2Settings initialSettings) {
         this(server, bootstrap, new DefaultHttp2FrameWriter(), frameLogger, initialSettings);
     }
 
     // Visible for testing
     Http2Codec(boolean server, Http2StreamChannelBootstrap bootstrap, Http2FrameWriter frameWriter,
-               Http2FrameLogger frameLogger, Http2Settings initialSettings) {
+                Http2FrameLogger frameLogger, Http2Settings initialSettings) {
         frameCodec = new Http2FrameCodec(server, frameWriter, frameLogger, initialSettings);
         multiplexCodec = new Http2MultiplexCodec(server, bootstrap);
     }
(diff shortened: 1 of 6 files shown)


normanmaurer:
@ejona86 PTAL as well. I think @Scottmitch mentioned to me that he suggested this at some point you  did not like it. Also /cc @buchgr

buchgr:
I would argue there is no good argument for having an inheritance relationship (what does 'less hacky' really mean?), but there are a few good ones against it.

1. There is simply no inheritance relationship. Http2FrameCodec is NOT a Http2ConnectionHandler. Just as Cat is not a Fruit.
2. We have benchmarked it in the past and we couldn't measure any performance improvements by introducing this inheritance relationship.
3. Last time I checked, the "vision" was for the frame codec to be Netty's public HTTP/2 API and to get rid of the old API. By having Http2FrameCodec inherit from Http2ConnectionHandler it also inherits it's public methods, which we definitely don't want on the Http2FrameCodec.
4. Right now, the Http2FrameCodec cleanly hides away the Http2ConnectionHandler and reuses it's functionality. Once, the frame codec API is complete (poke poke https://github.com/netty/netty/pull/5914 :P) we can make the old API package private and start ripping it apart where it makes sense to better fit the needs of the frame codec and to then *actually* improve performance.

- Jakob

normanmaurer:
@buchgr thanks for your comments and sorry if hacky sounded harsh this was not what it should have been :) Basically I just found the code I pushed here in a branch and wanted to bring it up again. If it makes no sense I am more then happy to just delete it.  That said I think how Http2FrameCodec uses internally the other handlers is a bit ugly but maybe it good enough for now and we will take care of it with #5914

buchgr:
> sorry if hacky sounded harsh

It didn't - no worries :-).

> I think how Http2FrameCodec uses internally the other handlers is a bit ugly

Could you elaborate your thinking on this? It just seems to me that inheritance couples the new and old API very tightly, while the current way of composing does less so [1]

[1] https://en.wikipedia.org/wiki/Composition_over_inheritance

normanmaurer:
Let me close this for now

ejona86:
I was against this before because of "the Http2FrameCodec cleanly hides away the Http2ConnectionHandler and reuses it's functionality." I wanted the strong separation of "new" and "old" APIs to make it very obvious which is which. It's complicated enough as-is; in my mind it'd be harder to figure out what is going on and how things should behave using extending. Yes, the hack has its own complexity, so it is a trade-off.

I do agree that using the Http2ConnectionHandler like we are is hackish, but I blame that on Http2ConnectionHandler, in how you have to extend it. I do think there's things we can do to improve the situation, but they are more invasive in Http2ConnectionHandler and other classes.

PR #6453, opened 2017-02-24
Enable HTTPS algorithm when directly creating a SslHandler from a Ssl…

slandelle:
…Context

Motivation:

`SslContext` has some helper methods to directly create SslHandler,
without having to first create the `SSLEngine`.

They are currently unsecure when used client side, as they don’t set
the « HTTP » endpointIdentificationAlgorithm, hence hostname
verification is not enabled.

IMHO, Netty should enforce security by default.

If one doesn’t want the HTTPS behavior, eg disabling hostname
verification, or a server that doesn’t support presenting hostname in
SslHello, he has to use the other methods: create the SSLEngine and
pass it to the handler.

Modification:

Always enable HTTPS algorithm in `newHandler` methods.

Result:

`newHandler` now returns `SslHandler` that are secured on the client
side when running JDK7+.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslContext.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslContext.java
@@ -25,18 +25,20 @@ import io.netty.handler.ssl.ApplicationProtocolConfig.Protocol;
 import io.netty.handler.ssl.ApplicationProtocolConfig.SelectedListenerFailureBehavior;
 import io.netty.handler.ssl.ApplicationProtocolConfig.SelectorFailureBehavior;
 import io.netty.util.internal.EmptyArrays;
+import io.netty.util.internal.PlatformDependent;
 
-import javax.net.ssl.KeyManager;
-import javax.net.ssl.KeyManagerFactory;
 import javax.crypto.Cipher;
 import javax.crypto.EncryptedPrivateKeyInfo;
 import javax.crypto.NoSuchPaddingException;
 import javax.crypto.SecretKey;
 import javax.crypto.SecretKeyFactory;
 import javax.crypto.spec.PBEKeySpec;
+import javax.net.ssl.KeyManager;
+import javax.net.ssl.KeyManagerFactory;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLException;
+import javax.net.ssl.SSLParameters;
 import javax.net.ssl.SSLSessionContext;
 import javax.net.ssl.TrustManager;
 import javax.net.ssl.TrustManagerFactory;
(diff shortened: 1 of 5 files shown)


slandelle:
ping @normanmaurer

normanmaurer:
@slandelle please fix Unit tests

slandelle:
@normanmaurer done

Scottmitch:
Although perhaps a good policy to have this is a breaking change ... this may cause handshakes to fail if connection address and certificate hostname names SANs/CN don't line up (for w/e reason). I'm not against making the change but just want to be sure everyone's on board. The user can enable hostname verification as it stands today (I understand folks may forget and this may expose vulnerabilities).

slandelle:
@Scottmitch I do have mixed feelings regarding this change. On one side, I think that Netty should enforce security by default. On the other end, adapting Netty's test suite for this PR made me realize that there could be tons of code out in the wild that might expect the current unsafe behavior :(

Then, if fixing current behavior has to wait for a next major release, would adding a warning in those methods' javadoc suffice? I bet it would go pretty unnoticed.

normanmaurer:
@Scottmitch @slandelle I think I would go with the javadocs warning for now. This will reduce the risk of breakage. That said another thing we could do is enable it by default when a system property is set . Not sure this will buy us much tho.

handler/src/main/java/io/netty/handler/ssl/SslContext.java
```diff
@@ -857,27 +859,41 @@ public final boolean isServer() {
...
      * @return a new {@link SslHandler}
      */
     public final SslHandler newHandler(ByteBufAllocator alloc, String peerHost, int peerPort) {
-        return new SslHandler(newEngine(alloc, peerHost, peerPort), startTls);
+        SSLEngine sslEngine = newEngine(alloc, peerHost, peerPort);
+        setHttpsEndpointIdentificationAlgorithm(sslEngine);
+        return new SslHandler(sslEngine, startTls);
+    }
+
+    private void setHttpsEndpointIdentificationAlgorithm(SSLEngine sslEngine) {
```
netkins: MINOR Make "setHttpsEndpointIdentificationAlgorithm" a "static" method. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2325)

normanmaurer:
@slandelle ping...

slandelle:
@normanmaurer I can provide a new PR with javadoc update. Then, would we keep this PR around for a next major version?

normanmaurer:
@slandelle yep we can keep it around for the next major. @Scottmitch already noted it as TODO for the next major one.

slandelle:
@normanmaurer Will provide a PR later today or tomorrow.

normanmaurer:
@slandelle thx

Scottmitch:
Let me close this for now ... lets do this for the next major release.

PR #6458, opened 2017-02-27
Optimize / Cleanup QueryStringDecoder / QueryStringEncoder

normanmaurer:
Motivation:

QueryStringDecoder and QueryStringEncoder contained some code that could either cleaned-up or optimized.

Modifications:

- Fix typos in exception messages and javadocs
- Precompile Pattern
- Make use of StringUtil.EMPTY_STRING

Result:

Faster and cleaner code.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
@@ -26,6 +26,10 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
+import static io.netty.util.internal.ObjectUtil.checkNotNull;
+import static io.netty.util.internal.ObjectUtil.checkPositive;
+import static io.netty.util.internal.StringUtil.EMPTY_STRING;
+
 /**
  * Splits an HTTP query string into a path string and key-value parameter pairs.
  * This decoder is for one time use only.  Create a new instance for each URI:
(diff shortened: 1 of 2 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
```diff
@@ -140,26 +137,19 @@ public QueryStringDecoder(URI uri, Charset charset) {
...
         if (rawPath != null) {
             hasPath = true;
         } else {
-            rawPath = "";
+            rawPath = EMPTY_STRING;
             hasPath = false;
         }
         // Also take care of cut of things like "http://localhost"
-        this.uri = rawPath + (uri.getRawQuery() == null? "" : '?' + uri.getRawQuery());
+        this.uri = rawPath + (uri.getRawQuery() == null? EMPTY_STRING : '?' + uri.getRawQuery());
```
slandelle: Isn't there a useless intermediate concat when hitting the second branch?

Would this be better?
`uri.getRawQuery() == null? rawPath : rawPath + "?" + uri.getRawQuery()`

normanmaurer: +1

CodingFabian:
lgtm

codec-http/src/main/java/io/netty/handler/codec/http/QueryStringEncoder.java
```diff
@@ -37,6 +40,7 @@
  */
 public class QueryStringEncoder {
 
+    private static final Pattern PATTERN = Pattern.compile("\\+");
```
trustin: Not sure which is better, but how about:

```java
Pattern.compile("+", Pattern.LITERAL)
```

normanmaurer: yeah let us do this :)

normanmaurer:
Cherry-picked into 4.1 (d675febf07d14d4dff82471829f974369705655a) and 4.0 (67224a8da1c528b8f6f5343f73fbc24733bc9f78)

PR #6462, opened 2017-02-27
WebSocketServerProtocolHandshakeHandler improvements

doom369:
Motivation:

Make code easier to read, make `WebSocketServerProtocolHandshakeHandler.getWebSocketLocation` method faster.

Modification:

WebSocket path check moved to separate method. Get header operation moved out from concat operation.

Result:

`WebSocketServerProtocolHandshakeHandler.getWebSocketLocation` is faster as `OptimizeStringConcat` could be applied. Code easier to read.

```
Benchmark                                          Mode  Cnt         Score         Error  Units
ConcatBenchmark.getWebSocketLocationHostMovedOut  thrpt   40  30327006.732 ± 1016440.196  ops/s
ConcatBenchmark.getWebSocketLocationOld           thrpt   40  13422614.312 ±  292622.139  ops/s
```

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketServerProtocolHandshakeHandler.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketServerProtocolHandshakeHandler.java
@@ -62,12 +62,7 @@ class WebSocketServerProtocolHandshakeHandler extends ChannelInboundHandlerAdapt
     @Override
     public void channelRead(final ChannelHandlerContext ctx, Object msg) throws Exception {
         final FullHttpRequest req = (FullHttpRequest) msg;
-        if (checkStartsWith) {
-            if (!req.uri().startsWith(websocketPath)) {
-                ctx.fireChannelRead(msg);
-                return;
-            }
-        } else if (!req.uri().equals(websocketPath)) {
+        if (isNotWebSocketPath(req)) {
             ctx.fireChannelRead(msg);
             return;
         }
@@ -110,6 +105,10 @@ class WebSocketServerProtocolHandshakeHandler extends ChannelInboundHandlerAdapt
         }
     }
 
+    private boolean isNotWebSocketPath(FullHttpRequest req) {
+        return checkStartsWith ? !req.uri().startsWith(websocketPath) : !req.uri().equals(websocketPath);
+    }
+
     private static void sendHttpResponse(ChannelHandlerContext ctx, HttpRequest req, HttpResponse res) {
         ChannelFuture f = ctx.channel().writeAndFlush(res);
         if (!isKeepAlive(req) || res.status().code() != 200) {
@@ -123,6 +122,7 @@ class WebSocketServerProtocolHandshakeHandler extends ChannelInboundHandlerAdapt
             // SSL in use so use Secure WebSockets
             protocol = "wss";
         }
-        return protocol + "://" + req.headers().get(HttpHeaderNames.HOST) + path;
+        String host = req.headers().get(HttpHeaderNames.HOST);
+        return protocol + "://" + host + path;
     }
 }


codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketServerProtocolHandshakeHandler.java
```diff
@@ -110,6 +105,14 @@ public void operationComplete(ChannelFuture future) throws Exception {
         }
     }
 
+    private boolean isNotWebSocketPath(FullHttpRequest req) {
+        if (checkStartsWith) {
+            return !req.uri().startsWith(websocketPath);
+        }
+
+        return !req.uri().equals(websocketPath);
```
normanmaurer: @doom369 what about make this:

```java
return checkStartsWith ? !req.uri().startsWith(websocketPath) : !req.uri().equals(websocketPath);
```

doom369: For it is the same, I don't mind. Let me fix.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketServerProtocolHandshakeHandler.java
```diff
@@ -110,6 +105,10 @@ public void operationComplete(ChannelFuture future) throws Exception {
         }
     }
 
+    private boolean isNotWebSocketPath(FullHttpRequest req) {
+        return checkStartsWith ? !req.uri().startsWith(websocketPath) : !req.uri().equals(websocketPath);
```
Scottmitch: I wonder if removing a branch here will change anything:

```java
        return checkStartsWith && !req.uri().startsWith(websocketPath) ||
               !checkStartsWith && !req.uri().equals(websocketPath);
```

doom369: @Scottmitch sorry, didn't get your point.

Scottmitch: Your code currently has a branch statement: `if <condition> then <statement> else <statement>` however the suggestion above removes a branch at the expense of evaluating the same boolean twice. Although the difference is likely in the noise there is a possibility these two will have different characteristics.

doom369: Do you think it worth it? This is just handshake that happens once on establishing connection. The goal here was to make code cleaner and easier to read, while proposed change is opposite.

normanmaurer: I agree with @doom369 ... Let us just merge it as it is.

normanmaurer:
Cherry-picked into 4.1 (d1b02257248adc250cc3ac61cb5efd179675519b)

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketServerProtocolHandshakeHandler.java
Scottmitch: I don't think this is much different from a readability standpoint. The reason why I included it was because there was some performance benefit claimed by this PR. I don't think it is necessary ... just a curiosity if part of the justification for this PR is performance.

doom369: @Scottmitch performance improvement was for `WebSocketServerProtocolHandshakeHandler.getWebSocketLocation` method where `host` variable was moved out from concat operation and thus concat operation became available for `OptimizeStringConcat` optimization (less allocations + faster).

Scottmitch: gotcha ... thanks for clarifying. lgtm

fenik17: @doom369 cool. didn't know what concat optimization can stumble on the such thing..

> thus concat operation became available for OptimizeStringConcat optimization

doom369: @fenik17 sad, but true :). Here is benchmark you can play with : 

```
@Threads(1)
@Measurement(iterations = 20)
@Warmup(iterations = 10)
public class ConcatBenchmark extends AbstractMicrobenchmark {

    HttpRequest req;
    ChannelPipeline cp;
    Channel channel;
    String path;

    @Setup
    public void setup() {
        path = "/websockets";
        req = new DefaultHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.GET, path);
        req.headers().set(HttpHeaderNames.HOST, "benchmark.test");
        channel = new EmbeddedChannel();
        channel.connect(new InetSocketAddress(0));
        cp = channel.pipeline();
    }

    @Benchmark
    public String getWebSocketLocationOld() {
        return getWebSocketLocationOld(cp, req, path);
    }

    @Benchmark
    public String getWebSocketLocationHostMovedOut() {
        return getWebSocketLocationHostMovedOut(cp, req, path);
    }

    private static String getWebSocketLocationOld(ChannelPipeline cp, HttpRequest req, String path) {
        String protocol = "ws";
        if (cp.get(SslHandler.class) != null) {
            // SSL in use so use Secure WebSockets
            protocol = "wss";
        }
        return protocol + "://" + req.headers().get(HttpHeaderNames.HOST) + path;
    }

    private static String getWebSocketLocationHostMovedOut(ChannelPipeline cp, HttpRequest req, String path) {
        String protocol = "ws";
        if (cp.get(SslHandler.class) != null) {
            // SSL in use so use Secure WebSockets
            protocol = "wss";
        }
        String host = req.headers().get(HttpHeaderNames.HOST);
        return protocol + "://" + host + path;
    }

}
```

PR #6465, opened 2017-02-27
Metrics exposed by PooledByteBufAllocator needs to be correctly synch…

normanmaurer:
…ronized

Motivation:

As we may access the metrics exposed of PooledByteBufAllocator from another thread then the allocations happen we need to ensure we synchronize on the PoolArena to ensure correct visibility.

Modifications:

Synchronize on the PoolArena to ensure correct visibility.

Result:

Fix multi-thread issues on the metrics

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/PoolChunkList.java
+++ b/buffer/src/main/java/io/netty/buffer/PoolChunkList.java
@@ -27,6 +27,7 @@ import static java.lang.Math.*;
 
 final class PoolChunkList<T> implements PoolChunkListMetric {
     private static final Iterator<PoolChunkMetric> EMPTY_METRICS = Collections.<PoolChunkMetric>emptyList().iterator();
+    private final PoolArena<T> arena;
     private final PoolChunkList<T> nextList;
     private final int minUsage;
     private final int maxUsage;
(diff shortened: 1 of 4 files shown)


normanmaurer:
@Scottmitch PTAL ... noticed this today. Please note we holds the lock on the PoolArena for the methods where things are updated already.

buffer/src/main/java/io/netty/buffer/PoolSubpage.java
```diff
@@ -200,27 +200,35 @@ private long toHandle(int bitmapIdx) {
 
     @Override
     public String toString() {
-        if (!doNotDestroy) {
-            return "(" + memoryMapIdx + ": not in use)";
-        }
+        synchronized (chunk.arena) {
```
Scottmitch: consider building the string outside the synchronized block.

buffer/src/main/java/io/netty/buffer/PoolChunkList.java
```diff
@@ -190,37 +192,41 @@ private static int minUsage0(int value) {
...
 
-        StringBuilder buf = new StringBuilder();
-        for (PoolChunk<T> cur = head;;) {
-            buf.append(cur);
-            cur = cur.next;
-            if (cur == null) {
-                break;
+            StringBuilder buf = new StringBuilder();
+            for (PoolChunk<T> cur = head;;) {
+                buf.append(cur);
```
Scottmitch: consider building the string outside the synchronized block. Trade off between potentially inconsistent state in the string and reducing the critical area.

normanmaurer: let me move the creation of the StringBuilder and the toString() call outside. The rest can stay inside as the PoolChunk itself will also need to hold the synchronized lock anyway.

buffer/src/main/java/io/netty/buffer/PoolSubpage.java
normanmaurer: +1

normanmaurer:
Cherry-picked into 4.1 (a7fe6c0) and 4.0 (d54ed19)

Issue #6466, opened 2017-02-27
Exception: bio is null via SSL#bioLengthByteBuffer


rkapsi:
Netty 4.1.9.Final-SNAPSHOT with netty-tcnative 2.0.0.Beta6

First and only time I've seen it so far. Happened on a H2 connection.

```java
2017-02-27 21:50:45,339 [WorkerThread-0] AbstractChannelHandlerContext WARN : Failed to mark a promise as failure because it has succeeded already: DefaultChannelPromise@7c0ddec1(success)
java.lang.Exception: bio is null
	at io.netty.internal.tcnative.SSL.$$YJP$$bioLengthByteBuffer(Native Method) ~[netty-all-4.1.9.Final-SNAPSHOT.jar:4.1.9.Final-SNAPSHOT]
	at io.netty.internal.tcnative.SSL.bioLengthByteBuffer(SSL.java) ~[netty-all-4.1.9.Final-SNAPSHOT.jar:4.1.9.Final-SNAPSHOT]
	at io.netty.handler.ssl.ReferenceCountedOpenSslEngine.wrap(ReferenceCountedOpenSslEngine.java:483) ~[netty-all-4.1.9.Final-SNAPSHOT.jar:4.1.9.Final-SNAPSHOT]
	at javax.net.ssl.SSLEngine.wrap(SSLEngine.java:509) ~[?:1.8.0_102]
	at io.netty.handler.ssl.SslHandler.wrap(SslHandler.java:836) ~[netty-all-4.1.9.Final-SNAPSHOT.jar:4.1.9.Final-SNAPSHOT]
	at io.netty.handler.ssl.SslHandler.wrap(SslHandler.java:659) ~[netty-all-4.1.9.Final-SNAPSHOT.jar:4.1.9.Final-SNAPSHOT]
	at io.netty.handler.ssl.SslHandler.wrapAndFlush(SslHandler.java:631) ~[netty-all-4.1.9.Final-SNAPSHOT.jar:4.1.9.Final-SNAPSHOT]
	at io.netty.handler.ssl.SslHandler.flush(SslHandler.java:612) ~[netty-all-4.1.9.Final-SNAPSHOT.jar:4.1.9.Final-SNAPSHOT]
	at io.netty.handler.ssl.SslHandler.flush(SslHandler.java:1441) ~[netty-all-4.1.9.Final-SNAPSHOT.jar:4.1.9.Final-SNAPSHOT]
[...]

PR #6467, opened 2017-02-28
OpenSslEngine may use networkBIO after calling shutdown

Scottmitch:
Motivation:
To ensure that all bytes queued in OpenSSL/tcnative internal buffers we invoke SSL_shutdown again to stimulate OpenSSL to write any pending bytes. If this call fails we may call SSL_free and the associated shutdown method to free resources. At this time we may attempt to use the networkBIO which has already been freed and get a NPE.

Modifications:
- Don't call bioLengthByteBuffer(networkBIO) if we have called shutdown() in ReferenceCountedOpenSslEngine

Result:
Fixes https://github.com/netty/netty/issues/6466


Scottmitch:
@rkapsi - can you verify this fixes the issue?

I will try to come up with a unit test tomorrow.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -479,9 +478,12 @@ public final SSLEngineResult wrap(
...
                     // OpenSSL can give us.
-                    boolean didShutdown = doSSLShutdown();
-                    bytesProduced = bioLengthBefore - SSL.bioLengthByteBuffer(networkBIO);
-                    return newResultMayFinishHandshake(didShutdown ? NEED_WRAP : NOT_HANDSHAKING, 0, bytesProduced);
+                    final int shutdownResult = doSSLShutdown();
+                    if (shutdownResult < 0) {
+                        return newResultMayFinishHandshake(NOT_HANDSHAKING, 0, bytesProduced);
+                    }
+                    bytesProduced = bioLengthBefore - shutdownResult;
+                    return newResultMayFinishHandshake(NEED_WRAP, 0, bytesProduced);
```
normanmaurer: Nit: 

```java
return shutdownResult < 0 ? newResultMayFinishHandshake(NOT_HANDSHAKING, 0, bytesProduced) : 
            newResultMayFinishHandshake(NEED_WRAP, 0, bioLengthBefore - shutdownResult);
```

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -34,7 +34,6 @@
 
 import java.nio.ByteBuffer;
 import java.nio.ReadOnlyBufferException;
-import java.security.AlgorithmConstraints;
```
normanmaurer: @Scottmitch was this an unused import we did miss ?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
Scottmitch: This will not work as `bytesProduced` must be set because it is used to advance the buffer position.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
Scottmitch: Yah it looks like we don't want this and it was introduced recently [1]. We store a reference to an algorithm constraint as an `Object algorithmConstraints` to avoid referencing this JDK 1.7+ class I guess.

[1] https://github.com/netty/netty/commit/325cc84a2ebbf4f5e362ec9ca09202b4647bd100#diff-cfd417915faad637e4278b54adafd9baR37

normanmaurer: Shame on me 👎 ... Thanks!

normanmaurer:
@Scottmitch just merge if a unit test is not feasible

Scottmitch:
4.1 (70bcb40) 4.0 (417eea8)

Scottmitch:
Creating a unit test for this requires more investigation. Reducing the BIO size gets us into this spot but I'm not sure how to make `SSL_shutdown` return an error.

rkapsi:
@Scottmitch I'll try to verify before EOD. Trying to squeeze multiple things through our release pipeline.

PR #6469, opened 2017-02-28
Ensure OpenSSL was loaded before try to use OpenSslCertificateException

normanmaurer:
Motivation:

We need to ensure we correctly load the OpenSSL libary before we try to validate the provided error code in OpenSslCertificateException. This is needed as we now directly obtain the error numbers via native calls in CertificateVerifier.

Modifications:

- Expose the valid error numbers in OpenSslCertificateException
- Ensure OpenSSL library is correctly loaded before trying to access CertificateVerifier static fields.

Result:

No more linking errors possible during loading.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/OpenSslCertificateException.java
+++ b/handler/src/main/java/io/netty/handler/ssl/OpenSslCertificateException.java
@@ -16,6 +16,9 @@
 package io.netty.handler.ssl;
 
 import io.netty.internal.tcnative.CertificateVerifier;
+import io.netty.util.collection.IntCollections;
+import io.netty.util.collection.IntObjectHashMap;
+import io.netty.util.collection.IntObjectMap;
 
 import java.security.cert.CertificateException;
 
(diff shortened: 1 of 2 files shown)


handler/src/main/java/io/netty/handler/ssl/OpenSslCertificateException.java
```diff
@@ -70,10 +424,16 @@ public int errorCode() {
...
-                    + CertificateVerifier.X509_V_ERR_DANE_NO_MATCH +
-                    ". See https://www.openssl.org/docs/manmaster/apps/verify.html .");
+        if (ERROR_MAP == null) {
+            // We could not load OpenSSL before, so fail here.
+            OpenSsl.ensureAvailability();
+
+            // We should never reach here.
+            assert false;
+        }
+        if (ERROR_MAP.get(errorCode) != Boolean.TRUE) {
```
netkins: BLOCKER NullPointerException might be thrown as 'ERROR_MAP' is nullable here [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2259)

handler/src/main/java/io/netty/handler/ssl/OpenSslCertificateException.java
```diff
@@ -24,6 +27,357 @@
...
+    public static final int X509_V_ERR_CERT_REJECTED;
+    public static final int X509_V_ERR_SUBJECT_ISSUER_MISMATCH;
+    public static final int X509_V_ERR_AKID_SKID_MISMATCH;
+    public static final int X509_V_ERR_AKID_ISSUER_SERIAL_MISMATCH;
+    public static final int X509_V_ERR_KEYUSAGE_NO_CERTSIGN;
+    public static final int X509_V_ERR_UNABLE_TO_GET_CRL_ISSUER;
+    public static final int X509_V_ERR_UNHANDLED_CRITICAL_EXTENSION;
+    public static final int X509_V_ERR_KEYUSAGE_NO_CRL_SIGN;
+    public static final int X509_V_ERR_UNHANDLED_CRITICAL_CRL_EXTENSION;
+    public static final int X509_V_ERR_INVALID_NON_CA;;
```
netkins: MINOR Remove this empty statement. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AEmptyStatementUsageCheck)

handler/src/main/java/io/netty/handler/ssl/OpenSslCertificateException.java
normanmaurer: thats fine... we should never reach this and we have an assert

Scottmitch:
I'm not sure this change is necessary. Can you just put a call to `OpenSsl.ensureAvailability()` in a static block in `OpenSslCertificateException`?

normanmaurer:
@Scottmitch I could... The reason why I did not was that I was thinking maybe some references the class even if OpenSSL could not be loaded. Not sure this is a valid thing to do but was not sure. WDYT ?

Scottmitch:
> maybe some references the class even if OpenSSL could not be loaded

I don't think this is valid. The error codes come from `tcnative` and the class name is `OpenSSL`.

normanmaurer:
@Scottmitch ok fair enough... let me change this then :) Makes the code easier as well ;)

handler/src/main/java/io/netty/handler/ssl/OpenSslCertificateException.java
```diff
@@ -24,6 +27,286 @@
  * SSL Record. This only work when {@link SslProvider#OPENSSL} is used.
  */
 public final class OpenSslCertificateException extends CertificateException {
+    public static final int X509_V_OK;
```
Scottmitch: can we avoid duplicating these values? Can these just be local variables (if they are necessary at all)?

handler/src/main/java/io/netty/handler/ssl/OpenSslCertificateException.java
```diff
@@ -24,6 +27,286 @@
...
+    public static final int X509_V_ERR_SUITE_B_LOS_NOT_ALLOWED;
+    public static final int X509_V_ERR_SUITE_B_CANNOT_SIGN_P_384_WITH_P_256;
+    public static final int X509_V_ERR_HOSTNAME_MISMATCH;
+    public static final int X509_V_ERR_EMAIL_MISMATCH;
+    public static final int X509_V_ERR_IP_ADDRESS_MISMATCH;
+    public static final int X509_V_ERR_DANE_NO_MATCH;
+    private static final IntObjectMap<Boolean> ERROR_MAP;
+
+    static {
+        OpenSsl.ensureAvailability();
```
Scottmitch: if someone uses the interface like: `new OpenSslCertificateException(.., CertificateVerifier.X509_V_ERR_UNSPECIFIED)` then the `CertificateVerifier` class will already be loaded and loading from this class won't help right?

handler/src/main/java/io/netty/handler/ssl/OpenSslCertificateException.java
normanmaurer: The idea was to expose them to the user so the user will have these in one place which is not internal. We can also remove these and add them later if needed but my hope was that this will guide the users to not use the ones exposed by `CertificateVerifier`.

handler/src/main/java/io/netty/handler/ssl/OpenSslCertificateException.java
normanmaurer: correct. That said it helps when the user uses `new OpenSslCertificateException(.., 2);` as otherwise we may fail when try to verify the errorCode.

handler/src/main/java/io/netty/handler/ssl/OpenSslCertificateException.java
Scottmitch: Lets delay this change until we can do it in one PR (remove old internal APIs and recommend this new API).

handler/src/main/java/io/netty/handler/ssl/OpenSslCertificateException.java
Scottmitch: > That said it helps when the user uses new OpenSslCertificateException(.., 2);

This seems dangerous and likely to run into issues where the expected value is not supported by the underlying OpenSSL library. I don't think we should encourage this type of usage.

normanmaurer: @Scottmitch so I am confused by this statement... So how do you expect users to use this then if you not want them to use `CertificateVerifier` stuff (which I think makes sense): https://github.com/netty/netty/pull/6469#discussion_r103590972

Scottmitch: Currently they should use `CertificateVerifier ` and call `OpenSsl.ensureAvailability()` before they use it.

In the future we should remove the `static final int` values from `CertificateVerifier` and instead move them to `OpenSslCertificateException` as done in this PR.

In general using a raw int that is not derived from tcnative should be avoided as it may lead to values being used which are not supported by the underlying OpenSSL library/version and lead to unexpected behavior.

normanmaurer: Still I not get why you think we should remove the init of OpenSSL in the static block as added here.... also be aware that if you use an übt that is not supported by Tonstufe it will throw an exception when init this class

Scottmitch: > we should remove the init of OpenSSL in the static block as added here

Until we can remove the defines from `CertificateVerifier` this PR just adds duplicate definitions in tcnative and doesn't solve the issue of "ensuring tcnative is loaded before static defines from the library are used".  Users who currently use `CertificateVerifier` that fail to call `OpenSsl.ensureAvailability()` first will still get a link exception. The same issue exists if anyone references the static defines in tcnatvie's `SSL` class (or logically any other reference to the tcnative library). I think once we move tcnative into Netty we will be better equipped to tackle this problem.

normanmaurer: I am not sure I agree as I think it still makes sense to add it here as like I said otherwise OpenSslException will fail in the constructor even if you just pass an int and not reference `CertificateVerifier`by yourself

normanmaurer:
Let me close this for now

PR #6471, opened 2017-02-28
Mark transport udt as deprecated as barchart-udt is not maintained an…

normanmaurer:
…ymore

Motivation:

barchart-udt is not maintained anymore so there is not way for us to get fixes for udt. Because of this we should mark the transport as deprecated.

Modifications:

Deprecate all udt classes.

Result:

transport udt is deprecated and so the user knows it will be removed in the future.

### DIFF
--- a/transport-udt/src/main/java/io/netty/channel/udt/DefaultUdtChannelConfig.java
+++ b/transport-udt/src/main/java/io/netty/channel/udt/DefaultUdtChannelConfig.java
@@ -39,7 +39,10 @@ import static io.netty.channel.udt.UdtChannelOption.SYSTEM_SEND_BUFFER_SIZE;
 
 /**
  * The default {@link UdtChannelConfig} implementation.
+ *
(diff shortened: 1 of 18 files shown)


normanmaurer:
Cherry-picked into 4.1 (4734ef61a57b15df3e313b071956218fde98da2c) and 4.0 (ed370dcebb35ff12f7e1d4f2dfc882bdcd2e8078)

johnou:
@Andrei-Pozolotin fyi

dorkbox:
I just found this issue after working on some internal issues (which depend on netty and UDT) ... so I apologize for the very late delay. I don't update at the same frequency as Netty.

I've been maintaining UDT (both java + C) for use in my own network client/server application (built on Netty), which also happens to primarily use UDT. 

Should I also pickup the Netty UDT implementation?


https://github.com/dorkbox/barchart-udt
https://github.com/dorkbox/Network-UDT
https://github.com/dorkbox/UDT

PR #6472, opened 2017-02-28
Add javadoc warning on SslContext#newHandler client-side

slandelle:
Motivation:

SslContext#newHandler currently creates underlying SSLEngine without
enabling HTTPS endpointIdentificationAlgorithm. This behavior in
unsecured when used on the client side.

We can’t harden the behavior for now, as it would break existing
behavior, for example tests using self signed certificates.

Proper hardening will happen in a future major version when we can
break behavior.

Modifications:

Add javadoc warnings with code snippets.

Result:

Existing unsafe behavior and workaround documented.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslContext.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslContext.java
@@ -859,7 +859,21 @@ public abstract class SslContext {
      * Creates a new {@link SslHandler}.
      * <p>If {@link SslProvider#OPENSSL_REFCNT} is used then the returned {@link SslHandler} will release the engine
      * that is wrapped. If the returned {@link SslHandler} is not inserted into a pipeline then you may leak native
-     * memory!
+     * memory!</p>
+     * <p><b>Beware</b>: the underlying generated {@link SSLEngine} won't have
+     * <a href="https://wiki.openssl.org/index.php/Hostname_validation">hostname verification</a> enabled by default.
+     * If you create {@link SslHandler} for the client side and want proper security, we advice that you configure
+     * the {@link SSLEngine} (see {@link javax.net.ssl.SSLParameters#setEndpointIdentificationAlgorithm(String)}):</p>
+     * <pre>
+     * SSLEngine sslEngine = sslHandler.engine();
+     * SSLParameters sslParameters = sslEngine.getSSLParameters();
+     * // only available since Java 7
+     * sslParameters.setEndpointIdentificationAlgorithm("HTTPS");
+     * sslEngine.setSSLParameters(sslParameters);
+     * </pre>
+     *
+     * @param alloc If supported by the SSLEngine then the SSLEngine will use this to allocate ByteBuf objects.
+     *
      * @return a new {@link SslHandler}
      */
     public final SslHandler newHandler(ByteBufAllocator alloc) {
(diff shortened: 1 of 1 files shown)


slandelle:
@normanmaurer PTAL

handler/src/main/java/io/netty/handler/ssl/SslContext.java
```diff
@@ -859,7 +859,21 @@ public final boolean isServer() {
      * Creates a new {@link SslHandler}.
      * <p>If {@link SslProvider#OPENSSL_REFCNT} is used then the returned {@link SslHandler} will release the engine
      * that is wrapped. If the returned {@link SslHandler} is not inserted into a pipeline then you may leak native
-     * memory!
+     * memory!</p>
+     * <p><b>Beware</b>: underlying generated {@link SSLEngine} won't have hostname verification enabled!
+     * If you create {@link SslHandler} for the client side and want proper security, we advice that you don't use
+     * this method. Instead, create a {@link SslHandler} manually with a properly configured {@link SSLEngine}:</p>
```
normanmaurer: I think I would just say that you need to enable it on the `SSLEngine`. This is also possible when you use `newHandler(....)`. 

```java
SslHandler handler = ...;
hander.engine().setSSLParameters(...)
```

handler/src/main/java/io/netty/handler/ssl/SslContext.java
```diff
@@ -870,7 +884,20 @@ public final SslHandler newHandler(ByteBufAllocator alloc) {
...
+     * <pre>
+     * SSLEngine sslEngine = sslContext.newEngine(byteBufAllocator, peerHost, peerPort);
+     * SSLParameters sslParameters = sslEngine.getSSLParameters();
+     * // only available since Java 7
+     * sslParameters.setEndpointIdentificationAlgorithm("HTTPS");
+     * sslEngine.setSSLParameters(sslParameters);
+     * SslHandler sslHandler = new SslHandler(sslEngine);
```
normanmaurer: se above

slandelle:
@normanmaurer Done. Let me know if you want me to remove the code snippet.

handler/src/main/java/io/netty/handler/ssl/SslContext.java
```diff
@@ -859,7 +859,20 @@ public final boolean isServer() {
      * Creates a new {@link SslHandler}.
      * <p>If {@link SslProvider#OPENSSL_REFCNT} is used then the returned {@link SslHandler} will release the engine
      * that is wrapped. If the returned {@link SslHandler} is not inserted into a pipeline then you may leak native
-     * memory!
+     * memory!</p>
+     * <p><b>Beware</b>: the underlying generated {@link SSLEngine} won't have hostname verification enabled by default.
```
normanmaurer: @slandelle maybe add a link to an explanation what this even is ? (hostname verification )

normanmaurer:
@Scottmitch PTAL as well

slandelle:
@normanmaurer links added

handler/src/main/java/io/netty/handler/ssl/SslContext.java
```diff
@@ -870,7 +884,20 @@ public final SslHandler newHandler(ByteBufAllocator alloc) {
...
+     * the {@link SSLEngine} (see {@link javax.net.ssl.SSLParameters#setEndpointIdentificationAlgorithm(String)}):</p>
+     * <pre>
+     * SSLEngine sslEngine = sslHandler.engine();
+     * SSLParameters sslParameters = sslEngine.getSSLParameters();
+     * // only available since Java 7
+     * sslParameters.setEndpointIdentificationAlgorithm("HTTPS");
+     * sslEngine.setSSLParameters(sslParameters);
+     * </pre>
+     *
+     * @param alloc the ByteBufAllocator
```
Scottmitch: If supported by the `SSLEngine` then the `SSLEngine` will use this to allocate `ByteBuf` objects.

normanmaurer: +1

normanmaurer: @slandelle please make this change , after it I will pull in

slandelle:
@normanmaurer @Scottmitch done

normanmaurer:
Cherry-picked into 4.1 (79e24d1a174e4ef1ecc542bb584600fee9977f02) and 4.0 (6ab43b08ff2e296ea4fa296d7db5b3186ca8d5b8)

PR #6374, opened 2017-02-14
Correct expect header handling

jasontedor:
Motivation:

Today, the HTTP codec in Netty responds to HTTP/1.1 requests containing an "expect: 100-continue" header and a content-length that exceeds the max content length for the server with a 417 status (Expectation Failed). This is a violation of the HTTP specification. The purpose of this commit is to address this situation by modifying the HTTP codec to respond in this situation with a 413 status (Request Entity Too Large). Additionally, the HTTP codec ignores expectations in the expect header that are currently unsupported. This commit also addresses this situation by responding with a 417 status.

Handling the expect header is tricky business as the specification (RFC 2616) is more complicated than it needs to be. The specification defines the legitimate values for this header as "100-continue" and defines the notion of expectatation extensions. Further, the specification defines a 417 status (Expectation Failed) and this is where implementations go astray. The intent of the specification was for servers to respond with 417 status when they do not support the expectation in the expect header.

The key sentence from the specification follows:

> The server MUST respond with a 417 (Expectation Failed) status if any of the expectations cannot be met or, if there are other problems with the request, some other 4xx status.

That is, a server should respond with a 417 status if and only if there is an expectation that the server does not support (whether it be 100-continue, or another expectation extension), and should respond with another 4xx status code if the expectation is supported but there is something else wrong with the request.

Modifications:

This commit modifies the HTTP codec by changing the handling for the expect header in the HTTP object aggregator. In particular, the codec will now respond with 417 status if any expectation other than 100-continue is present in the expect header, the codec will respond with 413 status if the 100-continue expectation is present in the expect header and the content-length is larger than the max content length for the aggregator, and otherwise the codec will respond with 100 status.

Result:

The HTTP codec can now be used to correctly reply to clients that send a 100-continue expectation with a content-length that is too large for the server with a 413 status, and servers that use the HTTP codec will now no longer ignore expectations that are not supported (any value other than 100-continue).

PR #6474, opened 2017-03-01
introduce HttpServerExpectContinueHandler to handle "expect continue" messages

### LINKED ISSUE #5180: Should HttpServerCodec issue 100 continue responses when there's no HttpObjectAggregator in the pipeline?
A http server pipeline with both HttpServerCodec and HttpObjectAggregator installed will do the right thing for requests with 'expect: 100 continue' headers but removing the object aggregator leaves users rolling their own handling as seen in [netty's http server hello world example]( https://github.com/netty/netty/blob/4.1/example/src/main/java/io/netty/example/http/helloworld/HttpHelloWorldServerHandler.java#L48).

If HttpServerCodec is intended to be the 'batteries included' protocol implementation it seems like it could take the burden off application writers by issuing 100 CONTINUE responses when there is no HttpObjectAggregator instance in the pipeline thereby saving clients the timeout before proceeding with the payload.

This behavior exists in 4.1.0.CR7 (and has for a long time I think).

dschobel:
It is generally useful to have origin http servers respond to
"expect: continue-100" as soon as possible but applications without a
HttpObjectAggregator in their pipelines must use boiler plate to do so.

Modifications:

Introduce the HttpServerExpectContinueHandler handler to make it easier.

Result:

Less boiler plate for http application authors.

Fixes #5180

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.codec.http;
+
+import io.netty.buffer.Unpooled;
+import io.netty.channel.ChannelFutureListener;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.util.ReferenceCountUtil;
+
+import static io.netty.handler.codec.http.HttpHeaderNames.CONTENT_LENGTH;
+import static io.netty.handler.codec.http.HttpResponseStatus.CONTINUE;
+import static io.netty.handler.codec.http.HttpVersion.HTTP_1_1;
+
+/**
+ * Sends a <a href="https://www.w3.org/Protocols/rfc2616/rfc2616-sec8.html#sec8.2.3">100 CONTINUE</a>
+ * {@link HttpResponse} to {@link HttpRequest}s which contain a 'expect: 100-continue' header. It
+ * should only be used for applications which do <b>not</b> install the {@link HttpObjectAggregator}.
+ * <p>
+ * By default it accepts all expectations.
+ * <p>
+ * Since {@link HttpServerExpectContinueHandler} expects {@link HttpRequest}s it should be added after {@link
+ * HttpServerCodec} but before any other handlers that might send a {@link HttpResponse}. <blockquote>
+ * <pre>
+ *  {@link io.netty.channel.ChannelPipeline} p = ...;
+ *  ...
+ *  p.addLast("serverCodec", new {@link HttpServerCodec}());
+ *  p.addLast("respondExpectContinue", <b>new {@link HttpServerExpectContinueHandler}()</b>);
+ *  ...
+ *  p.addLast("handler", new HttpRequestHandler());
+ *  </pre>
+ * </blockquote>
+ */
+public class HttpServerExpectContinueHandler extends ChannelInboundHandlerAdapter {
+
+    private static final FullHttpResponse EXPECTATION_FAILED = new DefaultFullHttpResponse(
+            HTTP_1_1, HttpResponseStatus.EXPECTATION_FAILED, Unpooled.EMPTY_BUFFER);
+
+    private static final FullHttpResponse ACCEPT = new DefaultFullHttpResponse(
+            HTTP_1_1, CONTINUE, Unpooled.EMPTY_BUFFER);
+
+    static {
+        EXPECTATION_FAILED.headers().set(CONTENT_LENGTH, 0);
+        ACCEPT.headers().set(CONTENT_LENGTH, 0);
+    }
+
+    /**
+     * Produces a {@link HttpResponse} for {@link HttpRequest}s which define an expectation. Returns {@code null} if the
+     * request should be rejected. See {@link #rejectResponse(HttpRequest)}.
+     */
+    protected HttpResponse acceptMessage(@SuppressWarnings("unused") HttpRequest request) {
+        return ACCEPT.retainedDuplicate();
+    }
+
+    /**
+     * Returns the appropriate 4XX {@link HttpResponse} for the given {@link HttpRequest}.
+     */
+    protected HttpResponse rejectResponse(@SuppressWarnings("unused") HttpRequest request) {
+        return EXPECTATION_FAILED.retainedDuplicate();
+    }
+
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+        if (msg instanceof HttpRequest) {
+            HttpRequest req = (HttpRequest) msg;
+
+            if (HttpUtil.is100ContinueExpected(req)) {
+                HttpResponse accept = acceptMessage(req);
+
+                if (accept == null) {
+                    // the expectation failed so we refuse the request.
+                    HttpResponse rejection = rejectResponse(req);
+                    ReferenceCountUtil.release(msg);
+                    ctx.writeAndFlush(rejection).addListener(ChannelFutureListener.CLOSE_ON_FAILURE);
+                    return;
+                }
+
+                ctx.writeAndFlush(accept).addListener(ChannelFutureListener.CLOSE_ON_FAILURE);
+                req.headers().remove(HttpHeaderNames.EXPECT);
+            }
+        }
+        super.channelRead(ctx, msg);
+    }
+}
--- a/codec-http/src/test/java/io/netty/handler/codec/http/HttpServerExpectContinueHandlerTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/HttpServerExpectContinueHandlerTest.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.codec.http;
+
+import io.netty.channel.embedded.EmbeddedChannel;
+import io.netty.util.ReferenceCountUtil;
+import org.junit.Test;
+
+import static org.hamcrest.CoreMatchers.*;
+import static org.junit.Assert.*;
+
+public class HttpServerExpectContinueHandlerTest {
+
+    @Test
+    public void shouldRespondToExpectedHeader() {
+        EmbeddedChannel channel = new EmbeddedChannel(new HttpServerExpectContinueHandler() {
+            @Override
+            protected HttpResponse acceptMessage(HttpRequest request) {
+                HttpResponse response = new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.CONTINUE);
+                response.headers().set("foo", "bar");
+                return response;
+            }
+        });
+        HttpRequest request = new DefaultFullHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.GET, "/");
+        HttpUtil.set100ContinueExpected(request, true);
+
+        channel.writeInbound(request);
+        HttpResponse response = channel.readOutbound();
+
+        assertThat(response.status(), is(HttpResponseStatus.CONTINUE));
+        assertThat(response.headers().get("foo"), is("bar"));
+        ReferenceCountUtil.release(response);
+
+        HttpRequest processedRequest = channel.readInbound();
+        assertFalse(processedRequest.headers().contains(HttpHeaderNames.EXPECT));
+        ReferenceCountUtil.release(processedRequest);
+        assertFalse(channel.finishAndReleaseAll());
+    }
+
+    @Test
+    public void shouldAllowCustomResponses() {
+        EmbeddedChannel channel = new EmbeddedChannel(
+            new HttpServerExpectContinueHandler() {
+                @Override
+                protected HttpResponse acceptMessage(HttpRequest request) {
+                    return null;
+                }
+
+                @Override
+                protected HttpResponse rejectResponse(HttpRequest request) {
+                    return new DefaultFullHttpResponse(HttpVersion.HTTP_1_1,
+                            HttpResponseStatus.REQUEST_ENTITY_TOO_LARGE);
+                }
+            }
+        );
+
+        HttpRequest request = new DefaultFullHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.GET, "/");
+        HttpUtil.set100ContinueExpected(request, true);
+
+        channel.writeInbound(request);
+        HttpResponse response = channel.readOutbound();
+
+        assertThat(response.status(), is(HttpResponseStatus.REQUEST_ENTITY_TOO_LARGE));
+        ReferenceCountUtil.release(response);
+
+        // request was swallowed
+        assertTrue(channel.inboundMessages().isEmpty());
+        assertFalse(channel.finishAndReleaseAll());
+    }
+}
--- a/example/src/main/java/io/netty/example/http/helloworld/HttpHelloWorldServerHandler.java
+++ b/example/src/main/java/io/netty/example/http/helloworld/HttpHelloWorldServerHandler.java
@@ -45,9 +45,6 @@ public class HttpHelloWorldServerHandler extends ChannelInboundHandlerAdapter {
         if (msg instanceof HttpRequest) {
             HttpRequest req = (HttpRequest) msg;
 
-            if (HttpUtil.is100ContinueExpected(req)) {
-                ctx.write(new DefaultFullHttpResponse(HTTP_1_1, CONTINUE));
-            }
             boolean keepAlive = HttpUtil.isKeepAlive(req);
             FullHttpResponse response = new DefaultFullHttpResponse(HTTP_1_1, OK, Unpooled.wrappedBuffer(CONTENT));
             response.headers().set(CONTENT_TYPE, "text/plain");
--- a/example/src/main/java/io/netty/example/http/helloworld/HttpHelloWorldServerInitializer.java
+++ b/example/src/main/java/io/netty/example/http/helloworld/HttpHelloWorldServerInitializer.java
@@ -19,6 +19,7 @@ import io.netty.channel.ChannelInitializer;
 import io.netty.channel.ChannelPipeline;
 import io.netty.channel.socket.SocketChannel;
 import io.netty.handler.codec.http.HttpServerCodec;
+import io.netty.handler.codec.http.HttpServerExpectContinueHandler;
 import io.netty.handler.ssl.SslContext;
 
 public class HttpHelloWorldServerInitializer extends ChannelInitializer<SocketChannel> {
@@ -36,6 +37,7 @@ public class HttpHelloWorldServerInitializer extends ChannelInitializer<SocketCh
             p.addLast(sslCtx.newHandler(ch.alloc()));
         }
         p.addLast(new HttpServerCodec());
+        p.addLast(new HttpServerExpectContinueHandler());
         p.addLast(new HttpHelloWorldServerHandler());
     }
 }


codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
```diff
@@ -0,0 +1,57 @@
...
+ * </blockquote>
+ */
+public class HttpServerExpectContinueHandler extends ChannelInboundHandlerAdapter {
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+        if (msg instanceof HttpRequest) {
+            HttpRequest req = (HttpRequest) msg;
+
+            if (HttpUtil.is100ContinueExpected(req)) {
+                ctx.writeAndFlush(new DefaultFullHttpResponse(HTTP_1_1, CONTINUE));
```
jasontedor: I do not think there should be an unconditional continue reply to an Expect: 100-continue header. The point of sending such a header in the first place is to avoid sending a large body if the server would then reject it (e.g., 413, or, as another example, if authentication would fail).

dschobel: @jasontedor applications that need payload size limits for 413s will use the `HttpObjectAggregator`, this handler is strictly for use with applications that want to see content chunks.

jasontedor: I do not think that should matter. Whether or not the handler is intended for applications that want to see content chunks, a handler designed for the Expect: 100-continue header should still handle the header properly.

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
```diff
@@ -0,0 +1,57 @@
...
+                req.headers().remove(HttpHeaderNames.EXPECT);
+            }
+            super.channelRead(ctx, msg);
+        } else {
+            super.channelRead(ctx, msg);
```
fenik17: `super.channelRead(ctx, msg);` can be once beyond the if-else statement

codec-http/src/test/java/io/netty/handler/codec/http/HttpServerExpectContinueHandlerTest.java
```diff
@@ -0,0 +1,40 @@
...
+    HttpUtil.set100ContinueExpected(request, true);
+
+    channel.writeInbound(request);
+    HttpResponse response = channel.readOutbound();
+
+    assertThat(response.status(), is(HttpResponseStatus.CONTINUE));
+
+    HttpRequest processedRequest = channel.readInbound();
+    assertFalse(processedRequest.headers().contains(HttpHeaderNames.EXPECT));
+  }
```
fenik17: wrong whitespaces offset (2 instead of 4)

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
Scottmitch: Agreed we should at least give the user some control over what is done in this situation. @jasontedor - do you think an overridable method is good enough?

dschobel: @Scottmitch I'm going to post a version which does what you propose and expose a protected `acceptMessage` method which defaults to `true` since that's the desired behavior at every spot in netty (ditto finagle) that uses `HttpUtil.is100ContinueExpected` outside `HttpObjectAggregator`.

normanmaurer: @dschobel <3

dschobel:
new version that allows overriding the acceptance criteria. thanks for the feedback everybody.

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
```diff
@@ -0,0 +1,83 @@
...
+
+            if (HttpUtil.is100ContinueExpected(req)) {
+                if (acceptMessage(req)) {
+                    ctx.writeAndFlush(new DefaultFullHttpResponse(HTTP_1_1, CONTINUE));
+                    req.headers().remove(HttpHeaderNames.EXPECT);
+                    super.channelRead(ctx, msg);
+                } else {
+                    // the expection failed so we refuse the request with a final 417 response.
+                    ReferenceCountUtil.release(msg);
+                    ctx.writeAndFlush(EXPECTATION_FAILED.retainedDuplicate());
```
Scottmitch: I think we should let the user override this error code as the "expectation failed" response is only for expect requests which are not supported by the server, see PR https://github.com/netty/netty/pull/6374 for more details.

jasontedor: Indeed. I think the default here should just be the top-level 400 status code, bad request.

normanmaurer: @dschobel let us know once you have addressed this

dschobel: @Scottmitch @jasontedor @normanmaurer overridable makes sense but 417 seems like it would be more a more informative default since it indicates the expectation as the error source, what's the reason to prefer 400 which describes "malformed syntax"?

The 417 as default behavior jives with the conclusion of the linked PR:

> That is, a server should respond with a 417 status if and only if there is an expectation that the server does not support (whether it be 100-continue, or another expectation extension), and should respond with another 4xx status code if the expectation is supported but there is something else wrong with the request.

Since if users override `acceptMessage` without the error code then by definition the server does not understand why it is rejecting a message other than it is expectation related.

Thoughts?

Scottmitch: Generally sounds good to me. To clarify I think the default behavior should be:
- Handle 100-continue
- fallback to 417

The user should be able to override this behavior to do the following:
- Determine if a CONTINUE response should be sent, or some other response.
- add support for new/custom expectations.

dschobel:
Posted new version w/ overridable responses. 

I'm not thrilled about splitting the rejection logic between two methods but this is the only way I could see of including the 417 default.

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+        if (msg instanceof HttpRequest) {
+            HttpRequest req = (HttpRequest) msg;
+
+            if (HttpUtil.is100ContinueExpected(req)) {
+                if (acceptMessage(req)) {
+                    ctx.writeAndFlush(new DefaultFullHttpResponse(HTTP_1_1, CONTINUE));
```
Scottmitch: It seems likely the the user will want to customize this response (add headers, etc...). The response may depend upon the request too (tracing, etc..), right?

dschobel:
now with overridable accept messages.

dschobel:
ping @Scottmitch @normanmaurer

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+        if (msg instanceof HttpRequest) {
+            HttpRequest req = (HttpRequest) msg;
+
+            if (HttpUtil.is100ContinueExpected(req)) {
+                HttpResponse accept = acceptMessage(req);
+                if (accept != null) {
+                    ctx.writeAndFlush(accept);
```
normanmaurer: should we add a `ChannelFutureListener.CLOSE_ON_FAILURE` ?

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
```diff
@@ -0,0 +1,92 @@
...
+                    req.headers().remove(HttpHeaderNames.EXPECT);
+                    super.channelRead(ctx, msg);
+                } else {
+                    // the expection failed so we refuse the request.
+                    HttpResponse rejection = rejectResponse(req);
+                    ReferenceCountUtil.release(msg);
+                    ctx.writeAndFlush(rejection);
```
normanmaurer: should we add a `ChannelFutureListener.CLOSE_ON_FAILURE` ?

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
dschobel: depends, the client is meant to giveup waiting for a 100 continue and eventually send its request regardless so I think the session can continue successfully without this message being delivered and so the server should not hang up.

if you mean that a write failure indicates that there are transport level problems and so we should give up on the session, then yes, we should close on failure.

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
dschobel: I think this one is definitely yes.

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
normanmaurer: yes `CLOSE_ON_FAILURE`

dschobel:
now w/ ChannelFutureListener.CLOSE_ON_FAILURE (thanks norman!)

normanmaurer:
@dschobel please fix the checkstyle error: https://garage.netty.io/teamcity/viewLog.html?buildId=18586&buildTypeId=pulls_netty&guest=1

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
```diff
@@ -0,0 +1,93 @@
...
+     * request should be rejected. See {@link #rejectResponse(HttpRequest)}.
+     */
+    protected HttpResponse acceptMessage(@SuppressWarnings("unused") HttpRequest request) {
+        return new DefaultFullHttpResponse(HTTP_1_1, CONTINUE);
+    }
+
+    /**
+     * Returns the appropriate 4XX {@link HttpResponse} for the given {@link HttpRequest}.
+     */
+    protected HttpResponse rejectResponse(@SuppressWarnings("unused") HttpRequest request) {
```
netkins: MAJOR Remove this unused method parameter "request". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
```diff
@@ -0,0 +1,93 @@
...
+
+    static {
+        EXPECTATION_FAILED.headers().set(CONTENT_LENGTH, 0);
+    }
+
+    /**
+     * Produces a {@link HttpResponse} for {@link HttpRequest}s which define an expectation. Returns {@code null} if the
+     * request should be rejected. See {@link #rejectResponse(HttpRequest)}.
+     */
+    protected HttpResponse acceptMessage(@SuppressWarnings("unused") HttpRequest request) {
```
netkins: MAJOR Remove this unused method parameter "request". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+        if (msg instanceof HttpRequest) {
+            HttpRequest req = (HttpRequest) msg;
+
+            if (HttpUtil.is100ContinueExpected(req)) {
+                HttpResponse accept = acceptMessage(req);
+                if (accept != null) {
```
netkins: CRITICAL Redundant nullcheck of accept, which is known to be non-null in io.netty.handler.codec.http.HttpServerExpectContinueHandler.channelRead(ChannelHandlerContext, Object) [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=findbugs%3ARCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE)

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
```diff
@@ -0,0 +1,93 @@
...
+                    ctx.writeAndFlush(accept).addListener(ChannelFutureListener.CLOSE_ON_FAILURE);
+                    req.headers().remove(HttpHeaderNames.EXPECT);
+                    super.channelRead(ctx, msg);
+                } else {
+                    // the expection failed so we refuse the request.
```
Scottmitch: nit: `expection` -> `expectation`

codec-http/src/test/java/io/netty/handler/codec/http/HttpServerExpectContinueHandlerTest.java
```diff
@@ -0,0 +1,77 @@
...
+import io.netty.channel.embedded.EmbeddedChannel;
+import org.junit.Test;
+
+import static org.hamcrest.CoreMatchers.*;
+import static org.junit.Assert.*;
+
+public class HttpServerExpectContinueHandlerTest {
+
+    @Test
+    public void test_RespondToExpectHeader() {
```
Scottmitch: nit: make the method name more Javaish :) `test_RespondToExpectHeader` -> `testRespondToExpectHeader`

codec-http/src/test/java/io/netty/handler/codec/http/HttpServerExpectContinueHandlerTest.java
```diff
@@ -0,0 +1,77 @@
...
+        });
+        HttpRequest request = new DefaultFullHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.GET, "/");
+        HttpUtil.set100ContinueExpected(request, true);
+
+        channel.writeInbound(request);
+        HttpResponse response = channel.readOutbound();
+
+        assertThat(response.status(), is(HttpResponseStatus.CONTINUE));
+        assertThat(response.headers().get("foo"), is("bar"));
+
```
Scottmitch: add a call to `ReferenceCountUtil.release(response);` after you are done using it.

codec-http/src/test/java/io/netty/handler/codec/http/HttpServerExpectContinueHandlerTest.java
```diff
@@ -0,0 +1,77 @@
...
+        HttpRequest processedRequest = channel.readInbound();
+        assertFalse(processedRequest.headers().contains(HttpHeaderNames.EXPECT));
+    }
```
Scottmitch: add a call to `ReferenceCountUtil.release(processedRequest);` after you are done using it.

Scottmitch: also add a call to `assertFalse(channel.finishAndReleaseAll());` at the end of the test

codec-http/src/test/java/io/netty/handler/codec/http/HttpServerExpectContinueHandlerTest.java
```diff
@@ -0,0 +1,77 @@
...
+        );
+
+        HttpRequest request = new DefaultFullHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.GET, "/");
+        HttpUtil.set100ContinueExpected(request, true);
+
+        channel.writeInbound(request);
+        HttpResponse response = channel.readOutbound();
+
+        assertThat(response.status(), is(HttpResponseStatus.REQUEST_ENTITY_TOO_LARGE));
+
```
Scottmitch: add a call to `ReferenceCountUtil.release(response);` after you are done using it.

codec-http/src/test/java/io/netty/handler/codec/http/HttpServerExpectContinueHandlerTest.java
```diff
@@ -0,0 +1,77 @@
...
+        // request was swallowed
+        assertTrue(channel.inboundMessages().isEmpty());
+    }
```
Scottmitch: also add a call to `assertFalse(channel.finishAndReleaseAll());` at the end of the test

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
```diff
@@ -0,0 +1,93 @@
...
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.util.ReferenceCountUtil;
+
+import static io.netty.handler.codec.http.HttpHeaderNames.CONTENT_LENGTH;
+import static io.netty.handler.codec.http.HttpResponseStatus.CONTINUE;
+import static io.netty.handler.codec.http.HttpVersion.HTTP_1_1;
+
+/**
+ * HttpServerExpectContinueHandler sends a <a href="https://www.w3.org/Protocols/rfc2616/rfc2616-sec8.html#sec8.2.3">
```
trustin: How about just `Sends a ...`? Otherwise it's gonna look pretty long in the Javadoc.

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
```diff
@@ -0,0 +1,93 @@
...
+    static {
+        EXPECTATION_FAILED.headers().set(CONTENT_LENGTH, 0);
+    }
+
+    /**
+     * Produces a {@link HttpResponse} for {@link HttpRequest}s which define an expectation. Returns {@code null} if the
+     * request should be rejected. See {@link #rejectResponse(HttpRequest)}.
+     */
+    protected HttpResponse acceptMessage(@SuppressWarnings("unused") HttpRequest request) {
+        return new DefaultFullHttpResponse(HTTP_1_1, CONTINUE);
```
trustin: - `, Unpooled.EMPTY_BUFFER`
- Define a static final field and duplicate?

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
trustin: False positive

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
```diff
@@ -0,0 +1,93 @@
...
+                HttpResponse accept = acceptMessage(req);
+                if (accept != null) {
+                    ctx.writeAndFlush(accept).addListener(ChannelFutureListener.CLOSE_ON_FAILURE);
+                    req.headers().remove(HttpHeaderNames.EXPECT);
+                    super.channelRead(ctx, msg);
+                } else {
+                    // the expection failed so we refuse the request.
+                    HttpResponse rejection = rejectResponse(req);
+                    ReferenceCountUtil.release(msg);
+                    ctx.writeAndFlush(rejection).addListener(ChannelFutureListener.CLOSE_ON_FAILURE);
```
trustin: How about early-returning here and move the call `super.channelRead(ctx, msg)` to the bottom of the method? i.e.

```java
if (msg instanceof HttpRequest) {
    if (HttpUtil.is100ContinueExpected(req)) {
        if (accept == null) {
            ctx.writeAndFlush(rejection)....;
            return;
        }
    }
}

super.channelRead(ctx, msg);
```

codec-http/src/test/java/io/netty/handler/codec/http/HttpServerExpectContinueHandlerTest.java
trustin: Another bikeshedder: `test` prefix is so 80. How about `shouldRespondToExpectedHeader`? :-D

fenik17: [Don’t prefix your test names with “test”](http://blog.socosomi.com/dont-prefix-your-test-names-with-test/)

normanmaurer:
@dschobel please address comments and then I will pull in. Thanks!

normanmaurer:
@dschobel ping...

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
```diff
@@ -0,0 +1,97 @@
...
+    static {
+        EXPECTATION_FAILED.headers().set(CONTENT_LENGTH, 0);
+        ACCEPT.headers().set(CONTENT_LENGTH, 0);
+    }
+
+    /**
+     * Produces a {@link HttpResponse} for {@link HttpRequest}s which define an expectation. Returns {@code null} if the
+     * request should be rejected. See {@link #rejectResponse(HttpRequest)}.
+     */
+    protected HttpResponse acceptMessage(@SuppressWarnings("unused") HttpRequest request) {
```
netkins: MAJOR Remove this unused method parameter "request". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+        if (msg instanceof HttpRequest) {
+            HttpRequest req = (HttpRequest) msg;
+
+            if (HttpUtil.is100ContinueExpected(req)) {
+                HttpResponse accept = acceptMessage(req);
+
+                if (accept == null) {
```
netkins: CRITICAL Redundant nullcheck of accept, which is known to be non-null in io.netty.handler.codec.http.HttpServerExpectContinueHandler.channelRead(ChannelHandlerContext, Object) [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=findbugs%3ARCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE)

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java
```diff
@@ -0,0 +1,97 @@
...
+     * request should be rejected. See {@link #rejectResponse(HttpRequest)}.
+     */
+    protected HttpResponse acceptMessage(@SuppressWarnings("unused") HttpRequest request) {
+        return ACCEPT.retainedDuplicate();
+    }
+
+    /**
+     * Returns the appropriate 4XX {@link HttpResponse} for the given {@link HttpRequest}.
+     */
+    protected HttpResponse rejectResponse(@SuppressWarnings("unused") HttpRequest request) {
```
netkins: MAJOR Remove this unused method parameter "request". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

dschobel:
@normanmaurer feedback has been incorporated, the netkins errors look like false positives to me so I think this is gtg.

Scottmitch:
4.1 (b1cb05954)

Scottmitch:
thanks @dschobel !

PR #6478, opened 2017-03-01
Cleanup EPOLL native exceptions

Scottmitch:
Motivation:
Exceptions generated from transport-native-epoll may include duplicate error string description or inconsistent usage of the method name in the string description.

Modifications:
- Ensure the method name from static exceptions and dynamic exceptions is of the same format
- Remove duplicate string rational from the exception messages

Result:
More consistent error messages with no duplicate error description.


Scottmitch:
4.1 (93f5f62) 4.0 (5954cc1)

PR #6480, opened 2017-03-01
Allow GOAWAY to be sent from handlers after the Http2MultiplexCodec

exell-christopher:
Motivation:
When a server is load-shedding or preparing to shut-down gracefully, it would be nice to be able to issue a GOAWAY so that existing requests can finish, but new requests are turned away. The Http2FrameCodec seems to have all the correct handling for this, but the Http2MultiplexCodec does not allow for anything other than an Http2StreamFrame to be written. As far as I can tell server developers have no way to send a GOAWAY without extending the Http2MultiplexCodec itself which doesn't seem ideal.

Modification:
Update the write() functionality of the the Http2MultiplexCodec to allow Http2GoAwayFrame and Http2StreamFrame to be written.

Added one test, but feel like there may need to be more, just not sure at which level they need to be added.

My read through of the spec indicates that this shouldn't violate any rules.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
@@ -369,27 +369,36 @@ public final class Http2MultiplexCodec extends ChannelDuplexHandler {
 
         @Override
         protected void doWrite(Object msg) {
-            if (!(msg instanceof Http2StreamFrame)) {
-                ReferenceCountUtil.release(msg);
-                throw new IllegalArgumentException("Message must be an Http2StreamFrame: " + msg);
-            }
-            Http2StreamFrame frame = (Http2StreamFrame) msg;
-            ChannelPromise promise = ctx.newPromise();
-            if (isStreamIdValid(frame.streamId())) {
-                ReferenceCountUtil.release(frame);
-                throw new IllegalArgumentException("Stream id must not be set on the frame. Was: " + frame.streamId());
-            }
-            if (!isStreamIdValid(streamId())) {
-                if (!(frame instanceof Http2HeadersFrame)) {
-                    throw new IllegalArgumentException("The first frame must be a headers frame. Was: " + frame.name());
+
+            if (msg instanceof Http2StreamFrame) {
+                Http2StreamFrame frame = (Http2StreamFrame) msg;
+                ChannelPromise promise = ctx.newPromise();
+                if (isStreamIdValid(frame.streamId())) {
+                    ReferenceCountUtil.release(frame);
+                    throw new IllegalArgumentException("Stream id must not be set on the frame. Was: "
+                        + frame.streamId());
                 }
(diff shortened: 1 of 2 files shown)


normanmaurer:
LGTM... @Scottmitch @nmittler PTAL

normanmaurer:
Also maybe @buchgr

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
```diff
@@ -343,6 +343,31 @@ public void settingChannelOptsAndAttrsOnBootstrap() {
...
+        };
+
+        Http2StreamChannelBootstrap b = new Http2StreamChannelBootstrap();
+        b.parentChannel(parentChannel).handler(childChannelInitializer);
+        Channel childChannel = b.connect().channel();
+        assertTrue(childChannel.isActive());
+
+        Http2GoAwayFrame goAwayFrame = parentChannel.readOutbound();
+        assertNotNull(goAwayFrame);
+
```
Scottmitch: we need to call `goAwayFrame.release()` when we are done with `goAwayFrame`

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -369,27 +369,35 @@ protected void doClose() throws Exception {
...
+                Http2StreamFrame frame = (Http2StreamFrame) msg;
+                ChannelPromise promise = ctx.newPromise();
+                if (isStreamIdValid(frame.streamId())) {
+                    ReferenceCountUtil.release(frame);
+                    throw new IllegalArgumentException("Stream id must not be set on the frame. Was: "
+                        + frame.streamId());
+                }
+                if (!isStreamIdValid(streamId())) {
+                    if (!(frame instanceof Http2HeadersFrame)) {
+                        throw new IllegalArgumentException("The first frame must be a headers frame. Was: "
```
Scottmitch: looks like we missed a call to `ReferenceCountUtil.release(frame);` previously ... can you add this?

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
exell-christopher: Fixing.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
exell-christopher: Will do.

exell-christopher:
Changes Committed.

Scottmitch:
4.1 (52aecab)

Thanks @exell-christopher !

ejona86:
FYI: I think this was previously possible by writing the GOAWAY to the Channel.parent().

PR #6483, opened 2017-03-02
HTTP/2 move internal HPACK classes to the http2 package

Scottmitch:
Motivation:
The internal.hpack classes are no longer exposed in our public APIs and can be made package private in the http2 package.

Modifications:
- Make the hpack classes package private in the http2 package

Result:
Less APIs exposed as public.


normanmaurer:
@Scottmitch please fix compiler errors in microbenchmark module

Scottmitch:
@normanmaurer - ptal

Scottmitch:
4.1 (f9001b9)

PR #6484, opened 2017-03-02
Correct SslContextBuilder javadocs

normanmaurer:
Motivation:

SslContextBuilder sill state the KeyManagerFactory and TrustManagerFactory are only supported when SslProvider.JDK is used. This is not correct anymore.

Modifications:

Fix javadocs.

Result:

Correct javadocs.


normanmaurer:
will merge once CI pass

normanmaurer:
Cherry-picked into 4.1 (7e7e10f) and 4.0 (944cc38)

PR #6460, opened 2017-02-27
Allow to obtain informations of used direct and heap memory for ByteBufAllocator implementations

normanmaurer:
Motivation:

Often its useful for the user to be able to get some stats about the memory allocated via an allocator.

Modifications:

Allow to obtain the used heap and direct memory for an allocator
Add test case

Result:

Fixes [#6341]

PR #6485, opened 2017-03-02
Expose ByteBufAllocator metric in a more general way

normanmaurer:
Motivation:

PR [#6460] added a way to access the used memory of an allocator. The used naming was not very good and how things were exposed are not consistent.

Modifications:

- Add a new ByteBufAllocatorMetric and ByteBufAllocatorMetricProvider interface
- Let the ByteBufAllocator implementations implement ByteBufAllocatorMetricProvider
- Move exposed stats / metric from PooledByteBufAllocator to PooledByteBufAllocatorMetric and mark old methods as `@Deprecated`.

Result:

More consistent way to expose metric / stats for ByteBufAllocator

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/ByteBufAllocatorMetricProvider.java
+++ b/buffer/src/main/java/io/netty/buffer/ByteBufAllocatorMetricProvider.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.buffer;
+
+public interface ByteBufAllocatorMetricProvider {
+
+    /**
+     * Returns a {@link ByteBufAllocatorMetric} for a {@link ByteBufAllocator}.
+     */
+    ByteBufAllocatorMetric metric();
+}
--- a/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocatorMetric.java
+++ b/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocatorMetric.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.buffer;
+
+import io.netty.util.internal.StringUtil;
+
+import java.util.List;
+
+/**
+ * Exposed metric for {@link PooledByteBufAllocator}.
+ */
+@SuppressWarnings("deprecation")
+public final class PooledByteBufAllocatorMetric implements ByteBufAllocatorMetric {
+
+    private final PooledByteBufAllocator allocator;
+
+    PooledByteBufAllocatorMetric(PooledByteBufAllocator allocator) {
+        this.allocator = allocator;
+    }
+
+    /**
+     * Return the number of heap arenas.
+     */
+    public int numHeapArenas() {
+        return allocator.numHeapArenas();
+    }
+
+    /**
+     * Return the number of direct arenas.
+     */
+    public int numDirectArenas() {
+        return allocator.numDirectArenas();
+    }
+
+    /**
+     * Return a {@link List} of all heap {@link PoolArenaMetric}s that are provided by this pool.
+     */
+    public List<PoolArenaMetric> heapArenas() {
+        return allocator.heapArenas();
+    }
+
+    /**
+     * Return a {@link List} of all direct {@link PoolArenaMetric}s that are provided by this pool.
+     */
+    public List<PoolArenaMetric> directArenas() {
+        return allocator.directArenas();
+    }
+
+    /**
+     * Return the number of thread local caches used by this {@link PooledByteBufAllocator}.
+     */
+    public int numThreadLocalCaches() {
+        return allocator.numThreadLocalCaches();
+    }
+
+    /**
+     * Return the size of the tiny cache.
+     */
+    public int tinyCacheSize() {
+        return allocator.tinyCacheSize();
+    }
+
+    /**
+     * Return the size of the small cache.
+     */
+    public int smallCacheSize() {
+        return allocator.smallCacheSize();
+    }
+
+    /**
+     * Return the size of the normal cache.
+     */
+    public int normalCacheSize() {
+        return allocator.normalCacheSize();
+    }
+
+    /**
+     * Return the chunk size for an arena.
+     */
+    public int chunkSize() {
+        return allocator.chunkSize();
+    }
+
+    @Override
+    public long usedHeapMemory() {
+        return allocator.usedHeapMemory();
+    }
+
+    @Override
+    public long usedDirectMemory() {
+        return allocator.usedDirectMemory();
+    }
+
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder(256);
+        sb.append(StringUtil.simpleClassName(this))
+                .append("(usedHeapMemory: ").append(usedHeapMemory())
+                .append("; usedDirectMemory: ").append(usedDirectMemory())
+                .append("; numHeapArenas: ").append(numHeapArenas())
+                .append("; numDirectArenas: ").append(numDirectArenas())
+                .append("; tinyCacheSize: ").append(tinyCacheSize())
+                .append("; smallCacheSize: ").append(smallCacheSize())
+                .append("; normalCacheSize: ").append(normalCacheSize())
+                .append("; numThreadLocalCaches: ").append(numThreadLocalCaches())
+                .append("; chunkSize: ").append(chunkSize()).append(')');
+        return sb.toString();
+    }
+}
--- a/buffer/src/main/java/io/netty/buffer/UnpooledByteBufAllocator.java
+++ b/buffer/src/main/java/io/netty/buffer/UnpooledByteBufAllocator.java
@@ -17,16 +17,16 @@ package io.netty.buffer;
 
 import io.netty.util.internal.LongCounter;
 import io.netty.util.internal.PlatformDependent;
+import io.netty.util.internal.StringUtil;
 
 import java.nio.ByteBuffer;
 
 /**
  * Simplistic {@link ByteBufAllocator} implementation that does not pool anything.
  */
-public final class UnpooledByteBufAllocator extends AbstractByteBufAllocator implements InstrumentedByteBufAllocator {
+public final class UnpooledByteBufAllocator extends AbstractByteBufAllocator implements ByteBufAllocatorMetricProvider {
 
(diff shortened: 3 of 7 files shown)


buffer/src/main/java/io/netty/buffer/ByteBufAllocatorMetricProvider.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.buffer;
+
+public interface ByteBufAllocatorMetricProvider {
```
Scottmitch: Have you considered the following approach:

```java
public interface ByteBufAllocatorMetricProvider<T extends ByteBufAllocatorMetric> {
  T allocatorMetric();
}

public interface PooledByteBufAllocatorMetric extends ByteBufAllocatorMetric {
  int numHeapArenas();
  int numDirectArenas();
  ...
}

publc class class PooledByteBufAllocator implements
  ByteBufAllocatorMetricProvider<PooledByteBufAllocatorMetric>,
  PooledByteBufAllocatorMetric {
  @Override
  public PooledByteBufAllocatorMetric allocatorMetric() {
    return this;
  }

  // No need to deprecate numHeapArenas() etc...

  // Or if you really don't want PooledByteBufAllocatorMetric as part of the API for
  // PooledByteBufAllocator you could always move that to an internal class
}
```

normanmaurer: yeah I have but I liked more to mark the "old methods" as `@Deprecated` and not add a type param to the interface

Scottmitch: >  I liked more to mark the "old methods" as @Deprecated

I see ... I wonder if this is necessary though as it also creates more work later to clean up these APIs. Is it a problem for the allocator to also provide its metrics ... seems natural, no?

>  not add a type param to the interface

Doesn't it make sense in this case? Presumably the Unpooled and Pooled allocator (and custom allocator X) will want to expose different types of metrics, no?

normanmaurer: @Scottmitch I think it is more natural to only have the methods on the metric itself or on the allocator. As we said we wanted to expose a metric pojo we should be consistent and make use of it and not have these also on other objects IMHO.
 
I am not strong on the type thingy and can add it if you like... Just not sure its really needed as we can just override it as I did.

Scottmitch: I'm less concerned about where the methods are implemented (on the allocator, not on the allocator) and more concerned about having a well defined interface for them. That way folks should be able to limit access to what they pass around, have a well defined interface (which may allow for easier mocking/testing), and users to respect principle of least privilege.

normanmaurer: @Scottmitch which confuses me... as if we deprecate them on the allocator and so tell the user to use the metric interface it would be more clear, or I am missing anything ?

Scottmitch: I'm not sure it is strictly necessary to *not* have the methods accessible via the `PooledByteBufAllocator` directly.  How the user consumes these methods is up to them. Having the interface just makes the contract clear and allows users to pass this more restrictive interface if that is all that is necessary to get the job done.

normanmaurer: Ok then let me just say I would prefer to just expose it via the interface and deprecate the stuff so we can remove it later. Feels cleaner to me.

normanmaurer:
@trustin PTAL as well

buffer/src/main/java/io/netty/buffer/PooledByteBufAllocatorMetric.java
```diff
@@ -0,0 +1,104 @@
...
+    @Override
+    public long usedHeapMemory() {
+        return allocator.usedHeapMemory();
+    }
+
+    @Override
+    public long usedDirectMemory() {
+        return allocator.usedDirectMemory();
+    }
+}
```
trustin: `toString` please

buffer/src/main/java/io/netty/buffer/UnpooledByteBufAllocator.java
```diff
@@ -206,15 +216,30 @@ protected void freeDirect(ByteBuffer buffer) {
...
+        @Override
+        public long usedHeapMemory() {
+            return heapCounter.value();
+        }
+
+        @Override
+        public long usedDirectMemory() {
+            return directCounter.value();
         }
     }
```
trustin: Ditto - `toString()` please.

normanmaurer:
@trustin addressed 👍

buffer/src/main/java/io/netty/buffer/PooledByteBufAllocatorMetric.java
```diff
@@ -0,0 +1,122 @@
...
+    public long usedDirectMemory() {
+        return allocator.usedDirectMemory();
+    }
+
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder(128);
+        sb.append(StringUtil.simpleClassName(this)).append('(')
+                .append("usedHeapMemory: ").append(usedHeapMemory()).append("; ")
+                .append("usedDirectMemory: ").append(usedDirectMemory()).append("; ")
```
netkins: MINOR StringBuffer (or StringBuilder).append is called 2 consecutive times with literal Strings. Use a single append with a single combined String. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AConsecutiveLiteralAppends)

buffer/src/main/java/io/netty/buffer/PooledByteBufAllocatorMetric.java
```diff
@@ -0,0 +1,122 @@
...
+                .append("numHeapArenas: ").append(numHeapArenas()).append("; ")
```
netkins: MINOR StringBuffer (or StringBuilder).append is called 2 consecutive times with literal Strings. Use a single append with a single combined String. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AConsecutiveLiteralAppends)

buffer/src/main/java/io/netty/buffer/PooledByteBufAllocatorMetric.java
```diff
@@ -0,0 +1,122 @@
...
+                .append("numDirectArenas: ").append(numDirectArenas()).append("; ")
```
netkins: MINOR StringBuffer (or StringBuilder).append is called 2 consecutive times with literal Strings. Use a single append with a single combined String. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AConsecutiveLiteralAppends)

buffer/src/main/java/io/netty/buffer/PooledByteBufAllocatorMetric.java
```diff
@@ -0,0 +1,122 @@
...
+                .append("tinyCacheSize: ").append(tinyCacheSize()).append("; ")
```
netkins: MINOR StringBuffer (or StringBuilder).append is called 2 consecutive times with literal Strings. Use a single append with a single combined String. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AConsecutiveLiteralAppends)

buffer/src/main/java/io/netty/buffer/PooledByteBufAllocatorMetric.java
```diff
@@ -0,0 +1,122 @@
...
+                .append("smallCacheSize: ").append(smallCacheSize()).append("; ")
```
netkins: MINOR StringBuffer (or StringBuilder).append is called 2 consecutive times with literal Strings. Use a single append with a single combined String. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AConsecutiveLiteralAppends)

buffer/src/main/java/io/netty/buffer/PooledByteBufAllocatorMetric.java
```diff
@@ -0,0 +1,122 @@
...
+                .append("normalCacheSize: ").append(normalCacheSize()).append("; ")
```
netkins: MINOR StringBuffer (or StringBuilder).append is called 2 consecutive times with literal Strings. Use a single append with a single combined String. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AConsecutiveLiteralAppends)

buffer/src/main/java/io/netty/buffer/PooledByteBufAllocatorMetric.java
```diff
@@ -0,0 +1,122 @@
...
+                .append("numThreadLocalCaches: ").append(numThreadLocalCaches()).append("; ")
```
netkins: MINOR StringBuffer (or StringBuilder).append is called 2 consecutive times with literal Strings. Use a single append with a single combined String. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AConsecutiveLiteralAppends)

buffer/src/main/java/io/netty/buffer/PooledByteBufAllocatorMetric.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @Override
+    public long usedDirectMemory() {
+        return allocator.usedDirectMemory();
+    }
+
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder(128);
```
netkins: MAJOR StringBuffer constructor is initialized with size 128, but has at least 165 characters appended. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AInsufficientStringBufferDeclaration)

buffer/src/main/java/io/netty/buffer/PooledByteBufAllocatorMetric.java
```diff
@@ -0,0 +1,122 @@
...
+        sb.append(StringUtil.simpleClassName(this)).append('(')
```
netkins: MINOR StringBuffer (or StringBuilder).append is called 2 consecutive times with literal Strings. Use a single append with a single combined String. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AConsecutiveLiteralAppends)

buffer/src/main/java/io/netty/buffer/PooledByteBufAllocatorMetric.java
```diff
@@ -0,0 +1,122 @@
...
+                .append("usedHeapMemory: ").append(usedHeapMemory()).append("; ")
```
netkins: MINOR StringBuffer (or StringBuilder).append is called 2 consecutive times with literal Strings. Use a single append with a single combined String. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AConsecutiveLiteralAppends)

buffer/src/main/java/io/netty/buffer/PooledByteBufAllocatorMetric.java
Scottmitch: @normanmaurer - maybe 256?

buffer/src/main/java/io/netty/buffer/PooledByteBufAllocatorMetric.java
Scottmitch: +1 Plz cleanup the string builder comments

buffer/src/main/java/io/netty/buffer/UnpooledByteBufAllocator.java
```diff
@@ -97,13 +97,24 @@ public boolean isDirectBufferPooled() {
...
+
+    private void decrementDirect(int amount) {
+        metric.directCounter.add(-amount);
+    }
+
+    private void incrementHeap(int amount) {
+        metric.heapCounter.add(amount);
+    }
+
+    private void decrementHeap(int amount) {
```
Scottmitch: package private because they are accessed inside another class's scope?

normanmaurer:
@Scottmitch I think I addressed all your comments

normanmaurer:
will merge once CI finishes.

normanmaurer:
Cherry-picked into 4.1 (3ad3356892e841b5e2cd7f6e87882fdf3fa2e68d) and 4.0 (4564f44)

PR #6486, opened 2017-03-02
Correct usages of internalNioBuffer

Scottmitch:
Motivation:
There are numerous usages of internalNioBuffer which hard code 0 for the index when the intention was to use the readerIndex().

Modifications:
- Remove hard coded 0 for the index and use readerIndex()

Result:
We are less susceptible to using the wrong index, and don't make assumptions about the ByteBufAllocator.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
+++ b/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
@@ -588,7 +588,7 @@ public final class ByteBufUtil {
             dst = alloc.buffer(length);
         }
         try {
-            final ByteBuffer dstBuf = dst.internalNioBuffer(0, length);
+            final ByteBuffer dstBuf = dst.internalNioBuffer(dst.readerIndex(), length);
             final int pos = dstBuf.position();
             CoderResult cr = encoder.encode(src, dstBuf, true);
             if (!cr.isUnderflow()) {
@@ -635,7 +635,7 @@ public final class ByteBufUtil {
             try {
                 buffer.writeBytes(src, readerIndex, len);
                 // Use internalNioBuffer(...) to reduce object creation.
-                decodeString(decoder, buffer.internalNioBuffer(0, len), dst);
+                decodeString(decoder, buffer.internalNioBuffer(buffer.readerIndex(), len), dst);
             } finally {
                 // Release the temporary buffer again.
                 buffer.release();
@@ -1087,7 +1087,7 @@ public final class ByteBufUtil {
                     ByteBuf heapBuffer =  buf.alloc().heapBuffer(length);
                     try {
                         heapBuffer.writeBytes(buf, index, length);
-                        decoder.decode(heapBuffer.internalNioBuffer(0, length));
+                        decoder.decode(heapBuffer.internalNioBuffer(heapBuffer.readerIndex(), length));
                     } finally {
                         heapBuffer.release();
                     }
--- a/buffer/src/test/java/io/netty/buffer/AbstractByteBufTest.java
+++ b/buffer/src/test/java/io/netty/buffer/AbstractByteBufTest.java
@@ -2120,14 +2120,14 @@ public abstract class AbstractByteBufTest {
 
     private void testInternalNioBuffer(int a) {
         ByteBuf buffer = newBuffer(2);
-        ByteBuffer buf = buffer.internalNioBuffer(0, 1);
+        ByteBuffer buf = buffer.internalNioBuffer(buffer.readerIndex(), 1);
         assertEquals(1, buf.remaining());
 
         byte[] data = new byte[a];
         PlatformDependent.threadLocalRandom().nextBytes(data);
         buffer.writeBytes(data);
 
-        buf = buffer.internalNioBuffer(0, a);
+        buf = buffer.internalNioBuffer(buffer.readerIndex(), a);
         assertEquals(a, buf.remaining());
 
         for (int i = 0; i < a; i++) {
@@ -2972,7 +2972,8 @@ public abstract class AbstractByteBufTest {
 
     @Test(expected = IllegalReferenceCountException.class)
     public void testInternalNioBufferAfterRelease() {
-        releasedBuffer().internalNioBuffer(0, 1);
+        ByteBuf releasedBuffer = releasedBuffer();
+        releasedBuffer.internalNioBuffer(releasedBuffer.readerIndex(), 1);
     }
 
     @Test(expected = IllegalReferenceCountException.class)
--- a/buffer/src/test/java/io/netty/buffer/EmptyByteBufTest.java
+++ b/buffer/src/test/java/io/netty/buffer/EmptyByteBufTest.java
@@ -50,7 +50,7 @@ public class EmptyByteBufTest {
         assertThat(empty.nioBuffer().position(), is(0));
         assertThat(empty.nioBuffer().limit(), is(0));
         assertThat(empty.nioBuffer(), is(sameInstance(empty.nioBuffer())));
-        assertThat(empty.nioBuffer(), is(sameInstance(empty.internalNioBuffer(0, 0))));
+        assertThat(empty.nioBuffer(), is(sameInstance(empty.internalNioBuffer(empty.readerIndex(), 0))));
     }
 
     @Test
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -198,7 +198,6 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
                         handler.singleBuffer[0] = toByteBuffer(out, writerIndex,
                             out.writableBytes());
                         result = opensslEngine.unwrap(in.nioBuffers(readerIndex, len), handler.singleBuffer);
-                        out.writerIndex(writerIndex + result.bytesProduced());
                     } finally {
                         handler.singleBuffer[0] = null;
                     }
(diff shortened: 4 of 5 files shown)


buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -588,7 +588,7 @@ static ByteBuf encodeString0(ByteBufAllocator alloc, boolean enforceHeap, CharBu
             dst = alloc.buffer(length);
         }
         try {
-            final ByteBuffer dstBuf = dst.internalNioBuffer(0, length);
+            final ByteBuffer dstBuf = dst.internalNioBuffer(dst.readerIndex(), length);
```
normanmaurer: thats fine that said it should not make any difference as we just allocated the buffer so the readerIndex() must be 0

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -635,7 +635,7 @@ static String decodeString(ByteBuf src, int readerIndex, int len, Charset charse
             try {
                 buffer.writeBytes(src, readerIndex, len);
                 // Use internalNioBuffer(...) to reduce object creation.
-                decodeString(decoder, buffer.internalNioBuffer(0, len), dst);
+                decodeString(decoder, buffer.internalNioBuffer(buffer.readerIndex(), len), dst);
```
normanmaurer: Same comment as above

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -1087,7 +1087,7 @@ public static boolean isText(ByteBuf buf, int index, int length, Charset charset
                     ByteBuf heapBuffer =  buf.alloc().heapBuffer(length);
                     try {
                         heapBuffer.writeBytes(buf, index, length);
-                        decoder.decode(heapBuffer.internalNioBuffer(0, length));
+                        decoder.decode(heapBuffer.internalNioBuffer(heapBuffer.readerIndex(), length));
```
normanmaurer: Same comment as above

buffer/src/test/java/io/netty/buffer/AbstractByteBufTest.java
```diff
@@ -2972,7 +2972,8 @@ public void testNioBufferAfterRelease1() {
 
     @Test(expected = IllegalReferenceCountException.class)
     public void testInternalNioBufferAfterRelease() {
-        releasedBuffer().internalNioBuffer(0, 1);
+        ByteBuf releasedBuffer = releasedBuffer();
+        releasedBuffer.internalNioBuffer(releasedBuffer.readerIndex(), 1);
```
normanmaurer: Same comments as above... should not make any difference.

buffer/src/test/java/io/netty/buffer/EmptyByteBufTest.java
```diff
@@ -50,7 +50,7 @@ public void testNioBuffer() {
         assertThat(empty.nioBuffer().position(), is(0));
         assertThat(empty.nioBuffer().limit(), is(0));
         assertThat(empty.nioBuffer(), is(sameInstance(empty.nioBuffer())));
-        assertThat(empty.nioBuffer(), is(sameInstance(empty.internalNioBuffer(0, 0))));
+        assertThat(empty.nioBuffer(), is(sameInstance(empty.internalNioBuffer(empty.readerIndex(), 0))));
```
normanmaurer: also should make no difference

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
```diff
@@ -261,7 +261,7 @@ private void flushBufferedData(ByteBuf out) {
             ByteBuffer outNioBuffer = out.internalNioBuffer(idx + HEADER_LENGTH, out.writableBytes() - HEADER_LENGTH);
             int pos = outNioBuffer.position();
             // We always want to start at position 0 as we take care of reusing the buffer in the encode(...) loop.
-            compressor.compress(buffer.internalNioBuffer(0, flushableBytes), outNioBuffer);
+            compressor.compress(buffer.internalNioBuffer(buffer.readerIndex(), flushableBytes), outNioBuffer);
```
normanmaurer: Good catch... This is a bug.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -395,26 +395,21 @@ private ByteBuf writeEncryptedData(final ByteBuffer src, int len) {
...
             final ByteBuf buf = alloc.directBuffer(len);
             try {
-                final long addr = memoryAddress(buf);
-
-                sslRead = SSL.readFromSSL(ssl, addr, len);
+                sslRead = SSL.readFromSSL(ssl, memoryAddress(buf), len);
                 if (sslRead > 0) {
                     dst.limit(pos + sslRead);
-                    buf.getBytes(0, dst);
+                    buf.getBytes(buf.readerIndex(), dst);
```
normanmaurer: As we just allocated the buffer this should make no difference. Not saying it makes things worse or so just saying that it will not have a different outcome then before

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -646,8 +641,7 @@ public final SSLEngineResult wrap(
                 } else {
                     assert bioReadCopyBuf.readableBytes() <= dst.remaining() : "The destination buffer " + dst +
                             " didn't have enough remaining space to hold the encrypted content in " + bioReadCopyBuf;
-                    dst.put(bioReadCopyBuf.internalNioBuffer(bioReadCopyBuf.readerIndex(),
-                                                      bioReadCopyBuf.readerIndex() + bytesProduced));
+                    dst.put(bioReadCopyBuf.internalNioBuffer(bioReadCopyBuf.readerIndex(), bytesProduced));
```
normanmaurer: Good catch, this is a bug!

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -828,7 +827,7 @@ private SSLEngineResult wrap(ByteBufAllocator alloc, SSLEngine engine, ByteBuf i
                 newDirectIn = alloc.directBuffer(readableBytes);
                 newDirectIn.writeBytes(in, readerIndex, readableBytes);
                 in0 = singleBuffer;
-                in0[0] = newDirectIn.internalNioBuffer(0, readableBytes);
+                in0[0] = newDirectIn.internalNioBuffer(newDirectIn.readerIndex(), readableBytes);
```
normanmaurer: Again this makes no difference a we just allocated the buffer. It will always be 0.

transport/src/test/java/io/netty/channel/ChannelOutboundBufferTest.java
```diff
@@ -52,7 +52,7 @@ public void testNioBuffersSingleBacked() {
         assertEquals(0, buffer.nioBufferCount());
 
         ByteBuf buf = copiedBuffer("buf1", CharsetUtil.US_ASCII);
-        ByteBuffer nioBuf = buf.internalNioBuffer(0, buf.readableBytes());
+        ByteBuffer nioBuf = buf.internalNioBuffer(buf.readerIndex(), buf.readableBytes());
```
normanmaurer: same

transport/src/test/java/io/netty/channel/ChannelOutboundBufferTest.java
```diff
@@ -109,7 +109,7 @@ public void testNioBuffersExpand2() {
         assertEquals(65, buffer.nioBufferCount());
         for (int i = 0;  i < buffer.nioBufferCount(); i++) {
             if (i < 65) {
-                assertEquals(buffers[i], buf.internalNioBuffer(0, buf.readableBytes()));
+                assertEquals(buffers[i], buf.internalNioBuffer(buf.readerIndex(), buf.readableBytes()));
```
normanmaurer: same

transport/src/test/java/io/netty/channel/ChannelOutboundBufferTest.java
```diff
@@ -84,7 +84,7 @@ public void testNioBuffersExpand() {
         ByteBuffer[] buffers = buffer.nioBuffers();
         assertEquals(64, buffer.nioBufferCount());
         for (int i = 0;  i < buffer.nioBufferCount(); i++) {
-            assertEquals(buffers[i], buf.internalNioBuffer(0, buf.readableBytes()));
+            assertEquals(buffers[i], buf.internalNioBuffer(buf.readerIndex(), buf.readableBytes()));
```
normanmaurer: same

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
Scottmitch: If the allocator always returns a new buffer and the reader index is guaranteed to be 0 ... then it makes no difference. However I'm not sure we need to make those assumptions in this code.

normanmaurer: I think if the buffer not has a reader index of 0 its a bug as we ask for a new buffer. That said the change is ok as it makes it more consistent with other usages

codec/src/main/java/io/netty/handler/codec/compression/Lz4FrameEncoder.java
Scottmitch: https://github.com/netty/netty/pull/6487

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
Scottmitch: see PR https://github.com/netty/netty/pull/6488

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
Scottmitch: This behavior is not a strict requirement of the `ByteBufAllocator` interface (although it is reasonable behavior and our default implementations should do this).

Scottmitch:
4.1 (2cff918) 4.0 (aaa0155)

Issue #6481, opened 2017-03-01
SSL_read failed: OpenSSL error


rkapsi:
Netty 4.1.9-SNAPSHOT with netty-tcnative 2.0.0.Beta6

I'll try to provide a repro but I'm seeing the following error in our application when I upgrade to Netty 4.1.9-SNAPSHOT and netty-tcnative 2.0.0.Beta6.

From what I can tell, the SSL hanshake completes, I receive the HTTP request, I respond with `ctx.writeAndFlush(FullHttpResponse)`, I see the response data in curl but it cuts off randomly and both ends report the following errors:

```
// Server: Netty w/ OpenSSL on port 8443
ReferenceCountedOpenSslEngine DEBUG: SSL_read failed: OpenSSL error: error:1409442E:SSL routines:ssl3_read_bytes:tlsv1 alert protocol version
```

```
// Client: curl -k https://localhost:8443
curl: (56) SSL read: error:1408F10B:SSL routines:SSL3_GET_RECORD:wrong version number, errno 0
```

An another observation is that curl seems to be a bit more stable. Some requests do succeed but Browsers (I tested Chrome and FF) on the other hand fail pretty reliably (every time). The problem goes away as soon as I switch back to Netty 4.1.8.

This resembles what we observed in the #6466 ticket (minus the Exception).

Scottmitch:
> Netty 4.1.9-SNAPSHOT

What commit hash are you using?
What version of OpenSSL?
Is there anything interesting about the payload/scenario that you thing may help reproduce? For example do you have to repeat the request many times, pipelining, etc..?

Scottmitch:
can you reproduce with a modified version of the [http hello world](https://github.com/netty/netty/blob/4.1/example/src/main/java/io/netty/example/http/helloworld/HttpHelloWorldServer.java) example?

Scottmitch:
can you also include some verbose output from curl `-iv`.

Can you reproduce with `openssl s_client -host localhost -port 8443` (then enter `GET / HTTP/1.1` after the handshake)?

normanmaurer:
@rkapsi also a big thank you for trying the SNAPSHOTS and provide feedback!

Scottmitch:
Also do you try to enable `SSL_MODE_ENABLE_PARTIAL_WRITE` for Netty?

rkapsi:
Still trying to come up with a reproducer but I've narrowed it down to a commit.

```
[...]

PR #6492, opened 2017-03-03
SslHandler and OpenSslEngine miscalculation of wrap destination buffe…

Scottmitch:
…r size

Motivation:
When we do a wrap operation we calculate the maximum size of the destination buffer ahead of time, and return a BUFFER_OVERFLOW exception if the destination buffer is not big enough. However if there is a CompositeByteBuf the wrap operation may consist of multiple ByteBuffers and each incurs its own overhead during the encryption. We currently don't account for the overhead required for encryption if there are multiple ByteBuffers and we assume the overhead will only apply once to the entire input size. If there is not enough room to write an entire encrypted packed into the BIO SSL_write will return -1 despite having actually written content to the BIO. We then attempt to retry the write with a bigger buffer, but because SSL_write is stateful the remaining bytes from the previous operation are put into the BIO. This results in sending the second half of the encrypted data being sent to the peer which is not of proper format and the peer will be confused and ultimately not get the expected data (which may result in a fatal error). In this case because SSL_write returns -1 we have no way to know how many bytes were actually consumed and so the best we can do is ensure that we always allocate a destination buffer with enough space so we are guaranteed to complete the write operation synchronously.

Modifications:
- SslHandler#allocateNetBuf should take into account how many ByteBuffers will be wrapped and apply the encryption overhead for each
- Include the TLS header length in the overhead computation

Result:
Fixes https://github.com/netty/netty/issues/6481

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -57,6 +57,7 @@ import javax.net.ssl.SSLSessionContext;
 import javax.security.cert.X509Certificate;
 
 import static io.netty.handler.ssl.OpenSsl.memoryAddress;
+import static io.netty.handler.ssl.SslUtils.SSL_RECORD_HEADER_LENGTH;
 import static io.netty.util.internal.EmptyArrays.EMPTY_CERTIFICATES;
 import static io.netty.util.internal.EmptyArrays.EMPTY_JAVAX_X509_CERTIFICATES;
 import static io.netty.util.internal.ObjectUtil.checkNotNull;
@@ -107,15 +108,14 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
      * allow up to 255 bytes. 16 bytes is the max for PKC#5 (which handles it the same way as PKC#7) as we use a block
      * size of 16. See <a href="https://tools.ietf.org/html/rfc5652#section-6.3">rfc5652#section-6.3</a>.
      *
-     * 16 (IV) + 48 (MAC) + 1 (Padding_length field) + 15 (Padding) + 1 (ContentType) + 2 (ProtocolVersion) + 2 (Length)
+     * TLS Header (5) + 16 (IV) + 48 (MAC) + 1 (Padding_length field) + 15 (Padding) + 1 (ContentType) +
+     * 2 (ProtocolVersion) + 2 (Length)
      *
      * TODO: We may need to review this calculation once TLS 1.3 becomes available.
      */
-    static final int MAX_ENCRYPTION_OVERHEAD_LENGTH = 15 + 48 + 1 + 16 + 1 + 2 + 2;
+    static final int MAX_TLS_RECORD_OVERHEAD_LENGTH = SSL_RECORD_HEADER_LENGTH + 16 + 48 + 1 + 15 + 1 + 2 + 2;
 
-    static final int MAX_ENCRYPTED_PACKET_LENGTH = MAX_PLAINTEXT_LENGTH + MAX_ENCRYPTION_OVERHEAD_LENGTH;
-
-    private static final int MAX_ENCRYPTION_OVERHEAD_DIFF = Integer.MAX_VALUE - MAX_ENCRYPTION_OVERHEAD_LENGTH;
+    static final int MAX_ENCRYPTED_PACKET_LENGTH = MAX_PLAINTEXT_LENGTH + MAX_TLS_RECORD_OVERHEAD_LENGTH;
 
     private static final AtomicIntegerFieldUpdater<ReferenceCountedOpenSslEngine> DESTROYED_UPDATER =
             AtomicIntegerFieldUpdater.newUpdater(ReferenceCountedOpenSslEngine.class, "destroyed");
@@ -561,7 +561,7 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
                     }
                 }
 
-                if (dst.remaining() < calculateOutNetBufSize(srcsLen)) {
+                if (dst.remaining() < calculateOutNetBufSize(srcsLen, endOffset - offset)) {
                     // Can not hold the maximum packet so we need to tell the caller to use a bigger destination
                     // buffer.
                     return new SSLEngineResult(BUFFER_OVERFLOW, getHandshakeStatus(), 0, 0);
@@ -772,7 +772,7 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
                 }
             }
 
-            if (len < SslUtils.SSL_RECORD_HEADER_LENGTH) {
+            if (len < SSL_RECORD_HEADER_LENGTH) {
                 return newResultMayFinishHandshake(BUFFER_UNDERFLOW, status, 0, 0);
             }
 
@@ -782,7 +782,7 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
                 throw new NotSslRecordException("not an SSL/TLS record");
             }
 
-            if (packetLength - SslUtils.SSL_RECORD_HEADER_LENGTH > capacity) {
+            if (packetLength - SSL_RECORD_HEADER_LENGTH > capacity) {
                 // No enough space in the destination buffer so signal the caller
                 // that the buffer needs to be increased.
                 return newResultMayFinishHandshake(BUFFER_OVERFLOW, status, 0, 0);
@@ -1606,9 +1606,8 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
         return destroyed != 0;
     }
 
-    static int calculateOutNetBufSize(int pendingBytes) {
-        return min(MAX_ENCRYPTED_PACKET_LENGTH, MAX_ENCRYPTION_OVERHEAD_LENGTH
-                + min(MAX_ENCRYPTION_OVERHEAD_DIFF, pendingBytes));
+    static int calculateOutNetBufSize(int pendingBytes, int numComponents) {
+        return (int) min(Integer.MAX_VALUE, pendingBytes + (long) MAX_TLS_RECORD_OVERHEAD_LENGTH * numComponents);
     }
 
     private final class OpenSslSession implements SSLSession, ApplicationProtocolAccessor {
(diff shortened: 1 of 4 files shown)


handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
```diff
@@ -560,4 +564,127 @@ public void operationComplete(Future<Channel> future) {
...
+        for (int i = 0; i < providers.length; ++i) {
+            for (int j = 0; j < providers.length; ++j) {
+                compositeBufSizeEstimationGuaranteesSynchronousWrite(providers[i], providers[j]);
+            }
+        }
+    }
+
+    private void compositeBufSizeEstimationGuaranteesSynchronousWrite(SslProvider serverProvider,
+                                                                      SslProvider clientProvider)
+            throws CertificateException, SSLException {
```
normanmaurer: formatting looks strange

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
```diff
@@ -560,4 +564,127 @@ public void operationComplete(Future<Channel> future) {
...
+            if (cc != null) {
+                cc.close().syncUninterruptibly();
+            }
+            if (sc != null) {
+                sc.close().syncUninterruptibly();
+            }
+            group.shutdownGracefully();
+
+            ReferenceCountUtil.release(sslServerCtx);
+            ReferenceCountUtil.release(sslClientCtx);
```
normanmaurer: call `ssc.delete();`

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
Scottmitch: how would you prefer this be formatted?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -107,11 +107,12 @@
      * allow up to 255 bytes. 16 bytes is the max for PKC#5 (which handles it the same way as PKC#7) as we use a block
      * size of 16. See <a href="https://tools.ietf.org/html/rfc5652#section-6.3">rfc5652#section-6.3</a>.
      *
-     * 16 (IV) + 48 (MAC) + 1 (Padding_length field) + 15 (Padding) + 1 (ContentType) + 2 (ProtocolVersion) + 2 (Length)
+     * 5 (TLS Header) + 16 (IV) + 48 (MAC) + 1 (Padding_length field) + 15 (Padding) + 1 (ContentType) +
```
normanmaurer: So we missed the 5 bytes ?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1606,8 +1607,8 @@ private boolean isDestroyed() {
         return destroyed != 0;
     }
 
-    static int calculateOutNetBufSize(int pendingBytes) {
-        return min(MAX_ENCRYPTED_PACKET_LENGTH, MAX_ENCRYPTION_OVERHEAD_LENGTH
+    static int calculateOutNetBufSize(int pendingBytes, int numComponents) {
+        return min(MAX_ENCRYPTED_PACKET_LENGTH, MAX_ENCRYPTION_OVERHEAD_LENGTH * numComponents
```
normanmaurer: why do we need to multiply this by the number of components ? Shouldn't it not matter as we only need to worry about the summary of the readable bytes of all of them as we can only create one record with all of the components in it.

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
```diff
@@ -560,4 +564,127 @@ public void operationComplete(Future<Channel> future) {
             ReferenceCountUtil.release(sslClientCtx);
         }
     }
+
+    @Test(timeout = 30000)
+    public void testCompositeBufSizeEstimationGuaranteesSynchronousWrite() throws CertificateException, SSLException {
```
normanmaurer: also would be nice if we had this as test that not depends on the `SslHandler` but only uses the SSLEngine itself

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
normanmaurer:
```java
private void compositeBufSizeEstimationGuaranteesSynchronousWrite(
        SslProvider serverProvider, SslProvider clientProvider) throws CertificateException, SSLException {
```
Something like this maybe ?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
Scottmitch: yes

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
Scottmitch: SslHandler is involved in the estimation process so it has a role to play in this failure.

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
Scottmitch: done

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
Scottmitch: done

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
Scottmitch: This is described in the commit message but if we have a `CompositeByteBuf` we break that down into `ByteBuffer[]` and call `SSL_write` for each of these. So each `ByteBuffer` will incur the tls record overhead. I wanted to review this more because it may be the case that we can apply the encryption overhead to all the bytes, and other apply the tls record overhead (header) for each `ByteBuffer`. I think this case ... so let me refine this estimate.

Scottmitch:
@normanmaurer - PTAL

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1606,9 +1605,9 @@ private boolean isDestroyed() {
         return destroyed != 0;
     }
 
-    static int calculateOutNetBufSize(int pendingBytes) {
-        return min(MAX_ENCRYPTED_PACKET_LENGTH, MAX_ENCRYPTION_OVERHEAD_LENGTH
-                + min(MAX_ENCRYPTION_OVERHEAD_DIFF, pendingBytes));
+    static int calculateOutNetBufSize(int pendingBytes, int numComponents) {
+        int amt = pendingBytes + MAX_TLS_RECORD_OVERHEAD_LENGTH * numComponents;
```
Scottmitch: @normanmaurer - I consulted with @davidben regarding the question in https://github.com/netty/netty/pull/6492#discussion_r104241664 ... each call to SSL_write will incur the `encryption+tls header` amount of overhead.

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -226,16 +226,13 @@ SSLEngineResult unwrap(SslHandler handler, ByteBuf in, int readerIndex, int len,
...
                 return handler.maxPacketBufferSize;
             }
         };
 
         static SslEngineType forEngine(SSLEngine engine) {
-            if (engine instanceof OpenSslEngine) {
-                return TCNATIVE;
-            }
-            return JDK;
+            return engine instanceof ReferenceCountedOpenSslEngine ? TCNATIVE : JDK;
```
johnou: JDK SslEngine when using OpenSslEngine and finalizer? 😕 Shouldn't this remain OpenSslEngine?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
johnou: Is Integer.MAX_VALUE safe?

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -193,7 +193,7 @@ SSLEngineResult unwrap(SslHandler handler, ByteBuf in, int readerIndex, int len,
                      * we can use a special {@link OpenSslEngine#unwrap(ByteBuffer[], ByteBuffer[])} method
                      * that accepts multiple {@link ByteBuffer}s without additional memory copies.
                      */
-                    OpenSslEngine opensslEngine = (OpenSslEngine) handler.engine;
+                    ReferenceCountedOpenSslEngine opensslEngine = (ReferenceCountedOpenSslEngine) handler.engine;
```
johnou: classcastexception when using openssl with finalizer?

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
Scottmitch: > JDK SslEngine when using OpenSslEngine and finalizer?

I don't understand this statement. Can you explain.

Basically `ReferenceCountedOpenSslEngine` is the base class of `OpenSslEngine` and is the class which provides the interface which `SslEngineType#TCNATIVE` depends upon.

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
Scottmitch: I don't think so. `ReferenceCountedOpenSslEngine` is the base class of `OpenSslEngine`.

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
johnou: eugh, I had it in my head that ReferenceCountedOpenSslEngine extends OpenSslEngine.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: @Scottmitch ok thanks for verify

Scottmitch:
4.1 (53fc693) 4.0 (ccd6098)

PR #6494, opened 2017-03-04
Fix misordered 'assertEquals' arguments in tests

fenik17:
Motivation:

Wrong order of the arguments in some 'assertEquals' applying.

Modifications:

Flip compared arguments.

Result:

Correct `assertEquals` usage.

### DIFF
--- a/buffer/src/test/java/io/netty/buffer/AbstractCompositeByteBufTest.java
+++ b/buffer/src/test/java/io/netty/buffer/AbstractCompositeByteBufTest.java
@@ -1004,8 +1004,7 @@ public abstract class AbstractCompositeByteBufTest extends AbstractByteBufTest {
         CompositeByteBuf cbuf = compositeBuffer();
         ByteBuf buf1 = buffer().writeByte((byte) 1);
         cbuf.addComponent(true, buf1);
-        ByteBuf buf2 = EMPTY_BUFFER;
-        cbuf.addComponent(true, buf2);
+        cbuf.addComponent(true, EMPTY_BUFFER);
         ByteBuf buf3 = buffer().writeByte((byte) 2);
         cbuf.addComponent(true, buf3);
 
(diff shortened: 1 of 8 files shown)


buffer/src/test/java/io/netty/buffer/AbstractCompositeByteBufTest.java
```diff
@@ -49,6 +49,7 @@
 /**
  * An abstract test class for composite channel buffers
  */
+@SuppressWarnings("deprecation")
```
Scottmitch: can we do this at a more fine granularity on the methods which cause the warnings?

buffer/src/test/java/io/netty/buffer/AbstractByteBufTest.java
```diff
@@ -2268,6 +2268,7 @@ private void testBytesInArrayMultipleThreads(final boolean slice) throws Excepti
         final CyclicBarrier barrier = new CyclicBarrier(11);
         for (int i = 0; i < 10; i++) {
             new Thread(new Runnable() {
+                @SuppressWarnings("MisorderedAssertEqualsArguments")
```
Scottmitch: which line(s) does this apply to? Based upon [1] it is not clear which line would cause this warning.

[1] https://searchcode.com/codesearch/view/7358059/

buffer/src/test/java/io/netty/buffer/AbstractCompositeByteBufTest.java
fenik17: Not related with PR. I just removed this.

buffer/src/test/java/io/netty/buffer/AbstractByteBufTest.java
fenik17: Static analysis in IntelliJ was triggered "MisorderedAssertEqualsArguments" warning on the `assertArrayEquals()` statements.
I made supression more targeted through `//noinspection`.

fenik17:
@Scottmitch addressed.

buffer/src/test/java/io/netty/buffer/AbstractByteBufTest.java
```diff
@@ -2281,10 +2281,12 @@ public void run() {
                         byte[] array = new byte[8];
                         buf.readBytes(array);
 
+                        //noinspection MisorderedAssertEqualsArguments
```
Scottmitch: I would prefer not to add comments for specific IDEs in the code if possible. https://github.com/netty/netty/pull/6494#discussion_r104469753 it is not clear to me why this warning is triggered based upon the description ... neither of the arguments to the `assert` call are a literal ... do you have a better description of this warning than I provided in https://github.com/netty/netty/pull/6494#discussion_r104469753?

fenik17: As I understand `MisorderedAssertEqualsArguments` triggering warning on the `assertEquals` and `assertArrayEquals` if the first argument is not a constant. I can't found better description.. Only this [commit to IDEA](http://git.jetbrains.org/?p=idea/community.git;a=commit;h=638b01d6de0e53bf29ccd72b1d6ee6ed707009b5).
Just keep as it is without any suppressions?

normanmaurer: yep SGTM

fenik17: pushed

Scottmitch: yah remove the suppression ... seems like a IntelliJ bug based upon the description.

fenik17:
@Scottmitch comments addressed :)

Scottmitch:
4.1 (2993760) 4.0 (e140b8b)

Issue #6036, opened 2016-11-18
Http2FrameCodecTest use of spy for mocking


Scottmitch:
Using `spy()` from mocking can cause exceptions to be thrown while stepping through and debugging a test. This can prevent from debugging and create a makings distinguishing between real issues and mocking framework issues difficult. We should consider not using `spy()` or work with the mocking folks to identify the root cause.

https://github.com/netty/netty/pull/5996#discussion_r87312045

Scottmitch:
@buchgr @normanmaurer /cc

normanmaurer:
@Scottmitch I will take a crack on this....

PR #6496, opened 2017-03-05
Accept two ways to start HTTP/2 over clear text(version 2)

chhsiao90:
Refs: #6430 , #5857

Still not quite sure what it really mean "play nicely with HttpServerUpgradeHandler".
I tried to implement this with finagle-http2-like design, and make it more general.
I think it's better than previous PR, but if it still not in a good direction, please kindly let me know 😄 

Motivation:

HTTP/2 support two ways to start on a no-tls tcp connection,
http/1.1 upgrade and prior knowlege methodology to start HTTP/2.
Currently, the http2-server from example only support
starting by upgrade. I think we can do a simple dispatch by peek first
bytes from inbound that match to prior knowledge preface or not and
determine which handlers to set into pipeline.

Modifications:

Add ClearTextHttp2ServerUpgradeHandler to support start HTTP/2 via clear
text with two approach. And update example/http2-server to support
this functionality.

Result:

netty HTTP/2 and the example http2-server accept for two ways to start
HTTP/2 over clear text.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandler.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandler.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.codec.http2;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufUtil;
+import io.netty.channel.ChannelHandler;
+import io.netty.channel.ChannelHandlerAdapter;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.ByteToMessageDecoder;
+import io.netty.handler.codec.http.HttpServerCodec;
+import io.netty.handler.codec.http.HttpServerUpgradeHandler;
+import io.netty.util.internal.UnstableApi;
+
+import java.util.List;
+
+import static io.netty.buffer.Unpooled.unreleasableBuffer;
+import static io.netty.handler.codec.http2.Http2CodecUtil.connectionPrefaceBuf;
+
+import static io.netty.util.internal.ObjectUtil.checkNotNull;
+
+/**
+ * Performing cleartext upgrade, by h2c HTTP upgrade or Prior Knowledge.
+ * This handler config pipeline for h2c upgrade when handler added.
+ * And will update pipeline once it detect the connection is starting HTTP/2 by
+ * prior knowledge or not.
+ */
+@UnstableApi
+public final class CleartextHttp2ServerUpgradeHandler extends ChannelHandlerAdapter {
+    private static final ByteBuf CONNECTION_PREFACE = unreleasableBuffer(connectionPrefaceBuf());
+
+    private final HttpServerCodec httpServerCodec;
+    private final HttpServerUpgradeHandler httpServerUpgradeHandler;
+    private final ChannelHandler http2ServerHandler;
+
+    /**
+     * Creates the channel handler provide cleartext HTTP/2 upgrade from HTTP
+     * upgrade or prior knowledge
+     *
+     * @param httpServerCodec the http server codec
+     * @param httpServerUpgradeHandler the http server upgrade handler for HTTP/2
+     * @param http2ServerHandler the http2 server handler, will be added into pipeline
+     *                           when starting HTTP/2 by prior knowledge
+     */
+    public CleartextHttp2ServerUpgradeHandler(HttpServerCodec httpServerCodec,
+                                              HttpServerUpgradeHandler httpServerUpgradeHandler,
+                                              ChannelHandler http2ServerHandler) {
+        this.httpServerCodec = checkNotNull(httpServerCodec, "httpServerCodec");
+        this.httpServerUpgradeHandler = checkNotNull(httpServerUpgradeHandler, "httpServerUpgradeHandler");
+        this.http2ServerHandler = checkNotNull(http2ServerHandler, "http2ServerHandler");
+    }
+
+    @Override
+    public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
+        ctx.pipeline()
+           .addBefore(ctx.name(), null, new PriorKnowledgeHandler())
+           .addBefore(ctx.name(), null, httpServerCodec)
+           .replace(this, null, httpServerUpgradeHandler);
+    }
+
+    /**
+     * Peek inbound message to determine current connection wants to start HTTP/2
+     * by HTTP upgrade or prior knowledge
+     */
+    private final class PriorKnowledgeHandler extends ByteToMessageDecoder {
+        @Override
+        protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws Exception {
+            int prefaceLength = CONNECTION_PREFACE.readableBytes();
+            int bytesRead = Math.min(in.readableBytes(), prefaceLength);
+
+            if (!ByteBufUtil.equals(CONNECTION_PREFACE, CONNECTION_PREFACE.readerIndex(),
+                                    in, in.readerIndex(), bytesRead)) {
+                ctx.pipeline().remove(this);
+            } else if (bytesRead == prefaceLength) {
+                // Full h2 preface match, removed source codec, using http2 codec to handle
+                // following network traffic
+                ctx.pipeline()
+                   .remove(httpServerCodec)
+                   .remove(httpServerUpgradeHandler)
+                   .replace(this, null, http2ServerHandler);
+                ctx.fireUserEventTriggered(PriorKnowledgeUpgradeEvent.INSTANCE);
+            }
+        }
+    }
+
+    /**
+     * User event that is fired to notify about HTTP/2 protocol is started.
+     */
+    public static final class PriorKnowledgeUpgradeEvent {
+        private static final PriorKnowledgeUpgradeEvent INSTANCE = new PriorKnowledgeUpgradeEvent();
+
+        private PriorKnowledgeUpgradeEvent() {
+        }
+    }
+}
--- a/codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
+++ b/codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
@@ -1,0 +1,195 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.codec.http2;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.channel.embedded.EmbeddedChannel;
+import io.netty.handler.codec.http.DefaultHttpHeaders;
+import io.netty.handler.codec.http.HttpMethod;
+import io.netty.handler.codec.http.HttpRequest;
+import io.netty.handler.codec.http.HttpServerCodec;
+import io.netty.handler.codec.http.HttpServerUpgradeHandler;
+import io.netty.handler.codec.http.HttpServerUpgradeHandler.UpgradeCodec;
+import io.netty.handler.codec.http.HttpServerUpgradeHandler.UpgradeCodecFactory;
+import io.netty.handler.codec.http.HttpServerUpgradeHandler.UpgradeEvent;
+import io.netty.handler.codec.http.HttpVersion;
+import io.netty.handler.codec.http.LastHttpContent;
+import io.netty.handler.codec.http2.CleartextHttp2ServerUpgradeHandler.PriorKnowledgeUpgradeEvent;
+import io.netty.handler.codec.http2.Http2Stream.State;
+import io.netty.util.CharsetUtil;
+import io.netty.util.ReferenceCountUtil;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.*;
+
+/**
+ * Tests for {@link CleartextHttp2ServerUpgradeHandler}
+ */
+public class CleartextHttp2ServerUpgradeHandlerTest {
+    private EmbeddedChannel channel;
+
+    private Http2FrameListener frameListener;
+
+    private Http2ConnectionHandler http2ConnectionHandler;
+
+    private List<Object> userEvents;
+
+    @Before
+    public void setUp() {
+        frameListener = mock(Http2FrameListener.class);
+
+        http2ConnectionHandler = new Http2ConnectionHandlerBuilder().frameListener(frameListener).build();
+
+        UpgradeCodecFactory upgradeCodecFactory = new UpgradeCodecFactory() {
+            @Override
+            public UpgradeCodec newUpgradeCodec(CharSequence protocol) {
+                return new Http2ServerUpgradeCodec(http2ConnectionHandler);
+            }
+        };
+
+        userEvents = new ArrayList<Object>();
+
+        HttpServerCodec httpServerCodec = new HttpServerCodec();
+        HttpServerUpgradeHandler upgradeHandler = new HttpServerUpgradeHandler(httpServerCodec, upgradeCodecFactory);
+
+        CleartextHttp2ServerUpgradeHandler handler = new CleartextHttp2ServerUpgradeHandler(
+                httpServerCodec, upgradeHandler, http2ConnectionHandler);
+        channel = new EmbeddedChannel(handler, new ChannelInboundHandlerAdapter() {
+            @Override
+            public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
+                userEvents.add(evt);
+            }
+        });
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        channel.finishAndReleaseAll();
+    }
+
+    @Test
+    public void priorKnowledge() throws Exception {
+        channel.writeInbound(Http2CodecUtil.connectionPrefaceBuf());
+
+        ByteBuf settingsFrame = settingsFrameBuf();
+
+        assertFalse(channel.writeInbound(settingsFrame));
+
+        assertEquals(1, userEvents.size());
+        assertTrue(userEvents.get(0) instanceof PriorKnowledgeUpgradeEvent);
+
+        assertEquals(100, http2ConnectionHandler.connection().local().maxActiveStreams());
+        assertEquals(65535, http2ConnectionHandler.connection().local().flowController().initialWindowSize());
+
+        verify(frameListener).onSettingsRead(
+                any(ChannelHandlerContext.class), eq(expectedSettings()));
+    }
+
+    @Test
+    public void upgrade() throws Exception {
+        String upgradeString = "GET / HTTP/1.1\r\n" +
+                               "Host: example.com\r\n" +
+                               "Connection: Upgrade, HTTP2-Settings\r\n" +
+                               "Upgrade: h2c\r\n" +
+                               "HTTP2-Settings: AAMAAABkAAQAAP__\r\n\r\n";
+        ByteBuf upgrade = Unpooled.buffer().writeBytes(upgradeString.getBytes(CharsetUtil.US_ASCII));
+
+        assertFalse(channel.writeInbound(upgrade));
+
+        assertEquals(1, userEvents.size());
+
+        Object userEvent = userEvents.get(0);
+        assertTrue(userEvent instanceof UpgradeEvent);
+        assertEquals("h2c", ((UpgradeEvent) userEvent).protocol());
+        ReferenceCountUtil.release(userEvent);
+
+        assertEquals(100, http2ConnectionHandler.connection().local().maxActiveStreams());
+        assertEquals(65535, http2ConnectionHandler.connection().local().flowController().initialWindowSize());
+
+        assertEquals(1, http2ConnectionHandler.connection().numActiveStreams());
+        assertNotNull(http2ConnectionHandler.connection().stream(1));
+
+        Http2Stream stream = http2ConnectionHandler.connection().stream(1);
+        assertEquals(State.HALF_CLOSED_REMOTE, stream.state());
+        assertFalse(stream.isHeadersSent());
+    }
+
+    @Test
+    public void priorKnowledgeInFragments() throws Exception {
+        ByteBuf connectionPreface = Http2CodecUtil.connectionPrefaceBuf();
+        assertFalse(channel.writeInbound(connectionPreface.readBytes(5), connectionPreface));
+
+        ByteBuf settingsFrame = settingsFrameBuf();
+        assertFalse(channel.writeInbound(settingsFrame));
+
+        assertEquals(1, userEvents.size());
+        assertTrue(userEvents.get(0) instanceof PriorKnowledgeUpgradeEvent);
+
+        assertEquals(100, http2ConnectionHandler.connection().local().maxActiveStreams());
+        assertEquals(65535, http2ConnectionHandler.connection().local().flowController().initialWindowSize());
+
+        verify(frameListener).onSettingsRead(
+                any(ChannelHandlerContext.class), eq(expectedSettings()));
+    }
+
+    @Test
+    public void downgrade() throws Exception {
+        String requestString = "GET / HTTP/1.1\r\n" +
+                         "Host: example.com\r\n\r\n";
+        ByteBuf inbound = Unpooled.buffer().writeBytes(requestString.getBytes(CharsetUtil.US_ASCII));
+
+        assertTrue(channel.writeInbound(inbound));
+
+        Object firstInbound = channel.readInbound();
+        assertTrue(firstInbound instanceof HttpRequest);
+        HttpRequest request = (HttpRequest) firstInbound;
+        assertEquals(HttpMethod.GET, request.method());
+        assertEquals("/", request.uri());
+        assertEquals(HttpVersion.HTTP_1_1, request.protocolVersion());
+        assertEquals(new DefaultHttpHeaders().add("Host", "example.com"), request.headers());
+
+        ((LastHttpContent) channel.readInbound()).release();
+
+        assertNull(channel.readInbound());
+    }
+
+    private static ByteBuf settingsFrameBuf() {
+        ByteBuf settingsFrame = Unpooled.buffer();
+        settingsFrame.writeMedium(12); // Payload length
+        settingsFrame.writeByte(0x4); // Frame type
+        settingsFrame.writeByte(0x0); // Flags
+        settingsFrame.writeInt(0x0); // StreamId
+        settingsFrame.writeShort(0x3);
+        settingsFrame.writeInt(100);
+        settingsFrame.writeShort(0x4);
+        settingsFrame.writeInt(65535);
+
+        return settingsFrame;
+    }
+
+    private static Http2Settings expectedSettings() {
+        return new Http2Settings().maxConcurrentStreams(100).initialWindowSize(65535);
+    }
+}
--- a/example/src/main/java/io/netty/example/http2/helloworld/server/HelloWorldHttp2Handler.java
+++ b/example/src/main/java/io/netty/example/http2/helloworld/server/HelloWorldHttp2Handler.java
@@ -18,6 +18,7 @@ package io.netty.example.http2.helloworld.server;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufUtil;
 import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.http.FullHttpRequest;
 import io.netty.handler.codec.http.HttpServerUpgradeHandler;
 import io.netty.handler.codec.http2.DefaultHttp2Headers;
 import io.netty.handler.codec.http2.Http2ConnectionDecoder;
@@ -48,6 +49,14 @@ public final class HelloWorldHttp2Handler extends Http2ConnectionHandler impleme
         super(decoder, encoder, initialSettings);
     }
 
+    private static Http2Headers http1HeadersToHttp2Headers(FullHttpRequest request) {
+        return new DefaultHttp2Headers()
+                .authority(request.headers().get("Host"))
+                .method("GET")
+                .path(request.uri())
+                .scheme("http");
+    }
+
     /**
      * Handles the cleartext HTTP upgrade event. If an upgrade occurred, sends a simple response via HTTP/2
      * on stream 1 (the stream specifically reserved for cleartext HTTP upgrade).
@@ -55,11 +64,9 @@ public final class HelloWorldHttp2Handler extends Http2ConnectionHandler impleme
     @Override
     public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
         if (evt instanceof HttpServerUpgradeHandler.UpgradeEvent) {
-            // Write an HTTP/2 response to the upgrade request
-            Http2Headers headers =
-                    new DefaultHttp2Headers().status(OK.codeAsText())
-                    .set(new AsciiString(UPGRADE_RESPONSE_HEADER), new AsciiString("true"));
-            encoder().writeHeaders(ctx, 1, headers, 0, true, ctx.newPromise());
+            HttpServerUpgradeHandler.UpgradeEvent upgradeEvent =
+                    (HttpServerUpgradeHandler.UpgradeEvent) evt;
+            onHeadersRead(ctx, 1, http1HeadersToHttp2Headers(upgradeEvent.upgradeRequest()), 0 , true);
         }
         super.userEventTriggered(ctx, evt);
     }
--- a/example/src/main/java/io/netty/example/http2/helloworld/server/Http2ServerInitializer.java
+++ b/example/src/main/java/io/netty/example/http2/helloworld/server/Http2ServerInitializer.java
@@ -28,6 +28,7 @@ import io.netty.handler.codec.http.HttpServerCodec;
 import io.netty.handler.codec.http.HttpServerUpgradeHandler;
 import io.netty.handler.codec.http.HttpServerUpgradeHandler.UpgradeCodec;
 import io.netty.handler.codec.http.HttpServerUpgradeHandler.UpgradeCodecFactory;
+import io.netty.handler.codec.http2.CleartextHttp2ServerUpgradeHandler;
 import io.netty.handler.codec.http2.Http2CodecUtil;
 import io.netty.handler.codec.http2.Http2ServerUpgradeCodec;
 import io.netty.handler.ssl.SslContext;
@@ -88,9 +89,12 @@ public class Http2ServerInitializer extends ChannelInitializer<SocketChannel> {
     private void configureClearText(SocketChannel ch) {
         final ChannelPipeline p = ch.pipeline();
         final HttpServerCodec sourceCodec = new HttpServerCodec();
+        final HttpServerUpgradeHandler upgradeHandler = new HttpServerUpgradeHandler(sourceCodec, upgradeCodecFactory);
+        final CleartextHttp2ServerUpgradeHandler cleartextHttp2ServerUpgradeHandler =
+                new CleartextHttp2ServerUpgradeHandler(sourceCodec, upgradeHandler,
+                                                       new HelloWorldHttp2HandlerBuilder().build());
 
-        p.addLast(sourceCodec);
-        p.addLast(new HttpServerUpgradeHandler(sourceCodec, upgradeCodecFactory));
+        p.addLast(cleartextHttp2ServerUpgradeHandler);
         p.addLast(new SimpleChannelInboundHandler<HttpMessage>() {
             @Override
             protected void channelRead0(ChannelHandlerContext ctx, HttpMessage msg) throws Exception {


codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
```diff
@@ -0,0 +1,89 @@
...
+        PriorKnowlegeHandler() {
+            prefaceString = Http2CodecUtil.connectionPrefaceBuf().retain();
+        }
+
+        @Override
+        protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws Exception {
+            int prefaceLength = prefaceString.readableBytes();
+            int bytesRead = Math.min(in.readableBytes(), prefaceLength);
+
+            if (!ByteBufUtil.equals(prefaceString, prefaceString.readerIndex(), in, in.readerIndex(), bytesRead)) {
```
chhsiao90: Here has some performance concern, that it will always compare from the start of preface when the prior knowledge preface was separate into several fragments.
But I think that's fine because the preface is not very long (24 bytes)
If still concern, please kindly let me know!

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * Providing how starting http2 protocol, by h2c HTTP upgrade, or by h2 with Prior Knowledge.
+ * This handler will set up {@link PriorKnowlegeHandler}, {@link HttpServerCodec},
+ * and {@link UpgradeCodecFactory} into the pipeline when handler added.
+ * And will set up pipeline correctly once it detect the protocol is h2c or h2.
+ */
+public class ClearTextHttp2ServerUpgradeHandler extends ChannelHandlerAdapter {
+
+    private HttpServerCodec httpServerCodec;
+    private ChannelHandler priorKnowledgeUpgradeHandler;
```
chhsiao90: I used priorKnowledgeUpgradeHandler to decouple the using of ```Http2Codec``` or ```Http2ConnectionHandler```

chhsiao90:
Hi @nhnFreespirit @mosesn @Scottmitch @normanmaurer PTAL, thank you.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Providing how starting http2 protocol, by h2c HTTP upgrade, or by h2 with Prior Knowledge.
+ * This handler will set up {@link PriorKnowlegeHandler}, {@link HttpServerCodec},
+ * and {@link UpgradeCodecFactory} into the pipeline when handler added.
+ * And will set up pipeline correctly once it detect the protocol is h2c or h2.
+ */
+public class ClearTextHttp2ServerUpgradeHandler extends ChannelHandlerAdapter {
+
+    private HttpServerCodec httpServerCodec;
```
fenik17: Fields could be 'final'.

chhsiao90:
Sure! Thanks for the advice.

example/src/main/java/io/netty/example/http2/helloworld/server/HelloWorldHttp2Handler.java
```diff
@@ -48,18 +49,24 @@
         super(decoder, encoder, initialSettings);
     }
 
+    private static Http2Headers http1HeadersToHttp2Headers(FullHttpRequest request) {
```
netkins: MINOR Make "http1HeadersToHttp2Headers" a "static" method. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2325)

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
chhsiao90: Sure! Thanks for the advice.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
```diff
@@ -0,0 +1,88 @@
...
+import static io.netty.buffer.Unpooled.unreleasableBuffer;
+import static io.netty.handler.codec.http2.Http2CodecUtil.connectionPrefaceBuf;
+
+/**
+ * Providing how starting http2 protocol, by h2c HTTP upgrade, or by h2 with Prior Knowledge.
+ * This handler will set up {@link PriorKnowlegeHandler}, {@link HttpServerCodec},
+ * and {@link UpgradeCodecFactory} into the pipeline when handler added.
+ * And will set up pipeline correctly once it detect the protocol is h2c or h2.
+ */
+public class ClearTextHttp2ServerUpgradeHandler extends ChannelHandlerAdapter {
```
mosesn: nit: Cleartext is one word, so "Text" should not be capitalized.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
```diff
@@ -0,0 +1,88 @@
...
+import io.netty.handler.codec.http.HttpServerUpgradeHandler;
+import io.netty.handler.codec.http.HttpServerUpgradeHandler.UpgradeCodecFactory;
+
+import java.util.List;
+
+import static io.netty.buffer.Unpooled.unreleasableBuffer;
+import static io.netty.handler.codec.http2.Http2CodecUtil.connectionPrefaceBuf;
+
+/**
+ * Providing how starting http2 protocol, by h2c HTTP upgrade, or by h2 with Prior Knowledge.
```
mosesn: I think h2 is http/2 over TLS, I'm not sure whether prior knowledge has a name other than just prior knowledge.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
```diff
@@ -0,0 +1,88 @@
...
+                // Full h2 preface match, removed source codec, using http2 codec to handle
+                // following network traffic
+                ctx.pipeline().remove(httpServerCodec);
+                ctx.pipeline().remove(httpServerUpgradeHandler);
+
+                ctx.pipeline().addAfter(ctx.name(), null, priorKnowledgeUpgradeHandler);
+                out.add(in.readBytes(in.readableBytes()));
+
+                ctx.pipeline().remove(this);
+            } else {
```
mosesn: we don't need the else case here.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
```diff
@@ -0,0 +1,88 @@
...
+
+    class PriorKnowlegeHandler extends ByteToMessageDecoder {
+        @Override
+        protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws Exception {
+            int prefaceLength = CONNECTION_PREFACE.readableBytes();
+            int bytesRead = Math.min(in.readableBytes(), prefaceLength);
+
+            if (!ByteBufUtil.equals(CONNECTION_PREFACE, CONNECTION_PREFACE.readerIndex(),
+                                    in, in.readerIndex(), bytesRead)) {
+                out.add(in.readBytes(in.readableBytes()));
```
mosesn: do we need to call readBytes here?  can we pass it in directly?

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
```diff
@@ -0,0 +1,88 @@
...
+                ctx.pipeline().remove(this);
+            } else if (bytesRead == prefaceLength) {
+                // Full h2 preface match, removed source codec, using http2 codec to handle
+                // following network traffic
+                ctx.pipeline().remove(httpServerCodec);
+                ctx.pipeline().remove(httpServerUpgradeHandler);
+
+                ctx.pipeline().addAfter(ctx.name(), null, priorKnowledgeUpgradeHandler);
```
mosesn: won't this be an NPE if ctx.name() == null?  how about using `replace` instead?

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
```diff
@@ -0,0 +1,88 @@
...
+                out.add(in.readBytes(in.readableBytes()));
```
mosesn: can we pass `in` here directly?

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
```diff
@@ -0,0 +1,88 @@
...
+                                              ChannelHandler priorKnowledgeUpgradeHandler) {
+        this.httpServerCodec = httpServerCodec;
+        this.httpServerUpgradeHandler = httpServerUpgradeHandler;
+        this.priorKnowledgeUpgradeHandler = priorKnowledgeUpgradeHandler;
+    }
+
+    @Override
+    public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
+        ctx.pipeline()
+           .addBefore(ctx.name(), null, new PriorKnowlegeHandler())
```
mosesn: here and elsewhere, this seems to rely on the assumption that ctx.name() isn't named null.  can we do it without using addBefore?

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
mosesn: it seems like this is installed with the name `null` by `ClearTextHttp2ServerUpgradeHandler`.  how does this work?

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
```diff
@@ -0,0 +1,88 @@
...
+    private final HttpServerCodec httpServerCodec;
+    private final ChannelHandler priorKnowledgeUpgradeHandler;
+    private final HttpServerUpgradeHandler httpServerUpgradeHandler;
+
+    public ClearTextHttp2ServerUpgradeHandler(HttpServerCodec httpServerCodec,
+                                              HttpServerUpgradeHandler httpServerUpgradeHandler,
+                                              ChannelHandler priorKnowledgeUpgradeHandler) {
+        this.httpServerCodec = httpServerCodec;
+        this.httpServerUpgradeHandler = httpServerUpgradeHandler;
+        this.priorKnowledgeUpgradeHandler = priorKnowledgeUpgradeHandler;
```
mosesn: based on context, it seems like this is supposed to be the http2codec?  this isn't really clear based on the name, can you either document the arguments or change the name to be a bit clearer?

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
```diff
@@ -0,0 +1,88 @@
...
+        protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws Exception {
+            int prefaceLength = CONNECTION_PREFACE.readableBytes();
+            int bytesRead = Math.min(in.readableBytes(), prefaceLength);
+
+            if (!ByteBufUtil.equals(CONNECTION_PREFACE, CONNECTION_PREFACE.readerIndex(),
+                                    in, in.readerIndex(), bytesRead)) {
+                out.add(in.readBytes(in.readableBytes()));
+                ctx.pipeline().remove(this);
+            } else if (bytesRead == prefaceLength) {
+                // Full h2 preface match, removed source codec, using http2 codec to handle
```
mosesn: we should send the Http2 UpgradeEvent too, so that handlers outside of the CleartextHttp2ServerUpgradeHandler can handle the upgrades uniformly.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.List;
+
+import static io.netty.buffer.Unpooled.unreleasableBuffer;
+import static io.netty.handler.codec.http2.Http2CodecUtil.connectionPrefaceBuf;
+
+/**
+ * Providing how starting http2 protocol, by h2c HTTP upgrade, or by h2 with Prior Knowledge.
+ * This handler will set up {@link PriorKnowlegeHandler}, {@link HttpServerCodec},
+ * and {@link UpgradeCodecFactory} into the pipeline when handler added.
+ * And will set up pipeline correctly once it detect the protocol is h2c or h2.
```
mosesn: these docs aren't that clear right now.  I think rather than explaining the implementation details, it's better to describe the outcome.  The part where we have the least details "will set up pipeline correctly" should be the focus of this javadoc.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
normanmaurer: nope... but you should use `in.readRetainedSlice(in.readableBytes())` to eliminate the memory copy.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
normanmaurer: @mosesn `ctx.name()` can never be null. When its installed with `null` a name will be generated for you and assigned.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
normanmaurer: Same comment as above... Just remove the line

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
normanmaurer: +1

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
normanmaurer: @mosesn the name can never be null as explained in the other comment.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
```diff
@@ -0,0 +1,88 @@
...
+    @Override
+    public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
+        ctx.pipeline()
+           .addBefore(ctx.name(), null, new PriorKnowlegeHandler())
+           .addBefore(ctx.name(), null, httpServerCodec)
+           .addBefore(ctx.name(), null, httpServerUpgradeHandler)
+           .remove(this);
+    }
+
+    class PriorKnowlegeHandler extends ByteToMessageDecoder {
```
normanmaurer: private final ?

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
normanmaurer: final

normanmaurer: Also please mark it via `@UnstableApi`

codec-http2/src/test/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandlerTest.java
```diff
@@ -0,0 +1,195 @@
...
+        channel.finishAndReleaseAll();
+    }
+
+    @Test
+    public void priorKnowledge() throws Exception {
+        channel.writeInbound(Http2CodecUtil.connectionPrefaceBuf());
+
+        ByteBuf settingsFrame = settingsFrameBuf();
+
+        channel.writeInbound(settingsFrame);
```
normanmaurer: assert the return value

codec-http2/src/test/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandlerTest.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        channel.finishAndReleaseAll();
+    }
+
+    @Test
+    public void priorKnowledge() throws Exception {
+        channel.writeInbound(Http2CodecUtil.connectionPrefaceBuf());
```
normanmaurer: assert the return value

codec-http2/src/test/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandlerTest.java
```diff
@@ -0,0 +1,195 @@
...
+    @Test
+    public void upgrade() throws Exception {
+        String upgradeString = "GET / HTTP/1.1\r\n" +
+                               "Host: example.com\r\n" +
+                               "Connection: Upgrade, HTTP2-Settings\r\n" +
+                               "Upgrade: h2c\r\n" +
+                               "HTTP2-Settings: AAMAAABkAAQAAP__\r\n\r\n";
+        ByteBuf upgrade = Unpooled.buffer().writeBytes(upgradeString.getBytes(CharsetUtil.US_ASCII));
+
+        channel.writeInbound(upgrade);
```
normanmaurer: assert the return value

codec-http2/src/test/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandlerTest.java
```diff
@@ -0,0 +1,195 @@
...
+        inOrder.verify(http2ConnectionHandler).onHttpServerUpgrade(
+                new Http2Settings().maxConcurrentStreams(100).initialWindowSize(65535));
+        inOrder.verify(http2ConnectionHandler).userEventTriggered(
+                any(ChannelHandlerContext.class), isA(UpgradeEvent.class));
+    }
+
+    @Test
+    public void priorKnowledgeInFragments() throws Exception {
+        ByteBuf connectionPreface = Http2CodecUtil.connectionPrefaceBuf();
+        channel.writeInbound(connectionPreface.readBytes(5), connectionPreface);
```
normanmaurer: assert return value

codec-http2/src/test/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandlerTest.java
```diff
@@ -0,0 +1,195 @@
...
+
+        ByteBuf settingsFrame = settingsFrameBuf();
+        channel.writeInbound(settingsFrame);
```
normanmaurer: assert return value

codec-http2/src/test/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandlerTest.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    @Test
+    public void downgrade() throws Exception {
+        String requestString = "GET / HTTP/1.1\r\n" +
+                         "Host: example.com\r\n\r\n";
+        ByteBuf inbound = Unpooled.buffer().writeBytes(requestString.getBytes(CharsetUtil.US_ASCII));
+
+        try {
+            channel.writeInbound(inbound);
```
normanmaurer: assert return value

codec-http2/src/test/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandlerTest.java
```diff
@@ -0,0 +1,195 @@
...
+
+            assertTrue(unhandledInbounds.get(1) instanceof LastHttpContent);
+        } finally {
+            for (Object obj : unhandledInbounds) {
+                ReferenceCountUtil.release(obj);
+            }
+        }
+    }
+
+    private ByteBuf settingsFrameBuf() {
```
normanmaurer: static

codec-http2/src/test/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandlerTest.java
```diff
@@ -0,0 +1,195 @@
...
+        settingsFrame.writeInt(0x0); // StreamId
+        settingsFrame.writeShort(0x3);
+        settingsFrame.writeInt(100);
+        settingsFrame.writeShort(0x4);
+        settingsFrame.writeInt(65535);
+
+        return settingsFrame;
+    }
+
+    private Http2Settings expectedSettings() {
```
normanmaurer: static

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
chhsiao90: Agreed, maybe just prior knowledge would be better.
h2 would make some confused!

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
chhsiao90: Thanks for the advise, I will try improved this javadoc, and focus on "will set up pipeline correctly"

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
chhsiao90: OK, I will think a better name for it.
But it cannot be http2codec as the [comment](https://github.com/netty/netty/pull/6430#discussion_r103378622) said it should not force Http2Codec into pipeline.
So it can be configured as Http2ConnectionHandler.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
chhsiao90: Thanks, that advise was helpful!
Still not familiar with pipeline and buffer handling.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
chhsiao90: Is that Http2UpgradeEvent will make confused about it was come from h2c or prior knowledge?
Should I create another event for it?
WDYT

chhsiao90:
Thanks for the review @mosesn @normanmaurer
All of the comments were helpful!
I will address these and push it later!

codec-http2/src/test/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandlerTest.java
chhsiao90: Done by https://github.com/netty/netty/pull/6496/commits/94628b69d07283027a3d9bc955e49061e50fedd9

codec-http2/src/test/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandlerTest.java
chhsiao90: Done by https://github.com/netty/netty/pull/6496/commits/94628b69d07283027a3d9bc955e49061e50fedd9

codec-http2/src/test/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandlerTest.java
chhsiao90: Done by https://github.com/netty/netty/pull/6496/commits/94628b69d07283027a3d9bc955e49061e50fedd9

codec-http2/src/test/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandlerTest.java
chhsiao90: Done by https://github.com/netty/netty/pull/6496/commits/94628b69d07283027a3d9bc955e49061e50fedd9

codec-http2/src/test/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandlerTest.java
chhsiao90: Done by https://github.com/netty/netty/pull/6496/commits/94628b69d07283027a3d9bc955e49061e50fedd9

codec-http2/src/test/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandlerTest.java
chhsiao90: Done by https://github.com/netty/netty/pull/6496/commits/94628b69d07283027a3d9bc955e49061e50fedd9

codec-http2/src/test/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandlerTest.java
chhsiao90: Done by https://github.com/netty/netty/pull/6496/commits/94628b69d07283027a3d9bc955e49061e50fedd9

codec-http2/src/test/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandlerTest.java
chhsiao90: Done by https://github.com/netty/netty/pull/6496/commits/94628b69d07283027a3d9bc955e49061e50fedd9

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
chhsiao90: Done by https://github.com/netty/netty/pull/6496/commits/94628b69d07283027a3d9bc955e49061e50fedd9

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
chhsiao90: Done by https://github.com/netty/netty/pull/6496/commits/94628b69d07283027a3d9bc955e49061e50fedd9

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
chhsiao90: Done by https://github.com/netty/netty/pull/6496/commits/94628b69d07283027a3d9bc955e49061e50fedd9

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
chhsiao90: Done by https://github.com/netty/netty/pull/6496/commits/94628b69d07283027a3d9bc955e49061e50fedd9

chhsiao90:
Think I almost address all comments except one [comment](https://github.com/netty/netty/pull/6496#discussion_r104697896) about raise UpgradeEvent.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
chhsiao90: I think we should create another event for that.
Because the UpgradeEvent sent during HTTP upgrade had package-scope constructor and it had protocol and upgradeRequest for it's arguments.
There is some difficulty about reuse the UpgradeEvent like HTTP upgrade do.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
chhsiao90: Done by https://github.com/netty/netty/pull/6496/commits/4ffb05a244b094c81272077c84bdefca56e2bb5e

And I think I'm not very good at write document or naming a good variable name 😟 .
So if it seems weird or unclear please kindly let me know if you had some better idea!
Thanks.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
chhsiao90: Had renamed it by https://github.com/netty/netty/pull/6496/commits/9573cba689fd2b07a3397ac50ed174583a4dab30

But it seems that I didn't commit about the file moving because I forgot to set for case-sensitive commit.
I will take case of it when all things looks great and I will rebase and squash it!

fenik17: >  I forgot to set for case-sensitive commit.

It's a reason of the build failure..

chhsiao90: OK... let me do it now 😆

fenik17: hm... still failed.

> ClearTextHttp2ServerUpgradeHandlerTest.java:49: The name of the outer type and the file do not match.

looks like CIs bug )

chhsiao90:
Uh... I forgot the unit test...

Nikolay Fedorovskikh <notifications@github.com> 於 2017年3月9日週四 下午8:57 寫道：

fenik17:
Indeed. It's test. CI work fine )

> Uh... I forgot the unit test...

codec-http2/src/main/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandler.java
```diff
@@ -0,0 +1,100 @@
...
+            int prefaceLength = CONNECTION_PREFACE.readableBytes();
+            int bytesRead = Math.min(in.readableBytes(), prefaceLength);
+
+            if (!ByteBufUtil.equals(CONNECTION_PREFACE, CONNECTION_PREFACE.readerIndex(),
+                                    in, in.readerIndex(), bytesRead)) {
+                ctx.pipeline().remove(this);
+            } else if (bytesRead == prefaceLength) {
+                // Full h2 preface match, removed source codec, using http2 codec to handle
+                // following network traffic
+                ctx.pipeline().remove(httpServerCodec);
```
Scottmitch: consider saving `ctx.pipeline()` in a local variable or chaining calls like is done above in `handlerAdded`

codec-http2/src/main/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandler.java
```diff
@@ -0,0 +1,100 @@
...
+                ctx.pipeline().remove(httpServerUpgradeHandler);
+
+                ctx.pipeline().addAfter(ctx.name(), null, http2ServerHandler);
+
+                ctx.pipeline().remove(this);
+            } else {
+                // Wait for further bytes
```
Scottmitch: consider just removing the `else {}` if there is nothing todo.

codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
```diff
@@ -0,0 +1,176 @@
...
+        http2ConnectionHandler = new Http2ConnectionHandlerBuilder().frameListener(frameListener).build();
+        http2ConnectionHandler = spy(http2ConnectionHandler);
+
+        upgradeCodecFactory = new UpgradeCodecFactory() {
+            @Override
+            public UpgradeCodec newUpgradeCodec(CharSequence protocol) {
+                return new Http2ServerUpgradeCodec(http2ConnectionHandler);
+            }
+        };
+        upgradeCodecFactory = spy(upgradeCodecFactory);
```
Scottmitch: lets avoid using `spy`. https://github.com/netty/netty/issues/6036 still appears to be causing issues.

codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
```diff
@@ -0,0 +1,176 @@
...
+    private Http2FrameListener frameListener;
+
+    private Http2ConnectionHandler http2ConnectionHandler;
+
+    @Before
+    public void setUp() {
+        frameListener = mock(Http2FrameListener.class);
+
+        http2ConnectionHandler = new Http2ConnectionHandlerBuilder().frameListener(frameListener).build();
+        http2ConnectionHandler = spy(http2ConnectionHandler);
```
Scottmitch: lets avoid using `spy`. https://github.com/netty/netty/issues/6036 still appears to be causing issues.

codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
```diff
@@ -0,0 +1,176 @@
...
+        inOrder.verify(upgradeCodecFactory).newUpgradeCodec("h2c");
+        inOrder.verify(http2ConnectionHandler).onHttpServerUpgrade(
+                new Http2Settings().maxConcurrentStreams(100).initialWindowSize(65535));
+        inOrder.verify(http2ConnectionHandler).userEventTriggered(
+                any(ChannelHandlerContext.class), isA(UpgradeEvent.class));
+    }
+
+    @Test
+    public void priorKnowledgeInFragments() throws Exception {
+        ByteBuf connectionPreface = Http2CodecUtil.connectionPrefaceBuf();
```
Scottmitch: `Http2CodecUtil.connectionPrefaceBuf()` will return a retained slice ... it seems like the pipeline is not hooked up correctly because the reference count doesn't get decremented at the end of this test ... lets re-evaluate once the `spy` calls have been removed.

codec-http2/src/main/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandler.java
```diff
@@ -0,0 +1,100 @@
...
+            if (!ByteBufUtil.equals(CONNECTION_PREFACE, CONNECTION_PREFACE.readerIndex(),
+                                    in, in.readerIndex(), bytesRead)) {
+                ctx.pipeline().remove(this);
+            } else if (bytesRead == prefaceLength) {
+                // Full h2 preface match, removed source codec, using http2 codec to handle
+                // following network traffic
+                ctx.pipeline().remove(httpServerCodec);
+                ctx.pipeline().remove(httpServerUpgradeHandler);
+
+                ctx.pipeline().addAfter(ctx.name(), null, http2ServerHandler);
```
Scottmitch: `addAfter` -> `replace(this, null, http2ServerHandler);` ?

codec-http2/src/main/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandler.java
chhsiao90: Done by https://github.com/netty/netty/pull/6496/commits/20bee40963957cb4a4bd1aa114699a48bc53e0c4 (I forgot this one as other reviewers had point out this problem...)

codec-http2/src/main/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandler.java
chhsiao90: Done by https://github.com/netty/netty/pull/6496/commits/20bee40963957cb4a4bd1aa114699a48bc53e0c4

codec-http2/src/main/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandler.java
chhsiao90: Done by https://github.com/netty/netty/pull/6496/commits/20bee40963957cb4a4bd1aa114699a48bc53e0c4

codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
chhsiao90: Wow, you point out some problem that I hadn't noticed!
It seems that unreleaseable ByteBuf can't never be released... 

Minimun unit test
```
@Test
    public void prefaceRefCount() {
        ByteBuf preface = Http2CodecUtil.connectionPrefaceBuf();
        int prevCount = preface.refCnt();

        preface.release();
        assertEquals(prevCount - 1, preface.refCnt());
    }
```

So there is also encounter same problem at [Http2ConnectionHandler.java#L296](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java#L296) that ByteBuf never release reference count.

Is this an issue ? Should I create another issue for this ?

codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
chhsiao90: OK, I will remove ```spy```!

codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
chhsiao90: OK, I will remove ```spy```!

chhsiao90: Addressd by https://github.com/netty/netty/pull/6496/commits/1071905d0a9a766ad609472bfcc1e4758488f33f

Without ```spy```, we need investigate the h2 connection state to make sure all things work.
For example, we can check stream(id=1) is half closed when an h2c upgrade,  and the connection settings for both situation.

codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
chhsiao90: Addressd by https://github.com/netty/netty/pull/6496/commits/1071905d0a9a766ad609472bfcc1e4758488f33f

chhsiao90:
Hi,

Not to push, just want to let you know that I'm still waiting for some advise at [comment#1](https://github.com/netty/netty/pull/6496#discussion_r105149470) and [comment#2](https://github.com/netty/netty/pull/6496#discussion_r105323785) so that I can move on this PR.

Please take your time for this, just a reminder, thank you for your time to review this! 😆

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
normanmaurer: @mosesn @Scottmitch WDYT ?

codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
normanmaurer: I think this is fine in this case.

mosesn:
@chhsiao90 sorry, things have been crazy for me.  will get to it tomorrow!  I promise.  please nag me if I don't.  `@mnnakamura` on twitter, or just update this thread.

chhsiao90:
No problem, just take your time 😄

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
mosesn: Hmm, this is a sort of sticky situation.  I think for now you should add a new event, but in the long run it would be nice to make the events a bit more consistent.  The client-side events are totally different from the server-side events, and both kinds of events are difficult to extend.

mosesn:
LGTM except for the event.

codec-http2/src/main/java/io/netty/handler/codec/http2/ClearTextHttp2ServerUpgradeHandler.java
chhsiao90: Done by https://github.com/netty/netty/pull/6496/commits/9ff1c7fd0d06c7b58e9db1c49fbb119cc22d0781 - If the name is not ok or it need to move out as an independent class file, please kindly let me know, thanks!

mosesn:
How about we call it PriorKnowledge.UpgradeEvent?

codec-http2/src/main/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandler.java
```diff
@@ -0,0 +1,106 @@
...
+     * @param httpServerUpgradeHandler the http server upgrade handler for HTTP/2
+     * @param http2ServerHandler the http2 server handler, will be added into pipeline
+     *                           when starting HTTP/2 by prior knowledge
+     */
+    public CleartextHttp2ServerUpgradeHandler(HttpServerCodec httpServerCodec,
+                                              HttpServerUpgradeHandler httpServerUpgradeHandler,
+                                              ChannelHandler http2ServerHandler) {
+        this.httpServerCodec = httpServerCodec;
+        this.httpServerUpgradeHandler = httpServerUpgradeHandler;
+        this.http2ServerHandler = http2ServerHandler;
```
normanmaurer: @chhsiao90 can you please add null check via `ObjectUtil.checkNonNull(...)`

chhsiao90:
@mosesn -
Sorry I'm not sure is that ```PriorKnowledge.UpgradeEvent``` means ```PriorKnowledgeUpgradeEvent``` , currently we don't have a class or an enum named PriorKnowledge.

codec-http2/src/main/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandler.java
chhsiao90: Sure! https://github.com/netty/netty/pull/6496/commits/6c0349e59aa864975f2029acd1dad9d4521ff334

mosesn:
oh, I see, I misread and was thinking of it in comparison with INSTANCE.  Yeah, maybe PriorKnowledgeUpgradeEvent would be good?

chhsiao90:
Sure! Had renamed the event name 😄

normanmaurer:
@Scottmitch @nmittler PTAL as well again.

codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
Scottmitch: I think we have an issue here (not related to this PR). I will submit a followup PR

codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
```diff
@@ -0,0 +1,196 @@
...
+        Object firstInbound = channel.readInbound();
+        assertTrue(firstInbound instanceof HttpRequest);
+        HttpRequest request = (HttpRequest) firstInbound;
+        assertEquals(HttpMethod.GET, request.method());
+        assertEquals("/", request.uri());
+        assertEquals(HttpVersion.HTTP_1_1, request.protocolVersion());
+        assertEquals(new DefaultHttpHeaders().add("Host", "example.com"), request.headers());
+
+        Object secondInbound = channel.readInbound();
+        assertTrue(secondInbound instanceof LastHttpContent);
```
Scottmitch: Please release this object. You can kill 2 birds with 1 stone here:

```java
((LastHttpContent) secondInbound).release();
```

codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
Scottmitch: see https://github.com/netty/netty/pull/6586

codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
chhsiao90: Done!

chhsiao90:
Thanks for the review!
I had addressed the last comment and rebase/squash the commit.
And all tests had passed after rebase.

normanmaurer:
@Scottmitch Please merge once you had a final look.

Scottmitch:
4.1 (0ee36fe)

Thanks @chhsiao90 !

Issue #5430, opened 2016-06-20
TODO-List for next major version


normanmaurer:
This issue is a TODO list for things to do for next major version.

Scottmitch:
- drop support for JDKs older than jdk8 (or w/e the latest supported JDK version is at the time)
- combine `ChannelInboundHandler` and `ChannelOutboundHandler`
- Investigate classes which override finalizer and see if we can make them implement `ReferenceCounted` instead (e.g. `OpenSslEngine`).
- Fix generics for the `Promise` related interfaces (e.g. `GenericFutureListener`).
- Investigate `DefaultPromise` dropping support for the executor changing. This complicates things for a case that is not common in practice.
- codec-http's `HttpHeaders` interface should directly extend `Headers`

Scottmitch:
- ChannelPipeline methods to return `Future` (of some kind) instead of `this`. The current interface requires careful coordination/synchronization for implementations of ChannelPipeline to ensure operations occur before the method returns. Returning a `Future` can allow users to achieve any ordering required if modifying the pipeline from multiple threads by adding listeners to the `Future`, and potentially avoid explicit synchronization internally.

johnou:
CompletableFuture > Future if possible.

Scottmitch:
Make ByteBuf and CompositeByteBuf an interface. We can use `default` methods for future extensibiility after API is finalized.

Scottmitch:
@johnou - We may be able to offer CompletableFuture/CompletionStage conversion routines, but using these in the core may prove to be difficult because of Netty's threading model has certain restrictions. Also `CompletionStage`'s API is quite large and can lead to trouble in my experience.

Scottmitch:
- Revisit the "void promise" concept, implementation, and APIs. A void promise requires special treatment at multiple layers in order to use correctly, and is often not accounted for at all layers. We should re-analyze the value of "void promise", and if similar goals can be achieved which doesn't require special treatment throughout the code.

johnou:
Make NioEventLoopGroup final? 😃

slandelle:
Remove JZlib based encoder/decoder and jzlib optional dependency. IMHO, those should be flagged as deprecated in Netty 4.1.

Scottmitch:
Enable hostname verification by default

johnou:
Remove deprecated methods / classes.

[...]

PR #6498, opened 2017-03-06
Make NioEventLoop group not final

rockwotj:
Motivation: I want to be able to subclass `NioEventLoop` so I can slightly tweak behavior or add functionality such as timing how long each event in the EventLoop takes.

Modification: Remove the `final` modifier from `NioEventLoop`

### DIFF
--- a/transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
+++ b/transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
@@ -53,7 +53,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
  * {@link Selector} and so does the multi-plexing of these in the event loop.
  *
  */
-public final class NioEventLoop extends SingleThreadEventLoop {
+public class NioEventLoop extends SingleThreadEventLoop {
 
     private static final InternalLogger logger = InternalLoggerFactory.getInstance(NioEventLoop.class);
 


normanmaurer:
@rockwotj I would love to not make it non final honestly but better understand what exactly want to do and come up with a solution which will work for other eventloops as well. Could you show me how you planed to extend it ?

rockwotj:
@normanmaurer totally. The whole idea is I really want to know if there are tasks that are slow and locking up the event loop, since the netty version we have pins a channel to a particular `EventLoop`. I don't know if that's changed in a newer version, as I've seen plans to do so.

This is what I'd like to do (in scala):

```scala
class MyEventLoop(parent: NioEventLoopGroup, factory: ThreadFactory, selector: SelectorProvider)
  extends NioEventLoop(parent, factory, selector) {
  override protected def newTaskQueue(): util.Queue[Runnable] = new TaskTimingQueue
}

private class TaskTimingQueue extends ConcurrentLinkedQueue[Runnable] {
  override def add(task: Runnable): Boolean = {
    super.add(new WrappedTimedTask(task))
  }
}

private class WrappedTimedTask(underlying: Runnable) extends Runnable {
  override def run(): Unit = NettyTimingInfo.timeTask(underlying.run())
}

// Use this as my childGroup
class MyEventLoopGroup(nThreads: Int, threadFactory: ThreadFactory)
  extends NioEventLoopGroup(nThreads, threadFactory) {
  override def newChild(threadFactory: ThreadFactory, args: AnyRef*): EventExecutor = {
    new FbEventLoop(this, threadFactory, args(0).asInstanceOf[SelectorProvider])
  }
}
```

Currently I can't do this at all and classes such as `AbstractNioChannel` check to see if the [event loop is an instance](https://github.com/netty/netty/blob/HEAD/transport/src/main/java/io/netty/channel/nio/AbstractNioChannel.java#L379) of `NioEventLoop`, so I can't provide an wrapped implementation if I wanted to try and do this either.

Do you have better ideas for something like this? I feel very restricted by netty to add anything such as this.

johnou:
@rockwotj how about at the event loop group thread factory level? I implemented something similar which essentially marks Netty threads as non blocking IO threads and if a DAO call tries to use one it blocks the invocation..

```java
bossGroup = new EpollEventLoopGroup(1, new NonBlockingIOThreadFactory("Epoll"));
workerGroup = new EpollEventLoopGroup(0, new NonBlockingIOThreadFactory("Epoll"));
bootstrap.group(bossGroup, workerGroup).channel(EpollServerSocketChannel.class);
```

```java
public class NonBlockingIOThreadFactory extends DefaultThreadFactory {
    
    public NonBlockingIOThreadFactory(String poolName) {
        super(poolName);
    }
    
    @Override
    public Thread newThread(Runnable r) {
        return super.newThread(() -> {
            BlockingIODetector.preventBlockingIOinThisThread();
            r.run();
        });
    }
    
    @Override
    protected Thread newThread(Runnable r, String name) {
        return super.newThread(() -> {
            BlockingIODetector.preventBlockingIOinThisThread();
            r.run();
        }, name);
    }
}
```

rockwotj:
I can't do that for the `NioEventLoopGroup` because there is only a single thread that is executed, I want to know how long each task takes on the EventLoop, that solution doesn't quite work for me.

Scottmitch:
This has come up a few times [1][2] and we have found more preferred alternatives in both situations. Will the wrapping approach described in https://github.com/netty/netty/pull/6441#issuecomment-281994334 work for you?

[1] https://github.com/netty/netty/pull/6441
[2] https://github.com/netty/netty/pull/4819#issuecomment-179826173

rockwotj:
@Scottmitch like I said before, I cannot do that, because there are places that expect the concrete implementation of `NioEventLoop` ie [here](https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/nio/AbstractNioChannel.java#L119) and [here](https://github.com/netty/netty/blob/795f318c3c11ec0520e7acd963ad4b310c287c20/transport/src/main/java/io/netty/channel/nio/AbstractNioChannel.java#L379)

Scottmitch:
Agreed there are limitations with the current APIs and assumptions made in the implementations (as discussed in the referenced issues). However as long as the Netty code sees the unwrapped objects you can work around these limitations. As you can see these classes / interfaces were not designed with wrapping / extension in mind and making these classes non-final will likely expose other extensibility issues. I put a note in https://github.com/netty/netty/issues/5430#issuecomment-284551318 to improve this scenario in the next API breaking release but until then lets be sure we exhaust other options first.

rockwotj:
For this specific use case, I can use a dirty reflection hack. But it makes me very sad.

```scala
package io.netty.channels.nio

class FbEventLoopGroup(nThreads: Int, threadFactory: ThreadFactory)
  extends NioEventLoopGroup(nThreads, threadFactory) {

  override def newChild(threadFactory: ThreadFactory, args: AnyRef*): EventExecutor = {
    val eventLoop = new NioEventLoop(this, threadFactory, args(0).asInstanceOf[SelectorProvider])
    // Dirty, dirty hack
    val eventLoopQueue = classOf[SingleThreadEventExecutor].getDeclaredField("taskQueue")
    eventLoopQueue.setAccessible(true)
    val currentQueue = eventLoopQueue.get(eventLoop).asInstanceOf[JQueue[Runnable]]
    eventLoopQueue.set(eventLoop, new TaskTimingQueue(currentQueue))
    eventLoop
  }
}

// Guava ForwardingQueue
private class TaskTimingQueue(override val delegate: JQueue[Runnable]) extends ForwardingQueue[Runnable] {
  override def add(task: Runnable): Boolean = {
    super.add(new WrappedTimedTask(task))
  }

  override def offer(task: Runnable): Boolean = {
    super.offer(new WrappedTimedTask(task))
  }
}

private class WrappedTimedTask(underlying: Runnable) extends Runnable {
  override def run(): Unit = NettyTimingInfo.timeTask { underlying.run() }
}
```

Scottmitch:
> I can use a dirty reflection hack. But it makes me very sad.

Me too 😢 

To clarify you want to collect metrics on everything submitted to `EventLoop` queue, not just your application stuff?  Instrumenting `Runnable`s in combination with monitoring the queue lengths over queue drain operations (been meaning to add this) may generally useful for visibility. Not the best extensibility story, but would adding support for these use cases be "good enough" for now?

rockwotj:
So I basically want percentiles of how long tasks take in the `EventLoop` queue. We sometimes have tasks that take a long time (and block other channel's traffic) and we want visibility into when this happens.

That would seem "good enough" for my specific use case.

Scottmitch:
ok sounds good @rockwotj ... I may not have cycles to get to this for a few weeks so maybe use the reflection hack for now if you need something urgent.

Scottmitch:
see https://github.com/netty/netty/issues/6524

rockwotj:
Thanks @Scottmitch!

PR #6500, opened 2017-03-06
Minor doc updates

jasobrown:
Motivation:
    
    Make the FileRegion comments about which transports are supported more accurate.
    Also, eleminate any outstanding references to FileRegion.transfered as the method was renamed for spelling.
    
    Modifications:
    
    Class-level comment on FileRegion, can call renamed method.
    
    Result:
    
    More accurate documentation and less calls to deprecated methods.


Scottmitch:
4.1 (3861b7d) 4.0 (21457d8)

Scottmitch:
thanks @jasobrown !

PR #6501, opened 2017-03-06
OpenSsl tests incomplete check for supporting key manager

Scottmitch:
Motivaiton:
It is possible that if the OpenSSL library supports the interfaces required to use the KeyManagerFactory, but we fail to get the io.netty.handler.ssl.openssl.useKeyManagerFactory system property (or this property is set to false) that SSLEngineTest based unit tests which use a KeyManagerFactory will fail.

Modifications:
- We should check if the OpenSSL library supports the KeyManagerFactory interfaces and if the system property allows them to be used in OpenSslEngineTests

Result:
Unit tests which use OpenSSL and KeyManagerFactory will be skipped instead of failing.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
+++ b/handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
@@ -144,13 +144,17 @@ public final class OpenSsl {
                             certBio = ReferenceCountedOpenSslContext.toBIO(cert.cert());
                             SSL.setCertificateChainBio(ssl, certBio, false);
                             supportsKeyManagerFactory = true;
-                            useKeyManagerFactory = AccessController.doPrivileged(new PrivilegedAction<Boolean>() {
-                                @Override
-                                public Boolean run() {
-                                    return SystemPropertyUtil.getBoolean(
-                                            "io.netty.handler.ssl.openssl.useKeyManagerFactory", true);
-                                }
-                            });
+                            try {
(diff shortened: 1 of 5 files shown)


handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
```diff
@@ -144,13 +144,17 @@
...
                             SSL.setCertificateChainBio(ssl, certBio, false);
                             supportsKeyManagerFactory = true;
-                            useKeyManagerFactory = AccessController.doPrivileged(new PrivilegedAction<Boolean>() {
-                                @Override
-                                public Boolean run() {
-                                    return SystemPropertyUtil.getBoolean(
-                                            "io.netty.handler.ssl.openssl.useKeyManagerFactory", true);
-                                }
-                            });
+                            try {
```
netkins: MAJOR Refactor this code to not nest more than 5 if/for/while/switch/try statements. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS134)
MAJOR Extract this nested try block into a separate method. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1141)

normanmaurer:
Cherry-picked into 4.1 (6bb661302f45173a381c0cab5be7b64710d188f8) and 4.0 (ed57a10d96191cbf821e3ecccdb3e0e4fb3cabf8)

normanmaurer:
@Scottmitch thanks a lot buddy!

PR #6504, opened 2017-03-07
Use system property to detect if root is running the program

normanmaurer:
Motivation:

We forked a new process to detect if the program is run by root. We should better just use user.name system property

Modifications:

- Change PlatformDependent.isRoot0() to read the user.name system property to detect if root runs the program
- Directly init IS_ROOT in the static block

Result:

Less heavy way to detect if the program is run by root.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent.java
@@ -15,7 +15,6 @@
  */
 package io.netty.util.internal;
 
-import io.netty.util.CharsetUtil;
 import io.netty.util.internal.logging.InternalLogger;
 import io.netty.util.internal.logging.InternalLoggerFactory;
 import org.jctools.queues.MpscArrayQueue;
@@ -27,13 +26,8 @@ import org.jctools.queues.atomic.SpscLinkedAtomicQueue;
 import org.jctools.util.Pow2;
 import org.jctools.util.UnsafeAccess;
 
-import java.io.BufferedReader;
 import java.io.File;
-import java.io.IOException;
-import java.io.InputStreamReader;
 import java.lang.reflect.Method;
-import java.net.InetSocketAddress;
-import java.net.ServerSocket;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.security.AccessController;
@@ -78,7 +72,7 @@ public final class PlatformDependent {
 
     private static final boolean IS_ANDROID = isAndroid0();
     private static final boolean IS_WINDOWS = isWindows0();
-    private static volatile Boolean IS_ROOT;
+    private static final boolean MAYBE_SUPER_USER;
 
     private static final int JAVA_VERSION = javaVersion0();
 
@@ -162,6 +156,8 @@ public final class PlatformDependent {
         }
         DIRECT_MEMORY_LIMIT = maxDirectMemory;
         logger.debug("io.netty.maxDirectMemory: {} bytes", maxDirectMemory);
+
+        MAYBE_SUPER_USER = maybeSuperUser0();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -921,92 +910,19 @@ private static boolean isRoot0() {
...
-                    }
+        try {
+            final String username = AccessController.doPrivileged(new PrivilegedAction<String>() {
+                @Override
+                public String run() {
+                    return System.getProperty("user.name");
                 }
-            }
+            });
+            // Check for root and toor as some BSDs have a toor user that is basically the same as root.
```
johnou: https://www.freebsd.org/doc/faq/security.html#idp60124648 wow :)

normanmaurer: Yes... Still remember the times when I used freebsd :)

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -921,92 +910,19 @@ private static boolean isRoot0() {
...
+            return "root".equals(username) || "toor".equals(username);
```
Scottmitch: seems like this could be fooled into thinking you have root privileges root when you do not. what if someone has a user name `toor`? Is there a better alternative to this?

normanmaurer: I think we could check if its freebsd and if not not check for toor ? That said its only used for logging when setting "broadcast" fails. So not a big deal... We could rename it to `isMaybeRoot()` to make it more clear that even if it returns something "wrong" its not such a big deal ?

Scottmitch: `maybeRoot()` with comments explaining the limitations of the method sounds better.

fenik17: `isRootOrToor()` :smile:

normanmaurer: Or maybe 'maybeSuperUser()' ?

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -916,97 +905,23 @@ private static boolean isWindows0() {
...
-            } finally {
-                if (ss != null) {
-                    try {
-                        ss.close();
-                    } catch (Exception e) {
-                        // Ignore.
-                    }
+    private static boolean maybeSuperUser0() {
+        try {
+            final String username = AccessController.doPrivileged(new PrivilegedAction<String>() {
```
Scottmitch: just use `SystemPropertyUtil.get("user.name");` instead?

normanmaurer: +1... Anything else ?

Scottmitch: last comment. we shouldn't have to catch anything in this method after this change, right?

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -916,97 +905,23 @@ private static boolean isWindows0() {
...
+                @Override
+                public String run() {
+                    return System.getProperty("user.name");
                 }
+            });
+            if (isWindows()) {
+                return "Administrator".equals(username);
```
fenik17: For Windows exist better way: http://tech.chitgoks.com/2010/10/21/detect-if-windows-user-has-admin-rights-using-java/

common/src/main/java/io/netty/util/internal/PlatformDependent.java
normanmaurer: correct

common/src/main/java/io/netty/util/internal/PlatformDependent.java
normanmaurer: com.sun is better to be avoided :(

fenik17: it is truth too )

normanmaurer:
Cherry-picked into 4.1 (9ade81ab5b56fcd0574688dd899dc7c07685caf7) and 4.0 (242c040)

PR #6510, opened 2017-03-07
Support for handling SSL and non-SSL in pipeline

### LINKED ISSUE #6428: SSL Handshake failure closes connection disallowing a response

johnou:
Motivation:

Some pipelines require support for both SSL and non-SSL messaging.

Modifications:

Add utility decoder to support both SSL and non-SSL handlers based on the initial message.

Result:

Less boilerplate code to write for developers.

Fixes #6428.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.ssl;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufAllocator;
+import io.netty.channel.ChannelHandler;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.ByteToMessageDecoder;
+import io.netty.util.ReferenceCountUtil;
+import io.netty.util.internal.ObjectUtil;
+
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLParameters;
+import java.util.List;
+
+/**
+ * {@link OptionalSslHandler} is a utility decoder to support both SSL and non-SSL handlers
+ * based on the first message received.
+ */
+public class OptionalSslHandler extends ByteToMessageDecoder {
+
+    private final SslContext sslContext;
+
+    public OptionalSslHandler(SslContext sslContext) {
+        this.sslContext = ObjectUtil.checkNotNull(sslContext, "sslContext");
+    }
+
+    @Override
+    protected void decode(ChannelHandlerContext context, ByteBuf in, List<Object> out) throws Exception {
+        if (in.readableBytes() < SslUtils.SSL_RECORD_HEADER_LENGTH) {
+            return;
+        }
+        if (SslHandler.isEncrypted(in)) {
+            handleSsl(context);
+        } else {
+            handleNonSsl(context);
+        }
+    }
+
+    private void handleSsl(ChannelHandlerContext context) {
+        SslHandler sslHandler = null;
+        try {
+            sslHandler = newSslHandler(context, sslContext);
+            context.pipeline().replace(this, newSslHandlerName(), sslHandler);
+            sslHandler = null;
+        } finally {
+            // Since the SslHandler was not inserted into the pipeline the ownership of the SSLEngine was not
+            // transferred to the SslHandler.
+            if (sslHandler != null) {
+                ReferenceCountUtil.safeRelease(sslHandler.engine());
+            }
+        }
+    }
+
+    private void handleNonSsl(ChannelHandlerContext context) {
+        ChannelHandler handler = newNonSslHandler(context);
+        if (handler != null) {
+            context.pipeline().replace(this, newNonSslHandlerName(), handler);
+        } else {
+            context.pipeline().remove(this);
+        }
+    }
+
+    /**
+     * Optionally specify the SSL handler name, this method may return {@code null}.
+     * @return the name of the SSL handler.
+     */
+    protected String newSslHandlerName() {
+        return null;
+    }
+
+    /**
+     * Override to configure the SslHandler eg. {@link SSLParameters#setEndpointIdentificationAlgorithm(String)}.
+     * The hostname and port is not known by this method so servers may want to override this method and use the
+     * {@link SslContext#newHandler(ByteBufAllocator, String, int)} variant.
+     *
+     * @param context the {@link ChannelHandlerContext} to use.
+     * @param sslContext the {@link SSLContext} to use.
+     * @return the {@link SslHandler} which will replace the {@link OptionalSslHandler} in the pipeline if the
+     * traffic is SSL.
+     */
+    protected SslHandler newSslHandler(ChannelHandlerContext context, SslContext sslContext) {
+        return sslContext.newHandler(context.alloc());
+    }
+
+    /**
+     * Optionally specify the non-SSL handler name, this method may return {@code null}.
+     * @return the name of the non-SSL handler.
+     */
+    protected String newNonSslHandlerName() {
+        return null;
+    }
+
+    /**
+     * Override to configure the ChannelHandler.
+     * @param context the {@link ChannelHandlerContext} to use.
+     * @return the {@link ChannelHandler} which will replace the {@link OptionalSslHandler} in the pipeline
+     * or {@code null} to simply remove the {@link OptionalSslHandler} if the traffic is non-SSL.
+     */
+    protected ChannelHandler newNonSslHandler(ChannelHandlerContext context) {
+        return null;
+    }
+}
--- a/handler/src/test/java/io/netty/handler/ssl/OptionalSslHandlerTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/OptionalSslHandlerTest.java
@@ -1,0 +1,121 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.ssl;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import io.netty.channel.ChannelHandler;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelPipeline;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.junit.MockitoJUnitRunner;
+
+import static org.mockito.Mockito.*;
+
+@RunWith(MockitoJUnitRunner.class)
+public class OptionalSslHandlerTest {
+
+    private static final String SSL_HANDLER_NAME = "sslhandler";
+    private static final String HANDLER_NAME = "handler";
+
+    @Mock
+    private ChannelHandlerContext context;
+
+    @Mock
+    private SslContext sslContext;
+
+    @Mock
+    private ChannelPipeline pipeline;
+
+    @Before
+    public void setUp() throws Exception {
+        when(context.pipeline()).thenReturn(pipeline);
+    }
+
+    @Test
+    public void handlerRemoved() throws Exception {
+        OptionalSslHandler handler = new OptionalSslHandler(sslContext);
+        final ByteBuf payload = Unpooled.copiedBuffer("plaintext".getBytes());
+        try {
+            handler.decode(context, payload, null);
+            verify(pipeline).remove(handler);
+        } finally {
+            payload.release();
+        }
+    }
+
+    @Test
+    public void handlerReplaced() throws Exception {
+        final ChannelHandler nonSslHandler = Mockito.mock(ChannelHandler.class);
+        OptionalSslHandler handler = new OptionalSslHandler(sslContext) {
+            @Override
+            protected ChannelHandler newNonSslHandler(ChannelHandlerContext context) {
+                return nonSslHandler;
+            }
+
+            @Override
+            protected String newNonSslHandlerName() {
+                return HANDLER_NAME;
+            }
+        };
+        final ByteBuf payload = Unpooled.copiedBuffer("plaintext".getBytes());
+        try {
+            handler.decode(context, payload, null);
+            verify(pipeline).replace(handler, HANDLER_NAME, nonSslHandler);
+        } finally {
+            payload.release();
+        }
+    }
+
+    @Test
+    public void sslHandlerReplaced() throws Exception {
+        final SslHandler sslHandler = Mockito.mock(SslHandler.class);
+        OptionalSslHandler handler = new OptionalSslHandler(sslContext) {
+            @Override
+            protected SslHandler newSslHandler(ChannelHandlerContext context, SslContext sslContext) {
+                return sslHandler;
+            }
+
+            @Override
+            protected String newSslHandlerName() {
+                return SSL_HANDLER_NAME;
+            }
+        };
+        final ByteBuf payload = Unpooled.wrappedBuffer(new byte[] { 22, 3, 1, 0, 5 });
+        try {
+            handler.decode(context, payload, null);
+            verify(pipeline).replace(handler, SSL_HANDLER_NAME, sslHandler);
+        } finally {
+            payload.release();
+        }
+    }
+
+    @Test
+    public void decodeBuffered() throws Exception {
+        OptionalSslHandler handler = new OptionalSslHandler(sslContext);
+        final ByteBuf payload = Unpooled.wrappedBuffer(new byte[] { 22, 3 });
+        try {
+            handler.decode(context, payload, null);
+            verifyZeroInteractions(pipeline);
+        } finally {
+            payload.release();
+        }
+    }
+}


johnou:
@dlmarion PTAL

handler/src/test/java/io/netty/handler/ssl/OptionalSslHandlerTest.java
```diff
@@ -0,0 +1,98 @@
...
+            @Override
+            protected SslHandler sslHandler(SslContext sslContext, ChannelHandlerContext context) {
+                return sslHandler;
+            }
+        };
+        final ByteBuf payload = Unpooled.wrappedBuffer(new byte[] { 22, 3, 1, 0, 5 });
+        try {
+            handler.decode(context, payload, null);
+            verify(pipeline).replace(handler, HANDLER_NAME, sslHandler);
+        } finally {
```
fenik17: IMHO, try-finally is not necessary in the tests :)

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.List;
+
+/**
+ * {@link OptionalSslHandler} is a utility decoder to support both SSL and non-SSL handlers
+ * based on the initial message.
+ */
+public class OptionalSslHandler extends ByteToMessageDecoder {
+
+    private static final int SSL_RECORD_HEADER_LENGTH = 5;
+
```
normanmaurer: You can just reuse what is defined in SslUtils ?

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,94 @@
...
+    private final SslContext sslContext;
+    private final String handlerName;
+
+    public OptionalSslHandler(SslContext sslContext, String handlerName) {
+        this.sslContext = sslContext;
+        this.handlerName = handlerName;
```
normanmaurer: Null checks

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,94 @@
...
+        }
+    }
+
+    /**
+     * Override to configure the SslHandler eg. enable HttpsEndpointIdentificationAlgorithm.
+     * @param sslContext the {@link SSLContext} to use.
+     * @param context the {@link ChannelHandlerContext} to use.
+     * @return the {@link SslHandler} to replace this handler with.
+     */
+    protected SslHandler sslHandler(SslContext sslContext, ChannelHandlerContext context) {
```
normanmaurer: newSslHandler

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,94 @@
...
+        return sslContext.newHandler(context.alloc());
+    }
+
+    /**
+     * Override to configure the ChannelHandler.
+     * @param context the {@link ChannelHandlerContext} to use.
+     * @return the {@link ChannelHandler} to replace this handler with or null to remove.
+     */
+    protected ChannelHandler nonSslHandler(ChannelHandlerContext context) {
```
normanmaurer: NewNonSslHandler

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,94 @@
...
+        return null;
+    }
+
+    @Override
+    public boolean isSingleDecode() {
```
normanmaurer: final

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,94 @@
...
+
+    private void handleSsl(ChannelHandlerContext context) {
+        ChannelPipeline pipeline = context.pipeline();
+        pipeline.replace(this, handlerName, sslHandler(sslContext, context));
+    }
+
+    private void handleNonSsl(ChannelHandlerContext context) {
+        ChannelHandler handler = nonSslHandler(context);
+        if (handler != null) {
+            context.pipeline().replace(this, handlerName, handler);
```
normanmaurer: I wonder if we should not better have separate handlerNames for ssl and non ssl

handler/src/test/java/io/netty/handler/ssl/OptionalSslHandlerTest.java
johnou: I'd rather keep it as is, verify may throw an exception if there is a regression and this pattern is used in other Ssl unit tests.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
johnou: sgtm

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
johnou: of course, originally wasn't in this package so it wasn't visible, done.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
johnou: done.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
johnou: done.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
johnou: done.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
johnou: done.

johnou:
@normanmaurer PTAL

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,95 @@
...
+    protected SslHandler newSslHandler(SslContext sslContext, ChannelHandlerContext context) {
+        return sslContext.newHandler(context.alloc());
+    }
+
+    /**
+     * Override to configure the ChannelHandler.
+     * @param context the {@link ChannelHandlerContext} to use.
+     * @return the {@link ChannelHandler} to replace this handler with or null to remove.
+     */
+    protected ChannelHandler newNonSslHandler(ChannelHandlerContext context) {
```
netkins: MAJOR Remove this unused method parameter "context". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,95 @@
...
+ * {@link OptionalSslHandler} is a utility decoder to support both SSL and non-SSL handlers
+ * based on the initial message.
+ */
+public class OptionalSslHandler extends ByteToMessageDecoder {
+
+    private final SslContext sslContext;
+    private final String sslHandlerName;
+    private final String nonSslHandlerName;
+
+    public OptionalSslHandler(SslContext sslContext, String sslHandlerName, String nonSslHandlerName) {
```
normanmaurer: @johnou it feels a bit weird that we require a `nonSslHandlerName` but by default we will not add any. Consider just remove this and add another `protected` methods that the user can override while by default it uses `null`.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,95 @@
...
+     * Override to configure the ChannelHandler.
+     * @param context the {@link ChannelHandlerContext} to use.
+     * @return the {@link ChannelHandler} to replace this handler with or null to remove.
+     */
+    protected ChannelHandler newNonSslHandler(ChannelHandlerContext context) {
+        return null;
+    }
+
+    @Override
+    public final boolean isSingleDecode() {
```
normanmaurer: @johnou actually this is not needed and it should just work as the handler is removed anyway.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,95 @@
...
+     * @param sslContext the {@link SSLContext} to use.
+     * @param context the {@link ChannelHandlerContext} to use.
+     * @return the {@link SslHandler} to replace this handler with.
+     */
+    protected SslHandler newSslHandler(SslContext sslContext, ChannelHandlerContext context) {
+        return sslContext.newHandler(context.alloc());
+    }
+
+    /**
+     * Override to configure the ChannelHandler.
```
normanmaurer: the ChannelHandler that should be added if the traffic is not ssl... or something like that.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,95 @@
...
+     * @param context the {@link ChannelHandlerContext} to use.
+     * @return the {@link ChannelHandler} to replace this handler with or null to remove.
```
normanmaurer: {@code null}

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+
+    /**
+     * Override to configure the SslHandler eg. enable HttpsEndpointIdentificationAlgorithm.
+     * @param sslContext the {@link SSLContext} to use.
+     * @param context the {@link ChannelHandlerContext} to use.
+     * @return the {@link SslHandler} to replace this handler with.
+     */
+    protected SslHandler newSslHandler(SslContext sslContext, ChannelHandlerContext context) {
```
normanmaurer: switch the arguments to be more consistent

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,95 @@
...
+ */
+public class OptionalSslHandler extends ByteToMessageDecoder {
+
+    private final SslContext sslContext;
+    private final String sslHandlerName;
+    private final String nonSslHandlerName;
+
+    public OptionalSslHandler(SslContext sslContext, String sslHandlerName, String nonSslHandlerName) {
+        this.sslContext = ObjectUtil.checkNotNull(sslContext, "sslContext");
+        this.sslHandlerName = ObjectUtil.checkNotNull(sslHandlerName, "sslHandlerName");
```
normanmaurer: actually sslHandlerName can be null. In this case a custom name will be used.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,95 @@
...
+        if (SslHandler.isEncrypted(in)) {
+            handleSsl(context);
+        } else {
+            handleNonSsl(context);
+        }
+    }
+
+    private void handleSsl(ChannelHandlerContext context) {
+        ChannelPipeline pipeline = context.pipeline();
+        pipeline.replace(this, sslHandlerName, newSslHandler(sslContext, context));
```
normanmaurer: just merge the two lines

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
normanmaurer: also consider adding a second constructor which will just take an `SslContext`

handler/src/test/java/io/netty/handler/ssl/OptionalSslHandlerTest.java
```diff
@@ -0,0 +1,99 @@
...
+            }
+        };
+        final ByteBuf payload = Unpooled.wrappedBuffer(new byte[] { 22, 3, 1, 0, 5 });
+        try {
+            handler.decode(context, payload, null);
+            verify(pipeline).replace(handler, SSL_HANDLER_NAME, sslHandler);
+        } finally {
+            payload.release();
+        }
+    }
```
normanmaurer: Also please add a test which verifies that it buffers until the full 5 bytes are received.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
Scottmitch: +1. instead of caring about names as member variables can we just make methods that return `null` by default (this will use a default name). For example:

```java
protected String newSslHandlerName(SslHandler handler) {
  return null;
}
```

Also keep in mind if the SslHandler is created ... but not inserted in the pipeline (user code throws) ... we must call `release()` on the engine. See https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/SniHandler.java#L359

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
johnou: done.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
johnou: done.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
johnou: done.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
johnou: does it hurt to leave it? consistent with HAProxy implementation.

normanmaurer: I think I would just remove it... and I will fix it for HAProxy as well

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
normanmaurer: +1 sounds good.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
johnou: done.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
johnou: sgtm.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
johnou: done.

johnou: 👍

handler/src/test/java/io/netty/handler/ssl/OptionalSslHandlerTest.java
johnou: ow io.netty.handler.codec.ByteToMessageDecoder#callDecode has protected access which takes care of the buffering, might need to setup a full pipeline and client.

normanmaurer: nah... I think it will be good enough to check that the handler is still in the pipeline if you pass a buffer with not enough readable bytes into decode(...) as it will return early.

dlmarion:
FWIW, I just ran my tests against the latest implementation with no issues.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * {@link OptionalSslHandler} is a utility decoder to support both SSL and non-SSL handlers
+ * based on the initial message.
+ */
+public class OptionalSslHandler extends ByteToMessageDecoder {
+
+    private final SslContext sslContext;
+
+    public OptionalSslHandler(SslContext sslContext) {
+        this.sslContext = sslContext;
```
normanmaurer: missing null check here.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,116 @@
...
+        ChannelHandler handler = newNonSslHandler(context);
+        if (handler != null) {
+            context.pipeline().replace(this, newNonSslHandlerName(), handler);
+        } else {
+            context.pipeline().remove(this);
+        }
+    }
+
+    /**
+     * Override to specify the SSL handler name.
```
normanmaurer: @johnou add comment that this is allowed to return `{@code null}`

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,116 @@
...
+     * @param context the {@link ChannelHandlerContext} to use.
+     * @param sslContext the {@link SSLContext} to use.
+     * @return the {@link SslHandler} to be added to the pipeline if the traffic is SSL.
+     */
+    protected SslHandler newSslHandler(ChannelHandlerContext context, SslContext sslContext) {
+        return sslContext.newHandler(context.alloc());
+    }
+
+    /**
+     * Override to specify the non-SSL handler name.
```
normanmaurer: @johnou add docs that its allowed to return {@code null}

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,116 @@
...
+     * @return the name of the non-SSL handler.
+     */
+    protected String newNonSslHandlerName() {
+        return null;
+    }
+
+    /**
+     * Override to configure the ChannelHandler.
+     * @param context the {@link ChannelHandlerContext} to use.
+     * @return the {@link ChannelHandler} to be added to the pipeline if the traffic is not SSL, {@code null} to remove.
```
normanmaurer: `@return the {@link ChannelHandler} to with which the {@link  OptionalSslHandler} will be replaced or {@code null} to just remove the {@link  OptionalSslHandler}.`

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
johnou: done.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
johnou: done.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
johnou: sgtm.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
johnou: got me.

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,111 @@
...
+    /**
+     * Optionally specify the SSL handler name, this method may return {@code null}.
+     * @return the name of the SSL handler.
+     */
+    protected String newSslHandlerName() {
+        return null;
+    }
+
+    /**
+     * Override to configure the SslHandler eg. enable HttpsEndpointIdentificationAlgorithm.
```
Scottmitch: Add a comment to the effect: `the hostname and port is not known by this method so servers may want to override this method and use the {@link SslHandler#newHandlerByteBufAllocator, string, int)} variant.`

Also replace `HttpsEndpointIdentificationAlgorithm` with `{@link SSLParamters#setEndpointIdentificationAlgorithm(String)}`

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,111 @@
...
+     */
+    protected String newNonSslHandlerName() {
+        return null;
+    }
+
+    /**
+     * Override to configure the ChannelHandler.
+     * @param context the {@link ChannelHandlerContext} to use.
+     * @return the {@link ChannelHandler} to with which the {@link OptionalSslHandler} will be replaced or {@code null}
+     * to just remove the {@link OptionalSslHandler}.
```
netkins: MAJOR Remove this unused method parameter "context". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,111 @@
...
+
+    private void handleSsl(ChannelHandlerContext context) {
+        SslHandler sslHandler = null;
+        try {
+            sslHandler = newSslHandler(context, sslContext);
+            context.pipeline().replace(this, newSslHandlerName(), sslHandler);
+            sslHandler = null;
+        } finally {
+            // Since the SslHandler was not inserted into the pipeline the ownership of the SSLEngine was not
+            // transferred to the SslHandler.
```
netkins: BLOCKER Change this condition so that it does not always evaluate to "false" [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2583)

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,111 @@
...
+        return null;
+    }
+
+    /**
+     * Override to configure the ChannelHandler.
+     * @param context the {@link ChannelHandlerContext} to use.
+     * @return the {@link ChannelHandler} to with which the {@link OptionalSslHandler} will be replaced or {@code null}
+     * to just remove the {@link OptionalSslHandler}.
+     */
+    protected ChannelHandler newNonSslHandler(ChannelHandlerContext context) {
```
netkins: MAJOR Remove this unused method parameter "context". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,111 @@
...
+    private void handleSsl(ChannelHandlerContext context) {
+        SslHandler sslHandler = null;
+        try {
+            sslHandler = newSslHandler(context, sslContext);
+            context.pipeline().replace(this, newSslHandlerName(), sslHandler);
+            sslHandler = null;
+        } finally {
+            // Since the SslHandler was not inserted into the pipeline the ownership of the SSLEngine was not
+            // transferred to the SslHandler.
+            if (sslHandler != null) {
```
netkins: BLOCKER Change this condition so that it does not always evaluate to "false" [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2583)

johnou:
@normanmaurer @Scottmitch thanks for the feedback, believe I have addressed all your comments.

normanmaurer:
I will let @Scottmitch take a final look before pulling in. @johnou nice work!

Scottmitch:
4.1 (cfebaa3) 4.0 (3fbdae3)

Thanks @johnou !

johnou:
Cheers!

handler/src/main/java/io/netty/handler/ssl/OptionalSslHandler.java
```diff
@@ -0,0 +1,117 @@
...
+        return null;
+    }
+
+    /**
+     * Override to configure the ChannelHandler.
+     * @param context the {@link ChannelHandlerContext} to use.
+     * @return the {@link ChannelHandler} which will replace the {@link OptionalSslHandler} in the pipeline
+     * or {@code null} to simply remove the {@link OptionalSslHandler} if the traffic is non-SSL.
+     */
+    protected ChannelHandler newNonSslHandler(ChannelHandlerContext context) {
```
netkins: MAJOR Remove this unused method parameter "context". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

dlmarion:
Thanks to all for helping get this resolved.

johnou:
@dlmarion feel free to buy me a beer next time I'm in your area.

normanmaurer:
Ah... beer driven development ;)

PR #6513, opened 2017-03-08
SslContext to support TLS/SSL protocols

Scottmitch:
Motivation:
SslContext and SslContextBuilder do not support a way to specify the desired TLS protocols. This currently requires that the user extracts the SSLEngine once a context is built and manually call SSLEngine#setEnabledProtocols(String[]). Something this critical should be supported at the SslContext level.

Modifications:
- SslContextBuilder should accept a list of protocols to configure for each SslEngine

Result:
SslContext consistently sets the supported TLS/SSL protocols.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/JdkSslContext.java
+++ b/handler/src/main/java/io/netty/handler/ssl/JdkSslContext.java
@@ -52,7 +52,7 @@ public class JdkSslContext extends SslContext {
     private static final InternalLogger logger = InternalLoggerFactory.getInstance(JdkSslContext.class);
 
     static final String PROTOCOL = "TLS";
-    static final String[] PROTOCOLS;
+    static final String[] DEFAULT_PROTOCOLS;
     static final List<String> DEFAULT_CIPHERS;
     static final Set<String> SUPPORTED_CIPHERS;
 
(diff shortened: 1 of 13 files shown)


handler/src/main/java/io/netty/handler/ssl/JdkSslContext.java
```diff
@@ -52,7 +52,7 @@
     private static final InternalLogger logger = InternalLoggerFactory.getInstance(JdkSslContext.class);
 
     static final String PROTOCOL = "TLS";
-    static final String[] PROTOCOLS;
+    static final String[] DEFAULT_PROTOCOLS;
```
netkins: CRITICAL Make this array "private". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1873)

normanmaurer:
Is the commit message incorrect as it mentions setEnabledCipherSuites? I think it should be setEnabledProtocols .

Scottmitch:
@normanmaurer - yip cpy/pasta mistake.

Scottmitch:
@normanmaurer - commit message fixed ... any other comments?

normanmaurer:
@Scottmitch nope... ship it

Scottmitch:
4.1 (a230428) 4.0 (3b997d8)

PR #6517, opened 2017-03-08
Add support for IBM J9 Http2 cipher names.

CodingFabian:
Motivation:
Http2SecurityUtil currently lists HTTP/2 ciphers as documented by
JSSE docs [1] and the IANA [2] using the TLS_ prefix.
In some IBM J9 implementations the SSL_ prefix is used, which is also
covered by the JSSE.

[1] http://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html
[2] http://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml

Modifications:
Add both variants of the cipher names (prefixed with SSL_ in additon to TLS_)

Result:
HTTP/2 connections can now be created using the SslProvider.JDK on IBM J9
and potentially other JVMs which use the SSL_ prefix.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2SecurityUtil.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2SecurityUtil.java
@@ -34,26 +34,51 @@ public final class Http2SecurityUtil {
      * href="https://wiki.mozilla.org/Security/Server_Side_TLS#Non-Backward_Compatible_Ciphersuite">Mozilla Cipher
      * Suites</a> in accordance with the <a
      * href="https://tools.ietf.org/html/draft-ietf-httpbis-http2-16#section-9.2.2">HTTP/2 Specification</a>.
+     *
+     * According to the <a href="http://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html">
+     * JSSE documentation</a> "the names mentioned in the TLS RFCs prefixed with TLS_ are functionally equivalent
+     * to the JSSE cipher suites prefixed with SSL_".
+     * Both variants are used to support JVMs supporting the one or the other.
      */
     public static final List<String> CIPHERS;
 
     private static final List<String> CIPHERS_JAVA_MOZILLA_INCREASED_SECURITY = Collections.unmodifiableList(Arrays
             .asList(
             /* Java 8 */
-            "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384", /* openssl = ECDHE-ECDSA-AES256-GCM-SHA384 */
-            "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256", /* openssl = ECDHE-ECDSA-AES128-GCM-SHA256 */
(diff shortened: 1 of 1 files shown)


CodingFabian:
i converted the end of line comments mentioning the openssl name to "headers" of the tuple of the ciphername with tls and ssl prefixx

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2SecurityUtil.java
```diff
@@ -34,26 +34,48 @@
      * href="https://wiki.mozilla.org/Security/Server_Side_TLS#Non-Backward_Compatible_Ciphersuite">Mozilla Cipher
      * Suites</a> in accordance with the <a
      * href="https://tools.ietf.org/html/draft-ietf-httpbis-http2-16#section-9.2.2">HTTP/2 Specification</a>.
+     * 
+     * On OpenJDK, the ciphers start with TLS_, while on IBM J9 they are named SSL_.
```
normanmaurer: @CodingFabian  nit: On OpenJDK and Oracle JDK ?

CodingFabian: Well i consider Oracle JDK to be OpenJDK, but I happily will adjust to your preferred wording.
But the TLS_ names work correctly on Azul Zulu, so we should add that, or say OpenJDK based JVMs?

normanmaurer: OpenJDK based sounds good

Scottmitch:
> Http2SecurityUtil currently lists Http 2 ciphers by their OpenJDK names

This is not true. They are the cipher names listed by the JSSE docs [1] and the IANA [2]. There are comments which list the OpenSSL variants for informational purposes.

Please add a link to the JSSE cipher names doc and the following text to the commit message [1]:

> The names mentioned in the TLS RFCs prefixed with TLS_ are functionally equivalent to the JSSE cipher suites prefixed with SSL_.

[1] http://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html
[2] http://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2SecurityUtil.java
Scottmitch: I think its better to just list that the JSSE allows for the "TLS" and "SSL" prefix to be interchanged [1]:

> The names mentioned in the TLS RFCs prefixed with TLS_ are functionally equivalent to the JSSE cipher suites prefixed with SSL_.

[1] http://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html

normanmaurer:
@Scottmitch good find ... @CodingFabian please do as @Scottmitch advised

CodingFabian:
@Scottmitch @normanmaurer done as requested, taking some of your proposed wording and added a bit of my own to both javadoc and commit message.

Sorry for not knowing the roots of the implementation, i just found out that its not working on J9 :)

normanmaurer:
Cherry-picked into 4.1 (a94b23df7d367ecb57ed89a9e882e2f47df59306)

Issue #6507, opened 2017-03-07
UnorderedThreadPoolEventExecutor consumes 100% CPU when idle


meshcow:
### Expected behavior
0% CPU consumption when idle
### Actual behavior
almost 100% CPU consumption when idle
### Steps to reproduce
After a submitted task has been successfully executed, all the available cores get busy with doing some stuff.
```java
import io.netty.util.concurrent.UnorderedThreadPoolEventExecutor;

public class TestUnorderedThreadPoolEventExecutor {

    public static void main(String[] args) {
        UnorderedThreadPoolEventExecutor executor = new UnorderedThreadPoolEventExecutor(1);
        // compare: ScheduledThreadPoolExecutor executor = new ScheduledThreadPoolExecutor(1);
        executor.submit(() -> {
            return "completed!";
        });
    }
}
```
### Netty version
```
<groupId>io.netty</groupId>
<artifactId>netty-all</artifactId>
<version>4.1.8.Final</version>
```
#### Linux
Linux 3.13.0-042stab120.3 x86_64 x86_64 x86_64 GNU/Linux
java version "1.8.0_112"
Java(TM) SE Runtime Environment (build 1.8.0_112-b15)
Java HotSpot(TM) 64-Bit Server VM (build 25.112-b15, mixed mode)
#### OSX
Darwin 15.6.0 Darwin Kernel Version 15.6.0: Mon Jan  9 23:07:29 PST 2017; root:xnu-3248.60.11.2.1~1/RELEASE_X86_64 x86_64
java version "1.8.0_121"
Java(TM) SE Runtime Environment (build 1.8.0_121-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.121-b13, mixed mode)

### `top -H` on Linux
```
top - 13:37:31 up 92 days, 47 min,  2 users,  load average: 0.57, 0.13, 0.04
Threads: 127 total,   3 running, 124 sleeping,   0 stopped,   0 zombie
%Cpu(s): 96.4 us,  0.5 sy,  0.0 ni,  3.1 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st
KiB Mem:   2097152 total,  2082604 used,    14548 free,        0 buffers
KiB Swap:        0 total,        0 used,        0 free.  1245400 cached Mem

  PID USER      PR  NI    VIRT    RES    SHR S %CPU %MEM     TIME+ COMMAND                                                                                                                                
[...]

PR #6518, opened 2017-03-08
UnorderedThreadPoolEventExecutor consumes 100% CPU when idle

normanmaurer:
Motivation:

When UnorderedThreadPoolEventExecutor.execute / submit etc is called it will consume up to 100 % CPU even after the task was executed.

Modifications:

Add a special wrapper which we will be used in execute(...) to wrap the submitted Runnable. This is needed as  ScheduledThreadPoolExecutor.execute(...) will delegate to submit(...) which will then use decorateTask(...). The problem with this is that decorateTask(...) needs to ensure we only do our own decoration if we not call from execute(...) as otherwise we may end up creating an endless loop because DefaultPromise will call  EventExecutor.execute(...) when notify the listeners of the promise.

Result:

Fixes [#6507].

### DIFF
--- a/common/src/main/java/io/netty/util/concurrent/UnorderedThreadPoolEventExecutor.java
+++ b/common/src/main/java/io/netty/util/concurrent/UnorderedThreadPoolEventExecutor.java
@@ -30,6 +30,8 @@ import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+
 /**
  * {@link EventExecutor} implementation which makes no guarantees about the ordering of task execution that
  * are submitted because there may be multiple threads executing these tasks.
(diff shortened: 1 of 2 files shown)


normanmaurer:
@Scottmitch PTAL... unfortunally there is really no way for a unit test :(

Scottmitch:
`otherwise we may end up to great` -> can you clarify what you mean by this?

normanmaurer:
@Scottmitch actually I have an idea for a unit test.... stay tuned

common/src/main/java/io/netty/util/concurrent/UnorderedThreadPoolEventExecutor.java
```diff
@@ -201,6 +204,11 @@ public void shutdown() {
         return (Future<T>) super.submit(task);
     }
 
+    @Override
+    public void execute(Runnable command) {
+        super.schedule(new NonNotifyRunnable(command), 0, TimeUnit.NANOSECONDS);
```
netkins: MAJOR Unnecessary use of fully qualified name 'TimeUnit.NANOSECONDS' due to existing static import 'java.util.concurrent.TimeUnit.NANOSECONDS' [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUnnecessaryFullyQualifiedName)

normanmaurer:
@Scottmitch yeah... could not fine a better solution for the test as without the loop it sometimes passed.

normanmaurer:
Cherry-picked into 4.1 (c6a3cae269f9305be7de088fe89500bb1137a2fe) and 4.0 (b32f3777eab63ced56e2895388d06a83deb2acfa)

PR #6527, opened 2017-03-09
Use SystemPropertyUtil to access system properties

normanmaurer:
Motivation:

We should use SystemPropertyUtil to access system properties and so always handle SecurityExceptions.

Modifications:

Use SystemPropertyUtil everywhere.

Result:

Better and consist handling of SecurityException.


normanmaurer:
@jasontedor PTAL as well

normanmaurer:
Cherry-picked into 4.1 (9e6e1a3e7b451779a0aa8db07271a7befd69e02c) and 4.0 (c6bc0b302105a5a6ecae49af8ad6926bbf1379ac)

PR #6529, opened 2017-03-09
Recognizing TLS Extension "HeartBeat" as Valid TLS Packet

kaja-mohideen:
Motivation:

The widely used SSL Implementation, OpenSSL, already supports Heartbeat Extension; both sending and responding to Heartbeat Messages. But, since Netty is not recognizing that extension as valid packet, peers won't be able to use this extension.

Modification:

Update SslUtils.java to recognize Heartbeat Extension as valid tls packet.

Result:

With this change, softwares using Netty + OpenSSL will be able to respond for TLS Heartbeat requests (actually taken care by OpenSSL - no need of any extra implementation from Clients)

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslUtils.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslUtils.java
@@ -182,6 +182,7 @@ final class SslUtils {
             case SSL_CONTENT_TYPE_ALERT:
             case SSL_CONTENT_TYPE_HANDSHAKE:
             case SSL_CONTENT_TYPE_APPLICATION_DATA:
+            case SSL_CONTENT_TYPE_EXTENSION_HEARTBEAT:
                 tls = true;
                 break;
             default:
--- a/handler/src/main/java/io/netty/handler/ssl/SslUtils.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslUtils.java
@@ -49,6 +49,11 @@ final class SslUtils {
     static final int SSL_CONTENT_TYPE_APPLICATION_DATA = 23;
 
     /**
+     * HeartBeat Extension
+     */
+    static final int SSL_CONTENT_TYPE_EXTENSION_HEARTBEAT = 24;
+
+    /**
      * the length of the ssl record header (in bytes)
      */
     static final int SSL_RECORD_HEADER_LENGTH = 5;
@@ -90,6 +95,7 @@ final class SslUtils {
             case SSL_CONTENT_TYPE_ALERT:
             case SSL_CONTENT_TYPE_HANDSHAKE:
             case SSL_CONTENT_TYPE_APPLICATION_DATA:
+            case SSL_CONTENT_TYPE_EXTENSION_HEARTBEAT:
                 tls = true;
                 break;
             default:


normanmaurer:
@kaja-mohideen thanks a lot... Can you sign the ICLA ?

https://netty.io/s/icla ?

kaja-mohideen:
I have just signed the CCLA at https://docs.google.com/forms/d/e/1FAIpQLSfXTK6SnWWFbR50DFhoZq2floCFOUBH3kG8sZP77im5Rknctg/viewform?formkey=dG9wTmhoeGNGd1MtdFdtVXl4TlVSNlE6MQ as suggested by my employer.

Please let me know if anything more needs to be done.

normanmaurer:
@kaja-mohideen ok sounds good... Let me wait for the CI to complete

normanmaurer:
+1... @kaja-mohideen please address @Scottmitch s comment and ping us once done

kaja-mohideen:
Thanks for reviewing @Scottmitch 

I have done the changes. Kindly verify.

@normanmaurer

johnou:
@normanmaurer doesn't this open the possibility of exploiting the heartbleed vulnerability if the system is running < OpenSSL 1.0.1g?

johnou:
That said maybe it's not such an issue now because heartbeat is disabled by default and @kaja-mohideen would need to modify the compile flags and package tcnative himself https://github.com/netty/netty-tcnative/commit/3e10806c98033a3b754e28fb69ca2dd4a0cfe1f1

johnou:
Maybe a non-issue, I just tried tcnative against 1.0.1f and compilation failed with "error LNK2001: unresolved external symbol SSL_add0_chain_cert" so running tcnative with OpenSSL 1.0.1f will probably result in linkage errors.

johnou:
@normanmaurer awkward silence?

normanmaurer:
@kaja-mohideen @johnou sorry I am currently traveling so had not much time to think about it. Will come back to you soon.

johnou:
@normanmaurer have a good trip!

normanmaurer:
@nmittler @Scottmitch @davidben WDYT ?

davidben:
I actually know nothing about this extension. BoringSSL does not support it, but I assume that if this code is more lax than necessary, it's fine because the actual SSL stack will process it more thoroughly.

kaja-mohideen:
Is there anything that I can help with to get this merged?

johnou:
From a security stand point I'd like to see heartbeats disabled by default and an option to enable them eg. via property @normanmaurer wdyt?

normanmaurer:
@trustin @nmittler WDYT ?

normanmaurer:
@trustin do you think we need a property for this or not ?

trustin:
I'm not sure we need such a knob here because the TLS records are meant to be processed by the SSLEngine implementation.

trustin:
@normanmaurer We don't. If we are going to add such a know, it should be available when creating SslContext or SslEngine.

normanmaurer:
@trustin agreed... let me just merge this one

normanmaurer:
Cherry-picked into 4.1 (13cd69c5ec3bba7d3efbd99ba171e9d3591a8f86) and 4.0 (bc53086)

@kaja-mohideen sorry for the slow turn-around

kaja-mohideen:
@normanmaurer No problem. Glad that I have did some tiny contribution back to the lib which I use in every day basis :-)

Scottmitch:
@johnou - As @trustin said it is not the SslHandler's responsibility to filter out so I don't think we should rely on SslHandler to protect anyone against vulnerabilities of specific implementations of the SSLEngine. In fact in PR https://github.com/netty/netty/pull/6365 I'm exploring having the SslHandler not worry about the framing and just let the SSLEngine (which knows the details of the the SSL/TLS protocols) make these decisions. Folks are responsible to keep software (Netty, JDK, and OpenSSL) up to date to avoid vulnerabilities in older versions of software.

normanmaurer:
@Scottmitch +1

PR #6537, opened 2017-03-13
ChunkedWriteHandler flushes too often

mcobrien:
Motivation:

ChunkedWriteHandler queues written messages and actually writes them
when flush is called. In its doFlush method, it needs to flush after
each chunk is written to preserve memory. However, non-chunked messages
(those that aren't of type ChunkedInput) are treated in the same way,
which means that flush is called after each message is written.

Modifications:

Moved the call to flush() inside the if block that tests if the message
is an instance of ChunkedInput. To ensure flush is called at least once,
the existing boolean flushed is checked at the end of doFlush. This
check was previously in ChunkedWriteHandler.flush(), but wasn't checked in
other invocations of doFlush, e.g. in channelInactive.

Result:

When this handler is present in a pipeline, writing a series
of non-chunked messages will be flushed as the developer intended.

Notes:

Hi, hope I've understood how this is supposed to work.

It seems the behaviour of flushing after non-chunked messages was
introduced in https://github.com/netty/netty/commit/b57d9f307f61161b62a335734438fb0c49260c84.

Previously doFlush called flush when it encountered a ChannelPromise (added at the end of the
queue in flush()) or after each chunk.

Making this change provides a nice perf boost in a library I'm writing.

### DIFF
--- a/handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java
+++ b/handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java
@@ -134,10 +134,7 @@ public class ChunkedWriteHandler extends ChannelDuplexHandler {
 
     @Override
     public void flush(ChannelHandlerContext ctx) throws Exception {
-        if (!doFlush(ctx)) {
-            // Make sure to flush at least once.
-            ctx.flush();
-        }
+        doFlush(ctx);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


axiak:
@normanmaurer ^

normanmaurer:
@axiak will have a look soon... thx

normanmaurer:
@mcobrien change looks good! Thanks a lot.

Can you please sign our ICLA:

https://netty.io/s/icla

mcobrien:
thanks @normanmaurer! signed :)

normanmaurer:
Cherry-picked into 4.0 (2407b5b) and 4.1 (61f53c4d07bb5a883673f31b798046427fb7d55f)

normanmaurer:
@mcobrien thanks a lot! Sorry it took me so long to merge it :(

PR #6543, opened 2017-03-14
Replace system.out.println with InternalLoggerFactory

asajim:
Motivation:

There are two files that still use `system.out.println` to log their status

Modification:

Replace `system.out.println` with a `debug` function inside an instance of `InternalLoggerFactory`

Result:

Introduce an instance of `InternalLoggerFactory` in class `AbstractMicrobenchmark.java` and `AbstractSharedExecutorMicrobenchmark.java`

### DIFF
--- a/microbench/src/main/java/io/netty/microbench/util/AbstractMicrobenchmark.java
+++ b/microbench/src/main/java/io/netty/microbench/util/AbstractMicrobenchmark.java
@@ -33,11 +33,15 @@ public class AbstractMicrobenchmark extends AbstractMicrobenchmarkBase {
 
     protected static final int DEFAULT_FORKS = 2;
 
+    private final InternalLogger logger = InternalLoggerFactory.getInstance(getClass());
+
     public static final class HarnessExecutor extends ThreadPoolExecutor {
         public HarnessExecutor(int maxThreads, String prefix) {
             super(maxThreads, maxThreads, 0, TimeUnit.MILLISECONDS,
                     new LinkedBlockingQueue<Runnable>(), new DefaultThreadFactory(prefix));
-            System.out.println("Using harness executor");
+            if (logger.isDebugEnabled()) {
+                logger.debug("Using harness executor");
+            }
         }
     }
 
--- a/microbench/src/main/java/io/netty/microbench/util/AbstractMicrobenchmark.java
+++ b/microbench/src/main/java/io/netty/microbench/util/AbstractMicrobenchmark.java
@@ -28,68 +28,69 @@ import org.openjdk.jmh.annotations.Fork;
 import org.openjdk.jmh.runner.options.ChainedOptionsBuilder;
 
 /**
- * Default implementation of the JMH microbenchmark adapter.  There may be context switches introduced by this harness.
+ * Default implementation of the JMH microbenchmark adapter. There may be
+ * context switches introduced by this harness.
  */
 @Fork(AbstractMicrobenchmark.DEFAULT_FORKS)
 public class AbstractMicrobenchmark extends AbstractMicrobenchmarkBase {
 
-    protected static final int DEFAULT_FORKS = 2;
+	protected static final int DEFAULT_FORKS = 2;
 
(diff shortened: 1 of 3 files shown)


microbench/src/main/java/io/netty/microbench/util/AbstractSharedExecutorMicrobenchmark.java
```diff
@@ -64,7 +65,9 @@ public static void executor(EventLoop eventLoop) {
     public static final class DelegateHarnessExecutor extends AbstractEventExecutor {
         private static EventLoop executor;
         public DelegateHarnessExecutor(int maxThreads, String prefix) {
-            System.out.println("Using DelegateHarnessExecutor executor " + this);
+            if (logger.isDebugEnabled()) {
```
fenik17: redundant check

microbench/src/main/java/io/netty/microbench/util/AbstractMicrobenchmark.java
```diff
@@ -33,11 +33,15 @@
...
     protected static final int DEFAULT_FORKS = 2;
 
+    private final InternalLogger logger = InternalLoggerFactory.getInstance(getClass());
+
     public static final class HarnessExecutor extends ThreadPoolExecutor {
         public HarnessExecutor(int maxThreads, String prefix) {
             super(maxThreads, maxThreads, 0, TimeUnit.MILLISECONDS,
                     new LinkedBlockingQueue<Runnable>(), new DefaultThreadFactory(prefix));
-            System.out.println("Using harness executor");
+            if (logger.isDebugEnabled()) {
```
fenik17: redundant check

microbench/src/main/java/io/netty/microbench/util/AbstractSharedExecutorMicrobenchmark.java
```diff
@@ -64,7 +65,9 @@ public static void executor(EventLoop eventLoop) {
...
+                logger.debug("Using DelegateHarnessExecutor executor " + this);
```
fenik17: 'logger.debug("Using DelegateHarnessExecutor executor {}", this);'

fenik17:
IMHO, it is fully acceptably to use `System.out` in the benchmarks. Why not?

fenik17:
@asajim compilation error becouse you missed imports.

microbench/src/main/java/io/netty/microbench/util/AbstractSharedExecutorMicrobenchmark.java
```diff
@@ -64,7 +65,9 @@ public static void executor(EventLoop eventLoop) {
     public static final class DelegateHarnessExecutor extends AbstractEventExecutor {
         private static EventLoop executor;
         public DelegateHarnessExecutor(int maxThreads, String prefix) {
-            System.out.println("Using DelegateHarnessExecutor executor " + this);
+            if (logger.isDebugEnabled()) {
+                logger.debug("Using DelegateHarnessExecutor executor {}" + this);
```
CodingFabian: this would need to be ", this"

asajim:
> IMHO, it is fully acceptably to use System.out in the benchmarks. Why not?

@fenik17 I was thinking that logging in any file inside src/main/java  folder should use InternalLogger to ensure that the developer would not missed the logging message since most developer already familiar with the logging class itself. I know that microbench is a benchmark project, but prevention always a good idea, right?

normanmaurer:
@asajim can you please sigh the ICLA and ping me once done ?

https://netty.io/s/icla

microbench/src/main/java/io/netty/microbench/util/AbstractMicrobenchmark.java
normanmaurer: @asajim just remove this check

asajim:
@normanmaurer I've just signed it. Thanks

normanmaurer:
Squashed and cherry-picked into 4.1 (016629fe3b65b3797e7e9de39a661bac196c41e0)

PR #6544, opened 2017-03-15
Adding method to assert XML decoder framing works

jeff303:
Motivation:

In an effort to better understand how the XmlFrameDecoder works, I consulted the tests to find a method that would reframe the inputs as per the Javadocs for that class.  I couldn't find any methods that seemed to be doing it, so I wanted to add one to reinforce my understanding.

Modification:

Add a new test method to XmlFrameDecoder to assert that the reframing works as described.

Result:

New test method is added to XmlFrameDecoder


normanmaurer:
@jeff303 thanks a lot!

Cherry-picked into 4.1 (476d2aea76847ccae6831d0f41a4d06d7a6e3b77)

PR #6552, opened 2017-03-20
http2: Http2StreamChannel now shares options of its parent channel

### LINKED ISSUE #6551: Http2StreamChannel doesn't inherit config of the parent channel
# Problem

As you probably aware, Finagle is a bit special when it comes to Netty 4 usage. For example, we neither use pooling nor prefer direct buffers at the moment. We're getting close to the pooled/direct utopia, but we're not there yet.

That said, Finagle's HTTP/2 pipelines are unpooled and yet we've discovered that inbound messages on the server-side are pooled in some cases. Turns out, `Http2StreamChannel` doesn't inherit the channel config from the parent channel (in our case, a channel with unpooled allocator) and just [defaults to a standard setup effectively enabling pooling](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/AbstractHttp2StreamChannel.java#L58).

A possible workaround is to override the default allocator (`-Dio.netty.allocator.type`).

Is there any reason to not pass parent's channel config to the `Http2StreamChannel` constructor to make sure it re-uses its settings? I'm happy to make a PR if that sounds like a reasonable solution.

I'm not excluding the possibility that this is done intentionally for HTTP/2 and there is a good reason to not reuse parent's channel config. I apologize if that's the case and I'm eager to hear your thoughts about this.

# Env Details

* Netty version: 4.1.9-Final
* JVM version: 1.8.0_111-b14
* OS: Darwin 15.6.0

vkostyukov:
Fixes #6551

# Motivation

`Http2StreamChannel` ignores options of its parent channel when being created. That leads to surprising results when, for example, unpooled allocator could be silently replaced with pooled allocator (default setting).

# Modification

Copy parent channel's options over to the `Http2StreamChannel`.

# Result

Channel options are now consistent between `Http2StreamChannel` and its parent channel. Newly added test passes on this branch and fails on master.

# Implementation Notes

My initial idea was to just return `parent().config()` in `Http2StreamChannel.config` implementation but I gave up on it by the following reasons:

1) It immediately failed the `Http2MultiplexCodecTest.channelReadShouldRespectAutoRead` test since `child.config().setAutoRead(true)` was no longer triggering `child.read()` request (it triggered `child.parent().read()` request instead). I assumed it was essential to keep this behaviour (`Http2StreamChannel.config().setAutoRead(true)` triggers read) in place.

2) Just referencing parent channel's options means they will be amended by `Http2StreamChannelBootstrap.options` at the moment the first HTTP/2 stream channel is created, which is not necessary what we want.

Presumably, I might be missing some context here so please let me know how do you feel about this approach. I'm happy to change the direction in this PR if people think it's better to reference parent's options (do not copy) and solve that auto-read problem in some ad-hoc way.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
@@ -305,6 +305,12 @@ public final class Http2MultiplexCodec extends ChannelDuplexHandler {
         }
         channel.pipeline().addLast(handler);
 
+        // We need to copy parent's channel options into a child's options to make
+        // sure they share same allocator, same receive buffer allocator, etc.
+        //
+        // See https://github.com/netty/netty/issues/6551
+        channel.config().setOptions(parentChannel.config().getOptions());
+
         initOpts(channel, options);
         initAttrs(channel, attrs);
 
--- a/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
+++ b/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
@@ -39,6 +39,7 @@ import org.junit.Before;
 import org.junit.Test;
 
 import static io.netty.util.ReferenceCountUtil.release;
+
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
@@ -341,6 +342,28 @@ public class Http2MultiplexCodecTest {
         assertFalse(channel.config().isAutoRead());
         assertSame(mark, channel.config().getWriteBufferWaterMark());
         assertEquals("bar", channel.attr(key).get());
+    }
+
+    @Test
+    public void childChannelShouldShareParentsChannelOptions() {
+        EmbeddedChannel parent = new EmbeddedChannel();
+        parent.config().setAutoRead(false);
+        parent.config().setWriteSpinCount(42);
+
+        Http2StreamChannelBootstrap b = new Http2StreamChannelBootstrap();
+        parent.pipeline().addLast(new Http2MultiplexCodec(true, b));
+
+        Channel child = b
+            .parentChannel(parent)
+            .handler(new TestChannelInitializer())
+            .connect()
+            .channel();
+
+        assertFalse(child.config().isAutoRead());
+        assertEquals(child.config().getWriteSpinCount(), 42);
+
+        child.close();
+        assertTrue(!parent.finish());
     }
 
     @Test


mosesn:
@vkostyukov can you also update the commit message to include your description notes?

normanmaurer:
@Scottmitch @nmittler PTAL as well

nmittler:
This seems fine to me. 

@ejona86 want to take a quick peek?

fangqiang:
ok

vkostyukov:
Thanks for the review, @normanmaurer @nmittler!

If we're too worried about copying everything over, I'd be totally down to just copying allocators (both regular and receive buffers) along with auto-read properties for now. Let me know how do you feel about that.

normanmaurer:
Cherry-picked into 4.1 (7467106630ad6b2522f7953c244546b91b107efe).

@vkostyukov thanks!

ejona86:
Copying the options makes some sense. Although copying auto-read doesn't sound great since it seems there isn't a strong relation between a parent and a child. Copying watermarks also probably doesn't make sense in the same regard. Looking through ChannelOption, it seems _only_ ByteBufAllocator and MessageSizeEstimator makes sense (maybe RecvByteBufAllocator, but it isn't really used on the child). Everything else is either non-applicable or seems to be independent of the parent.

vkostyukov:
/I think/ I agree with @ejona86.

I'm happy to open a new PR that only copies the allocator and the message size estimator (along with updated tests). @normanmaurer Please, let me know if you want me to do that.

normanmaurer:
@vkostyukov @ejona86 I am not super strong on this but I wonder why you think it is not good to just copy all. This way we are at least "consistent".

ejona86:
I simply looked at the various options.

* `ALLOCATOR`. Looks like it would be useful to copy. This is a non-exclusive resource acquisition configuration option, so copying it makes sense in the same way that we may reuse the parent's event loop.
* `ALLOW_HALF_CLOSURE`. This seems independent between parent and child. Since the two will  have different handlers and are working with different APIs, I see no reason to believe it should be the same. Having it be the same just seems more prone to issue, since handlers on the child have to protect themselves for the different behavior.
* `AUTO_CLOSE`. This seems independent. Same as `ALLOW_HALF_CLOSURE`.
* `AUTO_READ`. This seems independent. Same as `ALLOW_HALF_CLOSURE`.
* `CONNECT_TIMEOUT_MILLIS`. This seems independent. The TCP creation happens once. I don't see how that would relate to per-stream creation timeout.
* `DATAGRAM_CHANNEL_ACTIVE_ON_REGISTRATION`. Don't entirely know, but it seems maybe N/A
* `IP_*`. N/A
* `MESSAGE_SIZE_ESTIMATOR`. Seems independent, but may be useful for the child to receive a default. The parent wouldn't "naturally" know how to estimate sizes for the child, and the parent would be handling bytes whereas the child is handling messages. But the parent is in a position to know what messages may be sent by the child, and so could coordinate a single estimator that works for both
* `RCVBUF_ALLOCATOR`. As an allocator it would seem to make sense to copy. But we aren't actually allocating anything for the child reading. And since this same option replaces the `MAX_MESSAGES_PER_READ`, which seems independent, it seems this shouldn't be copied.
* `SINGLE_EVENTEXECUTOR_PER_GROUP`. I don't know what this does.
* `SO_*`. N/A
* `TCP_NODELAY`. N/A
* `WRITE_BUFFER_WATER_MARK`. On face value it could be considered dependent, but I think that's actually probably not the case. Since writing from the child to the parent is messages instead of bytes, doesn't involve write syscalls, and there is lower latency, it actually seems independent.
* `WRITE_SPIN_COUNT`. N/A

mosesn:
@ejona86 from a principle of least surprise point of view, it might make sense to stick with copying everything.  It's simpler to explain, "it copies everything" than "it copies ByteBufAllocator and SizeEstimator".  For every case where it's independent, it's probably OK for them to start as the same, and then the child can configure it if it needs to.  For every case where it's N/A, no harm no foul.

normanmaurer:
That's exactly my POV here as well

ejona86:
From a principle of least surprise, then I'd suggest not copying anything. You can't willy-nilly change the value of `ALLOW_HALF_CLOSE`, `AUTO_CLOSE`, and `AUTO_READ` because it changes the channel behavior and it is _highly unlikely_ that code "naturally" works correctly independent of the value. Most pipelines would be prone to leaking channels if `ALLOW_HALF_CLOSE` is `true`. Most pipelines would just stall if `AUTO_READ` is disabled. Even `CONNECT_TIMEOUT_MILLIS` seems bug-causing. Copying those options _would cause bugs_. Auto-copying the allocator is a nice-to-have.

From a conceptual point of view, it makes sense _not_ to copy, since channel options configure a channel and a byte-based channel that uses TCP between two computers is fundamentally different than a message-based channel that is in-process.

Even copying `EventLoop` between different channel types is not safe in the general case (if we aren't talking specifically about h2childchan). Many channels require a specific type of `EventLoop`; it just so happens that the `Http2StreamChannel` can work with any `EventLoop`.

`ALLOCATOR` is a special case because it is for resource management and different implementations have no _semantic_ differences. Copying `ALLOCATOR` could cause bugs, in that `buffer()` could swap from heap to direct or vise versa or the default buffer size is different, but code is generally expected to handle that already.

I feel like we're strongly in special-case land if we're talking about auto-copying configuration.

Also, at this point I'd suggest not copying `MESSAGE_SIZE_ESTIMATOR`. I added it at the last minute in my earlier comment. Given `Http2StreamChannelBootstrap` (which is a bit new to me), it seems copying the estimator has no real value.

If we're so against a special case, then I wonder why not to avoid copying and instead just rely on users specifying child channel defaults, like is done with `ServerBootstrap`, which seems to be the best example of prior-art. That already seems available available via `Http2StreamChannelBootstrap`.

mosesn:
@ejona86 actually, reexamining some of these, like AUTO_READ, the default is that it's on, so if I disabled it, I would expect it to be off everywhere, including for child channels.  It seems like a pretty weird behavior if it's not propagated by default.  Same for AUTO_CLOSE and ALLOW_HALF_READ.  I think the common case would be that you want to propagate it, not that you want to have to change the default for every child channel.

I don't agree that it makes sense not to copy because it's a byte channel instead of a message channel.  I would expect properties to be consistent, especially for things like AUTO_READ which require that your code have substantively different semantics.

For the channels that require a specific event loop, I would expect that their children require the same event loop in general.

I think it makes more sense for `Http2StreamChannelBoostrap` to have options that differ from what the parent is configured with.  Otherwise, we'll have to wire something together that ensures that `Http2StreamChannelBootstrap` also gets the options we configure the `ServerBootstrap` with every time we want to share a configuration parameter (which is probably the common case).

Scottmitch:
FWIW I agree with @ejona86. It seems more typical for a user to explicitly specify the channel options in a `BootStrap` object instead of copying the parent channel's options. As described by @ejona86 copying may produce questionable (and may even undesirable) behavior for some options, and who knows about options that may be added in the future. If the intention is to preserve some/all options from the parent channel this could be done explicitly in a method such as `initChannel` (or some utility could be created to do this if it is common enough).

ejona86:
> like AUTO_READ, the default is that it's on, so if I disabled it, I would expect it to be off everywhere, including for child channels.

Disabling AUTO_READ is to manually manage TCP flow control. In HTTP/2, this is a "[bad idea](http://httpwg.org/specs/rfc7540.html#rfc.section.5.2.2)," except for short times when under heavy load. If I was implementing something that used that, I'd change the option over time; have it enabled most of the time and only disable it when things are bad (and centrally manage that; the handler would be ignorant). That would cause child-channels to randomly receive one of the configurations and not revert when the parent changes.

The inverse is also true: with gRPC we'd highly likely disable AUTO_READ on the child channels, but leave it on for the parent. Stream-level flow control is hugely important for streaming in gRPC. But as I already mentioned, HTTP/2 requires "disabling" TCP flow control and promptly reading from the socket.

The flow control on the child is completely separate from that of the parent; they are different counters and seem to be used in different situations. This is not surprising, since the parent and child are processing different protocols. The parent doesn't care what the child's value is, and vise versa. A strong point of h2childchan was that the parent and child are orthogonal and independent.

The HTTP/2 handler is huge and complex, but there's only one. It easily has reason to enable niche configuration and use a complex solution to a problem. But handlers on the child channel will be numerous; I don't see how they would deal with the various options without explicit testing first. This applies to the current options and any new options created in the future.

> For the channels that require a specific event loop, I would expect that their children require the same event loop in general.

Look at the communication between the parent and the child in the HTTP/2 case. It shares _no relation_ to NIO or LocalChannel or EmbeddedChannel or whatever else the parent channel is using. The child channel implementation is chosen by a _handler_, not by the parent's channel type. They are _completely_ unrelated.

> Otherwise, we'll have to wire something together that ensures that Http2StreamChannelBootstrap also gets the options we configure the ServerBootstrap with every time we want to share a configuration parameter (which is probably the common case).

If it is the common case for you, I'd suggest your `ChannelInitializer` copy the things you care. That could happen either on the parent channel (when creating `Http2MultiplexCodec`) or on the child channel (with `ctx.channel().parent()`). I would probably do the copying on the parent channel to avoid child configuration changing over time.

mosesn:
Talked to @vkostyukov and he helped explain your argument to me, I'm convinced.  Let's switch the behavior back.

normanmaurer:
Alright... let me revert this one then

vkostyukov:
Thanks, Norman! I'm sorry it's been so controversial and yet I'm glad we talked about that clarified certain things. I grateful to @ejona86 for passionately digging into that.

normanmaurer:
@vkostyukov no worries... And thanks again to @ejona86 ! :)

normanmaurer:
Done 40bead56c4

PR #6553, opened 2017-03-20
fix failure logging of value on channel option set

boazavital:
Pass value to failure log as value instead of accidentally passing channel.

Motivation:

The code accidentally passes channel twice instead of value, resulting in logs like:
`Failed to set channel option 'SO_SNDBUF' with value '[id: 0x2c5b2eb4]' for channel '[id: 0x2c5b2eb4]'`

Modifications:

Pass value instead of channel where it needs to be.

Result:

`Failed to set channel option 'SO_SNDBUF' with value '0' for channel '[id: 0x9bd3c5b8]'`

### DIFF
--- a/transport/src/main/java/io/netty/bootstrap/AbstractBootstrap.java
+++ b/transport/src/main/java/io/netty/bootstrap/AbstractBootstrap.java
@@ -461,7 +461,7 @@ public abstract class AbstractBootstrap<B extends AbstractBootstrap<B, C>, C ext
             }
         } catch (Throwable t) {
             logger.warn(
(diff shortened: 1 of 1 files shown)


normanmaurer:
@boazavital thanks a lot ... can you please change the commit message to match our template:

boazavital:
Of course, updated. Note that the link explicitly says "unless the commit is trivial", which I believe applies to this. If that's no longer true, consider changing the wiki.

normanmaurer:
Cherry-picked into 4.1 (bf08ed0b93e54e8e21f23e9d948fd2732dc52171) and 4.0 (99dbf1ba7740df576a22c9a6d4b7ab72c86c8fc6)

PR #6554, opened 2017-03-20
Adding new length-based frame decoder for cases when the length field…

jeff303:
… is delimited, not fixed length, adding unit test

Motivation:

I am working on parsing Syslog over TCP ([RFC 6587](https://tools.ietf.org/html/rfc6587#section-3.4.1)).  In this format, the frame length is in a field that is separated by a space from the data itself.  I wasn't able to find a built-in decoder in Netty to handle it, since the existing LengthFieldBasedFrameDecoder relies in a fixed-width length field.

Modification:

Added new decoder class and unit test, made io.netty.handler.codec.DelimiterBasedFrameDecoder#indexOf public so new class can call it

Result:

New decoder class and unit test

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
+++ b/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
@@ -72,6 +72,10 @@ public final class ByteBufUtil {
     static final int WRITE_CHUNK_SIZE = 8192;
     static final ByteBufAllocator DEFAULT_ALLOCATOR;
 
+    static final String BOM_STR = "\uFEFF";
+    static final byte[] BOM_UTF8 = BOM_STR.getBytes(CharsetUtil.UTF_8);
+    static final byte[] BOM_US_ASCII = BOM_STR.getBytes(CharsetUtil.US_ASCII);
+
     static {
         String allocType = SystemPropertyUtil.get(
                 "io.netty.allocator.type", PlatformDependent.isAndroid() ? "unpooled" : "pooled");
@@ -1471,6 +1475,37 @@ public final class ByteBufUtil {
             out.write(in, inOffset, len);
             outLen -= len;
         } while (outLen > 0);
+    }
+
+    /**
+     * Determines the length of the byte-order mark as it occurs at the beginning of the
+     * specified {@link ByteBuf} input.  Returns the length if found, otherwise -1.
+     *
+     * @param buf the input buffer, in which to locate the BOM (U+FEFF)
+     * @param charset the charset that should be used to interpret the buffer
+     * @return the length of the byte-order mark, if it occurs in the input, or -1 if the BOM does not occur
+     */
+    public static int lengthOfByteOrderMark(ByteBuf buf, Charset charset) {
+        byte[] bom;
+        if (CharsetUtil.UTF_8.equals(charset)) {
+            bom = BOM_UTF8;
+        } else if (CharsetUtil.US_ASCII.equals(charset)) {
+            bom = BOM_US_ASCII;
+        } else {
+            bom = BOM_STR.getBytes(charset);
+        }
+
+        int i;
+        for (i = 0; i < bom.length && i < buf.readableBytes(); i++) {
+            if (bom[i] != buf.getByte(i)) {
+                break;
+            }
+        }
+        if (i == bom.length) {
+            return i;
+        } else {
+            return 0;
+        }
     }
 
     private ByteBufUtil() { }
--- a/buffer/src/test/java/io/netty/buffer/ByteBufUtilLengthOfBomTest.java
+++ b/buffer/src/test/java/io/netty/buffer/ByteBufUtilLengthOfBomTest.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2014 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.buffer;
+
+import com.google.common.base.Charsets;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+import static org.hamcrest.CoreMatchers.*;
+import static org.junit.Assert.*;
+
+@RunWith(Parameterized.class)
+public class ByteBufUtilLengthOfBomTest {
+
+    private final Charset charset;
+    private final int expectedBomLength;
+    private final boolean injectBom;
+
+    public ByteBufUtilLengthOfBomTest(Charset charset, int expectedBomLength, boolean injectBom) {
+        this.charset = charset;
+        this.expectedBomLength = expectedBomLength;
+        this.injectBom = injectBom;
+    }
+
+    @Test
+    public void lengthOfBom() throws IOException {
+        final byte[] bytes = ((injectBom ? ByteBufUtil.BOM_STR : "") + "Hello there").getBytes(charset);
+
+        ByteBuf buf = Unpooled.copiedBuffer(bytes);
+        try {
+            int length = ByteBufUtil.lengthOfByteOrderMark(buf, charset);
+            assertThat(length, equalTo(expectedBomLength));
+        } finally {
+            buf.release();
+        }
+    }
+
+    @Parameterized.Parameters(name = "{index}: charset = {0}, expectedBomLength = {1}, injectBom = {2}")
+    public static Collection<Object> data() {
+        List<Object> params = new ArrayList<Object>();
+        params.add(new Object[]{Charsets.UTF_8, 3, true});
+        params.add(new Object[]{Charsets.UTF_16BE, 2, true});
+        params.add(new Object[]{Charsets.UTF_16LE, 2, true});
+        params.add(new Object[]{Charsets.UTF_8, 0, false});
+        params.add(new Object[]{Charsets.UTF_16BE, 0, false});
+        params.add(new Object[]{Charsets.UTF_16LE, 0, false});
+        return params;
+    }
+
+}
--- a/codec/src/main/java/io/netty/handler/codec/BaseLengthBasedFrameDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/BaseLengthBasedFrameDecoder.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.codec;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.serialization.ObjectDecoder;
+
+import java.util.List;
+
+/**
+ * Encapsulates common logic and state for message decoders that attempt to detect frames
+ * of a certain length
+ */
+public abstract class BaseLengthBasedFrameDecoder extends ByteToMessageDecoder {
+
+    protected final int maxFrameLength;
+    protected final boolean failFast;
+    protected boolean discardingTooLongFrame;
+    protected long tooLongFrameLength;
+    protected long bytesToDiscard;
+
+    protected BaseLengthBasedFrameDecoder(int maxFrameLength, boolean failFast) {
+        this.maxFrameLength = maxFrameLength;
+        this.failFast = failFast;
+    }
+
+    /**
+     * Fail the decoding due to a frame that is too long, as per {@link #maxFrameLength}
+     *
+     * @param firstDetectionOfTooLongFrame true if this is the first time the too-long frame was detected
+     * @return true if all bytes from the too-long frame were discarded
+     */
+    protected boolean failIfNecessary(boolean firstDetectionOfTooLongFrame) {
+        if (bytesToDiscard == 0) {
+            // Reset to the initial state and tell the handlers that
+            // the frame was too large.
+            final long prevTooLongFrameLength = tooLongFrameLength;
+            tooLongFrameLength = 0;
+            discardingTooLongFrame = false;
+            if (!failFast || firstDetectionOfTooLongFrame) {
+                fail(prevTooLongFrameLength);
+            }
+            return true;
+        }
+        // Keep discarding and notify handlers if necessary.
+        if (failFast && firstDetectionOfTooLongFrame) {
+            fail(tooLongFrameLength);
+        }
+        return false;
+    }
+
+    protected void resetStateOnTooLongFrame() {
+        // do nothing by default
+    }
+
+    /**
+     * Extract the sub-region of the specified buffer.
+     * <p>
+     * If you are sure that the frame and its content are not accessed after
+     * the current {@link #decode(ChannelHandlerContext, ByteBuf)}
+     * call returns, you can even avoid memory copy by returning the sliced
+     * sub-region (i.e. <tt>return buffer.slice(index, length)</tt>).
+     * It's often useful when you convert the extracted frame into an object.
+     * Refer to the source code of {@link ObjectDecoder} to see how this method
+     * is overridden to avoid memory copy.
+     */
+    protected ByteBuf extractFrame(ByteBuf buffer, int index, int length) {
+        return buffer.retainedSlice(index, length);
+    }
+
+    /**
+     * @deprecated - use {@link #extractFrame(ByteBuf, int, int)} instead, since the
+     * {@link ChannelHandlerContext} is not actually needed for implementation
+     */
+    @Deprecated
+    protected ByteBuf extractFrame(@SuppressWarnings({ "unused", "squid:S1172" }) ChannelHandlerContext ignored,
+                                   ByteBuf buffer, int index, int length) {
+        return buffer.retainedSlice(index, length);
+    }
+
+    @Override
+    protected final void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws Exception {
+        Object decoded = decode(ctx, in);
+        if (decoded != null) {
+            out.add(decoded);
+        }
+    }
+
+    protected abstract Object decode(@SuppressWarnings({ "unused", "squid:S1172" }) ChannelHandlerContext ctx,
+                                     ByteBuf in) throws Exception;
+
+    protected void discardingTooLongFrame(ByteBuf in) {
+        long bytesToDiscard = this.bytesToDiscard;
+        int localBytesToDiscard = (int) Math.min(bytesToDiscard, in.readableBytes());
+        in.skipBytes(localBytesToDiscard);
+        bytesToDiscard -= localBytesToDiscard;
+        this.bytesToDiscard = bytesToDiscard;
+
+        failIfNecessary(false);
+    }
+
+    protected void fail(long frameLength) {
+        resetStateOnTooLongFrame();
+        if (frameLength > 0) {
+            throw new TooLongFrameException(
+                "Adjusted frame length exceeds " + maxFrameLength +
+                    ": " + frameLength + " - discarded");
+        } else {
+            throw new TooLongFrameException(
+                "Adjusted frame length exceeds " + maxFrameLength +
+                    " - discarding");
+        }
+    }
+}
--- a/codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
@@ -1,0 +1,181 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.codec;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufUtil;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.util.CharsetUtil;
+
+import java.nio.charset.Charset;
+
+/**
+ * <p> Frames input by looking for a length field (in bytes) that is delimited, as opposed to being a fixed byte length.
+ *  Only single-byte encoding (i.e. UTF-8) of the length field and delimiter characters are supported, although the
+ * remaining frame can be anything (i.e. is not interpreted by this decoder). </p><p> Uses the same approach as the
+ * {@link DelimiterBasedFrameDecoder} to find delimiter in input </p>
+ */
+public class DelimitedLengthFieldBasedFrameDecoder extends BaseLengthBasedFrameDecoder {
+
+    public static final Charset CHARSET = CharsetUtil.UTF_8;
+
+    private final int lengthAdjustment;
+    private final ByteBuf delimiter;
+    private final boolean trimLengthString;
+
+    private boolean consumingLength = true;
+    private long frameLength;
+
+    private final int maxPossibleLengthInBytes;
+
+    public DelimitedLengthFieldBasedFrameDecoder(
+            int maxFrameLength,
+            int lengthAdjustment,
+            boolean failFast,
+            ByteBuf delimiter,
+            boolean trimLengthString
+    ) {
+        super(maxFrameLength, failFast);
+        if (delimiter == null) {
+            throw new NullPointerException("delimiter");
+        }
+        if (!delimiter.isReadable()) {
+            throw new IllegalArgumentException("empty delimiter");
+        }
+
+        this.delimiter = delimiter.slice(delimiter.readerIndex(), delimiter.readableBytes());
+        this.lengthAdjustment = lengthAdjustment;
+        this.trimLengthString = trimLengthString;
+
+        maxPossibleLengthInBytes = String.valueOf(maxFrameLength).length();
+    }
+
+    protected int indexWithin(ByteBuf haystack, ByteBuf needle) {
+        return DelimiterBasedFrameDecoder.indexWithin(haystack, needle);
+    }
+
+    /**
+     * Create a frame out of the {@link ByteBuf} and return it.
+     *
+     * @param ctx the {@link ChannelHandlerContext} which this {@link ByteToMessageDecoder} belongs to
+     * @param in the {@link ByteBuf} from which to read data
+     *
+     * @return the {@link ByteBuf} which represents the frame or {@code null} if no frame could be created.
+     */
+    @Override
+    protected Object decode(@SuppressWarnings({ "unused", "squid:S1172" }) ChannelHandlerContext ctx, ByteBuf in)
+            throws Exception {
+        if (discardingTooLongFrame) {
+            long bytesToDiscard = this.bytesToDiscard;
+            int localBytesToDiscard = (int) Math.min(bytesToDiscard, in.readableBytes());
+            in.skipBytes(localBytesToDiscard);
+            bytesToDiscard -= localBytesToDiscard;
+            this.bytesToDiscard = bytesToDiscard;
+
+            failIfNecessary(false);
+
+            // if exception was not thrown, reader has been advanced by readable bytes already, so return null to
+            // continue processing too long frame
+            return null;
+        }
+
+        if (consumingLength) {
+            int delimIndex = indexWithin(in, delimiter);
+            if (delimIndex < 0) {
+                return null;
+            }
+
+            frameLength = extractFrameLength(in, delimIndex);
+
+            if (frameLength < 0) {
+                throw new CorruptedFrameException("negative pre-adjustment length field: " + frameLength);
+            }
+
+            // this will not overflow because lengthAdjustment is an int
+            frameLength += lengthAdjustment;
+
+            //consume length field and delimiter bytes
+            in.skipBytes(delimIndex + delimiter.capacity());
+
+            //consume delimiter bytes
+            consumingLength = false;
+        }
+
+        if (frameLength > maxFrameLength) {
+            long discard = frameLength - in.readableBytes();
+            tooLongFrameLength = frameLength;
+
+            if (discard < 0) {
+                // buffer contains more bytes then the frameLength so we can discard all now
+                // will not overflow because frameLength must be less than in.readableBytes, an int
+                in.skipBytes((int) frameLength);
+            } else {
+                // Enter the discard mode and discard everything received so far.
+                discardingTooLongFrame = true;
+                bytesToDiscard = discard;
+                in.skipBytes(in.readableBytes());
+            }
+            failIfNecessary(true);
+            return null;
+        }
+
+        // never overflows because it's less than maxFrameLength
+        int frameLengthInt = (int) frameLength;
+        if (in.readableBytes() < frameLengthInt) {
+            // need more bytes available to read actual frame
+            return null;
+        }
+
+        // extract frame
+        int readerIndex = in.readerIndex();
+        ByteBuf frame = extractFrame(in, readerIndex, frameLengthInt);
+        in.readerIndex(readerIndex + frameLengthInt);
+
+        // the frame is now entirely present, reset state vars
+        consumingLength = true;
+        frameLength = 0;
+
+        return frame;
+    }
+
+    private int extractFrameLength(ByteBuf in, int delimIndex) {
+        if (delimIndex > maxPossibleLengthInBytes) {
+            throw new TooLongFrameException(
+                    "Length field delimiter offset (" + delimIndex + ") exceeds maximum possible byte length for" +
+                    " specified frame size (" + maxPossibleLengthInBytes + ")"
+            );
+        }
+        final String lengthStr = in.toString(in.readerIndex(), delimIndex, CHARSET);
+        try {
+            return Integer.parseInt(trimLengthString? lengthStr.trim() : lengthStr);
+
+        } catch (NumberFormatException e) {
+            throw new CorruptedFrameException(
+                    String.format(
+                            "Invalid length field decoded: %s",
+                            lengthStr
+                    ),
+                    e
+            );
+        }
+    }
+
+    @Override
+    protected void resetStateOnTooLongFrame() {
+        consumingLength = true;
+        frameLength = 0;
+    }
+}
--- a/codec/src/main/java/io/netty/handler/codec/DelimiterBasedFrameDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/DelimiterBasedFrameDecoder.java
@@ -233,7 +233,7 @@ public class DelimiterBasedFrameDecoder extends ByteToMessageDecoder {
         int minFrameLength = Integer.MAX_VALUE;
         ByteBuf minDelim = null;
         for (ByteBuf delim: delimiters) {
-            int frameLength = indexOf(buffer, delim);
+            int frameLength = indexWithin(buffer, delim);
             if (frameLength >= 0 && frameLength < minFrameLength) {
                 minFrameLength = frameLength;
                 minDelim = delim;
@@ -250,10 +250,10 @@ public class DelimiterBasedFrameDecoder extends ByteToMessageDecoder {
                 discardingTooLongFrame = false;
                 buffer.skipBytes(minFrameLength + minDelimLength);
 
-                int tooLongFrameLength = this.tooLongFrameLength;
-                this.tooLongFrameLength = 0;
+                final int prevTooLongFrameLength = tooLongFrameLength;
+                tooLongFrameLength = 0;
                 if (!failFast) {
-                    fail(tooLongFrameLength);
+                    fail(prevTooLongFrameLength);
                 }
                 return null;
             }
@@ -309,8 +309,13 @@ public class DelimiterBasedFrameDecoder extends ByteToMessageDecoder {
      * Returns the number of bytes between the readerIndex of the haystack and
      * the first needle found in the haystack.  -1 is returned if no needle is
      * found in the haystack.
+     *
+     * @param haystack the buffer to search for an occurence of the needle
+     * @param needle the buffer to find in the haystack
+     *
+     * @return the first relative byte index of the needle found within the haystack, or -1 if not found
      */
-    private static int indexOf(ByteBuf haystack, ByteBuf needle) {
+    protected static int indexWithin(ByteBuf haystack, ByteBuf needle) {
         for (int i = haystack.readerIndex(); i < haystack.writerIndex(); i ++) {
             int haystackIndex = i;
             int needleIndex;
--- a/codec/src/main/java/io/netty/handler/codec/LengthFieldBasedFrameDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/LengthFieldBasedFrameDecoder.java
@@ -184,19 +184,14 @@ import io.netty.channel.ChannelHandlerContext;
  * </pre>
  * @see LengthFieldPrepender
  */
-public class LengthFieldBasedFrameDecoder extends ByteToMessageDecoder {
+public class LengthFieldBasedFrameDecoder extends BaseLengthBasedFrameDecoder {
 
     private final ByteOrder byteOrder;
-    private final int maxFrameLength;
     private final int lengthFieldOffset;
     private final int lengthFieldLength;
     private final int lengthFieldEndOffset;
     private final int lengthAdjustment;
     private final int initialBytesToStrip;
-    private final boolean failFast;
-    private boolean discardingTooLongFrame;
-    private long tooLongFrameLength;
-    private long bytesToDiscard;
 
     /**
      * Creates a new instance.
@@ -301,10 +296,8 @@ public class LengthFieldBasedFrameDecoder extends ByteToMessageDecoder {
     public LengthFieldBasedFrameDecoder(
             ByteOrder byteOrder, int maxFrameLength, int lengthFieldOffset, int lengthFieldLength,
             int lengthAdjustment, int initialBytesToStrip, boolean failFast) {
-
+        super(checkPositive(maxFrameLength, "maxFrameLength"), failFast);
         this.byteOrder = checkNotNull(byteOrder, "byteOrder");
-
-        checkPositive(maxFrameLength, "maxFrameLength");
 
         checkPositiveOrZero(lengthFieldOffset, "lengthFieldOffset");
 
@@ -318,31 +311,11 @@ public class LengthFieldBasedFrameDecoder extends ByteToMessageDecoder {
                     "lengthFieldLength (" + lengthFieldLength + ").");
         }
 
-        this.maxFrameLength = maxFrameLength;
         this.lengthFieldOffset = lengthFieldOffset;
         this.lengthFieldLength = lengthFieldLength;
         this.lengthAdjustment = lengthAdjustment;
         this.lengthFieldEndOffset = lengthFieldOffset + lengthFieldLength;
         this.initialBytesToStrip = initialBytesToStrip;
-        this.failFast = failFast;
-    }
-
-    @Override
-    protected final void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws Exception {
-        Object decoded = decode(ctx, in);
-        if (decoded != null) {
-            out.add(decoded);
-        }
-    }
-
-    private void discardingTooLongFrame(ByteBuf in) {
-        long bytesToDiscard = this.bytesToDiscard;
-        int localBytesToDiscard = (int) Math.min(bytesToDiscard, in.readableBytes());
-        in.skipBytes(localBytesToDiscard);
-        bytesToDiscard -= localBytesToDiscard;
-        this.bytesToDiscard = bytesToDiscard;
-
-        failIfNecessary(false);
     }
 
     private static void failOnNegativeLengthField(ByteBuf in, long frameLength, int lengthFieldEndOffset) {
@@ -434,7 +407,7 @@ public class LengthFieldBasedFrameDecoder extends ByteToMessageDecoder {
         // extract frame
         int readerIndex = in.readerIndex();
         int actualFrameLength = frameLengthInt - initialBytesToStrip;
-        ByteBuf frame = extractFrame(ctx, in, readerIndex, actualFrameLength);
+        ByteBuf frame = extractFrame(in, readerIndex, actualFrameLength);
         in.readerIndex(readerIndex + actualFrameLength);
         return frame;
     }
@@ -473,40 +446,10 @@ public class LengthFieldBasedFrameDecoder extends ByteToMessageDecoder {
         return frameLength;
     }
 
-    private void failIfNecessary(boolean firstDetectionOfTooLongFrame) {
-        if (bytesToDiscard == 0) {
-            // Reset to the initial state and tell the handlers that
-            // the frame was too large.
-            long tooLongFrameLength = this.tooLongFrameLength;
-            this.tooLongFrameLength = 0;
-            discardingTooLongFrame = false;
-            if (!failFast || firstDetectionOfTooLongFrame) {
-                fail(tooLongFrameLength);
-            }
-        } else {
-            // Keep discarding and notify handlers if necessary.
-            if (failFast && firstDetectionOfTooLongFrame) {
-                fail(tooLongFrameLength);
-            }
-        }
-    }
-
     /**
      * Extract the sub-region of the specified buffer.
      */
     protected ByteBuf extractFrame(ChannelHandlerContext ctx, ByteBuf buffer, int index, int length) {
         return buffer.retainedSlice(index, length);
-    }
-
-    private void fail(long frameLength) {
-        if (frameLength > 0) {
-            throw new TooLongFrameException(
-                            "Adjusted frame length exceeds " + maxFrameLength +
-                            ": " + frameLength + " - discarded");
-        } else {
-            throw new TooLongFrameException(
-                            "Adjusted frame length exceeds " + maxFrameLength +
-                            " - discarding");
-        }
     }
 }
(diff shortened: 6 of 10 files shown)


jeff303:
This is very much a work-in-progress, and I expect I will need to make more changes here.  I wanted to get it out early to get feedback, to see whether it's something that's even worth pursuing (as opposed to keeping it on our own project's code).  Known changes still needed:

- adding Javadocs
- potentially more unit test methods
- determining utility class/approach for random seed management

codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,217 @@
...
+   * <p>
+   * If you are sure that the frame and its content are not accessed after
+   * the current {@link #decode(ChannelHandlerContext, ByteBuf)}
+   * call returns, you can even avoid memory copy by returning the sliced
+   * sub-region (i.e. <tt>return buffer.slice(index, length)</tt>).
+   * It's often useful when you convert the extracted frame into an object.
+   * Refer to the source code of {@link ObjectDecoder} to see how this method
+   * is overridden to avoid memory copy.
+   */
+  protected ByteBuf extractFrame(ChannelHandlerContext ctx, ByteBuf buffer, int index, int length) {
```
netkins: MAJOR Remove this unused method parameter "ctx". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,217 @@
...
+  private void failIfNecessary(boolean firstDetectionOfTooLongFrame) {
+    if (bytesToDiscard == 0) {
+      // Reset to the initial state and tell the handlers that
+      // the frame was too large.
+      long tooLongFrameLength = this.tooLongFrameLength;
+      this.tooLongFrameLength = 0;
+      discardingTooLongFrame = false;
+      consumingLength = false;
+      this.frameLength = 0;
+      if (!failFast || failFast && firstDetectionOfTooLongFrame) {
```
netkins: BLOCKER Change this condition so that it does not always evaluate to "true" [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2583)

codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,217 @@
...
+  protected ByteBuf extractFrame(ChannelHandlerContext ctx, ByteBuf buffer, int index, int length) {
+    return buffer.slice(index, length).retain();
+  }
+
+  private void fail(long frameLength) {
+    if (frameLength > 0) {
+      throw new TooLongFrameException("Adjusted frame length exceeds " + maxFrameLength + ": " + frameLength + " - " +
+          "discarded");
+    } else {
+      throw new TooLongFrameException("Adjusted frame length exceeds " + maxFrameLength + " - discarding");
```
fenik17: Consider use ternary operator inside concatination:
```java
private void fail(long frameLength) {
    throw new TooLongFrameException("Adjusted frame length exceeds " + maxFrameLength 
        + (frameLength > 0 ? ": " + frameLength : "") + " - discarding");
}
```

codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
fenik17: @netkins +1.

Change to: ```if (!failFast || firstDetectionOfTooLongFrame)```

fenik17: But maybe here was meant `failFast || !firstDetectionOfTooLongFrame`?
I'm not sure that I understand the meaning of the `failFast` flag..

codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,217 @@
...
+  private final boolean failFast;
+  private final Charset lengthFieldCharset;
+  private final ByteBuf delimiter;
+  private final boolean trimLengthString;
+
+  private boolean discardingTooLongFrame;
+  private long tooLongFrameLength;
+  private long bytesToDiscard;
+  private boolean consumingLength = true;
+  private long frameLength;
```
fenik17: Why it's `long` while the `maxFrameLength` is `int`?

codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,217 @@
...
+      int localBytesToDiscard = (int) Math.min(bytesToDiscard, in.readableBytes());
+      in.skipBytes(localBytesToDiscard);
+      bytesToDiscard -= localBytesToDiscard;
+      this.bytesToDiscard = bytesToDiscard;
+
+      failIfNecessary(false);
+    }
+
+    if (consumingLength) {
+      int delimIndex = indexOf(in, delimiter);
```
fenik17: Consider extracting a search of the delimeter into the separate `protected` method. This will allow us to search by a more complex rule if need be.

codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,217 @@
...
+      if (delimIndex < 0) {
+        return null;
+      }
+
+      final String lengthStr = in.toString(in.readerIndex(), delimIndex, lengthFieldCharset);
```
fenik17: ditto. Consider extracting a getting of the `frameLength` into a separate protected method.

codec/src/test/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  public static int nextInt(final int startInclusive, final int endExclusive) {
+
+    // TODO: add management of random seed for repeatable tests
+    if (startInclusive == endExclusive) {
+      return startInclusive;
+    }
+
+    return startInclusive + new Random().nextInt(endExclusive - startInclusive);
```
fenik17: Use `ThreadLocalRandom`.

codec/src/test/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,114 @@
...
+      slices.add(byteList.subList(byteInd, sliceIndex));
+      byteInd = sliceIndex;
+    }
+    // System.out.println(String.format("Slice from %d through %d", byteInd, byteList.size()));
+    slices.add(byteList.subList(byteInd, byteList.size()));
+
+    return slices;
+  }
+
+  public static int nextInt(final int startInclusive, final int endExclusive) {
```
fenik17: private?

codec/src/test/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,114 @@
...
+        channel.writeInbound(Unpooled.copiedBuffer(Bytes.toArray(chunk)));
+      }
+    } else {
+      channel.writeInbound(Unpooled.copiedBuffer(frame, charset));
+    }
+    ByteBuf in = (ByteBuf) channel.readInbound();
+    Assert.assertEquals(value, in.toString(charset));
+  }
+
+  public static List<List<Byte>> getRandomByteSlices(byte[] bytes) {
```
fenik17: private?

fenik17:
@jeff303 use a Netty's code style please..

jeff303:
@fenik17 , apologies for not following the style.  I have applied the code style formatting now as per [this doc](http://netty.io/wiki/setting-up-development-environment.html#wiki-h3-6).

I have introduced a new abstract base class for my new class, as well as LengthFieldBasedFrameDecoder, to put common methods/state in.

codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,160 @@
...
+        return DelimiterBasedFrameDecoder.indexOf(haystack, needle);
+    }
+
+    /**
+     * Create a frame out of the {@link ByteBuf} and return it.
+     *
+     * @param ctx the {@link ChannelHandlerContext} which this {@link ByteToMessageDecoder} belongs to
+     * @param in the {@link ByteBuf} from which to read data
+     *
+     * @return frame           the {@link ByteBuf} which represent the frame or {@code null} if no frame could be
```
fenik17: excess spaces

codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,160 @@
...
+        this.lengthAdjustment = lengthAdjustment;
+        this.trimLengthString = trimLengthString;
+    }
+
+    /**
+     * Returns the number of bytes between the readerIndex of the haystack and
+     * the first needle found in the haystack.  -1 is returned if no needle is
+     * found in the haystack.
+     */
+    private static int indexOf(ByteBuf haystack, ByteBuf needle) {
```
fenik17: suggest make it protected and non-static.

codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,160 @@
...
+                        ),
+                        e
+                );
+            }
+
+            if (frameLength < 0) {
+                throw new CorruptedFrameException("negative pre-adjustment length field: " + frameLength);
+            }
+
+            frameLength += lengthAdjustment;
```
fenik17: Here may be overflowing.

codec/src/main/java/io/netty/handler/codec/BaseLengthBasedFrameDecoder.java
```diff
@@ -0,0 +1,85 @@
...
+     * <p>
+     * If you are sure that the frame and its content are not accessed after
+     * the current {@link #decode(ChannelHandlerContext, ByteBuf)}
+     * call returns, you can even avoid memory copy by returning the sliced
+     * sub-region (i.e. <tt>return buffer.slice(index, length)</tt>).
+     * It's often useful when you convert the extracted frame into an object.
+     * Refer to the source code of {@link ObjectDecoder} to see how this method
+     * is overridden to avoid memory copy.
+     */
+    protected ByteBuf extractFrame(ChannelHandlerContext ctx, ByteBuf buffer, int index, int length) {
```
netkins: MAJOR Remove this unused method parameter "ctx". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,160 @@
...
+ * Uses the same approach as the {@link DelimiterBasedFrameDecoder} to find delimiter in input
+ */
+public class DelimitedLengthFieldBasedFrameDecoder extends BaseLengthBasedFrameDecoder {
+
+    private final int lengthAdjustment;
+    private final Charset lengthFieldCharset;
+    private final ByteBuf delimiter;
+    private final boolean trimLengthString;
+
+    private boolean discardingTooLongFrame;
```
netkins: MAJOR "discardingTooLongFrame" is the name of a field in "BaseLengthBasedFrameDecoder". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2387)

codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,160 @@
...
+    private long bytesToDiscard;
```
netkins: MAJOR "bytesToDiscard" is the name of a field in "BaseLengthBasedFrameDecoder". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2387)

codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,160 @@
...
+
+            //consume length field and delimiter bytes
+            in.skipBytes(delimIndex + delimiter.capacity());
+
+            //consume delimiter bytes
+            consumingLength = false;
+        }
+
+        if (frameLength > maxFrameLength) {
+            long discard = frameLength - in.readableBytes();
```
netkins: MAJOR Cast one of the operands of this substraction operation to a "long". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2184)

codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,180 @@
...
+            in.skipBytes(localBytesToDiscard);
+            bytesToDiscard -= localBytesToDiscard;
+            this.bytesToDiscard = bytesToDiscard;
+
+            failIfNecessary(false);
+
+            // if exception was not thrown, reader has been advanced by readable bytes already, so return null to
+            // continue processing too long frame
+            // TODO: check whether also needed in LengthFieldBasedFrameDecoder; I believe it is
+            return null;
```
jeff303: @fenik17 , I believe this return statement will also need to be added to make LengthFieldBasedFrameDecoder work properly, even before my commit.  I will come up with a test case when I get a chance.

codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,179 @@
...
+            int localBytesToDiscard = (int) Math.min(bytesToDiscard, in.readableBytes());
+            in.skipBytes(localBytesToDiscard);
+            bytesToDiscard -= localBytesToDiscard;
+            this.bytesToDiscard = bytesToDiscard;
+
+            failIfNecessary(false);
+
+            // if exception was not thrown, reader has been advanced by readable bytes already, so return null to
+            // continue processing too long frame
+            // TODO: check whether also needed in LengthFieldBasedFrameDecoder; I believe it is
```
netkins: INFO Complete the task associated to this TODO comment. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1135)

codec/src/test/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,169 @@
...
+    private static int nextInt(final int startInclusive, final int endExclusive) {
+        if (startInclusive == endExclusive) {
+            return startInclusive;
+        }
+
+        return startInclusive + ThreadLocalRandom.current().nextInt(endExclusive - startInclusive);
+    }
+
+    @Test
+    public void testDelimitedLengths() throws Exception {
```
fenik17: "test" prefix in the test-method name is redundant for JUnit (newer v.3).

codec/src/test/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.List;
+
+public class DelimitedLengthFieldBasedFrameDecoderTest {
+
+    private static List<List<Byte>> getRandomByteSlices(byte[] bytes) {
+        List<Byte> byteList = Bytes.asList(bytes);
+
+        int numSlices = nextInt(2, 10);
+        List<Integer> sliceIndexes = new ArrayList<Integer>();
+        for (int i = 1; i <= numSlices; i++) {
```
fenik17: This is somehow unusual - start a cycle from a 1 :)

codec/src/test/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+        try {
+            ch.writeInbound(Unpooled.copiedBuffer(longString, charset));
+            Assert.fail("TooLongFrameException should have been thrown");
+        } catch (TooLongFrameException ignored) {
+            //ignored
+        }
+        Assert.assertNull(ch.readInbound());
+
+        ch.close();
```
fenik17: `assertFalse(ch.finishAndReleaseAll());`

codec/src/test/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,169 @@
...
+        List<Byte> byteList = Bytes.asList(bytes);
+
+        int numSlices = nextInt(2, 10);
+        List<Integer> sliceIndexes = new ArrayList<Integer>();
+        for (int i = 1; i <= numSlices; i++) {
+            sliceIndexes.add(nextInt(0, bytes.length));
+        }
+        Collections.sort(sliceIndexes);
+
+        List<List<Byte>> slices = new LinkedList<List<Byte>>();
```
fenik17: Better use an `ArrayList` (whith initial size).

codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,179 @@
...
+        if (delimiter == null) {
+            throw new NullPointerException("delimiter");
+        }
+        if (!delimiter.isReadable()) {
+            throw new IllegalArgumentException("empty delimiter");
+        }
+
+        this.delimiter = delimiter.slice(delimiter.readerIndex(), delimiter.readableBytes());
+        this.lengthFieldCharset = lengthFieldCharset;
+
```
fenik17: nit: remove empty line.

codec/src/test/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,169 @@
...
+            }
+        }
+        if (!expectFrameTooLarge) {
+            ByteBuf in = channel.readInbound();
+            Assert.assertEquals(value, in.toString(charset));
+            in.release();
+        }
+    }
+
+    private String makeFrame(String value, Charset charset) {
```
fenik17: `static`

codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,179 @@
...
+            return null;
+        }
+
+        if (consumingLength) {
+            int delimIndex = indexOf(in, delimiter);
+            if (delimIndex < 0) {
+                return null;
+            }
+
+            final String lengthStr = in.toString(in.readerIndex(), delimIndex, lengthFieldCharset);
```
fenik17: Second parameter in the `toString()` is a `length`:
`in.toString(in.readerIndex(), delimIndex - in.readerIndex(), lengthFieldCharset);`
It's not catched in the tests?

I still suggest to extract the reading and parsing of a frame length into separate method. And it makes sense to add the check for a big difference between `in.readerIndex()` and `delimIndex` (to prevent a possible DDoS):

```java
// Max size of the frame length field, even taking into a whitespaces presence
private static final int MAX_LENGTH_OF_LENGTH_FIELD = 64;

...

protected int readFrameLength(ByteBuf in, int from, int to) {
    final int lengthStrSize = from - to;
    if (lengthStrSize < 1 || lengthStrSize > MAX_LENGTH_OF_LENGTH_FIELD) {
        throw new CorruptedFrameException(String.format(
            "wrong size of the pre-adjustment length field: %d (expected between 1 and %d)",
            lengthStrSize, MAX_LENGTH_OF_LENGTH_FIELD));
    }

    final String lengthStr = in.toString(from, lengthStrSize, lengthFieldCharset);
    final int lengthValue;
    try {
        lengthValue = Integer.parseInt(trimLengthString? lengthStr.trim() : lengthStr);
    } catch (NumberFormatException nfe) {
        throw new CorruptedFrameException(
            String.format("invalid length field decoded (in %s charset): %s",
                lengthFieldCharset.name(), lengthStr), nfe);
    }
	
    if (lengthValue < 0) {
        throw new CorruptedFrameException("negative pre-adjustment length field: " + lengthStr);
    }
	
    return lengthValue;
}

...

frameLength = readFrameLength(in, in.readerIndex(), delimIndex);
```

codec/src/test/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,169 @@
...
+            Assert.fail("TooLongFrameException should have been thrown");
+        } catch (TooLongFrameException ignored) {
+            //ignored
+        }
+        Assert.assertNull(ch.readInbound());
+
+        ch.close();
+    }
+
+    private EmbeddedChannel getTestChannel(Charset charset, int maxFrameLength, int lengthAdjustment,
```
fenik17: static

codec/src/test/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,169 @@
...
+                        lengthAdjustment,
+                        failFast,
+                        Unpooled.copiedBuffer(" ", charset),
+                        charset,
+                        true
+                )
+        );
+    }
+
+    private void writeStringAndAssert(EmbeddedChannel channel, String value, Charset charset,
```
fenik17: static

codec/src/main/java/io/netty/handler/codec/marshalling/MarshallingDecoder.java
```diff
@@ -81,7 +81,7 @@ protected Object decode(ChannelHandlerContext ctx, ByteBuf in) throws Exception
     }
 
     @Override
-    protected ByteBuf extractFrame(ChannelHandlerContext ctx, ByteBuf buffer, int index, int length) {
+    protected ByteBuf extractFrame(ByteBuf buffer, int index, int length) {
```
fenik17: It's API'a change. We're should to leave old method. 
/cc @normanmaurer

codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
jeff303: delimIndex is actually already a length measure.  It comes from indexOf, which effectively finds the index of the delimiter, then subtracts the readerIndex.  I am adding a new unit test which asserts this (and fails when subtracting the readerIndex).

I'm not sure I follow about the DDoS.  Can you please elaborate a bit there?

codec/src/main/java/io/netty/handler/codec/marshalling/MarshallingDecoder.java
jeff303: Hopefully understood this correctly.  Added the previous method back and delegate call to the new method.

jeff303:
I looked into the TODO previously mentioned, regarding the existing LengthFieldBasedFrameDecoder return statement.  It turns out, there is no bug there.   Here is the relevant snippet.  The final return statement will always return null in the case I was concerned with (i.e. continuing to consume a too-long frame where there are still more `bytesToDiscard`.

```
        if (discardingTooLongFrame) {
            long bytesToDiscard = this.bytesToDiscard;
            int localBytesToDiscard = (int) Math.min(bytesToDiscard, in.readableBytes());
            in.skipBytes(localBytesToDiscard);
            bytesToDiscard -= localBytesToDiscard;
            this.bytesToDiscard = bytesToDiscard;

            failIfNecessary(false);
        }

        if (in.readableBytes() < lengthFieldEndOffset) {
            return null;
        }
```

I added a new test method to assert this behavior: `LengthFieldBasedFrameDecoderTest#goodFrameAfterTooLongFrame`, and I know there is probably a better way to do the `assertByteBufsMatch` method, or ideally an existing method.

codec/src/main/java/io/netty/handler/codec/BaseLengthBasedFrameDecoder.java
```diff
@@ -0,0 +1,103 @@
...
+     * If you are sure that the frame and its content are not accessed after
+     * the current {@link #decode(ChannelHandlerContext, ByteBuf)}
+     * call returns, you can even avoid memory copy by returning the sliced
+     * sub-region (i.e. <tt>return buffer.slice(index, length)</tt>).
+     * It's often useful when you convert the extracted frame into an object.
+     * Refer to the source code of {@link ObjectDecoder} to see how this method
+     * is overridden to avoid memory copy.
+     */
+
+    protected ByteBuf extractFrame(ChannelHandlerContext ctx, ByteBuf buffer, int index, int length) {
```
netkins: MAJOR Remove this unused method parameter "ctx". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
fenik17: > delimIndex is actually already a length measure. 

Yep, I see at now. The variable name misled me. 
IMO, we should rename method `indexOf()` or change his behavior to match the name.

> I'm not sure I follow about the DDoS. Can you please elaborate a bit there?

Theoretically possible attack on memory (OOME) if the frame size is large and long time will come data without a separator. When the delimiter comes, we'll have to copy a large chunk of buffer into a string and try to convert it into a number. If there is still a space in the beginning or at the end, we get 2 large strings in the heap.
Maybe it's just my paranoia... It seems to me that in such places there are no superfluous checks )

codec/src/test/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,185 @@
...
+        writeStringAndAssert(ch, v3, charset, true, true);
+
+        writeStringAndAssert(ch, v1, charset, false, false);
+
+        Assert.assertFalse(ch.finishAndReleaseAll());
+    }
+
+    @Test
+    public void multipleFramesSingleBuffer() throws Exception {
+        Charset charset = CharsetUtil.ISO_8859_1;
```
fenik17: It makes sense also to testing with non one-byte encoding (for example UTF-16/UTF-8).

codec/src/main/java/io/netty/handler/codec/marshalling/MarshallingDecoder.java
fenik17: correctly. But better to swap calls and make new method as `static`:

```java
protected static ByteBuf extractFrame(ByteBuf buffer, int index, int length) {
    return buffer.slice(index, length);
}

@Override
protected ByteBuf extractFrame(ChannelHandlerContext ctx, ByteBuf buffer, int index, int length) {
    return extractFrame(buffer, index, length);
}
```

fenik17: And remove the method `extractFrame(ByteBuf buffer, int index, int length)` from `MarshallingDecoder`. It's present in the `BaseLengthBasedFrameDecoder`.

codec/src/main/java/io/netty/handler/codec/BaseLengthBasedFrameDecoder.java
```diff
@@ -0,0 +1,103 @@
...
+            // Reset to the initial state and tell the handlers that
+            // the frame was too large.
+            long tooLongFrameLength = this.tooLongFrameLength;
+            this.tooLongFrameLength = 0;
+            discardingTooLongFrame = false;
+            if (!failFast || firstDetectionOfTooLongFrame) {
+                fail(tooLongFrameLength);
+            }
+            return true;
+        } else {
```
fenik17: nit: the `else` statement is redundant.

codec/src/main/java/io/netty/handler/codec/marshalling/MarshallingDecoder.java
fenik17: Looked into. At now a param `ctx` is needless. I think it will be correct to deprecate a current `extractFrame` (with `ctx`) in the `LengthFieldBasedFrameDecoder`, make new method without `ctx`, call a new method from the old, and override only new method in the `MarshallingDecoder`.

jeff303:
Added a new check that I believe will address the DDoS concern, at least partially (i.e. ensure that delimiter relative offset is not more than max possible byte length of length field given the encoding), and a unit test for that check.

Regarding the `extractFrame` method, I kept the deprecated old version only in `MarshallingDecoder`, since that is what I understood needed to remain unchanged.

@fenik17 , still need some assistance with the multibyte issue.  Unignore the `DelimitedLengthFieldBasedFrameDecoderTest#multiByteStrings` method, and set a breakpoint in `ByteBufUtil#encodeString0` with this condition: `src.length() == 1 && charset.displayName().equals("UTF-16")`

It seems that this encoding (which is actually of the separator string) is writing 4 bytes for the single space character, which is unexpected by me (should be 2 bytes for a single space in UTF-16).  Not sure why that is happening, but it's causing the delimiter to not be found at decode time.

codec/src/main/java/io/netty/handler/codec/marshalling/MarshallingDecoder.java
```diff
@@ -80,8 +80,9 @@ protected Object decode(ChannelHandlerContext ctx, ByteBuf in) throws Exception
         }
     }
 
-    @Override
-    protected ByteBuf extractFrame(ChannelHandlerContext ctx, ByteBuf buffer, int index, int length) {
-        return buffer.slice(index, length);
+    @Deprecated
+    protected ByteBuf extractFrame(@SuppressWarnings("unused") ChannelHandlerContext ctx, ByteBuf buffer, int index,
```
netkins: MAJOR Add the missing @deprecated Javadoc tag. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AMissingDeprecatedCheck)
MAJOR Remove this unused method parameter "ctx". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

fenik17:
> Regarding the extractFrame method, I kept the deprecated old version only in MarshallingDecoder, since that is what I understood needed to remain unchanged.

No.. MarshallingDecoder just override the method from LengthFieldBasedFrameDecoder. To save an API compatibility we're should leave the both old methods in LengthFieldBasedFrameDecoder and MarshallingDecoder. I meant a such new structure:

```java
class BaseLengthBasedFrameDecoder 

protected ByteBuf extractFrame(ByteBuf buffer, int index, int length) {
    return buffer.retainedSlice(index, length);
}

-----------------------
class LengthFieldBasedFrameDecoder

/**
 * @deprecated use {@link BaseLengthBasedFrameDecoder#extractFrame(ChannelHandlerContext, ByteBuf, int, int)}
 */
@Deprecated
protected ByteBuf extractFrame(ChannelHandlerContext ignored, ByteBuf buffer, int index, int length) {
    return extractFrame(buffer, index, length);
}

-----------------------
class MarshallingDecoder

@Override
protected ByteBuf extractFrame(ByteBuf buffer, int index, int length) {
    return buffer.slice(index, length);
}

@Override
protected ByteBuf extractFrame(ChannelHandlerContext ignored, ByteBuf buffer, int index, int length) {
    return extractFrame(buffer, index, length);
}
```

Method from the  `BaseLengthBasedFrameDecoder` should be non static. And link to the `ObjectDecoder` in his javadoc is inactual..


> need some assistance with the multibyte issue

Tests failed because encoded in UTF-16 and UTF-32 strings have a BOM (Byte order mark) at the start. But, even if you take the UTF-16LE (UTF-32LE/BE) encoding which haven't a BOM ... tests will fail due to a `frameLength` field contains the number of characters but not namber of bytes. It also will fail with UTF-8 and non one-byte characters (I tried string "4 блин"). Thus, a current realization works fine only with a latin encoding..

codec/src/main/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoder.java
```diff
@@ -0,0 +1,178 @@
...
+
+        // the frame is now entirely present, reset state vars
+        consumingLength = true;
+        frameLength = 0;
+
+        return frame;
+    }
+
+    private int extractFrameLength(ByteBuf in, int delimIndex) {
+        if (delimIndex > maxPossibleLengthInBytes) {
```
fenik17: This does not take into account possible whitespaces which are assumed by the `trimLengthString` param.

fenik17:
@jeff303 what do you think about the encodings? Do you intend to support not only one-byte encodings? For the fixed-sized encodings it's will be not very difficult. For the dynamic-sized there are some possible solutions...

jeff303:
@fenik17 , I would like to support them if I can figure out how.  I got tied up last week with some production issues, etc. so didn't get any time to spend on this.  Planning to pick it up again this week.  Any suggestions you have would be appreciated.

fenik17:
Perhaps the simplest and universal way is to convert all to String..

jeff303:
@fenik17 , we were overcomplicating.  In my original use case (ex: RFC 6587), the delimited length field actually specifies message size in bytes, not characters.  So it's a moot point (i.e. the sender of the frames is responsible for correctly calculating the byte lengths independent of encoding).

To be very precise, for my purpose (this RFC), I don't need to support multibyte encodings because it only permits single byte encodings.  However, I feel this may be useful in general so I started to get it working.  UTF-16 variants are working now, after doing some tricks with the byte order marks, but UTF-32 is still failing.  I may be OK with not supporting UTF-32, since it's not typically used for transmission anyway.

fenik17:
In this case, it's better to not support the encodings. I don't see the point of writing slow, over-complicated and hard-to-support code that no one will use. Work only with bytes, a `MSG-LEN` part consider by `NONZERO-DIGIT` (%d49-57). Let the user take care of the need to form the correct delimiter - cutting out the BOM or something else. And it will be great if you add an example of using new decoder based on rfc6587..
This is only my humble opinion :)

codec/src/main/java/io/netty/handler/codec/BaseLengthBasedFrameDecoder.java
```diff
@@ -0,0 +1,110 @@
...
+    protected ByteBuf extractFrame(ByteBuf buffer, int index, int length) {
+        return buffer.retainedSlice(index, length);
+    }
+
+    /**
+     * @deprecated - use {@link #extractFrame(ByteBuf, int, int)} instead, since the
+     * {@link ChannelHandlerContext} is not actually needed for implementation
+     */
+    @Deprecated
+    protected ByteBuf extractFrame(ChannelHandlerContext ignored, ByteBuf buffer, int index, int length) {
```
netkins: MAJOR Remove this unused method parameter "ignored". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

codec/src/main/java/io/netty/handler/codec/marshalling/MarshallingDecoder.java
```diff
@@ -80,8 +80,17 @@ protected Object decode(ChannelHandlerContext ctx, ByteBuf in) throws Exception
         }
     }
 
-    @Override
-    protected ByteBuf extractFrame(ChannelHandlerContext ctx, ByteBuf buffer, int index, int length) {
+    protected ByteBuf extractFrame(ByteBuf buffer, int index, int length) {
```
netkins: MAJOR Add the "@Override" annotation above this method signature [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1161)

codec/src/main/java/io/netty/handler/codec/marshalling/MarshallingDecoder.java
```diff
@@ -80,8 +80,17 @@ protected Object decode(ChannelHandlerContext ctx, ByteBuf in) throws Exception
...
         return buffer.slice(index, length);
     }
+
+    /**
+     * @deprecated - use {@link #extractFrame(ByteBuf, int, int)} instead, since the
+     * {@link ChannelHandlerContext} is not actually needed for implementation
+     */
+    @Deprecated
+    protected ByteBuf extractFrame(@SuppressWarnings("unused") ChannelHandlerContext ignored, ByteBuf buffer, int index,
```
netkins: MAJOR Add the "@Override" annotation above this method signature [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1161)

codec/src/main/java/io/netty/handler/codec/DelimiterBasedFrameDecoder.java
```diff
@@ -16,8 +16,10 @@
 package io.netty.handler.codec;
 
 import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufUtil;
```
netkins: MINOR Remove this unused import 'io.netty.buffer.ByteBufUtil'. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

codec/src/main/java/io/netty/handler/codec/DelimiterBasedFrameDecoder.java
```diff
@@ -16,8 +16,10 @@
...
 import io.netty.channel.ChannelHandlerContext;
 
+import java.nio.charset.Charset;
```
netkins: MINOR Remove this unused import 'java.nio.charset.Charset'. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

codec/src/main/java/io/netty/handler/codec/BaseLengthBasedFrameDecoder.java
```diff
@@ -0,0 +1,111 @@
...
+    protected ByteBuf extractFrame(ByteBuf buffer, int index, int length) {
+        return buffer.retainedSlice(index, length);
+    }
+
+    /**
+     * @deprecated - use {@link #extractFrame(ByteBuf, int, int)} instead, since the
+     * {@link ChannelHandlerContext} is not actually needed for implementation
+     */
+    @Deprecated
+    protected ByteBuf extractFrame(@SuppressWarnings("unused") ChannelHandlerContext ignored, ByteBuf buffer,
```
netkins: MAJOR Remove this unused method parameter "ignored". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

jeff303:
@fenik17 , I did as you suggested and got rid of all the charset support, assumed single byte encoding for length field digits (via UTF-8).  I can't quite figure out how to access the SonarQube report - is this just running SonarLint?

fenik17:
@jeff303 sonarqube reports through a @netkins [comments](https://github.com/netty/netty/pull/6554#discussion_r112738182).

codec/src/main/java/io/netty/handler/codec/LengthFieldBasedFrameDecoder.java
```diff
@@ -354,7 +338,8 @@ protected final void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object>
      * @return  frame           the {@link ByteBuf} which represent the frame or {@code null} if no frame could
      *                          be created.
      */
-    protected Object decode(ChannelHandlerContext ctx, ByteBuf in) throws Exception {
+    protected Object decode(@SuppressWarnings({ "unused", "squid:S1172" }) ChannelHandlerContext ctx, ByteBuf in)
```
netkins: MAJOR Add the "@Override" annotation above this method signature [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1161)

jeff303:
@fenik17 , thanks, finally resolved the issues.  It took me a while to realize that it simply left the old comments up if they were still relevant with new commit.

codec/src/main/java/io/netty/handler/codec/BaseLengthBasedFrameDecoder.java
```diff
@@ -0,0 +1,112 @@
...
+    protected ByteBuf extractFrame(ByteBuf buffer, int index, int length) {
+        return buffer.retainedSlice(index, length);
+    }
+
+    /**
+     * @deprecated - use {@link #extractFrame(ByteBuf, int, int)} instead, since the
+     * {@link ChannelHandlerContext} is not actually needed for implementation
+     */
+    @Deprecated
+    protected ByteBuf extractFrame(@SuppressWarnings({ "unused", "squid:S1172" }) ChannelHandlerContext ignored,
```
fenik17: There is no need to solve all of CI reports (by @netkins). They are often false positive.. I think, supressing here is redundant (and in other deprecated methods below).

fenik17:
@jeff303 I will take a look in a while.

normanmaurer:
Sorry it took so long but this one totally fall through the cracks. I will try to review this week.

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -1242,5 +1242,16 @@ private static boolean isUtf8(ByteBuf buf, int index, int length) {
         return true;
     }
 
+    public static int lengthOfByteOrderMark(ByteBuf buf, Charset charset) {
+        final byte[] bom = "\uFEFF".getBytes(charset);
```
carl-mastrangelo: This is surprisingly expensive to create, at least two allocs if memory serves.  You should make this a final static var for UTF-8 and ASCII

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -1242,5 +1242,16 @@ private static boolean isUtf8(ByteBuf buf, int index, int length) {
...
+        int length = 0;
+        for (int i = 0; i < bom.length && i < buf.capacity(); i++) {
```
carl-mastrangelo: capacity is the total amount of bytes the buffer has, shouldn't this be just the readable bytes?

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -1242,5 +1242,16 @@ private static boolean isUtf8(ByteBuf buf, int index, int length) {
         return true;
     }
 
+    public static int lengthOfByteOrderMark(ByteBuf buf, Charset charset) {
+        final byte[] bom = "\uFEFF".getBytes(charset);
+        int length = 0;
```
carl-mastrangelo: If you define i outside of the loop, you can return it directly. and skip the length field.

codec/src/main/java/io/netty/handler/codec/BaseLengthBasedFrameDecoder.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    protected abstract Object decode(@SuppressWarnings({ "unused", "squid:S1172" }) ChannelHandlerContext ctx,
+                                     ByteBuf in) throws Exception;
+
+    protected void fail(long frameLength) {
+        resetStateOnTooLongFrame();
+        throw new TooLongFrameException(
+                "Adjusted frame length exceeds " + maxFrameLength +
+                (frameLength > 0? ": " + frameLength + " - discarded" : " - discarding"));
```
carl-mastrangelo: add space after `0`

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
jeff303: Done

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
jeff303: Done

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
jeff303: I think that's right.  Done.

codec/src/main/java/io/netty/handler/codec/BaseLengthBasedFrameDecoder.java
jeff303: Done

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -1312,5 +1316,24 @@ private static boolean isUtf8(ByteBuf buf, int index, int length) {
         return true;
     }
 
+    public static int lengthOfByteOrderMark(ByteBuf buf, Charset charset) {
```
carl-mastrangelo: can you clarify what this function does?   If the BOM is not found, this can return both 0 and 1.   Should it return -1 instead?    

Also, public methods should have a javadoc.

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -66,6 +66,10 @@ protected CharBuffer initialValue() throws Exception {
 
     static final ByteBufAllocator DEFAULT_ALLOCATOR;
 
+    private static final String BOM_STR = "\uFEFF";
+    private static final byte[] BOM_UTF8 = BOM_STR.getBytes(CharsetUtil.UTF_8);
+    private static final byte[] BOM_US_ASCII = BOM_STR.getBytes(CharsetUtil.US_ASCII);
```
carl-mastrangelo: Can you sanity check that this isn't just question marks like `??`

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
jeff303: I have added Javadoc with explanation.  As of now, this method is only used from the test code I added (`DelimitedLengthFieldBasedFrameDecoderTest`), so if desired I can refactor to only keep it there.  This was the place where it seemed like it should live (given my limited exposure to the codebase).

Updated the behavior to return -1 if BOM is not found at the beginning of the buffer.

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
jeff303: I'm afraid I don't quite follow.  You mean check in the tests?

carl-mastrangelo: Oh, I mean just print it out and see what the chars actually are.

codec/src/main/java/io/netty/handler/codec/DelimiterBasedFrameDecoder.java
```diff
@@ -311,8 +311,13 @@ private void fail(long frameLength) {
...
      * the first needle found in the haystack.  -1 is returned if no needle is
      * found in the haystack.
+     *
+     * @param haystack the buffer to search for an occurence of the needle
+     * @param needle the buffer to find in the haystack
+     *
+     * @return the first relative byte index of the needle found within the haystack, or -1 if not found
      */
-    private static int indexOf(ByteBuf haystack, ByteBuf needle) {
+    protected static int indexWithin(ByteBuf haystack, ByteBuf needle) {
```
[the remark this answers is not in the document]
jeff303: I'm not entirely sure, but the behavior is definitely different.  For my PR, I did nothing but rename this method since I found the previous name confusing (but would be happy to change it back).  If I try switching out the invocation of this method with `ByteBufUtil.indexOf`, (and flip the args, since they are in the opposite order for that method), the tests no longer pass and I haven't yet had the time to dig into why.

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -1312,5 +1316,36 @@ private static boolean isUtf8(ByteBuf buf, int index, int length) {
         return true;
     }
 
+    /**
+     * Determines the length of the byte-order mark as it occurs at the beginning of the
+     * specified {@link ByteBuf} input.  Returns the length if found, otherwise -1.
```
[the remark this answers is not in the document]
jeff303: This seems reasonable to me.  Will change.

buffer/src/test/java/io/netty/buffer/ByteBufUtilLengthOfBomTest.java
```diff
@@ -0,0 +1,72 @@
...
+    private final Charset charset;
+    private final int expectedBomLength;
+
+    public ByteBufUtilLengthOfBomTest(Charset charset, int expectedBomLength) {
+        this.charset = charset;
+        this.expectedBomLength = expectedBomLength;
+    }
+
+    @Test
+    public void lengthOfBom() throws IOException {
```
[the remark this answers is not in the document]
jeff303: Will do.

codec/src/test/java/io/netty/handler/codec/DelimitedLengthFieldBasedFrameDecoderTest.java
```diff
@@ -0,0 +1,239 @@
...
+        if (!expectFrameTooLarge) {
+            ByteBuf in = channel.readInbound();
+            Assert.assertEquals(value, in.toString(CHARSET));
+            in.release();
+        }
+    }
+
+    private static String makeFrame(String value) {
+        final byte[] bytes = value.getBytes(CHARSET);
+        final int bomLength = ByteBufUtil.lengthOfByteOrderMark(Unpooled.copiedBuffer(bytes), CHARSET);
```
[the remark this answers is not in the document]
jeff303: When I use `Unpooled.copiedBuffer` to create a `ByteBuf` from a String in a particular charset (i.e. this overload: `Unpooled#copiedBuffer(java.lang.CharSequence, java.nio.charset.Charset)`), the BOM is inserted into the resulting bytes.  In the data I'm interested in consuming (i.e. RFC 6587), the BOM is not actually part of the data.  Hence, the "length" field should not include the length of the BOM.  Therefore, when I build the frame from a String, I need to subtract the length of the BOM from the total byte length in order to build this portion of the test data.

codec/src/main/java/io/netty/handler/codec/marshalling/MarshallingDecoder.java
```diff
@@ -81,7 +81,18 @@ protected Object decode(ChannelHandlerContext ctx, ByteBuf in) throws Exception
     }
 
     @Override
-    protected ByteBuf extractFrame(ChannelHandlerContext ctx, ByteBuf buffer, int index, int length) {
+    protected ByteBuf extractFrame(ByteBuf buffer, int index, int length) {
```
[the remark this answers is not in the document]
jeff303: I made this change in response to comments from @fenik17 above.  Those were many months ago so it's not too fresh in my mind, but I am happy to change this back if that's the consensus (and you are correct that my change does not in any way concern this).

normanmaurer:
Sorry for dropping the ball.. @jeff303 are you still interested in this change if so I will try to get it over the finish line.

jeff303:
> Sorry for dropping the ball.. @jeff303 are you still interested in this change if so I will try to get it over the finish line.

Yes, I'm still interested.  I will rebase and update here.  Also, is the target branch still `4.1`, or something else?

normanmaurer:
@jeff303 yes 4.1 please let me know once you did and I will have another look.

jeff303:
> @jeff303 yes 4.1 please let me know once you did and I will have another look.

Should be rebased and up to date now.  Thanks.

jeff303:
Rebased onto 4.1 again.

normanmaurer:
Sorry but never where able to get this over the finish line and it seems like there is not much interesting the community (no-one else needed something like this it seems). Let me close this for now. Thanks for all your work and sorry for not merging it

PR #6556, opened 2017-03-21
EmptyByteBuf allows writing ByteBufs with 0 readable bytes

bryce-anderson:
Motivation:

The contract of `ByteBuf.writeBytes(ByteBuf src)` is such that it will throw an `IndexOutOfBoundsException` if `src.readableBytes()` is greater than `this.writableBytes()`. The EmptyByteBuf class will throw the exception, even if the source buffer has zero readable bytes, in violation of the contract.

Modifications:

Use the helper method `checkLength(..)` to check the length and throw the exception, if appropriate.

Result:

Conformance with the stated behavior of ByteBuf.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/EmptyByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/EmptyByteBuf.java
@@ -760,7 +760,7 @@ public final class EmptyByteBuf extends ByteBuf {
 
     @Override
     public ByteBuf writeBytes(ByteBuf src) {
-        throw new IndexOutOfBoundsException();
+        return checkLength(src.readableBytes());
     }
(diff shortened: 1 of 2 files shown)


normanmaurer:
@bryce-anderson thanks a lot!

normanmaurer:
Will pull in once the CI finishes its run.

bryce-anderson:
@normanmaurer, awesome. Thanks for the quick review!

normanmaurer:
Cherry-picked into 4.1 (aa2f16f3143e46412e3c043b097e0f70f5026196) and 4.0 (f235563b4de1079f6d51e911a56d76dc865f2e3e)

Issue #87, opened 2011-11-30
Build Issue with maven-antrun-plugin: "Plugin execution not covered by lifecycle configuration"


veebs:
Just installed Eclipse Indigo Service Release 1

It comes with M2E - Maven Integration for Eclipse 1.1

When I load up the Netty project, I get the following error:

```
Plugin execution not covered by lifecycle configuration: org.apache.maven.plugins:maven-antrun-plugin:1.7:run (execution: write-version, phase: validate)   pom.xml /netty  line 320    Maven Project Build Lifecycle Mapping Problem
```

```
      <plugin>
        <artifactId>maven-antrun-plugin</artifactId>
        <version>1.7</version>
        <executions>
          <execution>
            <id>write-version</id>
            <phase>validate</phase>
            <goals>
              <goal>run</goal>
            </goals>
         ...
        </execution>
      </plugin>
```

Seems like `maven-antrun-plugin` does not like the the `validate` phase.  

The error goes away if I set the phase to `package` - but I am not sure if this is what is required.

chandu1988:
??????

chandu1988:
Description Resource	Path	Location	Type
Mapping file orm.xml does not have supported content for this JPA platform.	orm.xml	/spring-social-tutorial/src/main/java/META-INF	 JPA Problem

JavaHelper:
Could someone please guide on this issue http://stackoverflow.com/questions/35523664/java-util-concurrent-executionexception-org-apache-catalina-lifecycleexception/35923065#35923065 ?

PR #6563, opened 2017-03-24
non-latin character broken on HttpHeader by HttpObjectDecoder

ByunghyunBang:
I encountered non-latin character broken on HttpHeader by HttpObjectDecoder.

Steps to reproduce:
1. request with non-ascii character in Request Header
$ curl -v http://localhost/testurl -H"Host:www.foo.com" -H"foo:한글中文"
2. Then, request.headers().get("foo") have broken characters. "????"

$ java -version
java version "1.8.0_101"
Java(TM) SE Runtime Environment (build 1.8.0_101-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.101-b13, mixed mode)

Operating system: Ubuntu Linux 16.04 64-bit

$ uname -a
Linux myhost 4.4.0-66-generic #87-Ubuntu SMP Fri Mar 3 15:29:05 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

Motivation:

Modification:

Result:

Fixes #<GitHub issue number>. 

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectDecoder.java
@@ -805,7 +805,7 @@ public abstract class HttpObjectDecoder extends ByteToMessageDecoder {
 
         @Override
         public boolean process(byte value) throws Exception {
-            char nextByte = (char) value;
+            char nextByte = (char) (value & 0xFF);
             if (nextByte == HttpConstants.CR) {
                 return true;
             }


normanmaurer:
@ByunghyunBang thats expected as headers are US_ASCII as defined by rfc

ByunghyunBang:
I am developing http proxy with netty.
As you said, header name should be ASCII charactor, but header values can be from various encoding in act.

Currently netty is receiving HTTP request by ByteBuf and store it as "String" on HttpObjectDecoder. During this operation, all character on ByteBuf is moving to char[] on String without breaking encoding.
But in process() function, type casting from byte to char is not consider msb (sign-bit). So the value over 127 can be casted wrong value. (ex :  ec in byte -> ffec in char)

Please think this point.

fenik17:
See the [RFC7230](https://tools.ietf.org/html/rfc7230):

>  Historically, HTTP has allowed field content with text in the
>   ISO-8859-1 charset [ISO-8859-1], supporting other charsets only
>   through use of [RFC2047] encoding.  In practice, most HTTP header
>   field values use only a subset of the US-ASCII charset [USASCII].
>   Newly defined header fields SHOULD limit their field values to
>   US-ASCII octets.  A recipient SHOULD treat other octets in field
>   content (obs-text) as opaque data.

fenik17:
so.. maybe Netty should be just ignoring a not-ascii simbols.

fenik17:
@ByunghyunBang try that: 
`curl -v http://localhost/testurl -H "Host:www.foo.com" -H "Content-Type: application/x-www-form-urlencoded; charset=utf-8" --data-ascii "foo:한글中文"`

ByunghyunBang:
I think this problem is NOT a RFC issue BUT a wrong casting from byte to char.

Good example is like these :

1.AsciiString.b2c()
    public static char b2c(byte b) {
        return (char) (b & 0xFF);
    }
2. HttpPostStandardRequestDecoder.skipControlCharacters
    void skipControlCharacters() {
        ...
        while (sao.pos < sao.limit) {
            char c = (char) (sao.bytes[sao.pos++] & 0xFF);
        ...
    }

normanmaurer:
@trustin @Scottmitch wdyt ?

trustin:
I believe it's pretty common in CJK world that header values contain non-ASCII characters. It would be very useful for them if we preserve all the bits. i.e. +1 for @ByunghyunBang's suggestion.

However, we may want to disable this behavior by default and ignore the header entries that violates the RFC to avoid potential issues that non-CJK world users might experience.

normanmaurer:
@trustin I wonder what "issues" you would expect. Can you clarify ?

trustin:
@normanmaurer A user might assume the header values decoded by Netty will always be strictly 7-bit ASCII and write his/her code based on that assumption, which can cause an unexpected exception on their side.

I think this PR is fine to be merged as-is because we didn't care much about non-ASCII characters yet anyway.

fenik17:
@trustin +1.

normanmaurer:
Cherry-picked into 4.1 (71b338ce17036bcb448c6c73e9a35b4971254733) and 4.0 (1bbe4893ec666cf0036f57004e897e6917c96f09)

normanmaurer:
@ByunghyunBang thanks for the PR...

ByunghyunBang:
Thank you for accepting my PR. :+1:

PR #5624, opened 2016-08-03
Do not log on explicit no unsafe

jasontedor:
Motivation:

When Netty components are initialized, Netty attempts to determine if it
has access to unsafe. If Netty is not able to access unsafe (because of
security permissions, or because the JVM was started with an explicit
flag to tell Netty to not look for unsafe), Netty logs an info-level
message that looks like a warning:

Your platform does not provide complete low-level API for accessing
direct buffers reliably. Unless explicitly requested, heap buffer will
always be preferred to avoid potential system unstability.

This log message can appear in applications that depend on Netty for
networking, and this log message can be scary to end-users of such
platforms. This log message should not be emitted if the application was
started with an explicit flag telling Netty to not look for unsafe.

Modifications:

This commit refactors the unsafe detection logic to expose whether or
not the JVM was started with a flag telling Netty to not look for
unsafe. With this exposed information, the log message on unsafe being
unavailable can be modified to not be emitted when Netty is explicitly
told to not look for unsafe.

Result:

No log message is produced when unsafe is unavailable because Netty was
told to not look for it.

Issue #1977, opened 2013-11-11
Create a new system property that switches buffers to little endian


jacques-n:
We use Netty's ByteBuf implementations extensively within Apache Drill.  However, to date we've had to fork the buffer code so that we can change one line: deciding whether to flip the bits in PooledUnsafeDirectByteBuf.  (line 33 of the tip of master).

It would preferable if we could have that set via SystemProperty rather than fork the codebase.  We will generate a pull request to support this work but wanted to see whether there was any concern about this before generating the pull request.  I was thinking to model after the similar system property use in ResourceLeakDetector.

PR #6568, opened 2017-03-25
Allow to free direct buffers on java9 again

normanmaurer:
Motivation:

Java9 adds a new method to Unsafe which allows to free direct ByteBuffer via the cleaner without the need to use an commandline arguments.

Modifications:

- Add Cleaner interface
- Add CleanerJava9 which will be used when using Java9+ and take care of release direct ByteBuffer
- Let Cleaner0 implement Cleaner

Result:

Be able to free direct ByteBuffer on Java9+ again without any commandline arguments.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/CleanerJava6.java
+++ b/common/src/main/java/io/netty/util/internal/CleanerJava6.java
@@ -29,41 +29,32 @@ import java.nio.ByteBuffer;
  *
  * For more details see <a href="https://github.com/netty/netty/issues/2604">#2604</a>.
  */
-final class Cleaner0 {
+final class CleanerJava6 implements Cleaner {
     private static final long CLEANER_FIELD_OFFSET;
     private static final Method CLEAN_METHOD;
-    private static final boolean CLEANER_IS_RUNNABLE;
 
-    private static final InternalLogger logger = InternalLoggerFactory.getInstance(Cleaner0.class);
+    private static final InternalLogger logger = InternalLoggerFactory.getInstance(CleanerJava6.class);
 
     static {
-        ByteBuffer direct = ByteBuffer.allocateDirect(1);
-        Field cleanerField;
         long fieldOffset = -1;
         Method clean = null;
-        boolean cleanerIsRunnable = false;
         Throwable error = null;
         if (PlatformDependent0.hasUnsafe()) {
+            ByteBuffer direct = ByteBuffer.allocateDirect(1);
             try {
-                cleanerField = direct.getClass().getDeclaredField("cleaner");
+                Field cleanerField = direct.getClass().getDeclaredField("cleaner");
                 fieldOffset = PlatformDependent0.objectFieldOffset(cleanerField);
                 Object cleaner = PlatformDependent0.getObject(direct, fieldOffset);
-                try {
-                    // Cleaner implements Runnable from JDK9 onwards.
-                    Runnable runnable = (Runnable) cleaner;
-                    runnable.run();
-                    cleanerIsRunnable = true;
-                } catch (ClassCastException ignored) {
-                    clean = cleaner.getClass().getDeclaredMethod("clean");
-                    clean.invoke(cleaner);
-                }
+                clean = cleaner.getClass().getDeclaredMethod("clean");
+                clean.invoke(cleaner);
             } catch (Throwable t) {
                 // We don't have ByteBuffer.cleaner().
                 fieldOffset = -1;
                 clean = null;
-                cleanerIsRunnable = false;
                 error = t;
             }
+        } else {
+            error = new UnsupportedOperationException("sun.misc.Unsafe unavailable");
         }
         if (error == null) {
             logger.debug("java.nio.ByteBuffer.cleaner(): available");
@@ -72,31 +63,24 @@ final class Cleaner0 {
         }
         CLEANER_FIELD_OFFSET = fieldOffset;
         CLEAN_METHOD = clean;
-        CLEANER_IS_RUNNABLE = cleanerIsRunnable;
-
-        // free buffer if possible
-        freeDirectBuffer(direct);
     }
 
-    static void freeDirectBuffer(ByteBuffer buffer) {
-        if (CLEANER_FIELD_OFFSET == -1 || !buffer.isDirect()) {
+    static boolean isSupported() {
+        return CLEANER_FIELD_OFFSET != -1;
+    }
+
+    @Override
+    public void freeDirectBuffer(ByteBuffer buffer) {
+        if (!buffer.isDirect()) {
             return;
         }
-        assert CLEAN_METHOD != null || CLEANER_IS_RUNNABLE:
-                "CLEANER_FIELD_OFFSET != -1 implies CLEAN_METHOD != null or CLEANER_IS_RUNNABLE == true";
         try {
             Object cleaner = PlatformDependent0.getObject(buffer, CLEANER_FIELD_OFFSET);
             if (cleaner != null) {
-                if (CLEANER_IS_RUNNABLE) {
-                    ((Runnable) cleaner).run();
-                } else {
-                    CLEAN_METHOD.invoke(cleaner);
-                }
+                CLEAN_METHOD.invoke(cleaner);
             }
-        } catch (Throwable t) {
-            // Nothing we can do here.
+        } catch (Throwable cause) {
+            PlatformDependent0.throwException(cause);
         }
     }
-
-    private Cleaner0() { }
 }
--- a/common/src/main/java/io/netty/util/internal/CleanerJava9.java
+++ b/common/src/main/java/io/netty/util/internal/CleanerJava9.java
@@ -1,0 +1,82 @@
+/*
+* Copyright 2017 The Netty Project
+*
+* The Netty Project licenses this file to you under the Apache License,
+* version 2.0 (the "License"); you may not use this file except in compliance
+* with the License. You may obtain a copy of the License at:
+*
+*   http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+* WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+* License for the specific language governing permissions and limitations
+* under the License.
+*/
+package io.netty.util.internal;
+
+import io.netty.util.internal.logging.InternalLogger;
+import io.netty.util.internal.logging.InternalLoggerFactory;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.nio.ByteBuffer;
+
+/**
+ * Provide a way to clean a ByteBuffer on Java9+.
+ */
+final class CleanerJava9 implements Cleaner {
+    private static final InternalLogger logger = InternalLoggerFactory.getInstance(CleanerJava9.class);
+
+    private static final Method INVOKE_CLEANER;
+
+    static {
+        final Method method;
+        final Throwable error;
+        if (PlatformDependent0.hasUnsafe()) {
+            ByteBuffer buffer = ByteBuffer.allocateDirect(1);
+            Object maybeInvokeMethod;
+            try {
+                // See https://bugs.openjdk.java.net/browse/JDK-8171377
+                Method m = PlatformDependent0.UNSAFE.getClass().getDeclaredMethod("invokeCleaner", ByteBuffer.class);
+                m.invoke(PlatformDependent0.UNSAFE, buffer);
+                maybeInvokeMethod = m;
+            } catch (NoSuchMethodException e) {
+                maybeInvokeMethod = e;
+            } catch (InvocationTargetException e) {
+                maybeInvokeMethod = e;
+            } catch (IllegalAccessException e) {
+                maybeInvokeMethod = e;
+            }
+            if (maybeInvokeMethod instanceof Throwable) {
+                method = null;
+                error = (Throwable) maybeInvokeMethod;
+            } else {
+                method = (Method) maybeInvokeMethod;
+                error = null;
+            }
+        } else {
+            method = null;
+            error = new UnsupportedOperationException("sun.misc.Unsafe unavailable");
+        }
+        if (error == null) {
+            logger.debug("java.nio.ByteBuffer.cleaner(): available");
+        } else {
+            logger.debug("java.nio.ByteBuffer.cleaner(): unavailable", error);
+        }
+        INVOKE_CLEANER = method;
+    }
+
+    static boolean isSupported() {
+        return INVOKE_CLEANER != null;
+    }
+
+    @Override
+    public void freeDirectBuffer(ByteBuffer buffer) {
+        try {
+            INVOKE_CLEANER.invoke(PlatformDependent0.UNSAFE, buffer);
+        } catch (Throwable cause) {
+            PlatformDependent0.throwException(cause);
+        }
+    }
+}
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent.java
@@ -67,9 +67,6 @@ public final class PlatformDependent {
     private static final Pattern MAX_DIRECT_MEMORY_SIZE_ARG_PATTERN = Pattern.compile(
             "\\s*-XX:MaxDirectMemorySize\\s*=\\s*([0-9]+)\\s*([kKmMgG]?)\\s*$");
 
-    // this must be initialized before any code below triggers initialization of PlatformDependent0
-    private static final boolean IS_EXPLICIT_NO_UNSAFE = explicitNoUnsafe0();
-
     private static final boolean IS_ANDROID = isAndroid0();
     private static final boolean IS_WINDOWS = isWindows0();
     private static final boolean MAYBE_SUPER_USER;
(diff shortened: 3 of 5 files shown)


common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -281,10 +282,6 @@ public Object run() {
 
         logger.debug("java.nio.DirectByteBuffer.<init>(long, int): {}",
                 DIRECT_BUFFER_CONSTRUCTOR != null ? "available" : "unavailable");
-
-        if (direct != null) {
-            freeDirectBuffer(direct);
```
normanmaurer: removing this makes it easier to break cycle references and its not super important anyway

common/src/main/java/io/netty/util/internal/Cleaner0.java
```diff
@@ -29,10 +29,9 @@
  *
  * For more details see <a href="https://github.com/netty/netty/issues/2604">#2604</a>.
  */
-final class Cleaner0 {
+final class Cleaner0 implements Cleaner {
```
netkins: CRITICAL Initialization circularity between io.netty.util.internal.Cleaner0 and io.netty.util.internal.CleanerJava9 [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=findbugs%3AIC_INIT_CIRCULARITY)
CRITICAL Initialization circularity between io.netty.util.internal.Cleaner0 and io.netty.util.internal.PlatformDependent [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=findbugs%3AIC_INIT_CIRCULARITY)
CRITICAL Initialization circularity between io.netty.util.internal.Cleaner0 and io.netty.util.internal.PlatformDependent0 [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=findbugs%3AIC_INIT_CIRCULARITY)

common/src/main/java/io/netty/util/internal/CleanerJava9.java
```diff
@@ -0,0 +1,76 @@
...
+import sun.misc.Unsafe;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.nio.ByteBuffer;
+
+/**
+ * Provide a way to clean a ByteBuffer on Java9+.
+ */
+final class CleanerJava9 implements Cleaner {
```
netkins: CRITICAL Initialization circularity between io.netty.util.internal.CleanerJava9 and io.netty.util.internal.PlatformDependent [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=findbugs%3AIC_INIT_CIRCULARITY)
CRITICAL Initialization circularity between io.netty.util.internal.CleanerJava9 and io.netty.util.internal.PlatformDependent0 [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=findbugs%3AIC_INIT_CIRCULARITY)

johnou:
@normanmaurer how about the non unsafe version? https://bugs.openjdk.java.net/browse/JDK-8148117

normanmaurer:
@johnou not sure i get your question... Can you explain ?

common/src/main/java/io/netty/util/internal/Cleaner0.java
normanmaurer: I think its fine... @Scottmitch @nmittler @trustin PTAL

common/src/main/java/io/netty/util/internal/CleanerJava9.java
normanmaurer: I think its fine... @Scottmitch @nmittler @trustin PTAL

johnou:
@normanmaurer nvm, I misread "If, at some point in the future, it is determined that sun.misc.Cleaner is in fact a critical internal API ( with static usage ), 
then it can be reinstated as a subtype of jdk.internal.ref.Cleaner, providing the same public API. " as though there would already be a public API.

normanmaurer:
@trustin PTAL as well

common/src/main/java/io/netty/util/internal/Cleaner0.java
fenik17: It's a possible deadlock such as [guava issue #1977](https://github.com/google/guava/issues/1977)?

normanmaurer: I think I found the cycle... stay tuned

common/src/main/java/io/netty/util/internal/CleanerJava9.java
```diff
@@ -0,0 +1,76 @@
...
+* distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+* WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+* License for the specific language governing permissions and limitations
+* under the License.
+*/
+package io.netty.util.internal;
+
+import io.netty.util.internal.logging.InternalLogger;
+import io.netty.util.internal.logging.InternalLoggerFactory;
+import sun.misc.Unsafe;
```
netkins: MINOR Remove this unused import 'sun.misc.Unsafe'. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

common/src/main/java/io/netty/util/internal/CleanerJava9.java
```diff
@@ -0,0 +1,75 @@
...
+* distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+* WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+* License for the specific language governing permissions and limitations
+* under the License.
+*/
+package io.netty.util.internal;
+
+import io.netty.util.internal.logging.InternalLogger;
+import io.netty.util.internal.logging.InternalLoggerFactory;
+
```
netkins: MINOR Remove this unused import 'sun.misc.Unsafe'. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

common/src/main/java/io/netty/util/internal/Cleaner0.java
normanmaurer: fixed

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -244,7 +246,7 @@ public Object run() {
                 public Object run() {
                     try {
                         Class<?> bitsClass =
-                                Class.forName("java.nio.Bits", false, PlatformDependent.getSystemClassLoader());
+                                Class.forName("java.nio.Bits", false, getSystemClassLoader());
```
normanmaurer: This fixed a cycle reference

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -281,10 +283,35 @@ public Object run() {
...
                 DIRECT_BUFFER_CONSTRUCTOR != null ? "available" : "unavailable");
+    }
+
+    static boolean isExplicitNoUnsafe() {
+        return IS_EXPLICIT_NO_UNSAFE;
+    }
 
-        if (direct != null) {
-            freeDirectBuffer(direct);
+    private static boolean explicitNoUnsafe0() {
```
normanmaurer: I moved this from `PlatformDependent` to remove the a cycle reference.

common/src/main/java/io/netty/util/internal/CleanerJava9.java
```diff
@@ -0,0 +1,75 @@
...
+            method = null;
+            logger.debug("java.nio.ByteBuffer.cleaner(): unavailable", (Throwable) maybeInvokeMethod);
+        } else {
+            method = (Method) maybeInvokeMethod;
+            logger.debug("java.nio.ByteBuffer.cleaner(): available");
+        }
+        INVOKE_CLEANER = method;
+    }
+
+    private static void freeDirectBuffer0(ByteBuffer buffer) {
```
Scottmitch: is this method necessary? seems like it is only used in a single place

common/src/main/java/io/netty/util/internal/Cleaner0.java
```diff
@@ -72,31 +63,28 @@
...
+        freeDirectBuffer0(direct);
+    }
+
+    @Override
+    public void freeDirectBuffer(ByteBuffer buffer) {
+        freeDirectBuffer0(buffer);
     }
 
-    static void freeDirectBuffer(ByteBuffer buffer) {
+    private static void freeDirectBuffer0(ByteBuffer buffer) {
```
Scottmitch: I think we can remove this method, and just move the code into `freeDirectBuffer`. We invoke the cleaner above `clean.invoke(cleaner);` so we may be double freeing this buffer

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -122,7 +120,7 @@ public Random current() {
             logger.debug("-Dio.netty.noPreferDirect: {}", !DIRECT_BUFFER_PREFERRED);
         }
 
-        if (!hasUnsafe() && !isAndroid() && !IS_EXPLICIT_NO_UNSAFE) {
+        if (!hasUnsafe() && !isAndroid() && !PlatformDependent0.isExplicitNoUnsafe()) {
```
Scottmitch: `isExplicitNoUnsafe` should be reflected in `hasUnsafe` ... can we just remove the `!PlatformDependent0.isExplicitNoUnsafe()` part?

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -158,6 +156,19 @@ public Random current() {
         logger.debug("io.netty.maxDirectMemory: {} bytes", maxDirectMemory);
 
         MAYBE_SUPER_USER = maybeSuperUser0();
+
+        if (hasUnsafe() && !isAndroid()) {
+            // only direct to method if we are not running on android.
+            // See https://github.com/netty/netty/issues/2604
+            CLEANER = javaVersion() >= 9 ? new CleanerJava9() : new Cleaner0();
```
Scottmitch: consider renaming `Cleaner0` to `CleanerJava6`.

normanmaurer:
@Scottmitch addressed all of it... Good to merge ?

common/src/main/java/io/netty/util/internal/CleanerJava6.java
```diff
@@ -29,39 +29,29 @@
...
+                clean = cleaner.getClass().getDeclaredMethod("clean");
+                clean.invoke(cleaner);
             } catch (Throwable t) {
                 // We don't have ByteBuffer.cleaner().
                 fieldOffset = -1;
                 clean = null;
-                cleanerIsRunnable = false;
                 error = t;
             }
         }
```
Scottmitch: the conditional below checks if `error == null` and then says "cleaner is available". However this may not be the case if we have no unsafe. We should adjust the logic below or set the `error = <exception indicating unsafe isn't available>`.

common/src/main/java/io/netty/util/internal/CleanerJava6.java
```diff
@@ -72,31 +62,21 @@
...
 
-    static void freeDirectBuffer(ByteBuffer buffer) {
+    @Override
+    public void freeDirectBuffer(ByteBuffer buffer) {
         if (CLEANER_FIELD_OFFSET == -1 || !buffer.isDirect()) {
             return;
         }
-        assert CLEAN_METHOD != null || CLEANER_IS_RUNNABLE:
-                "CLEANER_FIELD_OFFSET != -1 implies CLEAN_METHOD != null or CLEANER_IS_RUNNABLE == true";
+        assert CLEAN_METHOD != null;
```
Scottmitch: nit: remove this assert ... we will very likely NPE via a JVM trap below anyways.

common/src/main/java/io/netty/util/internal/CleanerJava6.java
```diff
@@ -72,31 +62,21 @@
...
             if (cleaner != null) {
-                if (CLEANER_IS_RUNNABLE) {
-                    ((Runnable) cleaner).run();
-                } else {
-                    CLEAN_METHOD.invoke(cleaner);
-                }
+                CLEAN_METHOD.invoke(cleaner);
             }
         } catch (Throwable t) {
             // Nothing we can do here.
```
Scottmitch: can we log a debug statement, or is this expected to happen frequently?

common/src/main/java/io/netty/util/internal/CleanerJava9.java
```diff
@@ -0,0 +1,71 @@
...
+            maybeInvokeMethod = e;
+        }
+        if (maybeInvokeMethod instanceof Throwable) {
+            method = null;
+            logger.debug("java.nio.ByteBuffer.cleaner(): unavailable", (Throwable) maybeInvokeMethod);
+        } else {
+            method = (Method) maybeInvokeMethod;
+            logger.debug("java.nio.ByteBuffer.cleaner(): available");
+        }
+        INVOKE_CLEANER = method;
```
Scottmitch: we should also log something similar to the `Java6` class if the cleaner is not supported.

common/src/main/java/io/netty/util/internal/CleanerJava9.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    public void freeDirectBuffer(ByteBuffer buffer) {
+        if (INVOKE_CLEANER == null) {
```
Scottmitch: suggestion: instead of checking this condition every time we could have `PlatformDependent` detect if the cleaner is properly initialized, and if not just use the NOOP cleaner. We should do the same thing for the java6 class:

```java
class CleanerJava9 {

static {
..
}

static boolean isInitialized() {
  return INVOKE_CLEANER != null;
}
..
}

class PlatformDependent {
..
if (java9) {
  CLEANER = CleanerJava9.isInitialized() ? new CleanerJava9() : NOOP;
} else {
  CLEANER = CleanerJava6.isInitialized() ? new CleanerJava6() : NOOP;
}
...
}
```

common/src/main/java/io/netty/util/internal/CleanerJava6.java
normanmaurer: i would argue this should never happen as we tried it before in the static block. Maybe best to throw an Error . WDYT ?

Scottmitch: yah if we never expect it to happen lets re-throw it. What about `PlatformDependent.throwException(t);`?

common/src/main/java/io/netty/util/internal/CleanerJava6.java
```diff
@@ -72,31 +64,24 @@
         }
         CLEANER_FIELD_OFFSET = fieldOffset;
         CLEAN_METHOD = clean;
-        CLEANER_IS_RUNNABLE = cleanerIsRunnable;
+    }
 
-        // free buffer if possible
-        freeDirectBuffer(direct);
+    public static boolean isSupported() {
```
Scottmitch: package private?

common/src/main/java/io/netty/util/internal/CleanerJava9.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+        if (error == null) {
+            logger.debug("java.nio.ByteBuffer.cleaner(): available");
+        } else {
+            logger.debug("java.nio.ByteBuffer.cleaner(): unavailable", error);
+        }
+        INVOKE_CLEANER = method;
+    }
+
+    public static boolean isSupported() {
```
Scottmitch: package private?

common/src/main/java/io/netty/util/internal/CleanerJava9.java
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Provide a way to clean a ByteBuffer on Java9+.
+ */
+final class CleanerJava9 implements Cleaner {
+    private static final InternalLogger logger = InternalLoggerFactory.getInstance(CleanerJava9.class);
+
+    private static final Method INVOKE_CLEANER;
+
+    static {
+        Method method;
```
Scottmitch: final?

common/src/main/java/io/netty/util/internal/CleanerJava9.java
```diff
@@ -0,0 +1,82 @@
...
+        Throwable error;
```
Scottmitch: final?

normanmaurer:
Cherry-picked into 4.1 (7b6119a0a478b40f12e85c0e92e603f9f6ee5b6c) and 4.0 (295c24e246470dfdcc63066133ebcaa23c5144a9)

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -122,7 +127,7 @@ public Random current() {
             logger.debug("-Dio.netty.noPreferDirect: {}", !DIRECT_BUFFER_PREFERRED);
         }
 
-        if (!hasUnsafe() && !isAndroid() && !IS_EXPLICIT_NO_UNSAFE) {
```
jasontedor: Why was this removed?  It means that logging for this shows up even if no unsafe was explicitly set which was the point of adding this in the first place It means that with Netty 4.1.10.Final our users are back to seeing a message that they don't understand that makes them think their system is broken. Unless there is a good reason that this is a removed (I'm sorry if I'm missing it), I would like to open a PR to add it back.

jasontedor: Relates #5624

normanmaurer:
There was no good reason and it's my fault to miss this :(

Can you please do a PR and also add a comment in the code why it should never be removed again.

Sorry again for the trouble

common/src/main/java/io/netty/util/internal/PlatformDependent.java
jasontedor: Also, it seems to me that this change was not needed for the stated purpose of this change (to allow freeing direct buffers on JDK 9)? Again, maybe I'm missing something and if so I'm sorry.

Adding unrelated changes to PRs makes it hard to follow what is going on. I follow this repository, if I see a change titled "Allow to free direct buffers on java9 again" I think I know what it's about and I skip reading the code change. If an unrelated code change is slipped in, how can I as a follower know this without reading every code change which I think is an unreasonable assumption?

Had this been in a separate PR I would have seen it and would have responded then to please not remove it.

jasontedor:
Thank you for understanding @normanmaurer! I will open a PR soon.

> Can you please do a PR and also add a comment in the code why it should never be removed again.

I will do this! 😄

jasontedor:
I opened #6696.

Issue #6564, opened 2017-03-24
When the number of pending unencrypted bytes is high, SslHandler tries to allocate a huge outNet buffer.


trustin:
### Expected behavior

When a user attempts to write a large buffer (e.g. 1GB) via SslHandler, SslHandler should not attempt to allocate an outNet buffer that's as large as the buffer given by the user, because

- We may not be able to flush fast enough to deallocate it soon.
- A user may get an OOME when under backpressure due to double memory allocation.
- It's inefficient to allocate a direct buffer that's bigger than 16MiB (pooled allocator's chunk size)

### Actual behavior

When a user attempts to write a large buffer, SslHandler should not allocate a huge buffer upfront but only allocate up to certain threshold. e.g. Always keep the number of pending outNet bytes low even if `SslHandler.pendingUnencryptedWrites.bytes` is high.

### Steps to reproduce

- Give JVM relatively small amount of heap and direct memory (e.g. 128m)
- On an SSL connection, attempt to send a buffer that's slightly larger than half of the available direct memory. 
- SslHandler will attempt to allocate a buffer that's as large as the buffer being sent and end up with an OOME.

### Minimal yet complete reproducer code (or URL to code)

I reproduced the problem by running `./gradlew core:check` for [Armeria](https://github.com/line/armeria), but it's not a minimal reproducer obviously. You get the idea though, right? ;-)

### Netty version

4.1.9.Final. No problem with 4.1.8.Final.

### JVM version (e.g. `java -version`)

```
java version "1.8.0_112"
Java(TM) SE Runtime Environment (build 1.8.0_112-b15)
Java HotSpot(TM) 64-Bit Server VM (build 25.112-b15, mixed mode)
```

### OS version (e.g. `uname -a`)

`Linux porori 4.9.0-0.bpo.2-amd64 #1 SMP Debian 4.9.13-1~bpo8+1 (2017-02-27) x86_64 GNU/Linux`

trustin:
I believe [this](https://github.com/netty/netty/commit/53fc69390130f7a5a33d454a59ad214f6b4176f5#diff-cfd417915faad637e4278b54adafd9baL1609) is related. /cc @Scottmitch

trustin:
A workaround is probably for a user to split the large buffer into multiple 16KiB ones, although I didn't try yet.

normanmaurer:
[...]

PR #6569, opened 2017-03-26
Limit the maximum size of the allocated outbound buffer to MAX_ENCRYP…

normanmaurer:
…TED_PACKET_LENGTH

Motivation:

We should limit the size of the allocated outbound buffer to MAX_ENCRYPTED_PACKET_LENGTH to ensure we not cause an OOME when the user tries to encrypt a very big buffer.

Modifications:

Limit the size of the allocated outbound buffer to MAX_ENCRYPTED_PACKET_LENGTH

Result:

Fixes [#6564]

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -1611,7 +1611,8 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
     }
 
     static int calculateOutNetBufSize(int pendingBytes, int numComponents) {
-        return (int) min(Integer.MAX_VALUE, pendingBytes + (long) MAX_TLS_RECORD_OVERHEAD_LENGTH * numComponents);
+        return (int) min(MAX_ENCRYPTED_PACKET_LENGTH,
+                pendingBytes + (long) MAX_TLS_RECORD_OVERHEAD_LENGTH * numComponents);
     }
 
     private final class OpenSslSession implements SSLSession, ApplicationProtocolAccessor {
--- a/handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
@@ -318,7 +318,7 @@ public class OpenSslEngineTest extends SSLEngineTest {
 
     @Test
     public void testCalculateOutNetBufSizeOverflow() {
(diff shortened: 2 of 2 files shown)


normanmaurer:
@trustin PTAL.... I limited the size to MAX_ENCRYPTED_PACKET_LENGTH as most of the times the buffer should be smaller anyway and allocations are "cheap" when these can be served from within the  `PooledByteBufAllocator`

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1611,7 +1611,8 @@ private boolean isDestroyed() {
     }
 
     static int calculateOutNetBufSize(int pendingBytes, int numComponents) {
-        return (int) min(Integer.MAX_VALUE, pendingBytes + (long) MAX_TLS_RECORD_OVERHEAD_LENGTH * numComponents);
+        return (int) Math.min(MAX_ENCRYPTED_PACKET_LENGTH,
```
netkins: MAJOR Unnecessary use of fully qualified name 'Math.min' due to existing static import 'java.lang.Math.min' [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUnnecessaryFullyQualifiedName)

normanmaurer:
@Scottmitch WDYT ?

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
```diff
@@ -318,7 +318,7 @@ public void testSrcsLenOverFlowCorrectlyHandled() throws Exception {
 
     @Test
     public void testCalculateOutNetBufSizeOverflow() {
-        assertEquals(MAX_VALUE,
+        assertEquals(OpenSslEngine.MAX_ENCRYPTED_PACKET_LENGTH,
```
Scottmitch: `OpenSslEngine.MAX_ENCRYPTED_PACKET_LENGTH` -> `MAX_ENCRYPTED_PACKET_LENGTH`

normanmaurer:
Cherry-picked into 4.0 (29a83e7df9dc70c12502226b176f9549dd1af508) and 4.1 (ed1071d327cdd77bf520befda7f62081374e77bc)

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
normanmaurer: done

trustin:
Thanks @normanmaurer for working on this.

PR #6570, opened 2017-03-26
Add 'io.netty.tryAllocateUninitializedArray' system property which al…

normanmaurer:
…lows to allocate byte[] without memset in Java9+

Motivation:

Java9 added a new method to Unsafe which allows to allocate a byte[] without memset it. This can have a massive impact in allocation times when the byte[] is big. This change allows to enable this when using Java9 with the io.netty.tryAllocateUninitializedArray property when running Java9+.
Please note that you will need to open up the jdk.internal.misc package via '--add-opens java.base/jdk.internal.misc=ALL-UNNAMED' as well.

Modifications:

Allow to allocate byte[] without memset on Java9+

Result:

Better performance when allocate big heap buffers and using java9.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/PoolArena.java
+++ b/buffer/src/main/java/io/netty/buffer/PoolArena.java
@@ -672,6 +672,10 @@ abstract class PoolArena<T> implements PoolArenaMetric {
                     directMemoryCacheAlignment);
         }
 
+        private static byte[] newByteArray(int size) {
+            return PlatformDependent.allocateUninitializedArray(size);
+        }
+
         @Override
         boolean isDirect() {
             return false;
@@ -679,12 +683,12 @@ abstract class PoolArena<T> implements PoolArenaMetric {
 
         @Override
         protected PoolChunk<byte[]> newChunk(int pageSize, int maxOrder, int pageShifts, int chunkSize) {
-            return new PoolChunk<byte[]>(this, new byte[chunkSize], pageSize, maxOrder, pageShifts, chunkSize, 0);
+            return new PoolChunk<byte[]>(this, newByteArray(chunkSize), pageSize, maxOrder, pageShifts, chunkSize, 0);
         }
 
         @Override
         protected PoolChunk<byte[]> newUnpooledChunk(int capacity) {
-            return new PoolChunk<byte[]>(this, new byte[capacity], capacity, 0);
+            return new PoolChunk<byte[]>(this, newByteArray(capacity), capacity, 0);
         }
 
         @Override
--- a/buffer/src/main/java/io/netty/buffer/UnpooledUnsafeHeapByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/UnpooledUnsafeHeapByteBuf.java
@@ -30,6 +30,11 @@ class UnpooledUnsafeHeapByteBuf extends UnpooledHeapByteBuf {
     }
 
     @Override
+    byte[] allocateArray(int initialCapacity) {
+        return PlatformDependent.allocateUninitializedArray(initialCapacity);
+    }
+
+    @Override
     public byte getByte(int index) {
         checkIndex(index);
         return _getByte(index);
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent.java
@@ -97,6 +97,7 @@ public final class PlatformDependent {
     private static final long DIRECT_MEMORY_LIMIT;
     private static final ThreadLocalRandomProvider RANDOM_PROVIDER;
     private static final Cleaner CLEANER;
+    private static final int UNINITIALIZED_ARRAY_ALLOCATION_THRESHOLD;
 
     public static final boolean BIG_ENDIAN_NATIVE_ORDER = ByteOrder.nativeOrder() == ByteOrder.BIG_ENDIAN;
 
@@ -160,7 +161,13 @@ public final class PlatformDependent {
             }
         }
         DIRECT_MEMORY_LIMIT = maxDirectMemory;
-        logger.debug("io.netty.maxDirectMemory: {} bytes", maxDirectMemory);
+        logger.debug("-Dio.netty.maxDirectMemory: {} bytes", maxDirectMemory);
+
+        int tryAllocateUninitializedArray =
+                SystemPropertyUtil.getInt("io.netty.uninitializedArrayAllocationThreshold", 1024);
+        UNINITIALIZED_ARRAY_ALLOCATION_THRESHOLD = javaVersion() >= 9 && PlatformDependent0.hasAllocateArrayMethod() ?
+                tryAllocateUninitializedArray : -1;
+        logger.debug("-Dio.netty.uninitializedArrayAllocationThreshold: {}", UNINITIALIZED_ARRAY_ALLOCATION_THRESHOLD);
 
         MAYBE_SUPER_USER = maybeSuperUser0();
 
@@ -175,6 +182,11 @@ public final class PlatformDependent {
         } else {
             CLEANER = NOOP;
         }
+    }
+
+    public static byte[] allocateUninitializedArray(int size) {
+        return UNINITIALIZED_ARRAY_ALLOCATION_THRESHOLD < 0 || UNINITIALIZED_ARRAY_ALLOCATION_THRESHOLD > size ?
+                new byte[size] : PlatformDependent0.allocateUninitializedArray(size);
     }
 
     /**
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent0.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent0.java
@@ -40,7 +40,9 @@ final class PlatformDependent0 {
     private static final long BYTE_ARRAY_BASE_OFFSET;
     private static final Constructor<?> DIRECT_BUFFER_CONSTRUCTOR;
     private static final boolean IS_EXPLICIT_NO_UNSAFE = explicitNoUnsafe0();
+    private static final Method ALLOCATE_ARRAY_METHOD;
 
+    private static final Object INTERNAL_UNSAFE;
     static final Unsafe UNSAFE;
 
     // constants borrowed from murmur3
@@ -59,12 +61,15 @@ final class PlatformDependent0 {
     static {
         final ByteBuffer direct;
         Field addressField = null;
+        Method allocateArrayMethod = null;
         Unsafe unsafe;
+        Object internalUnsafe = null;
 
         if (isExplicitNoUnsafe()) {
             direct = null;
             addressField = null;
             unsafe = null;
+            internalUnsafe = null;
         } else {
             direct = ByteBuffer.allocateDirect(1);
 
@@ -187,6 +192,7 @@ final class PlatformDependent0 {
             BYTE_ARRAY_BASE_OFFSET = -1;
             UNALIGNED = false;
             DIRECT_BUFFER_CONSTRUCTOR = null;
+            ALLOCATE_ARRAY_METHOD = null;
         } else {
             Constructor<?> directBufferConstructor;
             long address = -1;
@@ -237,7 +243,6 @@ final class PlatformDependent0 {
                 }
             }
             DIRECT_BUFFER_CONSTRUCTOR = directBufferConstructor;
-
             ADDRESS_FIELD_OFFSET = objectFieldOffset(addressField);
             BYTE_ARRAY_BASE_OFFSET = UNSAFE.arrayBaseOffset(byte[].class);
             boolean unaligned;
@@ -279,7 +284,64 @@ final class PlatformDependent0 {
             }
 
             UNALIGNED = unaligned;
+
+            Object maybeException = AccessController.doPrivileged(new PrivilegedAction<Object>() {
+                @Override
+                public Object run() {
+                    try {
+                        // Java9 has jdk.internal.misc.Unsafe and not all methods are propergated to
+                        // sun.misc.Unsafe
+                        Class<?> internalUnsafeClass = getClassLoader(PlatformDependent0.class)
+                                .loadClass("jdk.internal.misc.Unsafe");
+                        Method method = internalUnsafeClass.getDeclaredMethod("getUnsafe");
+                        return method.invoke(null);
+                    } catch (Throwable e) {
+                        return e;
+                    }
+                }
+            });
+
+            if (!(maybeException instanceof Throwable)) {
+                internalUnsafe = maybeException;
+                final Object finalInternalUnsafe = internalUnsafe;
+                maybeException = AccessController.doPrivileged(new PrivilegedAction<Object>() {
+                    @Override
+                    public Object run() {
+                        try {
+                            return finalInternalUnsafe.getClass().getDeclaredMethod(
+                                    "allocateUninitializedArray", Class.class, int.class);
+                        } catch (NoSuchMethodException e) {
+                            return e;
+                        } catch (SecurityException e) {
+                            return e;
+                        }
+                    }
+                });
+
+                if (maybeException instanceof Method) {
+                    try {
+                        Method m = (Method) maybeException;
+                        byte[] bytes = (byte[]) m.invoke(finalInternalUnsafe, byte.class, 8);
+                        assert bytes.length == 8;
+                        allocateArrayMethod = m;
+                    } catch (IllegalAccessException e) {
+                        maybeException = e;
+                    } catch (InvocationTargetException e) {
+                        maybeException = e;
+                    }
+                }
+            }
+            ALLOCATE_ARRAY_METHOD = allocateArrayMethod;
+
+            if (maybeException instanceof Throwable) {
+                logger.debug("jdk.internal.misc.Unsafe.allocateUninitializedArray(int): unavailable",
+                        (Throwable) maybeException);
+            } else {
+                logger.debug("jdk.internal.misc.Unsafe.allocateUninitializedArray(int): available");
+            }
         }
+
+        INTERNAL_UNSAFE = internalUnsafe;
 
         logger.debug("java.nio.DirectByteBuffer.<init>(long, int): {}",
                 DIRECT_BUFFER_CONSTRUCTOR != null ? "available" : "unavailable");
@@ -341,6 +403,20 @@ final class PlatformDependent0 {
 
     static ByteBuffer allocateDirectNoCleaner(int capacity) {
         return newDirectBuffer(UNSAFE.allocateMemory(capacity), capacity);
+    }
+
+    static boolean hasAllocateArrayMethod() {
+        return ALLOCATE_ARRAY_METHOD != null;
+    }
+
+    static byte[] allocateUninitializedArray(int size) {
+        try {
+            return (byte[]) ALLOCATE_ARRAY_METHOD.invoke(INTERNAL_UNSAFE, byte.class, size);
+        } catch (IllegalAccessException e) {
+            throw new Error(e);
+        } catch (InvocationTargetException e) {
+            throw new Error(e);
+        }
     }
 
     static ByteBuffer newDirectBuffer(long address, int capacity) {
--- a/microbench/src/main/java/io/netty/microbench/internal/UnitializedArrayBenchmark.java
+++ b/microbench/src/main/java/io/netty/microbench/internal/UnitializedArrayBenchmark.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.microbench.internal;
+
+import io.netty.microbench.util.AbstractMicrobenchmark;
+import io.netty.util.internal.PlatformDependent;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Param;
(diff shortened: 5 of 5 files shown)


normanmaurer:
See also https://shipilev.net/jvm-anatomy-park/7-initialization-costs/

buffer/src/main/java/io/netty/buffer/UnpooledUnsafeHeapByteBuf.java
```diff
@@ -29,6 +29,10 @@
         super(alloc, initialCapacity, maxCapacity);
     }
 
+    byte[] allocateArray(int initialCapacity) {
```
netkins: MAJOR Add the "@Override" annotation above this method signature [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1161)

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -99,6 +99,7 @@
     private static final AtomicLong DIRECT_MEMORY_COUNTER;
     private static final long DIRECT_MEMORY_LIMIT;
     private static final ThreadLocalRandomProvider RANDOM_PROVIDER;
+    private static final boolean TRY_ALLOCATE_UNINITIAlIZED_ARRAY;
```
netkins: MINOR Rename this constant name to match the regular expression '^([a-z][a-zA-Z0-9]*|[A-Z][A-Z0-9_v]*)$'. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS00115)

buffer/src/main/java/io/netty/buffer/PoolArena.java
```diff
@@ -672,19 +672,23 @@ private void destroyPoolChunkLists(PoolChunkList<T>... chunkLists) {
                     directMemoryCacheAlignment);
         }
 
+        private static byte[] newByteArray(int size) {
+            return HAS_UNSAFE ? PlatformDependent.allocateUninitializedArray(size) : new byte[size];
```
fenik17: a HAS_UNSAFE check not needed here. It's also checked in the allocateUninitializedArray().

normanmaurer: +1

common/src/main/java/io/netty/util/internal/PlatformDependent.java
normanmaurer: false positive ?

jasontedor: I think it's because of the lowercase `l` in the middle of `UNINITIAlIZED`?

normanmaurer: Doh ... you are right

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -105,7 +107,7 @@ public Object run() {
             // http://www.mail-archive.com/jdk6-dev@openjdk.java.net/msg00698.html
             if (unsafe != null) {
                 final Unsafe finalUnsafe = unsafe;
-                final Object maybeException = AccessController.doPrivileged(new PrivilegedAction<Object>() {
+                Object maybeException = AccessController.doPrivileged(new PrivilegedAction<Object>() {
```
jasontedor: Sorry, I'm missing why you dropped `final` here?

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -277,8 +279,40 @@ public Object run() {
...
+                    allocateArrayMethod = null;
+                }
+            } else {
+                allocateArrayMethod = null;
+            }
+            ALLOCATE_ARRAY_METHOD = allocateArrayMethod;
         }
 
+        logger.debug("sun.misc.Unsafe.allocateUninitializedArray: {}",
+                ALLOCATE_ARRAY_METHOD != null ? "available" : "unavailable");
```
jasontedor: Log any exception that caused it to be unavailable?

trustin:
How about making this behavior default and adding a system property that
disables it?

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
normanmaurer: let me re-add it again. Messed up while refactoring.

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
normanmaurer: +1

normanmaurer:
@trustin not sure if we want to change the default behaviour or not. Thought it may make sense to keep the old behaviour just to be a bit more conservative . WDYT ? Also @nmittler @Scottmitch

trustin:
@normanmaurer I think it's OK, because we never guaranteed that a buffer allocated by `ByteBufAllocator` is zeroed out. We may need to leave some note about this 'potentially' breaking change in the release note. I doubt it will be a problem though.

normanmaurer:
@trustin also the user will need to open up access as explained in the commit message:

```
Please note that you will need to open up the jdk.internal.misc package via '--add-opens java.base/jdk.internal.misc=ALL-UNNAMED' as well.
```

So maybe its fine as you said...

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -277,8 +283,74 @@ public Object run() {
...
+                @Override
+                public Object run() {
+                    try {
+                        // Java9 has jdk.internal.misc.Unsafe and not all methods are propergated to
+                        // sun.misc.Unsafe
+                        Class<?> internalUnsafeClass = getClassLoader(PlatformDependent0.class)
+                                .loadClass("jdk.internal.misc.Unsafe");
+                        Method method = internalUnsafeClass.getDeclaredMethod("getUnsafe");
+                        return method.invoke(null);
+                    } catch (ClassNotFoundException e) {
```
fenik17: Maybe catch only an `Exception` instead of all?

nmittler:
@normanmaurer I agree with @trustin here, I think it's fine

> I think it's OK, because we never guaranteed that a buffer allocated by ByteBufAllocator is zeroed out. We may need to leave some note about this 'potentially' breaking change in the release note. I doubt it will be a problem though.

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
normanmaurer: I like to just be explicit about what to catch and what not.

normanmaurer:
@nmittler @trustin ok changed the default to be `true`. PTAL again

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -277,8 +283,74 @@ public Object run() {
...
+                    }
+                } else {
+                    allocateArrayMethod = null;
+                }
+            } else {
+                allocateArrayMethod = null;
+            }
+            ALLOCATE_ARRAY_METHOD = allocateArrayMethod;
+
+            logger.debug("jdk.internal.misc.Unsafe.allocateUninitializedArray: {}",
```
trustin: Array: -> Array(): ?

normanmaurer: +1... let me fix this

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -277,8 +283,74 @@ public Object run() {
...
+                        assert bytes.length == 8;
+                    } catch (IllegalAccessException e) {
+                        allocateArrayMethod = null;
+                        maybeException = e;
+                    } catch (InvocationTargetException e) {
+                        allocateArrayMethod = null;
+                        maybeException = e;
+                    }
+                } else {
+                    allocateArrayMethod = null;
```
nmittler: nit: initialize allocateArrayMethod to `null` above so you don't need this else?

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -277,8 +283,74 @@ public Object run() {
...
+                }
+            } else {
+                allocateArrayMethod = null;
```
nmittler: nit: same comment

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -316,6 +388,19 @@ static ByteBuffer allocateDirectNoCleaner(int capacity) {
         return newDirectBuffer(UNSAFE.allocateMemory(capacity), capacity);
     }
 
+    static byte[] allocateUninitializedArray(int size) {
+        if (ALLOCATE_ARRAY_METHOD == null) {
+            return new byte[size];
+        }
+        try {
+            return (byte[]) ALLOCATE_ARRAY_METHOD.invoke(INTERNAL_UNSAFE, byte.class, size);
```
nmittler: Do you have benchmarks to show the impact of using reflection here? I'm concerned that we're trading the cost of memset for the cost of a reflective call.

Ideally, we would just call `Unsafe. allocateUninitializedArray` but that mean that we would have to require Java 9 to build.

normanmaurer:
Will add a benchmark

normanmaurer:
So I added a jmh benchmark and as @nmittler suspected there is some overhead involved in doing the reflection thingy. So I think after running the benchmark a few times I would suggest we do the allocation based on the size either one way or the other. I seems to pay of for allocations >= 100.

@Scottmitch @trustin @nmittler WDYT ?

```

Result "io.netty.microbench.internal.UnitializedArrayBenchmark.allocateUninitializedByteArray":
  103,443 ±(99.9%) 45,122 ns/op [Average]
  (min, avg, max) = (91,138, 103,443, 127,497), stdev = 16,091
  CI (99.9%): [58,321, 148,564] (assumes normal distribution)


# Run complete. Total time: 00:07:43

Benchmark                                                 (size)  Mode  Cnt     Score      Error  Units
UnitializedArrayBenchmark.allocateInitializedByteArray         1  avgt    6     7,938 ±    1,381  ns/op
UnitializedArrayBenchmark.allocateInitializedByteArray        10  avgt    6     9,046 ±    1,725  ns/op
UnitializedArrayBenchmark.allocateInitializedByteArray       100  avgt    6    27,600 ±    6,358  ns/op
UnitializedArrayBenchmark.allocateInitializedByteArray      1000  avgt    6   176,902 ±   79,175  ns/op
UnitializedArrayBenchmark.allocateInitializedByteArray     10000  avgt    6  1030,709 ±  686,290  ns/op
UnitializedArrayBenchmark.allocateInitializedByteArray    100000  avgt    6  8115,312 ± 1132,039  ns/op
UnitializedArrayBenchmark.allocateUninitializedByteArray       1  avgt    6    18,466 ±    2,311  ns/op
UnitializedArrayBenchmark.allocateUninitializedByteArray      10  avgt    6    19,117 ±    2,626  ns/op
UnitializedArrayBenchmark.allocateUninitializedByteArray     100  avgt    6    22,222 ±   14,708  ns/op
UnitializedArrayBenchmark.allocateUninitializedByteArray    1000  avgt    6   102,574 ±  147,389  ns/op
UnitializedArrayBenchmark.allocateUninitializedByteArray   10000  avgt    6   104,798 ±   71,891  ns/op
UnitializedArrayBenchmark.allocateUninitializedByteArray  100000  avgt    6   103,443 ±   45,122  ns/op

Benchmark result is saved to /Users/norman/Documents/workspace/netty/microbench/target/reports/performance/UnitializedArrayBenchmark.json
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 463.698 sec - in io.netty.microbench.internal.UnitializedArrayBenchmark```

**Comment by normanmaurer** [comment]
Also @fenik17 ...

**Comment by normanmaurer** [comment]
Once we compile with java9 we can remove the reflection thingy.

File: common/src/main/java/io/netty/util/internal/PlatformDependent0.java (hunk 10 above)
Comment (normanmaurer) [inline common/src/main/java/io/netty/util/internal/PlatformDependent0.java]: See https://github.com/netty/netty/pull/6570#issuecomment-290071182

**Comment by fenik17** [comment]
great! 
So, if it work only for Java 9, should we explicitly check a version?
`TRY_ALLOCATE_UNINITIALIZED_ARRAY = JAVA_VERSION >= 8 && tryAllocateUninitializedArray && hasUnsafe();`

**Comment by normanmaurer** [comment]
@fenik17 done

File: microbench/src/main/java/io/netty/microbench/internal/UnitializedArrayBenchmark.java [hunk 11]
```diff
@@ -0,0 +1,80 @@
...
+        }
+        if (!PlatformDependent.hasUnsafe()) {
+            throw new IllegalStateException("Needs Unsafe");
+        }
+    }
+
+    @Setup(Level.Iteration)
+    public void tearDown() throws Exception {
+        // Trigger a GC and sleep for 2 seconds to give it some time to finish GC
+        System.gc();
```
netkins: CRITICAL Don't try to be smarter than the JVM, remove this call to run the garbage collector. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1215)

netkins: CRITICAL Don't try to be smarter than the JVM, remove this call to run the garbage collector. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1215)

nmittler:
@normanmaurer I think 100 bytes may be a little low and it may not behave the same way on all systems. Maybe playing it safe (no pun intended :)) and bumping the threshold up to something like 1KiB would be better. WDYT?

normanmaurer:
@nmittler works for me... @Scottmitch @trustin WDYT ?

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -99,6 +99,7 @@
     private static final AtomicLong DIRECT_MEMORY_COUNTER;
     private static final long DIRECT_MEMORY_LIMIT;
     private static final ThreadLocalRandomProvider RANDOM_PROVIDER;
+    private static final boolean TRY_ALLOCATE_UNINITIALIZED_ARRAY;
```
Scottmitch: can we instead make this an `int` which indicates the size to start allocating using unsafe? Something like `<0` means don't use unsafe, `<=0` means use unsafe for arrays of this size or greater. Default of 512b or 1024b based upon the benchmarks sgtm.

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
Scottmitch: If the exception is not expected do we want this to prevent the application from starting? It looks like we don't really care what the exception is, we just won't use `jdk.internal.misc.Unsafe`.

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -277,8 +283,69 @@ public Object run() {
...
+                        maybeException = e;
+                    } catch (InvocationTargetException e) {
+                        maybeException = e;
+                    }
+                }
+            }
+            ALLOCATE_ARRAY_METHOD = allocateArrayMethod;
+
+            logger.debug("jdk.internal.misc.Unsafe.allocateUninitializedArray(): {}",
+                    ALLOCATE_ARRAY_METHOD != null ? "available" : "unavailable", maybeException);
```
Scottmitch: the way the logging format is we may not have enough `{}` if `maybeException` is not an exception. Should we only include `maybeException` if an exception occurred and have 2 log statements?

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -316,6 +383,19 @@ static ByteBuffer allocateDirectNoCleaner(int capacity) {
         return newDirectBuffer(UNSAFE.allocateMemory(capacity), capacity);
     }
 
+    static byte[] allocateUninitializedArray(int size) {
+        if (ALLOCATE_ARRAY_METHOD == null) {
```
Scottmitch: typically `PlatformDependent` takes care of calling methods in `PlatformDependent0` only if appropriate, and `PlatformDependent0` fails if it is called in an incorrectly initialized environment. Can we follow that convention here too? That will avoid duplicate logic checks in `PlatformDependent` and here.

microbench/src/main/java/io/netty/microbench/internal/UnitializedArrayBenchmark.java
Scottmitch: was this impacting the benchmark, and is this not something JMH accounts for?

common/src/main/java/io/netty/util/internal/PlatformDependent.java
normanmaurer: Sure...

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
normanmaurer: Imho it should never throw an other exception so I thought it may be good not not blindly catch everything

microbench/src/main/java/io/netty/microbench/internal/UnitializedArrayBenchmark.java
normanmaurer: I did not test if it has an impact removing it. I can tell tho

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -157,9 +158,18 @@ public Random current() {
         DIRECT_MEMORY_LIMIT = maxDirectMemory;
         logger.debug("io.netty.maxDirectMemory: {} bytes", maxDirectMemory);
 
+        boolean tryAllocateUninitializedArray =
+                SystemPropertyUtil.getBoolean("io.netty.tryAllocateUninitializedArray", true);
+        TRY_ALLOCATE_UNINITIALIZED_ARRAY = tryAllocateUninitializedArray && hasUnsafe() && javaVersion() >= 9;
+        logger.debug("io.netty.tryAllocateUninitializedArray: {}", tryAllocateUninitializedArray);
```
trustin: Not directly related with this PR, but weren't we used to prepend `-D` to system property names when logging them?

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
trustin: Do we build with JDK 8 now? Then we could delegate this into a separate class that uses `MethodHandle`?

normanmaurer:
@trustin yes we build with java 8.... Good idea about the `MethodHandle`! Let me check as I never used this before :)

normanmaurer:
@trustin let me just merge it with reflection for now. We may want to consider using method handles for a few things in a followup pr.

normanmaurer:
@Scottmitch @trustin @nmittler ok I addressed all comments and used `1024` as default threshold. Also I removed the `System.gc` from the benchmark as it seems to be fine even without it. PTAL again.

fenik17:
> Good idea about the MethodHandle! 

Or even better to use LambdaMetafactory.

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -155,11 +156,22 @@ public Random current() {
...
         }
         DIRECT_MEMORY_LIMIT = maxDirectMemory;
-        logger.debug("io.netty.maxDirectMemory: {} bytes", maxDirectMemory);
+        logger.debug("-Dio.netty.maxDirectMemory: {} bytes", maxDirectMemory);
+
+        int tryAllocateUninitializedArray =
+                SystemPropertyUtil.getInt("io.netty.tryAllocateUninitializedArray", 1024);
+        TRY_ALLOCATE_UNINITIALIZED_ARRAY = javaVersion() >= 9 && PlatformDependent0.hasAllocateArrayMethod() ?
+                tryAllocateUninitializedArray : -1;
+        logger.debug("-Dio.netty.tryAllocateUninitializedArray: {}", tryAllocateUninitializedArray);
```
nmittler: Do you want to log `tryAllocateUninitializedArray` or `TRY_ALLOCATE_UNINITIALIZED_ARRAY` here? Perhaps the latter would be more helpful for debugging? Or just log both?

nmittler: @normanmaurer if you use `Integer.MAX_VALUE` instead of `-1` as the opt-out case (i.e. never have to worry about negative values), then the condition below will become simpler:

```java
return size > TRY_ALLOCATE_UNINITIALIZED_ARRAY ?
    PlatformDependent0.allocateUninitializedArray(size) :
    new byte[size];
```

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -99,6 +99,7 @@
     private static final AtomicLong DIRECT_MEMORY_COUNTER;
     private static final long DIRECT_MEMORY_LIMIT;
     private static final ThreadLocalRandomProvider RANDOM_PROVIDER;
+    private static final int TRY_ALLOCATE_UNINITIALIZED_ARRAY;
```
nmittler: nit: maybe rename to make it more clear how it's used? Something like `UNINITIALIZED_ARRAY_ALLOCATION_THRESHOLD`?

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -155,11 +156,22 @@ public Random current() {
             }
         }
         DIRECT_MEMORY_LIMIT = maxDirectMemory;
-        logger.debug("io.netty.maxDirectMemory: {} bytes", maxDirectMemory);
+        logger.debug("-Dio.netty.maxDirectMemory: {} bytes", maxDirectMemory);
+
+        int tryAllocateUninitializedArray =
+                SystemPropertyUtil.getInt("io.netty.tryAllocateUninitializedArray", 1024);
```
nmittler: nit: If you rename as suggested above, you'll probably want to rename the property to be consistent.

common/src/main/java/io/netty/util/internal/PlatformDependent.java
normanmaurer: fine by me

common/src/main/java/io/netty/util/internal/PlatformDependent.java
normanmaurer: yep let me only log TRY_ALLOCATE_UNINITIALIZED_ARRAY

normanmaurer: I think I would prefer to use -1 as this is the value that we usually use to opt-out

common/src/main/java/io/netty/util/internal/PlatformDependent.java
Scottmitch: lgtm after this comment is addressed.

normanmaurer:
Cherry-picked into 4.1 (e482d933f755d42f2ed0a29776e2599ae07f1806) and 4.0 (ee198f9c35579de4e72240975bb40c05691fca22)

Issue #6574, opened 2017-03-27
IllegalArgumentException for "address" in newDirectBuffer


Erik-OShaughnessy:
### Observed behavior

Running any Apache Spark job results in an IllegalArgumentException.

```
java.lang.IllegalArgumentException: address: -2251768783175664 (expected: >= 0)
	at io.netty.util.internal.ObjectUtil.checkPositiveOrZero(ObjectUtil.java:75)
	at io.netty.util.internal.PlatformDependent0.newDirectBuffer(PlatformDependent0.java:280)
	at io.netty.util.internal.PlatformDependent0.allocateDirectNoCleaner(PlatformDependent0.java:276)
	at io.netty.util.internal.PlatformDependent.allocateDirectNoCleaner(PlatformDependent.java:486)
	at io.netty.buffer.PoolArena$DirectArena.allocateDirect(PoolArena.java:711)
	at io.netty.buffer.PoolArena$DirectArena.newChunk(PoolArena.java:700)
	at io.netty.buffer.PoolArena.allocateNormal(PoolArena.java:237)
	at io.netty.buffer.PoolArena.allocate(PoolArena.java:213)
	at io.netty.buffer.PoolArena.allocate(PoolArena.java:141)
	at io.netty.buffer.PooledByteBufAllocator.newDirectBuffer(PooledByteBufAllocator.java:296)
	at io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:177)
	at io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:168)
	at io.netty.channel.nio.AbstractNioChannel.newDirectBuffer(AbstractNioChannel.java:404)
	at io.netty.channel.nio.AbstractNioByteChannel.filterOutboundMessage(AbstractNioByteChannel.java:275)
	at io.netty.channel.AbstractChannel$AbstractUnsafe.write(AbstractChannel.java:746)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.write(DefaultChannelPipeline.java:1251)
	at io.netty.channel.AbstractChannelHandlerContext.invokeWrite0(AbstractChannelHandlerContext.java:733)
	at io.netty.channel.AbstractChannelHandlerContext.invokeWrite(AbstractChannelHandlerContext.java:725)
	at io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:810)
[...]

PR #6575, opened 2017-03-28
Allow negative memoryAddress when calling PlatformDependent0.newDirec…

normanmaurer:
…tBuffer(...)

Motivation:

When UNSAFE.allocateMemory is returning an address whose high bit is set we currently throw an IllegalArgumentException. This is not correct as it may return a negative number on at least sparc.

Modifications:

- Allow to pass in negative memoryAddress
- Add unit tests

Result:

Correctly validate the memoryAddress and so also work on sparc as expected. Fixes [#6574].

### DIFF
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent0.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent0.java
@@ -317,7 +317,6 @@ final class PlatformDependent0 {
     }
 
     static ByteBuffer newDirectBuffer(long address, int capacity) {
-        ObjectUtil.checkPositiveOrZero(address, "address");
         ObjectUtil.checkPositiveOrZero(capacity, "capacity");
 
         try {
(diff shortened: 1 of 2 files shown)


common/src/main/java/io/netty/util/internal/ObjectUtil.java
```diff
@@ -78,6 +78,28 @@ public static long checkPositiveOrZero(long i, String name) {
     }
 
     /**
+     * Checks that the given argument is not zero. If it is, throws {@link IllegalArgumentException}.
+     * Otherwise, returns the argument.
+     */
+    public static int checkNotZero(int i, String name) {
+        if (i == 0) {
+            throw new IllegalArgumentException(name + ": " + i + " (expected: != 0)");
```
fenik17: `name + ": 0 (expected: != 0)"`

normanmaurer:
Let me just merge this as it only remove a check and add unit tests.

normanmaurer:
Cherry-picked into 4.1 (5c1c14286d29d57acddf32b227722f7b2e113ed4) and 4.0 (d05e20dcbe7519eeed5856d4819193b5e2c0243e)

PR #6582, opened 2017-03-29
UnsafeByteBufUtil errors and simplification

Scottmitch:
Motiviation:
UnsafeByteBufUtil has some bugs related to using an incorrect index, and also omitting the array paramter when dealing with byte[] objects. There is also some simplification possible with respect to type casting, and minor formatting consistentcy issues.

Modifications:
- Ensure indexing is correct when dealing with native memory
- Ensure array is used when dealing with heap memory
- Remove unecessary casts when using long
- Fix formating and alignment

Result:
UnsafeByteBufUtil is more correct and won't access direct memory when heap arrays are used.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
+++ b/buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
@@ -57,30 +57,24 @@ final class UnsafeByteBufUtil {
 
     static int getUnsignedMedium(long address) {
         if (UNALIGNED) {
-            if (BIG_ENDIAN_NATIVE_ORDER) {
-                return (PlatformDependent.getByte(address) & 0xff) |
-                        (PlatformDependent.getShort(address + 1) & 0xffff) << 8;
-            }
-            return (Short.reverseBytes(PlatformDependent.getShort(address)) & 0xffff) << 8 |
-                    PlatformDependent.getByte(address + 2) & 0xff;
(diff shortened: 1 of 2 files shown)


normanmaurer:
@Scottmitch so there should be failing tests on little endian already without this patch ?

normanmaurer:
@Scottmitch could we add some unit tests that try to use `ByteBuf.getInt(...)` and compare it with `ByteBuf.nioBuffer().getInt(...)` same for the other methods. This way we ensure we do everything correctly

Scottmitch:
@normanmaurer - I added some unit tests. PTAL

Scottmitch:
The code which dealt with medium on machines which allowed unaligned memory access is broken, but consistently broken between write/read. So client/server running Netty on the same endian platform should work. However I would expect issues between machines with different endianness, or even talking to non-Netty software when the medium code was used (surprised this has not come up yet).

`getUnsignedMediumLE` was completely broken because of an off by 1 bug for the aligned access code.

normanmaurer:
It it's fixed now ? Or are you saying it's still broken?

Scottmitch:
I was describing the state before this PR ... this PR *should* correct the issues described above.

normanmaurer:
@Scottmitch gotcha... LGTM after CI run completes. Great work

Scottmitch:
4.1 (ef21d5f) 4.0 (99c2d1e)

trustin:
Thank you so much, @Scottmitch !

Issue #6549, opened 2017-03-16
HttpServerKeepAliveHandler closes connection on 204 response


sschepens:
### Expected behavior

Netty Http Server responds 204 without body, maintaining the connection open.

### Actual behavior

`HttpServerKeepAliveHandler` closes the connection unless the response is `Informational` or has `Content-Length` set, but 204 responses are not Informational but should not carry a `Content-Length` header [rfc](https://tools.ietf.org/html/rfc7230#section-3.3.2).
```
A server MUST NOT send a Content-Length header field in any response
   with a status code of 1xx (Informational) or 204 (No Content).  A
   server MUST NOT send a Content-Length header field in any 2xx
   (Successful) response to a CONNECT request
```
A condition regarding 204 responses should be added [here](https://github.com/netty/netty/blob/149916d052180dd2b0e0e98598f349454ea200ba/codec-http/src/main/java/io/netty/handler/codec/http/HttpServerKeepAliveHandler.java#L115).

### Steps to reproduce

Create Http Server adding `https://github.com/netty/netty/blob/149916d052180dd2b0e0e98598f349454ea200ba/codec-http/src/main/java/io/netty/handler/codec/http/HttpServerKeepAliveHandler.java#L115`

Send a 204 response without body and `Content-Length`.

Scottmitch:
see https://github.com/netty/netty/pull/6587

PR #6587, opened 2017-03-30
HttpServerKeepAliveHandler 204 response with no Content-Length should…

Scottmitch:
… keepalive

Motivation:
https://tools.ietf.org/html/rfc7230#section-3.3.2 states that a 204 response MUST NOT include a Content-Length header. If the HTTP version permits keep alive these responses should be treated as keeping the connection alive even if there is no Content-Length header.

Modifications:
- HttpServerKeepAliveHandler#isSelfDefinedMessageLength should account for 204 respones

Result:
Fixes https://github.com/netty/netty/issues/6549.


Scottmitch:
@sschepens PTAL

sschepens:
seems ok :+1: 
thanks @Scottmitch !

Scottmitch:
4.1 (b041f1a) 4.0 (ece022c)

Issue #6573, opened 2017-03-27
Netty DNS resolver ignores provided DNS server address


tomasherman:
### Expected behavior
In netty 4.1.8.Final, following code works fine. However, in 4.1.9 it does not send the request to provided address and sends it to default name server stream instead.

```scala
import java.net.{InetAddress, InetSocketAddress}

import io.netty.channel.nio.NioEventLoopGroup
import io.netty.channel.socket.nio.NioDatagramChannel
import io.netty.resolver.dns.{DnsNameResolverBuilder, DnsServerAddresses}

/** Simple dns resolver for testing stuff
  */
object Cli extends App {
  private val bossGroup = new NioEventLoopGroup()

  val res = new DnsNameResolverBuilder(bossGroup.next())
            .channelType(classOf[NioDatagramChannel])
            .nameServerAddresses(DnsServerAddresses.singleton(new InetSocketAddress("customdns.com", 33220)))
            .build()

  println(res.resolve("testaddress.com").await().get())
}
```

If you run this code with netty 4.1.9 in debug mode, you can see it's sending requests to system resolvers.

I suspect the problem is the `doResolveUncached` in `DnsNameResolver` where you check whether the address stream is null. I think you should check whether the nameServerAddress is null and if it is, you should default to system resolvers. if it is defined however, you should make it into a stream and use it. But maybe im missing something.

### Netty version
4.1.8.Final is fine, 4.1.9.Final is broken
### JVM version (e.g. `java -version`)
1.8
### OS version (e.g. `uname -a`)
mac os

normanmaurer:
@Scottmitch can you please check as this was introduced by your change in 54c9ecf682eeafaaf7c826903c60f1c783b84dea . I think @tomasherman is correct that we should ensure we not override the name servers set by the user in this case.

Scottmitch:
temporary workaround: Use [NoopDnsServerAddressStreamProvider](https://github.com/netty/netty/blob/4.1/resolver-dns/src/main/java/io/netty/resolver/dns/NoopDnsServerAddressStreamProvider.java#L22).

[...]

PR #6588, opened 2017-03-31
DNS Resolve ambiguity in which DNS servers are used during resolution

Scottmitch:
Motivation:
Recently DnsServerAddressStreamProvider was introduced to allow control for each query as to which DNS server should be used for resolution to respect the local host's default DNS server configuration. However resolver-dns also accepts a stream of DNS servers to use by default, but this stream is not host name aware. This creates an ambiguity as to which method is used to determine the DNS server to user during resolution, and in which order. We can remove this ambiguity and provide a more general API by just supporting DnsServerAddressStreamProvider.

Modifications:
- Remove the fixed DnsServerAddresses and instead only accept a DnsServerAddressStreamProvider.
- Add utility methods to help use DnsServerAddressStreamProvider for a single entry, a list of entries, and get the default for the current machine.

Result:
Fixes https://github.com/netty/netty/issues/6573.

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProvider.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProvider.java
@@ -29,8 +29,9 @@ public interface DnsServerAddressStreamProvider {
     /**
      * Ask this provider for the name servers to query for {@code hostname}.
      * @param hostname The hostname for which to lookup the DNS server addressed to use.
-     * @return The {@link DnsServerAddressStream} which should be used to resolve {@code hostname} or {@code null} to
-     * use the default resolvers.
+     *                 If this is the final {@link DnsServerAddressStreamProvider} to be queried then generally empty
+     *                 string or {@code '.'} correspond to the default {@link DnsServerAddressStream}.
+     * @return The {@link DnsServerAddressStream} which should be used to resolve {@code hostname}.
      */
     DnsServerAddressStream nameServerAddressStream(String hostname);
 }
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProviders.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProviders.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.resolver.dns;
+
+import io.netty.util.internal.UnstableApi;
+
+/**
+ * Utility methods related to {@link DnsServerAddressStreamProvider}.
+ */
+@UnstableApi
+public final class DnsServerAddressStreamProviders {
+    // TODO(scott): how is this done on Windows? This may require a JNI call to GetNetworkParams
+    // https://msdn.microsoft.com/en-us/library/aa365968(VS.85).aspx.
+    private static final DnsServerAddressStreamProvider DEFAULT_DNS_SERVER_ADDRESS_STREAM_PROVIDER =
+            UnixResolverDnsServerAddressStreamProvider.parseSilently();
+
+    private DnsServerAddressStreamProviders() {
+    }
+
+    /**
+     * A {@link DnsServerAddressStreamProvider} which inherits the DNS servers from your local host's configuration.
+     * <p>
+     * Note that only macOS and Linux are currently supported.
+     * @return A {@link DnsServerAddressStreamProvider} which inherits the DNS servers from your local host's
+     * configuration.
+     */
+    public static DnsServerAddressStreamProvider platformDefault() {
+        return DEFAULT_DNS_SERVER_ADDRESS_STREAM_PROVIDER;
+    }
+}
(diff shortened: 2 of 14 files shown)


Scottmitch:
@tomasherman - PTAL.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProvider.java
```diff
@@ -25,12 +25,29 @@
  * /etc/resolver</a>.
  */
 @UnstableApi
-public interface DnsServerAddressStreamProvider {
+public abstract class DnsServerAddressStreamProvider {
+    // TODO(scott): how is this done on Windows? This may require a JNI call to GetNetworkParams
+    // https://msdn.microsoft.com/en-us/library/aa365968(VS.85).aspx.
+    private static final DnsServerAddressStreamProvider DEFAULT_DNS_SERVER_ADDRESS_STREAM_PROVIDER =
+            UnixResolverDnsServerAddressStreamProvider.parseSilently();
```
netkins: MAJOR Initialization of io.netty.resolver.dns.DnsServerAddressStreamProvider accesses class io.netty.resolver.dns.UnixResolverDnsServerAddressStreamProvider, which isn't initialized yet [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=findbugs%3AIC_SUPERCLASS_USES_SUBCLASS_DURING_INITIALIZATION)

resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProvider.java
```diff
@@ -25,12 +25,29 @@
  * /etc/resolver</a>.
  */
 @UnstableApi
-public interface DnsServerAddressStreamProvider {
+public abstract class DnsServerAddressStreamProvider {
+    // TODO(scott): how is this done on Windows? This may require a JNI call to GetNetworkParams
```
netkins: INFO Complete the task associated to this TODO comment. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1135)

normanmaurer:
@trustin PTAL as well

resolver-dns/src/test/java/io/netty/resolver/dns/SearchDomainTest.java
```diff
@@ -33,20 +33,19 @@
...
 import static org.junit.Assert.assertTrue;
-import static org.hamcrest.Matchers.not;
-import static org.hamcrest.core.StringContains.containsString;
 
 public class SearchDomainTest {
 
     private DnsNameResolverBuilder newResolver() {
         return new DnsNameResolverBuilder(group.next())
             .channelType(NioDatagramChannel.class)
-            .nameServerAddresses(DnsServerAddresses.singleton(dnsServer.localAddress()))
```
normanmaurer: @Scottmitch I wonder if we should maybe keep the method, mark it as `@deprecated` and just delegate to `nameServerProvider(...)`. This will allow more easily to upgrade for people. I know its marked as unstable but maybe its worth it. WDYT ? Also /cc @trustin

resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProvider.java
normanmaurer: @Scottmitch Please address

resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProvider.java
```diff
@@ -25,12 +25,29 @@
...
+
+    /**
+     * A {@link DnsServerAddressStreamProvider} which inherits the DNS servers from your local host's configuration.
+     * <p>
+     * Note that only macOS and Linux are currently supported.
+     * @return A {@link DnsServerAddressStreamProvider} which inherits the DNS servers from your local host's
+     * configuration.
      */
-    DnsServerAddressStream nameServerAddressStream(String hostname);
+    public static DnsServerAddressStreamProvider platformDefault() {
```
trustin: Could be placed before non-static members?

resolver-dns/src/test/java/io/netty/resolver/dns/SearchDomainTest.java
trustin: +1 although I'm not strong on this given that its an unstable API anyways.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProvider.java
Scottmitch: whoops ... thanks @netkins 👍

resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProvider.java
Scottmitch: I will move this method into another class and change this back to an interface. The motivation for this method creates a circular class loading issue.

resolver-dns/src/test/java/io/netty/resolver/dns/SearchDomainTest.java
Scottmitch: My preference is to break the interface, and force the user to deal with the change in behavior as opposed to silently change behavior behind the scenes. `@deprecated` methods are easy to miss and may result in unexpected/undesirable behavior (as indicated in the issue which motivated this PR). `@deprecated` methods also create additional maintain burden and may limit design decisions. IMO since the API is unstable we should take advantage and make the change.

Scottmitch: We may be able to adapt what the user provides in `nameServerAddresses`, but there is still an ambiguity as to what the precedence should be. I would rather rely on the user explicitly specifying the precedence than guess.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProviders.java
```diff
@@ -0,0 +1,43 @@
...
+package io.netty.resolver.dns;
+
+import io.netty.util.internal.UnstableApi;
+
+/**
+ * Utility methods related to {@link DnsServerAddressStreamProvider}.
+ */
+@UnstableApi
+public final class DnsServerAddressStreamProviders {
+    // TODO(scott): how is this done on Windows? This may require a JNI call to GetNetworkParams
```
netkins: INFO Complete the task associated to this TODO comment. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1135)

Scottmitch:
4.1 (e074df2)

I will merge this change now and I can add back the deprecated method as a followup if you guys feel strongly about it.

PR #6589, opened 2017-03-31
Expose HTTP/2 HpackDecoder

### LINKED ISSUE #6579: Expose the hpack decoder
Hello,

grpc is using the hpack decoder in [GrpcHttp2HeadersDecoder](https://github.com/grpc/grpc-java/blob/master/netty/src/main/java/io/grpc/netty/GrpcHttp2HeadersDecoder.java) and in netty 4.1.9.Final its visibility has been [changed to private](https://github.com/zer0se7en/netty/pull/2/commits/f9001b9fc07a71a9d6eaf0462470416780302107).
This is preventing grpc to be used with netty 4.1.9.Final and with netty-tcnative 2.0.0.Final.

Would it be possible to expose back this feature to the public api?

Thank you,

Reference discussion: https://groups.google.com/forum/#!topic/grpc-io/jkgIbSC9PlI

nmittler:
Motivation:

gRPC (and potentially other libraries) has an optimized header processor that requires direct access to the HpackDecoder.

Modifications:

Make the HpackDecoder and its constructors public.

Result:

Fixes #6579

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
@@ -49,7 +49,7 @@ import static io.netty.util.AsciiString.EMPTY_STRING;
 import static io.netty.util.internal.ObjectUtil.checkPositive;
 import static io.netty.util.internal.ThrowableUtil.unknownStackTrace;
 
-final class HpackDecoder {
+public class HpackDecoder {
     private static final Http2Exception DECODE_ULE_128_DECOMPRESSION_EXCEPTION = unknownStackTrace(
             connectionError(COMPRESSION_ERROR, "HPACK - decompression failure"), HpackDecoder.class,
             "decodeULE128(..)");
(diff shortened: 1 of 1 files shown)


codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
```diff
@@ -49,7 +49,7 @@
 import static io.netty.util.internal.ObjectUtil.checkPositive;
 import static io.netty.util.internal.ThrowableUtil.unknownStackTrace;
 
-final class HpackDecoder {
+public class HpackDecoder {
```
normanmaurer: if we make it public again can we at least make it final ?

nmittler: Yup - done. I wasn't sure at first whether gRPC needed to extend but it looks like it doesn't.

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
```diff
@@ -96,15 +96,15 @@
...
+    public HpackDecoder(long maxHeaderListSize, int initialHuffmanDecodeCapacity) {
         this(maxHeaderListSize, initialHuffmanDecodeCapacity, DEFAULT_HEADER_TABLE_SIZE);
     }
 
     /**
      * Exposed Used for testing only! Default values used in the initial settings frame are overriden intentionally
      * for testing but violate the RFC if used outside the scope of testing.
      */
-    HpackDecoder(long maxHeaderListSize, int initialHuffmanDecodeCapacity, int maxHeaderTableSize) {
+    public HpackDecoder(long maxHeaderListSize, int initialHuffmanDecodeCapacity, int maxHeaderTableSize) {
```
Scottmitch: having this be public can be dangerous as indicated in the comments. Can this stay package private?

Scottmitch:
Lets track followup discussion in #6591

PR #6594, opened 2017-04-01
HTTP/2 Make DefaultHttp2HeadersDecoder's Http2Headers object creation extensible

### LINKED ISSUE #6591: Avoid exposing HPACK classes
HPACK is an internal detail of HTTP/2 and if possible it would be nice to avoid exposing this in the public interface of the codec-http2. Recently PR https://github.com/netty/netty/pull/6589 exposed `HpackDecoder` for extensibility reasons. In this particular case the extensibility point seemed to be focused on creating a different type of `Http2Headers` object than `DefaultHttp2Headers`. I would like to propose an alternative approach to exposing `HpackDecoder` and instead make `DefaultHttp2HeadersDecoder`'s creation of the `Http2Header` object overridable. In this particular case [GrpcHttp2HeadersDecoder](https://github.com/grpc/grpc-java/blob/master/netty/src/main/java/io/grpc/netty/GrpcHttp2HeadersDecoder.java) is duplicating logic which also exists in `DefaultHttp2HeadersDecoder` and IMO would be nice to remove this duplication and consolidate the logic in `DefaultHttp2HeadersDecoder`.

Related:
https://github.com/netty/netty/issues/6579
https://github.com/netty/netty/pull/6589

Scottmitch:
Motivation:
It is generally useful to override DefaultHttp2HeadersDecoder's creation of a new Http2Headers object so more optimized versions can be substituted if the use case allows for it.

Modifications:
- DefaultHttp2HeadersDecoder should support an overridable method to generate the new Http2Headers object for each decode operation

Result:
DefaultHttp2HeadersDecoder is more extensible.
Fixes #6591.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
@@ -112,7 +112,7 @@ public class DefaultHttp2HeadersDecoder implements Http2HeadersDecoder, Http2Hea
     @Override
     public Http2Headers decodeHeaders(int streamId, ByteBuf headerBlock) throws Http2Exception {
         try {
-            final Http2Headers headers = new DefaultHttp2Headers(validateHeaders, (int) headerArraySizeAccumulator);
+            final Http2Headers headers = newHeaders();
             hpackDecoder.decode(streamId, headerBlock, headers);
             headerArraySizeAccumulator = HEADERS_COUNT_WEIGHT_NEW * headers.size() +
                                          HEADERS_COUNT_WEIGHT_HISTORICAL * headerArraySizeAccumulator;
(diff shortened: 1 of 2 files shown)


Scottmitch:
@nmittler @carl-mastrangelo @ejona86 - FYI

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
```diff
@@ -126,4 +126,28 @@ public Http2Headers decodeHeaders(int streamId, ByteBuf headerBlock) throws Http
...
+    /**
+     * A weighted moving average estimating how many headers are expected during the decode process.
+     * @return an estimate of how many headers are expected during the decode process.
+     */
+    protected final int numberOfHeadersGuess() {
+        return (int) headerArraySizeAccumulator;
+    }
+
+    /**
+     * Determine if the headers should be validated as a result of the decode operation.
```
carl-mastrangelo: Determines

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
```diff
@@ -126,4 +126,28 @@ public Http2Headers decodeHeaders(int streamId, ByteBuf headerBlock) throws Http
...
+     * @return {@code true} if the headers should be validated as a result of the decode operation.
+     */
+    protected final boolean validateHeaders() {
+        return validateHeaders;
+    }
+
+    /**
+     * Create a new {@link Http2Headers} object which will store the results of the decode operation.
+     * @return a new {@link Http2Headers} object which will store the results of the decode operation.
+     */
```
carl-mastrangelo: 100% optional: if you use errorprone at all, there is an annotation called ForOverride that lets people know that a method is meant to be overriden.  It statically checks that no one outside of this class calls the method.

Scottmitch: I'm not necessarily against this but lets consider adding that dependency in a followup PR to keep the change isolated.

Scottmitch:
4.1 (225d10e)

PR #6598, opened 2017-04-03
 remove redundant compare

loststarsXD:
Motivation:

Compare before CAS is redundant, cause CAS is not a lock.

Modification:

 remove redundant compare

Result:

just keep CAS


loststarsXD:
@Scottmitch a volatile read will be cheaper than a CAS operation， I will close it.

PR #6599, opened 2017-04-03
Expose the Http2FrameWriter Constructor Argument

mosesn:
Motivation:

DefaultHttp2FrameWriter has constructors that it would be a hassle to
expose as configuration parameters on Http2Codec.  We should instead
allow it to be passed as an argument.

Modifications:

Mark the base constructor for Http2Codec as public.

Result:

Http2Codec's underlying DefaultHttp2FrameWriter can be configured more
flexibly.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Codec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Codec.java
@@ -62,6 +62,7 @@ public final class Http2Codec extends ChannelDuplexHandler {
      *
      * @param server {@code true} this is a server
      * @param bootstrap bootstrap used to instantiate child channels for remotely-created streams.
+     * @param frameLogger handler used to log all frames
      */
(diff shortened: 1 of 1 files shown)


Scottmitch:
IIUC the goal is to minimize the amount of "old API" that is publicly exposed in the "new API". If the constructor pattern is getting too complex for the "new API" maybe we should consider a builder pattern, or some alternative means to simplify construction?

mosesn:
Either switching to the builder pattern or making this constructor public works for me.

mosesn:
@Scottmitch I've written the other solution in #6605 so pick your poison =)

ejona86:
Yeah, exposing the writer seems fishy. As short-term that could be fine, but we need to know what problem is being solved and address it separately. It seems #6605 provides more clarity. I'll respond some there.

mosesn:
OK, will close this one for now and we can continue in #6605.  If we reconsider, we can reopen this ticket.

PR #5786, opened 2016-09-01
Expose the ChannelHandlerContext from SniHandler's select() step to t…

rkapsi:
…he user.

Motivation

I'm looking to harden our SSL impl. a little bit and add some guards agaist certain types of abuse. One can think of invalid hostname strings in the SNI extenstion or invalid SNI handshakes altogether. This will require measuring, velocity tracking and other things.

Modifications

Adding a protected `lookup(ctx, hostname)` method that is called from SniHandler's `select(...)` method which users can override and implement custom behaviour. The default implementation will simply call the AsyncMapper.

Result

It's possible to get a hold onto the ChannelHandlerContext. Users can override that method and do something with it right there or they can delegate it to something else. SniHandler is happy as long as a `Future<SslContext>` is being returned.

PR #6603, opened 2017-04-04
Supplier interface for SslContext

### LINKED ISSUE #6542: Ability to pass along attributes via SslContext
This is related to the #6158 OCSP pull request.

The #6158 PR is demonstrating the most simple way of obtaining and dealing with OCSP stapling but it gets a bit more complicated in practice. In particular when you look at how `SniHandler` and `AsyncMapping` work (which we use).

1. `AsyncMapping` returns `Future<SslContext>`
2. `SniHandler` uses `Future<SslContext>` to construct `SslEngine` + `SslHandler`

... but how do I pass along the info that is needed to obtain a staple or the OCSP staple itself? I'm currently using the delegate pattern around `SslContext` but it gets quickly rather messy and error prone in conjunction with reference counting. It'd be great if it was possible to pass along arbitrary attributes via `SslContext` (e.g. via `AttributeMap` or simply as a `Map<Object, Object>`).

A more general way of looking at is the desire to pass configuration parameters from `SslContext` to "things" that use it. One scenario is described above but other use-cases could be other config parameters for the `SslEngine` such as `setEnabledProtocols()` or some custom behavior for the `SslHandler` etc etc.

Here's some pseudo code to illustrate it. WDYT - happy to PR it.
 
```java
public class MySniHandler extends SniHandler {

  public static final AttributeKey<OcspKey> OCSP_KEY = AttributeKey.valueOf(OcspKey.class, "OCSP_KEY");
  
  public static final AttributeKey<String[]> PROTOCOLS_KEY = AttributeKey.valueOf(String[].class, "PROTOCOLS_KEY");
  
  private final OcspCache<OcspKey, Future<byte[]>> ocspCache;
  
  public MySniHandler(AsyncMapping<String, SslContext> mapping, OcspCache<OcspKey, Future<byte[]>> ocspCache) {
    super(mapping);
    this.ocspCache = ocspCache;
  }

  @Override
  protected void replaceHandler(ChannelHandlerContext ctx, String hostname, SslContext sslContext) throws Exception {
    SslHandler sslHandler = sslContext.newHandler(ctx.alloc());
    ReferenceCountedOpenSslEngine engine = (ReferenceCountedOpenSslEngine) sslHandler.engine();
    
    OcspKey ocspKey = sslContext.attr(OCSP_KEY).get();
    
    // I (the server) may or may not have a staple and I don't want
    // to make the client wait for it.
    Future<byte[]> future = ocspCache.get(ocspKey);
    if (future.isSuccess()) {
      engine.setOcspResponse(future.getNow());
    }
    
    String[] protocols = sslContext.attr(PROTOCOLS_KEY).get();
    engine.setEnabledProtocols(protocols);
    
    ctx.pipeline().replace(this, SslHandler.class.getName(), sslHandler);
  }
}
```

rkapsi:
Motivation

There are use-cases (in particular with SniHandler+AsyncMapper) where it's desireable to be able to return more than just the SslContext from the mapper. It could be configuration data, debug information etc. The only way to do that right now is to employ a delegation pattern for SslContext but it's very maintanace intensive and always subject to breakage.

Modifications

Introduce a generic Supplier interface (may be replaced by JDK8's own version at a future point) and let SniHandler depend on Future<Supplier<SslContext>> instead of Future<SslContext>. The user can implement a custom Supplier that can be used to pass arbitary data  beyond the SslContext itself into SniHandler.

Result

Fixes #6542

### DIFF
--- a/common/src/main/java/io/netty/util/AsyncMapping.java
+++ b/common/src/main/java/io/netty/util/AsyncMapping.java
@@ -18,11 +18,11 @@ package io.netty.util;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.Promise;
 
-public interface AsyncMapping<U, R> {
+public interface AsyncMapping<IN, OUT> {
 
     /**
      * Returns the {@link Future} that will provide the result of the mapping. The given {@link Promise} will
      * be fulfilled when the result is available.
      */
-    Future<R> map(U input, Promise<R> promise);
+    Future<OUT> map(IN input, Promise<OUT> promise);
 }
--- a/handler/src/main/java/io/netty/handler/ssl/FixedSslContextSupplier.java
+++ b/handler/src/main/java/io/netty/handler/ssl/FixedSslContextSupplier.java
@@ -21,10 +21,10 @@ import io.netty.util.internal.ObjectUtil;
 /**
  * A {@link Supplier} for {@link SslContext}s.
  */
-public final class SslContextSupplier implements Supplier<SslContext> {
+final class FixedSslContextSupplier implements Supplier<SslContext> {
     private final SslContext sslContext;
 
-    public SslContextSupplier(SslContext sslContext) {
+    public FixedSslContextSupplier(SslContext sslContext) {
         this.sslContext = ObjectUtil.checkNotNull(sslContext, "sslContext");
     }
 
--- a/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
@@ -30,6 +30,7 @@ import io.netty.handler.codec.DecoderException;
 import io.netty.util.AsyncMapping;
 import io.netty.util.CharsetUtil;
 import io.netty.util.DomainNameMapping;
+import io.netty.util.FixedSupplier;
 import io.netty.util.Mapping;
 import io.netty.util.ReferenceCountUtil;
 import io.netty.util.Supplier;
@@ -364,7 +365,7 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
         } else {
             // If the user returned something other than a SslContext or Supplier<SslContext>
             // in the Future<?> then let it blow up with a ClassCastException
-            supplier = new FixedSslContextSupplier((SslContext) value);
+            supplier = new FixedSupplier<SslContext>((SslContext) value);
         }
 
         selection = new Selection(supplier, hostname);
@@ -486,7 +487,7 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
         private static void onOperationComplete(Future<? extends SslContext> future,
                 Promise<Supplier<? extends SslContext>> promise) {
             if (future.isSuccess()) {
-                promise.setSuccess(new FixedSslContextSupplier(future.getNow()));
+                promise.setSuccess(new FixedSupplier<SslContext>(future.getNow()));
             } else {
                 promise.setFailure(future.cause());
             }
@@ -507,7 +508,7 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
             final Supplier<SslContext> supplier;
             try {
                 SslContext sslContent = mapping.map(input);
-                supplier = new FixedSslContextSupplier(sslContent);
+                supplier = new FixedSupplier<SslContext>(sslContent);
             } catch (Throwable cause) {
                 return promise.setFailure(cause);
             }
--- a/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
@@ -92,7 +92,7 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
      * @param mapping the mapping of domain name to {@link SslContext}
      */
     public SniHandler(AsyncMapping<? super String, ? extends SslContext> mapping) {
-        this (new AsyncMappingAdapter(mapping));
+        this(new AsyncMappingAdapter(mapping));
     }
 
     /**
@@ -115,7 +115,7 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
     /**
      * @return the selected {@link Supplier} for {@link SslContext}
      */
-    public Supplier<? extends SslContext> sslContextSupplier() {
+    protected Supplier<? extends SslContext> sslContextSupplier() {
         return selection.supplier;
     }
 
@@ -364,7 +364,7 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
         } else {
             // If the user returned something other than a SslContext or Supplier<SslContext>
             // in the Future<?> then let it blow up with a ClassCastException
-            supplier = new SslContextSupplier((SslContext) value);
+            supplier = new FixedSslContextSupplier((SslContext) value);
         }
 
         selection = new Selection(supplier, hostname);
@@ -455,18 +455,18 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
     }
 
     private static final class AsyncMappingAdapter implements SslContextMapping {
-        private final AsyncMapping<? super String, ? extends SslContext> mapping;
-
-        private AsyncMappingAdapter(AsyncMapping<? super String, ? extends SslContext> mapping) {
-            this.mapping = ObjectUtil.checkNotNull(mapping, "mapping");
-        }
+        private final AsyncMapping<String, SslContext> mapping;
 
         @SuppressWarnings("unchecked")
+        AsyncMappingAdapter(AsyncMapping<? super String, ? extends SslContext> mapping) {
+            this.mapping = (AsyncMapping<String, SslContext>) ObjectUtil.checkNotNull(mapping, "mapping");
+        }
+
         @Override
         public Future<Supplier<? extends SslContext>> map(String input,
                 final Promise<Supplier<? extends SslContext>> promise) {
 
-            Future<SslContext> future = ((AsyncMapping<String, SslContext>) mapping).map(input,
+            Future<SslContext> future = mapping.map(input,
                     ImmediateEventExecutor.INSTANCE.<SslContext>newPromise());
 
             if (future.isDone()) {
@@ -475,11 +475,7 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
                 future.addListener(new FutureListener<SslContext>() {
                     @Override
                     public void operationComplete(Future<SslContext> future) {
-                        try {
-                            onOperationComplete(future, promise);
-                        } catch (Exception err) {
-                            promise.setFailure(err);
-                        }
+                        onOperationComplete(future, promise);
                     }
                 });
             }
@@ -489,9 +485,8 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
 
         private static void onOperationComplete(Future<? extends SslContext> future,
                 Promise<Supplier<? extends SslContext>> promise) {
-
             if (future.isSuccess()) {
-                promise.setSuccess(new SslContextSupplier(future.getNow()));
+                promise.setSuccess(new FixedSslContextSupplier(future.getNow()));
             } else {
                 promise.setFailure(future.cause());
             }
@@ -501,7 +496,7 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
     private static final class MappingAdapter implements SslContextMapping {
         private final Mapping<? super String, ? extends SslContext> mapping;
 
-        private MappingAdapter(Mapping<? super String, ? extends SslContext> mapping) {
+        MappingAdapter(Mapping<? super String, ? extends SslContext> mapping) {
             this.mapping = ObjectUtil.checkNotNull(mapping, "mapping");
         }
 
@@ -512,7 +507,7 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
             final Supplier<SslContext> supplier;
             try {
                 SslContext sslContent = mapping.map(input);
-                supplier = new SslContextSupplier(sslContent);
+                supplier = new FixedSslContextSupplier(sslContent);
             } catch (Throwable cause) {
                 return promise.setFailure(cause);
             }
--- a/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
@@ -465,7 +465,7 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
 
         @Override
         public Future<Supplier<? extends SslContext>> map(String input,
-                final Promise<Supplier<? extends SslContext>> promise) {
+                                                          final Promise<Supplier<? extends SslContext>> promise) {
 
             Future<SslContext> future = mapping.map(input,
                     ImmediateEventExecutor.INSTANCE.<SslContext>newPromise());
@@ -485,7 +485,7 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
         }
 
         private static void onOperationComplete(Future<? extends SslContext> future,
-                Promise<Supplier<? extends SslContext>> promise) {
+                                                Promise<Supplier<? extends SslContext>> promise) {
             if (future.isSuccess()) {
                 promise.setSuccess(new FixedSupplier<SslContext>(future.getNow()));
             } else {
@@ -503,7 +503,7 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
 
         @Override
         public Future<Supplier<? extends SslContext>> map(String input,
-                Promise<Supplier<? extends SslContext>> promise) {
+                                                          Promise<Supplier<? extends SslContext>> promise) {
 
             final Supplier<SslContext> supplier;
             try {
--- a/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
@@ -32,8 +32,10 @@ import io.netty.util.CharsetUtil;
 import io.netty.util.DomainNameMapping;
 import io.netty.util.Mapping;
 import io.netty.util.ReferenceCountUtil;
+import io.netty.util.Supplier;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.FutureListener;
+import io.netty.util.concurrent.ImmediateEventExecutor;
 import io.netty.util.concurrent.Promise;
 import io.netty.util.internal.ObjectUtil;
 import io.netty.util.internal.PlatformDependent;
@@ -56,7 +58,7 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
             InternalLoggerFactory.getInstance(SniHandler.class);
     private static final Selection EMPTY_SELECTION = new Selection(null, null);
 
-    protected final AsyncMapping<String, SslContext> mapping;
+    protected final SslContextMapping mapping;
 
     private boolean handshakeFailed;
     private boolean suppressRead;
@@ -70,7 +72,7 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
      * @param mapping the mapping of domain name to {@link SslContext}
      */
     public SniHandler(Mapping<? super String, ? extends SslContext> mapping) {
-        this(new AsyncMappingAdapter(mapping));
+        this(new MappingAdapter(mapping));
     }
 
     /**
@@ -89,9 +91,18 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
      *
      * @param mapping the mapping of domain name to {@link SslContext}
      */
-    @SuppressWarnings("unchecked")
     public SniHandler(AsyncMapping<? super String, ? extends SslContext> mapping) {
-        this.mapping = (AsyncMapping<String, SslContext>) ObjectUtil.checkNotNull(mapping, "mapping");
+        this (new AsyncMappingAdapter(mapping));
+    }
+
+    /**
+     * Creates a SNI detection handler with configured {@link SslContext}
+     * maintained by {@link SslContextMapping}
+     *
+     * @param mapping the mapping of domain name to {@link SslContext}
+     */
+    public SniHandler(SslContextMapping mapping) {
+        this.mapping = ObjectUtil.checkNotNull(mapping, "mapping");
     }
 
     /**
@@ -102,10 +113,19 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
     }
 
     /**
+     * @return the selected {@link Supplier} for {@link SslContext}
+     */
+    public Supplier<? extends SslContext> sslContextSupplier() {
+        return selection.supplier;
+    }
+
+    /**
+     * @see #sslContextSupplier()
      * @return the selected {@link SslContext}
      */
     public SslContext sslContext() {
-        return selection.context;
+        Supplier<? extends SslContext> supplier = sslContextSupplier();
+        return supplier != null ? supplier.get() : null;
     }
 
     @Override
@@ -277,23 +297,23 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
     }
 
     private void select(final ChannelHandlerContext ctx, final String hostname) throws Exception {
-        Future<SslContext> future = lookup(ctx, hostname);
+        Future<?> future = lookup(ctx, hostname);
         if (future.isDone()) {
             if (future.isSuccess()) {
-                onSslContext(ctx, hostname, future.getNow());
+                onSslContextOrSupplier(ctx, hostname, future.getNow());
             } else {
                 throw new DecoderException("failed to get the SslContext for " + hostname, future.cause());
             }
         } else {
             suppressRead = true;
-            future.addListener(new FutureListener<SslContext>() {
+            future.addListener(new FutureListener<Object>() {
                 @Override
-                public void operationComplete(Future<SslContext> future) throws Exception {
+                public void operationComplete(Future<Object> future) throws Exception {
                     try {
                         suppressRead = false;
                         if (future.isSuccess()) {
                             try {
-                                onSslContext(ctx, hostname, future.getNow());
+                                onSslContextOrSupplier(ctx, hostname, future.getNow());
                             } catch (Throwable cause) {
                                 ctx.fireExceptionCaught(new DecoderException(cause));
                             }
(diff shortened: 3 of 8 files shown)


handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -56,7 +58,7 @@
             InternalLoggerFactory.getInstance(SniHandler.class);
     private static final Selection EMPTY_SELECTION = new Selection(null, null);
 
-    protected final AsyncMapping<String, SslContext> mapping;
+    protected final SslContextMapping mapping;
```
rkapsi: ATTENTION: I think this is the only thing that breaks the API (unfortunately). If that's OK then I'd actually recommend to make this field private.

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -102,10 +113,19 @@ public String hostname() {
     }
 
     /**
+     * @return the selected {@link Supplier} for {@link SslContext}
+     */
+    public Supplier<? extends SslContext> sslContextSupplier() {
```
netkins: MAJOR Remove usage of generic wildcard type. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1452)

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -70,7 +72,7 @@
      * @param mapping the mapping of domain name to {@link SslContext}
      */
     public SniHandler(Mapping<? super String, ? extends SslContext> mapping) {
-        this(new AsyncMappingAdapter(mapping));
+        this(new MappingAdapter(mapping));
```
netkins: MAJOR Avoid instantiation through private constructors from outside of the constructors class. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AAccessorClassGeneration)

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -313,31 +333,58 @@ public void operationComplete(Future<SslContext> future) throws Exception {
...
      *
-     * @see AsyncMapping#map(Object, Promise)
+     * @see SslContextMapping#map(String, Promise)
+     * @deprecated the return value will at some point change from {@code Future<SslContext>} to
+     *   {@code Future<Supplier<? extends SslContext>>}.
      */
-    protected Future<SslContext> lookup(ChannelHandlerContext ctx, String hostname) throws Exception {
-        return mapping.map(hostname, ctx.executor().<SslContext>newPromise());
+    @Deprecated
+    protected Future<?> lookup(ChannelHandlerContext ctx, String hostname) throws Exception {
```
netkins: MAJOR Remove usage of generic wildcard type. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1452)

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -89,9 +91,18 @@ public SniHandler(DomainNameMapping<? extends SslContext> mapping) {
      *
      * @param mapping the mapping of domain name to {@link SslContext}
      */
-    @SuppressWarnings("unchecked")
     public SniHandler(AsyncMapping<? super String, ? extends SslContext> mapping) {
-        this.mapping = (AsyncMapping<String, SslContext>) ObjectUtil.checkNotNull(mapping, "mapping");
+        this (new AsyncMappingAdapter(mapping));
```
netkins: MAJOR Avoid instantiation through private constructors from outside of the constructors class. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AAccessorClassGeneration)

handler/src/main/java/io/netty/handler/ssl/SslContextSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package io.netty.handler.ssl;
+
+import io.netty.util.Supplier;
+import io.netty.util.internal.ObjectUtil;
+
+/**
+ * A {@link Supplier} for {@link SslContext}s.
+ */
+public final class SslContextSupplier implements Supplier<SslContext> {
```
Scottmitch: `SslContextSupplier` -> `FixedSslContextSupplier`.

Can you make this more general purpose?

```java
public final class FIxedSupplier<T> implements Supplier<T> {
   private final T object;
...
}
```

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -407,31 +454,78 @@ public void flush(ChannelHandlerContext ctx) throws Exception {
         ctx.flush();
     }
 
-    private static final class AsyncMappingAdapter implements AsyncMapping<String, SslContext> {
+    private static final class AsyncMappingAdapter implements SslContextMapping {
+        private final AsyncMapping<? super String, ? extends SslContext> mapping;
+
+        private AsyncMappingAdapter(AsyncMapping<? super String, ? extends SslContext> mapping) {
```
Scottmitch: make this package private

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
Scottmitch: My assumption is it is not commonly directly referenced, and since you introduced this change we know you don't depend upon it, or are at least willing to absorb the change.

I'm inclined to allow it, but I have a hunch @normanmaurer may not be so lenient.

good reminder to only expose what is necessary https://github.com/netty/netty/pull/5786#r77206033 ... especially member variables which are very difficult to change after the fact.

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
Scottmitch: also remove extra space after `this`

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
Scottmitch: can this method be protected?

handler/src/main/java/io/netty/handler/ssl/SslContextMapping.java
```diff
@@ -0,0 +1,27 @@
...
+
+import io.netty.util.AsyncMapping;
+import io.netty.util.Supplier;
+
+/**
+ * A map of hostnames to {@link Supplier}s of {@link SslContext}s.
+ *
+ * @see SniHandler
+ */
+public interface SslContextMapping extends AsyncMapping<String, Supplier<? extends SslContext>> {
```
Scottmitch: is the intention with the wildcards to make it easier to work with different types of concrete SslContext, or at the differences between least JDK/OpenSSL? what do we lose if we get rid of the wild cards?

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -313,31 +333,58 @@ public void operationComplete(Future<SslContext> future) throws Exception {
...
-     * users can override this method to implement custom behavior.
+     * The default implementation will simply call {@link SslContextMapping#map(String, Promise)} but users can
+     * override this method to implement custom behavior.
+     *
+     * NOTE: This method returns a {@code Future<?>} to accommodate the old and new way of doing things without
+     * braking the API. Users who override this method are advised to change their return value from
+     * {@code Future<SslContext>} to {@code Future<Supplier<? extends SslContext>>}.
      *
-     * @see AsyncMapping#map(Object, Promise)
+     * @see SslContextMapping#map(String, Promise)
```
Scottmitch: `SslContextMapping#map(Object, Promise)`

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -313,31 +333,58 @@ public void operationComplete(Future<SslContext> future) throws Exception {
     }
 
     /**
-     * The default implementation will simply call {@link AsyncMapping#map(Object, Promise)} but
-     * users can override this method to implement custom behavior.
+     * The default implementation will simply call {@link SslContextMapping#map(String, Promise)} but users can
```
Scottmitch: `SslContextMapping#map(Object, Promise)`

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -313,31 +333,58 @@ public void operationComplete(Future<SslContext> future) throws Exception {
...
             selection = EMPTY_SELECTION;
             PlatformDependent.throwException(cause);
         }
     }
 
     /**
+     * @see #replaceHandler(ChannelHandlerContext, String, SslContext)
+     */
+    protected void replaceHandler(ChannelHandlerContext ctx, String hostname,
+            Supplier<? extends SslContext> supplier) throws Exception {
```
Scottmitch: nit: fix indentation

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -407,31 +454,78 @@ public void flush(ChannelHandlerContext ctx) throws Exception {
...
+        private AsyncMappingAdapter(AsyncMapping<? super String, ? extends SslContext> mapping) {
+            this.mapping = ObjectUtil.checkNotNull(mapping, "mapping");
+        }
+
+        @SuppressWarnings("unchecked")
+        @Override
+        public Future<Supplier<? extends SslContext>> map(String input,
+                final Promise<Supplier<? extends SslContext>> promise) {
+
+            Future<SslContext> future = ((AsyncMapping<String, SslContext>) mapping).map(input,
```
Scottmitch: can we just do this cast once in the constructor? This should preserve the previous behavior.

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -407,31 +454,78 @@ public void flush(ChannelHandlerContext ctx) throws Exception {
...
+                    }
+                });
+            }
+
+            return promise;
+        }
+
+        private static void onOperationComplete(Future<? extends SslContext> future,
+                Promise<Supplier<? extends SslContext>> promise) {
+
```
Scottmitch: nit: kill line

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -407,31 +454,78 @@ public void flush(ChannelHandlerContext ctx) throws Exception {
...
+                        }
+                    }
+                });
+            }
+
+            return promise;
+        }
+
+        private static void onOperationComplete(Future<? extends SslContext> future,
+                Promise<Supplier<? extends SslContext>> promise) {
```
Scottmitch: nit: fix indentation

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -407,31 +454,78 @@ public void flush(ChannelHandlerContext ctx) throws Exception {
...
+
+            if (future.isDone()) {
+                onOperationComplete(future, promise);
+            } else {
+                future.addListener(new FutureListener<SslContext>() {
+                    @Override
+                    public void operationComplete(Future<SslContext> future) {
+                        try {
+                            onOperationComplete(future, promise);
+                        } catch (Exception err) {
```
Scottmitch: consider removing this ... if an exception is thrown we don't handle it for the `future.isDone()` path ... and it looks like the `onOperationComplete` is not expected to throw.

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -313,31 +333,58 @@ public void operationComplete(Future<SslContext> future) throws Exception {
...
+     * The default implementation will simply call {@link SslContextMapping#map(String, Promise)} but users can
+     * override this method to implement custom behavior.
+     *
+     * NOTE: This method returns a {@code Future<?>} to accommodate the old and new way of doing things without
+     * braking the API. Users who override this method are advised to change their return value from
+     * {@code Future<SslContext>} to {@code Future<Supplier<? extends SslContext>>}.
      *
-     * @see AsyncMapping#map(Object, Promise)
+     * @see SslContextMapping#map(String, Promise)
+     * @deprecated the return value will at some point change from {@code Future<SslContext>} to
```
Scottmitch: can you add this new method now, reference it from the deprecated statement, and only use the new method so we have stronger type guarantees with the generics and less casting?

common/src/main/java/io/netty/util/AsyncMapping.java
```diff
@@ -18,11 +18,11 @@
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.Promise;
 
-public interface AsyncMapping<IN, OUT> {
+public interface AsyncMapping<U, R> {
```
Scottmitch: can you explain the motivation for this change? `U` doesn't seem very clear to me.

rkapsi: For me `IN` and `OUT` are strange but `U` and `R` commonly used in JDK8+ functional interfaces and classes that build on top of it. E.g. [CompletableFuture](http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#completedFuture-U-).

`U` as "user supplied value" and `R` as "return value".

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
rkapsi: Because `SslContextMapping extends AsyncMapping<String, ...>` it's `SslContextMapping#map(String, Promise)`.

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
rkapsi: What's the convention here? Most code seems to use "Indent continuation" of 2x the normal indention. You want it to be block formatted?

```java
private static void onOperationComplete(Future<? extends SslContext> future,
                                        Promise<Supplier<? extends SslContext>> promise) {

}
```

handler/src/main/java/io/netty/handler/ssl/SslContextMapping.java
rkapsi: That would be the thought but I think the interface should be this instead:

```java
interface SslContextMapping<T extends SslContext> extends AsyncMapping<String, T> {}
```

From Netty's POV we don't gain or lose anything if it's just SslContext. From the user's POV we lose the ability to clamp down on the generics and facilitate the compiler for type/input validation.

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
rkapsi: Not without breaking the API but it's possible that we're the only ones overriding it in our extended SniHandler. :thinking:

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -102,10 +113,19 @@ public String hostname() {
     }
 
     /**
+     * @return the selected {@link Supplier} for {@link SslContext}
+     */
+    protected Supplier<? extends SslContext> sslContextSupplier() {
```
netkins: MAJOR Remove usage of generic wildcard type. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1452)

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
Scottmitch: Either Intellij is confused or javadoc references are subject to type erasure. Based upon a reference in the JDKs [HashMap](http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/HashMap.java#l552) I suspect its the latter. Can you confirm this?

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -313,31 +333,58 @@ public void operationComplete(Future<SslContext> future) throws Exception {
...
+     * The default implementation will simply call {@link SslContextMapping#map(String, Promise)} but users can
+     * override this method to implement custom behavior.
      *
-     * @see AsyncMapping#map(Object, Promise)
+     * NOTE: This method returns a {@code Future<?>} to accommodate the old and new way of doing things without
+     * braking the API. Users who override this method are advised to change their return value from
+     * {@code Future<SslContext>} to {@code Future<Supplier<? extends SslContext>>}.
+     *
+     * @see SslContextMapping#map(String, Promise)
+     * @deprecated the return value will at some point change from {@code Future<SslContext>} to
```
Scottmitch: Originally in https://github.com/netty/netty/pull/6603#discussion_r110068929 but the conversation was collapsed

me
> can you add this new method now, reference it from the deprecated statement, and only use the new method so we have stronger type guarantees with the generics and less casting?

@rkapsi 
> Not without breaking the API but it's possible that we're the only ones overriding it in our extended SniHandler. 🤔

I don't think we have to break the API. What about just adding a new method with a different name?

```java
    /**
     * The default implementation will simply call {@link SslContextMapping#map(Object, Promise)} but users can
     * override this method to implement custom behavior.
     *
     * NOTE: This method returns a {@code Future<?>} to accommodate the old and new way of doing things without
     * braking the API. Users who override this method are advised to change their return value from
     * {@code Future<SslContext>} to {@code Future<Supplier<? extends SslContext>>}.
     *
     * @see SslContextMapping#map(Object, Promise)
     * @deprecated Use {@link #lookupSupplier(io.netty.channel.ChannelHandlerContext, String)}.
     */
    @Deprecated
    protected Future<SslContext> lookup(ChannelHandlerContext ctx, String hostname) throws Exception {
        final Promise<SslContext> promise = ctx.executor().newPromise();
        lookupSupplier(ctx, hostname).addListener(new GenericFutureListener<Future<Supplier<? extends SslContext>>>() {
            @Override
            public void operationComplete(Future<Supplier<? extends SslContext>> future) throws Exception {
                if (future.isSuccess()) {
                    // todo: do we care about null here, or if the supplier throws?
                    promise.setSuccess(future.getNow().get());
                } else if (future.isCancelled()) {
                    promise.cancel(false);
                } else {
                    promise.setFailure(future.cause());
                }
            }
        });
        return promise;
    }

    protected Future<Supplier<? extends SslContext>> lookupSupplier(ChannelHandlerContext ctx,
                                                                    String hostname) throws Exception {
        Promise<Supplier<? extends SslContext>> promise = ctx.executor().newPromise();
        return mapping.map(hostname, promise);
    }
```

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
Scottmitch: Generally line the arguments up, unless the max line length would be exceeded.

handler/src/main/java/io/netty/handler/ssl/SslContextMapping.java
Scottmitch: Do you think different supplier types would be co-mingled or even co-exist? I would assume most users use either JDK or OpenSSL, but not both. I'm wondering if it is practically useful from the user's perspective. If not then perhaps we can get rid of these generics. If there is not a concrete use case I would prefer to just use `Supplier<SslContext>` for now to keep things simple. We can mark the interfaces as unstable and add the wildcards later if there is no immediate use for them.

handler/src/main/java/io/netty/handler/ssl/FixedSslContextSupplier.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package io.netty.handler.ssl;
+
+import io.netty.util.Supplier;
+import io.netty.util.internal.ObjectUtil;
+
+/**
+ * A {@link Supplier} for {@link SslContext}s.
+ */
+final class FixedSslContextSupplier implements Supplier<SslContext> {
```
Scottmitch: from https://github.com/netty/netty/pull/6603#discussion_r110066594. why not something like:

```java
    public final class FixedSupplier<T> implements Supplier<T> {
        private final T value;

        public FixedSupplier(T value) {
            this.value = value;
        }

        @Override
        public T get() {
            return null;
        }
    }
```

common/src/main/java/io/netty/util/AsyncMapping.java
Scottmitch: Does `U` mean anything, or is it an alternative when `T` is already taken? `P` for parameter type, and `R` for return type would make more sense to me than `U`. However I'm not sure it is worth moving away `IN` and `OUT` which seems more descriptive. Can we handle this as a followup PR?

handler/src/main/java/io/netty/handler/ssl/FixedSslContextSupplier.java
rkapsi: That works. Our app is in JDK8+ and under normal circumstances I'd simply write `() -> value` and everything else would be something more specific than a simple implementation.

In that regard, is there a roadmap for Netty being JDK8+?

rkapsi:
Strange, I can't reply to the `U` vs. `IN` comment. Anyways, I use `U` in `@FunctionalInterface` for **U**ser provided arguments and `R` is the return type. I'll revert for a cleaner focused PR.

common/src/main/java/io/netty/util/AsyncMapping.java
rkapsi: Strange, I can't reply to the `U` vs. `IN` comment. Anyways, I use `U` in `@FunctionalInterface` for **U**ser provided arguments and `R` is the return type. I'll revert for a cleaner focused PR.

handler/src/main/java/io/netty/handler/ssl/SslContextMapping.java
rkapsi: So... The co-existing thing is not too crazy. Our Netty app can listen on multiple network interfaces (think Anycast, ECMP), each with it's own slightly different config and we could totally do a JDK vs. OPENSSL showdown (in that regard it'd be actually nicer to let openssl, libressl, boringssl co-exist). But for that we don't need open ended suppliers and `Supplier<SslContext>` would be fine (if not desirable).

I'm thinking in terms of being able to pin my API which is going to be consumed by yet somebody else to a certain implementation of SslContext.

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
rkapsi: They're subject to type erasure but notice that the `IN` parameter of `SslContextMapping` is a concrete type with concrete argument types. There's no casting I could do to erase the String from the interface.

```java
interface SslContextMapping<T> extends AsyncMapping<String, T>
```

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
rkapsi: There is unfortunately a little problem with that. What does SniHandler's `select(...)` method call? If it's `lookup(...)` then the whole Supplier thing gets rendered useless because we do want `replaceHandler(..., Supplier)` to be called.

If it's `lookupSupplier(...)` then here's no way to accommodate the `lookup(...)` into it.

handler/src/main/java/io/netty/handler/ssl/FixedSslContextSupplier.java
Scottmitch: > is there a roadmap for Netty being JDK8+?

This is not planned for 4.x. However we will very likely raise the minimum required JDK version for future versions of Netty (see https://github.com/netty/netty/issues/5430#issuecomment-227178314))

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
rkapsi: WDYT of the following...

Turn `SniHandler` into an `AbstractSniHandler` with an `protected abstract` `select(...)` method.

Let there be a "new" `SniHandler` that extends `AbstractSniHandler` and let it implement the current behavior of using SslContext directly.

Let there be a "real new" `SniWithSupplierHandler` (better name TBD) that extends `AbstractSniHandler` and implements the Supplier based SslContext acquisition. In some ways we could omit this (and the Supplier stuff altogether) because everyone wanting to go this route can built it themselves on top of AbstractSniHandler and we can revisit the issue when more people ask for a concrete impl.

Scottmitch: Good point. Let me clarify an approach:
1. `select` should be calling `lookupSupplier`
2. `lookupSupplier` should call `lookup` (by default)
3. `lookup` should call the `mapping` function.

This should preserve backwards compatibility by default. The only exception is still the type change of `mapping`.

```diff
diff --git a/handler/src/main/java/io/netty/handler/ssl/SniHandler.java b/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
index ec17e0e..201b25f 100644
--- a/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
@@ -15,11 +15,6 @@
  */
 package io.netty.handler.ssl;
 
-import java.net.IDN;
-import java.net.SocketAddress;
-import java.util.List;
-import java.util.Locale;
-
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufUtil;
 import io.netty.channel.ChannelHandlerContext;
@@ -35,6 +30,7 @@ import io.netty.util.ReferenceCountUtil;
 import io.netty.util.Supplier;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.FutureListener;
+import io.netty.util.concurrent.GenericFutureListener;
 import io.netty.util.concurrent.ImmediateEventExecutor;
 import io.netty.util.concurrent.Promise;
 import io.netty.util.internal.ObjectUtil;
@@ -42,6 +38,11 @@ import io.netty.util.internal.PlatformDependent;
 import io.netty.util.internal.logging.InternalLogger;
 import io.netty.util.internal.logging.InternalLoggerFactory;
 
+import java.net.IDN;
+import java.net.SocketAddress;
+import java.util.List;
+import java.util.Locale;
+
 /**
  * <p>Enables <a href="https://tools.ietf.org/html/rfc3546#section-3.1">SNI
  * (Server Name Indication)</a> extension for server side SSL. For clients
@@ -297,23 +298,23 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
     }
 
     private void select(final ChannelHandlerContext ctx, final String hostname) throws Exception {
-        Future<?> future = lookup(ctx, hostname);
+        Future<Supplier<? extends SslContext>> future = lookupSupplier(ctx, hostname);
         if (future.isDone()) {
             if (future.isSuccess()) {
-                onSslContextOrSupplier(ctx, hostname, future.getNow());
+                onSslContextSupplier(ctx, hostname, future.getNow());
             } else {
                 throw new DecoderException("failed to get the SslContext for " + hostname, future.cause());
             }
         } else {
             suppressRead = true;
-            future.addListener(new FutureListener<Object>() {
+            future.addListener(new FutureListener<Supplier<? extends SslContext>>() {
                 @Override
-                public void operationComplete(Future<Object> future) throws Exception {
+                public void operationComplete(Future<Supplier<? extends SslContext>> future) throws Exception {
                     try {
                         suppressRead = false;
                         if (future.isSuccess()) {
                             try {
-                                onSslContextOrSupplier(ctx, hostname, future.getNow());
+                                onSslContextSupplier(ctx, hostname, future.getNow());
                             } catch (Throwable cause) {
                                 ctx.fireExceptionCaught(new DecoderException(cause));
                             }
@@ -332,22 +333,70 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
         }
     }
 
+    // Move these somewhere more accessible
+    private interface Function<X, Y> {
+        Y apply(X in);
+    }
+
+    private static <X, Y extends Future<X>, Z> GenericFutureListener<Y> cascadeTo(final Promise<Z> outPromise,
+                                                                                  final Function<X, Z> mapper) {
+        return new GenericFutureListener<Y>() {
+            @Override
+            public void operationComplete(Y future) throws Exception {
+                if (future.isSuccess()) {
+                    final Z result;
+                    try {
+                        result = mapper.apply(future.getNow());
+                    } catch (Throwable cause) {
+                        outPromise.setFailure(cause);
+                        return;
+                    }
+                    outPromise.setSuccess(result);
+                } else if (future.isCancelled()) {
+                    outPromise.cancel(false);
+                } else {
+                    outPromise.setFailure(future.cause());
+                }
+            }
+        };
+    }
+    // Move these somewhere more accessible
+
     /**
-     * The default implementation will simply call {@link SslContextMapping#map(String, Promise)} but users can
+     * The default implementation will simply call {@link SslContextMapping#map(Object, Promise)} but users can
      * override this method to implement custom behavior.
      *
      * NOTE: This method returns a {@code Future<?>} to accommodate the old and new way of doing things without
      * braking the API. Users who override this method are advised to change their return value from
      * {@code Future<SslContext>} to {@code Future<Supplier<? extends SslContext>>}.
      *
-     * @see SslContextMapping#map(String, Promise)
-     * @deprecated the return value will at some point change from {@code Future<SslContext>} to
-     *   {@code Future<Supplier<? extends SslContext>>}.
+     * @see SslContextMapping#map(Object, Promise)
+     * @deprecated Use {@link #lookupSupplier(io.netty.channel.ChannelHandlerContext, String)}.
      */
     @Deprecated
-    protected Future<?> lookup(ChannelHandlerContext ctx, String hostname) throws Exception {
-        Promise<Supplier<? extends SslContext>> promise = ctx.executor().newPromise();
-        return mapping.map(hostname, promise);
+    protected Future<SslContext> lookup(ChannelHandlerContext ctx, String hostname) throws Exception {
+        final Promise<SslContext> promise = ctx.executor().newPromise();
+        mapping.map(hostname, ctx.executor().<Supplier<? extends SslContext>>newPromise())
+                .addListener(cascadeTo(promise, new SniHandler.Function<Supplier<? extends SslContext>, SslContext>() {
+            @Override
+            public SslContext apply(Supplier<? extends SslContext> in) {
+                return in.get();
+            }
+        }));
+        return promise;
+    }
+
+    protected Future<Supplier<? extends SslContext>> lookupSupplier(ChannelHandlerContext ctx,
+                                                                    String hostname) throws Exception {
+        final Promise<Supplier<? extends SslContext>> promise = ctx.executor().newPromise();
+        lookup(ctx, hostname)
+                .addListener(cascadeTo(promise, new SniHandler.Function<SslContext, Supplier<? extends SslContext>>() {
+            @Override
+            public Supplier<? extends SslContext> apply(SslContext in) {
+                return new FixedSslContextSupplier(in);
+            }
+        }));
+        return promise;
     }
 
     /**
@@ -356,17 +405,8 @@ public class SniHandler extends ByteToMessageDecoder implements ChannelOutboundH
      * @see #select(ChannelHandlerContext, String)
      */
     @SuppressWarnings("unchecked")
-    private void onSslContextOrSupplier(ChannelHandlerContext ctx, String hostname, Object value) {
-        // NOTE: See JavaDoc of #lookup(...) to see why this if-else is necessary.
-        final Supplier<? extends SslContext> supplier;
-        if (value instanceof Supplier<?>) {
-            supplier = (Supplier<? extends SslContext>) value;
-        } else {
-            // If the user returned something other than a SslContext or Supplier<SslContext>
-            // in the Future<?> then let it blow up with a ClassCastException
-            supplier = new FixedSslContextSupplier((SslContext) value);
-        }
-
+    private void onSslContextSupplier(ChannelHandlerContext ctx, String hostname,
+                                      Supplier<? extends SslContext> supplier) {
         selection = new Selection(supplier, hostname);
         try {
             replaceHandler(ctx, hostname, supplier);

```

rkapsi: I think #6623 is a better approach as it retains compatibility, doesn't penalize the existing API with more object allocations and opens the door to use any `Future<T>`.

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
normanmaurer: @rkapsi yes as @Scottmitch noted this is a breaking change so sorry but no we can not do this.

rkapsi:
@normanmaurer @Scottmitch let's decline this and do #6623 instead.

Scottmitch:
closing in favor of  #6623

PR #6605, opened 2017-04-05
Supply a builder for Http2Codec

mosesn:
Motivation:

DefaultHttp2FrameWriter has constructors that it would be a hassle to
expose as configuration parameters on Http2Codec. We should instead
make a builder for Http2Codec.

Modifications:

Get rid of the public constructors on Http2Codec and instead make sure
you can always use the builder where you would have used the constructor
before.

Result:

Http2Codec can be configured more flexibly, and the SensitivityDetector
can be configured.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecBuilder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecBuilder.java
@@ -55,7 +55,7 @@ public final class Http2CodecBuilder {
     public Http2CodecBuilder(boolean server, Http2StreamChannelBootstrap bootstrap) {
         this.bootstrap = checkNotNull(bootstrap, "bootstrap");
         this.server = server;
-        this.initialSettings = new Http2Settings();
+        this.initialSettings = Http2Settings.defaultSettings();
         this.frameLogger = HTTP2_FRAME_LOGGER;
         this.headersSensitivityDetector = null;
     }
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecBuilder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecBuilder.java
@@ -1,0 +1,130 @@
+/*
+ * Copyright 2016 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.codec.http2;
+
+import io.netty.channel.ChannelHandler;
+import io.netty.handler.codec.http2.Http2HeadersEncoder.SensitivityDetector;
+import io.netty.util.internal.UnstableApi;
+
+import static io.netty.handler.logging.LogLevel.INFO;
+import static io.netty.util.internal.ObjectUtil.checkNotNull;
+
+/**
+ * A builder for {@link Http2Codec}.
+ */
+@UnstableApi
+public final class Http2CodecBuilder {
+    private static final Http2FrameLogger HTTP2_FRAME_LOGGER = new Http2FrameLogger(INFO, Http2Codec.class);
+
+    private Http2StreamChannelBootstrap bootstrap;
+    private boolean server;
+    private Http2Settings initialSettings;
+    private Http2FrameLogger frameLogger;
+    private SensitivityDetector headersSensitivityDetector;
+
+    /**
+     * Creates a new {@link Http2Codec} builder.
+     *
+     * @param server {@code true} this is a server
+     * @param streamHandler the handler added to channels for remotely-created streams. It must be
+     *     {@link ChannelHandler.Sharable}. {@code null} if the event loop from the parent channel should be used.
+     */
+    public Http2CodecBuilder(boolean server, ChannelHandler streamHandler) {
+        this(server, new Http2StreamChannelBootstrap().handler(streamHandler));
+    }
+
+    /**
+     * Creates a new {@link Http2Codec} builder.
+     *
+     * @param server {@code true} this is a server
+     * @param bootstrap bootstrap used to instantiate child channels for remotely-created streams.
+     */
+    public Http2CodecBuilder(boolean server, Http2StreamChannelBootstrap bootstrap) {
+        this.bootstrap = checkNotNull(bootstrap, "bootstrap");
+        this.server = server;
+        this.initialSettings = new Http2Settings();
+        this.frameLogger = HTTP2_FRAME_LOGGER;
+        this.headersSensitivityDetector = null;
+    }
+
+    /**
+     * Specifies the initial settings to send to peer.
+     *
+     * @param initialSettings non default initial settings to send to peer
+     * @return {@link Http2CodecBuilder} the builder for the {@link Http2Codec}
+     */
+    public Http2CodecBuilder initialSettings(Http2Settings initialSettings) {
+        this.initialSettings = initialSettings;
+        return this;
+    }
+
+    /**
+     * Returns the initial settings to send to peer.
+     */
+    public Http2Settings initialSettings() {
+        return initialSettings;
+    }
+
+    /**
+     * Specifies the frame logger to log messages with.
+     *
+     * @param frameLogger handler used to log all frames
+     * @return {@link Http2CodecBuilder} the builder for the {@link Http2Codec}
+     */
+    public Http2CodecBuilder frameLogger(Http2FrameLogger frameLogger) {
+        this.frameLogger = frameLogger;
+        return this;
+    }
+
+    /**
+     * Returns the frame logger to log messages with.
+     */
+    public Http2FrameLogger frameLogger() {
+        return frameLogger;
+    }
+
+    /**
+     * Specifies the headers sensitivity detector.
+     *
+     * @param headersSensitivityDetector decides whether headers should be considered sensitive or not
+     * @return {@link Http2CodecBuilder} the builder for the {@link Http2Codec}
+     */
+    public Http2CodecBuilder headersSensitivityDetector(SensitivityDetector headersSensitivityDetector) {
+        this.headersSensitivityDetector = headersSensitivityDetector;
+        return this;
+    }
+
+    /**
+     * Returns the headers sensitivity detector.
+     */
+    public SensitivityDetector headersSensitivityDetector() {
+        return headersSensitivityDetector;
+    }
+
+    private Http2FrameWriter frameWriter() {
+        return headersSensitivityDetector() == null ?
+            new DefaultHttp2FrameWriter() :
+            new DefaultHttp2FrameWriter(headersSensitivityDetector());
+    }
+
+    /**
+     * Builds/creates a new {@link Http2Codec} instance using this builder's current settings.
+     */
+    public Http2Codec build() {
+        return new Http2Codec(server, bootstrap,
+            frameWriter(), frameLogger(), initialSettings());
+    }
+}
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecBuilder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 The Netty Project
+ * Copyright 2017 The Netty Project
  *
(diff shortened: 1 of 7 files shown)


mosesn:
alternate approach to solving the same problem as #6599.  I don't have a strong preference, just wanted to provide both implementations so we can merge it whenever we come to a conclusion.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecBuilder.java
```diff
@@ -0,0 +1,131 @@
...
+import static io.netty.handler.logging.LogLevel.INFO;
+
+/**
+ * A builder for {@link Http2Codec}.
+ */
+@UnstableApi
+public final class Http2CodecBuilder {
+    private static final Http2FrameLogger HTTP2_FRAME_LOGGER = new Http2FrameLogger(INFO, Http2Codec.class);
+
+    private Http2StreamChannelBootstrap bootstrap;
```
fenik17: final?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecBuilder.java
```diff
@@ -0,0 +1,131 @@
...
+    private boolean server;
```
fenik17: final?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecBuilder.java
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * Returns the frame logger to log messages with.
+     */
+    public Http2FrameLogger frameLogger() {
+        return frameLogger;
+    }
+
+    /**
+     * Specifies the initial settings to send to peer.
```
ejona86: This JavaDoc looks wrong.

ejona86:
This seems fair. It is a very narrow API with obvious objectives. Even though it mixes the "old" and "new" APIs, the interface used here could easily be duplicated in the future without issue.

I do question if we'll want to solve the problem a different way eventually, like putting bits in the Http2Headers or add a noop method to Http2Headers and have applications override the implementation. If we were serious about §7.1.3, then that seems necessary to support proxies that comply with:
> An intermediary MUST NOT re-encode a value that uses the never
   indexed literal representation with another representation that would
   index it.  If HPACK is used for re-encoding, the never indexed
   literal representation MUST be used.

mosesn:
@ejona86 @fenik17 I've changed the commit to reflect your suggestions, thanks!

mosesn:
@normanmaurer @ejona86 @Scottmitch @fenik17 anything else you want to see?  what do you think of this approach?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecBuilder.java
```diff
@@ -0,0 +1,132 @@
+/*
+ * Copyright 2016 The Netty Project
```
fenik17: 2017

fenik17:
I like the builders :)

mosesn:
thanks @fenik17, I fixed the date.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecBuilder.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /**
+     * Creates a new {@link Http2Codec} builder.
+     *
+     * @param server {@code true} this is a server
+     * @param bootstrap bootstrap used to instantiate child channels for remotely-created streams.
+     */
+    public Http2CodecBuilder(boolean server, Http2StreamChannelBootstrap bootstrap) {
+        this.bootstrap = bootstrap;
```
normanmaurer: `ObjectUtil.checkNotNull(....)`

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecBuilder.java
```diff
@@ -0,0 +1,132 @@
...
+            return new DefaultHttp2FrameWriter(headersSensitivityDetector());
+        }
+    }
+
+    /**
+     * Builds/creates a new {@link Http2Codec} instance using this builder's current settings.
+     */
+    public Http2Codec build() {
+        return new Http2Codec(this.server, this.bootstrap,
+            this.frameWriter(), this.frameLogger(), this.initialSettings());
```
normanmaurer: nit: remove this. everywhere as its not needed.

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2CodecTest.java
```diff
@@ -72,16 +72,18 @@ public void setUp() throws InterruptedException {
...
                     }
                 });
         serverChannel = sb.bind(serverAddress).sync().channel();
 
+        Http2Codec codec = new Http2CodecBuilder(false, new TestChannelInitializer()).build();
         Bootstrap cb = new Bootstrap()
                 .channel(LocalChannel.class)
                 .group(group)
-                .handler(new Http2Codec(false, new TestChannelInitializer()));
+                .handler(codec);
```
normanmaurer: just build directly here ?

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2CodecTest.java
```diff
@@ -72,16 +72,18 @@ public void setUp() throws InterruptedException {
                     @Override
                     protected void initChannel(Channel ch) throws Exception {
                         serverConnectedChannel = ch;
-                        ch.pipeline().addLast(new Http2Codec(true, serverLastInboundHandler));
+                        Http2Codec codec = new Http2CodecBuilder(true, serverLastInboundHandler).build();
+                        ch.pipeline().addLast(codec);
```
normanmaurer: merge these two lines ?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecBuilder.java
```diff
@@ -0,0 +1,132 @@
...
+    public SensitivityDetector headersSensitivityDetector() {
+        return headersSensitivityDetector;
+    }
+
+    private Http2FrameWriter frameWriter() {
+        if (headersSensitivityDetector() == null) {
+            return new DefaultHttp2FrameWriter();
+        } else {
+            return new DefaultHttp2FrameWriter(headersSensitivityDetector());
+        }
```
normanmaurer: nit:

```java
return headersSensitivityDetector == null ? new DefaultHttp2FrameWriter() : new DefaultHttp2FrameWriter(headersSensitivityDetector());
```

normanmaurer:
@mosesn please address and let me know once you are done :)

mosesn:
@normanmaurer fixed!

normanmaurer:
@mosesn can you please rebase and fix the compiler error ? https://garage.netty.io/teamcity/viewLog.html?buildId=19260&buildTypeId=pulls_netty&guest=1

mosesn:
@normanmaurer I can't figure out how to reproduce that locally.  `mvn compile -pl example` doesn't seem to work.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecBuilder.java
normanmaurer: @mosesn also please fix this

mosesn: whoops, on it.

mosesn:
@normanmaurer OK, I think I've fixed the issues in example . . . let's see what the CI says.

normanmaurer:
@trustin PTAL

Scottmitch:
@mosesn - can you remove the `cache` file?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecBuilder.java
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Creates a new {@link Http2Codec} builder.
+     *
+     * @param server {@code true} this is a server
+     * @param bootstrap bootstrap used to instantiate child channels for remotely-created streams.
+     */
+    public Http2CodecBuilder(boolean server, Http2StreamChannelBootstrap bootstrap) {
+        this.bootstrap = checkNotNull(bootstrap, "bootstrap");
+        this.server = server;
+        this.initialSettings = new Http2Settings();
```
Scottmitch: can you make a common default settings object, and share with [AbstractHttp2ConnectionHandlerBuilder.java](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/AbstractHttp2ConnectionHandlerBuilder.java#L82)? Consider putting it in `Http2CodecUtil`.

mosesn:
OK, this is very weird, I could have sworn I had tackled @fenik17's comments, but it looks like I didn't . . . I think squashing along the way is probably the issue, since I had to `git push -f`, every time, so I'm going to do incremental commits from now on and squash at the end.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecBuilder.java
mosesn: @Scottmitch this seems risky since the settings object is mutable.

normanmaurer:
@Scottmitch please pull in when satisfied.

normanmaurer:
@Scottmitch ping ....

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecBuilder.java
Scottmitch: @mosesn - Good point. We have a few options, but I think we should have a consistent default settings between the two APIs.

1. Expose a read-only variant (may require refactoring into an interface etc...)
2. Expose a copy (may require a copy constructor).

mosesn: @Scottmitch I think the current implementation implicitly does API 2.  is there more you want me to do there?

mosesn:
@Scottmitch @normanmaurer sorry to nag, but I would really like to get this in before the next release.  Is there anything else you want me to do?  @Scottmitch I think I'm already doing what you suggested.  If I'm not, could you please elaborate?

normanmaurer:
@mosesn the release happened on the weekend :(

@Scottmitch PTAL again ...

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecBuilder.java
Scottmitch: The goal is to ensure consistency in the initial settings between the "new API" and the "old API". The PR as it stands introduces a different initial settings object for the "new API" than we are using for the "old API". In addition the settings introduced in this PR doesn't limit the `MAX_HEADER_LIST_SIZE` and may leave implementations exposed to unnecessarily large headers by default. The suggestion was to make the settings object built in [AbstractHttp2ConnectionHandlerBuilder.java](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/AbstractHttp2ConnectionHandlerBuilder.java#L82) available in a common place, and use it here. I added an option 3 which should not be much additional effort.

mosesn: oh, my mistake, I get it now.  yeah, I'll create a static method which returns a new Http2Settings object.  I'll put it on Http2Settings.

mosesn: @Scottmitch for what it's worth, this reduces consistency.  [Http2Codec][0] previously didn't use that setting.  I agree this is the right default though.  Should we still go forward with this?

[0]: https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Codec.java#L60-L68

normanmaurer:
@mosesn thanks... will merge once the CI completes.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecBuilder.java
Scottmitch: The "old" and the "new" API should be consistent. I think it is a bug/oversight that the Http2Codec is not using an setting objects equivalent to the one in `AbstractHttp2ConnectionHandlerBuilder`. Since we are changing the code in this PR I'm fine with making the change in this PR ... but it can also be a followup PR if you prefer.

mosesn: @Scottmitch rad, the current version has the static method change.

normanmaurer:
Squashed and cherry-picked into 4.1 (cf26227c6ca6cf405d2b4140e47548a31da1b305).

@mosesn thanks!

Scottmitch:
delayed lgtm .. thanks @mosesn

mosesn:
thanks @Scottmitch @normanmaurer !

Issue #6607, opened 2017-04-05
4.1.9 breaks epoll selector rebuild


nhw76:
The following commit appears to break the epoll selector rebuild logic in rebuildSelector():

https://github.com/netty/netty/commit/795f318c3c11ec0520e7acd963ad4b310c287c20

If the logic is fired by an excessive number of premature returns, you will hit an exception like this:

```
java.lang.ClassCastException: io.netty.channel.nio.SelectedSelectionKeySetSelector cannot be cast to java.nio.channels.spi.AbstractSelector
        at java.nio.channels.spi.AbstractSelectableChannel.register(AbstractSelectableChannel.java:212)
        at io.netty.channel.nio.NioEventLoop.rebuildSelector0(NioEventLoop.java:347)
        at io.netty.channel.nio.NioEventLoop.rebuildSelector(NioEventLoop.java:318)
        at io.netty.channel.nio.NioEventLoop.select(NioEventLoop.java:769)
        at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:388)
        at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:858)
        at java.lang.Thread.run(Thread.java:745)
```

Looks like a case of replacing the selector member with the unwrappedSelector() method may have been missed?

Scottmitch:
see #6608

Scottmitch:
please re-open if PR #6608 didn't fix the issue

md-5:
https://hub.spigotmc.org/jira/browse/SPIGOT-3418
User is suggesting this issue is there in 4.1.13

normanmaurer:
Let me check

normanmaurer:
@md-5 it is fixed in 4.1.13 and to prove this I added a unit-test. Please check:
https://github.com/netty/netty/pull/6971

md-5:
Hmm thanks for looking, this is strange.
It is an isolated report so far, I can only think maybe the user ended up with a different Netty version than the one we bundled.

md-5:
Apologies @normanmaurer , looks like a packaging error on our behalf caused the incorrect version to be used.

Thanks for looking anyway

normanmaurer:
No worries... more unit tests FTW

PR #6608, opened 2017-04-05
NioEventLoop#rebuildSelector0 throws ClassCastException

Scottmitch:
Motivation:
Commit 795f318 simplified some code related to the special case Set for the selected keys and introduced a Selector wrapper to make sure this set was properly reset. However the JDK makes assumptions about the type of Selector and this type is not extensible. This means whenever we call into the JDK we must provide the unwrapped version of the Selector or we get a ClassCastException. We missed a case of unwrapping in NioEventLoop#rebuildSelector0.

Modificaitons:
- NioEventLoop#openSelector should return a tuple so we can atomically set the wrapped and unwrapped Selector
- NioEventLoop#rebuildSelector0 should use the unwrapped version of the selector

Result:
Fixes https://github.com/netty/netty/issues/6607.

### DIFF
--- a/transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
+++ b/transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
@@ -146,11 +146,29 @@ public final class NioEventLoop extends SingleThreadEventLoop {
             throw new NullPointerException("selectStrategy");
         }
         provider = selectorProvider;
-        selector = openSelector();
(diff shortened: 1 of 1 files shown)


Scottmitch:
@nhw76 - Can you verify this fixes the issue?

Scottmitch:
4.1 (c37267d) 4.0 (be94aa0)

PR #6613, opened 2017-04-06
NetUtil IPv6 bugs related to IPv4 and compression

Scottmitch:
Motivation:
NetUtil#getByName and NetUtil#isValidIpV6Address do not strictly enforce the format of IPv4 addresses that are allowed to be embedded in IPv6 addresses as specified in https://tools.ietf.org/html/rfc4291#section-2.5.5. This may lead to invalid addresses being parsed, or invalid addresses being considered valid.

Modifications:
- NetUtil#isValidIpV6Address should enforce the IPv4-Compatible and IPv4-Mapped are the only valid formats for including IPv4 addresses as specified in https://tools.ietf.org/html/rfc4291#section-2.5.5
- NetUtil#getByName should more stritcly parse IPv6 addresses which contain IPv4 addresses as specified in https://tools.ietf.org/html/rfc4291#section-2.5.5

Result:
NetUtil#getByName and NetUtil#isValidIpV6Address respect the IPv6 RFC which defines the valid formats for embedding IPv4 addresses.

### DIFF
--- a/common/src/main/java/io/netty/util/NetUtil.java
+++ b/common/src/main/java/io/netty/util/NetUtil.java
@@ -314,8 +314,7 @@ public final class NetUtil {
     }
 
     /**
-     * Creates an byte[] based on an ipAddressString. No error handling is
-     * performed here.
+     * Creates an byte[] based on an ipAddressString. No error handling is performed here.
      */
     public static byte[] createByteArrayFromIpAddressString(String ipAddressString) {
 
@@ -343,110 +342,9 @@ public final class NetUtil {
                 ipAddressString = ipAddressString.substring(0, percentPos);
             }
 
-            StringTokenizer tokenizer = new StringTokenizer(ipAddressString, ":.", true);
-            ArrayList<String> hexStrings = new ArrayList<String>();
-            ArrayList<String> decStrings = new ArrayList<String>();
-            String token = "";
-            String prevToken = "";
-            int doubleColonIndex = -1; // If a double colon exists, we need to
-            // insert 0s.
-
-            // Go through the tokens, including the seperators ':' and '.'
-            // When we hit a : or . the previous token will be added to either
-            // the hex list or decimal list. In the case where we hit a ::
-            // we will save the index of the hexStrings so we can add zeros
-            // in to fill out the string
-            while (tokenizer.hasMoreTokens()) {
-                prevToken = token;
-                token = tokenizer.nextToken();
-
-                if (":".equals(token)) {
-                    if (":".equals(prevToken)) {
-                        doubleColonIndex = hexStrings.size();
-                    } else if (!prevToken.isEmpty()) {
-                        hexStrings.add(prevToken);
-                    }
-                } else if (".".equals(token)) {
-                    decStrings.add(prevToken);
-                }
-            }
-
-            if (":".equals(prevToken)) {
-                if (":".equals(token)) {
-                    doubleColonIndex = hexStrings.size();
-                } else {
-                    hexStrings.add(token);
-                }
-            } else if (".".equals(prevToken)) {
-                decStrings.add(token);
-            }
-
-            // figure out how many hexStrings we should have
-            // also check if it is a IPv4 address
-            int hexStringsLength = 8;
-
-            // If we have an IPv4 address tagged on at the end, subtract
-            // 4 bytes, or 2 hex words from the total
-            if (!decStrings.isEmpty()) {
-                hexStringsLength -= 2;
-            }
-
-            // if we hit a double Colon add the appropriate hex strings
-            if (doubleColonIndex != -1) {
-                int numberToInsert = hexStringsLength - hexStrings.size();
-                for (int i = 0; i < numberToInsert; i ++) {
-                    hexStrings.add(doubleColonIndex, "0");
-                }
-            }
-
-            byte[] ipByteArray = new byte[IPV6_BYTE_COUNT];
-
-            // Finally convert these strings to bytes...
-            for (int i = 0; i < hexStrings.size(); i ++) {
-                convertToBytes(hexStrings.get(i), ipByteArray, i << 1);
-            }
-
-            // Now if there are any decimal values, we know where they go...
-            for (int i = 0; i < decStrings.size(); i ++) {
-                ipByteArray[i + 12] = (byte) (Integer.parseInt(decStrings.get(i)) & 255);
-            }
-            return ipByteArray;
+            return getIPv6ByName(ipAddressString, true);
         }
         return null;
-    }
-
-    /**
-     * Converts a 4 character hex word into a 2 byte word equivalent
-     */
-    private static void convertToBytes(String hexWord, byte[] ipByteArray, int byteIndex) {
-
-        int hexWordLength = hexWord.length();
-        int hexWordIndex = 0;
-        ipByteArray[byteIndex] = 0;
-        ipByteArray[byteIndex + 1] = 0;
-        int charValue;
-
-        // high order 4 bits of first byte
-        if (hexWordLength > 3) {
-            charValue = getIntValue(hexWord.charAt(hexWordIndex ++));
-            ipByteArray[byteIndex] |= charValue << 4;
-        }
-
-        // low order 4 bits of the first byte
-        if (hexWordLength > 2) {
-            charValue = getIntValue(hexWord.charAt(hexWordIndex ++));
-            ipByteArray[byteIndex] |= charValue;
-        }
-
-        // high order 4 bits of second byte
-        if (hexWordLength > 1) {
-            charValue = getIntValue(hexWord.charAt(hexWordIndex ++));
-            ipByteArray[byteIndex + 1] |= charValue << 4;
-        }
-
-        // low order 4 bits of the first byte
-        charValue = getIntValue(hexWord.charAt(hexWordIndex));
-        ipByteArray[byteIndex + 1] |= charValue & 15;
     }
 
     private static int getIntValue(char c) {
@@ -581,17 +479,40 @@ public final class NetUtil {
                     numberOfPeriods ++;
                     if (numberOfPeriods > 3) {
                         return false;
+                    } else if (numberOfPeriods == 1) {
+                        // Verify this address is of the correct structure to contain an IPv4 address.
+                        // It must be IPv4-Mapped or IPv4-Compatible
+                        // (see https://tools.ietf.org/html/rfc4291#section-2.5.5).
+                        int j = i - word.length() - 2; // index of character before the previous ':'.
+                        final int beginColonIndex = ipAddress.lastIndexOf(':', j);
+                        if (beginColonIndex == -1) {
+                            return false;
+                        }
+                        char tmpChar = ipAddress.charAt(j);
+                        if (isValidIPv4MappedChar(tmpChar)) {
+                            if (j - beginColonIndex != 4 ||
+                                !isValidIPv4MappedChar(ipAddress.charAt(j - 1)) ||
+                                !isValidIPv4MappedChar(ipAddress.charAt(j - 2)) ||
+                                !isValidIPv4MappedChar(ipAddress.charAt(j - 3))) {
+                                return false;
+                            }
+                            j -= 5;
+                        } else if (tmpChar == '0' || tmpChar == ':') {
+                            --j;
+                        } else {
+                            return false;
+                        }
+                        for (; j >= startOffset; --j) {
+                            tmpChar = ipAddress.charAt(j);
+                            if (tmpChar != '0' && tmpChar != ':') {
+                                return false;
+                            }
+                        }
                     }
                     if (!isValidIp4Word(word.toString())) {
                         return false;
                     }
                     if (numberOfColons != 6 && !doubleColon) {
-                        return false;
-                    }
-                    // a special case ::1:2:3:4:5:d.d.d.d allows 7 colons with an
-                    // IPv4 ending, otherwise 7 :'s is bad
-                    if (numberOfColons == 7 && ipAddress.charAt(startOffset) != ':' &&
-                        ipAddress.charAt(1 + startOffset) != ':') {
                         return false;
                     }
                     word.delete(0, word.length());
@@ -635,7 +556,7 @@ public final class NetUtil {
         // Check if we have an IPv4 ending
         if (numberOfPeriods > 0) {
             // There is a test case with 7 colons and valid ipv4 this should resolve it
-            if (numberOfPeriods != 3 || !(isValidIp4Word(word.toString()) && numberOfColons < 7)) {
+            if (numberOfPeriods != 3 || !(isValidIp4Word(word.toString()) && (numberOfColons < 7 || doubleColon))) {
                 return false;
             }
         } else {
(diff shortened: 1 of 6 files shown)


Scottmitch:
@trustin @fenik17 - PTAL. Please suggest additional unit tests if you think of interesting test cases.

common/src/main/java/io/netty/util/NetUtil.java
```diff
@@ -803,21 +828,25 @@ public static Inet6Address getByName(CharSequence ip, boolean ipv4Mapped) {
                 break;
             case '.':
                 ++ipv4Seperators;
-                if (i - begin > IPV4_MAX_CHAR_BETWEEN_SEPARATOR
+                tmp = i - begin; // tmp is the length of the current segment.
+                if (tmp > IPV4_MAX_CHAR_BETWEEN_SEPARATOR
```
netkins: MAJOR Reduce the number of conditional operators (19) used in the expression (maximum allowed 5). [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1067)

common/src/main/java/io/netty/util/NetUtil.java
```diff
@@ -748,11 +747,37 @@ public static Inet6Address getByName(CharSequence ip) {
...
+     * defined in <a href="http://tools.ietf.org/html/rfc4291#section-2.5.5">rfc 4291 section 2</a> is supported.
+     * @param ip {@link CharSequence} IP address to be converted to a {@link Inet6Address}
+     * @param ipv4Mapped
+     * <ul>
+     * <li>{@code true} To allow IPv4 mapped inputs to be translated into {@link Inet6Address}</li>
+     * <li>{@code false} Consider IPv4 mapped addresses as invalid.</li>
+     * </ul>
+     * @return byte array representation of the {@code ip} or {@code null} if not a valid IP address.
+     */
+    private static byte[] getIPv6ByName(CharSequence ip, boolean ipv4Mapped) {
```
netkins: MAJOR The Cyclomatic Complexity of this method "getIPv6ByName" is 94 which is greater than 25 authorized. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AMethodCyclomaticComplexity)

fenik17:
Can we add support for IPv6 addresses with `[]` and `%` into the `getIPv6ByName()` method? Somethink like that:
```java
public static byte[] createByteArrayFromIpAddressString(String ipAddressString) {
    if (isValidIpV4Address(ipAddressString)) {
        ....
    }
    return getIPv6ByName(ipAddressString, true);
}

private static byte[] getIPv6ByName(CharSequence ip, boolean ipv4Mapped) {
    ...
    int startIdx; // instead of '0' below
    int endIdx;   // instead of 'ipLength - 1' below
    if (ip.charAt(0) == '[') {
        startIdx = 1;
        endIdx = ip.length() - 2;
    } else {
        startIdx = 0;
        endIdx = ip.length() - 1;
    }
    ...
    loop: for (int i = startIdx; i <= endIdx; i++) {
        ...
        case '%':
            endIdx = i - 1;
            break loop;
        }
        ...
    }
    ...
    if (startIdx == 1 && (ipv6Seperators == 0 || ip.charAt(ip.length() - 1) != ']')) {
        // "[x.x.x.x]" - invalid ipv4
        // "[x:x:x::x" - invalid ipv6
        return null;
    }
    ...
}
```

This will allow us not to call a 'isValidIpV6Address()' method in the `createByteArrayFromIpAddressString()`, and make `getByName` more universal. Also, we will be able to test `getByName` on a par with the `isValidIpV4Address()` and `isValidIpV6Address()`:

```java
    @Test
    public void testIsValidIpV4Address() {
        for (String host : validIpV4Hosts.keySet()) {
            assertTrue(host, NetUtil.isValidIpV4Address(host));
            assertNotNull(host, NetUtil.getByName(host));
        }
        for (String host : invalidIpV4Hosts.keySet()) {
            assertFalse(host, NetUtil.isValidIpV4Address(host));
            assertNull(host, NetUtil.getByName(host));
        }
    }

    @Test
    public void testIsValidIpV6Address() {
        for (String host : validIpV6Hosts.keySet()) {
            assertTrue(host, NetUtil.isValidIpV6Address(host));
            assertNotNull(host, NetUtil.getByName(host));
        }
        for (String host : invalidIpV6Hosts.keySet()) {
            assertFalse(host, NetUtil.isValidIpV6Address(host));
            assertNull(host, NetUtil.getByName(host));
        }
    }
```

fenik17:
In addition to the previous comment - `InetAddress.getByName()` from the JDK supports an `[]` and `%`.

common/src/main/java/io/netty/util/NetUtil.java
```diff
@@ -679,6 +634,50 @@ private static boolean isValidNumericChar(char c) {
...
+        // The IPv4 compatible is deprecated, but it allows parsing of plain IPv4 addressed into IPv6-Mapped addresses.
+        return b0 == b1 && (b0 == 0 || b0 == -1);
+    }
+
+    private static boolean isValidIPv4Mapped(byte[] bytes, int currentIndex) {
+        if (currentIndex > 12 || currentIndex < 2 ||
+            !isValidIPv4MappedSeparators(bytes[currentIndex - 1], bytes[currentIndex - 2])) {
+            return false;
+        }
+        currentIndex -= 2;
```
fenik17: Maybe `-= 3`? With `-2` test failed for ip "::ffff:1.2.3.4" (whithout `Unsafe`).

fenik17: Netkins doesn't run tests with `-Dio.netty.noUnsafe=true`?

common/src/main/java/io/netty/util/NetUtil.java
```diff
@@ -803,21 +828,25 @@ public static Inet6Address getByName(CharSequence ip, boolean ipv4Mapped) {
...
-                                           i == 1 && !isValidNumericChar(ip.charAt(0))))) {
+                        || ipv4Seperators == 1 && (!ipv4Mapped ||
+                                             // We also parse pure IPv4 addresses as IPv4-Mapped for ease of use.
+                                                   currentIndex != 0 && !isValidIPv4Mapped(bytes, currentIndex))
+                        || (tmp == 3 && (!isValidNumericChar(ip.charAt(i - 1)) ||
+                                         !isValidNumericChar(ip.charAt(i - 2)) ||
+                                         !isValidNumericChar(ip.charAt(i - 3))) ||
+                            tmp == 2 && (!isValidNumericChar(ip.charAt(i - 1)) ||
+                                         !isValidNumericChar(ip.charAt(i - 2))) ||
+                            tmp == 1 && !isValidNumericChar(ip.charAt(i - 1)))) {
```
fenik17: all of last part need only for the `ipv4Seperators == 1`:
```
|| ( 
    ipv4Seperators == 1 &&
    (
          (!ipv4Mapped || currentIndex != 0 && !isValidIPv4Mapped(bytes, currentIndex))
       || (tmp == 3 && (!isValidNumericChar(ip.charAt(i - 1)) || !isValidNumericChar(ip.charAt(i - 2)) || !isValidNumericChar(ip.charAt(i - 3))) ||
           tmp == 2 && (!isValidNumericChar(ip.charAt(i - 1)) || !isValidNumericChar(ip.charAt(i - 2))) ||
           tmp == 1 && !isValidNumericChar(ip.charAt(i - 1)))
    )
  ) {
```

fenik17:
@Scottmitch fix the typos `ipv6Seperators` -> `ipv6Separators`, `ipv4Seperators` -> `ipv4Separators` please :)

common/src/main/java/io/netty/util/NetUtil.java
Scottmitch: good catch ... should be `-=3` ... the indexing is off the for unsafe case.

> Netkins doesn't run tests with -Dio.netty.noUnsafe=true?

No. It just assumes the default for the platform.

common/src/main/java/io/netty/util/NetUtil.java
Scottmitch: Good catch. The validation after we see the first `ipv4Separators` is done in the default case below.

Scottmitch:
> Can we add support for IPv6 addresses with [] and % into a getIPv6ByName() method?

Lets do a followup PR for this enhancement.

common/src/main/java/io/netty/util/NetUtil.java
```diff
@@ -802,22 +827,26 @@ public static Inet6Address getByName(CharSequence ip, boolean ipv4Mapped) {
                 begin = -1;
                 break;
             case '.':
-                ++ipv4Seperators;
-                if (i - begin > IPV4_MAX_CHAR_BETWEEN_SEPARATOR
-                        || ipv4Seperators > IPV4_SEPARATORS
-                        || (ipv6Seperators > 0 && (currentIndex + compressLength < 12))
+                ++ipv4Separators;
+                tmp = i - begin; // tmp is the length of the current segment.
+                if (tmp > IPV4_MAX_CHAR_BETWEEN_SEPARATOR
```
netkins: MAJOR Reduce the number of conditional operators (19) used in the expression (maximum allowed 5). [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1067)

common/src/main/java/io/netty/util/NetUtil.java
```diff
@@ -850,18 +879,18 @@ public static Inet6Address getByName(CharSequence ip, boolean ipv4Mapped) {
...
-                    ipv4Seperators != IPV4_SEPARATORS ||
+                    ipv4Separators != IPV4_SEPARATORS ||
                     currentIndex >= bytes.length) {
                 return null;
             }
-            if (ipv6Seperators == 0) {
+            if (ipv6Separators == 0) {
                 compressLength = 12;
-            } else if (ipv6Seperators >= IPV6_MIN_SEPARATORS &&
+            } else if (ipv6Separators >= IPV6_MIN_SEPARATORS &&
```
netkins: MAJOR Reduce the number of conditional operators (8) used in the expression (maximum allowed 5). [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1067)

fenik17:
`convertToBytes()` can be removed at now.

> Lets do a followup PR for this enhancement.

Ok.

Scottmitch:
> convertToBytes() can be removed at now.

+1

fenik17:
> Please suggest additional unit tests if you think of interesting test cases.

**0:0:0:0:0:ffff::10.0.0.1**

```
isValidIpV6Address("0:0:0:0:0:ffff::10.0.0.1") => false
getByName("0:0:0:0:0:ffff::10.0.0.1")          => not null
```

normanmaurer:
Let me merge this...

fenik17:
The `getByName()` method still does not check IPv4-Mapped addresses, e.g. for "1:0:0:0:0:ffff:10.0.0.1" or "0:0:0:0:0:ffff::10.0.0.1" it return a not null.
But `isValidIpV6Address()` has such check.

fenik17:
I can fix it in another PR if @Scottmitch have not time..

Scottmitch:
@fenik17 - Thanks for pointing out this unit test .. let me add this unit test and fix the issue

common/src/main/java/io/netty/util/NetUtil.java
```diff
@@ -850,18 +845,18 @@ public static Inet6Address getByName(CharSequence ip, boolean ipv4Mapped) {
...
                            ip.charAt(ipLength - 1) != ':' &&
```
fenik17: this check - `ip.charAt(ipLength - 1) != ':'` - seems superfluous or wrong for the `ipv4Separators > 0`.

Scottmitch: agreed ... this should be caught in the loop above.

fenik17:
One more unit test for invalid address: "1:0:0:0:0:ffff:10.0.0.1".

fenik17:
Maybe better use `equals()` with zeroed array instead of `getLong() / getInt()` in the `isValidIPv4Mapped()`.

common/src/main/java/io/netty/util/NetUtil.java
```diff
@@ -850,18 +847,17 @@ public static Inet6Address getByName(CharSequence ip, boolean ipv4Mapped) {
...
                 return null;
             }
-            if (ipv6Seperators == 0) {
+            if (ipv6Separators == 0) {
                 compressLength = 12;
-            } else if (ipv6Seperators >= IPV6_MIN_SEPARATORS &&
-                           ip.charAt(ipLength - 1) != ':' &&
-                           (!isCompressed && (ipv6Seperators == 6 && ip.charAt(0) != ':') ||
-                            isCompressed && (ipv6Seperators + 1 < IPV6_MAX_SEPARATORS &&
+            } else if (ipv6Separators >= IPV6_MIN_SEPARATORS &&
```
netkins: MAJOR Reduce the number of conditional operators (7) used in the expression (maximum allowed 5). [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1067)

Scottmitch:
> Maybe better use equals() with zeroed array instead of getLong() / getInt() in the isValidIPv4Mapped()?

I considered doing this initially ... I will just add a `isZero` method to PlatformDependent instead to make this more generally available an independently unit tested. This way we won't need an intermediate allocation.

Scottmitch:
4.1 (9cb858fcf) 4.0 (241cdc677)

common/src/test/java/io/netty/util/NetUtilTest.java
```diff
@@ -87,7 +92,11 @@
             "[2001:0000:4136:e378:8000:63bf:3fff:fdd2]", "200100004136e378800063bf3ffffdd2",
             "0:1:2:3:4:5:6:789a",                        "0000000100020003000400050006789a",
             "0:1:2:3::f",                                "0000000100020003000000000000000f",
-            "0:0:0:0:0:0:10.0.0.1",                      "0000000000000000000000000a000001",
+            "0:0:0:0:0:0:10.0.0.1",                      "00000000000000000000ffff0a000001",
+            "0:0:0:0:0::10.0.0.1",                       "00000000000000000000ffff0a000001",
+            "::0:0:0:0:0:10.0.0.1",                      "00000000000000000000ffff0a000001",
+            "0:0:0:0:0:0::10.0.0.1",                     "00000000000000000000ffff0a000001",
```
fenik17: @Scottmitch oops. Is it really valid?

Scottmitch: Yes we allow `IPv4 compatible` addresses as input. The rational is to be liberal in what we accept but more strict in what we generate (we only generate IPv4 mapped).

```
// We allow IPv4 Mapped (https://tools.ietf.org/html/rfc4291#section-2.5.5.1)
// and IPv4 compatible (https://tools.ietf.org/html/rfc4291#section-2.5.5.1).
// The IPv4 compatible is deprecated, but it allows parsing of plain IPv4 addressed into IPv6-Mapped addresses.
```

fenik17: I'm about of count of zeros and "::" after. There can not be more bits in place of "::".
This is confirmed by examples [1] and some online-validators: [2] and [3].

[1] https://tools.ietf.org/html/rfc4291#section-2.2
[2] http://formvalidation.io/validators/ip/ ("Try it" in the end)
[3] https://www.helpsystems.com/intermapper/ipv6-test-address-validation

fenik17: So... I think the "0:0:0:0:0::10.0.0.1", "0:0:0:0::10.0.0.1" ...  "::10.0.0.1" also is invalid in the mapped mode, because it's compessed form of "0:0:0:0:0:0:10.0.0.1". But now it parsed as "::ffff:10.0.0.1".

Scottmitch: Thanks for pointing this out. I agree there is a bug. Let me come up with a PR.

Scottmitch: See https://github.com/netty/netty/pull/6680

PR #6614, opened 2017-04-07
HTTP/2 Allow more time for EventLoopGroup to shutdown in test that us…

Scottmitch:
…e LocalChannel

Motivation:
The CI servers have reported leaks while building the HTTP/2 unit tests. The unit tests attempt to wait for the channels to be closed before exiting the test, but we should wait in case there are any tasks pending on the EventLoopGroup's task queues.

Modifications:
- Change the Future.sync() operations to Future.syncUninterruptibly()
- HTTP/2 unit tests which use local channel should wait for 5 seconds before shutting down the EventLoopGroups

Result:
More likely that any cleanup related tasks will execute before the unit tests are shutdown.

### DIFF
--- a/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2CodecTest.java
+++ b/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2CodecTest.java
@@ -36,9 +36,9 @@ import org.junit.BeforeClass;
 import org.junit.Test;
 
 import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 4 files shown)


Scottmitch:
Related to https://github.com/netty/netty/issues/6577. I have investigated this in the past, but don't have time to dig in now. It seems that the writes are making it to the LocalChannel and therefore the LocalChannel is responsible for releasing the objects. This PR shouldn't make anything worse so I will merge it as a stop gap until I have more time to investigate.

Scottmitch:
4.1 (1bc5bc6)

PR #6615, opened 2017-04-07
add common DNS record types

flier:
Motivation:

add more common DNS record types

Modification:

define, decode and encode A, AAAA, CNAME, MX, NS, SOA, SRV and TXT record types

Result:

### DIFF
--- a/codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsAAAARecord.java
+++ b/codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsAAAARecord.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.codec.dns;
+
+import io.netty.util.internal.StringUtil;
+import io.netty.util.internal.UnstableApi;
+
+import java.net.Inet6Address;
+
+import static io.netty.util.internal.ObjectUtil.checkNotNull;
+
+@UnstableApi
+public class DefaultDnsAAAARecord extends AbstractDnsRecord implements DnsAAAARecord {
+    private final Inet6Address address;
+
+    /**
+     * Creates a new AAAA record.
+     *
+     * @param name the domain name
+     * @param dnsClass the class of the record, usually one of the following:
+     *                 <ul>
+     *                     <li>{@link #CLASS_IN}</li>
+     *                     <li>{@link #CLASS_CSNET}</li>
+     *                     <li>{@link #CLASS_CHAOS}</li>
+     *                     <li>{@link #CLASS_HESIOD}</li>
+     *                     <li>{@link #CLASS_NONE}</li>
+     *                     <li>{@link #CLASS_ANY}</li>
+     *                 </ul>
+     * @param timeToLive the TTL value of the record
+     * @param address the IPv6 Internet address this AAAA record resolves to.
+     */
+    public DefaultDnsAAAARecord(String name, int dnsClass, long timeToLive, Inet6Address address) {
+        super(name, DnsRecordType.AAAA, dnsClass, timeToLive);
+        this.address = checkNotNull(address, "address");
+    }
+
+    @Override
+    public Inet6Address address() { return address; }
+
+    @Override
+    public String toString() {
+        final StringBuilder buf = new StringBuilder(64).append(StringUtil.simpleClassName(this)).append('(');
+        final DnsRecordType type = type();
+        buf.append(name().isEmpty()? "<root>" : name())
+                .append(' ')
+                .append(timeToLive())
+                .append(' ');
+
+        DnsMessageUtil.appendRecordClass(buf, dnsClass())
+                .append(' ')
+                .append(type.name());
+
+        buf.append(' ')
+                .append(address.getHostAddress());
+
+        return buf.toString();
+    }
+}
--- a/codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsAAAARecord.java
+++ b/codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsAAAARecord.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 The Netty Project
+ * Copyright 2018 The Netty Project
  *
  * The Netty Project licenses this file to you under the Apache License,
  * version 2.0 (the "License"); you may not use this file except in compliance
@@ -22,6 +22,9 @@ import java.net.Inet6Address;
 
 import static io.netty.util.internal.ObjectUtil.checkNotNull;
 
+/**
+ * Default {@link DnsAAAARecord} implementation.
+ */
 @UnstableApi
 public class DefaultDnsAAAARecord extends AbstractDnsRecord implements DnsAAAARecord {
     private final Inet6Address address;
@@ -51,9 +54,27 @@ public class DefaultDnsAAAARecord extends AbstractDnsRecord implements DnsAAAARe
     public Inet6Address address() { return address; }
 
     @Override
+    public boolean equals(Object o) {
+        if (this == o) { return true; }
+        if (o == null || getClass() != o.getClass()) { return false; }
+        if (!super.equals(o)) { return false; }
+        DefaultDnsAAAARecord that = (DefaultDnsAAAARecord) o;
+        return address.equals(that.address);
+    }
+
+    @Override
+    public int hashCode() {
+        int hashCode = super.hashCode();
+
+        hashCode = hashCode * 31 + address.hashCode();
+
+        return hashCode;
+    }
+
+    @Override
     public String toString() {
         final StringBuilder buf = new StringBuilder(64).append(StringUtil.simpleClassName(this)).append('(');
-        final DnsRecordType type = type();
+
         buf.append(name().isEmpty()? "<root>" : name())
                 .append(' ')
                 .append(timeToLive())
@@ -61,7 +82,7 @@ public class DefaultDnsAAAARecord extends AbstractDnsRecord implements DnsAAAARe
 
         DnsMessageUtil.appendRecordClass(buf, dnsClass())
                 .append(' ')
-                .append(type.name());
+                .append(type().name());
 
         buf.append(' ')
                 .append(address.getHostAddress());
--- a/codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsMxRecord.java
+++ b/codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsMxRecord.java
@@ -28,7 +28,7 @@ import static io.netty.util.internal.ObjectUtil.checkNotNull;
 @UnstableApi
 public class DefaultDnsMxRecord extends AbstractDnsRecord implements DnsMxRecord {
     private final int preference;
-    private final String hostname;
+    private final String exchange;
 
     /**
      * Creates a new MX record.
@@ -44,19 +44,19 @@ public class DefaultDnsMxRecord extends AbstractDnsRecord implements DnsMxRecord
      *                     <li>{@link #CLASS_ANY}</li>
      *                 </ul>
      * @param timeToLive the TTL value of the record
-     * @param hostname the canonical or primary name this MX record resolves to.
+     * @param exchange the canonical or primary name this MX record resolves to.
      */
-    public DefaultDnsMxRecord(String name, int dnsClass, long timeToLive, int preference, String hostname) {
+    public DefaultDnsMxRecord(String name, int dnsClass, long timeToLive, int preference, String exchange) {
         super(name, DnsRecordType.MX, dnsClass, timeToLive);
         this.preference = preference;
-        this.hostname = IDN.toASCII(checkNotNull(hostname, "hostname"));
+        this.exchange = IDN.toASCII(checkNotNull(exchange, "exchange"));
     }
 
     @Override
     public int preference() { return preference; }
 
     @Override
-    public String hostname() { return hostname; }
+    public String exchange() { return exchange; }
 
     @Override
     public boolean equals(Object o) {
@@ -64,7 +64,7 @@ public class DefaultDnsMxRecord extends AbstractDnsRecord implements DnsMxRecord
         if (o == null || getClass() != o.getClass()) { return false; }
         if (!super.equals(o)) { return false; }
         DefaultDnsMxRecord that = (DefaultDnsMxRecord) o;
-        return preference == that.preference && hostname.equals(that.hostname);
+        return preference == that.preference && exchange.equals(that.exchange);
     }
 
     @Override
@@ -72,7 +72,7 @@ public class DefaultDnsMxRecord extends AbstractDnsRecord implements DnsMxRecord
         int hashCode = super.hashCode();
 
         hashCode = hashCode * 31 + preference;
-        hashCode = hashCode * 31 + hostname.hashCode();
+        hashCode = hashCode * 31 + exchange.hashCode();
 
         return hashCode;
     }
@@ -93,7 +93,7 @@ public class DefaultDnsMxRecord extends AbstractDnsRecord implements DnsMxRecord
         buf.append(' ')
                 .append(preference)
                 .append(' ')
-                .append(hostname);
+                .append(exchange);
 
         return buf.toString();
     }
--- a/codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsMxRecord.java
+++ b/codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsMxRecord.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.codec.dns;
+
+import io.netty.util.internal.StringUtil;
+import io.netty.util.internal.UnstableApi;
+
+import static io.netty.util.internal.ObjectUtil.checkNotNull;
+
+@UnstableApi
+public class DefaultDnsMxRecord extends AbstractDnsRecord implements DnsMxRecord {
+    private final int preference;
+    private final String hostname;
+
+    /**
+     * Creates a new MX record.
+     *
+     * @param name the domain name
+     * @param dnsClass the class of the record, usually one of the following:
+     *                 <ul>
+     *                     <li>{@link #CLASS_IN}</li>
+     *                     <li>{@link #CLASS_CSNET}</li>
+     *                     <li>{@link #CLASS_CHAOS}</li>
+     *                     <li>{@link #CLASS_HESIOD}</li>
+     *                     <li>{@link #CLASS_NONE}</li>
+     *                     <li>{@link #CLASS_ANY}</li>
+     *                 </ul>
+     * @param timeToLive the TTL value of the record
+     * @param hostname the canonical or primary name this MX record resolves to.
+     */
+    public DefaultDnsMxRecord(String name, int dnsClass, long timeToLive, int preference, String hostname) {
+        super(name, DnsRecordType.MX, dnsClass, timeToLive);
+        this.preference = preference;
+        this.hostname = checkNotNull(hostname, "hostname");
+    }
+
+    @Override
+    public int preference() { return preference; }
+
+    @Override
+    public String hostname() { return hostname; }
+
+    @Override
+    public String toString() {
+        final StringBuilder buf = new StringBuilder(64).append(StringUtil.simpleClassName(this)).append('(');
+        final DnsRecordType type = type();
+        buf.append(name().isEmpty()? "<root>" : name())
+                .append(' ')
+                .append(timeToLive())
+                .append(' ');
+
+        DnsMessageUtil.appendRecordClass(buf, dnsClass())
+                .append(' ')
+                .append(type.name());
+
+        buf.append(' ')
+                .append(preference)
+                .append(' ')
+                .append(hostname);
+
+        return buf.toString();
+    }
+}
--- a/codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsMxRecord.java
+++ b/codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsMxRecord.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 The Netty Project
+ * Copyright 2018 The Netty Project
  *
  * The Netty Project licenses this file to you under the Apache License,
  * version 2.0 (the "License"); you may not use this file except in compliance
@@ -18,8 +18,13 @@ package io.netty.handler.codec.dns;
 import io.netty.util.internal.StringUtil;
 import io.netty.util.internal.UnstableApi;
 
+import java.net.IDN;
+
 import static io.netty.util.internal.ObjectUtil.checkNotNull;
 
+/**
+ * Default {@link DnsMxRecord} implementation.
+ */
 @UnstableApi
 public class DefaultDnsMxRecord extends AbstractDnsRecord implements DnsMxRecord {
     private final int preference;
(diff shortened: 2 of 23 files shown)


flier:
seems CI environment broken

> [07:07:04][io.netty:netty-handler] Failed to execute goal on project netty-handler: Could not resolve dependencies for project io.netty:netty-handler:jar:4.1.10.Final-SNAPSHOT: Failure to find io.netty:netty-tcnative:jar:linux-x86_64:2.0.1.Final-SNAPSHOT in https://oss.sonatype.org/content/repositories/snapshots was cached in the local repository, resolution will not be reattempted until the update interval of sonatype-nexus-snapshots has elapsed or updates are forced

normanmaurer:
@flier please rebase as well so it will not fail anymore

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsSoaRecord.java
```diff
@@ -0,0 +1,121 @@
...
+     * @param timeToLive the TTL value of the record
+     * @param primaryNameServer the name server that was the original or primary source of data for this zone.
+     * @param responsibleAuthorityMailbox the mailbox of the person responsible for this zone.
+     * @param serialNumber the version number of the original copy of the zone.
+     * @param refreshInterval the time interval before the zone should be refreshed.
+     * @param retryInterval the time interval that should elapse before a failed refresh should be retried.
+     * @param expireLimit the time value that specifies the upper limit on the time interval
+     * @param minimumTTL the minimum TTL field that should be exported with any RR from this zone.
+     */
+    public DefaultDnsSoaRecord(
```
netkins: MAJOR Constructor has 10 parameters, which is greater than 7 authorized. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS00107)

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsSoaRecord.java
```diff
@@ -0,0 +1,121 @@
...
+        buf.append(name().isEmpty()? "<root>" : name())
+                .append(' ')
+                .append(timeToLive())
+                .append(' ');
+
+        DnsMessageUtil.appendRecordClass(buf, dnsClass())
+                .append(' ')
+                .append(type.name());
+
+        buf.append(' ')
```
netkins: MINOR StringBuffer (or StringBuilder).append is called 2 consecutive times with literal Strings. Use a single append with a single combined String. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AConsecutiveLiteralAppends)

flier:
@normanmaurer I have rebase it to `4.1` branch, anything else need me do?

normanmaurer:
@flier sorry have been super busy. Will check soon

normanmaurer:
@trustin and @Scottmitch can you check pleasE ?

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsRecordDecoder.java
```diff
@@ -90,6 +94,57 @@ protected DnsRecord decodeRecord(
             String name, DnsRecordType type, int dnsClass, long timeToLive,
             ByteBuf in, int offset, int length) throws Exception {
 
+        if (type == DnsRecordType.A) {
```
Scottmitch: any reason why a `switch` wasn't used here ... seems more natural.

flier: sure, I can change to use switch, I use `if` only because of old code use it.

flier: `DnsRecordType.A` is a `static final DnsRecordType`, seems can't be use as a case statement label 

```java
public class DnsRecordType implements Comparable<DnsRecordType> {

    /**
     * Address record RFC 1035 Returns a 32-bit IPv4 address, most commonly used
     * to map hostnames to an IP address of the host, but also used for DNSBLs,
     * storing subnet masks in RFC 1101, etc.
     */
    public static final DnsRecordType A = new DnsRecordType(0x0001, "A");
```

flier: Besides, we hard to convert `DnsRecordType` to a `Enum`, because it use a private constructor wrap value of the unknown type, but `Enum` need a constant value.

```java
    public static DnsRecordType valueOf(int intValue) {
        DnsRecordType result = BY_TYPE.get(intValue);
        if (result == null) {
            return new DnsRecordType(intValue);
        }
        return result;
    }

    private DnsRecordType(int intValue) {
        this(intValue, "UNKNOWN");
    }
```

flier:
>can we make the decoding/encoding to/from POJOs, byte[], and String objects optional and/or implemented at a higher level? The DNS resolver can function without this additional overhead and I'm assuming this will be unnecessary for most users of the DNS resolver.

@Scottmitch  Do you means to replace `ByteBuf` with `byte[]`, or move those encoder/decoder to POJO like `DefaultDnsARecord` which may implements a `DnsSerializer`/`DnsSerializer` interface?

Scottmitch:
Sorry I rushed my first pass and didn't notice the object wrappers were being used instead of the raw record parsing in some places.

It seems like we now have some duplication of logic between the type hierarchy, the `DnsRecordType` checking, and `DnsRawRecord` usage. Did you consider moving away from the `DnsRecordType` checks and usage of `DnsRawRecord` in favor of the `instanceof` checks in some areas (e.g. `DnsNameResolverContext#onResponse`)? Is it generally expected there are other record types we don't yet support in the type hierarchy? If the correct decoder is used this seems like an indication of a programming error or not yet supported record type.

For example:
```java
    private static Map<String, String> buildAliasMap(DnsResponse response) {
        final int answerCount = response.count(DnsSection.ANSWER);
        Map<String, String> cnames = null;
        for (int i = 0; i < answerCount; i ++) {
            final DnsRecord r = response.recordAt(DnsSection.ANSWER, i);
            if (!(r instanceof DnsCNameRecord)) {
                continue;
            }
            if (cnames == null) {
                cnames = new HashMap<String, String>(min(8, answerCount));
            }

            cnames.put(r.name().toLowerCase(Locale.US), ((DnsCNameRecord) r).hostname().toLowerCase(Locale.US));
        }

        return cnames != null? cnames : Collections.<String, String>emptyMap();
    }
```

Also now that we are deserializing into objects maybe we should consider doing the lower case conversion up front during the decode (and only 1 time) to ensure consistency in comparisons and map operations:

https://www.ietf.org/rfc/rfc1035.txt
> Note that while upper and lower case letters are allowed in domain
names, no significance is attached to the case.  That is, two names with
the same spelling but different case are to be treated as if identical.

@normanmaurer - WDYT?

flier:
hi @normanmaurer and @Scottmitch, 

anything else need be change?

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsAAAARecord.java
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright 2017 The Netty Project
```
trustin: 2018 for all new files

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsAAAARecord.java
```diff
@@ -0,0 +1,71 @@
...
+        this.address = checkNotNull(address, "address");
+    }
+
+    @Override
+    public Inet6Address address() { return address; }
+
+    @Override
+    public String toString() {
+        final StringBuilder buf = new StringBuilder(64).append(StringUtil.simpleClassName(this)).append('(');
+        final DnsRecordType type = type();
```
trustin: Can be inlined

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsAAAARecord.java
```diff
@@ -0,0 +1,71 @@
...
+
+import io.netty.util.internal.StringUtil;
+import io.netty.util.internal.UnstableApi;
+
+import java.net.Inet6Address;
+
+import static io.netty.util.internal.ObjectUtil.checkNotNull;
+
+@UnstableApi
+public class DefaultDnsAAAARecord extends AbstractDnsRecord implements DnsAAAARecord {
```
trustin: `AAAA` -> `Aaaa`?

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsMxRecord.java
```diff
@@ -0,0 +1,76 @@
...
+     *                     <li>{@link #CLASS_HESIOD}</li>
+     *                     <li>{@link #CLASS_NONE}</li>
+     *                     <li>{@link #CLASS_ANY}</li>
+     *                 </ul>
+     * @param timeToLive the TTL value of the record
+     * @param hostname the canonical or primary name this MX record resolves to.
+     */
+    public DefaultDnsMxRecord(String name, int dnsClass, long timeToLive, int preference, String hostname) {
+        super(name, DnsRecordType.MX, dnsClass, timeToLive);
+        this.preference = preference;
```
trustin: Range check?

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsMxRecord.java
```diff
@@ -0,0 +1,76 @@
...
+        this.hostname = checkNotNull(hostname, "hostname");
```
trustin: IDN conversion, just like we do in `super()`?

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsMxRecord.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Override
+    public int preference() { return preference; }
+
+    @Override
+    public String hostname() { return hostname; }
```
trustin: Do we have a better name for this? I guess a user may be confused with `name` and `hostname`.

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsAAAARecord.java
```diff
@@ -0,0 +1,71 @@
...
+        final StringBuilder buf = new StringBuilder(64).append(StringUtil.simpleClassName(this)).append('(');
+        final DnsRecordType type = type();
+        buf.append(name().isEmpty()? "<root>" : name())
+                .append(' ')
+                .append(timeToLive())
+                .append(' ');
+
+        DnsMessageUtil.appendRecordClass(buf, dnsClass())
+                .append(' ')
+                .append(type.name());
```
trustin: I see code duplication in record implementations for line 55-64. Could we dedup?

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsNsRecord.java
```diff
@@ -0,0 +1,73 @@
...
+     */
+    public DefaultDnsNsRecord(
+            String name, int dnsClass, long timeToLive, String hostname) {
+        super(name, DnsRecordType.NS, dnsClass, timeToLive);
+        this.hostname = checkNotNull(hostname, "hostname");
+    }
+
+    @Override
+    public String hostname() {
+        return hostname;
```
trustin: Confusing with `name()`

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsNsRecord.java
```diff
@@ -0,0 +1,73 @@
...
+     *                     <li>{@link #CLASS_NONE}</li>
+     *                     <li>{@link #CLASS_ANY}</li>
+     *                 </ul>
+     * @param timeToLive the TTL value of the record
+     * @param hostname the hostname which should be authoritative for the specified class and domain.
+     */
+    public DefaultDnsNsRecord(
+            String name, int dnsClass, long timeToLive, String hostname) {
+        super(name, DnsRecordType.NS, dnsClass, timeToLive);
+        this.hostname = checkNotNull(hostname, "hostname");
```
trustin: IDN conversion

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsRecordDecoder.java
```diff
@@ -90,6 +94,57 @@ protected DnsRecord decodeRecord(
...
+            byte[] buf = new byte[length];
```
trustin: Check if the length is correct?

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsRecordDecoder.java
```diff
@@ -90,6 +94,57 @@ protected DnsRecord decodeRecord(
...
+            in.getBytes(offset, buf);
+            return new DefaultDnsARecord(
+                    name, dnsClass, timeToLive, (Inet4Address) InetAddress.getByAddress(buf));
+        }
+        if (type == DnsRecordType.AAAA) {
+            byte[] buf = new byte[length];
```
trustin: Check if the length is correct?

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsRecordDecoder.java
```diff
@@ -90,6 +94,57 @@ protected DnsRecord decodeRecord(
...
+            in.getBytes(offset, buf);
+            return new DefaultDnsAAAARecord(
+                    name, dnsClass, timeToLive, (Inet6Address) InetAddress.getByAddress(buf));
+        }
+        if (type == DnsRecordType.CNAME) {
+            return new DefaultDnsCNameRecord(
+                    name, dnsClass, timeToLive, decodeName0(in.duplicate().setIndex(offset, offset + length)));
+        }
+        if (type == DnsRecordType.MX) {
+            ByteBuf buf = in.retainedDuplicate().setIndex(offset, offset + length);
```
trustin: Check if the content is at least longer than 2 bytes?

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsRecordDecoder.java
```diff
@@ -90,6 +94,57 @@ protected DnsRecord decodeRecord(
...
+            int preference = buf.readShort();
```
trustin: `readUnsignedShort()` unless preference can be negative?

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsRecordDecoder.java
```diff
@@ -90,6 +94,57 @@ protected DnsRecord decodeRecord(
...
+        }
+        if (type == DnsRecordType.SOA) {
+            ByteBuf buf = in.retainedDuplicate().setIndex(offset, offset + length);
+            String primaryNameServer = decodeName0(buf);
+            String responsibleAuthorityMailbox = decodeName0(buf);
+            int serialNumber = buf.readInt();
+            int refreshInterval = buf.readInt();
+            int retryInterval = buf.readInt();
+            int expireLimit = buf.readInt();
+            int minimumTTL = buf.readInt();
```
trustin: Make sure that the length is greater than its minimum possible value?

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsRecordDecoder.java
```diff
@@ -90,6 +94,57 @@ protected DnsRecord decodeRecord(
...
+            return new DefaultDnsSoaRecord(
+                    name, dnsClass, timeToLive,
+                    primaryNameServer, responsibleAuthorityMailbox,
+                    serialNumber, refreshInterval, retryInterval, expireLimit, minimumTTL);
+        }
+        if (type == DnsRecordType.SRV) {
+            ByteBuf buf = in.retainedDuplicate().setIndex(offset, offset + length);
+            int priority = buf.readShort();
+            int weight = buf.readShort();
+            int port = buf.readShort();
```
trustin: Ditto - length check and `readUnsignedShort()`

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsRecordDecoder.java
```diff
@@ -98,6 +153,14 @@ protected DnsRecord decodeRecord(
             return new DefaultDnsPtrRecord(
                     name, dnsClass, timeToLive, decodeName0(in.duplicate().setIndex(offset, offset + length)));
         }
+
+        if (type == DnsRecordType.TXT) {
+            String text = in.retainedDuplicate().setIndex(offset, offset + length).toString(CharsetUtil.US_ASCII);
```
trustin: - TXT record cannot be longer than 255 bytes.
- Please make sure to check the maximum/minimum lengths of all record types you added.

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsRecordDecoder.java
```diff
@@ -98,6 +153,14 @@ protected DnsRecord decodeRecord(
...
+
+            return new DefaultDnsTxtRecord(
+                    name, dnsClass, timeToLive, text);
```
trustin: Could be a single line instead of two.

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsRecordEncoder.java
```diff
@@ -49,8 +49,22 @@ public final void encodeQuestion(DnsQuestion question, ByteBuf out) throws Excep
     public void encodeRecord(DnsRecord record, ByteBuf out) throws Exception {
         if (record instanceof DnsQuestion) {
             encodeQuestion((DnsQuestion) record, out);
+        } else if (record instanceof DnsARecord) {
+            encodeARecord((DnsARecord) record, out);
+        } else if (record instanceof DnsAAAARecord) {
+            encodeAAAARecord((DnsAAAARecord) record, out);
```
trustin: `AAAA` -> `Aaaa`?

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsRecordEncoder.java
```diff
@@ -140,7 +199,7 @@ private void encodeRawRecord(DnsRawRecord record, ByteBuf out) throws Exception
         out.writeBytes(content, content.readerIndex(), contentLen);
     }
 
-    protected void encodeName(String name, ByteBuf buf) throws Exception {
+    public static void encodeName(String name, ByteBuf buf) throws Exception {
```
trustin: Not your fault but it'd be very nice if you could add Javadoc for this. 🙏

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsSoaRecord.java
```diff
@@ -0,0 +1,120 @@
...
+            String primaryNameServer, String responsibleAuthorityMailbox,
+            int serialNumber, int refreshInterval, int retryInterval, int expireLimit, int minimumTTL) {
+        super(name, DnsRecordType.SOA, dnsClass, timeToLive);
+        this.primaryNameServer = checkNotNull(primaryNameServer, "primaryNameServer");
+        this.responsibleAuthorityMailbox = checkNotNull(responsibleAuthorityMailbox, "responsibleAuthorityMailbox");
+        this.serialNumber = serialNumber;
+        this.refreshInterval = refreshInterval;
+        this.retryInterval = retryInterval;
+        this.expireLimit = expireLimit;
+        this.minimumTTL = minimumTTL;
```
trustin: - IDN conversion on `primaryNameServer` or reject non-ASCII domain name.
- Please add range checks for all integer parameters (if necessary)

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsSrvRecord.java
```diff
@@ -0,0 +1,96 @@
...
+     * @param weight the weight field specifies a relative weight for entries with the same priority.
+     * @param port the port on this target host of this service.
+     * @param target the domain name of the target host this SRV record resolves to.
+     */
+    public DefaultDnsSrvRecord(
+            String name, int dnsClass, long timeToLive, int priority, int weight, int port, String target) {
+        super(name, DnsRecordType.SRV, dnsClass, timeToLive);
+        this.priority = priority;
+        this.weight = weight;
+        this.port = port;
```
trustin: Range check

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsTxtRecord.java
```diff
@@ -0,0 +1,83 @@
...
+import static io.netty.util.internal.ObjectUtil.checkNotNull;
+
+public class DefaultDnsTxtRecord extends AbstractDnsRecord implements DnsTxtRecord {
+    private static final Pattern RE_KEY = Pattern.compile("`([`=\\s])");
+    private static final Pattern RE_VALUE = Pattern.compile("``");
+
+    private final String key;
+    private final String value;
+
+    public DefaultDnsTxtRecord(String name, int dnsClass, long timeToLive, String text) {
```
trustin: How about adding a constructor that accepts `String key, String value` instead of `String text`?

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsTxtRecord.java
```diff
@@ -0,0 +1,83 @@
...
+            key = "";
+            value = s;
+        }
+
+        key = RE_KEY.matcher(key).replaceAll("$1").toLowerCase(Locale.US);
+        value = RE_VALUE.matcher(value).replaceAll("`");
+
+        return new String[]{key, value};
+    }
+}
```
trustin: Missing `toString()`

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsAAAARecord.java
flier: I think `AAAA` will be better, because most DNS related document direct use `AAAA`

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsMxRecord.java
flier: I think the `preference` value only be used to compare with each other, it could be any number, right?

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsTxtRecord.java
flier: sure, let's move `decodeTxt` to `DefaultDnsRecordDecoder`, and pass the decoded `key` and `value` to `DefaultDnsTxtRecord`

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsMxRecord.java
flier: sure, we can use `exchange` instead of `hostname`

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsNsRecord.java
flier: we can use `nameserver` or `domain` instead of `hostname`

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsTxtRecord.java
flier: added

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsRecordEncoder.java
flier: added

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsRecordDecoder.java
flier: done.

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsRecordDecoder.java
flier: done

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsRecordDecoder.java
flier: done

codec-dns/src/main/java/io/netty/handler/codec/dns/DefaultDnsRecordDecoder.java
flier: I'm not sure why the TXT record cannot be longer than 255, because [RFC 1035](https://tools.ietf.org/html/rfc1035#section-3.3.14) and [RFC 1464](https://tools.ietf.org/html/rfc1464) haven't limits, it seems base on the DNS server implementations, even the Wiki give a suggestion

> Value: This can be free form text data of any type. Each word is treated as a separate string unless one or more strings are enclosed in quotes. The maximum length of a string is 255 bytes, but any number of strings can be present in each record, subject to the standard DNS RDATA size limit of 65535 bytes.[8]

normanmaurer:
Let me close this for now as we never were able to finish this one. Sorry

PR #6629, opened 2017-04-16
Fix regression in QueryStringEncoder

blucas:
Motivation:

Commit d675febf07d14d4dff82471829f974369705655a introduced a regression in QueryStringEncoder, resulting in whitespace being converted into a literal `+` sign instead of `%20`.

Modification:

Modify `encodeComponent` to pattern match and replace on the result of the call to `URLEncoder#encode`

Result:

Fixes regression


normanmaurer:
@blucas LGTM... will pull in soon

normanmaurer:
Cherry-picked into 4.1 (dcd322dda2dfd1e0567017d2e02c53728c310032) and 4.0 (9627cdd0297cb3faabde5b340902ca258fb5ffb7)

Issue #6627, opened 2017-04-14
Some ScheduledExecutorServices like AbstractScheduledEventExecutor break the contract of the interface


dapengzhang0:
### Expected behavior
As the javadoc of `ScheduledExecutorServices` stated: 

```
Zero and negative delays (but not periods) are also allowed in schedule methods, 
and are treated as requests for immediate execution.
```

### Actual behavior
For passing a negative delay, `AbstractScheduledEventExecutor#schedule` throws
`IllegalArgumentException`

### Netty version
4.1.3 Final

normanmaurer:
Sounds like a bug.... let me fix it

normanmaurer:
@dapengzhang0 PTAL https://github.com/netty/netty/pull/6633

normanmaurer:
Fixed by https://github.com/netty/netty/pull/6633

PR #6633, opened 2017-04-18
AbstractScheduledEventExecutor.schedule(...) must accept delay <= 0.

normanmaurer:
Motivation:

As the javadoc of ScheduledExecutorService state:

Zero and negative delays (but not periods) are also allowed in schedule methods,and are treated as requests for immediate execution.

Modifications:

- Correctly handle delay <= 0.
- Add unit tests.

Result:

Fixes [#6627].

### DIFF
--- a/common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java
+++ b/common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java
@@ -131,12 +131,11 @@ public abstract class AbstractScheduledEventExecutor extends AbstractEventExecut
     }
 
     @Override
-    public  ScheduledFuture<?> schedule(Runnable command, long delay, TimeUnit unit) {
+    public ScheduledFuture<?> schedule(Runnable command, long delay, TimeUnit unit) {
         ObjectUtil.checkNotNull(command, "command");
         ObjectUtil.checkNotNull(unit, "unit");
         if (delay < 0) {
-            throw new IllegalArgumentException(
-                    String.format("delay: %d (expected: >= 0)", delay));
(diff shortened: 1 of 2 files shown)


common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java
```diff
@@ -131,27 +131,30 @@ protected final boolean hasScheduledTasks() {
...
         ObjectUtil.checkNotNull(command, "command");
         ObjectUtil.checkNotNull(unit, "unit");
-        if (delay < 0) {
-            throw new IllegalArgumentException(
-                    String.format("delay: %d (expected: >= 0)", delay));
+        ScheduledFutureTask<Void> task = new ScheduledFutureTask<Void>(
+                this, command, null, ScheduledFutureTask.deadlineNanos(unit.toNanos(delay)));
+        if (delay <= 0) {
+            execute(task);
+            return task;
```
dapengzhang0: This return might be blocked by `execute(task)` and can't return immediately. 
The following can be regarded as "immediate execution" in some sense
```
if (delay <0) {
    delay = 0;
}
return schedule(new ScheduledFutureTask<Void>(
    this, command, null, ScheduledFutureTask.deadlineNanos(unit.toNanos(delay))));
```

normanmaurer: Good point let me adjust it

normanmaurer:
done

normanmaurer:
Cherry-picked into 4.1 (1b0b8f80cd) and 4.0 (59a8292e36)

Issue #6622, opened 2017-04-11
VoidChannelPromise eats exceptions...


ninja-:
### Expected behavior

```
voidPromise
ChannelPromise voidPromise()
Return a special ChannelPromise which can be reused for different operations.
It's only supported to use it for write(Object, ChannelPromise).

Be aware that the returned ChannelPromise will not support most operations and should only be used if you want to save an object allocation for every write operation. 
You will not be able to detect if the operation was complete, only if it failed as the implementation will call ChannelPipeline.fireExceptionCaught(Throwable) in this case.
```

when writing a packet with a void promise, an exception should be thrown on failure.

### Actual behavior
the exception is eaten and it's not forwarded to the channel pipeline

### Minimal yet complete reproducer code (or URL to code)
This is CLEARLY a Bug.
in this function https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java#L736
when exception is thrown the logic goes to here
https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java#L834
and...the exception is ignored.
If the promise is void, it should be forwarded to the pipeline as stated in the API...

cc @Scottmitch @normanmaurer 
### Netty version
4.1.9-FINAL

ninja-:
```patch

From 7beda8e105e05eb3b3c7e0c6b2f99a069370137f Mon Sep 17 00:00:00 2001
Date: Tue, 11 Apr 2017 17:34:36 +0200
Subject: [PATCH] Fix encoder exceptions being discarded when writing with a VoidPromise.

Motivation:

According to documentation of Channel#voidPromise(), if encoding of the message throws an exception, exception should be passed to the channel pipeline.
Unfortunately, that's not the case, which this commit fixes.
I've been ripping my hair out for hours trying to find out a memory leak without any clue.
There was an exception thrown in my handler which made the reason very obvious, but it was being silently discarded.
Fixes #6622

Modifications:

Fix the logic of notifyOutboundHandler to handle a case when the promise is not void.

Result:

Encoder exceptions will be properly propagated to the pipeline when used with void promise, in accordance to javadocs of Channel#voidPromise()
---
 transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java | 2 ++
 1 file changed, 2 insertions(+), 0 deletions(-)

[...]

PR #6634, opened 2017-04-18
VoidChannelPromise not notified when exception is thrown.

normanmaurer:
Motivation:

When a VoidChannelPromise is used by the user we need to ensure we propergate the exception through the ChannelPipeline otherwise the exception will just be swallowed and so the user has no idea whats going on.

Modifications:

- Always call tryFailure / trySuccess even when we use the VoidChannelPromise
- Add unit test

Result:

Fixes [#6622].

### DIFF
--- a/transport/src/main/java/io/netty/channel/ChannelOutboundBuffer.java
+++ b/transport/src/main/java/io/netty/channel/ChannelOutboundBuffer.java
@@ -664,15 +664,15 @@ public final class ChannelOutboundBuffer {
     }
 
     private static void safeSuccess(ChannelPromise promise) {
-        if (!(promise instanceof VoidChannelPromise)) {
-            PromiseNotificationUtil.trySuccess(promise, null, logger);
-        }
+        // Only log if the given promise is not of type VoidChannelPromise as trySuccess(...) is expected to return
+        // false.
+        PromiseNotificationUtil.trySuccess(promise, null, promise instanceof VoidChannelPromise ? null : logger);
     }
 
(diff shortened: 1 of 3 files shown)


ninja-:
👍

transport/src/main/java/io/netty/channel/ChannelOutboundBuffer.java
```diff
@@ -664,15 +664,15 @@ public void run() {
...
+        PromiseNotificationUtil.trySuccess(promise, null, promise instanceof VoidChannelPromise ? null : logger);
     }
 
     private static void safeFail(ChannelPromise promise, Throwable cause) {
-        if (!(promise instanceof VoidChannelPromise)) {
-            PromiseNotificationUtil.tryFailure(promise, cause, logger);
-        }
+        // Only log if the given promise is not of type VoidChannelPromise as tryFailure(...) is expected to return
+        // false.
+        PromiseNotificationUtil.tryFailure(promise, cause, promise instanceof VoidChannelPromise ? null : logger);
```
trustin: How about moving `promise instanceof VoidChannelPromise` into `PromiseNotificationUtil`?

normanmaurer:
Doesn't work as these are in different jars

ninja-:
When you guys agree on a merge, can we do a version bump afterwards? Would make things so much easier for me :)

normanmaurer:
We want will do a release within the next 1,5 weeks

normanmaurer:
Cherry-picked into 4.1 (119383873d) and 4.0 (c62564cf4c)

PR #6638, opened 2017-04-18
Correctly release all buffers in UnpooledTest

normanmaurer:
Motivation:

We not correctly released all buffers in the UnpooledTest and so showed "bad" way of handling buffers to people that inspect our code to understand when a buffer needs to be released.

Modifications:

Explicit release all buffers.

Result:

Cleaner and more correct code.

### DIFF
--- a/buffer/src/test/java/io/netty/buffer/UnpooledTest.java
+++ b/buffer/src/test/java/io/netty/buffer/UnpooledTest.java
@@ -15,20 +15,17 @@
  */
 package io.netty.buffer;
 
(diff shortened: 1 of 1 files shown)


normanmaurer:
Sure I can do

normanmaurer:
@trustin actually I think it is ok as it is and we not use finally in most tests as well.

normanmaurer:
Cherry-picked into 4.1 (b662afeece) and 4.0 (b0fd0619db57c8b733574862a923390b59ab362b)

PR #6636, opened 2017-04-18
Http2FrameCodec leaking direct buffer.

durigon:
Motivation:

Http2FrameCodec  is incrementing the reference count in the onDataRead method for the Http2DataFrame buffer.
This may result in Leaking direct buffer. We need to ensure we not retain the buffer as we may leak memory.

Modifications:

Not call retain for Http2FrameCodec.onDataRead.

Result:

No more leak for Http2FrameCodec.onDataRead.

Motivation:

Explain here the context, and why you're making that change.
What is the problem you're trying to solve.

Modification:

Describe the modifications you've done.

Result:

Fixes #<GitHub issue number>. 

If there is no issue then describe the changes introduced by this PR.

PR #6642, opened 2017-04-19
Correctly manage buffer life-cycle in http2 multiplex example

normanmaurer:
Motivation:

We not correctly managed the life-cycle of the buffer / frames in our http2 multiplex example which lead to a memory leak.

Modifications:

- Correctly release frame if not echo'ed back the remote peer.
- Not retain content before echo back to remote peer.

Result:

No more leak in the example, fixes [#6636].

### DIFF
--- a/example/src/main/java/io/netty/example/http2/helloworld/multiplex/server/HelloWorldHttp2Handler.java
+++ b/example/src/main/java/io/netty/example/http2/helloworld/multiplex/server/HelloWorldHttp2Handler.java
@@ -65,7 +65,10 @@ public class HelloWorldHttp2Handler extends ChannelDuplexHandler {
      */
     public void onDataRead(ChannelHandlerContext ctx, Http2DataFrame data) throws Exception {
         if (data.isEndStream()) {
-            sendResponse(ctx, data.content().retain());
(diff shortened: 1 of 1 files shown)


normanmaurer:
@buchgr FYI...

example/src/main/java/io/netty/example/http2/helloworld/multiplex/server/HelloWorldHttp2Handler.java
```diff
@@ -65,7 +65,10 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
      */
     public void onDataRead(ChannelHandlerContext ctx, Http2DataFrame data) throws Exception {
         if (data.isEndStream()) {
-            sendResponse(ctx, data.content().retain());
+            sendResponse(ctx, data.content());
+        } else {
+            // We not send back the response to the remote-peer so we need to release it.
```
trustin: - We not -> We do not
- Insert comma (,) between 'remote-peer' and 'so'.

normanmaurer:
Addressed comments of @trustin and cherry-picked into 4.1 (38483e8790)

Issue #6620, opened 2017-04-11
Base64 decoder can handle buffers larger or equal to 715827883 bytes


trollheim:
for byte buffers larger than 715827882 bytes this code 
ByteBuf bb = Unpooled.buffer(size);
Base64.decode(bb, 0, size);
throws exception
Exception in thread "main" java.lang.IllegalArgumentException: initialCapacity: -536870911 (expectd: 0+)
 
It is happening  because line 120 in  io.netty.handler.codec.base64.Base64 class, `int len43 = len * 4 / 3;` can cause integer overflow for values larger than Integer.MAX_VALUE/3.
changing order of operations to be division followed by multiplication should solve this issue.
Problem was spotted in version 4.1.8.Final.

normanmaurer:
Can you please open a or with a fix? Even better when containing a unit test

normanmaurer:
Fixed by https://github.com/netty/netty/pull/6643

PR #6404, opened 2017-02-17
Improve performance of Base64.decode and encode methods.

normanmaurer:
Motivation:

The decode and encode method uses getByte(...) and setByte(...) in loops which can be very expensive because of bounds / reference-count checking. Beside this it also slows-down a lot when paranoid leak-detection is enabled as it will track each access.

Modifications:

- Pack bytes into int / short and so reduce operations on the ByteBuf
- Use ByteProcessor to reduce getByte calls.

Result:

Better performance in general. Also when you run the build with -Pleak the handler module will build in 1/4 of the time it took before.

PR #6643, opened 2017-04-19
Fix possible overflow when calculate in the size of the out buffer in…

normanmaurer:
… Base64

Motivation:

We not correctly guarded against overflow and so call Base64.encode(...) with a big buffer may lead to an overflow when calculate the size of the out buffer.

Modifications:

Correctly guard against overflow.

Result:

Fixes [#6620].

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/base64/Base64.java
+++ b/codec/src/main/java/io/netty/handler/codec/base64/Base64.java
@@ -120,11 +120,7 @@ public final class Base64 {
             throw new NullPointerException("dialect");
         }
 
-        int len43 = (len << 2) / 3;
-        ByteBuf dest = allocator.buffer(
-                len43 +
-                        (len % 3 > 0 ? 4 : 0) + // Account for padding
-                        (breakLines ? len43 / MAX_LINE_LENGTH : 0)).order(src.order()); // New lines
+        ByteBuf dest = allocator.buffer(encodedBufferSize(len, breakLines)).order(src.order());
         byte[] alphabet = alphabet(dialect);
         int d = 0;
         int e = 0;
@@ -197,6 +193,21 @@ public final class Base64 {
             }
             encode3to4LittleEndian(inBuff, numSigBytes, dest, destOffset, alphabet);
         }
+    }
+
+    // package-private for testing
+    static int encodedBufferSize(int len, boolean breakLines) {
+        // Cast len to long to prevent overflow
+        long len43 = ((long) len << 2) / 3;
+
+        // Account for padding
+        long ret = (len43 + 3) & ~3;
+
+        if (breakLines) {
+            ret += len43 / MAX_LINE_LENGTH;
+        }
+
+        return ret < Integer.MAX_VALUE ? (int) ret : Integer.MAX_VALUE;
     }
 
     private static int toInt(byte value) {
@@ -312,6 +323,11 @@ public final class Base64 {
         return new Decoder().decode(src, off, len, allocator, dialect);
     }
 
+    // package-private for testing
+    static int decodedBufferSize(int len) {
+        return len - (len >>> 2);
+    }
+
     private static final class Decoder implements ByteProcessor {
         private final byte[] b4 = new byte[4];
         private int b4Posn;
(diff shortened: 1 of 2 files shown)


codec/src/main/java/io/netty/handler/codec/base64/Base64.java
```diff
@@ -120,11 +120,10 @@ public static ByteBuf encode(
...
         }
 
-        int len43 = (len << 2) / 3;
-        ByteBuf dest = allocator.buffer(
-                len43 +
-                        (len % 3 > 0 ? 4 : 0) + // Account for padding
-                        (breakLines ? len43 / MAX_LINE_LENGTH : 0)).order(src.order()); // New lines
+        int len43 = len43(len);
+        ByteBuf dest = allocator.buffer((int) Math.min(Integer.MAX_VALUE, (long) len43 +
+                (len % 3 > 0 ? 4 : 0) + // Account for padding
```
fenik17: This is not correct alignement... And further the value can be increased.
I suggest extract the entire calculation of the buffer size in a separate method (without superfluous casts int<->long and overflow check) :

```java
static int getEncodeBufferSize(int len, boolean breakLines) {
    long ret = len + len / 3;
    if (breakLines) {
        ret += ret / MAX_LINE_LENGTH;
    }
    // align by 4
    ret = (ret + 3) & ~3;
    return ret < Integer.MAX_VALUE ? (int) ret : Integer.MAX_VALUE;
}
```

codec/src/main/java/io/netty/handler/codec/base64/Base64.java
```diff
@@ -312,6 +316,11 @@ public static ByteBuf decode(
         return new Decoder().decode(src, off, len, allocator, dialect);
     }
 
+    // package-private for testing
+    static int len34(int len) {
+        return (len * 3) >>> 2;
```
fenik17: This don't fix overflow. `(2147483647 * 3) >>> 2 == 536870911`, but must be `1610612735`.
I already offered an [option earlier](https://github.com/netty/netty/pull/6404#discussion_r101889166):

`int len34 = len - (len >>> 2);`

normanmaurer:
@fenik17 maybe you want to provide a PR with a proper fix ?

fenik17:
@normanmaurer no need to open another PR. Just, I remembered that I had already seen a `len34` :)
And, maybe, I'm wrong about alignment (don't know for why this "Account for padding").

normanmaurer:
@fenik17 PTAL again

codec/src/main/java/io/netty/handler/codec/base64/Base64.java
```diff
@@ -199,6 +195,21 @@ private static void encode3to4(
...
+        long len43 = ((long) len << 2) / 3;
+
+        // Account for padding
+        long ret = (len43 + 3) & ~3;
+
+        if (breakLines) {
+            ret += len43 / MAX_LINE_LENGTH;
+        }
+
+        return ret < Integer.MAX_VALUE ? (int) ret : Integer.MAX_VALUE;
```
Scottmitch: if we overflow at this point can we still produce meaningful results, or are we dead?

normanmaurer: as this is the "worse" case size that we need we may still be able to do so. At worse we will later on produce an IndexOutOfBoundsException when filling the buffer. So I think it is ok to just return Integer.MAX_VALUE here.

normanmaurer:
Cherry-picked into 4.1 (34ff9cf5f2) and 4.0 (ae5dba53932e629acf86f1c6a1b455ed61a1312c)

codec/src/main/java/io/netty/handler/codec/base64/Base64.java
Scottmitch: if there is a chance it may succeed then that sgtm ... if we know it will fail and are relying on `IndexOutOfBoundsException` to be thrown later we should just throw not instead of wasting resources.

normanmaurer:
Yes it still may success

PR #6644, opened 2017-04-19
Fix IllegalArgumentException when release a wrapped ByteBuffer on Java9

normanmaurer:
Motivation:

Unsafe.invokeCleaner(...) checks if the passed in ByteBuffer is a slice or duplicate and if so throws an IllegalArgumentException on Java9. We need to ensure we never try to free a ByteBuffer that was provided by the user directly as we not know if its a slice / duplicate or not.

Modifications:

Never try to free a ByteBuffer that was passed into UnpooledUnsafeDirectByteBuf constructor by an user (via Unpooled.wrappedBuffer(....)).

Result:

Build passes again on Java9


normanmaurer:
@Scottmitch @trustin PTAL... with this change we are able to build netty again with latest java9 release. This will also fix the build problems on the CI with Java9

normanmaurer:
For example this is the error before the fix:
https://garage.netty.io/teamcity/viewLog.html?buildId=19240&buildTypeId=netty_BuildOraclejdk9&tab=buildLog

normanmaurer:
@trustin PTAL when you have a chance as I would love to pull this in and make the CI happy again

normanmaurer:
Cherry-picked into 4.1 (bf0beb772c83fb5fa2240369ec7602a65000e089) and 4.0 (193bb2a56bc143202bc1754f20db806baf1612ed)

PR #6534, opened 2017-03-12
Add support for SNIMatcher when using SslProvider.OPENSSL* and Java8+

normanmaurer:
Motivation:

Java8 adds support for SNIMatcher to reject SNI when the hostname not matches what is expected. We not supported doing this when using SslProvider.OPENSSL*.

Modifications:

- Add support for SNIMatcher when using SslProvider.OPENSSL*
- Add unit tests

Result:

SNIMatcher now support with our own SSLEngine as well.

PR #6650, opened 2017-04-20
Fix buffer leak in local transport when a close triggers the close of…

normanmaurer:
… a remote peer and there are still messages in the inbound buffer.

Motivation:

We need to release all the buffers that may be put into our inbound queue since we closed the Channel to ensure we not leak any memory. This is fine as it basically gives the same guarantees as TCP which  means even if the promise was notified before its not really guaranteed that the "remote peer" will see the buffer at all.

Modifications:

Ensure we release all buffers in the inbound buffer if a doClose() is called.

Result:

No more leaks.

### DIFF
--- a/transport/src/main/java/io/netty/channel/local/LocalChannel.java
+++ b/transport/src/main/java/io/netty/channel/local/LocalChannel.java
@@ -62,7 +62,7 @@ public class LocalChannel extends AbstractChannel {
 
     private final ChannelConfig config = new DefaultChannelConfig(this);
     // To further optimize this we could write our own SPSC queue.
-    private final Queue<Object> inboundBuffer = PlatformDependent.newSpscQueue();
+    final Queue<Object> inboundBuffer = PlatformDependent.newSpscQueue();
     private final Runnable readTask = new Runnable() {
         @Override
         public void run() {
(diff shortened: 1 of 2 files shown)


normanmaurer:
This fixes the leak reported in https://github.com/netty/netty/pull/6534.

transport/src/main/java/io/netty/channel/local/LocalChannel.java
```diff
@@ -246,34 +246,42 @@ protected void doClose() throws Exception {
...
+                        PlatformDependent.throwException(cause);
                     }
-                    PlatformDependent.throwException(cause);
                 }
+            } finally {
+                // We need to release all the buffers that may be put into our inbound queue since we closed the Channel
+                // to ensure we not leak any memory. This is fine as it basically gives the same guarantees as TCP which
+                // means even if the promise was notified before its not really guaranteed that the "remote peer" will
+                // see the buffer at all.
+                releaseInboundBuffers();
```
Scottmitch: to be safe maybe we should only do this if state was not closed, and we should consider wrapping the above code `finishPeerRead` in this finally block too ... it is possible this method may throw if the peer's event loop rejects the close runnable.

normanmaurer: So you basically are saying that we should just move this to the above if block ?

Scottmitch: Feel free to refactor but this is what I was thinking of:

```java
    @Override
    protected void doClose() throws Exception {
        final LocalChannel peer = this.peer;
        boolean stateChangedToClosed = state != State.CLOSED;
        try {
            if (stateChangedToClosed) {
                // Update all internal state before the closeFuture is notified.
                if (localAddress != null) {
                    if (parent() == null) {
                        LocalChannelRegistry.unregister(localAddress);
                    }
                    localAddress = null;
                }

                // State change must happen before finishPeerRead to ensure writes are released either in doWrite or
                // channelRead.
                state = State.CLOSED;

                // Preserve order of event and force a read operation now before the close operation is processed.
                finishPeerRead(this);

                ChannelPromise promise = connectPromise;
                if (promise != null) {
                    // Use tryFailure() instead of setFailure() to avoid the race against cancel().
                    promise.tryFailure(DO_CLOSE_CLOSED_CHANNEL_EXCEPTION);
                    connectPromise = null;
                }
            }

            if (peer != null) {
                this.peer = null;
                // Need to execute the close in the correct EventLoop (see https://github.com/netty/netty/issues/1777).
                // Also check if the registration was not done yet. In this case we submit the close to the EventLoop
                // to make sure its run after the registration completes
                // (see https://github.com/netty/netty/issues/2144).
                EventLoop peerEventLoop = peer.eventLoop();
                final boolean peerIsActive = peer.isActive();
                if (peerEventLoop.inEventLoop() && !registerInProgress) {
                    peer.tryClose(peerIsActive);
                } else {
                    try {
                        peerEventLoop.execute(new Runnable() {
                            @Override
                            public void run() {
                                peer.tryClose(peerIsActive);
                            }
                        });
                    } catch (Throwable cause) {
                        logger.warn("Releasing Inbound Queues for channels {}-{} because exception occurred!",
                                this, peer, cause);
                        if (peerEventLoop.inEventLoop()) {
                            peer.releaseInboundBuffers();
                        } else {
                            // inboundBuffers is a SPSC so we may leak if the event loop is shutdown prematurely or
                            // rejects the close Runnable but give a best effort.
                            peer.close();
                        }
                    }
                }
            }
        } finally {
            if (stateChangedToClosed) {
                // We need to release all the buffers that may be put into our inbound queue since we closed the Channel
                // to ensure we not leak any memory. This is fine as it basically gives the same guarantees as TCP which
                // means even if the promise was notified before its not really guaranteed that the "remote peer" will
                // see the buffer at all.
                releaseInboundBuffers();
            }
        }
    }
```

normanmaurer: I see... let me do something similar

normanmaurer:
Cherry-picked into 4.1 (c663a94359b70069b67b8d7f7bc9d0c43dca7674) and 4.0 (a20d1c930d33fa9a74afd25d0b75a4fb3336e0dd)

PR #6651, opened 2017-04-20
Only use test SslProviders that are supported in SslHandlerTest.testC…

normanmaurer:
…ompositeBufSizeEstimationGuaranteesSynchronousWrite().

Motivation:

We need to ensure we only try to to test with the SslProviders that are supported when running the SslHandlerTest.testCompositeBufSizeEstimationGuaranteesSynchronousWrite test.

Modifications:

Skip SslProvider.OPENSSL* if not supported.

Result:

No more test-failures if openssl is not installed on the system.

### DIFF
--- a/handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
@@ -571,13 +571,19 @@ public class SslHandlerTest {
             throws CertificateException, SSLException, ExecutionException, InterruptedException {
         SslProvider[] providers = SslProvider.values();
         for (int i = 0; i < providers.length; ++i) {
-            for (int j = 0; j < providers.length; ++j) {
-                compositeBufSizeEstimationGuaranteesSynchronousWrite(providers[i], providers[j]);
+            SslProvider serverProvider = providers[i];
+            if (isSupported(serverProvider)) {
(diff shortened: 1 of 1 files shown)


handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
```diff
@@ -569,15 +571,28 @@ public void operationComplete(Future<Channel> future) {
...
+                        continue;
+                    }
+                default:
+                    providers.add(provider);
+            }
+        }
+
+        for (int i = 0; i < providers.size(); ++i) {
+            for (int j = 0; j < providers.size(); ++j) {
+                compositeBufSizeEstimationGuaranteesSynchronousWrite(providers.get(i), providers.get(j));
```
Scottmitch: nit: avoid the intermediate collection and just check if each is supported:

```java
// move this to some test utility ?
public static boolean isSupported(SslProvider provider) {
  switch (provider) {
    case OPENSSL:
    case OPENSSL_REFCNT:
       return OpenSsl.isAvailable();
    default:
      return true;
  }
}
```

normanmaurer:
Cherry-picked into 4.1 (4dd6c14ba275384ce8561ec4a5f139a740a79f7f) and 4.0 (9300a14d08307a98a739e3a394a8e8937f8dbc90)

PR #6655, opened 2017-04-22
Avoid infinite loop in HTTP/2 distributor toString()

ejona86:
Motivation:

Although effectively unused, the toString() of
WeightedFairQueueByteDistributor.State is useful for debugging. It
accidentally had an infinite loop, as it would recurse infinitely
between a parent and its child, which makes it less useful for
debugging.

Modifications:

Prune the infinite loop by using the parent's streamId instead of the
parent's toString().

Result:

Faster, less stack-overflowing toString()


normanmaurer:
@ejona86 can you please rebase ?

Scottmitch:
4.1 (c6ad9338b)

Scottmitch:
Build failure isn't related to this change.

normanmaurer:
Yep build failure is fixed in current 4.1 branch already :)

PR #6656, opened 2017-04-22
Remove the comment which is a bit misleading

### LINKED ISSUE #6652: Misleading comment in `ServerBootstrap`
The mechanism of registering handlers seems to have been changed since #5566 has been fixed, making [this](https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/bootstrap/ServerBootstrap.java#L178) comment misleading.

I did not have time to inspect the whole history, but the user's `initChannel` is now not called after the current invocation returns. As an evidence, the code patched with the following change on tag `4.1.9.Final` also passes [this test case](https://github.com/netty/netty/blob/4.1/transport/src/test/java/io/netty/bootstrap/ServerBootstrapTest.java#L81), which was introduced to ensure that #5566 was fixed.

```
$ git diff
diff --git a/transport/src/main/java/io/netty/bootstrap/ServerBootstrap.java b/transport/src/main/java/io/netty/bootstrap/ServerBootstrap.java
index f9042af46..b73239ffa 100644
--- a/transport/src/main/java/io/netty/bootstrap/ServerBootstrap.java
+++ b/transport/src/main/java/io/netty/bootstrap/ServerBootstrap.java
@@ -174,7 +174,9 @@ public class ServerBootstrap extends AbstractBootstrap<ServerBootstrap, ServerCh
                 if (handler != null) {
                     pipeline.addLast(handler);
                 }
-
+                pipeline.addLast(new ServerBootstrapAcceptor(
+                        ch, currentChildGroup, currentChildHandler, currentChildOptions, currentChildAttrs));
+                /*
                 // We add this handler via the EventLoop as the user may have used a ChannelInitializer as handler.
                 // In this case the initChannel(...) method will only be called after this method returns. Because
                 // of this we need to ensure we add our handler in a delayed fashion so all the users handler are
@@ -185,7 +187,7 @@ public class ServerBootstrap extends AbstractBootstrap<ServerBootstrap, ServerCh
                         pipeline.addLast(new ServerBootstrapAcceptor(
                                 ch, currentChildGroup, currentChildHandler, currentChildOptions, currentChildAttrs));
                     }
-                });
+                });*/
             }
         });
     }
```

Please consider changing the comment, it could be very confusing for newcomers.

### Expected behavior

N/A

### Actual behavior

N/A

### Steps to reproduce

N/A

N/A

### Netty version

4.1.9.Final

java version "1.8.0_121"
Java(TM) SE Runtime Environment (build 1.8.0_121-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.121-b13, mixed mode)

Linux herberteuler 4.9.22 #1-NixOS SMP Wed Apr 12 10:41:42 UTC 2017 x86_64 GNU/Linux

herberteuler:
Motivation:

The invocation of `initChannel` of `ChannelInitializer` has been moved to as early as during `handlerAdded` is invoked in 26aa34853, whereas it was only invoked during `channelRegistered` was invoked before that. So the comment does not describe how handlers are added in normal circumstances anymore.

This change deletes the comment, hopefully reduce confusions.

Modification:

The comment is removed. However, the code is kept as-is since there might be unusual cases, and adding `ServerBootstrapAcceptor` via the event loop is always safe to enforce the correct order.

Result:

N/A

Fixes #6652.

### DIFF
--- a/transport/src/main/java/io/netty/bootstrap/ServerBootstrap.java
+++ b/transport/src/main/java/io/netty/bootstrap/ServerBootstrap.java
@@ -175,10 +175,6 @@ public class ServerBootstrap extends AbstractBootstrap<ServerBootstrap, ServerCh
                     pipeline.addLast(handler);
                 }
 
-                // We add this handler via the EventLoop as the user may have used a ChannelInitializer as handler.
-                // In this case the initChannel(...) method will only be called after this method returns. Because
-                // of this we need to ensure we add our handler in a delayed fashion so all the users handler are
-                // placed in front of the ServerBootstrapAcceptor.
                 ch.eventLoop().execute(new Runnable() {
                     @Override
                     public void run() {


normanmaurer:
Cherry-picked into 4.1 (2d38a4453c360bcd72dbefe4ec04adcdb19f8ddd) and 4.0 (9f1ceb9e57).

@herberteuler thanks!

Issue #6657, opened 2017-04-22
IpSubnetFilterRule throws ClassCastException when non expected IP version comes


doom369:
### Expected behavior

No error. `IpSubnetFilterRule.match` returns false.

### Actual behavior

`IpSubnetFilterRule.match` throws `ClassCastException`.

### Steps to reproduce

```
IpSubnetFilterRule rule = new IpSubnetFilterRule("0.0.0.0", 0, IpFilterRuleType.ACCEPT);

//remoteAddress.getAddress() instanceof Inet6Address
rule.matches(remoteAddress); //throws ClassCastException
```

### Minimal yet complete reproducer code (or URL to code)

```
    @Test
    public void testIpv4SubnetMaskCorrectlyHandlesIpv6() {
        IpSubnetFilterRule rule = new IpSubnetFilterRule("0.0.0.0", 0, IpFilterRuleType.ACCEPT);
        Assert.assertFalse(rule.matches(newSockAddress("2001:db8:abcd:0000::1")));
    }
```

### Netty version

4.1.9-Final

normanmaurer:
Fixed by https://github.com/netty/netty/pull/6658

PR #6658, opened 2017-04-22
#6657 do not throw ClassCastException when rule subnet version doesn'…

doom369:
…t match remote IP version

Motivation:

`IpSubnetFilterRule.match` throws excpetion while should return false for another IP version.

Modification:

Added `instanceOf` check for `IpSubnetFilterRule.match` method. So Rule will not throw `ClassCastException` when remote address has another IP version.

Result:

Fix for https://github.com/netty/netty/issues/6657

### DIFF
--- a/handler/src/main/java/io/netty/handler/ipfilter/IpSubnetFilterRule.java
+++ b/handler/src/main/java/io/netty/handler/ipfilter/IpSubnetFilterRule.java
@@ -91,9 +91,12 @@ public final class IpSubnetFilterRule implements IpFilterRule {
 
         @Override
         public boolean matches(InetSocketAddress remoteAddress) {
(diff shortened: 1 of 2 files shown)


johnou:
Seems like a bit of a hack, why not support IPV6?

doom369:
@johnou could you please explain what do you mean? `IpSubnetFilterRule` supports both IPv4 and IPv6. However, in case I have only 1 rule for IPv4 and remote address is IPv6 I'm getting exception, while filter should return just `false`.

johnou:
wow, okay I see what you mean, after checking out RuleBasedIpFilter this makes sense.

normanmaurer:
Cherry-picked into 4.1 (e78ccd6d52450f30e648abca22b8f2afec62f45d).

@doom369 thanks!

PR #6659, opened 2017-04-22
HTTP/2 StreamByteDistributor improve parameter validation

Scottmitch:
Motivation:
Each StreamByteDistributor may allow for priority in different ways, but there are certain characteristics which are invalid regardless of the distribution algorithm. We should validate these invalid characteristics at the flow controller level.

Modifications:
- Disallow negative stream IDs from being used. These streams may be accepted by the WeightedFairQueueByteDistributor and cause state for other valid streams to be evicted.
- Improve unit tests to verify limits are enforced.

Result:
Boundary conditions related to the priority parameters are validated more strictly.


Scottmitch:
@ejona86 - FYI

normanmaurer:
@Scottmitch please fix check-style

Scottmitch:
4.1 (d21f2adb9)

PR #6667, opened 2017-04-24
Improvement : allocate less objects for multipart form parsing.

doom369:
Motivation:

MultipartRequestDecoder allocates unnecessary objects during multi-part form parsing.

Modification:

`toLowerCase().startsWith` constructions replaced with `regionMatches(true)`.

Result:

Less allocations, faster matching.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
@@ -764,12 +764,12 @@ public class HttpPostMultipartRequestDecoder implements InterfaceHttpPostRequest
                     }
                 } else {
                     for (int i = 1; i < contents.length; i++) {
-                        if (contents[i].toLowerCase().startsWith(HttpHeaderValues.CHARSET.toString())) {
+                        final String charsetHeader = HttpHeaderValues.CHARSET.toString();
+                        if (contents[i].regionMatches(true, 0, charsetHeader, 0, charsetHeader.length())) {
                             String values = StringUtil.substringAfter(contents[i], '=');
                             Attribute attribute;
                             try {
-                                attribute = factory.createAttribute(request, HttpHeaderValues.CHARSET.toString(),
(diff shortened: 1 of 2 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
```diff
@@ -764,12 +764,12 @@ private InterfaceHttpData findMultipartDisposition() {
                     }
                 } else {
                     for (int i = 1; i < contents.length; i++) {
-                        if (contents[i].toLowerCase().startsWith(HttpHeaderValues.CHARSET.toString())) {
+                        final String charsetHeader = HttpHeaderValues.CHARSET.toString();
+                        if (contents[i].regionMatches(true, 0, charsetHeader, 0, charsetHeader.length())) {
```
fenik17: Or maybe this?
```java
if (HttpHeaderValues.CHARSET.regionMatches(true, 0, contents[i], 0, HttpHeaderValues.CHARSET.length())) {
    ...
}
```

doom369: It may change existing flows. My commit doesn't change any flows, just performance improvement. This part is not covered well with tests so I wouldn't risk here. Maybe later after deeper code review.

fenik17: Honestly, I don't see any risk here..
By the way, it would be nice to cach `contents[i]` in a variable.

fenik17: Although no, I think you variant will be faster due to the fact that `charAt()` is not used.

normanmaurer:
Cherry-picked into 4.0 (d7a0e9b8411f744a5f190a821273524a2628944e) and 4.1 (a45f9d7939e0e121011e09526015c2719007a6f5)

normanmaurer:
@doom369 thx

Issue #6671, opened 2017-04-25
ProxyHandler.connectFuture and its execution order


vkostyukov:
As you probably aware, Finagle has adopted `ProxyHandler` in its Netty 4 implementation (see [Netty4ProxyConnectHandler](https://github.com/twitter/finagle/blob/develop/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/proxy/Netty4ProxyConnectHandler.scala)) and it works just great for us, besides one quite surprising behavior its API exposes.

We've recently discovered that `ProxyHandler.connectFuture` [gets notified/satisfied before it removes its underlying proxy codec pieces from the pipeline](https://github.com/netty/netty/blob/4.1/handler-proxy/src/main/java/io/netty/handler/proxy/ProxyHandler.java#L290-L298), which means any possible subsequent write, triggered on `connectFuture.onSuccess` (which I would assume quite an intuitive way of using that future) will fail given the pipeline is malformed. Normally, we'd assume that `ProxyHandler` will queue those wires and drain them later when the pipeline is cleaned up. Although, this is not what exactly happens since when the connect is finished (we set that flag before notifying the listeners) [we just bypass outbound messages](https://github.com/netty/netty/blob/4.1/handler-proxy/src/main/java/io/netty/handler/proxy/ProxyHandler.java#L397-L399). Simply speaking, the following would never work (sorry for Scala):

```scala
proxyHandler.connectFuture.addListener(new GenericFutureListener[NettyFuture[Channel]] {
  override def operationComplete(future: NettyFuture[Channel]): Unit = {
    if (future.isSuccess) {
      ctx.write(something) // will fail since proxyHandler's codec is still in the pipeline
    }
}
```

This is slightly misleading so I suggested we either:

 - document it
 - run the promise **after** we remove the proxy codec

I'm happy to open a PR as usual (with docs or the fix). What do you think about that?

PS: We've workaround that by running `ctx.wirte` in the event loop (`ctx.executor.schedule()`) such that it's queued.

normanmaurer:
let me fix it by notify the promise after removal :)

normanmaurer:
@vkostyukov please check https://github.com/netty/netty/pull/6672

normanmaurer:
Fixed by https://github.com/netty/netty/pull/6672

vkostyukov:
Thanks a lot, @normanmaurer for such a quick turn around!

normanmaurer:
[...]

PR #6672, opened 2017-04-26
Notify connect promise of ProxyHandler after codecs are removed

normanmaurer:
Motivation:

We need to notify the promise after the codecs are removed to allow writing from the listener and ensure we not try to do any encoding anymore. If we not do we will end up with corrupt data.

Modifications:

Notify promise after codecs are removed.

Result:

Fixes [#6671].

### DIFF
--- a/handler-proxy/src/main/java/io/netty/handler/proxy/ProxyHandler.java
+++ b/handler-proxy/src/main/java/io/netty/handler/proxy/ProxyHandler.java
@@ -283,11 +283,9 @@ public abstract class ProxyHandler extends ChannelDuplexHandler {
 
     private void setConnectSuccess() {
         finished = true;
-        if (connectTimeoutFuture != null) {
-            connectTimeoutFuture.cancel(false);
-        }
+        cancelConnectTimeoutFuture();
 
-        if (connectPromise.trySuccess(ctx.channel())) {
+        if (!connectPromise.isDone()) {
             boolean removedCodec = true;
 
             removedCodec &= safeRemoveEncoder();
@@ -303,13 +301,12 @@ public abstract class ProxyHandler extends ChannelDuplexHandler {
                 if (flushedPrematurely) {
                     ctx.flush();
                 }
+                connectPromise.trySuccess(ctx.channel());
             } else {
                 // We are at inconsistent state because we failed to remove all codec handlers.
                 Exception cause = new ProxyConnectException(
                         "failed to remove all codec handlers added by the proxy handler; bug?");
-                failPendingWrites(cause);
-                ctx.fireExceptionCaught(cause);
-                ctx.close();
+                failPendingWritesAndClose(cause);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


normanmaurer:
@vkostyukov PTAL

handler-proxy/src/main/java/io/netty/handler/proxy/ProxyHandler.java
```diff
@@ -287,7 +287,7 @@ private void setConnectSuccess() {
             connectTimeoutFuture.cancel(false);
         }
 
-        if (connectPromise.trySuccess(ctx.channel())) {
+        if (!connectPromise.isDone()) {
```
vkostyukov: I guess an alternative solution would be to move `finished = true;` statement right after we remove both codecs (so that the subsequent `write` triggered in the listener will be enqueued). I honestly can't tell the difference between these two approaches so LGTM!

normanmaurer: @vkostyukov yeah... I think my fix is fine as well. So if @trustin and @Scottmitch reviewed it I will merge it.

handler-proxy/src/main/java/io/netty/handler/proxy/ProxyHandler.java
```diff
@@ -287,7 +287,7 @@ private void setConnectSuccess() {
...
             boolean removedCodec = true;
 
             removedCodec &= safeRemoveEncoder();
```
Scottmitch: consider using boolean operation while we are here (instead of bitwise operations)

handler-proxy/src/main/java/io/netty/handler/proxy/ProxyHandler.java
```diff
@@ -287,7 +287,7 @@ private void setConnectSuccess() {
             connectTimeoutFuture.cancel(false);
```
Scottmitch: should we do `connectTimeoutFuture = null;` here (and other spots where we call `connectTimeoutFuture.cancel(..);`)?

handler-proxy/src/main/java/io/netty/handler/proxy/ProxyHandler.java
Scottmitch: If we expect this to be called frequently consider using a non-volatile variable to avoid duplicate processing. For example can we use `finished` instead?

handler-proxy/src/main/java/io/netty/handler/proxy/ProxyHandler.java
```diff
@@ -347,11 +349,12 @@ private void setConnectFailure(Throwable cause) {
                     exceptionMessage(cause.toString()), cause);
         }
 
-        if (connectPromise.tryFailure(cause)) {
+        if (!connectPromise.isDone()) {
             safeRemoveDecoder();
             safeRemoveEncoder();
 
             failPendingWrites(cause);
```
Scottmitch: consider extracting this into a private method and share code with `setConnectSuccess`'s `else` clause above.

handler-proxy/src/main/java/io/netty/handler/proxy/ProxyHandler.java
normanmaurer: @Scottmitch nope not expected to be called frequently. So I would prefer to keep it as it is

handler-proxy/src/main/java/io/netty/handler/proxy/ProxyHandler.java
normanmaurer: @Scottmitch let me do this is a followup.

normanmaurer:
Cherry-picked into 4.1 (8b13d9656a)

PR #6676, opened 2017-04-27
Netty BOM inherit from parent and its own dependency management

### LINKED ISSUE #6674: Netty BOM inherit from parent and its own dependency management
A BOM was recently added in #5994 but is not valid.

A BOM should only expose dependency management for the artifacts that the project produces (so in this case, netty artifacts). It should not have an opinion about third party dependencies (otherwise including that BOM in a project potentially overrides decisions that were made about them).

One way to fix that would be to stop inheriting from `netty-parent`. Also, I'd set the version of each dependency in the bom rather than using `${project.version}`. If you're including the bom for netty X.Y.Z, it should be X.Y.Z. and nothing else (Maven does not allow to override that property, which is good so it's not strictly necessary but the intent is more clear if you hardcode the current version).

Here is an example of BOM for reference: http://repo.spring.io/release/org/springframework/spring-framework-bom/4.3.8.RELEASE/spring-framework-bom-4.3.8.RELEASE.pom

We'd like to [use this bom in Spring Boot](https://github.com/spring-projects/spring-boot/issues/8553) but we can't for that reason.

### LINKED ISSUE #6677: Netty bom does not provide dependency management for `netty-all`
The netty BOM should provide dependency management for all netty artifacts. I don't see `netty-all` in the list. Can you please add it?

johnou:
Motivation:

A BOM should only expose dependency management for the artifacts that the project produces (so in this case, netty artifacts). It should not have an opinion about third party dependencies (otherwise including that BOM in a project potentially overrides decisions that were made about them).

Modifications:

Stop inheriting from netty-parent and explicitly set the versions for the artifacts.

Result:

Valid BOM.

Fixes #6674, fixes #6677.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -15,136 +15,176 @@
   ~ under the License.
   -->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
-
   <modelVersion>4.0.0</modelVersion>
   <parent>
-    <groupId>io.netty</groupId>
-    <artifactId>netty-parent</artifactId>
-    <version>4.1.10.Final-SNAPSHOT</version>
+    <groupId>org.sonatype.oss</groupId>
+    <artifactId>oss-parent</artifactId>
+    <version>7</version>
+    <relativePath />
   </parent>
 
+  <groupId>io.netty</groupId>
   <artifactId>netty-bom</artifactId>
+  <version>4.1.10.Final-SNAPSHOT</version>
   <packaging>pom</packaging>
 
   <name>Netty/BOM</name>
+  <description>Netty (Bill of Materials)</description>
+  <url>http://netty.io/</url>
+
+  <organization>
+    <name>The Netty Project</name>
+    <url>http://netty.io/</url>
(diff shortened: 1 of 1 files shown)


johnou:
@snicoll PTAL.

bom/pom.xml
```diff
@@ -15,15 +15,17 @@
...
+    <groupId>org.sonatype.oss</groupId>
+    <artifactId>oss-parent</artifactId>
+    <version>7</version>
+    <relativePath />
   </parent>
 
+  <groupId>io.netty</groupId>
   <artifactId>netty-bom</artifactId>
+  <version>4.1.10.Final-SNAPSHOT</version>
   <packaging>pom</packaging>
```
snicoll: According to [this page](https://maven.apache.org/guides/mini/guide-central-repository-upload.html) you need to add the `<url>`, `<license>` and `<scm>` tags to be able to deploy it to Maven central. Check the sample BOM in the related issue for reference.

johnou: roger.

johnou:
@snicoll amended commit, thanks for checking it out 👍

normanmaurer:
@snicoll does this look correct to you ?

snicoll:
@normanmaurer looks good to me. Thanks!

normanmaurer:
Cherry-picked into 4.1 (0123190214c59d764a6d3c79a9570cd20fe6715a) and 4.0 (8e928375c1)

@johnou thanks!

PR #6684, opened 2017-04-28
Mark conscrypt as optional

normanmaurer:
Motivation:

Conscrypt is not needed when using the handler module, so it should be marked as optional

Modifications:

Mark conscrypt as optional

Result:

Be able to use handler module without conscrypt


normanmaurer:
Cherry-picked into 4.1 (3ba4da8a79) and 4.0 (d495ddb0a204cae2cdb601cdb1eb6eddd7095cea)

Issue #6678, opened 2017-04-27
Improve warning when native libraries extracted to volume with `noexec` set


mkw:
### Expected behavior

A Docker-version of our application suddenly failed to work when the `--tmpfs` flag was added to the Docker run command.  Digging into the cause resulted in a surprise that this causes Docker to mount the temp directory with `noexec`; the Netty stack trace (below) does clearly report the `failed to map segment from shared object` error, but this is a bit ambiguous to anyone who did not expect Docker to do this.  Given how popular containers in general, and Docker in particular, has gotten, it seems appropriate to try to log a better message.

### Actual behavior

Only warning is:

```
java.lang.UnsatisfiedLinkError: /tmp/libnetty-transport-native-epoll5589732826809244917.so: /tmp/libnetty-transport-native-epoll5589732826809244917.so: failed to map segment from shared object: Operation not permitted
	at java.lang.ClassLoader$NativeLibrary.load(Native Method) ~[?:1.8.0_121]
	at java.lang.ClassLoader.loadLibrary0(ClassLoader.java:1941) ~[?:1.8.0_121]
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1824) ~[?:1.8.0_121]
	at java.lang.Runtime.load0(Runtime.java:809) ~[?:1.8.0_121]
	at java.lang.System.load(System.java:1086) ~[?:1.8.0_121]
	at io.netty.util.internal.NativeLibraryUtil.loadLibrary(NativeLibraryUtil.java:36) ~[netty-common-4.1.9.Final.jar!/:4.1.9.Final]
	at io.netty.util.internal.NativeLibraryLoader.loadLibrary(NativeLibraryLoader.java:259) ~[netty-common-4.1.9.Final.jar!/:4.1.9.Final]
	at io.netty.util.internal.NativeLibraryLoader.load(NativeLibraryLoader.java:226) ~[netty-common-4.1.9.Final.jar!/:4.1.9.Final]
	at io.netty.channel.epoll.Native.loadNativeLibrary(Native.java:269) ~[netty-transport-native-epoll-4.1.9.Final-linux-x86_64.jar!/:4.1.9.Final]
	at io.netty.channel.epoll.Native.<clinit>(Native.java:64) ~[netty-transport-native-epoll-4.1.9.Final-linux-x86_64.jar!/:4.1.9.Final]
[...]

PR #6685, opened 2017-04-28
Add explicit message when `noexec` prevents library loading.

mkw:
Motivation:

Docker's `--tmpfs` flag mounts the temp volume with `noexec` by default, resulting in an UnsatisfiedLinkError.  While this is good security practice, it is a surprising failure from a seemingly innocuous flag.

Modifications:

Add a best-effort attempt to detect the `noexec` flag in `NativeLibraryLoader`.  This solution is (probably) limited to Oracle/OpenJDK VM's and requires Java 7 or higher, and also requires `--add-opens=java.base/sun.nio.fs=...` for Java 9.  It avoids error message parsing and it does not directly attempt to read mount information.

Result:

Fixes [#6678].

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
@@ -356,6 +356,7 @@ public final class HttpUtil {
     /**
      * Fetch charset from message's Content-Type header.
      *
+     * @param message entity to fetch Content-Type header from
      * @return the charset from message's Content-Type header or {@link io.netty.util.CharsetUtil#ISO_8859_1}
      * if charset is not presented or unparsable
      */
@@ -364,17 +365,55 @@ public final class HttpUtil {
     }
 
     /**
+     * Fetch charset from Content-Type header value.
+     *
+     * @param contentTypeValue Content-Type header value to parse
+     * @return the charset from message's Content-Type header or {@link io.netty.util.CharsetUtil#ISO_8859_1}
+     * if charset is not presented or unparsable
+     */
+    public static Charset getCharset(CharSequence contentTypeValue) {
+        if (contentTypeValue != null) {
+            return getCharset(contentTypeValue, CharsetUtil.ISO_8859_1);
+        } else {
+            return CharsetUtil.ISO_8859_1;
+        }
+    }
+
+    /**
      * Fetch charset from message's Content-Type header.
      *
+     * @param message entity to fetch Content-Type header from
+     * @param defaultCharset result to use in case of empty, incorrect or doesn't conain required part header value
      * @return the charset from message's Content-Type header or {@code defaultCharset}
      * if charset is not presented or unparsable
      */
     public static Charset getCharset(HttpMessage message, Charset defaultCharset) {
-        CharSequence charsetCharSequence = getCharsetAsSequence(message);
-        if (charsetCharSequence != null) {
-            try {
-                return Charset.forName(charsetCharSequence.toString());
-            } catch (UnsupportedCharsetException unsupportedException) {
+        CharSequence contentTypeValue = message.headers().get(HttpHeaderNames.CONTENT_TYPE);
+        if (contentTypeValue != null) {
+            return getCharset(contentTypeValue, defaultCharset);
+        } else {
+            return defaultCharset;
+        }
+    }
+
+    /**
+     * Fetch charset from Content-Type header value.
+     *
+     * @param contentTypeValue Content-Type header value to parse
+     * @param defaultCharset result to use in case of empty, incorrect or doesn't contain required part header value
+     * @return the charset from message's Content-Type header or {@code defaultCharset}
+     * if charset is not presented or unparsable
+     */
+    public static Charset getCharset(CharSequence contentTypeValue, Charset defaultCharset) {
+        if (contentTypeValue != null) {
+            CharSequence charsetCharSequence = getCharsetAsSequence(contentTypeValue);
+            if (charsetCharSequence != null) {
+                try {
+                    return Charset.forName(charsetCharSequence.toString());
+                } catch (UnsupportedCharsetException unsupportedException) {
+                    return defaultCharset;
+                }
+            } else {
                 return defaultCharset;
             }
         } else {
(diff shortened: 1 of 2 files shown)


common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -368,4 +380,65 @@ private static void closeQuietly(Closeable c) {
...
+            try {
+                // Use reflection for so Java 6 fails gracefully.
+                toPathMethod = File.class.getMethod("toPath");
+                Class<?> pathClass = Class.forName("java.nio.file.Path");
+                getFileStoreMethod = Class.forName("java.nio.file.Files").getMethod("getFileStore", pathClass);
+                Class<?> unixFileStoreClass = Class.forName("sun.nio.fs.UnixFileStore");
+                mountEntryMethod = unixFileStoreClass.getDeclaredMethod("entry");
+                mountEntryMethod.setAccessible(true);
+                Class<?> unixMountEntryClass = Class.forName("sun.nio.fs.UnixMountEntry");
+                hasOptionMethod = unixMountEntryClass.getDeclaredMethod("hasOption", String.class);
```
normanmaurer: Just use `ReflectionUtils` maybe ?

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -224,6 +224,18 @@ public static void load(String name, ClassLoader loader) {
             out = null;
 
             loadLibrary(loader, tmpFile.getPath(), true);
+        } catch (UnsatisfiedLinkError e) {
+            try {
+                if (tmpFile.isFile() && tmpFile.canRead() && NoexecFileStoreDetector.isOnNoexecFileStore(tmpFile)) {
```
netkins: BLOCKER NullPointerException might be thrown as 'tmpFile' is nullable here [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2259)

mkw:
The revised commit uses `ReflectionUtil#trySetAccessible(AccessibleObject)`.  It was a little confusing to use, given that all the places that use that method except `io.netty.channel.nio.NioEventLoop#openSelector()` catch `SecurityException`, oddly.  It is not at all obvious to me why the `SecurityException` from `AccessibleObject#setAccessible()` is treated differently from  any other `SecurityException`.  For this particular use case, I treated both the same, for now, but I assume that I may be asked to change that.

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -224,6 +224,19 @@ public static void load(String name, ClassLoader loader) {
...
             loadLibrary(loader, tmpFile.getPath(), true);
+        } catch (UnsatisfiedLinkError e) {
+            try {
+                if (tmpFile != null && tmpFile.isFile() && tmpFile.canRead() &&
+                    NoexecFileStoreDetector.isOnNoexecFileStore(tmpFile)) {
+                    logger.info("{} exists and is readable but cannot be loaded because device mounted with noexec; " +
+                                "use -Dio.netty.native.workdir=[path] to set native working directory separately.",
+                                tmpFile.getPath());
+                }
+            } catch (Throwable t) {
```
normanmaurer: t -> ignored

normanmaurer: just a nit...

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -368,4 +381,70 @@ private static void closeQuietly(Closeable c) {
...
+                GET_FILE_STORE_METHOD == null || TO_PATH_METHOD == null) {
+                return false;
+            }
+            try {
+                Object path = TO_PATH_METHOD.invoke(file);
+                Object fileStore = GET_FILE_STORE_METHOD.invoke(null, path);
+                Object mountEntry = MOUNT_ENTRY_METHOD.invoke(fileStore);
+                return (Boolean) HAS_OPTION_METHOD.invoke(mountEntry, "noexec");
+            } catch (Throwable t) {
+                logger.debug("Error checking if {} is on a file store mounted with noexec", file, t);
```
Scottmitch: can we let this method throw and catch/log at a higher level? Currently we just catch a `Throwable` above and ignore it ... seems more intuitive to handle this in a single spot and if higher layers want to do something with the exception they can.

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -368,4 +381,70 @@ private static void closeQuietly(Closeable c) {
...
+
+        static {
+            Method toPathMethod = null;
+            Method getFileStoreMethod = null;
+            Method mountEntryMethod = null;
+            Method hasOptionMethod = null;
+            try {
+                // Use reflection so Java 6 fails gracefully.
+                toPathMethod = File.class.getMethod("toPath");
+                Class<?> pathClass = Class.forName("java.nio.file.Path");
```
Scottmitch: I would prefer if we explicitly reference the type instead of relying upon implicit string/reflection lookup:

```java
if (PlatformDependent.javaVersion() >= 7) {
// reference java.nio.file.Path directly
} else {
  MOUNT_ENTRY_METHOD = null;
  ...
}
```

You will also have to add an expiation to the `pom.xml` to allow this jdk7+ API to be referenced.

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -368,4 +381,70 @@ private static void closeQuietly(Closeable c) {
...
+                getFileStoreMethod = Class.forName("java.nio.file.Files").getMethod("getFileStore", pathClass);
+
+                mountEntryMethod = Class.forName("sun.nio.fs.UnixFileStore").getDeclaredMethod("entry");
+                hasOptionMethod =
+                        Class.forName("sun.nio.fs.UnixMountEntry").getDeclaredMethod("hasOption", String.class);
+
+                Throwable mountEntryThrowable = ReflectionUtil.trySetAccessible(mountEntryMethod);
+                Throwable hasOptionThrowable = ReflectionUtil.trySetAccessible(hasOptionMethod);
+                if (mountEntryThrowable != null || hasOptionThrowable != null) {
+                    if (!(mountEntryThrowable instanceof SecurityException) ||
```
Scottmitch: what exception are you expecting if a module accessibility is the root cause of the failure?

Issue #3063, opened 2014-10-28
WebSocketServerProtocolHandler breaks ChannelHandler's exceptionCaught behavior


mereshow:
`WebSocketServerProtocolHandler` overrides `exceptionCaught` method. If an exception happens in the pipeline before `WebSocketServerProtocolHandler` it closes the channel, leaving the Handlers that follow it unable to handle the `Exception`.
This is the `WebSocketServerProtocolHandler's` `exceptionCaught` code:

``` java
@Override
    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
        if (cause instanceof WebSocketHandshakeException) {
            FullHttpResponse response = new DefaultFullHttpResponse(
                    HTTP_1_1, HttpResponseStatus.BAD_REQUEST, Unpooled.wrappedBuffer(cause.getMessage().getBytes()));
            ctx.channel().writeAndFlush(response).addListener(ChannelFutureListener.CLOSE);
        } else {
            ctx.close();
        }
    }
```

This is my scenario: after an successful handshake, the server has this pipeline:
`Pipeline = [wsdecoder, wsencoder, WS403Responder, WebSocketServerProtocolHandler, domainHandler, DefaultChannelPipeline$TailContext#0]`

My server logic is in domainHandler. I set `maxFrameSize` in `WebSocketServerProtocolHandler` to be 1MB. If I send a file bigger than this wsdecoder (`WebSocket08FrameDecoder`) calls protocolViolation, which sends an `CloseWebSocketFrame` to the client and throws an `CorruptedFrameException`. This exception is then captured by `WebSocketServerProtocolHandler's` `exceptionCaught` that closes the channel and doesn't call the next handler's `exceptionCaught` (my domain handler) leaving me no chance to change the behavior.

This is the `WebSocket08FrameDecoder's` `protocolViolation` code:

``` java
private void protocolViolation(ChannelHandlerContext ctx, CorruptedFrameException ex) {
        state = State.CORRUPT;
        if (ctx.channel().isActive()) {
            Object closeMessage;
            if (receivedClosingHandshake) {
                closeMessage = Unpooled.EMPTY_BUFFER;
            } else {
                closeMessage = new CloseWebSocketFrame(1002, null);
            }
            ctx.writeAndFlush(closeMessage).addListener(ChannelFutureListener.CLOSE);
        }
        throw ex;
    }
```

Thanks for this amazing framework, I'm learning a lot thanks to you :-)

Saludos.

[...]

PR #6688, opened 2017-05-01
Fire exceptionCaught before exception-caused close for WebSockets.

mkw:
Motivation:

WebSocket decoding throws exceptions on failure that should cause the pipline to close.  These are currently ignored in the `WebSocketProtocolHandler` and `WebSocketServerProtocolHandler`.  In
particular, this means that messages exceding the max message size will cause the channel to close with no reported failure.

Modifications:

Re-fire the event just before closing the socket to allow it to be handled appropriately.

Result:

Closes [#3063].

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketProtocolHandler.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketProtocolHandler.java
@@ -39,6 +39,7 @@ abstract class WebSocketProtocolHandler extends MessageToMessageDecoder<WebSocke
 
     @Override
     public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
+        ctx.fireExceptionCaught(cause);
         ctx.close();
     }
 }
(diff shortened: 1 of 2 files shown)


mkw:
If you want a try/finally around it to guarantee closure, let me know.  It did not appear that other similar code bothered with that, but this change may surprise someone.

normanmaurer:
LGTM... @trustin @Scottmitch happy with it as well ?

Scottmitch:
`ChannelHandlerContext#fire*` methods are not permitted to throw ... so no `try/finally` is necessary.

normanmaurer:
Cherry-picked into 4.1 (e70fbe316d63c28c81e175095297b2205e656eca) and 4.0 (5c1480b17799aa5e88edabfec0f10ca166295f69)

normanmaurer:
@mkw thanks so much....

fenik17:
@normanmaurer 4.0 didn't pushed?

normanmaurer:
@fenik17 hmm its in: https://github.com/netty/netty/commit/5c1480b17799aa5e88edabfec0f10ca166295f69

fenik17:
@normanmaurer thanks! now see.

Issue #6689, opened 2017-05-01
HttpContentCompressor.beginEncode


mskaric:
There is a problem in condition. Please remove ! sign.

### Expected behavior

String contentEncoding = headers.headers().get(HttpHeaderNames.CONTENT_ENCODING);
        if (contentEncoding != null &&
            HttpHeaderValues.IDENTITY.contentEqualsIgnoreCase(contentEncoding)) {
            return null;
        }

### Actual behavior

String contentEncoding = headers.headers().get(HttpHeaderNames.CONTENT_ENCODING);
        if (contentEncoding != null &&
            !HttpHeaderValues.IDENTITY.contentEqualsIgnoreCase(contentEncoding)) {
            return null;
        }

normanmaurer:
Can you please open a PR and add a unit-test as well ?


On 1. May 2017 at 13:45:41, mskaric (notifications@github.com) wrote:

There is a problem in condition. Please remove ! sign.
Expected behavior

String contentEncoding =
headers.headers().get(HttpHeaderNames.CONTENT_ENCODING);
if (contentEncoding != null &&
HttpHeaderValues.IDENTITY.contentEqualsIgnoreCase(contentEncoding)) {
return null;
}
Actual behavior

String contentEncoding =
headers.headers().get(HttpHeaderNames.CONTENT_ENCODING);
if (contentEncoding != null &&
!HttpHeaderValues.IDENTITY.contentEqualsIgnoreCase(contentEncoding)) {
return null;
}
Steps to reproduce Minimal yet complete reproducer code (or URL to code) Netty
version JVM version (e.g. java -version) OS version (e.g. uname -a)

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/netty/netty/issues/6689>, or mute the thread

mskaric:
Unfortunately I really have no time to delve into that :). I just wanted to
help. I made this code to go around the problem and now it works as
expected. Hopefully you don't mind.
Setting identity should turn off the compression e.g. return null.


public class WebContentCompressor extends HttpContentCompressor{

    @Override
    protected Result beginEncode(HttpResponse headers, String
acceptEncoding) throws Exception {
        String
encoding=headers.headers().get(HttpHeaderNames.CONTENT_ENCODING);
        if(encoding!=null &&
HttpHeaderValues.IDENTITY.contentEqualsIgnoreCase(encoding))
              return null;
[...]

PR #6690, opened 2017-05-02
Not try to compresses HttpMessage if IDENTITY header value is set.

normanmaurer:
Motivation:

If Content-Encoding: IDENTITY is used we should not try to compress the http message but just let it pass-through.

Modifications:

Remove "!"

Result:

Fixes [#6689]

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
@@ -101,8 +101,7 @@ public class HttpContentCompressor extends HttpContentEncoder {
     @Override
     protected Result beginEncode(HttpResponse headers, String acceptEncoding) throws Exception {
         String contentEncoding = headers.headers().get(HttpHeaders.Names.CONTENT_ENCODING);
-        if (contentEncoding != null &&
-            !HttpHeaders.Values.IDENTITY.equalsIgnoreCase(contentEncoding)) {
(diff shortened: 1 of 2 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
```diff
@@ -102,7 +102,7 @@ public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
     protected Result beginEncode(HttpResponse headers, String acceptEncoding) throws Exception {
         String contentEncoding = headers.headers().get(HttpHeaders.Names.CONTENT_ENCODING);
         if (contentEncoding != null &&
```
fenik17: A `contentEncoding != null` is redundant here.

normanmaurer: +1

normanmaurer:
Cherry-picked into 4.0 (4e885d4061a2936e703c90be6ea2312a437a1cc2) and 4.1 (a3e496a521)

bgleib:
@normanmaurer This change breaks the use case of doing higher-level encoding of the response payload. We were compressing responses with brotli and setting the Content-Encoding header to "br", which previously worked fine because this code would notice the non-empty Content-Encoding and not attempt to re-encode. With this change Netty now tries to re-encode the message despite Content-Encoding already being set.

normanmaurer:
Fixed by https://github.com/netty/netty/pull/6785

PR #6691, opened 2017-05-02
Make HttpHeadersEncoder.encodeHeader package private to match class v…

anuraaga:
... visibility.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpHeadersEncoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpHeadersEncoder.java
@@ -26,7 +26,7 @@ final class HttpHeadersEncoder {
     private HttpHeadersEncoder() {
     }
 
-    public static void encoderHeader(CharSequence name, CharSequence value, ByteBuf buf) throws Exception {
+    static void encoderHeader(CharSequence name, CharSequence value, ByteBuf buf) throws Exception {
         final int nameLen = name.length();
         final int valueLen = value.length();
         final int entryLen = nameLen + valueLen + 4;


anuraaga:
@trustin Any thoughts on this PR?

trustin:
Sounds OK-ish to me, but let me leave the answer to the guys who wrote the majority of this codec. The class could be renamed to avoid the potential confusion it may give to users. (They may think it's a handler or a codec.)

normanmaurer:
@anuraaga I am fine with the change... @Scottmitch fine as well ?

Scottmitch:
Unless there is a reason to make this public ... I vote to make the method package private. I would expect this is an internal implementation detail of codec-http.

anuraaga:
I think that's reasonable too. Just for reference though, I do have a use case here for serializing headers into a ByteBuf in a compliant way.

https://github.com/anuraaga/armeria/blob/e0c75e14ce4725b8e5d89fd72641dd21d35d6b1b/grpc/src/main/java/com/linecorp/armeria/server/grpc/ArmeriaServerCall.java#L415

I feel like simple utilities like this one that implement the http spec have use cases and low maintenance cost, so maybe it's ok to make public, but what do you think?

Scottmitch:
Exposing internal APIs may seem innocuous at the time, but in Netty these can have a very long lifetime and we are stuck supporting them. I'm not sure what is being done in the code you referenced above but seems like a custom protocol that happens to have some overlap with HTTP? Although I see the appeal of making this method public I'm still in the "make it package private" camp.

anuraaga:
@Scottmitch OK - updated PR to reduce visibility of the method instead.

normanmaurer:
Cherry-picked int 4.1 (ba5d1880bc7178b9ec90eeae02bdeb9456b01a0a) . Thanks!

Scottmitch:
thanks @anuraaga !

PR #6696, opened 2017-05-03
Do not log on explicit no unsafe again

jasontedor:
Motivation:

Users should not see a scary log message when Netty is initialized if Netty configuration explicitly disables unsafe. The log message that produces this warning was previously guarded but the guard was lost.

Modifications:

This commit brings back the guard against the scary log message if unsafe is explicitly disabled.

Result:

No log message is produced when unsafe is unavailable because Netty was told to not look for it.

Relates #5624, relates #6568

### DIFF
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent.java
@@ -128,7 +128,11 @@ public final class PlatformDependent {
             logger.debug("-Dio.netty.noPreferDirect: {}", !DIRECT_BUFFER_PREFERRED);
         }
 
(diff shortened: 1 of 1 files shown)


normanmaurer:
@jasontedor LGTM... Thanks a lot again. Will merge once the CI finishes

jasontedor:
Thanks @normanmaurer!

normanmaurer:
Cherry-picked into 4.1 (9a0fd3a7b892438fccff3d63946756160a8647fe) and 4.0 (bf2ab608ec)

jasontedor:
Thanks for merging @normanmaurer.

PR #6697, opened 2017-05-03
Do not try to use cleaner if no unsafe

jasontedor:
Motivation:

If unsafe is unavailable, we can not use the cleaner anyway. If we try to set it up, we get an annoying log message about unsafe being unavailable (when debug logging is enabled). We know this will fail, so we should not even bother and avoid the log message.

Modifications:

This commit adds a guard against setting up the cleaner if it is not going to be available because unsafe is unavailable.

Result:

We do not try to set up the cleaner if unsafe is unavailable, and we do not get an annoying log message.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent.java
@@ -171,7 +171,7 @@ public final class PlatformDependent {
 
         MAYBE_SUPER_USER = maybeSuperUser0();
 
-        if (!isAndroid()) {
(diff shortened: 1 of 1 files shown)


normanmaurer:
Cherry-picked into 4.1 (02a2738cd21038af7e437e67bb4aa1410b3ad43d) and 4.0 (821aec0999)

normanmaurer:
@jasontedor thanks a lot

jasontedor:
Thank you @normanmaurer!

PR #6702, opened 2017-05-04
InternalThreadLocalMap#stringBuilder: ensure memory overhead

fengjiachun:
Motivation:

InternalThreadLocalMap#stringBuilder: ensure memory overhead

Modification:

If the capacity of StringBuilder is greater than 65536 then release it on the next time you get StringBuilder and re-create a StringBuilder. 

I'm not sure if this is a good idea.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/InternalThreadLocalMap.java
+++ b/common/src/main/java/io/netty/util/internal/InternalThreadLocalMap.java
@@ -35,6 +35,7 @@ import java.util.WeakHashMap;
  */
 public final class InternalThreadLocalMap extends UnpaddedInternalThreadLocalMap {
 
+    private static final int STRING_BUILDER_MAX_CAPACITY = 1024 << 6;
     private static final int DEFAULT_ARRAY_LIST_INITIAL_CAPACITY = 8;
 
     public static final Object UNSET = new Object();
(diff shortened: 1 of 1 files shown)


common/src/main/java/io/netty/util/internal/InternalThreadLocalMap.java
```diff
@@ -164,7 +165,7 @@ public int size() {
 
     public StringBuilder stringBuilder() {
         StringBuilder builder = stringBuilder;
-        if (builder == null) {
+        if (builder == null || builder.capacity() > STRING_BUILDER_MAX_CAPACITY /* ensure memory overhead */ ) {
```
jasontedor: I don't think you want to create a new `StringBuilder` but instead trim it to some size if it exceeds that size.

fengjiachun: you are right

normanmaurer:
@fengjiachun LGTM...

normanmaurer:
Cherry-picked into 4.1 (963cd22a05) and 4.0 (8715a7866c)

jasontedor:
Hmm...my suggestion wasn't incorporated here? 😄

normanmaurer:
@jasontedor doh missed it. Let me do a followup later today

jasontedor:
@normanmaurer I got you, see #6708.

normanmaurer:
@jasontedor thanks a lot buddy.

PR #6704, opened 2017-05-05
Make DnsNameResolverTest pass on Java7

normanmaurer:
Motivation:

IDN.toUnicode(...) removes trailing dots when used in Java7 while it not does on java8.

Modifications:

Check if we should test with the trailing dot removed or not.

Result:

Test pass on Java7 as well.


normanmaurer:
Cherry-picked into 4.1 (f65885fc54)

Scottmitch:
thanks @normanmaurer

PR #6707, opened 2017-05-05
Add explicit message when `noexec` prevents library loading.

mkw:
Add explicit message when `noexec` prevents library loading.

Motivation:

Docker's `--tmpfs` flag mounts the temp volume with `noexec` by default, resulting in an UnsatisfiedLinkError.  While this is good security practice, it is a surprising failure from a seemingly innocuous flag.

Modifications:

Add a best-effort attempt to detect the `noexec` flag in `NativeLibraryLoader`.  This solution is (probably) limited to Oracle/OpenJDK VM's and requires Java 7 or higher, and also requires `--add-opens=java.base/sun.nio.fs=...` for Java 9.  It avoids error message parsing and it does not directly attempt to read mount information.

Also upgraded animal-sniffer-maven-plugin and added configuration to make it ignore code annotated with `io.netty.util.SuppressForbidden` to address @Scottmitch's feedback from original PR.

Result:

Fixes [#6678].

### DIFF
--- a/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
+++ b/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
@@ -18,8 +18,8 @@ package io.netty.util.internal;
 import io.netty.util.internal.logging.InternalLogger;
 import io.netty.util.internal.logging.InternalLoggerFactory;
 
-import java.io.Closeable;
 import java.io.ByteArrayOutputStream;
+import java.io.Closeable;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
@@ -27,10 +27,14 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.lang.reflect.Method;
 import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.attribute.PosixFilePermission;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.Arrays;
+import java.util.EnumSet;
 import java.util.Locale;
+import java.util.Set;
 
 /**
  * Helper class to load JNI resources.
@@ -227,6 +231,20 @@ public final class NativeLibraryLoader {
             out = null;
 
             loadLibrary(loader, tmpFile.getPath(), true);
+        } catch (UnsatisfiedLinkError e) {
+            try {
+                if (tmpFile != null && tmpFile.isFile() && tmpFile.canRead() &&
+                    !NoexecVolumeDetector.canExecuteExecutable(tmpFile)) {
+                    logger.info("{} exists but cannot be executed even when execute permissions set; " +
+                                "check volume for \"noexec\" flag; use -Dio.netty.native.workdir=[path] " +
+                                "to set native working directory separately.",
+                                tmpFile.getPath());
+                }
+            } catch (Throwable t) {
+                logger.debug("Error checking if {} is on a file store mounted with noexec", tmpFile, t);
+            }
+            // Re-throw to fail the load
+            throw e;
         } catch (Exception e) {
             throw (UnsatisfiedLinkError) new UnsatisfiedLinkError(
                     "could not load a native library: " + name).initCause(e);
@@ -370,5 +388,43 @@ public final class NativeLibraryLoader {
 
     private NativeLibraryLoader() {
         // Utility
+    }
+
+    private static final class NoexecVolumeDetector {
+
+        private static boolean canExecuteExecutable(File file) throws IOException {
+            if (PlatformDependent.javaVersion() < 7) {
+                // Pre-JDK7, the Java API did not directly support POSIX permissions; instead of implementing a custom
+                // work-around, assume true, which disables the check.
+                return true;
+            }
+
+            // If we can already execute, there is nothing to do.
+            if (file.canExecute()) {
+                return true;
+            }
+
+            // On volumes, with noexec set, even files with the executable POSIX permissions will fail to execute.
+            // The File#canExecute() method honors this behavior, probaby via parsing the noexec flag when initializing
+            // the UnixFileStore, though the flag is not exposed via a public API.  To find out if library is being
+            // loaded off a volume with noexec, confirm or add executalbe permissions, then check File#canExecute().
+
+            Set<PosixFilePermission> existingFilePermissions = Files.getPosixFilePermissions(file.toPath());
+            Set<PosixFilePermission> executePermissions = EnumSet.of(PosixFilePermission.OWNER_EXECUTE,
+                                                                     PosixFilePermission.GROUP_EXECUTE,
+                                                                     PosixFilePermission.OTHERS_EXECUTE);
+            if (existingFilePermissions.containsAll(executePermissions)) {
+                return false;
+            }
+
+            Set<PosixFilePermission> newPermissions = EnumSet.copyOf(existingFilePermissions);
+            newPermissions.addAll(executePermissions);
+            Files.setPosixFilePermissions(file.toPath(), newPermissions);
+            return file.canExecute();
+        }
+
+        private NoexecVolumeDetector() {
+            // Utility
+        }
     }
 }
--- a/pom.xml
+++ b/pom.xml
@@ -709,6 +709,13 @@
 
             <!-- Resolver -->
             <ignore>java.net.InetAddress</ignore>
+
+            <!-- NoexecVolumeDetector -->
+            <ignore>java.nio.file.attribute.PosixFilePermission</ignore>
+            <ignore>java.nio.file.Files</ignore>
+            <ignore>java.nio.file.LinkOption</ignore>
+            <ignore>java.nio.file.Path</ignore>
+            <ignore>java.io.File</ignore>
           </ignores>
         </configuration>
         <executions>


mkw:
I accidentally closed #6685 cleaning up branches in our clone.  This is a second attempt and addresses the comments that @Scottmitch had on that PR, being:

- Instead of using reflection for Java7 API's, it checks the java version in a guard clause and directly uses Files/FileStore/Path.  This requires either ignoring _all_ references to these fairly generic NIO APIs in Java7 or upgrading sniffer and using annotations for excluding the method.  I chose the latter since my first shot at the former did not work, and it was not clear why.
- It documents the InaccessibleObjectException expected when Jigsaw blocks access to the private API's used to extract mount information.
- It removes a redundant throw/catch when checking for noexec.

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -368,4 +385,59 @@ private static void closeQuietly(Closeable c) {
...
+                } catch (Throwable t) {
+                    logger.debug("Failed to initialize NativeLibraryLoader.NoexecFileStoreDetector", t);
+                }
+            }
+            MOUNT_ENTRY_METHOD = mountEntryMethod;
+            HAS_OPTION_METHOD = hasOptionMethod;
+        }
+
+        @SuppressForbidden(reason = "to obtain mount information about file stores")
+        private static boolean isOnNoexecFileStore(File file)
```
normanmaurer: remove "private" to make it package-private

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -368,4 +385,59 @@ private static void closeQuietly(Closeable c) {
...
+                throws IOException, InvocationTargetException, IllegalAccessException {
+            if (PlatformDependent.javaVersion() < 7 || HAS_OPTION_METHOD == null || MOUNT_ENTRY_METHOD == null) {
```
normanmaurer: you can remove `PlatformDependent.javaVersion() < 7` as these methods will be null in this case anyway.

pom.xml
```diff
@@ -706,7 +706,7 @@
              be used even when compiling with java 1.7+ -->
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>animal-sniffer-maven-plugin</artifactId>
-        <version>1.11</version>
+        <version>1.15</version>
```
normanmaurer: why this change ?

pom.xml
```diff
@@ -721,6 +721,9 @@
             <artifactId>java16</artifactId>
             <version>1.1</version>
           </signature>
+          <annotations>
+            <annotation>io.netty.util.SuppressForbidden</annotation>
```
normanmaurer: Why this change ?

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -368,4 +385,59 @@ private static void closeQuietly(Closeable c) {
...
+                return false;
+            }
+            FileStore fileStore = Files.getFileStore(file.toPath());
+            Object unixMountEntry = MOUNT_ENTRY_METHOD.invoke(fileStore);
+            return (Boolean) HAS_OPTION_METHOD.invoke(unixMountEntry, "noexec");
```
Scottmitch: I guess there is no way to detect an issue by avoiding internal APIs?
- Checking if the directory is readable and/or [File#canExecute](https://docs.oracle.com/javase/7/docs/api/java/io/File.html#canExecute())
- Use [FileStore#getAttribute](https://docs.oracle.com/javase/7/docs/api/java/nio/file/FileStore.html#getAttribute(java.lang.String)) or some other attribute API offered by NIO

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -15,18 +15,22 @@
...
+import java.io.Closeable;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.net.URL;
+import java.nio.file.FileStore;
```
normanmaurer: @johnou good point!

@mkw we need to use reflection here or move it to another class that is only loaded if java version >= 7

pom.xml
mkw: My mistake -- undoing -- thought it was needed for annotation-based suppression (but I'm also about to undo that).

pom.xml
mkw: Idea is to more finely target suppression of the animal sniffer errors, but I agree that reflection is better; undoing and fixing.

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
mkw: I didn't realize it, but it appears that [`File#canExecute`](https://docs.oracle.com/javase/7/docs/api/java/io/File.html#canExecute()) _can_ be used for this: it does require that I temporarily try to update the POSIX permissions, but that is far preferable to the internal API mess that I'd created.  Expect an updated PR soon.

mkw:
Sorry for the long delay on my end, as well.  The revised version uses only public Java API's and reflection to avoid problems with JDK6.

Previously, @Scottmitch had asked to avoid reflection, but so far as I can tell I can only suppress the warnings from Animal Sniffer by suppressing warnings about _all_ uses of classes like `java.nio.file.Files`, which seems overly broad.  Note that the Animal Sniffer Maven plugin does support a custom "suppress only here" annotation, but it is not configured in Netty.  Were it to be configured, I suspect that the class list in the main pom.xml should be mostly dropped and replaced with annotations across the project.

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -371,4 +390,62 @@ private static void closeQuietly(Closeable c) {
...
         // Utility
     }
+
+    private static final class NoexecVolumeDetector {
+
+        private static boolean canExecuteExecutable(File file)
+                throws ClassNotFoundException, NoSuchMethodException, InvocationTargetException,
+                       IllegalAccessException {
+            if (PlatformDependent.javaVersion() < 7) {
+                // Assume true because we
```
Scottmitch: incomplete comment ... please complete this comment or remove it.

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -371,4 +390,62 @@ private static void closeQuietly(Closeable c) {
...
+            // If we can already execute, there is nothing to do.
+            if (file.canExecute()) {
+                return true;
+            }
+
+            // On volumes, with noexec set, even files with the executable POSIX permissions will fail to execute.
+            // So, check for executable permissions, then find out if the file can be executed.
+
+            // Everything is done reflectively to prevent the animal-sniffer plugin from complainin.
+            Class<?> filesClass = Class.forName("java.nio.file.Files");
```
Scottmitch: instead of using reflection to load these resources can we reference the class directly? We just need to be careful about class loading on older JDKs.

```java
private static boolean canExecuteFile(File f) {
  return PlatformDependent.javaVersion() < 7 ? true : Java7FilesUtil. canExecuteFile(f);
}

// In another class file ... Java7FilesUtil.java
final class Java7FilesUtil {
  static boolean canExecuteFile(File f) {
    ...
  }
}
```

mkw: First, if I'm reading, `io.netty.util.internal.PlatformDependent#newConcurrentDeque` correctly, it _appears_ that it is fine to use Java7 classes provided that they are protected by a version check -- no?

Whether or not this is true, I would be happy to do this non-reflectively, but doing so would require adding the classes to the Animal Sniffer configuration.  My understanding of how Animal Sniffer works is that doing so would suppress the usages of these classes globally within the Netty codebase.  Given how natural it has been for developers to use `java.nio.file.*`, this seems unwise.

As of [Animal Sniffer 1.11](http://blog.soebes.de/blog/2014/03/25/animal-sniffer-version-1-dot-11-released/), it is possible to configure an annotation to suppress warnings in specific classes or method, but so far as I can tell, this is not yet configured for the project.  If you want, I could make a separate issue that extended `io.netty.util.SuppressForbidden` or added something similar (`io.netty.util.IgnoreJDK7` and `io.netty.util.IgnoreJDK8`?), added these to relevant scopes, and removed the global class exclusions from the sniffer configuration.

Please advise.

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
mkw: Sorry -- was rushing yesterday and didn't re-read my comments.  They should be much clearer now.

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
Scottmitch: > it appears that it is fine to use Java7 classes provided that they are protected by a version check -- no?

No. If you need to use features not available in JDK6 you need a version check. Netty's minimum compatibility is currently at JDK6 so it must run against this JDK and we can't attempt to load any classes which won't exist in this context.

> My understanding of how Animal Sniffer works is that doing so would suppress the usages of these classes globally within the Netty codebase ... If you want, I could make a separate issue

I would prefer to have the code reference actual types and update the Animal sniffer configuration to provide more fine grained exceptions. For the purposes of this PR it is fine to add an exception using the current approach and we can update the animal sniffer exceptions in a followup PR.

normanmaurer: The important here is that you "MUST" import the java7 classes in another class that you only access if the version check match. Otherwise you will get a classloader exception when try to use java6

mkw: I'm going to apply this rule, but this block, inside`io.netty.util.internal.PlatformDependent` confuses me:
```
    /**
     * Returns a new concurrent {@link Deque}.
     */
    public static <C> Deque<C> newConcurrentDeque() {
        if (javaVersion() < 7) {
            return new LinkedBlockingDeque<C>();
        } else {
            return new ConcurrentLinkedDeque<C>();
        }
    }
```
The Javadoc of `java.util.concurrent.ConcurrentLinkedDeque` includes `@since 1.7`, and it is imported on line 42.

Assuming the rule is "you 'MUST' import the java7 classes in another class that you only access if the version check match," shouldn't PlatformDependant fail to load entirely on JDK6?  Or am I misunderstanding something?

Scottmitch: You found a bug (/cc @normanmaurer) 👍 

This should be:

```java
return new java.util.concurrent.ConcurrentLinkedDeque<C>();
```

normanmaurer: will look into it

normanmaurer: Fixed in https://github.com/netty/netty/pull/6822

mkw: @Scottmitch The non-reflective version is in, though it requires suppressing checks for uses of all of:

* java.nio.file.attribute.PosixFilePermission
* java.nio.file.Files
* java.nio.file.LinkOption
* java.nio.file.Path
* java.io.File (so it can call `File#toPath()`)

Note that the research in [this repository](https://github.com/mkw/java6test/tree/master) seems to have convinced @normanmaurer that the Java7 check at the beginning of the method should work (see the comments on #6822) in Java6 with the imports present.

Scottmitch:
4.1 (ed5fcbb77) 4.0 (aff9e7ac0)

Scottmitch:
thanks @mkw !

PR #6708, opened 2017-05-05
Trim thread local string builder if large

jasontedor:
Motivation:

A previous change allocated a new thread local string builder if it was getting too large. This is a good change, these string builders can accidentally get too large and then never shrunk and that is sort of a memory leak. However, the change allocates an entirely new string builder which is more allocations than necessary. Instead, we can trim the string builder if its too large, this only allocates an extra backing array instead of a whole new object.

Modifications:

If the string builder is above a threshold, we trim the string builder and then ensure its capacity is reasonable to we do not allocate too much as we start using the string builder.

Result:

The thread local string builder do not serve as a memory yet we do not allocate too many new objects.

Relates #6702

### DIFF
--- a/common/src/main/java/io/netty/util/internal/InternalThreadLocalMap.java
+++ b/common/src/main/java/io/netty/util/internal/InternalThreadLocalMap.java
@@ -35,7 +35,6 @@ import java.util.WeakHashMap;
  */
 public final class InternalThreadLocalMap extends UnpaddedInternalThreadLocalMap {
 
-    private static final int STRING_BUILDER_MAX_CAPACITY = 1024 << 6;
     private static final int DEFAULT_ARRAY_LIST_INITIAL_CAPACITY = 8;
 
     public static final Object UNSET = new Object();
@@ -164,13 +163,17 @@ public final class InternalThreadLocalMap extends UnpaddedInternalThreadLocalMap
     }
 
     public StringBuilder stringBuilder() {
-        StringBuilder builder = stringBuilder;
-        if (builder == null || builder.capacity() > STRING_BUILDER_MAX_CAPACITY) {
-            stringBuilder = builder = new StringBuilder(512);
+        final int stringBuilderCapacity = 1024;
+        if (stringBuilder == null) {
+            stringBuilder = new StringBuilder(stringBuilderCapacity);
         } else {
-            builder.setLength(0);
+            if (stringBuilder.capacity() > stringBuilderCapacity) {
+                stringBuilder.setLength(stringBuilderCapacity);
+                stringBuilder.trimToSize();
+            }
+            stringBuilder.setLength(0);
         }
-        return builder;
+        return stringBuilder;
     }
 
     public Map<Charset, CharsetEncoder> charsetEncoderCache() {


common/src/main/java/io/netty/util/internal/InternalThreadLocalMap.java
```diff
@@ -164,13 +164,17 @@ public int size() {
     }
 
     public StringBuilder stringBuilder() {
-        StringBuilder builder = stringBuilder;
-        if (builder == null || builder.capacity() > STRING_BUILDER_MAX_CAPACITY) {
```
fenik17: Constant can be deleted now.

jasontedor: Thanks, I pushed.

common/src/main/java/io/netty/util/internal/InternalThreadLocalMap.java
```diff
@@ -164,13 +163,17 @@ public int size() {
...
-            stringBuilder = builder = new StringBuilder(512);
+        final StringBuilder builder = stringBuilder;
+        if (builder == null) {
+            return stringBuilder = new StringBuilder(512);
```
netkins: MAJOR Extract the assignment out of this expression. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AAssignmentInSubExpressionCheck)

common/src/main/java/io/netty/util/internal/InternalThreadLocalMap.java
fengjiachun: Is that 2 times of memory copy?

fengjiachun:
@jasontedor Is that 2 times of memory copy?

common/src/main/java/io/netty/util/internal/InternalThreadLocalMap.java
fingthinking: How about use code like this at the 170 line:
```java
if(builder.capacity() > 512){
            builder.setLength(512);
            builder.trimToSize();
        }
        builder.setLength(0);
```
with this only one time array copy 512 length.

jasontedor: @fingthinking That's cleaner, thanks.

normanmaurer: @jasontedor I wonder if we should allow to adjust the max size via a system property ?

jasontedor: @normanmaurer That doesn't seem necessary to me but if you think it is I would be happy to add it.

normanmaurer: maybe just upper this to 1024 (as it is per thread anyway) and then its good enough for now. we can add a property later if we need.

@jasontedor WDYT ?

jasontedor: Sure, that sounds good to me. I'll push later.

normanmaurer: @jasontedor thanks... ping me once done.

jasontedor: @normanmaurer I pushed.

normanmaurer:
@jasontedor thanks buddy!

Cherry-picked into 4.1 (d88cd23bfc3d1cc2422ca73d63b281990bbcb8a5) and 4.0 (b30409890c372e916946514e0352fef8ce82529e)

jasontedor:
Thanks @normanmaurer.

Issue #6709, opened 2017-05-06
SslHandler does not work with little endian buffers


sudheeshkatkam:
### Expected behavior
SslHandler to work regardless of LE or BE buffers
### Actual behavior
SslHandler uses [SslUtils#getEncryptedPacketLength](https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/SslUtils.java#L123) which does not work with little endian buffers.

Looks like the issue was fixed in [16750ff](https://github.com/netty/netty/commit/16750ffc2802e6098e624df56ea6fbbe5c1055c3) but the change was not carried forward. I suspect there is an alternative then?

From the doc in the fix:
> Reads a big-endian short integer from the buffer.  Please note that we do not use
> {@link ChannelBuffer#getShort(int)} because it might be a little-endian buffer.

### Minimal yet complete reproducer code (or URL to code)
Code to reproduce the issue is here: [8b415c1](https://github.com/sudheeshkatkam/netty/commit/8b415c1949e0320b4fa1b7a2c7430c1af4eabd52). This bootstraps the chat server and client with a little endian bytebuf allocator, and adds details to console output.

To reproduce the failure (handshake timeout), start server using this command (in root directory):
> ./run-example.sh -Dlebba=true securechat-server

And then the client, using this command:

> ./run-example.sh -Dlebba=true securechat-client

To fix the use, apply a modified version of [16750ff](https://github.com/netty/netty/commit/16750ffc2802e6098e624df56ea6fbbe5c1055c3) which is here: [f7da006](https://github.com/sudheeshkatkam/netty/commit/f7da006152eaaee40b3482b7126e2f1f2bd90b3d).
### Netty version
4.1
### JVM version (e.g. `java -version`)
java version "1.8.0_77"
Java(TM) SE Runtime Environment (build 1.8.0_77-b03)
Java HotSpot(TM) 64-Bit Server VM (build 25.77-b03, mixed mode)

normanmaurer:
@sudheeshkatkam will check... thanks!

normanmaurer:
@sudheeshkatkam PTAL https://github.com/netty/netty/pull/6713

sudheeshkatkam:
LGTM!

normanmaurer:
Fixed by https://github.com/netty/netty/pull/6713

PR #6713, opened 2017-05-08
Ensure SslUtils and so SslHandler works when using with Little-Endian…

normanmaurer:
… buffers.

Motivation:

We not correctly handle LE buffers when try to read the packet length out of the buffer and just assume it always is a BE buffer.

Modifications:

Correctly account for the endianess of the buffer when reading the packet lenght.

Result:

Fixes [#6709].

### DIFF
--- a/handler/src/test/java/io/netty/handler/ssl/SslUtilsTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/SslUtilsTest.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.ssl;
+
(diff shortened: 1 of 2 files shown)


handler/src/test/java/io/netty/handler/ssl/SslUtilsTest.java
```diff
@@ -0,0 +1,67 @@
...
+        cTOsLE.flip();
+
+        assertTrue(engineBE.wrap(empty, cTOsBE).bytesProduced() > 0);
+        cTOsBE.flip();
+
+        ByteBuf bufferLE = Unpooled.buffer().order(ByteOrder.LITTLE_ENDIAN).writeBytes(cTOsLE);
+        ByteBuf bufferBE = Unpooled.buffer().writeBytes(cTOsBE);
+
+        // Test that the packet-length for BE and LE is the same
+        assertEquals(SslUtils.getEncryptedPacketLength(bufferBE, 0),
```
Scottmitch: nit: add static import for `SslUtils.getEncryptedPacketLength`

Scottmitch:
@sudheeshkatkam - can you verify that you have run these tests on your LE machine?

normanmaurer:
@Scottmitch I think this has actually nothing to do with running on LE or not (at least not 100%). It has to do which buffers you feed into the `SslHandler` (which is the test ;))

sudheeshkatkam:
I had tested to make sure the reproducer code (mentioned in https://github.com/netty/netty/issues/6709) works.

normanmaurer:
@Scottmitch ready to merge ?

handler/src/test/java/io/netty/handler/ssl/SslUtilsTest.java
normanmaurer: done...

normanmaurer:
Cherry-picked into 4.1 (732948b3c4) and 4.0 (eeb8c221e0)

Issue #2973, opened 2014-10-07
WriteTimeoutHandler missed cleanup


Scottmitch:
It seems like the `WriteTimeoutHandler` does not always cancel the timeout task it schedules in all the same cases that the and IdleStateHandler does.  For example the `handlerRemoved` method is not overridden in `WrtieTimeoutHandler`.  There may be other cases where the timeout is not canceled when it should be and further investigation is needed.

normanmaurer:
Fixed

PR #6715, opened 2017-05-08
OpenSslEngine wrap may generate bad data if multiple src buffers

Scottmitch:
Motivation:
SSL_write requires a fixed amount of bytes for overhead related to the encryption process for each call. OpenSslEngine#wrap(..) will attempt to encrypt multiple input buffers until MAX_PLAINTEXT_LENGTH are consumed, but the size estimation provided by calculateOutNetBufSize may not leave enough room for each call to SSL_write. If SSL_write is not able to completely write results to the destination buffer it will keep state and attempt to write it later. Netty doesn't account for SSL_write keeping state and assumes all writes will complete synchronously (by attempting to allocate enough space to account for the overhead) and feeds the same data to SSL_write again later which results in corrupted data being generated.

Modifications:
- OpenSslEngine#wrap should only produce a single TLS packet according to the SSLEngine API specificaiton [1].
[1] https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLEngine.html#wrap-java.nio.ByteBuffer:A-int-int-java.nio.ByteBuffer-
- OpenSslEngine#wrap should only consider a single buffer when determining if there is enough space to write, because only a single buffer will ever be consumed.

Result:
OpenSslEngine#wrap will no longer produce corrupted data due to incorrect accounting of space required in the destination buffers.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -610,9 +610,9 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
                     }
                 }
 
-                if (dst.remaining() < calculateOutNetBufSize(srcsLen, endOffset - offset)) {
-                    // Can not hold the maximum packet so we need to tell the caller to use a bigger destination
-                    // buffer.
+                // we will only produce a single TLS packet, and we don't aggregate src buffers,
+                // so we always fix the number of buffers to 1 when checking if the dst buffer is large enough.
+                if (dst.remaining() < calculateOutNetBufSize(srcsLen, 1)) {
                     return new SSLEngineResult(BUFFER_OVERFLOW, getHandshakeStatus(), 0, 0);
                 }
 
@@ -638,9 +638,7 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
                         bytesProduced += bioLengthBefore - pendingNow;
                         bioLengthBefore = pendingNow;
 
-                        if (bytesConsumed == MAX_PLAINTEXT_LENGTH || bytesProduced == dst.remaining()) {
-                            return newResultMayFinishHandshake(status, bytesConsumed, bytesProduced);
-                        }
+                        return newResultMayFinishHandshake(status, bytesConsumed, bytesProduced);
                     } else {
                         int sslError = SSL.getError(ssl, bytesWritten);
                         if (sslError == SSL.SSL_ERROR_ZERO_RETURN) {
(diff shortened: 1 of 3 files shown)


handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -638,9 +638,7 @@ public final SSLEngineResult wrap(
                         bytesProduced += bioLengthBefore - pendingNow;
                         bioLengthBefore = pendingNow;
 
-                        if (bytesConsumed == MAX_PLAINTEXT_LENGTH || bytesProduced == dst.remaining()) {
-                            return newResultMayFinishHandshake(status, bytesConsumed, bytesProduced);
-                        }
```
normanmaurer: @Scottmitch why is this not needed anymore ?

Scottmitch: This is explained in the commit message:

> OpenSslEngine#wrap should only produce a single TLS packet according to the SSLEngine API specificaiton [1].
[1] https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLEngine.html#wrap-java.nio.ByteBuffer:A-int-int-java.nio.ByteBuffer-

Scottmitch: If we consume multiple packets we must be more careful about how much overhead we account for. This is done in PR https://github.com/netty/netty/pull/6365

normanmaurer: Got it... Sorry I missed this before.

normanmaurer:
@Scottmitch LGTM... please merge once the CI passes

Scottmitch:
3 minutes isn't enough on the CI servers ... bumping to 5 minutes.

Scottmitch:
Note that performance may be impacted as a result of this PR (due to strictly limiting the amount of data encrypted per wrap call), but PR https://github.com/netty/netty/pull/6365 will improve upon the ability to wrap (and unwrap) multiple TLS packets per call.

Scottmitch:
4.1 (141089998) 4.0 (aef6dbc08)

lukaszx0:
grpc-java was failing  when upgraded to `4.1.10.Final` with `netty-tcnative 2.0.1.Final` (see #2973). I just tried the latest netty snapshots and it is passing now - thanks @Scottmitch !

For posterity, if you're app is failing with weird `javax.net.ssl.SSLException: Unsupported record version Unknown-126.182` ([whole stacktrace](https://gist.github.com/lukaszx0/8e144e01a68b468d78ae4dc0d2d68f60)) upgrade to `>4.1.11.Final` should fix it (as it'll contain this PR)

normanmaurer:
@lukaszx0 thanks for reporting back.

Scottmitch:
@lukaszx0 - Thanks for confirming. We are aware of the implications of this issue and are in the process of generating a new release now. Standby.

Issue #6716, opened 2017-05-08
OpenSsl.java line 143 creates file and doesn't delete it


rkapsi:
The OpenSsl class at line [143](https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/OpenSsl.java#L143) creates a SelfSignedCertificate but doesn't delete it once it's done with it.

### Expected behavior

File getting deleted once it's done with it

### Actual behavior

File may or may not getting deleted depending on if `File#deleteOnExit()` gets triggered.

normanmaurer:
@rkapsi good catch. Let me fix this very quickly.

normanmaurer:
@rkapsi fixed by https://github.com/netty/netty/pull/6718

normanmaurer:
Fixed by https://github.com/netty/netty/pull/6718

PR #6718, opened 2017-05-08
Correctly delete SelfSignedCertificate once done with it.

normanmaurer:
Motivation:

In OpenSsl init code we create a SelfSignedCertificate which we not explicitly delete. This can lead to have the deletion delayed.

Modifications:

Delete the SelfSignedCertificate once done with it.

Result:

Fixes [#6716]

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
+++ b/handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
@@ -122,6 +122,7 @@ public final class OpenSsl {
                 final long sslCtx = SSLContext.make(SSL.SSL_PROTOCOL_ALL, SSL.SSL_MODE_SERVER);
                 long privateKeyBio = 0;
                 long certBio = 0;
+                SelfSignedCertificate cert = null;
                 try {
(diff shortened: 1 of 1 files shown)


normanmaurer:
@rkapsi PTAL

handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
```diff
@@ -157,6 +158,10 @@ public Boolean run() {
                             }
                         } catch (Throwable ignore) {
                             logger.debug("KeyManagerFactory not supported.");
+                        } finally {
+                            if (cert != null) {
+                                cert.delete();
```
rkapsi: Looks good if it's safe to delete the file before free'ing the pointers in the outer `try-finally`.

normanmaurer: @rkapsi yes its safe as we "copy" to get the pointers

normanmaurer:
Cherry-picked into 4.1 (ec935c5a7b91c77a0e0e55f54978fb0cf8338dbe) and 4.0 (881e05aba56717b437d994b4c481647c18f9ee1f)

PR #6719, opened 2017-05-08
The SNI extension value is ASCII encoded but Netty uses UTF-8

### LINKED ISSUE #6717: The value of the SNI extension is ASCII encoded but Netty uses UTF-8
The value of the SNI extension is ASCII encoded but Netty's SNI code tries to decode the `byte[]` as UTF-8.

The was flagged by @fenik17 in this https://github.com/netty/netty/pull/6623#discussion_r111070719

https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/AbstractSniHandler.java#L188

https://tools.ietf.org/html/rfc6066#page-6

> "HostName" contains the fully qualified DNS hostname of the server,
   as understood by the client.  The hostname is represented as a byte
   string using ASCII encoding without a trailing dot.  This allows the
   support of internationalized domain names through the use of A-labels
   defined in [RFC5890].  DNS hostnames are case-insensitive.  The
   algorithm to compare hostnames is described in [RFC5890], Section
   2.3.2.4.

### Expected behavior

Do as the RFC says?

### Actual behavior

Not doing as the RFC says!

rkapsi:
Motivation

RFC 6066 (https://tools.ietf.org/html/rfc6066#page-6) says that the hostname in the SNI extension is ASCII encoded but Netty decodes it using UTF-8.

Modifications

Use ASCII instead of UTF-8

Result

Fixes #6717

### DIFF
--- a/handler/src/test/java/io/netty/handler/ssl/SniHandlerTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/SniHandlerTest.java
@@ -160,38 +160,71 @@ public class SniHandlerTest {
             SniHandler handler = new SniHandler(mapping);
             EmbeddedChannel ch = new EmbeddedChannel(handler);
 
-            // hex dump of a client hello packet, which contains hostname "CHAT4。LEANCLOUD。CN"
-            String tlsHandshakeMessageHex1 = "16030100";
-            // part 2
-            String tlsHandshakeMessageHex = "bd010000b90303a74225676d1814ba57faff3b366" +
-                    "3656ed05ee9dbb2a4dbb1bb1c32d2ea5fc39e0000000100008c0000001700150000164348" +
-                    "415434E380824C45414E434C4F5544E38082434E000b000403000102000a00340032000e0" +
-                    "00d0019000b000c00180009000a0016001700080006000700140015000400050012001300" +
-                    "0100020003000f0010001100230000000d0020001e0601060206030501050205030401040" +
-                    "20403030103020303020102020203000f00010133740000";
-
             try {
-                // Push the handshake message.
-                // Decode should fail because SNI error
+                // hex dump of a client hello packet, which contains hostname "CHAT4.LEANCLOUD.CN"
+                String tlsHandshakeMessageHex1 = "16030100";
+                // part 2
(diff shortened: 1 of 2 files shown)


rkapsi:
@normanmaurer Interesting, there is an unit test that's in violation of the RFC.

https://github.com/netty/netty/blob/4.1/handler/src/test/java/io/netty/handler/ssl/SniHandlerTest.java#L163

rkapsi:
It appears [DomainNameMapping](https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/DomainNameMapping.java#L128) is doing some unnecessary ASCII to ASCII encoding.

handler/src/test/java/io/netty/handler/ssl/SniHandlerTest.java
```diff
@@ -198,6 +192,45 @@ public void testServerNameParsing() throws Exception {
...
+                    "0100020003000f0010001100230000000d0020001e0601060206030501050205030401040" +
+                    "20403030103020303020102020203000f00010133740000";
+
+            // Push the handshake message.
+            // Decode should fail because of the badly encoded "HostName" string in the SNI extension
+            // that isn't ASCII as per RFC 6066 - https://tools.ietf.org/html/rfc6066#page-6
+            ch.writeInbound(Unpooled.wrappedBuffer(DatatypeConverter.parseHexBinary(tlsHandshakeMessageHex1)));
+            ch.writeInbound(Unpooled.wrappedBuffer(DatatypeConverter.parseHexBinary(tlsHandshakeMessageHex)));
+        } finally {
+            releaseAll(leanContext, leanContext2, nettyContext);
```
Scottmitch: please also call `ch.finishAndReleaseAll();` if `ch != null`

normanmaurer:
@rkapsi please do as suggest and rebase. Then let us know

rkapsi:
@Scottmitch @normanmaurer - done. Applied `finhsgAndReleaseAll()` also to `testServerNameParsing()` (was a custom for-loop).

Scottmitch:
4.1 (2db4f2557)

Thanks @rkapsi !

PR #6723, opened 2017-05-09
Correctly detect if Ocsp is supported

normanmaurer:
Motivation:

We only used the openssl version to detect if Ocsp is supported or not which is not good enough as even the version is correct it may be compiled without support for OCSP (like for example on ubuntu).

Modifications:

Try to enable OCSP while static init OpenSsl and based on if this works return true or false when calling OpenSsl.isOcspSupported().

Result:

Correctly detect if OSCP is supported.


normanmaurer:
@rkapsi FYI

rkapsi:
Oooh, thx.

normanmaurer:
Cherry-picked into 4.1 (c053c5144d412535a0efb79fa5da4b7346138dee) and 4.0 (71af54ce72)

PR #6727, opened 2017-05-09
Simplify JUnit assertions

fenik17:
Motivation:

Some JUnit assert calls can be replaced by simpler.

Modifications:

Replacement with a more suitable methods.

Result:

More informative JUnit reports.


normanmaurer:
Cherry-picked into 4.1 (94e9448ae3).

@fenik17 thanks <3

PR #6728, opened 2017-05-09
ByteToMessageDecoder#handlerRemoved may release cumulation buffer pre…

Scottmitch:
…maturely

Motivation:
ByteToMessageDecoder#handlerRemoved will immediately release the cumulation buffer, but it is possible that a child class may still be using this buffer, and therefore use a dereferenced buffer.

Modifications:
- ByteToMessageDecoder#handlerRemoved and ByteToMessageDecoder#decode should coordinate to avoid the case where a child class is using the cumulation buffer but ByteToMessageDecoder releases that buffer.

Result:
Child classes of ByteToMessageDecoder are less likely to reference a released buffer.

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
@@ -129,11 +129,24 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
         }
     };
 
+    private static final byte STATE_INIT = 0;
+    private static final byte STATE_CALLING_CHILD_DECODE = 1;
+    private static final byte STATE_HANDLER_REMOVED_PENDING = 2;
+
     ByteBuf cumulation;
     private Cumulator cumulator = MERGE_CUMULATOR;
     private boolean singleDecode;
     private boolean decodeWasNull;
     private boolean first;
+    /**
+     * A bitmask where the bits are defined as
+     * <ul>
+     *     <li>{@link #STATE_INIT}</li>
+     *     <li>{@link #STATE_CALLING_CHILD_DECODE}</li>
+     *     <li>{@link #STATE_HANDLER_REMOVED_PENDING}</li>
+     * </ul>
+     */
+    private byte decodeState = STATE_INIT;
     private int discardAfterReads = 16;
     private int numReads;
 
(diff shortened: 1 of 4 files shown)


codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
```diff
@@ -109,25 +102,59 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) t
...
-        ByteBuf expected = Unpooled.wrappedBuffer(new byte[] {'b'});
-        ByteBuf b = channel.readInbound();
-        assertEquals(expected, b);
-        assertNull(channel.readInbound());
-        expected.release();
-        b.release();
+    }
+
+    @Test
+    public void handlerRemovedWillNotReleaseBufferIfDecodeInProgress() {
```
normanmaurer: can you also have this test in `ReplayingDecoderTest` ?

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -134,6 +134,8 @@ public ByteBuf cumulate(ByteBufAllocator alloc, ByteBuf cumulation, ByteBuf in)
     private boolean singleDecode;
     private boolean decodeWasNull;
     private boolean first;
+    private boolean callingChildDecode;
+    private boolean handlerRemovedPending;
```
Scottmitch: 2 booleans was a bit simpler but I will just use a `byte` if we want to be more definitive about memory usage.

normanmaurer: was not about memory but more about "logic" as it can only be in one state of both. But this is just a nit. So feel free to ignore

Scottmitch: done ...  I changed to a `byte` anyways :)

codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
Scottmitch: good catch ... done

Scottmitch:
4.1 (ce2ce9d7a) 4.0 (23fc5d7af)

PR #6733, opened 2017-05-11
Correctly include all modules during build

normanmaurer:
Motivation:

To ensure the release plugin works correctly we need to ensure all modules are included during build.

Modification:

- Include all modules
- Skip compilation and tests for native code when not supported but still include the module and build the jar

Result:

Build and release works again


all/pom.xml
```diff
@@ -35,12 +35,117 @@
...
+        </repository>
+      </repositories>
+
+      <dependencies>
+        <!-- Depend on all our native jars -->
+        <dependency>
+          <groupId>${project.groupId}</groupId>
+          <artifactId>netty-transport-native-epoll</artifactId>
+          <version>${project.version}</version>
+          <classifier>linux-x86_64</classifier>
```
Scottmitch: I know these are the platforms we typically release from but should we use `${jni.classifier}` here in case we change that, or someone else wants to generate an artifact on some other system?

all/pom.xml
```diff
@@ -35,12 +35,117 @@
...
+          <name>openbsd</name>
+        </os>
       </activation>
+      <dependencies>
+        <!-- All release modules -->
+        <dependency>
+          <groupId>${project.groupId}</groupId>
+          <artifactId>netty-transport-native-kqueue</artifactId>
+          <version>${project.version}</version>
+          <classifier>osx-x86_64</classifier>
```
Scottmitch: is this correct? Just use `${jni.classifier}`?

all/pom.xml
```diff
@@ -340,14 +445,6 @@
       <scope>compile</scope>
       <optional>true</optional>
```
Scottmitch: why `epoll` and not `kqueue`? can you add a comment?

all/pom.xml
normanmaurer: sure we can do

all/pom.xml
normanmaurer: sure we can do

all/pom.xml
normanmaurer: actually I think we should remove epoll here as well. We should only pull this into the all if -Puber is used.

all/pom.xml
normanmaurer: actually no this will not work as this will be called during the release on mac or linux and we need to fetch both.

all/pom.xml
normanmaurer: actually no this will not work as this will be called during the release on mac or linux and we need to fetch both.

all/pom.xml
Scottmitch: all is always included by the top level pom ... so how does this work with the different native artifacts for snapshots?

all/pom.xml
Scottmitch: oOh interesting ... sgtm

all/pom.xml
Scottmitch: please add a comment

normanmaurer:
Cherry-picked into 4.1 (9e62c79574)

all/pom.xml
normanmaurer: discussed offline and added a comment for this

all/pom.xml
normanmaurer: done

PR #6734, opened 2017-05-12
Add uber-staging and uber-snapshot profile that can be used to genera…

normanmaurer:
…te uber all jars.

Motivation:

As we now include native code for multiple platforms we need to generate an uber all jar before release it from the staging repository. For this the uber-staging profile can be used. To create a snapshot uber jar the uber-snapshot profile can be used.

Modifications:

- Add uber-staging and uber-snapshot profile
- Correct comment in pom.xml file to show usage.

Result:

Easier to create snapshot and release uber jars.

### DIFF
--- a/all/pom.xml
+++ b/all/pom.xml
@@ -36,7 +36,7 @@
   <profiles>
     <!-- If the uber profile is used it will automatically fetch the missing native jar from maven and add it to the all jar as well. -->
     <profile>
-      <id>uber</id>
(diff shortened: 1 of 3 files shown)


normanmaurer:
Cherry-picked into 4.1 (827c4096567b0ef10bab192b1140acd1d1d15b87)

Scottmitch:
lgtm

Issue #6738, opened 2017-05-12
Missing netty-transport-native-* entries in the bom can lead to mismatched versions


wilkinsona:
### Expected behavior

Netty's bom provides dependency management for all of Netty's modules, thereby ensuring that a consistent version is used.

### Actual behavior

Netty's bom does not include dependency management for:

- `netty-transport-native-epoll`
- `netty-transport-native-kqueue`
- `netty-transport-native-unix-common`

This can lead to mismatched version of Netty's modules when using the bom to control their versions

### Steps to reproduce

1. Depend on something that uses Netty, including one of the `transport-native` modules listed above
2. Import Netty's bom to control the version of Netty that's used

### Minimal yet complete reproducer code (or URL to code)

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
	<modelVersion>4.0.0</modelVersion>

	<groupId>com.example</groupId>
	<artifactId>netty-bom-problem</artifactId>
	<version>0.0.1-SNAPSHOT</version>
	<packaging>jar</packaging>

	<name>netty-bom-problem</name>

	<dependencyManagement>
		<dependencies>
			<dependency>
				<groupId>io.netty</groupId>
				<artifactId>netty-bom</artifactId>
				<version>4.1.11.Final</version>
				<type>pom</type>
				<scope>import</scope>
			</dependency>
		</dependencies>
	</dependencyManagement>

	<dependencies>
		<dependency>
			<groupId>io.projectreactor.ipc</groupId>
			<artifactId>reactor-netty</artifactId>
			<version>0.6.3.RELEASE</version>
		</dependency>
	</dependencies>

</project>
```
 
This pom yields the following dependencies:

```
[INFO] The following files have been resolved:
[INFO]    io.netty:netty-common:jar:4.1.11.Final:compile
[INFO]    io.netty:netty-handler:jar:4.1.11.Final:compile
[INFO]    io.projectreactor:reactor-core:jar:3.0.7.RELEASE:compile
[INFO]    org.reactivestreams:reactive-streams:jar:1.0.0:compile
[INFO]    io.projectreactor.ipc:reactor-ipc:jar:0.6.2.RELEASE:compile
[...]

PR #6740, opened 2017-05-12
Correctly include native modules

normanmaurer:
Motivation:

We need to include the native modules on the bom to ensure there is no version missmatch.

Modifications:

Add native modules.

Result:

Fixes [#6738]

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -186,7 +186,23 @@
         <artifactId>netty-all</artifactId>
         <version>4.1.12.Final-SNAPSHOT</version>
       </dependency>
+      <dependency>
+        <groupId>io.netty</groupId>
+        <artifactId>netty-transport-unix-common</artifactId>
+        <version>4.1.12.Final-SNAPSHOT</version>
+      </dependency>
+      <dependency>
+        <groupId>io.netty</groupId>
+        <artifactId>netty-transport-native-epoll</artifactId>
+        <version>4.1.12.Final-SNAPSHOT</version>
+        <classifier>linux-x86_64</classifier>
+      </dependency>
+      <dependency>
+        <groupId>io.netty</groupId>
+        <artifactId>netty-transport-native-kqueue</artifactId>
+        <version>4.1.12.Final-SNAPSHOT</version>
+        <classifier>osx-x86_64</classifier>
+      </dependency>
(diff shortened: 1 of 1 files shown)


normanmaurer:
@wilkinsona PTAL

wilkinsona:
Looks good to me

bom/pom.xml
```diff
@@ -186,7 +186,28 @@
         <artifactId>netty-all</artifactId>
         <version>4.1.12.Final-SNAPSHOT</version>
       </dependency>
+      <dependency>
+        <groupId>io.netty</groupId>
+        <artifactId>netty-transport-unix-common</artifactId>
+        <version>4.1.12.Final-SNAPSHOT</version>
+      </dependency>
+      <dependency>
+        <groupId>${project.groupId}</groupId>
```
johnou: io.netty for both transport-native dependencies like all other modules?

fenik17:
Maybe make `version` as property in this pom? For changing only one line when a new version released.

johnou:
@fenik17 is that going to play well with release:update-versions? Changing all versions is automatic as part of the release process anyway.

fenik17:
@johnou yep, you are right.. Spring Boot [used properties only for the third-party libraries](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-dependencies/pom.xml).

bom/pom.xml
normanmaurer: +1

bom/pom.xml
```diff
@@ -186,7 +186,28 @@
...
+        <groupId>io.netty</groupId>
+        <artifactId>netty-transport-unix-common</artifactId>
+        <version>4.1.12.Final-SNAPSHOT</version>
+      </dependency>
+      <dependency>
+        <groupId>io.netty</groupId>
+        <artifactId>netty-transport-native-epoll</artifactId>
+        <version>4.1.12.Final-SNAPSHOT</version>
+        <classifier>linux-x86_64</classifier>
+        <scope>compile</scope>
```
johnou: @normanmaurer why define scope and optional tags?

normanmaurer:
Cherry-picked into 4.1 (8811041cc6) and 4.0 (ffdb5e59f1)

PR #6744, opened 2017-05-14
Optionally throw exceptions on invalid cookies decoding

### LINKED ISSUE #6730: Cookie decoder stopped throwing exception on malformed cookie name
We want to switch from 4.0 to 4.1, but with 4.1 cookie decoding changed and on illegal characters cookie decoding procedure doesn't throw exception anymore. The piece of code that was throwing exception was in `io.netty.handler.codec.http.cookie.DefaultCookie#DefaultCookie`.

### Expected behavior
When a cookie contains illegal characters in name/value cookie decoder could, based on some parameter, throw an exception, so the server could handle it.
### Actual behavior
There are two modes in CookieDecoder: LAX or STRICT. LAX allowes illegal characters, STRICT swallows malformed cookie. In both cases there is no feedback.
### Steps to reproduce
Tests in `io.netty.handler.codec.http.cookie.ServerCookieDecoderTest` are enough to highlight this situation, specially `io.netty.handler.codec.http.cookie.ServerCookieDecoderTest#testRejectCookieValueWithSemicolon`
### Netty version
Netty 4.1.11.Final

alobodzki:
Motivation:

To provide quick feedback to developers netty based server could fail
requests fast on malformed cookies.

Modifications:

Additional flag on a decoder that would turn on/off throwing exceptions
on cookie validation error.

Result:

Fixes #6730
Developers will have choice whether they want to have a validation that
is swallowing malformed cookies, or get an exception to let the server
handle it.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDecoder.java
@@ -32,27 +32,39 @@ public abstract class CookieDecoder {
     private final InternalLogger logger = InternalLoggerFactory.getInstance(getClass());
 
     private final boolean strict;
+    private final CookieErrorHandler cookieErrorHandler;
 
     protected CookieDecoder(boolean strict) {
         this.strict = strict;
+        cookieErrorHandler = new CookieErrorHandler() {
+            @Override
+            public void handle(String message, Object... arguments) {
+                logger.debug(String.format(message, arguments));
+            }
+        };
+    }
+
+    protected CookieDecoder(boolean strict, CookieErrorHandler cookieErrorHandler) {
+        this.strict = strict;
+        this.cookieErrorHandler = cookieErrorHandler;
     }
 
     protected DefaultCookie initCookie(String header, int nameBegin, int nameEnd, int valueBegin, int valueEnd) {
         if (nameBegin == -1 || nameBegin == nameEnd) {
-            logger.debug("Skipping cookie with null name");
+            cookieErrorHandler.handle("Skipping cookie with null name");
             return null;
         }
 
         if (valueBegin == -1) {
-            logger.debug("Skipping cookie with null value");
+            cookieErrorHandler.handle("Skipping cookie with null value");
             return null;
         }
 
         CharSequence wrappedValue = CharBuffer.wrap(header, valueBegin, valueEnd);
         CharSequence unwrappedValue = unwrapValue(wrappedValue);
         if (unwrappedValue == null) {
-            logger.debug("Skipping cookie because starting quotes are not properly balanced in '{}'",
-                    wrappedValue);
+            cookieErrorHandler.handle("Skipping cookie because starting quotes are not properly balanced" +
+                    " in '%s'", wrappedValue);
             return null;
         }
 
@@ -60,10 +72,8 @@ public abstract class CookieDecoder {
 
         int invalidOctetPos;
         if (strict && (invalidOctetPos = firstInvalidCookieNameOctet(name)) >= 0) {
-            if (logger.isDebugEnabled()) {
-                logger.debug("Skipping cookie because name '{}' contains invalid char '{}'",
-                        name, name.charAt(invalidOctetPos));
-            }
+            cookieErrorHandler.handle("Skipping cookie because name '%s' contains invalid char '%s'",
+                    name, name.charAt(invalidOctetPos));
             return null;
         }
 
@@ -71,7 +81,7 @@ public abstract class CookieDecoder {
 
         if (strict && (invalidOctetPos = firstInvalidCookieValueOctet(unwrappedValue)) >= 0) {
             if (logger.isDebugEnabled()) {
-                logger.debug("Skipping cookie because value '{}' contains invalid char '{}'",
+                cookieErrorHandler.handle("Skipping cookie because value '%s' contains invalid char '%s'",
                         unwrappedValue, unwrappedValue.charAt(invalidOctetPos));
             }
             return null;
@@ -80,5 +90,9 @@ public abstract class CookieDecoder {
         DefaultCookie cookie = new DefaultCookie(name, unwrappedValue.toString());
         cookie.setWrap(wrap);
         return cookie;
+    }
+
+    public interface CookieErrorHandler {
+       void handle(String message, Object... arguments);
     }
 }
--- a/codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
@@ -56,6 +56,11 @@ public final class ServerCookieDecoder extends CookieDecoder {
         super(strict);
     }
 
+
+    private ServerCookieDecoder(ServerCookieDecoderBuilder builder) {
+        super(builder.strict, builder.cookieErrorHandler);
+    }
+
     /**
      * Decodes the specified Set-Cookie HTTP header value into a {@link Cookie}.
      *
@@ -151,5 +156,24 @@ public final class ServerCookieDecoder extends CookieDecoder {
         }
 
         return cookies;
+    }
+
+    public static class ServerCookieDecoderBuilder {
+        private boolean strict;
+        private CookieErrorHandler cookieErrorHandler;
+
+        public ServerCookieDecoderBuilder strict(boolean strict) {
+            this.strict = strict;
+            return this;
+        }
+
+        public ServerCookieDecoderBuilder cookieErrorHandler(CookieErrorHandler cookieErrorHandler) {
+            this.cookieErrorHandler = cookieErrorHandler;
+            return this;
+        }
+
+        public ServerCookieDecoder build() {
+            return new ServerCookieDecoder(this);
+        }
     }
 }
--- a/codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
@@ -56,7 +56,6 @@ public final class ServerCookieDecoder extends CookieDecoder {
         super(strict);
     }
 
-
     private ServerCookieDecoder(ServerCookieDecoderBuilder builder) {
         super(builder.strict, builder.cookieErrorHandler);
     }
--- a/codec-http/src/test/java/io/netty/handler/codec/http/cookie/ServerCookieDecoderTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/cookie/ServerCookieDecoderTest.java
@@ -86,7 +86,8 @@ public class ServerCookieDecoderTest {
         c = it.next();
         assertEquals("__utmz", c.name());
         assertEquals("48461872.1258140131.1.1.utmcsr=overstock.com|" +
-                        "utmccn=(referral)|utmcmd=referral|utmcct=/Home-Garden/Furniture/Clearance/clearance/32/dept.html",
+                        "utmccn=(referral)|utmcmd=referral|utmcct=/Home-Garden" +
+                        "/Furniture/Clearance/clearance/32/dept.html",
                 c.value());
 
         c = it.next();
--- a/codec-http/src/test/java/io/netty/handler/codec/http/cookie/ServerCookieDecoderTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/cookie/ServerCookieDecoderTest.java
@@ -20,6 +20,7 @@ import org.junit.Test;
 import java.util.Iterator;
 import java.util.Set;
 
+import static io.netty.handler.codec.http.cookie.ServerCookieDecoder.ServerCookieDecoderBuilder;
 import static org.junit.Assert.*;
 
 public class ServerCookieDecoderTest {
@@ -56,12 +57,12 @@ public class ServerCookieDecoderTest {
     @Test
     public void testDecodingGoogleAnalyticsCookie() {
         String source =
-            "ARPT=LWUKQPSWRTUN04CKKJI; " +
-            "kw-2E343B92-B097-442c-BFA5-BE371E0325A2=unfinished_furniture; " +
-            "__utma=48461872.1094088325.1258140131.1258140131.1258140131.1; " +
-            "__utmb=48461872.13.10.1258140131; __utmc=48461872; " +
-            "__utmz=48461872.1258140131.1.1.utmcsr=overstock.com|utmccn=(referral)|" +
-                    "utmcmd=referral|utmcct=/Home-Garden/Furniture/Clearance/clearance/32/dept.html";
+                "ARPT=LWUKQPSWRTUN04CKKJI; " +
+                        "kw-2E343B92-B097-442c-BFA5-BE371E0325A2=unfinished_furniture; " +
+                        "__utma=48461872.1094088325.1258140131.1258140131.1258140131.1; " +
+                        "__utmb=48461872.13.10.1258140131; __utmc=48461872; " +
+                        "__utmz=48461872.1258140131.1.1.utmcsr=overstock.com|utmccn=(referral)|" +
+                        "utmcmd=referral|utmcct=/Home-Garden/Furniture/Clearance/clearance/32/dept.html";
         Set<Cookie> cookies = ServerCookieDecoder.STRICT.decode(source);
         Iterator<Cookie> it = cookies.iterator();
         Cookie c;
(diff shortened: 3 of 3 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDecoder.java
```diff
@@ -32,27 +32,49 @@
...
 
     protected CookieDecoder(boolean strict) {
-        this.strict = strict;
+        this(strict, false);
     }
 
     protected DefaultCookie initCookie(String header, int nameBegin, int nameEnd, int valueBegin, int valueEnd) {
         if (nameBegin == -1 || nameBegin == nameEnd) {
-            logger.debug("Skipping cookie with null name");
+            if (logger.isDebugEnabled()) {
```
fenik17: The check for `isDebugEnabled()` is redundant in the such cases. I think, it's should be logging or exception thrown (not both):
```java
private final boolean needReport;

protected CookieDecoder(boolean strict, boolean reportException) {
    this.strict = strict;
    this.reportException = reportException;
    needReport = reportException || logger.isDebugEnabled();
}

private final void reportError(String message) {
    if (reportException) {
        throw new IllegalArgumentException(message);
    } else {
        logger.debug(message);
    }
}

...
if (needReport) {
    reportError("Some error " + xxx + " ...");
}
```

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDecoder.java
```diff
@@ -62,7 +84,12 @@ protected DefaultCookie initCookie(String header, int nameBegin, int nameEnd, in
         if (strict && (invalidOctetPos = firstInvalidCookieNameOctet(name)) >= 0) {
             if (logger.isDebugEnabled()) {
                 logger.debug("Skipping cookie because name '{}' contains invalid char '{}'",
-                        name, name.charAt(invalidOctetPos));
+                        name, invalidOctetPos);
```
fenik17: For why this change? An `invalidOctetPos` is not a char. The same below.

codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
```diff
@@ -152,4 +152,23 @@ private ServerCookieDecoder(boolean strict) {
 
         return cookies;
     }
+
+    public static class ServerCookieDecoderBuilder {
```
fenik17: final?

codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
```diff
@@ -152,4 +152,23 @@ private ServerCookieDecoder(boolean strict) {
...
+            return this;
+        }
+
+        public ServerCookieDecoderBuilder reportError(boolean reportError) {
+            this.reportError = reportError;
+            return this;
+        }
+
+        public ServerCookieDecoder build() {
+            return new ServerCookieDecoder(strict, reportError);
```
netkins: MAJOR Avoid instantiation through private constructors from outside of the constructors class. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AAccessorClassGeneration)

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDecoder.java
```diff
@@ -32,27 +32,49 @@
...
     }
 
     protected DefaultCookie initCookie(String header, int nameBegin, int nameEnd, int valueBegin, int valueEnd) {
         if (nameBegin == -1 || nameBegin == nameEnd) {
-            logger.debug("Skipping cookie with null name");
+            if (logger.isDebugEnabled()) {
+                logger.debug("Skipping cookie with null name");
+            }
+            if (reportException) {
+                throw new IllegalArgumentException("Skipping cookie with null name");
```
slandelle: Invalid error message: you're not skipping, you're crashing.

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDecoder.java
```diff
@@ -32,27 +32,49 @@
...
         }
 
         if (valueBegin == -1) {
-            logger.debug("Skipping cookie with null value");
+            if (logger.isDebugEnabled()) {
+                logger.debug("Skipping cookie with null value");
+            }
+            if (reportException) {
+                throw new IllegalArgumentException("Skipping cookie with null value");
+            }
```
slandelle: Same as above: redundant check + invalid exception message.

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDecoder.java
```diff
@@ -32,27 +32,49 @@
...
-            logger.debug("Skipping cookie because starting quotes are not properly balanced in '{}'",
-                    wrappedValue);
+            if (logger.isDebugEnabled()) {
+                logger.debug("Skipping cookie because starting quotes are not properly balanced in '{}'", wrappedValue);
+            }
+            if (reportException) {
+                throw new IllegalArgumentException(
+                        String.format("Skipping cookie because starting quotes are not properly balanced in '%s'",
+                                wrappedValue));
+            }
```
slandelle: same as above

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDecoder.java
```diff
@@ -62,7 +84,12 @@ protected DefaultCookie initCookie(String header, int nameBegin, int nameEnd, in
...
             if (logger.isDebugEnabled()) {
                 logger.debug("Skipping cookie because name '{}' contains invalid char '{}'",
-                        name, name.charAt(invalidOctetPos));
+                        name, invalidOctetPos);
+            }
+            if (reportException) {
+                throw new IllegalArgumentException(
+                        String.format("Skipping cookie because name '%s' contains invalid char '%c'",
+                                name, invalidOctetPos));
             }
```
slandelle: same as above

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDecoder.java
```diff
@@ -72,7 +99,12 @@ protected DefaultCookie initCookie(String header, int nameBegin, int nameEnd, in
         if (strict && (invalidOctetPos = firstInvalidCookieValueOctet(unwrappedValue)) >= 0) {
             if (logger.isDebugEnabled()) {
                 logger.debug("Skipping cookie because value '{}' contains invalid char '{}'",
-                        unwrappedValue, unwrappedValue.charAt(invalidOctetPos));
+                    unwrappedValue, invalidOctetPos);
+            }
+            if (reportException) {
+                throw new IllegalArgumentException(
+                        String.format("Skipping cookie because value '%s' contains invalid char '%c'",
+                                unwrappedValue, invalidOctetPos));
```
slandelle: same as above

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDecoder.java
alobodzki: Looks good, but your proposed version introduces computation that is not present in my, or original version as the string will have to be computed, so it will be executed on each request. In my code the JIT will just remove the whole branch and the change will be neutral, do you agree?

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDecoder.java
alobodzki: Good point, I missed that.

codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
alobodzki: Yes.

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDecoder.java
fenik17: No.. In my example string computation will be only if `needReport` is true.

alobodzki: Yes, your code is better, cheers.

normanmaurer:
@slandelle happy with it now ?

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDecoder.java
```diff
@@ -32,47 +32,61 @@
...
     protected CookieDecoder(boolean strict) {
-        this.strict = strict;
+        this(strict, false);
     }
 
     protected DefaultCookie initCookie(String header, int nameBegin, int nameEnd, int valueBegin, int valueEnd) {
         if (nameBegin == -1 || nameBegin == nameEnd) {
-            logger.debug("Skipping cookie with null name");
+            if (needReport) {
+                reportError(String.format("Invalid cookie with null name found in header '%s'", header));
```
Scottmitch: would it make more sense to format these as a `String format, Object... args`? That way you can avoid allocation if the logging level isn't "high" enough.

normanmaurer: +1

fenik17: @Scottmitch logging level is taken into account the `needReport` field :)

Scottmitch: I would prefer not to have coupling between `needReport` and the logging level. The log methods (`log.debug`, etc..) should be responsible for enforcing whether logging action is taken.

fenik17: @Scottmitch have you any ideas how avoid duplicate a message computation for logging and exception? It can be simple with lambdas.. But, we can't use lambdas.

Scottmitch: can you show me how you would do it with lambdas?

fenik17: @Scottmitch something like this:
```java
...
reportError(() -> String.format("Invalid cookie with null name found in header '%s'", header));
...

private void reportError(Supplier<String> messageSupplier) {
    if (reportException) {
        throw new IllegalArgumentException(messageSupplier.get());
    }
        logger.debug(messageSupplier.get());
    }
}
```

Such approach (lazy string computation through lambdas) is used e.g. in the `log4j2` [1].

[1] https://logging.apache.org/log4j/2.x/manual/api.html#LambdaSupport

fenik17: I propose to replace the field `needReport` with the method:
```java
private boolean isNeedReportError() {
    return reportException || logger.isDebugEnabled();
}
```
This will fix the problem with changing logging level in runtime.

fenik17: There is another option with override..

in the `CookieDecoder`:
```java
protected boolean isNeedReportError() {
    return logger.isDebugEnabled();
}

protected void reportError(String message) {
    logger.debug(message);
}

...
if (isNeedReportError()) {
    reportError(String.format("..."));
}
...
```
in the `ServerCookieDecoder`:
```java
public static final ServerCookieDecoder FAIL_FAST = new ServerCookieDecoder(true) {
    @Override
    protected boolean isNeedReportError() {
        return true;
    }

    @Override
    protected void reportError(String message) {
        throw new IllegalArgumentException(message);
    }
}
```


I like this more.

normanmaurer: @Scottmitch WDYT ?

Scottmitch: @fenik17 - sounds reasonable. Lambdas in this case are mostly syntactic sugar. You can just create your own interface and implement the necessary methods. We have our own `Supplier`s (e.g. `BooleanSupplier`) and intend to remove them in favor of the JDK8's when we can drop support for older JDK versions ... you can create an internal interface or maybe even consider a more visible `Supplier<T>` if necessary.

fenik17: @Scottmitch I can provide PR with the such approach after the current PR will be merged.

Scottmitch: followup PR sgtm

alobodzki:
@fenik17 I like the lambda approach better, doesn't make sense to merge this commit and build on top of it. I added some example implementation of this feature.

codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
```diff
@@ -152,4 +156,23 @@ private ServerCookieDecoder(boolean strict) {
...
+            return this;
+        }
+
+        public ServerCookieDecoderBuilder cookieErrorHandler(CookieErrorHandler cookieErrorHandler) {
+            this.cookieErrorHandler = cookieErrorHandler;
+            return this;
+        }
+
+        public ServerCookieDecoder build() {
+            return new ServerCookieDecoder(this);
```
netkins: MAJOR Avoid instantiation through private constructors from outside of the constructors class. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AAccessorClassGeneration)
MAJOR Avoid instantiation through private constructors from outside of the constructors class. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AAccessorClassGeneration)

codec-http/src/test/java/io/netty/handler/codec/http/cookie/ServerCookieDecoderTest.java
```diff
@@ -99,49 +101,49 @@ public void testDecodingGoogleAnalyticsCookie() {
     public void testDecodingLongValue() {
         String longValue =
                 "b___$Q__$ha__<NC=MN(F__%#4__<NC=MN(F__2_d____#=IvZB__2_F____'=KqtH__2-9____" +
-                "'=IvZM__3f:____$=HbQW__3g'____%=J^wI__3g-____%=J^wI__3g1____$=HbQW__3g2____" +
```
normanmaurer: please revert formatting changes

codec-http/src/test/java/io/netty/handler/codec/http/cookie/ServerCookieDecoderTest.java
```diff
@@ -56,12 +57,12 @@ public void testDecodingMultipleCookies() {
...
-            "__utma=48461872.1094088325.1258140131.1258140131.1258140131.1; " +
-            "__utmb=48461872.13.10.1258140131; __utmc=48461872; " +
-            "__utmz=48461872.1258140131.1.1.utmcsr=overstock.com|utmccn=(referral)|" +
-                    "utmcmd=referral|utmcct=/Home-Garden/Furniture/Clearance/clearance/32/dept.html";
+                "ARPT=LWUKQPSWRTUN04CKKJI; " +
+                        "kw-2E343B92-B097-442c-BFA5-BE371E0325A2=unfinished_furniture; " +
+                        "__utma=48461872.1094088325.1258140131.1258140131.1258140131.1; " +
+                        "__utmb=48461872.13.10.1258140131; __utmc=48461872; " +
+                        "__utmz=48461872.1258140131.1.1.utmcsr=overstock.com|utmccn=(referral)|" +
+                        "utmcmd=referral|utmcct=/Home-Garden/Furniture/Clearance/clearance/32/dept.html";
```
normanmaurer: please revert formatting changes

codec-http/src/test/java/io/netty/handler/codec/http/cookie/ServerCookieDecoderTest.java
```diff
@@ -173,10 +175,36 @@ public void testDecodingOldRFC2965Cookies() {
 
     @Test
     public void testRejectCookieValueWithSemicolon() {
-        Set<Cookie> cookies = ServerCookieDecoder.STRICT.decode("name=\"foo;bar\";");
+        Set<Cookie> cookies = ServerCookieDecoder.STRICT.decode("name=\"foo,bar\";");
```
normanmaurer: revert ?

codec-http/src/main/java/io/netty/handler/codec/http/cookie/CookieDecoder.java
```diff
@@ -81,4 +91,8 @@ protected DefaultCookie initCookie(String header, int nameBegin, int nameEnd, in
         cookie.setWrap(wrap);
         return cookie;
     }
+
+    public interface CookieErrorHandler {
```
Scottmitch: As discussed in https://github.com/netty/netty/pull/6744/#discussion_r116452809 I think adding a `STRICT` and `LAX` options should be good enough and we can make this interface an implementation detail (not expose it publicly).

normanmaurer:
@alobodzki ping

normanmaurer:
Closing this due no activity.

Issue #6741, opened 2017-05-13
kqueue bootstrap 'Failed to set channel option 'IP_TOS'' (4.1.11.Final)


jhrussell:
Running into an error trying to set IP_TOS while using the new kqueue native transport.  This works for both NIO and EPOLL transports.  And by works, I mean at the very least it doesn't throw any errors.

### Expected behavior
No errors.  Perhaps even setting channel option IP_TOS.

### Actual behavior
Warning/exception thrown.  The channel appears to be bootstrapped okay otherwise.

> 2017-05-12T17:41:46.835 [KQueueEventLoopGroup-4-2] WARN  io.netty.bootstrap.ServerBootstrap - Failed to set channel option 'IP_TOS' with value '184' for channel '[id: 0x413f8ecb, L:/127.0.0.1:8554 - R:/127.0.0.1:55500]'
> io.netty.channel.ChannelException: setsockopt() failed: Invalid argument
> 	at io.netty.channel.unix.Socket.setTrafficClass(Native Method)
> 	at io.netty.channel.unix.Socket.setTrafficClass(Socket.java:385)
> 	at io.netty.channel.kqueue.KQueueSocketChannelConfig.setTrafficClass(KQueueSocketChannelConfig.java:288)
> 	at io.netty.channel.kqueue.KQueueSocketChannelConfig.setOption(KQueueSocketChannelConfig.java:116)
> 	at io.netty.bootstrap.AbstractBootstrap.setChannelOption(AbstractBootstrap.java:459)
> 	at io.netty.bootstrap.AbstractBootstrap.setChannelOptions(AbstractBootstrap.java:451)
> 	at io.netty.bootstrap.ServerBootstrap$ServerBootstrapAcceptor.channelRead(ServerBootstrap.java:248)
> 	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
> 	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
> 	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
> 	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1334)
> 	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
[...]

PR #6749, opened 2017-05-18
KQueueSocket#setTrafficClass exceptions

Scottmitch:
Motivation:
MacOS will throw an error when attempting to set the IP_TOS socket option if IPv6 is available, and also when getting the value for IP_TOS.

Modifications:
- Socket#setTrafficClass and Socket#getTrafficClass should try to use IPv6 first, and check if the error code indicates the protocol is not supported before trying IPv4

Result:
Fixes https://github.com/netty/netty/issues/6741.

### DIFF
--- a/transport-native-epoll/src/test/java/io/netty/channel/epoll/EpollSocketTest.java
+++ b/transport-native-epoll/src/test/java/io/netty/channel/epoll/EpollSocketTest.java
@@ -19,6 +19,7 @@ import io.netty.channel.unix.DomainSocketAddress;
 import io.netty.channel.unix.PeerCredentials;
 import io.netty.channel.unix.tests.SocketTest;
 import io.netty.channel.unix.tests.UnixTestUtils;
+import org.junit.BeforeClass;
 import org.junit.Test;
 
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


Scottmitch:
@jhrussell - Please verify this fixes the issue.

jhrussell:
I'm not 100% sure I'm properly set up to build and and test this change locally, but from my attempt to do so, this appears to work.

transport-native-epoll/src/test/java/io/netty/channel/epoll/EpollSocketTest.java
```diff
@@ -19,15 +19,22 @@
...
 
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assume.assumeTrue;
 
 public class EpollSocketTest extends SocketTest<LinuxSocket> {
+    @BeforeClass
+    public static void loadJNI() {
+        assumeTrue(Epoll.isAvailable());
```
normanmaurer: I wonder if we should fix this in an extra commit as this will also need to be backported to 4.0 imho and is not really related to this PR

Scottmitch: This is already done in 4.0 [1] and the file structure isn't the same anymore so the back port would be a bit messy. Can you clarify what action you want me to take?

[1] https://github.com/netty/netty/blob/4.0/transport-native-epoll/src/test/java/io/netty/channel/unix/SocketTest.java#L31

Scottmitch: See https://github.com/netty/netty/pull/6753

Scottmitch:
4.1 (4c6d946fb)

PR #6753, opened 2017-05-18
SocketTest assumeTrue(EPOLL)

Scottmitch:
Motivation:
SocketTest should verify that EPOLL is available before running the test.

Modifications:
- Add a @BeforeClass method which assumeTrue if EPOLL is available

Result:
More robust tests than get skipped if EPOLL isn't available


Scottmitch:
4.0 (02be19a5d)

PR #6756, opened 2017-05-19
Share code that is needed to support shaded native libraries.

normanmaurer:
Motivation:

For our native libraries in netty we support shading, to have this work on runtime the user needs to set a system property. This code should shared.

Modifications:

Move logic to NativeLbiraryLoader and so share for all native libs.

Result:

Less code duplication and also will work for netty-tcnative out of the box once it support shading


normanmaurer:
Cherry-picked into 4.1 (201d9b6536) and 4.0 (305b7dd6e31940e1b4e2aefcf3e6bbed7a5d1093)

PR #6760, opened 2017-05-19
SslHandler#handlerRemoved0() shouldn't care about the SSLEngine being…

rkapsi:
… a specific type but only if it's ReferenceCounted

Motivation

SslHandler should release any type of SSLEngine if it implements the ReferenceCounted interface

Modifications

Change condition to check for ReferenceCounted interface

Result

Better use of interfaces


normanmaurer:
Thx let me check

normanmaurer:
Cherry-picked into 4.1 (915bf5f5b7d0afa40c9921bea0025d80cb29a51a) and 4.0 (7d04299938d88bf833898fb2614a505128342174)

@rkapsi thanks!

Scottmitch:
thanks @rkapsi !

PR #6766, opened 2017-05-22
MQTT CONNECT message should allow arbitrary byte strings in password and will message fields.

### LINKED ISSUE #6750: codec.mqtt: wrong password and willMessage field types in MqttConnectPayload/MqttEncoder (should be byte[])

oakad:
Motivation

    Mandated by OASIS MQTT v3.1.1 spec

Modifications

    Store password and will message data as byte[] fields in MqttConnectPayload.
    Modify MqttEncoder and MqttDecoder accordingly.
    MqttConnectPayload will handle the string conversion if required by users
    of older protocol versions.

Result

    Fixes #6750

Signed-off-by: Alex Dubov <oakad@yahoo.com>

### DIFF
--- a/codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttConnectPayload.java
+++ b/codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttConnectPayload.java
@@ -16,6 +16,7 @@
 
 package io.netty.handler.codec.mqtt;
 
+import io.netty.util.CharsetUtil;
 import io.netty.util.internal.StringUtil;
 
 /**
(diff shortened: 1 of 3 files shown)


Scottmitch:
Can you fix the check style issues? Building locally should catch these issues.

```
[08:50:45][io.netty:netty-codec-mqtt] /opt/teamcity-agent/work/a00bdf77e621d005/codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttEncoder.java:116: Line is longer than 120 characters (found 124).
[08:50:45][io.netty:netty-codec-mqtt] /opt/teamcity-agent/work/a00bdf77e621d005/codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttEncoder.java:116:1: File contains tab characters (this is the first instance).
```

Scottmitch:
Also can you sign the [ICLA](http://netty.io/s/icla)?

PR #6769, opened 2017-05-23
MQTT CONNECT message should allow arbitrary byte strings in password and will message fields.

### LINKED ISSUE #6750: codec.mqtt: wrong password and willMessage field types in MqttConnectPayload/MqttEncoder (should be byte[])

oakad:
Motivation

    Mandated by OASIS MQTT v3.1.1 spec

Modifications

    Store password and will message data as byte[] fields in MqttConnectPayload.
    Modify MqttEncoder and MqttDecoder accordingly.
    MqttConnectPayload will handle the string conversion if required by users
    of older protocol versions.

Result

    Fixes #6750

Signed-off-by: Alex Dubov <oakad@yahoo.com>

### DIFF
--- a/codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttConnectPayload.java
+++ b/codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttConnectPayload.java
@@ -16,6 +16,7 @@
 
 package io.netty.handler.codec.mqtt;
 
+import io.netty.util.CharsetUtil;
 import io.netty.util.internal.StringUtil;
 
 /**
@@ -25,9 +26,9 @@ public final class MqttConnectPayload {
 
     private final String clientIdentifier;
     private final String willTopic;
-    private final String willMessage;
+    private final byte[] willMessage;
     private final String userName;
-    private final String password;
+    private final byte[] password;
 
     public MqttConnectPayload(
             String clientIdentifier,
@@ -35,6 +36,19 @@ public final class MqttConnectPayload {
             String willMessage,
             String userName,
             String password) {
+        this.clientIdentifier = clientIdentifier;
+        this.willTopic = willTopic;
+        this.willMessage = willMessage.getBytes(CharsetUtil.UTF_8);
+        this.userName = userName;
+        this.password = password.getBytes(CharsetUtil.UTF_8);
+    }
+
+     public MqttConnectPayload(
+            String clientIdentifier,
+            String willTopic,
+            byte[] willMessage,
+            String userName,
+            byte[] password) {
         this.clientIdentifier = clientIdentifier;
         this.willTopic = willTopic;
         this.willMessage = willMessage;
@@ -51,6 +65,10 @@ public final class MqttConnectPayload {
     }
 
     public String willMessage() {
+        return willMessageBytes() != null ? new String(willMessageBytes(), CharsetUtil.UTF_8) : null;
+    }
+
+    public byte[] willMessageBytes() {
         return willMessage;
     }
 
@@ -59,6 +77,10 @@ public final class MqttConnectPayload {
     }
 
     public String password() {
+        return passwordBytes() != null ? new String(passwordBytes(), CharsetUtil.UTF_8) : null;
+    }
+
+    public byte[] passwordBytes() {
         return password;
     }
 
--- a/codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttDecoder.java
+++ b/codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttDecoder.java
@@ -339,21 +339,21 @@ public final class MqttDecoder extends ReplayingDecoder<DecoderState> {
         int numberOfBytesConsumed = decodedClientId.numberOfBytesConsumed;
 
         Result<String> decodedWillTopic = null;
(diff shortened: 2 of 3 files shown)


Scottmitch:
for future reference ... you don't have to close the old PR (https://github.com/netty/netty/pull/6766) and open a new one ... just push new commits to the same branch and the PR will reflect the updated code.

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttConnectPayload.java
```diff
@@ -51,6 +65,10 @@ public String willTopic() {
     }
 
     public String willMessage() {
```
Scottmitch: please deprecate this method an pointing folks to `willMessageBytes()`.
also consider adding to the javadocs that this method may create a new string on each invocation.

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttConnectPayload.java
```diff
@@ -59,6 +77,10 @@ public String userName() {
     }
 
     public String password() {
```
Scottmitch: please deprecate this method an pointing folks to `passwordBytes()`.
also consider adding to the javadocs that this method may create a new string on each invocation.

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttDecoder.java
```diff
@@ -430,6 +430,22 @@ private static MqttFixedHeader decodeFixedHeader(ByteBuf buffer) {
...
+        final Result<Integer> decodedSize = decodeMsbLsb(buffer);
+        int size = decodedSize.value;
+        int numberOfBytesConsumed = decodedSize.numberOfBytesConsumed;
+        if (size < minBytes || size > maxBytes) {
+            buffer.skipBytes(size);
+            numberOfBytesConsumed += size;
+            return new Result<byte[]>(null, numberOfBytesConsumed);
+        }
+        byte[] b = new byte[size];
+        buffer.getBytes(buffer.readerIndex(), b);
```
Scottmitch: replace the call to `getBytes` and `skipBytes` with a call to `buffer.readBytes(b);`

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttConnectPayload.java
```diff
@@ -37,6 +38,19 @@ public MqttConnectPayload(
...
+    }
+
+     public MqttConnectPayload(
+            String clientIdentifier,
+            String willTopic,
+            byte[] willMessage,
+            String userName,
+            byte[] password) {
+        this.clientIdentifier = clientIdentifier;
+        this.willTopic = willTopic;
```
normanmaurer: I think we need to call `willTopic.clone()` and `password.clone()` here to ensure the user not later modify it.

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttConnectPayload.java
normanmaurer: +1

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttConnectPayload.java
```diff
@@ -51,6 +65,10 @@ public String willTopic() {
     }
 
     public String willMessage() {
+        return willMessageBytes() != null ? new String(willMessageBytes(), CharsetUtil.UTF_8) : null;
+    }
+
+    public byte[] willMessageBytes() {
```
normanmaurer: Do we need to call willMessage.clone() to ensure the user not modifies it ?

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttConnectPayload.java
```diff
@@ -59,6 +77,10 @@ public String userName() {
     }
 
     public String password() {
+        return passwordBytes() != null ? new String(passwordBytes(), CharsetUtil.UTF_8) : null;
+    }
+
+    public byte[] passwordBytes() {
```
normanmaurer: same as above

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttDecoder.java
normanmaurer: +1

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttConnectPayload.java
oakad: MQTT v3.1 requires string, MQTT v3.1.1 and above requires bytes. So technically, this can not be deprecated as long as v3.1 is around.

oakad:
I will stick to a forked version, thanks.

Sorry for the noise.

Scottmitch:
@oakad - Do you mean to say you don't think this change is correct or that you don't want to contribute this change? Unless you think this is incorrect we will likely pull it in and make the necessary changes from review.

PR #6771, opened 2017-05-23
Don't add null to SmtpResponse.details()

mcobrien:
Motivation:

If the remote server returns an invalid response in the form `000 \r\n`
(i.e. a three digit code, then space, but no details), null is added
as a singletonList to the response being constructed.

This seems unexpected and it would be easier to handle an empty
`details` list in client code.

Modifications:

If `detail` is null (because `frame.isReadable()` returned false after
reading the separator), initialise `DefaultSmtpResponse` with an empty
list instead of a list containing a single null value.

Result:

When encountering this malformed server response, a `DefaultSmtpResponse`
with a code but no details will be created.

@normanmaurer

### DIFF
--- a/codec-smtp/src/main/java/io/netty/handler/codec/smtp/SmtpResponseDecoder.java
+++ b/codec-smtp/src/main/java/io/netty/handler/codec/smtp/SmtpResponseDecoder.java
@@ -69,7 +69,11 @@ public final class SmtpResponseDecoder extends LineBasedFrameDecoder {
                         details.add(detail);
                     }
                 } else {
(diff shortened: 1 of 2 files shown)


normanmaurer:
@mcobrien please fix the check style error

mcobrien:
sorry about that, fixed now!

normanmaurer:
@mcobrien awesome... lgtm

normanmaurer:
Cherry-picked into 4.1 (12a57545690aec0a263305442769fe0cf726aa44)

Issue #6770, opened 2017-05-23
Channel.Unsafe.closeForcibly() fails for EpollSocketChannel if not registered to event loop


trustin:
### Expected behavior

`Channel.Unsafe.closeForcibly()` should close the connection immediately even if the channel is not registered to an event loop.

### Actual behavior

The connection is closed, but an exception is accompanied.
```
05:53:55.405 [armeria-server-boss-epoll-26-1] WARN  io.netty.channel.AbstractChannel - Failed to close a channel.
java.lang.IllegalStateException: channel not registered to an event loop
	at io.netty.channel.AbstractChannel.eventLoop(AbstractChannel.java:158)
	at io.netty.channel.epoll.AbstractEpollChannel.doDeregister(AbstractEpollChannel.java:140)
	at io.netty.channel.epoll.AbstractEpollChannel.doClose(AbstractEpollChannel.java:117)
	at io.netty.channel.epoll.AbstractEpollStreamChannel.doClose(AbstractEpollStreamChannel.java:721)
	at io.netty.channel.AbstractChannel$AbstractUnsafe.closeForcibly(AbstractChannel.java:703)
```

### Steps to reproduce

Accept an incoming connection from an epoll server socket and close the accepted connection using `Channel.Unsafe.closeForcibly()` before it's registered to an event loop.

### Minimal yet complete reproducer code (or URL to code)

```java
ServerBootstrap sb = new ServerBootstrap();
...
sb.handler(new ChannelInboundHandlerAdapter() {
    @Override
    public void channelRead(ctx, msg) {
        Channel childChannel = (Channel) msg;
        if (isBlacklisted(childChannel)) {
            ((SocketChannelConfig) childChannel.config()).setSoLinger(0); // Send RST
            childChannel.unsafe().closeForcibly();
        } else {
            ctx.fireChannelRead(msg);
        }
    }
});
```

### Netty version

4.1.11

normanmaurer:
@trustin does it fail with nio as well or only epoll ?

trustin:
Epoll only. By the way, the connection seems to be closed because `socket.close()` is in the `finally` block. It raises an exception while trying to deregister.

normanmaurer:
@trustin will check

trustin:
Thanks, @normanmaurer !

Scottmitch:
To clarify ... the issue is an exception is thrown, not that the channel isn't closed, correct?

Scottmitch:
[...]

PR #6772, opened 2017-05-23
Only try to deregister from EventLoop when the native Channel was reg…

normanmaurer:
…istered before.

Motivation:

We only can call eventLoop() if we are registered on an EventLoop yet. As we just did this without checking we spammed the log with an error that was harmless.

Modifications:

Check if registered on eventLoop before try to deregister on close.

Result:

Fixes [#6770]

### DIFF
--- a/testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketCloseForciblyTest.java
+++ b/testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketCloseForciblyTest.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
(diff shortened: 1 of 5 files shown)


normanmaurer:
@trustin FYI

testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketCloseForciblyTest.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+package io.netty.testsuite.transport.socket;
+
+import io.netty.bootstrap.Bootstrap;
+import io.netty.bootstrap.ServerBootstrap;
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.channel.socket.SocketChannel;
+import org.junit.Assert;
```
netkins: MINOR Remove this unused import 'org.junit.Assert'. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketCloseForciblyTest.java
```diff
@@ -0,0 +1,53 @@
...
+import org.junit.Test;
+
+import java.util.concurrent.atomic.AtomicReference;
```
netkins: MINOR Remove this unused import 'java.util.concurrent.atomic.AtomicReference'. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

trustin:
Thanks a lot, @normanmaurer !

normanmaurer:
Cherry-picked into 4.1 (2f8fe2af01c1bd5f6e02a3a10ea949318705a325) and 4.0 (a45f63aea02d82bcccf0455a9e95f0b3d5dabeba)

Issue #6768, opened 2017-05-22
Adding handlers to a closed Channel's pipeline


rkapsi:
It appears the a ChannelPipeline will happily add a handler to a closed Channel's pipeline and will call `handlerAdded(...)` but will not call `handlerRemoved(...)` or blow up in some other way to let the caller know that things aren't in a good state.

Just wondering if there wouldn't be edge cases where we'd potentially leak resources (e.g. `SslHandler#handlerRemoved0(...)` which is supposed to release a reference counted SSLEngine).

### Expected behavior

Fail with Exception or `handlerRemoved(...)` getting called after `handlerAdded(...)` 

### Actual behavior

`handlerAdded(...)`  getting called as if everything is ok

### Minimal yet complete reproducer code (or URL to code)

```java
@Test
public void testHandlerAfterClose() throws Exception {
  EmbeddedChannel channel = new EmbeddedChannel();
  channel.close().syncUninterruptibly();
  
  assertFalse(channel.isActive());
  
  final CountDownLatch addedLatch = new CountDownLatch(1);
  final CountDownLatch removedLatch = new CountDownLatch(1);
  
  ChannelPipeline pipeline = channel.pipeline();
  pipeline.addLast(new ChannelHandlerAdapter() {
    @Override
    public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
      addedLatch.countDown();
    }

    @Override
    public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
      removedLatch.countDown();
    }
  });
  
  assertTrue(addedLatch.await(1L, TimeUnit.SECONDS));
  assertTrue(removedLatch.await(1L, TimeUnit.SECONDS));
}
```

normanmaurer:
@rkapsi will check

Scottmitch:
thanks for reporting @rkapsi ... sounds like a bug to me

normanmaurer:
@Scottmitch @rkapsi I wonder what you think is the better option here. Throw an exception when calling add or just remove it directly after it is added automatically ? I guess 1) is better. WDYT ?

rkapsi:
I'd go with Exception as well.

normanmaurer:
@Scottmitch sounds good to you as well ?

Scottmitch:
Exception should be fine (static exception?), hopefully we have been diligent in handling exceptions in cases where ReferenceCounted things are added to the pipeline :sweat_smile:

normanmaurer:
[...]

PR #6774, opened 2017-05-23
Not add ChannelHandler to ChannelPipeline once the pipeline was destr…

normanmaurer:
…oyed.

Motivation:

ChannelPipeline will happily add a handler to a closed Channel's pipeline and will call handlerAdded(...) but will not call handlerRemoved(...).

Modifications:

Check if pipeline was destroyed and if so not add the handler at all but propergate an exception.

Result:

Fixes [#6768]

### DIFF
--- a/transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
+++ b/transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
@@ -67,6 +67,7 @@ public class DefaultChannelPipeline implements ChannelPipeline {
     private Map<EventExecutorGroup, EventExecutor> childExecutors;
     private MessageSizeEstimator.Handle estimatorHandle;
     private boolean firstRegistration = true;
+    private boolean destroyed;
 
     /**
      * This is the head of a linked list that is processed by {@link #callHandlerAddedForAllHandlers()} and so process
@@ -138,6 +139,12 @@ public class DefaultChannelPipeline implements ChannelPipeline {
         return channel;
     }
 
+    private void checkDestroyed(ChannelHandler handler) {
+        if (destroyed) {
+            throw new ChannelPipelineException("Channel already closed, can not add handler " + handler);
+        }
+    }
+
     @Override
     public final ChannelPipeline addFirst(String name, ChannelHandler handler) {
         return addFirst(null, name, handler);
@@ -147,6 +154,7 @@ public class DefaultChannelPipeline implements ChannelPipeline {
     public final ChannelPipeline addFirst(EventExecutorGroup group, String name, ChannelHandler handler) {
         final AbstractChannelHandlerContext newCtx;
         synchronized (this) {
+            checkDestroyed(handler);
             checkMultiplicity(handler);
             name = filterName(name, handler);
 
@@ -196,6 +204,7 @@ public class DefaultChannelPipeline implements ChannelPipeline {
     public final ChannelPipeline addLast(EventExecutorGroup group, String name, ChannelHandler handler) {
         final AbstractChannelHandlerContext newCtx;
         synchronized (this) {
+            checkDestroyed(handler);
             checkMultiplicity(handler);
 
             newCtx = newContext(group, filterName(name, handler), handler);
@@ -246,6 +255,7 @@ public class DefaultChannelPipeline implements ChannelPipeline {
         final AbstractChannelHandlerContext newCtx;
         final AbstractChannelHandlerContext ctx;
         synchronized (this) {
+            checkDestroyed(handler);
             checkMultiplicity(handler);
             name = filterName(name, handler);
             ctx = getContextOrDie(baseName);
@@ -306,6 +316,7 @@ public class DefaultChannelPipeline implements ChannelPipeline {
         final AbstractChannelHandlerContext ctx;
 
         synchronized (this) {
+            checkDestroyed(handler);
             checkMultiplicity(handler);
             name = filterName(name, handler);
             ctx = getContextOrDie(baseName);
@@ -516,6 +527,7 @@ public class DefaultChannelPipeline implements ChannelPipeline {
 
         final AbstractChannelHandlerContext newCtx;
         synchronized (this) {
+            checkDestroyed(newHandler);
             checkMultiplicity(newHandler);
             if (newName == null) {
                 newName = generateName(newHandler);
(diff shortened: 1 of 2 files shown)


normanmaurer:
@rkapsi PTAL

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -138,13 +139,26 @@ public final Channel channel() {
         return channel;
     }
 
+    private boolean checkDestroyed(ChannelHandler handler) {
+        if (destroyed) {
+            fireExceptionCaught(
```
Scottmitch: not throwing an exception or any immediate indication that the insertion operation failed will be problematic for cases which assume the pipeline operation will assume reference count ownership of the handler (e.g. SniHandler inserting a SslHandler which uses the pipeline methods to manage reference count lifetime, or immediately release if exception is thrown [1]).

[1] https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/SniHandler.java#L135

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -838,6 +868,7 @@ public final ChannelPipeline fireChannelUnregistered() {
      * See: https://github.com/netty/netty/issues/3156
      */
     private synchronized void destroy() {
+        destroyed = true;
```
fenik17: maybe it make sense to check that `destroyed` is false here?

normanmaurer:
@scottmitch I am a bit on a fence here as throwing an exception is a bit of a backward incompatible change. We could argue that this is a bug and so its ok but I am not sure. @trustin WDYT ?

rkapsi:
LGTM

Maybe if you're worried about unexpected side effects then add a system property that will by default log ERROR/WARN and throw only if the flag is explicitly enabled. I'd be happy to play guinea pig.

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
normanmaurer: @Scottmitch ok let me just throw

normanmaurer:
@Scottmitch @rkapsi @trustin PTAL again

Scottmitch:
@normanmaurer - I think we should either:

1. Change behavior and throw. Note that the API permits for unchecked exceptions so it could be argued the API allows for this and our implementation just didn't take advantage.
2. Make sure if handlerAdded is called that handlerRemoved is also called.

I don't have a strong preference between the two approaches.

normanmaurer:
@Scottmitch yeah I just ended up throwing.

normanmaurer:
Cherry-picked into 4.1 (4aa800259674bab303edd9427802574fcb4a3131) and 4.0 (4cc98134da611562b7433aa4fb9937ab24d553fb)

normanmaurer:
@mosesn I would like to pick your brain on this as a consumer of netty. What do you think about this change ?

transport/src/test/java/io/netty/channel/DefaultChannelPipelineTest.java
```diff
@@ -1073,6 +1073,38 @@ public void testNotPinExecutor() {
...
+                }
+
+                @Override
+                public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
+                    removedHandler.set(true);
+                }
+            });
+            fail();
+        } catch (ChannelPipelineException e) {
+            // expected
```
mosesn: if you happen to revisit this, it would be nice to ensure that we enter this catch block

mosesn:
@normanmaurer I think this seems like a useful change.  It does seem potentially racy, but since you'll throw an exception if you race, then it's easy for implementors to handle the race, although they have to figure out that it's something that could happen first, which might not be easy to discover.  We might have to fix some code, but it doesn't *look* like it will be that painful.

My major concern is that it looks like it's possible that there is a valid reason for adding a new handler after a destroy has been triggered.  Maybe it makes more sense to prevent adding a new handler after the destroy has finished?

normanmaurer:
@mosesn thats actually a good point. We should most likely just ensure we can during destroy but not after its complete. Gimme a few minutes to do a followup and a testcase

normanmaurer:
@mosesn I just wonder if we may miss some cases if we allow to add handlers during destroy. Mainly because a user may add a handler infront or after the current handler so it may be possible that we not traverse the just added handler anymore while doing the destroy operation. Thoughts ?

mosesn:
@normanmaurer it's possible, but that's just a technical difficulty.  you could imagine adding a state machine where we set the state to be `Destroying` and if a handler gets added while it's in the `Destroying` state we add it to a data structure of things to double check were destroyed later.  Then after we finish destroying new entrants, then we can mark it `Destroyed`.

normanmaurer:
@mosesn yeah but this could also lead to have some kind of endless loop when the user add new handlers during removal all the time. Need to think a bit about it.

@mosesn just to clarify, you are not concerned about throwing exceptions like we do after this PR, correct ?

mosesn:
Yes, that's right.  I *think* it should be OK.

normanmaurer:
@mosesn we could also just call handlerRemoved directly after handlerAdded in this case but I think this would even be more weird . Agree ?

mosesn:
@normanmaurer agree.

normanmaurer:
@mosesn @rkapsi after thinking more and more I am tempted to just revert this fix for now and cut the release. I would prefer to give it a few more thoughts and I think its not super urgent to be fixed as it exists forever anyway. Agree ?

mosesn:
@normanmaurer I think you're right, that's the prudent thing to do.

rkapsi:
@normanmaurer - yea totally fine. #6768 was the result of me trying to find what ended up being the APR pointer leak. I was basically looking into the possibility of how SslContext or rather SslEngines could leak. I haven't seen it actually happen on our end and consider it just a possibility.

bryce-anderson:
@normanmaurer, I took a quick look, and it looks good to me. That said, I agree with your decision to not rush it into the release: it can wait.

normanmaurer:
done 4.1 (80aa5dcdcc870b9bd82cae4b0be2d7c9bc893491) and 4.0 (01ab204ae5a55fe627d248a6961cfd53f5ebc8df)

Scottmitch:
+1 for re-open the issue and fix after the release seeing as this issue has existing for a while

PR #5959, opened 2016-10-31
Also wakup EventLoop for write tasks

normanmaurer:
Motivation:

In 64c3f58279809141a851becba445a20badc8a610 I added an optimization to not wakupe the EventLoop if an write happenede from outside the EventLoop. This optimization did not give any real performance gain while disallow some implementations like auto-flush in an ChannelHandler.

Modifications:

Always wakup the EventLoop on writes.

Result:

More flexible and generic implementation.

PR #6779, opened 2017-05-24
Add ChannelOption which allows to wakeup EventLoop for write tasks (follow up to #5959)

NiteshKant:
This is a follow up on #5959. 

Following changes are done on top of the original commit by @normanmaurer 

- Added a benchmark to write variable number of items from outside the event loop with and without this new option enabled. Here is the result from the benchmark:

```
 Benchmark                                             (wakeUpOnWrite)  (writeCount)   Mode  Cnt      Score     Error  Units
WakeupOnEachWriteBenchmark.writeFromOutsideEventLoop             true             1  thrpt   20  45211.545 ± 304.884  ops/s
WakeupOnEachWriteBenchmark.writeFromOutsideEventLoop             true           100  thrpt   20  14645.364 ±  79.898  ops/s
WakeupOnEachWriteBenchmark.writeFromOutsideEventLoop             true          1000  thrpt   20   1761.138 ±  35.237  ops/s
WakeupOnEachWriteBenchmark.writeFromOutsideEventLoop            false             1  thrpt   20  44881.363 ± 429.171  ops/s
WakeupOnEachWriteBenchmark.writeFromOutsideEventLoop            false           100  thrpt   20  14787.720 ±  70.798  ops/s
WakeupOnEachWriteBenchmark.writeFromOutsideEventLoop            false          1000  thrpt   20   1783.418 ±  32.192  ops/s
```

- Instead of looking up this option on every write, cache the value on `channelActive()`. It would have been better if we could access the config in `AbstractChannelHandlerContext` constructor but the config isn't available with the channel for Head and Tail pipeline context.

- Added `AbstractChannelBenchmark` which is taken from my earlier [auto-flush PR](https://github.com/netty/netty/pull/5716/files#diff-0aa69930da7db31a252387d999826fba).

### DIFF
--- a/transport/src/main/java/io/netty/channel/DefaultChannelConfig.java
+++ b/transport/src/main/java/io/netty/channel/DefaultChannelConfig.java
@@ -23,18 +23,7 @@ import java.util.Map.Entry;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 
-import static io.netty.channel.ChannelOption.ALLOCATOR;
-import static io.netty.channel.ChannelOption.AUTO_CLOSE;
-import static io.netty.channel.ChannelOption.AUTO_READ;
-import static io.netty.channel.ChannelOption.CONNECT_TIMEOUT_MILLIS;
-import static io.netty.channel.ChannelOption.MAX_MESSAGES_PER_READ;
-import static io.netty.channel.ChannelOption.MESSAGE_SIZE_ESTIMATOR;
-import static io.netty.channel.ChannelOption.RCVBUF_ALLOCATOR;
-import static io.netty.channel.ChannelOption.SINGLE_EVENTEXECUTOR_PER_GROUP;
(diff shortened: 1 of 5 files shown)


transport/src/main/java/io/netty/channel/DefaultChannelConfig.java
```diff
@@ -429,4 +424,12 @@ private boolean getPinEventExecutorPerGroup() {
         return pinEventExecutor;
     }
 
+    private ChannelConfig setWakeupOnWrite(boolean wakeupOnWrite) {
+        this.wakeupOnWrite = wakeupOnWrite;
+        return this;
+    }
+
+    private boolean getWakeupOnWrite() {
```
netkins: MAJOR A getX() method which returns a boolean should be named isX() [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3ABooleanGetMethodName)

NiteshKant: This is a private method and there exists another method `getPinEventExecutorPerGroup()` which returns a boolean. I can change both but not sure if it is needed.

PR #6780, opened 2017-05-25
Handle the possibility of HTTP/0.9 with a better error message

### LINKED ISSUE #6739: Handle missing version in HTTP request line (a.k.a HTTP/0.9)
I don't know if it's worth pursuing but I came yesterday across a network appliance that seemed to perform malformed HTTP based health checks against our Netty application.

The HTTP requests looked like this `GET /\r\n` without the version and [HttpVersion line 63](https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/HttpVersion.java#L63) was throwing Exceptions.

Turns out it's HTTP/0.9 as per [RFC 1945](https://www.ietf.org/rfc/rfc1945.txt) (see section 3.1)

> If the protocol version is not specified, the recipient must assume that the message is in the simple HTTP/0.9 format.

rkapsi:
Motivation

RFC 1945 (see section 3.1) says that request lines may not have a version in which case the request is assumed to be HTTP/0.9. We don't necessarily want to support that but the existing Exception should indicate the possibility of the request being HTTP/0.9 and give the user a chance to track it down.

Modifications

Indicate in the Exception's message that the request is possibly HTTP/0.9.

Result

Fixes #6739

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpVersion.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpVersion.java
@@ -60,7 +60,7 @@ public class HttpVersion implements Comparable<HttpVersion> {
         text = text.trim();
 
         if (text.isEmpty()) {
-            throw new IllegalArgumentException("text is empty");
+            throw new IllegalArgumentException("text is empty (possibly HTTP/0.9)");
         }
 
         // Try to match without convert to uppercase first as this is what 99% of all clients


normanmaurer:
Cherry-picked into 4.1 (b419bd152a7ea4946bff824ce1c6eff62d2c56ce) and 4.0 (0435d8410a)

Issue #6745, opened 2017-05-15
QueryStringDecoder does not decode correctly path part with '+' (plus) sign in it


croudet:
### Expected behavior
'+' is a valid char for path part of URI, it should not be replaced by space.

### Actual behavior

'+' is replaced with space.

### Minimal yet complete reproducer code (or URL to code)
```
public void testURIPathDecoding() throws URISyntaxException {
    String path = "/a/path/with/a+in.it";
    URI uri = new URI("http", "www.bbbb.com", path, null);
    System.out.println(path.equals(uri.getPath()));
    System.out.println(path.equals(uri.getRawPath()));
    QueryStringDecoder dec = new QueryStringDecoder(uri.getRawPath());
    System.out.println(dec.path());
    System.out.println(path.equals(dec.path()));
    System.out.println(path.equals(dec.uri()));
  }

true
true
/a/path/with/a in.it
false
true
```
### Netty version
 4.1.11
### OS version (e.g. `uname -a`)
Any

fenik17:
See https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.1
> **application/x-www-form-urlencoded**  
> This is the default content type. 
> ... Space characters are replaced by `+`...

And [Javadoc for QueryStringDecoder](https://github.com/netty/netty/blob/d675febf07d14d4dff82471829f974369705655a/codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java#L44) says that:

> This decoder can also decode the content of an HTTP POST request whose content type is **application/x-www-form-urlencoded**

croudet:
I am talking about the path part of an URL.
See grammar in  https://www.w3.org/Addressing/URL/url-spec.txt

```
httpaddress             h t t p :   / / hostport [  / path ] [ ? search ]   
path                    void |  segment  [  / path ] 
segment                 xpalphas
xpalpha                 xalpha | +   
xpalphas                xpalpha [ xpalphas ]   
```

fenik17:
@croudet yes, you are right. It's a bug..

fenik17:
A `QueryStringDecoder` also  doesn't cut a `fragment` from query string (as prescribed by [RFC 3986](http://www.faqs.org/rfcs/rfc3986.html))
```java
System.out.println(new QueryStringDecoder(new URI("p?a=b#anchor")).parameters().get("a").get(0));
System.out.println(new QueryStringDecoder("p?a=b#anchor").parameters().get("a").get(0));
[...]

PR #6781, opened 2017-05-25
Fixes in QueryStringDecoder

fenik17:
Motivation:

QueryStringDecoder has several problems:
- doesn't decode correctly path part with `+` (plus) sign in it,
- doesn't cut a `fragment` (after `#`) from query string (see RFC 3986),
- doesn't work correctly with encoding,
- treat `%%` as a percent character escaping (it's don't described in RFC).

Modifications:

- leave `+` chars in a `path` part of uri string,
- ignore `fragment` part (after `#`),
- correctly work with encoding,
- don't treat `%%` as escaping for the `%` (to match the javadoc and RFC).

Result:

Fixed issues from #6745.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
@@ -19,16 +19,20 @@ import io.netty.util.CharsetUtil;
 
 import java.net.URI;
 import java.net.URLDecoder;
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.CharacterCodingException;
 import java.nio.charset.Charset;
+import java.nio.charset.CharsetDecoder;
+import java.nio.charset.CoderResult;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
-import static io.netty.util.internal.ObjectUtil.checkNotNull;
-import static io.netty.util.internal.ObjectUtil.checkPositive;
-import static io.netty.util.internal.StringUtil.EMPTY_STRING;
+import static io.netty.util.internal.ObjectUtil.*;
+import static io.netty.util.internal.StringUtil.*;
 
 /**
  * Splits an HTTP query string into a path string and key-value parameter pairs.
@@ -63,11 +67,10 @@ public class QueryStringDecoder {
 
     private final Charset charset;
     private final String uri;
-    private final boolean hasPath;
     private final int maxParams;
+    private int pathEndIdx;
     private String path;
     private Map<String, List<String>> params;
-    private int nParams;
 
     /**
      * Creates a new decoder that decodes the specified URI. The decoder will
@@ -106,14 +109,12 @@ public class QueryStringDecoder {
      * specified charset.
      */
     public QueryStringDecoder(String uri, Charset charset, boolean hasPath, int maxParams) {
-        checkNotNull(uri, "uri");
-        checkNotNull(charset, "charset");
-        checkPositive(maxParams, "maxParams");
+        this.uri = checkNotNull(uri, "uri");
+        this.charset = checkNotNull(charset, "charset");
+        this.maxParams = checkPositive(maxParams, "maxParams");
 
-        this.uri = uri;
-        this.charset = charset;
-        this.maxParams = maxParams;
-        this.hasPath = hasPath;
+        // `-1` means that path end index will be initialized lazily
+        pathEndIdx = hasPath ? -1 : 0;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
```diff
@@ -300,56 +282,96 @@ public static String decodeComponent(final String s, final Charset charset) {
...
-        final int size = s.length();
-        boolean modified = false;
-        for (int i = 0; i < size; i++) {
-            final char c = s.charAt(i);
-            if (c == '%' || c == '+') {
-                modified = true;
+        return decodeComponent(s, 0, s.length(), charset, false);
+    }
+
+    private static String decodeComponent(String s, int from, int toExcluded, Charset charset, boolean isPath) {
```
netkins: MAJOR The Cyclomatic Complexity of this method "decodeComponent" is 27 which is greater than 25 authorized. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AMethodCyclomaticComplexity)

codec-http/src/test/java/io/netty/handler/codec/http/QueryStringDecoderTest.java
```diff
@@ -179,9 +218,7 @@ public void testUrlDecoding() throws Exception {
                 final String decoded = QueryStringDecoder.decodeComponent(encoded);
                 Assert.assertEquals(expected, decoded);
             } catch (IllegalArgumentException e) {
-                Assert.assertTrue("String \"" + e.getMessage() + "\" does"
-                                  + " not contain \"" + expected + '"',
-                                  e.getMessage().contains(expected));
+                Assert.assertEquals(e.getMessage(), expected);
```
fenik17: oops.. will swap an args later.

fenik17: fixed

normanmaurer:
@Scottmitch PTAL as well

fenik17:
@normanmaurer thanks! I included [one more fix](https://github.com/netty/netty/issues/6745#issuecomment-304530375)... Found this when I thought about how to reduce the cyclomatic complexity, to which @netkins complained ))

codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
```diff
@@ -300,74 +291,108 @@ public static String decodeComponent(final String s, final Charset charset) {
...
-            return s;
+        if (firstEscaped == -1) {
+            return s.substring(from, toExcluded);
         }
-        final byte[] buf = new byte[size];
-        int pos = 0;  // position in `buf'.
-        for (int i = 0; i < size; i++) {
+
+        CharsetDecoder decoder = CharsetUtil.decoder(charset);
+        int decodedCapacity = len / 3;
```
Scottmitch: can you add a comment explaining `len / 3`? something to the effect of each encoded byte takes 3 characters.

codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
```diff
@@ -300,74 +291,108 @@ public static String decodeComponent(final String s, final Charset charset) {
...
+                }
+            } catch (CharacterCodingException ex) {
+                throw new IllegalStateException(ex);
             }
+            strBuf.append(charBuf.flip());
+        }
+        return strBuf.toString();
+    }
+
+    private static byte decodeHexByte(String s, int pos, int len) {
```
Scottmitch: reminder: consider using consistent approach between this class an [ByteBufUtil.java](https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java#L892) (pick the "better" of the two).

codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
fenik17: @Scottmitch I added the comment. Also, replaced
```java
int decodedCapacity = len / 3;
```
by
```java
int decodedCapacity = (toExcluded - firstEscaped) / 3;
```

codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
normanmaurer: @fenik17 did you see this ?

fenik17: @normanmaurer I'm not sure, is this a reminder for me or not? :)

Scottmitch: doesn't have to be addressed as part of this PR ... I meant to create an issue but got tracked. https://github.com/netty/netty/issues/6797

Scottmitch:
4.1 (270e9d66c) 4.0 (4189ce1b8)

Scottmitch:
thanks @fenik17 !

PR #6783, opened 2017-05-26
Correctly parse package-prefix if the jni lib is contained in a path …

normanmaurer:
…that matches the library name.

Motivation:

We used strstr to find the path to the library, which fails if the library is contained in a directory that also matches the library name.

Modifications:

- Introduce netty_unix_util_strstr_last which will return a pointer which points to the last accourance and so not fails if the direct also matches the library name.

Result:

Be able to load the library in all cases.

### DIFF
--- a/transport-native-unix-common/src/main/c/netty_unix_util.c
+++ b/transport-native-unix-common/src/main/c/netty_unix_util.c
@@ -51,8 +51,20 @@ char* netty_unix_util_rstrstr(char* s1rbegin, const char* s1rend, const char* s2
     return NULL;
 }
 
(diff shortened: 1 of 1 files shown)


normanmaurer:
@Scottmitch PTAL... this was ported from https://github.com/netty/netty-tcnative/pull/272

normanmaurer:
Let me just merge this in as you reviewed the code in netty-tcnative anyway as well.

normanmaurer:
Cherry-picked into 4.1 (0b0309624a5c13f6059a647a73b3c6d9799c604f) and 4.0 (8a0a6b944f)

PR #6787, opened 2017-05-27
RedisDecoder infinite loop

Scottmitch:
Motivation:
RedisDecoder can get into an infinite loop while decoding bulk strings if the final \r and \n to indicate the end of content are split on ByteBuf boundaries.

Modifications:
- We should break out of the decode loop if remainingBulkLength is 0 and we don't have enough data to read EOL

Result:
No more infinite loop in RedisDecoder#decodeBulkStringContent.

### DIFF
--- a/codec-redis/src/test/java/io/netty/handler/codec/redis/RedisDecoderTest.java
+++ b/codec-redis/src/test/java/io/netty/handler/codec/redis/RedisDecoderTest.java
@@ -27,9 +27,15 @@ import org.junit.Test;
 
 import java.util.List;
 
-import static io.netty.handler.codec.redis.RedisCodecTestUtil.*;
-import static org.hamcrest.CoreMatchers.*;
-import static org.junit.Assert.*;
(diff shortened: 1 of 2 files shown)


codec-redis/src/test/java/io/netty/handler/codec/redis/RedisDecoderTest.java
```diff
@@ -52,6 +58,16 @@ public void teardown() throws Exception {
     }
 
     @Test
+    public void splitEOLDoesNotInfiniteLoop() throws Exception {
+        channel.writeInbound(byteBufOf("$6\r\nfoobar\r"));
+        channel.writeInbound(byteBufOf("\n"));
```
normanmaurer: Add asserts to the methods above

Scottmitch: done

Scottmitch:
4.1 (742ee7642)

normanmaurer:
@Scottmitch thanks!

PR #6790, opened 2017-05-30
Improvement : "if" construction replaced with "switch" for ServerCookieDecoder

doom369:
Motivation:

Usually, cookies don't start with special chars, so `Switch` statement is more efficient for cookie decoder as it performs less operations.

Modification:

`If` statement replaced with `switch`. Added 2 more tests for corner cases.

Result:

Cookie decoding is a bit faster.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
@@ -87,10 +87,17 @@ public final class ServerCookieDecoder extends CookieDecoder {
                     break loop;
                 }
                 char c = header.charAt(i);
-                if (c == '\t' || c == '\n' || c == 0x0b || c == '\f'
-                        || c == '\r' || c == ' ' || c == ',' || c == ';') {
-                    i++;
-                    continue;
+                switch (c) {
+                    case '\t':
+                    case '\n':
+                    case 0x0b:
+                    case '\f':
+                    case '\r':
+                    case ' ':
+                    case ',':
+                    case ';':
+                        i++;
+                        continue;
                 }
                 break;
             }
--- a/codec-http/src/test/java/io/netty/handler/codec/http/cookie/ServerCookieDecoderTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/cookie/ServerCookieDecoderTest.java
@@ -34,6 +34,26 @@ public class ServerCookieDecoderTest {
     }
 
     @Test
+    public void testDecodingSingleCookieWithSemicolon() {
+        String cookieString = "myCookie=myValue;";
+        Set<Cookie> cookies = ServerCookieDecoder.STRICT.decode(cookieString);
+        assertEquals(1, cookies.size());
+        Cookie cookie = cookies.iterator().next();
+        assertNotNull(cookie);
+        assertEquals("myValue", cookie.value());
+    }
+
+    @Test
+    public void testDecodingSingleCookieWithStartNewLine() {
+        String cookieString = "\t\nmyCookie=myValue";
+        Set<Cookie> cookies = ServerCookieDecoder.STRICT.decode(cookieString);
+        assertEquals(1, cookies.size());
+        Cookie cookie = cookies.iterator().next();
+        assertNotNull(cookie);
+        assertEquals("myValue", cookie.value());
+    }
+
+    @Test
     public void testDecodingMultipleCookies() {
         String c1 = "myCookie=myValue;";
         String c2 = "myCookie2=myValue2;";


fenik17:
@doom369 consider using a `BitSet` as it done in the `CookieUtil`.

fenik17:
In any case, it would be nice to see benchmark results.

doom369:
@fenik17 switch is O(1) or O(log(n)) in the worst case. I don't see how bitset could be faster.

```
@Fork(1)
@State(Scope.Thread)
@Warmup(iterations = 10, time = 1, timeUnit = TimeUnit.SECONDS, batchSize = 1000)
@Measurement(iterations = 20, time = 1, timeUnit = TimeUnit.SECONDS, batchSize = 1000)
public class SwitchVsIfNettyCookie2 {

    char c;

    @Setup
    public void init() {
        c = 'm';
    }

    @Benchmark
    public boolean testSwitch() throws Exception {
        switch (c) {
            case '\t':
            case '\n':
            case 0x0b:
            case '\f':
            case '\r':
            case ' ':
            case ',':
            case ';':
                return true;
        }
        return false;
    }

    @Benchmark
    public boolean testIf() throws Exception {
        return c == '\t' || c == '\n' || c == 0x0b || c == '\f'
                || c == '\r' || c == ' ' || c == ',' || c == ';';
    }

}
```

```
# Run complete. Total time: 00:01:00

Benchmark                           Mode  Cnt       Score      Error  Units
SwitchVsIfNettyCookie2.testIf      thrpt   20  274212.590 ± 1010.933  ops/s
SwitchVsIfNettyCookie2.testSwitch  thrpt   20  347754.160 ± 2944.634  ops/s
```

slandelle:
> Cookie decoding is a bit faster.

I'm not sure you can draw this conclusion from your benchmark, as you're not testing the actual code, but a way shorter method. I think you'd need to test the 2 actual implementations.

Then, as @fenik17 said, it would be nice to compare with a BitSet based impl too.

fenik17:
I think, in this place one of the fastest approach would be to use a simple bit mask:

```java
    private static long mask = setBits(0, '\t', '\n', 0x0b, '\f', '\r', ' ', ',', ';');

    private static long setBits(long mask, int... bitPos) {
        for (int pos : bitPos) {
            assert pos < 64 && pos >= 0;
            mask |= (1L << pos);
        }
        return mask;
    }

    private static boolean isSetBit(long mask, int bitPos) {
        return bitPos < 64 && (mask & (1L << bitPos)) != 0;
    }
```

It has complexity O(1) in all cases and does not carry the overhead as in BitSet operations.

doom369:
@fenik17 no need to guess. We have JMH for that :).

@slandelle 
> as you're not testing the actual code

I'm testing a specific change I did. However, use case is real.

slandelle:
@doom369 The actual method that would uses this piece of code is way larger than what you've tested. I'm just saying there's a good chance the performance improvement you see in your test would be dwarfed/nullified once injected in the real code. In that case, IMHO, readability should prevail over perf improvement within error margin. If so, the switch version could be better. My2cents.

doom369:
@slandelle I understand. And yes, most probably those change will be dwarfed/nullified in full decoding flow as this is just the tiny micro-optimization. However, it is very simple and straightforward, took minutes to fix and review. And that's the reason I submitted it. 

I see many other places that will involve much discussisons and that's why I'm quiet there :).

codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
```diff
@@ -87,10 +87,17 @@ private ServerCookieDecoder(boolean strict) {
                     break loop;
                 }
                 char c = header.charAt(i);
-                if (c == '\t' || c == '\n' || c == 0x0b || c == '\f'
-                        || c == '\r' || c == ' ' || c == ',' || c == ';') {
-                    i++;
-                    continue;
+                switch (c) {
```
netkins: MAJOR Add a default case to this switch. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3ASwitchLastCaseIsDefaultCheck)

fenik17:
> no need to guess. We have JMH for that :).

np )
```java
Benchmark                     Mode  Cnt       Score      Error  Units
BitSetBenchmark.testBitMask  thrpt   20  447646,355 ± 4455,793  ops/s
BitSetBenchmark.testBitSet   thrpt   20  369771,190 ± 2982,951  ops/s
BitSetBenchmark.testIf       thrpt   20  339016,457 ± 2166,366  ops/s
BitSetBenchmark.testSwitch   thrpt   20  372332,634 ± 4399,075  ops/s
```
These results are still within the margin of error, in other test cases the picture can be different. I agree with @slandelle: `switch`s variant can be accepted as more readable.

<details>

<summary>Benchmarks code</summary>

```java
@Fork(1)
@State(Scope.Thread)
@Warmup(iterations = 10, time = 1, timeUnit = TimeUnit.SECONDS, batchSize = 1000)
@Measurement(iterations = 20, time = 1, timeUnit = TimeUnit.SECONDS, batchSize = 1000)
public class BitSetBenchmark extends AbstractMicrobenchmark {

    private static final BitSet bitSet = new BitSet(64);
    private static final long bitMask = setBits(0, '\t', '\n', 0x0b, '\f', '\r', ' ', ',', ';');

    static {
        for (char c : new char[] { '\t', '\n', 0x0b, '\f', '\r', ' ', ',', ';' }) {
            bitSet.set(c);
        }
    }

    private static long setBits(long mask, int... bitPos) {
        for (int pos : bitPos) {
            assert pos < 64 && pos >= 0;
            mask |= (1L << pos);
        }
        return mask;
    }

    private static boolean hasBit(long mask, int bitPos) {
        return bitPos < 64 && (mask & (1L << bitPos)) != 0;
    }

    private char c = 'm';

    @Setup
    public void init() {
        c = 'm';
    }

    @Benchmark
    public boolean testSwitch() {
        switch (c) {
        case '\t':
        case '\n':
        case 0x0b:
        case '\f':
        case '\r':
        case ' ':
        case ',':
        case ';':
            return true;
        }
        return false;
    }

    @Benchmark
    public boolean testIf() {
        return c == '\t' || c == '\n' || c == 0x0b || c == '\f'
               || c == '\r' || c == ' ' || c == ',' || c == ';';
    }

    @Benchmark
    public boolean testBitSet() {
        return bitSet.get(c);
    }

    @Benchmark
    public boolean testBitMask() {
        return hasBit(bitMask, c);
    }
}
```

</details>

Scottmitch:
@doom369 - Thanks for the PR. Can you please include a JMH benchmark which uses `ServerCookieDecoder` and include results before/after?

doom369:
@fenik17 thanks for benchmark!
@Scottmitch no difference on benchmarks for `ServerCookieDecoder`. So closing it. This change doesn't worth it. Thanks for your comments.

PR #6792, opened 2017-05-30
ByteBufUtil#decodeHexDump

Scottmitch:
Motivation:
ByteBufUtil provides a hexDump method. For debugging purposes it is often useful to decode that hex dump to get the original content, but no such method exists.

Modifications:
- Add ByteBufUtil#decodeHexDump

Result:
ByteBufUtil#decodeHexDump is available to make debugging easier.


buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -869,6 +877,25 @@ private static String hexDump(ByteBuf buffer, int fromIndex, int length) {
             return new String(buf);
         }
 
+        private static byte[] decodeHexDump(String hexDump, int fromIndex, int length) {
```
normanmaurer: Why not just change `String` to `CharSequence` ?

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -125,6 +126,13 @@ public static String hexDump(byte[] array, int fromIndex, int length) {
     }
 
     /**
+     * Decodes a string generated by {@link #hexDump(byte[])}
+     */
+    public static byte[] decodeHexDump(String hexDump) {
```
normanmaurer: Why not just change `String` to `CharSequence` ?

Scottmitch: The method was meant to be symmetrical with `hexDump` which produces a `String` ... but since the implementation doesn't depend on `String` I can change to `CharSequence`

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
Scottmitch: done

Scottmitch:
@normanmaurer - Comments addressed.

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
normanmaurer: +1...

buffer/src/test/java/io/netty/buffer/ByteBufUtilTest.java
```diff
@@ -23,13 +23,33 @@
...
+    public void decodeHexEvenLength() {
+        decodeHex(256);
+    }
+
+    @Test
+    public void decodeHexOddLength() {
+        decodeHex(257);
+    }
+
+    private void decodeHex(int len) {
```
normanmaurer: static ?

Scottmitch:
@normanmaurer - 2nd round addressed ... anything else?

normanmaurer:
@Scottmitch ship it

Scottmitch:
4.1 (b71abcedd)

Issue #6791, opened 2017-05-30
io.netty.internal.tcnative.* classes bundled with netty-all JAR


rkapsi:
I believe none of the `io.netty.internal.tcnative.*` classes is supposed to be bundled with the netty-all JAR but they are...

```bash
$ unzip netty-all-4.1.11.Final.jar | grep tcnative
   creating: io/netty/internal/tcnative/
  inflating: io/netty/internal/tcnative/SSL.class  
  inflating: io/netty/internal/tcnative/Buffer.class  
  inflating: io/netty/internal/tcnative/NativeStaticallyReferencedJniMethods.class  
  inflating: io/netty/internal/tcnative/SSLContext.class  
  inflating: io/netty/internal/tcnative/CertificateRequestedCallback.class  
  inflating: io/netty/internal/tcnative/SniHostNameMatcher.class  
  inflating: io/netty/internal/tcnative/Library.class  
  inflating: io/netty/internal/tcnative/SessionTicketKey.class  
  inflating: io/netty/internal/tcnative/CertificateRequestedCallback$KeyMaterial.class  
  inflating: io/netty/internal/tcnative/CertificateVerifier.class 
```
... causing classpath problems if one is trying to use the netty-all JAR along with a custom TCN build.

normanmaurer:
@rkapsi will check

normanmaurer:
@rkapsi Please check https://github.com/netty/netty/pull/6793

PR #6793, opened 2017-05-30
Correctly exclude netty-tcnative classes from the all jar.

normanmaurer:
Motivation:

We need to exclude the netty-tcnative classes from the all jar when building it as otherwise there may be classloader issues.

Modifications:

Add io/netty/internal/tcnative to the exclude list.

Result:

Fixes [#6791]

### DIFF
--- a/all/pom.xml
+++ b/all/pom.xml
@@ -551,7 +551,7 @@
               <goal>unpack-dependencies</goal>
             </goals>
             <configuration>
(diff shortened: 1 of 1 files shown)


normanmaurer:
@rkapsi PTAL

rkapsi:
@normanmaurer LGTM

Scottmitch:
4.1 (1504abd47) 4.0 (62d9b5b22)

normanmaurer:
@Scottmitch thanks for cherry-picking :) Please remember to use 4.0.x as milestone if you also cherry-pick into 4.0

PR #6803, opened 2017-06-01
OpenSslEngine return NEED_WRAP if the destination buffered filled

### LINKED ISSUE #6796: Recent commit breaks Mutual TLS
Commit f20063d26b640cb4121e092b5a3a0bb4c0a10eb5 appears to break Mutual TLS in the sense that the connection just hangs. There're no Exceptions as far as I can tell and the only failure I'm getting is when the events in [SslHandler#channelInactive(...)](https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/SslHandler.java#L898) fire.

Interestingly, the problem occurs only in our production environment (which has more load, latencies and RTTs between servers). The problem doesn't occur on my dev workstation (i.e. unable to reproduce in the form of a unit test).

Reverting the commit fixes the problem for me.

### Netty version

4.1.12.Final-SNAPSHOT w/ TCN 2.0.2.Final-SNAPSHOT (openssl-static)

Scottmitch:
Motivation:
If the destination buffer is completely filled during a call to OpenSslEngine#wrap(..) we may return NEED_UNWRAP because there is no data pending in the SSL buffers. However during a handshake if the SSL buffers were just drained, and filled up the destination buffer it is possible OpenSSL may produce more data on the next call to SSL_write. This means we should keep trying to call SSL_write as long as the destination buffer is filled and only return NEED_UNWRAP when the destination buffer is not full and there is no data pending in OpenSSL's buffers.

Modifications:
- If the handshake produces data in OpenSslEngine#wrap(..) we should return NEED_WRAP if the destination buffer is completely filled

Result:
OpenSslEngine returns the correct handshake status from wrap().
Fixes #6796.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -571,8 +571,11 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
                     bytesProduced = bioLengthBefore - SSL.bioLengthByteBuffer(networkBIO);
 
                     if (bytesProduced > 0) {
-                        // It's important we call this before wrapStatus() as wrapStatus() may shutdown the engine.
+                        // If we have filled up the dst buffer and we have not finished the handshake we should try to
+                        // wrap again. Otherwise we should only try to wrap again if there is still data pending in
+                        // SSL buffers.
                         return newResult(mayFinishHandshake(status != FINISHED ?
+                                         bytesProduced == bioLengthBefore ? NEED_WRAP :
                                          getHandshakeStatus(SSL.bioLengthNonApplication(networkBIO)) : FINISHED),
                                          0, bytesProduced);
                     }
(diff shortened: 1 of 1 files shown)


handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -779,8 +780,13 @@ private void wrapNonAppData(ChannelHandlerContext ctx, boolean inUnwrap) throws
                     // If this is not enough we will increase the buffer in wrap(...).
                     out = allocateOutNetBuf(ctx, 2048, 1);
                 }
+
                 SSLEngineResult result = wrap(alloc, engine, Unpooled.EMPTY_BUFFER, out);
 
+                if (previousNeedUnwrap) {
```
Scottmitch: changes to SslHandler are just to support debugging and will be removed if this PR is accepted.

Scottmitch:
@rkapsi - would you mind trying this PR with OPENSSL and report back the same log statements you did as in https://github.com/netty/netty/issues/6796#issuecomment-305486288?

rkapsi:
@Scottmitch - will do first thing in the morning tomorrow

Scottmitch:
@rkapsi verified this patch works as expected in https://github.com/netty/netty/issues/6796#issuecomment-305766517

Scottmitch:
@rkapsi - Thanks for verifying! I reverted the debug code ... would you mind verifying that https://github.com/netty/netty/issues/6796 is fixed without any changes to `SslHandler`?

Scottmitch:
4.1 (24f801c7d) 4.0 (77bc0f2f3)

Scottmitch:
I will pull this in now as even if it doesn't fix #6796 it is still an issue, and the behavior from the log statements provided in https://github.com/netty/netty/issues/6796#issuecomment-305766517 demonstrates the correct behavior.

rkapsi:
@Scottmitch running 24f801c7d1d1e66193d6ff0da3b93ce427b2b329 built from HEAD of 4.1 now. Looks good, no issues. :100:

Issue #6578, opened 2017-03-28
Strange pauses before handling Http2 Header Frames in 4.1.9


nhnFreespirit:
After upgrading From Netty 4.1.8 to 4.1.9 we are seeing strange pauses in our TSL HTTP/2 server, specifically it seems to have long (several seconds) pause before processing header frames in some cases.

I have bisected this with a local build of Netty, and that pointed to this commit as being the cause https://github.com/netty/netty/commit/6353c229fd11c6e6306011d55993775f7685d122

I do not have much in the form of a standalone repro right now as we don't really understand what is happening yet, but we are actively investigating and wanted to hear if anyone has any ideas?

Will update this ticket as we learn more.

nhnFreespirit:
I should probably add that the current version of the server is using the "old" Netty HTTP/2 API to build our pipeline.

normanmaurer:
@Scottmitch can you please have a look ?

Scottmitch:
@nhnFreespirit - Are you doing renegotiations, and is this pause related to handshake/setup time? Or does this pause occur after the handshake is setup and you are sending application data?

fsareshwala:
@Scottmitch I looked at this problem briefly today with Wireshark and it looks like the pause is after the handshake is complete and application data has started being sent. I am currently working on decrypting the data being sent back and forth between our client and server so that I can see where exactly things are getting blocked. I should have more information regarding that tomorrow.

Scottmitch:
@fsareshwala - Thanks for the update. Keep me posted. It would be nice to keep some of this optimization but if necessary we can also revert it.

normanmaurer:
@Scottmitch +1

fsareshwala:
I have a bit more information to share. We were able to decrypt the data going being transferred between our services and see this in Wireshark:

```
```

fsareshwala:
I have a bit more information to share. We were able to decrypt the data going being transferred between our services.

In the Wireshark trace below, we have two services communicating with each other. The netty server is listening on port `8443` and the ssl client is using port `55507`. Packet 75 is where the client sends its request. It receives no response from the netty server for about 9 seconds at which point the client sends a `PING` request. Receiving no response from the netty server for another 0.5 seconds, the client closes the connection.

```
#   timestamp   src port    dst port    protocol    size    summary
15  2.331795    55507       8443        TLSv1.2     159     Client Hello
[...]

PR #6806, opened 2017-06-02
Revert "Revert "SslHandler avoid calling wrap/unwrap when unnecessary""

Scottmitch:
Motivation:
PR https://github.com/netty/netty/pull/6803 corrected an error in the return status of the OpenSslEngine. We should now be able to restore the SslHandler optimization.

Modifications:
- This reverts commit 7f3b75a5091dcd6d882102fdb92daa6931e02c30.

Result:
SslHandler optimization is restored.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -766,7 +766,7 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
      * {@link #setHandshakeFailure(ChannelHandlerContext, Throwable)}.
      * @return {@code true} if this method ends on {@link SSLEngineResult.HandshakeStatus#NOT_HANDSHAKING}.
      */
-    private void wrapNonAppData(ChannelHandlerContext ctx, boolean inUnwrap) throws SSLException {
+    private boolean wrapNonAppData(ChannelHandlerContext ctx, boolean inUnwrap) throws SSLException {
         ByteBuf out = null;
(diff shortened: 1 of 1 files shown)


Scottmitch:
@fsareshwala @nhnFreespirit - Please evaluate as requested in https://github.com/netty/netty/issues/6578#issuecomment-305841065

normanmaurer:
@nhnFreespirit ping...

nhnFreespirit:
Hey.

Sorry for leaving you hanging like this but I had to go do something completely different for a while. I should still have a local (if still somewhat complex) test case for this lying around somewhere, so let me see if I can get around to setting that up and running it today.

nhnFreespirit:
Getting my test case back up and running was a bit more involved than anticipated as our codebase have moved on quite a bit since last time. I did not see any issues when running the code from the pr/6806 branch.

That said, I also failed at reproducing the actual problem (by reverting 24f801c7d1d1e66193d6ff0da3b93ce427b2b329 as well) so it is possible that other factors make the original problem no longer appear...

I'm fine with merging this. I know what symptoms to look out for if the issue should re-appear and what to test if it does.

Scottmitch:
Thanks @nhnFreespirit ... I plan to merge this after 4.1.13 is released. It would be great if you could try a snapshot which includes this PR at that time.

Scottmitch:
4.1 (81fb2eede) 4.0 (623cb8d95)

Scottmitch:
@nhnFreespirit - Please report back if you see issues in the next release.

PR #6809, opened 2017-06-02
Adjust Content-Length header when encoding Full Responses

### LINKED ISSUE #6808: HttpContentCompressor/HttpContentEncoder should reset the content-length header for FullHttpMessages
### Expected behavior
When the `HttpContentEncoder` is tasked with compressing a full `HttpMessage` that already contains a `Content-Length` header it should perform the compression and _adjust_ the `Content-Length`.

### Actual behavior
All messages compressed result in `Transfer-Encoding: Chunked` and don't have a `Content-Length` header.

```java
@Test
    public void testAmendContentLength() throws Exception {
        EmbeddedChannel ch = new EmbeddedChannel(new HttpContentCompressor());
        ch.writeInbound(newRequest());

        ByteBuf content = Unpooled.copiedBuffer("Hello, world!", CharsetUtil.UTF_8);
        HttpMessage res = new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.OK, content);
        HttpUtil.setContentLength(res, content.readableBytes());

        ch.writeOneOutbound(res);

        HttpMessage compressedResponse = ch.readOutbound();
        assertThat(res.headers().get(HttpHeaderNames.CONTENT_ENCODING), is("gzip"));
        assertTrue(HttpUtil.isContentLengthSet(compressedResponse));
    }
```

### Netty version 4.1 branch at SHA 24f801c7d1d1 (should affect most versions)

bryce-anderson:
Motivation:
If a full HttpResponse with a Content-Length header is encoded by the HttpContentEncoder subtypes the Content-Length header is removed and the message is set to Transfer-Encoder: chunked. This is an unnecessary loss of information about the message content.

Modifications:
- If a full HttpResponse has a Content-Length header, the header is adjusted after encoding instead of converting the message to Chunked-Transfer encoding.

Result:
Fixes #6808.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentEncoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentEncoder.java
@@ -164,18 +164,21 @@ public abstract class HttpContentEncoder extends MessageToMessageCodec<HttpReque
                 // so that the message looks like a decoded message.
                 res.headers().set(HttpHeaderNames.CONTENT_ENCODING, result.targetContentEncoding());
 
-                // Make the response chunked to simplify content transformation.
-                res.headers().remove(HttpHeaderNames.CONTENT_LENGTH);
-                res.headers().set(HttpHeaderNames.TRANSFER_ENCODING, HttpHeaderValues.CHUNKED);
-
                 // Output the rewritten response.
                 if (isFull) {
                     // Convert full message into unfull one.
                     HttpResponse newRes = new DefaultHttpResponse(res.protocolVersion(), res.status());
                     newRes.headers().set(res.headers());
                     out.add(newRes);
-                    // Fall through to encode the content of the full response.
+
+                    ensureContent(res);
+                    encodeFullResponse(newRes, (HttpContent) res, out);
+                    break;
                 } else {
+                    // Make the response chunked to simplify content transformation.
+                    res.headers().remove(HttpHeaderNames.CONTENT_LENGTH);
+                    res.headers().set(HttpHeaderNames.TRANSFER_ENCODING, HttpHeaderValues.CHUNKED);
+
                     out.add(res);
                     state = State.AWAIT_CONTENT;
                     if (!(msg instanceof HttpContent)) {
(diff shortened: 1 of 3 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/HttpContentEncoder.java
```diff
@@ -205,6 +208,25 @@ protected void encode(ChannelHandlerContext ctx, HttpObject msg, List<Object> ou
         }
     }
 
+    private void encodeFullResponse(HttpResponse newRes, HttpContent content, List<Object> out) {
+        int existingMessages = out.size();
+        assert encodeContent(content, out);
```
fenik17: in runtime an `assert` block is ignoring if `-ea` option has not specified.

Scottmitch: We don't want this to be an `assert`, right? `assert` statements can be disabled via command line and optimized away by JIT.

codec-http/src/test/java/io/netty/handler/codec/http/HttpContentCompressorTest.java
```diff
@@ -198,14 +198,54 @@ public void testChunkedContentWithTrailingHeader() throws Exception {
...
 
-        FullHttpResponse res = new DefaultFullHttpResponse(
+        FullHttpResponse fullRes = new DefaultFullHttpResponse(
                 HttpVersion.HTTP_1_1, HttpResponseStatus.OK,
                 Unpooled.copiedBuffer("Hello, World", CharsetUtil.US_ASCII));
-        res.headers().set(HttpHeaderNames.CONTENT_LENGTH, res.content().readableBytes());
+        fullRes.headers().set(HttpHeaderNames.CONTENT_LENGTH, fullRes.content().readableBytes());
+        ch.writeOutbound(fullRes);
+
+        Object o = ch.readOutbound();
```
Scottmitch: consider just directly casting to the object you expect like the other tests ... a class cast exception will be thrown if the expected type is not read.

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentEncoder.java
bryce-anderson: Removed.

codec-http/src/test/java/io/netty/handler/codec/http/HttpContentCompressorTest.java
bryce-anderson: Done.

normanmaurer:
@Scottmitch yeah LGTM... @bryce-anderson can you please squash ? After this I will pull in

bryce-anderson:
Squished. Thanks for the reviews, @fenik17, @Scottmitch and @normanmaurer.

normanmaurer:
Cherry-picked into 4.0 (729796687f) and 4.1 (9fa3e556f3a223c366f5ee7f706dd9645a06e196)

bryce-anderson:
@normanmaurer, @Scottmitch  I just occurred to me: do any of the content encoders emit a raw `ByteBuf`? If so, this PR will break those cases since they won't be included in the adjusted content-length header. It looks like the HttpObjectEncoder is happy to encode a `ByteBuf`, so I think I may have just introduced a bug, which I'm happy to fix.

bryce-anderson:
Okay, looks like my worries were unfounded: all content emitted from the compressors is wrapped in `DefaultHttpContent` in the `fetchEncoderOutput` method. Sorry for the noise.

PR #6811, opened 2017-06-03
Change QueryStringDecoder to use ByteBufUtil.decodeHexDump

### LINKED ISSUE #6797: QueryStringDecoder and ByteBufUtil decoding hex string duplication
QueryStringDecoder has methods which deal with decoding hex into bytes (e.g. `decodeHexByte`, `decodeHexNibble`) and `ByteBufUtil` has `decodeHexDump`. Can we share code between these two classes?

sunrenjie:
Motivations:
1. There are duplicated implementations of decoding hex strings. #6797
2. ByteBufUtil.HexUtil.decodeHexDump does not handle substring start
index properly. In addition, there is not enough protection against
illegal arguments.

Modifications:
1. ByteBufUtil.HexUtil.decodeHexDump is enhanced.
2. Change QueryStringDecoder to use ByteBufUtil.decodeHexDump and remove
its own implementation.
3. Update tests accordingly.

Result:
Fixed #6797 in real sense and made hex decoding function more robust.


sunrenjie:
@normanmaurer @Scottmitch PTAL

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -878,19 +885,24 @@ private static String hexDump(ByteBuf buffer, int fromIndex, int length) {
...
             }
 
-            int endIndex = fromIndex + length - 1;
             byte[] bytes = new byte[length >>> 1];
 
-            for (; fromIndex < endIndex; fromIndex += 2) {
-                bytes[fromIndex >>> 1] = (byte) ((Character.digit(hexDump.charAt(fromIndex), 16) << 4) +
-                                                  Character.digit(hexDump.charAt(fromIndex + 1), 16));
+            for (int i = 0; i < length; i += 2) {
+                int d1 = Character.digit(hexDump.charAt(fromIndex + i), 16);
```
fenik17: In my tests, a `QueryStringDecoder.decodeHexNibble()` faster than `Character.digit()` by ~30%.
Also `Character.digit()` supports some exotic cases:
```java
Character.digit('\uFF21', 16) => 10 // 'Ａ'
vs normal
Character.digit('\u0041', 16) => 10 // 'A'
```
Not sure that this is correctly for hex-encoding.

codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
```diff
@@ -330,7 +331,10 @@ private static String decodeComponent(String s, int from, int toExcluded, Charse
 
             byteBuf.clear();
             do {
-                byteBuf.put(decodeHexByte(s, i, toExcluded));
+                if (i + 3 > toExcluded) {
+                    throw new IllegalArgumentException("unterminated escape sequence at index " + i + " of: " + s);
+                }
+                byteBuf.put(decodeHexDump(s, i + 1, 2));
```
fenik17: It's too wasteful to allocate new byte array on each encoded byte...

Scottmitch: this will incur an intermediate `byte[]` array and copy operation. We consider creating a variant of `decodeHexDump` which can directly write into a `ByteBuf` ... something like:

```java
public static void decodHexDump(CharSequence hexDump, int start, int len, ByteBuf result) {
...
}
```

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
Scottmitch: +1 ... I have the same comment below.

codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
Scottmitch: +1 ... I have the same comment below.

fenik17: > directly write into a ByteBuf

Or maybe just a function to deconing one byte?

Scottmitch: I guess either would work in this cause however writing into a `ByteBuf` may be useful if users want to avoid the `byte[]` translation all together ... and this way users don't *have* to know how to properly iterate `hexDump` and the required characteristics.

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
sunrenjie: @fenik17 Yes, the docs explicitly states so. QueryStringDecoder.decodeHexNibble() is simple and fits our purposes better, I think.

codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
```diff
@@ -354,40 +358,7 @@ private static String decodeComponent(String s, int from, int toExcluded, Charse
...
-    }
-
-    /**
-     * Helper to decode half of a hexadecimal number from a string.
-     * @param c The ASCII character of the hexadecimal number to decode.
-     * Must be in the range {@code [0-9a-fA-F]}.
-     * @return The hexadecimal value represented in the ASCII character
-     * given, or {@code -1} if the character is invalid.
-     */
-    private static int decodeHexNibble(final char c) {
```
sunrenjie: @Scottmitch
Test result given by @fenik17 above shows that "QueryStringDecoder.decodeHexNibble() faster than Character.digit() by ~30%." This agrees with the fact that Character.digit() tries to address a larger set of letters (ASCII latin+full-width latin; not exactly what we'd expect in escaped URL strings anyway) with a much more complex algorithm, etc.

So, let's, say, continue to use decodeHexNibble and drop this PR?

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
sunrenjie: @Scottmitch If I call decodeHexDump() with fromIndex=1000, length=2, "bytes[fromIndex >>> 1] = ... "will be referencing bytes[500] while the bytes array is of length 1.

codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
fenik17: Ok, it makes sense. Just one moment: here `ByteBuffer`, not `ByteBuf`.

codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
fenik17: > and drop this PR?

No.. Just use `decodeHexNibble()` instead of `Character.digit(x, 16)`.

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
Scottmitch: makes sense ... that was a bug ... previously we were only decoding from 0 so this went unnoticed.

codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
Scottmitch: @fenik17 - Are you suggesting adding a utility method for `ByteBuffer` instead of `ByteBuf`?

codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
Scottmitch: Move `decodeHexNibble` into `ByteBufUtil` and make it `private static` and use it instead of `Character.digit`

codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
fenik17: @Scottmitch I just pointing that in `QueryStringDecoder.decodeComponent()` used a `ByteBuffer`. If we want to use there the method you proposed, will have to make a variant with `ByteBuffer`.

Scottmitch: Oh I see what you are saying now ... makes sense. Not ideal to add a `ByteBuffer` method in `ByteBufUtil` ... maybe we should change `QueryStringDecoder` to use a `ByteBuf` (so we can take advantage of our allocators) and then use `ByteBuf#nioBuffer()` to do the decoding.

fenik17: Use `ByteBuf#nioBuffer()` will be too expensive. It smells of overengineering )
In `QueryStringDecoder.decodeComponent()` needed only decode two characters to one byte...

Scottmitch: > Use ByteBuf#nioBuffer() will be too expensive. It smells of overengineering )

Perhaps ... if it is too much I'm not against directly adding a `ByteBuffer` utility method.

normanmaurer: I agree with @fenik17 lets keep it simple...

sunrenjie:
@fenik17 @Scottmitch @normanmaurer 
Hi, PTAL.
I took the conservative approach and didn't introduce a new version of decodeHexDump() with write-back to ByteBuf/ByteBuffer, etc. Instead, it is merely corrected and kept there. Maybe we could wait for it to become useful in other ways and let it evolve, then.

Scottmitch:
4.1 (629b83e0a) 4.0 (6096f7820)

Scottmitch:
thanks @sunrenjie !

PR #6814, opened 2017-06-05
Allow in RFC2965 cookies comma as a cookies separator

alobodzki:
Motivation:

To be more complient with RFC2965, which is being partially supported
within the decoder, the code should also allow comma to be a cookies separator:

"Note: For backward compatibility, the separator in the Cookie header
   is semi-colon (;) everywhere.  A server SHOULD also accept comma (,)
   as the separator between cookie-values for future compatibility."

Modifications:

In tokenizing part of a decoder additional logic condition added, that if
the cookie has RFC2965 compliant format, then comma is also considered
as a cookies separator.

Result:

In case the cookie has RFC2965 format comma will be allowed as
a cookies separator.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
@@ -75,7 +75,7 @@ public final class ServerCookieDecoder extends CookieDecoder {
         boolean rfc2965Style = false;
         if (header.regionMatches(true, 0, RFC2965_VERSION, 0, RFC2965_VERSION.length())) {
             // RFC 2965 style cookie, move to after version value
-            i = header.indexOf(';') + 1;
+            i = indexOfLegacyDelimiter(header, 0) + 1;
             rfc2965Style = true;
         }
 
@@ -103,7 +103,7 @@ public final class ServerCookieDecoder extends CookieDecoder {
             for (;;) {
 
                 char curChar = header.charAt(i);
-                if (curChar == ';') {
+                if (curChar == ';' || (rfc2965Style && curChar == ',')) {
                     // NAME; (no value till ';')
                     nameEnd = i;
                     valueBegin = valueEnd = -1;
@@ -121,8 +121,14 @@ public final class ServerCookieDecoder extends CookieDecoder {
 
                     valueBegin = i;
                     // NAME=VALUE;
-                    int semiPos = header.indexOf(';', i);
-                    valueEnd = i = semiPos > 0 ? semiPos : headerLen;
+                    int delimiterPos;
+                    if (rfc2965Style) {
+                        delimiterPos = indexOfLegacyDelimiter(header, i);
+                    } else {
+                        delimiterPos = header.indexOf(';', i);
+                    }
+
+                    valueEnd = i = delimiterPos > 0 ? delimiterPos : headerLen;
                     break;
                 } else {
                     i++;
@@ -151,5 +157,15 @@ public final class ServerCookieDecoder extends CookieDecoder {
         }
 
         return cookies;
+    }
+
+    private static int indexOfLegacyDelimiter(String header, int offset) {
+        for (int pos = offset; pos < header.length(); pos++) {
+            char c = header.charAt(pos);
+            if (c == ',' || c == ';') {
+                return pos;
+            }
+        }
+        return -1;
     }
 }
--- a/codec-http/src/test/java/io/netty/handler/codec/http/cookie/ServerCookieDecoderTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/cookie/ServerCookieDecoderTest.java
@@ -172,6 +172,48 @@ public class ServerCookieDecoderTest {
     }
 
     @Test
+    public void testDecodingOldRFC2965CookiesWithCommaSeparatorAfterVersion() {
+        String source = "$Version=\"1\", " +
+                "Part_Number1=\"Riding_Rocket_0023\"; $Path=\"/acme/ammo\", " +
+                "Part_Number2=\"Rocket_Launcher_0001\"; $Path=\"/acme\"";
+
+        Set<Cookie> cookies = ServerCookieDecoder.STRICT.decode(source);
+        Iterator<Cookie> it = cookies.iterator();
+        Cookie c;
+
+        c = it.next();
+        assertEquals("Part_Number1", c.name());
+        assertEquals("Riding_Rocket_0023", c.value());
+
+        c = it.next();
+        assertEquals("Part_Number2", c.name());
+        assertEquals("Rocket_Launcher_0001", c.value());
+
+        assertFalse(it.hasNext());
+    }
+
+    @Test
+    public void testDecodingOldRFC2965CookiesWithCommaSeparatorBetweenCookies() {
+        String source = "$Version=\"1\"; " +
+                "Part_Number1=\"Riding_Rocket_0023\"; $Path=\"/acme/ammo\", " +
+                "Part_Number2=\"Rocket_Launcher_0001\"; $Path=\"/acme\"";
+
+        Set<Cookie> cookies = ServerCookieDecoder.STRICT.decode(source);
+        Iterator<Cookie> it = cookies.iterator();
+        Cookie c;
+
+        c = it.next();
+        assertEquals("Part_Number1", c.name());
+        assertEquals("Riding_Rocket_0023", c.value());
+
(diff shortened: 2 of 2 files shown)


slandelle:
RFC2965 has been deprecated for a very long time by RFC6265.
IMHO, upgrading servers and user agents would be a better solution than trying to support dead standards.
Moreover, one of the goals of the new cookie support implementation that was introduced last year was to move from RFC2965 to RFC6265.

Scottmitch:
I agree with @slandelle ... my preference would be to just focus on RFCs which have not been rendered obsolete if possible.

alobodzki:
@slandelle @Scottmitch of course, but there is a partial support in StyxCookieDecoder for rfc2965, so let's just remove it, at least it the intention will be clear, that only rfc6265 is supported. Or split it into two separate handlers. If not, I would go with this change.

Scottmitch:
What is `StyxCookieDecoder` and why can't this be updated to use the RFC6265 focused cookie code in Netty? Unfortunately we can't remove code without a major version bump, and this is why the old cookie code has been deprecated and references the new cookie code.

slandelle:
I think that's some autocompletion failure and @alobodzki refers to `ServerCookieDecoder` that still has some RFC2965 compatibility code for backward compatibility reasons (see `RFC2965_XXX` constants and such).

alobodzki:
@Scottmitch Yes, @slandelle is correct.

codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
```diff
@@ -121,8 +121,14 @@ private ServerCookieDecoder(boolean strict) {
 
                     valueBegin = i;
                     // NAME=VALUE;
-                    int semiPos = header.indexOf(';', i);
-                    valueEnd = i = semiPos > 0 ? semiPos : headerLen;
+                    int delimiterPos = header.indexOf(';', i);
+                    if (rfc2965Style) {
```
Scottmitch: if `rfc2965Style` should we just do a loop which searches for the first index of `;` or `,` to avoid 2 `indexOf` calls and iterations (which may not find a `,` character)

alobodzki:
@Scottmitch Sorry for a delay, I updated code to address your changes.

codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
```diff
@@ -75,7 +75,7 @@ private ServerCookieDecoder(boolean strict) {
         boolean rfc2965Style = false;
         if (header.regionMatches(true, 0, RFC2965_VERSION, 0, RFC2965_VERSION.length())) {
             // RFC 2965 style cookie, move to after version value
-            i = header.indexOf(';') + 1;
+            i = getDelimiterPos(header, 0) + 1;
```
fenik17: `getDelimiterPos(header, 0)` -> `getDelimiterPos(header, RFC2965_VERSION.length())` ?

fenik17: Also, if the delimiter is not found, the decoder will return the `$Version`->'1'? Is it's fine?

codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
```diff
@@ -152,4 +158,15 @@ private ServerCookieDecoder(boolean strict) {
 
         return cookies;
     }
+
+    private int getDelimiterPos(String header, int i) {
```
fenik17: static

codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
```diff
@@ -152,4 +158,15 @@ private ServerCookieDecoder(boolean strict) {
...
+        int delimiterPos = -1;
+        for (int j = i; j < header.length(); j++) {
+            if (header.charAt(j) == ',' || header.charAt(j) == ';') {
```
fenik17: cache the `header.charAt(j)` into variable.

codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
```diff
@@ -152,4 +158,15 @@ private ServerCookieDecoder(boolean strict) {
...
+                delimiterPos = j;
```
fenik17: A bit better `return j` here and `return -1` at the end.

codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
```diff
@@ -121,8 +121,14 @@ private ServerCookieDecoder(boolean strict) {
 
                     valueBegin = i;
                     // NAME=VALUE;
-                    int semiPos = header.indexOf(';', i);
-                    valueEnd = i = semiPos > 0 ? semiPos : headerLen;
+                    int delimiterPos;
+                    if (rfc2965Style) {
+                        delimiterPos = getDelimiterPos(header, i);
+                    } else {
```
netkins: MAJOR Refactor this code to not nest more than 5 if/for/while/switch/try statements. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS134)

codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
netkins: MINOR Make "getDelimiterPos" a "static" method. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2325)

slandelle:
@alobodzki Again, I think you should provide more information about what motivates your PR.

RFC2965 is super legacy, and the fact that there's still some partial support lurking around because it can't be removed until next major release doesn't feel to me like a good reason to improve it.

IMHO, the only valid motivation would be that there are actually a non negligible number of user-agents out there that only support RFC2965 and not RFC6265, a la Javascript framework "should be support IE5?".
On the contrary, if this would only benefit your organization because it has some legacy user-agents that can't be upgraded/managers won't spend a dime on, I don't think this should make it in Netty.

codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
alobodzki: getDelimiterPos is just like indexOf, second parameter is an offset, could you expand what you mean in your first question?
Second question coment: the implementation is doing stuff like before expanded by adding comma as a delimiter. getDelimiterPos returns -1 in case ;, is not found.

alobodzki:
@slandelle I strongy agree with you, about old legacy part, but the point is that the decoder code is providing partial implementation of that spec, which makes it worse, because people think that it is providing full support for a spec. Our people did think like that, I don't know what kind of clients they are using, that generate this version of cookies. Either the support for that old spec should be moved to some legacy decoder or removed completely. If it is removed we will have a justification to write our own decoder. For now for some cases it works, for some other cases it is splitting header string at wrong positions.

codec-http/src/main/java/io/netty/handler/codec/http/cookie/ServerCookieDecoder.java
fenik17: > what you mean in your first question?

This code work if `header` starts with the `$Version`:
```java
if (header.regionMatches(true, 0, RFC2965_VERSION, 0, RFC2965_VERSION.length()))
```
not needed to scan `header` from `0`, we can scan immediately from the `"$Version".length()`.

> Second question coment

This comment also is relevant for the current implementation. A `getDelimiterPos` returns `-1` if a `;` is not found, but then it's incremented: `i=-1+1`. Just try execute this:
```java
System.out.println(ServerCookieDecoder.STRICT.decode("$Version=1").size());
System.out.println(ServerCookieDecoder.STRICT.decode("$Version=1;").size());
```
It's looks like a bug,.

alobodzki:
@Scottmitch @slandelle I think I addressed the issues mentioned - is there anything more I can do for this PR?

alobodzki:
@Scottmitch did my commit addressed your change request?

normanmaurer:
Let's close this for now it seems like the RFC is obsolete and this is sitting here for a long time

Issue #6804, opened 2017-06-02
JDKZlibEncoder or JDKZlibDecoder can't work correct with large data


lsqlebai:
### Expected behavior
work 
### Actual behavior
decode error(protobuf decode error)
### Steps to reproduce
i have 2 server. and them communicate with tcp.
and the pipeline init code double with them.
when the send data > 16000 byte (maybe less). the decode side can't correct decode (protobuf decode error). 
### Minimal yet complete reproducer code (or URL to code)
decode error code:
Code:

        ChannelPipeline pipeline = ch.pipeline();
        pipeline.addLast(new ProtobufVarint32FrameDecoder());
        pipeline.addLast(ZlibCodecFactory.newZlibDecoder(ZlibWrapper.NONE));
        pipeline.addLast(new ProtobufDecoder(GameServer.ServerData.getDefaultInstance()));
        pipeline.addLast(new ProtobufVarint32LengthFieldPrepender());
        pipeline.addLast(ZlibCodecFactory.newZlibEncoder(ZlibWrapper.NONE));
        pipeline.addLast(new ProtobufEncoder());

i had 2 test:
     remove zlib, can be work.like this:
     
      ChannelPipeline pipeline = ch.pipeline();
        //receive
        pipeline.addLast(new ProtobufVarint32FrameDecoder());
        pipeline.addLast(new ProtobufDecoder(GameServer.ServerData.getDefaultInstance()));
        //send
        pipeline.addLast(new ProtobufVarint32LengthFieldPrepender());
        pipeline.addLast(new ProtobufEncoder());

change to the JZlibDecoder and JZlibEncoder can be work. like this:
        
        ChannelPipeline pipeline = ch.pipeline();
        //receive
        pipeline.addLast(new ProtobufVarint32FrameDecoder());
        pipeline.addLast(new JZlibDecoder(ZlibWrapper.NONE));
        pipeline.addLast(new ProtobufDecoder(GameServer.ServerData.getDefaultInstance()));
        //send
        pipeline.addLast(new ProtobufVarint32LengthFieldPrepender());
        pipeline.addLast(new JZlibEncoder(ZlibWrapper.NONE));
        pipeline.addLast(new ProtobufEncoder());
### Netty version
4.1.7
### JVM version (e.g. `java -version`)
java version "1.8.0_101"
### OS version (e.g. `uname -a`)
windows

normanmaurer:
[...]

PR #6816, opened 2017-06-05
JdkZlibDecoder and JZlibDecoder consistency

Scottmitch:
Motivation:
JdkZlibDecoder will allocate a new buffer when the previous buffer is filled with inflated data, but JZlibDecoder will attempt to use the same buffer by resizing. This leads to inconsistent results when these two decoders that are intended to be functionality equivalent.

Modifications:
- JdkZlibDecoder should attempt to resize and reuse the existing buffer instead of creating multiple buffers

Result:
Fixes https://github.com/netty/netty/issues/6804

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/compression/JZlibDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/compression/JZlibDecoder.java
@@ -107,16 +107,15 @@ public class JZlibDecoder extends ZlibDecoder {
                 z.next_in = array;
                 z.next_in_index = 0;
             }
-            int oldNextInIndex = z.next_in_index;
+            final int oldNextInIndex = z.next_in_index;
 
             // Configure output.
-            int maxOutputLength = inputLength << 1;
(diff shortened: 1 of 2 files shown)


normanmaurer:
@Scottmitch while I am not against the change its important to note that its an implementation detail and nothing the user should relay on IMHO

Scottmitch:
@normanmaurer - I agree its an implementation detail ... however I don't think it makes sense for these two decoders to have different behavior (from our perspective). `ProtobufDecoder#decoder` is a `MessageToMessageDecoder` and therefore doesn't have access to multiple `ByteBuf` objects at once so some additional aggregation would be necessary at this level. Also this PR potentially reduces the size on each buffer re-size in `JZlibDecoder` to what is necessary based on what has been decoded.

normanmaurer:
@Scottmitch ok sounds good... Just wanted to clarify that this behaviour could break at any point again as its an implementation detail so the user should not depend on it. Ship it

Scottmitch:
4.1 (e06cb82c4) 4.0 (19be3fe5f)

PR #6817, opened 2017-06-06
password and willMessage field types should be byte[] in codec.mqtt

### LINKED ISSUE #6750: codec.mqtt: wrong password and willMessage field types in MqttConnectPayload/MqttEncoder (should be byte[])

sbcd90:
Motivation:

Update the mqtt-codec based on mqtt spec (3.1.3.5).

Modification:

- Changes made to the file `MqttConnectPayload.java`.
- Subsequent changes have been made to files `MqttDecoder.java`, `MqttEncoder.java`, `MqttMessageBuilders.java`.
- Test cases have been updated.

Result:

Fixes #6750 .

### DIFF
--- a/codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttConnectPayload.java
+++ b/codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttConnectPayload.java
@@ -16,6 +16,7 @@
 
 package io.netty.handler.codec.mqtt;
 
+import io.netty.util.CharsetUtil;
 import io.netty.util.internal.StringUtil;
 
 /**
@@ -25,16 +26,34 @@ public final class MqttConnectPayload {
 
     private final String clientIdentifier;
     private final String willTopic;
-    private final String willMessage;
+    private final byte[] willMessage;
     private final String userName;
-    private final String password;
+    private final byte[] password;
 
+    /**
+     * @deprecated use {@link MqttConnectPayload#MqttConnectPayload(String, String, byte[], String, byte[])} instead
+     */
+    @Deprecated
     public MqttConnectPayload(
             String clientIdentifier,
             String willTopic,
             String willMessage,
             String userName,
             String password) {
+        this(
+          clientIdentifier,
+          willTopic,
+          willMessage.getBytes(CharsetUtil.UTF_8),
+          userName,
+          password.getBytes(CharsetUtil.UTF_8));
+    }
+
+    public MqttConnectPayload(
+            String clientIdentifier,
+            String willTopic,
+            byte[] willMessage,
+            String userName,
+            byte[] password) {
         this.clientIdentifier = clientIdentifier;
         this.willTopic = willTopic;
         this.willMessage = willMessage;
@@ -50,7 +69,15 @@ public final class MqttConnectPayload {
         return willTopic;
     }
 
+    /**
+     * @deprecated use {@link MqttConnectPayload#willMessageInBytes()} instead
+     */
+    @Deprecated
     public String willMessage() {
+        return new String(willMessage, CharsetUtil.UTF_8);
+    }
+
+    public byte[] willMessageInBytes() {
         return willMessage;
     }
 
@@ -58,7 +85,15 @@ public final class MqttConnectPayload {
         return userName;
     }
 
+    /**
+     * @deprecated use {@link MqttConnectPayload#passwordInBytes()} instead
+     */
+    @Deprecated
     public String password() {
+        return new String(password, CharsetUtil.UTF_8);
+    }
+
+    public byte[] passwordInBytes() {
         return password;
     }
 
--- a/codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttDecoder.java
+++ b/codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttDecoder.java
@@ -339,21 +339,21 @@ public final class MqttDecoder extends ReplayingDecoder<DecoderState> {
         int numberOfBytesConsumed = decodedClientId.numberOfBytesConsumed;
 
         Result<String> decodedWillTopic = null;
-        Result<String> decodedWillMessage = null;
+        Result<byte[]> decodedWillMessage = null;
         if (mqttConnectVariableHeader.isWillFlag()) {
             decodedWillTopic = decodeString(buffer, 0, 32767);
             numberOfBytesConsumed += decodedWillTopic.numberOfBytesConsumed;
-            decodedWillMessage = decodeAsciiString(buffer);
+            decodedWillMessage = decodeByteArray(buffer);
             numberOfBytesConsumed += decodedWillMessage.numberOfBytesConsumed;
         }
         Result<String> decodedUserName = null;
-        Result<String> decodedPassword = null;
+        Result<byte[]> decodedPassword = null;
         if (mqttConnectVariableHeader.hasUserName()) {
             decodedUserName = decodeString(buffer);
             numberOfBytesConsumed += decodedUserName.numberOfBytesConsumed;
         }
         if (mqttConnectVariableHeader.hasPassword()) {
-            decodedPassword = decodeString(buffer);
+            decodedPassword = decodeByteArray(buffer);
             numberOfBytesConsumed += decodedPassword.numberOfBytesConsumed;
         }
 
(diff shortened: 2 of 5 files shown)


Scottmitch:
This is a duplicate of https://github.com/netty/netty/pull/6769 ... fyi @oakad if you don't plan on resolving the comment in your PR will will focus on this PR instead.

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttConnectPayload.java
```diff
@@ -50,15 +50,15 @@ public String willTopic() {
         return willTopic;
     }
 
-    public String willMessage() {
+    public byte[] willMessage() {
```
Scottmitch: Please use the same approach (add a new method) as in https://github.com/netty/netty/pull/6769 and avoid breaking the public API.

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttDecoder.java
```diff
@@ -445,6 +438,21 @@ private static MqttFixedHeader decodeFixedHeader(ByteBuf buffer) {
         return new Result<String>(s, numberOfBytesConsumed);
     }
 
+    private static Result<byte[]> decodeByteArray(ByteBuf buffer, int minBytes, int maxBytes) {
+        final Result<Integer> decodedSize = decodeMsbLsb(buffer);
+        int size = decodedSize.value;
+        int numberOfBytesConsumed = decodedSize.numberOfBytesConsumed;
+        if (size < minBytes || size > maxBytes) {
```
Scottmitch: can we remove `minBytes` and `maxBytes` parameters? `maxBytes` is always `MAX_VALUE` and `minBytes` is always `0`. can we instead just throw an exception (or return early) if `size < 0`?

sbcd90:
Hi @Scottmitch , I have made the suggested changes.

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttConnectPayload.java
```diff
@@ -25,16 +26,30 @@
 
     private final String clientIdentifier;
     private final String willTopic;
-    private final String willMessage;
+    private final byte[] willMessage;
     private final String userName;
-    private final String password;
+    private final byte[] password;
 
     public MqttConnectPayload(
```
Scottmitch: should we deprecate this constructor in favor of the new constructor which takes `byte[]`?

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttConnectPayload.java
```diff
@@ -51,6 +66,10 @@ public String willTopic() {
     }
 
     public String willMessage() {
```
Scottmitch: Please deprecate this method and refer to the `willMessageInBytes` in the `@deprecated` javadocs.

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttConnectPayload.java
```diff
@@ -59,6 +78,10 @@ public String userName() {
     }
 
     public String password() {
```
Scottmitch: Please deprecate this method and refer to the `passwordInBytes ` in the `@deprecated` javadocs.

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttDecoder.java
```diff
@@ -445,6 +434,21 @@ private static MqttFixedHeader decodeFixedHeader(ByteBuf buffer) {
         return new Result<String>(s, numberOfBytesConsumed);
     }
 
+    private static Result<byte[]> decodeByteArray(ByteBuf buffer) {
+        final Result<Integer> decodedSize = decodeMsbLsb(buffer);
+        int size = decodedSize.value;
+        int numberOfBytesConsumed = decodedSize.numberOfBytesConsumed;
+        if (size < 0) {
+            buffer.skipBytes(size);
```
Scottmitch: this is invalid and will throw an exception.

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttDecoder.java
```diff
@@ -445,6 +434,21 @@ private static MqttFixedHeader decodeFixedHeader(ByteBuf buffer) {
...
+            numberOfBytesConsumed += size;
```
Scottmitch: do you really want to decrement `numberOfBytesConsumed`?

Scottmitch: can you add a unit test which hits this code?

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttMessageBuilders.java
```diff
@@ -120,6 +121,11 @@ public ConnectBuilder willTopic(String willTopic) {
         }
 
         public ConnectBuilder willMessage(String willMessage) {
```
Scottmitch: deprecate this method and reference the new method in javadocs?

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttMessageBuilders.java
```diff
@@ -146,6 +152,11 @@ public ConnectBuilder username(String username) {
         }
 
         public ConnectBuilder password(String password) {
```
Scottmitch: deprecate this method and reference the new method in javadocs?

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttDecoder.java
sbcd90: This part of the code will be hit only in a hypothetical situation when the encoded bytebuf have some issues like length of a message field, say password is set to -1 etc.

However, if the assumption is, the encoded bytebuf is created out of the MqttEncoder, then this part of the code can be safely commented out.

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttDecoder.java
```diff
@@ -17,6 +17,8 @@
 package io.netty.handler.codec.mqtt;
 
 import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufAllocator;
```
netkins: MINOR Remove this unused import 'io.netty.buffer.ByteBufAllocator'. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttDecoder.java
```diff
@@ -17,6 +17,8 @@
...
+import io.netty.buffer.EmptyByteBuf;
```
netkins: MINOR Remove this unused import 'io.netty.buffer.EmptyByteBuf'. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttDecoder.java
```diff
@@ -445,6 +434,20 @@ private static MqttFixedHeader decodeFixedHeader(ByteBuf buffer) {
         return new Result<String>(s, numberOfBytesConsumed);
     }
 
+    private static Result<byte[]> decodeByteArray(ByteBuf buffer) {
+        final Result<Integer> decodedSize = decodeMsbLsb(buffer);
+        int size = decodedSize.value;
+        int numberOfBytesConsumed = decodedSize.numberOfBytesConsumed;
+        if (size < 0) {
+            numberOfBytesConsumed += size;
```
Scottmitch: I still don't understand this line of code. 2 bytes have still been consumed either way even if the data is invalid. This value is ultimately set to `bytesRemainingInVariablePart` which may be used later during the decode process. You mentioned this was a [hypothetical situation](https://github.com/netty/netty/pull/6817#discussion_r120771580) but it is not clear what will happen if this situation does occur (if its possible at all). If this situation is possible please add a unit test, and if it isn't possible can we remove this line?

sbcd90:
Hi @Scottmitch, removed that line of code.

codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttDecoder.java
```diff
@@ -445,6 +434,16 @@ private static MqttFixedHeader decodeFixedHeader(ByteBuf buffer) {
...
     }
 
+    private static Result<byte[]> decodeByteArray(ByteBuf buffer) {
+        final Result<Integer> decodedSize = decodeMsbLsb(buffer);
+        int size = decodedSize.value;
+        int numberOfBytesConsumed = decodedSize.numberOfBytesConsumed;
+        byte[] bytes = new byte[size];
+        buffer.readBytes(bytes);
+        numberOfBytesConsumed += size;
+        return new Result<byte[]>(bytes, numberOfBytesConsumed);
```
fenik17: Now a `numberOfBytesConsumed` is not needed.
```java
return new Result<byte[]>(bytes, decodedSize.numberOfBytesConsumed + size);
```

sbcd90:
Hi @fenik17 , @Scottmitch , do you think of any further changes?

normanmaurer:
@sbcd90 can you please sign our ICLA and let me know once done:

https://netty.io/s/icla

sbcd90:
@normanmaurer , I have signed the ICLA.

normanmaurer:
Cherry-picked into 4.1 (66c83f7b74b3ea28baa4a0702d621aba8331993a).

@sbcd90 thanks!

PR #6822, opened 2017-06-06
Remove import of java.util.concurrent.ConcurrentLinkedDeque to be abl…

normanmaurer:
…e to use netty with java6

Motivation:

We need to not import java.util.concurrent.ConcurrentLinkedDeque to be able to use netty with java6

Modifications:

Remove import and use FQCN

Result:

Works on java6 as well

### DIFF
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent.java
@@ -39,7 +39,6 @@ import java.util.Map;
 import java.util.Queue;
 import java.util.Random;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.ConcurrentMap;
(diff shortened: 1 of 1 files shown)


mkw:
So, after some research, I do not believe this is necessary.  Whether a fully qualified reference or import is use does not appear to matter; both work fine.  See the test repository that I created [here](https://github.com/mkw/java6test):

* Requires Docker (should honor Docker Machine environment variables, but I used Docker for Mac) for executing the tests.
* Build and execute with `./gradlew check` (invokes the wrapper script so no need to install Gradle); requires at least Java 7 to build, though Java 8 works fine.
* All it does is check for the Java specification version and determine what Deque implementation to use.  One class uses a fully qualified reference and the other uses an import.  Both work.

normanmaurer:
@mkw you are right... I did more checking and it only is problematic if its part of a method signature. So Let me just close this one

Scottmitch:
@mkw - thanks for following up!

PR #6825, opened 2017-06-07
Correctly handle ByteBuf implementations which have no memoryAddress …

normanmaurer:
…when writing to native transport

Motivation:

Commit 3c4dfed08ae2ce300f6ac1d0794d42a14c344b79 introduced a regression in handling buffers that have no memoryAddress.

Modifications:

Fix regression and also add unit tests.

Result:

It's possible again to write buffers without memory address.

### DIFF
--- a/transport-native-unix-common/src/main/java/io/netty/channel/unix/UnixChannelUtil.java
+++ b/transport-native-unix-common/src/main/java/io/netty/channel/unix/UnixChannelUtil.java
@@ -29,6 +29,6 @@ public final class UnixChannelUtil {
      * (We check this because otherwise we need to make it a new direct buffer.)
      */
     public static boolean isBufferCopyNeededForWrite(ByteBuf byteBuf) {
-        return !(byteBuf.hasMemoryAddress() || byteBuf.isDirect() && byteBuf.nioBufferCount() <= IOV_MAX);
+        return !byteBuf.hasMemoryAddress() || !byteBuf.isDirect() || byteBuf.nioBufferCount() > IOV_MAX;
     }
 }
--- a/testsuite/src/main/java/io/netty/testsuite/transport/socket/DatagramUnicastTest.java
+++ b/testsuite/src/main/java/io/netty/testsuite/transport/socket/DatagramUnicastTest.java
@@ -37,7 +37,7 @@ public class DatagramUnicastTest extends AbstractDatagramTest {
 
     private static final byte[] BYTES = {0, 1, 2, 3};
     private enum WrapType {
-        NONE, DUP, SLICE,
+        NONE, DUP, SLICE, READ_ONLY
     }
 
     @Test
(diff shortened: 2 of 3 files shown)


normanmaurer:
@louxiu  FYI as well

louxiu:
My fault.... `UnixChannelUtilTest` is not executed. Maybe it should be moved to `transport-native-unix-common`.

transport-native-unix-common/src/main/java/io/netty/channel/unix/UnixChannelUtil.java
```diff
@@ -29,6 +29,6 @@ private UnixChannelUtil() {
      * (We check this because otherwise we need to make it a new direct buffer.)
      */
     public static boolean isBufferCopyNeededForWrite(ByteBuf byteBuf) {
-        return !(byteBuf.hasMemoryAddress() || byteBuf.isDirect() && byteBuf.nioBufferCount() <= IOV_MAX);
+        return !byteBuf.hasMemoryAddress() || !(byteBuf.isDirect() && byteBuf.nioBufferCount() <= IOV_MAX);
```
Scottmitch: IMHO the following is a bit simpler: `!byteBuf.hasMemoryAddress() || !byteBuf.isDirect() || byteBuf.nioBufferCount() > IOV_MAX` .. each term can be evaluated independently and no need to negate a group.

normanmaurer: sure why not... let me adjust

normanmaurer:
Cherry-picked into 4.0 (a2bd9a4399) and 4.1 (047da1108674f0c425f0f00d159cdec41d697268)

normanmaurer:
@louxiu will do a followup for this

louxiu:
This commit fails on `UnixChannelUtilTest`

normanmaurer:
@louxiu @tonyshenkk I forgot to backport the fix for the test to 4.1. See https://github.com/netty/netty/commit/925def9b56854a9f3ea57c791af6a5050bcd438b

louxiu:
A bit confused about this case. `assertCompositeByteBufIsBufferCopyNeededForWrite(alloc, 2, 0, true);` 
I think `CompositeByteBuf` with multi direct buffer(`hasMemoryAddress` should be true?) don't need a another copy.

normanmaurer:
@louxiu no as hasMemoryAddress() should only be true if it only has exactly one memory address.

louxiu:
When `CompositeByteBuf` with multi direct buffer is added to IovArray, it will be converted to nio buffer. The memory address of  `CompositeByteBuf` is not required?  Memory address of all direct buffers contained in it can be got.

normanmaurer:
@louxiu but this is not the contract of the `memoryAddress()` method so we will need to special check something here. Let me check later today.

tonyshenkk:
as implement in 4.0.23.Final in io.netty.channel.epoll.EpollDatagramChannel#filterOutboundMessage(Object)

it is no need to copy byteBuff if the byteBuf has memoryAddress or is composited direct byteBuf

normanmaurer:
@louxiu I just double-checked and this will only work if all the contained buffers in the CompositeByteBuffer are direct. So for this to work we will need to check this first. Can you please open another issue to track this "feature" ?

louxiu:
Glad to do that.

tonyshenkk:
I have pull a request to Fix this issue, at    https://github.com/netty/netty/pull/6872

PR #6826, opened 2017-06-07
Allow to access memoryAddress of wrapped ByteBuf for ReadOnlyByteBuf

normanmaurer:
Motivation:

We should allow to access the memoryAddress of the wrapped ByteBuf when using ReadOnlyByteBuf for peformance reasons. If a user act on a memoryAddress its his responsible anyway to do nothing "stupid".

Modifications:

Delegate to wrapped ByteBuf.

Result:

Less performance overhead for various operations and also when writing to a native transport (which needs the memoryAddress).


normanmaurer:
Cherry-picked into 4.1 (7922757575) and 4.0 (5a68d18905)

PR #6829, opened 2017-06-08
Use FQCN to prevent classloader issues on java6

normanmaurer:
Motivation:

We need to use FQCN to prevent classloader issues for classes that are > Java6. This is a cleanup of ed5fcbb7732d9ffd16d9675fe40711360763c9a4.

Modifications:

Just remove the imports and use FQCN.

Result:

No classloader issues with java6

### DIFF
--- a/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
+++ b/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
@@ -27,8 +27,6 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.lang.reflect.Method;
 import java.net.URL;
-import java.nio.file.Files;
-import java.nio.file.attribute.PosixFilePermission;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


normanmaurer:
@johnou FYI

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -409,17 +407,20 @@ private static boolean canExecuteExecutable(File file) throws IOException {
             // the UnixFileStore, though the flag is not exposed via a public API.  To find out if library is being
             // loaded off a volume with noexec, confirm or add executalbe permissions, then check File#canExecute().
 
-            Set<PosixFilePermission> existingFilePermissions = Files.getPosixFilePermissions(file.toPath());
-            Set<PosixFilePermission> executePermissions = EnumSet.of(PosixFilePermission.OWNER_EXECUTE,
-                                                                     PosixFilePermission.GROUP_EXECUTE,
-                                                                     PosixFilePermission.OTHERS_EXECUTE);
+            // Note: We use FQCN to not break when netty is used in java6
```
hepin1989: should add a note like：
Please not change it,trust me.

normanmaurer:
Cherry-picked into 4.1 (f208b147a6) and 4.0 (164a1f181351005fe552551ea40607097e1e79cf)

PR #6830, opened 2017-06-08
Not force to run autoconf and compile multiple times

normanmaurer:
Motivation:

We should not force autoconf and compile as this will result in multiple executions and so slow down the build.

Modifications:

Remove force declarations

Result:

Faster build of native modules


Scottmitch:
if the environment changes will it automatically be detected, or are there some command line arguments we can use to force the configuration?

normanmaurer:
@Scottmitch if you run `mvn clean package` it will just be fine.

normanmaurer:
Cherry-picked into 4.1 (b6c27b9f6b) and 4.0 (a4d81dbf32)

PR #6833, opened 2017-06-08
Minor improvements for append params in MessageFormatter

fenik17:
Motivation:

Empty arrays can be detected through `Array.getLength` (without type checking).

Modifications:

Use `Array.getLength` to avoid checking type for empty arrays and checking arrays length by zero in appropriate methods.

Result:

Less code. Faster append for empty arrays.


normanmaurer:
@fenik17 do you have a reference to this (like that it is really faster?)

fenik17:
@normanmaurer JMH showed none performance changes :cry:. Just close )

PR #6838, opened 2017-06-11
Do proper bounds-checking in HpackHuffmanDecoder to reduce overhead o…

normanmaurer:
…f IndexOutOfBoundsException creation

Motivation:

HpackHuffmanDecoder.Decoder did not do any bound-checking but just catched IndexOutOfBoundsException to detect if the array needs to grow. This can be very expensive because of fillInStackTrace()

Modifications:

Add proper bounds checking and grow the array if needed without catching IndexOutOfBoundsException.

Result:

Less overhead if the array needs to grow.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/HpackHuffmanDecoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/HpackHuffmanDecoder.java
@@ -230,16 +230,13 @@ final class HpackHuffmanDecoder {
         }
 
         private void append(int i) {
-            try {
-                bytes[index] = (byte) i;
-            } catch (IndexOutOfBoundsException ignore) {
+            if (bytes.length == index) {
                 // Always just expand by INITIAL_SIZE
                 byte[] newBytes = new byte[bytes.length + initialCapacity];
                 System.arraycopy(bytes, 0, newBytes, 0, bytes.length);
                 bytes = newBytes;
-                bytes[index] = (byte) i;
             }
-            index++;
+            bytes[index++] = (byte) i;
         }
     }
 }


Scottmitch:
Checking for the exception after the fact was shown to improve performance over the long hall (https://github.com/netty/netty/commit/a7713069a13b47efbf197a1c7416500c45d14d7c), but I would guess that checking the index wouldn't be subject to larger penalties if the initial size is too small. Let me revert this portion of `HttpObjectDecoder` to avoid these potential issues.

normanmaurer:
@Scottmitch +1 ... Also are you ok with merge this one ?

Scottmitch:
see https://github.com/netty/netty/pull/6845

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackHuffmanDecoder.java
```diff
@@ -230,15 +230,14 @@ AsciiString end() throws Http2Exception {
...
-                bytes[index] = (byte) i;
-            } catch (IndexOutOfBoundsException ignore) {
+            if (bytes.length >= index) {
                 // Always just expand by INITIAL_SIZE
                 byte[] newBytes = new byte[bytes.length + initialCapacity];
                 System.arraycopy(bytes, 0, newBytes, 0, bytes.length);
                 bytes = newBytes;
-                bytes[index] = (byte) i;
             }
+            bytes[index] = (byte) i;
```
Scottmitch: Consider doing the following:

```java
bytes[index++] = (byte) i;
```

this would be consistent with https://github.com/netty/netty/pull/6845

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackHuffmanDecoder.java
```diff
@@ -230,15 +230,14 @@ AsciiString end() throws Http2Exception {
         }
 
         private void append(int i) {
-            try {
-                bytes[index] = (byte) i;
-            } catch (IndexOutOfBoundsException ignore) {
+            if (bytes.length >= index) {
```
Scottmitch: is `index == bytes.length` enough? (wondering if we can be consistent with https://github.com/netty/netty/pull/6845).

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackHuffmanDecoder.java
```diff
@@ -230,15 +230,14 @@ AsciiString end() throws Http2Exception {
...
                 // Always just expand by INITIAL_SIZE
                 byte[] newBytes = new byte[bytes.length + initialCapacity];
```
Scottmitch: I wonder if this is worth re-visiting (followup PR). Perhaps we should be more aggressive at first (double it until a certain threshold is crossed)?

normanmaurer: maybe... I will open an issue.

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackHuffmanDecoder.java
normanmaurer: yes

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackHuffmanDecoder.java
normanmaurer: sure

normanmaurer:
@Scottmitch PTAL again... all addressed an followup issue https://github.com/netty/netty/issues/6846 created

normanmaurer:
Cherry-picked into 4.1 (3f0085c267)

Issue #6831, opened 2017-06-08
FixedChannelPool throw java.lang.IllegalStateException: Channel is unhealthy not offering it back to pool


fian46:
Hi, i am trying to write some TCP stuff with FixedChannelPool. The documentation says i have to return any channel that borrow from pool even if the channel already dead, but when i do return the channel, FixedChannelPool throw java.lang.IllegalStateException: Channel is unhealthy not offering it back to pool.
Is this bug ?
if i don't release the channel to FixedChannelPool its never give any connection when dead channel count = max_pool_size.
if i just ignore the exception, everything works fine, but i think this should never happend.

### Expected behavior
i expect no exception thrown
### Actual behavior
throw  java.lang.IllegalStateException: Channel is unhealthy not offering it back to pool.
### Steps to reproduce
1. Create FixedChannelPool.
2. call aquire channel
3. kill channel
4. call release channel, but the returned ChannelFuture call get()
5. TA DA throw  java.lang.IllegalStateException: Channel is unhealthy not offering it back to pool.

### Minimal yet complete reproducer code (or URL to code)
`public class MaybeBug {`
`    @Test`
`    public void test() throws InterruptedException, ExecutionException{`
`        Bootstrap config = new Bootstrap();`
`        config.remoteAddress("localhost", 1234);`
`        config.option(ChannelOption.TCP_NODELAY, true);`
`        config.channel(NioSocketChannel.class);`
`        config.group(new NioEventLoopGroup(1));`
`        FixedChannelPool pool = new FixedChannelPool(config, new AbstractChannelPoolHandler() {`
`            @Override`
`            public void channelCreated(Channel ch) throws Exception {`
`                /**`
`                 * ignore`
`                 */`
`            }`
`        }, 2);`
`        Channel fresh = pool.acquire().get();`
`        fresh.close().get();`
`        pool.release(fresh).get();`
`    }
}`
### Netty version
4.1.5
### JVM version (e.g. `java -version`)
1.8
### OS version (e.g. `uname -a`)
Windows 10

NB: sorry for bad english.

normanmaurer:
@fian46 please check https://github.com/netty/netty/pull/6843

normanmaurer:
Fixed by https://github.com/netty/netty/pull/6843

PR #6843, opened 2017-06-12
Not fail the promise when a closed Channel is offered back to the Cha…

normanmaurer:
…nnelPool

Motivation:

We should not fail the promise when a closed Channel is offereed back to the ChannelPool as we explicit mention that the Channel must always be returned.

Modifications:

- Not fail the promise
- Add test-case

Result:

Fixes [#6831]

### DIFF
--- a/transport/src/main/java/io/netty/channel/pool/SimpleChannelPool.java
+++ b/transport/src/main/java/io/netty/channel/pool/SimpleChannelPool.java
@@ -43,9 +43,6 @@ public class SimpleChannelPool implements ChannelPool {
     private static final AttributeKey<SimpleChannelPool> POOL_KEY = AttributeKey.newInstance("channelPool");
     private static final IllegalStateException FULL_EXCEPTION = ThrowableUtil.unknownStackTrace(
             new IllegalStateException("ChannelPool full"), SimpleChannelPool.class, "releaseAndOffer(...)");
-    private static final IllegalStateException UNHEALTHY_NON_OFFERED_TO_POOL = ThrowableUtil.unknownStackTrace(
-            new IllegalStateException("Channel is unhealthy not offering it back to pool"),
-            SimpleChannelPool.class, "releaseAndOffer(...)");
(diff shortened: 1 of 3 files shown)


normanmaurer:
PTAL @fian46

transport/src/main/java/io/netty/channel/pool/SimpleChannelPool.java
```diff
@@ -327,7 +324,7 @@ private void releaseAndOfferIfHealthy(Channel channel, Promise<Void> promise, Fu
             releaseAndOffer(channel, promise);
         } else { //channel ont healthy, just releasing it.
```
Scottmitch: `ont` -> `not`

transport/src/test/java/io/netty/channel/pool/FixedChannelPoolTest.java
```diff
@@ -300,6 +300,35 @@ public void run() {
         channel.close().syncUninterruptibly();
     }
 
+    @Test
+    public void testReleaseClosed() {
+        EventLoopGroup group = new LocalEventLoopGroup(1);
```
Scottmitch: shouldn't we be closing this at the end of the test? consider creating this in a `@BeforeClass` and cleaning up in `@AfterClass`.

normanmaurer: yes! Let me update

transport/src/main/java/io/netty/channel/pool/SimpleChannelPool.java
normanmaurer: +1

normanmaurer:
Cherry-picked into 4.1 (94c0ef3c96189b3fb02ee68a4e229d1e3ba6e2df) and 4.0 (8b66dc7c16)

fian46:
nice already fixed and closed

PR #6845, opened 2017-06-12
AppendableCharSequence not to depend upon IndexOutOfBoundsException f…

Scottmitch:
…or resize

Motivation:
AppendableCharSequence depends upon IndexOutOfBoundsException to trigger a resize operation under the assumption that the resize operation will be rare if the initial size guess is good. However if the initial size guess is not good then the performance will be more unpredictable and likely suffer.

Modifications:
- Check the position in AppendableCharSequence#append to determine if a resize is necessary

Result:
More predictable performance in AppendableCharSequence#append.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/AppendableCharSequence.java
+++ b/common/src/main/java/io/netty/util/internal/AppendableCharSequence.java
@@ -68,12 +68,12 @@ public final class AppendableCharSequence implements CharSequence, Appendable {
 
     @Override
     public AppendableCharSequence append(char c) {
-        try {
-            chars[pos++] = c;
-        } catch (IndexOutOfBoundsException e) {
-            expand();
-            chars[pos - 1] = c;
+        if (pos == chars.length) {
+            char[] old = chars;
+            chars = new char[old.length << 1];
+            System.arraycopy(old, 0, chars, 0, old.length);
         }
+        chars[pos++] = c;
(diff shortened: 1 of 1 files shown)


Scottmitch:
4.1 (1cc4607f0) 4.0 (4587800d8)

common/src/main/java/io/netty/util/internal/AppendableCharSequence.java
```diff
@@ -68,12 +68,12 @@ public AppendableCharSequence subSequence(int start, int end) {
...
     @Override
     public AppendableCharSequence append(char c) {
-        try {
-            chars[pos++] = c;
-        } catch (IndexOutOfBoundsException e) {
-            expand();
-            chars[pos - 1] = c;
+        if (pos == chars.length) {
+            char[] old = chars;
+            chars = new char[old.length << 1];
```
johnou: I guess this blows up if old overflows with something like negative index not allowed?

Scottmitch: Yes the intention is to just let `NegativeArraySizeException` propagate as I'm not sure there is much value in throwing `IllegalStateException` instead.

johnou: sgtm

PR #6848, opened 2017-06-13
Mention that the IdleStateHandler should be placed before any decoder…

normanmaurer:
… / encoder in the pipeline.

Motivation:

To ensure that the IdleStateHandler works as expected its important to put it in the right position into the ChannelPipeline.

Modifications:

Add docs to make it clear.

Result:

Better docs.

### DIFF
--- a/handler/src/main/java/io/netty/handler/timeout/IdleStateHandler.java
+++ b/handler/src/main/java/io/netty/handler/timeout/IdleStateHandler.java
@@ -21,11 +21,12 @@ import io.netty.channel.Channel.Unsafe;
 import io.netty.channel.ChannelDuplexHandler;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
+import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInitializer;
 import io.netty.channel.ChannelOutboundBuffer;
+import io.netty.channel.ChannelPipeline;
 import io.netty.channel.ChannelPromise;
-import io.netty.util.concurrent.EventExecutor;
 
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


handler/src/main/java/io/netty/handler/timeout/IdleStateHandler.java
```diff
@@ -34,6 +35,9 @@
  * Triggers an {@link IdleStateEvent} when a {@link Channel} has not performed
  * read, write, or both operation for a while.
  *
+ * The {@link IdleStateHandler} should be placed as first {@link ChannelHandler}
```
rkapsi: I'd argue that it depends on the use-case and is possibly dangerous advise. E.g. putting it before a decoder (server side) will expose the user to certain DoS attacks.

rkapsi: s/expose/leave vulnerable

normanmaurer: @rkapsi... can you give an example of such an attack ?

rkapsi: Sure... Say you use HttpServerCodec.

An attacker can open a connection and very slowly trickle bytes of a valid HTTP request and just staying below the IdleStateHandler's threshold. Allowing the attacker to open many connections at a relative low bandwidth cost.

rkapsi: Putting it after the decoder will force the attacker to deliver a HTTP request within the specified threshold.

Scottmitch: Agreed it depends on what you are trying to achieve.

normanmaurer: @rkapsi @Scottmitch @trustin so you say we should better not put this statement into the java docs at all ?

rkapsi: Depends. :D

I don't think it's bad advise if you keep DoS attack vectors, malicious or buggy behavior out of the picture. But ultimately it boils down to realizing that idleness is measured on event level and the meaning thereof depends on the position in the pipeline.

Scottmitch: agreed ... I would say omit it because it is not a requirement to use `IdleStateHandler` as it depends upon the applications desired result. Maybe explaining the behavior it will provide relative to other handlers would be better?

normanmaurer:
let me close this

PR #6850, opened 2017-06-13
Let DnsNameResolver constructor use a default value for searchDomains

slandelle:
Motivation:

It’s currently complicated to extend `DnsNameResolver` as the default
value for `searchDomain` is package private.

Modifications:

* let `DnsNameResolver` accept a null `searchDomains` and then default
to `DEFAULT_SEARCH_DOMAINS`, just like it’s being done with
`resolvedAddressTypes`.
* set default `DnsNameResolverBuilder#searchDomains` value to null to
avoid cloning internal `DnsNameResolver.DEFAULT_SEARCH_DOMAINS` in
`DnsNameResolver` constructor.

Result:

More versatile `DnsNameResolver` constructor.
No array copy when using default search domains.

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
@@ -195,6 +195,7 @@ public class DnsNameResolver extends InetNameResolver {
      * @param dnsServerAddressStreamProvider The {@link DnsServerAddressStreamProvider} used to determine the name
      *                                       servers for each hostname lookup.
      * @param searchDomains the list of search domain
+     *                      (can be null, if so, will try to default to the underlying platform ones)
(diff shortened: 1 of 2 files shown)


normanmaurer:
@slandelle LGTM in general but can you just update the javadoc and mention that null is allowed and the default will be used in this case: https://github.com/slandelle/netty/blob/21e003cd98547d694b36e11ca616b089bf8de9a0/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java#L197

slandelle:
@normanmaurer done

normanmaurer:
@slandelle thanks a lot.... 

Cherry-picked into 4.1 (ca5ed7c114120e6d45522be0bb231fedae065cf3)

slandelle:
Thanks for merging!

Issue #6852, opened 2017-06-13
Netty bom does not provide dependency management for netty-transport-native-epoll


snicoll:
### Expected behavior

The bom provides consistent versions for all of Netty's modules

### Actual behavior

No dependency management for ` io.netty:netty-transport-native-epoll` is available, leading to a potential mismatch

### Steps to reproduce

Import the bom and try to resolve `netty-transport-native-epoll` without a version.

### Minimal yet complete reproducer code (or URL to code)

N/A

### Netty version

4.1.12.Final

### JVM version (e.g. `java -version`)

N/A

### OS version (e.g. `uname -a`)

N/A

Scottmitch:
Are you interested in submitting a PR?

Scottmitch:
Oh I see PR https://github.com/netty/netty/pull/6853 now ... thanks!

normanmaurer:
Fixed by https://github.com/netty/netty/pull/6853

PR #6853, opened 2017-06-13
 Add dependency management for missing entries

snicoll:
Motivation:

The bom does not provide entries for a number of netty modules, in
particular those that are deployed with classifiers. As a result, they
can't be used without defining a version.

Modifications:

Provide dependency management for the missing modules.

Result:

Fixes [#6852]

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -138,6 +138,11 @@
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
+        <artifactId>netty-dev-tools</artifactId>
+        <version>4.1.13.Final-SNAPSHOT</version>
+      </dependency>
+      <dependency>
+        <groupId>io.netty</groupId>
         <artifactId>netty-handler</artifactId>
         <version>4.1.13.Final-SNAPSHOT</version>
       </dependency>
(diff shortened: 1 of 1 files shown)


snicoll:
@normanmaurer I've signed the icla.

Scottmitch:
Seems strange that an entry is required with and without the classifier ... can you explain why this is required?

snicoll:
That's how maven works I am afraid. If you have dep management without the classifier it won't work if you add a classifier. And vice versa.

normanmaurer:
@snicoll I think we should only have the one with the classifier /cc @Scottmitch

snicoll:
We do use the one without in Spring Boot though. Are we doing something  wrong maybe?

normanmaurer:
@snicoll the one without the classifier will not contain any native code, so yes I think so

Scottmitch:
In general the artifact with the dependency is the only one that is functional. It may be useful to include both dependencies if you want your library/application to include EPOLL code regardless if EPOLL is available or not. You just need to be careful to check `Epoll.isAvailable()` before you use it. @snicoll - do you have an issue where the classifier is dynamic or not matching the end user's configuration and therefore causing issues?

snicoll:
I am not educated enough to answer that question but @mp911de can since he's working on Lettuce and that's the component using it.

On a more general note though, as long as you publish an artifact as part of Netty, it should be present in the dependency management of the bom. I'd argue that either this should be fixed or you should stop publishing that jar.

If relying to the jar without the classifier is the wrong thing to do, I am sure Mark can act on it based on your advice. Thanks!

mp911de:
It might make sense to include both artifacts from a stability perspective, to not break builds. From a functional perspective, it makes sense to only include `netty-transport-native-epoll` with the classifier, otherwise the dependency is pointless in user code. IIRC `netty-transport-native-kqueue` is included only with a classifier.

normanmaurer:
@mp911de +1 sp @snicoll I think all is good as we include it with the classifier already. correct ?

snicoll:
No. It's not. You need one entry without it. See my precious comment.

normanmaurer:
@snicoll ah I missed this. Can you then also add one for the kqueue one ?

snicoll:
Happy to do it even though that sounds like a different pr right?

normanmaurer:
@snicoll just adjust the commit message and the PR details to cover it and we are good :)

snicoll:
@normanmaurer done

normanmaurer:
Thanks... cherry-picked into 4.1 (2af895994dde5d17fb56e191b15a12bbe17df8da)

Issue #4312, opened 2015-10-03
HttpRequestDecoder fires channelReadComplete on every network read


scf37:
Expected: HttpRequestDecoder  should  fire channelReadComplete event when last available HttpMessage have been consumed

Observed: channelReadComplete on every network packet received.

How to reproduce:

```
ch.pipeline().addLast(new HttpRequestDecoder());
ch.pipeline().addLast(new HttpResponseEncoder());
ch.pipeline().addLast(new MyChannelInboundHandler()); //overrides channelReadComplete 
```

When connected to such server using Windows telnet (IMPORTANT because it flushes on every keystroke), channelReadComplete fires on every keystroke instead of once - when I finish entering http request.

It seems that ByteToMessageDecoder blindly forwards channelReadComplete while handling channelRead correctly.

Netty 4.0.28.Final. Oracle JDK 1.8.0_60. Windows 10.

scf37:
Some background: I'm using HttpRequestDecoder with AUTO_READ=false and need to know when to call ChannelContext.read()

Scottmitch:
@scf37 - I'm not sure that `channelReadComplete` behaves like you are expecting.

See [ChanneHandler](https://github.com/netty/netty/blob/master/transport/src/main/java/io/netty/channel/ChannelHandler.java#L221):
```java
/**
 * Invoked when the last message read by the current read operation has been consumed by
 * {@link #channelRead(ChannelHandlerContext, Object)}.  If {@link ChannelOption#AUTO_READ} is off, no further
 * attempt to read an inbound data from the current {@link Channel} will be made until
 * {@link ChannelHandlerContext#read()} is called.
 */
void channelReadComplete(ChannelHandlerContext ctx) throws Exception;
```

`channelReadComplete()` is not designed to signify any specific protocol event, but instead notifies you that no more reads will be performed on a particular channel before returning back to the event loop.  In your case a "complete http request" will be signified by an object of type [LastHttpContent](https://github.com/netty/netty/blob/master/codec-http/src/main/java/io/netty/handler/codec/http/LastHttpContent.java).

In general you can not control how your peers will write nor how bytes will arrive at your application. `autoRead=false` is a way to give your application more control as to when reads are done on the socket. So if you are expecting more data to come (and your application can handle more data) then you are responsible for invoking `Channel.read()` at your discretion.

scf37:
[...]

PR #6856, opened 2017-06-14
Only call ctx.fireChannelReadComplete() if ByteToMessageDecoder decod…

normanmaurer:
…ed at least one message.

Motivation:

Its wasteful and also confusing that channelReadComplete() is called even if there was no message forwarded to the next handler.

Modifications:

- Only call ctx.fireChannelReadComplete() if at least one message was decoded
- Add unit test

Result:

Less confusing behavior. Fixes [#4312].

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/spdy/SpdyFrameCodec.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/spdy/SpdyFrameCodec.java
@@ -125,13 +125,10 @@ public class SpdyFrameCodec extends ByteToMessageDecoder
 
     @Override
     public void channelReadComplete(ChannelHandlerContext ctx) throws Exception {
-        if (!read) {
-            if (!ctx.channel().config().isAutoRead()) {
-                ctx.read();
-            }
-        }
+        boolean wasRead = read;
         read = false;
-        super.channelReadComplete(ctx);
+
+        channelReadComplete(ctx, wasRead);
     }
 
     @Override
--- a/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
@@ -313,15 +313,18 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
 
     @Override
     public void channelReadComplete(ChannelHandlerContext ctx) throws Exception {
+        channelReadComplete(ctx, !decodeWasNull);
+    }
+
+    protected final void channelReadComplete(ChannelHandlerContext ctx, boolean readData) throws Exception {
         numReads = 0;
         discardSomeReadBytes();
-        if (decodeWasNull) {
-            decodeWasNull = false;
-            if (!ctx.channel().config().isAutoRead()) {
-                ctx.read();
-            }
+        decodeWasNull = false;
+        if (readData) {
+            ctx.fireChannelReadComplete();
+        } else if (!ctx.channel().config().isAutoRead()) {
+            ctx.read();
         }
-        ctx.fireChannelReadComplete();
     }
 
     protected final void discardSomeReadBytes() {
--- a/codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
+++ b/codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
@@ -26,6 +26,8 @@ import org.junit.Test;
 import java.util.List;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingDeque;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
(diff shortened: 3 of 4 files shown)


Scottmitch:
Looks like there is some breakage with this PR:

```
[11:02:35][io.netty:netty-testsuite] Tests in error: 
[11:02:35][io.netty:netty-testsuite]   SocketSpdyEchoTest.testSpdyEcho:150->AbstractComboTestsuiteTest.run:67->testSpdyEcho:155->testSpdyEcho:212 ? TestTimedOut
[11:02:35][io.netty:netty-testsuite]   SocketSpdyEchoTest.testSpdyEchoNotAutoRead:160->AbstractComboTestsuiteTest.run:67->testSpdyEchoNotAutoRead:165->testSpdyEcho:212 ? TestTimedOut
```

normanmaurer:
@Scottmitch will check

normanmaurer:
@Scottmitch @trustin ok fixed my stupid bug (brainfart), so please check again

normanmaurer:
Also I think we may want to wait till the next release before pull it in to give it a bit more testing

codec-http/src/main/java/io/netty/handler/codec/spdy/SpdyFrameCodec.java
```diff
@@ -125,13 +125,18 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) t
...
+        read = false;
+        if (!wasRead) {
             if (!ctx.channel().config().isAutoRead()) {
                 ctx.read();
             }
         }
-        read = false;
         super.channelReadComplete(ctx);
+
+        if (wasRead) {
```
Scottmitch: `super.channelReadComplete(ctx);` may also call `ctx.fireChannelReadComplete();` ... are we sure this method won't be called twice?

normanmaurer:
@Scottmitch PTAL again...

Scottmitch:
@normanmaurer - Test appears to have timed out on CI server ... see PR https://github.com/netty/netty/pull/6990.

codec-http/src/main/java/io/netty/handler/codec/spdy/SpdyFrameCodec.java
```diff
@@ -125,13 +125,19 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) t
...
+
+        if (!wasRead) {
             if (!ctx.channel().config().isAutoRead()) {
                 ctx.read();
             }
+        } else {
+            ctx.fireChannelReadComplete();
         }
-        read = false;
-        super.channelReadComplete(ctx);
```
Scottmitch: `ByteToMessageDecoder` has some state which is updated in its `channelReadComplete` ... shouldn't we still delegate?

codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
```diff
@@ -305,4 +307,37 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) t
...
+            protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws Exception {
+                if (in.readableBytes() > 1) {
+                    readCompleteExpected.set(true);
+                    out.add(in.readBytes(in.readableBytes()));
+                }
+            }
+        }, new ChannelInboundHandlerAdapter() {
+            @Override
+            public void channelReadComplete(ChannelHandlerContext ctx) throws Exception {
+                assertTrue(readCompleteExpected.get());
```
Scottmitch: should we could how many times read complete is called and verify it is called the expected amount?

codec-http/src/main/java/io/netty/handler/codec/spdy/SpdyFrameCodec.java
normanmaurer: yeah we should reset the state... Actually the same "issue" exists in `SslHandler`. Let me come up with something.

codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
normanmaurer: done

normanmaurer:
@Scottmitch PTAL again... not sure I am super in love with the fix for resetting the state tho.

codec-http/src/main/java/io/netty/handler/codec/spdy/SpdyFrameCodec.java
```diff
@@ -125,13 +125,16 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) t
 
     @Override
     public void channelReadComplete(ChannelHandlerContext ctx) throws Exception {
+        resetReadState();
```
Scottmitch: What about adding a method to `ByteToMessageDecoder` like this:

```java
@Override
public void channelReadComplete(ChannelHandlerContext ctx) throws Exception {
  channelReadComplete(ctx, !decodeWasNull);
}

protected final void channelReadComplete(ChannelHandlerContext ctx, boolean readData) throws Exception {
        numReads = 0;
        discardSomeReadBytes();
        decodeWasNull = false;
        if (readData) {
          ctx.fireChannelReadComplete(); 
        } else if (ctx.channel().config().isAutoRead()) {
          ctx.read();
        }
}
```

Then in SpdyFrameCodec.java you can do this:

```java
public void channelReadComplete(ChannelHandlerContext ctx) throws Exception {
  // Not sure if it is necessary to reset before the call ... but just in case.
  boolean wasRead = read;
  read = false;
  channelReadComplete(ctx, wasRead);
}
```

normanmaurer: good idea... PTAL again

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1167,14 +1167,15 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) t
...
-        discardSomeReadBytes();
-
         flushIfNeeded(ctx);
-        readIfNeeded(ctx);
-
+        boolean readData = firedChannelRead;
         firedChannelRead = false;
-        ctx.fireChannelReadComplete();
+        // if readData is false channelReadComplete(....) will take care of calling read.
+        if (readData && !handshakePromise.isDone() && !ctx.channel().config().isAutoRead()) {
```
Scottmitch: can we share this logic with `readIfNeeded`'s logic? maybe have `readIfNeeded` take a `boolean readData` parameter if necessary?

normanmaurer: @Scottmitch the branch is different...

Scottmitch: +1 upon closer inspection the conditional is different ... fine as is

normanmaurer:
need to investigate the test-failures.

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -313,15 +313,18 @@ static void fireChannelRead(ChannelHandlerContext ctx, CodecOutputList msgs, int
...
         discardSomeReadBytes();
-        if (decodeWasNull) {
-            decodeWasNull = false;
-            if (!ctx.channel().config().isAutoRead()) {
-                ctx.read();
-            }
+        decodeWasNull = false;
+        if (readData) {
+            ctx.fireChannelReadComplete();
+        } else if (ctx.channel().config().isAutoRead()) {
```
normanmaurer: doh! I missed `!` here.

normanmaurer:
Cherry-picked into 4.1 (d63bb4811ed8ccd5d9e45853f3ac6aee9da7ecab) and 4.0 (29d5bac7ac)

normanmaurer:
This was reverted via:
https://github.com/netty/netty/pull/7121

Issue #6846, opened 2017-06-12
Reconsider growing strategy in HpackHuffmanDecoder


normanmaurer:
At the moment we just increase my intialCapacity but it may make sense to double it up to a certain size:

https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/HpackHuffmanDecoder.java#L237

Scottmitch:
thanks for reporting @normanmaurer! My guess is this will have a measurable impact if the initial size is too small (currently defaults to 32).

normanmaurer:
@Scottmitch agree... I would say we should double up to 512 bytes . WDYT ? I can do a pr

Scottmitch:
It will depend upon the use case but can we derive a guess from https://github.com/http2/http_samples or some other sample size? We can start with 512 or 1024 and refine later if you don't have much cycles to investigate.

normanmaurer:
@Scottmitch good idea... let me check

normanmaurer:
@Scottmitch here you go https://github.com/netty/netty/pull/6857. I think I will have not much time to long into it in more detail so I just used 1024 for now

Scottmitch:
1024 sounds like a good start to me ... we can revise/improve later if necessary.

normanmaurer:
Fixed by https://github.com/netty/netty/pull/6857

PR #6857, opened 2017-06-14
Use more aggressive expanding strategy in HpackHuffmanDecoder

normanmaurer:
Motivation:

Before we always expanded the buffer by the initialCapacity which by default is 32 bytes. This may lead to many expansions of the buffer until we finally reached the point that the buffer can fin everything.

Modifications:

Double the buffer size until the threshold of >= 1024 is hit. After this will grow it by the initialCapacity

Result:

Less expansion of the buffer (and so allocations / copies) when the intialCapacity is not big enough.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/HpackHuffmanDecoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/HpackHuffmanDecoder.java
@@ -231,8 +231,11 @@ final class HpackHuffmanDecoder {
 
         private void append(int i) {
             if (bytes.length == index) {
-                // Always just expand by INITIAL_SIZE
-                byte[] newBytes = new byte[bytes.length + initialCapacity];
+                // Choose an expanding strategy depending on how big the buffer already is.
+                // 1024 was choosen as a good guess and we may be able to investigate more if there are better choices.
+                // See also https://github.com/netty/netty/issues/6846
+                final int newLength = bytes.length >= 1024 ? bytes.length + initialCapacity : bytes.length << 1;
+                byte[] newBytes = new byte[newLength];
                 System.arraycopy(bytes, 0, newBytes, 0, bytes.length);
                 bytes = newBytes;
             }


Scottmitch:
can you add a reference to https://github.com/netty/netty/issues/6846 in the commit message?

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackHuffmanDecoder.java
```diff
@@ -231,8 +231,15 @@ AsciiString end() throws Http2Exception {
 
         private void append(int i) {
             if (bytes.length == index) {
-                // Always just expand by INITIAL_SIZE
-                byte[] newBytes = new byte[bytes.length + initialCapacity];
+                final int newLength;
```
Scottmitch: nit: consider using tertiary for less boiler plate:

```java
final int newLength = bytes.length >= 1024 ? bytes.length + initialCapacity : bytes.length << 1;
```

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackHuffmanDecoder.java
```diff
@@ -231,8 +231,15 @@ AsciiString end() throws Http2Exception {
...
+                // Choose an expanding strategy depending on how big the buffer already is.
+                if (bytes.length >= 1024) {
```
Scottmitch: Consider leaving a comment as to how this value was derived (educated guess, average header length from source X, etc..)

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackHuffmanDecoder.java
normanmaurer: sure why not

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackHuffmanDecoder.java
normanmaurer: sure can do

normanmaurer:
Addressed comments. Let me pull in

normanmaurer:
Cherry-picked into 4.1 (575baf5050)

PR #6858, opened 2017-06-14
Small performance improvement for HttpRequestEncoder

doom369:
Motivation:

Right now `HttpRequestEncoder` does insertion of slash for url like `http://localhost?pararm=1` before the question mark. It is done not effectively.

Modification:

Code:
```
new StringBuilder(len + 1)
                .append(uri, 0, index)
                .append(SLASH)
                .append(uri, index, len)
                .toString();
```

Replaced with:
```
new StringBuilder(uri)
                .insert(index, SLASH)
                .toString();
```

Result:

Faster `HttpRequestEncoder`. Additional small test. Attached benchmark in PR.

```
Benchmark                                      Mode  Cnt        Score        Error  Units
HttpRequestEncoderInsertBenchmark.newEncoder  thrpt   40  3704843.303 ±  98950.919  ops/s
HttpRequestEncoderInsertBenchmark.oldEncoder  thrpt   40  3284236.960 ± 134433.217  ops/s
```

### DIFF
--- a/codec-http/src/test/java/io/netty/handler/codec/http/HttpRequestEncoderTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/HttpRequestEncoderTest.java
@@ -35,6 +35,7 @@ public class HttpRequestEncoderTest {
                 HttpMethod.GET, "http://localhost"));
         String req = buffer.toString(Charset.forName("US-ASCII"));
         assertEquals("GET http://localhost/ HTTP/1.1\r\n", req);
+        buffer.release();
     }
 
     @Test
@@ -45,6 +46,18 @@ public class HttpRequestEncoderTest {
                 "http://localhost:9999?p1=v1"));
         String req = buffer.toString(Charset.forName("US-ASCII"));
         assertEquals("GET http://localhost:9999/?p1=v1 HTTP/1.1\r\n", req);
+        buffer.release();
+    }
+
+    @Test
+    public void testUriWithEmptyPath() throws Exception {
+        HttpRequestEncoder encoder = new HttpRequestEncoder();
+        ByteBuf buffer = Unpooled.buffer(64);
+        encoder.encodeInitialLine(buffer, new DefaultHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.GET,
+                "http://localhost:9999/?p1=v1"));
+        String req = buffer.toString(Charset.forName("US-ASCII"));
+        assertEquals("GET http://localhost:9999/?p1=v1 HTTP/1.1\r\n", req);
+        buffer.release();
     }
 
     @Test
@@ -55,6 +68,7 @@ public class HttpRequestEncoderTest {
                 HttpMethod.GET, "http://localhost/"));
         String req = buffer.toString(Charset.forName("US-ASCII"));
         assertEquals("GET http://localhost/ HTTP/1.1\r\n", req);
+        buffer.release();
     }
 
     @Test
@@ -65,6 +79,7 @@ public class HttpRequestEncoderTest {
                 HttpMethod.GET, "/"));
         String req = buffer.toString(Charset.forName("US-ASCII"));
         assertEquals("GET / HTTP/1.1\r\n", req);
+        buffer.release();
     }
 
     @Test
@@ -75,6 +90,7 @@ public class HttpRequestEncoderTest {
                 HttpMethod.GET, ""));
         String req = buffer.toString(Charset.forName("US-ASCII"));
         assertEquals("GET / HTTP/1.1\r\n", req);
+        buffer.release();
     }
 
     @Test
@@ -85,5 +101,6 @@ public class HttpRequestEncoderTest {
                 HttpMethod.GET, "/?url=http://example.com"));
         String req = buffer.toString(Charset.forName("US-ASCII"));
         assertEquals("GET /?url=http://example.com HTTP/1.1\r\n", req);
+        buffer.release();
     }
 }
--- a/microbench/src/main/java/io/netty/handler/codec/http/HttpRequestEncoderInsertBenchmark.java
+++ b/microbench/src/main/java/io/netty/handler/codec/http/HttpRequestEncoderInsertBenchmark.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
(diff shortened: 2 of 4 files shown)


microbench/src/main/java/io/netty/handler/codec/http/package-info.java
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright 2015 The Netty Project
```
normanmaurer: Fix year

doom369:
@normanmaurer done.

fenik17:
@doom369 hm... An array copying is faster than two appends. Did you check uri with long query (length 1000 chars after `?`) and short (length 3 chars)?

doom369:
@fenik17 

> An array copying is faster than two appends.

This could be easily explained. As every `append` is manual array copy that is slower than regular arraycopy.

I just did the test. Results are always better with new Ecnoder.

codec-http/src/test/java/io/netty/handler/codec/http/HttpRequestEncoderTest.java
```diff
@@ -48,6 +48,16 @@ public void testUriWithoutPath2() throws Exception {
     }
 
     @Test
+    public void testUriWithoutPath3() throws Exception {
+        HttpRequestEncoder encoder = new HttpRequestEncoder();
+        ByteBuf buffer = Unpooled.buffer(64);
+        encoder.encodeInitialLine(buffer, new DefaultHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.GET,
+                "http://localhost:9999/?p1=v1"));
```
thc202: The path should be removed?

doom369: @thc202 no, as far as I know.

thc202: I mean for the test, the name does not match what's being tested (the URI has a path but the name of the test says that it's without one).

doom369: agree, renamed.

fenik17:
@doom369 
> This could be easily explained. As every append is manual array copy that could be slower than system arraycopy. So with 2 append we have 2 manual array copies.

Yep. I meant two manual array copy instead of one manual array copy and one system arraycopy (
`new StringBuilder(uri)` - it's manual array copy of the entire string).

doom369:
> new StringBuilder(uri) - it's manual array copy of the entire string)

No, it is `System.arraycopy`

fenik17:
> No, it is System.arraycopy

You're right. I looked to CharSequence instead of String.

microbench/src/main/java/io/netty/handler/codec/http/HttpRequestEncoderInsertBenchmark.java
```diff
@@ -0,0 +1,117 @@
...
+            AsciiString method = request.method().asciiName();
+            ByteBufUtil.copy(method, method.arrayOffset(), buf, method.length());
+            buf.writeByte(SP);
+
+            // Add / as absolute path if no is present.
+            // See http://tools.ietf.org/html/rfc2616#section-5.1.2
+            String uri = request.uri();
+
+            if (uri.isEmpty()) {
+                uri += SLASH;
```
netkins: MAJOR Use "Character.toString" instead. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2131)

microbench/src/main/java/io/netty/handler/codec/http/HttpRequestEncoderInsertBenchmark.java
```diff
@@ -0,0 +1,117 @@
...
+            } else {
+                int start = uri.indexOf("://");
+                if (start != -1 && uri.charAt(0) != SLASH) {
+                    int startIndex = start + 3;
+                    // Correctly handle query params.
+                    // See https://github.com/netty/netty/issues/2732
+                    int index = uri.indexOf(QUESTION_MARK, startIndex);
+                    if (index == -1) {
+                        if (uri.lastIndexOf(SLASH) <= startIndex) {
+                            uri += SLASH;
```
netkins: MAJOR Use "Character.toString" instead. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2131)

codec-http/src/test/java/io/netty/handler/codec/http/HttpRequestEncoderTest.java
```diff
@@ -48,6 +48,16 @@ public void testUriWithoutPath2() throws Exception {
     }
 
     @Test
+    public void testUriWithEmptyPath() throws Exception {
+        HttpRequestEncoder encoder = new HttpRequestEncoder();
+        ByteBuf buffer = Unpooled.buffer(64);
+        encoder.encodeInitialLine(buffer, new DefaultHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.GET,
+                "http://localhost:9999/?p1=v1"));
+        String req = buffer.toString(Charset.forName("US-ASCII"));
+        assertEquals("GET http://localhost:9999/?p1=v1 HTTP/1.1\r\n", req);
```
Scottmitch: Looks like we should add `buffer.release();` here and for all tests in this class.

doom369: Good point. I'll fix.

doom369: @Scottmitch done.

microbench/src/main/java/io/netty/handler/codec/http/HttpRequestEncoderInsertBenchmark.java
doom369: @normanmaurer do I need to apply this fix as it was copied from existing class?

normanmaurer:
@Scottmitch want to take a final look ?

normanmaurer:
Cherry-picked into 4.1 (dd69a813d46c2db2b8071445ecddb4bf939ba2e9) and 4.0 (4c7a9652a00ac8f0c9a42a8a0b356783ca59bb88)

PR #6862, opened 2017-06-15
HttpObjectAggregator should NOT add `Content-Length: 0` for 1xx and 204 FullHttpResponses

monkey-mas:
Motivation:

If a client receives a 1xx or 204 response, `HttpObjectAggregator` adds `Content-Length` header as these responses don't have the header field according to RFC7230 section 3.3.2. This results in violation of
the specification even if a client receives an appropriate response.

Modification:

`HttpObjectAggregator` adds `Content-Length` header if it doesn't violate the specification.

Result:

`FullHttpResponse` of 1xx or 204 doesn't contain Content-Length, which satisfies RFC7230, if a server follows the specification, i.e. sending 1xx or 204 responses without `Content-Length`.

### DIFF
--- a/codec-http/src/test/java/io/netty/handler/codec/http/HttpObjectAggregatorTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/HttpObjectAggregatorTest.java
@@ -28,6 +28,7 @@ import org.junit.Test;
 import org.mockito.Mockito;
 
 import java.nio.channels.ClosedChannelException;
+import java.util.ArrayList;
 import java.util.List;
 
 import static io.netty.handler.codec.http.HttpHeadersTestUtils.of;
@@ -499,5 +500,48 @@ public class HttpObjectAggregatorTest {
         assertEquals(replacedRep.decoderResult(), aggregatedRep.decoderResult());
         aggregatedRep.release();
         replacedRep.release();
+    }
+
+    @Test
+    public void testAggregateResponseWithoutContentLength() {
+        final List<HttpResponseStatus> statusesWithoutContentLength = new ArrayList<HttpResponseStatus>();
+        statusesWithoutContentLength.add(HttpResponseStatus.CONTINUE);
+        statusesWithoutContentLength.add(HttpResponseStatus.SWITCHING_PROTOCOLS);
+        statusesWithoutContentLength.add(HttpResponseStatus.PROCESSING);
+        statusesWithoutContentLength.add(HttpResponseStatus.NO_CONTENT);
+
+        for (final HttpResponseStatus status : statusesWithoutContentLength) {
+            testAggregateResponseContentLength(status, false);
+        }
+    }
+
+    @Test
+    public void testAggregateResponseWithContentLength() {
+        final List<HttpResponseStatus> statusesWithContentLength = new ArrayList<HttpResponseStatus>();
+        statusesWithContentLength.add(HttpResponseStatus.OK);
+        statusesWithContentLength.add(HttpResponseStatus.NOT_MODIFIED);
+        statusesWithContentLength.add(HttpResponseStatus.BAD_REQUEST);
+        statusesWithContentLength.add(HttpResponseStatus.INTERNAL_SERVER_ERROR);
+
+        for (final HttpResponseStatus status : statusesWithContentLength) {
+            testAggregateResponseContentLength(status, true);
+        }
+    }
+
+    private void testAggregateResponseContentLength(HttpResponseStatus status, boolean canContainContentLength) {
+        EmbeddedChannel embedder = new EmbeddedChannel(new HttpObjectAggregator(1024 * 1024));
+
+        HttpResponse rep = new DefaultHttpResponse(HttpVersion.HTTP_1_1, status);
+        HttpContent lastChunk = LastHttpContent.EMPTY_LAST_CONTENT;
+        assertFalse(embedder.writeInbound(rep));
+        assertTrue(embedder.writeInbound(lastChunk));
+
+        FullHttpResponse aggregatedRep = embedder.readInbound();
+        assertEquals(canContainContentLength, HttpUtil.isContentLengthSet(aggregatedRep));
+        if (canContainContentLength) {
+            assertEquals(0, HttpUtil.getContentLength(aggregatedRep, -1));
+        }
+
+        aggregatedRep.release();
     }
 }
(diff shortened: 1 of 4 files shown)


monkey-mas:
Let me please ping @bryce-anderson as he understands my motivation :)

- One thing I'm not sure is if it's really important(if it's a _must_) to set `Content-Length` header for a response of HEAD request. [RFC7230 section 3.3.2.](https://tools.ietf.org/html/rfc7230#section-3.3.2) says:
```
     A server MAY send a Content-Length header field in a response to a
     HEAD request (Section 4.3.2 of [RFC7231]); a server MUST NOT send
     Content-Length in such a response unless its field-value equals the
     decimal number of octets that would have been sent in the payload
     body of a response if the same request had used the GET method.
```

- Alternatively, we can _intentionally_ remove `Content-Length` header field when 1xx or 204 responses contain the header as it violates the specification. If this sounds good, then we might write something like; 
```
if (HttpUtil.isContentLengthSet(aggregated))) {
    if (HttpUtil.mustNotHaveContentLength(aggregated)) {
        aggregated.headers().remove(CONTENT_LENGTH);
    } else {
        aggregated.headers().set(
            CONTENT_LENGTH,
            String.valueOf(aggregated.content().readableBytes()));
    }
}
```

bryce-anderson:
I think @monkey-mas has found another bug here: for a HEAD request that doesn't contain a Content-Length header, the `HttpObjectAggregator` looks to be adding one (which is presumably set to `0`) when in fact the body may be larger that 0 and the content-length header just wasn't set.

codec-http/src/test/java/io/netty/handler/codec/http/HttpObjectAggregatorTest.java
```diff
@@ -500,4 +501,47 @@ public void testReplaceAggregatedResponse() {
...
+        assertFalse(embedder.writeInbound(rep));
+        assertTrue(embedder.writeInbound(lastChunk));
+
+        FullHttpResponse aggregatedRep = embedder.readInbound();
+        assertEquals(canContainContentLength, HttpUtil.isContentLengthSet(aggregatedRep));
+        if (canContainContentLength) {
+            assertEquals(0, HttpUtil.getContentLength(aggregatedRep, -1));
+        }
+
+        aggregatedRep.release();
```
normanmaurer: @monkey-mas please also call `embedded.finish()` and assert the return value of it

codec-http/src/test/java/io/netty/handler/codec/http/HttpObjectAggregatorTest.java
```diff
@@ -500,4 +501,47 @@ public void testReplaceAggregatedResponse() {
...
+        statusesWithContentLength.add(HttpResponseStatus.NOT_MODIFIED);
+        statusesWithContentLength.add(HttpResponseStatus.BAD_REQUEST);
+        statusesWithContentLength.add(HttpResponseStatus.INTERNAL_SERVER_ERROR);
+
+        for (final HttpResponseStatus status : statusesWithContentLength) {
+            testAggregateResponseContentLength(status, true);
+        }
+    }
+
+    private void testAggregateResponseContentLength(HttpResponseStatus status, boolean canContainContentLength) {
```
normanmaurer: nit: static

monkey-mas:
> I think @monkey-mas has found another bug here: for a HEAD request that doesn't contain a Content-Length header, the HttpObjectAggregator looks to be adding one (which is presumably set to 0) when in fact the body may be larger that 0 and the content-length header just wasn't set.

Good point! Now I'm not sure if it's good for us to apply my first patch.

Just an idea but what if we simply _not_ set `Content-Length` header field by removing the code [here](https://github.com/netty/netty/blob/netty-4.1.12.Final/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectAggregator.java#L216-L226)? Then we can avoid some inconsistency like Bryce pointed out.

I'd say it _should_ be okay because
- clients shouldn't rely on `Content-Length` header field as it's possible that a response to a HEAD request doesn't contain one as the RFC says.
- once a response is decoded to Netty's `Response` data, we don't have to depend on `Content-Length` header field to obtain the size of it? (I'm not sure though.) If this is correct then an _aggregated_ response doesn't necessarily have a `Content-Length` header field.

What's more we can
- respect server's actual responses
- solve the corner case for 1xx and 204 responses (with regard to `Content-Length` existence) without adding my patch as we don't set the header field when calling `finishAggregation(...)`.

What do you think?
@normanmaurer @bryce-anderson

bryce-anderson:
I would advocate for simply ensuring that the message is not a HEAD response before setting the content-length.

monkey-mas:
> I would advocate for simply ensuring that the message is not a HEAD response before setting the content-length.

Cool! Do you think it's possible to tell if a response is one for a HEAD request? `HttpObjectAggregator` doesn't know if a response is for a HEAD request I suppose.

bryce-anderson:
> Cool! Do you think it's possible to tell if a response is one for a HEAD request?
> HttpObjectAggregator doesn't know if a response is for a HEAD request I suppose.

To be honest I didn't fully look into it for this PR. Typically this is done in Netty by observing the inbound messages and keeping a record of the important aspects. An example can be found [here](https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentEncoder.java#L108). That Said, maybe addressing that in this PR is a bit much and would be better as a separate commit or maybe not worth the effort at all.

monkey-mas:
> maybe addressing that in this PR is a bit much and would be better as a separate commit or maybe not worth the effort at all.

Got it, thanks for your comment :) and sorry for my late response. Alternatively we might create `HttpClientObjectAggregator` and `HttpServerObjectAggregator` (instead of just having one `HttpObjectAggregator` to divide its responsibility) like we have `HttpServerCodec` and `HttpClientCodec` to achieve our goal. Anyway, I'll give it a shot. Let me ping you once I'm ready for it.

normanmaurer:
@monkey-mas any news ?

normanmaurer:
@monkey-mas @bryce-anderson ping

monkey-mas:
@normanmaurer 
Sorry for not being able to reply sooner. Will check this weekend at the latest :-)

normanmaurer:
@monkey-mas let me close this for now and please reopen if you are ready and still want to contribute it.

monkey-mas:
Sorry about it. Haven't got enough time recently... Let me reopen it once I'm ready.

Issue #5861, opened 2016-09-25
HttpUtil.getContentLength(HttpMessage, long) throws unexpected NumberFormatException


ewie:
The Javadocs of `HttpUtil.getContentLength(HttpMessage, long)` and its `int` overload state that the provided default value is returned if the Content-Length value is not a number. `NumberFormatException` is thrown instead [1] causing the method to behave like `HttpUtil.getContentLength(HttpMessage)` where the exception is explicitly documented for this case.

[1] https://github.com/netty/netty/blob/149916d052180dd2b0e0e98598f349454ea200ba/codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java#L187

normanmaurer:
@ewie sounds like a bug... let me fix it.

normanmaurer:
@ewie should be fixed by https://github.com/netty/netty/pull/5877

normanmaurer:
Fixed by https://github.com/netty/netty/pull/5877

PR #6866, opened 2017-06-19
Remove content-length header leniency

jasontedor:
Motivation:

If the content-length does not parse as a number, leniency causes this to instead be parsed as the default value. This leads to bodies being silently ignored on requests which can be incredibly dangerous. Instead, if the content-length header is invalid, an exception should be thrown for upstream handling.

Modifications:

This commit removes the leniency in parsing the content-length header by allowing a number format exception, if thrown, to escape from the method rather than falling back to the default value.

Result:

In invalid content-length header will not be silently ignored.

Relates #5861

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectAggregator.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectAggregator.java
@@ -151,7 +151,11 @@ public class HttpObjectAggregator
 
     @Override
     protected boolean isContentLengthInvalid(HttpMessage start, int maxContentLength) {
-        return getContentLength(start, -1L) > maxContentLength;
+        try {
+            return getContentLength(start, -1L) > maxContentLength;
+        } catch (final NumberFormatException e) {
+            return false;
+        }
     }
 
     @Override
--- a/codec-http/src/test/java/io/netty/handler/codec/http/HttpUtilTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/HttpUtilTest.java
@@ -25,10 +25,14 @@ import java.util.Collections;
 import java.util.List;
 
 import static io.netty.handler.codec.http.HttpHeadersTestUtils.of;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.hasToString;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 public class HttpUtilTest {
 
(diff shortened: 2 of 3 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectAggregator.java
```diff
@@ -151,7 +151,11 @@ protected boolean isAggregated(HttpObject msg) throws Exception {
 
     @Override
     protected boolean isContentLengthInvalid(HttpMessage start, int maxContentLength) {
-        return getContentLength(start, -1L) > maxContentLength;
+        try {
+            return getContentLength(start, -1L) > maxContentLength;
+        } catch (final NumberFormatException e) {
+            return false;
```
fenik17: why `false`?

fenik17: also, this change can lead to incorrect error message "content length exceeded xxx bytes." if `NumberFormatException` will be thrown.

jasontedor: Because `true` here means that the content length exceeds the max content length, that's not the case if the content length is invalid, so false.

fenik17: Then the name of the method is incorrect. Unfortunately, this is a public API...

jasontedor: No, it doesn't. Returning `false` prevents the [oversized message handler from incorrectly being invoked in MessageAggregator](https://github.com/netty/netty/blob/7460d90a67bd9724d8fdac839c384bf36cbbce37/codec/src/main/java/io/netty/handler/codec/MessageAggregator.java#L236-L240), instead [allowing the number format exception to escape because the decoder result will be failure](https://github.com/netty/netty/blob/7460d90a67bd9724d8fdac839c384bf36cbbce37/codec/src/main/java/io/netty/handler/codec/MessageAggregator.java#L242).

fenik17: > No, it doesn't. 

Yes, I already understood. I was confused by the name of the method..

jasontedor: Yes, the method name is awful but there's nothing we can do about it. 😦

Scottmitch: `invalid` can have multiple meanings in this context ... check out the javadoc for more clarity.

codec-http/src/test/java/io/netty/handler/codec/http/HttpUtilTest.java
```diff
@@ -130,12 +134,42 @@ public void testGetMimeType() {
...
-        assertNull(message.headers().get(HttpHeaderNames.CONTENT_LENGTH));
+    public void testGetContentLengthThrowsNumberFormatException() {
+        final HttpMessage message = new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.OK);
+        message.headers().set(HttpHeaderNames.CONTENT_LENGTH, "bar");
+        try {
+            HttpUtil.getContentLength(message);
+            fail();
+        } catch (final NumberFormatException e) {
+            // a number format exception is expected here
+            assertThat(e, hasToString(containsString("For input string: \"bar\"")));
```
Scottmitch: can you please avoid checking for string contents. This may vary based upon local and is generally fragile. IMHO checking for `NumberFormatException` should be "good enough" and not subject to false failures if implicitly defined strings change behind the scenes.

codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectAggregator.java
jasontedor: I agree, the Javadocs help but it's exactly because "invalid" has multiple meanings as you say that it's ambiguous and thus leads to the confusion as above. 😇

codec-http/src/test/java/io/netty/handler/codec/http/HttpUtilTest.java
jasontedor: Okay, I will address tomorrow. Thanks for the feedback!

jasontedor: @Scottmitch I made the change that you requested (I squashed and force pushed already).

jasontedor:
@Scottmitch Thanks, back to you.

Scottmitch:
4.1 (9ad74e72e)

Scottmitch:
thanks @jasontedor !

jasontedor:
Thanks @Scottmitch!

normanmaurer:
@jasontedor <3

PR #6867, opened 2017-06-19
NullPointerException in Lz4FrameEncoder

Scottmitch:
Motivation:
Lz4FrameEncoder maintains internal state, but the life cycle of the buffer is not consistently managed. The buffer is allocated in handlerAdded but freed in close, but the buffer can still be used until handlerRemoved is called.

Modifications:
- Move the cleanup of the buffer from close to handlerRemoved
- Explicitly throw an EncoderException from Lz4FrameEncoder if the encode operation has finished and there isn't enough space to write data

Result:
No more NPE in Lz4FrameEncoder on the buffer.

### DIFF
--- a/codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
+++ b/codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
@@ -15,30 +15,46 @@
  */
 package io.netty.handler.codec.compression;
 
-import java.io.InputStream;
-import java.util.zip.Checksum;
-
-import io.netty.util.internal.PlatformDependent;
-import org.junit.Assert;
-import org.junit.Before;
-import org.junit.Test;
-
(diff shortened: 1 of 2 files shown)


codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameEncoderTest.java
```diff
@@ -221,4 +237,73 @@ public void testAllocatingAroundBlockSize() {
         Assert.assertTrue(channel.releaseOutbound());
         Assert.assertFalse(channel.releaseInbound());
     }
+
+    @Test
+            //(timeout = 3000)
```
normanmaurer: @Scottmitch left-over ?

Scottmitch: yip just fixed this :)

Scottmitch:
4.1 (14ea69cdc)

PR #6868, opened 2017-06-19
Deduplicate and simplify code in HttpPostMultipartRequestDecoder

fenik17:
Motivation:

- A `HttpPostMultipartRequestDecoder` contains two pairs of the same methods: `readFileUploadByteMultipartStandard`+`readFileUploadByteMultipart` and `loadFieldMultipartStandard`+`loadFieldMultipart`.
- These methods use `NotEnoughDataDecoderException` to detecting not last data chunk (exception handling is very expensive).
- These methods can be greatly simplified.
- Methods `loadFieldMultipart` and `loadFieldMultipartStandard` has an unnecessary catching for the `IndexOutOfBoundsException`.

Modifications:

- Remove duplicate methods.
- Replace handling `NotEnoughDataDecoderException` by the return of a boolean result.
- Simplify code.

Result:

The code is cleaner and easier to support. Less exception handling logic.


normanmaurer:
@Scottmitch @trustin PTAL as well.

normanmaurer:
Cherry-picked into 4.1 (3e9f617504adb6f3ac984a9320020dd2c202739e)

PR #6869, opened 2017-06-20
Deploy transport-native-unix-common-tests

normanmaurer:
Motivation:

To be able to easily build only one of the native sub-modules its needed that all the dependencies can be fetched from maven. At the moment we dont deploy transport-native-unix-common and so an attempt to just build for example the native epoll transport fails with:

[ERROR] Failed to execute goal on project netty-transport-native-epoll: Could not resolve dependencies for project io.netty:netty-transport-native-epoll:jar:4.1.13.Final-SNAPSHOT: Could not find artifact io.netty:netty-transport-native-unix-common-tests:jar:4.1.13.Final-SNAPSHOT in sonatype-nexus-snapshots (https://oss.sonatype.org/content/repositories/snapshots) -> [Help 1]

Modifications:

Deploy jar

Result:

All dependencies on maven repository.

### DIFF
--- a/transport-native-unix-common-tests/pom.xml
+++ b/transport-native-unix-common-tests/pom.xml
@@ -29,11 +29,6 @@
     Tests for the static library which contains common unix utilities.
   </description>
 
-  <properties>
-    <!-- Don't deploy this artifact to Maven Central -->
-    <maven.deploy.skip>true</maven.deploy.skip>
-  </properties>
-
   <dependencies>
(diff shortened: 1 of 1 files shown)


normanmaurer:
@Scottmitch yes... pulling in

normanmaurer:
Cherry-picked into 4.1 (568fa998b1dfd0af5c5571d89b61beb6d6e6c54b)

PR #6874, opened 2017-06-20
Optimizations in QueryStringEncoder

fenik17:
Motivation:

A life cycle of QueryStringEncoder is simple: create, append params, convert to String. Current realization collect params in the list, and calculate an URI string in `toString` method. We can simplify this: don't store params to the list, and immediately append parameters to the `StringBuilder`.

Modifications:

- Remove list for params and remove a tuple class `Param`.
- Use one common `StringBuilder` and append parameters into it.
- Resolve `TODO` in the `encodeParam` method.

Result:

Less allocations (no `ArrayList`, no `Param` tuples). Second `toString` call is faster.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/QueryStringEncoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/QueryStringEncoder.java
@@ -23,9 +23,6 @@ import java.net.URISyntaxException;
 import java.net.URLEncoder;
 import java.nio.charset.Charset;
 import java.nio.charset.UnsupportedCharsetException;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.regex.Pattern;
 
 /**
  * Creates an URL-encoded URI from a path string and key-value parameter pairs.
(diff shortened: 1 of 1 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/QueryStringEncoder.java
```diff
@@ -86,42 +92,30 @@ public URI toUri() throws URISyntaxException {
...
-            this.name = name;
+        // replace all '+' with "%20"
+        int idx = s.indexOf('+');
+        if (idx == -1) {
+            sb.append(s);
+            return;
+        }
+        sb.append(s, 0, idx).append("%20");
+        int size = s.length();
+        for (idx++; idx < size; idx++) {
```
normanmaurer: move the first `idx++` out of the for declaration to make it more clear ?

fenik17: done

Scottmitch:
4.1 (d672a5a48) 4.0 (76df5a591)

PR #6875, opened 2017-06-20
OpenSSL CHACHA20 CipherSuiteConverter updates

Scottmitch:
Motivation:
For historical reasons OpenSSL's internal naming convention for CHACHA20 based cipher suites does not include the HMAC algorithm in the cipher name. This will prevent the CHACHA20 cipher suites from being used if the RFC cipher names are specified.

Modifications:
- Add a special case for CHACHA20 cipher name conversions in CipherSuiteConverter
- Update OPENSSL_CIPHERSUITE_PATTERN to accommodate the new naming scheme for CHACHA20 cipher suites

Result:
CipherSuiteConverter now works with CHACHA20 cipher suites.


normanmaurer:
Cherry-picked into 4.1 (6d029ad3acb2214192c73657d6febcbfa44dadf6) and 4.0 (05e5aaf430)

PR #6876, opened 2017-06-21
HTTP/2 HelloWorld clear text NPE

Scottmitch:
Motivation:
HelloWorldHttp2Handler throws a NPE when converting from HTTP/1.x headers to HTTP/2 headers because there is no Host header.

Modifications:
- HelloWorldHttp2Handler should check if the Host header is present before setting it in the HTTP/2 headers

Result:
No more NPE in HelloWorldHttp2Handler.


normanmaurer:
Cherry-picked into 4.1 (6b25909f6b586ac22c32e570ed82cc8fc7b188c8)

PR #6877, opened 2017-06-21
Http2FrameLogger Updates

Scottmitch:
Motivation:
The Http2FrameLogger uses a custom format when logging events. We should use the more familiar format of 'channel event type: details' and single line logging for more consistent debugging.

Modifications:
- Http2FrameLogger should not use a StringBuilder and instead should directly use the Logger
- Http2FrameLogger should use the more consistent format defined above

Result:
Http2FrameLogger's logging formate is more consistent with other log events.


normanmaurer:
@trustin PTAL as well

Scottmitch:
4.1 (5934ae8fd)

I can address concerns in a followup PR

PR #6878, opened 2017-06-21
Fix compiler warnings in netty Epoll and unix common

carl-mastrangelo:
Motivation:
Google requires stricter compilation by adding `-Werror` and enabling many other warnings.   

Modification:
Fix compiler warnings.  the explanation for each change is in its commit.

Result:
Easier compilation.

cc: @Scottmitch

### DIFF
--- a/transport-native-epoll/src/main/c/netty_epoll_native.c
+++ b/transport-native-epoll/src/main/c/netty_epoll_native.c
@@ -315,7 +315,9 @@ static jstring netty_epoll_native_kernelVersion(JNIEnv* env, jclass clazz) {
 }
 
 static jboolean netty_epoll_native_isSupportingSendmmsg(JNIEnv* env, jclass clazz) {
-    if (sendmmsg) {
+    // Use & to avoid warnings with -Wtautological-pointer-compare when sendmmsg is
+    // not weakly defined.
+    if (&sendmmsg != NULL) {
         return JNI_TRUE;
     }
     return JNI_FALSE;
--- a/transport-native-unix-common/src/main/c/netty_unix_socket.c
+++ b/transport-native-unix-common/src/main/c/netty_unix_socket.c
@@ -581,7 +581,7 @@ static jint netty_unix_socket_connectDomainSocket(JNIEnv* env, jclass clazz, jin
 static jint netty_unix_socket_recvFd(JNIEnv* env, jclass clazz, jint fd) {
     int socketFd;
     struct msghdr descriptorMessage = { 0 };
-    struct iovec iov[1] = { 0 };
+    struct iovec iov[1] = { { 0 } };
     char control[CMSG_SPACE(sizeof(int))] = { 0 };
     char iovecData[1];
 
@@ -629,7 +629,7 @@ static jint netty_unix_socket_recvFd(JNIEnv* env, jclass clazz, jint fd) {
 
 static jint netty_unix_socket_sendFd(JNIEnv* env, jclass clazz, jint socketFd, jint fd) {
     struct msghdr descriptorMessage = { 0 };
-    struct iovec iov[1] = { 0 };
+    struct iovec iov[1] = { { 0 } };
     char control[CMSG_SPACE(sizeof(int))] = { 0 };
     char iovecData[1];
 


transport-native-epoll/src/main/c/netty_epoll_native.c
```diff
@@ -315,7 +315,7 @@ static jstring netty_epoll_native_kernelVersion(JNIEnv* env, jclass clazz) {
 }
 
 static jboolean netty_epoll_native_isSupportingSendmmsg(JNIEnv* env, jclass clazz) {
-    if (sendmmsg) {
+    if (&sendmmsg) {
```
Scottmitch: > When compiling on Linux, `sendmmsg` is always present, so the function is always nonnull.

The reason why we do the soft linking is to accommodate for older kernels/glibs ([added in glibc 2.14 and kernel 3.0](http://man7.org/linux/man-pages/man2/sendmmsg.2.html))where this function is not present.

IIUC what we want here is the following:

```c
return sendmmsg != NULL ? JNI_TRUE : JNI_FALSE;
```

carl-mastrangelo: But if the function is present, the compiler will think the if is always true and warn.  The function signature is only weak if that preprocessor define is set.  In my case, it is not defined as weak, which makes the function always non NULL.

The code proposed has the same problem.  If you try compiling this library with -Werror,-Wpointer-bool-conversion on a recent Linux kernel I think you will see what I am talking about.

normanmaurer: @carl-mastrangelo I am a bit confused... so if you change it to use `&` will this still work as what we tried to do ?

carl-mastrangelo: Yes, it just silences the warning.

normanmaurer: @Scottmitch happy as well ?

Scottmitch: There are 2 things at play here:

1. Compile time availability (guarded by `IO_NETTY_SENDMMSG_NOT_FOUND`)
2. Link time availability (guarded by `__attribute__((weak))`)

IIUC your configuration makes it such that `IO_NETTY_SENDMMSG_NOT_FOUND` is not defined and therefore you have the function pointer `sendmmsg` defined at compile time. However this doesn't say anything about what happens at run time (imagine linking this artifact on a pre 3.0 kernel). The compiler warning doesn't doesn't seem to be taking into account the soft linking portion of the method definition. Are you sure that the address for the method pointer will also be `0` at run time? This seems to be the case in the environment I tried (see code below) (gcc 5.4.0, ldd 2.23, Ubuntu 16.10) but I'm only confident that the pointer will point to `NULL`. Is there a reference somewhere you can point me to?


```c
#include <stdlib.h>
#include <stdio.h>

extern int softlinkmethod() __attribute__((weak));

int main() {
    int* f = NULL;
    printf("ref: %p deref: %p\n", f, &f);
    printf("ref: %p deref: %p\n", softlinkmethod, &softlinkmethod);
    if (softlinkmethod != NULL) {
        printf("1 in if\n");
    } else {
        printf("2 in else\n");
    }
    if (&softlinkmethod) {
        printf("3 in if\n");
    } else {
        printf("4 in else\n");
    }
    return 0;
}

/*int softlinkmethod() {
  return 1;
}*/
```

#### stdout
```
ref: (nil) deref: 0x7ffca533ba70
ref: (nil) deref: (nil)
2 in else
4 in else
```

Scottmitch: Here is something that maybe useful (for clang):

https://clang.llvm.org/docs/AttributeReference.html
> A weakly-linked declaration may or may not be present a run-time, and a program can determine whether the declaration is present by checking whether the address of that declaration is non-NULL.

Scottmitch: https://developer.apple.com/library/content/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/WeakLinking.html

Using Weakly Linked Symbols
```c
extern int MyWeakLinkedFunction() __attribute__((weak_import));
 
int main()
{
    int result = 0;
 
    if (MyWeakLinkedFunction != NULL)
    {
        result = MyWeakLinkedFunction();
    }
 
    return result;
}
```

carl-mastrangelo: But this solution doesn't work.  The symbol is *not* weakly defined.  If `IO_NETTY_SENDMMSG_NOT_FOUND` is set, it's a compile time error for redefining it:

```
#define _GNU_SOURCE
#include <sys/socket.h>

struct mmsghdr {
    struct msghdr msg_hdr;  /* Message header */
    unsigned int  msg_len;  /* Number of bytes transmitted */
};

extern int sendmmsg(int sockfd, struct mmsghdr* msgvec, unsigned int vlen, unsigned int flags) __attribute__((weak));

int main(int argc, char ** argv) {
  return 0;
}
```
```
Wed Jun 21 03:02:12$ gcc send.c 
send.c:5:8: error: redefinition of ‘struct mmsghdr’
 struct mmsghdr {
        ^
In file included from send.c:2:0:
/usr/include/x86_64-linux-gnu/sys/socket.h:101:8: note: originally defined here
 struct mmsghdr
        ^
send.c:12:12: error: conflicting types for ‘sendmmsg’
 extern int sendmmsg(int sockfd, struct mmsghdr* msgvec, unsigned int vlen, unsigned int flags) __attribute__((weak));
            ^
In file included from send.c:2:0:
/usr/include/x86_64-linux-gnu/sys/socket.h:193:12: note: previous declaration of ‘sendmmsg’ was here
 extern int sendmmsg (int __fd, struct mmsghdr *__vmessages,
            ^
```

Also, to be clear, sendmmsg is *NOT* a function pointer, it is a function.  It cannot be null if it is defined.  Did you try compiling with `-Wpointer-bool-conversion` ? 

The real issue is that this file is selectively defining the method as weak, which is the true problem

I talked with several other C++ people before I sent this out, and this seems to be the least worst solution.  The next bets solution would be something like:


```
int (*sendmmsg_fp)(int sockfd, struct mmsghdr* msgvec, unsigned int vlen, unsigned int flags);

#ifdef IO_NETTY_SENDMMSG_NOT_FOUND
extern int sendmmsg(int sockfd, struct mmsghdr* msgvec, unsigned int vlen, unsigned int flags) __attribute__((weak));
sendmmsg_fp = sendmmsg;

#ifndef __USE_GNU
struct mmsghdr {
    struct msghdr msg_hdr;  /* Message header */
    unsigned int  msg_len;  /* Number of bytes transmitted */
};
#endif
#else
sendmmsg_fp = sendmmsg;
#endif
```

Then, everywhere else in this file, use `sendmmsg_fp`.  However, that's a lot more work, and not as clean.

Scottmitch: > But this solution doesn't work. The symbol is not weakly defined. If IO_NETTY_SENDMMSG_NOT_FOUND is set, it's a compile time error for redefining it:

I'm not suggesting redefining anything. The purpose of `IO_NETTY_SENDMMSG_NOT_FOUND` is to avoid this and should remain.

> The real issue is that this file is selectively defining the method as weak, which is the true problem

This is the approach taken to supporting compiling on pre 3.0 kernels, post 3.0 kernels, and linking artifacts built on pre 3.0 kernels on post 3.0 kernels (and visa versa). Are you aware of an alternative approach?

> Did you try compiling with -Wpointer-bool-conversion ?

I think the warning in question is `-Wtautological-pointer-compare`? We should be explicitly checking against `NULL` instead of the implicit boolean check either way. clang actually gives the below recommendation ... this warning seems to be a bit overzealous in this case do to the linking nature of the problem but I guess it shouldn't hurt (assuming the address of operator is always safe the function isn't defined ... as I said seems to be this way in practice but I would be interested in a link which explains in more detail)

```
scott@scott-lubuntu:~$ clang -Werror test.c -o test
test.c:14:10: error: comparison of function 'sendmmsg' not equal to a null pointer is always true [-Werror,-Wtautological-pointer-compare]
     if (sendmmsg != NULL) {
         ^~~~~~~~    ~~~~
test.c:14:10: note: prefix with the address-of operator to silence this warning
     if (sendmmsg != NULL) {
         ^
         &
```

carl-mastrangelo: Hmm, you are right about the warning.  When I change `if(sendmmsg)` to `if(sendmmsg != null)` I get the same warning as you do.

I agree the warning is overzealous.  Talking to other C++ people around me, the point of the warning is to protect against accidentally doing:

```
if (sendmssg()) { ... }
```

Which is apparently a common mistake.  

I have upodated the code to use !=, PTAL.

normanmaurer:
@carl-mastrangelo can you also please squash FTW :)

carl-mastrangelo:
@normanmaurer done

normanmaurer:
Will merge once CI completes

normanmaurer:
Cherry-picked into 4.1 (b98561552292240aed69b80bd5bc1cdca67a9137) and 4.0 (101757e861)

PR #6884, opened 2017-06-21
Make Native loading work better with Java 8

carl-mastrangelo:
Motivation:
Enable static linking for Java 8.  These commits are the same as those introduced to netty tcnative.  The goal is to allow lots of JNI libraries to be statically linked together without having conflict `JNI_OnLoad` methods.  

Modification:
* add JNI_OnLoad suffixes to enable static linking
* Add static names to the list of libraries that try to be loaded
* Enable compiling with JNI 1.8
* Sort includes
 
Result:
Enable statically linked JNI code.

cc: @nmittler @Scottmitch

### DIFF
--- a/transport-native-epoll/src/main/c/netty_epoll_native.c
+++ b/transport-native-epoll/src/main/c/netty_epoll_native.c
@@ -36,12 +36,14 @@
 #include <inttypes.h>
 #include <link.h>
 #include <time.h>
-#include "netty_unix_filedescriptor.h"
-#include "netty_unix_socket.h"
-#include "netty_unix_errors.h"
-#include "netty_unix_util.h"
-#include "netty_unix_limits.h"
+
 #include "netty_epoll_linuxsocket.h"
+#include "netty_unix_errors.h"
+#include "netty_unix_filedescriptor.h"
+#include "netty_unix_jni.h"
+#include "netty_unix_limits.h"
+#include "netty_unix_socket.h"
+#include "netty_unix_util.h"
 
 // TCP_FASTOPEN is defined in linux 3.7. We define this here so older kernels can compile.
 #ifndef TCP_FASTOPEN
@@ -573,7 +575,7 @@ static jint netty_epoll_native_JNI_OnLoad(JNIEnv* env, const char* packagePrefix
       return JNI_ERR;
     }
 
-    return JNI_VERSION_1_6;
+    return NETTY_JNI_VERSION;
 }
 
 static void netty_epoll_native_JNI_OnUnLoad(JNIEnv* env) {
@@ -584,12 +586,14 @@ static void netty_epoll_native_JNI_OnUnLoad(JNIEnv* env) {
     netty_epoll_linuxsocket_JNI_OnUnLoad(env);
 }
 
-jint JNI_OnLoad(JavaVM* vm, void* reserved) {
+// Invoked by the JVM when statically linked
+jint JNI_OnLoad_netty_transport_native_epoll(JavaVM* vm, void* reserved) {
     JNIEnv* env;
-    if ((*vm)->GetEnv(vm, (void**) &env, JNI_VERSION_1_6) != JNI_OK) {
+    if ((*vm)->GetEnv(vm, (void**) &env, NETTY_JNI_VERSION) != JNI_OK) {
         return JNI_ERR;
     }
-
+    char* packagePrefix = NULL;
+#ifndef NETTY_NOT_DYNAMIC
     Dl_info dlinfo;
     jint status = 0;
     // We need to use an address of a function that is uniquely part of this library, so choose a static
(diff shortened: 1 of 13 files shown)


transport-native-epoll/src/main/java/io/netty/channel/epoll/Native.java
```diff
@@ -188,7 +188,11 @@ private static void loadNativeLibrary() {
         if (!name.startsWith("linux")) {
             throw new IllegalStateException("Only supported on Linux");
         }
-        NativeLibraryLoader.load("netty-transport-native-epoll", PlatformDependent.getClassLoader(Native.class));
+        String []libraryNames = new String[]{
+          "netty-transport-native-epoll",
+          "netty_transport_native_epoll"
```
Scottmitch: will it break folks to just change to using underscores?

carl-mastrangelo: yes, since the name of the so library extracted has  dashes in it.  If you want to change that name too, it would be possible.

carl-mastrangelo:
I can make the changes in kqueue, but didn't because I don't have a way of actually testing it.  I'll send a commit in a sec.

carl-mastrangelo:
KQueue changes made, PTAL.

transport-native-epoll/src/main/java/io/netty/channel/epoll/Native.java
Scottmitch: > If you want to change that name too

Yes this was what I had in mind. We should have control over the library name and the library loading. I don't think we need to support loading mixed versions (older version libraries with the `-` in the names with newer version library loader which expects `_` in the name)  ... let me know if I'm missing something.

Scottmitch:
> I can make the changes in kqueue, but didn't because I don't have a way of actually testing it. I'll send a commit in a sec.

Somebody needs a Mac 😜

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/Native.java
```diff
@@ -100,7 +100,11 @@ private static void loadNativeLibrary() {
         if (!name.startsWith("mac") && !name.contains("bsd") && !name.startsWith("darwin")) {
             throw new IllegalStateException("Only supported on BSD");
         }
-        NativeLibraryLoader.load("netty-transport-native-kqueue", PlatformDependent.getClassLoader(Native.class));
+        String []libraryNames = new String[]{
+            "netty-transport-native-kqueue",
+            "netty_transport_native_kqueue"
+        }
```
Scottmitch: semi-colon plz

transport-native-epoll/src/main/java/io/netty/channel/epoll/Native.java
normanmaurer: +1 to what @Scottmitch said

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/Native.java
```diff
@@ -100,7 +100,11 @@ private static void loadNativeLibrary() {
         if (!name.startsWith("mac") && !name.contains("bsd") && !name.startsWith("darwin")) {
             throw new IllegalStateException("Only supported on BSD");
         }
-        NativeLibraryLoader.load("netty-transport-native-kqueue", PlatformDependent.getClassLoader(Native.class));
+        String []libraryNames = new String[]{
```
normanmaurer: also nit: `new String[] {`

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/Native.java
carl-mastrangelo: D'oh, my Go code is showing.

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/Native.java
carl-mastrangelo: Done, here and in epoll.

transport-native-epoll/src/main/java/io/netty/channel/epoll/Native.java
carl-mastrangelo: grepping through I don't see where the shared object gets its name.  Is it just derived from the artifact Id?  I wouldn't mind making the change in this PR, but I'm afraid I don't know maven that well.

Scottmitch: I would need to experiment if the `maven-hawtjni-plugin` makes this configurable ([1] maybe useful), and if not we could always just mv the artifact after it is generated (via a `maven-antrun-plugin` execution).

[1] https://github.com/fusesource/hawtjni/blob/477c8ccac78c3695ebcf6299d8b201adb3394d34/maven-hawtjni-plugin/src/main/java/org/fusesource/hawtjni/maven/PackageJarMojo.java#L60

normanmaurer:
@carl-mastrangelo can you please squash... ?

normanmaurer:
@Scottmitch I think we can merge this as it is ... I will check if we can change the name of the native lib easily as a followup

transport-native-epoll/src/main/c/netty_epoll_native.c
```diff
@@ -614,11 +618,21 @@ jint JNI_OnLoad(JavaVM* vm, void* reserved) {
...
         // Something is wrong but nothing we can do about this :(
         return;
     }
     netty_epoll_native_JNI_OnUnLoad(env);
 }
+
+void JNI_OnUnload(JavaVM* vm, void* reserved) {
+  JNI_OnUnload_netty_transport_native_epoll(vm, reserved);
+}
+
```
Scottmitch: extra newline?

carl-mastrangelo:
@normanmaurer Done

transport-native-epoll/src/main/c/netty_epoll_native.c
carl-mastrangelo: @Scottmitch  Dpme

normanmaurer:
@carl-mastrangelo ok... just try to build on macOS as well... if all works I will merge 👍

normanmaurer:
Cherry-picked into 4.1 (83de77fbe5385130ba9afc666fecbe9b59bd8032)

Issue #6844, opened 2017-06-12
DNS resolution in a Java Docker container fails


jacek-:
Netty version: `4.1.12.Final`
Docker image: `openjdk:8-jre`

Context:
Running DNS resolution in a Java docker container fails. DNS resolution should complete successfully and return the IP address of the sibling docker container. Instead it throws the following exception:

    Exception in thread "main" java.util.concurrent.ExecutionException: java.net.UnknownHostException: failed to resolve 'recomp-hdb' after 3 queries
        at io.netty.util.concurrent.AbstractFuture.get(AbstractFuture.java:41)
        at TestNetty.main(TestNetty.java:15)
    Caused by: java.net.UnknownHostException: failed to resolve 'recomp-hdb' after 3 queries
        at io.netty.resolver.dns.DnsNameResolverContext.finishResolve(DnsNameResolverContext.java:688)
        at io.netty.resolver.dns.DnsNameResolverContext.tryToFinishResolve(DnsNameResolverContext.java:620)
        at io.netty.resolver.dns.DnsNameResolverContext$3.operationComplete(DnsNameResolverContext.java:313)
        at io.netty.util.concurrent.DefaultPromise.notifyListener0(DefaultPromise.java:507)
        ...

DNS resolution with plain Java code works as expected and returns the IP address.

The Netty-based code that throws the exception:

    import io.netty.channel.nio.NioEventLoopGroup;
    import io.netty.channel.socket.nio.NioDatagramChannel;
    import io.netty.resolver.dns.DnsNameResolverBuilder;
    import io.netty.resolver.dns.DnsNameResolver;
    import io.netty.resolver.dns.DnsServerAddresses;
    public class TestNetty {
      private static NioEventLoopGroup bossGroup = new NioEventLoopGroup();
      
      public static void main(String[] args) throws Exception {
        DnsNameResolver res = new DnsNameResolverBuilder(bossGroup.next())
            .channelType(NioDatagramChannel.class)
            .build();

        System.out.println(res.resolve("recomp-hdb").await().get());
      }
    }

The plain Java code which works fine:

    import java.net.InetAddress;
    public class Test {
        public static void main(String args[]) throws Exception {
                InetAddress address = InetAddress.getByName("recomp-hdb");
                System.out.println(address.getHostAddress());
        }
    }

Steps to reproduce:
1. Create a docker compose file that defines two docker containers, one of which is based on openjdk image.
2. Build the code shown above, changing the name of the sibling container according to your docker compose.
[...]

PR #6885, opened 2017-06-21
DNS Resolver Search Domain Bugs

Scottmitch:
Motivation:
The DNS resolver supports search domains. However the ndots are not correctly enforced. The search domain should only be appended under the following scenario [1]:

> Resolver queries having fewer than ndots dots (default is 1) in them will be attempted using each component of the search path in turn until a match is found.

The DNS resolver current appends the search domains if ndots is 0 which should never happen (because no domain can have less than 0 dots).

[1] https://linux.die.net/man/5/resolv.conf

Modifications:
- Parse /etc/resolv.conf to get the default value for ndots on Unix platforms
- The search domain shouldn't be used if ndots is 0
- Avoid failing a promise to trigger the search domain queries in DnsNameResolverContext#resolve

Result:
More correct usage of search domains in the DNS resolver.
Fixes https://github.com/netty/netty/issues/6844.

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
@@ -64,9 +64,9 @@ import java.util.Iterator;
 import java.util.List;
 
 import static io.netty.resolver.dns.DefaultDnsServerAddressStreamProvider.DNS_PORT;
+import static io.netty.resolver.dns.UnixResolverDnsServerAddressStreamProvider.parseEtcResolverFirstNdots;
 import static io.netty.util.internal.ObjectUtil.checkNotNull;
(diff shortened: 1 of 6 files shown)


Scottmitch:
@vietj - PTAL

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -129,6 +130,14 @@
             searchDomains = EmptyArrays.EMPTY_STRINGS;
         }
         DEFAULT_SEARCH_DOMAINS = searchDomains;
+
+        int ndots;
+        try {
+            ndots = parseEtcResolverFirstNdots();
+        } catch (Exception e) {
```
normanmaurer: nit: `e -> ignore`

Scottmitch:
4.1 (6cd086050)

PR #6887, opened 2017-06-22
Replace DnsNameResolverContext#trace special code with an implementat…

Scottmitch:
…ion of DnsQueryLifecycleObserver

Motivation:
DnsQueryLifecycleObserver is designed to capture the life cycle of every query. DnsNameResolverContext has a custom trace mechanism which consists of a StringBuilder and manual calls throughout the class. We can remove some special case code in DnsNameResolverContext and instead use a special implementation of DnsQueryLifecycleObserver when trace is enabled.

Modifications:
- Remove all references to the boolean trace variables in DnsNameResolverContext and DnsNameResolver
- Introduce TraceDnsQueryLifecycleObserver which will be used when trace is enabled and will log similar data as what trace currently provides

Result:
Less special case code in DnsNameResolverContext and instead delegate to TraceDnsQueryLifecycleObserver to capture trace information.

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/BiDnsQueryLifecycleObserver.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/BiDnsQueryLifecycleObserver.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.resolver.dns;
+
+import io.netty.channel.ChannelFuture;
+import io.netty.handler.codec.dns.DnsQuestion;
+import io.netty.handler.codec.dns.DnsResponseCode;
+import io.netty.util.internal.UnstableApi;
+
+import java.net.InetSocketAddress;
+import java.util.List;
+
+import static io.netty.util.internal.ObjectUtil.checkNotNull;
+
+/**
+ * Combines two {@link DnsQueryLifecycleObserver} into a single {@link DnsQueryLifecycleObserver}.
+ */
+@UnstableApi
+public final class BiDnsQueryLifecycleObserver implements DnsQueryLifecycleObserver {
+    private final DnsQueryLifecycleObserver a;
+    private final DnsQueryLifecycleObserver b;
+
+    /**
+     * Create a new instance.
+     * @param a The {@link DnsQueryLifecycleObserver} that will receive events first.
+     * @param b The {@link DnsQueryLifecycleObserver} that will receive events second.
+     */
+    public BiDnsQueryLifecycleObserver(DnsQueryLifecycleObserver a, DnsQueryLifecycleObserver b) {
+        this.a = checkNotNull(a, "a");
+        this.b = checkNotNull(b, "b");
+    }
+
+    @Override
+    public void queryWritten(InetSocketAddress dnsServerAddress, ChannelFuture future) {
+        try {
+            a.queryWritten(dnsServerAddress, future);
+        } finally {
+            b.queryWritten(dnsServerAddress, future);
+        }
+    }
+
+    @Override
+    public void queryCancelled(int queriesRemaining) {
+        try {
+            a.queryCancelled(queriesRemaining);
+        } finally {
+            b.queryCancelled(queriesRemaining);
+        }
+    }
+
+    @Override
+    public DnsQueryLifecycleObserver queryRedirected(List<InetSocketAddress> nameServers) {
+        try {
+            a.queryRedirected(nameServers);
+        } finally {
+            b.queryRedirected(nameServers);
+        }
+        return this;
+    }
+
+    @Override
+    public DnsQueryLifecycleObserver queryCNAMEd(DnsQuestion cnameQuestion) {
+        try {
+            a.queryCNAMEd(cnameQuestion);
+        } finally {
+            b.queryCNAMEd(cnameQuestion);
+        }
+        return this;
+    }
+
+    @Override
+    public DnsQueryLifecycleObserver queryNoAnswer(DnsResponseCode code) {
+        try {
+            a.queryNoAnswer(code);
+        } finally {
+            b.queryNoAnswer(code);
+        }
+        return this;
+    }
+
+    @Override
+    public void queryFailed(Throwable cause) {
+        try {
+            a.queryFailed(cause);
+        } finally {
+            b.queryFailed(cause);
+        }
+    }
+
+    @Override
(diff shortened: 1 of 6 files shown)


Scottmitch:
@normanmaurer - WDYT? we may lose some special case trace notifications ... but I'm not sure if they were even used.

resolver-dns/src/main/java/io/netty/resolver/dns/BiDnsQueryLifecycleObserver.java
```diff
@@ -0,0 +1,96 @@
+package io.netty.resolver.dns;
```
normanmaurer: Missing license header

resolver-dns/src/main/java/io/netty/resolver/dns/BiDnsQueryLifecycleObserver.java
```diff
@@ -0,0 +1,96 @@
...
+import java.net.InetSocketAddress;
+import java.util.List;
+
+import static io.netty.util.internal.ObjectUtil.checkNotNull;
+
+/**
+ * Combines two {@link DnsQueryLifecycleObserver} into a single {@link DnsQueryLifecycleObserver}.
+ */
+@UnstableApi
+public final class BiDnsQueryLifecycleObserver implements DnsQueryLifecycleObserver {
```
normanmaurer: Not sure but why we not just support a a list of observers? A.k.a varargs

resolver-dns/src/main/java/io/netty/resolver/dns/BiDnsQueryLifecycleObserverFactory.java
```diff
@@ -0,0 +1,31 @@
...
+import io.netty.handler.codec.dns.DnsQuestion;
+import io.netty.util.internal.UnstableApi;
+
+import static io.netty.util.internal.ObjectUtil.checkNotNull;
+
+/**
+ * Combines two {@link DnsQueryLifecycleObserverFactory} into a single {@link DnsQueryLifecycleObserverFactory}.
+ */
+@UnstableApi
+public final class BiDnsQueryLifecycleObserverFactory implements DnsQueryLifecycleObserverFactory {
```
normanmaurer: Same as above

resolver-dns/src/main/java/io/netty/resolver/dns/TraceDnsQueryLifeCycleObserverFactory.java
```diff
@@ -0,0 +1,30 @@
+package io.netty.resolver.dns;
```
normanmaurer: Missing license header

normanmaurer:
In General I think it's a good idea 👍

resolver-dns/src/main/java/io/netty/resolver/dns/BiDnsQueryLifecycleObserver.java
Scottmitch: Just an optimization to avoid allocation of collections/arrays. My rational was we can always add a `MultiDnsQueryLifecycleObserver` later ... does this sound reasonable?

resolver-dns/src/main/java/io/netty/resolver/dns/BiDnsQueryLifecycleObserver.java
Scottmitch: fixed

resolver-dns/src/main/java/io/netty/resolver/dns/BiDnsQueryLifecycleObserverFactory.java
Scottmitch: fixed

resolver-dns/src/main/java/io/netty/resolver/dns/TraceDnsQueryLifeCycleObserverFactory.java
Scottmitch: fixed

Scottmitch:
@normanmaurer - comments addressed. PTAL.

resolver-dns/src/main/java/io/netty/resolver/dns/BiDnsQueryLifecycleObserver.java
normanmaurer: I think this is not needed... just make it a List and pass in an unmodifiableList . This way there is no extra allocation and it's more generic. WDYT?

resolver-dns/src/main/java/io/netty/resolver/dns/TraceDnsQueryLifecycleObserver.java
```diff
@@ -0,0 +1,87 @@
...
+    public void queryFailed(Throwable cause) {
+        if (dnsServerAddress != null) {
+            logger.log(level, "from {} : {} failure", dnsServerAddress, question, cause);
+        } else {
+            logger.log(level, "{} query never written and failed", question, cause);
+        }
+    }
+
+    @Override
+    public void querySucceed() {
```
netkins: MAJOR Add a nested comment explaining why this method is empty, throw an UnsupportedOperationException or complete the implementation. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1186)

resolver-dns/src/main/java/io/netty/resolver/dns/BiDnsQueryLifecycleObserver.java
Scottmitch: Creation of the list (or varargs array) requires allocation ... this is what I'm trying to avoid and suggested if there is more than 2 required in the future a `MultiDnsQueryLifecycleObserver` can be used which accepts a list/varargs.

normanmaurer: You only need to create the list one time when create the factory and the pass the same instance to each observer. So it's not really expensive.... I would prefer to not optimize and introduce just another impl for something that can be done with a generic impl which is more or less kind of free

Scottmitch: > You only need to create the list one time when create the factory and the pass the same instance to each observer

The me clarify to make sure we are on the same page...

There is `BiDnsQueryLifecycleObserverFactory` which makes it so 2 `DnsQueryLifecycleObserverFactory` can be used at the same time. `BiDnsQueryLifecycleObserverFactory#newDnsQueryLifecycleObserver` will be called on every DNS query, and this must call both factories to generate two new `DnsQueryLifecycleObserver`s and pass these to a new `BiDnsQueryLifecycleObserver`. We would need to create a list for the factory one time, but then on every query we would need to create a list to give to `BiDnsQueryLifecycleObserver` ... creation of a list on every query is what I want to avoid.

normanmaurer: yeah we are on the same page but I still not get why you would need these extra allocs. What is wrong with create one List when create the Factory, make it unmodifiable and then pass the `same` instance to the newly created Observer all the time ? I don't get why this is not possible or why you think this would cause more allocations then what we have here... Creating a new factory should be very rare and so should be no concern as you usually not create and destroy `DnsResolver`instances frequently.

normanmaurer: @Scottmitch after talking to you offline I realise I am just stupid and you are right... we can't reuse :) So good as it is

Scottmitch:
4.1 (1767814a4)

PR #6888, opened 2017-06-22
Move methods for decode hex dump into StringUtil

fenik17:
Motivation:

PR #6811 introduced a public utility methods to decode hex dump and its parts, but they are not visible from `netty-common`.

Modifications:

1. Move `decodeHexByte`, `decodeHexDump` and `decodeHexNibble` methods into `StringUtil`.
2. Apply these methods where applicable.
3. Remove similar methods from other locations (e.g. `HpackHex` test class).

Result:

Less code duplication.


buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -129,21 +128,21 @@ public static String hexDump(byte[] array, int fromIndex, int length) {
      * Decode a 2-digit hex byte from within a string.
      */
     public static byte decodeHexByte(CharSequence s, int pos) {
```
Scottmitch: should we deprecate these methods and point to the StringUtil counterparts?

normanmaurer: I guess no as StringUtil is internal

normanmaurer:
Cherry-picked into 4.1 (01eb428b39156e61624dae303ae9772ba3424b36) and 4.0 (a341cd14d5b3181bf65e5bad374bc736fc0a1c69)

PR #6889, opened 2017-06-22
Move "fallthrough" statement to where fall actually happens

carl-mastrangelo:
Motivation:
Static analysis looks for error prone switch case statements.
Accidental fall through is one such case, but it is sometimes
intentional.  To indicate this, the "//fallthrough" comment can be
added before the fall.

The code in question has this comment, but it is *after* the fall
so the static analysis flags it.

This is described in http://errorprone.info/bugpattern/FallThrough

Modifications:
Move fall through comment to where the fall actually occurs

Result:
More compatible with Error Prone tools

### DIFF
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
@@ -284,8 +284,8 @@ final class EpollEventLoop extends SingleThreadEventLoop {
                         if (wakenUp == 1) {
                             Native.eventFdWrite(eventFd.intValue(), 1L);
                         }
-                    default:
                         // fallthrough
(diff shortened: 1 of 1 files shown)


normanmaurer:
Cherry-picked into 4.1 (7baef4fbe89559b10fbeed7e17b16f953ff3b7ab) and 4.0 (8834a7e46d1d19680dd0896a0545067236ed9863)

Scottmitch:
thanks @carl-mastrangelo! in the future please remember to also check kqueue for similar updates ... they are very similar and likely have the same issues.

4.1 (1df722f65)

carl-mastrangelo:
@Scottmitch oops, sorry, I didn't think.

Our build system is ultra linux specific, so I didn't even notice the errorprone warning for kqueue, since it won't compile.

Issue #6894, opened 2017-06-23
channelInactive() not getting called when ctx.close() is called on the ChannelHandlerContext


lothbrok72:
### Expected behavior
channelInactive() gets called when ctx.close() is called on the ChannelHandlerContext

### Actual behavior
channelInactive() does not get called when ctx.close() is called on the ChannelHandlerContext.

### Steps to reproduce
We have a test where we write invalid xml to a test channel. ctx.close() is called within exceptionCaught() and I would expect channelInactive to fire. The test is as below. Assert.assertTrue(listener.connectionClosed) is currently failing whereas it was previously passing when using netty-handler 4.0.28.Final. 

        listener.connectionError = null;
        Assert.assertFalse(listener.connectionClosed);
        TestUtils.writeInbound(channel, xml5);
        Assert.assertEquals(null, readNext());
        Assert.assertNotNull(listener.connectionError);
        Assert.assertTrue(listener.connectionClosed);

### Minimal yet complete reproducer code (or URL to code)

    @Override
    public void exceptionCaught(ChannelHandlerContext ctx, Throwable exc) {
        if (connectionListener == null) {
            return;
        }
        // Treat unexpected exceptions as fatal to the connection
        try {
            connectionListener.connectionError(exc);
        } finally {
            ctx.close();
        }
    }

    @Override
    public void channelInactive(ChannelHandlerContext ctx) throws Exception {
        if (connectionListener == null) {
            return;
        }
        connectionListener.connectionClosed();
    }

### Netty version
netty-all 4.1.11.FINAL

normanmaurer:
@lothbrok72 can you come up with a test-case please ?

lothbrok72:
@normanmaurer Our current test case (Invalid XML file, should cause a connection exception and connection close)
- We write a file containing invalid xml to an EmbeddedChannel
- call poll() on the CollectingListener
- check that a connectionError has been thrown
- check that the connection has been closed i.e. channelInactive was called.

I can provide more code of out test setup if required.

Also, seems to work fine when I call ctx.channel().close() instead of ctx.close().

normanmaurer:
So this happens with EmbeddedChannel correct? Like the ChannelHandler is in the pipeline of an EmbeddedChannel?

lothbrok72:
Yes it is an EmbeddedChannel:

    @Before
    public void before() {
        listener = new CollectingListener();
        sync = new SyncStreamRestart();
        Debugger debugger = new ConsoleDebugger();
[...]

PR #6897, opened 2017-06-23
Correctly run all pending tasks for EmbeddedChannel when the Channel …

normanmaurer:
…was closed.

Motivation:

When a user called ctx.close() and used the EmbeddedChannel we did not correctly run all pending tasks which means channelInactive was never called.

Modifications:

Ensure we run all pending tasks after all operations that may change the Channel
 state and are part of the Channel.Unsafe impl.

Result:

Fixes [#6894].

### DIFF
--- a/transport/src/main/java/io/netty/channel/embedded/EmbeddedChannel.java
+++ b/transport/src/main/java/io/netty/channel/embedded/EmbeddedChannel.java
@@ -35,6 +35,7 @@ import io.netty.channel.ChannelPromise;
 import io.netty.channel.DefaultChannelConfig;
 import io.netty.channel.DefaultChannelPipeline;
 import io.netty.channel.EventLoop;
+import io.netty.channel.RecvByteBufAllocator;
 import io.netty.util.ReferenceCountUtil;
 import io.netty.util.internal.ObjectUtil;
 import io.netty.util.internal.PlatformDependent;
@@ -701,7 +702,12 @@ public class EmbeddedChannel extends AbstractChannel {
 
     @Override
     protected AbstractUnsafe newUnsafe() {
-        return new DefaultUnsafe();
+        return new EmbeddedUnsafe();
+    }
+
+    @Override
+    public Unsafe unsafe() {
+        return ((EmbeddedUnsafe) super.unsafe()).wrapped;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


normanmaurer:
@trustin can you check https://github.com/netty/netty/pull/6897 ?

transport/src/main/java/io/netty/channel/embedded/EmbeddedChannel.java
```diff
@@ -697,6 +697,13 @@ protected void handleInboundMessage(Object msg) {
         public void connect(SocketAddress remoteAddress, SocketAddress localAddress, ChannelPromise promise) {
             safeSetSuccess(promise);
         }
+
+        @Override
+        protected void afterClose() {
+            // Ensure we rull al pending tasks when the channel is closed.
```
trustin: run all

transport/src/test/java/io/netty/channel/embedded/EmbeddedChannelTest.java
```diff
@@ -541,6 +542,25 @@ protected void handleOutboundMessage(Object msg) {
         }
     }
 
+    @Test(timeout = 2000)
```
trustin: Does this have to be this short? Seems like it wants flakiness on a slow machine.

normanmaurer:
@trustin thanks.. .fixed your nits

transport/src/main/java/io/netty/channel/AbstractChannel.java
```diff
@@ -678,6 +682,7 @@ public void run() {
                     fireChannelInactiveAndDeregister(wasActive);
                 }
             }
+            afterClose();
```
Scottmitch: can you add a comment as to why this cannot be accomplished in `doClose()`? It seems strange that this operation is done synchronously and the `doClose()` call may use `closeExecutor`.

normanmaurer: yep will do...

normanmaurer: let me think a bit more about this bugfix

normanmaurer:
@Scottmitch PTAL again... I think this is better

normanmaurer:
Cherry-picked into 4.1 (8adb30bbe2785094c92d085757cb1fdc2c1edec9) and 4.0 (98140cb98f).

Thanks again @Scottmitch for the comments and so let me think more about this and come up with a better solution

PR #6900, opened 2017-06-24
Exhaust non-exhaustive switch cases

carl-mastrangelo:
Motivation:
ErrorProne warns about missing cases in switch statements that
appear as an oversight.

Modifcation:
Add the last case to statement to ensure all cases are covered.

Result:
Able to enable Error Prone static analysis

Note:  Sorry about sending so many tiny PRs.  Each compilation failure gets fixed and rolled into a commit.  Trying to compile a single package at a time takes a lot of work, so some fixes are not obvious until later.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/JdkSslContext.java
+++ b/handler/src/main/java/io/netty/handler/ssl/JdkSslContext.java
@@ -244,6 +244,10 @@ public class JdkSslContext extends SslContext {
                 case REQUIRE:
                     engine.setNeedClientAuth(true);
                     break;
+                case NONE:
+                    break; // exhaustive cases
+                default:
+                    throw new Error("Unknown auth " + clientAuth);
             }
         }
         return apn.wrapperFactory().wrapSslEngine(engine, apn, isServer());


normanmaurer:
@carl-mastrangelo no worries... thanks for all the patches!

handler/src/main/java/io/netty/handler/ssl/JdkSslContext.java
```diff
@@ -244,6 +244,8 @@ private SSLEngine configureAndWrapEngine(SSLEngine engine) {
                 case REQUIRE:
                     engine.setNeedClientAuth(true);
                     break;
+                case NONE:
+                    break; // exhaustive cases
```
normanmaurer: @carl-mastrangelo should we also add:

```java
default: 
    throw Error()
```

This way we will ensure we never miss one ;)

carl-mastrangelo: Done.

normanmaurer:
Cherry-picked into 4.1 (322fe8ec85d72271a5ac0ed3f0149c4ea1323168) and 4.0 (344360af308bd952642fb596d8c55900984282ce)

PR #6904, opened 2017-06-26
Add content-security-policy and x-frame-options to HttpHeaderNames

normanmaurer:
Motivation:

These headers can be used to prevent clickjacking.

Modifications:

Add static fields for content-security-policy and x-frame-options

Result:

Expose general useful names


normanmaurer:
Cherry-picked into 4.1 (cc1a209185)

trustin:
Hmm.. why are these fields `CharSequence` while others are `AsciiString`?

normanmaurer:
@trustin because I am stupid :( Good catch let me do a followup

normanmaurer:
Fixed via 3b5dd4e9dc0b6e350d27a2628b3f37b2483fe4a9.... Thanks @trustin for the catch!

PR #6907, opened 2017-06-27
Close channels that are released to a closed FixedChannelPool.

normanmaurer:
Motivation:

Channels returned to a FixedChannelPool after closing it remain active.

Since channels that where acquired from the pool are not closed during the close operation, they remain open even after releasing the channel back to the pool where they are then in accessible and become in-effect a connection leak.

Modification:

Close the released channel on releasing back to a closed pool.

Result:

Much harder to create a connection leak by closing an active
FixedChannelPool instance.


normanmaurer:
Cherry-picked into 4.1 (32b3f58f6308f0d7cc22fad4a94406c4fe16a402) and 4.0 (b6af2bac8b6f12365236370dfe124e2f422d2f7c)

PR #6911, opened 2017-06-27
KQueue.isAvailable() should return false on MacOS < 10.12

normanmaurer:
Motivation:

We currently dont support MacOS < 10.12 so we should return faluse from KQueue.isAvailable() if we are running on an older version.

Modifications:

Check for MacOS version and return false if too old.

Result:

KQueue.isAvailable() will return false if used on an unsupported OSX version.

### DIFF
--- a/transport-native-kqueue/src/main/java/io/netty/channel/kqueue/KQueue.java
+++ b/transport-native-kqueue/src/main/java/io/netty/channel/kqueue/KQueue.java
@@ -17,7 +17,10 @@ package io.netty.channel.kqueue;
 
 import io.netty.channel.unix.FileDescriptor;
 import io.netty.util.internal.PlatformDependent;
+import io.netty.util.internal.SystemPropertyUtil;
 import io.netty.util.internal.UnstableApi;
+
+import java.util.Locale;
 
 /**
  * If KQueue is available the JNI resources will be loaded when this class loads.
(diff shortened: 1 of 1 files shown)


Scottmitch:
`transport-native-kqueue` is compatible with 10.11, it just doesn't support linking with 10.11 when built on 10.12 (which is what we release with). This would prevent folks folk using `transport-native-kqueue` if they build their own.

normanmaurer:
@Scottmitch let me come up with a better way.

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/KQueue.java
```diff
@@ -46,8 +49,29 @@
         if (cause != null) {
             UNAVAILABILITY_CAUSE = cause;
         } else {
-            UNAVAILABILITY_CAUSE = PlatformDependent.hasUnsafe() ? null :
+            cause = PlatformDependent.hasUnsafe() ? null :
                     new IllegalStateException("sun.misc.Unsafe not available");
+            if (cause == null) {
+                if (SystemPropertyUtil.get("os.name", "").toLowerCase(Locale.US).contains("mac")) {
```
netkins: MAJOR Merge this if statement with the enclosing one. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1066)

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/KQueue.java
```diff
@@ -46,8 +49,29 @@
...
+                    String version = SystemPropertyUtil.get("os.version", "");
+                    String[] versionParts = SystemPropertyUtil.get("os.version", "")
+                            .toLowerCase(Locale.US).split("\\.");
+                    if (versionParts.length < 2) {
+                        cause = new IllegalStateException("MacOS version " + version + " not supported");
+                    } else {
+                        try {
+                            int majorVersion = Integer.parseInt(versionParts[0]);
+                            int minorVersion = Integer.parseInt(versionParts[1]);
+                            if (majorVersion < 10 || (majorVersion == 10 && minorVersion < 12)) {
```
netkins: MAJOR Refactor this code to not nest more than 5 if/for/while/switch/try statements. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS134)

Scottmitch:
I propose closing this PR in favor of https://github.com/netty/netty/pull/6915

normanmaurer:
Thanks @Scottmitch Lets do it...

PR #6912, opened 2017-06-27
Avoid a double check ByteBuf#ensureWritable in ByteBufUtil

fenik17:
Motivation:

Methods `ByteBufUtil#writeUtf8` and `ByteBufUtil#writeAscii` contains a check `ByteBuf#ensureWritable` before the calling `ByteBuf#writeBytes`. But the `ByteBuf#writeBytes` also do a such check inside.

Modifications:

Make checks more targeted.

Result:

Less redundant method calls.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
+++ b/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
@@ -444,13 +444,11 @@ public final class ByteBufUtil {
      * This method returns the actual number of bytes written.
      */
     public static int writeUtf8(ByteBuf buf, CharSequence seq) {
-        final int len = seq.length();
-        buf.ensureWritable(utf8MaxBytes(seq));
-
         for (;;) {
             if (buf instanceof AbstractByteBuf) {
                 AbstractByteBuf byteBuf = (AbstractByteBuf) buf;
-                int written = writeUtf8(byteBuf, byteBuf.writerIndex, seq, len);
+                byteBuf.ensureWritable(utf8MaxBytes(seq));
+                int written = writeUtf8(byteBuf, byteBuf.writerIndex, seq, seq.length());
                 byteBuf.writerIndex += written;
                 return written;
             } else if (buf instanceof WrappedByteBuf) {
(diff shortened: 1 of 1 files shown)


buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -444,13 +444,11 @@ public static ByteBuf writeUtf8(ByteBufAllocator alloc, CharSequence seq) {
...
      */
     public static int writeUtf8(ByteBuf buf, CharSequence seq) {
-        final int len = seq.length();
-        buf.ensureWritable(utf8MaxBytes(seq));
-
         for (;;) {
             if (buf instanceof AbstractByteBuf) {
                 AbstractByteBuf byteBuf = (AbstractByteBuf) buf;
-                int written = writeUtf8(byteBuf, byteBuf.writerIndex, seq, len);
+                byteBuf.ensureWritable(utf8MaxBytes(seq));
```
normanmaurer: Doesn't this mean w will check multiple times now when unwrap the buffers ?

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackEncoder.java
```diff
@@ -257,15 +257,9 @@ private void encodeStringLiteral(ByteBuf out, CharSequence string) {
...
-                AsciiString asciiString = (AsciiString) string;
-                out.writeBytes(asciiString.array(), asciiString.arrayOffset(), asciiString.length());
-            } else {
-                // Only ASCII is allowed in http2 headers, so its fine to use this.
-                // https://tools.ietf.org/html/rfc7540#section-8.1.2
-                out.writeCharSequence(string, CharsetUtil.ISO_8859_1);
-            }
+            // Only ASCII is allowed in http2 headers, so its fine to use this.
+            // https://tools.ietf.org/html/rfc7540#section-8.1.2
+            ByteBufUtil.writeAscii(out, string);
```
normanmaurer: Why not just leave it as it is ? To remove code-duplication ? Also This not belongs to this PR imho

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
fenik17: No. This code branch is finishes with `return`.

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackEncoder.java
fenik17: Yes, it's unrelated. Sorry.. reverted.

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
normanmaurer: @fenik17 ah true... thanks

normanmaurer:
Cherry-picked into 4.1 (f35047765f00de1d642d23bc0f980fa8eceacf3b)

PR #6913, opened 2017-06-27
SocksCmdRequest and SocksCmdResponse are trying to convert host from …

fenik17:
…IDN for the non-DOMAIN address types

Motivation:

In the `SocksCmdRequest` and `SocksCmdResponse` constructors a host param converts from IDN to ascii compatible form regardless address type.

Modifications:

Use `IDN#toASCII` only for `DOMAIN` address type.

Result:

More correct host handling in socks commands.

### DIFF
--- a/codec-socks/src/main/java/io/netty/handler/codec/socks/SocksCmdRequest.java
+++ b/codec-socks/src/main/java/io/netty/handler/codec/socks/SocksCmdRequest.java
@@ -51,9 +51,11 @@ public final class SocksCmdRequest extends SocksRequest {
                 }
                 break;
             case DOMAIN:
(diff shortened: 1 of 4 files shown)


normanmaurer:
@fenik17 can you add a test-case as well ?

fenik17:
@normanmaurer I added tests that check that the host is encoded as ascii for `DOMAIN` address type and don't encode the host for `UNKNOWN`.

fenik17:
Also fixed getters `host()`.

normanmaurer:
@Scottmitch want to check as well ?

normanmaurer:
Cherry-picked into 4.1 (8d0e0922a512a08a287c25bed6462fa244738603)

PR #6915, opened 2017-06-28
transport-native-kqueue libraries should not lazy link

Scottmitch:
Motivation:
We rely upon the linker being non-lazy to test compatibility the native library compatibility for kqueue, but the default mode of operation is to lazy link.

Modifications:
- We should modify the build scripts to inform the linker that this library should not be lazy linked
- Error messages changes
dyld: lazy symbol binding failed: Symbol not found: _clock_gettime

java.lang.UnsatisfiedLinkError: unsupported JNI version 0xFFFFFFFF required by .../libnetty-transport-native-kqueue.dylib

Result:
Link errors are detected upon library load time.

### DIFF
--- a/transport-native-kqueue/pom.xml
+++ b/transport-native-kqueue/pom.xml
@@ -79,6 +79,8 @@
                   <configureArgs>
                     <arg>${jni.compiler.args.ldflags}</arg>
                     <arg>${jni.compiler.args.cflags}</arg>
+                    <!-- weak_import was added in 10.2 so ensure we explicitly set the artifact -->
+                    <arg>MACOSX_DEPLOYMENT_TARGET=10.2</arg>
                   </configureArgs>
                 </configuration>
                 <goals>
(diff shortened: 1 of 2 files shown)


Scottmitch:
@normanmaurer - I tested this with an artifact built on 10.12.5 and ran on 10.10.5. Both platforms can also build their own artifacts and link/run with those artifacts. If you have a 10.11 host to try that would also be nice.

transport-native-kqueue/pom.xml
```diff
@@ -79,6 +79,8 @@
                   <configureArgs>
                     <arg>${jni.compiler.args.ldflags}</arg>
                     <arg>${jni.compiler.args.cflags}</arg>
+                    <!-- weak_import was added in 10.2 so ensure we explicitly set the artifact -->
+                    <arg>MACOSX_DEPLOYMENT_TARGET=10.2</arg>
```
normanmaurer: @Scottmitch in the comment you say `weak_import` but we not use this in the ldflags. Are you refer to `-z now` ? Can you clarify this ?

transport-native-unix-common/pom.xml
```diff
@@ -102,6 +102,8 @@
                       <env key="CFLAGS" value="-O3 -Werror -Wno-attributes -fPIC -fno-omit-frame-pointer -Wunused-variable" />
                       <env key="LDFLAGS" value="-Wl,--no-as-needed -lrt" />
                       <env key="LIB_NAME" value="${nativeLibName}" />
+                      <!-- weak_import was added in 10.2 so ensure we explicitly set the artifact -->
+                      <env key="MACOSX_DEPLOYMENT_TARGET" value="10.2" />
```
normanmaurer: see above

transport-native-kqueue/pom.xml
Scottmitch: Sure. The artifact failed to link on my 10.10 machine because the weak linking feature was not enabled. This meant that accept4 (which is defined as `__attribute__((weak_import))`) failed to link (which is expected) but the linker threw an error.

normanmaurer: Ah got it... can you just add a bit more context in the comment then 👍

Scottmitch: yip done

normanmaurer:
@Scottmitch ship it... also should we do the same for tcnative ?

Scottmitch:
4.1 (07a641900)

Issue #6906, opened 2017-06-27
Exception occours if client sends reset frame while the server is sending headers


qwangseu:
when I use netty http2 to build a server, exception will occour if the client sents a reset frame to the server while the server is sending headers (netty version is 4.1.8.final),
```
io.netty.handler.codec.http2.Http2Exception: Request stream 2414839 is not correct for server connection
   at io.netty.handler.codec.http2.Http2Exception.connectionError(Http2Exception.java:85)
   at io.netty.handler.codec.http2.DefaultHttp2Connection$DefaultEndpoint.checkNewStreamAllowed(DefaultHttp2Connection.java:1095)
   at io.netty.handler.codec.http2.DefaultHttp2Connection$DefaultEndpoint.createStream(DefaultHttp2Connection.java:942)
   at io.netty.handler.codec.http2.DefaultHttp2Connection$DefaultEndpoint.createStream(DefaultHttp2Connection.java:960)
   at io.netty.handler.codec.http2.DefaultHttp2Connection$DefaultEndpoint.createStream(DefaultHttp2Connection.java:868)
   at io.netty.handler.codec.http2.DefaultHttp2ConnectionEncoder.writeHeaders(DefaultHttp2ConnectionEncoder.java:158)
   at io.netty.handler.codec.http2.DefaultHttp2ConnectionEncoder.writeHeaders(DefaultHttp2ConnectionEncoder.java:148)
   at com.c.t.h2c.netty.NettyServerHandler.sendResponseHeaders(NettyServerHandler.java:293)
   at com.c.t.h2c.netty.NettyServerHandler.write(NettyServerHandler.java:253)
   at io.netty.channel.AbstractChannelHandlerContext.invokeWrite0(AbstractChannelHandlerContext.java:739)
   at io.netty.channel.AbstractChannelHandlerContext.invokeWrite(AbstractChannelHutor$5.run(SingleThreadEventExecutor.java:858)
   at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:144)
   at java.lang.Thread.run(Thread.java:745)
```

https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java#L156
when the server receives reset frame  it will remove the stream from the streamMap, but  it will create the stream when it writes headers if it finds the stream is null。

[...]

PR #6916, opened 2017-06-28
DefaultHttp2ConnectionEncoder#writeHeaders shouldn't send GO_AWAY if …

Scottmitch:
…stream is closed

Motivation:
DefaultHttp2ConnectionEncoder#writeHeaders attempts to find a stream object, and if one doesn't exist it tries to create one. However in the event that the local endpoint has received a RST_STREAM frame before writing the response headers we attempt to create a stream. Since this stream ID is for the incorrect endpoint we then generate a GO_AWAY for what appears to be a protocol error, but can instead be failed locally.

Modifications:
- Just fail the local promise in the above situation instead of sending a GO_AWAY

Result:
Less severe consequences if the server asynchronously sends headers after a RST_STREAM has been received.
Fixes https://github.com/netty/netty/issues/6906.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
@@ -126,8 +126,7 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
                     // Allowed sending DATA frames in these states.
                     break;
                 default:
-                    throw new IllegalStateException(String.format(
-                            "Stream %d in unexpected state: %s", stream.id(), stream.state()));
+                    throw new IllegalStateException("Stream " + stream.id() + " in unexpected state " + stream.state());
             }
         } catch (Throwable e) {
             data.release();
@@ -153,7 +152,15 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
         try {
             Http2Stream stream = connection.stream(streamId);
             if (stream == null) {
-                stream = connection.local().createStream(streamId, endOfStream);
+                try {
+                    stream = connection.local().createStream(streamId, endOfStream);
+                } catch (Http2Exception cause) {
+                    if (connection.remote().mayHaveCreatedStream(streamId)) {
+                        promise.tryFailure(new IllegalStateException("Stream no longer exists: " + streamId, cause));
+                        return promise;
+                    }
+                    throw cause;
+                }
             } else {
                 switch (stream.state()) {
                     case RESERVED_LOCAL:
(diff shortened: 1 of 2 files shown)


Scottmitch:
@ejona86 @carl-mastrangelo - FYI

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -153,7 +152,15 @@ public ChannelFuture writeHeaders(final ChannelHandlerContext ctx, final int str
         try {
             Http2Stream stream = connection.stream(streamId);
             if (stream == null) {
-                stream = connection.local().createStream(streamId, endOfStream);
+                try {
+                    stream = connection.local().createStream(streamId, endOfStream);
+                } catch (Throwable cause) {
+                    if (connection.remote().mayHaveCreatedStream(streamId)) {
+                        promise.tryFailure(new IllegalStateException("Stream no longer exists: " + streamId, cause));
```
Scottmitch: this error message is consistent with one below ... I can change the format to move the streamID if folks prefer.

Zeymo: connection.remote().mayHaveCreatedStream(streamId) will also be false and throw cause...I think this issue is no bussniess about goway... what we expect is just return and done the stream rather than  throw any exception

BTW client send RST_STREAM but not send GO_AWAY

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -153,7 +152,15 @@ public ChannelFuture writeHeaders(final ChannelHandlerContext ctx, final int str
         try {
             Http2Stream stream = connection.stream(streamId);
             if (stream == null) {
-                stream = connection.local().createStream(streamId, endOfStream);
+                try {
```
netkins: MAJOR Extract this nested try block into a separate method. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1141)

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -153,7 +152,15 @@ public ChannelFuture writeHeaders(final ChannelHandlerContext ctx, final int str
...
+                    stream = connection.local().createStream(streamId, endOfStream);
+                } catch (Throwable cause) {
```
normanmaurer: Can we only catch `Http2Exception` here and let everything else bubble up always ?

Scottmitch: Currently it looks like we should be able todo this ... however if the implementation changes and any other exception type escapes (e.g. `IllegalState`) we will send a GO_AWAY. Maybe this is OK though?

normanmaurer: Yeah I think that's fine... so +1 to change

Scottmitch: done

Scottmitch:
4.1 (bc46a99ea)

PR #6918, opened 2017-06-28
Performance improvement: Character.isWhitespace replaced with naive ==.

doom369:
Motivation:

Right now netty uses `Character.isWhitespace` check for whitespaces in some places of `HttpObjectDecoder` and few others http classes. However, this method is not intrinsic and slower than naive comparing with equal operator `==` (as it uses table lookup).

Modification:

`Character.isWhitespace` repalced with `==`. `!Character.isWhitespace` replaced with `!=`.

Result:

Tests show that `splitHeader` method of `HttpRequestDecoder` is 30% faster.

Benchmark for `Character.isWhitespace` vs `==` :

```
@Fork(1)
@State(Scope.Thread)
@Warmup(iterations = 10, time = 1, batchSize = 1000)
@Measurement(iterations = 20, time = 1, batchSize = 1000)
public class CharacterCompare {

    private char c = 'c';

    @Benchmark
    public boolean characterIsWhitespaceChar() {
        return Character.isWhitespace(c);
    }

    @Benchmark
    public boolean naiveCheckChar() {
        return c == ' ';
    }
}
```

Result: 
```
Benchmark                                    Mode  Cnt       Score       Error  Units
CharacterCompare.characterIsWhitespaceChar  thrpt   20  287834.304 ± 16050.006  ops/s
CharacterCompare.naiveCheckChar             thrpt   20  355577.043 ± 39929.517  ops/s
```

Benchmark for `splitHeader` of `HttpRequestDecoder` (you need to make fields and method public to make it work) :

```
@State(Scope.Benchmark)
@Warmup(iterations = 10)
@Measurement(iterations = 20)
public class HttpObjectDecoderSplitHeadersBenchmark extends AbstractMicrobenchmark {

    private HttpRequestDecoder httpRequestDecoder = new HttpRequestDecoder();
    private AppendableCharSequence header = new AppendableCharSequence(23).append("Content-Type: text/html");

    @Benchmark
    public CharSequence oldDecoder() throws Exception {
        httpRequestDecoder.splitHeader(header);
        return httpRequestDecoder.name;
    }

    @Benchmark
    public CharSequence newDecoder() throws Exception {
        httpRequestDecoder.splitHeaderNew(header);
        return httpRequestDecoder.name;
    }

}
```

Result:

```
Benchmark                                           Mode  Cnt         Score         Error  Units
HttpObjectDecoderSplitHeadersBenchmark.newDecoder  thrpt   40  21289487.337 ±  986998.323  ops/s
HttpObjectDecoderSplitHeadersBenchmark.oldDecoder  thrpt   40  14586426.110 ± 1328832.579  ops/s
```

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpMethod.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpMethod.java
@@ -127,7 +127,7 @@ public class HttpMethod implements Comparable<HttpMethod> {
 
         for (int i = 0; i < name.length(); i ++) {
             char c = name.charAt(i);
-            if (Character.isISOControl(c) || Character.isWhitespace(c)) {
+            if (Character.isISOControl(c) || c == ' ') {
                 throw new IllegalArgumentException("invalid character in name");
             }
         }
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectDecoder.java
@@ -681,7 +681,7 @@ public abstract class HttpObjectDecoder extends ByteToMessageDecoder {
         hex = hex.trim();
         for (int i = 0; i < hex.length(); i ++) {
             char c = hex.charAt(i);
-            if (c == ';' || Character.isWhitespace(c) || Character.isISOControl(c)) {
+            if (c == ';' || c == ' ' || Character.isISOControl(c)) {
                 hex = hex.substring(0, i);
                 break;
             }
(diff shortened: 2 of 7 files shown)


fenik17:
This is change in behaviour.

https://tools.ietf.org/html/rfc7230#section-3.5
```
   Although the request-line and status-line grammar rules require that
   each of the component elements be separated by a single SP octet,
   recipients MAY instead parse on whitespace-delimited word boundaries
   and, aside from the CRLF terminator, treat any form of whitespace as
   the SP separator while ignoring preceding or trailing whitespace;
   such whitespace includes one or more of the following octets: SP,
   HTAB, VT (%x0B), FF (%x0C), or bare CR.  
```

doom369:
@fenik17 you are right. Somehow I missed javadocs explanation.

PR #6928, opened 2017-06-30
Reflectively load Optional SSL Dependencies

carl-mastrangelo:
Motivation:
Static analyzers complain if there are references to absent classes
in a jar file.  It is possible to remove said files, but for the
SSL handlers in netty doing so only raises more warnings.

Modification:
Jdk(Alpn|Npn) Ssl classes are modified to be more defensive about
loading non present SSL Engines via reflection.

Result:
Less complainy static analysis

This change is a slightly more burdensome change, since compile
time analysis is intentionally broken.  I understand if you might
not want this PR, though since it is useful to my team, I thought
I would pose it.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/JdkAlpnApplicationProtocolNegotiator.java
+++ b/handler/src/main/java/io/netty/handler/ssl/JdkAlpnApplicationProtocolNegotiator.java
@@ -15,14 +15,19 @@
  */
 package io.netty.handler.ssl;
 
+import io.netty.util.internal.logging.InternalLogger;
+import io.netty.util.internal.logging.InternalLoggerFactory;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import javax.net.ssl.SSLEngine;
+
 
 /**
  * The {@link JdkApplicationProtocolNegotiator} to use if you need ALPN and are using {@link SslProvider#JDK}.
  */
 public final class JdkAlpnApplicationProtocolNegotiator extends JdkBaseApplicationProtocolNegotiator {
-    private static final boolean AVAILABLE = ConscryptAlpnSslEngine.isAvailable() || JettyAlpnSslEngine.isAvailable();
-    private static final SslEngineWrapperFactory ALPN_WRAPPER = AVAILABLE ? new AlpnWrapper() : new FailureWrapper();
+    private static final SslEngineWrapperFactory ALPN_WRAPPER = new AlpnWrapper();
 
     /**
      * Create a new instance.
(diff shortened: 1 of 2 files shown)


handler/src/main/java/io/netty/handler/ssl/JdkAlpnApplicationProtocolNegotiator.java
```diff
@@ -15,14 +15,19 @@
...
 import javax.net.ssl.SSLEngine;
 
+
 /**
  * The {@link JdkApplicationProtocolNegotiator} to use if you need ALPN and are using {@link SslProvider#JDK}.
  */
 public final class JdkAlpnApplicationProtocolNegotiator extends JdkBaseApplicationProtocolNegotiator {
-    private static final boolean AVAILABLE = ConscryptAlpnSslEngine.isAvailable() || JettyAlpnSslEngine.isAvailable();
-    private static final SslEngineWrapperFactory ALPN_WRAPPER = AVAILABLE ? new AlpnWrapper() : new FailureWrapper();
+    private static final SslEngineWrapperFactory ALPN_WRAPPER = new AlpnWrapper();
```
netkins: MAJOR Avoid instantiation through private constructors from outside of the constructors class. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AAccessorClassGeneration)

normanmaurer:
@trustin @Scottmitch WDYT ?

carl-mastrangelo:
One alternative is to move the reflection directly into the Jetty* and Conscrypt* classes.  They already do reflection, but the amount of change would be much larger that this version.

normanmaurer:
@carl-mastrangelo I kind of like the idea of having all reflection in in one place tho

carl-mastrangelo:
@normanmaurer I investigated if the reflection could be put into the leaf classes.  For Jetty, it would be possible, but would be significantly uglier.  The code would need to resort to using dynamic Proxies in order to create `ALPN.ServerProvider` and `ALPN.ClientProvider` instances.  

For conscrypt, it won't work because it depends on an abstract class `HandshakeListener` which cannot be instantiated reflectively.  From looking around, using javassist could be used to create one, but that may be a bridge too far.

Scottmitch:
We already depend upon properties of the class loaders to delay class loading until first reference for features which are available in newer JDK versions and for our other native libraries. I'm not sure this PR provides much value at the expense of making the class references implicit through reflection. If we get it wrong and load classes which are not available that would be a programing error, and we will fix it, but IIUC for now we are OK.

normanmaurer:
@carl-mastrangelo I agree with @Scottmitch ... WDYT ?

carl-mastrangelo:
I can roll this back.

PR #6939, opened 2017-07-04
Clarify the appointment of the intermediate collection

fenik17:
Motivation:
An intermediate list is creating in the `EpollEventLoop#closeAll` to prevent ConcurrentModificationException. But this is not the obvious purpose has no comment.

Modifications:
Add comment to clarify the appointment of the intermediate collection.

Result:
More clear code.

### DIFF
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
@@ -350,6 +350,8 @@ final class EpollEventLoop extends SingleThreadEventLoop {
         } catch (IOException ignore) {
             // ignore on close
         }
+        // Using the intermediate collection to prevent ConcurrentModificationException.
+        // In the `close()` method, the channel is deleted from `channels` map.
         Collection<AbstractEpollChannel> array = new ArrayList<AbstractEpollChannel>(channels.size());
 
         for (AbstractEpollChannel channel: channels.values()) {


transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
```diff
@@ -350,14 +350,9 @@ private void closeAll() {
...
         }
-        Collection<AbstractEpollChannel> array = new ArrayList<AbstractEpollChannel>(channels.size());
-
-        for (AbstractEpollChannel channel: channels.values()) {
-            array.add(channel);
-        }
-
-        for (AbstractEpollChannel ch: array) {
-            ch.unsafe().close(ch.unsafe().voidPromise());
+        for (PrimitiveEntry<AbstractEpollChannel> entry : channels.entries()) {
```
normanmaurer: @Scottmitch @trustin do you remember why we did this ? Was it to ensure we not remove entries from the map while doing the iteration ?

fenik17: Perhaps this is a rudiment, preserved from the initial commit:

https://github.com/fenik17/netty/blob/9330172f803f340df677c370464126cd6112204a/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java#L292

also a `channels` map is not thread-safe.

Scottmitch: I would assume it is to avoid ConcurrentModificationException. close() -> doClose() -> doDeregister() -> EpollEventLoop#remove() -> channels#remove

normanmaurer: Yeah that's what I assume as well. Couldn't that happen @fenik17?

fenik17: You are right, guys. Thanks for not letting me do something stupid!
I updated a PR.

Scottmitch:
4.1 (016e00bf7) 4.0 (df09b84e4)

Issue #6934, opened 2017-07-02
java.lang.AbstractMethodError: java.nio.channels.SocketChannel.bind


Donaldhan:
my os is win10, jdk is 1.7.0.17,netty is 4.1.12. I try run the  MsgEchoServer/Client example in udt package, 
then  MsgEchoServer is started, when start the MsgEchoClient,get this errror:

[INFO ] 2017-07-02 21:57:25 io.netty.handler.logging.LoggingHandler [id: 0xdf676102] REGISTERED
[INFO ] 2017-07-02 21:57:25 io.netty.handler.logging.LoggingHandler [id: 0xdf676102] CONNECT: /192.168.32.129:10020
Exception in thread "main" java.lang.AbstractMethodError: java.nio.channels.SocketChannel.bind(Ljava/net/SocketAddress;)Ljava/nio/channels/SocketChannel;
	at io.netty.util.internal.SocketUtils$4.run(SocketUtils.java:96)
	at io.netty.util.internal.SocketUtils$4.run(SocketUtils.java:93)
	at java.security.AccessController.doPrivileged(Native Method)
	at io.netty.util.internal.SocketUtils.bind(SocketUtils.java:93)
	at io.netty.channel.udt.nio.NioUdtMessageConnectorChannel.doBind(NioUdtMessageConnectorChannel.java:101)
	at io.netty.channel.udt.nio.NioUdtMessageConnectorChannel.doConnect(NioUdtMessageConnectorChannel.java:112)
	at io.netty.channel.nio.AbstractNioChannel$AbstractNioUnsafe.connect(AbstractNioChannel.java:254)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.connect(DefaultChannelPipeline.java:1266)
	at io.netty.channel.AbstractChannelHandlerContext.invokeConnect(AbstractChannelHandlerContext.java:545)
	at io.netty.channel.AbstractChannelHandlerContext.connect(AbstractChannelHandlerContext.java:530)
	at io.netty.handler.logging.LoggingHandler.connect(LoggingHandler.java:201)
	at io.netty.channel.AbstractChannelHandlerContext.invokeConnect(AbstractChannelHandlerContext.java:545)
	at io.netty.channel.AbstractChannelHandlerContext.connect(AbstractChannelHandlerContext.java:530)
	at io.netty.channel.AbstractChannelHandlerContext.connect(AbstractChannelHandlerContext.java:512)
[...]

PR #6940, opened 2017-07-04
AbstractMethodError with barchart-udt (builded for java 6)

fenik17:
Motivation:

`SocketChannelUDT` from barchart-udt does not have the java 7 `public abstract SocketChannel bind(SocketAddress local)` method. Calling the abstract method `SocketChannel.bind(SocketAddress localAddress)` for `SocketChannelUDT` leads to an `AbstractMethodError` runtime error.

Modifications:

Make workaround with explicit call of `SocketChannelUDT.bind(SocketAddress local)` as it done in `NioUdtByteConnectorChannel`.

Result:

Fixes [#6934].


normanmaurer:
Cherry-picked into 4.1 (1ba265ad4d14163e346996a5c0a663b4600e7d29) and 4.0 (0a6df7b764821ed5b84c753489953287527455b1)

PR #6943, opened 2017-07-05
Also use realloc when shrink the buffer.

normanmaurer:
Motivation:

We should also use realloc when shrink the buffer to eliminate extra allocations / memory copies when possible.

Modifications:

Use realloc for expanding and shrinking when possible.

Result:

Less memory copies and allocations


normanmaurer:
Cherry-picked into 4.1 (83db2b07b4) and 4.0 (20059f130dbfbf3f7e2ef6877e82f1fea6fe8725)

PR #6944, opened 2017-07-05
UnixResolverDnsServerAddressStreamProviderTest failure

Scottmitch:
Motivation:
InetSocketAddress#getHostName() may attempt a reverse lookup which may lead to test failures because the expected address will not match.

Modifications:
- Use InetSocketAddress#getHostString() which will not attempt any lookups and instead return the original String

Result:
UnixResolverDnsServerAddressStreamProviderTest is more reliable.


Scottmitch:
Build failure unrelated and captured in https://github.com/netty/netty/issues/6946

Scottmitch:
4.1 (d040c939e)

PR #6952, opened 2017-07-07
 Configures HTTP2 pipeline with more proper way

### LINKED ISSUE #6881: HTTP/2 Prior Knowledge does not appear to work with Http2Codec and Http2ServerDowngrader
### Expected behavior

The `CleartextHttp2ServerUpgradeHandler` should work with both HTTP Upgrade and Prior Knowledge when downgrading HTTP/2 requests internally.

### Actual behavior

HTTP Upgrade works, but not Prior Knowledge. In the latter case, nghttp reports: `(last_stream_id=0, error_code=PROTOCOL_ERROR(0x01), opaque_data(96)=[HTTP/2 client preface string missing or corrupt. Hex dump for received bytes: 000000040100000000])`

### Steps to reproduce

1. Fire up `io.netty.example.http2.helloworld.server.Http2Server` with the modifications described in the next section
2. Install `nghttp`
3. Run `nghttp --upgrade --verbose --no-dep http://localhost:8080/`; it should succeed
4. Run `nghttp --verbose --no-dep http://localhost:8080/`; it fails like so:

```
[  0.002] Connected
[  0.002] send SETTINGS frame <length=12, flags=0x00, stream_id=0>
          (niv=2)
          [SETTINGS_MAX_CONCURRENT_STREAMS(0x03):100]
          [SETTINGS_INITIAL_WINDOW_SIZE(0x04):65535]
[  0.002] send HEADERS frame <length=33, flags=0x05, stream_id=1>
          ; END_STREAM | END_HEADERS
          (padlen=0)
          ; Open new stream
          :method: GET
          :path: /
          :scheme: http
          :authority: localhost:8080
          accept: */*
          accept-encoding: gzip, deflate
          user-agent: nghttp2/1.23.1
[  0.005] recv SETTINGS frame <length=6, flags=0x00, stream_id=0>
          (niv=1)
          [SETTINGS_MAX_HEADER_LIST_SIZE(0x06):8192]
[  0.005] send SETTINGS frame <length=0, flags=0x01, stream_id=0>
          ; ACK
          (niv=0)
[  0.006] recv GOAWAY frame <length=104, flags=0x00, stream_id=0>
          (last_stream_id=0, error_code=PROTOCOL_ERROR(0x01), opaque_data(96)=[HTTP/2 client preface string missing or corrupt. Hex dump for received bytes: 000000040100000000])
Some requests were not processed. total=1, processed=0
```

Run the `Http2Server` example code, using the following modified `configureClearText` method:

```java
    private void configureClearText(SocketChannel ch) {
        final ChannelPipeline p = ch.pipeline();
        final HttpServerCodec sourceCodec = new HttpServerCodec();

        final Http2Codec http2Codec = new Http2CodecBuilder(true, new ChannelInitializer<Channel>() {
            @Override
            protected void initChannel(Channel channel) throws Exception {
                ChannelPipeline p = channel.pipeline();
                p.addLast("downgrader", new Http2ServerDowngrader(true));
                p.addLast("chunk-aggregator", new HttpObjectAggregator(1048576));
                p.addLast("http1-handler", new HelloWorldHttp1Handler("Downgrade"));
            }
        }).build();

        final UpgradeCodecFactory upgradeCodecFactory = new UpgradeCodecFactory() {
            @Override
            public UpgradeCodec newUpgradeCodec(CharSequence protocol) {
                if (AsciiString.contentEquals(Http2CodecUtil.HTTP_UPGRADE_PROTOCOL_NAME, protocol)) {
                    return new Http2ServerUpgradeCodec("server-upgrade-codec", http2Codec);
                } else {
                    return null;
                }
            }
        };

        final HttpServerUpgradeHandler upgradeHandler = new HttpServerUpgradeHandler(sourceCodec, upgradeCodecFactory);

        final CleartextHttp2ServerUpgradeHandler cleartextHttp2ServerUpgradeHandler =
            new CleartextHttp2ServerUpgradeHandler(sourceCodec, upgradeHandler, http2Codec);

        p.addLast(cleartextHttp2ServerUpgradeHandler);

        p.addLast(new SimpleChannelInboundHandler<HttpMessage>() {
            @Override
            protected void channelRead0(ChannelHandlerContext ctx, HttpMessage msg) throws Exception {
                // If this handler is hit then no upgrade has been attempted and the client is just talking HTTP.
                System.err.println("Directly talking: " + msg.protocolVersion() + " (no upgrade was attempted)");
                ChannelPipeline pipeline = ctx.pipeline();
                ChannelHandlerContext thisCtx = pipeline.context(this);
                pipeline.addAfter(thisCtx.name(), null, new HelloWorldHttp1Handler("Direct. No Upgrade Attempted."));
                pipeline.replace(this, null, new HttpObjectAggregator(maxHttpContentLength));
                ctx.fireChannelRead(ReferenceCountUtil.retain(msg));
            }
        });

        p.addLast(new UserEventLogger());
    }
```

### Netty version

Latest `4.1` branch

java 8u131

macOS 10.12.4

chhsiao90:
Motivation:

When we use pipeline.replace and we still had ongoing inbound, then
there will be some problem that inbound message would go to wrong
handlers. So we add handler first, and remove self after add, so that
the next handler will be the correct one.

Modifications:

Uses remove after addAfter instead of replace.

Result:

Fixed #6881

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandler.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandler.java
@@ -89,8 +89,11 @@ public final class CleartextHttp2ServerUpgradeHandler extends ChannelHandlerAdap
                 // following network traffic
                 ctx.pipeline()
                    .remove(httpServerCodec)
-                   .remove(httpServerUpgradeHandler)
-                   .replace(this, null, http2ServerHandler);
+                   .remove(httpServerUpgradeHandler);
+
+                ctx.pipeline().addAfter(ctx.name(), null, http2ServerHandler);
+                ctx.pipeline().remove(this);
+
                 ctx.fireUserEventTriggered(PriorKnowledgeUpgradeEvent.INSTANCE);
             }
         }
--- a/codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
+++ b/codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
@@ -212,7 +212,7 @@ public class CleartextHttp2ServerUpgradeHandlerTest {
             }
         });
 
-        channel.writeInbound(Http2CodecUtil.connectionPrefaceBuf());
+        assertFalse(channel.writeInbound(Http2CodecUtil.connectionPrefaceBuf()));
 
         ByteBuf settingsFrame = settingsFrameBuf();
 
--- a/codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
+++ b/codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
@@ -17,8 +17,10 @@ package io.netty.handler.codec.http2;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
+import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.channel.ChannelInitializer;
 import io.netty.channel.embedded.EmbeddedChannel;
 import io.netty.handler.codec.http.DefaultHttpHeaders;
 import io.netty.handler.codec.http.HttpMethod;
@@ -35,7 +37,6 @@ import io.netty.handler.codec.http2.Http2Stream.State;
 import io.netty.util.CharsetUtil;
 import io.netty.util.ReferenceCountUtil;
 import org.junit.After;
-import org.junit.Before;
 import org.junit.Test;
 
 import java.util.ArrayList;
@@ -56,8 +57,7 @@ public class CleartextHttp2ServerUpgradeHandlerTest {
 
     private List<Object> userEvents;
 
-    @Before
-    public void setUp() {
+    private void setUpServerChannel() {
         frameListener = mock(Http2FrameListener.class);
 
         http2ConnectionHandler = new Http2ConnectionHandlerBuilder().frameListener(frameListener).build();
@@ -91,6 +91,8 @@ public class CleartextHttp2ServerUpgradeHandlerTest {
 
     @Test
     public void priorKnowledge() throws Exception {
+        setUpServerChannel();
+
         channel.writeInbound(Http2CodecUtil.connectionPrefaceBuf());
 
         ByteBuf settingsFrame = settingsFrameBuf();
@@ -109,6 +111,8 @@ public class CleartextHttp2ServerUpgradeHandlerTest {
 
     @Test
     public void upgrade() throws Exception {
+        setUpServerChannel();
+
         String upgradeString = "GET / HTTP/1.1\r\n" +
                                "Host: example.com\r\n" +
                                "Connection: Upgrade, HTTP2-Settings\r\n" +
@@ -138,6 +142,8 @@ public class CleartextHttp2ServerUpgradeHandlerTest {
 
     @Test
     public void priorKnowledgeInFragments() throws Exception {
+        setUpServerChannel();
+
         ByteBuf connectionPreface = Http2CodecUtil.connectionPrefaceBuf();
         assertFalse(channel.writeInbound(connectionPreface.readBytes(5), connectionPreface));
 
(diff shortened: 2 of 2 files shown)


codec-http2/src/main/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandler.java
```diff
@@ -89,8 +89,11 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) t
...
                 ctx.pipeline()
                    .remove(httpServerCodec)
-                   .remove(httpServerUpgradeHandler)
-                   .replace(this, null, http2ServerHandler);
+                   .remove(httpServerUpgradeHandler);
+
+                ctx.pipeline().addAfter(ctx.name(), null, http2ServerHandler);
+                ctx.pipeline().remove(this);
+
                 ctx.fireUserEventTriggered(PriorKnowledgeUpgradeEvent.INSTANCE);
```
rschmitt: Is this used by anything?

codec-http2/src/main/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandler.java
```diff
@@ -89,8 +89,11 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) t
                 // following network traffic
                 ctx.pipeline()
                    .remove(httpServerCodec)
-                   .remove(httpServerUpgradeHandler)
-                   .replace(this, null, http2ServerHandler);
+                   .remove(httpServerUpgradeHandler);
+
```
rschmitt: Fair enough, but this seems like a workaround for a deeper problem in `replace`. Is there anything in the documented contract of `replace` that would cause us to expect it to behave differently than `addAfter` followed by `remove`?

codec-http2/src/main/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandler.java
rschmitt: (The `PriorKnowledgeUpgradeEvent`, that is.)

chhsiao90: In netty itself, no. But maybe the api user would like to know there is an upgrade event fired.

chhsiao90:
About the test, I think [CleartextHttp2ServerUpgradeHandlerTest](https://github.com/netty/netty/blob/4.1/codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java) might enough. If the changed break something, the test will failed.
But if still need a test for #6881 , I think I can address it.

rschmitt:
Well, obviously a new test is needed, because this bug passed all the tests.

normanmaurer:
@chhsiao90 as @rschmitt said we need a test :) Will look at the code soon.

chhsiao90:
Sure, I will add the unit test later!

codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
```diff
@@ -175,6 +183,48 @@ public void downgrade() throws Exception {
...
+        CleartextHttp2ServerUpgradeHandler handler = new CleartextHttp2ServerUpgradeHandler(
+                httpServerCodec, upgradeHandler, http2Codec);
+        channel = new EmbeddedChannel(handler, new ChannelInboundHandlerAdapter() {
+            @Override
+            public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
+                userEvents.add(evt);
+            }
+        });
+
+        channel.writeInbound(Http2CodecUtil.connectionPrefaceBuf());
```
normanmaurer: @chhsiao90 please assert the return value.

chhsiao90: Addressed by 7e4fa2dfdd832d00a0ad76d7ea4ee8533791ad7c

Please let me know if the pr is ready for merge, I can squash the commits.

Scottmitch:
I quickly read over the issue description and it seems like we are treating a symptom of the problem instead of the root cause. I think this warrants more investigation. For example `ByteToMessageDecoder` does aggregate and takes care to forward on data when a replace occurs, however `Http2Codec` doesn't seem to be as careful when manipulating the pipeline ... is this correct, and/or are there other issues at play here?

normanmaurer:
Squashed and cherry-picked into 4.1 (8320a45c156f87afcdb545218f49d561b42843f9)

Issue #6951, opened 2017-07-06
Writing to released ByteBuf


fenik17:
What is the concept of checking the possibility of accessing to `ByteBuf`?

Probably a bug. A `writeCharSequence` (`setCharSequence`) can allow writing to released buffer:
```java
    ByteBuf buf = Unpooled.directBuffer(16);
    buf.release();
    buf.writeCharSequence("1", CharsetUtil.US_ASCII); // <-- no exceptions
```
With an `UnpooledHeapByteBuf` this code leads to `IllegalReferenceCountException` due to call  `ensureAccessible()` in the `capacity()`. IIUC this done only to prevent NPE for released buffed. Perhaps it's better to rewrite the method like this:
```java
    public int capacity() {
        byte[] array = this.array;
        if (array != null) {
            return array.length;
        }
        // array is null if refCnt == 0
        ensureAccessible();
        return 0;
    }
```

Also, a `setCharSequence` - the only `set` method which can increase capacity...

Scottmitch:
We should check the ref count before writing to the underlying storage. Typically method use [checIndex](https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java#L1346) to accomplish this. I'm AFK but can look in more detail later.

normanmaurer:
That's a bug... looking now

normanmaurer:
Ok I have a fix... stay tuned

normanmaurer:
@fenik17 please check https://github.com/netty/netty/pull/6953

fenik17:
@normanmaurer #6953 this is OK. Thanks! But what about this:
> a setCharSequence - the only set method which can increase capacity...

IMHO, this is not fine.. For example:
```java
        ByteBuf buf = Unpooled.directBuffer(1);
        buf.setCharSequence(0, "AB", CharsetUtil.US_ASCII); // no exception. capacity == 2

        buf.resetWriterIndex().capacity(1);
        buf.setBytes(0, new byte[]{1, 2}); // IndexOutOfBoundsException
```
I think only `write` methods should increase capacity.

fenik17:
> I think only write methods should increase capacity.

Although, if fix it now, someone's code can breaks...

normanmaurer:
@fenik17 let me investigate later today. Need to finish some other stuff first.

normanmaurer:
[...]

PR #6953, opened 2017-07-07
Add tests to ensure an IllegalReferenceCountException is thrown if set/writeCharSequence is called on a released buffer

normanmaurer:
Motivation:

    We need to ensure we not allow calling set/writeCharsequence on an released ByteBuf.

    Modifications:

    Add test-cases

    Result:

    Proves fix of [#6951].

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/AbstractUnpooledSlicedByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/AbstractUnpooledSlicedByteBuf.java
@@ -261,7 +261,7 @@ abstract class AbstractUnpooledSlicedByteBuf extends AbstractDerivedByteBuf {
     @Override
     public CharSequence getCharSequence(int index, int length, Charset charset) {
         checkIndex0(index, length);
-        return buffer.getCharSequence(idx(index), length, charset);
+        return unwrap().getCharSequence(idx(index), length, charset);
     }
 
     @Override
@@ -384,11 +384,6 @@ abstract class AbstractUnpooledSlicedByteBuf extends AbstractDerivedByteBuf {
         checkIndex0(index, src.remaining());
         unwrap().setBytes(idx(index), src);
         return this;
-    }
-
-    @Override
-    public int setCharSequence(int index, CharSequence sequence, Charset charset) {
-        return super.setCharSequence(idx(index), sequence, charset);
     }
 
     @Override
--- a/buffer/src/test/java/io/netty/buffer/AbstractByteBufTest.java
+++ b/buffer/src/test/java/io/netty/buffer/AbstractByteBufTest.java
@@ -2502,6 +2502,10 @@ public abstract class AbstractByteBufTest {
 
     private ByteBuf releasedBuffer() {
         ByteBuf buffer = newBuffer(8);
+
+        // Clear the buffer so we are sure the reader and writer indices are 0.
+        // This is important as we may return a slice from newBuffer(...).
+        buffer.clear();
         assertTrue(buffer.release());
         return buffer;
     }
@@ -2764,6 +2768,30 @@ public abstract class AbstractByteBufTest {
         } finally {
             buffer.release();
         }
+    }
+
+    @Test(expected = IllegalReferenceCountException.class)
+    public void testSetUsAsciiCharSequenceAfterRelease() {
+        testSetCharSequenceAfterRelease0(CharsetUtil.US_ASCII);
+    }
+
+    @Test(expected = IllegalReferenceCountException.class)
+    public void testSetIso88591CharSequenceAfterRelease() {
+        testSetCharSequenceAfterRelease0(CharsetUtil.ISO_8859_1);
+    }
+
+    @Test(expected = IllegalReferenceCountException.class)
+    public void testSetUtf8CharSequenceAfterRelease() {
+        testSetCharSequenceAfterRelease0(CharsetUtil.UTF_8);
+    }
+
+    @Test(expected = IllegalReferenceCountException.class)
+    public void testSetUtf16CharSequenceAfterRelease() {
+        testSetCharSequenceAfterRelease0(CharsetUtil.UTF_16);
+    }
+
+    private void testSetCharSequenceAfterRelease0(Charset charset) {
+        releasedBuffer().setCharSequence(0, "x", charset);
     }
 
     @Test(expected = IllegalReferenceCountException.class)
@@ -3087,6 +3115,30 @@ public abstract class AbstractByteBufTest {
     }
 
     @Test(expected = IllegalReferenceCountException.class)
+    public void testWriteUsAsciiCharSequenceAfterRelease() {
+        testWriteCharSequenceAfterRelease0(CharsetUtil.US_ASCII);
+    }
+
+    @Test(expected = IllegalReferenceCountException.class)
+    public void testWriteIso88591CharSequenceAfterRelease() {
+        testWriteCharSequenceAfterRelease0(CharsetUtil.ISO_8859_1);
+    }
+
+    @Test(expected = IllegalReferenceCountException.class)
+    public void testWriteUtf8CharSequenceAfterRelease() {
+        testWriteCharSequenceAfterRelease0(CharsetUtil.UTF_8);
+    }
+
+    @Test(expected = IllegalReferenceCountException.class)
+    public void testWriteUtf16CharSequenceAfterRelease() {
+        testWriteCharSequenceAfterRelease0(CharsetUtil.UTF_16);
+    }
+
+    private void testWriteCharSequenceAfterRelease0(Charset charset) {
+        releasedBuffer().writeCharSequence("x", charset);
+    }
+
+    @Test(expected = IllegalReferenceCountException.class)
     public void testForEachByteAfterRelease() {
         releasedBuffer().forEachByte(new TestByteProcessor());
     }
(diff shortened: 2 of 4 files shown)


normanmaurer:
@Scottmitch I was wondering if we should just check the `ensureAccessible()` in `ensureWritable(...)` wdyt ?

buffer/src/test/java/io/netty/buffer/AbstractByteBufTest.java
```diff
@@ -2766,6 +2767,22 @@ public void testSetBytesAfterRelease3() {
     }
 
     @Test(expected = IllegalReferenceCountException.class)
+    public void testSetCharSequenceAfterRelease() {
+        testSetCharSequenceAfterRelease0(CharsetUtil.US_ASCII);
+        testSetCharSequenceAfterRelease0(CharsetUtil.ISO_8859_1);
```
fenik17: This line will not be executed.

normanmaurer: stupid me ... You are right. let me fix it

buffer/src/test/java/io/netty/buffer/AbstractByteBufTest.java
```diff
@@ -2766,6 +2767,22 @@ public void testSetBytesAfterRelease3() {
...
+        testSetCharSequenceAfterRelease0(CharsetUtil.UTF_8);
+    }
+
+    private void testSetCharSequenceAfterRelease0(Charset charset) {
+        ByteBuf buffer = buffer();
+        try {
+            releasedBuffer().setCharSequence(0, "Test", charset);
+        } finally {
+            buffer.release();
```
fenik17: Why need second release? I think such will be enough:
```java
@Test(expected = IllegalReferenceCountException.class)
public void testSetCharSequenceAfterReleaseUtf8() {
    releasedBuffer().setCharSequence(0, "Test", CharsetUtil.UTF_8);
}

@Test(expected = IllegalReferenceCountException.class)
public void testSetCharSequenceAfterReleaseAscii() {
    releasedBuffer().setCharSequence(0, "Test", CharsetUtil.US_ASCII);
}
```
This approach is used in other tests for checking `IllegalReferenceCountException`.

buffer/src/test/java/io/netty/buffer/SlicedByteBufTest.java
```diff
@@ -118,6 +118,12 @@ public void testGetReadOnlyHeapDst() {
         super.testGetReadOnlyHeapDst();
     }
 
+    @Test(expected = IndexOutOfBoundsException.class)
```
fenik17: `IndexOutOfBoundsException` -> `IllegalReferenceCountException`?

normanmaurer:
@fenik17 fixed all the problems

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -464,6 +464,7 @@ public static int writeUtf8(ByteBuf buf, CharSequence seq) {
 
     // Fast-Path implementation
     static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq, int len) {
+        buffer.ensureAccessible();
```
Scottmitch: it seems more natural/consistent for the ByteBuf implementation to enforce this "internal" constraint instead of enforcing this externally. However I guess this is the equivalent of the "set" operations in the ByteBuf implementations which typically provide the accessible check ... so maybe this is OK.

Scottmitch: > @Scottmitch I was wondering if we should just check the ensureAccessible() in ensureWritable(...) wdyt ?

Just saw this message ... yes this sounds better to me. However we should revisit usage of `ensureWritable` because we may be able to change `checkIndex` to `checkIndex0`.

normanmaurer: @Scottmitch ok let me check

normanmaurer: https://github.com/netty/netty/pull/6978

normanmaurer:
@Scottmitch @fenik17 PTAL again just rebased.

normanmaurer:
@fenik17 sorry but PTAL again... I just pushed some more changes I missed before.

buffer/src/test/java/io/netty/buffer/AbstractByteBufTest.java
```diff
@@ -2501,6 +2502,10 @@ public void testLittleEndianWithExpand() {
 
     private ByteBuf releasedBuffer() {
         ByteBuf buffer = newBuffer(8);
+
+        // Clear the buffer so we are sure the reader and writer indices are 0.
+        // This is important as we may return a slice from newBuffer(...).
+        buffer.clear();
```
normanmaurer: This is needed as otherwise the test may fail with `IndexOutOfBoundsException` and not `IllegalReferenceCountException` when executed with a slice buffer

normanmaurer:
@Scottmitch yep.. will merge #6957 first and then rebase this one

buffer/src/test/java/io/netty/buffer/RetainedSlicedByteBufTest.java
```diff
@@ -26,8 +25,7 @@
     protected ByteBuf newBuffer(int length, int maxCapacity) {
         Assume.assumeTrue(maxCapacity == Integer.MAX_VALUE);
         ByteBuf wrapped = Unpooled.wrappedBuffer(new byte[length * 2]);
-        ByteBuf buffer = wrapped.retainedSlice(length > 1 ?
-                        PlatformDependent.threadLocalRandom().nextInt(length - 1) + 1 : 0, length);
+        ByteBuf buffer = wrapped.retainedSlice(length > 1 ? 1 : 0, length);
```
normanmaurer: please note this is needed as otherwise you will sometimes get test errors as we may create a slice which is not large enough to write the requested bytes in. For example we may need 4 bytes and specify a length of 8 (which should be good enough). Now because of `threadLocalRandom().nextInt(...) + 1` we may end up with an offset that is >= 5 and so the write will fail.

buffer/src/test/java/io/netty/buffer/SlicedByteBufTest.java
```diff
@@ -34,8 +32,7 @@
     protected ByteBuf newBuffer(int length, int maxCapacity) {
         Assume.assumeTrue(maxCapacity == Integer.MAX_VALUE);
         ByteBuf buffer = Unpooled.wrappedBuffer(
-                new byte[length * 2], length > 1 ?
-                        PlatformDependent.threadLocalRandom().nextInt(length - 1) + 1 : 0, length);
+                new byte[length * 2], length > 1 ? 1 : 0, length);
```
normanmaurer: See https://github.com/netty/netty/pull/6953/files#r128322865

normanmaurer:
@Scottmitch @fenik17 PTAL again... This basically only adds the test-cases to prove this was fixes now (as the fix was done before).

buffer/src/test/java/io/netty/buffer/RetainedSlicedByteBufTest.java
Scottmitch: Would this give us more variability in the test while still ensuring we have at least `length` bytes?
```java
start = /* random number from 0 to length * 2 - length */
ByteBuf buffer = wrapped.retainedSlice(start, length);
```

normanmaurer:
@Scottmitch adjusted how we calculate the slice we want to take in the tests and also found a bug in our sliced buffer implementations which I fixed here (the test failed before after adjusting how we calculate the slice).

normanmaurer:
@Scottmitch @fenik17 PTAL again... sorry

buffer/src/test/java/io/netty/buffer/SlicedByteBufTest.java
```diff
@@ -31,14 +30,18 @@
...
-    protected ByteBuf newBuffer(int length, int maxCapacity) {
+    protected final ByteBuf newBuffer(int length, int maxCapacity) {
         Assume.assumeTrue(maxCapacity == Integer.MAX_VALUE);
-        ByteBuf buffer = Unpooled.wrappedBuffer(
-                new byte[length * 2], length > 1 ?
-                        PlatformDependent.threadLocalRandom().nextInt(length - 1) + 1 : 0, length);
-        assertEquals(0, buffer.readerIndex());
-        assertEquals(length, buffer.writerIndex());
-        return buffer;
+        int offset = length == 0 ? 1 : PlatformDependent.threadLocalRandom().nextInt(length);
```
Scottmitch: if length is 0 we will allocate a buffer of size 0, but take a slice starting from index 1 ... won't this be IOOB?

buffer/src/test/java/io/netty/buffer/RetainedSlicedByteBufTest.java
```diff
@@ -16,22 +16,16 @@
...
-        ByteBuf buffer = wrapped.retainedSlice(length > 1 ?
-                        PlatformDependent.threadLocalRandom().nextInt(length - 1) + 1 : 0, length);
-        wrapped.release();
 
-        assertEquals(0, buffer.readerIndex());
-        assertEquals(length, buffer.writerIndex());
-        return buffer;
+    @Override
+    protected ByteBuf newSlice(ByteBuf buffer, int offset, int length) {
+        ByteBuf slice =  buffer.retainedSlice(offset, length);
```
Scottmitch: nit: remove double space after `=`

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -673,6 +673,20 @@ public int setCharSequence(int index, CharSequence sequence, Charset charset) {
...
+        int written = setCharSequenceOptimized(index, sequence, charset, expand);
+        if (written >= 0) {
+            return written;
+        }
+        byte[] bytes = sequence.toString().getBytes(charset);
+        if (expand) {
+            ensureWritable0(bytes.length);
+            // setBytes(...) will take care of checking the indices.
+        }
+        setBytes(index, bytes);
```
Scottmitch: good catch ... I guess the slice would try to apply its adjustment here too.

buffer/src/test/java/io/netty/buffer/SlicedByteBufTest.java
normanmaurer: @Scottmitch you are right... if its 0 we should use 0 ... brainfart

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
normanmaurer: yep

normanmaurer:
@Scottmitch addressed

fenik17:
@normanmaurer IIUC the `set/writeCharSequence` methods in `PooledSlicedByteBuf` before worked wrong, because they were not overriden? Can you add tests for sliced buffers with an `adjustment` > 0?

fenik17:
At least, if I now (on `illegal_access_sequence` branch) remove overriden from `PooledSlicedByteBuf`, all tests passing.

normanmaurer:
@fenik17 adding more tests... stay tuned.

buffer/src/main/java/io/netty/buffer/AbstractUnpooledSlicedByteBuf.java
```diff
@@ -387,11 +387,6 @@ public ByteBuf setBytes(int index, ByteBuffer src) {
     }
 
     @Override
-    public int setCharSequence(int index, CharSequence sequence, Charset charset) {
-        return super.setCharSequence(idx(index), sequence, charset);
```
fenik17: Yeah, `_setByte()` inside `setCharSequence()` already use `idx(`)  :)

fenik17:
I think, now is Ok.

buffer/src/main/java/io/netty/buffer/AbstractUnpooledSlicedByteBuf.java
normanmaurer: yep

normanmaurer:
Cherry-picked into 4.1 (06f64948d5da798ed75c892f6cbba18566857937) and also only cherry-picked how we calculate the slice offset and length in the test to 4.0 (b88537bb7a)

PR #6954, opened 2017-07-07
Http2ConnectionHandler: allow graceful shutdown to wait forever

zpencer:
Motivation:

There should be a way to allow graceful shutdown to wait for all open streams to close without a timeout. Using `gracefulShutdownTimeoutMillis` with a large value is a bit of a hack, and has a gotcha that sufficiently large values will overflow the long, resulting in a `ClosingChannelFutureListener` that executes immediately.

Modification:

Add a new boolean flag to express waiting until all streams are closed.

Result:

We can now shutdown the connection without a forced timeout.

Fixes #<GitHub issue number>. 

See motivation.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/AbstractHttp2ConnectionHandlerBuilder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/AbstractHttp2ConnectionHandlerBuilder.java
@@ -153,19 +153,11 @@ public abstract class AbstractHttp2ConnectionHandlerBuilder<T extends Http2Conne
      * Sets the graceful shutdown timeout of the {@link Http2Connection} in milliseconds.
      */
     protected B gracefulShutdownTimeoutMillis(long gracefulShutdownTimeoutMillis) {
-        if (gracefulShutdownTimeoutMillis < 0) {
+        if (gracefulShutdownTimeoutMillis < -1) {
             throw new IllegalArgumentException("gracefulShutdownTimeoutMillis: " + gracefulShutdownTimeoutMillis +
-                                               " (expected: >= 0)");
+                                               " (expected: -1 for indefinite or >= 0)");
         }
         this.gracefulShutdownTimeoutMillis = gracefulShutdownTimeoutMillis;
-        return self();
-    }
-
-    /**
-     * Sets the graceful shutdown timeout to be indefinite.
-     */
-    protected B gracefulShutdownIndefiniteWait() {
-        gracefulShutdownTimeoutMillis = -1;
         return self();
     }
 
@@ -434,11 +426,7 @@ public abstract class AbstractHttp2ConnectionHandlerBuilder<T extends Http2Conne
         }
 
         // Setup post build options
-        if (gracefulShutdownTimeoutMillis == -1) {
-            handler.gracefulShutdownIndefiniteWait();
-        } else {
-            handler.gracefulShutdownTimeoutMillis(gracefulShutdownTimeoutMillis);
-        }
+        handler.gracefulShutdownTimeoutMillis(gracefulShutdownTimeoutMillis);
         if (handler.decoder().frameListener() == null) {
             handler.decoder().frameListener(frameListener);
         }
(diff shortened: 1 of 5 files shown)


codec-http2/src/main/java/io/netty/handler/codec/http2/AbstractHttp2ConnectionHandlerBuilder.java
```diff
@@ -85,6 +85,7 @@
     private Http2Settings initialSettings = Http2Settings.defaultSettings();
     private Http2FrameListener frameListener;
     private long gracefulShutdownTimeoutMillis = DEFAULT_GRACEFUL_SHUTDOWN_TIMEOUT_MILLIS;
+    private boolean gracefulShutdownIndefiniteWait;
```
Scottmitch: instead of adding extra state can we just allow `gracefulShutdownTimeoutMillis` to be negative and that represent infinite time?

normanmaurer: +1

codec-http2/src/main/java/io/netty/handler/codec/http2/AbstractHttp2ConnectionHandlerBuilder.java
```diff
@@ -152,11 +153,23 @@ protected long gracefulShutdownTimeoutMillis() {
...
+                                               " (expected: >= 0)");
+        }
         this.gracefulShutdownTimeoutMillis = gracefulShutdownTimeoutMillis;
         return self();
     }
 
     /**
+     * Sets the graceful shutdown timeout to be indefinite.
+     */
+    protected B gracefulShutdownIndefiniteWait() {
```
normanmaurer: remove this one and just have `gracefulShutdownTimeoutMillis(....)` accept -1 ?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -112,6 +113,14 @@ public void gracefulShutdownTimeoutMillis(long gracefulShutdownTimeoutMillis) {
         this.gracefulShutdownTimeoutMillis = gracefulShutdownTimeoutMillis;
     }
 
+    /**
+     * Configures this endpoint to wait until all streams are closed before closing the connection during the graceful
+     * shutdown process.
+     */
+    public void gracefulShutdownIndefiniteWait() {
```
normanmaurer: see above

codec-http2/src/main/java/io/netty/handler/codec/http2/AbstractHttp2ConnectionHandlerBuilder.java
zpencer: I was originally trying to keep the existing API behavior the same, but allowing -1 is certainly more natural here. Changed.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
zpencer: Changed here as well.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -454,8 +455,12 @@ private void doGracefulShutdown(ChannelHandlerContext ctx, ChannelFuture future,
             future.addListener(new ClosingChannelFutureListener(ctx, promise));
         } else {
             // If there are active streams we should wait until they are all closed before closing the connection.
-            closeListener = new ClosingChannelFutureListener(ctx, promise,
-                                                             gracefulShutdownTimeoutMillis, MILLISECONDS);
+            if (gracefulShutdownTimeoutMillis == -1) {
```
Scottmitch: nit: we could also just use `< 0` here. at this point we don't care if it is a specific token value (e.g. `-1`) or some other negative number.

normanmaurer:
squashed and cherry-picked into 4.1 (732b145842).

@zpencer thanks!

PR #6962, opened 2017-07-11
HTTP/2 enforce HTTP message flow

Scottmitch:
Motivation:
codec-http2 currently does not strictly enforce the HTTP/1.x semantics with respect to the number of headers defined in RFC 7540 Section 8.1 [1]. We currently don't validate the number of headers nor do we validate that the trailing headers should indicate EOS.

[1] https://tools.ietf.org/html/rfc7540#section-8.1

Modifications:
- DefaultHttp2ConnectionDecoder should only allow decoding of a single headers and a single trailers
- DefaultHttp2ConnectionEncoder should only allow encoding of a single headers and optionally a single trailers

Result:
Constraints of RFC 7540 restricting the number of headers/trailers is enforced.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpStatusClass.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpStatusClass.java
@@ -74,6 +74,27 @@ public enum HttpStatusClass {
         return UNKNOWN;
     }
 
+    /**
+     * Returns the class of the specified HTTP status code.
+     * @param code Just the numeric portion of the http status code.
+     */
+    public static HttpStatusClass valueOf(CharSequence code) {
+        if (code != null && code.length() == 3) {
+            char c0 = code.charAt(0);
+            return isDigit(c0) && isDigit(code.charAt(1)) && isDigit(code.charAt(2)) ? valueOf(digit(c0) * 100)
+                                                                                     : UNKNOWN;
+        }
+        return UNKNOWN;
+    }
+
+    private static int digit(char c) {
+        return c - '0';
+    }
+
+    private static boolean isDigit(char c) {
+        return c >= '0' && c <= '9';
+    }
+
     private final int min;
     private final int max;
     private final AsciiString defaultReasonPhrase;
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
@@ -373,13 +373,16 @@ public class DefaultHttp2Connection implements Http2Connection {
      * Simple stream implementation. Streams can be compared to each other by priority.
      */
     private class DefaultStream implements Http2Stream {
-        private static final byte SENT_STATE_RST = 0x1;
-        private static final byte SENT_STATE_HEADERS = 0x2;
-        private static final byte SENT_STATE_PUSHPROMISE = 0x4;
+        private static final byte META_STATE_SENT_RST = 1;
+        private static final byte META_STATE_SENT_HEADERS = 1 << 1;
+        private static final byte META_STATE_SENT_TRAILERS = 1 << 2;
+        private static final byte META_STATE_SENT_PUSHPROMISE = 1 << 3;
+        private static final byte META_STATE_RECV_HEADERS = 1 << 4;
+        private static final byte META_STATE_RECV_TRAILERS = 1 << 5;
         private final int id;
         private final PropertyMap properties = new PropertyMap();
         private State state;
-        private byte sentState;
+        private byte metaState;
 
         DefaultStream(int id, State state) {
             this.id = id;
@@ -398,35 +401,60 @@ public class DefaultHttp2Connection implements Http2Connection {
 
         @Override
         public boolean isResetSent() {
-            return (sentState & SENT_STATE_RST) != 0;
+            return (metaState & META_STATE_SENT_RST) != 0;
         }
 
         @Override
         public Http2Stream resetSent() {
-            sentState |= SENT_STATE_RST;
+            metaState |= META_STATE_SENT_RST;
             return this;
         }
 
         @Override
-        public Http2Stream headersSent() {
-            sentState |= SENT_STATE_HEADERS;
+        public Http2Stream headersSent(boolean isInformational) {
+            if (!isInformational) {
+                metaState |= isHeadersSent() ? META_STATE_SENT_TRAILERS : META_STATE_SENT_HEADERS;
+            }
             return this;
         }
 
         @Override
         public boolean isHeadersSent() {
-            return (sentState & SENT_STATE_HEADERS) != 0;
+            return (metaState & META_STATE_SENT_HEADERS) != 0;
+        }
+
+        @Override
+        public boolean isTrailersSent() {
+            return (metaState & META_STATE_SENT_TRAILERS) != 0;
+        }
+
+        @Override
+        public Http2Stream headersReceived(boolean isInformational) {
+            if (!isInformational) {
+                metaState |= isHeadersReceived() ? META_STATE_RECV_TRAILERS : META_STATE_RECV_HEADERS;
+            }
+            return this;
+        }
+
+        @Override
+        public boolean isHeadersReceived() {
+            return (metaState & META_STATE_RECV_HEADERS) != 0;
+        }
+
+        @Override
+        public boolean isTrailersReceived() {
+            return (metaState & META_STATE_RECV_TRAILERS) != 0;
         }
 
         @Override
         public Http2Stream pushPromiseSent() {
-            sentState |= SENT_STATE_PUSHPROMISE;
+            metaState |= META_STATE_SENT_PUSHPROMISE;
             return this;
         }
 
         @Override
         public boolean isPushPromiseSent() {
-            return (sentState & SENT_STATE_PUSHPROMISE) != 0;
+            return (metaState & META_STATE_SENT_PUSHPROMISE) != 0;
         }
 
         @Override
@@ -599,7 +627,7 @@ public class DefaultHttp2Connection implements Http2Connection {
         }
 
         @Override
-        public Http2Stream headersSent() {
+        public Http2Stream headersSent(boolean isInformational) {
             throw new UnsupportedOperationException();
         }
 
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
@@ -16,6 +16,7 @@ package io.netty.handler.codec.http2;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.http.HttpStatusClass;
 import io.netty.handler.codec.http2.Http2Connection.Endpoint;
 import io.netty.util.internal.UnstableApi;
 import io.netty.util.internal.logging.InternalLogger;
@@ -23,6 +24,7 @@ import io.netty.util.internal.logging.InternalLoggerFactory;
 
 import java.util.List;
 
+import static io.netty.handler.codec.http.HttpStatusClass.INFORMATIONAL;
 import static io.netty.handler.codec.http2.Http2CodecUtil.DEFAULT_PRIORITY_WEIGHT;
 import static io.netty.handler.codec.http2.Http2Error.INTERNAL_ERROR;
 import static io.netty.handler.codec.http2.Http2Error.PROTOCOL_ERROR;
@@ -282,6 +284,14 @@ public class DefaultHttp2ConnectionDecoder implements Http2ConnectionDecoder {
                 return;
             }
 
+            boolean isInformational = !connection.isServer() &&
+                    HttpStatusClass.valueOf(headers.status()) == INFORMATIONAL;
+            if ((isInformational || !endOfStream) && stream.isHeadersReceived() || stream.isTrailersReceived()) {
+                throw streamError(streamId, PROTOCOL_ERROR,
+                                  "Stream %d received too many headers EOS: %s state: %s",
+                                  streamId, endOfStream, stream.state());
+            }
+
             switch (stream.state()) {
                 case RESERVED_REMOTE:
                     stream.open(endOfStream);
@@ -305,6 +315,7 @@ public class DefaultHttp2ConnectionDecoder implements Http2ConnectionDecoder {
                             stream.state());
             }
 
+            stream.headersReceived(isInformational);
             encoder.flowController().updateDependencyTree(streamId, streamDependency, weight, exclusive);
 
             listener.onHeadersRead(ctx, streamId, headers, streamDependency, weight, exclusive, padding, endOfStream);
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
@@ -21,10 +21,12 @@ import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelPromise;
 import io.netty.channel.CoalescingBufferQueue;
+import io.netty.handler.codec.http.HttpStatusClass;
 import io.netty.util.internal.UnstableApi;
 
 import java.util.ArrayDeque;
 
+import static io.netty.handler.codec.http.HttpStatusClass.INFORMATIONAL;
 import static io.netty.handler.codec.http2.Http2CodecUtil.DEFAULT_PRIORITY_WEIGHT;
 import static io.netty.handler.codec.http2.Http2Error.PROTOCOL_ERROR;
 import static io.netty.handler.codec.http2.Http2Exception.connectionError;
@@ -145,6 +147,15 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
         return writeHeaders(ctx, streamId, headers, 0, DEFAULT_PRIORITY_WEIGHT, false, padding, endStream, promise);
     }
 
+    private static boolean validateHeadersSentState(Http2Stream stream, Http2Headers headers, boolean isServer,
+                                                    boolean endOfStream) {
+        boolean isInformational = isServer && HttpStatusClass.valueOf(headers.status()) == INFORMATIONAL;
+        if ((isInformational || !endOfStream) && stream.isHeadersSent() || stream.isTrailersSent()) {
+            throw new IllegalStateException("Stream " + stream.id() + " sent too many headers EOS: " + endOfStream);
+        }
+        return isInformational;
+    }
+
     @Override
     public ChannelFuture writeHeaders(final ChannelHandlerContext ctx, final int streamId,
             final Http2Headers headers, final int streamDependency, final short weight,
@@ -180,6 +191,7 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
             // for this stream.
             Http2RemoteFlowController flowController = flowController();
             if (!endOfStream || !flowController.hasFlowControlled(stream)) {
+                boolean isInformational = validateHeadersSentState(stream, headers, connection.isServer(), endOfStream);
                 if (endOfStream) {
                     final Http2Stream finalStream = stream;
                     final ChannelFutureListener closeStreamLocalListener = new ChannelFutureListener() {
@@ -190,6 +202,7 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
                     };
                     promise = promise.unvoid().addListener(closeStreamLocalListener);
                 }
+
                 ChannelFuture future = frameWriter.writeHeaders(ctx, streamId, headers, streamDependency,
                                                                 weight, exclusive, padding, endOfStream, promise);
                 // Writing headers may fail during the encode state if they violate HPACK limits.
@@ -197,7 +210,7 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
                 if (failureCause == null) {
                     // Synchronously set the headersSent flag to ensure that we do not subsequently write
                     // other headers containing pseudo-header fields.
-                    stream.headersSent();
+                    stream.headersSent(isInformational);
                 } else {
                     lifecycleManager.onError(ctx, failureCause);
                 }
@@ -451,6 +464,7 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
 
         @Override
         public void write(ChannelHandlerContext ctx, int allowedBytes) {
+            boolean isInformational = validateHeadersSentState(stream, headers, connection.isServer(), endOfStream);
             if (promise.isVoid()) {
                 promise = ctx.newPromise();
             }
@@ -461,7 +475,7 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
             // Writing headers may fail during the encode state if they violate HPACK limits.
             Throwable failureCause = f.cause();
             if (failureCause == null) {
-                stream.headersSent();
+                stream.headersSent(isInformational);
             } else {
                 lifecycleManager.onError(ctx, failureCause);
             }
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Stream.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Stream.java
@@ -130,14 +130,40 @@ public interface Http2Stream {
     <V> V removeProperty(Http2Connection.PropertyKey key);
 
     /**
-     * Indicates that headers has been sent to the remote on this stream.
+     * Indicates that headers have been sent to the remote endpoint on this stream. The first call to this method would
+     * be for the initial headers (see {@link #isHeadersSent()}} and the second call would indicate the trailers
+     * (see {@link #isTrailersReceived()}).
+     * @param isInformational {@code true} if the headers contain an informational status code (for responses only).
      */
-    Http2Stream headersSent();
+    Http2Stream headersSent(boolean isInformational);
 
     /**
-     * Indicates whether or not headers was sent to the remote endpoint.
+     * Indicates whether or not headers were sent to the remote endpoint.
      */
     boolean isHeadersSent();
+
+    /**
+     * Indicates whether or not trailers were sent to the remote endpoint.
+     */
+    boolean isTrailersSent();
+
+    /**
+     * Indicates that headers have been received. The first call to this method would be for the initial headers
+     * (see {@link #isHeadersReceived()}} and the second call would indicate the trailers
+     * (see {@link #isTrailersReceived()}).
+     * @param isInformational {@code true} if the headers contain an informational status code (for responses only).
+     */
+    Http2Stream headersReceived(boolean isInformational);
+
+    /**
+     * Indicates whether or not the initial headers have been received.
+     */
+    boolean isHeadersReceived();
+
+    /**
+     * Indicates whether or not the trailers have been received.
+     */
+    boolean isTrailersReceived();
 
     /**
      * Indicates that a push promise was sent to the remote endpoint.
--- a/codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoderTest.java
+++ b/codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoderTest.java
@@ -24,6 +24,7 @@ import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelPipeline;
 import io.netty.channel.ChannelPromise;
 import io.netty.channel.DefaultChannelPromise;
+import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http2.Http2RemoteFlowController.FlowControlled;
 import io.netty.util.concurrent.ImmediateEventExecutor;
 import junit.framework.AssertionFailedError;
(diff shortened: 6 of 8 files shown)


Scottmitch:
@nmittler @carl-mastrangelo @ejona86 - FYI

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
```diff
@@ -398,35 +401,56 @@ public final State state() {
...
+            return (metaState & META_STATE_SENT_HEADERS) != 0;
+        }
+
+        @Override
+        public boolean isTrailersSent() {
+            return (metaState & META_STATE_SENT_TRAILERS) != 0;
+        }
+
+        @Override
+        public Http2Stream headersReceived() {
```
Scottmitch: I could also modify this API to include a parameter `boolean endOfStream` and move the validation `if (stream.isTrailersReceived() || stream.isHeadersReceived() && !endOfStream) { throw } ` inside this method.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -282,6 +282,12 @@ public void onHeadersRead(ChannelHandlerContext ctx, int streamId, Http2Headers
                 return;
             }
 
+            if (stream.isTrailersReceived() || stream.isHeadersReceived() && !endOfStream) {
```
netkins: BLOCKER NullPointerException might be thrown as 'stream' is nullable here [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2259)

Scottmitch: I don't think this is an issue ... `shouldIgnoreHeadersOrDataFrame` won't let a `null` stream slip by here.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -161,6 +161,8 @@ public ChannelFuture writeHeaders(final ChannelHandlerContext ctx, final int str
                     }
                     throw cause;
                 }
+            } else if (stream.isTrailersSent() || stream.isHeadersSent() && !endOfStream) {
+                throw new IllegalStateException("Stream " + stream.id() + " sent too many headers EOS: " + endOfStream);
```
carl-mastrangelo: dumb question:  why is the promise not failed here, like on line 159?

Scottmitch: It is failed in the `catch` block below. The only exception to this control flow is when the stream no longer exists ... we catch early and prevent `lifecycleManager.onError` from being called.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
```diff
@@ -398,35 +401,56 @@ public final State state() {
...
 
         @Override
         public Http2Stream resetSent() {
-            sentState |= SENT_STATE_RST;
+            metaState |= META_STATE_SENT_RST;
             return this;
         }
 
         @Override
         public Http2Stream headersSent() {
```
ejona86: I don't see any use of this method.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -305,6 +311,7 @@ public void onHeadersRead(ChannelHandlerContext ctx, int streamId, Http2Headers
                             stream.state());
             }
 
+            stream.headersReceived();
```
ejona86: This doesn't seem to consider 1xx response headers.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
normanmaurer: +1

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -456,6 +458,12 @@ public void write(ChannelHandlerContext ctx, int allowedBytes) {
             }
             promise.addListener(this);
 
+            if (stream.isTrailersSent() || stream.isHeadersSent() && !endOfStream) {
+                promise.tryFailure(new IllegalStateException("Stream " + stream.id() + " sent too many headers EOS: " +
```
normanmaurer: just a nit... `headers, EOS:`

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoderTest.java
```diff
@@ -343,10 +345,107 @@ public void headersWriteShouldOpenStreamForPush() throws Exception {
     }
 
     @Test
+    public void trailersDoNotEndStreamThrows() {
+        writeAllFlowControlledFrames();
+        final int streamId = 6;
+        ChannelPromise promise = newPromise();
```
normanmaurer: should we assert the state of the promise after the write call ?

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoderTest.java
```diff
@@ -343,10 +345,107 @@ public void headersWriteShouldOpenStreamForPush() throws Exception {
...
+
+        verify(writer, times(1)).writeHeaders(eq(ctx), eq(streamId), eq(EmptyHttp2Headers.INSTANCE), eq(0),
+                eq(DEFAULT_PRIORITY_WEIGHT), eq(false), eq(0), eq(false), eq(promise));
+    }
+
+    @Test
+    public void trailersDoNotEndStreamWithDataThrows() {
+        writeAllFlowControlledFrames();
+        final int streamId = 6;
+        ChannelPromise promise = newPromise();
```
normanmaurer: should we assert the state of the promise after the write call ?

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoderTest.java
```diff
@@ -343,10 +345,107 @@ public void headersWriteShouldOpenStreamForPush() throws Exception {
...
+
+    @Test
+    public void tooManyHeadersEOSThrows() {
+        tooManyHeadersThrows(true);
+    }
+
+    private void tooManyHeadersThrows(boolean eos) {
+        writeAllFlowControlledFrames();
+        final int streamId = 6;
+        ChannelPromise promise = newPromise();
```
normanmaurer: should we assert the state of the promise after the write call ?

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoderTest.java
```diff
@@ -343,10 +345,107 @@ public void headersWriteShouldOpenStreamForPush() throws Exception {
...
+        encoder.writeHeaders(ctx, streamId, EmptyHttp2Headers.INSTANCE, 0, false, promise);
+        ChannelPromise promise2 = newPromise();
```
normanmaurer: should we assert the state of the promise after the write call ?

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoderTest.java
```diff
@@ -343,10 +345,107 @@ public void headersWriteShouldOpenStreamForPush() throws Exception {
...
+
+    @Test
+    public void tooManyHeadersWithDataEOSThrows() {
+        tooManyHeadersWithDataThrows(true);
+    }
+
+    private void tooManyHeadersWithDataThrows(boolean eos) {
+        writeAllFlowControlledFrames();
+        final int streamId = 6;
+        ChannelPromise promise = newPromise();
```
normanmaurer: should we assert the state of the promise after the write call ?

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoderTest.java
```diff
@@ -343,10 +345,107 @@ public void headersWriteShouldOpenStreamForPush() throws Exception {
...
+        encoder.writeHeaders(ctx, streamId, EmptyHttp2Headers.INSTANCE, 0, false, promise);
+
+        Http2Stream stream = connection.stream(streamId);
+        when(remoteFlow.hasFlowControlled(eq(stream))).thenReturn(true);
+
+        ChannelPromise promise2 = newPromise();
```
normanmaurer: should we assert the state of the promise after the write call ?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
Scottmitch: It is used in in a few places [1][2].

[1] https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java#L200
[2] https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java#L464

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
Scottmitch: +1. I will update the PR to account for this on the send/receive side.

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoderTest.java
Scottmitch: I don't think it is necessary because we are mocking out the underlying transport ... we mostly care when things have been failed (because it is done at the encoder level).

Scottmitch:
@ejona86 - The PR has been corrected to support informational headers flow. PTAL.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Stream.java
```diff
@@ -130,16 +130,53 @@ public boolean remoteSideOpen() {
...
     /**
+     * Indicates whether or not trailers were sent to the remote endpoint.
+     */
+    boolean isTrailersSent();
+
+    /**
+     * Indicates that headers have been received. The first call to this method would be for the initial headers
+     * (see {@link #isHeadersReceived()}} and the second call would indicate the trailers
+     * (see {@link #isTrailersReceived()}).
+     * @param {@code true} if the headers contain an informational status code (for responses only).
```
ejona86: It looks like `@param` is missing the argument name.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -282,6 +285,16 @@ public void onHeadersRead(ChannelHandlerContext ctx, int streamId, Http2Headers
                 return;
             }
 
+            boolean isInformational = false;
+            if ((!connection.isServer() &&
+                    (isInformational = HttpStatusClass.valueOf(headers.status()) == INFORMATIONAL)
+                    && stream.isInformationalHeadersReceived()) ||
```
ejona86: This looks like it only permits the first informational header. You can actually have an unlimited number of them before the first "real" header. Ditto for sending.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
ejona86: Doh. Sorry. I mistook headersSent() it for a new method.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Stream.java
```diff
@@ -130,16 +130,53 @@ public boolean remoteSideOpen() {
     <V> V removeProperty(Http2Connection.PropertyKey key);
 
     /**
-     * Indicates that headers has been sent to the remote on this stream.
+     * Indicates that headers have been sent to the remote endpoint on this stream. The first call to this method would
+     * be for the initial headers (see {@link #isHeadersSent()}} and the second call would indicate the trailers
+     * (see {@link #isTrailersReceived()}).
+     * @param {@code true} if the headers contain an informational status code (for responses only).
```
ejona86: Missing argument name after `@param`

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -145,6 +147,17 @@ public ChannelFuture writeHeaders(ChannelHandlerContext ctx, int streamId, Http2
         return writeHeaders(ctx, streamId, headers, 0, DEFAULT_PRIORITY_WEIGHT, false, padding, endStream, promise);
     }
 
+    private static boolean validateHeadersSentState(Http2Stream stream, Http2Headers headers, boolean isServer,
+                                                    boolean endOfStream) {
+        boolean isInformational = false;
+        if ((isServer && (isInformational = HttpStatusClass.valueOf(headers.status()) == INFORMATIONAL)
```
netkins: MAJOR Extract the assignment out of this expression. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AAssignmentInSubExpressionCheck)

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -282,6 +285,16 @@ public void onHeadersRead(ChannelHandlerContext ctx, int streamId, Http2Headers
                 return;
             }
 
+            boolean isInformational = false;
+            if ((!connection.isServer() &&
+                    (isInformational = HttpStatusClass.valueOf(headers.status()) == INFORMATIONAL)
```
netkins: MAJOR Extract the assignment out of this expression. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AAssignmentInSubExpressionCheck)

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
netkins: BLOCKER NullPointerException might be thrown as 'stream' is nullable here [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2259)

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -16,13 +16,16 @@
 
 import io.netty.buffer.ByteBuf;
 import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.http.HttpResponseStatus;
```
netkins: MINOR Remove this unused import 'io.netty.handler.codec.http.HttpResponseStatus'. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -282,6 +285,16 @@ public void onHeadersRead(ChannelHandlerContext ctx, int streamId, Http2Headers
                 return;
             }
 
+            boolean isInformational = false;
+            if ((!connection.isServer() &&
+                    (isInformational = HttpStatusClass.valueOf(headers.status()) == INFORMATIONAL)
+                    && stream.isInformationalHeadersReceived()) ||
+                stream.isTrailersReceived() || stream.isHeadersReceived() && !endOfStream) {
```
netkins: BLOCKER NullPointerException might be thrown as 'stream' is nullable here [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2259)

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
Scottmitch: Yes you are right ... I wasn't aware that infinite could be received/sent. Thanks for pointing this out.

Are you aware of any practical use cases for more than 1 info response?

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoderTest.java
normanmaurer: ok fair enough

Scottmitch:
@ejona86 - Updated to support multiple info headers ... PTAL.

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoderTest.java
```diff
@@ -343,10 +345,215 @@ public void headersWriteShouldOpenStreamForPush() throws Exception {
...
+        encoder.writeHeaders(ctx, streamId, EmptyHttp2Headers.INSTANCE, 0, false, promise2);
+
+        ChannelPromise promise3 = newPromise();
+        ChannelFuture future = encoder.writeHeaders(ctx, streamId, EmptyHttp2Headers.INSTANCE, 0, eos, promise3);
+        assertTrue(future.isDone());
+        if (eos) {
+            assertTrue(future.isSuccess());
+        } else {
+            assertFalse(future.isSuccess());
```
normanmaurer: just a nit but you can simplify this:

```java
assertEquals(eos, future.isSuccess());
```

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoderTest.java
```diff
@@ -343,10 +345,215 @@ public void headersWriteShouldOpenStreamForPush() throws Exception {
...
+        }
```
normanmaurer: You can simplify:

```
assertEquals(eos, future.isSuccess());
```

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -282,6 +284,16 @@ public void onHeadersRead(ChannelHandlerContext ctx, int streamId, Http2Headers
                 return;
             }
 
+            boolean isInformational = false;
+            if ((!connection.isServer() &&
+                    (isInformational = HttpStatusClass.valueOf(headers.status()) == INFORMATIONAL)
+                    && stream.isHeadersReceived()) ||
```
netkins: BLOCKER NullPointerException might be thrown as 'stream' is nullable here [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2259)

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -282,6 +284,16 @@ public void onHeadersRead(ChannelHandlerContext ctx, int streamId, Http2Headers
...
+                stream.isTrailersReceived() || stream.isHeadersReceived() && !endOfStream) {
```
netkins: BLOCKER NullPointerException might be thrown as 'stream' is nullable here [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2259)

codec-http/src/main/java/io/netty/handler/codec/http/HttpStatusClass.java
```diff
@@ -74,6 +74,29 @@ public static HttpStatusClass valueOf(int code) {
...
+     * Returns the class of the specified HTTP status code.
+     * @param code Just the numeric portion of the http status code.
+     */
+    public static HttpStatusClass valueOf(CharSequence code) {
+        if (code != null && code.length() == 3) {
+            char c0 = code.charAt(0);
+            char c1 = code.charAt(1);
+            char c2 = code.charAt(2);
+            return isDigit(c0) && isDigit(c1) && isDigit(c2) ?
+                    valueOf(digit(c0) * 100 + digit(c1) * 10 + digit(c2)) : UNKNOWN;
```
fenik17: To determine the class, it will be enough: ```valueOf((c0 - '0') * 100)```.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
fenik17: +1. It can be looks like this:
```java
boolean isInformational = isServer && HttpStatusClass.valueOf(headers.status()) == INFORMATIONAL;
if ((isInformational || !endOfStream) && stream.isHeadersSent() || stream.isTrailersSent()) {
    ...
}
```

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
fenik17: ditto as suggested below.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
Scottmitch: sgtm

codec-http/src/main/java/io/netty/handler/codec/http/HttpStatusClass.java
Scottmitch: I was trying to balance how much was determined in this method vs delegating to `valueOf(int)`. However since both are in the scope of this `Enum` I guess I can be more aggressive here and duplication of logic won't be too bad.

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoderTest.java
Scottmitch: done

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoderTest.java
Scottmitch: done

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -282,6 +284,14 @@ public void onHeadersRead(ChannelHandlerContext ctx, int streamId, Http2Headers
                 return;
             }
 
+            boolean isInformational = !connection.isServer() &&
+                    HttpStatusClass.valueOf(headers.status()) == INFORMATIONAL;
+            if ((isInformational || !endOfStream) && stream.isHeadersReceived() || stream.isTrailersReceived()) {
```
netkins: BLOCKER NullPointerException might be thrown as 'stream' is nullable here [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2259)
BLOCKER NullPointerException might be thrown as 'stream' is nullable here [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2259)

codec-http/src/main/java/io/netty/handler/codec/http/HttpStatusClass.java
fenik17: We can even delete the check `isDigit(c0)` :)

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -282,6 +284,14 @@ public void onHeadersRead(ChannelHandlerContext ctx, int streamId, Http2Headers
...
+                throw streamError(stream.id(), PROTOCOL_ERROR,
```
fenik17: `stream.id()` -> `streamId`

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
```diff
@@ -373,13 +373,16 @@ final DefaultPropertyKey verifyKey(PropertyKey key) {
...
     private class DefaultStream implements Http2Stream {
-        private static final byte SENT_STATE_RST = 0x1;
-        private static final byte SENT_STATE_HEADERS = 0x2;
-        private static final byte SENT_STATE_PUSHPROMISE = 0x4;
+        private static final byte META_STATE_SENT_RST = 1;
+        private static final byte META_STATE_SENT_HEADERS = 1 << 1;
+        private static final byte META_STATE_SENT_TRAILERS = 1 << 2;
+        private static final byte META_STATE_SENT_PUSHPROMISE = 1 << 3;
+        private static final byte META_STATE_RECV_HEADERS = 1 << 4;
+        private static final byte META_STATE_RECV_TRAILERS = 1 << 5;
```
carl-mastrangelo: Optional: you might consider using EnumSet for this.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -145,6 +147,15 @@ public ChannelFuture writeHeaders(ChannelHandlerContext ctx, int streamId, Http2
         return writeHeaders(ctx, streamId, headers, 0, DEFAULT_PRIORITY_WEIGHT, false, padding, endStream, promise);
     }
 
+    private static boolean validateHeadersSentState(Http2Stream stream, Http2Headers headers, boolean isServer,
+                                                    boolean endOfStream) {
+        boolean isInformational = isServer && HttpStatusClass.valueOf(headers.status()) == INFORMATIONAL;
+        if ((isInformational || !endOfStream) && stream.isHeadersSent() || stream.isTrailersSent()) {
+            throw new IllegalStateException("Stream " + stream.id() + " sent too many headers EOS: " + endOfStream);
```
carl-mastrangelo: nit: IllegalStateException is for calling Java methods out of order, not for network related errors.

From an API standpoint, I would expect a method named "validate..."  to also have a throws clause.

Scottmitch: > IllegalStateException ...

Lets consider this for a followup PR. This PR is just being consistent with the existing approach.

> From an API standpoint

IIUC is generally not common practice to include `throws` clauses for runtime exceptions. Can you can submit a followup PR to make the changes you have in mind to clarify?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
Scottmitch: I will just stick with manual bit masking for now ... less object overhead and self contained.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
Scottmitch: sure

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
carl-mastrangelo: > IIUC is generally not common practice to include throws clauses for runtime exceptions. Can you can submit a followup PR to make the changes you have in mind to clarify?

I am coming from a Guava mindset, which has a special exception for "The remote side messed up"  called VerificationException.  I think you don't depend on Guava, but it's a useful idea.  If there isn't any recovery path from such an exception, why not just make it RuntimeException, or IORuntimeException, or some such?

Followup PR for ISE is fine.  ISE is really meant for thinks like calling `read()` after `close()`.  It's a programmer error.

Scottmitch: @carl-mastrangelo - Yip I'm not against having a special exceptions type for these types of things. Lets do it in a followup PR so it can be applied consistently and the change is isolated.

Scottmitch:
4.1 (a91df58ca)

Scottmitch:
@ejona86 - if there are any remaining issues I can submit a followup PR.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
ejona86: _Practical_! Bah! Such high expectations! :smile: I'll say in _practice_ they're pretty unheard of. But as far as semi-practical use-cases:

1. Multiple 1xx status codes. You con conceive of another 1xx status code you may want to mix with 100... say [WebDav's 102 Processing](https://tools.ietf.org/html/rfc2518#section-10.1). You send 100 to tell the client you have resources available, send 102 to say you have received the request and are working on the response, and then 200 when you done.

2. Sending 102 multiple times looks useful. It's supposed to tell the client not to time out yet. If it took a _long_ time to process you could understand sending that multiple times. Granted, in that case you may want to rearchitect. But it would be practical...

3. I had thought _someone_ would be using 100 Continue multiple times as a sort of checkpointing system. It turns out [103 Checkpoint](https://web.archive.org/web/20151013212135/http://code.google.com/p/gears/wiki/ResumableHttpRequestsProposal) is a quasi-thing. [There](http://advanced-servlets.yahoogroups.narkive.com/9TbXg1mU/sending-multiple-100-continue-responses-before-a-200-ok) [is](https://stackoverflow.com/questions/22818059/several-100-continue-received-from-the-server) [evidence](https://github.com/php-mod/curl/issues/19) that _somebody_ is sending 100 multiple times, but none of them explain why. And it definitely seems rare.

ejona86:
LGTM

PR #6964, opened 2017-07-12
OpenSslEngine protocol selection must be contiguous

Scottmitch:
Motivation:
TLS doesn't support a way to advertise non-contiguous versions from the client's perspective, and the client just advertises the max supported version. The TLS protocol also doesn't support all different combinations of discrete protocols, and instead assumes contiguous ranges. OpenSSL has some unexpected behavior (e.g. handshake failures) if non-contiguous protocols are used even where there is a compatible set of protocols and ciphers. For these reasons this method will determine the minimum protocol and the maximum protocol and enabled a contiguous range from [min protocol, max protocol] in OpenSSL.

Modifications:
- ReferenceCountedOpenSslEngine#setEnabledProtocols should determine the min/max protocol versions and enable a contiguous range

Result:
OpenSslEngine is more consistent with the JDK's SslEngineImpl and no more unexpected handshake failures due to protocol selection quirks.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -93,6 +93,18 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
             new SSLException("renegotiation unsupported"), ReferenceCountedOpenSslEngine.class, "beginHandshake()");
     private static final ResourceLeakDetector<ReferenceCountedOpenSslEngine> leakDetector =
             ResourceLeakDetectorFactory.instance().newResourceLeakDetector(ReferenceCountedOpenSslEngine.class);
+    private static final int OPENSSL_OP_NO_PROTOCOL_INDEX_SSLV2 = 0;
+    private static final int OPENSSL_OP_NO_PROTOCOL_INDEX_SSLV3 = 1;
+    private static final int OPENSSL_OP_NO_PROTOCOL_INDEX_TLSv1 = 2;
+    private static final int OPENSSL_OP_NO_PROTOCOL_INDEX_TLSv1_1 = 3;
+    private static final int OPENSSL_OP_NO_PROTOCOL_INDEX_TLSv1_2 = 4;
+    private static final int[] OPENSSL_OP_NO_PROTOCOLS = new int[] {
+            SSL.SSL_OP_NO_SSLv2,
+            SSL.SSL_OP_NO_SSLv3,
+            SSL.SSL_OP_NO_TLSv1,
+            SSL.SSL_OP_NO_TLSv1_1,
+            SSL.SSL_OP_NO_TLSv1_2
+    };
     /**
      * <a href="https://www.openssl.org/docs/man1.0.2/crypto/X509_check_host.html">The flags argument is usually 0</a>.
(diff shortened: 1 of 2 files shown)


Scottmitch:
Thanks to @davidben for explaining the current state of affairs in OpenSSL/BoringSSL and limitations of the TLS protocol w.r.t to non-contiguous versions.

Scottmitch:
This PR is dependent upon https://github.com/netty/netty/pull/6965

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1363,54 +1375,77 @@ private static boolean isProtocolEnabled(int opts, int disableMask, String proto
...
-                    opts |= SSL.SSL_OP_NO_TLSv1;
-                }
-                if (!tlsv1_1) {
-                    opts |= SSL.SSL_OP_NO_TLSv1_1;
+                for (int i = 0; i < minProtocolIndex; ++i) {
+                    opts |= OPENSSL_OP_NO_PROTOCOLS[i];
                 }
-                if (!tlsv1_2) {
-                    opts |= SSL.SSL_OP_NO_TLSv1_2;
+                assert maxProtocolIndex != Integer.MAX_VALUE;
```
netkins: MAJOR Unnecessary use of fully qualified name 'Integer.MAX_VALUE' due to existing static import 'java.lang.Integer.MAX_VALUE' [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUnnecessaryFullyQualifiedName)

handler/src/main/java/io/netty/handler/ssl/SslUtils.java
```diff
@@ -71,6 +76,47 @@
      */
     static final int NOT_ENCRYPTED = -2;
 
+    static final String[] DEFAULT_CIPHER_SUITES = new String[] {
```

Scottmitch:
4.1 (615299007) 4.0 (5fd6c9d5d)

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1363,54 +1375,77 @@ private static boolean isProtocolEnabled(int opts, int disableMask, String proto
...
-                    opts |= SSL.SSL_OP_NO_TLSv1;
-                }
-                if (!tlsv1_1) {
-                    opts |= SSL.SSL_OP_NO_TLSv1_1;
+                for (int i = 0; i < minProtocolIndex; ++i) {
+                    opts |= OPENSSL_OP_NO_PROTOCOLS[i];
                 }
-                if (!tlsv1_2) {
-                    opts |= SSL.SSL_OP_NO_TLSv1_2;
+                assert maxProtocolIndex != MAX_VALUE;
```
johnou: Max protocol index is never assigned max value, maybe you intended to assert minProtocolIndex != OPENSSL_OP_NO_PROTOCOLS.length?

Scottmitch: Nah this to clarify the assumption because otherwise the next line will have a bug due to overflow `for (int i = maxProtocolIndex + 1; i < OPENSSL_OP_NO_PROTOCOLS.length; ++i)`

PR #6965, opened 2017-07-12
Unify default cipher suites betweek JDK and OpenSSL

Scottmitch:
Motivation:
Currently the default cipher suites are set independently between JDK and OpenSSL. We should use a common approach to setting the default ciphers. Also the OpenSsl default ciphers are expressed in terms of the OpenSSL cipher name conventions, which is not correct and may be exposed to the end user. OpenSSL should also use the RFC cipher names like the JDK defaults.

Modifications:
- Move the default cipher definition to a common location and use it in JDK and OpenSSL initialization
- OpenSSL should not expose OpenSSL cipher names externally

Result:
Common initialization and OpenSSL doesn't expose custom cipher names.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslUtils.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslUtils.java
@@ -24,7 +24,12 @@ import io.netty.handler.codec.base64.Base64Dialect;
 
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
+import java.util.List;
+import java.util.Set;
+
 import javax.net.ssl.SSLHandshakeException;
+
+import static java.util.Arrays.asList;
 
 /**
  * Constants for SSL packets.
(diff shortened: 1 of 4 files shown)


Scottmitch:
This was discovered while developing PR https://github.com/netty/netty/pull/6964

handler/src/main/java/io/netty/handler/ssl/SslUtils.java
```diff
@@ -71,6 +74,55 @@
      */
     static final int NOT_ENCRYPTED = -2;
 
+    static final String[] DEFAULT_CIPHER_SUITES = new String[] {
+        // XXX: Make sure to sync this list with OpenSslEngineFactory.
```
johnou: @Scottmitch we can drop this comment now right? :)

Scottmitch: +1 copy/pasta fail

handler/src/main/java/io/netty/handler/ssl/SslUtils.java
```diff
@@ -71,6 +74,54 @@
...
+            if (supported.contains(n)) {
+                enabled.add(n);
+            }
+        }
+    }
+
+    static void useFallbackCiphersIfDefaultIsEmpty(List<String> defaultCiphers, Iterable<String> fallbackCiphers) {
+        if (defaultCiphers.isEmpty()) {
+            for (String cipher : fallbackCiphers) {
+                if (cipher.startsWith("SSL_") || cipher.contains("_RC4_")) {
```
johnou: Why SSL_ here? I only see _RC4_ in jdk. Could this break IBM j9?

handler/src/main/java/io/netty/handler/ssl/SslUtils.java
```diff
@@ -71,6 +74,54 @@
...
+                    continue;
+                }
+                defaultCiphers.add(cipher);
+            }
+        }
+    }
+
+    static void useFallbackCiphersIfDefaultIsEmpty(List<String> defaultCiphers, String[] fallbackCiphers) {
```
johnou: Instead of maintaining two identical methods why not just wrap the string primitive array with Arrays.asList?

handler/src/main/java/io/netty/handler/ssl/SslUtils.java
Scottmitch: `SSL_` cipher may be returned by the JDK in `getSupportedCiphers` however the JDK should accept the `TLS_` ciphers. The reason we exclude `SSL_` is to prevent duplicate entries, and some JDKs may not support this custom naming scheme (but all should support the `TLS_` naming scheme).

There may be some older cipher names in the JSSE which have the `SSL_` prefix (established before the `TLS_` ... e.g. `SSL_RSA_WITH_RC4_128_SHA`) however most of these ciphers are old/obsolete and I don't think we want to enable them by default anyways. If someone wants these older ciphers they should explicitly enable them instead of relying on the default (which in this case would mean none of the actual `SslUtils#DEFAULT_CIPHER_SUITES` ciphers are supported).

handler/src/main/java/io/netty/handler/ssl/SslUtils.java
Scottmitch: Sure I can do this ... the original motivation was to avoid the intermediate allocation/iteration/copy ... however this method is called infrequently so it shouldn't matter.

netkins: MINOR Consider using varargs for methods or constructors which take an array the last parameter. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUseVarargs)

handler/src/main/java/io/netty/handler/ssl/SslUtils.java
```diff
@@ -71,6 +74,54 @@
      */
     static final int NOT_ENCRYPTED = -2;
 
+    static final String[] DEFAULT_CIPHER_SUITES = new String[] {
```

handler/src/main/java/io/netty/handler/ssl/SslUtils.java
johnou: Yep makes sense, only saw it used in static blocks but I don't have access to a full source code checkout atm.

handler/src/main/java/io/netty/handler/ssl/SslUtils.java
johnou: Thanks for the explanation, was wondering if it's worth adding debug logging explaining they are being excluded but given the logging from the other classes and that it is a fall back is probably fine.

Scottmitch:
4.1 (43ae9748d) 4.0 (f57d8a150)

Issue #6966, opened 2017-07-12
Netty DNS Tests fail with custom search domain


carl-mastrangelo:
### Expected behavior
tests pass
### Actual behavior
tests fail
### Steps to reproduce
Add this to `/etc/resolv.conf`:

```
search floop
```

and run tests

### Netty version
4.1

output:

```
Tests run: 28, Failures: 0, Errors: 11, Skipped: 0, Time elapsed: 7.491 sec <<< FAILURE! - in io.netty.resolver.dns.DnsNameResolverTest
secondDnsServerShouldBeUsedBeforeCNAMEFirstServerNotStarted(io.netty.resolver.dns.DnsNameResolverTest)  Time elapsed: 1.855 sec  <<< ERROR!
io.netty.resolver.dns.DnsNameResolverContext$SearchDomainUnknownHostException: Search domain query failed. Original hostname: 'netty.io' failed to resolve 'netty.io.floop' after 4 queries 

testResolveNotDecodeUnicode(io.netty.resolver.dns.DnsNameResolverTest)  Time elapsed: 0.218 sec  <<< ERROR!
io.netty.resolver.dns.DnsNameResolverContext$SearchDomainUnknownHostException: Search domain query failed. Original hostname: 'xn--mller-kva.de' failed to resolve 'xn--mller-kva.de.floop'

testNonCachedResolve(io.netty.resolver.dns.DnsNameResolverTest)  Time elapsed: 0.144 sec  <<< ERROR!
io.netty.resolver.dns.DnsNameResolverContext$SearchDomainUnknownHostException: Search domain query failed. Original hostname: 'blogspot.in' failed to resolve 'blogspot.in.floop'

testResolveAAAAorA(io.netty.resolver.dns.DnsNameResolverTest)  Time elapsed: 0.808 sec  <<< ERROR!
io.netty.resolver.dns.DnsNameResolverContext$SearchDomainUnknownHostException: Search domain query failed. Original hostname: 'blogspot.in' failed to resolve 'blogspot.in.floop'

aAndAAAAQueryShouldTryFirstDnsServerBeforeSecond(io.netty.resolver.dns.DnsNameResolverTest)  Time elapsed: 0.197 sec  <<< ERROR!
io.netty.resolver.dns.DnsNameResolverContext$SearchDomainUnknownHostException: Search domain query failed. Original hostname: 'netty.io' failed to resolve 'netty.io.floop' after 5 queries 

[...]

PR #6967, opened 2017-07-12
DNS Resovler tests should be more explicit about ndots

Scottmitch:
Motivation:
The DNS resolver may use default configuration inherited from the environment. This means the ndots value may change and result in test failure if the tests don't explicitly set the assumed value.

Modifications:
- Explicitly set ndots in resolver-dns unit tests so we don't fail if the environment overrides the search domain and ndots

Result:
Unit tests are less dependent upon the enviroment they run in.
Fixes https://github.com/netty/netty/issues/6966.


Scottmitch:
@carl-mastrangelo - PTAL

Scottmitch:
> As in, if I tried using DnsNameResolver do I have to set ndots there too?

If you don't set values programmatically we attempt to inherit them from the environment. If you need to be explicit for a particular use case (e.g. ignore the environment) then you should programmatically set the desired values.

Scottmitch:
4.1 (b249714a2)

PR #6972, opened 2017-07-16
Record release when enable detailed leak detection

louxiu:
Motivation:
It would be easier to find where is missing release call in several retain release calls on a ByteBuf

Modifications:
Remove final modifier on SimpleLeakAwareByteBuf release function and override it to record release in AdvancedLeakAwareByteBuf and AdvancedLeakAwareCompositeByteBuf

Result:
Release will be recorded when enable detailed leak detection


normanmaurer:
@trustin @Scottmitch WDYT... I think this is not a bad idea...

normanmaurer:
Cherry-picked into 4.1 (0ad99310f5c02363f9402872d9b3d803654a794d) and 4.0 (c4213bc4de3ba0141d6535c6ffb95c87fbb837e4)

PR #6973, opened 2017-07-16
Explicit specify hostaddress during tests to ensure testsuite pass on…

normanmaurer:
… docker (mac)

Motivation:

When run the current testsuite on docker (mac) it will fail a few tests with:

io.netty.channel.AbstractChannel$AnnotatedConnectException: connect(..) failed: Cannot assign requested address: /0:0:0:0:0:0:0:0%0:46607
Caused by: java.net.ConnectException: connect(..) failed: Cannot assign requested address

Modifications:

Specify host explicit as done in other tests to only use ipv6 when really supported.

Result:

Build pass on docker as well

### DIFF
--- a/testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketConnectTest.java
+++ b/testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketConnectTest.java
@@ -21,6 +21,8 @@ import io.netty.channel.Channel;
 import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.testsuite.util.TestUtils;
+import io.netty.util.NetUtil;
 import io.netty.util.concurrent.ImmediateEventExecutor;
 import io.netty.util.concurrent.Promise;
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


normanmaurer:
Discovered this while creating a docker file to make it easier to setup build env for netty.

normanmaurer:
As these are just test-fixes I merged it into 4.1 (e6a399a7785b41dc941e8037257a30a0df1d92d4) and 4.0 (9eeb86db4c1a1d83d1648f97244ce320f1cbc0ef)

testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketConnectTest.java
```diff
@@ -82,7 +84,7 @@ public void testChannelEventsFiredWhenClosedDirectly(ServerBootstrap sb, Bootstr
         Channel cc = null;
         try {
             sb.childHandler(new ChannelInboundHandlerAdapter());
-            sc = sb.bind(0).syncUninterruptibly().channel();
+            sc = sb.bind(NetUtil.LOCALHOST, TestUtils.getFreePort()).syncUninterruptibly().channel();
```
Scottmitch: IIRC `TestUtils.getFreePort()` has been observed to fail when test are run in quick succession due to bind failures (address already in use). Can we just use `new InetSocketAddress(0)` which will pick an ephemeral port instead?

normanmaurer: Sure we can use 0 as port... let me update this in all tests then ....

normanmaurer: See https://github.com/netty/netty/pull/6986

PR #6974, opened 2017-07-17
Calculate correct lastRecords size

louxiu:
Motivation:
ResourceLeakDetector records at most MAX_RECORDS+1 records

Modifications:
Increase records size after add to lastRecords

Result:
ResourceLeakDetector will record at most MAX_RECORDS records

### DIFF
--- a/common/src/main/java/io/netty/util/ResourceLeakDetector.java
+++ b/common/src/main/java/io/netty/util/ResourceLeakDetector.java
@@ -364,17 +364,18 @@ public class ResourceLeakDetector<T> {
         }
 
         private void record0(Object hint, int recordsToSkip) {
-            if (creationRecord != null) {
+            // Check MAX_RECORDS > 0 here to avoid similar check before remove from and add to lastRecords
+            if (creationRecord != null && MAX_RECORDS > 0) {
                 String value = newRecord(hint, recordsToSkip);
 
                 synchronized (lastRecords) {
                     int size = lastRecords.size();
                     if (size == 0 || !lastRecords.getLast().equals(value)) {
+                        if (size >= MAX_RECORDS) {
+                            lastRecords.removeFirst();
+                            ++removedRecords;
+                        }
                         lastRecords.add(value);
-                    }
-                    if (size > MAX_RECORDS) {
-                        lastRecords.removeFirst();
-                        ++removedRecords;
                     }
                 }
             }


normanmaurer:
LGTM... @Scottmitch @trustin PTAL as well

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -371,6 +371,7 @@ private void record0(Object hint, int recordsToSkip) {
                     int size = lastRecords.size();
                     if (size == 0 || !lastRecords.getLast().equals(value)) {
                         lastRecords.add(value);
+                        size += 1;
```
Scottmitch: please use the increment operator `++size` if incrementing by 1. Also we should remove from `lastRecords` before adding to avoid re-sizing larger than the max size (which we will never fill).

normanmaurer: @Scottmitch good idea!

louxiu: Done

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -364,18 +364,18 @@ public void record(Object hint) {
         }
 
         private void record0(Object hint, int recordsToSkip) {
-            if (creationRecord != null) {
+            if (creationRecord != null && MAX_RECORDS > 0) {
```
normanmaurer: @louxiu I think this is not correct as this change behaviour. Before the user could use MAX_RECORDS = 0 to basically enforce there is no limit at all.  Now we not enter the block all. Can you explain why you think this is correct here ?

louxiu: 1. If `MAX_RECORDS == 0` and `lastRecords.size() == 0`, then `size >= MAX_RECORDS` is true, `lastRecords.removeFirst()` will raise error. Even if we check `size > 0 && size >= MAX_RECORDS` before remove, there will always be one record added to lastRecords.

2. I don't get this "Before the user could use MAX_RECORDS = 0 to basically enforce there is no limit at all.". I can't find any special check for `MAX_RECORDS = 0`.

normanmaurer: @louxiu maybe I miss-read the diff but it seems when `MAX_RECORDS == 0` previous to this change a new record was created and added to lastRecords which is not true anymore. I just want to ensure this is what you want

louxiu: I must admit this change is a bit confusing. :-(

normanmaurer: so maybe revert this part ?

louxiu: I would prefer to keep the change. Maybe we should add extra comment or find a better way to handle `MAX_RECORDS == 0`;

normanmaurer: @louxiu then please add a comment to explain it.

louxiu: @normanmaurer done

normanmaurer:
@Scottmitch happy with the change ?

Scottmitch:
@louxiu - can you sign the [ICLA](http://netty.io/s/icla)?

louxiu:
@Scottmitch done

Scottmitch:
4.1 (96e06aa74) 4.0 (e8e317bdb)

PR #6975, opened 2017-07-17
SSLEngineTest should not depend on OpenSsl* class.

normanmaurer:
Motivation:

6152990073607602b2263109139cf829f3d2f7e4 introduced a test-case in SSLEngineTest which used OpenSsl.* which should not be done as this is am abstract bass class that is also used for non OpenSsl tests.

Modifications:

Move the protocol definations into SslUtils.

Result:

Cleaner code.


handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1351,20 +1351,20 @@ public final void setEnabledCipherSuites(String[] cipherSuites) {
                 return enabled.toArray(new String[1]);
             }
         }
-        if (isProtocolEnabled(opts, SSL.SSL_OP_NO_TLSv1, OpenSsl.PROTOCOL_TLS_V1)) {
-            enabled.add(OpenSsl.PROTOCOL_TLS_V1);
+        if (isProtocolEnabled(opts, SSL.SSL_OP_NO_TLSv1, SslUtils.PROTOCOL_TLS_V1)) {
+            enabled.add(SslUtils.PROTOCOL_TLS_V1);
```
Scottmitch: consider just adding a static import for these ... less verbose and they are used multiple times in this file.

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
```diff
@@ -375,36 +375,36 @@ public void testWrapWithDifferentSizesTLSv1() throws Exception {
...
-        testWrapWithDifferentSizes(OpenSsl.PROTOCOL_TLS_V1, "AECDH-RC4-SHA");
-        testWrapWithDifferentSizes(OpenSsl.PROTOCOL_TLS_V1, "DHE-RSA-SEED-SHA");
-        testWrapWithDifferentSizes(OpenSsl.PROTOCOL_TLS_V1, "AECDH-AES256-SHA");
-        testWrapWithDifferentSizes(OpenSsl.PROTOCOL_TLS_V1, "ECDHE-RSA-DES-CBC3-SHA");
-        testWrapWithDifferentSizes(OpenSsl.PROTOCOL_TLS_V1, "ADH-CAMELLIA256-SHA");
-        testWrapWithDifferentSizes(OpenSsl.PROTOCOL_TLS_V1, "DHE-RSA-CAMELLIA256-SHA");
-        testWrapWithDifferentSizes(OpenSsl.PROTOCOL_TLS_V1, "ECDHE-RSA-AES256-SHA");
-        testWrapWithDifferentSizes(OpenSsl.PROTOCOL_TLS_V1, "DHE-RSA-CAMELLIA128-SHA");
-        testWrapWithDifferentSizes(OpenSsl.PROTOCOL_TLS_V1, "ECDHE-RSA-RC4-SHA");
+        testWrapWithDifferentSizes(SslUtils.PROTOCOL_TLS_V1, "ADH-AES128-SHA");
```
Scottmitch: Just use static import ... in the future we won't have to change as many lines if we move the definition again.

handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
```diff
@@ -1691,13 +1691,13 @@ public void testHandshakeCompletesWithNonContiguousProtocolsTLSv1_2CipherOnly()
         clientSslCtx = SslContextBuilder.forClient()
                 .trustManager(InsecureTrustManagerFactory.INSTANCE)
                 .ciphers(Arrays.asList(sharedCipher))
-                .protocols(OpenSsl.PROTOCOL_TLS_V1_2, OpenSsl.PROTOCOL_TLS_V1)
+                .protocols(SslUtils.PROTOCOL_TLS_V1_2, SslUtils.PROTOCOL_TLS_V1)
```
Scottmitch: consider using a static import.

normanmaurer:
Comments addressed and cherry-picked into 4.1 (64a3e6c69c1ba49ccd7fe0410e6b2131cda9315c) and 4.0 (25364f101d)

PR #6976, opened 2017-07-17
Use array initializer expression

normanmaurer:
Motivation:

Code introduced in 6152990073607602b2263109139cf829f3d2f7e4 can be cleaned up and use array initializer expressions.

Modifications:

Use array initializer expressions.

Result:

Cleaner code.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslUtils.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslUtils.java
@@ -76,7 +76,7 @@ final class SslUtils {
      */
     static final int NOT_ENCRYPTED = -2;
 
-    static final String[] DEFAULT_CIPHER_SUITES = new String[] {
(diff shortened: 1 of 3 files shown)


handler/src/main/java/io/netty/handler/ssl/SslUtils.java
```diff
@@ -76,7 +76,7 @@
      */
     static final int NOT_ENCRYPTED = -2;
 
-    static final String[] DEFAULT_CIPHER_SUITES = new String[] {
+    static final String[] DEFAULT_CIPHER_SUITES = {
```

normanmaurer: false positive

normanmaurer:
As this is just cleanup I merged it.

4.1 (d4b9f3e4aa) and 4.0 (1a212ec9ff)

Scottmitch:
lgtm

PR #6978, opened 2017-07-17
AbstractByteBuf.ensureWritable(...) should check if buffer was released

normanmaurer:
Motivation:

AbstractByteBuf.ensureWritable(...) should check if buffer was released and if so throw an IllegalReferenceCountException

Modifications:

Ensure we throw in all cases.

Result:

More consistent and correct behaviour

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
@@ -266,7 +266,8 @@ public abstract class AbstractByteBuf extends ByteBuf {
         return this;
     }
 
-    private void ensureWritable0(int minWritableBytes) {
+    final void ensureWritable0(int minWritableBytes) {
+        ensureAccessible();
         if (minWritableBytes <= writableBytes()) {
             return;
         }
@@ -286,6 +287,7 @@ public abstract class AbstractByteBuf extends ByteBuf {
 
     @Override
     public int ensureWritable(int minWritableBytes, boolean force) {
+        ensureAccessible();
         if (minWritableBytes < 0) {
             throw new IllegalArgumentException(String.format(
                     "minWritableBytes: %d (expected: >= 0)", minWritableBytes));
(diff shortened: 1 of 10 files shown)


buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -267,6 +267,7 @@ public ByteBuf ensureWritable(int minWritableBytes) {
     }
 
     private void ensureWritable0(int minWritableBytes) {
+        ensureAccessible();
```
johnou: this feels backwards, isn't it usually the public method eg. mymethod invoking the private mymethod0?

fenik17: @johnou an `ensureAccessible()` method is protected, not is public.

johnou: I was looking at `public int ensureWritable(int minWritableBytes, boolean force)` but it even with the modifier as protected it still feels backwards.

fenik17: @johnou you thinks that it's better to move the call of `ensureAccessible` from `ensureWritable0` to `ensureWritable`?

fenik17: @johnou there is some sense in this..

normanmaurer: @johnou I think it is fine as otherwise I will need to duplicate some code.

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -1095,7 +1084,6 @@ public int writeBytes(InputStream in, int length)
 
     @Override
     public int writeBytes(ScatteringByteChannel in, int length) throws IOException {
-        ensureAccessible();
         ensureWritable(length);
         int writtenBytes = setBytes(writerIndex, in, length);
```
Scottmitch: the implementation of this method may call `checkIndex` but can now call `checkIndex0`. For example `PooledDirectByteBuf#setBytes(..)` and `PooledHeapByteBuf#setBytes(..)` call `checkIndex`. We should inspect all usages of `checkIndex` in buffer implementations which inherit from `AbstractByteBuf` as suggested in https://github.com/netty/netty/pull/6953#discussion_r126544205.

Scottmitch: Discussed with @normanmaurer offline ... both methods are public so we would have to create additional methods that bypass the additional call to ensureWritable (e.g `writeBytes` calls `setBytes0` which calls `checkIndex0`) ... lets not worry about this for now.

normanmaurer: +1...

normanmaurer:
Cherry-picked into 4.1 (d125adec38a0fa64011e5d77c5e4eadad69d4368) and 4.0 (44a7475002fd9834de44224e5ec05ad9a5f489d2)

Issue #6968, opened 2017-07-13
TlsReadTest tests are misleading


nayato:
Tests rely on channel having AutoRead = false and verify that read was issued through setting readIssued. In reality read method is called during construction of EmbeddedChannel - because AutoRead is true at this time, before AutoRead is set to false in test.

normanmaurer:
@nayato can you submit a pr with something that you think makes things more clear ?

Scottmitch:
@nayato - please check if https://github.com/netty/netty/pull/6982 is what you had in mind.

normanmaurer:
@nayato does https://github.com/netty/netty/pull/6987 work for you ?

normanmaurer:
Fixed by https://github.com/netty/netty/pull/6987

PR #6982, opened 2017-07-18
SslHandlerTest#TlsReadTest auto read initialization

Scottmitch:
Motivation:
SslHandlerTest#TlsReadTest attempts to carefully orchestrate when reads are done and that they are done independent of auto read being enabled in some scenarios. However the use of EmbeddedChannel enables auto read in the constructor before the unit test has a change to effectively disable it. This gives a false sense of security when verifying that reads were done independently of auto read.

Modifications:
- EmbeddedChannel should allow some channel options in the constructor and apply them before initializing the channel handlers

Result:
Fixes https://github.com/netty/netty/issues/6968


normanmaurer:
@Scottmitch I know I suggested the `Map` thing but after thinking a bit more about it I wonder if this may be better:

https://github.com/netty/netty/pull/6987

If you like it and we think it is better I will add more tests to `EmbeddedChannelTest` and also update the commit message. Just wanted to show it first without burning too many cycles.

Scottmitch:
Lets go with PR #6987 instead ... controlling the registration time should allow for even more flexibility.

PR #6985, opened 2017-07-18
MQTT unknown message type isn't handled as decoding error

### LINKED ISSUE #6984: MQTT unknown message type isn't handled as decoding error
### Expected behavior

When there is an MQTT decoding error, an invalidMessage is created and a DecodeResult with failure and related cause exception is filled with right information so that the user application can see what's the problem. It should happens in any step in the decoding phase.

### Actual behavior

It doesn't work if the problem is in the fixed header with a not valid MQTT message type.
Poisoning a message with the first byte set to 0xF0 (so 15 that is not a valid MQTT message type), the IllegalArgumentException is raised and propagated. There is no way to have a failed DecodeResult as in the other scenarios.

### Steps to reproduce

Building a server pipeline with the MqttDecoder and then after opening the connection sending a packet with just two bytes 0xF0 0x00.

### Netty version

4.1.13 Final

ppatierno:
Motivation:

MQTT unknown message type isn't handled as decoding error

Modification:

Catching exception during the MQTT decoding of the fixed header
Adding a unit test for unknown MQTT message type

Result:

Fixes #6984.

### DIFF
--- a/codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttDecoder.java
+++ b/codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttDecoder.java
@@ -71,11 +71,15 @@ public final class MqttDecoder extends ReplayingDecoder<DecoderState> {
     @Override
     protected void decode(ChannelHandlerContext ctx, ByteBuf buffer, List<Object> out) throws Exception {
         switch (state()) {
-            case READ_FIXED_HEADER:
+            case READ_FIXED_HEADER: try {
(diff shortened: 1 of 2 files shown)


normanmaurer:
@ppatierno seems like you not signed the ICLA yet. please do and let me know once you did:

https://netty.io/s/icla

ppatierno:
@normanmaurer just done !

normanmaurer:
Cherry-picked into 4.1 (b8d3d9655039b40bbe5df0446bfa50eb3543dd60) , @ppatierno thanks!

PR #6986, opened 2017-07-18
Update tests to not use TestUtils.getFreePort() and so ensure we not …

normanmaurer:
…try to use a port that is used by the system in the meantime.

Motivation:

We should not try to detect a free port in tests put just use 0 when bind so there is no race in which the system my bind something to the port we choosen before.

Modifications:

- Remove the usage of TestUtils.getFreePort() in the testsuite
- Remove hack to workaround bind errors which will not happen anymore now

Result:

Less flacky tests.

### DIFF
--- a/testsuite/src/main/java/io/netty/testsuite/transport/socket/AbstractDatagramTest.java
+++ b/testsuite/src/main/java/io/netty/testsuite/transport/socket/AbstractDatagramTest.java
@@ -20,15 +20,13 @@ import io.netty.buffer.ByteBufAllocator;
 import io.netty.channel.ChannelOption;
 import io.netty.testsuite.transport.AbstractComboTestsuiteTest;
 import io.netty.testsuite.transport.TestsuitePermutation;
-import io.netty.testsuite.util.TestUtils;
 import io.netty.util.NetUtil;
 
 import java.net.InetSocketAddress;
+import java.net.SocketAddress;
 import java.util.List;
 
 public abstract class AbstractDatagramTest extends AbstractComboTestsuiteTest<Bootstrap, Bootstrap> {
-
-    protected volatile InetSocketAddress addr;
 
     protected AbstractDatagramTest() {
         super(Bootstrap.class, Bootstrap.class);
@@ -41,16 +39,16 @@ public abstract class AbstractDatagramTest extends AbstractComboTestsuiteTest<Bo
 
     @Override
     protected void configure(Bootstrap bootstrap, Bootstrap bootstrap2, ByteBufAllocator allocator) {
-        addr = new InetSocketAddress(
-                NetUtil.LOCALHOST4, TestUtils.getFreePort());
-        bootstrap.localAddress(addr);
         bootstrap.option(ChannelOption.ALLOCATOR, allocator);
-        bootstrap2.localAddress(0).remoteAddress(addr);
         bootstrap2.option(ChannelOption.ALLOCATOR, allocator);
     }
 
-    protected void refreshLocalAddress(Bootstrap bootstrap) {
-        addr = new InetSocketAddress(NetUtil.LOCALHOST4, TestUtils.getFreePort());
-        bootstrap.localAddress(addr);
+    protected SocketAddress newSocketAddress() {
+        // We use LOCALHOST4 as we use InternetProtocolFamily.IPv4 when creating the DatagramChannel and its
+        // not supported to bind to and IPV6 address in this case.
+        //
+        // See also http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/e74259b3eadc/
+        // src/share/classes/sun/nio/ch/DatagramChannelImpl.java#l684
+        return new InetSocketAddress(NetUtil.LOCALHOST4, 0);
     }
 }
--- a/testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketConnectionAttemptTest.java
+++ b/testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketConnectionAttemptTest.java
@@ -23,7 +23,6 @@ import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.channel.ChannelOption;
 import io.netty.util.internal.SocketUtils;
-import io.netty.testsuite.util.TestUtils;
 import io.netty.util.NetUtil;
 import io.netty.util.concurrent.GlobalEventExecutor;
 import io.netty.util.concurrent.Promise;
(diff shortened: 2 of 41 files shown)


Scottmitch:
@normanmaurer - please address the build failures

testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketShutdownOutputByPeerTest.java
```diff
@@ -68,6 +70,9 @@ public void testShutdownOutput(ServerBootstrap sb) throws Throwable {
             Thread.sleep(100);
             assertEquals(1, h.halfClosureCount.intValue());
         } finally {
+            if (s != null) {
```
netkins: BLOCKER Change this condition so that it does not always evaluate to "true" [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2583)

testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketShutdownOutputByPeerTest.java
```diff
@@ -68,6 +70,9 @@ public void testShutdownOutput(ServerBootstrap sb) throws Throwable {
...
+                sc.close();
```
netkins: BLOCKER NullPointerException might be thrown as 'sc' is nullable here [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2259)

normanmaurer:
@Scottmitch fixed everything... PTAL again

testsuite/src/main/java/io/netty/testsuite/transport/socket/AbstractDatagramTest.java
```diff
@@ -41,16 +39,11 @@ protected AbstractDatagramTest() {
...
         bootstrap.option(ChannelOption.ALLOCATOR, allocator);
-        bootstrap2.localAddress(0).remoteAddress(addr);
         bootstrap2.option(ChannelOption.ALLOCATOR, allocator);
     }
 
-    protected void refreshLocalAddress(Bootstrap bootstrap) {
-        addr = new InetSocketAddress(NetUtil.LOCALHOST4, TestUtils.getFreePort());
-        bootstrap.localAddress(addr);
+    protected SocketAddress newSocketAddress() {
+        return new InetSocketAddress(NetUtil.LOCALHOST4, 0);
```
Scottmitch: it looks like the "loop 3 times" has been removed in `DatagramUnicastTest.java` ... do we still have to force ipv4?

testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketConnectionAttemptTest.java
```diff
@@ -96,7 +95,7 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
 
         cb.handler(handler);
         cb.option(ChannelOption.ALLOW_HALF_CLOSURE, halfClosure);
-        ChannelFuture future = cb.connect(NetUtil.LOCALHOST, TestUtils.getFreePort()).awaitUninterruptibly();
+        ChannelFuture future = cb.connect(NetUtil.LOCALHOST, BAD_PORT).awaitUninterruptibly();
```
Scottmitch: Do we really want to test connect is refused to a port that cannot be bound to ... or do we want to test that connection is refused to a port that could be bound to ... but that no-one is currently bound to?

testsuite/src/main/java/io/netty/testsuite/transport/udt/UDTClientServerConnectionTest.java
```diff
@@ -340,19 +338,16 @@ public void channelRead0(ChannelHandlerContext ctx, Object msg) throws Exception
      */
     @Test
     public void connection() throws Exception {
-
-        final String host = "localhost";
-        final int port = 1234;
-
         log.info("Starting server.");
-        final Server server = new Server(host, port);
+        // Using LOCALHOST4 as UDT transport not supports IPV6 :(
```
Scottmitch: `not supports` -> `does not support`

testsuite/src/main/java/io/netty/testsuite/transport/socket/AbstractDatagramTest.java
normanmaurer: yes as we construct the DatagramChannel with Ipv4 protocol

testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketConnectionAttemptTest.java
normanmaurer: hmmm true... maybe we should just use a port that no-one will use like tcp port 4 which is not assigned:

```
#                 4/tcp    Unassigned
```
This is from /etc/services...

wdyt ?

testsuite/src/main/java/io/netty/testsuite/transport/udt/UDTClientServerConnectionTest.java
normanmaurer: will fix

testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketConnectionAttemptTest.java
Scottmitch: Hum it would be preferred to crawl the ports but this is likely good enough. I guess put a `UNASSIGNED_PORT` constant in `TestUtils` (or something similar).

testsuite/src/main/java/io/netty/testsuite/transport/socket/AbstractDatagramTest.java
Scottmitch: I guess the JDK only supports IPv4 with UDP is what you mean?

Scottmitch: consider adding a comment here

Scottmitch: With a link to http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/e74259b3eadc/src/share/classes/sun/nio/ch/DatagramChannelImpl.java#l684

testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketConnectionAttemptTest.java
normanmaurer: The problem with crawling is that it may be possible the server bind the port in between. Let me use 4 for now.

normanmaurer:
@Scottmitch all addressed... PTAL again

normanmaurer:
Cherry-picked into 4.1 (3cdff36821) and 4.0 (928a9e9eaa)

PR #6987, opened 2017-07-18
Allow to delay registration when creating a EmbeddedChannel

normanmaurer:
Motivation:

    Some ChannelOptions must be set before the Channel is really registered to have the desired effect.

    Modifications:

    Add another constructor argument which allows to not register the EmbeddedChannel to its EventLoop until the user calls register().

    Result:

    More flexible usage of EmbeddedChannel. Also Fixes [#6968].

### DIFF
--- a/handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
@@ -160,27 +160,32 @@ public class SslHandlerTest {
         }
 
         public void test(final boolean dropChannelActive) throws Exception {
-          SSLEngine engine = SSLContext.getDefault().createSSLEngine();
-          engine.setUseClientMode(true);
+            SSLEngine engine = SSLContext.getDefault().createSSLEngine();
+            engine.setUseClientMode(true);
 
-          EmbeddedChannel ch = new EmbeddedChannel(
-              this,
-              new SslHandler(engine),
-              new ChannelInboundHandlerAdapter() {
-                @Override
-                public void channelActive(ChannelHandlerContext ctx) throws Exception {
-                  if (!dropChannelActive) {
-                    ctx.fireChannelActive();
-                  }
-                }
-              }
-          );
-          ch.config().setAutoRead(false);
-          assertFalse(ch.config().isAutoRead());
+            EmbeddedChannel ch = new EmbeddedChannel(false, false,
+                    this,
+                    new SslHandler(engine),
+                    new ChannelInboundHandlerAdapter() {
+                        @Override
+                        public void channelActive(ChannelHandlerContext ctx) throws Exception {
+                            if (!dropChannelActive) {
+                                ctx.fireChannelActive();
+                            }
+                        }
+                    }
+            );
+            ch.config().setAutoRead(false);
+            assertFalse(ch.config().isAutoRead());
 
-          assertTrue(ch.writeOutbound(Unpooled.EMPTY_BUFFER));
-          assertTrue(readIssued);
-          assertTrue(ch.finishAndReleaseAll());
+            ch.register();
+
+            assertTrue(readIssued);
+            readIssued = false;
+
+            assertTrue(ch.writeOutbound(Unpooled.EMPTY_BUFFER));
+            assertTrue(readIssued);
+            assertTrue(ch.finishAndReleaseAll());
        }
     }
 
--- a/transport/src/main/java/io/netty/channel/embedded/EmbeddedChannel.java
+++ b/transport/src/main/java/io/netty/channel/embedded/EmbeddedChannel.java
@@ -111,13 +111,26 @@ public class EmbeddedChannel extends AbstractChannel {
     }
 
     /**
+     * Create a new instance with the pipeline initialized with the specified handlers.
+     *
+     * @param register {@code true} if this {@link Channel} is registered to the {@link EventLoop} in the
+     *                 constructor. If {@code false} the user will need to call {@link #register()}.
+     * @param hasDisconnect {@code false} if this {@link Channel} will delegate {@link #disconnect()}
+     *                      to {@link #close()}, {@link false} otherwise.
+     * @param handlers the {@link ChannelHandler}s which will be add in the {@link ChannelPipeline}
+     */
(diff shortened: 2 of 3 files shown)


Scottmitch:
don't we still want to register/initialize the handlers?

normanmaurer:
@Scottmitch these will be automatically initialised when the channel will be registered.

transport/src/main/java/io/netty/channel/embedded/EmbeddedChannel.java
```diff
@@ -130,11 +143,25 @@ public EmbeddedChannel(ChannelId channelId, final ChannelHandler... handlers) {
...
+    public EmbeddedChannel(ChannelId channelId, boolean hasDisconnect, ChannelHandler... handlers) {
+        this(channelId, true, hasDisconnect, handlers);
+    }
+
+    /**
+     * Create a new instance with the channel ID set to the given ID and the pipeline
+     * initialized with the specified handlers.
+     *
+     * @param channelId the {@link ChannelId} that will be used to identify this channel
+     * @param hasDisconnect {@code false} if this {@link Channel} will delegate {@link #disconnect()}
```
normanmaurer: the param for register is missing... I will add it once we decide if we like this pr or not :)

Scottmitch:
@normanmaurer - it looks like the semantics of the test have changed though ... I don't see where the registration is done.

normanmaurer:
https://github.com/netty/netty/pull/6987/files#diff-9c1e6958b4f819c24efad4753e2545b0R181 ?

transport/src/main/java/io/netty/channel/embedded/EmbeddedChannel.java
```diff
@@ -174,9 +201,19 @@ protected void initChannel(Channel ch) throws Exception {
                 }
             }
         });
+        if (register) {
+            ChannelFuture future = loop.register(this);
```
netkins: MAJOR Remove this useless assignment to local variable "future". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1854)

transport/src/main/java/io/netty/channel/embedded/EmbeddedChannel.java
```diff
@@ -174,9 +201,19 @@ protected void initChannel(Channel ch) throws Exception {
...
+            assert future.isDone();
+        }
+    }
 
+    public void register()  throws Exception {
```
Scottmitch: nit: remove double space after `()`

transport/src/main/java/io/netty/channel/embedded/EmbeddedChannel.java
```diff
@@ -111,13 +111,26 @@ public EmbeddedChannel(boolean hasDisconnect, ChannelHandler... handlers) {
     }
 
     /**
+     * Create a new instance with the pipeline initialized with the specified handlers.
+     *
+     * @param register {@code true} if this {@link Channel} is directly registered on the {@link EventLoop}.
```
Scottmitch: `directly registered on the {@link EventLoop}` -> `registered to the {@link EventLoop} in the constructor`

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
```diff
@@ -160,27 +160,32 @@ public void read(ChannelHandlerContext ctx) throws Exception {
...
-          ch.config().setAutoRead(false);
-          assertFalse(ch.config().isAutoRead());
-
-          assertTrue(ch.writeOutbound(Unpooled.EMPTY_BUFFER));
-          assertTrue(readIssued);
-          assertTrue(ch.finishAndReleaseAll());
+            SSLEngine engine = SSLContext.getDefault().createSSLEngine();
+            engine.setUseClientMode(true);
+
+            EmbeddedChannel ch = new EmbeddedChannel(false, true,
```
Scottmitch: `hasDisconnect` is typically false by default ... should we preserve this behavior? Should we add a constructor that doesn't take this argument for now?

normanmaurer: yes we should... I think I would just not add yet another constructor for now.

normanmaurer:
@Scottmitch PTAL again

transport/src/main/java/io/netty/channel/embedded/EmbeddedChannel.java
normanmaurer: false positive...

transport/src/main/java/io/netty/channel/embedded/EmbeddedChannel.java
```diff
@@ -111,13 +111,26 @@ public EmbeddedChannel(boolean hasDisconnect, ChannelHandler... handlers) {
...
     /**
+     * Create a new instance with the pipeline initialized with the specified handlers.
+     *
+     * @param register {@code true} if this {@link Channel} is registered to the {@link EventLoop} in the
+     *                 constructor. If {@code false} the user will need to call {@link #register()}.
+     * @param hasDisconnect {@code false} if this {@link Channel} will delegate {@link #disconnect()}
+     *                      to {@link #close()}, {@link false} otherwise.
+     * @param handlers the {@link ChannelHandler}s which will be add in the {@link ChannelPipeline}
+     */
+    public EmbeddedChannel(boolean register, boolean hasDisconnect, ChannelHandler... handlers) {
```
Scottmitch: based upon https://github.com/netty/netty/pull/6987#discussion_r128170295 I was under the impression that we would just support `public EmbeddedChannel(boolean register, ChannelHandler... handlers)` for now.

normanmaurer: we cant as this clash with the existing constructor:

https://github.com/netty/netty/blob/66d82a22ea4dac414a101993342d93878a7e5ea2/transport/src/main/java/io/netty/channel/embedded/EmbeddedChannel.java#L109

normanmaurer:
Cherry-picked into 4.1 (ef22e65b570dd91591aafd5fadae73ded4bbfa8d) and 4.0 (b3d469e76601cdc217a1ef63d43938e0099fa9c7)

PR #5465, opened 2016-06-29
Remove unsafe char[] access in PlatformDependent

Scottmitch:
Motivation:
PlatformDependent attempts to use reflection to get the underlying char[] (or byte[]) from String objects. This is fragile as if the String implementation does not utilize the full array, and instead uses a subset of the array, this optimization is invalid. OpenJDK6 and some earlier versions of OpenJDK7 String have the capability to use a subsection of the underlying char[].

Modifications:
- PlatformDependent should not attempt to use the underlying array from String (or other data types) via reflection

Result:
PlatformDependent hash code generation for CharSequence does not depend upon specific JDK implementation details.

PR #6988, opened 2017-07-18
Use the correct murmur3 C1 value

kashike:
Introduced in a7f7d9c8e004aa13801424e6e665b1ec525a822f (#5465).

This fixes a (what I presume to be) typo, using the same value for both `C1` and `C2`.

https://en.wikipedia.org/wiki/MurmurHash
https://github.com/google/guava/blob/master/guava/src/com/google/common/hash/Murmur3_32HashFunction.java#L60

### DIFF
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent0.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent0.java
@@ -49,7 +49,7 @@ final class PlatformDependent0 {
 
     // constants borrowed from murmur3
     static final int HASH_CODE_ASCII_SEED = 0xc2b2ae35;
-    static final int HASH_CODE_C1 = 0x1b873593;
+    static final int HASH_CODE_C1 = 0xcc9e2d51;
     static final int HASH_CODE_C2 = 0x1b873593;
 
     /**


normanmaurer:
@kashike I think you are correct here...

@Scottmitch PTAL

normanmaurer:
@kashike  also can you please sign our ICLA:

https://netty.io/s/icla

kashike:
I've already signed it. :-)

normanmaurer:
@kashike ha! thanks :)

Scottmitch:
4.1 (c43e09da5)

I re-ran the `HashCollisionTest` (with some additional values) ... there was 1 more collision after this change but not on an http header so all is good ... thanks @kashike!

testn:
Why didn't the test code catch this?

Scottmitch:
We are not re-implementing murmur's algorithm. We have a custom hash algorithm and just happen to use the same constants as murmur (not required). This hash algorithm is designed to balance speed vs collision for a specific use case (case insensitive ascii headers). This hash algorithm was faster than murmur (or a variant of murmur) for this case when last evaluated and no collisions amongst known header names where detected.

Issue #6979, opened 2017-07-17
bom still uses snapshot for netty-transport-native-unix-common version


tsachev:
### Expected behavior
released bom should not contain snapshot versions
### Actual behavior
it still has snapshot version
### Minimal yet complete reproducer code (or URL to code)
see
http://repo1.maven.org/maven2/io/netty/netty-bom/4.1.13.Final/netty-bom-4.1.13.Final.pom
and
http://repo1.maven.org/maven2/io/netty/netty-bom/4.0.49.Final/netty-bom-4.0.49.Final.pom
### Netty version
4.1.13.Final and 4.0.49.Final
### OS version (e.g. `uname -a`)

`netty-transport-unix-common` should be `netty-transport-native-unix-common`

Seem that this was addressed by @wilkinsona in #6849 , but something went wrong in the end probably when merging/cherry-picking and #6852 also used the wrong artifact ids

in `4.1.13.Final`
```
      <dependency>
        <groupId>io.netty</groupId>
        <artifactId>netty-transport-unix-common</artifactId>
        <version>4.1.13.Final-SNAPSHOT</version>
        <classifier>linux-x86_64</classifier>
      </dependency>
      <dependency>
        <groupId>io.netty</groupId>
        <artifactId>netty-transport-unix-common</artifactId>
        <version>4.1.13.Final-SNAPSHOT</version>
        <classifier>osx-x86_64</classifier>
      </dependency>
```
in `4.0.49.Final`
```
      <dependency>
        <groupId>io.netty</groupId>
        <artifactId>netty-transport-unix-common</artifactId>
        <version>4.0.48.Final-SNAPSHOT</version>
      </dependency>
```

normanmaurer:
@tsachev will investigate later today.

normanmaurer:
Please check https://github.com/netty/netty/pull/6989

normanmaurer:
Fixed by https://github.com/netty/netty/pull/6989

PR #6989, opened 2017-07-18
Correct typo in artifactId of dependency in bom pom.xml

normanmaurer:
Motivation:

    There was a typo in a dependency in the bom pom.xml which lead to have it specify a non-existing artifact and also so not have the maven release plugin update the version correctly.

    Modifications:

    Rename netty-transport-unix-common to netty-transport-native-unix-common and also fix the version.

    Result:

    Fixes [#6979]

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -198,14 +198,14 @@
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
-        <artifactId>netty-transport-unix-common</artifactId>
-        <version>4.1.13.Final-SNAPSHOT</version>
+        <artifactId>netty-transport-native-unix-common</artifactId>
+        <version>4.1.14.Final-SNAPSHOT</version>
         <classifier>linux-x86_64</classifier>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
-        <artifactId>netty-transport-unix-common</artifactId>
-        <version>4.1.13.Final-SNAPSHOT</version>
+        <artifactId>netty-transport-native-unix-common</artifactId>
+        <version>4.1.14.Final-SNAPSHOT</version>
         <classifier>osx-x86_64</classifier>
       </dependency>
       <dependency>


normanmaurer:
@tsachev and @wilkinsona PTAL ... This should fix [#6979]

Scottmitch:
I thought we had this at one point though ... and there was something with the release process that prevented using the variable? I may be confusing some other issue though...

wilkinsona:
That was my concern too. It looks fine in terms of using the variable consistently, but I don't know anything about Netty's release process so I also don't know if it'll work as intended.

normanmaurer:
@Scottmitch @wilkinsona the problem is that if we use the version directly the maven release plugin will NOT update the version for the dependencies that also have the classifier listed. So I think this is the way to go and also "consistent" with how we add the dependencies in the all module.

tsachev:
I do not have an opinion on using project.version (unless it makes the bom useless as parent, but that is not its purpose anyway).
There might be something in the release process that replaces `${project.version}` with the actual project version I don't know.

But #6979 is about using wrong artifactId which manifests itself in the bom, because maven-release-plugin does not see this non existing module and fails to update them.

`<artifactId>netty-transport-unix-common</artifactId>` should become `<artifactId>netty-transport-native-unix-common</artifactId>` everywhere. Note the missing *native* in the name.

normanmaurer:
@tsachev doh! I missed this ... so I think this is the real issue then.. Let me update the pr to just fix the artifactId, this should be good enough.

normanmaurer:
@Scottmitch @wilkinsona @tsachev PTAL again

wilkinsona:
LGTM

normanmaurer:
Cherry-picked into 4.1 (deb5c4520492e2bed02ccd58cd43318f1a52db3d) and committed a updated version to 4.0 (ae66f3ae9edfc76136a3c5d736d2be91fe157748)

Scottmitch:
lgtm

PR #6990, opened 2017-07-19
Increase timeout for DnsNameResolverTest

Scottmitch:
Motivation:
DnsNameResolverTest has been observed to timeout on the CI servers. We should increase the timeout from 5 seconds to 30 seconds.

Modifications:
- Increase timeout from 5 to 30 seconds.

Result:
Less false failures due to slower CI machines.


normanmaurer:
Cherry-picked into 4.1 (0afe4e09643c457c452ef0d9563c46f8e7c24ad1)

PR #6992, opened 2017-07-19
Support JDK9-native ALPN

### LINKED ISSUE #6933: Netty 4.1 should support ALPN on JDK9
This is obviously a known issue, but it's worth tracking. It may also be tricky to pull this off while compiling against JDK6; in particular, invoking `SSLEngine#setHandshakeApplicationProtocolSelector` requires passing a `BiFunction` implementation. I see a few ways of accomplishing this, assuming that calling this method turns out to be necessary:

1. Take advantage of [JEP 238](http://openjdk.java.net/jeps/238) (Multi-Release JAR Files) to compile and link a small amount of code directly against JDK9
2. Use reflection to call `java.lang.invoke.MethodHandleProxies#asInterfaceInstance`
3. Use a library like ASM to manually generate a `BiFunction` implementation at runtime

rschmitt:
Motivation:
Netty is unable to take advantage of JDK9's built-in ALPN support.

Modifications:

When running on JDK9 or later, Netty will invoke Java's built-in methods
for specifying the supported (server) or requested (client) application
layer protocols and for obtaining the negotiated protocol after the
handshake has taken place. The relevant methods are new in Java 9, so
they are called using core reflection. As all of the methods in question
are public, there are no issues involving security, Jigsaw
encapsulation, or even performance (as all the Method instances are
cached in static final fields).

Result:
Fixes #6933.


handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+
+    static SSLEngine configureAlpn(SSLEngine engine, JdkApplicationProtocolNegotiator applicationNegotiator) {
+        List<String> supportedProtocols = applicationNegotiator.protocols();
+        SSLParameters params = engine.getSSLParameters();
+        try {
+            setApplicationProtocols(params, supportedProtocols);
+        } catch (Exception ex) {
+            throw new RuntimeException(ex);
```
normanmaurer: IllegalStateException

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+        engine.setSSLParameters(params);
+        return engine;
+    }
+
+    static String getApplicationProtocol(SSLEngine sslEngine) {
+        try {
+            return (String) getApplicationProtocol.invoke(sslEngine);
+        } catch (Exception ex) {
+            throw new RuntimeException(ex);
```
normanmaurer: IllegalStateException

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+
+    private static void setApplicationProtocols(SSLParameters parameters, List<String> supportedProtocols) {
+        String[] protocolArray = supportedProtocols.toArray(new String[]{});
```
normanmaurer: `EmptyArrays.EMPTY_STRINGS`

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,90 @@
...
+        try {
+            setApplicationProtocols.invoke(parameters, new Object[]{ protocolArray });
+        } catch (Exception ex) {
+            throw new RuntimeException(ex);
```
normanmaurer: IllegalStateException

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -574,11 +575,16 @@ public SSLEngine engine() {
      */
     public String applicationProtocol() {
         SSLSession sess = engine().getSession();
-        if (!(sess instanceof ApplicationProtocolAccessor)) {
-            return null;
+        if (sess instanceof ApplicationProtocolAccessor) {
+            return ((ApplicationProtocolAccessor) sess).getApplicationProtocol();
+        } else if (Java9SslUtils.supportsAlpn(engine())) {
```
normanmaurer: nit you can remove the `} else {` as you return in the `if`.

normanmaurer:
@rschmitt also please sign our ICLA:

https://netty.io/s/icla

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,90 @@
...
+    static {
+        Method setter = null;
+        Method getter = null;
+
+        if (PlatformDependent.javaVersion() >= 9) {
+            try {
+                getter = SSLEngine.class.getMethod("getApplicationProtocol");
+                setter = SSLParameters.class.getMethod("setApplicationProtocols", String[].class);
+            } catch (Throwable t) {
+                throw new RuntimeException("Unable to initialize Java9SslUtils", t);
```
normanmaurer: I think this may be a bit to aggressive as these methods could also throw if a SecurityManager is in place or I am wrong ?

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,90 @@
...
+            }
+        }
+
+        getApplicationProtocol = getter;
+        setApplicationProtocols = setter;
```
normanmaurer: I think we should also try to invoke these methods here.

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,90 @@
...
+
+    static boolean supportsAlpn(SSLEngine engine) {
+        if (PlatformDependent.javaVersion() < 9) {
+            return false;
+        }
+
+        try {
+            getApplicationProtocol.invoke(engine);
+            return true;
+        } catch (Exception ignore) {
```
normanmaurer: Is this good enough ? I think a security manager could also throw runtime exceptions ?

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,90 @@
...
+            return false;
+        }
+    }
+
+    static SSLEngine configureAlpn(SSLEngine engine, JdkApplicationProtocolNegotiator applicationNegotiator) {
+        List<String> supportedProtocols = applicationNegotiator.protocols();
+        SSLParameters params = engine.getSSLParameters();
+        try {
+            setApplicationProtocols(params, supportedProtocols);
+        } catch (Exception ex) {
```
normanmaurer: Is this good enough ? I think a security manager could also throw runtime exceptions ?

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,90 @@
...
+            throw new RuntimeException(ex);
+        }
+        engine.setSSLParameters(params);
+        return engine;
+    }
+
+    static String getApplicationProtocol(SSLEngine sslEngine) {
+        try {
+            return (String) getApplicationProtocol.invoke(sslEngine);
+        } catch (Exception ex) {
```
normanmaurer: Is this good enough ? I think a security manager could also throw runtime exceptions ?

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,90 @@
...
+            throw new RuntimeException(ex);
+        }
+    }
+
+    private static void setApplicationProtocols(SSLParameters parameters, List<String> supportedProtocols) {
+        String[] protocolArray = supportedProtocols.toArray(new String[]{});
+        try {
+            setApplicationProtocols.invoke(parameters, new Object[]{ protocolArray });
+        } catch (Exception ex) {
```
normanmaurer: Is this good enough ? I think a security manager could also throw runtime exceptions ?

johnou: http://docs.oracle.com/javase/7/docs/api/java/lang/Class.html#getMethod(java.lang.String,%20java.lang.Class...) on the other hand may throw a SecurityException, nice catch @normanmaurer!

johnou: @rschmitt please wrap with `AccessController.doPrivileged(new PrivilegedAction..`

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,101 @@
...
+                    }
+                });
+                setter = AccessController.doPrivileged(new PrivilegedExceptionAction<Method>() {
+                    @Override
+                    public Method run() throws Exception {
+                        return SSLParameters.class.getMethod("setApplicationProtocols", String[].class);
+                    }
+                });
+            } catch (Throwable t) {
+                throw new RuntimeException("Unable to initialize Java9SslUtils", t);
```
johnou: Instead of throwing a runtime exception inside the static init block we could assign a new static supportsAlpn boolean (including the invoke test) then simply return that in the function below.

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
johnou: runtime exception extends exception?

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -574,11 +575,17 @@ public SSLEngine engine() {
      */
     public String applicationProtocol() {
         SSLSession sess = engine().getSession();
-        if (!(sess instanceof ApplicationProtocolAccessor)) {
-            return null;
+        if (sess instanceof ApplicationProtocolAccessor) {
+            return ((ApplicationProtocolAccessor) sess).getApplicationProtocol();
+        }
+        if (Java9SslUtils.supportsAlpn(engine())) {
```
johnou: Missing PlatformDependent.javaVersion() >= 9 check no?

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
normanmaurer: brainfart... all good 👍

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,109 @@
...
+                getter = AccessController.doPrivileged(new PrivilegedExceptionAction<Method>() {
+                    @Override
+                    public Method run() throws Exception {
+                        return SSLEngine.class.getMethod("getApplicationProtocol");
+                    }
+                });
+                setter = AccessController.doPrivileged(new PrivilegedExceptionAction<Method>() {
+                    @Override
+                    public Method run() throws Exception {
+                        return SSLParameters.class.getMethod("setApplicationProtocols", String[].class);
```
johnou: If I use Netty in my project but not Netty's alpn and this is invoked, will it cause problems because setApplicationProtocols is invoked with an empty array?

johnou: Probably not because it doesn't look like a static method.

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
johnou: @normanmaurer or was the issue with classloading only if we invoke a method which contains a class which isn't on the class path?

normanmaurer:
@rschmitt did you see my question about testing ? Will this be enough to ensure the ALPN tests are run for java9 as well ?

normanmaurer:
@Scottmitch PTAL as well... Also @trustin if you have cycles.

normanmaurer:
@rschmitt ping :)

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
rschmitt: I think you're referring to the problem where a class is statically linked against a class that is not present at runtime, such as in the case where javac is configured to link against the JDK8 standard library while producing bytecode compatible with Java 6. The advantage of using reflection in Java9SslUtils is that it is Java 6 compatible on all three levels (source, target, and library). In recent revisions of the PR, the static initializer also tolerates runtimes older than JDK9, so it is always safe to call supportsAlpn.

rschmitt:
@normanmaurer It looks like the existing test coverage is insufficient, so I'll need to add some test cases specific to Java 9.

normanmaurer:
@rschmitt can you ping me once you have tests added ?

normanmaurer:
@rschmitt can you please rebase :)

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,180 @@
...
+
+            SSLEngine sslEngine = (SSLEngine) args[0];
+            List<String> protocols = (List<String>) args[1];
+            for (String p : supportedProtocols) {
+                if (protocols.contains(p)) {
+                    return p;
+                }
+            }
+
+            return "";
```
normanmaurer: return StringUtil.EMPTY

handler/src/main/java/io/netty/handler/ssl/JdkAlpnApplicationProtocolNegotiator.java
```diff
@@ -1,136 +0,0 @@
...
- */
-package io.netty.handler.ssl;
-
-import io.netty.buffer.ByteBufAllocator;
-import javax.net.ssl.SSLEngine;
-
-/**
- * The {@link JdkApplicationProtocolNegotiator} to use if you need ALPN and are using {@link SslProvider#JDK}.
- */
-public final class JdkAlpnApplicationProtocolNegotiator extends JdkBaseApplicationProtocolNegotiator {
```
normanmaurer: @rschmitt we can not remove the class as it was public.

handler/src/main/java/io/netty/handler/ssl/JdkApplicationProtocolNegotiator.java
```diff
@@ -17,120 +17,115 @@
...
 import java.util.List;
-import java.util.Set;
+
+import io.netty.handler.ssl.ApplicationProtocolConfig.Protocol;
+import io.netty.util.internal.PlatformDependent;
 
 /**
  * JDK extension methods to support {@link ApplicationProtocolNegotiator}
  */
-public interface JdkApplicationProtocolNegotiator extends ApplicationProtocolNegotiator {
```
normanmaurer: we can not remove the interface as it was public

handler/src/main/java/io/netty/handler/ssl/JdkNpnApplicationProtocolNegotiator.java
```diff
@@ -1,121 +0,0 @@
...
- */
-package io.netty.handler.ssl;
-
-import io.netty.buffer.ByteBufAllocator;
-import javax.net.ssl.SSLEngine;
-
-/**
- * The {@link JdkApplicationProtocolNegotiator} to use if you need NPN and are using {@link SslProvider#JDK}.
- */
-public final class JdkNpnApplicationProtocolNegotiator extends JdkBaseApplicationProtocolNegotiator {
```
normanmaurer: we can not remove this class as it was public

handler/src/main/java/io/netty/handler/ssl/JdkApplicationProtocolNegotiator.java
```diff
@@ -17,120 +17,115 @@
...
+            for (String p : supportedProtocols) {
+                if (protocols.contains(p)) {
+                    engineWrapper.getSession().setApplicationProtocol(p);
+                    return p;
+                }
+            }
+
+            if (failIfNoCommonProtocols) {
+                throw new SSLHandshakeException("Selected protocol is not supported");
+            } else {
```
normanmaurer: you can remove the else as in the if we throw

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,180 @@
...
+            setHandshakeApplicationProtocolSelector.invoke(sslEngine, biFunction);
+        } catch (Exception ex) {
+            throw new IllegalStateException(ex);
+        }
+    }
+
+    private static final class AlpnSelector implements InvocationHandler {
+        private final List<String> supportedProtocols;
+
+        private AlpnSelector(List<String> supportedProtocols) {
```
normanmaurer: use package private constructor as you call it from the outer class.

handler/src/main/java/io/netty/handler/ssl/JdkApplicationProtocolNegotiator.java
```diff
@@ -17,120 +17,115 @@
...
+
+    private final List<String> supportedProtocols;
+    private final boolean failIfNoCommonProtocols;
+    private final Protocol protocol;
+
+    JdkApplicationProtocolNegotiator(
+        List<String> supportedProtocols,
+        boolean failIfNoCommonProtocols,
+        Protocol protocol
+    ) {
```
normanmaurer: put all the constructor args on one line

handler/src/main/java/io/netty/handler/ssl/JdkApplicationProtocolNegotiator.java
```diff
@@ -17,120 +17,115 @@
...
+        this.supportedProtocols = supportedProtocols;
```
normanmaurer: wraps this with `Collections.unmodifiableList(...)` as we return the List via a getter.

handler/src/main/java/io/netty/handler/ssl/JdkAlpnApplicationProtocolNegotiator.java
rschmitt: Could you clarify Netty's compatibility contract? I ask because commit 4448b8f42f599e79db1744cd8f5fdfee702c195e recently changed the type signature of `SslEngineWrapperFactory#wrapSslEngine` in a non-backwards-compatible way, despite the fact that that is technically a public method declared on a public interface. Similarly, I judged these classes to be internal Netty classes, as they are not exposed to end users through `ApplicationProtocolConfig` and are, like `wrapSslEngine`, just an implementation detail of Netty's ALPN support.

normanmaurer: doh! You are right... @nmittler any way to fix this ?

normanmaurer: @rschmitt as you stated it may not be a big deal but if possible I would love to not do this as the class is public. If I could change it I would have made the class package private from the start but we missed it :(

Scottmitch: FWIW 4448b8f hasn't been included in a release yet, so we can still restore API breakage with no harm done. At one point we didn't have `ApplicationProtocolConfig` so this class was public ... @nmittler, @normanmaurer, and I discussed removing this class from the public interface when introducing conscrypt support and decided not to due to preserving public APIs.

Scottmitch: Is it possible to deprecate this class and point to the new class ... and just not use this class in Netty anymore?

handler/src/main/java/io/netty/handler/ssl/JdkApplicationProtocolNegotiator.java
Scottmitch: Same with the other interface that were removed inside this API

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,180 @@
...
+    static {
+        boolean support = false;
+        Method setter = null;
+        Method getter = null;
+        Method selector = null;
+        Class biFunction = null;
+
+        if (PlatformDependent.javaVersion() >= 9) {
+            try {
+                getter = AccessController.doPrivileged(new PrivilegedExceptionAction<Method>() {
```
nmittler: Just curious ... why do you need `doPrivileged` here?

johnou: asserting privileges for class.getmethod 

http://docs.oracle.com/javase/7/docs/api/java/lang/Class.html#getMethod(java.lang.String,%20java.lang.Class...) 

SecurityException - If a security manager, s, is present and any of the following conditions is met:
invocation of s.checkMemberAccess(this, Member.PUBLIC) denies access to the method
the caller's class loader is not the same as or an ancestor of the class loader for the current class and invocation of s.checkPackageAccess() denies access to the package of this class

handler/src/main/java/io/netty/handler/ssl/JdkAlpnApplicationProtocolNegotiator.java
normanmaurer: @Scottmitch so I fixed the problem introduced by  https://github.com/netty/netty/commit/4448b8f42f599e79db1744cd8f5fdfee702c195e via https://github.com/netty/netty/pull/7110 . Any chance to also fix the breakage here after rebasing ?

normanmaurer: Ups sorry @rschmitt ^^

normanmaurer:
Let me close this in favour of https://github.com/netty/netty/pull/7126.

Issue #6993, opened 2017-07-19
Netty Epoll fails in IPv6 only environment


carl-mastrangelo:
https://github.com/netty/netty/blob/83de77fbe5385130ba9afc666fecbe9b59bd8032/transport-native-unix-common/src/main/c/netty_unix_socket.c#L198

In this code, it tries to set dual stack sockets, but fails when running in an IPv6 only environment.  Such an environment exists in Google by blocking all IPv4 functionality in the kernel in order to prove code is IPv6 compatible.  

cc: @pmarks-net

normanmaurer:
@Scottmitch please remember to set the milestone when you close issues so it is listed for the correct version.

PR #6994, opened 2017-07-19
Allowed Netty Epoll/Kqueue to work in IPv6 Only environments

carl-mastrangelo:
Motivation:
In some environments, IPv4 may be disabled (at a kernel level).
Google has such an environment for testing v4 -> v6 transition
paths.  This give confidence that code is v6 ready.

Modifications:
Change native socket code to ignore failures of trying to enter
dual stack mode.  This change has been made to Google's internal
JDK, and will/should be upstreamed to OpenJDK eventually.

Results:
Netty works in IPv6 only environments

Fixes: #6993

### DIFF
--- a/transport-native-unix-common/src/main/c/netty_unix_socket.c
+++ b/transport-native-unix-common/src/main/c/netty_unix_socket.c
@@ -188,17 +188,40 @@ static int socket_type(JNIEnv* env) {
     }
 }
 
+static void netty_unix_socket_optionHandleError(JNIEnv* env, int err, char* method) {
+    if (err == EBADF) {
+        netty_unix_errors_throwClosedChannelException(env);
+    } else {
+        netty_unix_errors_throwChannelExceptionErrorNo(env, method, err);
+    }
+}
+
+static void netty_unix_socket_setOptionHandleError(JNIEnv* env, int err) {
+    netty_unix_socket_optionHandleError(env, err, "setsockopt() failed: ");
+}
+
+static int netty_unix_socket_setOption0(jint fd, int level, int optname, const void* optval, socklen_t len) {
+    return setsockopt(fd, level, optname, optval, len);
+}
+
 static jint _socket(JNIEnv* env, jclass clazz, int type) {
     int fd = nettyNonBlockingSocket(socketType, type, 0);
     if (fd == -1) {
         return -errno;
     } else if (socketType == AF_INET6) {
-        // Allow to listen /connect ipv4 and ipv6
+        // Try to allow listen /connect ipv4 and ipv6
         int optval = 0;
-        if (netty_unix_socket_setOption(env, fd, IPPROTO_IPV6, IPV6_V6ONLY, &optval, sizeof(optval)) < 0) {
-            // Something went wrong so close the fd and return here. setOption(...) itself throws the exception already.
-            close(fd);
-            return -1;
+        if (netty_unix_socket_setOption0(fd, IPPROTO_IPV6, IPV6_V6ONLY, &optval, sizeof(optval)) < 0) {
+            if (errno != EAFNOSUPPORT) {
+              netty_unix_socket_setOptionHandleError(env, errno);
+              // Something went wrong so close the fd and return here. setOption(...) itself throws the exception already.
+              close(fd);
+              return -1;
+            }
+            // else we failed to enable dual stack mode.
+            // It is assumed the socket is re‐stricted to sending and receiving IPv6 packets only.
+            // Don't close fd and don't return -1. At best we can do is log.
+            // TODO: bubble this up to an actual Logger.
         }
     }
     return fd;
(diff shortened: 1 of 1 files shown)


transport-native-unix-common/src/main/c/netty_unix_socket.c
```diff
@@ -188,17 +188,48 @@ static int socket_type(JNIEnv* env) {
...
         int optval = 0;
-        if (netty_unix_socket_setOption(env, fd, IPPROTO_IPV6, IPV6_V6ONLY, &optval, sizeof(optval)) < 0) {
-            // Something went wrong so close the fd and return here. setOption(...) itself throws the exception already.
-            close(fd);
-            return -1;
+        if (netty_unix_socket_setOption0(fd, IPPROTO_IPV6, IPV6_V6ONLY, &optval, sizeof(optval)) < 0) {
+            if (errno == EAFNOSUPPORT) {
+              // Something went wrong, but don't close fd and don't return -1.   At best we can do is log.
+              // TODO: bubble this up to an actual Logger.
+              fprintf(stderr, "setsockopt() failed: Unable to set IPV6_V6ONLY.  Continuing.\n");
```
normanmaurer: Please address the TODO

normanmaurer: Or just remove the fprintf for now

transport-native-unix-common/src/main/c/netty_unix_socket.c
```diff
@@ -289,26 +320,10 @@ static jobject _recvFrom(JNIEnv* env, jint fd, void* buffer, jint pos, jint limi
     return createDatagramSocketAddress(env, &addr, res);
 }
 
-static void netty_unix_socket_optionHandleError(JNIEnv* env, int err, char* method) {
-    if (err == EBADF) {
```
normanmaurer: Please revert these changes as it seems this just moves code and so is not related

Scottmitch: IIUC the static methods have to move up to avoid their first use case or we need to declare the method before the first use.

transport-native-unix-common/src/main/c/netty_unix_socket.c
```diff
@@ -321,14 +336,6 @@ int netty_unix_socket_getOption(JNIEnv* env, jint fd, int level, int optname, vo
     return rc;
 }
 
-int netty_unix_socket_setOption(JNIEnv* env, jint fd, int level, int optname, const void* optval, socklen_t len) {
```
Scottmitch: IIUC this method doesn't have to move (it isn't static and is declared in the header file)?

transport-native-unix-common/src/main/c/netty_unix_socket.c
Scottmitch: Why do we want to suppress the error here? `socket_type` accounts for `EAFNOSUPPORT` so if we get it here it seems like a programming error. Can we just propagate the error via `netty_unix_socket_setOptionHandleError` in this case too?

transport-native-unix-common/src/main/c/netty_unix_socket.c
carl-mastrangelo: netty_unix_socket_setOption0 needs to be defined before _socket() uses it, and I wanted to keep most of these functions together.  

I can move the minimal number of functions to do so, or move _socket down.  Preference?

transport-native-unix-common/src/main/c/netty_unix_socket.c
carl-mastrangelo: done.

transport-native-unix-common/src/main/c/netty_unix_socket.c
Scottmitch: Moving just the static methods sgtm

transport-native-unix-common/src/main/c/netty_unix_socket.c
```diff
@@ -188,17 +188,47 @@ static int socket_type(JNIEnv* env) {
...
     } else if (socketType == AF_INET6) {
-        // Allow to listen /connect ipv4 and ipv6
+        // Try to allow listen /connect ipv4 and ipv6
         int optval = 0;
-        if (netty_unix_socket_setOption(env, fd, IPPROTO_IPV6, IPV6_V6ONLY, &optval, sizeof(optval)) < 0) {
-            // Something went wrong so close the fd and return here. setOption(...) itself throws the exception already.
-            close(fd);
-            return -1;
+        if (netty_unix_socket_setOption0(fd, IPPROTO_IPV6, IPV6_V6ONLY, &optval, sizeof(optval)) < 0) {
+            if (errno == EAFNOSUPPORT) {
```
Scottmitch: Why do we want to suppress the error here? socket_type accounts for EAFNOSUPPORT so if we get it here it seems like a programming error. Can we just propagate the error via netty_unix_socket_setOptionHandleError in this case too?

Scottmitch: This was collapsed in https://github.com/netty/netty/pull/6994#discussion_r128362622

transport-native-unix-common/src/main/c/netty_unix_socket.c
carl-mastrangelo: Sure, it doesn't have to, but some teams prefer to keep related code adjacent / contiguous if possible.  If you don't have such a requirement I can move the bare minimum.

transport-native-unix-common/src/main/c/netty_unix_socket.c
carl-mastrangelo: That code checks if AF_INET6 is allowed, but not single stack mode which is what `_socket()` is asking for.

transport-native-unix-common/src/main/c/netty_unix_socket.c
carl-mastrangelo: Done.

transport-native-unix-common/src/main/c/netty_unix_socket.c
Scottmitch: Can we remote the `if` since it doesn't do anything? What about the following:

```c
if (errno != EAFNOSUPPORT) {
...
}
// else we failed to enable dual stack mode.
// It is assumed the socket is re‐stricted to sending and receiving IPv6 packets only.
```

carl-mastrangelo: Done.

Scottmitch:
4.1 (3d22b2424)

I will not back port this to 4.0 because the code has diverged in this area.

normanmaurer:
@Scottmitch not backporting sounds good.

PR #6995, opened 2017-07-19
Automatically detect shaded packagePrefix

### LINKED ISSUE #6963: -Dio.netty.packagePrefix should not be required when shading
### Expected behavior

A library may use shading transparently to the user and would be able to be combined with other libraries that also use Netty. Similarly, an application should be able to use shading transparently to the user.

### Actual behavior

The shader has to choose whether to rename the native components of epoll and tcnative. If they are renamed, then `-Dio.netty.packagePrefix` must be set at runtime, which may not be possible programatically. If they aren't renamed, then epoll and tcnative can fail when mixed with other libraries because: 1) there is a name collision of the `META-INF/native/` files in the class path and 2) [`System.loadLibrary(String)` does nothing if it previously loaded a particular name](http://docs.oracle.com/javase/8/docs/api/java/lang/Runtime.html#loadLibrary-java.lang.String-).

### Possible fixes

`NativeLibraryLoader` [currently](https://github.com/netty/netty/blob/f208b147a6c1748092f1d23f71865e9082b00534/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java#L188) defaults to a package prefix of `""`:
```java
String name = SystemPropertyUtil.get("io.netty.packagePrefix", "").replace('.', '-') + originalName;
```

Instead of defaulting to `""`, I think it should at least attempt to find a package prefix from the class path. There are many options. A sampling:
```java
// Semi-readable implicit
String maybeShaded = NativeLibraryLoader.class.getName();
// Use ! instead of . to avoid shading utilities from modifying the string
String expected = "io!netty!util!internal!NativeLibraryLoader".replace('!', '.');
if (maybeShaded.endsWith(expected)) {
  return maybeShaded.substring(0, maybeShaded.length() - expected.size());
} else {
  throw ...;
}

// Quick and dirty implicit
String name = NativeLibraryLoader.class.getName();
// 42 == "io.netty.util.internal.NativeLibraryLoader".length()
return name.substring(0, name.length() - 42);

// Pre-configured (explicit). User created a file for Netty to read
String resourceName = obfuscateConcat("/META-INF/native/", "io.netty.packagePrefix");
// Could also be a relative, but then users would need to create a file in
// io/netty/util/internal or similar subpackage (I don't see any class in
// io.netty package).
// String resourceName = "packagePrefix";
InputStream is = NativeLibraryLoader.class.getResourceAsStream(resourceName);
if (is == null) {
  return "";
}
return asStringUtf8(is);
```

In my mind, when shaded, a `""` fallback is counter-productive. It gives the appearance that Netty has been properly shaded when in fact it still collides. As a user, I'd much rather epoll and tcnative obviously fail when improperly shaded. So for me an implicit solution has the _benefit_ of failing when misconfigured. But we can mix-and-match behaviors pretty easily.

Depending on the solution, supporting `io.netty.packagePrefix` may not be necessary any more. An explicit configuration by the user (via `io.netty.packagePrefix` or similar) may be a necessary supplement to implicit solutions since shading may not use prefixing; `io.netty` may have been rewritten as `a.b` or `com.example.shadednetty`.

I'd be willing to submit a PR for necessary changes, although I'd hope we could agree on an approach first.

ejona86:
Motivation:

Shading requires renaming binary components (.so, .dll; for tcnative,
epoll, etc). But the rename then requires setting the
io.netty.packagePrefix system property on the command line or runtime,
which is either a burden or not feasible.

If you don't rename the binary components everything appears to
work, until a dependency on a second version of the binary component is
added. At that point, only one version of the binary will be loaded...
which is what shading is supposed to prevent. So for valid shading, the
binaries must be renamed.

Modifications:

Automatically detect the package prefix by comparing the actual class
name to the non-shaded expected class name. The expected class name must
be obfuscated to prevent shading utilities from changing it.

Result:

When shading and using binary components, runtime configuration is no
longer necessary.

Pre-existing shading users that were not renaming the binary components
will break, because the packagePrefix previously defaulted to "". Since
these pre-existing users had broken configurations that only _appeared_
to work, this breakage is considered a Good Thing. Users may workaround
this breakage temporarily by setting -Dio.netty.packagePrefix= to
restore packagePrefix to "".

Fixes #6963

----

I have manually tested this with my in-progress shading in gRPC. The default was detected and behaved the same as specifying `io.netty.packagePrefix` before and after this change. Specifying `-Dio.netty.packagePrefix=` overrode the default. My shading config is still a bit broken, so I'm getting some cipher suite failures, but those were pre-existing and unrelated to defaulting the packagePrefix.

I could have returned `""` instead of throwing `UnsatisfiedLinkError` when the prefix is un-detected, but felt it was better to fail early instead of letting the binary component's linking to fail.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
+++ b/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
@@ -181,11 +181,31 @@ public final class NativeLibraryLoader {
     }
 
     /**
+     * The shading prefix added to this class's full name.
+     *
+     * @throws UnsatisfiedLinkError if the shader used something other than a prefix
+     */
+    private static String calculatePackagePrefix() {
+        String maybeShaded = NativeLibraryLoader.class.getName();
+        // Use ! instead of . to avoid shading utilities from modifying the string
+        String expected = "io!netty!util!internal!NativeLibraryLoader".replace('!', '.');
+        if (!maybeShaded.endsWith(expected)) {
+            throw new UnsatisfiedLinkError(String.format(
+                    "Could not find prefix added to %s to get %s. When shading, only adding a "
+                    + "package prefix is supported", expected, maybeShaded));
+        }
+        return maybeShaded.substring(0, maybeShaded.length() - expected.length());
+    }
+
+    /**
      * Load the given library with the specified {@link ClassLoader}
      */
     public static void load(String originalName, ClassLoader loader) {
         // Adjust expected name to support shading of native libraries.
-        String name = SystemPropertyUtil.get("io.netty.packagePrefix", "").replace('.', '-') + originalName;
+        String implicitPackagePrefix = calculatePackagePrefix();
+        // The system property should not be necessary; it can be removed in the future.
+        String packagePrefix = SystemPropertyUtil.get("io.netty.packagePrefix", implicitPackagePrefix);
+        String name = packagePrefix.replace('.', '-') + originalName;
 
         String libname = System.mapLibraryName(name);
         String path = NATIVE_RESOURCE_HOME + libname;


Scottmitch:
4.1 (e5a31a428) 4.0 (a54c2a7ae)

Scottmitch:
thanks @ejona86!

We will remove the system property after this gets included in a release.

PR #6996, opened 2017-07-20
Explicitly cast from unsigned ints to ints in Netty Epoll

carl-mastrangelo:
Motivation:

`SetIntArrayRegion` takes an array of `jint`s as its last arg, but
netty is passing in an array of unsigned ints.  This makes the
linter unhappy, and complains when `-Werror=pointer-sign` is on.

Java doesn't really have unsigned ints (except for chars) so
it doesn't really make sense to pass unsigned ints back up.

Modifications:

Explicitly cast from unsigned ints to jint.

Result:
Less naggy linter.


transport-native-epoll/src/main/c/netty_epoll_linuxsocket.c
```diff
@@ -183,39 +183,39 @@ static void netty_epoll_linuxsocket_getTcpInfo(JNIEnv* env, jclass clazz, jint f
...
+     cArray[11] = (jint) tcp_info.tcpi_rcv_mss;
+     cArray[12] = (jint) tcp_info.tcpi_unacked;
+     cArray[13] = (jint) tcp_info.tcpi_sacked;
+     cArray[14] = (jint) tcp_info.tcpi_lost;
+     cArray[15] = (jint) tcp_info.tcpi_retrans;
+     cArray[16] = (jint) tcp_info.tcpi_fackets;
+     cArray[17] = (jint) tcp_info.tcpi_last_data_sent;
+     cArray[18] = (jint) tcp_info.tcpi_last_ack_sent;
+     cArray[19] = (jint) tcp_info.tcpi_last_data_recv;
+     cArray[20] = (jint) tcp_info.tcpi_last_ack_recv;
```
Scottmitch: @normanmaurer - can you clarify why this was not an array of `long` or why we don't return a class which mirrors the `tcp_info` struct (I'm assuming this has to do with object allocation overhead).

normanmaurer: because int[] is good enough and no long is needed. Also yes a class was not needed here as well.

Scottmitch: > int[] is good enough

There are unsigned values. Is it expected users of this API will know the `int` is really unsigned and do the `long` conversion them self, or do we somehow know these values never use the MSB?

normanmaurer: actually you are right these may hold values that will be truncated when cast to int. That said I think we can not fix this as we would need to change the return value of multiple methods that are public. We could deprecate these tho and add new methods that return a long.

carl-mastrangelo: I can make that change, but it would be API breaking in EpollTcpInfo.  Java is going to complain that users are trying to put a long in an int.

normanmaurer: let me have a look how we can "fix this without breaking stuff".

carl-mastrangelo: fwiw, casting them to int is not truncating them it just treats them as negative.  I view this as a java shortcoming and not something readily fixable.  At Google we use the Guava classes [UnsignedInts](http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/primitives/UnsignedInts.html) and [UnsignedInteger](http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/primitives/UnsignedInteger.html) to express that a number may appear negative.  

Using Guava is probably not possible for you, but it does mean consumers of Netty APIs have a way of interpretting the signed ints that come from tcp info.

Also, I don't think using longs really solves this.  Consumers still have to do bounds checking on the result, and arithmetic with longs and preexisting ints is more confusing than staying entirely with ints.

Scottmitch: Agreed ... the data is not lost ... one approach to "fix" this is to just to clarify expectations on the API. For example saying "some values are unsigned" and a link in the javadocs to the `tcp_info` structure in some linux header file maybe "good enough".

 > Consumers still have to do bounds checking on the result

In general yes ... long can still be negative ... however practically the underlying unsigned 32 bits will not overflow so the API can clarify the valid range of these values and practically users can depend upon these constraints instead of always having to bounds check.

normanmaurer:
@carl-mastrangelo I think this is a proper fix which also fix the problem with truncated values (without API breakage) : https://github.com/netty/netty/pull/7011

transport-native-epoll/src/main/c/netty_epoll_linuxsocket.c
normanmaurer: See https://github.com/netty/netty/pull/7011... This should fix it without api breakage as well.

normanmaurer:
Let me close this as https://github.com/netty/netty/pull/7011 fixes it probably

Issue #3927, opened 2015-07-01
FileRegion support in HTTP/2


Apache9:
I'm currently working on [HDFS-7966 New Data Transfer Protocol via HTTP/2](https://issues.apache.org/jira/browse/HDFS-7966).

The code is based on netty and we find that the data frame only accepts ByteBuf as its content so we can not use FileRegion to send block data.

I know that one can not use FileRegion if there is an SslHandler in the pipeline, but netty still supports clear text HTTP/2 protocol right? So is there any other reason that we do not support FileRegion in HTTP/2?

Thanks.

nmittler:
The entire HTTP/2 stack (as was the case for SPDY) is `ByteBuf` based.  I'm not sure how we could support `FileRegion` without significant refactoring.  Or did you have some ideas?

Apache9:
I could have a try to see if we can support it with acceptable modifications on the existing code.

Thanks.

Apache9:
I have developed a proof-of-concept(#3928). Two unknown problems:
1. FileRegion does not support slice, so I write an ugly split method...
2. All 'size's are int in the original code, but FileRegion's count could be long. I just cast the count to int in the POC implementation.

Scottmitch:
@nmittler - I'll assign to you for now ... it seems like you have been following this more closely with your POCs PRs.

Scottmitch:
@Apache9 @nmittler - There doesn't seem to be much movement on this issue recently ... I'll move this to 4.1.1 for now

Apache9:
Sorry about this...
I tried in #4611 and @nmittler thought it is better to have a common interface for `ByteBuf` and `FileRegion` whose index is int based. I explained the difficulty and haven't seen any progress after that...

nmittler:
@Apache9 @Scottmitch apologies guys .. I haven't had the bandwidth to get back to this and probably won't any time soon.

If someone else would want to take this on, the original interface that I sketched out for `ReadableObject` seems like reasonable start (#3965 and #3969). It will require some work, but I don't think it's incredibly hard. The main changes would be:

- Use int-based indices rather than long. This will avoid any potential loss of VM optimizations for looping, etc., as well as minimize the impact on current code.
[...]

PR #6997, opened 2017-07-20
Add slice support for FileRegion.

Apache9:
Motivation:
Let codec such as HTTP/2 can write FileRegion chunk by chunk.

Modifications:
Add a transferIndex which is similar to readerIndex in ByteBuf for FileRegion.
Add two transferBytesTo methods which are similar to readBytes in ByteBuf for FileRegion.
Add slice and transferSlice methods which are similar to slice and readSlice in ByteBuf for FileRegion.
Make the return value of position to be the offset related to the FileChannel returned by channel method. And for DefaultFileRegion, the return value of the original position method is the offset related to the underlying FileChannel, so this does not break compatibility.
The new methods are added to a SliceableFileRegion interface which extends the old FileRegion interface, will move the new methods to FileRegion at the next minor release.
Add a FileChannelBasedFileRegion which makes it easier for implementation file channel related file regions.
Add a FileChannelBasedSlicedFileRegion which is package private.
Add FileChannelBasedSlicedFileRegionTest.
The native transport has not been modified yet so it can not accept the FileChannelBasedSlicedFileRegion right now, will address it in another PR.

Result:
Passed all FileRegion related testsuites.

Fixes
The first step of #3927

### DIFF
--- a/transport/src/main/java/io/netty/channel/DefaultFileRegion.java
+++ b/transport/src/main/java/io/netty/channel/DefaultFileRegion.java
@@ -17,6 +17,7 @@ package io.netty.channel;
 
 import io.netty.util.AbstractReferenceCounted;
 import io.netty.util.IllegalReferenceCountException;
+import io.netty.util.ReferenceCounted;
 import io.netty.util.internal.logging.InternalLogger;
 import io.netty.util.internal.logging.InternalLoggerFactory;
 
@@ -32,7 +33,7 @@ import java.nio.channels.WritableByteChannel;
  * Be aware that the {@link FileChannel} will be automatically closed once {@link #refCnt()} returns
  * {@code 0}.
  */
-public class DefaultFileRegion extends AbstractReferenceCounted implements FileRegion {
+public class DefaultFileRegion extends FileChannelBasedFileRegion implements FileRegion {
 
     private static final InternalLogger logger = InternalLoggerFactory.getInstance(DefaultFileRegion.class);
     private final File f;
@@ -87,16 +88,12 @@ public class DefaultFileRegion extends AbstractReferenceCounted implements FileR
         this.f = f;
     }
 
-    /**
-     * Returns {@code true} if the {@link FileRegion} has a open file-descriptor
-     */
+    @Override
     public boolean isOpen() {
         return file != null;
     }
 
-    /**
-     * Explicitly open the underlying file-descriptor if not done yet.
-     */
+    @Override
     public void open() throws IOException {
         if (!isOpen() && refCnt() > 0) {
             // Only open if this DefaultFileRegion was not released yet.
@@ -120,26 +117,27 @@ public class DefaultFileRegion extends AbstractReferenceCounted implements FileR
         return transferred;
     }
 
+    @Deprecated
     @Override
     public long transferred() {
         return transferred;
     }
 
+    @Deprecated
     @Override
     public long transferTo(WritableByteChannel target, long position) throws IOException {
         long count = this.count - position;
         if (count < 0 || position < 0) {
-            throw new IllegalArgumentException(
-                    "position out of range: " + position +
-                    " (expected: 0 - " + (this.count - 1) + ')');
+            throw new IllegalArgumentException(String.format(
+                    "position out of range: %d (expected: 0 - %d)", position, this.count - 1));
         }
         if (count == 0) {
-            return 0L;
+            return 0;
         }
         if (refCnt() == 0) {
             throw new IllegalReferenceCountException(0);
         }
-        // Call open to make sure fc is initialized. This is a no-oop if we called it before.
+        // Call open to make sure fc is initialized. This is a no-op if we called it before.
         open();
 
         long written = file.transferTo(this.position + position, count, target);
@@ -149,8 +147,20 @@ public class DefaultFileRegion extends AbstractReferenceCounted implements FileR
         return written;
     }
 
-    @Override
-    protected void deallocate() {
+    private final AbstractReferenceCounted refCnt = new AbstractReferenceCounted() {
+
+        @Override
+        public ReferenceCounted touch(Object hint) {
+            return DefaultFileRegion.this;
+        }
+
+        @Override
+        protected void deallocate() {
+            close();
+        }
+    };
+
+    private void close() {
         FileChannel file = this.file;
 
         if (file == null) {
@@ -168,24 +178,51 @@ public class DefaultFileRegion extends AbstractReferenceCounted implements FileR
     }
 
     @Override
+    public int refCnt() {
+        return refCnt.refCnt();
+    }
+
+    @Override
+    public boolean release() {
+        return refCnt.release();
+    }
+
+    @Override
+    public boolean release(int decrement) {
+        return refCnt.release(decrement);
+    }
+
+    @Override
     public FileRegion retain() {
-        super.retain();
+        refCnt.retain();
         return this;
     }
 
     @Override
     public FileRegion retain(int increment) {
-        super.retain(increment);
+        refCnt.retain(increment);
         return this;
     }
 
     @Override
     public FileRegion touch() {
+        refCnt.touch();
         return this;
     }
 
     @Override
     public FileRegion touch(Object hint) {
+        refCnt.touch(hint);
         return this;
+    }
+
+    @Override
+    public SliceableFileRegion unwrap() {
+        return null;
+    }
+
+    @Override
+    protected FileChannel channel() {
+        return file;
     }
 }
--- a/transport/src/main/java/io/netty/channel/FileChannelBasedFileRegion.java
+++ b/transport/src/main/java/io/netty/channel/FileChannelBasedFileRegion.java
@@ -1,0 +1,152 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.channel;
+
+import io.netty.util.IllegalReferenceCountException;
+
+import java.io.IOException;
+import java.nio.channels.FileChannel;
+import java.nio.channels.WritableByteChannel;
+
+/**
+ * Base class for {@link FileRegion} implementations which based on a {@link FileChannel}.
+ */
+public abstract class FileChannelBasedFileRegion implements SliceableFileRegion {
+
+    private long transferIndex;
+    private long transferred;
+
+    @Override
+    public long transferBytesTo(WritableByteChannel target, long length) throws IOException {
+        long written = transferBytesTo(target, transferIndex, length);
+        if (written > 0) {
+            transferIndex += written;
+        }
+        return written;
+    }
+
+    @Override
+    public long transferIndex() {
+        return transferIndex;
+    }
+
+    @Override
+    public FileChannelBasedFileRegion transferIndex(long index) {
+        if (index < 0 || index > count()) {
+            throw new IndexOutOfBoundsException(
+                    String.format("transferIndex: %d (expected: 0 <= transferIndex <= count(%d))",
+                            index, count()));
+        }
+        this.transferIndex = index;
+        return this;
+    }
+
+    @Override
+    public boolean isTransferable() {
+        return transferIndex < count();
+    }
+
+    @Override
+    public long transferableBytes() {
+        return count() - transferIndex;
+    }
+
+    @Override
+    public long transferBytesTo(WritableByteChannel target, long position, long length) throws IOException {
+        if (position < 0 || position > count()) {
+            throw new IllegalArgumentException(String
+                    .format("position out of range: %d (expected: 0 - %d)", position, count() - 1));
+        }
+        if (length < 0) {
+            throw new IllegalArgumentException("negative length " + length);
+        }
+        length = Math.min(count() - position, length);
+        if (length == 0) {
+            return 0;
+        }
+        if (refCnt() == 0) {
+            throw new IllegalReferenceCountException(0);
+        }
+        // Call open to make sure fc is initialized. This is a no-op if we called it before.
+        open();
+        return channel().transferTo(position() + position, length, target);
+    }
+
+    @Override
+    public SliceableFileRegion slice(long index, long length) {
+        return new FileChannelBasedSlicedFileRegion(this, index, length);
+    }
+
+    @Override
+    public SliceableFileRegion retainedSlice(long index, long length) {
+        SliceableFileRegion sliced = slice(index, length);
+        sliced.retain();
+        return sliced;
+    }
+
+    @Override
+    public SliceableFileRegion transferSlice(long length) {
+        SliceableFileRegion sliced = slice(transferIndex(), length);
+        transferIndex += length;
+        return sliced;
+    }
+
+    @Override
+    public SliceableFileRegion transferRetainedSlice(long length) {
+        SliceableFileRegion sliced = transferSlice(length);
+        sliced.retain();
+        return sliced;
+    }
+
+    /**
+     * Explicitly open the underlying file-descriptor if not done yet.
+     * <p>
+     * The implementation should be idempotent.
+     */
+    public abstract void open() throws IOException;
+
+    /**
+     * Returns {@code true} if the {@link FileChannelBasedFileRegion} has a open file-descriptor
+     */
+    public abstract boolean isOpen();
+
+    /**
+     * Return the underlying file channel instance of this file region. Will return {@code null} if
+     * the underlying file channel has not been opened yet.
+     */
+    protected abstract FileChannel channel();
+
+    // The methods below are only used to keep compatible with old APIs.
+    @Override
+    public long transfered() {
+        return transferred;
+    }
+
+    @Override
+    public long transferred() {
+        return transferred;
+    }
+
+    @Override
+    public long transferTo(WritableByteChannel target, long position) throws IOException {
+        // pass Long.MAX_VALUE to indicate that transfer as much as possible.
+        long written = transferBytesTo(target, position, Long.MAX_VALUE);
+        if (written > 0) {
+            transferred += written;
+        }
+        return written;
+    }
+}
--- a/transport/src/main/java/io/netty/channel/FileChannelBasedSlicedFileRegion.java
+++ b/transport/src/main/java/io/netty/channel/FileChannelBasedSlicedFileRegion.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.channel;
+
+import java.io.IOException;
+import java.nio.channels.FileChannel;
+
+/**
+ * A derived {@link FileChannelBasedFileRegion} which shares reference counting with its parent.
+ */
+final class FileChannelBasedSlicedFileRegion extends FileChannelBasedFileRegion {
+
+    private final FileChannelBasedFileRegion parent;
+    private final long offset;
+    private final long count;
+
+    FileChannelBasedSlicedFileRegion(FileChannelBasedFileRegion parent, long index, long count) {
+        if (index < 0 || index > parent.count() - count) {
+            throw new IndexOutOfBoundsException(parent + ".slice(" + index + ", " + count + ')');
+        }
+
+        if (parent instanceof FileChannelBasedSlicedFileRegion) {
+            FileChannelBasedSlicedFileRegion sliced = (FileChannelBasedSlicedFileRegion) parent;
+            this.parent = sliced.parent;
+            this.offset = sliced.offset + index;
+        } else {
+            this.parent = parent;
+            this.offset = index;
+        }
+        this.count = count;
+    }
+
+    @Override
+    public long position() {
+        return parent.position() + offset;
+    }
+
+    @Override
+    public long count() {
+        return count;
+    }
+
+    @Override
+    public FileChannelBasedSlicedFileRegion retain() {
+        parent.retain();
+        return this;
+    }
+
+    @Override
+    public FileChannelBasedSlicedFileRegion retain(int increment) {
+        parent.retain(increment);
+        return this;
+    }
+
+    @Override
+    public FileChannelBasedSlicedFileRegion touch() {
+        parent.touch();
+        return this;
+    }
+
+    @Override
+    public FileChannelBasedSlicedFileRegion touch(Object hint) {
+        parent.touch(hint);
+        return this;
+    }
+
+    @Override
+    public int refCnt() {
+        return parent.refCnt();
+    }
+
+    @Override
+    public boolean release() {
+        return parent.release();
+    }
+
+    @Override
+    public boolean release(int decrement) {
+        return parent.release(decrement);
+    }
+
+    @Override
+    public FileChannelBasedFileRegion unwrap() {
+        return parent;
+    }
+
+    @Override
+    public boolean isOpen() {
+        return parent.isOpen();
+    }
+
+    @Override
+    public void open() throws IOException {
+        parent.open();
+    }
+
+    @Override
+    protected FileChannel channel() {
+        return parent.channel();
+    }
+}
--- a/transport/src/main/java/io/netty/channel/FileRegion.java
+++ b/transport/src/main/java/io/netty/channel/FileRegion.java
@@ -67,7 +67,12 @@ public interface FileRegion extends ReferenceCounted {
 
     /**
      * Returns the bytes which was transfered already.
+     *
+     * @deprecated Will be removed in the next minor release of netty. Please use the new methods added in
+     *             {@link FileChannelBasedFileRegion} instead and they will be moved to this interface in the next minor
+     *             release.
      */
+    @Deprecated
     long transferred();
 
     /**
@@ -84,7 +89,11 @@ public interface FileRegion extends ReferenceCounted {
      *                  transfer start from {@link #position()}th byte and
      *                  <tt>{@link #count()} - 1</tt> will make the last
      *                  byte of the region transferred.
+     * @deprecated Will be removed in the next minor release of netty. Please use the new methods added in
+     *             {@link FileChannelBasedFileRegion} instead and they will be moved to this interface in the next minor
+     *             release.
      */
+    @Deprecated
     long transferTo(WritableByteChannel target, long position) throws IOException;
 
     @Override
--- a/transport/src/main/java/io/netty/channel/SliceableFileRegion.java
+++ b/transport/src/main/java/io/netty/channel/SliceableFileRegion.java
@@ -1,0 +1,131 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.channel;
+
+import java.io.IOException;
+import java.nio.channels.WritableByteChannel;
+
+/**
+ * A {@link FileRegion} which supports slice.
+ * <p>
+ * We introduce this interface because we can not add new methods to an interface until next
+ * minor release. All the methods here will be moved to FileRegion when we release netty 4.2.
+ */
+public interface SliceableFileRegion extends FileRegion {
+
+    /**
+     * Returns the {@code transferIndex} of this file region.
+     */
+    long transferIndex();
+
+    /**
+     * Sets the {@code transferIndex} of file region.
+     *
+     * @throws IndexOutOfBoundsException
+     *             if the specified {@code transferIndex} is less than {@code 0} or greater than
+     *             {@link #count()}
+     */
+    SliceableFileRegion transferIndex(long index);
+
+    /**
+     * Returns {@code true} if and only if {@code (this.count - this.transferIndex)} is greater than
+     * {@code 0}.
+     */
+    boolean isTransferable();
+
+    /**
+     * Returns the number of readable bytes which is equal to {@code (count - transferIndex)}.
+     */
+    long transferableBytes();
+
+    /**
+     * Transfers the content of this file region to the specified channel.
+     *
+     * @param target
+     *            the destination of the transfer
+     * @param length
+     *            the maximum number of bytes to transfer
+     */
+    long transferBytesTo(WritableByteChannel target, long length) throws IOException;
+
+    /**
+     * Transfers the content of this file region to the specified channel.
+     * <p>
+     * This method does not modify {@link #transferIndex()}.
+     *
+     * @param target
+     *            the destination of the transfer
+     * @param position
+     *            the relative offset of the file where the transfer begins from. For example,
+     *            <tt>0</tt> will make the transfer start from {@link #position()}th byte and
+     *            <tt>{@link #count()} - 1</tt> will make the last byte of the region transferred.
+     * @param length
+     *            the maximum number of bytes to transfer
+     */
+    long transferBytesTo(WritableByteChannel target, long position, long length) throws IOException;
+
+    /**
+     * Returns a slice of this file region's sub-region. This method does not modify
+     * {@code transferIndex} of this buffer.
+     * <p>
+     * Also be aware that this method will NOT call {@link #retain()} and so the reference count
+     * will NOT be increased.
+     */
+    SliceableFileRegion slice(long index, long length);
+
+    /**
+     * Returns a slice of this file region's sub-region. This method does not modify
+     * {@code transferIndex} of this buffer.
+     * <p>
+     * Note that this method returns a {@linkplain #retain() retained} file region unlike
+     * {@link #slice(long, long)}.
+     */
+    SliceableFileRegion retainedSlice(long index, long length);
+
+    /**
+     * Returns a new slice of this file region's sub-region starting at the current
+     * {@code transferIndex} and increases the {@code transferIndex} by the size of the new slice (=
+     * {@code length}).
+     * <p>
+     * Also be aware that this method will NOT call {@link #retain()} and so the reference count
+     * will NOT be increased.
+     *
+     * @param length
+     *            the size of the new slice
+     */
+    SliceableFileRegion transferSlice(long length);
+
+    /**
+     * Returns a new slice of this file region's sub-region starting at the current
+     * {@code transferIndex} and increases the {@code transferIndex} by the size of the new slice (=
+     * {@code length}).
+     * <p>
+     * Note that this method returns a {@linkplain #retain() retained} file region unlike
+     * {@link #transferSlice(long)}.
+     *
+     * @param length
+     *            the size of the new slice
+     */
+    SliceableFileRegion transferRetainedSlice(long length);
+
+    /**
+     * Return the underlying file region instance if this file region is a wrapper of another file
+     * region.
+     *
+     * @return {@code null} if this file region is not a wrapper
+     */
+    SliceableFileRegion unwrap();
+}
--- a/transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
+++ b/transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
@@ -26,6 +26,7 @@ import io.netty.channel.ChannelOutboundBuffer;
 import io.netty.channel.ChannelPipeline;
 import io.netty.channel.FileRegion;
 import io.netty.channel.RecvByteBufAllocator;
+import io.netty.channel.SliceableFileRegion;
 import io.netty.channel.socket.ChannelInputShutdownEvent;
 import io.netty.channel.socket.ChannelInputShutdownReadComplete;
 import io.netty.util.internal.StringUtil;
@@ -157,12 +158,70 @@ public abstract class AbstractNioByteChannel extends AbstractNioChannel {
         }
     }
 
+    private enum WriteStatus {
+        SET_OP_WRITE, // We have not written out the whole message and output channel is not writable
+        INCOMPLETE, // We have not written out the whole message after the spin count
+        DONE // we have written out the whole message
+    }
+
+    private WriteStatus doWriteFileRegion(ChannelOutboundBuffer in, FileRegion region,
+            int writeSpinCount) throws Exception {
+        if (region instanceof SliceableFileRegion) {
+            SliceableFileRegion sliceableRegion = (SliceableFileRegion) region;
+            if (!sliceableRegion.isTransferable()) {
+                return WriteStatus.DONE;
+            }
+            WriteStatus status = WriteStatus.INCOMPLETE;
+            long flushedAmount = 0;
+
+            for (int i = writeSpinCount - 1; i >= 0; i--) {
+                long localFlushedAmount = doWriteFileRegion(sliceableRegion);
+                if (localFlushedAmount == 0) {
+                    status = WriteStatus.SET_OP_WRITE;
+                    break;
+                }
+
+                flushedAmount += localFlushedAmount;
+                if (!sliceableRegion.isTransferable()) {
+                    status = WriteStatus.DONE;
+                    break;
+                }
+            }
+            in.progress(flushedAmount);
+            return status;
+        } else {
+            // TODO: consider remove this branch once we move the methods in SliceableFileRegion
+            // into FileRegion in the next minor release.
+            if (region.transferred() >= region.count()) {
+                return WriteStatus.DONE;
+            }
+            WriteStatus status = WriteStatus.INCOMPLETE;
+            long flushedAmount = 0;
+
+            for (int i = writeSpinCount - 1; i >= 0; i--) {
+                long localFlushedAmount = doWriteFileRegion(region);
+                if (localFlushedAmount == 0) {
+                    status = WriteStatus.SET_OP_WRITE;
+                    break;
+                }
+
+                flushedAmount += localFlushedAmount;
+                if (region.transferred() >= region.count()) {
+                    status = WriteStatus.DONE;
+                    break;
+                }
+            }
+            in.progress(flushedAmount);
+            return status;
+        }
+    }
+
     @Override
     protected void doWrite(ChannelOutboundBuffer in) throws Exception {
         int writeSpinCount = -1;
 
         boolean setOpWrite = false;
-        for (;;) {
+        loop: for (;;) {
             Object msg = in.current();
             if (msg == null) {
                 // Wrote all messages.
@@ -208,36 +267,23 @@ public abstract class AbstractNioByteChannel extends AbstractNioChannel {
                 }
             } else if (msg instanceof FileRegion) {
                 FileRegion region = (FileRegion) msg;
-                boolean done = region.transferred() >= region.count();
-
-                if (!done) {
-                    long flushedAmount = 0;
-                    if (writeSpinCount == -1) {
-                        writeSpinCount = config().getWriteSpinCount();
-                    }
-
-                    for (int i = writeSpinCount - 1; i >= 0; i--) {
-                        long localFlushedAmount = doWriteFileRegion(region);
-                        if (localFlushedAmount == 0) {
-                            setOpWrite = true;
-                            break;
-                        }
-
-                        flushedAmount += localFlushedAmount;
-                        if (region.transferred() >= region.count()) {
-                            done = true;
-                            break;
-                        }
-                    }
-
-                    in.progress(flushedAmount);
+                if (writeSpinCount == -1) {
+                    writeSpinCount = config().getWriteSpinCount();
                 }
-
-                if (done) {
-                    in.remove();
-                } else {
-                    // Break the loop and so incompleteWrite(...) is called.
-                    break;
+                switch (doWriteFileRegion(in, region, writeSpinCount)) {
+                    case SET_OP_WRITE:
+                        setOpWrite = true;
+                        // the fall through is intentional, setOpWrite also means we have not
+                        // written out the whole message.
+                    case INCOMPLETE:
+                        // Break the loop and so incompleteWrite(...) is called.
+                        break loop;
+                    case DONE:
+                        in.remove();
+                        break;
+                    default:
+                        // Should not reach here.
+                        throw new Error();
                 }
             } else {
                 // Should not reach here.
--- a/transport/src/main/java/io/netty/channel/oio/AbstractOioByteChannel.java
+++ b/transport/src/main/java/io/netty/channel/oio/AbstractOioByteChannel.java
@@ -26,6 +26,7 @@ import io.netty.channel.ChannelOutboundBuffer;
 import io.netty.channel.ChannelPipeline;
 import io.netty.channel.FileRegion;
 import io.netty.channel.RecvByteBufAllocator;
+import io.netty.channel.SliceableFileRegion;
 import io.netty.channel.socket.ChannelInputShutdownEvent;
 import io.netty.util.internal.StringUtil;
 
@@ -203,7 +204,15 @@ public abstract class AbstractOioByteChannel extends AbstractOioChannel {
                     readableBytes = newReadableBytes;
                 }
                 in.remove();
+            } else if (msg instanceof SliceableFileRegion) {
+                SliceableFileRegion region = (SliceableFileRegion) msg;
+                long transferIndex = region.transferIndex();
+                doWriteFileRegion(region);
+                in.progress(region.transferIndex() - transferIndex);
+                in.remove();
             } else if (msg instanceof FileRegion) {
+                // TODO: consider remove this branch once we move the methods in SliceableFileRegion
+                // into FileRegion in the next minor release.
                 FileRegion region = (FileRegion) msg;
                 long transferred = region.transferred();
                 doWriteFileRegion(region);
--- a/transport/src/main/java/io/netty/channel/oio/OioByteStreamChannel.java
+++ b/transport/src/main/java/io/netty/channel/oio/OioByteStreamChannel.java
@@ -19,6 +19,7 @@ import io.netty.buffer.ByteBuf;
 import io.netty.channel.Channel;
 import io.netty.channel.FileRegion;
 import io.netty.channel.RecvByteBufAllocator;
+import io.netty.channel.SliceableFileRegion;
 
 import java.io.EOFException;
 import java.io.IOException;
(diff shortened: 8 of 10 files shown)


Apache9:
The first step of #3927

Apache9:
ping @Scottmitch @nmittler @normanmaurer . Thanks.

transport/src/main/java/io/netty/channel/FileRegion.java
```diff
@@ -84,9 +86,86 @@
...
+     *
+     * @param target    the destination of the transfer
+     * @param position  the relative offset of the file where the transfer
+     *                  begins from.  For example, <tt>0</tt> will make the
+     *                  transfer start from {@link #position()}th byte and
+     *                  <tt>{@link #count()} - 1</tt> will make the last
+     *                  byte of the region transferred.
+     * @param length    the maximum number of bytes to transfer
+     */
+    long transferBytesTo(WritableByteChannel target, long position, long length) throws IOException;
```
normanmaurer: We can not add methods to the interface until we release a new minor / major version. So please remove this and just check for the actual implementation with an instanceof when you need these methods.

transport/src/main/java/io/netty/channel/FileRegion.java
```diff
@@ -84,9 +86,86 @@
...
+     * Return the underlying file region instance if this file region is a wrapper of another file region.
+     *
+     * @return {@code null} if this file region is not a wrapper
+     */
+    FileRegion unwrap();
+
+    /**
+     * Return the underlying file channel instance of this file region.
+     */
+    FileChannel channel() throws IOException;
```
normanmaurer: see above also I think this should be not part of the `FileRegion` interface at all as the FileRegion should not be specific to `FileChannel`.

transport/src/main/java/io/netty/channel/FileRegion.java
normanmaurer: This is true for all new methods on the interface

transport/src/main/java/io/netty/channel/DefaultFileRegion.java
```diff
@@ -168,24 +183,55 @@ protected void deallocate() {
...
+    @Override
+    public DefaultFileRegion retain() {
+        refCnt.retain();
         return this;
     }
 
     @Override
-    public FileRegion retain(int increment) {
-        super.retain(increment);
+    public DefaultFileRegion retain(int increment) {
```
normanmaurer: we can not change the return type as this would be a breaking change.

transport/src/main/java/io/netty/channel/DefaultFileRegion.java
```diff
@@ -168,24 +183,55 @@ protected void deallocate() {
...
     @Override
-    public FileRegion touch() {
+    public DefaultFileRegion touch() {
+        refCnt.touch();
         return this;
     }
 
     @Override
-    public FileRegion touch(Object hint) {
+    public DefaultFileRegion touch(Object hint) {
```
normanmaurer: we can not change the return type as this is a breaking change

transport/src/test/java/io/netty/channel/SlicedFileRegionTest.java
```diff
@@ -0,0 +1,105 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Test for {@link SlicedFileRegion}.
+ */
+public class SlicedFileRegionTest {
+
+    private static byte[] TEST_DATA = new byte[1024 * 64];
```
normanmaurer: final

transport/src/test/java/io/netty/channel/SlicedFileRegionTest.java
```diff
@@ -0,0 +1,105 @@
...
+        FileOutputStream out = new FileOutputStream(TEST_FILE);
+        try {
+            out.write(TEST_DATA);
+        } finally {
+            out.close();
+        }
+    }
+
+    @Test
+    public void test() throws IOException {
```
normanmaurer: use a more meaningful name.

transport/src/main/java/io/netty/channel/SlicedFileRegion.java
```diff
@@ -0,0 +1,122 @@
...
+package io.netty.channel;
+
+import java.io.IOException;
+import java.nio.channels.FileChannel;
+import java.nio.channels.WritableByteChannel;
+
+/**
+ * A derived {@link FileRegion} which shares reference counting with its parent.
+ */
+public class SlicedFileRegion extends AbstractFileRegion {
```
normanmaurer: make this final and package private for now as its an implementation detail.

transport/src/main/java/io/netty/channel/FileRegion.java
```diff
@@ -84,9 +86,86 @@
...
+     */
+    long transferableBytes();
+
+    /**
+     * Returns a slice of this file region's sub-region. This method does not modify {@code transferIndex} of this
+     * buffer.
+     * <p>
+     * Also be aware that this method will NOT call {@link #retain()} and so the reference count will NOT be increased.
+     */
+    FileRegion slice(long index, long length);
```
normanmaurer: can you explain what the slices are good for ?

Apache9: For splitting big chunk into data frames in HTTP/2 implementation.

normanmaurer: ok.

transport/src/main/java/io/netty/channel/FileRegion.java
Apache9: OK. So it is free for me to add a new AbstractFileRegion class and make DefaultFileRegion implement it, and until netty 4.2 release I can move the new methods in AbstractFileRegion up to the interface?

normanmaurer: yep I think this is fine. (Please also add a todo to move to interface at some point.

Apache9: Got it. Thanks for the explaination. Will fix this soon.

transport/src/main/java/io/netty/channel/DefaultFileRegion.java
Apache9: One more question, is this because we may break the code which extends DefaultFileRegion, or that we want to maintain drop in replacement between patch releases of netty since the JVM may include the return value to generate the method signature? I ask this because the Native.sendfile uses DefaultFileRegion as the parameter and I want to change it to AbstractFileRegion if possible in the follow on PRs which add native support for SlicedFileRegion.

normanmaurer: its because the user may extend it.

Apache9: Got it. Thanks.

Apache9:
@normanmaurer I've modified the code based on your comments. Now the new methods are only in AbstractFileRegion. In the XXXByteStreamChannel implementations I've added a new branch to handle AbstractFileRegion and also added a comment to say that please remove the branch once we release netty 4.2 and move the methods in AbstractFileRegion to FileRegion.

transport/src/main/java/io/netty/channel/oio/OioByteStreamChannel.java
```diff
@@ -128,21 +129,44 @@ protected void doWriteFileRegion(FileRegion region) throws Exception {
...
+                if (!r.isTransferable()) {
+                    return;
+                }
             }
-            written += localWritten;
-
-            if (written >= region.count()) {
-                return;
+        } else {
+            // TODO: consider remove this branch once we move the methods in AbstractFileRegion
```
netkins: INFO Complete the task associated to this TODO comment. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1135)

transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
```diff
@@ -206,7 +207,42 @@ protected void doWrite(ChannelOutboundBuffer in) throws Exception {
...
+                }
+                in.progress(flushedAmount);
+                if (done) {
+                    in.remove();
+                } else {
+                    // Break the loop and so incompleteWrite(...) is called.
+                    break;
+                }
             } else if (msg instanceof FileRegion) {
+                // TODO: consider remove this branch once we move the methods in AbstractFileRegion
```
netkins: INFO Complete the task associated to this TODO comment. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1135)

transport/src/main/java/io/netty/channel/socket/nio/NioSocketChannel.java
```diff
@@ -380,8 +381,15 @@ protected int doWriteBytes(ByteBuf buf) throws Exception {
 
     @Override
     protected long doWriteFileRegion(FileRegion region) throws Exception {
-        final long position = region.transferred();
-        return region.transferTo(javaChannel(), position);
+        if (region instanceof AbstractFileRegion) {
+            AbstractFileRegion r = (AbstractFileRegion) region;
+            return r.transferBytesTo(javaChannel(), r.transferableBytes());
+        } else {
+            // TODO: consider remove this branch once we move the methods in AbstractFileRegion
```
netkins: INFO Complete the task associated to this TODO comment. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1135)

transport/src/main/java/io/netty/channel/oio/AbstractOioByteChannel.java
```diff
@@ -203,7 +204,15 @@ protected void doWrite(ChannelOutboundBuffer in) throws Exception {
...
                 }
                 in.remove();
+            } else if (msg instanceof AbstractFileRegion) {
+                AbstractFileRegion region = (AbstractFileRegion) msg;
+                long transferIndex = region.transferIndex();
+                doWriteFileRegion(region);
+                in.progress(region.transferIndex() - transferIndex);
+                in.remove();
             } else if (msg instanceof FileRegion) {
+                // TODO: consider remove this branch once we move the methods in AbstractFileRegion
```
netkins: INFO Complete the task associated to this TODO comment. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1135)

Apache9:
Any other concerns?  @normanmaurer Thanks.

transport/src/main/java/io/netty/channel/AbstractFileRegion.java
```diff
@@ -0,0 +1,184 @@
...
+        }
+        this.transferIndex = index;
+        return this;
+    }
+
+    /**
+     * Returns {@code true} if and only if {@link #transferableBytes()} is greater than {@code 0}.
+     */
+    public boolean isTransferable() {
+        return transferIndex < count();
```
normanmaurer: use `transferableBytes() > 0` so the impl match what you describe in the javadocs

transport/src/main/java/io/netty/channel/AbstractFileRegion.java
```diff
@@ -0,0 +1,184 @@
...
+        if (position < 0 || position > count()) {
+            throw new IllegalArgumentException("position out of range: " + position
+                    + " (expected: 0 - " + (this.count() - 1) + ')');
+        }
+        if (length < 0) {
+            throw new IllegalArgumentException("negative length " + length);
+        }
+        length = Math.min(count() - position, length);
+        if (length == 0) {
+            return 0L;
```
normanmaurer: nit: You can remove the `L`

transport/src/main/java/io/netty/channel/AbstractFileRegion.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    /**
+     * Returns a slice of this file region's sub-region. This method does not modify
+     * {@code transferIndex} of this buffer.
+     * <p>
+     * Also be aware that this method will NOT call {@link #retain()} and so the reference count
+     * will NOT be increased.
+     */
+    public AbstractFileRegion slice(long index, long length) {
```
normanmaurer: This should return a FileRegion imho.

transport/src/main/java/io/netty/channel/AbstractFileRegion.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    /**
+     * Returns a slice of this file region's sub-region. This method does not modify
+     * {@code transferIndex} of this buffer.
+     * <p>
+     * Note that this method returns a {@linkplain #retain() retained} file region unlike
+     * {@link #slice()}.
+     */
+    public AbstractFileRegion retainedSlice(long index, long length) {
```
normanmaurer: This should return a `FileRegion` imho

transport/src/main/java/io/netty/channel/AbstractFileRegion.java
```diff
@@ -0,0 +1,184 @@
...
+     * {@code transferIndex} and increases the {@code transferIndex} by the size of the new slice (=
+     * {@code length}).
+     * <p>
+     * Also be aware that this method will NOT call {@link #retain()} and so the reference count
+     * will NOT be increased.
+     *
+     * @param length
+     *            the size of the new slice
+     */
+    public AbstractFileRegion transferSlice(long length) {
```
normanmaurer: This should return a `FileRegion` imho

transport/src/main/java/io/netty/channel/AbstractFileRegion.java
```diff
@@ -0,0 +1,184 @@
...
+     * {@code transferIndex} and increases the {@code transferIndex} by the size of the new slice (=
+     * {@code length}).
+     * <p>
+     * Note that this method returns a {@linkplain #retain() retained} file region unlike
+     * {@link #transferSlice()}.
+     *
+     * @param length
+     *            the size of the new slice
+     */
+    public AbstractFileRegion transferRetainedSlice(long length) {
```
normanmaurer: This should return a `FileRegion` imho

transport/src/main/java/io/netty/channel/AbstractFileRegion.java
```diff
@@ -0,0 +1,184 @@
...
+        return sliced;
+    }
+
+    /**
+     * Return the underlying file region instance if this file region is a wrapper of another file
+     * region.
+     *
+     * @return {@code null} if this file region is not a wrapper
+     */
+    public abstract AbstractFileRegion unwrap();
```
normanmaurer: This should return a `FileRegion` imho

transport/src/main/java/io/netty/channel/AbstractFileRegion.java
```diff
@@ -0,0 +1,184 @@
...
+
+    /**
+     * Return the underlying file channel instance of this file region.
+     */
+    public abstract FileChannel channel() throws IOException;
```
normanmaurer: I really dislike that we need to expose this at all as a FileRegion should not really depend on a FileChannel. This is an implementation detail.

transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
```diff
@@ -184,7 +185,7 @@ protected void doWrite(ChannelOutboundBuffer in) throws Exception {
                 if (writeSpinCount == -1) {
                     writeSpinCount = config().getWriteSpinCount();
                 }
-                for (int i = writeSpinCount - 1; i >= 0; i --) {
+                for (int i = writeSpinCount - 1; i >= 0; i--) {
```
normanmaurer: revert

transport/src/main/java/io/netty/channel/AbstractFileRegion.java
Apache9: Yeah I think we do not need to move this method to the FileRegion interface. This is just used to simplify the implementation of sub classes, and also for native transport such as epoll and kqueue. So maybe rename AbstractFileRegion to FileChannelBasedFileRegion?

transport/src/main/java/io/netty/channel/AbstractFileRegion.java
Apache9: I just follow the implementation of AbstractByteBuf.  `readableBytes` returns `writerIndex - readerIndex` while `isReadable` returns `writerIndex > readerIndex`. I'm OK with using `transferableBytes() > 0`, just want to make sure that there is no other reasons why ByteBuf use `writerIndex > readerIndex` instead of `readableBytes() > 0`

transport/src/main/java/io/netty/channel/AbstractFileRegion.java
Apache9: Here we need the returned FileRegion also has the newly added method, otherwise we need to do instanceof and cast everywhere after a slice. We can change to `FileRegion` after moving the methods to FileRegion interface?

transport/src/main/java/io/netty/channel/AbstractFileRegion.java
Apache9: I changed the javadoc to 'Returns {@code true} if and only if {@code (this.count - this.transferIndex)} is greater than {@code 0}.'

Apache9:
Introduce a new interface called SliceableFileRegion. The new methods are all added to this interface. And also introduce a FileChannelBasedFileRegion class to place the common code. The `FileChannel channel()` method is marked as protected since in JNI seems we are free to call protected methods.

@normanmaurer  PTAL. Thanks.

Apache9:
Fix checkstyle problem.

Ping @normanmaurer .

Thanks.

transport/src/main/java/io/netty/channel/oio/OioByteStreamChannel.java
```diff
@@ -128,21 +129,44 @@ protected void doWriteFileRegion(FileRegion region) throws Exception {
...
+                if (!r.isTransferable()) {
+                    return;
+                }
             }
-            written += localWritten;
-
-            if (written >= region.count()) {
-                return;
+        } else {
+            // TODO: consider remove this branch once we move the methods in SliceableFileRegion
```
netkins: INFO Complete the task associated to this TODO comment. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1135)

transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
```diff
@@ -206,7 +207,42 @@ protected void doWrite(ChannelOutboundBuffer in) throws Exception {
...
+                }
+                in.progress(flushedAmount);
+                if (done) {
+                    in.remove();
+                } else {
+                    // Break the loop and so incompleteWrite(...) is called.
+                    break;
+                }
             } else if (msg instanceof FileRegion) {
+                // TODO: consider remove this branch once we move the methods in SliceableFileRegion
```
netkins: INFO Complete the task associated to this TODO comment. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1135)

transport/src/main/java/io/netty/channel/socket/nio/NioSocketChannel.java
```diff
@@ -380,8 +381,15 @@ protected int doWriteBytes(ByteBuf buf) throws Exception {
 
     @Override
     protected long doWriteFileRegion(FileRegion region) throws Exception {
-        final long position = region.transferred();
-        return region.transferTo(javaChannel(), position);
+        if (region instanceof SliceableFileRegion) {
+            SliceableFileRegion r = (SliceableFileRegion) region;
+            return r.transferBytesTo(javaChannel(), r.transferableBytes());
+        } else {
+            // TODO: consider remove this branch once we move the methods in SliceableFileRegion
```
netkins: INFO Complete the task associated to this TODO comment. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1135)

transport/src/main/java/io/netty/channel/oio/AbstractOioByteChannel.java
```diff
@@ -203,7 +204,15 @@ protected void doWrite(ChannelOutboundBuffer in) throws Exception {
...
                 }
                 in.remove();
+            } else if (msg instanceof SliceableFileRegion) {
+                SliceableFileRegion region = (SliceableFileRegion) msg;
+                long transferIndex = region.transferIndex();
+                doWriteFileRegion(region);
+                in.progress(region.transferIndex() - transferIndex);
+                in.remove();
             } else if (msg instanceof FileRegion) {
+                // TODO: consider remove this branch once we move the methods in SliceableFileRegion
```
netkins: INFO Complete the task associated to this TODO comment. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1135)

normanmaurer:
@ejona86 @carl-mastrangelo @Scottmitch  PTAL as well

transport/src/main/java/io/netty/channel/SliceableFileRegion.java
```diff
@@ -0,0 +1,131 @@
...
+ */
+package io.netty.channel;
+
+import java.io.IOException;
+import java.nio.channels.WritableByteChannel;
+
+/**
+ * A {@link FileRegion} which supports slice.
+ * <p>
+ * We introduce this interface is only because that we can not add new methods to an interface until
```
ejona86: "is only because that" seems to have a grammar issue. Maybe: "We introduce this interface because we..."

transport/src/main/java/io/netty/channel/FileChannelBasedSlicedFileRegion.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    @Override
+    public long count() {
+        return count;
+    }
+
+    @Override
+    public long transferTo(WritableByteChannel target, long position) throws IOException {
+        throw new UnsupportedOperationException();
```
ejona86: If `transferTo()` and `transferred()` are unsupported, then what's the point of `SliceableFileRegion` extending `FileRegion`? It's simply impossible to use this `SliceableFileRegion` as a `FileRegion`. They could be separate interfaces and although `DefaultFileRegion` could implement both, the slices from it would only implement `SliceableFileRegion`.

ejona86: Why isn't `transferTo()` supported? What would be wrong with this sketch of an implementation:
```java
private final long transferred;

public long transferred() { return transferred; }
￼
public long transferTo(WritableByteChannel target, long position) throws IOException {
  long written = transferBytesTo(target, position, this.count - position);
  this.transferred += written;
  return written;
}
```

Apache9: For now `SliceableFileRegion` just extends `FileRegion`, the intention here is to move the methods in this interface up to `FileRegion` at the next minor release. The old `transferTo` and `trnsferred` methods are marked as deprecated and some new methods are provided. You can either use the old methods, or the new methods, but you can not use them both at the same time. So you mean it is better to declare a completely new interface and remove the old `FileRegion` interface in the future?

Apache9: These methods are expected to be removed in the future and new implementation classes are not required to implement them. If you use the new methods introduced in `SliceableFileRegion` then you need to use the new `transferBytesTo` methods. Of course this is only my expectation, if this is not the correct upgrading method of netty, please tell me. I can modify the code to respect the netty rule. Thanks.

ejona86: > You can either use the old methods, or the new methods

Who do you mean by "you". _I_ can't use a method that throws `UnsupportedOperationException`. So if I am given an instance of `FileRegion`, I don't know if I can use it. The important detail is the the _creator_ of the instance is different than the _consumer_ of the instance.

> but you can not use them both at the same time

That's not what's implemented. In the current implementation the _consumer_ doesn't get a choice. As far the API goes, the consumer is the only one that matters; the creator knows the concrete type and so has little need for interfaces.

> So you mean it is better to declare a completely new interface and remove the old FileRegion interface in the future?

With the current implementation, yes. But you could also change the implementation.

Apache9: The only way to create a `SlicedFileRegion` is to use the methods declared in `SliceableFileRegion`, then you also need to use the `transferBytesTo` methods declared in `SliceableFileRegion`.  And you can still use `DefaultFileRegion` in the old way. That's what I mean. Sorry I do not speak English very well.

ejona86: > These methods are expected to be removed in the future

So this argument feels like "because we're lazy," because the methods _are_ there today. "Let's not spend the effort of implementing them if they will be removed later." Which is a fair argument, but not by itself.

My implementation sketch would seem to indicate it isn't very hard to add support for the old method. Some edge cases might need to be checked, but that still seems to be pretty easy. So assuming my proposed implementation works, the "lazy" here seems to be quite extreme, as we don't want to add ~10 straight-forward lines to maintain compatibility. But I'd accept that maybe there's an issue with my implementation, which could dramatically change the cost/benefit.

> new implementation classes are not required to implement them.

This is a circular argument. I'm claiming that implementations should implement them, and that I don't see a strong reason why they shouldn't implement them.

> If you use the new methods introduced in SliceableFileRegion then you need to use the new transferBytesTo methods.

As I mention in the other comment, the problem is that these are two different "you"s. The code creating the instance is different than the one consuming it, otherwise they wouldn't need to communicate via an interface.

ejona86: > The only way to create a SlicedFileRegion is to use the methods declared in SliceableFileRegion

Not quite, since `DefaultFileRegion.slice(long, long)` is concrete, but that's a technicality. You're right that anyone slicing probably doesn't have the concrete type; I was wrong there. But I don't think it really changes my point that the producer and consumer are different pieces of code.

> Sorry I do not speak English very well.

Oh, wow. I didn't realize you were a non-native speaker. Your English is great!

Apache9: I just want to follow the `ByteBuf` way to make the code a bit easier to understand when adding the slice support. And to be honest I'm also a user of netty and I just want this feature, no matter who implement it. So if you think the code is BS and you already have an implementation then just implement it in netty. Or you can just modify my code and implement the compatibility code. I do not mind. Thanks.

ejona86: I don't think your code should get thrown out. I think there are two easy alternatives to address the problem I brought up:

1. Implement `transferTo()`. It seems easy to implement, so why not? This lets pre-existing code continue working with the slices, since slices would be implementing `FileRegion` properly.
2. Don't have slices extend `FileRegion`, since they don't implement the interface. That'd mean `SliceableFileRegion` would be a completely separate interface from `FileRegion`. Classes could still implement both interfaces though, if they supported both sets of operations.

To me, both of those options seem like fairly small changes to this PR. I'd prefer (1), but I'd be okay with (2).

Which solution do we prefer? (Which may be "3. do nothing; it's great as-is")

Apache9: The problem of the old `transferTo` method is that, a user can call `thansferTo` multiple times with the same position but the number of `transferred` bytes will be increased every time. Of course this is not the correct way to use `FileRegion` but anyway, it makes it a bit difficult to map the behavior to the new index based solution.

So just follow what I have done to the `DefaultFileRegion` in this PR? Implement two orthogonal transfer APIs? When given a `FileRegion` then you can use the old API, and for `SliceableFileRegion`, you'd better use the new API but it is still OK to use the old API, but it is not allowed to use the old API at first and then switch to the new API, and vice versa. Is it OK? Thanks.

ejona86: > it makes it a bit difficult to map the behavior to the new index based solution.

I'd generally expect the two APIs to be independent: doing calls on one of the APIs doesn't update the indexes used by the other. I could even accept the two APIs being mutually exclusive: doing calls on one of the APIs breaks the other API (e.g., it might throw an exception).

> When given a FileRegion then you can use the old API, and for SliceableFileRegion, you'd better use the new API but it is still OK to use the old API, but it is not allowed to use the old API at first and then switch to the new API, and vice versa.

Yeah, that sounds fine. I'm not sure I'd go as far as saying it's "not allowed" to mix the APIs, but simply "it makes no sense". The result is the same either way, but the reasoning is a bit more clear.

That's why I think you could use this code, maybe as-is:
```java
private long transferred;

public long transferred() { return transferred; }
￼
public long transferTo(WritableByteChannel target, long position) throws IOException {
  long written = transferBytesTo(target, position, this.count - position);
  if (written > 0) {
    this.transferred += written;
  }
  return written;
}
```

Apache9: OK, got it. Will provide a new patch soon. Thanks.

Apache9:
@ejona86 I've added the support for `transferTo`. PTAL.

And I didn't remove the related code in `DefaultFileRegion` as we will access the `transferred` field in native transport. I think we can discuss whether we can remove these code in the follow on native transport PR. 

Thanks.

transport/src/main/java/io/netty/channel/FileChannelBasedFileRegion.java
```diff
@@ -0,0 +1,165 @@
...
+        // FileRegion.position(), so here we need to add FileRegion.position() back before
+        // subtraction, and when calling transferTo, just use position, do not need to add
+        // FileRegion.position again.
+        long count = position() + count() - position;
+        if (count < 0 || position < 0) {
+            throw new IllegalArgumentException("position out of range: " + position
+                    + " (expected: 0 - " + (count() - 1) + ')');
+        }
+        if (count == 0) {
+            return 0L;
```
ejona86: nit: `0`

transport/src/main/java/io/netty/channel/FileChannelBasedFileRegion.java
```diff
@@ -120,4 +129,37 @@ public SliceableFileRegion transferRetainedSlice(long length) {
...
+
+    // The methods below are only used to keep compatible with old APIs.
+    @Override
+    public long transferred() {
+        return transferred;
+    }
+
+    @Override
+    public long transferTo(WritableByteChannel target, long position) throws IOException {
+        // for old style API, the position passed in is not relative, it contains
```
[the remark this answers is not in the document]
Apache9: Yeah, misused... We should use `transferred` as the position. Fixed.

transport/src/main/java/io/netty/channel/FileChannelBasedFileRegion.java
Apache9: Done.

transport/src/main/java/io/netty/channel/SliceableFileRegion.java
Apache9: Done.

transport/src/main/java/io/netty/channel/SliceableFileRegion.java
```diff
@@ -21,8 +21,8 @@
 /**
  * A {@link FileRegion} which supports slice.
  * <p>
- * We introduce this interface is only because that we can not add new methods to an interface until
- * next minor release. All the methods here will be moved to FileRegion when we release netty 4.2.
+ * We introduce this interface because we we can not add new methods to an interface until next
```
ejona86: Accidental double "we" :stuck_out_tongue:

transport/src/main/java/io/netty/channel/SliceableFileRegion.java
```diff
@@ -0,0 +1,131 @@
...
+     * will NOT be increased.
+     */
+    SliceableFileRegion slice(long index, long length);
+
+    /**
+     * Returns a slice of this file region's sub-region. This method does not modify
+     * {@code transferIndex} of this buffer.
+     * <p>
+     * Note that this method returns a {@linkplain #retain() retained} file region unlike
+     * {@link #slice()}.
```
Scottmitch: this link is dead -> `{@link #slice(long, long)}`

transport/src/main/java/io/netty/channel/SliceableFileRegion.java
```diff
@@ -0,0 +1,131 @@
...
+     */
+    SliceableFileRegion transferSlice(long length);
+
+    /**
+     * Returns a new slice of this file region's sub-region starting at the current
+     * {@code transferIndex} and increases the {@code transferIndex} by the size of the new slice (=
+     * {@code length}).
+     * <p>
+     * Note that this method returns a {@linkplain #retain() retained} file region unlike
+     * {@link #transferSlice()}.
```
Scottmitch: this link is dead -> `{@link #transferSlice(long)}.`

transport/src/main/java/io/netty/channel/FileChannelBasedFileRegion.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    @Override
+    public SliceableFileRegion slice(long index, long length) {
+        return new FileChannelBasedSlicedFileRegion(this, index, length);
+    }
+
+    @Override
+    public SliceableFileRegion retainedSlice(long index, long length) {
+        SliceableFileRegion sliced = slice(index, length);
```
Scottmitch: typically we override the methods which return the same type (e.g. `retain`, `touch`, etc...) so we can do things such as `return slice(index, length).retain();`. Is there any reason why we are not doing that here?

transport/src/main/java/io/netty/channel/FileChannelBasedFileRegion.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    @Override
+    public SliceableFileRegion transferRetainedSlice(long length) {
+        SliceableFileRegion sliced = transferSlice(length);
+        sliced.retain();
+        return sliced;
+    }
+
+    private long transferred;
```
Scottmitch: can you move this member variable to the top of the class declaration?

transport/src/main/java/io/netty/channel/FileChannelBasedSlicedFileRegion.java
```diff
@@ -0,0 +1,117 @@
...
+import java.io.IOException;
+import java.nio.channels.FileChannel;
+
+/**
+ * A derived {@link FileChannelBasedFileRegion} which shares reference counting with its parent.
+ */
+final class FileChannelBasedSlicedFileRegion extends FileChannelBasedFileRegion {
+
+    private final FileChannelBasedFileRegion parent;
+
```
Scottmitch: nit: remove extra lines?

transport/src/main/java/io/netty/channel/FileChannelBasedSlicedFileRegion.java
```diff
@@ -0,0 +1,117 @@
...
+    private final long offset;
+
+    private final long count;
+
+    public FileChannelBasedSlicedFileRegion(FileChannelBasedFileRegion parent, long index,
+            long count) {
```
Scottmitch: nit: move onto previous line?

transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
```diff
@@ -206,7 +207,42 @@ protected void doWrite(ChannelOutboundBuffer in) throws Exception {
                     // Break the loop and so incompleteWrite(...) is called.
                     break;
                 }
+            } else if (msg instanceof SliceableFileRegion) {
+                SliceableFileRegion region = (SliceableFileRegion) msg;
```
Scottmitch: I wonder if we should break each of these cases into its own private method. Folks may not be using each one of these types and JIT maybe able to optimize better if some unused code is factored out into methods.

@normanmaurer @ejona86 - wdyt?

transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
Scottmitch: can you create a followup issue for this, and link it to this PR?

transport/src/test/java/io/netty/channel/FileChannelBasedSlicedFileRegionTest.java
```diff
@@ -0,0 +1,174 @@
...
+        try {
+            out.write(TEST_DATA);
+        } finally {
+            out.close();
+        }
+    }
+
+    private byte[] transferBytesTo(SliceableFileRegion region, int length) throws IOException {
+        ByteArrayOutputStream bos = new ByteArrayOutputStream();
+        WritableByteChannel ch = Channels.newChannel(bos);
```
Scottmitch: does `ch` need to be closed?

transport/src/test/java/io/netty/channel/FileChannelBasedSlicedFileRegionTest.java
```diff
@@ -0,0 +1,174 @@
...
+        while (bos.size() < length && region.isTransferable()) {
+            region.transferBytesTo(ch, length - bos.size());
+        }
+        return bos.toByteArray();
+    }
+
+    private byte[] transferTo(FileRegion region) throws IOException {
+        ByteArrayOutputStream bos = new ByteArrayOutputStream();
+        WritableByteChannel ch = Channels.newChannel(bos);
```
Scottmitch: does `ch` need to be closed?

transport/src/main/java/io/netty/channel/FileChannelBasedFileRegion.java
Apache9: The `retain` and `touch` method in `DefaultFileRegion` return `FileRegion` and this is class is extended by `DefaultFileRegion` so we can not declared a `retain` or `touch` method which returns a `SliceableFileRegion`. And as @normanmaurer said above, users may extend `DefaultFileRegion` so we can not change the return value type of these methods in `DefaultFileRegion`. Maybe we could do this when releasing 4.2?

transport/src/main/java/io/netty/channel/SliceableFileRegion.java
Apache9: Done.

transport/src/main/java/io/netty/channel/SliceableFileRegion.java
Apache9: Done.

transport/src/main/java/io/netty/channel/FileChannelBasedFileRegion.java
Apache9: Done.

transport/src/main/java/io/netty/channel/FileChannelBasedSlicedFileRegion.java
Apache9: Done.

transport/src/main/java/io/netty/channel/FileChannelBasedSlicedFileRegion.java
Apache9: Done.

transport/src/test/java/io/netty/channel/FileChannelBasedSlicedFileRegionTest.java
Apache9: Not really, it just wraps a `ByteArrayOutputStream`. But anyway, no harm to close it. Fixed with a try finally.

transport/src/test/java/io/netty/channel/FileChannelBasedSlicedFileRegionTest.java
Apache9: Ditto.

transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
Apache9: I'm not sure whether JVM would do better(maybe it will not inline the unused branch and make a smaller doWrite method?) but I think at least it will make the method easier to read and understand. I can do the refactoring if no objections. Thanks.

Apache9:
@Scottmitch Filed #7155 for the follow on work. Do not know why I can not reply to the comment directly...

transport/src/main/java/io/netty/channel/FileChannelBasedFileRegion.java
```diff
@@ -0,0 +1,160 @@
...
+    public long transferred() {
+        return transferred;
+    }
+
+    @Override
+    public long transferTo(WritableByteChannel target, long position) throws IOException {
+        long count = count() - position;
+        if (count < 0 || position < 0) {
+            throw new IllegalArgumentException("position out of range: " + position
+                    + " (expected: 0 - " + (count() - 1) + ')');
```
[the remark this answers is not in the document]
fenik17: @carl-mastrangelo one of the IntelliJ IDEA's inspections: "Reports String literals of length one being used in concatenation. These literals may be replaced by equivalent character literals, gaining some performance enhancement."

See https://stackoverflow.com/questions/24859500/concatenate-char-literal-x-vs-single-char-string-literal-x

transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
Scottmitch: refactoring out not-commonly called code paths can reduce the byte code size of a method and make it easier to inline ... we don't have todo that as part of this PR but in theory seems like a good thing todo.

transport/src/main/java/io/netty/channel/FileChannelBasedFileRegion.java
Scottmitch: thanks for clarifying ... lgtm

transport/src/main/java/io/netty/channel/FileChannelBasedFileRegion.java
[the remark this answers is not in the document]
Apache9: So let's keep it as a character? The code is just the same with `DefaultFileRegion.transferTo`.

transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
```diff
@@ -208,36 +268,19 @@ protected void doWrite(ChannelOutboundBuffer in) throws Exception {
...
+                if (writeSpinCount == -1) {
+                    writeSpinCount = config().getWriteSpinCount();
                 }
-
-                if (done) {
-                    in.remove();
-                } else {
-                    // Break the loop and so incompleteWrite(...) is called.
-                    break;
+                switch (doWriteFileRegion(in, region, writeSpinCount)) {
```
netkins: MAJOR Add a default case to this switch. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3ASwitchLastCaseIsDefaultCheck)

transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
```diff
@@ -157,6 +158,65 @@ public final void read() {
...
+                flushedAmount += localFlushedAmount;
+                if (!sliceableRegion.isTransferable()) {
+                    status = WriteStatus.DONE;
+                    break;
+                }
+            }
+            in.progress(flushedAmount);
+            return status;
+        } else {
+            // TODO: consider remove this branch once we move the methods in SliceableFileRegion
```
netkins: INFO Complete the task associated to this TODO comment. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1135)

transport/src/main/java/io/netty/channel/FileChannelBasedFileRegion.java
```diff
@@ -0,0 +1,160 @@
...
+    public long transferableBytes() {
+        return count() - transferIndex;
+    }
+
+    @Override
+    public long transferBytesTo(WritableByteChannel target, long position, long length)
+            throws IOException {
+        if (position < 0 || position > count()) {
+            throw new IllegalArgumentException("position out of range: " + position
+                    + " (expected: 0 - " + (this.count() - 1) + ')');
```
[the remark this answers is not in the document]
Apache9: Done.

transport/src/main/java/io/netty/channel/FileChannelBasedFileRegion.java
```diff
@@ -0,0 +1,160 @@
...
+
+    @Override
+    public SliceableFileRegion transferRetainedSlice(long length) {
+        SliceableFileRegion sliced = transferSlice(length);
+        sliced.retain();
+        return sliced;
+    }
+
+    /**
+     * Explicitly open the underlying file-descriptor if not done yet.
```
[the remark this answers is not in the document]
Apache9: Done.

Apache9:
Any other concerns? @Scottmitch @carl-mastrangelo . Thanks.

transport/src/main/java/io/netty/channel/FileChannelBasedFileRegion.java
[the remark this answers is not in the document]
carl-mastrangelo: @fenik17 if the reason for the char is to silence IDE warnings, I get that.  If it's a perf thing, this code path is unlikely to be executed a lot.  That would not be a good reason.

Personally I prefer format strings as I find them easier to read.

transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
```diff
@@ -157,12 +158,71 @@ public final void read() {
...
+            int writeSpinCount) throws Exception {
+        if (region instanceof SliceableFileRegion) {
+            SliceableFileRegion sliceableRegion = (SliceableFileRegion) region;
+            if (!sliceableRegion.isTransferable()) {
+                return WriteStatus.DONE;
+            }
+            WriteStatus status = WriteStatus.INCOMPLETE;
+            long flushedAmount = 0;
+
+            for (int i = writeSpinCount - 1; i >= 0; i--) {
```
carl-mastrangelo: Why not use a plain incrementing for loop?

transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
```diff
@@ -157,12 +158,71 @@ public final void read() {
...
+        }
+    }
+
     @Override
     protected void doWrite(ChannelOutboundBuffer in) throws Exception {
         int writeSpinCount = -1;
 
         boolean setOpWrite = false;
-        for (;;) {
+        LOOP: for (;;) {
```
carl-mastrangelo: Upper case should be reserved for constants.  is a label really a constant?

transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
Scottmitch: this is preexisting IIRC ... the index isn't used and we don't have to declare an additional constant for `end` this way.

transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
Apache9: This is only my style...

What do we suggest in netty? 'Loop'? 'loop'?

normanmaurer: `loop:`

transport/src/main/java/io/netty/channel/FileChannelBasedFileRegion.java
```diff
@@ -0,0 +1,162 @@
...
+        if (count == 0) {
+            return 0;
+        }
+        if (refCnt() == 0) {
+            throw new IllegalReferenceCountException(0);
+        }
+        // Call open to make sure fc is initialized. This is a no-op if we called it before.
+        open();
+
+        long written = channel().transferTo(position() + position, count, target);
```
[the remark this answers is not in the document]
Apache9: The reference count check is in the `channel()` method... The method used to be public so we need to do reference count check. Let me do some refactoring here.

transport/src/main/java/io/netty/channel/FileChannelBasedFileRegion.java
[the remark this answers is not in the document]
fenik17: Yes, at this place it is just to suppress IDE warnings and to the be consistence with Netty's code style. 
+1 for use `String.format()` for the exceptions messages.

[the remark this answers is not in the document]
Apache9: Done.

transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
Apache9: Done.

transport/src/main/java/io/netty/channel/FileChannelBasedFileRegion.java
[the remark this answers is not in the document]
Apache9: Make `transferTo` call `transferBytesTo`, add ref count check.

Apache9:
Any more concerns?  Thanks.

Apache9:
Ping @normanmaurer  @Scottmitch

transport/src/main/java/io/netty/channel/FileChannelBasedFileRegion.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    @Override
+    public long transferableBytes() {
+        return count() - transferIndex;
+    }
+
+    @Override
+    public long transferBytesTo(WritableByteChannel target, long position, long length)
+            throws IOException {
```
Scottmitch: nit: put this on the previous line

Apache9: Done. So what's is expected line length? 120?

Apache9:
@Scottmitch  Rebase and squash done.

transport/src/main/java/io/netty/channel/FileChannelBasedFileRegion.java
normanmaurer: yes... 120

Scottmitch:
@normanmaurer - Please pull this in after the next release if you are satisfied.

Apache9:
4.1.16 is out, is it OK to pull in this now? There are several follow on issues, such as adding native transport support, and also need to start trying to add FileRegion support for HTTP/2.

Thanks.

normanmaurer:
@Apache9 I will try to have a final look this week

Apache9:
Any updates here? @normanmaurer 
Thanks.

Scottmitch:
@Apache9 - sorry for delay ... thanks for your patience ... hang in there we will pull this in soon.

Apache9:
Thanks @Scottmitch  . Wait for your good new.

Apache9:
Ping @Scottmitch @normanmaurer .

Thanks.

normanmaurer:
@Apache9 sorry I know you wait forever :( Its on my to do list for this week.

Apache9:
Wait for your good news.

normanmaurer:
@Apache9 sorry if I asked this before but can you please tell me again what exactly you need this for ? Is it "just" so you can write a File in slices without the need to open multiple FileChannels ?

Apache9:
@normanmaurer We need this to add the FileRegion support for HTTP/2. We can not embed a large FileRegion in a single data frame when we actually write it out, we need to split it to several small ones.

Thanks.

normanmaurer:
@Apache9 ok so its related to what I thought... Now a maybe stupid question but why could you not just open a FileChannel and create multiple new DefaultFileRegion using this FileChannel with different position and count arguments ? The only "problem" I can see here is that you need to ensure you not close the FileChannel under the hood when not all are written yet.

Apache9:
So you mean we should expose an API in HTTP/2 with something like a FileRegion but not a FileRegion? Maybe Http2FileRegion? Which will have the slice support and returns a FileRegion for slice? I'm OK with it. Just want to know your decision.

Thanks.

normanmaurer:
@Apache9 why you can't just write a FileRegion with the correct size from the beginning ? Or you want to "enqueue" something into the HTTP2 layer and dynamically slice out the correct stuff ?

Apache9:
Oh, maybe I need to say more. For the end users of the HTTP/2 API, it is impossible for them to write a `FileRegion` with exactly the correct size for a data frame, there are flow controls, and there maybe others writing on the same connection but different streams, so the next data frame available to them will be changed dynamically. In the framework we will do splitting and merging of data frames.

So I think the correct way to do this is to give end users the ability to write a very large `FileRegion` or something like a `FileRegion`, and we will split it to several small data frames in the framework.

Thanks.

normanmaurer:
@Apache9 ok got it...

So I think what I would do is to have another abstraction which basically provides a "Stream" of FileRegion on demand. I think this would minimise the changeset a lot. I am not convinced that we really need the slice stuff on the FileRegion itself.

Let me open a PR that shows the idea... Gimme a few minutes ;)

normanmaurer:
@Apache9 check https://github.com/netty/netty/pull/7435 ... No unit tests included yet ;)

normanmaurer:
@Apache9 sorry for all the delay and back and forth by @Scottmitch and myself were discussion some alternatives approves and at the end we think we may want to mark `FileRegion` as `@deprecated` and introduce some new API. Hopefully I can write up some proposal soon.

Apache9:
@normanmaurer Thanks. Wait for your good news.

normanmaurer:
Let me close this as this is not really something we want to do for 4.1.

Issue #7002, opened 2017-07-20
ReadOnlyBufferException regression due to "SslHandler aggregation of plaintext data on write"


ejona86:
### Actual behavior

gRPC is now seeing a `ReadOnlyBufferException`:

```
Caused by: java.nio.ReadOnlyBufferException
	at io.netty.buffer.ReadOnlyByteBuf.capacity(ReadOnlyByteBuf.java:413)
	at io.netty.buffer.AbstractByteBuf.ensureWritable(AbstractByteBuf.java:315)
	at io.netty.buffer.WrappedByteBuf.ensureWritable(WrappedByteBuf.java:214)
	at io.netty.handler.ssl.SslHandler$SslHandlerCoalescingBufferQueue.attemptCopyToCumulation(SslHandler.java:1851)
	at io.netty.handler.ssl.SslHandler$SslHandlerCoalescingBufferQueue.compose(SslHandler.java:1817)
	at io.netty.channel.AbstractCoalescingBufferQueue.remove(AbstractCoalescingBufferQueue.java:154)
	at io.netty.handler.ssl.SslHandler.wrap(SslHandler.java:752)
	at io.netty.handler.ssl.SslHandler.wrapAndFlush(SslHandler.java:731)
	at io.netty.handler.ssl.SslHandler.flush(SslHandler.java:712)
        ...
```

### Steps to reproduce

```sh
$ git clone https://github.com/grpc/grpc-java.git # I'm at 8572f5ff6
$ cd grpc-java
$ patch -p1 <<EOF
diff --git a/build.gradle b/build.gradle
index 45d87574..83d62a14 100644
--- a/build.gradle
+++ b/build.gradle
@@ -90,7 +90,7 @@ subprojects {
         protocPluginBaseName = 'protoc-gen-grpc-java'
         javaPluginPath = "$rootDir/compiler/build/exe/java_plugin/$protocPluginBaseName$exeSuffix"
 
-        nettyVersion = '4.1.13.Final'
+        nettyVersion = '4.1.14.Final-SNAPSHOT'
         guavaVersion = '19.0'
         protobufVersion = '3.3.1'
         protocVersion = '3.3.0' // TODO(carl-mastrangelo): set this from protobufVersion when versions align again
EOF
$ ./gradlew :grpc-interop-testing:installDist -PskipCodegen=true
$ ./run-test-server.sh
$ # In different terminal
$ ./run-test-client.sh --server_host_override=foo.test.google.fr --use_test_ca=true
```

[...]

PR #7004, opened 2017-07-21
ByteBufs which are not resizable should not throw in ensureWritable(i…

Scottmitch:
…nt,boolean)

Motivation:
ByteBuf#ensureWritable(int,boolean) returns an int indicating the status of the resize operation. For buffers that are unmodifiable or cannot be resized this method shouldn't throw but just return 1.
ByteBuf#ensureWriteable(int) should throw unmodifiable buffers.

Modifications:
- ReadOnlyByteBuf should be updated as described above.
- Add a unit test to SslHandler which verifies the read only buffer can be tolerated in the aggregation algorithm.

Result:
Fixes https://github.com/netty/netty/issues/7002.

### DIFF
--- a/buffer/src/test/java/io/netty/buffer/AbstractPooledByteBufTest.java
+++ b/buffer/src/test/java/io/netty/buffer/AbstractPooledByteBufTest.java
@@ -15,7 +15,13 @@
  */
 package io.netty.buffer;
 
-import static org.junit.Assert.*;
+import org.junit.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.greaterThanOrEqualTo;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
 
 public abstract class AbstractPooledByteBufTest extends AbstractByteBufTest {
 
@@ -33,5 +39,24 @@ public abstract class AbstractPooledByteBufTest extends AbstractByteBufTest {
         assertEquals(0, buffer.writerIndex());
         assertEquals(0, buffer.readerIndex());
         return buffer;
+    }
+
+    @Test
+    public void ensureWritableWithEnoughSpaceShouldNotThrow() {
+        ByteBuf buf = newBuffer(1, 10);
+        buf.ensureWritable(3);
+        assertThat(buf.writableBytes(), is(greaterThanOrEqualTo(3)));
+        buf.release();
+    }
+
+    @Test(expected = IndexOutOfBoundsException.class)
+    public void ensureWritableWithNotEnoughSpaceShouldThrow() {
+        ByteBuf buf = newBuffer(1, 10);
+        try {
+            buf.ensureWritable(11);
+            fail();
+        } finally {
+            buf.release();
+        }
     }
 }
--- a/buffer/src/test/java/io/netty/buffer/ReadOnlyByteBufTest.java
+++ b/buffer/src/test/java/io/netty/buffer/ReadOnlyByteBufTest.java
@@ -21,12 +21,23 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.ByteBuffer;
+import java.nio.ReadOnlyBufferException;
 import java.nio.channels.GatheringByteChannel;
 import java.nio.channels.ScatteringByteChannel;
 
-import static io.netty.buffer.Unpooled.*;
-import static org.junit.Assert.*;
-import static org.mockito.Mockito.*;
+import static io.netty.buffer.ByteBufUtil.ensureWritableSuccess;
+import static io.netty.buffer.Unpooled.BIG_ENDIAN;
+import static io.netty.buffer.Unpooled.EMPTY_BUFFER;
+import static io.netty.buffer.Unpooled.LITTLE_ENDIAN;
+import static io.netty.buffer.Unpooled.buffer;
+import static io.netty.buffer.Unpooled.unmodifiableBuffer;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
 /**
  * Tests read-only channel buffers
@@ -179,6 +190,37 @@ public class ReadOnlyByteBufTest {
     @Test
     public void shouldIndicateNotWritableAnyNumber() {
         assertFalse(unmodifiableBuffer(buffer(1)).isWritable(1));
+    }
+
+    @Test
+    public void ensureWritableIntStatusShouldFailButNotThrow() {
+        ensureWritableIntStatusShouldFailButNotThrow(false);
+    }
+
+    @Test
+    public void ensureWritableForceIntStatusShouldFailButNotThrow() {
+        ensureWritableIntStatusShouldFailButNotThrow(true);
+    }
+
+    private void ensureWritableIntStatusShouldFailButNotThrow(boolean force) {
+        ByteBuf buf = buffer(1);
+        ByteBuf readOnly = buf.asReadOnly();
+        int result = readOnly.ensureWritable(1, force);
+        assertEquals(1, result);
+        assertFalse(ensureWritableSuccess(result));
+        readOnly.release();
+    }
+
+    @Test(expected = ReadOnlyBufferException.class)
+    public void ensureWritableShouldThrow() {
+        ByteBuf buf = buffer(1);
+        ByteBuf readOnly = buf.asReadOnly();
+        try {
+            readOnly.ensureWritable(1);
+            fail();
+        } finally {
+            buf.release();
+        }
     }
 
     @Test
--- a/buffer/src/test/java/io/netty/buffer/SlicedByteBufTest.java
+++ b/buffer/src/test/java/io/netty/buffer/SlicedByteBufTest.java
@@ -22,7 +22,10 @@ import org.junit.Test;
 
 import java.nio.ByteBuffer;
 
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 /**
  * Tests sliced channel buffers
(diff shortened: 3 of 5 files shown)


Scottmitch:
@ejona86 - PTAL.

buffer/src/test/java/io/netty/buffer/SlicedByteBufTest.java
```diff
@@ -224,4 +228,26 @@ public void testWriteIso88591CharSequenceExpand() {
...
+        buf.ensureWritable(1);
+        assertEquals(originalWriterIndex + 1, buf.writerIndex());
+        assertEquals(1, buf.readableBytes());
+    }
+
+    @Test(expected = IndexOutOfBoundsException.class)
+    public void ensureWritableWithNotEnoughSpaceShouldThrow() {
+        ByteBuf buf = buffer(1);
+        try {
+            buf.writerIndex(buf.writerIndex() + 2);
```
normanmaurer: Just use the following code to not depend on the knowledge that we double the requested length of the buffer:

```java
buf.writerIndex(buf.writerIndex() + buf.writableBytes());
```

buffer/src/test/java/io/netty/buffer/SlicedByteBufTest.java
```diff
@@ -224,4 +228,26 @@ public void testWriteIso88591CharSequenceExpand() {
...
+        int originalWriterIndex = buf.writerIndex();
+        buf.writerIndex(buf.writerIndex() + 1);
+        buf.ensureWritable(1);
+        assertEquals(originalWriterIndex + 1, buf.writerIndex());
+        assertEquals(1, buf.readableBytes());
+    }
+
+    @Test(expected = IndexOutOfBoundsException.class)
+    public void ensureWritableWithNotEnoughSpaceShouldThrow() {
+        ByteBuf buf = buffer(1);
```
normanmaurer: I think you want to use `newBuffer(1)` to actually use a slice

buffer/src/test/java/io/netty/buffer/SlicedByteBufTest.java
```diff
@@ -224,4 +228,26 @@ public void testWriteIso88591CharSequenceExpand() {
     public void testWriteUtf16CharSequenceExpand() {
         super.testWriteUtf16CharSequenceExpand();
     }
+
+    @Test
+    public void ensureWritableWithEnoughSpaceShouldNotThrow() {
+        ByteBuf buf = buffer(1);
```
normanmaurer: I think you want to use `newBuffer(2)` to actually use a slice and also ensure we have enough space.

buffer/src/test/java/io/netty/buffer/ReadOnlyByteBufTest.java
```diff
@@ -182,6 +193,28 @@ public void shouldIndicateNotWritableAnyNumber() {
     }
 
     @Test
+    public void ensureWritableIntStatusShouldFailButNotThrow() {
+        ByteBuf buf = buffer(1);
+        ByteBuf readOnly = buf.asReadOnly();
+        int result = unmodifiableBuffer(readOnly).ensureWritable(1, false);
```
normanmaurer: why wrap in `unmodifiableBuffer` ? you called as `asReadOnly()` before so its already a read-only buffer.

buffer/src/test/java/io/netty/buffer/AbstractPooledByteBufTest.java
```diff
@@ -34,4 +38,26 @@ protected ByteBuf newBuffer(int length, int maxCapacity) {
         assertEquals(0, buffer.readerIndex());
         return buffer;
     }
+
+    @Test
+    public void ensureWritableWithEnoughSpaceShouldNotThrow() {
+        ByteBuf buf = buffer(1);
```
normanmaurer: you want to use `newBuffer(...)` to ensure you actually use a pooled buffer.

normanmaurer: Also you should use enough bytes to be sure that the buffer is big enough and not depend on the impl detail that we will use big enough buffer when we request a size of 1. Use `4` or something like that.

buffer/src/test/java/io/netty/buffer/AbstractPooledByteBufTest.java
```diff
@@ -34,4 +38,26 @@ protected ByteBuf newBuffer(int length, int maxCapacity) {
...
+        int originalWriterIndex = buf.writerIndex();
+        buf.writerIndex(buf.writerIndex() + 1);
+        buf.ensureWritable(1);
+        assertEquals(originalWriterIndex + 1, buf.writerIndex());
+        assertEquals(1, buf.readableBytes());
+    }
+
+    @Test(expected = IndexOutOfBoundsException.class)
+    public void ensureWritableWithNotEnoughSpaceShouldThrow() {
+        ByteBuf buf = buffer(1);
```
normanmaurer: you want to use `newBuffer(...)` to ensure you actually use a pooled buffer.

buffer/src/test/java/io/netty/buffer/AbstractPooledByteBufTest.java
```diff
@@ -34,4 +38,26 @@ protected ByteBuf newBuffer(int length, int maxCapacity) {
...
+        try {
+            buf.writerIndex(buf.writerIndex() + 2);
```
normanmaurer: you want to use `buf.writerIndex(buf.writerIndex() + buf.writableBytes())` if I am not wrong.

ejona86:
The core change LGTM.

buffer/src/test/java/io/netty/buffer/SlicedByteBufTest.java
Scottmitch: Agreed this test isn't right. I will allocate a new buffer manually here because I care about the size.

buffer/src/test/java/io/netty/buffer/ReadOnlyByteBufTest.java
Scottmitch: copy/pasta ... I will fix

Scottmitch:
@normanmaurer - copy/paste got outa control in the unit tests ... issues should hopefully be fixed ... PTAL.

buffer/src/test/java/io/netty/buffer/AbstractPooledByteBufTest.java
```diff
@@ -34,4 +40,22 @@ protected ByteBuf newBuffer(int length, int maxCapacity) {
         assertEquals(0, buffer.readerIndex());
         return buffer;
     }
+
+    @Test
+    public void ensureWritableWithEnoughSpaceShouldNotThrow() {
+        ByteBuf buf = newBuffer(1, 10);
+        buf.ensureWritable(3);
+        assertThat(buf.writableBytes(), is(greaterThanOrEqualTo(3)));
```
normanmaurer: missing `buf.release()`

buffer/src/test/java/io/netty/buffer/SlicedByteBufTest.java
```diff
@@ -227,4 +231,33 @@ public void testWriteIso88591CharSequenceExpand() {
     public void testWriteUtf16CharSequenceExpand() {
         super.testWriteUtf16CharSequenceExpand();
     }
+
+    @Test
+    public void ensureWritableWithEnoughSpaceShouldNotThrow() {
+        ByteBuf buf = Unpooled.buffer(10);
```
normanmaurer: why not also just use `newBuffer(...)` which will take care of returning slice ?

buffer/src/test/java/io/netty/buffer/SlicedByteBufTest.java
```diff
@@ -227,4 +231,33 @@ public void testWriteIso88591CharSequenceExpand() {
...
+        int originalReadableBytes = slice.readableBytes();
+        slice.ensureWritable(originalWriterIndex - slice.writerIndex());
+        assertEquals(originalWriterIndex, slice.writerIndex());
+        assertEquals(originalReadableBytes, slice.readableBytes());
+        slice.release();
+    }
+
+    @Test(expected = IndexOutOfBoundsException.class)
+    public void ensureWritableWithNotEnoughSpaceShouldThrow() {
+        ByteBuf buf = Unpooled.buffer(10);
```
normanmaurer: same as above

buffer/src/test/java/io/netty/buffer/AbstractPooledByteBufTest.java
normanmaurer: I guess this is also the reason why the build failed with a leak detected.

buffer/src/test/java/io/netty/buffer/SlicedByteBufTest.java
Scottmitch: I didn't want to be subject to the size manipulation in the `newBuffer` method and I would have to unwrap the buffer to adjust underlying indexes ... however I can do this incase we use a different allocator in the future.

Scottmitch:
@normanmaurer - PTAL

buffer/src/test/java/io/netty/buffer/SlicedByteBufTest.java
```diff
@@ -227,4 +230,33 @@ public void testWriteIso88591CharSequenceExpand() {
     public void testWriteUtf16CharSequenceExpand() {
         super.testWriteUtf16CharSequenceExpand();
     }
+
+    @Test
+    public void ensureWritableWithEnoughSpaceShouldNotThrow() {
+        ByteBuf slice = newBuffer(10);
+        ByteBuf unwrapped = slice.unwrap();
+        slice.unwrap().writerIndex(unwrapped.writerIndex() + 5);
```
normanmaurer: nit: i think you could also just use `unwrapped .´ here ?

buffer/src/test/java/io/netty/buffer/SlicedByteBufTest.java
```diff
@@ -227,4 +230,33 @@ public void testWriteIso88591CharSequenceExpand() {
...
+        assertEquals(originalWriterIndex, slice.writerIndex());
+        assertEquals(originalReadableBytes, slice.readableBytes());
+        slice.release();
+    }
+
+    @Test(expected = IndexOutOfBoundsException.class)
+    public void ensureWritableWithNotEnoughSpaceShouldThrow() {
+        ByteBuf slice = newBuffer(10);
+        ByteBuf unwrapped = slice.unwrap();
+        slice.unwrap().writerIndex(unwrapped.writerIndex() + 5);
```
normanmaurer: nit: I think you could just used `unwrapped.` here ?

buffer/src/test/java/io/netty/buffer/SlicedByteBufTest.java
Scottmitch: fixed

buffer/src/test/java/io/netty/buffer/SlicedByteBufTest.java
Scottmitch: fixed

Scottmitch:
4.1 (452fd3624) 4.0 (0090c9cad)

Issue #7001, opened 2017-07-20
HttpPostRequestEncoderTest failures introduced by 08748344d852a07011be6e74bb5ddc6dcf221c45


Scottmitch:
After commit 08748344d852a07011be6e74bb5ddc6dcf221c45 some tests in `HttpPostRequestEncoderTest` are now failing.

> Failed tests: 
  HttpPostRequestEncoderTest.testDataIsMultipleOfChunkSize2:412->checkNextChunkSize:423 Chunk is not 8096 bytes expected:<8096> but was:<8094>

> Failed tests: 
  HttpPostRequestEncoderTest.testDataIsMultipleOfChunkSize1:390->checkNextChunkSize:423 Chunk is not 8096 bytes expected:<8096> but was:<8093>

normanmaurer:
@Scottmitch strange not failing here :( Will investigate

violetagg:
@Scottmitch I tested yesterday on my local machine and they were ok + the PR voter also passed with all tests green. I'll also check again.

violetagg:
I'm not able to reproduce this failure neither on Mac nor Ubuntu
===
maven 3.3.9

java version "1.8.0_131"
Java(TM) SE Runtime Environment (build 1.8.0_131-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.131-b11, mixed mode)

Linux bosh-lite 4.2.0-36-generic #42~14.04.1-Ubuntu SMP Fri May 13 17:27:22 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
===
maven 3.3.9

java version "1.8.0_141"
Java(TM) SE Runtime Environment (build 1.8.0_141-b15)
Java HotSpot(TM) 64-Bit Server VM (build 25.141-b15, mixed mode)

Darwin MacBook-Pro-2.local 16.7.0 Darwin Kernel Version 16.7.0: Thu Jun 15 17:36:27 PDT 2017; root:xnu-3789.70.16~2/RELEASE_X86_64 x86_64

normanmaurer:
Fixed in https://github.com/netty/netty/pull/7005.

normanmaurer:
Fixed by https://github.com/netty/netty/pull/7005

PR #7005, opened 2017-07-21
Fix flacky multipart test introduced by 08748344d852a07011be6e74bb5dd…

normanmaurer:
…c6dcf221c45.

Motivation:

08748344d852a07011be6e74bb5ddc6dcf221c45 introduced two new tests which did not take into account that the multipart delimiter can be between 2 and 16 bytes long.

Modifications:

Take the multipart delimiter length into account.

Result:

Fixes [#7001]


normanmaurer:
@violetagg PTAL

violetagg:
LGTM 

Thanks,
Violeta

normanmaurer:
Cherry-picked into 4.1 (d141ba11bf) so the CI is stable again,

PR #7007, opened 2017-07-21
Delete temporary self-signed certs in SSLEngineTest-based tests

ejona86:
Motivation:

Lots of usages of SelfSignedCertificates were not deleting the certs at
the end of the test. This includes setupHandlers() which is used by
extending classes. Although these files will be deleted at JVM exit and
deleting them early does not free the JVM from trying to delete them at
shutdown, it's good practice to delete eagerly and since users sometimes
use tests as a form of documentation, it'd be good for them to see the
explicit deletes.

Modifications:

Add missing delete() calls to ½ of the SelfSignedCertificates-using
tests.

Result:

Tests that more clearly communicates which resources are created and
may accumulate without early delete.

CC @normanmaurer

### DIFF
--- a/handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
@@ -1073,6 +1073,7 @@ public abstract class SSLEngineTest {
         } finally {
             cleanupClientSslEngine(clientEngine);
             cleanupServerSslEngine(serverEngine);
+            ssc.delete();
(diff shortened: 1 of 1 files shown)


normanmaurer:
Cherry-picked into 4.1 (2fbce6d470862b8896a44cddf1a061b3b738d206) and 4.0 (62a207b5e268ebe483cbcaca0142510ec3b28870)

normanmaurer:
@ejona86 thanks buddy!

PR #7008, opened 2017-07-21
Unify EpollEventLoopGroup initialization.

carl-mastrangelo:
Motivation:
`Epoll.ensureAvailability()` is called multiple times, once in
static initialization and in a couple of the constructors.  This is
redundant and confusing to read.

Modifications:
Move `Epoll.ensureAvailability()` call into an instance initializer
and remove all other references.  This ensures that every EELG
checks availability, while still delaying the check until
construction.  This pattern is used when there are multiple ctors,
as in this class.

Result:
Easier to read code.

### DIFF
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoopGroup.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoopGroup.java
@@ -33,8 +33,8 @@ import java.util.concurrent.ThreadFactory;
  * it only works on linux.
  */
 public final class EpollEventLoopGroup extends MultithreadEventLoopGroup {
-    static {
-        // Ensure JNI is initialized by the time this class is loaded by this time!
+    {
+        // Ensure JNI is initialized by the time this class is loaded.
         Epoll.ensureAvailability();
     }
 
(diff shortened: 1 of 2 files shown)


transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoopGroup.java
```diff
@@ -33,8 +33,8 @@
  * it only works on linux.
  */
 public final class EpollEventLoopGroup extends MultithreadEventLoopGroup {
-    static {
-        // Ensure JNI is initialized by the time this class is loaded by this time!
+    {
```
netkins: MAJOR Move the contents of this initializer to a standard constructor or to field initializers. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1171)

carl-mastrangelo:
FYI, this still differs slightly from the KQueue impl, which uses a static initializer rather than an instance initializer like this one does.

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoopGroup.java
normanmaurer: seems like I learn something new... When is code called that is in an `{}` block but is not `static` @carl-mastrangelo ? Never saw this, even was not aware this is valid java code 😆

johnou: static is called when the class is loaded (once, when an object is instantiated or you call a static method), instance initializer is not invoked when the class is loaded (will not be invoked if you invoke a static method but will be invoked every time you instantiate an object of that class).

johnou: I think the big difference here is that if the static initializer throws an exception and causes a ExceptionInInitializerError, following instantiations of that class will result in NoClassDefFoundError. whereas the instance initializer will allow you to instantiate the object multiple times even if it throws an exception.

normanmaurer: @johnou so its basically a way to ensure its called for each constructor ?

johnou: @normanmaurer yep, allows you to avoid duplicating code when you have multiple constructors and comes in handy with anonymous classes (where you cannot use a constructor).

carl-mastrangelo: Just for fun:  instance initializers are pretty unusual:
* they run before the constructor
* they cannot explicitly throw exceptions, unlike static initializers
* there can be multiple of them, which means there are several methods named "<init>" in the class file.

Like @johnou said, they are also useful in anonymous class.  They are also what power [Double Brace initializers](http://wiki.c2.com/?DoubleBraceInitialization).

normanmaurer:
Sounds good... can you do the same for kqueue and so make it consistent?

carl-mastrangelo:
@normanmaurer Done.

normanmaurer:
Cherry-picked into 4.1 (d8f4547f5cbe95e3b46260556a27bdc7f76e0249) and partly into 4.0 (c7256dbebe0cb42292420d668400f3791eab0006)

normanmaurer:
@carl-mastrangelo thanks buddy

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/KQueueEventLoopGroup.java
```diff
@@ -30,7 +30,7 @@
 
 @UnstableApi
 public final class KQueueEventLoopGroup extends MultithreadEventLoopGroup {
-    static {
+    {
```
netkins: MAJOR Move the contents of this initializer to a standard constructor or to field initializers. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1171)

Scottmitch:
I missed the instance initializer portion of this PR .. see PR https://github.com/netty/netty/pull/7019 for followup

Issue #5052, opened 2016-03-27
Netty-tcnative with statically-linked boringssl uber-jar won't load


jchambers:
I created a simple project to try out the new statically-linked netty-tcnative uber-jar, and was surprised to discover that things weren't working as expected. Here are my dependencies (I chose `netty-codec-http2` for similarity with another project I'm working on):

```xml
<dependencies>
    <dependency>
        <groupId>io.netty</groupId>
        <artifactId>netty-tcnative-boringssl-static</artifactId>
        <version>1.1.33.Fork15</version>
    </dependency>
    <dependency>
        <groupId>io.netty</groupId>
        <artifactId>netty-codec-http2</artifactId>
        <version>4.1.0.CR4</version>
    </dependency>
</dependencies>
```

…and here's the app:

```java
public static void main(final String[] args) {
    System.out.println("OpenSSL available? " + OpenSsl.isAvailable());
    System.out.println("ALPN available?    " + OpenSsl.isAlpnSupported());

    if (OpenSsl.unavailabilityCause() != null) {
        System.out.println("Unavailability cause:");
        OpenSsl.unavailabilityCause().printStackTrace(System.out);
    }
}
```

When I run it under OS X (I'm afraid I don't have other environments handy right now), I get this:

```
OpenSSL available? false
ALPN available?    false
Unavailability cause:
java.lang.IllegalArgumentException: Failed to load any of the given libraries: [netty-tcnative-osx-x86_64, netty-tcnative]
	at io.netty.util.internal.NativeLibraryLoader.loadFirstAvailable(NativeLibraryLoader.java:167)
	at io.netty.handler.ssl.OpenSsl.<clinit>(OpenSsl.java:78)
	at TcnativeBoringsslUberJarSandboxApp.main(TcnativeBoringsslUberJarSandboxApp.java:6)
```

Setting a breakpoint at [`NativeLibraryLoader.java#L164`](https://github.com/netty/netty/blob/4653dc1d05cfac8766bed63c4dd28c8d06128423/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java#L164), it looks like we're running into this for both `netty-tcnative-osx-x86_64` and `netty-tcnative`:

```
java.lang.UnsatisfiedLinkError: no netty-tcnative-osx-x86_64 in java.library.path
```

[...]

PR #7011, opened 2017-07-22
Correctly handle unsigned int values returned from TCP_INFO

normanmaurer:
Motivation:

We used an int[] to store all values that are returned in the struct for TCP_INFO which is not good enough as it uses usigned int values.

Modifications:

- Change int[] to long[] and correctly cast values.

Result:

No more truncated values.

### DIFF
--- a/transport-native-epoll/src/main/c/netty_epoll_linuxsocket.c
+++ b/transport-native-epoll/src/main/c/netty_epoll_linuxsocket.c
@@ -178,46 +178,47 @@ static jint netty_epoll_linuxsocket_isIpTransparent(JNIEnv* env, jclass clazz, j
      return optval;
 }
 
-static void netty_epoll_linuxsocket_getTcpInfo(JNIEnv* env, jclass clazz, jint fd, jintArray array) {
+static void netty_epoll_linuxsocket_getTcpInfo(JNIEnv* env, jclass clazz, jint fd, jlongArray array) {
      struct tcp_info tcp_info;
      if (netty_unix_socket_getOption(env, fd, IPPROTO_TCP, TCP_INFO, &tcp_info, sizeof(tcp_info)) == -1) {
          return;
      }
-     unsigned int cArray[32];
-     cArray[0] = tcp_info.tcpi_state;
-     cArray[1] = tcp_info.tcpi_ca_state;
-     cArray[2] = tcp_info.tcpi_retransmits;
-     cArray[3] = tcp_info.tcpi_probes;
-     cArray[4] = tcp_info.tcpi_backoff;
-     cArray[5] = tcp_info.tcpi_options;
-     cArray[6] = tcp_info.tcpi_snd_wscale;
-     cArray[7] = tcp_info.tcpi_rcv_wscale;
-     cArray[8] = tcp_info.tcpi_rto;
-     cArray[9] = tcp_info.tcpi_ato;
-     cArray[10] = tcp_info.tcpi_snd_mss;
-     cArray[11] = tcp_info.tcpi_rcv_mss;
-     cArray[12] = tcp_info.tcpi_unacked;
-     cArray[13] = tcp_info.tcpi_sacked;
-     cArray[14] = tcp_info.tcpi_lost;
-     cArray[15] = tcp_info.tcpi_retrans;
-     cArray[16] = tcp_info.tcpi_fackets;
-     cArray[17] = tcp_info.tcpi_last_data_sent;
-     cArray[18] = tcp_info.tcpi_last_ack_sent;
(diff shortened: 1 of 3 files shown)


normanmaurer:
@carl-mastrangelo PTAL as well.

johnou:
@normanmaurer looks like tests failed. 

https://garage.netty.io/teamcity/viewLog.html?buildId=22521&buildTypeId=pulls_netty&guest=1

io.netty.channel.epoll (1)
 EpollSocketChannelTest.testTcpInfo 	
 
First failure:	7011	#5052 	Changes (337) 	22 Jul 17 14:37
java.lang.AssertionError
    at io.netty.channel.epoll.EpollSocketChannelTest.assertTcpInfo0(EpollSocketChannelTest.java:74)
    at io.netty.channel.epoll.EpollSocketChannelTest.testTcpInfo(EpollSocketChannelTest.java:39)

normanmaurer:
Will check... tests passed locally here 😱

normanmaurer:
@johnou doh!... my fault. Should be fixed now.

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollTcpInfo.java
```diff
@@ -61,133 +61,133 @@
  */
 public final class EpollTcpInfo {
 
-    final int[] info = new int[32];
+    final long[] info = new long[32];
 
     public int state() {
-        return info[0] & 0xFF;
+        return (int) info[0];
```
johnou: What about introducing a new int safe cast utility method? 

From openjdk 8..

```java
/**
     * Returns the value of the {@code long} argument;
     * throwing an exception if the value overflows an {@code int}.
     *
     * @param value the long value
     * @return the argument as an int
     * @throws ArithmeticException if the {@code argument} overflows an int
     * @since 1.8
     */
    public static int toIntExact(long value) {
        if ((int)value != value) {
            throw new ArithmeticException("integer overflow");
        }
        return (int)value;
    }
```

transport-native-epoll/src/main/java/io/netty/channel/epoll/LinuxSocket.java
```diff
@@ -154,7 +154,7 @@ public static LinuxSocket newSocketDomain() {
     private static native int getTcpUserTimeout(int fd) throws IOException;
     private static native int isIpFreeBind(int fd) throws IOException;
     private static native int isIpTransparent(int fd) throws IOException;
-    private static native void getTcpInfo(int fd, int[] array) throws IOException;
+    private static native void getTcpInfo(int fd, long[] array) throws IOException;
```
netkins: MINOR Consider using varargs for methods or constructors which take an array the last parameter. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUseVarargs)

normanmaurer: thats fine...

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollTcpInfo.java
normanmaurer: I think we not need this for now.

transport-native-epoll/src/main/c/netty_epoll_linuxsocket.c
```diff
@@ -178,46 +178,47 @@ static jint netty_epoll_linuxsocket_isIpTransparent(JNIEnv* env, jclass clazz, j
...
-     cArray[27] = tcp_info.tcpi_advmss;
-     cArray[28] = tcp_info.tcpi_reordering;
-     cArray[29] = tcp_info.tcpi_rcv_rtt;
-     cArray[30] = tcp_info.tcpi_rcv_space;
-     cArray[31] = tcp_info.tcpi_total_retrans;
-
-     (*env)->SetIntArrayRegion(env, array, 0, 32, cArray);
+     jlong cArray[32];
+     // Expand to 64 bits, then cast away unsigned-ness.
+     cArray[0] = (jlong) (unsigned long long) tcp_info.tcpi_state;
```
Scottmitch: lets use fixed width types `uint64_t` so we can be consistent with java's fixed width `jlong`.

normanmaurer: yep... doing now

normanmaurer:
Cherry-picked into 4.1 (efb2d141c1de9ef19443bb04678ef753894fcbe2) and 4.0 (26d97f586ebe5e430af75008b4e9f6a9dc463688)

PR #7013, opened 2017-07-23
change default max message size in MqttDecoder

### LINKED ISSUE #7012: small Default max message size in MqttDecoder
### Expected behavior
According to [MQTT 3.1.1 spec, section 2.2.3](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/csprd02/mqtt-v3.1.1-csprd02.html#_Toc385349759), MQTT packets size can take up to 268 435 455 bytes. I think this value should be a default. 

### Actual behavior
The current default value of MQTT packet to decode is 8092.

Sammers21:
### Motivation:
I think that default max message size in MqttDecoder should be equal to the maximum suggested by the protocol.

### Modification:
changed default max message size in MqttDecoder from 8 092 to 268 435 455.

Result:

Fixes #7012

### DIFF
--- a/codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttDecoder.java
+++ b/codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttDecoder.java
@@ -39,7 +39,7 @@ import static io.netty.handler.codec.mqtt.MqttCodecUtil.validateFixedHeader;
  */
 public final class MqttDecoder extends ReplayingDecoder<DecoderState> {
 
-    private static final int DEFAULT_MAX_BYTES_IN_MESSAGE = 8092;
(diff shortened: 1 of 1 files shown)


normanmaurer:
Using `268 435 455` as default sounds dangerous as this would mean that this amount of memory are allowed to be buffered in memory. This could easily lead to OOME.

Sammers21:
@normanmaurer, you are right. It is pretty dangerous.

johnou:
imo makes more sense to introduce a static init block and read a new netty property. defaulting to the original value 8092.

Sammers21:
@johnou, I know that I can use it. The question in default value

Sammers21:
So, @normanmaurer is right it can cause OOME

PR #7017, opened 2017-07-24
Respect DNS port that is specified via JNDI

normanmaurer:
Motivation:

JNDI allows to specify an port so we should respect it.

Modifications:

Use the specified port and if none is specifed use 53.

Result:

Correct handling of JNDI configured DNS.


normanmaurer:
See http://docs.oracle.com/javase/7/docs/technotes/guides/jndi/jndi-dns.html

normanmaurer:
Cherry-picked into 4.1 (486f962252)

PR #7019, opened 2017-07-25
Epoll and KQueue to use static initializers

Scottmitch:
Motivation:
Epoll and KQueue transports use an instance initializer instead of a static initializer to ensure the JNI libraries are loaded before use. However the availability of the JNI libraries will not change after class load time so static initializers should be sufficient.

Modifications:
- Change the instance initializer to a static initializer for Epoll and KQueue initialization in their respective EventLoops

Result:
Less initialization required at instance creation time.

### DIFF
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoopGroup.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoopGroup.java
@@ -33,8 +33,8 @@ import java.util.concurrent.ThreadFactory;
  * it only works on linux.
  */
 public final class EpollEventLoopGroup extends MultithreadEventLoopGroup {
-    {
-        // Ensure JNI is initialized by the time this class is loaded.
+    static {
+        // Ensure JNI is initialized by the time this class is loaded by this time!
         Epoll.ensureAvailability();
     }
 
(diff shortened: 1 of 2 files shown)


Scottmitch:
@carl-mastrangelo - FYI

carl-mastrangelo:
These instances are created seldomly so the amount of extra work should be small, no?  Also, as mentioned in the previous PR, moving init to static leads to confusing errors when static initialization fails (for whatever reason).  Leaving in instance initializers turns a NoClassDefFoundError into the actual one.  If you look at my commentary on the original PR, you can see I had to set special VM flags to get at the actual error, which is above what I would expect others to know how to do.  Lastly, it doesn't seem so bad to do it at instance creation time.  There are no static members or code in these classes that actually depends native code being available at classload time.

I don't think the benefits outweigh the downsides.

normanmaurer:
I agree with @carl-mastrangelo here... @Scottmitch any concerns ?

johnou:
@carl-mastrangelo while I did point out a minor short-coming in the static initializer use, we avoid that trap by invoking `Epoll.isAvailable()` before trying to create the EpollEventLoopGroup instance which seems to do the trick.

Scottmitch:
thanks for the feedback @carl-mastrangelo ... since instance creation is expected to be relatively low and having the instance initializers provides more useful feedback let me close this PR.

Issue #7021, opened 2017-07-25
ipv6 support test in netty_unix_socket.c misses some cases


dschobel:
We ran into a problem with the native epoll transport on [Aurora](https://aurora.apache.org/) when ipv6 networking is disabled where [the test to determine if ipv6 is supported](https://github.com/netty/netty/blob/4.1/transport-native-unix-common/src/main/c/netty_unix_socket.c#L179-L188) successfully created a ipv6 socket but all connections to it fail because, as mentioned, ipv6 networking is disabled.

This seems like strange behavior in Aurora and we can work around it with the `preferIPv4Stack` flag but I wonder whether netty should have a stronger test and try to connect to the test socket to confirm that ipv6 networking is really working.

normanmaurer:
@dschobel can you give some more details how it fails ?

dschobel:
sure, so the test linked above returns AF_INET6 because it can create the socket. when it comes time to connect though, the connection fails with `connect: Cannot assign requested address`

normanmaurer:
Interesting... will the same error happen when you try to bind to an IPv6 address?

normanmaurer:
@dschobel can you check if this fixes it for you: https://github.com/netty/netty/pull/7022

dschobel:
> will the same error happen when you try to bind to an IPv6 address?

`bind` looks like it succeeds. when running this https://gist.github.com/dschobel/0af22b79179c6f4fd3aa5267d87f7d3f in the same aurora container it prints out:

```
successfully bound!
successfully listening!
listening on a00:9157:: port number 37207
```

normanmaurer:
@dschobel and if you add a `connect` to the address it fails ?

dschobel:
@normanmaurer creating the ipv6 socket + then connecting is what failed to begin with, did you mean that I should try bind before connect?

normanmaurer:
@dschobel I asked you if you could try to do the connect and so give me a c reproducer so I can see what is going on :)

dschobel:
https://gist.github.com/dschobel/e9f29aa318304d2192d4357a02653da1 was my test case which is basically inlining netty's native socket code when the ip address is 0.0.0.0 or ::/0

normanmaurer:
@dschobel hmm this looks not correct to me as you use wild card addresses and try to connect to these here:
https://gist.github.com/dschobel/e9f29aa318304d2192d4357a02653da1#file-connect-c-L66

[...]

PR #7022, opened 2017-07-25
More bullet-proof way fo detecting if ipv6 is supported or not when u…

normanmaurer:
…sing the native transport.

Motivation:

We should try to bind to an ipv6 only socket before we enable ipv6 support in the native transport as it may not work due setup of the platform.

Modifications:

Try to bind to ::1 use IPV6 later on if this works

Result:

Fixes [#7021].

### DIFF
--- a/transport-native-unix-common/src/main/c/netty_unix_socket.c
+++ b/transport-native-unix-common/src/main/c/netty_unix_socket.c
@@ -183,8 +183,16 @@ static int socket_type(JNIEnv* env) {
         }
         return AF_INET6;
     } else {
+        // Explicitly try to bind to ::1 to ensure IPV6 can really be used.
+        // See https://github.com/netty/netty/issues/7021.
+        struct sockaddr_in6 addr;
+        memset(&addr, 0, sizeof(addr));
+        addr.sin6_family = AF_INET6;
+        addr.sin6_addr.s6_addr[15] = 1; /* [::1]:0 */
+        int res = bind(fd, (struct sockaddr *)&addr, sizeof(addr));
+
         close(fd);
-        return AF_INET6;
+        return res == 0 ? AF_INET6 : AF_INET;
     }
 }
 


normanmaurer:
@dschobel and @carl-mastrangelo as well...

transport-native-unix-common/src/main/c/netty_unix_socket.c
```diff
@@ -183,6 +189,27 @@ static int socket_type(JNIEnv* env) {
         }
         return AF_INET6;
     } else {
+        int optval = 1;
+        if (netty_unix_socket_setOption0(fd, IPPROTO_IPV6, IPV6_V6ONLY, &optval, sizeof(optval)) < 0) {
+            struct sockaddr_in6 ip6addr;
+
+            ip6addr.sin6_family = AF_INET6;
```
carl-mastrangelo: I'm not a C pro, but would this be clearer as:

```
memset(&ip6addr, 0, sizeof(struct sockaddr_in6))
ip6addr.sin6_family = AF_INET6;
ip6addr.sin6_addr = in6addr_any;
```

It's a few lines shorter, and shows you didn't miss any fields.  Thoughts?

transport-native-unix-common/src/main/c/netty_unix_socket.c
```diff
@@ -183,6 +189,27 @@ static int socket_type(JNIEnv* env) {
         }
         return AF_INET6;
     } else {
+        int optval = 1;
+        if (netty_unix_socket_setOption0(fd, IPPROTO_IPV6, IPV6_V6ONLY, &optval, sizeof(optval)) < 0) {
```
carl-mastrangelo: optional: can this be saved somewhere?  it would be nice to not make more syscalls than necessary per connection.

normanmaurer: @carl-mastrangelo what you want to save ? the return value ?

transport-native-unix-common/src/main/c/netty_unix_socket.c
normanmaurer: yeah this is actually better.

normanmaurer: that said let me also explicit set the port to `0`

transport-native-unix-common/src/main/c/netty_unix_socket.c
carl-mastrangelo: For the same input (ipv4Preferred)  yeah.  Like I said, I am not as familiar with native code here, but some things make me antsy.  For example, calling bind on the same Address below, won't that be racy if two sockets are created at almost the same time?  Woudn't that imply you need to do so_reuseaddr ?  If lots of connections came in quick, it would waste ports.  

Up to you, but as is I think it slows down reading; it isn't as "obviously correct" as it could be.

transport-native-unix-common/src/main/c/netty_unix_socket.c
```diff
@@ -183,6 +189,27 @@ static int socket_type(JNIEnv* env) {
...
+            struct sockaddr_in6 ip6addr;
+
+            memset(&ip6addr, 0, sizeof(struct sockaddr_in6))
```
carl-mastrangelo: missing semi colon

transport-native-unix-common/src/main/c/netty_unix_socket.c
normanmaurer: @carl-mastrangelo maybe I am missing something but `socket_type` will only be called when the native lib is loaded so I don't see how we would get into any trouble here. Can you explain ?

transport-native-unix-common/src/main/c/netty_unix_socket.c
normanmaurer: doh!...

transport-native-unix-common/src/main/c/netty_unix_socket.c
carl-mastrangelo: Oh, if its only called once thats fine.  I thought it was called for each socket.  NM.

transport-native-unix-common/src/main/c/netty_unix_socket.c
```diff
@@ -169,6 +169,12 @@ static jbyteArray createInetSocketAddressArray(JNIEnv* env, const struct sockadd
     return bArray;
 }
 
+
+static int netty_unix_socket_setOption0(jint fd, int level, int optname, const void* optval, socklen_t len) {
+    return setsockopt(fd, level, optname, optval, len);
+}
+
+
```
Scottmitch: nit: kill extra line

Scottmitch:
@dschobel - Can you verify if this fixes the issue?

normanmaurer:
@Scottmitch it does not fix it... working on it

carl-mastrangelo:
In talking with @pmarks-net , I think this code is not correct.  Some things:

* Setting `IPV6_V6ONLY` on an `AF_INET6` socket should always succeed.  It would be *interesting* if it did not.     
* Binding to the IPV6 any address should always work, even with no IPV6 adapter.
* **Connecting** to the bound address will fail, because the kernel will make the source the loopback address `::1`, which has no adapter.


@dschobel, can you verify that any off the above are not true on Aurora?

normanmaurer:
@dschobel did you see @carl-mastrangelo s comment ?

dschobel:
@carl-mastrangelo @normanmaurer I should be able to get it tested this week, if you want to write the test C then it'll be faster

normanmaurer:
@dschobel sorry I am super busy atm :(

normanmaurer:
@dschobel any news ?

dschobel:
Sorry been swamped, I haven't forgotten

On Aug 8, 2017 3:23 AM, "Norman Maurer" <notifications@github.com> wrote:

@dschobel <https://github.com/dschobel> any news ?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/netty/netty/pull/7022#issuecomment-320873360>, or mute
the thread

dschobel:
answers @carl-mastrangelo @normanmaurer 

> Setting IPV6_V6ONLY on an AF_INET6 socket should always succeed. It would be interesting if it did not.

I've never seen this fail so consider it confirmed for aurora.

> Binding to the IPV6 any address should always work, even with no IPV6 adapter.

Bind also succeeds when ipv6 networking is disabled. See: https://github.com/netty/netty/issues/7021#issuecomment-317848303


> Connecting to the bound address will fail, because the kernel will make the source the loopback address ::1, which has no adapter.

confirmed as well, connecting is the only way I've reliably seen the lack of ipv6 proven so that's what I suggested for the test in https://github.com/netty/netty/issues/7021#issuecomment-317791174

normanmaurer:
@dschobel ok I think I know now what exactly we need to test.  will work on a pr.

normanmaurer:
@carl-mastrangelo @dschobel I wonder if we could check if the loopback interface supports IPV6 via `getifaddrs` in our "is-ipv6 supported code" .

See http://man7.org/linux/man-pages/man3/getifaddrs.3.html

pmarks-net:
I've encountered two solutions to this "`connect([::]:port)` fails when IPv6 loopback is disabled" problem:

1. Probe for `::1`, and disable IPv6 if it's missing.  For example, see `grpc_ipv6_loopback_available()`: https://github.com/grpc/grpc/blob/master/src/core/lib/iomgr/socket_utils_common_posix.c (though it's worth noting that this approach fails if the machine has exhausted all TCP ports.)

2. Probe for `::1`, and crash with an error on startup if it's missing.  We do this for a lot of Google-internal code, where it's feasible to fix the problem instead of building workarounds.

normanmaurer:
@pmarks-net thanks will check ...

dschobel:
@normanmaurer @pmarks-net I tested `grpc_ipv6_loopback_available`'s logic in aurora and binding to ::1 correctly detects that ipv6 is unavailable so that seems like a winner.

normanmaurer:
@dschobel thanks for reporting back!

normanmaurer:
@dschobel @carl-mastrangelo PTAL again

normanmaurer:
@pmarks-net thanks again for the tip.

dschobel:
LGTM  👍

normanmaurer:
Cherry-picked into 4.1 (7290cbc48aea4971aa66c64c2f43d8a2611012ab) and 4.0 (937ceaa6663dbe5bbd1bc89bd6666e1c0f31edb2)

PR #7028, opened 2017-07-26
Make DelegatingChannelPromiseNotifier use Vararg overload

carl-mastrangelo:
Motivation:
ErrorProne complains that the array override doesn't match the
vararg super call.  See http://errorprone.info/bugpattern/Overrides

Additionally, almost every other Future uses the vararg form, so
it would be stylistically consistent to keep it that way.

Modifications:
Use vararg override.

Result:
Cleaner, less naggy code.

cc: @Scottmitch

### DIFF
--- a/transport/src/test/java/io/netty/channel/DelegatingChannelPromiseNotifierTest.java
+++ b/transport/src/test/java/io/netty/channel/DelegatingChannelPromiseNotifierTest.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.channel;
+
+import io.netty.util.concurrent.Future;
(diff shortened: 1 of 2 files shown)


normanmaurer:
@carl-mastrangelo is this binary compatible ? I would assume yes but I am not sure ....

carl-mastrangelo:
@normanmaurer Yes, and I think it was added this cycle.  (july 10th).  

The problem with the code as is is that it violates the definition of the parent class:

```
DelegatingChannelPromiseNotifier dcpn = ...;
dcpn.addListeners(singleListener); // FAILS!!!!
Promise f = dcpn;
dcpn.addListeners(singleListener); // SUCCESS
````

**Comment by normanmaurer** [comment]
@carl-mastrangelo ah got it ! Can you just also add a small unit test that covers this... this would be awesome :)

**Comment by carl-mastrangelo** [comment]
@normanmaurer will do, but I am away from my work computer atm.  Will send something out by tomorrow.

**Comment by normanmaurer** [comment]
@carl-mastrangelo <3 thanks...  I will wait with pulling in this then after the test is there.

**Comment by Scottmitch** [comment]
lgtm ... not sure how this happened ... I can't imagine I typed all this by hand but my IDE (which has been updated recently) generates the correct code.

**Comment by carl-mastrangelo** [comment]
@normanmaurer PTAL

File: transport/src/test/java/io/netty/channel/DelegatingChannelPromiseNotifierTest.java [hunk 1]
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright 2012 The Netty Project
```
Scottmitch: 2017

carl-mastrangelo: Ah, I thought you might just always use 2012, but I guessed wrong.

normanmaurer:
Thanks!

Cherry-picked into 4.1 (60250f3795d4082e59df6806dade2966ec583859)

PR #7031, opened 2017-07-28
First call channelReadComplete(...) before flush(...) for better perf…

normanmaurer:
…ormance

Motivation:

In Http2ConnectionHandler we call flush(...) in channelReadComplete(...) to ensure we update the flow-controller and write stuff to the remote peer. We should better flip the order and so may be able to pick up more bytes.

Modifications:

Change order of calls.

Result:

Better performance


normanmaurer:
also @buchgr @ejona86 @carl-mastrangelo ... I noticed this when working on the child channel pr and tried to optimise flushing.

carl-mastrangelo:
LGTM.

Seems safe enough.  gRPC doesn't actually do writes within read callbacks, (or at least code search doesn't think we do).  All gRPC writes take a thread hop anyways to respond to a write.

normanmaurer:
Cherry-picked into 4.1 (d56f403c69)

PR #7032, opened 2017-07-28
Only flush on channelReadComplete(...) in http2 hello world examples.

normanmaurer:
Motivation:

In our http1 hello world example we only flush on channelReadComplete(...) to make better use of gathering writes. We should do the same in http2.

Modifications:

Only flush in channelReadComplete(...)

Result:

Better performance and more consistent examples.

### DIFF
--- a/example/src/main/java/io/netty/example/http2/helloworld/frame/server/HelloWorldHttp2Handler.java
+++ b/example/src/main/java/io/netty/example/http2/helloworld/frame/server/HelloWorldHttp2Handler.java
@@ -61,10 +61,15 @@ public class HelloWorldHttp2Handler extends ChannelDuplexHandler {
         }
     }
 
+    @Override
+    public void channelReadComplete(ChannelHandlerContext ctx) throws Exception {
+        ctx.flush();
+    }
+
     /**
      * If receive a frame with end-of-stream set, send a pre-canned response.
      */
-    public void onDataRead(ChannelHandlerContext ctx, Http2DataFrame data) throws Exception {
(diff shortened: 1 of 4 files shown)


normanmaurer:
also @ejona86 @buchgr  @carl-mastrangelo  FYI

carl-mastrangelo:
Does FallbackRequestHandler need to be updated too?

normanmaurer:
@carl-mastrangelo `FallbackRequestHandler` already does this :)

carl-mastrangelo:
Perhaps I don't understand, your other change was to move flushing after channelReadComplete for http2 usage.  FallbackRequestHandler  is an http2 class correct?  Does that mean the channelReadComplete method on it should be removed since it is now covered by Http2ConnectionHandler ?

normanmaurer:
@carl-mastrangelo this is true when the `Http2ConnectionHandler` is in the pipeline but I not wanted to depend on this for `FallbackRequestHandler` as it is not really http2 related and so may give the user the wrong impression.

carl-mastrangelo:
In that case, why does HelloWorldHttp2Handler flush then?  That seems to be http2 specific.

normanmaurer:
@carl-mastrangelo I think you talk about the one for the frame and multiple codec. Because the user is not aware that `Http2ConnectionHandler` is in the pipeline and its an implementation detail IMHO.

normanmaurer:
let me pull this in... we can do a followup if needed.

normanmaurer:
Cherry-picked into 4.1 (580ac8cd41a426c30aeee363a9bf3dab2f9e303f)

PR #7033, opened 2017-07-28
Fix false-positive leak detection report when ReferenceCountedOpenSsl…

normanmaurer:
…Engine constructor throws.

 Motivation:

    We need to ensure we only create the ResourceLeak when the constructor not throws.

    Modifications:

    Ensure ResourceLeakDetector.track(...) is only called if the constructor of ReferenceCoundedOpenSslEngine not throws.

    Result:

    No more false-positves.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -235,7 +235,6 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
                                   int peerPort, boolean jdkCompatibilityMode, boolean leakDetection) {
         super(peerHost, peerPort);
         OpenSsl.ensureAvailability();
-        leak = leakDetection ? leakDetector.track(this) : null;
         this.alloc = checkNotNull(alloc, "alloc");
         apn = (OpenSslApplicationProtocolNegotiator) context.applicationProtocolNegotiator();
         session = new OpenSslSession(context.sessionContext());
@@ -288,6 +287,7 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
                 PlatformDependent.throwException(cause);
             }
         }
+        leak = leakDetection ? leakDetector.track(this) : null;
     }
 
     /**
--- a/handler/src/test/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngineTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngineTest.java
@@ -15,7 +15,9 @@
  */
 package io.netty.handler.ssl;
 
+import io.netty.handler.ssl.util.InsecureTrustManagerFactory;
 import io.netty.util.ReferenceCountUtil;
+import org.junit.Test;
 
 import javax.net.ssl.SSLEngine;
 
@@ -53,5 +55,15 @@ public class ReferenceCountedOpenSslEngineTest extends OpenSslEngineTest {
     @Override
     protected void cleanupServerSslEngine(SSLEngine engine) {
         ReferenceCountUtil.release(engine);
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void testNotLeakOnException() throws Exception {
+        clientSslCtx = SslContextBuilder.forClient()
+                .trustManager(InsecureTrustManagerFactory.INSTANCE)
+                .sslProvider(sslClientProvider())
+                .build();
+
+        clientSslCtx.newEngine(null);
     }
 }


normanmaurer:
Notices this when doing something stupid and so had the constructor throw :)

handler/src/test/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngineTest.java
```diff
@@ -54,4 +56,14 @@ protected void cleanupServerSslContext(SslContext ctx) {
     protected void cleanupServerSslEngine(SSLEngine engine) {
         ReferenceCountUtil.release(engine);
     }
+
+    @Test(expected = NullPointerException.class)
+    public void testNotLeakOnException() throws Exception {
```
normanmaurer: this produced a leak and so would fail the build when run with `-Pleak` on the ci.

normanmaurer:
also @carl-mastrangelo

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -236,56 +236,66 @@ protected void deallocate() {
         super(peerHost, peerPort);
         OpenSsl.ensureAvailability();
```
carl-mastrangelo: This could leak too, right?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -236,56 +236,66 @@ protected void deallocate() {
...
+            Lock readerLock = context.ctxLock.readLock();
+            readerLock.lock();
+            final long finalSsl;
             try {
-                networkBIO = SSL.bioNewByteBuffer(ssl, context.getBioNonApplicationBufferSize());
+                finalSsl = SSL.newSSL(context.ctx, !context.isClient());
+            } finally {
+                readerLock.unlock();
+            }
+            synchronized (this) {
```
carl-mastrangelo: I think it would be safer if you just assigned `refCnt` here, after you have created `ssl`.   That would fix the problem of exceptions being thrown in initializers, or in super(), or in OpenSsl.ensureAvailability();

normanmaurer: I thought about this but the subclass calls release() in the finalizer so I want to ensure everything is setup already.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: the leak is created after this so it may be fine ?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -236,56 +236,66 @@ protected void deallocate() {
         super(peerHost, peerPort);
         OpenSsl.ensureAvailability();
         leak = leakDetection ? leakDetector.track(this) : null;
```
Scottmitch: can we just create this leak detector as the last line in the constructor (with a comment for justification)? This will also minimize the "risk" for letting `this` escaping the constructor and using unitized fields of this object (even though practically it shouldn't). The member variable is `final` so we shouldn't have to worry about re-ordering with instructions outside the scope of the constructor.

normanmaurer:
Yep that's a better idea... let me do this

normanmaurer:
Cherry-picked into 4.1 (c5b5d363601a0d9ae294034c1e68a6de54a0958f) and 4.0 (3a97fdb8be6d6011836c997ec48d1c4643e9391a) after addressing the comment of @Scottmitch

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -288,6 +287,7 @@ protected void deallocate() {
                 PlatformDependent.throwException(cause);
             }
         }
+        leak = leakDetection ? leakDetector.track(this) : null;
```
Scottmitch: would be nice to add a comment explaining why this is at the end so we don't move it by accident later.

normanmaurer:
Sure let me do as a followup commit tomorrow when I am back at my laptop

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: Done... d9d3d65716

PR #7035, opened 2017-07-29
Use addAfter in Http2Codec.handlerAdded(....) to ensure we not end up…

normanmaurer:
… with incorrect handler order

Motivation:

When pipeline.replace(...) is used with the Http2Codec we will end up with an incorrect pipeline view for the handler that was replaced because the Http2Codec will add handler in front of itselv but the handler that will be replaced thinks that the Http2Codec is the handler that will be "next" in the pipeline.

Modifications:

- Use addAfter(...) in Http2Codec.handlerAdded(....) and so ensure correct ordering.
- Port tests written by @chhsiao90 in [#6952] to verify the fix.

Result:

Fixes [#6881]

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Codec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Codec.java
@@ -40,8 +40,11 @@ public final class Http2Codec extends ChannelDuplexHandler {
 
     @Override
     public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
-        ctx.pipeline().addBefore(ctx.executor(), ctx.name(), null, frameCodec);
-        ctx.pipeline().addBefore(ctx.executor(), ctx.name(), null, multiplexCodec);
+        // We use addAfter(...) because if we would use addBefore(...) we could end up with an incorrect
+        // order of handlers in the pipeline when replace(...) is used with the Http2Codec.
+        // See https://github.com/netty/netty/issues/6881.
+        ctx.pipeline().addAfter(ctx.executor(), ctx.name(), null, multiplexCodec);
+        ctx.pipeline().addAfter(ctx.executor(), ctx.name(), null, frameCodec);
 
         ctx.pipeline().remove(this);
     }
--- a/codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
+++ b/codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
@@ -17,8 +17,10 @@ package io.netty.handler.codec.http2;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
+import io.netty.channel.Channel;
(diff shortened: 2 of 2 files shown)


normanmaurer:
@chhsiao90 @rschmitt PTAL

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Codec.java
```diff
@@ -40,8 +40,11 @@ Http2FrameCodec frameCodec() {
 
     @Override
     public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
-        ctx.pipeline().addBefore(ctx.executor(), ctx.name(), null, frameCodec);
-        ctx.pipeline().addBefore(ctx.executor(), ctx.name(), null, multiplexCodec);
+        // We use addAfter(...) because if we would use addBefore(...) we could end up with an incorrect
+        // order of handlers in the pipeline when replace(...) is used with the Http2Codec.
+        // See https://github.com/netty/netty/issues/6881.
+        ctx.pipeline().addAfter(ctx.executor(), ctx.name(), null, multiplexCodec);
```
Scottmitch: should we use `ctx.pipeline().addAfter(ctx.name(), null, frameCodec);` instead? Don't we want to use the same executor as the channel?

normanmaurer: nope we want to use the same executor as the one that was used by the `Http2Codec`, just like we did before.

normanmaurer: @Scottmitch good to merge ?

normanmaurer:
Also maybe @ejona86 :)

ejona86:
This is confusing to me a couple ways:

1) From what I can tell, the order in the pipeline is correct, even with the replace. The problem is that the ctx for old handlers isn't what it assumed it would be for _inbound_ events.
2) The problem isn't the replace in my mind, but the _inbound_ event. I feel like this change would break any _outbound_ event.

This seems like a bug in `CleartextHttp2ServerUpgradeHandler`, in that it assumes the handler it wants to send to will share the current ctx after the replace. That's not safe, as demonstrated here, as the ctx is reused. Instead, it seems `CleartextHttp2ServerUpgradeHandler` should do an 'addAfter' followed by a `remove()` and then `ctx.fireUserEventTriggered()`. Normally this isn't a problem because you aren't trying to trigger any inbound or outbound events from the `ctx`, but the moment you do, the only safe thing in my mind is for the handler to avoid replace as then you have no "bookmark" of where you should trigger events.

normanmaurer:
@ejona86 so to be clear you think https://github.com/netty/netty/pull/6952 is more correct in terms of fixing it ?

ejona86:
@normanmaurer, yeah, #6952 looks to be exactly what I was thinking the fix should be (I hadn't looked at it earlier, only this PR and the issue). I feel that replace invalidates the ctx; you have no clue what the ctx represents after a replace. There is no way to use replace and have both `ctx.write()` and `ctx.fireUserEventTriggered()` function predictably.

normanmaurer:
@ejona86 ok fair enough... let me pull in #6952 for now then.

PR #7036, opened 2017-07-30
Ensure Http2FrameCodec uses Http2Settings.defaultSettings()

normanmaurer:
Motivation:

Http2FrameCodec should use Http2Settings.defaultSettings() when no Http2Settings were specified by the user.

Modifications:

Replace new Http2Settings() with Http2Settings.defaultSettings()

Result:

Use correct Http2Settings by default when using Http2FrameCodec in all cases.


normanmaurer:
@ejona86 @carl-mastrangelo FYI

normanmaurer:
Cherry-picked into 4.1 (2988fb8eeb7cc136b6b7660bcc5774cccc313744)

PR #7037, opened 2017-07-30
Allow IP_TRANSPARENT to be set on redirecting socket

albertok:
Motivation:

TProxy support is not complete, the IP_TRANSPARENT option can currently only be set on EpollServerSocket. Setting the option on an EpollSocket is also required so as to be able to bind a socket to a non-local address as described in ip(7)
http://man7.org/linux/man-pages/man7/ip.7.html

"TProxy redirection with the iptables TPROXY target also
requires that this option be set on the redirected socket."
 
Modifications:

Added IP_TRANSPARENT socket option to EpollSocketChannelConfig
 
Result:

A redirecting socket can be created with a non-local IP address as required for TProxy


normanmaurer:
@albertok LGTM... let me squash and merge once the ci finish.

normanmaurer:
Squashed and cherry-picked into 4.1 (21b7ab1f25f8c7032d74a82305bc12f18f863faf) and 4.0 (de6a18e10)

@albertok thanks

PR #7038, opened 2017-07-30
Fix hash function and hash table size in Snappy

fenik17:
Motivation:

1. Hash function in the Snappy encoding is wrong probably: used `+` instead of `*`. See the reference implementation [1].
2. Size of the hash table is calculated, but not applied.

Modifications:

1. Fix hash function: replace addition by multiplication.
2. Allocate hash table with calculated size.
3. Use an `Integer.numberOfLeadingZeros` trick for calculate log2.
4. Release buffers in tests.

Result:

1. Better compression. In the test `encodeAndDecodeLongTextUsesCopy` now compressed size is 175 instead of 180 before this change.
2. No extra allocations for hash table.
3. A bit faster the calc of shift (less expensive math operations).

[1] https://github.com/google/snappy/blob/513df5fb5a2d51146f409141f9eb8736935cc486/snappy.cc#L67

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/compression/Snappy.java
+++ b/codec/src/main/java/io/netty/handler/codec/compression/Snappy.java
@@ -72,7 +72,7 @@ public final class Snappy {
         final int baseIndex = inIndex;
 
         final short[] table = getHashTable(length);
-        final int shift = 32 - (int) Math.floor(Math.log(table.length) / Math.log(2));
+        final int shift = Integer.numberOfLeadingZeros(table.length) + 1;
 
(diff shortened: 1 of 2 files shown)


normanmaurer:
Cherry-picked into 4.1 (6ab9c177acdd87c7fd0a10f128e3ea21402bc05e) and 4.0 (287e14739)

@fenik17 thanks!

fenik17:
@normanmaurer is not pushed into 4.1?

normanmaurer:
@fenik17 ups... fixed :)

fenik17:
Спасибо :)

PR #7042, opened 2017-08-02
Unify KQueue and Epoll wait timeout approach

Scottmitch:
Motivation:
KQueueEventLoop and EpollEventLoop implement different approaches to applying a timeout of their respective poll calls. Epoll attempts to ensure the desired timeout is satisfied at the java layer and at the JNI layer, but it should be sufficient to account for spurious wakups at the JNI layer. Epoll timeout granularity is also limited to milliseconds which may be too large for some latency sensitive applications.

Modifications:
- Make EpollEventLoop wait method look like KQueueEventLoop
- Epoll should support a finer timeout granularity via timerfd_create. We can hide most of these details behind the epollWait0 JNI call to avoid crossing additional JNI boundaries.

Result:
More consistent timeout approach between KQueue and Epoll.

### DIFF
--- a/transport-native-epoll/src/main/c/netty_epoll_native.c
+++ b/transport-native-epoll/src/main/c/netty_epoll_native.c
@@ -27,6 +27,7 @@
 #include <netinet/tcp.h>
 #include <sys/types.h>
 #include <sys/socket.h>
+#include <sys/timerfd.h>
 #include <unistd.h>
 #include <arpa/inet.h>
 #include <fcntl.h>
@@ -49,16 +50,6 @@
 #ifndef TCP_FASTOPEN
 #define TCP_FASTOPEN 23
 #endif
-
-/**
- * On older Linux kernels, epoll can't handle timeout
- * values bigger than (LONG_MAX - 999ULL)/HZ.
- *
- * See:
- *   - https://github.com/libevent/libevent/blob/master/epoll.c#L138
- *   - http://cvs.schmorp.de/libev/ev_epoll.c?revision=1.68&view=markup
- */
-#define MAX_EPOLL_TIMEOUT_MSEC (35*60*1000)
 
 // optional
 extern int epoll_create1(int flags) __attribute__((weak));
@@ -85,8 +76,6 @@ jfieldID packetScopeIdFieldId = NULL;
 jfieldID packetPortFieldId = NULL;
 jfieldID packetMemoryAddressFieldId = NULL;
 jfieldID packetCountFieldId = NULL;
-
-clockid_t waitClockId = 0; // initialized by netty_unix_util_initialize_wait_clock
 
 // util methods
 static int getSysctlValue(const char * property, int* returnValue) {
@@ -117,18 +106,25 @@ static jint netty_epoll_native_eventFd(JNIEnv* env, jclass clazz) {
     jint eventFD = eventfd(0, EFD_CLOEXEC | EFD_NONBLOCK);
 
     if (eventFD < 0) {
-        int err = errno;
-        netty_unix_errors_throwChannelExceptionErrorNo(env, "eventfd() failed: ", err);
+        netty_unix_errors_throwChannelExceptionErrorNo(env, "eventfd() failed: ", errno);
     }
     return eventFD;
+}
+
+static jint netty_epoll_native_timerFd(JNIEnv* env, jclass clazz) {
+    jint timerFD = timerfd_create(CLOCK_MONOTONIC, TFD_CLOEXEC | TFD_NONBLOCK);
+
+    if (timerFD < 0) {
+        netty_unix_errors_throwChannelExceptionErrorNo(env, "timerfd_create() failed: ", errno);
+    }
+    return timerFD;
 }
 
 static void netty_epoll_native_eventFdWrite(JNIEnv* env, jclass clazz, jint fd, jlong value) {
     jint eventFD = eventfd_write(fd, (eventfd_t) value);
 
     if (eventFD < 0) {
-        int err = errno;
-        netty_unix_errors_throwChannelExceptionErrorNo(env, "eventfd_write() failed: ", err);
+        netty_unix_errors_throwChannelExceptionErrorNo(env, "eventfd_write() failed: ", errno);
     }
 }
 
@@ -138,6 +134,15 @@ static void netty_epoll_native_eventFdRead(JNIEnv* env, jclass clazz, jint fd) {
     if (eventfd_read(fd, &eventfd_t) != 0) {
         // something is serious wrong
         netty_unix_errors_throwRuntimeException(env, "eventfd_read() failed");
+    }
+}
+
+static void netty_epoll_native_timerFdRead(JNIEnv* env, jclass clazz, jint fd) {
+    uint64_t timerFireCount;
+
+    if (read(fd, &timerFireCount, sizeof(uint64_t)) < 0) {
+        // it is expected that this is only called where there is known to be activity, so this is an error.
+        netty_unix_errors_throwChannelExceptionErrorNo(env, "read() failed: ", errno);
     }
 }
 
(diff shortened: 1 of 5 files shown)


Scottmitch:
This still needs more testing, but thought I would get it upstream for review.

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
```diff
@@ -371,6 +368,10 @@ private void processReady(EpollEventArray events, int ready) {
                 // consume wakeup event
                 Native.eventFdRead(eventFd.intValue());
             } else {
+                if (fd == timerFd.intValue()) {
+                    continue;
```
Scottmitch: note we don't really care how many timeout events fired so I'm not using `timerfd_gettime`

transport-native-epoll/src/main/c/netty_epoll_native.c
```diff
@@ -169,42 +168,37 @@ static jint netty_epoll_native_epollCreate(JNIEnv* env, jclass clazz) {
...
+          return -err;
         }
-      } else {
-        return -err;
-      }
+        do {
+            result = epoll_wait(efd, ev, len, -1);
+            if (result > 0) {
+              // Detect timeout, and preserve the epoll_wait API.
+              return result == 1 && ev[0].data.fd == timerFd ? 0 : result;
```
normanmaurer: just a nit: we use 4 spaces and not 2 :)

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
normanmaurer: @Scottmitch +1... not sure about this as well here.

transport-native-epoll/src/main/c/netty_epoll_native.c
```diff
@@ -169,42 +168,37 @@ static jint netty_epoll_native_epollCreate(JNIEnv* env, jclass clazz) {
...
+            if (result >= 0) {
+              return result;
+            }
+        } while((err = errno) == EINTR);
+    } else {
+        struct itimerspec ts;
+        memset(&ts.it_interval, 0, sizeof(struct timespec));
+        ts.it_value.tv_sec = tvSec;
+        ts.it_value.tv_nsec = tvNsec;
+        result = timerfd_settime(timerFd, 0, &ts, NULL);
```
normanmaurer: @Scottmitch would be interesting to see how this compares in terms of perf to the old way of doing time based epoll calls.

transport-native-epoll/src/main/c/netty_epoll_native.c
```diff
@@ -169,42 +168,37 @@ static jint netty_epoll_native_epollCreate(JNIEnv* env, jclass clazz) {
...
+        // Zeros = poll (aka return immediately).
+        do {
+            result = epoll_wait(efd, ev, len, 0);
+            if (result >= 0) {
+              return result;
+            }
+        } while((err = errno) == EINTR);
+    } else {
+        struct itimerspec ts;
+        memset(&ts.it_interval, 0, sizeof(struct timespec));
```
normanmaurer: do we even need this memset here as we set all struct fields anyway... ?

transport-native-epoll/src/main/c/netty_epoll_native.c
```diff
@@ -123,6 +112,16 @@ static jint netty_epoll_native_eventFd(JNIEnv* env, jclass clazz) {
     return eventFD;
 }
 
+static jint netty_epoll_native_timerFd(JNIEnv* env, jclass clazz) {
+    jint timerFD = timerfd_create(CLOCK_MONOTONIC, TFD_CLOEXEC | TFD_NONBLOCK);
+
+    if (timerFD < 0) {
+        int err = errno;
+        netty_unix_errors_throwChannelExceptionErrorNo(env, "timerfd_create() failed: ", err);
```
normanmaurer: just a nit: you could merge the two lines. Not need to store `errno` to err first.

transport-native-epoll/src/main/c/netty_epoll_native.c
```diff
@@ -50,16 +51,6 @@
...
 
-/**
- * On older Linux kernels, epoll can't handle timeout
- * values bigger than (LONG_MAX - 999ULL)/HZ.
- *
- * See:
- *   - https://github.com/libevent/libevent/blob/master/epoll.c#L138
- *   - http://cvs.schmorp.de/libev/ev_epoll.c?revision=1.68&view=markup
- */
-#define MAX_EPOLL_TIMEOUT_MSEC (35*60*1000)
```
normanmaurer: less hacks FTW.

transport-native-epoll/src/main/c/netty_epoll_native.c
Scottmitch: Yes we do. Notice this is setting `it_interval` to 0 indicating we don't want a re-ocurring event, as opposed to `it_value` which is the "one shot" timer.

transport-native-epoll/src/main/c/netty_epoll_native.c
Scottmitch: sgtm. IDE doesn't do as well in C++ so it more manual and likely to be missed :(

transport-native-epoll/src/main/c/netty_epoll_native.c
Scottmitch: agreed. I will throw together some perf tests, but it would be nice to have some more real world evaluation. Perhaps someone involved in https://issues.apache.org/jira/browse/CASSANDRA-13651 can run some tests ... can you reach out?

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
Scottmitch: Based upon https://linux.die.net/man/2/timerfd_settime I wouldn't expect data to pile up. IIUC `read` just allows you to get a count of how many times the timer fired, or will block/EAGAIN, since the last time the timer was changed or since the last `read`. Since we reset the interval each time

> read(2)
If the timer has already expired one or more times since its settings were last modified using timerfd_settime(), or since the last successful read(2), then the buffer given to read(2) returns an unsigned 8-byte integer (uint64_t) containing the number of expirations that have occurred. (The returned value is in host byte order, i.e., the native byte order for integers on the host machine.)
If no timer expirations have occurred at the time of the read(2), then the call either blocks until the next timer expiration, or fails with the error EAGAIN if the file descriptor has been made nonblocking (via the use of the fcntl(2) F_SETFL operation to set the O_NONBLOCK flag).

transport-native-epoll/src/main/c/netty_epoll_native.c
normanmaurer: done...

Scottmitch:
FYI this approach is also used in [libevent](https://github.com/libevent/libevent/blob/master/epoll.c)

normanmaurer:
@Scottmitch cool :) like i said the code looks good once we did more testing we should pull it in

Scottmitch:
Test failures appear to be unrelated ... coincidentally a test timeout using NIO and OIO see https://github.com/netty/netty/issues/7056.

normanmaurer:
@Scottmitch want to merge ?

Scottmitch:
@normanmaurer - I haven't had time to perf test ... but yes I want to merge :)

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
```diff
@@ -204,43 +221,23 @@ public void setIoRatio(int ioRatio) {
...
+        // So we need to check task queue again before calling epoll_wait. If we don't, the task might be pended
+        // until epoll_wait was timed out. It might be pended until idle timeout if IdleStateHandler existed
+        // in pipeline.
+        if (oldWakeup && hasTasks()) {
+            return epollWaitNow();
+        }
 
-            int selectedKeys = Native.epollWait(epollFd.intValue(), events, (int) timeoutMillis);
-            selectCnt ++;
+        long totalDelay = delayNanos(System.nanoTime());
```
rkapsi: Use `TimeUnit` to avoid time math and the magic constant?

```java
long delayInNanos = delayNanos(System.nanoTime());
int delayInSeconds = (int)min(TimeUnit.NANOSECONDS.toSeconds(delayInNanos), Integer.MAX_VALUE);
return Native.epollWait(epollFd, events, timerFd, delayInSeconds, 
    (int) min(delayInNanos - TimeUnit.SECONDS.toNanos(delayInSeconds), Integer.MAX_VALUE));
```

Scottmitch: I would prefer going through the additional layers of indirection at this level. For example `TimeUnit.SECONDS.toNanos` has some additional indirection and overflow checking. We can also use `Math.min` here which is intrinsic.

normanmaurer:
@Scottmitch ship it...

transport-native-epoll/src/main/c/netty_epoll_native.c
```diff
@@ -141,6 +137,15 @@ static void netty_epoll_native_eventFdRead(JNIEnv* env, jclass clazz, jint fd) {
     }
 }
 
+static void netty_epoll_native_timerFdRead(JNIEnv* env, jclass clazz, jint fd) {
+    uint64_t timerFireCount;
+
+    if (read(fd, &timerFireCount, sizeof(uint64_t)) < 0) {
```
carl-mastrangelo: not to stick my nose in, but technically this can return fewer than 8 bytes right?

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
```diff
@@ -114,6 +124,13 @@ public Integer call() throws Exception {
                         // ignore
                     }
                 }
+                if (timerFd != null) {
+                    try {
+                        timerFd.close();
+                    } catch (Exception e) {
+                        // ignore
```
carl-mastrangelo: loggable?

Scottmitch:
4.1 (fe2dd973e9)

normanmaurer:
@Scottmitch thanks!

Scottmitch:
Also back ported to 4.0 (35e7e2a) for maintenance purposes

PR #7043, opened 2017-08-02
(refactor) some opportunities to use lambda functions

desmorto:
Motivation:

Modification:

Result:

Fixes #<GitHub issue number>. 


normanmaurer:
We can not do this as we still support Java 6 as runtime

PR #7044, opened 2017-08-02
Upgrading to Conscrypt 1.0.0.RC9.

nmittler:
Motivation:

Starting with 1.0.0.RC9, conscrypt supports a buffer allocator.

Modifications:

- Updated the creation process for the engine to pass through the
ByteBufAllocator.
- Wrap a ByteBufAllocator with an adapter for conscrypt.
- Added a property to optionally control whether conscrypt uses
Netty's buffer allocator.

Result:

Netty+conscrypt will support using Netty's ByteBufAllocator.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ConscryptAlpnSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ConscryptAlpnSslEngine.java
@@ -214,8 +214,7 @@ abstract class ConscryptAlpnSslEngine extends JdkSslEngine {
 
         BufferAdapter(ByteBuf nettyBuffer) {
             this.nettyBuffer = nettyBuffer;
-            nettyBuffer.writerIndex(nettyBuffer.capacity());
-            this.buffer = nettyBuffer.nioBuffer();
+            this.buffer = nettyBuffer.nioBuffer(0, nettyBuffer.capacity());
         }
 
         @Override
--- a/handler/src/main/java/io/netty/handler/ssl/ConscryptAlpnSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ConscryptAlpnSslEngine.java
@@ -212,7 +212,7 @@ abstract class ConscryptAlpnSslEngine extends JdkSslEngine {
         private final ByteBuf nettyBuffer;
         private final ByteBuffer buffer;
 
-        private BufferAdapter(ByteBuf nettyBuffer) {
+        BufferAdapter(ByteBuf nettyBuffer) {
             this.nettyBuffer = nettyBuffer;
             nettyBuffer.writerIndex(nettyBuffer.capacity());
             this.buffer = nettyBuffer.nioBuffer();
--- a/handler/src/main/java/io/netty/handler/ssl/ConscryptAlpnSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ConscryptAlpnSslEngine.java
@@ -19,6 +19,8 @@ import static io.netty.handler.ssl.SslUtils.toSSLHandshakeException;
 import static io.netty.util.internal.ObjectUtil.checkNotNull;
 import static java.lang.Math.min;
 
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufAllocator;
 import io.netty.handler.ssl.JdkApplicationProtocolNegotiator.ProtocolSelectionListener;
 import io.netty.handler.ssl.JdkApplicationProtocolNegotiator.ProtocolSelector;
 import java.lang.reflect.Method;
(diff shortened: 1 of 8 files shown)


handler/src/main/java/io/netty/handler/ssl/ConscryptAlpnSslEngine.java
```diff
@@ -173,4 +194,45 @@ private static boolean isConscryptEngine(SSLEngine engine, Class<?> enginesClass
...
+    private static final class BufferAllocatorAdapter extends BufferAllocator {
+        private final ByteBufAllocator alloc;
+
+        BufferAllocatorAdapter(ByteBufAllocator alloc) {
+            this.alloc = alloc;
+        }
+
+        @Override
+        public AllocatedBuffer allocateDirectBuffer(int capacity) {
+            return new BufferAdapter(alloc.directBuffer(capacity));
```
netkins: MAJOR Avoid instantiation through private constructors from outside of the constructors class. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AAccessorClassGeneration)

handler/src/test/java/io/netty/handler/ssl/ConscryptJdkSslEngineInteropTest.java
```diff
@@ -73,4 +73,10 @@ public void testMutualAuthValidClientCertChainTooLongFailOptionalClientAuth() th
     @Override
     public void testMutualAuthValidClientCertChainTooLongFailRequireClientAuth() throws Exception {
     }
+
+    @Override
+    protected boolean mySetupMutualAuthServerIsValidServerException(Throwable cause) {
```
nmittler: The fact that we didn't need it before was a bug. Conscrypt wasn't properly propagating the alert back to the client. Now it behaves the same way as netty-tcnative.

handler/src/main/java/io/netty/handler/ssl/ConscryptAlpnSslEngine.java
```diff
@@ -173,4 +194,45 @@ private static boolean isConscryptEngine(SSLEngine engine, Class<?> enginesClass
...
+        }
+    }
+
+    private static final class BufferAdapter extends AllocatedBuffer {
+        private final ByteBuf nettyBuffer;
+        private final ByteBuffer buffer;
+
+        private BufferAdapter(ByteBuf nettyBuffer) {
+            this.nettyBuffer = nettyBuffer;
+            nettyBuffer.writerIndex(nettyBuffer.capacity());
```
nmittler: Ah right ... IIRC this is so that the call to`nettyBuffer.nioBuffer` returns a properly sized buffer. I suspect there may be a better way to do this?

handler/src/main/java/io/netty/handler/ssl/ConscryptAlpnSslEngine.java
nmittler: done.

handler/src/main/java/io/netty/handler/ssl/ConscryptAlpnSslEngine.java
```diff
@@ -173,4 +194,45 @@ private static boolean isConscryptEngine(SSLEngine engine, Class<?> enginesClass
...
+    }
+
+    private static final class BufferAdapter extends AllocatedBuffer {
+        private final ByteBuf nettyBuffer;
+        private final ByteBuffer buffer;
+
+        BufferAdapter(ByteBuf nettyBuffer) {
+            this.nettyBuffer = nettyBuffer;
+            nettyBuffer.writerIndex(nettyBuffer.capacity());
+            this.buffer = nettyBuffer.nioBuffer();
```
normanmaurer: @nmittler so you basically just want to get the `ByteBuffer` of the whole `ByteBuf` ?

If so use:

```java
this.buffer = nettyBuffer.nioBuffer(0, nettyBuffer.capacity());
```
 
no need to adjust the writerIndex ;)

nmittler: ah right, makes sense. done.

nmittler:
@normanmaurer any other comments before I merge?

handler/src/test/java/io/netty/handler/ssl/ConscryptJdkSslEngineInteropTest.java
normanmaurer: got it...

normanmaurer:
@nmittler once the CI finish of course ;)

nmittler:
@normanmaurer of course :)

nmittler:
@normanmaurer I suspect the failing tests are unrelated ... can you confirm?

nmittler:
@normanmaurer I forget... does this need to be cherry-picked in another branch as well?

Scottmitch:
lgtm

@nmittler - Yip we keep the sslengine stuff consistent between 4.1 and 4.0 because it would be a maintenance headache otherwise.

4.0 (3681f14e0)

nmittler:
@Scottmitch ah right ... thanks for cherry-picking!

PR #7048, opened 2017-08-04
Remove io.netty.packagePrefix system property

Scottmitch:
Motivation:
Now that the NativeLibraryLoader implicitly detects the shaded package prefix we no longer need the io.netty.packagePrefix system property.

Modifications:
- Remove io.netty.packagePrefix processing from NativeLibraryLoader

Result:
Code is cleaner.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
+++ b/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
@@ -202,10 +202,7 @@ public final class NativeLibraryLoader {
      */
     public static void load(String originalName, ClassLoader loader) {
         // Adjust expected name to support shading of native libraries.
-        String implicitPackagePrefix = calculatePackagePrefix();
-        // The system property should not be necessary; it can be removed in the future.
-        String packagePrefix = SystemPropertyUtil.get("io.netty.packagePrefix", implicitPackagePrefix);
-        String name = packagePrefix.replace('.', '-') + originalName;
+        String name = calculatePackagePrefix().replace('.', '-') + originalName;
 
         String libname = System.mapLibraryName(name);
         String path = NATIVE_RESOURCE_HOME + libname;


Scottmitch:
@ejona86 - FYI

normanmaurer:
@Scottmitch will this not break people that still depend on it ?

Scottmitch:
@normanmaurer - I wanted to wait until we pushed out a release first. However I think the implicit derivation of the package prefix should be sufficient.

normanmaurer:
@Scottmitch ok sounds good... just wanted to double-check. Ship it :)

normanmaurer:
Cherry-picked into 4.1 (a75ac747f06483060070b2a303bf8917a6d7847c) and 4.0 (984016e527).

@Scottmitch thanks !

cchantep:
Thx for the work. BTW I got a failure after the loading of the native lib using this mechanism: https://groups.google.com/forum/?fromgroups#!topic/netty/2w7Ctx9w2Uo

cchantep:
I'm wondering whether the `packagePrefix` should be used when calling [`env->GetMethodID(..)`](https://github.com/netty/netty/blob/4.1/transport-native-unix-common/src/main/c/netty_unix_socket.c#L1025)? As classes from netty-transport-native-unix-common are shaded along with kqueue or epoll? To look for something like `"(Ljava/lang/String;IILfoo/io/netty/channel/unix/DatagramSocketAddress;)V"` ?

normanmaurer:
I will try to check what you did tomorrow latest. We used it with success in the past but I think we never tried Datagram stuff while shading. Does it only happens with Datagram or also without ?

Also can you please open a new issues and not add to old ones that are closed ?

cchantep:
Hi @normanmaurer ,

Thanks for the quick reply.

The issue happens for kqueue-osx_x86-64 or epoll-linux_x86-64 (both based on unix-common as far as I see). The error details are visible [there](https://github.com/cchantep/netty-native-shading#mac-os-x-kqueue).

Do you prefer to check before creating a new issue?

Best regards

normanmaurer:
Just create the issue with all the details. At worse we can just close it later once we found the root cause

cchantep:
Done at https://github.com/netty/netty/issues/8090

PR #7049, opened 2017-08-04
Ensure no java.lang.UnsupportedClassVersionError are thrown if runnin…

normanmaurer:
…g on Java7 and try to check if conscrypt is available.

Motivation:

We need to ensure we not try to load any conscrypt classes directly (which means without using reflection) in the same class that is used to check if conscrypt is available. This is needed as otherwise we will have the following problem when try to use netty on java7:

java.lang.UnsupportedClassVersionError: org/conscrypt/BufferAllocator : Unsupported major.minor version 52.0
	at io.netty.handler.ssl.ConscryptJdkSslEngineInteropTest.checkConscrypt(ConscryptJdkSslEngineInteropTest.java:49)

This regression was introduced by 4448b8f42f599e79db1744cd8f5fdfee702c195e and detected on the CI when using:

mvn clean package -DtestJavaHome=$JAVA7_HOME

Modifications:

Move the detection code in an extra class and use it.

Result:

Works correctly also when using Java7.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/Conscrypt.java
+++ b/handler/src/main/java/io/netty/handler/ssl/Conscrypt.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.ssl;
+
+import io.netty.util.internal.PlatformDependent;
+
+import javax.net.ssl.SSLEngine;
+import java.lang.reflect.Method;
+
+/**
+ * Contains methods that can be used to detect if conscrypt is usable.
+ */
+final class Conscrypt {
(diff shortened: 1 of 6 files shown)


normanmaurer:
@nmittler PTAL as it was your commit that introduced the regression. Beers on you =P

handler/src/main/java/io/netty/handler/ssl/Conscrypt.java
```diff
@@ -0,0 +1,70 @@
...
+import io.netty.util.internal.PlatformDependent;
+
+import javax.net.ssl.SSLEngine;
+import java.lang.reflect.Method;
+
+/**
+ * Contains methods that can be used to detect if conscrypt is usable.
+ */
+final class Conscrypt {
+    // This class must ensure it not try to load any conscrypt classes without using reflection to also allow
```
Scottmitch: `it not try` -> `it does not try`

Scottmitch: Consider rewording and adding to class level javadocs:

This class exists to avoid loading other conscrypt related classes using features only available in JDK8+, because we need to maintain JDK6+ runtime compatibility and JDK7+ compile compatibility.

normanmaurer: actually we need jdk6+ runtime compatibility only ;). The issue happens because when we try to load the class in java7 it tries to load the conscrypt classes which will fail as conscrypt is compiled with java8 ;) But yeah let me reword

normanmaurer: reworded

normanmaurer:
let me just merge this one... as @Scottmitch approved.

normanmaurer:
Cherry-picked into 4.1 (32f497760f4fc18b797d53ed26b690f268f2ee2b) and 4.0 (04df3e21ed33bb7e05103bfdbc413480a3963391)

Issue #6969, opened 2017-07-14
CorruptedFrameException when streaming Json array over multiple writes


violetagg:
### Expected behavior
One should be able to stream Json array over multiple writes

### Actual behavior
The exception is thrown
```
io.netty.handler.codec.CorruptedFrameException: invalid JSON received at byte position 43: 5d202020202020
	at io.netty.handler.codec.json.JsonObjectDecoder.decode(JsonObjectDecoder.java:170)
	at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:489)
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:428)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:265)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1334)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
	at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:926)
	at io.netty.channel.embedded.EmbeddedChannel.writeInbound(EmbeddedChannel.java:274)
	at io.netty.handler.codec.json.JsonObjectDecoderTest.testStreamJsonArrayOverMultipleWrites(JsonObjectDecoderTest.java:89)
```

### Steps to reproduce
The test case linked to this issue reproduces the exception above.
This commit https://github.com/netty/netty/commit/231e6a5b7d92a7455b5c64f8e4a71a6fcc73cfd1 is the cause for the problem.
If I comment the `reset` invocation the scenario is working again.
https://github.com/netty/netty/blob/4.1/codec/src/main/java/io/netty/handler/codec/json/JsonObjectDecoder.java#L95

[...]

PR #7050, opened 2017-08-04
Handle partially decoded elements while streaming Json array

violetagg:
Motivation:

`insideString` and `openBraces` need a proper handling when streaming
Json array over multiple writes and an element decoding was started but
not completed.
Related to #6969

Modifications:

If the `idx` is reset:
- `insideString` has to be reset to `false` in order to indicate that
  array element will be decoded from the beginning
- `openBraces` has to be reset to `1` to indicate that Json array
  decoding is in progress.

Result:
Json array is properly decoded when in streaming mode


normanmaurer:
@Scottmitch PTAL as well.

normanmaurer:
Cherry-picked into 4.1 (db4781282f364b3a57b829db4362d77f449b612c). Thanks @violetagg

PR #7051, opened 2017-08-04
Fix incorrect javadocs in Http2RemoteFlowController

normanmaurer:
Motivation:

The javadocs of Http2RemoteFlowController.isWritable(...) are incorrect.

Modifications:

Update javadocs to reflect reality.

Result:

Correct javadocs.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2RemoteFlowController.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2RemoteFlowController.java
@@ -67,11 +67,12 @@ public interface Http2RemoteFlowController extends Http2FlowController {
     /**
      * Determine if the {@code stream} has bytes remaining for use in the flow control window.
      * <p>
-     * Note that this only takes into account HTTP/2 flow control. It does <strong>not</strong> take into account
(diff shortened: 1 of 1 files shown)


normanmaurer:
@ejona86 FYI... found this while working on the http2 child channel pr.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2RemoteFlowController.java
```diff
@@ -67,11 +67,11 @@
     /**
      * Determine if the {@code stream} has bytes remaining for use in the flow control window.
      * <p>
-     * Note that this only takes into account HTTP/2 flow control. It does <strong>not</strong> take into account
-     * the underlying {@link io.netty.channel.Channel#isWritable()}.
+     * Note that it will take into account channel writability, and so will return {@code false} if
```
Scottmitch: `Note that it will take` -> `Note that this method respects channel writability. The channel must be writable for this method to return {@code true}.`

normanmaurer: addressed

normanmaurer:
Addressed @Scottmitch s comment and cherry-picked into 4.1 (348745608)

PR #7052, opened 2017-08-04
asm 6.0_BETA was released so we should use it when building on java9

normanmaurer:
Motivation:

We used asm 6.0_ALPHA when building on java9 as the latest stable release not works with java9. asm 6.0_BETA was just released so we should update.

Modifications:

Upgrade asm version

Result:

Not use ALPHA release anymore


Scottmitch:
test failure seems unrelated. see https://github.com/netty/netty/issues/7055

normanmaurer:
Cherry-picked into 4.1 (bd029503ae60ea5583879e1e1de597e49ebb6069) and 4.0 (a527bdb6c6f87f2bc381b486b6031e5e1cb2c404)

Issue #6695, opened 2017-05-03
io.netty.handler.codec.EncoderException: MessageToMessageCodec$1 must produce at least one message.


mskaric:
I'm not sure whether this is an error or normal behavior. But i am sure that the error message is misleading.
### Expected behavior
No error
### Actual behavior
io.netty.handler.codec.EncoderException: MessageToMessageCodec$1 must produce at least one message.
### Steps to reproduce
try to write empty string only like this:
ctx.write(new DefaultHttpContent(Unpooled.copiedBuffer("", CharsetUtil.UTF_8)), ctx.voidPromise());

### Netty version
4.1.10
### JVM version (e.g. `java -version`)
java version "1.8.0_131"
Java(TM) SE Runtime Environment (build 1.8.0_131-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.131-b11, mixed mode)
### OS version (e.g. `uname -a`)
linux 4.4.0-71-generic #92-Ubuntu SMP

normanmaurer:
Can you also show your codec impl or share which included codec was used?

mskaric:
Of course I can Norman. That is the least I can do.
The problem shows itself when I use chunked transfer e.g. I don't use content-length header.
I think the problem is because 0 length chunk should be the last chunk.
My workaround is to check whether business layer is sending 0 length chunk and ignore it completely but I think this solution should be implemented in HttpResponseEncoder class.

 ChannelPipeline p = ch.pipeline();
        if (sslCtx != null) {
            p.addLast(sslCtx.newHandler(ch.alloc()));
        }        
        p.addLast("TIMER", new ReadTimeoutHandler(30));                        
        p.addLast("DECODER",new HttpRequestDecoder());        
        p.addLast("AGGREGATOR",new HttpObjectAggregator(4194304)); //4MB
        p.addLast("ENCODER",new HttpResponseEncoder());        
        p.addLast("COMPRESOR",new WebContentCompressor());        
        p.addLast(dataEventExecutorGroup,new WebServerHandler(handler));

mskaric:
This is the solution I am using:

public void write(ByteBuf buf) {
        **if(buf.writerIndex()==buf.readerIndex())
            return;**
        if (writeResponse) {
            ctx.write(response);
            writeResponse = false;
        }
        ctx.write(new DefaultHttpContent(buf), ctx.voidPromise());
    }

leopinho:
[...]

PR #7057, opened 2017-08-06
Add testcases to prove HttpResponseEncoder correctly handles empty co…

normanmaurer:
…ntent

Motivation:

Issue #6695 states that there is an issue when writing empty content via HttpResponseEncoder.

Modifications:

Add two test-cases.

Result:

Verified that all works as expected.


normanmaurer:
Cherry-picked into 4.1 (918fde82f8efa64a6f9d3abb8e108c773f2ccf46) and 4.0 (0b1b39adb1)

Issue #3756, opened 2015-05-07
Duplicate code in NativeLibraryLoader


lime62:
`NativeLibraryLoader` has some functions which are similar to `PlatformDependent`'s.

For example, 
- tmpdir()
- toDirectory()
- isWindow()

I wonder if these can be replaced with functions of `PlatformDependent`.
WDYT?

normanmaurer:
Fixed by https://github.com/netty/netty/pull/7058

PR #7058, opened 2017-08-06
Remove code-duplication from NativeLibraryLoader

normanmaurer:
Motivation:

NativeLibraryLoader has some code-duplication that can be removed.

Modifications:

Remove duplicated code and just use provided methods of PlatformDependent.

Result:

Less code duplication, fixes [#3756].

### DIFF
--- a/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
+++ b/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
@@ -31,7 +31,6 @@ import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.Arrays;
 import java.util.EnumSet;
-import java.util.Locale;
 import java.util.Set;
 
 /**
@@ -43,13 +42,10 @@ public final class NativeLibraryLoader {
     private static final InternalLogger logger = InternalLoggerFactory.getInstance(NativeLibraryLoader.class);
 
     private static final String NATIVE_RESOURCE_HOME = "META-INF/native/";
-    private static final String OSNAME;
     private static final File WORKDIR;
     private static final boolean DELETE_NATIVE_LIB_AFTER_LOADING;
 
     static {
-        OSNAME = SystemPropertyUtil.get("os.name", "").toLowerCase(Locale.US).replaceAll("[^a-z0-9]+", "");
-
         String workdir = SystemPropertyUtil.get("io.netty.native.workdir");
         if (workdir != null) {
             File f = new File(workdir);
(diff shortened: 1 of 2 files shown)


common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -1075,7 +1075,7 @@ private static File tmpdir0() {
     }
 
     @SuppressWarnings("ResultOfMethodCallIgnored")
-    private static File toDirectory(String path) {
+    static File toDirectory(String path) {
```
Scottmitch: is this change necessary?

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -43,13 +43,13 @@
...
-    private static final String OSNAME;
+    private static final boolean OSX;
     private static final File WORKDIR;
     private static final boolean DELETE_NATIVE_LIB_AFTER_LOADING;
 
     static {
-        OSNAME = SystemPropertyUtil.get("os.name", "").toLowerCase(Locale.US).replaceAll("[^a-z0-9]+", "");
-
+        String osname = SystemPropertyUtil.get("os.name", "").toLowerCase(Locale.US);
+        OSX = osname.startsWith("macosx") || osname.startsWith("osx");
```
Scottmitch: would it make sense to move the OS X platform detection to `PlatformDependent`?

common/src/main/java/io/netty/util/internal/PlatformDependent.java
normanmaurer: yes as we use it in ´NativeLibraryLoader´

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
normanmaurer: sure we can do but as there were no other places where this is used I did not bother.

common/src/main/java/io/netty/util/internal/PlatformDependent.java
Scottmitch: It looks like you removed all usages in this PR, or am I missing something?

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
Scottmitch: Yip understood. My rational is if we need todo macOS specific stuff in the future `PlatformDependent` is the natural place to look, but we may miss it (and duplicate it) if the logic lives in this class. Since `PlatformDependent` is internal and duplication removal was the motivation for this PR it seemed like it warranted a discussion.

common/src/main/java/io/netty/util/internal/PlatformDependent.java
normanmaurer: if that is true... which may be true that at the end I was able to remove all of them you are right... let me check

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
normanmaurer: true, let me move it.

common/src/main/java/io/netty/util/internal/PlatformDependent.java
normanmaurer: ok you are right... thanks :)

normanmaurer:
Cherry-picked into 4.1 (bdb0a39c8a) and 4.0 (1a0ef664c5d5371654c5a1680fb11df0ddd5efeb)

PR #7059, opened 2017-08-07
Http2FrameLogger frame labels incorrect

Scottmitch:
Motivation:
The labels identifying the frame types in Http2FrameLogger are not always correct.

Modification:
- Correct the string labels to indicate the right frame type in Http2FrameLogger

Result:
Logs are more correct.


Scottmitch:
4.1 (0f8ecdcad)

PR #7060, opened 2017-08-07
Immediate caching the wrapped string value in AsciiString

fenik17:
Motivation:
The `AsciiString#toString` method calculate string value and cache it into field. If an `AsciiString` created from the `String` value, we can avoid rebuilding strings if we cache them immediately when creating `AsciiString`. It would be useful for constants strings, which already stored in the JVMs string table, or in cases where an unavoidable `#toString `method call is assumed.

Modifications:
- Add new static method `AsciiString#cache(String)` which save string value into cache field.
- Apply a "benign" data race in the `#hashCode` and `#toString` methods.

Result:
Less memory usage in some `AsciiString` use cases.

### DIFF
--- a/common/src/main/java/io/netty/util/AsciiString.java
+++ b/common/src/main/java/io/netty/util/AsciiString.java
@@ -45,7 +45,7 @@ import static io.netty.util.internal.ObjectUtil.checkNotNull;
  * {@link #arrayChanged()} so the state of this class can be reset.
  */
 public final class AsciiString implements CharSequence, Comparable<CharSequence> {
-    public static final AsciiString EMPTY_STRING = new AsciiString("");
+    public static final AsciiString EMPTY_STRING = cached("");
     private static final char MAX_CHAR_VALUE = 255;
 
     public static final int INDEX_NOT_FOUND = -1;
@@ -1121,10 +1121,12 @@ public final class AsciiString implements CharSequence, Comparable<CharSequence>
      */
     @Override
     public int hashCode() {
-        if (hash == 0) {
-            hash = PlatformDependent.hashCodeAscii(value, offset, length);
+        int h = hash;
+        if (h == 0) {
+            h = PlatformDependent.hashCodeAscii(value, offset, length);
+            hash = h;
         }
-        return hash;
+        return h;
     }
 
     @Override
@@ -1148,11 +1150,12 @@ public final class AsciiString implements CharSequence, Comparable<CharSequence>
      */
     @Override
     public String toString() {
-        if (string != null) {
-            return string;
+        String cache = string;
+        if (cache == null) {
+            cache = toString(0);
+            string = cache;
         }
-        string = toString(0);
-        return string;
+        return cache;
     }
 
     /**
@@ -1382,6 +1385,18 @@ public final class AsciiString implements CharSequence, Comparable<CharSequence>
      */
     public static AsciiString of(CharSequence string) {
         return string.getClass() == AsciiString.class ? (AsciiString) string : new AsciiString(string);
+    }
+
+    /**
+     * Returns an {@link AsciiString} containing the given string and retains/caches the input
+     * string for later use in {@link #toString()}.
+     * Used for the constants (which already stored in the JVM's string table) and in cases
+     * where the guaranteed use of the {@link #toString()} method.
+     */
+    public static AsciiString cached(String string) {
+        AsciiString asciiString = new AsciiString(string);
+        asciiString.string = string;
+        return asciiString;
     }
 
     /**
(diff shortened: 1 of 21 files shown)


Scottmitch:
Build failure looks unrelated https://github.com/netty/netty/issues/7062.

This may not necessarily mean less memory usage. It may mean more depending upon the use case. If someone converts from `String` to `AsciiString` and doesn't go back to `String` then this will actually retain more memory. Do you have a more specific use case in mind?

fenik17:
@Scottmitch in Netty, in the vast majority of cases, an `AsciiString` which is created from the string, is going back to the `String` (e.g. `HttpConversionUtils.toHttp2Path`) or created from constants (e.g. `HttpHeaderNames`) which already stored in jvms string table.

As an alternative, we can add a new constructor `AsciiString(String value, boolean immediateCacheValue)` and use it in Netty. WDYT?

Scottmitch:
> (e.g. HttpConversionUtils.toHttp2Path) 

Are you referring to these translations like [1]? There are some of the pseudo headers from HTTP/2 which are represented as `String` in the H1 APIs. However note that these cases typically don't start from a string, and are parsed as bytes from the wire (or cached in hpack). I wouldn't expect lazy string creation to have much impact here. Also this conversion utility is only required if going from H2 types to H1 types which is not a requirement of either codec.

[1] https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/HttpConversionUtil.java#L276

> or created from constants

Hum this is a good point if the strings are already retained in the static tables there is no need to duplicate these. However I don't think this will have too much impact because after the first use they should be cached ... unless there is some threading visibility issue (which this PR may not address).

> in the vast majority of cases

I don't have enough evidence of this yet. For example one concern is the [HttpHeaders#newEntity](https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaders.java#L1168) which is the header creation API from 4.0. However this use case also cached bytes and `String` [2] so maybe its OK.

[2] https://github.com/netty/netty/blob/4.0/codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderEntity.java#L35

@normanmaurer - any thoughts?

fenik17:
@Scottmitch thanks!
I totally agree with your point. PR has been updated. PTAL again, please :)

common/src/main/java/io/netty/util/AsciiString.java
```diff
@@ -1121,10 +1121,12 @@ public boolean matches(String expr) {
      */
     @Override
     public int hashCode() {
-        if (hash == 0) {
-            hash = PlatformDependent.hashCodeAscii(value, offset, length);
+        int h = hash;
```
Scottmitch: is the temporary variable change necessary? The member variable `hash` is set as a result of the computation, and not used during the computation ... so I wouldn't expect this to have any impact.

common/src/main/java/io/netty/util/AsciiString.java
```diff
@@ -1385,6 +1388,19 @@ public static AsciiString of(CharSequence string) {
     }
 
     /**
+     * Returns an {@link AsciiString} containing the given string.
```
Scottmitch: `containing the given string.` -> `and retains/caches the input string for later use in {@link #toString()}`

common/src/main/java/io/netty/util/AsciiString.java
fenik17: It's fix for a non-benign data race. See [1], [2].

[1] https://shipilev.net/blog/2016/close-encounters-of-jmm-kind/#wishful-benign-is-resilient
[2] http://openjdk.5641.n7.nabble.com/RFR-8166842-String-hashCode-has-a-non-benign-data-race-td286991.html

common/src/main/java/io/netty/util/AsciiString.java
fenik17: fixed

common/src/main/java/io/netty/util/AsciiString.java
normanmaurer: @fenik17 interesting read... that said I think it would only be problematic if we would have a constructor that directly set the `hash` or I am wrong ? That said I think it not harm so I am not against this change :)

fenik17: > it would only be problematic if we would have a constructor that directly set the hash

No. This is about a situation where two threads at the same time are call the `hashCode()` (or `toString()`) method. Theoretically, one of them can get wrong result `0` (or `null`) due to double reading from non-volatile/non-final field, because the compiler can reorder the field reads.

normanmaurer: got it... let me pull this in

normanmaurer:
Cherry-picked into 4.1 (4875a2aad4c82204caedb2d121fb5ef03c09fe44) . @fenik17 thanks

PR #7064, opened 2017-08-08
Ensure we null out the previous set InetAddress on java.net.DatagramP…

normanmaurer:
…acket when using OioDatagramChannel.

Motivation:

We need to ensure we always null out (or set) the address on the java.net.DatagramPacket when doing read or write operation as the same instance is used across different calls.

Modifications:

Null out the address if needed.

Result:

Ensure the correct remote address is used when connect / disconnect between calls and also mix these with calls that directly specify the remote address for adatagram packets.


normanmaurer:
@Scottmitch noticed this while browsing the source for something else.

normanmaurer:
Cherry-picked into 4.1 (85f5d6bf0521732358ed6dc5ca6fbcbda5fc207b) and 4.0 (e4c85537aeae0105b801be5dd25031e80e615b2e)

PR #7065, opened 2017-08-08
We should prefer heap buffers when using the OIO transport to reduce …

normanmaurer:
…memory copies.

Motivation:

When using the OIO transport we need to act on byte[] when writing and reading from / to the underyling Socket. So we should ensure we use heap buffers by default to reduce memory copies.

Modifications:

Ensure we prefer heap buffers by default for the OIO transport.

Result:

Possible less memory copies.

### DIFF
--- a/transport/src/main/java/io/netty/channel/local/LocalChannel.java
+++ b/transport/src/main/java/io/netty/channel/local/LocalChannel.java
@@ -24,6 +24,7 @@ import io.netty.channel.ChannelPipeline;
 import io.netty.channel.ChannelPromise;
 import io.netty.channel.DefaultChannelConfig;
 import io.netty.channel.EventLoop;
+import io.netty.channel.PreferHeapByteBufAllocator;
 import io.netty.channel.SingleThreadEventLoop;
 import io.netty.util.ReferenceCountUtil;
 import io.netty.util.concurrent.Future;
--- a/transport-rxtx/src/main/java/io/netty/channel/rxtx/DefaultRxtxChannelConfig.java
+++ b/transport-rxtx/src/main/java/io/netty/channel/rxtx/DefaultRxtxChannelConfig.java
@@ -19,6 +19,7 @@ import io.netty.buffer.ByteBufAllocator;
 import io.netty.channel.ChannelOption;
 import io.netty.channel.DefaultChannelConfig;
 import io.netty.channel.MessageSizeEstimator;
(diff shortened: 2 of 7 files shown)


normanmaurer:
Noticed this when looking for something else.

transport-rxtx/src/main/java/io/netty/channel/rxtx/RxtxChannel.java
```diff
@@ -20,6 +20,7 @@
 import gnu.io.SerialPort;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelPromise;
+import io.netty.channel.PreferHeapByteBufAllocator;
```
netkins: MINOR Remove this unused import 'io.netty.channel.PreferHeapByteBufAllocator'. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

transport/src/main/java/io/netty/channel/local/LocalChannel.java
```diff
@@ -24,6 +24,7 @@
 import io.netty.channel.ChannelPromise;
 import io.netty.channel.DefaultChannelConfig;
 import io.netty.channel.EventLoop;
+import io.netty.channel.PreferHeapByteBufAllocator;
```
Scottmitch: Can we get rid of `io.netty.channel.local.PreferHeapByteBufAllocator` now?

normanmaurer: nope it was moved the `io.netty.channel` (it was in the local package before) as we use it in OIO as well now.

normanmaurer:
@Scottmitch no idea... hopefully not too many ;)

transport/src/main/java/io/netty/channel/local/LocalChannel.java
Scottmitch: It doesn't look like it was "moved", but instead "copied" and there are 2 of them now. I'm wondering if we can get rid of the old one (`io.netty.channel.local.PreferHeapByteBufAllocator`) which no longer appears to be used.

normanmaurer: If this is the case my git foo is weak and I messed up... let me check

normanmaurer:
Cherry-picked into 4.1 (d0c43c9e42d393ff0cc9f1ab97551e5778ec0b7e)

PR #7077, opened 2017-08-10
ISSUE-7076 Fix NPE in MQTT Connect

### LINKED ISSUE #7076: MQTT NPE if password or willMessage is null.
### Expected behavior
methods should return null not NPE

### Actual behavior
throws NPE

### Steps to reproduce
Send MQTT connect message, with a null password, and then call .password or .willMesage
https://github.com/netty/netty/pull/6817
This is caused by the following change where null byte[] for these two fields is not being null handled.

Work around is to use new methods and null check manually, but would be good to save random bugs to fix the changed logic in deprecated methods.

      String password = connect.payload().passwordInBytes() == null ? null : new String(connect.payload().passwordInBytes(), CharsetUtil.UTF_8);
      String willMessage = connect.payload().willMessageInBytes() == null ? null : new String(connect.payload().willMessageInBytes(), CharsetUtil.UTF_8);

### Netty version
4.1.13

michaelandrepearce:
Add null checks before converting to string to avoid NPE.

Motivation:

Fixing an introduced NPE by #6817

Modification:

Add null checks.

Result:

Fixes #7076 .

### DIFF
--- a/codec-mqtt/src/test/java/io/netty/handler/codec/mqtt/MqttConnectPayloadTest.java
+++ b/codec-mqtt/src/test/java/io/netty/handler/codec/mqtt/MqttConnectPayloadTest.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.netty.handler.codec.mqtt;
+
+import static org.junit.Assert.assertNull;
+
+import io.netty.util.CharsetUtil;
+import org.junit.Test;
+
+public class MqttConnectPayloadTest {
+
+    @Test
+    public void testNullWillMessage() throws Exception {
+        String clientIdentifier = "clientIdentifier";
+        String willTopic = "willTopic";
+        byte[] willMessage = null;
+        String userName = "userName";
+        byte[] password = "password".getBytes(CharsetUtil.UTF_8);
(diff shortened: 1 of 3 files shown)


normanmaurer:
@michaelandrepearce can you please sign our ICLA and let me know once done:

https://netty.io/s/icla

Thanks!

normanmaurer:
@michaelandrepearce also would be nice if you could just add some simple unit tests for this :)

michaelandrepearce:
Already signed please check

codec-mqtt/src/test/java/io/netty/handler/codec/mqtt/MqttConnectPayloadTest.java
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright 2014 The Netty Project
```
normanmaurer: @michaelandrepearce please change to 2017

codec-mqtt/src/test/java/io/netty/handler/codec/mqtt/MqttConnectPayloadTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+        Assert.assertNull(mqttConnectPayload.willMessageInBytes());
+        Assert.assertNull(mqttConnectPayload.willMessage());
+
+        builder = new MqttMessageBuilders.ConnectBuilder();
+        builder.willMessage((byte[]) null);
+
+        mqttConnectPayload = builder.build().payload();
+
+        Assert.assertNull(mqttConnectPayload.willMessageInBytes());
```
normanmaurer: consider add static import for `Assert.assertNull`

codec-mqtt/src/test/java/io/netty/handler/codec/mqtt/MqttConnectPayloadTest.java
michaelandrepearce: done.

codec-mqtt/src/test/java/io/netty/handler/codec/mqtt/MqttConnectPayloadTest.java
michaelandrepearce: Done.

michaelandrepearce:
Have just pushed both.

normanmaurer:
Cherry-picked into 4.1 (1ba592049c).

@michaelandrepearce thanks!

Issue #7062, opened 2017-08-08
EpollDomainSocketGatheringWriteTest#testGatheringWriteBig failure


Scottmitch:
Observed while building PR https://github.com/netty/netty/pull/7060

https://garage.netty.io/teamcity/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=23077#_focus=132185&state=132185

```
[22:16:05][io.netty:netty-transport-native-epoll] io.netty.channel.epoll.EpollDomainSocketGatheringWriteTest
[22:16:05][io.netty.channel.epoll.EpollDomainSocketGatheringWriteTest] testGatheringWriteWithCompositeNotAutoRead
[22:16:05][io.netty.channel.epoll.EpollDomainSocketGatheringWriteTest] testGatheringWriteBig
[22:16:05]
[testGatheringWriteBig] [Test Output]
00:11:59.243 [Time-limited test] INFO  i.n.c.e.EpollDomainSocketGatheringWriteTest - Running: testGatheringWriteBig 1 of 1 (ServerBootstrap(ServerBootstrapConfig(group: EpollEventLoopGroup, channelFactory: EpollServerDomainSocketChannel.class, localAddress: /tmp/NETTY3571910213550174173UDS, options: {ALLOCATOR=UnpooledByteBufAllocator(directByDefault: true)}, childGroup: EpollEventLoopGroup, childOptions: {ALLOCATOR=UnpooledByteBufAllocator(directByDefault: true)})) + Bootstrap(BootstrapConfig(group: EpollEventLoopGroup, channelFactory: EpollDomainSocketChannel.class, options: {ALLOCATOR=UnpooledByteBufAllocator(directByDefault: true)}, resolver: io.netty.resolver.DefaultAddressResolverGroup@1a5ee7a3))) with UnpooledByteBufAllocator
[...]

PR #7084, opened 2017-08-10
SocketGatheringWriteTest improvements

Scottmitch:
Motivation:
SocketGatherWriteTest has been observed to fail and it has numerous issues which when resolved may help reduce the test failures.

Modifications:
- A volatile counter and a spin/sleep loop is used to trigger test termination. Incrementing a volatile is generally bad practice and can be avoided in this situation. This mechanism can be replaced by a promise. This mechanism should also trigger upon exception or channel inactive.
- The TestHandler maintains an internal buffer, but it is not released. We now only create a buffer on the server side and release it after comparing the expected results.
- The composite buffer creation logic can be simplified, also the existing composite buffer doesn't take into account the buffer's reader index when building buf2.

Result:
Cleaner test.

### DIFF
--- a/testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketGatheringWriteTest.java
+++ b/testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketGatheringWriteTest.java
@@ -26,6 +26,8 @@ import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelOption;
 import io.netty.channel.SimpleChannelInboundHandler;
 import io.netty.testsuite.util.TestUtils;
+import io.netty.util.concurrent.ImmediateEventExecutor;
+import io.netty.util.concurrent.Promise;
 import io.netty.util.internal.StringUtil;
 import org.junit.AfterClass;
 import org.junit.Rule;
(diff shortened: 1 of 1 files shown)


Scottmitch:
Discovered this while investigating https://github.com/netty/netty/issues/7062

testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketGatheringWriteTest.java
```diff
@@ -175,49 +166,84 @@ private void testGatheringWrite0(
...
-                Throwable cause) throws Exception {
+        public final void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
             if (exception.compareAndSet(null, cause)) {
-                ctx.close();
+                handleException(ctx, cause);
             }
+            super.exceptionCaught(ctx, cause);
+        }
+
+        void handleException(ChannelHandlerContext ctx, Throwable cause) {
```
netkins: MAJOR Remove this unused method parameter "cause". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1172)

Scottmitch:
Build failure is attributed to https://github.com/netty/netty/issues/7056

normanmaurer:
Cherry-picked into 4.1 (0bf614d9e9395320f1712fc8e47b245df648fee0) and 4.0 (a044f1705a)

PR #7085, opened 2017-08-10
SocketStringEchoTest improvements

Scottmitch:
Motivation:
SocketStringEchoTest has been observed to fail and it has numerous issues which when resolved may help reduce the test failures.

Modifications:
- A volatile counter and a spin/sleep loop is used to trigger test termination. Incrementing a volatile is generally bad practice and can be avoided in this situation. This mechanism can be replaced by a promise. This mechanism should also trigger upon exception or channel inactive.
- Asserts are done in the Netty threads. Although these should result in a exceptionCaught the test may not observe these failures because it is spinning waiting for the count to reach the desired value.

Result:
Cleaner test.


Scottmitch:
Discovered while investigating https://github.com/netty/netty/issues/7056

normanmaurer:
Cherry-picked into 4. 1 (ce4b4da5838dbb343c216f51af05ce3e75b1a39a) and 4.0 (0fe8eb52bc4ef328c9e3f2c7558c1732691098c1)

PR #7086, opened 2017-08-11
Strip http 'expect' headers when expectation response is produced

### LINKED ISSUE #7075: should HttpObjectAggregator strip 'expect' header after sending 100 continue response?
There is a behavior difference in how [HttpObjectAggregator](https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectAggregator.java#L162) and [HttpServerExpectContinueHandler](https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/HttpServerExpectContinueHandler.java#L92) 
deal with requests which set 'expect: 100-continue' headers and meet expectations.

`HttpServerExpectContinueHandler` removes the expect header from the client request and `HttpObjectAggregator` does not.

In the case where `HttpObjectAggregator` has sent the 100 response it seems that there is no value in propagating the header further so it would make sense to also remove the header.

This is the behavior as of Netty 4.1.12.Final

dschobel:
Motivation:

HttpObjectAggregator differs from HttpServerExpectContinueHandler's handling
of expect headers by not stripping the 'expect' header when a response
is generated.

Modifications:

HttpObjectAggregator removes the 'expect' header in cases where it generates
a response.

Result:

Consistent and correct behavior between HttpObjectAggregator and HttpServerExpectContinueHandler.

Fixes #7075.

### DIFF
--- a/codec-http/src/test/java/io/netty/handler/codec/http/HttpObjectAggregatorTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/HttpObjectAggregatorTest.java
@@ -371,6 +371,23 @@ public class HttpObjectAggregatorTest {
     }
 
     @Test
+    public void testValidRequestWith100ContinueAndDecoder() {
+        EmbeddedChannel embedder = new EmbeddedChannel(new HttpRequestDecoder(), new HttpObjectAggregator(100));
+        embedder.writeInbound(Unpooled.copiedBuffer(
+            "GET /upload HTTP/1.1\r\n" +
+                "Expect: 100-continue\r\n" +
+                "Content-Length: 0\r\n\r\n", CharsetUtil.US_ASCII));
+
+        FullHttpResponse response = embedder.readOutbound();
+        assertEquals(HttpResponseStatus.CONTINUE, response.status());
+        FullHttpRequest request = embedder.readInbound();
+        assertFalse(request.headers().contains(HttpHeaderNames.EXPECT));
+        request.release();
+        response.release();
+        assertFalse(embedder.finish());
+    }
+
+    @Test
     public void testOversizedRequestWith100ContinueAndDecoder() {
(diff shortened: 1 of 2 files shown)


codec-http/src/test/java/io/netty/handler/codec/http/HttpObjectAggregatorTest.java
```diff
@@ -371,6 +371,21 @@ private static void runUnsupportedExceptHeaderExceptionTest(final boolean close)
...
 
     @Test
+    public void testValidRequestWith100ContinueAndDecoder() {
+        EmbeddedChannel embedder = new EmbeddedChannel(new HttpRequestDecoder(), new HttpObjectAggregator(100));
+        embedder.writeInbound(Unpooled.copiedBuffer(
+            "GET /upload HTTP/1.1\r\n" +
+                "Expect: 100-continue\r\n" +
+                "Content-Length: 0\r\n\r\n", CharsetUtil.US_ASCII));
+
+        FullHttpResponse response = embedder.readOutbound();
```
normanmaurer: @dschobel you need to call `response.release()` as well.

codec-http/src/test/java/io/netty/handler/codec/http/HttpObjectAggregatorTest.java
```diff
@@ -371,6 +371,21 @@ private static void runUnsupportedExceptHeaderExceptionTest(final boolean close)
...
+        assertEquals(HttpResponseStatus.CONTINUE, response.status());
+        FullHttpRequest request = embedder.readInbound();
+        assertFalse(request.headers().contains(HttpHeaderNames.EXPECT));
+        request.release();
```
normanmaurer: can you also call `embedder.finish()` and assert that it returns false ?

normanmaurer:
Cherry-picked into 4.1 (52e19d5c631d9cbe6cc41b5ea34d7493160f26ee).

@dschobel thanks!

PR #7088, opened 2017-08-11
SocketGatheringWriteTest increase timeouts

Scottmitch:
Motivation:
EpollDomainSocketGatheringWriteTest. testGatheringWriteBig takes on average about 20-25 seconds on the CI servers, but there is a 30 second timeout being applied which leads to what maybe false positive test failures.

Modifications:
- Increase the test timeout to 120 seconds globally and 60 seconds to wait for all writes per test

Result:
Higher timeout for potentially less false positive test failures.


Scottmitch:
Related to #7062

normanmaurer:
@Scottmitch ship it

Scottmitch:
4.1 (8cb5d0fa8) 4.0 (8e016eddf)

PR #7090, opened 2017-08-11
Ensure we call promise.setUncancellable() before trying to process in…

normanmaurer:
… DefaultHttp2StreamChannel.

Motivation:

We should call promise.setUncancellable() in DefaultHttp2StreamChannel.Unsafe impl to detect if the operation was cancelled.

Modifications:

Add promise.setUncancellable() calls

Result:

More correct handling of cancelled promises


normanmaurer:
Cherry-picked into 4.1 (08284dbbcd4f4e4a3ad7756cbb9eb71d180fcdfa)

Issue #7069, opened 2017-08-08
Loading epoll transport from java.library.path fails


eun-ice:
### Expected behavior

On restricted linux systems I want to use epoll transport without the JVM being able to write an .so and load it into JVM. For example grsecurity hardened systems do only allow this if the JVM is running as a trusted user (usually root)

### Actual behavior

Putting libnetty-tcnative-linux-x86_64.so in java.library.path works fine, but libnetty_transport_native_epoll.so fails to load.

### Steps to reproduce

Put libnetty_transport_native_epoll.so and libnetty-tcnative-linux-x86_64.so in a path specified in -Djava.library.path (or LD_LIBRARY_PATH env variable).

Start the JVM with a -Dio.netty.native.workdir=/xyz that is mounted with "noexec" with any Netty code using Epoll.

What happens:

libnetty_transport_native_epoll.so fails to load with:

java.lang.UnsatisfiedLinkError: unsupported JNI version 0xFFFFFFFF required by /.../lib/libnetty_transport_native_epoll.so

Initialization of the .so fails, because netty_unix_util_parse_package_prefix in netty_unix_util.c from transport-native-unix-common module looks for "netty-transport-native-epoll" in the file name. However, if the .so is named libnetty-transport-native-epoll.so, it is not loaded at all.

### Minimal yet complete reproducer code (or URL to code)

Any epoll example.

### Netty version

Current trunk.

### JVM version (e.g. `java -version`)

Java(TM) SE Runtime Environment (build 1.8.0_131-b11)

### OS version (e.g. `uname -a`)

Linux 4.9 with grsecurity

### Proposed fixes

I can think of two ways fixing this

1) Modify io.netty.channel.epoll.Native to properly load the lib if it is named libnetty-transport-native-epoll.so

2) Modify netty_unix_util.c to also work if the name of the .so is libnetty_transport_native_epoll.so.

This pull request implements solution 2): https://github.com/netty/netty/pull/7068

Scottmitch:
IIRC @normanmaurer you were going to submit a followup PR to only support the `_` names as a followup to @carl-mastrangelo's PR https://github.com/netty/netty/pull/6884 (see https://github.com/netty/netty/pull/6884#issuecomment-310574798). Once this is done I would imagine this quirk will go away?

eun-ice:
[...]

PR #7091, opened 2017-08-11
Use underscores in our native library names to allow loading on restr…

normanmaurer:
Motivation:

At the moment we try to load the library using multiple names which includes names using - but also _ . We should just use _ all the time.

Modifications:

Replace - with _

Result:

Fixes [#7069]

### DIFF
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/Native.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/Native.java
@@ -188,11 +188,7 @@ public final class Native {
         if (!name.startsWith("linux")) {
             throw new IllegalStateException("Only supported on Linux");
         }
-        String []libraryNames = new String[] {
-          "netty-transport-native-epoll",
-          "netty_transport_native_epoll"
-        };
-        NativeLibraryLoader.loadFirstAvailable(PlatformDependent.getClassLoader(Native.class), libraryNames);
+        NativeLibraryLoader.load("netty_transport_native_epoll", PlatformDependent.getClassLoader(Native.class));
     }
 
     private Native() {
--- a/transport-native-kqueue/src/main/java/io/netty/channel/kqueue/Native.java
+++ b/transport-native-kqueue/src/main/java/io/netty/channel/kqueue/Native.java
@@ -100,11 +100,7 @@ final class Native {
         if (!name.startsWith("mac") && !name.contains("bsd") && !name.startsWith("darwin")) {
             throw new IllegalStateException("Only supported on BSD");
         }
-        String []libraryNames = new String[] {
-            "netty-transport-native-kqueue",
-            "netty_transport_native_kqueue"
-        };
-        NativeLibraryLoader.loadFirstAvailable(PlatformDependent.getClassLoader(Native.class), libraryNames);
+        NativeLibraryLoader.load("netty_transport_native_kqueue", PlatformDependent.getClassLoader(Native.class));
     }
 
     private Native() {
--- a/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
+++ b/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
@@ -111,7 +111,7 @@ public final class NativeLibraryLoader {
      */
     public static void load(String originalName, ClassLoader loader) {
         // Adjust expected name to support shading of native libraries.
-        String name = calculatePackagePrefix().replace('.', '-') + originalName;
+        String name = calculatePackagePrefix().replace('.', '_') + originalName;
 
         String libname = System.mapLibraryName(name);
         String path = NATIVE_RESOURCE_HOME + libname;
--- a/transport-native-epoll/pom.xml
+++ b/transport-native-epoll/pom.xml
@@ -144,6 +144,7 @@
               <execution>
                 <id>build-native-lib</id>
                 <configuration>
+                  <name>netty_transport_native_epoll</name>
                   <nativeSourceDirectory>${project.basedir}/src/main/c</nativeSourceDirectory>
                   <libDirectory>${project.build.outputDirectory}</libDirectory>
                   <!-- We use Maven's artifact classifier instead.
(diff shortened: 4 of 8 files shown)


normanmaurer:
@eun-ice PTAL

normanmaurer:
Also @carl-mastrangelo

johnou:
@normanmaurer netty-tcnative next on the list?

normanmaurer:
@johnou PR for tcnative is here: https://github.com/netty/netty-tcnative/pull/295

Once merged I will release it and update the netty loading code for it as well.

normanmaurer:
ups something is wrong... let me check and ping once done @Scottmitch

eun-ice:
@normanmaurer does it make sense for me to test your PRs at the moment?

normanmaurer:
@eun-ice yep you can test this one now.. Fixed the problem

normanmaurer:
@Scottmitch PTAL

eun-ice:
@normanmaurer I tried tcnative without https://github.com/netty/netty-tcnative/pull/295 since I couldn't get tcnative to work with.

With this PR netty one trys to load netty_transport_native_epoll once, but without success:

```
22:49:23,140 DEBUG io.netty.util.internal.NativeLibraryLoader:81 - Unable to load the library 'netty_transport_native_epoll', trying next name...
java.lang.UnsatisfiedLinkError: could not load a native library: netty_transport_native_epoll
        at io.netty.util.internal.NativeLibraryLoader.load(NativeLibraryLoader.java:173)
        at io.netty.util.internal.NativeLibraryLoader.loadFirstAvailable(NativeLibraryLoader.java:81)
        at io.netty.channel.epoll.Native.loadNativeLibrary(Native.java:191)
        at io.netty.channel.epoll.Native.<clinit>(Native.java:61)
        at io.netty.channel.epoll.Epoll.<clinit>(Epoll.java:33)
		...		
Caused by: java.io.IOException: Access denied
        at java.io.UnixFileSystem.createFileExclusively(Native Method)
        at java.io.File.createTempFile(File.java:2024)
        at io.netty.util.internal.NativeLibraryLoader.load(NativeLibraryLoader.java:141)
        ... 28 more

```

eun-ice:
By the way, you can easily reproduce this by setting -Dio.netty.native.workdir to something the JVM is not allowed to write at all. Then netty is forced to load from java.library.path.

E.g. on linux
`LD_LIBRARY_PATH=/path/to/libs java -jar... -Dio.netty.native.workdir=/mnt/cdrom/`

That way java will add /path/to/libs to java.library.path.

eun-ice:
I have been able to get your PR to work by adding this patch:

`diff --git a/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java b/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
index 43da6b863f..3e41873416 100644
--- a/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
+++ b/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
@@ -125,10 +125,13 @@ public final class NativeLibraryLoader {
             }
         }
 
-        if (url == null) {
-            // Fall back to normal loading of JNI stuff
+        try {
+            // first try to load from java.library.path
             loadLibrary(loader, name, false);
+            logger.info("{} was loaded from java.libary.path", name);
             return;
+            logger.info("{} cannot be loaded from java.libary.path", name, ex);
         }
 
         int index = libname.lastIndexOf('.');
`

eun-ice:
@normanmaurer Fix for NativeLibraryLoader (based on this pull request): https://github.com/netty/netty/pull/7097

Scottmitch:
@normanmaurer - please clarify the commit message with some details from https://github.com/netty/netty-tcnative/pull/295#issuecomment-321950426 to clarify the motivation.

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/Native.java
```diff
@@ -100,11 +100,8 @@ private static void loadNativeLibrary() {
         if (!name.startsWith("mac") && !name.contains("bsd") && !name.startsWith("darwin")) {
             throw new IllegalStateException("Only supported on BSD");
         }
-        String []libraryNames = new String[] {
-            "netty-transport-native-kqueue",
-            "netty_transport_native_kqueue"
-        };
-        NativeLibraryLoader.loadFirstAvailable(PlatformDependent.getClassLoader(Native.class), libraryNames);
+        NativeLibraryLoader.loadFirstAvailable(PlatformDependent.getClassLoader(Native.class),
```
Scottmitch: can we just call `load` now?

transport-native-epoll/src/main/java/io/netty/channel/epoll/Native.java
```diff
@@ -188,11 +188,8 @@ private static void loadNativeLibrary() {
         if (!name.startsWith("linux")) {
             throw new IllegalStateException("Only supported on Linux");
         }
-        String []libraryNames = new String[] {
-          "netty-transport-native-epoll",
-          "netty_transport_native_epoll"
-        };
-        NativeLibraryLoader.loadFirstAvailable(PlatformDependent.getClassLoader(Native.class), libraryNames);
+        NativeLibraryLoader.loadFirstAvailable(PlatformDependent.getClassLoader(Native.class),
```
Scottmitch: can we just call `load` now?

Scottmitch:
Build failure looks unrelated ... captured in https://github.com/netty/netty/issues/7056#issuecomment-321996849

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/Native.java
normanmaurer: yes :)

transport-native-epoll/src/main/java/io/netty/channel/epoll/Native.java
normanmaurer: yes :)

normanmaurer:
@Scottmitch PTAL again.

normanmaurer:
Cherry-picked into 4.1 (19dcb15062ff93c5968c78df2de3c8ef6a181036) and 4.0 (c1d9967780)

PR #7096, opened 2017-08-12
FIX endless loop in ByteBufUtil#writeAscii

louxiu:
Motivation:

Missing return in ByteBufUtil#writeAscii causes endless loop

Modifications:

Add return after write finished

Result:

ByteBufUtil#writeAscii is ok


normanmaurer:
Will merge once ci completes

normanmaurer:
Cherry-Picked into 4.1 (8be9a63c1c4beb3f43ffcc3ea978e67f138b5050) and 4.0 (c62cc6e24424c3cbcc49eb8beb9fbb264767154a).

@louxiu thanks!

PR #7097, opened 2017-08-12
Make NativeLibraryLoader check java.library.path first

eun-ice:
Motivation:

On restricted systems (e.g. grsecurity), it might not be possible to write a .so on disk and load it afterwards. On those system Netty should check java.library.path for libraries to load.

Modifications:

Changed NativeLibraryLoader.java to first try to load libs from java.library.path before exporting the .so to disk.

Result:

Libraries load fine on restricted systems.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
+++ b/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
@@ -125,10 +125,13 @@ public final class NativeLibraryLoader {
             }
         }
 
-        if (url == null) {
-            // Fall back to normal loading of JNI stuff
+        try {
+            // first try to load from java.library.path
             loadLibrary(loader, name, false);
+            logger.info("{} was loaded from java.libary.path", name);
             return;
+        } catch (Exception ex) {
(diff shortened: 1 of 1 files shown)


normanmaurer:
Can you please rebase and ensure the PR only contain the relevant changes?

eun-ice:
@normanmaurer It is already only one commit in my repo, see https://github.com/eun-ice/netty/commits/native_load_underscore 

However during creating the PR github silently creates a merge commit that is not necessary. So I'm stumped.

My full change is the patch in https://github.com/netty/netty/pull/7091#issuecomment-321968832

eun-ice:
@normanmaurer I think what I could do is create a PR with two commits, one being your PR from native_load_underscore and one being my patch to it. If that helps, just let me know and I will do it.

Scottmitch:
lets wait until @normanmaurer's other PRs are pulled in (should be in the next week) as there are a few changes related to renaming/reloading libraries ... it will be easier to evaluate once these other PRs are pulled in.

normanmaurer:
@eun-ice please rebase on latest 4.1 :)

eun-ice:
@normanmaurer  I created a new PR: #7102

PR #7098, opened 2017-08-12
Increase test timeout for SocketStringEchoTest

Scottmitch:
Motivation:
SocketStringEchoTest has been observed to fail on CI servers, but the stack traces still indicate work was being done.

Modifications:
- Increase the test timeout

Result:
Tests have more time to complete, and hopefully less false positive test failures.


Scottmitch:
Hopefully this helps with https://github.com/netty/netty/issues/7056

Scottmitch:
4.1 (505993a2f) 4.0 (043e84779)

Scottmitch:
We don't really develop off of 4.0 any more .. but interesting that this test didn't have any timeout on 4.0 before this change.

Issue #6576, opened 2017-03-28
ByteBuf does not have access methods for little endian floats and doubles


trustin:
### Expected behavior

`ByteBuf` has `get/read/set/writeFloatLE()` and `get/read/set/writeDoubleLE()`.

### Actual behavior

`ByteBuf` does not have the little endian variant of float/double access methods.

### Netty version

4.0.x and 4.1.x

trustin:
Probably not a good idea to add these methods to 4.1 but probably we could in 4.2+

normanmaurer:
@trustin I guess we still could as `ByteBuf` is an abstract class and just implement them directly there. We then could implement these in an optimized way in the sub-classes. WDYT ?

trustin:
Sounds good to me as long as we put some noticeable note in the announcement. :-)

normanmaurer:
@trustin want to take a crack on this or should I do ?

trustin:
Will find some time this week since this isn't an urgent one. :-)

normanmaurer:
@trustin ok cool :)

martin3361:
I came to a point in our game project where we need this and we cannot use it

+1

normanmaurer:
@martin3361 if you want to you could provide a pr for this.

martin3361:
I'm not sure if I can since I'm not familiar with endianness and how to implement it

martin3361:
Here is what I use for workaround:

	private void writeFloat(ByteBuf in, float value) {
		in.writeBytes(ByteBuffer.allocate(4).order(ByteOrder.LITTLE_ENDIAN).putFloat(value).array());
	}

normanmaurer:
Fixed by https://github.com/netty/netty/pull/7099

binakot:
I'm working on gps devices server and I really need this asap 🚜 

+1

PR #7099, opened 2017-08-12
Support the little endian floats and doubles by ByteBuf

fenik17:
Motivation:
`ByteBuf` does not have the little endian variant of float/double access methods.

Modifications:
Add support for little endian floats and doubles into `ByteBuf`.

Result:
`ByteBuf` has get/read/set/writeFloatLE() and get/read/set/writeDoubleLE() methods. Fixes [#6576].


normanmaurer:
Cherry-picked into 4.1 (0774c9145642bb9f08d9961dd0c25352dd9a944a).

@fenik17 thanks!

PR #7102, opened 2017-08-15
Make NativeLibraryLoader check java.library.path first

eun-ice:
Motivation:

On restricted systems (e.g. grsecurity), it might not be possible to write a .so on disk and load it afterwards. On those system Netty should check java.library.path for libraries to load.

Modifications:

Changed NativeLibraryLoader.java to first try to load libs from java.library.path before exporting the .so to disk.

Result:

Libraries load fine on restricted systems.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
+++ b/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
@@ -113,6 +113,17 @@ public final class NativeLibraryLoader {
         // Adjust expected name to support shading of native libraries.
         String name = calculatePackagePrefix().replace('.', '_') + originalName;
 
+        try {
+            // first try to load from java.library.path
+            loadLibrary(loader, name, false);
+            logger.debug("{} was loaded from java.libary.path", name);
+            return;
+        } catch (Throwable ex) {
+            logger.debug(
+                    "{} cannot be loaded from java.libary.path, "
+                    + "now trying export to -Dio.netty.native.workdir: {}", name, WORKDIR, ex);
+        }
+
         String libname = System.mapLibraryName(name);
         String path = NATIVE_RESOURCE_HOME + libname;
 
@@ -123,12 +134,6 @@ public final class NativeLibraryLoader {
             } else {
                 url = loader.getResource(NATIVE_RESOURCE_HOME + "lib" + name + ".jnilib");
             }
-        }
-
-        if (url == null) {
-            // Fall back to normal loading of JNI stuff
-            loadLibrary(loader, name, false);
-            return;
         }
 
         int index = libname.lastIndexOf('.');


common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -125,10 +125,13 @@ public static void load(String originalName, ClassLoader loader) {
...
 
-        if (url == null) {
-            // Fall back to normal loading of JNI stuff
+        try {
+            // first try to load from java.library.path
             loadLibrary(loader, name, false);
+            logger.info("{} was loaded from java.libary.path", name);
             return;
+        } catch (Exception ex) {
+            logger.info("{} cannot be loaded from java.libary.path", name, ex);
```
normanmaurer: @eun-ice I wonder if we should make this a debug or even only log it if it also fails to load completely as it may be confusing to the user.

eun-ice: I will make it a debug message.

eun-ice:
@normanmaurer PR updated as you requested.

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -125,10 +125,13 @@ public static void load(String originalName, ClassLoader loader) {
...
 
-        if (url == null) {
-            // Fall back to normal loading of JNI stuff
+        try {
+            // first try to load from java.library.path
             loadLibrary(loader, name, false);
+            logger.info("{} was loaded from java.libary.path", name);
             return;
+        } catch (Exception ex) {
+            logger.debug("{} cannot be loaded from java.libary.path", name, ex);
```
normanmaurer: can you add some thing like "cannot be loaded from java.libary.path, trying temporary directory now" or something like that

eun-ice: what about "cannot be loaded from java.libary.path, now trying io.netty.native.workdir"

normanmaurer: @eun-ice yes... another idea (and maybe even the better) would be to try the workdir first and only if it fails try java.library.path. This way the old "order" is maintained. WDYT ?

eun-ice: Or even 
logger.debug("{} cannot be loaded from java.libary.path, now trying export to " + WORKDIR, name, ex);

eun-ice: Regarding to the order: I think in terms of being cautious with resource usage, trying to load an existing file first before we write something to /tmp is the better approach. On my system, writing to /tmp will work but loading the lib will fail, resulting in an unneeded tmp file write.

eun-ice: @normanmaurer regarding the log statement, what abou:             
`logger.debug("{} cannot be loaded from java.libary.path, now trying export to -Dio.netty.native.workdir: {}", name, WORKDIR, ex);
`

normanmaurer: fair enough.. Yeah please update the log statement.

eun-ice:
@normanmaurer log message updated

eun-ice:
@normanmaurer had to fix the PR.

eun-ice:
I have tested on Linux with loading form /tmp and with loading from java.library.path as well as OSX loading from /tmp. Everything works fine.

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -125,10 +125,15 @@ public static void load(String originalName, ClassLoader loader) {
...
-            // Fall back to normal loading of JNI stuff
+        try {
+            // first try to load from java.library.path
             loadLibrary(loader, name, false);
+            logger.info("{} was loaded from java.libary.path", name);
             return;
+        } catch (Throwable ex) {
+            logger.debug(
+                    "{} cannot be loaded from java.libary.path, "
+                    + "now trying export to -Dio.netty.native.workdir: {}", name, WORKDIR, ex);
```
carl-mastrangelo: What happens when there are more args passed than verbs in the format string?  The Throwable at the end doesn't seem to be treated special.

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -125,10 +125,15 @@ public static void load(String originalName, ClassLoader loader) {
             }
         }
 
-        if (url == null) {
-            // Fall back to normal loading of JNI stuff
+        try {
+            // first try to load from java.library.path
             loadLibrary(loader, name, false);
+            logger.info("{} was loaded from java.libary.path", name);
```
carl-mastrangelo: This should be debug level, like in loadFirstAvailable.

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
eun-ice: the throwable is treated specially, see io.netty.util.internal.logging.MessageFormatter arrayFormat

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
eun-ice: I personally prefer the info level, since it would point to potential out-of-date libraries on disk, if someone messes up the java.library.path, but I will change it to logger.debug if you think debug is better.

normanmaurer:
Debug please

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
carl-mastrangelo: IMO log statements should be actionable.  In the case that it does load successfully from `java.library.path`, there isn't really much to do.  If it loads successfully but doesn't work right, that would warrant a log statement, which I think is what debug is good for.

eun-ice:
Debug it is!

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
eun-ice: Changed to debug, thanks for the feedback.

eun-ice:
@normanmaurer Tests on Linux (With and without grsecurity) and Mac OSX are absolutely fine. Is there anything else I can do?

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -125,10 +125,15 @@ public static void load(String originalName, ClassLoader loader) {
             }
         }
 
-        if (url == null) {
-            // Fall back to normal loading of JNI stuff
+        try {
+            // first try to load from java.library.path
             loadLibrary(loader, name, false);
```
Scottmitch: can you move this code above `libname` and other variables which may not be necessary?

eun-ice: Will do.

eun-ice: @Scottmitch PR is updated

Scottmitch:
@eun-ice - can you sign the Netty ICLA: https://netty.io/wiki/developer-guide.html

eun-ice:
@Scottmitch  just did. Also signed the CCLA since I don't know what applies here - as CTO lines between individual and corporate contribution get a little blurry sometimes

Scottmitch:
4.1 (da86b85a28) 4.0 (3c6ffe7b5c)

Scottmitch:
thanks @eun-ice !

eun-ice:
I have to thank you @Scottmitch . I love netty :-)

normanmaurer:
@eun-ice <3

normanmaurer:
@Scottmitch thanks for merging

Issue #7103, opened 2017-08-15
This DirectMemory should be released


MKudo:
I faced the huge direct memory usage in my application (using Netty). I research it and found this ReadOnlyByteBufferBuf copy method.

https://github.com/netty/netty/blob/eb7f751ba519cbcab47d640cd18757f09d077b55/buffer/src/main/java/io/netty/buffer/ReadOnlyByteBufferBuf.java#L419

It seams to create DirectByteBuffer and put it to UnpooledDirectByteBuf constructor. In this constructor, 'doNotFree field' will set to true. Then this DirectByteBuffer seams be not freed (by code. But GC will free this memory).

I hope Netty free this memory.

ninja-:
ByteBuffers are cleaned by JVM on garbage collection.
doNotFree is true, but since it's UnpooledDirectByteBuf they will both be garbage collected properly.
So my analysis here is that the code is correct.

You should use netty's ByteBufs whenever possible.

MKudo:
> You should use netty's ByteBufs whenever possible, and also use the pooled allocator and avoid using Unpooled*
Thank you your advice. Unfortunately I only using Cassandra driver (which using Netty). If I can find out fix point, I will send the request it for Cassandra driver developer.


But, UnpooledDirectByteBuf seams to try to free the memory directory. (I know if there are no Cleaner, it will do nothing and GC will free the memory.) And my hope is simply try this free.


In realloc case(capacity(int, int))
https://github.com/netty/netty/blob/66b17310418230d12c21f1f91287c22022d8d847/buffer/src/main/java/io/netty/buffer/UnpooledDirectByteBuf.java#L156
call setByteBuffer method. And it will call freeDirect method.
https://github.com/netty/netty/blob/66b17310418230d12c21f1f91287c22022d8d847/buffer/src/main/java/io/netty/buffer/UnpooledDirectByteBuf.java#L122


In dealloc case, deallocate method also call freeDirect method.
https://github.com/netty/netty/blob/66b17310418230d12c21f1f91287c22022d8d847/buffer/src/main/java/io/netty/buffer/UnpooledDirectByteBuf.java#L667

MKudo:
In my understand. The problem came from ReadOnlyByteBufferBuf's miss use of UnpooledDirectByteBuf constructor. The UnpooledDirectByteBuf's constructor is used by Unpooled#wrappedBuffer.
[...]

PR #7107, opened 2017-08-15
Use the ByteBufAllocator when copy a ReadOnlyByteBufferBuf and so als…

normanmaurer:
…o be able to release it without the GC when the Cleaner is present.

Motivation:

In ReadOnlyByteBufferBuf.copy(...) we just allocated a ByteBuffer directly and wrapped it. This way it was not possible for us to free the direct memory that was used by the copy without the GC.

Modifications:

- Ensure we use the allocator when create the copy and so be able to release direct memory in a timely manner
- Add unit test
- Depending on if the to be copied buffer is direct or heap based we also allocate the same type on copy.

Result:

Fixes [#7103].


normanmaurer:
Cherry-picked into 4.1 (e487db783673c9c153a011c8156083bbd86bf8e6) and 4.0 (1b357872a30c46873162b2d825669a01f2474e23)

PR #4132, opened 2015-08-24
HTTP to HTTP/2 translation errors (round 2)

Scottmitch:
Motivation:
Commit 0d8ce23c83ad7fa67245c721ba4d47437458ddfc failed to fix the Host header processing. Host is not a URI but is instead defined in https://tools.ietf.org/html/rfc3986#section-3.2.2 as host        = IP-literal / IPv4address / reg-name

Modifications:
- Host should not be treated as a URI.
- We should be more explicit about required fields, and unexpected input by throwing exceptions.

Result:
Translation from HTTP/1.x to HTTP/2 is more correct.

PR #7109, opened 2017-08-15
Use the index-based AsciiString constructor instead of substring() usage

fenik17:
Motivation:
The construction `new AsciiString(string.substring(...))` can be replaced with the `new AsciiString(string, start, length)` to avoid extra allocation.

Modifications:
Apply the described replacement in `HttpConversionUtil#setHttp2Authority`.

Result:
Less allocations.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/HttpConversionUtil.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/HttpConversionUtil.java
@@ -446,16 +446,19 @@ public final class HttpConversionUtil {
         return path.isEmpty() ? EMPTY_REQUEST_PATH : new AsciiString(path);
     }
 
-    private static void setHttp2Authority(String authority, Http2Headers out) {
+    // package-private for testing only
+    static void setHttp2Authority(String authority, Http2Headers out) {
         // The authority MUST NOT include the deprecated "userinfo" subcomponent
         if (authority != null) {
-            int endOfUserInfo = authority.indexOf('@');
-            if (endOfUserInfo < 0) {
-                out.authority(new AsciiString(authority));
-            } else if (endOfUserInfo + 1 < authority.length()) {
-                out.authority(new AsciiString(authority.substring(endOfUserInfo + 1)));
+            if (authority.isEmpty()) {
+                out.authority(EMPTY_STRING);
             } else {
-                throw new IllegalArgumentException("authority: " + authority);
+                int start = authority.indexOf('@') + 1;
+                int length = authority.length() - start;
+                if (length == 0) {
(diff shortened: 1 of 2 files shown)


normanmaurer:
@fenik17 can you include a unit test for the throws and the "normal" case ?

fenik17:
@normanmaurer done.

codec-http2/src/main/java/io/netty/handler/codec/http2/HttpConversionUtil.java
```diff
@@ -446,16 +446,18 @@ private static AsciiString toHttp2Path(URI uri) {
         return path.isEmpty() ? EMPTY_REQUEST_PATH : new AsciiString(path);
     }
 
-    private static void setHttp2Authority(String authority, Http2Headers out) {
+    static void setHttp2Authority(String authority, Http2Headers out) {
```
normanmaurer: @fenik17 just one nit... Can you add a comment such:

```
package-private for testing only.
```

fenik17: @normanmaurer +1. I'll add later today.

normanmaurer: @fenik17 thanks... let me know once you are done.

codec-http2/src/main/java/io/netty/handler/codec/http2/HttpConversionUtil.java
```diff
@@ -446,16 +446,18 @@ private static AsciiString toHttp2Path(URI uri) {
...
         if (authority != null) {
             int endOfUserInfo = authority.indexOf('@');
             if (endOfUserInfo < 0) {
                 out.authority(new AsciiString(authority));
-            } else if (endOfUserInfo + 1 < authority.length()) {
-                out.authority(new AsciiString(authority.substring(endOfUserInfo + 1)));
             } else {
-                throw new IllegalArgumentException("authority: " + authority);
+                int start = endOfUserInfo + 1;
+                if (start >= authority.length()) {
```
Scottmitch: the overflow condition should also be checked by the `AsciiString` constructor ... I guess this is just to provide a more descriptive exception message?

fenik17: > I guess this is just to provide a more descriptive exception message?

I suppose so. It's from #4132 )

fenik17:
@normanmaurer comment added. And the check for empty string. PTAL again, please.

normanmaurer:
Cherry-picked into 4.1 (ba2745665367a13931ab3999e171d51f50df030c). @fenik17 thanks

PR #7110, opened 2017-08-15
Revert SslEngineWrapperFactory api breakage introduced by 4448b8f42f59…

normanmaurer:
…9e79db1744cd8f5fdfee702c195e.

Motivation:

Commit 4448b8f42f599e79db1744cd8f5fdfee702c195e introduced some API breakage which we need to revert before we release.

Modifications:

- Introduce an AllocatorAwareSslEngineWrapperFactory which expose an extra method that takes a ByteBufAllocator as well.
- Revert API changes to SslEngineWrapperFactory.

Result:

API breakage reverted.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/JdkApplicationProtocolNegotiator.java
+++ b/handler/src/main/java/io/netty/handler/ssl/JdkApplicationProtocolNegotiator.java
@@ -32,6 +32,29 @@ public interface JdkApplicationProtocolNegotiator extends ApplicationProtocolNeg
          * Abstract factory pattern for wrapping an {@link SSLEngine} object. This is useful for NPN/APLN support.
          *
          * @param engine The engine to wrap.
+         * @param applicationNegotiator The application level protocol negotiator
+         * @param isServer <ul>
+         * <li>{@code true} if the engine is for server side of connections</li>
+         * <li>{@code false} if the engine is for client side of connections</li>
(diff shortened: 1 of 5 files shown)


handler/src/main/java/io/netty/handler/ssl/JdkApplicationProtocolNegotiator.java
```diff
@@ -32,6 +32,29 @@
...
          *
          * @param engine The engine to wrap.
+         * @param applicationNegotiator The application level protocol negotiator
+         * @param isServer <ul>
+         * <li>{@code true} if the engine is for server side of connections</li>
+         * <li>{@code false} if the engine is for client side of connections</li>
+         * </ul>
+         * @return The resulting wrapped engine. This may just be {@code engine}.
+         */
+        SSLEngine wrapSslEngine(
```
Scottmitch: should we deprecate the old interface/method and point to the new abstract class?

normanmaurer: We can but it's also exposed as the return type of a method on an interface. Let me just pull it in for now and we can do a followup if needed

normanmaurer:
Cherry-picked into 4.1 (5d9a5d3e8d81ec26be609bafb000833a7745a21b) and 4.0 (ff0d857845)

PR #7111, opened 2017-08-15
Add explicit fall through comments for errorprone.info

perezd:
Motivation:

When compiling this code and running it through errorprone[1], this message appears:
```
StringUtil.java:493: error: [FallThrough] Switch case may fall through; add a `// fall through` comment if it was deliberate
                    case LINE_FEED:
                    ^
    (see http://errorprone.info/bugpattern/FallThrough)
```
By adding that comment, it silences the error and also makes clear the intention of that statement.

[1]http://errorprone.info/index

Modification:

Add simple comment.

Result:

Errorprone is happier with the code.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/StringUtil.java
+++ b/common/src/main/java/io/netty/util/internal/StringUtil.java
@@ -491,6 +491,7 @@ public final class StringUtil {
                         }
                         // double-quote appears without being enclosed with double-quotes
                     case LINE_FEED:
+                        // fall through
                     case CARRIAGE_RETURN:
                         // special characters appears without being enclosed with double-quotes
                         throw newInvalidEscapedCsvFieldException(value, i);
--- a/common/src/main/java/io/netty/util/internal/StringUtil.java
+++ b/common/src/main/java/io/netty/util/internal/StringUtil.java
@@ -490,8 +490,8 @@ public final class StringUtil {
                             break;
                         }
                         // double-quote appears without being enclosed with double-quotes
+                    // fall through
                     case LINE_FEED:
-                        // fall through
                     case CARRIAGE_RETURN:
                         // special characters appears without being enclosed with double-quotes
                         throw newInvalidEscapedCsvFieldException(value, i);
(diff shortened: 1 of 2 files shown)


perezd:
Woops, I put that in the wrong place. Fixing.

common/src/main/java/io/netty/util/internal/StringUtil.java
```diff
@@ -490,6 +490,7 @@ public static CharSequence unescapeCsv(CharSequence value) {
                             break;
                         }
                         // double-quote appears without being enclosed with double-quotes
+                    // fall through
                     case LINE_FEED:
```
johnou: missing a second fall through comment for this case

common/src/main/java/io/netty/util/internal/StringUtil.java
```diff
@@ -490,6 +490,7 @@ public static CharSequence unescapeCsv(CharSequence value) {
                             break;
                         }
                         // double-quote appears without being enclosed with double-quotes
+                    // fall through
```
johnou: fix indenting?

transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
```diff
@@ -439,8 +439,8 @@ protected void run() {
                         if (wakenUp.get()) {
                             selector.wakeup();
                         }
+                    // fall through
```
johnou: indenting

common/src/main/java/io/netty/util/internal/StringUtil.java
perezd: Actually, it seems like you don't need it there if the case is fully empty. Just verified.

perezd:
OK sorry about the indenting, fixed!

normanmaurer:
squashed and cherry-picked into 4.1 (346900443234b2d39b07007b8878fe4f2905569c)

perezd:
Hey I Found more, sorry I should have warned. I'll open another request.

normanmaurer:
@perezd no worries... Just keep them coming :)

Issue #6755, opened 2017-05-18
SslHandlerTest#testCompositeBufSizeEstimationGuaranteesSynchronousWrite failure


Scottmitch:
Encountered while building PR https://github.com/netty/netty/pull/6493 on CI servers.

https://garage.netty.io/teamcity/viewLog.html?buildId=20267&buildTypeId=pulls_netty&tab=buildLog#_focus=40344

```
[19:57:50][io.netty:netty-handler] testCompositeBufSizeEstimationGuaranteesSynchronousWrite(io.netty.handler.ssl.SslHandlerTest)  Time elapsed: 122.061 sec  <<< ERROR!
[19:57:50][io.netty:netty-handler] java.util.concurrent.ExecutionException: java.lang.IllegalStateException: client closed
[19:57:50][io.netty:netty-handler] 	at io.netty.handler.ssl.SslHandlerTest.compositeBufSizeEstimationGuaranteesSynchronousWrite(SslHandlerTest.java:702)
[19:57:50][io.netty:netty-handler] 	at io.netty.handler.ssl.SslHandlerTest.testCompositeBufSizeEstimationGuaranteesSynchronousWrite(SslHandlerTest.java:585)
[19:57:50][io.netty:netty-handler] Caused by: java.lang.IllegalStateException: client closed
[19:57:50][io.netty:netty-handler] 
[19:57:50][io.netty:netty-handler] Running io.netty.handler.ssl.SniHandlerTest
[19:57:50][io.netty:netty-handler] 15:56:47.281 [threadDeathWatcher-5-1] DEBUG io.netty.buffer.PoolThreadCache - Freed 5 thread-local buffer(s) from thread: threadDeathWatcher-5-1
[19:57:51][io.netty:netty-handler] io.netty.handler.ssl.SslHandlerTest
[19:57:51][io.netty.handler.ssl.SslHandlerTest] testCompositeBufSizeEstimationGuaranteesSynchronousWrite
[19:57:51][testCompositeBufSizeEstimationGuaranteesSynchronousWrite] java.util.concurrent.ExecutionException: java.lang.IllegalStateException: client closed
[19:57:51]
[testCompositeBufSizeEstimationGuaranteesSynchronousWrite] java.util.concurrent.ExecutionException: java.lang.IllegalStateException: client closed
	at io.netty.handler.ssl.SslHandlerTest.compositeBufSizeEstimationGuaranteesSynchronousWrite(SslHandlerTest.java:702)
[...]

PR #7114, opened 2017-08-16
Increase visibility for SslHandlerTest#testCompositeBufSizeEstimation…

Scottmitch:
…GuaranteesSynchronousWrite

Motivation:
SslHandlerTest#testCompositeBufSizeEstimationGuaranteesSynchronousWrite has been observed to fail on CI servers, but it is not clear why.

Modifications:
- Add more visibility into what the state was and what the condition that caused the failure was.

Result:
More visibility when the test fails.


Scottmitch:
This is to help with https://github.com/netty/netty/issues/6755

Scottmitch:
4.1 (053e6184f2) 4.0 (ad552a5efd)

PR #7118, opened 2017-08-17
Update netty-tcnative native library names to use underscores.

normanmaurer:
Motivation:

We recently changed netty-tcnative to use underscores in its native library names.

Modifications:

Update code to use underscores when loading native library.

Result:

More consistent code.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
+++ b/handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
@@ -410,14 +410,12 @@ public final class OpenSsl {
         Set<String> libNames = new LinkedHashSet<String>(4);
         // First, try loading the platform-specific library. Platform-specific
         // libraries will be available if using a tcnative uber jar.
-        libNames.add("netty-tcnative-" + os + '-' + arch);
+        libNames.add("netty_tcnative-" + os + '-' + arch);
         if (LINUX.equalsIgnoreCase(os)) {
             // Fedora SSL lib so naming (libssl.so.10 vs libssl.so.1.0.0)..
-            libNames.add("netty-tcnative-" + os + '-' + arch + "-fedora");
+            libNames.add("netty_tcnative-" + os + '-' + arch + "-fedora");
         }
         // finally the default library.
-        libNames.add("netty-tcnative");
-        // in Java 8, statically compiled JNI code is namespaced
         libNames.add("netty_tcnative");
 
         NativeLibraryLoader.loadFirstAvailable(SSL.class.getClassLoader(),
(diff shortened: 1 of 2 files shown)


handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
```diff
@@ -410,14 +410,12 @@ private static void loadTcNative() throws Exception {
...
             // Fedora SSL lib so naming (libssl.so.10 vs libssl.so.1.0.0)..
-            libNames.add("netty-tcnative-" + os + '-' + arch + "-fedora");
+            libNames.add("netty_tcnative-" + os + '-' + arch + "-fedora");
         }
         // finally the default library.
-        libNames.add("netty-tcnative");
-        // in Java 8, statically compiled JNI code is namespaced
         libNames.add("netty_tcnative");
 
         NativeLibraryLoader.loadFirstAvailable(SSL.class.getClassLoader(),
```
Scottmitch: it seems like we may not actually use Library class from tnactive to load the native library ... if that is the case we also need to apply the shading prefix here too. And also there is an earlier check in OpenSsl.java to check if its already on the class path [1] which should also take into account the package prefix.

we really need to move tcnative into netty to simplify and reduce duplication 😭 

[1] `Class.forName("io.netty.internal.tcnative.SSL", false, OpenSsl.class.getClassLoader());`

normanmaurer: @Scottmitch actually I think this is not really related to this PR at all and should be tracked in another one. Or I miss-understood you ?

normanmaurer: Talked with @Scottmitch offline... this is not related to this pr and we will do a followup.

normanmaurer:
Cherry-picked into 4.1 (5de38051c973b5558bb30379a2f44db7d6f309c3) and 4.0 (a95bc4dcc4)

handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
Scottmitch: actually I think we can just not support the use case where netty-tcnative and netty have different package prefixes if we are assuming netty-tcnative will move into the netty project soon.

If we do want to support different package prefix (or for example one shaded and the other one not) then we will have to use the native library loading from `Library` in netty-tcnative (which would be able to infer the package prefix), as opposed to `NativeLibraryLoader` which lives in Netty and cannot as easily infer the package prefix for classes in another jar.

normanmaurer: @Scottmitch sounds good... This also means no other pr ;)

PR #7119, opened 2017-08-17
ByteToMessageDecoder must call fireChannelReadComplete if any channel…

normanmaurer:
…Read(...) produced data.

Motivation:

Commit d63bb4811ed8ccd5d9e45853f3ac6aee9da7ecab introduced a regression which could lead to missing calls of fireChannelReadComplete(). The problem could happen when we had multiple channelRead(...) calls and the last call of it not produced any data while the previous ones did. In this case we missed to call fireChannelReadComplete() as decodeWasNull was previous set to true while it should be false.

Modifications:

- Correctly only update decodeWasNull to true if no previous channelRead(...) call produced data.
- Add unit test.

Result:

Correctly call fireChannelReadComplete() in all cases.

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
@@ -136,7 +136,7 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
     ByteBuf cumulation;
     private Cumulator cumulator = MERGE_CUMULATOR;
     private boolean singleDecode;
-    private boolean decodeWasNull;
+    private boolean decodeWasNull = true;
     private boolean first;
     /**
      * A bitmask where the bits are defined as
@@ -280,7 +280,9 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
                 }
 
                 int size = out.size();
-                decodeWasNull = !out.insertSinceRecycled();
+                // Only update decodeWasNull if we did not produce data in a previous channelRead(...) call.
+                decodeWasNull &= !out.insertSinceRecycled();
+
                 fireChannelRead(ctx, out, size);
                 out.recycle();
             }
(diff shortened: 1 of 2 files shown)


codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -280,7 +280,11 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
                 }
 
                 int size = out.size();
-                decodeWasNull = !out.insertSinceRecycled();
+                if (decodeWasNull) {
+                    // Only update decodeWasNull if we did not produce data in a previous channelRead(...) call.
+                    decodeWasNull = !out.insertSinceRecycled();
```
Scottmitch: consider avoiding a conditional/branch statement here and just do the assignment.

```java
decodeWasNull = decodeWasNull && !out.insertSinceRecycled();
```

codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
```diff
@@ -344,4 +344,50 @@ public void channelReadComplete(ChannelHandlerContext ctx) throws Exception {
...
+                assertTrue(readCompleteExpected.get());
+                readCompleteCount.incrementAndGet();
+            }
+        });
+
+        channel.pipeline().fireChannelRead(Unpooled.wrappedBuffer(new byte[] {'a'}));
+        channel.pipeline().fireChannelReadComplete();
+        assertEquals(0, readCompleteCount.get());
+
+        channel.pipeline().fireChannelRead(Unpooled.wrappedBuffer(new byte[] {'b'}));
```
Scottmitch: add static import for `Unpooled.wrappedBuffer`?

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
normanmaurer: used bitwise at the end :)

normanmaurer:
@Scottmitch actually after more debugging I think what we tried to archive with https://github.com/netty/netty/commit/d63bb4811ed8ccd5d9e45853f3ac6aee9da7ecab is not safe to do... I would prefer to revert https://github.com/netty/netty/commit/d63bb4811ed8ccd5d9e45853f3ac6aee9da7ecab for now and think more about it in general. WDYT ?

normanmaurer:
Let me close this one for now... I will investigate

PR #7122, opened 2017-08-17
Do not define JNI_OnLoad when not dynamic

carl-mastrangelo:
Motivation:
Due to an oversight (by myself), linking two JNI modules with
duplicate symbols fails in linking.  This only seems to happen
some of the time (the behavior seems to be different between GCC
and Clang toolchains).  For instance, including both netty tcnative
and netty epoll fails to link because of duplicate JNI_OnLoad
symobols.

Modification:
Do not define the JNI_OnLoad and JNI_OnUnload symbols when
compiling for static linkage, as indicated by the NETTY_NOT_DYNAMIC
preprocessor define.  They are never directly called when
statically linked.

Result:
Able to statically compile epoll and tcnative code into a single
binary.

### DIFF
--- a/transport-native-epoll/src/main/c/netty_epoll_native.c
+++ b/transport-native-epoll/src/main/c/netty_epoll_native.c
@@ -595,7 +595,7 @@ jint JNI_OnLoad_netty_transport_native_epoll(JavaVM* vm, void* reserved) {
         return JNI_ERR;
     }
     char* packagePrefix = NULL;
-#ifndef NETTY_NOT_DYNAMIC
+#ifndef NETTY_BUILD_STATIC
     Dl_info dlinfo;
     jint status = 0;
     // We need to use an address of a function that is uniquely part of this library, so choose a static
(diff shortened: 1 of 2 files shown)


normanmaurer:
Ship it

carl-mastrangelo:
I'm not super familiar with your merge process, I'll defer to you on this one.

transport-native-epoll/src/main/c/netty_epoll_native.c
```diff
@@ -620,9 +620,11 @@ jint JNI_OnLoad_netty_transport_native_epoll(JavaVM* vm, void* reserved) {
     return ret;
 }
 
+#ifndef NETTY_NOT_DYNAMIC
```
Scottmitch: would it be possible to avoid double negatives? `if not defined netty not dynamic` -> `if defined netty dynamic` -> `#ifdef NETTY_DYNAMIC_LINK`?

carl-mastrangelo: i picked the negative form because if undefined, the default behavior (dyanmic) should happen.  

Perhaps `NETTY_BUILD_STATIC`?

Scottmitch: `NETTY_BUILD_STATIC` sgtm

carl-mastrangelo: Done, and updated commit message.

Scottmitch:
we will merge after CI completes.

normanmaurer:
Cherry-picked into 4.1 (e4af881bdbae26ebb9b5bfd3b0417be9b1a8115d)

@carl-mastrangelo thanks!

PR #7125, opened 2017-08-18
Use Constructor for reflective class instantiation.

carl-mastrangelo:
Motivation:
Calling `newInstance()` on a Class object can bypass compile time
checked Exception propagation.  This is noted in Java Puzzlers,
as well as in ErrorProne:
http://errorprone.info/bugpattern/ClassNewInstance

Modifications:
Use the niladic constructor to create a new instance.

Result:
Compile time safety for checked exceptions


carl-mastrangelo:
One additional addon for this would be in ReflectiveChannelFactory, keeping a ref to the ctor rather than the class.  This would throw an error early if there wasn't a publicly visible, no-arg ctor present on the channel.

normanmaurer:
@carl-mastrangelo thanks!

normanmaurer:
Cherry-picked into 4.1 (03d89c222279a8f1d9b70061a0e5e36c6fcf88fc) and 4.0 (ca8967bc0b6ecc5c4e8ea8aa1d0ae42235872795)

PR #7126, opened 2017-08-18
Support JDK9-native ALPN

### LINKED ISSUE #6933: Netty 4.1 should support ALPN on JDK9

normanmaurer:
Motivation:

Netty is unable to use Java9s ALPN support atm.

Modifications:

When running on Java9+ we invoke the correct methods that are exposed on the Java9+ implementation of SSLEngine and so be able to support ALPN.
This patch is based on the work of @rschmitt and so https://github.com/netty/netty/pull/6992.

Result:

Fixes #6933.


normanmaurer:
@rschmitt PTAL as well... This pr is based on your pr https://github.com/netty/netty/pull/6992 but not breaks the current API. 

This one was tested on latest java9 release and all ALPN related tests pass 🎉

vietj:
good job @normanmaurer @rschmitt

handler/src/main/java/io/netty/handler/ssl/Java9SslEngineWrapper.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.net.ssl.SSLException;
+
+import static io.netty.handler.ssl.SslUtils.toSSLHandshakeException;
+
+final class Java9SslEngineWrapper extends JdkSslEngine {
+    private final JdkApplicationProtocolNegotiator applicationNegotiator;
+
+    Java9SslEngineWrapper(SSLEngine engine, JdkApplicationProtocolNegotiator applicationNegotiator) {
+        super(engine);
+        this.applicationNegotiator =  applicationNegotiator;
```
johnou: Do we care about double spaces?

handler/src/main/java/io/netty/handler/ssl/Java9SslEngineWrapper.java
```diff
@@ -0,0 +1,43 @@
...
+        Java9SslUtils.configureAlpn(this, applicationNegotiator);
+    }
+
+    void selectProtocol() throws SSLException {
+        if (getUseClientMode()) {
```
johnou: And server mode? Throw unsupported operation exception?

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1428,7 +1430,12 @@ private boolean setHandshakeSuccessIfStillHandshaking() {
     /**
      * Notify all the handshake futures about the successfully handshake
      */
-    private void setHandshakeSuccess() {
+    private void setHandshakeSuccess() throws SSLException {
+        if (engine instanceof Java9SslEngineWrapper) {
+            // Call selectProtocol() to ensure our SelectionListener is called.
```
johnou: Does this need to be invoked every setHandshakeSuccess?

handler/src/main/java/io/netty/handler/ssl/Java9SslEngineWrapper.java
normanmaurer: Yes need to fix

handler/src/main/java/io/netty/handler/ssl/Java9SslEngineWrapper.java
normanmaurer: This way I not need any checks in SslHandler....

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
normanmaurer: I guess it would be better to only call it once... let me fix

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,186 @@
...
+
+        private AlpnSelector(JdkSslEngine jdkSslEngine, JdkApplicationProtocolNegotiator applicationNegotiator) {
+            this.jdkSslEngine = jdkSslEngine;
+            this.applicationNegotiator = applicationNegotiator;
+            protocols = Collections.unmodifiableSet(new LinkedHashSet<String>(applicationNegotiator.protocols()));
+        }
+
+        @SuppressWarnings("unchecked")
+        @Override
+        public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
```
netkins: MINOR Consider using varargs for methods or constructors which take an array the last parameter. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUseVarargs)

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,186 @@
...
+
+    private static void installSelector(final JdkSslEngine sslEngine,
+                                        final JdkApplicationProtocolNegotiator applicationProtocolNegotiator) {
+        checkNotNull(applicationProtocolNegotiator, "applicationProtocolNegotiator");
+
+        Object biFunction = AccessController.doPrivileged(new PrivilegedAction<Object>() {
+            @Override
+            public Object run() {
+                return Proxy.newProxyInstance(Java9SslUtils.class.getClassLoader(), new Class[]{ BI_FUNCTION_CLASS },
+                        new AlpnSelector(sslEngine, applicationProtocolNegotiator));
```
netkins: MAJOR Avoid instantiation through private constructors from outside of the constructors class. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AAccessorClassGeneration)

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
normanmaurer: actually this is good as it is

handler/src/main/java/io/netty/handler/ssl/Java9SslEngineWrapper.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.net.ssl.SSLException;
+
+import static io.netty.handler.ssl.SslUtils.toSSLHandshakeException;
+
+final class Java9SslEngineWrapper extends JdkSslEngine {
+    private final JdkApplicationProtocolNegotiator applicationNegotiator;
+
+    Java9SslEngineWrapper(SSLEngine engine, JdkApplicationProtocolNegotiator applicationNegotiator) {
+        super(engine);
+        this.applicationNegotiator = applicationNegotiator;
```
carl-mastrangelo: ObjectUtil.checkNotNull ?

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,186 @@
...
+            this.applicationNegotiator = applicationNegotiator;
+            protocols = Collections.unmodifiableSet(new LinkedHashSet<String>(applicationNegotiator.protocols()));
+        }
+
+        @SuppressWarnings("unchecked")
+        @Override
+        public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+            if (!"apply".equals(method.getName())) {
+                throw new UnsupportedOperationException("Invoked unknown method " + method.getName());
+            }
```
carl-mastrangelo: nit: check that the args are the correct count and type?

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,186 @@
...
+
+            JdkApplicationProtocolNegotiator.ProtocolSelector selector = applicationNegotiator.protocolSelectorFactory()
+                    .newSelector(jdkSslEngine, protocols);
+
+            try {
+                String selected = selector.select((List<String>) args[1]);
+                return selected == null ? StringUtil.EMPTY_STRING : selected;
+            } catch (Exception cause) {
```
carl-mastrangelo: should this be logged?

handler/src/main/java/io/netty/handler/ssl/JdkAlpnApplicationProtocolNegotiator.java
```diff
@@ -22,7 +22,8 @@
  * The {@link JdkApplicationProtocolNegotiator} to use if you need ALPN and are using {@link SslProvider#JDK}.
  */
 public final class JdkAlpnApplicationProtocolNegotiator extends JdkBaseApplicationProtocolNegotiator {
-    private static final boolean AVAILABLE = Conscrypt.isAvailable() || JettyAlpnSslEngine.isAvailable();
+    private static final boolean AVAILABLE = Conscrypt.isAvailable() || JettyAlpnSslEngine.isAvailable() ||
+            Java9SslUtils.supportsAlpn();
```
carl-mastrangelo: optional, but I would put jdk ahead of alpn.

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
normanmaurer: yep thats a good idea I think... at least with an assert...

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
normanmaurer: maybe... I am not sure.. @Scottmitch ?

handler/src/main/java/io/netty/handler/ssl/JdkAlpnApplicationProtocolNegotiator.java
normanmaurer: sure why not...

handler/src/main/java/io/netty/handler/ssl/Java9SslEngineWrapper.java
normanmaurer: not needed as we instance if only internally and we know it can't be null.

normanmaurer:
@carl-mastrangelo PTAL again.. added asserts FTW

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,187 @@
...
+            protocols = Collections.unmodifiableSet(new LinkedHashSet<String>(applicationNegotiator.protocols()));
+        }
+
+        @SuppressWarnings("unchecked")
+        @Override
+        public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+            // Assert that it is actually called via the BiConsumer that we expect.
+            assert "apply".equals(method.getName());
+            assert args.length == 2;
+            assert args[0] instanceof SSLEngine;
```
normanmaurer: as we cast arg[1] below I not added another assert here.

handler/src/main/java/io/netty/handler/ssl/Java9SslEngineWrapper.java
carl-mastrangelo: You might consider renaming selectProtocol to clientSelectProtocol, or selectProtocolIfClient or something like.

normanmaurer: `selectProtocolIfClient` sounds good.

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,187 @@
...
+            this.jdkSslEngine = jdkSslEngine;
+            this.applicationNegotiator = applicationNegotiator;
+            protocols = Collections.unmodifiableSet(new LinkedHashSet<String>(applicationNegotiator.protocols()));
+        }
+
+        @SuppressWarnings("unchecked")
+        @Override
+        public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+            // Assert that it is actually called via the BiConsumer that we expect.
+            assert "apply".equals(method.getName());
```
carl-mastrangelo: Include what the name was?  i.e. 

`assert "apply".equals(method.getName()); : "unexpected method " + method.getName();`

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,187 @@
...
+            assert args.length == 2;
```
carl-mastrangelo: same comment as above.

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
carl-mastrangelo: out of curiosity, what exception do you expect here that you want to ignore?

handler/src/main/java/io/netty/handler/ssl/JdkAlpnApplicationProtocolNegotiator.java
```diff
@@ -130,6 +132,9 @@ public SSLEngine wrapSslEngine(SSLEngine engine, ByteBufAllocator alloc,
                 return isServer ? JettyAlpnSslEngine.newServerEngine(engine, applicationNegotiator)
                         : JettyAlpnSslEngine.newClientEngine(engine, applicationNegotiator);
             }
+            if (Java9SslUtils.supportsAlpn()) {
```
carl-mastrangelo: in regard to the above, can you move this above the `if (JettyAlpnSslEngine.isAvailable()) {` ?  Jetty is a horrible hack and should be avoided if possible.

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
normanmaurer: We actually throw an exception if we want it to fail for example:

https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/JdkBaseApplicationProtocolNegotiator.java#L205

handler/src/main/java/io/netty/handler/ssl/JdkAlpnApplicationProtocolNegotiator.java
normanmaurer: +1

normanmaurer:
@Scottmitch ready to merge FTW ?

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,187 @@
...
+                });
+                getter.invoke(engine);
+
+                setter = AccessController.doPrivileged(new PrivilegedExceptionAction<Method>() {
+                    @Override
+                    public Method run() throws Exception {
+                        return SSLParameters.class.getMethod("setApplicationProtocols", String[].class);
+                    }
+                });
+                setter.invoke(engine.getSSLParameters(), new Object[] { EmptyArrays.EMPTY_STRINGS });
```
normanmaurer: the point was to try to use the method via reflection to ensure nothing will blow-up on runtime

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,187 @@
...
+    static {
+        Method setter = null;
+        Method getter = null;
+        Method selector = null;
+        Class<?> biFunction = null;
+
+        int version = PlatformDependent.javaVersion();
+        if (version >= 9) {
+            try {
+                SSLEngine engine = SSLContext.getDefault().createSSLEngine();
```
normanmaurer: @ejona86 the idea of this method was just to ensure all the reflective calls will work before we try to use these at runtime. I think if we change this to `SSLContext.getInstance("TLS")` this should be good enough, no ?

normanmaurer: @ejona86 actually I think I understand now what you are suggesting. Let me try this...

normanmaurer: @ejona86 after thinking a bt more about it and hacking on it I think we should just use `SSLContext.getInstance("TLS")` for now and be happy.  The problem is that we still have constructors that take `JdkApplicationProtocolNegotiator` and so its hard to to make this work without any hacks. That said I deprecated these in another PR:

https://github.com/netty/netty/pull/7138

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
ejona86: Oh, okay. FYI: setHandshakeApplicationProtocolSelector isn't tried here. That doesn't really bother me, though.

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
ejona86: Hmm... that is even more troublesome. Yeah, swapping to `SSLContext.getInstance("TLS")` is enough at this point.

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -573,12 +576,11 @@ public SSLEngine engine() {
...
      */
     public String applicationProtocol() {
-        SSLSession sess = engine().getSession();
-        if (!(sess instanceof ApplicationProtocolAccessor)) {
-            return null;
+        SSLSession sess = engine.getSession();
+        if (sess instanceof ApplicationProtocolAccessor) {
+            return ((ApplicationProtocolAccessor) sess).getApplicationProtocol();
         }
-
```
Scottmitch: should we try to use http://download.java.net/java/jdk9/docs/api/javax/net/ssl/SSLEngine.html#getApplicationProtocol-- here?

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -564,7 +564,10 @@ public final void setCloseNotifyReadTimeoutMillis(long closeNotifyReadTimeoutMil
      * Returns the {@link SSLEngine} which is used by this handler.
      */
     public SSLEngine engine() {
-        return engine;
+        // If its an instanceof Java9SslEngineWrapper we need to unwrap it before returning it to the user
+        // as Java9SslEngineWrapper does not delegate all the new methods that are added to SSLEngine in Java9.
+        return engine instanceof Java9SslEngineWrapper ?
```
Scottmitch: discussed with @normanmaurer offline. This may lead to treating the SslSession object differently in JDK9 and lower ... can we avoid this?

handler/src/main/java/io/netty/handler/ssl/Java9SslEngineWrapper.java
```diff
@@ -0,0 +1,43 @@
...
+        Java9SslUtils.configureAlpn(this, applicationNegotiator);
+    }
+
+    void selectProtocolIfClient() throws SSLException {
+        if (getUseClientMode()) {
+            try {
+                applicationNegotiator.protocolListenerFactory().newListener(
+                        this, applicationNegotiator.protocols())
+                        .selected(Java9SslUtils.getApplicationProtocol(getWrappedEngine()));
```
Scottmitch: If `Java9SslUtils.getApplicationProtocol(getWrappedEngine())` returns `null` (and maybe empty) (assuming the handshake is done) shouldn't we call [ProtocolSelectionListener#unsupported](https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/JdkApplicationProtocolNegotiator.java#L104)?

http://download.java.net/java/jdk9/docs/api/javax/net/ssl/SSLEngine.html#getApplicationProtocol--
> Returns:
null if it has not yet been determined if application protocols might be used for this connection, an empty String if application protocols values will not be used, or a non-empty application protocol String if a value was successfully negotiated.

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,190 @@
...
+            // Assert that it is actually called via the BiConsumer that we expect.
+            assert "apply".equals(method.getName()) : "unexpected method " + method.getName();
+            assert args.length == 2 : "unexpected lengths of arguments " +  args.length;
+            assert args[0] instanceof SSLEngine;
+
+            JdkApplicationProtocolNegotiator.ProtocolSelector selector = applicationNegotiator.protocolSelectorFactory()
+                    .newSelector(jdkSslEngine, protocols);
+
+            try {
+                String selected = selector.select((List<String>) args[1]);
```
Scottmitch: under what conditions can we invoke `ProtocolSelector#unsupported()`? Will this callback be called if the peer doesn't advertise ALPN, and if so what will the arguments be? Consider adding a comment.

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,190 @@
...
+        try {
+            SET_APPLICATION_PROTOCOL.invoke(parameters, new Object[]{ protocolArray });
+        } catch (Exception ex) {
+            throw new IllegalStateException(ex);
+        }
+    }
+
+    private static void installSelector(final JdkSslEngine sslEngine,
+                                        final JdkApplicationProtocolNegotiator applicationProtocolNegotiator) {
+        checkNotNull(applicationProtocolNegotiator, "applicationProtocolNegotiator");
```
normanmaurer: +1

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -87,7 +87,7 @@
  * the instance depends upon are released. Otherwise if any method of this class is called which uses the
  * the {@link ReferenceCountedOpenSslContext} JNI resources the JVM may crash.
  */
-public class ReferenceCountedOpenSslEngine extends SSLEngine implements ReferenceCounted {
+public class ReferenceCountedOpenSslEngine extends SSLEngine implements ReferenceCounted, ApplicationProtocolAccessor {
```
netkins: MAJOR The Cyclomatic Complexity of this class is 520 which is greater than 300 authorized. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AClassCyclomaticComplexity)

handler/src/main/java/io/netty/handler/ssl/Java9SslEngine.java
```diff
@@ -0,0 +1,110 @@
...
+        return verifyProtocolSelection(super.wrap(srcs, offset, len, dst));
+    }
+
+    @Override
+    public SSLEngineResult unwrap(ByteBuffer src, ByteBuffer dst) throws SSLException {
+        return verifyProtocolSelection(super.unwrap(src, dst));
+    }
+
+    @Override
+    public SSLEngineResult unwrap(ByteBuffer src, ByteBuffer[] dsts) throws SSLException {
```
netkins: MINOR Consider using varargs for methods or constructors which take an array the last parameter. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUseVarargs)

handler/src/main/java/io/netty/handler/ssl/Java9SslEngine.java
```diff
@@ -0,0 +1,110 @@
...
+
+import javax.net.ssl.SSLEngine;
+import javax.net.ssl.SSLEngineResult;
+import javax.net.ssl.SSLException;
+
+import java.nio.ByteBuffer;
+import java.util.List;
+import java.util.function.BiFunction;
+
+
```
Scottmitch: nit: kill extra line

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,229 @@
...
+import java.util.function.BiFunction;
+
+import io.netty.util.internal.EmptyArrays;
+import io.netty.util.internal.PlatformDependent;
+import io.netty.util.internal.StringUtil;
+import io.netty.util.internal.logging.InternalLogger;
+import io.netty.util.internal.logging.InternalLoggerFactory;
+
+final class Java9SslUtils {
+    private static final InternalLogger log = InternalLoggerFactory.getInstance(Java9SslUtils.class);
```
Scottmitch: nit: `log` -> `logger` for consistency?

handler/src/main/java/io/netty/handler/ssl/Java9SslEngine.java
```diff
@@ -0,0 +1,110 @@
...
+import static io.netty.handler.ssl.SslUtils.toSSLHandshakeException;
+
+final class Java9SslEngine extends JdkSslEngine {
+    private final JdkApplicationProtocolNegotiator applicationNegotiator;
+
+    Java9SslEngine(SSLEngine engine, JdkApplicationProtocolNegotiator applicationNegotiator) {
+        super(engine);
+        this.applicationNegotiator = applicationNegotiator;
+        Java9SslUtils.configureAlpn(this, applicationNegotiator);
```
Scottmitch: can we do this outside the scope of the constructor to avoid allowing `this` to escape? maybe in `wrapEngine` or create a static factory method in here?

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,229 @@
...
+        AlpnSelector(Java9SslEngine wrappedEngine, JdkApplicationProtocolNegotiator applicationNegotiator) {
+            this.wrappedEngine = wrappedEngine;
+            this.applicationNegotiator = applicationNegotiator;
+            protocols = Collections.unmodifiableSet(new LinkedHashSet<String>(applicationNegotiator.protocols()));
+        }
+
+        @Override
+        public String apply(SSLEngine sslEngine, List<String> strings) {
+
+            JdkApplicationProtocolNegotiator.ProtocolSelector selector = applicationNegotiator.protocolSelectorFactory()
```
Scottmitch: nit: consider adding a static import for `ProtocolSelector` to help with formatting.

handler/src/main/java/io/netty/handler/ssl/Java9SslEngine.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    private SSLEngineResult verifyProtocolSelection(SSLEngineResult result) throws SSLException {
+        if (result.getHandshakeStatus() == SSLEngineResult.HandshakeStatus.FINISHED && getUseClientMode()) {
+            String protocol = getApplicationProtocol();
+            JdkApplicationProtocolNegotiator.ProtocolSelectionListener selectionListener =
```
Scottmitch: nit: consider adding a static import for `ProtocolSelectionListener` to help with formatting.

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
```diff
@@ -0,0 +1,229 @@
...
+                    .newSelector(wrappedEngine, protocols);
+
+            try {
+                String selected = selector.select(strings);
+                return selected == null ? StringUtil.EMPTY_STRING : selected;
```
Scottmitch: https://github.com/netty/netty/pull/7126#discussion_r134666394

> under what conditions can we invoke `ProtocolSelector#unsupported()`? Will this BiFunction be called if the peer doesn't advertise ALPN, and if so what will the arguments be? Consider adding a comment.

handler/src/main/java/io/netty/handler/ssl/JdkAlpnApplicationProtocolNegotiator.java
```diff
@@ -16,13 +16,17 @@
...
+import io.netty.util.internal.PlatformDependent;
+
 import javax.net.ssl.SSLEngine;
 
 /**
  * The {@link JdkApplicationProtocolNegotiator} to use if you need ALPN and are using {@link SslProvider#JDK}.
  */
 public final class JdkAlpnApplicationProtocolNegotiator extends JdkBaseApplicationProtocolNegotiator {
-    private static final boolean AVAILABLE = Conscrypt.isAvailable() || JettyAlpnSslEngine.isAvailable();
+    private static final boolean AVAILABLE = Conscrypt.isAvailable() ||
```
Scottmitch: nit: consider reformatting:

```java
    private static final boolean AVAILABLE = Conscrypt.isAvailable() ||
                                             jdkAlpnSupported() ||
                                             JettyAlpnSslEngine.isAvailable();
```

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1827,15 +1828,19 @@ final boolean checkSniHostnameMatch(String hostname) {
         return Java8SslUtils.checkSniHostnameMatch(matchers, hostname);
     }
 
-    private final class OpenSslSession implements SSLSession, ApplicationProtocolAccessor {
+    @Override
+    public synchronized String getApplicationProtocol() {
```
Scottmitch: can `applicationProtocol` be volatile instead of requiring synchronization?

handler/src/main/java/io/netty/handler/ssl/Java9SslEngine.java
```diff
@@ -0,0 +1,110 @@
...
+    private final JdkApplicationProtocolNegotiator applicationNegotiator;
+
+    Java9SslEngine(SSLEngine engine, JdkApplicationProtocolNegotiator applicationNegotiator) {
+        super(engine);
+        this.applicationNegotiator = applicationNegotiator;
+        Java9SslUtils.configureAlpn(this, applicationNegotiator);
+    }
+
+    private SSLEngineResult verifyProtocolSelection(SSLEngineResult result) throws SSLException {
+        if (result.getHandshakeStatus() == SSLEngineResult.HandshakeStatus.FINISHED && getUseClientMode()) {
```
Scottmitch: it would be nice not to have to call this on every wrap/unwrap. It looks like this is not necessary on the server. Is it work having a `Java9ServerSslEngine` and `Java9ClientSslEngine` ... and the `Java9ServerSslEngine` wouldn't have todo this at all? `Java9ClientSslEngine ` wouldn't need to have `setHandshakeApplicationProtocolSelector` or `getHandshakeApplicationProtocolSelector` either.

handler/src/main/java/io/netty/handler/ssl/Java9SslEngine.java
normanmaurer: Good idea

handler/src/main/java/io/netty/handler/ssl/Java9SslEngine.java
normanmaurer: I can do this that said we still need the set and get as this must delegate to the wrapped engine to be able to delegate all methods of the java9 SSLEngine

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: Yep I think this should be fine

handler/src/main/java/io/netty/handler/ssl/Java9SslUtils.java
normanmaurer: Let me find out...

handler/src/main/java/io/netty/handler/ssl/Java9SslEngine.java
normanmaurer: after thinking more about this I think it is fine to just keep one implementation and just reverse the checks as the JIT should be smart enough to know it will never match when the engine is server side. if it really shows up to be an issue we can do a followup.

handler/src/main/java/io/netty/handler/ssl/JdkAlpnApplicationProtocolNegotiator.java
normanmaurer: done

handler/src/main/java/io/netty/handler/ssl/Java9SslEngine.java
```diff
@@ -0,0 +1,175 @@
...
+            if (protocol.isEmpty()) {
+                // ALPN is not supported
+                selector.unsupported();
+            }
+        }
+    }
+
+    Java9SslEngine(SSLEngine engine, JdkApplicationProtocolNegotiator applicationNegotiator, boolean isServer) {
+        super(engine);
+        if (isServer) {
```
netkins: MINOR Consider using varargs for methods or constructors which take an array the last parameter. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUseVarargs)

handler/src/main/java/io/netty/handler/ssl/Java9SslEngine.java
Scottmitch: > as the JIT should be smart enough to know it will never match when the engine is server side.

This would be an interesting optimization ... conditional code elimination based upon the usage of the class. Are you assuming the JIT would effectively? create the two different classes behind the scenes?

normanmaurer: basically I think the JIT should figure out that a boolean will always match or not match and if not just optimise the branch away ;)

That said after addressing your comment I need extra code for server and client side anyway.

handler/src/main/java/io/netty/handler/ssl/Java9SslEngine.java
```diff
@@ -0,0 +1,175 @@
...
+            selectionListener = applicationNegotiator.protocolListenerFactory()
+                    .newListener(this, applicationNegotiator.protocols());
+            alpnSelector = null;
+            Java9SslUtils.setApplicationProtocols(engine, applicationNegotiator.protocols());
+        }
+    }
+
+    private SSLEngineResult verifyProtocolSelection(SSLEngineResult result) throws SSLException {
+        if (result.getHandshakeStatus() == SSLEngineResult.HandshakeStatus.FINISHED) {
+            String protocol = getApplicationProtocol();
```
Scottmitch: this call can be moved into the conditional below

handler/src/test/java/io/netty/handler/ssl/JdkSslEngineTest.java
```diff
@@ -314,15 +333,15 @@ protected Provider serverSslContextProvider() {
...
                                                         String... supportedProtocols) {
         return new ApplicationProtocolConfig(protocol,
                 SelectorFailureBehavior.FATAL_ALERT,
                 SelectedListenerFailureBehavior.FATAL_ALERT,
                 supportedProtocols);
     }
 
-    private ApplicationProtocolConfig acceptingNegotiator(Protocol protocol,
+    private static ApplicationProtocolConfig acceptingNegotiator(Protocol protocol,
                                                           String... supportedProtocols) {
```
Scottmitch: nit: fix argument alignment

handler/src/test/java/io/netty/handler/ssl/JdkSslEngineTest.java
```diff
@@ -314,15 +333,15 @@ protected Provider serverSslContextProvider() {
         return provider;
     }
 
-    private ApplicationProtocolConfig failingNegotiator(Protocol protocol,
+    private static ApplicationProtocolConfig failingNegotiator(Protocol protocol,
                                                         String... supportedProtocols) {
```
Scottmitch: nit: fix argument alignment

handler/src/main/java/io/netty/handler/ssl/Java9SslEngine.java
```diff
@@ -0,0 +1,175 @@
...
+        return verifyProtocolSelection(super.wrap(srcs, offset, len, dst));
+    }
+
+    @Override
+    public SSLEngineResult unwrap(ByteBuffer src, ByteBuffer dst) throws SSLException {
+        return verifyProtocolSelection(super.unwrap(src, dst));
+    }
+
+    @Override
+    public SSLEngineResult unwrap(ByteBuffer src, ByteBuffer[] dsts) throws SSLException {
```
netkins: MINOR Consider using varargs for methods or constructors which take an array the last parameter. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUseVarargs)

normanmaurer:
Cherry-picked into 4.1 (1065e0f26e0d47a67c479b0fad81efab5d9438d9) and 4.0 (c78abce9548b8df2e4c49ec9e93906d27cea1ebe)

PR #7127, opened 2017-08-18
Include JNIEXPORT on exported symbols

carl-mastrangelo:
Motivation:
As noticed in https://stackoverflow.com/questions/45700277/
compilation can fail if the definition of a method doesn't
match the declaration.  It's easy enough to add this in, and make
it easy to compile.

Modifications:
Add JNIEXPORT to the entry points.

* On Windows this adds: `__declspec(dllexport)`
* On Mac this adds: `__attribute__((visibility("default")))`
* On Linux (GCC 4.2+) this adds: ` __attribute__((visibility("default")))`
* On other it doesn't add anything.

Result:
Easier compilation

### DIFF
--- a/transport-native-epoll/src/main/c/netty_epoll_native.c
+++ b/transport-native-epoll/src/main/c/netty_epoll_native.c
@@ -621,7 +621,7 @@ jint JNI_OnLoad_netty_transport_native_epoll(JavaVM* vm, void* reserved) {
 }
 
 #ifndef NETTY_BUILD_STATIC
-jint JNI_OnLoad(JavaVM* vm, void* reserved) {
+JNIEXPORT jint JNI_OnLoad(JavaVM* vm, void* reserved) {
     return JNI_OnLoad_netty_transport_native_epoll(vm, reserved);
 }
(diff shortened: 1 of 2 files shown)


carl-mastrangelo:
Analogue to https://github.com/netty/netty-tcnative/commit/1343d60faba61844379010bdabf218984ac8eb12

Scottmitch:
It would be nice to backport the EPOLL portion to 4.0 ... but looks like your related changes haven't been backported which may require some more work.

carl-mastrangelo:
@Scottmitch do you know approximate usage for each version?

Scottmitch:
@carl-mastrangelo - Nope ... we don't have great visibility into usage in general. Although I would say 4.1 is the current "development" branch and 4.0 is more or less in maintenance mode ... it generally makes maintenance easier to back port things when possible though (unless the feature doesn't exist in 4.0 of course).

carl-mastrangelo:
Merged as 7f1051b6ca9d1a146f4142844bfead43bce452c8

@Scottmitch I can send a backport PR if you'd like.  For now I'll leave this as is.

PR #7129, opened 2017-08-19
Decouple DnsCache and DnsCacheEntry

Scottmitch:
Motivation:
DnsCache (an interface) is coupled to DnsCacheEntry (a final class). This means that DnsCache implementations can't implement their own DnsCacheEntry objects if the default behavior isn't appropriate.

Modifications:
- DnsCacheEntry should be moved to DefaultDnsCache as it is an implementation detail
- DnsCache#cache(..) should return a new DnsCacheEntry
- The methods which from DnsCacheEntry that were used outside the scope of DefaultDnsCache should be moved into an interface

Result:
DnsCache is more extensible and not tightly coupled to a default implementation of DnsCacheEntry.

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
@@ -17,6 +17,7 @@ package io.netty.resolver.dns;
 
 import io.netty.channel.EventLoop;
 import io.netty.handler.codec.dns.DnsRecord;
+import io.netty.util.concurrent.ScheduledFuture;
 import io.netty.util.internal.PlatformDependent;
 import io.netty.util.internal.UnstableApi;
 
@@ -38,7 +39,8 @@ import static io.netty.util.internal.ObjectUtil.checkPositiveOrZero;
 @UnstableApi
 public class DefaultDnsCache implements DnsCache {
 
-    private final ConcurrentMap<String, List<DnsCacheEntry>> resolveCache = PlatformDependent.newConcurrentHashMap();
+    private final ConcurrentMap<String, List<DefaultDnsCacheEntry>> resolveCache =
+                                                            PlatformDependent.newConcurrentHashMap();
     private final int minTtl;
     private final int maxTtl;
     private final int negativeTtl;
@@ -95,8 +97,9 @@ public class DefaultDnsCache implements DnsCache {
 
     @Override
     public void clear() {
-        for (Iterator<Map.Entry<String, List<DnsCacheEntry>>> i = resolveCache.entrySet().iterator(); i.hasNext();) {
-            final Map.Entry<String, List<DnsCacheEntry>> e = i.next();
+        for (Iterator<Map.Entry<String, List<DefaultDnsCacheEntry>>> i = resolveCache.entrySet().iterator();
+             i.hasNext();) {
+            final Map.Entry<String, List<DefaultDnsCacheEntry>> e = i.next();
             i.remove();
             cancelExpiration(e.getValue());
         }
@@ -106,8 +109,9 @@ public class DefaultDnsCache implements DnsCache {
     public boolean clear(String hostname) {
         checkNotNull(hostname, "hostname");
         boolean removed = false;
-        for (Iterator<Map.Entry<String, List<DnsCacheEntry>>> i = resolveCache.entrySet().iterator(); i.hasNext();) {
-            final Map.Entry<String, List<DnsCacheEntry>> e = i.next();
+        for (Iterator<Map.Entry<String, List<DefaultDnsCacheEntry>>> i = resolveCache.entrySet().iterator();
+             i.hasNext();) {
+            final Map.Entry<String, List<DefaultDnsCacheEntry>> e = i.next();
             if (e.getKey().equals(hostname)) {
                 i.remove();
                 cancelExpiration(e.getValue());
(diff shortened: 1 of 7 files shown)


resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -122,19 +126,19 @@ private static boolean emptyAdditionals(DnsRecord[] additionals) {
     }
 
     @Override
-    public List<DnsCacheEntry> get(String hostname, DnsRecord[] additionals) {
+    public List<? extends DnsCacheEntry> get(String hostname, DnsRecord[] additionals) {
```
netkins: MINOR Consider using varargs for methods or constructors which take an array the last parameter. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUseVarargs)

resolver-dns/src/main/java/io/netty/resolver/dns/NoopDnsCache.java
```diff
@@ -47,13 +47,14 @@ public boolean clear(String hostname) {
     }
 
     @Override
-    public List<DnsCacheEntry> get(String hostname, DnsRecord[] additionals) {
+    public List<? extends DnsCacheEntry> get(String hostname, DnsRecord[] additionals) {
```
netkins: MINOR Consider using varargs for methods or constructors which take an array the last parameter. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUseVarargs)

resolver-dns/src/main/java/io/netty/resolver/dns/DnsCache.java
```diff
@@ -49,17 +49,19 @@
      * @param additionals the additional records
      * @return the cached entries
      */
-    List<DnsCacheEntry> get(String hostname, DnsRecord[] additionals);
+    List<? extends DnsCacheEntry> get(String hostname, DnsRecord[] additionals);
```
netkins: MAJOR Remove usage of generic wildcard type. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1452)
MINOR Consider using varargs for methods or constructors which take an array the last parameter. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUseVarargs)

resolver-dns/src/main/java/io/netty/resolver/dns/DnsCache.java
```diff
@@ -49,17 +49,19 @@
...
+     * Create a new {@link DnsCacheEntry} and cache a resolved address for a given hostname.
      * @param hostname the hostname
      * @param additionals the additional records
      * @param address the resolved address
      * @param originalTtl the TLL as returned by the DNS server
      * @param loop the {@link EventLoop} used to register the TTL timeout
+     * @return The {@link DnsCacheEntry} corresponding to this cache entry.
      */
-    void cache(String hostname, DnsRecord[] additionals, InetAddress address, long originalTtl, EventLoop loop);
+    DnsCacheEntry cache(String hostname, DnsRecord[] additionals, InetAddress address, long originalTtl,
```
normanmaurer: @Scottmitch if we want to do this we also should do the same for the other `cache` method.

normanmaurer: I must say I think how the interface works is a bit strange in the way that it act as cache but also "leak" the cache objects.

Scottmitch: I debated having a factory but the cache is the one that knows what type of object it will create. Having a factory would mean additional parameters in the resolver, and the cache/factory would have to be coordinated such that the type of object generated by the factory is compatible with the cache. Also if the cache wants to have some implementation details about the type of object this would be leaked by the factory and/or require casting.

Previously a cache object was created outside the cache before this method was called, and then again inside this method (which seems even more leaky). The `DnsCacheEntry` API only exposes "read only" accessors so I don't think we have to worry about external modifications. Someone could of course attempt to cast the object, but the `DefaultDnsCacheEntry` is package private so I don't think that is likely to be a problem.

Scottmitch: I debated doing this but it isn't currently necessary. The primary user of the cache (Netty) does not need the cache entry for the failure case. This allows the cache implementation to not even allocate an object if not necessary. We could also allow `null` to be returned but I figured it would be less complicated to just keep `void` until it is actually needed.

I'm not strong here and I can change if you prefer.

normanmaurer: @Scottmitch I am not saying it was any better before but just that I not like how the API works in general. Maybe its ok for now tho.

normanmaurer: I would prefer to change it to make it consistent

Scottmitch: done

Scottmitch: We can always submit a followup PR, but for now this PR mostly maintains the original API which some enhancements.

normanmaurer: Agree... ship it for now

Scottmitch:
@normanmaurer - PTAL

Scottmitch:
4.1 (566566db3a)

PR #7135, opened 2017-08-21
DelegatingSslContext should also be able to configure the SslHandler

normanmaurer:
Motivation:

DelegatingSslContext at the moment intercept newEngine calls and allow to init the SslEngine after it is created. The problem here is that this may not work the SSLEngine that is wrapped in the SslHandler when calling newHandler(...). This is because some SslContext implementations not delegate to newEngine(...) when creating the SslHandler to allow some optimizations. For this we should also allow to init the SslHandler after its creation and by default just delegate to initEngine(...).

Modifications:

Allow the user to also init the SslHandler after creation while by default init its SSLEngine after creation.

Result:

More flexible and correct code.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
@@ -379,12 +379,12 @@ public abstract class ReferenceCountedOpenSslContext extends SslContext implemen
     }
 
     @Override
-    final SslHandler newHandler(ByteBufAllocator alloc, boolean startTls) {
+    protected final SslHandler newHandler(ByteBufAllocator alloc, boolean startTls) {
         return new SslHandler(newEngine0(alloc, null, -1, false), startTls, false);
     }
 
     @Override
-    final SslHandler newHandler(ByteBufAllocator alloc, String peerHost, int peerPort, boolean startTls) {
+    protected final SslHandler newHandler(ByteBufAllocator alloc, String peerHost, int peerPort, boolean startTls) {
         return new SslHandler(newEngine0(alloc, peerHost, peerPort, false), startTls, false);
     }
 
--- a/handler/src/test/java/io/netty/handler/ssl/DelegatingSslContextTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/DelegatingSslContextTest.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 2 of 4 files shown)


handler/src/test/java/io/netty/handler/ssl/DelegatingSslContextTest.java
```diff
@@ -0,0 +1,60 @@
...
+        SslHandler handler = delegating.newHandler(UnpooledByteBufAllocator.DEFAULT);
+        Assert.assertArrayEquals(new String[] { SslUtils.PROTOCOL_TLS_V1_1 }, handler.engine().getEnabledProtocols());
+
+        handler = delegating.newHandler(UnpooledByteBufAllocator.DEFAULT, "localhost", 9090);
+        Assert.assertArrayEquals(new String[] { SslUtils.PROTOCOL_TLS_V1_1 }, handler.engine().getEnabledProtocols());
+    }
+
+    private static SslContext newDelegatingSslContext() throws Exception {
+        return new DelegatingSslContext(new JdkSslContext(SSLContext.getDefault(),
+                false, ClientAuth.NONE)) {
```
Scottmitch: nit: put these args on the previous line

handler/src/test/java/io/netty/handler/ssl/DelegatingSslContextTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Test
+    public void testInitEngineOnNewSslHandler() throws Exception {
+        SslContext delegating = newDelegatingSslContext();
+
+        SslHandler handler = delegating.newHandler(UnpooledByteBufAllocator.DEFAULT);
+        Assert.assertArrayEquals(new String[] { SslUtils.PROTOCOL_TLS_V1_1 }, handler.engine().getEnabledProtocols());
+
+        handler = delegating.newHandler(UnpooledByteBufAllocator.DEFAULT, "localhost", 9090);
+        Assert.assertArrayEquals(new String[] { SslUtils.PROTOCOL_TLS_V1_1 }, handler.engine().getEnabledProtocols());
```
Scottmitch: can we just use `EXPECTED_PROTOCOLS` here?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -379,12 +379,12 @@ public final SSLEngine newEngine(ByteBufAllocator alloc, String peerHost, int pe
     }
 
     @Override
-    final SslHandler newHandler(ByteBufAllocator alloc, boolean startTls) {
+    public final SslHandler newHandler(ByteBufAllocator alloc, boolean startTls) {
```
Scottmitch: why `public` instead of `protected`? The base class has this method as `protected` and it doesn't look like these methods need to be `public`.

handler/src/test/java/io/netty/handler/ssl/DelegatingSslContextTest.java
normanmaurer: yes

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
normanmaurer: yes protected is better... sorry 👍

normanmaurer:
@Scottmitch all addressed...

normanmaurer:
Cherry-picked into 4.1 (2beb5fc8ee) and 4. 0(cd80df7547)

PR #7136, opened 2017-08-21
SslHandlerTest#testCompositeBufSizeEstimationGuaranteesSynchronousWri…

Scottmitch:
…te print SslProvider on failure

Motivation:
When SslHandlerTest#testCompositeBufSizeEstimationGuaranteesSynchronousWrite fails it would be useful to know the SslProvider type

Modifications:
- Print the sever and client SslProvider upon failure

Result:
Failures include more info to help diagnose issues.


Scottmitch:
4.1 (cbce95eae1) 4.0 (9180b98e13)

PR #7138, opened 2017-08-22
Deprecate ApplicationProtocolNegotiator and its implementation as peo…

normanmaurer:
…ple should use ApplicationProtocolConfig

Motivation:

We should deprecate ApplicationProtocolNegotiator as the users should use ApplicationProtocolConfig these days.

Modifications:

Add deprecation annotations and javadocs.

Result:

Be able to make package-private in next major release.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ApplicationProtocolNegotiator.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ApplicationProtocolNegotiator.java
@@ -25,7 +25,10 @@ import java.util.List;
  * <li><a href="https://technotes.googlecode.com/git/nextprotoneg.html">Next Protocol Negotiation</a></li>
  * <li><a href="http://tools.ietf.org/html/rfc7301">Application-Layer Protocol Negotiation</a></li>
  * </ul>
+ *
+ * @deprecated use {@link ApplicationProtocolConfig}
  */
+@SuppressWarnings("deprecation")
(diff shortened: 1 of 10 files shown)


handler/src/main/java/io/netty/handler/ssl/ApplicationProtocolNegotiator.java
```diff
@@ -25,7 +25,10 @@
  * <li><a href="https://technotes.googlecode.com/git/nextprotoneg.html">Next Protocol Negotiation</a></li>
  * <li><a href="http://tools.ietf.org/html/rfc7301">Application-Layer Protocol Negotiation</a></li>
  * </ul>
+ *
+ * @deprecated
```
Scottmitch: add a comment about using `ApplicationProtocolConfig` instead?

Scottmitch: same comment in a few other places

normanmaurer:
@Scottmitch yes this should be package-private

normanmaurer:
@Scottmitch PTAL again

normanmaurer:
Cherry-picked into 4.1 (6e859469ca49f2029ad68f1d32e5b8bfa054eeba) and 4.0 (cb5227655a3a4682e8aa24ba69032fb525122dfe)

PR #7140, opened 2017-08-24
OpenSsl should use _ instead of -

Scottmitch:
Motivation:
netty-tcnative recently change the name of the native libraries from using - to _.

Modifications:
- OpenSsl should use _ instead of - even for the classifiers to be consistent with netty-tcnative

Result:
Loading netty-tcnative works.


Scottmitch:
4.1 (c93e58c453) 4.0 (6c7730328d)

normanmaurer:
LGTM

PR #7141, opened 2017-08-24
Include more detail why Unsafe is not available

carl-mastrangelo:
Motivation:
PD and PD0 Both try to find and use Unsafe.  If unavailable, they
try to log why and continue on.  However, it is not always easy to
enable this logging.  Chaining exceptions together is much easier
to reach, and the original exception is relevant when Unsafe is
needed.

Modifications:
* Make PD log why PD0 could not be loaded with a trace level log
* Make PD0 remember why Unsafe wasn't available
* Expose unavailability cause through PD for higher level use.
* Make Epoll and KQueue include the reason when failing

Result:
Easier debugging in hard to reconfigure environments

### DIFF
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent0.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent0.java
@@ -44,6 +44,7 @@ final class PlatformDependent0 {
     private static final int JAVA_VERSION = javaVersion0();
     private static final boolean IS_ANDROID = isAndroid0();
 
+    private static final Throwable UNSAFE_UNAVAILABILITY_CAUSE;
     private static final Object INTERNAL_UNSAFE;
     static final Unsafe UNSAFE;
 
@@ -64,12 +65,14 @@ final class PlatformDependent0 {
         final ByteBuffer direct;
         Field addressField = null;
         Method allocateArrayMethod = null;
+        Throwable unsafeUnavailabilityCause = null;
         Unsafe unsafe;
         Object internalUnsafe = null;
 
         if (isExplicitNoUnsafe()) {
             direct = null;
             addressField = null;
+            unsafeUnavailabilityCause = new UnsupportedOperationException("Unsafe explicitly disabled");
             unsafe = null;
             internalUnsafe = null;
         } else {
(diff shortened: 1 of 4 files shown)


carl-mastrangelo:
@normanmaurer still a little new on code review practice for Netty.  Google has separate LGTM and Approval; does Netty use the same model?

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -101,9 +104,10 @@ public Object run() {
...
             // the runtime permission accessClassInPackage.sun.misc
-            if (maybeUnsafe instanceof Exception) {
+            if (maybeUnsafe instanceof Throwable) {
                 unsafe = null;
-                logger.debug("sun.misc.Unsafe.theUnsafe: unavailable", (Exception) maybeUnsafe);
+                unsafeUnavailabilityCause = (Throwable) maybeUnsafe;
+                logger.debug("sun.misc.Unsafe.theUnsafe: unavailable", (Throwable) maybeUnsafe);
             } else {
                 unsafe = (Unsafe) maybeUnsafe;
                 logger.debug("sun.misc.Unsafe.theUnsafe: available");
```
Scottmitch: there are a few more cases below where it would be useful to set the `unsafeUnavailabilityCause` (where ever `unsafe` is set to `null`.

Scottmitch:
@carl-mastrangelo - there is no "hard and fast rule" for reviews/approvals but in general we try to get at least 1 other contributor's review, unless it is a "trivial" change. The more reviewers the better, and @normanmaurer and my self will often review after the fact if we are too busy and another contributor has already reviewed. The more complex the contribution/review the more justification there is to wait for more contributors ... but use your judgment, and ping us if we don't get to something within a few days.

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
normanmaurer: +1

normanmaurer:
@carl-mastrangelo can you address @Scottmitch s comments and update the PR ?

carl-mastrangelo:
@Scottmitch Thanks for the review.  I only say one other place where unsafe was set to null.  PTAL

johnou:
@carl-mastrangelo think there is one more 

```java
                   logger.debug("java.nio.Buffer.address: unavailable", (Throwable) maybeAddressField);

                    // If we cannot access the address of a direct buffer, there's no point of using unsafe.
                    // Let's just pretend unsafe is unavailable for overall simplicity.
                    unsafe = null;
```

carl-mastrangelo:
D'oh I'm blind.  Fixed.

There are a couple places that assume the return value won't be null (as in, if it isn't a Throwable, it must be nonnull).  I'm not sure these are exactly right, but I don't feel confident enough to put those assertions in.

normanmaurer:
@carl-mastrangelo thanks a lot!

Cherry-picked into 4.1 (c891c9c13fe9a04c714536e2c2be14ffc5b7b26e) and 4.0 (3e2236465f)

PR #7143, opened 2017-08-24
AbstractChannel should call doClose even after shutdownOutput

Scottmitch:
Motivation:
ShutdownOutput now fails all pending writes in the ChannelOutboundBuffer and sets it to null. However the Close code path uses the ChannelOutboundBuffer as an indication that the close operation is in progress and exits early and will not call doClose. This will lead to the Channel not actually being fully closed.

Bug introduced by 237a4da1b76ef7f303eca8f6ec8539efc9fe4d26

Modifications:
- AbstractChannel#close shouldn't exit early just because outboundBuffer is null, and instead should use additional state closeInitiated to avoid duplicate close operations

Result:
AbstractChannel#close(..) after AbstractChannel#shutdownOutbound() will still invoke doClose and cleanup Channel state.

### DIFF
--- a/transport/src/main/java/io/netty/channel/AbstractChannel.java
+++ b/transport/src/main/java/io/netty/channel/AbstractChannel.java
@@ -66,6 +66,7 @@ public abstract class AbstractChannel extends DefaultAttributeMap implements Cha
     private volatile SocketAddress remoteAddress;
     private volatile EventLoop eventLoop;
     private volatile boolean registered;
+    private boolean closeInitiated;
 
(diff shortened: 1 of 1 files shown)


normanmaurer:
LGTM... I will pull in once CI passes and also do another pr to add a unit test. @Scottmitch now get some sleep =P

normanmaurer:
Cherry-picked into 4.1 (c80fdc8241e15455632d10effa156b1ba99e25a8) and 4.0 (2c6a8b46767ba4338774241e89761072a5cafd86)

PR #7146, opened 2017-08-24
Makes LengthFieldBasedFrameDecoder::decode inlineable

franz1981:
Motivation:

The decode method is too large to be inlined with default compiler settings, hence the uncommon paths need to be packed and moved away form the common one.

Modifications:

The uncommon paths of the decode call (eg failures with thrown exceptions) are packed and moved in private methods in order to reduce the size of the common one
and let it being inlined.

Result:

The decode method is being inlined if the stack depth allows it.

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/LengthFieldBasedFrameDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/LengthFieldBasedFrameDecoder.java
@@ -15,12 +15,12 @@
  */
 package io.netty.handler.codec;
 
+import java.nio.ByteOrder;
+import java.util.List;
+
 import io.netty.buffer.ByteBuf;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.codec.serialization.ObjectDecoder;
-
-import java.nio.ByteOrder;
-import java.util.List;
 
 /**
  * A decoder that splits the received {@link ByteBuf}s dynamically by the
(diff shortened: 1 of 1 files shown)


franz1981:
I've a couple of flamegraphs with unfloded the inlined/not inlined the calls to it before and after the change

Before:
image

After:
image

normanmaurer:
@franz1981 can you please sigh our icla and let me know once done:

https://netty.io/s/icla

franz1981:
@normanmaurer Done, I've to wait something after it (email or similar)? 
Anyway Thanks!
Just as a note: looking at the graphs seems that ByteToMessageDecoder could need the same kind of optimization, I need to check with JitWatch too...

normanmaurer:
@franz1981 cool... no I will just merge once the CI completes. And yes we love contributions :)

franz1981:
@normanmaurer :100:  Glad to know! I work most of the time on the Artemis broker (that is using a lot of Netty), hence it is a pleasure to give back something good to this project :+1:

codec/src/main/java/io/netty/handler/codec/LengthFieldBasedFrameDecoder.java
```diff
@@ -346,6 +346,52 @@ protected final void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object>
...
+        } else {
+            // Enter the discard mode and discard everything received so far.
+            discardingTooLongFrame = true;
+            bytesToDiscard = discard;
+            in.skipBytes(in.readableBytes());
+        }
+        failIfNecessary(true);
+    }
+
+    private void failOnFrameLengthLessThanInitialBytesToStrip(ByteBuf in, long frameLength, int frameLengthInt) {
```
johnou: Can this be static?

codec/src/main/java/io/netty/handler/codec/LengthFieldBasedFrameDecoder.java
```diff
@@ -346,6 +346,52 @@ protected final void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object>
...
+        failIfNecessary(false);
+    }
+
+    private void failOnNegativeLengthField(ByteBuf in, long frameLength) {
+        in.skipBytes(lengthFieldEndOffset);
+        throw new CorruptedFrameException(
+           "negative pre-adjustment length field: " + frameLength);
+    }
+
+    private void failOnFrameLengthLessThanLengthFieldEndOffset(ByteBuf in, long frameLength) {
```
johnou: Static?

codec/src/main/java/io/netty/handler/codec/LengthFieldBasedFrameDecoder.java
franz1981: I can make static with ease:

- failOnNegativeLengthField
- failOnFrameLengthLessThanLengthFieldEndOffset
- failOnFrameLengthLessThanInitialBytesToStrip
But they are on the "slow paths" and throwing exceptions, so I do not see great benefits doing it: there are other reasons (besides performances) that I'm not seeing to perform such change?

johnou: I wouldn't go out of the way to transform all methods to static, I was under the impression that it was a common coding style to use static helper / utility methods which are stateless.

franz1981: Got it...All the methods I've moved away from the decode are not stateless in their original form (copied from the original decode method), but can be transformed in that way, as I pointed in the last comment.

normanmaurer:
Correct... if it can be static please make it static 👍

johnou:
@franz1981 could you please provide the graph legend too? Bright green / lime == not eligible for inlineable?

franz1981:
@normanmaurer @johnou I can change the method I've listed making them static and squash the whole here
@johnou Yes, sorry, I've missed that..
I've followed (partially) this guide: http://psy-lob-saw.blogspot.it/2017/02/flamegraphs-intro-fire-for-everyone.html and quoting it: 

> Aqua frames are inlined frames and green are "real".
Red are native code (kernel code or anything called via JNI/JNA) while yellow are mostly GC (JVM) native stuff.

normanmaurer:
@franz1981 yep please do 👍

franz1981:
@normanmaurer @johnou Done, at a first look it doesn't look like the byte code size seems not increased (besides this.* used to prepare the operand of the moved fail* calls)

johnou:
@franz1981 looking forward to the next pull request for ByteToMessageDecoder ;)

normanmaurer:
Cherry-picked into 4.1 (6780183a89066a4de0449d3be57caa8dcf1be9ce) and 4.0 (ff22b6573dcf0c28110ebc8d539690038df507fb).

@franz1981 thanks a lot!

PR #7148, opened 2017-08-25
AutoClose behavior may infinite loop

Scottmitch:
Motivation:
If AutoClose is false and there is a IoException then AbstractChannel will not close the channel but instead just fail flushed element in the ChannelOutboundBuffer. AbstractChannel also notifies of writability changes, which may lead to an infinite loop if the peer has closed its read side of the socket because we will keep accepting more data but continuously fail because the peer isn't accepting writes.

Modifications:
- If the transport throws on a write we should acknowledge that the output side of the channel has been shutdown and cleanup. If the channel can't accept more data because it is full, and still healthy it is not expected to throw. However if the channel is not healthy it will throw and is not expected to accept any more writes. In this case we should shutdown the output for Channels that support this feature and otherwise just close.
- Connection-less protocols like UDP can remain the same because the channel may disconnected temporarily.
- Make sure AbstractUnsafe#shutdownOutput is called because the shutdown on the socket may throw an exception.

Result:
More correct handling of write failure when AutoClose is false.

### DIFF
--- a/testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketHalfClosedTest.java
+++ b/testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketHalfClosedTest.java
@@ -28,15 +28,19 @@ import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.channel.ChannelInitializer;
 import io.netty.channel.ChannelOption;
 import io.netty.channel.RecvByteBufAllocator;
+import io.netty.channel.SimpleChannelInboundHandler;
 import io.netty.channel.socket.ChannelInputShutdownEvent;
 import io.netty.channel.socket.ChannelInputShutdownReadComplete;
+import io.netty.channel.socket.ChannelOutputShutdownEvent;
 import io.netty.channel.socket.DuplexChannel;
 import io.netty.util.UncheckedBooleanSupplier;
 import org.junit.Test;
 
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
 public class SocketHalfClosedTest extends AbstractSocketTest {
@@ -50,8 +54,8 @@ public class SocketHalfClosedTest extends AbstractSocketTest {
         testAllDataReadAfterHalfClosure(false, sb, cb);
     }
 
-    public void testAllDataReadAfterHalfClosure(final boolean autoRead,
-                                                ServerBootstrap sb, Bootstrap cb) throws Throwable {
+    private void testAllDataReadAfterHalfClosure(final boolean autoRead,
+                                                 ServerBootstrap sb, Bootstrap cb) throws Throwable {
         final int totalServerBytesWritten = 1024 * 16;
         final int numReadsPerReadLoop = 2;
         final CountDownLatch serverInitializedLatch = new CountDownLatch(1);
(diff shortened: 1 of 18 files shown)


transport/src/main/java/io/netty/channel/AbstractServerChannel.java
```diff
@@ -58,6 +60,13 @@ protected void doDisconnect() throws Exception {
         throw new UnsupportedOperationException();
     }
 
+    @UnstableApi
+    @Override
+    protected final void doShutdownOutput(Throwable cause) throws Exception {
+        super.doShutdownOutput(cause);
```
johnou: Should this be wrapped in try finally since do shutdown throws exception?

Scottmitch: I don't think this is necessary because if this method throws AbstractChannel will catch and close the connection.

normanmaurer: @Scottmitch ah ok... sounds good then

Scottmitch:
@normanmaurer - unit test added ... will merge when CI completes

testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketHalfClosedTest.java
```diff
@@ -50,8 +54,8 @@ public void testAllDataReadAfterHalfClosure(ServerBootstrap sb, Bootstrap cb) th
         testAllDataReadAfterHalfClosure(false, sb, cb);
     }
 
-    public void testAllDataReadAfterHalfClosure(final boolean autoRead,
-                                                ServerBootstrap sb, Bootstrap cb) throws Throwable {
+    private void testAllDataReadAfterHalfClosure(final boolean autoRead,
```
netkins: MINOR Make "testAllDataReadAfterHalfClosure" a "static" method. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2325)

testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketHalfClosedTest.java
```diff
@@ -151,6 +155,200 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
...
+    }
+
+    public void testAutoCloseFalseDoesShutdownOutput(ServerBootstrap sb, Bootstrap cb) throws Throwable {
+        testAutoCloseFalseDoesShutdownOutput(false, false, sb, cb);
+        testAutoCloseFalseDoesShutdownOutput(false, true, sb, cb);
+        testAutoCloseFalseDoesShutdownOutput(true, false, sb, cb);
+        testAutoCloseFalseDoesShutdownOutput(true, true, sb, cb);
+    }
+
+    private void testAutoCloseFalseDoesShutdownOutput(boolean allowHalfClosed,
```
netkins: MINOR Make "testAutoCloseFalseDoesShutdownOutput" a "static" method. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2325)

testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketHalfClosedTest.java
```diff
@@ -162,8 +360,8 @@ public void testAllDataReadClosure(ServerBootstrap sb, Bootstrap cb) throws Thro
         testAllDataReadClosure(false, true, sb, cb);
     }
 
-    public void testAllDataReadClosure(final boolean autoRead, final boolean allowHalfClosed,
-                                       ServerBootstrap sb, Bootstrap cb) throws Throwable {
+    private void testAllDataReadClosure(final boolean autoRead, final boolean allowHalfClosed,
```
netkins: MINOR Make "testAllDataReadClosure" a "static" method. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2325)

testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketHalfClosedTest.java
Scottmitch: fixed

testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketHalfClosedTest.java
Scottmitch: fixed

testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketHalfClosedTest.java
Scottmitch: fixed

Scottmitch:
4.1 (89ecb4b4a4)

Scottmitch:
Backport to 4.0 is in https://github.com/netty/netty/pull/7153

Issue #7150, opened 2017-08-25
Netty Epoll.isAvailable() check doesn't handle lib loading issue on 32bit JVM


michaelandrepearce:
### Expected behavior
Method should check os and jvm variables if it can safely load the libs expected and return false and not cause JVM warning by loading incompatible libs

https://issues.apache.org/jira/browse/ARTEMIS-1099

### Actual behavior
using a 32 bit JVM:
Java HotSpot(TM) Server VM warning: You have loaded library /tmp/libnetty-transport-native-epoll7550636318037030956.so which might have disabled stack guard. The VM will try to fix the stack guard now.
It's highly recommended that you fix the library with 'execstack -c <libfile>', or link it with '-z noexecstack'.

### Steps to reproduce
Run netty native epoll on 32bit jvm 

### Minimal yet complete reproducer code (or URL to code)
see:
https://issues.apache.org/jira/browse/ARTEMIS-1099

### Netty version
4.1.x - present

### JVM version (e.g. `java -version`)
32bit any
### OS version (e.g. `uname -a`)
linux

michaelandrepearce:
This seems similar to (possible duplicate ) https://github.com/netty/netty/issues/7120

johnou:
@michaelandrepearce are you packaging your own epoll artifact? There is currently no 32 bit support for epoll. Iirc detecting actual architecture from jvm is not straightforward.

michaelandrepearce:
@johnou No we're not, the isAvailable() check needs to take this into account though in its checks to detect if it is available, before trying to load the native libs.

johnou:
@michaelandrepearce but I am running a custom packaged epoll artifact which runs on 32 bit architecture so the suggested changes will break our application :(

Scottmitch:
@michaelandrepearce - can you describe the implication of the "Actual Behavior"? I would like to understand what the impacts of this issue are. For example is one of the following happening:
- Is it that the library is loaded and later crashes or does something unexpected?
- Does the load fail, and the error message is printed out, and you just don't want to see the error message?

michaelandrepearce:
@scottmitch 
It loads the lib but due to the bit complied on invocation (used by is available) it causes a native stack error and then the jvm has to unload i gracefully. It does continue, but this reliance that the jvm can catch this safely unload really should be avoided.

michaelandrepearce:
[...]

PR #7151, opened 2017-08-25
ISSUE-7150 Add checks to load libs only in support environments.

michaelandrepearce:
To avoid attempts of native lib loading, only load the libs in environments currently  the native libs are build/provided for.

Motivation:

Netty should only attempt to load native libs, for environments that the native libs are provided, this avoids un-needed jvm warnings on isAvailable checks.

Modification:

Add checks for epoll, to only load in 64bit linux jvm
Add checks for kqueue, to only load in 64bit osx jvm

Result:

Fixes #https://github.com/netty/netty/issues/7150

### DIFF
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent.java
@@ -73,6 +73,7 @@ public final class PlatformDependent {
 
     private static final boolean IS_WINDOWS = isWindows0();
     private static final boolean IS_OSX = isOsx0();
+    private static final boolean IS_LINUX = isLinux0();
 
     private static final boolean MAYBE_SUPER_USER;
 
@@ -214,6 +215,13 @@ public final class PlatformDependent {
      */
     public static boolean isOsx() {
         return IS_OSX;
+    }
+
+    /**
+     * Return {@code true} if the JVM is running on Linux
+     */
+    public static boolean isLinux() {
+        return IS_LINUX;
     }
 
     /**
@@ -941,6 +949,17 @@ public final class PlatformDependent {
             logger.debug("Platform: MacOS");
         }
         return osx;
+    }
+
+    private static boolean isLinux0() {
+        String osname = SystemPropertyUtil.get("os.name", "").toLowerCase(Locale.US)
+                                          .replaceAll("[^a-z0-9]+", "");
+        boolean linux = osname.startsWith("linux");
+
+        if (linux) {
+            logger.debug("Platform: Linux");
+        }
+        return linux;
     }
 
     private static boolean maybeSuperUser0() {
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/Epoll.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/Epoll.java
@@ -26,36 +26,40 @@ public final class Epoll {
     private static final Throwable UNAVAILABILITY_CAUSE;
 
     static  {
-        Throwable cause = null;
-        FileDescriptor epollFd = null;
-        FileDescriptor eventFd = null;
-        try {
-            epollFd = Native.newEpollCreate();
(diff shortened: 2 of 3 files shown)


michaelandrepearce:
@johnou looks like pr build isn't working post recent tag. is some post tagging wash-up normally needed?

ninja-:
@michaelandrepearce AFAIR you need approval from one of the project admins to get your branch to run on CI...so wait for it  :D

transport-native-epoll/src/main/java/io/netty/channel/epoll/Epoll.java
```diff
@@ -26,36 +26,40 @@
...
-            eventFd = Native.newEventFd();
-        } catch (Throwable t) {
-            cause = t;
-        } finally {
-            if (epollFd != null) {
-                try {
-                    epollFd.close();
-                } catch (Exception ignore) {
-                    // ignore
+        if (PlatformDependent.isLinux() && PlatformDependent.bitMode() == 64) {
```
johnou: This will break custom packaged epoll artifacts.

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -943,6 +951,17 @@ private static boolean isOsx0() {
         return osx;
     }
 
+    private static boolean isLinux0() {
+        String osname = SystemPropertyUtil.get("os.name", "").toLowerCase(Locale.US)
+                                          .replaceAll("[^a-z0-9]+", "");
+        boolean linux = osname.startsWith("linux");
```
johnou: iirc this is not enough to detect most Linux distributions.

johnou:
```java
boolean linux = (osname.contains("nix") || osname.contains("nux")
                    || osname.contains("aix"));
```

michaelandrepearce: This would be incorrect.

 Aix != Linux, like wise picking up nix is the ending for unix again != Linux.

Epoll is Linux 2.6+ only

transport-native-epoll/src/main/java/io/netty/channel/epoll/Epoll.java
michaelandrepearce: The code supports what the project delivers in it's compiled binaries.

If the project where to deliver 32 but and 64 bit binaries then this would be changed but then also foldering the builds also would be needed eg like how other projects handle native libs, example jnr with jffi

johnou: I disagree, the same could be said about Netty openssl but the documentation gives instructions on how to build 32 bit artifacts (we actually build an uber jar which works on both 32bit and 64bit architectures). 

http://netty.io/wiki/forked-tomcat-native.html

michaelandrepearce: So how is that working to detect the right bit architecture and load the correct compiled lib within the uber? Is it like as I mentioned the jnr approach for jffi, or are you just compiling to 32bit architecture

johnou: The openssl uber jar uses the architecture as part of the filename in the jar.

michaelandrepearce: Cool so it's more akin to what I'm suggesting on the issue comment trace.

Are you able to comment on the issue discussions to answer my one query there?

normanmaurer:
Stand back... I am working on a fix

normanmaurer:
Let me close this and please check https://github.com/netty/netty/pull/7163.

michaelandrepearce:
@normanmaurer sure, the approach in your PR is what i was getting to on the issue discussion as an alternative option (better).

PR #7158, opened 2017-08-28
Fix ServerCookieEncoder javadoc, close #7115

### LINKED ISSUE #7115: Multiple Set-Cookie HTTP headers should not be folded into one
According to [RFC 6265](https://tools.ietf.org/html/rfc6265#section-3):

> Origin servers SHOULD NOT fold multiple Set-Cookie header fields into
   a single header field.  The usual mechanism for folding HTTP headers
   fields (i.e., as defined in [RFC2616]) might change the semantics of
   the Set-Cookie header field because the %x2C (",") character is used
   by Set-Cookie in a way that conflicts with such folding.

However `ServerCookieEncoder` [Javadoc](http://netty.io/4.1/api/io/netty/handler/codec/http/cookie/ServerCookieEncoder.html) states: 

> A RFC6265 compliant cookie encoder to be used server side, so some fields are sent (Version is typically ignored). As Netty's Cookie merges Expires and MaxAge into one single field, only Max-Age field is sent. Note that multiple cookies are supposed to be sent at once in a single "Set-Cookie" header.
 // Example
 HttpRequest req = ...;
 res.setHeader("Cookie", ServerCookieEncoder.encode("JSESSIONID", "1234"));

This is contradicting to the RFC and misleading.

Also the example is outdated with the `setHeader` method, has `req` in one line and `res` in another line, and should use "Set-Cookie" instead of "Cookie".

slandelle:
Motivation:

ServerCookieEncoder’s javadoc contains some invalid copy-pasting from
ClientCookieEncoder.

Modifications:
* As per RFC6265, multiple cookies are sent as separate Set-Cookie
response headers.
* Fix code sample

Result:

Proper javadoc

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/ServerCookieEncoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/ServerCookieEncoder.java
@@ -26,12 +26,12 @@ import java.util.List;
  *
  * As Netty's Cookie merges Expires and MaxAge into one single field, only Max-Age field is sent.
  *
- * Note that multiple cookies are supposed to be sent at once in a single "Set-Cookie" header.
+ * Note that multiple cookies must be sent as separate "Set-Cookie" headers.
  *
  * <pre>
  * // Example
- * {@link HttpRequest} req = ...;
- * res.setHeader("Cookie", {@link ServerCookieEncoder}.encode("JSESSIONID", "1234"));
+ * {@link HttpResponse} res = ...;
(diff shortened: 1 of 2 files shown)


normanmaurer:
@slandelle is this only for 4.1 ?

slandelle:
It's not only for 4.1. You should be able to cherry-pick on 4.0.

normanmaurer:
@slandelle thanks

Cherry-picked into 4.1 (d9d0e633dc69b9b93139d822eaa17b97ff3826ab) and 4.0 (5ccedc5e5b)

PR #7161, opened 2017-08-29
HttpObjectEncoder buffer size estimation

Scottmitch:
Motivation:
HttpObjectEncoder allocates a new buffer when encoding the initial line and headers, and also allocates a buffer when encoding the trailers. The allocation always uses the default size of 256. This may lead to consistent under allocation and require a few resize/copy operations which can cause GC/memory pressure.

Modifications:
- Introduce a weighted average which tracks the historical size of encoded data and uses this as an estimate for future buffer allocations

Result:
Better approximation of buffer sizes.


normanmaurer:
@Scottmitch ship it... nice change

Scottmitch:
4.1 (9bd6d8129e)

normanmaurer:
I also cherry-picked into 4.0 (4732fabb1664acae0a545b6496066acb8d39ecb7). Was easy to do and makes maintenance easier.

Issue #7159, opened 2017-08-28
AssertionError error thrown during Epoll channel cleanup


gemmellr:
We received a bug report [1] that our client which uses Netty 4.1.14, was failing some user tests due to throwing an assertion error:

> java.lang.AssertionError
        at io.netty.channel.epoll.EpollEventLoop.remove(EpollEventLoop.java:159)
        at io.netty.channel.epoll.AbstractEpollChannel.doDeregister(AbstractEpollChannel.java:142)
        at io.netty.channel.epoll.AbstractEpollChannel.doClose(AbstractEpollChannel.java:118)
        at io.netty.channel.epoll.AbstractEpollStreamChannel.doClose(AbstractEpollStreamChannel.java:703)
        at io.netty.channel.AbstractChannel$AbstractUnsafe.doClose0(AbstractChannel.java:685)
        at io.netty.channel.AbstractChannel$AbstractUnsafe.access$700(AbstractChannel.java:419)
        at io.netty.channel.AbstractChannel$AbstractUnsafe$5.run(AbstractChannel.java:646)
        at io.netty.util.concurrent.GlobalEventExecutor$TaskRunner.run(GlobalEventExecutor.java:233)
        at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:138)
        at java.lang.Thread.run(Thread.java:748)

Looking into the stracktrace, it seems that it is complaining the EpollEventLoop.remove() operation is not being run on the event loop thread. From the stack, it seems to be running on the GlobalEventExecutor.

Following the trace back to see how that might happen, I supposed it could be from that executor being returned here: https://github.com/netty/netty/blob/netty-4.1.14.Final/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollSocketChannel.java#L210 and used to schedule the task. If so, I also noticed the line before the executor was returned has actually already run the method throwing the assertion later when run on the executor. From the code it seemed like this would only happen with so_linger set, and indeed looking into the test further showed it does appear to be using options to set that. 

[1] https://issues.apache.org/jira/browse/QPIDJMS-314

normanmaurer:
@gemmellr let me check

normanmaurer:
@gemmellr fixed by https://github.com/netty/netty/pull/7162. Please check

gemmellr:
[...]

PR #7162, opened 2017-08-29
Fix assertion error when closing / shutdown native channel and SO_LINGER is set.

normanmaurer:
Motivation:

When SO_LINGER is used we run doClose() on the GlobalEventExecutor by default so we need to ensure we schedule all code that needs to be run on the EventLoop on the EventLoop in doClose. Beside this there are also threading issues when calling shutdownOutput(...)

Modifications:

- Schedule removal from EventLoop to the EventLoop
- Correctly handle shutdownOutput and shutdown in respect with threading-model
- Add unit tests

Result:

Fixes [#7159].

### DIFF
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollChannel.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollChannel.java
@@ -171,7 +171,25 @@ abstract class AbstractEpollChannel extends AbstractChannel implements UnixChann
             }
 
             if (isRegistered()) {
-                doDeregister();
+                // Need to check if we are on the EventLoop as doClose() may be triggered by the GlobalEventExecutor
+                // if SO_LINGER is used.
+                //
+                // See https://github.com/netty/netty/issues/7159
+                EventLoop loop = eventLoop();
+                if (loop.inEventLoop()) {
+                    doDeregister();
+                } else {
+                    loop.execute(new Runnable() {
+                        @Override
+                        public void run() {
+                            try {
+                                doDeregister();
+                            } catch (Exception cause) {
+                                pipeline().fireExceptionCaught(cause);
+                            }
+                        }
+                    });
+                }
             }
         } finally {
             socket.close();
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollChannel.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollChannel.java
@@ -184,7 +184,7 @@ abstract class AbstractEpollChannel extends AbstractChannel implements UnixChann
                         public void run() {
                             try {
                                 doDeregister();
-                            } catch (Exception cause) {
+                            } catch (Throwable cause) {
                                 pipeline().fireExceptionCaught(cause);
                             }
                         }
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollServerChannel.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollServerChannel.java
@@ -75,12 +75,8 @@ public abstract class AbstractEpollServerChannel extends AbstractEpollChannel im
 
     @UnstableApi
     @Override
-    protected final void doShutdownOutput(Throwable cause) throws Exception {
-        try {
-            super.doShutdownOutput(cause);
-        } finally {
-            close();
-        }
+    protected final void doShutdownOutput() throws Exception {
+        doClose();
     }
 
     abstract Channel newChildChannel(int fd, byte[] remote, int offset, int len) throws Exception;
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollServerChannel.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollServerChannel.java
@@ -23,7 +23,6 @@ import io.netty.channel.ChannelPipeline;
 import io.netty.channel.ChannelPromise;
 import io.netty.channel.EventLoop;
 import io.netty.channel.ServerChannel;
-import io.netty.util.internal.UnstableApi;
 
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
@@ -71,12 +70,6 @@ public abstract class AbstractEpollServerChannel extends AbstractEpollChannel im
     @Override
     protected Object filterOutboundMessage(Object msg) throws Exception {
         throw new UnsupportedOperationException();
-    }
-
-    @UnstableApi
-    @Override
-    protected final void doShutdownOutput() throws Exception {
-        doClose();
     }
 
     abstract Channel newChildChannel(int fd, byte[] remote, int offset, int len) throws Exception;
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollStreamChannel.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollStreamChannel.java
@@ -545,47 +545,13 @@ public abstract class AbstractEpollStreamChannel extends AbstractEpollChannel im
 
     @UnstableApi
     @Override
-    protected final void doShutdownOutput(Throwable cause) throws Exception {
-        try {
-            // The native socket implementation may throw a NotYetConnected exception when we attempt to shut it down.
-            // However NIO doesn't propagate an exception in the same situation (write failure), and we just want to
-            // update the socket state to flag that it has been shutdown. So don't use a voidPromise but instead create
-            // a new promise and ignore the results.
-            shutdownOutput0(newPromise());
-        } finally {
-            super.doShutdownOutput(cause);
-        }
-    }
-
-    private void shutdownOutput0(final ChannelPromise promise) {
-        try {
-            try {
-                socket.shutdown(false, true);
-            } finally {
-                ((AbstractUnsafe) unsafe()).shutdownOutput();
-            }
-            promise.setSuccess();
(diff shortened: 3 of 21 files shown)


gemmellr:
As mentioned on #7159 (but not marked as referenced here for some reason), I gave this a try and it resolved the issue seen.

transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollChannel.java
```diff
@@ -171,7 +171,24 @@ protected void doClose() throws Exception {
             }
 
             if (isRegistered()) {
-                doDeregister();
+                // Need to check if we are on the EventLoop as doClose() may be triggered by the GlobalEventExecutor
+                // if SO_LINGER is used.
+                //
+                // See https://github.com/netty/netty/issues/7159
+                if (eventLoop().inEventLoop()) {
```
Scottmitch: save a reference to `eventLoop()`

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/AbstractKQueueChannel.java
```diff
@@ -137,9 +137,27 @@ protected void doClose() throws Exception {
...
-                // events which are pending in kqueue to avoid referencing a deleted pointer at a later time.
-                doDeregister();
+                // since we rely upon jniSelfRef to be consistent we make sure that we clear this reference out for
+                // all events which are pending in kqueue to avoid referencing a deleted pointer at a later time.
+
+                // Need to check if we are on the EventLoop as doClose() may be triggered by the GlobalEventExecutor
+                // if SO_LINGER is used.
+                //
+                // See https://github.com/netty/netty/issues/7159
+                if (eventLoop().inEventLoop()) {
```
Scottmitch: save a reference to `eventLoop()`

normanmaurer:
So I looked more into this and this needs more work... Stay tuned for an updated PR..

gemmellr:
Not sure if the changes are complete yet, but just a note to say that I gave the current updated version a try and was again able to run the original tests repeatedly without observing any failure.

normanmaurer:
@gemmellr thanks! Will complete this during this week.

normanmaurer:
@Scottmitch @ejona86 PTAL again... this took quite some changes :(

transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollStreamChannel.java
```diff
@@ -676,27 +633,29 @@ public ChannelFuture shutdown() {
...
-                    shutdown0(promise);
-                }
-            });
-        } else {
-            EventLoop loop = eventLoop();
-            if (loop.inEventLoop()) {
-                shutdown0(promise);
-            } else {
-                loop.execute(new Runnable() {
+        shutdownOutput().addListener(new ChannelFutureListener() {
```
Scottmitch: if we are on the event loop and so_linger isn't set this will complete synchronously. consider adding the typical check for if the future is complete before adding a `ChannelFutureListener` in both these cases.

transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollServerChannel.java
```diff
@@ -75,12 +75,8 @@ protected Object filterOutboundMessage(Object msg) throws Exception {
...
     @UnstableApi
     @Override
-    protected final void doShutdownOutput(Throwable cause) throws Exception {
-        try {
-            super.doShutdownOutput(cause);
-        } finally {
-            close();
-        }
+    protected final void doShutdownOutput() throws Exception {
+        doClose();
```
Scottmitch: should we just make this the default method behavior in `AbstractChannel` and the channels which don't want to close can override and do a no-op?

transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollChannel.java
```diff
@@ -171,7 +171,25 @@ protected void doClose() throws Exception {
...
+                if (loop.inEventLoop()) {
+                    doDeregister();
+                } else {
+                    loop.execute(new Runnable() {
+                        @Override
+                        public void run() {
+                            try {
+                                doDeregister();
+                            } catch (Exception ignore) {
+                                // Just ignore
```
Scottmitch: log this and/or fire the exception in the pipeline?

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/AbstractKQueueStreamChannel.java
```diff
@@ -372,51 +376,8 @@ protected Object filterOutboundMessage(Object msg) {
...
-                socket.shutdown(true, true);
-            } finally {
-                ((AbstractUnsafe) unsafe()).shutdownOutput();
-            }
-            promise.setSuccess();
-        } catch (Throwable cause) {
-            promise.setFailure(cause);
-        }
+    protected final void doShutdownOutput() throws Exception {
+        socket.shutdown(false, true);
```
Scottmitch: +1 much cleaner ... thanks!

transport/src/main/java/io/netty/channel/AbstractChannel.java
```diff
@@ -616,24 +616,70 @@ public final void close(final ChannelPromise promise) {
...
+                } finally {
+                    closeOutboundBufferForShutdown(pipeline, outboundBuffer, shutdownCause);
+                }
+            }
+        }
+
+        private void closeOutboundBufferForShutdown(
+                ChannelPipeline pipeline, ChannelOutboundBuffer buffer, Throwable cause) {
+            buffer.failFlushed(cause, false);
+            buffer.close(cause, true);
```
Scottmitch: should we surround this code with a try/catch and fire an exception in the pipeline and/or log if something goes wrong?

transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollServerChannel.java
normanmaurer: @Scottmitch I thought about this but did not do it as I was thinking that this may have "bad effects" for people the implemented their own transport. WDYT ?

transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollStreamChannel.java
normanmaurer: yep let me fix and also for other transports.

transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollChannel.java
normanmaurer: done

transport/src/main/java/io/netty/channel/AbstractChannel.java
normanmaurer: I think it is fine to just let it bubble up as if it throws there is something seriously wrong.

Scottmitch: If the task is executed on the EventLoop it won't bubble up will it? It will just log a warning ... is that good enough?

normanmaurer: Thats true... That said I am not super concerned as if close(...) throws we have a bug in our impl so it should never happen. If you feel better I can definitely add a catch tho. WDYT ?

transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollServerChannel.java
Scottmitch: Hum it is possible but I wonder if its better to explicitly close in this case instead of silently doing nothing. It seems easier to detect this behavior rather than doing nothing and continuing to queue writes. Also if someone has a custom protocol and is calling `shutdownOutput` it seems likely they will expect some action to be taken and not just a noop.

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/AbstractKQueueChannel.java
```diff
@@ -137,9 +137,28 @@ protected void doClose() throws Exception {
...
+                EventLoop loop = eventLoop();
+                if (loop.inEventLoop()) {
+                    doDeregister();
+                } else {
+                    loop.execute(new Runnable() {
+                        @Override
+                        public void run() {
+                            try {
+                                doDeregister();
+                            } catch (Exception cause) {
```
Scottmitch: `Exception` -> `Throwable`? Otherwise some exceptions may not be propagated.

transport/src/main/java/io/netty/channel/AbstractChannel.java
Scottmitch: If its a programming error on our end and not expected to ever happen then logging a warning should be "good enough".

transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollServerChannel.java
normanmaurer: good point... let me do this.

normanmaurer:
@Scottmitch addressed your last round of comments. Will do some more testing and if everything works as expected I will merge this one

normanmaurer:
Cherry-picked into 4.1 (aa8bdb5d6b7f89d56712c2d177c27857a17d79cd) ... Opened another PR for 4.0 to ensure I not messed up when merging things and resolving conflicts.

PR #7163, opened 2017-08-29
Only load native transport if running architecture match the compiled…

normanmaurer:
… library architecture.

Motivation:

We should only try to load the native artifacts if the architecture we are currently running on is the same as the one the native libraries were compiled for.

Modifications:

Include architecture in native lib name and append the current arch when trying to load these. This will fail then if its not the same as the arch of the compiled arch.

Result:

Fixes [#7150].

### DIFF
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent.java
@@ -92,6 +92,8 @@ public final class PlatformDependent {
     private static final File TMPDIR = tmpdir0();
 
     private static final int BIT_MODE = bitMode0();
+    private static final String NORMALIZED_ARCH = normalizeArch(SystemPropertyUtil.get("os.arch", ""));
+    private static final String NORMALIZED_OS = normalizeOs(SystemPropertyUtil.get("os.name", ""));
 
     private static final int ADDRESS_SIZE = addressSize0();
     private static final boolean USE_DIRECT_BUFFER_NO_CLEANER;
@@ -1224,6 +1226,99 @@ public final class PlatformDependent {
         default:
             return hash;
         }
+    }
+
+    public static String normalizedArch() {
+        return NORMALIZED_ARCH;
+    }
+
+    public static String normalizedOs() {
+        return NORMALIZED_OS;
+    }
+
+    private static String normalize(String value) {
+        return value.toLowerCase(Locale.US).replaceAll("[^a-z0-9]+", "");
+    }
+
+    private static String normalizeArch(String value) {
+        value = normalize(value);
+        if (value.matches("^(x8664|amd64|ia32e|em64t|x64)$")) {
+            return "x86_64";
+        }
+        if (value.matches("^(x8632|x86|i[3-6]86|ia32|x32)$")) {
+            return "x86_32";
+        }
+        if (value.matches("^(ia64|itanium64)$")) {
+            return "itanium_64";
+        }
+        if (value.matches("^(sparc|sparc32)$")) {
+            return "sparc_32";
+        }
+        if (value.matches("^(sparcv9|sparc64)$")) {
+            return "sparc_64";
+        }
+        if (value.matches("^(arm|arm32)$")) {
+            return "arm_32";
+        }
+        if ("aarch64".equals(value)) {
+            return "aarch_64";
+        }
+        if (value.matches("^(ppc|ppc32)$")) {
+            return "ppc_32";
+        }
+        if ("ppc64".equals(value)) {
+            return "ppc_64";
+        }
+        if ("ppc64le".equals(value)) {
+            return "ppcle_64";
+        }
+        if ("s390".equals(value)) {
+            return "s390_32";
+        }
+        if ("s390x".equals(value)) {
+            return "s390_64";
+        }
+
+        return "unknown";
+    }
+
+    private static String normalizeOs(String value) {
+        value = normalize(value);
+        if (value.startsWith("aix")) {
+            return "aix";
+        }
+        if (value.startsWith("hpux")) {
+            return "hpux";
+        }
+        if (value.startsWith("os400")) {
+            // Avoid the names such as os4000
+            if (value.length() <= 5 || !Character.isDigit(value.charAt(5))) {
+                return "os400";
+            }
+        }
+        if (value.startsWith("linux")) {
+            return "linux";
+        }
+        if (value.startsWith("macosx") || value.startsWith("osx")) {
+            return "osx";
+        }
+        if (value.startsWith("freebsd")) {
+            return "freebsd";
+        }
+        if (value.startsWith("openbsd")) {
+            return "openbsd";
+        }
+        if (value.startsWith("netbsd")) {
+            return "netbsd";
+        }
+        if (value.startsWith("solaris") || value.startsWith("sunos")) {
+            return "sunos";
+        }
+        if (value.startsWith("windows")) {
+            return "windows";
+        }
+
+        return "unknown";
     }
 
     private static final class AtomicLongCounter extends AtomicLong implements LongCounter {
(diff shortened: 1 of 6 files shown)


normanmaurer:
@michaelandrepearce @doom369 PTAL

normanmaurer:
also @johnou

doom369:
Thanks! Will try to check soon.

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/Native.java
```diff
@@ -111,7 +111,8 @@ private static void loadNativeLibrary() {
         if (!name.startsWith("mac") && !name.contains("bsd") && !name.startsWith("darwin")) {
```
michaelandrepearce: Should this be replaced with calls to PlatformDependent to get this information? And missing ones added there?

transport-native-epoll/src/main/java/io/netty/channel/epoll/Native.java
```diff
@@ -196,7 +196,8 @@ private static void loadNativeLibrary() {
         if (!name.startsWith("linux")) {
```
michaelandrepearce: Should this be replaced with calls to PlatformDependent to get this information? And missing ones added there?

handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
```diff
@@ -404,14 +402,14 @@ static long memoryAddress(ByteBuf buf) {
...
-        String arch = normalizeArch(SystemPropertyUtil.get("os.arch", ""));
+        String os = PlatformDependent.normalizedOs();
+        String arch = PlatformDependent.normalizedArch();
 
         Set<String> libNames = new LinkedHashSet<String>(4);
         // First, try loading the platform-specific library. Platform-specific
         // libraries will be available if using a tcnative uber jar.
         libNames.add("netty_tcnative_" + os + '_' + arch);
-        if (LINUX.equalsIgnoreCase(os)) {
+        if ("linux".equalsIgnoreCase(os)) {
```
michaelandrepearce: Should this be replaced with calls to PlatformDependent to get if linux?

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -1226,6 +1228,99 @@ static int hashCodeAsciiSafe(byte[] bytes, int startPos, int length) {
...
+            return "netbsd";
+        }
+        if (value.startsWith("solaris") || value.startsWith("sunos")) {
+            return "sunos";
+        }
+        if (value.startsWith("windows")) {
+            return "windows";
+        }
+
+        return "unkown";
```
johnou: typo, "unknown"

transport-native-epoll/pom.xml
```diff
@@ -179,7 +179,7 @@
                       <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
                     </manifest>
                     <manifestEntries>
-                      <Bundle-NativeCode>META-INF/native/libnetty_transport_native_epoll.so; osname=linux; processor=x86_64,*</Bundle-NativeCode>
+                      <Bundle-NativeCode>META-INF/native/libnetty_transport_native_epoll_${os.detected.arch}.so; osname=linux; processor=${os.detected.arch},*</Bundle-NativeCode>
```
johnou: :ok_hand:

michaelandrepearce: an alternative option is, could remove <platform>.</platform> hawtjni will then place these in folders correctly based on the arch

normanmaurer: this would have been more complex as we would also need to adjust our code that loads the native libs etc.

common/src/main/java/io/netty/util/internal/PlatformDependent.java
normanmaurer: fixed

handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
normanmaurer: maybe... I did want to keep the change focused ;)

transport-native-epoll/src/main/java/io/netty/channel/epoll/Native.java
normanmaurer: maybe... I did want to keep the change focused ;)

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/Native.java
normanmaurer: maybe... I did want to keep the change focused

transport-native-epoll/pom.xml
michaelandrepearce: Fair enough, as noted this approach works +1

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/Native.java
michaelandrepearce: Ok no worries any obj if i go do this after then? new pr post merge of this?

normanmaurer: let us merge this one first and then we can see if it makes sense to move some of the stuff (I think it may make sense tho ;))

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -1226,6 +1228,99 @@ static int hashCodeAsciiSafe(byte[] bytes, int startPos, int length) {
...
+        value = normalize(value);
+        if (value.startsWith("aix")) {
+            return "aix";
+        }
+        if (value.startsWith("hpux")) {
+            return "hpux";
+        }
+        if (value.startsWith("os400")) {
+            // Avoid the names such as os4000
+            if (value.length() <= 5 || !Character.isDigit(value.charAt(5))) {
```
netkins: MAJOR Merge this if statement with the enclosing one. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1066)

normanmaurer: I would like to keep it as it is for now as the code was just moved and it kind of mirrors what is in the maven-os-plugin.

transport-native-epoll/pom.xml
Scottmitch: This will break folks that are shading ... I don't have a problem with this but just good to note.

common/src/main/java/io/netty/util/internal/PlatformDependent.java
Scottmitch: It would be nice to consolidate this platform logic somewhere ... seems like we keep copying/moving it around.

transport-native-epoll/pom.xml
normanmaurer: @Scottmitch imho how these files are named is an implementation detail so it should be ok ?

common/src/main/java/io/netty/util/internal/PlatformDependent.java
normanmaurer: agree but the os plugin is not part of netty :)

transport-native-epoll/pom.xml
ejona86: When shading you have to rename them, that means having some level of knowledge how they are named.

The recent change from `-` to `_` broke existing configuration for shading as well. Breakages like that should be "okay," but avoiding useless breakages would be good. This breakage seems unavoidable.

normanmaurer: @ejona86 agree, we should only do if there is a good reason. I think this is one.

johnou: @normanmaurer worth while a mention in the release notes then :+1:

normanmaurer:
Cherry-picked not 4.1 (0fffc844d6d4c8e196a8448a45e7ff542446f708) and 4.0 (1ee2930515c594757769e08eace028eeb2450920).

I will ensure we mention it in the release notes.

PR #7165, opened 2017-08-30
Use threadsafe setter on Atomic Updaters

carl-mastrangelo:
Motivation:
The documentation for field updates says:

> Note that the guarantees of the `compareAndSet`
> method in this class are weaker than in other atomic classes.
> Because this class cannot ensure that all uses of the field
> are appropriate for purposes of atomic access, it can
> guarantee atomicity only with respect to other invocations of
> `compareAndSet` and `set` on the same updater.

This implies that volatiles shouldn't use normal assignment; the
updater should set them.

Modifications:
Use setter for field updaters that make use of compareAndSet.

Result:
Concurrency compliant code

### DIFF
--- a/common/src/main/java/io/netty/util/HashedWheelTimer.java
+++ b/common/src/main/java/io/netty/util/HashedWheelTimer.java
@@ -97,8 +97,8 @@ public class HashedWheelTimer implements Timer {
     public static final int WORKER_STATE_INIT = 0;
     public static final int WORKER_STATE_STARTED = 1;
     public static final int WORKER_STATE_SHUTDOWN = 2;
-    @SuppressWarnings({ "unused", "FieldMayBeFinal", "RedundantFieldInitialization" })
-    private volatile int workerState = WORKER_STATE_INIT; // 0 - init, 1 - started, 2 - shut down
+    @SuppressWarnings({ "unused", "FieldMayBeFinal" })
+    private volatile int workerState; // 0 - init, 1 - started, 2 - shut down
 
     private final long tickDuration;
     private final HashedWheelBucket[] wheel;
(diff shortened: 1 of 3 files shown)


common/src/main/java/io/netty/util/HashedWheelTimer.java
```diff
@@ -253,6 +253,7 @@ public HashedWheelTimer(
         if (ticksPerWheel <= 0) {
             throw new IllegalArgumentException("ticksPerWheel must be greater than 0: " + ticksPerWheel);
         }
+        WORKER_STATE_UPDATER.set(this, WORKER_STATE_INIT);
```
johnou: isn't this redundant because the variable int with no default value defined defaults to 0?

carl-mastrangelo: Yes, but it is what the original code did.  (also, the JLS says that the default initialization of the field will always be zero, so I don't think it technically needs to be done).   I think it serves mostly as documentation.

normanmaurer: yeah I think its not needed but also not harms really as instances of `HashedWheelTimer` should not be created very frequently.

normanmaurer:
@carl-mastrangelo can you please check the compilation error ?

carl-mastrangelo:
@normanmaurer Done.  


I also updated `AbstractReferenceCountedByteBuf`  to use the setter for init.

normanmaurer:
Cherry-picked into 4.1 (7528e5a11e6edeb847d7913437a1c88b1bb30e9d) and 4.0 (3c24d0e062f77f51324fce75f5ee438c001e7a0d)

normanmaurer:
@carl-mastrangelo thanks !

PR #7168, opened 2017-08-31
Allow to construct UnpooledByteBufAllocator that explictly always use…

normanmaurer:
… sun.misc.Cleaner

Motivation:

When the user want to have the direct memory explicitly managed by the GC (just as java.nio does) it is useful to be able to construct an UnpooledByteBufAllocator that allows this without the chances to see any memory leak.

Modifications:

Allow to explicitly disable the usage of reflection to construct direct ByteBufs and so be sure these will be collected by GC.

Result:

More flexible way to use the UnpooledByteBufAllocator.


normanmaurer:
Cherry-picked into 4.1 (bca35b0449) and 4.0 (30b62266ad)

PR #7174, opened 2017-09-01
Make UnpooledDirectByteBuf, UnpooledHeapByteBuf and UnpooledUnsafeDir…

normanmaurer:
…ectByteBuf constructors public.

Motivation:

The constrcutors a protected atm but the classes are public. We should make the constructors public as well to make it easier to write your own ByteBufAllocator.

Modifications:

Change constructors to be public and add some javadocs.

Result:

Easier to create own ByteBufAllocator.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/UnpooledDirectByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/UnpooledDirectByteBuf.java
@@ -28,9 +28,9 @@ import java.nio.channels.GatheringByteChannel;
 import java.nio.channels.ScatteringByteChannel;
 
 /**
- * A NIO {@link ByteBuffer} based buffer.  It is recommended to use {@link Unpooled#directBuffer(int)}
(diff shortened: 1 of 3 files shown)


normanmaurer:
@Scottmitch WDYT ? Should we do this or not ... The downside that I can think of is that it may confuse users in a way that they think they should use these directly while this is only true if you write your own allocator.

Scottmitch:
I think this makes sense. We could add some additional comments on the constructor if you think that is useful. `Typically these objects are only constructed via ByteBufAllocator...`

normanmaurer:
I think it is fine as it is and it is also fine in terms of binary compatibility:

https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html#jls-13.4.7

normanmaurer:
Cherry-picked into 4.1 (9d56439aa1c221d38ee3b526c63d268c2e6f16c7) and 4.0 (671ea0ad9a)

Issue #7172, opened 2017-08-31
too many releases in Http2FrameCodec in cleartext upgrade


mosesn:
### Expected behavior
No refcounting issues

### Actual behavior
io.netty.util.IllegalReferenceCountException: refCnt: 0, decrement: 1

### Steps to reproduce
Try to use HttpServerUpgradeHandler with Http2MultiplexCodec.  The request will come in, and netty writes an upgrade response to the wire.  After that succeeds, netty turns the request into a user event, with a ref count of 1.  Its ref count is incremented to 2, then it's sent to Http2FrameCodec through a userEventTriggered.  It expects that after the userEventTriggered is called, it'll come back with a ref count of 1 again.

However, inside of Http2FrameCodec#userEventTriggered, it decrements the ref count until it hits 0.  In particular, `InboundHttpToHttp2Adapter.handle(ctx, connection(), decoder().frameListener(), upgrade.upgradeRequest())` and `upgrade.release()` each decrement the ref count once.

### Minimal yet complete reproducer code (or URL to code)
https://github.com/twitter/finagle/blob/develop/finagle-http2/src/main/scala/com/twitter/finagle/http2/Http2CleartextServerInitializer.scala#L104-L108

try upgrading this to 4.1.15

### Netty version
4.1.15

PR #7176, opened 2017-09-01
Fix reference count issue when using Http2FrameCodec / Http2Multiplex…

normanmaurer:
…Codec with HttpServerUpgradeHandler

Motivation:

When using  Http2FrameCodec / Http2MultiplexCodec with HttpServerUpgradeHandler reference count exception will be triggered.

Modifications:

- Correctly retain before calling InboundHttpToHttp2Adapter.handle
- Add unit test

Result:

Fixes [#7172].

### DIFF
--- a/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2FrameCodecTest.java
+++ b/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2FrameCodecTest.java
@@ -24,9 +24,13 @@ import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelPromise;
 import io.netty.channel.embedded.EmbeddedChannel;
 import io.netty.handler.codec.UnsupportedMessageTypeException;
+import io.netty.handler.codec.http.DefaultFullHttpRequest;
+import io.netty.handler.codec.http.FullHttpRequest;
 import io.netty.handler.codec.http.HttpMethod;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http.HttpScheme;
+import io.netty.handler.codec.http.HttpServerUpgradeHandler;
+import io.netty.handler.codec.http.HttpVersion;
 import io.netty.handler.codec.http2.Http2Exception.StreamException;
 import io.netty.handler.codec.http2.Http2Stream.State;
 import io.netty.handler.logging.LogLevel;
@@ -40,6 +44,8 @@ import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 import org.mockito.ArgumentCaptor;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
 
 import java.net.InetSocketAddress;
 import java.util.HashSet;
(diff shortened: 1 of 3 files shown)


normanmaurer:
@mosesn PTAL

codec-http2/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
normanmaurer: this is needed to be able to mock final classes.

See:
https://github.com/mockito/mockito/wiki/What%27s-new-in-Mockito-2#mock-the-unmockable-opt-in-mocking-of-final-classesmethods

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2FrameCodecTest.java
```diff
@@ -641,6 +647,30 @@ public void operationComplete(ChannelFuture future) throws Exception {
...
+        });
+        when(upgradeEvent.release()).thenAnswer(new Answer<Boolean>() {
+            @Override
+            public Boolean answer(InvocationOnMock invocationOnMock) throws Throwable {
+                return request.release();
+            }
+        });
+
+        when(upgradeEvent.upgradeRequest()).thenReturn(request);
+        channel.pipeline().fireUserEventTriggered(upgradeEvent);
```
mosesn: can we also check the reference count on upgradeEvent directly on the request here?

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2FrameCodecTest.java
```diff
@@ -641,6 +647,30 @@ public void operationComplete(ChannelFuture future) throws Exception {
         assertTrue(listenerExecuted.get());
     }
 
+    @Test
+    public void upgradeEventNoRefCntError() throws Http2Exception {
+        frameListener.onHeadersRead(http2HandlerCtx, Http2CodecUtil.HTTP_UPGRADE_STREAM_ID, request, 31, false);
+
+        final FullHttpRequest request = new DefaultFullHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.GET, "/");
+        final HttpServerUpgradeHandler.UpgradeEvent upgradeEvent = mock(HttpServerUpgradeHandler.UpgradeEvent.class);
```
mosesn: why do we mock this?  can we construct a real upgrade event here?

normanmaurer: the constructor is package-private.

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2FrameCodecTest.java
normanmaurer: sure... that said without the fix it fails the test because it tries to release everything that is stored in the `EmbeddedChannel`, so I think its not really needed.

mosesn: yeah, I figured it might be nice to help avoid regressions where it's +1 instead of -1.

normanmaurer: good point let me do this

normanmaurer: done

normanmaurer:
Cherry-picked into 4.1 (379ac890f4dbec15d19714711f85455a12112c3f)

PR #7177, opened 2017-09-04
Ensure the tests complete on java7 and java9 as well.

normanmaurer:
Motivation:

379ac890f4dbec15d19714711f85455a12112c3f introduced the usage of the inline mock maker. This unfortunally not work on java7 and java9.

Modifications:

Just use reflection to create the event for now.

Result:

Netty tests pass again on java7 and java9 as well.


normanmaurer:
Cherry-picked into 4.1 (5c572f0f63da22392d477697c4c289072915d483) to make the CI happy again

Issue #7178, opened 2017-09-05
The ChannelHandler which is used to construct Http2MultiplexCodecBuilder will also be added to out bound streams


Apache9:
### Expected behavior
The `ChannelHandler` which is used to construct `Http2MultiplexCodecBuilder` should only be added to remotely-created streams.
### Actual behavior
It is also added to out bound streams.
### Steps to reproduce
Modify `Http2MultiplexCodecBuilderTest`, replace the code in `setUp`

```
        Bootstrap cb = new Bootstrap()
                .channel(LocalChannel.class)
                .group(group)
                .handler(new Http2MultiplexCodecBuilder(false, new TestChannelInitializer()).build());
```
to
```
        Bootstrap cb = new Bootstrap()
                .channel(LocalChannel.class)
                .group(group)
                .handler(new Http2MultiplexCodecBuilder(false, new ChannelInitializer<Channel>() {

                    @Override
                    protected void initChannel(Channel ch) throws Exception {
                        throw new RuntimeException("should not be call for outbound stream");
                    }
                }).build());
```
Then `createOutboundStream` and `multipleOutboundStreams` will both fail.
### Minimal yet complete reproducer code (or URL to code)
Just apply this patch
https://github.com/Apache9/netty/commit/2dcbae3109bbab5699dbfd64176694c8d43f69ca
### Netty version
4.1.15.Final
### JVM version (e.g. `java -version`)
java version "1.8.0_121"
Java(TM) SE Runtime Environment (build 1.8.0_121-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.121-b13, mixed mode)
### OS version (e.g. `uname -a`)
Linux zhangduo-ubuntu 4.4.0-93-generic #116-Ubuntu SMP Fri Aug 11 21:17:51 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

Apache9:
The stracktrace of the exception when running modified `Http2MultiplexCodecBuilderTest`.
```
15:49:13.120 [defaultEventLoop-1-1] WARN  io.netty.channel.ChannelInitializer - Failed to initialize a channel. Closing: [id: 0x70bc891d, L:local:E:e07decb9 - R:local:io.netty.handler.codec.http2.http2multiplexcodecbuildertest](HTTP/2 - -1)
java.lang.RuntimeException: should not be call for outbound stream
[...]

PR #7180, opened 2017-09-05
Not add inboundStreamHandler for outbound streams created by Http2Mul…

normanmaurer:
…tiplexCodec.

Motivation:

We must not add the inboundStreamHandler for outbound streams creates by Http2MultiplexCodec as the user will specify a handler via Http2StreamChannelBootstrap.

Modifications:

- Check if the stream is for outbound and if so not add the inboundStreamHandler to the pipeline
- Update tests so this is covered.

Result:

Fixes [#7178]

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
@@ -239,7 +239,7 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
                     break;
                 }
                 // fall-trough
-                DefaultHttp2StreamChannel childChannel = new DefaultHttp2StreamChannel(stream);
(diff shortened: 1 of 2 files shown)


normanmaurer:
Also @Apache9.

ejona86:
This seems fine for now. I think we'll eventually want to improve `Http2StreamChannelBootstrap` so that the handler is added before the channel is registered (e.g., we could pass `handler` to `newOutboundStream()`). But it looks like `Http2StreamChannelBootstrap` could use other improvements during `init()`, like how late it sets channel options/attributes. It'll probably need a refactor later.

normanmaurer:
+1... baby steps ;)

Apache9:
+1

normanmaurer:
Cherry-picked into 4.1 (870b5f5e4b4eef8ea3baf650fa9c592533b2fda2)

Issue #7179, opened 2017-09-05
Http2MultiplexCodec can not work together with ChunkedWriteHandler


Apache9:
### Expected behavior
When used at server side, the `ChunkedWriteHandler` should be able to write out data chunk by chunk(multiple `Http2DataFrame`s)
### Actual behavior
The `ChunkedWriteHandler` always finds out that the channel is not writable so nothing can be written out.
### Steps to reproduce
Use `ChunkedWriteHandler` at server side.
### Minimal yet complete reproducer code (or URL to code)
See this UT
https://github.com/Apache9/netty/commit/001ec097a56c3a818819875acba881cb06dc804d
### Netty version
4.1.15.Final
### JVM version (e.g. `java -version`)
java version "1.8.0_144"
Java(TM) SE Runtime Environment (build 1.8.0_144-b01)
Java HotSpot(TM) 64-Bit Server VM (build 25.144-b01, mixed mode)
### OS version (e.g. `uname -a`)
Linux ubuntu 4.10.0-27-generic #30~16.04.2-Ubuntu SMP Thu Jun 29 16:07:46 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

Apache9:
The problem is that the `writable` field for server side `DefaultHttp2StreamChannel` is always `false`. 

At client side, the `writable` will be changed after calling the `onStreamActive` in the `Http2ConnectionAdapter` of `DefaultHttp2RemoteFlowController`. It will finally reach the `DefaultHttp2StreamChannel.writabilityChanged`.

But at server side, the calling trace is broken up at `Http2FrameCodec.Http2RemoteFlowControllerListener.writabilityChanged`. In this method, the `Http2FrameStream` gotten from the `Http2Stream` is null, so we give up.

https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java#L588

No idea why it is null only at server side...

normanmaurer:
@Apache9 can I reuse some of your test-code within netty and include it as part of the project under the AL2 license ?

Apache9:
Sure.

normanmaurer:
Fixed by https://github.com/netty/netty/pull/7181

Issue #7082, opened 2017-08-10
h2childchan: DefaultHttp2StreamChannel initial writability


Scottmitch:
Currently the initial writability of `DefaultHttp2StreamChannel` is false if we have not yet created the underlying `Http2Stream` object. Typically a channel is default writability is true, unless conditions otherwise prohibit this. In the case of `DefaultHttp2StreamChannel` we should investigate if the following policy makes sense: the default state is writable unless the connection window has collapsed, and we may also want to consider the parent channel's writability.

https://github.com/netty/netty/pull/6896#discussion_r132339750

normanmaurer:
This is "fixed" by https://github.com/netty/netty/pull/7181 ... To make it short no it makes no sense.

normanmaurer:
Fixed by #7181

Scottmitch:
@normanmaurer - Can you explain why the policy suggested here "makes no sense"? The policy implemented in #7181 seems incomplete and may incorrectly report writability in a few cases:
- depending on initial settings for outbound streams (https://github.com/netty/netty/pull/7181#discussion_r139296524)
- depending on connection window
- depending on channel writability

PR #7181, opened 2017-09-06
Correctly update writability state of Http2StreamChannel created by H…

normanmaurer:
…ttp2MultiplexCodec.

Motivation:

We missed to mark the Http2StreamChannel as writable in some cases which could lead to the situation that a Channel never becomes writable. Also when a Http2StreamChannel was created we always marked it non-writable at the beginning which means if the user will only start writing once the Channel becomes writable it will never happen as it only became writable after the first header was written.

Modifications:

- Correctly handle updates for writability in all cases
- Change unit tests to cover this.

Result:

Fixes [#7179].

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
@@ -582,6 +582,11 @@ public class Http2FrameCodec extends Http2ConnectionHandler {
         ctx.fireExceptionCaught(cause);
     }
 
+    final boolean isWritable(DefaultHttp2FrameStream stream) {
+        Http2Stream s = stream.stream;
+        return s != null && connection().remote().flowController().isWritable(s);
+    }
+
     private final class Http2RemoteFlowControllerListener implements Http2RemoteFlowController.Listener {
         @Override
         public void writabilityChanged(Http2Stream stream) {
@@ -601,7 +606,7 @@ public class Http2FrameCodec extends Http2ConnectionHandler {
     static class DefaultHttp2FrameStream implements Http2FrameStream {
 
         private volatile int id = -1;
-        private volatile Http2Stream stream;
+        volatile Http2Stream stream;
 
         DefaultHttp2FrameStream setStreamAndProperty(PropertyKey streamKey, Http2Stream stream) {
             assert id == -1 || stream.id() == id;
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
@@ -231,16 +231,17 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
 
     @Override
     final void onHttp2StreamStateChanged(ChannelHandlerContext ctx, Http2FrameStream stream) {
+        Http2MultiplexCodecStream s = (Http2MultiplexCodecStream) stream;
+
         switch (stream.state()) {
             case HALF_CLOSED_REMOTE:
             case OPEN:
-                if (((Http2MultiplexCodecStream) stream).channel != null) {
+                if (s.channel != null) {
                     // ignore if child channel was already created.
                     break;
                 }
                 // fall-trough
-                DefaultHttp2StreamChannel childChannel = new DefaultHttp2StreamChannel(stream, false);
-                ChannelFuture future = ctx.channel().eventLoop().register(childChannel);
+                ChannelFuture future = ctx.channel().eventLoop().register(new DefaultHttp2StreamChannel(s, false));
                 if (future.isDone()) {
                     registerDone(future);
                 } else {
@@ -248,7 +249,7 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
                 }
                 break;
             case CLOSED:
-                DefaultHttp2StreamChannel channel = ((Http2MultiplexCodecStream) stream).channel;
+                DefaultHttp2StreamChannel channel = s.channel;
                 if (channel != null) {
                     channel.streamClosed();
                 }
@@ -395,17 +396,25 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
         READ_PROCESSED_OK_TO_PROCESS_MORE
     }
 
+    private boolean initialWritability(DefaultHttp2FrameStream stream) {
+        // If the stream id is not valid yet we will just mark the channel as writable as we will be notified
+        // about non-writability state as soon as the first Http2HeaderFrame is written (if needed).
+        // This should be good enough and simplify things a lot.
+        return !isStreamIdValid(stream.id()) || isWritable(stream);
+    }
+
     // TODO: Handle writability changes due writing from outside the eventloop.
     private final class DefaultHttp2StreamChannel extends DefaultAttributeMap implements Http2StreamChannel {
         private final Http2StreamChannelConfig config = new Http2StreamChannelConfig(this);
         private final Http2ChannelUnsafe unsafe = new Http2ChannelUnsafe();
         private final ChannelId channelId;
         private final ChannelPipeline pipeline;
-        private final Http2FrameStream stream;
+        private final DefaultHttp2FrameStream stream;
         private final ChannelPromise closePromise;
         private final boolean outbound;
 
         private volatile boolean registered;
+        // We start with the writability of the channel when creating the StreamChannel.
         private volatile boolean writable;
 
         private boolean closePending;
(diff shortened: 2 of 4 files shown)


normanmaurer:
@Apache9 please check as well... This should fix #7179

Apache9:
Not very familiar with the modifications here but my UT passes.

Thanks @normanmaurer  for the quick fix.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
```diff
@@ -582,6 +582,11 @@ void onHttp2FrameStreamException(ChannelHandlerContext ctx, Http2FrameStreamExce
         ctx.fireExceptionCaught(cause);
     }
 
+    final boolean isWritable(DefaultHttp2FrameStream stream) {
+        Http2Stream s = stream.stream;
+        return s == null ? false : connection().remote().flowController().isWritable(s);
```
netkins: MAJOR Ternary operators that can be simplified with || or && [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3ASimplifiedTernary)

normanmaurer:
@Apache9 can you verify again. I did miss something before.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -395,17 +396,25 @@ void onBytesConsumed(@SuppressWarnings("unused") ChannelHandlerContext ctx,
         READ_PROCESSED_OK_TO_PROCESS_MORE
     }
 
+    private boolean initialWritability(DefaultHttp2FrameStream stream) {
+        // If the stream id is not valid yet we will just mark the channel as writable as we will be notified
+        // about non-writability state as soon as the first Http2HeaderFrame is written (if needed).
+        // This should be good enough and simplify things a lot.
+        return isStreamIdValid(stream.id()) ? isWritable(stream) : true;
```
netkins: MAJOR Ternary operators that can be simplified with || or && [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3ASimplifiedTernary)

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
```diff
@@ -582,6 +582,11 @@ void onHttp2FrameStreamException(ChannelHandlerContext ctx, Http2FrameStreamExce
         ctx.fireExceptionCaught(cause);
     }
 
+    final boolean isWritable(DefaultHttp2FrameStream stream) {
+        Http2Stream s = stream.stream;
+        return s != null && connection().remote().flowController().isWritable(s);
```
netkins: MAJOR Ternary operators that can be simplified with || or && [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3ASimplifiedTernary)

normanmaurer: false positive ?

Apache9:
@normanmaurer  It still works. Thanks.

normanmaurer:
@Apache9 perfect... thanks :)

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -395,17 +396,25 @@ void onBytesConsumed(@SuppressWarnings("unused") ChannelHandlerContext ctx,
         READ_PROCESSED_OK_TO_PROCESS_MORE
     }
 
+    private boolean initialWritability(DefaultHttp2FrameStream stream) {
+        // If the stream id is not valid yet we will just mark the channel as writable as we will be notified
+        // about non-writability state as soon as the first Http2HeaderFrame is written (if needed).
+        // This should be good enough and simplify things a lot.
+        return !isStreamIdValid(stream.id()) || isWritable(stream);
```
netkins: MAJOR Ternary operators that can be simplified with || or && [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3ASimplifiedTernary)

normanmaurer:
@ejona86 ping

normanmaurer:
Also related and fixes https://github.com/netty/netty/issues/7082

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
normanmaurer: for client side its triggered by this code path:

https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java#L397

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -1011,6 +1020,9 @@ public void operationComplete(ChannelFuture future) throws Exception {
             private void firstWriteComplete(ChannelFuture future, ChannelPromise promise) {
                 Throwable cause = future.cause();
                 if (cause == null) {
+                    // As we just finished our first write which made the stream-id valid we need to re-evulate
```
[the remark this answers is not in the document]
normanmaurer: fixed

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
ejona86: Oh, I see now. I don't feel bad missing that slight-of-hand. That does look like it will break when hitting MAX_CONCURRENT_STREAMS since StreamBufferingEncoder won't create the stream immediately.

So if `s == null` here, then things are weird/broken elsewhere, so it shouldn't really matter whether we consider it writable or not.

normanmaurer:
Cherry-picked into 4.1 (bf0a53179a3057ac075c1192f80917ddd5ee1d9e)

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
Scottmitch: This may not be correct in all cases. The peer can send a settings frame with initial_window_size of 0 which would mean new streams are not writable. A server/client may do this if they are being overwhelmed and/or was overly optimistic about how many resources it initially advertised ... not sure how common it is in practice.

We should consider checking the peer's initial window size if this is a locally created stream.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -1011,6 +1020,9 @@ public void operationComplete(ChannelFuture future) throws Exception {
             private void firstWriteComplete(ChannelFuture future, ChannelPromise promise) {
                 Throwable cause = future.cause();
                 if (cause == null) {
+                    // As we just finished our first write which made the stream-id valid we need to re-evaluate
+                    // the writability of the channel.
+                    writabilityChanged(Http2MultiplexCodec.this.isWritable(stream));
```
Scottmitch: can we rename `writabilityChanged` to `checkForWritabilityChange` (or something similar)? At this point we are not indicating there is a writability change ... we are just checking if there has been a change.

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecBuilderTest.java
```diff
@@ -74,7 +75,34 @@ public void setUp() throws InterruptedException {
...
+
+                                    @Override
+                                    public void channelWritabilityChanged(ChannelHandlerContext ctx) throws Exception {
+                                        writable |= ctx.channel().isWritable();
+                                        super.channelWritabilityChanged(ctx);
+                                    }
+
+                                    @Override
+                                    public void channelInactive(ChannelHandlerContext ctx) throws Exception {
+                                        assertTrue(writable);
```
Scottmitch: I guess we currently don't have a test which closes a channel after it has written to fast, to much, or the peer stops reading? Seems like a potentially "flaky" check to have here ... if conditions/tests change ... can you add a comment?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
ejona86: The writability needs to be true so the application provides the initial headers. After we have the headers the stream exists and this won't be used any more. So the application shouldn't end up writing any DATA.

normanmaurer: I think what @ejona86 said is correct and so it should be correct @Scottmitch

Scottmitch: Ahh I see ... yes this makes sense. Although we don't make the same consideration for trailers. We don't really know when the trailers will be written so it is possible the user will write data until not writable and then want to write trailers but if they are using channel writability to conditionally write then they will have to wait. Writing trailers maybe less common, but it seems strange to exhibit different behavior in these scenarios.

ejona86: This has nothing to do with HEADERS vs DATA and how they are impacted by HTTP/2 flow control. It has everything to do with "the stream doesn't exist before the initial headers." So I don't really see the trailers as being weird at all. If client could send multiple headers (say, like server can with 1xx codes) then we'd still want this same logic.

Scottmitch: Yip understood this is to facilitate stream creation.

PR #7182, opened 2017-09-06
Not log notify failure for DelegatingChannelPromiseNotifier when prom…

normanmaurer:
…ise is VoidChannelPromise

Motivation:

We should not log by default if the promise is a VoidChannelPromise as its try* methods will always return false.

Modifications:

Do an instanceof check to determine if we should log or not by default

Result:

No more noise in the logs when using a VoidChannelPromise.


normanmaurer:
Noticed this while running the http2 multiplex tests.

normanmaurer:
Pulling this in as its an easy one and there is no risk. 4.1 (a739d89792dbc21ad2f1384e018e547dadff2763)

Issue #7173, opened 2017-08-31
NPE in cleartext upgrade


mosesn:
### Expected behavior
No null pointer exceptions

### Actual behavior
When an HTTP/1.1 upgrade request comes in, we call `onHttpServerUpgrade` in Http2ConnectionHandler.  This makes us create a stream, which eventually calls `new DefaultHttp2StreamChannel`, which is fine, except that it assumes that handlerAdded has already been called on Http2MultiplexCodec . . . which is always true except when we're doing a cleartext upgrade.

### Steps to reproduce

Make an HttpServerUpgradeHandler, with an Http2MultiplexCodec.  Send an h2c upgrade request.

### Minimal yet complete reproducer code (or URL to code)

Try to upgrade this to netty 4.1.15.

https://github.com/twitter/finagle/blob/develop/finagle-http2/src/main/scala/com/twitter/finagle/http2/Http2CleartextServerInitializer.scala#L74

### Netty version
4.1.15.Final

mosesn:
Digging a little deeper, the exactly problem here is because we call `prepareUpgradeResponse` before we call `upgradeTo` in Http2ServerUpgradeCodec.

mosesn:
I was able to get around this problem by setting two private fields in HttpMultiplexCodec (both named ctx).  @normanmaurer if I merge a fix into netty, what's the ETA on a release?

normanmaurer:
@mosesn end of next week sounds doable... so yeah please provide a pr if you already have one

mosesn:
@normanmaurer I don't have one yet, but I'm happy to take a stab today.  Where do you think we should fix this?  HttpServerUpgradeHandler seems pretty stable (hasn't been changed in a year) so I'm more inclined to do it in Http2ServerUpgradeCodec.

Looks like maybe we could try dumping all the code from prepareUpgradeResponse into upgradeTo?

https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ServerUpgradeCodec.java#L132

What do you think?

normanmaurer:
@mosesn sounds reasonable to me.

normanmaurer:
@mosesn actually I think we need another fix... let me come back to you asap.

normanmaurer:
[...]

PR #7184, opened 2017-09-06
Fix NPE when using Http2ServerUpgradeCodec with Http2FrameCodec and H…

normanmaurer:
Motiviation:

At the moment an NPE is thrown if someone tries to use the Http2ServerUpgradeCodec with Http2FrameCodec and Http2MultiplexCodec.

Modifications:

- Ensure the handler was added to the pipeline before calling on*Upgrade(...) methods.
- Add tests
- Fix adding of handlers after upgrade.

Result:

Fixes [#7173].

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
@@ -156,8 +156,13 @@ public class Http2ConnectionHandler extends ByteToMessageDecoder implements Http
         if (connection().isServer()) {
             throw connectionError(PROTOCOL_ERROR, "Client-side HTTP upgrade requested for a server");
         }
-        if (prefaceSent() || decoder.prefaceReceived()) {
-            throw connectionError(PROTOCOL_ERROR, "HTTP upgrade must occur before HTTP/2 preface is sent or received");
+        if (!prefaceSent()) {
+            // If the preface was not sent yet it most likely means the handler was not added to the pipeline before
+            // calling this method.
+            throw connectionError(INTERNAL_ERROR, "HTTP upgrade must occur after preface was sent");
+        }
+        if (decoder.prefaceReceived()) {
+            throw connectionError(PROTOCOL_ERROR, "HTTP upgrade must occur before HTTP/2 preface is received");
         }
 
         // Create a local stream used for the HTTP cleartext upgrade.
@@ -172,8 +177,13 @@ public class Http2ConnectionHandler extends ByteToMessageDecoder implements Http
         if (!connection().isServer()) {
             throw connectionError(PROTOCOL_ERROR, "Server-side HTTP upgrade requested for a client");
         }
-        if (prefaceSent() || decoder.prefaceReceived()) {
-            throw connectionError(PROTOCOL_ERROR, "HTTP upgrade must occur before HTTP/2 preface is sent or received");
+        if (!prefaceSent()) {
+            // If the preface was not sent yet it most likely means the handler was not added to the pipeline before
+            // calling this method.
+            throw connectionError(INTERNAL_ERROR, "HTTP upgrade must occur after preface was sent");
+        }
+        if (decoder.prefaceReceived()) {
+            throw connectionError(PROTOCOL_ERROR, "HTTP upgrade must occur before HTTP/2 preface is received");
         }
 
         // Apply the settings but no ACK is necessary.
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ServerUpgradeCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ServerUpgradeCodec.java
@@ -49,11 +49,14 @@ public class Http2ServerUpgradeCodec implements HttpServerUpgradeHandler.Upgrade
     private static final InternalLogger logger = InternalLoggerFactory.getInstance(Http2ServerUpgradeCodec.class);
     private static final List<CharSequence> REQUIRED_UPGRADE_HEADERS =
             Collections.singletonList(HTTP_UPGRADE_SETTINGS_HEADER);
+    private static final ChannelHandler[] EMPTY_HANDLERS = new ChannelHandler[0];
 
     private final String handlerName;
     private final Http2ConnectionHandler connectionHandler;
-    private final ChannelHandler upgradeToHandler;
+    private final ChannelHandler[] handlers;
     private final Http2FrameReader frameReader;
+
+    private Http2Settings settings;
 
     /**
      * Creates the codec using a default name for the connection handler when adding to the
@@ -62,7 +65,7 @@ public class Http2ServerUpgradeCodec implements HttpServerUpgradeHandler.Upgrade
      * @param connectionHandler the HTTP/2 connection handler
      */
     public Http2ServerUpgradeCodec(Http2ConnectionHandler connectionHandler) {
-        this(null, connectionHandler);
+        this(null, connectionHandler, EMPTY_HANDLERS);
     }
 
     /**
@@ -72,7 +75,7 @@ public class Http2ServerUpgradeCodec implements HttpServerUpgradeHandler.Upgrade
      * @param http2Codec the HTTP/2 multiplexing handler.
      */
     public Http2ServerUpgradeCodec(Http2MultiplexCodec http2Codec) {
-        this(null, http2Codec);
+        this(null, http2Codec, EMPTY_HANDLERS);
     }
 
     /**
@@ -83,7 +86,7 @@ public class Http2ServerUpgradeCodec implements HttpServerUpgradeHandler.Upgrade
      * @param connectionHandler the HTTP/2 connection handler
      */
     public Http2ServerUpgradeCodec(String handlerName, Http2ConnectionHandler connectionHandler) {
-        this(handlerName, connectionHandler, connectionHandler);
+        this(handlerName, connectionHandler, EMPTY_HANDLERS);
     }
 
     /**
@@ -93,7 +96,7 @@ public class Http2ServerUpgradeCodec implements HttpServerUpgradeHandler.Upgrade
      * @param http2Codec the HTTP/2 multiplexing handler.
      */
     public Http2ServerUpgradeCodec(String handlerName, Http2MultiplexCodec http2Codec) {
-        this(handlerName, http2Codec, http2Codec);
+        this(handlerName, http2Codec, EMPTY_HANDLERS);
     }
 
     /**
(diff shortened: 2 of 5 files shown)


normanmaurer:
@mosesn PTAL

codec-http2/src/main/java/io/netty/handler/codec/http2/AbstractHttp2ConnectionHandlerBuilder.java
```diff
@@ -106,6 +104,24 @@
     private Boolean encoderEnforceMaxConcurrentStreams;
     private Boolean encoderIgnoreMaxHeaderListSize;
     private int initialHuffmanDecodeCapacity = DEFAULT_INITIAL_HUFFMAN_DECODE_CAPACITY;
+    private boolean sendPreface = true;
+
+    /**
+     * Returns {@code true} if the preface should be send once the channel become active. If you want to do a clear-
```
mosesn: s/send/sent

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ServerUpgradeCodec.java
```diff
@@ -151,8 +148,23 @@ public boolean prepareUpgradeResponse(ChannelHandlerContext ctx, FullHttpRequest
...
+
+        } catch (Http2Exception e) {
+            ctx.fireExceptionCaught(e);
+            ctx.close();
+            return;
+        }
+
+        if (handlers != null) {
+            final String name = ctx.pipeline().context(connectionHandler).name();
+            for (int i = handlers.length - 1; i >= 0; -- i) {
```
mosesn: ffti: s/-- i/i--

normanmaurer:
Doh! You are right let me fix it

normanmaurer:
@mosesn I think its correct now... Also adjusted tests to ensure we really send the preface etc.

mosesn:
LGTM @normanmaurer!

normanmaurer:
@ejona86 ping

normanmaurer:
Also @carl-mastrangelo

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ServerUpgradeCodec.java
```diff
@@ -151,8 +148,23 @@ public boolean prepareUpgradeResponse(ChannelHandlerContext ctx, FullHttpRequest
...
+        } catch (Http2Exception e) {
+            ctx.fireExceptionCaught(e);
+            ctx.close();
+            return;
+        }
+
+        if (handlers != null) {
+            final String name = ctx.pipeline().context(connectionHandler).name();
+            for (int i = handlers.length - 1; i >= 0; i--) {
+                ctx.pipeline().addAfter(name, null, handlers[i]);
```
carl-mastrangelo: can upgradeTo be called multiple times?  If not, does the codec live possibly a long time?  If both of those things, then should handlers be null'ed out after this call?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ServerUpgradeCodec.java
```diff
@@ -151,8 +148,23 @@ public boolean prepareUpgradeResponse(ChannelHandlerContext ctx, FullHttpRequest
...
+            connectionHandler.onHttpServerUpgrade(settings);
+
+        } catch (Http2Exception e) {
+            ctx.fireExceptionCaught(e);
+            ctx.close();
+            return;
+        }
+
+        if (handlers != null) {
+            final String name = ctx.pipeline().context(connectionHandler).name();
```
carl-mastrangelo: Can handlers be looked up by identity?  Also, what if there are multiple handlers with the same name in the pipeline (i.e. an H2 on top of H2 channel would be present twice in the pipe).

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2ServerUpgradeCodecTest.java
```diff
@@ -0,0 +1,83 @@
...
+        testUpgrade(new Http2FrameCodecBuilder(true).httpClearTextUpgrade(true).build());
+    }
+
+    @Test
+    public void testUpgradeToHttp2MultiplexCodec() {
+        testUpgrade(new Http2MultiplexCodecBuilder(true, new HttpInboundHandler()).httpClearTextUpgrade(true).build());
+    }
+
+    private static void testUpgrade(Http2ConnectionHandler handler) {
+        FullHttpRequest request = new DefaultFullHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.GET, "/");
```
carl-mastrangelo: Optional: if you treat you test code as examples of how to use the SUT, consider using the http OPTIONS method, and the `*` path.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ServerUpgradeCodec.java
normanmaurer: I think this should be fine as we lookup by instance... I am missing anything ?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -109,6 +110,14 @@ protected Http2ConnectionHandler(Http2ConnectionDecoder decoder, Http2Connection
         decoder = new DefaultHttp2ConnectionDecoder(connection, encoder, frameReader);
     }
 
+    void sendPrefaceOnHandlerAdded(boolean sendPrefaceOnHandlerAdded) {
```
ejona86: This isn't really clear from a semantic perspective. Does it mean that the preface won't be sent or that it will be sent earlier or later? `httpClearTextUpgrade` is clearer from that perspective, as the preface will be sent from `onHttpServerUpgrade()`. That's very non-obvious given its current name.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -162,6 +171,7 @@ public void onHttpClientUpgrade() throws Http2Exception {
 
         // Create a local stream used for the HTTP cleartext upgrade.
         connection().local().createStream(HTTP_UPGRADE_STREAM_ID, true);
+        sendPreface();
```
ejona86: Client doesn't send preface in this way for Upgrade. Instead, the SETTINGS goes in the initial HTTP/1 Headers that contained the Upgrade request. This works because you can send redundant SETTINGS at any time, but is misleading and seems like it could simply be removed.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
normanmaurer: @ejona86 so you suggest changing the method name ? Any good idea ?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
normanmaurer: true... So yeah I think we only need this for servers.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
ejona86: `httpClearTextUpgrade` seemed strictly better, given the code is already aware of `onHttpServerUpgrade()` and that's what it's called on the builder.

normanmaurer:
@ejona86 @carl-mastrangelo addressed comments...

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
mosesn: Maybe I'm interpreting the spec differently, but I think clients do need to send the preface this way.

https://tools.ietf.org/html/rfc7540#section-3.2

```
   The first HTTP/2 frame sent by the server MUST be a server connection
   preface (Section 3.5) consisting of a SETTINGS frame (Section 6.5).
   Upon receiving the 101 response, the client MUST send a connection
   preface (Section 3.5), which includes a SETTINGS frame.
```

ejona86: @mosesn, you're right. That's weird given the `HTTP2-Settings` header but so be it.

Sorry @normanmaurer!

ejona86:
@normanmaurer, so having some more time to look at this and in light of @mosesn's discovery, why not invert the `prefaceSent()` check (to assert that the preface was sent) in `onHttp2ServerUpgrade()` and `onHttp2ClientUpgrade()` so that we make use of the `handlerAdded()`? That removes the need for the `httpClearTextUpgrade` configuration and reduces most of the special-caseness.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ServerUpgradeCodec.java
carl-mastrangelo: Nope, I was thinking of a different API.  I thought the name was derived from the class

normanmaurer:
@ejona86 thats a great idea! Let me do this..

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -165,8 +156,11 @@ public void onHttpClientUpgrade() throws Http2Exception {
         if (connection().isServer()) {
             throw connectionError(PROTOCOL_ERROR, "Client-side HTTP upgrade requested for a server");
         }
-        if (prefaceSent() || decoder.prefaceReceived()) {
-            throw connectionError(PROTOCOL_ERROR, "HTTP upgrade must occur before HTTP/2 preface is sent or received");
+        if (!prefaceSent()) {
+            throw connectionError(PROTOCOL_ERROR, "HTTP upgrade must occur after preface was sent");
```
mosesn: this seems backwards according to the spec, since you don't send the preface until you know the upgrade has succeeded.  can we add a note that this is an implementation detail in netty (maybe a comment?)

ejona86: Should this be INTERNAL instead? I agree it should be made more obvious to be a Netty-internal.

normanmaurer: yeah sounds good.

normanmaurer:
@ejona86 @mosesn adjusted error, added two more tests and added a comment. Will pull in once CI completes

ejona86:
The INTERNAL+comment LGTM

normanmaurer:
Cherry-picked into 4.1 (15611dadbbf8f8c217c03ce33da89beb54602023)

PR #7187, opened 2017-09-06
Remove double comparing of content out of the DefaultHttp2GoAwayFrame…

fenik17:
….equals()

Motivation:
In `DefaultHttp2GoAwayFrame.equals()` a content compared twice: explicitly and in the `super` method.

Modifications:
Remove explicit content comparision.
Make `hashCode()` consistent with `equals()`.

Result:
A `DefaultHttp2GoAwayFrame.equals()` work faster.


normanmaurer:
Cherry-picked into 4.1 (e50075508611e3cdabf60f37d9ba52f0be46c414)

PR #7188, opened 2017-09-06
Optimize DefaultChannelId.equals

fenik17:
Motivation:
A `DefaultChannelId` has final `hashCode` field calculated in the constructor. We can use it in `equals` to the fast return for different objects.

Modifications:
Use `hashCode` field in `DefaultChannelId.equals()`.

Result:
Fast `equals` on negative scenarios.

### DIFF
--- a/transport/src/main/java/io/netty/channel/DefaultChannelId.java
+++ b/transport/src/main/java/io/netty/channel/DefaultChannelId.java
@@ -274,7 +274,14 @@ public final class DefaultChannelId implements ChannelId {
 
     @Override
     public boolean equals(Object obj) {
-        return this == obj || (obj instanceof DefaultChannelId && Arrays.equals(data, ((DefaultChannelId) obj).data));
+        if (this == obj) {
+            return true;
+        }
+        if (!(obj instanceof DefaultChannelId)) {
+            return false;
+        }
+        DefaultChannelId other = (DefaultChannelId) obj;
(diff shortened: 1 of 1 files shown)


normanmaurer:
@fenik17 can you split this into two prs as these are different issues ?

fenik17:
@normanmaurer sure. and one #7193

transport/src/main/java/io/netty/channel/DefaultChannelId.java
```diff
@@ -274,7 +274,14 @@ public int compareTo(final ChannelId o) {
...
     public boolean equals(Object obj) {
-        return this == obj || (obj instanceof DefaultChannelId && Arrays.equals(data, ((DefaultChannelId) obj).data));
+        if (this == obj) {
+            return true;
+        }
+        if (!(obj instanceof DefaultChannelId)) {
+            return false;
+        }
+        DefaultChannelId other = (DefaultChannelId) obj;
+        return hashCode == other.hashCode && Arrays.equals(data, other.data);
```
johnou: Short circuit ftw

normanmaurer: <3

normanmaurer:
Cherry-picked into 4.1 (e40469057483c6b3ce9cf69af12dd19a68392cfd). @fenik17 thanks!

PR #7191, opened 2017-09-07
Print out the actual cause when an assertion failure happens during D…

normanmaurer:
…atagramUnicastTest.testSimpleSendWithConnect

Motivation:

We recently saw an assertion failure when running DatagramUnicastTest.testSimpleSendWithConnect.

Modifications:

- Adding more debug infos
- Ensure we always correctly release the buffers.

Result:

More informations when tests fail.


normanmaurer:
Let me pull this in so I will get more details on failure.

normanmaurer:
Cherry-picked into 4.1 (c0396818cafc4d3b6a5c8230b9f91b180e813645)

PR #7193, opened 2017-09-07
Fix hashCode() in Http2StreamChannelId

fenik17:
Motivation:
In `Http2StreamChannelId` a `hashCode()` is not consistent with `equals()`.

Modifications:
Make a `Http2StreamChannelId.hashCode()` consistent with `equals()`.

Result:
Faster hash map's operations where the `Http2StreamChannelId` as keys.


normanmaurer:
Cherry-picked into 4.1 (de9e666493469d40fa9107a78f9ae421a05f9cf4).

@fenik17 thanks!

PR #7198, opened 2017-09-08
Addition of a new jSerialComm transport

Ziver:
Motivation:

This is an alternative to the Rxtx transport.
One of the main difference is that there is no
native lib installation requirement with jSerialComm
as it handles the native dependencies automatically.
Which makes the deployment/installation less complicated.

Modifications:

Addition of a new transport and associated example code.
A new depndency to jSerialComm

Result:

A new transport is available.

### DIFF
--- a/example/pom.xml
+++ b/example/pom.xml
@@ -116,6 +116,11 @@
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
+      <artifactId>netty-transport-jserialcomm</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
       <artifactId>netty-transport-udt</artifactId>
       <version>${project.version}</version>
     </dependency>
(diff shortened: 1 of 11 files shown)


normanmaurer:
@Ziver first thanks for this PR this looks great. That said I am not sure this should be part of Netty itself and not just another project that implements a custom netty transport. That is because if we merge this we will need to maintain it while not have any possibility to even test the code. This also reminds me that I wanted to mark the serial transport as `@deprecated` as there we basically have the same problem. 

I think transports that ship as part of netty should be easy for us to test and to maintain.

normanmaurer:
See https://github.com/netty/netty/pull/7200

Ziver:
@normanmaurer that makes sense. I'll put it in its own repo.

Here is the link to the new repo for anyone that would be interested:
https://github.com/Ziver/Netty-Transport-jSerialComm

normanmaurer:
@Ziver maybe it also makes sense to add it here: 

http://netty.io/wiki/related-projects.html

normanmaurer:
(you can do this by yourself)

PR #7200, opened 2017-09-10
Mark transport-rxtx as @deprecated

normanmaurer:
Motivation:

transport-rxtx has no tests and there is really no easy way to add some. Beside this this transport is not really well maintained.

Modifications:

Mark transport-rxtx as @deprecated so we can drop it in next major version.

Result:

Notify users of plan to drop the transport.

### DIFF
--- a/transport-rxtx/src/main/java/io/netty/channel/rxtx/package-info.java
+++ b/transport-rxtx/src/main/java/io/netty/channel/rxtx/package-info.java
@@ -16,5 +16,8 @@
 
 /**
  * A serial and parallel port communication transport based on <a href="http://rxtx.qbang.org/">RXTX</a>.
+ *
+ * @deprecated this transport will be removed in the next major version.
  */
+@Deprecated
 package io.netty.channel.rxtx;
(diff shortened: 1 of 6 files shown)


normanmaurer:
@Scottmitch you are cool with this as well ? I suspect yes ;)

normanmaurer:
Cherry-picked into 4.1 (8c8779669e1acdd8307e5faafb7ca9b378348831) and 4.0 (791ce25d760762c2d87b7765fe7c55344169b93a)

Scottmitch:
lgtm

Scottmitch:
@normanmaurer - for the next major release we should consider start using an "incubator" for new protocol implementations. Once they reach a certain maturity, and have a dedicated developer, then we can promote them into the main project.

transport-rxtx/src/main/java/io/netty/channel/rxtx/package-info.java
```diff
@@ -16,5 +16,8 @@
 
 /**
  * A serial and parallel port communication transport based on <a href="http://rxtx.qbang.org/">RXTX</a>.
+ *
+ * @deprecated this transport will be removed in the next major version.
  */
+@Deprecated
```
fenik17: Compiler in IntelliJ IDEA still is not allow `@Deprecated` in package-info's:
> Error:(23, 1) java: modifier deprecated not allowed here

normanmaurer: Fixed it... ff592c8d5c54d6a81b17050f980c5d57299a5148

normanmaurer:
@Scottmitch thats a good idea.

PR #7201, opened 2017-09-11
Use the correct osname in the Bundle-NativeCode declaration.

normanmaurer:
Motivation:

We need to ensure we use the correct osname in the Bundle-NativeCode declaration as declared in:

https://www.osgi.org/developer/specifications/reference/

Modifications:

Update osname to match the spec.

Result:

Correct Bundle-NativeCode entry in the MANIFEST

### DIFF
--- a/transport-native-kqueue/pom.xml
+++ b/transport-native-kqueue/pom.xml
@@ -110,7 +110,7 @@
                       <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
                     </manifest>
                     <manifestEntries>
-                      <Bundle-NativeCode>META-INF/native/libnetty_transport_native_kqueue_${os.detected.arch}.jnilib; osname=darwin, processor=${os.detected.arch}"</Bundle-NativeCode>
+                      <Bundle-NativeCode>META-INF/native/libnetty_transport_native_kqueue_${os.detected.arch}.jnilib; osname=MacOSX, processor=${os.detected.arch}"</Bundle-NativeCode>
                     </manifestEntries>
                     <index>true</index>
                     <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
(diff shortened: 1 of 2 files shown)


normanmaurer:
PTAL... noticed this while looking for something.

https://www.osgi.org/developer/specifications/reference/

transport-native-kqueue/pom.xml
```diff
@@ -110,7 +110,7 @@
                       <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
                     </manifest>
                     <manifestEntries>
-                      <Bundle-NativeCode>META-INF/native/libnetty_transport_native_kqueue_${os.detected.arch}.jnilib; osname=darwin, processor=${os.detected.arch}"</Bundle-NativeCode>
+                      <Bundle-NativeCode>META-INF/native/libnetty_transport_native_kqueue_${os.detected.arch}.jnilib; osname=MacOS, processor=${os.detected.arch}"</Bundle-NativeCode>
```
ejona86: MacOSX

normanmaurer: @ejona86 isn't both valid ?
https://www.osgi.org/developer/specifications/reference/#ReferenceOS

normanmaurer:
@carl-mastrangelo yes...

transport-native-kqueue/pom.xml
ejona86: It looks like MacOS was like OS 9, the thing that is not based on Unix. I'd have expected them to be aliases if they referred to the same thing, not separate rows.

normanmaurer: hmm.. ok let me use macosx just to be safe :)

normanmaurer:
Cherry-picked into 4.1 (426938307ddcbd6f2db3e86afe6390b6fbe74918) and 4.0 (23dc34effc73771a246a09cacefa2ec77157164a)

PR #7202, opened 2017-09-13
Use offset finding eol avoid repeated scaning.

MengJie:
Motivation:

A large frame will be componsed by many packages. Every time the package
arrived, findEndOfLine will be called from the start of the buffer. It
will cause the complexity of reading frame equal to  O(n^2). This can be
eliminated by using a offset to mark the last scan position, when new
package arrived, just find the delimter from the mark. The complexity
will be O(n).

Modification:
Add a offset to mark the last scan position.

Result:
Better performance for read large frame.

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/LineBasedFrameDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/LineBasedFrameDecoder.java
@@ -71,7 +71,6 @@ public class LineBasedFrameDecoder extends ByteToMessageDecoder {
         this.maxLength = maxLength;
         this.failFast = failFast;
         this.stripDelimiter = stripDelimiter;
-        this.offset = 0;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


codec/src/main/java/io/netty/handler/codec/LineBasedFrameDecoder.java
```diff
@@ -68,6 +71,7 @@ public LineBasedFrameDecoder(final int maxLength, final boolean stripDelimiter,
         this.maxLength = maxLength;
         this.failFast = failFast;
         this.stripDelimiter = stripDelimiter;
+        this.offset = 0;
```
normanmaurer: this is not needed as `offset` will be 0

normanmaurer:
@MengJie can you please sign the ICLA and let me know once done:

https://netty.io/s/icla

normanmaurer:
@MengJie seems like something is wrong with your PR as it produces test errors:

https://garage.netty.io/teamcity/viewLog.html?buildId=24306&buildTypeId=pulls_netty&guest=1

normanmaurer:
Cherry-picked into 4.1 (80b8a91b704c6418581a9d540998d7f4446faaef) and 4.0 (70684c2247bc3a0cfcf44ad339509180ddd54738)

normanmaurer:
@MengJie thanks!

MengJie:
@normanmaurer thanks!

Issue #7205, opened 2017-09-13
MQTT encode doesn't complain if password is set but username not


ppatierno:
### Expected behavior

The MQTT decoder should raise an exception trying to build a CONNECT packet where password field is set but not the username one (as by MQTT 3.1/3.1.1 spec).

### Actual behavior

The CONNECT packet is built and hasUsername and hasPassword aren't handled in the right way.

normanmaurer:
Fixed by https://github.com/netty/netty/pull/7206#pullrequestreview-62812166

PR #7206, opened 2017-09-13
[#7205] Check on wrong CONNECT packet with password but without username

ppatierno:
Added checking on password set but username not in the CONNECT encoding method

Changed the way how hasUsername and hasPassword are set in the CONNECT packet
Added unit test with a wrong CONNECT packet having password set but not username

Motivation:

Modification:

Result:

Fixes #<GitHub issue number>. 

### DIFF
--- a/codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttEncoder.java
+++ b/codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttEncoder.java
@@ -20,6 +20,7 @@ import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.DecoderException;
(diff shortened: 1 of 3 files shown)


normanmaurer:
@ppatierno please fix the check-style error. After this LGTM:
https://garage.netty.io/teamcity/viewLog.html?buildId=24245&buildTypeId=pulls_netty&tab=buildLog

ppatierno:
@normanmaurer on the link you provided me I only see that there is an error about checkstyle but not what the error is. Running the `mvn checkstyle:checkstyle` command locally, I don't receive any error. Any thoughts ?

ppatierno:
@normanmaurer it happens running `mvn test` (?) btw found it, the fix is on its way ;)

ppatierno:
@normanmaurer done !

normanmaurer:
Squashes and adjusted commit message and cherry-picked into 4.1 (901c66fa813a759c9dbfe64eed10ba4d6095d4ea).

normanmaurer:
@ppatierno thanks

PR #7212, opened 2017-09-15
Remove allocation from ResourceLeakDetector

carl-mastrangelo:
Motivation:
RLD allocates an ArrayDeque in anticipation of recording access
points.  If the leak detection level is less than ADVANCED though,
the dequeue is never used.  Since SIMPLE is the default level,
there is a minor perf win to not preemptively allocate it.

This showed up in garbage profiling when creation a high number of
buffers.

Modifications:
Only allocate the dequeue if it will be used.

Result:
Less garbage created.

### DIFF
--- a/common/src/main/java/io/netty/util/ResourceLeakDetector.java
+++ b/common/src/main/java/io/netty/util/ResourceLeakDetector.java
@@ -27,6 +27,7 @@ import java.util.ArrayDeque;
 import java.util.Deque;
 import java.util.concurrent.ConcurrentMap;
 
+import static io.netty.util.internal.EmptyArrays.EMPTY_OBJECTS;
 import static io.netty.util.internal.StringUtil.EMPTY_STRING;
 import static io.netty.util.internal.StringUtil.NEWLINE;
 import static io.netty.util.internal.StringUtil.simpleClassName;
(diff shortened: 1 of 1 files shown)


common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -407,9 +408,14 @@ public String toString() {
...
-                array = lastRecords.toArray();
-                removedRecords = this.removedRecords;
+            if (lastRecords != null) {
+                synchronized (lastRecords) {
+                    array = lastRecords.toArray();
+                    removedRecords = this.removedRecords;
+                }
+            } else {
+                removedRecords = 0;
+                array = new Object[0];
```
normanmaurer: @carl-mastrangelo use `EmptyArrays.EMPTY_OBJECTS` to reduce allocations ;)

carl-mastrangelo: Done.

normanmaurer:
Cherry-picked into 4.1 (b32cd26a961c71cc5ad0e9f88fa8c15fbfd89930) and 4.0 (9df8082c39050f206780ab81e96624e3801f418f)

normanmaurer:
@carl-mastrangelo thanks!

PR #7213, opened 2017-09-15
NativeLibraryLoader logging clarify

Scottmitch:
Motivation:
NativeLibraryLoader only logs a debug statement if the library is successfully loaded from java.library.path, but will load failure statements the load for java.library.path which can mislead users to believe the load actually failed when it may have succeeded.

Modifications:
- Always load a debug statement when a library was successfully loaded

Result:
NativeLibraryLoader log statements more clear.


normanmaurer:
@Scottmitch LGTM after you fixed the typo in the commit message:

```
 but will load
```

should be:

```
 but will log
```

Scottmitch:
4.1 (b1332bf12e) 4.0 (74da4c54bc)

PR #7217, opened 2017-09-16
Reduce performance overhead of ResourceLeakDetector

normanmaurer:
Motiviation:

The ResourceLeakDetector helps to detect and troubleshoot resource leaks and is often used even in production enviroments with a low level. Because of this its import that we try to keep the overhead as low as overhead. Most of the times no leak is detected (as all is correctly handled) so we should keep the overhead for this case as low as possible.

Modifications:

- Only call getStackTrace() if a leak is reported as it is a very expensive native call. Also handle the filtering and creating of the String in a lazy fashion
- Remove the need to mantain a Queue to store the last access records
- Add benchmark

Result:

Huge decrease of performance overhead.

Before the patch:

```
Benchmark                                           (recordTimes)   Mode  Cnt     Score     Error  Units
ResourceLeakDetectorRecordBenchmark.record                      8  thrpt   20  4358.367 ± 116.419  ops/s
ResourceLeakDetectorRecordBenchmark.record                     16  thrpt   20  2306.027 ±  55.044  ops/s
ResourceLeakDetectorRecordBenchmark.recordWithHint              8  thrpt   20  4220.979 ± 114.046  ops/s
ResourceLeakDetectorRecordBenchmark.recordWithHint             16  thrpt   20  2250.734 ±  55.352  ops/s
```

With this patch:

```
Benchmark                                           (recordTimes)   Mode  Cnt      Score      Error  Units
ResourceLeakDetectorRecordBenchmark.record                      8  thrpt   20  71398.957 ± 2695.925  ops/s
ResourceLeakDetectorRecordBenchmark.record                     16  thrpt   20  38643.963 ± 1446.694  ops/s
ResourceLeakDetectorRecordBenchmark.recordWithHint              8  thrpt   20  71677.882 ± 2923.622  ops/s
ResourceLeakDetectorRecordBenchmark.recordWithHint             16  thrpt   20  38660.176 ± 1467.732  ops/s
```

### DIFF
--- a/common/src/main/java/io/netty/util/ResourceLeakDetector.java
+++ b/common/src/main/java/io/netty/util/ResourceLeakDetector.java
@@ -23,11 +23,10 @@ import io.netty.util.internal.logging.InternalLoggerFactory;
 
 import java.lang.ref.PhantomReference;
 import java.lang.ref.ReferenceQueue;
-import java.util.ArrayDeque;
-import java.util.Deque;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.concurrent.ConcurrentMap;
 
-import static io.netty.util.internal.EmptyArrays.EMPTY_OBJECTS;
 import static io.netty.util.internal.StringUtil.EMPTY_STRING;
 import static io.netty.util.internal.StringUtil.NEWLINE;
 import static io.netty.util.internal.StringUtil.simpleClassName;
@@ -40,7 +39,10 @@ public class ResourceLeakDetector<T> {
 
     private static final String PROP_MAX_RECORDS = "io.netty.leakDetection.maxRecords";
     private static final int DEFAULT_MAX_RECORDS = 4;
+    private static final String PROP_MAX_SAMPLED_RECORDS = "io.netty.leakDetection.maxSampledRecords";
+
     private static final int MAX_RECORDS;
+    private static final int MAX_SAMPLED_RECORDS;
 
     /**
      * Represents the level of resource leak detection.
@@ -109,11 +111,14 @@ public class ResourceLeakDetector<T> {
         Level level = Level.parseLevel(levelStr);
 
         MAX_RECORDS = SystemPropertyUtil.getInt(PROP_MAX_RECORDS, DEFAULT_MAX_RECORDS);
+        long maxRecordsSampled = SystemPropertyUtil.getLong(PROP_MAX_SAMPLED_RECORDS, MAX_RECORDS * 10L);
+        MAX_SAMPLED_RECORDS = Math.max((int) Math.min(Integer.MAX_VALUE, maxRecordsSampled), MAX_RECORDS);
 
         ResourceLeakDetector.level = level;
         if (logger.isDebugEnabled()) {
             logger.debug("-D{}: {}", PROP_LEVEL, level.name().toLowerCase());
             logger.debug("-D{}: {}", PROP_MAX_RECORDS, MAX_RECORDS);
+            logger.debug("-D{}: {}", PROP_MAX_SAMPLED_RECORDS, MAX_SAMPLED_RECORDS);
         }
     }
 
@@ -245,27 +250,29 @@ public class ResourceLeakDetector<T> {
 
         if (level.ordinal() < Level.PARANOID.ordinal()) {
             if ((PlatformDependent.threadLocalRandom().nextInt(samplingInterval)) == 0) {
-                reportLeak(level);
+                reportLeak();
                 return new DefaultResourceLeak(obj);
-            } else {
-                return null;
             }
-        } else {
-            reportLeak(level);
-            return new DefaultResourceLeak(obj);
+            return null;
+        }
+        reportLeak();
+        return new DefaultResourceLeak(obj);
+    }
+
+    private void clearRefQueue() {
+        for (;;) {
+            @SuppressWarnings("unchecked")
+            DefaultResourceLeak ref = (DefaultResourceLeak) refQueue.poll();
+            if (ref == null) {
+                break;
+            }
+            ref.close();
         }
     }
 
-    private void reportLeak(Level level) {
+    private void reportLeak() {
         if (!logger.isErrorEnabled()) {
-            for (;;) {
-                @SuppressWarnings("unchecked")
-                DefaultResourceLeak ref = (DefaultResourceLeak) refQueue.poll();
-                if (ref == null) {
-                    break;
-                }
-                ref.close();
-            }
+            clearRefQueue();
             return;
         }
 
@@ -328,11 +335,14 @@ public class ResourceLeakDetector<T> {
     @SuppressWarnings("deprecation")
     private final class DefaultResourceLeak extends PhantomReference<Object> implements ResourceLeakTracker<T>,
             ResourceLeak {
-        private final String creationRecord;
-        private final Deque<String> lastRecords;
-        private final int trackedHash;
+        private final Record head;
 
-        private int removedRecords;
+        // This will be updated once a new Record will be created and added
+        private Record tail;
+
+        private final int trackedHash;
+        private int numRecords;
+        private int droppedRecords;
 
         DefaultResourceLeak(Object referent) {
             super(referent, refQueue);
@@ -343,41 +353,42 @@ public class ResourceLeakDetector<T> {
             // It's important that we not store a reference to the referent as this would disallow it from
             // be collected via the PhantomReference.
             trackedHash = System.identityHashCode(referent);
-
-            Level level = getLevel();
-            if (level.ordinal() >= Level.ADVANCED.ordinal()) {
-                creationRecord = newRecord(null, 3);
-                lastRecords = new ArrayDeque<String>();
-            } else {
-                creationRecord = null;
-                lastRecords = null;
-            }
+            head = tail = getLevel().ordinal() >= Level.ADVANCED.ordinal() ? new Record() : null;
             allLeaks.put(this, LeakEntry.INSTANCE);
         }
 
         @Override
         public void record() {
-            record0(null, 3);
+            record0(null);
         }
 
         @Override
         public void record(Object hint) {
-            record0(hint, 3);
+            record0(hint);
         }
 
-        private void record0(Object hint, int recordsToSkip) {
+        private void record0(Object hint) {
             // Check MAX_RECORDS > 0 here to avoid similar check before remove from and add to lastRecords
-            if (creationRecord != null && MAX_RECORDS > 0) {
-                String value = newRecord(hint, recordsToSkip);
+            if (head != null && MAX_RECORDS > 0) {
+                synchronized (head) {
+                    if (tail == null) {
+                        // already closed
+                        return;
+                    }
 
-                synchronized (lastRecords) {
-                    int size = lastRecords.size();
-                    if (size == 0 || !lastRecords.getLast().equals(value)) {
-                        if (size >= MAX_RECORDS) {
-                            lastRecords.removeFirst();
-                            ++removedRecords;
-                        }
-                        lastRecords.add(value);
+                    Record record = hint == null ? new Record() : new Record(hint);
+                    tail.next = record;
+                    tail = record;
+
+                    // Enforce a limit so our linked-list not grows too large and cause a GC storm later on when we
+                    // unlink it. The reason why we choose a different limit to MAX_RECORDS is that we will not handle
+                    // duplications here as its very expensive. We will filter these out when we actually
+                    // detected a leak.
+                    if (numRecords == MAX_SAMPLED_RECORDS) {
+                        head.next = head.next.next;
+                        droppedRecords++;
+                    } else {
+                        numRecords++;
                     }
                 }
             }
@@ -386,7 +397,21 @@ public class ResourceLeakDetector<T> {
         @Override
         public boolean close() {
             // Use the ConcurrentMap remove method, which avoids allocating an iterator.
-            return allLeaks.remove(this, LeakEntry.INSTANCE);
+            if (allLeaks.remove(this, LeakEntry.INSTANCE)) {
+                // Call clear so the reference is not even enqueued.
+                clear();
+
+                if (head != null) {
+                    synchronized (head) {
+                        // Allow to GC all the records.
+                        head.next = null;
+                        numRecords = 0;
+                        tail = null;
+                    }
+                }
+                return true;
+            }
+            return false;
         }
 
         @Override
@@ -403,39 +428,64 @@ public class ResourceLeakDetector<T> {
 
         @Override
         public String toString() {
-            if (creationRecord == null) {
+            if (head == null) {
                 return EMPTY_STRING;
             }
 
-            final Object[] array;
-            final int removedRecords;
-            if (lastRecords != null) {
-                synchronized (lastRecords) {
-                    array = lastRecords.toArray();
-                    removedRecords = this.removedRecords;
+            final String creationRecord;
+            final String[] array;
+            int idx = 0;
+            String last = null;
+            final int dropped;
+
+            synchronized (head) {
+                if (tail == null) {
+                    // Already closed
+                    return EMPTY_STRING;
                 }
-            } else {
-                removedRecords = 0;
-                array = EMPTY_OBJECTS;
+                dropped = droppedRecords;
+                creationRecord = head.toString();
+                array = new String[numRecords];
+                Record record = head.next;
+                while (record != null) {
+                    String recordStr = record.toString();
+                    if (last == null || !last.equals(recordStr)) {
+                        array[idx ++] = recordStr;
+                        last = recordStr;
+                    }
+                    record = record.next;
+                }
             }
+
+            int removed = idx > MAX_RECORDS ? idx - MAX_RECORDS : 0;
 
             StringBuilder buf = new StringBuilder(16384).append(NEWLINE);
-            if (removedRecords > 0) {
+            if (removed > 0) {
                 buf.append("WARNING: ")
-                .append(removedRecords)
-                .append(" leak records were discarded because the leak record count is limited to ")
-                .append(MAX_RECORDS)
-                .append(". Use system property ")
-                .append(PROP_MAX_RECORDS)
-                .append(" to increase the limit.")
-                .append(NEWLINE);
+                   .append(removed)
+                   .append(" leak records were discarded because the leak record count is limited to ")
+                   .append(MAX_RECORDS)
+                   .append(". Use system property ")
+                   .append(PROP_MAX_RECORDS)
+                   .append(" to increase the limit.")
+                   .append(NEWLINE);
             }
-            buf.append("Recent access records: ")
-            .append(array.length)
-            .append(NEWLINE);
+            if (dropped > 0) {
+                buf.append(dropped)
+                   .append(" leak records were not sampled because the leak record sample count is limited to ")
+                   .append(MAX_SAMPLED_RECORDS)
+                   .append(". Use system property ")
+                   .append(PROP_MAX_SAMPLED_RECORDS)
+                   .append(" to increase the limit.")
+                   .append(NEWLINE);
+            }
 
-            if (array.length > 0) {
-                for (int i = array.length - 1; i >= 0; i --) {
+            int records = idx - removed;
+            buf.append("Recent access records: ").append(records).append(NEWLINE);
+
+            if (records > 0) {
+                // The array may not be completely filled so we need to take this into account.
+                for (int i = records - 1; i >= 0; i --) {
                     buf.append('#')
                        .append(i + 1)
                        .append(':')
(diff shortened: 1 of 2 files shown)


normanmaurer:
@carl-mastrangelo PTAL... After I had a look at your change (https://github.com/netty/netty/commit/b32cd26a961c71cc5ad0e9f88fa8c15fbfd89930) I noticed we can do even better 👍

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -328,11 +325,11 @@ protected void reportInstancesLeak(String resourceType) {
     @SuppressWarnings("deprecation")
     private final class DefaultResourceLeak extends PhantomReference<Object> implements ResourceLeakTracker<T>,
```
carl-mastrangelo: Since you aren't doing any cleanup here, why not use a WeakReference?

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -343,42 +340,29 @@ protected void reportInstancesLeak(String resourceType) {
...
-                        }
-                        lastRecords.add(value);
-                    }
+            if (head != null && MAX_RECORDS > 0) {
+                Record record = hint == null ? new Record() : new Record(hint);
+
+                synchronized (head) {
+                    tail.next = record;
+                    tail = record;
+                    numRecords++;
```
carl-mastrangelo: Since I can't comment on the actual line:

On line 373 where you call close.  It looks like you should call clear() too, which will avoid the ref being added to the queue at all.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -403,22 +387,30 @@ public boolean close(T trackedObject) {
...
-                synchronized (lastRecords) {
-                    array = lastRecords.toArray();
-                    removedRecords = this.removedRecords;
+            final String creationRecord = head.toString();
+            final String[] array;
+            int idx = 0;
+
+            synchronized (head) {
+                String last = null;
+                array = new String[numRecords];
```
carl-mastrangelo: can last moved out of the sync block?

microbench/src/main/java/io/netty/microbench/util/ResourceLeakDetectorRecordBenchmark.java
```diff
@@ -0,0 +1,80 @@
...
+    ResourceLeakDetector<Object> detector = new ResourceLeakDetector<Object>(
+            Object.class, 1, Integer.MAX_VALUE) {
+        @Override
+        protected void reportTracedLeak(String resourceType, String records) { }
+
+        @Override
+        protected void reportUntracedLeak(String resourceType) { }
+
+        @Override
+        protected void reportInstancesLeak(String resourceType) { }
```
netkins: MAJOR Add a nested comment explaining why this method is empty, throw an UnsupportedOperationException or complete the implementation. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1186)

microbench/src/main/java/io/netty/microbench/util/ResourceLeakDetectorRecordBenchmark.java
```diff
@@ -0,0 +1,80 @@
...
+    };
+
+    @Param({ "8", "16" })
+    private int recordTimes;
+    private ResourceLeakDetector.Level level;
+
+    ResourceLeakDetector<Object> detector = new ResourceLeakDetector<Object>(
+            Object.class, 1, Integer.MAX_VALUE) {
+        @Override
+        protected void reportTracedLeak(String resourceType, String records) { }
```
netkins: MAJOR Add a nested comment explaining why this method is empty, throw an UnsupportedOperationException or complete the implementation. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1186)

microbench/src/main/java/io/netty/microbench/util/ResourceLeakDetectorRecordBenchmark.java
```diff
@@ -0,0 +1,80 @@
...
+
+        @Override
+        protected void reportUntracedLeak(String resourceType) { }
```
netkins: MAJOR Add a nested comment explaining why this method is empty, throw an UnsupportedOperationException or complete the implementation. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1186)

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -460,47 +453,45 @@ public String toString() {
             "io.netty.buffer.AdvancedLeakAwareByteBuf.recordLeakNonRefCountingOperation("
     };
 
-    static String newRecord(Object hint, int recordsToSkip) {
-        StringBuilder buf = new StringBuilder(4096);
+    private static final class Record extends Throwable {
+        private final String hintString;
+        private Record next;
```
netkins: MAJOR Make this "next" field final. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1165)

common/src/main/java/io/netty/util/ResourceLeakDetector.java
normanmaurer: good idea!

common/src/main/java/io/netty/util/ResourceLeakDetector.java
normanmaurer: yes

common/src/main/java/io/netty/util/ResourceLeakDetector.java
normanmaurer: @carl-mastrangelo good question... Let me think about it and may do another PR later on. I think this is not related to this PR.

normanmaurer:
@carl-mastrangelo PTAL again

common/src/main/java/io/netty/util/ResourceLeakDetector.java
normanmaurer: false positive

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -343,50 +340,42 @@ protected void reportInstancesLeak(String resourceType) {
...
             }
         }
 
         @Override
         public boolean close() {
             // Use the ConcurrentMap remove method, which avoids allocating an iterator.
-            return allLeaks.remove(this, LeakEntry.INSTANCE);
+            if (allLeaks.remove(this, LeakEntry.INSTANCE)) {
+                // Call clear so the reference is not even enqueued.
+                //clear();
```
carl-mastrangelo: uncomment?

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -460,47 +458,45 @@ public String toString() {
...
+            StringBuilder buf = new StringBuilder(4096);
+            if (hintString != null) {
+                buf.append("\tHint: ").append(hintString).append(NEWLINE);
+            }
+
+            // Append the stack trace.
+            StackTraceElement[] array = getStackTrace();
+            // Skip the first three elements.
+            outer: for (int i = 3; i < array.length; i++) {
+                String estr = array[i].toString();
```
carl-mastrangelo: nit: if you remove the `(` characters from the elements in `STACK_TRACE_ELEMENT_EXCLUSIONS`, you don't have to call `toString` here.  You can just call `getMethodName`.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -460,47 +458,45 @@ public String toString() {
...
+
+        @Override
+        public String toString() {
+            StringBuilder buf = new StringBuilder(4096);
+            if (hintString != null) {
+                buf.append("\tHint: ").append(hintString).append(NEWLINE);
+            }
+
+            // Append the stack trace.
+            StackTraceElement[] array = getStackTrace();
```
carl-mastrangelo: One other idea for perf:  Rather than construct the string here from the stack trace, why not just pass the stack elements and the hint up?

This would have two benefits:  The critical section in `DefaultResourceLeak` would be much smaller.  Second: it would allocate less garbage.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
normanmaurer: still testing some stuff :)

common/src/main/java/io/netty/util/ResourceLeakDetector.java
normanmaurer: With pass up you mean return it ?

common/src/main/java/io/netty/util/ResourceLeakDetector.java
normanmaurer: good idea.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
carl-mastrangelo: So rather than construct and return a String in `Record`, have `DefaultResourceLeak` call `getStackTrace()` and access the hint directly.  i.e. don't use `toString()`.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
Scottmitch: actually I think you want to make this field package private.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
Scottmitch: For example ... just construct one big `StringBuilder` and construct it once instead of constructing a string building for each record, and copying the result into another `StringBuilder`.

normanmaurer: I see that said I not think it really matters as this will only be called if a leak is detected so it should not be called frequently. If its called frequently you have more serious problems to fix imho ;)

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -403,22 +392,30 @@ public boolean close(T trackedObject) {
...
+            final String[] array;
+            int idx = 0;
+            String last = null;
+
+            synchronized (head) {
+                array = new String[numRecords];
+                Record record = head.next;
+                while (record != null) {
+                    String recordStr = record.toString();
+                    if (last == null || !last.equals(recordStr)) {
```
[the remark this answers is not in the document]
normanmaurer: I just copied what was there before. I would need to check in more detail but from a quick glance it looks ok.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
normanmaurer: actually this is not quite true as I also need to take the classname into account. That said I think we can optimise it.

normanmaurer:
Another number to share is that running `mvn -Pleak clean package` in the handler module before this patch took over 10 mins on my machine... With this patch it takes 2 minutes ;)

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -453,54 +474,52 @@ public String toString() {
...
 
-        // Append the stack trace.
-        StackTraceElement[] array = new Throwable().getStackTrace();
-        for (StackTraceElement e: array) {
-            if (recordsToSkip > 0) {
-                recordsToSkip --;
-            } else {
-                String estr = e.toString();
+        private final String hintString;
+        Record next;
```
netkins: MAJOR Make this "next" field final. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1165)

common/src/main/java/io/netty/util/ResourceLeakDetector.java
Scottmitch: sound reasonable to me ... low priority IIUC ... are we missing something @carl-mastrangelo?

common/src/main/java/io/netty/util/ResourceLeakDetector.java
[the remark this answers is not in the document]
Scottmitch: oh I see ... this was done previously at the record time, but I guess you are delaying the check bcz you may be able to avoid it all together unless there is a leak ... lgtm!

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -109,7 +109,7 @@ static Level parseLevel(String levelStr) {
         Level level = Level.parseLevel(levelStr);
 
         MAX_RECORDS = SystemPropertyUtil.getInt(PROP_MAX_RECORDS, DEFAULT_MAX_RECORDS);
-
+        MAX_RECORDS_SAMPLED = MAX_RECORDS * 100;
```
Scottmitch: check for overflow? folks may set `MAX_RECORDS` to very high to make sure all records are captured.

netkins: MAJOR Cast one of the operands of this multiplication operation to a "long". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2184)

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -403,39 +423,56 @@ public boolean close(T trackedObject) {
...
+                    record = record.next;
                 }
-            } else {
-                removedRecords = 0;
-                array = EMPTY_OBJECTS;
             }
 
+            int removed = idx > MAX_RECORDS ? idx - MAX_RECORDS : 0;
+
+            long discarded = removed + dropped;
```
netkins: MAJOR Cast one of the operands of this addition operation to a "long". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2184)

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -453,54 +490,52 @@ public String toString() {
...
 
-        // Append the stack trace.
-        StackTraceElement[] array = new Throwable().getStackTrace();
-        for (StackTraceElement e: array) {
-            if (recordsToSkip > 0) {
-                recordsToSkip --;
-            } else {
-                String estr = e.toString();
+        private final String hintString;
+        Record next;
```
netkins: MAJOR Make this "next" field final. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1165)

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -453,54 +490,52 @@ public String toString() {
...
+            }
+
+            // Append the stack trace.
+            StackTraceElement[] array = getStackTrace();
+            // Skip the first three elements.
+            for (int i = 3; i < array.length; i++) {
+                StackTraceElement element = array[i];
+
+                // Strip the noisy stack trace elements.
+                if (STACK_TRACE_ELEMENT_EXCLUSIONS.contains(element.getClassName() + '.' + element.getMethodName())) {
```
carl-mastrangelo: Hmm, you could probably do even better:  just interleave the class names and method names in STACK_TRACE_ELEMENT_EXCLUSIONS, and then iterate over it.  

You would be much more likely to get the super fast == check that happens for internned strings, and you wouldn't allocate here at all, even on the stack.   

I think you are using `contains` to work around shading, but if you use `ReferenceCountUtil.getClassName()` when you initialize STACK_TRACE_ELEMENT_EXCLUSIONS, I think you can use equals.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
carl-mastrangelo: @Scottmitch passing in the string builder sounds best to me, but I suppose this will only get called when there is a problem.

@normanmaurer optional, up to you.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -343,41 +348,42 @@ protected void reportInstancesLeak(String resourceType) {
...
+                    }
+
+                    Record record = hint == null ? new Record() : new Record(hint);
+                    tail.next = record;
+                    tail = record;
+
+                    // Enforce a limited so our linked-list not grows too large and cause a GC storm later on when we
+                    // unlink it. The reason why we choose a different limit to MAX_RECORDS is that we will not handle
+                    // duplications here as it is very expensive and only will filter these out when we actually
+                    // detected a lead.
```
carl-mastrangelo: s/lead/leak/

common/src/main/java/io/netty/util/ResourceLeakDetector.java
normanmaurer: @carl-mastrangelo not sure I get what you are suggesting.. Can you give an example ? Also again it may not even matter much as its only called once a leak happens.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
normanmaurer: lol good catch

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -343,41 +348,42 @@ protected void reportInstancesLeak(String resourceType) {
...
-
-                synchronized (lastRecords) {
-                    int size = lastRecords.size();
-                    if (size == 0 || !lastRecords.getLast().equals(value)) {
-                        if (size >= MAX_RECORDS) {
-                            lastRecords.removeFirst();
-                            ++removedRecords;
-                        }
-                        lastRecords.add(value);
+            if (head != null && MAX_RECORDS > 0) {
```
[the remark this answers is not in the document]
normanmaurer: @Scottmitch I think we still want to have the creation record which is stored in head. So nope.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
normanmaurer: MAX_RECORDS_SAMPLED is a long but I think it may make sense to just cap it on `Integer.MAX_VALUE`

common/src/main/java/io/netty/util/ResourceLeakDetector.java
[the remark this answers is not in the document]
normanmaurer: exactly.. Mostly because toString() is expensive here as it calls getStackTrace()

common/src/main/java/io/netty/util/ResourceLeakDetector.java
normanmaurer: I will just ignore this for now and may do another pr if it is a problem later on.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
carl-mastrangelo: I have another change to this file, I'll show it after you submit your PR.

normanmaurer: awesome... Thanks

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -109,7 +109,8 @@ static Level parseLevel(String levelStr) {
         Level level = Level.parseLevel(levelStr);
 
         MAX_RECORDS = SystemPropertyUtil.getInt(PROP_MAX_RECORDS, DEFAULT_MAX_RECORDS);
-
+        long maxRecordsSampled = MAX_RECORDS * 10L;
```
normanmaurer: @Scottmitch do you think we should add another property to change the `10L` here ? I guess it is fine for now and we can just add it later on if needed tho.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -453,54 +503,52 @@ public String toString() {
...
-                // Strip the noisy stack trace elements.
-                boolean excluded = false;
-                for (String exclusion: STACK_TRACE_ELEMENT_EXCLUSIONS) {
-                    if (estr.startsWith(exclusion)) {
-                        excluded = true;
-                        break;
-                    }
-                }
+        Record(Object hint) {
+            // This needs to be generated even if toString() is never called as it may change later on.
```
netkins: MAJOR Make this "next" field final. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1165)

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -109,11 +111,14 @@ static Level parseLevel(String levelStr) {
         Level level = Level.parseLevel(levelStr);
 
         MAX_RECORDS = SystemPropertyUtil.getInt(PROP_MAX_RECORDS, DEFAULT_MAX_RECORDS);
+        long maxRecordsSampled = SystemPropertyUtil.getLong(PROP_MAX_SAMPLED_RECORDS, MAX_RECORDS * 10);
```
netkins: MAJOR Cast one of the operands of this multiplication operation to a "long". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2184)

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -453,54 +503,52 @@ public String toString() {
...
 
-        // Append the stack trace.
-        StackTraceElement[] array = new Throwable().getStackTrace();
-        for (StackTraceElement e: array) {
-            if (recordsToSkip > 0) {
-                recordsToSkip --;
-            } else {
-                String estr = e.toString();
+        private final String hintString;
+        Record next;
```
netkins: MAJOR Make this "next" field final. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1165)

normanmaurer:
Cherry-picked into 4.1 (3c8c7fc7e9) and 4.0 (4560197923).

PR #7223, opened 2017-09-18
extract self() method to avoid too many @SuppressWarnings

skyao:
There are many @SuppressWarnings("unchecked") in the code for the same purpose that we want to do this return:

```java
public B someMethod() {
    ......
    return (B) this;
}
```

Modification:

Add a method self() and reuse in all these return lines:

```java
private B self() {
    return (B) this;
}
```
Result:

Then only one @SuppressWarnings("unchecked") left in the code.

### DIFF
--- a/transport/src/main/java/io/netty/bootstrap/AbstractBootstrap.java
+++ b/transport/src/main/java/io/netty/bootstrap/AbstractBootstrap.java
@@ -78,7 +78,6 @@ public abstract class AbstractBootstrap<B extends AbstractBootstrap<B, C>, C ext
      * The {@link EventLoopGroup} which is used to handle all the events for the to-be-created
      * {@link Channel}
      */
-    @SuppressWarnings("unchecked")
     public B group(EventLoopGroup group) {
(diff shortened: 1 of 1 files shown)


normanmaurer:
LGTM... will merge once the CI completes.

johnou:
@skyao what's the actual benefit? thisBootstrap also sounds a bit weird.

skyao:
The benefi is to reduce the count of `@SuppressWarnings("unchecked") ` from n (in this class it is 7) to just ONE.

This style is common in many open source project, I just want to follow the style and move it to netty.

If you agree with this, I would try a search and update all similar cases.

normanmaurer:
@skyao we use the name `self()` for other cases like this. Can you please change `thisBootstrap()` to `self()`?

normanmaurer:
Cherry-picked into 4.1 (47f016bf564e9b86571fcb04d42f3ae327d0f58d) and 4.0 (5e6dd3778e).

@skyao thanks!

PR #7224, opened 2017-09-18
Correctly filter out native tcnative lib

normanmaurer:
Motivation:

c93e58c453147ab5b34a708a85530d2372bbac81 changed to use _ for the tcnative lib name but missed to also adjust the filtering.

Modifications:

Fix filtering to look for _

Result:

Not include native tcnative lib as expected.

### DIFF
--- a/all/pom.xml
+++ b/all/pom.xml
@@ -551,7 +551,7 @@
               <goal>unpack-dependencies</goal>
             </goals>
             <configuration>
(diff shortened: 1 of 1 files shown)


normanmaurer:
Verified locally that it fixed it... Will just pull in.

4.1 (8067686349) and 4.0 (0465f68ed0)

PR #7225, opened 2017-09-18
Fix assertion error when closing / shutdown native channel and SO_LIN…

normanmaurer:
…GER is set.

Motivation:

When SO_LINGER is used we run doClose() on the GlobalEventExecutor by default so we need to ensure we schedule all code that needs to be run on the EventLoop on the EventLoop in doClose. Beside this there are also threading issues when calling shutdownOutput(...)

Modifications:

- Schedule removal from EventLoop to the EventLoop
- Correctly handle shutdownOutput and shutdown in respect with threading-model
- Add unit tests

Result:

Fixes [#7159].

### DIFF
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollStreamChannel.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollStreamChannel.java
@@ -559,29 +559,8 @@ public abstract class AbstractEpollStreamChannel extends AbstractEpollChannel im
 
     @UnstableApi
     @Override
-    protected final void doShutdownOutput(Throwable cause) throws Exception {
-        try {
-            // The native socket implementation may throw a NotYetConnected exception when we attempt to shut it down.
-            // However NIO doesn't propagate an exception in the same situation (write failure), and we just want to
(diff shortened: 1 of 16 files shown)


transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollStreamChannel.java
```diff
@@ -559,25 +559,13 @@ protected Object filterOutboundMessage(Object msg) {
...
-            shutdownOutput0(newPromise());
-        } finally {
-            super.doShutdownOutput(cause);
-        }
+    protected final void doShutdownOutput() throws Exception {
+        fd().shutdown(false, true);
     }
 
-    protected void shutdownOutput0(final ChannelPromise promise) {
+    private void shutdownInput0(final ChannelPromise promise) {
```
netkins: MAJOR Private method 'shutdownInput0' is never used. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AUnusedPrivateMethod)

normanmaurer:
Cherry-picked into 4.0 (dec1679c6e)

Issue #7222, opened 2017-09-18
NativeLibraryLoader should check the result of ClassLoader#getResource method


vietj:
### Expected behavior

`NativeLibraryLoader` uses `ClassLoader#getResource` method that can return nulls when the resource cannot be found. The returned `url` variable should be checked for nullity and fail in a more usable manner than a `NullPointerException` 

### Actual behavior

`NativeLibraryLoader` fails with a `NullPointerException` (wrapped in `UnsatisfiedLinkError`) which does not give much information about the real cause of the failure.

```
java.lang.UnsatisfiedLinkError: could not load a native library: netty_transport_native_kqueue
	at io.netty.util.internal.NativeLibraryLoader.load(NativeLibraryLoader.java:178)
	at io.netty.channel.kqueue.Native.loadNativeLibrary(Native.java:114)
	at io.netty.channel.kqueue.Native.<clinit>(Native.java:54)
	at io.netty.channel.kqueue.KQueue.<clinit>(KQueue.java:33)
	at io.vertx.core.spi.transport.KQeueTransport.isAvailable(KQeueTransport.java:39)
	at io.vertx.core.spi.transport.NioTransport.nativeTransport(NioTransport.java:47)
	at io.vertx.core.spi.Transport.<clinit>(Transport.java:34)
	at io.vertx.core.impl.VertxImpl.<init>(VertxImpl.java:153)
	at io.vertx.core.impl.VertxImpl.<init>(VertxImpl.java:143)
	at io.vertx.core.impl.VertxFactoryImpl.vertx(VertxFactoryImpl.java:42)
	at io.vertx.core.Vertx.vertx(Vertx.java:91)
	at io.vertx.test.core.VertxTestBase.setUp(VertxTestBase.java:69)
	at io.vertx.test.core.HttpTestBase.setUp(HttpTestBase.java:46)
	at io.vertx.test.core.AsyncTestBase.before(AsyncTestBase.java:76)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
[...]

PR #7228, opened 2017-09-19
NativeLibraryLoader should check the result of ClassLoader#getResourc…

normanmaurer:
…e method

Motivation:

NativeLibraryLoader uses ClassLoader#getResource method that can return nulls when the resource cannot be found. The returned url variable should be checked for nullity and fail in a more usable manner than a NullPointerException

Modifications:

Fail with a FileNotFoundException

Result:

Fixes [#7222].


normanmaurer:
Added a unit test FTW

normanmaurer:
Cherry-picked into 4.1 (4d5f0e7ad5d20a754a484d0ce9ae2d771c0f8c1e) and 4.0 (8fafb316a989818518eb31f709e7d0eed7e4c8f7) as it was an easy one...

Issue #7171, opened 2017-08-31
New HTTP/2 child-channel API lacks InboundHttp2ToHttpAdapter


Bennett-Lynch:
The new HTTP/2 child-channel API is great, but it appears to lack support for something equivalent to InboundHttp2ToHttpAdapter (and outbound as well).

Is there anything planned for the new API to support this?  HTTP/2 is largely focused on multiplexing, and having the support to treat requests as HTTP/1-like is very helpful.

mosesn:
@Bennett-Lynch on the server-side, there's https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ServerDowngrader.java, you might be able to do something analogous on the client-side.

Bennett-Lynch:
Thanks. I wasn't aware of this class. Is there a client-side solution other than writing my own?

mosesn:
@Bennett-Lynch not that I'm aware of.  The client-side API is still relatively fresh.

normanmaurer:
@Bennett-Lynch @mosesn is correct... If you end up rolling your own please consider providing a PR for netty so we can include it :)

normanmaurer:
@Bennett-Lynch stay tuned I think we should have something pretty soon.

normanmaurer:
Was fixed by #7233

PR #7233, opened 2017-09-20
Adapt Http2ServerDowngrader to work with clients

lionelli1:
Motivation:
Http2ServerDowngrader is specifically built for server channels where
inbound Http2StreamFrames are converted into HttpRequests, and outbound
HttpResponses are converted into Http2StreamFrames. It can be easily
made to be more generic to work with client channels where inbound
Http2StreamFrames are converted into HttpResponses, and outbound
HttpRequests are converted into Http2StreamFrames.

Modification:
- Renamed Http2ServerDowngrader to a more general
  Http2StreamFrameToHttpObjectCodec
- Made it take in an "isServer" parameter to determine whether encoding
  inbound Http2StreamFrames should create HttpRequests (for server) or
  HttpResponses (for client)

Result:
Now Http2StreamFrameToHttpObjectCodec can be used to translate
Http2StreamFrame to HttpObject for both server and client.

Fixes https://github.com/netty/netty/issues/7171


lionelli1:

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
```diff
@@ -16,46 +16,58 @@
...
-public class Http2ServerDowngrader extends MessageToMessageCodec<Http2StreamFrame, HttpObject> {
-
+public class Http2StreamFrameToHttpObjectCodec extends MessageToMessageCodec<Http2StreamFrame, HttpObject> {
+    private final boolean isServer;
     private final boolean validateHeaders;
 
-    public Http2ServerDowngrader(boolean validateHeaders) {
+    private volatile HttpScheme scheme;
+
+    public Http2StreamFrameToHttpObjectCodec(final boolean isServer,
```
normanmaurer: @lionelli1 nit: either make both final or none of these :)

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
```diff
@@ -71,18 +83,17 @@ protected void decode(ChannelHandlerContext ctx, Http2StreamFrame frame, List<Ob
             Http2Headers headers = headersFrame.headers();
 
             if (headersFrame.isEndStream()) {
-                if (headers.method() == null) {
+                if ((headers.method() == null) && (headers.status() == null)) {
```
normanmaurer: remove `(...)` around checks ?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
```diff
@@ -132,4 +143,44 @@ protected void encode(ChannelHandlerContext ctx, HttpObject obj, List<Object> ou
...
+
+        if (isSsl(ctx.channel())) {
+            scheme = HttpScheme.HTTPS;
+        }
+    }
+
+    protected boolean isSsl(final Channel ch) {
+        final Channel channel = (ch instanceof Http2StreamChannel) ? ch.parent() : ch;
+        final ChannelPipeline p = channel.pipeline();
+        return (null != p.get(SslHandler.class)) || (null != p.get(SniHandler.class));
```
normanmaurer: remove `(...)` around the checks its not needed.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
```diff
@@ -132,4 +143,44 @@ protected void encode(ChannelHandlerContext ctx, HttpObject obj, List<Object> ou
...
+        return isServer ?
+                HttpConversionUtil.toFullHttpRequest(id, headers, alloc, validateHeaders) :
+                HttpConversionUtil.toFullHttpResponse(id, headers, alloc, validateHeaders);
+    }
+
+    @Override
+    public void handlerAdded(final ChannelHandlerContext ctx) throws Exception {
+        super.handlerAdded(ctx);
+
+        if (isSsl(ctx.channel())) {
```
normanmaurer: maybe just use:

```java
scheme = isSsl(...) ? HttpScheme.HTTPS : HttpScheme.HTTP;
```

This way its more explicit about what is used and what not

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
```diff
@@ -132,4 +143,44 @@ protected void encode(ChannelHandlerContext ctx, HttpObject obj, List<Object> ou
...
+            scheme = HttpScheme.HTTPS;
+        }
+    }
+
+    protected boolean isSsl(final Channel ch) {
```
normanmaurer: make this take `ChannelHandlerContext`

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
lionelli1: oops, missed that one out. done!

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
lionelli1: removed

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
lionelli1: yup, changed

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
lionelli1: done

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
lionelli1: removed

lionelli1:
@normanmaurer i've addressed your review comments. back to you. thanks!

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
```diff
@@ -16,46 +16,58 @@
...
  */
 @UnstableApi
-public class Http2ServerDowngrader extends MessageToMessageCodec<Http2StreamFrame, HttpObject> {
-
+public class Http2StreamFrameToHttpObjectCodec extends MessageToMessageCodec<Http2StreamFrame, HttpObject> {
+    private final boolean isServer;
     private final boolean validateHeaders;
 
-    public Http2ServerDowngrader(boolean validateHeaders) {
+    private volatile HttpScheme scheme;
```
Scottmitch: it looks like this variable is always access in the EventLoop thread ... does this have to be volatile?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
```diff
@@ -132,4 +143,43 @@ protected void encode(ChannelHandlerContext ctx, HttpObject obj, List<Object> ou
...
+        super.handlerAdded(ctx);
+
+        scheme = isSsl(ctx) ? HttpScheme.HTTPS : HttpScheme.HTTP;
+    }
+
+    protected boolean isSsl(final ChannelHandlerContext ctx) {
+        final Channel ch = ctx.channel();
+        final Channel connChannel = (ch instanceof Http2StreamChannel) ? ch.parent() : ch;
+        final ChannelPipeline p = connChannel.pipeline();
+        return null != p.get(SslHandler.class) || null != p.get(SniHandler.class);
```
Scottmitch: if SNI is in the pipeline we should wait until a `SslHandshakeCompletionEvent` is received in the pipeline before we consider the connection to be `HTTPS`.

For example:
```java
@Override
public void userEventTriggered(ChannelHandlerContext ctx, Object evt) {
  if (evt instanceof SslHandshakeCompletionEvent) {
    // SslHandler should be in the pipeline
  }
}
```

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodecTest.java
```diff
@@ -0,0 +1,799 @@
...
+            goaway.release();
+            frame.release();
+        }
+    }
+
+    private SslHandler buildSsl() throws Exception {
+        final SslContext ctx = SslContextBuilder.forClient().build();
+
+        // this SslHandler bypasses Http2StreamFrames so we can verify the
+        // object written
```
Scottmitch: nit: put this line on the previous line

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
normanmaurer: +1... not need for volatile

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
normanmaurer: +1

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodecTest.java
normanmaurer: +1

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
normanmaurer: @lionelli1 @Scottmitch that said we need to ensure we do this as part of the "correct" pipeline which may be the parent pipeline or not depending on if `Http2MultiplexCodex` is used or not.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
lionelli1: good point. removed volatile

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodecTest.java
lionelli1: done

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
lionelli1: but for most of the use cases where this handler is added at the child stream, the SslHandshakeCompletionEvent would've happened at the parent channel by the time it is added to the stream channels. the new child channels won't be notified of this event upon creation, would it?

normanmaurer: yep ... maybe its ok for now to just do this only if the its not a "child channel" for now

lionelli1: ok changed to checking on SslHandler with the assumption that this handler will mostly be used on an `Http2StreamChannel`

normanmaurer:
Cherry-picked into 4.1 (e069079aff5111b45d333bcce5432260fad99076)

PR #7234, opened 2017-09-20
Expose all defaults on PooledByteBufAllocator

carl-mastrangelo:
Motivation:
Most, but not all defaults are statically exposed on
PooledByteBufAllocator.  This makes it cumbersome to make a custom
allocator where most of the defaults remain the same.

Modification:
Expose useCacheForAllThreads, and Direct preferred.  The latter is
needed because it is under the internal package, and public code
should probably not depend on it.

Result:
More customizeable allocators


normanmaurer:
Cherry-picked into 4.1 (003c8cc7abcfbe0c7a3ce97a23413ba4c336034a) and 4.0 (9e95d4382b08f99a7134c0f88979a8ba9f94d3e9)

PR #7241, opened 2017-09-21
Fix regression in reporting leaks introduced by 3c8c7fc7e9c27f87e64aa…

normanmaurer:
…d5bd1df6c58ed8ef36e.

Motivation:

3c8c7fc7e9c27f87e64aad5bd1df6c58ed8ef36e introduced some changes to the ResourceLeakDetector that introduced a regression and so would always log that paranoid leak detection should be enabled even it was already.

Modifications:

Correctly not clear the recorded stacktraces when we process the reference queue so we can log these.

Result:

ResourceLeakDetector works again as expected.

### DIFF
--- a/common/src/main/java/io/netty/util/ResourceLeakDetector.java
+++ b/common/src/main/java/io/netty/util/ResourceLeakDetector.java
@@ -266,7 +266,7 @@ public class ResourceLeakDetector<T> {
             if (ref == null) {
                 break;
             }
-            ref.close();
+            ref.dispose();
         }
     }
 
@@ -284,9 +284,7 @@ public class ResourceLeakDetector<T> {
                 break;
             }
 
-            ref.clear();
-
-            if (!ref.close()) {
+            if (!ref.dispose()) {
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


normanmaurer:
my fault :(

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -394,6 +392,11 @@ private void record0(Object hint) {
             }
         }
 
+        boolean dispose() {
+            clear();
+            return allLeaks.remove(this, LeakEntry.INSTANCE);
```
carl-mastrangelo: I noticed this when I did my CL, but hadn't come up with a solution.   Apparently allLeaks was never cleared, even since 2013!.

normanmaurer: it was as it was done as part of `close()` the problem was that it also cleared the linked-list of records that we tried to access then on `toString()` and so was not able to print anything 👎

normanmaurer:
Cherry-picked into 4.1 (c3298a3836c244c94fae9a9cbc1598ea3820d357) and 4.0 (2053d07cab5968acc9308ca27f4c808e11afed8b)

Scottmitch:
@normanmaurer - is it is worth re-running the benchmarks? seems like this may have preemptively thrown tracking info away?

normanmaurer:
@Scottmitch I did and the numbers are the same as we still did all the recording just failed to log them when there was a real leak.

PR #7245, opened 2017-09-23
Adding stable JDK9 automatic module names derived from main packages

### LINKED ISSUE #7218: Add 'Automatic-Module-Name' to MANIFEST for forward compatibility with JDK9
When a non-modular library (e.g. built for JDK8 or JDK9) is used within a modular JDK9 application, Java creates an automatic module for it. Modules should be named from the root packages they own, but the default algorithm uses JAR file name instead. If Netty migrates to JDK9 and Jigsaw one day, this may cause serious dependency issues (JVM will refuse to start, if the same module is required with two different names, e.g. `netty` and `io.netty`).

It is possible to specify the automatic module name explicitly through a Manifest entry `Automatic-Module-Name`, while the project is still built against JDK7.

I suggest to add `Automatic-Module-Name` to the manifests to acquire your root packages as module names, since JDK9 is going to be released within a few days.

zyxist:
Motivation:

`Automatic-Module-Name` entry provides a stable JDK9 module name, when Netty is used in a modular JDK9 applications. More info: http://blog.joda.org/2017/05/java-se-9-jpms-automatic-modules.html

When Netty migrates to JDK9 in the future, the entry can be replaced by actual `module-info` descriptor.

Modification:

The POM-s are configured to put the correct module names to the manifest.

Result:

Fixes #7218.

### DIFF
--- a/codec-redis/pom.xml
+++ b/codec-redis/pom.xml
@@ -28,6 +28,10 @@
 
   <name>Netty/Codec/Redis</name>
 
+  <properties>
+    <javaModuleName>io.netty.codec.redis</javaModuleName>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/codec-socks/pom.xml
+++ b/codec-socks/pom.xml
@@ -28,6 +28,10 @@
 
   <name>Netty/Codec/Socks</name>
 
+  <properties>
+    <javaModuleName>io.netty.codec.socks</javaModuleName>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -30,6 +30,7 @@
   <name>Netty/Common</name>
 
   <properties>
+    <javaModuleName>io.netty.common</javaModuleName>
     <collection.template.dir>${project.basedir}/src/main/templates</collection.template.dir>
     <collection.template.test.dir>${project.basedir}/src/test/templates</collection.template.test.dir>
     <collection.src.dir>${project.build.directory}/generated-sources/collections/java</collection.src.dir>
--- a/transport-native-epoll/pom.xml
+++ b/transport-native-epoll/pom.xml
@@ -27,6 +27,7 @@
   <packaging>jar</packaging>
 
   <properties>
+    <javaModuleName>io.netty.transport.epoll</javaModuleName>
     <!-- Needed by the native transport as we need the memoryAddress of the ByteBuffer -->
     <argLine.java9.extras>--add-exports java.base/sun.security.x509=ALL-UNNAMED --add-opens=java.base/java.nio=ALL-UNNAMED</argLine.java9.extras>
     <unix.common.lib.name>netty-unix-common</unix.common.lib.name>
--- a/transport-native-kqueue/pom.xml
+++ b/transport-native-kqueue/pom.xml
@@ -353,6 +353,7 @@
   </profiles>
 
   <properties>
+    <javaModuleName>io.netty.transport.kqueue</javaModuleName>
     <!-- Needed by the native transport as we need the memoryAddress of the ByteBuffer -->
     <argLine.java9.extras>--add-exports java.base/sun.security.x509=ALL-UNNAMED --add-opens=java.base/java.nio=ALL-UNNAMED</argLine.java9.extras>
     <unix.common.lib.name>netty-unix-common</unix.common.lib.name>
--- a/transport-native-unix-common/pom.xml
+++ b/transport-native-unix-common/pom.xml
@@ -30,6 +30,8 @@
   </description>
 
   <properties>
+    <javaModuleName>io.netty.transport.unix.common</javaModuleName>
+
     <exe.make>make</exe.make>
     <exe.compiler>gcc</exe.compiler>
     <exe.archiver>ar</exe.archiver>
--- a/README.md
+++ b/README.md
@@ -24,3 +24,40 @@ Note that this is build-time requirement.  JDK 5 (for 3.x) or 6 (for 4.0+) is en
 ## Branches to look
 
 Development of all versions takes place in each branch whose name is identical to `<majorVersion>.<minorVersion>`.  For example, the development of 3.9 and 4.0 resides in [the branch '3.9'](https://github.com/netty/netty/tree/3.9) and [the branch '4.0'](https://github.com/netty/netty/tree/4.0) respectively.
+
+## Usage with JDK 9
+
+Netty can be used in modular JDK9 applications as a collection of automatic modules. The module names follow the
+reverse-DNS style, and are derived from subproject names rather than root packages due to historical reasons. They
+are listed below:
+
+ * `io.netty.buffer`
+ * `io.netty.codec`
+ * `io.netty.codec.dns`
+ * `io.netty.codec.haproxy`
+ * `io.netty.codec.http`
+ * `io.netty.codec.http2`
+ * `io.netty.codec.memcache`
+ * `io.netty.codec.mqtt`
+ * `io.netty.codec.redis`
+ * `io.netty.codec.smtp`
+ * `io.netty.codec.socks`
+ * `io.netty.codec.stomp`
+ * `io.netty.codec.xml`
+ * `io.netty.common`
+ * `io.netty.handler`
+ * `io.netty.handler.proxy`
+ * `io.netty.resolver`
+ * `io.netty.resolver.dns`
+ * `io.netty.transport`
+ * `io.netty.transport.epoll` (`native` omitted - reserved keyword in Java)
+ * `io.netty.transport.kqueue` (`native` omitted - reserved keyword in Java)
+ * `io.netty.transport.unix.common` (`native` omitted - reserved keyword in Java)
+ * `io.netty.transport.rxtx`
+ * `io.netty.transport.sctp`
+ * `io.netty.transport.udt`
+
+
+
+Automatic modules do not provide any means to declare dependencies, so you need to list each used module separately
+in your `module-info` file.
--- a/buffer/pom.xml
+++ b/buffer/pom.xml
@@ -28,6 +28,10 @@
 
   <name>Netty/Buffer</name>
 
+  <properties>
+    <javaModuleName>io.netty.buffer</javaModuleName>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/codec-dns/pom.xml
+++ b/codec-dns/pom.xml
@@ -28,6 +28,10 @@
 
   <name>Netty/Codec/DNS</name>
 
+  <properties>
+    <javaModuleName>io.netty.codec.dns</javaModuleName>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/codec-haproxy/pom.xml
+++ b/codec-haproxy/pom.xml
@@ -28,6 +28,10 @@
 
   <name>Netty/Codec/HAProxy</name>
 
+  <properties>
+    <javaModuleName>io.netty.codec.haproxy</javaModuleName>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/codec-http/pom.xml
+++ b/codec-http/pom.xml
@@ -28,6 +28,10 @@
 
   <name>Netty/Codec/HTTP</name>
 
+  <properties>
+    <javaModuleName>io.netty.codec.http</javaModuleName>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/codec-http2/pom.xml
+++ b/codec-http2/pom.xml
@@ -28,6 +28,10 @@
 
   <name>Netty/Codec/HTTP2</name>
 
+  <properties>
+    <javaModuleName>io.netty.codec.http2</javaModuleName>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>${project.groupId}</groupId>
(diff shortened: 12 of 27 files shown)


common/pom.xml
```diff
@@ -30,6 +30,7 @@
   <name>Netty/Common</name>
 
   <properties>
+    <javaModuleName>io.netty.util</javaModuleName>
```
johnou: What's wrong with common?

transport-native-epoll/pom.xml
```diff
@@ -27,6 +27,7 @@
   <packaging>jar</packaging>
 
   <properties>
+    <javaModuleName>io.netty.channel.epoll</javaModuleName>
```
johnou: channel or transport?

transport-native-kqueue/pom.xml
```diff
@@ -353,6 +353,7 @@
   </profiles>
 
   <properties>
+    <javaModuleName>io.netty.channel.kqueue</javaModuleName>
```
johnou: Channel or transport?

transport-native-unix-common/pom.xml
```diff
@@ -30,6 +30,8 @@
   </description>
 
   <properties>
+    <javaModuleName>io.netty.channel.unix</javaModuleName>
```
johnou: Unix common no? It is a shared command module, not a unix implementation.

common/pom.xml
zyxist: The module names should be derived from the main package within them. This is why I chose `util` instead of `common` and `channel` instead of `transport`.

Mark Reinhold: *Strongly recommend that all modules be named according to the reverse Internet domain-name convention. A module's name should correspond to the name of its principal exported API package, which should also follow that convention.*

http://blog.joda.org/2017/04/java-se-9-jpms-module-naming.html

transport-native-unix-common/pom.xml
zyxist: If this module is just an "internal" stuff for the rest of Netty, I agree that we don't have to configure an automatic module name for it. It will be used just as a classpath dependency, which is possible for automatic modules (but not for true modules).

The only issue is that the manifest generator is in the parent POM, and removing this line would create an empty module name in the manifest.

johnou:
@mp911de ptal

mp911de:
Looks technically okay, I'm a bit concerned regarding the naming. During the process of finding a consensus for naming rules and automatic naming, a series of controversial approaches were discussed of which each one comes with its benefits and consequences.

I find it quite hard to reason why module names diverge so far from the actual artifact name if the artifact is not modular in the first place (not speaking about any prefixes like `io.` or such). From a user perspective, I wish to know a scheme for module names so I can derive the module name for each artifact, with as little required input as possible. I don't want to be required to look up module names from the documentation or look into each jar file to inspect the package structure.

For now (no `module-info.java` present), module names are more of a cosmetic/documentation arrangement that could easily change once netty provides module descriptors.

One could argue that artifacts are named the wrong way, such as `netty-transport-native-epoll` should be named `netty-channel-epoll` which would then match `io.netty.channel.epoll`, but I don't want to open a discussion about artifact names, just briefly mention how things are named.

From this standing point, I'd argue to align module names to the artifact names to document the `io.netty` module namespace (that would be `io.netty.transport.native.epoll` for `netty-transport-native-epoll`) and to avoid confusion around module names.

Another possibility could be to drop the `io.` prefix and use the artifact name (replace `-` with `.`) as the automatic module name (like [Spring Framework 5](https://github.com/spring-projects/spring-framework/blob/2d6364eb834791bf5a35ec274a71d8dbccab8b67/build.gradle#L178). Unless artifact names are going to change, the module names are stable as well. 

Module names (without a descriptor) for the sake of the name do not provide true value to Java 9 projects and need to be revisited at a later stage anyway once a descriptor will be added.

zyxist:
OK, I understand your point of view :). Originally the things were named `channel`, etc. but later the terminology was changed. The artifacts represent new names, but packages for obvious reasons must keep the old names. And I understand that the users can benefit from the module names that are better aligned with the new terminology.

At the same time, you made some statements about modules that can lead you (or the users of Netty) into the trouble in the future.

First of all, there is a good explanation at Stephen's blog about the relationship modules vs. artifacts, and the further conclusion about naming:

 - http://blog.joda.org/2017/04/java-se-9-jpms-modules-are-not-artifacts.html
 - http://blog.joda.org/2017/04/java-se-9-jpms-module-naming.html

He suggests to name the module from the main exported package, so finding out the module name should not be that confusing for the users (it could be even calculated by IDE-s). His explanation is convincing for me:

 - *"Modules are a group of packages. As such, the module name must be related to the package names."*
 - *"Creating a module with a particular name takes ownership of that package name and everything beneath it."*
 - *"As the owner of that namespace, any sub-packages may be grouped into sub-modules as desired so long as no package is in two modules."*

Netty is not a new project, so the recommendations can be relaxed, but I think that at least reverse DNS style is important.

> For now (no module-info.java present), module names are more of a cosmetic/documentation arrangement that could easily change once netty provides module descriptors.

> Module names (without a descriptor) for the sake of the name do not provide true value to Java 9 projects and need to be revisited at a later stage anyway once a descriptor will be added.

I'd be VERY careful with changing/aligning the module names in the future. They are not cosmetic for modular projects that would use Netty as a dependency, even if Netty is not using modules right now. If you change it at a later stage, when there will be many projects that depend on existing name, they will start falling into a module hell. Suppose that you have two modules A and B. A uses the old Netty name, and B uses the new Netty name. Now you won't be able to use both A and B in the same project, because Java will see two distinct modules with the same packages and classify them as a *package split*, which is illegal in Jigsaw. Such project will not even compile. Take a look at examples given by Stephen - there is a good example, what would happen:

 - http://blog.joda.org/2017/05/java-se-9-jpms-automatic-modules.html

And this is why I opened this issue/PR. If this were a cosmetic thing for Jigsaw, I wouldn't care, too. I won't be very happy, if my project stops compiling in the future because of module name changes :).

> From this standing point, I'd argue to align module names to the artifact names to document the io.netty module namespace (that would be io.netty.transport.native.epoll for netty-transport-native-epoll) and to avoid confusion around module names.

Although I'm more convinced by Stephen's explanation, this proposal sounds OK for me. It keeps the reverse DNS style, but it also takes into account that Netty has some history behind it, and new things should be named in the new way. I'll make the necessary changes tomorrow.

zyxist:
Hello, sorry for the delay. I fixed the module names according to the discussion. Please check if everything is OK.

codec-redis/pom.xml
```diff
@@ -28,6 +28,10 @@
 
   <name>Netty/Codec/Redis</name>
 
+  <properties>
+    <javaModuleName>io.netty.handler.codec.redis</javaModuleName>
```
mp911de: Shouldn't this be rather `io.netty.codec.redis`?

codec-socks/pom.xml
```diff
@@ -28,6 +28,10 @@
 
   <name>Netty/Codec/Socks</name>
 
+  <properties>
+    <javaModuleName>io.netty.handler.codec.socks</javaModuleName>
```
mp911de: Shouldn't this be rather `io.netty.codec.socks`?

normanmaurer:
@carl-mastrangelo PTAL as well.

johnou:
@zyxist missed updating two of the modules.

zyxist:
These names are not missed. The discussion was about things that are named differently (that is: *channel* was the old name, *transport* is the new name). `codec` is still a codec, so I kept the adjustment to the root package names.

johnou:
@zyxist I mean io.netty.handler.codec.redis and io.netty.handler.codec.socks

zyxist:
Oh, I see - thanks :).

zyxist:
I added one more change - removing the "native" word from module names, because it is a reserved keyword in Java, and one would not be able to require such a module in `module-info.java`.

https://github.com/netty/netty/issues/7266

normanmaurer:
@Scottmitch @carl-mastrangelo please review (any of you) and pull in once satisfied to 4.1 and 4.0. Also ensure @zyxist squashed it before.

normanmaurer:
@zyxist can you please squash the commits ?

zyxist:
Done.

johnou:
@zyxist please bump the maven compiler plugin version to 3.7.0 https://cwiki.apache.org/confluence/display/MAVEN/Java+9+-+Jigsaw

zyxist:
@johnou -> do you plan to deliver separate JAR archives for JDK9 and JDK6/7/8? Your source and target Java version is JDK6, it is technically not possible to use `module-info` files with JDK6. To support `module-info` descriptors, the plugin actually performs the build twice and produces two JAR archives: one for older Java version, one for Java 9. This affects the entire build pipeline and is quite big amount of work to do that goes out of scope of this Pull Request.

Just bumping the plugin version without additional configuration gives nothing.

Can you merge this Pull Request to give Netty a stable module name, to unlock me and other users, and meanwhile start work on updating your build pipeline as a separate task?

johnou:
Aha makes sense, that would be up to the maintainers.

LordAkkarin:
I'm sorry if I have to chime in and correct above's information, but a specialized configuration of the current maven-compiler-plugin (as [outlined in its documentation](https://maven.apache.org/plugins/maven-compiler-plugin/examples/module-info.html)) will not result in two separate jars but rather a single jar consisting of classes which are targeted at both releases. This method has, for instance, been [adopted by slf4j already](https://github.com/qos-ch/slf4j/blob/master/pom.xml#L174). Given that versions earlier than Java 9 will not attempt to load the module-info class (unless somehow instructed to do so), this method guarantees full backward compatibility.

That being said, I do agree that the definition of the legacy headers (as implemented by this PR) should suffice at the moment to allow consistent references to netty through Jigsaw enabled projects at the moment. A module-info should, however, still be the ultimate target here (especially given the implementation of the [jlink](http://openjdk.java.net/jeps/282) tool which seems to require such definitions to work at all).

normanmaurer:
So @LordAkkarin @johnou @zyxist @mp911de I am correct in my understanding that this is ready to be merged and good as it is for now ?

johnou:
@normanmaurer yep.

johnou:
@normanmaurer actually one thing just came to mind..

johnou:
@zyxist what happens to modules where javaModuleName is not defined? eg. dev-tools does it end up with an empty string header for Automatic-Module-Name in the manifest? does this cause problems? or is it simply ignored because it is empty?

normanmaurer:
@zyxist ping...

doom369:
@johnou @normanmaurer nothing will happen. Modules without `javaModuleName` will get the automatic module name derived from jar name. So it is safe to leave it as it. Also, this module has no java classes, so it is totally fine.

johnou:
@doom369 so modules without the javaModuleName property defined in properties will end up with an empty tag [1] then fall back to the automatic module name derived from the jar name?

[1] root pom.xml

```xml
<manifestEntries>
    <Automatic-Module-Name>${javaModuleName}</Automatic-Module-Name>
</manifestEntries>
```

doom369:
@johnou in general - yes (modules without this property will have empty `Automatic-Module-Name` field in manifest file). However, it has nothing to with `dev-tools`. As it is not actually included to any artifact jar. It is only used to execute the specfic maven plugin.

johnou:
@doom369 my main concern was the behaviour of a jar with an empty Automatic-Module-Name, dev-tools was a bad example.

doom369:
@johnou empty name is not allowed for` Automatic-Module-Name`. You'll get 

`Automatic-Module-Name: : Invalid module name: '' is not a Java identifier`

Just checked :).

doom369:
@Scottmitch @carl-mastrangelo sorry for pushing, could you please review this PR? So those of us who moved to Java 9 could finish modularization.

johnou:
@doom369 tbh I'm still not convinced with the usage of Maven properties and Automatic-Module-Name in the root pom.xml within pluginManagement, it seems brittle and hacky (the manifest for the testsuite jar has the invalid entry "Automatic-Module-Name: "). Instead why not override the maven-jar-plugin config in each module pom.xml like so..

```xml
  <build>
    <plugins>
      <plugin>
        <artifactId>maven-jar-plugin</artifactId>
        <executions>
          <execution>
            <id>default-jar</id>
            <configuration>
              <archive>
                <manifestEntries>
                  <Automatic-Module-Name>${javaModuleName}</Automatic-Module-Name>
                </manifestEntries>
              </archive>
            </configuration>
          </execution>
        </executions>
      </plugin>
...
```
I think this approach seems more like the Maven way?

mp911de:
@johnou Defining the plugin in the parent's `pluginManagement` is the approach to go. You don't need to repeat yourself in each `pom.xml`, only set the property in all submodules. I think that should be done for every other sub-module to comply with the requirements coming from the parent pom /cc @zyxist

johnou:
@mp911de yep that's what I was getting at, it should be all submodules if it's defined in the parent pom or defined in each submodule if that's not desirable for whatever reason.

normanmaurer:
@zyxist can you ensure you add it for all sub-modules ? After this I will merge it

normanmaurer:
@zyxist ping

doom369:
@normanmaurer PR is good. All needed modules are with correct names. Do mean some specific module?

normanmaurer:
Cherry-picked into 4.1 (e8540c2b7a1243618002c857be5ba463984e1b76)

normanmaurer:
@zyxist thanks!

udaychandra:
Got hold of the latest netty-transport-native-kqueue:4.1.18.Final:osx-x86_64 jar from maven central to use it in a Java 9 modular project. The build fails because it cannot find 'io.netty.transport.kqueue' module. I still don't see the 'Automatic-Module-Name' entry in the manifest file of the jar. Am I missing something?

The jar command's "--describe-module" option complains about the same. Reported error when I ran "jar --file <path-to-jar> --describe-module" is:

netty.transport.native.kqueue: Invalid module name: 'native' is not a Java identifier

PR #7250, opened 2017-09-26
Propagate all exceptions when loading native code

### LINKED ISSUE #7249: Epoll static loading broken in 4.1.16
### Expected behavior
Be able to load Epoll statically

### Actual behavior
Loading fails because the shared object can't be found.

carl-mastrangelo:
Motivation:
There are 2 motivations, the first depends on the second:

Loading Netty Epoll statically stopped working in 4.1.16, due to
`Native` always loading the arch specific shared object.  In a
static binary, there is no arch specific SO.

Second, there are a ton of exceptions that can happen when loading
a native library.  When loading native code, Netty tries a bunch of
different paths but a failure in any given may not be fatal.

Additionally: turning on debug logging is not always feasible so
exceptions get silently swallowed.

Modifications:

* Change Epoll and Kqueue to try the static load second
* Modify NativeLibraryLoader to record all the locations where
  exceptions occur.
* Attempt to use `addSuppressed` from Java 7 if available.

Alternatives Considered:

An alternative would be to record log messages at each failure.  If
all load attempts fail, the log messages are printed as warning,
else as debug. The problem with this is there is no `LogRecord` to
create like in java.util.logging.  Buffering the args to
logger.log() at the end of the method loses the call site, and
changes the order of events to be confusing.

Another alternative is to teach NativeLibraryLoader about loading
the SO first, and then the static version.  This would consolidate
the code fore Epoll, Kqueue, and TCNative.   I think this is the
long term better option, but this PR is changing a lot already.
Someone else can take a crack at it later

Results:
Epoll Still Loads and easier debugging.

Fixes #7249

### DIFF
--- a/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
+++ b/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
@@ -30,8 +30,10 @@ import java.lang.reflect.Method;
 import java.net.URL;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.EnumSet;
+import java.util.List;
 import java.util.Set;
 
 /**
@@ -77,16 +79,20 @@ public final class NativeLibraryLoader {
      *         if none of the given libraries load successfully.
      */
     public static void loadFirstAvailable(ClassLoader loader, String... names) {
+        List<Throwable> suppressed = new ArrayList<Throwable>();
         for (String name : names) {
             try {
                 load(name, loader);
                 return;
             } catch (Throwable t) {
+                suppressed.add(t);
                 logger.debug("Unable to load the library '{}', trying next name...", name, t);
             }
         }
-        throw new IllegalArgumentException("Failed to load any of the given libraries: "
-                                           + Arrays.toString(names));
+        IllegalArgumentException iae =
+                new IllegalArgumentException("Failed to load any of the given libraries: " + Arrays.toString(names));
+        ThrowableUtil.addSuppressedAndClear(iae, suppressed);
+        throw iae;
     }
 
     /**
@@ -112,12 +118,13 @@ public final class NativeLibraryLoader {
     public static void load(String originalName, ClassLoader loader) {
         // Adjust expected name to support shading of native libraries.
         String name = calculatePackagePrefix().replace('.', '_') + originalName;
-
+        List<Throwable> suppressed = new ArrayList<Throwable>();
         try {
             // first try to load from java.library.path
             loadLibrary(loader, name, false);
             return;
         } catch (Throwable ex) {
+            suppressed.add(ex);
             logger.debug(
                     "{} cannot be loaded from java.libary.path, "
                     + "now trying export to -Dio.netty.native.workdir: {}", name, WORKDIR, ex);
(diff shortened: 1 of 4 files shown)


common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -257,25 +281,36 @@ public Object run() {
...
+            try {
+                // The helper class is NOT found in target ClassLoader, we have to define the helper class.
+                final byte[] classBinary = classToByteArray(helper);
+                return AccessController.doPrivileged(new PrivilegedAction<Class<?>>() {
+                    @Override
+                    public Class<?> run() {
+                        try {
+                            // Define the helper class in the target ClassLoader,
+                            //  then we can call the helper to load the native library.
+                            Method defineClass = ClassLoader.class.getDeclaredMethod("defineClass", String.class,
```
carl-mastrangelo: Also, this looks like a possible bug.  Shouldn't `loader` be used to lookup `defineClass`, not `ClassLoader.class` ?

cc: @ejona86

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -257,25 +281,36 @@ public Object run() {
...
-                    } catch (Exception e) {
-                        throw new IllegalStateException("Define class failed!", e);
+        } catch (ClassNotFoundException e1) {
+            try {
+                // The helper class is NOT found in target ClassLoader, we have to define the helper class.
+                final byte[] classBinary = classToByteArray(helper);
+                return AccessController.doPrivileged(new PrivilegedAction<Class<?>>() {
+                    @Override
+                    public Class<?> run() {
+                        try {
```
netkins: MAJOR Extract this nested try block into a separate method. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1141)

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -201,19 +216,29 @@ public static void load(String originalName, ClassLoader loader) {
      * @param absolute - Whether the native library will be loaded by path or by name
      */
     private static void loadLibrary(final ClassLoader loader, final String name, final boolean absolute) {
+        Throwable suppressed = null;
         try {
-            // Make sure the helper is belong to the target ClassLoader.
-            final Class<?> newHelper = tryToLoadClass(loader, NativeLibraryUtil.class);
-            loadLibraryByHelper(newHelper, name, absolute);
+            try {
```
netkins: MAJOR Extract this nested try block into a separate method. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1141)

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
normanmaurer: good question...

normanmaurer:
@Scottmitch PTAL as well.

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
ejona86: Since `loader` is known to be an instanceof `ClassLoader` using `ClassLoader.class` here is fine. It stands out as strange because the reflection is to avoid accessibility, not because the class is unknown at compile time. So it's good and fine as-is, just most other reflection usages couldn't do this.

normanmaurer:
Cherry-picked into 4.1 (d3ca087f6b35f8db3417324a62d43790f3d5fc62) and 4.0 (4a572e06d63fb1eb277f86300219c33032a3cec8)

Issue #7251, opened 2017-09-26
ApplicationProtocolNegotiationHandler does not use fallback when with Java 9 and extension ignored  


vietj:
### Expected behavior

The fallback given to the `ApplicationProtocolNegotiationHandler ` is provided in the `configurePipeline` method.

### Actual behavior

The `configurePipeline` method is called with an empty string instead of the given fallback protocol. This is because the engine returns an empty string and the `ApplicationProtocolNegotiationHandler ` checks the nullity of the protocol string, it should also check wether it's empty or not.

### Steps to reproduce

A bootstrap setup with an `SslHandler` and the `ApplicationProtocolNegotiationHandler` using JDK engine. The bootstrap connects to an SSL server that does not support ALPN and ignores the extension.

### Minimal yet complete reproducer code (or URL to code)

None provided (obvious case, can contribute something next week)

### Netty version

4.1.15.Final

### JVM version (e.g. `java -version`)

```
java version "9"
Java(TM) SE Runtime Environment (build 9+181)
Java HotSpot(TM) 64-Bit Server VM (build 9+181, mixed mode)
```
### OS version (e.g. `uname -a`)

```
Darwin julien 15.6.0 Darwin Kernel Version 15.6.0: Fri Feb 17 10:21:18 PST 2017; root:xnu-3248.60.11.4.1~1/RELEASE_X86_64 x86_64
```

normanmaurer:
@vietj I am not sure I really understand the issue :( Can you point me to the code you are talking about. For example which `configurePipeline` method ?

vietj:
here : https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/ApplicationProtocolNegotiationHandler.java#L92

I think `protocol != null` should be replaced by `protocol != null && protocol.length() > 0`

normanmaurer:
Ok got it... I think we need a different fix but I will take care today

normanmaurer:
Fixed by https://github.com/netty/netty/pull/7258

PR #7258, opened 2017-09-28
Fix Java9SslEngine implementation of ApplicationProtocolAccessor and …

normanmaurer:
…so fix ApplicationProtocolNegationHandler

Motivation:

Java9SslEngine did not correctly implement ApplicationProtocolAccessor and so returned an empty String when no extension was used while the interface contract is to return null. This lead to the situation that ApplicationProtocolNegationHandler did not correct work.

Modifications:

- Rename ApplicationProtocolAccessor.getApplicationProtocol() to getNegotiatedApplicationProtocol() which resolves the clash with the method exposed by Java9s SSLEngine.
- Correctly implement getNegotiatedApplicationProtocol() for Java9Sslengine
- Add delegate in Java9Sslengine.getApplicationProtocol() which is provided by Java9
- Adjust tests to test the correct behaviour.

Result:

Fixes [#7251].

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ApplicationProtocolAccessor.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ApplicationProtocolAccessor.java
@@ -26,5 +26,5 @@ interface ApplicationProtocolAccessor {
      * @return the application-level protocol name or
      *         {@code null} if the negotiation failed or the client does not have ALPN/NPN extension
      */
-    String getApplicationProtocol();
+    String getNegotiatedApplicationProtocol();
 }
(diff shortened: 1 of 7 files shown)


normanmaurer:
@vietj PTAL

normanmaurer:
@vietj did you verify the fix as well with your testcase ?

normanmaurer:
Cherry-picked into 4.1 (bb1833f22c1e43cd5ea2d75fbc2c5587d062ac04) and 4.0 (5500a01f995cb7bada5495e334d52da482475f89)

Issue #7255, opened 2017-09-27
A lot of ChannelInputShutdownReadComplete events fired for NIO channel with ALLOW_HALF_CLOSURE


khitrin:
### Expected behavior
Single `ChannelInputShutdownReadComplete` event will be fired for closed socket.

### Actual behavior
`ChannelInputShutdownReadComplete` event fired on each NIO event loop run.
`NioEventLoop` begin consuming a lot of CPU.

### Steps to reproduce
Create upstream connection channel with enabled `ALLOW_HALF_CLOSURE` and disabled `AUTO_READ`, do a single HTTP keep-alive request and wait until connection being closed by remote server.

### Netty version
4.1.16.Final

### JVM version (e.g. `java -version`)
Java(TM) SE Runtime Environment (build 1.8.0_131-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.131-b11, mixed mode)

### OS version (e.g. `uname -a`)
Darwin 16.7.0 Darwin Kernel Version 16.7.0: Thu Jun 15 17:36:27 PDT 2017; root:xnu-3789.70.16~2/RELEASE_X86_64 x86_64

macOS Sierra 10.12.6

Also confirmed on Linux.

### Suspicious code fragment

from `io.netty.channel.nio.AbstractNioByteChannel::read`

https://github.com/netty/netty/blob/f57d8f00e17a3cbd127aeb2b499c3f2787c002e4/transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java#L124-L133

`readPending` not cleared when `allocHandle.lastBytesRead() <= 0`.

Negative read bytes indicates socket closed state, but each consecutive selection of this socket key with `OP_READ` will succeed immediately (may be related to http://bugs.java.com/bugdatabase/view_bug.do?bug_id=4531726).

### Fix proposal

Call `removeReadOp()` on closed sockets.

khitrin:
Thank you for fast response and this nice and similar fix for all channel implementations.

But, after checking code for every four affected classes, I afraid that this problem can hurt non-NIO implementations and channels with auto-read.

khitrin:
Thank you for fast response. It will fix issue for non-autoread channels (original issue), but after looking at affected code I'm afraid that channels with `AUTO_READ` can be hurt with this wakeups.

normanmaurer:
@khitrin so you are saying it is not fixed with the PR ?

khitrin:
[...]

PR #7259, opened 2017-09-28
Set readPending to false when EOF is detected while issue an read

normanmaurer:
Motivation:

We need to set readPending to false when we detect a EOF while issue a read as otherwise we may not unregister from the Selector / Epoll / KQueue and so keep on receving wakeups.

Modifications:

Set readPending to false when EOF is detected.

Result:

Fixes [#7255].


normanmaurer:
I still need to add  tests as well.

Scottmitch:
I would like to propose https://github.com/netty/netty/pull/7331 as an alternative.

Scottmitch:
Fixed by #7331

PR #7263, opened 2017-09-28
Enable setting custom SSLParameters on Engine when using SniHandler

nmardiks:
Enable setting custom SSLParameters on Engine when using SniHandler (specifically required for use with UseCipherSuitesOrder)

Motivation:

Trying to use SniHandler with custom SSL Parameters being applied to the engine. Specifically, I need to setUseCipherSuitesOrder(true), which I could not find any other way of doing.
Due to when SniHandler runs and how, it is too late to apply in any other place outside of it, since once replace() had been called, the handshake had already started.

Modification:

Adding a simple setter to SniHandler which is being applied immediately after new engine is created and before it is being used (replace() called on pipeline)

Result:

One can set custom SSLParameters to be used by SniHandler.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
@@ -26,6 +26,8 @@ import io.netty.util.concurrent.Promise;
 import io.netty.util.internal.ObjectUtil;
 import io.netty.util.internal.PlatformDependent;
 
+import javax.net.ssl.SSLParameters;
+
 /**
  * <p>Enables <a href="https://tools.ietf.org/html/rfc3546#section-3.1">SNI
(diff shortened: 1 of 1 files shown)


normanmaurer:
@nmardiks I think this PR is not really needed as you can just wrap you `SslContext` with `DelegatingSslContext` and apply changes to the `SSLEngine` by implement `initEngine(...)`: https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/DelegatingSslContext.java#L97

Let me close this one for now and please reopen if you think there is an other issue here.

nmardiks:
Thanks, @normanmaurer makes perfect sense and looks like it would have solved my problem (I was on 4.0.36 and couldn't see this as it was only introduced on 4.0.38).

PR #7265, opened 2017-09-29
HttpPostMultipartRequestDecoder should decode header field parameters

dminkovsky:
Motivation:

I am receiving a `mutlipart/form_data` upload from a [Mailgun](https://www.mailgun.com/) webhook. This webhook used to send parts like this:

```
--74e78d11b0214bdcbc2f86491eeb4902
Content-Disposition: form-data; name="attachment-2"; filename="attached_ÑÐ°Ð¹Ð».txt"
Content-Type: text/plain
Content-Length: 32

This is the content of the file

--74e78d11b0214bdcbc2f86491eeb4902--
```

but now it posts parts like this:

```
--74e78d11b0214bdcbc2f86491eeb4902
Content-Disposition: form-data; name="attachment-2"; filename*=utf-8''attached_%D1%84%D0%B0%D0%B9%D0%BB.txt

This is the content of the file

--74e78d11b0214bdcbc2f86491eeb4902--
```
 
This new format uses field parameter encoding described in [RFC 5987](https://tools.ietf.org/html/rfc5987#section-3.2). More about this encoding can be found [here](https://stackoverflow.com/a/20592910/741970). 

Netty does not parse this format. The result is the filename is not decoded and the part is not parsed into a `FileUpload`.

Modification:

- Added failing test in `HttpPostRequestDecoderTest.java` and updated `HttpPostMultipartRequestDecoder.java`
- Refactored to please Netkins

Result:

Fixes #7265 (this):

- `HttpPostMultipartRequestDecoder ` identifies the RFC 5987 format and parses it.
- Previous functionality is retained.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
@@ -22,6 +22,7 @@ import io.netty.handler.codec.http.HttpHeaderNames;
 import io.netty.handler.codec.http.HttpHeaderValues;
 import io.netty.handler.codec.http.HttpRequest;
 import io.netty.handler.codec.http.LastHttpContent;
+import io.netty.handler.codec.http.QueryStringDecoder;
 import io.netty.handler.codec.http.multipart.HttpPostBodyUtil.SeekAheadOptimize;
 import io.netty.handler.codec.http.multipart.HttpPostBodyUtil.TransferEncodingMechanism;
 import io.netty.handler.codec.http.multipart.HttpPostRequestDecoder.EndOfDataDecoderException;
@@ -691,18 +692,7 @@ public class HttpPostMultipartRequestDecoder implements InterfaceHttpPostRequest
                         String[] values = contents[i].split("=", 2);
                         Attribute attribute;
                         try {
-                            String name = cleanString(values[0]);
-                            String value = values[1];
-
-                            // See http://www.w3.org/Protocols/rfc2616/rfc2616-sec19.html
-                            if (HttpHeaderValues.FILENAME.contentEquals(name)) {
-                                // filename value is quoted string so strip them
-                                value = value.substring(1, value.length() - 1);
-                            } else {
-                                // otherwise we need to clean the value
-                                value = cleanString(value);
-                            }
-                            attribute = factory.createAttribute(request, name, value);
+                            attribute = getContentDispositionAttribute(values);
                         } catch (NullPointerException e) {
                             throw new ErrorDataDecoderException(e);
                         } catch (IllegalArgumentException e) {
@@ -803,6 +793,38 @@ public class HttpPostMultipartRequestDecoder implements InterfaceHttpPostRequest
                 throw new ErrorDataDecoderException("Filename not found");
             }
         }
+    }
+
+    private static final String FILENAME_ENCODED = HttpHeaderValues.FILENAME.toString() + '*';
+
+    private Attribute getContentDispositionAttribute(String... values) {
+        String name = cleanString(values[0]);
+        String value = values[1];
+
+        // Filename can be token, quoted or encoded. See https://tools.ietf.org/html/rfc5987
+        if (HttpHeaderValues.FILENAME.contentEquals(name)) {
+            // Value is quoted or token. Strip if quoted:
+            int last = value.length() - 1;
+            if (last > 0 &&
+              value.charAt(0) == HttpConstants.DOUBLE_QUOTE &&
+              value.charAt(last) == HttpConstants.DOUBLE_QUOTE) {
+                value = value.substring(1, last);
+            }
+        } else if (FILENAME_ENCODED.equals(name)) {
+            try {
+                name = HttpHeaderValues.FILENAME.toString();
+                String[] split = value.split("'", 3);
+                value = QueryStringDecoder.decodeComponent(split[2], Charset.forName(split[0]));
+            } catch (ArrayIndexOutOfBoundsException e) {
+                 throw new ErrorDataDecoderException(e);
+            } catch (UnsupportedCharsetException e) {
+                throw new ErrorDataDecoderException(e);
+            }
+        } else {
+            // otherwise we need to clean the value
+            value = cleanString(value);
+        }
+        return factory.createAttribute(request, name, value);
     }
 
     /**
--- a/codec-http/src/test/java/io/netty/handler/codec/http/multipart/HttpPostRequestDecoderTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/multipart/HttpPostRequestDecoderTest.java
@@ -33,6 +33,7 @@ import io.netty.handler.codec.http.LastHttpContent;
 import io.netty.util.CharsetUtil;
 import org.junit.Test;
 
+import java.net.URLEncoder;
 import java.nio.charset.UnsupportedCharsetException;
 import java.util.Arrays;
 
@@ -489,6 +490,140 @@ public class HttpPostRequestDecoderTest {
         } finally {
             decoder.destroy();
             content.release();
+        }
+    }
+
+    // https://github.com/netty/netty/pull/7265
+    @Test
+    public void testDecodeContentDispositionFieldParameters() throws Exception {
+
+        final String boundary = "74e78d11b0214bdcbc2f86491eeb4902";
+
+        String encoding = "utf-8";
+        String filename = "attached_файл.txt";
+        String filenameEncoded = URLEncoder.encode(filename, encoding);
+
+        final String body = "--" + boundary + "\r\n" +
+          "Content-Disposition: form-data; name=\"file\"; filename*=" + encoding + "''" + filenameEncoded + "\r\n" +
+          "\r\n" +
+          "foo\r\n" +
+          "\r\n" +
+          "--" + boundary + "--";
+
+        final DefaultFullHttpRequest req = new DefaultFullHttpRequest(HttpVersion.HTTP_1_1,
+                                                                      HttpMethod.POST,
+                                                                      "http://localhost",
+                                                                      Unpooled.wrappedBuffer(body.getBytes()));
+
+        req.headers().add(HttpHeaderNames.CONTENT_TYPE, "multipart/form-data; boundary=" + boundary);
+        final DefaultHttpDataFactory inMemoryFactory = new DefaultHttpDataFactory(false);
+        final HttpPostRequestDecoder decoder = new HttpPostRequestDecoder(inMemoryFactory, req);
+        assertFalse(decoder.getBodyHttpDatas().isEmpty());
+        InterfaceHttpData part1 = decoder.getBodyHttpDatas().get(0);
+        assertTrue("the item should be a FileUpload", part1 instanceof FileUpload);
+        FileUpload fileUpload = (FileUpload) part1;
+        assertEquals("the filename should be decoded", filename, fileUpload.getFilename());
+        decoder.destroy();
+        req.release();
+    }
+
+    // https://github.com/netty/netty/pull/7265
+    @Test
+    public void testDecodeWithLanguageContentDispositionFieldParameters() throws Exception {
+
+        final String boundary = "74e78d11b0214bdcbc2f86491eeb4902";
+
+        String encoding = "utf-8";
+        String filename = "attached_файл.txt";
+        String language = "anything";
+        String filenameEncoded = URLEncoder.encode(filename, encoding);
+
+        final String body = "--" + boundary + "\r\n" +
+          "Content-Disposition: form-data; name=\"file\"; filename*=" +
+              encoding + "'" + language + "'" + filenameEncoded + "\r\n" +
+          "\r\n" +
+          "foo\r\n" +
+          "\r\n" +
+          "--" + boundary + "--";
+
+        final DefaultFullHttpRequest req = new DefaultFullHttpRequest(HttpVersion.HTTP_1_1,
+                                                                      HttpMethod.POST,
+                                                                      "http://localhost",
+                                                                      Unpooled.wrappedBuffer(body.getBytes()));
+
+        req.headers().add(HttpHeaderNames.CONTENT_TYPE, "multipart/form-data; boundary=" + boundary);
+        final DefaultHttpDataFactory inMemoryFactory = new DefaultHttpDataFactory(false);
+        final HttpPostRequestDecoder decoder = new HttpPostRequestDecoder(inMemoryFactory, req);
+        assertFalse(decoder.getBodyHttpDatas().isEmpty());
+        InterfaceHttpData part1 = decoder.getBodyHttpDatas().get(0);
+        assertTrue("the item should be a FileUpload", part1 instanceof FileUpload);
+        FileUpload fileUpload = (FileUpload) part1;
+        assertEquals("the filename should be decoded", filename, fileUpload.getFilename());
+        decoder.destroy();
+        req.release();
+    }
+
+    // https://github.com/netty/netty/pull/7265
+    @Test
+    public void testDecodeMalformedNotEncodedContentDispositionFieldParameters() throws Exception {
+
+        final String boundary = "74e78d11b0214bdcbc2f86491eeb4902";
+
+        final String body = "--" + boundary + "\r\n" +
+          "Content-Disposition: form-data; name=\"file\"; filename*=not-encoded\r\n" +
+          "\r\n" +
+          "foo\r\n" +
+          "\r\n" +
+          "--" + boundary + "--";
+
+        final DefaultFullHttpRequest req = new DefaultFullHttpRequest(HttpVersion.HTTP_1_1,
+                                                                      HttpMethod.POST,
+                                                                      "http://localhost",
+                                                                      Unpooled.wrappedBuffer(body.getBytes()));
+
+        req.headers().add(HttpHeaderNames.CONTENT_TYPE, "multipart/form-data; boundary=" + boundary);
+
+        final DefaultHttpDataFactory inMemoryFactory = new DefaultHttpDataFactory(false);
+
+        try {
+            new HttpPostRequestDecoder(inMemoryFactory, req);
+            fail("Was expecting an ErrorDataDecoderException");
+        } catch (HttpPostRequestDecoder.ErrorDataDecoderException e) {
+            assertTrue(e.getCause() instanceof ArrayIndexOutOfBoundsException);
+        } finally {
+            req.release();
+        }
+    }
+
+    // https://github.com/netty/netty/pull/7265
+    @Test
+    public void testDecodeMalformedBadCharsetContentDispositionFieldParameters() throws Exception {
+
+        final String boundary = "74e78d11b0214bdcbc2f86491eeb4902";
+
+        final String body = "--" + boundary + "\r\n" +
+          "Content-Disposition: form-data; name=\"file\"; filename*=not-a-charset''filename\r\n" +
+          "\r\n" +
+          "foo\r\n" +
+          "\r\n" +
+          "--" + boundary + "--";
+
+        final DefaultFullHttpRequest req = new DefaultFullHttpRequest(HttpVersion.HTTP_1_1,
+                                                                      HttpMethod.POST,
+                                                                      "http://localhost",
+                                                                      Unpooled.wrappedBuffer(body.getBytes()));
+
+        req.headers().add(HttpHeaderNames.CONTENT_TYPE, "multipart/form-data; boundary=" + boundary);
+
+        final DefaultHttpDataFactory inMemoryFactory = new DefaultHttpDataFactory(false);
+
+        try {
+            new HttpPostRequestDecoder(inMemoryFactory, req);
+            fail("Was expecting an ErrorDataDecoderException");
+        } catch (HttpPostRequestDecoder.ErrorDataDecoderException e) {
+            assertTrue(e.getCause() instanceof UnsupportedCharsetException);
+        } finally {
+            req.release();
         }
     }
 }


codec-http/src/test/java/io/netty/handler/codec/http/multipart/HttpPostRequestDecoderTest.java
```diff
@@ -491,4 +492,37 @@ public void testFormEncodeIncorrect() throws Exception {
...
+                                                                      "http://localhost",
+                                                                      Unpooled.wrappedBuffer(body.getBytes()));
+
+        req.headers().add(HttpHeaderNames.CONTENT_TYPE, "multipart/form-data; boundary=" + boundary);
+        final DefaultHttpDataFactory inMemoryFactory = new DefaultHttpDataFactory(false);
+        final HttpPostRequestDecoder decoder = new HttpPostRequestDecoder(inMemoryFactory, req);
+        assertFalse(decoder.getBodyHttpDatas().isEmpty());
+        InterfaceHttpData part1 = decoder.getBodyHttpDatas().get(0);
+        assertTrue("the item should be a FileUpload", part1 instanceof FileUpload);
+        FileUpload fileUpload = (FileUpload) part1;
```
dminkovsky: CCE here before this patch; the part is present but it's an `Attribute`.

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
```diff
@@ -694,10 +695,22 @@ private InterfaceHttpData findMultipartDisposition() {
                             String name = cleanString(values[0]);
                             String value = values[1];
 
+                            boolean shouldDecode = false;
+                            int last = name.length() - 1;
+                            if (name.charAt(last) == '*') {
```
netkins: MAJOR Refactor this code to not nest more than 5 if/for/while/switch/try statements. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS134)

netkins: MAJOR Refactor this code to not nest more than 5 if/for/while/switch/try statements. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS134)

dminkovsky:
First commit on this patch PR shows the patch. 

Second commit was required to please netkins. I was already 5 indents in.

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
```diff
@@ -805,6 +795,35 @@ private InterfaceHttpData findMultipartDisposition() {
         }
     }
 
+    private Attribute getContentDispositionAttribute(String[] values) {
```
netkins: MINOR Consider using varargs for methods or constructors which take an array the last parameter. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUseVarargs)

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
```diff
@@ -805,6 +795,35 @@ private InterfaceHttpData findMultipartDisposition() {
...
+            name = name.substring(0, last);
+        }
+
+        // See http://www.w3.org/Protocols/rfc2616/rfc2616-sec19.html
+        if (HttpHeaderValues.FILENAME.contentEquals(name)) {
+            // filename value is quoted string so strip them
+            if (!shouldDecode) {
+                value = value.substring(1, value.length() - 1);
+            } else {
+                String[] split = value.split("''", 2);
```
fenik17: Consider replacing `String#split` with use precompiled `Pattern` constant or a `value.indexOf("''")`.

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
```diff
@@ -805,6 +795,35 @@ private InterfaceHttpData findMultipartDisposition() {
         }
     }
 
+    private Attribute getContentDispositionAttribute(String... values) {
+        Attribute attribute;
```
fenik17: Why is it necessary to declare the `attribute` here?

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
```diff
@@ -805,6 +795,35 @@ private InterfaceHttpData findMultipartDisposition() {
...
+        int last = name.length() - 1;
+        if (name.charAt(last) == '*') {
+            shouldDecode = true;
+            name = name.substring(0, last);
+        }
+
+        // See http://www.w3.org/Protocols/rfc2616/rfc2616-sec19.html
+        if (HttpHeaderValues.FILENAME.contentEquals(name)) {
+            // filename value is quoted string so strip them
+            if (!shouldDecode) {
```
fenik17: IIUC the RFC 5987 prescribe decoding not only `filename` header value? Otherwise we should not cut `*` from name..

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
dminkovsky: Thank you. I don't have much experience in performant Java string work, so I was following the lead of existing code a [few lines up](https://github.com/dminkovsky/netty/blob/149916d052180dd2b0e0e98598f349454ea200ba/codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java#L707) where you have `value.split("=", 2)`. Does the compiler/runtime optimize this case because it's a unit-length string and therefore effectively a char?

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
dminkovsky: That is how it was before my patch on [4.1](https://github.com/dminkovsky/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java#L708). For the actual test/patch, please see my first commit (https://github.com/netty/netty/pull/7265/commits/e4a98530078894b1132a929fe007c7e7b854a9d7). My second commit, which this comes from, was done because Netkins complained I had gone above 5 indents and should refactor. I would obviously rather not include this commit in this PR.

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
dminkovsky: I thought about that too. But I'm not sure that's the case (what other headers are there? they aren't being tested for flow control in this code). I decided to just cover this case and establish how it might be done if someone encountered this problem with other headers.

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
fenik17: Yep. For one-char string a `String.split` in OpenJDK [doesn't use `Pattern`](http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/lang/String.java#l2324). But the string `"''"` has two chars.

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
fenik17: Ok. But now there is no need for this. Just use fast return without variable:
```java
    return factory.createAttribute(request, name, value);
```

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
fenik17: I mean, if we want apply new RFC only for `filename` header, we should not change the `name` attribute for other headers. But now you are doing this: `name = name.substring(0, last);`

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
dminkovsky: Ah okay thanks. I opted for the precompiled pattern because it seemed simpler.

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
dminkovsky: Ah yes that makes sense. I think I avoid this now. Could add a test for it...

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
dminkovsky: Yes of course. Fixed this.

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
```diff
@@ -805,6 +796,35 @@ private InterfaceHttpData findMultipartDisposition() {
         }
     }
 
+    private static Pattern doubleSingleQuote = Pattern.compile("''");
```
fenik17: `final`

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
```diff
@@ -805,6 +796,35 @@ private InterfaceHttpData findMultipartDisposition() {
...
+
+    private Attribute getContentDispositionAttribute(String... values) {
+        String name = cleanString(values[0]);
+        String value = values[1];
+
+        boolean shouldDecode = false;
+        int last = name.length() - 1;
+        if (name.charAt(last) == '*' && HttpHeaderValues.FILENAME.contentEquals(name.substring(0, last))) {
```
fenik17: Would be better to reduce the possible allocations using `filename*` constant. Something like this:
```java
    private static final String FILENAME_ENCODED = HttpHeaderValues.FILENAME.toString() + '*';

    private Attribute getContentDispositionAttribute(String... values) {
        String name = cleanString(values[0]);
        String value = values[1];

        // See http://www.w3.org/Protocols/rfc2616/rfc2616-sec19.html
        if (HttpHeaderValues.FILENAME.contentEquals(name)) {
            // filename value is quoted string so strip them
            value = value.substring(1, value.length() - 1);
        } else if (FILENAME_ENCODED.equals(name)) {
            // filename value is encoded. See https://tools.ietf.org/html/rfc5987
            name = name.substring(0, name.length() - 1);
            String[] split = doubleSingleQuote.split(value, 2);
            value = QueryStringDecoder.decodeComponent(split[1], Charset.forName(split[0]));
        } else {
            // otherwise we need to clean the value
            value = cleanString(value);
        }
        return factory.createAttribute(request, name, value);
    }
```

fenik17:
@dminkovsky RFC says that `ext-value` can optional contains language information:

`foo: bar; title*=iso-8859-1'en'%A3%20rates`

Maybe it makes sense to support for this too? At least, take this into account when parsing the value.

dminkovsky:
Yes, I thought about that too, but what would we do with it?

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
dminkovsky: Thanks. Pushed this version.

fenik17:
> Yes, I thought about that, but what would we do with it?

Ignoring? Just split value through single quote `"'"`.
```java
String[] split = value.split("'", 3);
value = QueryStringDecoder.decodeComponent(split[2], Charset.forName(split[0]))
```

dminkovsky:
I’m sorry that’s not what I meant. I meant: if we got the language code
value—like you said by splitting a different way—then what would we do with
that value?

пн, 2 окт. 2017 г. в 1:12, Nikolay Fedorovskikh <notifications@github.com>:

fenik17:
I suggest not doing anything with the language field. Just skip this.

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
```diff
@@ -805,6 +796,29 @@ private InterfaceHttpData findMultipartDisposition() {
...
+    private static final String FILENAME_ENCODED = HttpHeaderValues.FILENAME.toString() + '*';
+
+    private Attribute getContentDispositionAttribute(String... values) {
+        String name = cleanString(values[0]);
+        String value = values[1];
+
+        // See http://www.w3.org/Protocols/rfc2616/rfc2616-sec19.html
+        if (HttpHeaderValues.FILENAME.contentEquals(name)) {
+            // filename value is quoted string so strip them
+            value = value.substring(1, value.length() - 1);
```
carl-mastrangelo: consider checking that the first and last chars are actually quotes.

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
```diff
@@ -805,6 +796,29 @@ private InterfaceHttpData findMultipartDisposition() {
...
+        } else if (FILENAME_ENCODED.equals(name)) {
+            // filename value is encoded. See https://tools.ietf.org/html/rfc5987
+            name = name.substring(0, name.length() - 1);
+            String[] split = DOUBLE_SINGLE_QUOTE.split(value, 2);
+            value = QueryStringDecoder.decodeComponent(split[1], Charset.forName(split[0]));
```
carl-mastrangelo: Here too, you might consider check that the array size is 2 so that you don't throw an NPE.

codec-http/src/test/java/io/netty/handler/codec/http/multipart/HttpPostRequestDecoderTest.java
```diff
@@ -491,4 +492,37 @@ public void testFormEncodeIncorrect() throws Exception {
             content.release();
         }
     }
+
+    // https://github.com/netty/netty/pull/7265
+    @Test
+    public void testDecodeContentDispositionFieldParameters() throws Exception {
```
carl-mastrangelo: add a negative case?  perhaps a malformed header?

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
dminkovsky: this was here before this PR. should i modify this behavior in this PR?

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
dminkovsky: Considered this, but if not an NPE, then what? There are other spots in the code around this PR that split and don't check.

codec-http/src/test/java/io/netty/handler/codec/http/multipart/HttpPostRequestDecoderTest.java
dminkovsky: Yes, makes sense. Will do.

dminkovsky:
@fenik17 just looked again at your [previous comment](https://github.com/netty/netty/pull/7265#issuecomment-333442257) and yes, of course, it's much better this way. thank you for your reviews!

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
```diff
@@ -805,6 +796,28 @@ private InterfaceHttpData findMultipartDisposition() {
         }
     }
 
+    private static final String FILENAME_ENCODED = HttpHeaderValues.FILENAME.toString() + '*';
```
netkins: MAJOR Remove this unused "DOUBLE_SINGLE_QUOTE" private field. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1068)

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
dminkovsky: Adding this check too.

codec-http/src/test/java/io/netty/handler/codec/http/multipart/HttpPostRequestDecoderTest.java
dminkovsky: Added a few tests.

codec-http/src/test/java/io/netty/handler/codec/http/multipart/HttpPostRequestDecoderTest.java
```diff
@@ -491,4 +492,136 @@ public void testFormEncodeIncorrect() throws Exception {
...
+          "Content-Disposition: form-data; name=\"file\"; filename*=" + encoding + "''" + filenameEncoded + "\r\n" +
+          "\r\n" +
+          "foo\r\n" +
+          "\r\n" +
+          "--" + boundary + "--";
+
+        final DefaultFullHttpRequest req = new DefaultFullHttpRequest(HttpVersion.HTTP_1_1,
+                                                                      HttpMethod.POST,
+                                                                      "http://localhost",
+                                                                      Unpooled.wrappedBuffer(body.getBytes()));
```
carl-mastrangelo: @normanmaurer do your tests require you to deref all buffers?

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
```diff
@@ -805,6 +795,37 @@ private InterfaceHttpData findMultipartDisposition() {
...
+
+    private Attribute getContentDispositionAttribute(String... values) {
+        String name = cleanString(values[0]);
+        String value = values[1];
+
+        // See http://www.w3.org/Protocols/rfc2616/rfc2616-sec19.html
+        if (HttpHeaderValues.FILENAME.contentEquals(name)) {
+            // filename value is quoted string so strip them
+            int last = value.length() - 1;
+            if (value.charAt(0) == HttpConstants.DOUBLE_QUOTE && value.charAt(last) == HttpConstants.DOUBLE_QUOTE) {
```
fenik17: ```if (last > 0 && ...```

dminkovsky: Ah yes, thank you
```
value         = token / quoted-string
```
https://tools.ietf.org/html/rfc5987#section-3.2.1

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
```diff
@@ -805,6 +795,38 @@ private InterfaceHttpData findMultipartDisposition() {
...
+            // Value is quoted or token. Strip if quoted:
+            int last = value.length() - 1;
+            if (last > 0 &&
+              value.charAt(0) == HttpConstants.DOUBLE_QUOTE &&
+              value.charAt(last) == HttpConstants.DOUBLE_QUOTE) {
+                value = value.substring(1, last);
+            }
+        } else if (FILENAME_ENCODED.equals(name)) {
+            try {
+                name = name.substring(0, name.length() - 1);
```
fenik17: We can avoid an extra allocation:
```java
name = HttpHeaderValues.FILENAME.toString();
```

dminkovsky:
By the way, could this make 4.1.17? I see only a 4.0 tag. I am wondering because I am on 4.1

normanmaurer:
Everything that is merged into 4.0 will also be merged Into 4.1

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
dminkovsky: Yes, makes sense. Thanks.

dminkovsky:
Good to hear. Thank you.

codec-http/src/test/java/io/netty/handler/codec/http/multipart/HttpPostRequestDecoderTest.java
normanmaurer: yep I think we should call req.release()

normanmaurer: @dminkovsky please address this

dminkovsky: @normanmaurer done. Please have a look. Thank you.

normanmaurer:
Cherry-picked into 4.1 (8aeba78ecc9c9c3806ca729120c57e521f3d0501) and 4.0 (b12dcb0627b015a4f244771037062ee9de1a8cf3).

@dminkovsky thanks!

fenik17:
@normanmaurer this is not pushed into 4.0?

normanmaurer:
@fenik17 somehow I did not... Thanks for pinging. Cherry-picked as 82b710376

Issue #7266, opened 2017-09-30
Unable to use Netty native transports in modularized Java 9 projects


doom369:
Due to automatic name generation in Java 9 it is impossible to use Netty native transports - Epoll, KQueue in modularized Java 9 projects. For example, automatic module name for the epoll transport would be : 

`netty.transport.native.epoll`

And `native` is the reserved keyword which doesn't allow the module to be exported to any other module. I created also [question in SO](https://stackoverflow.com/questions/46501388/how-to-deal-with-native-keyword-in-auto-generated-module-names-in-java-9) in case there any workarounds.

However, there is actually very simple fix without moving the whole netty to java 9 that netty can apply right now.

We can add to manifest next line and it will solve the issue for Java 9. I checked it already and it works.

`<Automatic-Module-Name>netty.transport.epoll</Automatic-Module-Name>`

WDYT?

johnou:
@doom369 ptal https://github.com/netty/netty/issues/7218

doom369:
@johnou thanks. I saw that. My PR just fixes the concrete issue with epoll and kqueue. I think there is no need right now to give precise names for other modules until netty will provide `module-info.java` files.

zyxist:
@doom369 -> actually, there is a need to give them now: http://blog.joda.org/2017/05/java-se-9-jpms-automatic-modules.html

I'll remove the `native` word from my PR.

Scottmitch:
Wasn't this fixed recently?

normanmaurer:
@Scottmitch yep.. good catch

PR #7268, opened 2017-09-30
Added automatic module name to the jar manifest for native transports

doom369:
Motivation:

Right now it is impossible to add native transports to modularized java 9 projects as the automatic generated name contains reserved keyword `native`.

Fix for #7266.

Modification:

Added predefined module names for native transports (without `native` keyword) to jar manifest to avoid automatic modules name:

```
netty.transport.epoll
netty.transport.kqueue
```

Result:

Now native transports could be added to modularized java 9 projects.


normanmaurer:
@doom369 isn't this a duplicate of https://github.com/netty/netty/pull/7245 ?

doom369:
@normanmaurer it depends on your decision. Fill free to close.

doom369:
Fixes applied in https://github.com/netty/netty/pull/7245

PR #7269, opened 2017-09-30
Remove unnecessary instantiation of HttpResponseStatus

fenik17:
Motivation:
- Inside of the `HttpResponseStatus#equals` checks only status code. No need to create new instance of `HttpResponseStatus` for comparison with response status.
- RFC says: `the HTTP version and reason phrase aren't important` [1].

Modifications:
Use comparison by status code without creating new `HttpResponseStatus`.

Result:
Less allocations, more clear code.

[1] https://tools.ietf.org/html/draft-ietf-hybi-thewebsocketprotocol-00

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker00.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker00.java
@@ -24,7 +24,6 @@ import io.netty.handler.codec.http.HttpHeaderNames;
 import io.netty.handler.codec.http.HttpHeaderValues;
 import io.netty.handler.codec.http.HttpHeaders;
 import io.netty.handler.codec.http.HttpMethod;
-import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http.HttpVersion;
 import io.netty.util.AsciiString;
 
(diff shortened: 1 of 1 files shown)


normanmaurer:
Cherry-picked into 4.1 (78786e6052d108c29b407aa93a505692d0a8fe20) and 4.0 (272f26f68c).

@fenik17 thanks!

slandelle:
Why not use `!response.status().equals(HttpResponseStatus.SWITCHING_PROTOCOLS)`???

fenik17:
@slandelle to emphasize that only code matters and not to be confused between "Switching Protocols" and "WebSocket Protocol Handshake".

slandelle:
@fenik17 Sorry, I don't get you. The meaning of this code is "assert the response status code is the Switching Protocol one" (which is actually bad as it doesn't allow things like redirects and authorization required, but that's another issue).

fenik17:
@slandelle I just wanted to emphasize that the reason phrase is not important. Only the response code is important. From RFC:

>            HTTP/1.1 101 WebSocket Protocol Handshake
>
>        This line may be sent differently if necessary, but must match
>        the Status-Line production defined in the HTTP specification,
>        with the Status-Code having the value 101.

Without introspect into the implementation of the `HttpResponseStatus.equals()`, this is not clear.
Feel free to open PR with the variant that you think is right.

slandelle:
Will do. IMHO:
* "RFC says: the HTTP version and reason phrase aren't important [1].": Absolutely, `HttpResponseStatus#equals` exactly implements that.
* Don't use a hardcoded 101 value, use the intended constant instead

PR #7271, opened 2017-10-02
Use predefined HttpResponseStatus constant instead of hardcoded 101 code

slandelle:
Motivation:

#7269 removed an unnecessary instanciation for verifying WebSocket
handshake status code.
But it uses a hardcoded status code value for 101 instead of using the
intended `HttpResponseStatus#SWITCHING_PROTOCOLS` constant.

Modidication:

Compare actual `HttpResponseStatus` against predefined constant. Note
that `HttpResponseStatus#equals` is implemented in respect with the RFC
(only honor code, not text) so it’s intended to be used this way.

Result:

Cleaner code, use intended constant instead of hard coded value.


normanmaurer:
Cherry-picked into 4.1 (940f15f0d2cbdbb05397832a0b3e8ac4d5dd1420) and 4.0 (9b34ec6784)

@slandelle thanks for the cleanup

PR #7276, opened 2017-10-04
Do not treat errors as decoder exception

jasontedor:
Motivation: Today when Netty encounters a general error while decoding it treats this as a decoder exception. However, for fatal causes this should not be treated as such, instead the fatal error should be carried up the stack without the callee having to unwind causes.

Modifications: Instead of translating any error to a decoder exception, we let those unwind out the stack (note that finally blocks still execute).

Result: Fatal errors will not be treated as innocent decoder exceptions.

Relates #6096

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
@@ -265,8 +265,8 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
                 callDecode(ctx, cumulation, out);
             } catch (DecoderException e) {
                 throw e;
-            } catch (Throwable t) {
-                throw new DecoderException(t);
+            } catch (Exception e) {
+                throw new DecoderException(e);
             } finally {
                 if (cumulation != null && !cumulation.isReadable()) {
                     numReads = 0;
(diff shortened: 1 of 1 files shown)


normanmaurer:
@jasontedor I like the change and imho its a good start in the right direction. Did you also check all the other decoder / encoders to ensure the same is done there ?

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -265,8 +265,8 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
                 callDecode(ctx, cumulation, out);
             } catch (DecoderException e) {
                 throw e;
-            } catch (Throwable t) {
-                throw new DecoderException(t);
+            } catch (Exception e) {
+                throw new DecoderException(e);
```
carl-mastrangelo: optional: maybe don't wrap RuntimeExceptions either?

normanmaurer: I think we should just handle `Error` special here imho

jasontedor:
> Did you also check all the other decoder / encoders to ensure the same is done there ?

No, this specific one is hurting us right now as some of our users are seeing decoder exceptions hiding out of memory errors and I want to get a fix in quickly. I can take a more careful look at the others soon.

jasontedor:
@normanmaurer I've now checked all the other places and I have another patch ready to push. My preference would be to get this PR in (since it's ready) and then I can immediately open a follow-up. What do you think?

normanmaurer:
@jasontedor works for me...

normanmaurer:
Cherry-picked into 4.1 (5eca326c3568566f334b31b18f7866af19a989c8) and 4.0 (4ad6ed17ebb62959c382b8d9372372e2911c68f6) . @jasontedor thanks

Issue #7230, opened 2017-09-19
Method setHandshakeTimeout of SslHandler does not work in server side


BillyYccc:
### Expected behavior
I want to enable `setHandshakeTimeout()` in server side. I call the method setHandshakeTimeout() like this `sslHandler.setHandshakeTimeout(1,TimeUnit.MILLISECONDS);` The value of HandshakeTimeout is set to a very small value so that the client and server can not complete the handshake since the handshake is timeout. 

### Actual behavior
But it can not work when I use this in server.The connection can still be established between server and client. However when I do the same in client, the SSL connection can not be established , this is what I expect to see. 

### Steps to reproduce
1. call the `sslHandler.setHandshakeTimeoutMillis(1);` or `sslHandler.setHandshakeTimeout(1,TimeUnit.MILLISECONDS);` in server
2. do connect and SSL connection can be successfully established
3. do 1 and 2 again in client side, and SSL connection can not be established

### Minimal yet complete reproducer code (or URL to code)
You may use [SecureChat](http://netty.io/4.1/xref/io/netty/example/securechat/package-summary.html) example or any other SSL example to reproduce

### Netty version
4.1.15.Final

### JVM version (e.g. `java -version`)
java version "1.8.0_131"
Java(TM) SE Runtime Environment (build 1.8.0_131-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.131-b11, mixed mode)

normanmaurer:
@BillyYccc please check https://github.com/netty/netty/pull/7277

normanmaurer:
Fixed

PR #7277, opened 2017-10-05
SslHandler.setHandshakeTimeout*(...) should also been enforced on the…

normanmaurer:
… server side.

Motivation:

We should also enforce the handshake timeout on the server-side to allow closing connections which will not finish the handshake in an expected amount of time.

Modifications:

- Enforce the timeout on the server and client side
- Add unit test.

Result:

Fixes [#7230].

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -1536,11 +1536,15 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
     public void handlerAdded(final ChannelHandlerContext ctx) throws Exception {
         this.ctx = ctx;
 
-        if (ctx.channel().isActive() && engine.getUseClientMode()) {
-            // Begin the initial handshake.
-            // channelActive() event has been fired already, which means this.channelActive() will
-            // not be invoked. We have to initialize here instead.
-            handshake(null);
+        if (ctx.channel().isActive()) {
+            if (engine.getUseClientMode()) {
+                // Begin the initial handshake.
+                // channelActive() event has been fired already, which means this.channelActive() will
+                // not be invoked. We have to initialize here instead.
+                handshake(null);
+            } else {
+                applyHandshakeTimeout(null);
+            }
         } else {
             // channelActive() event has not been fired yet.  this.channelOpen() will be invoked
             // and initialization will occur there.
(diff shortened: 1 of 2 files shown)


handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
```diff
@@ -85,6 +85,36 @@
...
+    private static void testHandshakeTimeout(boolean client) throws Exception {
+        SSLEngine engine = SSLContext.getDefault().createSSLEngine();
+        engine.setUseClientMode(client);
+        SslHandler handler = new SslHandler(engine);
+        handler.setHandshakeTimeoutMillis(1);
+
+        EmbeddedChannel ch = new EmbeddedChannel(handler);
+        try {
+            while (!handler.handshakeFuture().isDone()) {
+                Thread.sleep(10);
```
netkins: CRITICAL Remove this use of "Thread.sleep()". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2925)

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -32,6 +32,7 @@
 import io.netty.channel.ChannelPipeline;
 import io.netty.channel.ChannelPromise;
 import io.netty.channel.ChannelPromiseNotifier;
+import io.netty.channel.embedded.EmbeddedChannel;
```
netkins: MINOR Remove this unused import 'io.netty.channel.embedded.EmbeddedChannel'. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
normanmaurer: This is needed...

johnou:
@normanmaurer looks like the default handshake timeout is 10 seconds which will now be applied to server mode?

normanmaurer:
@johnou yep... do you think this is too aggressive ?

johnou:
@normanmaurer unfortunately I don't have any metrics for handshakes.. I will discuss it with my team and get back to you.

johnou:
@normanmaurer we are fine with the 10s timeout, heads up @rkapsi

rkapsi:
@johnou @normanmaurer 10s timeout for default is fine for us (and happens to be our configured value). Thx for the heads up @johnou.

rkapsi:
Maybe useful in this context... We observe the p95 time for successful handshakes around 1-1.5 seconds and p99 around 3-5 seconds.

normanmaurer:
@ejona86 @louiscryan any concerns with this change?

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
ejona86: Swapping from `isDone()` to `await(10, MILLISECONDS)` is probably better.

ejona86:
Seems fine to me.

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
normanmaurer: @ejona86 actually this will not work as then we will try to call await() on the EventLoop thread which will then throw an exception.

normanmaurer:
Cherry-picked into 4.1 (521e87984dcff49d556a5696ed5de0b72a8a1797) and 4.0 (aa3b5c0055d77a4b7f5c99585c71dbfe66b1f577)

PR #7278, opened 2017-10-05
Ensure setting / getting the traffic class on an ipv4 only system wor…

normanmaurer:
…ks when using the native transport.

Motivation:

We tried to set IPV6 opts on an ipv4 only system and so failed to set / get the traffic opts. This resulted in a test-error when trying to compile netty on ipv4 only systems.

Modifications:

Use the correct opts depending on if the system is ipv4 only or not.

Result:

Be able to build and use on ipv4 only systems.

### DIFF
--- a/transport-native-unix-common/src/main/c/netty_unix_socket.c
+++ b/transport-native-unix-common/src/main/c/netty_unix_socket.c
@@ -755,18 +755,26 @@ static void netty_unix_socket_setSoLinger(JNIEnv* env, jclass clazz, jint fd, ji
 }
 
 static void netty_unix_socket_setTrafficClass(JNIEnv* env, jclass clazz, jint fd, jint optval) {
-    /* Try to set the ipv6 equivalent, but don't throw if this is an ipv4 only socket. */
-    int rc = netty_unix_socket_setOption0(fd, IPPROTO_IPV6, IPV6_TCLASS, &optval, sizeof(optval));
-    if (rc < 0 && errno != ENOPROTOOPT) {
-        netty_unix_socket_setOptionHandleError(env, errno);
-    }
+    if (socketType == AF_INET6) {
+        // This call will put an exception on the stack to be processed once the JNI calls completes if
+        // setsockopt failed and return a negative value.
+        int rc = netty_unix_socket_setOption(env, fd, IPPROTO_IPV6, IPV6_TCLASS, &optval, sizeof(optval));
 
-    /* Linux allows both ipv4 and ipv6 families to be set */
+        if (rc >= 0) {
+/* Linux allows both ipv4 and ipv6 families to be set */
 #ifdef __linux__
-      else {
+            // Previous call successful now try to set also for ipv4
+            if (netty_unix_socket_setOption0(fd, IPPROTO_IP, IP_TOS, &optval, sizeof(optval)) == -1) {
+                if (errno != ENOPROTOOPT) {
+                    // throw exception
+                    netty_unix_socket_setOptionHandleError(env, errno);
+                }
+            }
+#endif
+        }
+    } else {
         netty_unix_socket_setOption(env, fd, IPPROTO_IP, IP_TOS, &optval, sizeof(optval));
     }
-#endif
 }
 
 static jint netty_unix_socket_isKeepAlive(JNIEnv* env, jclass clazz, jint fd) {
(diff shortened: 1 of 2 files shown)


normanmaurer:
@Scottmitch this was the cause of the build failure on docker (which is ipv4 only).

normanmaurer:
@carl-mastrangelo ping

transport-native-unix-common/src/main/c/netty_unix_socket.c
```diff
@@ -755,18 +755,16 @@ static void netty_unix_socket_setSoLinger(JNIEnv* env, jclass clazz, jint fd, ji
 }
 
 static void netty_unix_socket_setTrafficClass(JNIEnv* env, jclass clazz, jint fd, jint optval) {
-    /* Try to set the ipv6 equivalent, but don't throw if this is an ipv4 only socket. */
-    int rc = netty_unix_socket_setOption0(fd, IPPROTO_IPV6, IPV6_TCLASS, &optval, sizeof(optval));
-    if (rc < 0 && errno != ENOPROTOOPT) {
-        netty_unix_socket_setOptionHandleError(env, errno);
-    }
+    if (socketType == AF_INET6) {
+        netty_unix_socket_setOption(env, fd, IPPROTO_IPV6, IPV6_TCLASS, &optval, sizeof(optval));
```
pmarks-net: I'm concerned that this change deletes *all* error handling for setTrafficClass().  The simplest approach for the `AF_INET6` branch is to try both, and throw an error if both fail, e.g.:

    const int rc6 = netty_unix_socket_setOption(env, fd, IPPROTO_IPV6, IPV6_TCLASS, &optval, sizeof(optval));
    const int rc6_errno = rc6 < 0 ? errno : 0;
    const int rc4 = netty_unix_socket_setOption(env, fd, IPPROTO_IP, IP_TOS, &optval, sizeof(optval));
    const int rc4_errno = rc4 < 0 ? errno : 0;
    if (rc6 < 0 && rc4 < 0) {
        // some logic to figure out which errno is more relevant.
    }

This removes the `__linux__` special case, assuming that other platforms provide a reasonable errno.

transport-native-unix-common/src/main/c/netty_unix_socket.c
```diff
@@ -814,15 +812,23 @@ static jint netty_unix_socket_getSoLinger(JNIEnv* env, jclass clazz, jint fd) {
...
     int optval;
-    if (netty_unix_socket_getOption0(fd, IPPROTO_IPV6, IPV6_TCLASS, &optval, sizeof(optval)) == -1) {
-        if (errno != ENOPROTOOPT || netty_unix_socket_getOption0(fd, IPPROTO_IP, IP_TOS, &optval, sizeof(optval)) == -1) {
-            netty_unix_socket_getOptionHandleError(env, errno);
-            return -1;
+    if (socketType == AF_INET6) {
+        if (netty_unix_socket_getOption0(fd, IPPROTO_IPV6, IPV6_TCLASS, &optval, sizeof(optval)) == -1) {
+            if (errno == ENOPROTOOPT) {
+                if (netty_unix_socket_getOption(env, fd, IPPROTO_IP, IP_TOS, &optval, sizeof(optval)) == -1) {
+                    return -1;
```
pmarks-net: Shouldn't this call `getOptionHandleError()` before *every* `return -1`?

transport-native-unix-common/src/main/c/netty_unix_socket.c
```diff
@@ -755,18 +755,16 @@ static void netty_unix_socket_setSoLinger(JNIEnv* env, jclass clazz, jint fd, ji
...
-        netty_unix_socket_setOptionHandleError(env, errno);
-    }
+    if (socketType == AF_INET6) {
+        netty_unix_socket_setOption(env, fd, IPPROTO_IPV6, IPV6_TCLASS, &optval, sizeof(optval));
 
-    /* Linux allows both ipv4 and ipv6 families to be set */
+        /* Linux allows both ipv4 and ipv6 families to be set */
 #ifdef __linux__
-      else {
         netty_unix_socket_setOption(env, fd, IPPROTO_IP, IP_TOS, &optval, sizeof(optval));
```
carl-mastrangelo: this might be confusing.  If the ipv6 call succeeds, but the v4 call fails, is the socket still usable?

transport-native-unix-common/src/main/c/netty_unix_socket.c
normanmaurer: @pmarks-net no...  `getOptionHandleError` is only needed when calling `netty_unix_socket_getOption0` (in which case we do it in this method, but not when using  ´netty_unix_socket_getOption´ (no trailing `0`).

transport-native-unix-common/src/main/c/netty_unix_socket.c
normanmaurer: @pmarks-net maybe I am missing something but as far as i can tell it does correctly handle errors. Please note we call `netty_unix_socket_setOption` and NOT `netty_unix_socket_setOption0`.

normanmaurer:
@carl-mastrangelo @pmarks-net PTAL again

normanmaurer:
@pmarks-net like you said if this is a problem is existed before and we can address it in another PR once we are sure. Thanks for the review again!

normanmaurer:
Cherry-picked into 4.1 (d8e187ff2c30ce75d610e0079164841d919d839e)

PR #7279, opened 2017-10-05
Do not treat errors as decoder exception (redux)

jasontedor:
Motivation: Today when Netty encounters a general error while decoding it treats this as a decoder exception. However, for fatal causes this should not be treated as such, instead the fatal error should be carried up the stack without the callee having to unwind causes. This was probably done for byte to byte message decoder but is now done for all decoders.

Modifications: Instead of translating any error to a decoder exception, we let those unwind out the stack (note that finally blocks still execute) except in places where an event needs to fire where we fire with the error instead of wrapping in a decoder exception.

Result: Fatal errors will not be treated as innocent decoder exceptions.

Relates #6096

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/ReplayingDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/ReplayingDecoder.java
@@ -418,7 +418,7 @@ public abstract class ReplayingDecoder<S> extends ByteToMessageDecoder {
             }
         } catch (DecoderException e) {
             throw e;
-        } catch (Throwable cause) {
+        } catch (Exception cause) {
             throw new DecoderException(cause);
         }
     }
--- a/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
@@ -100,7 +100,11 @@ public class SniHandler extends AbstractSniHandler<SslContext> {
     protected final void onLookupComplete(ChannelHandlerContext ctx,
                                           String hostname, Future<SslContext> future) throws Exception {
         if (!future.isSuccess()) {
-            throw new DecoderException("failed to get the SslContext for " + hostname, future.cause());
+            final Throwable cause = future.cause();
+            if (cause instanceof Error) {
+                throw (Error) cause;
+            }
+            throw new DecoderException("failed to get the SslContext for " + hostname, cause);
         }
 
         SslContext sslContext = future.getNow();
--- a/codec-haproxy/src/main/java/io/netty/handler/codec/haproxy/HAProxyMessageDecoder.java
+++ b/codec-haproxy/src/main/java/io/netty/handler/codec/haproxy/HAProxyMessageDecoder.java
@@ -355,16 +355,16 @@ public class HAProxyMessageDecoder extends ByteToMessageDecoder {
         fail(ctx, "header length (" + length + ") exceeds the allowed maximum (" + maxLength + ')', null);
     }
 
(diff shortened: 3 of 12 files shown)


jasontedor:
This is a follow up to #7276.

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -100,7 +100,11 @@ public SslContext sslContext() {
     protected final void onLookupComplete(ChannelHandlerContext ctx,
                                           String hostname, Future<SslContext> future) throws Exception {
         if (!future.isSuccess()) {
-            throw new DecoderException("failed to get the SslContext for " + hostname, future.cause());
+            if (future.cause() instanceof Error) {
```
normanmaurer: @jasontedor nit: Can you store the cause in a temp variable as we access it multiple times and each time this is a volatile access.

jasontedor: I've made this change, squashed, and force pushed.

codec/src/main/java/io/netty/handler/codec/ReplayingDecoder.java
```diff
@@ -418,7 +418,7 @@ protected void callDecode(ChannelHandlerContext ctx, ByteBuf in, List<Object> ou
             }
         } catch (DecoderException e) {
             throw e;
-        } catch (Throwable cause) {
+        } catch (Exception cause) {
```
johnou: based on the javadoc of ReplayingDecoder I think we might need special handling here like we now have for Sni.. https://github.com/jasontedor/netty/blob/8ad05f00703ed42480e4b83c5d550dce68259b60/codec/src/main/java/io/netty/handler/codec/ReplayingDecoder.java#L77

```
* {@link ReplayingDecoder} passes a specialized {@link ByteBuf}
 * implementation which throws an {@link Error} of certain type when there's not
 * enough data in the buffer.
```

johnou: @jasontedor https://github.com/netty/netty/blob/eb7f751ba519cbcab47d640cd18757f09d077b55/common/src/main/java/io/netty/util/Signal.java

johnou: actually it's caught further up, might be okay.

normanmaurer: @johnou yep I think its fine

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -100,7 +100,12 @@ public SslContext sslContext() {
     protected final void onLookupComplete(ChannelHandlerContext ctx,
                                           String hostname, Future<SslContext> future) throws Exception {
         if (!future.isSuccess()) {
-            throw new DecoderException("failed to get the SslContext for " + hostname, future.cause());
+            final Throwable cause = future.cause();
+            if (cause instanceof Error) {
+                throw (Error) cause;
+            } else {
```
normanmaurer: nit: you can also remove the else here.

jasontedor: @normanmaurer I made this change, squashed, and force pushed here.

normanmaurer:
Cherry-picked into 4.1 (3fe1f715117d03d1597fc066ec2f8bc7abb51473) and 4.0 (b5b53daf9)

PR #7281, opened 2017-10-06
Fail the connectPromise with the correct exception if the connection …

normanmaurer:
…is refused when using the native kqueue transport.

Motivation:

Due a bug we happen to sometimes fail the connectPromise with a ClosedChannelException when using the kqueue transport and the remote peer refuses the connection. We need to ensure we fail it with the correct exception.

Modifications:

Call finishConnect() before calling close() to ensure we preserve the correct exception.

Result:

KQueueSocketConnectionAttemptTest.testConnectionRefused will pass always on macOS.


carl-mastrangelo:
Does a parallel change in AbstractEpollChannel need to be made too?

normanmaurer:
@carl-mastrangelo nope... it is an `kqueue only` fix

normanmaurer:
Cherry-picked into 4.1 (9bcf31977c69d2b70c486ac7af3c07e9f0db29c9)

Issue #6100, opened 2016-12-03
Upgrade ASM dependency to 6.0.0 final once released


normanmaurer:
We need to upgrade our ASM dependency to 6.0.0 final once its released to also support compiling on java9.

See  https://issues.apache.org/jira/browse/MSHADE-242

PR #7283, opened 2017-10-07
Upgrade dependencies to versions which use ASM 6.0.0+

johnou:
Motivation:

We need to upgrade our dependencies to versions which use ASM 6.0.0+ to support compiling on java9.

Modifications:

Update animal-sniffer-maven-plugin and maven-shade-plugin.

Result:

Fixes https://github.com/netty/netty/issues/6100

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -77,10 +77,6 @@
         <argLine.java9>--add-modules java.xml.bind ${argLine.java9.extras}</argLine.java9>
         <!-- Not use alpn agent as Java9 supports alpn out of the box -->
         <argLine.alpnAgent />
-        <!-- This is needed as earlier ASM versions not support java9 -->
-        <!-- https://issues.apache.org/jira/browse/MSHADE-242 -->
-        <!-- https://github.com/netty/netty/issues/6100 -->
-        <asm.version>6.0_BETA</asm.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -704,6 +691,9 @@
             <ignore>java.util.zip.CRC32</ignore>
             <ignore>java.util.zip.Adler32</ignore>
 
+            <!-- NioDatagramChannel implementation -->
+            <ignore>java.net.ProtocolFamily</ignore>
```
johnou: After updating animal-sniffer-maven-plugin..

```text
[INFO] --- animal-sniffer-maven-plugin:1.16:check (default) @ netty-transport ---
[INFO] Checking unresolved references to org.codehaus.mojo.signature:java16:1.1
[ERROR] /home/johno/Workspaces/netty/transport/src/main/java/io/netty/channel/socket/nio/NioDatagramChannel.java:104: Undefined reference: java.net.ProtocolFamily
```

johnou:
@normanmaurer shall I rebase on 4.0?

normanmaurer:
Cherry-picked into 4.1 (f301edfb9d91fedcdde63468eccde266efb05e50) and 4.0 (78a0699ef8c48edba760e82d84e31827490eaceb)

Issue #7285, opened 2017-10-08
Improve "Discarded inbound message" warning


doom369:
Right now, when the message reaches the end of the pipeline I can see a message:

> Discarded inbound message PooledUnsafeDirectByteBuf(ridx: 0, widx: 2, cap: 2) that reached at the tail of the pipeline. Please check your pipeline configuration.

However, in case of multiple pipelines, you can't find exact "wrong" pipeline. As above warning doesn't have any info regarding the flow channel. 

I think it would be a good idea to add something like : 

 `logger.debug("Handlers : {}.", Arrays.toString(ctx.pipeline().names().toArray()));`

 to the warning above, so end user could easily identify the wrong pipeline.

WDYT?

PR #7286, opened 2017-10-08
#7285 Improved "Discarded inbound message" warning

doom369:
Motivation:

On servers with many pipelines or dynamic pipelines, it is easy for end user to make mistake during pipeline configuration. Current message:

`Discarded inbound message PooledUnsafeDirectByteBuf(ridx: 0, widx: 2, cap: 2) that reached at the tail of the pipeline. Please check your pipeline configuration.`

Is not always meaningful and doesn't allow to find the wrong pipeline quickly.

Modification:

Added additional log message that identifies pipeline handlers. This will allow for the end users quickly find the problem pipeline.

Result:

Meaningful warning when the message reaches the end of the pipeline. Fixes https://github.com/netty/netty/issues/7285

### DIFF
--- a/transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
+++ b/transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
@@ -1243,6 +1243,10 @@ public class DefaultChannelPipeline implements ChannelPipeline {
 
         @Override
         public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+            if (logger.isDebugEnabled()) {
+                logger.debug("Reached end of the pipeline for handlers : {}.",
+                        ctx.pipeline().names());
+            }
             onUnhandledInboundMessage(msg);
         }
 


transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -1243,6 +1244,8 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws E
 
         @Override
         public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+            logger.debug("Reached end of the pipeline for handlers : {}.",
+                    Arrays.toString(ctx.pipeline().names().toArray()));
```
fenik17: `Arrays.toString()` and `toArray()` is not needed. Also will be better with check for logger level bcz `.names()` is not free:
```java
    logger.debug("Reached end of the pipeline for handlers : {}.", ctx.pipeline().names());
}
```

doom369: @fenik17 good points. Let me fix.

normanmaurer:
@doom369 wouldn't it be good enough to just log the Channel itself ?

doom369:
@normanmaurer channel doesn't have enough info. For example, in case of port unification, you still don't know exactly pipeline.

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -1243,6 +1243,10 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws E
 
         @Override
         public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+            if (logger.isDebugEnabled()) {
+                logger.debug("Reached end of the pipeline for handlers : {}.",
```
normanmaurer: Can we also include the Channel itself.

normanmaurer:
@carl-mastrangelo @Scottmitch WDYT ?

carl-mastrangelo:
Not opposed to the change itself, but it would be nice if the debug messages happened in the same place.   Specifically in `onUnhandledInboundMessage`, which also has access to the names, and correctly handles exceptions with respect to releasing the object.

normanmaurer:
@doom369 can you address @carl-mastrangelo s comment ?

carl-mastrangelo:
@doom369 I think that should be okay to factor it into a private method, and leave `onUnhandledInboundMessage` forwarding to it.

normanmaurer:
Let me close this due no response

PR #7288, opened 2017-10-09
Correct filling an `origin` header for WS client

fenik17:
Motivation:
An `origin`/`sec-websocket-origin` header value in websocket client is filling incorrect in some cases:
- Hostname is not converting to lower-case as prescribed by RFC 6354 (see [1]).
- Selecting a `http` scheme when source URI has `wss`/`https` scheme and non-standard port.

Modifications:
- Convert uri-host to lower-case.
- Use a `https` scheme if source URI scheme is `wss`/`https`, or if source scheme is null and port == 443.

Result:
Correct filling an `origin` header for WS client.

[1] https://tools.ietf.org/html/rfc6454#section-4

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
@@ -39,6 +39,7 @@ import io.netty.util.internal.ThrowableUtil;
 
 import java.net.URI;
 import java.nio.channels.ClosedChannelException;
+import java.util.Locale;
 
 /**
  * Base class for web socket client handshake implementations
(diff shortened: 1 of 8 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
```diff
@@ -477,14 +469,30 @@ static CharSequence websocketHostValue(URI wsURL) {
         return NetUtil.toSocketAddressString(host, port);
     }
 
-    static CharSequence websocketOriginValue(String host, int wsPort) {
-        String originValue = (wsPort == HttpScheme.HTTPS.port() ?
-                HttpScheme.HTTPS.name() : HttpScheme.HTTP.name()) + "://" + host;
-        if (wsPort != HttpScheme.HTTP.port() && wsPort != HttpScheme.HTTPS.port()) {
+    static CharSequence websocketOriginValue(URI wsURL) {
+        String scheme = wsURL.getScheme();
+        String schemePrefix;
```
normanmaurer: final

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
```diff
@@ -477,14 +469,30 @@ static CharSequence websocketHostValue(URI wsURL) {
...
+        int port = wsURL.getPort();
+        int defaultPort;
```
normanmaurer: final

fenik17:
@normanmaurer done.

normanmaurer:
Cherry-picked into 4.1 (dc98eae5a5a3ac4d4e7b50d4d38018993d52616b) and 4.0 (ea68fcfc4683fb9cf529680a94410abb7dc257be)

normanmaurer:
@fenik17 thanks!

PR #7291, opened 2017-10-11
DnsResolver CNAME redirect bug

Scottmitch:
Motviation:
DnsNameResolverContext#followCname attempts to build a query to follow a CNAME, but puts the original hostname in the DnsQuery instead of the CNAME hostname. This will result in not following CNAME redirects correctly.

Result:
- DnsNameResolverContext#followCname should use the CNAME instead of the original hostname when building the DnsQuery

Result:
More correct handling of redirect queries.

### DIFF
--- a/resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
+++ b/resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
@@ -425,7 +425,7 @@ public class DnsNameResolverTest {
     }
 
     @Test
-    public void  testResolveA() throws Exception {
+    public void testResolveA() throws Exception {
         DnsNameResolver resolver = newResolver(ResolvedAddressTypes.IPV4_ONLY)
                 // Cache for eternity
                 .ttl(Integer.MAX_VALUE, Integer.MAX_VALUE)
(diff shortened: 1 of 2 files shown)


resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
```diff
@@ -742,6 +742,81 @@ public void testResolveAllEmptyIpv6() {
...
+    public void testCNAMEResolveAllIpv4() throws IOException, InterruptedException {
+        testCNAMERecursiveResolve(true);
+    }
+
+    @Test
+    public void testCNAMEResolveAllIpv6() throws IOException, InterruptedException {
+        testCNAMERecursiveResolve(false);
+    }
+
+    public void testCNAMERecursiveResolve(boolean ipv4Preferred) throws IOException, InterruptedException {
```
normanmaurer: nit: make this private

Scottmitch: done

Scottmitch:
4.1 (dcb828f02f)

normanmaurer:
@Scottmitch thanks!

PR #7300, opened 2017-10-13
Minor improvements in ByteBufOutputStream

fenik17:
Motivation:
In the `ByteBufOutputStream` we can use an appropriate methods of `ByteBuf`
to reduce calls of virtual methods and do not copying converting logic.

Modifications:
- Use an appropriate methods of `ByteBuf`
- Remove redundant conversions (int -> byte, int -> char).
- Use `ByteBuf#writeCharSequence` in the `writeBytes(String)'.

Result:
Less code duplication. A `writeBytes(String)` method is faster.
No unnecessary conversions. More consistent and cleaner code.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/ByteBufOutputStream.java
+++ b/buffer/src/main/java/io/netty/buffer/ByteBufOutputStream.java
@@ -15,6 +15,8 @@
  */
 package io.netty.buffer;
 
+import io.netty.util.CharsetUtil;
+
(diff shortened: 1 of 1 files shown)


normanmaurer:
Cherry-picked into 4.1 (09dd6a5d4d96fed5364187e5df678f063ab5c6e0) and 4.0 (134ee2dbb)

normanmaurer:
@fenik17 thanks

PR #7301, opened 2017-10-13
Fixes a javadoc for ByteBufUtil#copy method

fenik17:
Motivation:
Javadoc of the `ByteBufUtil#copy(AsciiString, int, ByteBuf, int, int)` is incorrect.

Modifications:
Fix it.

Result:
The description of the `#copy` method is not misleading.


normanmaurer:
Cherry-picked into 4.1 (17e1a26d64fa2d2bfcd068bbaf68803bad949b42)

PR #7304, opened 2017-10-14
Make Comparators Serializable

idelpivnitskiy:
Motivation:

Objects of `java.util.TreeMap` or `java.util.TreeSet` will become non-Serializable if instantiated with Comparators, which are not also Serializable. This can result in unexpected and difficult-to-diagnose bugs.

Modifications:

Implements Serializable for all classes, which implements Comparator.

Result:

Proper Comparators which will not force collections to non-Serializable mode.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/WeightedFairQueueByteDistributor.java
@@ -25,6 +25,7 @@ import io.netty.util.internal.PriorityQueueNode;
 import io.netty.util.internal.SystemPropertyUtil;
 import io.netty.util.internal.UnstableApi;
 
+import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.Iterator;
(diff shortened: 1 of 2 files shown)


codec-http/src/main/java/io/netty/handler/codec/spdy/SpdySession.java
```diff
@@ -322,7 +323,8 @@ void clearPendingWrites(Throwable cause) {
         }
     }
 
-    private final class StreamComparator implements Comparator<Integer> {
+    private final class StreamComparator implements Comparator<Integer>, Serializable {
```
netkins: CRITICAL Make this inner class static [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2066)

idelpivnitskiy: Hm... If I will follow this rule, I have to make Serializable a huge chain of classes: StreamComparator -> SpdySession -> StreamState -> PendingWrite -> ChannelPromise...

We can remove back `Serializable` from `StreamComparator`, because it used only once - inside method `SpdySession.activeStreams()` and after that will never be serialized. So, in this case, it's not a critical problem.

normanmaurer:
Cherry-picked into 4.1 (4793daa589ffeb52886409a01e607d3dbe42551a)

PR #7306, opened 2017-10-14
Make methods 'static' where it possible

idelpivnitskiy:
Motivation:

Even if it's a super micro-optimization (most JVM could optimize such cases in runtime), in theory (and according to some performance tests) it may help a bit. It also makes a code more clear and allows you to access such methods in the test scope directly, without an instance of the class.

Modifications:

Add 'static' modifier for all methods, where it possible. Mostly in test scope.

Result:

Cleaner code with proper 'static' modifiers.

### DIFF
--- a/buffer/src/test/java/io/netty/buffer/ReadOnlyByteBufTest.java
+++ b/buffer/src/test/java/io/netty/buffer/ReadOnlyByteBufTest.java
@@ -202,7 +202,7 @@ public class ReadOnlyByteBufTest {
         ensureWritableIntStatusShouldFailButNotThrow(true);
     }
 
-    private void ensureWritableIntStatusShouldFailButNotThrow(boolean force) {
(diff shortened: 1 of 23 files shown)


common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
```diff
@@ -370,7 +370,7 @@ protected boolean runAllTasks() {
      *
      * @return {@code true} if at least one task was executed.
      */
-    protected final boolean runAllTasksFrom(Queue<Runnable> taskQueue) {
+    protected static boolean runAllTasksFrom(Queue<Runnable> taskQueue) {
```
idelpivnitskiy: not sure about this one

normanmaurer:
Cherry-picked into 4.1 (50a067a8f72b518fd1c9a2539cd777e00625436e) and 4.0 (a1328419f)

PR #7322, opened 2017-10-21
Correctly update Channel writability when queueing data in SslHandler.

normanmaurer:
Motivation:

A regression was introduced in 86e653e which had the effect that the writability was not updated for a Channel while queueing data in the SslHandler.

Modifications:

- Factor out code that will increment / decrement pending bytes and use it in AbstractCoalescingBufferQueue and PendingWriteQueue
- Add test-case

Result:

Channel writability changes are triggered again.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -352,7 +352,7 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
     private boolean sentFirstMessage;
     private boolean flushedBeforeHandshake;
     private boolean readDuringHandshake;
-    private final SslHandlerCoalescingBufferQueue pendingUnencryptedWrites = new SslHandlerCoalescingBufferQueue(16);
+    private SslHandlerCoalescingBufferQueue pendingUnencryptedWrites;
     private Promise<Channel> handshakePromise = new LazyChannelPromise();
     private final LazyChannelPromise sslClosePromise = new LazyChannelPromise();
 
@@ -375,6 +375,7 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
     private volatile long handshakeTimeoutMillis = 10000;
     private volatile long closeNotifyFlushTimeoutMillis = 3000;
     private volatile long closeNotifyReadTimeoutMillis;
+    volatile int wrapDataSize = MAX_PLAINTEXT_LENGTH;
 
     /**
      * Creates a new instance.
@@ -474,7 +475,7 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
      */
     @UnstableApi
     public final void setWrapDataSize(int wrapDataSize) {
-        pendingUnencryptedWrites.wrapDataSize = wrapDataSize;
+        this.wrapDataSize = wrapDataSize;
     }
 
     /**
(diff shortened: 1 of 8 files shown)


normanmaurer:
@smaldini @violetagg PTAL as well.

violetagg:
+1 LGTM and my tests are ok

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1838,7 +1838,7 @@ protected ByteBuf removeEmptyValue() {
             return null;
         }
 
-        private static boolean attemptCopyToCumulation(ByteBuf cumulation, ByteBuf next, int wrapDataSize) {
```
Scottmitch: nit: this could be kept static and moved outside the scope of this class.

transport/src/main/java/io/netty/channel/AbstractCoalescingBufferQueue.java
```diff
@@ -30,21 +30,40 @@
...
+    private final PendingBytesTracker tracker;
     private int readableBytes;
 
-    public AbstractCoalescingBufferQueue(int initSize) {
+    /**
+     * Create a new instance.
+     *
+     * @param initSize theinitial size of the underlying queue.
+     */
+    protected AbstractCoalescingBufferQueue(int initSize) {
```
Scottmitch: this is now unused ... consider removing it ... this class is labeled with `@UnstableApi`.

transport/src/main/java/io/netty/channel/AbstractCoalescingBufferQueue.java
```diff
@@ -291,6 +308,17 @@ private void incrementReadableBytes(int increment) {
         if (nextReadableBytes < readableBytes) {
             throw new IllegalStateException("buffer queue length overflow: " + readableBytes + " + " + increment);
         }
+        if (tracker != null) {
+            tracker.incrementPendingOutboundBytes(increment);
+        }
         readableBytes = nextReadableBytes;
```
Scottmitch: update the state of this class before calling out to `tracker` ... tracker may call into user code which may result in re-entry.

transport/src/main/java/io/netty/channel/PendingBytesTracker.java
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.channel;
+
+abstract class PendingBytesTracker implements MessageSizeEstimator.Handle {
```
Scottmitch: consider making this an interface ... it has no code.

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
normanmaurer: sure why not...

transport/src/main/java/io/netty/channel/PendingBytesTracker.java
```diff
@@ -0,0 +1,104 @@
...
+    private final MessageSizeEstimator.Handle estimatorHandle;
+
+    private PendingBytesTracker(MessageSizeEstimator.Handle estimatorHandle) {
+        this.estimatorHandle = ObjectUtil.checkNotNull(estimatorHandle, "estimatorHandle");
+    }
+
+    @Override
+    public int size(Object msg) {
```
Scottmitch: make this `final` for now?

normanmaurer:
Cherry-picked into 4.1 (55b501d0d47c24a5db84386be11994b2eaf14579) . 4.0 is not affected

PR #7328, opened 2017-10-23
Added QueryStringDecoder.rawPath() and rawQuery()

nedtwigg:
Motivation:

Before this commit, it is impossible to access the path component of the
URI before it has been decoded.  This makes it impossible to distinguish
between the following URIs:

/user/title?key=value
/user%2Ftitle?key=value

The user could already access the raw uri value, but they had to calculate
pathEndIdx themselves, even though it might already be cached inside
QueryStringDecoder.

Result:

The user can easily and efficiently access the undecoded path and query.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
@@ -182,6 +182,21 @@ public class QueryStringDecoder {
         return params;
     }
 
+    /**
+     * Returns the raw path string of the URI.
+     */
+    public String rawPath() {
+        return uri.substring(0, pathEndIdx());
+    }
+
+    /**
+     * Returns raw query string of the URI.
+     */
+    public String rawQuery() {
+        int start = pathEndIdx() + 1;
+        return start < uri.length() ? uri.substring(start) : EMPTY_STRING;
+    }
+
     private int pathEndIdx() {
         if (pathEndIdx == -1) {
             pathEndIdx = findPathEndIndex(uri);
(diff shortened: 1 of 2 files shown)


fenik17:
The `pathEndIdx` method is the implementation details. It's not a good idea to put it outside, imho. Maybe better add two methods to getting a raw components?

```java
    public String rawPath() {
        return uri.substring(0, pathEndIdx());
    }

    public String rawQuery() {
        return uri.substring(pathEndIdx() + 1, uri.length());
    }
```

nedtwigg:
That was my first instinct, except that every public method on `QueryStringDecoder` right now is calculated lazily and cached, but such an implementation of `rawPath()` and `rawQuery()` would not be.

Caching is an implementation detail, but since every method `QueryStringDecoder` currently has the same performance behavior, I figured it would be less controversial to make `pathEndIdx()` public than to add new methods that have different performance characteristics than the existing methods.

I'm happy with either, I'll leave that up to you :)

normanmaurer:
@nedtwigg I think I would do what @fenik17 suggested.

fenik17:
@nedtwigg
> I figured it would be less controversial to make pathEndIdx() public than to add new methods that have different performance characteristics than the existing methods.

Getting raw components is not a direct destination of the `QueryStringDecoder`. I think it will not be a big problem if caching issues are left to the discretion of the user.

nedtwigg:
Roger, I reverted my previous commit, and added `rawPath()` and `rawQuery()`, and added some tests for `QueryStringDecoderTest`.  If you'd rather I force-pushed a branch without the revert, just lemme know.

fenik17:
Add tests with getting empty string by rawQuery(), please. For example:
```java
new QueryStringDecoder("path/")
new QueryStringDecoder("path?")
new QueryStringDecoder("")
```
This is for the case when `start >= uri.length()`.

fenik17:
I think it's time to squash commits and update the issue/commit description.

nedtwigg:
Squashed with updated commit description.

normanmaurer:
@nedtwigg can you please sign our icla and let me know once done:

https://netty.io/s/icla

nedtwigg:
Signed.

normanmaurer:
Cherry-picked into 4.1 (dcbbae7f9006f5b995d055702dd1c9f55f0e0c9d) and 4.0 (ac8f66bb7b029f48940e8fe165a962f84e2c531d)

normanmaurer:
@nedtwigg thanks!

PR #7330, opened 2017-10-23
Trigger user event when H2 conn preface & SETTINGS frame is sent

lionelli1:
Motivation:
Previously client Http2ConnectionHandler trigger a user event
immediately when the HTTP/2 connection preface is sent. Any attempt to
immediately send a new request would cause the server to terminate the
connection, as it might not have received the SETTINGS frame from the
client. Per RFC7540 Section 3.5, the preface "MUST be followed by a
SETTINGS frame (Section 6.5), which MAY be empty."
(https://tools.ietf.org/html/rfc7540#section-3.5)

This event could be made more meaningful if it also indicates that the
initial client SETTINGS frame has been sent to signal that the channel
is ready to send new requests.

Modification:
- Renamed event to Http2ConnectionPrefaceAndSettingsFrameWrittenEvent.
- Modified Http2ConnectionHandler to trigger the user event only if it
  is a client and it has sent both the preface and SETTINGS frame.

Result:
It is now safe to use the event as an indicator that the HTTP/2
connection is ready to send new requests.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
@@ -231,7 +231,7 @@ public class Http2FrameCodec extends Http2ConnectionHandler {
      */
     @Override
     public final void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
-        if (evt instanceof Http2ConnectionPrefaceWrittenEvent) {
+        if (evt == Http2ConnectionPrefaceAndSettingsFrameWrittenEvent.INSTANCE) {
             // The user event implies that we are on the client.
             tryExpandConnectionFlowControlWindow(connection());
         } else if (evt instanceof UpgradeEvent) {
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
@@ -364,15 +364,20 @@ public class Http2ConnectionHandler extends ByteToMessageDecoder implements Http
 
             prefaceSent = true;
 
-            if (!connection().isServer()) {
+            final boolean isClient = !connection().isServer();
+            if (isClient) {
                 // Clients must send the preface string as the first bytes on the connection.
                 ctx.write(connectionPrefaceBuf()).addListener(ChannelFutureListener.CLOSE_ON_FAILURE);
-                ctx.fireUserEventTriggered(Http2ConnectionPrefaceWrittenEvent.INSTANCE);
             }
 
             // Both client and server must send their initial settings.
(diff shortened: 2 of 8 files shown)


lionelli1:

normanmaurer:
@ejona86 @carl-mastrangelo PTAL... imho this is the correct thing to do but would like to have your review as well as it may also affect GRPC.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
```diff
@@ -231,7 +231,7 @@ void handlerAdded0(@SuppressWarnings("unsed") ChannelHandlerContext ctx) throws
      */
     @Override
     public final void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
-        if (evt instanceof Http2ConnectionPrefaceWrittenEvent) {
+        if (evt instanceof Http2ConnectionPrefaceAndSettingsFrameWrittenEvent) {
```
carl-mastrangelo: `if evt == Http2ConnectionPrefaceAndSettingsFrameWrittenEvent.INSTANCE`

lionelli1: good point since this is the only instance of this class. changed.

Scottmitch:
thanks @lionelli1 !

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
```diff
@@ -231,7 +231,7 @@ void handlerAdded0(@SuppressWarnings("unsed") ChannelHandlerContext ctx) throws
      */
     @Override
     public final void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
-        if (evt instanceof Http2ConnectionPrefaceWrittenEvent) {
+        if (evt == Http2ConnectionPrefaceAndSettingsFrameWrittenEvent.INSTANCE) {
```
Scottmitch: lets be consistent and either use `evt == Http2ConnectionPrefaceAndSettingsFrameWrittenEvent.INSTANCE` or `instanceof` ... `instanceof` allows us to be less dependent on the class hierarchy but I think either way will work.

normanmaurer: @lionelli1 @Scottmitch I vote for using `==` here. @lionelli1 please make the change and squash. After this is done i will merge it.

lionelli1: ok! replaced all `instanceof` with `==` and squashed my changes. thanks!

normanmaurer:
Cherry-picked into 4.1 (baf273aea88b5c8124cacd4b4bf17a6807bf88fc).

@lionelli1 thanks a lot!

PR #7331, opened 2017-10-23
Set readPending to false when ever a read is done

### LINKED ISSUE #7255: A lot of ChannelInputShutdownReadComplete events fired for NIO channel with ALLOW_HALF_CLOSURE

Scottmitch:
Motivation:
readPending is currently only set to false if data is delivered to the application, however this may result in duplicate events being received from the selector in the event that the socket was closed.

Modifications:
- We should set readPending to false before each read attempt for all
transports besides NIO.
- Based upon the Javadocs it is possible that NIO may have spurious
wakeups [1]. In this case we should be more cautious and only set
readPending to false if data was actually read.

[1] https://docs.oracle.com/javase/7/docs/api/java/nio/channels/SelectionKey.html
That a selection key's ready set indicates that its channel is ready for some operation category is a hint, but not a guarantee, that an operation in such a category may be performed by a thread without causing the thread to block.

Result:
Notification from the selector (or simulated events from kqueue/epoll ET) in the event of socket closure.
Fixes #7255

### DIFF
--- a/transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
+++ b/transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
@@ -126,6 +126,16 @@ public abstract class AbstractNioByteChannel extends AbstractNioChannel {
                         byteBuf.release();
                         byteBuf = null;
                         close = allocHandle.lastBytesRead() < 0;
+                        if (close) {
+                            // Based upon the Javadocs it is possible that NIO may have spurious wake ups [1]. In this
+                            // case we should be more cautious and only set readPending to false if data was actually
+                            // read.
+                            // [1] https://docs.oracle.com/javase/7/docs/api/java/nio/channels/SelectionKey.html
+                            // That a selection key's ready set indicates that its channel is ready for some operation
+                            // category is a hint, but not a guarantee, that an operation in such a category may be
+                            // performed by a thread without causing the thread to block.
+                            readPending = false;
+                        }
                         break;
                     }
 
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollStreamChannel.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollStreamChannel.java
@@ -727,7 +727,6 @@ public abstract class AbstractEpollStreamChannel extends AbstractEpollChannel im
                 EpollRecvByteAllocatorHandle allocHandle) {
             if (byteBuf != null) {
                 if (byteBuf.isReadable()) {
-                    readPending = false;
                     pipeline.fireChannelRead(byteBuf);
                 } else {
                     byteBuf.release();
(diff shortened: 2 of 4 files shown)


Scottmitch:
@grossws @khitrin - would you mind verifying this PR?

khitrin:
@Scottmitch The PR resolves #7255. Thank you for this fix.

normanmaurer:
@Scottmitch I wonder if this patch is correct... Shouldn't we really only set readPending to false if something was read from the socket ? I think in practice we may get a spurious wakeup try to read something but there is nothing on the socket to be read. In this case I think we want to keep on reading. 

Maybe I am missing something ?

normanmaurer:
@Scottmitch sounds good to me... thanks for digging :)

Scottmitch:
@normanmaurer - code updated. PTAL.

transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
```diff
@@ -126,6 +126,9 @@ public final void read() {
                         byteBuf.release();
                         byteBuf = null;
                         close = allocHandle.lastBytesRead() < 0;
+                        if (close) {
+                            readPending = false;
```
normanmaurer: @Scottmitch can you just please add a comment why we handle it special here compared to the native transports ?

Scottmitch: done

Scottmitch:
4.1 (413c7c2cd8)

PR #7332, opened 2017-10-23
Fixes a LoggingHandler#format method with two arguments

fenik17:
Motivation:
Bug in capacity calculation: occurs auto convert to string instead of sum up.

Modifications:
Use `eventName.length()` in sum.

Result:
Less trash in logs.


normanmaurer:
Cherry-picked into 4.1 (dcc39e5b21eda7bdd9563ed86253a693923f6432).

@fenik17 thanks!

PR #7333, opened 2017-10-23
Use WeakReferences for Resource Leaks

carl-mastrangelo:
Motivation:
Phantom references are for cleaning up resources that were
forgotten, which means they keep their referent alive.   This
means garbage is kept around until the refqueue is drained, rather
than when the reference is unreachable.

Modification:
Use Weak References instead of Phantoms

Result:
More punctual leak detection.


normanmaurer:
@Scottmitch PTAL as well if you have time... Tested and works as expected.

normanmaurer:
Cherry-picked into 4.1 (e62e6df4acd8d66623fdce271e009d3fe76d203d) and 4.0 (e5d4e28e4644d6df97ab17488843fce1734dcc5b).

@carl-mastrangelo thanks buddy!

PR #7334, opened 2017-10-23
Http2StreamFrameToHttpObjectCodec should handle 100-Continue properly

lionelli1:
Motivation:
Http2StreamFrameToHttpObjectCodec was not properly encoding and
decoding 100-Continue HttpResponse/Http2SettingsFrame properly. It was
encoding 100-Continue FullHttpResponse as an Http2SettingFrame with
endStream=true, causing the child channel to terminate. It was not
decoding 100-Continue Http2SettingsFrame (endStream=false) as
FullHttpResponse. This should be fixed as it would cause http2 child
stream to prematurely close, and could cause HttpObjectAggregator to
fail if it's in the pipeline.

Modification:
- Fixed encode() to properly encode 100-Continue FullHttpResponse as
  Http2SettingsFrame with endStream=false
- Fixed decode() to properly decode 100-Continue Http2SettingsFrame
  (endStream=false) as a FullHttpResponse

Result:
Now Http2StreamFrameToHttpObjectCodec should be properly handling
100-Continue responses.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
@@ -20,23 +20,27 @@ import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.Unpooled;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandler;
+import io.netty.channel.ChannelHandler.Sharable;
 import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.EncoderException;
 import io.netty.handler.codec.MessageToMessageCodec;
 import io.netty.handler.codec.http.DefaultHttpContent;
 import io.netty.handler.codec.http.DefaultLastHttpContent;
 import io.netty.handler.codec.http.FullHttpMessage;
+import io.netty.handler.codec.http.FullHttpResponse;
 import io.netty.handler.codec.http.HttpContent;
 import io.netty.handler.codec.http.HttpHeaderNames;
 import io.netty.handler.codec.http.HttpHeaderValues;
 import io.netty.handler.codec.http.HttpMessage;
 import io.netty.handler.codec.http.HttpObject;
 import io.netty.handler.codec.http.HttpRequest;
+import io.netty.handler.codec.http.HttpResponse;
+import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http.HttpScheme;
 import io.netty.handler.codec.http.HttpUtil;
 import io.netty.handler.codec.http.HttpVersion;
 import io.netty.handler.codec.http.LastHttpContent;
 import io.netty.handler.ssl.SslHandler;
-import io.netty.util.ReferenceCountUtil;
 import io.netty.util.internal.UnstableApi;
 
 import java.util.List;
@@ -51,6 +55,7 @@ import java.util.List;
  * is a single header.
  */
 @UnstableApi
+@Sharable
 public class Http2StreamFrameToHttpObjectCodec extends MessageToMessageCodec<Http2StreamFrame, HttpObject> {
     private final boolean isServer;
     private final boolean validateHeaders;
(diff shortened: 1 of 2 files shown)


lionelli1:

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
```diff
@@ -120,6 +133,17 @@ private void encodeLastContent(LastHttpContent last, List<Object> out) {
 
     @Override
     protected void encode(ChannelHandlerContext ctx, HttpObject obj, List<Object> out) throws Exception {
+        // 100-continue is typically a FullHttpResponse, but the decoded
+        // Http2HeadersFrame should not be marked as endStream=true
+        if (obj instanceof FullHttpResponse) {
```
lionelli1: in the event of separate sequential HttpResponse (status=100) and LastHttpContent objects being written, the following objects would be encoded:

```
1. Http2HeadersFrame(status:100, endStream=false)
2. Http2HeadersFrame(endStream=true)
```
unless we maintain some kind of state in this handler to swallow the LastHttpContent, this will result in the stream to close immediately after 100-continue is written. i was hoping that we don't have to maintain that state, as devs could simply write a FullHttpResponse instead...

normanmaurer: @lionelli1 I think we should just mention in the java docs that this is needed and throw an exception if the dev tries to not use a FullHttpResponse here. Can you do this and also add a unit test ?

normanmaurer:
@lionelli1 also it seems there is a resource leak which means you not release the buffer correctly.

Please run with `mvn -Pleak clean package` and fix it.

normanmaurer:
@carl-mastrangelo I need to investigate but I think there may be something wrong with your leak-detector changes as we had no trace logged for the leak report here:

https://garage.netty.io/teamcity/viewLog.html?buildId=25308&buildTypeId=pulls_netty&guest=1

normanmaurer:
@carl-mastrangelo never mind.... I was blind and need more coffee ;) All good

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
lionelli1: good point. changed and added unit test to verify behavior.

lionelli1:
@normanmaurer fixed the leak. thanks for the pointer!

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
normanmaurer: <3

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
```diff
@@ -118,8 +133,36 @@ private void encodeLastContent(LastHttpContent last, List<Object> out) {
...
+        if (obj instanceof HttpResponse) {
+            final HttpResponse res = (HttpResponse) obj;
+            if (res.status().equals(HttpResponseStatus.CONTINUE)) {
+                if (res instanceof FullHttpResponse) {
+                    final Http2Headers headers = toHttp2Headers(res);
+                    out.add(new DefaultHttp2HeadersFrame(headers, false));
+                    return;
+                } else {
+                    throw new EncoderException(
+                            "100-continue response must be a FullHttpResponse");
```
Scottmitch: consider using the `HttpResponseStatus.CONTINUE` instead of `100-continue` for consistency.

lionelli1: fixed.

normanmaurer:
@lionelli1 thanks... Cherry-picked into 4.1 (424bb09d2424114ee0f81a397f1079ccd21ebf64)

Issue #7337, opened 2017-10-24
Possible buffer leak in SslHandler?


Anon43590:
Excuse me if I'm missing something obvious, but it seems to me that in SslHandler.java, "buf" removed from "pendingUnencryptedWrites" on line 751 may leak if "wrap(alloc, engine, buf, out)" called on line 762 throws. The wrap function on line 912 is marked as "throws SSLException", which I assume is something that could possibly happen and not be a fatal error for the entire server. So it seems like the aforementioned buffer had ought to be released if such an exception occurs, perhaps in the finally block on line 803.

I only noticed this randomly while helping a coworker to track down an unrelated issue, so it's not like I've actually seen this leak happen.

normanmaurer:
You are correct... let me fix it.

Anon43590:
Cool, thanks

normanmaurer:
@Anon43590 please check https://github.com/netty/netty/pull/7338

Anon43590:
@normanmaurer looks correct to me, thanks

PR #7338, opened 2017-10-24
Fix possible leak in SslHandler if wrap(...) throws.

normanmaurer:
Motivation:

We can end up with a buffer leak if SSLEngine.wrap(...) throws.

Modifications:

Correctly release the ByteBuf if SSLEngine.wrap(...) throws.

Result:

Fixes [#7337].

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -742,13 +742,14 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
         ChannelPromise promise = null;
         ByteBufAllocator alloc = ctx.alloc();
         boolean needUnwrap = false;
+        ByteBuf buf = null;
         try {
             final int wrapDataSize = this.wrapDataSize;
             // Only continue to loop if the handler was not removed in the meantime.
             // See https://github.com/netty/netty/issues/5860
             while (!ctx.isRemoved()) {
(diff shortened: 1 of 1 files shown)


handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -801,6 +804,10 @@ private void wrap(ChannelHandlerContext ctx, boolean inUnwrap) throws SSLExcepti
                 }
             }
         } finally {
+            // Ownership of buffer was not transfered, release it.
```
Scottmitch: `transfered ` -> `transferred`

eun-ice:
I have seen SSLEngine.wrap throw with OpenSSL in production and some weird half-closed SSL states.
So thank you for this fix!

normanmaurer:
Cherry-picked into 4.1 (92b786e2f33a44f16be224305fe4fb716422b14a). 4.0 is not affected

jchambers:
I'm trying to figure out if this might have been the cause of some leaks I've been chasing down. My understanding is that:

1. This problem has been around for at least six months, and possibly longer.
2. In practical terms, this problem is most likely to manifest on a lossy connection, and generally should not me an issue on high-quality, low-loss connections.

Is that accurate? If it is, then I think this solves our problem. Thanks kindly!

normanmaurer:
Yes I think it’s correct

jchambers:
Thank you!

PR #7339, opened 2017-10-24
Take the architecture into account when loading netty-tcnative

normanmaurer:
Motivation:

We should ensure we only try to load the netty-tcnative version that was compiled for the architecture we are using.

Modifications:

Include architecture into native lib name.

Result:

Only load native lib if the architecture is supported.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
+++ b/handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
@@ -406,15 +406,17 @@ public final class OpenSsl {
         String arch = PlatformDependent.normalizedArch();
 
         Set<String> libNames = new LinkedHashSet<String>(4);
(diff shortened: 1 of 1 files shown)


normanmaurer:
This depends on https://github.com/netty/netty-tcnative/pull/315.

normanmaurer:
@doom369 also FYI...

normanmaurer:
@carl-mastrangelo it depends on https://github.com/netty/netty-tcnative/pull/315 which ensures there is always the arch present. I am missing something?

normanmaurer:
@carl-mastrangelo PTAL again... I think this should work now also with your requirement (its the same we do for the native transports as well)

normanmaurer:
Cherry-picked into 4.1 (ad1f0d46b38b711b45456768135df055eec04280) and 4.0 (4a429692b3)

PR #7345, opened 2017-10-26
Support running Netty (in particular netty-tcnative) in the bootstrap class loader

trask:
Motivation:

Fix NullPointerExceptions that occur when running netty-tcnative inside the bootstrap class loader.

Modifications:

Replace `loader.getResource(...)` with `ClassLoader.getSystemResource(...)` when loader is null.

Replace `loader.loadClass(...)` with `Class.forName(..., false, loader)` which works when loader is both null and non-null.

### DIFF
--- a/common/src/test/java/io/netty/util/internal/NativeLibraryLoaderTest.java
+++ b/common/src/test/java/io/netty/util/internal/NativeLibraryLoaderTest.java
@@ -17,6 +17,7 @@ package io.netty.util.internal;
 
 import org.junit.Test;
 
+import java.lang.reflect.Method;
 import java.io.FileNotFoundException;
 import java.util.UUID;
 
@@ -25,6 +26,8 @@ import static org.junit.Assert.fail;
 
 public class NativeLibraryLoaderTest {
 
+    private static final Method getSupressedMethod = getGetSuppressed();
+
     @Test
     public void testFileNotFound() {
         try {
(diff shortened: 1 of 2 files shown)


normanmaurer:
@trask I wonder if we could somehow have an unit test.

ejona86:
For the test, you could add a binary to the bootstrap class loader, but just as a weird name "test-bs-loading.so" and load it in the test. But that'd be pretty annoying.

normanmaurer:
@ejona86 @trask if its too hard to add a test than dont worry... I just would prefer to have a unit test so we ensure we not break it again in the future.

@trask let me know what you think.

trask:
@normanmaurer @ejona86 unit test added, let me know what you think, the test caught one additional NPE, see additional change in latest push:

```
if (loader == null) {
    // cannot defineClass inside bootstrap class loader
    throw e1;
}
```

ejona86:
Oh, hey! _Much_ easier test. Right! The file doesn't even need to exist.

normanmaurer:
Cherry-picked into 4.1 (58e74e9fee1c12eb3a73c38109cfd081750992b6) and 4.0 (b12a1e3fa66b61a7ed625b2d4ee2c901171a2074).

@trask thanks!

trask:
Thanks for the quick review and merge

common/src/test/java/io/netty/util/internal/NativeLibraryLoaderTest.java
```diff
@@ -25,13 +26,56 @@
 
 public class NativeLibraryLoaderTest {
 
+    private static final Method getSupressedMethod = getGetSuppressed();
```
carl-mastrangelo: ThrowableUtil might have been a better place for this.

trask: I thought it was better not to put test-only code (getSuppressed) inside of non- test code (ThrowableUtil), but if that's not a concern I can move it there.

Issue #7264, opened 2017-09-29
SSL failures Netty 4.1.13 -> 4.1.14


rkapsi:
I have fallen a little bit behind in the Netty releases and am in the process of upgrading from Netty 4.1.13 w/ netty-tcnative 2.0.5 (linux, openssl, custom static build) to Netty 4.1.14 with same netty-tcnative version.

We're observing a failure where SSL is slowly failing and becoming more unreliable over a course of a couple of hours. Unreliable as in handshakes failing/hanging. It's a degrading failure with no apparent external metrics such as servers running of memory or more CPU being consumed. 

I'm going to bisect the 4.1.14 commits and hopefully find the source but if you've got ideas then let me know.

johnou:
@rkapsi any exceptions or thread dumps to share?

rkapsi:
@johnou there are no exceptions (I have `INFO` level enabled for the `io.netty.handler.ssl` and `io.netty.internal.tcnative` packages). I haven't bothered with thread dumps (yet). My bisect may take a few days (some non Netty things that I need to do first).

normanmaurer:
@rkapsi unfortunately there is nothing that comes into my mind :(

rkapsi:
Hey there, sorry for the slow progress. Here's what I have so far as I'm homing in with git bisect. I believe df568c739e2a73b4a1aea533a4fea934fdf9d0f7 is the last good commit and 7cfe4161823dec6192543e916b927e7de40190be ended rather catastrophically because it was missing commit 74140dbf53456c30600d7870316c5ac79878a003. I'll very tomorrow the last good commit but the bisect for the next few will be possibly a PITA because they appear to be tangled commits.

![screenshot from 2017-10-09 18-08-05](https://user-images.githubusercontent.com/191635/31360818-21634fae-ad1e-11e7-9b1b-a91b0d0a9198.png)

```
$ git bisect log
git bisect start
# bad: [8cc1071881e90b0130bdd35a0441abcd0df6ffa9] [maven-release-plugin] prepare release netty-4.1.14.Final
git bisect bad 8cc1071881e90b0130bdd35a0441abcd0df6ffa9
# good: [2a376eeb1b14b1f2e23e1c30ac2f2a213dbea25b] [maven-release-plugin] prepare for next development iteration
git bisect good 2a376eeb1b14b1f2e23e1c30ac2f2a213dbea25b
[...]

PR #7352, opened 2017-10-31
OpenSslEngine support unwrap plaintext greater than 2^14 and avoid infinite loop

Scottmitch:
Motivation:
If SslHandler sets jdkCompatibilityMode to false and ReferenceCountedOpenSslEngine sets jdkCompatibilityMode to true there is a chance we will get stuck in an infinite loop if the peer sends a TLS packet with length greater than 2^14 (the maximum length allowed in the TLS 1.2 RFC [1]). However there are legacy implementations which actually send larger TLS payloads than 2^14 (e.g. OpenJDK's SSLSessionImpl [2]) and in this case ReferenceCountedOpenSslEngine will return BUFFER_OVERFLOW in an attempt to notify that a larger buffer is to be used, but if the buffer is already at max size this process will repeat indefinitely.

[1] https://tools.ietf.org/html/rfc5246#section-6.2.1
[2] http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/d5a00b1e8f78/src/share/classes/sun/security/ssl/SSLSessionImpl.java#l793

Modifications:
- Support TLS payload sizes greater than 2^14 in ReferenceCountedOpenSslEngine
- ReferenceCountedOpenSslEngine should throw an exception if a
BUFFER_OVERFLOW is impossible to rectify

Result:
No more infinite loop in ReferenceCountedOpenSslEngine due to
BUFFER_OVERFLOW and large TLS payload lengths.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -58,15 +58,18 @@ import javax.net.ssl.SSLSessionContext;
 import javax.security.cert.X509Certificate;
 
 import static io.netty.handler.ssl.OpenSsl.memoryAddress;
-import static io.netty.util.internal.EmptyArrays.EMPTY_CERTIFICATES;
-import static io.netty.util.internal.EmptyArrays.EMPTY_JAVAX_X509_CERTIFICATES;
-import static io.netty.util.internal.ObjectUtil.checkNotNull;
 import static io.netty.handler.ssl.SslUtils.PROTOCOL_SSL_V2;
 import static io.netty.handler.ssl.SslUtils.PROTOCOL_SSL_V2_HELLO;
 import static io.netty.handler.ssl.SslUtils.PROTOCOL_SSL_V3;
 import static io.netty.handler.ssl.SslUtils.PROTOCOL_TLS_V1;
 import static io.netty.handler.ssl.SslUtils.PROTOCOL_TLS_V1_1;
 import static io.netty.handler.ssl.SslUtils.PROTOCOL_TLS_V1_2;
+import static io.netty.handler.ssl.SslUtils.SSL_RECORD_HEADER_LENGTH;
+import static io.netty.internal.tcnative.SSL.SSL_MAX_PLAINTEXT_LENGTH;
+import static io.netty.internal.tcnative.SSL.SSL_MAX_RECORD_LENGTH;
+import static io.netty.util.internal.EmptyArrays.EMPTY_CERTIFICATES;
+import static io.netty.util.internal.EmptyArrays.EMPTY_JAVAX_X509_CERTIFICATES;
+import static io.netty.util.internal.ObjectUtil.checkNotNull;
 import static java.lang.Integer.MAX_VALUE;
 import static java.lang.Math.min;
 import static javax.net.ssl.SSLEngineResult.HandshakeStatus.FINISHED;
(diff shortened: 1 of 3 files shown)


Scottmitch:
this depends upon https://github.com/netty/netty-tcnative/pull/318

/cc @rkapsi - This may be related to https://github.com/netty/netty/issues/7264

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
```diff
@@ -873,6 +875,222 @@ public void channelInactive(ChannelHandlerContext ctx) {
...
+                                            });
+                                        } else {
+                                            donePromise.tryFailure(sslEvt.cause());
+                                        }
+                                    }
+                                    ctx.fireUserEventTriggered(evt);
+                                }
+
+                                @Override
+                                public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
```
normanmaurer: Consider using AssertionError

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
```diff
@@ -873,6 +875,222 @@ public void channelInactive(ChannelHandlerContext ctx) {
...
+                                    donePromise.tryFailure(new IllegalStateException("server exception sentData: " +
+                                            sentData + " writeCause: " + writeCause, cause));
+                                }
+
+                                @Override
+                                public void channelInactive(ChannelHandlerContext ctx) {
+                                    donePromise.tryFailure(new IllegalStateException("server closed sentData: " +
+                                            sentData + " writeCause: " + writeCause));
```
normanmaurer: Consider using AssertionError

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
```diff
@@ -873,6 +875,222 @@ public void channelInactive(ChannelHandlerContext ctx) {
...
+                                        SslHandshakeCompletionEvent sslEvt = (SslHandshakeCompletionEvent) evt;
+                                        if (!sslEvt.isSuccess()) {
+                                            donePromise.tryFailure(sslEvt.cause());
+                                        }
+                                    }
+                                }
+
+                                @Override
+                                public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
+                                    donePromise.tryFailure(new IllegalStateException("client exception. bytesSeen: " +
```
normanmaurer: Consider using AssertionError

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
```diff
@@ -873,6 +875,222 @@ public void channelInactive(ChannelHandlerContext ctx) {
...
+                                            bytesSeen, cause));
+                                }
+
+                                @Override
+                                public void channelInactive(ChannelHandlerContext ctx) {
+                                    donePromise.tryFailure(new IllegalStateException("client closed. bytesSeen: " +
+                                            bytesSeen));
```
normanmaurer: Consider using AssertionError

normanmaurer:
@Scottmitch I wonder if we should alsways use jdkCompatible as default with this knowledge just to eliminate any surprises when people deploy their impl and use java on the remote peer as well

Scottmitch:
> I wonder if we should alsways use jdkCompatible as default with this knowledge just to eliminate any surprises when people deploy their impl and use java on the remote peer as well

This bug is only visible when `jdkCompatible=true` so I don't think changing the default would help in this case. I think we should leave the default as `jdkCompatible=false` when creating from our `SslContext`. If this is not desired the user should be able to change the default.

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
Scottmitch: `AssertionError(String, Throwable)` is jdk 1.7+ and is noisy in the IDE. I'll just stick with `IllegalStateException` here ... end result should be the same.

normanmaurer:
@Scottmitch ah yeah sorry... brain-fart . Should not review PRs via my phone 😢

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
normanmaurer: sounds good to me... Thanks for clarify

Scottmitch:
4.1 (fbe0e3506e) 4.0 (5c71b47122)

Issue #7353, opened 2017-10-31
java.lang.NullPointerException: ssl at ReferenceCountedOpenSslEngine.rejectRemoteInitiatedRenegotiation


floragunncom:
### Expected behavior
No NPE

### Actual behavior
java.lang.NullPointerException: ssl (see stracktrace below) at ReferenceCountedOpenSslEngine.rejectRemoteInitiatedRenegotiation

### Steps to reproduce
Happens randomly

### Netty version
4.1.13.Final with tcnative 2.0.5.Final and OpenSSL 1.0.2l  25 May 2017

### JVM version (e.g. `java -version`)
java version "1.8.0_121"
Java(TM) SE Runtime Environment (build 1.8.0_121-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.121-b13, mixed mode)

### OS version (e.g. `uname -a`)
Darwin MacBook-Pro.local 14.5.0 Darwin Kernel Version 14.5.0: Sun Jun  4 21:40:08 PDT 2017; root:xnu-2782.70.3~1/RELEASE_X86_64 x86_64

<pre>
io.netty.handler.codec.DecoderException: java.lang.NullPointerException: ssl
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:459) ~[netty-codec-4.1.13.Final.jar:4.1.13.Final]
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:265) ~[netty-codec-4.1.13.Final.jar:4.1.13.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362) [netty-transport-4.1.13.Final.jar:4.1.13.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348) [netty-transport-4.1.13.Final.jar:4.1.13.Final]
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340) [netty-transport-4.1.13.Final.jar:4.1.13.Final]
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1334) [netty-transport-4.1.13.Final.jar:4.1.13.Final]
[...]

PR #7357, opened 2017-11-01
Fix possible NPE in ReferenceCountedOpenSslEngine.rejectRemoteInitiat…

normanmaurer:
…edRenegotiation()

Motivation:

ReferenceCountedOpenSslEngine.rejectRemoteInitiatedRenegotiation() is called in a finally block to ensure we always check for renegotiation. The problem here is that sometimes we will already shutdown the engine before we call the method which will lead to an NPE in this case as the ssl pointer was already destroyed.

Modifications:

Check that the engine is not destroyed yet before calling SSL.getHandshakeCount(...)

Result:

Fixes [#7353].


normanmaurer:
Cherry-picked into 4.1 (7321418eb550ff5737b11acbc2843b29b466e968) and 4.0 (26417cf2ad625ab458c8559dd326c0e0027c64dc)

PR #7361, opened 2017-11-02
HttpObjectEncoder and MessageAggregator EMPTY_BUFFER usage

Scottmitch:
Motivation:
HttpObjectEncoder and MessageAggregator treat buffers that are not readable special. If a buffer is not readable, then an EMPTY_BUFFER is written and the actual buffer is ignored. If the buffer has already been released then this will not be correct as the promise will be completed, but in reality the original content shouldn't have resulted in any write because it was invalid.

Modifications:
- HttpObjectEncoder should retain/write the original buffer instead of using EMPTY_BUFFER
- MessageAggregator should retain/write the original ByteBufHolder instead of using EMPTY_BUFFER

Result:
Invalid write operations which happen to not be readable correctly reflect failed status in the promise, and do not result in any writes to the channel.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java
@@ -21,6 +21,7 @@ import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.FileRegion;
 import io.netty.handler.codec.MessageToMessageEncoder;
 import io.netty.util.CharsetUtil;
+import io.netty.util.ReferenceCountUtil;
 import io.netty.util.internal.StringUtil;
 
 import java.util.Iterator;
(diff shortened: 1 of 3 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java
```diff
@@ -109,10 +110,12 @@ protected void encode(ChannelHandlerContext ctx, Object msg, List<Object> out) t
         //     ch.write(Unpooled.EMPTY_BUFFER).addListener(ChannelFutureListener.CLOSE);
         //
         // See https://github.com/netty/netty/issues/2983 for more information.
```
ddossot: Should this comment be reworded? It's no more an `EMPTY_BUFFER` related issue but one more generally related to non-readable buffers.

codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java
```diff
@@ -109,10 +110,12 @@ protected void encode(ChannelHandlerContext ctx, Object msg, List<Object> out) t
...
-
-        if (msg instanceof ByteBuf && !((ByteBuf) msg).isReadable()) {
-            out.add(EMPTY_BUFFER);
-            return;
+        if (msg instanceof ByteBuf) {
+            final ByteBuf emptyBufPotential = (ByteBuf) msg;
```
ddossot: `potentialEmptyBuf`? That would be similar to `potentialDup` [here](https://github.com/netty/netty/blob/4.1/codec-http2/src/test/java/io/netty/handler/codec/http2/HashCollisionTest.java#L148).

codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java
Scottmitch: I think its OK for the comment to reference EMPTY_BUFFER as the "typical example" however the code also supports any empty buffer, and this PR doesn't change that behavior. The next question is do we want/need to continue to support this behavior which is less clear.

codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java
Scottmitch: variable name updated to be more consistent.

normanmaurer:
Cherry-picked into 4.1 (93130b172a61815354267f0f3d8f5af52de39754) and 4.0 (756e1579b376a57a49114f97cde2e04b386b4801)

Issue #7362, opened 2017-11-02
WebSockets V00 handshake broken


Scottmitch:
https://github.com/netty/netty/commit/7995afee8f1cb9047709239321d84ccb279fe4d1 modifies `HttpClientCoded#isContentAlwaysEmpty` to return `true` if the `statusCode == 101` which breaks the WebSocket V00 handshake [1]. The V00 handshake includes a payload in the request which is echoed back in the response. It could be argued that V00 is now irrelevant and folks should use the finalized RFC variant [2], but either way this feature is now broken. We have a few options:

1. Remove support for V00
2. Special case this somehow when V00 is used, and continue to support V00's handshake with a payload
3. Revert this fix
4. Update the `HttpClientCoded#isContentAlwaysEmpty` to check against `HttpStatusClass#INFORMATIONAL`.

Based upon the HTTP/1.x RFC [3] we shouldn't do 3, and we should actually also do 4.

Option 1 and 2 are debatable. It would be nice to do 2 because the V00 handshake is prohibited by the HTTP/1.x RFC.

[1] https://tools.ietf.org/html/draft-ietf-hybi-thewebsocketprotocol-00
[2] https://tools.ietf.org/html/rfc6455#section-1.3
[3] https://tools.ietf.org/html/rfc7231#section-6.2
> 1xx responses are terminated by the first empty line after
   the status-line (the empty line signaling the end of the header
   section).

/cc @pkolaczk

normanmaurer:
We should definitely special case this as we can not just drop the support.

normanmaurer:
Fixed by https://github.com/netty/netty/pull/7363. @Scottmitch @pkolaczk PTAL

normanmaurer:
Fixed by https://github.com/netty/netty/pull/7363

PR #7363, opened 2017-11-03
Correctly handle WebSockets 00 when using HttpClientCodec.

normanmaurer:
Motivation:

7995afee8f1cb9047709239321d84ccb279fe4d1 introduced a change that broke special handling of WebSockets 00.

Modifications:

Correctly delegate to super method which has special handling for WebSockets 00.

Result:

Fixes [#7362].


normanmaurer:
@Scottmitch @pkolaczk PTAL

pkolaczk:
looks good

normanmaurer:
Added a unit test as well

normanmaurer:
Cherry-picked into 4.1 (e0bbff74f7097f000472785982ad86c0ce891567) and 4.0 (4f4d800bb)

PR #7364, opened 2017-11-04
HTTP/2 write of released buffer should not write and should fail the …

Scottmitch:
…promise

Motivation:
HTTP/2 allows writes of 0 length data frames. However in some cases EMPTY_BUFFER is used instead of the actual buffer that was written. This may mask writes of released buffers or otherwise invalid buffer objects. It is also possible that if the buffer is invalid AbstractCoalescingBufferQueue will not release the aggregated buffer nor fail the associated promise.

Modifications:
- DefaultHttp2FrameCodec should take care to fail the promise, even if releasing the data throws
- AbstractCoalescingBufferQueue should release any aggregated data and fail the associated promise if something goes wrong during aggregation

Result:
More correct handling of invalid buffers in HTTP/2 code.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
@@ -15,7 +15,6 @@
 package io.netty.handler.codec.http2;
 
 import io.netty.buffer.ByteBuf;
-import io.netty.buffer.Unpooled;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandlerContext;
@@ -386,10 +385,10 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
             if (!endOfStream) {
                 if (queuedData == 0) {
                     // There's no need to write any data frames because there are only empty data frames in the queue
-                    // and it is not end of stream yet. Just complete their promises by writing an empty buffer.
+                    // and it is not end of stream yet. Just complete their promises by getting the buffer corresponding
+                    // to 0 bytes and writing it to the channel (to preserve notification order).
                     ChannelPromise writePromise = ctx.newPromise().addListener(this);
-                    queue.remove(0, writePromise).release();
-                    ctx.write(Unpooled.EMPTY_BUFFER, writePromise);
+                    ctx.write(queue.remove(0, writePromise), writePromise);
                     return;
                 }
 
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2FrameWriter.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2FrameWriter.java
@@ -173,13 +173,18 @@ public class DefaultHttp2FrameWriter implements Http2FrameWriter, Http2FrameSize
                 }
             } while (!lastFrame);
         } catch (Throwable t) {
-            if (needToReleaseHeaders) {
-                header.release();
+            try {
+                if (needToReleaseHeaders) {
+                    header.release();
+                }
+                if (needToReleaseData) {
+                    data.release();
+                }
+            } finally {
+                promiseAggregator.setFailure(t);
+                promiseAggregator.doneAllocatingPromises();
             }
-            if (needToReleaseData) {
-                data.release();
-            }
-            promiseAggregator.setFailure(t);
+            return promiseAggregator;
         }
         return promiseAggregator.doneAllocatingPromises();
     }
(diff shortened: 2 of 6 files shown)


codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2FrameWriter.java
```diff
@@ -264,27 +269,35 @@ public ChannelFuture writeSettingsAck(ChannelHandlerContext ctx, ChannelPromise
     }
 
     @Override
-    public ChannelFuture writePing(ChannelHandlerContext ctx, boolean ack, ByteBuf data,
-            ChannelPromise promise) {
-        boolean releaseData = true;
-        SimpleChannelPromiseAggregator promiseAggregator =
+    public ChannelFuture writePing(ChannelHandlerContext ctx, boolean ack, ByteBuf data, ChannelPromise promise) {
+        final SimpleChannelPromiseAggregator promiseAggregator =
```
Scottmitch: it is assumed this operation won't throw ... it makes the code simpler and if object allocation fails (of an internal object) we are very unlikely to be able to make progress. However if folks prefer the code can account for this operation throwing too.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2FrameWriter.java
```diff
@@ -264,27 +269,35 @@ public ChannelFuture writeSettingsAck(ChannelHandlerContext ctx, ChannelPromise
...
                 new SimpleChannelPromiseAggregator(promise, ctx.channel(), ctx.executor());
         try {
             verifyPingPayload(data);
             Http2Flags flags = ack ? new Http2Flags().ack(true) : new Http2Flags();
+            int payloadLength = data.readableBytes();
             ByteBuf buf = ctx.alloc().buffer(FRAME_HEADER_LENGTH);
```
carl-mastrangelo: buffer should be released in the catch below right?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2FrameWriter.java
```diff
@@ -393,21 +413,27 @@ public ChannelFuture writeWindowUpdate(ChannelHandlerContext ctx, int streamId,
...
+            try {
                 payload.release();
+            } finally {
+                promiseAggregator.setFailure(t);
+                promiseAggregator.doneAllocatingPromises();
             }
+            return promiseAggregator;
+        }
+        try {
+            ctx.write(payload, promiseAggregator.newPromise());
```
carl-mastrangelo: should this be a voidPromise?

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2ConnectionRoundtripTest.java
```diff
@@ -649,6 +653,79 @@ public void run() throws Http2Exception {
         assertFalse(clientChannel.isOpen());
     }
 
+    private enum WriteEmptyBufferMode {
+        SingleEndOfStream,
```
carl-mastrangelo: this is rather surprising, I have almost never seen enums not use CONSTANT_CASE

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -388,8 +387,7 @@ public void write(ChannelHandlerContext ctx, int allowedBytes) {
                     // There's no need to write any data frames because there are only empty data frames in the queue
                     // and it is not end of stream yet. Just complete their promises by writing an empty buffer.
```
fenik17: fix a comment too?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2FrameWriter.java
Scottmitch: no because it is written to `ctx` which transfers reference count ownership.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2FrameWriter.java
Scottmitch: can you explain why you think this? We don't want `promise` to be completed until the 2 writes in this method (frame header, and payload) are complete. `SimpleChannelPromiseAggregator` facilitates the combination of multiple promises events into a single promise notification, but it needs to know how many promise events to expect. `newPromise` actually just increments a counter and `doneAllocatingPromises` signifies the counter value is completed and notification is now allowed.

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2ConnectionRoundtripTest.java
Scottmitch: updated

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
Scottmitch: done

Scottmitch:
4.1 (35b0cd58fb)

PR #7366, opened 2017-11-04
ReadOnlyHttpHeaders

Scottmitch:
Motivation:
For use cases that create headers, but do not need to modify them a read only variant of HttpHeaders would be useful and may be able to provide better iteration performance for encoding.

Modifications:
- Introduce ReadOnlyHttpHeaders that is backed by a flat array

Result:
ReadOnlyHttpHeaders exists for non-modifiable HttpHeaders use cases.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/ReadOnlyHttpHeaders.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/ReadOnlyHttpHeaders.java
@@ -1,0 +1,339 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.codec.http;
+
+import io.netty.util.AsciiString;
+import io.netty.util.internal.UnstableApi;
+
+import java.util.AbstractMap.SimpleImmutableEntry;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.LinkedHashSet;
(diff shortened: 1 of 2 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/ReadOnlyHttpHeaders.java
```diff
@@ -0,0 +1,322 @@
...
+    private final AsciiString[] keyValuePairs;
+
+    public ReadOnlyHttpHeaders(boolean validateHeaders, AsciiString... keyValuePairs) {
+        if ((keyValuePairs.length & 1) != 0) {
+            throw newInvalidArraySizeException();
+        }
+        if (validateHeaders) {
+            validateHeaders(keyValuePairs);
+        }
+        this.keyValuePairs = keyValuePairs;
```
normanmaurer: do we need to call `clone()` to ensure we make a copy as people could still hold an reference on the original one ?

codec-http/src/main/java/io/netty/handler/codec/http/ReadOnlyHttpHeaders.java
```diff
@@ -0,0 +1,322 @@
...
+    @Override
+    public short getShort(CharSequence name, short defaultValue) {
+        AsciiString value = get0(name);
+        return value == null ? defaultValue : value.parseShort();
+    }
+
+    @Override
+    public Long getTimeMillis(CharSequence name) {
+        AsciiString value = get0(name);
+        return value == null ? null : CharSequenceValueConverter.INSTANCE.convertToTimeMillis(value);
```
normanmaurer: just delegate to `getTimeMillis(name, null)`?

codec-http/src/main/java/io/netty/handler/codec/http/ReadOnlyHttpHeaders.java
Scottmitch: We do not do this for `ReadOnlyHttp2Headers`, and instead leave it up to the user if they want to pay the cost to avoid external modifications. Let me clarify in the java doc.

codec-http/src/main/java/io/netty/handler/codec/http/ReadOnlyHttpHeaders.java
Scottmitch: Can you clarify? There is `getTimeMillis(CharSequence name, long defaultValue)` ... but this takes a primitive `long` (not a `Long`) so I can't pass `null`.

Scottmitch:
4.1 (8618a3351c)

Issue #7262, opened 2017-09-28
Get ClosedChannelException when close channel before SSL handshake


Jianzhu91:
Hi, recently I encounter a problem when using `Netty` and its `SlsHandler`. I would like to close a channel before SSL handshake based on my throttling configuration. I try to implement as below, but I will receive `ClosedChannelException` when I close the channel.

I add my `ThrottleHandler` before `SslHandler` in my `ChannelInitializer` . It overrides the `channelActive` method in order to determine whether we should do the SSL handshake or close the channel before the handshake. The `SslHandler` is setting in a server mode.

From my understanding, the SSL handshake should starts when the channel become active, so my closing operation should run before the SSL handshakes starts. However, I still receive a `ClosedChannelException` when closing the channel. This exception is triggered in `channelInactive` method in `SslHandler`. Does the handshake starts when the channel is active, or it just starts when you add `SslHandler` to the pipeline?

### Expected behavior
The channel should be closed normally without any exception. The handshake should not have started.

### Actual behavior
The `SslHandler` triggers `ClosedChannelException` when I close the channel before the handshake.

    java.nio.channels.ClosedChannelException
        at io.netty.handler.ssl.SslHandler.channelInactive(...)(Unknown Source) ~[netty-all-4.1.jar:4.1.6.Final]

### Steps to reproduce
Normally start a simple server which enables the SSL/TLS feature by adding a `SslHandler` into the `ChannelInitializer`. The `SslHandler` is setup with a Java `SSLEngine` in server mode. Then add the `ThrottleHandler` (codes are listed below) before `SslHandler`.

### Minimal yet complete reproducer code (or URL to code)
    @ChannelHandler.Sharable
    public class ThrottleHandler extends ChannelInboundHandlerAdapter {
        private ThrottleConfig config;
        private ThrottleConfig config;

        public ThrottleHandler(ThrottleConfig config) {
            super();
            this.config = config;
        }

        @Override
        public void channelActive(ChannelHandlerContext ctx) throws Exception {
            if (config.isThrottled) {
                log.info("Close a connection by throttle handler");
                ctx.close();
                return;
            }

            ctx.pipeline().remove(this);
            ctx.fireChannelActive();
        }
    }

    public class AbstractChannelInitializer extends ChannelInitializer<SocketChannel> {
        private ThrottleHandler throttleHandler;
        private SslHandler handshakeHandler;

        @Override
        protected void initChannel(SocketChannel channel) throws Exception {
[...]

PR #7368, opened 2017-11-05
Dont fire an SslHandshakeEvent if the handshake was not started at all.

normanmaurer:
Motivation:

We should not fire a SslHandshakeEvent if the channel is closed but the handshake was not started.

Modifications:

- Add a variable to SslHandler which tracks if an handshake was started yet or not and depending on this fire the event.
- Add a unit test

Result:

Fixes [#7262].

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -371,6 +371,7 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
     private boolean sentFirstMessage;
     private boolean flushedBeforeHandshake;
     private boolean readDuringHandshake;
+    private boolean handshakeStarted;
     private SslHandlerCoalescingBufferQueue pendingUnencryptedWrites;
     private Promise<Channel> handshakePromise = new LazyChannelPromise();
     private final LazyChannelPromise sslClosePromise = new LazyChannelPromise();
@@ -864,7 +865,7 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
 
     /**
      * This method will not call
-     * {@link #setHandshakeFailure(ChannelHandlerContext, Throwable, boolean)} or
+     * {@link #setHandshakeFailure(ChannelHandlerContext, Throwable, boolean, boolean)} or
      * {@link #setHandshakeFailure(ChannelHandlerContext, Throwable)}.
      * @return {@code true} if this method ends on {@link SSLEngineResult.HandshakeStatus#NOT_HANDSHAKING}.
      */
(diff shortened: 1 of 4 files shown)


handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1576,6 +1582,7 @@ public void handlerAdded(final ChannelHandlerContext ctx) throws Exception {
             } else {
                 applyHandshakeTimeout(null);
             }
+            handshakeStarted = true;
```
Scottmitch: should we set this before we call handshake, incase some sort of error happens internally?

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1501,16 +1502,21 @@ private void setHandshakeFailure(ChannelHandlerContext ctx, Throwable cause, boo
...
             pendingUnencryptedWrites.releaseAndFailAll(ctx, cause);
         }
     }
 
-    private void notifyHandshakeFailure(Throwable cause) {
+    private void notifyHandshakeFailure(Throwable cause, boolean notify) {
         if (handshakePromise.tryFailure(cause)) {
-            SslUtils.notifyHandshakeFailure(ctx, cause);
+            if (notify) {
+                SslUtils.notifyHandshakeFailure(ctx, cause);
```
Scottmitch: would it make sense to move the conditional logic into `SslUtils.notifyHandshakeFailure`? SniHandler may fire an exception if it sees something that is not an SSL record, and the handshake may not have started.

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
normanmaurer: sure why not.

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
normanmaurer: I guess we could.

normanmaurer:
@Scottmitch comments addressed

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
johnou: same thing in channelActive?

normanmaurer:
@johnou @Scottmitch PTAL again

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1592,14 +1593,19 @@ public void handlerAdded(final ChannelHandlerContext ctx) throws Exception {
...
-                // not be invoked. We have to initialize here instead.
-                handshake(null);
-            } else {
-                applyHandshakeTimeout(null);
-            }
+            startHandshakeProceessing();
+        }
+    }
+
+    private void startHandshakeProceessing() {
```
johnou: startHandshakeProceessing => startHandshakeProcessing

normanmaurer:
@johnou typo fixed.

johnou:
Lgtm

normanmaurer:
Cherry-picked into 4.1 (c921742a4281817052e06ce3bf28c859699e2cbf) and 4.0 (ae7436813a32f84a90e7069b1300b09ba4617dd8)

PR #7373, opened 2017-11-06
SslHandler leak

Scottmitch:
Motivation:
SslHandler only supports ByteBuf objects, but will not release objects of other types. SslHandler will also not release objects if its internal state is not correctly setup.

Modifications:
- Release non-ByteBuf objects in write
- Release all objects if the SslHandler queue is not setup

Result:
Less leaks in SslHandler.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -34,6 +34,7 @@ import io.netty.channel.ChannelPromise;
 import io.netty.channel.ChannelPromiseNotifier;
 import io.netty.handler.codec.ByteToMessageDecoder;
 import io.netty.handler.codec.UnsupportedMessageTypeException;
+import io.netty.util.ReferenceCountUtil;
 import io.netty.util.ReferenceCounted;
 import io.netty.util.concurrent.DefaultPromise;
 import io.netty.util.concurrent.EventExecutor;
@@ -656,6 +657,7 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
             pendingUnencryptedWrites.releaseAndFailAll(ctx,
                     new ChannelException("Pending write on removal of SslHandler"));
         }
+        pendingUnencryptedWrites = null;
         if (engine instanceof ReferenceCounted) {
             ((ReferenceCounted) engine).release();
         }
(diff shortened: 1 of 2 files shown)


handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
```diff
@@ -143,6 +144,33 @@ public void testTruncatedPacket() throws Exception {
...
+            }
+        };
+        try {
+            ch.write(referenceCounted).get();
+            fail();
+        } catch (ExecutionException e) {
+            assertThat(e.getCause(), is(instanceOf(UnsupportedMessageTypeException.class)));
+        }
+        assertEquals(0, referenceCounted.refCnt());
+        ch.finishAndReleaseAll();
```
normanmaurer: @Scottmitch assert the return value ?

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -701,10 +703,15 @@ public void read(ChannelHandlerContext ctx) throws Exception {
     @Override
     public void write(final ChannelHandlerContext ctx, Object msg, ChannelPromise promise) throws Exception {
         if (!(msg instanceof ByteBuf)) {
-            promise.setFailure(new UnsupportedMessageTypeException(msg, ByteBuf.class));
-            return;
+            UnsupportedMessageTypeException exception = new UnsupportedMessageTypeException(msg, ByteBuf.class);
+            ReferenceCountUtil.safeRelease(msg);
+            promise.setFailure(exception);
```
normanmaurer: inline the exception creation here ?

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
Scottmitch: done

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
Scottmitch: I wanted to wait to release until after `msg` was used on the line above. Currently is just uses the `class` but I didn't want to feed it a released object. wdyt?

normanmaurer: ah right.. good point.

Scottmitch:
4.1 (570d96d8c2) 4.0 (040130dee9)

PR #7376, opened 2017-11-06
Keep all leak records up to the target amount

carl-mastrangelo:
Motivation:
When looking for a leak, its nice to be able to request at least a
number of leaks.

Modification:

* Made all leak records up to the target amoutn recorded, and only
  then enable backing off.
* Enable recording more than 32 elements.  Previously the shift
  amount made this impossible.

Result:
Ability to record all accesses.

### DIFF
--- a/common/src/main/java/io/netty/util/ResourceLeakDetector.java
+++ b/common/src/main/java/io/netty/util/ResourceLeakDetector.java
@@ -386,15 +386,21 @@ public class ResourceLeakDetector<T> {
         /**
          * This method works by exponentially backing off as more records are present in the stack. Each record has a
          * 1 / 2^n chance of dropping the top most record and replacing it with itself. This has a number of convenient
-         * properties. First, the first record is always recorded. Second, the very last access will always be
-         * recorded. Third, an arbitrary number of accesses can be accepted, rather than just the last few. Fourth,
-         * it is easy to keep a precise record of the number of elements in the stack, since each element has to know
-         * how tall the stack is.
+         * properties:
+         *
+         * <ol>
+         * <li>  The current record is always recorded. This is due to the compare and swap dropping the top most
+         *       record, rather than the to-be-pushed record.
+         * <li>  The very last access will always be recorded. This comes as a property of 1.
+         * <li>  It is possible to retain more records than the target, based upon the probability distribution.
+         * <li>  It is easy to keep a precise record of the number of elements in the stack, since each element has to
+         *     know how tall the stack is.
+         * </ol>
          *
          * In this particular implementation, there are also some advantages. A thread local random is used to decide
          * if something should be recorded. This means that if there is a deterministic access pattern, it is now
-         * possible to see what other accesses occur, rather than always dropping them. Second, there is roughly a
-         * linear ramp up to {@link #TARGET_RECORDS}, after which backoff occurs. This matches typical access patterns,
+         * possible to see what other accesses occur, rather than always dropping them. Second, after
+         * {@link #TARGET_RECORDS} accesses, backoff occurs. This matches typical access patterns,
          * where there are either a high number of accesses (i.e. a cached buffer), or low (an ephemeral buffer), but
          * not many in between.
          *
@@ -415,9 +421,14 @@ public class ResourceLeakDetector<T> {
                         // already closed.
                         return;
                     }
-                    int numElements = oldHead.pos + 1;
-                    if (dropped = PlatformDependent.threadLocalRandom().nextInt(1 << numElements) >= TARGET_RECORDS) {
-                        prevHead = oldHead.next;
+                    final int numElements = oldHead.pos + 1;
+                    if (numElements >= TARGET_RECORDS) {
+                        final int backOffFactor = Math.min(numElements - TARGET_RECORDS, 30);
+                        if (dropped = PlatformDependent.threadLocalRandom().nextInt(1 << backOffFactor) != 0) {
+                            prevHead = oldHead.next;
+                        }
+                    } else {
+                        dropped = false;
                     }
                     newHead = hint != null ? new Record(prevHead, hint) : new Record(prevHead);
                 } while (!headUpdater.compareAndSet(this, oldHead, newHead));
@@ -458,10 +469,7 @@ public class ResourceLeakDetector<T> {
 
         @Override
         public String toString() {
-            Record oldHead;
-            do {
-                oldHead = headUpdater.get(this);
-            } while (!headUpdater.compareAndSet(this, oldHead, null));
+            Record oldHead = headUpdater.getAndSet(this, null);
             if (oldHead == null) {
                 // Already closed
                 return EMPTY_STRING;


common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -411,13 +411,17 @@ private void record0(Object hint) {
...
                         // already closed.
                         return;
                     }
-                    int numElements = oldHead.pos + 1;
-                    if (dropped = PlatformDependent.threadLocalRandom().nextInt(1 << numElements) >= TARGET_RECORDS) {
-                        prevHead = oldHead.next;
+                    final int numElements = oldHead.pos + 1;
+                    if (numElements >= TARGET_RECORDS) {
+                        final int backOffFactor = numElements - TARGET_RECORDS;
+                        if (dropped = PlatformDependent.threadLocalRandom().nextInt(1 << backOffFactor) != 0) {
```
normanmaurer: @carl-mastrangelo shouldn't you not update `dropped` it anymore once you set `dropped` to `true` ?

carl-mastrangelo: dropped needs to be reset because if it loses the compareAndSet race, it may retry without dropping a record the second time around.

carl-mastrangelo: It could be added to an else clause instead of being done unconditionally at the top.

Scottmitch: `backOffFactor` may overflow. if `backOffFactor == 31` this will result in an IllegalArgumentException. It maybe unlikely as `numElements` grows, but consider protecting for this case just incase we are unlucky.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -411,13 +411,17 @@ private void record0(Object hint) {
...
                     if ((prevHead = oldHead = headUpdater.get(this)) == null) {
                         // already closed.
                         return;
                     }
-                    int numElements = oldHead.pos + 1;
-                    if (dropped = PlatformDependent.threadLocalRandom().nextInt(1 << numElements) >= TARGET_RECORDS) {
-                        prevHead = oldHead.next;
+                    final int numElements = oldHead.pos + 1;
+                    if (numElements >= TARGET_RECORDS) {
+                        final int backOffFactor = numElements - TARGET_RECORDS;
```
Scottmitch: IIUC the behavior of this code is as follow: We retain every record until we hit `TARGET_RECORDS`, and then we  retain subsequent records according to the probability distribution which decreases exponentially as records are added.

I'm wondering if this approach/complexity is justified vs. just retain until max count, and then drop (maybe always retain the last record). Either way it seems like we would have to touch atomic state, and the outcome of seeing "X records were dropped" most likely means that you will have to increase `TARGET_RECORDS` to understand the full context of the leak anyways.

Scottmitch:
@carl-mastrangelo - On line 467 there is the following code:

```java
            Record oldHead;
            do {
                oldHead = headUpdater.get(this);
            } while (!headUpdater.compareAndSet(this, oldHead, null));
```

Consider simplifying to:

```java
Record oldHead = headUpdater.getAndSet(this, null);
```

carl-mastrangelo:
@Scottmitch done.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
carl-mastrangelo: Done

common/src/main/java/io/netty/util/ResourceLeakDetector.java
carl-mastrangelo: Your understanding is correct.  

Whether or not the complexity is justified is something I can't decide, since I haven't had to debug anything complicated.     That said:

*  Retaining until max count has a problem: there isn't a good answer for duplicates.  In the previous version, 10x the number of references were actually stored in order to deduplicate them.  This solution pushes back on them using a single knob
* This solution is (IMHO) simpler than the previous one.  There is only one pointer to deal with, rather than two.  
* This is an online solution.  You don't have to restart your program to have a chance of seeing all the accesses.  Consider debugging a problem that would be inconvenient to restart.  Each leak will get a different set of traces.  It also handles odd cases, like having too many or too few traces.

The backoff is what makes this scale.  It avoids doing writes in the good case, while not overly hurting the leaking case.

Scottmitch: > Retaining until max count has a problem: there isn't a good answer for duplicates. 

Can you provide an example as to where this approach helps with duplicates? IIUC the approaches are the same until the limit is reached, and then the probability of retention decreases as records increase. You may do better with "duplicates", but you may not ... in the end this just retains some extra frames, right?

> There is only one pointer to deal with, rather than two.

Would the approach have to change significantly? From a concurrency perspective we would just be either dropping new records, or replacing the last record. If the priority portion was removed what else would have to change?

> Each leak will get a different set of traces. It also handles odd cases, like having too many or too few traces.

As the number of traces increases, the probability of retaining a trace decreases. So it is possible to see other traces it becomes more unlikely. If the number of records exceeds `TARGET_RECORDS` it still seems likely that you will have to increase the count to understand what is going on. Once the limit is exceeded the stack frames may no longer be "sequential", which seems likely to cause confusion. If we think there is value in a probabilistic retention it seems like at a minimum the leak report should indicate which frames are no longer "sequential".

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -411,13 +411,17 @@ private void record0(Object hint) {
...
                     if ((prevHead = oldHead = headUpdater.get(this)) == null) {
                         // already closed.
                         return;
                     }
-                    int numElements = oldHead.pos + 1;
-                    if (dropped = PlatformDependent.threadLocalRandom().nextInt(1 << numElements) >= TARGET_RECORDS) {
-                        prevHead = oldHead.next;
+                    final int numElements = oldHead.pos + 1;
+                    if (numElements >= TARGET_RECORDS) {
+                        final int backOffFactor = Math.min(numElements - TARGET_RECORDS, 31);
```
Scottmitch: `1<<31 == -2147483648` which will throw an IllegalArgumentException. this should max out at `30` or you could consider using `(1 << backOffFactor) - 1` instead.

carl-mastrangelo: I picked 30.  Done.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
carl-mastrangelo: So high level: there are places where this new code falls short.  There are cases that will be dropped, and shortcomings.  I still feel it is better than what was previously there.  This code has a number of upsides as described in the comments and the PR description.  Let's not let the perfect be the enemy of the good.

> Can you provide an example as to where this approach helps with duplicates? IIUC the approaches are the same until the limit is reached, and then the probability of retention decreases as records increase. You may do better with "duplicates", but you may not ... in the end this just retains some extra frames, right?

So to be clear, this code doesn't implement what I am about to describe but:  If the backoff code does decide to run, it would be okay to see if the top most stack element is the same, since backoff makes this happen more seldomly.  Right now it discards most duplicates, which the previous code couldn't do.

> Would the approach have to change significantly? From a concurrency perspective we would just be either dropping new records, or replacing the last record. If the priority portion was removed what else would have to change?

Well, I haven't thought extensively about the implementation, but I think the behavior is incorrect.  Previously:

> I'm wondering if this approach/complexity is justified vs. just retain until max count, and then drop

This would ensure that you always have to raise the number of retained refs extremely high.  Often its the later refs that are important, which the code as is does (and keeps some of the middle refs).    It would be better if only the last few accesses were kept rather than the first few, but thats an even more complicated thing.

> If we think there is value in a probabilistic retention it seems like at a minimum the leak report should indicate which frames are no longer "sequential".

I think there is value (after about 40 refs, it starts becoming difficult to reason about the access order).  I'm not sure what you mean by sequential, since the accesses are concurrent.  Could you clarify?

Scottmitch: > This code has a number of upsides as described in the comments

Can you clarify some of the statements in the comments:

How is this accomplished:
`Second, the very last access will always be recorded`

"Arbitrary" maybe a bit strong here. "It is possible to retain more records than the limit, based upon the probability distribution..." seems more accurate.
`Third, an arbitrary number of accesses can be accepted, rather than just the last few`

Why does the priority distribution have any relation to the stack storage and counting elements?
`Fourth, it is easy to keep a precise record of the number of elements in the stack, since each element has to know how tall the stack is`

This is only true after we have reached the limit. So the thread local random is additional complexity on top of enforcing the limit.
`A thread local random is used to decide if something should be recorded. This means that if there is a deterministic access pattern, it is now possible to see what other accesses occur, rather than always dropping them.`

> Right now it discards most duplicates, which the previous code couldn't do.

I still don't understand how "it discards most duplicates". There is no duplicate detection, but instead records are randomly thrown away. When I think of "discarding duplicates" I think of logic like "it is good to throw this away because I know I have already recorded it (or even it is likely I have already recorded it after some threshold is exceeded)".

> Well, I haven't thought extensively about the implementation, but I think the behavior is incorrect.

I don't understand why dropping records is considered "incorrect" in this case. We are discussing the behavior when limits are exceeded, and in the current PR a "best effort" is made to retain more, but not too much more, as opposed to retaining a fixed amount. Either way records will likely be dropped so why is one more "correct" than the other. IIUC the proposed value is we "may" be able to get a few extra frames which "may" provide more context, and that is what I'm questioning ... I don't understand a real use case where this would be beneficial.

> Often its the later refs that are important, which the code as is does (and keeps some of the middle refs).

Once the limit is reached the current code will drop with increasing probability. This means it is increasingly likely to *not* retain the later refs after the limit is exceeded.

> I'm not sure what you mean by sequential, since the accesses are concurrent. Could you clarify?

Yes the behavior isn't quite "sequential" (that is what the quotes are for), but it maybe sequential depending upon the access pattern (e.g. always accessed from inside the event loop), or "sequential" enough to understand how the object propagated around and untimely to understand the reference count ownership transitions. The interesting part is once we start using a probability distribution restoring the context may become more difficult to piece together because of the non-deterministic holes.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -386,15 +386,19 @@ public void record(Object hint) {
         /**
          * This method works by exponentially backing off as more records are present in the stack. Each record has a
          * 1 / 2^n chance of dropping the top most record and replacing it with itself. This has a number of convenient
-         * properties. First, the first record is always recorded. Second, the very last access will always be
-         * recorded. Third, an arbitrary number of accesses can be accepted, rather than just the last few. Fourth,
-         * it is easy to keep a precise record of the number of elements in the stack, since each element has to know
-         * how tall the stack is.
+         * properties:
+         *
+         * 1.  The current record is always recorded. This is due to the compare and swap dropping the top most record,
```
Scottmitch: nit: use java doc / html formatting (e.g `<ol><li></li></ol>`)

common/src/main/java/io/netty/util/ResourceLeakDetector.java
carl-mastrangelo: Updated to clarify comments.   Some more clarifying things:

> I still don't understand how "it discards most duplicates".
In the previous version of this code, all accesses to a refcounted object were kept up to the max limit (10x the expected limit) in order to deduplicate the accesses on a leak.   I dont know why 10 was picked, but it seems to be a balance between trying to get the target number of unique accesses and limiting memory usage.  

The new version has a better story:  Ignore caring about duplicates all together.  There are dupes, and that's okay.  IMO the interesting record is the very last one, which was likely preceded by lots of dupes.  The previous code had to try and deal with this case.  The new code handles it.  Most of the dupes are discarded at record time, rather than retained.  The interesting accesses are mixed in.  They aren't for sure going to be recorded, but they at least have a chance.

By keeping a fixed number, or having an exact group of accesses saved, you _ensure_ that there is an access pattern that is useful but is missed.

carl-mastrangelo:
updates made and squashed

normanmaurer:
@Scottmitch agreed... let me pull this in once the CI finishes and then cut a release. /cc @carl-mastrangelo

Scottmitch:
4.1 (ef5ebb40c9) 4.0 (849147f428)

Issue #7378, opened 2017-11-07
Channel promise of a write is successful, although SSL handshake fails.


s-gheldd:
### Expected behavior
When a SSL handshake fails, a channel promise of a write operation should fail aswell. 

### Actual behavior
The channel promise succeeds. 

### Steps to reproduce

    $ git clone https://github.com/s-gheldd/netty-reproducer.git
    $ cd netty-reproducer
    $ mvn test

### Minimal yet complete reproducer code (or URL to code)

https://github.com/s-gheldd/netty-reproducer

### Netty version

`4.1.16.Final`

(last version that had the expected behavior `4.1.13.Final`)

### JVM version (e.g. `java -version`)

    java version "1.7.0_80"
    Java(TM) SE Runtime Environment (build 1.7.0_80-b15)
    Java HotSpot(TM) 64-Bit Server VM (build 24.80-b11, mixed mode)

### OS version (e.g. `uname -a`)

    Darwin ghelds-MacBook-Pro.local 15.6.0 Darwin Kernel Version 15.6.0: Mon Oct  2 22:20:08 PDT 2017; root:xnu-3248.71.4~1/RELEASE_X86_64 x86_64

Scottmitch:
See https://github.com/netty/netty/pull/7380

PR #7380, opened 2017-11-07
SslHandler promise completion incorrect if write doesn't immediately

Scottmitch:
complete

Motivation:
SslHandler removes a Buffer/Promise pair from
AbstractCoalescingBufferQueue when wrapping data. However it is possible
the SSLEngine will not consume the entire buffer. In this case
SslHandler adds the Buffer back to the queue, but doesn't add the
Promise back to the queue. This may result in the promise completing
immediately in finishFlush, and generally not correlating to the
completion of writing the corresponding Buffer

Modifications:
- AbstractCoalescingBufferQueue#addFirst should also support adding the
ChannelPromise
- In the event of a handshake timeout we should immediately fail pending
writes immediately to get a more accurate exception

Result:
Fixes https://github.com/netty/netty/issues/7378.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -794,7 +794,10 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
                     return;
                 } else {
                     if (buf.isReadable()) {
-                        pendingUnencryptedWrites.addFirst(buf);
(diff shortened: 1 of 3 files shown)


Scottmitch:
@s-gheldd - FYI. Thanks for the reproducer!

s-gheldd:
Thanks for the extremely quick reaction.

Scottmitch:
4.1 (8c5eeb581e)

PR #7381, opened 2017-11-07
AbstractCoalescingBufferQueue addFirst void promise handling

Scottmitch:
Motivation:
AbstractCoalescingBufferQueue#add accounts for void promises, but AbstractCoalescingBufferQueue#addFirst does not. These methods should be consistent.

Modifications:
- AbstractCoalescingBufferQueue#addFirst should account for void promises and share code with AbstractCoalescingBufferQueue#add

Result:
More correct void promise handling in AbstractCoalescingBufferQueue.


Scottmitch:
unit test added

Scottmitch:
4.1 (7511c15187)

PR #7382, opened 2017-11-08
Use Parameterized to run SslHandler tests with different SslProviders.

normanmaurer:
Motivation:

At the moment use loops to run SslHandler tests with different SslProviders which is error-prone and also make it difficult to understand with which provider these failed.

Modifications:

- Move unit tests that should run with multiple SslProviders to extra class.
- Use junit Parameterized to run with different SslProvider combinations

Result:

Easier to understand which SslProvider produced test failures


normanmaurer:
Cherry-picked into 4.1 (2adb8bd80f5d19d049ffb7257172cb8d6f1273d2) and 4.0 (d83de7f81)

PR #7384, opened 2017-11-08
Use Netty's DefaultPriorityQueue instead of JDK's PriorityQueue for s…

anuraaga:
…cheduled tasks.

Motivation:

`AbstractScheduledEventExecutor` uses a standard `java.util.PriorityQueue` to keep track of task deadlines. `ScheduledFuture.cancel` removes tasks from this `PriorityQueue`. Unfortunately, `PriorityQueue.remove` has `O(n)` performance since it must search for the item in the entire queue before removing it. This is fast when the future is at the front of the queue (e.g., already triggered) but not when it's randomly located in the queue.

Many servers will use `ScheduledFuture.cancel` on all requests, e.g., to manage a request timeout. As these cancellations will be happen in arbitrary order, when there are many scheduled futures, `PriorityQueue.remove` is a bottleneck and greatly hurts performance with many concurrent requests (>10K).

Modification:

Use netty's `DefaultPriorityQueue` for scheduling futures instead of the JDK. `DefaultPriorityQueue` is almost identical to the JDK version except it is able to remove futures without searching for them in the queue. This means `DefaultPriorityQueue.remove` has `O(log n)` performance.

Result:

Before - cancelling futures has varying performance, capped at `O(n)`
After - cancelling futures has stable performance, capped at `O(log n)`

Benchmark results

After - cancelling in order and in reverse order have similar performance within `O(log n)` bounds
```
Benchmark                                           (num)   Mode  Cnt       Score      Error  Units
ScheduledFutureTaskBenchmark.cancelInOrder            100  thrpt   20  137779.616 ± 7709.751  ops/s
ScheduledFutureTaskBenchmark.cancelInOrder           1000  thrpt   20   11049.448 ±  385.832  ops/s
ScheduledFutureTaskBenchmark.cancelInOrder          10000  thrpt   20     943.294 ±   12.391  ops/s
ScheduledFutureTaskBenchmark.cancelInOrder         100000  thrpt   20      64.210 ±    1.824  ops/s
ScheduledFutureTaskBenchmark.cancelInReverseOrder     100  thrpt   20  167531.096 ± 9187.865  ops/s
ScheduledFutureTaskBenchmark.cancelInReverseOrder    1000  thrpt   20   33019.786 ± 4737.770  ops/s
ScheduledFutureTaskBenchmark.cancelInReverseOrder   10000  thrpt   20    2976.955 ±  248.555  ops/s
ScheduledFutureTaskBenchmark.cancelInReverseOrder  100000  thrpt   20     362.654 ±   45.716  ops/s
```

Before - cancelling in order and in reverse order have significantly different performance at higher queue size, orders of magnitude worse than the new implementation.
```
Benchmark                                           (num)   Mode  Cnt       Score       Error  Units
ScheduledFutureTaskBenchmark.cancelInOrder            100  thrpt   20  139968.586 ± 12951.333  ops/s
ScheduledFutureTaskBenchmark.cancelInOrder           1000  thrpt   20   12274.420 ±   337.800  ops/s
ScheduledFutureTaskBenchmark.cancelInOrder          10000  thrpt   20     958.168 ±    15.350  ops/s
ScheduledFutureTaskBenchmark.cancelInOrder         100000  thrpt   20      53.381 ±    13.981  ops/s
ScheduledFutureTaskBenchmark.cancelInReverseOrder     100  thrpt   20  123918.829 ±  3642.517  ops/s
ScheduledFutureTaskBenchmark.cancelInReverseOrder    1000  thrpt   20    5099.810 ±   206.992  ops/s
ScheduledFutureTaskBenchmark.cancelInReverseOrder   10000  thrpt   20      72.335 ±     0.443  ops/s
ScheduledFutureTaskBenchmark.cancelInReverseOrder  100000  thrpt   20       0.743 ±     0.003  ops/s
```

### DIFF
--- a/common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java
+++ b/common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java
@@ -54,7 +54,9 @@ public abstract class AbstractScheduledEventExecutor extends AbstractEventExecut
     PriorityQueue<ScheduledFutureTask<?>> scheduledTaskQueue() {
         if (scheduledTaskQueue == null) {
             scheduledTaskQueue = new DefaultPriorityQueue<ScheduledFutureTask<?>>(
-                    SCHEDULED_FUTURE_TASK_COMPARATOR, 11);
+                    SCHEDULED_FUTURE_TASK_COMPARATOR,
+                    // Use same initial capacity as java.util.PriorityQueue
+                    11);
         }
         return scheduledTaskQueue;
     }
--- a/common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java
+++ b/common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java
@@ -15,9 +15,11 @@
  */
 package io.netty.util.concurrent;
 
+import io.netty.util.internal.DefaultPriorityQueue;
 import io.netty.util.internal.ObjectUtil;
+import io.netty.util.internal.PriorityQueue;
 
-import java.util.PriorityQueue;
+import java.util.Comparator;
 import java.util.Queue;
 import java.util.concurrent.Callable;
 import java.util.concurrent.Executors;
@@ -28,7 +30,15 @@ import java.util.concurrent.TimeUnit;
  */
 public abstract class AbstractScheduledEventExecutor extends AbstractEventExecutor {
 
-    Queue<ScheduledFutureTask<?>> scheduledTaskQueue;
+    private static final Comparator<ScheduledFutureTask<?>> SCHEDULED_FUTURE_TASK_COMPARATOR =
+            new Comparator<ScheduledFutureTask<?>>() {
+                @Override
+                public int compare(ScheduledFutureTask<?> o1, ScheduledFutureTask<?> o2) {
+                    return o1.compareTo(o2);
+                }
+            };
+
+    PriorityQueue<ScheduledFutureTask<?>> scheduledTaskQueue;
 
     protected AbstractScheduledEventExecutor() {
     }
@@ -41,14 +51,15 @@ public abstract class AbstractScheduledEventExecutor extends AbstractEventExecut
         return ScheduledFutureTask.nanoTime();
     }
 
-    Queue<ScheduledFutureTask<?>> scheduledTaskQueue() {
+    PriorityQueue<ScheduledFutureTask<?>> scheduledTaskQueue() {
         if (scheduledTaskQueue == null) {
-            scheduledTaskQueue = new PriorityQueue<ScheduledFutureTask<?>>();
+            scheduledTaskQueue = new DefaultPriorityQueue<ScheduledFutureTask<?>>(
+                    SCHEDULED_FUTURE_TASK_COMPARATOR, 11);
         }
         return scheduledTaskQueue;
     }
 
-    private static  boolean isNullOrEmpty(Queue<ScheduledFutureTask<?>> queue) {
+    private static boolean isNullOrEmpty(Queue<ScheduledFutureTask<?>> queue) {
         return queue == null || queue.isEmpty();
     }
 
@@ -59,7 +70,7 @@ public abstract class AbstractScheduledEventExecutor extends AbstractEventExecut
      */
     protected void cancelScheduledTasks() {
         assert inEventLoop();
-        Queue<ScheduledFutureTask<?>> scheduledTaskQueue = this.scheduledTaskQueue;
+        PriorityQueue<ScheduledFutureTask<?>> scheduledTaskQueue = this.scheduledTaskQueue;
         if (isNullOrEmpty(scheduledTaskQueue)) {
             return;
         }
@@ -71,7 +82,7 @@ public abstract class AbstractScheduledEventExecutor extends AbstractEventExecut
             task.cancelWithoutRemove(false);
         }
 
-        scheduledTaskQueue.clear();
+        scheduledTaskQueue.clearIgnoringIndexes();
     }
 
     /**
@@ -205,7 +216,7 @@ public abstract class AbstractScheduledEventExecutor extends AbstractEventExecut
 
     final void removeScheduled(final ScheduledFutureTask<?> task) {
         if (inEventLoop()) {
-            scheduledTaskQueue().remove(task);
+            scheduledTaskQueue().removeTyped(task);
         } else {
             execute(new Runnable() {
                 @Override
--- a/common/src/main/java/io/netty/util/internal/DefaultPriorityQueue.java
+++ b/common/src/main/java/io/netty/util/internal/DefaultPriorityQueue.java
@@ -78,6 +78,11 @@ public final class DefaultPriorityQueue<T extends PriorityQueueNode> extends Abs
     }
 
     @Override
+    public void clearIgnoringIndexes() {
+        size = 0;
+    }
+
+    @Override
     public boolean offer(T e) {
         if (e.priorityQueueIndex(this) != INDEX_NOT_IN_QUEUE) {
             throw new IllegalArgumentException("e.priorityQueueIndex(): " + e.priorityQueueIndex(this) +
--- a/common/src/main/java/io/netty/util/internal/PriorityQueue.java
+++ b/common/src/main/java/io/netty/util/internal/PriorityQueue.java
@@ -34,4 +34,13 @@ public interface PriorityQueue<T> extends Queue<T> {
      * @param node An object which is in this queue and the priority may have changed.
      */
     void priorityChanged(T node);
+
+    /**
+     * Removes all of the elements from this {@link PriorityQueue} without calling
+     * {@link PriorityQueueNode#priorityQueueIndex(DefaultPriorityQueue)} or explicitly removing references to them to
+     * allow them to be garbage collected. This should only be used when it is certain that the nodes will not be
+     * re-inserted into this or any other {@link PriorityQueue} and it is known that the {@link PriorityQueue} itself
+     * will be garbage collected after this call.
+     */
+    void clearIgnoringIndexes();
 }
--- a/microbench/src/main/java/io/netty/microbench/concurrent/ScheduledFutureTaskBenchmark.java
+++ b/microbench/src/main/java/io/netty/microbench/concurrent/ScheduledFutureTaskBenchmark.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.microbench.concurrent;
+
+import io.netty.channel.DefaultEventLoop;
+import io.netty.channel.EventLoop;
+import io.netty.microbench.util.AbstractMicrobenchmark;
+import io.netty.util.concurrent.Future;
+import io.netty.util.concurrent.ScheduledFuture;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
+
+@State(Scope.Benchmark)
+public class ScheduledFutureTaskBenchmark extends AbstractMicrobenchmark {
+
+    static final EventLoop executor = new DefaultEventLoop();
+
+    @State(Scope.Thread)
+    public static class FuturesHolder {
+
+        private static final Callable<Void> NO_OP = new Callable<Void>() {
+            @Override
+            public Void call() throws Exception {
+                return null;
+            }
+        };
+
+        @Param({ "100", "1000", "10000", "100000" })
+        int num;
+
+        final List<ScheduledFuture<Void>> futures = new ArrayList<ScheduledFuture<Void>>();
+
+        @Setup(Level.Invocation)
+        public void reset() {
+            futures.clear();
+            executor.submit(new Runnable() {
+                @Override
+                public void run() {
+                    for (int i = 1; i <= num; i++) {
+                        futures.add(executor.schedule(NO_OP, i, TimeUnit.HOURS));
+                    }
+                }
+            }).syncUninterruptibly();
+        }
+    }
+
+    @TearDown(Level.Trial)
+    public void stop() throws Exception {
+        executor.shutdownGracefully().syncUninterruptibly();
+    }
+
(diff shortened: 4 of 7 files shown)


anuraaga:
Build failure is the same one I had when building locally, before running `mvn install`. Snapshots haven't been published?

microbench/src/main/java/io/netty/microbench/concurrent/ScheduledFutureTaskBenchmark.java
```diff
@@ -0,0 +1,94 @@
...
+    public void cancelInOrder(Blackhole bh, final FuturesHolder futures) {
+        for (int i = 0; i < futures.num; i++) {
+            bh.consume(cancelFuture(futures.futures.get(i)));
+        }
+    }
+
+    @Benchmark
+    public void cancelInReverseOrder(Blackhole bh, final FuturesHolder futures) {
+        for (int i = futures.num - 1; i >= 0; i--) {
+            bh.consume(cancelFuture(futures.futures.get(i)));
```
Scottmitch: this benchmark has non trivial amount of overhead. suggestions:

- execute the test in the eventloop thread. executing a Runnable an blocking on the future is likely to have a high enough overhead to make the actual queue difference in the noise
- consider trying alternatives to Blackhole (e.g. return some value from the method that is the result of the computation).

common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java
```diff
@@ -15,19 +15,28 @@
...
 
+    private static final Comparator<ScheduledFutureTask<?>> SCHEDULED_FUTURE_TASK_COMPARATOR =
+            new Comparator<ScheduledFutureTask<?>>() {
+                @Override
+                public int compare(ScheduledFutureTask<?> o1, ScheduledFutureTask<?> o2) {
+                    return o1.compareTo(o2);
+                }
+            };
+
     Queue<ScheduledFutureTask<?>> scheduledTaskQueue;
```
Scottmitch: nit: if you use `io.netty.util.internal.PriorityQueue` then you have access to `removeTyped` which should remove some casting overhead.

microbench/src/main/java/io/netty/microbench/concurrent/ScheduledFutureTaskBenchmark.java
anuraaga: Can you give some pointers on running the test on the eventloop thread? Tried using `AbstractSharedExecutorMicrobenchmark` but it seemed to deadlock between the benchmark thread (JMH isn't asynchronous) and scheduler, which makes sense. Do you mean to use `AuxCounters`?

FWIW, even with the overhead, the benchmark results are pretty clear IMO.

common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java
Scottmitch: Also `clear()` is more expensive by default on `DefaultPriorityQueue` because it iterates through all tasks to reset the queue index. Consider adding another method which doesn't iterate through the collection to fix indexes and instead just sets size to `0` (e.g. `clearIgnoreIndexes()`).

microbench/src/main/java/io/netty/microbench/concurrent/ScheduledFutureTaskBenchmark.java
Scottmitch: > Tried using AbstractSharedExecutorMicrobenchmark

The goal here is to supply JMH with the thread (an EventLoop) to run the benchmarks on. Its been a while since this was used, and I'm not surprised if there are issues here.

> Can you give some pointers on running the test on the eventloop thread?

The approach attempted by `AbstractSharedExecutorMicrobenchmark` would be ideal, but if there are issues with this approach then consider executing a runnable on the event loop which does all the work, and waiting for that to continue. This would reduce the context switching and Runnable allocation back/forth to just 1 per benchmark iteration instead of 1 for every schedule call.

> FWIW, even with the overhead, the benchmark results are pretty clear IMO.

I would expect the results to be more pronounced. We don't want to benchmark promise notification and context switching for each execution. Including "extras" in the benchmark makes the end results more volatile and more difficult to asses.

anuraaga: Ah got it - thanks for the suggestion went with scheduling once per invocation.

common/src/main/java/io/netty/util/internal/DefaultPriorityQueue.java
```diff
@@ -78,6 +78,14 @@ public void clear() {
     }
 
     @Override
+    public void clearIgnoringIndexes() {
+        for (int i = 0; i < size; ++i) {
```
Scottmitch: If we have to iterate the array then I don't think this method is worth it. My intention for suggesting this method would be to avoid iteration of the queue. My suggestion would be we do the following:

1. don't iterate the array, and just set size to 0. (we should clarify in javadocs): it is useful in cases were GC nepotism is known not to be an issue because the queue object itself will be made eligible for GC.
2. set queue to EMPTY_ARRAY, and size to 0. the negative would be we may have to ramp up again if the queue is used after this method is called.

Option (2) is the more conservative approach. We should clarify the behavior in the javadocs, and if we need to in the future we could consider providing a "resize hint" to avoid the size ramp up overhead.

common/src/main/java/io/netty/util/internal/PriorityQueue.java
```diff
@@ -34,4 +34,10 @@
      * @param node An object which is in this queue and the priority may have changed.
      */
     void priorityChanged(T node);
+
+    /**
+     * Removes all of the elements from this {@link PriorityQueue}, without resetting nodes. This should only be used
```
Scottmitch: `without resetting nodes` -> `without calling {@link PriorityQueueNode#priorityQueueIndex(DefaultPriorityQueue)} on any of the nodes.

microbench/src/main/java/io/netty/microbench/concurrent/ScheduledFutureTaskBenchmark.java
```diff
@@ -0,0 +1,92 @@
...
+        @Param({ "100", "1000", "10000", "100000" })
+        private int num;
+
+        private final List<ScheduledFuture<Void>> futures = new ArrayList<ScheduledFuture<Void>>();
+
+        @Setup(Level.Invocation)
+        public void reset() {
+            futures.clear();
+            for (int i = 1; i <= num; i++) {
+                futures.add(executor.schedule(new Callable<Void>() {
```
Scottmitch: consider executing this code on the event loop too (just like in the benchmarks).

microbench/src/main/java/io/netty/microbench/concurrent/ScheduledFutureTaskBenchmark.java
```diff
@@ -0,0 +1,92 @@
...
+            public void run() {
+                for (int i = 0; i < futures.num; i++) {
+                    futures.futures.get(i).cancel(false);
+                }
+            }
+        }).syncUninterruptibly();
+    }
+
+    @Benchmark
+    public Future<?> cancelInReverseOrder(Blackhole bh, final FuturesHolder futures) {
```
Scottmitch: remove `Blackhole bh`

microbench/src/main/java/io/netty/microbench/concurrent/ScheduledFutureTaskBenchmark.java
```diff
@@ -0,0 +1,92 @@
...
+        executor.shutdownGracefully();
+    }
+
+    @Benchmark
+    public Future<?> cancelInOrder(final FuturesHolder futures) {
+        return executor.submit(new Runnable() {
+            @Override
+            public void run() {
+                for (int i = 0; i < futures.num; i++) {
+                    futures.futures.get(i).cancel(false);
```
Scottmitch: `futures.futures` is unclear. consider clarifying by renaming `futuresHolder.futures`

microbench/src/main/java/io/netty/microbench/concurrent/ScheduledFutureTaskBenchmark.java
```diff
@@ -0,0 +1,92 @@
...
+
+    private static final EventLoop executor = new DefaultEventLoop();
+
+    @State(Scope.Thread)
+    public static class FuturesHolder {
+
+        @Param({ "100", "1000", "10000", "100000" })
+        private int num;
+
+        private final List<ScheduledFuture<Void>> futures = new ArrayList<ScheduledFuture<Void>>();
```
Scottmitch: consider making this package private because it is used outside the class

microbench/src/main/java/io/netty/microbench/concurrent/ScheduledFutureTaskBenchmark.java
```diff
@@ -0,0 +1,92 @@
...
+@State(Scope.Benchmark)
+public class ScheduledFutureTaskBenchmark extends AbstractMicrobenchmark {
+
+    private static final EventLoop executor = new DefaultEventLoop();
+
+    @State(Scope.Thread)
+    public static class FuturesHolder {
+
+        @Param({ "100", "1000", "10000", "100000" })
+        private int num;
```
Scottmitch: consider making this package private because it is used outside the class

common/src/main/java/io/netty/util/internal/PriorityQueue.java
normanmaurer: +1

normanmaurer:
@anuraaga please also rebase your commit and update the benchmark numbers :)

microbench/src/main/java/io/netty/microbench/concurrent/ScheduledFutureTaskBenchmark.java
```diff
@@ -0,0 +1,92 @@
...
+                        return null;
+                    }
+                }, i, TimeUnit.HOURS));
+            }
+        }
+    }
+
+    @TearDown(Level.Trial)
+    public void stop() throws Exception {
+        executor.shutdownGracefully();
```
normanmaurer: I wonder if we should call `sync()` here just to ensure it is really "gone". @Scottmitch WDYT ?

microbench/src/main/java/io/netty/microbench/concurrent/ScheduledFutureTaskBenchmark.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
+
+@State(Scope.Benchmark)
+public class ScheduledFutureTaskBenchmark extends AbstractMicrobenchmark {
+
+    private static final EventLoop executor = new DefaultEventLoop();
+
+    @State(Scope.Thread)
+    public static class FuturesHolder {
```
normanmaurer: final

common/src/main/java/io/netty/util/internal/DefaultPriorityQueue.java
anuraaga: Went with option 1 since it looks like this is only called when shutting down an event loop, so indeed the queue should always be made GC-able.

microbench/src/main/java/io/netty/microbench/concurrent/ScheduledFutureTaskBenchmark.java
anuraaga: Unfortunately JMH doesn't allow `@State` classes to be `final`

microbench/src/main/java/io/netty/microbench/concurrent/ScheduledFutureTaskBenchmark.java
anuraaga: Added sync for good measure

microbench/src/main/java/io/netty/microbench/concurrent/ScheduledFutureTaskBenchmark.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
+
+@State(Scope.Benchmark)
+public class ScheduledFutureTaskBenchmark extends AbstractMicrobenchmark {
+
+    static final EventLoop executor = new DefaultEventLoop();
+
+    @State(Scope.Thread)
+    public static class FuturesHolder {
```
normanmaurer: make this class final ?

anuraaga: Had replied to the previous comment too, but unfortunately JMH doesn't allow @State classes to be final

normanmaurer: @anuraaga ah missed this... Thanks, TIL ;)

normanmaurer:
@Scottmitch PTAL again and pull in once satisfied. Looks like a good change to me...

@anuraaga thanks again!

normanmaurer:
Adding @carl-mastrangelo as reviewer as well

common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java
```diff
@@ -41,14 +51,15 @@ protected static long nanoTime() {
         return ScheduledFutureTask.nanoTime();
     }
 
-    Queue<ScheduledFutureTask<?>> scheduledTaskQueue() {
+    PriorityQueue<ScheduledFutureTask<?>> scheduledTaskQueue() {
         if (scheduledTaskQueue == null) {
-            scheduledTaskQueue = new PriorityQueue<ScheduledFutureTask<?>>();
+            scheduledTaskQueue = new DefaultPriorityQueue<ScheduledFutureTask<?>>(
+                    SCHEDULED_FUTURE_TASK_COMPARATOR, 11);
```
carl-mastrangelo: optional: add an inline comment for what 11 means?

common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java
```diff
@@ -205,7 +216,7 @@ public void run() {
 
     final void removeScheduled(final ScheduledFutureTask<?> task) {
         if (inEventLoop()) {
-            scheduledTaskQueue().remove(task);
+            scheduledTaskQueue().removeTyped(task);
```
carl-mastrangelo: one alternative (which the JDK ScheduledExecutorService does) is that it marks the task as failed but leaves the task in place.  This results in an O(1) time op, but allows the queue to possibly get large.  

Since your change is already trading memory for cpu, it may be worth doing.  The downside is that an uncleared queue will eat memory and possibly get promoted.

common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java
anuraaga: Added comment

common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java
anuraaga: I think the fixed, small overhead of the queue index is better than having weak bounds on the queue size by not removing futures. For example, with a server serving 1K qps, a user setting a request timeout of 10s would have a queue with 10K items at all times, while reconfiguring to 20s would have a queue with 20K items at all times - this doesn't seem like good, predictable behavior.

Also, leaving it in would not really be `O(1)` but just delay the `O(log n)` to the future when the task gets unscheduled, as removal from the front of a queue is always `log n` bubbles. Perhaps this could be worked around with some batch 'remove from front' algorithm, but doesn't seem worth the complexity IMO.

Scottmitch: I would prefer to keep the current behavior for this PR, and consider alternative behavior in a followup PR.

Scottmitch:
4.0 Doesn't have the PriorityQueue and other changes ... I will not back port.

normanmaurer:
+1 .... 4.1 only sounds good

PR #7387, opened 2017-11-08
Fix check logic in Http2MultiplexCodecBuilder

sunng87:
Motivation:

There is an logic issue when checking if ChannelHandler is sharable.

Modification:

Corrected `||` to `&&`. Added test cases to reproduce this issue.


Scottmitch:
4.1 (73e8122fc1)

PR #7388, opened 2017-11-08
Correctly convert empty HttpContent to ByteBuf

normanmaurer:
Motivation:

93130b172a61815354267f0f3d8f5af52de39754 introduced a regression where we not "converted" an empty HttpContent to ByteBuf and just passed it on in the pipeline. This can lead to the situation that other handlers in the pipeline will see HttpContent instances which is not expected.

Modifications:

- Correctly convert HttpContent to ByteBuf when empty
- Add unit test.

Result:

Handlers in the pipeline will see the expected message type.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java
@@ -17,18 +17,17 @@ package io.netty.handler.codec.http;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufUtil;
+import io.netty.buffer.Unpooled;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.FileRegion;
 import io.netty.handler.codec.MessageToMessageEncoder;
 import io.netty.util.CharsetUtil;
-import io.netty.util.ReferenceCountUtil;
 import io.netty.util.internal.StringUtil;
 
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map.Entry;
 
-import static io.netty.buffer.Unpooled.EMPTY_BUFFER;
 import static io.netty.buffer.Unpooled.directBuffer;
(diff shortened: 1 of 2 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java
```diff
@@ -151,7 +152,7 @@ protected void encode(ChannelHandlerContext ctx, Object msg, List<Object> out) t
                     } else {
                         // Need to produce some output otherwise an
                         // IllegalStateException will be thrown
-                        out.add(EMPTY_BUFFER);
+                        out.add(encodeAndRetain(msg));
```
Scottmitch: should we ensure that we only write a 0 length buffer in this case?

codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java
```diff
@@ -150,8 +150,11 @@ protected void encode(ChannelHandlerContext ctx, Object msg, List<Object> out) t
                         out.add(buf);
                     } else {
                         // Need to produce some output otherwise an
-                        // IllegalStateException will be thrown
-                        out.add(EMPTY_BUFFER);
+                        // IllegalStateException will be thrown as we did not write anything
+                        // Its ok to just write an EMPTY_BUFFER as if there are reference count issues these will be
+                        // propagated as the caller of the encode(...) method will release the original
+                        // buffer.
```
Scottmitch: also is this true: `writing an empty buffer will not actually write anything on the wire, so if there is a user error with msg it will not be visible externally`

normanmaurer: done

normanmaurer:
Cherry-picked into 4.1 (3554646a600d8b20570cad498d7a5c5ce7a966de) and 4.0 (85faf9e0252969d322aedf2b31321eb09ffb8e90)

PR #7390, opened 2017-11-08
Move EmptyHttpHeaders.INSTANCE initialization to inner class.

mcnulty:
Move EmptyHttpHeaders.INSTANCE initialization to inner class.

    Motivation:

    If the HttpUtil class is initialized before HttpHeaders or
    EmptyHttpHeaders, EmptyHttpHeaders.INSTANCE will be null. This
    can lead to NPEs in code that relies on this field being
    non-null. One example is the
    LastHttpContent.EMPTY_LAST_CONTENT.trailingHeaders method.

    Modifications:

    Move HttpUtil.EMPTY_HEADERS to a private static final inner class
    of EmptyHttpHeaders called InstanceInitializer.

    Result:

    Any initialization order of HttpUtil, EmptyHttpHeaders or
    HttpHeaders will result in EmptyHttpHeaders.INSTANCE being initialized
    correctly.

The 1st and 3rd commits in this PR contain some tests that when run in isolation (perhaps via an IDE), trigger a certain initialization order of the classes. As a result, it might not makes sense to merge the 1st and 3rd commits, as they won't really be testing the desired conditions when run with other tests.

### DIFF
--- a/codec-http/src/test/java/io/netty/handler/codec/http/HttpRequestEncoderTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/HttpRequestEncoderTest.java
@@ -18,6 +18,8 @@ package io.netty.handler.codec.http;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.channel.embedded.EmbeddedChannel;
+import io.netty.handler.codec.DecoderResult;
+import io.netty.util.CharsetUtil;
 import io.netty.util.IllegalReferenceCountException;
 import org.junit.Test;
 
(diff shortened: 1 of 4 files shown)


codec-http/src/test/java/io/netty/handler/codec/http/HttpRequestEncoderTest.java
```diff
@@ -154,4 +157,81 @@ public void testEmptydBufferShouldPassThrough() throws Exception {
...
+
+    @Test
+    public void testForChunkedRequestNpe() throws Exception {
+        EmbeddedChannel channel = new EmbeddedChannel(new HttpRequestEncoder());
+        channel.writeOutbound(new CustomHttpRequest());
+        channel.writeOutbound(new DefaultHttpContent(Unpooled.copiedBuffer("test", CharsetUtil.US_ASCII)));
+        channel.writeOutbound(LastHttpContent.EMPTY_LAST_CONTENT);
+    }
+
+    private static class CustomHttpRequest implements HttpRequest {
```
Scottmitch: why is this object necessary? can you just use `DefaultHttpRequest` or is class loading impacted? Can you add a comment if it is necessary?

codec-http/src/test/java/io/netty/handler/codec/http/HttpRequestEncoderTest.java
```diff
@@ -154,4 +157,81 @@ public void testEmptydBufferShouldPassThrough() throws Exception {
         channel.finishAndReleaseAll();
         assertEquals(0, buffer.refCnt());
     }
+
+    @Test
+    public void testForChunkedRequestNpe() throws Exception {
+        EmbeddedChannel channel = new EmbeddedChannel(new HttpRequestEncoder());
+        channel.writeOutbound(new CustomHttpRequest());
+        channel.writeOutbound(new DefaultHttpContent(Unpooled.copiedBuffer("test", CharsetUtil.US_ASCII)));
+        channel.writeOutbound(LastHttpContent.EMPTY_LAST_CONTENT);
```
Scottmitch: please call `channel.finishAndReleaseAll()`, and assert the return value.

codec-http/src/test/java/io/netty/handler/codec/http/HttpRequestEncoderTest.java
```diff
@@ -111,7 +114,7 @@ public void testEmptyAbsPath() throws Exception {
         for (ByteBuf buffer : getBuffers()) {
             HttpRequestEncoder encoder = new HttpRequestEncoder();
             encoder.encodeInitialLine(buffer, new DefaultHttpRequest(HttpVersion.HTTP_1_1,
-                    HttpMethod.GET, ""));
+                                                                     HttpMethod.GET, ""));
```
Scottmitch: please revert these formatting changes.

mcnulty:
@Scottmitch Thanks for taking a look at the changes. 

I was not actually expecting that the tests in the first and third commits would be merged. The tests won't fail when run with together with the other tests because the static fields they are testing will likely have already been initialized. I only included them to help illustrate the underlying issue. Sorry for the confusion.

Would it be simpler if I remove those commits from the PR and instead include the source for the tests as comments in this PR?

codec-http/src/test/java/io/netty/handler/codec/http/HttpRequestEncoderTest.java
mcnulty: Yes. If DefaultHttpRequest is used, the HttpHeaders class will be initialized before HttpUtil and the test won't trigger the issue.

codec-http/src/test/java/io/netty/handler/codec/http/HttpRequestEncoderTest.java
normanmaurer: also please assert the return values of `writeOutbound`

codec-http/src/test/java/io/netty/handler/codec/http/HttpRequestEncoderTest.java
normanmaurer: final

codec-http/src/test/java/io/netty/handler/codec/http/HttpRequestEncoderTest.java
```diff
@@ -123,7 +126,7 @@ public void testQueryStringPath() throws Exception {
         for (ByteBuf buffer : getBuffers()) {
             HttpRequestEncoder encoder = new HttpRequestEncoder();
             encoder.encodeInitialLine(buffer, new DefaultHttpRequest(HttpVersion.HTTP_1_1,
-                    HttpMethod.GET, "/?url=http://example.com"));
+                                                                     HttpMethod.GET, "/?url=http://example.com"));
```
normanmaurer: please revert formatting changes

codec-http/src/test/java/io/netty/handler/codec/http/HttpRequestEncoderTest.java
Scottmitch: @mcnulty - Please add a java doc or comment including this rational.

Scottmitch:
@mcnulty - Including the tests is OK IMHO ... even if its possible they won't trigger the issue due to class loader already being "primed". As long as they will catch the issue when run in isolation. @normanmaurer - WDYT?

normanmaurer:
@mcnulty agreee with @Scottmitch here. Just keep the test. After you addressed the comments I will pull in

mcnulty:
Sounds good, I'll address the comments and update the PR.

normanmaurer:
@mcnulty let me know once done...

mcnulty:
@normanmaurer the PR has been updated. I think I addressed all the feedback. Let me know if you would like to see anything else adjusted.

mcnulty:
@normanmaurer done.

normanmaurer:
Cherry-picked into 4.1 (48b4502d1d19208fb46d1ecbe4fb1f3b42248df2)

normanmaurer:
@mcnulty thanks!

Issue #7393, opened 2017-11-09
pooled ByteBuf copy will create unpooled ByteBuf


las1991:
-Dio.netty.allocator.type : pooled

When i call CompositeByteBuf.copy , but it create a Unpooled ByteBuf.
and then my server frequently GC.
I think this method should be written like this :

`public ByteBuf copy(int index, int length) {`
`checkIndex(index, length);`
`ByteBuf dst = ByteBufAllocator.DEFAULT.buffer(length);`
`if (length != 0) {`
`copyTo(index, length, toComponentIndex(index), dst);`
`}`
`return dst;`
`    }`

This is just one of my Suggestions, if right, hope to fix it later. If not, please correct it.

ninja-:
I think changing copy() to create pooled buffers would be api breaking and would cause memory leaks in user apps...(no need to call .release() right now)
Personally I avoid using copy() for the reason you mentioned and replace it with:
1. create new pooled buffer with the same size
2. newBuf.readBytes(buf)

ninja-:
I would also recommend avoiding CompositeByteBuf for most use cases, especially for high performance

normanmaurer:
@ninja- I disagree with your point about API breakage. You are supposed to call release() if you not do its a bug. That said I will need to have a look whats going on here.

normanmaurer:
@las1991 please check https://github.com/netty/netty/pull/7395 ... This should fix it

las1991:
@normanmaurer thank you

PR #7395, opened 2017-11-09
CompositeBytebuf.copy() and copy(...) should respect the allocator

normanmaurer:
Motivation:

When calling CompositeBytebuf.copy() and copy(...) we currently use Unpooled to allocate the buffer. This is not really correct and may produce more GC then needed. We should use the allocator that was used when creating the CompositeByteBuf to allocate the new buffer which may be for example the PooledByteBufAllocator.

Modifications:

- Use alloc() to allocate the new buffer.
- Add tests
- Fix tests that depend on the copy to be backed by an byte-array without checking hasArray() first.

Result:

Fixes [#7393].

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
@@ -1333,7 +1333,7 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
     @Override
     public ByteBuf copy(int index, int length) {
         checkIndex(index, length);
-        ByteBuf dst = Unpooled.buffer(length);
+        ByteBuf dst = allocBuffer(length);
         if (length != 0) {
             copyTo(index, length, toComponentIndex(index), dst);
         }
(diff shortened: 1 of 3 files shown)


buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
```diff
@@ -1333,7 +1333,7 @@ public int setBytes(int index, FileChannel in, long position, int length) throws
     @Override
     public ByteBuf copy(int index, int length) {
         checkIndex(index, length);
-        ByteBuf dst = Unpooled.buffer(length);
+        ByteBuf dst = alloc().buffer(length);
```
fenik17: Maybe `allocBuffer(length)`?

normanmaurer: @fenik17 good catch... let me do this and add one more assert in the tests 👍

normanmaurer:
Cherry-picked into 4.1 (cc069722a2d3ecdd6ec1dc9ae8245eca3e8d734f) and 4.0 (4c525d1e514ac6373c2cd628401e363703d9e178)

PR #7397, opened 2017-11-10
Correctly handle 205 Reset Content response with transfer-encoding

violetagg:
Motivation:

According to RFC 7231 the server may choose to:
```
indicate a zero-length payload for the response by including a
Transfer-Encoding header field with a value of chunked and a message
body consisting of a single chunk of zero-length
```
https://tools.ietf.org/html/rfc7231#page-53

In such cases the exception below appears during decoding phase:
```
java.lang.IllegalArgumentException: invalid version format: 0
	at io.netty.handler.codec.http.HttpVersion.<init>(HttpVersion.java:121)
	at io.netty.handler.codec.http.HttpVersion.valueOf(HttpVersion.java:76)
	at io.netty.handler.codec.http.HttpResponseDecoder.createMessage(HttpResponseDecoder.java:118)
	at io.netty.handler.codec.http.HttpObjectDecoder.decode(HttpObjectDecoder.java:219)
```

Modifications:

HttpObjectDecoder.isContentAlwaysEmpty specifies content NOT empty
when 205 Reset Content response

Result:

There is no `IllegalArgumentException: invalid version format: 0`
when handling 205 Reset Content response with transfer-encoding

### DIFF
--- a/codec-http/src/test/java/io/netty/handler/codec/http/HttpResponseDecoderTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/HttpResponseDecoderTest.java
@@ -369,6 +369,28 @@ public class HttpResponseDecoderTest {
     }
 
     @Test
+    public void testResetContentResponseWithTransferEncoding() {
+        EmbeddedChannel ch = new EmbeddedChannel(new HttpResponseDecoder());
+        assertTrue(ch.writeInbound(Unpooled.copiedBuffer(
+                "HTTP/1.1 205 Reset Content\r\n" +
+                "Transfer-Encoding: chunked\r\n" +
+                "\r\n" +
+                "0\r\n" +
+                "\r\n",
+                CharsetUtil.US_ASCII)));
+
+        HttpResponse res = ch.readInbound();
+        assertThat(res.protocolVersion(), sameInstance(HttpVersion.HTTP_1_1));
+        assertThat(res.status(), is(HttpResponseStatus.RESET_CONTENT));
+
+        LastHttpContent lastContent = ch.readInbound();
+        assertThat(lastContent.content().isReadable(), is(false));
+        lastContent.release();
+
+        assertThat(ch.finish(), is(false));
+    }
+
+    @Test
     public void testLastResponseWithTrailingHeader() {
         EmbeddedChannel ch = new EmbeddedChannel(new HttpResponseDecoder());
         ch.writeInbound(Unpooled.copiedBuffer(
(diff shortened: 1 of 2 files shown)


codec-http/src/test/java/io/netty/handler/codec/http/HttpResponseDecoderTest.java
```diff
@@ -369,6 +369,29 @@ public void testLastResponseWithHeaderRemoveTrailingSpaces() {
...
+        HttpResponse res = ch.readInbound();
+        assertThat(res.protocolVersion(), sameInstance(HttpVersion.HTTP_1_1));
+        assertThat(res.status(), is(HttpResponseStatus.RESET_CONTENT));
+
+        LastHttpContent lastContent = ch.readInbound();
+        assertThat(lastContent.content().isReadable(), is(false));
+        lastContent.release();
+
+        assertThat(ch.finish(), is(false));
+        assertThat(ch.readInbound(), is(nullValue()));
```
normanmaurer: you can remove this line as if `ch.finish()` returns false it means there is nothing in the inbound and outbound queue.

codec-http/src/test/java/io/netty/handler/codec/http/HttpResponseDecoderTest.java
```diff
@@ -369,6 +369,29 @@ public void testLastResponseWithHeaderRemoveTrailingSpaces() {
     }
 
     @Test
+    public void testResetContentResponseWithTransferEncoding() {
+        EmbeddedChannel ch = new EmbeddedChannel(new HttpResponseDecoder());
+        ch.writeInbound(Unpooled.copiedBuffer(
```
normanmaurer: please assert the return value

violetagg: done

codec-http/src/test/java/io/netty/handler/codec/http/HttpResponseDecoderTest.java
violetagg: done

normanmaurer:
@Lukasa do you mind also checking this one ?

Lukasa:
This looks fine. 205 is kinda strange: it’s defined to always have a zero-length payload, not to not have a payload. Put another way, it should always have a body with an effective content length of zero.

normanmaurer:
Cherry-picked into 4.1 (72a216625f16617579f0c0102a7d6624865148d6) and 4.0 (24d12d35b1)

normanmaurer:
@violetagg thanks!

PR #7400, opened 2017-11-10
H2C upgrades should be ineligible for flow control

### LINKED ISSUE #7280: throwing exception on h2c upgrade
I occasionally see a failure when I'm doing an h2c upgrade, although not very frequently.

The exception looks like:

```
WAR [20171004-22:30:36.811] twitter: Unhandled exception in connection with /10.70.113.107:40092, shutting down connection
WAR [20171004-22:30:36.811] twitter: io.netty.handler.codec.http2.Http2Exception: Attempting to return too many bytes for stream 0
WAR [20171004-22:30:36.811] twitter:     at io.netty.handler.codec.http2.Http2Exception.connectionError(Http2Exception.java:85)
WAR [20171004-22:30:36.811] twitter:     at io.netty.handler.codec.http2.Http2Exception.streamError(Http2Exception.java:128)
WAR [20171004-22:30:36.811] twitter:     at io.netty.handler.codec.http2.DefaultHttp2LocalFlowController$DefaultState.returnProcessedBytes(DefaultHttp2LocalFlowController.java\
:432)
WAR [20171004-22:30:36.811] twitter:     at io.netty.handler.codec.http2.DefaultHttp2LocalFlowController$DefaultState.consumeBytes(DefaultHttp2LocalFlowController.java:441)
WAR [20171004-22:30:36.811] twitter:     at io.netty.handler.codec.http2.DefaultHttp2LocalFlowController.consumeBytes(DefaultHttp2LocalFlowController.java:190)
WAR [20171004-22:30:36.811] twitter:     at io.netty.handler.codec.http2.Http2FrameCodec.consumeBytes(Http2FrameCodec.java:315)
WAR [20171004-22:30:36.811] twitter:     at io.netty.handler.codec.http2.Http2MultiplexCodec.onBytesConsumed(Http2MultiplexCodec.java:384)
WAR [20171004-22:30:36.811] twitter:     at io.netty.handler.codec.http2.Http2MultiplexCodec$DefaultHttp2StreamChannel$Http2ChannelUnsafe.doRead0(Http2MultiplexCodec.java:947)
WAR [20171004-22:30:36.811] twitter:     at io.netty.handler.codec.http2.Http2MultiplexCodec$DefaultHttp2StreamChannel.fireChildRead(Http2MultiplexCodec.java:734)
WAR [20171004-22:30:36.811] twitter:     at io.netty.handler.codec.http2.Http2MultiplexCodec.onHttp2StreamFrame(Http2MultiplexCodec.java:286)
WAR [20171004-22:30:36.811] twitter:     at io.netty.handler.codec.http2.Http2MultiplexCodec.onHttp2Frame(Http2MultiplexCodec.java:221)
WAR [20171004-22:30:36.811] twitter:     at io.netty.handler.codec.http2.Http2FrameCodec$FrameListener.onDataRead(Http2FrameCodec.java:527)
WAR [20171004-22:30:36.811] twitter:     at io.netty.handler.codec.http2.InboundHttpToHttp2Adapter.handle(InboundHttpToHttp2Adapter.java:67)
WAR [20171004-22:30:36.811] twitter:     at io.netty.handler.codec.http2.Http2FrameCodec.userEventTriggered(Http2FrameCodec.java:250)
WAR [20171004-22:30:36.811] twitter:     at io.netty.channel.AbstractChannelHandlerContext.invokeUserEventTriggered(AbstractChannelHandlerContext.java:329)
WAR [20171004-22:30:36.811] twitter:     at io.netty.channel.AbstractChannelHandlerContext.invokeUserEventTriggered(AbstractChannelHandlerContext.java:315)
WAR [20171004-22:30:36.811] twitter:     at io.netty.channel.AbstractChannelHandlerContext.fireUserEventTriggered(AbstractChannelHandlerContext.java:307)
WAR [20171004-22:30:36.811] twitter:     at io.netty.handler.codec.http.HttpServerUpgradeHandler$1.operationComplete(HttpServerUpgradeHandler.java:329)
WAR [20171004-22:30:36.811] twitter:     at io.netty.handler.codec.http.HttpServerUpgradeHandler$1.operationComplete(HttpServerUpgradeHandler.java:318)
WAR [20171004-22:30:36.811] twitter:     at io.netty.util.concurrent.DefaultPromise.notifyListener0(DefaultPromise.java:507)
WAR [20171004-22:30:36.811] twitter:     at io.netty.util.concurrent.DefaultPromise.notifyListenersNow(DefaultPromise.java:481)
WAR [20171004-22:30:36.811] twitter:     at io.netty.util.concurrent.DefaultPromise.notifyListeners(DefaultPromise.java:420)
WAR [20171004-22:30:36.811] twitter:     at io.netty.util.concurrent.DefaultPromise.addListener(DefaultPromise.java:163)
WAR [20171004-22:30:36.811] twitter:     at io.netty.channel.DefaultChannelPromise.addListener(DefaultChannelPromise.java:93)
WAR [20171004-22:30:36.811] twitter:     at io.netty.channel.DefaultChannelPromise.addListener(DefaultChannelPromise.java:28)
WAR [20171004-22:30:36.811] twitter:     at io.netty.handler.codec.http.HttpServerUpgradeHandler.upgrade(HttpServerUpgradeHandler.java:318)
WAR [20171004-22:30:36.811] twitter:     at io.netty.handler.codec.http.HttpServerUpgradeHandler.decode(HttpServerUpgradeHandler.java:239)
WAR [20171004-22:30:36.811] twitter:     at io.netty.handler.codec.http.HttpServerUpgradeHandler.decode(HttpServerUpgradeHandler.java:40)
WAR [20171004-22:30:36.811] twitter:     at io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:88)
WAR [20171004-22:30:36.811] twitter:     at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
```

And it comes from [here](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2LocalFlowController.java?utf8=%E2%9C%93#L432-L433):

```
        private void returnProcessedBytes(int delta) throws Http2Exception {
            if (processedWindow - delta < window) {
                throw streamError(stream.id(), INTERNAL_ERROR,
                        "Attempting to return too many bytes for stream %d", stream.id());
            }
            processedWindow -= delta;
        }
```

Any idea what's going on here?  Is the issue that the client isn't respecting flow control?

mosesn:
Motivation:

When the h2c upgrade request is too big, the Http2FrameCodec complains
it's too big for flow control reasons, even though it's ineligible for
flow control.

Modifications:

Specially mark upgrade streams and make Http2FrameCodec know not to try
to flow control on those streams.

Result:

Servers won't barf when they receive an upgrade request with a fat
payload.

[Fixes #7280]

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2LocalFlowController.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2LocalFlowController.java
@@ -53,7 +53,6 @@ public class DefaultHttp2LocalFlowController implements Http2LocalFlowController
     private ChannelHandlerContext ctx;
     private float windowUpdateRatio;
     private int initialWindowSize = DEFAULT_WINDOW_SIZE;
-    private boolean cleartextUpgradeConnection;
 
     public DefaultHttp2LocalFlowController(Http2Connection connection) {
         this(connection, DEFAULT_WINDOW_UPDATE_RATIO, false);
@@ -180,9 +179,6 @@ public class DefaultHttp2LocalFlowController implements Http2LocalFlowController
         if (numBytes == 0) {
             return false;
         }
-        if (cleartextUpgradeConnection && stream.id() == Http2CodecUtil.HTTP_UPGRADE_STREAM_ID) {
-            return false;
-        }
 
         // Streams automatically consume all remaining bytes when they are closed, so just ignore
         // if already closed.
@@ -281,11 +277,6 @@ public class DefaultHttp2LocalFlowController implements Http2LocalFlowController
             // Immediately consume the bytes for the connection window.
             connectionState.consumeBytes(dataLength);
         }
-    }
-
-    @Override
-    public void designateCleartextUpgradeConnection() {
-        cleartextUpgradeConnection = true;
     }
 
     private FlowState connectionState() {
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2LocalFlowController.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2LocalFlowController.java
@@ -53,6 +53,7 @@ public class DefaultHttp2LocalFlowController implements Http2LocalFlowController
     private ChannelHandlerContext ctx;
     private float windowUpdateRatio;
     private int initialWindowSize = DEFAULT_WINDOW_SIZE;
+    private boolean cleartextUpgradeConnection;
 
     public DefaultHttp2LocalFlowController(Http2Connection connection) {
         this(connection, DEFAULT_WINDOW_UPDATE_RATIO, false);
@@ -179,6 +180,9 @@ public class DefaultHttp2LocalFlowController implements Http2LocalFlowController
         if (numBytes == 0) {
             return false;
         }
+        if (cleartextUpgradeConnection && stream.id() == Http2CodecUtil.HTTP_UPGRADE_STREAM_ID) {
+            return false;
+        }
 
         // Streams automatically consume all remaining bytes when they are closed, so just ignore
         // if already closed.
@@ -277,6 +281,11 @@ public class DefaultHttp2LocalFlowController implements Http2LocalFlowController
             // Immediately consume the bytes for the connection window.
             connectionState.consumeBytes(dataLength);
         }
+    }
+
+    @Override
+    public void designateCleartextUpgradeConnection() {
+        cleartextUpgradeConnection = true;
     }
 
     private FlowState connectionState() {
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
@@ -190,8 +190,7 @@ public class Http2ConnectionHandler extends ByteToMessageDecoder implements Http
         encoder.remoteSettings(settings);
 
         // Create a stream in the half-closed state.
-        decoder.flowController().designateCleartextUpgradeConnection();
-        Http2Stream stream = connection().remote().createStream(HTTP_UPGRADE_STREAM_ID, true);
+        connection().remote().createStream(HTTP_UPGRADE_STREAM_ID, true);
     }
 
     @Override
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
@@ -190,7 +190,8 @@ public class Http2ConnectionHandler extends ByteToMessageDecoder implements Http
         encoder.remoteSettings(settings);
 
         // Create a stream in the half-closed state.
-        connection().remote().createStream(HTTP_UPGRADE_STREAM_ID, true);
+        decoder.flowController().designateCleartextUpgradeConnection();
+        Http2Stream stream = connection().remote().createStream(HTTP_UPGRADE_STREAM_ID, true);
     }
 
     @Override
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
@@ -145,7 +145,6 @@ public class Http2FrameCodec extends Http2ConnectionHandler {
     private static final InternalLogger LOG = InternalLoggerFactory.getInstance(Http2FrameCodec.class);
 
     private final PropertyKey streamKey;
-    private final PropertyKey upgradeKey;
 
     private final Integer initialFlowControlWindowSize;
 
@@ -162,7 +161,6 @@ public class Http2FrameCodec extends Http2ConnectionHandler {
         connection().addListener(new ConnectionListener());
         connection().remote().flowController().listener(new Http2RemoteFlowControllerListener());
         streamKey = connection().newKey();
-        upgradeKey = connection().newKey();
         initialFlowControlWindowSize = initialSettings.initialWindowSize();
     }
 
@@ -249,7 +247,6 @@ public class Http2FrameCodec extends Http2ConnectionHandler {
                 }
                 upgrade.upgradeRequest().headers().setInt(
                         HttpConversionUtil.ExtensionHeaderNames.STREAM_ID.text(), HTTP_UPGRADE_STREAM_ID);
-                stream.setProperty(upgradeKey, true);
                 InboundHttpToHttp2Adapter.handle(
                         ctx, connection(), decoder().frameListener(), upgrade.upgradeRequest().retain());
             } finally {
@@ -315,14 +312,6 @@ public class Http2FrameCodec extends Http2ConnectionHandler {
 
     final boolean consumeBytes(int streamId, int bytes) throws Http2Exception {
         Http2Stream stream = connection().stream(streamId);
-        // upgraded requests are ineligible for stream control
-        if (streamId == Http2CodecUtil.HTTP_UPGRADE_STREAM_ID) {
-            Boolean upgraded = stream.getProperty(upgradeKey);
-            if (Boolean.TRUE.equals(upgraded)) {
-                return true;
-            }
-        }
-
         return connection().local().flowController().consumeBytes(stream, bytes);
     }
 
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
@@ -145,6 +145,7 @@ public class Http2FrameCodec extends Http2ConnectionHandler {
     private static final InternalLogger LOG = InternalLoggerFactory.getInstance(Http2FrameCodec.class);
 
     private final PropertyKey streamKey;
+    private final PropertyKey upgradeKey;
 
     private final Integer initialFlowControlWindowSize;
 
@@ -161,6 +162,7 @@ public class Http2FrameCodec extends Http2ConnectionHandler {
         connection().addListener(new ConnectionListener());
         connection().remote().flowController().listener(new Http2RemoteFlowControllerListener());
         streamKey = connection().newKey();
+        upgradeKey = connection().newKey();
         initialFlowControlWindowSize = initialSettings.initialWindowSize();
     }
 
@@ -247,6 +249,7 @@ public class Http2FrameCodec extends Http2ConnectionHandler {
                 }
                 upgrade.upgradeRequest().headers().setInt(
                         HttpConversionUtil.ExtensionHeaderNames.STREAM_ID.text(), HTTP_UPGRADE_STREAM_ID);
+                stream.setProperty(upgradeKey, true);
                 InboundHttpToHttp2Adapter.handle(
                         ctx, connection(), decoder().frameListener(), upgrade.upgradeRequest().retain());
             } finally {
@@ -312,6 +315,14 @@ public class Http2FrameCodec extends Http2ConnectionHandler {
 
     final boolean consumeBytes(int streamId, int bytes) throws Http2Exception {
         Http2Stream stream = connection().stream(streamId);
+        // upgraded requests are ineligible for stream control
+        if (streamId == Http2CodecUtil.HTTP_UPGRADE_STREAM_ID) {
+            Boolean upgraded = stream.getProperty(upgradeKey);
+            if (Boolean.TRUE.equals(upgraded)) {
+                return true;
+            }
+        }
+
         return connection().local().flowController().consumeBytes(stream, bytes);
     }
 
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
@@ -318,7 +318,7 @@ public class Http2FrameCodec extends Http2ConnectionHandler {
         // upgraded requests are ineligible for stream control
         if (streamId == Http2CodecUtil.HTTP_UPGRADE_STREAM_ID) {
             Boolean upgraded = stream.getProperty(upgradeKey);
-            if (upgraded != null && upgraded) {
+            if (Boolean.TRUE.equals(upgraded)) {
                 return true;
             }
         }
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
@@ -145,6 +145,7 @@ public class Http2FrameCodec extends Http2ConnectionHandler {
     private static final InternalLogger LOG = InternalLoggerFactory.getInstance(Http2FrameCodec.class);
 
     private final PropertyKey streamKey;
+    private final PropertyKey upgradeKey;
 
     private final Integer initialFlowControlWindowSize;
 
@@ -161,6 +162,7 @@ public class Http2FrameCodec extends Http2ConnectionHandler {
         connection().addListener(new ConnectionListener());
         connection().remote().flowController().listener(new Http2RemoteFlowControllerListener());
         streamKey = connection().newKey();
+        upgradeKey = connection().newKey();
         initialFlowControlWindowSize = initialSettings.initialWindowSize();
     }
 
@@ -247,6 +249,7 @@ public class Http2FrameCodec extends Http2ConnectionHandler {
                 }
                 upgrade.upgradeRequest().headers().setInt(
                         HttpConversionUtil.ExtensionHeaderNames.STREAM_ID.text(), HTTP_UPGRADE_STREAM_ID);
+                stream.setProperty(upgradeKey, true);
                 InboundHttpToHttp2Adapter.handle(
                         ctx, connection(), decoder().frameListener(), upgrade.upgradeRequest().retain());
             } finally {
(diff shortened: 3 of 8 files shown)


mosesn:
@ejona86 @normanmaurer I could use some help with this one–my test seems like it should be OK, but it doesn't seem to fail when I remove my fix.  So I don't know where the exception has disappeared off to.  Does one of you mind taking a look and helping me out with this?

normanmaurer:
@mosesn add this diff to your pr:

```
diff --git a/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2FrameCodecTest.java b/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2FrameCodecTest.java
index 8da52d076..4d29f5465 100644
--- a/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2FrameCodecTest.java
+++ b/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2FrameCodecTest.java
@@ -21,6 +21,7 @@ import io.netty.buffer.UnpooledByteBufAllocator;
 import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.embedded.EmbeddedChannel;
 import io.netty.handler.codec.UnsupportedMessageTypeException;
@@ -37,6 +38,7 @@ import io.netty.handler.codec.http2.Http2Stream.State;
 import io.netty.handler.logging.LogLevel;
 import io.netty.util.AbstractReferenceCounted;
+import io.netty.util.ReferenceCountUtil;
 import io.netty.util.ReferenceCounted;
 import io.netty.util.concurrent.DefaultPromise;
 import io.netty.util.concurrent.GlobalEventExecutor;
@@ -668,6 +670,27 @@ public class Http2FrameCodecTest {
 
     @Test
     public void upgradeWithoutFlowControlling() throws Exception {
+        channel.pipeline().addAfter(http2HandlerCtx.name(), null, new ChannelInboundHandlerAdapter() {
+            @Override
+            public void channelRead(final ChannelHandlerContext ctx, Object msg) throws Exception {
+                if (msg instanceof Http2DataFrame) {
+                    // Simulate consuming the frame and update the flow-controller.
+                    Http2DataFrame data = (Http2DataFrame) msg;
+                    ctx.writeAndFlush(new DefaultHttp2WindowUpdateFrame(data.initialFlowControlledBytes())
+                            .stream(data.stream())).addListener(new ChannelFutureListener() {
+                        public void operationComplete(ChannelFuture future) throws Exception {
+                            Throwable cause = future.cause();
+                            if (cause != null) {
+                                ctx.fireExceptionCaught(cause);
+                        }
+                    });
+                }
+                ReferenceCountUtil.release(msg);
+            }
+        });
+
         frameListener.onHeadersRead(http2HandlerCtx, Http2CodecUtil.HTTP_UPGRADE_STREAM_ID, request, 31, false);
 
         // Using reflect as the constructor is package-private and the class is final.
```

This basically simulates what an user would do normally and what the `Http2MultiplexCodec` does as well.

mosesn:
Trying it out now, thanks!

normanmaurer:
@carl-mastrangelo @Scottmitch @ejona86 any comments ?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
```diff
@@ -312,6 +315,13 @@ private void increaseInitialConnectionWindow(int deltaBytes) throws Http2Excepti
 
     final boolean consumeBytes(int streamId, int bytes) throws Http2Exception {
         Http2Stream stream = connection().stream(streamId);
+        if (streamId == Http2CodecUtil.HTTP_UPGRADE_STREAM_ID) {
```
ejona86: A comment here would be nice, since it is a workaround.

mosesn:
@ejona86 comment added, and I squashed the whole thing.  thanks!

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
```diff
@@ -312,6 +315,14 @@ private void increaseInitialConnectionWindow(int deltaBytes) throws Http2Excepti
 
     final boolean consumeBytes(int streamId, int bytes) throws Http2Exception {
         Http2Stream stream = connection().stream(streamId);
+        // upgraded requests are ineligible for stream control
+        if (streamId == Http2CodecUtil.HTTP_UPGRADE_STREAM_ID) {
+            Boolean upgraded = stream.getProperty(upgradeKey);
+            if (upgraded != null && upgraded) {
```
normanmaurer: one nit: I think you could shortcut this via `Boolean.TRUE.equals(upgrade)`.

mosesn: @normanmaurer I'd rather not depend on escape analysis to avoid the boxing allocation, but I can change it if you feel strongly.

normanmaurer: @mosesn I am not sure why any escape analysis would be needed here.. Can you explain to me why you think this would be the case ?

`upgraded` will be a `Boolean` anyway so I am not sure what you refer to.

mosesn: @normanmaurer that's an excellent point.  I'll change it.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
```diff
@@ -312,6 +315,14 @@ private void increaseInitialConnectionWindow(int deltaBytes) throws Http2Excepti
...
+                return true;
```
Scottmitch: this method isn't documented very well (internal method I guess), but the methods it calls and the method which use it say `* @return {@code true} if a frame has been written as a result of this method call.`. This doesn't write a window update, so do we really want to return `true` here, and if so can we update the javadocs?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
```diff
@@ -145,6 +145,7 @@
     private static final InternalLogger LOG = InternalLoggerFactory.getInstance(Http2FrameCodec.class);
 
     private final PropertyKey streamKey;
+    private final PropertyKey upgradeKey;
```
Scottmitch: can you add some javadocs indicating why this `PropertyKey` is necessary? (/cc @ejona86 https://github.com/netty/netty/issues/7280#issuecomment-334593812). If we are just trying to prevent the server from making contributions to flow control because the stream is created in "half closed local" on the client [1] then can we just do: 

```java
if (streamId == HTTP_UPGRADE_STREAM_ID && connection().isServer()) {
  return false;
}
...
```

Should be done in a lower layer (e.g. `DefaultHttp2LocalFlowController#consumeBytes`)?

[1] https://tools.ietf.org/html/rfc7540#section-5.1.1
> HTTP/1.1 requests that are upgraded to HTTP/2 (see Section 3.2) are
   responded to with a stream identifier of one (0x1).  After the
   upgrade completes, stream 0x1 is "half-closed (local)" to the client.
   Therefore, stream 0x1 cannot be selected as a new stream identifier
   by a client that upgrades from HTTP/1.1.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
mosesn: I don't know.  @normanmaurer what do you think?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
mosesn: @Scottmitch I think h2 clients use 0x1 as the first stream id–I think that designation is only true for h2c clients.

Scottmitch: Good call.

Any reason why we cannot move this `upgradeKey` into lower levels (e.g. when the stream is created in `Http2ConnectionHandler` and in `DefaultHttp2LocalFlowController#consumeBytes`)? It seems like this should always happen (even if not using the frame codec).

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
Scottmitch: the return status ultimately drives if we need to flush eventually or not to enable for "flush consolidation" ... so if there are no new frames generated I'm not sure why we would return `true`.

mosesn: That's compelling to me, I'm not sure I understand this well enough to draft a new comment though.  Do you mind suggesting a comment and I'll use that?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
mosesn: Sure, I can try moving it.

mosesn: @Scottmitch hmm, I actually don't know how to thread the upgradeKey through.  any ideas?

Scottmitch: Good question. This isn't exactly straightforward, and seems like we may need to make some sort of API change. Here are some options:

1. Let the `Http2LocalFlowController` create the upgrade stream. We would need some sort of API such as `Http2LocalFlowController#createUpgradeStream(...)` and the `Http2LocalFlowController` could then keep the key local.
2. Add an API on the `Http2LocalFlowController` to notify that a remote initiated upgrade stream was created.
3. Add a method on the `Http2Stream` interface to determine if the stream was a remote initiated update. The `Http2LocalFlowController` could use this method to determine if consumeBytes behavior. This state can be set at the `Http2ConnectionHandler` level, potentially with an additional method/overload for `Http2Connection#createStream`. 

Option 1 leaks some responsibilities between layers, and we would have to be careful the correct endpoint and initial stream state is used.
Option 2 increases the API footprint of `Http2LocalFlowController`, but doesn't leave much room for error and keeps concerns separated.
Option 3 will require additional APIs and also possibly additional state for every stream (e.g. a boolean) that *may* only be used for a single stream.

If there are no other options my preference would be for option (2).

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
Scottmitch: My assumption is that we want to `return false;` here. If this is the case then we shouldn't need any additional comments (no WINDOW_UPDATE is written so returning `false` doesn't change existing behavior). If this is not the case then I don't understand why we need/want to return `true` so I don't yet have a suggestion.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
normanmaurer: I also think 2) would be good.

Scottmitch: @ejona86 - any better ideas?

normanmaurer: also @carl-mastrangelo @nmittler

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
```diff
@@ -312,6 +315,14 @@ private void increaseInitialConnectionWindow(int deltaBytes) throws Http2Excepti
 
     final boolean consumeBytes(int streamId, int bytes) throws Http2Exception {
         Http2Stream stream = connection().stream(streamId);
+        // upgraded requests are ineligible for stream control
+        if (streamId == Http2CodecUtil.HTTP_UPGRADE_STREAM_ID) {
+            Boolean upgraded = stream.getProperty(upgradeKey);
+            if (Boolean.TRUE.equals(upgraded)) {
+                return true;
```
Scottmitch: continuation of https://github.com/netty/netty/pull/7400#discussion_r152084904 ...

This still doesn't make sense to me. Seems like if property is not null we should just return false, because we are not writing a window update and do not require a flush as a result of this method call.

mosesn: whoops, I just forgot to change it.

mosesn:
@Scottmitch I've changed it to do what you suggested in option 2, but I unfortunately think that I'm reaching the limits of netty competence.  This bit of the codebase is a little too complicated for me.  I've also borked my local version, so tests hang every time I run them.  If this isn't what you wanted, it might make sense to go back to the Http2FrameCodec version and then someone else can fix it up later.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -190,7 +190,8 @@ public void onHttpServerUpgrade(Http2Settings settings) throws Http2Exception {
         encoder.remoteSettings(settings);
 
         // Create a stream in the half-closed state.
-        connection().remote().createStream(HTTP_UPGRADE_STREAM_ID, true);
+        decoder.flowController().designateCleartextUpgradeConnection();
+        Http2Stream stream = connection().remote().createStream(HTTP_UPGRADE_STREAM_ID, true);
```
Scottmitch: nit: kill `Http2Stream stream` variable?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2LocalFlowController.java
```diff
@@ -179,6 +180,9 @@ public boolean consumeBytes(Http2Stream stream, int numBytes) throws Http2Except
         if (numBytes == 0) {
             return false;
         }
+        if (cleartextUpgradeConnection && stream.id() == Http2CodecUtil.HTTP_UPGRADE_STREAM_ID) {
```
Scottmitch: nit: combine conditional with previous conditional?

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2FrameCodecTest.java
```diff
@@ -666,6 +668,47 @@ public void upgradeEventNoRefCntError() throws Exception {
...
+        // Check if we could make it accessible which may fail on java9.
+        Assume.assumeTrue(ReflectionUtil.trySetAccessible(constructor) == null);
+
+        String longString = new String(new char[70000]).replace("\0", "*");
+        DefaultFullHttpRequest request =
+            new DefaultFullHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.GET, "/", bb(longString));
+
+        HttpServerUpgradeHandler.UpgradeEvent upgradeEvent = constructor.newInstance(
+            "HTTP/2", request);
+        channel.pipeline().fireUserEventTriggered(upgradeEvent);
```
Scottmitch: this test doesn't drive `Http2ConnectionHandler` through the necessary series of events to prepare for the upgrade. `onHttpServerUpgrade` is never called (and there is some stuff that must occur before that). Consider creating a unit test which uses `Http2ServerUpgradeCodec` (or simulate the events necessary) so `Http2ConnectionHandler`'s stat is properly updated.

mosesn: @Scottmitch I don't understand this, and unfortunately don't have time to tackle it.  Instead of letting this PR languish, how about I revert this commit, go back to the previous implementation, which worked but didn't cover all cases, and I'll file a ticket to fix it later.  Otherwise, I'll try to return to this when I have time (probably sometime next year).

Scottmitch: > I don't understand this

The test is not following the same steps that `Http2ServerUpgradeCodec` would, and therefore state is not properly initialized. Your code changes which invoke `designateCleartextUpgradeConnection` are not called.

> how about I revert this commit, go back to the previous implementation

The approach seems correct but if you don't want to fix the unit test reverting is fine as long as https://github.com/netty/netty/pull/7400#discussion_r152634982 is addressed.

> I've also borked my local version, so tests hang every time I run them

Can you clarify what "local version" means here, and which tests hang? Netty tests, finagle tests, other tests?

mosesn: I have a clone of the netty repository, and the netty tests hang.

mosesn:
@Scottmitch I've reverted to the old version and fixed the issue you mentioned.  PTAL.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java
ejona86: I know I'm late in replying... I don't have any better ideas. I agree that this should be done at a lower level, but it wasn't clear to me how we'd want to do that (which is what options you discussed). I would have needed to dig in much more than I have time to do. So I figured workaround here as a first-go, and we either fix it up later or just accept it and bide our time until the old API goes away.

normanmaurer: I am ok with this... @Scottmitch WDYT ?

Scottmitch:
thanks @mosesn !

Issue #7365, opened 2017-11-04
EventLoop.schedule with big delay fails


zyclonite:
this behaviour was discovered by accident, so there is no actual usecase behind it but it is definitely not possible to use Long.MAX_VALUE in milliseconds for a scheduled event

### Expected behavior
runnable should trigger after given delay
getting a proper exception if something goes wrong

### Actual behavior
working with delays smaller of around 10_000_000_000ms on all platforms i tested with (osx, linux)
failing on osx with already a delay of 15_768_000_000ms (5 years)
with the following exception
```
Nov 04, 2017 2:08:39 PM io.netty.channel.nio.NioEventLoop handleLoopException
WARNING: Unexpected exception in the selector loop.
java.io.IOException: Invalid argument
	at sun.nio.ch.KQueueArrayWrapper.kevent0(Native Method)
	at sun.nio.ch.KQueueArrayWrapper.poll(KQueueArrayWrapper.java:198)
	at sun.nio.ch.KQueueSelectorImpl.doSelect(KQueueSelectorImpl.java:117)
	at sun.nio.ch.SelectorImpl.lockAndDoSelect(SelectorImpl.java:86)
	at sun.nio.ch.SelectorImpl.select(SelectorImpl.java:97)
	at io.netty.channel.nio.NioEventLoop.select(NioEventLoop.java:746)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:391)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:858)
	at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:144)
	at java.lang.Thread.run(Thread.java:748)
```
failing on linux with around 946_080_000_000 (300 years)
which immediately triggers the runnable without exception

### Minimal yet complete reproducer code (or URL to code)
```
public class Main {
    public static void main(String[] args) {
        final ThreadFactory threadFactory = new DefaultThreadFactory("test-eventloop");
        NioEventLoopGroup eventLoopGroup = new NioEventLoopGroup(1, threadFactory);
        eventLoopGroup.setIoRatio(50);
        EventLoop el = eventLoopGroup.next();
        Runnable toRun = Main::handle;
        Future<?> future = el.schedule(toRun, 1000L * 60 * 60 * 24 * 365 * 300, TimeUnit.MILLISECONDS);
        System.out.println("timer scheduled");
    }

[...]

PR #7402, opened 2017-11-13
EventLoop.schedule with big delay fails

normanmaurer:
Motivation:

Using a very huge delay when calling schedule(...) may cause an Selector error when calling select(...) later on. We should gaurd against such a big value.

Modifications:

- Add guard against a very huge value.
- Added tests.

Result:

Fixes [#7365]

### DIFF
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
@@ -37,6 +37,7 @@ import java.util.Collection;
 import java.util.Queue;
 import java.util.concurrent.Callable;
 import java.util.concurrent.Executor;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
 import static java.lang.Math.min;
@@ -77,6 +78,9 @@ final class EpollEventLoop extends SingleThreadEventLoop {
     };
     private volatile int wakenUp;
     private volatile int ioRatio = 50;
+
+    // See http://man7.org/linux/man-pages/man2/timerfd_create.2.html.
+    static final long MAX_SCHEDULED_DAYS = TimeUnit.SECONDS.toDays(999999999);
 
     EpollEventLoop(EventLoopGroup parent, Executor executor, int maxEvents,
                    SelectStrategy strategy, RejectedExecutionHandler rejectedExecutionHandler) {
@@ -447,6 +451,14 @@ final class EpollEventLoop extends SingleThreadEventLoop {
             // release native memory
             iovArray.release();
             events.free();
+        }
+    }
+
+    @Override
+    protected void validateScheduled(long amount, TimeUnit unit) {
+        long days = unit.toDays(amount);
+        if (days > MAX_SCHEDULED_DAYS) {
+            throw new IllegalArgumentException("days: " + days + " (expected: < " + MAX_SCHEDULED_DAYS + ')');
         }
     }
 }
--- a/transport-native-kqueue/src/main/java/io/netty/channel/kqueue/KQueueEventLoop.java
+++ b/transport-native-kqueue/src/main/java/io/netty/channel/kqueue/KQueueEventLoop.java
@@ -33,6 +33,7 @@ import java.io.IOException;
 import java.util.Queue;
 import java.util.concurrent.Callable;
 import java.util.concurrent.Executor;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
 import static io.netty.channel.kqueue.KQueueEventArray.deleteGlobalRefs;
@@ -75,6 +76,8 @@ final class KQueueEventLoop extends SingleThreadEventLoop {
 
     private volatile int wakenUp;
     private volatile int ioRatio = 50;
+
+    static final long MAX_SCHEDULED_DAYS = 365 * 3;
 
     KQueueEventLoop(EventLoopGroup parent, Executor executor, int maxEvents,
                     SelectStrategy strategy, RejectedExecutionHandler rejectedExecutionHandler) {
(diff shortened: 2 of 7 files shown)


transport-native-kqueue/src/main/java/io/netty/channel/kqueue/KQueueEventLoop.java
```diff
@@ -76,6 +77,8 @@ public Integer call() throws Exception {
     private volatile int wakenUp;
     private volatile int ioRatio = 50;
 
+    static final long MAX_SCHEDULED_DAYS = 356 * 3;
```
normanmaurer: please note that I used this as it seems to work on linux / Windows and macOS.

Scottmitch: Why do we care about windows and linux here? This is in the kqueue transport.

transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
```diff
@@ -113,6 +113,8 @@ public Void run() {
         }
     }
 
+    static final long MAX_SCHEDULED_DAYS = 356 * 3;
```
Scottmitch: add a comment explaining where this number comes from.

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/KQueueEventLoop.java
Scottmitch: https://www.freebsd.org/cgi/man.cgi?query=kqueue&sektion=2#end

>  The timeout value is limited to 24	hours; longer timeouts will be
     silently reinterpreted as 24 hours.

Do we need to do anything for kqueue? IIRC we verified that scheduled tasks are actually expired even if we are woken up as a result of timeout.

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
```diff
@@ -78,6 +79,9 @@ public Integer call() throws Exception {
     private volatile int wakenUp;
     private volatile int ioRatio = 50;
 
+    // See http://man7.org/linux/man-pages/man2/timerfd_create.2.html.
```
Scottmitch: Is this limit defined in the docs, or is this just based upon the size of the data types? If its data type based this may not be correct and depends upon the type widths. For example `long` may be unsigned 32, but IIRC it is only guaranteed to be smaller than `int`. `time_t` maybe 32 or 64 byte signed integer. If this is based upon datatypes we should get this information via JNI ... hopefully in a limits file somewhere?

make this a java doc comment?

normanmaurer: its defined in the manpage to be the max you can set.

transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
normanmaurer: will do... It came from testing ;)

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
Scottmitch: > EINVAL new_value is not properly initialized (one of the tv_nsec
              falls outside the range zero to 999,999,999).

Gotcha ... consider adding to javadocs.

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
```diff
@@ -78,6 +79,9 @@ public Integer call() throws Exception {
...
+    static final long MAX_SCHEDULED_DAYS = TimeUnit.SECONDS.toNanos(999999999) + TimeUnit.NANOSECONDS.toSeconds(999999999);
```
carl-mastrangelo: `TimeUnit.NANOSECONDS.toSeconds(999999999)`

That evaluates to zero.  Is this calculation correct?

normanmaurer: we can just remove the `TimeUnit.NANOSECONDS.toSeconds(999999999)` here and keep it simple. Will do

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/KQueueEventLoop.java
normanmaurer: @Scottmitch sorry this was confusing. I tested the value with NIO on linux, windows and macOs ;)

normanmaurer: @Scottmitch oh my ... This is a nice "bug". I wonder if this is still true and why they not just fixed it... any idea ?

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
carl-mastrangelo: So if the variable is named `MAX_SCHEDULED_DAYS`   why are there no "day" like constants in the arithmetic?  It's kind of odd to put a "seconds" value into a "days" variable.

normanmaurer: carl because I am stupid... I forgot to push the latest change :(

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/KQueueEventLoop.java
carl-mastrangelo: s/356/365/ ?

Scottmitch:
@normanmaurer - kicking off another build now ... failure looked related to (hopefully transient) CI issue.

normanmaurer:
I will pick this up again... sorry totally forgot about it

PR #7403, opened 2017-11-13
Don't try to use UnixResolverDnsServerAddressStreamProvider when on W…

normanmaurer:
…indows.

Motivation:

We should not try to use UnixResolverDnsServerAddressStreamProvider when on Windows as it will log some error that will produce noise and may confuse users.

Modifications:

Just use DefaultDnsServerAddressStreamProvider if windows is used.

Result:

Less noise in the logs. This was reported in vert.x: https://github.com/eclipse/vert.x/issues/2204

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProviders.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProviders.java
@@ -15,6 +15,7 @@
  */
 package io.netty.resolver.dns;
 
+import io.netty.util.internal.PlatformDependent;
 import io.netty.util.internal.UnstableApi;
 
 /**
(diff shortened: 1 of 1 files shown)


resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProviders.java
```diff
@@ -25,7 +26,10 @@
     // TODO(scott): how is this done on Windows? This may require a JNI call to GetNetworkParams
     // https://msdn.microsoft.com/en-us/library/aa365968(VS.85).aspx.
     private static final DnsServerAddressStreamProvider DEFAULT_DNS_SERVER_ADDRESS_STREAM_PROVIDER =
-            UnixResolverDnsServerAddressStreamProvider.parseSilently();
+            // If on windows just use the DefaultDnsServerAddressStreamProvider.INSTANCE as otherwise
+            // we will log some error which may be confusing.
+            PlatformDependent.isWindows() ? DefaultDnsServerAddressStreamProvider.INSTANCE :
```
Scottmitch: If the issue is "the logs are confusing" ... then should we either:

- not log in `DnsServerAddressStreamProvider#parseSilently()` if on windows? 
- make parseSilently more strict and just throw if it fails ... and handle the errors at this level because we are introducing conditionals based on platform here

normanmaurer: honestly I think we should not try to call something called `Unix*` on windows, thats why I did the change here.

Scottmitch: makes sense ... I guess it still makes sense for parseSilently to tolerate a failure on unix based systems too ... in case the system configuration is not as expected.

normanmaurer: +1

normanmaurer:
Cherry-picked into 4.1 (0013567cd89bd3c3b042f44ccdcafc21fc126704)

PR #7404, opened 2017-11-13
Improved error message in FixedChannelPool

lutovich:
Motivation:

`FixedChannelPool` allows users to configure `acquireTimeoutMillis` and expects given value to be greater or equal to zero when timeout action is supplied. However, validation error message said that value is expected to be greater or equal to one. Code performs check against zero.

Modifications:

Changed error message to say that value greater or equal to zero is expected. Added test to check that zero is an acceptable value.

Result:

Exception with right error message is thrown.


normanmaurer:
Cherry-picked into 4.1 (25d146038cd825311b93bbabeb3f979796bd71d3) and 4.0 (33d7348fdab117ef7b20f8a150db33acd8f6c223).

@lutovich thanks!

PR #7405, opened 2017-11-13
HttpHeaders valuesIterator and contains improvements

Scottmitch:
Motivation:
In order to determine if a header contains a value we currently rely
upon getAll(..) and regular expressions. This operation is commonly used
during the encode and decode stage to determine the transfer encoding
(e.g. HttpUtil#isTransferEncodingChunked). This operation requires an
intermediate collection and possibly regular expressions for the
CombinedHttpHeaders use case which can be expensive.

Modifications:
- Add a valuesIterator to HttpHeaders and specializations of this method
for DefaultHttpHeaders, ReadOnlyHttpHeaders, and CombinedHttpHeaders.

Result:
Less intermediate collections and allocation overhead when determining
if HttpHeaders contains a name/value pair.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaders.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaders.java
@@ -31,6 +31,9 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 
+import static io.netty.util.AsciiString.contentEquals;
+import static io.netty.util.AsciiString.contentEqualsIgnoreCase;
+import static io.netty.util.AsciiString.trim;
 import static io.netty.util.internal.ObjectUtil.checkNotNull;
 
 /**
@@ -1146,7 +1149,7 @@ public abstract class HttpHeaders implements Iterable<Map.Entry<String, String>>
      */
     @Deprecated
     public static boolean equalsIgnoreCase(CharSequence name1, CharSequence name2) {
-        return AsciiString.contentEqualsIgnoreCase(name1, name2);
+        return contentEqualsIgnoreCase(name1, name2);
     }
 
     @Deprecated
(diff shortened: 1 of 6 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaders.java
```diff
@@ -1576,32 +1595,64 @@ public boolean contains(String name, String value, boolean ignoreCase) {
...
-            return false;
-        }
-
-        for (String v: values) {
-            if (contains(v, value, ignoreCase)) {
-                return true;
+        Iterator<? extends CharSequence> itr = valueCharSequenceIterator(name);
+        if (ignoreCase) {
+            while (itr.hasNext()) {
+                if (containsCommaSeparatedTrimmed(itr.next(), value, true)) {
```
normanmaurer: why not use the same loop and pass `ignoreCase` into `containsCommaSeparatedTrimmed(....)` ?

common/src/main/java/io/netty/util/AsciiString.java
```diff
@@ -1007,6 +1007,32 @@ public AsciiString toUpperCase() {
 
     /**
      * Copies this string removing white space characters from the beginning and end of the string.
+     * @param c The {@link CharSequence} to trim.
+     * @return a new string with characters {@code <= \\u0020} removed from the beginning and the end.
+     */
+    public static CharSequence trim(CharSequence c) {
+        if (c.getClass() == AsciiString.class) {
```
normanmaurer: why we not use an instanceof here as well ?

windie: checkNotNull(c)?

normanmaurer: @windie I think its fine as it will throw directly anyway.

windie:
+1 @normanmaurer 's comments. Otherwise, LGTM

common/src/main/java/io/netty/util/AsciiString.java
Scottmitch: we use a direct class check because `AsciiString` is a `final` class. This is consistent with other checks in this class.

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaders.java
Scottmitch: +1 good catch ... the code structure was previously different where the loop contents were unique but now I can simplify.

Scottmitch:
4.1 (0a47c590fe)

PR #7406, opened 2017-11-13
HttpMethod#valueOf improvement

Scottmitch:
Motivation:
HttpMethod#valueOf shows up on profiler results in the top set of
results. Since it is a relatively simple operation it can be improved in
isolation.

Modifications:
- Introduce a special case map which assigns each HttpMethod to a unique
index in an array and provides constant time lookup from a hash code
algorithm. When the bucket is matched we can then directly do equality
comparison instead of potentially following a linked structure when
HashMap has hash collisions.

Result:
~10% improvement in benchmark results for HttpMethod#valueOf

Benchmark                                     Mode  Cnt   Score   Error   Units
HttpMethodMapBenchmark.newMapKnownMethods    thrpt   16  31.831 ± 0.928  ops/us
HttpMethodMapBenchmark.newMapMixMethods      thrpt   16  25.568 ± 0.400  ops/us
HttpMethodMapBenchmark.newMapUnknownMethods  thrpt   16  51.413 ± 1.824  ops/us
HttpMethodMapBenchmark.oldMapKnownMethods    thrpt   16  29.226 ± 0.330  ops/us
HttpMethodMapBenchmark.oldMapMixMethods      thrpt   16  21.073 ± 0.247  ops/us
HttpMethodMapBenchmark.oldMapUnknownMethods  thrpt   16  49.081 ± 0.577  ops/us

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpMethod.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpMethod.java
@@ -15,11 +15,10 @@
  */
 package io.netty.handler.codec.http;
 
-import static io.netty.util.internal.ObjectUtil.checkNotNull;
 import io.netty.util.AsciiString;
 
-import java.util.HashMap;
-import java.util.Map;
+import static io.netty.util.internal.MathUtil.findNextPositivePowerOfTwo;
+import static io.netty.util.internal.ObjectUtil.checkNotNull;
 
 /**
  * The request method of HTTP or its derived protocols, such as
@@ -86,18 +85,19 @@ public class HttpMethod implements Comparable<HttpMethod> {
      */
     public static final HttpMethod CONNECT = new HttpMethod("CONNECT");
 
-    private static final Map<String, HttpMethod> methodMap = new HashMap<String, HttpMethod>();
+    private static final EnumNameMap<HttpMethod> methodMap;
 
     static {
-        methodMap.put(OPTIONS.toString(), OPTIONS);
-        methodMap.put(GET.toString(), GET);
-        methodMap.put(HEAD.toString(), HEAD);
-        methodMap.put(POST.toString(), POST);
-        methodMap.put(PUT.toString(), PUT);
-        methodMap.put(PATCH.toString(), PATCH);
-        methodMap.put(DELETE.toString(), DELETE);
-        methodMap.put(TRACE.toString(), TRACE);
-        methodMap.put(CONNECT.toString(), CONNECT);
+        methodMap = new EnumNameMap<HttpMethod>(
+                new EnumNameMap.Node<HttpMethod>(OPTIONS.toString(), OPTIONS),
+                new EnumNameMap.Node<HttpMethod>(GET.toString(), GET),
+                new EnumNameMap.Node<HttpMethod>(HEAD.toString(), HEAD),
+                new EnumNameMap.Node<HttpMethod>(POST.toString(), POST),
+                new EnumNameMap.Node<HttpMethod>(PUT.toString(), PUT),
+                new EnumNameMap.Node<HttpMethod>(PATCH.toString(), PATCH),
+                new EnumNameMap.Node<HttpMethod>(DELETE.toString(), DELETE),
+                new EnumNameMap.Node<HttpMethod>(TRACE.toString(), TRACE),
+                new EnumNameMap.Node<HttpMethod>(CONNECT.toString(), CONNECT));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/HttpMethod.java
```diff
@@ -173,4 +173,44 @@ public String toString() {
...
+        }
+
+        private static int hashCode(String name) {
+            // This hash code needs to produce a unique index for each HttpMethod. If new methods are added this
+            // algorithm will need to be adjusted. The goal is to have each enum name's hash value correlate to a unique
+            // index in the values array.
+            return name.hashCode() >>> 6;
+        }
+
+        private static final class Node<T> {
```
normanmaurer: This one is exposed via a package-private constructor so it should be package-private as well.

codec-http/src/main/java/io/netty/handler/codec/http/HttpMethod.java
```diff
@@ -173,4 +173,44 @@ public String toString() {
...
+        T get(String name) {
+            EnumNameMap.Node<T> node = values[hashCode(name) & valuesMask];
+            return node == null || !node.key.equals(name) ? null : node.value;
+        }
+
+        private static int hashCode(String name) {
+            // This hash code needs to produce a unique index for each HttpMethod. If new methods are added this
+            // algorithm will need to be adjusted. The goal is to have each enum name's hash value correlate to a unique
+            // index in the values array.
+            return name.hashCode() >>> 6;
```
normanmaurer: can you add a comment "how this would need to be adjusted" ? Like how does the `>>> 6` play into account here

microbench/src/main/java/io/netty/handler/codec/http/HttpMethodMapBenchmark.java
```diff
@@ -0,0 +1,215 @@
...
+        knownMethods = new String[] {
+                "OPTIONS",
+                "GET",
+                "HEAD",
+                "POST",
+                "PUT",
+                "PATCH",
+                "DELETE",
+                "TRACE",
+                "CONNECT"
```
normanmaurer: consider using `HttpMethod.OPTIONS.getName()` etc

microbench/src/main/java/io/netty/handler/codec/http/HttpMethodMapBenchmark.java
```diff
@@ -0,0 +1,215 @@
...
+@State(Scope.Benchmark)
+@Warmup(iterations = 5)
+@Measurement(iterations = 8)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+public class HttpMethodMapBenchmark extends AbstractMicrobenchmark {
+    private static final Map<String, HttpMethod> oldMap = new HashMap<String, HttpMethod>();
+    private static final SimpleStringMap<HttpMethod> newMap;
+    private static final String[] knownMethods;
+    private static final String[] mixMethods;
+    private static final String[] unknownMethods;
```
normanmaurer: use `uppercase` naming to be inline with the JLS

microbench/src/main/java/io/netty/handler/codec/http/HttpMethodMapBenchmark.java
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        private static int hashCode(String name) {
+            // This hash code needs to produce a unique index for each HttpMethod. If new methods are added this
+            // algorithm will need to be adjusted. The goal is to have each enum name's hash value correlate to a unique
+            // index in the values array.
+            return name.hashCode() >>> 6;
+        }
+
+        private static final class Node<T> {
```
normanmaurer: same comment as in the actual impl in HTTP

codec-http/src/main/java/io/netty/handler/codec/http/HttpMethod.java
Scottmitch: `private` inner class with `package private` constructor is common practice. Can you clarify why this needs to change? The inner class is private (only visible to the outer class), but the constructor is `package private` because its internal to the class and the outer class needs to construct it.

codec-http/src/main/java/io/netty/handler/codec/http/HttpMethod.java
Scottmitch: I don't have a generic algorithm to determine an appropriate hash code. The only constraint for this class to function as intended is given all the possible `HttpMethod` values, they must each have a unique entry in the array. If this is not the case the constructor will throw and let you know there is a collision (so it won't break silently). The hash code of each value `>>> 6` just happens to be unique relative to each other modulo `16`.

I'll add some additional comments ... let me know if you think there is anything outside of what is described above that would be helpful.

microbench/src/main/java/io/netty/handler/codec/http/HttpMethodMapBenchmark.java
Scottmitch: done

codec-http/src/main/java/io/netty/handler/codec/http/HttpMethod.java
normanmaurer: I read at some point that the JVM does some extra steps while dispatch stuff in this case while it not does when it is package-private or public. Maybe this is not true anymore.

normanmaurer: never mind... I somehow thought the EnumNameMap is package-private ...

microbench/src/main/java/io/netty/handler/codec/http/HttpMethodMapBenchmark.java
Scottmitch: It was this way originally but this doesn't simulate how the HttpObjectDecoder actually works (it builds a new String object), and was unfairly biased toward the new approach due to the `equals` short circuiting on reference equality checks.

I will add comments.

normanmaurer: ok

codec-http/src/main/java/io/netty/handler/codec/http/HttpMethod.java
```diff
@@ -173,4 +173,46 @@ public String toString() {
...
+        private final int valuesMask;
+
+        EnumNameMap(EnumNameMap.Node<T>... nodes) {
+            values = (EnumNameMap.Node<T>[]) new EnumNameMap.Node[findNextPositivePowerOfTwo(nodes.length)];
+            valuesMask = values.length - 1;
+            for (EnumNameMap.Node<T> node : nodes) {
+                int i = hashCode(node.key) & valuesMask;
+                if (values[i] != null) {
+                    throw new IllegalArgumentException("index " + i + " collision between values: [" +
+                            values[i].key + ", " + node.key + "]");
```
normanmaurer: nit: `"]"`/ `']'`

Scottmitch:
4.1 (93b144b7b4)

Issue #7351, opened 2017-10-30
Test failure in OpenJDK 8u141-jdk Docker container


mkw:
### Expected behavior

The Epoll implementation works in the OpenJDK 8u141-jdk Docker container.

I know that Netty doesn't specifically target Netty, but this is the official OpenJDK build (https://hub.docker.com/_/openjdk/), and is used to work.  I assume that something changed in the OS being used to build the official OpenJDK image.  I noticed this line in the Netty build output:

```
[INFO] --- build-helper-maven-plugin:1.10:regex-property (regex-combined-sendmmsg) @ netty-transport-native-epoll ---
[INFO] No match to regex '.*IO_NETTY_SENDMSSG_NOT_FOUND.*' found in '4.9.49-mobyldd (Debian GLIBC 2.24-11+deb9u1) 2.24'. The initial value '4.9.49-mobyldd (Debian GLIBC 2.24-11+deb9u1) 2.24' is left as-is...
```

### Actual behavior

Run:

`docker run -it -v `pwd`:`pwd` -v ~/.m2:/root/.m2 -w `pwd` netty-build mvn install --fae -pl transport-native-epoll`

In the output, one of the tests fails because getsockopt() throws "Operation not supported" somewhere in the native code:

```
Tests run: 9, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.001 sec <<< FAILURE! - in io.netty.channel.epoll.EpollSocketTest
testTrafficClass(io.netty.channel.epoll.EpollSocketTest)  Time elapsed: 0 sec  <<< ERROR!
io.netty.channel.ChannelException: getsockopt() failed: Operation not supported
```

In my own application code, I saw the following fragment of a stack trace when trying to get a list of supported options on a channel:

```
    io.netty.channel.ChannelException: getsockopt() failed: Operation not supported
        at io.netty.channel.unix.Socket.getTrafficClass(Native Method)
        at io.netty.channel.unix.Socket.getTrafficClass(Socket.java:368)
        at io.netty.channel.epoll.EpollSocketChannelConfig.getTrafficClass(EpollSocketChannelConfig.java:197)
        at io.netty.channel.epoll.EpollSocketChannelConfig.getOption(EpollSocketChannelConfig.java:88)
        at io.netty.channel.DefaultChannelConfig.getOptions(DefaultChannelConfig.java:94)
        at io.netty.channel.epoll.EpollSocketChannelConfig.getOptions(EpollSocketChannelConfig.java:58)
[...]

PR #7407, opened 2017-11-14
EpollSocketChannelConfig.getOptions() must not throw if TCP_FASTOPEN_…

normanmaurer:
…CONNECT is not supported

Motivation:

If a user calls EpollSocketChannelConfig.getOptions() and TCP_FASTOPEN_CONNECT is not supported we throw an exception.

Modifications:

- Just return 0 if ENOPROTOOPT is set.
- Add testcase

Result:

getOptions() works as expected.

### DIFF
--- a/transport-native-epoll/src/main/c/netty_epoll_linuxsocket.c
+++ b/transport-native-epoll/src/main/c/netty_epoll_linuxsocket.c
@@ -256,7 +256,14 @@ static jint netty_epoll_linuxsocket_isTcpQuickAck(JNIEnv* env, jclass clazz, jin
 
 static jint netty_epoll_linuxsocket_isTcpFastOpenConnect(JNIEnv* env, jclass clazz, jint fd) {
     int optval;
-    if (netty_unix_socket_getOption(env, fd, IPPROTO_TCP, TCP_FASTOPEN_CONNECT, &optval, sizeof(optval)) == -1) {
+    int optlen = sizeof(optval);
+    // We call getsockopt directly so we can handle ENOPROTOOPT by ourself.
+    if (getsockopt(fd, IPPROTO_TCP, TCP_FASTOPEN_CONNECT, &optval, &optlen) == -1) {
+        if (errno == ENOPROTOOPT) {
+            // Not supported by the system, so just return 0.
+            return 0;
+        }
+        netty_unix_socket_getOptionHandleError(env, errno);
         return -1;
     }
     return optval;
--- a/transport-native-epoll/src/test/java/io/netty/channel/epoll/EpollServerSocketChannelConfigTest.java
+++ b/transport-native-epoll/src/test/java/io/netty/channel/epoll/EpollServerSocketChannelConfigTest.java
@@ -17,12 +17,14 @@ package io.netty.channel.epoll;
 
 import io.netty.bootstrap.ServerBootstrap;
 import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.channel.ChannelOption;
 import io.netty.channel.EventLoopGroup;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
 import java.net.InetSocketAddress;
+import java.util.Map;
 
 import static org.junit.Assert.*;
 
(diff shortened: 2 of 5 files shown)


normanmaurer:
Was reported here https://github.com/netty/netty/issues/7351#issuecomment-344080229

mkw:
You beat me!  Thanks.  I was trying for a patch that would handle is more like TCP_FASTOPEN.  It appears that support for that is checked on startup and the set value stored on the Java side, only being set natively when its supported.

As part of that, I made a CentOS Docker build image after I noticed that you do the releases on RHEL 6.9.  Granted, still limited to the host system Kernel (4.9.x on Docker for Mac right now), but still possibly a bit closer to official.  Here it is in case it's useful:

```
FROM centos:6.9

RUN yum install -y \
        apr-devel \
        autoconf \
        automake \
        glibc-devel \
        libaio-devel \
        libtool \
        lksctp-tools \
        lsb-core \
        make \
        openssl-devel \
        tar \
        java-1.8.0-openjdk \
        java-1.8.0-openjdk-devel

ARG MAVEN_VERSION=3.5.2
ARG MAVEN_SHA=707b1f6e390a65bde4af4cdaf2a24d45fc19a6ded00fff02e91626e3e42ceaff
ARG MAVEN_BASE_URL=http://www-us.apache.org/dist/maven/maven-3/${MAVEN_VERSION}/binaries

RUN mkdir -p /usr/share/maven && \
    curl -fsSL -o /tmp/apache-maven.tar.gz ${MAVEN_BASE_URL}/apache-maven-$MAVEN_VERSION-bin.tar.gz && \
    echo "${MAVEN_SHA}  /tmp/apache-maven.tar.gz" | sha256sum -c - && \
    tar -xzf /tmp/apache-maven.tar.gz -C /usr/share/maven --strip-components=1 && \
    rm -f /tmp/apache-maven.tar.gz && \
    ln -s /usr/share/maven/bin/mvn /usr/bin/mvn

ENV JAVA_HOME /usr/lib/jvm/java
ENV MAVEN_HOME /usr/share/maven

VOLUME /root/.m2
VOLUME /opt/netty

WORKDIR /opt/netty

CMD ["/bin/bash"]
```

normanmaurer:
@carl-mastrangelo @Scottmitch any thought / comments ?

normanmaurer:
Cherry-picked into 4.1 (f115bf50cbda70964130236a09fd3045f1b250be)

transport-native-epoll/src/main/c/netty_epoll_linuxsocket.c
```diff
@@ -256,7 +256,14 @@ static jint netty_epoll_linuxsocket_isTcpQuickAck(JNIEnv* env, jclass clazz, jin
 
 static jint netty_epoll_linuxsocket_isTcpFastOpenConnect(JNIEnv* env, jclass clazz, jint fd) {
     int optval;
-    if (netty_unix_socket_getOption(env, fd, IPPROTO_TCP, TCP_FASTOPEN_CONNECT, &optval, sizeof(optval)) == -1) {
+    int optlen = sizeof(optval);
+    // We call getsockopt directly so we can handle ENOPROTOOPT by ourself.
+    if (getsockopt(fd, IPPROTO_TCP, TCP_FASTOPEN_CONNECT, &optval, &optlen) == -1) {
```
Scottmitch: ideally we wouldn't have to call `getsockopt` directly and instead use a method from `netty_unix_socket` (e.g. expose `netty_unix_socket_getOption0` as something like `netty_unix_socket_getOptionNoErrorHandling`)

normanmaurer: If you pe der I can add it to the header file and call it... let me do a prq

Scottmitch: sgtm ... just lets us isolate the OS system calls more easily and can be useful when there are differences between platforms.

normanmaurer: @Scottmitch https://github.com/netty/netty/pull/7425 PTAL

PR #7411, opened 2017-11-15
Allow to detect failed query caused by an Timeout / IO error and also…

normanmaurer:
… not cache these.

Motivation:

At the moment there is not way for the user to know if resolving a domain was failed because the domain was unkown or because of an IO error / timeout. If it was caused by an timeout / IO error the user may want to retry the query. Also if the query was failed because of an IO error / timeout we should not cache it.

Modifications:

- Add DnsNameResolverTimeoutException and include it in the UnkownHostException if the domain could not be resolved because of an timeout. This will allow the user to retry the query when inspecting the cause.
- Do not cache IO errors / timeouts
- Add unit test

Result:

Easier for users to implement retries for DNS querys and not cache IO errors / timeouts.

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsCache.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsCache.java
@@ -65,6 +65,8 @@ public interface DnsCache {
 
     /**
      * Cache the resolution failure for a given hostname.
+     * Be aware this <strong>won't</strong> be called with timeout / cancel / transport exceptions.
+      *
      * @param hostname the hostname
      * @param additionals the additional records
      * @param cause the resolution failure
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
@@ -137,14 +137,19 @@ abstract class DnsNameResolverContext<T> {
                 private int searchDomainIdx = initialSearchDomainIdx;
                 @Override
                 public void operationComplete(Future<T> future) throws Exception {
-                    if (future.isSuccess()) {
+                    Throwable cause = future.cause();
+                    if (cause == null) {
                         promise.trySuccess(future.getNow());
-                    } else if (searchDomainIdx < searchDomains.length) {
-                        doSearchDomainQuery(hostname + '.' + searchDomains[searchDomainIdx++], this);
-                    } else if (!startWithoutSearchDomain) {
-                        internalResolve(promise);
                     } else {
-                        promise.tryFailure(new SearchDomainUnknownHostException(future.cause(), hostname));
+                        if (DnsNameResolver.isTransportOrTimeoutError(cause)) {
+                            promise.tryFailure(new SearchDomainUnknownHostException(cause, hostname));
+                        } else if (searchDomainIdx < searchDomains.length) {
+                            doSearchDomainQuery(hostname + '.' + searchDomains[searchDomainIdx++], this);
+                        } else if (!startWithoutSearchDomain) {
+                            internalResolve(promise);
+                        } else {
+                            promise.tryFailure(new SearchDomainUnknownHostException(cause, hostname));
+                        }
                     }
                 }
             });
@@ -166,6 +171,9 @@ abstract class DnsNameResolverContext<T> {
         SearchDomainUnknownHostException(Throwable cause, String originalHostname) {
             super("Search domain query failed. Original hostname: '" + originalHostname + "' " + cause.getMessage());
             setStackTrace(cause.getStackTrace());
+
+            // Preserve the cause
+            initCause(cause.getCause());
         }
 
         @Override
@@ -189,11 +197,11 @@ abstract class DnsNameResolverContext<T> {
         assert recordTypes.length > 0;
         final int end = recordTypes.length - 1;
         for (int i = 0; i < end; ++i) {
-            if (!query(hostname, recordTypes[i], nameServerAddressStream.duplicate(), promise)) {
+            if (!query(hostname, recordTypes[i], nameServerAddressStream.duplicate(), promise, null)) {
                 return;
             }
         }
-        query(hostname, recordTypes[end], nameServerAddressStream, promise);
+        query(hostname, recordTypes[end], nameServerAddressStream, promise, null);
     }
 
     /**
@@ -283,19 +291,20 @@ abstract class DnsNameResolverContext<T> {
 
     private void query(final DnsServerAddressStream nameServerAddrStream, final int nameServerAddrStreamIndex,
                        final DnsQuestion question,
-                       final Promise<T> promise) {
+                       final Promise<T> promise, Throwable cause) {
         query(nameServerAddrStream, nameServerAddrStreamIndex, question,
-                parent.dnsQueryLifecycleObserverFactory().newDnsQueryLifecycleObserver(question), promise);
+                parent.dnsQueryLifecycleObserverFactory().newDnsQueryLifecycleObserver(question), promise, cause);
     }
 
     private void query(final DnsServerAddressStream nameServerAddrStream,
                        final int nameServerAddrStreamIndex,
                        final DnsQuestion question,
                        final DnsQueryLifecycleObserver queryLifecycleObserver,
-                       final Promise<T> promise) {
+                       final Promise<T> promise,
+                       final Throwable cause) {
         if (nameServerAddrStreamIndex >= nameServerAddrStream.size() || allowedQueries == 0 || promise.isCancelled()) {
             tryToFinishResolve(nameServerAddrStream, nameServerAddrStreamIndex, question, queryLifecycleObserver,
-                               promise);
+                               promise, cause);
             return;
         }
 
@@ -319,21 +328,22 @@ abstract class DnsNameResolverContext<T> {
                     return;
                 }
 
+                final Throwable queryCause = future.cause();
                 try {
-                    if (future.isSuccess()) {
+                    if (queryCause == null) {
                         onResponse(nameServerAddrStream, nameServerAddrStreamIndex, question, future.getNow(),
                                    queryLifecycleObserver, promise);
                     } else {
                         // Server did not respond or I/O error occurred; try again.
-                        queryLifecycleObserver.queryFailed(future.cause());
-                        query(nameServerAddrStream, nameServerAddrStreamIndex + 1, question, promise);
+                        queryLifecycleObserver.queryFailed(queryCause);
+                        query(nameServerAddrStream, nameServerAddrStreamIndex + 1, question, promise, queryCause);
                     }
                 } finally {
                     tryToFinishResolve(nameServerAddrStream, nameServerAddrStreamIndex, question,
                                        // queryLifecycleObserver has already been terminated at this point so we must
                                        // not allow it to be terminated again by tryToFinishResolve.
                                        NoopDnsQueryLifecycleObserver.INSTANCE,
-                                       promise);
+                                       promise, queryCause);
                 }
             }
         });
@@ -366,7 +376,7 @@ abstract class DnsNameResolverContext<T> {
             // Retry with the next server if the server did not tell us that the domain does not exist.
             if (code != DnsResponseCode.NXDOMAIN) {
                 query(nameServerAddrStream, nameServerAddrStreamIndex + 1, question,
-                      queryLifecycleObserver.queryNoAnswer(code), promise);
+                      queryLifecycleObserver.queryNoAnswer(code), promise, null);
             } else {
                 queryLifecycleObserver.queryFailed(NXDOMAIN_QUERY_FAILED_EXCEPTION);
             }
@@ -420,7 +430,7 @@ abstract class DnsNameResolverContext<T> {
 
                 if (!nameServers.isEmpty()) {
                     query(parent.uncachedRedirectDnsServerStream(nameServers), 0, question,
-                          queryLifecycleObserver.queryRedirected(unmodifiableList(nameServers)), promise);
+                          queryLifecycleObserver.queryRedirected(unmodifiableList(nameServers)), promise, null);
                     return true;
                 }
             }
@@ -601,7 +611,8 @@ abstract class DnsNameResolverContext<T> {
                             final int nameServerAddrStreamIndex,
                             final DnsQuestion question,
                             final DnsQueryLifecycleObserver queryLifecycleObserver,
-                            final Promise<T> promise) {
+                            final Promise<T> promise,
+                            final Throwable cause) {
         // There are no queries left to try.
         if (!queriesInProgress.isEmpty()) {
             queryLifecycleObserver.queryCancelled(allowedQueries);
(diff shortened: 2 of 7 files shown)


normanmaurer:
@slandelle FYI

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
```diff
@@ -314,21 +315,22 @@ public void operationComplete(Future<AddressedEnvelope<DnsResponse, InetSocketAd
...
 
+                final Throwable queryCause = future.cause();
                 try {
-                    if (future.isSuccess()) {
+                    if (queryCause == null) {
                         onResponse(nameServerAddrStream, nameServerAddrStreamIndex, question, future.getNow(),
                                    queryLifecycleObserver, promise);
                     } else {
                         // Server did not respond or I/O error occurred; try again.
                         queryLifecycleObserver.queryFailed(future.cause());
```
Scottmitch: `future.cause()` -> `queryCause`

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
```diff
@@ -698,10 +701,15 @@ private void finishResolve(Promise<T> promise) {
                 .append(' ');
             }
         }
-        final UnknownHostException cause = new UnknownHostException(buf.toString());
-
-        resolveCache.cache(hostname, additionals, cause, parent.ch.eventLoop());
-        promise.tryFailure(cause);
+        final UnknownHostException unknownHostException = new UnknownHostException(buf.toString());
+        if (cause == null) {
+            // Only cache if the failure was not because of an IO error / timeout that was caused by the query
```
Scottmitch: couldn't the cache decide if it wants to retain or not instead of always making that decision here?

resolver-dns/src/main/java/io/netty/resolver/dns/DnsQueryContext.java
```diff
@@ -213,12 +213,13 @@ private void setFailure(String message, Throwable cause) {
            .append(" (no stack trace available)");
 
         final DnsNameResolverException e;
-        if (cause != null) {
-            e = new DnsNameResolverException(nameServerAddr, question(), buf.toString(), cause);
+        if (cause == null) {
+            // This was caused by an timeout so use DnsNameResolverTimeoutException to allow the user to
```
Scottmitch: IIUC this may not always be true. This method is also called when a write failure occurs (from `onQueryWriteCompletion`), right? Consider passing this exception (or some indication as to what the exception is) in to this method where the root cause is known instead of trying to infer it here.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
```diff
@@ -617,30 +620,30 @@ void tryToFinishResolve(final DnsServerAddressStream nameServerAddrStream,
...
+                          promise, cause);
                 }
                 return;
             }
 
             queryLifecycleObserver.queryFailed(NAME_SERVERS_EXHAUSTED_EXCEPTION);
 
             // .. and we could not find any A/AAAA records.
-            if (!triedCNAME) {
+            if (cause == null && !triedCNAME) {
```
Scottmitch: is it possible that the current query failed, but the CNAME query may succeed? If so we may be prematurely giving up on the query.

normanmaurer: I think if the A / AAA record query failed due of an timeout or IO error we want to not even try the CNAME and just retry or whatever the user wants to do. Not strong on this tho.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsQueryContext.java
normanmaurer: Yes and in this case `cause != null`. it is only null if we call it because of a timeout

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
normanmaurer: I dont think it will make any sense to put this in the cache ever as it was failed due a timeout or IO error to the dns server (especially an timeout is most likely caused because the dns server enforce some query limit, like public google name servers).

normanmaurer:
@Scottmitch btw I wonder if we should also add another option to DnsNameResolver to allow configure a retry on timeout / IO error. I think that is usually what you want and expect.

normanmaurer:
@Scottmitch another thing I could do is create a "wrapper" that wraps DnsResolver that will do the retries.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
normanmaurer: @Scottmitch That said I see you point that the cache could handle it.... If we want to make this part of the cache impl I would at least suggest we not cache it in our default impl.  WDYT ?

normanmaurer:
@Scottmitch also note that I added a second commit to show how this could be used to implement retries as part of the `DnsResolver`

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
Lukasa: FWIW, I don't think caching transport errors ever really makes much sense: the cache has no particular insight into whether these errors are likely to be transient or permanent, and has to pick some arbitrary TTL to use for this cache entry. Kinda ends up making no sense.

normanmaurer: @Lukasa +1 ... @Scottmitch you are strong about it ?

normanmaurer:
@vietj @slandelle FYI as well...

normanmaurer:
@trustin @Scottmitch @vietj @slandelle also please note that this pr has two commits and the second one is only to show how you could implement the retry in the `DnsResolver`. I am not sure if we should do this here or if the user should do this by himself / herself. I kind of think we may want to do this for the user.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
Scottmitch: I think we should either:

- Use the cache API consistently, and let the cache decide (the assumption is the cache already has a policy for this)
- Remove the error path from the caching API completely if we think it is of little use

It doesn't make sense to me to conditionally call the cache API.

normanmaurer: @Scottmitch removing the error-path from the caching API makes no sense as this would mean we not cache resolve failures that happen because there a no `A` or `AAAA` records present.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
Scottmitch: Isn't it possible the CNAME query may actually try a different DNS server (via the `DnsServerAddressStream`), or be cached?

normanmaurer: Yes its possible and I think this is even a problem... We should only try CNAME if an A or AAAA record not exists (and this was the case for the failure), not if we failed because of a timeout.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
Scottmitch: Agreed. One of the PR goals is state as:

> At the moment there is not way for the user to know if resolving a domain was failed because the domain was unkown or because of an IO error / timeout

How will a user differentiate between the two scenarios? Is the intention that a user will get a failure, and inspect the cause of that failure? Seems like you don't want the cache to use the same technique to w/e take action it may want to (clear existing state, cache failure for some amount of time ... seems to be the current assumption of the API, etc...).

normanmaurer: The user would check `UnkownHostException.getCause() instanceof DnsNameResolverException`... Like I said I am not super strong on this but honestly I can't think of any good use-case that would make sense to have the cache handle this. But maybe I am wrong... That said if you prefer I can push it to the cache and just make our default implement ignore these. WDYT ?

resolver-dns/src/main/java/io/netty/resolver/dns/DnsQueryContext.java
Scottmitch: > it is only null if we call it because of a timeout

IIUC here is the stack trace that may call this method in a non-timeout case:

`DnsQueryContext#setFailure`
`DnsNameResolverContext#onQueryWriteCompletion`
`DnsNameResolverContext#writeQuery`

This may be the result of a transport error, not a timeout, no?

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
normanmaurer: one thing that would be "hard" to do is retries as I did in the second commit here if the cache handles it. As then there is no good way to understand if it makes sense to retry or not as if its in the cache we would basically retry and get the same from the cache all the time

resolver-dns/src/main/java/io/netty/resolver/dns/DnsQueryContext.java
normanmaurer: @Scottmitch but only the one for the timeout will use `setFailure` and passing `null` into it as `cause`. So whenever `cause == null` we know its a timeout. I am missing something ?

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
Scottmitch: > The user would check UnkownHostException.getCause() instanceof DnsNameResolverException

Is it reasonable to provide a utility method which provides this functionality? Propagating the cause exception list can be easy to get wrong. For example what if we later put this exception 2 levels deep instead of at the top level?

> That said if you prefer I can push it to the cache and just make our default implement ignore these.

I'm also not strong, but it seems like the assumption is existing implementations of `DnsCache` are accounting for these scenarios ... unless we think this is an API bug. Either way I think it would be useful to clarify the API. Either saying something like `this won't be called with timeout/cancel/transport exceptions` or `this will be called for timeout/cancel/transport exceptions and if the cache wants to ignore these you should use <utility method> to detect these types of exceptions`.

normanmaurer: > Is it reasonable to provide a utility method which provides this functionality? Propagating the cause exception list can be easy to get wrong. For example what if we later put this exception 2 levels deep instead of at the top level?

Yep sure I can add such a method. Something like `isCausedByTransportError(...)` or whatever. Any good idea for a name ? 

 > I'm also not strong, but it seems like the assumption is existing implementations of DnsCache are accounting for these scenarios ... unless we think this is an API bug. Either way I think it would be useful to clarify the API. Either saying something like this won't be called with timeout/cancel/transport exceptions or this will be called for timeout/cancel/transport exceptions and if the cache wants to ignore these you should use <utility method> to detect these types of exceptions.

Please see https://github.com/netty/netty/pull/7411#discussion_r152081386

Scottmitch: > As then there is no good way to understand if it makes sense to retry or not as if its in the cache we would basically retry and get the same from the cache all the time

I would imagine the retry layers and the cache layer are independent. If the retry layer wants to retry due to a transport failure/timeout it is free todo so. If the cache wants to return a result that fails the request immediately it is also free todo so. Do we need to couple the 2 layers?

Also does clarifying the javadocs either way make sense?

normanmaurer: @Scottmitch what I am saying is that if we want to have the retry logic as part of the `DnsNameResolver` it will only work if the cache does not cache transport errors as otherwise the cache will just return the same result for each retry. 

And yep clarification of the java docs makes sense once we agree and know what to do ;)

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
Scottmitch: Oh I see ... for timeout/cancel this makes sense ... what about transport error though?

normanmaurer: I think its the same here... any reason why you think this is not the case ?

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
Scottmitch: I'm fine with the change ... the scenarios I'm thinking of are theoretical and in the optimization territory. Since the APIs are experimental we can always change this later if necessary.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsQueryContext.java
Scottmitch: ahh my mistake. yes what you are saying makes sense. lgtm.

normanmaurer:
@Scottmitch PTAL again

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
Scottmitch: if its a transport failure ... this may not apply to the CNAME query because it may use a different transport, right?

resolver-dns/src/main/java/io/netty/resolver/dns/DnsCache.java
```diff
@@ -65,6 +65,8 @@ DnsCacheEntry cache(String hostname, DnsRecord[] additionals, InetAddress addres
 
     /**
      * Cache the resolution failure for a given hostname.
+     * Be aware this <strong>won't</strong> be called with timeout / cancel /transport exceptions.
```
Scottmitch: nit `/transport` -> `/ transport` for consistency

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
```diff
@@ -617,30 +620,34 @@ void tryToFinishResolve(final DnsServerAddressStream nameServerAddrStream,
...
                 }
                 return;
             }
 
             queryLifecycleObserver.queryFailed(NAME_SERVERS_EXHAUSTED_EXCEPTION);
 
             // .. and we could not find any A/AAAA records.
-            if (!triedCNAME) {
+
+            // If cause != null we know this was caused by a timeout / cancel /transport exception. In this case we
```
Scottmitch: nit `/transport` -> `/ transport` for consistency

normanmaurer:
Cherry-picked into 4.1 (12a413bf024e64946b28ed9aa3065a31b9420ad2).

normanmaurer:
Reopen this one as I reverted it ... I need to do some more changes because of some changes that were merged in 4.1 in the meantime

normanmaurer:
Now for real ;)

Cherry-picked into 4.1 (c1b1d6268aca287f310c5feefa6606eb6f3f6a70)

PR #7414, opened 2017-11-17
Fix DN resolution when ndots is greater than 1

shyiko:
Motivation:

DN resolution does not fall back to the "original name" lookup after search list is checked. This results in a failure to resolve any name (outside of search list) that has number of dots less than resolv.conf's ndots value (which, for example, is often the case in the context of Kubernetes where kubelet passes on resolv.conf containing "options ndots:5").

It also does not go through the search list in a situation described in resolv.conf man:
"The default for n[dots] is 1, meaning that if there are any dots in a name, the name will be tried first as an absolute name before any search list elements are appended to it."

Modifications:

DnsNameResolverContext::resolve was updated to match Go's https://github.com/golang/go/blob/release-branch.go1.9/src/net/dnsclient_unix.go#L338 logic.

Result:
DnsNameResolverContext::resolve will now try to resolve "original name" if search list yields no results when number of dots in the original name is less than resolv.conf's ndots value. It will also go through the search list in case "origin name" resolution fails and number of dots is equal or larger than resolv.conf's ndots value.

(ICLA signed)

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
@@ -125,31 +125,41 @@ abstract class DnsNameResolverContext<T> {
     }
 
     void resolve(final Promise<T> promise) {
-        if (parent.searchDomains().length == 0 || parent.ndots() == 0 || StringUtil.endsWith(hostname, '.')) {
+        final String[] searchDomains = parent.searchDomains();
+        if (searchDomains.length == 0 || parent.ndots() == 0 || StringUtil.endsWith(hostname, '.')) {
             internalResolve(promise);
         } else {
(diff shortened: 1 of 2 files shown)


normanmaurer:
@slandelle PTAL as well

slandelle:
Thanks! I guess @vietj might be interested too.

normanmaurer:
@slandelle good point +1

normanmaurer:
@slandelle @vietj thoughts ?

slandelle:
@normanmaurer LGTM

normanmaurer:
Cherry-picked into 4.1 (844d804aba4b459215f05c9ade59157f6c2cfb4b)

normanmaurer:
@shyiko thanks!

Scottmitch:
small followup: https://github.com/netty/netty/pull/7424

PR #7415, opened 2017-11-17
Correctly propagate channelInactive even if cleanup throws

normanmaurer:
Motivation:

Its possible that cleanup() will throw if invalid data is passed into the wrapped EmbeddedChannel. We need to ensure we still call channelInactive(...) in this case.

Modifications:

- Correctly forward Exceptions caused by cleanup()
- Ensure all content is released when cleanup() throws
- Add unit tests

Result:

Correctly handle the case when cleanup() throws.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentDecoder.java
@@ -195,13 +195,13 @@ public abstract class HttpContentDecoder extends MessageToMessageDecoder<HttpObj
 
     @Override
     public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
-        cleanup();
+        cleanupSafely(ctx);
         super.handlerRemoved(ctx);
     }
 
     @Override
     public void channelInactive(ChannelHandlerContext ctx) throws Exception {
-        cleanup();
+        cleanupSafely(ctx);
         super.channelInactive(ctx);
     }
 
@@ -214,17 +214,18 @@ public abstract class HttpContentDecoder extends MessageToMessageDecoder<HttpObj
     private void cleanup() {
         if (decoder != null) {
             // Clean-up the previous decoder if not cleaned up correctly.
-            if (decoder.finish()) {
-                for (;;) {
-                    ByteBuf buf = decoder.readInbound();
-                    if (buf == null) {
-                        break;
-                    }
-                    // Release the buffer
-                    buf.release();
-                }
-            }
+            decoder.finishAndReleaseAll();
             decoder = null;
+        }
+    }
+
+    private void cleanupSafely(ChannelHandlerContext ctx) {
+        try {
+            cleanup();
+        } catch (Throwable cause) {
+            // If cleanup throws any error we need to propagate it through the pipeline
+            // so we don't fail to propagate pipeline events.
+            ctx.fireExceptionCaught(cause);
         }
     }
 
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentEncoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentEncoder.java
@@ -289,31 +289,31 @@ public abstract class HttpContentEncoder extends MessageToMessageCodec<HttpReque
 
     @Override
     public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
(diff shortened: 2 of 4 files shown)


normanmaurer:
Reported in Vert.x as https://github.com/eclipse/vert.x/issues/2212

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentDecoder.java
```diff
@@ -214,20 +214,21 @@ public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
...
+            decoder.finishAndReleaseAll();
             decoder = null;
         }
     }
 
+    private void cleanupSafely(ChannelHandlerContext ctx) {
+        try {
+            cleanup();
+        } catch (Exception e) {
+            // If cleanup throws any error we need to propergate it through the pipeline and then ensure we call
```
ejona86: s/propergate/propagate/ Ditto elsewhere.

normanmaurer:
@ejona86 the test basically simulate the situation where `EmbeddedChannel.finish()` (the channel that is used internally by the compressor / decompressor) will throw when called. Before this caused to not receive the `channelInactive(...)` event.

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentDecoder.java
normanmaurer: addressed

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentEncoder.java
```diff
@@ -289,34 +289,34 @@ private boolean encodeContent(HttpContent c, List<Object> out) {
...
-            }
+            encoder.finishAndReleaseAll();
             encoder = null;
         }
     }
 
+    private void cleanupSafely(ChannelHandlerContext ctx) {
+        try {
+            cleanup();
+        } catch (Exception e) {
```
Scottmitch: should this be `Throwable` like in `AbstractChannelHandlerContext` so nothing escapes the pipeline?

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentDecoder.java
```diff
@@ -214,20 +214,21 @@ public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
...
             decoder = null;
         }
     }
 
+    private void cleanupSafely(ChannelHandlerContext ctx) {
+        try {
+            cleanup();
+        } catch (Exception e) {
+            // If cleanup throws any error we need to propagate it through the pipeline and then ensure we call
+            // super.channelInactive(...) so the event is not lost.
```
Scottmitch: this is also used for `handlerRemoved` ... should this comment be more generic (e.g. "so we don't fail to propagate pipeline events...")

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentEncoder.java
normanmaurer: @Scottmitch I used `Exception` as we want to review our usage of `catch (Throwable...)` anyway as most likely we should do this most of the times. That said I am not strong on this.

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentDecoder.java
normanmaurer: Yep let me adjust it.

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentEncoder.java
Scottmitch: I see ... this may fall into the "we won't cleanup correctly if we let Throwable slip through" category ... because the pipeline won't be notified of events.

normanmaurer: True enough... let me just use `Throwable` for now

johnou: Or keep exception and use try finally?

normanmaurer:
Cherry-picked into 4.1 (7f4ade7e7d6f907a30926c54e16abefe88e81a8e) and 4.0 (4375b6da22)

Issue #7418, opened 2017-11-18
Sending DefaultLastContent with trailers causes "io.netty.handler.codec.EncoderException: HttpServerCodec$HttpServerResponseEncoder must produce at least one message."


pkolaczk:
### Expected behavior
Writing `DefaultLastContent` object with trailing headers sends an empty chunk with trailing headers.

### Actual behavior
Writing `DefaultLastContent` with trailing headers causes EncoderException:
```
io.netty.handler.codec.EncoderException: HttpServerCodec$HttpServerResponseEncoder must produce at least one message.
        at io.netty.handler.codec.MessageToMessageEncoder.write(MessageToMessageEncoder.java:98) ~[netty-all-4.0.52.Final.jar:4.0.52.Final]
        at io.netty.channel.CombinedChannelDuplexHandler.write(CombinedChannelDuplexHandler.java:348) ~[netty-all-4.0.52.Final.jar:4.0.52.Final]
        at io.netty.channel.AbstractChannelHandlerContext.invokeWrite0(AbstractChannelHandlerContext.java:732) ~[netty-all-4.0.52.Final.jar:4.0.52.Final]
        at io.netty.channel.AbstractChannelHandlerContext.invokeWrite(AbstractChannelHandlerContext.java:724) ~[netty-all-4.0.52.Final.jar:4.0.52.Final]
        at io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:809) ~[netty-all-4.0.52.Final.jar:4.0.52.Final]
        at io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:717) ~[netty-all-4.0.52.Final.jar:4.0.52.Final]
        at io.netty.handler.logging.LoggingHandler.write(LoggingHandler.java:254) ~[netty-all-4.0.52.Final.jar:4.0.52.Final]
        at io.netty.channel.AbstractChannelHandlerContext.invokeWrite0(AbstractChannelHandlerContext.java:732) ~[netty-all-4.0.52.Final.jar:4.0.52.Final]
        at io.netty.channel.AbstractChannelHandlerContext.invokeWriteAndFlush(AbstractChannelHandlerContext.java:795) ~[netty-all-4.0.52.Final.jar:4.0.52.Final]
        at io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:807) ~[netty-all-4.0.52.Final.jar:4.0.52.Final]
[...]

PR #7419, opened 2017-11-19
Add testcase for writing empty last content with headers.

normanmaurer:
Motivation:

https://github.com/netty/netty/issues/7418 reported an issue with writing a LastHttpContent with trailers set.

Modifications:

Add unit test to ensure this issue is fixed in latest netty release.

Result:

Ensure code is correct.

### DIFF
--- a/codec-http/src/test/java/io/netty/handler/codec/http/HttpRequestEncoderTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/HttpRequestEncoderTest.java
@@ -156,16 +156,26 @@ public class HttpRequestEncoderTest {
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


normanmaurer:
@pkolaczk your reported bug is only present in the latest 4.0 code so this is only against 4.1 to ensure its "fixed" there.

normanmaurer:
@Scottmitch please note this should only go into 4.1 ... 4.0 should get https://github.com/netty/netty/pull/7420

normanmaurer:
Cherry-picked into 4.1 (e5e4c18c1be1d32c5bcb9e6aa0ee300e9c89c18d)

PR #7420, opened 2017-11-19
Fix writing of empty LastHttpContent with trailers

normanmaurer:
Motivation:

4732fabb1664acae0a545b6496066acb8d39ecb7 introduced a regression in HttpObjectEncoder which will lead to buffer leak and IllegalStateException when a LastHttpContent with trailers is written.

Modifications:

- Correctly add the buffer to the encoded list.
- Add testcases

Result:

Fixes [#7418]


normanmaurer:
@Scottmitch this was introduced by https://github.com/netty/netty/commit/4732fabb1664acae0a545b6496066acb8d39ecb7. Seems like something went wrong when you backported the change to 4.0 as the problem is not present in 4.1 :(

normanmaurer:
@pkolaczk PTAL... this should fix it

Scottmitch:
yah that back port looks bad :(. I didn't intend to back port any of the `padSizeForAccumulation` stuff. It seems like you actually back ported it (https://github.com/netty/netty/pull/7161#issuecomment-326241072) 😜

normanmaurer:
@Scottmitch doh! Hah yeah it seems so... All the blame and shame on me 👎

normanmaurer:
Cherry-picked into 4.0 (8009c4c8a0514dc63b2cc6c437f9a8ebc2176579)

PR #7421, opened 2017-11-20
HttpConversionUtils avoid intermediate collection allocation

Scottmitch:
Modifications:
HttpConversionUtil#toLowercaseMap requires an intermediate List to be allocated. This can be avoided with the recently added value iterator methods.

Modifications:
- Use HttpHeaders#valueCharSequenceIterator instead of getAll

Result:
Less intermediate object allocation and copying.


Scottmitch:
@mosesn - PTAL

Scottmitch:
4.1 (a3e41ba6eb)

PR #7422, opened 2017-11-20
HttpConversionUtil TE filtering robustness

Scottmitch:
Motivation:
HttpConversionUtil#toHttp2Headers has special code to filter the TE header name. However this filtering code may result in adding the <TE, TRAILERS> tuple in scenarios that are not appropriate. For example if a value containing trailers is seen it will be added, but the value could not actually be equal to trailers. Also CSV values are not supported.

Modifications:
- Account for CSV header values
- Account for the value containing 'trailers' but not actually being equal to 'trailers'

Result:
More robust parsing of the TE header.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/HttpConversionUtil.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/HttpConversionUtil.java
@@ -24,7 +24,6 @@ import io.netty.handler.codec.http.FullHttpMessage;
 import io.netty.handler.codec.http.FullHttpRequest;
 import io.netty.handler.codec.http.FullHttpResponse;
 import io.netty.handler.codec.http.HttpHeaderNames;
-import io.netty.handler.codec.http.HttpHeaderValues;
 import io.netty.handler.codec.http.HttpHeaders;
 import io.netty.handler.codec.http.HttpMessage;
 import io.netty.handler.codec.http.HttpMethod;
@@ -38,8 +37,13 @@ import io.netty.util.internal.UnstableApi;
 
 import java.net.URI;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map.Entry;
 
+import static io.netty.handler.codec.http.HttpHeaderNames.CONNECTION;
+import static io.netty.handler.codec.http.HttpHeaderNames.COOKIE;
+import static io.netty.handler.codec.http.HttpHeaderNames.TE;
+import static io.netty.handler.codec.http.HttpHeaderValues.TRAILERS;
 import static io.netty.handler.codec.http.HttpScheme.HTTP;
 import static io.netty.handler.codec.http.HttpScheme.HTTPS;
 import static io.netty.handler.codec.http.HttpUtil.isAsteriskForm;
@@ -48,11 +52,15 @@ import static io.netty.handler.codec.http2.Http2Error.PROTOCOL_ERROR;
 import static io.netty.handler.codec.http2.Http2Exception.connectionError;
 import static io.netty.handler.codec.http2.Http2Exception.streamError;
 import static io.netty.util.AsciiString.EMPTY_STRING;
+import static io.netty.util.AsciiString.contentEqualsIgnoreCase;
+import static io.netty.util.AsciiString.indexOf;
+import static io.netty.util.AsciiString.trim;
 import static io.netty.util.ByteProcessor.FIND_COMMA;
 import static io.netty.util.ByteProcessor.FIND_SEMI_COLON;
 import static io.netty.util.internal.ObjectUtil.checkNotNull;
 import static io.netty.util.internal.StringUtil.isNullOrEmpty;
 import static io.netty.util.internal.StringUtil.length;
+import static io.netty.util.internal.StringUtil.unescapeCsvFields;
 
 /**
  * Provides utility methods and constants for the HTTP/2 to HTTP conversion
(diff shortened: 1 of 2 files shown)


Scottmitch:
@mosesn - PTAL

codec-http2/src/main/java/io/netty/handler/codec/http2/HttpConversionUtil.java
```diff
@@ -446,18 +450,28 @@ public static Http2Headers toHttp2Headers(HttpHeaders inHeaders, boolean validat
...
             Entry<CharSequence, CharSequence> entry = iter.next();
             final AsciiString aName = AsciiString.of(entry.getKey()).toLowerCase();
             if (!HTTP_TO_HTTP2_HEADER_BLACKLIST.contains(aName) &&
                 !connectionBlacklist.contains(aName)) {
                 // https://tools.ietf.org/html/rfc7540#section-8.1.2.2 makes a special exception for TE
-                if (aName.contentEqualsIgnoreCase(HttpHeaderNames.TE)) {
-                    if (AsciiString.containsIgnoreCase(entry.getValue(), HttpHeaderValues.TRAILERS)) {
-                        out.add(HttpHeaderNames.TE, HttpHeaderValues.TRAILERS);
+                if (aName.contentEqualsIgnoreCase(TE)) {
+                    if (AsciiString.indexOf(entry.getValue(), ',', 0) == -1) {
```
mosesn: how would you feel about splitting this section out into a method?  this method is huge, not that easy to read right now.

codec-http2/src/main/java/io/netty/handler/codec/http2/HttpConversionUtil.java
```diff
@@ -446,18 +450,28 @@ public static Http2Headers toHttp2Headers(HttpHeaders inHeaders, boolean validat
...
+                        if (AsciiString.contentEqualsIgnoreCase(entry.getValue(), TRAILERS)) {
```
mosesn: should we trim this?

codec-http2/src/main/java/io/netty/handler/codec/http2/HttpConversionUtil.java
```diff
@@ -446,18 +450,28 @@ public static Http2Headers toHttp2Headers(HttpHeaders inHeaders, boolean validat
...
+                            out.add(TE, TRAILERS);
+                        }
+                    } else {
+                        List<CharSequence> teValues = StringUtil.unescapeCsvFields(entry.getValue());
+                        for (CharSequence teValue : teValues) {
+                            if (AsciiString.contentEqualsIgnoreCase(teValue, TRAILERS)) {
```
mosesn: trim?

codec-http2/src/main/java/io/netty/handler/codec/http2/HttpConversionUtil.java
mosesn: ffti: should we use value.forEachByte here, for consistency?

normanmaurer: If easy to do +1

codec-http2/src/main/java/io/netty/handler/codec/http2/HttpConversionUtil.java
normanmaurer: good question... @carl-mastrangelo @Scottmitch @ejona86 WDYT ?

codec-http2/src/main/java/io/netty/handler/codec/http2/HttpConversionUtil.java
normanmaurer: good question... @carl-mastrangelo @Scottmitch @ejona86 WDYT ?

ejona86: Wouldn't hurt, but doesn't seem necessary. Although it's probably reading too much into the language, the spec does say 'it MUST NOT contain any value other than "trailers".' :)

codec-http2/src/main/java/io/netty/handler/codec/http2/HttpConversionUtil.java
Scottmitch: I don't think we want to force the conversion here. We will only ever use the `TRAILERS` value which is already an `AsciiString`, and we may get to take advantage of intrinsics in case the value is a `String`. In the other cases it makes more sense to force to `AsciiString` because we will either use the value for encoding later and/or take `subSequence`s without copying.

codec-http2/src/main/java/io/netty/handler/codec/http2/HttpConversionUtil.java
Scottmitch: Agreed with @ejona86 ... I would prefer to avoid trimming unless an argument can be made that the RFC requires it.

Scottmitch: Maybe we should trim just to be safe:

https://tools.ietf.org/html/rfc7230#section-3.2
> header-field   = field-name ":" OWS field-value OWS

https://tools.ietf.org/html/rfc7230#section-3.2.3
>  OWS            = *( SP / HTAB )
                    ; optional whitespace

codec-http2/src/main/java/io/netty/handler/codec/http2/HttpConversionUtil.java
Scottmitch: Lets trim just to be safe .. https://github.com/netty/netty/pull/7422/files#r152127214

codec-http2/src/main/java/io/netty/handler/codec/http2/HttpConversionUtil.java
Scottmitch: done

normanmaurer:
Cherry-picked into 4.1 (7d213240ca768d6dd35ef2336b1fda757bd4df3c)

PR #7424, opened 2017-11-20
Reduce conditionals in DnsNameResovlerContext

Scottmitch:
Motivation:
Minor cleanup from 844d804 just to reduce the conditional statements and indentation level.

Modifications:
- combine the else + if into an else if statement

Result:
Code cleaned up.


Scottmitch:
4.1 (907ed79069)

PR #7425, opened 2017-11-21
Not directly call getsockopt but use exported helper function

normanmaurer:
Motivation:

To better isolate OS system calls we should not call getsockopt directly but use our netty_unix_socket_getOption0 function. See is a followup of f115bf5.

Modifications:

Export netty_unix_socket_getOption0 by declaring it in the header file and use it

Result:

Better isolation of system calls.

### DIFF
--- a/transport-native-epoll/src/main/c/netty_epoll_linuxsocket.c
+++ b/transport-native-epoll/src/main/c/netty_epoll_linuxsocket.c
@@ -256,9 +256,8 @@ static jint netty_epoll_linuxsocket_isTcpQuickAck(JNIEnv* env, jclass clazz, jin
 
 static jint netty_epoll_linuxsocket_isTcpFastOpenConnect(JNIEnv* env, jclass clazz, jint fd) {
     int optval;
-    int optlen = sizeof(optval);
-    // We call getsockopt directly so we can handle ENOPROTOOPT by ourself.
-    if (getsockopt(fd, IPPROTO_TCP, TCP_FASTOPEN_CONNECT, &optval, &optlen) == -1) {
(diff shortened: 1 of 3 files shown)


transport-native-epoll/src/main/c/netty_epoll_linuxsocket.c
```diff
@@ -256,9 +256,8 @@ static jint netty_epoll_linuxsocket_isTcpQuickAck(JNIEnv* env, jclass clazz, jin
 
 static jint netty_epoll_linuxsocket_isTcpFastOpenConnect(JNIEnv* env, jclass clazz, jint fd) {
     int optval;
-    int optlen = sizeof(optval);
-    // We call getsockopt directly so we can handle ENOPROTOOPT by ourself.
-    if (getsockopt(fd, IPPROTO_TCP, TCP_FASTOPEN_CONNECT, &optval, &optlen) == -1) {
+    // We call netty_unix_socket_getOption0 directly so we can handle ENOPROTOOPT by ourself.
+    if (netty_unix_socket_getOption0(fd, IPPROTO_TCP, TCP_FASTOPEN_CONNECT, &optval, sizeof(optval)) == -1) {
```
carl-mastrangelo: should this happen in kqueue too?

normanmaurer: nope its linux only (TCP_FASTOPEN_CONNECT)

normanmaurer:
Cherry-picked into 4.1 (b1cc5835acff2df9e5e5b06b70e88b7e9d3c9490), 4.0 does not have the issue

PR #7426, opened 2017-11-21
Add RetryingInetNameResolver which can wrap a DnsNameResolver and implements retries

normanmaurer:
Add RetryingInetNameResolver which can wrap a DnsNameResolver and implements retries

Motivation:

If a resolution fails because of a timeout or IO error its often useful to retry it as the Nameserver may just be overloaded or empower some kind of rate limiting.

Modifications:

- Add RetryingInetNameResolver which can be used to handle retries
- Add unit tests

Result:

More robust resolver handling.

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/RetryingInetNameResolver.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/RetryingInetNameResolver.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.resolver.dns;
+
+import io.netty.resolver.InetNameResolver;
+import io.netty.util.concurrent.Future;
+import io.netty.util.concurrent.FutureListener;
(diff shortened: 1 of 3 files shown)


normanmaurer:
This depends on https://github.com/netty/netty/pull/7411... Please only review 
e2d51bc

normanmaurer:
@vietj @slandelle PTAL

normanmaurer:
ok this one now only depend on 4.1

normanmaurer:
@vietj @slandelle PTAL

resolver-dns/src/main/java/io/netty/resolver/dns/RetryingInetNameResolver.java
```diff
@@ -0,0 +1,96 @@
...
+        this.resolver = resolver;
+        this.retries = ObjectUtil.checkPositiveOrZero(retries, "retries");
+    }
+
+    @Override
+    protected void doResolve(final String inetHost, final Promise<InetAddress> promise) throws Exception {
+        resolver.resolve(inetHost).addListener(new RetryFutureListener<InetAddress>(retries, inetHost, promise) {
+            @Override
+            protected Future<InetAddress> resolveAgain(String inetHost) {
+                return resolver.resolve(inetHost);
```
NiteshKant: Should this also provide a way to backoff retries? A bad configuration here may result in a retry storm.

normanmaurer: @NiteshKant I think we should be ok without...

normanmaurer:
@slandelle @vietj any comments ?

normanmaurer:
@trustin @vietj @Scottmitch thoughts ?

slandelle:
@normanmaurer on sick leave... will have a look as soon as I can...

normanmaurer:
@slandelle @trustin ping

resolver-dns/src/main/java/io/netty/resolver/dns/RetryingInetNameResolver.java
trustin: Why is it OK without backoff?

normanmaurer:
lets close this one

PR #7427, opened 2017-11-21
Implement HttpHeaders#toString

slandelle:
Motivation:

For debugging/logging purpose, it would be convenient to have
HttpHeaders#toString implemented.

Modification:

Override HttpHeaders#toString to print the headers.

Result:

Convenient HttpHeaders#toString

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaders.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaders.java
@@ -19,6 +19,7 @@ import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufUtil;
 import io.netty.handler.codec.DateFormatter;
 import io.netty.handler.codec.Headers;
+import io.netty.handler.codec.HeadersUtils;
 import io.netty.util.AsciiString;
 import io.netty.util.CharsetUtil;
 
@@ -1686,5 +1687,10 @@ public abstract class HttpHeaders implements Iterable<Map.Entry<String, String>>
      */
     public boolean contains(CharSequence name, CharSequence value, boolean ignoreCase) {
         return contains(name.toString(), value.toString(), ignoreCase);
+    }
+
+    @Override
+    public String toString() {
+        return HeadersUtils.toString(getClass(), iteratorCharSequence(), size());
     }
 }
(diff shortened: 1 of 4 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaders.java
```diff
@@ -1687,4 +1687,20 @@ public final String getAsString(CharSequence name) {
     public boolean contains(CharSequence name, CharSequence value, boolean ignoreCase) {
         return contains(name.toString(), value.toString(), ignoreCase);
     }
+
+    @Override
+    public String toString() {
```
normanmaurer: @slandelle did you consider overriding this in `DefaultHttpHeaders` and have special implementation that not needs to allocate an extra List etc ?

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaders.java
```diff
@@ -1687,4 +1687,20 @@ public final String getAsString(CharSequence name) {
...
+        StringBuilder sb = new StringBuilder("HttpHeaders(");
```
normanmaurer: I think we should better use some "sane" initialCapacity here and then just call `append`.

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaders.java
```diff
@@ -1687,4 +1687,20 @@ public final String getAsString(CharSequence name) {
...
+
+        List<Entry<String, String>> entries = entries();
+        if (!entries.isEmpty()) {
+            String separator = ", ";
```
normanmaurer: Store this outside of the methods ?

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaders.java
slandelle: My intent was solely for debugging, so I went with the most straightforward implementation. Will change to use the iterator.

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaders.java
slandelle: ok

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaders.java
slandelle: ok

slandelle:
@normanmaurer PTAL again

slandelle:
@normanmaurer Please wait a few minutes, I'll make more consistent with DefaultHeaders.

slandelle:
@normanmaurer PTAL again. I added a helper in HeadersUtil to share toString implementation between DefaultHeaders and HttpHeaders.

normanmaurer:
Cherry-picked into 4.1 (e420f857fa0b9ab1df37359f7c0108475696b8dd)

Issue #7428, opened 2017-11-21
Usage of DefaultHttpHeader.names() outcome may result in java.lang.ArrayStoreException


mikkokar:
`DefaultHttpHeader.names()` exposes HTTP header names as a `Set<String>`. Converting the resulting set to an array using `toArray(String[])` throws an exception: `java.lang.ArrayStoreException: io.netty.util.AsciiString`. 

### Expected behavior

I expect following code to return an array of String objects:

```
        Set<String> nettyHeaders = new DefaultHttpHeaders(true)
                .add(HttpHeaderNames.CONTENT_LENGTH, "50")
                .names();
        
        String[] namesArray = nettyHeaders.toArray(new String[nettyHeaders.size()]);
```

### Actual behavior

But the code above throws an exception:

```
java.lang.ArrayStoreException: io.netty.util.AsciiString

	at io.netty.handler.codec.HeadersUtils$DelegatingStringSet.fillArray(HeadersUtils.java:234)
	at io.netty.handler.codec.HeadersUtils$DelegatingStringSet.toArray(HeadersUtils.java:227)
	at io.netty.handler.codec.http.DefaultHttpHeadersTest.providesHeaderNamesAsArray(DefaultHttpHeadersTest.java:222)        ...
```
### Steps to reproduce

See below.

### Minimal yet complete reproducer code (or URL to code)

I have constructed this test to demonstrate the issue:

```
    @Test
    public void providesHeaderNamesAsArray() throws Exception {
        Set<String> nettyHeaders = new DefaultHttpHeaders(true)
                .add(HttpHeaderNames.CONTENT_LENGTH, "50")
                .names();

        String[] namesArray = nettyHeaders.toArray(new String[nettyHeaders.size()]);
        assertArrayEquals(namesArray, new String[]{"content-length"});
    }
```

### Root Cause

The problem lies in the DelegatingStringSet class in HeaderUtils.java, which adapts the underlying `CharSequence` to a `String`. Well, the adaptation is implemented for the `iterator()` only, but not for `toArray()` method. 

### Netty version

4.1.17 and latest Netty 4.1 branch.

### JVM version (e.g. `java -version`)

jdk 1.8.0_131

normanmaurer:
@mikkokar thanks! Will check

normanmaurer:
@mikkokar PTAL https://github.com/netty/netty/pull/7429

normanmaurer:
Fixed via https://github.com/netty/netty/pull/7429

PR #7429, opened 2017-11-21
DefaultHttpHeader.names().toArray(...) may throw ArrayStoreException

normanmaurer:
Motivation:

DefaultHttpHeader.names() exposes HTTP header names as a Set<String>. Converting the resulting set to an array using toArray(String[]) throws an exception: java.lang.ArrayStoreException: io.netty.util.AsciiString.

Modifications:

- Remove our custom implementation of toArray(...) (and others) by just extending AbstractCollection.
- Add unit test

Result:

Fixes [#7428].

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/HeadersUtils.java
+++ b/codec/src/main/java/io/netty/handler/codec/HeadersUtils.java
@@ -15,6 +15,7 @@
  */
 package io.netty.handler.codec;
 
+import java.util.AbstractCollection;
 import java.util.AbstractList;
 import java.util.Collection;
 import java.util.Iterator;
@@ -84,7 +85,7 @@ public final class HeadersUtils {
     private static final class StringEntryIterator implements Iterator<Entry<String, String>> {
         private final Iterator<Entry<CharSequence, CharSequence>> iter;
 
-        public StringEntryIterator(Iterator<Entry<CharSequence, CharSequence>> iter) {
+        StringEntryIterator(Iterator<Entry<CharSequence, CharSequence>> iter) {
             this.iter = iter;
         }
 
@@ -145,7 +146,7 @@ public final class HeadersUtils {
     private static final class StringIterator<T> implements Iterator<String> {
         private final Iterator<T> iter;
 
-        public StringIterator(Iterator<T> iter) {
+        StringIterator(Iterator<T> iter) {
             this.iter = iter;
         }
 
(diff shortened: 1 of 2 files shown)


normanmaurer:
@mikkokar PTAL as well

mikkokar:
Thank you @normanmaurer for fixing this. That's what I also did to `fillArray` locally.

codec/src/main/java/io/netty/handler/codec/HeadersUtils.java
```diff
@@ -231,7 +231,9 @@ public boolean contains(Object o) {
         private void fillArray(Object[] arr) {
             Iterator<T> itr = allNames.iterator();
             for (int i = 0; i < size(); ++i) {
```
carl-mastrangelo: i think the iterator says you need to call hasNext()?

normanmaurer: @carl-mastrangelo yep this would be more correct... let me do this.

codec/src/main/java/io/netty/handler/codec/HeadersUtils.java
```diff
@@ -230,8 +230,10 @@ public boolean contains(Object o) {
 
         private void fillArray(Object[] arr) {
```
carl-mastrangelo: Why is this not passed in as a String[] ?

Scottmitch: Good point. I guess its because the signature of `public <X> X[] toArray(X[] a)` is generic and it is still shared between both `toArray` methods. However we should be able to avoid forcing the `toString` conversion in the case of the `public Object[] toArray()` method. Also it the creation of a `Object[]` array here may not be correct ... we should consider doing something similar to `AbstractCollection`:

```java
T[] r = a.length >= size ? a :
                  (T[])java.lang.reflect.Array
                  .newInstance(a.getClass().getComponentType(), size);
```

normanmaurer: @Scottmitch I disagree about not doing the string conversion. I think we must do it even if `Object[]` or `CharSequence[]` is passed in as we define it as `Set<String>` which should give the guarantee that only `String` will be stored in there. I will change how we create the Array tho.

normanmaurer:
@Scottmitch @carl-mastrangelo please note that I found an "easier" fix which also takes care of correctly creating the array and remove a lot of code. I just extended `AbstractCollection` and so we get all for free ;) PTAL again

normanmaurer:
Cherry-picked into 4.1 (7aca99f9869634d1f65e017e62b5520cd8df8cf5). 4.0 is not affected

carl-mastrangelo:
@normanmaurer +1

Issue #6548, opened 2017-03-16
Unable to initialize PlatformDependent0 (Netty 4.1.8.Final)


shumy:
### Expected behavior
No errors

### Actual behavior
java.lang.NoClassDefFoundError: Could not initialize class io.netty.util.internal.PlatformDependent0
io.netty.common[io.netty.channel.DefaultChannelId] : Could not invoke Process.myPid(); not Android?
java.lang.ClassNotFoundException: android/os/Process

### Steps to reproduce
On Ubuntu 16:
git clone https://github.com/shumy/leffe-wg
cd leffe-wg
gradle
./run.sh

### Minimal yet complete reproducer code (or URL to code)
https://github.com/shumy/leffe-wg

### Netty version
4.1.8.Final

### JVM version (e.g. `java -version`)
Java 8 on Apache Felix OSGi 

### OS version (e.g. `uname -a`)
Linux micael 4.4.0-66-generic #87-Ubuntu SMP Fri Mar 3 15:29:05 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

### OSGi bundles
    0|Active     |    0|System Bundle (5.4.0)|5.4.0
    1|Active     |    1|Guava: Google Core Libraries for Java (18.0.0)|18.0.0
    2|Active     |    1|Jackson-annotations (2.7.0)|2.7.0
    3|Active     |    1|Jackson-core (2.7.4)|2.7.4
    4|Active     |    1|jackson-databind (2.7.4)|2.7.4
    5|Active     |    1|com.github.shumy.leffe-wg (1.0.0)|1.0.0
    6|Active     |    1|Netty/Buffer (4.1.8.Final)|4.1.8.Final
    7|Active     |    1|Netty/Codec (4.1.8.Final)|4.1.8.Final
    8|Active     |    1|Netty/Codec/DNS (4.1.8.Final)|4.1.8.Final
    9|Active     |    1|Netty/Codec/HTTP (4.1.8.Final)|4.1.8.Final
   10|Active     |    1|Netty/Codec/HTTP2 (4.1.8.Final)|4.1.8.Final
   11|Active     |    1|Netty/Codec/Socks (4.1.8.Final)|4.1.8.Final
   12|Active     |    1|Netty/Common (4.1.8.Final)|4.1.8.Final
   13|Active     |    1|Netty/Handler (4.1.8.Final)|4.1.8.Final
[...]

PR #5640, opened 2016-08-05
Mark initialization of unsafe as privileged

jasontedor:
Motiviation:

Preparing platform dependent code for using unsafe requires executing
privileged code. The privileged code for initializing unsafe is executed
in a manner that would require all code leading up to the initialization
to have the requisite permissions. Yet, in a restrictive environment
(e.g., under a security policy that only grants the requisite
permissions the Netty common jar but not to application code triggering
the Netty initialization), then initializing unsafe will not succeed
even if the security policy would otherwise permit it.

Modifications:

This commit marks the necessary blocks as privileged. This enables
access to the necessary resources for initialization unsafe. The idea is
that we are saying the Netty code is trusted, and as long as the Netty
code has been granted the necessary permissions, then we will allow the
caller access to these resources even though the caller itself might not
have the requisite permissions.

Result:

Unsafe can be initialized in a restrictive security environment.

PR #7432, opened 2017-11-22
Guard against NoClassDefFoundError when trying to load Unsafe.

normanmaurer:
Motivation:

OSGI and other enviroments may not allow to even load Unsafe which will lead to an NoClassDefFoundError when trying to access it. We should guard against this.

Modifications:

Catch NoClassDefFoundError when trying to load Unsafe.

Result:

Be able to use netty with a strict OSGI config.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent0.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent0.java
@@ -96,6 +96,10 @@ final class PlatformDependent0 {
                         return e;
                     } catch (IllegalAccessException e) {
                         return e;
+                    } catch (NoClassDefFoundError e) {
+                        // Also catch NoClassDefFoundError in case someone uses for example OSGI and it made
+                        // Unsafe unloadable.
+                        return e;
                     }
                 }
             });


common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -96,6 +96,10 @@ public Object run() {
                         return e;
                     } catch (IllegalAccessException e) {
                         return e;
+                    } catch (NoClassDefFoundError e) {
```
johnou: I think LinkageError might be a little more robust here, would love to hear what @jasontedor has to say.

normanmaurer: Yeah maybe that would be a good idea!

jasontedor: Thanks for the ping @johnou. I think the most specific error possible should be caught. If this throws a `NoClassDefFoundError` in an OSGI situation, that I think that's what should be caught here only.

jasontedor: Also, before merging, I would recommend thinking about my [comment](https://github.com/netty/netty/issues/6548#issuecomment-346465502) on the original issue. It's hard to support something that you do not have experience with, tests for, etc. so maybe the best path is not to make a code change here at all (because a future refactoring might "break" it again). Instead, there's an out here for users in OSGI already: either add the dependency needed, or set a flag to skip trying to use unsafe.

johnou: There is actually a Netty OSGI test suite 💯

normanmaurer: yep ... I will try to add a test.. not idea if I will be successful tho =P

johnou: @normanmaurer not as easy as trying to load the class with the PaxExam runner is it? :cheese:

jasontedor: Apparently it does not do much of anything as it's not catching an issue for OSGI environments that do not add the necessary dependency for unsafe that has been there for over a year (#5640). And I question having that with two of the main maintainers saying that they do not have OSGI knowledge to maintain support for it. And none of this negates my point that there's an out for OSGI already without a change needed in Netty.

johnou: @jasontedor I would still argue that this specific exception is not 100% related to OSGI and that Netty should be able to detect if unsafe is available or not and if it is not, fallback.

jasontedor: Okay, I agree with you.

normanmaurer:
Unfortunately I was not able to write a test with paxexam (maybe because I don't know enough about it). That said I have read multiple sources in the internet that state that the `NoClassDefFoundError` is what is thrown by OSGi containers in this case. So let me merge this one and hopefully we can find someone in the future who will add a unit test .

normanmaurer:
Cherry-picked into 4.1 (65cacc9b15fc88f221ca4c3dde4c301d32864c08) and 4.0 (7ceea2223634da6dd73dde2a96acf06e8a4401dd)

Issue #7260, opened 2017-09-28
Slice support for FileRegion in JNI transports


Scottmitch:
Recently `slice` support was added to file region `FileRegion` for `NIO` and `OIO` https://github.com/netty/netty/pull/6997. We should do the same for `KQueue` and `EPOLL`. There are some challenges related to how `FileRegion` and `FileChannel` are handled in these JNI transports. There are assumptions made about the type of `FileRegion` made in order to get the integer file descriptor [1][2]. There are also assumptions made about `DefaultFileRegion` and its internals in JNI code [3][4]. There is some support for types other than `DefaultFileRegion` via `EpollSocketWritableByteChannel` [5], but this doesn't take advantage of `sendFile`. We should investigate alternative approaches which may help unify the two approaches in most cases and remove assumptions of types in JNI. For example: we should investigate moving extracting the file's FD (and the reflection involved) up to the java layer, creating a new interface for extensibility to give access to the file's FD, and pass the file's FD (and other necessary arguments) directly to JNI.

[1] https://github.com/netty/netty/blob/4.1/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollStreamChannel.java#L471 
[2] https://github.com/netty/netty/blob/4.1/transport-native-kqueue/src/main/java/io/netty/channel/kqueue/AbstractKQueueStreamChannel.java#L307
[3] https://github.com/netty/netty/blob/4.1/transport-native-epoll/src/main/c/netty_epoll_native.c#L282
[4] https://github.com/netty/netty/blob/4.1/transport-native-kqueue/src/main/c/netty_kqueue_bsdsocket.c#L46
[5] https://github.com/netty/netty/blob/4.1/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollStreamChannel.java#L401

Scottmitch:
@Apache9 @normanmaurer - FYI

Apache9:
Thanks @Scottmitch , nice writing up.

I think for FileChannelBasedFileRegion, we an try to get the FileChannel and then the FD, and then pass it to JNI.

PR #7435, opened 2017-11-23
Add FileRegionStream which can be used to provide a stream of FileReg…

normanmaurer:
…ion backed by some resource.

Motivation:

To support difference use-cases (like for example HTTP/2) we need the ability to dynamicly create FileRegion instances from a FileChannel in specific sizes.

Modifications:

- Add FileRegionStream and a DefaultFileRegionStream implementation which allows to create DefaultFileRegion instances using the same FileChannel instance.
- Allow to create DefaultFileRegion instances from a FileRegion and allow the user to take responsibility to close the FileChannel. This can be useful for users to write the same FileChannel multiple times without the overhead of calling
 open(...) all the times. This is important as open(...) may block and is also expensive.

Result:

Be able to create a stream of FileRegions.

### DIFF
--- a/transport/src/main/java/io/netty/channel/DefaultFileRegionStream.java
+++ b/transport/src/main/java/io/netty/channel/DefaultFileRegionStream.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.channel;
+
+import io.netty.util.AbstractReferenceCounted;
+import io.netty.util.IllegalReferenceCountException;
+import io.netty.util.internal.UnstableApi;
+import io.netty.util.internal.logging.InternalLogger;
+import io.netty.util.internal.logging.InternalLoggerFactory;
+
+import java.io.IOException;
+import java.nio.channels.FileChannel;
+
+import static io.netty.util.internal.ObjectUtil.*;
+
+/**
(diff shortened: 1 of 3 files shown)


normanmaurer:
This is an alternative way of implement what we need for HTTP/2 and so could close https://github.com/netty/netty/pull/6997.

@carl-mastrangelo @ejona86 @Scottmitch @Apache9  I think I would prefer to do something like this to minimise the scope of changes. I think this is good enough.

Apache9:
So we are going to expose this in the HTTP/2 API? Overall LGTM. This is only a POC so I'm not going to point out some corner cases as I think you will fix them in the next patch. :)

normanmaurer:
yep... I want to get the opinion of the others on this approach before spending more time on it.

transport/src/main/java/io/netty/channel/DefaultFileRegionStream.java
```diff
@@ -0,0 +1,106 @@
...
+     * Create a new instance
+     *
+     * @param file              the {@link FileChannel} which should be transfered
+     * @param position          the position from which the transfer should start
+     * @param count             the number of bytes to transfer
+     */
+    public DefaultFileRegionStream(FileChannel file, long position, long count) {
+        this.file = checkNotNull(file, "file");
+        this.position = checkPositiveOrZero(position, "position");
+        endPosition = position + checkPositiveOrZero(count, "count");
```
carl-mastrangelo: nit: this could be negative

transport/src/main/java/io/netty/channel/DefaultFileRegionStream.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    @Override
+    public FileRegion read(long bytes) {
+        if (refCnt() == 0) {
+            throw new IllegalReferenceCountException(0);
+        }
+
+        if (position + bytes > endPosition) {
```
carl-mastrangelo: same here about negative

Scottmitch:
@normanmaurer - the approach in PR https://github.com/netty/netty/pull/6997 seems consistent with the pre-existing approach of taking subsections of content vs consuming subsections of content, reference counting not being conditional, and wrapping from `ByteBuf`, `ByteBuffer`, etc... The approach in PR #6997 also seems more likely what I would expect the behavior to be if we could break APIs on `FIleRegion` itself. Other than the change-set being smaller what are the advantages of this PR vs https://github.com/netty/netty/pull/6997?

Scottmitch:
also related to https://github.com/netty/netty/issues/7260

Apache9:
So what's the final solution here?

PR #7438, opened 2017-11-24
Add tests for HttpObjectDecoder related to limits

normanmaurer:
Motivation:

HttpObjectDecoder will throw a TooLongFrameException when either the max size for the initial line or the header size was exceeed. We have no tests for this.

Modifications:

Add test cases.

Result:

More tests.


normanmaurer:
Cherry-picked into 4.1 (9ef83234f36441cd63f56e1c1a5f796e5e253b8b) and 4.0 (ae4a272e46)

Issue #7383, opened 2017-11-08
AbstractChannel#filterOutboundMessage may mask writing a released empty buffer


Scottmitch:
AbstractChannel attempts to "filter" messages which are written [1]. A goal of this process is to copy from heap to direct if necessary. However implementations of this method [2][3] may translate a buffer with 0 readable bytes to EMPTY_BUFFER. This may mask a user error where an empty buffer is written but already released.

[1] https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/AbstractChannel.java#L877
[2] https://github.com/netty/netty/blob/4.1/transport-native-kqueue/src/main/java/io/netty/channel/kqueue/AbstractKQueueChannel.java#L252
[3] https://github.com/netty/netty/blob/4.1/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollChannel.java#L306

PR #7440, opened 2017-11-25
Not use safeRelease(...) but release(...) to release non-readable hol…

normanmaurer:
…ders to ensure we not mask errors.

Motivation:

AbstractChannel attempts to "filter" messages which are written [1]. A goal of this process is to copy from heap to direct if necessary. However implementations of this method [2][3] may translate a buffer with 0 readable bytes to EMPTY_BUFFER. This may mask a user error where an empty buffer is written but already released.

Modifications:

Replace safeRelease(...) with release(...) to ensure we propagate reference count issues.

Result:

Fixes [#7383]


normanmaurer:
@Scottmitch PTAL

normanmaurer:
Cherry-picked into 4.1 (251bb1a73961f785ad12dbb3ae92830eca70b8cd) and 4.0 (ea8ff3b4f)

PR #7441, opened 2017-11-26
The constants calculation in compile-time

fenik17:
Motivation:
Allow pre-computing calculation of the constants for compiler where it could be. Similar fix in OpenJDK: [1].

Modifications:
- Use parentheses.
- Simplify static initialization of `BYTE2HEX_*` arrays in `StringUtil`.

Result:
Less bytecode, possible faster calculations at runtime.

[1] https://bugs.openjdk.java.net/browse/JDK-4477961

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/compression/FastLz.java
+++ b/codec/src/main/java/io/netty/handler/codec/compression/FastLz.java
@@ -168,7 +168,7 @@ final class FastLz {
                         readU16(input, inOffset + ip - 1) == readU16(input, inOffset + ip + 1)) {
                     distance = 1;
                     ip += 3;
-                    ref = anchor - 1 + 3;
+                    ref = anchor + (3 - 1);
 
                     /*
                      * goto match;
--- a/common/src/main/java/io/netty/util/internal/StringUtil.java
+++ b/common/src/main/java/io/netty/util/internal/StringUtil.java
@@ -48,19 +48,9 @@ public final class StringUtil {
 
     static {
         // Generate the lookup table that converts a byte into a 2-digit hexadecimal integer.
-        int i;
-        for (i = 0; i < 10; i++) {
-            BYTE2HEX_PAD[i] = "0" + i;
-            BYTE2HEX_NOPAD[i] = String.valueOf(i);
-        }
-        for (; i < 16; i++) {
-            char c = (char) ('a' + i - 10);
-            BYTE2HEX_PAD[i] = "0" + c;
-            BYTE2HEX_NOPAD[i] = String.valueOf(c);
-        }
-        for (; i < BYTE2HEX_PAD.length; i++) {
+        for (int i = 0; i < BYTE2HEX_PAD.length; i++) {
             String str = Integer.toHexString(i);
-            BYTE2HEX_PAD[i] = str;
+            BYTE2HEX_PAD[i] = i > 0xf ? str : ('0' + str);
             BYTE2HEX_NOPAD[i] = str;
         }
     }
@@ -226,10 +216,10 @@ public final class StringUtil {
             return c - '0';
         }
         if (c >= 'A' && c <= 'F') {
-            return c - 'A' + 0xA;
+            return c - ('A' - 0xA);
         }
         if (c >= 'a' && c <= 'f') {
-            return c - 'a' + 0xA;
+            return c - ('a' - 0xA);
         }
         return -1;
     }


fenik17:
code:
```java 
return c - ('A' - 0xA);
```

bytecode:
```
L4
    LINENUMBER 219 L4
    ILOAD 0
    BIPUSH 55
    ISUB
    IRETURN
```

---------
code:
```java
return c - 'A' + 0xA;
```

bytecode:
```
L4
    LINENUMBER 219 L4
    ILOAD 0
    BIPUSH 65
    ISUB
    BIPUSH 10
    IADD
    IRETURN
```

common/src/main/java/io/netty/util/internal/StringUtil.java
```diff
@@ -48,19 +48,9 @@
...
-            BYTE2HEX_PAD[i] = "0" + i;
-            BYTE2HEX_NOPAD[i] = String.valueOf(i);
-        }
-        for (; i < 16; i++) {
-            char c = (char) ('a' + i - 10);
-            BYTE2HEX_PAD[i] = "0" + c;
-            BYTE2HEX_NOPAD[i] = String.valueOf(c);
-        }
-        for (; i < BYTE2HEX_PAD.length; i++) {
+        for (int i = 0; i < BYTE2HEX_PAD.length; i++) {
```
fenik17: Simplification in this code block is done instead of a just correction `('a' + i - 10) -> ('a' + (i - 10))` which relates for current PR.

normanmaurer: @fenik17 got it... that said I think it not really matters here as its in a static block.

normanmaurer:
@fenik17 TIL! This is super interesting and a good catch! Thanks again for the contribution.

adriancole:
how much longer until there's a unit test: bytecode representing java < N ops?

handler/src/main/java/io/netty/handler/logging/LoggingHandler.java
```diff
@@ -347,13 +348,13 @@ private static String formatByteBufHolder(ChannelHandlerContext ctx, String even
...
         if (length == 0) {
-            StringBuilder buf = new StringBuilder(chStr.length() + 1 + eventName.length() + 2 + msgStr.length() + 4);
+            StringBuilder buf = new StringBuilder(chStr.length() + eventName.length() + msgStr.length() + (1 + 2 + 4));
             buf.append(chStr).append(' ').append(eventName).append(", ").append(msgStr).append(", 0B");
             return buf.toString();
         } else {
             int rows = length / 16 + (length % 15 == 0? 0 : 1) + 4;
             StringBuilder buf = new StringBuilder(
-                    chStr.length() + 1 + eventName.length() + 2 + msgStr.length() + 2 + 10 + 1 + 2 + rows * 80);
+                    chStr.length() + eventName.length() + msgStr.length() + (1 + 2 + 2 + 10 + 1 + 2) + rows * 80);
```
jchambers: Moving these constants away from the things they're modifying makes it slightly less clear what's going on (acknowledging that's already true for `2 + 10 + 1 + 2`). I wonder if, given the number of small constants, it would make more sense to just smoosh everything down into a single constant (18), but add a comment explaining where it came from.

jasontedor: I agree with @jchambers, yet I feel even more strongly about the loss of clarity.

I think it should be left as-is. I doubt this code is a performance bottleneck for anything (since it's inside a debugging handler), and I doubt these calculations are even a performance bottleneck for this handler in general. Rearranging these constants or collapsing leads to less clear code, and is the epitome of micro-optimizing at the expense of code quality. I know this quote is misused and misunderstood often, but here I think it's apt:

>Programmers waste enormous amounts of time thinking about, or worrying about, the speed of noncritical parts of their programs, and these attempts at efficiency actually have a strong negative impact when debugging and maintenance are considered.

Sorry @fenik17, I think you do wonderful work on the Netty codebase and I enjoy reading your PRs, but this particular change in the logging handler I think would be a mistake.

normanmaurer: @fenik17 I think I agree with @jchambers @jasontedor here ... clear code is more important here. So either I would do what @jchambers suggested or revert this part.

codec/src/main/java/io/netty/handler/codec/compression/FastLz.java
```diff
@@ -168,7 +168,7 @@ static int compress(final byte[] input, final int inOffset, final int inLength,
                         readU16(input, inOffset + ip - 1) == readU16(input, inOffset + ip + 1)) {
                     distance = 1;
                     ip += 3;
-                    ref = anchor - 1 + 3;
+                    ref = anchor + (3 - 1);
```
jchambers: This might also be an opportunity to just do `anchor + 2`, but explain in a comment where the `2` comes from.

handler/src/main/java/io/netty/handler/logging/LoggingHandler.java
fenik17: I agree. This is a bit ugly code. Reverted..
I know how to do better, and maybe come back here later :)

codec/src/main/java/io/netty/handler/codec/compression/FastLz.java
fenik17: I'm not sure how to formulate the correct comment here..

fenik17:
@adriancole 
> how much longer until there's a unit test: bytecode representing java < N ops?

Sorry, I didn't understand.. Can you clarify your question?

codec/src/main/java/io/netty/handler/codec/compression/FastLz.java
jchambers: I'll confess that I'm not sure, either.

normanmaurer:
@Scottmitch any preference ?

Scottmitch:
@normanmaurer - I agree that we should balance the optimizations vs clarity/understandability, and in general only change if we can demonstrate some practical value. However the current form of this PR seems relatively isolated and doesn't impact understandability IMHO.

normanmaurer:
Cherry-picked into 4.1 (c9668ce40f80f73763165a8bfe70969de830204c)

PR #7445, opened 2017-11-27
Dont use ThreadDeathWatcher to cleanup PoolThreadCache if FastThreadL…

normanmaurer:
…ocalThread with wrapped Runnable is used

Motivation:

We dont need to use the ThreadDeathWatcher if we use a FastThreadLocalThread for which we wrap the Runnable and ensure we call FastThreadLocal.removeAll() once the Runnable completes.

Modifications:

- Dont use a ThreadDeathWatcher if we are sure we will call FastThreadLocal.removeAll()
- Add unit test.

Result:

Less overhead / running theads if you only allocate / deallocate from FastThreadLocalThreads.

### DIFF
--- a/buffer/src/test/java/io/netty/buffer/PooledByteBufAllocatorTest.java
+++ b/buffer/src/test/java/io/netty/buffer/PooledByteBufAllocatorTest.java
@@ -240,19 +240,28 @@ public class PooledByteBufAllocatorTest extends AbstractByteBufAllocatorTest<Poo
         assertFalse(lists.get(5).iterator().hasNext());
     }
 
-    // The ThreadDeathWatcher sleeps 1s, give it double that time.
-    @Test (timeout = 2000)
-    public void testThreadCacheDestroyedByThreadDeathWatcher() {
+    @Test (timeout = 4000)
+    public void testThreadCacheDestroyedByThreadDeathWatcher() throws InterruptedException {
+        testThreadCacheDestroyedByThreadDeathWatcher(false);
+    }
+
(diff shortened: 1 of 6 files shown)


buffer/src/test/java/io/netty/buffer/PooledByteBufAllocatorTest.java
```diff
@@ -243,31 +243,51 @@ public void testFreePoolChunk() {
...
+
+                buf.release();
+            }
+        };
+
+        for (int i = 0; i < numArenas; i++) {
+            if (useRunnable) {
+                new FastThreadLocalThread(task).start();
+            } else {
+                new FastThreadLocalThread() {
```
ejona86: Missing `start()`? Does that mean another assertion should be added to notice?

normanmaurer: yep missed and yep let me add another assertion as well

normanmaurer: @ejona86 good catch!

normanmaurer: Added a `CountdownLatch` so the test would fail with a timeout if we not start these. @ejona86 PTAL again

normanmaurer:
Cherry-picked into 4.1 (09a05b680da6e44918b76599b78e8e0cb0c19109) and 4.0 (6507f237d14da298816f4822369e8e3a29cab8db)

PR #7446, opened 2017-11-27
Dont use the ThreadDeathWatcher to release memory back to PoolArena a…

normanmaurer:
…fter PoolThreadCache is not used anymore

Motivation:

At the moment we use the ThreadDeathWatcher to release memory back to the PoolArena after the PoolThreadCache is not used anymore if the cache is used by a non-FastThreadLocalThread. This will need to spawn on extra thread and iterate over all the Threads each second.

Modifications:

Better just use a finalizer to return the memory back to the PoolArena after the cache is not used anymore and the cache belonged to a non-FastThreadLocalThread.

Result:

Less running threads / overhead.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/PoolThreadCache.java
+++ b/buffer/src/main/java/io/netty/buffer/PoolThreadCache.java
@@ -66,7 +66,8 @@ final class PoolThreadCache {
 
     PoolThreadCache(PoolArena<byte[]> heapArena, PoolArena<ByteBuffer> directArena,
                     int tinyCacheSize, int smallCacheSize, int normalCacheSize,
-                    int maxCachedBufferCapacity, int freeSweepAllocationThreshold) {
+                    int maxCachedBufferCapacity, int freeSweepAllocationThreshold,
+                    boolean useThreadDeathWatcher) {
         if (maxCachedBufferCapacity < 0) {
             throw new IllegalArgumentException("maxCachedBufferCapacity: "
                     + maxCachedBufferCapacity + " (expected: >= 0)");
@@ -112,14 +113,16 @@ final class PoolThreadCache {
             numShiftsNormalHeap = -1;
         }
 
-        // We only need to watch the thread when any cache is used.
-        if (tinySubPageDirectCaches != null || smallSubPageDirectCaches != null || normalDirectCaches != null
-                || tinySubPageHeapCaches != null || smallSubPageHeapCaches != null || normalHeapCaches != null) {
-            // Only check freeSweepAllocationThreshold when there are caches in use.
-            if (freeSweepAllocationThreshold < 1) {
-                throw new IllegalArgumentException("freeSweepAllocationThreshold: "
-                        + freeSweepAllocationThreshold + " (expected: > 0)");
-            }
+        // Only check if there are caches in use.
+        if ((tinySubPageDirectCaches != null || smallSubPageDirectCaches != null || normalDirectCaches != null
+                || tinySubPageHeapCaches != null || smallSubPageHeapCaches != null || normalHeapCaches != null)
+                && freeSweepAllocationThreshold < 1) {
+            throw new IllegalArgumentException("freeSweepAllocationThreshold: "
+                    + freeSweepAllocationThreshold + " (expected: > 0)");
+        }
+
+        if (useThreadDeathWatcher) {
+
             freeTask = new Runnable() {
                 @Override
                 public void run() {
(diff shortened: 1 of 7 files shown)


normanmaurer:
PTAL... This depends on https://github.com/netty/netty/pull/7445.

ejona86:
Hmm... since finalize is deprecated in Java 9, do we want to do something else? We could keep the cleaner thread around and use `WeakReference`+`ReferenceQueue`. The thread would sit on `remove()`, although it will probably still need to do periodic wakeups. I don't know the code well enough to know if we could avoid the thread altogether.

normanmaurer:
@ejona86 hmm yeah we could have some kind of "ReaperThread" that will cleanup this stuff... Even this would be better then the `ThreadDeatchWatcher` imho... 

So I think a `finalize()` may be ok for now as we will need to do another PR anyway to get rid of all `finalize` overrides in all the classes. So for now I think this PR is at least an improvement.

WDYT ?

ejona86:
I've found adding threads within a library to be difficult. If we remove it now, then depending how much time passes I'd anticipate re-adding it to cause breakage.

Using `ReferenceQueue` is much closer to the current code than `finalize`, although it is likely true `ThreadDeathWatcher` is the wrong API. And just because there are (a few) other `finalize()` usages doesn't mean we should paint ourselves into a corner and make the problem worse.

Is there a reason to try hard to kill `ThreadDeathWatcher` now at the expense of dealing with the migration twice and any thread-creation fallout?

normanmaurer:
@ejona86 actually I would argue that if we would run on java9 only we should use the `Cleaner` for it which we not do atm: https://docs.oracle.com/javase/9/docs/api/java/lang/ref/Cleaner.html

WYTD ?

ejona86:
"on java9 only", I'd agree with that, but given support is currently Java 6, it's a pretty moot. Cleaner when on Java 9 sounds good, but we will need a fallback. If we chose `finalize` since it's easy, that could be fine. Note that Cleaner does create a thread (albeit, in a very specific manner).

Hmm... classloader leak... I guess the failure case is the ThreadLocal instance could be thrown away, and let the storage evaporate (since Thread uses WeakReference for the mapping), but the Thread itself remain running. So ThreadDeathWatcher continues running unawares, and in fact maybe even transitively retains the ThreadLocal because it retains the ClassLoader.

So it seems the problem is the false assumption that the ThreadLocal storage's life ends at the same time as the thread. I guess that's why Cleaner has "The context class loader of the thread is set to the system class loader." Aaaand... we couldn't do the same approach, because the Cleaner trick relies on the fact that Cleaner is in the system class loader.

Does my analysis feel right? If so, then it seems `finalize` is likely the only option, short of disabling caching on the other threads to begin with.

buffer/src/main/java/io/netty/buffer/PoolThreadCache.java
```diff
@@ -508,4 +486,16 @@ protected Entry newObject(Handle<Entry> handle) {
             }
         };
     }
+
+    // We create an instanceof this if we need to call free() as part of the GC of the cache.
```
ejona86: s/instanceof/instance of/

ejona86:
Wait... if my analysis is correct, then even `Cleaner` wouldn't work, since it maintains strong references to `Runnables` retains the `ClassLoader` (which retains `UnpaddedInternalThreadLocalMap.class` which has the static `ThreadLocal`). `Cleaner` would need the references to be non-static (or properly cleared). My head hurts a little thinking about it (mostly because of normal "there's still a reference, in the finalize"), but it does seem `finalize` would work. But it also seems we're backing ourselves into the corner.

I feel like something else needs to give. Either the caching needs to go, or an API to clear the cache, or some slow caching path, or something. For the slow caching path, maybe we could force all non-`FastThreadLocalThread`s to share a single (possibly-sharded) cache and just throw a lock on it, but we would still need to figure out a good time to clear it.

buffer/src/test/java/io/netty/buffer/PooledByteBufAllocatorTest.java
```diff
@@ -290,8 +290,9 @@ public void run() {
             }
         }
 
-        // Wait for the ThreadDeathWatcher to have destroyed all thread caches
+        // Wait for the GC to have destroyed all thread caches
         while (allocator.metric().numThreadLocalCaches() > 0) {
+            System.gc();
```
carl-mastrangelo: Also System.runFinalization()

buffer/src/main/java/io/netty/buffer/PoolThreadCache.java
```diff
@@ -508,4 +486,16 @@ protected Entry newObject(Handle<Entry> handle) {
...
+    private final class PoolThreadCacheFinalizer {
+        @Override
+        protected void finalize() throws Throwable {
+            try {
+                free();
```
carl-mastrangelo: is it safe to call free multiple times?

normanmaurer:
@carl-mastrangelo @ejona86 WDYT about the approach: https://github.com/netty/netty/pull/7463 ?

normanmaurer:
Let me close this.. as we should better use #7463

PR #7448, opened 2017-11-27
Provide a test for #6548 using the OSGi test suite

timothyjward:
The OSGi Test suite runs without access to sun.misc.Unsafe, and so is a good place to put a test to avoid regressing #6548.

The included test is a simple echo from a client to a server and back which covers a broad set of netty features.

Signed-off-by: Tim Ward <timothyjward@apache.org>

Motivation:

Bug #6548 appeared because it is not easy to test Netty's behaviour in the absence of sun.misc.Unsafe. This test uses OSGi to limit the loadabilty of sun.misc, and then uses Netty to send some data over the network

Modification:

An additional test has been added to the OSGi test suite to verify using Netty in the absence of Unsafe

Result:

Enhances the fix for #6548 by adding a test.

### DIFF
--- a/testsuite-osgi/src/test/java/io/netty/osgitests/OsgiBundleTest.java
+++ b/testsuite-osgi/src/test/java/io/netty/osgitests/OsgiBundleTest.java
@@ -16,28 +16,55 @@
 
 package io.netty.osgitests;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.ops4j.pax.exam.CoreOptions.frameworkProperty;
 import static org.ops4j.pax.exam.CoreOptions.junitBundles;
 import static org.ops4j.pax.exam.CoreOptions.mavenBundle;
 import static org.ops4j.pax.exam.CoreOptions.systemProperty;
 import static org.ops4j.pax.exam.CoreOptions.wrappedBundle;
+import static org.osgi.framework.Constants.FRAMEWORK_BOOTDELEGATION;
 
 import java.io.BufferedReader;
 import java.io.File;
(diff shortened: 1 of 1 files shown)


testsuite-osgi/src/test/java/io/netty/osgitests/OsgiBundleTest.java
```diff
@@ -104,4 +133,127 @@ public void testResolvedBundles() {
         // No-op, as we just want the bundles to be resolved. Just check if we tested something
         assertFalse("At least one bundle needs to be tested", BUNDLES.isEmpty());
     }
+
+    @Test
+    public void testSimpleEcho() throws InterruptedException {
+
+        assertFalse(PlatformDependent.hasUnsafe());
```
normanmaurer: Wouldn’t it be good enough to only do this? This should throw otherwise

timothyjward: That is a useful check - I just thought it might be a good idea to do a more comprehensive sniff test by actually firing up a Channel in OSGi. Feel free to modify!

normanmaurer: I would prefer to keep the test as simple as possible

normanmaurer:
This is the test for https://github.com/netty/netty/pull/7432.

normanmaurer:
@timothyjward simplified and cherry-picked into 4.1 (460d1251219e741e034efc083c7f8cd6f57c7238) and 4.0 (8d58e9a3d2).

Thanks again!

PR #7449, opened 2017-11-28
Throw FileNotFoundException when connecting to a missing UDS path

carl-mastrangelo:
Motivation:
Exception handling is nicer when a more specific Exception is thrown

Modification:
Add a static reference for ENOENT, and throw FNFE if it is returned

Result:
More precise exception handling


normanmaurer:
Cherry-picked into 4.1 (d8cb9ce09f5e591c244f2c3b04ea5bf054865ebb) and 4.0 (777b400fd6c6ef1f948c0f8da964565f804d5f22)

PR #7450, opened 2017-11-28
Fixed default OpenSsl cipher suites

SgtSilvio:
Motivation:

The default enabled cipher suites of the OpenSsl engine are not set to
SslUtils#DEFAULT_CIPHER_SUITES. Instead all available cipher suites are
enabled. This should happen only as a fallback.

Modifications:

Moved the line in the static initializer in OpenSsl which adds the
SslUtils#DEFAULT_CIPHER_SUITES to the default enabled cipher suites up
before the fallback.

Result:

The default enabled cipher suites of the OpenSsl engine are set to the
available ones of the SslUtils#DEFAULT_CIPHER_SUITES.
The default enabled cipher suites of the OpenSsl engine are only set to
all available cipher suites if no one of the
SslUtils#DEFAULT_CIPHER_SUITES is supported.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
+++ b/handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
@@ -189,9 +189,9 @@ public final class OpenSsl {
                 availableJavaCipherSuites.add(CipherSuiteConverter.toJava(cipher, "SSL"));
             }
 
+            addIfSupported(availableJavaCipherSuites, defaultCiphers, DEFAULT_CIPHER_SUITES);
             useFallbackCiphersIfDefaultIsEmpty(defaultCiphers, availableJavaCipherSuites);
             DEFAULT_CIPHERS = Collections.unmodifiableList(defaultCiphers);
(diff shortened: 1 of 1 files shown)


normanmaurer:
@SgtSilvio good catch!

Can you sign our ICLA and maybe also add a unit test:

https://netty.io/s/icla

johnou:
cc @rkapsi @slandelle @CodingFabian

rkapsi:
Nice catch, +1 for unit test.

normanmaurer:
Cherry-picked into 4.1(3c8f4b81d7c5dd7ac0c221585623ccf5346380cd) and 4.0 (6583b5271c)

normanmaurer:
Also added a test-case.

SgtSilvio:
Thank you for your quick response, sorry that I did not find time for writing the unit test myself.

normanmaurer:
@SgtSilvio no worries... thanks for the fix :)

Scottmitch:
oof ... good catch!

PR #7454, opened 2017-11-30
FIX: force a read operation for peer instead of self

louxiu:
Motivation:
When A is in `writeInProgress` and call self close, A should
`finishPeerRead` for B(A' peer).

Modifications:
Call `finishPeerRead` with peer in `LocalChannel#doClose`

Result:
Clear confuse of code logic

### DIFF
--- a/transport/src/test/java/io/netty/channel/local/LocalChannelTest.java
+++ b/transport/src/test/java/io/netty/channel/local/LocalChannelTest.java
@@ -391,6 +391,73 @@ public class LocalChannelTest {
     }
 
     @Test
+    public void testCloseAfterWriteInSameEventLoopPreservesOrder() throws InterruptedException {
+        Bootstrap cb = new Bootstrap();
+        ServerBootstrap sb = new ServerBootstrap();
+        final CountDownLatch messageLatch = new CountDownLatch(3);
+        final ByteBuf data = Unpooled.wrappedBuffer(new byte[1024]);
+
+        try {
+            cb.group(sharedGroup)
+                    .channel(LocalChannel.class)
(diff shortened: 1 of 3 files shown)


louxiu:
This commit doesn't fix any bug for LocalChannel. IMHO, the original code is just incorrect. 
@normanmaurer @Scottmitch  PTAL

louxiu:
Add another [commit](https://github.com/netty/netty/pull/7454/commits/8849a3ae8a2e2808742aa672ed9500e2ab332002) to solve an order problem in same event group.

normanmaurer:
@louxiu please check the test failure

louxiu:
@normanmaurer done.  This commit improves mismatch test in `SniClientTest `.

transport/src/test/java/io/netty/channel/local/LocalChannelTest.java
```diff
@@ -390,6 +391,67 @@ public void operationComplete(ChannelFuture future) throws Exception {
...
+                    .channel(LocalChannel.class)
+                    .handler(new ChannelInboundHandlerAdapter() {
+                        @Override
+                        public void channelActive(ChannelHandlerContext ctx) throws Exception {
+                            ctx.writeAndFlush(data.retainedDuplicate());
+                        }
+
+                        @Override
+                        public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+                            assertEquals(data, msg);
```
Scottmitch: consider surfacing this exception in the JUnit thread. One option is to set an AtomicReference<Throwable> and check it after the latch await returns.

louxiu:
@Scottmitch Done it another way. PTAL

PR #7457, opened 2017-11-30
EDNS Client Subnet is not encoded correctly when source prefix length is not a multiple of 8.

thomdev:
Motivation:
When the ECS source prefix length is not a mutiple of 8, the last byte the address inside the
ECS OPT record is not padded properly.
    
Modifications:
DefaultDnsRecordEncoder.padWithZeros(...) was modified to add padding from the least
significant bits.
    
Result:
ECS encoding bug fixed.


normanmaurer:
@thomdev thanks a lot Thomas :)

normanmaurer:
I will squash and merge once the CI completes.

normanmaurer:
@thomdev Thanks again...

I squashed and cherry-picked into 4.1 (32746c53c1089550f79bd2d6eeba4d65dab1f2d3)

PR #7460, opened 2017-12-01
Http2ConnectionHandler Http2ConnectionPrefaceAndSettingsFrameWrittenE…

Scottmitch:
…vent propagation

Motivation:
Http2ConnectionHandler uses ctx.fireUserEvent to propagate the Http2ConnectionPrefaceAndSettingsFrameWrittenEvent through the pipeline. This will propagate the event to the next inbound handler in the pipeline. If the user extends Http2ConnectionHandler the Http2ConnectionPrefaceAndSettingsFrameWrittenEvent may be missed and initialization dependent upon this event will not be run.

Modifications:
- Http2ConnectionHandler should use userEventTriggered instead of ctx.fireUserEvent

Result:
Classes that extend Http2ConnectionHandler will see the Http2ConnectionPrefaceAndSettingsFrameWrittenEvent user event.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
@@ -233,7 +233,7 @@ public class Http2ConnectionHandler extends ByteToMessageDecoder implements Http
         private ByteBuf clientPrefaceString;
         private boolean prefaceSent;
 
-        public PrefaceDecoder(ChannelHandlerContext ctx) {
+        public PrefaceDecoder(ChannelHandlerContext ctx) throws Exception {
             clientPrefaceString = clientPrefaceString(encoder.connection());
             // This handler was just added to the context. In case it was handled after
             // the connection became active, send the connection preface now.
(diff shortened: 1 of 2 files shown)


normanmaurer:
@Scottmitch can you also add a unit test ?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -375,8 +375,10 @@ private void sendPreface(ChannelHandlerContext ctx) {
                     ChannelFutureListener.CLOSE_ON_FAILURE);
 
             if (isClient) {
-                ctx.fireUserEventTriggered(
-                        Http2ConnectionPrefaceAndSettingsFrameWrittenEvent.INSTANCE);
+                // If this handled is extended by the user and we directly fire the userEvent from this context then
```
normanmaurer: nit: s/handled/handler/

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -375,8 +375,10 @@ private void sendPreface(ChannelHandlerContext ctx) {
...
+                // the user will not see the event. We should fire the event through the entire pipeline so it is seen
```
normanmaurer: also this comment is not correct as we not fire it through the entire pipeline ;) Please adjust

normanmaurer:
Cherry-picked into 4.1 (7cced5576f0800b5df2b2e166ff6b24925addec4)

normanmaurer:
@Scottmitch thanks!

Issue #7458, opened 2017-12-01
Epoll deadlock during classload


carl-mastrangelo:
### Netty version
4.1.17

This was a tricky one reported by one of our gRPC users.  There is a deadlock that can occur during initialization of netty Epoll code.     The core of the problem is that the Epoll class load JNI code, which then calls back into java code to load more classes.  I think it would be best if all JNI code in netty did NOT call back into Java to avoid these cycle problems.  

In the reported case there are two event loops that are trying to load.    The abbreviated interleaving that causes the lock is listed below, using T1 and T2 as the threads.

1.  T1 calls `Epoll.<clinit>`
2.  T1 calls `Runtime.loadLibrary0`, which is synchronized on the `Runtime` class.  The **`Runtime.class` lock is held.**
3.  T2 creates a new channel, resulting in `AbstractChannel.newId` calling `DefaultChannelId.newInstance`.  
4.  T2 calls `DefaultChannelId.<clinit>`, which calls `MacAddressUtil.<clinit>` which finally calls  ` java.net.NetworkInterface.<clinit>`.  
5.  T2 **Now hold the classloader lock for NetworkInterface**
6.  T2  `NetworkInterface.<clinit>`   attempts to call `System.loadLibrary` for library `net` which then blocks trying to acquire the `Runtime.class` lock
7.  T1 now continues, calling `netty_epoll_native_JNI_OnLoad`, followed by `netty_unix_socket_JNI_OnLoad`
8.  T1 tries to initialize `NetUtil` with `(*env)->FindClass(env, netty_unix_util_prepend(packagePrefix, "io/netty/util/NetUtil"));`
9.  T1 In `NetUtil.<clinit>` tries to initialize `NetworkInterface`, but waits to acquire the class init lock.

At this point both threads hold each others' locks, and progress ceases.

/cc @Scottmitch @normanmaurer @ejona86

normanmaurer:
@carl-mastrangelo I think we will need to at least call `FindClass` and there is nothing we can do about this... Are you sure this is because of `FindClass` and not because we call `NetUtil.isIpV4StackPreferred()` ?

normanmaurer:
@Scottmitch @carl-mastrangelo @ejona86 PTAL https://github.com/netty/netty/pull/7461 for a fix and some more details

Scottmitch:
[...]

PR #7461, opened 2017-12-01
Ensure we not run into classloading deadlocks while trying to load th…

normanmaurer:
…e native transports.

Motivation:

Modifications:

Use Boolean.getBoolean("java.net.preferIPv4Stack")  to detect if ipv4 should be prefered.

Result:

Fixes [#7458].

### DIFF
--- a/transport-native-unix-common/src/main/c/netty_unix_socket.c
+++ b/transport-native-unix-common/src/main/c/netty_unix_socket.c
@@ -40,8 +40,6 @@ static jclass datagramSocketAddressClass = NULL;
 static jmethodID datagramSocketAddrMethodId = NULL;
 static jmethodID inetSocketAddrMethodId = NULL;
 static jclass inetSocketAddressClass = NULL;
-static jclass netUtilClass = NULL;
-static jmethodID netUtilClassIpv4PreferredMethodId = NULL;
 static int socketType;
 static const char* ip4prefix = "::ffff:";
 static const unsigned char wildcardAddress[] = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 };
@@ -169,9 +167,7 @@ static jbyteArray createInetSocketAddressArray(JNIEnv* env, const struct sockadd
     return bArray;
 }
 
-static int socket_type(JNIEnv* env) {
-    jboolean ipv4Preferred = (*env)->CallStaticBooleanMethod(env, netUtilClass, netUtilClassIpv4PreferredMethodId);
-
+static int socket_type(JNIEnv* env, jboolean ipv4Preferred) {
     if (ipv4Preferred) {
         // User asked to use ipv4 explicitly.
         return AF_INET;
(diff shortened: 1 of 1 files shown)


normanmaurer:
@carl-mastrangelo PTAL... This should fix #7458. Its kind of the same as openjdk does as well:

http://code.metager.de/source/xref/openjdk/jdk8/jdk/src/share/native/java/net/net_util.c

Unfortunately there is no way to retrieve a system property without calling back into java. Dont get me started on this.

transport-native-unix-common/src/main/c/netty_unix_socket.c
```diff
@@ -1001,27 +997,24 @@ jint netty_unix_socket_JNI_OnLoad(JNIEnv* env, const char* packagePrefix) {
...
-        // position method was not found.. something is wrong so bail out
-        netty_unix_errors_throwRuntimeException(env, "failed to get method ID: NetUild.isIpV4StackPreferred()");
+
+    jmethodID booleanClassGetBooleanMethodId = (*env)->GetStaticMethodID(env, booleanClass, "getBoolean", "(Ljava/lang/String;)Z" );
+    if (booleanClassGetBooleanMethodId == NULL) {
+        netty_unix_errors_throwRuntimeException(env, "failed to get method ID: Boolean.getBoolean(String)");
         return JNI_ERR;
     }
 
+    jboolean preferIPv4Stack = (*env)->CallStaticBooleanMethod(env, booleanClass, booleanClassGetBooleanMethodId, "java.net.preferIPv4Stack");
```
Scottmitch: what happens if a security manager prevents this call from being made?

normanmaurer: no idea ... @jasontedor any idea ?

That said its the same that the JDK does

transport-native-unix-common/src/main/c/netty_unix_socket.c
```diff
@@ -1001,27 +997,24 @@ jint netty_unix_socket_JNI_OnLoad(JNIEnv* env, const char* packagePrefix) {
...
-    nettyClassName = netty_unix_util_prepend(packagePrefix, "io/netty/util/NetUtil");
-    jclass localNetUtilClass = (*env)->FindClass(env, nettyClassName);
-    free(nettyClassName);
-    nettyClassName = NULL;
-    if (localNetUtilClass == NULL) {
+
+    // We retrieve java.net.preferIPv4Stack directly via Boolean.getBoolean(...) and not via NetUtil to ensure
+    // we not run into deadlocks while trying to load classes that may depend on each other.
+    // See https://github.com/netty/netty/issues/7458
+    jclass booleanClass = (*env)->FindClass(env, "java/lang/Boolean");
```
carl-mastrangelo: This is really hacky.  IMO the better long term solution would be to pass in all the dependencies (like isipv4Preferred) using some `Java_io_netty_epoll_init()` like method.  

As is it's probably fine.

transport-native-unix-common/src/main/c/netty_unix_socket.c
```diff
@@ -1001,27 +997,24 @@ jint netty_unix_socket_JNI_OnLoad(JNIEnv* env, const char* packagePrefix) {
...
-        // out-of-memory!
-        netty_unix_errors_throwOutOfMemoryError(env);
-        return JNI_ERR;
-    }
-    netUtilClassIpv4PreferredMethodId = (*env)->GetStaticMethodID(env, netUtilClass, "isIpV4StackPreferred", "()Z" );
-    if (netUtilClassIpv4PreferredMethodId == NULL) {
-        // position method was not found.. something is wrong so bail out
-        netty_unix_errors_throwRuntimeException(env, "failed to get method ID: NetUild.isIpV4StackPreferred()");
+
+    jmethodID booleanClassGetBooleanMethodId = (*env)->GetStaticMethodID(env, booleanClass, "getBoolean", "(Ljava/lang/String;)Z" );
```
carl-mastrangelo: `s/" )/")/`

transport-native-unix-common/src/main/c/netty_unix_socket.c
carl-mastrangelo: Maybe check if there is a pending exception after this call and return early?

transport-native-unix-common/src/main/c/netty_unix_socket.c
normanmaurer: Sure I can try this as well just thought this is good enough as it’s the same as openjdk does without a problem... WDYT?

carl-mastrangelo: OpenJDK is in a unique position because it also controls all the classes that are being loaded at the time.  The JDK can take shortcuts that in general outside code can't.  (side note: this is also true for concurrent classes!).  


Like I said this is fine as is for working around the current issue.   However, if any other native loading happens in any of the other JNI_OnLoad  methods they would need to be fixed too.  My concern was that it will be long term brittle.

normanmaurer: @carl-mastrangelo got it... let me try to find time to come-up with another pr.

Scottmitch: +1 for passing info into JNI if possible instead of JNI reaching back to java.

normanmaurer:
Let me close this one in favour of https://github.com/netty/netty/pull/7470

PR #7463, opened 2017-12-01
Introduce ObjectCleaner and use it in FastThreadLocal to ensure FastThreadLocal.onRemoval(...) is called

normanmaurer:
Motivation:

    There is no guarantee that FastThreadLocal.onRemoval(...) is called if the FastThreadLocal is used by "non" FastThreacLocalThreads. This can lead to all sort of problems, like for example memory leaks as direct memory is not correctly cleaned up etc.

    Beside this we use ThreadDeathWatcher to check if we need to release buffers back to the pool when thread local caches are collected. In the past ThreadDeathWatcher was used which will need to "wakeup" every second to check if the registered Threads are still alive. If we can ensure FastThreadLocal.onRemoval(...) is called we do not need this anymore.

    Modifications:

    - Introduce ObjectCleaner and use it to ensure FastThreadLocal.onRemoval(...) is always called when a Thread is collected.
    - Deprecate ThreadDeathWatcher
    - Add unit tests.

    Result:

    Consistent way of cleanup FastThreadLocals when a Thread is collected.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java
+++ b/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java
@@ -436,18 +436,13 @@ public class PooledByteBufAllocator extends AbstractByteBufAllocator implements 
             final PoolArena<ByteBuffer> directArena = leastUsedArena(directArenas);
 
             Thread current = Thread.currentThread();
-            boolean fastThread = current instanceof FastThreadLocalThread;
             if (useCacheForAllThreads || current instanceof FastThreadLocalThread) {
-                // If our FastThreadLocalThread will call FastThreadLocal.removeAll() we not need to use
-                // the ThreadDeathWatcher to release memory from the PoolThreadCache once the Thread dies.
-                boolean useTheadWatcher = fastThread ?
-                        !((FastThreadLocalThread) current).willCleanupFastThreadLocals() : true;
                 return new PoolThreadCache(
                         heapArena, directArena, tinyCacheSize, smallCacheSize, normalCacheSize,
-                        DEFAULT_MAX_CACHED_BUFFER_CAPACITY, DEFAULT_CACHE_TRIM_INTERVAL, useTheadWatcher);
+                        DEFAULT_MAX_CACHED_BUFFER_CAPACITY, DEFAULT_CACHE_TRIM_INTERVAL);
             }
-            // No caching for non FastThreadLocalThreads.
-            return new PoolThreadCache(heapArena, directArena, 0, 0, 0, 0, 0, false);
+            // No caching so just use 0 as sizes.
+            return new PoolThreadCache(heapArena, directArena, 0, 0, 0, 0, 0);
         }
 
         @Override
--- a/common/src/main/java/io/netty/util/Recycler.java
+++ b/common/src/main/java/io/netty/util/Recycler.java
@@ -285,7 +285,7 @@ public abstract class Recycler<T> {
         static WeakOrderQueue allocate(Stack<?> stack, Thread thread) {
             // We allocated a Link so reserve the space
             return reserveSpace(stack.availableSharedCapacity, LINK_CAPACITY)
-                    ? WeakOrderQueue.newQueue(stack, thread) : null;
+                    ? newQueue(stack, thread) : null;
         }
 
         private static boolean reserveSpace(AtomicInteger availableSharedCapacity, int space) {
--- a/common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
+++ b/common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
@@ -17,6 +17,7 @@ package io.netty.util.concurrent;
 
 import io.netty.util.internal.InternalThreadLocalMap;
 import io.netty.util.internal.PlatformDependent;
+import io.netty.util.internal.ObjectCleaner;
 
 import java.util.Collections;
 import java.util.IdentityHashMap;
@@ -131,8 +132,30 @@ public class FastThreadLocal<V> {
     /**
      * Returns the current value for the current thread
      */
+    @SuppressWarnings("unchecked")
     public final V get() {
-        return get(InternalThreadLocalMap.get());
+        final InternalThreadLocalMap threadLocalMap = InternalThreadLocalMap.get();
+        Object v = threadLocalMap.indexedVariable(index);
+        if (v != InternalThreadLocalMap.UNSET) {
+            return (V) v;
+        }
+
+        V value = initialize(threadLocalMap);
+        Thread current = Thread.currentThread();
+        if (!FastThreadLocalThread.willCleanupFastThreadLocals(current)) {
+            // We will need to ensure we will trigger remove(InternalThreadLocalMap) so everything will be released
+            // and FastThreadLocal.onRemoval(...) will be called.
+            ObjectCleaner.register(current, new Runnable() {
+                @Override
+                public void run() {
+                    remove(threadLocalMap);
+
+                    // It's fine to not call InternalThreadLocalMap.remove() here as this will only be triggered once
+                    // the Thread is collected by GC. In this case the ThreadLocal will be gone away already.
+                }
+            });
+        }
+        return value;
     }
 
     /**
--- a/common/src/main/java/io/netty/util/concurrent/FastThreadLocalThread.java
+++ b/common/src/main/java/io/netty/util/concurrent/FastThreadLocalThread.java
@@ -89,4 +89,13 @@ public class FastThreadLocalThread extends Thread {
     public boolean willCleanupFastThreadLocals() {
         return cleanupFastThreadLocals;
     }
+
+    /**
+     * Returns {@code true} if {@link FastThreadLocal#removeAll()} will be called once {@link Thread#run()} completes.
+     */
+    @UnstableApi
+    public static boolean willCleanupFastThreadLocals(Thread thread) {
+        return thread instanceof FastThreadLocalThread &&
+                ((FastThreadLocalThread) thread).willCleanupFastThreadLocals();
+    }
 }
--- a/common/src/test/java/io/netty/util/concurrent/FastThreadLocalTest.java
+++ b/common/src/test/java/io/netty/util/concurrent/FastThreadLocalTest.java
@@ -75,4 +75,61 @@ public class FastThreadLocalTest {
             throw t;
         }
     }
+
+    @Test(timeout = 4000)
+    public void testOnRemoveCalledForFastThreadLocal() throws Exception {
+        testOnRemoveCalled(true);
+    }
+
+    @Test(timeout = 4000)
+    public void testOnRemoveCalledForNonFastThreadLocal() throws Exception {
+        testOnRemoveCalled(false);
+    }
+
+    private static void testOnRemoveCalled(boolean fastThreadLocal) throws Exception {
+
+        final TestFastThreadLocal threadLocal = new TestFastThreadLocal();
+        final TestFastThreadLocal threadLocal2 = new TestFastThreadLocal();
+
+        Runnable runnable = new Runnable() {
+            @Override
+            public void run() {
+                assertEquals(Thread.currentThread().getName(), threadLocal.get());
+                assertEquals(Thread.currentThread().getName(), threadLocal2.get());
+            }
+        };
+        Thread thread = fastThreadLocal ? new FastThreadLocalThread(runnable) : new Thread(runnable);
+        thread.start();
+        thread.join();
+
+        String threadName = thread.getName();
+
+        // Null this out so it can be collected
+        thread = null;
+
+        // Loop until onRemoval(...) was called. This will fail the test if this not works due a timeout.
+        while (threadLocal.onRemovalCalled.get() == null || threadLocal2.onRemovalCalled.get() == null) {
+            System.gc();
+            System.runFinalization();
+            Thread.sleep(50);
+        }
+
+        assertEquals(threadName, threadLocal.onRemovalCalled.get());
+        assertEquals(threadName, threadLocal2.onRemovalCalled.get());
+    }
+
+    private static final class TestFastThreadLocal extends FastThreadLocal<String> {
+
+        final AtomicReference<String> onRemovalCalled = new AtomicReference<String>();
+
+        @Override
+        protected String initialValue() throws Exception {
+            return Thread.currentThread().getName();
+        }
+
+        @Override
+        protected void onRemoval(String value) throws Exception {
+            onRemovalCalled.set(value);
+        }
+    }
 }
--- a/buffer/src/main/java/io/netty/buffer/PoolThreadCache.java
+++ b/buffer/src/main/java/io/netty/buffer/PoolThreadCache.java
@@ -20,7 +20,6 @@ package io.netty.buffer;
 import io.netty.buffer.PoolArena.SizeClass;
 import io.netty.util.Recycler;
 import io.netty.util.Recycler.Handle;
-import io.netty.util.ThreadDeathWatcher;
 import io.netty.util.internal.MathUtil;
 import io.netty.util.internal.PlatformDependent;
 import io.netty.util.internal.logging.InternalLogger;
@@ -56,9 +55,6 @@ final class PoolThreadCache {
     private final int numShiftsNormalHeap;
     private final int freeSweepAllocationThreshold;
 
-    private final Thread deathWatchThread;
-    private final Runnable freeTask;
-
     private int allocations;
 
     // TODO: Test if adding padding helps under contention
@@ -66,8 +62,7 @@ final class PoolThreadCache {
 
     PoolThreadCache(PoolArena<byte[]> heapArena, PoolArena<ByteBuffer> directArena,
                     int tinyCacheSize, int smallCacheSize, int normalCacheSize,
-                    int maxCachedBufferCapacity, int freeSweepAllocationThreshold,
-                    boolean useThreadDeathWatcher) {
+                    int maxCachedBufferCapacity, int freeSweepAllocationThreshold) {
         if (maxCachedBufferCapacity < 0) {
             throw new IllegalArgumentException("maxCachedBufferCapacity: "
                     + maxCachedBufferCapacity + " (expected: >= 0)");
@@ -119,25 +114,6 @@ final class PoolThreadCache {
                 && freeSweepAllocationThreshold < 1) {
             throw new IllegalArgumentException("freeSweepAllocationThreshold: "
                     + freeSweepAllocationThreshold + " (expected: > 0)");
-        }
-
-        if (useThreadDeathWatcher) {
-
-            freeTask = new Runnable() {
-                @Override
-                public void run() {
-                    free0();
-                }
-            };
-
-            deathWatchThread = Thread.currentThread();
-
-            // The thread-local cache will keep a list of pooled buffers which must be returned to
-            // the pool when the thread is not alive anymore.
-            ThreadDeathWatcher.watch(deathWatchThread, freeTask);
-        } else {
-            freeTask = null;
-            deathWatchThread = null;
         }
     }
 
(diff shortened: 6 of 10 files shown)


normanmaurer:
There is still some TODO to fix ... Just wanted to show the idea

johnou:
@normanmaurer could the same approach be used with Recycler instead of finalize?

normanmaurer:
@johnou not sure what you refer to with finalize.... any pointers?

johnou:
@normanmaurer 

io.netty.util.Recycler.WeakOrderQueue#finalize
```java
        @Override
        protected void finalize() throws Throwable {
            try {
                super.finalize();
            } finally {
                // We need to reclaim all space that was reserved by this WeakOrderQueue so we not run out of space in
                // the stack. This is needed as we not have a good life-time control over the queue as it is used in a
                // WeakHashMap which will drop it at any time.
                Link link = head;
                while (link != null) {
                    reclaimSpace(LINK_CAPACITY);
                    link = link.next;
                }
            }
        }
```

normanmaurer:
Yes most likely

common/src/main/java/io/netty/util/internal/AutomaticCleanupReference.java
```diff
@@ -0,0 +1,94 @@
...
+    private static final ReferenceQueue<Object> REFERENCE_QUEUE = new ReferenceQueue<Object>();
+
+    static {
+        AccessController.doPrivileged(new PrivilegedAction<Object>() {
+            @Override
+            public Object run() {
+                Thread cleanupThread = new Thread(new Runnable() {
+                    @Override
+                    public void run() {
+                        for (;;) {
```
ejona86: So this runs forever? Isn't that _guaranteed_ to be a classloader leak via this Runnable?

common/src/main/java/io/netty/util/internal/ThreadCleaner.java
```diff
@@ -30,38 +29,45 @@
...
-                cleanupThread.setContextClassLoader(null);
-                cleanupThread.setName("ThreadCleanerReaper");
-                cleanupThread.setDaemon(true);
-                cleanupThread.start();
-                return null;
+                }
+                CLEANER_RUNNING.set(false);
+
+                // Its important to first access the LIVE_SET and then CLEANER_RUNNING to ensure correct
+                // behavior in multi-htraded enviroments.
```
ejona86: s/htraded/threaded/

common/src/main/java/io/netty/util/internal/ThreadCleaner.java
```diff
@@ -0,0 +1,131 @@
...
+            }
+        }
+    };
+
+    /**
+     * Register the given {@link Thread} for which the {@link Runnable} will be executed once there are no references
+     * to the object anymore, which typically happens once the {@link Thread} dies.
+     *
+     * This should only be used if there are no other ways to execute some cleanup once the {@link Thread} dies as
+     * its not a cheap way to handle the cleanup.
```
ejona86: nit: s/its/it's/

common/src/main/java/io/netty/util/internal/ThreadCleaner.java
```diff
@@ -0,0 +1,131 @@
...
+            cleanupThread.setName("ThreadCleanerReaper");
+
+            // This Thread is not a daemon as it will die once all references to the registered Threads will go away.
+            cleanupThread.setDaemon(false);
+            cleanupThread.start();
+        }
+    }
+
+    private ThreadCleaner() {
+        // Only contains a  static method.
```
ejona86: nit: there's a double space

common/src/main/java/io/netty/util/internal/ThreadCleaner.java
```diff
@@ -0,0 +1,131 @@
...
+     * its not a cheap way to handle the cleanup.
+     */
+    public static void register(Thread thread, Runnable cleanupTask) {
+        ThreadCleanerReference reference = new ThreadCleanerReference(thread,
+                ObjectUtil.checkNotNull(cleanupTask, "cleanupTask"));
+        // Its important to add the reference to the LIVE_SET before we access CLEANER_RUNNING to ensure correct
+        // behavior in multi-threaded enviroments.
+        LIVE_SET.add(reference);
+
+        // Check if there is arleady
```
ejona86: s/arleady/already/

ejona86:
> I guess the failure case is the ThreadLocal instance could be thrown away, and let the storage evaporate (since Thread uses WeakReference for the mapping), but the Thread itself remain running.

So, this turned out to be semi-wrong. Or rather, if you leave a Thread running, "you're gonna have a bad time." I realized that even using `finalize()` you can't necessarily clean up resources if the Thread lives because ThreadLocal is only lazily cleaned up in the ThreadLocalMap. Any values held are strong references, and so they might be retained for the rest of the Thread's lifetime.

So I'm not trying to solve that any more :-). And that makes multiple solutions we've considered solvable. The only thing we need to be wary of is clearing the Thread.contextClassLoader.

buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java
```diff
@@ -440,14 +441,26 @@ protected synchronized PoolThreadCache initialValue() {
             if (useCacheForAllThreads || current instanceof FastThreadLocalThread) {
                 // If our FastThreadLocalThread will call FastThreadLocal.removeAll() we not need to use
                 // the ThreadDeathWatcher to release memory from the PoolThreadCache once the Thread dies.
```
johnou: update javadoc?

common/src/main/java/io/netty/util/Recycler.java
```diff
@@ -416,7 +416,10 @@ protected void finalize() throws Throwable {
         // to scavenge those that can be reused. this permits us to incur minimal thread synchronisation whilst
         // still recycling all items.
         final Recycler<T> parent;
-        final Thread thread;
+
+        // We store the Thread in a WeakReference as otherwise we may be the only ones that still hold a strong
+        // Reference to the Thread itself after it died.
+        final WeakReference<Thread> threadRef;
```
johnou: should this be in the same pull request?

carl-mastrangelo:
Out of curiosity: why not spawn the cleaner thread when the original thread goes away?    That would keep the number of active threads down.   

Also, why not do the cleanup in a `finalize()` method?   This is actually garbage collection, so it seems reasonable to do it there and not spawn another thread.

buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java
normanmaurer: +1 ... Let me do this

common/src/main/java/io/netty/util/Recycler.java
normanmaurer: I did it in the same as I noticed it here because the Thread was never GCed. Maybe I should do it in a separate PR tho..

common/src/main/java/io/netty/util/internal/ThreadCleaner.java
```diff
@@ -0,0 +1,131 @@
...
+import java.lang.ref.ReferenceQueue;
+import java.lang.ref.WeakReference;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * Allows a way to register some {@link Runnable} that will executed once there are no references to the {@link Thread}
+ * anymore. This typically happens once the {@link Thread} dies / completes.
+ */
+public final class ThreadCleaner {
```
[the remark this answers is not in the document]
normanmaurer: nope... `WeakReference` was good enough.

normanmaurer:
@carl-mastrangelo answers / comments below....


> Out of curiosity: why not spawn the cleaner thread when the original thread goes away, rather than when it is first attached? That would keep the number of active threads down.

How would you detect that the original Thread goes away ? Also I think its not so bad to have one extra thread that is doing the cleanup for everything. I will most likely use it in other places as well at some point.

> Also, why not do the cleanup in a finalize() method? This is actually garbage collection, so it seems reasonable to do it there and not spawn another thread.

I actually had a PR (#7446) with a `finalize()` at some point but @ejona86 asked me to not do this as `finalize()` is marked as `@deprecated` in Java9 and so should not be used anymore.

johnou:
@carl-mastrangelo Object.finalize is deprecated, the recommended replacements are PhantomReference and the new Cleaner class.

common/src/main/java/io/netty/util/Recycler.java
normanmaurer: @johnou https://github.com/netty/netty/pull/7499 ... Even added a test-case :)

normanmaurer:
@johnou @ejona86 @carl-mastrangelo PTAL again as I changed the approach a bit in a followup commit that I pushed. So the "real problem" why we needed all of this is because of the user uses a non `FastThreadLocalThread` there is no guarantee that we will call `FastThreadLocal.removeAll()` at all once the Thread goes away. The cleanup of the `PoolThreadCache` is done in `FastThreadLocal.onRemove(...)` that will be triggered by the `removeAll()`.

Because of this I now moved the usage of `ThreadCleaner.register(...)` into the `FastThreadLocal` itself which has the affect that we are guaranteed that `onRemoval(...)` is always called (even if we use a non `FastThreadLocal`. I think this is a more general fix to the problem and also fixes a possible memory leak that could happen `NativeDatagramPacketArray` if non `FastThreadLocal` threads are used to access it as `onRemoval` would never been called before for these:

https://github.com/netty/netty/blob/4.1/transport-native-epoll/src/main/java/io/netty/channel/epoll/NativeDatagramPacketArray.java#L43

WDYT ?

common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
```diff
@@ -57,21 +58,25 @@ public static void removeAll() {
...
-                    tlv.remove(threadLocalMap);
-                }
-            }
+            removeAll(threadLocalMap);
         } finally {
             InternalThreadLocalMap.remove();
         }
     }
 
+    private static void removeAll(InternalThreadLocalMap threadLocalMap) {
```
ejona86: Is there a reason this was split out into its own function?

I had expected the `Runnable` passed to `ThreadCleaner` would use it, but it uses `remove(...)` instead. It's a bit hard to determine what's the difference between the two methods.

normanmaurer: Actually sorry it should use `removeAll` and I messed up :(

ejona86:
> Because of this I now moved the usage of ThreadCleaner.register(...) into the FastThreadLocal itself which has the affect that we are guaranteed that onRemoval(...) is always called (even if we use a non FastThreadLocal

It also means that you'll use the ThreadCleaner even if nothing stored in the FastThreadLocal needs cleaning. Normally that'd be rare, but if caching is disabled for non-FastThreadLocalThreads, then it'd be more frequent. Not sure it matters at all.

normanmaurer:
@ejona86 yeah I agree... That said I think it may be still the preferred way to fix it and ensure `onRemoval` is always called and so can depend on it... WDYT ?

common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
normanmaurer: Sorry what I wanted to say was that we not need to factor out removeAll and can just call remove... will fix it.

common/src/main/java/io/netty/util/concurrent/FastThreadLocalThread.java
```diff
@@ -89,4 +89,13 @@ public final void setThreadLocalMap(InternalThreadLocalMap threadLocalMap) {
     public boolean willCleanupFastThreadLocals() {
         return cleanupFastThreadLocals;
     }
+
+    /**
+     * Returns {@code true} if {@link FastThreadLocal#removeAll()} will be called once {@link Thread#run()} completes.
+     */
+    @UnstableApi
+    public static boolean willCleanupFastThreadLocals(Thread thread) {
+        return thread instanceof FastThreadLocalThread && ((FastThreadLocalThread) thread)
```
carl-mastrangelo: might be clearer to break on `&&`

common/src/main/java/io/netty/util/internal/ThreadCleaner.java
```diff
@@ -0,0 +1,131 @@
...
+
+    // This will hold a reference to the ThreadCleanerReference which will be removed once we called cleanup()
+    private static final Set<ThreadCleanerReference> LIVE_SET = new ConcurrentSet<ThreadCleanerReference>();
+    private static final ReferenceQueue<Object> REFERENCE_QUEUE = new ReferenceQueue<Object>();
+    private static final AtomicBoolean CLEANER_RUNNING = new AtomicBoolean(false);
```
carl-mastrangelo: nit: caps are usually for constants.

common/src/main/java/io/netty/util/internal/ThreadCleaner.java
```diff
@@ -0,0 +1,131 @@
...
+            // Set to null to ensure we not create classloader leaks by holds a strong reference to the inherited
+            // classloader.
+            // See:
+            // - https://github.com/netty/netty/issues/7290
+            // - https://bugs.openjdk.java.net/browse/JDK-7008595
+            cleanupThread.setContextClassLoader(null);
+            cleanupThread.setName("ThreadCleanerReaper");
+
+            // This Thread is not a daemon as it will die once all references to the registered Threads will go away.
+            cleanupThread.setDaemon(false);
```
carl-mastrangelo: Daemon is for not blocking shutdown.  In a shutdown scenario, how important is it to clean up refs?

ejona86:
> That said I think it may be still the preferred way to fix it and ensure `onRemoval` is always called and so can depend on it... WDYT ?

I'm fine with it. It might be unnecessary in some cases, but doesn't seem like it hurts much and does remove a type of bug-prone behavior.

common/src/main/java/io/netty/util/internal/ThreadCleaner.java
normanmaurer: actually I think we could also just use a non-daemon thread as the thread will go away once there is no references anymore.

common/src/main/java/io/netty/util/internal/ThreadCleaner.java
normanmaurer: I think CAPS are for things that are static final so I think this is fine. Sure the content can be changed but this is ok IMHO.

normanmaurer:
@carl-mastrangelo PTAL again

common/src/main/java/io/netty/util/internal/ThreadCleaner.java
```diff
@@ -0,0 +1,132 @@
...
+    };
+
+    /**
+     * Register the given {@link Thread} for which the {@link Runnable} will be executed once there are no references
+     * to the object anymore, which typically happens once the {@link Thread} dies.
+     *
+     * This should only be used if there are no other ways to execute some cleanup once the {@link Thread} dies as
+     * it is not a cheap way to handle the cleanup.
+     */
+    public static void register(Thread thread, Runnable cleanupTask) {
```
Scottmitch: this class is `internal` ... consider:
1. exposing a method where the user can directly create a `ThreadCleanerReference` to avoid allocating a `Runnable` and `ThreadCleanerReference`
2. Make the class package private and you can use the same technique as above to reduce object allocation.

common/src/main/java/io/netty/util/internal/ThreadCleaner.java
```diff
@@ -0,0 +1,132 @@
...
+                // Its important to first access the LIVE_SET and then CLEANER_RUNNING to ensure correct
+                // behavior in multi-threaded environments.
+                if (LIVE_SET.isEmpty() || !CLEANER_RUNNING.compareAndSet(false, true)) {
+                    // There was nothing added after we set STARTED to false or some other cleanup Thread
+                    // was started already so its safe to let this Thread complete now.
+                    break;
+                }
+            }
+            if (interrupted) {
+                // As we catched the InterruptedException above we should mark the Thread as interrupted.
```
Scottmitch: `catched` -> `caught`

common/src/main/java/io/netty/util/internal/ThreadCleaner.java
normanmaurer: I am not concerned about this as this should be something that is done very frequently and I think this design is cleaner.

common/src/main/java/io/netty/util/internal/ThreadCleaner.java
```diff
@@ -0,0 +1,132 @@
...
+     */
+    public static void register(Thread thread, Runnable cleanupTask) {
+        ThreadCleanerReference reference = new ThreadCleanerReference(thread,
+                ObjectUtil.checkNotNull(cleanupTask, "cleanupTask"));
+        // Its important to add the reference to the LIVE_SET before we access CLEANER_RUNNING to ensure correct
+        // behavior in multi-threaded environments.
+        LIVE_SET.add(reference);
+
+        // Check if there is already
+        if (CLEANER_RUNNING.compareAndSet(false, true)) {
```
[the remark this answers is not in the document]
normanmaurer: again I think we should not care as this should not happen very frequently and the CAS is the thing that I would be concerned last about (as the other stuff will be the "expensive")

normanmaurer:
@Scottmitch yeah I think how we test here is the best we can do for now...

common/src/main/java/io/netty/util/internal/ThreadCleaner.java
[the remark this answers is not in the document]
Scottmitch: fair enough ... I guess even if folks are doing get/remove this may not buy much.

common/src/main/java/io/netty/util/internal/ThreadCleaner.java
Scottmitch: Agreed the current APIs are nicer. However it is internal and doesn't need to be generic right? We could potentially save 1 object allocation per thread local. Since it is internal we can always change it later but maybe nice to keep thread local overhead down if we don't have to be generic.

normanmaurer: @Scottmitch as discussed offline I will merge it as it is for now and we can address this if we find out its an issue later on. The current design is nicer :)

carl-mastrangelo:
cc: @zpencer

common/src/main/java/io/netty/util/internal/ThreadCleaner.java
Scottmitch: sgtm ... internal == change later if necessary

normanmaurer: exactly 👍

normanmaurer:
@mosesn you may be interested as well.

normanmaurer:
Cherry-picked into 4.1 (e329ca1cf3cecd41c9cf1a5a7be369a65f002ab5) and 4.0 (00ae6274c015804546f5f12d2b795ec4d105ee71)

common/src/test/java/io/netty/util/concurrent/FastThreadLocalTest.java
```diff
@@ -75,4 +75,61 @@ public void run() {
...
+
+    private static void testOnRemoveCalled(boolean fastThreadLocal) throws Exception {
+
+        final TestFastThreadLocal threadLocal = new TestFastThreadLocal();
+        final TestFastThreadLocal threadLocal2 = new TestFastThreadLocal();
+
+        Runnable runnable = new Runnable() {
+            @Override
+            public void run() {
+                assertEquals(Thread.currentThread().getName(), threadLocal.get());
```
zpencer: A bit late to this PR, but what happens if code sets a FastThreadLocal and the thread exits before it's read? It looks like the cleaner code is only registered on the `get` path.

normanmaurer: @zpencer good point. Let me fix this.

normanmaurer: @zpencer https://github.com/netty/netty/pull/7534 PTAL

PR #7468, opened 2017-12-04
SslHandler aggregation prefer copy over CompositeByteBuf

Scottmitch:
Motivation:
SslHandler will do aggregation of writes by default in an attempt to improve goodput and reduce the number of discrete buffers which must be accumulated. However if aggregation is not possible then a CompositeByteBuf is used to accumulate multiple buffers. Using a CompositeByteBuf doesn't provide any of the benefits of better goodput and in the case of small + large writes (e.g. http/2 frame header + data) this can reduce the amount of data that can be passed to writev by about half. This has the impact of increasing latency as well as reducing goodput.

Modifications:
- SslHandler should prefer copying instead of using a CompositeByteBuf

Result:
Better goodput (and potentially improved latency) at the cost of copy operations.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -1882,7 +1882,7 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
                 return composite;
             }
             return attemptCopyToCumulation(cumulation, next, wrapDataSize) ? cumulation :
-                    composeIntoComposite(alloc, cumulation, next);
+                    copyAndCompose(alloc, cumulation, next);
         }
 
         @Override
--- a/transport/src/main/java/io/netty/channel/AbstractCoalescingBufferQueue.java
+++ b/transport/src/main/java/io/netty/channel/AbstractCoalescingBufferQueue.java
@@ -17,16 +17,16 @@ package io.netty.channel;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.CompositeByteBuf;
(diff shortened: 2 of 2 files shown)


handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1882,7 +1882,7 @@ protected ByteBuf compose(ByteBufAllocator alloc, ByteBuf cumulation, ByteBuf ne
                 return composite;
             }
             return attemptCopyToCumulation(cumulation, next, wrapDataSize) ? cumulation :
-                    composeIntoComposite(alloc, cumulation, next);
+                    copyAndCompose(alloc, cumulation, next);
```
normanmaurer: @Scottmitch I wonder if we want to make this configurable... Does this make any sense ?

johnou:
@Scottmitch any benchmarks you can share?

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
Scottmitch: It is possible that folks may not want to copy under some situations (e.g. if each packet is over a certain threshold), but I'm not sure if providing this level of control is justified just yet. I propose we wait until it is necessary until we increase the API footprint. I think we could do something like `BiPredicate<ByteBuf, ByteBuf>` to determine if we should copy or use a Composite when the time comes.  WDYT?

normanmaurer: Yep fine as it is

Scottmitch:
@johnou - Unfortunately not.

Scottmitch:
4.1 (d7c977dd71)

PR #7469, opened 2017-12-04
DefaultMaxMessagesRecvByteBufAllocator support an option to ignore ma…

Scottmitch:
…ybeMoreData

Motivation:
If large amounts of data is being transferred it is difficult to correlate the amount we attempt to read vs the maximum amount that the OS will actually buffer and deliver to the application. For exmaple some OSes may dynicamlly update the SO_RCVBUF size or otherwise dynamically adjust how much data is delieved to the application. In these circumstances it can reduce latency to just call read() on the socket another time to see if there is really any data remaining instead of giving up the maxMessagesPerRead quantum and going back to the selector to read later.

Motifications:
- Add DefaultMaxMessagesRecvByteBufAllocator#respectMaybeMoreData which provides a way to ignore the maybeMoreData function which may not account for the current data pending, and if it does this maybe racy.

Result:
Option to always use the full maxMessagesPerRead quantum before going back to the selector.

### DIFF
--- a/transport/src/main/java/io/netty/channel/DefaultMaxMessagesRecvByteBufAllocator.java
+++ b/transport/src/main/java/io/netty/channel/DefaultMaxMessagesRecvByteBufAllocator.java
@@ -25,6 +25,7 @@ import io.netty.util.UncheckedBooleanSupplier;
  */
 public abstract class DefaultMaxMessagesRecvByteBufAllocator implements MaxMessagesRecvByteBufAllocator {
     private volatile int maxMessagesPerRead;
+    private volatile boolean respectMaybeMoreData = true;
 
     public DefaultMaxMessagesRecvByteBufAllocator() {
         this(1);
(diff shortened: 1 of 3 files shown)


transport/src/main/java/io/netty/channel/DefaultMaxMessagesRecvByteBufAllocator.java
```diff
@@ -48,6 +49,38 @@ public MaxMessagesRecvByteBufAllocator maxMessagesPerRead(int maxMessagesPerRead
...
+     * <ul>
+     *     <li>{@code true} to stop reading if we think there is no more data. This may save a system call to read from
+     *          the socket, but if data has arrived in a racy fashion we may give up our {@link #maxMessagesPerRead()}
+     *          quantum and have to wait for the selector to notify us of more data.</li>
+     *     <li>{@code false} to keep reading (up to {@link #maxMessagesPerRead()}) or until there is no data when we
+     *          attempt to read.</li>
+     * </ul>
+     * @return {@code this}.
+     */
+    public MaxMessagesRecvByteBufAllocator respectMaybeMoreData(boolean respectMaybeMoreData) {
```
normanmaurer: Should we change the return type to `DefaultMaxMessagesRecvByteBufAllocator`? Also should we override this methods in the other impls and return the correct type there as well ?

Scottmitch: done

Scottmitch:
4.1 (95b02e49ac)

PR #7470, opened 2017-12-04
Not call java methods from within JNI init code to prevent class load…

normanmaurer:
…ing deadlocks.

Motivation:

Modifications:

Add a new Socket.initialize() method that will be called when init the library and pass everything needed to the JNI level so we not need to call back to java.

Result:

Fixes [#7458].


normanmaurer:
@carl-mastrangelo @Scottmitch PTAL

normanmaurer:
Let me merge this... @Scottmitch if you have any followup concerns we can address these later on.

normanmaurer:
Cherry-picked into 4.1 (9afa857cc187a7a786002a29c15986ce5d0d53f3). I will open another PR for 4.0 to ensure I not mess up when cherry-pick and resolving merge conflicts.

PR #7475, opened 2017-12-05
Ensure we not try to call `select` when the `AbstractSniHandler` was …

normanmaurer:
…already removed from the pipeline.

Motivation:

We tried to call `select` after we closed the channel (and so removed all the handlers from the pipeline) when we detected a non SSL record. This would cause an exception like this:

```
Caused by: java.util.NoSuchElementException: io.netty.handler.ssl.SniHandler
	at io.netty.channel.DefaultChannelPipeline.getContextOrDie(DefaultChannelPipeline.java:1098)
	at io.netty.channel.DefaultChannelPipeline.replace(DefaultChannelPipeline.java:506)
	at io.netty.handler.ssl.SniHandler.replaceHandler(SniHandler.java:133)
	at io.netty.handler.ssl.SniHandler.onLookupComplete(SniHandler.java:113)
	at io.netty.handler.ssl.AbstractSniHandler.select(AbstractSniHandler.java:225)
	at io.netty.handler.ssl.AbstractSniHandler.decode(AbstractSniHandler.java:218)
	at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:489)
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:428)
	... 40 more
```

Modifications:

- Ensure we rethrow the NotSslRecordException when detecting it (and closing the channel). This will also ensure we not call `select(...)`
- Not catch `Throwable` but only `Exception`
- Add test case.

Result:

Correctly handle the case of an non SSL record.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/AbstractSniHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/AbstractSniHandler.java
@@ -98,7 +98,7 @@ public abstract class AbstractSniHandler<T> extends ByteToMessageDecoder impleme
                             // SSLv3 or TLS
                             if (majorVersion == 3) {
                                 final int packetLength = in.getUnsignedShort(readerIndex + 3) +
-                                                         SslUtils.SSL_RECORD_HEADER_LENGTH;
+                                        SslUtils.SSL_RECORD_HEADER_LENGTH;
 
                                 if (readableBytes < packetLength) {
                                     // client hello incomplete; try again to decode once more data is ready.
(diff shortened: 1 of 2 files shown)


normanmaurer:
PTAL @rkapsi

normanmaurer:
Found this while reviewing another PR.

rkapsi:
@normanmaurer nice catch and +1 for catching Exception.

It appears we worked around it without realizing it's a bug or rather our refCnt aware version of CusomSniHandler needs to deal with these kind of things in `onLookupComplete(...)`.

normanmaurer:
Cherry-picked into 4.1 (1453f8d18b635bdd2d0ef9df9334c88c993a7092) and 4.0 (d729ae716)

PR #7476, opened 2017-12-05
Only try to match SSLException message when debug logging is enabled.

normanmaurer:
Motivation:

We only want to log for the particular case when debug logging is enabled so we not need to try to match the message if this is not the case.

Modifications:

Guard with logger.isDebugEnabled()

Result:

Less overhead when debug logging is not enabled.


normanmaurer:
Cherry-picked into 4.1 (ca1e1fcddf4c9d6b76b596f826ada899649b052d) and 4.0 (54a6b0c5424c027c83492a4e5e460f1cdf455374)

PR #4703, opened 2016-01-13
Correctly handle non handshake commands when using SniHandler

normanmaurer:
Motivation:

As we can only handle handshake commands to parse SNI we should try to skip alert and change cipher spec commands a few times before we fallback to use a default SslContext.

Modifications:

- Use default SslContext if no application data command was received
- Use default SslContext if after 4 commands we not received a handshake command
- Simplify code
- Eliminate multiple volatile fields
- Rename SslConstants to SslUtils
- Share code between SslHandler and SniHandler by moving stuff to SslUtils

Result:

Correct handling of non handshake commands and cleaner code.

PR #7479, opened 2017-12-06
Don't loop over TLS records for SNI

Lukasa:
Motivation:

The AbstractSniHandler previously was willing to tolerate up to three non-handshake records before a ClientHello that contained an SNI extension field. This is, so far as I can tell, completely unnecessary: no TLS implementation will be sending alerts or change cipher spec messages before ClientHello.

Given that it was not possible to determine why this loop is in the code to begin with, it's probably just best to remove it.

Modifications:

Remove the for loop.

Result:

The AbstractSniHandler will more rapidly determine whether it should pass the records on to the default SSL handler.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/AbstractSniHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/AbstractSniHandler.java
@@ -42,9 +42,6 @@ import java.util.Locale;
  */
 public abstract class AbstractSniHandler<T> extends ByteToMessageDecoder implements ChannelOutboundHandler {
 
-    // Maximal number of ssl records to inspect before fallback to the default SslContext.
-    private static final int MAX_SSL_RECORDS = 4;
-
     private static final InternalLogger logger =
             InternalLoggerFactory.getInstance(AbstractSniHandler.class);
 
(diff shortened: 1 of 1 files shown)


normanmaurer:
@rkapsi as well PTAL

rkapsi:
Here's the PR #4703 that introduced it ~2 years ago. Is it one of these things that are technically possible because the spec allows it but practically they're nonsense?

normanmaurer:
@rkapsi yep I talked with @cconroy about it but cant really remember anymore why I thought this would be a good change :(

handler/src/main/java/io/netty/handler/ssl/AbstractSniHandler.java
```diff
@@ -57,156 +54,152 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) t
...
-                                }
+                        SslUtils.notifyHandshakeFailure(ctx, e, true);
+                        throw e;
+                    }
+                    if (len == SslUtils.NOT_ENOUGH_DATA ||
+                        writerIndex - readerIndex - SslUtils.SSL_RECORD_HEADER_LENGTH < len) {
+                        // Not enough data
+                        return;
+                    }
+                    break;
```
Scottmitch: what is the expected behavior when we are in this state? Previously we consumed/ignored that data and moved on, but now we leave the content, bail out and move on. Will this content prevent us from making progress? For example will we just keep reading more data but never make progress because the alert/change_cipher is in the way?

Lukasa: No, I don't think so: if we break out we hit the fallback state and assume there was no SNI present, which allows the actual TLS implementation to sort this mess out.

carl-mastrangelo: Could you add a comment to that effect?  It wasn't obvious from reading this diff alone.

handler/src/main/java/io/netty/handler/ssl/AbstractSniHandler.java
```diff
@@ -57,156 +54,152 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) t
...
-
-                                if (readableBytes < packetLength) {
-                                    // client hello incomplete; try again to decode once more data is ready.
-                                    return;
-                                }
+                        SslUtils.notifyHandshakeFailure(ctx, e, true);
+                        throw e;
+                    }
+                    if (len == SslUtils.NOT_ENOUGH_DATA ||
+                        writerIndex - readerIndex - SslUtils.SSL_RECORD_HEADER_LENGTH < len) {
```
carl-mastrangelo: should this be `in.readableBytes()` ?

Lukasa: What's "this" in this comment? `len`?

Lukasa:
Ok, I've updated this patch. Waiting for some feedback from @carl-mastrangelo.

handler/src/main/java/io/netty/handler/ssl/AbstractSniHandler.java
Scottmitch: Maybe @carl-mastrangelo is talking about `writerIndex - readerIndex` ... which maybe now we can use `readableBytes`?

handler/src/main/java/io/netty/handler/ssl/AbstractSniHandler.java
```diff
@@ -57,156 +54,153 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) t
...
+                        // Not enough data
                         return;
                     }
+                    // SNI can't be present in an ALERT or CHANGE_CIPHER_SPEC record, so we'll fall back and assume
+                    // no SNI is present. Let's let the actual TLS implementation sort this out.
```
Scottmitch: `AbstractSniHandler` is typically responsible for inserting a `SslHandler` into the pipeline after `SNI`. At this point it is likely that there will be no `SslHandler` in the pipeline. IIUC it is likely at this point that we are out of sync with our peer so should we call some overridable method which by default throws some exception and/or closes the channel? Otherwise the content will remain in `in` and it is not clear how we will make progress.

handler/src/main/java/io/netty/handler/ssl/AbstractSniHandler.java
Scottmitch: @carl-mastrangelo - Ping

carl-mastrangelo: I just found out this tab was still open.  yes that is what I meant.

Scottmitch: thanks @carl-mastrangelo ... @Lukasa ^^

normanmaurer:
@netty-bot test this please

normanmaurer:
@Lukasa sorry that it took so long...

PR #7480, opened 2017-12-06
Not call java methods from within JNI init code to prevent class load…

normanmaurer:
…ing deadlocks.

Motivation:

Modifications:

Add a new Socket.initialize() method that will be called when init the library and pass everything needed to the JNI level so we not need to call back to java.

Result:

Fixes [#7458].


normanmaurer:
Cherry-picked into 4.0 (f921ea344e)

PR #7491, opened 2017-12-10
Reduce Object allocations in CompositeByteBuf.

normanmaurer:
Motivation:

We used subList in CompositeByteBuf to remove ranges of elements from the internal storage. Beside this we also used an foreach loop in a few cases which will crate an Iterator.

Modifications:

- Use our own sub-class of ArrayList which exposes removeRange(...). This allows to remove a range of elements without an extra allocation.
- Use an old style for loop to iterate over the elements to reduce object allocations.

Result:

Less allocations.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
@@ -48,7 +48,7 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
 
     private final ByteBufAllocator alloc;
     private final boolean direct;
-    private final List<Component> components;
+    private final ComponentList components;
     private final int maxNumComponents;
 
     private boolean freed;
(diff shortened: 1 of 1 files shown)


buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
```diff
@@ -1612,8 +1614,9 @@ public CompositeByteBuf discardReadBytes() {
         // Discard everything if (readerIndex = writerIndex = capacity).
         int writerIndex = writerIndex();
         if (readerIndex == writerIndex && writerIndex == capacity()) {
-            for (Component c: components) {
-                c.freeIfNecessary();
+            int size = components.size();
+            for (int i = 0 ; i < size; i++) {
```
carl-mastrangelo: `s/0 ;/0;/`

normanmaurer: good catch... addressed

normanmaurer:
Cherry-picked into 4.1 (1988cd041d3d5eaa5f52e58d3d3f773b9ef17fe7) and 4.0 (4a8748aa3c68dc6833f32a30e3739046b77016f1)

Issue #7483, opened 2017-12-07
SingleThreadEventExecutor ignores startThread failures


rvansa:
### Expected behavior
SingleThreadEventExecutor should throw an exception when it can't start the thread second time.

### Actual behavior
When `doStartThread` throws an exception, e.g. due to the actual executor being depleted of threads and throwing in its rejected execution handler, the STEE ends up in started state anyway. If we try to execute another task in this executor, it will be queued but the thread won't be started anymore and the task will linger forever.

### Netty version
4.1.9

### JVM version (e.g. `java -version`)
1.8

normanmaurer:
@rvansa looking now

normanmaurer:
@rvansa PTAL  #7492

normanmaurer:
Fixed by https://github.com/netty/netty/pull/7492

PR #7492, opened 2017-12-11
SingleThreadEventExecutor ignores startThread failures

normanmaurer:
Motivation:

When doStartThread throws an exception, e.g. due to the actual executor being depleted of threads and throwing in its rejected execution handler, the STEE ends up in started state anyway. If we try to execute another task in this executor, it will be queued but the thread won't be started anymore and the task will linger forever.

Modifications:

- Ensure we not update the internal state if the startThread() method throws.
- Add testcase

Result:

Fixes [#7483]

### DIFF
--- a/common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
+++ b/common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
@@ -16,6 +16,7 @@
 package io.netty.util.concurrent;
 
 import io.netty.util.internal.ObjectUtil;
+import io.netty.util.internal.PlatformDependent;
 import io.netty.util.internal.SystemPropertyUtil;
 import io.netty.util.internal.UnstableApi;
 import io.netty.util.internal.logging.InternalLogger;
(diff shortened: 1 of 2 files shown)


normanmaurer:
@rvansa PTAL as well

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
```diff
@@ -582,7 +583,14 @@ private boolean runShutdownHooks() {
         gracefulShutdownTimeout = unit.toNanos(timeout);
 
         if (oldState == ST_NOT_STARTED) {
-            doStartThread();
+            try {
+                doStartThread();
+            } catch (RejectedExecutionException cause) {
```
rvansa: `j.u.c.RejectedExecutionHandler` is not limited to throwing REEs, why not `Throwable`?

normanmaurer: We try to not introduce more Throwable catches if we rethrow these. Maybe `Exception` ?

rvansa: Sounds good.

normanmaurer: @rvansa PTAL again just pushed

normanmaurer:
@rvansa PTAL again

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
```diff
@@ -582,7 +583,18 @@ private boolean runShutdownHooks() {
...
 
         if (oldState == ST_NOT_STARTED) {
-            doStartThread();
+            try {
+                doStartThread();
+            } catch (Throwable cause) {
+                STATE_UPDATER.set(this, ST_TERMINATED);
+                terminationFuture.tryFailure(cause);
+
+                if (!(cause instanceof Exception)) {
```
carl-mastrangelo: optional:

`if (cause instanceof Error)) {`

normanmaurer: I think I would like to keep it as it is for now just in case someone creates a special `Throwable` sub-type

normanmaurer:
Cherry-picked into 4.1 (5ad35a157cac27004b492703708ecebc4442a55a) ... 4.0 does not have the problem as it not accepts an Executor.

Issue #7290, opened 2017-10-10
ThreadDeathWatcher causes custom classLoader script memory leaks


juebanlin:
### Steps to reproduce
`package net.jueb.redAlert.script.game.serverScript.sys;

import io.netty.buffer.ByteBufAllocator;
import net.jueb.redAlert.core.base.message.AppByteMessage;
import net.jueb.redAlert.core.base.message.MsgCode;
import net.jueb.redAlert.game.ServerConfig;
import net.jueb.redAlert.game.ServerMain;
import net.jueb.redAlert.game.factory.ScriptFactory;
import net.jueb.util4j.hotSwap.annationScriptFactory.annations.AnnationScript;

/**
 * 调试脚本
 * @author Administrator
 */
@AnnationScript(id = MsgCode.Game_Sys_Debug)
public class DebugScript extends AbstractSysScript{

	@Override
	public void action() {
		ClassLoader customLoader=ServerConfig.classProvider.getClassLoader();
		if(getClass().getClassLoader()==customLoader)
		{//this is hotswap script
			//test1 
			ByteBufAllocator.DEFAULT.ioBuffer();// init ThreadDeathWatcher watcher thread
			//or test2 ,use netty client write msg to init ThreadDeathWatcher
			ServerMain.getInstance().getLogClient().sendMessage(new AppByteMessage(AppByteMessage.Heart_Req));
			//so ,this script started netty thread
		}
	}
	public static void main(String[] args) {
		//run DebugScript action method
		ScriptFactory.getInstance().buildAction(MsgCode.Game_Sys_Debug).run();
	}
}`
### Minimal yet complete reproducer code (or URL to code)
![default](https://user-images.githubusercontent.com/5427272/31369054-5110b35e-adb4-11e7-9796-a4bb6b2a5b88.png)

### Netty version
4.1.6

juebanlin:
I hope you can initialize manually

Or pre initialization ThreadDeathWatcher

normanmaurer:
@juebanlin sorry but I not understand what exactly is the problem and how to reproduce. Can you elaborate ?

juebanlin:
@normanmaurer 

There is a custom classloader loaded script, if script created PooledByteBuf, may trigger to create a thread, see ThreadDeathWatcher schedule method script classloader, and then lead to the Unrecyclable, finally lead to the hotswap system is a memory leak

normanmaurer:
@juebanlin I still not understand it completely... Can you provide a reproducer ?

[...]

PR #7493, opened 2017-12-11
Ensure ThreadDeathWatcher and GlobalEventExecutor will not cause clas…

normanmaurer:
…sloader leaks.

Motivation:

ThreadDeathWatcher and GlobalEventExecutor may create and start a new thread from various other threads and so inherit the classloader. We need to ensure we not inherit to allow recycling the classloader.

Modifications:

Use Thread.setContextClassLoader(null) to ensure we not hold a strong reference to the classloader and so not leak it.

Result:

Fixes [#7290].


normanmaurer:
@juebanlin PTAL as wel

normanmaurer:
@ejona86 yep 😭

normanmaurer:
Cherry-picked into 4.1 (63bae0956aeb42e875ad303a2281989ad92acbf2) and 4.0 (11cfda3bbb)

PR #7494, opened 2017-12-11
Ability to scoop up events that reach the tail of the ChannelPipeline

### LINKED ISSUE #7484: Ability to define Head+Tail ChannelHandler
I'm working on a new Channel type that extends AbstractChannel and is in some ways a hybrid version of a LocalChannel or EmbeddedChannel. Its purpose is to act as a proxy between two other Channels (say EpollChannel or NioChannel) and also provide a disposable section of a ChannelPipeline with a proper EventLoop.

This is pretty straight forward and just requires handlers that scoop up the various events at the ends of the pipeline and to some extent it'd be the Channel itself (more on that later).

The problem is that there's no elegant way of ensuring the position of a handler and `addLast(...)` is in particular a pain in the neck as it's the goto method for users to manipulate the pipeline. Kind of defeating the purpose of trying to give users (of our Netty based software) a section of pipeline they can safely manipulate without breaking everything to the left (head) or right (tail) of it.

Technically I wouldn't want to introduce any new API as it's all there. It just needs to be a bit more extensible.

Idea 1: Change `AbstractChannel` to depend on `ChannelPipeline` instead of `DefaultChannelPipeline` (ditto `Unsafe` instead of `AbstractUnsafe`). This would allow me to use delegation/composition and override `addFirst/Last(...)` and ensure that users can't add their handlers before/after the required head/tail handlers. Not my favorite (final and concrete classes are IMHO overused in Netty ;)).

Idea 2: Change `DefaultChannelPipeline` and add a bunch of protected methods alongside `onUnhandledInboundMessage(...)` for all the other events that are currently being swallowed (say `channelReadComplete(...)` on the `TailContext` that I'd like to translate into a `flush()`, or `channelWritabilityChanged(...)` for managing back pressure).

Idea 3: Let ChannelPipeline fire a "pipeline modified event" every time a handler gets added or removed. My head and tail handlers could simply look for it in the `userEventTriggered(...)` method and move themselves if they're not head or tail. 

I think number 2 is my favorite as it's most aligned with the intent to scoop up the events that reach the ends of the pipeline. If I could I'd also break up AbstractChannel's dependency on concrete classes from number one but that is not needed in this context. Number 3 I'm not sure about. It's simple but has problems. For one I can create a loop if there are two competing handlers and I wouldn't want to introduce this new event for every Netty user out there if none of them cares about it.

rkapsi:
Motivation

There is currently no way to enforce the position of a handler in a ChannelPipeline and assume you wanted to write something like a custom Channel type that acts as a proxy between two other Channels.

ProxyChannel(Channel client, Channel server) {
  client calls write(msg) -> server.write(msg)
  client calls flush() -> server.flush()
  server calls fireChannelRead(msg) -> client.write(msg)
  server calls fireChannelReadComplete() -> client.flush()
}

In order to make it work reliably one needs to be able to scoop up the various events at the head and tail of the pipeline. The head side of the pipeline is covered by Unsafe and it's also relatively safe to count on the user to not use the addFirst() method to manipulate the pipeline. The tail side is always at a risk of getting broken because addLast() is the goto method to add handlers.

Modifications

Adding a few extra methods to DefaultChannelPipeline that expose some of the events that reach the pipeline's TailContext.

Result

Fixes #7484

### DIFF
--- a/transport/src/test/java/io/netty/channel/DefaultChannelPipelineTailTest.java
+++ b/transport/src/test/java/io/netty/channel/DefaultChannelPipelineTailTest.java
@@ -1,0 +1,408 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.channel;
+
+import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertTrue;
+
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import io.netty.bootstrap.Bootstrap;
+
+public class DefaultChannelPipelineTailTest {
+
+    private static EventLoopGroup GROUP;
+
+    @BeforeClass
+    public static void init() {
+        GROUP = new DefaultEventLoopGroup(1);
+    }
+
+    @AfterClass
+    public static void destroy() {
+        GROUP.shutdownGracefully();
+    }
+
+    @Test
+    public void testOnUnhandledInboundChannelActive() throws Exception {
+        final CountDownLatch latch = new CountDownLatch(1);
+        MyChannel myChannel = new MyChannel() {
+            @Override
+            protected void onUnhandledInboundChannelActive() {
+                latch.countDown();
+            }
+        };
+
+        Bootstrap bootstrap = new Bootstrap()
+                .channelFactory(new MyChannelFactory(myChannel))
+                .group(GROUP)
+                .handler(new ChannelInboundHandlerAdapter())
+                .remoteAddress(new InetSocketAddress(0));
+
+        Channel channel = bootstrap.connect()
+                .sync().channel();
+
+        try {
+            assertTrue(latch.await(1L, TimeUnit.SECONDS));
+        } finally {
+            channel.close();
+        }
+    }
+
+    @Test
+    public void testOnUnhandledInboundChannelInactive() throws Exception {
+        final CountDownLatch latch = new CountDownLatch(1);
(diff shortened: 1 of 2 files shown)


transport/src/test/java/io/netty/channel/DefaultChannelPipelineTailTest.java
```diff
@@ -0,0 +1,396 @@
...
+import java.util.concurrent.TimeUnit;
+
+import org.junit.AfterClass;
+import org.junit.Test;
+
+import io.netty.bootstrap.Bootstrap;
+
+public class DefaultChannelPipelineTailTest {
+
+    private static final EventLoopGroup GROUP = new DefaultEventLoopGroup();
```
normanmaurer: put this in a `@BeforeClass` ?

normanmaurer: Also you may just want to use 1 thread :)

normanmaurer:
@carl-mastrangelo @Scottmitch @ejona86 I would like to have your opinion on this.. I think this not really hurts and gives more flexibility ... So why not :)

transport/src/test/java/io/netty/channel/DefaultChannelPipelineTailTest.java
```diff
@@ -0,0 +1,402 @@
...
+        };
+
+        Bootstrap bootstrap = new Bootstrap()
+                .channelFactory(new MyChannelFactory(myChannel))
+                .group(GROUP)
+                .handler(new ChannelInboundHandlerAdapter())
+                .remoteAddress(new InetSocketAddress(0));
+
+        Channel channel = bootstrap.connect()
+                .syncUninterruptibly().channel();
```
carl-mastrangelo: sync() normally?  You do want to be interruptible

transport/src/test/java/io/netty/channel/DefaultChannelPipelineTailTest.java
```diff
@@ -0,0 +1,402 @@
...
+
+        try {
+            channel.pipeline().fireExceptionCaught(new IOException("testOnUnhandledInboundException"));
+            assertTrue(latch.await(1L, TimeUnit.SECONDS));
```
carl-mastrangelo: you might want to check that you got the right exception here.   Any exception, perhaps one you weren't expecting, would make this test pass.

normanmaurer:
Cherry-picked into 4.1 (79ed1c68714117ffa03aad6473504b4ed142593e). @rkapsi thanks!

RoganDawes:
If I may ask, is the intention that users should always extend the Channel implementation to be able to create their own variation of ChannelPipeline?

I am working on a generic proxy tool that should be able to handle all sorts of Channel implementations supported by Netty, and naturally have the same requirement as the original author, to be able to handle events that run off the end of the pipeline. Making users choose my extended version of NioChannel, etc, doesn't seem terribly clean to me.

An alternative of specifying a ChannelPipelineFactory as we had in v3, applying to the AbstractChannel class would solve this problem, with the default implementation simply constructing a new DefaultChannelPipeline every time.

I know that the design of ChannelPipelineFactory in v3 was problematic, because it lacked the capabilities of ChannelInitializer, but if all it is tasked with doing is constructing a "bare" DefaultChannelPipeline, that does not seem so bad to me?

PR #7499, opened 2017-12-13
Ensure Thread can be collected in a timely manner if Recycler.Stack h…

normanmaurer:
…olds a reference to it.

Motivation:

In our Recycler implementation we store a reference to the current Thread in the Stack that is stored in a FastThreadLocal. The Stack itself is referenced in the DefaultHandle itself. A problem can arise if a user stores a Reference to an Object that holds a reference to the DefaultHandle somewhere and either not remove the reference at all or remove it very late. In this case the Thread itself can not be collected as its still referenced in the Stack that is referenced by the DefaultHandle.

Modifications:

- Use a WeakReference to store the reference to the Thread in the Stack
- Add a test case

Result:

Ensure a Thread can be collected in a timely manner in all cases even if it used the Recycler.

### DIFF
--- a/common/src/test/java/io/netty/util/RecyclerTest.java
+++ b/common/src/test/java/io/netty/util/RecyclerTest.java
@@ -18,7 +18,9 @@ package io.netty.util;
 import org.junit.Test;
 
 import java.util.Random;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 
 import static org.junit.Assert.*;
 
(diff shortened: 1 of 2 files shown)


normanmaurer:
@ejona86 whatever we do with https://github.com/netty/netty/pull/7463 we should merge this one in all cases :)

common/src/test/java/io/netty/util/RecyclerTest.java
```diff
@@ -34,6 +36,43 @@ protected HandledObject newObject(
...
+        };
+        assertFalse(collected.get());
+        thread.start();
+        thread.join();
+
+        // Null out so it can be collected.
+        thread = null;
+
+        // Loop until the Thread was collected. If we can not collect it the Test will fail due of a timeout.
+        while (!collected.get()) {
```
Scottmitch: This test is a bit speculative, and dependent upon VM/GC behavior. Hopefully all GCs behaves as expected and will reclaim/finalize the object in time. I'm not sure of a better way to test this, but this maybe subject to sporadic test failures (maybe unlikely though?).

normanmaurer: @Scottmitch I think its the best we can do... If there will be ever a better API I am happy to revisit it.

normanmaurer:
Cherry-picked into 4.1 (0276b6e0f641687b145c810ab143be56143fc5ac) and 4.0 (b386ee3ea)

common/src/test/java/io/netty/util/RecyclerTest.java
Scottmitch: If it starts failing I guess we can take a closer look and consider disabling it if necessary.

PR #7500, opened 2017-12-13
Fix failing h2spec tests 8.1.2.1 related to pseudo-headers validation

### LINKED ISSUE #5761: HTTP/2: Many specification compliancy issues
I ran [h2spec](https://github.com/summerwind/h2spec) against the hello world HTTP/2 server (with removed HTTP upgrade logic). The result: 70 tests, 37 passed, 1 skipped, **32 failed**. 

From skimming over the list of errors, the failed tests seem to be mostly due to Netty responding with wrong error codes. [Here](https://gist.github.com/buchgr/9b8bc81592258008641e20e6d96f61c2) is a detailed list of what went wrong. 

cc: @Scottmitch @nmittler @louiscryan @ejona86

madgnome:
**Motivation:**

According to the [spec](http://httpwg.org/specs/rfc7540.html#HttpHeaders):

> All pseudo-header fields MUST appear in the header block before regular
> header fields. Any request or response that contains a pseudo-header
> field that appears in a header block after
> a regular header field MUST be treated as malformed (Section 8.1.2.6).

> Pseudo-header fields are only valid in the context in which they are defined.
> Pseudo-header fields defined for requests MUST NOT appear in responses;
> pseudo-header fields defined for responses MUST NOT appear in requests.
> Pseudo-header fields MUST NOT appear in trailers.
> Endpoints MUST treat a request or response that contains undefined or
> invalid pseudo-header fields as malformed (Section 8.1.2.6).

> Clients MUST NOT accept a malformed response. Note that these requirements
> are intended to protect against several types of common attacks against HTTP;
> they are deliberately strict because being permissive can expose
> implementations to these vulnerabilities.

**Modifications:**

- Introduce new constructor to DefaultHttp2Headers allowing
  header insertion order to be preserved (useful for testing)
- Assign a PseudoHeaderType (REQUEST|RESPONSE) to PseudoHeaderName
- Group variables related to header decoding to the inner class
  HeadersDecodingState
- Add pseudo-header validation to DefaultHttp2HeaderDecoder

**Result:**

- Requests with unknown pseudo-field headers are rejected
- Requests with containing response specific pseudo-headers are rejected
- Requests where pseudo-header appear after regular header are rejected
- Fixes #5761 tests 8.1.2.1

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Headers.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Headers.java
@@ -16,6 +16,7 @@ package io.netty.handler.codec.http2;
 
 import static io.netty.handler.codec.http2.Http2Error.PROTOCOL_ERROR;
 import static io.netty.handler.codec.http2.Http2Exception.connectionError;
+import static io.netty.handler.codec.http2.Http2Headers.PseudoHeaderName.hasPseudoHeaderFormat;
 import static io.netty.util.AsciiString.CASE_SENSITIVE_HASHER;
 import static io.netty.util.AsciiString.isUpperCase;
 import io.netty.handler.codec.CharSequenceValueConverter;
@@ -30,7 +31,7 @@ public class DefaultHttp2Headers
         extends DefaultHeaders<CharSequence, CharSequence, Http2Headers> implements Http2Headers {
     private static final ByteProcessor HTTP2_NAME_VALIDATOR_PROCESSOR = new ByteProcessor() {
         @Override
-        public boolean process(byte value) throws Exception {
+        public boolean process(byte value) {
             return !isUpperCase(value);
         }
     };
@@ -197,7 +198,7 @@ public class DefaultHttp2Headers
             this.next = next;
 
             // Make sure the pseudo headers fields are first in iteration order
-            if (key.length() != 0 && key.charAt(0) == ':') {
+            if (hasPseudoHeaderFormat(key)) {
                 after = firstNonPseudo;
                 before = firstNonPseudo.before();
             } else {
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
@@ -115,7 +115,7 @@ public class DefaultHttp2HeadersDecoder implements Http2HeadersDecoder, Http2Hea
     public Http2Headers decodeHeaders(int streamId, ByteBuf headerBlock) throws Http2Exception {
         try {
             final Http2Headers headers = newHeaders();
-            hpackDecoder.decode(streamId, headerBlock, headers);
+            hpackDecoder.decode(streamId, headerBlock, headers, validateHeaders);
             headerArraySizeAccumulator = HEADERS_COUNT_WEIGHT_NEW * headers.size() +
                                          HEADERS_COUNT_WEIGHT_HISTORICAL * headerArraySizeAccumulator;
             return headers;
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
@@ -45,6 +45,8 @@ import static io.netty.handler.codec.http2.Http2Error.COMPRESSION_ERROR;
 import static io.netty.handler.codec.http2.Http2Error.INTERNAL_ERROR;
 import static io.netty.handler.codec.http2.Http2Error.PROTOCOL_ERROR;
 import static io.netty.handler.codec.http2.Http2Exception.connectionError;
+import static io.netty.handler.codec.http2.Http2Headers.PseudoHeaderName.getPseudoHeader;
+import static io.netty.handler.codec.http2.Http2Headers.PseudoHeaderName.hasPseudoHeaderFormat;
 import static io.netty.util.AsciiString.EMPTY_STRING;
 import static io.netty.util.internal.ObjectUtil.checkPositive;
 import static io.netty.util.internal.ThrowableUtil.unknownStackTrace;
@@ -119,7 +121,7 @@ final class HpackDecoder {
      * <p>
      * This method assumes the entire header block is contained in {@code in}.
      */
-    public void decode(int streamId, ByteBuf in, Http2Headers headers) throws Http2Exception {
+    public void decode(int streamId, ByteBuf in, Http2Headers headers, boolean validateHeaders) throws Http2Exception {
         int index = 0;
         long headersLength = 0;
         int nameLength = 0;
@@ -127,6 +129,7 @@ final class HpackDecoder {
         byte state = READ_HEADER_REPRESENTATION;
         boolean huffmanEncoded = false;
         CharSequence name = null;
+        HeaderType headerType = null;
         IndexType indexType = IndexType.NONE;
         while (in.isReadable()) {
             switch (state) {
@@ -146,7 +149,10 @@ final class HpackDecoder {
                                 state = READ_INDEXED_HEADER;
                                 break;
                             default:
-                                headersLength = indexHeader(index, headers, headersLength);
+                                HpackHeaderField indexedHeader = getIndexedHeader(index);
+                                headerType = validate(indexedHeader.name, headerType, validateHeaders);
+                                headersLength = addHeader(headers, indexedHeader.name, indexedHeader.value,
+                                        headersLength);
                         }
                     } else if ((b & 0x40) == 0x40) {
                         // Literal Header Field with Incremental Indexing
@@ -162,6 +168,7 @@ final class HpackDecoder {
                             default:
                                 // Index was stored as the prefix
                                 name = readName(index);
+                                headerType = validate(name, headerType, validateHeaders);
                                 nameLength = name.length();
                                 state = READ_LITERAL_HEADER_VALUE_LENGTH_PREFIX;
                         }
@@ -188,6 +195,7 @@ final class HpackDecoder {
                             default:
                             // Index was stored as the prefix
                             name = readName(index);
+                                headerType = validate(name, headerType, validateHeaders);
                             nameLength = name.length();
                             state = READ_LITERAL_HEADER_VALUE_LENGTH_PREFIX;
                         }
@@ -200,13 +208,16 @@ final class HpackDecoder {
                     break;
 
                 case READ_INDEXED_HEADER:
-                    headersLength = indexHeader(decodeULE128(in, index), headers, headersLength);
+                    HpackHeaderField indexedHeader = getIndexedHeader(decodeULE128(in, index));
+                    headerType = validate(indexedHeader.name, headerType, validateHeaders);
+                    headersLength = addHeader(headers, indexedHeader.name, indexedHeader.value, headersLength);
                     state = READ_HEADER_REPRESENTATION;
                     break;
 
                 case READ_INDEXED_HEADER_NAME:
                     // Header Name matches an entry in the Header Table
                     name = readName(decodeULE128(in, index));
+                    headerType = validate(name, headerType, validateHeaders);
                     nameLength = name.length();
                     state = READ_LITERAL_HEADER_VALUE_LENGTH_PREFIX;
                     break;
@@ -243,6 +254,7 @@ final class HpackDecoder {
                     }
 
                     name = readStringLiteral(in, nameLength, huffmanEncoded);
+                    headerType = validate(name, headerType, validateHeaders);
 
                     state = READ_LITERAL_HEADER_VALUE_LENGTH_PREFIX;
                     break;
@@ -256,6 +268,7 @@ final class HpackDecoder {
                             state = READ_LITERAL_HEADER_VALUE_LENGTH;
                             break;
                         case 0:
+                            headerType = validate(name, headerType, validateHeaders);
                             headersLength = insertHeader(headers, name, EMPTY_STRING, indexType, headersLength);
                             state = READ_HEADER_REPRESENTATION;
                             break;
@@ -288,6 +301,7 @@ final class HpackDecoder {
                     }
 
                     CharSequence value = readStringLiteral(in, valueLength, huffmanEncoded);
+                    headerType = validate(name, headerType, validateHeaders);
                     headersLength = insertHeader(headers, name, value, indexType, headersLength);
                     state = READ_HEADER_REPRESENTATION;
                     break;
@@ -386,6 +400,34 @@ final class HpackDecoder {
         hpackDynamicTable.setCapacity(dynamicTableSize);
     }
 
+    private HeaderType validate(CharSequence name, HeaderType previousHeaderType,
+                                final boolean validateHeaders) throws Http2Exception {
+        if (!validateHeaders) {
+            return null;
+        }
+
+        if (hasPseudoHeaderFormat(name)) {
+            if (previousHeaderType == HeaderType.REGULAR_HEADER) {
+                throw connectionError(PROTOCOL_ERROR, "Pseudo-header field '%s' found after regular header.", name);
+            }
+
+            final Http2Headers.PseudoHeaderName pseudoHeader = getPseudoHeader(name);
+            if (pseudoHeader == null) {
+                throw connectionError(PROTOCOL_ERROR, "Invalid HTTP/2 pseudo-header '%s' encountered.", name);
+            }
+
+            final HeaderType currentHeaderType = pseudoHeader.isRequestOnly() ?
+                    HeaderType.REQUEST_PSEUDO_HEADER : HeaderType.RESPONSE_PSEUDO_HEADER;
+            if (previousHeaderType != null && currentHeaderType != previousHeaderType) {
+                throw connectionError(PROTOCOL_ERROR, "Mix of request and response pseudo-headers.");
+            }
+
+            return currentHeaderType;
+        }
+
+        return HeaderType.REGULAR_HEADER;
+    }
+
     private CharSequence readName(int index) throws Http2Exception {
         if (index <= HpackStaticTable.length) {
             HpackHeaderField hpackHeaderField = HpackStaticTable.getEntry(index);
@@ -398,14 +440,12 @@ final class HpackDecoder {
         throw READ_NAME_ILLEGAL_INDEX_VALUE;
     }
 
-    private long indexHeader(int index, Http2Headers headers, long headersLength) throws Http2Exception {
+    private HpackHeaderField getIndexedHeader(int index) throws Http2Exception {
         if (index <= HpackStaticTable.length) {
-            HpackHeaderField hpackHeaderField = HpackStaticTable.getEntry(index);
-            return addHeader(headers, hpackHeaderField.name, hpackHeaderField.value, headersLength);
+            return HpackStaticTable.getEntry(index);
         }
         if (index - HpackStaticTable.length <= hpackDynamicTable.length()) {
-            HpackHeaderField hpackHeaderField = hpackDynamicTable.getEntry(index - HpackStaticTable.length);
-            return addHeader(headers, hpackHeaderField.name, hpackHeaderField.value, headersLength);
+            return hpackDynamicTable.getEntry(index - HpackStaticTable.length);
         }
         throw INDEX_HEADER_ILLEGAL_INDEX_VALUE;
     }
@@ -503,5 +543,14 @@ final class HpackDecoder {
         }
 
         throw DECODE_ULE_128_DECOMPRESSION_EXCEPTION;
+    }
+
+    /**
+     * HTTP/2 header types.
+     */
+    private enum HeaderType {
+        REGULAR_HEADER,
+        REQUEST_PSEUDO_HEADER,
+        RESPONSE_PSEUDO_HEADER
     }
 }
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Headers.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Headers.java
@@ -35,38 +35,44 @@ public interface Http2Headers extends Headers<CharSequence, CharSequence, Http2H
         /**
          * {@code :method}.
          */
-        METHOD(":method"),
+        METHOD(":method", true),
 
         /**
          * {@code :scheme}.
          */
-        SCHEME(":scheme"),
+        SCHEME(":scheme", true),
 
         /**
          * {@code :authority}.
          */
-        AUTHORITY(":authority"),
+        AUTHORITY(":authority", true),
 
         /**
          * {@code :path}.
          */
-        PATH(":path"),
+        PATH(":path", true),
 
         /**
          * {@code :status}.
          */
-        STATUS(":status");
+        STATUS(":status", false);
+
+        private static final char PSEUDO_HEADER_PREFIX = ':';
+        private static final byte PSEUDO_HEADER_PREFIX_BYTE = (byte) PSEUDO_HEADER_PREFIX;
 
         private final AsciiString value;
-        private static final CharSequenceMap<AsciiString> PSEUDO_HEADERS = new CharSequenceMap<AsciiString>();
+        private final boolean requestOnly;
+        private static final CharSequenceMap<PseudoHeaderName> PSEUDO_HEADERS = new CharSequenceMap<PseudoHeaderName>();
+
         static {
             for (PseudoHeaderName pseudoHeader : PseudoHeaderName.values()) {
-                PSEUDO_HEADERS.add(pseudoHeader.value(), AsciiString.EMPTY_STRING);
+                PSEUDO_HEADERS.add(pseudoHeader.value(), pseudoHeader);
             }
         }
 
-        PseudoHeaderName(String value) {
+        PseudoHeaderName(String value, boolean requestOnly) {
             this.value = AsciiString.cached(value);
+            this.requestOnly = requestOnly;
         }
 
         public AsciiString value() {
@@ -75,10 +81,42 @@ public interface Http2Headers extends Headers<CharSequence, CharSequence, Http2H
         }
 
         /**
+         * Indicates whether the specified header follows the pseudo-header format (begins with ':' character)
+         *
+         * @return {@code true} if the header follow the pseudo-header format
+         */
+        public static boolean hasPseudoHeaderFormat(CharSequence headerName) {
+            if (headerName instanceof AsciiString) {
+                final AsciiString asciiHeaderName = (AsciiString) headerName;
+                return asciiHeaderName.length() > 0 && asciiHeaderName.byteAt(0) == PSEUDO_HEADER_PREFIX_BYTE;
+            } else {
+                return headerName.length() > 0 && headerName.charAt(0) == PSEUDO_HEADER_PREFIX;
+            }
+        }
+
+        /**
          * Indicates whether the given header name is a valid HTTP/2 pseudo header.
          */
         public static boolean isPseudoHeader(CharSequence header) {
             return PSEUDO_HEADERS.contains(header);
+        }
+
+        /**
+         * Returns the {@link PseudoHeaderName} corresponding to the specified header name.
+         *
+         * @return corresponding {@link PseudoHeaderName} if any, {@code null} otherwise.
+         */
+        public static PseudoHeaderName getPseudoHeader(CharSequence header) {
+            return PSEUDO_HEADERS.get(header);
+        }
+
+        /**
+         * Indicates whether the pseudo-header is to be used in a request context.
+         *
+         * @return {@code true} if the pseudo-header is to be used in a request context
+         */
+        public boolean isRequestOnly() {
+            return requestOnly;
         }
     }
 
--- a/codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
+++ b/codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
@@ -71,7 +71,7 @@ public class HpackDecoderTest {
         byte[] b = StringUtil.decodeHexDump(encoded);
         ByteBuf in = Unpooled.wrappedBuffer(b);
         try {
-            hpackDecoder.decode(0, in, mockHeaders);
+            hpackDecoder.decode(0, in, mockHeaders, true);
         } finally {
             in.release();
         }
@@ -167,7 +167,7 @@ public class HpackDecoderTest {
         try {
             final long expectedHeaderSize = 4026531870L; // based on the input above
             hpackDecoder.setMaxHeaderTableSize(expectedHeaderSize);
-            hpackDecoder.decode(0, in, mockHeaders);
+            hpackDecoder.decode(0, in, mockHeaders, true);
             assertEquals(expectedHeaderSize, hpackDecoder.getMaxHeaderTableSize());
         } finally {
             in.release();
@@ -180,7 +180,7 @@ public class HpackDecoderTest {
         ByteBuf in = Unpooled.wrappedBuffer(input);
         try {
             hpackDecoder.setMaxHeaderTableSize(4026531870L - 1); // based on the input above ... 1 less than is above.
-            hpackDecoder.decode(0, in, mockHeaders);
+            hpackDecoder.decode(0, in, mockHeaders, true);
         } finally {
             in.release();
         }
@@ -191,7 +191,7 @@ public class HpackDecoderTest {
         byte[] input = {0, (byte) 0x80, 0};
         ByteBuf in = Unpooled.wrappedBuffer(input);
         try {
-            hpackDecoder.decode(0, in, mockHeaders);
+            hpackDecoder.decode(0, in, mockHeaders, true);
             verify(mockHeaders, times(1)).add(EMPTY_STRING, EMPTY_STRING);
         } finally {
             in.release();
@@ -203,7 +203,7 @@ public class HpackDecoderTest {
         byte[] input = {0, (byte) 0x81, -1};
         ByteBuf in = Unpooled.wrappedBuffer(input);
         try {
-            hpackDecoder.decode(0, in, mockHeaders);
+            hpackDecoder.decode(0, in, mockHeaders, true);
         } finally {
             in.release();
         }
@@ -214,7 +214,7 @@ public class HpackDecoderTest {
         byte[] input = {0, (byte) 0x84, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF};
         ByteBuf in = Unpooled.wrappedBuffer(input);
         try {
-            hpackDecoder.decode(0, in, mockHeaders);
+            hpackDecoder.decode(0, in, mockHeaders, true);
         } finally {
             in.release();
         }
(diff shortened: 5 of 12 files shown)


codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Headers.java
```diff
@@ -70,6 +70,7 @@ public void validateName(CharSequence name) {
     };
 
     private HeaderEntry<CharSequence, CharSequence> firstNonPseudo = head;
+    private boolean keepInsertionOrder;
```
normanmaurer: final ?

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
```diff
@@ -498,4 +477,46 @@ static long decodeULE128(ByteBuf in, long result) throws Http2Exception {
...
     }
+
+    private final class HeadersDecodingState {
+        private boolean sawRegularHeader;
+        private long headersLength;
+
+        private final Http2Headers headers;
+        private final int streamId;
+
+        private HeadersDecodingState(final int streamId, final Http2Headers headers) {
```
normanmaurer: use a package-private constructor if you call it from outer-class

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
```diff
@@ -498,4 +477,46 @@ static long decodeULE128(ByteBuf in, long result) throws Http2Exception {
...
+            this.streamId = streamId;
+            this.headers = headers;
+        }
+
+        private void addHeader(CharSequence name, CharSequence value) throws Http2Exception {
```
normanmaurer: make package-private if called from outclass.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Headers.java
```diff
@@ -22,64 +22,89 @@
...
             this.value = AsciiString.cached(value);
+            this.type = type;
         }
 
         public AsciiString value() {
             // Return a slice so that the buffer gets its own reader index.
             return value;
         }
 
+        public PseudoHeaderType getType() {
```
normanmaurer: remove `get`  as we also not have it on value etc.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Headers.java
```diff
@@ -22,64 +22,89 @@
...
          */
         public static boolean isPseudoHeader(CharSequence header) {
             return PSEUDO_HEADERS.contains(header);
         }
+
+        public static PseudoHeaderName getPseudoHeader(CharSequence header) {
+            return PSEUDO_HEADERS.get(header);
+        }
+
+        public static boolean hasPseudoHeaderFormat(CharSequence header) {
```
normanmaurer: add java docs to the two methods.

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
```diff
@@ -498,4 +477,46 @@ static long decodeULE128(ByteBuf in, long result) throws Http2Exception {
 
         throw DECODE_ULE_128_DECOMPRESSION_EXCEPTION;
     }
+
+    private final class HeadersDecodingState {
```
normanmaurer: static ?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
```diff
@@ -152,4 +162,29 @@ protected final boolean validateHeaders() {
...
+        for (final Map.Entry<CharSequence, CharSequence> header : headers) {
+            final CharSequence key = header.getKey();
+            if (hasPseudoHeaderFormat(key)) {
+                final Http2Headers.PseudoHeaderName pseudoHeader = getPseudoHeader(key);
+                if (pseudoHeader == null) {
+                    throw connectionError(PROTOCOL_ERROR, "Invalid HTTP/2 pseudo-header '%s' encountered", key);
+                }
+
+                if (pseudoHeader.getType() == Http2Headers.PseudoHeaderType.REQUEST) {
+                    isRequest = true;
```
normanmaurer: just replace this by:

```java
if (isResponse) {
    throw connectionError(PROTOCOL_ERROR, "Mix of request and response pseudo-headers.");
}
```

This way you can fast fail the loop.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
```diff
@@ -152,4 +162,29 @@ protected final boolean validateHeaders() {
...
+                } else {
+                    isResponse = true;
```
normanmaurer: just replace this by:

```java
if (isRequest) {
    throw connectionError(PROTOCOL_ERROR, "Mix of request and response pseudo-headers.");
}
```

This way you can fast fail the loop.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
```diff
@@ -152,4 +162,29 @@ protected final boolean validateHeaders() {
...
+                }
+            }
+        }
+
+        if (isRequest && isResponse) {
```
normanmaurer: This can be removed once you addressed the comments above

normanmaurer:
@madgnome thanks for the PR... I left a few comments.

Also please sign our ICLA:

https://netty.io/s/icla

madgnome:
ICLA signed.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
```diff
@@ -152,4 +162,35 @@ protected final boolean validateHeaders() {
...
+
+        for (final Map.Entry<CharSequence, CharSequence> header : headers) {
+            final CharSequence key = header.getKey();
+            if (hasPseudoHeaderFormat(key)) {
+                final Http2Headers.PseudoHeaderName pseudoHeader = getPseudoHeader(key);
+                if (pseudoHeader == null) {
+                    throw connectionError(PROTOCOL_ERROR, "Invalid HTTP/2 pseudo-header '%s' encountered", key);
+                }
+
+                if (pseudoHeader.type() == Http2Headers.PseudoHeaderType.REQUEST) {
```
fenik17: This can be simplified:
```
Http2Headers.PseudoHeaderType prevType = null;
for (...) {
   ...
   if (pseudoHeader.type() != prevType)) {
      requestAndResponsePseudoHeadersMixed();
   }
   prevType = pseudoHeader.type();
}

**Comment by normanmaurer** [comment]
@madgnome I guess with this we can also re-enable the test here: https://github.com/netty/netty/blob/4.1/testsuite-http2/pom.xml#L76 ?

**Comment by madgnome** [comment]
Indeed! will update.

**Comment by normanmaurer** [comment]
@mosesn PTAL as well

**Comment by mosesn** [comment]
Migrating to hard fail when connection headers are out of order makes me a bit nervous, but we can back it out and figure out a migration strategy if people try this in the wild and discover it's actually an issue.  🚢 🇮🇹

File: codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Headers.java [hunk 11]
```diff
@@ -70,6 +70,7 @@ public void validateName(CharSequence name) {
     };
 
     private HeaderEntry<CharSequence, CharSequence> firstNonPseudo = head;
+    private final boolean keepInsertionOrder;
```
Scottmitch: I would prefer to not have this in production code if it is only to support testing. I would prefer to isolate this to the unit tests if possible.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
```diff
@@ -152,4 +162,24 @@ protected final boolean validateHeaders() {
     protected Http2Headers newHeaders() {
         return new DefaultHttp2Headers(validateHeaders, (int) headerArraySizeAccumulator);
     }
+
+    protected void validate(Http2Headers headers) throws Http2Exception {
+        Http2Headers.PseudoHeaderType prevType = null;
```
Scottmitch: in addition to verifying the type is the same as the previous, shouldn't we also provide an "expected type" to satisfy the following requirements:

> Pseudo-header fields defined for requests MUST NOT appear in responses;
pseudo-header fields defined for responses MUST NOT appear in requests.

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
```diff
@@ -121,13 +122,15 @@
...
         int index = 0;
-        long headersLength = 0;
         int nameLength = 0;
         int valueLength = 0;
         byte state = READ_HEADER_REPRESENTATION;
         boolean huffmanEncoded = false;
         CharSequence name = null;
         IndexType indexType = IndexType.NONE;
+
+        HeadersDecodingState headersDecodingState = new HeadersDecodingState(streamId, headers,
```
Scottmitch: can we avoid addition object allocation in the headers decode operation? This maybe a frequently called operation and if we can avoid intermediate allocations that would be preferred.

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
```diff
@@ -498,4 +478,51 @@ static long decodeULE128(ByteBuf in, long result) throws Http2Exception {
...
+
+        void addHeader(CharSequence name, CharSequence value) throws Http2Exception {
+            long newHeaderLength = this.headersLength + name.length() + value.length();
+            if (newHeaderLength > maxHeaderListSizeGoAway) {
+                headerListSizeExceeded(maxHeaderListSizeGoAway);
+            }
+
+            boolean pseudoHeaderFormat = hasPseudoHeaderFormat(name);
+            if (sawRegularHeader && pseudoHeaderFormat) {
+                throw connectionError(PROTOCOL_ERROR, "Pseudo-header field found after regular header.");
```
Scottmitch: It seems like we validate the pseudo header requirements in `DefaultHttp2HeadersDecoder` and in here in `HpackDecoder`. Can we validate in one or the other instead of both?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Headers.java
madgnome: Yeah, wasn't sold on that either. 

Other solutions are:
* Duplicate DefaultHttp2Headers in test code
* Remove `final` on `DefaultHttp2Headers#newHeaderEntry` and extends it in test to implement an in insertion order iterator.

Any preference?

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
madgnome: Agree, I wanted to put everything into `DefaultHttp2HeadersDecoder` but the way `DefaultHttp2Headers` works we lose the insertion order. 

Do you think it would be better to have everything in `HpackDecoder` or should I find a way to retrieve insertion order from DefaultHttp2Headers?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
madgnome: I'm going through h2spec failing tests and trying to fix them. The following tests should cover that

```
            <excludeSpec>8.1.2.3 - Sends a HEADERS frame with empty ":path" pseudo-header field</excludeSpec>
            <excludeSpec>8.1.2.3 - Sends a HEADERS frame that omits ":method" pseudo-header field</excludeSpec>
            <excludeSpec>8.1.2.3 - Sends a HEADERS frame that omits ":scheme" pseudo-header field</excludeSpec>
            <excludeSpec>8.1.2.3 - Sends a HEADERS frame that omits ":path" pseudo-header field</excludeSpec>
```

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Headers.java
normanmaurer: @madgnome I thin I would go with the "removing final".

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
normanmaurer: @madgnome @Scottmitch  +1 I think we can fix these step by step

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
normanmaurer: +1... would be nice if we could just store these directly

madgnome:
Ok, I've changed things a bit:

- I've undo all my changes to `HPackDecoder` 
- I've moved the validation for out of order headers to `DefaultHttp2HeadersDecoder` (now all the validation happen in the same place 🎆 )
-  For this to work it uses `keepInsertionOrder` in `DefaultHttp2Headers`. Another solution would be to create a `DecodedHttp2Headers` extending `DefaultHttp2Headers` which keep insertion order.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
Scottmitch: I brought it up because it is mentioned in the commit message. If the intention is not to fix this in the current PR, please remove this text from the commit message.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Headers.java
Scottmitch: Removing final sounds OK to me ... the question is how much would we have to duplicate in the tests? It sounds like what you want todo is different than what `DefaultHttp2Headers` is designed todo (enforce HTTP/2 constraints). `DefaultHttp2Headers` also doesn't provide much code besides the `Http2HeaderEntry` (which it looks like would be replaced anyways) and `HTTP2_NAME_VALIDATOR` (which is package private and maybe we can share?). So maybe duplication isn't too bad here?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
madgnome: We check that pseudo headers are not mixed (no request header with response header or vice versa ) so it does meet the requirement in a way

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Headers.java
madgnome: I've changed the implementation of `DefaultHttp2HeadersDecoder` to instantiate `DefaultHttp2Headers` with `keepInsertionOrder` to `true`, doing that allows us to move all the validation to `DefaultHttp2HeadersDecoder` and to keep `HPackDecoder` untouched.

If you think `keepInsertionOrder` still shouldn't exist, I could implement a `DecoderHttp2Headers` class that would be use by the decoder to do that.

Scottmitch: IIUC `keepInsertionOrder` is still only necessary for tests. If this is the case I would prefer to remove it and isolate this behavior to unit tests.

madgnome: Oh damn, good catch. I was a bit too eager cleaning some code before commiting and remove the important part

```
protected Http2Headers newHeaders() {
        return new DefaultHttp2Headers(validateHeaders, (int) headerArraySizeAccumulator, true);
    }
```

`DefaultHttp2HeadersDecoder` will use `keepInsertionOrder`

Scottmitch: ping me when behavior has been isolated to unit tests and I will take another look

madgnome: You can see there https://github.com/netty/netty/pull/7500/files#diff-32bd1e577a26442a725a78acaf5822daR163 how we use `keepInsertionOrder` in prod code.

(Tests are back green)

madgnome: @Scottmitch are you happy with `keepInsertionOrder` in `DefaultHttp2Headers` now that it's used by production code?

Otherwise I'll create an `DecoderHttp2Headers` extending `DefaultHttp2Headers` which keep insertion order and use that in the decoder.

mosesn:
This approach seems reasonable to me.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
```diff
@@ -150,6 +160,33 @@ protected final boolean validateHeaders() {
      * @return a new {@link Http2Headers} object which will store the results of the decode operation.
      */
     protected Http2Headers newHeaders() {
-        return new DefaultHttp2Headers(validateHeaders, (int) headerArraySizeAccumulator);
+        return new DefaultHttp2Headers(validateHeaders, (int) headerArraySizeAccumulator, true);
```
Scottmitch: note that headers are modifiable by default. So after the headers are decoded they maybe modified and conceivably add another pseudo header (e.g. a proxy use case). I'm hesitant to generate a `Http2Headers` objects which does not enforce the iteration order required by the specification under these constraints. IMHO it would be safer just to isolate this behavior to the unit tests if that is where it is required.

normanmaurer: @Scottmitch @madgnome +1

madgnome: Oh damn, excellent point.

> IMHO it would be safer just to isolate this behavior to the unit tests if that is where it is required.

We need to know the insertion order to be able to detect if the headers are out of order when we decode them. 

Back to the drawing board...

madgnome:
Ok, I gave it another go.

Changes compared to previous commits:

- Validation is done entirely in `HPackDecoder`. I tried my best to limit object allocation
- New `InOrderHttp2Headers` class in test package to help with testing (No modification to existing `DefaultHttp2Headers` behaviour

normanmaurer:
@ejona86 @carl-mastrangelo any comment ?

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
```diff
@@ -382,6 +396,28 @@ private void setDynamicTableSize(long dynamicTableSize) throws Http2Exception {
         hpackDynamicTable.setCapacity(dynamicTableSize);
     }
 
+    private Http2Headers.HeaderType validate(CharSequence name, Http2Headers.HeaderType previousHeaderType)
+        throws Http2Exception {
+        if (hasPseudoHeaderFormat(name)) {
```
Scottmitch: Netty typically allows users to opt out of header validation which may otherwise add overhead. For example in a "trusted" scenario it may be desirable to avoid the additional overhead (map lookup, string inspection, etc...). I would suggest we propagate the `validateHeaders` parameter in `DefaultHttp2HeadersDecoder` to this class and conditionally do the validation.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Headers.java
```diff
@@ -35,51 +46,78 @@
...
             this.value = AsciiString.cached(value);
+            this.type = type;
         }
 
         public AsciiString value() {
             // Return a slice so that the buffer gets its own reader index.
             return value;
         }
 
+        public HeaderType type() {
```
Scottmitch: does `HeaderType` need to be public? From the perspective of a `PseudoHeaderName` it is only allowed in either `request` or `response`, correct? If that is the case can we just have a method like this:

```java
// true -> the pseudo header can only be used in requests, false -> the pseudo header can only be used in responses
public boolean isRequestOnly() {
}
```

codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
```diff
@@ -512,4 +512,91 @@ public void testAccountForHeaderOverhead() throws Exception {
...
+        ByteBuf in = Unpooled.buffer(200);
+        try {
+            HpackEncoder hpackEncoder = new HpackEncoder(true);
+
+            Http2Headers toEncode = new DefaultHttp2Headers();
+            toEncode.add(":test", "1");
+            hpackEncoder.encodeHeaders(1, in, toEncode, NEVER_SENSITIVE);
+
+            // decode the headers, we should get an exception, but
+            // the decoded headers object should contain all of the headers
```
Scottmitch: this makes it sound like you would expect `decoded` to contain `:test=1` after the decode operation, but that is not the case. Can you clarify this comment?

codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
```diff
@@ -512,4 +512,91 @@ public void testAccountForHeaderOverhead() throws Exception {
...
+            Http2Headers decoded = new DefaultHttp2Headers();
+
+            expectedException.expect(Http2Exception.class);
+            hpackDecoder.decode(1, in, decoded);
+        } finally {
+            in.release();
```
Scottmitch: should we verify that `decoded` contains `:status=200`?

codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
```diff
@@ -512,4 +512,91 @@ public void testAccountForHeaderOverhead() throws Exception {
...
```
Scottmitch: should we verify that `decoded` contains `:method=200`?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Headers.java
```diff
@@ -35,51 +46,78 @@
...
+            return PSEUDO_HEADERS.get(header);
+        }
+
+        /**
+         * Indicates whether the specified header follows the pseudo-header format (begins with ':' character)
+         *
+         * @return {@code true} if the header follow the pseudo-header format
+         */
+        public static boolean hasPseudoHeaderFormat(CharSequence header) {
+            return header.length() > 0 && header.charAt(0) == PSEUDO_HEADER_PREFIX;
```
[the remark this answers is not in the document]
carl-mastrangelo: Another idea: skip the header.length() check and just catch the exception on index out of bounds.  Most headers will not need the check.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Headers.java
madgnome: `HeaderType` can be `REGULAR_HEADER` , `REQUEST_PSEUDO_HEADER` and `RESPONSE_PSEUDO_HEADER`, we use it to track the current header type in the decoder to help with validation. 

We can make `HeaderType` package private or if we absolutely don't want it, we could try to rely on `Http2Headers.PseudoHeaderName` only but that would make the code more confusing I think.

codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
madgnome: I think it's only necessary if we expect consumers to use the `Http2Headers` after a failed decoding attempt. It's currently not the case in `DefaultHttp2HeadersDecoder`

codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
madgnome: - [X] Remove comment

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Headers.java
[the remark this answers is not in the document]
madgnome: - [X] Add optimisation for AsciiString

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
madgnome: - [X] Propage `validateHeaders`

codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
Scottmitch: sgtm

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Headers.java
```diff
@@ -35,51 +46,86 @@
...
-        PATH(":path"),
+        PATH(":path", REQUEST_PSEUDO_HEADER),
 
         /**
          * {@code :status}.
          */
-        STATUS(":status");
+        STATUS(":status", RESPONSE_PSEUDO_HEADER);
+
+        private static char PSEUDO_HEADER_PREFIX = ':';
```
Scottmitch: make this variable `final`.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Headers.java
madgnome: Yeah makes perfect sense, I'll move `HeaderType` to `HPackDecoder` .

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Headers.java
```diff
@@ -35,51 +46,86 @@
...
+
+        /**
+         * Indicates whether the specified header follows the pseudo-header format (begins with ':' character)
+         *
+         * @return {@code true} if the header follow the pseudo-header format
+         */
+        public static boolean hasPseudoHeaderFormat(CharSequence header) {
+            try {
+                if (header instanceof AsciiString) {
+                    return ((AsciiString) header).byteAt(0) == PSEUDO_HEADER_PREFIX;
```
Scottmitch: consider adding a new `private static final byte PSEUDO_HEADER_PREFIX_BYTE = (byte) ';';` and using it here to avoid any casting.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Headers.java
```diff
@@ -35,51 +46,86 @@
...
+        public static PseudoHeaderName getPseudoHeader(CharSequence header) {
+            return PSEUDO_HEADERS.get(header);
+        }
+
+        /**
+         * Indicates whether the specified header follows the pseudo-header format (begins with ':' character)
+         *
+         * @return {@code true} if the header follow the pseudo-header format
+         */
+        public static boolean hasPseudoHeaderFormat(CharSequence header) {
```
Scottmitch: nit: `header` -> `headerName`?

Scottmitch: potential optimization: we should use this in [DefaultHttp2Headers](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Headers.java#L200) and use [HeadersBenchmark](https://github.com/netty/netty/blob/4.1/microbench/src/main/java/io/netty/microbench/headers/HeadersBenchmark.java#L51) to ensure there is benefit of this approach.

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
```diff
@@ -398,14 +438,12 @@ private CharSequence readName(int index) throws Http2Exception {
...
+    private HpackHeaderField getIndexedHeader(int index) throws Http2Exception {
         if (index <= HpackStaticTable.length) {
-            HpackHeaderField hpackHeaderField = HpackStaticTable.getEntry(index);
-            return addHeader(headers, hpackHeaderField.name, hpackHeaderField.value, headersLength);
+            return HpackStaticTable.getEntry(index);
         }
         if (index - HpackStaticTable.length <= hpackDynamicTable.length()) {
-            HpackHeaderField hpackHeaderField = hpackDynamicTable.getEntry(index - HpackStaticTable.length);
-            return addHeader(headers, hpackHeaderField.name, hpackHeaderField.value, headersLength);
+            return  hpackDynamicTable.getEntry(index - HpackStaticTable.length);
```
Scottmitch: nit: remove double space after `return`

codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
```diff
@@ -160,27 +164,13 @@ public void testDecodeULE128LongOverflow2() throws Http2Exception {
...
     }
 
-    @Test
-    public void testSetTableSizeWithMaxUnsigned32BitValueSucceeds() throws Http2Exception {
-        byte[] input = {(byte) 0x3F, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF, (byte) 0x0E};
-        ByteBuf in = Unpooled.wrappedBuffer(input);
-        try {
-            final long expectedHeaderSize = 4026531870L; // based on the input above
-            hpackDecoder.setMaxHeaderTableSize(expectedHeaderSize);
-            hpackDecoder.decode(0, in, mockHeaders);
```
Scottmitch: is there a reason for the method movement? if not, can you undo this movement of this unit test and just add the extra parameter here?

codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
```diff
@@ -430,40 +454,6 @@ public void testLiteralNeverIndexedWithLargeValue() throws Http2Exception {
...
-            toEncode.add("test_2", "2");
-            toEncode.add("long", String.format("%0100d", 0).replace('0', 'A'));
-            toEncode.add("test_3", "3");
-            hpackEncoder.encodeHeaders(1, in, toEncode, NEVER_SENSITIVE);
-
-            // decode the headers, we should get an exception, but
-            // the decoded headers object should contain all of the headers
-            Http2Headers decoded = new DefaultHttp2Headers();
-            try {
-                hpackDecoder.decode(1, in, decoded);
```
Scottmitch: is there a reason for the method movement? if not, can you undo this movement of this unit test and just add the extra parameter here?

codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
```diff
@@ -520,7 +510,126 @@ public void testIncompleteHeaderFieldRepresentation() throws Http2Exception {
...
+        try {
+            HpackEncoder hpackEncoder = new HpackEncoder(true);
+
+            Http2Headers toEncode = new DefaultHttp2Headers();
+            toEncode.add(":status", "200");
+            toEncode.add(":method", "GET");
+            hpackEncoder.encodeHeaders(1, in, toEncode, NEVER_SENSITIVE);
+
+            // decode the headers, we should get an exception, but
+            // the decoded headers object should contain all of the headers
```
Scottmitch: I still don't completely understand this comment. I guess it is trying to say `:status` should still be contained in `decoded` ... which is why I previously asked if we should validate the header contents after the fact. However if this is not necessary (which I agree it doesn't seem necessary) can we remove/clarify this part of the comment?

Same question in the tests below too.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Headers.java
madgnome: - [X] Move HeaderType to HPackDecoder

codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
madgnome: good catch, I'll move that back.

codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
madgnome: good catch, I'll move that back.

codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
madgnome: Bad copy/paste, will delete

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
madgnome: - [X] remove space

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Headers.java
madgnome: - [x] Make final

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Headers.java
madgnome: - [x] add byte format for prefix

codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
madgnome: - [x] Move

codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
madgnome: - [x] Move

codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
madgnome: - [x] delete comment

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Headers.java
madgnome: - [x] use this in `DefaultHttp2Headers`
- [x] benchmark

 I've modified the code a bit after benchmarking it

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Headers.java
```diff
@@ -35,51 +35,84 @@
...
+        /**
+         * Returns the {@link PseudoHeaderName} corresponding to the specified header name.
+         *
+         * @return corresponding {@link PseudoHeaderName} if any, {@code null} otherwise.
+         */
+        public static PseudoHeaderName getPseudoHeader(CharSequence header) {
+            return PSEUDO_HEADERS.get(header);
+        }
+
+        public boolean isRequestOnly() {
```
Scottmitch: can you plz add some javadocs here?

madgnome: - [x] add javadoc

madgnome:
All feedbacks were actioned. 🤞

Scottmitch:
4.1 (3e6b54bb59)

Scottmitch:
thanks for your patience @madgnome !

normanmaurer:
👍

Issue #5226, opened 2016-05-09
HttpObjectAggregator.replace headers may enforce validation when they should not


Scottmitch:
`HttpObjectAggregator.replace` creates a new `DefaultFullHttpRequest` and does not specify a value for the `validateHeaders` during construction. This would mean that header validation is not consistent from the user's perspective and may enforce validation when the user does not want validation ... which may throw unexpected/undesired errors.

This issue also exists in other implementations of `replace` in codec-http.

normanmaurer:
@Scottmitch I will take care.

normanmaurer:
Fixed by https://github.com/netty/netty/pull/7503

PR #7503, opened 2017-12-13
Only enable validation of headers if original headers were validating…

normanmaurer:
… as well.

Motiviation:

In our replace(...) methods we always used validation for the newly created headers while the original headers may not use validation at all.

Modifications:

- Only use validation if the original headers used validation as well.
- Ensure we create a copy of the headers in replace(...).

Result:

Fixes [#5226]

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/DefaultFullHttpRequest.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/DefaultFullHttpRequest.java
@@ -142,7 +142,10 @@ public class DefaultFullHttpRequest extends DefaultHttpRequest implements FullHt
 
     @Override
     public FullHttpRequest replace(ByteBuf content) {
-        return new DefaultFullHttpRequest(protocolVersion(), method(), uri(), content, headers(), trailingHeaders());
+        FullHttpRequest request = new DefaultFullHttpRequest(protocolVersion(), method(), uri(), content,
+                HttpHeaders.copy(headers()), HttpHeaders.copy(trailingHeaders()));
+        request.setDecoderResult(decoderResult());
+        return request;
     }
 
     @Override
--- a/codec-http/src/main/java/io/netty/handler/codec/http/DefaultFullHttpRequest.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/DefaultFullHttpRequest.java
@@ -143,7 +143,7 @@ public class DefaultFullHttpRequest extends DefaultHttpRequest implements FullHt
     @Override
     public FullHttpRequest replace(ByteBuf content) {
         FullHttpRequest request = new DefaultFullHttpRequest(protocolVersion(), method(), uri(), content,
-                HttpHeaders.copy(headers()), HttpHeaders.copy(trailingHeaders()));
+                headers().copy(), trailingHeaders().copy());
         request.setDecoderResult(decoderResult());
         return request;
     }
--- a/codec-http/src/main/java/io/netty/handler/codec/http/DefaultFullHttpResponse.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/DefaultFullHttpResponse.java
@@ -150,7 +150,7 @@ public class DefaultFullHttpResponse extends DefaultHttpResponse implements Full
     @Override
     public FullHttpResponse replace(ByteBuf content) {
         FullHttpResponse response = new DefaultFullHttpResponse(protocolVersion(), status(), content,
-                HttpHeaders.copy(headers()), HttpHeaders.copy(trailingHeaders()));
+                headers().copy(), trailingHeaders().copy());
         response.setDecoderResult(decoderResult());
         return response;
     }
--- a/codec-http/src/main/java/io/netty/handler/codec/http/DefaultFullHttpResponse.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/DefaultFullHttpResponse.java
@@ -149,7 +149,10 @@ public class DefaultFullHttpResponse extends DefaultHttpResponse implements Full
 
     @Override
     public FullHttpResponse replace(ByteBuf content) {
-        return new DefaultFullHttpResponse(protocolVersion(), status(), content, headers(), trailingHeaders());
+        FullHttpResponse response = new DefaultFullHttpResponse(protocolVersion(), status(), content,
+                HttpHeaders.copy(headers()), HttpHeaders.copy(trailingHeaders()));
+        response.setDecoderResult(decoderResult());
+        return response;
     }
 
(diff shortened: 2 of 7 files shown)


codec-http2/src/test/java/io/netty/handler/codec/http2/InboundHttp2ToHttpAdapterTest.java
```diff
@@ -607,6 +608,9 @@ public void run() throws Http2Exception {
             verify(serverListener, times(2)).messageReceived(requestCaptor.capture());
             capturedRequests = requestCaptor.getAllValues();
             assertEquals(2, capturedRequests.size());
+            // We expect to not have this header i nthe captured request so remove it now.
+            assertNotNull(request.headers().remove("x-http2-stream-weight"));
```
normanmaurer: Note this change was needed as we not correctly copied the headers before in our replace(...) methods.

codec-http/src/main/java/io/netty/handler/codec/http/DefaultFullHttpRequest.java
```diff
@@ -142,7 +142,10 @@ public FullHttpRequest retainedDuplicate() {
 
     @Override
     public FullHttpRequest replace(ByteBuf content) {
-        return new DefaultFullHttpRequest(protocolVersion(), method(), uri(), content, headers(), trailingHeaders());
+        FullHttpRequest request = new DefaultFullHttpRequest(protocolVersion(), method(), uri(), content,
+                HttpHeaders.copy(headers()), HttpHeaders.copy(trailingHeaders()));
+        request.setDecoderResult(decoderResult());
```
normanmaurer: Also we missed this before :(

codec-http/src/main/java/io/netty/handler/codec/http/DefaultFullHttpResponse.java
```diff
@@ -149,7 +149,10 @@ public FullHttpResponse retainedDuplicate() {
 
     @Override
     public FullHttpResponse replace(ByteBuf content) {
-        return new DefaultFullHttpResponse(protocolVersion(), status(), content, headers(), trailingHeaders());
+        FullHttpResponse response = new DefaultFullHttpResponse(protocolVersion(), status(), content,
+                HttpHeaders.copy(headers()), HttpHeaders.copy(trailingHeaders()));
+        response.setDecoderResult(decoderResult());
```
normanmaurer: Also we missed this before :(

codec/src/main/java/io/netty/handler/codec/DefaultHeaders.java
```diff
@@ -952,6 +952,16 @@ private T thisT() {
         return (T) this;
     }
 
+    /**
+     * Returns a deep copy of this instance.
+     */
+    public DefaultHeaders<K, V, T> copy() {
+        DefaultHeaders<K, V, T> copy = new DefaultHeaders<K, V, T>(
+                hashingStrategy, valueConverter, nameValidator, entries.length);
+        copy.set(this);
```
Scottmitch: `set` will also do `clear` which will do some unnecessary work (array filling, pointer updating, assignments). We should be able to just use `addImpl` instead of `set`.

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaders.java
```diff
@@ -1693,4 +1693,18 @@ public boolean contains(CharSequence name, CharSequence value, boolean ignoreCas
     public String toString() {
         return HeadersUtils.toString(getClass(), iteratorCharSequence(), size());
     }
+
+    /**
+     * Returns a deap copy of the passed in {@link HttpHeaders}.
+     */
+    public static HttpHeaders copy(HttpHeaders headers) {
+        final HttpHeaders copy;
```
Scottmitch: nit: consider simplifying this to a tertiary statement:

```java
return (headers instanceof DefaultHttpHeaders) ? ((DefaultHttpHeaders) headers).copy()
                : new DefaultHttpHeaders().set(headers);
```

codec-http/src/main/java/io/netty/handler/codec/http/DefaultHttpHeaders.java
```diff
@@ -42,7 +42,7 @@
 /**
  * Default implementation of {@link HttpHeaders}.
  */
-public class DefaultHttpHeaders extends HttpHeaders {
+public class DefaultHttpHeaders extends HttpHeaders implements Cloneable {
```
Scottmitch: this class doesn't implement `clone` so should we get rid of the `implements Cloneable`?

codec-http2/src/test/java/io/netty/handler/codec/http2/InboundHttp2ToHttpAdapterTest.java
```diff
@@ -607,6 +608,9 @@ public void run() throws Http2Exception {
             verify(serverListener, times(2)).messageReceived(requestCaptor.capture());
             capturedRequests = requestCaptor.getAllValues();
             assertEquals(2, capturedRequests.size());
+            // We expect to not have this header i nthe captured request so remove it now.
```
Scottmitch: `i nthe` -> `in the`

Scottmitch: `We expect to not have` -> `We do not expect to have`

codec-http/src/main/java/io/netty/handler/codec/http/DefaultHttpHeaders.java
normanmaurer: @Scottmitch doh.. good catch, thats a left-over as I first wanted to just use `clone()` for it.

codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaders.java
```diff
@@ -1693,4 +1693,12 @@ public boolean contains(CharSequence name, CharSequence value, boolean ignoreCas
     public String toString() {
         return HeadersUtils.toString(getClass(), iteratorCharSequence(), size());
     }
+
+    /**
+     * Returns a deap copy of the passed in {@link HttpHeaders}.
+     */
+    public static HttpHeaders copy(HttpHeaders headers) {
```
Scottmitch: can we just make `copy()` a method of the `HttpHeaders` class because it is an abstract class? Then we don't have to use the static method outside and can just call `.copy()` on the object.

normanmaurer:
Cherry-picked into 4.1 (942b993f2b9781ff2126ff92a6be5b975dfc72ed)

PR #7506, opened 2017-12-14
Remove WeakOrderedQueue from WeakHashMap when FastThreadLocal value w…

normanmaurer:
…as removed if possible.

Motivation:

We should remove the WeakOrderedQueue from the WeakHashMap directly if possible and only depend on the semantics of the WeakHashMap if there is no other way for us to cleanup it.

Modifications:

Override onRemoval(...) to remove the WeakOrderedQueue if possible.

Result:

Less overhead and quicker collection of WeakOrderedQueue for some cases.

### DIFF
--- a/common/src/main/java/io/netty/util/Recycler.java
+++ b/common/src/main/java/io/netty/util/Recycler.java
@@ -114,6 +114,16 @@ public abstract class Recycler<T> {
             return new Stack<T>(Recycler.this, Thread.currentThread(), maxCapacityPerThread, maxSharedCapacityFactor,
                     ratioMask, maxDelayedQueuesPerThread);
         }
+
+        @Override
+        protected void onRemoval(Stack<T> value) {
+            // Let us remove the WeakOrderQueue from the WeakHashMap directly if its safe to remove some overhead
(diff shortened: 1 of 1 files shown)


common/src/main/java/io/netty/util/Recycler.java
```diff
@@ -114,6 +114,16 @@ public void recycle(Object object) {
             return new Stack<T>(Recycler.this, Thread.currentThread(), maxCapacityPerThread, maxSharedCapacityFactor,
                     ratioMask, maxDelayedQueuesPerThread);
         }
+
+        @Override
+        protected void onRemoval(Stack<T> value) {
+            // Let us remove the WeakOrderQueue from the WeakHashMap directly if its safe to remove some overhead
```
johnou: Invoke super.onRemoval?

normanmaurer:
No need it’s empty / noop

normanmaurer:
Cherry-picked into 4.1 (640a22df9efb41e3d29b79916938c1c315be2872) and 4.0 (4e62f9fd95906f0a871404f4ef4f10795dedb889)

PR #7507, opened 2017-12-14
FileDescriptor writev core dump

Scottmitch:
Motivation:
FileDescriptor#writev calls JNI code, and that JNI code dereferences a NULL pointer which crashes the application. This occurs when writing a single CompositeByteBuf object with more than one component.

Modifications:
- Initialize the iovec iterator properly to avoid the core dump
- Fix the array length calculation if we aren't able to fit all the ByteBuffer objects in the iovec array

Result:
No more core dump.

### DIFF
--- a/transport-native-unix-common/src/main/c/netty_unix_filedescriptor.c
+++ b/transport-native-unix-common/src/main/c/netty_unix_filedescriptor.c
@@ -111,19 +111,19 @@ static jlong netty_unix_filedescriptor_writevAddresses(JNIEnv* env, jclass clazz
     return _writev(env, clazz, fd, iov, length);
 }
 
-static jlong netty_unix_filedescriptor_writev(JNIEnv* env, jclass clazz, jint fd, jobjectArray buffers, jint offset, jint length, jlong maxBytesToWrite) {
+static jlong netty_unix_filedescriptor_writev(JNIEnv* env, jclass clazz, jint fd, jobjectArray buffers, const jint offset, jint length, jlong maxBytesToWrite) {
     struct iovec iov[length];
-    struct iovec* iovptr = NULL;
+    struct iovec* iovptr = iov;
     int i;
     int num = offset + length;
     if (posFieldId != NULL && limitFieldId != NULL) {
-        for (i = offset; i < num; i++) {
+        for (i = offset; i < num; ++i) {
             jobject bufObj = (*env)->GetObjectArrayElement(env, buffers, i);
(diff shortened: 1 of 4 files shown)


transport-native-unix-common/src/main/c/netty_unix_filedescriptor.c
```diff
@@ -163,13 +163,13 @@ static jlong netty_unix_filedescriptor_writev(JNIEnv* env, jclass clazz, jint fd
             (*env)->DeleteLocalRef(env, bufObj);
         }
     } else if (limitFieldId != NULL) {
-        for (i = offset; i < num; i++) {
+        for (i = offset; i < num; ++i) {
             jobject bufObj = (*env)->GetObjectArrayElement(env, buffers, i);
             jint pos = (*env)->CallIntMethod(env, bufObj, posId, NULL);
             jint limit = (*env)->GetIntField(env, bufObj, limitFieldId);
```
todor-m: What if you have a pending exception? Will this code work without calling `ExceptionOccurred/ExceptionClear`?

Scottmitch: We typically don't call `ExceptionOccurred` in JNI unless there is a specific need to. We generally let the exceptions be thrown (if any occurred) after transitioning back from JNI to Java.

Do you see a problem with this approach on this line/method, or just a general question?

todor-m: The exception will be thrown when you return from the method (if I remember correctly). I think JNI will keep returning 0s (or some other garbage) if you don't clear the exception. Think about what can happen when pos and limit are 0s (or a bad bufObj).

todor-m: Are you accounting for the case when JNI will start returning bad values because of an exception?

todor-m: It is more like a general question since it applies to all the JNI calls in netty, not that particular one.

Scottmitch: Lets consider this as a separate issue as the behavior w.r.t to JNI exception hasn't changed in this PR and scope is larger.  It may boil down to how paranoid/defensive do we want to be. Perhaps we should check the state before we make the system call.

Scottmitch: Thanks for bringing this up @todor-m ... see https://github.com/netty/netty/issues/7508

PR #7512, opened 2017-12-16
codec-http2: Better keep track of nameLength in HpackDecoder.decode

### LINKED ISSUE #7511: double counting header name length in http/2
### Expected behavior

A header with a name of length X and a value of length Y gets considered as being length X + Y in total.

### Actual behavior

A header with a name of length X and a value of length Y gets considered as being length 2X + Y in total.  I think this is where the bug is:

https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java#L262

### Steps to reproduce

Use hpack decoder to decode a header.

### Netty version

4.1.16.Final

jdk8

mosesn:
Motivation:

http/2 counts header sizes somewhat inconsistently.  Sometimes, headers
which are substantively less than the header list size will be measured
as longer than the header list size.

Modifications:

Keep better track of the nameLength of a given name, so that we don't
accidentally end up reusing a nameLength.

Result:

More consistent measurement of header list size.

Fixes #7511.

### DIFF
--- a/codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
+++ b/codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
@@ -455,4 +455,26 @@ public class HpackDecoderTest {
             in.release();
         }
     }
+
+    @Test
+    public void testDecodeCountsNamesOnlyOnce() throws Http2Exception {
+        ByteBuf in = Unpooled.buffer(200);
+        try {
+            hpackDecoder.setMaxHeaderListSize(3500, 4000);
+            HpackEncoder hpackEncoder = new HpackEncoder(true);
+
+            // encode headers that are slightly larger than maxHeaderListSize
+            // but smaller than maxHeaderListSizeGoAway
+            Http2Headers toEncode = new DefaultHttp2Headers();
+            toEncode.add(String.format("%03000d", 0).replace('0', 'f'), "value");
+            toEncode.add("accept", "value");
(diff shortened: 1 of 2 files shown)


codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
```diff
@@ -455,4 +455,28 @@ public void testDecodeLargerThanMaxHeaderListSizeButSmallerThanMaxHeaderListSize
...
+            HpackEncoder hpackEncoder = new HpackEncoder(true);
+
+            // encode headers that are slightly larger than maxHeaderListSize
+            // but smaller than maxHeaderListSizeGoAway
+            Http2Headers toEncode = new DefaultHttp2Headers();
+            toEncode.add(String.format("%03000d", 0).replace('0', 'f'), "value");
+            toEncode.add("accept", "value");
+            hpackEncoder.encodeHeaders(1, in, toEncode, NEVER_SENSITIVE);
+
+            // decode the headers, we should get an exception, but
```
normanmaurer: this comment is a bit confusing... when we "get an exception" shouldn't we need to catch it ?

mosesn: oh, I just copy/pasted this from the previous test.  I'll just delete it.

mosesn: update, PTAL

Scottmitch:
thanks @mosesn !

PR #7513, opened 2017-12-16
Improvement : added overloaded method in order to avoid unnecessary allocation

doom369:
Motivation : 

Avoid unnecessary array allocation when using the function with varargs in the `DefaultChannelPipeline` class.

Modifications : 

Added `addLast` and `addFirst` overloaded methods with 1 handler instead of varargs.

Result : 

No array allocation when using simple construction like `pipeline.addLast(new Handler());`


normanmaurer:
Cherry-picked into 4.1 (f9888acfdd38b6184e7728ed61adbf37be8be37d).

@doom369 thanks

PR #7521, opened 2017-12-18
Added configurable ByteBuf bounds checking

franz1981:
Motivation:

The JVM isn't always able to hoist out/reduce bounds checking (due to ref counting operations etc etc) hence making it configurable could improve performances for most CPU intensive use cases.

Modifications:

Each AbstractByteBuf bounds check has been tested against a new static final configuration property similar to checkAccessible ie io.netty.buffer.bytebuf.checkBounds.

Result:

Any user could disable ByteBuf bounds checking in order to get extra performances.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
@@ -43,13 +43,22 @@ import static io.netty.util.internal.MathUtil.isOutOfBounds;
  */
 public abstract class AbstractByteBuf extends ByteBuf {
     private static final InternalLogger logger = InternalLoggerFactory.getInstance(AbstractByteBuf.class);
-    private static final String PROP_MODE = "io.netty.buffer.bytebuf.checkAccessible";
+    private static final String LEGACY_PROP_CHECK_ACCESSIBLE = "io.netty.buffer.bytebuf.checkAccessible";
+    private static final String PROP_CHECK_ACCESSIBLE = "io.netty.buffer.checkAccessible";
     private static final boolean checkAccessible;
+    private static final String PROP_CHECK_BOUNDS = "io.netty.buffer.checkBounds";
+    private static final boolean checkBounds;
 
     static {
-        checkAccessible = SystemPropertyUtil.getBoolean(PROP_MODE, true);
+        if (SystemPropertyUtil.contains(PROP_CHECK_ACCESSIBLE)) {
+            checkAccessible = SystemPropertyUtil.getBoolean(PROP_CHECK_ACCESSIBLE, true);
+        } else {
+            checkAccessible = SystemPropertyUtil.getBoolean(LEGACY_PROP_CHECK_ACCESSIBLE, true);
+        }
+        checkBounds = SystemPropertyUtil.getBoolean(PROP_CHECK_BOUNDS, true);
         if (logger.isDebugEnabled()) {
-            logger.debug("-D{}: {}", PROP_MODE, checkAccessible);
+            logger.debug("-D{}: {}", PROP_CHECK_ACCESSIBLE, checkAccessible);
+            logger.debug("-D{}: {}", PROP_CHECK_BOUNDS, checkBounds);
         }
     }
 
@@ -97,11 +106,18 @@ public abstract class AbstractByteBuf extends ByteBuf {
         return readerIndex;
     }
 
+    private static void checkIndexBounds(final int readerIndex, final int writerIndex, final int capacity) {
+        if (readerIndex < 0 || readerIndex > writerIndex || writerIndex > capacity) {
+            throw new IndexOutOfBoundsException(String.format(
+                    "readerIndex: %d, writerIndex: %d (expected: 0 <= readerIndex <= writerIndex <= capacity(%d))",
+                    readerIndex, writerIndex, capacity));
+        }
+    }
+
     @Override
     public ByteBuf readerIndex(int readerIndex) {
-        if (readerIndex < 0 || readerIndex > writerIndex) {
-            throw new IndexOutOfBoundsException(String.format(
-                    "readerIndex: %d (expected: 0 <= readerIndex <= writerIndex(%d))", readerIndex, writerIndex));
+        if (checkBounds) {
+            checkIndexBounds(readerIndex, writerIndex, capacity());
         }
         this.readerIndex = readerIndex;
         return this;
@@ -114,10 +130,8 @@ public abstract class AbstractByteBuf extends ByteBuf {
 
     @Override
     public ByteBuf writerIndex(int writerIndex) {
-        if (writerIndex < readerIndex || writerIndex > capacity()) {
-            throw new IndexOutOfBoundsException(String.format(
-                    "writerIndex: %d (expected: readerIndex(%d) <= writerIndex <= capacity(%d))",
-                    writerIndex, readerIndex, capacity()));
+        if (checkBounds) {
+            checkIndexBounds(readerIndex, writerIndex, capacity());
         }
         this.writerIndex = writerIndex;
         return this;
@@ -125,10 +139,8 @@ public abstract class AbstractByteBuf extends ByteBuf {
 
     @Override
     public ByteBuf setIndex(int readerIndex, int writerIndex) {
-        if (readerIndex < 0 || readerIndex > writerIndex || writerIndex > capacity()) {
-            throw new IndexOutOfBoundsException(String.format(
-                    "readerIndex: %d, writerIndex: %d (expected: 0 <= readerIndex <= writerIndex <= capacity(%d))",
-                    readerIndex, writerIndex, capacity()));
+        if (checkBounds) {
+            checkIndexBounds(readerIndex, writerIndex, capacity());
         }
         setIndex0(readerIndex, writerIndex);
         return this;
@@ -271,11 +283,12 @@ public abstract class AbstractByteBuf extends ByteBuf {
         if (minWritableBytes <= writableBytes()) {
             return;
         }
-
-        if (minWritableBytes > maxCapacity - writerIndex) {
-            throw new IndexOutOfBoundsException(String.format(
-                    "writerIndex(%d) + minWritableBytes(%d) exceeds maxCapacity(%d): %s",
-                    writerIndex, minWritableBytes, maxCapacity, this));
+        if (checkBounds) {
+            if (minWritableBytes > maxCapacity - writerIndex) {
+                throw new IndexOutOfBoundsException(String.format(
+                        "writerIndex(%d) + minWritableBytes(%d) exceeds maxCapacity(%d): %s",
+                        writerIndex, minWritableBytes, maxCapacity, this));
+            }
         }
 
         // Normalize the current capacity to the power of 2.
@@ -618,15 +631,21 @@ public abstract class AbstractByteBuf extends ByteBuf {
         return this;
     }
 
+    private static void checkReadableBounds(final ByteBuf src, final int length) {
+        if (length > src.readableBytes()) {
+            throw new IndexOutOfBoundsException(String.format(
+                    "length(%d) exceeds src.readableBytes(%d) where src is: %s", length, src.readableBytes(), src));
+        }
+    }
+
     @Override
     public ByteBuf setBytes(int index, ByteBuf src, int length) {
         checkIndex(index, length);
         if (src == null) {
             throw new NullPointerException("src");
         }
-        if (length > src.readableBytes()) {
-            throw new IndexOutOfBoundsException(String.format(
-                    "length(%d) exceeds src.readableBytes(%d) where src is: %s", length, src.readableBytes(), src));
+        if (checkBounds) {
+            checkReadableBounds(src, length);
         }
 
         setBytes(index, src, src.readerIndex(), length);
@@ -889,9 +908,11 @@ public abstract class AbstractByteBuf extends ByteBuf {
 
     @Override
     public ByteBuf readBytes(ByteBuf dst, int length) {
-        if (length > dst.writableBytes()) {
-            throw new IndexOutOfBoundsException(String.format(
-                    "length(%d) exceeds dst.writableBytes(%d) where dst is: %s", length, dst.writableBytes(), dst));
+        if (checkBounds) {
+            if (length > dst.writableBytes()) {
+                throw new IndexOutOfBoundsException(String.format(
+                        "length(%d) exceeds dst.writableBytes(%d) where dst is: %s", length, dst.writableBytes(), dst));
+            }
         }
         readBytes(dst, dst.writerIndex(), length);
         dst.writerIndex(dst.writerIndex() + length);
@@ -1065,9 +1086,8 @@ public abstract class AbstractByteBuf extends ByteBuf {
 
     @Override
     public ByteBuf writeBytes(ByteBuf src, int length) {
-        if (length > src.readableBytes()) {
-            throw new IndexOutOfBoundsException(String.format(
-                    "length(%d) exceeds src.readableBytes(%d) where src is: %s", length, src.readableBytes(), src));
+        if (checkBounds) {
+            checkReadableBounds(src, length);
         }
         writeBytes(src, src.readerIndex(), length);
         src.readerIndex(src.readerIndex() + length);
@@ -1355,26 +1375,30 @@ public abstract class AbstractByteBuf extends ByteBuf {
         checkIndex0(index, fieldLength);
     }
 
-    final void checkIndex0(int index, int fieldLength) {
-        if (isOutOfBounds(index, fieldLength, capacity())) {
+    private static void checkRangeBounds(final int index, final int fieldLength, final int capacity) {
+        if (isOutOfBounds(index, fieldLength, capacity)) {
             throw new IndexOutOfBoundsException(String.format(
-                    "index: %d, length: %d (expected: range(0, %d))", index, fieldLength, capacity()));
+                    "index: %d, length: %d (expected: range(0, %d))", index, fieldLength, capacity));
+        }
+    }
+
+    final void checkIndex0(int index, int fieldLength) {
+        if (checkBounds) {
+            checkRangeBounds(index, fieldLength, capacity());
         }
     }
 
     protected final void checkSrcIndex(int index, int length, int srcIndex, int srcCapacity) {
         checkIndex(index, length);
-        if (isOutOfBounds(srcIndex, length, srcCapacity)) {
-            throw new IndexOutOfBoundsException(String.format(
-                    "srcIndex: %d, length: %d (expected: range(0, %d))", srcIndex, length, srcCapacity));
+        if (checkBounds) {
+            checkRangeBounds(srcIndex, length, srcCapacity);
         }
     }
 
     protected final void checkDstIndex(int index, int length, int dstIndex, int dstCapacity) {
         checkIndex(index, length);
-        if (isOutOfBounds(dstIndex, length, dstCapacity)) {
-            throw new IndexOutOfBoundsException(String.format(
-                    "dstIndex: %d, length: %d (expected: range(0, %d))", dstIndex, length, dstCapacity));
+        if (checkBounds) {
+            checkRangeBounds(dstIndex, length, dstCapacity);
         }
     }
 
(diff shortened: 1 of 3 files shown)


franz1981:
Some perf numbers here: 
```
Benchmark                             (checkBounds)   Mode  Cnt          Score          Error  Units
HeapByteBufBenchmark.getByte                   true  thrpt   20  332684594.313 ±  8953313.584  ops/s
HeapByteBufBenchmark.getByte                  false  thrpt   20  331470006.429 ±  5992112.659  ops/s
HeapByteBufBenchmark.getByteUnsafe             true  thrpt   20  297034653.544 ±  7067998.751  ops/s
HeapByteBufBenchmark.getByteUnsafe            false  thrpt   20  350080973.276 ±  2829875.724  ops/s
HeapByteBufBenchmark.getInt                    true  thrpt   20  233769231.764 ± 37982264.776  ops/s
HeapByteBufBenchmark.getInt                   false  thrpt   20  246525191.924 ±  6646627.756  ops/s
HeapByteBufBenchmark.getIntUnsafe              true  thrpt   20  298039670.527 ±  1739762.174  ops/s
HeapByteBufBenchmark.getIntUnsafe             false  thrpt   20  340834739.094 ± 11627875.418  ops/s
HeapByteBufBenchmark.getLong                   true  thrpt   20  188164538.785 ±  4310305.428  ops/s
HeapByteBufBenchmark.getLong                  false  thrpt   20  189511566.299 ±  4289526.979  ops/s
HeapByteBufBenchmark.getLongUnsafe             true  thrpt   20  262591794.664 ±  4195639.130  ops/s
HeapByteBufBenchmark.getLongUnsafe            false  thrpt   20  324980975.085 ± 13722245.684  ops/s
HeapByteBufBenchmark.getMedium                 true  thrpt   20  265704119.637 ± 17150786.218  ops/s
HeapByteBufBenchmark.getMedium                false  thrpt   20  260680356.073 ± 13971219.764  ops/s
HeapByteBufBenchmark.getMediumUnsafe           true  thrpt   20  240690889.348 ± 11765957.460  ops/s
HeapByteBufBenchmark.getMediumUnsafe          false  thrpt   20  266946759.429 ± 15270851.535  ops/s
HeapByteBufBenchmark.getShort                  true  thrpt   20  301194012.966 ±  3316910.632  ops/s
HeapByteBufBenchmark.getShort                 false  thrpt   20  287463722.495 ± 11626358.454  ops/s
HeapByteBufBenchmark.getShortUnsafe            true  thrpt   20  288905335.287 ±  2652739.151  ops/s
HeapByteBufBenchmark.getShortUnsafe           false  thrpt   20  348658308.647 ±  5516655.619  ops/s
```

franz1981:
While this one to show the direct ones too:
```
Benchmark                                    (checkBounds)   Mode  Cnt           Score          Error  Units
BaseByteBufBenchmark.setByteBufDirect                 true  thrpt   20   913441876.824 ± 32621169.332  ops/s
BaseByteBufBenchmark.setByteBufDirect                false  thrpt   20  1145345090.958 ± 57174168.802  ops/s
BaseByteBufBenchmark.setByteBufDirectPooled           true  thrpt   20   869004223.911 ± 62690580.273  ops/s
BaseByteBufBenchmark.setByteBufDirectPooled          false  thrpt   20  1142072093.640 ± 28696919.049  ops/s
BaseByteBufBenchmark.setByteBufHeap                   true  thrpt   20   838566890.117 ± 88192014.727  ops/s
BaseByteBufBenchmark.setByteBufHeap                  false  thrpt   20  1131150064.668 ± 49330465.530  ops/s
BaseByteBufBenchmark.setByteBufferDirect              true  thrpt   20   955755981.625 ± 57827866.657  ops/s
BaseByteBufBenchmark.setByteBufferDirect             false  thrpt   20   985091640.816 ± 34782870.293  ops/s
BaseByteBufBenchmark.setByteBufferHeap                true  thrpt   20   619423886.978 ± 21110693.427  ops/s
BaseByteBufBenchmark.setByteBufferHeap               false  thrpt   20   622036829.778 ± 35497214.491  ops/s
```

fenik17:
We will never turn off bords checks on production... Can you show a comparison with current version, where there is no extra `if`?

franz1981:
@fenik17  afaik it is optimized away but I will figured it out in the size in byte codes of the method impacting just on inlining, anyway I will provide soon a comparison.For further info please take a look here: http://pveentjer.blogspot.it/2017/01/final-static-boolean-jit.html
There are already Agrona's Unsafe buffer that employ a similar method to allow users to trade efficiency with security.

franz1981:
@fenik17 
It seems as I've posted above: the additional `if` has been removed hence it is for free:
```
Benchmark                              Mode  Cnt          Score          Error  Units
HeapByteBufBenchmark.getByteUnsafe    thrpt   20  286342717.658 ± 10601001.635  ops/s
HeapByteBufBenchmark.getIntUnsafe     thrpt   20  280028987.454 ±  9262047.662  ops/s
HeapByteBufBenchmark.getLongUnsafe    thrpt   20  282384007.227 ±  5972554.961  ops/s
HeapByteBufBenchmark.getMediumUnsafe  thrpt   20  252613116.587 ±  1326148.403  ops/s
HeapByteBufBenchmark.getShortUnsafe   thrpt   20  280816212.215 ±  8164328.001  ops/s
```

fenik17:
@franz1981 Ok. Thanks!

franz1981:
@normanmaurer 
I've noticed that the numbers on `ByteBufBenchmark` seems too high, especially if compared to the others in `HeapByteBufBenchmark` ie ~1000 M ops/sec vs 300 M ops/sec
I've noticed that `ByteBufBenchmark` isn't returning the `ByteBuf`/`ByteBuffer` used, hence probably it could make sense to turn
```
    @Benchmark
    public void setByteBufDirectPooled() {
        directBufferPooled.setByte(0, BYTE);
    }
```
into
```
    @Benchmark
    public ByteBuf setByteBufDirectPooled() {
        return directBufferPooled.setByte(0, BYTE);
    }
```
That would help to better measure bounds checking cost and avoid any dead code elimination that won't happen in a real production environment, wdyt?

slandelle:
@franz1981 Absolutely. You need to either return a value or explicitly consume with a BlackHole, otherwise JIT will eliminate your code.

franz1981:
@slandelle 
TBH I've run it with the Linux Perf Asm profiler and it doesn't seems to happen, but as you rightly said the JIT is allowed to do it so It probably makes sense to change it: better safe than sorry :+1: 
Probably it could be addressed with a separate PR or It makes sense to do it into this one?

slandelle:
> better safe than sorry

Actually, I'm pretty sure `ByteBufBenchmark` is broken, see [JMH doc](http://hg.openjdk.java.net/code-tools/jmh/file/1ddf31f810a3/jmh-samples/src/main/java/org/openjdk/jmh/samples/JMHSample_08_DeadCode.java).

> Probably it could be addressed with a separate PR or It makes sense to do it into this one?

I'd be tempted to go with a separate PR.

normanmaurer:
+1 please open a pr

franz1981:
@slandelle 
> Actually, I'm pretty sure ByteBufBenchmark is broken, see JMH doc.

Good point!Anyway consider that's not the same exact case: the JVM can't say if the `ByteBuf`/`ByteBuffer` could be printed/used further after `@benchmark` because is a instance member, but It would be better to fix it anyway :+1: 

@normanmaurer 
yep will do :+1:

franz1981:
@normanmaurer @slandelle 
Done here: https://github.com/netty/netty/pull/7524

It would be interesting to add front/back padding (and alignment too) to avoid any interactions with the surrounding memory: I think it could make the results more stable especially when is needed to compare vs `ByteBuffer`

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -99,9 +102,11 @@ public int readerIndex() {
 
     @Override
     public ByteBuf readerIndex(int readerIndex) {
-        if (readerIndex < 0 || readerIndex > writerIndex) {
-            throw new IndexOutOfBoundsException(String.format(
-                    "readerIndex: %d (expected: 0 <= readerIndex <= writerIndex(%d))", readerIndex, writerIndex));
+        if (checkBounds) {
+            if (readerIndex < 0 || readerIndex > writerIndex) {
```
normanmaurer: merge the two above branches.

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -114,21 +119,25 @@ public int writerIndex() {
 
     @Override
     public ByteBuf writerIndex(int writerIndex) {
-        if (writerIndex < readerIndex || writerIndex > capacity()) {
-            throw new IndexOutOfBoundsException(String.format(
-                    "writerIndex: %d (expected: readerIndex(%d) <= writerIndex <= capacity(%d))",
-                    writerIndex, readerIndex, capacity()));
+        if (checkBounds) {
+            if (writerIndex < readerIndex || writerIndex > capacity()) {
```
normanmaurer: merge the branches

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -114,21 +119,25 @@ public int writerIndex() {
...
     }
 
     @Override
     public ByteBuf setIndex(int readerIndex, int writerIndex) {
-        if (readerIndex < 0 || readerIndex > writerIndex || writerIndex > capacity()) {
-            throw new IndexOutOfBoundsException(String.format(
-                    "readerIndex: %d, writerIndex: %d (expected: 0 <= readerIndex <= writerIndex <= capacity(%d))",
-                    readerIndex, writerIndex, capacity()));
+        if (checkBounds) {
+            if (readerIndex < 0 || readerIndex > writerIndex || writerIndex > capacity()) {
```
normanmaurer: merge the branches

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -271,11 +280,12 @@ final void ensureWritable0(int minWritableBytes) {
         if (minWritableBytes <= writableBytes()) {
             return;
         }
-
-        if (minWritableBytes > maxCapacity - writerIndex) {
-            throw new IndexOutOfBoundsException(String.format(
-                    "writerIndex(%d) + minWritableBytes(%d) exceeds maxCapacity(%d): %s",
-                    writerIndex, minWritableBytes, maxCapacity, this));
+        if (checkBounds) {
+            if (minWritableBytes > maxCapacity - writerIndex) {
```
normanmaurer: merge branches

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -624,9 +634,11 @@ public ByteBuf setBytes(int index, ByteBuf src, int length) {
         if (src == null) {
             throw new NullPointerException("src");
         }
-        if (length > src.readableBytes()) {
-            throw new IndexOutOfBoundsException(String.format(
-                    "length(%d) exceeds src.readableBytes(%d) where src is: %s", length, src.readableBytes(), src));
+        if (checkBounds) {
+            if (length > src.readableBytes()) {
```
normanmaurer: merge branches

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -887,9 +899,11 @@ public ByteBuf readBytes(ByteBuf dst) {
 
     @Override
     public ByteBuf readBytes(ByteBuf dst, int length) {
-        if (length > dst.writableBytes()) {
-            throw new IndexOutOfBoundsException(String.format(
-                    "length(%d) exceeds dst.writableBytes(%d) where dst is: %s", length, dst.writableBytes(), dst));
+        if (checkBounds) {
+            if (length > dst.writableBytes()) {
```
normanmaurer: merge branches

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -1063,9 +1077,11 @@ public ByteBuf writeBytes(ByteBuf src) {
 
     @Override
     public ByteBuf writeBytes(ByteBuf src, int length) {
-        if (length > src.readableBytes()) {
-            throw new IndexOutOfBoundsException(String.format(
-                    "length(%d) exceeds src.readableBytes(%d) where src is: %s", length, src.readableBytes(), src));
+        if (checkBounds) {
+            if (length > src.readableBytes()) {
```
normanmaurer: merge branches

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -1354,25 +1370,31 @@ protected final void checkIndex(int index, int fieldLength) {
     }
 
     final void checkIndex0(int index, int fieldLength) {
-        if (isOutOfBounds(index, fieldLength, capacity())) {
-            throw new IndexOutOfBoundsException(String.format(
-                    "index: %d, length: %d (expected: range(0, %d))", index, fieldLength, capacity()));
+        if (checkBounds) {
+            if (isOutOfBounds(index, fieldLength, capacity())) {
```
normanmaurer: merge branches

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -1354,25 +1370,31 @@ protected final void checkIndex(int index, int fieldLength) {
...
         }
     }
 
     protected final void checkSrcIndex(int index, int length, int srcIndex, int srcCapacity) {
         checkIndex(index, length);
-        if (isOutOfBounds(srcIndex, length, srcCapacity)) {
-            throw new IndexOutOfBoundsException(String.format(
-                    "srcIndex: %d, length: %d (expected: range(0, %d))", srcIndex, length, srcCapacity));
+        if (checkBounds) {
+            if (isOutOfBounds(srcIndex, length, srcCapacity)) {
```
normanmaurer: merge branches ?

franz1981:
@normanmaurer 
> basically looking good to me... just wondering if you want to merge the branches

I'm taking care of them (hopefully without introducing any bugs :P

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
franz1981: I've not merged this one because I don't want to loose the distinctions of the `dst.writableBytes` and `src.readableBytes` message logs in favour of something more generic to replace both.
If you want me to merge them please suggest me a proper meaningful message to help users when the checks will fail :+1:

franz1981:
@normanmaurer I've added further configurable points of bounds checking ie probably will squeeze a lil' more of performance too
Please check if they seems fine too :)

franz1981:
@normanmaurer I'm seeing several errors on CI, there is something I could do for them Re this PR?
Seeing everything red on the PR is scary! :)

franz1981:
@normanmaurer Anyone has had the chance to try if the perf gained by this one is good enough?

normanmaurer:
@franz1981 sorry not yet... just returned from vacation and trying to process my backlog.

franz1981:
@normanmaurer np bud, mine was more a broadcast curiosity :) 
Sadly (and luckly for all the folks on Github) a `@all` don't exist yet AFAIK :+1:

normanmaurer:
I am ok with the change ... I will defer to @trustin @carl-mastrangelo @Scottmitch @ejona86

ejona86:
Removing bounds checking for the unsafe versions is pretty scary to me, as it makes basically every call to ByteBuf the equivalent of an Unsafe call, as it can read/write all memory. That opens a lot of possible vulnerabilities, and it means code calling ByteBuf is now semi-responsible for bounds checking; we'd feel pressure to start merging any PRs that add such bounds checks. That sounds weaker than buffers in C, because at least they know there isn't bounds checking.

The Spectre stuff is semi-related, because it allows the equivalent of disabling bounds checking. If we did something like [index masking](https://webkit.org/blog/8048/what-spectre-and-meltdown-mean-for-webkit/), I could kinda live with that. But I'd still be concerned it'd devolve to having additional bounds checking outside of ByteBuf because this feature may be enabled.

franz1981:
I understand all your points: good points indeed :)
I can just quote a sentence from one of the authors of Agrona's lib that is shipping a similar feature:
> Bounds checking is one of those thorny issues and topics of ongoing debate. Avoiding bounds checks can result in faster code, but introduces the potential to cause a segfault and bring down the JVM. Agrona's buffers give you the choice to disable bounds checking through the commandline property agrona.disable.bounds.checks, but bounds check by default. This means that their use is safe, but if application profiling of tested code determines that bounds checking is a bottleneck then it can be removed.

It is not a (common) user tool but IMO is pretty important as a benchmarking/test tool and in extreme cases as a proper tool. For sure It needs to be enabled by default exactly like others equally dangerous (ie PROP_MODE): it is rather different from the Specter case IMO because there are real use cases where it makes sense and not on any user hands.

franz1981:
@ejona86 +100 for the beautiful link you sent: I'm enjoying it so much 👍

ejona86:
Oh, I agree it's totally different from Spectre. I brought it up because index masking may be a compromise.

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -97,11 +100,18 @@ public int readerIndex() {
         return readerIndex;
     }
 
+    private static void checkIndexesBounds(final int readerIndex, final int writerIndex, final int capacity) {
```
trustin: `checkIndexBounds()`?

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -1355,26 +1369,30 @@ protected final void checkIndex(int index, int fieldLength) {
         checkIndex0(index, fieldLength);
     }
 
-    final void checkIndex0(int index, int fieldLength) {
-        if (isOutOfBounds(index, fieldLength, capacity())) {
+    private static void checkIsOutOfBounds(final int index, final int fieldLength, final int capacity) {
```
trustin: `checkRangeBounds`?

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -44,12 +44,15 @@
 public abstract class AbstractByteBuf extends ByteBuf {
     private static final InternalLogger logger = InternalLoggerFactory.getInstance(AbstractByteBuf.class);
     private static final String PROP_MODE = "io.netty.buffer.bytebuf.checkAccessible";
+    private static final String BOUNDS_MODE = "io.netty.buffer.bytebuf.checkBounds";
```
trustin: - `PROP_MODE` to `PROP_CHECK_ACCESSIBLE` and `BOUNDS_PROP` to `PROP_CHECK_BOUNDS`?
- `buffer.bytebuf` sounds redundant to me although it is consistent with `PROP_MODE`. How about using `io.netty.buffer.checkBounds` and accepting both `io.netty.buffer.bytebuf.checkAccessible` and `io.netty.buffer.checkAccessible`?

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -44,12 +44,15 @@
...
+    private static final boolean checkBounds;
     private static final boolean checkAccessible;
 
     static {
         checkAccessible = SystemPropertyUtil.getBoolean(PROP_MODE, true);
+        checkBounds = SystemPropertyUtil.getBoolean(BOUNDS_MODE, true);
         if (logger.isDebugEnabled()) {
-            logger.debug("-D{}: {}", PROP_MODE, checkAccessible);
+            logger.debug("-D{}: {} -D{}: {}", PROP_MODE, checkAccessible, BOUNDS_MODE, checkBounds);
```
trustin: and how about splitting into two log messages?

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
franz1981: > PROP_MODE to PROP_CHECK_ACCESSIBLE and BOUNDS_PROP to PROP_CHECK_BOUNDS

:+1: 

> How about using io.netty.buffer.checkBounds

:+1: 

> and accepting both io.netty.buffer.bytebuf.checkAccessible and io.netty.buffer.checkAccessible?

Sounds good, may I add some log on deprecated uses of `io.netty.buffer.bytebuf.checkAccessible` too, wdyt?

trustin: > may I add some log on deprecated uses of io.netty.buffer.bytebuf.checkAccessible too, wdyt?

👍

franz1981:
I hope to have addressed all the changes requested, but haven't had yet any chance to try this change with some actual use case (ie benchmarks of Netty's heavy-user application with tons of decoding/encoding): any of you have had time to try it? 
Just curious...with [Aeron](https://github.com/real-logic/aeron)'s data-structure was quite interesting and useful to try it considering that is the same exact feature :+1:

franz1981:
@trustin Hi! Looking at the reviewers summary seems that I haven't addressed all the requested changes, but I can't find any..there is anything I can do to improve what I have done here? :+1:

trustin:
Great job, @franz1981 !

normanmaurer:
So @ejona86 what are your thoughts ?

ejona86:
It does worry me to have such a gaping "unsafe" hole. But my attitude toward the PR has changed. I think it is fine and won't cause much maintenance trouble since calling code already tries to avoid IndexOutOfBoundsException.

normanmaurer:
@ejona86 mind approving then ?

franz1981:
@normanmaurer @ejona86 @trustin I've just taken a look to the awesome presentation of Remi Forax about how to help the JVM to trust final fields on https://speakerdeck.com/forax/java-dot-lang-dot-invoke-whisper-to-the-jit-for-speed-gain .
If this PR is ok and will be merged i'm thinking to provide (in the future) another PR with a per-instance configuration to disable bounds checking where the boolean check won't have any cost when disabled.
In this way anyone using `ByteBuf` on trusty code could disable it leaving the other uses safe.

franz1981:
@normanmaurer thanks to the summer I have some time where I can help :) 
There is anything I need to help/improve on this PR?

franz1981:
@normanmaurer I have already signed it the first PR I have sent some time ago...or it has been expired and I need to sign it twice?

normanmaurer:
@franz1981 stupid me... I did miss-spell your name when looking it up 👎  All good.

Issue #4283, opened 2015-09-25
Allow writing long-length objects in HTTP/2 flow control


nmittler:
This is just to capture the data byte we use for `streamableBytes`.  It should be changed to `long` to support very large objects (e.g. FileRegion).

@Scottmitch FYI

Scottmitch:
@nmittler - So your thinking that each stream's `pendingBytes` should be of data type `long` (and then what ever else needs to change to accommodate this)?

nmittler:
@Scottmitch yeah that's basically it (for now). The `writableBytes` would still be an int, but streams may have more than `MAX_INT` to send.

Issue #7509, opened 2017-12-15
HTTP/2 UniformStreamByteDistributor has an infinite loop problem when transferring content that exceeds 2 GB.


durigon:
### Expected behavior
no infinite loop 

### Actual behavior
HTTP/2 UniformStreamByteDistributor has an infinite loop problem when transferring content that exceeds 2GB.

- [Http2CodecUtil streamableBytes - If pendingBytes is out of the integer range, streamableBytes may be zero.](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecUtil.java#L226)
- [Since there is a frame and windowSize is greater than 0, writing 0 bytes is infinitely repeated.](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/UniformStreamByteDistributor.java#L170)
- [UniformStreamByteDistributor  StreamState.streamableBytes](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/UniformStreamByteDistributor.java#L118)

- Workaround : Add flow controller so that HTTP/2 stream pending bytes do not exceed the integer(pendingBytes, streamableBytes).

### Steps to reproduce
Use curl with limit-rate to request more than 2Gbytes of content In multiple terminal(More than 6 terminals)
`curl -v https://localhost/3gbytes.bin -o/dev/null --http2 --insecure --limit-rate 1M `

### Netty version
netty-4.1.12.Final

### JVM version (e.g. `java -version`)
#java -version
java version "1.8.0_131"
Java(TM) SE Runtime Environment (build 1.8.0_131-b11)

### OS version (e.g. `uname -a`)
uname -a
Linux sunggonsong 2.6.32-696.13.2.el6.x86_64 #1 SMP Thu Oct 5 21:22:16 UTC 2017 x86_64 x86_64 x86_64 GNU/Linu

normanmaurer:
@durigon just to ensure did you try 4.1.18.Final as well ?

durigon:
#4283 and #4585 are still in process, so it seems that it has not been fixed.

Scottmitch:
I think this will be resolved once https://github.com/netty/netty/issues/4283 is resolved.

Scottmitch:
@durigon - Can you run your tests with https://github.com/netty/netty/pull/7523 ?

durigon:
[...]

PR #7523, opened 2017-12-19
HTTP/2 support pending data larger than Integer.MAX_VALUE

Scottmitch:
Motivation:
Currently the remote flow controller limits the maximum amount of pending data to Integer.MAX_VALUE. The overflow handling is also not very graceful in that it may lead to infinite loops, or otherwise no progress being made.

Modifications:
- StreamByteDistributor and RemoteFlowController should support pending bytes of type long.

Result:
Fixes https://github.com/netty/netty/issues/4283


Scottmitch:
Also may make the symptoms from https://github.com/netty/netty/issues/7509 go away.

Scottmitch:
verified by @durigon (see https://github.com/netty/netty/issues/7509#issuecomment-352983892)

PR #7524, opened 2017-12-19
Fixed JMH ByteBuf benchmark to avoid dead code elimination

franz1981:
Motivation:

The JMH doc suggests to use BlackHoles to avoid dead code elimination hence would be better to follow this best practice.

Modifications:

Each benchmark method is returning the ByteBuf/ByteBuffer to avoid the JVM to perform any dead code elimination.

Result:

The results are more reliable and comparable to the others provided by other ByteBuf benchmarks (eg HeapByteBufBenchmark)


normanmaurer:
Cherry-picked into 4.1 (1cf268724430131ce5c7f4c2068498d429125ac0). 

@franz1981 thanks!

PR #7525, opened 2017-12-19
Add a hint of ownership transfer when calling EmbeddedChannel.read*()…

normanmaurer:
… methods.

Motivation:

As shown in issues it is sometimes hard to understand why a leak was reported when the user just calles EmbeddedChannel.readInbound() / EmbeddedChannel.readOutbound() and drop the message on the floor.

Modifications:

Add a hint before handover the message to the user and transfer the ownership.

Result:

Easier debugging of leaks caused by EmbeddedChannel.read*().

### DIFF
--- a/transport/src/main/java/io/netty/channel/embedded/EmbeddedChannel.java
+++ b/transport/src/main/java/io/netty/channel/embedded/EmbeddedChannel.java
@@ -40,7 +40,6 @@ import io.netty.util.ReferenceCountUtil;
 import io.netty.util.internal.ObjectUtil;
 import io.netty.util.internal.PlatformDependent;
 import io.netty.util.internal.RecyclableArrayList;
-import io.netty.util.internal.UnstableApi;
 import io.netty.util.internal.logging.InternalLogger;
 import io.netty.util.internal.logging.InternalLoggerFactory;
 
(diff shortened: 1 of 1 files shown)


transport/src/main/java/io/netty/channel/embedded/EmbeddedChannel.java
```diff
@@ -289,15 +288,23 @@ public boolean isActive() {
      */
     @SuppressWarnings("unchecked")
     public <T> T readInbound() {
-        return (T) poll(inboundMessages);
+        T message = (T) poll(inboundMessages);
+        if (message != null) {
+            ReferenceCountUtil.touch(message, "Caller of readInbound() will handle the message from this point");
```
normanmaurer: The message is inspired by https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java#L1021.

normanmaurer:
Cherry-picked into 4.1 (0c5014b1057f9549c3fb346890162c66bc3631ea)

transport/src/main/java/io/netty/channel/embedded/EmbeddedChannel.java
carl-mastrangelo: ReferenceCountUtil.touch handles null I think.

PR #7527, opened 2017-12-20
Fix DefaultHttpDataFactory cleanup bug

ngocdaothanh:
Fix DefaultHttpDataFactory cleanup bug

Motivation:

DefaultHttpDataFactory uses HttpRequest as map keys.

Because of the implementation of "hashCode"" and "equals" in DefaultHttpRequest,
if we use normal maps, HttpDatas of different requests may end up in the same map entry,
causing cleanup bug.

Consider this example:
- Suppose that request1 is equal to request2, causing their HttpDatas to be stored in one single map entry.
- request1 is cleaned up first, while request2 is still being decoded.
- Consequently request2's HttpDatas are suddenly gone, causing NPE, or worse loss of data.

This bug can be reproduced by starting the HttpUploadServer example,
then run this command:

```
ab -T 'application/x-www-form-urlencoded' -n 100 -c 5 -p post.txt http://localhost:8080/form
```

post.txt file content:

```
a=1&b=2
```

There will be errors like this:

```
java.lang.NullPointerException
	at io.netty.handler.codec.http.multipart.MemoryAttribute.getValue(MemoryAttribute.java:64)
	at io.netty.handler.codec.http.multipart.MixedAttribute.getValue(MixedAttribute.java:243)
	at io.netty.example.http.upload.HttpUploadServerHandler.writeHttpData(HttpUploadServerHandler.java:271)
	at io.netty.example.http.upload.HttpUploadServerHandler.readHttpDataChunkByChunk(HttpUploadServerHandler.java:230)
	at io.netty.example.http.upload.HttpUploadServerHandler.channelRead0(HttpUploadServerHandler.java:193)
	at io.netty.example.http.upload.HttpUploadServerHandler.channelRead0(HttpUploadServerHandler.java:66)
	at io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:105)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
	at io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:102)
	at io.netty.handler.codec.MessageToMessageCodec.channelRead(MessageToMessageCodec.java:111)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
	at io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:310)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:284)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1412)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
	at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:943)
	at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:141)
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:645)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:580)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:497)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:459)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:886)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.lang.Thread.run(Thread.java:748)
```

Modifications:

Keep identity of requests by using IdentityHashMap

Result:

DefaultHttpDataFactory is fixed.

The ConcurrentHashMap is replaced with a synchronized map, but I think the performance won't be affected much in real web apps.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/multipart/DefaultHttpDataFactory.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/multipart/DefaultHttpDataFactory.java
@@ -15,25 +15,28 @@
  */
 package io.netty.handler.codec.http.multipart;
 
+import io.netty.handler.codec.http.DefaultHttpRequest;
 import io.netty.handler.codec.http.HttpConstants;
 import io.netty.handler.codec.http.HttpRequest;
-import io.netty.util.internal.PlatformDependent;
 
 import java.io.IOException;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.IdentityHashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 
 /**
- * Default factory giving Attribute and FileUpload according to constructor
+ * Default factory giving {@link Attribute} and {@link FileUpload} according to constructor.
  *
- * Attribute and FileUpload could be :<br>
- * - MemoryAttribute, DiskAttribute or MixedAttribute<br>
- * - MemoryFileUpload, DiskFileUpload or MixedFileUpload<br>
- * according to the constructor.
+ * <p>According to the constructor, {@link Attribute} and {@link FileUpload} can be:</p>
+ * <ul>
+ * <li>MemoryAttribute, DiskAttribute or MixedAttribute</li>
+ * <li>MemoryFileUpload, DiskFileUpload or MixedFileUpload</li>
+ * </ul>
  */
 public class DefaultHttpDataFactory implements HttpDataFactory {
 
@@ -57,9 +60,14 @@ public class DefaultHttpDataFactory implements HttpDataFactory {
     private Charset charset = HttpConstants.DEFAULT_CHARSET;
 
     /**
-     * Keep all HttpDatas until cleanAllHttpData() is called.
+     * Keep all {@link HttpData}s until cleaning methods are called.
+     * We need to use {@link IdentityHashMap} because different requests may be equal.
+     * See {@link DefaultHttpRequest#hashCode} and {@link DefaultHttpRequest#equals}.
+     * Similarly, when removing data items, we need to check their identities because
+     * different data items may be equal.
      */
-    private final Map<HttpRequest, List<HttpData>> requestFileDeleteMap = PlatformDependent.newConcurrentHashMap();
+    private final Map<HttpRequest, List<HttpData>> requestFileDeleteMap =
+            Collections.synchronizedMap(new IdentityHashMap<HttpRequest, List<HttpData>>());
 
     /**
      * HttpData will be in memory if less than default size (16KB).
@@ -109,7 +117,7 @@ public class DefaultHttpDataFactory implements HttpDataFactory {
     }
 
     /**
-     * @return the associated list of Files for the request
+     * @return the associated list of {@link HttpData} for the request
      */
     private List<HttpData> getList(HttpRequest request) {
         List<HttpData> list = requestFileDeleteMap.get(request);
@@ -125,15 +133,15 @@ public class DefaultHttpDataFactory implements HttpDataFactory {
         if (useDisk) {
             Attribute attribute = new DiskAttribute(name, charset);
             attribute.setMaxSize(maxSize);
-            List<HttpData> fileToDelete = getList(request);
-            fileToDelete.add(attribute);
+            List<HttpData> list = getList(request);
+            list.add(attribute);
             return attribute;
         }
         if (checkSize) {
             Attribute attribute = new MixedAttribute(name, minSize, charset);
             attribute.setMaxSize(maxSize);
-            List<HttpData> fileToDelete = getList(request);
-            fileToDelete.add(attribute);
+            List<HttpData> list = getList(request);
+            list.add(attribute);
             return attribute;
         }
         MemoryAttribute attribute = new MemoryAttribute(name);
@@ -146,15 +154,15 @@ public class DefaultHttpDataFactory implements HttpDataFactory {
         if (useDisk) {
             Attribute attribute = new DiskAttribute(name, definedSize, charset);
             attribute.setMaxSize(maxSize);
-            List<HttpData> fileToDelete = getList(request);
-            fileToDelete.add(attribute);
+            List<HttpData> list = getList(request);
+            list.add(attribute);
             return attribute;
         }
         if (checkSize) {
             Attribute attribute = new MixedAttribute(name, definedSize, minSize, charset);
             attribute.setMaxSize(maxSize);
-            List<HttpData> fileToDelete = getList(request);
-            fileToDelete.add(attribute);
+            List<HttpData> list = getList(request);
+            list.add(attribute);
             return attribute;
         }
         MemoryAttribute attribute = new MemoryAttribute(name, definedSize);
@@ -186,16 +194,16 @@ public class DefaultHttpDataFactory implements HttpDataFactory {
                 attribute.setMaxSize(maxSize);
             }
             checkHttpDataSize(attribute);
-            List<HttpData> fileToDelete = getList(request);
-            fileToDelete.add(attribute);
+            List<HttpData> list = getList(request);
+            list.add(attribute);
             return attribute;
         }
         if (checkSize) {
             Attribute attribute = new MixedAttribute(name, value, minSize, charset);
             attribute.setMaxSize(maxSize);
             checkHttpDataSize(attribute);
-            List<HttpData> fileToDelete = getList(request);
-            fileToDelete.add(attribute);
+            List<HttpData> list = getList(request);
+            list.add(attribute);
             return attribute;
         }
         try {
@@ -217,8 +225,8 @@ public class DefaultHttpDataFactory implements HttpDataFactory {
                     contentTransferEncoding, charset, size);
             fileUpload.setMaxSize(maxSize);
             checkHttpDataSize(fileUpload);
-            List<HttpData> fileToDelete = getList(request);
-            fileToDelete.add(fileUpload);
+            List<HttpData> list = getList(request);
+            list.add(fileUpload);
             return fileUpload;
         }
         if (checkSize) {
@@ -226,8 +234,8 @@ public class DefaultHttpDataFactory implements HttpDataFactory {
                     contentTransferEncoding, charset, size, minSize);
             fileUpload.setMaxSize(maxSize);
             checkHttpDataSize(fileUpload);
-            List<HttpData> fileToDelete = getList(request);
-            fileToDelete.add(fileUpload);
+            List<HttpData> list = getList(request);
+            list.add(fileUpload);
             return fileUpload;
         }
         MemoryFileUpload fileUpload = new MemoryFileUpload(name, filename, contentType,
@@ -239,20 +247,42 @@ public class DefaultHttpDataFactory implements HttpDataFactory {
 
     @Override
     public void removeHttpDataFromClean(HttpRequest request, InterfaceHttpData data) {
-        if (data instanceof HttpData) {
-            List<HttpData> fileToDelete = getList(request);
-            fileToDelete.remove(data);
+        if (!(data instanceof HttpData)) {
+            return;
+        }
+
+        // Do not use getList because it adds empty list to requestFileDeleteMap
+        // if request is not found
+        List<HttpData> list = requestFileDeleteMap.get(request);
+        if (list == null) {
+            return;
+        }
+
+        // Can't simply call list.remove(data), because different data items may be equal.
+        // Need to check identity.
+        Iterator<HttpData> i = list.iterator();
+        while (i.hasNext()) {
+            HttpData n = i.next();
+            if (n == data) {
+                i.remove();
+
+                // Remove empty list to avoid memory leak
+                if (list.isEmpty()) {
+                    requestFileDeleteMap.remove(request);
+                }
+
+                return;
+            }
         }
     }
 
     @Override
     public void cleanRequestHttpData(HttpRequest request) {
-        List<HttpData> fileToDelete = requestFileDeleteMap.remove(request);
-        if (fileToDelete != null) {
-            for (HttpData data: fileToDelete) {
-                data.delete();
+        List<HttpData> list = requestFileDeleteMap.remove(request);
+        if (list != null) {
+            for (HttpData data : list) {
+                data.release();
             }
-            fileToDelete.clear();
         }
     }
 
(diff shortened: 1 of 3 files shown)


ngocdaothanh:
This is a very serious bug.
Can the Netty team cut a new version very soon please (perfect as an Xmas present)?
This bug was introduced long ago when branch 4.1 was first released.

Scottmitch:
@ngocdaothanh - can you provide a unit test too please?

ngocdaothanh:
OK I'll add one.

ngocdaothanh:
I've added `DefaultHttpDataFactoryTest`

ngocdaothanh:
Added copyright header to fix compilation error

codec-http/src/test/java/io/netty/handler/codec/http/multipart/DefaultHttpDataFactoryTest.java
```diff
@@ -0,0 +1,60 @@
...
+        List<HttpData> list1 = factory.getList(req1);
+        assertEquals(0, list1.size());
+
+        // Add something to list1
+        list1.add(new MemoryAttribute("name", "value"));
+        assertEquals(1, list1.size());
+
+        // req2 equals req1
+        HttpRequest req2 = new DefaultHttpRequest(HTTP_1_1, POST, "/form");
+        assertTrue(req1.hashCode() == req2.hashCode());
```
Scottmitch: assetEquals(req1.hashCode(), req2.hashCode());

codec-http/src/test/java/io/netty/handler/codec/http/multipart/DefaultHttpDataFactoryTest.java
```diff
@@ -0,0 +1,60 @@
...
+        assertTrue(req1.equals(req2));
+
+        // But req2 should be identified differently from req1
+        List<HttpData> list2 = factory.getList(req2);
+        assertEquals(0, list2.size());
+
+        // req1 should be identified by itself
+        List<HttpData> list3 = factory.getList(req1);
+        assertEquals(1, list3.size());
+        assertTrue(list1.get(0) == list3.get(0));
```
Scottmitch: `assertTrue` -> `assertEquals`

codec-http/src/main/java/io/netty/handler/codec/http/multipart/DefaultHttpDataFactory.java
```diff
@@ -110,8 +115,10 @@ public void setMaxLimit(long maxSize) {
 
     /**
      * @return the associated list of Files for the request
+     *
+     * Visible for testing.
      */
-    private List<HttpData> getList(HttpRequest request) {
+    List<HttpData> getList(HttpRequest request) {
```
Scottmitch: it would be ideal if we could test using the public API of the class instead of depending on the internals (which are subject to change over time).

ngocdaothanh:
> it would be ideal if we could test using the public API

Thanks for the review.
I've update the PR in that direction.

Following that direction, I've also found and fix 2 more bugs, at `cleanAllHttpData` and  `removeHttpDataFromClean`, a big win.

ngocdaothanh:
As explained, I think this is a very serious bug, but not noticed for a long time (around 1 year), because luckily (or unluckily) most requests, attributes, and file uploads are different.

I hope a new version of Netty will be released very soon.

codec-http/src/main/java/io/netty/handler/codec/http/multipart/DefaultHttpDataFactory.java
```diff
@@ -241,7 +246,16 @@ public FileUpload createFileUpload(HttpRequest request, String name, String file
     public void removeHttpDataFromClean(HttpRequest request, InterfaceHttpData data) {
         if (data instanceof HttpData) {
             List<HttpData> fileToDelete = getList(request);
-            fileToDelete.remove(data);
+
+            // Can't simply call fileToDelete.remove(data), need to check identity
```
normanmaurer: can you explain why ?

ngocdaothanh: Thanks, I've improved the comment to explain why.

codec-http/src/test/java/io/netty/handler/codec/http/multipart/DefaultHttpDataFactoryTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+        // Clean up by req1
+        factory.cleanRequestHttpData(req1);
+
+        // Assert that data belonging to req1 has been cleaned up, but not req2
+        assertNull(attribute1.getByteBuf());
+        assertNull(file1.getByteBuf());
+        assertNotNull(attribute2.getByteBuf());
+        assertNotNull(file2.getByteBuf());
+    }
```
Scottmitch: The reference counts here are not cleaned up ... but that maybe because of a bug in `DefaultHttpDataFactory`. Lets discuss on the previous comments.

codec-http/src/main/java/io/netty/handler/codec/http/multipart/DefaultHttpDataFactory.java
```diff
@@ -261,7 +278,9 @@ public void cleanAllHttpData() {
         Iterator<Entry<HttpRequest, List<HttpData>>> i = requestFileDeleteMap.entrySet().iterator();
         while (i.hasNext()) {
             Entry<HttpRequest, List<HttpData>> e = i.next();
-            i.remove();
+
+            // Calling i.remove() here will cause "java.lang.IllegalStateException: Entry was removed"
+            // at e.getValue() below
 
             List<HttpData> fileToDelete = e.getValue();
             if (fileToDelete != null) {
```
Scottmitch: the reference count ownership of this class seems suspect. For example in `cleanRequestHttpData` and `cleanAllHttpData` we call `delete()`, but it seems like we should instead be calling `release()` (which then calls `delete()` if the reference count reaches 0).

codec-http/src/main/java/io/netty/handler/codec/http/multipart/DefaultHttpDataFactory.java
```diff
@@ -241,7 +248,17 @@ public FileUpload createFileUpload(HttpRequest request, String name, String file
     public void removeHttpDataFromClean(HttpRequest request, InterfaceHttpData data) {
         if (data instanceof HttpData) {
             List<HttpData> fileToDelete = getList(request);
```
Scottmitch: this method's usage of `getList` seems like it maybe problematic. `getList` will create a new list (which contains `request`) and add it to `requestFileDeleteMap` if a pre-existing list cannot be found. However this method will not remove that list if there are no more items in the list after removal, which may lead to a memory leak. If `fileToDelete.isEmpty()` after an item is removed, shouldn't we also remove the associated entry from `requestFileDeleteMap`?

Also it seems strange to create a new list and add it to `requestFileDeleteMap` if it doesn't exist in this case. Would it make sense to just release the `request` and skip the add/remove from the `requestFileDeleteMap ` map?

ngocdaothanh: 👍 Fixed

codec-http/src/main/java/io/netty/handler/codec/http/multipart/DefaultHttpDataFactory.java
ngocdaothanh: 👍 Fixed and added test

codec-http/src/main/java/io/netty/handler/codec/http/multipart/DefaultHttpDataFactory.java
```diff
@@ -209,67 +225,99 @@ public Attribute createAttribute(HttpRequest request, String name, String value)
     }
 
     @Override
-    public FileUpload createFileUpload(HttpRequest request, String name, String filename,
+    public FileUpload createFileUpload(
+            HttpRequest request, String name, String filename,
             String contentType, String contentTransferEncoding, Charset charset,
-            long size) {
```
Scottmitch: can you revert the formatting and other unrelated related changes? it makes merging/reviewing/maintenance more difficult. if you think they are beneficial please submit them in a followup PR.

codec-http/src/main/java/io/netty/handler/codec/http/multipart/DefaultHttpDataFactory.java
```diff
@@ -209,67 +225,99 @@ public Attribute createAttribute(HttpRequest request, String name, String value)
...
+        while (i.hasNext()) {
+            HttpData n = i.next();
+            if (n == data) {
+                i.remove();
+                break;
+            }
+        }
+
+        // Remove empty list to avoid memory leak
+        if (list.isEmpty()) {
```
Scottmitch: this can be moved into the `if (n == data)` block above, after the `i.remove()` call because that is the only place we are modifying the list.

codec-http/src/main/java/io/netty/handler/codec/http/multipart/DefaultHttpDataFactory.java
```diff
@@ -209,67 +225,99 @@ public Attribute createAttribute(HttpRequest request, String name, String value)
...
-            fileToDelete.clear();
+        List<HttpData> list = requestFileDeleteMap.remove(request);
+        if (list == null) {
+            return;
+        }
+
+        for (HttpData data: list) {
+            data.release();
         }
+        list.clear();
```
Scottmitch: is this necessary? can we just let it be GCed instead of clearing it out first?

codec-http/src/main/java/io/netty/handler/codec/http/multipart/DefaultHttpDataFactory.java
```diff
@@ -209,67 +225,99 @@ public Attribute createAttribute(HttpRequest request, String name, String value)
...
+            // Calling i.remove() here will cause "java.lang.IllegalStateException: Entry was removed"
+            // at e.getValue() below
+
+            List<HttpData> list = e.getValue();
+            if (list != null) {
+                for (HttpData data : list) {
+                    data.release();
                 }
-                fileToDelete.clear();
+                list.clear();
```
Scottmitch: is this necessary? can we just let it be GCed instead of clearing it out first?

codec-http/src/main/java/io/netty/handler/codec/http/multipart/DefaultHttpDataFactory.java
```diff
@@ -209,67 +225,99 @@ public Attribute createAttribute(HttpRequest request, String name, String value)
...
 
-            List<HttpData> fileToDelete = e.getValue();
-            if (fileToDelete != null) {
-                for (HttpData data : fileToDelete) {
-                    data.delete();
+            // Calling i.remove() here will cause "java.lang.IllegalStateException: Entry was removed"
+            // at e.getValue() below
+
+            List<HttpData> list = e.getValue();
+            if (list != null) {
```
Scottmitch: can `list` ever be `null`? seems like that would be a programming error. if so consider just doing `assert list != null` and remove the runtime check.

codec-http/src/main/java/io/netty/handler/codec/http/multipart/DefaultHttpDataFactory.java
ngocdaothanh: Removed `list.clear()`

codec-http/src/main/java/io/netty/handler/codec/http/multipart/DefaultHttpDataFactory.java
ngocdaothanh: Remove unnecessary check

codec-http/src/main/java/io/netty/handler/codec/http/multipart/DefaultHttpDataFactory.java
ngocdaothanh: Removed `list.clear()`

codec-http/src/main/java/io/netty/handler/codec/http/multipart/DefaultHttpDataFactory.java
ngocdaothanh: Improved code

codec-http/src/main/java/io/netty/handler/codec/http/multipart/DefaultHttpDataFactory.java
ngocdaothanh: Reverted unnecessary formatting; we still should rename `fileToDelete` to `list`

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostStandardRequestDecoder.java
```diff
@@ -653,23 +653,19 @@ private static String decodeAttribute(String s, Charset charset) {
...
         destroyed = true;
 
         if (undecodedChunk != null && undecodedChunk.refCnt() > 0) {
             undecodedChunk.release();
             undecodedChunk = null;
         }
-
-        // release all data which was not yet pulled
-        for (int i = bodyListHttpDataRank; i < bodyListHttpData.size(); i++) {
-            bodyListHttpData.get(i).release();
```
ngocdaothanh: This `release` cause HttpPostRequestDecoderTest#testFormEncodeIncorrect to fail; we already release data at `cleanFiles` above

codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostStandardRequestDecoder.java
```diff
@@ -653,23 +653,19 @@ private static String decodeAttribute(String s, Charset charset) {
      */
     @Override
     public void destroy() {
-        checkDestroyed();
```
ngocdaothanh: `checkDestroyed` will be done at `cleanFiles` (one line below)

ngocdaothanh:
@Scottmitch Thanks, I've addressed for comments

Scottmitch:
@normanmaurer - I'll assign to you ... please feel free to pull in after you are satisfied.

normanmaurer:
Cherry-picked into 4.1 (2b4f6677917cbe4121c2b748ba7002f8ba4bb8b5)

Issue #7343, opened 2017-10-25
Investigate whether it's possible not to use finalizer for Recycler$WeakOrderQueue


luengnat:
Recycler$WeakOrderQueue has a finalize() method. It could potentially harm the GC performance. We should investigate whether it's possible not to use finalize() method.

johnou:
@luengnat feel free to investigate, you can see https://github.com/netty/netty/pull/5577 for background on why it was used.

johnou:
While it could potentially harm GC performance I would be interested in seeing a reproducer, if not then maybe just close this issue if it's not a problem..

johnou:
@normanmaurer something for the new ObjectCleaner?

normanmaurer:
@johnou once in, yes

normanmaurer:
Please check https://github.com/netty/netty/pull/7530

normanmaurer:
This was fixed

PR #7530, opened 2017-12-21
Replace finalizer() usage in Recycler.WeakOrderQueue with ObjectClean…

normanmaurer:
…er usage.

Motivation:

We recently introduced ObjectCleaner which can be used to ensure some cleanup action is done once an object becomes weakable reachable. We should use this in Recycler.WeakOrderQueue to reduce the overhead of using a finalizer() (which will cause the GC to process it two times).

Modifications:

Replace finalizer() usage with ObjectCleaner

Result:

Fixes [#7343]

### DIFF
--- a/common/src/main/java/io/netty/util/Recycler.java
+++ b/common/src/main/java/io/netty/util/Recycler.java
@@ -17,6 +17,7 @@
 package io.netty.util;
 
 import io.netty.util.concurrent.FastThreadLocal;
+import io.netty.util.internal.ObjectCleaner;
 import io.netty.util.internal.SystemPropertyUtil;
 import io.netty.util.internal.logging.InternalLogger;
 import io.netty.util.internal.logging.InternalLoggerFactory;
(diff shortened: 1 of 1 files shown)


normanmaurer:
also @johnou

common/src/main/java/io/netty/util/Recycler.java
```diff
@@ -236,41 +237,93 @@ public void recycle(Object object) {
 
         // Let Link extend AtomicInteger for intrinsics. The Link itself will be used as writerIndex.
         @SuppressWarnings("serial")
-        private static final class Link extends AtomicInteger {
+        static final class Link extends AtomicInteger {
             private final DefaultHandle<?>[] elements = new DefaultHandle[LINK_CAPACITY];
 
             private int readIndex;
             private Link next;
```
Scottmitch: these member variables are accessed outside the class ... should they be package private?

common/src/main/java/io/netty/util/Recycler.java
```diff
@@ -236,41 +237,93 @@ public void recycle(Object object) {
...
+            final WeakOrderQueue queue = new WeakOrderQueue(stack, thread);
             // Done outside of the constructor to ensure WeakOrderQueue.this does not escape the constructor and so
             // may be accessed while its still constructed.
             stack.setHead(queue);
+
+            // We need to reclaim all space that was reserved by this WeakOrderQueue so we not run out of space in
+            // the stack. This is needed as we not have a good life-time control over the queue as it is used in a
+            // WeakHashMap which will drop it at any time.
+            final Head head = queue.head;
+            ObjectCleaner.register(queue, head);
```
Scottmitch: I wonder if it makes sense to be able to create an ObjectCleaner instance ... so we can partition its usage and reduce contention on the static state within (e.g. the `Set`, ...). any thoughts? This would give us flexibility to partition it depending upon the use case (e.g. a static instance for all `Recycler` objects, another static instance for all `FastThreadLocal` objects, individual `Recycler` objects, etc...).

Scottmitch:
General question about `ObjectCleaner`. IIUC we are trading a `finalize` method for each object vs each "register" call will introduce:
-` WeakReference`
- `Runnable`

then in general for `ObjectCleaner` some static state:
- concurrent `Set`,
- `ReferenceQueue`
- background thread(s)
- a few atomic operations

It is not clear to me that the `ObjectCleaner` will provide an overall win here (memory, GC time, throughput, etc...). How can we quantify the differences before and after?

johnou:
@Scottmitch need to also consider that ```finalize``` is being deprecated.

normanmaurer:
@Scottmitch I think one of the main reasons for this changes are what @johnou said... `finalize` is deprecated now and we should try to avoid using it if we can.

common/src/main/java/io/netty/util/Recycler.java
ejona86: It seems like ConcurrentHashMap should do a good job of avoiding contention. If it becomes a problem then we can adjust the load factor. The bigger thing to me is only having one cleaner thread; I'd hope to keep it that way.

common/src/main/java/io/netty/util/Recycler.java
normanmaurer: yes

PR #7531, opened 2017-12-21
Add 32 bytes overhead per header entry when calculating headers length in HPackDecoder

### LINKED ISSUE #7514: hpackdecoder doesn't account for header overhead
Http/2 says for max header list size:

>    SETTINGS_MAX_HEADER_LIST_SIZE (0x6):  This advisory setting informs a
>       peer of the maximum size of header list that the sender is
>       prepared to accept, in octets.  The value is based on the
>       uncompressed size of header fields, including the length of the
>       name and value in octets plus an overhead of 32 octets for each
>       header field.
> 
>       For any given request, a lower limit than what is advertised MAY
>       be enforced.  The initial value of this setting is unlimited.

In HpackEncoder, we account for the 32 bytes, but when decoding, we don't.

The way I found this was by setting up a finagle proxy talking to a finagle server.  When I curled the server directly, I was way beneath the limit, but when I curled the proxy, I got 431s, since it correctly accounted for the 32 bytes overhead on the client-side in HpackEncoder.

madgnome:
**Motivation:**

According to the HTTP/2 Spec:

> SETTINGS_MAX_HEADER_LIST_SIZE (0x6): This advisory setting informs a
> peer of the maximum size of header list that the sender is
> prepared to accept, in octets. The value is based on the
> uncompressed size of header fields, including the length of the
> name and value in octets plus an overhead of 32 octets for each
> header field.

We were accounting for the 32 bytes when encoding in HpackEncoder,
but not when decoding in HPackDecoder.

**Modifications:**

- Add 32 bytes to the header list length for each entry when decoding
with HPackDecoder.

**Result:**

- We account for the 32 bytes overhead by header entry in HPackDecoder
- Fixes #7514

### DIFF
--- a/codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
+++ b/codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
@@ -35,14 +35,19 @@ import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.util.internal.StringUtil;
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 
 import static io.netty.handler.codec.http2.HpackDecoder.decodeULE128;
 import static io.netty.handler.codec.http2.Http2HeadersEncoder.NEVER_SENSITIVE;
 import static io.netty.util.AsciiString.EMPTY_STRING;
 import static io.netty.util.AsciiString.of;
 import static java.lang.Integer.MAX_VALUE;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.Matchers.greaterThanOrEqualTo;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 import static org.mockito.Mockito.mock;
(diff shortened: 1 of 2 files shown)


codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
```diff
@@ -477,4 +480,34 @@ public void testDecodeCountsNamesOnlyOnce() throws Http2Exception {
...
+            hpackEncoder.encodeHeaders(1, in, toEncode, NEVER_SENSITIVE);
+
+            Http2Headers decoded = new DefaultHttp2Headers();
+            try {
+                // SETTINGS_MAX_HEADER_LIST_SIZE is big enough for the header to fit...
+                assertThat(hpackDecoder.getMaxHeaderListSize(), is(greaterThanOrEqualTo(headerSize)));
+
+                hpackDecoder.decode(1, in, decoded);
+                fail();
+            } catch (Http2Exception e) {
```
normanmaurer: just change this to `catch (Http2Exception.HeaderListSizeException expected)` and remove the `assertTrue(...)` below ?

normanmaurer:
Also @mosesn

codec-http2/src/test/java/io/netty/handler/codec/http2/HpackDecoderTest.java
```diff
@@ -52,6 +57,9 @@
 import static org.mockito.Mockito.verifyNoMoreInteractions;
 
 public class HpackDecoderTest {
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
```
carl-mastrangelo: fyi, in the future, this should be final

Scottmitch: https://github.com/netty/netty/pull/7596

PR #7533, opened 2017-12-21
Use 198.51.100.254 as BAD_HOST for tests.

normanmaurer:
Motivation:

At the moment we use netty.io as BAD_HOST with an port that we know is timing out. This may change in the future so we should better use 198.51.100.254 which is specified as "for documentation only".

Modifications:

Replace netty.io with 198.51.100.254 in tests that depend on BAD_HOST.

Result:

More future proof code.

### DIFF
--- a/testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketTestPermutation.java
+++ b/testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketTestPermutation.java
@@ -44,7 +44,7 @@ import java.util.List;
 
 public class SocketTestPermutation {
 
-    static final String BAD_HOST = SystemPropertyUtil.get("io.netty.testsuite.badHost", "netty.io");
+    static final String BAD_HOST = SystemPropertyUtil.get("io.netty.testsuite.badHost", "198.51.100.254");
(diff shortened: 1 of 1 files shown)


Scottmitch:
> "for documentation only"

TIL. Where is this declared/defined?

normanmaurer:
@Scottmitch https://tools.ietf.org/html/rfc6890#section-2.2.2 ... didn't know this as well until @weissi told me about it :)

normanmaurer:
Cherry-picked into 4.1 (b6c42f6547e6621dd924707167db83b15652d7f6) and 4.0 (c6a9c4d2bd3c0969b13ac9f01d6637dd566aac57)

testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketTestPermutation.java
```diff
@@ -44,7 +44,7 @@
 
 public class SocketTestPermutation {
 
-    static final String BAD_HOST = SystemPropertyUtil.get("io.netty.testsuite.badHost", "netty.io");
+    static final String BAD_HOST = SystemPropertyUtil.get("io.netty.testsuite.badHost", "198.51.100.254");
```
carl-mastrangelo: question: will this fail the right way in a ipv6 only environment?

PR #7534, opened 2017-12-21
Ensure ObjectCleaner will also be used when FastThreadLocal.set is used.

normanmaurer:
Motivation:

e329ca1 introduced the user of ObjectCleaner in FastThreadLocal but we missed the case to register our cleaner task if FastThreadLocal.set was called only.

Modifications:

- Use ObjectCleaner also when FastThreadLocal.set is used.
- Add test case.

Result:

ObjectCleaner is always used.

### DIFF
--- a/common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
+++ b/common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
@@ -134,13 +134,18 @@ public class FastThreadLocal<V> {
      */
     @SuppressWarnings("unchecked")
     public final V get() {
-        final InternalThreadLocalMap threadLocalMap = InternalThreadLocalMap.get();
+        InternalThreadLocalMap threadLocalMap = InternalThreadLocalMap.get();
         Object v = threadLocalMap.indexedVariable(index);
         if (v != InternalThreadLocalMap.UNSET) {
             return (V) v;
         }
 
         V value = initialize(threadLocalMap);
+        registerCleaner(threadLocalMap);
+        return value;
+    }
+
+    private void registerCleaner(final InternalThreadLocalMap threadLocalMap) {
         Thread current = Thread.currentThread();
         if (!FastThreadLocalThread.willCleanupFastThreadLocals(current)) {
             // We will need to ensure we will trigger remove(InternalThreadLocalMap) so everything will be released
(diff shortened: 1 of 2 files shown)


normanmaurer:
@zpencer PTAL

common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
```diff
@@ -190,7 +194,13 @@ private V initialize(InternalThreadLocalMap threadLocalMap) {
      */
     public final void set(V value) {
         if (value != InternalThreadLocalMap.UNSET) {
-            set(InternalThreadLocalMap.get(), value);
+            InternalThreadLocalMap threadLocalMap = InternalThreadLocalMap.get();
+            boolean alreadySet = threadLocalMap.isIndexedVariableSet(index);
```
Scottmitch: consider making an overload of `set` which returns `alreadySet` because it already knows this information. it could be `private static`.

common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
```diff
@@ -134,13 +134,18 @@ public FastThreadLocal() {
      */
     @SuppressWarnings("unchecked")
     public final V get() {
-        final InternalThreadLocalMap threadLocalMap = InternalThreadLocalMap.get();
```
Scottmitch: was removing `final` intentional and for stylistic reasons?

normanmaurer: yeah I added it before when I needed it as I used it in the `Runnable` but as I moved this to an extra method I not need it anymore

common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
normanmaurer: This is not alway true... let me just merge it as it is for now

normanmaurer:
Cherry-picked into 4.1 (e004b4a35) and 4.0 (09484de7692a4643bb422f6f92392cd45de6f78b)

common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
Scottmitch: PR https://github.com/netty/netty/pull/7541 is what I had in mind

PR #7535, opened 2017-12-22
HpackDecoder#addHeader has an unused parameter

Scottmitch:
Motivation:
HpackDecoder#addHeader takes in the streamId as a parameter but no longer uses it.

Modifications:
- Remove the streamId parameter from HpackDecoder#addHeader

Result:
Less unused parameters in HpackDecoder.


normanmaurer:
Cherry-picked into 4.1 (336cee9b1f431b5b3d5539984a391409ff2d9917). 

@Scottmitch thanks!

PR #7536, opened 2017-12-22
Added exact utf8 length estimator and exposed writeUtf8 with custom space reservation on destination buffer

franz1981:
Motivation:

To avoid eager allocation of the destination and to perform length prefixed encoding of UTF-8 string with forward only access pattern

Modifications:

The original writeUtf8 is modified by allowing customization of the reserved bytes on the destination buffer and is introduced an exact UTF-8 length estimator.

Result:

Is now possible to perform length first encoding with UTF-8 well-formed char sequences following a forward only write access pattern on the destination buffer.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
+++ b/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
@@ -475,14 +475,29 @@ public final class ByteBufUtil {
     /**
      * Encode a {@link CharSequence} in <a href="http://en.wikipedia.org/wiki/UTF-8">UTF-8</a> and write
      * it to a {@link ByteBuf}.
-     *
+     * <p>
+     * It behaves like {@link #reserveAndWriteUtf8(ByteBuf, CharSequence, int)} with {@code reserveBytes}
+     * computed by {@link #utf8MaxBytes(CharSequence)}.<br>
      * This method returns the actual number of bytes written.
      */
     public static int writeUtf8(ByteBuf buf, CharSequence seq) {
+        return reserveAndWriteUtf8(buf, seq, utf8MaxBytes(seq));
+    }
+
+    /**
+     * Encode a {@link CharSequence} in <a href="http://en.wikipedia.org/wiki/UTF-8">UTF-8</a> and write
+     * it into {@code reserveBytes} of a {@link ByteBuf}.
+     * <p>
+     * The {@code reserveBytes} must be computed (ie eagerly using {@link #utf8MaxBytes(CharSequence)}
+     * or exactly with {@link #utf8Bytes(CharSequence)}) to ensure this method to not fail: for performance reasons
+     * the index checks will be performed using just {@code reserveBytes}.<br>
+     * This method returns the actual number of bytes written.
+     */
+    public static int reserveAndWriteUtf8(ByteBuf buf, CharSequence seq, int reserveBytes) {
         for (;;) {
             if (buf instanceof AbstractByteBuf) {
                 AbstractByteBuf byteBuf = (AbstractByteBuf) buf;
-                byteBuf.ensureWritable0(utf8MaxBytes(seq));
+                byteBuf.ensureWritable0(reserveBytes);
                 int written = writeUtf8(byteBuf, byteBuf.writerIndex, seq, seq.length());
                 byteBuf.writerIndex += written;
                 return written;
@@ -521,7 +536,7 @@ public final class ByteBufUtil {
                     // duplicate bounds checking with charAt. If an IndexOutOfBoundsException is thrown we will
                     // re-throw a more informative exception describing the problem.
                     c2 = seq.charAt(++i);
-                } catch (IndexOutOfBoundsException e) {
+                } catch (IndexOutOfBoundsException ignored) {
                     buffer._setByte(writerIndex++, WRITE_UTF_UNKNOWN);
                     break;
                 }
@@ -546,10 +561,76 @@ public final class ByteBufUtil {
     }
 
     /**
+     * Returns max bytes length of UTF8 character sequence of the given length.
+     */
+    public static int utf8MaxBytes(final int seqLength) {
+        return seqLength * MAX_BYTES_PER_CHAR_UTF8;
+    }
+
+    /**
      * Returns max bytes length of UTF8 character sequence.
+     * <p>
+     * It behaves like {@link #utf8MaxBytes(int)} applied to {@code seq} {@link CharSequence#length()}.
      */
     public static int utf8MaxBytes(CharSequence seq) {
-        return seq.length() * MAX_BYTES_PER_CHAR_UTF8;
+        return utf8MaxBytes(seq.length());
+    }
+
+    /**
+     * Returns the exact bytes length of UTF8 character sequence.
+     * <p>
+     * This method is producing the exact length according to {@link #writeUtf8(ByteBuf, CharSequence)}.
+     */
+    public static int utf8Bytes(final CharSequence seq) {
+        if (seq instanceof AsciiString) {
+            return seq.length();
+        }
+        int seqLength = seq.length();
+        int i = 0;
+        // ASCII fast path
+        while (i < seqLength && seq.charAt(i) < 0x80) {
+            ++i;
+        }
+        // !ASCII is packed in a separate method to let the ASCII case be smaller
+        return i < seqLength ? i + utf8Bytes(seq, i, seqLength) : i;
+    }
+
+    private static int utf8Bytes(final CharSequence seq, final int start, final int length) {
+        int encodedLength = 0;
+        for (int i = start; i < length; i++) {
+            final char c = seq.charAt(i);
+            // making it 100% branchless isn't rewarding due to the many bit operations necessary!
+            if (c < 0x800) {
+                // branchless version of: (c <= 127 ? 0:1) + 1
+                encodedLength += ((0x7f - c) >>> 31) + 1;
+            } else if (isSurrogate(c)) {
+                if (!Character.isHighSurrogate(c)) {
+                    encodedLength++;
+                    // WRITE_UTF_UNKNOWN
+                    continue;
+                }
+                final char c2;
+                try {
+                    // Surrogate Pair consumes 2 characters. Optimistically try to get the next character to avoid
+                    // duplicate bounds checking with charAt.
+                    c2 = seq.charAt(++i);
+                } catch (IndexOutOfBoundsException ignored) {
+                    encodedLength++;
+                    // WRITE_UTF_UNKNOWN
+                    break;
+                }
+                if (!Character.isLowSurrogate(c2)) {
+                    // WRITE_UTF_UNKNOWN + (Character.isHighSurrogate(c2) ? WRITE_UTF_UNKNOWN : c2)
+                    encodedLength += 2;
+                    continue;
+                }
+                // See http://www.unicode.org/versions/Unicode7.0.0/ch03.pdf#G2630.
+                encodedLength += 4;
+            } else {
+                encodedLength += 3;
+            }
+        }
+        return encodedLength;
     }
 
     /**
--- a/buffer/src/test/java/io/netty/buffer/ByteBufUtilTest.java
+++ b/buffer/src/test/java/io/netty/buffer/ByteBufUtilTest.java
@@ -263,6 +263,7 @@ public class ByteBufUtilTest {
         ByteBufUtil.writeUtf8(buf2, surrogateString);
 
         assertEquals(buf, buf2);
+        assertEquals(buf.readableBytes(), ByteBufUtil.utf8Bytes(surrogateString));
 
         buf.release();
         buf2.release();
@@ -281,6 +282,7 @@ public class ByteBufUtilTest {
         ByteBufUtil.writeUtf8(buf2, surrogateString);
 
         assertEquals(buf, buf2);
+        assertEquals(buf.readableBytes(), ByteBufUtil.utf8Bytes(surrogateString));
 
         buf.release();
         buf2.release();
@@ -299,6 +301,7 @@ public class ByteBufUtilTest {
         ByteBufUtil.writeUtf8(buf2, surrogateString);
 
         assertEquals(buf, buf2);
+        assertEquals(buf.readableBytes(), ByteBufUtil.utf8Bytes(surrogateString));
 
         buf.release();
         buf2.release();
@@ -318,6 +321,7 @@ public class ByteBufUtilTest {
         ByteBufUtil.writeUtf8(buf2, surrogateString);
 
         assertEquals(buf, buf2);
+        assertEquals(buf.readableBytes(), ByteBufUtil.utf8Bytes(surrogateString));
 
         buf.release();
         buf2.release();
@@ -337,7 +341,7 @@ public class ByteBufUtilTest {
         ByteBufUtil.writeUtf8(buf2, surrogateString);
 
         assertEquals(buf, buf2);
-
+        assertEquals(buf.readableBytes(), ByteBufUtil.utf8Bytes(surrogateString));
         buf.release();
         buf2.release();
     }
@@ -356,6 +360,7 @@ public class ByteBufUtilTest {
         ByteBufUtil.writeUtf8(buf2, surrogateString);
 
         assertEquals(buf, buf2);
+        assertEquals(buf.readableBytes(), ByteBufUtil.utf8Bytes(surrogateString));
 
         buf.release();
         buf2.release();
@@ -372,6 +377,7 @@ public class ByteBufUtilTest {
         ByteBufUtil.writeUtf8(buf2, surrogateString);
 
         assertEquals(buf, buf2);
+        assertEquals(buf.readableBytes(), ByteBufUtil.utf8Bytes(surrogateString));
 
         buf.release();
         buf2.release();
@@ -388,6 +394,7 @@ public class ByteBufUtilTest {
         ByteBufUtil.writeUtf8(buf2, surrogateString);
 
         assertEquals(buf, buf2);
+        assertEquals(buf.readableBytes(), ByteBufUtil.utf8Bytes(surrogateString));
 
         buf.release();
         buf2.release();
@@ -539,6 +546,42 @@ public class ByteBufUtilTest {
             }
         } finally {
             buffer.release();
+        }
+    }
+
+    @Test
+    public void testUtf8Bytes() {
+        final String s = "Some UTF-8 like äÄ∏ŒŒ";
+        checkUtf8Bytes(s);
+    }
+
+    @Test
+    public void testUtf8BytesWithSurrogates() {
+        final String s = "a\uD800\uDC00b";
+        checkUtf8Bytes(s);
+    }
+
+    @Test
+    public void testUtf8BytesWithNonSurrogates3Bytes() {
+        final String s = "a\uE000b";
+        checkUtf8Bytes(s);
+    }
+
+    @Test
+    public void testUtf8BytesWithNonSurrogatesNonAscii() {
+        final char nonAscii = (char) 0x81;
+        final String s = "a" + nonAscii + "b";
+        checkUtf8Bytes(s);
+    }
+
+    private static void checkUtf8Bytes(final CharSequence charSequence) {
+        final ByteBuf buf = Unpooled.buffer(ByteBufUtil.utf8MaxBytes(charSequence));
+        try {
+            final int writtenBytes = ByteBufUtil.writeUtf8(buf, charSequence);
+            final int utf8Bytes = ByteBufUtil.utf8Bytes(charSequence);
+            assertEquals(writtenBytes, utf8Bytes);
+        } finally {
+            buf.release();
         }
     }
 


normanmaurer:
@franz1981 can you rebase ... hopefully the error will be gone then

franz1981:
@normanmaurer Done! I will do the same for the PR on bounds checking I've pushed too!

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +562,65 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
+
+    /**
+     * Returns the exact bytes length of a well-formed UTF8 character sequence.
+     * <p>
+     * This method is producing a precise estimation only with well-formed UTF8 character sequences, otherwise
+     * it will not perform underrated estimations according to {@link #writeUtf8(ByteBuf, CharSequence)}.
+     */
+    public static int utf8Bytes(final CharSequence seq) {
+        int seqLength = seq.length();
+        //the algorithm is just adding the additional length of not ASCII chars to the baseline ASCII length
```
normanmaurer: nit: space after `//`

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +562,65 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
+        int encodedLength = seqLength;
+        int i = 0;
+        //ASCII
```
normanmaurer: nit: space after `//`

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +562,65 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
+        while (i < seqLength && seq.charAt(i) < 0x80) {
+            i++;
+        }
+        if (i < seqLength) {
```
normanmaurer: nit: space after `//`

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +562,65 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
+            //!ASCII
+            encodedLength += utf8BytesSlowPath(seq, i, seqLength);
+            //integer overflow on utf8BytesSlowPath
```
normanmaurer: nit: space after `//`

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +562,65 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
+        int seqLength = seq.length();
+        //the algorithm is just adding the additional length of not ASCII chars to the baseline ASCII length
+        int encodedLength = seqLength;
+        int i = 0;
+        //ASCII
+        while (i < seqLength && seq.charAt(i) < 0x80) {
+            i++;
+        }
+        if (i < seqLength) {
+            //!ASCII
```
normanmaurer: nit: space after `//`

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +562,65 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
+            encodedLength += utf8BytesSlowPath(seq, i, seqLength);
+            //integer overflow on utf8BytesSlowPath
+            if (encodedLength < seqLength) {
+                throw new IllegalArgumentException("UTF-8 encoded length overflow");
```
normanmaurer: Is this something that should ever happen.?

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +562,65 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
+            }
+        }
+        return encodedLength;
+    }
+
+    private static int utf8BytesSlowPath(final CharSequence sequence, final int start, final int length) {
+        int encodedLength = 0;
+        for (int i = start; i < length; i++) {
+            final int c = sequence.charAt(i);
+            //making it 100% brenchless isn't rewarding due to the many bit operations necessary!
```
normanmaurer: nit: space after `//`

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +562,65 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
+            if (c < 0x800) {
+                //branchless version of: c <= 127 ? 0:1
```
normanmaurer: nit: space after `//`

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +562,65 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
+                encodedLength += (0x7f - c) >>> 31;
+            } else {
+                encodedLength += 2;
+                if (MIN_SURROGATE <= c && c <= MAX_SURROGATE) {
+                    assert Character.codePointAt(sequence, i) != c : "Not well formed pair";
+                    i++;
```
normanmaurer: `++I` to make it consistent

normanmaurer:
@franz1981 so the idea is to "trade" cpu vs memory usage ?

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
franz1981: `CharSequence` to me means `StringBuilder/CharBuffer/String` as common cases and if you have such big instances you have for sure bigger troubles than just calculating the UTF 8 encoded length :P, but the API doesn't prohibit anyone to provide a custom implementation backed by an arbitrary long stream of chars that could cause that overflow..
If you think that it could be dropped to me it could be safe enough, but if it doesn't cost I will leave it there, wdyt?

franz1981:
@normanmaurer Yes and hoping that the `CharSequence` to be encoded isn't too long, there are good chances it will be already in the L1 cache ie computing its exact UTF-8 length won't cost too much.
That will make possible to write the encoded length + encoded char sequence on the destination buffer in an append only fashion.

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +562,65 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
      */
     public static int utf8MaxBytes(CharSequence seq) {
-        return seq.length() * MAX_BYTES_PER_CHAR_UTF8;
+        return utf8MaxBytes(seq.length());
+    }
+
+    /**
+     * Returns the exact bytes length of a well-formed UTF8 character sequence.
+     * <p>
+     * This method is producing a precise estimation only with well-formed UTF8 character sequences, otherwise
```
nicktrav: (Excuse the "drive-by" comments): This wording here is a little confusing. "precise estimation" is mentioned here, but the previous sentence mentions "exact", which seems more correct.

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +562,65 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
+            encodedLength += utf8BytesSlowPath(seq, i, seqLength);
+            // integer overflow on utf8BytesSlowPath
+            if (encodedLength < seqLength) {
+                throw new IllegalArgumentException("UTF-8 encoded length overflow");
+            }
+        }
+        return encodedLength;
+    }
+
+    private static int utf8BytesSlowPath(final CharSequence sequence, final int start, final int length) {
```
nicktrav: Without a comment or javadoc, the naming here is a little confusing. The method is not really calculating the number of bytes, rather it's calculating the bytes required in _addition_ to those needed for one-byte ASCII chars. e.g. this method returns `1` when scanning the sequence `aä` (which is three bytes in length).

Maybe `additionalNonAsciiBytes`? ... or just call it out in a comment.

franz1981: Good points!
I've avoided a javadoc mainly due to the visibility and its "one shot" usage of this method, but I will change for sure its name in something more autodocumented for sure :+1:

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
franz1981: > but the previous sentence mentions "exact", which seems more correct.

it is just another way to avoid repeating the same sentence `the exact bytes length`, but I understand that "estimation" could be interpreted as something less precise than "exact"

buffer/src/test/java/io/netty/buffer/ByteBufUtilTest.java
```diff
@@ -542,6 +539,15 @@ public void testIsTextWithInvalidIndexAndLength() {
         }
     }
 
+    @Test
+    public void testUtf8BytesIsExact() {
```
ejona86: There should be a test that includes a surrogate pair.

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +562,65 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
      */
     public static int utf8MaxBytes(CharSequence seq) {
-        return seq.length() * MAX_BYTES_PER_CHAR_UTF8;
+        return utf8MaxBytes(seq.length());
+    }
+
+    /**
+     * Returns the exact bytes length of a well-formed UTF8 character sequence.
+     * <p>
+     * This method is producing the exact length only with well-formed UTF8 character sequences, otherwise
```
ejona86: s/UTF-8/UTF-16/ ?

franz1981: @ejona86 
Thanks!

> I found computing the "additional" bytes to be confusing

what do you mean with confusing? As a design choice or how I've implemented it?
It is mainly due to a simpler handling of integer overflow (rare TBH) and for inlining reasons optimizing around (what I'm expecting to be)  the common case

ejona86: > what do you mean with confusing?

I found it harder to check correctness, since it adds another operation (conceptually), instead of absolute numbers you have to do subtraction to get relative numbers. "And it is 2...? But it is 3 bytes in UTF-8. Doh! But it was one char in UTF-16." Or "isn't there a missing `++`? Doh. It's two chars in UTF-16." In this sort of conversion logic I am quite trained to think it terms of "X is Y bytes in UTF-8".

franz1981: Now I got it :) 
Yes it is less straightforward if compared to an absolute computation and TBH I could have implement it that way too (ie absolute encoded lenght): it is nice to see how different devs consider one way more intuitive than others...and is cool :)
If many devs will consider it less simple to be maintained than an absolute computation i will change it for sure :+1:

normanmaurer:
@franz1981 please ping me once you addressed @ejona86 comments.

buffer/src/test/java/io/netty/buffer/ByteBufUtilTest.java
franz1981: @ejona86 Added blow this one :+1:

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
franz1981: @ejona86 
> s/UTF-8/UTF-16/ ?
I've used UTF8 as it is used in the already existing `ByteBufUtil`'s javadoc :
```
    /**
     * Returns max bytes length of UTF8 character sequence.
     * <p>
     * It behaves like {@link #utf8MaxBytes(int)} applied to {@code seq} {@link CharSequence#length()}.
     */
    public static int utf8MaxBytes(CharSequence seq) {
        return utf8MaxBytes(seq.length());
    }
```

franz1981:
@normanmaurer Done :+1:

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
ejona86: @franz1981, I wasn't talking about that first sentence. I was talking about the line I made the comment on:
> This method is producing the exact length only with well-formed UTF8 character sequences

The input is UTF-16 and is expected to be well-formed. The input isn't well-formed UTF-8.

buffer/src/test/java/io/netty/buffer/ByteBufUtilTest.java
```diff
@@ -542,6 +539,25 @@ public void testIsTextWithInvalidIndexAndLength() {
...
+        checkUtf8Bytes(utf8);
+    }
+
+    @Test
+    public void testUtf8BytesWithSurrogatesIsExact() {
+        final String utf8 = "a\uD800\uDC00b";
+        checkUtf8Bytes(utf8);
+    }
+
+    private static void checkUtf8Bytes(final CharSequence utf8) {
```
ejona86: `utf8` isn't encoded in UTF-8. Please name it something else.

buffer/src/test/java/io/netty/buffer/ByteBufUtilTest.java
```diff
@@ -542,6 +539,25 @@ public void testIsTextWithInvalidIndexAndLength() {
...
+        final ByteBuf buf = Unpooled.buffer(ByteBufUtil.utf8MaxBytes(utf8));
+        final int writtenBytes = ByteBufUtil.writeUtf8(buf, utf8);
```
ejona86: It worries me to see this same class used as the testing oracle. Could this be `utf8.toString().getBytes(CharsetUtil.UTF_8).length` or similar?

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
franz1981: oh!!! You're pretty right, I've missed it!! thanks!

buffer/src/test/java/io/netty/buffer/ByteBufUtilTest.java
franz1981: Makes sense, let me check what I can do!

franz1981: I've looked into it and probably it isn't the case here.
I'm not assuming that `ByteBufUtil::writeUtf8` and `String::getBytes` for UTF-8 charset will produce 1:1 the same output (eg just consider the `?` replacement done on `ByteBufUtil::writeUtf8`), hence I prefer to have these methods coupled as the `checUtf8Bytes` javadoc states.

ejona86: I guess that's fine because the return value will be too small or there will be an index out of range exception. Although there's not any test that triggers WRITE_UTF_UNKNOWN.

normanmaurer:
@Scottmitch PTAL

buffer/src/test/java/io/netty/buffer/ByteBufUtilTest.java
```diff
@@ -542,6 +539,25 @@ public void testIsTextWithInvalidIndexAndLength() {
...
+    public void testUtf8BytesWithSurrogatesIsExact() {
+        final String s = "a\uD800\uDC00b";
+        checkUtf8Bytes(s);
+    }
+
+    private static void checkUtf8Bytes(final CharSequence charSequence) {
+        final ByteBuf buf = Unpooled.buffer(ByteBufUtil.utf8MaxBytes(charSequence));
+        final int writtenBytes = ByteBufUtil.writeUtf8(buf, charSequence);
+        final int utf8Bytes = ByteBufUtil.utf8Bytes(charSequence);
+        Assert.assertEquals(writtenBytes, utf8Bytes);
```
Scottmitch: We must call `buf.release()` after this line.

buffer/src/test/java/io/netty/buffer/ByteBufUtilTest.java
```diff
@@ -542,6 +539,25 @@ public void testIsTextWithInvalidIndexAndLength() {
...
+        final String s = "Some UTF-8 like äÄ∏ŒŒ";
+        checkUtf8Bytes(s);
+    }
+
+    @Test
+    public void testUtf8BytesWithSurrogatesIsExact() {
+        final String s = "a\uD800\uDC00b";
+        checkUtf8Bytes(s);
+    }
+
```
Scottmitch: also add the following test cases:
- includes a non-surrogate pair which results in 3 bytes (e.g. `'\uE000'`)
- includes a non-surrogate pair non-ascii (e.g. `(char) 0x81`)

Scottmitch: please also add update `testWriteUtf8Surrogates` to include the new method:

```java
assertEquals(buf.readableBytes(), ByteBufUtil.utf8Bytes(surrogateString));
```

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +562,65 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
+        for (int i = start; i < length; i++) {
+            final int c = sequence.charAt(i);
+            // making it 100% branchless isn't rewarding due to the many bit operations necessary!
+            if (c < 0x800) {
+                // branchless version of: c <= 127 ? 0:1
+                encodedLength += (0x7f - c) >>> 31;
+            } else {
+                encodedLength += 2;
+                if (MIN_SURROGATE <= c && c <= MAX_SURROGATE) {
+                    assert Character.codePointAt(sequence, i) != c : "Not well formed pair";
```
Scottmitch: our encoding method `writeUtf8` accounts for invalid utf8 encodings just like the JDK. This may result in writing an `WRITE_UTF_UNKNOWN` if the encoding is not correct, but still proceed without an exception. So this method may not support some methods which are supported by `writeUtf8`. We should consider the following:
- change from `assert` to a runtime check. otherwise the encoding size may not be exact, and not detected if asserts for this package are not enabled.
- be more lenient about encoding issues and calculate the size according to `writeUtf8` (and the default JDK configuration).
- add a mode which can conditionally switch between the previous 2 approaches

franz1981:
@Scottmitch I've addressed everything you've rightly pointed me, thanks :+1:

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +560,79 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
+                    continue;
+                }
+                final char c2;
+                try {
+                    // Surrogate Pair consumes 2 characters. Optimistically try to get the next character to avoid
+                    // duplicate bounds checking with charAt.
+                    c2 = seq.charAt(++i);
+                } catch (IndexOutOfBoundsException e) {
+                    encodedLength++;
+                    //WRITE_UTF_UNKNOWN
```
normanmaurer: nit: space after `//`

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +560,79 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
+        for (int i = start; i < length; i++) {
+            final char c = seq.charAt(i);
+            // making it 100% branchless isn't rewarding due to the many bit operations necessary!
+            if (c < 0x800) {
+                // branchless version of: (c <= 127 ? 0:1) + 1
+                encodedLength += ((0x7f - c) >>> 31) + 1;
+            } else if (isSurrogate(c)) {
+                if (!Character.isHighSurrogate(c)) {
+                    encodedLength++;
+                    //WRITE_UTF_UNKNOWN
```
normanmaurer: nit: space after `//`

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +560,79 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
+                    // Surrogate Pair consumes 2 characters. Optimistically try to get the next character to avoid
+                    // duplicate bounds checking with charAt.
+                    c2 = seq.charAt(++i);
+                } catch (IndexOutOfBoundsException e) {
+                    encodedLength++;
+                    //WRITE_UTF_UNKNOWN
+                    break;
+                }
+                if (!Character.isLowSurrogate(c2)) {
+                    //WRITE_UTF_UNKNOWN + (Character.isHighSurrogate(c2) ? WRITE_UTF_UNKNOWN : c2)
```
normanmaurer: nit: space after `//`

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +560,79 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
+                    encodedLength++;
+                    //WRITE_UTF_UNKNOWN
+                    continue;
+                }
+                final char c2;
+                try {
+                    // Surrogate Pair consumes 2 characters. Optimistically try to get the next character to avoid
+                    // duplicate bounds checking with charAt.
+                    c2 = seq.charAt(++i);
+                } catch (IndexOutOfBoundsException e) {
```
normanmaurer: e -> ignored

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +560,79 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
+     * This method is producing the exact length according to {@link #writeUtf8(ByteBuf, CharSequence)}.
+     */
+    public static int utf8Bytes(final CharSequence seq) {
+        int seqLength = seq.length();
+        int encodedLength = 0;
+        int i = 0;
+        // ASCII fast path
+        while (i < seqLength && seq.charAt(i) < 0x80) {
+            ++i;
+            encodedLength++;
```
Scottmitch: instead of incrementing both variables in each loop iteration, can we set `encodedLength = i` outside this loop?

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +560,79 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
+        }
+        if (i < seqLength) {
```
Scottmitch: can we get rid of `encodedLength` and simplify to the following?

```java
return i < seqLength ? i + utf8Bytes(seq, i, seqLength) : i;
```

franz1981: I've missed that chance to simplify it further, yes definitely!

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +560,74 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
+    /**
+     * Returns the exact bytes length of UTF8 character sequence.
+     * <p>
+     * This method is producing the exact length according to {@link #writeUtf8(ByteBuf, CharSequence)}.
+     */
+    public static int utf8Bytes(final CharSequence seq) {
+        int seqLength = seq.length();
+        int i = 0;
+        // ASCII fast path
+        while (i < seqLength && seq.charAt(i) < 0x80) {
```
carl-mastrangelo: optional: make 0x80 a char to avoid promotion.

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +560,74 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
-        return seq.length() * MAX_BYTES_PER_CHAR_UTF8;
+        return utf8MaxBytes(seq.length());
+    }
+
+    /**
+     * Returns the exact bytes length of UTF8 character sequence.
+     * <p>
+     * This method is producing the exact length according to {@link #writeUtf8(ByteBuf, CharSequence)}.
+     */
+    public static int utf8Bytes(final CharSequence seq) {
```
carl-mastrangelo: another fast path would be checking if this is already an AsciiString

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +560,74 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
+        int seqLength = seq.length();
+        int i = 0;
+        // ASCII fast path
+        while (i < seqLength && seq.charAt(i) < 0x80) {
+            ++i;
```
carl-mastrangelo: If you are in the nitty gritty of optimizing this, consider iterating in reverse.  Checking larger indicies before smaller ones allows bounds check elimination, and usually the cache prefetcher does work backwards.  

If not, you can ignore.

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -545,11 +560,74 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
+
+    private static int utf8Bytes(final CharSequence seq, final int start, final int length) {
+        int encodedLength = 0;
+        for (int i = start; i < length; i++) {
+            final char c = seq.charAt(i);
+            // making it 100% branchless isn't rewarding due to the many bit operations necessary!
+            if (c < 0x800) {
+                // branchless version of: (c <= 127 ? 0:1) + 1
+                encodedLength += ((0x7f - c) >>> 31) + 1;
+            } else if (isSurrogate(c)) {
```
carl-mastrangelo: You might double check this branch.  I think java may not have so strict checking on surrogate ordering (like you could have two highs and two lows, or mixed highs and lows in odd orders).

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
franz1981: Re bounds checking I've just verified that with ascii cases it already works as expected, otherwise it won't make any difference due to `InlineSmallCode` that prevent the uncommon path from being inlined and forcing the JVM to put bounds check before the uncommon loop too.

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
franz1981: I'm not sure if you mean a promotion on Java or in the compiled ASM, so I've chosen the final source of truth, ASM :P
That's the `seq.charAt(i) < 0x80` part inside the unrolled loop:
```
0x00007f2ca0e5353b: nopl 0x0(%rax,%rax,1)  ;*aload_0
                                           ; - CompileTest::check@16 (line 26)
             L0001: movzwl 0x10(%r11,%rax,2),%r10d  ;*caload
                                                    ; - java.lang.String::charAt@27 (line 660)
                                                    ; - CompileTest::check@18 (line 26)
0x00007f2ca0e53546: cmp $0x80,%r10d
0x00007f2ca0e5354d: jge L0008  ;*if_icmpge
                               ; - CompileTest::check@26 (line 26)
0x00007f2ca0e5354f: movslq %eax,%r8
0x00007f2ca0e53552: movzwl 0x12(%r11,%r8,2),%r10d  ;*caload
                                                   ; - java.lang.String::charAt@27 (line 660)
                                                   ; - CompileTest::check@18 (line 26)
0x00007f2ca0e53558: cmp $0x80,%r10d
0x00007f2ca0e5355f: jge L0006  ;*if_icmpge
                               ; - CompileTest::check@26 (line 26)
0x00007f2ca0e53561: movzwl 0x14(%r11,%r8,2),%r10d  ;*caload
                                                   ; - java.lang.String::charAt@27 (line 660)
                                                   ; - CompileTest::check@18 (line 26)
0x00007f2ca0e53567: cmp $0x80,%r10d
0x00007f2ca0e5356e: jge L0007  ;*if_icmpge
                               ; - CompileTest::check@26 (line 26)
0x00007f2ca0e53570: movzwl 0x16(%r11,%r8,2),%r10d  ;*caload
                                                   ; - java.lang.String::charAt@27 (line 660)
                                                   ; - CompileTest::check@18 (line 26)
0x00007f2ca0e53576: cmp $0x80,%r10d
```
The only promotion I could spot here is `movzwl` that is zero extending to long the char/int returned by `seq.charAt`, but I can ensure you (because the ASM is the same) that putting a char cast or declaring `0x80` a char doesn't make difference: the ASM is 1:1 like this one.

normanmaurer:
@carl-mastrangelo @Scottmitch PTAL again and let me know once you are satisfied.

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
Scottmitch: the branching strategy here is designed to be consistent with `writeUtf8` above which we have unit tests  to ensure the same behavior with the default JDK encoder. Do you have a specific test case which maybe problematic?

franz1981: @carl-mastrangelo I can say that @Scottmitch took the words right out of my mouth :+1:

normanmaurer:
Ok I cherry-picked this into 4.1 (bc8e022601b205e6c38075999fe6ebd08d81f179).

@carl-mastrangelo if you have any followup comments please open a PR.

PR #7539, opened 2017-12-22
Implemented ASCII fast path on writeUtf8 to speedup encoding of ASCII prefixed char sequences

franz1981:
Motivation:

UTF-8 char sequences can be mostly or totally ASCII prefixed hence handling this case better would benefit the common case.

Modifications:

The original writeUtf8 is modified handling the ASCII fast path at first while falling back to the full-fat UTF-8 when the first not ASCII char is met.

Result:

Benchmarks shows an improvement for mostly ASCII char sequences with a negligible impact on full UTF-8 encoding.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
+++ b/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
@@ -503,8 +503,24 @@ public final class ByteBufUtil {
 
         // We can use the _set methods as these not need to do any index checks and reference checks.
         // This is possible as we called ensureWritable(...) before.
-        for (int i = 0; i < len; i++) {
-            char c = seq.charAt(i);
+        int i;
+        //ASCII fast-path
+        for (i = 0; i < len; i++) {
+            final char c = seq.charAt(i);
+            if (c < 0x80) {
+                buffer._setByte(writerIndex++, (byte) c);
+            } else {
+                writerIndex = writeUtf8SlowPath(buffer, writerIndex, seq, len, i);
+                break;
+            }
+        }
+        return writerIndex - oldWriterIndex;
+    }
+
+    private static int writeUtf8SlowPath(final AbstractByteBuf buffer, int writerIndex,
+                                         final CharSequence seq, final int len, int i) {
+        for (; i < len; i++) {
+            final char c = seq.charAt(i);
             if (c < 0x80) {
                 buffer._setByte(writerIndex++, (byte) c);
             } else if (c < 0x800) {
@@ -542,7 +558,7 @@ public final class ByteBufUtil {
                 buffer._setByte(writerIndex++, (byte) (0x80 | (c & 0x3f)));
             }
         }
-        return writerIndex - oldWriterIndex;
+        return writerIndex;
     }
 
     /**
--- a/microbench/src/main/java/io/netty/microbench/buffer/ByteBufUtilBenchmark.java
+++ b/microbench/src/main/java/io/netty/microbench/buffer/ByteBufUtilBenchmark.java
@@ -22,11 +22,14 @@ import io.netty.microbench.util.AbstractMicrobenchmark;
 import io.netty.util.CharsetUtil;
 import org.openjdk.jmh.annotations.Benchmark;
 import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Param;
 import org.openjdk.jmh.annotations.Scope;
 import org.openjdk.jmh.annotations.Setup;
 import org.openjdk.jmh.annotations.State;
 import org.openjdk.jmh.annotations.TearDown;
 import org.openjdk.jmh.annotations.Warmup;
+
+import java.util.Arrays;
 
 
 @State(Scope.Benchmark)
@@ -44,22 +47,27 @@ public class
 
     private StringBuilder utf8Sequence;
     private String utf8;
+    private String fullUtf8;
+
+    @Param({"8", "128", "1024"})
+    private int stringLength;
 
     @Setup
     public void setup() {
+        final int requiredBytes = 4 * stringLength;
         // Use buffer sizes that will also allow to write UTF-8 without grow the buffer
-        buffer = Unpooled.directBuffer(512);
-        wrapped = Unpooled.unreleasableBuffer(Unpooled.directBuffer(512));
-        asciiSequence = new StringBuilder(128);
-        for (int i = 0; i < 128; i++) {
+        buffer = Unpooled.directBuffer(requiredBytes);
+        wrapped = Unpooled.unreleasableBuffer(Unpooled.directBuffer(requiredBytes));
+        asciiSequence = new StringBuilder(stringLength);
+        for (int i = 0; i < stringLength; i++) {
             asciiSequence.append('a');
         }
         ascii = asciiSequence.toString();
 
         // Generate some mixed UTF-8 String for benchmark
-        utf8Sequence = new StringBuilder(128);
+        utf8Sequence = new StringBuilder(stringLength);
         char[] chars = "Some UTF-8 like äÄ∏ŒŒ".toCharArray();
-        for (int i = 0; i < 128; i++) {
+        for (int i = 0; i < stringLength; i++) {
             utf8Sequence.append(chars[i % chars.length]);
         }
         utf8 = utf8Sequence.toString();
@@ -67,6 +75,10 @@ public class
 
         asciiBuffer = Unpooled.copiedBuffer(ascii, CharsetUtil.US_ASCII);
         utf8Buffer = Unpooled.copiedBuffer(utf8, CharsetUtil.UTF_8);
+        final char utf8Char = 'Œ';
+        final char[] utf8Chars = new char[stringLength];
+        Arrays.fill(utf8Chars, utf8Char);
+        fullUtf8 = new String(utf8Chars);
     }
 
     @TearDown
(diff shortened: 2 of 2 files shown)


franz1981:
The benchmark results show interesting results
on 4.1:
```
Benchmark                               Mode  Cnt         Score        Error  Units
ByteBufUtilBenchmark.writeAscii        thrpt   20  13933422.759 ± 126652.679  ops/s
ByteBufUtilBenchmark.writeAsciiAsUtf8  thrpt   20  12456281.779 ± 279760.847  ops/s
ByteBufUtilBenchmark.writeFullUtf8     thrpt   20  34670225.086 ± 384342.511  ops/s
ByteBufUtilBenchmark.writeUtf8         thrpt   20   4614812.430 ±  60556.670  ops/s
```
With this PR:
```

Benchmark                               Mode  Cnt         Score        Error  Units
ByteBufUtilBenchmark.writeAscii        thrpt   20  13899817.109 ± 108070.691  ops/s
ByteBufUtilBenchmark.writeAsciiAsUtf8  thrpt   20  13828353.123 ± 131895.610  ops/s
ByteBufUtilBenchmark.writeFullUtf8     thrpt   20  36728458.579 ± 304574.246  ops/s
ByteBufUtilBenchmark.writeUtf8         thrpt   20   4709120.778 ±  36624.084  ops/s
```
Now the fast path of UTF-8 encoding on ASCII encoded char sequences has the same performance of the ASCII one with negligible impacts on the original performances (the benchmark show some improvement, but is in the error marging TBH).

microbench/src/main/java/io/netty/microbench/buffer/ByteBufUtilBenchmark.java
```diff
@@ -164,13 +166,25 @@ public void writeUtf8ViaArrayWrapped() {
     @Benchmark
     public void writeUtf8() {
         buffer.resetWriterIndex();
-        ByteBufUtil.writeUtf8(buffer, utf8Sequence);
+        ByteBufUtil.writeUtf8(buffer, utf8);
+    }
+
+    @Benchmark
+    public void writeAsciiAsUtf8() {
```
doom369: Those benchmarks are wrong. More info why - http://hg.openjdk.java.net/code-tools/jmh/file/fbe1b55eadf8/jmh-samples/src/main/java/org/openjdk/jmh/samples/JMHSample_08_DeadCode.java

franz1981: I agree on the paper and I was thinking to push another PR following the JMH best practices.
TBH what is happening in most of the already provided ones is different, thanks to `@TearDown` that for the JVM to avoid the benchmarked methods to be removed: the JVM cannot declare them as *dead code* because it can't say in advance what is happening on the tear down method.
This is confirmed by looking at the assembly produced, that shows on C2 optimization level that the code is not removed.

franz1981: Re http://hg.openjdk.java.net/code-tools/jmh/file/fbe1b55eadf8/jmh-samples/src/main/java/org/openjdk/jmh/samples/JMHSample_08_DeadCode.java it is a different case because the methods are side effect free from the calling method point of view

franz1981: Anyway considering that the results doesn't show big margins as I was expecting (in addition, I've a branch with changed JMH benchmarks too) I can close this PR :+1:

johnou: @franz1981 are you sure it's within the error margin and not a 1000000 ops/s increase?

```code
ByteBufUtilBenchmark.writeAsciiAsUtf8  thrpt   20  12456281.779 ± 279760.847  ops/s
                                                     279760
ByteBufUtilBenchmark.writeAsciiAsUtf8  thrpt   20  13828353.123 ± 131895.610  ops/s
                                                     131895
```

franz1981: @johnou You're right, probably marking it as "within the error margin" is too harsh :P
My concern is that it is not a *huge* improvement as I was expecting, just ~10%...

johnou: For us that's huge, ByteBufUtil.writeUtf8 is a hot path. Could I anyway request that you change the jmh to return the object to avoid copy / paste and running into that dead code trap?

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -503,7 +503,23 @@ static int writeUtf8(AbstractByteBuf buffer, int writerIndex, CharSequence seq,
...
+                break;
+            }
+        }
+        return writerIndex - oldWriterIndex;
+    }
+
+    private static int writeUtf8SlowPath(final AbstractByteBuf buffer, int writerIndex,
+                                         final CharSequence seq, final int len, int i) {
+        for (; i < len; i++) {
             char c = seq.charAt(i);
```
johnou: Mark c final for consistency?

microbench/src/main/java/io/netty/microbench/buffer/ByteBufUtilBenchmark.java
johnou: Hard to see from mobile but does the improvement come from the fact that writeUtf8 can now be inlined?

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
franz1981: Uhh I've missed it while copying the original body: good point!

microbench/src/main/java/io/netty/microbench/buffer/ByteBufUtilBenchmark.java
franz1981: Mmm difficult to say...The original bytecode size was small enough that it was already inlined: if I've to put my money I will bet it is due to a better usage of the micro ops cache of the CPU...
I've to run th Perf profiler to be sure of that, hoping that my consumer CPU has that kind of counters available!

johnou:
I wasn't able to reproduce the 10% performance increase.

```
Previous

Benchmark                               Mode  Cnt         Score        Error  Units
ByteBufUtilBenchmark.writeAsciiAsUtf8  thrpt  100  13228751.016 ± 156616.125  ops/s

Previous with final keyword on c variable

Benchmark                               Mode  Cnt         Score        Error  Units
ByteBufUtilBenchmark.writeAsciiAsUtf8  thrpt  100  13401118.480 ± 194586.268  ops/s
ByteBufUtilBenchmark.writeFullUtf8     thrpt  100  35272695.322 ± 274635.697  ops/s
ByteBufUtilBenchmark.writeUtf8         thrpt  100   4452171.899 ±  29517.474  ops/s

Optimised (this PR with modified benchmark to return int / ByteBuf)

Benchmark                               Mode  Cnt         Score        Error  Units
ByteBufUtilBenchmark.writeAsciiAsUtf8  thrpt  100  11426143.723 ± 236834.007  ops/s
ByteBufUtilBenchmark.writeFullUtf8     thrpt  100  37893558.826 ± 830499.204  ops/s
ByteBufUtilBenchmark.writeUtf8         thrpt  100   4364675.956 ±  96279.510  ops/s
```

franz1981:
@johnou I will retry with more forks too! Did you have the writeAsciiString results too of the same benchmarks runs? That's the real baseline for any improvements IMO
It would be great to have this same benchmark using 3 different lengths (short/medium|UUID-like|long) too

franz1981:
@johnou I've modified the benchmark with variable sized strings: these are the results with the new PR:
```
Benchmark                                    (stringLength)   Mode  Cnt         Score         Error  Units
ByteBufUtilBenchmark.writeAsciiString                     8  thrpt   20  72747959.436 ± 3591213.261  ops/s
ByteBufUtilBenchmark.writeAsciiString                   128  thrpt   20  12181677.898 ±   34360.577  ops/s
ByteBufUtilBenchmark.writeAsciiString                  1024  thrpt   20   1691206.827 ±   15469.176  ops/s
ByteBufUtilBenchmark.writeAsciiStringAsUtf8               8  thrpt   20  77754528.549 ±  763594.754  ops/s
ByteBufUtilBenchmark.writeAsciiStringAsUtf8             128  thrpt   20  11223187.865 ±  119018.589  ops/s
ByteBufUtilBenchmark.writeAsciiStringAsUtf8            1024  thrpt   20   1455686.602 ±   17138.850  ops/s
ByteBufUtilBenchmark.writeFullUtf8String                  8  thrpt   20  47160853.218 ±  553114.475  ops/s
ByteBufUtilBenchmark.writeFullUtf8String                128  thrpt   20   5311433.950 ±   73524.882  ops/s
ByteBufUtilBenchmark.writeFullUtf8String               1024  thrpt   20    643214.613 ±   29458.033  ops/s
ByteBufUtilBenchmark.writeUtf8String                      8  thrpt   20  78539446.469 ±  846538.650  ops/s
ByteBufUtilBenchmark.writeUtf8String                    128  thrpt   20   4804244.447 ±  162650.026  ops/s
ByteBufUtilBenchmark.writeUtf8String                   1024  thrpt   20    677367.937 ±   15832.357  ops/s
```

franz1981:
@johnou While these ones are with the original `writeUtf8`:
```
Benchmark                                    (stringLength)   Mode  Cnt         Score         Error  Units
ByteBufUtilBenchmark.writeAsciiString                     8  thrpt   20  74749874.645 ± 2671759.272  ops/s
ByteBufUtilBenchmark.writeAsciiString                   128  thrpt   20  11459849.272 ±  419816.311  ops/s
ByteBufUtilBenchmark.writeAsciiString                  1024  thrpt   20   1679524.664 ±   38366.004  ops/s
ByteBufUtilBenchmark.writeAsciiStringAsUtf8               8  thrpt   20  72946077.597 ±  480300.273  ops/s
ByteBufUtilBenchmark.writeAsciiStringAsUtf8             128  thrpt   20  13295598.495 ±  124586.538  ops/s
ByteBufUtilBenchmark.writeAsciiStringAsUtf8            1024  thrpt   20   1809970.123 ±   22364.444  ops/s
ByteBufUtilBenchmark.writeFullUtf8String                  8  thrpt   20  50472646.901 ±  359376.878  ops/s
ByteBufUtilBenchmark.writeFullUtf8String                128  thrpt   20   5427233.121 ±   26526.761  ops/s
ByteBufUtilBenchmark.writeFullUtf8String               1024  thrpt   20    701231.707 ±    5607.669  ops/s
ByteBufUtilBenchmark.writeUtf8String                      8  thrpt   20  72335423.290 ±  729530.583  ops/s
ByteBufUtilBenchmark.writeUtf8String                    128  thrpt   20   4599448.039 ±   95454.556  ops/s
ByteBufUtilBenchmark.writeUtf8String                   1024  thrpt   20    683134.535 ±    8943.448  ops/s
```

franz1981:
@johnou In conclusion: as you have noticed it doesn't seems a clear step forward compared to the original implementation and although it shows some improvements there are regressions as well: what I'm not 100% is why there is such great variance in the results while using the same implementation too, wdyt?
Maybe it could worth to try with more forks?

johnou:
Not sure if it's worth the trouble, if the method can be inlined and there is no excess object allocation or bounds checking I would leave it as is.

franz1981:
@johnou Agree and I'm pretty surprised too: a similar optimization is contained into the StandardCharset encoders of the OpenJDK too and I'm just guessing why they have use it and which scenarios it was meant to favour...
Re this PR i can close it but probably will push a new ones with the changes in the JMH benchmark, makes sense?

normanmaurer:
@franz1981 yeah I think we should close this one and just fix the benchmarks :)

Issue #7537, opened 2017-12-22
 DefaultChannelPromise.setSuccess(null) throw NPE  when I update netty-version from 4.0.36.Final to 4.1.16.Final


Lihuanghe:
### Actual behavior
 `futurn.setSuccess(null) throw NPE  when I update netty-version from 4.0.36.Final to 4.1.16.Final`
### Steps to reproduce
EveryTime when I call method doBusiness 
### Minimal yet complete reproducer code (or URL to code)
source like this:

```java
public ChannelPromise doBusiness(){
  // many other code
  //...
  //...
  
  DefaultChannelPromise futurn = new DefaultChannelPromise(null);
  futurn.setSuccess(null);  // netty-4.1.16-Final throw NPE
  return futurn;

}
```

### Netty version
4.1.16.Final
### JVM version (e.g. `java -version`)
sun-hotspot-jdk1.7
### OS version (e.g. `uname -a`)
centos6.5

normanmaurer:
@Lihuanghe this is expected and it was in fact a bug that it did not before. That said I think we should throw directly when the constructor is invoked to fail fast. 

If you want to create a `ChannelPromise` you should use `channel.newPromise()` or if you want a `Promise` use `executor.newPromise()`

Lihuanghe:
thank you .

PR #7540, opened 2017-12-22
Fail fast when DefaultChannelPromise is constructed with null as Chan…

normanmaurer:
…nel.

Motivation:

We should fail fast when DefaultChannelPromise is constructed with null as Channel as otherwise it will fail with a NPE once we call setSuccess / setFailure.

Modifications:

Add null check and test.

Result:

Fail fast.

### DIFF
--- a/transport/src/main/java/io/netty/channel/DefaultChannelPromise.java
+++ b/transport/src/main/java/io/netty/channel/DefaultChannelPromise.java
@@ -21,6 +21,8 @@ import io.netty.util.concurrent.EventExecutor;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.GenericFutureListener;
 
+import static io.netty.util.internal.ObjectUtil.checkNotNull;
+
 /**
  * The default {@link ChannelPromise} implementation.  It is recommended to use {@link Channel#newPromise()} to create
  * a new {@link ChannelPromise} rather than calling the constructor explicitly.
@@ -37,7 +39,7 @@ public class DefaultChannelPromise extends DefaultPromise<Void> implements Chann
      *        the {@link Channel} associated with this future
      */
     public DefaultChannelPromise(Channel channel) {
-        this.channel = channel;
+        this.channel = checkNotNull(channel, "channel");
     }
 
     /**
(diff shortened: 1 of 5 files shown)


normanmaurer:
Inspired by https://github.com/netty/netty/issues/7537

transport/src/main/java/io/netty/channel/DefaultChannelPromise.java
```diff
@@ -37,7 +39,7 @@
      *        the {@link Channel} associated with this future
      */
     public DefaultChannelPromise(Channel channel) {
-        this.channel = channel;
+        this.channel = checkNotNull(channel, "channel");
     }
 
     /**
```
Scottmitch: does the constructor below also need to have a null check?

normanmaurer: @Scottmitch I thought about this but we pass in null explicit in `Bootstrap` if `ChannelFactory` throws. We could use a "dummy" `Channel` here as replacement of null but not sure this is better. Thoughts ?

Scottmitch: Seems strange to have once constructor allow null and the other not allow null in this scenario. The case of "no channel" does seem exceptional and a "dummy" channel seems reasonable here to make the interface contract more clear.

normanmaurer:
Works for me... will take care once I am done with the stuff I am doing atm 🤪

transport/src/main/java/io/netty/channel/DefaultChannelPromise.java
normanmaurer: done... PTAL again

normanmaurer:
@Scottmitch PTAL again

normanmaurer:
Cherry-picked into 4.1 (1740f366eb728ea5a0a63d18e9042161673414cd) and 4.0 (bd6435f5532e5878ba0a9379a47bccff71e5fbbe)

transport/src/main/java/io/netty/channel/DefaultChannelPromise.java
```diff
@@ -37,7 +39,7 @@
      *        the {@link Channel} associated with this future
      */
     public DefaultChannelPromise(Channel channel) {
-        this.channel = channel;
+        this.channel = checkNotNull(channel, "channel");
```
carl-mastrangelo: Followup: this caused some tests in grpc to start failing I think.  We are probably at fault here, but just to let you know some code was doing this.

normanmaurer: @carl-mastrangelo thanks... that said I think it is a bug to pass in null :)

PR #7542, opened 2017-12-22
Http2FrameCodecTest increase timeout

Scottmitch:
Motivation:
Http2FrameCodecTest#newOutboundStream has a timeout of 1 second and has been observed to timeout on CI servers.

Modifications:
- Increase the timeout to 5 seconds

Result:
Less false positive test failures on CI servers.


normanmaurer:
Cherry-picked into 4.1 (adf2596c36946fc1ec1c363375e927dd72b51c1f)

PR #7543, opened 2017-12-22
Test output should include GC details

Scottmitch:
Motivation:
Our tests are often asynchronous and have timeouts to avoid hanging indefinitely. However sometimes the timeouts maybe set to low for the CI servers. It would be helpful to confirm if the application was busy with GC and if that was a contributing factor to the test timing out.

Modifications:
- Unit tests should run with -XX:+PrintGCDetails by default

Result:
More visibility into GC behavior in unit tests.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -95,6 +95,12 @@
       </properties>
     </profile>
     <profile>
+      <id>noPrintGC</id>
+      <properties>
+        <argLine.printGC>-D_</argLine.printGC>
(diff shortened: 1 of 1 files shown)


normanmaurer:
Cherry-picked into 4.1 (e17f438b104f7488fc18ff6c180f1eb5f58744a0) and 4.0 (db787ff5326500764487f0521a19257c4a0ae75d)

PR #7544, opened 2017-12-23
OpenSslEngine: Remove renegotiation support

### LINKED ISSUE #6320: OpenSslEngine should not call SSL_set_state
OpenSslEngine currently calls `SSL_set_state(SSL_ST_ACCEPT)` during the renegotiation process. The OpenSSL team has advised against this [1]. We should investigate and avoid calling `SSL_ST_ACCEPT` if possible.

[1] https://github.com/openssl/openssl/issues/2551#issuecomment-277644153

FauxFaux:
This is the minimum number of changes necessary to remove the calls to `SSL.setState()`, which are causing problems for `netty-tcnative` building on modern linux distributions.

~~I haven't signed the contributor license agreement~~. I have signed the individual contributor license agreement.

Please read the associated ticket (#6320) for some discussion.

---

### Motivation:

`SSL.setState()` has gone from openssl 1.1. Calling it is, and probably always has been, incorrect. Doing renogitation in this manner is potentially insecure. There have been at least two insecure renegotiation vulnerabilities in users of the OpenSSL library.

Renegotiation is not necessary for correct operation of the TLS protocol.

BoringSSL has already eliminated this functionality, and the tests (now deleted) were not running on BoringSSL.

### Modifications:

If the connection setup has completed, always return that negotiation is not supported. Previously this was done only if we were the client.

Remove the tests for this functionality.

### Result:

Fixes #6320.


johnou:
Only nit I have is that renegotiation is not spelt correctly in the commit message.

FauxFaux:
Oh wow, that's gone really badly, hasn't it. I'll push a new version.

johnou:
@FauxFaux are you also handling a pull request for removing ```SSL.setState``` in https://github.com/netty/netty-tcnative?

johnou:
cc @rkapsi @snazy

FauxFaux:
I've signed the "icla", and raised a pull request for netty-tcnative: https://github.com/netty/netty-tcnative/pull/325

Whoever does the releases needs to be aware that these need to be released in order.

normanmaurer:
@FauxFaux first of thanks for this... In the past we discussed this and by the time decided that we did not want to remove renegotiation as this may break apps that depend on it. I know in practise you should not really use it but there may be people that depend on it. 

That said I am not sure if we changed our mind and are fine to just drop it.  I am personally fine with it... @Scottmitch @carl-mastrangelo @ejona86 @trustin WDYT ?

ruckc:
In HTTPD, I believe it uses renegotiation to require SSL client  certificate authentication only some connections (based on which path is requested).  Would this remove that ability?

FauxFaux:
@normanmaurer: I'm not a direct/heavy user of netty; so that decision is up to you folks. My two cents is that, in general security, being secure is more important than backwards compatibility; if people are using this thinking it provides security and it turns out that it doesn't; they are being severly mislead.

I'm not enough of an openssl developer to convince you that the current code is *wrong*, but I've done enough security engineering/cryptography to label it as *certainly scary*. I'm also not enough of an openssl/netty developer to attempt to fix the code to not call `setState`. There were some thoughts on the original issue, which I'll relink here:
 * https://github.com/netty/netty/issues/6320#issuecomment-292554546
 * https://github.com/openssl/openssl/issues/2551#issuecomment-277417205

---

@ruckc: I don't know enough about client cert auth in TLS to know the answer to that.
Is it supported on boringssl, which has had this code disabled forever?
Do you have sample netty code that could be integrated into the test suite? It's either not covered by the existing test suite, or not broken.

This change is necessary to support netty *at all* going forward. Conditional client cert auth is more of a minority feature. If there's no current software which is broken by this change, I would still advise making it.

Lukasa:
> In HTTPD, I believe it uses renegotiation to require SSL client certificate authentication only some connections (based on which path is requested). Would this remove that ability?

I doubt the Netty code supports handling this as a client: to do so require pretty deep integration into the info callbacks in OpenSSL to spot the server’s certificate request. However, this shouldn’t break that: OpenSSL will transparently handle processing a renegotiation initiated by the remote peer.

What this will break is servers attempting to behave like HTTPD by attempting to delay client auth until later in the connection lifecycle. My 2c is that I think that use case probably shouldn’t be supported anyway, because TLS renegotiation is a terrible misfeature, but that doesn’t mean people aren’t using it.

rkapsi:
No objections on our end and being able to use openssl 1.1 is :+1: (very interested in being able to use TLS 1.3).

doom369:
+1

normanmaurer:
@mosesn any concerns ?

mosesn:
@normanmaurer no, the feature I was thinking of was resumption, I just forgot what it was called.  🚢  🇮🇹

Scottmitch:
The additional complexity/hacks required to support it (in OpenSSL and Netty), the potential for mishaps during the renegotiation protocol implications, and it has been removed from TLS 1.3 make for a compelling case. If there are no strong objections my preference is to remove renegotiation support.

Scottmitch:
4.0 (8ffa828cbb)

Scottmitch:
Followup PR for more removal: https://github.com/netty/netty/pull/7558

Issue #7546, opened 2017-12-28
Using Netty on Android, throw out abnormal javax.net.ssl.SSLException: failed to initialize the server-side SSL context


litangyu:
### Expected behavior
I want to use the Netty in Android, and create a capture software.

Can you change the code into a BKS KeyStore library that is compatible with Android.

I find the method of name buildKeyStore in SslContext, which uses KeyStore.getInstance ("JKS"). I think if you use KeyStore.getInstance (KeyStore.getDefaultType ()), it will be better.

My English is not very good. It may not express what I mean. But I'm looking forward to getting back to me, or giving me some advice.

### Actual behavior

~~~
12-28 19:01:36.910 23808-24612/me.lty.myapplication W/System.err: javax.net.ssl.SSLException: failed to initialize the server-side SSL context
12-28 19:01:36.910 23808-24612/me.lty.myapplication W/System.err:     at io.netty.handler.ssl.JdkSslServerContext.newSSLContext(JdkSslServerContext.java:278)
12-28 19:01:36.910 23808-24612/me.lty.myapplication W/System.err:     at io.netty.handler.ssl.JdkSslServerContext.<init>(JdkSslServerContext.java:237)
12-28 19:01:36.910 23808-24612/me.lty.myapplication W/System.err:     at io.netty.handler.ssl.SslContext.newServerContextInternal(SslContext.java:415)
12-28 19:01:36.910 23808-24612/me.lty.myapplication W/System.err:     at io.netty.handler.ssl.SslContextBuilder.build(SslContextBuilder.java:441)
12-28 19:01:36.910 23808-24612/me.lty.myapplication W/System.err:     at me.lty.myapplication.proxy.handler.HttpProxyServerHandle.channelRead(HttpProxyServerHandle.java:140)
12-28 19:01:36.910 23808-24612/me.lty.myapplication W/System.err:     at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
12-28 19:01:36.910 23808-24612/me.lty.myapplication W/System.err:     at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
12-28 19:01:36.910 23808-24612/me.lty.myapplication W/System.err:     at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
[...]

PR #7547, opened 2017-12-28
Remove direct usage of JKS and SunX509

normanmaurer:
Motivation:

When using netty on android or with for example a IBM JVM it may not be able to build a SslContext as we hardcoded the use of JKS and SunX509 (which both may not be present).

Modifications:

- Use the default algorithm / type which can be override via a System property
- Remove System property check as its redundant with KeyManagerFactory.getDefaultAlgorithm()

Result:

More portable code. Fixes [#7546].


Scottmitch:
4.0 (d34a930e4b)

carl-mastrangelo:
Latent LGTM.

PR #7558, opened 2018-01-02
Remove remote initiated renegotiation support

Scottmitch:
Motivation:
We recently removed support for renegotiation, but there are still some hooks to attempt to allow remote initiated renegotiation to succeed. The remote initated renegotiation can be even more problematic from a security stand point and should also be removed.

Modifications:
- Remove state related to remote iniated renegotiation from OpenSslEngine

Result:
More renegotiation code removed from the OpenSslEngine code path.


Scottmitch:
@normanmaurer - note that I didn't touch the `SslHandler#renegotate` method because it doesn't rely upon any special OpenSSL API calls. However we can be more aggressive and disable this too (don't think its strictly necessary though).

normanmaurer:
Cherry-picked into 4.1 (062d9269120bac0064408025b78788613d41d36d) and 4.0 (ccecc20124abf83540d5d0258c0fba988eefb798)

PR #7559, opened 2018-01-02
Update domains in DnsNameResolverTest

Scottmitch:
Motivation:
DnsNameResolverTest has not been updated in a while.

Modifications:
- Update the DOMAINS definition in DnsNameResolverTest

Result:
More current domain names.


Scottmitch:
Note that a build failure was observed because `blogspot.in` failed to be resolved during PR build https://github.com/netty/netty/pull/7402 [1]. This domain name is no longer in the top 100 from the automated domain name script we use.

[1] https://garage.netty.io/teamcity/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=27280#_focus=82870

PR #7562, opened 2018-01-03
Fix StompFrame.copy() does not copy headers, close #7561

### LINKED ISSUE #7561: StompFrame.copy() does not copy headers
### Expected behavior
Call stompFrame.copy() do copies of content and headers.

### Actual behavior
After call stompFrame.copy() only content was copy.

 `StompFrame stompFrame = new DefaultStompFrame(StompCommand.CONNECT);`
 `stompFrame.headers().set(StompHeaders.HOST, "localhost");`
 `StompFrame copyFrame = stompFrame.copy();`
 `Assert.assertTrue(copyFrame.headers().isEmpty());`

### Netty version
4.1.20.Final

java version "1.8.0_151"
Java(TM) SE Runtime Environment (build 1.8.0_151-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.151-b12, mixed mode)

Darwin Kernel Version 16.7.0: Mon Nov 13 21:56:25 PST 2017; root:xnu-3789.72.11~1/RELEASE_X86_64 x86_64

amizurov:
Motivation:
  Incorrect behavior for StompFrame.copy() method.

Modification:
 Added copying of frame headers

Result:
 When you call the StompFrame.copy() method, the headers are also copied.

Fixes [#7561].

### DIFF
--- a/codec-stomp/src/main/java/io/netty/handler/codec/stomp/DefaultStompFrame.java
+++ b/codec-stomp/src/main/java/io/netty/handler/codec/stomp/DefaultStompFrame.java
@@ -31,10 +31,16 @@ public class DefaultStompFrame extends DefaultStompHeadersSubframe implements St
     }
 
     public DefaultStompFrame(StompCommand command, ByteBuf content) {
-        super(command);
+        this(command, content, null);
+    }
+
+    public DefaultStompFrame(StompCommand command, ByteBuf content, StompHeaders headers) {
+        super(command, headers);
+
         if (content == null) {
             throw new NullPointerException("content");
         }
+
         this.content = content;
     }
 
@@ -60,7 +66,7 @@ public class DefaultStompFrame extends DefaultStompHeadersSubframe implements St
 
     @Override
     public StompFrame replace(ByteBuf content) {
-        return new DefaultStompFrame(command, content);
+        return new DefaultStompFrame(command, content, new DefaultStompHeaders().set(headers));
     }
 
     @Override
--- a/codec-stomp/src/main/java/io/netty/handler/codec/stomp/DefaultStompFrame.java
+++ b/codec-stomp/src/main/java/io/netty/handler/codec/stomp/DefaultStompFrame.java
@@ -34,7 +34,7 @@ public class DefaultStompFrame extends DefaultStompHeadersSubframe implements St
         this(command, content, null);
     }
 
-    public DefaultStompFrame(StompCommand command, ByteBuf content, StompHeaders headers) {
+    DefaultStompFrame(StompCommand command, ByteBuf content, DefaultStompHeaders headers) {
         super(command, headers);
 
         if (content == null) {
@@ -66,7 +66,7 @@ public class DefaultStompFrame extends DefaultStompHeadersSubframe implements St
 
     @Override
     public StompFrame replace(ByteBuf content) {
-        return new DefaultStompFrame(command, content, new DefaultStompHeaders().set(headers));
+        return new DefaultStompFrame(command, content, headers.copy());
     }
 
     @Override
--- a/codec-stomp/src/main/java/io/netty/handler/codec/stomp/DefaultStompHeadersSubframe.java
+++ b/codec-stomp/src/main/java/io/netty/handler/codec/stomp/DefaultStompHeadersSubframe.java
@@ -24,13 +24,19 @@ public class DefaultStompHeadersSubframe implements StompHeadersSubframe {
 
     protected final StompCommand command;
     protected DecoderResult decoderResult = DecoderResult.SUCCESS;
-    protected final StompHeaders headers = new DefaultStompHeaders();
+    protected final StompHeaders headers;
 
     public DefaultStompHeadersSubframe(StompCommand command) {
+        this(command, null);
+    }
+
+    public DefaultStompHeadersSubframe(StompCommand command, StompHeaders headers) {
         if (command == null) {
             throw new NullPointerException("command");
         }
+
         this.command = command;
+        this.headers = headers == null ? new DefaultStompHeaders() : headers;
     }
 
     @Override
--- a/codec-stomp/src/main/java/io/netty/handler/codec/stomp/DefaultStompHeadersSubframe.java
+++ b/codec-stomp/src/main/java/io/netty/handler/codec/stomp/DefaultStompHeadersSubframe.java
@@ -24,13 +24,13 @@ public class DefaultStompHeadersSubframe implements StompHeadersSubframe {
 
     protected final StompCommand command;
     protected DecoderResult decoderResult = DecoderResult.SUCCESS;
-    protected final StompHeaders headers;
+    protected final DefaultStompHeaders headers;
 
     public DefaultStompHeadersSubframe(StompCommand command) {
         this(command, null);
     }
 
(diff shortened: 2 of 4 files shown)


Scottmitch:
thanks @amizurov ... can you sign the [ICLA as referenced in the developer guide](https://netty.io/wiki/developer-guide.html)?

codec-stomp/src/main/java/io/netty/handler/codec/stomp/DefaultStompFrame.java
```diff
@@ -60,7 +66,7 @@ public StompFrame retainedDuplicate() {
 
     @Override
     public StompFrame replace(ByteBuf content) {
-        return new DefaultStompFrame(command, content);
+        return new DefaultStompFrame(command, content, new DefaultStompHeaders().set(headers));
```
Scottmitch: Please use the same strategy as https://github.com/netty/netty/commit/942b993f2b9781ff2126ff92a6be5b975dfc72ed#diff-17c17dd59c9912dd08e4dc95fa04bcab. In summary adding a `copy()` method on `DefaultStompHeaders` which leverages `DefaultHeaders#copy()`.

amizurov: I looked to this strategy but `DefaultStompHeaders extends DefaultHeader implements StompHeaders` we can't add `StompHeaders#copy()` method it will be clashes. Maybe we can add method with name like a `StompHeaders#deepCopy()` or another way ?

Scottmitch: We can't modify `StompHeaders` because it is an interface and that would be an API breaking change. However we can modify `DefaultStompHeaders` and do the following:

```java
    @Override
    public DefaultStompHeaders copy() {
        DefaultStompHeaders copy = new DefaultStompHeaders();
        copy.addImpl(this);
        return copy;
    }
```

We could also modify `DefaultHeaders` so we can consolidate the `copy.addImpl`, and any other work which maybe necessary in the future, but this maybe unnecessary.

amizurov: So we can, but in `DefaultStompHeadersSubframe` used interface `StompHeaders`, how we can used implementation method maybe I misunderstood you.

amizurov: @Scottmitch I understand that you are busy, but I would like to quickly close this defect.

amizurov:
@Scottmitch , @normanmaurer Guys what we do with this pull request ? For a long time no activity.

codec-stomp/src/main/java/io/netty/handler/codec/stomp/DefaultStompFrame.java
Scottmitch: > how we can access to method that not present in interface without casting or other trick? 

Unfortunately before jdk8 (which provides the `default` key word) extending interfaces is not possible without breaking API changes. So we have a few options:

1. Add a `DefaultStompHeaders copy()` method (as suggested above)
- change the member variable in `DefaultStompHeadersSubframe` so it is declared as type `DefaultStompHeaders`.
- make the new constructor in `DefaultStompHeadersSubframe` which currently takes a `StompHeaders` instead take a `DefaultStompHeaders` and make it package private
2. "break" the StompHeaders API and add a `StompHeaders copy()` method (and also override this method in `DefaultStompHeaders` as suggested above).

@normanmaurer - my guess is you would prefer the "non-breaking" API change?

normanmaurer:
Yes

amizurov:
Thanks a lot for your comments, please check that is all right.

codec-stomp/src/test/java/io/netty/handler/codec/stomp/DefaultStompFrameTest.java
```diff
@@ -0,0 +1,52 @@
...
+        AsciiString copyHeaderName = new AsciiString("foo");
+        AsciiString copyHeaderValue = new AsciiString("bar");
+        copyFrame.headers().set(copyHeaderName, copyHeaderValue);
+
+        assertFalse(sourceFrame.headers().contains(copyHeaderName, copyHeaderValue));
+        assertTrue(copyFrame.headers().contains(copyHeaderName, copyHeaderValue));
+
+        assertEquals(1, sourceFrame.headers().size());
+        assertEquals(2, copyFrame.headers().size());
+        assertNotEquals(sourceFrame.headers(), copyFrame.headers());
```
normanmaurer: I think you also need to call `release()` on the frames ?

amizurov: You mean that `release()` not affect a copied frame or that is work correct by spec ?

normanmaurer: Basically add:

```java
sourceFrame.release();
copyFrame.release();
```

This is needed as `DefaultStompFrame` implementations `ByteBufHolder` to ensure there are no memory leaks.

normanmaurer:
@Scottmitch please pull in once satisfied .

codec-stomp/src/main/java/io/netty/handler/codec/stomp/DefaultStompHeadersSubframe.java
```diff
@@ -24,13 +24,19 @@
 
     protected final StompCommand command;
     protected DecoderResult decoderResult = DecoderResult.SUCCESS;
-    protected final StompHeaders headers = new DefaultStompHeaders();
+    protected final DefaultStompHeaders headers;
```
normanmaurer: @amizurov can you explain why the changes to this class are needed at all ? I think you can revert the changes

amizurov: So, when we create `new DefaultStompFrame` in `DefaultStompFrame #replace(...)` method we pass headers.copy() to the constructor.

normanmaurer: But not to this class

amizurov: It's super class for `DefaultStompFrame`, else it's will be look like this: 
 ```java
DefaultStompFrame(StompCommand command, ByteBuf content, DefaultStompHeaders 
headers)  {
        super(command);

        if (headers != null) {
            this.headers.set(headers);
        }

        if (content == null) {
            throw new NullPointerException("content");
        }
        this.content = content;
    }

Scottmitch:
@normanmaurer - since you have an outstanding question I'll let you merge when satisfied :)

codec-stomp/src/main/java/io/netty/handler/codec/stomp/DefaultStompHeadersSubframe.java
normanmaurer: never mind missed it is the super-class.

normanmaurer:
Squashed and cherry-picked into 4.1 (4afabd3f3c400109b82951e8bec95753103ea915).

@amizurov thanks a lot!

PR #7563, opened 2018-01-03
IovArray#add return value resulted in more ByteBufs being added durin…

Scottmitch:
…g iteration

Motivation:
IovArray implements MessageProcessor, and the processMessage method will continue to be called during iteration until it returns true. A recent commit b215794de31f28355e4469fcc04782f55076c80c changed the return value to only return true if any component of a CompositeByteBuf was added as a result of the method call. However this results in the iteration continuing, and potentially subsequent smaller buffers maybe added, which will result in out of order writes and generally corrupts data.

Modifications:
- IovArray#add should return false so that the MessageProcessor#processMessage will stop iterating.

Result:
Native transports which use IovArray will not corrupt data during gathering writes of CompositeByteBuf objects.

### DIFF
--- a/testsuite/src/main/java/io/netty/testsuite/transport/socket/CompositeBufferGatheringWriteTest.java
+++ b/testsuite/src/main/java/io/netty/testsuite/transport/socket/CompositeBufferGatheringWriteTest.java
@@ -21,14 +21,18 @@ import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.CompositeByteBuf;
 import io.netty.channel.Channel;
+import io.netty.channel.ChannelConfig;
 import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.channel.ChannelInitializer;
+import io.netty.channel.ChannelOption;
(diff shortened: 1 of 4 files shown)


testsuite/src/main/java/io/netty/testsuite/transport/socket/CompositeBufferGatheringWriteTest.java
```diff
@@ -133,6 +137,147 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
...
+                            compositeByteBuf.addComponent(true,
+                                    expectedContent.retainedSlice(expectedContent.readerIndex() + offset, 50));
+                            offset += 50;
+                            compositeByteBuf.addComponent(true,
+                                    expectedContent.retainedSlice(expectedContent.readerIndex() + offset, 200));
+                            offset += 200;
+                            ctx.write(compositeByteBuf);
+
+                            // Write a single buffer that is smaller than the second component of the CompositeByteBuf
+                            // above and will fit in the
```
trustin: sentence not finished?

trustin:
Seems like the data corruption is now gone. Thanks, @Scottmitch !

johnou:
@trustin would you please better explain the situation in which you experienced data corruption because of this regression?

Scottmitch:
@johnou - as described in the commit message writing `CompositeByteBuf` may cause issues in general with the JNI based transports. We will get a release out this week which include this fix.

PR #7564, opened 2018-01-04
Reduce memory allocations in StompSubframeDecoder.readHeaders

xehonk:
Motivation:
When decoding stomp frames a lot of unnecessary character arrays are created while parsing the headers. For every header, an array is created to read the line into and then more when splitting the line at the colon.

Modifications:
- Parse key and value of a header while reading the line instead of afterwards.
- Reuse a single AppendableCharSequence.
- Reduce initial size of AppendableCharSequence when reading the command as it is expected to be short.

Result:
Allocations when parsing stomp frames are reduced significantly.

### DIFF
--- a/codec-stomp/src/main/java/io/netty/handler/codec/stomp/StompSubframeDecoder.java
+++ b/codec-stomp/src/main/java/io/netty/handler/codec/stomp/StompSubframeDecoder.java
@@ -196,7 +196,7 @@ public class StompSubframeDecoder extends ReplayingDecoder<State> {
     }
 
     private StompCommand readCommand(ByteBuf in) {
-        String commandStr = readLine(in, maxLineLength);
+        String commandStr = readLine(in, 16);
         StompCommand command = null;
         try {
             command = StompCommand.valueOf(commandStr);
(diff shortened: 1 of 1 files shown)


normanmaurer:
@Scottmitch please merge once satisfied

codec-stomp/src/main/java/io/netty/handler/codec/stomp/StompSubframeDecoder.java
```diff
@@ -266,28 +259,71 @@ private static void skipControlCharacters(ByteBuf buffer) {
...
+                } else if (validateHeaders) {
+                    invalidHeader(key, buf.toString());
+                }
+                return true;
+            } else {
+                if (lineLength >= maxLineLength) {
+                    invalidLineLength();
+                }
+                if (nextByte == StompConstants.COLON && key != null) {
+                    valid = false;
```
fenik17: Maybe we should call an `invalidHeader()` here?

xehonk: While the code would certainly look better, it would have side effects:
- the generated error message would not contain the entire header
- in case validation is disabled, the next header would be read starting after this point, not after the line break

fenik17: Ok :)

normanmaurer:
Cherry-picked into 4.1 (27ff15319c2e566c44e333b9bd59cf210c130c2f).

@xehonk thanks a lot!

PR #7567, opened 2018-01-05
HttpResponseStatus object allocation reduction

Scottmitch:
Motivation:
Usages of HttpResponseStatus may result in more object allocation then necessary due to not looking for cached objects and the AsciiString parsing method not being used due to CharSequence method being used instead.

Modifications:
- HttpResponseDecoder should attempt to get the HttpResponseStatus from cache instead of allocating a new object
- HttpResponseStatus#parseLine(CharSequence) should check if the type is AsciiString and redirect to the AsciiString parsing method which may not require an additional toString call
- HttpResponseStatus#parseLine(AsciiString) can be optimized and doesn't require and may not require object allocation

Result:
Less allocations when dealing with HttpResponseStatus.


codec-http/src/main/java/io/netty/handler/codec/http/HttpResponseStatus.java
```diff
@@ -446,6 +447,20 @@ public static HttpResponseStatus valueOf(int code) {
         return new HttpResponseStatus(code);
     }
 
+    /**
+     * Returns the {@link HttpResponseStatus} represented by the specified {@code code} and {@code reasonPhrase}.
+     * If the specified code is a standard HTTP getStatus {@code code} and {@code reasonPhrase}, a cached instance
```
normanmaurer: getStatus/status

codec-http/src/main/java/io/netty/handler/codec/http/HttpResponseStatus.java
```diff
@@ -456,79 +471,25 @@ public static HttpResponseStatus valueOf(int code) {
...
-        }
-
-        public HttpResponseStatus status() {
-            if (state <= 1) {
-                parseStatus(string.length());
-                state = 3;
-            }
-            return status;
+    /**
+     * Parses the specified HTTP status line into a {@link HttpResponseStatus}.  The expected formats of the line are:
```
normanmaurer: Nit: double space after the dot

codec-http/src/main/java/io/netty/handler/codec/http/HttpResponseStatus.java
```diff
@@ -447,7 +449,21 @@ public static HttpResponseStatus valueOf(int code) {
...
+     * Returns the {@link HttpResponseStatus} represented by the specified {@code code} and {@code reasonPhrase}.
+     * If the specified code is a standard HTTP status {@code code} and {@code reasonPhrase}, a cached instance
+     * will be returned. Otherwise, a new instance will be returned.
+     * @param code The response code value.
+     * @param reasonPhrase The response code reason phrase.
+     * @return the {@link HttpResponseStatus} represented by the specified {@code code} and {@code reasonPhrase}.
+     */
+    public static HttpResponseStatus valueOf(int code, String reasonPhrase) {
+        HttpResponseStatus responseStatus = valueOf(code);
+        return responseStatus.reasonPhrase().contentEquals(reasonPhrase) ? responseStatus :
```
fenik17: contentEquals -> equals

codec-http/src/main/java/io/netty/handler/codec/http/HttpResponseStatus.java
```diff
@@ -447,7 +449,21 @@ public static HttpResponseStatus valueOf(int code) {
...
+                new HttpResponseStatus(code, reasonPhrase);
```
fenik17: @Scottmitch If the `code` is not represented in this class, will happen a double instatuation of `HttpResponseStatus`: first in `valueOf(code)`, and second at this line. I think we can avoid this. WDYT?

fenik17:
I propose one more little optimization. We can reduce the amount of  `int` to `string` conversions in constructor:

```java
        this.code = code;
        String codeStr = Integer.toString(code); // <- extract explicit variable
        codeAsText = new AsciiString(codeStr); // use it 1
        this.reasonPhrase = reasonPhrase;
        if (bytes) {
            this.bytes = (codeStr + ' ' + reasonPhrase).getBytes(CharsetUtil.US_ASCII); // use it 2
        } else {
            this.bytes = null;
        }
```

This only affects the class initialization.

Also maybe make sense to make a `codeClass` as `final` and initialize it in the constructor. Or just remove this field, because it is rarely used and `HttpStatusClass.valueOf(code)` have a low cost.

normanmaurer:
@Scottmitch PTAL

codec-http/src/main/java/io/netty/handler/codec/http/HttpResponseStatus.java
Scottmitch: I don't see any clear guidance as if it should be treated as case insensitive or not. can you point me to the RFC which defines the case sensitivity here? If possible I would like to avoid object allocation even if the case doesn't match up.

https://tools.ietf.org/html/rfc7230#section-3.1.2
> A client SHOULD ignore the reason-phrase content.
> reason-phrase  = *( HTAB / SP / VCHAR / obs-text )

codec-http/src/main/java/io/netty/handler/codec/http/HttpResponseStatus.java
Scottmitch: good catch ... yes we can avoid the double allocation

Scottmitch:
> I propose one more little optimization.

done. thanks!

> Or just remove this field, because it is rarely used and HttpStatusClass.valueOf(code) have a low cost.

Lets consider this as a followup PR. It is used and I would like to understand the impacts for those cases before making the change.

Scottmitch:
@fenik17 - one outstanding question https://github.com/netty/netty/pull/7567#discussion_r163131678

codec-http/src/main/java/io/netty/handler/codec/http/HttpResponseStatus.java
normanmaurer: IMHO `contentEquals` is fine here.

fenik17: @Scottmitch I mean just, if we compare two string, it's better to use `equals` instead of `contentEquals`. This is not related with case sensitivity.

Scottmitch:
@normanmaurer - you just "commented" but didn't "approve" ... are you satisfied?

normanmaurer:
Yes... just approved as well

PR #7577, opened 2018-01-13
Fixes #7566 by handling concatenated GZIP streams.

### LINKED ISSUE #7566: JdkZlibDecoder ignores everything but first GZIP stream, non-compliant with RFC 1952
### Expected behavior
gzip content is fully consumed
### Actual behavior
gzip content is partially consumed
### Steps to reproduce
n/a
n/a
### Netty version
4.1.17
n/a
n/a

I'm using Spring WebClient to read a large (12 million records) gzipped response. I've set up the Netty pipeline as follows:
```
ctx.addHandlerLast(new ReadTimeoutHandler(readTimeoutMillis, TimeUnit.MILLISECONDS));
ctx.addHandlerLast(new Slf4JLoggingHandler());
if (forceDecompression) {
    io.netty.handler.codec.http.HttpHeaders httpHeaders = new ReadOnlyHttpHeaders(
            true,
            CONTENT_ENCODING, GZIP,
            CONTENT_TYPE, APPLICATION_JSON
    );
    HttpResponseHeadersHandler headersModifier = new HttpResponseHeadersHandler(httpHeaders);
    ctx.addHandlerFirst(headersModifier);
}
ctx.addHandlerLast(new HttpContentDecompressor());
```
The `HttpResponseHeadersHandler` overwrites headers in the response because the server doesn't send an `Content-Encoding` header. Details and implementation can be found [here](https://github.com/netty/netty/issues/7553#issuecomment-355249883).

Problem is, processing stops after about 130,000 records. No errors are shown. I've debugged and found out that the `JdkZlibDecoder` used for decompression stops reading and enters the finished block. I've tests that work fine for fewer number of records (100 or so).

However, using a different client (okhttp), if I don't set the `Content-Encoding` header and wrap the input stream with `GZIPInputStream`, everything works as expected.

So, question is, if `GZIPInputStream` doesn't have a problem, why does `JdkZlibDecoder`? And since it does, why won't it report what it thinks is the issue?

```
>> GET http://myserver.mycompany.com/.../businesses.20180104.json.gz
<< 200 OK
<< connection -> [keep-alive]
<< accept-ranges -> [bytes]
<< content-disposition -> [attachment; filename="businesses.20180104.json.gz"; filename*=UTF-8''businesses.20180104.json.gz] 
<< content-type -> [application/x-gzip]
<< content-length -> [3384998203]
<< date -> [Fri, 05 Jan 2018 00:43:32 GMT]
<< etag -> [0e49d5fa7ba9f68058bfbb4a98bef032c3a73871]
<< last-modified -> [Thu, 04 Jan 2018 23:54:26 GMT]
<< x-artifactory-id -> [9732f56568ea1e3d:59294f65:160b8066066:-8000]
<< x-checksum-md5 -> [451ca1b1414e7b511de874e61fd33eb2]
<< x-artifactory-filename -> [businesses.20180104.json.gz]
<< server -> [Artifactory/5.3.0]
<< x-checksum-sha1 -> [0e49d5fa7ba9f68058bfbb4a98bef032c3a73871]
```

For full disclosure, Initially posted a [question](https://stackoverflow.com/q/48094553/839733) on stackoverflow. Also related to https://github.com/square/okhttp/issues/3759.

asarkar:
Motivation:
According to [RFC 1952](https://tools.ietf.org/html/rfc1952), concatenation of valid gzip streams is also a valid gzip stream. `JdkZlibDecoder` only processed the first and discarded the rest.

Modifications:
- Introduced a constructor argument `decompressConcatenated` that if true, `JdkZlibDecoder` would continue to process the stream.

Result:
- Fixes #7566. 
- If `decompressConcatenated = true`, concatenated streams would be processed in
compliance with RFC 1952.
- If `decompressConcatenated = false` (default), existing behavior would remain.


yschimke:
Even if desired I suggested splitting out the maven changes into a separate PR.  I'll leave the review to Norman though.

asarkar:
@yschimke Created https://github.com/netty/netty/issues/7578 for Maven and corresponding PR https://github.com/netty/netty/pull/7579.

codec/src/main/java/io/netty/handler/codec/compression/JdkZlibDecoder.java
```diff
@@ -207,7 +217,15 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) t
             if (readFooter) {
                 gzipState = GzipState.FOOTER_START;
                 if (readGZIPFooter(in)) {
-                    finished = true;
+                    finished = !decompressConcatenated;
+
+                    if (finished) {
+                        inflater.end();
```
yschimke: It's really easy to have a slow native resource leak with Inflater if you don't close the inflater.  Make sure you close after the last stream.

codec/src/test/java/io/netty/handler/codec/compression/JdkZlibTest.java
```diff
@@ -35,4 +47,45 @@ protected ZlibDecoder createDecoder(ZlibWrapper wrapper) {
...
+        try {
+            byte[] bytes = IOUtils.toByteArray(getClass().getResourceAsStream("/multiple.gz"));
+
+            chDecoderGZip.writeInbound(Unpooled.copiedBuffer(bytes));
+            Queue<Object> messages = chDecoderGZip.inboundMessages();
+            assertEquals(2, messages.size());
+
+            for (String s : Arrays.asList("a", "b")) {
+                ByteBuf msg = (ByteBuf) messages.poll();
+                assertEquals(s, msg.toString(Charset.forName("UTF-8")));
```
normanmaurer: use `CharsetUtil.UTF_8`

codec/src/test/java/io/netty/handler/codec/compression/JdkZlibTest.java
```diff
@@ -35,4 +47,45 @@ protected ZlibDecoder createDecoder(ZlibWrapper wrapper) {
...
+                ReferenceCountUtil.release(msg);
+            }
+        } finally {
+            chDecoderGZip.close();
```
normanmaurer: use `assertFalse(chDecoderGZip.finish())` to verify that there is no more data to process.

codec/src/test/java/io/netty/handler/codec/compression/JdkZlibTest.java
```diff
@@ -35,4 +47,45 @@ protected ZlibDecoder createDecoder(ZlibWrapper wrapper) {
...
+
+            chDecoderGZip.writeInbound(Unpooled.copiedBuffer(bytes));
+            Queue<Object> messages = chDecoderGZip.inboundMessages();
+            assertEquals(1, messages.size());
+
+            ByteBuf msg = (ByteBuf) messages.poll();
+            assertEquals("a", msg.toString(Charset.forName("UTF-8")));
+            ReferenceCountUtil.release(msg);
+        } finally {
+            chDecoderGZip.close();
```
normanmaurer: use `assertFalse(chDecoderGZip.finish())` to verify that there is no more data to process.

codec/src/test/java/io/netty/handler/codec/compression/JdkZlibTest.java
```diff
@@ -35,4 +47,45 @@ protected ZlibDecoder createDecoder(ZlibWrapper wrapper) {
...
+
+    @Test
+    // verifies backward compatibility
+    public void testConcatenatedStreamsReadFirstOnly() throws IOException {
+        EmbeddedChannel chDecoderGZip = new EmbeddedChannel(createDecoder(ZlibWrapper.GZIP));
+
+        try {
+            byte[] bytes = IOUtils.toByteArray(getClass().getResourceAsStream("/multiple.gz"));
+
+            chDecoderGZip.writeInbound(Unpooled.copiedBuffer(bytes));
```
normanmaurer: assert return value

codec/src/test/java/io/netty/handler/codec/compression/JdkZlibTest.java
```diff
@@ -35,4 +47,45 @@ protected ZlibDecoder createDecoder(ZlibWrapper wrapper) {
...
+            Queue<Object> messages = chDecoderGZip.inboundMessages();
+            assertEquals(1, messages.size());
+
+            ByteBuf msg = (ByteBuf) messages.poll();
+            assertEquals("a", msg.toString(Charset.forName("UTF-8")));
```
normanmaurer: use CharsetUtil.UTF_8

codec/src/test/java/io/netty/handler/codec/compression/JdkZlibTest.java
```diff
@@ -35,4 +47,45 @@ protected ZlibDecoder createDecoder(ZlibWrapper wrapper) {
...
+    }
+
+    @Test
+    public void testConcatenatedStreamsReadFully() throws IOException {
+        EmbeddedChannel chDecoderGZip = new EmbeddedChannel(new JdkZlibDecoder(true));
+
+        try {
+            byte[] bytes = IOUtils.toByteArray(getClass().getResourceAsStream("/multiple.gz"));
+
+            chDecoderGZip.writeInbound(Unpooled.copiedBuffer(bytes));
```
normanmaurer: assert return value

codec/src/main/java/io/netty/handler/codec/compression/JdkZlibDecoder.java
normanmaurer: its already done

codec/src/main/java/io/netty/handler/codec/compression/JdkZlibDecoder.java
```diff
@@ -39,6 +39,7 @@
 
     // GZIP related
     private final ByteBufChecksum crc;
+    private boolean decompressConcatenated;
```
normanmaurer: final

codec/src/main/java/io/netty/handler/codec/compression/JdkZlibDecoder.java
asarkar: What's funny is that it wasn't closed before.

asarkar:
@normanmaurer Already signed ical before submitting PR. Are you not seeing it?

codec/src/main/java/io/netty/handler/codec/compression/JdkZlibDecoder.java
asarkar: Making this `final` forces boilerplate. Because of the `switch` in the constructor, the compiler thinks that the variable is not initialized, and it has to be explicitly initialized to `false` in all cases. I don't see any benefit in doing so since it's already `private`. WDYT?

codec/src/test/java/io/netty/handler/codec/compression/JdkZlibTest.java
asarkar: Done.

codec/src/test/java/io/netty/handler/codec/compression/JdkZlibTest.java
asarkar: Done. Due to JDK 6 compilation level, `StandardCharsets.UTF-8` can't be used.

codec/src/test/java/io/netty/handler/codec/compression/JdkZlibTest.java
asarkar: done.

codec/src/test/java/io/netty/handler/codec/compression/JdkZlibTest.java
asarkar: done

codec/src/test/java/io/netty/handler/codec/compression/JdkZlibTest.java
asarkar: Done. Due to JDK 6 compilation level, `StandardCharsets.UTF-8` can't be used.

codec/src/test/java/io/netty/handler/codec/compression/JdkZlibTest.java
asarkar: done

codec/src/main/java/io/netty/handler/codec/compression/JdkZlibDecoder.java
normanmaurer: I would prefer to make it final if it will never be changed. This may also allow optimisations for the JIT / JVM

codec/src/main/java/io/netty/handler/codec/compression/JdkZlibDecoder.java
normanmaurer: it was closed in `handlerRemoved` or I am missing anything ?

codec/src/main/java/io/netty/handler/codec/compression/JdkZlibDecoder.java
asarkar: Done.

codec/src/main/java/io/netty/handler/codec/compression/JdkZlibDecoder.java
asarkar: It was, I missed it. As a result, I've no removed the above `iniflater.end`, and the `else` block.

asarkar:
@normanmaurer This PR is ready to be merged.

normanmaurer:
Cherry-picked into 4.1 (6ff48dcbe311d2736249b02b31a8745e719d1304) and 4.0 (c870dbf3a7038b65903802b3c857e0f4e839c2de)

Issue #7565, opened 2018-01-04
Wrapped direct ByteBufs do not respect the position of the underlying buffer


sprsquish:
### Expected behavior

If I create a MappedByteBuffer then change it's position and limit then wrap that buffer, the ByteBuf should respect that position/offset.

### Actual behavior

`ReadOnlyUnsafeDirectByteBuf` does not respect the position of the underlying MappedByteBuffer and will start at the base address rather than the position. 

### Steps to reproduce

```scala
val f = java.nio.channels.FileChannel.open(java.nio.file.FileSystems.getDefault.getPath("somefile")) 
val m = f.map(java.nio.channels.FileChannel.MapMode.READ_ONLY, 0, f.size)

val b1 = io.netty.buffer.Unpooled.wrappedBuffer(m)

val dup = m.duplicate
dup.position(2)
dup.limit(4)

val b2 = io.netty.buffer.Unpooled.wrappedBuffer(dup)

b1.slice(2, 2) == b2 // this should be true, but is not because b2 will start at the wrapped buffer's 0 index
```

### Minimal yet complete reproducer code (or URL to code)

See above

### Netty version

4.1.19.Final

### JVM version (e.g. `java -version`)

java version "1.8.0_121"
Java(TM) SE Runtime Environment (build 1.8.0_121-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.121-b13, mixed mode)

### OS version (e.g. `uname -a`)

Linux sprsquish 4.10.0-42-generic #46-Ubuntu SMP Mon Dec 4 14:38:01 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

normanmaurer:
@sprsquish PTAL https://github.com/netty/netty/pull/7580 this should fix it. And thanks for the test code :)

normanmaurer:
Fixed by https://github.com/netty/netty/pull/7580

PR #7580, opened 2018-01-14
Correctly take position into account when wrap a ByteBuffer via ReadO…

normanmaurer:
…nlyUnsafeDirectByteBuf

Motivation:

We did not correctly take the position into account when wrapping a ByteBuffer via ReadOnlyUnsafeDirectByteBuf as we obtained the memory address from the original ByteBuffer and not the slice we take.

Modifications:

- Correctly use the slice to obtain memory address.
- Add test case.

Result:

Fixes [#7565].

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/ReadOnlyUnsafeDirectByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/ReadOnlyUnsafeDirectByteBuf.java
@@ -27,8 +27,10 @@ import java.nio.ByteBuffer;
 final class ReadOnlyUnsafeDirectByteBuf extends ReadOnlyByteBufferBuf {
     private final long memoryAddress;
 
-    ReadOnlyUnsafeDirectByteBuf(ByteBufAllocator allocator, ByteBuffer buffer) {
-        super(allocator, buffer);
+    ReadOnlyUnsafeDirectByteBuf(ByteBufAllocator allocator, ByteBuffer byteBuffer) {
+        super(allocator, byteBuffer);
+        // Use buffer as the super class will slice the passed in ByteBuffer which means the memoryAddress
+        // may be different if the position != 0.
         memoryAddress = PlatformDependent.directBufferAddress(buffer);
     }
 
--- a/buffer/src/test/java/io/netty/buffer/ReadOnlyDirectByteBufferBufTest.java
+++ b/buffer/src/test/java/io/netty/buffer/ReadOnlyDirectByteBufferBufTest.java
@@ -15,13 +15,18 @@
  */
 package io.netty.buffer;
 
+import io.netty.util.internal.PlatformDependent;
 import org.junit.Assert;
 import org.junit.Test;
 
 import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.ReadOnlyBufferException;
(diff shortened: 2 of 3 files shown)


normanmaurer:
@sprsquish PTAL

buffer/src/main/java/io/netty/buffer/ReadOnlyUnsafeDirectByteBuf.java
```diff
@@ -29,7 +29,9 @@
 
     ReadOnlyUnsafeDirectByteBuf(ByteBufAllocator allocator, ByteBuffer buffer) {
         super(allocator, buffer);
-        memoryAddress = PlatformDependent.directBufferAddress(buffer);
+        // Use this.buffer as the super class wills slice the passed in ByteBuffer which means the memoryAddress
```
asarkar: extra 's' in ''wills'

buffer/src/main/java/io/netty/buffer/ReadOnlyUnsafeDirectByteBuf.java
```diff
@@ -27,8 +27,10 @@
 final class ReadOnlyUnsafeDirectByteBuf extends ReadOnlyByteBufferBuf {
     private final long memoryAddress;
 
-    ReadOnlyUnsafeDirectByteBuf(ByteBufAllocator allocator, ByteBuffer buffer) {
-        super(allocator, buffer);
+    ReadOnlyUnsafeDirectByteBuf(ByteBufAllocator allocator, ByteBuffer wrapped) {
+        super(allocator, wrapped);
```
trustin: I guess the name `buffer` is correct since it's not wrapped just yet.

normanmaurer: Let me rename it to buf so we not make the mistake to use it in the future

normanmaurer:
Cherry-picked into 4.1 (819b870b8a08414b534ac755c55393a70d97e14c) and 4.0 (b817cefa225d8b65d3a06516b8325b493b6be218)

buffer/src/main/java/io/netty/buffer/ReadOnlyUnsafeDirectByteBuf.java
```diff
@@ -27,8 +27,10 @@
 final class ReadOnlyUnsafeDirectByteBuf extends ReadOnlyByteBufferBuf {
     private final long memoryAddress;
 
-    ReadOnlyUnsafeDirectByteBuf(ByteBufAllocator allocator, ByteBuffer buffer) {
-        super(allocator, buffer);
+    ReadOnlyUnsafeDirectByteBuf(ByteBufAllocator allocator, ByteBuffer byteBuffer) {
+        super(allocator, byteBuffer);
+        // Use buffer as the super class will slice the passed in ByteBuffer which means the memoryAddress
+        // may be different if the position != 0.
```
carl-mastrangelo: I must be blind, what changed here?

Scottmitch: I had to do a double take when I was reviewing as well. `super()` uses the `ByteBuffer` but the line below uses the `buffer` from the parent class.

normanmaurer: @carl-mastrangelo as @Scottmitch said... we now use the `ByteBuffer` stored in super. Before we used the one that was passed into the constructor.

Issue #7583, opened 2018-01-15
io.netty.resolver.dns.DnsNameResolver#doResolveCached race condition


johnou:
### Expected behavior

No exception.

### Actual behavior

```
2018-01-15 05:33:32,915 ERROR [Redisson-Epoll-1-15] (org.redisson.connection.DNSMonitor) Unable to resolve xxx.0001.use1.cache.amazonaws.com
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
    at java.util.ArrayList.rangeCheck(ArrayList.java:653) ~[?:1.8.0_144]
    at java.util.ArrayList.get(ArrayList.java:429) ~[?:1.8.0_144]
    at io.netty.resolver.dns.DnsNameResolver.doResolveCached(DnsNameResolver.java:613) ~[netty-resolver-dns-4.1.19.Final.jar:4.1.19.Final]
    at io.netty.resolver.dns.DnsNameResolver.doResolve(DnsNameResolver.java:593) ~[netty-resolver-dns-4.1.19.Final.jar:4.1.19.Final]
    at io.netty.resolver.dns.DnsNameResolver.doResolve(DnsNameResolver.java:527) ~[netty-resolver-dns-4.1.19.Final.jar:4.1.19.Final]
    at io.netty.resolver.SimpleNameResolver.resolve(SimpleNameResolver.java:63) ~[netty-resolver-4.1.19.Final.jar:4.1.19.Final]
    at io.netty.resolver.dns.InflightNameResolver.resolve(InflightNameResolver.java:100) ~[netty-resolver-dns-4.1.19.Final.jar:4.1.19.Final]
    at io.netty.resolver.dns.InflightNameResolver.resolve(InflightNameResolver.java:66) ~[netty-resolver-dns-4.1.19.Final.jar:4.1.19.Final]
    at io.netty.resolver.dns.InflightNameResolver.resolve(InflightNameResolver.java:51) ~[netty-resolver-dns-4.1.19.Final.jar:4.1.19.Final]
    at io.netty.resolver.InetSocketAddressResolver.doResolve(InetSocketAddressResolver.java:57) ~[netty-resolver-4.1.19.Final.jar:4.1.19.Final]
    at io.netty.resolver.InetSocketAddressResolver.doResolve(InetSocketAddressResolver.java:32) ~[netty-resolver-4.1.19.Final.jar:4.1.19.Final]
[...]

PR #7584, opened 2018-01-15
Fix concurrency issue in DnsNameResolver when DefaultDnsCache is used.

normanmaurer:
Motivation:

We need to ensure we only call List.* methods in the synchronized block as the returned List may not be thread-safe.

Modifications:

Do not call isEmpty() outside of the synchronized block.

Result:

Fixes [#7583]


normanmaurer:
@johnou FYI

johnou:
@normanmaurer lgtm

Scottmitch:
I will wait until https://github.com/netty/netty/pull/7587 is merged and this is re-based before reviewing. Please ping me when this happens.

normanmaurer:
@Scottmitch actually this is the "quick fix" for the mentioned issue and involves less changes. We should work on #7587 for a more complete fix tho but I am not sure if I have the cycles atm. So let me pull this one in as its low-risk and will fix the issue for now.

normanmaurer:
Cherry-picked into 4.1 (f3c6da32d79b1053ca4667191fa7a660a457b44e)

PR #7587, opened 2018-01-16
Fix race-condition when using DnsCache in DnsNameResolver

normanmaurer:
Motivation:

The usage of DnsCache in DnsNameResolver was racy in general. First of the isEmpty() was not called in a synchronized block while we depended on synchronized. The other problem was that this whole synchronization only worked if the DefaultDnsCache was used and the returned List was not wrapped by the user.

Modifications:

- Rewrite DefaultDnsCache to not depend on synchronization on the returned List.

Result:

Fixes [#7583] and other races.

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
@@ -23,11 +23,13 @@ import io.netty.util.internal.UnstableApi;
 
 import java.net.InetAddress;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
 import static io.netty.util.internal.ObjectUtil.checkNotNull;
 import static io.netty.util.internal.ObjectUtil.checkPositiveOrZero;
@@ -39,8 +41,8 @@ import static io.netty.util.internal.ObjectUtil.checkPositiveOrZero;
 @UnstableApi
 public class DefaultDnsCache implements DnsCache {
 
-    private final ConcurrentMap<String, List<DefaultDnsCacheEntry>> resolveCache =
-                                                            PlatformDependent.newConcurrentHashMap();
+    private final ConcurrentMap<String, Entries> resolveCache = PlatformDependent.newConcurrentHashMap();
+
     private final int minTtl;
     private final int maxTtl;
     private final int negativeTtl;
@@ -97,28 +99,21 @@ public class DefaultDnsCache implements DnsCache {
 
     @Override
     public void clear() {
-        for (Iterator<Map.Entry<String, List<DefaultDnsCacheEntry>>> i = resolveCache.entrySet().iterator();
-             i.hasNext();) {
-            final Map.Entry<String, List<DefaultDnsCacheEntry>> e = i.next();
-            i.remove();
-            cancelExpiration(e.getValue());
+        while (!resolveCache.isEmpty()) {
+            for (Iterator<Map.Entry<String, Entries>> i = resolveCache.entrySet().iterator(); i.hasNext();) {
+                Map.Entry<String, Entries> e = i.next();
+                i.remove();
+
+                e.getValue().cancelExpiration();
+            }
         }
     }
 
     @Override
     public boolean clear(String hostname) {
         checkNotNull(hostname, "hostname");
-        boolean removed = false;
-        for (Iterator<Map.Entry<String, List<DefaultDnsCacheEntry>>> i = resolveCache.entrySet().iterator();
-             i.hasNext();) {
-            final Map.Entry<String, List<DefaultDnsCacheEntry>> e = i.next();
-            if (e.getKey().equals(hostname)) {
-                i.remove();
-                cancelExpiration(e.getValue());
-                removed = true;
-            }
-        }
-        return removed;
+        Entries entries = resolveCache.remove(hostname);
+        return entries != null && entries.cancelExpiration();
     }
 
     private static boolean emptyAdditionals(DnsRecord[] additionals) {
@@ -129,22 +124,11 @@ public class DefaultDnsCache implements DnsCache {
     public List<? extends DnsCacheEntry> get(String hostname, DnsRecord[] additionals) {
         checkNotNull(hostname, "hostname");
         if (!emptyAdditionals(additionals)) {
-            return null;
+            return Collections.<DnsCacheEntry>emptyList();
         }
-        return resolveCache.get(hostname);
-    }
 
-    private List<DefaultDnsCacheEntry> cachedEntries(String hostname) {
-        List<DefaultDnsCacheEntry> oldEntries = resolveCache.get(hostname);
-        final List<DefaultDnsCacheEntry> entries;
-        if (oldEntries == null) {
-            List<DefaultDnsCacheEntry> newEntries = new ArrayList<DefaultDnsCacheEntry>(8);
-            oldEntries = resolveCache.putIfAbsent(hostname, newEntries);
-            entries = oldEntries != null? oldEntries : newEntries;
-        } else {
-            entries = oldEntries;
-        }
-        return entries;
+        Entries entries = resolveCache.get(hostname);
+        return entries == null ? null : entries.get();
     }
 
     @Override
@@ -157,22 +141,7 @@ public class DefaultDnsCache implements DnsCache {
         if (maxTtl == 0 || !emptyAdditionals(additionals)) {
             return e;
         }
-        final int ttl = Math.max(minTtl, (int) Math.min(maxTtl, originalTtl));
-        final List<DefaultDnsCacheEntry> entries = cachedEntries(hostname);
-
-        synchronized (entries) {
-            if (!entries.isEmpty()) {
-                final DefaultDnsCacheEntry firstEntry = entries.get(0);
-                if (firstEntry.cause() != null) {
-                    assert entries.size() == 1;
-                    firstEntry.cancelExpiration();
-                    entries.clear();
-                }
-            }
-            entries.add(e);
-        }
-
-        scheduleCacheExpiration(entries, e, ttl, loop);
+        cache0(e, Math.max(minTtl, (int) Math.min(maxTtl, originalTtl)), loop);
         return e;
     }
 
@@ -186,40 +155,34 @@ public class DefaultDnsCache implements DnsCache {
         if (negativeTtl == 0 || !emptyAdditionals(additionals)) {
             return e;
         }
-        final List<DefaultDnsCacheEntry> entries = cachedEntries(hostname);
 
-        synchronized (entries) {
-            final int numEntries = entries.size();
-            for (int i = 0; i < numEntries; i ++) {
-                entries.get(i).cancelExpiration();
-            }
-            entries.clear();
-            entries.add(e);
-        }
-
-        scheduleCacheExpiration(entries, e, negativeTtl, loop);
+        cache0(e, negativeTtl, loop);
         return e;
     }
 
-    private static void cancelExpiration(List<DefaultDnsCacheEntry> entries) {
-        final int numEntries = entries.size();
-        for (int i = 0; i < numEntries; i++) {
-            entries.get(i).cancelExpiration();
+    private void cache0(DefaultDnsCacheEntry e, int ttl, EventLoop loop) {
+        Entries entries = resolveCache.get(e.hostname());
+        if (entries == null) {
+            entries = new Entries(e);
+            Entries oldEntries = resolveCache.putIfAbsent(e.hostname(), entries);
+            if (oldEntries != null) {
+                entries = oldEntries;
+                entries.add(e);
+            }
         }
+
+        scheduleCacheExpiration(entries, e, ttl, loop);
     }
 
-    private void scheduleCacheExpiration(final List<DefaultDnsCacheEntry> entries,
+    private void scheduleCacheExpiration(final Entries entries,
                                          final DefaultDnsCacheEntry e,
                                          int ttl,
                                          EventLoop loop) {
         e.scheduleExpiration(loop, new Runnable() {
                     @Override
                     public void run() {
-                        synchronized (entries) {
-                            entries.remove(e);
-                            if (entries.isEmpty()) {
-                                resolveCache.remove(e.hostname());
-                            }
+                        if (entries.remove(e)) {
+                            resolveCache.remove(e.hostname);
                         }
                     }
                 }, ttl, TimeUnit.SECONDS);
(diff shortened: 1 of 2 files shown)


resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -289,4 +278,92 @@ public String toString() {
             }
         }
     }
+
+    private static final class Entries {
+        private final ReadWriteLock lock = new ReentrantReadWriteLock();
+
```
johnou:
```java
private final Lock r = lock.readLock();
private final Lock w = lock.writeLock();
```

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -289,4 +278,92 @@ public String toString() {
...
+        // or removed.
+        private List<DefaultDnsCacheEntry> entries;
+
+        Entries(DefaultDnsCacheEntry entry) {
+            entries = Collections.unmodifiableList(Collections.singletonList(entry));
+        }
+
+        List<DefaultDnsCacheEntry> get() {
+            Lock readLock = lock.readLock();
+            readLock.lock();
```
johnou: is the read lock necessary?

normanmaurer: Yes as we read the field... of we update it we use the write lock

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
normanmaurer: done

johnou:
lgtm!

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -289,4 +278,90 @@ public String toString() {
...
+                    entries = Collections.unmodifiableList(newEntries);
+                } finally {
+                    writeLock.unlock();
+                }
+            } else {
+                List<DefaultDnsCacheEntry> oldEntries;
+
+                writeLock.lock();
+                try {
+                    oldEntries = entries;
```
Scottmitch: if this field is treated as copy on write ... can we just use a volatile and avoid the locking? this would make things much more clear because we let the list escape the lock above in `get()`.

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -289,4 +278,90 @@ public String toString() {
...
+        Entries(DefaultDnsCacheEntry entry) {
+            entries = Collections.unmodifiableList(Collections.singletonList(entry));
+        }
+
+        List<DefaultDnsCacheEntry> get() {
+            readLock.lock();
+            try {
+                return entries;
+            } finally {
+                readLock.unlock();
```
Scottmitch: the list escapes the lock here which may violate what ever locking/synchronization enforced by this class internally.

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -186,16 +178,16 @@ public DnsCacheEntry cache(String hostname, DnsRecord[] additionals, Throwable c
         if (negativeTtl == 0 || !emptyAdditionals(additionals)) {
             return e;
         }
-        final List<DefaultDnsCacheEntry> entries = cachedEntries(hostname);
 
-        synchronized (entries) {
-            final int numEntries = entries.size();
-            for (int i = 0; i < numEntries; i ++) {
-                entries.get(i).cancelExpiration();
+        Entries entries = resolveCache.get(hostname);
```
Scottmitch: nit: share code with the previous method?

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -606,21 +606,19 @@ private boolean doResolveCached(String hostname,
 
         InetAddress address = null;
         Throwable cause = null;
-        synchronized (cachedEntries) {
-            final int numEntries = cachedEntries.size();
-            assert numEntries > 0;
+        final int numEntries = cachedEntries.size();
```
Scottmitch: move this into the conditional block where it is actually used. If the assertion is not there `get(0)` will throw, and we already check for `isEmpty()` above.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -736,22 +734,20 @@ private boolean doResolveAllCached(String hostname,
 
         List<InetAddress> result = null;
         Throwable cause = null;
-        synchronized (cachedEntries) {
-            final int numEntries = cachedEntries.size();
-            assert numEntries > 0;
+        final int numEntries = cachedEntries.size();
```
Scottmitch: move this into the conditional block where it is actually used. If the assertion is not there `get(0)` will throw, and we already check for `isEmpty()` above.

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
normanmaurer: Yeah I think this would work...  Let me do this

normanmaurer:
@johnou PTAL again... also @Scottmitch and @trustin

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -97,28 +101,30 @@ public int negativeTtl() {
...
+            List<DefaultDnsCacheEntry> entryList = e.getValue().get();
+            if (entryList != null) {
+                cancelExpiration(entryList);
+            }
         }
     }
 
     @Override
     public boolean clear(String hostname) {
         checkNotNull(hostname, "hostname");
```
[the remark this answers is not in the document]
normanmaurer: I think I would keep it for now to make it clear we not support null at all.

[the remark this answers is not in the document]
Scottmitch: not a blocker but the decision to "not support null" could be consolidated in a single place (where we add to `resolveCache`) instead of spread out throughout the code.

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -289,4 +253,92 @@ public String toString() {
             }
         }
     }
+
+    // Directly extend AtomicReference for intrinsics and also to keep memory overhead low.
+    private static final class Entries extends AtomicReference<List<DefaultDnsCacheEntry>> {
+
+        Entries(DefaultDnsCacheEntry entry) {
+            set(Collections.singletonList(entry));
```
Scottmitch: `set` -> `super`?

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -289,4 +253,92 @@ public String toString() {
...
+                    entries.get(i).cancelExpiration();
+                }
+            }
+        }
+
+        boolean remove(DefaultDnsCacheEntry entry) {
+            for (;;) {
+                List<DefaultDnsCacheEntry> entries = get();
+
+                if (entries.isEmpty()) {
```
Scottmitch: we haven't verified that `entry` is in the `entries` list and so do we need a special case for this condition here?

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -289,4 +253,92 @@ public String toString() {
...
+                    if (compareAndSet(entries, Collections.<DefaultDnsCacheEntry>emptyList())) {
+                        return false;
+                    }
+                } else {
+                    // Just size the new ArrayList as before as we may not find the entry we are looking for and not
```
Scottmitch: Can we clarify assumptions here? If it is not likely we will find the element then it seems beneficial to iterate first, and then only allocate/copy if we find the element. if it is likely we will find the element then the current approach (always allocate/copy) seems reasonable.

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -289,4 +253,92 @@ public String toString() {
...
+                    // want to cause an extra allocation / memory copy in this case.
+                    List<DefaultDnsCacheEntry> newEntries = new ArrayList<DefaultDnsCacheEntry>(entries.size());
+                    for (int i = 0; i < entries.size(); i++) {
+                        DefaultDnsCacheEntry e = entries.get(i);
+                        if (!e.equals(entry)) {
+                            newEntries.add(e);
+                        }
+                    }
+                    if (newEntries.isEmpty()) {
```
Scottmitch: would it be better to handle this by checking `if (entries.size() == 1 && entry.get(0).equals(newEntry)) { swap with empty collection }`? or is it possible that `entries` may contain multiple elements which equal `entry`?

resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
```diff
@@ -1045,12 +1045,12 @@ int dnsRedirectPort(InetAddress server) {
             QuerySucceededEvent succeededEvent = (QuerySucceededEvent) observer.events.poll();
 
             if (cache) {
-                assertNull(nsCache.cache.get("io.", null));
-                assertNull(nsCache.cache.get("netty.io.", null));
+                assertTrue(nsCache.cache.get("io.", null).isEmpty());
```
Scottmitch: does it ever make sense to return an empty list, as opposed to just null? if we return empty lists then conditionals tend to get more complicated (e.g. `if not empty and not null`), and the tests should likely be updated to reflect this too.

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
normanmaurer: @Scottmitch we should most likely find it. Thats why I choose the way I am doing it here.

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
normanmaurer: @Scottmitch it may be possible to have multiple entries so I think the way its handled now it more safe.

resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
normanmaurer: @Scottmitch I was thinking about this and decided that it may be better to return an empty list mainly because we not state in the javadocs that we will ever return null from the `DnsCache` interface.

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
normanmaurer: I think if its empty its safe to return just `false`.

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
normanmaurer: TIL: didn't know there is a constructor for this :)

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
Scottmitch: Please include comment with this rational.

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
Scottmitch: `DefaultDnsCacheEntry` doesn't implement `equals`, and on each `cache(..)` call a new element is created. Can you clarify under which conditions the `entries` array would contain the same element multiple times?

resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
Scottmitch: We have been returning `null` so perhaps it is best to update the javadocs. `null` is part of the language and there is no reliable way to express a java 6 API which won't return null, which means folks have to `null` check anyways at runtime.

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
Scottmitch: the question was related to https://github.com/netty/netty/pull/7587#discussion_r162137171 below. I was wondering if this special case could be removed if we first search for an item and then allocate an entry / array copy around it. If we don't find the element we can bail without a special case for empty.

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
normanmaurer: @Scottmitch actually scratch this... you are right it cant happen.

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -289,4 +252,97 @@ public String toString() {
...
+                    if (compareAndSet(entries, Collections.<DefaultDnsCacheEntry>emptyList())) {
+                        return false;
+                    }
+                } else {
+                    // Just size the new ArrayList as before as we may not find the entry we are looking for and not
+                    // want to cause an extra allocation / memory copy in this case.
+                    //
+                    // Its very likely we find the entry we are looking for so we directly create a new ArrayList
+                    // and fill it.
+                    List<DefaultDnsCacheEntry> newEntries = new ArrayList<DefaultDnsCacheEntry>(size);
```
Scottmitch: if it is "very likely" we will find the element should we create the `ArrayList` with `size - 1`?

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -289,4 +252,97 @@ public String toString() {
...
+                List<DefaultDnsCacheEntry> entries = getAndSet(Collections.singletonList(e));
+                cancelExpiration(entries);
+            }
+        }
+
+        boolean remove(DefaultDnsCacheEntry entry) {
+            for (;;) {
+                List<DefaultDnsCacheEntry> entries = get();
+                int size = entries.size();
+                if (size == 0 || (size == 1 && entries.get(0).equals(entry))) {
```
Scottmitch: if `size == 1` and `!entries.get(0).equals(entry)` then we don't need to go to the else condition (which will allocate/copy for no reason).

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
Scottmitch: Now that we have a special check for `size == 1` before this conditional block executes ... is this ever possible?

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
normanmaurer: honestly I don't care too much but at the end it ay be just good to be safe and not sorry... allocating an array which has one element more then needed is preferred over a memory copy because we need to resize imho.

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
normanmaurer: no... let me remove it.

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
normanmaurer: yep we can just return false

normanmaurer:
@Scottmitch and another time

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -289,4 +252,99 @@ public String toString() {
...
+                    // If the list is empty we just return early and so not allocate a new ArrayList.
+                    if (compareAndSet(entries, Collections.<DefaultDnsCacheEntry>emptyList())) {
+                        return false;
+                    }
+                } else if (size == 1) {
+                    if (entries.get(0).equals(entry)) {
+                        // If the list is empty we just return early and so not allocate a new ArrayList.
+                        if (compareAndSet(entries, Collections.<DefaultDnsCacheEntry>emptyList())) {
+                            return false;
+                        }
```
Scottmitch: it looks like we should return `true` in an else condition? consider simplifying the code the control flow is easier to follow:

```java
return entries.get(0).equals(entry) && compareAndSet(entries, Collections.<DefaultDnsCacheEntry>emptyList());
```

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
Scottmitch: fine with me ... just may be able to save some memory if its "very likely".

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
normanmaurer: doh! Yeah... let me do what you suggested.

normanmaurer: Actually no... there was something else wrong.

This is correct:

```java
                    if (entries.get(0).equals(entry)) {
                        // If the list is empty we just return early and so not allocate a new ArrayList.
                        if (compareAndSet(entries, Collections.<DefaultDnsCacheEntry>emptyList())) {
                            return false;
                        } else {
                            // try again as the CAS failed
                            continue;
                        }
                    }
                    return false;
```

@Scottmitch WDYT ?

normanmaurer: sorry this was not correct as well. Pushed a new version which should be correct

normanmaurer:
Cherry-picked into 4.1 (e305488c5ab9eb29abfb65ca103c55a4680f4c72)

Issue #7585, opened 2018-01-16
Provide a Docker image for reproducible builds


asarkar:
### Expected behavior
Developer can build on any platform using a Docker image without having to install additional native packages. Simply cloning the repo and following an one-liner build command should be enough.

We should provide a fully-configured Docker image, and also configure the TeamCity CI to use it (I hope it has support for Docker). That way, every build will be reproducible without the million variables that are involved today (OS, OS version, Linux distro, package version, Maven version, etc etc).
FWIW, the Maven version dependency will be eliminated with this PR https://github.com/netty/netty/issues/7578

Sample build command:
```
docker run -it -v /path/to/local/repo:/home/netty/netty netty/netty-build
```

The above command assumes that the default entrypoint to the Docker image is `mvn -f /home/netty/netty clean install`. Of course, the actual path and the build commands can vary, I just made these up.

### Actual behavior
Build is pretty much wild west. For building on Linux, additional native packages need to be installed. Haven't tried on Windows, but I couldn't get it working on a Mac using a Debian VM or Docker image as it kept running OOM. I don't accept having to install a bunch of additional packages on my system for various reasons, security being an obvious one.

### Steps to reproduce
Clone the repo and run `mvn clean install`. Watch it fail.

### Minimal yet complete reproducer code (or URL to code)
N/A

### Netty version
N/A

### JVM version (e.g. `java -version`)
N/A

### OS version (e.g. `uname -a`)
N/A

normanmaurer:
Fixed by https://github.com/netty/netty/pull/7594 ... @asarkar PTAL

normanmaurer:
Fixed by https://github.com/netty/netty/pull/7594

PR #7594, opened 2018-01-18
Provide a Docker image for reproducible builds

normanmaurer:
Motivation:

It would be good to provide a docker image for people that want to build netty on linux.

Modifications:

Add a docker file

Result:

People can more easily build netty. Fixes [#7585].

### DIFF
--- a/docker/Dockerfile-netty-centos6
+++ b/docker/Dockerfile-netty-centos6
@@ -1,0 +1,34 @@
+FROM centos:6
+MAINTAINER netty@googlegroups.com
+ENTRYPOINT /bin/bash
+
+ENV SOURCE_DIR $HOME/source
+ENV MAVEN_VERSION 3.5.2
+ENV JAVA_VERSION 1.8.0
+
+RUN mkdir $SOURCE_DIR
+WORKDIR $SOURCE_DIR
+
+# install dependencies
+RUN yum install -y \
+ apr-devel \
+ autoconf \
+ automake \
+ git \
+ glibc-devel \
+ java-$JAVA_VERSION-openjdk-devel \
+ libtool \ 
+ lksctp-tools \
+ lsb-core \
+ make \
+ openssl-devel \
+ tar \
+ wget
+
+
+RUN wget -q http://archive.apache.org/dist/maven/maven-3/$MAVEN_VERSION/binaries/apache-maven-$MAVEN_VERSION-bin.tar.gz && tar xfz apache-maven-$MAVEN_VERSION-bin.tar.gz && mv apache-maven-$MAVEN_VERSION /opt/
+
+RUN echo 'PATH=/opt/apache-maven-$MAVEN_VERSION/bin:$PATH' >> ~/.bashrc
+RUN echo 'export JAVA_HOME="/usr/lib/jvm/java-$JAVA_VERSION/"' >> ~/.bashrc
+
+RUN rm -rf $SOURCE_DIR
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,15 @@
+
+** Create a docker image **
+```
+docker build -f Dockerfile-netty-centos6 . -t netty-centos6
+```
+
+** Using the image **
+
+```
+cd /path/to/netty/
+```
+
+```
+docker run -it -v ~/.m2:/root/.m2 -v ~/.ssh:/root/.ssh -v ~/.gnupg:/root/.gnupg -v `pwd`:/code -w /code netty-centos6 bash
+```


docker/Dockerfile-netty-centos6
```diff
@@ -0,0 +1,22 @@
...
+
+RUN mkdir $SOURCE_DIR
+WORKDIR $SOURCE_DIR
+
+# install dependencies
+RUN yum install -y lsb-core autoconf automake libtool make tar glibc-devel openssl-devel apr-devel lksctp-tools java-$JAVA_VERSION-openjdk-devel git wget
+
+RUN wget -q http://archive.apache.org/dist/maven/maven-3/$MAVEN_VERSION/binaries/apache-maven-$MAVEN_VERSION-bin.tar.gz
+RUN tar xfz apache-maven-$MAVEN_VERSION-bin.tar.gz
+RUN mv apache-maven-$MAVEN_VERSION /opt/
```
ddossot: Consider merging this into one `RUN ... && \ ...` command to have a single layer created for this group of operation that is logically related.

docker/Dockerfile-netty-centos6
```diff
@@ -0,0 +1,22 @@
...
+
+ENV SOURCE_DIR $HOME/source
+ENV MAVEN_VERSION 3.5.2
+ENV JAVA_VERSION 1.8.0
+
+RUN mkdir $SOURCE_DIR
+WORKDIR $SOURCE_DIR
+
+# install dependencies
+RUN yum install -y lsb-core autoconf automake libtool make tar glibc-devel openssl-devel apr-devel lksctp-tools java-$JAVA_VERSION-openjdk-devel git wget
```
ddossot: Consider sorting and multi-lining for easier maintenance: https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#sort-multi-line-arguments

docker/Dockerfile-netty-centos6
```diff
@@ -0,0 +1,22 @@
+FROM centos:6
+MAINTAINER netty@googlegroups.com
```
asarkar: I believe Google groups need sign up, right? If that's the case, this should be an email that a user of the image can contact without having to sign up for Google groups - all they want to do is report issues.

docker/README.md
```diff
@@ -0,0 +1,15 @@
+
+** Create a docker image **
+```
+docker build -f Dockerfile-netty-centos6 . -t netty-centos6
```
asarkar: This should be published to Docker Cloud, probably under `netty/netty-centos6`. Then CI build changed to run the following line (`docker run`).

docker/Dockerfile-netty-centos6
```diff
@@ -0,0 +1,22 @@
+FROM centos:6
```
asarkar: I suggest starting from OpenJDK base image, so that you don't have to do the install yourself.
https://hub.docker.com/r/library/openjdk/tags/

docker/Dockerfile-netty-centos6
asarkar: I don't see Maven used anywhere, it has to be put on the path. Here's how to do that:
```
curl -sSL http://archive.apache.org/dist/maven/maven-3/$MAVEN_VERSION/binaries/apache-maven-$MAVEN_VERSION-bin.tar.gz | tar xz && \
ln -sf $(pwd)/apache-maven-$MAVEN_VERSION $(pwd)/maven && \
ln -sf $(pwd)/maven/bin/mvn /usr/local/bin/mvn && \
find maven/bin -type f -name "*.cmd" -exec rm -f {} +
```

docker/Dockerfile-netty-centos6
normanmaurer: no you can send an email to it even without sign up.

docker/Dockerfile-netty-centos6
normanmaurer: we need to ensure we use centos6 so the native stuff is build against the correct GLIBC version etc.

asarkar: I see, that's why I couldn't get it working on Debian. This PR is going to make life easier for a lot of people.

docker/Dockerfile-netty-centos6
johnou: It's defined in PATH below.

normanmaurer:
Cherry-picked into 4.1 (949c5153336441c5e06ac822821a59445daed209) and 4.0 (5c8450e215161b1ca6f4025c423fe9688f63ce8d)

PR #7596, opened 2018-01-19
HpackDecoderTest cleanup

Scottmitch:
Motivation:
public ExpectedException fields should be final.
setUp method signature has a throws, but it isn't necessary


normanmaurer:
Cherry-picked into 4.1 (322a062185edb7fd10ec9d820d470d9beb71bcea).

@Scottmitch thanks

PR #7597, opened 2018-01-19
ObjectCleaner may indefinitely block on ReferenceQueue#poll

Scottmitch:
Motivation:
ObjectCleaner polls a ReferenceQueue which will block indefinitely. However it is possible there is a race condition between the live set of objects being empty due to the WeakReference being cleaned/cleared and polling the queue. If this situation occurs the cleanup thread may never unblock if no more objects are added to the live set, and may result in an application's failure to gracefully close.

Modifications:
- ReferenceQueue.remove should use a timeout to compensate for the race condition, and avoid dead lock

Result:
No more dead lock in ObjectCleaner when polling the ReferenceQueue.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/ObjectCleaner.java
+++ b/common/src/main/java/io/netty/util/internal/ObjectCleaner.java
@@ -22,12 +22,16 @@ import java.lang.ref.WeakReference;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import static io.netty.util.internal.SystemPropertyUtil.getInt;
+import static java.lang.Math.max;
+
 /**
  * Allows a way to register some {@link Runnable} that will executed once there are no references to an {@link Object}
  * anymore.
  */
 public final class ObjectCleaner {
-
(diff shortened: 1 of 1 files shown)


common/src/main/java/io/netty/util/internal/ObjectCleaner.java
```diff
@@ -43,7 +47,7 @@ public void run() {
                 while (!LIVE_SET.isEmpty()) {
                     try {
                         AutomaticCleanerReference reference =
-                                (AutomaticCleanerReference) REFERENCE_QUEUE.remove();
+                                (AutomaticCleanerReference) REFERENCE_QUEUE.remove(REFERENCE_QUEUE_POLL_TIMEOUT_MS);
                         try {
                             reference.cleanup();
```
normanmaurer: we need to add a null check as it will return `null` when the timeout hits.

normanmaurer:
Cherry-picked into 4.1 (f72f162e1698226f7e8b1c96736e93e756b859dd) and 4.0 (9466593bbd)

normanmaurer:
@Scottmitch thanks!

PR #7598, opened 2018-01-19
ObjectCleaner should continue cleaning despite exceptions

Scottmitch:
Motivation:
ObjectCleaner inovkes a Runnable which may execute user code (FastThreadLocal#onRemoval) and therefore exceptions maybe thrown. If an exception is thrown the cleanup thread will exit prematurely and we may never finish cleaning up which will result in leaks.

Modifications:
- ObjectCleaner should suppress exceptions and continue cleaning

Result:
ObjectCleaner will reliably clean despite exceptions being thrown.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/ObjectCleaner.java
+++ b/common/src/main/java/io/netty/util/internal/ObjectCleaner.java
@@ -16,6 +16,8 @@
 package io.netty.util.internal;
 
 import io.netty.util.concurrent.FastThreadLocalThread;
+import io.netty.util.internal.logging.InternalLogger;
+import io.netty.util.internal.logging.InternalLoggerFactory;
 
 import java.lang.ref.ReferenceQueue;
 import java.lang.ref.WeakReference;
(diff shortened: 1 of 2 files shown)


Scottmitch:
the first commit is part of https://github.com/netty/netty/pull/7597

common/src/main/java/io/netty/util/internal/ObjectCleaner.java
```diff
@@ -41,17 +47,21 @@ public void run() {
...
                     } catch (InterruptedException ex) {
                         // Just consume and move on
                         interrupted = true;
+                        continue;
+                    }
+                    if (reference != null) {
+                        try {
+                            reference.cleanup();
+                        } catch (Throwable ignored) {
+                            // ignore exceptions, and don't log in case the logger throws an exception.
```
normanmaurer: I think we should try to log and just guard the logger itself. So there will be at least a chance the user see the exception

Scottmitch: We can log but here are potential motivation to not log:
- the JDK also does the same:
https://github.com/netroby/jdk9-dev/blob/master/jdk/src/java.base/share/classes/jdk/internal/ref/CleanerImpl.java#L153
- the logger may also block or have other side effects.

WDYT?

normanmaurer:
Cherry-picked into 4.1 (031bad60dcb70ea52697b09a2c17c62894b92c2b) and 4.0 (2fd6cb0a0f9ddaec03ef4837c06273d845dbb1fb)

johnou:
@normanmaurer shouldn't we break out of the never ending loop after marking the thread interrupted?

johnou:
Nevermind, it is outside of the for loop 🍺

PR #7600, opened 2018-01-20
ByteBuf.toString(Charset) is not thread-safe

thomdev:
Motivation:

Calling `ByteBuf#toString(Charset)` or `ByteBuf#toString(int, int, Charset)` on the same buffer from multiple threads at the same time produces unexpected results, such as various exceptions and/or corrupted output. This is because `ByteBufUtil#decodeString(...)` is taking the source ByteBuffer for `CharsetDecoder#decode()` from `ByteBuf#internalNioBuffer(int, int)`, which is not thread-safe.

Modification:

Call `ByteBuf#nioBuffer()` instead of `ByteBuf#internalNioBuffer()` to get the source buffer to pass to `CharsetDecoder#decode()`.

Result:

Fixes the possible race condition.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
+++ b/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
@@ -682,8 +682,7 @@ public final class ByteBufUtil {
             dst.clear();
         }
         if (src.nioBufferCount() == 1) {
-            // Use internalNioBuffer(...) to reduce object creation.
-            decodeString(decoder, src.internalNioBuffer(readerIndex, len), dst);
+            decodeString(decoder, src.nioBuffer(readerIndex, len), dst);
         } else {
             // We use a heap buffer as CharsetDecoder is most likely able to use a fast-path if src and dst buffers
             // are both backed by a byte array.
--- a/buffer/src/test/java/io/netty/buffer/AbstractByteBufTest.java
+++ b/buffer/src/test/java/io/netty/buffer/AbstractByteBufTest.java
@@ -2083,6 +2083,35 @@ public abstract class AbstractByteBufTest {
     }
 
     @Test
+    public void testToStringMultipleThreads() throws Exception {
+        ByteBuf copied = copiedBuffer("Hello, World!", CharsetUtil.ISO_8859_1);
+        buffer.clear();
+        buffer.writeBytes(copied);
+
+        final CountDownLatch latch = new CountDownLatch(60000);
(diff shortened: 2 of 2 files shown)


slandelle:
@thomdev Could you please explain how you got such race condition? Do you call `ButeBuf#toString` on the same buffer concurrently? If so, how does it make sense?

thomdev:
@slandelle We have a large ByteBuf that stores the content of 500K+ small strings, and an array of offsets. We are calling `ByteBuf#toString(int, int, Charset)` from multiple threads with different offset and size parameters.

slandelle:
Generally speaking, `ByteBuf` is not threadsafe. IMHO, it would be pretty sad if `ByteBuf#toString` were to cause additional an byte copy for the sake of making it threadsafe as it would hurt performance for all the users who don't do concurrent `toString` on the same `ByteBuf` instance.

Why don't you just do `byteBuf.copy(index, length).toString(charset)`?

thomdev:
There is a trade off between efficiency and safety. The current implementation of `ByteBuf.toString(int, int, Charset)` is error prone because the method sounds thread-safe (since it's not supposed to modify the `ByteBuf`) but it is not. BTW, the change I am proposing is slicing the existing `ByteBuffer` but is not doing a string copy. 

I would not be opposed to documenting the thread-safety issue rather than making the code change. I'll let @normanmaurer weigh in.

I agree with your workaround, but it is not obvious why it would be necessary and I am just trying to save the next guy the trouble-shooting pain I went through ;-)

normanmaurer:
@slandelle I think you miss understood the fix... it will not cause an extra memory / Byte copy but just call duplicate() on the ByteBuffer. So it’s just an extra object allocation.

I think @thomdev is correct . It’s a bug

slandelle:
@thomdev @normanmaurer Yes, I agree. For every `ByteBuffer` implementation, it doesn't create any additional copy. BTW, `ByteBuffer#nioBuffer`javadoc is wrong. `The returned buffer shares the content with this buffer` doesn't hold for at least `CompositeByteBuffer`.

normanmaurer:
Cherry-picked into 4.1 (3ae57cf3028e4cfddca141fff723d3189b79fa4c) and 4.0 (0c741eb4c).

@thomdev I slightly adjusted the test-code before cherry-picking to better propagate the exception when it failed.

normanmaurer:
@slandelle https://github.com/netty/netty/pull/7601 PTAL

PR #7601, opened 2018-01-21
Fix ByteBuf.nioBuffer(...) and nioBuffers(...) docs to reflect reality.

normanmaurer:
Motivation:

Depending on the implementation of ByteBuf nioBuffer(...) and nioBuffers(...) may either share the content or return a ByteBuffer that contains a copy of the content.

Modifications:

Fix javadocs.

Result:

Correct docs.


normanmaurer:
@slandelle PTAL

slandelle:
@normanmaurer LGTM

normanmaurer:
Cherry-picked into 4.1 (336bea9dc541413be4cd961417e2be456bdcac11) and 4.0 (0fb1417a4cd0fdf8399dbd27ff53574b0134fa3a).

@slandelle thanks for review

PR #7606, opened 2018-01-22
DefaultChannelPipeline will not invoke handler if events are fired from

Scottmitch:
handlerAdded

Motiviation:
DefaultChannelPipeline and AbstractChannelHandlerContext maintain state
which indicates if a ChannelHandler should be invoked or not. However
the state is updated to allow the handler to be invoked only after the
handlerAdded method completes. If the handlerAdded method generates
events which may result in other methods being invoked on that handler
they will be missed.

Modifications:
- DefaultChannelPipeline should set the state before calling
handlerAdded

Result:
DefaultChannelPipeline will allow events to be processed during the
handlerAdded process.

### DIFF
--- a/transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
+++ b/transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
@@ -611,8 +611,10 @@ public class DefaultChannelPipeline implements ChannelPipeline {
 
     private void callHandlerAdded0(final AbstractChannelHandlerContext ctx) {
         try {
-            ctx.handler().handlerAdded(ctx);
+            // We must call setAddComplete before calling handlerAdded. Otherwise if the handlerAdded method generates
+            // any pipeline events ctx.handler() will miss them because the state will not allow it.
             ctx.setAddComplete();
+            ctx.handler().handlerAdded(ctx);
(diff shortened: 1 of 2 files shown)


transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -611,8 +611,8 @@ private static void checkMultiplicity(ChannelHandler handler) {
 
     private void callHandlerAdded0(final AbstractChannelHandlerContext ctx) {
         try {
-            ctx.handler().handlerAdded(ctx);
             ctx.setAddComplete();
+            ctx.handler().handlerAdded(ctx);
```
normanmaurer: just do me a favour and add a comment that explains the ordering. Otherwise we may forget 👍

Scottmitch: done

normanmaurer:
@Scottmitch please fix check style.

normanmaurer:
Resolved the checkstyle error and cherry-picked into 4.1 (2d815fa75210dfeb578f380be6b71cdddf39d933) and 4.0 (848b2b23d74dc01472c5b270698affb095c0a943)

PR #7608, opened 2018-01-23
Reduce overhead of cancel flowcontrolled writes.

normanmaurer:
Motivation:

When we cancel the flowcontrolled writes we did create a new StreamException for each write that was enqueued. Creating Exceptions is very expensive due of filling the stacktrace.

Modifications:

Only create the StreamException once and reuse the same for all the flowcontrolled writes (per stream).

Result:

Less expensive to cancel flowcontrolled writes.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2RemoteFlowController.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2RemoteFlowController.java
@@ -480,13 +480,15 @@ public class DefaultHttp2RemoteFlowController implements Http2RemoteFlowControll
                 return;
             }
 
-            for (;;) {
-                FlowControlled frame = pendingWriteQueue.poll();
-                if (frame == null) {
-                    break;
-                }
-                writeError(frame, streamError(stream.id(), INTERNAL_ERROR, cause,
-                                              "Stream closed before write could take place"));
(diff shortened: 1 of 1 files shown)


codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2RemoteFlowController.java
```diff
@@ -480,13 +480,15 @@ private void cancel(Throwable cause) {
...
-                FlowControlled frame = pendingWriteQueue.poll();
-                if (frame == null) {
-                    break;
-                }
-                writeError(frame, streamError(stream.id(), INTERNAL_ERROR, cause,
-                                              "Stream closed before write could take place"));
+            FlowControlled frame = pendingWriteQueue.poll();
+            if (frame != null) {
+                // Only create exception once and reuse to reduce overhead of filling in the stacktrace.
+                Http2Exception exception = streamError(stream.id(), INTERNAL_ERROR, cause,
```
trustin: nit: final?

trustin: I always prefer having final local variables because it's good for readability. It is absolutely sure it will never change :-)

normanmaurer:
Merged into 4.1 (b769ec09346e227b9e33d581f26e2cb0b22911a7) as @trustin did approve and its an easy / non-risky change

PR #7610, opened 2018-01-23
Convert Http2Error.STREAM_CLOSED to ClosedChannelException when using…

normanmaurer:
… child channels.

Motivation:

We should convert Http2Exceptions that are produced because of STREAM_CLOSED to ClosedChannelException when hand-over to the child channel to make it more consistent with other transports.

Modifications:

- Check if STREAM_CLOSED is used and if so create a new ClosedChannelException (while preserve the original exception as cause) and use it in the child channel
- Ensure STREAM_CLOSED is used in DefaultHttp2RemoteFlowController when writes are failed because of a closed stream.
- Add testcase

Result:

More consistent and correct exception usage.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2RemoteFlowController.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2RemoteFlowController.java
@@ -27,6 +27,7 @@ import static io.netty.handler.codec.http2.Http2CodecUtil.MAX_WEIGHT;
 import static io.netty.handler.codec.http2.Http2CodecUtil.MIN_WEIGHT;
 import static io.netty.handler.codec.http2.Http2Error.FLOW_CONTROL_ERROR;
 import static io.netty.handler.codec.http2.Http2Error.INTERNAL_ERROR;
+import static io.netty.handler.codec.http2.Http2Error.STREAM_CLOSED;
 import static io.netty.handler.codec.http2.Http2Exception.streamError;
 import static io.netty.handler.codec.http2.Http2Stream.State.HALF_CLOSED_LOCAL;
 import static io.netty.util.internal.ObjectUtil.checkNotNull;
@@ -100,12 +101,12 @@ public class DefaultHttp2RemoteFlowController implements Http2RemoteFlowControll
             public void onStreamClosed(Http2Stream stream) {
                 // Any pending frames can never be written, cancel and
                 // write errors for any pending frames.
-                state(stream).cancel();
+                state(stream).cancel(STREAM_CLOSED, null);
             }
 
             @Override
             public void onStreamHalfClosed(Http2Stream stream) {
-                if (HALF_CLOSED_LOCAL.equals(stream.state())) {
+                if (HALF_CLOSED_LOCAL == stream.state()) {
                     /**
                      * When this method is called there should not be any
                      * pending frames left if the API is used correctly. However,
@@ -117,7 +118,7 @@ public class DefaultHttp2RemoteFlowController implements Http2RemoteFlowControll
                      *
                      * This is to cancel any such illegal writes.
                      */
-                    state(stream).cancel();
+                    state(stream).cancel(STREAM_CLOSED, null);
                 }
             }
         });
@@ -392,7 +393,7 @@ public class DefaultHttp2RemoteFlowController implements Http2RemoteFlowControll
                 // If a cancellation occurred while writing, call cancel again to
                 // clear and error all of the pending writes.
                 if (cancelled) {
-                    cancel(cause);
+                    cancel(INTERNAL_ERROR, cause);
                 }
             }
             return writtenBytes;
@@ -462,17 +463,11 @@ public class DefaultHttp2RemoteFlowController implements Http2RemoteFlowControll
         }
 
         /**
-         * Any operations that may be pending are cleared and the status of these operations is failed.
-         */
-        void cancel() {
-            cancel(null);
-        }
-
-        /**
          * Clears the pending queue and writes errors for each remaining frame.
+         * @param error the {@link Http2Error} to use.
          * @param cause the {@link Throwable} that caused this method to be invoked.
          */
-        private void cancel(Throwable cause) {
+        void cancel(Http2Error error, Throwable cause) {
             cancelled = true;
             // Ensure that the queue can't be modified while we are writing.
             if (writing) {
@@ -482,7 +477,7 @@ public class DefaultHttp2RemoteFlowController implements Http2RemoteFlowControll
             FlowControlled frame = pendingWriteQueue.poll();
             if (frame != null) {
                 // Only create exception once and reuse to reduce overhead of filling in the stacktrace.
-                final Http2Exception exception = streamError(stream.id(), INTERNAL_ERROR, cause,
+                final Http2Exception exception = streamError(stream.id(), error, cause,
                         "Stream closed before write could take place");
                 do {
                     writeError(frame, exception);
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
@@ -429,6 +429,7 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
         // We start with the writability of the channel when creating the StreamChannel.
         private volatile boolean writable;
 
+        private boolean outboundClosed;
         private boolean closePending;
         private boolean readInProgress;
         private Queue<Object> inboundBuffer;
@@ -877,6 +878,7 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
                 }
 
                 // The promise should be notified before we call fireChannelInactive().
+                outboundClosed = true;
                 closePromise.setSuccess();
                 promise.setSuccess();
 
@@ -980,7 +982,9 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
                     return;
                 }
 
-                if (!isActive()) {
+                if (!isActive() ||
+                        // Once the outbound side was closed we should not allow header / data frames
+                        outboundClosed && (msg instanceof Http2HeadersFrame || msg instanceof Http2DataFrame)) {
                     ReferenceCountUtil.release(msg);
                     promise.setFailure(CLOSED_CHANNEL_EXCEPTION);
                     return;
@@ -1046,7 +1050,8 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
                     writabilityChanged(Http2MultiplexCodec.this.isWritable(stream));
                     promise.setSuccess();
                 } else {
-                    promise.setFailure(cause);
+                    promise.setFailure(wrapStreamClosedError(cause));
+                    // If the first write fails there is not much we can do, just close
                     closeForcibly();
                 }
             }
@@ -1056,8 +1061,27 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
                 if (cause == null) {
                     promise.setSuccess();
                 } else {
-                    promise.setFailure(cause);
+                    Throwable error = wrapStreamClosedError(cause);
+                    promise.setFailure(error);
+
+                    if (error instanceof ClosedChannelException) {
+                        if (config.isAutoClose()) {
+                            // Close channel if needed.
+                            closeForcibly();
+                        } else {
+                            outboundClosed = true;
+                        }
+                    }
                 }
+            }
+
+            private Throwable wrapStreamClosedError(Throwable cause) {
+                // If the error was caused by STREAM_CLOSED we should use a ClosedChannelException to better
+                // mimic other transports and make it easier to reason about what exceptions to expect.
+                if (cause instanceof Http2Exception && ((Http2Exception) cause).error() == Http2Error.STREAM_CLOSED) {
+                    return new ClosedChannelException().initCause(cause);
+                }
+                return cause;
             }
 
             private Http2StreamFrame validateStreamFrame(Http2StreamFrame frame) {
--- a/codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2RemoteFlowControllerTest.java
+++ b/codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2RemoteFlowControllerTest.java
@@ -257,8 +257,8 @@ public abstract class DefaultHttp2RemoteFlowControllerTest {
         moreData.assertNotWritten();
 
         connection.stream(STREAM_A).close();
-        data.assertError();
-        moreData.assertError();
+        data.assertError(Http2Error.STREAM_CLOSED);
+        moreData.assertError(Http2Error.STREAM_CLOSED);
         verifyZeroInteractions(listener);
     }
 
@@ -1106,8 +1106,11 @@ public abstract class DefaultHttp2RemoteFlowControllerTest {
             return merged;
         }
 
-        public void assertError() {
+        public void assertError(Http2Error error) {
             assertNotNull(t);
+            if (error != null) {
+                assertSame(error, ((Http2Exception) t).error());
+            }
         }
     }
 }
--- a/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
+++ b/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
@@ -30,13 +30,15 @@ import io.netty.handler.codec.http.HttpScheme;
 import io.netty.handler.codec.http2.Http2Exception.StreamException;
 import io.netty.util.AsciiString;
 import io.netty.util.AttributeKey;
+
+import java.net.InetSocketAddress;
+import java.nio.channels.ClosedChannelException;
+import java.util.concurrent.atomic.AtomicBoolean;
+
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Test;
-
-import java.net.InetSocketAddress;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 import static io.netty.util.ReferenceCountUtil.release;
 import static org.junit.Assert.assertEquals;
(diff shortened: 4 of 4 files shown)


codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2RemoteFlowController.java
```diff
@@ -106,7 +106,7 @@ public void onStreamClosed(Http2Stream stream) {
 
             @Override
             public void onStreamHalfClosed(Http2Stream stream) {
-                if (HALF_CLOSED_LOCAL.equals(stream.state())) {
+                if (HALF_CLOSED_LOCAL == stream.state()) {
```
normanmaurer: I can also revert this change as its not really related but we should do this anyway as its an enum .

Scottmitch: no objection to this change here

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2RemoteFlowController.java
```diff
@@ -480,12 +480,14 @@ private void cancel(Throwable cause) {
                 return;
             }
 
+            // If cause == null this method was called via onStreamClosed(...) or onStreamHalfClosed(...)
+            Http2Error error = cause == null ? Http2Error.STREAM_CLOSED : INTERNAL_ERROR;
```
Scottmitch: `STREAM_CLOSED` has a specific meaning defined by the RFC and its use here may not follow the RFC's definition.

https://tools.ietf.org/html/rfc7540#section-7
> The endpoint received a frame after a stream was half-closed.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -772,6 +772,19 @@ void fireChildReadComplete() {
             }
         }
 
+        void fireChildExceptionCaught(Throwable cause) {
```
ejona86: I don't like this method, or at least its name. It is not obvious it has side effects other than propagating the exception. Either call wrap explicitly or maybe make this `wrapAndFireChildExceptionCaught` or similar.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2RemoteFlowController.java
ejona86: Using the STREAM_CLOSED seems fine to me. If we sent these frames, then the remote endpoint _would_ have "received a frame after the stream was half-closed." But I'm not as wild about the implicit error choice based on the cause. Could we pass that in explicitly instead? Functionally that will be the same.

However, that does expose one problem: when there is half-close and still-pending frames, that should kill the channel.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -772,6 +772,19 @@ void fireChildReadComplete() {
...
+            pipeline().fireExceptionCaught(wrapStreamClosedError(cause));
+        }
+
+        private Throwable wrapStreamClosedError(Throwable cause) {
+            // If the error was caused by STREAM_CLOSED we should use a ClosedChannelException to better
+            // mimic other transports and make it easier to reason about what exceptions to expect.
+            if (cause instanceof Http2Exception && ((Http2Exception) cause).error() == Http2Error.STREAM_CLOSED) {
+                return new ClosedChannelException().initCause(cause);
```
ejona86: So how correct is this, really? I see two potential problems:
1. The channel isn't actually closed: it's only guaranteed to be half-closed and you can still receive. I don't remember how half-closure behavior works in Netty.
2. The channel isn't actually closed: there's not an impending channelInactive.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -772,6 +772,19 @@ void fireChildReadComplete() {
             }
         }
 
+        void fireChildExceptionCaught(Throwable cause) {
+            pipeline().fireExceptionCaught(wrapStreamClosedError(cause));
```
ejona86: It seems wrong to use this wrapping for `fireExceptionCaught`. It doesn't seem "normal" for other Channels to `fireExceptionCaught(ChannelClosedException)`; other channels just fail the futures.

In the case of us receiving a RST_STREAM, I think that should be propagated on the child channel as a userEvent of type Http2ResetFrame. That's already being propagated from the FrameCodec as a read; we should just special-case it in Http2MultiplexCodec.onHttp2Frame. I'm not sure what that means we should do with this exception though, because it seems it _may_ be a duplicate, but that's not guaranteed.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
ejona86: Doh. You can semi-ignore this comment; I forgot to delete it. The other comments likely resolve the strangeness here. But the core idea of "the channel should be closed if we're triggering ClosedChannelException" remains.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2RemoteFlowController.java
normanmaurer: Sure I can also pass it in directly if it makes stuff more clear.

So you say if it’s half closed we should close it ?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
normanmaurer: Sure... just used this naming to be consistent with other names

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
normanmaurer: @ejona86 I think you are actual right and we should just call `channel.close()` here as well. WDYT ?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
normanmaurer: @ejona86 I think this brings us back to another "issue" I am currently working on but still not sure how to fix it in a good way. The problem is that for queued outbound frames (in DefaultRemoteFlowController) we will fail the promise but also call `HttpLifecycleHandler.onError(...)` so there is no easy way to understand if it was an "outbound" operation that failed and so the promise will be failed or if it was an inbound operation.  I wonder if we should call `onError(...)` at all for outbound operations.

Because of this we may call `fireExecptionCaught` with such an exception. That said I would love to fix this. @ejona86 any ideas ?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2RemoteFlowController.java
ejona86: If it's half closed and there are messages pending, then it seems good to close the channel.

normanmaurer: @Scottmitch agree ?

normanmaurer: I guess yeah we should also take AUTO_CLOSE into account here. Thats a good idea. Let me do this

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
normanmaurer: also @Scottmitch ^^

normanmaurer: This may be a good compromise for now https://github.com/netty/netty/pull/7618

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
ejona86: SGTM

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
ejona86: Just coming around to seeing the replies. Probably would have reduced some confusion if I read things in order :-D.

Yeah, I think that #7618 will improve things here.

Note: as far as I'm concerned, the h2childchan API has no place for Http2Exception. It's unneeded and just makes things confusing. We can still use it "internally," but I feel it is counter-productive to expose to users.

normanmaurer: @ejona86 hmm... dont you think they would want to receive it sometimes for better visibility what was wrong ?

ejona86: I didn't say there couldn't be exceptions. But why would those exceptions have an `Http2Error`? Why would they have _(shudders)_ a `ShutdownHint`? What is a user supposed to do with such a thing?

The biggest potential problem I see is keeping some distinction between stream- and channel-level exceptions. I think that may not be a big deal for outbound. It could be for inbound, but I also question exactly how many stream-level exceptions we should have for inbound.

Even if we end up keeping Http2Exception for h2childchan, I would hope it becomes less core to the API.

Scottmitch:
@normanmaurer - Ping me after this PR is rebased.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
Scottmitch: > In the case of us receiving a RST_STREAM, I think that should be propagated on the child channel as a userEvent of type Http2ResetFrame.

I'm not very keen on overloading userEvents to carry actual protocol data. Again userEvent is more for control messages between handlers in the pipeline and not intended for protocol  data. Can we treat this the same way would would treat a "half closed" channel, but it isn't optional for HTTP/2?

> Why would they have (shudders) a ShutdownHint

The user is not generally expected to consume this, and doesn't need to be exposed to the user.

> the h2childchan API has no place for Http2Exception

Agreed. For the child channel API we should try to be a consistent as possible with "regular channels" for exceptions and lifecycle.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
Scottmitch: Continuation of https://github.com/netty/netty/pull/7610#discussion_r163377907 ... I think we need to be careful about half closure (both for inbound and outbound events) instead of unconditionally closing.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
ejona86: > I'm not very keen on overloading userEvents to carry actual protocol data.

We'll totally need to use userEvents for some protocol data. Anything that is important even when you are pushing-back via flow control (imagine disabled AUTO_READ) seems like it must be a userEvent. This is virtually the same as saying that anything that is useful to reorder in front of data needs to be a userEvent. GoAway is one example. ResetStream is another, since some users will want to stop their processing immediately on receiving of the ResetStream and others will want to drain the inbound data. GoAway and ResetStream are both control messages.

Note that if you ignore the userEvents, you're still okay. You'll still see the channel close after you've read all the data, etc.

normanmaurer: @Scottmitch @ejona86 I think we should handle inbound `RST_STREAM` as `IOException("connection reset)"` to mimic other protocols and so also support half-closure. Let me do a pr to show the idea.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2RemoteFlowController.java
normanmaurer: @Scottmitch so I had a deeper look and I think we can detect the difference here. When we receive a RST_FRAME the `Http2FrameListener.onRstStreamRead(...)` is called. So I think it should be ok ?

normanmaurer:
@Scottmitch @ejona86 PTAL again... also I verified that we should never receive and "inbound" exception that we need to convert as in this case we will receive a `Http2ResetFrame` followed by a `streamClosed(...)`.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
normanmaurer: I kept it like this now as there is nothing special going on here anymore (the exception wrapping only happens for outbound).

normanmaurer:
Also @mosesn and @vkostyukov for awareness

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
Scottmitch: >  I think we should handle inbound RST_STREAM as IOException("connection reset)"

Agreed. Again I think we should be as consistent as possible with other channels which seems like `exceptionCaught` in this case. However in this case the user may also want to get the error code from the RST_STREAM frame. Can/should we put this info in the exception?

> anything that is useful to reorder in front of data needs to be a userEvent.

My main concern is we stay consistent with other channels.
- If there is a corresponding userEvent currently used then we should use it (e.g. stream half closed remote maybe ChannelInputShutdownEvent.INSTANCE](https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java#L86)).
- If there is an alternative approach used by other channels we should have a good reason to not follow this convention (e.g. RST_STREAM -> exceptionCaught).
- If we exhaust the above options and it makes sense to add a userEvent then lets do it.

I'm not yet convinced that RST_STREAM and GO_AWAY (see https://github.com/netty/netty/pull/7635) necessarily have to be user events.

normanmaurer:
Same here... I think these should not be user events

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
normanmaurer: Yeah I think we could just create a custom IOException that hold it... i would like to to do this as a follow up tho

Scottmitch: followup sgtm!

I would like to get on the same page as @ejona86 too :)

normanmaurer: Same here... I think these should not be user events

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
Scottmitch: My vote is to remove the method for now. We can add it back later if necessary.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -870,7 +875,8 @@ public void close(ChannelPromise promise) {
 
                     // Only ever send a reset frame if the connection is still alive as otherwise it makes no sense at
                     // all anyway.
-                    if (parent().isActive() && !streamClosedWithoutError && isStreamIdValid(stream().id())) {
+                    if (!outboundClosed && parent().isActive() &&
```
Scottmitch: IIUC `outboundClosed` maybe set to `true` as a result of detecting the connection is in `HALF_CLOSED_REMOTE` in the flow controller (via `writeComplete`). If this is the case we still may want to send a `RST_STREAM` if we have not yet closed the local side of the stream.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -991,7 +998,7 @@ public void write(Object msg, final ChannelPromise promise) {
                     return;
                 }
 
-                if (!isActive()) {
+                if (!isActive() || outboundClosed) {
```
Scottmitch: Same comment as above. We may not want to write "headers"/"data", but other control type frames may still be valid based upon the current usage of `outboundClosed`.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
normanmaurer: sure its not doing anything except delegating anyway atm

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
normanmaurer: @Scottmitch hm... maybe we should just not care about `outboundClosed` here

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
normanmaurer: yeah let us just fail headers / data to be inline with the RFC.

normanmaurer:
@Scottmitch hopefully I addressed your concerns / comments

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
ejona86: > I think we should handle inbound RST_STREAM as IOException("connection reset)" to mimic other protocols and so also support half-closure. Let me do a pr to show the idea.

I was thinking we do something like that as well (if the error code is something other than CANCEL, probably want to include it, though), and that's what most people would use. But receiving RST_STREAM(CANCEL) while pushing-back on flow control seems important for use cases where flow control is likely to be applied, as the notice can arrive much sooner and you can know to abort work without asking for more work.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -870,7 +871,8 @@ public void close(ChannelPromise promise) {
 
                     // Only ever send a reset frame if the connection is still alive as otherwise it makes no sense at
                     // all anyway.
-                    if (parent().isActive() && !streamClosedWithoutError && isStreamIdValid(stream().id())) {
+                    if (parent().isActive() &&
```
Scottmitch: revert if there is not change?

normanmaurer:
@ejona86 happy as well?

PR #7615, opened 2018-01-23
Replace reflective access of Throwable#addSuppressed with version guarded access

### LINKED ISSUE #7614: ThrowableUtil should work without needing reflection permissions
### Expected behavior
Netty works with a security manager without needing "java.lang.RuntimePermission" "accessDeclaredMembers". The `ThrowableUtil` class should be able to function without the `addSuppressed` method being available.

### Actual behavior
In a environment with a security manager, Netty initialization fails due to the use of [`getDeclaredMethod`](https://github.com/netty/netty/blob/netty-4.1.20.Final/common/src/main/java/io/netty/util/internal/ThrowableUtil.java#L35) in `ThrowableUtil`.

```
Caused by: java.security.AccessControlException: access denied ("java.lang.RuntimePermission" "accessDeclaredMembers")
    at java.base/java.security.AccessControlContext.checkPermission(AccessControlContext.java:472)
    at java.base/java.security.AccessController.checkPermission(AccessController.java:895)
    at java.base/java.lang.SecurityManager.checkPermission(SecurityManager.java:558)
    at java.base/java.lang.Class.checkMemberAccess(Class.java:2804)
    at java.base/java.lang.Class.getDeclaredMethod(Class.java:2428)
    at io.netty.util.internal.ThrowableUtil.getAddSuppressed(ThrowableUtil.java:35)
    at io.netty.util.internal.ThrowableUtil.<clinit>(ThrowableUtil.java:27)
```

### Steps to reproduce
Run Netty with a security manager that does not grant the "accessDeclaredMembers" permission.

### Reproducer
This is not minimal but see the code https://github.com/elastic/elasticsearch/pull/28345. Remove the accessDeclaredMembers permission and run `./gradlew :modules:transport-netty4:check`.

### Netty version
4.1.18.Final in the reproducer but confirmed with 4.1.20.Final as well.

```
$ java -version
java version "9.0.4"
Java(TM) SE Runtime Environment (build 9.0.4+11)
Java HotSpot(TM) 64-Bit Server VM (build 9.0.4+11, mixed mode)
```

Linux build 4.13.0-25-generic #29-Ubuntu SMP Mon Jan 8 21:14:41 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

jaymode:
Motivation:

In environments with a security manager, the reflective access to get the reference to
Throwable#addSuppressed can cause issues that result in Netty failing to load. The main
motivation in this pull request is to remove the use of reflection to prevent issues in
these environments.

Modifications:

ThrowableUtil no longer uses Class#getDeclaredMembers to get the Method that references
Throwable#addSuppressed and instead guards the call to Throwable#addSuppressed with a
Java version check.

Additionally, a annotation was added that suppresses the animal sniffer java16 signature
check on the given method. The benefit of the annotation is that it limits the exclusion
of Throwable to just the ThrowableUtil class and has string text indicating the reason
for suppressing the java16 signature check.

Result:

Netty no longer requires the use of Class#getDeclaredMethod for ThrowableUtil and will
work in environments restricted by a security manager without needing to grant reflection
permissions.

Fixes #7614

### DIFF
--- a/common/src/main/java/io/netty/util/SuppressJava6Requirement.java
+++ b/common/src/main/java/io/netty/util/SuppressJava6Requirement.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.netty.util;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation to suppress the Java 6 source code requirement checks for a method.
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target({ ElementType.METHOD })
+public @interface SuppressJava6Requirement {
+
+    String reason();
+}
--- a/common/src/test/java/io/netty/util/internal/NativeLibraryLoaderTest.java
+++ b/common/src/test/java/io/netty/util/internal/NativeLibraryLoaderTest.java
@@ -15,7 +15,6 @@
  */
 package io.netty.util.internal;
 
-import io.netty.util.SuppressJava6Requirement;
 import org.junit.Test;
 
 import java.io.FileNotFoundException;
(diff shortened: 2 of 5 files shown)


common/src/test/java/io/netty/util/internal/NativeLibraryLoaderTest.java
```diff
@@ -48,34 +46,28 @@ public void testFileNotFoundWithNullClassLoader() {
...
+            if (PlatformDependent.javaVersion() >= 7) {
                 verifySuppressedException(error, ClassNotFoundException.class);
             }
         }
     }
 
+    @SuppressJava6Requirement(reason = "uses Java 7+ Throwable#getSuppressed but is guarded by version checks")
     private static void verifySuppressedException(UnsatisfiedLinkError error,
             Class<?> expectedSuppressedExceptionClass) {
+        if (PlatformDependent.javaVersion() < 7) {
```
normanmaurer: I think we can remove this... If you really want to keep it make it an `assert`.

common/src/main/java/io/netty/util/SuppressJava6Requirement.java
```diff
@@ -0,0 +1,32 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation to suppress the Java 6 source code requirement checks for a method.
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target({ ElementType.METHOD })
+public @interface SuppressJava6Requirement {
```
normanmaurer: @jaymode can we move this to "internal" package ?

pom.xml
```diff
@@ -721,6 +721,9 @@
             <ignore>java.nio.file.Path</ignore>
             <ignore>java.io.File</ignore>
           </ignores>
+          <annotations>
+            <annotation>io.netty.util.SuppressJava6Requirement</annotation>
```
normanmaurer: @jaymode you also need to adjust his one as otherwise the build will fail:

https://garage.netty.io/teamcity/viewLog.html?buildId=27812&buildTypeId=pulls_netty&guest=1 

;)

normanmaurer:
Squashed and cherry-picked into 4.1 (f0c76cacc3f6eb3a7df3fdc06059d5ddaac3c442) and 4.0 (307229fa85f273b857a42a05f6e4a6f2818c2bc4)

PR #7618, opened 2018-01-24
Do not fire outbound exception throught the pipeline when using Http2…

normanmaurer:
…FrameCodec / Http2MultiplexCodec

Motivation:

Usually when using netty exceptions which happen for outbound operations should not be fired through the pipeline but only the ChannelPromise should be failed.

Modifications:

- Change Http2LifecycleManager.onError(...) to take also an boolean that indicate if the error was caused by an outbound operation
- Channel Http2ConnectionHandler.on*Error(...) methods to also take this boolean
- Change Http2FrameCodec to only fire exceptions through the pipeline if these are not outbound operations related
- Add unit test.

Result:

More consistent error handling when using Http2FrameCodec and Http2MultiplexCodec.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
@@ -211,7 +211,7 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
                     // other headers containing pseudo-header fields.
                     stream.headersSent(isInformational);
                 } else {
-                    lifecycleManager.onError(ctx, failureCause);
+                    lifecycleManager.onError(ctx, true, failureCause);
                 }
 
                 return future;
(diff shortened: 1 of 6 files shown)


normanmaurer:
also @mosesn and @buchgr

normanmaurer:
and maybe @vietj / @jchambers

jchambers:
Haven't done a deep analysis, but this seems completely reasonable to me in principle!

normanmaurer:
@ejona86 yeah I agree... Its a bit "messy" atm as we depend on have onError called for trying to send a RST for example. All of this is a lot easier when using the `Http2FrameCodec` or `Http2MultiplexCodec` as here it would be easy to just attach a ChannelFutureListener that would do the needed work. This is not really possible easily when using the "old API" as the user may just call methods on the encoder directly and we have no good place to keep all the state etc.  So with this in mind I think its more a step in the right direction then the final solution (which hopefully at the end will be to not have the "old API" at all. WDYT ?

normanmaurer:
So just to make it clear this is more a "step" in the right direction then an ultimate fix for long-term.

ejona86:
I think we're on the same page. I'm also fine with this for now and think it does help resolve lots of confusion without putting current implementations in a huge pinch if they are relying on the current behavior. The other stuff I was talking about is long-term.

PR #7619, opened 2018-01-24
Ensure async failures are correctly propagated to Http2LifecycleManag…

normanmaurer:
…er.onError(...) in all cases.

Motivation:

If DefaultHttp2ConnectionEncoder process outbound operation it sometimes missed to call Http2LifecycleManager.onError(...) when the operation was executed asynchronously.

Modifications:

Make best effort to update flags but still ensure failures are propageted to Http2LifecycleManager.onError(...) in all cases.

Result:

More consistent handling of errors.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
@@ -209,7 +209,23 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
                 if (failureCause == null) {
                     // Synchronously set the headersSent flag to ensure that we do not subsequently write
                     // other headers containing pseudo-header fields.
+                    //
+                    // This is best-afford as the operation is async it may still fail later on. We just update it here
+                    // directly as otherwise it may be challenging for the end-user to co-ordinate writes that depend
+                    // on the state.
                     stream.headersSent(isInformational);
+
+                    if (!future.isDone()) {
+                        future.addListener(new ChannelFutureListener() {
+                            @Override
+                            public void operationComplete(ChannelFuture future) throws Exception {
+                                Throwable cause = future.cause();
+                                if (cause != null) {
+                                    lifecycleManager.onError(ctx, cause);
+                                }
+                            }
+                        });
+                    }
                 } else {
                     lifecycleManager.onError(ctx, failureCause);
                 }
@@ -269,7 +285,7 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
     }
 
     @Override
-    public ChannelFuture writePushPromise(ChannelHandlerContext ctx, int streamId, int promisedStreamId,
+    public ChannelFuture writePushPromise(final ChannelHandlerContext ctx, int streamId, int promisedStreamId,
             Http2Headers headers, int padding, ChannelPromise promise) {
         try {
             if (connection.goAwayReceived()) {
@@ -285,7 +301,22 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
             // Writing headers may fail during the encode state if they violate HPACK limits.
             Throwable failureCause = future.cause();
             if (failureCause == null) {
+                // This is best-afford as the operation is async it may still fail later on. We just update it here
+                // directly as otherwise it may be challenging for the end-user to co-ordinate writes that depend
+                // on the state.
                 stream.pushPromiseSent();
+
+                if (!future.isDone()) {
+                    future.addListener(new ChannelFutureListener() {
+                        @Override
+                        public void operationComplete(ChannelFuture future) throws Exception {
+                            Throwable cause = future.cause();
+                            if (cause != null) {
+                                lifecycleManager.onError(ctx, cause);
+                            }
+                        }
+                    });
+                }
             } else {
                 lifecycleManager.onError(ctx, failureCause);
             }
@@ -462,7 +493,7 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
         }
 
         @Override
-        public void write(ChannelHandlerContext ctx, int allowedBytes) {
+        public void write(final ChannelHandlerContext ctx, int allowedBytes) {
             boolean isInformational = validateHeadersSentState(stream, headers, connection.isServer(), endOfStream);
             if (promise.isVoid()) {
                 promise = ctx.newPromise();
@@ -472,9 +503,24 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
             ChannelFuture f = frameWriter.writeHeaders(ctx, stream.id(), headers, streamDependency, weight, exclusive,
                                                        padding, endOfStream, promise);
             // Writing headers may fail during the encode state if they violate HPACK limits.
-            Throwable failureCause = f.cause();
+            final Throwable failureCause = f.cause();
             if (failureCause == null) {
+                // This is best-afford as the operation is async it may still fail later on. We just update it here
+                // directly as otherwise it may be challenging for the end-user to co-ordinate writes that depend
+                // on the state.
                 stream.headersSent(isInformational);
+
+                if (!f.isDone()) {
+                    f.addListener(new ChannelFutureListener() {
+                        @Override
+                        public void operationComplete(ChannelFuture future) throws Exception {
+                            Throwable cause = future.cause();
+                            if (cause != null) {
+                                lifecycleManager.onError(ctx, cause);
+                            }
+                        }
+                    });
+                }
             } else {
                 lifecycleManager.onError(ctx, failureCause);
             }


codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -285,7 +301,22 @@ public ChannelFuture writePushPromise(ChannelHandlerContext ctx, int streamId, i
             // Writing headers may fail during the encode state if they violate HPACK limits.
             Throwable failureCause = future.cause();
             if (failureCause == null) {
+                // This is best-afford as the operation is async it may still fail later on. We just update it here
+                // directly as otherwise it may be challenging for the end-user to co-ordinate writes that depend
+                // on the state.
                 stream.pushPromiseSent();
+
+                if (!future.isDone()) {
+                    future.addListener(new ChannelFutureListener() {
```
trustin: This entire `if` block could be extracted into a method for reduced duplication?

normanmaurer: sure let me do

normanmaurer: done

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -209,7 +209,15 @@ public void operationComplete(ChannelFuture future) throws Exception {
                 if (failureCause == null) {
                     // Synchronously set the headersSent flag to ensure that we do not subsequently write
                     // other headers containing pseudo-header fields.
+                    //
+                    // This is best-afford as the operation is async it may still fail later on. We just update it here
+                    // directly as otherwise it may be challenging for the end-user to co-ordinate writes that depend
+                    // on the state.
                     stream.headersSent(isInformational);
+
+                    if (!future.isDone()) {
```
trustin: Hmm, perhaps you can move this check into `notifyLifecycleManagerOnError()`?

trustin:
Off topic: It's not lifecycle but life cycle :-)

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
normanmaurer: sure

normanmaurer: done

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -209,7 +209,13 @@ public void operationComplete(ChannelFuture future) throws Exception {
                 if (failureCause == null) {
                     // Synchronously set the headersSent flag to ensure that we do not subsequently write
                     // other headers containing pseudo-header fields.
+                    //
+                    // This is best-afford as the operation is async it may still fail later on. We just update it here
```
ejona86: s/afford/effort/

Ditto elsewhere.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -428,6 +439,20 @@ public boolean merge(ChannelHandlerContext ctx, Http2RemoteFlowController.FlowCo
         }
     }
 
+    private void notifyLifecycleManagerOnError(ChannelFuture future, final ChannelHandlerContext ctx) {
+        if (!future.isDone()) {
```
ejona86: I'd feel better if this was an `assert !future.isDone()`, as if the future is done this method is being misused (`lifecycleManager.onError()` won't be called). But I guess that isn't strictly thread-safe :(, even though practically it probably would be safe.

normanmaurer: Actually no... its completely fine if `future.isDone()` is true  and this is called. I should just not be `successful` ;)

ejona86: How about you just remove this if then (add the listener unconditionally)? That is fully safe and only adds work if there was a race.

normanmaurer: @ejona86 maybe we misunderstand each other... so here is the calling code:

```java
            if (failureCause == null) {
                // This is best-effort as the operation is async it may still fail later on. We just update it here
                // directly as otherwise it may be challenging for the end-user to co-ordinate writes that depend
                // on the state.
                stream.pushPromiseSent();

                notifyLifecycleManagerOnError(future, ctx);
            } else {
                lifecycleManager.onError(ctx, failureCause);
            }
```

So if I remove the `!isDone()` check it will always add the listener even if the future was done but successfully. I wanted to avoid this...

ejona86: Ah, you're right. Then maybe the condition should be `if (!(future.isDone() && future.cause() == null))` as the check instead? I'd also accept renaming the method.

Right now there's tight coupling that's not easily visible between the caller and callee; you have to read both together to guarantee correctness. I want it to be more obvious "you are holding it wrong" (via rename) or for it to become fool-proof (the fully-precise check).

normanmaurer:
I would go for renaming if you have a better name ;)

ejona86:
I don't have any good ideas for a rename. But I think that's the same as my existing argument. (I decided `ifFutureNotDoneNotifyLifecycleManagerOnError` is not a good name. As well as `onlyUseThisIfYouKnowWhatItDoes`.)

There's a guiding principal I have. If I can't write good and concise Javadoc for a method, then maybe I should change the method. It applies equally well to method names.

As yet another option, you could hoist the `isDone()` check out of the method. Thus, instead of:
```java
            if (failureCause == null) {
                // This is best-effort as the operation is async it may still fail later on. We just update it here
                // directly as otherwise it may be challenging for the end-user to co-ordinate writes that depend
                // on the state.
                stream.pushPromiseSent();

                notifyLifecycleManagerOnError(future, ctx);
            } else {
                lifecycleManager.onError(ctx, failureCause);
            }
```

You'd have:
```java
            if (failureCause == null) {
                // This is best-effort as the operation is async it may still fail later on. We just update it here
                // directly as otherwise it may be challenging for the end-user to co-ordinate writes that depend
                // on the state.
                stream.pushPromiseSent();

                if (!future.isDone()) {
                    notifyLifecycleManagerOnError(future, ctx);
                }
            } else {
                lifecycleManager.onError(ctx, failureCause);
            }
```

normanmaurer:
@ejona86 thats what I had before and @trustin suggested to move the check into the method... That said it was more an suggestion then a hard-requirement so let me change it back if it makes things easier to parse / read for you.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -209,7 +209,16 @@ public void operationComplete(ChannelFuture future) throws Exception {
                 if (failureCause == null) {
                     // Synchronously set the headersSent flag to ensure that we do not subsequently write
                     // other headers containing pseudo-header fields.
+                    //
+                    // This is best-effort as the operation is async it may still fail later on. We just update it here
+                    // directly as otherwise it may be challenging for the end-user to co-ordinate writes that depend
```
Scottmitch: this isn't really for end users to consume, but is more for us to consume internally and ensure we correctly order and enforce protocol restrictions.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -474,7 +503,15 @@ public void write(ChannelHandlerContext ctx, int allowedBytes) {
             // Writing headers may fail during the encode state if they violate HPACK limits.
             Throwable failureCause = f.cause();
             if (failureCause == null) {
+                // This is best-effort as the operation is async it may still fail later on. We just update it here
+                // directly as otherwise it may be challenging for the end-user to co-ordinate writes that depend
+                // on the state.
                 stream.headersSent(isInformational);
+
+                if (!f.isSuccess()) {
```
Scottmitch: can you include a unit test which exercises this code path? I'm not sure this change is necessary.

`FlowControlledHeaders` adds itself as listener to `promise` before the `write*` operation. When the future completes `FlowControlledBase#operationComplete` will be called, if the future has failed `error(..)` will be called, which should call `lifecycleManager.onError(ctx, cause)`. If anything maybe we should remove the `else` condition below which also calls `lifecycleManager.onError(ctx, cause)`?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
normanmaurer: let me reword...

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
normanmaurer: @Scottmitch good catch. Let me remove it for the reason you stated.

normanmaurer:
@Scottmitch PTAL again

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -285,7 +292,14 @@ public ChannelFuture writePushPromise(ChannelHandlerContext ctx, int streamId, i
             // Writing headers may fail during the encode state if they violate HPACK limits.
             Throwable failureCause = future.cause();
             if (failureCause == null) {
+                // This is best-effort as the operation is async it may still fail later on.
+
```
Scottmitch: kill line

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -209,7 +209,14 @@ public void operationComplete(ChannelFuture future) throws Exception {
                 if (failureCause == null) {
                     // Synchronously set the headersSent flag to ensure that we do not subsequently write
                     // other headers containing pseudo-header fields.
+                    //
+                    // This is best-effort as the operation is async it may still fail later on.
```
Scottmitch: instead of `best-effort` maybe something like `this just sets internal stream state which is used elsewhere in the codec and doesn't necessarily mean the write will complete successfully` is more clear?

normanmaurer: works for me

PR #7623, opened 2018-01-25
Add method to wait for ObjectCleaner thread termination

jaymode:
Motivation:

In some applications, it is desireable to wait for all threads to finish executing especially those
that do not respond to InterruptedException immediately. The ObjectCleaner class runs a thread that
executes in the background and defers responding to a InterruptedException until all processing of
the cleanup tasks has finished.

Modifications:

Add a static volatile variable to track the current reference to the cleaner thread and add the
awaitInactivity method to ObjectCleaner. This method allows for the caller to wait for the
termination of the current thread.

Result:

Callers of awaitInactivity can wait for pending cleanups to be finished assuming their application
is shut down and ObjectCleaner#register is not called again.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/ObjectCleaner.java
+++ b/common/src/main/java/io/netty/util/internal/ObjectCleaner.java
@@ -16,12 +16,11 @@
 package io.netty.util.internal;
 
 import io.netty.util.concurrent.FastThreadLocalThread;
-import io.netty.util.internal.logging.InternalLogger;
-import io.netty.util.internal.logging.InternalLoggerFactory;
 
 import java.lang.ref.ReferenceQueue;
 import java.lang.ref.WeakReference;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import static io.netty.util.internal.SystemPropertyUtil.getInt;
@@ -72,6 +71,7 @@ public final class ObjectCleaner {
                 if (LIVE_SET.isEmpty() || !CLEANER_RUNNING.compareAndSet(false, true)) {
                     // There was nothing added after we set STARTED to false or some other cleanup Thread
                     // was started already so its safe to let this Thread complete now.
+                    ObjectCleaner.objectCleanerThread = null;
                     break;
                 }
             }
@@ -81,6 +81,8 @@ public final class ObjectCleaner {
             }
         }
     };
+
+    private static volatile Thread objectCleanerThread;
 
     /**
      * Register the given {@link Object} for which the {@link Runnable} will be executed once there are no references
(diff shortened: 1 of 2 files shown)


common/src/main/java/io/netty/util/internal/ObjectCleaner.java
```diff
@@ -72,6 +71,7 @@ public void run() {
                 if (LIVE_SET.isEmpty() || !CLEANER_RUNNING.compareAndSet(false, true)) {
                     // There was nothing added after we set STARTED to false or some other cleanup Thread
                     // was started already so its safe to let this Thread complete now.
+                    ObjectCleaner.objectCleanerThread = null;
```
Scottmitch: This may race with the thread setting the variable in the register method below.

common/src/main/java/io/netty/util/internal/ObjectCleaner.java
```diff
@@ -98,20 +100,56 @@ public static void register(Object object, Runnable cleanupTask) {
...
+        }
+
+        if (CLEANER_RUNNING.get()) {
+            Thread objectCleanerThread;
+            // loop here because there is a race between the setting of the atomic boolean and the
+            // setting of the objectCleanerThread
+            for (;;) {
+                objectCleanerThread = ObjectCleaner.objectCleanerThread;
+                if (objectCleanerThread != null) {
+                    break;
```
Scottmitch: As you indicated in your comments this is racy and may not reliably await depending on the circumstances.

common/src/main/java/io/netty/util/internal/ObjectCleaner.java
```diff
@@ -98,20 +100,56 @@ public static void register(Object object, Runnable cleanupTask) {
...
+     * Waits until the current cleaner thread has completed or the timeout has elapsed.
+     *
+     * Note: a new thread will be started if one is not running on a call to
+     * {@link #register(Object, Runnable)}, so this method is really only useful
+     * <strong>after</strong> your application has shut down and you want to wait for this
+     * thread to terminate.
+     *
+     * @return {@code true} if the cleaner thread has been terminated
+     */
+    public static boolean awaitInactivity(long timeout, TimeUnit unit) throws InterruptedException {
```
Scottmitch: This class is in an internal package and therefore not part of the Netty public api. The PR justification says it maybe useful to wait until all threads are complete. Can you provide more context here? The thread is currently not a deamon so the application should put terminate until the thread exits.

jaymode: I agree that my reasoning for this was not clear and more context is needed. In the elasticsearch tests, we shutdown a node, which includes stopping the Netty bootstrap(s). After this has been completed, our test framework checks for leaked threads and if there are some it waits for them, if the wait times out then it tries to interrupt all of the leaked threads, then waits again, and finally fails if the thread is still around.

Example (Netty 4.1.20.Final, JDK 9.04, Ubuntu 17.10):

```
  2> Jan 22, 2018 4:12:19 PM com.carrotsearch.randomizedtesting.ThreadLeakControl checkThreadLeaks
  2> WARNING: Will linger awaiting termination of 2 leaked thread(s).
  2> Jan 22, 2018 4:12:24 PM com.carrotsearch.randomizedtesting.ThreadLeakControl checkThreadLeaks
  2> SEVERE: 1 thread leaked from SUITE scope 
  2>    1) Thread[id=32, name=ObjectCleanerThread, state=TIMED_WAITING, group=TGRP-SSLClientAuthTests]
  2>         at java.base@9.0.4/java.lang.Object.wait(Native Method)
  2>         at java.base@9.0.4/java.lang.ref.ReferenceQueue.remove(ReferenceQueue.java:151)
  2>         at app//io.netty.util.internal.ObjectCleaner$1.run(ObjectCleaner.java:52)
  2>         at app//io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
  2>         at java.base@9.0.4/java.lang.Thread.run(Thread.java:844)
  2> Jan 22, 2018 4:12:24 PM com.carrotsearch.randomizedtesting.ThreadLeakControl tryToInterruptAll
  2> INFO: Starting to interrupt leaked threads:
  2>    1) Thread[id=32, name=ObjectCleanerThread, state=TIMED_WAITING, group=TGRP-SSLClientAuthTests]
  2> Jan 22, 2018 4:12:27 PM com.carrotsearch.randomizedtesting.ThreadLeakControl tryToInterruptAll
  2> SEVERE: There are still zombie threads that couldn't be terminated:
  2>    1) Thread[id=32, name=ObjectCleanerThread, state=TIMED_WAITING, group=TGRP-SSLClientAuthTests]
  2>         at java.base@9.0.4/java.lang.Object.wait(Native Method)
  2>         at java.base@9.0.4/java.lang.ref.ReferenceQueue.remove(ReferenceQueue.java:151)
  2>         at app//io.netty.util.internal.ObjectCleaner$1.run(ObjectCleaner.java:52)
  2>         at app//io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
  2>         at java.base@9.0.4/java.lang.Thread.run(Thread.java:844)
```

In the same JVM more tests may be run after, which is why I'd like to see this thread stop. We can add a `ThreadLeakFilter` to allow this thread to "leak" and not fail but I don't want to pollute the next test.

I used ThreadDeathWatcher#awaitInactivity as precedent for this method but did not pay attention to the fact that the class is in the internal package.

I'm interested to hear what ideas you or @normanmaurer might have.

jaymode:
I'm going to close this PR as I have found it doesn't actually help in my situation and #7627 will not either as the JVM itself is not terminating but Netty has been in the test. If the interrupt of the thread actually terminated the running, then I believe I wouldn't have this issue.

PR #7624, opened 2018-01-25
Add java-doc for implemented methods of io.netty.util.concurrent.Future#cancel(boolean mayInterruptIfRunning) 

### LINKED ISSUE #7609: should add java-doc for method DefaultPromise#cancel(boolean mayInterruptIfRunning)
### Expected behavior
implementation of method `io.netty.util.concurrent.DefaultPromise#cancel(boolean mayInterruptIfRunning)` actually ignored the param `mayInterruptIfRunning`.
so it need to make a comment like following:
(this comment is copied from `java.util.concurrent.CompletableFuture#cancel`):

>      * @param mayInterruptIfRunning this value has no effect in this
>      * implementation because interrupts are not used to control
>      * processing.
the other method `io.netty.util.concurrent.CompleteFuture#cancel` may do the same too.
### Actual behavior
there is no comment about the param `mayInterruptIfRunning`, which will make user confusing when call the `cancel` method with a param `mayInterruptIfRunning`.
### Netty version
4

laosijikaichele:
Motivation:

The methods implement `io.netty.util.concurrent.Future#cancel(boolean mayInterruptIfRunning)` which actually ignored the param mayInterruptIfRunning.We need to add comments for the `mayInterruptIfRunning` param.

Modifications:

Add comments for the `mayInterruptIfRunning` param.

Result:

People who call the `cancel` method will be more clear about the effect of `mayInterruptIfRunning` param.
Fixes #7609

### DIFF
--- a/common/src/main/java/io/netty/util/concurrent/CompleteFuture.java
+++ b/common/src/main/java/io/netty/util/concurrent/CompleteFuture.java
@@ -140,6 +140,9 @@ public abstract class CompleteFuture<V> extends AbstractFuture<V> {
         return false;
     }
 
+    /**
+     * @param mayInterruptIfRunning this value has no effect in this implementation.
+     */
     @Override
     public boolean cancel(boolean mayInterruptIfRunning) {
         return false;
--- a/common/src/main/java/io/netty/util/concurrent/CompleteFuture.java
+++ b/common/src/main/java/io/netty/util/concurrent/CompleteFuture.java
@@ -141,6 +141,8 @@ public abstract class CompleteFuture<V> extends AbstractFuture<V> {
     }
 
     /**
+     * {@inheritDoc}
+     *
      * @param mayInterruptIfRunning this value has no effect in this implementation.
      */
     @Override
--- a/transport/src/main/java/io/netty/channel/DelegatingChannelPromiseNotifier.java
+++ b/transport/src/main/java/io/netty/channel/DelegatingChannelPromiseNotifier.java
@@ -172,11 +172,6 @@ public final class DelegatingChannelPromiseNotifier implements ChannelPromise, C
         return delegate.getNow();
     }
 
-    /**
-     * {@inheritDoc}
-     *
-     * @param mayInterruptIfRunning this value has no effect in this implementation.
-     */
     @Override
     public boolean cancel(boolean mayInterruptIfRunning) {
         return delegate.cancel(mayInterruptIfRunning);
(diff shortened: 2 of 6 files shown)


common/src/main/java/io/netty/util/concurrent/CompleteFuture.java
```diff
@@ -140,6 +140,9 @@ public boolean isCancelled() {
         return false;
     }
 
+    /**
+     * @param mayInterruptIfRunning this value has no effect in this implementation.
```
Scottmitch: at a minimum we need {@inheritDoc}, and in general I'm not convinced this change is necessary as the `cancel` operation itself is indicated as best effort in the original javadocs and therefore the parameter may not be used in general.

laosijikaichele: if a param value does not used in the method, then it should be indicated by the javadoc, but the original javadoc from `java.util.concurrent.Future#cancel` does not clearly indicate it, so the JVM team itself indicate it when the implementation method does not use the param `mayInterruptIfRunning`, please refer the javadoc of `java.util.concurrent.CompletableFuture#cancel` and `java.util.concurrent.ForkJoinTask#cancel` ;
 which javadocs clearly indicate that `mayInterruptIfRunning` has no effect in the implementation.

transport/src/main/java/io/netty/channel/DelegatingChannelPromiseNotifier.java
```diff
@@ -172,6 +172,11 @@ public Void getNow() {
         return delegate.getNow();
     }
 
+    /**
+     * {@inheritDoc}
+     *
+     * @param mayInterruptIfRunning this value has no effect in this implementation.
```
normanmaurer: @laosijikaichele actually I think this is not correct as it just delegates it and if it is ignored or not depends on the implementation of `delegate`

laosijikaichele: @normanmaurer  agree, just deleted comment for this method.

normanmaurer:
@laosijikaichele please sign our ICLA: https://netty.io/s/icla

laosijikaichele:
@normanmaurer  signed done.

normanmaurer:
@Scottmitch fine with you?

normanmaurer:
Squashed and cherry-picked into 4.1 (9dd5c928f3ebf4b3fd8b6cf875d7b170a3a12a48) and 4.0 (16239d4ea3)

normanmaurer:
@laosijikaichele thanks!

PR #7626, opened 2018-01-25
Reduce object allocation by using same ChannelFutureListener instance.

normanmaurer:
Motivation:

When VoidChannelPromise.unvoid() was called we created a new ChannelFutureListener everytime. This is not needed as its stateless.

Modifications:

Reuse the ChannelFutureListener.

Result:

Less object allocations


normanmaurer:
Cherry-picked into 4.1 (3ca68ab254165025809c393999cd7f720f299b01)

Issue #7617, opened 2018-01-24
ObjectCleanerThread not be closed


hank-whu:
### Netty version
4.1.20.Final

### JVM version (e.g. `java -version`)
9.0.1

### OS version (e.g. `uname -a`)
Windows 10

Non-finished threads:

Thread[DestroyJavaVM,5,main]

Thread[ObjectCleanerThread,1,main]
  at java.base@9.0.1/java.lang.Object.wait(Native Method)
  at java.base@9.0.1/java.lang.ref.ReferenceQueue.remove(ReferenceQueue.java:151)
  at app//io.netty.util.internal.ObjectCleaner$1.run(ObjectCleaner.java:52)
  at app//io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
  at java.base@9.0.1/java.lang.Thread.run(Thread.java:844)

normanmaurer:
@hank-whu it should be after 10 seconds maximum if there are no more references: https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/internal/ObjectCleaner.java#L52

Are you sure there are no more references hold ?

johnou:
@normanmaurer why isn't there a break in the loop if the thread is interrupted? wouldn't that indicate the application is shutting down?

normanmaurer:
@johnou I think it should still cary on cleanup until there are no more references as otherwise we may leak direct memory etc. It should only exit once the `LIVE_SET` is empty imho.

johnou:
@normanmaurer then how about a daemon thread? Allowing the JVM process to terminate in a timely manner.

normanmaurer:
@johnou we decided against a daemon thread to make harder guarantees about cleanup. If it would be a deamon thread we may leak direct memory for example when someone hot-deploy stuff. 

Also /cc @Scottmitch , as both of us had the discussion :)

Scottmitch:
Yes we considered daemon but decided against it. Although I would expect most OSes we run on will clean up process resources after the process dies we can’t be sure of this. Since we allocate native resources outside the scope of the JVM the more conservative approach is to go non daemon. The mitigation factor was to have a timeout on the poll. This may cause issues if some of the jvm shutdown processing contains references to items registered in the cleaner but I don’t think this is likely.

hank-whu:
@normanmaurer 
shoud i wait 10 seconds after shutdown the EventLoopGroup ?
it doesn't seem like a good way.
[...]

PR #7627, opened 2018-01-25
ObjectCleanerThread must be a deamon thread to ensure the JVM can alw…

normanmaurer:
…ays terminate.

Motivation:

The ObjectCleanerThread must be a daemon thread as otherwise we may block the JVM from exit. By using a daemon thread we basically give the same garantees as the JVM when it comes to cleanup of resources (as the GC threads are also daemon threads and the CleanerImpl uses a deamon thread as well in Java9+).

Modifications:

Change ObjectCleanThread to be a daemon thread.

Result:

JVM shutdown will always be able to complete. Fixed [#7617].

### DIFF
--- a/common/src/main/java/io/netty/util/internal/ObjectCleaner.java
+++ b/common/src/main/java/io/netty/util/internal/ObjectCleaner.java
@@ -16,8 +16,6 @@
 package io.netty.util.internal;
 
 import io.netty.util.concurrent.FastThreadLocalThread;
-import io.netty.util.internal.logging.InternalLogger;
-import io.netty.util.internal.logging.InternalLoggerFactory;
 
 import java.lang.ref.ReferenceQueue;
 import java.lang.ref.WeakReference;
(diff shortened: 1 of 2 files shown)


common/src/test/java/io/netty/util/internal/ObjectCleanerTest.java
```diff
@@ -110,4 +112,35 @@ public void run() {
...
+                break;
+            }
+        }
+        assertNotNull(cleanerThread);
+        assertTrue(cleanerThread.isDaemon());
+
+        // Null out the temporary object to ensure it is enqueued for GC.
+        temporaryObject = null;
+
+        while (cleanerThread.isAlive()) {
```
johnou: never ending loop?

johnou:
cc @jaymode

common/src/test/java/io/netty/util/internal/ObjectCleanerTest.java
normanmaurer: Yeah this needs to be removed, done..

common/src/main/java/io/netty/util/internal/ObjectCleaner.java
```diff
@@ -116,9 +116,9 @@ public Void run() {
             });
             cleanupThread.setName(CLEANER_THREAD_NAME);
 
-            // This Thread is not a daemon as it will die once all references to the registered Objects will go away
-            // and its important to always invoke all cleanup tasks as these may free up direct memory etc.
-            cleanupThread.setDaemon(false);
+            // Mark this as a daemon thread to ensure that we the JVM can exit if this is the only thread that is
+            // running.
+            cleanupThread.setDaemon(true);
```
normanmaurer: @jaymode question... does this also need to be in a privileged block ?

jaymode: This is a good question. `Thread#setDaemon` does check permissions by calling `Thread#checkAccess`, but I think it is important to look at what happens there before wrapping in a doPrivileged block. In the default [SecurityManager](https://docs.oracle.com/javase/7/docs/api/java/lang/SecurityManager.html#checkAccess(java.lang.Thread)), the access control is only enforced for system threads. In that case, I think this should not be wrapped since we are not modifying a system thread. In elasticsearch we use [SecureSM](https://github.com/elastic/securesm/blob/master/src/main/java/org/elasticsearch/SecureSM.java), which has a stricter policy but even then I don't think you'd want to elevate privileges here since modifying threads within your own threadgroup or one of its descendants is allowed.

normanmaurer: @jaymode thanks a lot for the answer and sorry for bothering you but after your last PR I was wondering if I am doing something stupid ;)

normanmaurer:
Cherry-picked into 4.1 (1879433ae6df3add2dbde00722f8fef70c32af9c) and 4.0 (f5f7e6f9e5a4c0f01753ff56869c903e635bb472)

johnou:
@normanmaurer any chance of another release soon? we needed the 4.1.20 release for the iovarray fix but it also contained this regression with the object cleaner.

normanmaurer:
Yep will do this week

Issue #7628, opened 2018-01-25
closing an http/2 stream a second time throws an exception


mosesn:
h/t to @vkostyukov for finding this issue.  I'm just the messenger!

### Expected behavior

Closing an http/2 stream a second time should not have a performance impact

### Actual behavior

We found this stack trace when we ran the java async-profiler.

```
Total: 1087173410 (0.97%)  samples: 1084
  [ 0] java_lang_Throwable::fill_in_stack_trace(Handle, methodHandle, Thread*)
  [ 1] java_lang_Throwable::fill_in_stack_trace(Handle, methodHandle)
  [ 2] JVM_FillInStackTrace
  [ 3] Java_java_lang_Throwable_fillInStackTrace
  [ 4] java.lang.Throwable.fillInStackTrace
  [ 5] java.lang.Throwable.fillInStackTrace
  [ 6] java.lang.Throwable.<init>
  [ 7] java.lang.Exception.<init>
  [ 8] java.lang.RuntimeException.<init>
  [ 9] java.lang.IllegalStateException.<init>
  [10] io.netty.channel.VoidChannelPromise.fail
  [11] io.netty.channel.VoidChannelPromise.addListener
  [12] io.netty.channel.VoidChannelPromise.addListener
  [13] io.netty.handler.codec.http2.Http2MultiplexCodec$DefaultHttp2StreamChannel$Http2ChannelUnsafe.close
  [14] io.netty.channel.DefaultChannelPipeline$HeadContext.close
  [15] io.netty.channel.AbstractChannelHandlerContext.invokeClose
  [16] io.netty.channel.AbstractChannelHandlerContext.close
  [17] io.netty.channel.ChannelOutboundHandlerAdapter.close
  [18] io.netty.channel.AbstractChannelHandlerContext.invokeClose
  [19] io.netty.channel.AbstractChannelHandlerContext.close
  [20] io.netty.channel.ChannelDuplexHandler.close
  [21] io.netty.channel.AbstractChannelHandlerContext.invokeClose
  [22] io.netty.channel.AbstractChannelHandlerContext.close
  [23] io.netty.channel.ChannelDuplexHandler.close
  [24] io.netty.channel.AbstractChannelHandlerContext.invokeClose
  [25] io.netty.channel.AbstractChannelHandlerContext.close
  [26] io.netty.channel.ChannelDuplexHandler.close
  [27] io.netty.channel.AbstractChannelHandlerContext.invokeClose
[...]

PR #7631, opened 2018-01-26
Correctly handle multiple calls to DefaultHttp2StreamChannel.Unsafe.c…

normanmaurer:
…lose(...)

Motivation:

Calling DefaultHttp2StreamChannel.Unsafe.close(...) multiple times should not fail.

Modification:

- Correctly handle multiple calls to DefaultHttp2StreamChannel.Unsafe.close(...)
- Add unit test

Result:

Fixes [#7628]

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
@@ -32,7 +32,6 @@ import io.netty.channel.ChannelPromise;
 import io.netty.channel.DefaultChannelConfig;
 import io.netty.channel.DefaultChannelPipeline;
 import io.netty.channel.DefaultMaxMessagesRecvByteBufAllocator;
-import io.netty.channel.DelegatingChannelPromiseNotifier;
 import io.netty.channel.EventLoop;
 import io.netty.channel.MessageSizeEstimator;
 import io.netty.channel.RecvByteBufAllocator;
@@ -768,7 +767,7 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
             @SuppressWarnings("deprecation")
             private RecvByteBufAllocator.ExtendedHandle recvHandle;
             private boolean writeDoneAndNoFlush;
-            private ChannelPromise pendingClosePromise;
+            private boolean closeInitiated;
 
             @Override
             public void connect(final SocketAddress remoteAddress,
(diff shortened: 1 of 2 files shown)


normanmaurer:
@mosesn @vkostyukov PTAL

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -844,57 +843,58 @@ public void bind(SocketAddress localAddress, ChannelPromise promise) {
...
 
             @Override
-            public void close(ChannelPromise promise) {
+            public void close(final ChannelPromise promise) {
                 if (!promise.setUncancellable()) {
                     return;
                 }
-                if (closePromise.isDone()) {
-                    promise.setFailure(new ClosedChannelException());
+                if (closeInitiated) {
```
normanmaurer: This is the same code we use in `AbstractChannel`

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -844,57 +843,58 @@ public void bind(SocketAddress localAddress, ChannelPromise promise) {
...
+                    if (closePromise.isDone()) {
```
bryce-anderson: From my reading of this method, the first time `close` is called the `closePromise` will be completed on line 897, so this should always evaluate to true. The major behavior change is that instead of completing the promise like `promise.setFailure(new ClosedChannelException())` we now do `promise.setSuccess()` which appears to be consistent with the behavior of the `AbstractUnsafe` implementation.

So, it seems like the state `closeInitiated` is redundant with the state of `closePromise` and as such the block
```java
             if (closeInitiated) {
                    if (closePromise.isDone()) {
                        // Closed already.
                        promise.setSuccess();
                    } else if (!(promise instanceof VoidChannelPromise)) { // Only needed if no VoidChannelPromise.
                        // This means close() was called before so we just register a listener and return
                        closePromise.addListener(new ChannelFutureListener() {
                            @Override
                                promise.setSuccess();
                            }
                        });
                    }
                    return;
                }
```
can be reduced to
```java
if (closePromise.isDone()) {
        // Closed already.
        promise.setSuccess(); // safe on VoidPromise
        return;
}
```

normanmaurer: @bryce-anderson I think the difference here is that if we do it with the listener is that it will be done in the "correct" order if close(...) is called in a re-entrance way.

bryce-anderson: Ah, you're totally right. From looking at `AbstractUnsafe`, it appears that `closeFuture` is satisfied before the passed `promise` in the `doClose0` method, and that happens before calling the `fireChannelInactiveAndDeregister` method. If we switch the order of 
```java
promise.setSuccess();
closePromise.setSuccess();
```
to mirror that and place it before the inactive and deregister logic, we might avoid any reentrance issues outside of the write and flush of the RST frame, which also seems safe to move to after the satisfaction of `closePromise`.

normanmaurer: @bryce-anderson so you are happy with my patch ?

bryce-anderson: I think this is definitely a step in the right direction, so I'm good with this patch. I think one of our root problems is that the `StreamChannel` will report `.isOpen() == true` even within a `channelInactive` event because `.isOpen()` is predicated on the state of the `closePromise` which gets completed as the very last step. It's not clear to me one way or the other if this is expected behavior, but I suspect it is not.

More specifically, we encounter reentrant behavior because we predicate a call to `Channel.close()` on `Channel.isClosed()` in response to a `channelInactive` event.
See string of events here:
https://github.com/twitter/finagle/blob/develop/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/transport/ChannelTransport.scala#L188
https://github.com/twitter/finagle/blob/develop/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/transport/ChannelTransport.scala#L94
https://github.com/twitter/finagle/blob/develop/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/transport/ChannelTransport.scala#L141

It's entirely possible that we're misusing the event model.

normanmaurer: @bryce-anderson I think thats actually another issue... Let us fix the issue that was reported here first and then I will fix the other thing you brought up on another PR... WDYT ?

bryce-anderson: That seems reasonable. Do you want to open the issue or should I?

normanmaurer: please do @bryce-anderson

Scottmitch:
@normanmaurer - doh I didn't see this before I submitted https://github.com/netty/netty/pull/7643 ... this PR should be preferred bcz the memory should be less (you now have a boolean vs a ChannelPromise reference). Also it looks like VoidPromise is not allowed in this method via [AbstractChannelHandlerContext](https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java#L600) so not sure the `VoidPromise` conditional is necessary.

We can just pull in the other PR for the unit tests.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -844,57 +843,58 @@ public void bind(SocketAddress localAddress, ChannelPromise promise) {
...
-                    }
-                    promise.setSuccess();
-                    closePromise.setSuccess();
-                } finally {
-                    pendingClosePromise = null;
+                pipeline().fireChannelInactive();
+                if (isRegistered()) {
+                    deregister(unsafe().voidPromise());
                 }
+                promise.setSuccess();
```
Scottmitch: the order of this line and the next line should be reversed to be consistent with `AbstractChannel`. see https://github.com/netty/netty/pull/7643

normanmaurer:
Addressed comments and cherry-picked into 4.1 (b423a3578300b1dbcd8643336420d9683a023f51)

mosesn:
@normanmaurer sorry I didn't get a chance to review this, thanks for handling this so quickly!

normanmaurer:
@mosesn no worries :)

PR #7632, opened 2018-01-26
Reduce objects by directly implement interface in internal implementa…

normanmaurer:
…tions of DefaultHttp2RemoteFlowController

Motivation:

We can just implement the interfaces directly and so reduce object creation in DefaultHttp2RemoteFlowController.

Modifications:

Directly implement the interfaces.

Result:

Less object creation.


normanmaurer:
Small optimization that can be done...

PR #7633, opened 2018-01-26
Method to check if a Http2 header is present and has a given value

tsegismont:
Motivation:

With HTTP1, it's very easy to check if a header is present and has a
given value: you can simply invoke
io.netty.handler.codec.http.HttpHeaders#contains(java.lang.CharSequence, java.lang.CharSequence, boolean)

It is not possible to do the same with HTTP2. You have to get the list
of all headers (returned as String) and then iterate over it invoking
String#equals or String#equalsIgnoreCase

Modifications:

I've added io.netty.handler.codec.http2.Http2Headers#contains and
implemented it in DefaultHttp2Headers, EmptyHttp2Headers and ReadOnlyHttp2Headers.

Result:

You can use AsciiString constants to check if a header is present in a
consice and efficient manner.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Headers.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Headers.java
@@ -14,16 +14,16 @@
  */
 package io.netty.handler.codec.http2;
 
-import static io.netty.handler.codec.http2.Http2Error.PROTOCOL_ERROR;
-import static io.netty.handler.codec.http2.Http2Exception.connectionError;
-import static io.netty.util.AsciiString.CASE_SENSITIVE_HASHER;
-import static io.netty.util.AsciiString.isUpperCase;
 import io.netty.handler.codec.CharSequenceValueConverter;
 import io.netty.handler.codec.DefaultHeaders;
 import io.netty.util.AsciiString;
 import io.netty.util.ByteProcessor;
 import io.netty.util.internal.PlatformDependent;
 import io.netty.util.internal.UnstableApi;
+
+import static io.netty.handler.codec.http2.Http2Error.*;
+import static io.netty.handler.codec.http2.Http2Exception.*;
+import static io.netty.util.AsciiString.*;
 
 @UnstableApi
 public class DefaultHttp2Headers
@@ -181,6 +181,11 @@ public class DefaultHttp2Headers
     @Override
     public CharSequence status() {
         return get(PseudoHeaderName.STATUS.value());
+    }
+
+    @Override
+    public boolean contains(CharSequence name, CharSequence value, boolean caseInsensitive) {
+        return contains(name, value, caseInsensitive? CASE_INSENSITIVE_HASHER : CASE_SENSITIVE_HASHER);
     }
 
     @Override
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
@@ -17,6 +17,7 @@ package io.netty.handler.codec.http2;
 
 import io.netty.handler.codec.Headers;
 import io.netty.util.AsciiString;
+import io.netty.util.HashingStrategy;
 
 import java.util.ArrayList;
 import java.util.Collections;
@@ -27,9 +28,10 @@ import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Set;
 
-import static io.netty.handler.codec.CharSequenceValueConverter.INSTANCE;
-import static io.netty.handler.codec.http2.DefaultHttp2Headers.HTTP2_NAME_VALIDATOR;
-import static io.netty.util.internal.EmptyArrays.EMPTY_ASCII_STRINGS;
+import static io.netty.handler.codec.CharSequenceValueConverter.*;
+import static io.netty.handler.codec.http2.DefaultHttp2Headers.*;
+import static io.netty.util.AsciiString.*;
+import static io.netty.util.internal.EmptyArrays.*;
 
 /**
  * A variant of {@link Http2Headers} which only supports read-only methods.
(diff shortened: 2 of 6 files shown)


tsegismont:
@normanmaurer here's the pull request. While working on it, I noticed two things:

1. `ReadOnlyHttp2Headers#contains(java.lang.CharSequence, java.lang.CharSequence)` always ignores case for values

See this test https://github.com/tsegismont/netty/blob/a9b8bda56b5e65fffdbdf1e6cf998fa980144fcc/codec-http2/src/test/java/io/netty/handler/codec/http2/ReadOnlyHttp2HeadersTest.java#L138

2. `DefaultHttp2Headers#contains(java.lang.CharSequence, java.lang.CharSequence)` returns false if value is present but header has multiple value.

That looks incorrect to me. Or am I missing something?

normanmaurer:
@tsegismont yes this looks like a bug... want to provide a PR ?

tsegismont:
Separate PRs for both issues or new commits to this one?

normanmaurer:
@tsegismont separate please :)

tsegismont:
@normanmaurer no pb, but could we wait for this one to be merged? That would save me some conflict resolution :-)

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
```diff
@@ -769,6 +771,30 @@ public CharSequence status() {
...
+    public boolean contains(CharSequence name, CharSequence value, boolean caseInsensitive) {
+        final int nameHash = AsciiString.hashCode(name);
+        final HashingStrategy<CharSequence> strategy = caseInsensitive? CASE_INSENSITIVE_HASHER : CASE_SENSITIVE_HASHER;
+        final int valueHash = strategy.hashCode(value);
+
+        return contains(name, nameHash, value, valueHash, strategy, pseudoHeaders)
+               || contains(name, nameHash, value, valueHash, strategy, otherHeaders);
+    }
+
+    private static boolean contains(CharSequence name, int nameHash, CharSequence value, int valueHash,
```
Scottmitch: can we use this for `contains(CharSequence name, CharSequence value)`? looks like there is a bug in the value comparison which should be case sensitive.

Scottmitch: followup PR is fine for this

tsegismont:
@normanmaurer the fixes are ready but as commented previously I would like to wait for this PR to be merged first

>ReadOnlyHttp2Headers#contains(java.lang.CharSequence, java.lang.CharSequence) always ignores case for values

See https://github.com/tsegismont/netty/commit/f6bd5a0b1340277dc3620bcc9a04b24b1284d904

>DefaultHttp2Headers#contains(java.lang.CharSequence, java.lang.CharSequence) returns false if value is present but header has multiple values.

I was wrong. The problem was not related to headers having multiple values. The problems was that the base `contains()` implementation uses the `JAVA_HASHER`

See https://github.com/tsegismont/netty/commit/e024f349c5b6c298318d4752a1fcf02da5889b2b

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
normanmaurer: @Scottmitch I think that was what @tsegismont wanted to do one this is merged: https://github.com/netty/netty/pull/7633#issuecomment-360745218

tsegismont: @Scottmitch yep, see my other comments in the PR, I noticed the bug and Norman asked to fix it separately

Scottmitch:
> I would like to wait for this PR to be merged first

We will include both fixes in the next release, and if they are fixed in separate PRs we generally need to merge one then the other. Is this problematic for you?

normanmaurer:
I asked to do it in separate prs as it is not really related as one is a new feature and the other is a bug fix.

tsegismont:
>I asked to do it in separate prs as it is not really related as one is a new feature and the other is a bug fix.

@normanmaurer sure, makes sense

>Is this problematic for you?

@Scottmitch I just meant that https://github.com/tsegismont/netty/commit/f6bd5a0b1340277dc3620bcc9a04b24b1284d904 and https://github.com/tsegismont/netty/commit/e024f349c5b6c298318d4752a1fcf02da5889b2b depend on this PR, so it's easier for me to wait for it being merged before opening new ones.

Scottmitch:
> so it's easier for me to wait for it being merged before opening new ones.

Thanks for clarifying.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Headers.java
```diff
@@ -183,6 +183,11 @@ public CharSequence status() {
         return get(PseudoHeaderName.STATUS.value());
     }
 
+    @Override
+    public boolean contains(CharSequence name, CharSequence value, boolean caseInsensitive) {
+        return contains(name, value, caseInsensitive? CASE_INSENSITIVE_HASHER : CASE_SENSITIVE_HASHER);
```
carl-mastrangelo: s/e?/e ?/

codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
```diff
@@ -769,6 +771,30 @@ public CharSequence status() {
         return get(PseudoHeaderName.STATUS.value());
     }
 
+    @Override
+    public boolean contains(CharSequence name, CharSequence value, boolean caseInsensitive) {
+        final int nameHash = AsciiString.hashCode(name);
+        final HashingStrategy<CharSequence> strategy = caseInsensitive? CASE_INSENSITIVE_HASHER : CASE_SENSITIVE_HASHER;
+        final int valueHash = strategy.hashCode(value);
+
+        return contains(name, nameHash, value, valueHash, strategy, pseudoHeaders)
```
carl-mastrangelo: This should be in the reverse order.   Pseudo headers are checked more rarely, so they should be checked last.

Scottmitch: sgtm ... can you submit a followup PR?

carl-mastrangelo: Done, see #7642

PR #7634, opened 2018-01-26
Cleanup Http2MultiplexCodec by removing out-dated TODO

normanmaurer:
Motivation:

Http2MultiplexCodec contains some TODO that is outdated.

Modifications:

Remove TODO which is outdated

Result:

Cleaner code.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
@@ -155,9 +155,6 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
     private final ChannelHandler inboundStreamHandler;
 
     private int initialOutboundStreamWindow = Http2CodecUtil.DEFAULT_WINDOW_SIZE;
-    // TODO: We may be able to optimize when we really need to call flush(...) during channelReadComplete(...)
-    // by checking if this is true and only then call flush(...).
-    private boolean flushNeeded;
     private boolean parentReadInProgress;
     private int idCount;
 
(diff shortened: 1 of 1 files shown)


normanmaurer:
Some left-over from my work that can be removed now.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -155,9 +155,6 @@ public int guess() {
     private final ChannelHandler inboundStreamHandler;
 
     private int initialOutboundStreamWindow = Http2CodecUtil.DEFAULT_WINDOW_SIZE;
-    // TODO: We may be able to optimize when we really need to call flush(...) during channelReadComplete(...)
```
Scottmitch: do we have a followup issue to investigate this, or you feel it is no longer necessary?

normanmaurer: no longer necessary.

Scottmitch: Can you explain why? Is it because you are now convinced that calling flush is cheap enough if there is nothing to flush (despite the pipeline activity, and pipelines being deeper than in the "low level API")?

normanmaurer: Yes exactly... (+ its the same we do in the Http2ConnectionHandler).

Scottmitch: > (+ its the same we do in the Http2ConnectionHandler).

The pipeline is generally assumed to be more shallow because it is not very extensible. We can tweak later if necessary though (just wondering if perf analysis was done).

normanmaurer: yeah we can followup if needed... that said I dont think so for now

normanmaurer:
Cherry-picked into 4.1 (c43dc3364b7db4f4ea1de9128c7cbb76eb1379ba)

PR #7635, opened 2018-01-26
Http2MultiplexCodec should not fire Http2GoAwayFrame to child channel…

normanmaurer:
…s as user event

Motivation:

We should not fire the Http2GoAwayFrame to the child channels at all when using Http2MultiplexCodec. We should just pass it further through the pipeline so if a user is interested in the frame its possible to just handle directly in the parents ChannelPipeline.

Modifications:

Remove code that fires the frame throught the child channels

Result:

Less confusing behaviour.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
@@ -220,8 +220,6 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
         if (frame instanceof Http2StreamFrame) {
             Http2StreamFrame streamFrame = (Http2StreamFrame) frame;
             onHttp2StreamFrame(((Http2MultiplexCodecStream) streamFrame.stream()).channel, streamFrame);
-        } else if (frame instanceof Http2GoAwayFrame) {
-            onHttp2GoAwayFrame(ctx, (Http2GoAwayFrame) frame);
         } else if (frame instanceof Http2SettingsFrame) {
             Http2Settings settings = ((Http2SettingsFrame) frame).settings();
             if (settings.initialWindowSize() != null) {
@@ -315,28 +313,6 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
                 tail = childChannel;
             }
             childChannel.fireChannelReadPending = true;
-        }
-    }
-
-    private void onHttp2GoAwayFrame(ChannelHandlerContext ctx, final Http2GoAwayFrame goAwayFrame) {
-        try {
-            forEachActiveStream(new Http2FrameStreamVisitor() {
-                @Override
-                public boolean visit(Http2FrameStream stream) {
-                    final int streamId = stream.id();
-                    final DefaultHttp2StreamChannel childChannel = ((Http2MultiplexCodecStream) stream).channel;
-                    if (streamId > goAwayFrame.lastStreamId() && connection().local().isValidStreamId(streamId)) {
-                        childChannel.pipeline().fireUserEventTriggered(goAwayFrame.retainedDuplicate());
-                    }
-                    return true;
-                }
-            });
-        } catch (Http2Exception e) {
-            ctx.fireExceptionCaught(e);
-            ctx.close();
-        } finally {
-            // We need to ensure we release the goAwayFrame.
-            goAwayFrame.release();
         }
     }
 


normanmaurer:
Also @mosesn and @vkostyukov

Scottmitch:
The motivation for this change is not clear. _Why_ do we not want to tell the user about GO_AWAY? What if the user has some cleanup it needs to initiate in preparation for closure?

normanmaurer:
@Scottmitch IMHO we should not fire frames via `fireUserEventTriggered`. If the user really wants to do some cleanup then it should be done in a `ChannelInboundHandler` that sits after the `Http2MultiplexCodec`. This will receive the frame. IMHO the child channel should only receive frames that are related directly to its stream.

If we want to also pass the frame to the child channel (which I am not convinced yet we should) then we should at least use `fireChannelRead` and not `fireUserEventTriggered`.

Hope this makes it more clear

Scottmitch:
> IMHO we should not fire frames via fireUserEventTriggered

Agreed. I forgot we recently made a change to fire frames on the parent channel.

> Hope this makes it more clear

+1. Consider capturing this in the `Motivation` section of the commit.

ejona86:
> IMHO we should not fire frames via fireUserEventTriggered

But it doesn't seem appropriate to wait on delivering this until read is called. Why do I have to open up further flow control window to receive this frame?

> IMHO the child channel should only receive frames that are related directly to its stream.

This is possibly related to the stream. We see things like GOAWAY("max_age") and then the connection is killed. It is useful for the stream to be aware there was the goaway in those error cases.

Scottmitch:
> But it doesn't seem appropriate to wait on delivering this until read is called.

When you say "read" I'm assuming you talking about the `channel.read()`on the child channels (bcz the parent channel's auto read is always on). This argument makes a lot of sense if we want to make control messages available to child channels, and I feel like we have discussed it before and its all coming back to me now. I'm back on team @ejona86 for GO_AWAY on child channels and think we should at the very least add a comment explaining the rational for using a userEvent so we don't forget in the future.

ejona86:
> When you say "read" I'm assuming you talking about the channel.read() on the child channels (bcz the parent channel's auto read is always on).

Yes. That's what I was referring to. When AUTO_READ is off.

> and think we should at the very least add a comment explaining the rational for using a userEvent so we don't forget in the future.

\<Insert witty comment about flow control>

And FWIW, I could accept the argument that this sort of thing is too niche, but I wanted to make sure the use-case was clear.

Scottmitch:
> I could accept the argument that this sort of thing is too niche, but I wanted to make sure the use-case was clear.

I think its fine to provide additional info that would otherwise not be obtainable via the child channel API. As you previously said ignoring the event has no impact, but perhaps it will help folks initiate graceful cleanup on each child channel.

A related question: should we also want to propagate the event to the parent channel? For example if a proxy receives a GO_AWAY, maybe it makes sense to forward that on. If so, should we stick with "userEvents for control messages" theme?

normanmaurer:
@Scottmitch @ejona86 I am still not convinced... why is it not good enough to "just" fire it on he parent channel ?

normanmaurer:
@trustin so you are fine with not firing the go away on the child channel ? I think we are still not all in agreement what we should do or not ;)

trustin:
Obviously I didn't read the comments in detail but just code. :-P

How about just making `onHttp2GoAwayFrame` method protected and allow a user to override them? The default behavior may be debatable but it should be fine as long as it's customizable and the default behavior is well documented.

ejona86:
> why is it not good enough to "just" fire it on he parent channel ?

Looking at the code some more, I'm remembering more of what was going on. We only propagate when `streamId > goAwayFrame.lastStreamId()`. That is, we only propagate for streams closed because of the GOAWAY. Thus, this is just as important as propagating RST_STREAM(REFUSED_STREAM).

We could propagate RST_STREAM(REFUSED_STREAM) to the child streams instead of GOAWAY if we wanted, as it would be equivalent semantically (although the spec doesn't call this out). It is true that a child channel receiving the GOAWAY probably shouldn't look at the error code nor error message.

> For example if a proxy receives a GO_AWAY, maybe it makes sense to forward that on.

Generally not. GOAWAY is point-to-point. Most (L7) proxies would simply create a new connection to the backend and let the end-client be non-the-wiser. If a stream received the GOAWAY with the proxy, and the proxy does not have the request buffered, then it would likely send RST_STREAM(REFUSED_STREAM) to the client and let the client retry (even on the same connection).

Scottmitch:
> That is, we only propagate for streams closed because of the GOAWAY.

In this case it may make more sense to fire an exception and forcibly close the channel, because the peer has said this stream is not valid. `RST_STREAM(REFUSED_STREAM)` seems reasonable to me.

> why is it not good enough to "just" fire it on he parent channel ?

My rational was from the user perspective the child channels are logically independent from the parent channel. If the parent channel gets an event we don't provide a way to get all child channels. So we wouldn't want users to try and maintain this information them selves.

> if a stream received the GOAWAY with the proxy, and the proxy does not have the request buffered, then it would likely send RST_STREAM(REFUSED_STREAM)

Seems like you are saying that we should always propagate the go away (not just when `streamId > goAwayFrame.lastStreamId()` is true) which is what I assumed we were doing. This gives the child channel the opportunity to attempt some fashion of graceful shutdown.

ejona86:
> In this case it may make more sense to fire an exception and forcibly close the channel

Close the channel, yes. Fire an exception? Is that normal to do when the remote closes with TCP RST? I'd need to do more research how Netty already handles the remote resetting the connection (with and without half close support).

Note that in this we're a bit better off because no data will have been received, so there's more limited concern about flow control. But if we're handling RST_STREAM in other cases, we'd want to do something similar here.

>> if a stream received the GOAWAY with the proxy, and the proxy does not have the request buffered, then it would likely send RST_STREAM(REFUSED_STREAM)
>
> Seems like you are saying that we should always propagate the go away (not just when streamId > goAwayFrame.lastStreamId() is true) which is what I assumed we were doing.

No, the proxy behavior I was discussing would only apply to the streams impacted by the GOAWAY.

If we do always deliver the GOAWAY, then we would need some way for the child channels to determine if the stream was reset because of that GOAWAY. Since handlers on the child channel don't know their stream id, they can't use an ID comparison to determine if they are directly impacted. It seemed like you liked the idea faking RST_STREAM(REFUSED_STREAM) for the `lastStreamId` handling; if we did that, then child channels would be okay, but we would need to think carefully about which notification we deliver first.

> This gives the child channel the opportunity to attempt some fashion of graceful shutdown.

This seems complex and separated enough for an issue. Created https://github.com/netty/netty/issues/7651 and replied there.

Scottmitch:
> Is that normal to do when the remote closes with TCP RST?

Yes. fireExceptionCaught is called with an IoException and then channelInactive is called.

> This seems complex and separated enough for an issue.

Thanks!

normanmaurer:
@Scottmitch @ejona86 @carl-mastrangelo I totally lost track here now... What was the conclusion here 🤣  ?

ejona86:
Summary as best as I can tell:
* This was serving a purpose, to notify streams killed because of a GOAWAY details of why they were killed (and specifically in this instance, that the server did not process them).
* It seems it is good and fine to model this as a RST_STREAM(REFUSED_STREAM) from the stream's perspective. This is what a proxy would do, for instance. We would probably want to call out in some documentation somewhere that such a transform is done (primarily to avoid problems when debugging "who sent RST_STREAM?").

And then:
* close channel+fireExceptionCaught to signal the RST_STREAM.

I'm less certain about that last part because I really don't know how things mix with flow control and whether the RST_STREAM should cause any received data buffered to be trashed unconditionally (I thought the OS doesn't do that). It seems @Scottmitch knows what goes on here for TCP RST, and as a general rule I'm fine with copying what happens with TCP RST. We'd need to figure out what exception to use.

(In my design, I had said we'd propagate the RST_STREAM as a user event for HTTP/2-aware applications to observe immediately and then emulate the channel closure more like normal (which I thought would involve draining the inbound queue). @Scottmitch isn't wholly against using user events for control messages; discussion for one case to do so was moved to #7651. User events have the benefit of ignoring flow control.)

normanmaurer:
Closing this as https://github.com/netty/netty/pull/9290 is preferable.

PR #7636, opened 2018-01-26
ReadOnlyHttp2Headers.contains always ignores case for values

tsegismont:
Motivation:

When checking if a value is present, ReadOnlyHttp2Headers always ignores
case for values.

RFC 7540 says: https://tools.ietf.org/html/rfc7540#section-8.1.2
"header field names are strings of ASCII characters that are compared in a case-insensitive fashion"

But there is no such constraint on header values

Modifications:

Updated ReadOnlyHttp2Headers.contains to compare header value in a
case-sensitive way.

Result:

ReadOnlyHttp2Headers compares header names in a case-insensitive way,
values in a case-sensitive way.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/ReadOnlyHttp2Headers.java
@@ -430,29 +430,7 @@ public final class ReadOnlyHttp2Headers implements Http2Headers {
 
     @Override
     public boolean contains(CharSequence name, CharSequence value) {
-        final int nameHash = AsciiString.hashCode(name);
-        final int valueHash = AsciiString.hashCode(value);
-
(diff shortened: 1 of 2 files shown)


Scottmitch:
Can you sign the ICLA: https://netty.io/wiki/developer-guide.html?

tsegismont:
I believe Red Hat has signed the CCLA, hasn't it @normanmaurer ?

normanmaurer:
@tsegismont nope there is no CCLA from redhat... maybe there should ;)

https://netty.io/s/ccla

normanmaurer:
@tsegismont that said you can also just sign the ICLA which may be easier and faster

tsegismont:
@Scottmitch @normanmaurer just signed the ICLA

normanmaurer:
ICLA was signed.

Cherry-picked into 4.1 (7f23c34b55428224e102fadeba4bbed2187858d5)

normanmaurer:
@tsegismont thanks!

PR #7637, opened 2018-01-26
DefaultHttp2Headers#contains(CharSequence, CharSequence) does not work with String

tsegismont:
Motivation:

If you test a header value providing a String, contains() returns false.
This is due to the implementation inherited from DefaultHeaders using
the JAVA_HASHER.

JAVA_HASHER.equals returns false because a is a String and b an
AsciiString.

Modifications:

DefaultHttp2Headers overrides contains and uses CASE_SENSITIVE_HASHER.

Result:

You can test a header value with any CharSequence implementation.


Scottmitch:
will merge after we get the CLA sorted out

normanmaurer:
As the ICLA was signed, I cherry-picked it into 4.1 (0db652277f328c5b0d10afd16e19d78b897f5d82)

Issue #7638, opened 2018-01-26
The Http2MultiplexCodec StreamChannel will fire channelInactive without the channel.isOpen method reporting false


bryce-anderson:
### Expected behavior

When part of a HTTP/2 `StreamChannel` the `StreamChannel.isOpen()` method should report `false` within a call to a `ChannelInboundHandler`s `channelInactive` method.

### Actual behavior

Within the `channelInactive` method, `ctx.channel().isOpen() == true`

### Minimal yet complete reproducer code (or URL to code)

Within the `Http2MultiplexCodecTest:
```java
    @Test
    public void channelClosedWhenInactiveFired() {
        LastInboundHandler inboundHandler = streamActiveAndWriteHeaders(inboundStream);
        Http2StreamChannel childChannel = (Http2StreamChannel) inboundHandler.channel();

        final AtomicBoolean channelWasOpen = new AtomicBoolean(false);

        childChannel.pipeline().addLast(new ChannelInboundHandlerAdapter() {
            @Override
            public void channelInactive(ChannelHandlerContext ctx) throws Exception {
                channelWasOpen.set(ctx.channel().isOpen());
                super.channelInactive(ctx);
            }
        });

        childChannel.close().syncUninterruptibly();
        assertFalse(channelWasOpen.get());
    }
```

Result: 
java.lang.AssertionError
io.netty.handler.codec.http2.Http2MultiplexCodecTest.channelClosedWhenInactiveFired(Http2MultiplexCodecTest.java:471)

### Netty version

4.1

### JVM version (e.g. `java -version`)
openjdk version "1.8.0_152"

### OS version (e.g. `uname -a`)
Darwin

bryce-anderson:
See the discussion in https://github.com/netty/netty/pull/7631 for more details including a possible solution.

normanmaurer:
@bryce-anderson thanks a lot! I will fix it :)

normanmaurer:
@bryce-anderson PTAL https://github.com/netty/netty/pull/7640

normanmaurer:
Fixed

PR #7640, opened 2018-01-26
Http2MultiplexCodec.DefaultHttpStreamChannel.isOpen() / isActive() sh…

normanmaurer:
…oule be false when fireChannelActive() is called.

Motivation:

When part of a HTTP/2 StreamChannel the Http2StreamChannel.isOpen() / isActive() should report false within a call to a ChannelInboundHandlers channelInactive() method.

Modifications:

Fullfill promise before call fireChannelInactive()

Result:

Correctly update state / promise before notify handlers. Fixes [#7638]

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
@@ -886,12 +886,14 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
                         }
                     }
 
+                    // The promise should be notified before we call fireChannelInactive().
+                    promise.setSuccess();
+                    closePromise.setSuccess();
+
                     pipeline().fireChannelInactive();
                     if (isRegistered()) {
                         deregister(unsafe().voidPromise());
                     }
-                    promise.setSuccess();
-                    closePromise.setSuccess();
                 } finally {
                     pendingClosePromise = null;
                 }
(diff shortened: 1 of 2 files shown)


normanmaurer:
This is also consistent with `AbstractChannel`

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -886,12 +886,14 @@ public void close(ChannelPromise promise) {
                         }
                     }
 
+                    // The promise should be notified before we call fireChannelInactive().
+                    promise.setSuccess();
```
bryce-anderson: I also think the order of these two needs to be reversed, the following test fails:
```java
@Test
    public void channelClosedWhenClosePromiseCompletes() {
        LastInboundHandler inboundHandler = streamActiveAndWriteHeaders(inboundStream);
        Http2StreamChannel childChannel = (Http2StreamChannel) inboundHandler.channel();

        assertTrue(childChannel.isOpen());
        assertTrue(childChannel.isActive());

        final AtomicBoolean channelOpen = new AtomicBoolean(true);

        final ChannelPromise p = childChannel.newPromise();

        p.addListener(new ChannelFutureListener() {
            @Override
                channelOpen.set(future.channel().isOpen());
            }
        });

        childChannel.close(p).syncUninterruptibly();

        assertFalse(channelOpen.get());
    }
```

Reversing the order fixes the test above. I'm happy to file another bug if you prefer (perhaps that breaks something else I'm not aware of).

normanmaurer:
Yes please open another issue 👍

vkostyukov:
Thanks a lot, @normanmaurer!

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
Scottmitch: +1 ... there was actually another issue too so I just submitted https://github.com/netty/netty/pull/7643 with a modified version of the unit test above.

Issue #7641, opened 2018-01-26
The Promise passed to a StreamChannel.close(..) call will report channel().isOpen() == true in its listeners


bryce-anderson:
### Expected behavior

When the promise is completed, it's channel should report being closed in the handler.

### Actual behavior

The channel still reports open.

### Minimal yet complete reproducer code (or URL to code)
Test case added to `Http2MultiplexCodecTest`:
```java
    @Test
    public void channelClosedWhenClosePromiseCompletes() {
        LastInboundHandler inboundHandler = streamActiveAndWriteHeaders(inboundStream);
        Http2StreamChannel childChannel = (Http2StreamChannel) inboundHandler.channel();

        assertTrue(childChannel.isOpen());
        assertTrue(childChannel.isActive());

        final AtomicBoolean channelOpen = new AtomicBoolean(true);

        final ChannelPromise p = childChannel.newPromise();

        p.addListener(new ChannelFutureListener() {
            @Override
            public void operationComplete(ChannelFuture future) throws Exception {
                channelOpen.set(future.channel().isOpen());
            }
        });

        childChannel.close(p).syncUninterruptibly();

        assertFalse(channelOpen.get());
    }
```

### Netty version
4.1

### JVM version (e.g. `java -version`)
openjdk version "1.8.0_152"

### OS version (e.g. `uname -a`)
Darwin

bryce-anderson:
See https://github.com/netty/netty/pull/7640 for the discussion. The solution is to change the order of promise satisfaction [here](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java#L893) as `ch.isClose()` is predicated on the state of the `StreamChannel`s `closePromise` [here](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java#L511).

bryce-anderson:
Btw, I'm happy to contribute a fix for this. We make heavy use of Netty and are building out on these new API's so I'd personally love to get more involved. That said, fft just fix it if you think its expedient.

normanmaurer:
Sounds good 👌 that said I think you should not re-order the promises but use an extra volatile boolean that is updated before any of the promises is notified. This way the notifications of the promises is also consistent with AbstractChannel and so other transports...

So please provide a PR with the suggestion I made 👍

bryce-anderson:
[...]

PR #7643, opened 2018-01-27
Http2MultiplexChannel close promise completion consistency with Abstr…

Scottmitch:
…actChannel

Motivation:
The completion order of promises in Http2MultiplexChannel#close should be consistent with that of AbstractChannel. Otherwise this may result in Future listeners seeing incorrect channel state.

Modifications:
- Subsequent close operations should always result in successfully completing the close promise
- The channel's closePromise should be completed before the close(Promise) argument

Result:
More consistent behavior between Http2MultiplexChannel and AbstractChannel.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
@@ -856,7 +856,7 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
                     return;
                 }
                 if (closePromise.isDone()) {
-                    promise.setFailure(new ClosedChannelException());
+                    promise.setSuccess();
                     return;
                 }
                 if (pendingClosePromise != null) {
@@ -886,9 +886,10 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
                         }
                     }
 
-                    // The promise should be notified before we call fireChannelInactive().
-                    promise.setSuccess();
+                    // The promise should be notified before we call fireChannelInactive(), and the closePromise should
+                    // be notified first for consistency with AbstractChannel so listeners of
                     closePromise.setSuccess();
+                    promise.setSuccess();
 
                     pipeline().fireChannelInactive();
                     if (isRegistered()) {
--- a/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
+++ b/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
@@ -19,6 +19,7 @@ import io.netty.buffer.ByteBufUtil;
 import io.netty.buffer.UnpooledByteBufAllocator;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
+import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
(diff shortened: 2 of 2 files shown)


codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
```diff
@@ -367,6 +367,82 @@ void write(Object msg, ChannelPromise promise) {
...
+        Http2StreamChannel childChannel = (Http2StreamChannel) inboundHandler.channel();
+
+        assertTrue(childChannel.isOpen());
+        assertTrue(childChannel.isActive());
+        childChannel.close().syncUninterruptibly();
+
+        assertFalse(childChannel.isOpen());
+        assertFalse(childChannel.isActive());
+
+        childChannel.close(childChannel.voidPromise()).syncUninterruptibly();
```
bryce-anderson: Just out of curiosity, why is this not allowed? And FWIW, it seems that just closing the `Channel` with its `voidPromise()` is enough to trigger it it: we don't need to close it first. (which is consistent with the behavior of other Channel implementations such as `NioSocketChannel`).

carl-mastrangelo:
cc @zpencer

Scottmitch:
See https://github.com/netty/netty/issues/7641#issuecomment-360953005 for recommended approach regarding this PR (basically we will only take the unit tests).

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -886,9 +886,10 @@ public void close(ChannelPromise promise) {
                         }
                     }
 
-                    // The promise should be notified before we call fireChannelInactive().
-                    promise.setSuccess();
+                    // The promise should be notified before we call fireChannelInactive(), and the closePromise should
+                    // be notified first for consistency with AbstractChannel so listeners of
```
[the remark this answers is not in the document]
Scottmitch: agreed it is 👍 ... however I plan on reverting this portion of the PR in favor of @normanmaurer's PR (see #7641 (comment)) so let me just leave this as is for now and it will go away after rebase.

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
```diff
@@ -367,6 +367,82 @@ void write(Object msg, ChannelPromise promise) {
...
+        LastInboundHandler inboundHandler = streamActiveAndWriteHeaders(inboundStream);
+        Http2StreamChannel childChannel = (Http2StreamChannel) inboundHandler.channel();
+
+        assertTrue(childChannel.isOpen());
+        assertTrue(childChannel.isActive());
+
+        final AtomicBoolean channelOpen = new AtomicBoolean(true);
+        final AtomicBoolean channelActive = new AtomicBoolean(true);
+
+        childChannel.close().addListener(new ChannelFutureListener() {
```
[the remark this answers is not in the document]
Scottmitch: good point ... EmbeddedChannel will execute code in the junit thread.

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
Scottmitch: It is currently disallowed by [AbstractChannelHandlerContext](https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java#L600) and has been this way for a while. We may be able to consider allowing `void` promises (now that they are supported in `AbstractChannel`) but it may break folks.

Scottmitch: the original intention of this test was to demonstrate that the behavior was at least consistent with `AbstractChannel` in this regard because the code is a bit different in than AbstractChannel (which uses [safeSetSuccess](https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/AbstractChannel.java#L695))... but it doesn't provide much value so let me just remove it.

normanmaurer:
@Scottmitch thanks ... I only took the tests, slightly adjusted the commit message and cherry-picked into 4.1 (614b9e0f2)

PR #7646, opened 2018-01-29
Add the NOTICE of the forked portion of Apache Harmony

### LINKED ISSUE #7588: Netty NOTICE doesn't cover Apache Harmony's notice
... which is required by the ASLv2. 

https://www.apache.org/licenses/LICENSE-2.0 (section 4d.)

trustin:
Motivation:

According to the section 4d of ASLv2:

If the Work includes a "NOTICE" text file as part of its distribution, then any Derivative Works that You distribute must include a readable copy of the attribution notices contained within such NOTICE file, excluding those notices that do not pertain to any part of the Derivative Works, in at least one of the following places: within a NOTICE text file distributed as part of the Derivative Works; within the Source form or documentation, if provided along with the Derivative Works; or, within a display generated by the Derivative Works, if and wherever such third-party notices normally appear. The contents of the NOTICE file are for informational purposes only and do not modify the License. You may add Your own attribution notices within Derivative Works that You distribute, alongside or as an addendum to the NOTICE text from the Work, provided that such additional attribution notices cannot be construed as modifying the License.

Modifications:

- Add a subset of NOTICE.txt of Apache Harmony.

Result:

- Fixes #7588

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -61,6 +61,8 @@ facade for Java, which can be obtained at:
 This product contains a modified portion of 'Apache Harmony', an open source
 Java SE, which can be obtained at:
 
+  * NOTICE:
+    * license/NOTICE.harmony.txt
   * LICENSE:
     * license/LICENSE.harmony.txt (Apache License 2.0)
   * HOMEPAGE:
--- a/license/NOTICE.harmony.txt
+++ b/license/NOTICE.harmony.txt
@@ -1,0 +1,6 @@
+Apache Harmony
+
+Copyright 2006, 2010 The Apache Software Foundation.
+
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).


NOTICE.txt
```diff
@@ -62,6 +62,7 @@ This product contains a modified portion of 'Apache Harmony', an open source
 Java SE, which can be obtained at:
 
   * LICENSE:
```
ivankelly: Would be more accurate to put it under a NOTICE: header, but this is fine too.

trustin: Sure. Will update.

johnou: yeah but it's still in the license directory.

trustin: Yeah, but I'm not sure I want to add another top-level directory just for this. Perhaps when we have more of these?

normanmaurer:
Cherry-picked into 4.1 (bda6cb01c6b3f3304b3cb559d6694f130ad190fb)

PR #7648, opened 2018-01-29
Remove unused variable in DefaultHttp2StreamChannel

normanmaurer:
Motivation:

We should remove unused variable (was never read).

Modifications:

Remove unused variable (was never read).

Result:

Cleanup.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
@@ -441,7 +441,6 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
 
         // Keeps track of flush calls in channelReadComplete(...) and aggregate these.
         private boolean inFireChannelReadComplete;
-        private boolean flushPending;
 
         boolean fireChannelReadPending;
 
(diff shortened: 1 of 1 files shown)


normanmaurer:
Let me just merge this as its only cleanup.

normanmaurer:
Cherry-picked into 4.1 (6e6edb59e7)

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -1083,13 +1082,11 @@ public void flush() {
...
-                    // pending. We will ensure we trigger ctx.flush() after we processed all Channels later on and
+                    // If we are currently in the  channelReadComplete(...) call we should just ignore the flush.
+                    // We will ensure we trigger ctx.flush() after we processed all Channels later on and
                     // so aggregate the flushes. This is done as ctx.flush() is expensive when as it may trigger an
                     // write(...) or writev(...) operation on the socket.
-                    if (inFireChannelReadComplete) {
-                        flushPending = true;
-                    } else {
+                    if (!inFireChannelReadComplete) {
                         flush0(ctx);
```
ejona86: This flush looks (pre-existingly) broken. We're currently in `Unsafe` and this calls flush on the child channel. Shouldn't we be flushing the parent channel instead?

Scottmitch: The flush here in the child channel will eventually call flush on the parent channel, if we are not in `channelReadComplete` for the parent. IIUC the rational being that if we are in `channelReadComplete` for the parent then we are in the process of flushing the parent channel.

ejona86: nm. This is in the codec, not the child channel.

PR #7649, opened 2018-01-29
There is no any implementation of ByteBuf throws IndexOutOfBoundsException

kakashiio:
There is no any implementation of ByteBuf throws IndexOutOfBoundsException in resetWriterIndex

Motivation:

Fix the difference between the code and comment.

Modification:

Delete the comment which about throw IndexOutOfBoundsException on the method ByteBuf#resetWriterIndex 

Result:

Delete the comment which about throw IndexOutOfBoundsException on the method  ByteBuf#resetWriterIndex to make the comment has the same meaning with the code.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/ByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/ByteBuf.java
@@ -487,10 +487,6 @@ public abstract class ByteBuf implements ReferenceCounted, Comparable<ByteBuf> {
     /**
      * Repositions the current {@code writerIndex} to the marked
      * {@code writerIndex} in this buffer.
-     *
-     * @throws IndexOutOfBoundsException
-     *         if the current {@code readerIndex} is greater than the marked
-     *         {@code writerIndex}
      */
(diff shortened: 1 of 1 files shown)


normanmaurer:
@trustin I wonder if we should fix the code or the java docs ;) WDYT ?

normanmaurer:
@Scottmitch @ejona86 WDYT ?

ejona86:
Fixing the code seems better to me, as it seems an ugly state to get into. I also don't see any reason why it shouldn't be any different than `resetReaderIndex()`.

It appears `resetReaderIndex()` does throw today. It seems `AbstractByteBuf` should mirror what is done for reader index:
```diff
     @Override
     public ByteBuf resetWriterIndex() {
-        writerIndex = markedWriterIndex;
+        writerIndex(markedWriterIndex);
         return this;
     }
```

kakashiio:
That makes the code more robust ! Thanks !

normanmaurer:
Let me close this as there is another PR which fixes the impl.

Issue #7254, opened 2017-09-27
Illegal reflective access warnings when running Netty with Java 9


doom369:
### Expected behavior

No warnings.

### Actual behavior

Warnings when starting netty with java 9:

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by io.netty.util.internal.ReflectionUtil (file:/home/doom369/.m2/repository/io/netty/netty-common/4.1.16.Final/netty-common-4.1.16.Final.jar) to constructor java.nio.DirectByteBuffer(long,int)
WARNING: Please consider reporting this to the maintainers of io.netty.util.internal.ReflectionUtil
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

with `--illegal-access=warn` enabled :

```
WARNING: Illegal reflective access by io.netty.util.internal.ReflectionUtil (file:/home/doom369/.m2/repository/io/netty/netty-common/4.1.16.Final/netty-common-4.1.16.Final.jar) to constructor java.nio.DirectByteBuffer(long,int)
WARNING: Illegal reflective access by io.netty.util.internal.ReflectionUtil (file:/home/doom369/.m2/repository/io/netty/netty-common/4.1.16.Final/netty-common-4.1.16.Final.jar) to method java.nio.Bits.unaligned()
```

### Steps to reproduce

Run any netty server with Java 9.

### Netty version

4.1.16.Final

### JVM version (e.g. `java -version`)

```
java version "9"
Java(TM) SE Runtime Environment (build 9+181)
Java HotSpot(TM) 64-Bit Server VM (build 9+181, mixed mode)
```

Maybe it would be better to do not access above classes in case of java 9 to avoid warning?

rockwotj:
Can we re-open this to actually solve the issue, I'd hate for things to break later on because this never got fixed?

The warning is quite clear this will break later on.

doom369:
@rockwotj Actually when reflection access will be removed from Java 9 - netty will be ready. As fixed was already applied.

Netty was tested on EA builds were reflective access was removed and right now netty handles this case.

rockwotj:
I see what version of netty has these changes?

doom369:
It was many months ago - https://github.com/netty/netty/issues/6345

alsnkline:
This seems strange to be open and closed at that same time.
I'm getting these errors also and would love to know when/how to upgrade Netty to get rid of them.

[...]

PR #7650, opened 2018-01-29
Reflective setAccessible(true) will produce scary warnings on the con…

normanmaurer:
…sole when using java9+, dont do it

Motivation:

Reflective setAccessible(true) will produce scary warnings on the console when using java9+, while netty still works. That said users may feel uncomfortable with these warnings, we should not try to do it by default when using java9+.

Modifications:

Add io.netty.noReflectiveAccessible system property which controls if setAccessible(...) will be used. By default it will bet set to true when using java9+.

Result:

Fixes [#7254].

### DIFF
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent0.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent0.java
@@ -46,6 +46,8 @@ final class PlatformDependent0 {
 
     private static final Throwable UNSAFE_UNAVAILABILITY_CAUSE;
     private static final Object INTERNAL_UNSAFE;
+    private static final boolean IS_EXPLICIT_TRY_REFLECTION_SET_ACCESSIBLE = explicitTryReflectionSetAccessible0();
+
     static final Unsafe UNSAFE;
 
     // constants borrowed from murmur3
@@ -84,7 +86,9 @@ final class PlatformDependent0 {
                 public Object run() {
                     try {
                         final Field unsafeField = Unsafe.class.getDeclaredField("theUnsafe");
-                        Throwable cause = ReflectionUtil.trySetAccessible(unsafeField);
+                        // We always want to try using Unsafe as the access still works on java9 as well and
+                        // we need it for out native-transports and many optimizations.
+                        Throwable cause = ReflectionUtil.trySetAccessible(unsafeField, false);
                         if (cause != null) {
                             return cause;
                         }
@@ -218,7 +222,7 @@ final class PlatformDependent0 {
                                 try {
                                     final Constructor<?> constructor =
                                             direct.getClass().getDeclaredConstructor(long.class, int.class);
-                                    Throwable cause = ReflectionUtil.trySetAccessible(constructor);
+                                    Throwable cause = ReflectionUtil.trySetAccessible(constructor, true);
                                     if (cause != null) {
                                         return cause;
                                     }
@@ -267,7 +271,7 @@ final class PlatformDependent0 {
                         Class<?> bitsClass =
                                 Class.forName("java.nio.Bits", false, getSystemClassLoader());
                         Method unalignedMethod = bitsClass.getDeclaredMethod("unaligned");
-                        Throwable cause = ReflectionUtil.trySetAccessible(unalignedMethod);
+                        Throwable cause = ReflectionUtil.trySetAccessible(unalignedMethod, true);
                         if (cause != null) {
                             return cause;
                         }
@@ -795,6 +799,15 @@ final class PlatformDependent0 {
             logger.debug("Platform: Android");
         }
         return android;
+    }
+
+    private static boolean explicitTryReflectionSetAccessible0() {
+        // we disable reflective access
+        return SystemPropertyUtil.getBoolean("io.netty.tryReflectionSetAccessible", javaVersion() < 9);
+    }
+
+    static boolean isExplicitTryReflectionSetAccessible() {
+        return IS_EXPLICIT_TRY_REFLECTION_SET_ACCESSIBLE;
     }
 
     static int javaVersion() {
--- a/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2FrameCodecTest.java
+++ b/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2FrameCodecTest.java
@@ -719,7 +719,7 @@ public class Http2FrameCodecTest {
                 UpgradeEvent.class.getDeclaredConstructor(CharSequence.class, FullHttpRequest.class);
 
         // Check if we could make it accessible which may fail on java9.
-        Assume.assumeTrue(ReflectionUtil.trySetAccessible(constructor) == null);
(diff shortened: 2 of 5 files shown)


normanmaurer:
@doom369 FYI...

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -84,7 +86,9 @@
                 public Object run() {
                     try {
                         final Field unsafeField = Unsafe.class.getDeclaredField("theUnsafe");
-                        Throwable cause = ReflectionUtil.trySetAccessible(unsafeField);
+                        // We always want to try using Unsafe as the access still works on java9 as well and
+                        // we need it for out native-transports and many optimizations.
```
johnou: @normanmaurer have you looked into MethodHandles.byteBufferViewVarHandle [1] that creates a VarHandle that allows to do aligned/unaligned access on a ByteBuffer?

[1] http://download.java.net/java/jdk9/docs/api/java/lang/invoke/MethodHandles.html#byteBufferViewVarHandle-java.lang.Class-java.nio.ByteOrder-

http://mail.openjdk.java.net/pipermail/jigsaw-dev/2017-February/011477.html

doom369:
@normanmaurer 

`java.nio.DirectByteBuffer(long,int)` and `java.nio.Bits.unaligned()` both are not available under Java 9. So maybe it would be better just to skip reflection access try for those particular methods when running java9 instead of adding the additional property?

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
normanmaurer: @johnou we are compiling against java8 so it will not help at all (and most people not use java9 yet).

normanmaurer:
@doom369 the property will be "true" by default when using java9 and so we will skip it, or I am missing anything? The thing is you can open it up if you like to and so get the extra perf / features. Thats why I would love to at least still allow this.

doom369:
@normanmaurer I see... In case someone manually (via jvm option) will open access to jdk internal modules this will work. Doesn't seem like real case to me, but yeah. Nice to have.

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
johnou: @normanmaurer perhaps for Netty 5 then given that Java 8 and 9 will be EOL this year unless you fork out cash http://www.oracle.com/technetwork/java/eol-135779.html

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -797,6 +801,15 @@ private static boolean isAndroid0() {
         return android;
     }
 
+    private static boolean explicitNoReflectiveAccessible0() {
+        // we disable reflective access
+        return SystemPropertyUtil.getBoolean("io.netty.noReflectiveAccessible", javaVersion() >= 9);
```
Scottmitch: can we avoid the negative "no" in the name? IMHO it makes it more difficult to have to negate the logic to understand what is going on. What about `io.netty.tryReflectionSetAccessible`?

normanmaurer: @Scottmitch sure... I just used it because we use it for unsafe as well

normanmaurer:
@Scottmitch PTAL again

normanmaurer:
@doom369 I think it is definitely a real-case if you want max perf.

ejona86:
CC @zpencer

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -46,6 +46,8 @@
 
     private static final Throwable UNSAFE_UNAVAILABILITY_CAUSE;
     private static final Object INTERNAL_UNSAFE;
+    private static final boolean IS_EXPLICIT_TRY_REFLECTIVION_SET_ACCESSIBLE = explicitTryReflectionSetAccessible0();
```
zpencer: nit: typo in REFLECTION/REFLECTIVE

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -84,7 +86,9 @@
                 public Object run() {
                     try {
                         final Field unsafeField = Unsafe.class.getDeclaredField("theUnsafe");
-                        Throwable cause = ReflectionUtil.trySetAccessible(unsafeField);
+                        // We always want to try using Unsafe as the access still works on java9 as well and
+                        // we need it for out native-transports and many optimizations.
+                        Throwable cause = ReflectionUtil.trySetAccessible(unsafeField, false);
```
zpencer: Would the scary sounding message still be logged here for jdk9? Perhaps we can log a message for the user in addition to the code comments here.

normanmaurer: Nope... access to unsafe is handled special and so nothing is logged

normanmaurer:
Cherry-picked into 4.1 (e72c197aa3a88e97c10793ee7870775ab8807f92) and 4.0 (c62345f3a9)

normanmaurer:
Yep with JNI you can do whatever you want 👌😉

eolivelli:
@normanmaurer 
just to double check,
after upgrading Netty to this version (4.1.21Final) is it possible to have worse performance on Java 9 without setting the system property an re-allow reflective access ?

normanmaurer:
@eolivelli yes... that said there is not much we can do about it :(

eolivelli:
@normanmaurer clear, thank you

eolivelli:
@normanmaurer  last thought: can you consider adding a note here https://netty.io/news/2018/02/05/4-0-56-Final-4-1-21-Final.html I think this is an important flag to set and people may see degradation in performances.

slandelle:
@normanmaurer Wouldn't it make sense to ask the JCP for a public API for what you're doing?

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
johnou: @slandelle like this suggestion or is the public API not rich enough?

artash:
Just a side note here, I think this change _may_ affect the RSS of your java process. Your direct buffers will now be allocated with `ByteBuffer.allocateDirect()` and this will not only allocate the memory but also set it to zero, i.e. it'll call:
```
            base = UNSAFE.allocateMemory(size);
``` 
followed by 
```
        UNSAFE.setMemory(base, size, (byte) 0);
```
 whereas in the past netty would only call `Unsafe.allocateMemory()` and not set the memory.  This means the memory would be committed but won't be resident (in physical RAM) until it's written to. 

In other words the new version eagerly uses all the memory allocated and this may definitely have an impact in your free memory graphs.

itsTyrion:
> @eolivelli yes... that said there is not much we can do about it :(

Any, even vague, numbers? 1%, 5%, 10%, 20%?

itsTyrion:
@normanmaurer

PR #7653, opened 2018-01-30
Modify the `resetWriterIndex`. There is no reason to implements by assigning directly instead of invoking `writerIndex`.

kakashiio:
Modify the `resetWriterIndex`. There is no reason to implements by assigning directly instead of invoking `writerIndex`.

Motivation:

If someone invoke `writeByte()`,  `markWriterIndex()`, `readByte()` in order first, and then invoke `resetWriterIndex()` should be throw a `IndexOutOfBoundsException` to obey the rule that the buffer declared "0      <=      readerIndex   <=   writerIndex    <=    capacity".

Modification:

Changed the code `writerIndex = markedWriterIndex;` into `writerIndex(markedWriterIndex);` to make the check affect

Result:
Throw IndexOutOfBoundsException if any invalid happened in resetWriterIndex.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
@@ -167,7 +167,7 @@ public abstract class AbstractByteBuf extends ByteBuf {
 
     @Override
     public ByteBuf resetWriterIndex() {
-        writerIndex = markedWriterIndex;
+        writerIndex(markedWriterIndex);
         return this;
     }
 


johnou:
I wouldn't say no reason, by assigning directly the bounds check is skipped which could be a significant performance gain. In general when using mark and skip I wouldn't expect it to throw an IOOBE unless you are changing the size of the bytebuf?

normanmaurer:
@johnou actually I think this patch is correct and also reflects what is documented in the javadocs. That said a test-case would be nice 👍

normanmaurer:
@kakashiio ping

normanmaurer:
@kakashiio ping... can you add the tests ?

kakashiio:
@normanmaurer Okay,let me try tomorrow. I have a lot of work to do in the past several days.

normanmaurer:
Thanks a lot 👍

normanmaurer:
Thanks.. Merged with https://github.com/netty/netty/pull/7750 and opened https://github.com/netty/netty/pull/7751

Issue #7329, opened 2017-10-23
Expiry strategy improvement request for DefaultDnsCache


jianyuanzh:
### Expected behavior
Hostname configured with both IPv4 and IPv6 dns records. Netty dns resolver is configured to prefer IPv4 addresses, so I should always get IPv4 address.

### Actual behavior
but after several minutes run, it can only return ipv6 record.

### Steps to reproduce
Create a NettyDns resolver with DefaultDnsCache, and keep it running for several minutes. You can see finally only ipv6 record can be got.

### Minimal yet complete reproducer code (or URL to code)
the url have the issue is "mirrors.servercentral.com"

### Netty version
4.1.15.Final

### JVM version (e.g. `java -version`)
```
java -version
java version "1.8.0_131"
Java(TM) SE Runtime Environment (build 1.8.0_131-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.131-b11, mixed mode)
```
### OS version (e.g. `uname -a`)
```
uname -a
Darwin zhangjianyuans-MacBook-Pro.local 16.7.0 Darwin Kernel Version 16.7.0: Thu Jun 15 17:36:27 PDT 2017; root:xnu-3789.70.16~2/RELEASE_X86_64 x86_64
```

jianyuanzh:
The root cause for this issue if the expire strategy of the DefaultDnsCache. The DefaultDnsCache controls the expiry per record, and remove the record from the list if the record is expired. In this case, the ipv4 record ttl is 60, while the ipv6 record ttl is 21599. 

So, ipv4 record will surely be expired before ipv6 record. And ipv6 record is still recorded in DefaulDnsCache, so query the DefaultDnsCache can still get the record for the URL with the ipv6 record.  

But as mentioned above, we are expecting the ipv4 record.

normanmaurer:
@jianyuanzh hmm I think what you describe is actually what I would expect. Can you show me how exactly your build the `DnsNameResolver` ?

jianyuanzh:
@normanmaurer , here is my demo script to check the TTL for DNS record:
```
String dns = "mirrors.servercentral.com.";

EventLoopGroup loopGroup = new NioEventLoopGroup(1)

StringBuilder sb = new StringBuilder()
DnsNameResolver resolver = new DnsNameResolverBuilder(loopGroup.next())
        .channelType(NioDatagramChannel.class)
        .resolvedAddressTypes(ResolvedAddressTypes.IPV4_PREFERRED)
[...]

PR #7656, opened 2018-01-30
DefaultDnsCache should expire all records per hostname when one TTL i…

normanmaurer:
…s reached.

Motivation:

At the moment DefaultDnsCache will expire each record dependong on its own TTL. This may result in unexpected results for the end-user especially if the user for example uses IPV4_PREFERED but the cached AAAA records has a higher TTL then the A records and so the A record was removed. In this case we would only return the AAAA record and not even try to refresh.

Modifications:

Always expire all records for a hostname when one TTL is reached.

Result:

Fixes [#7329]

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
@@ -104,7 +104,7 @@ public class DefaultDnsCache implements DnsCache {
                 Map.Entry<String, Entries> e = i.next();
                 i.remove();
 
-                e.getValue().cancelExpiration();
+                e.getValue().clearAndCancel();
             }
         }
     }
@@ -113,7 +113,7 @@ public class DefaultDnsCache implements DnsCache {
     public boolean clear(String hostname) {
         checkNotNull(hostname, "hostname");
         Entries entries = resolveCache.remove(hostname);
-        return entries != null && entries.cancelExpiration();
+        return entries != null && entries.clearAndCancel();
     }
 
     private static boolean emptyAdditionals(DnsRecord[] additionals) {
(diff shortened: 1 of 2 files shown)


normanmaurer:
Also @slandelle

normanmaurer:
And @jianyuanzh

jianyuanzh:
Thanks @normanmaurer for the fix

trustin:
Then we don't really schedule an expiration task for every record. Only one task per hostname should be enough.

Given such an optimization will make this PR non-trivial, it's up to you to address this comment in a separate PR. What do you think?

normanmaurer:
@trustin I think I would do a followup pr once this one is merged as it not makes things worse in terms of scheduling. This will make things easier for me and easier to review. WDYT ? Is the rest of the PR ok ?

normanmaurer:
@jianyuanzh does it solve the problem for you ?

@slandelle any concerns ?

slandelle:
@normanmaurer LGTM

normanmaurer:
ok let me merge this one then.

normanmaurer:
Cherry-picked into 4.1 (b874edbf65e146cefd0f769767c587931e9f2552)

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -171,18 +171,27 @@ private void cache0(DefaultDnsCacheEntry e, int ttl, EventLoop loop) {
...
+    private void scheduleCacheExpiration(final DefaultDnsCacheEntry e,
                                          int ttl,
                                          EventLoop loop) {
         e.scheduleExpiration(loop, new Runnable() {
                     @Override
                     public void run() {
-                        if (entries.remove(e)) {
-                            resolveCache.remove(e.hostname);
+                        // We always remove all entries for a hostname once one entry expire. This is not the
+                        // most efficient to do but this way we can guarantee that if a DnsResolver
```
Scottmitch: what about keeping the cache purge behavior consistent, but triggering background queries as necessary (on demand, periodically, etc...) to refresh the expired entries? What is the DNS resolver's behavior in this circumstance?

normanmaurer: We may be able to optimize it but for now I think this fix is ok.... we may want to think more about how the whole cache is implemented and what informations are provided to it tho

PR #7658, opened 2018-01-30
ByteBufUtil.isText method should be safe to be called concurrently

normanmaurer:
Motivation:

ByteBufUtil.isText(...) may produce unexpected results if called concurrently on the same ByteBuffer.

Modifications:

- Don't use internalNioBuffer where it is not safe.
- Add unit test.

Result:

ByteBufUtil.isText is thread-safe.


normanmaurer:
I decided to remove the changes to `FixedCompositeByteBuf` and `CompositeByteBuf` in this PR. Will open another PR for these which also includes tests.

normanmaurer:
Cherry-picked into 4.1 (d2bd36fc4c9e200e7371c487b95b761bc9f3cd27) and 4.0 (f05c45e65d12e45a38a36467759b4f3f65b7bae8)

PR #7661, opened 2018-01-31
Add the threshold control whether the HTTP response should be compressed.

### LINKED ISSUE #7660: io.netty.handler.codec.http.HttpContentCompressor need compress threshold
### Expected behavior
the response body is compressed, when the response body exceeds the threshold
### Actual behavior
no threshold can be set
### Steps to reproduce
-
-
### Netty version
4.1.16.Final
java 1.8
-

teaey:
…mpressed.

Motivation:

Add the threshold control whether the HTTP response should be compressed.

Modifications:

Add filed io.netty.handler.codec.http.HttpContentCompressor#compressThreshold that control whether the HTTP response should be compressed.

Result:

Fixes #7660.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
@@ -32,6 +32,7 @@ public class HttpContentCompressor extends HttpContentEncoder {
     private final int compressionLevel;
     private final int windowBits;
     private final int memLevel;
+    private final int contentSizeThreshold;
     private ChannelHandlerContext ctx;
 
     /**
@@ -52,7 +53,7 @@ public class HttpContentCompressor extends HttpContentEncoder {
      *        compression level is {@code 6}.
      */
     public HttpContentCompressor(int compressionLevel) {
-        this(compressionLevel, 15, 8);
+        this(compressionLevel, 15, 8, 0);
     }
 
     /**
@@ -75,6 +76,33 @@ public class HttpContentCompressor extends HttpContentEncoder {
      *        at the expense of memory usage.  The default value is {@code 8}
      */
     public HttpContentCompressor(int compressionLevel, int windowBits, int memLevel) {
+        this(compressionLevel, windowBits, memLevel, 0);
+    }
+
+    /**
+     * Creates a new handler with the specified compression level, window size,
+     * and memory level..
+     *
+     * @param compressionLevel
+     *        {@code 1} yields the fastest compression and {@code 9} yields the
+     *        best compression.  {@code 0} means no compression.  The default
+     *        compression level is {@code 6}.
+     * @param windowBits
+     *        The base two logarithm of the size of the history buffer.  The
+     *        value should be in the range {@code 9} to {@code 15} inclusive.
+     *        Larger values result in better compression at the expense of
+     *        memory usage.  The default value is {@code 15}.
+     * @param memLevel
+     *        How much memory should be allocated for the internal compression
+     *        state.  {@code 1} uses minimum memory and {@code 9} uses maximum
+     *        memory.  Larger values result in better and faster compression
+     *        at the expense of memory usage.  The default value is {@code 8}
+     * @param contentSizeThreshold
+     *        The response body is compressed when the size of the response
+     *        body exceeds the threshold. The value should be a non negative
+     *        number. {@code 0} will enable compression for all responses.
+     */
+    public HttpContentCompressor(int compressionLevel, int windowBits, int memLevel, int contentSizeThreshold) {
         if (compressionLevel < 0 || compressionLevel > 9) {
             throw new IllegalArgumentException(
                     "compressionLevel: " + compressionLevel +
@@ -88,9 +116,14 @@ public class HttpContentCompressor extends HttpContentEncoder {
             throw new IllegalArgumentException(
                     "memLevel: " + memLevel + " (expected: 1-9)");
         }
+        if (contentSizeThreshold < 0) {
+            throw new IllegalArgumentException(
+                    "contentSizeThreshold: " + contentSizeThreshold + " (expected: non negative number)");
+        }
         this.compressionLevel = compressionLevel;
         this.windowBits = windowBits;
         this.memLevel = memLevel;
+        this.contentSizeThreshold = contentSizeThreshold;
     }
 
     @Override
@@ -100,6 +133,13 @@ public class HttpContentCompressor extends HttpContentEncoder {
 
     @Override
     protected Result beginEncode(HttpResponse headers, String acceptEncoding) throws Exception {
+        if (this.contentSizeThreshold > 0) {
+            if (headers instanceof HttpContent &&
+                    ((HttpContent) headers).content().readableBytes() < contentSizeThreshold) {
+                return null;
+            }
+        }
+
         String contentEncoding = headers.headers().get(HttpHeaderNames.CONTENT_ENCODING);
         if (contentEncoding != null) {
             // Content-Encoding was set, either as something specific or as the IDENTITY encoding
--- a/codec-http/src/test/java/io/netty/handler/codec/http/HttpContentCompressorTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/HttpContentCompressorTest.java
@@ -29,11 +29,7 @@ import static org.hamcrest.CoreMatchers.instanceOf;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.not;
 import static org.hamcrest.CoreMatchers.nullValue;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertSame;
-import static org.junit.Assert.assertThat;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
 
 public class HttpContentCompressorTest {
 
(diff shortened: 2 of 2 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
```diff
@@ -33,6 +33,7 @@
     private final int windowBits;
     private final int memLevel;
     private ChannelHandlerContext ctx;
+    private final int compressThreshold;
```
normanmaurer: nit: can you move this to the other final variables.

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
```diff
@@ -52,7 +53,7 @@ public HttpContentCompressor() {
      *        compression level is {@code 6}.
      */
     public HttpContentCompressor(int compressionLevel) {
-        this(compressionLevel, 15, 8);
+        this(compressionLevel, 15, 8,0);
```
normanmaurer: add whitespace after `,`

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
```diff
@@ -73,8 +74,11 @@ public HttpContentCompressor(int compressionLevel) {
      *        state.  {@code 1} uses minimum memory and {@code 9} uses maximum
      *        memory.  Larger values result in better and faster compression
      *        at the expense of memory usage.  The default value is {@code 8}
+     * @param compressThreshold
+     *        The response body is compressed,
+     *        when the response body exceeds the threshold
      */
-    public HttpContentCompressor(int compressionLevel, int windowBits, int memLevel) {
+    public HttpContentCompressor(int compressionLevel, int windowBits, int memLevel, int compressThreshold) {
```
normanmaurer: We need to keep the existing constructor to not break the API. Please just create an extra constructor and delegate to it.

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
```diff
@@ -73,8 +74,11 @@ public HttpContentCompressor(int compressionLevel) {
      *        state.  {@code 1} uses minimum memory and {@code 9} uses maximum
      *        memory.  Larger values result in better and faster compression
      *        at the expense of memory usage.  The default value is {@code 8}
+     * @param compressThreshold
+     *        The response body is compressed,
+     *        when the response body exceeds the threshold
```
normanmaurer: Explain that `<= 0` will disable this.

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
```diff
@@ -107,6 +112,12 @@ protected Result beginEncode(HttpResponse headers, String acceptEncoding) throws
             return null;
         }
 
+        if(this.compressThreshold > 0) {
```
normanmaurer: add space after `if`

teaey:
@normanmaurer  I suppose using strategy pattern which is better.
I will introduced a method eg:

```java
boolean needCompress(Response response);
```

normanmaurer:
@teaey please add a unit test as well.

teaey:
@normanmaurer Two methods are added to the HttpContentCompressorTest.

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
```diff
@@ -131,6 +164,18 @@ protected Result beginEncode(HttpResponse headers, String acceptEncoding) throws
                         wrapper, compressionLevel, windowBits, memLevel)));
     }
 
+    /**
+     * For customization should implement this method that
+     * performs the necessary business logic.
```
normanmaurer: @teaey please reword to:

Returns {@code true} if the compression should be used. May be override to change default behaviour

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
```diff
@@ -131,6 +164,18 @@ protected Result beginEncode(HttpResponse headers, String acceptEncoding) throws
...
+     */
+    protected boolean needCompress(HttpResponse response) {
```
normanmaurer: rename to `useCompression(...)`

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
```diff
@@ -75,6 +76,33 @@ public HttpContentCompressor(int compressionLevel) {
...
+     *        Larger values result in better compression at the expense of
+     *        memory usage.  The default value is {@code 15}.
+     * @param memLevel
+     *        How much memory should be allocated for the internal compression
+     *        state.  {@code 1} uses minimum memory and {@code 9} uses maximum
+     *        memory.  Larger values result in better and faster compression
+     *        at the expense of memory usage.  The default value is {@code 8}
+     * @param compressThreshold
+     *        The response body is compressed,
+     *        If the response body exceeds the threshold.
```
normanmaurer: nit: s/If/if/

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
```diff
@@ -75,6 +76,33 @@ public HttpContentCompressor(int compressionLevel) {
...
+     *        Default value is {@code 0} that means all responses will be compressed.
```
normanmaurer: change to: `{@code 0}` will enable compression for all responses.

teaey:
@normanmaurer All changed.

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
```diff
@@ -107,6 +136,10 @@ protected Result beginEncode(HttpResponse headers, String acceptEncoding) throws
             return null;
         }
 
+        if (!useCompression(headers)) {
```
normanmaurer: @teaey sorry one last nit... I think this should be the first thing to check as we not need to even compare headers if it is false. Please move it before the header check

teaey:
@normanmaurer I think so.

normanmaurer:
@ejona86 @Scottmitch @carl-mastrangelo @trustin WDYT ?

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
```diff
@@ -131,6 +164,18 @@ protected Result beginEncode(HttpResponse headers, String acceptEncoding) throws
                         wrapper, compressionLevel, windowBits, memLevel)));
     }
 
+    /**
+     * Returns {@code true} if the compression should be used.
+     * May be override to change default behaviour.
+     */
+    protected boolean useCompression(HttpResponse response) {
+        if (this.compressThreshold > 0) {
+            return response instanceof HttpContent &&
```
ejona86: This looks like it only works properly for a FullHttpResponse. If a HttpResponse was sent separately than a HttpContent(s), then this won't compress. Is that intended?

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
```diff
@@ -75,6 +76,33 @@ public HttpContentCompressor(int compressionLevel) {
...
+     *        The base two logarithm of the size of the history buffer.  The
+     *        value should be in the range {@code 9} to {@code 15} inclusive.
+     *        Larger values result in better compression at the expense of
+     *        memory usage.  The default value is {@code 15}.
+     * @param memLevel
+     *        How much memory should be allocated for the internal compression
+     *        state.  {@code 1} uses minimum memory and {@code 9} uses maximum
+     *        memory.  Larger values result in better and faster compression
+     *        at the expense of memory usage.  The default value is {@code 8}
+     * @param compressThreshold
```
ejona86: sizeThreshold? payloadSizeThreshold? compressThreshold just makes me question what it means (e.g., threshold of compression ratio?).

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
```diff
@@ -32,6 +32,7 @@
     private final int compressionLevel;
     private final int windowBits;
     private final int memLevel;
+    protected final int compressThreshold;
```
ejona86: Why is this protected?

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
```diff
@@ -91,6 +119,7 @@ public HttpContentCompressor(int compressionLevel, int windowBits, int memLevel)
         this.compressionLevel = compressionLevel;
         this.windowBits = windowBits;
         this.memLevel = memLevel;
+        this.compressThreshold = (compressThreshold < 0) ? 0 : compressThreshold;
```
ejona86: I'd feel better with throwing if negative.

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
teaey: Hi @ejona86. 

Users can implement the subclass to overwrite the useCompression method, which may access the compressThreshold.

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
teaey: Yes, this is the same as other fields.
@normanmaurer What's your opinion?

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
teaey: 👍 
What about contentSizeThreshold?

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
teaey: This is in line with the original class.

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
ejona86: SGTM

teaey:
@ejona86 Please review again.

ejona86:
@teaey, oh, I had been waiting for a reply on one of my comments. Since the most recent commit wasn't done in a separate commit I couldn't easily tell what changes you made.

normanmaurer:
Cherry-picked into 4.1 (06dcca1dbcc048b6ce7d954e4ecdf1d3ed6af034).

@teaey thanks!

Issue #7662, opened 2018-01-31
Proposal: add removeIfExists() method to ChannelPipeline


doom369:
In some cases, I have to remove the handler from the pipeline that may exist or may not exists in the pipeline. However, current Netty `remove()` implementation always throws `NoSuchElementException` that requires construction like this:

```
try {
    pipeline.remove(LetsEncryptHandler.class);
} catch (NoSuchElementException e) {
}
```

this could be simplified with:

`pipeline.removeIfExists(LetsEncryptHandler.class);`

WDYT?

albertok:
This would be very nice to have.
I've been using:
`if (pipeline.names.contains("someHandler")`

PR #7663, opened 2018-01-31
#7662 Proposal: add removeIfExists() method to DefaultChannelPipeline

doom369:
Motivation:

Sometimes it is very convenient to remove the handler from pipeline without throwing the exception in case those handler doesn't exist in the pipeline.

Modification:

Added 3 overloaded methods to `DefaultChannelPipeline`, but **not added** to `ChannelHandler` due to back compatibility.

Result:

Fixes https://github.com/netty/netty/issues/7662


transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -457,6 +457,26 @@ public final ChannelHandler remove(String name) {
...
+    public final <T extends ChannelHandler> T removeIfExists(ChannelHandler handler) {
+        return removeIfExists(context(handler));
+    }
+
+    @SuppressWarnings("unchecked")
+    private <T extends ChannelHandler> T removeIfExists(ChannelHandlerContext ctx) {
+        if (ctx == null) {
+            return null;
+        }
+        return (T) remove((AbstractChannelHandlerContext) ctx).handler();
```
normanmaurer: @doom369 I think this is not safe as in the meantime it could have been removed from another thread which would have this method throw a `NoSuchElementException`

doom369: @normanmaurer I don't see any place where NoSuchElementException could be thrown in above code.

normanmaurer: ah never mind... this should work...

normanmaurer:
What do others think ?

ejona86:
This seems fine. Not necessary, but a fine convenience. This is very unsurprising given the current remove throws if not found.

Issue #7639, opened 2018-01-26
Http2ConnectionHandler shutsdown channel too early when using SSL


carl-mastrangelo:
### Expected behavior
Connection read related failures should be noticed before buffered writes.

 
### Actual behavior
Channel is closed by failing buffered writes too earl

### Steps to reproduce.
https://github.com/grpc/grpc-java/issues/3991

In this issue, gRPC is adding an SSL handler followed by an Http2ConnectionHandler.   When added, the Http2ConnectionHandler writes the preface which is buffered by the SslHandler.   In the issue linked, the certificate sent by the server was incorrect and the client tries to handle the error.  The exact call sequence is 

1.  `SslHandler.decodeJdkCompatible`
2.  `SslHandler.handleUnwrapThrowable`
3.  `SslHandler.wrapAndFlush`
4.  `SslHandler.wrap`
5.  `promise.tryFailure(SSLENGINE_CLOSED);`

This goes through several promises and eventually fails the Write promise for the HTTP/2 preface sent.   Unfortunately, the write looks like:

```
ctx.write(connectionPrefaceBuf()).addListener(ChannelFutureListener.CLOSE_ON_FAILURE);
```

(We are still in the the scope of the `read`, remember).    This promise is failed, which closes the channel, which then ends all further processing by gRPC because we ignore errors after the channel has closed.    I believe this is wrong, because the `SslHandshakeCompletionEvent` get trigger _after_ the channel has been closed down, which is not the correct order.

As a work around, gRPC now adds the HTTP/2 handler after Ssl is established in order for the correct Ssl Exception to be returned.

As for a fix, I'm open to ideas.    My first guess was to make the `Http2ConnectionHandler` not use `CLOSE_ON_FAILURE`, but instead add an executable to the event loop to close the channel.  This would allow the read() to fail properly, followed by closing the connection.  @ejona86 points out that this may not work, because other write failure listeners may close the connection down early too.   I still think this is likely the best course of action.  

### Netty version
4.1.17

ejona86:
> My first guess was to make the Http2ConnectionHandler not use CLOSE_ON_FAILURE, but instead add an executable to the event loop to close the channel.

I brought up that other things related to this handling _may_ also be putting things on the event loop (or putting them there dependent of the reentrancy counter), such that the channel.close() still happens before they are processed. I didn't look into it at all.

Scottmitch:
[...]

PR #7667, opened 2018-01-31
SslHandler unwrap out of order promise/event notificaiton

Scottmitch:
Motivation:
SslHandler#decode methods catch any exceptions and attempt to wrap
before shutting down the engine. The intention is to write any alerts
which the engine may have pending. However the wrap process may also
attempt to write user data, and may also complete the associated
promises. If this is the case, and a promise listener closes the channel
then SslHandler may later propagate a SslHandshakeCompletionEvent user
event through the pipeline. Since the channel has already been closed
the user may no longer be paying attention to user events.

Modifications:
- Sslhandler#decode should first fail the associated handshake promise
and propagate the SslHandshakeCompletionEvent before attempting to wrap

Result:
Fixes https://github.com/netty/netty/issues/7639

### DIFF
--- a/handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
@@ -22,6 +22,8 @@ import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.Unpooled;
 import io.netty.channel.Channel;
+import io.netty.channel.ChannelFuture;
+import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
@@ -29,9 +31,13 @@ import io.netty.channel.ChannelInitializer;
 import io.netty.channel.ChannelOutboundHandlerAdapter;
 import io.netty.channel.ChannelPromise;
 import io.netty.channel.DefaultChannelId;
+import io.netty.channel.DefaultEventLoopGroup;
 import io.netty.channel.EventLoopGroup;
 import io.netty.channel.SimpleChannelInboundHandler;
 import io.netty.channel.embedded.EmbeddedChannel;
+import io.netty.channel.local.LocalAddress;
+import io.netty.channel.local.LocalChannel;
+import io.netty.channel.local.LocalServerChannel;
 import io.netty.channel.nio.NioEventLoopGroup;
 import io.netty.channel.socket.nio.NioServerSocketChannel;
 import io.netty.channel.socket.nio.NioSocketChannel;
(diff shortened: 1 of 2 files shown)


Scottmitch:
@carl-mastrangelo - can you verify this PR fixes the original issue?

carl-mastrangelo:
@Scottmitch Unfortunately I can't test it out.  Between 4.1.17 and 4.1.20 Netty stopped working for gRPC (tests fail plus https://github.com/grpc/grpc-java/issues/4025 ).   I haven't had a chance to update gRPC for this usage.   (too many Netty PRs in my backlog ;) )

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
```diff
@@ -431,6 +437,97 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exc
...
+
+            SslCompletionEvent evt = (SslCompletionEvent) events.take();
+            assertTrue(evt instanceof SslHandshakeCompletionEvent);
+            assertThat(evt.cause(), is(instanceOf(SSLException.class)));
+
+            ChannelFuture future = (ChannelFuture) events.take();
+            assertThat(future.cause(), is(instanceOf(SSLException.class)));
+
+            serverChannel.close();
+            clientChannel.close();
```
normanmaurer: @Scottmitch nit: also null these out so we not try to close again in the finally block.

normanmaurer: Also you may want to call `sync()`, not needed but still not bad I think

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
```diff
@@ -431,6 +437,97 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exc
...
+                .channel(LocalServerChannel.class)
+                .childHandler(new ChannelInitializer<Channel>() {
+                  @Override
+                  protected void initChannel(Channel ch) {
+                      ch.pipeline().addLast(sslServerCtx.newHandler(ch.alloc()));
+                      ch.pipeline().addLast(new ChannelInboundHandlerAdapter() {
+                          @Override
+                          public void channelActive(ChannelHandlerContext ctx) {
+                              ByteBuf buf = ctx.alloc().buffer(10);
+                              buf.writerIndex(buf.readerIndex() + 10);
```
normanmaurer: nit: you could also just call `buf.writeZero(10)`.

Scottmitch: done

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
Scottmitch: done

buffer/src/test/java/io/netty/buffer/AbstractByteBufTest.java
```diff
@@ -19,6 +19,7 @@
 import io.netty.util.CharsetUtil;
 import io.netty.util.IllegalReferenceCountException;
 import io.netty.util.internal.PlatformDependent;
+import javafx.application.Platform;
```
normanmaurer: Remove ?

buffer/src/test/java/io/netty/buffer/AbstractByteBufTest.java
```diff
@@ -115,6 +116,9 @@ public void dispose() {
 
     @Test
     public void comparableInterfaceNotViolated() {
+        ByteBuffer bb = ByteBuffer.allocateDirect(10);
+        ByteBuffer bbro = bb.asReadOnlyBuffer();
+        long f = PlatformDependent.directBufferAddress(bbro);
```
normanmaurer: Actually remove the change ?

ejona86:
@Scottmitch, I back-ported the patch to 4.1.17 and can confirm it fixes the issue.

handler/src/test/java/io/netty/handler/ssl/SslHandlerTest.java
```diff
@@ -431,6 +437,99 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exc
...
+            assertThat(evt.cause(), is(instanceOf(ClosedChannelException.class)));
+            assertTrue(events.isEmpty());
+        } finally {
+            if (serverChannel != null) {
+                serverChannel.close();
+            }
+            if (clientChannel != null) {
+                clientChannel.close();
+            }
+            group.shutdownGracefully();
```
normanmaurer: Nit: call ssc.delete()

Scottmitch: We don't consistently do this in tests. Lets do this in a followup PR if we think it is necessary.

Scottmitch:
4.1 (978a46cc0a) 4.0 (94e55692d4)

PR #7671, opened 2018-02-01
DnsNameResolverContext#followCname only uses first name server

Scottmitch:
Motivation:
When following a CNAME it is possible there are multiple name servers to query against. However DnsNameResolverContext#followCname explicitly only uses the first name server address when attempting the query. This may lead to resolution failures because we didn't try all the available name servers.

Modifications:
DnsNameResolverContext#followCname should not just try the first name server, but it should try all name servers

Result:
More complete CNAME resolution.

### DIFF
--- a/resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
+++ b/resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
@@ -49,6 +49,7 @@ import org.apache.directory.server.dns.messages.RecordClass;
 import org.apache.directory.server.dns.messages.RecordType;
 import org.apache.directory.server.dns.messages.ResourceRecord;
 import org.apache.directory.server.dns.messages.ResourceRecordModifier;
+import org.apache.directory.server.dns.messages.ResponseCode;
 import org.apache.directory.server.dns.store.DnsAttribute;
 import org.apache.directory.server.dns.store.RecordStore;
 import org.junit.AfterClass;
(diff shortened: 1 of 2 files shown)


resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
```diff
@@ -817,6 +819,108 @@ private static void testCNAMERecursiveResolve(boolean ipv4Preferred) throws IOEx
...
+        DnsNameResolver resolver = null;
+        try {
+            DnsCache nsCache = new DefaultDnsCache();
+            // What we want to test is the following:
+            // 1. Do a DNS query.
+            // 2. CNAME is returned, we want to lookup that CNAME on multiple DNS servers
+            // 3. The first DNS server should fail
+            // 4. The second DNS server should succeed
+            // This verifies that we do in fact follow multiple DNS servers in the CNAME resolution.
+            // The DnsCache is used for the name server cache, but doesn't provide a InetSocketAddress (only InetAddress
```
Scottmitch: @normanmaurer - the test is a bit convoluted ... seems like the name server cache should support `InetSocketAddress` instead of `InetAddress`.

normanmaurer: @Scottmitch I think it comes from the fact that NS records not allow to specify a port for an entry.

normanmaurer:
Cherry-picked into 4.1 (fe5c69bdd9d65ddd604a03bfc71081b5bc2ad40a).

@Scottmitch thanks!

PR #7676, opened 2018-02-01
Avoid register multiple cleaner task for same thread's FastThreadLoca…

kojilin:
…l index

Motivation:

Currently if user call set/remove/set/remove many times, it will create multiple cleaner task for same index. It may cause OOM due to long live thread will have more and more task in LIVE_SET.

Modification:

Add flag to avoid recreating tasks.

Result:
Only create 1 clean task. But use more space of indexedVariables.

### DIFF
--- a/common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
+++ b/common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
@@ -125,8 +125,11 @@ public class FastThreadLocal<V> {
 
     private final int index;
 
+    private final int cleanerFlagIndex;
+
     public FastThreadLocal() {
         index = InternalThreadLocalMap.nextVariableIndex();
+        cleanerFlagIndex = InternalThreadLocalMap.nextVariableIndex();
     }
 
     /**
@@ -147,19 +150,25 @@ public class FastThreadLocal<V> {
 
     private void registerCleaner(final InternalThreadLocalMap threadLocalMap) {
         Thread current = Thread.currentThread();
-        if (!FastThreadLocalThread.willCleanupFastThreadLocals(current)) {
-            // We will need to ensure we will trigger remove(InternalThreadLocalMap) so everything will be released
-            // and FastThreadLocal.onRemoval(...) will be called.
-            ObjectCleaner.register(current, new Runnable() {
-                @Override
-                public void run() {
-                    remove(threadLocalMap);
-
-                    // It's fine to not call InternalThreadLocalMap.remove() here as this will only be triggered once
-                    // the Thread is collected by GC. In this case the ThreadLocal will be gone away already.
-                }
-            });
+        if (FastThreadLocalThread.willCleanupFastThreadLocals(current) ||
+            threadLocalMap.indexedVariable(cleanerFlagIndex) != InternalThreadLocalMap.UNSET) {
+            return;
         }
+        // removeIndexedVariable(cleanerFlagIndex) isn't necessary because the finally cleanup is tied to the lifetime
+        // of the thread, and this Object will be discarded if the associated thread is GCed.
+        threadLocalMap.setIndexedVariable(cleanerFlagIndex, Boolean.TRUE);
+
+        // We will need to ensure we will trigger remove(InternalThreadLocalMap) so everything will be released
+        // and FastThreadLocal.onRemoval(...) will be called.
+        ObjectCleaner.register(current, new Runnable() {
+            @Override
+            public void run() {
+                remove(threadLocalMap);
+
+                // It's fine to not call InternalThreadLocalMap.remove() here as this will only be triggered once
+                // the Thread is collected by GC. In this case the ThreadLocal will be gone away already.
+            }
+        });
     }
 
     /**
--- a/common/src/test/java/io/netty/util/concurrent/FastThreadLocalTest.java
+++ b/common/src/test/java/io/netty/util/concurrent/FastThreadLocalTest.java
@@ -16,6 +16,7 @@
 
 package io.netty.util.concurrent;
 
+import io.netty.util.internal.ObjectCleaner;
 import org.junit.Before;
 import org.junit.Test;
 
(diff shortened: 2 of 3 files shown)


common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
```diff
@@ -147,19 +148,27 @@ public final V get() {
...
-                    // It's fine to not call InternalThreadLocalMap.remove() here as this will only be triggered once
-                    // the Thread is collected by GC. In this case the ThreadLocal will be gone away already.
-                }
-            });
+        if (FastThreadLocalThread.willCleanupFastThreadLocals(current)) {
+            return;
+        }
+        Object v = threadLocalMap.indexedVariable(addToCleanerFlagIndex);
+        if (v != InternalThreadLocalMap.UNSET) {
+            return;
```
johnou: @kojilin so the InternalThreadLocalMap is always the same instance?

common/src/test/java/io/netty/util/concurrent/FastThreadLocalTest.java
```diff
@@ -76,6 +80,30 @@ public void run() {
...
     }
 
+    @Test
+    public void testMultipleSetRemove() throws Exception {
+        final FastThreadLocal<String> threadLocal = new FastThreadLocal<String>();
+
+        final Field liveSetField = ObjectCleaner.class.getDeclaredField("LIVE_SET");
+        liveSetField.setAccessible(true);
+        final Set<?> liveSet = (Set<?>) liveSetField.get(ObjectCleaner.class);
+        final int sizeWhenStart = liveSet.size();
```
normanmaurer: can we make this work without reflection ? Maybe expose this as a getter on the `ObjectCleaner` ?

common/src/test/java/io/netty/util/concurrent/FastThreadLocalTest.java
```diff
@@ -76,6 +80,30 @@ public void run() {
...
+
+        Thread thread = new Thread(new Runnable() {
+            @Override
+            public void run() {
+                threadLocal.set("1");
+                threadLocal.remove();
+                threadLocal.set("2");
+                threadLocal.remove();
+            }
+        });
```
normanmaurer: Also can you please use at least two threads to ensure everything is fine even when we have more then one ?

common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
kojilin: ah...I misunderstood about this, I need to add flag for each index.

kojilin: Looks like I need to remove for each index.

common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
```diff
@@ -125,8 +125,11 @@ private static void removeFromVariablesToRemove(
 
     private final int index;
 
+    private final int cleanerFlagIndex;
+
     public FastThreadLocal() {
         index = InternalThreadLocalMap.nextVariableIndex();
+        cleanerFlagIndex = InternalThreadLocalMap.nextVariableIndex();
```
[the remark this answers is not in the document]
normanmaurer: I need to think about it.

normanmaurer:
@kojilin please fix the check style errors

normanmaurer:
@kojilin I may have a better fix... let me try something

normanmaurer:
@kojilin so the idea is to use a special Object like `UNSET` to mark that there was one value set at some point and only register the cleaner if `UNSET` was stored before. WDYT ?

normanmaurer:
@kojilin actually I think your solution is just fine... Can you just fix all the check style errors (which should show up when you run `mvn clean package` and push again ?

kojilin:
@normanmaurer Just update the style, thanks.

common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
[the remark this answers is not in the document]
trustin: Hmm, we may be able to incorporate this flag into a dedicated bitset in `InternalThreadLocalMap` but that would deserve another PR.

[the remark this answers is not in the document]
normanmaurer: yeah maybe we can improve it over time.. that said I am not super concerned about the need of the extra "slot" here... WDYT ?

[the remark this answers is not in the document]
normanmaurer: @trustin ^^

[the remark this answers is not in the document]
trustin: Hard to tell without numbers I guess. Not very concerned though.

common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
```diff
@@ -147,19 +150,26 @@ public final V get() {
...
-                }
-            });
+        if (FastThreadLocalThread.willCleanupFastThreadLocals(current)) {
+            return;
         }
+        Object v = threadLocalMap.indexedVariable(cleanerFlagIndex);
+        if (v != InternalThreadLocalMap.UNSET) {
+            return;
+        }
+        threadLocalMap.setIndexedVariable(cleanerFlagIndex, Boolean.TRUE);
```
Scottmitch: consider adding a comment that `threadLocalMap.removeIndexedVariable(cleanerFlagIndex);` isn't necessary because the finally cleanup is tied to the lifetime of the thread, and this Object will be discarded if the associated thread is GCed.

johnou:
@normanmaurer for 4.1.21 right?

trustin:
Yes, please.

normanmaurer:
Cherry-picked into 4.1 (1e70d3092d547b1d19f9bf3a4da231701d19de6a) and 4.0 (0f6284ddbf35298db20a3398d71fe4719f73e74f)

Issue #7673, opened 2018-02-01
JdkSslContext produces invalid supported ciphers


slandelle:
### Expected behavior

JdkSslContext should produce valid ciphers that are actually supported by the platform.
Typically, it should be possible to use them in a custom `CipherSuiteFilter` to configure a `SSLEngine`.

### Actual behavior

Netty 4.1.13.Final introduced a workaround for IBM certs naming scheme.
This workaround is based on IBM's statement:

> According to IBM's documentation [2] the "SSL_" prefix is "interchangeable" with the "TLS_" prefix.

This statement is wrong and transformed names cause `SSLEngine#setEnabledCipherSuites` to crash.

On my machine, the following ciphers get transformed with the `SSL` prefix being replaced with `TLS` and cause SSLEngine to crash:

* SSL_DH_anon_WITH_DES_CBC_SHA
* SSL_DHE_DSS_WITH_3DES_EDE_CBC_SHA
* SSL_DH_anon_WITH_3DES_EDE_CBC_SHA
* SSL_DHE_RSA_WITH_DES_CBC_SHA
* SSL_RSA_WITH_DES_CBC_SHA
* SSL_DHE_DSS_WITH_DES_CBC_SHA
* SSL_RSA_WITH_NULL_SHA
* SSL_RSA_WITH_3DES_EDE_CBC_SHA
* SSL_RSA_WITH_NULL_MD5
* SSL_DHE_RSA_WITH_3DES_EDE_CBC_SHA

### Steps to reproduce

See reproducer below.

### Minimal yet complete reproducer code (or URL to code)

```java
SSLContext sslcontext = SSLContext.getInstance("TLS");
sslcontext.init(null, null, null);
SSLEngine sslEngine = sslcontext.createSSLEngine();

// works
sslEngine.setEnabledCipherSuites(new String[] {"SSL_DH_anon_WITH_DES_CBC_SHA"});

// crashes with IllegalArgumentException: Unsupported ciphersuite TLS_DH_anon_WITH_DES_CBC_SHA
sslEngine.setEnabledCipherSuites(new String[] {"TLS_DH_anon_WITH_DES_CBC_SHA"});
```

### Netty version

Tested on 4.1.20.Final

### JVM version (e.g. `java -version`)

Java HotSpot(TM) 64-Bit Server VM (build 25.152-b16, mixed mode)

### OS version (e.g. `uname -a`)

Not relevant

normanmaurer:
[...]

PR #7677, opened 2018-02-01
JdkSslContext supported cipher suites incorrect

Scottmitch:
Motivation:
JdkSslContext builds the list of supported cipher suites, but assumes that ciphers prefixed with SSL_ and TLS_ will be interchangeable. However this is not the case and only applies to a small subset of ciphers. This results in the JdkSslContext attempting to use unsupported ciphers.

Modifications:
- When building the list of ciphers in JdkSslContext we should first check if the engine supports the TLS_ prefix cipher.

Result:
Fixes https://github.com/netty/netty/issues/7673

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/JdkSslContext.java
+++ b/handler/src/main/java/io/netty/handler/ssl/JdkSslContext.java
@@ -105,7 +105,13 @@ public class JdkSslContext extends SslContext {
             //[3] https://www.ibm.com/developerworks/community/forums/html/topic?id=9b5a56a9-fa46-4031-b33b-df91e28d77c2
             //[4] https://www.ibm.com/developerworks/rfe/execute?use_case=viewRfe&CR_ID=71770
             if (supportedCipher.startsWith("SSL_")) {
-                SUPPORTED_CIPHERS.add("TLS_" + supportedCipher.substring("SSL_".length()));
+                final String tlsPrefixedCipherName = "TLS_" + supportedCipher.substring("SSL_".length());
+                try {
+                    engine.setEnabledCipherSuites(new String[]{tlsPrefixedCipherName});
+                    SUPPORTED_CIPHERS.add(tlsPrefixedCipherName);
+                } catch (IllegalArgumentException ignored) {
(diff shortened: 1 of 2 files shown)


handler/src/main/java/io/netty/handler/ssl/JdkSslContext.java
```diff
@@ -105,7 +105,13 @@
             //[3] https://www.ibm.com/developerworks/community/forums/html/topic?id=9b5a56a9-fa46-4031-b33b-df91e28d77c2
             //[4] https://www.ibm.com/developerworks/rfe/execute?use_case=viewRfe&CR_ID=71770
             if (supportedCipher.startsWith("SSL_")) {
-                SUPPORTED_CIPHERS.add("TLS_" + supportedCipher.substring("SSL_".length()));
+                final String tlsPrefixedCipherName = "TLS_" + supportedCipher.substring("SSL_".length());
+                try {
+                    engine.setEnabledCipherSuites(new String[]{tlsPrefixedCipherName});
+                    SUPPORTED_CIPHERS.add(tlsPrefixedCipherName);
+                } catch (IllegalArgumentException ignored) {
+                    // The cipher is not supported ... move on to the next cipher.
```
slandelle: Shouldn't you fall back to adding the original name here?

Scottmitch: the original name has already been added above on line 97, and is known to be supported because its returned by `engine.getSupportedCipherSuites()`.

slandelle: Oh right

Scottmitch:
4.0 (8eb80a35e0)

normanmaurer:
@Scottmitch can you also note the sha for 4.1 ?

Scottmitch:
@normanmaurer - GitHub notes the sha above as part of the merge ... am I the only one that can see this bcz I merged it?

screen shot 2018-02-02 at 8 15 15 am

normanmaurer:
@Scottmitch ah yeah I did not see this 🤣

PR #7679, opened 2018-02-02
To detect Android, check the VM property rather than the classpath

carl-mastrangelo:
Motivation:
Some java binaries include android classes on their classpath, even
if they aren't actually android.  When this is true, `Unsafe` no
longer works, disabling the Epoll functionality.  A sample case is
for binaries that use the j2objc library.

Modifications:
Check the `java.vm.name` instead of the classpath.   Numerous
Google-internal Android libraries / binaries check this property
rather than the class path.

It is believed this is safe and works with bother ART and Dalvik
VMs, safe for Robolectric, and j2objc.

Results:
Unusually built java server binaries can still use Netty Epoll.

cc: 
@ericgribkoff

### DIFF
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent0.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent0.java
@@ -786,19 +786,19 @@ final class PlatformDependent0 {
     }
 
     private static boolean isAndroid0() {
-        boolean android;
-        try {
-            Class.forName("android.app.Application", false, getSystemClassLoader());
-            android = true;
-        } catch (Throwable ignored) {
-            // Failed to load the class uniquely available in Android.
-            android = false;
-        }
+        // Idea: Sometimes java binaries include Android classes on the classpath, even if it isn't actually Android.
+        // Rather than check if certain classes are present, just check the VM, which is tied to the JDK.
 
-        if (android) {
+        // Optional improvement: check if `android.os.Build.VERSION` is >= 24. On later versions of Android, the
+        // OpenJDK is used, which means `Unsafe` will actually work as expected.
+
+        // Android sets this property to Dalvik, regardless of whether it actually is.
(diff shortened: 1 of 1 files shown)


common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -786,19 +786,19 @@ static boolean isAndroid() {
...
+        // Idea: Sometimes java binaries include Android classes on the classpath, even if it isn't actually Android.
+        // Rather than check if certain classes are present, just check the VM, which is tied to the JDK.
+
+        // Optional improvement: check if `android.os.Build.VERSION` is >= 24. On later versions of Android, the
+        // OpenJDK is used, which means `Unsafe` will actually work as expected.
 
-        if (android) {
+        // Android sets this property to Dalvik, regardless of whether it actually is.
+        String vmName = SystemPropertyUtil.get("java.vm.name");
+        boolean isAndroid = vmName.equals("Dalvik");
```
fenik17: This is correct for `ART` runtime?

normanmaurer:
Cherry-picked into 4.1 (4ed961f4fee1884f58118896c542b4a33803d47e). 

@carl-mastrangelo thanks!

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
johnou: @normanmaurer depending on security manager or jvm could SystemPropertyUtil.get("java.vm.name") return null? might be safer to use yoda style equals here.

johnou: https://github.com/netty/netty/pull/7700

normanmaurer: hmm.. maybe not a bad idea. @johnou want to do a pr ?

PR #7683, opened 2018-02-02
Introduce a CipherSuiteFilter that defaults to supportedCiphers, close #7655

### LINKED ISSUE #7655: Provide a way to disable cipher suites filtering
### Expected behavior

There should be a way to have client side `SslContext` work against any website, regardless of security.

### Actual behavior

Supported ciphers are filtered against a hard coded whitelist (`SslUtils#DEFAULT_CIPHER_SUITES`).

I wonder if the whitelist way is a good thing. Wouldn't it filter out modern ciphers that would be introduced in JDK updates? Properly maintaining this blacklist would require Netty core devs to frequently update it. Wouldn't a blacklist work better: ban ciphers that are known as unsecure?

Then, my core issue is that there's no way to disable this security. On the client side, sometimes security is not the top most priority and users just want things to work, all the more as they do work with all the other user agents around (browsers, curl, Apache HttpComponents). Typical examples are load tests and web crawlers.

### Steps to reproduce

For example, it's not possible to connect to "https://www.gretalr.com" with a `JdkSslHandler` while it's possible with any other user agent I know of (reported against AHC as https://github.com/AsyncHttpClient/async-http-client/issues/1510).

```java
EventLoopGroup eventLoopGroup = new NioEventLoopGroup();
final SslContext sslContext = SslContextBuilder.forClient().build();
final CountDownLatch latch = new CountDownLatch(1);

Bootstrap b = new Bootstrap()
        .group(eventLoopGroup)
        .channel(NioSocketChannel.class)
        .handler(new ChannelInitializer<NioSocketChannel>() {
          @Override
          protected void initChannel(NioSocketChannel nioSocketChannel) {
            SSLEngine sslEngine = sslContext.newEngine(nioSocketChannel.alloc(), "www.gretalr.com", 443);
            SslHandler sslHandler = new SslHandler(sslEngine);
            sslHandler.handshakeFuture().addListener(whenHandshaked -> {
              if (whenHandshaked.isSuccess()) {
                System.out.println("All good");
              } else {
                whenHandshaked.cause().printStackTrace();
              }
              latch.countDown();
            });
            nioSocketChannel.pipeline().addLast(sslHandler);
          }
        });
Channel channel = b.connect("www.gretalr.com", 443).sync().channel();
latch.await();
channel.close();
eventLoopGroup.shutdownGracefully();
```

Note: I can connect fine if I create the SSLEngine from  a `javax.net.ssl.SSLContext` instead.

### Netty version

4.1.20.Final

1.8.0_152

not relevant

slandelle:
Motivation:

Sometimes, it would be convenient to be able to easily enable all
supported cipher suites, regardless of security.

Currently, the only way it to retrieve all supported ciphers and pass
them explicitly.

Modification:

Introduce a new CipherSuiteFilter that behaves like Identity
CipherSuiteFilter except that it defaults to supportedCiphers instead
of defaultCiphers when ciphers are null.

Result:

Convenient way to enabled all supported cipher suites.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/IdentityWithSupportedFallbackCipherSuiteFilter.java
+++ b/handler/src/main/java/io/netty/handler/ssl/IdentityWithSupportedFallbackCipherSuiteFilter.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.ssl;
+
+import java.util.List;
+import java.util.Set;
+
+/**
+ * This class behaves like {@link IdentityCipherSuiteFilter} except that it falls back to supportedCiphers
+ * instead of defaultCiphers when ciphers are null.
+ *
+ * Typical usage is to enable weak and insecure ciphers that are supported but filtered out in defaultCiphers.
+ */
+public final class IdentityWithSupportedFallbackCipherSuiteFilter implements CipherSuiteFilter {
+    public static final IdentityWithSupportedFallbackCipherSuiteFilter INSTANCE =
+            new IdentityWithSupportedFallbackCipherSuiteFilter();
+
+    private IdentityWithSupportedFallbackCipherSuiteFilter() { }
+
+    @Override
+    public String[] filterCipherSuites(Iterable<String> ciphers,
+                                       List<String> defaultCiphers,
+                                       Set<String> supportedCiphers) {
+        return ciphers == null ?
+                supportedCiphers.toArray(new String[supportedCiphers.size()]) :
+                IdentityCipherSuiteFilter.INSTANCE.filterCipherSuites(ciphers, defaultCiphers, supportedCiphers);
+    }
+}


normanmaurer:
@slandelle hmm... didnt you want to add a new filter and not modify an existing ? IT seems like this may produce unexpected behaviour for users that expect the old behaviour

slandelle:
o_0 that wasn’t my intent. Weird git stuff. Will have a look.

slandelle:
I do have both files in my fork, it wasn’t a renaming of existing IdentityCipherSuitesFilter. See https://github.com/slandelle/netty/tree/master/handler/src/main/java/io/netty/handler/ssl.
What happens if you cherry pick in your side?

normanmaurer:
@slandelle I didnt say it was a renaming just that you changed the behaviour of an existing class:

https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/SupportedCipherSuiteFilter.java

slandelle:
Ahhh, I wasn’t aware of this class and happened to pick the same name!!! AFK, will fix later

slandelle:
@normanmaurer PTAL again. I'm having a hard time coming up with a proper name...

normanmaurer:
@slandelle can you please fix check style errors ?

slandelle:
@normanmaurer Should be fine now

handler/src/main/java/io/netty/handler/ssl/IdentityWithSupportedFallbackCipherSuiteFilter.java
```diff
@@ -0,0 +1,41 @@
...
+ * under the License.
+ */
+package io.netty.handler.ssl;
+
+import java.util.List;
+import java.util.Set;
+
+/**
+ * This class behaves like {@link IdentityCipherSuiteFilter} except that it falls back to supportedCiphers
+ * instead of defaultCiphers when ciphers are null.
```
johnou: should be easy to unit test?

slandelle:
Actually, I think it would be better to modify existing IdentityCipherSuiteFilter instead of creating a new class.
Let me close this PR in favor of a new one.

PR #7684, opened 2018-02-02
Propagate full Unsafe unavailability reason in PlatformDependent

carl-mastrangelo:
Motivation:
It is not clear why Unsafe is unavailable when it is explicitly
disabled, or when Netty thinks it is running on Android.

Modification:
Change the "has" fields and methods to be causes.  A null cause
means Unsafe is present.  This catches all possible reason why
Unsafe might not be available.

Result:
Easier to debug Netty start up when logging cannot be turned on.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent.java
@@ -78,9 +78,9 @@ public final class PlatformDependent {
 
     private static final boolean CAN_ENABLE_TCP_NODELAY_BY_DEFAULT = !isAndroid();
 
-    private static final boolean HAS_UNSAFE = hasUnsafe0();
+    private static final Throwable UNSAFE_UNAVAILABILITY_CAUSE = unsafeUnavailabilityCause0();
     private static final boolean DIRECT_BUFFER_PREFERRED =
-            HAS_UNSAFE && !SystemPropertyUtil.getBoolean("io.netty.noPreferDirect", false);
+            UNSAFE_UNAVAILABILITY_CAUSE == null && !SystemPropertyUtil.getBoolean("io.netty.noPreferDirect", false);
     private static final long MAX_DIRECT_MEMORY = maxDirectMemory0();
 
     private static final int MPSC_CHUNK_SIZE =  1024;
@@ -249,14 +249,14 @@ public final class PlatformDependent {
      * direct memory access.
      */
     public static boolean hasUnsafe() {
-        return HAS_UNSAFE;
+        return UNSAFE_UNAVAILABILITY_CAUSE == null;
     }
 
     /**
      * Return the reason (if any) why {@code sun.misc.Unsafe} was not available.
      */
     public static Throwable getUnsafeUnavailabilityCause() {
-        return PlatformDependent0.getUnsafeUnavailabilityCause();
+        return UNSAFE_UNAVAILABILITY_CAUSE;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


carl-mastrangelo:
This is a sibling PR to https://github.com/netty/netty/pull/7679 .   In our environment, turning on logging is sometimes difficult.  Having programmatic access to the exceptions makes it much easier to debug, since we can dump the whole causal chain.


As a side note: I would be interested in making PlatformDependent more testable.  Right now changes to it are based on best guesses of a lot of people.   I think having this well tested would make it easier to refactor when java9 is more prevalent, and breaks lots of things.

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -369,32 +368,33 @@ public Object run() {
...
         } else {
-            tryUnsafe = SystemPropertyUtil.getBoolean("org.jboss.netty.tryUnsafe", true);
+            unsafePropName = "org.jboss.netty.tryUnsafe";
         }
 
-        if (!tryUnsafe) {
-            logger.debug("sun.misc.Unsafe: unavailable (io.netty.tryUnsafe/org.jboss.netty.tryUnsafe)");
-            return true;
+        if (!SystemPropertyUtil.getBoolean(unsafePropName, true)) {
+            String msg = String.format("sun.misc.Unsafe: unavailable (%s)", unsafePropName);
```
normanmaurer: Just use string concatation ? I think there is no much win in use String.format at all.

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -965,24 +965,24 @@ private static boolean maybeSuperUser0() {
...
+    private static Throwable unsafeUnavailabilityCause0() {
         if (isAndroid()) {
             logger.debug("sun.misc.Unsafe: unavailable (Android)");
-            return false;
+            return new UnsupportedOperationException("sun.misc.Unsafe: unavailable (Android)");
         }
 
         if (PlatformDependent0.isExplicitNoUnsafe()) {
-            return false;
+            PlatformDependent0.getUnsafeUnavailabilityCause();
```
trustin: Missing `return`, or `return new UnsupportedOperationException("disabled as requested via system property")`?

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
carl-mastrangelo: Done.

common/src/main/java/io/netty/util/internal/PlatformDependent.java
carl-mastrangelo: Done.

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -965,24 +965,24 @@ private static boolean maybeSuperUser0() {
...
 
-    private static boolean hasUnsafe0() {
+    private static Throwable unsafeUnavailabilityCause0() {
         if (isAndroid()) {
             logger.debug("sun.misc.Unsafe: unavailable (Android)");
-            return false;
+            return new UnsupportedOperationException("sun.misc.Unsafe: unavailable (Android)");
         }
 
         if (PlatformDependent0.isExplicitNoUnsafe()) {
```
Scottmitch: this conditional is already taken into account in `PlatformDependent0` when setting `PlatformDependent0.getUnsafeUnavailabilityCause()` ... can we just let `PlatformDependent0` be responsible for enforcing this conditional? For example:

```java
Throwable cause = PlatformDependent0.getUnsafeUnavailabilityCause();
if (cause != null) {
  return cause;
}
...
```

normanmaurer: @carl-mastrangelo ^^

carl-mastrangelo: Sorry for the delay, done!

normanmaurer:
@trustin @Scottmitch can you take another look ?

normanmaurer:
@Scottmitch can we merge this ?

Scottmitch:
@normanmaurer - yip ... lgtm sorry for delay

normanmaurer:
@Scottmitch no worries buddy... thanks

PR #7687, opened 2018-02-04
 Support Redis inline commands

### LINKED ISSUE #7686: Redis codec doesn't support inline commands
### Expected behavior
RedisDecoder and RedisEncoder are able to handle [RESP inline commands](https://redis.io/topics/protocol#inline-commands).

### Actual behavior
Inline command support is not implemented. Trying to decode an inline command leads to `RedisCodecException: Unknown RedisMessageType: <decimal code of first character>`.

### Netty version
4.1.20

mseitner:
Motivation:
The RESP protocol implementation lacked inline command
support.

Modifications:
Added logic to decode and encode inline commands.

Result:
Inline commands are supported. Fixes #7686.

### DIFF
--- a/codec-redis/src/main/java/io/netty/handler/codec/redis/AbstractStringRedisMessage.java
+++ b/codec-redis/src/main/java/io/netty/handler/codec/redis/AbstractStringRedisMessage.java
@@ -16,6 +16,7 @@
 package io.netty.handler.codec.redis;
 
 import io.netty.util.internal.ObjectUtil;
+import io.netty.util.internal.StringUtil;
 import io.netty.util.internal.UnstableApi;
 
 /**
@@ -38,4 +39,14 @@ public abstract class AbstractStringRedisMessage implements RedisMessage {
     public final String content() {
         return content;
     }
+
+    @Override
+    public String toString() {
+        return new StringBuilder(StringUtil.simpleClassName(this))
+                .append('[')
+                .append("content=")
+                .append(content)
+                .append(']').toString();
+    }
+
 }
--- a/codec-redis/src/main/java/io/netty/handler/codec/redis/RedisDecoder.java
+++ b/codec-redis/src/main/java/io/netty/handler/codec/redis/RedisDecoder.java
@@ -36,6 +36,7 @@ public final class RedisDecoder extends ByteToMessageDecoder {
 
     private final ToPositiveLongProcessor toPositiveLongProcessor = new ToPositiveLongProcessor();
 
+    private final boolean decodeInlineCommands;
     private final int maxInlineMessageLength;
     private final RedisMessagePool messagePool;
 
@@ -53,25 +54,44 @@ public final class RedisDecoder extends ByteToMessageDecoder {
     }
 
     /**
-     * Creates a new instance with default {@code maxInlineMessageLength} and {@code messagePool}.
+     * Creates a new instance with default {@code maxInlineMessageLength} and {@code messagePool}
+     * and inline command decoding disabled.
      */
     public RedisDecoder() {
-        // 1024 * 64 is max inline length of current Redis server implementation.
-        this(1024 * 64, FixedRedisMessagePool.INSTANCE);
+        this(false);
+    }
+
+    /**
+     * Creates a new instance with default {@code maxInlineMessageLength} and {@code messagePool}.
+     * @param decodeInlineCommands if {@code true}, inline commands will be decoded.
+     */
+    public RedisDecoder(boolean decodeInlineCommands) {
+        this(RedisConstants.REDIS_INLINE_MESSAGE_MAX_LENGTH, FixedRedisMessagePool.INSTANCE, decodeInlineCommands);
+    }
+
+    /**
+     * Creates a new instance with inline command decoding disabled.
+     * @param maxInlineMessageLength the maximum length of inline message.
+     * @param messagePool the predefined message pool.
+     */
+    public RedisDecoder(int maxInlineMessageLength, RedisMessagePool messagePool) {
+        this(maxInlineMessageLength, messagePool, false);
     }
 
     /**
      * Creates a new instance.
      * @param maxInlineMessageLength the maximum length of inline message.
      * @param messagePool the predefined message pool.
+     * @param decodeInlineCommands if {@code true}, inline commands will be decoded.
      */
-    public RedisDecoder(int maxInlineMessageLength, RedisMessagePool messagePool) {
+    public RedisDecoder(int maxInlineMessageLength, RedisMessagePool messagePool, boolean decodeInlineCommands) {
         if (maxInlineMessageLength <= 0 || maxInlineMessageLength > RedisConstants.REDIS_MESSAGE_MAX_LENGTH) {
             throw new RedisCodecException("maxInlineMessageLength: " + maxInlineMessageLength +
                                           " (expected: <= " + RedisConstants.REDIS_MESSAGE_MAX_LENGTH + ")");
         }
         this.maxInlineMessageLength = maxInlineMessageLength;
         this.messagePool = messagePool;
+        this.decodeInlineCommands = decodeInlineCommands;
     }
 
     @Override
@@ -127,13 +147,7 @@ public final class RedisDecoder extends ByteToMessageDecoder {
             return false;
         }
 
-        final int initialIndex = in.readerIndex();
-        type = RedisMessageType.valueOf(in.readByte());
-
-        if (type.length() == 0) {
-            // byte did not contain type info, make it readable again
-            in.readerIndex(initialIndex);
-        }
+        type = RedisMessageType.readFrom(in, decodeInlineCommands);
         state = type.isInline() ? State.DECODE_INLINE : State.DECODE_LENGTH;
         return true;
     }
--- a/codec-redis/src/main/java/io/netty/handler/codec/redis/RedisDecoder.java
+++ b/codec-redis/src/main/java/io/netty/handler/codec/redis/RedisDecoder.java
@@ -126,7 +126,14 @@ public final class RedisDecoder extends ByteToMessageDecoder {
         if (!in.isReadable()) {
             return false;
         }
+
+        final int initialIndex = in.readerIndex();
         type = RedisMessageType.valueOf(in.readByte());
+
+        if (type.length() == 0) {
+            // byte did not contain type info, make it readable again
+            in.readerIndex(initialIndex);
+        }
         state = type.isInline() ? State.DECODE_INLINE : State.DECODE_LENGTH;
         return true;
     }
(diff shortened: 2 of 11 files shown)


normanmaurer:
@ddossot PTAL

codec-redis/src/main/java/io/netty/handler/codec/redis/RedisMessageType.java
```diff
@@ -23,28 +23,35 @@
...
-    public byte value() {
+    public Byte value() {
         return value;
     }
 
+    /**
+     * Returns length of this type.
+     */
+    public int length() {
+        return (value != null) ? RedisConstants.TYPE_LENGTH : 0;
```
ddossot: nit: remove parenthesis around condition for consistency with other ternaries, for ex. in `RedisDecoder`

codec-redis/src/main/java/io/netty/handler/codec/redis/RedisMessageType.java
```diff
@@ -69,7 +76,7 @@ public static RedisMessageType valueOf(byte value) {
         case '*':
             return ARRAY_HEADER;
         default:
-            throw new RedisCodecException("Unknown RedisMessageType: " + value);
+            return INLINE_COMMAND;
```
ddossot: We're changing a catch-all exception to an accept-all use case. I'm wondering if there's something more accurate we could do, for ex. only recognize `a-zA-Z` as the marker of an inline command and restore the `RedisCodecException` for all other cases?

mseitner: Redis commands can be [renamed](https://redis.io/topics/security#disabling-of-specific-commands) for security reasons. Substitute commands are not limited to `a-zA-Z` though:

redis.conf
```
rename-command PING もしもし # hello in Japanese
```

Invoke renamed command (works in `redis-cli`, too)
```
$ echo もしもし | nc localhost 6379
+PONG
```

I don't see any good reason to be more restrictive than Redis itself is.

codec-redis/src/main/java/io/netty/handler/codec/redis/RedisMessageType.java
mseitner: Fixed!

codec-redis/src/main/java/io/netty/handler/codec/redis/RedisMessageType.java
ddossot: The intent was not to be more restrictive than Redis is. Since `rename-command` allows using any character in commands, there's unfortunately not much check we can do.

Scottmitch:
I wondering if we even need to support these inline-commands. It seems like an alternative variant on the original protocol for text-based clients which may require additional work to represent the original protocol [1]. It also introduces additional complexity for encoding/decoding (as indicated above in https://github.com/netty/netty/pull/7687#discussion_r166031138) and may allow for malformed input to be parsed.

[1] https://redis.io/topics/protocol#inline-commands
> Sometimes you have only telnet in your hands and you need to send a command to the Redis server

ddossot:
@Scottmitch The inline protocol being human centric, I reckon the main use case would be people building new cli tools for Redis, which is a narrow use case.

@mseitner Did you have other use cases in mind?

mseitner:
I'm working on a Redis proxy, so I have to be transparent towards the client. I got everything up and running, but of all things it was `redis-benchmark` that refused to work. Why? The first thing it does is calling `ping` as inline command, then as regular (bulk) command:

```
$ redis-benchmark -c 1 -n 1
====== PING_INLINE ======
  1 requests completed in 0.00 seconds
  1 parallel clients
  3 bytes payload
  keep alive: 1

100.00% <= 0 milliseconds
1000.00 requests per second

====== PING_BULK ======
  1 requests completed in 0.00 seconds
  1 parallel clients
  3 bytes payload
  keep alive: 1

[...]
```

However, my concrete use case should have nothing to do with the level of protocol support. Do we agree that inline commands are part of the RESP protocol? If so, they should be supported.

Regarding backward compatibility/consistency I suggest to make inline command support configurable via constructor argument, with the default being `false`. That should make both sides happy :-)

mseitner:
@ddossot @Scottmitch How should I continue?

normanmaurer:
@mseitner I would say as long as it is "optional" and the old behaviour is the default it should be "ok"

ddossot:
@mseitner +1 for optional support for inline-cmd, off by default. This would address my main concern, which I detailed above about becoming too lax by default.

codec-redis/src/main/java/io/netty/handler/codec/redis/RedisDecoder.java
```diff
@@ -126,7 +146,18 @@ private boolean decodeType(ByteBuf in) throws Exception {
...
         }
+
+        final int initialIndex = in.readerIndex();
         type = RedisMessageType.valueOf(in.readByte());
+
+        if (type == RedisMessageType.INLINE_COMMAND && !decodeInlineCommands) {
+            throw new RedisCodecException("Decoding of inline commands is disabled");
+        }
+
+        if (type.length() == 0) {
```
Scottmitch: the `length()` method isn't really used in a generic fashion here and is a bit misleading. For example we always read 1 `byte` above without considering `length()`. Can we instead handle this `INLINE_COMMAND` logic as a continuation of the special case above?

codec-redis/src/main/java/io/netty/handler/codec/redis/RedisEncoder.java
```diff
@@ -178,20 +186,27 @@ private void writeArrayHeader(ByteBufAllocator allocator, boolean isNull, long l
...
-            buf.writeByte(RedisMessageType.ARRAY_HEADER.value());
+            writeMessageType(buf, RedisMessageType.ARRAY_HEADER);
             buf.writeBytes(numberToBytes(length));
             buf.writeShort(RedisConstants.EOL_SHORT);
             out.add(buf);
         }
     }
 
+    private static void writeMessageType(ByteBuf buf, RedisMessageType type) {
+        if (type.length() == 0) {
```
Scottmitch: instead of `length` here, we could just let `RedisMessageType` write itself to the `buf`:

```java
enum RedisMessageType {
  public void serialize(ByteBuf but) {
    if (value != null) { buf.writeByte(value); }
  }
}
```

codec-redis/src/main/java/io/netty/handler/codec/redis/RedisMessageType.java
```diff
@@ -23,28 +23,35 @@
...
         this.value = value;
         this.inline = inline;
     }
 
     /**
-     * Returns prefix {@code byte} for this type.
+     * Returns prefix {@code byte} for this type or {@code null} if its length is zero.
      */
-    public byte value() {
+    public Byte value() {
```
Scottmitch: if we introduce the `serialize` method above, perhaps we don't have to change to `Byte`. We could use `\0` or some "token" to represent null.

codec-redis/src/main/java/io/netty/handler/codec/redis/RedisDecoder.java
mseitner: I'll move the entire type detection to `RedisMessageType`. This will also improve functional consistency with the suggested serialization change below.

codec-redis/src/main/java/io/netty/handler/codec/redis/RedisMessageType.java
mseitner: I don't see the benefits of a magic number to represent the absence(!) of a Redis type indicator.

mseitner:
Changed as promised, the entire type detection is now implemented in `RedisMessageType`.

codec-redis/src/main/java/io/netty/handler/codec/redis/AbstractStringRedisMessage.java
```diff
@@ -38,4 +39,14 @@
     public final String content() {
         return content;
     }
+
+    @Override
+    public String toString() {
+        return new StringBuilder(StringUtil.simpleClassName(this))
```
normanmaurer: Just use normal string concats or at least size the StringBuilder correctly ?

codec-redis/src/main/java/io/netty/handler/codec/redis/RedisMessageType.java
```diff
@@ -54,9 +55,32 @@ public boolean isInline() {
     }
 
     /**
-     * Return {@link RedisMessageType} for this type prefix {@code byte}.
+     * Determine {@link RedisMessageType} based on the type prefix {@code byte} read from given the buffer.
      */
-    public static RedisMessageType valueOf(byte value) {
+    public static RedisMessageType readFrom(ByteBuf in, boolean decodeInlineCommands) {
+        final int initialIndex = in.readerIndex();
+        final RedisMessageType type = valueOf(in.readByte());
```
normanmaurer: just use `in.getByte(in.readerIndex())` this way you will not need to reset the index

mseitner: How does that help? If I use the suggested method I instead have to manually advance the reader index in all other cases to "consume" the type indicator.

codec-redis/src/main/java/io/netty/handler/codec/redis/AbstractStringRedisMessage.java
mseitner: As you can [see](https://github.com/netty/netty/pull/7687/commits/a9f2e7686aa5c9d64661554055830fbad17ae935#diff-c1537befe2d2bd6dd21162a189fee058L37) I just pulled up the `toString()` method from existing sub-classes. The implementation is similar to that of [other (core) classes](https://github.com/netty/netty/blob/73f23c5faa967748bdda826278ac5e63f715ae09/buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java#L1332). Is a more complex implementation that calculates the needed size really worth it? I suggest to never call the `toString()` method in a productive environment anyway.

codec-redis/src/main/java/io/netty/handler/codec/redis/RedisMessageType.java
normanmaurer: never mind.. brain fart.

codec-redis/src/main/java/io/netty/handler/codec/redis/AbstractStringRedisMessage.java
normanmaurer: I think it worth it... that said I think we can improve it in another PR.

mseitner: Great! I'd be happy to help with that :-) Anything else that should be done?

normanmaurer:
Squashed and cherry-picked into 4.1 (c75bc1f25b623d0fb2fa4882bcc19dc30ff1649b).

Sorry for the slow turn-around

PR #7688, opened 2018-02-04
Change the encoded buffer of the published message to CompositeByteBuf

r16turbo:
Motivation:

Reduce the memory copy of the payload.

Modification:

Using the CompositeByteBuf.

Result:

Memory usage and performance may improve slightly.


r16turbo:
In the case of EpollSocketChannel, performance is degraded because CompositeByteBuf has no memory address.

Issue #7695, opened 2018-02-07
IllegalReferenceCountException


IgorAlelekov:
Running the io.netty.example.http.upload.HttpUploadServer example the io.netty.util.IllegalReferenceCountException is thrown, please see the trace below

### Netty version
4.1.21

### JVM version (e.g. `java -version`)
1.8.131

### OS version (e.g. `uname -a`)
win10

2018 2:48:00 PM io.netty.example.http.upload.HttpUploadServerHandler exceptionCaught
WARNING: 
io.netty.util.IllegalReferenceCountException: refCnt: 0, increment: 1
	at io.netty.util.AbstractReferenceCounted.release0(AbstractReferenceCounted.java:87)
	at io.netty.util.AbstractReferenceCounted.release(AbstractReferenceCounted.java:71)
	at io.netty.handler.codec.http.multipart.MixedAttribute.release(MixedAttribute.java:320)
	at io.netty.handler.codec.http.multipart.DefaultHttpDataFactory.cleanRequestHttpData(DefaultHttpDataFactory.java:284)
	at io.netty.handler.codec.http.multipart.HttpPostMultipartRequestDecoder.cleanFiles(HttpPostMultipartRequestDecoder.java:958)
	at io.netty.handler.codec.http.multipart.HttpPostMultipartRequestDecoder.destroy(HttpPostMultipartRequestDecoder.java:937)
	at io.netty.handler.codec.http.multipart.HttpPostRequestDecoder.destroy(HttpPostRequestDecoder.java:247)
	at io.netty.example.http.upload.HttpUploadServerHandler.reset(HttpUploadServerHandler.java:211)
	at io.netty.example.http.upload.HttpUploadServerHandler.channelRead0(HttpUploadServerHandler.java:199)
	at io.netty.example.http.upload.HttpUploadServerHandler.channelRead0(HttpUploadServerHandler.java:1)
	at io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:105)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
[...]

Issue #7689, opened 2018-02-05
4.1.20 HttpPostMultipartRequestDecoder destroy duplicate


luastar:
### Netty version
4.1.20

### 
when use HttpPostMultipartRequestDecoder.destroy=>cleanFiles，it will call factory(DefaultHttpDataFactory).cleanRequestHttpData，and call data.release()【in netty 4.1.19 is data.delete()】 but at the end of destroy call data.release() again。
then get a Exception 
java.lang.IllegalStateException: HttpPostMultipartRequestDecoder was destroyed already

doom369:
@luastar did you use Netty upload example? Could be same issue as I had - https://github.com/netty/netty/issues/7611

johnou:
@doom369 does the upload example need to be fixed?

doom369:
@johnou probably yes. `data.release();` from `readHttpDataChunkByChunk()` of `HttpUploadServerHandler` should be removed. Need to check it. I haven't chance to check.

PR #7696, opened 2018-02-07
#7695 no need to manually release chunk during upload in HttpUploadServer

doom369:
Motivation:

After https://github.com/netty/netty/pull/7527 fix there is no need to manually release chunks (`HttpData`) during file upload as they will be released on `HttpPostRequestDecoder.destroy()`.

Modification:

`HttpUploadServer` example doesn't release chunks manually (doesn't call `data.release()`).

Result:
Fixes https://github.com/netty/netty/issues/7695 and https://github.com/netty/netty/issues/7689

### DIFF
--- a/example/src/main/java/io/netty/example/http/upload/HttpUploadServerHandler.java
+++ b/example/src/main/java/io/netty/example/http/upload/HttpUploadServerHandler.java
@@ -225,12 +225,8 @@ public class HttpUploadServerHandler extends SimpleChannelInboundHandler<HttpObj
                         logger.info(" 100% (FinalSize: " + partialContent.length() + ")");
                         partialContent = null;
                     }
-                    try {
-                        // new value
-                        writeHttpData(data);
-                    } finally {
-                        data.release();
-                    }
+                    // new value
+                    writeHttpData(data);
                 }
             }
             // Check partial decoding for a FileUpload


example/src/main/java/io/netty/example/http/upload/HttpUploadServerHandler.java
```diff
@@ -225,12 +225,8 @@ private void readHttpDataChunkByChunk() {
...
                         partialContent = null;
                     }
-                    try {
-                        // new value
-                        writeHttpData(data);
-                    } finally {
-                        data.release();
-                    }
+                    // new value
+                    writeHttpData(data);
```
johnou: writeHttpData always guarantees to release data even if it throws an exception for some reason?

doom369: writeHttpData just appends the content of chunks to StringBuilder it doesn't and shouldn't release data.

Actual release happens in reset() method.

johnou: inside finally block?

doom369: Why should it?

johnou: Just asking if it invokes HttpPostRequestDecoder.destroy() in all cases to avoid memory leaks.

doom369: HttpPostRequestDecoder.destroy() invoked in exceptionCaught and in channelInactive. Not the best places, however, this should work fine in case of error.

johnou: aha ok!

pnowojski:
Could anyone confirm (@johnou @doom369 @fishjam) which of the fixes: this one, or https://github.com/netty/netty/issues/7814 is the valid one for this bug? After upgrading netty version we experienced the same problem (our file uploading looks like was based on this example) and I'm not sure whether dropping this `data.release()` call is save/official/approved by netty fix.

doom369:
@pnowojski my approach works for me well for a long period on the production system.

pnowojski:
Thanks for the information. We will probably go the same way.

PR #7698, opened 2018-02-08
Utility method for reading a string from a ByteBuf.

johnou:
Motivation:

Easier way to read a string efficiently from a ByteBuf.

Modifications:

Add utility method and unit test for reading a String from a ByteBuf.

Result:

Less boilerplate code in order to read a string from a ByteBuf.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
+++ b/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
@@ -439,6 +439,30 @@ public final class ByteBufUtil {
         }
     }
 
+    /**
+     * Reads a string from the given {@link ByteBuf} in an efficient manner.
+     * @param buffer The buffer to read from.
+     * @param length The amount of bytes to be read from the buffer.
+     * @param charset The charset of the string to be returned.
+     * @return {@link String} from the given {@link ByteBuf} with the provided length and charset.
+     */
+    public static String readString(ByteBuf buffer, int length, Charset charset) {
+        if (buffer.isDirect()) {
+            final String result = buffer.toString(buffer.readerIndex(), length, charset);
+            buffer.skipBytes(length);
+            return result;
+        } else if (buffer.hasArray()) {
+            final String result = new String(buffer.array(), buffer.arrayOffset() + buffer.readerIndex(), length,
+                    charset);
+            buffer.skipBytes(length);
+            return result;
+        } else {
+            final byte[] array = new byte[length];
+            buffer.readBytes(array);
+            return new String(array, charset);
+        }
+    }
+
     private static int firstIndexOf(ByteBuf buffer, int fromIndex, int toIndex, byte value) {
         fromIndex = Math.max(fromIndex, 0);
         if (fromIndex >= toIndex || buffer.capacity() == 0) {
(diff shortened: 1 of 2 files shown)


buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -439,6 +439,29 @@ public static ByteBuf readBytes(ByteBufAllocator alloc, ByteBuf buffer, int leng
...
+            final String result = buffer.toString(buffer.readerIndex(), length, charset);
+            buffer.skipBytes(length);
+            return result;
+        } else if (buffer.hasArray()) {
+            final String result = new String(buffer.array(), buffer.arrayOffset() + buffer.readerIndex(), length, charset);
+            buffer.skipBytes(length);
+            return result;
+        } else {
+            final byte[] array = new byte[length];
+            buffer.readBytes(array);
```
fenik17: This will work only for one-byte encoding.

johnou:
@normanmaurer ptal

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -439,6 +439,30 @@ public static ByteBuf readBytes(ByteBufAllocator alloc, ByteBuf buffer, int leng
...
+     * @param charset The charset of the string to be returned.
+     * @return {@link String} from the given {@link ByteBuf} with the provided length and charset.
+     */
+    public static String readString(ByteBuf buffer, int length, Charset charset) {
+        if (buffer.isDirect()) {
+            final String result = buffer.toString(buffer.readerIndex(), length, charset);
+            buffer.skipBytes(length);
+            return result;
+        } else if (buffer.hasArray()) {
+            final String result = new String(buffer.array(), buffer.arrayOffset() + buffer.readerIndex(), length,
```
Scottmitch: why wouldn't we just use `buffer.toString` in each case? If there is a reason for this perhaps we should update the `buffer.toString` implementations to "do the right thing".

johnou: @Scottmitch because they are two fundamentally different things, toString returns a view of the buffer contents and does not modify the readerIndex or writerIndex while this util method does change the readerIndex.

normanmaurer: @johnou `readCharSequence(...)` ? https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/ByteBuf.java#L1728

johnou: @normanmaurer that might do the trick, I'll benchmark the two and get back to you :ok_hand:

johnou:
```
Benchmark                                           Mode   Cnt         Score        Error  Units
ByteBufStringTest.testHeapBufferReadCharSequence    thrpt  100  13519792.363 ± 103400.876  ops/s
ByteBufStringTest.testHeapBufferReadString          thrpt  100  16900628.210 ± 112470.090  ops/s
ByteBufStringTest.testDirectBufferReadCharSequence  thrpt  100   6652487.724 ± 184603.697  ops/s
ByteBufStringTest.testDirectBufferReadString        thrpt  100   7291389.470 ± 147740.907  ops/s

```

normanmaurer: @johnou I am not convinced we should add these methods. If you really feel the need we can try to improve the others but I am not sure its needed.

johnou: @normanmaurer yeah makes sense to improve the existing method, I'll try and come back to it.

PR #7700, opened 2018-02-09
System property util might return null

johnou:
Motivation:

isAndroid0 should be robust.

Modifications:

yoda equals for string comparison.

Result:

No NPE.


johnou:
@normanmaurer ptal

normanmaurer:
@johnou thanks

Cherry-picked into 4.1 (01e46ed03aea4434299faf4c8596b537b27f767b)

PR #7701, opened 2018-02-09
Reduce the default number of objects retained by the Recycler per thread

Scottmitch:
Motivation:
The Recycler currently retains 32k objects per thread by default. The Recycler is used in more than just one place and may result in large amounts of memory bloat if spikes of traffic are observed.

Modifications:
- Reduce the Recyclers default capacity from 32k to 4k.

Result:
- Lower default capacity of the Recycler and less memory retained.


normanmaurer:
@ejona86 @carl-mastrangelo @mosesn any concerns ?

normanmaurer:
@Scottmitch thanks.... 4.1 (6cd5e8b0ca2df07835e53b649fbf000f15175306)

PR #7702, opened 2018-02-09
Load Conscrypt method via reflection only once

ejona86:
Motivation:

The code did reflection every method call which made the code slower and
harder to read with additional cases to consider.

Modifications:

Instead of loading the method and then throwing it away, save the Method
reference instead of the Class reference. Then also use more precise
exception handling for the method invocation.

Result:

Simpler, speedier code.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/Conscrypt.java
+++ b/handler/src/main/java/io/netty/handler/ssl/Conscrypt.java
@@ -18,6 +18,7 @@ package io.netty.handler.ssl;
 import io.netty.util.internal.PlatformDependent;
 
 import javax.net.ssl.SSLEngine;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 
 /**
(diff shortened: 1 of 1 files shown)


ejona86:
Unfortunately the diff is pretty ugly because I moved the code from getConscryptClass up to the top of the class so that methods called during initialization are not mixed with other static methods.

P.S... _or_ Norman can approve the PR before I write my peace.

ejona86:
Doh. Just realized I hadn't tested this again after the more fine-grain exceptions. It doesn't compile. One moment.

ejona86:
Okay. This time with the missing semicolon. Conscrypt test still runs all 93 tests.

normanmaurer:
When you are already on it can you also do another or to update to 1.0.0 ?

ejona86:
Sure. I'll send out a PR for bumping to Conscrypt 1.0.0.

normanmaurer:
@ejona86 thanks !

Cherry-picked into 4.1 (8b273983f05b75e32353b07bad0dac2ecbbca9c7)

PR #7703, opened 2018-02-09
Bump Conscrypt version to 1.0.0

ejona86:
Motivation:

Conscrypt is now 1.0. No more need to depend on release candidates.

Modifications:

Just the version bump. Things seemed compatible.

Result:

Depending on first guaranteed-api-stable release of Conscrypt.


normanmaurer:
@ejona86 thanks merged!

4.1 (0b0418b3688f5aa3d0ad259d34605e2a6745203c)

PR #7704, opened 2018-02-09
ByteBufUtil to not pool direct memory by default

Scottmitch:
Motivation:
ByteBufUtil by default will cache DirectByteBuffer objects, and the
associated direct memory (up to 64k). In combination with the Recycler which may
cache up to 32k elements per thread may lead to a large amount of direct
memory being retained per EventLoop thread. As traffic spikes come this
may be perceived as a memory leak because the memory in the Recycler
will never be reclaimed.

Modifications:
- By default we shouldn't cache DirectByteBuffer objects.

Result:
Less direct memory consumption due to caching DirectByteBuffer objects.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
+++ b/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
@@ -85,7 +85,7 @@ public final class ByteBufUtil {
 
         DEFAULT_ALLOCATOR = alloc;
 
-        THREAD_LOCAL_BUFFER_SIZE = SystemPropertyUtil.getInt("io.netty.threadLocalDirectBufferSize", 64 * 1024);
+        THREAD_LOCAL_BUFFER_SIZE = SystemPropertyUtil.getInt("io.netty.threadLocalDirectBufferSize", 0);
         logger.debug("-Dio.netty.threadLocalDirectBufferSize: {}", THREAD_LOCAL_BUFFER_SIZE);
 
         MAX_CHAR_BUFFER_SIZE = SystemPropertyUtil.getInt("io.netty.maxThreadLocalCharBufferSize", 16 * 1024);


normanmaurer:
@ejona86 @carl-mastrangelo any concerns ?

johnou:
@Scottmitch what was invoking threadLocalDirectBuffer so often?

Scottmitch:
@johnou - If you are asking was [ByteBufUtil#threadLocalDirectBuffer()](https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java#L722) being called, then yes.

ejona86:
This is a bit harder for me to tell the impact. I didn't realize this was even a thing. Is the idea that the calling code would end up using the normal buffer pooling instead (if configured on the allocator)?

It seems that is the case, except for io.netty.channel.unix.SocketWritableByteChannel which will use an unpooled direct bytebuf. That doesn't sound good in that case.

Scottmitch:
Build failure fixed by PR https://github.com/netty/netty/pull/7705

Scottmitch:
@ejona86 - In a sense, yes. The goal is to avoid pooling direct memory in thread locals by default. The current approach of thread local + Recycler can lead to direct memory leaks because we currently don't provide a way to purge the memory in the Recycler. So it is safer (by default) to leave the direct memory pooling to `PooledByteBufAllocator` and just use Recycler for Object instance pooling.

In regards to [SocketWritableByteChannel](https://github.com/netty/netty/blob/4.1/transport-native-unix-common/src/main/java/io/netty/channel/unix/SocketWritableByteChannel.java#L49) directly using `Unpooled` to allocate that needs more investigation (originates from [here](https://github.com/netty/netty/blame/2d38a4453c360bcd72dbefe4ec04adcdb19f8ddd/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollStreamChannel.java#L1282)), but I'm guessing its because it is created in the `flush` code path which doesn't have access to an allocator. /cc @normanmaurer 

Before this PR and PR https://github.com/netty/netty/pull/7701 ... it may have resulted in 32k direct buffers of up to 64k being retained per EventLoop thread (with no way to purge them).

johnou:
@Scottmitch could use the singleton? PooledByteBufAllocator.DEFAULT

Scottmitch:
@johnou - Yes we could use `PooledByteBufAllocator.DEFAULT`, however ideally we should use the allocator configured by the user.

/cc @normanmaurer - did you see the question in my previous comment?

PR #7705, opened 2018-02-10
NioDatagramChannel invalid usage of internalNioBuffer

Scottmitch:
Motivation:
NioDatagramChannel attempts to unpack a AddressedEnvelope and unconditionally uses internalNioBuffer. However if the ByteBuf is a CompositeByteBuf with more than 1 components, the write will fail and throw an exception.

Modifications:
- NioDatagramChannel should check the nioBufferCount before attempting
to use internalNioBuffer

Result:
No more failure to write UDP packets on NIO when a CompositeByteBuf is
used.

### DIFF
--- a/transport/src/main/java/io/netty/channel/socket/nio/NioDatagramChannel.java
+++ b/transport/src/main/java/io/netty/channel/socket/nio/NioDatagramChannel.java
@@ -286,7 +286,8 @@ public final class NioDatagramChannel
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


Scottmitch:
In order to test this we need https://github.com/netty/netty/pull/7704, and this PR already has unit tests which fail.

Scottmitch:
4.0 (be5b5a3b29)

PR #7711, opened 2018-02-13
Adding option to Http2MultiplexCodecBuilder to propagate SETTINGS fra…

shorea:
Motivation:

Allow the observation of SETTINGS frame by other handlers in the pipeline. For my particular use case this allows me to observe the value of MAX_CONCURRENT_STREAMS for a ChannelPool abstraction that supports HTTP/2 multiplexing. 

Modification:

Add a new option to the Http2MultiplexCodecBuilder to turn settings propagation off/on. The approach taken is similar to InboundHttp2ToHttpAdapter.

Result:

Settings can now be observed in the parent channel. Previously it was not possible (to my knowledge) to capture the settings when using Http2MultiplexCodec.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
@@ -66,7 +66,8 @@ import static java.lang.Math.min;
  * communication, closing of the channel is delayed until any inbound queue is drained with {@link
  * Channel#read()}, which follows the default behavior of channels in Netty. Applications are
  * free to close the channel in response to such events if they don't have use for any queued
- * messages.
+ * messages. Any connection level events like {@link Http2SettingsFrame} and {@link Http2GoAwayFrame}
+ * will be processed internally and also propagated down the pipeline for other handlers to act on.
  *
  * <p>Outbound streams are supported via the {@link Http2StreamChannelBootstrap}.
  *
@@ -164,8 +165,10 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
     // Need to be volatile as accessed from within the DefaultHttp2StreamChannel in a multi-threaded fashion.
     volatile ChannelHandlerContext ctx;
 
-    Http2MultiplexCodec(Http2ConnectionEncoder encoder, Http2ConnectionDecoder decoder, Http2Settings initialSettings,
-                    ChannelHandler inboundStreamHandler) {
+    Http2MultiplexCodec(Http2ConnectionEncoder encoder,
+                        Http2ConnectionDecoder decoder,
+                        Http2Settings initialSettings,
+                        ChannelHandler inboundStreamHandler) {
         super(encoder, decoder, initialSettings);
         this.inboundStreamHandler = inboundStreamHandler;
     }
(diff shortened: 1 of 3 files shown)


codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -223,6 +228,9 @@ final void onHttp2Frame(ChannelHandlerContext ctx, Http2Frame frame) {
             if (settings.initialWindowSize() != null) {
                 initialOutboundStreamWindow = settings.initialWindowSize();
             }
+            if (propagateSettings) {
+                ctx.fireChannelRead(frame);
```
normanmaurer: @Scottmitch @carl-mastrangelo @ejona86 I wonder if we should just do this always and not need a special config for it... /cc @shorea

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodecBuilder.java
```diff
@@ -149,6 +150,18 @@ public Http2MultiplexCodecBuilder initialHuffmanDecodeCapacity(int initialHuffma
         return super.initialHuffmanDecodeCapacity(initialHuffmanDecodeCapacity);
     }
 
+    /**
+     * Determines whether a {@link Http2SettingsFrame} should be propagated in the parent channel pipeline. Default is
+     * false and SETTINGS frames will be processed internally by {@link Http2MultiplexCodec}.
+     *
+     * @param propagateSettings True if settings should be propagated, false otherwise.
```
normanmaurer: nit: True -> `{@code true}`

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodecBuilder.java
```diff
@@ -149,6 +150,18 @@ public Http2MultiplexCodecBuilder initialHuffmanDecodeCapacity(int initialHuffma
         return super.initialHuffmanDecodeCapacity(initialHuffmanDecodeCapacity);
     }
 
+    /**
+     * Determines whether a {@link Http2SettingsFrame} should be propagated in the parent channel pipeline. Default is
+     * false and SETTINGS frames will be processed internally by {@link Http2MultiplexCodec}.
```
normanmaurer: nit: false -> `{@code false}`

normanmaurer: Also maybe make it clear that even if `true` there will also processed internally.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
ejona86: That'd be fair, and would be fine for Http2GoAwayFrame as well (so the idea generalizes).

normanmaurer: Sounds good to me... what do the others think ?

shorea: That sounds good to me, I'll make that change.

normanmaurer:
Thanks a lot

normanmaurer:
@shorea awesome! Can you also sign our ICLA and let me know once done:

https://netty.io/s/icla

@ejona86  @Scottmitch ptal

shorea:
@normanmaurer sure thing, I just need to get internal approval so I'll ping you later today when it's signed.

shorea:
Got internal approval and submitted the ICLA. Ready to merge whenever y'all are.

normanmaurer:
Cherry-picked into 4.1 (650406c0a).

@shorea Thanks for the contribution!

Scottmitch:
lgtm

PR #7714, opened 2018-02-13
enhance some performance

Adam74211:
Motivation:

Modification:

Result:

Fixes #<GitHub issue number>. 


normanmaurer:
@Adam74211 please fix the checkstyle error and only do changes that are necessary (moving imports is not).

normanmaurer:
@Adam74211 ping

normanmaurer:
Did it by myself in https://github.com/netty/netty/pull/7787. Thanks

Issue #7710, opened 2018-02-13
DefaultHeaders does not correctly implement interface


jkinkead:
### Expected behavior

One-argument convert-and-get methods (`get{Byte,Double,Float,Int,Long,Short}`) should return `null` when invoked on a header that can't be converted, per the documentation (e.g. [for `getInt(String)`](https://github.com/netty/netty/blob/4f982be91e6d61c7c7a9d8d63115e0ebe65b2774/codec/src/main/java/io/netty/handler/codec/Headers.java#L175-L176)).

Similarly, two-argument convert-and-get methods should return the default value provided when the header can't be converted.

### Actual behavior

Conversion getter methods throw `NumberFormatException`.

### Steps to reproduce

Set a header to a non-numeric value, then attempt to read it as a number.

### Minimal yet complete reproducer code (or URL to code)

```java
io.netty.handler.codec.http2.DefaultHttp2Headers headers = new io.netty.handler.codec.http2.DefaultHttp2Headers();

headers.set("example", "abc");                                                                  

// Throws NumberFormatException, counter to documentation.    
int value = headers.getInt("example", 123); 
```

### Netty version

4.1

normanmaurer:
Fixed by https://github.com/netty/netty/pull/7715

normanmaurer:
@jkinkead please check https://github.com/netty/netty/pull/7715

jkinkead:
Awesome, thanks for the quick fix! It looks good barring other reviewer's comment (should catch `RuntimeException`).

It might be worth adding docs to [the converter interface](https://github.com/netty/netty/blob/4.1/codec/src/main/java/io/netty/handler/codec/ValueConverter.java) for the exceptions.

normanmaurer:
Fixed by https://github.com/netty/netty/pull/7715

normanmaurer:
I will also update the interface with some docs.

PR #7715, opened 2018-02-13
Correctly handle the case when converting of value fails and return n…

normanmaurer:
…ull or default value.

Motivation:

Headers.get* methods should not throw an exception but return null or the default value if converting of the value fails.

Modifications:

- Correctly handle the case when ValueConverter throws an Exception.
- Add testcase.

Result:

Fixes [#7710].

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/DefaultHeaders.java
+++ b/codec/src/main/java/io/netty/handler/codec/DefaultHeaders.java
@@ -603,7 +603,11 @@ public class DefaultHeaders<K, V, T extends Headers<K, V, T>> implements Headers
     @Override
     public Boolean getBoolean(K name) {
         V v = get(name);
-        return v != null ? valueConverter.convertToBoolean(v) : null;
+        try {
+            return v != null ? valueConverter.convertToBoolean(v) : null;
+        } catch (RuntimeException ignore) {
+            return null;
+        }
     }
 
     @Override
@@ -615,7 +619,11 @@ public class DefaultHeaders<K, V, T extends Headers<K, V, T>> implements Headers
     @Override
     public Byte getByte(K name) {
         V v = get(name);
-        return v != null ? valueConverter.convertToByte(v) : null;
+        try {
+            return v != null ? valueConverter.convertToByte(v) : null;
+        } catch (RuntimeException ignore) {
+            return null;
+        }
     }
 
     @Override
@@ -627,7 +635,11 @@ public class DefaultHeaders<K, V, T extends Headers<K, V, T>> implements Headers
     @Override
     public Character getChar(K name) {
         V v = get(name);
-        return v != null ? valueConverter.convertToChar(v) : null;
+        try {
+            return v != null ? valueConverter.convertToChar(v) : null;
+        } catch (RuntimeException ignore) {
+            return null;
+        }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


codec/src/main/java/io/netty/handler/codec/DefaultHeaders.java
```diff
@@ -603,7 +603,11 @@ public T clear() {
     @Override
     public Boolean getBoolean(K name) {
         V v = get(name);
-        return v != null ? valueConverter.convertToBoolean(v) : null;
+        try {
+            return v != null ? valueConverter.convertToBoolean(v) : null;
+        } catch (Exception ignore) {
```
ejona86: nit: RuntimeException for clarity, although it is equivalent to what's here.

normanmaurer: @ejona86 We are talking about all the `catch` correct ? If so I think that makes sense @carl-mastrangelo @Scottmitch WDYT ?

ejona86: Yeah, I was talking about all of them

Scottmitch: `RuntimeException` sgtm

codec/src/main/java/io/netty/handler/codec/DefaultHeaders.java
```diff
@@ -603,7 +603,11 @@ public T clear() {
     @Override
     public Boolean getBoolean(K name) {
         V v = get(name);
-        return v != null ? valueConverter.convertToBoolean(v) : null;
+        try {
+            return v != null ? valueConverter.convertToBoolean(v) : null;
```
jkinkead: There is another bug here that I haven't filed: `convertToBoolean` has different behavior for `AsciiString` header values than for other `CharSequence` values.

`AsciiString` uses [this null-byte-check conversion](https://github.com/netty/netty/blob/4f982be91e6d61c7c7a9d8d63115e0ebe65b2774/common/src/main/java/io/netty/util/AsciiString.java#L1217-L1219), whereas other strings use [`Boolean.parseBoolean`](https://github.com/netty/netty/blob/4f982be91e6d61c7c7a9d8d63115e0ebe65b2774/codec/src/main/java/io/netty/handler/codec/CharSequenceValueConverter.java#L68-L73), which treats [only strings matching "true"](https://docs.oracle.com/javase/9/docs/api/java/lang/Boolean.html#parseBoolean-java.lang.String-) as truthy (case-insensitive).

normanmaurer: Please fill another bug or do a or to fix it

normanmaurer:
@ejona86 @Scottmitch addressed PTAL again

codec/src/main/java/io/netty/handler/codec/DefaultHeaders.java
jkinkead: Done, issue #7717 . Thanks!

normanmaurer:
Cherry-picked into 4.1 (02b7507a622af049b7061fb2016ebb3b56b32111)

Scottmitch:
lgtm

PR #7720, opened 2018-02-14
DefaultHeaders / CharSequenceValueConverter should treat boolean cons…

normanmaurer:
…istently.

Motivation:

HttpHeaders.getBoolean should return the same truth value for the same string value, regardless of the underlying type.

Modifications:

- Only treat values of true as Boolean.TRUE
- Add unit tests.

Result:

Consistent converting of values for all CharSequence implementations.

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/CharSequenceValueConverter.java
+++ b/codec/src/main/java/io/netty/handler/codec/CharSequenceValueConverter.java
@@ -67,7 +67,11 @@ public class CharSequenceValueConverter implements ValueConverter<CharSequence> 
     @Override
     public boolean convertToBoolean(CharSequence value) {
         if (value instanceof AsciiString) {
-            return ((AsciiString) value).parseBoolean();
+            AsciiString asciiString = (AsciiString) value;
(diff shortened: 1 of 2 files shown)


codec/src/main/java/io/netty/handler/codec/CharSequenceValueConverter.java
```diff
@@ -66,8 +66,15 @@ public CharSequence convertFloat(float value) {
 
     @Override
     public boolean convertToBoolean(CharSequence value) {
+        if (value instanceof String) {
```
ejona86: This case doesn't seem to be all that necessary. Not a big deal though.

normanmaurer: yep we can just remove it I think..

normanmaurer:
Cherry-picked into 4.1 (ad6af3250cfe75f848c7cd7a6cb012172002032c)

Scottmitch:
follow up PR to consider: https://github.com/netty/netty/pull/7728

PR #7721, opened 2018-02-14
Implement DefaultHeaders.HeaderEntry.equals()

ryu1-sakai:
Motivation:

HeaderEntry.equals() inherets Object.equals() which simply check if two objects are the same.
So it returns false even when two HeaderEntry objects have the same name and value.

Modifications:

Implement HeaderEntry.equals() that follows the specification of Map.Entry.equals().
https://docs.oracle.com/javase/9/docs/api/java/util/Map.Entry.html#equals-java.lang.Object-

Result:

HeaderEntry.equals() returns true if two HeaderEntry objects have the same name and value.

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/DefaultHeaders.java
+++ b/codec/src/main/java/io/netty/handler/codec/DefaultHeaders.java
@@ -1170,5 +1170,20 @@ public class DefaultHeaders<K, V, T extends Headers<K, V, T>> implements Headers
         public final String toString() {
             return key.toString() + '=' + value.toString();
         }
+
+        @Override
+        public boolean equals(Object o) {
+            if (!(o instanceof Map.Entry)) {
+                return false;
+            }
+            Map.Entry<?, ?> other = (Map.Entry<?, ?>) o;
+            return (getKey() == null ? other.getKey() == null : getKey().equals(other.getKey()))  &&
+                   (getValue() == null ? other.getValue() == null : getValue().equals(other.getValue()));
+        }
+
+        @Override
+        public int hashCode() {
+            return (key == null ? 0 : key.hashCode()) ^ (value == null ? 0 : value.hashCode());
+        }
     }
 }
--- a/codec/src/test/java/io/netty/handler/codec/DefaultHeadersTest.java
+++ b/codec/src/test/java/io/netty/handler/codec/DefaultHeadersTest.java
@@ -441,6 +441,24 @@ public class DefaultHeadersTest {
     }
 
     @Test
+    public void testEntryEquals() {
+        Map.Entry<CharSequence, CharSequence> same1 = newInstance().add("name", "value").iterator().next();
+        Map.Entry<CharSequence, CharSequence> same2 = newInstance().add("name", "value").iterator().next();
+        assertEquals(same1, same2);
+        assertEquals(same1.hashCode(), same2.hashCode());
+
+        Map.Entry<CharSequence, CharSequence> nameDifferent1 = newInstance().add("name1", "value").iterator().next();
+        Map.Entry<CharSequence, CharSequence> nameDifferent2 = newInstance().add("name2", "value").iterator().next();
+        assertNotEquals(nameDifferent1, nameDifferent2);
+        assertNotEquals(nameDifferent1.hashCode(), nameDifferent2.hashCode());
+
+        Map.Entry<CharSequence, CharSequence> valueDifferent1 = newInstance().add("name", "value1").iterator().next();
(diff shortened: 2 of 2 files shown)


codec/src/main/java/io/netty/handler/codec/DefaultHeaders.java
```diff
@@ -1170,5 +1170,16 @@ public final V setValue(V value) {
         public final String toString() {
             return key.toString() + '=' + value.toString();
         }
+
+        @Override
+        public boolean equals(Object o) {
```
normanmaurer: @ryu1-sakai I think we should also override `hashCode` if we override `equals`.

ryu1-sakai: @normanmaurer thank you for reminding me. Added `hashCode`.

normanmaurer:
@ryu1-sakai please fix check style error and ping me once done.

ryu1-sakai:
@normanmaurer fixed the following error. Thanks!

    DefaultHeaders.java:1180:42: 'cast' is not followed by whitespace.

normanmaurer:
@ryu1-sakai did you sign our ICLA ?: https://netty.io/s/icla

ryu1-sakai:
@normanmaurer ICLA also done. Thank you for the instruction.

codec/src/main/java/io/netty/handler/codec/DefaultHeaders.java
```diff
@@ -1170,5 +1170,21 @@ public final V setValue(V value) {
...
+            if (!(o instanceof Map.Entry)) {
+                return false;
+            }
+            @SuppressWarnings("rawtypes")
+            Map.Entry other = (Map.Entry) o;
```
ejona86: If you can, it's still be better to cast to `Map.Entry<?, ?>`. Removing the generics can lead to surprising cases (none here, but in general strongly avoid rawtypes).

normanmaurer: @ryu1-sakai please do what @ejona86 suggested and ping me once done :)

codec/src/test/java/io/netty/handler/codec/DefaultHeadersTest.java
```diff
@@ -440,6 +440,21 @@ public void iteratorSetValueShouldChangeHeaderValue() {
         assertFalse(headers.contains(of("name1"), of("value3")));
     }
 
+    @Test
+    public void testEntryEquals() {
+        Map.Entry<CharSequence, CharSequence> same1 = newInstance().add("name", "value").iterator().next();
+        Map.Entry<CharSequence, CharSequence> same2 = newInstance().add("name", "value").iterator().next();
+        assertEquals(same1, same2);
```
normanmaurer: @ryu1-sakai can you also add an assert for the hashcode ?

ryu1-sakai:
@normanmaurer I made the following two modifications

* s/`Map.Entry`/`Map.Entry<?, ?>`/
* Added asserts for the `hashCode()` to the test

normanmaurer:
Cherry-picked into 4.1 (c1d0d88f0ae50ce38f00daa9bf7039c1a8bf9a80).

@ryu1-sakai thanks!

codec/src/main/java/io/netty/handler/codec/DefaultHeaders.java
```diff
@@ -1170,5 +1170,20 @@ public final V setValue(V value) {
...
+                return false;
+            }
+            Map.Entry<?, ?> other = (Map.Entry<?, ?>) o;
+            return (getKey() == null ? other.getKey() == null : getKey().equals(other.getKey()))  &&
+                   (getValue() == null ? other.getValue() == null : getValue().equals(other.getValue()));
+        }
+
+        @Override
+        public int hashCode() {
+            return (key == null ? 0 : key.hashCode()) ^ (value == null ? 0 : value.hashCode());
```
Scottmitch: any particular reason for xor vs the more traditional "multiply by prime and add" approach used elsewhere in netty?

ryu1-sakai:
Simply followed the `Map.Entry.hashCode()` specification
https://docs.oracle.com/javase/9/docs/api/java/util/Map.Entry.html#hashCode--

codec/src/test/java/io/netty/handler/codec/DefaultHeadersTest.java
ryu1-sakai: Simply followed the `Map.Entry.hashCode()` specification
https://docs.oracle.com/javase/9/docs/api/java/util/Map.Entry.html#hashCode--

PR #7725, opened 2018-02-16
Get memory address from Unsafe for OpenSSL

carl-mastrangelo:
Motivation:
Profiling tcnative SSL code showed a non trivial percentage (1%)
of time spent in JNI code for InstaceOf.   This turned out to be
from `Buffer.address` which makes a JNI call, which safely checks
on each call that The ByteBuffer is direct.

Modification:
Prefer using the address field of the pojo rather than looking it
up with JNI.  This is the same approach taken by the `OpenSsl`
class.

Result:
Less JNI overhead


normanmaurer:
@carl-mastrangelo thanks a lot... I cherry-picked it after I fixed the one check-style error and run the test suite locally. 

4.1 (07cabb1f55cded9a4648281763bbc4283acb2784)

PR #7730, opened 2018-02-18
Epoll flush/writabilityChange deadlock

### LINKED ISSUE #7729: Downloading large file halts during transfer
### Expected behavior
When downloading a large file (50MB) from a netty based server, I expect to receive the entire file and that the download does not halt during the transfer.

### Actual behavior
Download of file halts during the transfer. Transfer eventually terminates due to idle timeout. 

### Steps to reproduce
1. Setup a Netty HTTP server using Epoll Edge Triggered
2. Set the low water mark to 8192 bytes (8k) and high to 32768 bytes (32k)
3. Transfer the file using 8k chunks, flushing when unwritable

Will try to create one

### Netty version
4.1.21.Final

java version "1.7.0_80"
Java(TM) SE Runtime Environment (build 1.7.0_80-b15)
Java HotSpot(TM) 64-Bit Server VM (build 24.80-b11, mixed mode)

Linux

### Notes
FYI, downgrading to Netty 4.1.17.Final fixes the issue.  I believe the issue has been introduced by https://github.com/netty/netty/commit/b215794de31f28355e4469fcc04782f55076c80c. I use a custom handler to chunk `FullHttpResponse` objects and send them outbound, so it may be an issue with that handler in conjunction with the netty commit mentioned above.

/cc @Scottmitch Since you did the work on that commit

Scottmitch:
Motivation:
b215794 recently introduced a change in behavior where writeSpinCount provided a limit for how many write operations were attempted per flush operation. However when the write quantum was meet the selector write flag was not cleared, and the channel unsafe flush0 method has an optimization which prematurely exits if the write flag is set. This may lead to no write progress being made under the following scenario:
- flush is called, but the socket can't accept all data, we set the write flag
- the selector wakes us up because the socket is writable, we write data and use the writeSpinCount quantum
- we then schedule a flush() on the EventLoop to execute later, however it the flush0 optimization prematurely exits because the write flag is still set

In this scenario the socket is still writable so the EventLoop may never notify us that the socket is writable, and therefore we may never attempt to flush data to the OS.

Modifications:
- When the writeSpinCount quantum is exceeded we should clear the selector write flag

Result:
Fixes #7729

### DIFF
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollChannel.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollChannel.java
@@ -507,14 +507,13 @@ abstract class AbstractEpollChannel extends AbstractChannel implements UnixChann
         }
 
         @Override
-        protected void flush0() {
+        protected final void flush0() {
             // Flush immediately only when there's no pending flush.
             // If there's a pending flush operation, event loop will call forceFlush() later,
             // and thus there's no need to call it now.
-            if (isFlagSet(Native.EPOLLOUT)) {
-                return;
+            if (!isFlagSet(Native.EPOLLOUT)) {
+                super.flush0();
             }
-            super.flush0();
         }
 
         /**
--- a/transport/src/main/java/io/netty/channel/nio/AbstractNioChannel.java
+++ b/transport/src/main/java/io/netty/channel/nio/AbstractNioChannel.java
@@ -356,10 +356,9 @@ public abstract class AbstractNioChannel extends AbstractChannel {
             // Flush immediately only when there's no pending flush.
             // If there's a pending flush operation, event loop will call forceFlush() later,
             // and thus there's no need to call it now.
-            if (isFlushPending()) {
-                return;
+            if (!isFlushPending()) {
+                super.flush0();
             }
-            super.flush0();
         }
 
         @Override
--- a/testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketConditionalWritabilityTest.java
+++ b/testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketConditionalWritabilityTest.java
@@ -1,0 +1,124 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.testsuite.transport.socket;
+
+import io.netty.bootstrap.Bootstrap;
+import io.netty.bootstrap.ServerBootstrap;
+import io.netty.buffer.ByteBuf;
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelDuplexHandler;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.channel.ChannelInitializer;
+import io.netty.channel.ChannelOption;
+import io.netty.channel.WriteBufferWaterMark;
+import io.netty.util.ReferenceCountUtil;
+import org.junit.Test;
+
+import java.util.concurrent.CountDownLatch;
+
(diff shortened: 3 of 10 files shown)


Scottmitch:
@doom369 @blucas - can you verify if this fixes the issues you reported in https://github.com/netty/netty/issues/7729?

doom369:
@Scottmitch works for me. Thanks for the quick fix!

normanmaurer:
@Scottmitch hmmm... can you explain why this was not an issue before your change b215794 ?

blucas:
@Scottmitch - I'll try to get some time tonight to test it.

Scottmitch:
@normanmaurer - prior to b215794 `writeSpinCount` was used on each write operation to try to write all data, and if not all attempted data was written we set`EPOLLOUT` (potentially in multiple places). `writeSpinCount` was not treated as a quantum in the sense that we would only ever attempt `writeSpinCount` writes per-flush, and then try again later (e.g. execute a task on the event loop to try again). It is now possible we write more than the high water mark (the channel doesn't accept data), exhaust the quantum, and then when we try to flush again the `EPOLLOUT` is set. So the key difference is the flush process may be interrupted and tried again later, where previously the `EPOLLOUT` was set and would "back door" `flush0` when ready.

normanmaurer:
@Scottmitch got it... Ok sounds good to me... @blucas let us know once you tried it.

Scottmitch:
@normanmaurer @doom369 @blucas - Note that I updated the PR. The end result should be the same (functionally), but I think this approach is more correct. Would you mind verifying again?

transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollChannel.java
```diff
@@ -507,14 +507,13 @@ EpollRecvByteAllocatorHandle newEpollHandle(RecvByteBufAllocator.ExtendedHandle
...
 
         @Override
-        protected void flush0() {
+        protected final void flush0() {
             // Flush immediately only when there's no pending flush.
             // If there's a pending flush operation, event loop will call forceFlush() later,
             // and thus there's no need to call it now.
-            if (isFlagSet(Native.EPOLLOUT)) {
-                return;
+            if (!isFlagSet(Native.EPOLLOUT)) {
```
normanmaurer: @Scottmitch seems like the code is now basically the same as before here... Consider revert the change of the file ? Seems like the only change that is "new" is the final keyword

transport/src/main/java/io/netty/channel/nio/AbstractNioChannel.java
```diff
@@ -356,10 +356,9 @@ protected final void flush0() {
             // Flush immediately only when there's no pending flush.
             // If there's a pending flush operation, event loop will call forceFlush() later,
             // and thus there's no need to call it now.
-            if (isFlushPending()) {
-                return;
+            if (!isFlushPending()) {
```
normanmaurer: consider reverting this as the code basically does the same as before.

transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollChannel.java
Scottmitch: yes it is the same as before ... just 1 less code and byte code (see below). Maybe the JIT can optimize this ... but simple enough to do it in code.

```java
    public void function1(boolean v) {
        if (!v) {
            return;
        }
        doWork();
    }

    public void function2(boolean v) {
        if (v) {
            doWork();
        }
    }
```

```
  public void function1(boolean);
    Code:
       0: iload_1
       1: ifne          5
       4: return
       5: aload_0
       6: invokevirtual #6                  // Method doWork:()V
       9: return

  public void function2(boolean);
    Code:
       0: iload_1
       1: ifeq          8
       4: aload_0
       5: invokevirtual #6                  // Method doWork:()V
       8: return
```

normanmaurer: @Scottmitch I guess at the end it really not matters.. So up to you.

normanmaurer:
Let us wait for confirmation of @blucas and @doom369 tho before merging.

blucas:
@Scottmitch - I've tested EPOLL Edge Triggered using both the reproducer and my actual app both with and without ssl.  Seems like this has fixed the issue I reported, thank you!
Please note, I have not tested NIO/KQueue implementations of this, so have no idea about if they have the same issue and if you're solution for those transports work.

Scottmitch:
@blucas - thanks for confirming!

@doom369 - can you confirm again after the code changes?

Scottmitch:
@blucas - we are aiming for a release (which includes this fix) on Wednesday.

doom369:
@Scottmitch works fine.

normanmaurer:
Thanks... merged

normanmaurer:
@Scottmitch doh! It seems like the test suite hangs on macOS now... Will investigate

Scottmitch:
KQueue issue fixed by https://github.com/netty/netty/pull/7738

PR #7731, opened 2018-02-19
Fix Snappy decoding of large 2-byte literal lengths and copy offsets

dnault:
Motivation:

The Snappy decoder was failing on valid inputs containing literals
with 2-byte lengths > 0x8000 or copies with 2-byte offsets >= 0x8000.

The decoder was also enforcing an artificially low offset limit of
0x7FFF, something the Snappy format description advises against,
and which prevents decoding valid inputs generated by other encoders.

Modifications:

Interpret 2-byte literal lengths and 2-byte copy offsets as unsigned
shorts, in accordance with the format description and reference
implementation.

Allow any positive offset value. Throw an appropriate exception
for negative values (which can theoretically occur due to arithmetic
overflow on 4-byte offsets, but are unlikely to occur in the wild).

Result:

The Snappy decoder can handle valid inputs that previously caused
it to throw exceptions.

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/compression/Snappy.java
+++ b/codec/src/main/java/io/netty/handler/codec/compression/Snappy.java
@@ -403,7 +403,7 @@ public final class Snappy {
             if (in.readableBytes() < 2) {
                 return NOT_ENOUGH_INPUT;
             }
-            length = in.readShortLE();
+            length = in.readUnsignedShortLE();
             break;
         case 62:
             if (in.readableBytes() < 3) {
@@ -495,7 +495,7 @@ public final class Snappy {
 
         int initialIndex = out.writerIndex();
         int length = 1 + (tag >> 2 & 0x03f);
-        int offset = in.readShortLE();
+        int offset = in.readUnsignedShortLE();
 
         validateOffset(offset, writtenSoFar);
 
@@ -565,7 +565,7 @@ public final class Snappy {
 
     /**
      * Validates that the offset extracted from a compressed reference is within
-     * the permissible bounds of an offset (4 <= offset <= 32768), and does not
+     * the permissible bounds of an offset (0 < offset < Integer.MAX_VALUE), and does not
      * exceed the length of the chunk currently read so far.
      *
      * @param offset The offset extracted from the compressed reference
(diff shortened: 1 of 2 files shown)


normanmaurer:
@dnault can you please sign the ICLA: https://netty.io/s/icla ?

dnault:
@normanmaurer Signed.

normanmaurer:
@lw346 can you have a look ? Also @idelpivnitskiy

codec/src/main/java/io/netty/handler/codec/compression/Snappy.java
```diff
@@ -565,20 +565,21 @@ private static int decodeCopyWith4ByteOffset(byte tag, ByteBuf in, ByteBuf out,
 
     /**
      * Validates that the offset extracted from a compressed reference is within
-     * the permissible bounds of an offset (4 <= offset <= 32768), and does not
+     * the permissible bounds of an offset (0 < offset < Integer.MAX_VALUE), and does not
```
lw346: The reason for it limiting to 32768 is as per https://github.com/google/snappy/blob/master/format_description.txt#L79:
> Note that since the current Snappy compressor works in 32 kB blocks and does not do matching across blocks, it will never produce a bitstream with offsets larger than about 32768.

However it then goes on to say
> However, the decompressor should not rely on this, as it may change in the future

So I agree generally that we shouldn't be relying on this but would go on to question the behaviour of other encoders as not being per the 32 kB block chunking spec :)

codec/src/main/java/io/netty/handler/codec/compression/Snappy.java
```diff
@@ -495,7 +495,7 @@ private static int decodeCopyWith2ByteOffset(byte tag, ByteBuf in, ByteBuf out,
 
         int initialIndex = out.writerIndex();
         int length = 1 + (tag >> 2 & 0x03f);
-        int offset = in.readShortLE();
+        int offset = in.readUnsignedShortLE();
```
lw346: The source text is ambiguous on whether these are signed or not, but given it wouldn't make sense to have an offset that looked ahead of the current block I think it is reasonable to assume these are unsigned 16-bit integers.

normanmaurer:
@dnault thanks... merged

PR #7735, opened 2018-02-19
Add an assert to ensure we not leak in the future.

normanmaurer:
Motivation:

We should add an assert to ensure we not leak in the future as we not do any special handling if trySuccess(...) returns false.

Modifications:

Add assert.

Result:

More robust code.

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
@@ -32,6 +32,7 @@ import io.netty.handler.codec.dns.DnsResponse;
 import io.netty.handler.codec.dns.DnsResponseCode;
 import io.netty.handler.codec.dns.DnsSection;
 import io.netty.util.ReferenceCountUtil;
+import io.netty.util.ReferenceCounted;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.FutureListener;
 import io.netty.util.concurrent.Promise;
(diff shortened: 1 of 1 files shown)


resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverContext.java
```diff
@@ -139,7 +140,13 @@ void resolve(final Promise<T> promise) {
                 public void operationComplete(Future<T> future) throws Exception {
                     Throwable cause = future.cause();
                     if (cause == null) {
-                        promise.trySuccess(future.getNow());
+                        T result = future.getNow();
+
+                        // Assert that we will not use any ReferenceCounted result here as trySuccess(...) may
+                        // return false in which case the ownership would not be transferred and special
+                        // handling would be required. This is not the case in our current implementation.
+                        assert !(result instanceof ReferenceCounted);
```
Scottmitch: `T` may also be a `List` if we are resolving all addresses. Not sure if we should account/care about collections if we are trying to add some protection for reference counting?

normanmaurer: I just wanted to keep it simple for now... I think we not need to also try to walk a List etc.

normanmaurer: @Scottmitch WDYT ?

normanmaurer:
Let me just close this for now...

PR #7737, opened 2018-02-20
Always start with the assumption that the writefilter is not enabled.

normanmaurer:
Motivation:

ce241bd11ec26008a692b87391b94d8efe69069d introduced code to not try flushing data to the socket when the writefilter is enabled. The problem with this was that we sometimes created an AbstractKqueueChannel that started with the writeFilter enabled and so we never had the chance to at least try to flush one time.

Modifications:

- Always start with the assumption that the write filter is not enabled. At worse we will try to write one time and then enable it if we can not flush out everything.
- Remove now unused constructor.

Result:

Fixes test timeout on macOS.


Scottmitch:
@normanmaurer - I close we fix this in favor of https://github.com/netty/netty/pull/7738

normanmaurer:
Yep sounds good as discussed offline

PR #7739, opened 2018-02-21
Copies from read-only heap ByteBuffer to direct ByteBuf can avoid stealth ByteBuf allocation and additional copies

franz1981:
Motivation:

Read-only heap ByteBuffer doesn't expose array: the existent method to perform copies to direct ByteBuf involves the creation of a (maybe pooled) additional heap ByteBuf instance and copy

Modifications:

To avoid stressing the allocator with additional (and stealth) heap ByteBuf allocations is provided a method to perform copies using the (pooled) internal NIO buffer

Result:

Copies from read-only heap ByteBuffer to direct ByteBuf won't create any intermediate ByteBuf

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
+++ b/buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
@@ -538,32 +538,46 @@ final class UnsafeByteBufUtil {
     }
 
     static void setBytes(AbstractByteBuf buf, long addr, int index, ByteBuffer src) {
-        buf.checkIndex(index, src.remaining());
-
-        int length = src.remaining();
+        final int length = src.remaining();
         if (length == 0) {
             return;
         }
 
         if (src.isDirect()) {
+            buf.checkIndex(index, length);
             // Copy from direct memory
             long srcAddress = PlatformDependent.directBufferAddress(src);
-            PlatformDependent.copyMemory(srcAddress + src.position(), addr, src.remaining());
+            PlatformDependent.copyMemory(srcAddress + src.position(), addr, length);
             src.position(src.position() + length);
         } else if (src.hasArray()) {
+            buf.checkIndex(index, length);
             // Copy from array
             PlatformDependent.copyMemory(src.array(), src.arrayOffset() + src.position(), addr, length);
             src.position(src.position() + length);
         } else {
-            ByteBuf tmpBuf = buf.alloc().heapBuffer(length);
-            try {
-                byte[] tmp = tmpBuf.array();
-                src.get(tmp, tmpBuf.arrayOffset(), length); // moves the src position too
-                PlatformDependent.copyMemory(tmp, tmpBuf.arrayOffset(), addr, length);
-            } finally {
-                tmpBuf.release();
+            if (length < 8) {
+                setSingleBytes(buf, addr, index, src, length);
+            } else {
+                //no need to checkIndex: internalNioBuffer is already taking care of it
+                assert buf.nioBufferCount() == 1;
+                final ByteBuffer internalBuffer = buf.internalNioBuffer(index, length);
+                internalBuffer.put(src);
             }
         }
+    }
+
+    private static void setSingleBytes(final AbstractByteBuf buf, final long addr, final int index,
+                                       final ByteBuffer src, final int length) {
+        buf.checkIndex(index, length);
+        final int srcPosition = src.position();
+        final int srcLimit = src.limit();
+        long dstAddr = addr;
+        for (int srcIndex = srcPosition; srcIndex < srcLimit; srcIndex++) {
+            final byte value = src.get(srcIndex);
+            PlatformDependent.putByte(dstAddr, value);
+            dstAddr++;
+        }
+        src.position(srcLimit);
     }
 
     static void getBytes(AbstractByteBuf buf, long addr, int index, OutputStream out, int length) throws IOException {
--- a/microbench/src/main/java/io/netty/microbench/buffer/ByteBufCopyBenchmark.java
+++ b/microbench/src/main/java/io/netty/microbench/buffer/ByteBufCopyBenchmark.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.microbench.buffer;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.PooledByteBufAllocator;
+import io.netty.buffer.Unpooled;
+import io.netty.microbench.util.AbstractMicrobenchmark;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.TearDown;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+public class ByteBufCopyBenchmark extends AbstractMicrobenchmark {
+    static {
+        System.setProperty("io.netty.buffer.bytebuf.checkAccessible", "false");
+    }
+
+    @Param({"7", "36", "128", "512" })
+    private int size;
+    @Param({"true", "false" })
+    private boolean directByteBuff;
+    @Param({"true", "false" })
+    private boolean directByteBuffer;
+    @Param({"false", "true" })
+    private boolean readonlyByteBuffer;
+    @Param({"true", "false" })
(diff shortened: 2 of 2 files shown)


franz1981:
I've added a benchmark to cover most of the combinations of ByteBuf/ByteBuffer types and performing long-stepped copies seems a good deal to avoid stressing the Netty allocator (or GC if no allocator is involved)

franz1981:
Some benchmark results
**4.1**
```
Benchmark                      (directByteBuff)  (directByteBuffer)  (longAlignedCopyByteBuf)  (longAlignedCopyByteBuffer)  (nativeOrderByteBuffer)  (pooledByteBuf)  (readonlyByteBuffer)  (size)   Mode  Cnt         Score         Error  Units
ByteBufCopyBenchmark.setBytes              true               false                      true                         true                     true             true                  true       7  thrpt   20  12687942.561 ±  350323.974  ops/s
ByteBufCopyBenchmark.setBytes              true               false                      true                         true                     true             true                  true      36  thrpt   20  12051904.908 ±  807558.702  ops/s
ByteBufCopyBenchmark.setBytes              true               false                      true                         true                     true             true                  true     128  thrpt   20  12021786.615 ±  833817.788  ops/s
ByteBufCopyBenchmark.setBytes              true               false                      true                         true                     true             true                  true     512  thrpt   20  10676406.038 ± 1049181.868  ops/s
```
**this PR**
```
Benchmark                      (directByteBuff)  (directByteBuffer)  (longAlignedCopyByteBuf)  (longAlignedCopyByteBuffer)  (nativeOrderByteBuffer)  (pooledByteBuf)  (readonlyByteBuffer)  (size)   Mode  Cnt         Score         Error  Units
ByteBufCopyBenchmark.setBytes              true               false                      true                         true                     true             true                  true       7  thrpt   20  60708777.468 ± 3154109.733  ops/s
ByteBufCopyBenchmark.setBytes              true               false                      true                         true                     true             true                  true      36  thrpt   20  23655322.019 ±  403197.526  ops/s
ByteBufCopyBenchmark.setBytes              true               false                      true                         true                     true             true                  true     128  thrpt   20  13444496.452 ±   80783.990  ops/s
ByteBufCopyBenchmark.setBytes              true               false                      true                         true                     true             true                  true     512  thrpt   20   3991287.147 ±   92231.444  ops/s
```
Starting from 512 bytes the performance is better in the original version (if an pooled allocator is involved), but that's valid just with (this) microbenchmark: with a real application where the allocator would be already stressed doing other things, the risk of falling back from the thread local buffers would make the results very different/less predicatable.
I can somehow simulate it with an ad hoc concurrent benchmark if is needed.
It is strange that I'm pushing a PR that contains an evident regression in performance :)

buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
```diff
@@ -538,30 +538,38 @@ static void setBytes(AbstractByteBuf buf, long addr, int index, byte[] src, int
     }
 
     static void setBytes(AbstractByteBuf buf, long addr, int index, ByteBuffer src) {
-        buf.checkIndex(index, src.remaining());
```
normanmaurer: @franz1981 I think this should always come first.

microbench/src/main/java/io/netty/microbench/buffer/ByteBufCopyBenchmark.java
```diff
@@ -0,0 +1,102 @@
+/*
+ * Copyright 2015 The Netty Project
```
normanmaurer: 2018

buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
```diff
@@ -538,30 +538,38 @@ static void setBytes(AbstractByteBuf buf, long addr, int index, byte[] src, int
...
-                src.get(tmp, tmpBuf.arrayOffset(), length); // moves the src position too
-                PlatformDependent.copyMemory(tmp, tmpBuf.arrayOffset(), addr, length);
-            } finally {
-                tmpBuf.release();
+            if (length < 8) {
+                buf.checkIndex(index, length);
+                final int srcPosition = src.position();
+                int srcIndex = srcPosition;
+                long dstAddr = addr;
+                for (int i = 0; i < length; i++) {
```
normanmaurer: You could also just use the srcIndex to calculate how often you need to loop. This way you would safe one increment op.

buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
```diff
@@ -538,30 +538,38 @@ static void setBytes(AbstractByteBuf buf, long addr, int index, byte[] src, int
...
             src.position(src.position() + length);
         } else {
-            ByteBuf tmpBuf = buf.alloc().heapBuffer(length);
-            try {
-                byte[] tmp = tmpBuf.array();
-                src.get(tmp, tmpBuf.arrayOffset(), length); // moves the src position too
-                PlatformDependent.copyMemory(tmp, tmpBuf.arrayOffset(), addr, length);
-            } finally {
-                tmpBuf.release();
+            if (length < 8) {
```
normanmaurer: @franz1981 I wonder if we want to make this whole `if` branch code into an extra method which we just call to make the rest small enough to be inlined (as I think this is the more likely hot path).

buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
```diff
@@ -538,30 +538,38 @@ static void setBytes(AbstractByteBuf buf, long addr, int index, byte[] src, int
...
+                long dstAddr = addr;
+                for (int i = 0; i < length; i++) {
+                    final byte value = src.get(srcIndex);
+                    PlatformDependent.putByte(dstAddr, value);
+                    srcIndex++;
+                    dstAddr++;
+                }
+                src.position(srcPosition + length);
+            } else {
+                final ByteBuffer internalBuffer = buf.internalNioBuffer(index, length);
```
normanmaurer: @franz1981 can we add an `assert` that buf.nioBufferCount() == 1 ?

microbench/src/main/java/io/netty/microbench/buffer/ByteBufCopyBenchmark.java
franz1981: Uh vintage, sorry :P

buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
franz1981: Let me check the size of the <8 byte branch first :+1:

buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
franz1981: I was thinking the same but it could be performed twice, because internalNio etc etc is doing the same

normanmaurer: Add a comment please then.

normanmaurer:
@franz1981 let me know once you addressed it.

buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
franz1981: @normanmaurer Added on `setSingleBytes`

franz1981:
@normanmaurer I've addressed all the points :+1:

buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
```diff
@@ -538,34 +538,48 @@ static void setBytes(AbstractByteBuf buf, long addr, int index, byte[] src, int
...
-                PlatformDependent.copyMemory(tmp, tmpBuf.arrayOffset(), addr, length);
-            } finally {
-                tmpBuf.release();
+            if (length < 8) {
+                setSingleBytes(buf, addr, index, src, length);
+            } else {
+                //no need to checkIndex: internalNioBuffer is already taking care of it
+                assert buf.nioBufferCount() == 1;
+                final ByteBuffer internalBuffer = buf.internalNioBuffer(index, length);
+                internalBuffer.put(src);
```
ejona86: Update src.position?

franz1981: AFAIK `ByteBuffer::put(ByteBuffer)` will do it

ejona86: The ByteBuffer has a copy of the indices which will be modified, but that won't be reflected in the ByteBuf.

ejona86:
CC @zpencer

buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
franz1981: And that's seems ok to me: `buf` indices are not supposed to be modified

ejona86: Oh, you're right. I had it inverted in my mind; we're reading from a ByteBuffer, not a ByteBuf.

franz1981: np ^^ I often cross my eyes on `ByteBufer::put` instead: why it isn't named as `ByteBufer::readFrom` or something more human readable? :(

normanmaurer:
Cherry-picked into 4.1 (ed46c4ed00ec8260d64eda19ae05aae03fcdaa20)

normanmaurer:
@franz1981 thanks!

PR #7743, opened 2018-02-22
Flush task should not flush messages that were written since last flu…

normanmaurer:
…sh attempt.

Motivation:

The flush task is currently using flush() which will have the affect of have the flush traverse the whole ChannelPipeline and also flush messages that were written since we gave up flushing. This is not really correct as we should only continue to flush messages that were flushed at the point in time when the flush task was submitted for execution if the user not explicit call flush() by him/herself.

Modification:

Call *Unsafe.flush0() via the flush task which will only continue flushing messages that were marked as flushed before.

Result:

More correct behaviour when the flush task is used.

### DIFF
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollStreamChannel.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/AbstractEpollStreamChannel.java
@@ -73,7 +73,9 @@ public abstract class AbstractEpollStreamChannel extends AbstractEpollChannel im
     private final Runnable flushTask = new Runnable() {
         @Override
         public void run() {
-            flush();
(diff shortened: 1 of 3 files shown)


normanmaurer:
I am still thinking about how to add a proper test... Please review in the meantime tho

Scottmitch:
unit test could involve using writing write quota elements (taking into account gathering writes) flushing, then writing 1 more element and not flushing and verify the last element is not actually flushed.  You will have to be careful to defeat gathering writes, or limit how much data a gathering write is allowed to write somehow.

normanmaurer:
@Scottmitch I had no time yet for a test... are you cool with merge it and do a test as a followup ... I am not sure when I have time as the test is a bit tricky :(

Scottmitch:
@normanmaurer - sure that is fine. I added tests previously to ensure the flush is actually done.

PR #7745, opened 2018-02-22
Make WebSocketServerHandshaker injectable in WebSocketServerProtocolH…

### LINKED ISSUE #7742: WebSocket server handshaker should be injectable.
While a websocket *client* handshaker can be injected in `WebSocketClientProtocolHandler`, a websocket server handshaker cannot in `WebSocketServerProtocolHandler`. 
Is it ok to make it injectable? I will create a PR if there's no problem.

yueki1993:
…andler and WebSocketServerProtocolHandshakeHandler.

Motivation:
See also #7742.
At present, WebSocketServerHandshaker is not injectable in WebSocketServerProtocolHandler and WebSocketServerProtocolHandshakeHandler.
When we want to implement an original WebSocketServerHandshaker, we need heavy lifting or copy-and-paste work.
This PR enables us to use easily our own websocket server handshakers.

Modification:
- Add constractors to WebSocketServerProtocolHandler and WebSocketServerProtocolHandshakeHandler that are passed a WebSocketServerHandshaker object.
- Due to the above modification, some final fields may not be initialized at constractors. Therefore I make them non-final variables.

Result:
Fixes #7742 .

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketServerProtocolHandshakeHandler.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketServerProtocolHandshakeHandler.java
@@ -43,6 +43,7 @@ class WebSocketServerProtocolHandshakeHandler extends ChannelInboundHandlerAdapt
     private final int maxFramePayloadSize;
     private final boolean allowMaskMismatch;
     private final boolean checkStartsWith;
+    private WebSocketServerHandshaker handshaker;
 
     WebSocketServerProtocolHandshakeHandler(String websocketPath, String subprotocols,
             boolean allowExtensions, int maxFrameSize, boolean allowMaskMismatch) {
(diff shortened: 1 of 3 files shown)


yueki1993:
sorry. i realized my PR contains a bug. i'll fix it and commit again

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketServerProtocolHandshakeHandler.java
```diff
@@ -38,11 +38,22 @@
...
 
     private final String websocketPath;
-    private final String subprotocols;
-    private final boolean allowExtensions;
-    private final int maxFramePayloadSize;
-    private final boolean allowMaskMismatch;
+    private String subprotocols;
+    private boolean allowExtensions;
+    private int maxFramePayloadSize;
+    private boolean allowMaskMismatch;
```
normanmaurer: these should stay final

yueki1993:
thanks for your review! I will fix them tonight.

yueki1993:
@normanmaurer I fixed them. Can you review this again please?

normanmaurer:
@yueki1993 after thinking more about this I don't like the approach. I will come up with another PR to address this. Thanks for bringing this up tho!

PR #7746, opened 2018-02-22
Close SSLEngine when connection fails.

carl-mastrangelo:
Motivation:
When using the JdkSslEngine, the ALPN class is used keep a reference
to the engine.   In the event that the TCP connection fails, the
SSLEngine is not removed from the map, creating a memory leak.

Modification:
Always close the SSLEngine regardless of if the channel became
active.  Also, record the SSLEngine was closed in all places.

Result:
Fixes: https://github.com/grpc/grpc-java/issues/3080

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -1510,6 +1510,7 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
         try {
             // Release all resources such as internal buffers that SSLEngine
             // is managing.
+            outboundClosed = true;
             engine.closeOutbound();
 
             if (closeInbound) {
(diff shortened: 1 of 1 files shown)


carl-mastrangelo:
I am not sure how to test this, let me know if there is a good place to do so.

normanmaurer:
@carl-mastrangelo Maybe create an SslHandler that wraps a custom `SSLEngine` and assert that `closeOutbound` was called for the case where the `Channel` was not active when `closeOutboundAndChannel` was called ?

normanmaurer:
@carl-mastrangelo does my suggestion make sense ?

carl-mastrangelo:
@normanmaurer yes, sorry for the delay, Intellij complains about the conscrypt dependency in netty now, so I can't run the test from the IDE any more.  This has made it harder to make changes.  

The custom SSLEngine is slightly annoying, since it has a ton of methods to override.

normanmaurer:
Let me just merge this and think about a followup for a testcase

PR #7750, opened 2018-02-26
add test case for readerIndex larger than writerIndex

kakashiio:
Motivation:

Show how to breaks the rule between readerIndex and writerIndex.

Modification:

Fixed the resetWriterIndex() which mentioned in  #7653


normanmaurer:
Thanks merged with https://github.com/netty/netty/pull/7653 and opened:

See https://github.com/netty/netty/pull/7751

PR #7755, opened 2018-02-28
Issue 7754

### LINKED ISSUE #7754: Feature Request: ability to set/get ResourceLeakDetector's targetRecords programmatically
### Description

Currently ResourceLeakDetector's leak detection target records number can be set as a JVM option: io.netty.leakDetection.targetRecords. Would it be possible to provide a programmatic way to set this, i.e., a setter method: 

ResourceLeakDetector.setTargetRecords(int)

This is similar to how the leak detection level can be set using a JVM option as well as using a setter.

Also a getter: ResourceLeakDetector.getTargetRecords() would be useful.

### Netty version
4.1.22.Final

kiril-me:
Motivation:

Make the programmatic change of ReasourceLeakDetector target records.

Modification:

Target records setter and getter was introduced.

Result:

The target record possible to change from java program.

Fixes #7754. 


common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -411,7 +425,7 @@ public void record(Object hint) {
          */
         private void record0(Object hint) {
             // Check TARGET_RECORDS > 0 here to avoid similar check before remove from and add to lastRecords
-            if (TARGET_RECORDS > 0) {
+            if (targetRecords > 0) {
```
normanmaurer: As this might change during the method I think we should cache it first.

normanmaurer: @kiril-me ping me once done

kiril-me:
@normanmaurer done!

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -91,6 +89,8 @@ static Level parseLevel(String levelStr) {
 
     private static Level level;
 
+    private static int targetRecords;
```
Scottmitch: is this intentionally not volatile? we may have visibility issues here, and the volatile read may increase cost of leak detection.

normanmaurer:
Yes I think for the same reason as level

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -158,6 +158,20 @@ public static Level getLevel() {
         return level;
     }
 
+    /**
+     * Sets the resource leak detection target records.
+     */
+    public static void setTargetRecords(int targetRecords) {
+        ResourceLeakDetector.targetRecords = targetRecords;
```
carl-mastrangelo: should it be relogged here since its changing?

common/src/main/java/io/netty/util/ResourceLeakDetector.java
carl-mastrangelo: Being able to change the leak level dynamically seems like an expensive ability for not much payoff.   I would feel more comfortable if I knew the expected use case.

normanmaurer: @kiril-me can you give some details ?

Issue #7758, opened 2018-03-01
Handshake failed on netty 4.1.21 running under Android 5.0


MarkVilkel:
Hi!
I've got an exception on Android 5.0 with netty 4.1.21:

```
javax.net.ssl.SSLHandshakeException: Handshake failed
    io.netty.handler.codec.DecoderException: javax.net.ssl.SSLHandshakeException: Handshake failed
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:459) [na:0.0]
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:265) [na:0.0]
Caused by: javax.net.ssl.SSLHandshakeException: Handshake failed
	at com.android.org.conscrypt.OpenSSLEngineImpl.unwrap(OpenSSLEngineImpl.java:436) [na:0.0]
    at javax.net.ssl.SSLEngine.unwrap(SSLEngine.java:1006) [na:0.0]
Caused by: javax.net.ssl.SSLProtocolException: SSL handshake terminated: ssl=0x9d892e00: Failure in SSL library, usually a protocol error
                                                                             error:1408E0F4:SSL routines:SSL3_GET_MESSAGE:unexpected message (external/openssl/ssl/s3_both.c:498 0xac8b4ce0:0x00000000)
	at com.android.org.conscrypt.NativeCrypto.SSL_do_handshake_bio(Native Method) [na:0.0]
    at com.android.org.conscrypt.OpenSSLEngineImpl.unwrap(OpenSSLEngineImpl.java:423) [na:0.0]
```
So Android app is unable to establish ssl connection.

I have had the same error when we were using mina library, however I've solved it by adding:
```
            if (!inNetBuffer.hasRemaining()) {
                res = new SSLEngineResult(SSLEngineResult.Status.BUFFER_UNDERFLOW, SSLEngineResult.HandshakeStatus.NEED_UNWRAP, 0, 0);

```
like it was suggested by Internets in org.apache.mina.filter.support.SSLHandler.unwrap0() method:
```
    private SSLEngineResult unwrap0() throws SSLException {
        SSLEngineResult res;
        do {
            if (SessionLog.isDebugEnabled(session)) {
                SessionLog.debug(session, "   inNetBuffer: " + inNetBuffer);
                SessionLog.debug(session, "   appBuffer: " + appBuffer);
            }
            if (!inNetBuffer.hasRemaining()) {
[...]

PR #7761, opened 2018-03-02
Workaround SSLEngine.unwrap(...) bug in Android 5.0

normanmaurer:
Motivation:

Android 5.0 sometimes not correctly update the bytesConsumed of the SSLEngineResult when consuming data from the input ByteBuffer. This will lead to handshake failures.

Modifications:

Add a workaround for Android 5.0

Result:

Be able to use netty on Android 5.0 by fixing https://github.com/netty/netty/issues/7758 .

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -289,9 +289,26 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
             SSLEngineResult unwrap(SslHandler handler, ByteBuf in, int readerIndex, int len, ByteBuf out)
                     throws SSLException {
                 int writerIndex = out.writerIndex();
-                final SSLEngineResult result = handler.engine.unwrap(toByteBuffer(in, readerIndex, len),
+                ByteBuffer inNioBuffer = toByteBuffer(in, readerIndex, len);
+                int position = inNioBuffer.position();
+                final SSLEngineResult result = handler.engine.unwrap(inNioBuffer,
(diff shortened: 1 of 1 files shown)


normanmaurer:
@MarkVilkel  reported it fixes his issues 👍

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -289,9 +289,26 @@ boolean jdkCompatibilityMode(SSLEngine engine) {
...
                 out.writerIndex(writerIndex + result.bytesProduced());
+
+                // This is a workaround for a bug in Android 5.0. Android 5.0 does not correctly update the
+                // SSLEngineResult.bytesConsumed() in some cases and just return 0.
+                //
+                // See:
+                //     - https://android-review.googlesource.com/c/platform/external/conscrypt/+/122080
+                //     - https://github.com/netty/netty/issues/7758
+                if (result.bytesConsumed() == 0) {
+                    int consumed = inNioBuffer.position() - position;
```
carl-mastrangelo: `assert PlatformDependent.isAndroid();` 

?

normanmaurer: I can do this ... that said I am not sure its necessary. @carl-mastrangelo if you are strong on it I will add it.... WDYT ?

carl-mastrangelo: Im not keen on it.

normanmaurer: ok let me merge it then

PR #7763, opened 2018-03-02
EmptyHeaders get with default value returns null

Scottmitch:
Motivation:
EmptyHeaders#get with a default value argument returns null. It should never return null, and instead it should return the default value.

Modifications:
- EmptyHeaders#get with a default value should return that default value

Result:
More correct implementation of the Headers API.


normanmaurer:
@Scottmitch I know its boring but can you add a test ? Maybe also for the other methods :)

normanmaurer:
@Scottmitch ping

PR #7770, opened 2018-03-05
Add testcase for c11b23bbc1f46116407bdc10779c4cace639ee83

normanmaurer:
Motivation:

c11b23bbc1f46116407bdc10779c4cace639ee83 added a fix for closing the SSLEngine otbound but no test was provided.

Modifications:

Add testcase.

Result:

More tests.


normanmaurer:
@carl-mastrangelo here is a test for your fix... PTAL

Issue #7752, opened 2018-02-27
IovArray.add(ByteBuf) throws if buffer is a composite of a single direct buffer


ldaley:
### Expected behavior

Not throwing unhandled `UnsupportedOperationException`.

### Actual behavior

Method throws unhandled `UnsupportedOperationException` which causes channel write operations to fail.

### Steps to reproduce

The problem is here: https://github.com/netty/netty/blob/4.1/transport-native-unix-common/src/main/java/io/netty/channel/unix/IovArray.java#L96

If `buf` is a `FixedCompositeByteBuf` with one direct buffer component, it will return `nioBufferCount() == 1` and `memoryAddress()` will throw `UnsupportedOperationException`. 

I can see two possible, complimentary fixes:

1. Make `memoryAddress()` work for composites that are one direct buffer
2. Change this if branch condition to `buf.nioBufferCount() == 1 && buf.hasMemoryAddress()`

Here's a stack trace showing the problem:

```
io.netty.channel.socket.ChannelOutputShutdownException: Channel output shutdown
	at io.netty.channel.AbstractChannel$AbstractUnsafe.shutdownOutput(AbstractChannel.java:641)
	at io.netty.channel.AbstractChannel$AbstractUnsafe.flush0(AbstractChannel.java:948)
	at io.netty.channel.epoll.AbstractEpollChannel$AbstractEpollUnsafe.flush0(AbstractEpollChannel.java:543)
	at io.netty.channel.AbstractChannel$AbstractUnsafe.flush(AbstractChannel.java:901)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.flush(DefaultChannelPipeline.java:1321)
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush0(AbstractChannelHandlerContext.java:776)
	at io.netty.channel.AbstractChannelHandlerContext.invokeFlush(AbstractChannelHandlerContext.java:768)
	at io.netty.channel.AbstractChannelHandlerContext.flush(AbstractChannelHandlerContext.java:749)
	at io.netty.channel.CombinedChannelDuplexHandler$DelegatingChannelHandlerContext.flush(CombinedChannelDuplexHandler.java:533)
	at io.netty.channel.ChannelOutboundHandlerAdapter.flush(ChannelOutboundHandlerAdapter.java:115)
	at io.netty.channel.CombinedChannelDuplexHandler.flush(CombinedChannelDuplexHandler.java:358)
[...]

PR #7773, opened 2018-03-10
FixedCompositeByteBuf should allow to access memoryAddress / array wh…

normanmaurer:
…en wrap a single buffer.

Motivation:

We should allow to access the memoryAddress / array of the FixedCompositeByteBuf when it only wraps a single ByteBuf. We do the same for CompositeByteBuf.

Modifications:

- Check how many buffers FixedCompositeByteBuf wraps and depending on it delegate the access to the memoryAddress / array
- Add unit tests.

Result:

Fixes [#7752].


normanmaurer:
Also @ldaley

carl-mastrangelo:
Was the branch `fixed_buffer_memoryaddress` supposed to stay around?

normanmaurer:
@carl-mastrangelo nope... deleted. Thanks

PR #7774, opened 2018-03-10
IovArray.add(...) should check if buffer has memory address.

normanmaurer:
Motivation:

We currently not check if the buffer has a memory address and just assume this is the case if the nioBufferCount() == 1.

Modifications:

- Check hasMemoryAddress() before trying to access it.
- Add unit case.

Result:

More correct and robust code. Related to [#7752].


normanmaurer:
Also @ldaley

ldaley:
👍

PR #7778, opened 2018-03-13
Set stream ID to the real value in Http2StreamFrameToHttpObjectCodec

jprante:
Motivation:

Http2StreamFrameToHttpObjectCodec is useful to convert HTTP/2 streams to Netty's HTTP requests/responses. But the HTTP/2 stream ID is always 0 after conversion.

Modification:

Instead of always using 0 for the stream ID,  the stream ID from frame.stream().id() is used.

Result:

The stream ID is set from the HTTP/2 stream and retained after conversion. This can be useful for users who wants to track HTTP/2 requests/responses, for example via header x-http2-stream-id.

This is a single line change. I think the contribution is so trivial that I did not sign the CLA yet.


normanmaurer:
Can you add a unit test as well ?

jprante:
Yes, but it will take a little while.

normanmaurer:
@jprante I guess `frame.stream()` returns null which may be because of how the tests are setup. Please fix the tests and ping me once done.

normanmaurer:
@jprante please check https://github.com/netty/netty/pull/7786.

normanmaurer:
And thanks for the fix ...

PR #7780, opened 2018-03-14
PendingWriteQueue to handle write operations with void future

kachayev:
Motivation:

Right now `PendingWriteQueue.removeAndWriteAll` collects all promises to
PromiseCombiner instance which sets listener to each given promise throwing
`IllegalStateException` on `VoidChannelPromise` which breaks while loop
and "reports" operation has failed (when in fact part of writes might be
actually written).

Modifications:

Check if the promise is not void before adding it to the `PromiseCombiner`
instance. The test case is added.

Result:

`PendingWriteQueue.removeAndWriteAll` successfully writes all pending
writes even in case void promise was used.


normanmaurer:
@kachayev can you sign our ICLA:

https://netty.io/s/icla

Please let me know once done .

kachayev:
@normanmaurer Done!

Issue #7781, opened 2018-03-15
ResourceLeakDetector does not retain the initial record


trustin:
### Expected behavior

The leak report printed by `ResourceLeakDetector` retains the stack trace of the moment the buffer was created.

### Actual behavior

The record with `Created at:` heading is not actually the creation record, e.g.

```
Created at:
    io.netty.buffer.AdvancedLeakAwareByteBuf.writeBytes(AdvancedLeakAwareByteBuf.java:616)
    ...
```

### Steps to reproduce

Not sure, but probably write an app that generates a lot of access records and leak?

### Minimal yet complete reproducer code (or URL to code)

Sorry, no.

### Netty version

4.1.22

### JVM version (e.g. `java -version`)

```
java version "1.8.0_144"
Java(TM) SE Runtime Environment (build 1.8.0_144-b01)
Java HotSpot(TM) 64-Bit Server VM (build 25.144-b01, mixed mode)
```

### OS version (e.g. `uname -a`)

```
Linux infinity 4.14.25-1-lts #1 SMP Fri Mar 9 22:02:42 CET 2018 x86_64 GNU/Linux
```

normanmaurer:
Will have a look

normanmaurer:
@trustin please check the PR.

PR #7782, opened 2018-03-15
Correctly record creation stacktrace in ResourceLeakDetector.

normanmaurer:
Motivation:

We missed to correctly record the stacktrace of the creation of an ResourceLeak record. This could either have the effect to log the wrote stacktrace for creation or not log a stacktrace at all if the object was dropped on the floor after it was created.

Modifications:

Correctly create a Record on creation of the object.

Result:

Fixes https://github.com/netty/netty/issues/7781.


normanmaurer:
@trustin this should fix your problem reported in #7781

normanmaurer:
@carl-mastrangelo this was introduced by one of your refactorings in `ResourceLeakDetector`.

normanmaurer:
(no blame....)

PR #7784, opened 2018-03-15
HttpProxyHandler generates invalid CONNECT url and Host header when address is resolved

slandelle:
Motivation:

HttpProxyHandler uses `NetUtil#toSocketAddressString` to compute
CONNECT url and Host header.

The url is correct when the address is unresolved, as
`NetUtil#toSocketAddressString` will then use
`getHoststring`/`getHostname`. If the address is already resolved, the
url will be based on the IP instead of the hostname.

There’s an additional minor issue with the Host header: default port
443 should be omitted.

Modifications:

* Only use NetUtil.toSocketAddressString when address is unresolved,
use `getHostname` otherwise
* Don’t append port to Host header when it’s 443

Result:

HttpProxyHandler performs properly when connecting to a resolved address

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
@@ -17,7 +17,9 @@ package io.netty.handler.codec.http;
 
 import io.netty.util.AsciiString;
 import io.netty.util.CharsetUtil;
+import io.netty.util.NetUtil;
 
+import java.net.InetSocketAddress;
 import java.net.URI;
 import java.util.ArrayList;
 import java.nio.charset.Charset;
@@ -510,5 +512,22 @@ public final class HttpUtil {
         } else {
             return contentTypeValue.length() > 0 ? contentTypeValue : null;
         }
+    }
+
+    /**
+     * Formats the host string of an address so it can be used for computing an HTTP component
+     * such as an URL or a Host header
+     * @param addr the address
+     * @return the formatted String
+     */
+    public static String formatHostnameForHttp(InetSocketAddress addr) {
+        String hostString = NetUtil.getHostname(addr);
+        if (NetUtil.isValidIpV6Address(hostString)) {
+            if (!addr.isUnresolved()) {
+                hostString = NetUtil.toAddressString(addr.getAddress());
+            }
+            return "[" + hostString + "]";
+        }
+        return hostString;
     }
 }
--- a/common/src/main/java/io/netty/util/NetUtil.java
+++ b/common/src/main/java/io/netty/util/NetUtil.java
@@ -897,8 +897,8 @@ public final class NetUtil {
         final StringBuilder sb;
 
         if (addr.isUnresolved()) {
-            String hostString = PlatformDependent.javaVersion() >= 7 ? addr.getHostString() : addr.getHostName();
-            sb = newSocketAddressStringBuilder(hostString, port, !isValidIpV6Address(hostString));
+            String hostname = getHostname(addr);
+            sb = newSocketAddressStringBuilder(hostname, port, !isValidIpV6Address(hostname));
         } else {
             InetAddress address = addr.getAddress();
             String hostString = toAddressString(address);
@@ -1066,6 +1066,16 @@ public final class NetUtil {
         }
 
         return b.toString();
+    }
+
+    /**
+     * Returns {@link InetSocketAddress#getHostString()} if Java >= 7,
+     * or {@link InetSocketAddress#getHostName()} otherwise.
+     * @param addr The address
+     * @return the host string
+     */
+    public static String getHostname(InetSocketAddress addr) {
+        return PlatformDependent.javaVersion() >= 7 ? addr.getHostString() : addr.getHostName();
     }
 
     /**
--- a/handler-proxy/src/main/java/io/netty/handler/proxy/HttpProxyHandler.java
+++ b/handler-proxy/src/main/java/io/netty/handler/proxy/HttpProxyHandler.java
@@ -29,11 +29,11 @@ import io.netty.handler.codec.http.HttpHeaders;
 import io.netty.handler.codec.http.HttpMethod;
 import io.netty.handler.codec.http.HttpResponse;
 import io.netty.handler.codec.http.HttpResponseStatus;
+import io.netty.handler.codec.http.HttpUtil;
 import io.netty.handler.codec.http.HttpVersion;
 import io.netty.handler.codec.http.LastHttpContent;
 import io.netty.util.AsciiString;
 import io.netty.util.CharsetUtil;
-import io.netty.util.NetUtil;
 
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
@@ -47,6 +47,7 @@ public final class HttpProxyHandler extends ProxyHandler {
     private final String username;
     private final String password;
     private final CharSequence authorization;
+    private final boolean ignoreDefaultPortsInConnectHostHeader;
     private HttpResponseStatus status;
     private HttpHeaders headers;
 
@@ -55,11 +56,18 @@ public final class HttpProxyHandler extends ProxyHandler {
     }
 
     public HttpProxyHandler(SocketAddress proxyAddress, HttpHeaders headers) {
+        this(proxyAddress, headers, false);
+    }
+
+    public HttpProxyHandler(SocketAddress proxyAddress,
+                            HttpHeaders headers,
+                            boolean ignoreDefaultPortsInConnectHostHeader) {
         super(proxyAddress);
         username = null;
         password = null;
         authorization = null;
         this.headers = headers;
+        this.ignoreDefaultPortsInConnectHostHeader = ignoreDefaultPortsInConnectHostHeader;
     }
 
     public HttpProxyHandler(SocketAddress proxyAddress, String username, String password) {
@@ -68,6 +76,14 @@ public final class HttpProxyHandler extends ProxyHandler {
 
     public HttpProxyHandler(SocketAddress proxyAddress, String username, String password,
                             HttpHeaders headers) {
+        this(proxyAddress, username, password, headers, false);
+    }
+
+    public HttpProxyHandler(SocketAddress proxyAddress,
+                            String username,
+                            String password,
+                            HttpHeaders headers,
+                            boolean ignoreDefaultPortsInConnectHostHeader) {
         super(proxyAddress);
         if (username == null) {
             throw new NullPointerException("username");
@@ -87,6 +103,7 @@ public final class HttpProxyHandler extends ProxyHandler {
         authzBase64.release();
 
         this.headers = headers;
+        this.ignoreDefaultPortsInConnectHostHeader = ignoreDefaultPortsInConnectHostHeader;
     }
 
     @Override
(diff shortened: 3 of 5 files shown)


normanmaurer:
@slandelle this produced multiple test errors... please check

slandelle:
Yes, I'll have a look.

But I think those tests are broken.
For example, [setting the SslHandler before the HttpProxyHandler](https://github.com/netty/netty/blob/4.1/handler-proxy/src/test/java/io/netty/handler/proxy/ProxyHandlerTest.java#L178-L179) in the pipeline causes the CONNECT request to be sent over HTTPS instead of clear HTTP.
Those tests only pass because the proxy is a fake one and is broken too.

normanmaurer:
@slandelle then fix the broken tests ;)

handler-proxy/src/main/java/io/netty/handler/proxy/HttpProxyHandler.java
```diff
@@ -127,13 +128,36 @@ protected void removeDecoder(ChannelHandlerContext ctx) throws Exception {
     @Override
     protected Object newInitialMessage(ChannelHandlerContext ctx) throws Exception {
         InetSocketAddress raddr = destinationAddress();
-        final String host = NetUtil.toSocketAddressString(raddr);
+
+        String hostString = PlatformDependent.javaVersion() >= 7 ? raddr.getHostString() : raddr.getHostName();
```
normanmaurer: @slandelle should we make this a utility method in the http package ? Like in HttpUtils ? I think this is something that is useful in general.

slandelle: Can do, sure. But does it make sense to introduce this now while work on Netty 5 targeting Java 8 is about to start?

normanmaurer: @slandelle I think it still does as netty 5 is most likely still be 1 year away

slandelle:
@normanmaurer @trustin Note that tests passes, but I didn't fix the logic in ProxyHandlerTest that looks broken to me. That would be a separate issue.

handler-proxy/src/main/java/io/netty/handler/proxy/HttpProxyHandler.java
slandelle: ok

normanmaurer:
@slandelle sounds good... if you have time we love PRs as you know ;)

handler-proxy/src/main/java/io/netty/handler/proxy/HttpProxyHandler.java
slandelle: This utility would be better suited for NetUtil than HttpUtil, don't you think?

normanmaurer: Not sure... isnt it kind of related how you format stuff for the http header ?

slandelle: Ah, ok, I thought you were only talking about the first line above your comment, not the whole block. Will do.

kachayev:
@slandelle @normanmaurer I have a question regarding

> default port 443 should be omitted

[RFC 7230](https://tools.ietf.org/html/rfc7230#section-5.4) and older [RFC2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23) on `Host` header and [RFC7231](https://tools.ietf.org/html/rfc7231#section-4.3.6) on `CONNECT` method do not require excluding 443 port from the header.

`RFC2616` says

> A "host" without any trailing port information implies the default port for the service requested

Which might be applied to 80 and 443 if we still honor 2616, but not exclusively to 443.

slandelle:
@kachayev From my experience with AHC), there are some servers that really expect with port. And that's how browsers compute Host headers too.

common/src/main/java/io/netty/util/NetUtil.java
```diff
@@ -912,8 +912,8 @@ public static String toSocketAddressString(InetSocketAddress addr) {
      */
     public static String toSocketAddressString(String host, int port) {
         String portStr = String.valueOf(port);
-        return newSocketAddressStringBuilder(
-                host, portStr, !isValidIpV6Address(host)).append(':').append(portStr).toString();
+        return newSocketAddressStringBuilder(host, portStr, !isValidIpV6Address(host))
+                .append(':').append(port).toString();
```
fenik17: Why did you make this change?

slandelle: reverted

kachayev:
@slandelle With the latest changes `getHostHeader` implementation makes perfect sense and correlates with RFCs (excluding both default ports). But [here](https://github.com/netty/netty/pull/7784/files#diff-95cd1e58ffefe4ca490c3855b1a4446cR131) you always call the method with `secure` set to `true`, when in fact I can use http-only proxy. And here `url` is calculated by checking only 443 (my point that it's better either honor both default ports or do not honor none of them).

slandelle:
@kachayev You wouldn't use this handler for http-only proxy, the mechanism is different and doesn't use a CONNECT request (directly connect to proxy and send absolute url instead of relative). I have to check for WebSockets then.

kachayev:
@slandelle "wouldn't use" doesn't mean it's not allowed. RFC7231 I mentioned above does not specify tunnel to be used **only** with a secure connection. CURL, for example, does not establish tunnel when connecting to HTTP proxy **by default** but you still can force this:

```shell
$ curl -vvv -o /dev/null -x http://localhost:8888 --proxytunnel http://www.google.com
* Rebuilt URL to: http://www.google.com/
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8888 (#0)
* Establish HTTP proxy tunnel to www.google.com:80
> CONNECT www.google.com:80 HTTP/1.1
> Host: www.google.com:80
> User-Agent: curl/7.54.0
> Proxy-Connection: Keep-Alive
>
< HTTP/1.1 200 Connection established
<
* Proxy replied OK to CONNECT request
> GET / HTTP/1.1
> Host: www.google.com
> User-Agent: curl/7.54.0
> Accept: */*
>
< HTTP/1.1 200 OK
```

slandelle:
@kachayev Which proxy do you use? Then, what do you suggest? Ignore 80 and 443? It's not possible to guess the actual scheme and current API doesn't let you specify.

kachayev:
@slandelle Charles for Mac, but that's irrelevant because a lot of proxies **do** support CONNECT over plain HTTP. My suggestion is not to ignore even default ports as it might break functionality in some cases (at least we cannot guarantee that it would not).

slandelle:
> My suggestion is not to ignore even default ports (neither 443 nor 80) as it might break functionality in some cases (at least we cannot guarantee that it would not).

From my experience, it was the other way around: things broke when the default port was there. And that's what AHC and browser does. I'm fine with not ignoring by default, but I'll introduce an option  to override.

slandelle:
@normanmaurer PTAL again

kachayev:
@slandelle 👍

handler-proxy/src/main/java/io/netty/handler/proxy/HttpProxyHandler.java
```diff
@@ -108,32 +125,39 @@ public String password() {
     }
 
     @Override
-    protected void addCodec(ChannelHandlerContext ctx) throws Exception {
+    protected void addCodec(ChannelHandlerContext ctx) {
```
normanmaurer: @slandelle you need to revert all the removals of `throws...` as its a breaking change. People who may have override the method and have `throws` on it will be in trouble.

slandelle:
@normanmaurer fixed

carl-mastrangelo:
cc: @zpencer

normanmaurer:
@carl-mastrangelo @ejona86 @trustin  any of you can verify this one ?

zpencer:
I can take a look

ejona86:
FWIW, the getHostString changes all look good. It's the `ignoreDefaultPortsInConnectHostHeader` that I needed to look at more (including the discussion here) before feeling comfortable approving.

handler-proxy/src/main/java/io/netty/handler/proxy/HttpProxyHandler.java
```diff
@@ -127,13 +144,20 @@ protected void removeDecoder(ChannelHandlerContext ctx) throws Exception {
     @Override
     protected Object newInitialMessage(ChannelHandlerContext ctx) throws Exception {
         InetSocketAddress raddr = destinationAddress();
-        final String host = NetUtil.toSocketAddressString(raddr);
+
+        String hostString = HttpUtil.formatForHttp(raddr);
+        int port = raddr.getPort();
+        String url = hostString + ":" + port;
+        String hostHeader = (ignoreDefaultPortsInConnectHostHeader && (port == 80 || port == 443)) ?
```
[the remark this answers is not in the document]
normanmaurer: @slandelle can you do this ?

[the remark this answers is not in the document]
slandelle: I'm not convinced it would make sense to test this single line of code.
Then, url and Host header computations are entangled so I can save allocations.
I can't test `HttpProxyHandler#newInitialMessage` because I would need to override `destinationAddress()` which is final.
I'm going to extract CONNECT request creation in a static method.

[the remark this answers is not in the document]
slandelle: Actually, it should be fine. Stay tuned.

slandelle:
@normanmaurer @zpencer Host header tests added, PTAL again.

normanmaurer:
@trustin any concerns ?

trustin:
Sorry! Will take a look tomorrow.

2018년 3월 22일 (목) 오후 10:40, Norman Maurer <notifications@github.com>님이 작성:

codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
```diff
@@ -511,4 +513,21 @@ public static CharSequence getMimeType(CharSequence contentTypeValue) {
             return contentTypeValue.length() > 0 ? contentTypeValue : null;
         }
     }
+
+    /**
+     * Formats an address to it can be used for computing an HTTP component such as an URL or a Host header
```
normanmaurer: s/to/so/ ?

codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
```diff
@@ -511,4 +513,21 @@ public static CharSequence getMimeType(CharSequence contentTypeValue) {
...
+     * @return the formatted String
+     */
+    public static String formatForHttp(InetSocketAddress addr) {
+        String hostString = NetUtil.getHostString(addr);
+        if (NetUtil.isValidIpV6Address(hostString)) {
+            if (!addr.isUnresolved()) {
+                hostString = NetUtil.toAddressString(addr.getAddress());
+            }
+            return "[" + hostString + "]";
+        } else {
```
normanmaurer: nit: you can remove the else as you return anyway.

common/src/main/java/io/netty/util/NetUtil.java
```diff
@@ -1068,6 +1068,16 @@ private static String toAddressString(byte[] bytes, int offset, boolean ipv4Mapp
         return b.toString();
     }
 
+    /**
+     * Returns {@link InetSocketAddress#getHostString()} if Java >= 7,
+     * or {@link InetSocketAddress#getHostName()} otherwise
```
normanmaurer: nit: add `.` on EOL.

slandelle:
@normanmaurer done

codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
```diff
@@ -511,4 +513,20 @@ public static CharSequence getMimeType(CharSequence contentTypeValue) {
             return contentTypeValue.length() > 0 ? contentTypeValue : null;
         }
     }
+
+    /**
+     * Formats an address so it can be used for computing an HTTP component such as an URL or a Host header
+     * @param addr the address
+     * @return the formatted String
+     */
+    public static String formatForHttp(InetSocketAddress addr) {
```
trustin: I guess a user of this method may think the return value will contain the port number. How about adding a boolean parameter `withPort`, so that this method adds `:<port>` when it's `true`?

slandelle: What about trying to find a better name instead? Maybe `formatHostnameForHttp`?
I'm not fond of expanding an API beyond what's being used internally so it can serve for whatever other purpose we're not aware of.
WDYT?

trustin: That's also fine.

slandelle: Done

normanmaurer:
@slandelle thanks!

slandelle:
@normanmaurer Thanks for merging

Issue #7749, opened 2018-02-26
Proposal: DnsAddressResolverGroup to use injectable DnsNameResolverBuilder


kachayev:
### Expected behavior

It should be possible to create `DnsAddressResolverGroup` having `DnsNameResolverBuilder` ready-to-go.

### Actual behavior

Right now, if I want setup custom settings for `DnsAddressResolver` I need to sub-class `DnsAddressResolverGroup` and override `newNameResolver` [method](https://github.com/netty/netty/blob/06dcca1dbcc048b6ce7d954e4ecdf1d3ed6af034/resolver-dns/src/main/java/io/netty/resolver/dns/DnsAddressResolverGroup.java#L97) as it creates new `DnsNameResolverBuilder` with "almost" default settings internally.

Ideally, I want to instantiate `DnsNameResolverBuilder` by my own and pass it to the `DnsAddressResolverGroup` as an argument. More than that, I don't need `channelFactory` and `nameServerProvider` being carried over `DnsAddressResolverGroup`: I can set them directly to the `DnsNameResolverBuilder` instance.

It probably would be even easier to create a new `AddressResolverGroup` implementation to work this way, but `InflightNameResolver` is not allowed to be used from the outside of the package, so if the user wants to benefit from it, it should follow `DnsAddressResolverGroup` API and override `newNameResolver`.

### Netty version

4.1.*

### Proposal

There're 2 options here:

1. Upgrade `DnsAddressResolverGroup` with a new constructor 

```java
public DnsAddressResolverGroup(DnsNameResolverBuilder builder);
```

and appropriate checker if we have `DnsNameResolverBuilder` or we need to create a new one.

2. Create a new `AddressResolverGroup` implementation, like `PluggableDnsAddressResolverGroup` (? naming is hard).

What do you think? Happy to create a PR if there are no objections here.

normanmaurer:
Fixed by https://github.com/netty/netty/pull/7793

Issue #5838, opened 2016-09-19
DnsNameResolverBuilder does not instantiate InflightNameResolver


luengnat:
wrt https://github.com/netty/netty/pull/5267, DnsNameResolverBuilder does not instantiate InflightNameResolver. The only place that InflightNameResolver is used is https://github.com/netty/netty/blob/4.1/resolver-dns/src/main/java/io/netty/resolver/dns/DnsAddressResolverGroup.java#L73. The method is already marked as deprecated and the replaced method (https://github.com/netty/netty/blob/4.1/resolver-dns/src/main/java/io/netty/resolver/dns/DnsAddressResolverGroup.java#L100) uses DnsNameResolverBuilder which instantiates DnsNameResolver instead of InflightNameResolver.

With the above implementation, this causes the DNS servers to be overloaded with requests when there are multiple DNS requests made concurrently. Is it possible to switch to InflightNameResolver everywhere by default?

Scottmitch:
@trustin - wdty?

luengnat:
@trustin nudge :)

trustin:
`DnsNameResolverBuilder`, by its name, is supposed to build a `DnsNameResolver`, not an `InflightNameResolver` that wraps a `DnsNameResolver`.

Perhaps we could:

- make `DnsNameResolver` an abstract class,
- move the implementation of `DnsNameResolver` to some other class such as `DnsNameResolverImpl`
- create a `DefaultDnsNameResolver` that 1) extends `DnsNameResolver` and 2) is composed of `InflightNameResolver` and `DnsNameResolverImpl`

luengnat:
* I'm not sure about the benefit of using DnsNameResolver by default. The default behavior does not look appealing given that it already has a DNS caching implemented but it does not handle the in-flight requests well.
* Wouldn't it be simpler to wrap around DnsNameResolver with InflightNameResolver  by default in DnsAddressResolverGroup.java?

luengnat:
@Scottmitch / @normanmaurer what do you think about the proposal?

Scottmitch:
@trustin's suggestion keeps separation of concerns and we could still implement it such that the `inflight` behavior is clarified and users can get a good default. I agree that most folks will likely want to use the `inflight` by default (@trustin WDYT).

trustin:
> Wouldn't it be simpler to wrap around DnsNameResolver with InflightNameResolver by default in DnsAddressResolverGroup.java?

[...]

PR #7793, opened 2018-03-19
DnsAddressResolverGroup to use pluggable DnsNameResolverBuilder

kachayev:
Motivation:

Right now to customize DNS name resolver when using DnsAddressResolverGroup
one should subclass implementation and override newNameResolver method when
in fact it's possible to collect all settings in a DnsNameResolverBuilder
instance. Described in #7749.

Modifications:

- Added new constructor for DnsNameResolverBuilder in order to delay
  EventLoop specification

- Added new single-argument constructor for DnsAddressResolverGroup and
  RoundRobinDnsAddressResolverGroup accepting DnsNameResolverBuilder
  instance

- DnsAddressResolverGroup to build a new resolver using DnsNameResolverBuilder
  given instead of creating a new one

Result:

Much easier to customize DNS settings w/o subclassing DnsAddressResolverGroup

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsAddressResolverGroup.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsAddressResolverGroup.java
@@ -42,23 +42,27 @@ import static io.netty.util.internal.PlatformDependent.newConcurrentHashMap;
 @UnstableApi
 public class DnsAddressResolverGroup extends AddressResolverGroup<InetSocketAddress> {
 
-    private final ChannelFactory<? extends DatagramChannel> channelFactory;
-    private final DnsServerAddressStreamProvider nameServerProvider;
+    private final DnsNameResolverBuilder dnsResolverBuilder;
 
     private final ConcurrentMap<String, Promise<InetAddress>> resolvesInProgress = newConcurrentHashMap();
     private final ConcurrentMap<String, Promise<List<InetAddress>>> resolveAllsInProgress = newConcurrentHashMap();
 
+    public DnsAddressResolverGroup(DnsNameResolverBuilder dnsResolverBuilder) {
+        this.dnsResolverBuilder = dnsResolverBuilder.copy();
+    }
+
     public DnsAddressResolverGroup(
             Class<? extends DatagramChannel> channelType,
             DnsServerAddressStreamProvider nameServerProvider) {
-        this(new ReflectiveChannelFactory<DatagramChannel>(channelType), nameServerProvider);
+        this(new DnsNameResolverBuilder());
+        dnsResolverBuilder.channelType(channelType).nameServerProvider(nameServerProvider);
     }
 
     public DnsAddressResolverGroup(
             ChannelFactory<? extends DatagramChannel> channelFactory,
             DnsServerAddressStreamProvider nameServerProvider) {
-        this.channelFactory = channelFactory;
-        this.nameServerProvider = nameServerProvider;
+        this(new DnsNameResolverBuilder());
+        dnsResolverBuilder.channelFactory(channelFactory).nameServerProvider(nameServerProvider);
     }
 
     @SuppressWarnings("deprecation")
@@ -70,7 +74,11 @@ public class DnsAddressResolverGroup extends AddressResolverGroup<InetSocketAddr
                     " (expected: " + StringUtil.simpleClassName(EventLoop.class));
         }
 
-        return newResolver((EventLoop) executor, channelFactory, nameServerProvider);
+        // we don't really need to pass channelFactory and nameServerProvider separately,
+        // but still keep this to ensure backward compatibility with (potentially) override methods
+        return newResolver((EventLoop) executor,
+                dnsResolverBuilder.channelFactory(),
+                dnsResolverBuilder.nameServerProvider());
     }
 
     /**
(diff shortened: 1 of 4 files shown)


normanmaurer:
@trustin any concerns ?

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverBuilder.java
```diff
@@ -347,6 +372,10 @@ public DnsNameResolverBuilder decodeIdn(boolean decodeIdn) {
      * @return a {@link DnsNameResolver}
      */
     public DnsNameResolver build() {
+        if (eventLoop == null) {
+            throw new IllegalStateException("eventLoop should be specified to build DnsNameResolver");
```
trustin: - build -> build a
- DnsNameResolver -> DnsNameResolver. (dot)

trustin:
@normanmaurer Just left a nit. Sounds OK to me.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverBuilder.java
kachayev: @trustin 🤔shoud other 2 `IllegalStateException` be updated with a trailing dot?

resolver-dns/src/main/java/io/netty/resolver/dns/DnsAddressResolverGroup.java
```diff
@@ -42,23 +42,27 @@
...
 
-    private final ChannelFactory<? extends DatagramChannel> channelFactory;
-    private final DnsServerAddressStreamProvider nameServerProvider;
+    private final DnsNameResolverBuilder dnsResolverBuilder;
 
     private final ConcurrentMap<String, Promise<InetAddress>> resolvesInProgress = newConcurrentHashMap();
     private final ConcurrentMap<String, Promise<List<InetAddress>>> resolveAllsInProgress = newConcurrentHashMap();
 
+    public DnsAddressResolverGroup(DnsNameResolverBuilder dnsResolverBuilder) {
+        this.dnsResolverBuilder = dnsResolverBuilder;
```
carl-mastrangelo: This won't work.   If someone makes a DnsNameResolverBuilder and then calls build, and then subsequently modifies the builder, the channelFactory of DnsAddressResolverGroup will change. 

That would be unexpected.

kachayev: @carl-mastrangelo Not sure I follow. If you passed `DnsNameResolverBuilder` to create a `DnsAddressResolverGroup`, what you would expect from changing `channelFactory` on the builder you've used?

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverBuilder.java
kachayev: @trustin Fixed the message.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsAddressResolverGroup.java
carl-mastrangelo: I wouldn't expect the DnsAddressResolverGroup to use a different channel factory if the DnsNameResolverBuilder changed channelFactory.    That is currently possible as is.

kachayev: @carl-mastrangelo Would a deep copy of the builder meet your expectations? This will prevent mutations propagation, but it might be not expected as well.

carl-mastrangelo: a deep copy would be good.   Immutability is highly recommended.  This is Effective Java item 15 in the the 2nd edition.

kachayev: @carl-mastrangelo Updated.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolverBuilder.java
trustin: A rule of thumb is:

- Do not end with a dot if it's not a sentence, e.g. 'not a number'
- End with a dot if it's a sentence, e.g. 'It's not a number.
- End with a dot if it's not a sentence but followed by a sentence, e.g. 'not a number. It has to be a number.;

kachayev: @trustin Got it, thanks!

normanmaurer:
@carl-mastrangelo can you have a final look ?

kachayev:
@carl-mastrangelo The intent of this PR is described above: to make `DnsAddressResolverGroup` configurable (right now it doable only by subclassing the group). There are no particular fields you're interested in here, as you actually need the entire builder. The only difference that this PR makes is that you can pass builder with all configuration settings you need instead of relying on the one created internally in the resolver group.

normanmaurer:
@kachayev thanks a lot... just merged

luengnat:
Do you think this is a good workaround for #5838?

PR #7796, opened 2018-03-19
Skip NPN tests when libressl 2.6.1+ is used.

normanmaurer:
Motivation:

    LibreSSL removed support for NPN in its 2.6.1+ releases.

    Modifications:

    Skip NPN tests in libressl 2.6.1+

    Result:

    Be able to run netty tests against libressl 2.6.1+ as well.

### DIFF
--- a/handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
@@ -131,8 +131,37 @@ public class OpenSslEngineTest extends SSLEngineTest {
         super.testClientHostnameValidationFail();
     }
 
+    private static boolean isNpnSupported(String versionString) {
+        String[] versionStringParts = versionString.split(" ", -1);
+        if (versionStringParts.length == 2 && "LibreSSL".equals(versionStringParts[0])) {
+            String[] versionParts = versionStringParts[1].split("\\.", -1);
+            if (versionParts.length == 3) {
+                int major = Integer.parseInt(versionParts[0]);
+                if (major < 2) {
+                    return true;
+                }
+                if (major > 2) {
+                    return false;
+                }
+                int minor = Integer.parseInt(versionParts[1]);
+                if (minor < 6) {
+                    return true;
+                }
+                if (minor > 6) {
+                    return false;
+                }
+                int bugfix = Integer.parseInt(versionParts[2]);
+                if (bugfix > 0) {
+                    return false;
+                }
+            }
+        }
+        return true;
+    }
     @Test
     public void testNpn() throws Exception {
+        String versionString = OpenSsl.versionString();
+        assumeTrue("LibreSSL 2.6.1 removed NPN support, detected " + versionString, isNpnSupported(versionString));
         ApplicationProtocolConfig apn = acceptingNegotiator(Protocol.NPN,
                 PREFERRED_APPLICATION_LEVEL_PROTOCOL);
         setupHandlers(apn);
--- a/handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
@@ -1009,7 +1009,7 @@ public abstract class SSLEngineTest {
         SslHandler handler = channel.pipeline().get(SslHandler.class);
         assertNotNull(handler);
         String appProto = handler.applicationProtocol();
-        assertEquals(appProto, expectedApplicationProtocol);
+        assertEquals(expectedApplicationProtocol, appProto);
 
         SSLEngine engine = handler.engine();
         if (engine instanceof Java9SslEngine) {


handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
```diff
@@ -1009,7 +1009,7 @@ private static void verifyApplicationLevelProtocol(Channel channel, String expec
         SslHandler handler = channel.pipeline().get(SslHandler.class);
         assertNotNull(handler);
         String appProto = handler.applicationProtocol();
-        assertEquals(appProto, expectedApplicationProtocol);
+        assertEquals(expectedApplicationProtocol, appProto);
```
normanmaurer: I did this change as it was using the wrong order here so the error message was confusing

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
```diff
@@ -131,8 +132,37 @@ public void testClientHostnameValidationFail() throws InterruptedException, SSLE
         super.testClientHostnameValidationFail();
     }
 
+    private static boolean isNpnSupported(String versionString) {
+        String[] versionStringParts = versionString.split(" ");
```
ejona86: I suggest never using one-arg version of split, because it's almost always a mistake. If you want the current behavior, explicitly pass `0` as the limit, but this is rare. But generally you want a negative value (for no limit) or a limit. It looks like here could be a limit of 2 or -1.

Ditto below, with a limit of 3 or -1.

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
```diff
@@ -131,8 +132,37 @@ public void testClientHostnameValidationFail() throws InterruptedException, SSLE
...
+                    return false;
+                }
+            }
+        }
+        return true;
+    }
     @Test
     public void testNpn() throws Exception {
+        String versionString = OpenSsl.versionString();
+        assumeFalse("LibreSSL 2.6.1 removed NPN support, detected " + versionString, isNpnSupported(versionString));
```
ejona86: Shouldn't this be an `assumeTrue`?

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
```diff
@@ -131,8 +132,37 @@ public void testClientHostnameValidationFail() throws InterruptedException, SSLE
...
+                }
+                int minor = Integer.parseInt(versionParts[1]);
+                if (minor < 6) {
+                    return true;
+                }
+                if (minor > 6) {
+                    return false;
+                }
+                int bugfix = Integer.parseInt(versionParts[2]);
+                if (bugfix > 0) {
```
ejona86: This looks wrong. Based on this, 2.6.0 would support NPN and 2.6.1 would not. This is probably why the `assumeFalse` is "working."

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
```diff
@@ -56,6 +56,7 @@
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assume.assumeFalse;
 import static org.junit.Assume.assumeTrue;
```
ejona86: In unrelated news, it looks like assumeTrue is duplicated...

ejona86: nm. One is assert, the other is assume.

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
normanmaurer: @ejona86 actually they removed the supported in 2.6.1 so I think this is correct.

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
normanmaurer: yep... brain fart.

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
normanmaurer: I changed it to -1

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
ejona86: O.o

Very well. :smile:

normanmaurer:
@ejona86 done

normanmaurer:
Verified it works with https://github.com/netty/netty-tcnative/pull/334

PR #7798, opened 2018-03-20
Add tests for EmptyHeaders

normanmaurer:
Motivation:

6e5fd9311fc0abab37fd442eebdc810aa0c3d6a1 fixed a bug in EmptyHeaders which was never noticed before because we had no tests.

Modifications:

Add tests for EmptyHeaders.

Result:

EmptyHeaders is tested now.


normanmaurer:
Merging as these are only tests.

PR #7799, opened 2018-03-20
Ensure we not schedule multiple timeouts for close notify

normanmaurer:
Motivation:

We should only schedule one timeout to wait for the close notify to be done.

Modifications:

Keep track of if we already scheduled a timeout for close notify and if so not schedule another one.

Result:

No duplicated timeouts.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -400,6 +400,7 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
     private boolean needsFlush;
 
     private boolean outboundClosed;
+    private boolean closeNotify;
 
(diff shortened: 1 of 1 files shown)


normanmaurer:
@doom369 FYI... related to https://stackoverflow.com/questions/49338730/how-to-deal-with-netty-last-write-attempt-timed-out-force-closing-the-connecti/49348029?noredirect=1#comment85739344_49348029

doom369:
@normanmaurer thanks! I'll be able to make new deploy in few days.

normanmaurer:
Yep on my to do list

normanmaurer:
@carl-mastrangelo I am a bit out of ideas how to add a unit test :(

normanmaurer:
@trustin can you check as well ?

normanmaurer:
Let me merge this now

PR #7801, opened 2018-03-21
NIO read spin event loop spin when half closed

Scottmitch:
Motivation:
AbstractNioByteChannel will detect that the remote end of the socket has
been closed and propagate a user event through the pipeline. However if
the user has auto read on, or calls read again, we may propagate the
same user events again. If the underlying transport continuously
notifies us that there is read activity this will happen in a spin loop
which consumes unnecessary CPU.

Modifications:
- AbstractNioByteChannel's unsafe read() should check if the input side
of the socket has been shutdown before processing the event. This is
consistent with EPOLL and KQUEUE transports.

Result:
No more read spin loop in NIO when the channel is half closed.

### DIFF
--- a/transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
+++ b/transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
@@ -28,7 +28,7 @@ import io.netty.channel.RecvByteBufAllocator;
 import io.netty.channel.internal.ChannelUtils;
 import io.netty.channel.socket.ChannelInputShutdownEvent;
 import io.netty.channel.socket.ChannelInputShutdownReadComplete;
-import io.netty.channel.socket.DefaultSocketChannelConfig;
+import io.netty.channel.socket.SocketChannelConfig;
 import io.netty.util.internal.StringUtil;
 
 import java.io.IOException;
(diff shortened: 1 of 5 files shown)


Scottmitch:
@RoganDawes - can you verify this PR fixes the issue you observed with NIO?

Scottmitch:
unit tests coming soon.

normanmaurer:
@RoganDawes ping

Scottmitch:
big thanks to @normanmaurer for the unit test ... sorry for the delay!

transport/src/main/java/io/netty/channel/nio/AbstractNioByteChannel.java
```diff
@@ -84,17 +85,27 @@ public ChannelMetadata metadata() {
         return METADATA;
     }
 
+    final boolean shouldBreakReadReady(ChannelConfig config) {
+        return isInputShutdown0() && (inputClosedSeenErrorOnRead || !isAllowHalfClosure(config));
+    }
+
+    private static boolean isAllowHalfClosure(ChannelConfig config) {
+        return config instanceof DefaultSocketChannelConfig &&
+                ((DefaultSocketChannelConfig) config).isAllowHalfClosure();
```
normanmaurer: Replace `DefaultSocketChannelConfig` with `SocketChannelConfig`. After this its ready to be merged :)

RoganDawes:
Sorry for the delay. I'm actually not quite sure how to test if this PR solves my problem. I pulled the latest changes, but this doesn't seem to be present, not sure how to get this branch to test :flushed:

normanmaurer:
@RoganDawes please just pull the latest 4.1 branch... it was just merged.

PR #7803, opened 2018-03-22
Add DnsNameResolver.resolveAll(DnsQuestion)

### LINKED ISSUE #7795: Expose DnsNameResolverContext (or something equivalent to it) to users
A user may want to reuse the logic related with search list enumeration, retries and error handling in his or her resolver implementations. For example, a user could extend or customize `DnsNameResolverContext` to implement DNS-based service discovery that retrieves [SRV records](https://en.wikipedia.org/wiki/SRV_record), while respecting the global resolver configuration (`/etc/resolv.conf`). Currently, the user would have to re-implement search list enumeration, retries and error handling by his or her own.

Let me take a stab of this if it sounds good.

trustin:
Motivation:

A user is currently expected to use DnsNameResolver.query() when he or
she wants to look up the full DNS records rather than just InetAddres.

However, query() only performs a single query. It does not handle
/etc/hosts file, redirection, CNAMEs or multiple name servers.

As a result, such a user has to duplicate all the logic in
DnsNameResolverContext.

Modifications:

- Refactor DnsNameResolverContext so that it can send queries for
  arbitrary record types.
  - Rename DnsNameResolverContext to DnsResolveContext
  - Add DnsAddressResolveContext which extends DnsResolveContext for
    A/AAAA lookup
  - Add DnsRecordResolveContext which extends DnsResolveContext for
    arbitrary lookup
- Add DnsNameResolverContext.resolveAll(DnsQuestion) and its variants
- Change DnsNameResolverContext.resolve() delegates the resolve request
  to resolveAll() for simplicity
- Move the code that decodes A/AAAA record content to DnsAddressDecoder

Result:

- Fixes #7795
- A user does not have to duplicate DnsNameResolverContext in his or her
  own code to implement the usual DNS resolver behavior.

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsAddressDecoder.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsAddressDecoder.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.resolver.dns;
+
+import java.net.IDN;
+import java.net.Inet4Address;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufHolder;
+import io.netty.handler.codec.dns.DnsRawRecord;
+import io.netty.handler.codec.dns.DnsRecord;
+
+/**
+ * Decodes an {@link InetAddress} from an A or AAAA {@link DnsRawRecord}.
+ */
+final class DnsAddressDecoder {
+
+    private static final int INADDRSZ4 = 4;
+    private static final int INADDRSZ6 = 16;
+
+    /**
+     * Decodes an {@link InetAddress} from an A or AAAA {@link DnsRawRecord}.
+     *
+     * @param record the {@link DnsRecord}, most likely a {@link DnsRawRecord}
+     * @param name the host name of the decoded address
+     * @param decodeIdn whether to convert {@code name} to a unicode host name
+     *
+     * @return the {@link InetAddress}, or {@code null} if {@code record} is not a {@link DnsRawRecord} or
+     *         its content is malformed
+     */
+    static InetAddress decodeAddress(DnsRecord record, String name, boolean decodeIdn) {
+        if (!(record instanceof DnsRawRecord)) {
+            return null;
+        }
+        final ByteBuf content = ((ByteBufHolder) record).content();
+        final int contentLen = content.readableBytes();
+        if (contentLen != INADDRSZ4 && contentLen != INADDRSZ6) {
+            return null;
+        }
+
+        final byte[] addrBytes = new byte[contentLen];
+        content.getBytes(content.readerIndex(), addrBytes);
+
+        try {
+            return InetAddress.getByAddress(decodeIdn ? IDN.toUnicode(name) : name, addrBytes);
+        } catch (UnknownHostException e) {
+            // Should never reach here.
+            throw new Error(e);
+        }
+    }
+
+    private DnsAddressDecoder() { }
+}
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsAddressDecoder.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsAddressDecoder.java
@@ -16,7 +16,6 @@
 package io.netty.resolver.dns;
 
 import java.net.IDN;
-import java.net.Inet4Address;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsAddressResolveContext.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsAddressResolveContext.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.resolver.dns;
+
+import static io.netty.resolver.dns.DnsAddressDecoder.decodeAddress;
+
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.List;
+
+import io.netty.channel.EventLoop;
+import io.netty.handler.codec.dns.DnsRecord;
+import io.netty.handler.codec.dns.DnsRecordType;
+
+final class DnsAddressResolveContext extends DnsResolveContext<InetAddress> {
+
+    private final DnsCache resolveCache;
+
+    DnsAddressResolveContext(DnsNameResolver parent, String hostname, DnsRecord[] additionals,
+                             DnsServerAddressStream nameServerAddrs, DnsCache resolveCache) {
+        super(parent, hostname, DnsRecord.CLASS_IN, parent.resolveRecordTypes(), additionals, nameServerAddrs);
+        this.resolveCache = resolveCache;
+    }
+
+    @Override
+    DnsResolveContext<InetAddress> newResolverContext(DnsNameResolver parent, String hostname,
+                                                      int dnsClass, DnsRecordType[] expectedTypes,
+                                                      DnsRecord[] additionals,
+                                                      DnsServerAddressStream nameServerAddrs) {
+        return new DnsAddressResolveContext(parent, hostname, additionals, nameServerAddrs, resolveCache);
+    }
+
+    @Override
+    InetAddress convertRecord(DnsRecord record, String hostname, DnsRecord[] additionals, EventLoop eventLoop) {
+        return decodeAddress(record, hostname, parent.isDecodeIdn());
+    }
+
+    @Override
+    boolean gotResult(List<InetAddress> finalResult) {
+        final int size = finalResult.size();
+        final Class<? extends InetAddress> inetAddressType = parent.preferredAddressType().addressType();
+        for (int i = 0; i < size; i++) {
+            InetAddress address = finalResult.get(i);
+            if (inetAddressType.isInstance(address)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    @Override
+    void cache(String hostname, DnsRecord[] additionals,
+               DnsRecord result, InetAddress convertedResult) {
+        resolveCache.cache(hostname, additionals, convertedResult, result.timeToLive(), parent.ch.eventLoop());
+    }
+
+    @Override
+    void cache(String hostname, DnsRecord[] additionals, UnknownHostException cause) {
+        resolveCache.cache(hostname, additionals, cause, parent.ch.eventLoop());
+    }
+}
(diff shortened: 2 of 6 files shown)


resolver-dns/src/main/java/io/netty/resolver/dns/DnsAddressDecoder.java
```diff
@@ -0,0 +1,55 @@
...
+import io.netty.buffer.ByteBufHolder;
+import io.netty.handler.codec.dns.DnsRawRecord;
+import io.netty.handler.codec.dns.DnsRecord;
+
+final class DnsAddressDecoder {
+
+    private static final int INADDRSZ4 = 4;
+    private static final int INADDRSZ6 = 16;
+
+    static InetAddress decodeAddress(DnsRecord r, String name, boolean decodeIdn) {
```
normanmaurer: javadocs please

resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
```diff
@@ -1142,6 +1223,7 @@ int dnsRedirectPort(InetAddress server) {
             assertEquals(dnsServer.localAddress(), writtenEvent1.dnsServerAddress);
             QueryRedirectedEvent redirectedEvent = (QueryRedirectedEvent) observer.events.poll();
 
+            System.err.println(redirectedEvent.nameServers);
```
normanmaurer: revert ?

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -527,6 +534,94 @@ protected void doResolve(String inetHost, Promise<InetAddress> promise) throws E
...
+     * @param additionals additional records ({@code OPT})
+     *
+     * @return the list of the {@link DnsRecord}s as the result of the resolution
+     */
+    public final Future<List<DnsRecord>> resolveAll(DnsQuestion question, Iterable<DnsRecord> additionals) {
+        return resolveAll(question, additionals, executor().<List<DnsRecord>>newPromise());
+    }
+
+    /**
+     * Resolves the {@link DnsRecord}s that is matched by the specified {@link DnsQuestion}. Unlike
```
normanmaurer: ...that are...

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -527,6 +534,94 @@ protected void doResolve(String inetHost, Promise<InetAddress> promise) throws E
...
+     * @param question the question
+     *
+     * @return the list of the {@link DnsRecord}s as the result of the resolution
+     */
+    public final Future<List<DnsRecord>> resolveAll(DnsQuestion question) {
+        return resolveAll(question, EMPTY_ADDITIONALS, executor().<List<DnsRecord>>newPromise());
+    }
+
+    /**
+     * Resolves the {@link DnsRecord}s that is matched by the specified {@link DnsQuestion}. Unlike
```
normanmaurer: ...that are...

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -527,6 +534,94 @@ protected void doResolve(String inetHost, Promise<InetAddress> promise) throws E
         doResolve(inetHost, EMPTY_ADDITIONALS, promise, resolveCache);
     }
 
+    /**
+     * Resolves the {@link DnsRecord}s that is matched by the specified {@link DnsQuestion}. Unlike
```
normanmaurer: ...that are...

normanmaurer:
@slandelle can you have a look as well ?

resolver-dns/src/main/java/io/netty/resolver/dns/DnsAddressDecoder.java
trustin: Added Javadocs for `DnsAddressDecoder`

resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
trustin: Oops! Reverted.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
trustin: Fixed all :-)

trustin:
What happened to SonarQube?

normanmaurer:
@trustin good question... need to check

trustin:
Any comments?

normanmaurer:
@slandelle ping

normanmaurer:
@Scottmitch any concerns ?

resolver-dns/src/main/java/io/netty/resolver/dns/DnsResolveContext.java
```diff
@@ -411,21 +444,21 @@ private boolean handleRedirect(
                 for (int i = 0; i < additionalCount; i++) {
                     final DnsRecord r = res.recordAt(DnsSection.ADDITIONAL, i);
 
-                    if (r.type() == DnsRecordType.A && !parent.supportsARecords() ||
+                    if (r.type() != question.type() ||
```
Scottmitch: dropping the redirect may have undesired results if other behavior in this class changes, or DNS server is not as expected so consider adding a comment here. IIUC we are assuming the following behavior:

- this DNS resolver will issue a query for each `A` and `AAAA` records if they are both supported.
- it is assumed the DNS server will only provide redirects of the same type as the query, and therefore it is safe to drop redirects where types don't match. In addition if the DNS server responds with mismatching (or maybe all types) types it is desirable to filter the redirects to avoid multiple queries for the same redirect.

resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
```diff
@@ -945,6 +948,84 @@ private static void testResolveAll0(ResolvedAddressTypes addressTypes, InetAddre
...
+                    buf.append(r.type().name());
+                    buf.append(' ');
+                    buf.append(recordContent.readUnsignedShort());
+                    buf.append(' ');
+                    buf.append(DnsResolveContext.decodeDomainName(recordContent));
+
+                    ReferenceCountUtil.release(r);
+                }
+
+                logger.info("{} has the following MX records:{}", hostname, buf);
```
Scottmitch: is it possible to instead use the test DNS server that runs locally and verify results are as expected?

resolver-dns/src/main/java/io/netty/resolver/dns/DnsAddressDecoder.java
```diff
@@ -0,0 +1,68 @@
...
+import io.netty.handler.codec.dns.DnsRawRecord;
+import io.netty.handler.codec.dns.DnsRecord;
+
+/**
+ * Decodes an {@link InetAddress} from an A or AAAA {@link DnsRawRecord}.
+ */
+final class DnsAddressDecoder {
+
+    private static final int INADDRSZ4 = 4;
+    private static final int INADDRSZ6 = 16;
```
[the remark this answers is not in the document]
trustin: They are private fields and highly unlikely to change, so I'd like to keep these as they are.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsAddressResolveContext.java
```diff
@@ -0,0 +1,74 @@
...
+        return new DnsAddressResolveContext(parent, hostname, additionals, nameServerAddrs, resolveCache);
+    }
+
+    @Override
+    InetAddress convertRecord(DnsRecord record, String hostname, DnsRecord[] additionals, EventLoop eventLoop) {
+        return decodeAddress(record, hostname, parent.isDecodeIdn());
+    }
+
+    @Override
+    boolean gotResult(List<InetAddress> finalResult) {
```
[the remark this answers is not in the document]
trustin: Renamed to `containsExpectedResult`

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -527,6 +534,94 @@ protected void doResolve(String inetHost, Promise<InetAddress> promise) throws E
...
+                    }
+                } else if (hostsFileEntry instanceof Inet6Address) {
+                    if (type == DnsRecordType.AAAA) {
+                        content = Unpooled.wrappedBuffer(hostsFileEntry.getAddress());
+                    }
+                }
+
+                if (content != null) {
+                    trySuccess(promise, Collections.<DnsRecord>singletonList(
+                            new DefaultDnsRawRecord(hostname, type, 86400, content)));
```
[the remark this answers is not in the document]
trustin: Added

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -527,6 +534,94 @@ protected void doResolve(String inetHost, Promise<InetAddress> promise) throws E
...
+                    return promise;
+                }
+            }
+        }
+
+        // It was not A/AAAA question or there was no entry in /etc/hosts.
```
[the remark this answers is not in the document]
trustin: Initial cache API was designed to cache the complete DNS response, but it caused various issues with reference counting because it kept `DnsRawRecord`. We may be able to make this change once we implement all `DnsRecord` implementations for all known DNS record types. For now, we can't use cache for arbitrary record types.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -638,40 +733,20 @@ private static void tryFailure(Promise<?> promise, Throwable cause) {
...
-            for (int i = 0; i < numEntries; i++) {
-                final InetAddress a = resolvedEntries.get(i).address();
-                if (addressType.isInstance(a)) {
-                    trySuccess(promise, a);
-                    return true;
+        final Promise<List<InetAddress>> allPromise = executor().newPromise();
+        doResolveAllUncached(hostname, additionals, allPromise, resolveCache);
+        allPromise.addListener(new FutureListener<List<InetAddress>>() {
+            @Override
+            public void operationComplete(Future<List<InetAddress>> future) throws Exception {
```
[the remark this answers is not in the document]
trustin: Fixed

resolver-dns/src/main/java/io/netty/resolver/dns/DnsRecordResolveContext.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Override
+    DnsRecord convertRecord(DnsRecord record, String hostname, DnsRecord[] additionals, EventLoop eventLoop) {
+        return ReferenceCountUtil.retain(record);
+    }
+
+    @Override
+    boolean gotResult(List<DnsRecord> finalResult) {
+        return true;
```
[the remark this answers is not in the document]
trustin: It never gives an empty list by contract. Let me update the Javadoc of this method.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsRecordResolveContext.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Override
+    void cache(String hostname, DnsRecord[] additionals, DnsRecord result, DnsRecord convertedResult) {
+        // Do not cache.
```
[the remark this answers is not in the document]
trustin: Added comment

resolver-dns/src/main/java/io/netty/resolver/dns/DnsResolveContext.java
```diff
@@ -618,24 +634,24 @@ void tryToFinishResolve(final DnsServerAddressStream nameServerAddrStream,
...
             // There are still some queries we did not receive responses for.
-            if (gotPreferredAddress()) {
-                // But it's OK to finish the resolution process if we got a resolved address of the preferred type.
+            if (finalResult != null && gotResult(finalResult)) {
+                // But it's OK to finish the resolution process if we got something expected.
                 finishResolve(promise, cause);
             }
 
-            // We did not get any resolved address of the preferred type, so we can't finish the resolution process.
+            // We did not get any thing interesting, so we can't finish the resolution process.
```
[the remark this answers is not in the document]
trustin: Fixed

resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
trustin: I actually checked if the result contains a valid MX record, just like `queryMx()` does. The log messages are for debugging. `queryMx()` logs the same. And the test case uses the test DNS server rather than a public DNS server.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsResolveContext.java
trustin: Actually, it was a bug. This change has to be reverted.

trustin:
Ready for another round.

trustin:
Thanks!

PR #7805, opened 2018-03-22
Copy direct ByteBuf with buffer to OutputStream [WIP]

### LINKED ISSUE #7804: Transferring a ByteBuf to Outputstream without duplicating memory
(running io.netty.buffer:4.1.21.Final)

I have a Java Output stream (java.io.OutputStream) I want to incrementally copy a ByteBuffer too.

The byte buffer is off heap and big.
I naively used:
    public abstract ByteBuf getBytes(int index, OutputStream out, int length) throws IOException;

assuming that it would copy all the data to the output stream.

It does, but to do so it copies the whole native buffer into the heap! I do not have that much heap, resulting in:

```
java.lang.OutOfMemoryError: Java heap space 
at io.netty.util.internal.PlatformDependent.allocateUninitializedArray(PlatformDependent.java:200) ~[40:io.netty.common:4.1.21.Final] 
at io.netty.buffer.UnpooledUnsafeHeapByteBuf.allocateArray(UnpooledUnsafeHeapByteBuf.java:34) ~[34:io.netty.buffer:4.1.21.Final] 
at io.netty.buffer.UnpooledByteBufAllocator$InstrumentedUnpooledUnsafeHeapByteBuf.allocateArray(UnpooledByteBufAllocator.java:144) ~[34:io.netty.buffer:4.1.21.Final] 
at io.netty.buffer.UnpooledHeapByteBuf.<init>(UnpooledHeapByteBuf.java:60) ~[34:io.netty.buffer:4.1.21.Final] 
at io.netty.buffer.UnpooledUnsafeHeapByteBuf.<init>(UnpooledUnsafeHeapByteBuf.java:29) ~[34:io.netty.buffer:4.1.21.Final] 
at io.netty.buffer.UnpooledByteBufAllocator$InstrumentedUnpooledUnsafeHeapByteBuf.<init>(UnpooledByteBufAllocator.java:139) ~[34:io.netty.buffer:4.1.21.Final] 
at io.netty.buffer.UnpooledByteBufAllocator.newHeapBuffer(UnpooledByteBufAllocator.java:82) ~[34:io.netty.buffer:4.1.21.Final] 
at io.netty.buffer.AbstractByteBufAllocator.heapBuffer(AbstractByteBufAllocator.java:166) ~[34:io.netty.buffer:4.1.21.Final] 
at io.netty.buffer.AbstractByteBufAllocator.heapBuffer(AbstractByteBufAllocator.java:157) ~[34:io.netty.buffer:4.1.21.Final] 
at io.netty.buffer.UnsafeByteBufUtil.getBytes(UnsafeByteBufUtil.java:572) ~[34:io.netty.buffer:4.1.21.Final] 
at io.netty.buffer.UnpooledUnsafeDirectByteBuf.getBytes(UnpooledUnsafeDirectByteBuf.java:365) ~[34:io.netty.buffer:4.1.21.Final] 
```

I checked io.netty.buffer.UnsafeByteBufUtil.getBytes(AbstractByteBuf, long, int, OutputStream, int)
and behold. It creates a heap bytebuffer of that size to copy the array into it and then to copy that to the output stream.

Bummer!

Shouldn't netty do this better by itself? it seems that I need to manually do this by using a (reasonably sized) byte array and read from byteBuf into that and flush that to my stream in a loop

CodingFabian:
Motivation:

Currently copying a direct byte buf copies it fully into the heap before writing it to an output stream.
This is not really useful in most cases. if i wanted it completely in memory i would use a byte[].

Modification:
copy the bytebuf contents via an 8k buffer into the output stream

Result:

Fixes #7804

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/UnpooledDirectByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/UnpooledDirectByteBuf.java
@@ -34,6 +34,8 @@ import java.nio.channels.ScatteringByteChannel;
  */
 public class UnpooledDirectByteBuf extends AbstractReferenceCountedByteBuf {
 
+    private static final int BUFFER_SIZE = 8192;
+
     private final ByteBufAllocator alloc;
 
     private ByteBuffer buffer;
@@ -482,36 +484,32 @@ public class UnpooledDirectByteBuf extends AbstractReferenceCountedByteBuf {
 
     @Override
     public ByteBuf getBytes(int index, OutputStream out, int length) throws IOException {
-        getBytes(index, out, length, false);
-        return this;
-    }
-
-    private void getBytes(int index, OutputStream out, int length, boolean internal) throws IOException {
         ensureAccessible();
         if (length == 0) {
-            return;
+            return this;
         }
 
         if (buffer.hasArray()) {
             out.write(buffer.array(), index + buffer.arrayOffset(), length);
         } else {
-            byte[] tmp = new byte[length];
-            ByteBuffer tmpBuf;
-            if (internal) {
-                tmpBuf = internalNioBuffer();
-            } else {
-                tmpBuf = buffer.duplicate();
+            byte[] tmp = new byte[Math.min(length, BUFFER_SIZE)];
+            int pos = 0;
+            int remaining = length;
+            while (remaining > 0) {
+                int chunkSize = Math.min(remaining, BUFFER_SIZE);
+                buffer.get(tmp, 0, chunkSize);
+                out.write(tmp, 0, chunkSize);
+                pos += chunkSize;
+                remaining -= chunkSize;
             }
-            tmpBuf.clear().position(index);
-            tmpBuf.get(tmp);
-            out.write(tmp);
         }
+        return this;
     }
 
     @Override
     public ByteBuf readBytes(OutputStream out, int length) throws IOException {
         checkReadableBytes(length);
-        getBytes(readerIndex, out, length, true);
+        getBytes(readerIndex, out, length);
         readerIndex += length;
         return this;
     }
--- a/buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
+++ b/buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
@@ -34,6 +34,7 @@ import static io.netty.util.internal.PlatformDependent.BIG_ENDIAN_NATIVE_ORDER;
 final class UnsafeByteBufUtil {
     private static final boolean UNALIGNED = PlatformDependent.isUnaligned();
     private static final byte ZERO = 0;
+    private static final int BUFFER_SIZE = 8192;
 
     static byte getByte(long address) {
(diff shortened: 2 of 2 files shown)


CodingFabian:
same problem also exists in io.netty.buffer.UnpooledDirectByteBuf.getBytes(int, OutputStream, int, boolean)
but due to time constraints I wanted to put this out there first for everybody to have a look ;)

buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
```diff
@@ -583,14 +584,15 @@ private static void setSingleBytes(final AbstractByteBuf buf, final long addr, f
     static void getBytes(AbstractByteBuf buf, long addr, int index, OutputStream out, int length) throws IOException {
         buf.checkIndex(index, length);
         if (length != 0) {
-            ByteBuf tmpBuf = buf.alloc().heapBuffer(length);
```
CodingFabian: isnt this just a very complicated way of making new byte[] ? if I missed some nuance, please let me know ;)

normanmaurer: @CodingFabian I think the idea was that if you use the pooled allocator you also have heap memory pooled by default and so if doing it like this can reduce allocations. makes sense ?

CodingFabian: except that it most likely doesnt work here anyway, but that might change witht the 8k buffer.
will add it back. thx

buffer/src/main/java/io/netty/buffer/UnpooledDirectByteBuf.java
```diff
@@ -482,36 +484,32 @@ public ByteBuf setBytes(int index, ByteBuffer src) {
...
         if (buffer.hasArray()) {
             out.write(buffer.array(), index + buffer.arrayOffset(), length);
         } else {
-            byte[] tmp = new byte[length];
-            ByteBuffer tmpBuf;
-            if (internal) {
-                tmpBuf = internalNioBuffer();
-            } else {
-                tmpBuf = buffer.duplicate();
+            byte[] tmp = new byte[Math.min(length, BUFFER_SIZE)];
```
CodingFabian: did not use alloc() here because it is unclear for me if array().arrayOffset() returns non 0.

ByteBuffer.get() cannot work with a destination offset.

normanmaurer:
@CodingFabian please fix the test errors:

https://garage.netty.io/teamcity/viewLog.html?buildId=29187&buildTypeId=pulls_netty&guest=1

CodingFabian:
not yet that far that I look at the tests. That underflow is quite confusing tbh.
currently looking for more places that could benefit from this change

normanmaurer:
@CodingFabian I just implemented it myself here https://github.com/netty/netty/pull/7813. All tests pass.

PTAL and hope you not mind

CodingFabian:
No I do not mind - Sorry I could not invest more time!

Issue #3698, opened 2015-04-28
Cannot use SCTP unordered flag


benevans:
Some applications, e.g. SIP over SCTP ([RFC4168](https://tools.ietf.org/html/rfc4168#section-5.1)), require the use of SCTP's unordered delivery flag.

Currently in Netty there is no way to set the unordered flag on an outgoing message. I assume the place to do this would be in `io.netty.channel.sctp.SctpMessage`, alongside the payload protocol ID and stream ID.

I'm happy to submit a PR but just wanted to check first in case there was some reason why this wasn't implemented. This applies to Netty 4.0 and later.

jestan:
@benevans Current io.netty.channel.sctp.SctpMessage is hiding some controls and, it could have been done better. Initially  it was designed to  work with SIGTRAN protocols and I didn't include ordering/completeness flags, which were not used at that time. Please feel free to send your pull request.

normanmaurer:
@benevans @jestan just be sure we need to keep API compat if you want to modify it for 4.0

normanmaurer:
@jestan I assigned it to you ;)

jestan:
@normanmaurer sure, we will do it without breaking the API.

normanmaurer:
@jestan @benevans thanks! Looking forward to a PR !

Issue #7740, opened 2018-02-21
HttpServerUpgrader should add the new handler before removing the old


mosesn:
There's a potential race condition in the netty HTTP server upgrade path.  The client sends an upgrade request, the server responds with a 101, and then the client sends another message.  If the server has partially read this message before it does the upgrade, then it will propagate the bytes, but if this happens *before* the replacement happens, then we can be in a sticky situations.

In the HTTP/2 case, the client sends the connection preface as soon as it receives the 101, so this isn't academic–we have run into this situation.  Here's what the stack trace looked like:

```
java.lang.AssertionError: assertion failed
	at scala.Predef$.assert(Predef.scala:156)
	at com.twitter.finagle.netty4.http.handler.UnpoolHttpHandler$.channelRead(UnpoolHttpHandler.scala:35)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
	at io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:108)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
	at io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:102)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
	at io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:102)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
[...]

PR #7807, opened 2018-03-23
HttpServerUpgradeHandler shouldn't wait for flush to reshape pipeline

bryce-anderson:
Motivation:

There is a race between both flushing the upgrade response and receiving
more data before the flush ChannelPromise can fire the listener and reshape
the pipeline. Since We have already committed to an upgrade by writing the
upgrade response, we should immediately prepare for handling the next 
protocol.

Modifications:

The pipeline reshaping logic in HttpServerUpgradeHandler has been moved
out of the ChannelFutureListener attached to the write of the upgrade
response and happens immediately after the writeAndFlush call, but
before the method returns.

Result:

The pipeline is no longer subject to receiving more data before the
pipeline has been reformed.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpServerUpgradeHandler.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpServerUpgradeHandler.java
@@ -314,31 +314,31 @@ public class HttpServerUpgradeHandler extends HttpObjectAggregator {
         // Create the user event to be fired once the upgrade completes.
         final UpgradeEvent event = new UpgradeEvent(upgradeProtocol, request);
 
-        final UpgradeCodec finalUpgradeCodec = upgradeCodec;
-        ctx.writeAndFlush(upgradeResponse).addListener(new ChannelFutureListener() {
-            @Override
-            public void operationComplete(ChannelFuture future) throws Exception {
-                try {
-                    if (future.isSuccess()) {
-                        // Perform the upgrade to the new protocol.
-                        sourceCodec.upgradeFrom(ctx);
-                        finalUpgradeCodec.upgradeTo(ctx, request);
+        // After writing the upgrade response we immediately prepare the
+        // pipeline for the next protocol to avoid a race between completion
+        // of the write future and receiving data before the pipeline is
+        // restructured.
+        try {
+            final ChannelFuture writeComplete = ctx.writeAndFlush(upgradeResponse);
+            // Perform the upgrade to the new protocol.
+            sourceCodec.upgradeFrom(ctx);
+            upgradeCodec.upgradeTo(ctx, request);
 
-                        // Notify that the upgrade has occurred. Retain the event to offset
-                        // the release() in the finally block.
-                        ctx.fireUserEventTriggered(event.retain());
+            // Remove this handler from the pipeline.
+            ctx.pipeline().remove(HttpServerUpgradeHandler.this);
 
-                        // Remove this handler from the pipeline.
-                        ctx.pipeline().remove(HttpServerUpgradeHandler.this);
-                    } else {
-                        future.channel().close();
-                    }
-                } finally {
-                    // Release the event if the upgrade event wasn't fired.
-                    event.release();
-                }
-            }
-        });
+            // Notify that the upgrade has occurred. Retain the event to offset
+            // the release() in the finally block.
+            ctx.fireUserEventTriggered(event.retain());
+
+            // Add the listener last to avoid firing upgrade logic after
+            // the channel is already closed since the listener may fire
+            // immediately if the write failed eagerly.
+            writeComplete.addListener(ChannelFutureListener.CLOSE_ON_FAILURE);
+        } finally {
+            // Release the event if the upgrade event wasn't fired.
+            event.release();
+        }
         return true;
     }
 
--- a/codec-http/src/test/java/io/netty/handler/codec/http/HttpServerUpgradeHandlerTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/HttpServerUpgradeHandlerTest.java
@@ -1,0 +1,144 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.codec.http;
+
+import java.util.Collection;
+import java.util.Collections;
+
+import org.junit.Test;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import io.netty.channel.ChannelDuplexHandler;
+import io.netty.channel.ChannelFuture;
+import io.netty.channel.ChannelFutureListener;
+import io.netty.channel.ChannelHandler;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.channel.ChannelPromise;
+import io.netty.channel.embedded.EmbeddedChannel;
+import io.netty.handler.codec.http.HttpServerUpgradeHandler.UpgradeCodec;
+import io.netty.handler.codec.http.HttpServerUpgradeHandler.UpgradeCodecFactory;
+import io.netty.util.CharsetUtil;
+
+import static org.junit.Assert.*;
+
+public class HttpServerUpgradeHandlerTest {
+
+    private class TestUpgradeCodec implements UpgradeCodec {
+        @Override
+        public Collection<CharSequence> requiredUpgradeHeaders() {
+            return Collections.<CharSequence>emptyList();
+        }
+
+        @Override
+        public boolean prepareUpgradeResponse(ChannelHandlerContext ctx, FullHttpRequest upgradeRequest,
+                                              HttpHeaders upgradeHeaders) {
+            return true;
+        }
+
+        @Override
+        public void upgradeTo(ChannelHandlerContext ctx, FullHttpRequest upgradeRequest) {
+            // Ensure that the HttpServerUpgradeHandler is still installed when this is called
+            assertEquals(ctx.pipeline().context(HttpServerUpgradeHandler.class), ctx);
+            assertNotNull(ctx.pipeline().get(HttpServerUpgradeHandler.class));
+
+            // Add a marker handler to signal that the upgrade has happened
+            ctx.pipeline().addAfter(ctx.name(), "marker", new ChannelInboundHandlerAdapter());
+          }
+    }
+
+    @Test
+    public void upgradesPipelineInSameMethodInvocation() {
+        final HttpServerCodec httpServerCodec = new HttpServerCodec();
+        final UpgradeCodecFactory factory = new UpgradeCodecFactory() {
+            @Override
+            public UpgradeCodec newUpgradeCodec(CharSequence protocol) {
+                return new TestUpgradeCodec();
+            }
+        };
+
+        ChannelHandler testInStackFrame = new ChannelDuplexHandler() {
+            // marker boolean to signal that we're in the `channelRead` method
+            private boolean inReadCall;
+            private boolean writeUpgradeMessage;
+            private boolean writeFlushed;
+
+            @Override
+            public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+                assertFalse(inReadCall);
+                assertFalse(writeUpgradeMessage);
+
+                inReadCall = true;
+                try {
+                    super.channelRead(ctx, msg);
+                    // All in the same call stack, the upgrade codec should receive the message,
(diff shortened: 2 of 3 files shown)


normanmaurer:
Can we have a unit test ?

bryce-anderson:
I *think* this should fix #3698, but am not 100% sure. Since the `HttpServerUpgradeHandler` is an `HttpObjectAggregator`, it should clear the inbound pipeline of the entire initial request, so unless the client sent and upgrade request and then tried to pipeline another (which is incorrect) the inbound socket channel should be clean at the point the pipeline is reshaped.

I'm also open to suggestions for tests. Currently there aren't any dedicated tests for `HttpServerUpgradeHandler`, but I may try and get one in.

codec-http/src/test/java/io/netty/handler/codec/http/HttpServerUpgradeHandlerTest.java
```diff
@@ -0,0 +1,108 @@
+package io.netty.handler.codec.http;
```
normanmaurer: please add a license header.

codec-http/src/test/java/io/netty/handler/codec/http/HttpServerUpgradeHandlerTest.java
```diff
@@ -0,0 +1,110 @@
...
+import io.netty.channel.embedded.EmbeddedChannel;
+import io.netty.handler.codec.http.HttpServerUpgradeHandler.UpgradeCodec;
+import io.netty.handler.codec.http.HttpServerUpgradeHandler.UpgradeCodecFactory;
+import io.netty.util.CharsetUtil;
+
+import static org.junit.Assert.*;
+
+public class HttpServerUpgradeHandlerTest {
+
+  private class TestUpgradeCodec implements UpgradeCodec {
```
normanmaurer: we use 4 spaces, please fix :)

codec-http/src/test/java/io/netty/handler/codec/http/HttpServerUpgradeHandlerTest.java
```diff
@@ -0,0 +1,110 @@
...
+        "HTTP2-Settings: AAMAAABkAAQAAP__\r\n\r\n";
+    ByteBuf upgrade = Unpooled.buffer().writeBytes(upgradeString.getBytes(CharsetUtil.US_ASCII));
+
+    assertFalse(channel.writeInbound(upgrade));
+
+    // Make sure the pipeline has been restructured even though the write hasn't flushed
+    assertNull(channel.pipeline().get(HttpServerCodec.class));
+    assertNotNull(channel.pipeline().get("marker"));
+
+    channel.finishAndReleaseAll();
```
normanmaurer: add assert

bryce-anderson:
Thanks for the feedback @normanmaurer, I'll address it this afternoon. Cc'ing @ejona86 as he may have comments about correctness.

ejona86:
Didn't look in great detail, but this makes me feel better. There's really no reason I can think of to delay until the write is complete and I agree it just seems to add a race.

bryce-anderson:
Thanks for the double check @ejona86. @normanmaurer, I think I addressed your comments. Thanks again for the review.

codec-http/src/test/java/io/netty/handler/codec/http/HttpServerUpgradeHandlerTest.java
```diff
@@ -0,0 +1,124 @@
...
+import io.netty.util.CharsetUtil;
+
+import static org.junit.Assert.*;
+
+public class HttpServerUpgradeHandlerTest {
+
+    private class TestUpgradeCodec implements UpgradeCodec {
+        @Override
+        public Collection<CharSequence> requiredUpgradeHeaders() {
+            return ImmutableList.of();
```
normanmaurer: Can we just use `Collections.<CharSequence>emptyList()` and so remove the dependency on the guava here ?

codec-http/src/test/java/io/netty/handler/codec/http/HttpServerUpgradeHandlerTest.java
```diff
@@ -0,0 +1,124 @@
...
+                    inReadCall = false;
+                }
+            }
+
+            @Override
+            public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) throws Exception {
+                // We ensure that we're in the read call, but we don't flush the message
+                // to make sure that the pipeline upgrade machinery isn't dependant on
+                // the message being flushed.
+                assertTrue(inReadCall);
```
normanmaurer: Call `super.write(....)` to ensure we not "leak the message" and also to ensure the promise is notified.

codec-http/src/test/java/io/netty/handler/codec/http/HttpServerUpgradeHandlerTest.java
```diff
@@ -0,0 +1,124 @@
...
+        HttpServerUpgradeHandler upgradeHandler = new HttpServerUpgradeHandler(httpServerCodec, factory);
+
+        EmbeddedChannel channel = new EmbeddedChannel(testInStackFrame, httpServerCodec, upgradeHandler);
+
+        String upgradeString = "GET / HTTP/1.1\r\n" +
+            "Host: example.com\r\n" +
+            "Connection: Upgrade, HTTP2-Settings\r\n" +
+            "Upgrade: h2c\r\n" +
+            "HTTP2-Settings: AAMAAABkAAQAAP__\r\n\r\n";
+        ByteBuf upgrade = Unpooled.buffer().writeBytes(upgradeString.getBytes(CharsetUtil.US_ASCII));
```
normanmaurer: just use `Unpooled.copiedBuffer(upgradeString, CharsetUitl.US_ASCII)`

normanmaurer:
@bryce-anderson  #3698 does not seem to be the correct issue to reference here... Can you check ?

codec-http/src/test/java/io/netty/handler/codec/http/HttpServerUpgradeHandlerTest.java
bryce-anderson: This was deliberate so that we could be confident that the pipeline reshaping happened in this call stack irrespective of the promise being satisfied.

That said, perhaps it's not something that is likely to regress. And we do need to do something with the outbound message, good catch.

bryce-anderson:
@normanmaurer, sorry, you're right: the issue it attempts to fix is #7740.

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerUpgradeHandler.java
```diff
@@ -313,32 +313,32 @@ private boolean upgrade(final ChannelHandlerContext ctx, final FullHttpRequest r
...
+            // Perform the upgrade to the new protocol.
+            sourceCodec.upgradeFrom(ctx);
+            upgradeCodec.upgradeTo(ctx, request);
+
+            // Remove this handler from the pipeline.
+            ctx.pipeline().remove(HttpServerUpgradeHandler.this);
+
+            // Notify that the upgrade has occurred. Retain the event to offset
+            // the release() in the finally block.
+            ctx.fireUserEventTriggered(event.retain());
```
nepthar: Is there a reason to delay this until the write succeeds? (I can't think of any)

codec-http/src/test/java/io/netty/handler/codec/http/HttpServerUpgradeHandlerTest.java
```diff
@@ -0,0 +1,126 @@
...
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.channel.ChannelPromise;
+import io.netty.channel.embedded.EmbeddedChannel;
+import io.netty.handler.codec.http.HttpServerUpgradeHandler.UpgradeCodec;
+import io.netty.handler.codec.http.HttpServerUpgradeHandler.UpgradeCodecFactory;
+import io.netty.util.CharsetUtil;
+
+import static org.junit.Assert.*;
+
+public class HttpServerUpgradeHandlerTest {
```
nepthar: this test LGTM, but as a sanity check, does it fail without your change?

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerUpgradeHandler.java
bryce-anderson: The write hasn't necessarily succeeded yet as `writeComplete` is a `ChannelPromise`. We should enqueue the write before reshaping the pipeline as the incoming codec _may_ try to write something when it gets added to the pipeline.

codec-http/src/test/java/io/netty/handler/codec/http/HttpServerUpgradeHandlerTest.java
bryce-anderson: No, not anymore. Originally it did, but now that the `write` override on line 96 calls `super.write` as @normanmaurer requested, it will succeed either way. See his second round of comments.

I'm happy to take suggestions on how to test this cleanly, but the only thing I can think of is to drop the promise (what I had originally, and it did fail without my patch) or to store it somewhere else and fail it later with the former being bad form and the latter being a pretty ugly mess.

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerUpgradeHandler.java
normanmaurer: Maybe just delay the write by using ctx.eventLoop().execute(...) and schedule the write via it ?

bryce-anderson: That did the trick, thanks @normanmaurer. I'd tried to defer the satisfaction of the promise via indirection through the executor, but the `flush()` call would run it. Your suggestion is much more reasonable anyway.

codec-http/src/test/java/io/netty/handler/codec/http/HttpServerUpgradeHandlerTest.java
bryce-anderson: Using normans suggestion it now fails without the changes to `HttpServerUpgradeHandler`.

normanmaurer: @bryce-anderson did you forget to push this change ?

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerUpgradeHandler.java
nepthar: Ah, shoot, my comment was really unclear - I wanted to note that we `fireUserEventTriggered` before the write is complete and ask if there could be a reason to delay that event until the write finished instead. I'm not sure what the contract around that user event is, but maybe some folks rely on that happening after the write?

codec-http/src/test/java/io/netty/handler/codec/http/HttpServerUpgradeHandlerTest.java
bryce-anderson: @normanmaurer yes, I had. Sorry about that.

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerUpgradeHandler.java
```diff
@@ -313,32 +313,32 @@ private boolean upgrade(final ChannelHandlerContext ctx, final FullHttpRequest r
 
         // Create the user event to be fired once the upgrade completes.
         final UpgradeEvent event = new UpgradeEvent(upgradeProtocol, request);
+        final ChannelFuture writeComplete = ctx.writeAndFlush(upgradeResponse);
```
Scottmitch: write and flush *shouldn't* throw ... but consider including it in the `try` block below just to have less assumptions about releasing the `event` relative to behavior of this method.

codec-http/src/test/java/io/netty/handler/codec/http/HttpServerUpgradeHandlerTest.java
```diff
@@ -0,0 +1,145 @@
...
+                } finally {
+                    inReadCall = false;
+                }
+            }
+
+            @Override
+            public void write(final ChannelHandlerContext ctx, final Object msg,
+                              final ChannelPromise promise) throws Exception {
+                // We ensure that we're in the read call and defer the write
+                // so we can
```
Scottmitch: nit: just put this on the previous line. line length of 120 is permitted.

codec-http/src/test/java/io/netty/handler/codec/http/HttpServerUpgradeHandlerTest.java
```diff
@@ -0,0 +1,145 @@
...
+                writeUpgradeMessage = true;
+                ctx.channel().eventLoop().execute(new Runnable() {
+                    @Override
+                    public void run() {
+                        ctx.write(msg, promise);
+                    }
+                });
+                promise.addListener(new ChannelFutureListener() {
+                    @Override
+                    public void operationComplete(ChannelFuture future) throws Exception {
```
Scottmitch: nit: remove `throws Exception`

normanmaurer:
@bryce-anderson please address @Scottmitch s comments and ping me once done. I will pull in then.

codec-http/src/test/java/io/netty/handler/codec/http/HttpServerUpgradeHandlerTest.java
bryce-anderson: The bigger issue was that the comment was incomplete. Sorry about that.

bryce-anderson:
Thanks for the feedback @Scottmitch. @normanmaurer, should be ready. Thanks again for the reviews folks.

normanmaurer:
Will merge once CI completes @bryce-anderson ... thanks!

PR #7813, opened 2018-03-28
Limit the number of bytes to use to copy the content of a direct buff…

### LINKED ISSUE #7804: Transferring a ByteBuf to Outputstream without duplicating memory

normanmaurer:
…er to an Outputstream

Motivation:

Currently copying a direct ByteBuf copies it fully into the heap before writing it to an output stream.
The can result in huge memory usage on the heap.

Modification:

copy the bytebuf contents via an 8k buffer into the output stream

Result:

Fixes #7804

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
+++ b/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
@@ -27,6 +27,8 @@ import io.netty.util.internal.SystemPropertyUtil;
 import io.netty.util.internal.logging.InternalLogger;
 import io.netty.util.internal.logging.InternalLoggerFactory;
 
+import java.io.IOException;
+import java.io.OutputStream;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.CharBuffer;
@@ -64,6 +66,7 @@ public final class ByteBufUtil {
     private static final int MAX_BYTES_PER_CHAR_UTF8 =
             (int) CharsetUtil.encoder(CharsetUtil.UTF_8).maxBytesPerChar();
 
+    static final int WRITE_CHUNK_SIZE = 8192;
     static final ByteBufAllocator DEFAULT_ALLOCATOR;
 
     static {
@@ -1390,6 +1393,44 @@ public final class ByteBufUtil {
             }
         }
         return true;
+    }
+
+    /**
+     * Read bytes from the given {@link ByteBuffer} into the given {@link OutputStream} using the {@code position} and
+     * {@code length}. The position and limit of the given {@link ByteBuffer} may be adjusted.
+     */
+    static void readBytes(ByteBufAllocator allocator, ByteBuffer buffer, int position, int length, OutputStream out)
+            throws IOException {
+        if (buffer.hasArray()) {
+            out.write(buffer.array(), position + buffer.arrayOffset(), length);
+        } else {
+            int chunkLen = Math.min(length, WRITE_CHUNK_SIZE);
+            buffer.clear().position(position);
+
+            if (allocator.isDirectBufferPooled()) {
+                // if direct buffers are pooled chances are good that heap buffers are pooled as well.
+                ByteBuf tmpBuf = allocator.heapBuffer(chunkLen);
+                try {
+                    byte[] tmp = tmpBuf.array();
+                    int offset = tmpBuf.arrayOffset();
+                    getBytes(buffer, tmp, offset, chunkLen, out, length);
+                } finally {
+                    tmpBuf.release();
+                }
+            } else {
+                getBytes(buffer, new byte[chunkLen], 0, chunkLen, out, length);
+            }
+        }
+    }
+
+    private static void getBytes(ByteBuffer inBuffer, byte[] in, int inOffset, int inLen, OutputStream out, int outLen)
+            throws IOException {
+        do {
+            int len = Math.min(inLen, outLen);
+            inBuffer.get(in, inOffset, len);
+            out.write(in, inOffset, len);
+            outLen -= len;
+        } while (outLen > 0);
     }
 
     private ByteBufUtil() { }
--- a/buffer/src/main/java/io/netty/buffer/UnpooledDirectByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/UnpooledDirectByteBuf.java
@@ -491,21 +491,7 @@ public class UnpooledDirectByteBuf extends AbstractReferenceCountedByteBuf {
         if (length == 0) {
             return;
         }
-
-        if (buffer.hasArray()) {
-            out.write(buffer.array(), index + buffer.arrayOffset(), length);
-        } else {
-            byte[] tmp = new byte[length];
-            ByteBuffer tmpBuf;
-            if (internal) {
-                tmpBuf = internalNioBuffer();
-            } else {
-                tmpBuf = buffer.duplicate();
-            }
-            tmpBuf.clear().position(index);
-            tmpBuf.get(tmp);
-            out.write(tmp);
-        }
+        ByteBufUtil.readBytes(alloc(), internal ? internalNioBuffer() : buffer.duplicate(), index, length, out);
     }
 
     @Override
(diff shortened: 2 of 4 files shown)


normanmaurer:
@CodingFabian PTAL as well

buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
```diff
@@ -583,18 +583,35 @@ private static void setSingleBytes(final AbstractByteBuf buf, final long addr, f
...
-            try {
-                byte[] tmp = tmpBuf.array();
-                int offset = tmpBuf.arrayOffset();
-                PlatformDependent.copyMemory(addr, tmp, offset, length);
-                out.write(tmp, offset, length);
-            } finally {
-                tmpBuf.release();
+            int len = Math.min(length, ByteBufUtil.WRITE_CHUNK_SIZE);
+            if (buf.alloc().isDirectBufferPooled()) {
+                // if direct buffers are pooled changes are good that heap buffers are pooled as well.
```
trustin: changes -> chances

buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
```diff
@@ -583,18 +583,34 @@ private static void setSingleBytes(final AbstractByteBuf buf, final long addr, f
...
+                } finally {
+                    tmpBuf.release();
+                }
+            } else {
+                getBytes(addr, new byte[len], 0, len, out, length);
             }
         }
     }
 
+    private static void getBytes(long addr, byte[] array, int offset, int chunkLen, OutputStream out, int length)
```
trustin: How about:
- addr -> inAddr
- array -> in
- offset -> inOffset
- chunkLen -> inLen
- length -> outLen

buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
```diff
@@ -583,18 +583,34 @@ private static void setSingleBytes(final AbstractByteBuf buf, final long addr, f
...
         if (length != 0) {
-            ByteBuf tmpBuf = buf.alloc().heapBuffer(length);
-            try {
-                byte[] tmp = tmpBuf.array();
-                int offset = tmpBuf.arrayOffset();
-                PlatformDependent.copyMemory(addr, tmp, offset, length);
-                out.write(tmp, offset, length);
-            } finally {
-                tmpBuf.release();
+            int len = Math.min(length, ByteBufUtil.WRITE_CHUNK_SIZE);
```
trustin: Can we just do: `Math.min(Math.min(length, ByteBufUtil.WRITE_CHUNK_SIZE), length)`? This would eliminate the reason to pass 2 length parameters to the `getBytes()` below.

normanmaurer: hmm wouldn't we still need it ? I am confused why we not need it with this change ... Can you explain

buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
normanmaurer: done

buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
trustin: Because I'm an idiot? Didn't notice `outLen` is updated at the end of the loop. :-)

buffer/src/main/java/io/netty/buffer/UnpooledDirectByteBuf.java
```diff
@@ -491,21 +491,7 @@ private void getBytes(int index, OutputStream out, int length, boolean internal)
...
-            if (internal) {
-                tmpBuf = internalNioBuffer();
-            } else {
-                tmpBuf = buffer.duplicate();
-            }
-            tmpBuf.clear().position(index);
-            tmpBuf.get(tmp);
-            out.write(tmp);
-        }
+        ByteBufUtil.readBytes(internal ? internalNioBuffer() : buffer.duplicate(), index, length, out);
```
CodingFabian: when I tried to improve this I was wondering if we can avoid that duplicate call?

normanmaurer: @CodingFabian nope we can't... `internal` will be false if you use `get*`. This is needed as it may be called from multiple threads at the same time which should be fine as you not expect to do any internal updates on the indices. If you use a `get*` method there will be no duplicate invoked

buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java
normanmaurer: lol ok that explains why =P

buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
```diff
@@ -1392,5 +1395,25 @@ private static boolean isUtf8(ByteBuf buf, int index, int length) {
...
 
+    /**
+     * Read bytes from the given {@link ByteBuffer} into the given {@link OutputStream} using the {@code position} and
+     * {@code length}. The position and limit of the given {@link ByteBuffer} may be adjusted.
+     */
+    static void readBytes(ByteBuffer buffer, int position, int length, OutputStream out) throws IOException {
+        if (buffer.hasArray()) {
+            out.write(buffer.array(), position + buffer.arrayOffset(), length);
+        } else {
+            byte[] tmp = new byte[Math.min(length, WRITE_CHUNK_SIZE)];
```
CodingFabian: so you decided against using alloc() to get this temp array?

normanmaurer: let me check if its possible here as well (in the `Unsafe` portion it is.

normanmaurer: fixed @CodingFabian

PR #7815, opened 2018-03-28
Fix possible stalls in SingleThreadEventExecutor

davidmoten:
Motivation:

I was trying to implement backpressure (only calling `channel.read()` when there was demand) and under high concurrency (>100 threads) was reliably experiencing stalls. As a consequence had a quick look at underlying classes and noticed possible stalls in `SingleThreadEventExecutor`. This doesn't fix my problem but should be fixed anyway.

Two problems were noticed in `SingleThreadEventExecutor` where it is possible that 
* a call to `execute(Runnable)` adds a task to the queue but does not run that task (until `execute` gets called again).

Modification:

The first fix was simply to add a task to the queue *before* calling `startThread`. As it stood it was possible that a thread was started and stopped before the call to `addTask` was made. The consequence being that the task would not be executed till the next call to `execute`.

Similarly, in `startThread` itself the `compareAndSet` call should be in a loop to retry.

Result:

I haven't added unit tests for these as I suspect they would be difficult to induce.

### DIFF
--- a/common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
+++ b/common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
@@ -762,11 +762,9 @@ public abstract class SingleThreadEventExecutor extends AbstractScheduledEventEx
         }
 
         boolean inEventLoop = inEventLoop();
-        if (inEventLoop) {
-            addTask(task);
-        } else {
+        addTask(task);
+        if (!inEventLoop) {
             startThread();
-            addTask(task);
             if (isShutdown() && removeTask(task)) {
                 reject();
             }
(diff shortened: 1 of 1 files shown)


davidmoten:
Looking a bit closer at the class the lifecycle is unclear to me. Is the thread started only once  (I notice it blocks waiting for queued items)?

normanmaurer:
@davidmoten its only started once... So I think these are not needed (as you said).

johnou:
@normanmaurer adding the task before starting the queue wouldn't be a bad idea though.

davidmoten:
I thought so too. At least prevent the thread from needlessly blocking on
an empty queue.

davidmoten:
Stalls were all my own fault. Backpressure working nicely now with Netty channel reads, thanks.

normanmaurer:
@davidmoten thanks... Can you do a pr with just the "blocking queue changes" ?

davidmoten:
sure

PR #7821, opened 2018-03-30
Better handling of streaming JSON data in JsonObjectDecoder

### LINKED ISSUE #7820: JsonObjectDecoder to better handle streaming data
### Description
There seems to be a long running issue with several iterations of fixes for streaming of different types of JSON data:
- The [6156](https://github.com/netty/netty/issues/6156) was the original issue which made the internal index reset together with resetting the decoder state, but it did not cover the case with streaming arrays.
- The [6969](https://github.com/netty/netty/issues/6969) fixed the problem with arrays, but did not cover the case with streaming objects.
- The [7722](https://github.com/netty/netty/issues/7722) is a (currently) opened issue which fixes the problem with streaming objects. While the described solution seems to workaround the problem, it causes multiple reads of the same bytes and technically does not fully recover the state (different branches are executed for the repeated reads).

A better approach seems to be to just properly recover the internal decoder index given the information about how far the decoder went from the readerIndex and the last readerIndex are already there which seems to cover both the cases with streaming arrays and objects.

### Netty version
4.1.22.Final

ikurovsky:
Motivation:

When the JsonObjectDecoder determines that the incoming buffer had some data discarded, it resets the internal index to readerIndex and attempts to adjust the state which does not correctly work for streams of JSON objects.

Modifications:

Reset the internal index to the value considering the previous reads.

Result:

JsonObjectDecoder correctly handles streams of both JSON objects and arrays with no state adjustments or repeatable reads.

Fixes #7820

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/json/JsonObjectDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/json/JsonObjectDecoder.java
@@ -91,11 +91,7 @@ public class JsonObjectDecoder extends ByteToMessageDecoder {
         }
 
         if (this.idx > in.readerIndex() && lastReaderIndex != in.readerIndex()) {
-            this.idx = in.readerIndex();
-            if (state == ST_DECODING_ARRAY_STREAM) {
-                insideString = false;
-                openBraces = 1;
-            }
+            this.idx = in.readerIndex() + (idx - lastReaderIndex);
         }
 
         // index of next byte to process.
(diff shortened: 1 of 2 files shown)


normanmaurer:
@ikurovsky can you please sign our icla and let me know once done:

https://netty.io/s/icla

thanks

ikurovsky:
@normanmaurer done

normanmaurer:
@ikurovsky thanks!

PR #7822, opened 2018-03-30
ChunkedWriteHandler: avoid writing two times the same message

### LINKED ISSUE #7819: promise already done caused by ChunkedWriteHandler writing the same (non chunked) message twice
Hello,

I get an exception "promise already done" apparently caused by the ChunkedWriteHandler which is called twice for the same message, one from my flush and the other time from the channelWritabilityChanged.

Not sure if the ChunkedWriteHandler.doFlush method is supposed to be called twice in the same write but it looks to work ok when it's called twice and the message is ChunkedInput.

Maybe lines 304 and 305 should be reversed from:
```
                ctx.write(pendingMessage, currentWrite.promise);
                this.currentWrite = null;
```
to:
```
                this.currentWrite = null;
                ctx.write(pendingMessage, currentWrite.promise);
```

A stacktrace is attached.
[stack-trace.txt](https://github.com/netty/netty/files/1863541/stack-trace.txt)

xpromache:
Motivation:

ChunkedWriteHandler.doFlush is called twice from the same write if the channelWritabilityChanged event is invoked during the write. The buffer is already written so no extra data is sent on the socket but it causes the "promise already done" exception to be thrown. 
This error happens only when the message is not ChunkedInput.

Modification:
Clear out the currentWrite reference before the ctx.write call, such that next time when the method is invoked the same object is not used twice.

Result:

Fixes #7819

### DIFF
--- a/handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java
+++ b/handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java
@@ -299,8 +299,8 @@ public class ChunkedWriteHandler extends ChannelDuplexHandler {
                 ctx.flush();
                 requiresFlush = false;
             } else {
-                ctx.write(pendingMessage, currentWrite.promise);
                 this.currentWrite = null;
+                ctx.write(pendingMessage, currentWrite.promise);
                 requiresFlush = true;
             }
 


normanmaurer:
@xpromache thx

PR #7827, opened 2018-03-31
NetUtil valid IP methods to accept CharSequence

Scottmitch:
Motivation:
NetUtil has methods to determine if a String is a valid IP address. These methods don't rely upon String specific methods and can use CharSequence instead.

Modifications:
- Use CharSequence instead of String for the IP validator methods.
- Avoid object allocation in AsciiString#indexOf(char,int) and reduce
byte code

Result:
No more copy operation required if a CharSequence exists.

### DIFF
--- a/common/src/main/java/io/netty/util/AsciiString.java
+++ b/common/src/main/java/io/netty/util/AsciiString.java
@@ -723,22 +723,16 @@ public final class AsciiString implements CharSequence, Comparable<CharSequence>
      * -1 if found no occurrence.
      */
     public int indexOf(char ch, int start) {
-        if (start < 0) {
-            start = 0;
-        }
-
-        final int thisLen = length();
-
         if (ch > MAX_CHAR_VALUE) {
-            return -1;
+            return INDEX_NOT_FOUND;
         }
 
-        try {
-            return forEachByte(start, thisLen - start, new IndexOfProcessor((byte) ch));
-        } catch (Exception e) {
-            PlatformDependent.throwException(e);
-            return -1;
+        for (int i = start < 0 ? arrayOffset() : start + arrayOffset(); i < length(); ++i) {
+            if (b2c(value[i]) == ch) {
+                return i - arrayOffset();
+            }
         }
+        return -1;
     }
 
     /**
@@ -1832,10 +1826,7 @@ public final class AsciiString implements CharSequence, Comparable<CharSequence>
             return INDEX_NOT_FOUND;
         }
         final int sz = cs.length();
-        if (start < 0) {
-            start = 0;
-        }
-        for (int i = start; i < sz; i++) {
+        for (int i = start < 0 ? 0 : start; i < sz; i++) {
             if (cs.charAt(i) == searchChar) {
                 return i;
             }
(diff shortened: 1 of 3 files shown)


Scottmitch:
Here are preliminary benchmark results via PR https://github.com/netty/netty/pull/7826 (on my noisy laptop):

```
With this change
Benchmark                         Mode  Cnt        Score        Error  Units
NetUtilBenchmark.useIsValidIpv4  thrpt   20  7953593.605 ± 153490.535  ops/s
NetUtilBenchmark.useIsValidIpv6  thrpt   20   708862.308 ±  22429.566  ops/s

Without this change
Benchmark                         Mode  Cnt        Score        Error  Units
NetUtilBenchmark.useIsValidIpv4  thrpt   20  7982387.937 ± 615567.787  ops/s
NetUtilBenchmark.useIsValidIpv6  thrpt   20   788664.794 ±  19830.386  ops/s
```

common/src/main/java/io/netty/util/NetUtil.java
```diff
@@ -461,7 +463,7 @@ public static String bytesToIpAddress(byte[] bytes, int offset, int length) {
         }
     }
 
-    public static boolean isValidIpV6Address(String ip) {
+    public static boolean isValidIpV6Address(CharSequence ip) {
```
normanmaurer: We also need to keep the old method as otherwise the change is not binary compatible.

common/src/main/java/io/netty/util/NetUtil.java
```diff
@@ -622,6 +624,16 @@ private static boolean isValidIPv4Mapped(byte[] bytes, int currentIndex, int com
                 PlatformDependent.isZero(bytes, 0, currentIndex - 3);
     }
 
+    /**
+     * Takes a string and parses it to see if it is a valid IPV4 address.
```
normanmaurer: String / charsequence

common/src/main/java/io/netty/util/AsciiString.java
```diff
@@ -723,22 +723,16 @@ public int indexOf(CharSequence subString, int start) {
...
-        } catch (Exception e) {
-            PlatformDependent.throwException(e);
-            return -1;
+        for (int i = start < 0 ? arrayOffset() : start + arrayOffset(); i < length(); ++i) {
+            if (b2c(value[i]) == ch) {
+                return i - arrayOffset();
+            }
         }
+        return -1;
     }
```
normanmaurer: INDEX_NOT_FOUND

common/src/main/java/io/netty/util/NetUtil.java
Scottmitch: good catch ... done

Scottmitch:
@normanmaurer - note that I also removed internal usage of `forEachByte` in `AsciiString` while I was in there.

common/src/main/java/io/netty/util/AsciiString.java
```diff
@@ -1879,6 +1851,10 @@ public static byte c2b(char c) {
         return (byte) ((c > MAX_CHAR_VALUE) ? '?' : c);
     }
 
+    public static byte c2b0(char c) {
+        return (byte) c;
```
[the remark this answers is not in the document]
Scottmitch: doh ... +1

normanmaurer:
@Scottmitch thanks.. merged

PR #7830, opened 2018-04-01
Don't check accessible in UnpooledHeapByteBuf#capacity

fenik17:
Motivation:
The `#ensureAccessible` method in `UnpooledHeapByteBuf#capacity` used
to prevent NPE if buffer is released and `array` is `null`. In all
other implementations of `ByteBuf` the accessible is not checked by
`capacity` method. We can assign an empty array to `array`
in the `deallocate` and don't worry about NPE in the `#capacity`.
This will help reduce the number of repeated calls of the
`#ensureAccessible` in many operations with `UnpooledHeapByteBuf`.

Modifications:
1. Remove `#ensureAccessible` call from `UnpooledHeapByteBuf#capacity`.
Use the `EmptyArrays#EMPTY_BYTES` instead of `null` in `#deallocate`.

2. Fix access checks in `AbstractUnsafeSwappedByteBuf` and
`AbstractByteBuf#slice` that relied on `#ensureAccessible`
in `UnpooledHeapByteBuf#capacity`. This was found by unit tests.

Result:
Less double calls of `#ensureAccessible` for `UnpooledHeapByteBuf`.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/AbstractUnsafeSwappedByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/AbstractUnsafeSwappedByteBuf.java
@@ -64,7 +64,7 @@ abstract class AbstractUnsafeSwappedByteBuf extends SwappedByteBuf {
 
     @Override
     public final int getInt(int index) {
-        wrapped.checkIndex0(index, 4);
+        wrapped.checkIndex(index, 4);
         int v = _getInt(wrapped, index);
         return nativeByteOrder ? v : Integer.reverseBytes(v);
     }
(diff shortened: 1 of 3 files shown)


fenik17:
@normanmaurer not sure, maybe we need also call `ensureAccessible` in the `slice` methods at `AbstractPooledDerivedByteBuf`?

buffer/src/main/java/io/netty/buffer/AbstractUnsafeSwappedByteBuf.java
```diff
@@ -76,21 +76,21 @@ public final int getUnsignedShort(int index) {
...
-        wrapped.checkIndex0(index, 2);
+        wrapped.checkIndex(index, 2);
         short v = _getShort(wrapped, index);
         return nativeByteOrder ? v : Short.reverseBytes(v);
     }
 
     @Override
     public final ByteBuf setShort(int index, int value) {
-        wrapped.checkIndex0(index, 2);
+        wrapped.checkIndex(index, 2);
```
normanmaurer: @fenik17 why this change ?

normanmaurer: this looks unrelated to the PR.

fenik17: Without this fixes some unit tests is failed. I described this in `Modifications` part (see the PR description). Should I extract this in a separate PR?

normanmaurer: ah got it... sorry missed it

normanmaurer:
@fenik17 yes please also make the change for `AbstractPooledDerivedByteBuf `.

fenik17:
> yes please also make the change for AbstractPooledDerivedByteBuf.

Will do.

fenik17:
@normanmaurer done: #7846

PR #7832, opened 2018-04-01
SctpMessageCompletionHandler may leak ByteBuf for fragmented messages.

normanmaurer:
Motivation:

SctpMessageCompletionHandler stores fragments in a Map but not release the stored ByteBuf when the handler is removed.

Modifications:

Release all buffers that are still in the Map when handlerRemoved(...) is called.

Result:

No more leaks for fragemented messages.

### DIFF
--- a/transport-sctp/src/main/java/io/netty/handler/codec/sctp/SctpMessageCompletionHandler.java
+++ b/transport-sctp/src/main/java/io/netty/handler/codec/sctp/SctpMessageCompletionHandler.java
@@ -68,4 +68,13 @@ public class SctpMessageCompletionHandler extends MessageToMessageDecoder<SctpMe
         }
         byteBuf.retain();
     }
+
+    @Override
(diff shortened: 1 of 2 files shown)


normanmaurer:
Will add a test soon.

LosAngelesHK:
Thanks for creating the request, Norman. By the way, I have another questions about the patch.

1. If above patch is included in Netty v4.1.23.Final, when does it release?
2. I think that I found out the root cause to trigger the memory leak. It is triggered by the continuous accepting and dropping of SCTP connections to my application (My application is acted as server). After viewing the GC log of my application, I found that the heap usage of my application is quite flat at the start of my application. After a certain period of time, the heap usage would be gradually increased until the direct buffer memory is exhausted. Why the heap usage is not gradually increased at the start of my application?

p.s. The accepting and dropping of SCTP connections are occurred continuously at the start of my application.

[Heap Usage Diagram](https://i.stack.imgur.com/3jRbv.jpg)

PR #7834, opened 2018-04-03
EPOLL - decouple schedule tasks from epoll_wait life cycle

### LINKED ISSUE #7829: Improve Timer Wakeup for Native Transports
from https://github.com/netty/netty/pull/7816#discussion_r178171275

This topic comes up in frequently enough that I usually forget the answer .... "why do we need to have a max wakeup time of 1 second instead of unbounded" for the EventLoop implementations. IIRC the answer is when we schedule tasks we don't check if the new time results in a sooner deadline than the previous wakeup on the EventLoop. I think I investigated other events from outside the EventLoop and they should be invoking `wakeup` (and we have an atomic boolean protecting wakeup to avoid too many system calls). I wonder if now that we are saving the `prevDeadlineNanos`, and we have a timer mechanism independent of `epoll_wait`,  if we can avoid waking up every second (and additional `timerfd_settime` system calls), and instead only wakeup when necessary.

Potential benefits:
- if there are no scheduled task, we only wakeup when I/O occurs
- we will be more responsive to schedule tasks (because we can adjust the wakeup to the minimum delay as new tasks are scheduled)

Potential negatives:
- in the degenerate case if someone schedules tasks in decreasing order of expiration (10 seconds from now, 9 seconds, 8 seconds, etc...) we will have to touch the the `timerfd_settime` on each schedule. For outside the EventLoop we wouldn't have to wakeup the `epoll_wait` call (assuming `timerfd_settime` is thread safe).

@normanmaurer @carl-mastrangelo - wdyt? We could do the same thing for kqueue too (we would need to use a timerfd equivalent there too though).

Scottmitch:
Motivation:
EPOLL supports decoupling the timed wakeup mechanism from the selector call. The EPOLL transport takes advantage of this in order to offer more fine grained timer resolution. However we are current calling timerfd_settime on each call to epoll_wait and this is expensive. We don't have to re-arm the timer on every call to epoll_wait and instead only have to arm the timer when a task is scheduled with an earlier expiration than any other existing scheduled task.

Modifications:
- Before scheduled tasks are added to the task queue, we determine if the new
  duration is the soonest to expire, and if so update with timerfd_settime. We
also drain all the tasks at the end of the event loop to make sure we service
any expired tasks and get an accurate next time delay.
- EpollEventLoop maintains a volatile variable which represents the next deadline to expire. This variable is modified inside the event loop thread (before calling epoll_wait) and out side the event loop thread (immediately to ensure proper wakeup time).
- Execute the task queue before the schedule task priority queue. This means we
  may delay the processing of scheduled tasks but it ensures we transfer all
pending tasks from the task queue to the scheduled priority queue to run the
soonest to expire scheduled task first.
- Deprecate IORatio on EpollEventLoop, and drain the executor and scheduled queue on each event loop wakeup. Coupling the amount of time we are allowed to drain the executor queue to a proportion of time we process inbound IO may lead to unbounded queue sizes and unpredictable latency.

Result:
Fixes #7829
- In most cases this results in less calls to timerfd_settime
- Less event loop wakeups just to check for scheduled tasks executed outside the event loop
- More predictable executor queue and scheduled task queue draining
- More accurate and responsive scheduled task execution


Scottmitch:
more testing / analysis is required, but figured I would push this upstream to see if everyone is on board.

@carl-mastrangelo - would you mind running the same benchmarks you ran for PR https://github.com/netty/netty/pull/7816?

common/src/main/java/io/netty/util/concurrent/ScheduledTask.java
```diff
@@ -0,0 +1,29 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.util.concurrent;
+
+/**
+ * Represents a task that is scheduled to expire some time in the future.
+ */
+public interface ScheduledTask {
```
normanmaurer: Can't we just use `Delayed` ?

https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Delayed.html

transport/src/main/java/io/netty/channel/SingleThreadEventLoop.java
```diff
@@ -31,7 +31,6 @@
  *
  */
 public abstract class SingleThreadEventLoop extends SingleThreadEventExecutor implements EventLoop {
-
```
normanmaurer: revert

transport-native-epoll/src/main/java/io/netty/channel/epoll/Native.java
```diff
@@ -102,23 +102,23 @@ public static FileDescriptor newTimerFd() {
     public static native void eventFdWrite(int fd, long value);
     public static native void eventFdRead(int fd);
     static native void timerFdRead(int fd);
+    static native void timerFdSetTime(int fd, int sec, int nsec);
```
normanmaurer: This should have a `throws` as it may throw a `ClosedChannelException`

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
```diff
@@ -46,15 +48,24 @@
...
         Epoll.ensureAvailability();
     }
 
+    private final Queue<ScheduledFutureTaskRunnable> scheduledQueue;
+    /**
+     * Note that we use deadline instead of delay because deadline is just a fixed number but delay requires interacting
+     * with the time source (e.g. calling System.nanoTime()) which can be expensive.
+     */
+    private final AtomicLong nextDeadlineNanos = new AtomicLong(MAXIMUM_DEADLINE);
+    private final AtomicInteger wakenUp = new AtomicInteger();
```
normanmaurer: Just use an `AtomicBoolean` as we use it as a boolean.

normanmaurer:
@Scottmitch Thanks for this... in general this looks ok. That said I would be interested to see how it compares to  #7816 and if it worth it compared to the "easier" solution that was used in  #7816

common/src/main/java/io/netty/util/concurrent/ScheduledTask.java
Scottmitch: We want to get the `deadline`, not the `delay`.

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
Scottmitch: I considered this, however it involves additional conditional behind the scenes. Since this is internal I think we can just use `AtomicInteger`.

I also moved away from the AtomicFieldUpdater because:
- the excepted number of instances for this class is low
- the atomic field updaters have an additional "instance check" which can be avoided.

transport-native-epoll/src/main/java/io/netty/channel/epoll/Native.java
Scottmitch: However this is a runtime exception ... would you prefer a `IOException` using the `errno` like in `epollWait`?

Scottmitch:
Agreed more analysis is necessary. This PR does a lot of things PR #7816 doesn't do, so lets see if I can quantify it some how :)

transport-native-epoll/src/main/java/io/netty/channel/epoll/Native.java
normanmaurer: yeah I think this would be better (using NativeIOException)

common/src/main/java/io/netty/util/concurrent/ScheduledTask.java
normanmaurer: got it

normanmaurer: @Scottmitch that said should we just make this part of the Abstract base class and mark as unstable ?

transport-native-epoll/src/main/java/io/netty/channel/epoll/Native.java
Scottmitch: I just went with `IOException` for now ... avoids additional conditional checking and the string description of the error is already included in the `IoExecution` thrown by the JNI code.

common/src/main/java/io/netty/util/concurrent/ScheduledTask.java
Scottmitch: What abstract base class? The issue is we need it to be accessible outside the package (`PromiseTask` and `ScheduledFutureTask` are package private). I considered putting a protected interface in `AbstractScheduledEventExecutor` and can do this to reduce the exposure of this API. I will push this change and we can discuss more if necessary.

On a related note ... the relationship between `ScheduledFutureTask` and `AbstractScheduledEventExecutor` is very tightly coupled. I would like to tease these two apart in general as there a assumptions needed about using the "correct" `nanoTime` that may be made less "leaky" for the next major release (I added something to the doc).

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
```diff
@@ -385,6 +446,26 @@ protected final boolean runAllTasksFrom(Queue<Runnable> taskQueue) {
         }
     }
 
+    /**
+     * What ever tasks are present in {@code taskQueue} when this method is invoked will be {@link Runnable#run()}.
+     * @param taskQueue the task queue to drain.
+     * @return {@code true} if at least {@link Runnable#run()} was called.
+     */
+    private boolean runExistingTasksFrom(Queue<Runnable> taskQueue) {
+        taskQueue.offer(BOOKEND_TASK);
```
carl-mastrangelo: I think this would be simpler if you called `size()` up front, and just did a normal for loop until that many tasks were executed.  If taskQueue is full, you run the risk of getting an NPE later on.

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
```diff
@@ -71,16 +82,15 @@ public int get() throws Exception {
...
             return EpollEventLoop.super.pendingTasks();
         }
     };
-    private volatile int wakenUp;
-    private volatile int ioRatio = 50;
 
     EpollEventLoop(EventLoopGroup parent, Executor executor, int maxEvents,
                    SelectStrategy strategy, RejectedExecutionHandler rejectedExecutionHandler) {
         super(parent, executor, false, DEFAULT_MAX_PENDING_TASKS, rejectedExecutionHandler);
+        scheduledQueue = PlatformDependent.newMpscQueue();
```
carl-mastrangelo: This is concerning.  I would have expected this to be a non threadsafe queue.   If it is threadsafe, that means that it could be concurrently modified.  That means that there is race in the code above, where calling `peek` and then `poll` could return different tasks.

common/src/main/java/io/netty/util/concurrent/ScheduledFutureTask.java
```diff
@@ -78,12 +83,17 @@ protected EventExecutor executor() {
...
         return deadlineNanos;
     }
 
     public long delayNanos() {
-        return Math.max(0, deadlineNanos() - nanoTime());
+        return deadlineToDelayNanos(deadlineNanos());
+    }
+
+    static long deadlineToDelayNanos(long deadlineNanos) {
+        return Math.max(0, deadlineNanos - nanoTime());
```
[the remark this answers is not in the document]
Scottmitch: I agree this could likely be changed, but is pre-existing behavior and doesn't impact this change. Note that we are sorting, and determining "next to expire", based upon deadline (not delay).

Lets discuss in a followup issue: https://github.com/netty/netty/issues/7840

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
Scottmitch: This is an unbounded MPSC queue and size computation requires an iteration. I think insertion/removal should be a preferred approach here. We can make the assumptions about unbounded queue more clear or be more defensive if that helps.

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
Scottmitch: >  That means that there is race in the code above, where calling peek and then poll could return different tasks.

This is operating on a different queue. Here are the queues in play:
- `SingleThreadEventExecutor#taskQueue` - MPSC used for calls to `execute`. This is also currently used by all other transports to put process `schedule` calls on the event loop thread, but this PR changes that behavior for EPOLL.
- `AbstractScheduledEventExecutor#scheduledTaskQueue` - priority queue of scheduled `Runnable`s sorted by deadline. This is NOT thread safe and only touched on the event loop thread.
- `EpollEventLoop#scheduleQueue` - MPSC used to hold `schedule` calls outside of the event loop thread. This is drained when the event loop thread wakes up, and the `Runnable` is appended to `scheduledTaskQueue` (described above).

The goal is to decouple the scheduled task queue (`schedule` calls from outside the event loop) from the executor tasks. This allows us to drain all of the scheduled tasks to get an accurate "next expire time" before calling `epoll_wait`. In order to ensure we wakeup from outside the event loop we have `nextDeadlineNanos`, which is atomic and indicates the "next deadline", and used to know when we should call `setTimerFd` from outside (and inside) the event loop.

The class hierarchy and spread of queues is unfortunate, but since EPOLL is the only transport to decouple the `schedule` and `execute` calls I figured best to add `scheduledQueue` here for now.

Scottmitch: Let me add a comment describing this :)

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
Scottmitch: Actually the queue is unbounded by default, but the user may specify a limit ... let me make this more robust.

Scottmitch:
@carl-mastrangelo - thanks for review! would you be able to run the same benchmark you did in PR #7816? I'm also planning on doing some analysis, but the more data points the better.

trustin:
Sorry I didn't have a chance to review your PR yet, @Scottmitch. Will find some time soon.

M-AJ:
How can we get this artifact jar or artificatId for a maven dependency?

normanmaurer:
@M-AJ I can build one for you if you are interested otherwise you would just apply the patch and build it yourself.

M-AJ:
@normanmaurer If you can build one, that'd be great! Really appreciate it! I've tried building it locally but the process takes a while for me.

Scottmitch:
I just rebased to make getting recent fixes easier ... sorry super busy at the moment ... this is still on the backlog and hopefully will get to this soon. However it would be much appreciated if others could also try this out though 🤒

normanmaurer:
@M-AJ you want a netty-all jar or just the epoll jar ?

normanmaurer:
Sorry it took me some time :( 

Try this one:
https://drive.google.com/open?id=1S4Ro2xJu0Mf2NepbaFR4zM75wKXdMhDx

normanmaurer:
@M-AJ ^^

M-AJ:
Thanks man, I'm going to be running some benchmarks. Will report how well it goes.

normanmaurer:
@M-AJ any update ?

M-AJ:
@normanmaurer Our server was running version 4.0.36.Final; therefore, I had to update our pipeline handlers, attribute keys and bytebuf wrappers to this new version. Unfortunately, after successful compilation, the server did not receive any packets after the upgrade. I'm planning on fixing the issue and bench marking early June when I finish these last few weeks of school.

transport-native-epoll/src/test/java/io/netty/channel/epoll/EpollTest.java
```diff
@@ -1,72 +0,0 @@
...
-            final FileDescriptor eventfd = Native.newEventFd();
-            Native.epollCtlAdd(epoll.intValue(), timerFd.intValue(), Native.EPOLLIN);
-            Native.epollCtlAdd(epoll.intValue(), eventfd.intValue(), Native.EPOLLIN);
-
-            final AtomicReference<Throwable> ref = new AtomicReference<Throwable>();
-            Thread t = new Thread(new Runnable() {
-                @Override
-                public void run() {
-                    try {
-                        assertEquals(1, Native.epollWait(epoll, eventArray, timerFd, -1, -1));
```
Scottmitch: epollWait no longer supports the `-1` special value

transport-native-epoll/src/main/c/netty_epoll_native.c
```diff
@@ -169,6 +169,31 @@ static jint netty_epoll_native_epollCreate(JNIEnv* env, jclass clazz) {
     return efd;
 }
 
+static void netty_epoll_native_timerFdSetTime(JNIEnv* env, jclass clazz, jint timerFd, jint tvSec, jint tvNsec) {
+    struct itimerspec ts;
+    memset(&ts.it_interval, 0, sizeof(struct timespec));
```
normanmaurer: nit: do you even need to meet as you assign all fields anyway

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
```diff
@@ -160,9 +162,112 @@ NativeDatagramPacketArray cleanDatagramPacketArray() {
...
+        //    when it is woken up and before it calls epoll_wait again and adjusts the timer accordingly.
+        // The result is we wait until we are in the EventLoop and doing the actual removal, and also processing
+        // regular polling in the EventLoop too.
+
+        execute(runnable);
+    }
+
+    @Override
+    protected boolean wakesUpForTask(Runnable task) {
+        return super.wakesUpForTask(task) || !(task instanceof ScheduledTaskRunnable);
```
normanmaurer: @Scottmitch maybe I am missing something but is this really correct ? I thought it should be:

```
return !(task instanceof ScheduledTaskRunnable) && super.wakesUpForTask(task);
```

At least from what I understand we should never need to wakeup for a `ScheduledTaskRunnable` as the timerfd will take care of it.

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
```diff
@@ -160,9 +162,112 @@ NativeDatagramPacketArray cleanDatagramPacketArray() {
...
+        nextDeadlineNanos.set(MAXIMUM_DEADLINE);
+
+        // Note that we don't actually call setTimerFd here, we don't want to interrupt the actual timerFd and let
+        // the end of the EventLoop determine what the timerFd value should be (after execute queue is drained).
+    }
+
     @Override
     protected void wakeup(boolean inEventLoop) {
-        if (!inEventLoop && WAKEN_UP_UPDATER.compareAndSet(this, 0, 1)) {
+        if (!inEventLoop && wakenUp.getAndSet(1) == 0) {
```
normanmaurer: why this change ? Because it may be able to use XADD ?

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
Scottmitch: Yes you are correct ... the logic here is flawed 🤦‍♂️

transport-native-epoll/src/main/c/netty_epoll_native.c
Scottmitch: the `it_interval` is not touched below so it is zero'ed out (we don't want a repeating timer).

```c
           struct timespec {
               time_t tv_sec;                /* Seconds */
               long   tv_nsec;               /* Nanoseconds */
           };

           struct itimerspec {
               struct timespec it_interval;  /* Timer interval */
               struct timespec it_value;     /* Initial expiration */
           };
```

normanmaurer: doh! missed this... good point

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
Scottmitch: Yes `XADD` also introduced a `AtomicLong` for `nextDeadlineNanos` so I switched this to an `AtomicInteger` because we don't expect many `EventLoop` instances and can avoid the additional checks done by the reference updater. I can revert this portion of the change if you prefer as it isn't critical for this PR.

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
```diff
@@ -386,6 +444,46 @@ protected final boolean runAllTasksFrom(Queue<Runnable> taskQueue) {
...
+    private boolean runExistingTasksUntilMaxTasks(Queue<Runnable> taskQueue) {
+        Runnable task = pollTaskFrom(taskQueue);
+        if (task == null) {
+            return false;
+        }
+        int i = 0;
+        do {
+            safeExecute(task);
+            task = pollTaskFrom(taskQueue);
+            if (task == null) {
```
normanmaurer: Should we also respect `BOOKEND_TASK ` here, or do we not expect to have one in here as we failed to add it before calling this and only will ever have one in the queue ?

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
```diff
@@ -386,6 +444,46 @@ protected final boolean runAllTasksFrom(Queue<Runnable> taskQueue) {
...
+
+    private boolean runExistingTasksUntilBookend(Queue<Runnable> taskQueue) {
+        Runnable task = pollTaskFrom(taskQueue);
+        if (task == BOOKEND_TASK) {
+            return false;
+        }
+        for (;;) {
+            safeExecute(task);
+            task = pollTaskFrom(taskQueue);
+            if (task == BOOKEND_TASK) {
```
normanmaurer: I am correct to say that we should never have `task == null` here and above ?

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
normanmaurer: no its fine... just was curious .

normanmaurer:
@Scottmitch also can you please update the description as it still contains the details for your "old approach"

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
Scottmitch: Right. `BOOKEND_TASK` shouldn't be encountered here as you indicated.

normanmaurer: maybe we should add an `assert` ?

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
Scottmitch: I wouldn't expect this because we don't allow `null` to be offered to the queue, and we only enter this method if we successfully added `BOOKEND_TASK` to the queue. Am I missing something?

normanmaurer: sounds about right... Just wanted to ensure I read it correctly. Maybe an assert would clarify ?

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
Scottmitch: We can if you prefer however if we don't nothing negative will happen.

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
Scottmitch: >  Maybe an assert would clarify ?

We can if you prefer, however if `null` does come up we will NPE and log an error.

Scottmitch:
@normanmaurer - I updated the commit message. let me know if it makes more sense now.

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
normanmaurer: Let us just add a comment... can you do it ?

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
normanmaurer: Can you just add a comment for it ?

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
Scottmitch: done

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
Scottmitch: done

common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java
```diff
@@ -85,6 +102,15 @@ protected void cancelScheduledTasks() {
         }
 
         scheduledTaskQueue.clearIgnoringIndexes();
+
+        // calling minimumDelayScheduledTaskRemoved was considered here to give a chance for EventLoop
+        // implementations the opportunity to cleanup any timerFds, but this is currently only called when the EventLoop
+        // is being shutdown, so the timerFd and associated polling mechanism will be destroyed anyways.
+    }
+
+    protected final ScheduledTask peekScheduledTaskDelayNanos() {
```
trustin: Missing Javadoc?

common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java
```diff
@@ -225,11 +256,46 @@ protected void validateScheduled(long amount, TimeUnit unit) {
...
-    <V> ScheduledFuture<V> schedule(final ScheduledFutureTask<V> task) {
+    /**
+     * Represents a task that is scheduled to expire some time in the future.
+     */
+    protected interface ScheduledTask {
+        /**
+         * Get the time stamp relative to some monotonically increasing time source for which this {@link ScheduledTask}
+         * expires.
+         * @return the time stamp relative to some monotonically increasing time source for which this
+         * {@link ScheduledTask} expires.
```
trustin: nit: indentation

common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java
```diff
@@ -225,11 +256,46 @@ protected void validateScheduled(long amount, TimeUnit unit) {
...
+    }
+
+    /**
+     * Represents a call to one of the {@link #schedule(Runnable, long, TimeUnit)} method variants from outside of this
+     * event executor thread (e.g. {@link #inEventLoop()} returns {@code false}).
+     */
+    protected interface ScheduledTaskRunnable extends Runnable, ScheduledTask {
+        /**
+         * Determine if this object is the result of adding or removing a scheduled task.
+         * @return {@code true} if this object is the result of adding or removing a scheduled task.
```
trustin: Doesn't this method return `false` on removal?

common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java
```diff
@@ -242,14 +308,51 @@ public void run() {
...
+            scheduledTaskQueue.removeTyped(task);
+        }
+    }
+
+    /**
+     * Execute a {@link ScheduledTaskRunnable} from outside the event loop thread. Note that schedule events which
+     * occur on the event loop thread do not interact with this method.
+     * @param runnable The {@link ScheduledTaskRunnable} to execute.
+     */
+    protected void executeScheduledRunnable(ScheduledTaskRunnable runnable) {
```
trustin: I'm curious if the parameter type of this method can be widen to `ScheduledTask` or even `Runnable`. If so, we could hide `ScheduledTaskRunnable` (and even `ScheduledTask`) completely from the public API, i.e. less maintenance burden.

If a subclass needs the information like deadline, we could just add it as an additional parameter.

common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java
```diff
@@ -242,14 +308,51 @@ public void run() {
...
+        execute(runnable);
+    }
+
+    /**
+     * The next task to expire (e.g. minimum delay) has been removed from the scheduled priority queue.
+     */
+    protected void minimumDelayScheduledTaskRemoved(@SuppressWarnings("unused") ScheduledTask task) {
```
trustin: Ditto - can `ScheduledTask` be widen to `Runnable`?

normanmaurer:
@netty-bot test this please

normanmaurer:
This is sitting here for a long time now... Let me merge this.

normanmaurer:
Want to do a PR to show the ideas ?

njhill:
@normanmaurer sure, stand by :)

njhill:
@normanmaurer see #9470

common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java
```diff
@@ -99,16 +120,21 @@ protected final Runnable pollScheduledTask() {
...
+
+    final Runnable pollScheduledTask(Queue<ScheduledFutureTask<?>> scheduledTaskQueue, long nanoTime,
+                                     boolean notifyMinimumDeadlineRemoved) {
+        assert scheduledTaskQueue != null;
+        assert inEventLoop();
 
-        if (scheduledTask.deadlineNanos() <= nanoTime) {
-            scheduledTaskQueue.remove();
+        ScheduledFutureTask<?> scheduledTask = scheduledTaskQueue.peek();
+        if (scheduledTask != null && scheduledTask.deadlineNanos() <= nanoTime) {
```
franz1981: Replace direct comparison using nanoTime with differences ie ```scheduledTask.deadlineNanos() - nanoTime <= 0```

Scottmitch: Netty's [ScheduledFutureTask offsets all time stamps](https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/concurrent/ScheduledFutureTask.java#L34) and they are generally compared using equality for deadlines while giving up half of the domain (e.g. overflow earlier). I agree it is confusing and we may want to revisit how time duration/expiration is done in Netty 5 in this area.

transport-native-epoll/src/main/c/netty_epoll_native.c
```diff
@@ -186,6 +186,31 @@ static jint netty_epoll_native_epollCreate(JNIEnv* env, jclass clazz) {
     return efd;
 }
 
+static void netty_epoll_native_timerFdSetTime(JNIEnv* env, jclass clazz, jint timerFd, jint tvSec, jint tvNsec) {
```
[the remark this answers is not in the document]
Scottmitch: No I didn't consider this.

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
```diff
@@ -397,6 +455,55 @@ protected final boolean runAllTasksFrom(Queue<Runnable> taskQueue) {
         }
     }
 
+    /**
+     * What ever tasks are present in {@code taskQueue} when this method is invoked will be {@link Runnable#run()}.
+     * @param taskQueue the task queue to drain.
+     * @return {@code true} if at least {@link Runnable#run()} was called.
+     */
+    private boolean runExistingTasksFrom(Queue<Runnable> taskQueue) {
+        return taskQueue.offer(BOOKEND_TASK) ? runExistingTasksUntilBookend(taskQueue)
```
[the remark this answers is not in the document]
Scottmitch: good idea! calling from consumer thread for MPSC shouldn't require looping.

common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java
njhill: "earlier" being after only 300 years, rather than 600 :)

common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java
Scottmitch: The way this suggestion was implemented makes things more complex IMHO. We are trying to communicate from here that the type of task is for scheduled execution, and this information is used conditionally at higher layers (e.g. should we wake up). This has been changed to introduce 3 methods (e.g. `beforeScheduledTaskSubmitted`, `afterScheduledTaskSubmitted`, `wakesUpForScheduledRunnable`) and 1 additional type is exposed `NonWakeupRunnable` publicly which may require conditional allocation/wrapping at higher layers. IOW we are losing information by just allocating a `Runnable` here and have to compensate by complicating method signatures and type hierarchy at other layers.

I'll submit a followup PR (since this plus an additional PR was recently reverted) so we can take a look. /cc @normanmaurer @njhill

njhill: Thanks @Scottmitch. I gave this quite a lot of thought so would be good to discuss more. A few initial responses though:
- The prior communication depended on the notion of an intermediate task which happens to involve enqueing some other scheduled task onto the internal scheduled-task queue (i.e. not the scheduled task itself), which is something we should not be exposing at all imho... that seems like an internal detail which EL impls shouldn't be aware of.
- I don't see `wakesUpForScheduledRunnable` that you referenced, do you mean `wakesUpForRunnable` or `executeScheduledRunnable`?
- The only reason I left `wakesUpForRunnable` there is that it's protected and so would be a breaking change to remove it, I'd suggest that we deprecate it. A public `NonWakeupRunnable` interface serves a similar purpose but makes much more sense imho. Whether a given task is required to be run immediately is in general a property of that task and should be independent of the executor/EL implementation. We could consider a `lazyExecute(Runnable)` method in addition to this (might be be nice to have on the `EventExecutor` interface in netty 5).
- I actually think we should remove `executeScheduledRunnable` altogether (possible since it's package-private), and instead just override `schedule(...)` in `SingleThreadEventExecutor`. This would further simplify things and avoid the double `Runnable` wrapping. Not sure it's also what you had in mind but it's a small change, I'll also push an example of that
- Only the changes to `EpollEventLoop` itself were reverted for the release, the reworked superclass abstractions are still there. I'm still very interested to find out what the issues were ... @normanmaurer any luck with a reproducer or more clues? :)

njhill: @Scottmitch I made the changes alluded to above here 3923176afebefe58bf85155d25ed10962f9dd660 but took things a bit further still, uplifting the before/after hooks to `AbstractScheduledEventExecutor` and introducing higher level idea of `lazyExecute` / `LazyRunnable` which could have general applicability.

I really think this is both simpler to exploit and more generally useful for subclasses (i.e. event loop impls).

Issue #7837, opened 2018-04-03
CorsHandler do writeAndFlush in override ChannelHandlerContext#write method


amizurov:
Do not really understand, why in `CorsHandler` invoked writeAndFlash ? It's strange side effect.
### Actual behavior

https://github.com/netty/netty/blob/473f6a6edb8353728ea8e711ac8241ddc21adf08/codec-http/src/main/java/io/netty/handler/codec/http/cors/CorsHandler.java#L198-L209
### Steps to reproduce
Add `CorsHandler` to pipeline then write response in your own handler, response flushed automatically.

### Netty version
4.1.22

normanmaurer:
@amizurov its a bug :(
 
Fixed https://github.com/netty/netty/pull/7839

amizurov:
@normanmaurer Thanks a lot, It's really quickly as well as Netty :)

PR #7839, opened 2018-04-03
CorsHandler.write(...) should not cause a flush.

normanmaurer:
Motivation:

CorsHandler.write(...) should not cause a flush but just trigger a write(...)

Modifications:

Replace writeAndFlush(...) with write(...)

Result:

Fixes https://github.com/netty/netty/issues/7837.


ejona86:
Note, the motivation is the commit description isn't all that helpful. Maybe:
> Unnecessary flushes reduce the amount of flush coalescing that can happen at higher levels and thus can increase number of packets (because of TCP_NODELAY) and lower throughput (due to syscalls, TLS frames, etc).

normanmaurer:
@ejona86 amend... thanks

PR #7841, opened 2018-04-04
add task before starting thread in SingleThreadEventExecutor.execute

davidmoten:
Motivation:

Minor performance optimisation that prevents thread from blocking due to task not having been added to queue. Discussed #7815.

Modification:

add task to the queue before starting the thread.

Result:

No additional tests.


normanmaurer:
@davidmoten can you sign our icily and let me know once done ?

https://netty.io/s/icla

davidmoten:
@normanmaurer signed

normanmaurer:
Thanks!

Issue #7814, opened 2018-03-28
IllegalReferenceCountException in HttpPostMultipartRequestDecoder.


ExIngus:
### Netty version 
4.1.22.Final

### JVM version (e.g. `java -version`) 
1.8.0_162

### OS version (e.g. `uname -a`)
windows 10

The error is observed in versions 4.1.20-4.1.22

```
HttpPostRequestDecoder decoder = new HttpPostRequestDecoder(request);
try {
...
} finally {
	decoder.destroy();
}
```
When you call decoder.destroy () MixedAttribute.release () for objects is performed 2 times.

[PostTest.zip](https://github.com/netty/netty/files/1855660/PostTest.zip)

```
io.netty.util.IllegalReferenceCountException: refCnt: 0, increment: 1
    	at io.netty.util.AbstractReferenceCounted.release0(AbstractReferenceCounted.java:87)
    	at io.netty.util.AbstractReferenceCounted.release(AbstractReferenceCounted.java:71)
    	at io.netty.handler.codec.http.multipart.MixedAttribute.release(MixedAttribute.java:320)
    	at io.netty.handler.codec.http.multipart.HttpPostMultipartRequestDecoder.destroy(HttpPostMultipartRequestDecoder.java:947)
    	at io.netty.handler.codec.http.multipart.HttpPostRequestDecoder.destroy(HttpPostRequestDecoder.java:247)
```

doom369:
https://github.com/netty/netty/pull/7696

ExIngus:
I do not call manually HttpData.release ().

doom369:
Are you sure? Please post the full sketch. Or minimal reproducer.

ExIngus:
Test in error in the attached archive to the first message. (PostTest.zip)

fishjam:
Maybe I found the reason, please confirm it.
- Reason:
   - same MixedAttribute instance saved both in DefaultHttpDataFactory.requestFileDeleteMap and HttpPostMultipartRequestDecoder.bodyListHttpData
   - when call HttpPostMultipartRequestDecoder.destroy,  it will release data in requestFileDeleteMap by cleanFiles(), and release data in bodyListHttpData.
   - call stack:
```
	at io.netty.handler.codec.http.multipart.DefaultHttpDataFactory.createAttribute(DefaultHttpDataFactory.java:141)   <== the Attribute will save in list
[...]

PR #7843, opened 2018-04-05
fix IllegalReferenceCountException in HttpPostStandardRequestDecoder

### LINKED ISSUE #7689: 4.1.20 HttpPostMultipartRequestDecoder destroy duplicate

fishjam:
## avoid duplicate release in http request decoder

- Motivation:
   - According my debug and record in #7814, same MixedAttribute instance saved both in DefaultHttpDataFactory.requestFileDeleteMap and HttpPostMultipartRequestDecoder.bodyListHttpData, After user call HttpPostStandardRequestDecoder.next() or  HttpPostMultipartRequestDecoder.next() to get and release the data, same instance remain in 
HttpDataFactory, so call removeHttpDataFromClean to remove it.

- Modification:
  - remove released data from  factory
  - add null check in MemoryAttribute to avoid NullException while I debug this issue.

- Result:
  - Fixes #7689 #7695  #7814

- Additional
  - I think it's bad design that require user to release the data after they call HttpPostStandardRequestDecoder.next(), But since it write in the comments and lots of people do it, so I call removeHttpDataFromClean.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java
@@ -380,7 +380,9 @@ public class HttpPostMultipartRequestDecoder implements InterfaceHttpPostRequest
         checkDestroyed();
 
         if (hasNext()) {
-            return bodyListHttpData.get(bodyListHttpDataRank++);
+            InterfaceHttpData data = bodyListHttpData.get(bodyListHttpDataRank++);
+            factory.removeHttpDataFromClean(request, data);
+            return data;
         }
         return null;
     }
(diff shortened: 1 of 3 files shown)


fishjam:
Is there anybody can confirm this PR?
@doom369 @normanmaurer 

It seems that Reference Count in netty is too complex and nobody can make sure that there is no memory leak or duplicpate release.  :(

fishjam:
Although there is ResourceLeakDetector in netty, but it reley on  GC, and can not handle all the resource handleing. So when I debug these issue, I write a class ResourceLeakHelper to track all the operation. I think it can be enabled when level is PARANOID or add new level.
I add my test code, please confirm it. If it's OK, then I will try to refactor it and create PR.

[NettyLeakTest.zip](https://github.com/netty/netty/files/1909907/NettyLeakTest.zip)

fishjam:
@DATuan91 I'm not sure that it can fix [Unexpected IllegalReferenceCountException on decode multipart request](https://github.com/netty/netty/issues/8575) because I  have never checked the issue.

But as I mentioned in the [comments](https://github.com/netty/netty/pull/7843#issuecomment-381294653), I have written ResourceLeakHelper class to track all the ref count operation, and then can find the root reason of IllegalReferenceCountException.

PR #7844, opened 2018-04-05
Fixes NPE in ClientCookieDecoder

fenik17:
Motivation:
NPE occurs in `ClientCookieDecoder` if a cookie starts with comma.

Modifications:
Check `cookieBuilder` for `null` in the return.

Result:
No fails NPE on invalid cookies.


normanmaurer:
@fenik17 thx

PR #7846, opened 2018-04-05
Checks accessibility in the #slice and #duplicate methods of ByteBuf

fenik17:
Motivation:
The `ByteBuf#slice` and `ByteBuf#duplicate` methods should check an accessibility to prevent creation slice or duplicate of released buffer. At now this works not in the all scenarios.

Modifications:
Add missed checks.

Result:
More correct and consistent behavior of `ByteBuf` methods.


sahujaunpuri:
Has anyone done it in an app that creates a nifty jar in  drag & drop?

normanmaurer:
@fenik17 thanks!

Issue #7657, opened 2018-01-30
upgrade requests should support streaming


mosesn:
### Expected behavior

Upgrade requests and responses should continue propagating chunks immediately.

### Actual behavior

Http{Server,Client}UpgradeHandler both extend HttpObjectAggregator, and aren't streamed properly.

### Considerations

h2c is the only protocol I know which can have a request body in the upgrade request, so I'm going to target a solution to this.  The difficulty is that in the upgrade, we typically have an HTTP/1.1 request, and then respond with a 101, and then reply over HTTP2.  Right now, the way it works is that we have an HTTP/1.1 server handler, we process the entire request through that, then send it to the HttpServerUpgradeHandler.  The HttpServerUpgradeHandler aggregates the entire request, replies with a 101, removes the HTTP/1.1 server handler, and then translates the HTTP/1.1 request to an HTTP/2 request and adds the HTTP/2 server handler to the pipeline.

We should make a couple small modifications here.  Instead of aggregating the entire request, the HttpServerUpgradeHandler should continuously translate messages for the server, and should simply mark the connection as upgrading.  As soon as the server receives the headers, it should send the 101, and then it should immediately start sending HTTP/2 responses.  However, this is complicated slightly by the HTTP/1.1 server, which is still in the pipeline.  I *think* it should ignore messages it doesn't know how to handle, so it should be OK, but is worth checking.

After the request has finished streaming, we should remove the HTTP/1.1 server codec and the HttpServerUpgradeHandler from the pipeline.

I haven't looked at the HttpClientUpgradeHandler as closely, but it looks like it should be pretty similar.  Does this seems reasonable @normanmaurer @Scottmitch?

mosesn:
By the way, on the client-side the responses are actually OK, because it only aggregates the 101 response.  It should only be requests that we're worrying about.

normanmaurer:
@mosesn this sound reasonable to me... @carl-mastrangelo @ejona86 WDYT ?

ejona86:
How is it that _responses_ can't be streamed?

Streaming the upgrade _request_ is not great, because the client's request is sent in HTTP/1's format and thus there's no multiplexing. This is mentioned in [§3.2 of RFC 7540](http://httpwg.org/specs/rfc7540.html#rfc.section.3.2):
> Requests that contain a payload body MUST be sent in their entirety before the client can send HTTP/2 frames. This means that a large request can block the use of the connection until it is completely sent.
>
[...]

PR #7848, opened 2018-04-05
Set the correct stream header state for h2c upgrades

### LINKED ISSUE #7847: Http2ConnectionHandler doesn't set the correct header state for streams generated from upgrades
### Expected behavior

When stream 1 is created from a h2c upgrade, the resultant stream should signal that it has sent the headers for the client, and that it has received the headers for the server.

### Actual behavior

The stream state doesn't acknowledge that the request headers have been sent/received.

Add the following test cases to the Http2ConnectionHandlerTest suite:

```java
    @Test
    public void clientUpgradeSetsStreamHeadersSent() throws Exception {
        when(local.createStream(STREAM_ID, true)).thenReturn(stream);
        handler = newHandler();
        handler.onHttpClientUpgrade();
        Http2Stream upgradeStream = handler.connection().stream(HTTP_UPGRADE_STREAM_ID);
        assertSame(stream, upgradeStream);
        verify(stream).headersSent(false);
    }

    @Test
    public void serverUpgradeSetsStreamHeadersReceived() throws Exception {
        when(connection.isServer()).thenReturn(true);
        when(remote.createStream(STREAM_ID, true)).thenReturn(stream);
        handler = newHandler();
        handler.onHttpServerUpgrade(new Http2Settings());
        Http2Stream upgradeStream = handler.connection().stream(HTTP_UPGRADE_STREAM_ID);
        assertSame(stream, upgradeStream);
        verify(stream).headersReceived(false);
    }
```

### Netty version

4.1 branch.

openjdk version "1.8.0_162"

OSX Darwin Kernel Version 17.5.0

bryce-anderson:
Motivation:
When a H2 session is established due to a h2c upgrade, the initial
request is sent via HTTP/1.1. When the h2 connection is created
from this pathway, the resultant session should acknowledge that
the upgrade stream has already had it's request headers sent.

Modification:
The `Http2ConnectionHandler.onHttpClientUpgrade` and
`Http2ConnectionHandler.onHttpServerUpgrade` methods set the 
correct state to reflect the received/sent request headers.

Result:
Fixes #7847.


bryce-anderson:
This seemed easy enough to change, so I just went ahead and did it. If #7847 makes incorrect assumptions fft close this.

ejona86:
CC @zpencer

ejona86:
So I'd agree this is closer, but it seems incomplete. In the GET case, won't the client have already half-closed? That doesn't seem handled here. But when doing a streaming POST the client won't have half-closed, and then it's not clear how the stream state is updated.

Could you expand on your motivation some? Did this break your application, or did confuse some part of the rest of the HTTP/2 stack? I'm sort of assuming for now that it just broke your application.

bryce-anderson:
@ejona86, this change only concerns the h2c upgrade, so my following comments should be taken in that context.

For upgrade requests, Stream 1 is implicitly "half-closed" from the client toward the server, so for all requests, not just GET requests, it's half closed. The current upgrade mechanisms (HttpServerUpgradeHandler) don't facilitate streaming an upgrade request so this doesn't attempt to address the complexity associated with that, though even in that case it would have still received the headers.

The state I'm trying to rectify is the [HTTP Request/Response Exchange](https://http2.github.io/http2-spec/#rfc.section.8.1). For the upgrade request, the client-to-server half of the stream is already complete, so the headers have been sent by the client and received by the server when we call the respective upgrade methods. This is mostly just bookkeeping, but it is important for the `onStreamError` method which calls to the `resetStream`. Without this change, a problem with the upgrade request can cause a client to omit a `RST` frame [here](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java#L768) since it thinks it hasn't sent the request at all.

Why this seems weird is that the stream state and request/response state are encoded in two orthogonal ways (`State state` and `byte metaState`, respectively) even though there is a bit of coupling between them from the protocols perspective. For example, a half-closed stream will always have completed the request headers.

ejona86:
I misunderstood the context of this bug, and forgot the stream was already half-closed. That made me mix things together with #7657. Sorry.

> Without this change, a problem with the upgrade request can cause a client to omit a RST frame here since it thinks it hasn't sent the request at all.

Ah, excellent. That sort of thing should be mentioned in "motivation".

> Why this seems weird is that the stream state and request/response state are encoded in two orthogonal ways (State state and byte metaState, respectively) even though there is a bit of coupling between them from the protocols perspective. For example, a half-closed stream will always have completed the request headers.

This is good "motivation" as well. And it points out that I don't think this fix is in the correct spot. It seems this is a bug within DefaultHttp2Connection.java. The `createStream(_, /*halfClosed =*/ true)` should not produce a stream in such a broken state. Either that method or `DefaultStream`'s constructor seem more fitting places for the fix.

bryce-anderson:
@ejona86, I'll see about making the `halfClosed = true` case produce a correct state. Thanks for the feedback.

bryce-anderson:
@ejona86, ptal. I'm not sure it's right: there are a lot of ways to do a state transition with a stream, so let me know if you see a problem.

normanmaurer:
@ejona86 can you check again please?

Issue #7825, opened 2018-03-30
HpackEncoder limits header list size to 8K by default


ejona86:
### Expected behavior
Sending a header list of arbitrary size would succeed, except when the receiver prohibits header list larger than a certain size.

RFC7540 defines SETTINGS_MAX_HEADER_LIST_SIZE's initial value to be unlimited.

### Actual behavior
If the receiver does not specify a maximum header list size in SETTINGS, Netty disallows header list sizes greater than 8K.

The problem seems to be [in HpackEncoder](https://github.com/netty/netty/blob/netty-4.1.22.Final/codec-http2/src/main/java/io/netty/handler/codec/http2/HpackEncoder.java#L89). While defaulting to 8K makes sense for HpackDecoder in order for it to protect itself against the remote, there's no need for such behavior in HpackEncoder. In fact, since SETTINGS_MAX_HEADER_LIST_SIZE is advisory and the content is locally generated, it could ignore the setting completely and only risk wasting bandwidth.

This was reported against gRPC at grpc/grpc-java#4284. CC @JLofgren

### Netty version
4.1.17.Final

I'd be happy to send a PR changing the initialization to use `MAX_HEADER_LIST_SIZE`. @Scottmitch, @normanmaurer, concerns?

ejona86:
Looks like I can call `defaultHttp2HeadersEncoder.maxHeaderListSize(MAX_HEADER_LIST_SIZE)` to workaround this. If done immediately on creation, it looks like it won't collide with any received SETTINGS requesting a lower value.

JLofgren:
I've created a demo of this bug here: https://github.com/JLofgren/demo-grpc-java-bug-4284

Locally I have built netty initializing the max header size in HpackEncoder to MAX_HEADER_LIST_SIZE and shown that this solves the issue. I am working on a PR, but am having some issues dotting all the i's of your PR process.

PR #7853, opened 2018-04-06
Do not enforce arbitrary max header list size in HpackEncoder

JLofgren:
Motivation:

When connecting to an HTTP/2 server that did not set any value for the
SETTINGS_MAX_HEADER_LIST_SIZE in the settings frame, the netty client was
imposing an arbitrary maximum header list size of 8kB. There should be no need
for the client to enforce such a limit if the server has not specified any
limit. This caused an issue for a grpc-java client that needed to send a large
header to a server via an Envoy proxy server. The error condition is
demonstrated here: https://github.com/JLofgren/demo-grpc-java-bug-4284

Fixes grpc-java issue - https://github.com/grpc/grpc-java/issues/4284
and netty issue #7825

Modifications:

In HpackEncoder use MAX_HEADER_LIST_SIZE as default maxHeader list size.

Result:

HpackEncoder will only enforce a max header list size if the server has
specified a limit in its settings frame.

Motivation:

Modification:

Result:

Fixes #<GitHub issue number>. 

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/HpackEncoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/HpackEncoder.java
@@ -41,7 +41,6 @@ import java.util.Arrays;
 import java.util.Map;
 
 import static io.netty.handler.codec.http2.HpackUtil.equalsConstantTime;
-import static io.netty.handler.codec.http2.Http2CodecUtil.DEFAULT_HEADER_LIST_SIZE;
 import static io.netty.handler.codec.http2.Http2CodecUtil.DEFAULT_HEADER_TABLE_SIZE;
 import static io.netty.handler.codec.http2.Http2CodecUtil.MAX_HEADER_LIST_SIZE;
 import static io.netty.handler.codec.http2.Http2CodecUtil.MAX_HEADER_TABLE_SIZE;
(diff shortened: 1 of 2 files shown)


normanmaurer:
Can you also add a unit test ?

JLofgren:
Added.

normanmaurer:
@carl-mastrangelo @Scottmitch @ejona86 PTAL

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackEncoder.java
```diff
@@ -86,7 +85,7 @@ public HpackEncoder(boolean ignoreMaxHeaderListSize) {
     public HpackEncoder(boolean ignoreMaxHeaderListSize, int arraySizeHint) {
         this.ignoreMaxHeaderListSize = ignoreMaxHeaderListSize;
         maxHeaderTableSize = DEFAULT_HEADER_TABLE_SIZE;
-        maxHeaderListSize = DEFAULT_HEADER_LIST_SIZE;
+        maxHeaderListSize = MAX_HEADER_LIST_SIZE;
```
Scottmitch: if you look at the comments for `DEFAULT_HEADER_LIST_SIZE` it describes that although the value is arbitrary it is not un-intentional, and technically not in violation of the spec. Having unlimited header size may leave the server exposed to memory exhaustion before it has the chance to communicate a lower limit. I think I would prefer to have the default configurable rather than just making it infinite. What about extracting the value from `AbstractHttp2ConnectionHandlerBuilder#initialSettings`?

JLofgren: @Scottmitch Are you saying there is possible scenario in the HTTP/2 protocol in which the receiver may be reading headers from the sender *before* the receiver has sent a settings frame to the sender?

JLofgren: I guess I should say: ... *before* the receiver's settings frame has been processed by the sender?

normanmaurer: @Scottmitch ping

Scottmitch: doh ... I thought this was in the decoder. I took a look at the decoder and looks like we are already using my suggestion of inferring the value from the initial settings. Sorry for the noise  ... lgtm.

Issue #7842, opened 2018-04-05
netty-dev-tools dependency not built early enough


davidmoten:
### Expected behavior
When I clone the repository and call `mvn clean install` it should complete without error.

### Actual behavior
`netty-common` build fails because `netty-dev-tools` has not been built yet.

### Steps to reproduce

as above

### Minimal yet complete reproducer code (or URL to code)
as above 

### Netty version
4.1.24.Final-SNAPSHOT

### JVM version (e.g. `java -version`)
1.8.0_161

### OS version (e.g. `uname -a`)
`Linux ubuntu1 3.19.0-33-generic #38~14.04.1-Ubuntu SMP Fri Nov 6 18:17:28 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux`

### How to fix
Fix might be to add `netty-dev-tools` to `common/pom.xml` as a *test* scoped dependency (with comments to indicate why it's there) so maven build is aware that `netty-common` depends on `netty-dev-tools`.

normanmaurer:
@davidmoten thanks for reporting

PR #7858, opened 2018-04-09
Add dev-tools dependency for commons

normanmaurer:
Motivation:

We need to add a dev-tools dependecy for commons as otherwise we may fail to fetch it before we try to use it.

Modifications:

Add dependency.

Result:

Fixes https://github.com/netty/netty/issues/7842


normanmaurer:
its an easy one.. let me merge this

Issue #7863, opened 2018-04-12
ArrayIndexOutOfBoundsException in AsciiString().indexOf()


Stefan-Weber:
### Actual behavior
Calling the method indexOf() in a special case throws ArrayIndexOutOfBoundsException.

java.lang.ArrayIndexOutOfBoundsException: 6
	at io.netty.util.AsciiString.indexOf(AsciiString.java:693)
	at ByteBufsTests.indexOfSuccess(ByteBufsTests.java:46)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
...

### Steps to reproduce
Run the following 2 lines of JUnit code, I hope this is enough for reproduction.

### Minimal yet complete reproducer code (or URL to code)
Assert.assertEquals("Test success in 4.1.(22|23).Final", 3, new AsciiString("012345").indexOf("345", 3));
Assert.assertEquals("Test fails with 'ArrayIndexOutOfBoundsException: 6' in 4.1.23.Final", -1, new AsciiString("012345").indexOf("abc", 3));

### Netty version
4.1.23.Final

### JVM version (e.g. `java -version`)
java version "1.8.0_66"
Java(TM) SE Runtime Environment (build 1.8.0_66-b17)
Java HotSpot(TM) 64-Bit Server VM (build 25.66-b17, mixed mode)

### OS version (e.g. `uname -a`)
Win7

normanmaurer:
@Stefan-Weber yes thanks... working on a patch

PR #7864, opened 2018-04-12
Fix ArrayIndexOutOfBoundsException in AsciiString indexOf(...) and la…

normanmaurer:
…stIndexOf(...)

Motivation:

AsciiString did not correctly calculate the bounds and so it was possible to see an ArrayIndexOutOfBoundsException.

Modifications:

Correctly calculate bounds.

Result:

Fixes https://github.com/netty/netty/issues/7863

### DIFF
--- a/common/src/main/java/io/netty/util/AsciiString.java
+++ b/common/src/main/java/io/netty/util/AsciiString.java
@@ -689,7 +689,7 @@ public final class AsciiString implements CharSequence, Comparable<CharSequence>
         }
         final byte firstCharAsByte = c2b0(firstChar);
         final int len = offset + start + length - subCount;
-        for (int i = start + offset; i <= len; ++i) {
+        for (int i = start + offset; i <= len && i < length; ++i) {
             if (value[i] == firstCharAsByte) {
                 int o1 = i, o2 = 0;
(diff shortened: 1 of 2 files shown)


normanmaurer:
@Scottmitch PTAL as this is a regression introduced by https://github.com/netty/netty/commit/602ee5444dcd2c9adad9bdffb93ae907ccad5649 I think.

Also @Stefan-Weber

normanmaurer:
Also @Scottmitch as you did the change...

common/src/main/java/io/netty/util/AsciiString.java
```diff
@@ -689,7 +689,7 @@ public int indexOf(CharSequence subString, int start) {
         }
         final byte firstCharAsByte = c2b0(firstChar);
         final int len = offset + start + length - subCount;
-        for (int i = start + offset; i <= len; ++i) {
+        for (int i = start + offset; i <= len && i < length; ++i) {
```
fenik17: I think it will be more correct:
```java
final int len = offset + length - subCount;
```
that is, do not add the `start` to the `len`.

normanmaurer: sounds good ... let me close this one

PR #7865, opened 2018-04-12
NPE in CorsHandler for unauthorized preflight request

gustavonalle:
**Motivation:**

NPE in `CorsHandler` if a pre-flight request is done using an Origin header which is
not allowed by any `CorsConfig` passed to the handler on creation.

**Modification:**

During the pre-flight, check the `CorsConfig` for `null` and handle it correctly by not returning
any access-control header

**Result:**

No more NPE for pre-flight requests with unauthorized origins.


normanmaurer:
@gustavonalle please fix checkstyle

gustavonalle:
Fixed

normanmaurer:
@gustavonalle thanks!

PR #7866, opened 2018-04-12
AsciiString#indexOf out of bounds

Scottmitch:
Motivation:
The bounds checking for AsciiString#indexOf and AsciiString#lastIndexOf is not correct and may lead to ArrayIndexOutOfBoundsException.

Modifications:
- Correct the bounds checking for AsciiString#indexOf and AsciiString#lastIndexOf

Result
Fixes https://github.com/netty/netty/issues/7863


normanmaurer:
@Scottmitch ship it once the CI completes

Issue #7860, opened 2018-04-10
HTTP2 Multiplex API support for unknown/custom frames


Bennett-Lynch:
### Expected behavior

The HTTP/2 spec explicitly permits extension of the protocol in the form of new frame types:
http://httpwg.org/specs/rfc7540.html#rfc.section.5.5

I would like/expect to be able to write/read custom frame types using the HTTP2 Multiplex API.  That is, if I write a Http2StreamFrame to a child channel with e.g., `frameType == (byte) 20`, I would be able to read the frame on the server's corresponding child channel.

### Actual behavior

The HTTP2 Multiplex API allows any object of type `Http2StreamFrame`:
https://github.com/netty/netty/blob/501662a77f8166382eefd18a8620983d27f59d4a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java#L993

The corresponding write, however, rejects any object that is of type `Http2Frame` and not one of the spec-defined frame types:
https://github.com/netty/netty/blob/501662a77f8166382eefd18a8620983d27f59d4a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java#L299-L304

Writing a custom implementation of `Http2StreamFrame` or `Http2UnknownFrame` to a child channel will be rejected by the `Http2FrameCodec`.

normanmaurer:
@Bennett-Lynch PTAL https://github.com/netty/netty/pull/7867

normanmaurer:
Fixed..

Bennett-Lynch:
Thanks for the quick fix!

normanmaurer:
Np

Bennett-Lynch:
@normanmaurer Were you able to confirm a corresponding inbound read from the outbound writeUnknownFrame() unit test you added?  I'm unable to detect a read on either the connection or child channel pipeline, and I'm having trouble pinpointing where it might be getting lost in the codecs.

normanmaurer:
In my simple unit test it seemed to work... that said maybe there is a bug somewhere. Please re-open this issue if you think there may be one and I will try to find some time to check

Bennett-Lynch:
@normanmaurer I'm hesitant to re-open as I might be doing something wrong.

When trying to write the following to a client's Http2 multiplex child channel:
```
Http2UnknownFrame unknownFrame = new DefaultHttp2UnknownFrame(
        (byte) 20,
        new Http2Flags().ack(false),
[...]

PR #7867, opened 2018-04-12
Allow to write Http2UnkownFrame when using Http2FrameCodec / Http2Mul…

normanmaurer:
…tiplexCodec

Motivation:

We should allow to write Http2UnkownFrame to allow custom extensions.

Modifications:

Allow to write Http2UnkownFrame
Add unit test

Result:

Fixes https://github.com/netty/netty/issues/7860.


normanmaurer:
Also @mosesn

PR #7875, opened 2018-04-18
Filter DNS results so they only contain the expected type when multip…

anuraaga:
…le types are present.

Motivation:

Currently, if a DNS server returns a non-preferred address type before the preferred one, then both will be returned as the result, and when only taking a single one, this usually ends up being the non-preferred type. However, the JDK requires lookups to only return the preferred type when possible to allow for backwards compatibility.

To allow a client to be able to resolve the appropriate address when running on a machine that does not support IPv6 but the DNS server returns IPv6 addresses before IPv4 addresses when querying.

Modification:

Filter the returned records to the expected type when both types are present.

Result:

Allows a client to run on a machine with IPv6 disabled even when a server returns both IPv4 and IPv6 results. Netty-based code can be a drop-in replacement for JDK-based code in such circumstances.

This PR filters results before returning them to respect JDK expectations. 

The test is a bit whacky, let me know if you have any ideas on making it cleaner.

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsAddressResolveContext.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsAddressResolveContext.java
@@ -19,6 +19,7 @@ import static io.netty.resolver.dns.DnsAddressDecoder.decodeAddress;
 
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.util.ArrayList;
 import java.util.List;
 
 import io.netty.channel.EventLoop;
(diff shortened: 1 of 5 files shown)


resolver-dns/src/main/java/io/netty/resolver/dns/DnsAddressResolveContext.java
```diff
@@ -61,6 +62,31 @@ boolean containsExpectedResult(List<InetAddress> finalResult) {
...
+            InetAddress address = unfiltered.get(i);
+            if (inetAddressType.isInstance(address)) {
+                numExpected++;
+            }
+        }
+        if (numExpected == size || numExpected == 0) {
+            // If all the results are the preferred type, or none of them are, then we don't need to do any filtering.
+            return unfiltered;
+        }
+        List<InetAddress> filtered = new ArrayList<InetAddress>(numExpected);
```
normanmaurer: @anuraaga can you just lazy create the ArrayList if needed and so remove the need to iterate two times ?

anuraaga: I wanted to avoid creating the array at all if there is no need to filter, which is pretty common. I can't see a way to do that with just one loop since all the records need to be confirmed before knowing whether it needs filtering. I could merge this into `containsExpectedResult` by changing it to return `null` when it currently returns `false` or a filtered list otherwise which cuts from three to two loops. What do you think?

normanmaurer: actually after thinking more about it I think its ok... let me merge it

Issue #7877, opened 2018-04-19
Eternal loop in Recycler.WeakOrderQueue.Head#run() that blocks ObjectCleaner thread and causes OOM.


syermakov:
Hi,

looks like in commit 6eb9674bf5ded128cb230a31987f14816b7f1886 there is a problem that causes an eternal loop in `Recycler.WeakOrderQueue.Head#run()`:
```
            public void run() {
                Link head = link;
                while (head != null) {
                    reclaimSpace(LINK_CAPACITY);
                    head = link.next;
                }
            }
```
should be 
```
            public void run() {
                Link head = link;
                while (head != null) {
                    reclaimSpace(LINK_CAPACITY);
                    head = head.next;
                }
            }
```
Note `head = link.next` vs `head = head.next`

### Expected behavior

Recycler.WeakOrderQueue.Head#run() should not block forever.

### Actual behavior

Recycler.WeakOrderQueue.Head#run() falls into an eternal loop and blocks ObjectCleaner thread.

### Steps to reproduce

Please see below...

### Minimal yet complete reproducer code (or URL to code)

Sorry for not providing a reproducer, this issues came out in our project's test suite as an OOM after a long run and we were able to find its cause only by analyzing the JVM heap dump. 

### Netty version

4.1.23.Final

### JVM version (e.g. `java -version`)

java version "1.8.0_162"
Java(TM) SE Runtime Environment (build 1.8.0_162-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.162-b12, mixed mode)

### OS version (e.g. `uname -a`)

Darwin MacBook-Pro-sy.local 17.5.0 Darwin Kernel Version 17.5.0: Mon Mar  5 22:24:32 PST 2018; root:xnu-4570.51.1~1/RELEASE_X86_64 x86_64

normanmaurer:
@syermakov OMG... looking. Thanks for reporting

normanmaurer:
Fixed by https://github.com/netty/netty/pull/7878

PR #7878, opened 2018-04-19
Fix eternal loop in Recycler.WeakOrderQueue.Head#run() that blocks Ob…

normanmaurer:
…jectCleaner thread

Motivation:

When trying to cleanup WeakOrderQueue by the ObjectCleaner we end up in an endless loop which will cause the ObjectCleaner to be not able to cleanup any other resources anymore.This bug was introduced by 6eb9674bf5ded128cb230a31987f14816b7f1886.

Modifications:

Correctly update link while cleanup

Result:

Fixes https://github.com/netty/netty/issues/7877

Motivation:

Modification:

Result:

Fixes #<GitHub issue number>. 


normanmaurer:
There is no good way for a test that I can think of but I think its very easy to see that this is correct.. Let me merge it

PR #7879, opened 2018-04-19
Update DatagramPacket.recipient() to return the actual destination IP

### LINKED ISSUE #4950: DatagramPacket.recipient() doesn't return the actual destination IP, but the IP the app is bound to
I'm using epoll (on previous versions of my app I was using NIO and I think the behaviour is the same) on an application that is listening on all available interfaces (0.0.0.0). When I try to get the actual UDP packet destination IP, DatagramPacket.recipient() returns 0.0.0.0 and not the real packet destination IP. It seems to always return the IP the app is bound to.

Please, let me know if you need some extra information.

SO: Ubuntu 14.04.4
Netty version: 4.0.32.Final

kubuzetto:
Motivation:

DatagramPacket.recipient() doesn't return the actual destination IP, but the IP the app is bound to.

Modification:

- `IP_RECVORIGDSTADDR` option is enabled for UDP sockets, which allows retrieval of ancillary information containing the original recipient.
- `_recvFrom(...)` function from `transport-native-unix-common/src/main/c/netty_unix_socket.c` is modified such that if `IP_RECVORIGDSTADDR` is set, `recvmsg` is used instead of `recvfrom`; enabling the retrieval of the original recipient.
- `DatagramSocketAddress` also contains a 'local' address, representing the recipient.
- `EpollDatagramChannel` is updated to return the retrieved recipient address instead of the address the channel is bound to.

Result:

Fixes #4950.

### DIFF
--- a/transport-native-unix-common/src/main/java/io/netty/channel/unix/DatagramSocketAddress.java
+++ b/transport-native-unix-common/src/main/java/io/netty/channel/unix/DatagramSocketAddress.java
@@ -28,10 +28,16 @@ public final class DatagramSocketAddress extends InetSocketAddress {
 
     // holds the amount of received bytes
     private final int receivedAmount;
+    private DatagramSocketAddress localAddress;
 
-    DatagramSocketAddress(String addr, int port, int receivedAmount) {
+    DatagramSocketAddress(String addr, int port, int receivedAmount, DatagramSocketAddress local) {
         super(addr, port);
         this.receivedAmount = receivedAmount;
+        localAddress = local;
+    }
+
+    public DatagramSocketAddress localAddress() {
+        return localAddress;
     }
 
     public int receivedAmount() {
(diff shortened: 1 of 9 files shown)


transport-native-unix-common/src/main/java/io/netty/channel/unix/DatagramSocketAddress.java
```diff
@@ -28,10 +28,16 @@
 
     // holds the amount of received bytes
     private final int receivedAmount;
+    private DatagramSocketAddress localAddress;
```
normanmaurer: final

kubuzetto:
@normanmaurer Testing this might be a bit difficult. As long we are not enabling `TPROXY` with `IP_TRANSPARENT` (which, AFAIK, requires either root privileges or `CAP_NET_ADMIN` capability) and capturing traffic that is not targeted to our socket, `EpollDatagramChannel.localAddress()` will be equal to `DatagramPacket.recipient()` anyway. I don't suppose modifying the `iptables` settings of an automated test environment is viable :smile: 

The best available test I can come up with is setting `IP_RECVORIGDSTADDR` and verifying that `EpollDatagramChannel.localAddress()` is actually equal to `DatagramPacket.recipient()`, but that doesn't tell us much.

normanmaurer:
@trustin @carl-mastrangelo mind taking a look as well ?

normanmaurer:
@kubuzetto thanks a lot!

kubuzetto:
glad to help :)

PR #7883, opened 2018-04-21
Implement isWritable and ensureWritable on ReadOnlyByteBufferBuf

RikkiGibson:
Motivation:

It should be possible to write a ReadOnlyByteBufferBuf to a channel without errors. However, ReadOnlyByteBufferBuf does not override isWritable and ensureWritable, which can cause some handlers to mistakenly [assume they can write to the ReadOnlyByteBufferBuf](https://github.com/netty/netty/tree/2e92a2f5cdadb1c5932a75d3a84b9b89e77bde30/handler/src/main/java/io/netty/handler/ssl/SslHandler.java#L1960), resulting in ReadOnlyBufferException.

Modification:

Added isWritable and ensureWritable method overrides on ReadOnlyByteBufferBuf to indicate that it is never writable. Added tests for these methods.

Result:

Can successfully write ReadOnlyByteBufferBuf to a channel with an SslHandler (or any other handler which may attempt to write to the ByteBuf it receives).

### DIFF
--- a/buffer/src/test/java/io/netty/buffer/ReadOnlyDirectByteBufferBufTest.java
+++ b/buffer/src/test/java/io/netty/buffer/ReadOnlyDirectByteBufferBufTest.java
@@ -43,6 +43,38 @@ public class ReadOnlyDirectByteBufferBufTest {
         buffer(allocate(1));
     }
 
+    @Test
+    public void shouldIndicateNotWritable() {
+        ByteBuf buf = buffer(allocate(8).asReadOnlyBuffer()).clear();
+        Assert.assertFalse(buf.isWritable());
+    }
+
+    @Test
+    public void shouldIndicateNotWritableAnyNumber() {
+        ByteBuf buf = buffer(allocate(8).asReadOnlyBuffer()).clear();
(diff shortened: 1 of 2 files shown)


buffer/src/test/java/io/netty/buffer/ReadOnlyDirectByteBufferBufTest.java
```diff
@@ -43,6 +43,38 @@ public void testConstructWithWritable() {
         buffer(allocate(1));
     }
 
+    @Test
+    public void shouldIndicateNotWritable() {
+        ByteBuf buf = buffer(allocate(8).asReadOnlyBuffer()).clear();
+        Assert.assertFalse(buf.isWritable());
```
normanmaurer: Please also call buf.release() in all tests (just like in the existing tests)

RikkiGibson:
@normanmaurer Added `try/finally` to release the ByteBuf. Let me know if anything else is needed. :+1:

normanmaurer:
@RikkiGibson thanks all good :)

Issue #7888, opened 2018-04-25
HttpResponseEncoder does not treat status 205 as a content-is-always-empty response


nicmunroe:
### Expected behavior
RFC 2616 (https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.2.6) states that responses with HTTP status code 205 should never return a payload: `The response MUST NOT include an entity`.

Additionally, the RFC 7230 section on content-length (https://tools.ietf.org/html/rfc7230#section-3.3.2) does not call out 205 responses as being in the allowed-to-lie-about-content-length category (only HEAD and 204 responses are in that category). Also from that RFC 7230 section, 205 responses are not in the must-never-send-content-length category (only 1xx, 204, and CONNECT with a 2xx are in that category).

So based on that I would expect that when Netty encodes a 205 response it should:
* Strip out any payload (other than the no-data chunk when transfer-encoding is chunked)
* Set content-length to 0 for non-chunked responses

NOTE: The HTTP spec is obviously difficult to follow on these corner cases. Please double check my logic on this issue - I believe my analysis to be correct based on my reading of those RFCs, but I could have interpreted it wrong, or I might be missing something in these RFCs or another spec doc entirely that explains how to handle 205 more directly than what I've cobbled together here.

### Actual behavior
Netty does not currently do any smart filtering for 205 responses like it does for other always-empty responses (204, 304, HEAD, etc). I can send a 205 response with a payload and that payload will be sent over the wire as if it was a 200.

Netty *does* appear to sort of pay attention to this on the decoding side - i.e. HttpRequestDecoder drops the payload on the floor if it sees an incoming 205 with a payload. So this issue is mainly about the HttpResponseEncoder side (unless you find something I missed).

### Steps to reproduce
Send a 205 response with a payload from a Netty server and monitor the bytes on the wire. You'll see the payload is sent, and content-length is not set to 0 for non-chunked responses.

### Minimal yet complete reproducer code (or URL to code)
``` java
import org.junit.*;

import java.io.IOException;
import java.net.ServerSocket;
import java.util.*;
import java.util.concurrent.CompletableFuture;
import java.util.concurrent.TimeUnit;

import io.netty.bootstrap.*;
import io.netty.buffer.*;
import io.netty.channel.*;
import io.netty.channel.nio.NioEventLoopGroup;
import io.netty.channel.socket.SocketChannel;
import io.netty.channel.socket.nio.*;
import io.netty.handler.codec.http.*;
[...]

PR #7891, opened 2018-04-25
Correctly handle responses with status 205 and payload.

normanmaurer:
Motivation:

HTTP responses with status of 205 should not contain a payload. We should enforce this.

Modifications:

Correctly handle responses with status 205 and payload by set Content-Length: 0 header and stripping out the content.

Result:

Fixes https://github.com/netty/netty/issues/7888


normanmaurer:
FYI @nicmunroe

nicmunroe:
LGTM (other than the s/204/205/ like @ejona86 mentioned). Thanks!

PR #7903, opened 2018-05-03
Allow to disable native transport and native ssl support via system p…

normanmaurer:
…roperty.

Motivation:

Sometimes it's useful to disable native transports / native ssl to debug a problem. We should allow to do so with a system property so people not need to adjust code for this.

Modifications:

Add system properties which allow to disable native transport and native ssl.

Result:

Easier to disable native code usage without code changes.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
+++ b/handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
@@ -72,45 +72,54 @@ public final class OpenSsl {
     static {
         Throwable cause = null;
 
-        // Test if netty-tcnative is in the classpath first.
-        try {
-            Class.forName("io.netty.internal.tcnative.SSL", false, OpenSsl.class.getClassLoader());
-        } catch (ClassNotFoundException t) {
-            cause = t;
-            logger.debug(
-                    "netty-tcnative not in the classpath; " +
-                    OpenSslEngine.class.getSimpleName() + " will be unavailable.");
-        }
+        if (SystemPropertyUtil.getBoolean("io.netty.handler.ssl.noOpenSsl", false)) {
+            cause = new UnsupportedOperationException(
+                    "OpenSSL was explicit disabled with -Dio.netty.handler.ssl.noOpenSsl=true");
 
-        // If in the classpath, try to load the native library and initialize netty-tcnative.
-        if (cause == null) {
+            logger.debug(
+                    "netty-tcnative explicit disabled; " +
+                            OpenSslEngine.class.getSimpleName() + " will be unavailable.", cause);
+        } else {
+            // Test if netty-tcnative is in the classpath first.
             try {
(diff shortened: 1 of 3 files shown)


handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
```diff
@@ -72,45 +72,54 @@
...
+                Class.forName("io.netty.internal.tcnative.SSL", false, OpenSsl.class.getClassLoader());
+            } catch (ClassNotFoundException t) {
                 cause = t;
                 logger.debug(
-                    "Failed to load netty-tcnative; " +
-                        OpenSslEngine.class.getSimpleName() + " will be unavailable, unless the " +
-                        "application has already loaded the symbols by some other means. " +
-                        "See http://netty.io/wiki/forked-tomcat-native.html for more information.", t);
+                        "netty-tcnative not in the classpath; " +
+                                OpenSslEngine.class.getSimpleName() + " will be unavailable.");
```
carl-mastrangelo: nit: include `t` in the debug?

handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
```diff
@@ -72,45 +72,54 @@
...
+                    initializeTcNative();
+
+                    // The library was initialized successfully. If loading the library failed above,
+                    // reset the cause now since it appears that the library was loaded by some other
+                    // means.
+                    cause = null;
+                } catch (Throwable t) {
+                    if (cause == null) {
+                        cause = t;
+                    }
```
carl-mastrangelo: optional:
```
else {
   ThrowableUtils.addSuppressed(cause, t);
}
```

handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
normanmaurer: I can do this as a followup if you like as this code was just moved.

handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
normanmaurer: I can do this as a followup if you like as this code was just moved.

normanmaurer:
@trustin done... will merge once ci pass.

jchambers:
Awesome! This is, indeed, very useful for testing.

Some notes for future readers who might also be interested in this change:

- This is dramatically easier to read with [whitespace-only changes hidden](https://github.com/netty/netty/pull/7903/files?w=1)
- The system property for disabling native SSL providers (i.e. netty-tcnative) is `io.netty.handler.ssl.noOpenSsl` (example: `-Dio.netty.handler.ssl.noOpenSsl=true`)
- The system property for disabling native transports (i.e. epoll and kqueue) is `io.netty.transport.noNative` (example: `-Dio.netty.transport.noNative=true`)

Issue #7882, opened 2018-04-21
AddressResolver.resolveAll doesn't return consistent result


mrniko:
### Expected behavior
AddressResolver.resolveAll always returns full list of ips.

### Actual behavior
AddressResolver.resolveAll returns some part of full list ips.

### Steps to reproduce
`dev.myredis.com` host has two ips: 127.0.0.2, 127.0.0.3.
I guess AddressResolver.resolveAll should always return these two ips, but it's not true on practice if DnsAddressResolverGroup object shared between multiple threads

If this object created per invocation then the problem is gone.
Seems like DnsAddressResolverGroup is not thread safe!

### Minimal yet complete reproducer code (or URL to code)
Here is my test case:
```java
        NioEventLoopGroup niogroup = new NioEventLoopGroup();
        
        DnsAddressResolverGroup group = new DnsAddressResolverGroup(NioDatagramChannel.class, DnsServerAddressStreamProviders.platformDefault());
        ExecutorService es = Executors.newFixedThreadPool(2);
        for (int i = 0; i < 100; i++) {
            es.execute(new Runnable() {
                @Override
                public void run() {
                    AddressResolver<InetSocketAddress> resolver = group.getResolver(niogroup.next());
                    try {
                        URI uri = new URI("redis://dev.myredis.com:6379");
                        Future<List<InetSocketAddress>> allNodes = resolver.resolveAll(InetSocketAddress.createUnresolved(uri.getHost(), uri.getPort()));
                        List<InetSocketAddress> list = allNodes.syncUninterruptibly().getNow();
                        System.out.println(list);
                    } catch (URISyntaxException e) {
                        // TODO Auto-generated catch block
                        e.printStackTrace();
                    }
                }
            });
        }

        Thread.sleep(10000);
        group.close();
        es.shutdown();
        niogroup.shutdownGracefully();
```

Output is:

[dev.myredis.com/127.0.0.3:6379, dev.myredis.com/127.0.0.2:6379]
[dev.myredis.com/127.0.0.3:6379, dev.myredis.com/127.0.0.2:6379]
[dev.myredis.com/127.0.0.2:6379, dev.myredis.com/127.0.0.3:6379]
[dev.myredis.com/127.0.0.2:6379, dev.myredis.com/127.0.0.3:6379]
[dev.myredis.com/127.0.0.3:6379, dev.myredis.com/127.0.0.2:6379]
[...]

PR #7905, opened 2018-05-03
Ensure no partial results are used via the DnsCache.

normanmaurer:
Motivation:

We need to ensure we only use the DnsCache from the EventLoop when try to resolve a query with it. This is because otherwise we may see only partial results for a previous query as we are not done adding everything. Beside this we also should delay commiting any cached entries until we have completed the query.

Modifications:

- Delay commiting any entry to the cache until the query is complete
- Ensure we acces the cache only from the EventLoop to never see partial results

Result:

Fixes https://github.com/netty/netty/issues/7882


normanmaurer:
Working on a test case

PR #7906, opened 2018-05-04
DefaultDnsCache should store more then one Entry per hostname.

normanmaurer:
Motivation:

Due a bug we did never store more then one address per hostname in DefaultDnsCache.

Modifications:

- Correctly store multiple entries per hostname
- Add tests

Result:

DefaultDnsCache correctly stores more then one entry. Also fixes https://github.com/netty/netty/issues/7882 .


trustin:
I wonder if using `Set` is a better idea, though. Perhaps `List` is better given there will only be <10 entries in most cases.

normanmaurer:
@trustin yes I think List is better and we also need the List as this is what we return to the user as well.

PR #7909, opened 2018-05-04
Add test for caching failed queries in DefaultDnsCache.

normanmaurer:
Motivation:

We had no test that validated the handling of caching failures for DefaultDnsCache.

Modifications:

Add testcase.

Result:

More tests FTW.


normanmaurer:
Let me just merge this as its a unit test only.

PR #7911, opened 2018-05-06
Remove HpackDecoder.maxHeaderListSizeGoAway

### LINKED ISSUE #7887: HpackDecoder's maxHeaderListSizeGoAway kills stream unnecessarily
Consider this a bit of a TODO for myself, although @Scottmitch or @normanmaurer may have opinions about avoiding killing the connection in this case.

### Expected behavior

Sending too large of headers would not unnecessarily kill the connection.

### Actual behavior

When maxHeaderListSizeGoAway is exceeded, HpackDecoder will throw a connection error, killing the entire connection.

```
                            // Check new header size against max header size
                            if ((long) index + nameLength > maxHeaderListSizeGoAway - headersLength) {
                                headerListSizeExceeded(maxHeaderListSizeGoAway);
                            }
```

Note, I'm not talking about the separate check later against maxHeaderListSize and throws a _stream_ error:
```
        // we have read all of our headers, and not exceeded maxHeaderListSizeGoAway see if we have
        // exceeded our actual maxHeaderListSize. This must be done here to prevent dynamic table
        // corruption
        if (headersLength > maxHeaderListSize) {
            headerListSizeExceeded(streamId, maxHeaderListSize, true);
        }
```

Even though both of those call `headerListSizeExceeded()`, they call different overloads of the method which will throw different types of exceptions (connection vs stream).

The value of maxHeaderListSizeGoAway is likely 1.25 the size of maxHeaderListSize, although I see there's some ways it could be something different and I didn't dig in more.

This killing of the connection seems unnecessary though, as:
1) It seems it would be pretty easy to throw away this batch of headers, by setting `headers` to a `Http2Headers` implementation that throws away further adds (similar complexity to `EmptyHttp2Headers`). This could be done any time `headersLength > maxHeaderListSize`.
2) When decode is called, the "entire header block is contained in {@code in}". So this protection doesn't protect us from buffering any more data; it only reduced the decoded headers size. That is, throwing away decoded headers as suggested in (1) does not seem to expose us to other problems.

Note also there are separate checks in `DefaultHttp2FrameReader.HeadersBlockBuilder` that protect against too large of compressed HEADERS+CONTINUATION. This is what constrains the size of `in`.

ejona86:
Motivation:

When a sender sends too large of headers it should not unnecessarily
kill the connection, as killing the connection is a heavy-handed
solution while SETTINGS_MAX_HEADER_LIST_SIZE is advisory and may be
ignored.

The maxHeaderListSizeGoAway limit in HpackDecoder is unnecessary because
any headers causing the list to exceeding the max size can simply be
thrown away. In addition, DefaultHttp2FrameReader.HeadersBlockBuilder
limits the entire block to maxHeaderListSizeGoAway. Thus individual
literals are limited to maxHeaderListSizeGoAway.

(Technically, literals are limited to 1.6x maxHeaderListSizeGoAway,
since the canonical Huffman code has a maximum compression ratio of
.625. However, the "unnecessary" limit in HpackDecoder was also being
applied to compressed sizes.)

Modifications:

Remove maxHeaderListSizeGoAway checking in HpackDecoder and instead
eagerly throw away any headers causing the list to exceed
maxHeaderListSize.

Result:

Fewer large header cases will trigger connection-killing.
DefaultHttp2FrameReader.HeadersBlockBuilder will still kill the
connection when maxHeaderListSizeGoAway is exceeded, however.

Fixes #7887

------

The Sink interface is a step toward splitting headerList processing out of the core Hpack decoding (verification is the only thing left). I can swap to referring to Http2HeadersSink directly, but it seemed more readable to use the interface as a "we're not assuming anything about the implementation."

CC @Scottmitch

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
@@ -22,6 +22,7 @@ import io.netty.util.internal.UnstableApi;
 import static io.netty.handler.codec.http2.Http2CodecUtil.DEFAULT_HEADER_LIST_SIZE;
 import static io.netty.handler.codec.http2.Http2CodecUtil.DEFAULT_INITIAL_HUFFMAN_DECODE_CAPACITY;
 import static io.netty.handler.codec.http2.Http2Error.COMPRESSION_ERROR;
+import static io.netty.handler.codec.http2.Http2Error.INTERNAL_ERROR;
 import static io.netty.handler.codec.http2.Http2Exception.connectionError;
 
 @UnstableApi
@@ -31,6 +32,7 @@ public class DefaultHttp2HeadersDecoder implements Http2HeadersDecoder, Http2Hea
 
     private final HpackDecoder hpackDecoder;
     private final boolean validateHeaders;
+    private long maxHeaderListSizeGoAway;
 
     /**
      * Used to calculate an exponential moving average of header sizes to get an estimate of how large the data
@@ -79,6 +81,8 @@ public class DefaultHttp2HeadersDecoder implements Http2HeadersDecoder, Http2Hea
     DefaultHttp2HeadersDecoder(boolean validateHeaders, HpackDecoder hpackDecoder) {
         this.hpackDecoder = ObjectUtil.checkNotNull(hpackDecoder, "hpackDecoder");
         this.validateHeaders = validateHeaders;
+        this.maxHeaderListSizeGoAway =
+                Http2CodecUtil.calculateMaxHeaderListSizeGoAway(hpackDecoder.getMaxHeaderListSize());
     }
 
     @Override
@@ -93,7 +97,12 @@ public class DefaultHttp2HeadersDecoder implements Http2HeadersDecoder, Http2Hea
 
     @Override
     public void maxHeaderListSize(long max, long goAwayMax) throws Http2Exception {
-        hpackDecoder.setMaxHeaderListSize(max, goAwayMax);
+        if (goAwayMax < max || goAwayMax < 0) {
+            throw connectionError(INTERNAL_ERROR, "Header List Size GO_AWAY %d must be non-negative and >= %d",
+                    goAwayMax, max);
+        }
+        hpackDecoder.setMaxHeaderListSize(max);
+        this.maxHeaderListSizeGoAway = goAwayMax;
     }
 
     @Override
@@ -103,7 +112,7 @@ public class DefaultHttp2HeadersDecoder implements Http2HeadersDecoder, Http2Hea
 
     @Override
     public long maxHeaderListSizeGoAway() {
-        return hpackDecoder.getMaxHeaderListSizeGoAway();
+        return maxHeaderListSizeGoAway;
     }
 
     @Override
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
@@ -42,7 +42,6 @@ import static io.netty.handler.codec.http2.Http2CodecUtil.MIN_HEADER_LIST_SIZE;
 import static io.netty.handler.codec.http2.Http2CodecUtil.MIN_HEADER_TABLE_SIZE;
 import static io.netty.handler.codec.http2.Http2CodecUtil.headerListSizeExceeded;
 import static io.netty.handler.codec.http2.Http2Error.COMPRESSION_ERROR;
-import static io.netty.handler.codec.http2.Http2Error.INTERNAL_ERROR;
 import static io.netty.handler.codec.http2.Http2Error.PROTOCOL_ERROR;
 import static io.netty.handler.codec.http2.Http2Exception.connectionError;
 import static io.netty.handler.codec.http2.Http2Headers.PseudoHeaderName.getPseudoHeader;
@@ -84,7 +83,6 @@ final class HpackDecoder {
 
     private final HpackDynamicTable hpackDynamicTable;
     private final HpackHuffmanDecoder hpackHuffmanDecoder;
-    private long maxHeaderListSizeGoAway;
     private long maxHeaderListSize;
     private long maxDynamicTableSize;
     private long encoderMaxDynamicTableSize;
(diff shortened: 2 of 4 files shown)


codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
```diff
@@ -93,7 +97,12 @@ public long maxHeaderTableSize() {
 
     @Override
     public void maxHeaderListSize(long max, long goAwayMax) throws Http2Exception {
-        hpackDecoder.setMaxHeaderListSize(max, goAwayMax);
+        if (goAwayMax < max || goAwayMax < 0) {
+            throw connectionError(INTERNAL_ERROR, "Header List Size GO_AWAY %d must be positive and >= %d",
+                    goAwayMax, max);
+        }
+        hpackDecoder.setMaxHeaderListSize(max);
+        this.maxHeaderListSizeGoAway = max;
```
ejona86: s/max/goAwayMax/

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
```diff
@@ -93,7 +97,12 @@ public long maxHeaderTableSize() {
 
     @Override
     public void maxHeaderListSize(long max, long goAwayMax) throws Http2Exception {
-        hpackDecoder.setMaxHeaderListSize(max, goAwayMax);
+        if (goAwayMax < max || goAwayMax < 0) {
+            throw connectionError(INTERNAL_ERROR, "Header List Size GO_AWAY %d must be positive and >= %d",
```
fenik17: "positive or zero"?

ejona86: Yeah, looks like it should be "non-negative". This was pre-existing code and just moved, but I can update it.

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
```diff
@@ -108,22 +106,31 @@
...
+        if (sink.exceededMaxLength()) {
+            headerListSizeExceeded(streamId, maxHeaderListSize, true);
+        }
+    }
+
     /**
      * Decode the header block into header fields.
      * <p>
      * This method assumes the entire header block is contained in {@code in}.
      */
```
normanmaurer: @ejona86 move this to the public method (as it was before) ?

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
```diff
@@ -341,27 +326,18 @@ public void setMaxHeaderTableSize(long maxHeaderTableSize) throws Http2Exception
         }
     }
 
-    public void setMaxHeaderListSize(long maxHeaderListSize, long maxHeaderListSizeGoAway) throws Http2Exception {
```
normanmaurer: @ejona86 should we just keep the method as well and deprecate it and ignore the second argument ?

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
```diff
@@ -553,4 +517,35 @@ static long decodeULE128(ByteBuf in, long result) throws Http2Exception {
         REQUEST_PSEUDO_HEADER,
         RESPONSE_PSEUDO_HEADER
     }
+
+    private interface Sink {
+        void appendToHeaderList(CharSequence name, CharSequence value);
+    }
+
+    private static class Http2HeadersSink implements Sink {
```
normanmaurer: final ?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
normanmaurer: ?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
normanmaurer: Whatever you prefer

normanmaurer:
@mosesn PTAL as well

mosesn:
I haven't looked at the PR like, but my vague memory of maxHeaderListSizeGoAway was that it was intended to help us defend against bad actors who might try to make the remote peer allocate too much memory in headers?  Is that something we don't need to worry about?

normanmaurer:
@mosesn I think @ejona86 explained in his commit message why this is not "needed". Did you check?

ejona86:
@mosesn, check out the description, where I mention DefaultHttp2FrameReader. Also, this is specifically for maxHeaderListSizeGoAway, the "GoAway" portion being important. Pieces for maxHeaderListSize are still in place.

mosesn:
Got it, thanks for explaining.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
ejona86: Done.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2HeadersDecoder.java
ejona86: It was a bug. `goAwayMax` wasn't used when it should have been. Fixed.

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
ejona86: Done. This had been public in my changes until the last minute. Oops.

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
ejona86: Unclear how many people are using that, but sure. It's easy. Done.

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
ejona86: I don't tend to put `final` on private classes, but sure. Done.

ejona86:
@normanmaurer, I've addressed the comments. I can squash when ready.

normanmaurer:
@ejona86 please squash... LGTM

ejona86:
Rebased.

normanmaurer:
@ejona86 thanks!

PR #7917, opened 2018-05-07
Add test that we handle thread.interrupt() in NioEventLoop

normanmaurer:
Motivation:

We added some code to guard against thread.interrupt() in NioEventLoop but did not added a test.

Modifications:

Add testcase.

Result:

Verify that we correctly handle interrupt().

### DIFF
--- a/transport/src/test/java/io/netty/channel/nio/NioEventLoopTest.java
+++ b/transport/src/test/java/io/netty/channel/nio/NioEventLoopTest.java
@@ -25,6 +25,7 @@ import io.netty.util.concurrent.Future;
 import org.junit.Test;
 
 import java.nio.channels.Selector;
+import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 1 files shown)


transport/src/test/java/io/netty/channel/nio/NioEventLoopTest.java
```diff
@@ -107,4 +108,47 @@ public void run() {
...
+            }).syncUninterruptibly();
+
+            loop.schedule(new Runnable() {
+                @Override
+                public void run() {
+                    latch.countDown();
+                }
+            }, 2, TimeUnit.SECONDS).syncUninterruptibly();
+
+            latch.await();
```
trustin: Could you explain why we need this part (counting down twice)?

normanmaurer: I just wanted to see that we still process tasks correctly

PR #7918, opened 2018-05-07
Fix tiny typo in FixedChannelPool error message

lutovich:
Motivation:

Closed `FixedChannelPool` fails acquire and release operations with `IllegalStateException`s. These exceptions had message "FixedChannelPooled was closed". Here "FixedChannelPooled" looks like a typo and should probably be "FixedChannelPool".

Modifications:

Changed exception message to "FixedChannelPool was closed". Added test to assert that acquire operation on a closed pool fails.

Result:

A tiny bit cleaner exception message.

### DIFF
--- a/transport/src/main/java/io/netty/channel/pool/FixedChannelPool.java
+++ b/transport/src/main/java/io/netty/channel/pool/FixedChannelPool.java
@@ -43,10 +43,10 @@ public class FixedChannelPool extends SimpleChannelPool {
             new TimeoutException("Acquire operation took longer then configured maximum time"),
             FixedChannelPool.class, "<init>(...)");
     static final IllegalStateException POOL_CLOSED_ON_RELEASE_EXCEPTION = ThrowableUtil.unknownStackTrace(
-            new IllegalStateException("FixedChannelPooled was closed"),
+            new IllegalStateException("FixedChannelPool was closed"),
             FixedChannelPool.class, "release(...)");
     static final IllegalStateException POOL_CLOSED_ON_ACQUIRE_EXCEPTION = ThrowableUtil.unknownStackTrace(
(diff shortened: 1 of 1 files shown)


transport/src/test/java/io/netty/channel/pool/FixedChannelPoolTest.java
```diff
@@ -310,6 +310,7 @@ public void run() {
             pool.release(channel).syncUninterruptibly();
             fail();
         } catch (IllegalStateException e) {
+            assertEquals("FixedChannelPool was closed", e.getMessage());
```
normanmaurer: @lutovich IMHO we should remove this assert as the message may change again in the future and people should really not depend on it.

transport/src/test/java/io/netty/channel/pool/FixedChannelPoolTest.java
```diff
@@ -318,6 +319,25 @@ public void run() {
         sc.close().syncUninterruptibly();
     }
 
+    @Test
+    public void testAcquireAfterClosePool() throws Exception {
```
normanmaurer: @lutovich remove ?

lutovich: Should the test be removed or the exception message assertion?

normanmaurer: I would say the test and if you want to do another PR to add more tests just do it there. The test has not really anything to do with the change of the message.

normanmaurer:
@lutovich please ping me once done...

lutovich:
@normanmaurer PR updated. It now fixes just the exception message.

normanmaurer:
@lutovich thanks!

PR #7920, opened 2018-05-08
Enfore upper limit for minTtl when using DefaultCacheEntry.

normanmaurer:
Motivation:

a598c3b69b55f930b91a8265f42d1a3cceb75ddd added a upper limit for ttl but missed to also do the same for minTtl.

Modifications:

- Add upper limit for minTtl
- Add testcase.

Result:

No more IllegalArgumentException possible.


normanmaurer:
Just merging this as the previous fix was incomplete.

PR #7921, opened 2018-05-08
Fix the error message in ReferenceCounted.release

windie:
Motivation:

When a buffer is over-released, the current error message of `IllegalReferenceCountException` is `refCnt: XXX, increment: XXX`, which is confusing. The correct message should be `refCnt: XXX, decrement: XXX`.

Modifications:

Pass `-decrement` to create `IllegalReferenceCountException`.

Result:

The error message will be `refCnt: XXX, decrement: XXX` when a buffer is over-released.


normanmaurer:
@windie nice one... Welcome back :)

normanmaurer:
@windie thanks!

hepin1989:
I was wondering to prepare a fix:).

PR #7922, opened 2018-05-09
Make build pass on Java 10

normanmaurer:
Motivation:

Java 10 is out so we should be able to build netty with it (and run the tests).

Modifications:

- Update Mockito and JBoss Marshalling to support Java 10
- Fix unit test to not depend on specific cipher which is not present in Java 10 anymore

Result:

Netty builds (and runs all tests) when using Java 10

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -68,6 +68,21 @@
   </developers>
 
   <profiles>
+    <!-- JDK10 -->
+    <profile>
+      <id>java10</id>
+      <activation>
+        <jdk>10</jdk>
+      </activation>
+      <properties>
+        <!-- Not use alpn agent as Java10 supports alpn out of the box -->
+        <argLine.alpnAgent />
+        <forbiddenapis.skip>true</forbiddenapis.skip>
+        <!-- Needed because of https://issues.apache.org/jira/browse/MENFORCER-275 -->
+        <enforcer.plugin.version>3.0.0-M1</enforcer.plugin.version>
+      </properties>
+    </profile>
+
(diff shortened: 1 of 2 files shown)


normanmaurer:
These changes allow to also run builds with Java 10 🎉

karlpauls:
@normanmaurer, done. See #7923

normanmaurer:
@karlpauls updated this one as well... thanks!

normanmaurer:
@karlpauls PTAL again :)

karlpauls:
@normanmaurer, lgtm - however, not sure the framework version increase needs to still be in this one. Shouldn't that come from the 4.1 branch?

normanmaurer:
Yep this one needs a rebase before merging

normanmaurer:
And rebased...

pom.xml
```diff
@@ -68,6 +68,21 @@
...
+      <properties>
+        <!-- Not use alpn agent as Java10 supports alpn out of the box -->
+        <argLine.alpnAgent />
+        <forbiddenapis.skip>true</forbiddenapis.skip>
+        <!-- Needed because of https://issues.apache.org/jira/browse/MENFORCER-275 -->
+        <enforcer.plugin.version>3.0.0-M1</enforcer.plugin.version>
+      </properties>
+      <activation>
+        <jdk>10</jdk>
+      </activation>
```
trustin: nit: probably better moving this section right after `<id/>` for readability?

normanmaurer: sure why not

PR #7923, opened 2018-05-09
Make osgi testing work with java>=9 by updating to felix 5.6.10

karlpauls:
Motivation:

Currently, the testing-osgi is set to skip if run with java>=9. That is not necessary when using a newer version of Felix.

Modification:

Update to Felix framework 5.6.10 (which has better jpms support), add some more --add-opens to not have WARN messages, and remove the skipOsgiTestsuite setting from the parent pom.

Result:

OSGi tests run and pass on java>=9.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -78,7 +78,6 @@
         <!-- Not use alpn agent as Java9 supports alpn out of the box -->
         <argLine.alpnAgent />
         <!-- Skip as maven plugin not works with Java9 yet --> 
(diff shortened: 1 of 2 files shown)


normanmaurer:
@karlpauls can you also sign our ICLA:

https://netty.io/s/icla 

Please let me know once done :)

karlpauls:
@normanmaurer done.

normanmaurer:
@karlpauls thanks a lot!

PR #7926, opened 2018-05-11
Ensure Netty can be build with Java 11

normanmaurer:
Motivation:

Java 11 will be out soon, so we should be able to build (and run tests) netty.

Modifications:

- Add dependency that is needed till Java 11
- Adjust tests so these also pass on Java 11 (SocketChannelImpl.close() behavious a bit differently now).

Result:

Build also works (and tests pass) on Java 11.

### DIFF
--- a/testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketHalfClosedTest.java
+++ b/testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketHalfClosedTest.java
@@ -37,6 +37,7 @@ import io.netty.util.UncheckedBooleanSupplier;
 import org.junit.Test;
 
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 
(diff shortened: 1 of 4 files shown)


normanmaurer:
Asked a few questions about the change on the nio-dev mailinglist (still awaiting moderation tho): http://mail.openjdk.java.net/pipermail/nio-dev/2018-May/thread.html

normanmaurer:
Related discussion: http://mail.openjdk.java.net/pipermail/nio-dev/2018-May/005102.html

testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketHalfClosedTest.java
```diff
@@ -327,8 +328,18 @@ protected void channelRead0(ChannelHandlerContext ctx, ByteBuf msg) throws Excep
                     public void operationComplete(ChannelFuture future) throws Exception {
                         future.channel().close().addListener(new ChannelFutureListener() {
                             @Override
-                            public void operationComplete(ChannelFuture future) throws Exception {
-                                followerCloseLatch.countDown();
+                            public void operationComplete(final ChannelFuture future) throws Exception {
+                                // This is a bit racy but there is no better way how to handle this in Java11.
+                                // The problem is that on close() the underlying FD will not actually be closed directly
+                                // but the close will be done after the Selector did process all events. Because of
```
carl-mastrangelo: Dumb question: if this test is the only thing using the event loop, this should not be racy right?  If you can limit the selectable things to just this test, there should be nothing more after reaching this point IIUC.

normanmaurer: The problem is that we depend on the CountdownLatch to tel us at what point we try to do the write that should fail and client / sever use different EventLoops so there is no good way without schedule the countDown() in the future

Issue #7901, opened 2018-04-30
Netty Client concurrent issue when close FixedChannelPool


caorong:
I use Netty as A tcp Client, use FixedChannelPool as channelPool.

when Close client，some time will cause follow exception

### Expected behavior

no exception

### Actual behavior

```text
[xxxxxx]2018-04-03 14:34:56 04 [io.netty.util.concurrent.DefaultPromise.rejectedExecution]-[ERROR] Failed to submit a listener notification task. Event loop shut down?
 java.util.concurrent.RejectedExecutionException: event executor terminated

        at io.netty.util.concurrent.SingleThreadEventExecutor.reject(SingleThreadEventExecutor.java:821)
        at io.netty.util.concurrent.SingleThreadEventExecutor.offerTask(SingleThreadEventExecutor.java:327)
        at io.netty.util.concurrent.SingleThreadEventExecutor.addTask(SingleThreadEventExecutor.java:320)
        at io.netty.util.concurrent.SingleThreadEventExecutor.execute(SingleThreadEventExecutor.java:746)
        at io.netty.util.concurrent.DefaultPromise.safeExecute(DefaultPromise.java:760)
        at io.netty.util.concurrent.DefaultPromise.notifyListeners(DefaultPromise.java:428)
        at io.netty.util.concurrent.DefaultPromise.setFailure(DefaultPromise.java:113)
        at io.netty.channel.DefaultChannelPromise.setFailure(DefaultChannelPromise.java:87)
        at io.netty.channel.AbstractChannelHandlerContext.safeExecute(AbstractChannelHandlerContext.java:1010)
        at io.netty.channel.AbstractChannelHandlerContext.close(AbstractChannelHandlerContext.java:610)
        at io.netty.channel.AbstractChannelHandlerContext.close(AbstractChannelHandlerContext.java:465)
        at io.netty.channel.DefaultChannelPipeline.close(DefaultChannelPipeline.java:972)
        at io.netty.channel.AbstractChannel.close(AbstractChannel.java:234)
        at io.netty.channel.pool.SimpleChannelPool.close(SimpleChannelPool.java:397)
        at io.netty.channel.pool.FixedChannelPool.access$1201(FixedChannelPool.java:38)
        at io.netty.channel.pool.FixedChannelPool$5.run(FixedChannelPool.java:458)
        at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:163)
[...]

PR #7927, opened 2018-05-11
Fixed|SimpleChannelPool.close() should only return after complete.

normanmaurer:
Motivation:

We need to ensure we only return from close() after all work is done as otherwise we may close the EventExecutor before we dispatched everything.

Modifications:

Correctly wait on operations to complete before return.

Result:

Fixes https://github.com/netty/netty/issues/7901.


normanmaurer:
@trustin ping

PR #7929, opened 2018-05-12
Only use jboss-marshalling when using Java 10

normanmaurer:
Motivation:

cff87de44cfaa4544360715dc83e5ad1aa2c0bd7 updated jboss-marshalling to 2.0.5.Final but this broke the ability to run tests with Java 7.

Modifications:

Only use 2.0.5.Final if compiled against Java 10 (as before 1.4.x works fine).

Result:

Be able to run tests with Java 7 on the CI.


normanmaurer:
Let me just merge this so the CI is green again for the job

PR #7932, opened 2018-05-14
pom.xml: adding suuport for AARCH64 architecture in

ossdev07:
Netty/Transport/Native/Epoll

AARCH 64 architecture support was missing in the exiting Netty-epoll code.
Now the Netty/Transport/Native/Epoll project can be used in aarch64 platform.

Motivation:

Modification:

Result:

Fixes #<GitHub issue number>. 

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -602,10 +602,11 @@
                 </requireMavenVersion>
                 <requireProperty>
                   <regexMessage>
-                    x86_64 JDK must be used.
+                    x86_64/aarch64 JDK must be used.
                   </regexMessage>
                   <property>os.detected.arch</property>
                   <regex>^x86_64$</regex>
+                  <regex>^aarch_64$</regex>
                 </requireProperty>
               </rules>
             </configuration>


normanmaurer:
@ossdev07 seems like this broke the enforcer plugin https://garage.netty.io/teamcity/viewLog.html?buildId=30251&buildTypeId=pulls_netty&guest=1

pom.xml
```diff
@@ -602,10 +602,11 @@
                 </requireMavenVersion>
                 <requireProperty>
                   <regexMessage>
-                    x86_64 JDK must be used.
+                    x86_64/aarch64 JDK must be used.
                   </regexMessage>
                   <property>os.detected.arch</property>
                   <regex>^x86_64$</regex>
+                  <regex>^aarch_64$</regex>
```
normanmaurer: I guess this needs to be:

`<regex>^(x86_64)|(aarch_64)$</regex>

johnou: or `^(x86_64|aarch_64)$`

PR #7934, opened 2018-05-14
Update the comment of io.netty.buffer.PoolChunk.java

zekaryu:
Motivation:

When I read the source code, I found that the comment of PoolChunk is out of date. In old versions of PoolChunk.java, the depth_of_id and x are combined as a short value  and stored in short[] memoryMap, but now the type of memoryMap is changed to byte[] to store depth_of_id , and byte[] depthMap is added to store x , so the description about memoryMap in the comment may confuses readers.

Modifications:

update the last passage of the comment of the PoolChunk class.

Result:

No change to any source code , just update


normanmaurer:
@zekaryu thanks!

zekaryu:
@normanmaurer haha, hope I can contribute more in the future!

normanmaurer:
@zekaryu hope as well =P

PR #7937, opened 2018-05-15
Avoid implicit allocations in Http2FrameLogger when logging is disabled

njhill:
Motivation:

Integer autoboxing in `Http2FrameLogger` (and possibly also the varargs arrays) showed non-negligible CPU and garbage contribution when profiling a gRPC service. grpc-java currently hardcodes use of `Http2FrameLogger`, set at `DEBUG` level.

Modifications:

Wrap offending log statements in conditional blocks.

Result:

Garbage won't be produced by `Http2FrameLogger` when set to a disabled logging level.


normanmaurer:
@njhill can you please sign our icla and let me know once done:

https://netty.io/s/icla

njhill:
Hi @normanmaurer I already did it before opening the PR.

normanmaurer:
@njhill thanks!

Issue #7935, opened 2018-05-14
Can't disable SSLv2 in SSLEngine


dschobel:
### Expected behavior

`SSLContext` / `SSLEngine` should allow applications to opt-out of SSLv2.

### Actual behavior

`SSLEngine` enables SSLv2.

### Steps to reproduce

```
import io.netty.handler.ssl.SslContextBuilder
import io.netty.buffer.PooledByteBufAllocator.DEFAULT
import com.twitter.util.security.{PrivateKeyFile, X509CertificateFile} // available at https://github.com/twitter/util/blob/develop/util-security/src/main/scala/com/twitter/util/security
import java.io.File
 
 
val enabled = List("TLSv1.2")
val certFile = "/var/folders/6q/3x79z17d2g55x6c5z0t4s9040000gn/T/svc-test-server.cert7047932748554435997.pem"
val keyFile = "/var/folders/6q/3x79z17d2g55x6c5z0t4s9040000gn/T/svc-test-server.pkcs85450648075735221716.pem"
 
val key = new PrivateKeyFile(new File(keyFile)).readPrivateKey().get
val cert = new X509CertificateFile(new File(certFile)).readX509Certificate().get
 
val builder = SslContextBuilder.forServer(key, cert)
val context = builder.build()
 
val engine = context.newEngine(DEFAULT)
engine.setUseClientMode(false)
 
engine.getEnabledProtocols // res1: Array[String] = Array(SSLv2Hello, TLSv1, TLSv1.1, TLSv1.2, SSLv2)
engine.setEnabledProtocols(enabled.toArray)
engine.getEnabledProtocols // res3: Array[String] = Array(SSLv2Hello, TLSv1.2, SSLv2)  <--- why is SSLv2 enabled?
```

### Netty version
4.1.25.Final

### JVM version (e.g. `java -version`)
1.8.0_45-b14

### OS version (e.g. `uname -a`)
OS X 10.13.4

normanmaurer:
@dschobel do you use the native impl or the JDK impl ? Also if you use the native impl what flavour of it :) ?

dschobel:
this is using netty-tcnative-boringssl-static version 2.0.8.Final

normanmaurer:
@dschobel interesting...  can you also verify with netty-tcnative (non static one) ? I ill need to dig into it a bit more I guess .

dschobel:
same issue with netty-tcnative.

[...]

PR #7940, opened 2018-05-15
Correctly detect if protocol is enabled when using netty-tcnative

normanmaurer:
Motivation:

We sometimes did not correctly detect when a protocol is not enabled when using netty-tcnative as we did not take into account when the option flag is 0 (as for example in BoringSSL for SSLv2).

Modifications:

- Correctly take an option flag with 0 into account.
- Add unit tests.

Result:

Fixes https://github.com/netty/netty/issues/7935.

### DIFF
--- a/handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
@@ -44,6 +44,7 @@ import io.netty.util.internal.EmptyArrays;
 import io.netty.util.internal.PlatformDependent;
 import io.netty.util.internal.StringUtil;
 import org.junit.After;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.mockito.ArgumentCaptor;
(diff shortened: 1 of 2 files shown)


normanmaurer:
@dschobel PTAL

handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
```diff
@@ -2355,4 +2362,47 @@ public void testWrapDoesNotZeroOutSrc() throws Exception {
...
+        SSLEngine server = ctx.newEngine(UnpooledByteBufAllocator.DEFAULT);
+
+        try {
+            Set<String> supported = new HashSet<String>(Arrays.asList(server.getSupportedProtocols()));
+            if (supported.contains(protocol)) {
+                server.setEnabledProtocols(server.getSupportedProtocols());
+                Assert.assertEquals(supported, new HashSet<String>(Arrays.asList(server.getSupportedProtocols())));
+
+                for (String disabled: disabledProtocols) {
+                    supported.remove(disabled);
```
carl-mastrangelo: supported.removeAll(disabledProtocols)

?

normanmaurer: does not work as disabledProtocols is `String[]` and not a `Collection`

PR #7941, opened 2018-05-15
Correctly clear the error stack in all cases when using ReferenceCoun…

normanmaurer:
…tedOpenSslEngine.

Motivation:

We missed to correctly clear the error stack in one case when using the ReferenceCountedOpenSslEngine. Because of this it was possible to pick up an error on an unrelated operation.

Modifications:

- Correctly clear the stack
- Add verification of empty error stack to tests.

Result:

Not possible to observe unrelated errors.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -1093,8 +1093,6 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
     }
 
     private SSLEngineResult sslReadErrorResult(int err, int bytesConsumed, int bytesProduced) throws SSLException {
-        String errStr = SSL.getErrorString(err);
-
         // Check if we have a pending handshakeException and if so see if we need to consume all pending data from the
         // BIO first or can just shutdown and throw it now.
         // This is needed so we ensure close_notify etc is correctly send to the remote peer.
@@ -1103,11 +1101,14 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
             if (handshakeException == null && handshakeState != HandshakeState.FINISHED) {
                 // we seems to have data left that needs to be transfered and so the user needs
                 // call wrap(...). Store the error so we can pick it up later.
-                handshakeException = new SSLHandshakeException(errStr);
+                handshakeException = new SSLHandshakeException(SSL.getErrorString(err));
             }
+            // We need to clear all errors so we not pick up anything that was left on the stack on the next
+            // operation. Note that shutdownWithError(...) will cleanup the stack as well so its only needed here.
+            SSL.clearError();
             return new SSLEngineResult(OK, NEED_WRAP, bytesConsumed, bytesProduced);
         }
-        throw shutdownWithError("SSL_read", errStr);
+        throw shutdownWithError("SSL_read", SSL.getErrorString(err));
     }
 
     private void closeAll() throws SSLException {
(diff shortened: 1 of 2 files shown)


handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1093,8 +1093,6 @@ public final SSLEngineResult unwrap(
     }
 
     private SSLEngineResult sslReadErrorResult(int err, int bytesConsumed, int bytesProduced) throws SSLException {
-        String errStr = SSL.getErrorString(err);
```
normanmaurer: This change was just done to eliminate the need to create it when its not needed.

handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
```diff
@@ -82,6 +84,12 @@ public static void checkOpenSsl() {
         assumeTrue(OpenSsl.isAvailable());
     }
 
+    @Override
+    public void tearDown() throws InterruptedException {
+        super.tearDown();
+        Assert.assertEquals("SSL error stack not correctly consumed", 0, SSL.getLastErrorNumber());
```
ddossot: nit: static import `Assert.assertEquals`

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1103,11 +1101,14 @@ private SSLEngineResult sslReadErrorResult(int err, int bytesConsumed, int bytes
             if (handshakeException == null && handshakeState != HandshakeState.FINISHED) {
                 // we seems to have data left that needs to be transfered and so the user needs
                 // call wrap(...). Store the error so we can pick it up later.
-                handshakeException = new SSLHandshakeException(errStr);
+                handshakeException = new SSLHandshakeException(SSL.getErrorString(err));
             }
+            // We need to clear all errors so we not pick up anything that was left on the stack on the next
+            // operation. Note that shutdownWithError(...) will cleanup the stack as well so its only needed here.
+            SSL.clearError();
```
johnou: @normanmaurer without this what behaviour did you observe?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1093,8 +1093,6 @@ public final SSLEngineResult unwrap(
     }
 
     private SSLEngineResult sslReadErrorResult(int err, int bytesConsumed, int bytesProduced) throws SSLException {
```
Scottmitch: future improvement: we don't actually use `err` for anything besides getting the string. We could add a method to netty-tcnative that is `SSL.getLastErrorString()` to save a JNI call. Looks like this is the only place that `getLastErrorNumber` is used in non-unit test code too.

normanmaurer: 👍

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: basically whatever test that ran after `OpenSslEngineTest.testMultipleRecordsInOneBufferWithNonZeroPositionJDKCompatabilityModeOff` failed when using BoringSSL. Thats also why I added asserts now to ensure the error stack is empty when we teardown.

PR #7943, opened 2018-05-15
Verify error stack is empty after each operation when using Reference…

normanmaurer:
…CountedOpenSslEngine.

Motivation:

https://github.com/netty/netty/pull/7941 proved that its easy to not correctly clear the error stack sometimes. We should do carefully test this.

Modifications:

Add a new SSLEngine wrapper that is used during tests, which verifies that the error stack is empty after each method call.

Result:

Better testing.

### DIFF
--- a/handler/src/test/java/io/netty/handler/ssl/OpenSslErrorStackAssertSSLEngine.java
+++ b/handler/src/test/java/io/netty/handler/ssl/OpenSslErrorStackAssertSSLEngine.java
@@ -1,0 +1,440 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


normanmaurer:
As a followup for https://github.com/netty/netty/pull/7941

handler/src/test/java/io/netty/handler/ssl/OpenSslErrorStackAssertSSLEngine.java
```diff
@@ -0,0 +1,440 @@
...
+    }
+
+    @Override
+    public boolean release() {
+        return getWrappedEngine().release();
+    }
+
+    @Override
+    public boolean release(int decrement) {
+        return getWrappedEngine().release();
```
carl-mastrangelo: missing param?

normanmaurer: ups... will fix it quickly.

normanmaurer: https://github.com/netty/netty/commit/47985c11c1d9f11b94018539108ed65f0c6a32ab

PR #7945, opened 2018-05-16
Correctly calculate and respect if we can correctly fullfil wrap for …

normanmaurer:
…alerts.

Motivation:

We previously did not correctly take into account when we could not wrap (and so produce) the full SSL record with an alert when the SSLEngine was closed.

There are two problems here:

- If we call wrap(...) with an empty dst buffer after closeOutbound() was called we will not notify the user if we could not store the whole SSLRecord into the dst buffer and so we may produce incomplete SSLRecords
- If we call wrap(...) with an empty dst buffer we need to ensure we can at least hold an alert.

Modifications:

Add unit test which failed before.

Result:

Correctly handle the case when the dst buffer is not big enough and and alert needs to be produced.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -607,8 +607,17 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
 
                 int bioLengthBefore = SSL.bioLengthByteBuffer(networkBIO);
 
-                // Explicit use outboundClosed as we want to drain any bytes that are still present.
+                // Explicitly use outboundClosed as we want to drain any bytes that are still present.
                 if (outboundClosed) {
+                    // If the outbound was closed we want to ensure we can produce the alert to the destination buffer.
+                    // This is true even if we not using jdkCompatibilityMode.
+                    //
+                    // We use a plaintextLength of 2 as we at least want to have an alert fit into it.
+                    // https://tools.ietf.org/html/rfc5246#section-7.2
+                    if (!isBytesAvailableEnoughForWrap(dst.remaining(), 2, 1)) {
+                        return new SSLEngineResult(BUFFER_OVERFLOW, getHandshakeStatus(), 0, 0);
+                    }
+
                     // There is something left to drain.
                     // See https://github.com/netty/netty/issues/6260
                     bytesProduced = SSL.bioFlushByteBuffer(networkBIO);
--- a/handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/OpenSslEngineTest.java
@@ -23,7 +23,6 @@ import io.netty.handler.ssl.util.InsecureTrustManagerFactory;
 import io.netty.handler.ssl.util.SelfSignedCertificate;
 import io.netty.internal.tcnative.SSL;
 import io.netty.util.internal.PlatformDependent;
-import org.junit.Assert;
 import org.junit.Assume;
 import org.junit.BeforeClass;
 import org.junit.Test;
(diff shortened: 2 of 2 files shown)


normanmaurer:
Please note without this patch we also had a few more test-failures when using `BoringSSL` as here we try to be more exact when calculate the needed length of the dst buffer (BoringSSL provides an API that we use via netty-tcnative).

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -534,7 +534,8 @@ final int maxEncryptedPacketLength0() {
      * via other synchronized blocks.
      */
     final int calculateMaxLengthForWrap(int plaintextLength, int numComponents) {
-        return (int) min(maxWrapBufferSize, plaintextLength + (long) maxWrapOverhead * numComponents);
+        // We assume as minimum 2 bytes for the plaintextLength so we can fit an alert into it.
```
Scottmitch: consider adding a link to https://tools.ietf.org/html/rfc5246#section-7.2 which defines the 2 bytes size.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -609,6 +610,12 @@ public final SSLEngineResult wrap(
 
                 // Explicit use outboundClosed as we want to drain any bytes that are still present.
```
Scottmitch: `Explicit` -> `Explicitly`

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
Scottmitch: would it be better to just use the minimum size for each wrap operation ... and then if there is an alert return a `BUFFER_OVERFLOW`? This would manifest in the following ways:

- if there is a non-fatal alert in the middle of writing data, we will try to write data, overflow, and have to allocate another buffer to finish the wrap operation.
- if there is an alert (fatal or not) after we finished writing data we will pick it up on the next wrap operation, or in the "outboundClose" case you added below

This suggestion is operating under the assumption that alerts are relatively rare when compared to moving data back and forth. We may end up having to allocate an extra buffer if we overflow, but again if this is assumed to be rare and will result in a few bytes saved per wrap operation. Buffer pooling may make the extra few bytes less of an issue, but not everyone can use pooling.

Which reminds me this allocation is completely internal to the `SslHandler` so we could potentially always used a pooled allocator here.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -609,6 +610,12 @@ public final SSLEngineResult wrap(
...
                 if (outboundClosed) {
+                    // If the outbound was closed we want to ensure we can produce the alert to the destination buffer.
+                    // This is true even if we not using jdkCompatibilityMode.
+                    if (!isBytesAvailableEnoughForWrap(dst.remaining(), 0, 1)) {
```
Scottmitch: If the intention here is to check if there is enough space for an alert can we explicitly use `2` instead of `0`? Also if we have `2` in multiple places lets make a define for it.

normanmaurer: yep thats basically it let me use 2 (can't harm)

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: like discussed offline I think we should just use `2` for now and then maybe investigate in a followup if there is a better way. I addressed your other comments tho

Scottmitch: discussed offline. My main concern with this approach is mixing the application and non-application path with a special case for ALERT data, but there maybe other non-application data that OpenSSL wants to send that is non-ALERT data which maybe larger, or we may just "get lucky" because of this extra fluff. The API for OpenSSL to communicate "I want to write non-application data, and you didn't give me enough space" seems to be call SSL_write and look for the `SSL_WANT_WRITE` error code [1]. I would prefer to rely upon this API if possible which means we may need to investigate prematurely existing the wrap loop for empty buffers [2].

[1] https://github.com/netty/netty/blob/cd7007bae2de23f00d8fa253da9e1cb9ea4e4293/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java#L784-L797
[2] https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java#L717

PR #7946, opened 2018-05-16
Adjust tests to also pass when using BoringSSL

normanmaurer:
Motivation:

Some of the tests failed when using BoringSSL as some protocol / cipher combinations are not supported and it uses a different alert when the cert is not valid yet.

Modification:

- Remove protocol / cipher combos that are not supported by BoringSSL
- Test for different alert when using BoringSSL

Result:

Not test failures when using BoringSSL.


normanmaurer:
Once all the boring related stuff is merged I will add a job to the CI to verify all of this on a regular basis.

PR #7948, opened 2018-05-16
Guard against calling malloc(0) when create ByteBuffer.

normanmaurer:
Motivation:

We did not guard against the case of calling malloc(0) when creating a ByteBuffer without a Cleaner. The problem is that malloc(0) can have different behaviour, it either return a null-pointer or a valid pointer that you can pass to free.

The real problem arise if Unsafe.allocateMemory(0) returns 0 and we use it as the memoryAddress of the ByteBuffer. The problem here is that native libraries test for 0 and handle it as a null-ptr. This is for example true in SSL.bioSetByteBuffer(...) which would throw a NPE when 0 is used as memoryAddress and so produced errors during SSL usage.

Modifications:

- Always allocate 1 byte as minimum (even if we ask for an empty buffer).
- Add unit test.

Result:

No more errors possible because of malloc(0).

### DIFF
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent0.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent0.java
@@ -427,7 +427,10 @@ final class PlatformDependent0 {
     }
 
     static ByteBuffer allocateDirectNoCleaner(int capacity) {
-        return newDirectBuffer(UNSAFE.allocateMemory(capacity), capacity);
+        // Calling malloc with capacity of 0 may return a null ptr or a memory address that can be used.
+        // Just use 1 to make it safe to use in all cases:
+        // See: http://pubs.opengroup.org/onlinepubs/009695399/functions/malloc.html
+        return newDirectBuffer(UNSAFE.allocateMemory(Math.max(1, capacity)), capacity);
     }
 
     static boolean hasAllocateArrayMethod() {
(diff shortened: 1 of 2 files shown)


common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -427,7 +427,10 @@ static ByteBuffer reallocateDirectNoCleaner(ByteBuffer buffer, int capacity) {
     }
 
     static ByteBuffer allocateDirectNoCleaner(int capacity) {
-        return newDirectBuffer(UNSAFE.allocateMemory(capacity), capacity);
+        // Calling malloc with capacity of 0 may return a null ptr or a memory address that can be used.
+        // Just use 1 to make it safe to use in all cases:
+        // See: http://pubs.opengroup.org/onlinepubs/009695399/functions/malloc.html
+        return newDirectBuffer(UNSAFE.allocateMemory(Math.max(1, capacity)), capacity);
```
carl-mastrangelo: won't this result in a buffer that has a longer length than it's capacity?

normanmaurer: @carl-mastrangelo generally speaking yes... that said `capacity()` will return the original `capacity()` so its not a problem.

Same is true in OpenJDK as well:
http://hg.openjdk.java.net/jdk-updates/jdk10u/file/523e40ed3fc7/src/java.base/share/classes/java/nio/Direct-X-Buffer.java.template#l122

normanmaurer:
Also finish setup a CI job with BoringSSL which currently fails because this change is needed: 

https://garage.netty.io/teamcity/viewLog.html?buildId=30395&buildTypeId=netty_BuildBoringsslStatic

This will be green once this is merged.

PR #7950, opened 2018-05-18
Add tests for the Conscrypt based SSLEngine.

normanmaurer:
Motivation:

We currently have only interopt tests for Conscrypt, we should also have non-interopt tests.

Modifications:

Add ConscryptSslEngineTest

Result:

More tests

### DIFF
--- a/handler/src/test/java/io/netty/handler/ssl/ConscryptSslEngineTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/ConscryptSslEngineTest.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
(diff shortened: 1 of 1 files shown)


handler/src/test/java/io/netty/handler/ssl/ConscryptSslEngineTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+import java.security.Provider;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+import static org.junit.Assume.assumeTrue;
+
+@RunWith(Parameterized.class)
+public class ConscryptSslEngineTest  extends SSLEngineTest {
```
ejona86: nit: duplicate space.

PR #7953, opened 2018-05-18
Don't calculate max direct memory twice in PlatformDependent

njhill:
Motivation:

I'm not sure if trivial changes like this are interesting :-) But I noticed that the `PlatformDependent.maxDirectMemory0()` method is called twice unnecessarily during static initialization (on the default path at least).

Modifications:

Use constant `MAX_DIRECT_MEMORY` already set to the same value instead of calling `maxDirectMemory0()` again.

Result:

A surely imperceivable reduction in operations performed at startup.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent.java
@@ -158,7 +158,7 @@ public final class PlatformDependent {
         } else {
             USE_DIRECT_BUFFER_NO_CLEANER = true;
             if (maxDirectMemory < 0) {
-                maxDirectMemory = maxDirectMemory0();
+                maxDirectMemory = MAX_DIRECT_MEMORY;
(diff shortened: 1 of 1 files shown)


fenik17:
// small offtopic
@normanmaurer 
```
    public static long maxDirectMemory() {
        return MAX_DIRECT_MEMORY;
    }
```
Looks like this method does not respect `io.netty.maxDirectMemory` property. Maybe would be more correct to return `DIRECT_MEMORY_LIMIT` instead of `MAX_DIRECT_MEMORY`?

njhill:
I've added another commit to address @fenik17's observation, can revert it if necessary (could change existing behaviour).

normanmaurer:
@njhill thanks... I cherry-picked the first commit as 06f3574e46e0623a45a93d36dbe7aa3d7455a995. Please open another PR for the other change.

PR #7955, opened 2018-05-18
Propagate pong frames in WebSocketProtocolHandler

### LINKED ISSUE #7954: PongWebSocketFrames are inaccessible in handlers downstream of a WebSocketClientProtocolHandler
This is probably more of a question / request rather than a bug, as it looks like this was a deliberate decision ([source](https://github.com/netty/netty/blob/netty-4.1.25.Final/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketProtocolHandler.java#L32-L35)).

The use-case is a handler being able to assert that a pong frame was received from a Netty-based proxy server.

### Expected behavior

`PongWebSocketFrame`s are visible to a channel handler installed _downstream_ of a `WebSocketClientProtocolHandler`.

### Actual behavior

Sending a `PingWebSocketFrame` from a client to a server results in receipt of a `PongWebSocketFrame` by the client in the `WebSocketClientProtocolHandler`, but this frame is not propagated further along the channel pipeline.

### Steps to reproduce

I've pushed a commit [here](https://github.com/nicktrav/netty/commit/4cae21eb36fe873d24073e730cbda9bb1994741b) which demonstrates the issue.

Check out commit and run:
```bash
$ mvn test -pl codec-http -Dtest=io.netty.handler.codec.http.websocketx.PongFrameIgnoreTest
```

### Netty version

`4.1.25.Final`. Also run against `HEAD` of `4.1` branch @ 987c44388.

```bash
$ java -version
java version "10.0.1" 2018-04-17
Java(TM) SE Runtime Environment 18.3 (build 10.0.1+10)
Java HotSpot(TM) 64-Bit Server VM 18.3 (build 10.0.1+10, mixed mode)
```

```bash
$ uname -a
Darwin nickt.local 17.3.0 Darwin Kernel Version 17.3.0: Thu Nov  9 18:09:22 PST 2017; root:xnu-4570.31.3~1/RELEASE_X86_64 x86_64
```

nicktrav:
Motivation:

Currently, on recipt of a PongWebSocketFrame, the
WebSocketProtocolHandler will drop the frame, rather than passing it
along so it can be referenced by other handlers.

Modifications:

Add boolean field to WebSocketProtocolHandler to indicate whether Pong
frames should be dropped or propagated, defaulting to "true" to preserve
existing functionality.

Add new constructors to the client and server implementations of
WebSocketProtocolHandler that allow for overriding the behavior for the
handling of Pong frames.

Result:

PongWebSocketFrames are passed along the channel, if specified.

Fixes #7954.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketProtocolHandler.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketProtocolHandler.java
@@ -22,6 +22,27 @@ import io.netty.handler.codec.MessageToMessageDecoder;
 import java.util.List;
 
 abstract class WebSocketProtocolHandler extends MessageToMessageDecoder<WebSocketFrame> {
+
+    private final boolean dropPongFrames;
+
+    /**
+     * Creates a new {@link WebSocketProtocolHandler} that will <i>drop</i> {@link PongWebSocketFrame}s.
+     */
+    WebSocketProtocolHandler() {
+        this(true);
+    }
+
+    /**
+     * Creates a new {@link WebSocketProtocolHandler}, given a parameter that determines whether or not to drop {@link
+     * PongWebSocketFrame}s.
+     *
+     * @param dropPongFrames
+     *            {@code true} if {@link PongWebSocketFrame}s should be dropped
+     */
+    WebSocketProtocolHandler(boolean dropPongFrames) {
+        this.dropPongFrames = dropPongFrames;
+    }
+
     @Override
     protected void decode(ChannelHandlerContext ctx, WebSocketFrame frame, List<Object> out) throws Exception {
         if (frame instanceof PingWebSocketFrame) {
@@ -29,8 +50,7 @@ abstract class WebSocketProtocolHandler extends MessageToMessageDecoder<WebSocke
             ctx.channel().writeAndFlush(new PongWebSocketFrame(frame.content()));
             return;
         }
-        if (frame instanceof PongWebSocketFrame) {
-            // Pong frames need to get ignored
+        if (frame instanceof PongWebSocketFrame && dropPongFrames) {
             return;
         }
 
--- a/codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketProtocolHandlerTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketProtocolHandlerTest.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.netty.handler.codec.http.websocketx;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import io.netty.channel.embedded.EmbeddedChannel;
+import io.netty.util.CharsetUtil;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+/**
+ * Tests common, abstract class functionality in {@link WebSocketClientProtocolHandler}.
+ */
+public class WebSocketProtocolHandlerTest {
+
+    @Test
+    public void testPingFrame() {
+        ByteBuf pingData = Unpooled.copiedBuffer("Hello, world", CharsetUtil.UTF_8);
+        EmbeddedChannel channel = new EmbeddedChannel(new WebSocketProtocolHandler() { });
+
+        PingWebSocketFrame inputMessage = new PingWebSocketFrame(pingData);
+        assertFalse(channel.writeInbound(inputMessage)); // the message was not propagated inbound
+
+        // a Pong frame was written to the channel
+        PongWebSocketFrame response = channel.readOutbound();
+        assertEquals(pingData, response.content());
+
+        pingData.release();
+        assertFalse(channel.finish());
+    }
+
+    @Test
+    public void testPongFrameDropFrameFalse() {
+        EmbeddedChannel channel = new EmbeddedChannel(new WebSocketProtocolHandler(false) { });
+
+        PongWebSocketFrame pingResponse = new PongWebSocketFrame();
+        assertTrue(channel.writeInbound(pingResponse));
+
+        assertPropagatedInbound(pingResponse, channel);
+
+        pingResponse.release();
+        assertFalse(channel.finish());
+    }
+
+    @Test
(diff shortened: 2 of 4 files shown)


codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketProtocolHandlerTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+        pingResponse.release();
+    }
+
+    @Test
+    public void testTextFrame() {
+        EmbeddedChannel channel = new EmbeddedChannel(new WebSocketProtocolHandler() {});
+
+        TextWebSocketFrame textFrame = new TextWebSocketFrame();
+        channel.writeInbound(textFrame);
```
normanmaurer: assert return value please

codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketProtocolHandlerTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+        pingData.release();
+    }
+
+    @Test
+    public void testPongFrame() {
+        EmbeddedChannel channel = new EmbeddedChannel(new WebSocketProtocolHandler() {});
+
+        PongWebSocketFrame pingResponse = new PongWebSocketFrame();
+        channel.writeInbound(pingResponse);
```
normanmaurer: assert return value please

codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketProtocolHandlerTest.java
```diff
@@ -0,0 +1,84 @@
...
+public class WebSocketProtocolHandlerTest {
+
+    @Test
+    public void testPingFrame() {
+        ByteBuf pingData = Unpooled.buffer();
+        pingData.writeCharSequence("Hello, world!", StandardCharsets.UTF_8);
+        EmbeddedChannel channel = new EmbeddedChannel(new WebSocketProtocolHandler() {});
+
+        PingWebSocketFrame inputMessage = new PingWebSocketFrame(pingData);
+        channel.writeInbound(inputMessage);
```
normanmaurer: assert return value please

codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketProtocolHandlerTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Tests common, abstract class functionality in {@link WebSocketClientProtocolHandler}.
+ */
+public class WebSocketProtocolHandlerTest {
+
+    @Test
+    public void testPingFrame() {
+        ByteBuf pingData = Unpooled.buffer();
+        pingData.writeCharSequence("Hello, world!", StandardCharsets.UTF_8);
```
normanmaurer: You can use use `Unpooled.copiedBuffer(...)` to simplify this a bit.

codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketProtocolHandlerTest.java
```diff
@@ -0,0 +1,84 @@
...
+    @Test
+    public void testTextFrame() {
+        EmbeddedChannel channel = new EmbeddedChannel(new WebSocketProtocolHandler() {});
+
+        TextWebSocketFrame textFrame = new TextWebSocketFrame();
+        channel.writeInbound(textFrame);
+
+        assertPropagatedInbound(textFrame, channel);
+
+        textFrame.release();
```
normanmaurer: Call `channel.finish()` and assert return value

codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketProtocolHandlerTest.java
```diff
@@ -0,0 +1,84 @@
...
+    @Test
+    public void testPongFrame() {
+        EmbeddedChannel channel = new EmbeddedChannel(new WebSocketProtocolHandler() {});
+
+        PongWebSocketFrame pingResponse = new PongWebSocketFrame();
+        channel.writeInbound(pingResponse);
+
+        assertPropagatedInbound(pingResponse, channel);
+
+        pingResponse.release();
```
normanmaurer: Call `channel.finish()` and assert return value

codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketProtocolHandlerTest.java
```diff
@@ -0,0 +1,84 @@
...
+        channel.writeInbound(inputMessage);
+
+        // the message was not propageted inbound
+        assertNull(channel.readInbound());
+
+        // a Pong frame was written to the channel
+        PongWebSocketFrame response = channel.readOutbound();
+        assertEquals(pingData, response.content());
+
+        pingData.release();
```
normanmaurer: Call `channel.finish()` and assert return value

nicktrav:
@normanmaurer - ICLA signed, feedback addressed (thanks!). PTAL.

codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketProtocolHandlerTest.java
```diff
@@ -0,0 +1,82 @@
...
+import static org.junit.Assert.*;
+
+/**
+ * Tests common, abstract class functionality in {@link WebSocketClientProtocolHandler}.
+ */
+public class WebSocketProtocolHandlerTest {
+
+    @Test
+    public void testPingFrame() {
+        ByteBuf pingData = Unpooled.copiedBuffer("Hello, world", StandardCharsets.UTF_8);
```
normanmaurer: Use `CharsetUtil.UTF_8`

normanmaurer:
@trustin thats a good point... @nicktrav can you do this ?

nicktrav:
@normanmaurer @trustin - have updated the PR, adding a param on the abstract class that defaults to "true" (preserving existing functionality).

I added some new constructors to the subclasses to allow this value to be overridden. It feels a little gross with the constructor telescoping, but I'm not sure if there's any other way, without breaking compatibility. We could probably consider using a builder here in the next major release, where a lot of this WS code is scheduled to be cleaned up, IIUC?

normanmaurer:
@nicktrav thanks a lot!

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketProtocolHandler.java
```diff
@@ -22,15 +22,35 @@
...
     @Override
     protected void decode(ChannelHandlerContext ctx, WebSocketFrame frame, List<Object> out) throws Exception {
         if (frame instanceof PingWebSocketFrame) {
             frame.content().retain();
             ctx.channel().writeAndFlush(new PongWebSocketFrame(frame.content()));
             return;
         }
-        if (frame instanceof PongWebSocketFrame) {
-            // Pong frames need to get ignored
+        if (frame instanceof PongWebSocketFrame && dropPongFrames) {
```
johnou: @normanmaurer I would suggest flipping the conditions, boolean is cheaper than instance of.

johnou: :+1: makes sense

PR #7961, opened 2018-05-23
Read until all data is consumed when EOF is detected even if readPend…

normanmaurer:
…ing is false and auto-read is disabled.

Motivation:

We should better always notify the user of EOF even if the user did not request any data as otherwise we may never be notified when the remote peer closes the connection. This should be ok as the amount of extra data we may read and so fire through the pipeline is limited by SO_RECVBUF.

Modifications:

Always drain the socket when EOF is detected.

Result:

No risk for the user to be not notified of EOF.

### DIFF
--- a/transport-native-kqueue/src/main/java/io/netty/channel/kqueue/AbstractKQueueChannel.java
+++ b/transport-native-kqueue/src/main/java/io/netty/channel/kqueue/AbstractKQueueChannel.java
@@ -406,7 +406,7 @@ abstract class AbstractKQueueChannel extends AbstractChannel implements UnixChan
         final void readReadyFinally(ChannelConfig config) {
             maybeMoreDataToRead = allocHandle.maybeMoreDataToRead();
 
-            if ((readPending || allocHandle.isReadEOF()) && maybeMoreDataToRead) {
+            if (allocHandle.isReadEOF() || (readPending && maybeMoreDataToRead)) {
                 // trigger a read again as there may be something left to read and because of ET we
                 // will not get notified again until we read everything from the socket
                 //
--- a/transport-native-kqueue/src/main/java/io/netty/channel/kqueue/AbstractKQueueChannel.java
+++ b/transport-native-kqueue/src/main/java/io/netty/channel/kqueue/AbstractKQueueChannel.java
@@ -405,15 +405,8 @@ abstract class AbstractKQueueChannel extends AbstractChannel implements UnixChan
 
         final void readReadyFinally(ChannelConfig config) {
             maybeMoreDataToRead = allocHandle.maybeMoreDataToRead();
-            // Check if there is a readPending which was not processed yet.
-            // This could be for two reasons:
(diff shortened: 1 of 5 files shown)


normanmaurer:
We could also allow this to be configured by an `ChannelOption`. That said I am not sure we need. WDYT ?

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/AbstractKQueueChannel.java
```diff
@@ -405,15 +405,8 @@ final void readReady(long numberBytesPending) {
...
-            // This could be for two reasons:
-            // * The user called Channel.read() or ChannelHandlerContext.read() in channelRead(...) method
-            // * The user called Channel.read() or ChannelHandlerContext.read() in channelReadComplete(...) method
-            //
-            // See https://github.com/netty/netty/issues/2254
-            if (!readPending && !config.isAutoRead()) {
-                clearReadFilter0();
-            } else if (readPending && maybeMoreDataToRead) {
+
+            if ((readPending || allocHandle.isReadEOF()) &&  maybeMoreDataToRead) {
```
carl-mastrangelo: micronit: extra space after &&

normanmaurer: fixed

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/AbstractKQueueChannel.java
```diff
@@ -405,15 +405,8 @@ final void readReady(long numberBytesPending) {
...
-            // This could be for two reasons:
-            // * The user called Channel.read() or ChannelHandlerContext.read() in channelRead(...) method
-            // * The user called Channel.read() or ChannelHandlerContext.read() in channelReadComplete(...) method
-            //
-            // See https://github.com/netty/netty/issues/2254
-            if (!readPending && !config.isAutoRead()) {
-                clearReadFilter0();
-            } else if (readPending && maybeMoreDataToRead) {
+
+            if ((readPending || allocHandle.isReadEOF()) && maybeMoreDataToRead) {
```
Scottmitch: Should this conditional instead be the following:

`(allocHandle.isReadEOF() || (readPending && maybeMoreDataToRead))`

This says: "if EOF we always drain all data, otherwise we only read if there maybe data pending and a read is requested"

This would allow us to remove the EOF check from the `maybeMoreDataToRead()` method ... which is now duplicated inside/outside the method. We can update the comments in the `maybeMoreDataToRead()` method to say "it is assumed EOF is handled externally" (in addition to other comments). Since we have to expose `isReadEOF()` now (and we already do this for EPOLL) it seems to make sense.

Scottmitch:
Thanks for taking care @normanmaurer !

normanmaurer:
@Scottmitch sorry it took so long :(

Issue #7964, opened 2018-05-24
DefaultHttpResponse does not implement equals and hashCode to cover response status


ewie:
### Expected behavior
Two DefaultHttpResponse instances with different response status should not be considered equal. Headers and HTTP version are already covered in DefaultHttpMessage. DefaultFullHttpResponse also covers content and trailing headers. I would assume the response status is considered as well.

### Actual behavior
Response status is ignored by equals and hashCode. DefaultHttpResponse does not override those methods.

### Steps to reproduce
Run example below.

### Minimal yet complete reproducer code
```java
import io.netty.handler.codec.http.DefaultHttpResponse;
import io.netty.handler.codec.http.HttpResponseStatus;
import io.netty.handler.codec.http.HttpVersion;

class ResponseStatusEquality
{
  public static void main(String... args)
  {
    var ok = new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.OK);
    var notFound = new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.NOT_FOUND);
    if (ok.equals(notFound)) {
      throw new AssertionError(); // should not throw!
    }
  }
}
```

### Netty version
4.1.25.Final

### JVM version (e.g. `java -version`)
openjdk version "10.0.1" 2018-04-17
OpenJDK Runtime Environment (build 10.0.1+10)
OpenJDK 64-Bit Server VM (build 10.0.1+10, mixed mode)

### OS version (e.g. `uname -a`)
Linux 4.16.10

normanmaurer:
sounds like a bug to me... looking now

normanmaurer:
@ewie thanks for reporting, fixed by https://github.com/netty/netty/pull/7965

PR #7965, opened 2018-05-24
Correctly take status into account when compare DefaultHttpResponse

normanmaurer:
Motivation:

DefaultHttpResponse did not respect its status when compute the hashCode and check for equality.

Modifications:

Correctly implement hashCode and equals

Result:

Fixes https://github.com/netty/netty/issues/7964.

### DIFF
--- a/codec-http/src/test/java/io/netty/handler/codec/http/DefaultHttpResponseTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/DefaultHttpResponseTest.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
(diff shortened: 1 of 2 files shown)


codec-http/src/test/java/io/netty/handler/codec/http/DefaultHttpResponseTest.java
```diff
@@ -0,0 +1,36 @@
...
+        HttpResponse ok = new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.OK);
+        HttpResponse notFound = new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.NOT_FOUND);
+        Assert.assertNotEquals(ok, notFound);
+    }
+
+    @Test
+    public void testEquals() {
+        HttpResponse ok = new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.OK);
+        HttpResponse ok2 = new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.OK);
+        Assert.assertEquals(ok, ok2);
```
ejona86: Assert the hash codes are equal as well.

PR #7967, opened 2018-05-24
AUTO_CLOSE should not be marked as deprecated.

normanmaurer:
Motivation:

A long time ago we deprecated AUTO_CLOSE but it turned out this feature is still useful.

Modifications:

- Remove `@Deprecated` tag for AUTO_CLOSE, setAutoClose(...) and isAutoClose(...)
- Fix javadocs on ChannelConfig to correctly tell the default value of AUTO_CLOSE.

Result:

Less warnings.


normanmaurer:
@Scottmitch done... adjusted the commit message.

Issue #7970, opened 2018-05-24
NioEventLoop has too low MAX_SCHEDULED_DAYS


carl-mastrangelo:
### Expected behavior
Create a scheduled task with some reasonable, arbitrary delay.

### Actual behavior

`NioEventLoop.validateScheduled`   throws an illegal argument exception when passing a delay of greater than 3 years. 

### Netty version
4.1.25

A user of gRPC noticed that when passing a really big deadline, the above exception was thrown.  The deadline in question was 100 years, which, while high, was previously allowed.   [EpollEventLoop](https://github.com/netty/netty/blob/4.1/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java) allows about 31 years, but this actually appears wrong.    The man page says the number of /nanoseconds/ must be less than 999999999, not the number of seconds, as implemented.

I think a more reasonable number would be the Long.MAX_VALUE of nanos, which is about 292 years.  If the underlying poller can't support, I think clamping the value is right.   The ScheduledFuture can keep its deadline, there would just be a spurious wake up. 

cc: @ejona86

normanmaurer:
This sounds like a better solution then throwing... let me work on a fix

PR #7972, opened 2018-05-25
Allow to schedule tasks up to Long.MAX_VALUE

normanmaurer:
Motivation:

We should allow to schedule tasks with a delay up to Long.MAX_VALUE as we did pre 4.1.25.Final.

Modifications:

Just ensure we not overflow and put the correct max limits in place when schedule a timer. At worse we will get a wakeup to early and then schedule a new timeout.

Result:

Fixes https://github.com/netty/netty/issues/7970.

### DIFF
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
@@ -37,7 +37,6 @@ import java.util.Collection;
 import java.util.Queue;
 import java.util.concurrent.Callable;
 import java.util.concurrent.Executor;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
 import static java.lang.Math.min;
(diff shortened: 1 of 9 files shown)


transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
```diff
@@ -237,7 +236,7 @@ private int epollWait(boolean oldWakeup) throws IOException {
         long totalDelay = delayNanos(System.nanoTime());
         int delaySeconds = (int) min(totalDelay / 1000000000L, Integer.MAX_VALUE);
         return Native.epollWait(epollFd, events, timerFd, delaySeconds,
-                (int) min(totalDelay - delaySeconds * 1000000000L, Integer.MAX_VALUE));
+                (int) min(MAX_SCHEDULED_TIMERFD_NS, min(totalDelay - delaySeconds * 1000000000L, Integer.MAX_VALUE)));
```
carl-mastrangelo: I think this can be simplified to drop Integer.MAX_VALUE?

normanmaurer: Sorry not sure I understand... what exactly you propose ?

carl-mastrangelo:
```java
return Native.epollWait(epollFd, events, timerFd, delaySeconds,
    (int) min(MAX_SCHEDULED_TIMERFD_NS, totalDelay - delaySeconds * 1000000000L));
```

normanmaurer: ah yeah... let me do this

normanmaurer: @carl-mastrangelo PTAL again

normanmaurer:
@carl-mastrangelo can you validate this one with GRPC as well ?

carl-mastrangelo:
@normanmaurer I did try to test this out, sadly its pretty time consuming to bring netty into Google if it isn't a released artifact.  I think code review is good enough here.

normanmaurer:
@carl-mastrangelo ok... 😢

PR #7974, opened 2018-05-26
Run the WebSocketClientHandshaker08Test on the 08 implementation

nicktrav:
Motivation:

The websockets abstract test suite does not run against the 08
implementation in the 08 version of the test suite.

Modifications:

Update the WebSocketClientHandshaker08Test to instantiate a new
WebSocketClientHandshaker08 rather than an 07 handshaker.

Result:

The WebSocketClientHandshaker08Test now tests the 08 implementation.


normanmaurer:
Thanks a lot

PR #7975, opened 2018-05-26
Set (and override) websocket handshake headers after custom headers

### LINKED ISSUE #7973: WebsocketClientHandshaker can fail if multiple X-Websocket-Key headers are present on the upgrade request
### Expected behavior

A websockets handshake should not fail if a `X-Websockets-Key` header exists in the custom headers passed to the handshaker.

### Actual behavior

The handshake fails if the server picks the incorrect key on which to compute its challenge response (as opposed to the key that is set by the `WebsocketsClientHandshaker`).

### Steps to reproduce

Use the `WebsocketsClientHandshaker` to send a request to a websockets server along with passing custom headers that also contain an `X-Websockets-Key` header.

If the server happens to pick the _last_ `X-Websockets-Key` in the request headers on which to compute the `X-Websockets-Accept` value, the client handshake will fail with the following, as the `X-Websockets-Accept` value it expects was calculated on the _first_ `X-Websockets-Key`:

```
20:04:35.727 [nioEventLoopGroup-3-1] WARN  i.n.channel.DefaultChannelPipeline - An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.
io.netty.handler.codec.http.websocketx.WebSocketHandshakeException: Invalid challenge. Actual: w7cyq/iodDhF6MfTw9LCsDqUyBQ=. Expected: +RND8EYF2d5MkuthghnnAyg9DN0=
	at io.netty.handler.codec.http.websocketx.WebSocketClientHandshaker13.verify(WebSocketClientHandshaker13.java:223)
	at io.netty.handler.codec.http.websocketx.WebSocketClientHandshaker.finishHandshake(WebSocketClientHandshaker.java:216)
	at io.netty.handler.codec.http.websocketx.WebSocketClientProtocolHandshakeHandler.channelRead(WebSocketClientProtocolHandshakeHandler.java:57)
... snip ...
```

Reproducer [here](https://github.com/nicktrav/netty/commit/219d1dd5c00a2d2d92ef8224392f88a2a386668c).

### Netty version

HEAD of 4.1 branch, as of 8a8576150.

```
$ java -version
java version "1.8.0_162"
Java(TM) SE Runtime Environment (build 1.8.0_162-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.162-b12, mixed mode)
```

```
$ uname -a
Darwin nickt.local 17.3.0 Darwin Kernel Version 17.3.0: Thu Nov  9 18:09:22 PST 2017; root:xnu-4570.31.3~1/RELEASE_X86_64 x86_64
```

nicktrav:
Motivation:

Currently, when passing custom headers to a WebSocketClientHandshaker,
if values are added for headers that are reserved for use in the
websocket handshake performed with the server, these custom values can
be used by the server to compute the websocket handshake challenge. If
the server computes the response to the challenge with the custom header
values, rather than the values computed by the client handshaker, the
handshake may fail.

Modifications:

Update the client handshaker implementations to add the custom header
values first, and then set the reserved websocket header values.

Result:

Reserved websocket handshake headers, if present in the custom headers
passed to the client handshaker, will not be propagated to the server.
Instead the client handshaker will propagate the values it generates.

Fixes #7973.


codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshakerTest.java
```diff
@@ -292,4 +301,34 @@ protected void channelRead0(ChannelHandlerContext ctx, FullHttpResponse msg) thr
...
+
+        // add values for the headers that are reserved for use in the websockets handshake
+        for (CharSequence header : getHandshakeHeaderNames()) {
+            inputHeaders.add(header, bogusHeaderValue);
+        }
+        inputHeaders.add(getProtocolHeaderName(), bogusSubProtocol);
+
+        String realSubProtocol = "realSubProtocol";
+        WebSocketClientHandshaker handshaker = newHandshaker(uri, realSubProtocol, inputHeaders);
+        FullHttpRequest request = handshaker.newHandshakeRequest();
```
normanmaurer: You need to release this one after you are done with it

normanmaurer:
@nicktrav thanks!

Issue #7969, opened 2018-05-24
Make Http2UnknownFrame extend Http2StreamFrame rather than Http2Frame


Bennett-Lynch:
This is a follow-up issue regarding: https://github.com/netty/netty/issues/7860

### Expected behavior

Writing an `Http2UnknownFrame` to a child channel will be written to the parent channel and underlying HTTP/2 connection via `Http2MultiplexCodec`.

When `Http2MultiplexCodec` reads an `Http2UnknownFrame`, it will be propagated to the correct child channel.

### Actual behavior

`Http2MultiplexCodec` rejects the write since `Http2UnknownFrame` is not an `instanceof Http2StreamFrame`.

https://github.com/netty/netty/blob/dfeb4b15b587bf5298555e969a01378853efab8f/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java#L1021-L1028

`Http2MultiplexCodec` does not propagate reads to child channels since `Http2UnknownFrame` is not an `instanceof Http2StreamFrame`.

https://github.com/netty/netty/blob/dfeb4b15b587bf5298555e969a01378853efab8f/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java#L219-L221

### Proposal

`Http2UnknownFrame` should extend `Http2StreamFrame`, rather than `Http2Frame`, as it already defines the two stream-related methods that the two interfaces differ in.

Should it possibly declare an `isEndStream` method as well, similar to `Http2DataFrame`?

normanmaurer:
@Bennett-Lynch let it extend `Http2StreamFrame` sounds like the correct thing to do. Would you be interested in proving a PR ?

I think I would leave the `isEndStream` as it is and just use the `Http2Flags` for it.

normanmaurer:
@Bennett-Lynch I had a bit time this morning so I just did it myself and adding unit tests:

https://github.com/netty/netty/pull/7976

PTAL

Bennett-Lynch:
Thanks, @normanmaurer -- change looks good to me.

Fine with using `Http2Flags` for detecting `isEndStream`, but may I ask your reasoning for it?  Maintaining a consistent interface across frame types seems more ideal, but I realize these UnknownFrames are a bit of an edge case.

normanmaurer:
@Bennett-Lynch because its redundant as it exists in the flags already (and you can have other flags as well).

PR #7976, opened 2018-05-28
Correctly let Http2UnkownFrame extend HttpStreamFrame and so be usabl…

normanmaurer:
…e with Http2MultiplexCodec.

Motivation:

This is a followup for #7860. In the fix for #7860 we only partly fixed the problem as Http2UnknownFrame did not correctly extend HttpStreamFrame and so only worked when using the Http2FrameCodec. We need to have it extend HttpStreamFrame as otherwise Http2MultiplexCodec will reject to handle it correctly.

Modifications:

- Let Http2UnknownFrame extend HttpStreamFrame
- Add unit tests for writing and reading Http2UnkownFrame instances when the Http2MultiplexCodec is used.

Result:

Fixes https://github.com/netty/netty/issues/7969.


normanmaurer:
@Bennett-Lynch PTAL as well

Bennett-Lynch:
LGTM.

PR #7979, opened 2018-05-28
Use jdk-11-ea+15 when try to build with java11

normanmaurer:
Motivation:

A new EA build for java 11 is out.

Modifications:

Update from ea+14 to ea+15

Result:

Use latest ea build


normanmaurer:
Just merging as its only a version bump for java and docker.

PR #7981, opened 2018-05-29
Don't trim stacktrace for exceptions when running mvn test

normanmaurer:
Motivation:

The maven surefire plugin will trim stacktraces by default which makes these kind of use-less when trying to understand why an test failed because one was thrown.

Modifications:

Configure the plugin to not trim the stacktrace.

Result:

Easier to debug test-failures.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -864,6 +864,8 @@
             </property>
           </properties>
           <jvm>${testJavaHome}/bin/java</jvm>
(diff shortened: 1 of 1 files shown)


normanmaurer:
Noticed this while debugging some stuff that only fails on CI... Why this is not the default, I have no idea 👎

PR #7984, opened 2018-05-29
Ignore some test-flakiness when using Java11+ due outstanding Java11 …

normanmaurer:
…bug.

Motivation:

Java11 disallow draining any remaining bytes from the socket if a write causes a connection reset. This should be completely safe to do. At the moment if a write is causing a connection-reset you basically loose all the pending bytes that are sitting on the socket and are waiting to be read.

This happens because SocketOutputStream.write(…) may call AbstractPlainSocketImpl.setConnectionReset(…). Once this method is called any read(…) call will just throw a SocketException without even attempt to read any remaining data.

This is related:
 - https://bugs.openjdk.java.net/browse/JDK-8199329
 - http://hg.openjdk.java.net/jdk/jdk/rev/92cca24c8807
 - http://mail.openjdk.java.net/pipermail/net-dev/2018-May/011511.html

Modifications:

Tolarate if remaining bytes could not be read when using OIO.

### DIFF
--- a/testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketHalfClosedTest.java
+++ b/testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketHalfClosedTest.java
@@ -33,7 +33,9 @@ import io.netty.channel.socket.ChannelInputShutdownEvent;
 import io.netty.channel.socket.ChannelInputShutdownReadComplete;
 import io.netty.channel.socket.ChannelOutputShutdownEvent;
 import io.netty.channel.socket.DuplexChannel;
+import io.netty.channel.socket.oio.OioSocketChannel;
 import io.netty.util.UncheckedBooleanSupplier;
+import io.netty.util.internal.PlatformDependent;
 import org.junit.Test;
 
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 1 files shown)


testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketHalfClosedTest.java
```diff
@@ -414,19 +416,26 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) {
...
+            checkPrematureClose(ctx);
         }
 
-        private void checkPrematureClose() {
+        private void checkPrematureClose(ChannelHandlerContext ctx) {
             if (bytesRead < expectedBytes || !seenOutputShutdown) {
-                causeRef.set(new IllegalStateException("leader premature close"));
-                doneLatch.countDown();
+                if (ctx.channel() instanceof OioSocketChannel && seenOutputShutdown
+                        && PlatformDependent.javaVersion() >= 11) {
```
trustin: Perhaps we should check the Java version before any other conditions because it's more likely to fail fast on most pre-11 deployments?

trustin: Ah, missed that it's a test class. No worries, then!

normanmaurer:
Here is a standalone reproducer which will not fail with Java10 and earlier but fails with latest java11 release:

```java
import java.io.InputStream;
import java.io.OutputStream;
import java.net.ServerSocket;
import java.net.Socket;
import java.net.SocketException;
import java.util.concurrent.CountDownLatch;
import java.util.concurrent.atomic.AtomicReference;

public class Reproducer {

    public static void main(String... args) throws Throwable {
        CountDownLatch readLatch = new CountDownLatch(1);
        CountDownLatch writeLatch = new CountDownLatch(1);
        AtomicReference<Throwable> error = new AtomicReference<>();
        byte[] bytes = new byte[512];

        try (ServerSocket server = new ServerSocket(0)) {
            Thread thread = new Thread(() -> {
                try (Socket client = new Socket()){
                    client.connect(server.getLocalSocketAddress());

                    int readCount = 0;

                    try (InputStream in = client.getInputStream()) {
                        in.read();

                        readCount++;
                        readLatch.countDown();
                        writeLatch.await();


                        OutputStream out = client.getOutputStream();
                        try {
                            for (;;) {
                                // Just write until it fails.
                                out.write(1);
                            }
                        } catch (SocketException expected) {
                        }

                        try {
                            for (;;) {
                                // Read until it fails.
                                in.read();
                                readCount += 1;
                            }
                        } catch (SocketException expected) {
                        }
                    }

                    if (readCount != bytes.length) {
                        throw new AssertionError("Was not able to read remaining data after write error," +
                                " read " + readCount + "/" + bytes.length + ".");
                    }
                } catch (Throwable cause) {
                    error.set(cause);
                }
            });

            thread.start();

            Socket accepted = server.accept();

            // Use SO_LINGER of 0 to ensure we trigger a connection-reset when close the accepted socket.
            accepted.setSoLinger(true, 0);
            accepted.getOutputStream().write(bytes);

            // Wait until we read the first byte. After this there should still be 511 bytes left.
            readLatch.await();

            // Close the accepted channel and signal to the other thread that it should try to write now.
            accepted.close();
            writeLatch.countDown();

            thread.join();

            // Rethrow any error that was produced by the other thread.
            Throwable cause = error.get();
            if (cause != null) {
                throw cause;
            }
        }
    }
}
```

PR #7985, opened 2018-05-29
unexpected sharable handler use

pifuant:
Motivation:

unexpected sharable handler use in example

Modification:

make sharable handler use well

Result:

make sharable handler use well in example


example/src/main/java/io/netty/example/echo/EchoServer.java
```diff
@@ -51,6 +51,7 @@ public static void main(String[] args) throws Exception {
         // Configure the server.
         EventLoopGroup bossGroup = new NioEventLoopGroup(1);
         EventLoopGroup workerGroup = new NioEventLoopGroup();
+        final EchoServerHandler server_handler = new EchoServerHandler();
```
normanmaurer: use camel-case please... `serverHandler`

pifuant:
@normanmaurer PTAL

normanmaurer:
@pifuant also please sign our icla: https://netty.io/s/icla

PR #7986, opened 2018-05-30
Lazy init SelectedSelectionKeySet

carryxyh:
Lazy init SelectedSelectionKeySet because if maybeSelectorImplClass instanceof Throwable, the initialed SelectedSelectionKeySet is never used.


normanmaurer:
@carryxyh thanks a lot... Cherry-picked after adjust the commit message:

https://github.com/netty/netty/commit/b53cf045a7718766da8edd9499c657e77b314bbb

carryxyh:
Ok, thanks a lot

Issue #7990, opened 2018-05-30
CharSequenceValueConverter.convertToByte works differently for AsciiString vs String


henrik-lindqvist:
### Netty version
4.1

Looking at function CharSequenceValueConverter.convertToByte (CharSequence):
https://github.com/netty/netty/blob/4.1/codec/src/main/java/io/netty/handler/codec/CharSequenceValueConverter.java#L79

For AsciiString it takes the first byte, but for any other kind of CharSequence it parse the number value of the entire sequence, i.e. using Byte.parseByte(). This is surely not expected behavior.

See also:
https://docs.oracle.com/javase/8/docs/api/java/lang/Byte.html#parseByte-java.lang.String-

normanmaurer:
@henrik-lindqvist yes this sounds like a bug...

@carl-mastrangelo @ejona86 @Scottmitch @trustin What you think is the correct behaviour here ?

Scottmitch:
I agree there is inconsistency here, and it is not just with `convertToByte` but also with other `convertTo` methods that have a special case for `AsciiString`. The `AsciiString` special case was done for performance reasons, but I think it is more correct to consider the entire value during conversion. We can keep the performance optimization and enforce a max length check too if we want.

We should also be consistent with convertToChar [1] which doesn't have an `AsciiString` optimization but should follow the same semantics.

[1] https://github.com/netty/netty/blob/4.1/codec/src/main/java/io/netty/handler/codec/CharSequenceValueConverter.java#L88

carl-mastrangelo:
My own 2cents: Delete it, or return null.   I don't think there is a reasonable default behavior.  Also, there is hardly anyone using this method, except `getByteAndRemove`, which is itself seldomly used.  Poking around on github's search shows no users.

ValueConverter makes no claims about what subclasses _should_ do, so who can say what will happen?  Further, if I had an instance of type ValueConverter, the implementation would be really important to know.   It cannot provide the abstraction it promises.

henrik-lindqvist:
To be consistent with convertByte is should parse the number, not use the first byte. See: 
https://github.com/netty/netty/blob/4.1/codec/src/main/java/io/netty/handler/codec/CharSequenceValueConverter.java#L74

normanmaurer:
@henrik-lindqvist PTAL https://github.com/netty/netty/pull/7994

PR #7994, opened 2018-06-01
Fix CharSequenceValueConverter.convertToByte implementation for Ascii…

normanmaurer:
…String

Motivation:

The implementation of CharSequenceValueConverter.convertToByte did not correctly handle AsciiString if the length != 1.

Modifications:

- Only use fast-path for AsciiString with length of 1.
- Add unit tests.

Result:

Fixes https://github.com/netty/netty/issues/7990

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/CharSequenceValueConverter.java
+++ b/codec/src/main/java/io/netty/handler/codec/CharSequenceValueConverter.java
@@ -77,7 +77,7 @@ public class CharSequenceValueConverter implements ValueConverter<CharSequence> 
 
     @Override
     public byte convertToByte(CharSequence value) {
-        if (value instanceof AsciiString) {
+        if (value instanceof AsciiString && value.length() == 1) {
             return ((AsciiString) value).byteAt(0);
         }
         return Byte.parseByte(value.toString());
--- a/codec/src/test/java/io/netty/handler/codec/CharSequenceValueConverterTest.java
+++ b/codec/src/test/java/io/netty/handler/codec/CharSequenceValueConverterTest.java
@@ -14,6 +14,7 @@
  */
 package io.netty.handler.codec;
 
+import io.netty.util.AsciiString;
 import org.junit.Test;
 
 import static org.junit.Assert.assertEquals;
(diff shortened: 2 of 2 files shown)


henrik-lindqvist:
Nope, still bugged. Just remove the special AsciiString handling.

normanmaurer:
@henrik-lindqvist it works like it should as far as I can tell ... Please provide a unit test that shows that it doesn't if you think there is something wrong. I tried multiple things and all worked as expected.

henrik-lindqvist:
An "33" or "0x21" input to convertToByte, either as AsciiString or CharSequence, should return (byte)'!' or (byte)33 output. So it's compatible with convertByte(33) which return "33". Correct fix:

    @Override
    public byte convertToByte(CharSequence value) {
        return Byte.parseByte(value.toString());
    }

carl-mastrangelo:
@henrik-lindqvist that *a* way of converting a CharSequence to a byte, but there are many ways.  As I said in #7990, there isn't a "right" conversion here.  

For example, what should happen when converting `0x33` or `033` ?   What about `257` ?  What about `Byte.parseByte("᧓᧓")`, which is U+19D3, and is interpreted by Java as 33 ?  There isn't going to be a right answer.

henrik-lindqvist:
Agreed, that why i reported the issue, you guys have to choose the "right" conversion. But it cant differentiate depending on input type (AsciiString or not). It's either just the first byte like convertToChar do for char, or a sequence of digits (e.g "123") which is parsed into a ASCII value, that's what Byte.parseByte() do.  As said, looking at what convertByte() return, it's probably the latter, unless convertByte() is also bugged.

normanmaurer:
@henrik-lindqvist Can you please provide a unit test that fails ? I am still a bit confused why you think there is an issue. Passing in `33` results in the same output for both at the moment as well, I agree it was not correct before in some case as we ignored the length but now I think it should be fine. With an actual reproducer it would be a lot easier for me to understand the issue.

henrik-lindqvist:
The existing test seems correct:

    @Test
    public void testByteFromAsciiString() {
        assertEquals(127, converter.convertToByte(AsciiString.of("127")));
    }

You might add (untested):

    @Test
    public void testByteFromString() {
        assertEquals(127, converter.convertToByte("127"));
    }

    @Test
    public void testByteToCharSequence() {
        assertTrue(AsciiString.contentEquals("127", converter.convertByte(127)));
    }


To me, it seems impossible the current code would pass, unless byteAt is somehow parsing numbers, or '1' has changed ASCII value to 127:

    @Override
    public byte convertToByte(CharSequence value) {
        if (value instanceof AsciiString && value.length() == 1) {
            return ((AsciiString) value).byteAt(0);
        }
        return Byte.parseByte(value.toString());
}

henrik-lindqvist:
Also add:

@Test
public void testByteFromAsciiString() {
    assertEquals(8, converter.convertToByte(AsciiString.of("8")));
}

henrik-lindqvist:
Another:

    @Test
    public void testByteFromAsciiString() {
        assertEquals(8, converter.convertToByte(AsciiString.of("8")));
   }

PR #7996, opened 2018-06-01
Mention the HttpObjectEncoder's state in the message of the IllegalSt…

rkapsi:
…ateException

Motivation

The HttpObjectEncoder raises an IllegalStateException due to an illegal state but doesn't mention what the state was. It could be useful for debugging purposes to figure out what happened.

Modifications

Mention the HttpObjectEncoder's state in the message of the IllegalStateException.

Result

An exception with more information what caused it.


normanmaurer:
@rkapsi thanks a lot

Issue #7988, opened 2018-05-30
NPE in testsuite logged when using Conscrypt.


normanmaurer:
Saw this during the build:

```
Running io.netty.handler.ssl.ConscryptJdkSslEngineInteropTest
18:01:39.304 [nioEventLoopGroup-676-3] DEBUG io.netty.buffer.PoolThreadCache - Freed 10 thread-local buffer(s) from thread: nioEventLoopGroup-676-3
18:01:39.304 [nioEventLoopGroup-676-2] DEBUG io.netty.buffer.PoolThreadCache - Freed 8 thread-local buffer(s) from thread: nioEventLoopGroup-676-2
[98.268s][info   ][gc,start      ] GC(54) Pause Young (G1 Evacuation Pause)
[98.268s][info   ][gc,task       ] GC(54) Using 8 workers of 8 for evacuation
[98.289s][info   ][gc,phases     ] GC(54)   Pre Evacuate Collection Set: 0.0ms
[98.289s][info   ][gc,phases     ] GC(54)   Evacuate Collection Set: 19.5ms
[98.289s][info   ][gc,phases     ] GC(54)   Post Evacuate Collection Set: 1.5ms
[98.289s][info   ][gc,phases     ] GC(54)   Other: 0.4ms
[98.289s][info   ][gc,heap       ] GC(54) Eden regions: 178->0(171)
[98.289s][info   ][gc,heap       ] GC(54) Survivor regions: 3->10(23)
[98.289s][info   ][gc,heap       ] GC(54) Old regions: 7->7
[98.289s][info   ][gc,heap       ] GC(54) Humongous regions: 54->54
[98.289s][info   ][gc,metaspace  ] GC(54) Metaspace: 33982K->33982K(1081344K)
[98.289s][info   ][gc            ] GC(54) Pause Young (G1 Evacuation Pause) 482M->140M(604M) 21.501ms
[98.290s][info   ][gc,cpu        ] GC(54) User=0.03s Sys=0.01s Real=0.02s
18:01:39.566 [nioEventLoopGroup-689-1] DEBUG io.netty.handler.ssl.SslHandler - [id: 0x4b609c89, L:/127.0.0.1:37424 - R:/127.0.0.1:45348] HANDSHAKEN: TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
[...]

PR #8001, opened 2018-06-04
Update conscrypt to 1.1.3 which fixes some NPEs during tests when u…

normanmaurer:
…sing conscrypt.

Motivation:

When using conscrypt some NPEs were logged, these were fixed in the latest release.

Modifications:

Update to conscrypt 1.1.3.

Result:

Fixes https://github.com/netty/netty/issues/7988.


normanmaurer:
Also /cc @flooey

flooey:
Cool, let me know if you see anything else weird in your tests and we'll look into it.

normanmaurer:
Seems like the NPE is gone and everything else looks fine as well. Thanks for the quick turnaround

PR #8005, opened 2018-06-05
Avoiding the usage of Jmh Invocation level on setup/teardown 

DiegoEliasCosta:
Motivation:

We are conducting a scientific study to investigate bad practices/anti-patterns on creating micro-benchmarks using JMH, and we found two instances of harmfull usage of Invocation level in your benchmark classes.

The usage of Invocation level for JMH fixture methods (setup/teardown) inccurs in a significant impact in
in the benchmark time (see [JMH Documentation ](http://hg.openjdk.java.net/code-tools/jmh/file/tip/jmh-core/src/main/java/org/openjdk/jmh/annotations/Level.java)). 

When the benchmark and the setup/teardown is too small (less than a milisecond) the Invocation level might saturate the system with timestamp requests and iteration synchronizations which introduce artificial latency, throughput, and scalability bottlenecks.

From our experiments Netty has two instances of such case, HeadersBenchmark and CodecInputListBenchmark. On those cases, the benchmark is far too small (<100ns) and the Invocation level setup offsets the measurement considerably.

Modification:

I have applied the simplest fix patch to include what was previously in the setup, inside the benchmark code. 

Result:

In most cases, the benchmarks run faster now, without the artificial latency introduced by the setup code. I have attached a csv with all benchmarks runs and the difference between Netty benchmark on those patches 
- We run the entire benchmark 10 times with default parameters (iterations, forks, warmups kept as default)
- Median score reported 
- Factor = 0 means no significant difference (Moore Median's Test)

Attached Summary: [netty-invocation-10trials-summary.zip](https://github.com/netty/netty/files/2071662/netty-invocation-10trials-summary.zip)

Environment: 
Tests run on a Computational server with CPU: E5-1660-3.3GHZ  (6 cores + HT), 64 GB RAM.

### DIFF
--- a/microbench/src/main/java/io/netty/handler/codec/CodecOutputListBenchmark.java
+++ b/microbench/src/main/java/io/netty/handler/codec/CodecOutputListBenchmark.java
@@ -45,19 +45,19 @@ public class CodecOutputListBenchmark extends AbstractMicrobenchmark {
 
     @Benchmark
     public void codecOutList() {
-    	codecOutputList = CodecOutputList.newInstance();
+        codecOutputList = CodecOutputList.newInstance();
         benchmarkAddAndClear(codecOutputList, elements);
     }
 
     @Benchmark
     public void recyclableArrayList() {
-    	recycleableArrayList = RecyclableArrayList.newInstance(16);
+        recycleableArrayList = RecyclableArrayList.newInstance(16);
         benchmarkAddAndClear(recycleableArrayList, elements);
     }
 
     @Benchmark
     public void arrayList() {
-    	arrayList = new ArrayList<Object>(16);
+        arrayList = new ArrayList<Object>(16);
         benchmarkAddAndClear(arrayList, elements);
     }
 
--- a/microbench/src/main/java/io/netty/handler/codec/CodecOutputListBenchmark.java
+++ b/microbench/src/main/java/io/netty/handler/codec/CodecOutputListBenchmark.java
@@ -18,10 +18,8 @@ package io.netty.handler.codec;
 import io.netty.microbench.util.AbstractMicrobenchmark;
 import io.netty.util.internal.RecyclableArrayList;
 import org.openjdk.jmh.annotations.Benchmark;
-import org.openjdk.jmh.annotations.Level;
 import org.openjdk.jmh.annotations.Param;
 import org.openjdk.jmh.annotations.Scope;
-import org.openjdk.jmh.annotations.Setup;
 import org.openjdk.jmh.annotations.State;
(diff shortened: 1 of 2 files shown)


microbench/src/main/java/io/netty/handler/codec/CodecOutputListBenchmark.java
```diff
@@ -54,16 +45,19 @@ public void destroy() {
 
     @Benchmark
     public void codecOutList() {
+        codecOutputList = CodecOutputList.newInstance();
         benchmarkAddAndClear(codecOutputList, elements);
```
normanmaurer: Shouldn't we call `codecOutputList.recycle()` here  and remove it from the `@TearDown` ?

microbench/src/main/java/io/netty/handler/codec/CodecOutputListBenchmark.java
```diff
@@ -54,16 +45,19 @@ public void destroy() {
...
     }
 
     @Benchmark
     public void recyclableArrayList() {
+        recycleableArrayList = RecyclableArrayList.newInstance(16);
         benchmarkAddAndClear(recycleableArrayList, elements);
```
normanmaurer: Shouldn't we call `recycleableArrayList.release()` here  and remove it from the `@TearDown` ?

microbench/src/main/java/io/netty/handler/codec/CodecOutputListBenchmark.java
DiegoEliasCosta: I thought about doing that on my first patch, but that would actually change what is being measured, and I couldn't do that without checking with you (the developers).

The way it was first implemented, the `destroy` method is called just **once** per entire benchmark execution (no Level on `@Teardown` specified, the default is Level.Trial). I think the call to the recycle does not seems to make sense here we are creating a new instance on every benchmark invocation.

Now if you also want to measure the `recycle` method call - which wasn't being measured before - I will include it there, otherwise I would actually remove it from the benchmark.

What do you think?

PS: The same analysis is true for the RecyclableArrayList

DiegoEliasCosta:
Thanks for the feedback @normanmaurer, I believe I have already signed the ICLA when first submitting the pull request.
I can sign it again, but can someone confirm if you have already received my signature?

microbench/src/main/java/io/netty/handler/codec/CodecOutputListBenchmark.java
normanmaurer: true... lets just keep it as it is for now

PR #8009, opened 2018-06-06
Remove dead code in Http2CodecUtil

### LINKED ISSUE #8002: Http2CodecUtil.java - emptyPingBuf() is no longer needed (as writePing takes a long)
https://github.com/netty/netty/blob/b192bf12ad2af92bba0f32c9d3127b1192e54670/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecUtil.java#L177

In https://github.com/netty/netty/commit/501662a77f8166382eefd18a8620983d27f59d4a - emptyPing data changed to long (instead of buffer data)

This should be updated.

bryce-anderson:
Motivation:

The `ByteBuffer emptyPingBuf()` method of Http2CodecUtils is has been dead
code since DefaultHttp2PingFrame switched from using a ByteBuf to represent
the 8 octets to a long.

Modifications:

Remove the method and the unused static ByteBuf.

Result:

Less dead code.

Fixes #8002


bryce-anderson:
h/t to @vsabella who reported the dead code.

PR #8015, opened 2018-06-11
Make UnpooledHeapByteBuf array methods protected

tbrooks8:
Motivation:

Currently there is not a clear way to provide a byte array to a netty
ByteBuf and be informed when it is released. This is a would be a
valuable addition for projects that integrate with netty but also pool
their own byte arrays.

Modification:

Modified the UnpooledHeapByteBuf class so that the freeArray method is
protected visibility instead of default. This will allow a user to
subclass the UnpooledHeapByteBuf, provide a byte array, and override
freeArray to return the byte array to a pool when it is called.
Additionally this makes this implementation equivalent to
UnpooledDirectByteBuf (freeDirect is protected).

Additionally allocateArray is also made protect to provide another override
option for subclasses.

Result:

Users can override UnpooledHeapByteBuf#freeArray and 
UnpooledHeapByteBuf#allocateArray.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/UnpooledByteBufAllocator.java
+++ b/buffer/src/main/java/io/netty/buffer/UnpooledByteBufAllocator.java
@@ -147,7 +147,7 @@ public final class UnpooledByteBufAllocator extends AbstractByteBufAllocator imp
         }
 
         @Override
-        void freeArray(byte[] array) {
+        protected void freeArray(byte[] array) {
             int length = array.length;
             super.freeArray(array);
             ((UnpooledByteBufAllocator) alloc()).decrementHeap(length);
@@ -167,7 +167,7 @@ public final class UnpooledByteBufAllocator extends AbstractByteBufAllocator imp
         }
 
         @Override
-        void freeArray(byte[] array) {
+        protected void freeArray(byte[] array) {
             int length = array.length;
             super.freeArray(array);
             ((UnpooledByteBufAllocator) alloc()).decrementHeap(length);
--- a/buffer/src/main/java/io/netty/buffer/UnpooledByteBufAllocator.java
+++ b/buffer/src/main/java/io/netty/buffer/UnpooledByteBufAllocator.java
@@ -140,7 +140,7 @@ public final class UnpooledByteBufAllocator extends AbstractByteBufAllocator imp
         }
 
         @Override
-        byte[] allocateArray(int initialCapacity) {
+        protected byte[] allocateArray(int initialCapacity) {
             byte[] bytes = super.allocateArray(initialCapacity);
             ((UnpooledByteBufAllocator) alloc()).incrementHeap(bytes.length);
             return bytes;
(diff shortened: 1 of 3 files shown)


normanmaurer:
@carl-mastrangelo @mosesn @bryce-anderson WDYT about this ?

bryce-anderson:
The symmetry with `UnpooledDirectByteBuf ` is appealing. This seems fine to me but I have limited context so please defer to more educated opinions.

tbrooks8:
> If you are bringing your own allocator, why not make your own Buf too?

I want to clarify that the use case I am describing for making a method `protected` opposed to `default` does not involve implementing a netty allocator. Instead in the use case I am describing, the user provides a byte array to an instance (as allowed by the `protected` constructor) and would just like some way to know when the `UnpooledHeapByteBuf` frees the array it is using (which might be the original array or might be a different array created in from `capacity(int)`). The identity check will be left to me to implement.

Why not implement `ByteBuf`? `UnpooledHeapByteBuf` provides ~500 lines of code operating on a byte array that is exactly what I am looking for. If there were a `HeapByteBuf` superclass with setters for the underlying array and the operations that changed the array (`capacity(int)`, etc) were all that was implemented by `UnpooledHeapByteBuf` I would just extend `HeapByteBuf`. But that's not exactly how it exists now and I'm not sure if that's what you would prefer this pull request be looking at?

So finally, I would like to emphasize the is an `Unpooled` buffer from netty's perspective. Which is why I am looking at using this class. All I would like is the ability to access the method that is called when netty switches out the underlying array. Setting aside my case, this seems like useful functionality as netty itself uses this to implement instrumented `ByteBufs` instead of implementing completely different `ByteBufs`.

carl-mastrangelo:
@tbrooks8 If I understand you correctly, you want to know when the array is freed, not what the array actually is.  If that is the case, why not override `deallocate` or `release` ?

tbrooks8:
`deallocate` does not provide the information that I need as the array might have changed due to a call in `capacity(int)`.

I mean it will work for a `ByteBuf` that cannot be resized, but it the `ByteBuf` has a max capacity that allows resizing then it not inform the user when we are done with the original array (at least for a while).

Here is the ideal use case:

```
public class NewByteBuf extends UnpooledHeapByteBuf {

    private final byte[] initialArray;
    private final Runnable returnArrayToPoolRunnable;

    protected NewByteBuf(byte[] initialArray, Runnable returnArrayToPoolRunnable) {
        super(UnpooledByteBufAllocator.DEFAULT, initialArray, Integer.MAX_VALUE);
        this.initialArray = initialArray;
        this.returnArrayToPoolRunnable = returnArrayToPoolRunnable;
    }

    @Override
    protected void freeArray(byte[] array) {
        super.freeArray(array);
        if (array == initialArray) {
            returnArrayToPoolRunnable.run();
        }
    }
}
```

normanmaurer:
@tbrooks8 after hearing more about what you try to do I think this change is "ok"... I just wonder why you not also make "allocateArray" protected and so also allow to hook in different ways how the byte[] is allocated. This should give you the "max" control, no ?

tbrooks8:
@normanmaurer I did not make `allocateArray` protected because my use case is more related to "seeding" a bytebuf with data that is pooled. But I agree that making it protected might be valuable and is symmetrical. I modified the PR and description.

normanmaurer:
@tbrooks8 please squash and update commit message to reflect the changes.

tbrooks8:
I modified the description when I make the `allocateArray` method protected. Is there some more changes you want to the description?

normanmaurer:
@tbrooks8 sorry missed this... merged :)

tbrooks8:
thanks @normanmaurer

PR #8021, opened 2018-06-13
Defer channelInactive and channelUnregistered events in Http2MultiplexCodec

### LINKED ISSUE #8018: Http2MultiplexCodec.DefaultHttp2StreamChannel.Http2ChannelUnsafe has a different order of close events than other channel implementation
The `HttpObjectAggregator`, when it encounters a too-large-message, [does the following](https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectAggregator.java#L281) when handling a too large response:
```java
ctx.close();
throw new TooLongFrameException("Response entity too large: " + oversized);
```
The tail of the pipeline gets an `exceptionCaught` event before receiving a `channelInactive` event because channel deregistration is always done '[later](https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/AbstractChannel.java#L813)'. This is not the case in the `Http2ChannelUnsafe` which triggers shutdown events [immediately](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java#L925). That means when using the `HttpObjectAggregator` in a stream channel, we will receive  the`channelInactive` event before the `exceptionCaught` event, making it difficult to handle gracefully.

### Netty version

4.1.26-SNAPSHOT

openjdk version "1.8.0_162"

Darwin

bryce-anderson:
Motivation:

There is an inconsistency between the order of events in the
StreamChannel implementation in Http2MultiplexCodec and other Channel
implementations that extend AbstractChannel where channelInactive and
channelUnregistered events are not performed 'later'. This can cause an
unexected order of events for ChannelHandler implementations that call
Channel.close() in response to some event.

Modification:

The Http2MultiplexCodec.DefaultHttp2StreamChannel.Http2ChannelUnsafe was
modified to bounce the deregistration and channelInactive events through
the parent channels EventLoop.

Result:

Stream events are now in the proper order.

Fixes #8018.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
@@ -43,11 +43,14 @@ import io.netty.util.ReferenceCounted;
 import io.netty.util.internal.StringUtil;
 import io.netty.util.internal.ThrowableUtil;
 import io.netty.util.internal.UnstableApi;
+import io.netty.util.internal.logging.InternalLogger;
+import io.netty.util.internal.logging.InternalLoggerFactory;
 
 import java.net.SocketAddress;
 import java.nio.channels.ClosedChannelException;
 import java.util.ArrayDeque;
 import java.util.Queue;
+import java.util.concurrent.RejectedExecutionException;
 
 import static io.netty.handler.codec.http2.Http2CodecUtil.HTTP_UPGRADE_STREAM_ID;
 import static io.netty.handler.codec.http2.Http2CodecUtil.isStreamIdValid;
@@ -103,6 +106,8 @@ import static java.lang.Math.min;
  */
 @UnstableApi
 public class Http2MultiplexCodec extends Http2FrameCodec {
+
+    private static final InternalLogger logger = InternalLoggerFactory.getInstance(DefaultHttp2StreamChannel.class);
 
     private static final ChannelFutureListener CHILD_CHANNEL_REGISTRATION_LISTENER = new ChannelFutureListener() {
         @Override
@@ -899,6 +904,8 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
                 closePending = false;
                 fireChannelReadPending = false;
 
+                final boolean wasActive = isActive();
+
                 // Only ever send a reset frame if the connection is still alive as otherwise it makes no sense at
                 // all anyway.
                 if (parent().isActive() && !streamClosedWithoutError && isStreamIdValid(stream().id())) {
@@ -922,10 +929,7 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
                 closePromise.setSuccess();
                 promise.setSuccess();
 
-                pipeline().fireChannelInactive();
-                if (isRegistered()) {
-                    deregister(unsafe().voidPromise());
-                }
+                fireChannelInactiveAndDeregister(voidPromise(), wasActive);
             }
 
             @Override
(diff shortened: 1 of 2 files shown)


bryce-anderson:
A fair amount of this was just adapted from AbstractChannel. It felt nice to keep the parity between them since most of the code is duplicated, but I'm happy to slim it down if people think that would be better.

As an side, we should consider refactoring AbstractChannel such that it can be extended by Http2StreamChannel once we get a chance to break API.

normanmaurer:
@bryce-anderson please fix check-style

bryce-anderson:
@normanmaurer, done. One of these days I'll do it before pushing... Sorry for the noise.

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
```diff
@@ -583,6 +584,57 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
...
+            }
+
+            @Override
+            public void channelUnregistered(ChannelHandlerContext ctx) throws Exception {
+                channelUnregistered.set(count.getAndIncrement());
+                super.channelUnregistered(ctx);
+            }
+        });
+
+        childChannel.pipeline().addBefore("sniffer", "thrower", new ChannelInboundHandlerAdapter() {
```
normanmaurer: You can simplify this by just adding this handler first.

bryce-anderson: How about simple `addFirst`?

normanmaurer: Sure... that said I think just adding it first is simpler. But whatever you prefer

bryce-anderson: I'm not 100% sure what you mean, I need the pipeline to be in this order (and note that `streamActiveAndWriteHeaders` adds a `LastInboundHandler` to the pipeline) so adding the 'thrower' first would necessitate using `addAfter` to place the 'sniffer' handler.

normanmaurer: Basically just add the handlers in the opposite order and use `addLast`:


```java
childChannel.pipeline().addLast(new ChannelInboundHandlerAdapter() { }); // this is the thrower
childChannel.pipeline(). addLast(new ChannelInboundHandlerAdapter() { }); // this is the sniffer

```

bryce-anderson: As it was written that wouldn't work because there is already a `ListInboundHandler` (added by the `streamActiveAndWrite that would soak up events. I've changed it to just an outbound stream (via `newOutboundStream`) and we can do what you propose which is cleaner. 

Thanks for the feedback!

normanmaurer:
@bryce-anderson thanks a lot!

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -935,15 +939,70 @@ public void closeForcibly() {
...
+                        // an open channel.  Their doDeregister() calls close(). Consequently,
+                        // close() calls deregister() again - no need to fire channelUnregistered, so check
+                        // if it was registered.
+                        if (registered) {
+                            registered = false;
+                            pipeline.fireChannelUnregistered();
+                        } else {
+                            promise.setFailure(new IllegalStateException("Not registered"));
+                        }
+                        safeSetSuccess(promise);
```
carl-mastrangelo: This should go in the if-clause above, right?

bryce-anderson: Yes, I believe you're right. I'll put up another PR to fix it. Thanks for the review.

bryce-anderson: It looks like the `AbstractChannel` always set the promise to success, so I've put up a PR https://github.com/netty/netty/pull/8029 to align the behavior.

PR #8022, opened 2018-06-13
Check if Log level is enabled before creating log statements

rkapsi:
Motivation

There is a cost to concatenating strings and calling methods that will be wasted if the Logger's level is not enabled.

Modifications

Check if Log level is enabled before producing log statement. These are just a few cases found by RegEx'ing in the code.

Result

Tiny bit more efficient code.


hc-codersatlas:
Wouldn't it be better to just switch to using parameters in this case? I'd remove the branching altogether (to check for log levels - since it does it under the hood anyway)

rkapsi:
Not really if parameters are non-values (i.e. method calls).

PR #8028, opened 2018-06-16
Issue #8025. Ignoring HEADER and DATA frames for streams that may hav…

### LINKED ISSUE #8025: HTTP/2 - Race Condition with RST_STREAM
### Expected behavior
When a RST_STREAM is sent to the remote, any future incoming DATA frames should be ignored.

### Actual behavior
Sometimes an error is thrown instead. (See https://github.com/grpc/grpc-java/issues/3548)

### Steps to reproduce
Send a RST_STREAM frame to a remote endpoint.

### Netty version
4.1.23.Final

Java 8

Any (observed on Mac and Linux).

For a bit more context. Sending a RST_STREAM will eventually trigger this code in Http2ConnectionHandler, https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java#L589 . This removes the stream from the active streams metadata and any subsequent DATA frames that come in will be considered as an unknown stream. I was thinking we could just check if the stream may have existed and if so just ignore the DATA frame altogether. So instead of [this](https://github.com/netty/netty/blob/501662a77f8166382eefd18a8620983d27f59d4a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java#L535) do something like this

```java
verifyStreamMayHaveExisted(streamId);
return true;
```

shorea:
…e existed in the past.

Motivation:

Correctly ignore frames that come in after a RST_STREAM is sent.

Modification:

Ignoring frames that are for a stream that may have existed in the past instead of throwing an exception.

Result:

Fixes #8025.

### DIFF
--- a/all/pom.xml
+++ b/all/pom.xml
@@ -20,7 +20,7 @@
   <parent>
     <groupId>io.netty</groupId>
     <artifactId>netty-parent</artifactId>
-    <version>4.1.38.Final</version>
+    <version>4.1.39.Final-SNAPSHOT</version>
   </parent>
 
   <artifactId>netty-all</artifactId>
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -25,7 +25,7 @@
 
   <groupId>io.netty</groupId>
   <artifactId>netty-bom</artifactId>
-  <version>4.1.38.Final</version>
+  <version>4.1.39.Final-SNAPSHOT</version>
   <packaging>pom</packaging>
 
   <name>Netty/BOM</name>
@@ -49,7 +49,7 @@
     <url>https://github.com/netty/netty</url>
     <connection>scm:git:git://github.com/netty/netty.git</connection>
     <developerConnection>scm:git:ssh://git@github.com/netty/netty.git</developerConnection>
-    <tag>netty-4.1.38.Final</tag>
+    <tag>HEAD</tag>
   </scm>
 
   <developers>
@@ -69,165 +69,165 @@
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-buffer</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-codec</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-codec-dns</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-codec-haproxy</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-codec-http</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-codec-http2</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-codec-memcache</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-codec-mqtt</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-codec-redis</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-codec-smtp</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-codec-socks</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-codec-stomp</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-codec-xml</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-common</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-dev-tools</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-handler</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-handler-proxy</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-resolver</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-resolver-dns</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-transport</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-transport-rxtx</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-transport-sctp</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-transport-udt</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-example</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-all</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-transport-native-unix-common</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-transport-native-unix-common</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
         <classifier>linux-x86_64</classifier>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-transport-native-unix-common</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
         <classifier>osx-x86_64</classifier>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-transport-native-epoll</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-transport-native-epoll</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
         <classifier>linux-x86_64</classifier>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-transport-native-kqueue</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>io.netty</groupId>
         <artifactId>netty-transport-native-kqueue</artifactId>
-        <version>4.1.38.Final</version>
+        <version>4.1.39.Final-SNAPSHOT</version>
         <classifier>osx-x86_64</classifier>
       </dependency>
     </dependencies>
--- a/buffer/pom.xml
+++ b/buffer/pom.xml
@@ -20,7 +20,7 @@
   <parent>
     <groupId>io.netty</groupId>
     <artifactId>netty-parent</artifactId>
-    <version>4.1.38.Final</version>
+    <version>4.1.39.Final-SNAPSHOT</version>
   </parent>
 
   <artifactId>netty-buffer</artifactId>
--- a/codec-dns/pom.xml
+++ b/codec-dns/pom.xml
@@ -20,7 +20,7 @@
   <parent>
     <groupId>io.netty</groupId>
     <artifactId>netty-parent</artifactId>
-    <version>4.1.38.Final</version>
+    <version>4.1.39.Final-SNAPSHOT</version>
   </parent>
 
   <artifactId>netty-codec-dns</artifactId>
--- a/codec-haproxy/pom.xml
+++ b/codec-haproxy/pom.xml
@@ -20,7 +20,7 @@
   <parent>
     <groupId>io.netty</groupId>
     <artifactId>netty-parent</artifactId>
-    <version>4.1.38.Final</version>
+    <version>4.1.39.Final-SNAPSHOT</version>
   </parent>
 
   <artifactId>netty-codec-haproxy</artifactId>
--- a/codec-http/pom.xml
+++ b/codec-http/pom.xml
@@ -20,7 +20,7 @@
   <parent>
     <groupId>io.netty</groupId>
     <artifactId>netty-parent</artifactId>
-    <version>4.1.38.Final</version>
+    <version>4.1.39.Final-SNAPSHOT</version>
   </parent>
 
   <artifactId>netty-codec-http</artifactId>
--- a/codec-http2/pom.xml
+++ b/codec-http2/pom.xml
@@ -20,7 +20,7 @@
   <parent>
     <groupId>io.netty</groupId>
     <artifactId>netty-parent</artifactId>
-    <version>4.1.38.Final</version>
+    <version>4.1.39.Final-SNAPSHOT</version>
   </parent>
 
   <artifactId>netty-codec-http2</artifactId>
--- a/codec-memcache/pom.xml
+++ b/codec-memcache/pom.xml
@@ -20,7 +20,7 @@
   <parent>
     <groupId>io.netty</groupId>
     <artifactId>netty-parent</artifactId>
-    <version>4.1.38.Final</version>
+    <version>4.1.39.Final-SNAPSHOT</version>
   </parent>
 
   <artifactId>netty-codec-memcache</artifactId>
--- a/codec-mqtt/pom.xml
+++ b/codec-mqtt/pom.xml
@@ -20,7 +20,7 @@
   <parent>
     <groupId>io.netty</groupId>
     <artifactId>netty-parent</artifactId>
-    <version>4.1.38.Final</version>
+    <version>4.1.39.Final-SNAPSHOT</version>
   </parent>
 
   <artifactId>netty-codec-mqtt</artifactId>
--- a/codec-redis/pom.xml
+++ b/codec-redis/pom.xml
@@ -20,7 +20,7 @@
   <parent>
     <groupId>io.netty</groupId>
     <artifactId>netty-parent</artifactId>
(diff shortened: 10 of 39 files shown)


codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoderTest.java
```diff
@@ -272,26 +272,30 @@ public void dataReadForUnknownStreamShouldApplyFlowControlAndFail() throws Excep
         }
     }
 
+    @Test(expected = Http2Exception.class)
+    public void dataReadForUnknownStreamThatNeverExistedShouldThrow() throws Exception {
+        when(connection.streamMayHaveExisted(STREAM_ID)).thenReturn(false);
+        when(connection.stream(STREAM_ID)).thenReturn(null);
+        final ByteBuf data = dummyData();
+        decode().onDataRead(ctx, STREAM_ID, data, 0, true);
```
normanmaurer: @shorea do we need to call `data.release()` in a finally block here ?

normanmaurer:
@carl-mastrangelo @ejona86 @mosesn @bryce-anderson PTAL

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoderTest.java
normanmaurer: @shorea ping me once you addressed it.

normanmaurer:
@ejona86 does this look ok to you (except the missing release call in the test) ?

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoderTest.java
shorea: @normanmaurer  just pushed a fix for this.

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoderTest.java
```diff
@@ -587,8 +600,21 @@ public void pushPromiseReadAfterGoAwayShouldAllowFramesForStreamCreatedByLocalEn
         verify(listener).onPushPromiseRead(eq(ctx), anyInt(), anyInt(), any(Http2Headers.class), anyInt());
     }
 
+    @Test
+    public void pushPromiseReadForUnknownStreamThatMayHaveExistedShouldBeIgnored() throws Exception {
```
bryce-anderson: I don't think we should be ignoring these, we should be resetting. The [spec states](https://http2.github.io/http2-spec/#rfc.section.5.1) (bottom of the closed state description):

> An endpoint might receive a PUSH_PROMISE frame after it sends RST_STREAM. PUSH_PROMISE causes a stream to become "reserved" even if the associated stream has been reset. Therefore, a RST_STREAM is needed to close an unwanted promised stream.

shorea: @bryce-anderson I agree but I also don't know how to fix this. In https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java#L766 we require the parent stream to be non-null. Do you have an idea in mind in how to make this compliant with the spec?

bryce-anderson: Do we need to use `shouldIgnoreHeadersOrDataFrame` for a PUSH_PROMISE? From a brief read it doesn't seem appropriate since it's using the parent streams ID in the `streamCreatedAfterGoAwaySent(streamId)` method but we should be considering the promised stream id as the GOAWAY frame we would have sent carries the last _peer-initiated_ stream id to be processed.

To fix the behavior can we do the following: ignore the frame if `streamCreatedAfterGoAwaySent(promisedStreamId) == true`, otherwise send a RST_STREAM if the parent stream is `null`, otherwise continue on with the existing code path?

shorea: +1, running the tests now.

bryce-anderson: Also, I suppose we should check if the parent stream ever could have existed and if the promised stream does get sent a reset, we need to make sure we 'observe' the promised stream id via `incrementExpectedStreamId(streamId);` which unfortunately is in the `DefaultHttp2ConnectionreservePushStream method`. A lot of protocol checks in there apply to the promise frame even if the parent has already been reset. Maybe the right thing to do is throw a `streamError` in there instead of a protocol error if the parent stream doesn't exist (which if it's valid would be because it's reset)?

shorea: @bryce-anderson let me know if the latest commit is in line with what you were thinking. I wasn't 100% sure what affect throwing a stream error has but it seems to eventually send a RST_STREAM. is that accurate?

https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java#L704

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
```diff
@@ -764,7 +764,9 @@ public boolean isServer() {
         @Override
         public DefaultStream reservePushStream(int streamId, Http2Stream parent) throws Http2Exception {
             if (parent == null) {
-                throw connectionError(PROTOCOL_ERROR, "Parent stream missing");
+                incrementExpectedStreamId(streamId);
+                throw streamError(streamId, Http2Error.REFUSED_STREAM, "PUSH_PROMISE sent on unknown stream, "
+                                                                       + "likely discarded due to a RST_STREAM.");
```
bryce-anderson: I think this is better, but I'm not sure we're going to reach it: we're still using `shouldIgnoreHeadersOrDataFrame` in `DefaultHttp2ConnectionDecoder.onPushPromiseRead` which will return early if the parent stream is `null`.

shorea: Yeah forgot to update the decoder as well. I'll push a fix today.

shorea: @bryce-anderson can you take another look?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -445,26 +445,24 @@ public void onPushPromiseRead(ChannelHandlerContext ctx, int streamId, int promi
...
-                  throw connectionError(PROTOCOL_ERROR,
-                      "Stream %d in unexpected state for receiving push promise: %s",
-                      parentStream.id(), parentStream.state());
+            if (parentStream != null) {
+                switch (parentStream.state()) {
+                    case OPEN:
+                    case HALF_CLOSED_LOCAL:
+                        // Allowed to receive push promise in these states.
+                        break;
+                    default:
```
bryce-anderson: If the stream was reset, we could be in the CLOSED state, and this will interpret it as a connection error. I don't know if we'll still have a `Http2Stream` at that point though, so it may not be possible.

shorea: I'm not sure it's possible either, per my understanding of the code we remove the stream from the connection when we send the RST_STREAM

shorea:
@normanmaurer can this be merged and released?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
normanmaurer: @mosesn @ejona86 can you please have a final look ....

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoderTest.java
```diff
@@ -272,26 +273,34 @@ public void dataReadForUnknownStreamShouldApplyFlowControlAndFail() throws Excep
...
-                fail();
-            } catch (Http2Exception e) {
-                verify(localFlow)
-                        .receiveFlowControlledFrame(eq((Http2Stream) null), eq(data), eq(padding), eq(true));
-                verify(localFlow).consumeBytes(eq((Http2Stream) null), eq(processedBytes));
-                verify(localFlow).frameWriter(any(Http2FrameWriter.class));
-                verifyNoMoreInteractions(localFlow);
+            decode().onDataRead(ctx, STREAM_ID, data, padding, true);
+            verify(localFlow)
+                .receiveFlowControlledFrame(eq((Http2Stream) null), eq(data), eq(padding), eq(true));
```
carl-mastrangelo: optional: if every arg is `eq`, you can remove the wrapping of each of them.

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoderTest.java
```diff
@@ -272,26 +273,34 @@ public void dataReadForUnknownStreamShouldApplyFlowControlAndFail() throws Excep
...
+            verify(localFlow).consumeBytes(eq((Http2Stream) null), eq(processedBytes));
+            verify(localFlow).frameWriter(any(Http2FrameWriter.class));
+            verifyNoMoreInteractions(localFlow);
 
-                // Verify that the event was absorbed and not propagated to the observer.
-                verify(listener, never()).onDataRead(eq(ctx), anyInt(), any(ByteBuf.class), anyInt(), anyBoolean());
-            }
+            // Verify that the event was absorbed and not propagated to the observer.
+            verify(listener, never()).onDataRead(eq(ctx), anyInt(), any(ByteBuf.class), anyInt(), anyBoolean());
```
carl-mastrangelo: should this be for any `ctx`?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -529,11 +527,11 @@ private boolean shouldIgnoreHeadersOrDataFrame(ChannelHandlerContext ctx, int st
                             ctx.channel(), frameName, streamId);
                     return true;
                 }
-                // Its possible that this frame would result in stream ID out of order creation (PROTOCOL ERROR) and its
-                // also possible that this frame is received on a CLOSED stream (STREAM_CLOSED after a RST_STREAM is
-                // sent). We don't have enough information to know for sure, so we choose the lesser of the two errors.
-                throw streamError(streamId, STREAM_CLOSED, "Received %s frame for an unknown stream %d",
-                                  frameName, streamId);
+                // If the stream could have existed in the past we assume this is a frame sent by the remote
```
Scottmitch: Just to be sure we stay in sync with the remote peer we can always throw a `streamError` to inform the remote peer that we don't understand the current frame. This will help provide visibly in the event the remote/local get out of sync for some reason. For example the spec recommends it for the PUSH_PROMISE case and in general allows multiple RST_STREAM frames in extraneous scenarios.

> An endpoint might receive a PUSH_PROMISE frame after it sends
      RST_STREAM.  PUSH_PROMISE causes a stream to become "reserved"
      even if the associated stream has been reset.  Therefore, a
      RST_STREAM is needed to close an unwanted promised stream.

> However, an endpoint MAY send additional RST_STREAM
   frames if it receives frames on a closed stream after more than a
   round-trip time.  This behavior is permitted to deal with misbehaving
   implementations.

Some considerations:
- there should be some limit to how many RST_STREAMs we are willing to send in this scenario before either going into graceful close or eventually giving up and hard closing. For example we don't want to indefinitely send RST_STREAMs. However this heuristic can be implemented in a followup PR and doesn't change the existing behavior.
- The spec says the following, but in this case we have discarded the local state and don't know if we have already sent a RST_STREAM at this time ... which the spec allows sending multiple RST_STREAMs (as indicated above).
> An endpoint MUST ignore frames that it
      receives on closed streams after it has sent a RST_STREAM frame.
      An endpoint MAY choose to limit the period over which it ignores
      frames and treat frames that arrive after this time as being in
      error.


```java
if (connection.streamMayHaveExisted(streamId)) {
  throw streamError(streamId, STREAM_CLOSED, "Received %s frame for an unknown stream %d", frameName, streamId);
} else {
  throw connectionError(PROTOCOL_ERROR, "Stream %d does not exist", streamId);
}
```

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoderTest.java
```diff
@@ -272,26 +273,34 @@ public void dataReadForUnknownStreamShouldApplyFlowControlAndFail() throws Excep
         }
     }
 
+    @Test(expected = Http2Exception.class)
```
Scottmitch: it would be good to verify which type of exception is being thrown in this case. For example a `StreamException` doesn't seem appropriate in this case, but would still pass the test.

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoderTest.java
```diff
@@ -272,26 +273,34 @@ public void dataReadForUnknownStreamShouldApplyFlowControlAndFail() throws Excep
...
     public void dataReadForUnknownStreamShouldApplyFlowControl() throws Exception {
         when(connection.stream(STREAM_ID)).thenReturn(null);
         final ByteBuf data = dummyData();
         int padding = 10;
         int processedBytes = data.readableBytes() + padding;
         try {
-            try {
-                decode().onDataRead(ctx, STREAM_ID, data, padding, true);
-                fail();
-            } catch (Http2Exception e) {
```
Scottmitch: As previously discussed it seems like sending a `RST_STREAM` in this scenario is still useful to inform the remote end that the data was not accepted, and to update its local state.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
shorea: @Scottmitch per your second consideration (if I'm understanding correctly), that would negate the original purpose of the PR. We want to ignore any data frames that come after the reset and not propagate failures to the stream.

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoderTest.java
shorea: +1

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoderTest.java
shorea: Good call, will simplify

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoderTest.java
shorea: +1

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
Scottmitch: I looked at the original issue and here is what the RFC says about the scenario when we have sent a RST_STREAM frame and receive a DATA frame:

https://tools.ietf.org/html/rfc7540#section-5.1
> An endpoint MUST ignore frames that it
      receives on closed streams after it has sent a RST_STREAM frame.
      An endpoint MAY choose to limit the period over which it ignores
      frames and treat frames that arrive after this time as being in
      error.

https://tools.ietf.org/html/rfc7540#section-6.1
> If a DATA frame is received
   whose stream is not in "open" or "half-closed (local)" state, the
   recipient MUST respond with a stream error (Section 5.4.2) of type
   STREAM_CLOSED.

https://tools.ietf.org/html/rfc7540#section-5.4.2
> A RST_STREAM is the last frame that an endpoint can send on a stream.
   The peer that sends the RST_STREAM frame MUST be prepared to receive
   any frames that were sent or enqueued for sending by the remote peer.
   These frames can be ignored, except where they modify connection
   state (such as the state maintained for header compression
   (Section 4.3) or flow control).

> Normally, an endpoint SHOULD NOT send more than one RST_STREAM frame
   for any stream.  However, an endpoint MAY send additional RST_STREAM
   frames if it receives frames on a closed stream after more than a
   round-trip time.  This behavior is permitted to deal with misbehaving
   implementations.

> To avoid looping, an endpoint MUST NOT send a RST_STREAM in response
   to a RST_STREAM frame.

Here is my interpretation of the above somewhat contradictory statements:
1. If state transitions in a coordinated fashion between peers into the CLOSED state and a DATA frame is received in the CLOSED state, then you MUST send a RST_STREAM.
2. Sending a RST_STREAM is an uncoordinated means of forcing the stream into the closed state, and therefore you may receive frames that are already in flight. This frames can be ignored if you are reasonably confident that the remote peer has not yet received the RST_STREAM, but you are also allowed to send multiple RST_STREAMs.
3. Receiving a RST_STREAM frame while the stream is in the CLOSED state should have no protocol impacts.

Given the above criteria my preferred way to handle the scenario is the following:
- If the stream may have existed ... tracking/estimating RTT time is error prone and can be expensive for what is in large part a boundary condition. The RFC permits sending multiple RST_STREAM frames, doing so will have no negative implications from a protocol perspective*. A simpler approach would be to continue to send the RST_STREAM frames which provides positive acknowledgement for good peers that this frame was not accepted, and have some heuristic to "fail hard" (e.g. connection error) to catch the case of "misbehaving peers".
- if the stream was never in existence ... protocol error
- reduce the log level in `io.grpc.netty.NettyServerHandler` for this case (or consider a more selective log strategy)

(*) of course this will consume network resources but it is viewed as a tradeoff to keep things simple from state management and heuristic perspective. If this is proven to consume non-negligible network resources for some use cases I would first like to explore handling the heuristic of "not sending the RST_STREAM bcz I think this falls within the RTT of the previous frame" up into the application layer (e.g. override `Http2ConnectionHandler#onStreamError`) before considering baking this into Netty.

In reference to @ejona86's comment https://github.com/netty/netty/issues/8025#issuecomment-397689321 IIUC his suggestion is to ignore the receipt of RST_STREAM on a closed stream, but we already do that [1][2].

[1] https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java#L341-L342
[2] https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java#L349

bryce-anderson:
I'm generally in agreement with @Scottmitch that sending a RST_STREAM is the most desirable behavior from a protocol perspective, though I do think the logging is aggressive. Can we change the log level of that exception? In general, a RST_STREAM with code STREAM_CLOSED (and some others like CANCELED) are not that interesting and probably better to default them to debug level. That said, logs could get complex to interpret if we start special casing by the error code.

I would like to keep the fixes to the PUSH_PROMISE behavior, namely that I think we're currently using the wrong stream id in `shouldIgnoreHeadersOrDataFrame` (we should be using the promised stream id) and we don't increment the last observed inbound id if we reset the promised stream. Of course that could be changed in a different PR.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
shorea: > If the stream may have existed ... tracking/estimating RTT time is error prone and can be expensive for what is in large part a boundary condition. The RFC permits sending multiple RST_STREAM frames, doing so will have no negative implications from a protocol perspective*. A simpler approach would be to continue to send the RST_STREAM frames which provides positive acknowledgement for good peers that this frame was not accepted, and have some heuristic to "fail hard" (e.g. connection error) to catch the case of "misbehaving peers".

How would we trigger a RST_STREAM without throwing a stream error?

shorea: @Scottmitch can you take a look at my latest comment?

Scottmitch: sorry was afk for a few days

> How would we trigger a RST_STREAM without throwing a stream error?

throwing a `StreamException` in the interface based API (e.g. `Http2ConnectionHandler`, `Http2ConnectionDecoder`, `Http2ConnectionEncoder`, etc..) will generally result in a RST_STREAM being sent. Independent of this you can also always call [writeRstStream](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameWriter.java#L121) on the `Http2ConnectionDecoder`.

The code I provided in the previous comment [1] https://github.com/netty/netty/pull/8028#discussion_r197092876 should be sufficient to cover the "If the stream may have existed" and "if the stream was never in existence" case. However this does't cover the heuristic to close the connection if receiving a frame for a stream in a closed state should be a protocol error instead of a stream error, and I think we can handle that as a followup PR.

[1]
```java
if (connection.streamMayHaveExisted(streamId)) {
  throw streamError(streamId, STREAM_CLOSED, "Received %s frame for an unknown stream %d", frameName, streamId);
} else {
  throw connectionError(PROTOCOL_ERROR, "Stream %d does not exist", streamId);
}
```

Scottmitch:
> though I do think the logging is aggressive. Can we change the log level of that exception?

@bryce-anderson - If log level is to high/noisy I'm +1 for reducing it. However based upon the issue description I thought  the logging was being done at a higher level? (/cc @shorea)

shorea:
@Scottmitch @normanmaurer would it be possible to do a netty release this week? A colleague is waiting on another change to be released, I'll have to revisit this PR per Scott's comments but we have a workaround for this particular issue.

normanmaurer:
I can cut one one on Monday I think

shorea:
Thanks @normanmaurer !

shorea:
@normanmaurer I forgot to follow up on this. Would it still be possible to get a release out this week?

normanmaurer:
@shorea yep... working on it as we speak.

normanmaurer:
@shorea done https://netty.io/news/2018/07/10/4-1-26-Final.html

normanmaurer:
@shorea any update ?

shorea:
@normanmaurer  apologies for the long delay. I'll have some time this week to revisit Scott's comments.

normanmaurer:
@shorea ok cool.... ping me once done

goral09:
Guys, what's the update on this? It's been a while and I keep seeing this in my project. Can we push this forward? Do you need any help?

PSNRigner:
@shorea Any news about this issue ? It has been running for months and we would like to see it resolved as it can be spamming logs for production services.

sebtomba:
This PR seems to be approved and checked. When can you merge it?

normanmaurer:
@shorea any update ?

PSNRigner:
It has been 2 months without news of @shorea , not very professional from him.
Can't anyone else get his hands on this issue to get it finally reviewed and done ? Some of us have been waiting for too long.

gagahjt:
when will that version solve the problem

normanmaurer:
Let me see if I can take bring this over the finish line...

rabbitonweb:
Really looking forward to have this merged :)

rubensayshi:
Is the last pending desired change the one suggested by @Scottmitch ?

It seems nobody wants to apply those changes?  
I can give it a shot ... I guess rebasing will be most of the effort at this point ...

> Given the above criteria my preferred way to handle the scenario is the following:
> 
> If the stream may have existed ... tracking/estimating RTT time is error prone and can be expensive for what is in large part a boundary condition. The RFC permits sending multiple RST_STREAM frames, doing so will have no negative implications from a protocol perspective*. A simpler approach would be to continue to send the RST_STREAM frames which provides positive acknowledgement for good peers that this frame was not accepted, and have some heuristic to "fail hard" (e.g. connection error) to catch the case of "misbehaving peers".
> if the stream was never in existence ... protocol error
> reduce the log level in io.grpc.netty.NettyServerHandler for this case (or consider a more selective log strategy)

normanmaurer:
This was re-opened as https://github.com/netty/netty/pull/9402 to make progress.

PR #8029, opened 2018-06-16
Don't fail the deregistration promise in Http2MultiplexCodec

bryce-anderson:
Motivation:

We deviate from the AbstractChannel implementation on deregistration by
failing the provided promise if the channel is already deregistered. In
contrast, AbstractChannel will always set the promise to successfully
done.

Modification:

Change the Http2MultiplexCodec.DefaultHttp2StreamChannel.Http2ChannelUnsafe
to always set the promise provided to deregister as done as is the case in
AbstractChannel.


bryce-anderson:
For reviewing ease, the portion of AbstractChannel that MultiplexCodec is trying to emulate can be found here: https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/AbstractChannel.java#L820

normanmaurer:
@carl-mastrangelo PTAL as you found it :)

PR #8034, opened 2018-06-18
Use java 11+ea18

normanmaurer:
Motivation:

Java 11+ea18 was released.

Modifications:

Update to latest version.

Result:

Testing with latest java 11 release.

### DIFF
--- a/docker/docker-compose.centos-6.111.yaml
+++ b/docker/docker-compose.centos-6.111.yaml
@@ -7,7 +7,7 @@ services:
     build:
       args:
         centos_version : "6"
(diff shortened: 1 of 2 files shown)


normanmaurer:
After this is merged we should be able to also revert 4b728cd5bc53195bced516f33a1ea0a0def5604e.

normanmaurer:
Just verified and as mentioned before we can revert the workaround commit as https://bugs.openjdk.java.net/browse/JDK-8203937 is part of the release.

bryce-anderson:
Also LGTM. Sorry for the late review.

PR #8035, opened 2018-06-18
Revert "Ignore some test-flakiness when using Java11+ due outstanding…

normanmaurer:
… Java11 bug. (#7984)"

Motivation:

This reverts commit 4b728cd5bc53195bced516f33a1ea0a0def5604e as it was fixes in Java 11 ea+17.

Modification:

Revert previous added workaround as this is fixed in Java 11 now.

Result:

No more workaround for test included.


normanmaurer:
This is not needed anymore as it was fixed lately:

https://bugs.openjdk.java.net/browse/JDK-8199329

PR #8036, opened 2018-06-19
Allow to cache keymaterial when using OpenSSL

normanmaurer:
Motiviation:

During profiling it showed that a lot of time during the handshake is spent by parsing the key / chain over and over again. We should cache these parsed structures if possible to reduce the overhead during handshake.

Modification:

- Use new APIs provided by https://github.com/netty/netty-tcnative/pull/360.
- Introduce OpenSslCachingX509KeyManagerFactory which allows to wrap another KeyManagerFactory and caches the key material provided by it.

Result:

In benchmarks handshake times have improved by 30 %.


normanmaurer:
This depends on https://github.com/netty/netty-tcnative/pull/360 and may still need some cleanup, testing etc but I thought it would be good to show it early on.

handler/src/main/java/io/netty/handler/ssl/OpenSslCachingKeyMaterialProvider.java
```diff
@@ -0,0 +1,65 @@
...
+            if (material == null) {
+                // No keymaterial should be used.
+                return null;
+            }
+
+            OpenSslKeyMaterial old = cache.putIfAbsent(alias, material);
+            if (old != null) {
+                // Some other thread cached a material in the meantime, thus we just return the material without
+                // incrementing the reference count. This instance will be just released once we are done with it.
+                return material;
```
Scottmitch: should we release `material`, do `old.retain()`, and return `old`? Seems strange to return an object that won't live in the cache and won't be returned on subsequent calls.

handler/src/main/java/io/netty/handler/ssl/OpenSslKeyMaterial.java
```diff
@@ -0,0 +1,124 @@
...
+    private static final ResourceLeakDetector<OpenSslKeyMaterial> leakDetector =
+            ResourceLeakDetectorFactory.instance().newResourceLeakDetector(OpenSslKeyMaterial.class);
+    private final ResourceLeakTracker<OpenSslKeyMaterial> leak;
+    private long chain;
+    private long privateKey;
+
+    OpenSslKeyMaterial(long chain, long privateKey) {
+        this.chain = chain;
+        this.privateKey = privateKey;
+        leak = leakDetector.track(this);
```
Scottmitch: instead of a constructor should we use a static factory method and optionally apply the leak record like in buffer and other cases with some sample rate?

handler/src/main/java/io/netty/handler/ssl/OpenSslKeyMaterial.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /**
+     * Returns the pointer to the {@code STACK_OF(X509)}.
+     */
+    public long chain() {
```
Scottmitch: `chain` -> `certificateChain` or `certificateChainAddress`?

handler/src/main/java/io/netty/handler/ssl/OpenSslKeyMaterialManager.java
```diff
@@ -79,101 +74,61 @@ void setKeyMaterial(ReferenceCountedOpenSslEngine engine) throws SSLException {
...
-        } catch (Exception e) {
-            throw new SSLException(e);
-        } finally {
-            freeBio(keyBio);
-            freeBio(keyCertChainBio);
-            freeBio(keyCertChainBio2);
+    private String chooseClientAlias(ReferenceCountedOpenSslEngine engine,
+                                       String[] keyTypes, X500Principal[] issuer) {
+        X509KeyManager manager = provider.keyManager();
+        if (manager instanceof X509ExtendedKeyManager) {
```
Scottmitch: did you consider have a class that extends `OpenSslKeyMaterialManager`, overrides these methods, and requires `X509ExtendedKeyManager` to avoid casting here (and below)?

handler/src/main/java/io/netty/handler/ssl/OpenSslKeyMaterialProvider.java
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Provides {@link OpenSslKeyMaterial} for a given alias.
+ */
+class OpenSslKeyMaterialProvider {
+
+    private final X509KeyManager keyManager;
+    private final String password;
+
+    OpenSslKeyMaterialProvider(X509KeyManager keyManager, String password) {
+        this.keyManager = keyManager;
```
Scottmitch: check not null?

handler/src/main/java/io/netty/handler/ssl/OpenSslKeyMaterialProvider.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /**
+     * Will be invoked once the provider should be destroyed.
+     */
+    void destroy() {
+        // NOOP.
+    }
+}
+
```
Scottmitch: nit: kill line

handler/src/main/java/io/netty/handler/ssl/OpenSslStaticX509KeyManagerFactory.java
```diff
@@ -0,0 +1,69 @@
...
+import java.security.cert.X509Certificate;
+
+/**
+ * Wraps another {@link KeyManagerFactory} and caches its chains / certs for an alias for better performance when using
+ * {@link SslProvider#OPENSSL} or {@link SslProvider#OPENSSL_REFCNT}.
+ *
+ * Because of the caching its important that the wrapped {@link KeyManagerFactory}s {@link X509KeyManager}s always
+ * return the same {@link X509Certificate} chain and {@link PrivateKey} for the same alias.
+ */
+public final class OpenSslStaticX509KeyManagerFactory extends KeyManagerFactory {
```
Scottmitch: `OpenSslStaticX509KeyManagerFactory ` -> `OpenSslCachingX509KeyManagerFactory `?

handler/src/main/java/io/netty/handler/ssl/OpenSslCachingKeyMaterialProvider.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+/**
+ * {@link OpenSslKeyMaterialProvider} that will cache the {@link OpenSslKeyMaterial} to reduce the overhead
+ * of parsing the chain and the key for generation of the material.
+ */
+final class OpenSslCachingKeyMaterialProvider extends OpenSslKeyMaterialProvider {
+
+    private final ConcurrentMap<String, OpenSslKeyMaterial> cache = new ConcurrentHashMap<String, OpenSslKeyMaterial>();
```
johnou: io.netty.util.internal.PlatformDependent#newConcurrentHashMap()

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslClientContext.java
```diff
@@ -100,18 +99,23 @@ static OpenSslSessionContext newSessionContext(ReferenceCountedOpenSslContext th
...
                 }
 
-                if (keyManagerFactory != null) {
-                    X509KeyManager keyManager = chooseX509KeyManager(keyManagerFactory.getKeyManagers());
-                    OpenSslKeyMaterialManager materialManager = useExtendedKeyManager(keyManager) ?
-                            new OpenSslExtendedKeyMaterialManager(
-                                    (X509ExtendedKeyManager) keyManager, keyPassword) :
-                            new OpenSslKeyMaterialManager(keyManager, keyPassword);
+                if (keyMaterialProvider != null) {
+                    OpenSslKeyMaterialManager materialManager = new OpenSslKeyMaterialManager(keyMaterialProvider);
```
Scottmitch: `ReferenceCountedOpenSslClientContext#keyMaterialManager()` always returns `null`, but this method is called in `ReferenceCountedOpenSslContext#destroy()` to free up resources. Should we use a similar strategy for initialization as `ReferenceCountedOpenSslServerContext` to make sure this doesn't leak?

handler/src/main/java/io/netty/handler/ssl/OpenSslKeyMaterialManager.java
```diff
@@ -79,101 +74,61 @@ void setKeyMaterial(ReferenceCountedOpenSslEngine engine) throws SSLException {
...
+                            keyMaterial.release();
+                        }
+                    }
                 }
             }
         }
     }
 
-    CertificateRequestedCallback.KeyMaterial keyMaterial(ReferenceCountedOpenSslEngine engine, String[] keyTypes,
+    void setkeyMaterial(ReferenceCountedOpenSslEngine engine, long certOut, long keyOut, String[] keyTypes,
```
Scottmitch: `setkeyMaterial` -> `setKeyMaterial`? Also since this method is specific to client side should this be `setClientKeyMaterial`, and the other method be `setServerKeyMaterial`?

handler/src/main/java/io/netty/handler/ssl/OpenSslKeyMaterialManager.java
normanmaurer: Yeah that said I thought I would just keep it simple for now. Its not clear to me if this would give any gain at all.

handler/src/main/java/io/netty/handler/ssl/OpenSslKeyMaterialProvider.java
normanmaurer: I did not check null as this is only created by internal classes which does never provide a null. I can add an assert if you like tho... Let me know

handler/src/main/java/io/netty/handler/ssl/OpenSslStaticX509KeyManagerFactory.java
normanmaurer: Sure why not... I thought `Static` may be better as it should signal the key material is always "static" and so never change. Let me rename it tho.

handler/src/main/java/io/netty/handler/ssl/OpenSslKeyMaterialManager.java
normanmaurer: Used the same naming as I used in tcnative... Good idea.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslClientContext.java
normanmaurer: Good catch, let me fix it.

handler/src/main/java/io/netty/handler/ssl/OpenSslKeyMaterial.java
normanmaurer: The sampling is done by `track` internally so I think its fine... Its also the same that we do in `ReferenceCountedOpenSslEngine`.

handler/src/main/java/io/netty/handler/ssl/OpenSslCachingKeyMaterialProvider.java
johnou: so it does, iirc Netty used to ship with a version of ConcurrentHashMap for old jres?

normanmaurer: we changed this a while ago.

handler/src/main/java/io/netty/handler/ssl/OpenSslKeyMaterial.java
Scottmitch: `ReferenceCountedOpenSslEngine` doesn't use the `leak` record on the "hot path" so it is mostly just a memory concern ... the class hierarchy was also less flexible when `ReferenceCountedOpenSslEngine` was introduced bcz it had to be put "underneath" `OpenSslEngine` (which already existed).

handler/src/main/java/io/netty/handler/ssl/OpenSslKeyMaterialProvider.java
Scottmitch: If its all internal and it is known to be not null through other means I'm OK with omitting the check.

handler/src/main/java/io/netty/handler/ssl/OpenSslKeyMaterial.java
normanmaurer: I agree... that said I am not convinced it worth it to add  a `LeakAwareOpenSslKeyMaterial` here.... if you are strong on is I can do it of-course.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslClientContext.java
normanmaurer: fixed

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -486,6 +485,12 @@ private void destroy() {
 
                 SSLContext.free(ctx);
                 ctx = 0;
+                OpenSslKeyMaterialProvider provider = sessionContext().provider;
```
Scottmitch: should `OpenSslSessionContext` just have a `destroy()` method on it, and internally it can take care of the `OpenSslKeyMaterialProvider#destroy()`? That way we don't have to expose more internals of the class.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslClientContext.java
```diff
@@ -148,13 +151,13 @@ static OpenSslSessionContext newSessionContext(ReferenceCountedOpenSslContext th
         } catch (Exception e) {
             throw new SSLException("unable to setup trustmanager", e);
```
Scottmitch: now that `OpenSslKeyMaterialProvider` has an associated life cycle should we be more careful about cleaning the object up in the event of an exception? Perhaps we can clean it up in a finally block and null it out if we make it all the way through the method successfully and transfer the ownership to the return value.

(same comment on the server side).

handler/src/main/java/io/netty/handler/ssl/OpenSslKeyMaterial.java
Scottmitch: Not strong on it. Always something we can do as a followup since the class is package private.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslClientContext.java
normanmaurer: done...

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslClientContext.java
```diff
@@ -120,6 +120,9 @@ static OpenSslSessionContext newSessionContext(ReferenceCountedOpenSslContext th
                 }
             }
         } catch (Exception e) {
+            if (keyMaterialProvider != null) {
```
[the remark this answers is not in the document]
normanmaurer: I think this should never happen. But if you want I can also guard against this and just do it in a finally block.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -815,4 +816,26 @@ private static long newBIO(ByteBuf buffer) throws Exception {
...
+     */
+    static OpenSslKeyMaterialProvider providerFor(KeyManagerFactory factory, String password) {
+        X509KeyManager keyManager = ReferenceCountedOpenSslContext.chooseX509KeyManager(factory.getKeyManagers());
+        if (factory instanceof OpenSslCachingX509KeyManagerFactory) {
+            // The user explicit used OpenSslCachingX509KeyManagerFactory which signals us that its fine to cache.
+            return new OpenSslCachingKeyMaterialProvider(keyManager, password);
+        } else {
+            // Try to detect if its safe to cache the material.
+            String className = keyManager.getClass().getName();
+            if (className.equals("sun.security.ssl.X509KeyManagerImpl") ||
```
Scottmitch: I wonder if it is better just to just rely upon the user to wrap with `OpenSslCachingX509KeyManagerFactory` instead of trying to infer based upon class type. Below is the javadoc on `X509KeyManagerImpl`, which is concerning:

>   it is explicitly designed to accommodate KeyStores that change over the lifetime of the process.

If a user isn't taking advantage of this behavior then they can manually wrap in `OpenSslCachingX509KeyManagerFactory`.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslClientContext.java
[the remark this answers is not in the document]
Scottmitch: I would prefer doing it in a finally block ... the semantics are more clear and if the code here (or underneath) then we don't have to worry about memory leaks.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
normanmaurer: @Scottmitch yeah.... I guess better safe then sorry.

normanmaurer:
@Scottmitch @ejona86 @carl-mastrangelo @mosesn @bryce-anderson I would like to include this in the next netty release as it significant improve performance.

See https://github.com/netty/netty-tcnative/pull/360#issuecomment-399221762

handler/src/main/java/io/netty/handler/ssl/OpenSslCachingKeyMaterialProvider.java
```diff
@@ -0,0 +1,64 @@
...
```
mosesn: I'm not sure I understand exactly what key material is.  This looks like a potential memory leak, should I not be worried?

normanmaurer: @mosesn the `OpenSslKeyMaterial` is the combo of certificate chain and private key for an alias that is provided by the `X509KeyManager`. The important thing to note here is that the `OpenSslKeyMaterial` holds the pointers to the openssl structures and so native memory until these are destroyed. Usually you have 1 or two aliases. For example one for `rsa` and one for `ec` if you have dual chains or just one of these (which is usually the case). 

The `OpenSslKeyMaterial` and so the memory is released when `destroy()` is called, which is done once the `OpenSslContext` is destroyed. 

Does this make it more clear ?

normanmaurer: just to make it even clearer... this only caches what your X509KeyManager returns so it can not be "abused" by a remote peer.

mosesn:
This looks good in general, can you please add tests?

normanmaurer:
@mosesn It’s tested via the *SSLEngineTest stuff but let me think about some more that I can add.

normanmaurer:
@mosesn added some more tests for you...

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslServerContext.java
```diff
@@ -107,87 +105,97 @@ static ServerContext newSessionContext(ReferenceCountedOpenSslContext thiz, long
                                            String keyPassword, KeyManagerFactory keyManagerFactory)
             throws SSLException {
         ServerContext result = new ServerContext();
+        OpenSslKeyMaterialProvider keyMaterialProvider = null;
         try {
-            SSLContext.setVerify(ctx, SSL.SSL_CVERIFY_NONE, VERIFY_DEPTH);
```
normanmaurer: note that most of the changes here is just because I added another level of `try {...} catch { ...}`

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslClientContext.java
```diff
@@ -90,71 +88,82 @@ static OpenSslSessionContext newSessionContext(ReferenceCountedOpenSslContext th
             throw new IllegalArgumentException(
                     "Either both keyCertChain and key needs to be null or none of them");
         }
+        OpenSslKeyMaterialProvider keyMaterialProvider = null;
         try {
```
normanmaurer: note that most of the changes here are just because I added another level of `try {...} catch { ...}`

normanmaurer:
@carl-mastrangelo haha... So I will just blame you if I broke something =P

handler/src/main/java/io/netty/handler/ssl/OpenSslCachingKeyMaterialProvider.java
```diff
@@ -0,0 +1,64 @@
...
+        return material.retain();
+    }
+
+    @Override
+    void destroy() {
+        // This should be fine as after destroy is called we do not expect to have any more calls to chooseKeyMaterial.
+        for (OpenSslKeyMaterial material : cache.values()) {
+            material.release();
+        }
+        cache.clear();
```
ejona86: FYI: it is possible to make the release()+clear() atomic by using `Iterator.remove()` while iterating over the values. No need to change though.

handler/src/main/java/io/netty/handler/ssl/OpenSslSessionContext.java
```diff
@@ -35,13 +35,15 @@
...
+    private final OpenSslKeyMaterialProvider provider;
 
     // IMPORTANT: We take the OpenSslContext and not just the long (which points the native instance) to prevent
     //            the GC to collect OpenSslContext as this would also free the pointer and so could result in a
     //            segfault when the user calls any of the methods here that try to pass the pointer down to the native
     //            level.
-    OpenSslSessionContext(ReferenceCountedOpenSslContext context) {
+    OpenSslSessionContext(ReferenceCountedOpenSslContext context, OpenSslKeyMaterialProvider provider) {
         this.context = context;
+        this.provider = provider;
```
ejona86: Since this variable is private and not used (other than destruction), it would seem a good idea to have a comment mention what is actually using this object elsewhere and why that corresponds to this object's lifetime.

normanmaurer:
@ejona86 done

PR #8037, opened 2018-06-20
fix grammar in javadoc

radai-rosenblatt:


normanmaurer:
@radai-rosenblatt thanks!

PR #8038, opened 2018-06-20
Reduce array allocations during CompositeByteBuf construction

njhill:
Motivation:

Eliminate avoidable backing array reallocations when constructing composite `ByteBuf`s from existing buffer arrays/`Iterable`s. This also applies to the `Unpooled.wrappedBuffer(...)` methods.

Modifications:

Ensure the initial components `ComponentList` is sized at least as large as the provided buffer array/`Iterable` in the `CompositeByteBuffer` constructors.

In single-arg `Unpooled.wrappedBuffer(...)` methods, set `maxNumComponents` to the count of provided buffers, rather than a fixed default of 16. It seems likely that most usage of these involves wrapping a list without subsequent modification, particularly since they return a `ByteBuf` rather
than `CompositeByteBuf`. If a different/larger max is required there are already the `wrappedBuffer(int, ...)` variants.

In fact the current behaviour could be considered inconsistent - if you call `Unpooled.wrappedBuffer(int, ByteBuf)` with a single buffer, you might expect to subsequently be able to add buffers to it (since you specified a max related to consolidation), but it will in fact return just a slice of the provided buffer.

Result:

Fewer and smaller allocations in some cases when using `CompositeByteBuf`s or `Unpooled.wrappedBuffer(...)` methods.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
@@ -61,7 +61,7 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
         this.alloc = alloc;
         this.direct = direct;
         this.maxNumComponents = maxNumComponents;
-        components = newList(maxNumComponents);
+        components = newList(0, maxNumComponents);
     }
 
     public CompositeByteBuf(ByteBufAllocator alloc, boolean direct, int maxNumComponents, ByteBuf... buffers) {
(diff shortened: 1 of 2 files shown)


normanmaurer:
@njhill can you sign our ICLA and let me know once you did it ?

https://netty.io/s/icla

njhill:
@normanmaurer I have already signed it, you merged another PR of mine recently :)

normanmaurer:
@njhill doh missed it... Just merged. Great work!

vincent-grosbois:
Hello!
could this change be backported into netty 3.10 ?
thanks!

normanmaurer:
@vincent-grosbois honestly 3.x is EOL for years so I think we will not do any more back ports. Sorry

vincent-grosbois:
actually a mistake on my part sorry... I'm trying to fix https://issues.apache.org/jira/browse/SPARK-24917
which is solved by this specific commit.
As a result I'd like the netty version used by Spark 2.2 and 2.3 to receive this fix, but I'm unable to find out which version it is (3.x is really old though so probably it's not that one)

vincent-grosbois:
After checking it seems this version is 4.0
Do you think it's possible to backport this fix in netty 4.0?
thanks!

PR #8040, opened 2018-06-20
Support composite buffer creation without array alloc and copy

njhill:
Motivation:

`Unpooled.unmodifiableBuffer()` is currently used to efficiently write arrays of `ByteBuf`s via `FixedCompositeByteBuf`, but involves an allocation and content-copy of the provided `ByteBuf` array which in many (most?) cases shouldn't be necessary.

Modifications:

Modify the internal `FixedCompositeByteBuf` class to support wrapping the provided `ByteBuf` array directly. Control this behaviour with a constructor flag and expose the "unsafe" version via a new
`Unpooled.wrappedUnmodifiableBuffer(ByteBuf...)` method.

Result:

Less garbage on IO paths. I would guess pretty much all existing usage of `unmodifiableBuffer()` could use the copy-free version but assume it's not safe to change its default behaviour.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/FixedCompositeByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/FixedCompositeByteBuf.java
@@ -27,6 +27,7 @@ import java.nio.ReadOnlyBufferException;
 import java.nio.channels.FileChannel;
 import java.nio.channels.GatheringByteChannel;
 import java.nio.channels.ScatteringByteChannel;
+import java.util.Arrays;
 import java.util.Collections;
 
 /**
@@ -39,10 +40,10 @@ final class FixedCompositeByteBuf extends AbstractReferenceCountedByteBuf {
     private final int capacity;
     private final ByteBufAllocator allocator;
     private final ByteOrder order;
-    private final Object[] buffers;
+    private final ByteBuf[] buffers;
     private final boolean direct;
 
-    FixedCompositeByteBuf(ByteBufAllocator allocator, ByteBuf... buffers) {
+    FixedCompositeByteBuf(ByteBufAllocator allocator, boolean copy, ByteBuf... buffers) {
         super(AbstractByteBufAllocator.DEFAULT_MAX_CAPACITY);
         if (buffers.length == 0) {
             this.buffers = EMPTY;
@@ -52,8 +53,7 @@ final class FixedCompositeByteBuf extends AbstractReferenceCountedByteBuf {
             direct = false;
         } else {
             ByteBuf b = buffers[0];
-            this.buffers = new Object[buffers.length];
-            this.buffers[0] = b;
+            this.buffers = copy ? Arrays.copyOf(buffers, buffers.length) : buffers;
             boolean direct = true;
             int nioBufferCount = b.nioBufferCount();
             int capacity = b.readableBytes();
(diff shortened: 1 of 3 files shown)


njhill:
Added commit with small optimization to copy case, to avoid reflection in `Arrays.copyOf()`.

By the way I'm not sure `wrappedUnmodifiableBuffer()` is a great name for the new method, maybe there's a more appropriate one :)

njhill:
Thanks @normanmaurer. If this is accepted, would it be reasonable to consider deprecating the existing method to discourage use in favour of the new one? In particular anywhere the varargs are used can definitely use the new one.

If there were unusual cases where an explicit array was repurposed before the created composite buffer was finished with, it could just be copied before wrapping. The new behaviour is more analogous to things like  `Arrays.asList()`.

normanmaurer:
@njhill yeah I think that sounds good... Are you happy to open another PR after this one is merged ?

njhill:
@normanmaurer certainly. Any suggestions for better name of the new method?

normanmaurer:
@carl-mastrangelo @ejona86 @mosesn @bryce-anderson naming ideas ?

buffer/src/main/java/io/netty/buffer/FixedCompositeByteBuf.java
```diff
@@ -52,8 +52,12 @@
             direct = false;
         } else {
             ByteBuf b = buffers[0];
-            this.buffers = new Object[buffers.length];
-            this.buffers[0] = b;
+            if (copy) {
+                this.buffers = new ByteBuf[buffers.length];
+                System.arraycopy(buffers, 0, this.buffers, 0, buffers.length);
```
carl-mastrangelo: Optional: consider using Arrays.copyOf, which I think avoids zeroing of the array.

buffer/src/main/java/io/netty/buffer/FixedCompositeByteBuf.java
```diff
@@ -39,10 +39,10 @@
     private final int capacity;
     private final ByteBufAllocator allocator;
     private final ByteOrder order;
-    private final Object[] buffers;
+    private final ByteBuf[] buffers;
     private final boolean direct;
 
-    FixedCompositeByteBuf(ByteBufAllocator allocator, ByteBuf... buffers) {
+    FixedCompositeByteBuf(ByteBufAllocator allocator, boolean copy, ByteBuf... buffers) {
```
carl-mastrangelo: As an alternative, why not make this never copy, and have callsites copy when they need to?  They have to know anyways in order to set `copy` properly.

buffer/src/main/java/io/netty/buffer/FixedCompositeByteBuf.java
njhill: I actually used that in my first pass (see initial commit), but realized it uses reflection which can be avoided this way.

Unless JDK impls differ on this, it does not look like Arrays.copyOf avoids zeroing the array. I think the closest thing to that is java 9's `Unsafe.allocateUninitializedArray()`, but that's only for primitives.

carl-mastrangelo: JDK has an instrinsic for Arrays.copyOf (and copyOfRange).  Protobuf uses it depending on if running Android or not:

https://github.com/google/protobuf/blob/9021f623e1420f513268a01a5ad43a23618a84ba/java/core/src/main/java/com/google/protobuf/ByteString.java#L94

njhill: Interesting! Thanks for schooling me @carl-mastrangelo, I will change it to use this.

normanmaurer:
@carl-mastrangelo PTAL again as you had the last comment :)

normanmaurer:
Thanks a lot @njhill ... I squashed the commits and cherry-picked as f164759ea

PR #8046, opened 2018-06-22
HpackDecoder treats invalid pseudo-headers as stream level errors

### LINKED ISSUE #8043: Invalid HTTP2 pseudo-headers kill the session instead of being treated as malformed
### Expected behavior

When a peer receives invalid pseudo headers the request or response should be treated as malformed as prescribed in section [8.1.2.1](https://http2.github.io/http2-spec/#rfc.section.8.1.2.1). This entails a stream level PROTOCOL_ERROR as described in section [8.1.2.6](https://http2.github.io/http2-spec/#rfc.section.8.1.2.6)

### Actual behavior

The [HpackDecoder](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java#L401) throws a connection level PROTOCOL_ERROR mid-decode, resulting in the connection being torn down.

### Steps to reproduce

Attempt to send a HTTP/2 request with an unknown pseudo header, pseudo headers out of order, or a mix of request and response type pseudo headers.

bryce-anderson:
Motivation:

The HTTP/2 spec dictates that invalid pseudo-headers should cause the
request/response to be treated as malformed (8.1.2.1), and the recourse
for that is to treat the situation as a stream error of type
PROTOCOL_ERROR (8.1.2.6). However, we're treating them as a connection
error with the connection being immediately torn down and the HPACK
state potentially being corrupted.

Modifications:

The HpackDecoder now throws a StreamException for validation failures
and throwing is deffered until the end of of the decode phase to ensure
that the HPACK state isn't corrupted by returning early.

Result:

Behavior more closely aligned with the HTTP/2 spec.

Fixes #8043.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
@@ -44,6 +44,7 @@ import static io.netty.handler.codec.http2.Http2CodecUtil.headerListSizeExceeded
 import static io.netty.handler.codec.http2.Http2Error.COMPRESSION_ERROR;
 import static io.netty.handler.codec.http2.Http2Error.PROTOCOL_ERROR;
 import static io.netty.handler.codec.http2.Http2Exception.connectionError;
+import static io.netty.handler.codec.http2.Http2Exception.streamError;
 import static io.netty.handler.codec.http2.Http2Headers.PseudoHeaderName.getPseudoHeader;
 import static io.netty.handler.codec.http2.Http2Headers.PseudoHeaderName.hasPseudoHeaderFormat;
 import static io.netty.util.AsciiString.EMPTY_STRING;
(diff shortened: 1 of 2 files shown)


codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
```diff
@@ -119,24 +120,25 @@
      * This method assumes the entire header block is contained in {@code in}.
      */
     public void decode(int streamId, ByteBuf in, Http2Headers headers, boolean validateHeaders) throws Http2Exception {
-        Http2HeadersSink sink = new Http2HeadersSink(headers, maxHeaderListSize);
-        decode(in, sink, validateHeaders);
+        Http2HeadersSink sink = new Http2HeadersSink(streamId, headers, maxHeaderListSize, validateHeaders);
+        decode(in, sink);
 
-        // we have read all of our headers. See if we have exceeded our maxHeaderListSize. We must
+        // we have read all of our headers so now perform validation steps. We must
```
mosesn: this switches to the imperative mid-sentence, which feels awkward

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
```diff
@@ -119,24 +120,25 @@
      * This method assumes the entire header block is contained in {@code in}.
      */
     public void decode(int streamId, ByteBuf in, Http2Headers headers, boolean validateHeaders) throws Http2Exception {
-        Http2HeadersSink sink = new Http2HeadersSink(headers, maxHeaderListSize);
-        decode(in, sink, validateHeaders);
+        Http2HeadersSink sink = new Http2HeadersSink(streamId, headers, maxHeaderListSize, validateHeaders);
```
Scottmitch: I know it wasn't introduced by this PR but it would be nice to not have to introduce object allocation for each decode. Seems like this can be handled as local state in the now private `decode` method.

bryce-anderson: Do you mind if we defer this for a later commit? I think the private `decode` method is really heavy at this time and don't really want to add more state to it, at least as part of another behavior change.

fwiw, I suspect this is a really good candidate for stack allocation and even if it's not, it's super short lived garbage that only happens once on a complete HEADERS block.

normanmaurer: I am +1 to defer this for a followup as its not introduced by this change

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
```diff
@@ -551,6 +556,14 @@ public void appendToHeaderList(CharSequence name, CharSequence value) {
             if (!exceededMaxLength) {
                 headers.add(name, value);
             }
+            // Only consider the first validation exception
+            if (validate && validationException == null) {
```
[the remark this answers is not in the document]
bryce-anderson: I think I must be misunderstanding your suggestion. If it was an `else if` with the conditional above it would not validate unless `exceededMaxLength == true`, in which case it doesn't matter.

I was deliberate in letting the header be added even if validation was going to fail since there is a world where we would be interested in the results, but I could see wanting to refactor the checks such that we only add the header if it passes both checks.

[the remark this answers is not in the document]
bryce-anderson: In the common (maybe all) case they will just be discarded. For the case of `exceededMaxLength == true` it made sense not to store it since it was too big, but the invalid pseudo-header case is obvious. I'll change it to not add if it fails either validation if for nothing else than to be consistent.

normanmaurer:
@bryce-anderson thanks!

Issue #8023, opened 2018-06-14
Math.random usage in Websocket code


hc-codersatlas:
Hi, minor issue, but consistency / performance reasons WebSocketUtil and related handshakers make use of Math.random instead of ThreadLocalRandom. Math.random is pretty slow...

normanmaurer:
@hc-codersatlas PRs welcome...

hc-codersatlas:
Sure, but have a question about that part of the code in WebSocketUtil (and a few other files):

`    static int randomNumber(int minimum, int maximum) {
        return (int) (Math.random() * maximum + minimum);
}`

Is this correct? Assuming Math.random() produces 0.9999 (for convenience let's say 1.0) - the result exceeds the maximum value, i.e. 1.0 * maximum + minimum > maximum.

normanmaurer:
looks like a bug to me... PR with fix and test welcome :)

normanmaurer:
Fixed by https://github.com/netty/netty/pull/8048

PR #8048, opened 2018-06-22
Fix random number generator in WebSocketUtil

kachayev:
Motivation:

Implementation of `WebSocketUtil/randomNumber` is incorrect and might violate
the API by returning a value > maximum specified.

Modifications:

* `WebSocketUtil/randomNumber` is reimplemented, the idea of the solution described
  in the comment in the code

* Implementation of `WebSocketUtil/randomBytes` changed to `nextBytes` method

* `ThreadLocalRandom` is used instead of `Math.random` to improve efficiency

Result:

`WebSocketUtil/randomNumber` always produces the correct result.

Covers #8023

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketUtil.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketUtil.java
@@ -20,6 +20,7 @@ import io.netty.buffer.Unpooled;
 import io.netty.handler.codec.base64.Base64;
 import io.netty.util.CharsetUtil;
 import io.netty.util.concurrent.FastThreadLocal;
+import io.netty.util.internal.PlatformDependent;
 
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
@@ -105,11 +106,7 @@ final class WebSocketUtil {
      */
     static byte[] randomBytes(int size) {
         byte[] bytes = new byte[size];
-
-        for (int index = 0; index < size; index++) {
-            bytes[index] = (byte) randomNumber(0, 255);
-        }
-
+        PlatformDependent.threadLocalRandom().nextBytes(bytes);
         return bytes;
     }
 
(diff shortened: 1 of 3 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketUtil.java
```diff
@@ -105,11 +106,7 @@ static String base64(byte[] data) {
      */
     static byte[] randomBytes(int size) {
         byte[] bytes = new byte[size];
-
-        for (int index = 0; index < size; index++) {
-            bytes[index] = (byte) randomNumber(0, 255);
-        }
-
+        ThreadLocalRandom.current().nextBytes(bytes);
```
normanmaurer: Use `PlatformDependent.threadLocalRandom()`

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketUtil.java
```diff
@@ -121,7 +118,28 @@ static String base64(byte[] data) {
      * @return A pseudo-random number
      */
     static int randomNumber(int minimum, int maximum) {
-        return (int) (Math.random() * maximum + minimum);
+        double fraction = ThreadLocalRandom.current().nextDouble();
```
normanmaurer: Same as above.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketUtil.java
kachayev: Oh, my bad 🤦‍♂️

codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketUtilTest.java
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright 2016 The Netty Project
```
normanmaurer: 2018

kachayev:
@normanmaurer Fixed!

normanmaurer:
@kachayev thanks!

Issue #8017, opened 2018-06-12
Object cleaner implementation detail leaks 


jasontedor:
This is a follow on to the discussion in #8014. The problem here is that the object cleaner thread is considered an internal implementation detail yet its details can leak.

In Elasticearch, we have a thread leak control that detects threads created by a test suite but are not stopped at the end of the test suite. This is part of broader detection in our test framework for leaked resources of any kind.

The problem here is that because the object cleaner thread can not be stopped or otherwise controlled, it fails our thread leak control as it will always be running in nearly any test suite that interacts with Netty.

We tried to upgrade to Netty 4.1.25 and it was quite a mess because of this thread: elastic/elasticsearch#31232. We tried putting a filter in place on relevant test suites but we ultimately decided the surface area here is too large and this is gross anyway, especially without any public API guarantees about the thread. We have ultimately decided to revert this upgrade (elastic/elasticsearch#31282) and are now considering ourselves blocked on upgrading until a better solution to the object cleaner thread problem can be found.

We are looking for a way to shutdown the cleaner (ideally by adding something to our shutdown code ([HTTP](https://github.com/elastic/elasticsearch/blob/f88b9e83cfdac7010736fd221282f1986d3b1b39/modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpServerTransport.java#L335-L357), [TCP](https://github.com/elastic/elasticsearch/blob/f88b9e83cfdac7010736fd221282f1986d3b1b39/modules/transport-netty4/src/main/java/org/elasticsearch/transport/netty4/Netty4Transport.java#L280-L304)) so that all tests pick it up and it's not as messy as needing to filter the thread from leak control everywhere).

Alternatively, a way to avoid the use of the object cleaner would be acceptable too.

normanmaurer:
@mosesn @bryce-anderson @ejona86 @carl-mastrangelo @trustin I would be interested in your take related to this one...

carl-mastrangelo:
Some ideas: maybe the ThreadFactory for the cleaner can be provided by the user?

ejona86:
Has this become an issue now because Recycler was converted to use ObjectCleaner to remove the finalizer? It doesn't sound like this is a case of FastThreadLocal needing 

Unfortunately there is no limit to the amount of time the thread may remain alive, because it requires a GC to be able to exit.

Brainstorm:
[...]

PR #8051, opened 2018-06-25
Avoid ObjectCleaner in FastThreadLocal when onRemoval is ignored

ejona86:
Motivation:

As discussed in #8017, ObjectCleaner leaks into the rest of the environment and
interacts with checks in some testing environments.

Modifications:

Avoid using ObjectCleaner in FastThreadLocal when the extending class is
ignoring removal notifications.

Result:

Reduced usage of ObjectCleaner, and when combined with disabling Recycler, may
allow tests to run without ObjectCleaner being used.

------

I've not actually tested that a sane selection of tests can now run without ObjectCleaner,
but even if we wouldn't be 100% there, this still seems like an easy step in the right
direction.

CC @jasontedor, @normanmaurer, @carl-mastrangelo, @hc-codersatlas, @vkostyukov

### DIFF
--- a/common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
+++ b/common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
@@ -123,10 +123,22 @@ public class FastThreadLocal<V> {
         variablesToRemove.remove(variable);
     }
 
+    // visible for tests
+    static boolean hasOnRemoval(Class<? extends FastThreadLocal> clazz) {
+      try {
+        Class<?> declaringClass = clazz.getMethod("onRemoval", Object.class).getDeclaringClass();
+        return !declaringClass.equals(FastThreadLocal.class);
+      } catch (Throwable t) {
+        return true;
+      }
+    }
+
     private final int index;
+    private final boolean hasOnRemoval;
 
     public FastThreadLocal() {
         index = InternalThreadLocalMap.nextVariableIndex();
+        hasOnRemoval = hasOnRemoval(getClass());
     }
 
     /**
@@ -147,7 +159,8 @@ public class FastThreadLocal<V> {
 
     private void registerCleaner(final InternalThreadLocalMap threadLocalMap) {
         Thread current = Thread.currentThread();
-        if (FastThreadLocalThread.willCleanupFastThreadLocals(current) || threadLocalMap.isCleanerFlagSet(index)) {
+        if (!hasOnRemoval || FastThreadLocalThread.willCleanupFastThreadLocals(current)
+                || threadLocalMap.isCleanerFlagSet(index)) {
             return;
         }
 
--- a/common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
+++ b/common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
@@ -18,6 +18,8 @@ package io.netty.util.concurrent;
 import io.netty.util.internal.InternalThreadLocalMap;
 import io.netty.util.internal.ObjectCleaner;
 import io.netty.util.internal.PlatformDependent;
+import io.netty.util.internal.logging.InternalLogger;
+import io.netty.util.internal.logging.InternalLoggerFactory;
 
 import java.util.Collections;
 import java.util.IdentityHashMap;
@@ -44,6 +46,7 @@ import java.util.Set;
  */
 public class FastThreadLocal<V> {
 
+    private static final InternalLogger logger = InternalLoggerFactory.getInstance(FastThreadLocal.class);
     private static final int variablesToRemoveIndex = InternalThreadLocalMap.nextVariableIndex();
 
     /**
@@ -124,13 +127,20 @@ public class FastThreadLocal<V> {
     }
 
     // visible for tests
-    static boolean hasOnRemoval(Class<? extends FastThreadLocal> clazz) {
-      try {
-        Class<?> declaringClass = clazz.getMethod("onRemoval", Object.class).getDeclaringClass();
-        return !declaringClass.equals(FastThreadLocal.class);
-      } catch (Throwable t) {
-        return true;
-      }
+    static boolean hasOnRemoval(Class<? extends FastThreadLocal> origClass) {
+        for (Class<?> clazz = origClass; clazz != FastThreadLocal.class; clazz = clazz.getSuperclass()) {
+            try {
+                // Can't use getMethod since onRemoval is protected
+                clazz.getDeclaredMethod("onRemoval", Object.class);
+                return true;
+            } catch (NoSuchMethodException ignore) {
+                continue;
+            } catch (Throwable t) {
+                logger.info("Failed to determine if onRemoval is overridden", t);
+                return true;
+            }
+        }
+        return false;
     }
 
     private final int index;
@@ -296,5 +306,6 @@ public class FastThreadLocal<V> {
     /**
      * Invoked when this thread local variable is removed by {@link #remove()}.
      */
+    // This method is referenced via reflection in hasOnRemoval()
     protected void onRemoval(@SuppressWarnings("UnusedParameters") V value) throws Exception { }
 }
--- a/common/src/test/java/io/netty/util/concurrent/FastThreadLocalTest.java
+++ b/common/src/test/java/io/netty/util/concurrent/FastThreadLocalTest.java
@@ -137,6 +137,18 @@ public class FastThreadLocalTest {
         assertEquals(4, ObjectCleaner.getLiveSetCount() - sizeWhenStart);
     }
 
+    public void testDetectOnRemoval_whenPresent() {
+        assertThat(FastThreadLocal.hasOnRemoval(TestFastThreadLocal.class), is(true));
+    }
+
+    public void testDetectOnRemoval_whenAbsent() {
+        class ExtendingFastThreadLocal<T> extends FastThreadLocal<T> {
+        }
+
+        assertThat(FastThreadLocal.hasOnRemoval(FastThreadLocal.class), is(false));
+        assertThat(FastThreadLocal.hasOnRemoval(ExtendingFastThreadLocal.class), is(false));
+    }
+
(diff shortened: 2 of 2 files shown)


common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
```diff
@@ -123,10 +123,22 @@ private static void removeFromVariablesToRemove(
         variablesToRemove.remove(variable);
     }
 
+    // visible for tests
+    static boolean hasOnRemoval(Class<? extends FastThreadLocal> clazz) {
+      try {
+        Class<?> declaringClass = clazz.getMethod("onRemoval", Object.class).getDeclaringClass();
+        return !declaringClass.equals(FastThreadLocal.class);
+      } catch (Throwable t) {
```
normanmaurer: t -> ignore

common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
```diff
@@ -123,10 +123,22 @@ private static void removeFromVariablesToRemove(
         variablesToRemove.remove(variable);
     }
 
+    // visible for tests
+    static boolean hasOnRemoval(Class<? extends FastThreadLocal> clazz) {
```
normanmaurer: Also we use 4 spaces not 2.

common/src/test/java/io/netty/util/concurrent/FastThreadLocalTest.java
```diff
@@ -137,6 +137,18 @@ public void run() {
         assertEquals(4, ObjectCleaner.getLiveSetCount() - sizeWhenStart);
     }
 
+    public void testDetectOnRemoval_whenPresent() {
```
normanmaurer: Missing `@Test` annotation

common/src/test/java/io/netty/util/concurrent/FastThreadLocalTest.java
```diff
@@ -137,6 +137,18 @@ public void run() {
...
+        assertThat(FastThreadLocal.hasOnRemoval(TestFastThreadLocal.class), is(true));
+    }
+
+    public void testDetectOnRemoval_whenAbsent() {
```
normanmaurer: Missing `@Test` annotation

common/src/test/java/io/netty/util/concurrent/FastThreadLocalTest.java
```diff
@@ -137,6 +137,18 @@ public void run() {
...
+        class ExtendingFastThreadLocal<T> extends FastThreadLocal<T> {
+        }
+
+        assertThat(FastThreadLocal.hasOnRemoval(FastThreadLocal.class), is(false));
```
normanmaurer: just use `assertFalse` ?

common/src/test/java/io/netty/util/concurrent/FastThreadLocalTest.java
```diff
@@ -137,6 +137,18 @@ public void run() {
...
+        assertThat(FastThreadLocal.hasOnRemoval(ExtendingFastThreadLocal.class), is(false));
```
normanmaurer: just use `assertFalse` ?

common/src/test/java/io/netty/util/concurrent/FastThreadLocalTest.java
```diff
@@ -137,6 +137,18 @@ public void run() {
         assertEquals(4, ObjectCleaner.getLiveSetCount() - sizeWhenStart);
     }
 
+    public void testDetectOnRemoval_whenPresent() {
+        assertThat(FastThreadLocal.hasOnRemoval(TestFastThreadLocal.class), is(true));
```
normanmaurer: just use `assertTrue` ?

common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
normanmaurer: I wonder if we want to log something

common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
```diff
@@ -123,10 +123,22 @@ private static void removeFromVariablesToRemove(
...
+      try {
+        Class<?> declaringClass = clazz.getMethod("onRemoval", Object.class).getDeclaringClass();
```
bryce-anderson: I know you added some tests for this already, but should we note that the method name itself is important due to reflective usage at the `onRemoval` declaration?

vkostyukov:
That looks totally reasonable to me so 👍 (pending opened comments from others).

common/src/main/java/io/netty/util/concurrent/FastThreadLocal.java
```diff
@@ -123,10 +126,29 @@ private static void removeFromVariablesToRemove(
         variablesToRemove.remove(variable);
     }
 
+    // visible for tests
+    static boolean hasOnRemoval(Class<? extends FastThreadLocal> origClass) {
+        for (Class<?> clazz = origClass; clazz != FastThreadLocal.class; clazz = clazz.getSuperclass()) {
+            try {
+                // Can't use getMethod since onRemoval is protected
```
normanmaurer: didn't know that :(

normanmaurer: @ejona86 couldn't you use the following ?

```java
static boolean hasOnRemoval(Class<? extends FastThreadLocal> origClass) {
    try {
        origClass.getMethod("onRemoval").getDeclaringClass() != FastThreadLocal.class
    } catch (Throwable cause) {
        // logging.....
        return true;
    }
}

```

normanmaurer: actually seems like you tried something like this... Can you explain why this not works ?

normanmaurer: @ejona86 I guess what you could also do is loop through all the classes as you did and call:

`getDeclaredMethods()` on each class and check if `onRemoval` exists there. This should work  with protected as well. See https://docs.oracle.com/javase/7/docs/api/java/lang/Class.html#getDeclaredMethods()

ejona86: `getMethod()` does not return protected methods, so it throws NoSuchMethodException virtually all the time.

From getMethod's docs:
> Returns a Method object that reflects the specified _public_ member method...

getDeclaredMethod does return protected methods, but it doesn't recurse to the superclass when the method is not found, so the looping was required.

ejona86: > I guess what you could also do is loop through all the classes as you did and call: getDeclaredMethods() on each class and check if onRemoval exists there.

Isn't that what I did?

normanmaurer: @ejona86 nope... This is what I would do:

```
    static boolean hasOnRemoval(Class<? extends FastThreadLocal> origClass) {
        for (Class<?> clazz = origClass; clazz != FastThreadLocal.class; clazz = clazz.getSuperclass()) {
            try {
                for (Method method: clazz.getDeclaredMethods()) {
                    if ("onRemoval".equals(method.getName())) {
                        Class[] paramTypes = method.getParameterTypes();
                        if (paramTypes.length == 1 && paramTypes[0] == Object.class) {
                            return true;
                        }
                    }
                }
            } catch (Throwable t) {
                logger.info("Failed to determine if onRemoval is overridden", t);
                return true;
            }
        }
        return false;
    }
```

This will not need to cause any exceptions

jasontedor: I will look more closely tomorrow but I think this will interact poorly with a security manager and so unlikely to be a solution for us. We require the security manager, run our tests with one, and are going to be unwilling to grant Netty the `accessDeclaredMembers` runtime permission (I think that is what will be needed here).

normanmaurer: @jasontedor thanks a lot...

johnou: could we use a marker interface?

normanmaurer: @jasontedor I am correct that the problem you mentioned is true for my suggested code but also for the one that is used in this PR, correct ?

ejona86: Tagging interface is a bit hard, because existing users of the class won't use it. The reason the reflection is necessary is to continue supporting the current API.

We could create a NoRemovalNotificationFastThreadLocal (NRNFTL; "pending marketing approval") that overrides onRemoval and makes the method `final`. Then FastThreadLocal could see if the current instance extends NRNFTL. That does mean most existing users of FastThreadLocal would need to be changed, which I think does make it a less-nice change. The in-tree ones could be changed to it easily, though.

The current API is unfortunate, because `onRemoval()` usage is the uncommon case. There's only three in-tree users: `PooledByteBufAllocator`, `Recycler`, and epoll's `NativeDatagramPacketArray`. PooledByteBufAllocator could avoid `get()ing` from the thread local when `useCacheForAllThreads == true`. Recycler could be disabled by default, or replaced (#8052). The `NativeDatagramPacketArray` code looks like it may only be called from the event loop.

normanmaurer: @ejona86 I agree... that said we could also mark `onRemoval` as `@Deprecated` and only handle it from the `FastLocalThread` (we did the same before which I "fixed").  I am somewhat sure we could remove the `onRemoval` usage in our internal use-cases in many if not in all cases. 

Only partial handle onRemoval is a bit unfortunate but maybe the best bet.

WDYT ?

Also /cc @jasontedor @johnou

ejona86:
Closing in favor of #8064

PR #5968, opened 2016-11-01
[#5904] Disable the Recycler by default

normanmaurer:
Motivation:

The Recycler was introduced as a light-weight object pool, to reduce GC pressure. We use it in various places, like pooling ByteBuf objects (not the actual memory, just the container), write tasks (When you call write(...) from outside the EventLoop) and many more.

This usage gave us a lot of trouble in the past and also users a hard time when they ended up with a lot of objects in the Reycler due writing to fast etc.

Modifications:

Disable the recycler by default so users need to explicit enable it to use it.

Result:

Recycler is not used anymore by default.

PR #8052, opened 2018-06-25
RFC: Reimplement Recycler to reduce impact of pathological cases

ejona86:
Motivation:

Recycler is hard to tune and reduces performance in same cases. Inspired by the
simplicity and large performance gains seen with Go's sync.Pool, which only
reuses objects between GCs, it seemed a more conservative approach could be
applied to Recycler. Note that this looks nothing like sync.Pool, as Recycler
has _very_ different trade-offs and environment.

Modifications:

Reimplement Recycler as a weak reference free list, with low rate of
recycler-enabled objects and heuristics for reusing older objects. Add a
nanobenchmark to better contrast the implementation options, especially with
the 'gc' profiler.

Result:

A Recycler that can improve performance while worst-case would have neglible
impact to GC, memory, and CPU use.

------

There's no rush for this and it wouldn't bother me to close it. I couldn't find the issue where we were talking about disabling the Recycler. This PR should not be considered an argument against doing that; this was a "fun thing" that had interesting enough results I thought I should share. _I honestly don't know if it's worth anyone's time to look into this as a "solution" to Recycler._

I've only run some nanobenchmarks against this. The "generational" behavior in particular is unproven, although it looks interesting. It's possible it is unnecessary and just a single LIFO linked list would be enough, but I went ahead with it because it was the more conservative approach and shouldn't be any worse than the single list approach.

I was a bit worried about whether having so many weak references could cause GC slowdowns; weak refs aren't something I tend to see at large scale. But I think many such concerns are addressed by the 1% sampling.

Although it didn't show up in the code at all, I should note it is _really_ easy to add statistics to the linked list like its length or how reused the objects were, and then have that available _after_ a GC threw the list away by keeping the stats in the WeakRef.

The new implementation doesn't need to use ObjectCleaner :smile:.

### DIFF
--- a/common/src/main/java/io/netty/util/Recycler.java
+++ b/common/src/main/java/io/netty/util/Recycler.java
@@ -37,21 +37,43 @@ import static java.lang.Math.min;
  * @param <T> the type of the pooled object
  */
 public abstract class Recycler<T> {
+    // By its very nature this class uses heuristics, and depending on the
+    // workload may be overeager or too conservative. However, to prevent
+    // unnecessary burden on users, it has a core design principal of "do no
+    // harm." Being conservative and providing less benefit is better than
+    // being disabled due to overeagerness and providing no benefit.
+    //
+    // Because this class causes objects to have longer lifetimes than
+    // previously, it can negatively impact memory usage. Thus it follows the
+    // rule: "do not be the _direct_ cause of an object to outlive its
+    // generation." It does this by using weak references when objects are not
+    // actively being used. Worst-case, it will retain one WeakReference per
+    // thread of additional memory. This class does _indirectly_ cause objects
+    // to outlive their generation, however. To help mitigate that, it attempts
+    // to follow: "prefer using objects in older generations over objects in
+    // newer generations." This keeps newer objects out of the application,
+    // giving them more opportunity to be garbage collected. This rule works
+    // well for CMS, but is imprecise for G1.
+    //
+    // Because this class needs to maintain state, it can negatively impact GC
+    // and memory usage. Thus it follows the rule: "avoid overhead for objects
+    // that won't be reused." While more sophisticated methods of following the
+    // rule are possible, the class uses a simple approach of making ~1% of the
+    // new objects be recyclable. If recycling is helping, the recyclable
+    // objects will accumulate.
 
     private static final InternalLogger logger = InternalLoggerFactory.getInstance(Recycler.class);
 
-    @SuppressWarnings("rawtypes")
-    private static final Handle NOOP_HANDLE = new Handle() {
+    private static final Handle<?> NOOP_HANDLE = new Handle<Object>() {
         @Override
         public void recycle(Object object) {
             // NOOP
         }
     };
-    private static final AtomicInteger ID_GENERATOR = new AtomicInteger(Integer.MIN_VALUE);
-    private static final int OWN_THREAD_ID = ID_GENERATOR.getAndIncrement();
-    private static final int DEFAULT_INITIAL_MAX_CAPACITY_PER_THREAD = 4 * 1024; // Use 4k instances as default.
+
+    // These constants aren't actually used, but they are needed to maintain the previous API.
+    private static final int DEFAULT_INITIAL_MAX_CAPACITY_PER_THREAD = 4 * 1024;
     private static final int DEFAULT_MAX_CAPACITY_PER_THREAD;
-    private static final int INITIAL_CAPACITY;
     private static final int MAX_SHARED_CAPACITY_FACTOR;
     private static final int MAX_DELAYED_QUEUES_PER_THREAD;
     private static final int LINK_CAPACITY;
@@ -99,32 +121,17 @@ public abstract class Recycler<T> {
                 logger.debug("-Dio.netty.recycler.ratio: {}", RATIO);
             }
         }
-
-        INITIAL_CAPACITY = min(DEFAULT_MAX_CAPACITY_PER_THREAD, 256);
     }
 
     private final int maxCapacityPerThread;
-    private final int maxSharedCapacityFactor;
-    private final int ratioMask;
-    private final int maxDelayedQueuesPerThread;
 
-    private final FastThreadLocal<Stack<T>> threadLocal = new FastThreadLocal<Stack<T>>() {
+    final FastThreadLocal<Stack<T>> threadLocal = new FastThreadLocal<Stack<T>>() {
         @Override
         protected Stack<T> initialValue() {
-            return new Stack<T>(Recycler.this, Thread.currentThread(), maxCapacityPerThread, maxSharedCapacityFactor,
-                    ratioMask, maxDelayedQueuesPerThread);
-        }
-
-        @Override
-        protected void onRemoval(Stack<T> value) {
-            // Let us remove the WeakOrderQueue from the WeakHashMap directly if its safe to remove some overhead
-            if (value.threadRef.get() == Thread.currentThread()) {
-               if (DELAYED_RECYCLED.isSet()) {
-                   DELAYED_RECYCLED.get().remove(value);
-               }
-            }
+            return new Stack<T>(Recycler.this);
         }
     };
+    final AtomicInteger currentGenId = new AtomicInteger();
 
     protected Recycler() {
         this(DEFAULT_MAX_CAPACITY_PER_THREAD);
@@ -140,15 +147,10 @@ public abstract class Recycler<T> {
 
     protected Recycler(int maxCapacityPerThread, int maxSharedCapacityFactor,
                        int ratio, int maxDelayedQueuesPerThread) {
-        ratioMask = safeFindNextPositivePowerOfTwo(ratio) - 1;
         if (maxCapacityPerThread <= 0) {
             this.maxCapacityPerThread = 0;
-            this.maxSharedCapacityFactor = 1;
-            this.maxDelayedQueuesPerThread = 0;
         } else {
             this.maxCapacityPerThread = maxCapacityPerThread;
-            this.maxSharedCapacityFactor = max(1, maxSharedCapacityFactor);
-            this.maxDelayedQueuesPerThread = max(0, maxDelayedQueuesPerThread);
         }
     }
 
@@ -158,12 +160,7 @@ public abstract class Recycler<T> {
             return newObject((Handle<T>) NOOP_HANDLE);
         }
         Stack<T> stack = threadLocal.get();
-        DefaultHandle<T> handle = stack.pop();
-        if (handle == null) {
-            handle = stack.newHandle();
-            handle.value = newObject(handle);
-        }
-        return (T) handle.value;
+        return stack.popOrCreate();
     }
 
     /**
@@ -176,20 +173,12 @@ public abstract class Recycler<T> {
         }
 
         DefaultHandle<T> h = (DefaultHandle<T>) handle;
-        if (h.stack.parent != this) {
+        if (h.parent != this) {
             return false;
         }
 
         h.recycle(o);
         return true;
-    }
-
-    final int threadLocalCapacity() {
-        return threadLocal.get().elements.length;
-    }
-
-    final int threadLocalSize() {
-        return threadLocal.get().size;
     }
 
     protected abstract T newObject(Handle<T> handle);
@@ -199,16 +188,15 @@ public abstract class Recycler<T> {
     }
 
     static final class DefaultHandle<T> implements Handle<T> {
-        private int lastRecycledId;
-        private int recycleId;
+        private final WeakReference<DefaultHandle<T>> selfRef = new WeakReference<DefaultHandle<T>>(this);
+        private final Recycler<T> parent;
+        private T value;
+        private WeakReference<DefaultHandle<T>> next;
+        final int genId;
 
-        boolean hasBeenRecycled;
-
-        private Stack<?> stack;
-        private Object value;
-
-        DefaultHandle(Stack<?> stack) {
-            this.stack = stack;
+        DefaultHandle(Recycler<T> parent) {
+            this.parent = parent;
+            this.genId = parent.currentGenId.get();
         }
 
         @Override
@@ -216,446 +204,89 @@ public abstract class Recycler<T> {
             if (object != value) {
                 throw new IllegalArgumentException("object does not belong to handle");
             }
-
-            Stack<?> stack = this.stack;
-            if (lastRecycledId != recycleId || stack == null) {
-                throw new IllegalStateException("recycled already");
-            }
-
-            stack.push(this);
-        }
-    }
-
-    private static final FastThreadLocal<Map<Stack<?>, WeakOrderQueue>> DELAYED_RECYCLED =
-            new FastThreadLocal<Map<Stack<?>, WeakOrderQueue>>() {
-        @Override
-        protected Map<Stack<?>, WeakOrderQueue> initialValue() {
-            return new WeakHashMap<Stack<?>, WeakOrderQueue>();
-        }
-    };
-
-    // a queue that makes only moderate guarantees about visibility: items are seen in the correct order,
-    // but we aren't absolutely guaranteed to ever see anything at all, thereby keeping the queue cheap to maintain
-    private static final class WeakOrderQueue {
-
-        static final WeakOrderQueue DUMMY = new WeakOrderQueue();
-
-        // Let Link extend AtomicInteger for intrinsics. The Link itself will be used as writerIndex.
-        @SuppressWarnings("serial")
-        static final class Link extends AtomicInteger {
-            private final DefaultHandle<?>[] elements = new DefaultHandle[LINK_CAPACITY];
-
-            private int readIndex;
-            Link next;
-        }
-
-        // This act as a place holder for the head Link but also will reclaim space once finalized.
-        // Its important this does not hold any reference to either Stack or WeakOrderQueue.
-        static final class Head {
-            private final AtomicInteger availableSharedCapacity;
-
-            Link link;
-
-            Head(AtomicInteger availableSharedCapacity) {
-                this.availableSharedCapacity = availableSharedCapacity;
-            }
-
-            /// TODO: In the future when we move to Java9+ we should use java.lang.ref.Cleaner.
-            @Override
-            protected void finalize() throws Throwable {
-                try {
-                    super.finalize();
-                } finally {
-                    Link head = link;
-                    link = null;
-                    while (head != null) {
-                        reclaimSpace(LINK_CAPACITY);
-                        Link next = head.next;
-                        // Unlink to help GC and guard against GC nepotism.
-                        head.next = null;
-                        head = next;
-                    }
-                }
-            }
-
-            void reclaimSpace(int space) {
-                assert space >= 0;
-                availableSharedCapacity.addAndGet(space);
-            }
-
-            boolean reserveSpace(int space) {
-                return reserveSpace(availableSharedCapacity, space);
-            }
-
-            static boolean reserveSpace(AtomicInteger availableSharedCapacity, int space) {
-                assert space >= 0;
-                for (;;) {
-                    int available = availableSharedCapacity.get();
-                    if (available < space) {
-                        return false;
-                    }
-                    if (availableSharedCapacity.compareAndSet(available, available - space)) {
-                        return true;
-                    }
-                }
-            }
-        }
-
-        // chain of data items
-        private final Head head;
-        private Link tail;
-        // pointer to another queue of delayed items for the same stack
-        private WeakOrderQueue next;
-        private final WeakReference<Thread> owner;
-        private final int id = ID_GENERATOR.getAndIncrement();
-
-        private WeakOrderQueue() {
-            owner = null;
-            head = new Head(null);
-        }
-
-        private WeakOrderQueue(Stack<?> stack, Thread thread) {
-            tail = new Link();
-
-            // Its important that we not store the Stack itself in the WeakOrderQueue as the Stack also is used in
-            // the WeakHashMap as key. So just store the enclosed AtomicInteger which should allow to have the
-            // Stack itself GCed.
-            head = new Head(stack.availableSharedCapacity);
-            head.link = tail;
-            owner = new WeakReference<Thread>(thread);
-        }
-
-        static WeakOrderQueue newQueue(Stack<?> stack, Thread thread) {
-            final WeakOrderQueue queue = new WeakOrderQueue(stack, thread);
-            // Done outside of the constructor to ensure WeakOrderQueue.this does not escape the constructor and so
-            // may be accessed while its still constructed.
-            stack.setHead(queue);
-
-            return queue;
-        }
-
-        private void setNext(WeakOrderQueue next) {
-            assert next != this;
-            this.next = next;
-        }
-
-        /**
-         * Allocate a new {@link WeakOrderQueue} or return {@code null} if not possible.
-         */
-        static WeakOrderQueue allocate(Stack<?> stack, Thread thread) {
-            // We allocated a Link so reserve the space
-            return Head.reserveSpace(stack.availableSharedCapacity, LINK_CAPACITY)
-                    ? newQueue(stack, thread) : null;
-        }
-
-        void add(DefaultHandle<?> handle) {
-            handle.lastRecycledId = id;
-
-            Link tail = this.tail;
-            int writeIndex;
-            if ((writeIndex = tail.get()) == LINK_CAPACITY) {
-                if (!head.reserveSpace(LINK_CAPACITY)) {
-                    // Drop it.
-                    return;
-                }
-                // We allocate a Link so reserve the space
-                this.tail = tail = tail.next = new Link();
-
-                writeIndex = tail.get();
-            }
-            tail.elements[writeIndex] = handle;
-            handle.stack = null;
-            // we lazy set to ensure that setting stack to null appears before we unnull it in the owning thread;
-            // this also means we guarantee visibility of an element in the queue if we see the index updated
-            tail.lazySet(writeIndex + 1);
-        }
-
-        boolean hasFinalData() {
-            return tail.readIndex != tail.get();
-        }
-
-        // transfer as many items as we can from this queue to the stack, returning true if any were transferred
-        @SuppressWarnings("rawtypes")
-        boolean transfer(Stack<?> dst) {
-            Link head = this.head.link;
-            if (head == null) {
-                return false;
-            }
-
-            if (head.readIndex == LINK_CAPACITY) {
-                if (head.next == null) {
-                    return false;
-                }
-                this.head.link = head = head.next;
-            }
-
-            final int srcStart = head.readIndex;
-            int srcEnd = head.get();
-            final int srcSize = srcEnd - srcStart;
-            if (srcSize == 0) {
-                return false;
-            }
-
-            final int dstSize = dst.size;
-            final int expectedCapacity = dstSize + srcSize;
-
-            if (expectedCapacity > dst.elements.length) {
-                final int actualCapacity = dst.increaseCapacity(expectedCapacity);
-                srcEnd = min(srcStart + actualCapacity - dstSize, srcEnd);
-            }
-
-            if (srcStart != srcEnd) {
-                final DefaultHandle[] srcElems = head.elements;
-                final DefaultHandle[] dstElems = dst.elements;
-                int newDstSize = dstSize;
-                for (int i = srcStart; i < srcEnd; i++) {
-                    DefaultHandle element = srcElems[i];
-                    if (element.recycleId == 0) {
-                        element.recycleId = element.lastRecycledId;
-                    } else if (element.recycleId != element.lastRecycledId) {
-                        throw new IllegalStateException("recycled already");
-                    }
-                    srcElems[i] = null;
-
-                    if (dst.dropHandle(element)) {
-                        // Drop the object.
-                        continue;
-                    }
-                    element.stack = dst;
-                    dstElems[newDstSize ++] = element;
-                }
-
-                if (srcEnd == LINK_CAPACITY && head.next != null) {
-                    // Add capacity back as the Link is GCed.
-                    this.head.reclaimSpace(LINK_CAPACITY);
-                    this.head.link = head.next;
-                }
-
-                head.readIndex = srcEnd;
-                if (dst.size == newDstSize) {
-                    return false;
-                }
-                dst.size = newDstSize;
-                return true;
-            } else {
-                // The destination stack is full already.
-                return false;
-            }
+            parent.threadLocal.get().push(this);
         }
     }
 
     static final class Stack<T> {
+        private static final WeakReference<?> NULL_REF = new WeakReference<DefaultHandle<?>>(null);
 
-        // we keep a queue of per-thread queues, which is appended to once only, each time a new thread other
-        // than the stack owner recycles: when we run out of items in our stack we iterate this collection
-        // to scavenge those that can be reused. this permits us to incur minimal thread synchronisation whilst
-        // still recycling all items.
         final Recycler<T> parent;
+        /** The maxId of every handle within the generation. The last generation accepts the remaining ids. */
+        private final int[] genMaxId = new int[] {-1, -1};
+        private final WeakReference[] genHead = new WeakReference[] {
+            NULL_REF, NULL_REF, NULL_REF
+        };
+        private int allocationCount;
 
-        // We store the Thread in a WeakReference as otherwise we may be the only ones that still hold a strong
-        // Reference to the Thread itself after it died because DefaultHandle will hold a reference to the Stack.
-        //
-        // The biggest issue is if we do not use a WeakReference the Thread may not be able to be collected at all if
-        // the user will store a reference to the DefaultHandle somewhere and never clear this reference (or not clear
-        // it in a timely manner).
-        final WeakReference<Thread> threadRef;
-        final AtomicInteger availableSharedCapacity;
-        final int maxDelayedQueues;
-
-        private final int maxCapacity;
-        private final int ratioMask;
-        private DefaultHandle<?>[] elements;
-        private int size;
-        private int handleRecycleCount = -1; // Start with -1 so the first one will be recycled.
-        private WeakOrderQueue cursor, prev;
-        private volatile WeakOrderQueue head;
-
-        Stack(Recycler<T> parent, Thread thread, int maxCapacity, int maxSharedCapacityFactor,
-              int ratioMask, int maxDelayedQueues) {
+        Stack(Recycler<T> parent) {
             this.parent = parent;
-            threadRef = new WeakReference<Thread>(thread);
-            this.maxCapacity = maxCapacity;
-            availableSharedCapacity = new AtomicInteger(max(maxCapacity / maxSharedCapacityFactor, LINK_CAPACITY));
-            elements = new DefaultHandle[min(INITIAL_CAPACITY, maxCapacity)];
-            this.ratioMask = ratioMask;
-            this.maxDelayedQueues = maxDelayedQueues;
         }
 
-        // Marked as synchronized to ensure this is serialized.
-        synchronized void setHead(WeakOrderQueue queue) {
-            queue.setNext(head);
-            head = queue;
-        }
-
-        int increaseCapacity(int expectedCapacity) {
-            int newCapacity = elements.length;
-            int maxCapacity = this.maxCapacity;
-            do {
-                newCapacity <<= 1;
-            } while (newCapacity < expectedCapacity && newCapacity < maxCapacity);
-
-            newCapacity = min(newCapacity, maxCapacity);
-            if (newCapacity != elements.length) {
-                elements = Arrays.copyOf(elements, newCapacity);
-            }
-
-            return newCapacity;
-        }
-
-        @SuppressWarnings({ "unchecked", "rawtypes" })
         DefaultHandle<T> pop() {
-            int size = this.size;
-            if (size == 0) {
-                if (!scavenge()) {
-                    return null;
+            for (int i = 0; i < genHead.length; i++) {
+                WeakReference<DefaultHandle<T>> head = genHead[i];
+                if (head == NULL_REF) {
+                    continue;
                 }
-                size = this.size;
+                DefaultHandle<T> node = head.get();
+                if (node == null) {
+                    gcDetected(i);
+                    genHead[i] = NULL_REF;
+                    continue;
+                }
+                genHead[i] = node.next;
+                node.next = null;
+                return node;
             }
-            size --;
-            DefaultHandle ret = elements[size];
-            elements[size] = null;
-            if (ret.lastRecycledId != ret.recycleId) {
-                throw new IllegalStateException("recycled multiple times");
-            }
-            ret.recycleId = 0;
-            ret.lastRecycledId = 0;
-            this.size = size;
-            return ret;
+            return null;
         }
 
-        boolean scavenge() {
-            // continue an existing scavenge, if any
-            if (scavengeSome()) {
-                return true;
-            }
+        void gcDetected(int genIndex) {
+            int i = genIndex;
+            if (i > 0) {
+                // Promote this generation
 
-            // reset our scavenge cursor
-            prev = null;
-            cursor = head;
-            return false;
-        }
-
-        boolean scavengeSome() {
-            WeakOrderQueue prev;
-            WeakOrderQueue cursor = this.cursor;
-            if (cursor == null) {
-                prev = null;
-                cursor = head;
-                if (cursor == null) {
-                    return false;
-                }
-            } else {
-                prev = this.prev;
-            }
-
-            boolean success = false;
-            do {
-                if (cursor.transfer(this)) {
-                    success = true;
-                    break;
-                }
-                WeakOrderQueue next = cursor.next;
-                if (cursor.owner.get() == null) {
-                    // If the thread associated with the queue is gone, unlink it, after
-                    // performing a volatile read to confirm there is no data left to collect.
-                    // We never unlink the first queue, as we don't want to synchronize on updating the head.
-                    if (cursor.hasFinalData()) {
-                        for (;;) {
-                            if (cursor.transfer(this)) {
-                                success = true;
-                            } else {
-                                break;
-                            }
-                        }
-                    }
-
-                    if (prev != null) {
-                        prev.setNext(next);
-                    }
+                int maxId;
+                if (i < genMaxId.length) {
+                    maxId = genMaxId[i];
                 } else {
-                    prev = cursor;
+                    // Eden generation
+
+                    maxId = parent.currentGenId.get();
+                    if (maxId == genMaxId[i - 1]) {
+                        // This is the first thread to detect the new generation
+                        parent.currentGenId.incrementAndGet();
+                    }
                 }
-
-                cursor = next;
-
-            } while (cursor != null && !success);
-
-            this.prev = prev;
-            this.cursor = cursor;
-            return success;
-        }
-
-        void push(DefaultHandle<?> item) {
-            Thread currentThread = Thread.currentThread();
-            if (threadRef.get() == currentThread) {
-                // The current Thread is the thread that belongs to the Stack, we can try to push the object now.
-                pushNow(item);
-            } else {
-                // The current Thread is not the one that belongs to the Stack
-                // (or the Thread that belonged to the Stack was collected already), we need to signal that the push
-                // happens later.
-                pushLater(item, currentThread);
+                genMaxId[i - 1] = maxId;
             }
         }
 
-        private void pushNow(DefaultHandle<?> item) {
-            if ((item.recycleId | item.lastRecycledId) != 0) {
+        void push(DefaultHandle<T> node) {
+            if (node.next != null) {
                 throw new IllegalStateException("recycled already");
             }
-            item.recycleId = item.lastRecycledId = OWN_THREAD_ID;
-
-            int size = this.size;
-            if (size >= maxCapacity || dropHandle(item)) {
-                // Hit the maximum capacity or should drop - drop the possibly youngest object.
-                return;
+            int i = 0;
+            for (; i < genMaxId.length && genMaxId[i] < node.genId; i++) {
+                // find generation index
             }
-            if (size == elements.length) {
-                elements = Arrays.copyOf(elements, min(size << 1, maxCapacity));
-            }
-
-            elements[size] = item;
-            this.size = size + 1;
+            node.next = genHead[i];
+            genHead[i] = node.selfRef;
         }
 
-        private void pushLater(DefaultHandle<?> item, Thread thread) {
-            // we don't want to have a ref to the queue as the value in our weak map
-            // so we null it out; to ensure there are no races with restoring it later
-            // we impose a memory ordering here (no-op on x86)
-            Map<Stack<?>, WeakOrderQueue> delayedRecycled = DELAYED_RECYCLED.get();
-            WeakOrderQueue queue = delayedRecycled.get(this);
-            if (queue == null) {
-                if (delayedRecycled.size() >= maxDelayedQueues) {
-                    // Add a dummy queue so we know we should drop the object
-                    delayedRecycled.put(this, WeakOrderQueue.DUMMY);
-                    return;
-                }
-                // Check if we already reached the maximum number of delayed queues and if we can allocate at all.
-                if ((queue = WeakOrderQueue.allocate(this, thread)) == null) {
-                    // drop object
-                    return;
-                }
-                delayedRecycled.put(this, queue);
-            } else if (queue == WeakOrderQueue.DUMMY) {
-                // drop object
-                return;
+        T popOrCreate() {
+            DefaultHandle<T> handle = pop();
+            if (handle != null) {
+                return handle.value;
             }
-
-            queue.add(item);
-        }
-
-        boolean dropHandle(DefaultHandle<?> handle) {
-            if (!handle.hasBeenRecycled) {
-                if ((++handleRecycleCount & ratioMask) != 0) {
-                    // Drop the object.
-                    return true;
-                }
-                handle.hasBeenRecycled = true;
+            if ((allocationCount++ & 63) == 0) {
+                // Create recyclable objects at a low rate. If they are helpful, they will accumulate
+                handle = new DefaultHandle<T>(parent);
+                handle.value = parent.newObject(handle);
+                return handle.value;
             }
-            return false;
-        }
-
-        DefaultHandle<T> newHandle() {
-            return new DefaultHandle<T>(this);
+            return parent.newObject((Handle<T>) NOOP_HANDLE);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


ejona86:
Note that producerConsumer is mainly there for the gc.alloc.rate.norm. The runtimes are coarse, but get the point across; I ran with -Dforks=1 -DwarmupIterations=5 -DmeasureIterations=3

```
before:
RecyclerBenchmark.plainNew                                               avgt    3     9.088 ±    2.504   ns/op
RecyclerBenchmark.plainNew:·gc.alloc.rate                                avgt    3  5597.032 ± 1510.210  MB/sec
RecyclerBenchmark.plainNew:·gc.alloc.rate.norm                           avgt    3    80.000 ±    0.001    B/op
RecyclerBenchmark.plainNew:·gc.churn.PS_Eden_Space                       avgt    3  5601.251 ±  158.963  MB/sec
RecyclerBenchmark.plainNew:·gc.churn.PS_Eden_Space.norm                  avgt    3    80.071 ±   20.797    B/op
RecyclerBenchmark.plainNew:·gc.churn.PS_Survivor_Space                   avgt    3     0.076 ±    0.218  MB/sec
RecyclerBenchmark.plainNew:·gc.churn.PS_Survivor_Space.norm              avgt    3     0.001 ±    0.003    B/op
RecyclerBenchmark.plainNew:·gc.count                                     avgt    3    99.000             counts
RecyclerBenchmark.plainNew:·gc.time                                      avgt    3    68.000                 ms
RecyclerBenchmark.producerConsumer                                       avgt    3   316.584 ±   77.286   ns/op
RecyclerBenchmark.producerConsumer:consumer                              avgt    3   316.583 ±   77.303   ns/op
RecyclerBenchmark.producerConsumer:producer                              avgt    3   316.584 ±   77.271   ns/op
RecyclerBenchmark.producerConsumer:·gc.alloc.rate                        avgt    3   264.830 ±  125.890  MB/sec
RecyclerBenchmark.producerConsumer:·gc.alloc.rate.norm                   avgt    3    65.978 ±   16.503    B/op
RecyclerBenchmark.producerConsumer:·gc.churn.PS_Eden_Space               avgt    3   273.072 ± 1725.475  MB/sec
RecyclerBenchmark.producerConsumer:·gc.churn.PS_Eden_Space.norm          avgt    3    67.844 ±  417.193    B/op
RecyclerBenchmark.producerConsumer:·gc.churn.PS_Survivor_Space           avgt    3     0.021 ±    0.380  MB/sec
RecyclerBenchmark.producerConsumer:·gc.churn.PS_Survivor_Space.norm      avgt    3     0.005 ±    0.094    B/op
RecyclerBenchmark.producerConsumer:·gc.count                             avgt    3     5.000             counts
RecyclerBenchmark.producerConsumer:·gc.time                              avgt    3     5.000                 ms
RecyclerBenchmark.recyclerGetAndOrphan                                   avgt    3    15.122 ±    1.601   ns/op
RecyclerBenchmark.recyclerGetAndOrphan:·gc.alloc.rate                    avgt    3  4704.758 ±  520.253  MB/sec
RecyclerBenchmark.recyclerGetAndOrphan:·gc.alloc.rate.norm               avgt    3   112.000 ±    0.001    B/op
RecyclerBenchmark.recyclerGetAndOrphan:·gc.churn.PS_Eden_Space           avgt    3  4754.013 ±   16.992  MB/sec
RecyclerBenchmark.recyclerGetAndOrphan:·gc.churn.PS_Eden_Space.norm      avgt    3   113.175 ±   12.780    B/op
RecyclerBenchmark.recyclerGetAndOrphan:·gc.churn.PS_Survivor_Space       avgt    3     0.173 ±    1.160  MB/sec
RecyclerBenchmark.recyclerGetAndOrphan:·gc.churn.PS_Survivor_Space.norm  avgt    3     0.004 ±    0.027    B/op
RecyclerBenchmark.recyclerGetAndOrphan:·gc.count                         avgt    3    84.000             counts
RecyclerBenchmark.recyclerGetAndOrphan:·gc.time                          avgt    3    55.000                 ms
RecyclerBenchmark.recyclerGetAndRecycle                                  avgt    3    14.302 ±    3.369   ns/op
RecyclerBenchmark.recyclerGetAndRecycle:·gc.alloc.rate                   avgt    3    ≈ 10⁻⁴             MB/sec
RecyclerBenchmark.recyclerGetAndRecycle:·gc.alloc.rate.norm              avgt    3    ≈ 10⁻⁵               B/op
RecyclerBenchmark.recyclerGetAndRecycle:·gc.count                        avgt    3       ≈ 0             counts

after (no 1% sampling):
RecyclerBenchmark.plainNew                                               avgt    3     8.983 ±    0.167   ns/op
RecyclerBenchmark.plainNew:·gc.alloc.rate                                avgt    3  5654.780 ±  154.695  MB/sec
RecyclerBenchmark.plainNew:·gc.alloc.rate.norm                           avgt    3    80.000 ±    0.001    B/op
RecyclerBenchmark.plainNew:·gc.churn.PS_Eden_Space                       avgt    3  5658.427 ± 1857.463  MB/sec
RecyclerBenchmark.plainNew:·gc.churn.PS_Eden_Space.norm                  avgt    3    80.050 ±   24.149    B/op
RecyclerBenchmark.plainNew:·gc.churn.PS_Survivor_Space                   avgt    3     0.062 ±    0.001  MB/sec
RecyclerBenchmark.plainNew:·gc.churn.PS_Survivor_Space.norm              avgt    3     0.001 ±    0.001    B/op
RecyclerBenchmark.plainNew:·gc.count                                     avgt    3   100.000             counts
RecyclerBenchmark.plainNew:·gc.time                                      avgt    3    63.000                 ms
RecyclerBenchmark.producerConsumer                                       avgt    3   264.363 ±  952.286   ns/op
RecyclerBenchmark.producerConsumer:consumer                              avgt    3   264.365 ±  952.373   ns/op
RecyclerBenchmark.producerConsumer:producer                              avgt    3   264.360 ±  952.199   ns/op
RecyclerBenchmark.producerConsumer:·gc.alloc.rate                        avgt    3   388.263 ± 1531.383  MB/sec
RecyclerBenchmark.producerConsumer:·gc.alloc.rate.norm                   avgt    3    78.489 ±    9.565    B/op
RecyclerBenchmark.producerConsumer:·gc.churn.PS_Eden_Space               avgt    3   383.914 ± 1746.368  MB/sec
RecyclerBenchmark.producerConsumer:·gc.churn.PS_Eden_Space.norm          avgt    3    77.384 ±   90.381    B/op
RecyclerBenchmark.producerConsumer:·gc.churn.PS_Survivor_Space           avgt    3     0.021 ±    0.380  MB/sec
RecyclerBenchmark.producerConsumer:·gc.churn.PS_Survivor_Space.norm      avgt    3     0.004 ±    0.084    B/op
RecyclerBenchmark.producerConsumer:·gc.count                             avgt    3     7.000             counts
RecyclerBenchmark.producerConsumer:·gc.time                              avgt    3     4.000                 ms
RecyclerBenchmark.recyclerGetAndOrphan                                   avgt    3    23.962 ±    3.783   ns/op
RecyclerBenchmark.recyclerGetAndOrphan:·gc.alloc.rate                    avgt    3  3816.896 ±  562.904  MB/sec
RecyclerBenchmark.recyclerGetAndOrphan:·gc.alloc.rate.norm               avgt    3   144.000 ±    0.001    B/op
RecyclerBenchmark.recyclerGetAndOrphan:·gc.churn.PS_Eden_Space           avgt    3  3847.989 ± 1789.199  MB/sec
RecyclerBenchmark.recyclerGetAndOrphan:·gc.churn.PS_Eden_Space.norm      avgt    3   145.180 ±   71.643    B/op
RecyclerBenchmark.recyclerGetAndOrphan:·gc.churn.PS_Survivor_Space       avgt    3     0.250 ±    1.317  MB/sec
RecyclerBenchmark.recyclerGetAndOrphan:·gc.churn.PS_Survivor_Space.norm  avgt    3     0.009 ±    0.051    B/op
RecyclerBenchmark.recyclerGetAndOrphan:·gc.count                         avgt    3    68.000             counts
RecyclerBenchmark.recyclerGetAndOrphan:·gc.time                          avgt    3    48.000                 ms
RecyclerBenchmark.recyclerGetAndRecycle                                  avgt    3    18.690 ±    0.170   ns/op
RecyclerBenchmark.recyclerGetAndRecycle:·gc.alloc.rate                   avgt    3    ≈ 10⁻⁴             MB/sec
RecyclerBenchmark.recyclerGetAndRecycle:·gc.alloc.rate.norm              avgt    3    ≈ 10⁻⁵               B/op
RecyclerBenchmark.recyclerGetAndRecycle:·gc.count                        avgt    3       ≈ 0             counts


# Note the runtime difference in recyclerGetAndRecycle and lack of change to recyclerGetAndOrphan
after (no 1% sampling, non-generational):
RecyclerBenchmark.plainNew                                               avgt    3     9.029 ±    1.148   ns/op
RecyclerBenchmark.plainNew:·gc.alloc.rate                                avgt    3  5629.562 ±  710.186  MB/sec
RecyclerBenchmark.plainNew:·gc.alloc.rate.norm                           avgt    3    80.000 ±    0.001    B/op
RecyclerBenchmark.plainNew:·gc.churn.PS_Eden_Space                       avgt    3  5605.200 ± 3104.981  MB/sec
RecyclerBenchmark.plainNew:·gc.churn.PS_Eden_Space.norm                  avgt    3    79.647 ±   35.767    B/op
RecyclerBenchmark.plainNew:·gc.churn.PS_Survivor_Space                   avgt    3     0.076 ±    0.580  MB/sec
RecyclerBenchmark.plainNew:·gc.churn.PS_Survivor_Space.norm              avgt    3     0.001 ±    0.008    B/op
RecyclerBenchmark.plainNew:·gc.count                                     avgt    3    99.000             counts
RecyclerBenchmark.plainNew:·gc.time                                      avgt    3    63.000                 ms
RecyclerBenchmark.producerConsumer                                       avgt    3   237.773 ± 1211.562   ns/op
RecyclerBenchmark.producerConsumer:consumer                              avgt    3   237.771 ± 1211.542   ns/op
RecyclerBenchmark.producerConsumer:producer                              avgt    3   237.775 ± 1211.582   ns/op
RecyclerBenchmark.producerConsumer:·gc.alloc.rate                        avgt    3   466.422 ± 2551.443  MB/sec
RecyclerBenchmark.producerConsumer:·gc.alloc.rate.norm                   avgt    3    82.588 ±   24.188    B/op
RecyclerBenchmark.producerConsumer:·gc.churn.PS_Eden_Space               avgt    3   439.561 ± 1746.286  MB/sec
RecyclerBenchmark.producerConsumer:·gc.churn.PS_Eden_Space.norm          avgt    3    81.939 ±  546.164    B/op
RecyclerBenchmark.producerConsumer:·gc.churn.PS_Survivor_Space           avgt    3     0.014 ±    0.219  MB/sec
RecyclerBenchmark.producerConsumer:·gc.churn.PS_Survivor_Space.norm      avgt    3     0.003 ±    0.045    B/op
RecyclerBenchmark.producerConsumer:·gc.count                             avgt    3     8.000             counts
RecyclerBenchmark.producerConsumer:·gc.time                              avgt    3     8.000                 ms
RecyclerBenchmark.recyclerGetAndOrphan                                   avgt    3    23.661 ±    4.677   ns/op
RecyclerBenchmark.recyclerGetAndOrphan:·gc.alloc.rate                    avgt    3  3864.795 ±  812.263  MB/sec
RecyclerBenchmark.recyclerGetAndOrphan:·gc.alloc.rate.norm               avgt    3   144.000 ±    0.001    B/op
RecyclerBenchmark.recyclerGetAndOrphan:·gc.churn.PS_Eden_Space           avgt    3  3903.223 ±   47.398  MB/sec
RecyclerBenchmark.recyclerGetAndOrphan:·gc.churn.PS_Eden_Space.norm      avgt    3   145.445 ±   30.515    B/op
RecyclerBenchmark.recyclerGetAndOrphan:·gc.churn.PS_Survivor_Space       avgt    3     0.277 ±    1.796  MB/sec
RecyclerBenchmark.recyclerGetAndOrphan:·gc.churn.PS_Survivor_Space.norm  avgt    3     0.010 ±    0.066    B/op
RecyclerBenchmark.recyclerGetAndOrphan:·gc.count                         avgt    3    69.000             counts
RecyclerBenchmark.recyclerGetAndOrphan:·gc.time                          avgt    3    43.000                 ms
RecyclerBenchmark.recyclerGetAndRecycle                                  avgt    3    13.730 ±    0.977   ns/op
RecyclerBenchmark.recyclerGetAndRecycle:·gc.alloc.rate                   avgt    3    ≈ 10⁻⁴             MB/sec
RecyclerBenchmark.recyclerGetAndRecycle:·gc.alloc.rate.norm              avgt    3    ≈ 10⁻⁵               B/op
RecyclerBenchmark.recyclerGetAndRecycle:·gc.count                        avgt    3       ≈ 0             counts


after (1% sampling; this is the code you actually see):
RecyclerBenchmark.plainNew                                               avgt    3     9.053 ±    2.028   ns/op
RecyclerBenchmark.plainNew:·gc.alloc.rate                                avgt    3  5616.314 ± 1301.369  MB/sec
RecyclerBenchmark.plainNew:·gc.alloc.rate.norm                           avgt    3    80.000 ±    0.001    B/op
RecyclerBenchmark.plainNew:·gc.churn.PS_Eden_Space                       avgt    3  5601.811 ± 3141.269  MB/sec
RecyclerBenchmark.plainNew:·gc.churn.PS_Eden_Space.norm                  avgt    3    79.781 ±   26.312    B/op
RecyclerBenchmark.plainNew:·gc.churn.PS_Survivor_Space                   avgt    3     0.090 ±    0.791  MB/sec
RecyclerBenchmark.plainNew:·gc.churn.PS_Survivor_Space.norm              avgt    3     0.001 ±    0.011    B/op
RecyclerBenchmark.plainNew:·gc.count                                     avgt    3    99.000             counts
RecyclerBenchmark.plainNew:·gc.time                                      avgt    3    67.000                 ms
RecyclerBenchmark.producerConsumer                                       avgt    3   194.900 ±  351.183   ns/op
RecyclerBenchmark.producerConsumer:consumer                              avgt    3   194.902 ±  351.094   ns/op
RecyclerBenchmark.producerConsumer:producer                              avgt    3   194.897 ±  351.273   ns/op
RecyclerBenchmark.producerConsumer:·gc.alloc.rate                        avgt    3   296.171 ±  552.715  MB/sec
RecyclerBenchmark.producerConsumer:·gc.alloc.rate.norm                   avgt    3    45.155 ±    4.967    B/op
RecyclerBenchmark.producerConsumer:·gc.churn.PS_Eden_Space               avgt    3   273.668 ± 1741.189  MB/sec
RecyclerBenchmark.producerConsumer:·gc.churn.PS_Eden_Space.norm          avgt    3    41.460 ±  244.824    B/op
RecyclerBenchmark.producerConsumer:·gc.churn.PS_Survivor_Space           avgt    3     0.035 ±    0.580  MB/sec
RecyclerBenchmark.producerConsumer:·gc.churn.PS_Survivor_Space.norm      avgt    3     0.005 ±    0.092    B/op
RecyclerBenchmark.producerConsumer:·gc.count                             avgt    3     5.000             counts
RecyclerBenchmark.producerConsumer:·gc.time                              avgt    3     3.000                 ms
RecyclerBenchmark.recyclerGetAndOrphan                                   avgt    3    12.599 ±    6.015   ns/op
RecyclerBenchmark.recyclerGetAndOrphan:·gc.alloc.rate                    avgt    3  4083.809 ± 1873.270  MB/sec
RecyclerBenchmark.recyclerGetAndOrphan:·gc.alloc.rate.norm               avgt    3    81.000 ±    0.001    B/op
RecyclerBenchmark.recyclerGetAndOrphan:·gc.churn.PS_Eden_Space           avgt    3  4072.825 ± 3085.454  MB/sec
RecyclerBenchmark.recyclerGetAndOrphan:·gc.churn.PS_Eden_Space.norm      avgt    3    80.764 ±   30.359    B/op
RecyclerBenchmark.recyclerGetAndOrphan:·gc.churn.PS_Survivor_Space       avgt    3     0.062 ±    0.378  MB/sec
RecyclerBenchmark.recyclerGetAndOrphan:·gc.churn.PS_Survivor_Space.norm  avgt    3     0.001 ±    0.007    B/op
RecyclerBenchmark.recyclerGetAndOrphan:·gc.count                         avgt    3    72.000             counts
RecyclerBenchmark.recyclerGetAndOrphan:·gc.time                          avgt    3    50.000                 ms
RecyclerBenchmark.recyclerGetAndRecycle                                  avgt    3    19.069 ±    4.756   ns/op
RecyclerBenchmark.recyclerGetAndRecycle:·gc.alloc.rate                   avgt    3    ≈ 10⁻⁴             MB/sec
RecyclerBenchmark.recyclerGetAndRecycle:·gc.alloc.rate.norm              avgt    3    ≈ 10⁻⁵               B/op
RecyclerBenchmark.recyclerGetAndRecycle:·gc.count                        avgt    3       ≈ 0             counts
```

common/src/main/java/io/netty/util/Recycler.java
```diff
@@ -38,21 +38,43 @@
...
         @Override
         public void recycle(Object object) {
             // NOOP
         }
     };
-    private static final AtomicInteger ID_GENERATOR = new AtomicInteger(Integer.MIN_VALUE);
-    private static final int OWN_THREAD_ID = ID_GENERATOR.getAndIncrement();
-    private static final int DEFAULT_INITIAL_MAX_CAPACITY_PER_THREAD = 4 * 1024; // Use 4k instances as default.
+
+    // These constants aren't actually used, but they are needed to maintain the previous API.
```
normanmaurer: Add TODO: to remove in next major release.

common/src/main/java/io/netty/util/Recycler.java
```diff
@@ -177,478 +174,120 @@ public final boolean recycle(T o, Handle<T> handle) {
...
-        return threadLocal.get().size;
-    }
-
     protected abstract T newObject(Handle<T> handle);
 
     public interface Handle<T> {
         void recycle(T object);
     }
 
     static final class DefaultHandle<T> implements Handle<T> {
```
normanmaurer: should we just extend `WeakReference` here to reduce object creation ?

common/src/main/java/io/netty/util/Recycler.java
```diff
@@ -177,478 +174,120 @@ public final boolean recycle(T o, Handle<T> handle) {
...
-        private int lastRecycledId;
-        private int recycleId;
-
-        boolean hasBeenRecycled;
+        private final WeakReference<DefaultHandle<T>> selfRef = new WeakReference<DefaultHandle<T>>(this);
```
normanmaurer: see above.

common/src/main/java/io/netty/util/Recycler.java
```diff
@@ -177,478 +174,120 @@ public final boolean recycle(T o, Handle<T> handle) {
...
-        private DefaultHandle<?>[] elements;
-        private int size;
-        private int handleRecycleCount = -1; // Start with -1 so the first one will be recycled.
-        private WeakOrderQueue cursor, prev;
-        private volatile WeakOrderQueue head;
-
-        Stack(Recycler<T> parent, Thread thread, int maxCapacity, int maxSharedCapacityFactor,
-              int ratioMask, int maxDelayedQueues) {
+        /** The maxId of every handle within the generation. The last generation accepts the remaining ids. */
+        private int[] genMaxId = new int[] {-1, -1};
```
normanmaurer: final ?

common/src/main/java/io/netty/util/Recycler.java
```diff
@@ -177,478 +174,120 @@ public final boolean recycle(T o, Handle<T> handle) {
...
+        private WeakReference[] genHead = new WeakReference[] {
```
normanmaurer: final

common/src/main/java/io/netty/util/Recycler.java
```diff
@@ -177,478 +174,120 @@ public final boolean recycle(T o, Handle<T> handle) {
...
-                handle.hasBeenRecycled = true;
+            node.next = genHead[i];
+            genHead[i] = node.selfRef;
+        }
+
+        T popOrCreate() {
+            DefaultHandle<T> handle = pop();
+            if (handle != null) {
+                return handle.value;
+            } else if ((allocationCount++ & 63) == 0) {
```
normanmaurer: nit: you can remove the else as we return anyway in the if.

common/src/main/java/io/netty/util/Recycler.java
```diff
@@ -177,478 +174,120 @@ public final boolean recycle(T o, Handle<T> handle) {
...
+                // Create recyclable objects at a low rate. If they are helpful, they will accumulate
+                handle = new DefaultHandle<T>(parent);
+                handle.value = parent.newObject(handle);
+                return handle.value;
+            } else {
```
normanmaurer: you can remove the else.

normanmaurer:
@ejona86 I like the idea... one thing I am not sure about is how this will work in the scenario where we allocate in one thread and recycle in another thread.  In the old implementation it was possible while still get a good "hit count". How do we account for this here ?

normanmaurer:
@mosesn @bryce-anderson @jasontedor @Scottmitch @vkostyukov @carl-mastrangelo I would like to get your thoughts on this as well. 

@vkostyukov as you mentioned you disabled the recycler in finagle in https://github.com/netty/netty/pull/5968 would it be possible to test this one as well ?

@ejona86 the issue you were searching for is https://github.com/netty/netty/pull/5968.

common/src/main/java/io/netty/util/Recycler.java
ejona86: The problem is that WeakReference is a fixed reference at construction time and we need WeakReference in the linked list. So the linked list alternates between the fixed WeakReference and the mutable DefaultHandle.

ejona86:
> one thing I am not sure about is how this will work in the scenario where we allocate in one thread and recycle in another thread

This handles that case by the recycling thread receiving "ownership" of the recycled instance (no pushLater). If that receiving thread needs to allocate in the future, it can use the recycled instance. So if coordinating between Netty threads, it should work well.

If the recycling thread won't ever need instances (it's not a producer as well) or there's an imbalance, then no (or limited) reuse will be performed. This is when the 1% sampling kicks in and predominantly uses non-recyclable instances, effectively as if the Recycler wasn't being used. Note that this is opposite to dropHandle() which drops objects when recycling; they are instead _created_ with the inability to be recycled.

It would be possible to have a larger shared pool to put things in when a particular generation's list grows too long, or something. But it would need synchronization and Java's GC seems better suited to that case.

I didn't entirely follow how `pushLater` worked in the previous implementation. That's partly because I spent time confused until I realized `DELAYED_RECYCLED` is a static ThreadLocal whereas `threadLocal` was non-static. gc.alloc.rate.norm before for producerConsumer is 66 B/op. I believe that's halved, because it counts an "op" from both producer and consumer, so it's 132 B/op. So either the benchmark is broken or the previous implementation wouldn't reuse if a consumer is not a producer.

bryce-anderson:
Based on the forewarning by @ejona86 I haven't read the implementation in depth so I can't comment on it one way or the other. In general this sounds like an improvement since it attempts to prevent unnecessary generation migration, but it's hard to say how much. In my limited experience, microbenchmarks haven't done a great job representing how changes that can affect GC behavior will affect real applications.

If we think that recyclers are here to stay then let me know and I'll sink the time into a thorough review.

Fwiw, we just disabled recyclers (https://github.com/twitter/finagle/commit/b8523a4541953d6bbc6a64b9409d75ff162afcd9) as @vkostyukov noted on issue #5968. Maybe he's interested in running some benchmarks of the patch and seeing how they do.

vkostyukov:
As Bryce mentioned, we decided to disable Recyclers in Finagle as we observed they come at a non-trivial CPU cost. Disabling them increased RPS/core score in our internal Finagle simulations (with a small penalty in allocations/req) - see [this comment for more details](https://github.com/netty/netty/pull/5968#issuecomment-399578989).

I'd be happy to try these changes on our internal setup. Will update this thread when I know more.

ejona86:
@vkostyukov, thanks for offering to try out this change. It'll be useful to know if you used CMS or G1.

normanmaurer:
@ejona86 yeah thats what I was talking about... I think it can "frequently" happen that allocations happen from a non EventLoop thread that will never recycle by itself. For example if people write `ByteBuf` from outside the `EventLoop`.

ejona86:
> I think it can "frequently" happen that allocations happen from a non EventLoop thread that will never recycle by itself.

I agree it can happen and isn't necessarily rare, although I do question if it happens frequently enough to make a difference in performance, especially since the code would need some slower paths to handle it. Did the previous Recycler implementation optimize that case? The benchmark makes it seem like not, but the code is not quite clear without me looking deeper.

normanmaurer:
@ejona86 could you incorporate the requested changes and @vkostyukov could you do some testings with this ?

common/src/main/java/io/netty/util/Recycler.java
ejona86: Done

common/src/main/java/io/netty/util/Recycler.java
ejona86: Done

common/src/main/java/io/netty/util/Recycler.java
ejona86: Changed.

common/src/main/java/io/netty/util/Recycler.java
ejona86: Done. While I generally prefer avoiding the unnecessary else, I'm not sure it is as clear in this case.

ejona86:
I rebased past #8064, but the resulting file should be the same. I did not squash.

normanmaurer:
@ejona86 @bryce-anderson @vkostyukov @mosesn how can me move this forward ? Any takers ?

ejona86:
Rebased past 338ef969318 (although I threw away the change to Recycler, because my new implementation didn't have the bug) and squashed commits.

normanmaurer:
@hepin1989 Is Alibaba maybe interested in testing this ?

normanmaurer:
@netty-bot test this please

normanmaurer:
@netty-bot test this please

normanmaurer:
@vkostyukov @bryce-anderson did you ever test this ?

vkostyukov:
@normanmaurer I tried it in the past but haven't observed anything material: some noise that we normally see in our load-tests. I haven't dig deeper though so wasn't really sure how useful my testing was.

normanmaurer:
@belliottsmith would you mind have a look as well ?

belliottsmith:
Sorry for the extremely slow response!

So, it took me a little bit to grok this patch, and I'm still not totally sure of its intent - I wonder if you could explain at a high level what you expect this algorithm to do, and why you think this would work well?

As it stands, it looks like it will lose its contents on every GC (or, at least, most or part of its contents, and potentially all on a major GC), so only those items that are in use across a GC boundary will be available for allocation (or ‘promotion’)?  If so, that seems a much weaker semantic of recycler, and will never reach a stable garbage-free point?

It’s possible you intended to use SoftReference?  This leads to its own semantic problems, but we can have a better discussion once we know what you intended as the algorithm, and what you anticipate its behaviour to be?

There's certainly some elegance to the idea of a self-tuning pool, although it is never truly self-tuning.  Here we’re externalising our tuning onto GC parameters, which are already notoriously hard to deal with, and must handle many competing pressures.

Right now, my understanding of Recycler’s main cause of pain is cross-thread recycling.  This is because we tried to absolutely minimise CPU cost without bounding memory cost, and I can see a lot of minor mitigations have been done to improve the situation, without properly tackling it.  

This problem is most pronounced when there are a LOT (perhaps hundreds) of threads recycling to each other, and explains the MAX_DELAYED_QUEUES_PER_THREAD option.  In this case, it's fairly likely there will be distinct producer and consumer threads, and as mentioned upthread, this patch doesn't address this problem. In effect it disables the recycler in this case, but with some extra GC burden as we have to wait a couple of GC cycles to discard the unusable objects.

There are a lot of things that can be done here, that are orthogonal to the main thrust of this patch and should probably be addressed independently IMO.  Some that spring to mind:

* Our WeakOrderQueue recycles 16 items at once (configurably); there’s no reason to actually prevent the allocating thread from seeing a partially complete batch of 16, though, so that we can recycle objects immediately
* Per-thread we can track the maximum number of allocated buffers we had in-flight at any one time; if we currently have fewer than this in our buffer, we could steal an object we would have recycled to another thread.  If the thread never allocates, we will cross-thread recycle as normal, but we will save any cross-thread recycling otherwise.
* We could automatically switch to a shared MpscQueue that is not weakly ordered if there are too many WeakOrderQueue
* We can and perhaps should impose system-wide recycler limits, either by memory or number, and have each Recycler and WeakOrderQueue take permits as necessary

FWIW, the benchmark as created is probably too limited to be of use IMO, as it only ever has one object in flight at a time.  In real use there may be hundreds or thousands of allocated objects that are in use and yet to be recycled.  I’ve pushed a suggestion for a general approach that would at least somewhat account for this behaviour here: https://github.com/belliottsmith/netty/tree/recycler-benchmark, though I haven't played with it extensively

We should also be benchmarking cross-thread recycling, in both the mixed producer/consumer and the isolated producer/consumer cases.

ejona86:
> So, it took me a little bit to grok this patch, and I'm still not totally sure of its intent - I wonder if you could explain at a high level what you expect this algorithm to do, and why you think this would work well?

Hmmm... I felt like the documentation I added at the very top of Recycler did that. Based on the other comments it seems it may be a case of "yes, it really does that."

> As it stands, it looks like it will lose its contents on every GC (or, at least, most or part of its contents, and potentially all on a major GC), so only those items that are in use across a GC boundary will be available for allocation (or ‘promotion’)? If so, that seems a much weaker semantic of recycler, and will never reach a stable garbage-free point?

Yes, it loses contents every GC. The problem with the previous Recycler implementation is that it could retain objects too long and didn't have a trigger to stop saving objects. This new one takes the approach that if it reuses an object 100 times between two GC's, then that's a win, even if the object is then GCed.

> MAX_DELAYED_QUEUES_PER_THREAD option.<snip>In effect it disables the recycler in this case, but with some extra GC burden as we have to wait a couple of GC cycles to discard the unusable objects.

Why does it need to wait a couple of GC cycles? All objects that can be cleaned up will be cleaned up, excepting a single WeakReference per thread.

Also, the sampling approach taken means that even if the recycler is not helping, it is likely to cause harm. Recyclable objects will only accumulate if they are being reused.

> It's probably best not to conflate these distinct problems in a single patch, given the complexities of each.

I wasn't "trying to fix Recycler" per se. I was addressing the concerns that 1) Recycler causes harm in some workloads, 2) it's hard to tell which work loads it helps and which it hurts, 3) manual tuning helps it, so the problems with (1) and (2) may just be poor tuning, and 4) we were seriously considering deleting Recycler completely.

In the face of that, I realized I could greatly reduce the cost/risk of Recycler by being okay with objects being GCed. Everything else was just gravy to further reduce cost/risk.

> Some that spring to mind:

But how do we validate those changes, and determine if they resolve issues with pathological environments without creating new issues? The problem is the Recycler performance can't be truly tested without a real heap and real application behavior. That means we need to run a wide variety of real application benchmarks with/without Recycler and see what happens. That's been obviously too high a cost to validate a plethera of optimizations and for iteration.

This new recycler on the other hand targets very low cost in general, and then chips away at easy gains that help in certain scenarios while being unlikely to harm other scenarios.

> FWIW, the benchmark as created is probably too limited to be of use IMO, as it only ever has one object in flight at a time.

That was not the point of the benchmark. The benchmark was to show the pure overhead of the approach. It was targeting the worst-case. In showing the worst-case performance loss is small, if we show a single use-case where it helps (via real-world benchmarks) then at that point it is already "safe" and "helpful."

Also, some of the benchmarks were just to increase confidence that the code was behaving as expected.

> In real use there may be hundreds or thousands of allocated objects that are in use and yet to be recycled.

Meh. Number of allocated objects doesn't really matter to this implementation, nor did it seem to matter with the previous implementation.

The point of the Recycler is not to reuse objects. There's no "performance gain" from that directly. The point of the Recycler is to reduce the frequency of GCs. And the cost of GCs depends directly on the actual heap used and application allocation behavior. So anything short of a real application benchmark provides little signal for whether the Recycler is helping.

belliottsmith:
> Meh.  Number of allocated objects doesn't really matter to this implementation, nor did it seem to matter with the previous implementation.

Surely it impacts the GC overhead, since the number of items in the recycler at the point we GC determines how many objects are collected?

> That was not the point of the benchmark. The benchmark was to show the pure overhead of the approach. It was targeting the worst-case. In showing the worst-case performance loss is small, if we show a single use-case where it helps (via real-world benchmarks) then at that point it is already "safe" and "helpful."

I think it ignored the GC burden in normal usage, unless I'm missing something?  Which, as you mention, is the main purpose of recycling.

> I was addressing the concerns that 1) Recycler causes harm in some workloads, 2) it's hard to tell which work loads it helps and which it hurts, 3) manual tuning helps it, so the problems with (1) and (2) may just be poor tuning, and 4) we were seriously considering deleting Recycler completely.
> The problem with the previous Recycler implementation is that it could retain objects too long and didn't have a trigger to stop saving objects.

I think it would be great to pool all of the data we have on this, and look at that before taking any drastic action.  Is there any real data on Recycler behaviour, and which scenarios it has caused problems?  

I've quoted one well understood fundamental flaw in its design, that can be remedied.  Do we know that any of the generalised malaise quoted is definitely not attributable to this flaw?

There's also an entirely separate question of what objects are recycled; the logic of recycling is very different for small objects, large objects, and objects of variable size, and each use-site should probably encode its own modifiers to recycler behaviour (e.g. small objects probably make no sense to cross-thread recycle, unless the threads both allocate and free).

> Everything else was just gravy to further reduce cost/risk.

What's your definition of cost/risk?  If we don't currently understand the behaviour today, changing multiple dimensions of behaviour leaves us with little extra knowledge, and a drastic change in multiple dimensions of behaviour seem (to me) to have a fairly high cost and risk for the community at large, unless the benefits are really pronounced.  Increasing the GC burden for applications is a non-trivial thing for them to debug on upgrade, and something many users would consider high cost.

ejona86:
> > Meh.  Number of allocated objects doesn't really matter to this implementation, nor did it seem to matter with the previous implementation.
>
> Surely it impacts the GC overhead, since the number of items in the recycler at the point we GC determines how many objects are collected?

Two things: 1) _retained_ objects mainly contribute to GC overhead, not collected objects. 2) the number of objects in the recycler should be within the rounding error of memory usage in an actual application (if it's not, that's a problem). The application's own retained objects will be contributing the majority GC cost per collection.

> I think it ignored the GC burden in normal usage, unless I'm missing something? Which, as you mention, is the main purpose of recycling.

Yes, it ignored directly measuring the GC burden, because that can only be measured in a real application. The GC burden comes from everything-that-is-not-the-recycler. It was mainly focused on CPU usage since that is where the worst-case performance loss should come from. It did measure allocation rate, which will contribute to GC burden, to get an idea of how it behaves.

GC burden = per-GC cost x frequency of GCs. The Recycler can only provide benefit by reducing the frequency of GCs.

> I think it would be great to pool all of the data we have on this, and look at that before taking any drastic action.

Recycler's in a no-man's land, but the only expected action to take before this PR was to delete it. The main discussion around the Recycler has been "delete it" because it keeps causing performance degradation and lots of people have disabled it to good effect. But deleting it is also risky because it could be contributing well in some cases. But there's only a history of decisions; not something that we can test changes against to iterate on the implementation. So it's sort of stuck in a "can't move forward; can't move backward" state, unless we expend a lot more energy on it.

This change tries to side-step the problem by having a zero-tuning system that works well when it works and does almost nothing when it doesn't.

> There's also an entirely separate question of what objects are recycled; the logic of recycling is very different for small objects, large objects, and objects of variable size, and each use-site should probably encode its own modifiers to recycler behaviour (e.g. small objects probably make no sense to cross-thread recycle, unless the threads both allocate and free).

I completely agree. This is one of the reasons I'd be in favor of trashing the Recycler, because the things being Recycled have mainly just been guesses as to what would be "helpful." And the current Recycler can hurt as much as it helps.

But with this proposed recycler, the risk of getting the trade-off wrong is much lower. I'd still be happy if we were more precise in when we choose to use the Recycler, but it's not something we're going to get 100% right as it is work-load dependent and this change decreases the chances of the Recycler being harmful.

Now I agree there may be cases where we should cross-thread recycle, although that probably is only "byte buffers" and we already do that. Today, Recycler is only usable for _cheap_ objects, because otherwise it will cache far too many. So this new Recycler is no worse off.

> What's your definition of cost/risk? If we don't currently understand the behaviour today,

We know the sources of the _negative_ effects today: memory usage and CPU use. Let's say you only do cross-thread operations. Then the Recycler is pure overhead. And it can contribute overhead via over-caching (memory) and managing the cache (CPU).

> changing multiple dimensions of behaviour leaves us with little extra knowledge

It avoids us needing to hand-optimize various use-cases as it is self-tuning and has low overhead when unhelpful. It also handles well the cases Recycler handles well. The Recycler is pretty old and we're really no better off with having a good benchmark suite than when it first began. I understand the value of trying to reach a local maxima, but I see no evidence we are approaching that right now with Recycler.

The only knowledge as an ecosystem I think we have today is "try turning off Recycler and see if it helps."

> Increasing the GC burden for applications is a non-trivial thing for them to debug on upgrade, and something many users would consider high cost.

In what use-case do you think this new implementation would increase GC burden compared to the previous implementation? I agree there is risk that we may overlook something, but I thought it was clear I had tried to cover the bases.

The only major risk I identified is we could cause objects to be promoted to older generations only to soon thereafter orphan them (because the weak reference linked list was broken). I pre-emptively added handling for that case that seems work work well for CMS. I agree it wouldn't work as well for G1. But it still seems better than the current Recycler.

We _know_ that the current Recycler increases CPU/GC burden for many applications, as many applications have disabled the Recycler completely. I feel like the bar is very low.

belliottsmith:
> Two things: 1) _retained_ objects mainly contribute to GC overhead, not collected objects

This is an empirical claim, and I'm unsure of the basis for it.  If anything, with long lived easily reused objects being tenured, and their non-allocation reducing the frequency of eden/survivor collections, there will as a result be fewer promotions (since there is more time for objects to become unused before GC), and ultimately fewer times even tracing these long lived objects.

> We know the sources of the _negative_ effects today: memory usage and CPU use

This isn't a source, it's the negative effects themselves?

> Let's say you only do cross-thread operations. Then the Recycler is pure overhead.

If this is an absolute certainty, we should simply disable cross-thread recycling right now.

> In what use-case do you think this new implementation would increase GC burden compared to the previous implementation?

In the simple case of the current recycler, once we get over the ratioMask (which I'm unsure of the point of), we never generate garbage.  This patch clearly does generate garbage for the entire application lifetime.  What more are you looking for?

> We _know_ that the current Recycler increases CPU/GC burden for many applications, as many applications have disabled the Recycler completely. I feel like the bar is very low.

There's a clear potential for selection bias here, which is that the design - when used as intended - never encounters these pathological cases.  So those users reaping the benefits without incident are obviously silent.

It's not clear to me why the assumption is that the entire design is broken, when it's a fairly simple and universal paradigm that's employed safely in other applications:

* We know there's a specific pathological case.  There's a trivial fix for that (turning off cross-thread recycling, or implementing a weak or soft reference approach like this patch exclusively for cross-thread theft-recycling), or a slightly more sophisticated one (actually bound the memory we permit in flight, or reduce the recycle latency in these cases).  
* We haven't ruled out the simplest possibility, which is that all of those people complaining are encountering this known pathological case.

Anyway, this feels like a difficult dispute to resolve without more to go on.  I've given my three cents (50% more than intended!), which I hope helps you guys reach a conclusion.

belliottsmith:
So, before completely signing off, I figure I should say what I think a reasonable course of action would be...

1. There seems to me clearly some value in maintaining the same-thread recycling behaviour we have today, even if it is simplified dramatically and has much stricter constraints put upon it (e.g. limiting the size of each thread's stack more than it is today)  
2. If we think cross-thread recycling is really bad (which it probably is, at least with applications with many threads), using a modified version of this patch's approach, where a thread recycling an object from another thread first puts it onto a queue whose head is a WeakReference, so that it's always guaranteed to be in Eden and collected swiftly (of which only one is maintained per recycler).  There would be no WeakReference allocated for a handle, so there would be at most one WeakReference per thread X recycler at any one time, and we would anticipate very few unused objects making it to tenured, though many may make it to Survivor II.

This would be a more modest change to the current design, probably retaining much of the benefit, and limiting any of the possible downsides that have been proposed or inferred.

As far as limits are concerned, given recyclers are thread-local, there could quite easily be global hard-limits imposed per-thread (across _all_ recyclers) or soft-limits imposed across the process (also across all recyclers), with each Recycler defining a Weigher for its items (using either a constant or simple arithmetic for dynamically sized objects).  This would guarantee bounded memory utilisation, and the CPU burden would remain low, as it would be a simple stack push/pop, with only two extremely simple single-threaded queues to consult and maintain.

ejona86:
@belliottsmith, part of the point of this PR is that it requires no tuning. At the end of the day I want to stop thinking about the Recycler and optimizing it for my (users') workloads. That was not possible previously. This PR sheds some hope for that reality, and it's half the LOC to boot!

While I value pre-existing code (which I agree is frequently undervalued), I do find suggestions of patching the Recycler in ways A, B, and C to start looking like a sunk cost fallacy. I agree the Recycler could be improved, but I think it is close to a local maximum.

> This is an empirical claim, and I'm unsure of the basis for it.

In a very large heap, minor GC costs easily come from the mark phase as the GC needs to scan the large heap in order to find references to the very small eden. This cost increases the more tenured objects mutate, as pages containing those objects have to be scanned to see if they reference an object in eden (see card marking). The sweep is relatively low cost since worst-case it copies all of eden but eden is small.

> If anything, with long lived easily reused objects being tenured, and their non-allocation reducing the frequency of eden/survivor collections, there will as a result be fewer promotions (since there is more time for objects to become unused before GC), and ultimately fewer times even tracing these long lived objects.

I agree reducing the GC rate is beneficial, and there can be multiple ways that comes to pass. But that's all beside the point. My point was that a benchmark needs to have a _real heap_ with _real behavior_ modifying that heap, since the GC needs to traverse that heap to determine what is trash. If the heap isn't filled with GBs of objects (real objects, not some synthetic construction), then you aren't actually seeing a realistic per-GC cost.

>> Let's say you only do cross-thread operations. Then the Recycler is pure overhead.
>
> If this is an absolute certainty, we should simply disable cross-thread recycling right now.

But you don't know a priori whether an object will be used for local or cross-thread recycling. The current Recycler has a high per-object allocation cost in terms of relative GC churn. `DefaultHandle` is 56 bytes in size. Consider `PendingWrite` which is using Recycler and is 48 bytes; the recycler allocation is larger than the object to be recycled.

> There's a trivial fix for that (turning off cross-thread recycling, or implementing a weak or soft reference approach like this patch exclusively for cross-thread theft-recycling), or a slightly more sophisticated one (actually bound the memory we permit in flight, or reduce the recycle latency in these cases).

All of this "it's trivial" seems to be ignoring the point that "it's not trivial." Or rather, "if it's so trivial please send a PR because it isn't trivial to me." Simply deciding between those options doesn't seem trivial to me, each having its own tradeoffs.

> where a thread recycling an object from another thread first puts it onto a queue whose head is a WeakReference, so that it's always guaranteed to be in Eden and collected swiftly (of which only one is maintained per recycler). There would be no WeakReference allocated for a handle, so there would be at most one WeakReference per thread X recycler at any one time, and we would anticipate very few unused objects making it to tenured, though many may make it to Survivor II.

Unfortunately it can't work that way because WeakReferences are "immutable" (the reference stored within the WeakReference object can't be changed, other than cleared). To have a WeakReference point at an object, you must allocate that WeakReference.

> As far as limits are concerned, given recyclers are thread-local, there could quite easily be global hard-limits imposed per-thread (across all recyclers) or soft-limits imposed across the process (also across all recyclers)

My problem with that is it is workload dependent. There isn't a "one size fits all" appropriate limit. This is the normal problem with free lists. Some applications may only need 3 objects and some may need over 1000. And many applications have needs that vary over time. We've not looking to limit the Big-O; we do that mostly already, I thought.

NickleHuang:
I think the object pool allows cache invalidation, so it should be resolved as a soft reference, just by making the object stored in the Stack a soft reference

NickleHuang:
@ejona86 hi，Man, what is your final solution? There is too much content above, I suggest using soft references to solve this problem, don't you think？

NickleHuang:
**don't say more,look this code : private DefaultHandle<?>[] elements ; it was used to save object , it's Strong Reference . thougth it's cache , why not use SoftReference instead it ? such as private SoftReference<DefaultHandle<?>>[] elements , we can in case of OOM ERROR**

normanmaurer:
@ejona86 let me try to push this over the finish line ... 

/cc @chrisvest any thoughts on this ?

ejona86:
@normanmaurer, I'd be fine rebasing it or the like. It's more of a question "does anyone want it" and "does the code look fine."

normanmaurer:
@ejona86 i really like the idea and would like to experiment with it

ejona86:
FWIW, G1 is also a lot more common now and there are situations with G1 where the "do no harm" principle doesn't hold. But it still holds promise to be better than the current recycler.

normanmaurer:
Let me take this over and do some experimental integration

common/src/main/java/io/netty/util/Recycler.java
```diff
@@ -176,486 +174,120 @@ public final boolean recycle(T o, Handle<T> handle) {
...
+                    maxId = genMaxId[i];
+                } else {
+                    // Eden generation
 
-                    if (prev != null) {
-                        prev.setNext(next);
+                    maxId = parent.currentGenId.get();
+                    if (maxId == genMaxId[i - 1]) {
+                        // This is the first thread to detect the new generation
+                        parent.currentGenId.incrementAndGet();
```
chrisvest, line 260: The check and the increment are not atomic, so isn't this racy?

common/src/main/java/io/netty/util/Recycler.java
```diff
@@ -176,486 +174,120 @@ public final boolean recycle(T o, Handle<T> handle) {
...
-                elements = Arrays.copyOf(elements, min(size << 1, maxCapacity));
+            int i = 0;
+            for (; i < genMaxId.length && genMaxId[i] < node.genId; i++) {
+                // find generation index
             }
-
-            elements[size] = item;
-            this.size = size + 1;
+            node.next = genHead[i];
+            genHead[i] = node.selfRef;
```
chrisvest, line 276: The array business is not obvious to me. Here we are relying on `genMaxId.length` being one less than `genHead.length`, which feels shaky.

common/src/main/java/io/netty/util/Recycler.java
```diff
@@ -176,486 +174,120 @@ public final boolean recycle(T o, Handle<T> handle) {
...
-
-        boolean dropHandle(DefaultHandle<?> handle) {
-            if (!handle.hasBeenRecycled) {
-                if ((++handleRecycleCount & ratioMask) != 0) {
-                    // Drop the object.
-                    return true;
-                }
-                handle.hasBeenRecycled = true;
+            if ((allocationCount++ & 63) == 0) {
+                // Create recyclable objects at a low rate. If they are helpful, they will accumulate
```
chrisvest, line 285: (Score:5, Insightful)

common/src/main/java/io/netty/util/Recycler.java
```diff
@@ -37,21 +37,44 @@
...
+    // newer generations." This keeps newer objects out of the application,
+    // giving them more opportunity to be garbage collected. This rule works
+    // well for CMS, but is imprecise for G1.
+    //
+    // Because this class needs to maintain state, it can negatively impact GC
+    // and memory usage. Thus it follows the rule: "avoid overhead for objects
+    // that won't be reused." While more sophisticated methods of following the
+    // rule are possible, the class uses a simple approach of making ~1% of the
+    // new objects be recyclable. If recycling is helping, the recyclable
+    // objects will accumulate.
```
chrisvest, line 63: Good comment.

chrisvest (commented):
Had a look at the code. Some parts have me stratching my head, but overall this is a lot simpler that the existing Recycler. Mostly because it no longer attempts to recycle objects across threads. I suspect the current implementation has bugs in that area anyway (e.g. #9640), so I think it's worth to explore the approach given here.

This needs a rebase to fix some conflicts, and more conflicts are in the pipeline with #11037.

common/src/main/java/io/netty/util/Recycler.java
ejona86: It'd be quite fair to use a `compareAndSet()` here to reduce mental burden and worries.

But I don't think the race matters _much_ as few `DefaultHandle()`s would be allocated within the race window. It mainly "wastes" an id. I do agree that any `DefaultHandles()`s allocated during the race window will be handled less than _ideal_, as they might not be treated as eden in some threads.

When writing the code, I thought of `currentGenId` as just a clock. The clock needs to tick frequently enough to separate generations, but if it runs too frequently the code would react reasonably. It was so long ago that I can't be certain, but I think I might have purposefully avoided CAS here to help prevent myself from assuming a nice, orderly progression to avoid overly relying on "clean" interactions between threads and make sure I handled "when things go wrong" cases.

common/src/main/java/io/netty/util/Recycler.java
ejona86: They are parallel arrays and need to have that relationship; they are _very_ closely related. We could call it out more in the documentation, if that would help.

common/src/main/java/io/netty/util/Recycler.java
chrisvest: Ok. If it's fine for them race here, then I think it'd be good to call that out in a comment.

common/src/main/java/io/netty/util/Recycler.java
chrisvest: The `genMaxId` field javadoc comment could be expanded to explain how these two arrays work together. I think that would be useful as this code is a little unusual :)

normanmaurer:
I really would like to get something like this in for netty 5.

NickleHuang:
netty5 think use Linux new feature IO_Uring? java aio isn't really asynchronous



---Original---
From: "Norman ***@***.***&gt;
Date: Wed, Aug 25, 2021 21:31 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [netty/netty] RFC: Reimplement Recycler to reduce impact of pathological cases (#8052)




 
I really would like to get something like this in for netty 5.
 
—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

normanmaurer:
@NickleHuang please don't hi-jack issues... Also see https://github.com/netty/netty-incubator-transport-io_uring

NickleHuang:
cool。I'm looking forward to it



---Original---
From: "Norman ***@***.***&gt;
Date: Wed, Aug 25, 2021 22:00 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [netty/netty] RFC: Reimplement Recycler to reduce impact of pathological cases (#8052)




 
@NickleHuang please don't hi-jack issues... Also see https://github.com/netty/netty-incubator-transport-io_uring
 
—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

chrisvest:
@ejona86 I think we can close this since #11858 was merged?

normanmaurer:
Let me close this for now

PR #8053, opened 2018-06-25
Update to java 11+ea19

normanmaurer:
Motivation:

A new java 11 EA version was released.

Modifications:

Update to java 11+ea19

Result:

Use latest java 11 release.


normanmaurer:
Just a version bump for java 11.

PR #8054, opened 2018-06-25
Get rid of deprecated SslContext methods in handler-proxy tests

kachayev:
Motivation:

`ProxyHandlerTest` package uses deprecated methods `SslContext.newServerContext` and
`SslContext.newClientContext`.

Modifications:

`SslContextBuilder` is used to build a server and client `SslContext`.

Result:

Reduced the usage of deprecated methods in the code.


normanmaurer:
@kachayev <3 Thanks a lot!

PR #8056, opened 2018-06-25
Also remove the id from the DnsQueryContextManager if query fails due…

normanmaurer:
… parent Channel activation error.

Motivation:

Whenever we fail the query we should also remove the id from the DnsQueryContextManager.

Modifications:

Remove the id from the DnsQueryContextManager if we fail the query because the channel failed to become active.

Result:

More correct code.


normanmaurer:
This bug should not have any noticeable effect but I thought we should fix it anyway.

Issue #8041, opened 2018-06-20
Http2MultiplexCodec in client mode can trigger a GOAWAY if it creates a stream after getting a GOAWAY


bryce-anderson:
### Expected behavior

When a client attempts to utilize a new stream channel using Http2MultiplexCodec _after_ receiving a GOAWAY, the client will gracefully fail the stream channel with an exception that signifies the stream wasn't processed, perhaps such a RST_STREAM(REFUSED_STREAM).

### Actual behavior

When a client attempts to create a new stream channel using Http2MultiplexCodec _after_ receiving a GOAWAY, the client will throw it's own PROTOCOL_EXCEPTION that gets sent to the server and the connection gets torn down. It's also debatable as to whether the server should be actively tearing down the connection due to that message since it is still processing client streams.

### Steps to reproduce

Establish a client connection using Http2MultiplexCodec, have the server send a GOAWAY with one outstanding client stream, and attempt to create another client stream. You can observe a client-sent GOAWAY(last processed stream: 0) with a message similar to: 
```
Cannot create stream 5 since this endpoint has received a GOAWAY frame with last stream id 3.
```

### Netty version
4.1.26.Final-SNAPSHOT (as of 6/20/2018)

bryce-anderson:
A simple solution might be to add something like
```java
} else if (connection().goAwayReceived()) {
    ReferenceCountUtil.release(frame);
    promise.setFailure(Http2Exception.streamError(-1, Http2Error.REFUSED_STREAM, "Received GOAWAY before stream was initialized.")
    );
    return;
}
```
[here](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java#L1102) in the write method of the `Http2MultiplexCodec.DefaultHttp2StreamChannel.Http2ChannelUnsafe`.

bryce-anderson:
And I'm happy to work on this once we have a consensus on direction.

Scottmitch:
> PROTOCOL_EXCEPTION that gets sent to the server and the connection gets torn down.

[...]

PR #8057, opened 2018-06-25
Http2MultiplexCodec outbound streams check if a GOAWAY has been issued before initializing

bryce-anderson:
Motivation:

When using the Http2MultiplexCodec for clients, if a stream channel gets
its first HEADERS write after receiving a GOAWAY from the server, it
will attempt to create a stream, which will throw a hard connection
exception, triggering session shutdown. This may tear down streams that
haven't finished and doesn't signal to the application that the stream
wasn't processed.

Modification:

Change the
HttpMultiplexCodec.DefaultHttp2StreamChannel.Http2ChannelUnsafe to check
if a GOAWAY has been received and if so synthesize a
RST_STREAM(REFUSED_STREAM) exception which is used to tear down the
child channel gracefully without sending anything to the server.

Result:

Attempting to use a Http2MultiplexCodec child stream after receiving a
GOAWAY from a server isn't fatal to the session.

Fixes: #8041.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
@@ -1096,6 +1096,17 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
                                         + frame.name()));
                                 return;
                             }
+
+                            if (connection().goAwayReceived()) {
+                                // The session is draining so we fail the write with a synthesized RST_STREAM
+                                // exception that signals that the stream was unprocessed.
+                                ReferenceCountUtil.release(frame);
+                                Http2Exception cause = Http2Exception.streamError(-1, Http2Error.REFUSED_STREAM,
+                                        "Received GOAWAY before stream was initialized.");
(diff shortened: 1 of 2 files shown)


codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -1167,6 +1182,7 @@ private void writeComplete(ChannelFuture future, ChannelPromise promise) {
                             outboundClosed = true;
                         }
                     }
+                    promise.setFailure(error);
```
bryce-anderson: Note that this was moved to this position to be consistent with `firstWriteFailed` and the general behavior of AbstractChannel, namely that we transition the state reflected by `Channel.isOpen` and `Channel.isActive`, before failing the outbound buffer, which would normally contain this promise.

normanmaurer:
I assigned a few more people for review

bryce-anderson:
I'm going to close this in favor of #8067, which is probably better all around. I'll open a separate PR to address the ordering issue concerning write failures and child channel teardown events racing with future listeners.

PR #8059, opened 2018-06-26
Expose channel count for FixedChannelPool

rockwotj:
Motivation:

We use FixedChannelPool in production, and we believe we have a leak that doesn't return sockets to the pool (but they should be closed), thus blocking us from creating new connections when we need them. I haven't confirmed this yet, but right now I have to resort to reflection to access this field which makes me sad.

Modification:

Expose the acquiredChannelCount field through a getter method.

Result:

Allows introspection of the pool size in FixedChannelPool.

### DIFF
--- a/transport/src/main/java/io/netty/channel/pool/FixedChannelPool.java
+++ b/transport/src/main/java/io/netty/channel/pool/FixedChannelPool.java
@@ -28,6 +28,7 @@ import io.netty.util.internal.ThrowableUtil;
 import java.nio.channels.ClosedChannelException;
 import java.util.ArrayDeque;
 import java.util.Queue;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
@@ -70,7 +71,7 @@ public class FixedChannelPool extends SimpleChannelPool {
     private final Queue<AcquireTask> pendingAcquireQueue = new ArrayDeque<AcquireTask>();
     private final int maxConnections;
     private final int maxPendingAcquires;
-    private int acquiredChannelCount;
+    private final AtomicInteger acquiredChannelCount = new AtomicInteger();
     private int pendingAcquireCount;
     private boolean closed;
 
@@ -229,6 +230,11 @@ public class FixedChannelPool extends SimpleChannelPool {
         this.maxPendingAcquires = maxPendingAcquires;
     }
 
+    /** Returns the number of acquired channels that this pool thinks it has. */
+    public int acquiredChannelCount() {
+        return acquiredChannelCount.get();
+    }
+
     @Override
     public Future<Channel> acquire(final Promise<Channel> promise) {
         try {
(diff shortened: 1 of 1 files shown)


transport/src/main/java/io/netty/channel/pool/FixedChannelPool.java
```diff
@@ -228,6 +228,11 @@ public void onTimeout(AcquireTask task) {
         this.maxConnections = maxConnections;
         this.maxPendingAcquires = maxPendingAcquires;
     }
+    
+    /** Returns the number of acquired channels that this pool thinks it has. */
+    public int getAcquiredChannelCount() {
+        return acquiredChannelCount;
```
normanmaurer: This is not safe as acquiredChannelCount is not atomic.

transport/src/main/java/io/netty/channel/pool/FixedChannelPool.java
```diff
@@ -228,6 +228,11 @@ public void onTimeout(AcquireTask task) {
         this.maxConnections = maxConnections;
         this.maxPendingAcquires = maxPendingAcquires;
     }
+    
+    /** Returns the number of acquired channels that this pool thinks it has. */
+    public int getAcquiredChannelCount() {
```
normanmaurer: Remove the `get` prefix as we usually not use this prefixes in Netty.

transport/src/main/java/io/netty/channel/pool/FixedChannelPool.java
rockwotj: I see three options, do you have a preference?

1) Make `acquiredChannelCount` atomic
2) Enforce the caller is calling this in the event loop
3) Return a future and call it in the event loop

normanmaurer: I guess making acquiredChannelCount atomic is good enough

rockwotj: Made it atomic.

transport/src/main/java/io/netty/channel/pool/FixedChannelPool.java
```diff
@@ -70,7 +71,7 @@
     private final Queue<AcquireTask> pendingAcquireQueue = new ArrayDeque<AcquireTask>();
     private final int maxConnections;
     private final int maxPendingAcquires;
-    private int acquiredChannelCount;
+    private AtomicInteger acquiredChannelCount = new AtomicInteger();
```
normanmaurer: final

rockwotj: good call, done.

normanmaurer:
@rockwotj can you please squash your commits and sign our ICLA:

https://netty.io/s/icla

Please let me know once done and I will merge this one. Thanks!

wsrspirit:
I think make acquiredChannelCount  atomic can not completely fix this bug,
even though AtomicInteger CAS method is atomic, but the following lines of code are not atomic,
so it can not filly solve this problem 

if (acquiredChannelCount.get() < maxConnections) {
...
acquiredChannelCount.incrementAndGet();
...
}

On the oter side, I hope FixedChannelPool not useing deque(in SimpleChannelPool), which means I can figure out sockets package based on seq num, how can I do?

rockwotj:
This code runs in an single event loop so it doesn't need to be atomic in that situation (note it being a normal `int` before.

rockwotj:
@normanmaurer done.

normanmaurer:
@rockwotj thanks!

@wsrspirit I dont understand your question. If you think there is any issue please open a new issue as this is not related to this change.

PR #8062, opened 2018-06-27
Store NativeDatagramPacketArray directly in the EpollEventLoop

normanmaurer:
Motivation:

We can store the NativeDatagramPacketArray directly in the EpollEventLoop. This removes the need of using FastThreadLocal.

Modifications:

- Store NativeDatagramPacketArray directly in the EpollEventLoop (just as we do with IovArray as well).

Result:

Less FastThreadLocal usage and more consistent code.

PR #8064, opened 2018-06-27
Remove usage of ObjectCleaner

normanmaurer:
Motivation:

ObjectCleaner does start a Thread to handle the cleaning of resources which leaks into the users application. We should not use it in netty itself to make things more predictable.

Modifications:

- Remove usage of ObjectCleaner and use finalize as a replacement when possible.
- Clarify javadocs for FastThreadLocal.onRemoval(...) to ensure its clear that remove() is not guaranteed to be called when the Thread completees and so this method is not enough to guarantee cleanup for this case.

Result:

Fixes https://github.com/netty/netty/issues/8017.

### DIFF
--- a/common/src/main/java/io/netty/util/Recycler.java
+++ b/common/src/main/java/io/netty/util/Recycler.java
@@ -17,7 +17,6 @@
 package io.netty.util;
 
 import io.netty.util.concurrent.FastThreadLocal;
-import io.netty.util.internal.ObjectCleaner;
 import io.netty.util.internal.SystemPropertyUtil;
 import io.netty.util.internal.logging.InternalLogger;
 import io.netty.util.internal.logging.InternalLoggerFactory;
@@ -244,10 +243,9 @@ public abstract class Recycler<T> {
             Link next;
         }
 
-        // This act as a place holder for the head Link but also will be used by the ObjectCleaner
-        // to return space that was before reserved. Its important this does not hold any reference to
-        // either Stack or WeakOrderQueue.
-        static final class Head implements Runnable {
+        // This act as a place holder for the head Link but also will reclaim space once finalized.
+        // Its important this does not hold any reference to either Stack or WeakOrderQueue.
+        static final class Head {
             private final AtomicInteger availableSharedCapacity;
 
             Link link;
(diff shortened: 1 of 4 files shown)


normanmaurer:
@jasontedor @johnou FYI as well... When working on this I also noticed that the docs not really said that `onRemoval(...)` is guaranteed to be called when a `Thread` completes.

normanmaurer:
So with this and https://github.com/netty/netty/pull/8062 there is no more need to use `ObjectCleaner`. The behaviour of `FastThreadLocal` changes basically back to before this commit:

https://github.com/netty/netty/commit/e329ca1cf3cecd41c9cf1a5a7be369a65f002ab5#diff-e0eb4e9a6ea15564e4ddd076c55978de

This was introduced in 4.1.20.Final

normanmaurer:
I have also another idea to still somehow guarantee the call of `onRemoval` without the extra thread. I will do some experiments here and open a PR to show the idea later today.

common/src/main/java/io/netty/util/Recycler.java
```diff
@@ -256,12 +254,17 @@ public void recycle(Object object) {
...
-                    reclaimSpace(LINK_CAPACITY);
-                    head = head.next;
+            protected void finalize() throws Throwable {
+                try {
+                    super.finalize();
+                } finally {
+                    Link head = link;
+                    while (head != null) {
+                        reclaimSpace(LINK_CAPACITY);
+                        head = head.next;
```
johnou: @nitsanw should be free from GC nepotism here right?

normanmaurer: @johnou actually I think you have a good point here. We should ensure we null out next. Let me fix this

normanmaurer: @johnou PTAL again

normanmaurer:
So my other idea was basically to have a `ReferenceQueue` and store some references in it that can handle the onRemoval when needed on a "sampling" kind of basis. While this works ( I have a working PR here) I am not sure this really is any better then just this PR alone.

johnou:
@normanmaurer I would stick with finalize until Java 9 to avoid adding unnecessary complexity unless there is a clear performance gain.

common/src/main/java/io/netty/util/Recycler.java
johnou: @normanmaurer :ok_hand:

common/src/main/java/io/netty/util/Recycler.java
```diff
@@ -256,12 +254,20 @@ public void recycle(Object object) {
...
-            public void run() {
-                Link head = link;
-                while (head != null) {
-                    reclaimSpace(LINK_CAPACITY);
-                    head = head.next;
+            protected void finalize() throws Throwable {
+                try {
+                    super.finalize();
+                } finally {
+                    Link head = link;
```
carl-mastrangelo: optional: destroy `link` after this

normanmaurer: done

PR #8067, opened 2018-06-27
DefaultHttp2Connection checks if a GOAWAY has been issued before assigning a new stream id

bryce-anderson:
Motivation:

If a GOAWAY has been issued by the peer, the endpoint should no longer issue
new stream id's. If it does we fail during stream creation which can
cause connection tear-down.

Modification:

Change the
DefaultHttp2Connection.DefaultEndpoint.incrementAndGetNextStreamId()
method to consider whether a GOAWAY has been received before
issuing a new stream id.

Result:

The Http2FrameCodec will fail writes that attempt to initialize a new
stream with a `Http2NoMoreStreamIdsException`.

Fixes: #8041.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
@@ -708,7 +708,8 @@ public class DefaultHttp2Connection implements Http2Connection {
 
         @Override
         public int incrementAndGetNextStreamId() {
-            return nextReservationStreamId >= 0 ? nextReservationStreamId += 2 : nextReservationStreamId;
+            return goAwayReceived() ? -1 :
+                   nextReservationStreamId >= 0 ? nextReservationStreamId += 2 : nextReservationStreamId;
         }
 
         private void incrementExpectedStreamId(int streamId) {
--- a/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2FrameCodecTest.java
+++ b/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2FrameCodecTest.java
@@ -44,6 +44,7 @@ import io.netty.util.concurrent.DefaultPromise;
 import io.netty.util.concurrent.GlobalEventExecutor;
 import io.netty.util.concurrent.Promise;
 import io.netty.util.internal.ReflectionUtil;
+import org.hamcrest.Matchers;
 import org.junit.After;
 import org.junit.Assume;
 import org.junit.Before;
@@ -58,13 +59,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 
 import static io.netty.handler.codec.http2.Http2CodecUtil.isStreamIdValid;
 import static org.hamcrest.Matchers.instanceOf;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertThat;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
 import static org.mockito.Mockito.any;
 import static org.mockito.Mockito.anyBoolean;
 import static org.mockito.Mockito.anyInt;
@@ -400,6 +395,20 @@ public class Http2FrameCodecTest {
         assertEquals(1, debugData.refCnt());
         assertEquals(State.OPEN, stream.state());
         assertTrue(channel.isActive());
+    }
+
+    @Test
+    public void outboundStreamAfterGoAwayFails() throws Exception {
+        final Http2FrameStream stream = frameCodec.newStream();
+        frameListener.onGoAwayRead(http2HandlerCtx, 0, Http2Error.NO_ERROR.code(), Unpooled.EMPTY_BUFFER);
+
+        assertTrue(frameCodec.connection().goAwayReceived());
+        assertTrue(channel.isOpen());
+        ChannelFuture f = channel.writeAndFlush(
+                new DefaultHttp2HeadersFrame(new DefaultHttp2Headers(), false).stream(stream));
+
+        assertFalse(f.isSuccess());
+        assertThat(f.cause(), instanceOf(Http2NoMoreStreamIdsException.class));
     }
 
     @Test


codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
```diff
@@ -708,7 +708,8 @@ public boolean isPushPromiseSent() {
 
         @Override
         public int incrementAndGetNextStreamId() {
-            return nextReservationStreamId >= 0 ? nextReservationStreamId += 2 : nextReservationStreamId;
+            return goAwayReceived() ? -1 :
```
ejona86: Would it be cleaner to instead set `nextReservationStreamId = -1` (or Integer.MIN_VALUE) in `goAwayReceived()`?

bryce-anderson:
Note that this will replace https://github.com/netty/netty/pull/8057.

As an aside, would it make sense for `Http2NoMoreStreamIdsException` to be a `StreamException(REFUSED_STREAM)`?

ejona86:
> As an aside, would it make sense for Http2NoMoreStreamIdsException to be a StreamException(REFUSED_STREAM)?

In some ways that would be fine, although with REFUSED_STREAM it is fine to re-issued the HTTP request on the same connection. Http2NoMoreStreamIdsException however is "stop using this connection" (like GOAWAY). That's an important difference in semantic, so we'd need to have some way to handle that.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
bryce-anderson: Perhaps. The connection between `nextReservationStreamId` and `nextStreamIdToCreate` is somewhat subtle, but it looks safe to me. I'm happy to change it if you like.

ejona86:
> That's an important difference in semantic, so we'd need to have some way to handle that.

In the Http2MultiplexCodec the cause of the failure may just be a race. If we're only concerned about the race and think the receiver of the exception is going to be aware of the GOAWAY, then REFUSED_STREAM can make sense. That may mean the lower layer uses Http2NoMoreStreamIdsException and Http2MultiplexCodec makes sure things are in a GOAWAY state and then transforms it to REFUSED_STREAM.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
Scottmitch: I wonder if we should instead change `Http2FrameCodec` from this....

```java
final int streamId = connection.local().incrementAndGetNextStreamId();
if (streamId < 0) {
  promise.setFailure(new Http2NoMoreStreamIdsException());
  return;
}
stream.id = streamId;
```

to this...

```java
final Http2Stream oldStream;
try {
  oldStream = connection.local().createStream(connection.local().incrementAndGetNextStreamId(), false);
} catch (Throwable cause) {
  promise.setFailure(cause);
  return;
}
stream.id = oldStream.id();
```

this way the `Http2FrameCodec` doesn't have to make assumptions about `Http2NoMoreStreamIdsException` or create any new failure cases.

bryce-anderson: Right now the stream creation is driven almost entirely in the `Connection[Encoder|Decoder]`s (with the exception being h2c upgrade pathways and a stream error handling pathway for decoding headers that were too large). It makes a fair amount of sense to have it in the Encoder/Decoder due to the requirement of monotonically increasing stream-id's: having stream creation in the Encoder/Decoder helps ensure that we put HEADERS frames on the socket in the right order.

That said, it would probably work. I also don't think it's a good idea to be dolling out stream id's that look valid but aren't, but that's somewhat orthogonal other than combining both changes would result in throwing a [`Http2NoMoreStreamIdsException`](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java#L873) instead of the synthetic RST_STREAM you have in https://github.com/netty/netty/pull/8069. That is arguably a good thing since, as @ejona86 noted, we don't want to trigger retries against the same session.

Scottmitch:
> although with REFUSED_STREAM it is fine to re-issued the HTTP request on the same connection. 

@ejona86 - the peer may interpret `REFUSED_STREAM` as "I can retry on the same connection" however without knowing the criteria that caused the exception has changed there is minimal chance of getting a different result. I think the value of `REFUSED_STREAM` would be to communicate that not processing is done and is is safe to retry in general (on the same connection or not) [1]. `Http2NoMoreStreamIdsException` uses `PROTOCOL_ERROR` as a result of seeing a negative stream ID but carries different semantics for the different use cases:

- outbound side - I have exhausted the streams for this connection and should use another, based on checked for the `Http2NoMoreStreamIdsException` exception locally.
- inbound side - the negative ID is valid so we should send a `PROTOCOL_ERROR`

[1] https://tools.ietf.org/html/rfc7540#section-8.1.4
> The REFUSED_STREAM error code can be included in a RST_STREAM
      frame to indicate that the stream is being closed prior to any
      processing having occurred.  Any request that was sent on the
      reset stream can be safely retried.

bryce-anderson:
Fwiw, the inbound side interacting with `Http2NoMoreStreamIdsException` should be a non-reachable case: the spec ([4.1](https://tools.ietf.org/html/rfc7540#section-4.1)) dictates that the highest order bit is reserved and not part of the stream id representation, which is how it's decoded from the byte stream.[[1](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2FrameReader.java#L192)][[2](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecUtil.java#L199)]

bryce-anderson:
PR #8069 wasn't effective at eliminating the protocol error of sending a RST_STREAM on an idle stream in our usage of `Http2MultiplexCodec`: we're still issuing a `RST_STREAM(CANCEL)` for streams that were discarded due to being initiated after a GOAWAY due to this code [here](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java#L911). It should check that the stream may have been created (`connection().local().mayHaveCreatedStream(id)`), but it would also have sufficed to not allow it to acquire a stream id.

Scottmitch:
@bryce-anderson - I see what you are saying about "dolling out stream id's that look valid but aren't" but I'm leaning toward not handling this in the `incrementAndGetNextStreamId` method for the following reasons:

- folks could presumably use some other means to determine the stream ID (in the lower level API), in which case we will still have to rely upon other mechanisms (additional conditionals on the code path which also have to be handled at lower layers).
- explicitly checking if goaway was received can provide more informative/correct error message (e.g. `Http2NoMoreStreamIdsException` will be deceptive on the outbound path).

Are you opposed to the `connection().local().mayHaveCreatedStream(id)` approach:

```java
// old code
if (parent().isActive() && !streamClosedWithoutError && isStreamIdValid(stream().id())) {
}

// new code
if (parent().isActive() && !streamClosedWithoutError && connection().local(). mayHaveCreatedStream(stream().id())) {
}
```

if we do decide to also handle it in `incrementAndGetNextStreamId`, instead of `goAwayReceived()` (which is focused on the local endpoint) we should likely use `lastStreamKnownByPeer >= 0` to avoid crossing endpoints (similar code was corrected in PR https://github.com/netty/netty/pull/8080).

bryce-anderson:
@Scottmitch, I'm not against checking `mayHaveCreatedStream`, in fact that's exactly what I have in https://github.com/netty/netty/pull/8086. Either this PR or that one should fix my draining issue. Having both seems like a good thing, though arguably it's unnecessary. I think 8086 is probably the better of the two, if we have to choose.

If you want to continue with this PR let me know and I'll fix it. You're right: `goAwayReceived()` is specific to `localEndpoint` but the `incrementAndGetNextStreamId()` method is on both endpoints. Good catch.

Scottmitch:
Thanks for the heads up about PR https://github.com/netty/netty/pull/8086 ... yes I would prefer this approach for now.

bryce-anderson:
@Scottmitch, sorry, to be clear, is _this approach_ referring to this PR or to PR https://github.com/netty/netty/pull/8086? If you mean't 8086, feel free to close this one.

bryce-anderson:
I think we decided this wasn't necessary. We can always reopen it if someone thinks it's a good idea on the future.

PR #8069, opened 2018-06-27
HTTP/2 outbound event after receiving go_away forces sending a go_away

Scottmitch:
Motivation:
If the local endpoint receives a GO_AWAY frame and then tries to write a stream with a streamId higher than the last know stream ID we will throw a connection error. This results in the local peer sending a GO_AWAY frame to the remote peer, but this is not necessary as the error can be isolated to the local endpoint and communicated via the ChannelFuture return value.

Modifications:
- Instead of throwing a connection error, throw a stream error that simulates the peer receiving the stream and replying with a RST

Result:
Connections are not closed abruptly when trying to create a stream on the local endpoint after a GO_AWAY frame is received.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
@@ -864,9 +864,9 @@ public class DefaultHttp2Connection implements Http2Connection {
         private void checkNewStreamAllowed(int streamId, State state) throws Http2Exception {
             assert state != IDLE;
             if (goAwayReceived() && streamId > localEndpoint.lastStreamKnownByPeer()) {
-                throw connectionError(PROTOCOL_ERROR, "Cannot create stream %d since this endpoint has received a " +
-                                                      "GOAWAY frame with last stream id %d.", streamId,
-                                                      localEndpoint.lastStreamKnownByPeer());
+                throw streamError(streamId, REFUSED_STREAM,
+                      "Cannot create stream %d since this endpoint has received a GOAWAY frame with last stream id %d.",
+                      streamId, localEndpoint.lastStreamKnownByPeer());
             }
             if (!isValidStreamId(streamId)) {
                 if (streamId < 0) {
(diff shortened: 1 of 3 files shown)


ejona86:
This is fine assuming the caller observes GOAWAY, as the caller won't retry on this connection. But it seems this case only happens if the caller is ignoring GOAWAY. If the caller is not observing GOAWAY, that should be an "ugly" failure of some sort. Is there a reason to believe the caller just hasn't been notified of the GOAWAY yet at this point?

bryce-anderson:
> Is there a reason to believe the caller just hasn't been notified of the GOAWAY yet at this point?

In the Http2MultiplexCodec case the stream channel doesn't receive GOAWAYs. Even if you checked the connection state (either via checking the connection status or by monitoring the parent channel for the GOAWAY) before creating the channel, the stream isn't assigned until the first HEADERS are written so you'd need to check connection state on the first stream channel write.

Scottmitch:
> If the caller is not observing GOAWAY, that should be an "ugly" failure of some sort.

The assumption with the existing code is the caller verifies the GOAWAY condition manually and/or explicitly does the stream creation (before writing the headers) and catches any exception thrown. In the former case this results in duplicating logic that Netty must also enforce to avoid protocol violations which is not ideal, and closing the connection as a result is not necessary.

> Is there a reason to believe the caller just hasn't been notified of the GOAWAY yet at this point?

The listener is currently invoked before the stat on the connection is updated ... this doesn't seem right because this may result in protocol violations if the caller synchronously creates another stream. However this seems like a separate issue.

bryce-anderson:
I like the changes to `Http2ConnectionHandler.onStreamError` and think that should be committed regardless of the changes to `DefaultHttp2Connection`.

I see this is a very similar situation as running out of stream identifiers as handled by `Http2FrameCodec` [here](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java#L356) and would expect the same exception between the two.

Unfortunately, this plays to @ejona86's concern about the session having an ugly failure mode by just retrying against the same session only to fail again. We can't change the `DefaultHttp2Connection` to throw a `Http2NoMoreStreamIdsException` since it's not a `StreamException`, but changing it to a `StreamException` might trigger the ugly failure mode.

fwiw, I think this would fix things for my use case.

Scottmitch:
discussed with @ejona86 offline briefly ... I will merge now an address any concerns in a followup PR ... however I have a followup PR which builds upon this to resolve the issue described in https://github.com/netty/netty/pull/8069#issuecomment-400770431

PR #8075, opened 2018-06-28
Replace toArray(new T[size]) with toArray(new T[0]) to eliminate zero…

normanmaurer:
…-out and allow the VM to optimize.

Motivation:

Using toArray(new T[0]) is usually the faster aproach these days. We should use it.

See also https://shipilev.net/blog/2016/arrays-wisdom-ancients/#_conclusion.

Modifications:

Replace toArray(new T[size]) with toArray(new T[0]).

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
@@ -391,7 +391,7 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
         }
 
         Collection<ByteBuf> col = (Collection<ByteBuf>) buffers;
-        return addComponents0(increaseIndex, cIndex, col.toArray(new ByteBuf[col.size()]), 0 , col.size());
+        return addComponents0(increaseIndex, cIndex, col.toArray(new ByteBuf[0]), 0 , col.size());
     }
 
     /**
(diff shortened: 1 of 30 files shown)


Bennett-Lynch:
Would shared zero-length array constants benefit at all here?

E.g.,
https://commons.apache.org/proper/commons-lang/apidocs/src-html/org/apache/commons/lang3/ArrayUtils.html#line.87

normanmaurer:
@Bennett-Lynch I suspect no... Let me ask Shipley.

njhill:
@Bennett-Lynch @normanmaurer it's addressed in the same article: https://shipilev.net/blog/2016/arrays-wisdom-ancients/#_caching_the_array

But I'm not sure that includes the deferred cost to GC (which could still be insignificant in grand scheme)?

normanmaurer:
@Bennett-Lynch did not notice it ... doh 👍

hc-codersatlas:
@njhill escape analysis probably takes care of that in this case.

Scottmitch:
Did we observe any performance gains as a result of this PR?

normanmaurer:
I did no do any specific benchmarking for this but just trusted shiplev

doom369:
@Scottmitch I tested this a lot on java 8, 9, 10 many days ago. In all cases `new T[0]` is faster. Also, some static analyzers and IDE inspections now mark `toArray(new T[size])` code as slow.

PR #8076, opened 2018-06-28
Lazy distribute attemptedBytes

carryxyh:
Lazy distribute attemptedBytes which is never be used if channel write failed and return a negative number.

### DIFF
--- a/transport/src/main/java/io/netty/channel/socket/nio/NioSocketChannel.java
+++ b/transport/src/main/java/io/netty/channel/socket/nio/NioSocketChannel.java
@@ -399,12 +399,12 @@ public class NioSocketChannel extends AbstractNioByteChannel implements io.netty
                     // Zero length buffers are not added to nioBuffers by ChannelOutboundBuffer, so there is no need
                     // to check if the total size of all the buffers is non-zero.
                     ByteBuffer buffer = nioBuffers[0];
-                    int attemptedBytes = buffer.remaining();
                     final int localWrittenBytes = ch.write(buffer);
                     if (localWrittenBytes <= 0) {
                         incompleteWrite(true);
                         return;
                     }
+                    int attemptedBytes = buffer.remaining();
                     adjustMaxBytesPerGatheringWrite(attemptedBytes, localWrittenBytes, maxBytesPerGatheringWrite);
                     in.removeBytes(localWrittenBytes);
                     --writeSpinCount;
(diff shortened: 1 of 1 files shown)


transport/src/main/java/io/netty/channel/socket/nio/NioSocketChannel.java
```diff
@@ -399,12 +399,12 @@ protected void doWrite(ChannelOutboundBuffer in) throws Exception {
                     // Zero length buffers are not added to nioBuffers by ChannelOutboundBuffer, so there is no need
                     // to check if the total size of all the buffers is non-zero.
                     ByteBuffer buffer = nioBuffers[0];
-                    int attemptedBytes = buffer.remaining();
                     final int localWrittenBytes = ch.write(buffer);
                     if (localWrittenBytes <= 0) {
                         incompleteWrite(true);
                         return;
                     }
+                    int attemptedBytes = buffer.remaining();
```
normanmaurer: This is not correct as the write will modify the amount of remaining bytes on success

transport/src/main/java/io/netty/channel/socket/nio/NioSocketChannel.java
```diff
@@ -414,13 +414,13 @@ protected void doWrite(ChannelOutboundBuffer in) throws Exception {
...
                     // to check if the total size of all the buffers is non-zero.
                     // We limit the max amount to int above so cast is safe
-                    long attemptedBytes = in.nioBufferSize();
                     final long localWrittenBytes = ch.write(nioBuffers, 0, nioBufferCnt);
                     if (localWrittenBytes <= 0) {
                         incompleteWrite(true);
                         return;
                     }
                     // Casting to int is safe because we limit the total amount of data in the nioBuffers to int above.
+                    long attemptedBytes = in.nioBufferSize();
```
normanmaurer: sorry this is also incorrect... nioBufferSize() will also not produce the correct value after write was called.

normanmaurer:
Because of the reasons stated in the review comments I think this PR is not correct and should be closed.

carryxyh:
Oh, did not know about it. thx.

transport/src/main/java/io/netty/channel/socket/nio/NioSocketChannel.java
carryxyh: is this attemptedBytes incorrect in default case ?
this is the value in ChannelOutboundBuffer and only be changed in `in.nioBuffers(1024, maxBytesPerGatheringWrite);`.
And the ch.write just write this ByteBuffer[] into channel, is this will change the attemptedBytes value?

carryxyh: I just debug here and get the same value between before write and after write, in whick case the attemptedBytes will changed? I will have a try :>

PR #8080, opened 2018-06-28
HTTP/2 goaway connection state update sequencing

Scottmitch:
Motivation:
The Http2Connection state is updated by the DefaultHttp2ConnectionDecoder after the frame listener is notified of the goaway frame. If the listener sends a frame synchronously this means the connection state will not know about the goaway it just received and we may send frames that are not allowed on the connection. This may also mean a stream object is created but it may never get taken out of the stream map unless some other event occurs (e.g. timeout).

Modifications:
- The Http2Connection state should be updated before the listener is notified of the goaway
- The Http2Connection state modification and validation should be self contained when processing a goaway instead of partially in the decoder.

Result:
No more creating streams and sending frames after a goaway has been sent or received.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
@@ -225,7 +225,11 @@ public class DefaultHttp2Connection implements Http2Connection {
     }
 
     @Override
-    public void goAwayReceived(final int lastKnownStream, long errorCode, ByteBuf debugData) {
+    public void goAwayReceived(final int lastKnownStream, long errorCode, ByteBuf debugData) throws Http2Exception {
+        if (localEndpoint.lastStreamKnownByPeer() >= 0 && localEndpoint.lastStreamKnownByPeer() < lastKnownStream) {
+            throw connectionError(PROTOCOL_ERROR, "lastStreamId MUST NOT increase. Current value: %d new value: %d",
+                    localEndpoint.lastStreamKnownByPeer(), lastKnownStream);
+        }
         localEndpoint.lastStreamKnownByPeer(lastKnownStream);
         for (int i = 0; i < listeners.size(); ++i) {
             try {
@@ -235,19 +239,15 @@ public class DefaultHttp2Connection implements Http2Connection {
             }
         }
 
-        try {
-            forEachActiveStream(new Http2StreamVisitor() {
-                @Override
-                public boolean visit(Http2Stream stream) {
-                    if (stream.id() > lastKnownStream && localEndpoint.isValidStreamId(stream.id())) {
-                        stream.close();
-                    }
-                    return true;
+        forEachActiveStream(new Http2StreamVisitor() {
+            @Override
+            public boolean visit(Http2Stream stream) {
+                if (stream.id() > lastKnownStream && localEndpoint.isValidStreamId(stream.id())) {
+                    stream.close();
                 }
-            });
-        } catch (Http2Exception e) {
-            PlatformDependent.throwException(e);
-        }
+                return true;
+            }
+        });
     }
 
     @Override
@@ -256,7 +256,20 @@ public class DefaultHttp2Connection implements Http2Connection {
     }
 
     @Override
-    public void goAwaySent(final int lastKnownStream, long errorCode, ByteBuf debugData) {
+    public boolean goAwaySent(final int lastKnownStream, long errorCode, ByteBuf debugData) throws Http2Exception {
+        if (remoteEndpoint.lastStreamKnownByPeer() >= 0) {
+            // Protect against re-entrancy. Could happen if writing the frame fails, and error handling
+            // treating this is a connection handler and doing a graceful shutdown...
+            if (lastKnownStream == remoteEndpoint.lastStreamKnownByPeer()) {
+                return false;
+            }
+            if (lastKnownStream > remoteEndpoint.lastStreamKnownByPeer()) {
+                throw connectionError(PROTOCOL_ERROR, "Last stream identifier must not increase between " +
+                                "sending multiple GOAWAY frames (was '%d', is '%d').",
+                        remoteEndpoint.lastStreamKnownByPeer(), lastKnownStream);
+            }
+        }
+
         remoteEndpoint.lastStreamKnownByPeer(lastKnownStream);
         for (int i = 0; i < listeners.size(); ++i) {
             try {
@@ -266,19 +279,16 @@ public class DefaultHttp2Connection implements Http2Connection {
             }
         }
 
-        try {
-            forEachActiveStream(new Http2StreamVisitor() {
-                @Override
-                public boolean visit(Http2Stream stream) {
-                    if (stream.id() > lastKnownStream && remoteEndpoint.isValidStreamId(stream.id())) {
-                        stream.close();
-                    }
-                    return true;
+        forEachActiveStream(new Http2StreamVisitor() {
+            @Override
+            public boolean visit(Http2Stream stream) {
+                if (stream.id() > lastKnownStream && remoteEndpoint.isValidStreamId(stream.id())) {
+                    stream.close();
                 }
-            });
-        } catch (Http2Exception e) {
-            PlatformDependent.throwException(e);
-        }
+                return true;
+            }
+        });
+        return true;
     }
 
     /**
@@ -863,10 +873,10 @@ public class DefaultHttp2Connection implements Http2Connection {
 
         private void checkNewStreamAllowed(int streamId, State state) throws Http2Exception {
             assert state != IDLE;
-            if (goAwayReceived() && streamId > localEndpoint.lastStreamKnownByPeer()) {
+            if (lastStreamKnownByPeer >= 0 && streamId > lastStreamKnownByPeer) {
                 throw streamError(streamId, REFUSED_STREAM,
-                      "Cannot create stream %d since this endpoint has received a GOAWAY frame with last stream id %d.",
-                      streamId, localEndpoint.lastStreamKnownByPeer());
+                        "Cannot create stream %d greater than Last-Stream-ID %d from GOAWAY.",
+                        streamId, lastStreamKnownByPeer);
             }
             if (!isValidStreamId(streamId)) {
                 if (streamId < 0) {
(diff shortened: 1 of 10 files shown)


codec-http2/src/main/java/io/netty/handler/codec/http2/Http2Connection.java
```diff
@@ -327,15 +327,17 @@
...
     boolean goAwaySent();
 
     /**
-     * Indicates that a {@code GOAWAY} was sent to the remote endpoint and sets the last known stream.
+     * Updates the local state of this {@link Http2Connection} as a result of a {@code GOAWAY} to send to the remote
+     * endpoint.
+     * @return {@code true} if the corresponding {@code GOAWAY} frame should be sent to the remote endpoint.
      */
-    void goAwaySent(int lastKnownStream, long errorCode, ByteBuf message);
+    boolean goAwaySent(int lastKnownStream, long errorCode, ByteBuf message) throws Http2Exception;
```
Scottmitch: note this API change. I think it makes more sense to have the connection validate these conditions, and is consistent with other connection level validation. Having the validation outside can lead to the issue fixed in this PR (the state was updated at the wrong time).

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
```diff
@@ -225,7 +225,11 @@ public boolean goAwayReceived() {
     }
 
     @Override
-    public void goAwayReceived(final int lastKnownStream, long errorCode, ByteBuf debugData) {
+    public void goAwayReceived(final int lastKnownStream, long errorCode, ByteBuf debugData) throws Http2Exception {
+        if (localEndpoint.lastStreamKnownByPeer() >= 0 && localEndpoint.lastStreamKnownByPeer() < lastKnownStream) {
+            throw connectionError(PROTOCOL_ERROR, "lastStreamId MUST NOT increase. Current value: %d new value: %d",
```
normanmaurer: I am correct that the debugData does not need to be released here as the ownership of the buffer is not transferred ?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
```diff
@@ -256,7 +256,20 @@ public boolean goAwaySent() {
...
-    public void goAwaySent(final int lastKnownStream, long errorCode, ByteBuf debugData) {
+    public boolean goAwaySent(final int lastKnownStream, long errorCode, ByteBuf debugData) throws Http2Exception {
+        if (remoteEndpoint.lastStreamKnownByPeer() >= 0) {
+            // Protect against re-entrancy. Could happen if writing the frame fails, and error handling
+            // treating this is a connection handler and doing a graceful shutdown...
+            if (lastKnownStream == remoteEndpoint.lastStreamKnownByPeer()) {
+                return false;
+            }
+            if (lastKnownStream > remoteEndpoint.lastStreamKnownByPeer()) {
+                throw connectionError(PROTOCOL_ERROR, "Last stream identifier must not increase between " +
```
normanmaurer: I am correct that the debugData does not need to be released here as the ownership of the buffer is not transferred ?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
```diff
@@ -266,19 +279,16 @@ public void goAwaySent(final int lastKnownStream, long errorCode, ByteBuf debugD
...
-            forEachActiveStream(new Http2StreamVisitor() {
-                @Override
-                public boolean visit(Http2Stream stream) {
-                    if (stream.id() > lastKnownStream && remoteEndpoint.isValidStreamId(stream.id())) {
-                        stream.close();
-                    }
-                    return true;
+        forEachActiveStream(new Http2StreamVisitor() {
+            @Override
+            public boolean visit(Http2Stream stream) {
```
normanmaurer: Seems like this is the same code as https://github.com/netty/netty/pull/8080/files#diff-9d991ca64301bcd5b3c00d053591f485R244 . Can we just create an `HttpStreamVisitor` imp that we share  by passing the needed stuff via the constructor ?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -541,6 +537,7 @@ private boolean shouldIgnoreHeadersOrDataFrame(ChannelHandlerContext ctx, int st
                                 ("Stream created after GOAWAY sent. Last known stream by peer " +
                                  connection.remote().lastStreamKnownByPeer()));
                 }
+                stream.close();
```
normanmaurer: can you add a comment why this was added ?

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionTest.java
```diff
@@ -425,7 +425,7 @@ public void reserveWithPushDisallowedShouldThrow() throws Http2Exception {
     @Test(expected = Http2Exception.class)
     public void goAwayReceivedShouldDisallowCreation() throws Http2Exception {
         server.goAwayReceived(0, 1L, Unpooled.EMPTY_BUFFER);
-        server.remote().createStream(3, true);
+        server.local().createStream(3, true);
```
normanmaurer: why this change ?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
Scottmitch: yes. I added a comment to the javadocs of the interface.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
Scottmitch: yes. I added a comment to the javadocs of the interface.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Connection.java
Scottmitch: I extracted the iteration into a shared method.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
Scottmitch: yip done.

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionTest.java
Scottmitch: the unit test was incorrect and highlighted a bug addressed by this PR.

The test simulates receiving a go away frame, then created a stream from the remote endpoint, and expected an error. However this is not correct. Receiving a goaway frame should prevent the *local* endpoint from creating streams and have no effect on the *remote* endpoint.

Scottmitch: I added some additional tests to catch all the cases and clarify.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -541,6 +537,10 @@ private boolean shouldIgnoreHeadersOrDataFrame(ChannelHandlerContext ctx, int st
                                 ("Stream created after GOAWAY sent. Last known stream by peer " +
                                  connection.remote().lastStreamKnownByPeer()));
                 }
+
+                // If we are ignoring frames for this stream it means that we no longer need to retain state for this
+                // stream and can close it, which should remove it from the active stream map.
+                stream.close();
```
bryce-anderson: It doesn't seem like the responsibility of the decoder to manage the streams life-cycle Under what conditions would this happen? It seems that sending a RST should always be coupled with closing the stream and you've covered the GOAWAY case above. From my limited digging, it looks like the [`Http2ConnectionHandler`](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java#L778) is the only way we'll hit the `isResetSent()` case and it's already setup a listener to close the session. I presume it's only delayed to make draining work correctly with resets.

normanmaurer: @Scottmitch can you chime in here ?

Scottmitch: This was added as a fail-safe because prior to this PR we would still create streams initiated by the remote endpoint even if we sent an goaway. This resulted in "leaked" stream objects (stream objects would be created, inserted into the `streamMap` but never removed because there would be no lifetime event to trigger removal). With this PR `DefaultHttp2Connection` no longer permits this and I don't think we should hit this code path because:

1. as you said sending a rst should close the stream, which will remove from the `streamMap`.
2. we should no longer create streams initiated by the remote peer (exceeding the last stream ID) after a go away was sent.

so I can remove this code block if we assume this scenario is invalid for any implementation of `Http2Connection`. We have unit tests which do exercise this code path in `DefaultHttp2ConnectionDecoderTest` but we can modify the test/mocks to not

bryce-anderson: I can't attest to other implementations of `Http2Connection`, but if that's a legit concern I don't mind if we leave it if the worst case scenario is it's dead code. I'll trust your judgment @Scottmitch.

Scottmitch: I'm not aware of any alternative implementation of `Http2Connection` and you have a good point this code change seems to be fixing an issue that stems from elsewhere. Now that `DefaultHttp2Connection` is fixed it should be sufficient to remove this with Netty's code, but I'll add a comment to clarify the assumptions.

PR #8084, opened 2018-06-29
Add OpenSslX509KeyManagerFactory which makes it even easier for peopl…

normanmaurer:
…e to get the maximum performance when using OpenSSL / LibreSSL / BoringSSL  with netty.

Motivation:

To make it even easier for people to get the maximum performance when using native SSL we should provide our own KeyManagerFactory implementation that people can just use to configure their key material.

Modifications:

- Add OpenSslX509KeyManagerFactory which users can use for maximum performance with native SSL
- Refactor some internal code to re-use logic and not duplicate it.

Result:

Easier to get the max performance out of native SSL implementation.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/OpenSslX509KeyManagerFactory.java
+++ b/handler/src/main/java/io/netty/handler/ssl/OpenSslX509KeyManagerFactory.java
@@ -1,0 +1,220 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.ssl;
+
+import io.netty.buffer.ByteBufAllocator;
+import io.netty.buffer.UnpooledByteBufAllocator;
+import io.netty.util.ReferenceCountUtil;
+import io.netty.util.internal.ObjectUtil;
+
+import javax.net.ssl.KeyManager;
+import javax.net.ssl.KeyManagerFactory;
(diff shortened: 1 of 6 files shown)


handler/src/main/java/io/netty/handler/ssl/OpenSslX509KeyManagerFactory.java
```diff
@@ -0,0 +1,220 @@
...
+
+            /**
+             * {@link OpenSslKeyMaterialProvider} implementation that pre-compute the {@link OpenSslKeyMaterial} for
+             * all aliases.
+             */
+            private static final class OpenSslPopulatedKeyMaterialProvider extends OpenSslKeyMaterialProvider {
+                private final Map<String, Object> materialMap;
+
+                OpenSslPopulatedKeyMaterialProvider(
+                        X509KeyManager keyManager, String password, Iterator<String> aliases) {
```
carl-mastrangelo: not sure I understand why the final arg is Iterator.     If this is a private class, and only has one callsite for the ctor, why not pass the iterable directly?   That would make the loop a for rather than a while, and not need to call `next` and `hasNext`

normanmaurer: @carl-mastrangelo it doesn't really matter to me... let me change it.

normanmaurer:
@carl-mastrangelo any other comments ? Also @mosesn @bryce-anderson @ejona86 any comments ? I would love to include this as part of the next release

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslServerContext.java
```diff
@@ -121,9 +121,16 @@ static ServerContext newSessionContext(ReferenceCountedOpenSslContext thiz, long
...
+                        char[] keyPasswordChars = keyStorePassword(keyPassword);
+                        KeyStore ks = buildKeyStore(keyCertChain, key, keyPasswordChars);
+                        if (ks.aliases().hasMoreElements()) {
+                            keyManagerFactory = new OpenSslX509KeyManagerFactory();
+                        } else {
+                            keyManagerFactory = new OpenSslCachingX509KeyManagerFactory(
+                                    KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm()));
+                        }
+                        keyManagerFactory.init(ks, keyPasswordChars);
+                        keyMaterialProvider = providerFor(keyManagerFactory, keyPassword);
```
bryce-anderson: This is now a duplicate of the else so it should be able to be pulled out of the branchs.

Issue #8072, opened 2018-06-28
Add TcpQuickAck Support


hc-codersatlas:
See https://bugs.openjdk.java.net/browse/JDK-8145635 - it's in Java 10.

jdk.net.ExtendedSocketOption (instead of StandardSocketOption is where it is)

For consistency as it's already in the native Epoll module.

hc-codersatlas:
Looks good. Can close once it's merged. Thanks!

PR #8085, opened 2018-06-29
Support the usage of SocketOption when nio is used and the java versi…

normanmaurer:
…on >= 7.

Motivation:

The JDK uses SocketOption since java7 to support configuration options on the underyling Channel. We should allow to create a ChannelOption from a given SocketOption if nio is used. This also allows us to expose the same featureset in terms of configuration as the java nio implementation does without any extra effort.

Modifications:

- Add NioChannelOption which allows to wrap an existing SocketOption which then can be applied to the nio transport.
- Add test-cases

Result:

Support the same configuration options as the JDK. Also fixes https://github.com/netty/netty/issues/8072.

### DIFF
--- a/transport/src/main/java/io/netty/channel/socket/nio/NioChannelOption.java
+++ b/transport/src/main/java/io/netty/channel/socket/nio/NioChannelOption.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.channel.socket.nio;
+
+import io.netty.channel.ChannelException;
+import io.netty.channel.ChannelOption;
+
+import java.io.IOException;
+import java.net.SocketOption;
+import java.nio.channels.NetworkChannel;
+import java.util.Set;
+
+/**
+ * Provides {@link ChannelOption} over a given {@link SocketOption} which is then passed through the underlying
+ * {@link NetworkChannel}.
+ */
+public final class NioChannelOption<T> extends ChannelOption<T> {
+
+    private final SocketOption<T> option;
+
+    @SuppressWarnings("deprecation")
+    private NioChannelOption(SocketOption<T> option) {
+        super(option.name());
+        this.option = option;
+    }
+
+    /**
+     * Returns a {@link ChannelOption} for the given {@link SocketOption}.
+     */
+    public static <T> ChannelOption<T> of(SocketOption<T> option) {
+        return new NioChannelOption<T>(option);
+    }
+
+    // Internal helper methods to remove code duplication between Nio*Channel implementations.
+    static <T> boolean setOption(NetworkChannel channel, NioChannelOption<T> option, T value) {
+        try {
+            channel.setOption(option.option, value);
+            return true;
+        } catch (UnsupportedOperationException ignore) {
+            return false;
+        } catch (IOException e) {
+            throw new ChannelException(e);
+        }
+    }
+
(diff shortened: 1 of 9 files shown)


transport/src/main/java/io/netty/channel/socket/nio/NioChannelOption.java
```diff
@@ -0,0 +1,75 @@
...
+     * Returns a {@link ChannelOption} for the given {@link SocketOption}.
+     */
+    public static <T> ChannelOption<T> of(SocketOption<T> option) {
+        return new NioChannelOption<T>(option);
+    }
+
+    // Internal helper methods to remove code duplication between Nio*Channel implementations.
+    static <T> void setOption(NetworkChannel channel, NioChannelOption<T> option, T value) {
+        try {
+            channel.setOption(option.option, value);
```
bryce-anderson: This can throw a couple exceptions that signal that the argument is invalid (`UnsupportedOperationException`, `IllegalArgumentException`). Should they be caught and turned into a return value of `false`? that would seem to better match the semantics of `boolean ChannelConfig.setOption(..)`.

normanmaurer: Good idea... let me do this

normanmaurer: @bryce-anderson done...

transport/src/main/java/io/netty/channel/socket/nio/NioChannelOption.java
```diff
@@ -0,0 +1,79 @@
...
+     * Returns a {@link ChannelOption} for the given {@link SocketOption}.
+     */
+    public static <T> ChannelOption<T> of(SocketOption<T> option) {
+        return new NioChannelOption<T>(option);
+    }
+
+    // Internal helper methods to remove code duplication between Nio*Channel implementations.
+    static <T> boolean setOption(NetworkChannel channel, NioChannelOption<T> option, T value) {
+        try {
+            channel.setOption(option.option, value);
```
bryce-anderson: I think we still need a `return true;`. Do we not want to catch the `IllegalArgumentException`? I could see either way being the right thing here, though I would have expected a return value of false in that case.

transport/src/main/java/io/netty/channel/socket/nio/NioChannelOption.java
```diff
@@ -0,0 +1,79 @@
...
+        } catch (UnsupportedOperationException ignore) {
+            return false;
```
normanmaurer: damn it... forgot to amend.

transport/src/main/java/io/netty/channel/socket/nio/NioChannelOption.java
normanmaurer: I think IllegalArgumentException should be thrown

normanmaurer:
@bryce-anderson also added more tests for unsupported options.

normanmaurer:
@mosesn @bryce-anderson @ejona86 @carl-mastrangelo any comments ? I would love to include this as part of the next release

transport/src/main/java/io/netty/channel/socket/nio/NioChannelOption.java
bryce-anderson: Good enough for me.

normanmaurer:
@mosesn @ejona86 @carl-mastrangelo ping ...

normanmaurer:
And @trustin  :)

transport/src/main/java/io/netty/channel/socket/nio/NioChannelOption.java
```diff
@@ -0,0 +1,80 @@
...
+    public static <T> ChannelOption<T> of(SocketOption<T> option) {
+        return new NioChannelOption<T>(option);
+    }
+
+    // Internal helper methods to remove code duplication between Nio*Channel implementations.
+    static <T> boolean setOption(NetworkChannel channel, NioChannelOption<T> option, T value) {
+        try {
+            channel.setOption(option.option, value);
+            return true;
+        } catch (UnsupportedOperationException ignore) {
```
ejona86: Did you consider checking supportedOptions instead of catching the exception?
```
if (!channel.supportedOptions().contains(option.value)) {
  return false;
}
```

Given the design of the API, I'd expect every implementation to have `supportedOptions()` be a simple getter. This is at least true for all the implementations in the JDK.

develar:
Have you tested it under JDK 6?

normanmaurer:
@develar there was a bug with java6 which has be fixed : 
https://github.com/netty/netty/pull/8168

PR #8086, opened 2018-06-29
Don't send a RST on close if the stream never existed

bryce-anderson:
Motivation:

When a Http2MultiplexCodec stream channel fails to write the first
HEADERS it will forcibly close, and that will trigger sending a
RST_STREAM, which is commonly a connection level protocol error. This is
because it has what looks like a valid stream id, but didn't check with
the connection as to whether the stream may have actually existed.

Modifications:

Instead of checking if the stream was just a valid looking id ( > 0) we
check with the connection as to whether it may have existed at all.

Result:

We no longer send a RST_STREAM frame from Http2MultiplexCodec for idle
streams.

----------------------------

This is mostly for discussion at this time, in part because I don't know the right way to test it. In the Http2MultiplexCodecTest we override the `write` method which is where we get a stream id assigned, so it would need to either be more invasive or go in a different suite. 

The issue is that the stream thinks it's alive because it has a valid looking stream id (set by the [`Http2FrameCodec`](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java#L359)) but it's not. This could be fixed via https://github.com/netty/netty/pull/8067, but not everyone agrees it's the right thing to do. That said, I don't see this commit as been predicated on the other, or vise versa.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
@@ -906,9 +906,10 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
 
                 final boolean wasActive = isActive();
 
-                // Only ever send a reset frame if the connection is still alive as otherwise it makes no sense at
-                // all anyway.
-                if (parent().isActive() && !streamClosedWithoutError && isStreamIdValid(stream().id())) {
+                // Only ever send a reset frame if the connection is still alive and if the stream may have existed
+                // as otherwise we may send a RST on a stream in an invalid state and cause a connection error.
+                if (parent().isActive() && !streamClosedWithoutError &&
+                        connection().streamMayHaveExisted(stream().id())) {
                     Http2StreamFrame resetFrame = new DefaultHttp2ResetFrame(Http2Error.CANCEL).stream(stream());
                     write(resetFrame, unsafe().voidPromise());
                     flush();
--- a/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
+++ b/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
@@ -309,6 +309,43 @@ public class Http2MultiplexCodecTest {
     }
 
     @Test
+    public void outboundStreamShouldNotWriteResetFrameOnClose_IfStreamDidntExist() {
+        writer = new Writer() {
+            private boolean headersWritten;
+            @Override
+            void write(Object msg, ChannelPromise promise) {
+                // We want to fail to write the first headers frame. This is what happens if the connection
+                // refuses to allocate a new stream due to having received a GOAWAY.
+                if (!headersWritten && msg instanceof Http2HeadersFrame) {
+                    headersWritten = true;
+                    Http2HeadersFrame headersFrame = (Http2HeadersFrame) msg;
+                    final TestableHttp2MultiplexCodec.Stream stream =
+                            (TestableHttp2MultiplexCodec.Stream) headersFrame.stream();
+                    stream.id = 1;
+                    promise.setFailure(new Exception("boom"));
+                } else {
+                    super.write(msg, promise);
+                }
+            }
+        };
+
+        childChannelInitializer.handler = new ChannelInboundHandlerAdapter() {
+            @Override
+            public void channelActive(ChannelHandlerContext ctx) throws Exception {
+                ctx.writeAndFlush(new DefaultHttp2HeadersFrame(new DefaultHttp2Headers()));
+                ctx.fireChannelActive();
+            }
+        };
+
+        Channel childChannel = newOutboundStream();
+        assertFalse(childChannel.isActive());
+
+        childChannel.close();
+        parentChannel.runPendingTasks();
+        assertTrue(parentChannel.outboundMessages().isEmpty());
+    }
+
+    @Test
     public void inboundRstStreamFireChannelInactive() {
         LastInboundHandler inboundHandler = streamActiveAndWriteHeaders(inboundStream);
         assertTrue(inboundHandler.isChannelActive());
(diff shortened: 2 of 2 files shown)


codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -908,7 +908,8 @@ public void operationComplete(ChannelFuture future) throws Exception {
 
                 // Only ever send a reset frame if the connection is still alive as otherwise it makes no sense at
                 // all anyway.
-                if (parent().isActive() && !streamClosedWithoutError && isStreamIdValid(stream().id())) {
+                if (parent().isActive() && !streamClosedWithoutError &&
+                        connection().streamMayHaveExisted(stream().id())) {
```
normanmaurer: Can you update the comment to reflect the change ?

bryce-anderson:
This might also be fixed by checking this condition in `Http2ConnectionHandler.resetUnknownStream`, though it causes a few tests to fail. That may only be because they don't mock the expected session state correctly, or it may be called too early. I'll look into it.

bryce-anderson:
It looks like the change in the `Http2ConnectionHandler` works with some fixups to some mocks. Either the change in `Http2ConnectionHander` or `Http2MultiplexCodec` should fix my problem. The `Http2ConnectionHandler` change should be more general, but I left the `Http2MultiplexCodec` change in there to avoid unnecessary work. I can remove that part if people don't think it's valuable or it turns out there is an unforeseen issue with the change to `Http2ConnectionHandler`.

bryce-anderson:
@normanmaurer, I think I've addressed your comment. This was expanded quite a bit to include the `Http2ConnectionHandler` which should make this useful for both the old API and the `Http2MultiplexCodec` API, so it probably just needs a re-review.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -906,9 +906,10 @@ public void operationComplete(ChannelFuture future) throws Exception {
 
                 final boolean wasActive = isActive();
 
-                // Only ever send a reset frame if the connection is still alive as otherwise it makes no sense at
-                // all anyway.
-                if (parent().isActive() && !streamClosedWithoutError && isStreamIdValid(stream().id())) {
+                // Only ever send a reset frame if the connection is still alive and if the stream ever actually
```
Scottmitch: `ever actually...` -> `may have exists as otherwise we may send a RST on a stream in an invalid state and cause a connection error`. consider rewording to clarify the implications of not checking the conditions.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -731,6 +731,12 @@ protected Http2FrameWriter frameWriter() {
      */
     private ChannelFuture resetUnknownStream(final ChannelHandlerContext ctx, int streamId, long errorCode,
                                              ChannelPromise promise) {
+
+        if (!connection().streamMayHaveExisted(streamId)) {
```
Scottmitch: is this necessary after the [recent change to streamError](https://github.com/netty/netty/commit/c321e8ea4a1349125510bd9156a4e9a4c3e32741#diff-ce33d9374618cacba50c2b6020df1c86R704), and with your changes to `Http2MultiplexCodec` in this PR? IIUC from the perspective of this class the only "invalid" way this can be called is if the "user" explicitly writes a RST_STREAM frame for a bogus stream.

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2ConnectionHandlerTest.java
```diff
@@ -518,14 +518,25 @@ public void encoderAndDecoderAreClosedOnChannelInactive() throws Exception {
...
                                         eq(STREAM_CLOSED.code()), eq(promise))).thenReturn(future);
         handler.resetStream(ctx, NON_EXISTANT_STREAM_ID, STREAM_CLOSED.code(), promise);
         verify(frameWriter).writeRstStream(eq(ctx), eq(NON_EXISTANT_STREAM_ID), eq(STREAM_CLOSED.code()), eq(promise));
     }
 
+    @Test
+    public void writeRstOnNonExistantStreamThatMayNotHaveExistShouldNotSucceed() throws Exception {
+        handler = newHandler();
+        when(connection.streamMayHaveExisted(NON_EXISTANT_STREAM_ID)).thenReturn(false);
+        when(frameWriter.writeRstStream(eq(ctx), eq(NON_EXISTANT_STREAM_ID),
```
Scottmitch: nit: consider removing this as it shouldn't be necessary because we don't expect it to be called and are validating it won't be called in this test.

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
```diff
@@ -286,7 +286,7 @@ public void idleOutboundStreamShouldNotWriteResetFrameOnClose() {
     }
 
     @Test
-    public void outboundStreamShouldWriteResetFrameOnClose_headersSent() {
+    public void outboundStreamShouldWriteResetFrameOnClose_headersSent() throws Exception {
```
Scottmitch: `throws Exception` can be removed here.

Scottmitch: Did we want a unit test in `Http2MultiplexCodecTest` that validates we don't try to write a reset under the conditions in question?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
bryce-anderson: That's right, the problem is that `Http2MultiplexCodec` is doing the bad thing of explicitly writing a RST_STREAM. I added this because it was easier to test and looked like it would fix a problem in both the multiplex and original API's, but only one should be sufficient. Should we fix both places, or just `Http2MultiplexCodec`?

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
bryce-anderson: done.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
bryce-anderson: done.

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2ConnectionHandlerTest.java
bryce-anderson: done.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
Scottmitch: I vote for just `Http2MultiplexCodec`. Better to fix the "root" cause where this feels more like a band-aid. If protecting applications from sending bogus RST is also useful in the future we can address it with a followup PR.

bryce-anderson: done.

bryce-anderson:
Any more feedback would be appreciated. I'd love to get this into 4.1.26-Final, it's the last (known) thing in the way of using the new API in our client.

Scottmitch:
lgtm .. thanks @bryce-anderson!

Issue #8090, opened 2018-07-01
Fails to load native library when shading Netty epoll or kqueue (osx or linux)


cchantep:
### Expected behavior

With shaded Netty libraries, succeed to call `foo.io.netty.channel.epoll.Epoll.ensureAvailability()` (with `foo.` being the package prefix) or `foo.io.netty.channel.kqueue.KQueue.ensureAvailability()` without `Exception`.

### Actual behavior

When calling the `.ensureAvailability()` or directly loading a Netty native lib, then an `UnsatisfiedLinkError` is thrown with root cause `RuntimeException: failed to get method ID: DatagramSocketAddress.<init>(String, int, int, DatagramSocketAddress) [error] at java.base/java.lang.ClassLoader$NativeLibrary.load(Native Method)`.

*Mac OS X/kqueue:*

```
[...]

PR #8091, opened 2018-07-02
Fix support for shading native libraries which was broken in b818852c…

normanmaurer:
…db0ee56ce3cf939e2faa538d519baabd.

Motivation:

b818852cdb0ee56ce3cf939e2faa538d519baabd broke support for shading the native libraries in netty as it missed to respect the package prefix that is used when shading.

Modifications:

Correctly respect package prefix for constructor argument and include the used classname when logging that we could not find the constructor.

Result:

Be able to shade native libraries of netty again.

### DIFF
--- a/transport-native-unix-common/src/main/c/netty_unix_socket.c
+++ b/transport-native-unix-common/src/main/c/netty_unix_socket.c
@@ -1010,21 +1010,32 @@ jint netty_unix_socket_JNI_OnLoad(JNIEnv* env, const char* packagePrefix) {
     dynamicMethods = NULL;
     char* nettyClassName = netty_unix_util_prepend(packagePrefix, "io/netty/channel/unix/DatagramSocketAddress");
     jclass localDatagramSocketAddressClass = (*env)->FindClass(env, nettyClassName);
-    free(nettyClassName);
-    nettyClassName = NULL;
     if (localDatagramSocketAddressClass == NULL) {
+        free(nettyClassName);
+        nettyClassName = NULL;
         // pending exception...
         return JNI_ERR;
     }
     datagramSocketAddressClass = (jclass) (*env)->NewGlobalRef(env, localDatagramSocketAddressClass);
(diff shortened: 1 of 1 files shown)


normanmaurer:
@cchantep please check as well ... thanks!

normanmaurer:
This fix was verified by @cchantep in https://github.com/netty/netty/issues/8090.

normanmaurer:
I will work on a integration test for this one so we will not break it again

transport-native-unix-common/src/main/c/netty_unix_socket.c
```diff
@@ -1010,21 +1010,32 @@ jint netty_unix_socket_JNI_OnLoad(JNIEnv* env, const char* packagePrefix) {
...
+        free(nettyClassName);
+        nettyClassName = NULL;
         // out-of-memory!
         netty_unix_errors_throwOutOfMemoryError(env);
         return JNI_ERR;
     }
-    datagramSocketAddrMethodId = (*env)->GetMethodID(env, datagramSocketAddressClass, "<init>", "(Ljava/lang/String;IILio/netty/channel/unix/DatagramSocketAddress;)V");
+
+    // Respect shading...
+    char parameters[1024] = {0};
```
johnou: Wouldn't it be safer to calculate the length instead of using an arbitrary size?

normanmaurer: We could... that said if 1024 is not enough I would be really surprised

cchantep:
@normanmaurer do you know whether it's published on snapshot repo ? To check with CI?

transport-native-unix-common/src/main/c/netty_unix_socket.c
johnou: I was thinking in terms of buffer overflow but I assume snprintf guards against that? couldn't we do a simple check that nettyClassName does not exceed 1024 then throw illegalargumentexception?

PR #8096, opened 2018-07-03
Deprecate Unpooled.unmodifiableBuffer(ByteBuf...)

njhill:
Motivation:

https://github.com/netty/netty/pull/8040 introduced `Unpooled.wrappedUnmodifiableBuffer(ByteBuf...)` which has the same behaviour but wraps the provided array directly. This is preferred for most uses (including varargs-based use) and if there are any unusual cases of an explicit array which is re-used before the `ByteBuf` is finished with, it can just be copied first.

Modifications:

Added `@Deprecated` annotation and corresponding javadoc addition to `Unpooled.unmodifiableBuffer(ByteBuf...)`.

Result:

`Unpooled.unmodifiableBuffer(ByteBuf...)` will be deprecated.


normanmaurer:
@njhill thanks!

PR #8099, opened 2018-07-04
netty_unix_socket free nettyClassName after using it

Scottmitch:
Motivation:
netty_unix_socket attempts to use nettyClassName in an error message, but previously freed the memory. We should wait to free the memory until after we use it.

Modifications:
- Free nettyClassName after using it in snprintf

Result:
More useful error message.


normanmaurer:
Oh boy.... sorry and thanks for the great catch

Issue #8107, opened 2018-07-09
Log4J2 Auto-detection


apimastery:
Not sure if I'm missing something...

[#5047](https://github.com/netty/netty/pull/5047) added Log4J2LoggerFactory and Log4J2Logger. It appears that Log4J2 auto-detection in InternalLoggerFactory wasn't added?
https://github.com/netty/netty/blob/fef462c04333950201ddbfd505c0af730a0c2ffd/common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java#L39

Cheers!

normanmaurer:
@apimastery good catch... See https://github.com/netty/netty/pull/8109

PR #5047, opened 2016-03-26
Add Log4J2LoggerFactory and Log4J2Logger

windie:
Motivation:

See #3095

Modifications:

Add Log4J2LoggerFactory and Log4J2Logger which is an InternalLogger implementation based on log4j2.

Result:

The user can use log4j2 directly without a special slf4j binding.

PR #8109, opened 2018-07-09
Auto-detect Log4J2 for logging if on the class-path

normanmaurer:
Motivation:

https://github.com/netty/netty/pull/5047 added Log4J2 support but missed to add code to try to auto-detect it.

Modifications:

Try to use Log4JLoggerFactory by default.

Result:

Fixes https://github.com/netty/netty/issues/8107.


windie:
LGTM. Thanks for fixing this.

Issue #8078, opened 2018-06-28
Avoid creating FileInputStream and FileOutputStream for obtaining FileChannel


leventov:
There are multiple places in the production code of the project, e. g. in `ChunkedNioFile`, where `FileInputStream` or `FileOutputStream` are created solely to obtain a `FileChannel` from them. Until Netty projects bumps the compatibility level to Java 7 (where `FileChannel.open()` methods were added), it's better to obtain a FileChannel by means of creating a `RandomAccessFile`, because the latter doesn't override `Object.finalize()`. Also, I would extract those pieces of code into a utility methods like `openChannelForRead(File)` and `openChannelForWrite(File)`.

hc-codersatlas:
Same goes for BufferedReader instances - same finalize problem. It's used in DNS and some other modules.

normanmaurer:
@leventov good catch... Check https://github.com/netty/netty/pull/8110

PR #8110, opened 2018-07-09
Avoid creating FileInputStream and FileOutputStream for obtaining Fil…

normanmaurer:
…eChannel

Motivation:

If all we need is the FileChannel we should better use RandomAccessFile as FileInputStream and FileOutputStream use a finalizer.

Modifications:

Replace FileInputStream and FileOutputStream with RandomAccessFile when possible.

Result:

Fixes https://github.com/netty/netty/issues/8078.

### DIFF
--- a/buffer/src/test/java/io/netty/buffer/ReadOnlyDirectByteBufferBufTest.java
+++ b/buffer/src/test/java/io/netty/buffer/ReadOnlyDirectByteBufferBufTest.java
@@ -21,9 +21,8 @@ import org.junit.Test;
 
 import java.io.ByteArrayInputStream;
 import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.RandomAccessFile;
 import java.nio.ByteBuffer;
 import java.nio.ReadOnlyBufferException;
 import java.nio.channels.FileChannel;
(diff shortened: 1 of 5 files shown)


buffer/src/test/java/io/netty/buffer/ReadOnlyDirectByteBufferBufTest.java
```diff
@@ -293,12 +292,12 @@ public void testWrapMemoryMapped() throws Exception {
         ByteBuf b2 = null;
 
         try {
-            output = new FileOutputStream(file).getChannel();
+            output = new RandomAccessFile(file, "rw").getChannel();
```
ejona86: This looks like it leaks, since you need to close `RandomAccessFile`. The documentation for `getChannel()` does not say closing the channel closes the `RandomAccessFile`. The documentation for `RandomAccessFile.close()` notes the opposite, that closing the file closes the channel. Looking at the source, it seems the channel holds an extra ref, so closing it does not close the `fd` held by `RandomAccessFile`.

normanmaurer: @ejona86 maybe I am missing something but we call `output.close()` in the finally block so it should be fine ?

ejona86: At risk of repeating myself: `output` is the _channel_. But you need to close the `RandomAccessFile`. It is subtle.

normanmaurer: @ejona86 sorry I missed this... This is not correct, closing the `FileChannel` here will also close the `RandomAccessFile` from which it was created. See also http://hg.openjdk.java.net/jdk8u/jdk8u-dev/jdk/file/4a1e42601d61/src/share/classes/sun/nio/ch/FileChannelImpl.java#l138

ejona86: Hmm... the decrement for the ref is also in RandomAccessFile, so the channel _must_ close the file. Okay. Very well. I continue to be amazed at how poor some documentation is to define resource management...

PR #8111, opened 2018-07-09
Do not log explicit no unsafe, fixes helper method.

seut:
Motivation:

Users should not see a scary log message when Netty is initialized if
Netty configuration explicitly disables unsafe. The log message that
produces this warning was previously guarded but by recent refactoring
a bug was introduced inside the guard helper method.

Modifications:

This commit brings back the guard against the scary log message if
unsafe is explicitly disabled.

Result:

No log message is produced when unsafe is unavailable because Netty was
told to not look for it.

Relates #5624, #6696

### DIFF
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent0.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent0.java
@@ -368,7 +368,7 @@ final class PlatformDependent0 {
     }
 
     static boolean isExplicitNoUnsafe() {
-        return EXPLICIT_NO_UNSAFE_CAUSE == null;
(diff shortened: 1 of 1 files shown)


seut:
@jasontedor I hope you don't mind that I've re-used your PR/commit message ;)

jasontedor:
@seut I don’t mind at all, thank you for fixing this. I had noticed a few weeks ago this was broken again and haven’t taken the time yet to submit a PR so I’m super thankful that you did! ❤️

normanmaurer:
@seut doh... thanks a lot !

seut:
@normanmaurer thanks.

PR #8113, opened 2018-07-10
Use Java11+ea28 during build.

normanmaurer:
Motivation:

We should ensure we use the latest Java11 EA during build to catch any regressions etc.

Modifications:

Update from ea19 to ea28.

Result:

Use latest Java11 release.

### DIFF
--- a/docker/docker-compose.centos-6.111.yaml
+++ b/docker/docker-compose.centos-6.111.yaml
@@ -7,7 +7,7 @@ services:
     build:
       args:
         centos_version : "6"
-        java_version : "1.11.0-19"
+        java_version : "1.11.0-28"
 
   test:
     image: netty:centos-6-1.11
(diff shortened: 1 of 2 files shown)


normanmaurer:
Don't merge this yet... seems like there is a "regression" in the latest EA release related to ssl.

normanmaurer:
Here we go http://mail.openjdk.java.net/pipermail/net-dev/2018-July/011624.html ... 

/cc @ejona86 as this most likely also affects conscrypt.

normanmaurer:
Used the wrong mailing list... The follow at http://mail.openjdk.java.net/pipermail/security-dev/2018-July/017536.html

ejona86:
@normanmaurer, thanks for the heads-up.

normanmaurer:
Bug-report is open now:

https://bugs.openjdk.java.net/browse/JDK-8207029

normanmaurer:
While they closed the mentioned issue as part of ea23 they unfortunately missed two casts: https://bugs.openjdk.java.net/browse/JDK-8208166.

normanmaurer:
Java11 EA28 pass out tests again!

PR #8123, opened 2018-07-11
Add integration test for shading native libraries.

normanmaurer:
Motivation:

It's easy to break the support for shading native libs as shown in https://github.com/netty/netty/issues/8090. We should have some testing to ensure all works as expected.

Modification:

Add new testsuite which verifies that shading our native transports work as expected.

Result:

Include test to verify shading of native code.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -273,6 +273,7 @@
     <module>testsuite-autobahn</module>
     <module>testsuite-http2</module>
     <module>testsuite-osgi</module>
+    <module>testsuite-shading</module>
     <module>microbench</module>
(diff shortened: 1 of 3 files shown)


normanmaurer:
Related to https://github.com/netty/netty/issues/8090. 

/cc @cchantep

normanmaurer:
@ejona86 don't like maven xml hell =P ?

ejona86:
@normanmaurer, I overheard a discussion. It was like, "_something something_ it makes Java look terse _something_." Hard to say what it was about.

Issue #8122, opened 2018-07-11
Only apply copyMemoy safepoint for java <= 8


hc-codersatlas:
Hi, as per https://bugs.openjdk.java.net/browse/JDK-8149596 (and the source code of Java.nio.Bits) safepoint polling is no longer required (or rather it shouldn't be done). It's included as part of Unsafe.copyMemory.

Can we add a check to PlatformDependent where this occurs for java version <= 8 so we don't essentially do safepoint polling twice?

Thanks.

normanmaurer:
sounds good.

hc-codersatlas:
Thanks for the PR. Looks good.

johnou:
b109, should we be checking build too?

normanmaurer:
@johnou honestly I think we either should either just assume its ok if people use java8 (and if its not they should really upgrade) or only not use it for java9 and higher. /cc @hc-codersatlas WDYT ?

hc-codersatlas:
@normanmaurer agree

PR #8124, opened 2018-07-11
Only use manual safepoint polling in PlatformDependent0.copyMemory(..…

normanmaurer:
….) when using java <= 8

Motivation:

Java9 and later does the safepoint polling by itself so there is not need for us to do it.

Modifications:

Check for java version before doing manual safepoint polling.

Result:

Less custom code and less overhead when using java9 and later. Fixes https://github.com/netty/netty/issues/8122.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent0.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent0.java
@@ -550,7 +550,16 @@ final class PlatformDependent0 {
     }
 
     static void copyMemory(long srcAddr, long dstAddr, long length) {
-        //UNSAFE.copyMemory(srcAddr, dstAddr, length);
+        // Manual safe-point polling is only needed prior Java9:
+        // See https://bugs.openjdk.java.net/browse/JDK-8149596
+        if (javaVersion() <= 8) {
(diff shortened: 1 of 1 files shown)


normanmaurer:
@hc-codersatlas FYI

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -550,24 +550,36 @@ static void putLong(byte[] data, int index, long value) {
...
-        while (length > 0) {
-            long size = Math.min(length, UNSAFE_COPY_THRESHOLD);
-            UNSAFE.copyMemory(src, srcOffset, dst, dstOffset, size);
-            length -= size;
-            srcOffset += size;
-            dstOffset += size;
+        // Manual safe-point polling is only needed prior Java9:
+        // See https://bugs.openjdk.java.net/browse/JDK-8149596
+        if (javaVersion() <= 8) {
+            //UNSAFE.copyMemory(src, srcOffset, dst, dstOffset, length);
```
bryce-anderson: Not a part of this commit, but is this commented out code necessary? Same above.

PR #8126, opened 2018-07-11
Fix possible NPE introduced by a137291ad161c970259d9a5d7cc206c026f08e…

normanmaurer:
…e7 when using SslProvider.OPENSSL and init via files or OpenSslX509KeyManagerFactory

Motivation:

a137291ad161c970259d9a5d7cc206c026f08ee7 introduced a way to get the most speed out of OpenSSL by not only caching keymaterial but pre-compute these. The problem was we missed to check for null before doing an instanceof check and then a cast which could lead to a NPE as we tried to cast null to Exception and throw it.

Modifications:

Add null check and unit test.

Result:

No more NPE when keymaterial was not found for requested alias.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/OpenSslX509KeyManagerFactory.java
+++ b/handler/src/main/java/io/netty/handler/ssl/OpenSslX509KeyManagerFactory.java
@@ -198,11 +198,14 @@ public final class OpenSslX509KeyManagerFactory extends KeyManagerFactory {
                 @Override
                 OpenSslKeyMaterial chooseKeyMaterial(ByteBufAllocator allocator, String alias) throws Exception {
                     Object value = materialMap.get(alias);
+                    if (value == null) {
+                        // There is no keymaterial for the requested alias, return null
+                        return null;
+                    }
                     if (value instanceof OpenSslKeyMaterial) {
                         return ((OpenSslKeyMaterial) value).retain();
-                    } else {
-                        throw (Exception) value;
                     }
+                    throw (Exception) value;
                 }
 
                 @Override
--- a/handler/src/test/java/io/netty/handler/ssl/OpenSslX509KeyManagerFactoryProviderTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/OpenSslX509KeyManagerFactoryProviderTest.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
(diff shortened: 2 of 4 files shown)


handler/src/main/java/io/netty/handler/ssl/OpenSslX509KeyManagerFactory.java
```diff
@@ -158,61 +158,65 @@ OpenSslKeyMaterialProvider newProvider() {
...
-                        initComplete = true;
-                    } finally {
-                        if (!initComplete) {
-                            destroy();
+    /**
+     * {@link OpenSslKeyMaterialProvider} implementation that pre-compute the {@link OpenSslKeyMaterial} for
+     * all aliases.
+     */
+    // Only package-private for testing.
+    static final class OpenSslPopulatedKeyMaterialProvider extends OpenSslKeyMaterialProvider {
```
normanmaurer: all the changes here are just because I moved it to the outer-class and made it package-private for testing.

handler/src/main/java/io/netty/handler/ssl/OpenSslX509KeyManagerFactory.java
```diff
@@ -158,61 +158,65 @@ OpenSslKeyMaterialProvider newProvider() {
...
-                void destroy() {
-                    for (Object material: materialMap.values()) {
-                        ReferenceCountUtil.release(material);
-                    }
-                    materialMap.clear();
-                }
+        @Override
+        OpenSslKeyMaterial chooseKeyMaterial(ByteBufAllocator allocator, String alias) throws Exception {
+            Object value = materialMap.get(alias);
+            if (value == null) {
```
normanmaurer: This is the actual fix... We missed the null check.

handler/src/test/java/io/netty/handler/ssl/OpenSslX509KeyManagerFactoryProviderTest.java
```diff
@@ -0,0 +1,38 @@
...
+        keystore.load(getClass().getResourceAsStream("mutual_auth_server.p12"), password);
+
+        OpenSslX509KeyManagerFactory kmf = new OpenSslX509KeyManagerFactory();
+        kmf.init(keystore, password);
+        return kmf;
+    }
+
+    @Override
+    protected OpenSslKeyMaterialProvider newMaterialProvider(KeyManagerFactory kmf, String password) {
+        return ((OpenSslX509KeyManagerFactory) kmf).newProvider();
```
Scottmitch: We could make the base class generic on the typically e of kmf to avoid casting here.

normanmaurer: I think its fine as it is actual as the base-class is not abstract

normanmaurer:
I will merge this one once the build pass

PR #8147, opened 2018-07-24
HTTP/2 child channel read cycle doesn't respect RecvByteBufAllocator and

Scottmitch:
may drop data if stream closed while auto read is off

Motivation:
Http2MultiplexCodec queues data internally if data is delivered from the
parent channel but the child channel did not request data. If the parent
channel notifies of a stream closure it is possible data in the queue
will be discarded before closing the channel.
Http2MultiplexCodec interacts with RecvByteBufAllocator to control the
child channel's demand for read. However it currently only ever reads a
maximum of one time per loop. This can thrash the read loop and bloat
the call stack if auto read is on, because channelReadComplete will
re-enter the read loop synchronously, and also neglect to deliver data
during the parent's read loop (if it is active). This also meant the
readPendingQueue was not utilized as originally intended (to extend the
child channel's read loop during the parent channel's read loop if
demand for data still existed).

Modifications:
- Modify the child channel's read loop to respect the
RecvByteBufAllocator, and append to the parents readPendingQueue if
appropriate.
- Stream closure notification behaves like EPOLL and KQUEUE transports
and reads all queued data, because the data is already queued in memory
and it is known there will be no more data. This will also replenish the
connection flow control window which may otherwise be constrained by a
closed stream.

Result:
More correct read loop and less risk of dropping data.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
@@ -31,10 +31,10 @@ import io.netty.channel.ChannelProgressivePromise;
 import io.netty.channel.ChannelPromise;
 import io.netty.channel.DefaultChannelConfig;
 import io.netty.channel.DefaultChannelPipeline;
-import io.netty.channel.DefaultMaxMessagesRecvByteBufAllocator;
 import io.netty.channel.EventLoop;
 import io.netty.channel.MessageSizeEstimator;
 import io.netty.channel.RecvByteBufAllocator;
+import io.netty.channel.RecvByteBufAllocator.Handle;
 import io.netty.channel.VoidChannelPromise;
 import io.netty.channel.WriteBufferWaterMark;
 import io.netty.util.DefaultAttributeMap;
@@ -56,7 +56,6 @@ import static io.netty.handler.codec.http2.Http2CodecUtil.HTTP_UPGRADE_STREAM_ID
 import static io.netty.handler.codec.http2.Http2CodecUtil.isStreamIdValid;
 import static io.netty.handler.codec.http2.Http2Error.INTERNAL_ERROR;
 import static io.netty.handler.codec.http2.Http2Exception.connectionError;
-
 import static java.lang.Math.min;
 
 /**
@@ -111,7 +110,7 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
 
     private static final ChannelFutureListener CHILD_CHANNEL_REGISTRATION_LISTENER = new ChannelFutureListener() {
         @Override
-        public void operationComplete(ChannelFuture future) throws Exception {
+        public void operationComplete(ChannelFuture future) {
             registerDone(future);
         }
     };
@@ -145,19 +144,6 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
         @Override
         public Handle newHandle() {
             return HANDLE_INSTANCE;
-        }
-    }
-
-    private static final class Http2StreamChannelRecvByteBufAllocator extends DefaultMaxMessagesRecvByteBufAllocator {
-
-        @Override
-        public MaxMessageHandle newHandle() {
-            return new MaxMessageHandle() {
-                @Override
-                public int guess() {
-                    return 1024;
-                }
-            };
         }
     }
 
@@ -230,7 +216,7 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
         while (ch != null) {
             DefaultHttp2StreamChannel curr = ch;
             ch = curr.next;
-            curr.next = null;
+            curr.next = curr.previous = null;
         }
         head = tail = null;
     }
@@ -244,7 +230,7 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
     final void onHttp2Frame(ChannelHandlerContext ctx, Http2Frame frame) {
         if (frame instanceof Http2StreamFrame) {
             Http2StreamFrame streamFrame = (Http2StreamFrame) frame;
-            onHttp2StreamFrame(((Http2MultiplexCodecStream) streamFrame.stream()).channel, streamFrame);
+            ((Http2MultiplexCodecStream) streamFrame.stream()).channel.fireChildRead(streamFrame);
         } else if (frame instanceof Http2GoAwayFrame) {
             onHttp2GoAwayFrame(ctx, (Http2GoAwayFrame) frame);
             // Allow other handlers to act on GOAWAY frame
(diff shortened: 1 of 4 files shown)


codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -230,7 +216,7 @@ public final void handlerRemoved0(ChannelHandlerContext ctx) throws Exception {
         while (ch != null) {
             DefaultHttp2StreamChannel curr = ch;
             ch = curr.next;
-            curr.next = null;
+            curr.next = curr.previous = null;
```
normanmaurer: @Scottmitch good one... Is this to allow the GC to kick in earlier or to eliminate GC Nepotism effects ?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -331,38 +317,48 @@ final void onHttp2FrameStreamException(ChannelHandlerContext ctx, Http2FrameStre
...
-            case READ_PROCESSED_OK_TO_PROCESS_MORE:
-                addChildChannelToReadPendingQueue(childChannel);
-                break;
-            case READ_IGNORED_CHANNEL_INACTIVE:
-            case READ_QUEUED:
-                // nothing to do:
-                break;
-            default:
-                throw new Error();
+    private boolean childChannelInReadPendingQueue(DefaultHttp2StreamChannel childChannel) {
```
normanmaurer: nit: can we prefix this with `is` ?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -1197,18 +1185,16 @@ private ChannelFuture write0(Object msg) {
 
             @Override
             public void flush() {
-                if (!writeDoneAndNoFlush) {
+                // If we are currently in the parent channel's read loop we should just ignore the flush.
+                // We will ensure we trigger ctx.flush() after we processed all Channels later on and
+                // so aggregate the flushes. This is done as ctx.flush() is expensive when as it may trigger an
+                // write(...) or writev(...) operation on the socket.
+                if (!writeDoneAndNoFlush || parentReadInProgress) {
```
normanmaurer: @Scottmitch I wonder if we should always flush if `channel.isWritable()` returns false here to try to make room again... Something like:

```java 
if (!writeDoneAndNoFlush || (parentReadInProgress && channel.isWritable()) {
}
```

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
```diff
@@ -743,9 +743,197 @@ public void callUnsafeCloseMultipleTimes() {
...
+        assertEquals(dataFrame3, inboundHandler.readInbound());
+        assertEquals(dataFrame4, inboundHandler.readInbound());
+        assertNull(inboundHandler.readInbound());
+
+        // Now we want to call channelReadComplete and simulate the end of the read loop.
+        parentChannel.flushInbound();
+
+        childChannel.closeFuture().syncUninterruptibly();
+
+        dataFrame1.release();
```
normanmaurer: nit: should we just release the frames directly after we do not need these anymore ?

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
```diff
@@ -743,9 +743,197 @@ public void callUnsafeCloseMultipleTimes() {
...
+        // Now we want to call channelReadComplete and simulate the end of the read loop.
+        parentChannel.flushInbound();
+
+        // 3 = 1 for initialization + 1 for read when auto read was off + 1 for when auto read was back on
+        assertEquals(3, channelReadCompleteCount.get());
+
+        dataFrame1.release();
+        dataFrame2.release();
+        dataFrame3.release();
+        dataFrame4.release();
```
normanmaurer: should we release the frames directly after we do not need these anymore ?

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
```diff
@@ -743,9 +743,197 @@ public void callUnsafeCloseMultipleTimes() {
...
+        parentChannel.flushInbound();
+
+        // 3 = 1 for initialization + 1 for first read of 2 items + 1 for second read of 2 items +
+        // 1 for parent channel readComplete
+        assertEquals(4, channelReadCompleteCount.get());
+
+        dataFrame1.release();
+        dataFrame2.release();
+        dataFrame3.release();
+        dataFrame4.release();
```
normanmaurer: see above

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -476,24 +461,22 @@ private boolean initialWritability(DefaultHttp2FrameStream stream) {
...
-        /** {@code true} if a close without an error was initiated **/
-        private boolean streamClosedWithoutError;
-
-        // Keeps track of flush calls in channelReadComplete(...) and aggregate these.
-        private boolean inFireChannelReadComplete;
-
-        boolean fireChannelReadPending;
-
         // Holds the reference to the next DefaultHttp2StreamChannel that should be processed in
         // channelReadComplete(...)
```
normanmaurer: nit: consider updating the java doc to reflect its now a double-linked list.

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
Scottmitch: I don't think it matter from a test perspective. This approach was just being consistent with the existing tests in this class. If we really wanted to be strict we would have to complicate the test logic a bit more ... but I thought we came to the conclusion that if the test passes this shouldn't matter, and if a test fails we are OK with leaks because something is broken anyways.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
Scottmitch: Perhaps, but I think we need a more comprehensive review of writability for the child channel implementation is warranted https://github.com/netty/netty/issues/8148.

For example the child channel writes will not actually be written to the parent channel until the parent channel's `flush()` is called. This is when flow control is done. So I wouldn't expect `childChannel.parent().isWritable()` to be impacted by this operation.

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
normanmaurer: Fair enough... can you open a follow-up issue ?

normanmaurer: sounds good

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
normanmaurer: ok thanks :)

normanmaurer:
@ejona86 @bryce-anderson want to take a look as well ?

normanmaurer:
@Scottmitch please ping us once you addressed the feedback

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java
```diff
@@ -743,9 +743,197 @@ public void callUnsafeCloseMultipleTimes() {
...
+        Http2DataFrame dataFrame2 = new DefaultHttp2DataFrame(bb("2")).stream(inboundStream);
+        Http2DataFrame dataFrame3 = new DefaultHttp2DataFrame(bb("3")).stream(inboundStream);
+        Http2DataFrame dataFrame4 = new DefaultHttp2DataFrame(bb("4")).stream(inboundStream);
+
+        assertEquals(new DefaultHttp2HeadersFrame(request).stream(inboundStream), inboundHandler.readInbound());
+
+        // We want to simulate the parent channel calling channelRead and delay calling channelReadComplete.
+        parentChannel.writeOneInbound(new Object());
+        codec.onHttp2Frame(dataFrame1);
+        assertEquals(dataFrame1, inboundHandler.readInbound());
```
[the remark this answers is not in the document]
Scottmitch: Agreed it surprised me too. The channel initialization involved with the EmbeddedChannel and mocking makes it so read is enabled by default, a read is done for the initial headers, and auto read is enabled at that point so another read is done. Since you can't "un-read" a data frame slips through. Ideally I would like to move away from the existing test setup because the mocking is actually changing expected behavior ... but I think that outside the scope of this PR.

rschmitt:
@Scottmitch On previous versions of Netty, I noticed a problem with the `FlowControlHandler` on HTTP/2 streams with `Http2MultiplexCodec` and auto read set to off. The sequence of events I observed was:

1. A handler downstream of the `FlowControlHandler` would call `read`
2. `FlowControlHandler` would look in its queue, find it empty, and call `read` upstream
3. The `Http2MultiplexCodec` would fire `channelReadComplete` without first sending any data downstream
4. The `FlowControlHandler` would [discard this event](https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/flow/FlowControlHandler.java#L155-L160), and the channel would stall

Is this one of the issues fixed by this commit?

Issue #8144, opened 2018-07-23
Using underscore character('_') as part of SNI leads JVM crash


r9liucc:
Hello guys, we use Netty as our network implementation of microservices framework. We plan to support https and thus enable authentication on both client side and server side. But we encounter a problem when we want to use underscore character('_') in TLS SNI.

#### 1. Netty Server implementation:
```
// define ssl provider
private static final SslProvider SSL_PROVIDER = SslProvider.OPENSSL;

// factory method for creating ssl context
public SslContext serverCtx(InputStream certificate, InputStream privateKey) {
        try {
            return SslContextBuilder.forServer(certificate, privateKey)
                                    .sslProvider(SSL_PROVIDER)
                                    .trustManager(TwoWayTrustManagerFactory.INSTANCE)
                                    .clientAuth(ClientAuth.REQUIRE)
                                    .build();
        } catch (Exception e) {
            throw new RuntimeException("failed to create ssl context", e);
        }
}

// SNI Matcher implementation
@Slf4j
public class MySNIMatcher extends SNIMatcher {

    private final byte[] hostIdentifier;

    public MySNIMatcher(Identifier hostIdentifier) {
        super(0);
        this.hostIdentifier = hostIdentifier.idDistKey().getBytes(StandardCharsets.US_ASCII);
    }

    @Override
    public boolean matches(SNIServerName sniServerName) {
        if (log.isDebugEnabled()) {
            log.debug("match sni {}", new String(sniServerName.getEncoded(), StandardCharsets.US_ASCII));
        }
        return Arrays.equals(sniServerName.getEncoded(), hostIdentifier);
    }
}

// Netty bootstrap
bootstrap.childHandler(new ChannelInitializer<Channel>() {
            @Override
            protected void initChannel(Channel ch) throws Exception {
                ChannelPipeline pipeline = ch.pipeline();

                if (conf.logEnable())
                    pipeline.addLast("logging", new LoggingHandler());

                if (conf.sslEnable()) {
                    // conf.sslCtx() will actually use serverCtx() to create context
                    SSLEngine engine = conf.sslCtx().newEngine(ch.alloc());
                    SSLParameters s = engine.getSSLParameters();
                    s.setSNIMatchers(new MySNIMatcher(...));
                    s.setEndpointIdentificationAlgorithm("HTTPS");
                    engine.setSSLParameters(s);
                    pipeline.addLast("ssl", new SslHandler(engine, true));
                }

                codec(pipeline);

[...]

PR #8150, opened 2018-07-25
Allow to validate sni hostname with underscore

normanmaurer:
Motivation:

We should allow to also validate sni hostname which contains for example underscore when using our native SSL impl. The JDK implementation does this as well.

Modifications:

- Construct the SNIHostName via byte[] and not String.
- Add unit test

Result:

Fixes https://github.com/netty/netty/issues/8144.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslServerContext.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslServerContext.java
@@ -19,6 +19,7 @@ import io.netty.buffer.ByteBufAllocator;
 import io.netty.internal.tcnative.SSL;
 import io.netty.internal.tcnative.SSLContext;
 import io.netty.internal.tcnative.SniHostNameMatcher;
+import io.netty.util.CharsetUtil;
 import io.netty.util.internal.PlatformDependent;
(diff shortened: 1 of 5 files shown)


normanmaurer:
@ejona86 because of how `SNIHostName` constructor is implemented.

normanmaurer:
@ejona86 yep exactly...

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslServerContext.java
```diff
@@ -244,7 +245,8 @@ void verify(ReferenceCountedOpenSslEngine engine, X509Certificate[] peerCerts, S
         public boolean match(long ssl, String hostname) {
             ReferenceCountedOpenSslEngine engine = engineMap.get(ssl);
             if (engine != null) {
-                return engine.checkSniHostnameMatch(hostname);
+                // TODO: In the next release of tcnative we should pass the byte[] directly in and not use a String.
```
Scottmitch: Is there an issue on tcnarive for this?

normanmaurer: Not yet... will open one

PR #8156, opened 2018-07-27
RFC: Add additional insertion methods to ChannelPipeline

Bennett-Lynch:
Motivation:

Not all users add ChannelHandlers to a pipeline by a specific name, and
the current "addBefore and "addAfter" methods only support referencing
the base handler by name. This imposes the unnecessary requirement for
users to add their handlers by name, when they could instead reference
the base handler by object reference or class type. This is arguably
cleaner and more concise in many cases.

Modifications:

* Create "addBefore" and "addAfter" methods that reference the base
handler by either object reference or class type.
* Create overloaded versions that allow the user to specify the name of
the new handler
* Create overloaded "replace" methods that do not require the user to
specify the name of the new handler

I have not (yet) created overloaded versions that take an
EventExecutorGroup, as this this would double the number of overloads
needed to support the new parameter combinations.

I have also not yet attempted to implement these new interface methods,
or add unit tests, as this is just an RFC/proposal.

### DIFF
--- a/transport/src/main/java/io/netty/channel/ChannelPipeline.java
+++ b/transport/src/main/java/io/netty/channel/ChannelPipeline.java
@@ -309,6 +309,79 @@ public interface ChannelPipeline
     ChannelPipeline addBefore(EventExecutorGroup group, String baseName, String name, ChannelHandler handler);
 
     /**
+     * Inserts a {@link ChannelHandler} before the specified {@link ChannelHandler} of this pipeline.
+     *
+     * @param  baseHandler   the existing {@link ChannelHandler}
+     * @param  newHandler    the {@link ChannelHandler} to insert before
+     *
+     * @return itself
+
+     * @throws NoSuchElementException
+     *         if the specified base handler does not exist in this pipeline
+     * @throws NullPointerException
+     *         if the specified old handler or new handler is
+     *         {@code null}
+     */
+    ChannelPipeline addBefore(ChannelHandler baseHandler, ChannelHandler newHandler);
+
+    /**
+     * Inserts a {@link ChannelHandler} before the specified {@link ChannelHandler} of this pipeline.
+     *
+     * @param  baseHandler   the existing {@link ChannelHandler}
+     * @param  newName       the name of the {@link ChannelHandler} to insert before
+     * @param  newHandler    the {@link ChannelHandler} to insert before
+     *
+     * @return itself
+
+     * @throws NoSuchElementException
+     *         if the specified base handler does not exist in this pipeline
+     * @throws IllegalArgumentException
+     *         if a handler with the specified new name already exists in this
+     *         pipeline, except for the handler to be replaced
+     * @throws NullPointerException
+     *         if the specified old handler or new handler is
+     *         {@code null}
+     */
+    ChannelPipeline addBefore(ChannelHandler baseHandler, String newName, ChannelHandler newHandler);
+    
+    /**
+     * Inserts a {@link ChannelHandler} before the {@link ChannelHandler} of the specified type in this pipeline.
+     *
+     * @param  baseHandlerType   the type of the existing {@link ChannelHandler}
+     * @param  newHandler        the {@link ChannelHandler} to insert before
+     *
+     * @return itself
+
+     * @throws NoSuchElementException
+     *         if the specified base handler does not exist in this pipeline
+     * @throws NullPointerException
+     *         if the specified old handler or new handler is
+     *         {@code null}
+     */
+    <T extends ChannelHandler> ChannelPipeline addBefore(Class<T> baseHandlerType, ChannelHandler newHandler);
+
+    /**
+     * Inserts a {@link ChannelHandler} before the {@link ChannelHandler} of the specified type in this pipeline.
+     *
+     * @param  baseHandlerType   the type of the existing {@link ChannelHandler}
+     * @param  newName           the name of the {@link ChannelHandler} to insert before
+     * @param  newHandler        the {@link ChannelHandler} to insert before
+     *
+     * @return itself
+
+     * @throws NoSuchElementException
+     *         if the specified base handler does not exist in this pipeline
+     * @throws IllegalArgumentException
+     *         if a handler with the specified new name already exists in this
+     *         pipeline, except for the handler to be replaced
+     * @throws NullPointerException
+     *         if the specified old handler or new handler is
+     *         {@code null}
+     */
+    <T extends ChannelHandler> ChannelPipeline addBefore(Class<T> baseHandlerType, String newName,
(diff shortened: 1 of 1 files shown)


Bennett-Lynch:
@normanmaurer Any thoughts?

normanmaurer:
@Bennett-Lynch sorry I am super busy atm... Will come back to you soon

Bennett-Lynch:
@normanmaurer Friendly reminder/ping. Could we poll a few people to get a feel for the general consensus regarding this change?

normanmaurer:
This sounds good to me but we will need to wait for next major version... @ejona86 @carl-mastrangelo thoughts ?

ejona86:
+1 for something along these lines. You can do what you need to today by calling `context(ChannelHandler|Class).name()`, but basing it off names always seemed weird to me.

Instead of having so many overloads, would it be better to use just `ChannelHandlerContext`? That would be slightly strange because the `ctx` isn't really useful to anyone but the Handler, but 1) often a handler is wanting to modify the pipeline relative to itself (so it wouldn't be strange), 2) it would avoid the pipeline from having to double-search, and 3) it avoids having tons of overloads.

Bennett-Lynch:
@ejona86 I've used workarounds similar to the `Class.name()` approach, but it's a bit awkward, since Netty defaults to a slightly different implementation: https://github.com/netty/netty/blob/00afb19d7a37de21b35ce4f6cb3fa7f74809f2ab/transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java#L440 ...requiring the user to manually specify *all* handler names (or hackily duplicate Netty's default naming logic).

Using `ChannelHandlerContext` would be an interesting way to reduce the overloads... but it'd come at the cost of readability, I think:

`ctx.pipeline().replace(Object.class, new Handler());`
vs.
`ctx.pipeline().replace(ctx.pipeline().context(Object.class), new Handler());`

The current `ChannelPipeline` implementations that take in a reference or class don't require the pipeline to double-search, do they?

The large number of overloads is my only concern with this approach, but I personally would favor them over the readability cost in having the user to manually fetch the `ChannelHandlerContext`.  IMO, the overloads only really reach a concerning number when you start factoring in support for `EventExecutorGroup`... could we reasonably just omit support for it, or offer the longer `ChannelHandlerContext` form for it?  Or maybe it's totally fine to offer that many overloads.

@normanmaurer Is there any reason this would require a new major version?  It's an API change, but it should be fully backwards compatible.

transport/src/main/java/io/netty/channel/ChannelPipeline.java
```diff
@@ -344,6 +417,79 @@
...
+     *         if the specified old handler or new handler is
+     *         {@code null}
+     */
+    ChannelPipeline addAfter(ChannelHandler baseHandler, ChannelHandler newHandler);
+
+    /**
+     * Inserts a {@link ChannelHandler} after the specified {@link ChannelHandler} of this pipeline.
+     *
+     * @param  baseHandler   the existing {@link ChannelHandler}
+     * @param  newName       the name of the {@link ChannelHandler} to insert after
```
ejona86: This language is wrong. It is "the name of the {@link ChannelHandler} to _be inserted_"

ejona86:
> I've used workarounds similar to the Class.name() approach,

No, not `Class.getName()` but `ChannelHandlerContext.name()`.

> but it'd come at the cost of readability

I wouldn't necessarily say "readability" but would go with "terseness." We shouldn't forget having 6 overrides has its own cost for readability. Also, I'm more comparing against the existing API instead of your proposed API. Both APIs would be more readable than what we have.

I forgot that `addBefore` and the like don't return `ChannelHandlerContext`. That makes the gains more limited.

Still, I find [places like this](https://github.com/netty/netty/blob/00afb19d7a37de21b35ce4f6cb3fa7f74809f2ab/handler-proxy/src/main/java/io/netty/handler/proxy/HttpProxyHandler.java#L130) fairly common. Those would still be natural.

The more I look at the existing API, the more ~random~ naturally evolved it looks:
1. `get` supports Class, String
2. `context`/`remove`/`replace` supports ChannelHandler, Class, String
3. `addAfter`/`addBefore` supports String

It is unfortunate we have a Cartesian product between "operation" x "way to identify old handler" x "whether to name new handler". The "whether to name new handler" is not included for cases (1) and (2). Without the "whether to name new handler" I think it'd be pretty obvious to support ChannelHandler and Class.

It would be possible to get rid of the "whether to name new handler" axis, but passing `null` to these methods would likely be a readability issue of its own.

> The current ChannelPipeline implementations that take in a reference or class don't require the pipeline to double-search, do they?

They always require a (single) linear search. Depending on the implementation, they may double-search.  If you look up the `ChannelHandlerContext` to learn the handler's `name`, then it is a search. So today it requires a double-search if you don't know the `name` a priori. If the argument is a `ctx`, then it is possible for zero searches to be necessary in certain cases.

> Is there any reason this would require a new major version? It's an API change, but it should be fully backwards compatible.

Modifying interfaces is not backward compatible, as any existing implementations are not guaranteed to support the new methods.

Bennett-Lynch:
@ejona86 Agree with you on all points. And you can unfortunately add whether or not to use `EventExecutorGroup` to that Cartesian product.

Also fair call out on the fact that this is an interface (kind of an unfortunate restriction since I don't think there's any real use case where users are defining their own implementations).  We could potentially make these changes just to `DefaultChannelPipeline` and save the interface change for a future major version?

All things considered (in terms of the trade offs in overloads vs. terseness, etc.), do you have a final opinion on the ideal API to expose?

RoganDawes:
I also have some requests relating to the ChannelPipeline adding and removing of handlers.

In my code (Mallet), I often do something like:
```
        @Override
        protected void initChannel(Channel ch) throws Exception {
                String name = ch.pipeline().context(this).name();
                ch.pipeline().addAfter(name, null, getChannelHandler());
        }
```
This seems pretty clumsy to me, firstly having to look up the context from the Channel (although that is probably a separate issue relating to ChannelInitializer), then getting the name() from ChannelHandlerContext, then finally iterating through the pipeline a second time to add the new ChannelHandler after the current name().

If the pipeline supported addAfter(ChannelHandlerContext ctx, String name, ChannelHandler... handlers), etc, I think this should be more efficient, and concise.

Bennett-Lynch:
@normanmaurer Can we close this out and convert it into a [Netty 5] issue?

normanmaurer:
Move this as discussion for netty 5.

njhill:
FYI @Bennett-Lynch @RoganDawes I've added another suggestion of how this could be done in netty 5  in other issue https://github.com/netty/netty/issues/8511#issuecomment-459195152

normanmaurer:
I think we can close this for now

PR #8160, opened 2018-07-29
Lazily initialize NativeDatagramPacketArray and IovArray in EpollEventLoop

### LINKED ISSUE #8159: Significant epoll native memory usage jump in 4.1.26.Final
### Expected behavior

Native memory use is stable between releases.

### Actual behavior

Significant jump in native memory usage when using epoll transport between 4.1.25.Final and 4.1.26.Final.

### Steps to reproduce

Observe process native memory usage when using `EpollEventLoopGroup`(s) between the above versions. Usage is comparable between NIO and epoll in version 4.1.25.Final, and also between 4.1.25.Final and 4.1.26.Final when using NIO.

This is particularly significant when using the default ELG size (cores *2) on large SMP machines, and/or when there are multiple ELGs in the JVM due to use of netty in different libraries. In this case the difference can run into hundreds of MB (despite the heap being only 512M), and easily breach cgroup mem limits.

I assume this is caused by https://github.com/netty/netty/pull/8062. Also in light of that change it appears this comment is no longer accurate: https://github.com/netty/netty/blob/d20370dd5f15c7e08c252e0b818c91113d7bea94/transport-native-epoll/src/main/java/io/netty/channel/epoll/NativeDatagramPacketArray.java#L103

### Netty version

4.1.26.Final, 4.1.27.Final, 4.1.28.Final

```
java version "1.8.0_171"
Java(TM) SE Runtime Environment (build 1.8.0_171-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.171-b11, mixed mode)
```

```
4.4.0-127-generic #153-Ubuntu SMP x86_64 GNU/Linux
```

njhill:
Motivation:

Avoid unnecessary native memory allocation if UDP / TCP isn't being used.

Modifications:

Create the reused `NativeDatagramPacketArray` and `IovArray` upon first use instead of `EpollEventLoop` construction. This more closely matches the prior `FastThreadLocal.initialValue()` behaviour changed in #8062.

Also correct related comment in `NativeDatagramPacketArray`.

Result:

Reduced native memory use in many cases when using epoll. Fixes #8159

### DIFF
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
@@ -61,8 +61,10 @@ final class EpollEventLoop extends SingleThreadEventLoop {
     private final IntObjectMap<AbstractEpollChannel> channels = new IntObjectHashMap<AbstractEpollChannel>(4096);
     private final boolean allowGrowing;
     private final EpollEventArray events;
-    private final IovArray iovArray = new IovArray();
-    private final NativeDatagramPacketArray datagramPacketArray = new NativeDatagramPacketArray();
+
+    // These are initialized on first use
+    private IovArray iovArray;
+    private NativeDatagramPacketArray datagramPacketArray;
 
     private final SelectStrategy selectStrategy;
     private final IntSupplier selectNowSupplier = new IntSupplier() {
@@ -144,7 +146,11 @@ final class EpollEventLoop extends SingleThreadEventLoop {
      * Return a cleared {@link IovArray} that can be used for writes in this {@link EventLoop}.
      */
     IovArray cleanIovArray() {
-        iovArray.clear();
+        if (iovArray == null) {
+            iovArray = new IovArray();
+        } else {
+            iovArray.clear();
+        }
         return iovArray;
     }
 
@@ -152,7 +158,11 @@ final class EpollEventLoop extends SingleThreadEventLoop {
      * Return a cleared {@link NativeDatagramPacketArray} that can be used for writes in this {@link EventLoop}.
      */
     NativeDatagramPacketArray cleanDatagramPacketArray() {
-        datagramPacketArray.clear();
+        if (datagramPacketArray == null) {
+            datagramPacketArray = new NativeDatagramPacketArray();
+        } else {
+            datagramPacketArray.clear();
+        }
         return datagramPacketArray;
     }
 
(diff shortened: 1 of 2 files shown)


transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
```diff
@@ -458,8 +468,12 @@ protected void cleanup() {
             }
         } finally {
             // release native memory
-            iovArray.release();
-            datagramPacketArray.release();
+            if (iovArray != null) {
+                iovArray.release();
+            }
+            if (datagramPacketArray != null) {
+                datagramPacketArray.release();
```
normanmaurer: can you also null out reference.

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
```diff
@@ -458,8 +468,12 @@ protected void cleanup() {
             }
         } finally {
             // release native memory
-            iovArray.release();
-            datagramPacketArray.release();
+            if (iovArray != null) {
+                iovArray.release();
```
normanmaurer: can you also null out reference.

normanmaurer:
Also please adjust the commit message:

```
Avoid unnecessary native memory allocation if UDP isn't being used.
```

```
Avoid unnecessary native memory allocation if UDP / TCP isn't being used.
```

**Comment by njhill** [comment]
> Avoid unnecessary native memory allocation if UDP / TCP isn't being used.

Sure, I can change this. But if only UDP and not TCP is used then both `datagramPacketArray` and `iovArray` are still used. Is this what you meant? Or this is meant to refer to the case where the ELG is created but not all ELs have channels registered (which will also have memory saving here from  fewer `iovArray`s)?

**Comment by normanmaurer** [comment]
Yeah if you use the group only for accepting new connections you save even more memory. Or if you only use it for UDP but not TCP you also save a bit.

**Comment by njhill** [comment]
Thanks @normanmaurer, I updated the commit with requested wording and ref null-ing.

**Comment by njhill** [comment]
@normanmaurer do you think it would be worthwhile to do the same thing with
```java
channels = new IntObjectHashMap<AbstractEpollChannel>(4096);
```
to save a bit of heap given how large the init capacity is? If so I can open another PR.

johnou:
@normanmaurer this hit us pretty bad, when do you expect to release 4.1.29.Final?

normanmaurer:
@njhill I don't think this is really needed imho as we always need the map anyway (if you will ever register a Channel).

@johnou within the next two weeks.

Issue #8166, opened 2018-08-02
In netty-4.1.28.Final version, an error occurs when running on jdk6.


koo-taejin:
### Actual behavior
```
io.netty.channel.ChannelException: Unable to create Channel from class class io.netty.channel.socket.nio.NioSocketChannel
Caused by: java.lang.NoClassDefFoundError: java/nio/channels/NetworkChannel
	at io.netty.channel.socket.nio.NioSocketChannel.<init>(NioSocketChannel.java:105)
	at io.netty.channel.socket.nio.NioSocketChannel.<init>(NioSocketChannel.java:94)
	at io.netty.channel.socket.nio.NioSocketChannel.<init>(NioSocketChannel.java:87)
	at io.netty.channel.socket.nio.NioSocketChannel.<init>(NioSocketChannel.java:80)
	... 32 more
Caused by: java.lang.ClassNotFoundException: java.nio.channels.NetworkChannel
	at java.net.URLClassLoader$1.run(URLClassLoader.java:202)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:190)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:306)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:301)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:247)
```

### Steps to reproduce
```
// run following test on jvm6

    @Test
    public void listenerTest() throws Exception {
        Bootstrap bootstrap = createBootstrap();
        Channel channel = bootstrap.connect(host, port).sync().channel();

    }

    public Bootstrap createBootstrap() {
        EventLoopGroup workerGroup = new NioEventLoopGroup();
        Bootstrap bootstrap = new Bootstrap();
        bootstrap.group(workerGroup).channel(NioSocketChannel.class)
                .handler(new ChannelInitializer<SocketChannel>() {
                    @Override
                    protected void initChannel(SocketChannel ch) throws Exception {
                        ch.pipeline().addLast(new HttpClientCodec());
                        ch.pipeline().addLast(new HttpObjectAggregator(65535));
                    }
                });
        return bootstrap;
    }
```

### Netty version
netty-4.1.28.Final

### JVM version (e.g. `java -version`)
1.6.0_45

### OS version (e.g. `uname -a`)
Windows10,  CentOS 6.6

I think this issue is problem, but I'm not sure.
If this issue is problem, May I make a correction?

[...]

PR #8168, opened 2018-08-02
Ensure NIO transport can be used on Java6 again.

normanmaurer:
Motivation:

952eeb8e1e3706b09d4d3f32f16d7f0e5c540cb5 introduced the possibility to use any JDK SocketOption when using the NIO transport but broke the possibility to use netty with java6.

Modifications:

Do not use java7 types in method signatures of the static methods in NioChannelOption to prevent class-loader issues on java6.

Result:

Fixes https://github.com/netty/netty/issues/8166.

### DIFF
--- a/transport/src/main/java/io/netty/channel/socket/nio/NioChannelOption.java
+++ b/transport/src/main/java/io/netty/channel/socket/nio/NioChannelOption.java
@@ -19,33 +19,39 @@ import io.netty.channel.ChannelException;
 import io.netty.channel.ChannelOption;
 
 import java.io.IOException;
-import java.net.SocketOption;
-import java.nio.channels.NetworkChannel;
+import java.nio.channels.Channel;
 import java.util.Set;
 
 /**
- * Provides {@link ChannelOption} over a given {@link SocketOption} which is then passed through the underlying
- * {@link NetworkChannel}.
+ * Provides {@link ChannelOption} over a given {@link java.net.SocketOption} which is then passed through the underlying
+ * {@link java.nio.channels.NetworkChannel}.
  */
 public final class NioChannelOption<T> extends ChannelOption<T> {
 
-    private final SocketOption<T> option;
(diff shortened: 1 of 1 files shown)


normanmaurer:
@koo-taejin can you verify this fix ?

transport/src/main/java/io/netty/channel/socket/nio/NioChannelOption.java
```diff
@@ -19,33 +19,38 @@
...
     /**
-     * Returns a {@link ChannelOption} for the given {@link SocketOption}.
+     * Returns a {@link ChannelOption} for the given {@link java.net.SocketOption}.
      */
-    public static <T> ChannelOption<T> of(SocketOption<T> option) {
+    public static <T> ChannelOption<T> of(java.net.SocketOption<T> option) {
         return new NioChannelOption<T>(option);
     }
 
+    // It's important to not use java.nio.channels.NetworkChannel in the method signature directly as this will lead
```
ejona86: This comment feels weak or misleading, as it's actually the callers of this method that cause the trouble. Maybe:
> It's important to not use java.nio.channels.NetworkChannel as otherwise the classes that sometimes call this method may not be used on Java 6, as method linking can happen eagerly even if this method was not actually called at runtime.

carl-mastrangelo:
@normanmaurer do you regularly test with multiple JDK versions?

normanmaurer:
@carl-mastrangelo yep... that said I need to add a job that tests with java6 as well :(

koo-taejin:
@normanmaurer I have checked that it works well. :)

normanmaurer:
@koo-taejin thanks!

develar:
Thanks a lot for fix, because master of IntelliJ IDEA community is affected (183). Do you have any plans to release 4.1.29 soon?

normanmaurer:
Yes this week :)

Issue #8170, opened 2018-08-02
NPE with BoringSSL when trying to init with unsupported Cipher


rkapsi:
Trying to run the code from #8165 will NPE when a client attempts to connect.

```java
java.lang.NullPointerException: null
	at io.netty.handler.ssl.ReferenceCountedOpenSslContext.destroy(ReferenceCountedOpenSslContext.java:489) ~[netty-all-4.1.28.Final.jar:4.1.28.Final]
	at io.netty.handler.ssl.ReferenceCountedOpenSslContext.access$100(ReferenceCountedOpenSslContext.java:75) ~[netty-all-4.1.28.Final.jar:4.1.28.Final]
	at io.netty.handler.ssl.ReferenceCountedOpenSslContext$2.deallocate(ReferenceCountedOpenSslContext.java:122) ~[netty-all-4.1.28.Final.jar:4.1.28.Final]
	at io.netty.util.AbstractReferenceCounted.release0(AbstractReferenceCounted.java:82) ~[netty-all-4.1.28.Final.jar:4.1.28.Final]
	at io.netty.util.AbstractReferenceCounted.release(AbstractReferenceCounted.java:71) ~[netty-all-4.1.28.Final.jar:4.1.28.Final]
	at io.netty.handler.ssl.ReferenceCountedOpenSslContext.release(ReferenceCountedOpenSslContext.java:601) ~[netty-all-4.1.28.Final.jar:4.1.28.Final]
	at io.netty.handler.ssl.ReferenceCountedOpenSslContext.<init>(ReferenceCountedOpenSslContext.java:309) ~[netty-all-4.1.28.Final.jar:4.1.28.Final]
	at io.netty.handler.ssl.OpenSslContext.<init>(OpenSslContext.java:43) ~[netty-all-4.1.28.Final.jar:4.1.28.Final]
	at io.netty.handler.ssl.OpenSslServerContext.<init>(OpenSslServerContext.java:347) ~[netty-all-4.1.28.Final.jar:4.1.28.Final]
	at io.netty.handler.ssl.OpenSslServerContext.<init>(OpenSslServerContext.java:335) ~[netty-all-4.1.28.Final.jar:4.1.28.Final]
	at io.netty.handler.ssl.SslContext.newServerContextInternal(SslContext.java:421) ~[netty-all-4.1.28.Final.jar:4.1.28.Final]
[...]

PR #8171, opened 2018-08-02
Fix NPE exception when using invalid cipher during building SslContext.

normanmaurer:
Motivation:

We missed to do a null check before trying to destroy the OpenSslSessionContext, which could lead to a NPE.

Modifications:

Add null check and tests.

Result:

Fix https://github.com/netty/netty/issues/8170.


normanmaurer:
@rkapsi thanks for reporting

PR #8176, opened 2018-08-06
CompositeByteBuf nioBuffer doesn't always alloc

vincent-grosbois:
In nioBuffer(int,int) in CompositeByteBuf , we create a sub-array of nioBuffers for the components that are in range, then concatenate all the components in range into a single bigger buffer. 
However, if the call to nioBuffers() returned only one sub-buffer, then we are copying it to a newly-allocated buffer "merged" for no reason.

Motivation:

Profiler for Spark shows a lot of time spent in put() method inside nioBuffer(), while usually no copy of data is required.

Modification:
This change skips this last step and just returns a duplicate of the single buffer returned by the call to nioBuffers(), which will in most implementation not copy the data

Result:
No copy when the source is only 1 buffer

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
@@ -1467,9 +1467,14 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
             }
         }
 
-        ByteBuffer merged = ByteBuffer.allocate(length).order(order());
         ByteBuffer[] buffers = nioBuffers(index, length);
 
+        if(buffers.length == 1)
+        {
+            return buffers[0].duplicate();
+        }
+        
+        ByteBuffer merged = ByteBuffer.allocate(length).order(order());
         for (ByteBuffer buf: buffers) {
             merged.put(buf);
         }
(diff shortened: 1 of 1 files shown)


vincent-grosbois:
Does anybody know why the check is failing?

johnou:
https://garage.netty.io/teamcity/viewLog.html?tab=buildLog&buildTypeId=pulls_netty&buildId=32120#_focus=709

[14:32:01][io.netty:netty-buffer] /opt/teamcity-agent/work/a00bdf77e621d005/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java:1475: trailing whitespace

vincent-grosbois:
should be ok now

normanmaurer:
@vincent-grosbois can you please sing our ICLA and let me know once done: https://netty.io/s/icla

normanmaurer:
@vincent-grosbois thanks! BTW i Wonder why spark needs to call
This method at all. Do you have a profiler snapshot that shows where it is called ?

vincent-grosbois:
I don't have a profiler snapshot but I have several out of memory errors whose call stack is close by in Spark :)
cf this method : https://github.com/apache/spark/blob/3f4bda7289f1bfbbe8b9bc4b516007f569c44d2e/common/network-common/src/main/java/org/apache/spark/network/protocol/MessageWithHeader.java#L139
buf.nioBuffer(buf.readerIndex(), length) is usually a call to a composite buffer and we are trying to read like 4KB at a time... so usually it's enough to just slice the underlying buffer and not realloc a new one are 4 KB is usually totally included in an underlying bytebuf

dbtsai:
@vincent-grosbois Have you tried this in Spark? Does this solve the out of memory errors you observe? We're cutting Spark 2.4 soon, and it looks to me a big issue and we might try to get this in. Thanks.

PR #8178, opened 2018-08-07
Set initial SNI hostname when creating an OpenSSL engine in client mode

jianglai:
Motivation:

When using the JDK SSL provider in client mode, the SNI host names (called serverNames in SslEngineImpl) is set to the peerHost (if available) that is used to initialize the SSL Engine:

http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/sun/security/ssl/SSLEngineImpl.java#l377

This allows one to call SslEngine.getSSLParameters() and inspect what is the SNI name to be sent. The same should be done in the OpenSSL provider as well. Currently even though the the SNI name is sent by the OpenSSL provider during handshake when the peerHost is specified, it is missing from the parameters.

Modification:

Set the sniHostNames field when SNI is to be used. Also verifies the peer is actually a hostname before setting it as the SNI name, which is consistent with JDK SSL provider's behavior.

Result:

SslEngine using the OpenSSL provider created in client mode with peerHost should initialize sniHostNames with the peerHost.

Calling SslEngine.getSSLParameters().getServerNames() should return a list that contains that name.

PS:

I'm not sure where to put tests. Please advise.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslUtils.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslUtils.java
@@ -21,6 +21,7 @@ import io.netty.buffer.ByteBufUtil;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.codec.base64.Base64;
 import io.netty.handler.codec.base64.Base64Dialect;
+import io.netty.util.NetUtil;
 
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
@@ -347,6 +348,17 @@ final class SslUtils {
                 src.readableBytes(), true, Base64Dialect.STANDARD, allocator);
         src.readerIndex(src.writerIndex());
         return dst;
+    }
+
+    /**
+     * Validate that the given hostname can be used in SNI extension.
+     */
+    static boolean isValidHostNameForSNI(String hostname) {
+        return hostname != null &&
+               hostname.indexOf('.') > 0 &&
+               !hostname.endsWith(".") &&
+               !NetUtil.isValidIpV4Address(hostname) &&
+               !NetUtil.isValidIpV6Address(hostname);
     }
 
     private SslUtils() {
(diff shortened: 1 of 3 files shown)


handler/src/main/java/io/netty/handler/ssl/Java8SslUtils.java
```diff
@@ -24,11 +24,20 @@
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import sun.net.util.IPAddressUtil;
```
normanmaurer: Can we NetUtil that is part of Netty ?

jianglai: Done.

handler/src/main/java/io/netty/handler/ssl/Java8SslUtils.java
```diff
@@ -29,6 +31,14 @@
 
     private Java8SslUtils() { }
 
+    static boolean isValidHostName(String hostname) {
```
normanmaurer: I think this can just be moved to SslUtils now as it has no java8 dependencies

jianglai: Done.

jianglai:
PTAL. All tests passed.

handler/src/main/java/io/netty/handler/ssl/SslUtils.java
```diff
@@ -349,6 +350,19 @@ static ByteBuf toBase64(ByteBufAllocator allocator, ByteBuf src) {
         return dst;
     }
 
+    /**
+     * Validate that the given hostname can be used in SNI extension.
+     *
+     * @see sun.security.ssl.Utilities#rawToSNIHostName(String)
```
normanmaurer: @jianglai please remove this as it points to some "private" class that may not be present depending on the JDK you use.

normanmaurer:
@jianglai please address the last one comment and squash. Will merge after this was done.

normanmaurer:
@jianglai also please sign the ICLA:

https://netty.io/s/icla

jianglai:
Removed reference to internal methods and squashed. PTAL.

Regarding the CLA, I am contributing as an employee of Google. Do I still need to sign an individual CLA? Please advise. Thank you!

jianglai:
Understood. ICLA signed.

normanmaurer:
@jianglai thanks!

PR #8177, opened 2018-08-07
Correctly handle DNS redirects for NS servers that have no ADDITIONAL…

normanmaurer:
… record

Motiviation:

We incorrectly did ignore NS servers during redirect which had no ADDITIONAL record. This could at worse have the affect that we failed the query completely as none of the NS servers had a ADDITIONAL record. Beside this using a DnsCache to cache authoritative nameservers does not work in practise as we we need different features and semantics when cache these servers (for example we also want to cache unresolved nameservers and resolve these on the fly when needed).

Modifications:

- Correctly take NS records into account that have no matching ADDITIONAL record
- Correctly handle multiple ADDITIONAL records for the same NS record
- Introduce AuthoritativeDnsServerCache as a replacement of the DnsCache when caching authoritative nameservers + adding default implementation
- Add an adapter layer to reduce API breakage as much as possible
- Replace DnsNameResolver.uncachedRedirectDnsServerStream(...) with uncachedRedirectDnsServerList(...)
- Add unit tests

Result:

Our DnsResolver now correctly handle redirects in all cases.

PR #8181, opened 2018-08-08
Correctly handle hostnames with and without trailing dot in the DefaultDnsCache and use it for searchdomains.

normanmaurer:
Correctly handle hostnames with and without trailing dot in the DefaultDnsCache and use it for searchdomains.

    Motivation:

    We should ensure we return the same cached entries for the hostname and hostname ending with dot. Beside this we also should use it for the searchdomains as well.

    Modifications:

    - Internally always use hostname with a dot as a key and so ensure we correctly handle it in the cache.
    - Also query the cache for each searchdomain
    - Add unit tests

    Result:

    Use the same cached entries for hostname with and without trailing dot. Query the cache for each searchdomain query as well

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
@@ -19,6 +19,7 @@ import io.netty.channel.EventLoop;
 import io.netty.handler.codec.dns.DnsRecord;
 import io.netty.util.concurrent.ScheduledFuture;
 import io.netty.util.internal.PlatformDependent;
+import io.netty.util.internal.StringUtil;
 import io.netty.util.internal.UnstableApi;
 
 import java.net.InetAddress;
@@ -115,7 +116,7 @@ public class DefaultDnsCache implements DnsCache {
     @Override
     public boolean clear(String hostname) {
         checkNotNull(hostname, "hostname");
-        Entries entries = resolveCache.remove(hostname);
+        Entries entries = resolveCache.remove(appendDot(hostname));
         return entries != null && entries.clearAndCancel();
     }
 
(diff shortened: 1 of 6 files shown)


normanmaurer:
@slandelle FYI as well

normanmaurer:
@trustin maybe just doing it in the default impl would be good enough. Let me do this

normanmaurer:
@trustin PTAL again

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -115,7 +116,7 @@ public void clear() {
     @Override
     public boolean clear(String hostname) {
         checkNotNull(hostname, "hostname");
-        Entries entries = resolveCache.remove(hostname);
+        Entries entries = resolveCache.remove(appendDot(hostname));
```
carl-mastrangelo: What happens if there is a search path?  If my search path is `netty.io`  then the domain name `foo` would be incorrectly cached as `foo.`  right?

normanmaurer: We should cache foo.netty.io in this case.

normanmaurer: @carl-mastrangelo based on your question (thanks!) I also ensured we correctly query the cache for the use-case (+ added a test).

resolver-dns/src/main/java/io/netty/resolver/dns/DnsResolveContext.java
```diff
@@ -213,12 +217,10 @@ public Throwable fillInStackTrace() {
         }
     }
 
-    private void doSearchDomainQuery(String hostname, FutureListener<List<T>> listener) {
+    void doSearchDomainQuery(String hostname, Promise<List<T>> nextPromise) {
         DnsResolveContext<T> nextContext = newResolverContext(parent, hostname, dnsClass, expectedTypes,
-                                                              additionals, nameServerAddrs);
-        Promise<List<T>> nextPromise = parent.executor().newPromise();
+                additionals, nameServerAddrs);
```
Scottmitch: nit: restore alignment?

trustin: I also see a lot of similar issues in #8177. Perhaps we should make sure we all are using same IntelliJ code style settings?

PR #8187, opened 2018-08-09
Keep the amount of scheduled tasks for DefaultDnsCache at a minimum

normanmaurer:
Motivation:

We are currently always remove all entries from the cache for a hostname if the lowest TTL was reached but schedule one for each of the cached entries. This is wasteful.

Modifications:

- Reimplement logic to schedule TTL to only schedule a new removal task if the requested TTL was actual lower then the one for the already scheduled task.
- Ensure we only remove from the internal map if we did not replace the Entries in the meantime.

Result:

Less overhead in terms of scheduled tasks for the DefaultDnsCache

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
@@ -17,7 +17,6 @@ package io.netty.resolver.dns;
 
 import io.netty.channel.EventLoop;
 import io.netty.handler.codec.dns.DnsRecord;
-import io.netty.util.concurrent.ScheduledFuture;
 import io.netty.util.internal.PlatformDependent;
 import io.netty.util.internal.StringUtil;
 import io.netty.util.internal.UnstableApi;
@@ -29,8 +28,11 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.Delayed;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 
 import static io.netty.util.internal.ObjectUtil.checkNotNull;
 import static io.netty.util.internal.ObjectUtil.checkPositiveOrZero;
@@ -41,12 +43,55 @@ import static io.netty.util.internal.ObjectUtil.checkPositiveOrZero;
  */
 @UnstableApi
 public class DefaultDnsCache implements DnsCache {
-
-    private final ConcurrentMap<String, Entries> resolveCache = PlatformDependent.newConcurrentHashMap();
-
     // Two years are supported by all our EventLoop implementations and so safe to use as maximum.
     // See also: https://github.com/netty/netty/commit/b47fb817991b42ec8808c7d26538f3f2464e1fa6
     private static final int MAX_SUPPORTED_TTL_SECS = (int) TimeUnit.DAYS.toSeconds(365 * 2);
+
+    private static final ScheduledFuture<?> CANCELLED = new ScheduledFuture<Object>() {
+
+        @Override
+        public boolean cancel(boolean mayInterruptIfRunning) {
+            return false;
+        }
+
+        @Override
+        public long getDelay(TimeUnit unit) {
+            // We ignore unit and always return the minimum value to ensure the TTL of the cancelled marker is
+            // the smallest.
+            return Long.MIN_VALUE;
+        }
+
+        @Override
+        public int compareTo(Delayed o) {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public boolean isCancelled() {
+            return true;
+        }
+
+        @Override
+        public boolean isDone() {
+            return true;
+        }
+
+        @Override
+        public Object get() {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public Object get(long timeout, TimeUnit unit) {
+            throw new UnsupportedOperationException();
+        }
+    };
+
+    private static final AtomicReferenceFieldUpdater<DefaultDnsCache.Entries, ScheduledFuture> FUTURE_UPDATER =
+            AtomicReferenceFieldUpdater.newUpdater(
+                    DefaultDnsCache.Entries.class, ScheduledFuture.class, "expirationFuture");
+
+    private final ConcurrentMap<String, Entries> resolveCache = PlatformDependent.newConcurrentHashMap();
     private final int minTtl;
     private final int maxTtl;
     private final int negativeTtl;
@@ -146,7 +191,7 @@ public class DefaultDnsCache implements DnsCache {
             return e;
         }
         cache0(appendDot(hostname), e,
-                Math.max(minTtl, Math.min(MAX_SUPPORTED_TTL_SECS, (int) Math.min(maxTtl, originalTtl))), loop);
+                Math.max(minTtl, (int) Math.min(maxTtl, originalTtl)), loop);
         return e;
     }
 
@@ -161,45 +206,20 @@ public class DefaultDnsCache implements DnsCache {
             return e;
         }
 
-        cache0(appendDot(hostname), e, Math.min(MAX_SUPPORTED_TTL_SECS, negativeTtl), loop);
+        cache0(appendDot(hostname), e, negativeTtl, loop);
         return e;
     }
 
     private void cache0(String hostname, DefaultDnsCacheEntry e, int ttl, EventLoop loop) {
         Entries entries = resolveCache.get(hostname);
         if (entries == null) {
-            entries = new Entries(e);
+            entries = new Entries(hostname);
             Entries oldEntries = resolveCache.putIfAbsent(hostname, entries);
             if (oldEntries != null) {
                 entries = oldEntries;
             }
         }
-        entries.add(e);
-
-        scheduleCacheExpiration(hostname, e, ttl, loop);
-    }
-
-    private void scheduleCacheExpiration(final String hostname, final DefaultDnsCacheEntry e,
-                                         int ttl,
-                                         EventLoop loop) {
-        e.scheduleExpiration(loop, new Runnable() {
-                    @Override
-                    public void run() {
-                        // We always remove all entries for a hostname once one entry expire. This is not the
-                        // most efficient to do but this way we can guarantee that if a DnsResolver
-                        // be configured to prefer one ip family over the other we will not return unexpected
-                        // results to the enduser if one of the A or AAAA records has different TTL settings.
-                        //
-                        // As a TTL is just a hint of the maximum time a cache is allowed to cache stuff it's
-                        // completely fine to remove the entry even if the TTL is not reached yet.
-                        //
-                        // See https://github.com/netty/netty/issues/7329
-                        Entries entries = resolveCache.remove(hostname);
-                        if (entries != null) {
-                            entries.clearAndCancel();
-                        }
-                    }
-                }, ttl, TimeUnit.SECONDS);
+        entries.add(e, ttl, loop);
     }
 
     @Override
@@ -217,7 +237,6 @@ public class DefaultDnsCache implements DnsCache {
         private final String hostname;
         private final InetAddress address;
         private final Throwable cause;
-        private volatile ScheduledFuture<?> expirationFuture;
 
         DefaultDnsCacheEntry(String hostname, InetAddress address) {
             this.hostname = checkNotNull(hostname, "hostname");
@@ -241,16 +260,8 @@ public class DefaultDnsCache implements DnsCache {
             return cause;
         }
 
-        void scheduleExpiration(EventLoop loop, Runnable task, long delay, TimeUnit unit) {
-            assert expirationFuture == null : "expiration task scheduled already";
-            expirationFuture = loop.schedule(task, delay, unit);
-        }
-
-        void cancelExpiration() {
-            ScheduledFuture<?> expirationFuture = this.expirationFuture;
-            if (expirationFuture != null) {
-                expirationFuture.cancel(false);
-            }
+        String hostname() {
+            return hostname;
         }
 
         @Override
@@ -264,13 +275,18 @@ public class DefaultDnsCache implements DnsCache {
     }
 
     // Directly extend AtomicReference for intrinsics and also to keep memory overhead low.
-    private static final class Entries extends AtomicReference<List<DefaultDnsCacheEntry>> {
+    private final class Entries extends AtomicReference<List<DefaultDnsCacheEntry>> implements Runnable {
 
-        Entries(DefaultDnsCacheEntry entry) {
-            super(Collections.singletonList(entry));
+        private final String hostname;
+        // Needs to be package-private to be able to access it via the AtomicReferenceFieldUpdater
+        volatile ScheduledFuture<?> expirationFuture;
+
+        Entries(String hostname) {
+            super(Collections.<DefaultDnsCacheEntry>emptyList());
+            this.hostname = hostname;
         }
 
-        void add(DefaultDnsCacheEntry e) {
+        void add(DefaultDnsCacheEntry e, int ttl, EventLoop loop) {
             if (e.cause() == null) {
                 for (;;) {
                     List<DefaultDnsCacheEntry> entries = get();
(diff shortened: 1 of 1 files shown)


normanmaurer:
This was inspired by this discussion:

https://github.com/netty/netty/pull/8177#discussion_r208804367

normanmaurer:
@slandelle @vietj  FYI as well.

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -266,22 +228,29 @@ public String toString() {
     }
 
     // Directly extend AtomicReference for intrinsics and also to keep memory overhead low.
-    private static final class Entries extends AtomicReference<List<DefaultDnsCacheEntry>> {
+    private final class Entries extends AtomicReference<List<DefaultDnsCacheEntry>> implements Runnable {
+
+        private final String hostname;
+
+        // This is guarded by a synchronized block.
```
carl-mastrangelo: Do you use `@GuardedBy` ?

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -327,15 +317,28 @@ boolean clearAndCancel() {
                 return false;
             }
 
-            cancelExpiration(entries);
+            ScheduledFuture<?> expirationFuture = this.expirationFuture;
```
Scottmitch: `this.expirationFuture` is modified in a synchronized block, but accessed outside a synchronized block here. Can you describe how visibility is guaranteed here and the implications if the change is not visible? 

An alternative approach would be to avoid using `synchronized` and using a token/marker object to indicate cancelled.

```diff
+diff --git a/resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java b/resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
index 6ce975ff7a..5004929512 100644
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
@@ -17,6 +17,8 @@ package io.netty.resolver.dns;
 
 import io.netty.channel.EventLoop;
 import io.netty.handler.codec.dns.DnsRecord;
+import io.netty.util.concurrent.Future;
+import io.netty.util.concurrent.GenericFutureListener;
 import io.netty.util.concurrent.ScheduledFuture;
 import io.netty.util.internal.PlatformDependent;
 import io.netty.util.internal.UnstableApi;
@@ -28,7 +30,10 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.Delayed;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
 
 import static io.netty.util.internal.ObjectUtil.checkNotNull;
@@ -232,8 +237,125 @@ public class DefaultDnsCache implements DnsCache {
 
         private final String hostname;
 
-        // This is guarded by a synchronized block.
-        private ScheduledFuture<?> expirationFuture;
+        // TODO: most methods don't matter, because the only meaningful method is getDelay, but cleanup this code.
+        private final ScheduledFuture<?> CANCELLED = new ScheduledFuture<Object>() {
+            @Override
+            public boolean isSuccess() {
+                return false;
+            }
+
+            @Override
+            public boolean isCancellable() {
+                return false;
+            }
+
+            @Override
+            public Throwable cause() {
+                return null;
+            }
+
+            @Override
+            public Future<Object> addListener(GenericFutureListener<? extends Future<? super Object>> listener) {
+                return null;
+            }
+
+            @Override
+            public Future<Object> addListeners(GenericFutureListener<? extends Future<? super Object>>... listeners) {
+                return null;
+            }
+
+            @Override
+            public Future<Object> removeListener(GenericFutureListener<? extends Future<? super Object>> listener) {
+                return null;
+            }
+
+            @Override
+            public Future<Object> removeListeners(GenericFutureListener<? extends Future<? super Object>>... listeners) {
+                return null;
+            }
+
+            @Override
+            public Future<Object> sync() throws InterruptedException {
+                return null;
+            }
+
+            @Override
+            public Future<Object> syncUninterruptibly() {
+                return null;
+            }
+
+            @Override
+            public Future<Object> await() throws InterruptedException {
+                return null;
+            }
+
+            @Override
+            public Future<Object> awaitUninterruptibly() {
+                return null;
+            }
+
+            @Override
+            public boolean await(long timeout, TimeUnit unit) throws InterruptedException {
+                return false;
+            }
+
+            @Override
+            public boolean await(long timeoutMillis) throws InterruptedException {
+                return false;
+            }
+
+            @Override
+            public boolean awaitUninterruptibly(long timeout, TimeUnit unit) {
+                return false;
+            }
+
+            @Override
+            public boolean awaitUninterruptibly(long timeoutMillis) {
+                return false;
+            }
+
+            @Override
+            public Object getNow() {
+                return null;
+            }
+
+            @Override
+            public boolean cancel(boolean mayInterruptIfRunning) {
+                return false;
+            }
+
+            @Override
+            public long getDelay(TimeUnit unit) {
+                return Long.MIN_VALUE;
+            }
+
+            @Override
+            public int compareTo(Delayed o) {
+                return 0;
+            }
+
+            @Override
+            public boolean isCancelled() {
+                return false;
+            }
+
+            @Override
+            public boolean isDone() {
+                return false;
+            }
+
+            @Override
+            public Object get() throws InterruptedException, ExecutionException {
+                return null;
+            }
+
+            @Override
+            public Object get(long timeout, TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
+                return null;
+            }
+        };
+        // TODO: make this atomic reference updater
+        private final AtomicReference<ScheduledFuture<?>> expirationFutureRef = new AtomicReference<ScheduledFuture<?>>();
 
         Entries(String hostname) {
             super(Collections.<DefaultDnsCacheEntry>emptyList());
@@ -296,19 +418,29 @@ public class DefaultDnsCache implements DnsCache {
         }
 
         private void scheduleCacheExpirationIfNeeded(int ttl, EventLoop loop) {
-            ScheduledFuture<?> oldScheduled;
-
-            synchronized (this) {
-                oldScheduled = this.expirationFuture;
-                if (oldScheduled == null || oldScheduled.getDelay(TimeUnit.SECONDS) > ttl) {
-                    this.expirationFuture = loop.schedule(this, ttl, TimeUnit.SECONDS);
+            for (;;) {
+                ScheduledFuture<?> oldFuture = expirationFutureRef.get();
+                if (oldFuture == null || oldFuture.getDelay(TimeUnit.SECONDS) > ttl) {
+                    ScheduledFuture<?> newFuture = loop.schedule(this, ttl, TimeUnit.SECONDS);
+                    // It is possible that
+                    // 1. task will fire in between this line, or
+                    // 2. multiple timers may be set if there is concurrency
+                    // (1) Shouldn't be a problem because we will fail the CAS and then the next loop will see CANCELLED
+                    //     so the ttl will not be less, and we will bail out of the loop.
+                    // (2) This is a trade-off to avoid concurrency resulting in contention on a synchronized block.
+                    //     IIUC this would be an overall reduction in existing scheduling so this is likely acceptable.
+                    if (expirationFutureRef.compareAndSet(oldFuture, newFuture)) {
+                        if (oldFuture != null) {
+                            oldFuture.cancel(true);
+                        }
+                        break;
+                    } else {
+                        newFuture.cancel(true);
+                    }
                 } else {
-                    oldScheduled = null;
+                    break;
                 }
             }
-            if (oldScheduled != null) {
-                oldScheduled.cancel(false);
-            }
         }
 
         boolean clearAndCancel() {
@@ -317,7 +449,7 @@ public class DefaultDnsCache implements DnsCache {
                 return false;
             }
 
-            ScheduledFuture<?> expirationFuture = this.expirationFuture;
+            ScheduledFuture<?> expirationFuture = expirationFutureRef.getAndSet(CANCELLED);
             if (expirationFuture != null) {
                 expirationFuture.cancel(false);
             }
```

Note existing unit tests pass but more review would be useful :)

normanmaurer: @Scottmitch marked as volatile for now to fix the visibility concerns (good catch). Let me think about your idea a bit. I guess at worse we could do this as a followup ? IMHO the existing approach in this PR is already a nice improvement. Also note that I would not expect a lot of condition on the synchronized block here as the synchronisation only happens when modification for the same hostname happens concurrently (which most likely will not happened).

normanmaurer:
@carl-mastrangelo consider to recheck as well :)

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -41,12 +43,53 @@
...
+        }
+
+        @Override
+        public long getDelay(TimeUnit unit) {
+            return Long.MIN_VALUE;
+        }
+
+        @Override
+        public int compareTo(Delayed o) {
+            return -1;
```
Scottmitch: this won't be used ... but consider making this play nicely with the interface: `return MathUtil.compare(getDelay(TimeUnit.NANOSECONDS), o.getDelay(TimeUnit.NANOSECONDS));`

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -41,12 +43,53 @@
...
+    private static final ScheduledFuture<?> CANCELLED = new ScheduledFuture<Object>() {
+
+        @Override
+        public boolean cancel(boolean mayInterruptIfRunning) {
+            return false;
+        }
+
+        @Override
+        public long getDelay(TimeUnit unit) {
+            return Long.MIN_VALUE;
```
Scottmitch: consider adding a comment saying that we explicitly don't want to follow the interface here and we ignore `unit` so that we return the minimum value for comparison purposes ... the units are always expected to be SECONDS ... etc...

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -301,21 +317,53 @@ void add(DefaultDnsCacheEntry e) {
                                 assert replacedEntry == null;
                                 replacedEntry = entry;
                             }
-                        }
+                            i++;
+                        } while (i < entries.size());
```
Scottmitch: consider moving the post-increment above to a pre-increment here for more compactness `i` -> `++i`

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -301,21 +317,53 @@ void add(DefaultDnsCacheEntry e) {
...
                         newEntries.add(e);
-                        if (compareAndSet(entries, newEntries)) {
+                        if (compareAndSet(entries, Collections.unmodifiableList(newEntries))) {
                             if (replacedEntry != null) {
```
Scottmitch: this is an `if` block with no code contents. consider removing the if block and just keep the comments.

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -290,7 +305,8 @@ void add(DefaultDnsCacheEntry e) {
                         // Create a new List for COW semantics
                         List<DefaultDnsCacheEntry> newEntries = new ArrayList<DefaultDnsCacheEntry>(entries.size() + 1);
                         DefaultDnsCacheEntry replacedEntry = null;
```
Scottmitch: it seems like this is no longer used ... can it be removed?

trustin:
I see code duplication between this PR and #8177. Could we extract this logic into a common class?

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
trustin: We could throw some exception here if we are sure it's not going to be used.

normanmaurer:
@trustin yes I will do this one this one is merged

normanmaurer:
@trustin @Scottmitch all addressed

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -301,21 +317,53 @@ void add(DefaultDnsCacheEntry e) {
...
+                set(Collections.singletonList(e));
+                scheduleCacheExpirationIfNeeded(ttl, loop);
+            }
+        }
+
+        private void scheduleCacheExpirationIfNeeded(int ttl, EventLoop loop) {
+            for (;;) {
+                ScheduledFuture<?> oldFuture = FUTURE_UPDATER.get(this);
+                if (oldFuture == null || oldFuture.getDelay(TimeUnit.SECONDS) > ttl) {
+                    ScheduledFuture<?> newFuture = loop.schedule(this, ttl, TimeUnit.SECONDS);
```
carl-mastrangelo: dont want to be too nitpicky, but this isn't quite correct.  I think you need to calculate System.nanoTime() as the start point outside of the loop, and deduct when scheduling:

```java
long start = System.nanoTime();
long ttlNanos = TimeUnit.SECONDS.toNanos(ttl);
// ...
for (;;) {
  loop.schedule(this, ttlNanos - (System.nanoTime() - start), TimeUnit.NANOSECONDS);
}

```

normanmaurer: I think you are right in general I am just not sure we need to be so picky here... that said I guess it can’t really harm

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -168,38 +213,13 @@ public DnsCacheEntry cache(String hostname, DnsRecord[] additionals, Throwable c
     private void cache0(String hostname, DefaultDnsCacheEntry e, int ttl, EventLoop loop) {
```
Scottmitch: while you are here :) `MAX_SUPPORTED_TTL_SECS` is used on line 194 and 209 ... but it looks unnecessary because it is already enforced in the constructor. consider removing.

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -289,33 +306,60 @@ void add(DefaultDnsCacheEntry e) {
...
+            }
+        }
+
+        private void scheduleCacheExpirationIfNeeded(int ttl, EventLoop loop) {
+            long start = System.nanoTime();
+            long ttlNanos = TimeUnit.SECONDS.toNanos(ttl);
+
+            for (;;) {
+                ScheduledFuture<?> oldFuture = FUTURE_UPDATER.get(this);
+                long calculatedTTL = ttlNanos - (System.nanoTime() - start);
```
Scottmitch: extension of 
https://github.com/netty/netty/pull/8187#discussion_r209771462 ... I'm not convinced this is really necessary as the TTL granularity is in seconds and isn't tightly synchronized anyways. The cost to call `nanoTime` is not "free" either and this adds complexity. My vote would be to remove it and simplify for now unless this is proven to be a problem.

If we do think this is necessary we should ensure we only call `nanoTime` once per iteration by introducing a local variable to store the "last result in". This code currently calls `nanoTime` twice for the first loop iteration.

/cc @normanmaurer @carl-mastrangelo

normanmaurer: I think I am with @Scottmitch here.. Let me revert the change and add a comment.

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -41,12 +43,55 @@
...
 @UnstableApi
 public class DefaultDnsCache implements DnsCache {
-
-    private final ConcurrentMap<String, Entries> resolveCache = PlatformDependent.newConcurrentHashMap();
-
     // Two years are supported by all our EventLoop implementations and so safe to use as maximum.
     // See also: https://github.com/netty/netty/commit/b47fb817991b42ec8808c7d26538f3f2464e1fa6
     private static final int MAX_SUPPORTED_TTL_SECS = (int) TimeUnit.DAYS.toSeconds(365 * 2);
+
+    private static final ScheduledFuture<?> CANCELLED = new ScheduledFuture<Object>() {
```
Scottmitch: reminder: share this with the authoritative DNS cache when merged.

normanmaurer: Yep

PR #8189, opened 2018-08-10
The response body is compressed when the content type of the response in types.

BerryWang1996:
Motivation:

The response body is compressed when the content type of the response in types.

Modifications:

Add parameter named "types" in constructor, and add related code in method named "beginEncode".

Result:

If content type in response body is not contained by parameter named types, the response body will not be compressed.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
@@ -37,7 +37,7 @@ public class HttpContentCompressor extends HttpContentEncoder {
     private final int windowBits;
     private final int memLevel;
     private final int contentSizeThreshold;
-    private Set<String> types = null;
+    private Set<String> types;
     private ChannelHandlerContext ctx;
 
     /**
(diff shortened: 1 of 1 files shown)


normanmaurer:
@BerryWang1996 what is the reason for the PR ? It seems like it not changes anything in terms of functionality.

normanmaurer:
@BerryWang1996 why you need this?

BerryWang1996:
@normanmaurer Download file failed when use HttpContentCompressor. The binary stream should not be compressed...I think set content type is flexible than before...

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java
```diff
@@ -103,6 +108,38 @@ public HttpContentCompressor(int compressionLevel, int windowBits, int memLevel)
...
+     * @param memLevel
+     *        How much memory should be allocated for the internal compression
+     *        state.  {@code 1} uses minimum memory and {@code 9} uses maximum
+     *        memory.  Larger values result in better and faster compression
+     *        at the expense of memory usage.  The default value is {@code 8}
+     * @param contentSizeThreshold
+     *        The response body is compressed when the size of the response
+     *        body exceeds the threshold. The value should be a non negative
+     *        number. {@code 0} will enable compression for all responses.
+     * @param types
```
normanmaurer: We should make this `String... types`

normanmaurer:
@BerryWang1996 please add a unit test

normanmaurer:
Closing due inactivity

PR #8191, opened 2018-08-10
add volatile to registered

huyuechy:
Motivation:

Modification:

Result:

Fixes #<GitHub issue number>. 


normanmaurer:
@huyuechy why you think this is needed ? This is only ever accessed when in a `synchronized(this)` block

johnou:
seems unnecessary to me too.

normanmaurer:
I will just close this... If you still think this is needed please re-open and add some details why you think so.

PR #8193, opened 2018-08-12
Rename SslHandler.close(...) to closeOutbound(...) as it is still useful and delegate to the methods.

normanmaurer:
Motivation:

    Sometimes the user may want to send a close_notify without closing the underlying Channel. For this we offered the SslHandler.close(...) methods which were marked as deeprecated. We should offer an way to still do this without the user calling deprecated methods.

    See https://stackoverflow.com/questions/51710231/using-nettys-sslhandlerclosechannelhandlercontext-channelpromise/51753742#comment90555949_51753742 .

    Modifications:

    - Remove deprecation of the SslHandler.close(...) method that exactly allows this and rename these to closeOutbound(...) as this is more clear.
    - Add close(...) methods that delegate to these and mark these as deprecated.

    Result:

    Be able to send close_notify without closing the Channel.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -93,8 +93,8 @@ import static io.netty.handler.ssl.SslUtils.getEncryptedPacketLength;
  *
  * <h3>Closing the session</h3>
  * <p>
- * To close the SSL session, the {@link #close()} method should be
- * called to send the {@code close_notify} message to the remote peer.  One
+ * To close the SSL session, the {@link #closeOutbound()} method should be
+ * called to send the {@code close_notify} message to the remote peer. One
  * exception is when you close the {@link Channel} - {@link SslHandler}
  * intercepts the close request and send the {@code close_notify} message
  * before the channel closure automatically.  Once the SSL session is closed,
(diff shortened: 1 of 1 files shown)


handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -621,43 +621,62 @@ public String applicationProtocol() {
...
+     * destroys the underlying {@link SSLEngine}. This will <strong>not</strong> close the underlying
+     * {@link Channel}. If you want to also close the {@link Channel} use {@link Channel#close()} or
+     * {@link ChannelHandlerContext#close()}
+     */
+    public ChannelFuture closeOutbound() {
+        return closeOutbound(ctx.newPromise());
+    }
+
+    /**
+     * See {@link #closeOutbound()}
```
trustin: Better copying the Javadoc from closeOutbound()

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -621,43 +621,62 @@ public String applicationProtocol() {
         return handshakePromise;
     }
 
-    /**
-     * Sends an SSL {@code close_notify} message to the specified channel and
-     * destroys the underlying {@link SSLEngine}.
-     *
-     * @deprecated use {@link Channel#close()} or {@link ChannelHandlerContext#close()}
+     /**
```
trustin: Indentation?

BaurzhanSakhariev:
I tried using `closeOutbound` but I don't understand how we can leverage open connection if `SslHandler` on the other side [closes connection](https://github.com/netty/netty/blob/849649b48d3c8fbf86565b8fa28a07f8b673adaf/handler/src/main/java/io/netty/handler/ssl/SslHandler.java#L1091-L1101) anyway. I did `closeOutbound` + listener with removal of SslHandler on server side and sslCloseFuture listener with sslHandler removal on client side - it looks like client sends `close_notify` in reply to server's `close_notify` but connection is lost so it's not possible to keep using same connection but with plaintext. 

Did I understand the purpose of `closeOutbound` correctly?

Issue #8196, opened 2018-08-14
(Nio|Epoll)EventLoop.pendingTasks should not block when called outside of event loop


vkostyukov:
### Expected behavior

`EventLoop.pendingTasks` should be (reasonably) cheap to invoke so it can be used within observability.

### Actual behavior

We've been experimenting with overriding `EventExecutorChooserFactory` with more superior load balancing algorithms and found out that it might be quite non-trivial to support pending-tasks load metric (pick the least loaded event loop).

After poking around the code, it seems when querying `pendingTasks` of either `NioEventLoop`  or `EpollEventLoop`, the method [blocks if called outside of said event loop](https://github.com/netty/netty/blob/630c82717dd8d822323c9e4a8ad51ea4f8a38f86/transport/src/main/java/io/netty/channel/nio/NioEventLoop.java#L262-L272). Put it this way, it's only cheap to query event loop's pending tasks from within the same event loop. Perhaps I miss some (historical) context for `pendingTask`, but for an uneducated reader, it just defeats the purpose of a method providing visibility into an event loop.

I was hoping to pick up some context on the current behavior of `pendingTasks` and see if we ever want/will to change it.

normanmaurer:
@vkostyukov thats a good point... We may want to "replace" the current implementation with an `AtomicInteger` that is updated . WDYT ?

vkostyukov:
@normanmaurer Yeah, that seems reasonable to me. Even if we won't be able to guarantee certain prevention it should be totally fine for the sake of observability (provided we document it).

I'd be happy to contribute a fix for this unless somebody else want to work on this.

normanmaurer:
@vkostyukov that would be awesome... Like you already stated we should just make this "as cheap as possible".

normanmaurer:
Perfect... let her know I am always happy to help 👍

Wingfril:
I'm not really sure why we even need to do that? Can't we just get the super.size() itself without worrying if it is in the event loop? It seems `MpscQueue` is no longer in the code, and so live locking shouldn't be an issue anymore? `SingleThreadEventExecutor` uses a `LinkedBlockingQueue` which has a size variable which is an `atomicInteger` already?

johnou:
[...]

PR #8197, opened 2018-08-15
Pending task size

Wingfril:
Make pendingTask an atomic integer

Motivation:

When trying to access the number of pending tasks in an event loop, 
it is fairly expensive if we are not in the event loop. 

Modification:

Create an atomic integer to keep track when tasks are added and removed from the queue. 

Result:

Fixes #<8196>. 
`.pendingTask` no longer traverses through the queue and can be accessed outside of the event loop.

### DIFF
--- a/common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
+++ b/common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
@@ -304,7 +304,7 @@ public abstract class SingleThreadEventExecutor extends AbstractScheduledEventEx
      * Return the number of tasks that are pending for processing.
      *
      * <strong>Be aware that this operation may be expensive as it depends on the internal implementation of the
-     * SingleThreadEventExecutor. So use it was care!</strong>
+     * SingleThreadEventExecutor. So use it with care!</strong>
      */
     public int pendingTasks() {
         return taskQueue.size();
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
@@ -73,12 +73,6 @@ final class EpollEventLoop extends SingleThreadEventLoop {
             return epollWaitNow();
         }
     };
-    private final Callable<Integer> pendingTasksCallable = new Callable<Integer>() {
-        @Override
-        public Integer call() throws Exception {
-            return EpollEventLoop.super.pendingTasks();
-        }
-    };
     private volatile int wakenUp;
     private volatile int ioRatio = 50;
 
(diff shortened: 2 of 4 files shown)


common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
```diff
@@ -80,6 +81,7 @@ public void run() {
                     SingleThreadEventExecutor.class, ThreadProperties.class, "threadProperties");
 
     private final Queue<Runnable> taskQueue;
+    private AtomicInteger pendingTaskSize = new AtomicInteger(0);
```
normanmaurer: I think all these changes should go into `EpollEventLoop` / `NioEventLoop` / `KQueueEventLoop` and the `SingleThreadEventExecutor` should not be changed. This is mainly because otherwise it may produce unexpected effects for people that extend the `SingleThreadEventExecutor`.

normanmaurer: This should be final btw.

vietj:
I'm not sure this would be accepted as the cost for volatile might be a problem.

Keeping a private event loop counter with a `lazySet` on a volatile copy might be better.

njhill:
> Keeping a private event loop counter with a `lazySet` on a volatile copy might be better.

I was about to suggest something similar, just wanted to clarify the implication is to now have 3 monotonic fields:
- `a` a volatile counter incremented atomically upon task enqueue
- `b` a non-volatile `int` incremented upon dequeue (by event loop only)
- `c` a volatile `lazySet` copy of the second value 

and then `pendingTasks` just returns `a` - `c`

Wingfril:
@njhill What do you mean by event loop only for `b`? Only decrement if `inEventLoop()` is true?

njhill:
@Wingfril dequeue only happens in the event loop whereas enqueue could be from an arbitrary thread (it's a MPSC queue). Both these counters will only ever increase, overflow shouldn't matter since you only care about the diff.

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
Wingfril: I'm not sure how I can track how many events were run with `runAllTasks(long timeoutNanos)` then?

vietj:
that being said the underlying actual `Queue` has a `size()` operation, is there anything wrong with it ?

njhill:
@vietj the hoops are being jumped through here are because of this comment in `XxxEventLoop`s:
```java
// As we use a MpscQueue we need to ensure pendingTasks() is only executed from within the EventLoop as
// otherwise we may see unexpected behavior (as size() is only allowed to be called by a single consumer).
// See https://github.com/netty/netty/issues/5297
```
however I think this actually no longer applies (see https://github.com/netty/netty/issues/8196#issuecomment-413503916). So yes probably the queue `size()` should just be used again now which obviates the need for this PR.

normanmaurer:
@Wingfril so please just remove the overrides that exists in the `Nio|KQueue|Epoll`EventLoop for `pendingTasks()` and we should be good here:

See https://github.com/netty/netty/issues/8196#issuecomment-413564384

All other changes are not needed. Please ping me once done .

normanmaurer:
@Wingfril please do what @njhill said and then squash all commits + adjust commit message to reflect the actual changes. After this I will merge ... Thanks again!

normanmaurer:
@Wingfril this looks good! Can you just now squash + update the commit message and then force-push ?

normanmaurer:
@Wingfril thanks a lot! I merged it after I adjusted the commit message a bit during squash:

https://github.com/netty/netty/commit/785473788f3f19531294802b727fb10a48938222

PR #8200, opened 2018-08-16
Allow to load PrivateKey via OpenSSL Engine

normanmaurer:
Motivation:

OpenSSL itself has an abstraction which allows you to customize some things. For example it is possible to load the PrivateKey from the engine. We should support this.

Modifications:

Add two new static methods to OpenSslX509KeyManagerFactory which allow to create an OpenSslX509KeyManagerFactory that loads its PrivateKey via the OpenSSL Engine directly.

Result:

More flexible usage of OpenSSL possible


normanmaurer:
This depends on 
https://github.com/netty/netty-tcnative/pull/380

handler/src/main/java/io/netty/handler/ssl/OpenSslX509KeyManagerFactory.java
```diff
@@ -218,4 +230,147 @@ void destroy() {
...
+                                return SslContext.ALIAS;
+                            }
+                        }
+                    }
+                    return null;
+                }
+
+                @Override
+                public void engineStore(OutputStream stream, char[] password)
+                        throws IOException, NoSuchAlgorithmException, CertificateException {
```
Scottmitch: nit: remove the `throws` from the method signature ... other methods in this class have similar opportunities.

handler/src/main/java/io/netty/handler/ssl/OpenSslX509KeyManagerFactory.java
```diff
@@ -218,4 +230,147 @@ void destroy() {
...
+
+    private static final class OpenSslEngineKeyStore extends KeyStore {
+        private OpenSslEngineKeyStore(final X509Certificate[] certificateChain) {
+            super(new KeyStoreSpi() {
+
+                private final Date creationDate = new Date();
+
+                @Override
+                public Key engineGetKey(String alias, char[] password)
+                        throws NoSuchAlgorithmException, UnrecoverableKeyException {
```
Scottmitch: `NoSuchAlgorithmException` can be removed from this method signature.

handler/src/main/java/io/netty/handler/ssl/OpenSslX509KeyManagerFactory.java
```diff
@@ -218,4 +230,147 @@ void destroy() {
...
+                }
+
+                @Override
+                public void engineLoad(InputStream stream, char[] password)
+                        throws IOException, NoSuchAlgorithmException, CertificateException {
+                    if (stream != null && password != null) {
+                        throw new UnsupportedOperationException();
+                    }
+                }
+            }, null, "native");
```
Scottmitch: `"native"` is also used in `OpenSslPrivateKey#getFormat` ... should we share a constant here?

handler/src/main/java/io/netty/handler/ssl/OpenSslPrivateKey.java
```diff
@@ -0,0 +1,188 @@
...
+import io.netty.internal.tcnative.SSL;
+import io.netty.util.AbstractReferenceCounted;
+import io.netty.util.IllegalReferenceCountException;
+
+import javax.security.auth.Destroyable;
+import java.security.PrivateKey;
+
+final class OpenSslPrivateKey extends AbstractReferenceCounted implements PrivateKey {
+
+    private volatile long privateKey;
```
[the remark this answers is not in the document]
normanmaurer: @Scottmitch yeah for visibility... That said it should never be a problem as everything is internal and the last reference should be released when either we are done using it in OpenSslKeyMaterialProvider or when the cache is destroyed. 

So what are you suggesting us to do here ?

handler/src/main/java/io/netty/handler/ssl/OpenSslX509KeyManagerFactory.java
normanmaurer: this was actually a bug... We should return `null` from `OpenSslPrivateKey#getFormat` as we do not support encoding.

handler/src/main/java/io/netty/handler/ssl/OpenSslPrivateKey.java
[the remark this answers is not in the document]
Scottmitch: > So what are you suggesting us to do here ?

Adding a `volatile` here doesn't seem to solve the race conditions involved with potential use of this API. I'm wondering if its needed at all.

handler/src/main/java/io/netty/handler/ssl/OpenSslPrivateKey.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    @Override
+    public String getAlgorithm() {
+        return null;
+    }
+
+    @Override
+    public String getFormat() {
+        return "native";
```
Scottmitch: https://github.com/netty/netty/pull/8200#discussion_r210673926 ... IIUC you meant to make this `null`?

normanmaurer:
Verified this works ... writing a test is not really easy tho :(

Issue #7272, opened 2017-10-02
If using more than one shaded netty-tcnative library in a JVM, only the first loads successfully


steveniemitz:
Now that netty can automatically detect the package prefix ( #6995 ), I attempted to shade netty into two different packages, lets call them `A` and `B`.

Package `A` loaded tcnative successfully (lets say it was shaded to `A.repackaged.io.netty`).

However, when package `B` attempted to load, I got an `UnsatisfiedLinkError` invoking `aprMajorVersion()` in `io.netty.internal.tcnative.Library`.

Digging in more, I do see the native library (from package `B`) being unpacked and loadLibrary being called on it (and not throwing an exception).  However, `lldb` reports that it is not loaded (eg it doesn't show up in `image list`, only the library from package `A` does), and further, running the JVM with `-verbose:jni` shows the JNI native methods for the shaded version of tcnative in package `A` being registered (again) instead of `B`. ex:

`[Registering JNI native method A.repackaged.io.netty.internal.tcnative.Library.initialize0]`

I've tested this on OSX, but not Linux yet.

johnou:
cc @ejona86

normanmaurer:
also @carl-mastrangelo

carl-mastrangelo:
@steveniemitz to clarify, you mean the *shaded* version of `io.netty.internal.tcnative.Library` right?  At leats on OpenJDK, most of the interesting Native library loading logic is in ClassLoader.  Have you tried stepping through it in a debugger yet?

steveniemitz:
Sorry, yeah my post was confusing there.  I do mean the shaded version of `Library`.

Working from memory right now, I have stepped through.  The call to (native) `System.loadLibrary` succeeds, and it does pass the expected correct shaded lib name.  I haven't stepped though the native code though.

Xeli:
Sorry to add noise to this issue, but how did you manage to shade netty with tcnative at all? I run into the problem that netty cannot load the native library (for details: [)](https://stackoverflow.com/questions/46650776/after-shading-io-netty-tcnative-stops-working)

ejona86:
@Xeli, you also need to rename the .so's, .jnilib, and .dll. See #6995 for some more information.

normanmaurer:
@steveniemitz so can you please clarify what exactly the issue is ?

steveniemitz:
Sorry,  I'm not sure how I could be more descriptive,  what is unclear here?

Scottmitch:
[...]

PR #8207, opened 2018-08-20
Ensure multiple shaded version of the same netty artifact can be load…

normanmaurer:
…ed as long as the shaded prefix is different

Motivation:

We should support to load multiple shaded versions of the same netty artifact as netty is often used in multiple dependencies.

This is related to https://github.com/netty/netty/issues/7272.

Modifications:

- Use -fvisibility=hidden when compiling and use __attribute__((visibility("default"))) for things we really want to have exported
- Ensure fields are declared as static so these are not exported
- Adjust testsuite-shading to use install_name_tool on MacOS to change the id of the lib. Otherwise the wrong may be used.

Result:

Be able to use multiple shaded versions of the same netty artifact.

### DIFF
--- a/testsuite-shading/src/test/java/io/netty/testsuite/shading/ShadingIT.java
+++ b/testsuite-shading/src/test/java/io/netty/testsuite/shading/ShadingIT.java
@@ -23,19 +23,26 @@ import java.lang.reflect.Method;
 
 public class ShadingIT {
 
+    private static final String SHADING_PREFIX = System.getProperty("shadingPrefix2");
+    private static final String SHADING_PREFIX2 = System.getProperty("shadingPrefix");
+
     @Test
     public void testShadingNativeTransport() throws Exception {
-        testShading0(PlatformDependent.isOsx() ? "io.netty.channel.kqueue.KQueue" : "io.netty.channel.epoll.Epoll");
+        String className = PlatformDependent.isOsx() ?
+                "io.netty.channel.kqueue.KQueue" : "io.netty.channel.epoll.Epoll";
+        testShading0(SHADING_PREFIX, className);
+        testShading0(SHADING_PREFIX2, className);
     }
 
     @Ignore("Figure out why this sometimes fail on the CI")
     @Test
     public void testShadingTcnative() throws Exception {
-        testShading0("io.netty.handler.ssl.OpenSsl");
+        String className = "io.netty.handler.ssl.OpenSsl";
+        testShading0(SHADING_PREFIX, className);
+        testShading0(SHADING_PREFIX2, className);
     }
 
-    private static void testShading0(String classname) throws Exception {
-        String shadingPrefix = System.getProperty("shadingPrefix");
+    private static void testShading0(String shadingPrefix, String classname) throws Exception {
         final Class<?> clazz = Class.forName(shadingPrefix + '.' + classname);
         Method method = clazz.getMethod("ensureAvailability");
         method.invoke(null);
--- a/transport-native-epoll/src/main/c/netty_epoll_native.c
+++ b/transport-native-epoll/src/main/c/netty_epoll_native.c
@@ -65,11 +65,11 @@ struct mmsghdr {
 #endif
 
 // Those are initialized in the init(...) method and cached for performance reasons
-jfieldID packetAddrFieldId = NULL;
-jfieldID packetScopeIdFieldId = NULL;
-jfieldID packetPortFieldId = NULL;
-jfieldID packetMemoryAddressFieldId = NULL;
-jfieldID packetCountFieldId = NULL;
+static jfieldID packetAddrFieldId = NULL;
+static jfieldID packetScopeIdFieldId = NULL;
+static jfieldID packetPortFieldId = NULL;
+static jfieldID packetMemoryAddressFieldId = NULL;
+static jfieldID packetCountFieldId = NULL;
 
 // util methods
 static int getSysctlValue(const char * property, int* returnValue) {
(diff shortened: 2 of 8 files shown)


transport-native-unix-common/src/main/c/netty_unix_errors.c
```diff
@@ -37,42 +37,42 @@ static char* exceptionMessage(char* msg, int error) {
 }
 
 // Exported C methods
```
Scottmitch: Can we clarify the terminology here? They are no longer exported from a "C" perspective ... they are more now strictly "Exported JNI methods, but not exported C methods".

transport-native-epoll/src/main/c/netty_epoll_linuxsocket.c
```diff
@@ -481,7 +481,7 @@ jint netty_epoll_linuxsocket_JNI_OnLoad(JNIEnv* env, const char* packagePrefix)
     return NETTY_JNI_VERSION;
 }
 
-void netty_epoll_linuxsocket_JNI_OnUnLoad(JNIEnv* env) {
+void __attribute__((visibility("hidden"))) netty_epoll_linuxsocket_JNI_OnUnLoad(JNIEnv* env) {
```
Scottmitch: Alternatively it looks like you can make the default visibility "hidden" via a compiler flag `-fvisibility` [1][2]. The gcc wiki even recommends using `hidden` as the default and then explicitly adding `__attribute__ ((visibility ("default")))` when you need to export symbols ... did you consider this approach?

[1] https://gcc.gnu.org/wiki/Visibility
[2] https://clang.llvm.org/docs/LTOVisibility.html

transport-native-unix-common/src/main/c/netty_unix_socket.c
```diff
@@ -229,7 +229,7 @@ static jint _socket(JNIEnv* env, jclass clazz, int type) {
     return fd;
 }
 
-int netty_unix_socket_initSockaddr(JNIEnv* env, jbyteArray address, jint scopeId, jint jport,
+int __attribute__((visibility("hidden"))) netty_unix_socket_initSockaddr(JNIEnv* env, jbyteArray address, jint scopeId, jint jport,
```
Scottmitch: consider grouping all the "exported" methods together in the file, and surrounding with a "Exported JNI methods ..." comment

testsuite-shading/src/test/java/io/netty/testsuite/shading/ShadingIT.java
```diff
@@ -25,17 +25,20 @@
...
     @Test
     public void testShadingNativeTransport() throws Exception {
-        testShading0(PlatformDependent.isOsx() ? "io.netty.channel.kqueue.KQueue" : "io.netty.channel.epoll.Epoll");
+        testShading0(System.getProperty("shadingPrefix2"),
+                     PlatformDependent.isOsx() ? "io.netty.channel.kqueue.KQueue" : "io.netty.channel.epoll.Epoll");
+        testShading0(System.getProperty("shadingPrefix"),
+                     PlatformDependent.isOsx() ? "io.netty.channel.kqueue.KQueue" : "io.netty.channel.epoll.Epoll");
     }
 
     @Ignore("Figure out why this sometimes fail on the CI")
```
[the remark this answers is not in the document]
normanmaurer: Yep

transport-native-epoll/src/main/c/netty_epoll_linuxsocket.c
normanmaurer: I can try this... will see how to do something similar with windows for tcnative later if it works

transport-native-unix-common/src/main/c/netty_unix_socket.c
normanmaurer: Actually these are not even exported JNI methods. They are just exported c methods within the library

transport-native-unix-common/src/main/c/netty_unix_errors.c
normanmaurer: See https://github.com/netty/netty/pull/8207#discussion_r211333017

normanmaurer:
@carl-mastrangelo `nm -gU /path/to/libname.dylib` ... This print all exported which listed a lot of stuff that could cause strange side-effects if picked up ;)

normanmaurer:
Also for the record I am working on something similar for netty-tcnative atm :)

normanmaurer:
@carl-mastrangelo @Scottmitch PTAL again... I am just using JNIEXPORT now to export the stuff we need which will do the correct thing depending in the compiler 🎉

transport-native-epoll/src/main/c/netty_epoll_native.c
```diff
@@ -545,25 +545,25 @@ static void JNI_OnUnload_netty_transport_native_epoll0(JavaVM* vm, void* reserve
...
 }
 
-// As we build with -fvisibility=hidden we need to ensure we mark the entry load and unload functions used by the
-// JVM as visible.
+// We build with -fvisibility=hidden so ensure we mark everything that needs to be visible with JNIEXPORT
+// http://mail.openjdk.java.net/pipermail/core-libs-dev/2013-February/014549.html
 
 // Invoked by the JVM when statically linked
-jint __attribute__((visibility("default"))) JNI_OnLoad_netty_transport_native_epoll(JavaVM* vm, void* reserved) {
+JNIEXPORT jint JNI_OnLoad_netty_transport_native_epoll(JavaVM* vm, void* reserved) {
```
carl-mastrangelo: When I approved earlier I thought this was due to wanting the explicit behavior of JNIEXPORT on linux/solaris.  However it seems now that the windows behavior (which would obviously not work here, but bear with me)  is now:

```c
#define JNIEXPORT __declspec(dllexport)
#define JNIIMPORT __declspec(dllimport)
```

Since you'll be making this for tcnative, are you sure this is what you want?

transport-native-unix-common/src/main/c/netty_unix_socket.c
Scottmitch: From the perspective of a native library I would say "the symbols are not exported, but you can always refer to them via function pointers". In this context we are referring to the function pointers to register/export the symbols with JNI.

ejona86:
@normanmaurer, don't forget to update the commit message now that you swapped to JNIEXPORT.

transport-native-epoll/src/main/c/netty_epoll_native.c
normanmaurer: Yep that is exactly what I want 😉 Will verify what is exported on windows as well tho

PR #8210, opened 2018-08-21
Try to monkey-patch library id when shading is used and we are on Mac…

normanmaurer:
…OS / OSX.

Motivation:

https://github.com/netty/netty/commit/ea4c315b45bf1bfa82cd75a7d3f38bff684e2ebc did ensure we support using multiple versions of the same shaded native library but the user still needed to run install_name_tool -id on MacOS to ensure the ID is unique.
This is kind of error prone and also means that the shading itself would need to be done on MacOS / OSX.

This is related to https://github.com/netty/netty/issues/7272.

Modifications:

- Monkey patch the shaded native lib on MacOS to ensure the id is unique while unpack it to the tempory location.

Result:

Easier way of using shaded native libs in netty.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
+++ b/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
@@ -50,6 +50,10 @@ public final class NativeLibraryLoader {
     private static final boolean DELETE_NATIVE_LIB_AFTER_LOADING;
     private static final boolean TRY_TO_PATCH_SHADED_ID;
 
+    // Just use a-Z and numbers as valid ID bytes.
+    private static final byte[] UNIQUE_ID_BYTES =
+            "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ".getBytes(CharsetUtil.US_ASCII);
+
     static {
         String workdir = SystemPropertyUtil.get("io.netty.native.workdir");
         if (workdir != null) {
@@ -243,7 +247,7 @@ public final class NativeLibraryLoader {
     }
 
     /**
-     * Try to patch shaded library to ensure it uses an unique id.
+     * Try to patch shaded library to ensure it uses a unique id.
      */
     private static void patchShadedLibraryId(byte[] bytes, String originalName, String name) {
         // Our native libs always have the name as part of their id so we can search for it and replace it
@@ -276,8 +280,10 @@ public final class NativeLibraryLoader {
             logger.debug(
                     "Found the id of the shaded native library {}, adjusting it to something unique.", name);
 
-            PlatformDependent.threadLocalRandom().nextBytes(nameBytes);
-            System.arraycopy(nameBytes, 0, bytes, idIdx, nameBytes.length);
+            // We should only use bytes as replacement that are in our UNIQUE_ID_BYTES array.
+            for (int i = 0; i < nameBytes.length; i++) {
+                bytes[idIdx++] = UNIQUE_ID_BYTES[PlatformDependent.threadLocalRandom().nextInt(UNIQUE_ID_BYTES.length)];
+            }
         }
     }
 
--- a/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
+++ b/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
@@ -247,17 +247,17 @@ public final class NativeLibraryLoader {
     }
 
     /**
-     * Try to patch shaded library to ensure it uses a unique id.
+     * Try to patch shaded library to ensure it uses a unique ID.
      */
     private static void patchShadedLibraryId(byte[] bytes, String originalName, String name) {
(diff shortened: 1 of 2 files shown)


common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -174,16 +176,32 @@ public static void load(String originalName, ClassLoader loader) {
 
             byte[] buffer = new byte[8192];
             int length;
-            while ((length = in.read(buffer)) > 0) {
-                out.write(buffer, 0, length);
+            if (PlatformDependent.isOsx() && !packagePrefix.isEmpty()) {
+                // We read the whole native lib into memory to make it easier to monkey-patch the id.
+                ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream(in.available());
```
trustin: Could use Math.max() just in case in.available() returns a very small value.

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -218,6 +236,45 @@ public static void load(String originalName, ClassLoader loader) {
         }
     }
 
+    /**
+     * Try to patch shaded library to ensure it uses an unique id.
```
trustin: a unique ID

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -218,6 +242,45 @@ public static void load(String originalName, ClassLoader loader) {
...
+        }
+
+        if (idIdx == -1) {
+            logger.debug("Was not able to find the id of the shaded native library {}, can't adjust it.", name);
+        } else {
+            // we found our id... now monkey-patch it!
+            logger.debug(
+                    "Found the id of the shaded native library {}, adjusting it to something unique.", name);
+
+            PlatformDependent.threadLocalRandom().nextBytes(nameBytes);
```
trustin: Perhaps we should limit the characters to alphabets and numbers and log the new name just in case it can be used for debugging? We might also want to preserve the first few characters?

normanmaurer: @trustin this is actually only part of the id as the id usually contains the full path of the native lib when it was compiled. So it does not really make sense to log. That said I think limit to numbers + alphabets is a good idea.

normanmaurer:
@ejona86 thats my man 🤣

normanmaurer:
Also just FYI I tested this of course :)

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -47,6 +48,11 @@
     private static final String NATIVE_RESOURCE_HOME = "META-INF/native/";
     private static final File WORKDIR;
     private static final boolean DELETE_NATIVE_LIB_AFTER_LOADING;
+    private static final boolean TRY_TO_PATCH_SHADED_ID;
+
+    // Just use a-Z and numbers as valid ID bytes.
+    private static final byte[] UNIQUE_ID_BYTES =
+            "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ".getBytes(CharsetUtil.US_ASCII);
```
carl-mastrangelo: You should probably just used a uuid.  Google internally has >1500 instances of this exact string when I searched our code.

normanmaurer: I mean does it really matter as we use a random to pick the bytes then

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
```diff
@@ -174,16 +186,32 @@ public static void load(String originalName, ClassLoader loader) {
...
+            if (TRY_TO_PATCH_SHADED_ID && PlatformDependent.isOsx() && !packagePrefix.isEmpty()) {
+                // We read the whole native lib into memory to make it easier to monkey-patch the id.
+                ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream(in.available());
+
+                while ((length = in.read(buffer)) > 0) {
+                    byteArrayOutputStream.write(buffer, 0, length);
+                }
+                byteArrayOutputStream.flush();
+                byte[] bytes = byteArrayOutputStream.toByteArray();
+                byteArrayOutputStream.close();
```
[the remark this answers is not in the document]
normanmaurer: @njhill yeah .. that said I think I would just keep it as it cant hurt and is at least consistent with the other stream usage.

common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java
johnou: @carl-mastrangelo running OS X? :grin:

normanmaurer: @carl-mastrangelo I will merge this one... We can do a followup if needed.

Issue #8208, opened 2018-08-20
Support hosts file encoded in unicode format


ramtech123:
### Expected behavior
Netty should be able to resolve IP address from hosts file when the file is encoded in unicode format.

Java is able to resolve hostname from unicode encoded hosts file. Disabling Vertx (Netty) hostname resolution using `-Dvertx.disableDnsResolver=true` flag in our application was the workaround that we used, until we finally identified root cause.

### Actual behavior
`HostsFileParser.parseSilently()` returns `HostsFileEntries` instance with zero entries if the hosts file is encoded in unicode format.

### Steps to reproduce
Open hosts file with text editor (Eg. Windows Notepad) as Administrator.
Save hosts file with unicode encoding.
Run the reproducer and observe the console output.

### Minimal yet complete reproducer code (or URL to code)
https://github.com/ramtech123/pocs/tree/master/netty-hosts-reproducer

### Netty version
4.1

### JVM version (e.g. `java -version`)
Java 1.8

### OS version (e.g. `uname -a`)
Tested with Windows10, Windows Server 2012, 2016

normanmaurer:
@ramtech123 thanks.. Check https://github.com/netty/netty/pull/8211. This allows you to specify the charset to use while decoding.

ejona86:
Based on my own testing with Windows Server 2016, it seems like Windows auto-detects the encoding. I tried with ANSI (ASCII), Unicode (UTF-16), and UTF-8 and they all worked. I saved a trash hostname in `hosts` (different for each test) and then did a ping. Ping noticed my trash hostname configuration each time.

ejona86:
https://serverfault.com/a/452269 seems to think that the encoding should be UTF-8. I wonder if Windows became more permissive in a later version of Windows. I wasn't able to find any official documentation as to the encoding.

normanmaurer:
@ejona86 I mean what we could do is test UTF-8, ASCII, Unicode (in this order) until we were able to parse something when on windows... WDYT ?

ramtech123:
@ejona86 , I also could not find any standards around hosts file encoding. The comment here https://serverfault.com/questions/926706/is-there-any-ietf-rfc-standard-defined-for-hosts-file-encoding indicates it never had a formal standard.

@normanmaurer just curious about "...until we were able to parse something when on windows" - don't we expect this situation in non-Windows machine?

normanmaurer:
@ramtech123 I never heard of using something linux unicode for `/etc/hosts`.

[...]

PR #8211, opened 2018-08-21
Allow to parse hosts file which is stored in a different encoding the…

normanmaurer:
…n the default system encoding.

Motivation:

We should support to parse and read a hosts file which is stored in a different encoding then the system default. Beside this when we are on windows we should just try to parse it with multiple different charset before giving up as there is no real standard what charset to use.

Modifications:

- Add more method overloads to HostsFileParser that take a Charset.
- Try to parse with multiple Charsets in DefaultHostsFileEntriesResolver when windows is used.
- Add unit test

Result:

Fixes https://github.com/netty/netty/issues/8208.


normanmaurer:
@ramtech123 FYI

normanmaurer:
@ejona86 @ramtech123 PTAL again

ramtech123:
@normanmaurer I verified the behavior with above changes and it is looking good 👍. Thanks

Issue #8220, opened 2018-08-24
Recycler will produce npe error when multiple recycled at different thread


zhaojigang:
Netty version: 4.1.29.Final-SNAPSHOT

    public void testMultipleRecycleAtDifferentThread() {
        Recycler<HandledObject> recycler = newRecycler(1024);
        final HandledObject object = recycler.get();
        final Thread thread1 = new Thread(new Runnable() {
            @Override
            public void run() {
                object.recycle();  // stack=null
            }
        });
        thread1.start();
        try {
            thread1.join();
        } catch (InterruptedException e) {
        }

        final Thread thread2 = new Thread(new Runnable() {
            @Override
            public void run() {
                object.recycle();  //  stack npe 
            }
        });
        thread2.start();
        try {
            thread2.join();
        } catch (InterruptedException e) {
        }
    }

PR #8222, opened 2018-08-24
fix recycler npe bug

zhaojigang:
Fix https://github.com/netty/netty/issues/8220
I add a unitTest in RecyclerTest#testMultipleRecycleAtDifferentThread()


zhaojigang:
I am so sorry that unitTest occur errors, I will add a newPullRequest

PR #8224, opened 2018-08-24
Recycler will produce npe error when multiple recycled at different thread

zhaojigang:
fix https://github.com/netty/netty/issues/8220

### DIFF
--- a/common/src/test/java/io/netty/util/RecyclerTest.java
+++ b/common/src/test/java/io/netty/util/RecyclerTest.java
@@ -81,10 +81,11 @@ public class RecyclerTest {
         object.recycle();
     }
 
-    @Test
-    public void testMultipleRecycleAtDifferentThread() {
+    @Test(expected = IllegalStateException.class)
+    public void testMultipleRecycleAtDifferentThread() throws InterruptedException {
         Recycler<HandledObject> recycler = newRecycler(1024);
         final HandledObject object = recycler.get();
+        final AtomicReference<IllegalStateException> exceptionStore = new AtomicReference<IllegalStateException>();
         final Thread thread1 = new Thread(new Runnable() {
             @Override
             public void run() {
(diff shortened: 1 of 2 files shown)


common/src/test/java/io/netty/util/RecyclerTest.java
```diff
@@ -81,6 +81,38 @@ public void testMultipleRecycle() {
...
+        } catch (InterruptedException e) {
+        }
+
+        final Thread thread2 = new Thread(new Runnable() {
+            @Override
+            public void run() {
+                try {
+                    object.recycle();
+                } catch (IllegalStateException e) {
+                }
```
normanmaurer: store the `IllegalStateException` in an `AtomicReference` and assert it after `thread2.join()` returns.

common/src/test/java/io/netty/util/RecyclerTest.java
```diff
@@ -81,6 +81,38 @@ public void testMultipleRecycle() {
...
+        final Thread thread1 = new Thread(new Runnable() {
+            @Override
+            public void run() {
+                object.recycle();
+            }
+        });
+        thread1.start();
+        try {
+            thread1.join();
+        } catch (InterruptedException e) {
```
normanmaurer: no need to catch this... just let it throw.

common/src/test/java/io/netty/util/RecyclerTest.java
```diff
@@ -81,6 +81,38 @@ public void testMultipleRecycle() {
...
+                    object.recycle();
+                } catch (IllegalStateException e) {
+                }
+            }
+        });
+        thread2.start();
+        try {
+            thread2.join();
+        } catch (InterruptedException e) {
+        }
```
normanmaurer: no need to catch this... just let it throw.

normanmaurer:
Also please fix the check style error : https://garage.netty.io/teamcity/viewLog.html?buildId=32568&buildTypeId=pulls_netty&guest=1

zhaojigang:
ok, thanks!!! I will do it now!

Issue #8165, opened 2018-08-02
TCN is failing with error:14179044:SSL routines:tls_construct_server_key_exchange:internal error


rkapsi:
We're seeing some ciphers such as `TLS_DHE_RSA_WITH_AES_128_GCM_SHA256` (a.k.a. `DHE-RSA-AES128-GCM-SHA256`) failing with TCN 2.0.12.

Here is a very crude repro.

```java
public static void main(String[] args) throws Exception {
    String cipher = "DHE-RSA-AES128-GCM-SHA256";
    
    SelfSignedCertificate cert = new SelfSignedCertificate();
    
    ChannelHandler handler = new ChannelInitializer<Channel>() {
      @Override
      protected void initChannel(Channel ch) throws Exception {
        ChannelPipeline pipeline = ch.pipeline();
        
        SslContext sslContext = SslContextBuilder.forServer(cert.certificate(), cert.privateKey())
          .sslProvider(SslProvider.OPENSSL)
          .ciphers(Arrays.asList(cipher))
          .build();
        
        pipeline.addLast(sslContext.newHandler(ch.alloc()));
        pipeline.addLast(new ChannelInboundHandlerAdapter() {
          @Override
          public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
            if (evt instanceof SslHandshakeCompletionEvent) {
              SslHandshakeCompletionEvent event = (SslHandshakeCompletionEvent)evt;
              if (!event.isSuccess()) {
                Throwable cause = event.cause();
                if (cause instanceof SSLHandshakeException) {
                  cause.printStackTrace();
                }
              }
            }
            
            ctx.fireUserEventTriggered(evt);
          }
          
          @Override
          public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
            cause.printStackTrace();
          }
        });
      }
    };
    
    ServerBootstrap bootstrap = new ServerBootstrap()
        .channel(NioServerSocketChannel.class)
        .group(new NioEventLoopGroup())
        .childHandler(handler);
    
    Channel channel = bootstrap.bind(0).sync().channel();
    InetSocketAddress isa = (InetSocketAddress)channel.localAddress();
    
    System.out.println("Server: " + OpenSsl.versionString());
    System.out.println(cipher + " is supported: " + OpenSsl.isCipherSuiteAvailable(cipher));
    
[...]

PR #8225, opened 2018-08-24
An unit test to ensure that cipher suites don't break/disappear betwe…

rkapsi:
…en releases.

Motivation

Ensure classes of cipher suites continue working between releases. Adding just a DHE check for now as it caused #8165 but this test can be expaned to other suites.

Modifications

Adding an unit test that checks for the presence of a cipher suite.

Result

Prevent #8165 from happening in the future.

### DIFF
--- a/handler/src/test/java/io/netty/handler/ssl/CipherSuiteCanaryTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/CipherSuiteCanaryTest.java
@@ -1,0 +1,257 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License, version
+ * 2.0 (the "License"); you may not use this file except in compliance with the
+ * License. You may obtain a copy of the License at:
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.netty.handler.ssl;
+
+import io.netty.bootstrap.Bootstrap;
+import io.netty.bootstrap.ServerBootstrap;
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandler;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.channel.ChannelInitializer;
+import io.netty.channel.ChannelPipeline;
+import io.netty.channel.DefaultEventLoopGroup;
+import io.netty.channel.EventLoopGroup;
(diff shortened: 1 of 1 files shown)


handler/src/test/java/io/netty/handler/ssl/CipherSuiteCanaryTest.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    @Test
+    public void testDiffieHellmanKeyExchange() throws Exception {
+        usingCipher("DHE-RSA-AES128-GCM-SHA256");
+    }
+
+    private static void usingCipher(final String cipher) throws Exception {
+        Assume.assumeTrue(OpenSsl.isAvailable());
+        assertTrue(OpenSsl.isCipherSuiteAvailable(cipher));
```
normanmaurer: I think this should be an `assumeTrue` as we also run our testsuite with `BoringSSL` for example

handler/src/test/java/io/netty/handler/ssl/CipherSuiteCanaryTest.java
```diff
@@ -0,0 +1,220 @@
...
+                                }
+                            });
+                        }
+                    };
+
+                    Channel client = client(server, clientHandler);
+                    try {
+                        client.writeAndFlush(request.slice());
+
+                        if (!clientPromise.await(5L, TimeUnit.SECONDS)) {
```
normanmaurer: just use `assertTrue(...)` for all of these ?

handler/src/test/java/io/netty/handler/ssl/CipherSuiteCanaryTest.java
```diff
@@ -0,0 +1,220 @@
...
+                            fail("This shouldn't have failed!");
+                        }
+
+                        if (!serverPromise.await(5L, TimeUnit.SECONDS)) {
+                            fail("This shouldn't have failed!");
+                        }
+
+                        if (!clientPromise.isSuccess()) {
```
normanmaurer: just use `clientPromise.sync()` which will rethrow ?

handler/src/test/java/io/netty/handler/ssl/CipherSuiteCanaryTest.java
```diff
@@ -0,0 +1,220 @@
...
+                            throw new IllegalStateException(clientPromise.cause());
+                        }
+
+                        if (!serverPromise.isSuccess()) {
```
normanmaurer: just use `clientPromise.sync()` which will rethrow ?

handler/src/test/java/io/netty/handler/ssl/CipherSuiteCanaryTest.java
```diff
@@ -0,0 +1,220 @@
...
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+/**
+ * The purpose of this unit test is to act as a canary and catch changes in supported cipher suites.
+ */
+public class CipherSuiteCanaryTest {
```
normanmaurer: I wonder if we can use `@Paramized` and use all the different `SslProvider` to test against each other.

normanmaurer:
@rkapsi can you please fix the unit test ?

Issue #8229, opened 2018-08-27
transport-native-* should be usable when sun.misc.Unsafe can not be used.


normanmaurer:
As JDK11 is on its way (which will be the LTS release) we should ensure that our native transports can be used even when sun.misc.Unsafe is not present.

PR #8231, opened 2018-08-28
Allow to use native transports when sun.misc.Unsafe is not present on…

normanmaurer:
… the system

Motivation:

We should be able to use the native transports (epoll / kqueue) even when sun.misc.Unsafe is not present on the system. 

Modifications:

- Correctly disable usage of sun.misc.Unsafe when -PnoUnsafe is used while running the build
- Correctly increment metric when UnpooledDirectByteBuf is allocated. This was uncovered once -PnoUnsafe usage was fixed.
- Implement fallbacks in all our native transport code for when sun.misc.Unsafe is not present.

Result:

Fixes https://github.com/netty/netty/issues/8229.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/UnpooledDirectByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/UnpooledDirectByteBuf.java
@@ -64,7 +64,7 @@ public class UnpooledDirectByteBuf extends AbstractReferenceCountedByteBuf {
         }
 
         this.alloc = alloc;
-        setByteBuffer(ByteBuffer.allocateDirect(initialCapacity));
+        setByteBuffer(allocateDirect(initialCapacity));
     }
 
     /**
--- a/transport-native-epoll/src/main/c/netty_epoll_native.c
+++ b/transport-native-epoll/src/main/c/netty_epoll_native.c
@@ -38,6 +38,7 @@
 #include <time.h>
 
 #include "netty_epoll_linuxsocket.h"
+#include "netty_unix_buffer.h"
 #include "netty_unix_errors.h"
 #include "netty_unix_filedescriptor.h"
 #include "netty_unix_jni.h"
@@ -452,6 +453,9 @@ static jint netty_epoll_native_JNI_OnLoad(JNIEnv* env, const char* packagePrefix
     if (netty_unix_socket_JNI_OnLoad(env, packagePrefix) == JNI_ERR) {
         return JNI_ERR;
     }
+    if (netty_unix_buffer_JNI_OnLoad(env, packagePrefix) == JNI_ERR) {
+        return JNI_ERR;
+    }
     if (netty_epoll_linuxsocket_JNI_OnLoad(env, packagePrefix) == JNI_ERR) {
         return JNI_ERR;
     }
@@ -501,6 +505,7 @@ static void netty_epoll_native_JNI_OnUnLoad(JNIEnv* env) {
     netty_unix_errors_JNI_OnUnLoad(env);
     netty_unix_filedescriptor_JNI_OnUnLoad(env);
     netty_unix_socket_JNI_OnUnLoad(env);
+    netty_unix_buffer_JNI_OnUnLoad(env);
     netty_epoll_linuxsocket_JNI_OnUnLoad(env);
 }
 
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventArray.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventArray.java
@@ -15,7 +15,10 @@
  */
 package io.netty.channel.epoll;
 
+import io.netty.channel.unix.Buffer;
 import io.netty.util.internal.PlatformDependent;
+
+import java.nio.ByteBuffer;
 
 /**
  * This is an internal datastructure which can be directly passed to epoll_wait to reduce the overhead.
@@ -41,6 +44,7 @@ final class EpollEventArray {
     // The offsiet of the data union in the epoll_event struct
     private static final int EPOLL_DATA_OFFSET = Native.offsetofEpollData();
 
+    private ByteBuffer memory;
     private long memoryAddress;
     private int length;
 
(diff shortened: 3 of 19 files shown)


transport-native-unix-common/src/main/java/io/netty/channel/unix/PreferredDirectByteBufAllocator.java
```diff
@@ -0,0 +1,130 @@
...
+ */
+package io.netty.channel.unix;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufAllocator;
+import io.netty.buffer.CompositeByteBuf;
+import io.netty.util.internal.UnstableApi;
+
+@UnstableApi
+public final class PreferredDirectByteBufAllocator implements ByteBufAllocator {
```
normanmaurer: This was needed as otherwise we may fail to read data if a `ByteBufAllocator` is used which returns heap buffers from `ioBuffer(...)` methods.

buffer/src/main/java/io/netty/buffer/UnpooledDirectByteBuf.java
```diff
@@ -64,7 +64,7 @@ public UnpooledDirectByteBuf(ByteBufAllocator alloc, int initialCapacity, int ma
         }
 
         this.alloc = alloc;
-        setByteBuffer(ByteBuffer.allocateDirect(initialCapacity));
+        setByteBuffer(allocateDirect(initialCapacity));
```
normanmaurer: This was needed to invoke the correct overridden method.

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventArray.java
```diff
@@ -77,28 +78,42 @@ int length() {
...
     /**
      * Return the file descriptor for the {@code epoll_event} on this index.
      */
     int fd(int index) {
-        return PlatformDependent.getInt(memoryAddress + index * EPOLL_EVENT_SIZE + EPOLL_DATA_OFFSET);
+        return getInt(index, EPOLL_DATA_OFFSET);
+    }
+
+    private int getInt(int index, int offset) {
+        if (PlatformDependent.hasUnsafe()) {
```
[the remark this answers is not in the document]
normanmaurer: yeah performance... This one is still a bit faster .

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/KQueueEventArray.java
```diff
@@ -89,52 +94,77 @@ private void checkSize() {
     void realloc(boolean throwIfFail) {
         // Double the capacity while it is "sufficiently small", and otherwise increase by 50%.
         int newLength = capacity <= 65536 ? capacity << 1 : capacity + capacity >> 1;
-        long newMemoryAddress = PlatformDependent.reallocateMemory(memoryAddress, newLength * KQUEUE_EVENT_SIZE);
```
ejona86: Just to confirm, we didn't need to copy the old contents of the memory like was done previously?

normanmaurer:
Also once this is merged we should ensure we run a CI job once a day with "-PnoUnsafe"

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/NativeLongArray.java
```diff
@@ -71,12 +89,8 @@ private void checkSize() {
     private void realloc() {
         // Double the capacity while it is "sufficiently small", and otherwise increase by 50%.
         int newLength = capacity <= 65536 ? capacity << 1 : capacity + capacity >> 1;
-        long newMemoryAddress = PlatformDependent.reallocateMemory(memoryAddress, newLength * SIZEOF_JLONG);
```
ejona86: Ditto here.

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/KQueueEventArray.java
normanmaurer: Good point... let me copy.

normanmaurer:
@ejona86 fixed and pushed (but for whatever reason it is not shown in the GitHub UI yet).

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/KQueueEventArray.java
normanmaurer: fixed

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/NativeLongArray.java
```diff
@@ -28,13 +32,27 @@
...
     void add(long value) {
-        checkSize();
-        PlatformDependent.putLong(memoryOffset(size++), value);
+        reallocIfNeeded();
+        if (PlatformDependent.hasUnsafe()) {
+            PlatformDependent.putLong(memoryOffset(size), value);
+        } else {
+            memory.putLong(idx(size), value);
+        }
+        ++size;
```
carl-mastrangelo: is this size the number of longs, or the amount of memory?  Below, capacity is compared to it, but thats in units of bytes, not longs.

normanmaurer: It’s the number of longs stored. I am not confused by your comment tho as capacity is also not bytes but the max number of longs we can hold. Can you point me to the place that is incorrect ?

transport-native-epoll/src/main/c/netty_epoll_native.c
```diff
@@ -452,6 +453,9 @@ static jint netty_epoll_native_JNI_OnLoad(JNIEnv* env, const char* packagePrefix
     if (netty_unix_socket_JNI_OnLoad(env, packagePrefix) == JNI_ERR) {
         return JNI_ERR;
     }
+    if (netty_unix_buffer_JNI_OnLoad(env, packagePrefix) == JNI_ERR) {
```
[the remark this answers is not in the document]
normanmaurer: I can do a follow up after this PR to fix all of it.... good idea

PR #8234, opened 2018-08-29
We should be able to use the ByteBuffer cleaner on java8 (and earlier…

normanmaurer:
… versions) even if sun.misc.Unsafe is not present.

Motivation:

At the moment we have a hard dependency on sun.misc.Unsafe to use the Cleaner on Java8 and earlier. This is not really needed as we can still use pure reflection if sun.misc.Unsafe is not present.

Modifications:

Refactor Cleaner6 to fallback to pure reflection if sun.misc.Unsafe is not present on system.

Result:

More timely releasing of direct memory on Java8 and earlier when sun.misc.Unsafe is not present.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/CleanerJava6.java
+++ b/common/src/main/java/io/netty/util/internal/CleanerJava6.java
@@ -21,6 +21,8 @@ import io.netty.util.internal.logging.InternalLoggerFactory;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.nio.ByteBuffer;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 
 
 /**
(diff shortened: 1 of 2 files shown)


normanmaurer:
I also added the missing SecurityManager stuff.

common/src/main/java/io/netty/util/internal/CleanerJava6.java
```diff
@@ -32,55 +34,112 @@
...
-                clean = cleaner.getClass().getDeclaredMethod("clean");
-                clean.invoke(cleaner);
-            } catch (Throwable t) {
-                // We don't have ByteBuffer.cleaner().
+                cleaner = PlatformDependent0.getObject(direct, fieldOffset);
+            } else {
                 fieldOffset = -1;
-                clean = null;
-                error = t;
+                cleanerField.setAccessible(true);
```
ejona86: Should this also go through the AccessController?

common/src/main/java/io/netty/util/internal/CleanerJava6.java
```diff
@@ -32,55 +34,112 @@
...
             logger.debug("java.nio.ByteBuffer.cleaner(): unavailable", error);
         }
+        CLEANER_FIELD = cleanerField;
         CLEANER_FIELD_OFFSET = fieldOffset;
         CLEAN_METHOD = clean;
     }
 
     static boolean isSupported() {
-        return CLEANER_FIELD_OFFSET != -1;
+        return CLEANER_FIELD_OFFSET != -1 || CLEANER_FIELD != null;
```
ejona86: Note: CLEANER_FIELD != null even if using unsafe.

normanmaurer: yeah sure but does it matter ?

normanmaurer:
@ejona86 PTAL again

common/src/main/java/io/netty/util/internal/CleanerJava6.java
ejona86: I'd agree it doesn't really matter. But I just wanted to point it out.

PR #8237, opened 2018-08-29
We should call the UnLoad methods when we detect an error during call…

normanmaurer:
…ing OnLoad

Motivation:

We should ensure we call *UnLoad when we detect an error during calling *OnLoad and previous *OnLoad calls were succesfull.

Modifications:

Correctly call *UnLoad when needed.

Result:

More correct code and no leaks when an error happens during loading the native lib.


normanmaurer:
This came up 
https://github.com/netty/netty/pull/8231#discussion_r213451757.

And was also done for netty-tcnative https://github.com/netty/netty-tcnative/pull/384

Issue #8217, opened 2018-08-23
Log4J2LoggerFactory depends on a log4j2 version which does not work on Java 6


jerrinot:
### Expected behavior
Netty logging works via log4j2 while on Java 6

### Actual behavior
Log4j2 2.4+ depends on Java 7. Netty depends on log4j2 methods which are present only in newer versions. When older log4j2 version is on a classpath then Netty fails to initialize itself due `AbstractMethodError` while logging. 

### Steps to reproduce
Have Netty 4.1.27 and log4j2 on a classpath and run `InternalLoggerFactory.getInstance("foo").debug("bar", new Object));`

### Minimal yet complete reproducer code (or URL to code)
```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <groupId>info.jerrinot</groupId>
    <artifactId>netty-logging-issue-reproducer</artifactId>
    <version>1.0-SNAPSHOT</version>

    <dependencies>
        <dependency>
            <groupId>io.netty</groupId>
            <artifactId>netty-common</artifactId>
            <version>4.1.27.Final</version>
        </dependency>

        <dependency>
            <groupId>org.apache.logging.log4j</groupId>
            <artifactId>log4j-core</artifactId>
            <version>2.3</version>
        </dependency>

        <dependency>
            <groupId>junit</groupId>
            <artifactId>junit</artifactId>
            <version>4.12</version>
        </dependency>
    </dependencies>
</project>
```

```java
package info.jerrinot.nettyloggingissue;

import io.netty.util.internal.logging.InternalLoggerFactory;
import org.junit.Test;

public class ReproducerTest {

    @Test
    public void testNettyLogging() {
        InternalLoggerFactory.getInstance("foo").debug("bar", new Object());
    }
}
```

### Netty version
4.1.27

### JVM version (e.g. `java -version`)
irrelevant
### OS version (e.g. `uname -a`)
irrelevant

jerrinot:
fwiw - we use OpenSSL in some integration tests. They have to run on Java 6. I end-up with implementing this nasty hack (test-code only):
```java
    private static void patchOpenSslLogging() {
[...]

PR #8239, opened 2018-08-29
Do not fail on runtime when an older version of Log4J2 is on the clas…

normanmaurer:
…spath.

Motivation:

At the moment we will just assume the correct version of log4j2 is used when we find it on the classpath. This may lead to an AbstractMethodError at runtime.

Modifications:

Let Log4J2Logger extend AbstractInternalLogger and check on class loading if we need to take special action to delegat to older versions of Log4J2.

Result:

Fixes https://github.com/netty/netty/issues/8217.


normanmaurer:
@ejona86 yeah that is probably a better idea.

normanmaurer:
Closing this in favour of https://github.com/netty/netty/pull/8240

PR #8240, opened 2018-08-30
Do not fail on runtime when an older version of Log4J2 is on the clas…

### LINKED ISSUE #8217: Log4J2LoggerFactory depends on a log4j2 version which does not work on Java 6

normanmaurer:
…spath.

Motivation:

At the moment we will just assume the correct version of log4j2 is used when we find it on the classpath. This may lead to an AbstractMethodError at runtime. We should not use log4j2 if the version is not correct.

Modifications:

Check on class loading if we can use Log4J2 or not.

Result:

Fixes #8217.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/logging/Log4J2Logger.java
+++ b/common/src/main/java/io/netty/util/internal/logging/Log4J2Logger.java
@@ -21,14 +21,42 @@ import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.spi.ExtendedLogger;
 import org.apache.logging.log4j.spi.ExtendedLoggerWrapper;
 
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+
 import static io.netty.util.internal.logging.AbstractInternalLogger.EXCEPTION_MESSAGE;
 
 class Log4J2Logger extends ExtendedLoggerWrapper implements InternalLogger {
 
     private static final long serialVersionUID = 5485418394879791397L;
+    private static final boolean VARARGS_ONLY;
+
+    static {
+        // Older Log4J2 versions have only log methods that takes the format + varargs. So we should not use
+        // Log4J2 if the version is too old.
+        // See https://github.com/netty/netty/issues/8217
+        VARARGS_ONLY = AccessController.doPrivileged(new PrivilegedAction<Boolean>() {
+            @Override
+            public Boolean run() {
+                try {
+                    Logger.class.getMethod("debug", String.class, Object.class);
+                    return false;
+                } catch (NoSuchMethodException ignore) {
+                    // Log4J2 version too old.
+                    return true;
+                } catch (SecurityException ignore) {
+                    // We could not detect the version so we will use Log4J2 if its on the classpath.
+                    return false;
+                }
+            }
+        });
+    }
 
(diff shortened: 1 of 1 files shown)


common/src/main/java/io/netty/util/internal/logging/Log4J2Logger.java
```diff
@@ -16,101 +16,193 @@
...
 
-    /** {@linkplain AbstractInternalLogger#EXCEPTION_MESSAGE} */
-    private static final String EXCEPTION_MESSAGE = "Unexpected exception:";
+    private final Logger logger;
 
     Log4J2Logger(Logger logger) {
-        super((ExtendedLogger) logger, logger.getName(), logger.getMessageFactory());
+        super(logger.getName());
+        if (VARARGS_ONLY) {
+            throw new UnsupportedOperationException("Log4J2 version missmatch");
```
carl-mastrangelo: mismatch*

normanmaurer:
@ejona86 just because it is cleaner to wrap and not extend and depend on the fact that their method signature match ours. Let me separate the change tho.

normanmaurer:
@ejona86 nevermind I was wrong... We need to extend ExtendedLoggerWrapper to have `%F` work correctly. Let me revert most of this pr.

normanmaurer:
@ejona86 adjusted...

normanmaurer:
Let me pull this in... we can do a followup if we feel this is needed.

Issue #7431, opened 2017-11-22
grpc mutual authentication issue


shank9918:
### Expected behavior
To enable ssl mutual authentication

### Actual behavior
Using PKCS11 as keystore provider and loading certs from HSM.
Client fails during mutual authentication with following error and stacktrace

[DEBUG] 2017-11-22 00:34:21.484 [grpc-default-worker-ELG-2-2] ReferenceCountedOpenSslClientContext - **request of key failed**
 javax.net.ssl.SSLException: java.lang.NullPointerException
        at io.netty.handler.ssl.OpenSslKeyMaterialManager.keyMaterial(OpenSslKeyMaterialManager.java:121) ~[netty-handler-4.1.16.Final.jar!/:4.1.16.Final]
        at io.netty.handler.ssl.ReferenceCountedOpenSslClientContext$OpenSslCertificateRequestedCallback.requested(ReferenceCountedOpenSslClientContext.java:249) [netty-handler-4.1.16.Final.jar!/:4.1.16.Final]
        at io.netty.internal.tcnative.SSL.readFromSSL(Native Method) ~[netty-tcnative-boringssl-static-2.0.6.Final.jar!/:2.0.6.Final]
        at io.netty.handler.ssl.ReferenceCountedOpenSslEngine.readPlaintextData(ReferenceCountedOpenSslEngine.java:482) [netty-handler-4.1.16.Final.jar!/:4.1.16.Final]

### Steps to reproduce
Works with JKS store provider but while using PKCS11 it fails.

### Minimal yet complete reproducer code (or URL to code)
Client code:
ManagedChannel channel = null;
try {

	channel = NettyChannelBuilder.forAddress(host, port)
			.sslContext(GrpcSslContexts.forClient()
					.keyManager(this.keyManagerFactory)
					.trustManager(this.trustManagerFactory)
					.build())
			.overrideAuthority(this.authority)
			.build();
} catch (SSLException e) {
	throw new IllegalStateException("gRPC client cannot be started", e);
}

Server Code:
Server server = NettyServerBuilder.forPort(Integer.valueOf(environment.getProperty("grpc.port")))
.sslContext(GrpcSslContexts.configure(SslContextBuilder.forServer(sslUtil.keyManagerFactory())
		.trustManager(sslUtil.trustManagerFactory()))
		.clientAuth(ClientAuth.REQUIRE)
		.build())
.addService(grpcServer)
.build();

### Netty version
<grpc.version>1.7.0</grpc.version>
[...]

PR #8253, opened 2018-09-03
PemPrivateKey.toPem(...) should throw IllegalArgumentException when P…

normanmaurer:
…rivateKey which does not support encoding is used.

Motivation:

At the moment when a PrivateKey is used that does not support encoding we throw a NPE when trying to convert the key. We should better throw an IllegalArgumentException with the details about what key we tried to encode.

Modifications:

- Check if PrivateKey.getEncoded() returns null and if so throw an IllegalArgumentException
- Add unit test.

Result:

Better handling of non-supported PrivateKey implementations.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/PemPrivateKey.java
+++ b/handler/src/main/java/io/netty/handler/ssl/PemPrivateKey.java
@@ -62,7 +62,7 @@ public final class PemPrivateKey extends AbstractReferenceCounted implements Pri
 
         byte[] bytes = key.getEncoded();
         if (bytes == null) {
-            throw new IllegalArgumentException(key + " does not support encoding");
+            throw new IllegalArgumentException(key.getClass().getName() + " does not support encoding");
         }
 
         ByteBuf encoded = Unpooled.wrappedBuffer(bytes);
--- a/handler/src/main/java/io/netty/handler/ssl/PemPrivateKey.java
+++ b/handler/src/main/java/io/netty/handler/ssl/PemPrivateKey.java
@@ -60,7 +60,12 @@ public final class PemPrivateKey extends AbstractReferenceCounted implements Pri
             return ((PemEncoded) key).retain();
         }
 
-        ByteBuf encoded = Unpooled.wrappedBuffer(key.getEncoded());
+        byte[] bytes = key.getEncoded();
+        if (bytes == null) {
+            throw new IllegalArgumentException(key + " does not support encoding");
+        }
+
+        ByteBuf encoded = Unpooled.wrappedBuffer(bytes);
         try {
             ByteBuf base64 = SslUtils.toBase64(allocator, encoded);
             try {
--- a/handler/src/test/java/io/netty/handler/ssl/PemEncodedTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/PemEncodedTest.java
@@ -24,7 +24,10 @@ import static org.junit.Assume.assumeTrue;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileInputStream;
+import java.security.PrivateKey;
 
+import io.netty.buffer.Unpooled;
+import io.netty.buffer.UnpooledByteBufAllocator;
 import org.junit.Test;
 
 import io.netty.handler.ssl.util.SelfSignedCertificate;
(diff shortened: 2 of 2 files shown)


normanmaurer:
This is related to https://github.com/netty/netty/issues/7431... This will at least give a better error to the user.

handler/src/main/java/io/netty/handler/ssl/PemPrivateKey.java
```diff
@@ -60,7 +60,12 @@ static PemEncoded toPEM(ByteBufAllocator allocator, boolean useDirect, PrivateKe
             return ((PemEncoded) key).retain();
         }
 
-        ByteBuf encoded = Unpooled.wrappedBuffer(key.getEncoded());
+        byte[] bytes = key.getEncoded();
+        if (bytes == null) {
+            throw new IllegalArgumentException(key + " does not support encoding");
```
ejona86: `s/key/key.getClass().getName()/`?

ejona86: Right now it looks like it would give the inherited Object.toString() here for the most part. And we would hope it doesn't print out the private key. It's really the class name that is most important, so the swap cleans up the message.

normanmaurer: Good point... let me change it

normanmaurer: @ejona86 done

handler/src/test/java/io/netty/handler/ssl/PemEncodedTest.java
```diff
@@ -69,6 +72,26 @@ private static void testPemEncoded(SslProvider provider) throws Exception {
         }
     }
 
+    @Test(expected = IllegalArgumentException.class)
```
carl-mastrangelo: fwiw, I would avoid using `expected`, because it isn't clear what the error is.  Below, getAlgorithm returns null, which might also trigger IllegalArgumentException.  I can't really tell without digging into the source.  

Instead, using `ExpectedException` JUnit rule let's you specify both the exception type, and match a partial string of the error message.  It's more resilient to catching the wrong exception, and it's easier to read.

trustin: Time to migrate to a better option like AssertJ or Spock?

trustin:
```java
assertThatThrownBy(() -> {
    ...
}).isInstanceOf(IllegalArgumentException.class);
```

handler/src/main/java/io/netty/handler/ssl/PemPrivateKey.java
```diff
@@ -60,7 +60,12 @@ static PemEncoded toPEM(ByteBufAllocator allocator, boolean useDirect, PrivateKe
             return ((PemEncoded) key).retain();
         }
 
-        ByteBuf encoded = Unpooled.wrappedBuffer(key.getEncoded());
+        byte[] bytes = key.getEncoded();
+        if (bytes == null) {
+            throw new IllegalArgumentException(key.getClass().getName() + " does not support encoding");
```
trustin: nit: If it's a sentence, it should end with a period.

handler/src/test/java/io/netty/handler/ssl/PemEncodedTest.java
carl-mastrangelo: I haven't used either, though my preferred option `Truth` says it's similar to AssertJ.  JUnit5 has `assertThrows` which could also be used.

trustin: I'm a big fan of AssertJ due to its method chaining design which reduces the amount of imports. Alternatively, Spock may be a good choice although it's Groovy.

carl-mastrangelo: Yeah, it looks like even the [Truth](https://google.github.io/truth/comparison) page likes AssertJ better.

normanmaurer: Honestly I dont care too much... its "just" a testing library. So whatever fits us is ok by me. I think the important is to be just consistent and us either A or B and not end up using everything.

normanmaurer: @carl-mastrangelo please feel free to submit a PR using `@expected` here.

PR #7519, opened 2017-12-18
Outputs the line number from where the logging request was issued

chaosky:
Motivation:

Outputs the line number from where the logging request was issued when debugging netty

Modification:

Add LocationAwareSlf4JLogger can output the correct line number

Result:

Users of this library can correctly output the line number of the log when using slf4j

PR #8258, opened 2018-09-04
Log the correct line-number when using SLF4j with netty if possible.

normanmaurer:
Motivation:

At the moment we do not log the correct line number in many cases as it will log the line number of the logger wrapper itself. Slf4j does have an extra interface that can be used to filter out these nad make it more usable with logging wrappers.

Modifications:

Detect if the returned logger implements LocationAwareLogger and if so make use of its extra methods to be able to log the correct origin of the log request.

Result:

Better logging when using slf4j.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/logging/LocationAwareSlf4JLogger.java
+++ b/common/src/main/java/io/netty/util/internal/logging/LocationAwareSlf4JLogger.java
@@ -1,0 +1,248 @@
+/*
+ * Copyright 2017 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.util.internal.logging;
+
+import org.slf4j.spi.LocationAwareLogger;
+
+import static org.slf4j.spi.LocationAwareLogger.*;
+
+/**
+ * <a href="http://www.slf4j.org/">SLF4J</a> logger which is location aware and so will log the correct origin of the
+ * logging event by filter out the wrapper itself.
+ */
+final class LocationAwareSlf4JLogger extends AbstractInternalLogger {
+
+    // IMPORTANT: All our log methods first check if the log level is enabled before call the wrapped
+    // LocationAwareLogger.log(...) method. This is done to reduce GC creation that is caused by varargs.
+
+    private static final String FQCN = LocationAwareSlf4JLogger.class.getName();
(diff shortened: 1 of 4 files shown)


normanmaurer:
This is a reworked version of https://github.com/netty/netty/pull/7519.

common/src/main/java/io/netty/util/internal/logging/LocationAwareSlf4JLogger.java
```diff
@@ -0,0 +1,248 @@
...
+ * <a href="http://www.slf4j.org/">SLF4J</a> logger which is location aware and so will log the correct origin of the
+ * logging event by filter out the wrapper itself.
+ */
+final class LocationAwareSlf4JLogger extends AbstractInternalLogger {
+
+    // IMPORTANT: All our log methods first check if the log level is enabled before call the wrapped
+    // LocationAwareLogger.log(...) method. This is done to reduce GC creation that is caused by varargs.
+
+    private static final String FQCN = LocationAwareSlf4JLogger.class.getName();
+    private static final long serialVersionUID = -8292030083201538180L;
```
ejona86: nit: could use 0 or 1 here (or your favorite number). Using the ugly id is only needed for compatibility with previous versions of the class before you added the `serialVersionUID`.

normanmaurer: I just pressed "generate" :D

common/src/main/java/io/netty/util/internal/logging/LocationAwareSlf4JLogger.java
```diff
@@ -0,0 +1,248 @@
...
+
+    @Override
+    public void info(String format, Object... argArray) {
+        if (isInfoEnabled()) {
+            log(INFO_INT, format, argArray);
+        }
+    }
+
+    @Override
+    public void info(String msg, Throwable t) {
```
carl-mastrangelo: t is unused?

common/src/main/java/io/netty/util/internal/logging/LocationAwareSlf4JLogger.java
```diff
@@ -0,0 +1,248 @@
...
+    public void warn(String format, Object argA, Object argB) {
+        if (isWarnEnabled()) {
+            log(WARN_INT, format, argA, argB);
+        }
+    }
+
+    @Override
+    public void warn(String msg, Throwable t) {
+        if (isWarnEnabled()) {
+            log(WARN_INT, msg);
```
carl-mastrangelo: same here

common/src/main/java/io/netty/util/internal/logging/Slf4JLoggerFactory.java
```diff
@@ -44,6 +46,12 @@ public Slf4JLoggerFactory() {
 
     @Override
     public InternalLogger newInstance(String name) {
-        return new Slf4JLogger(LoggerFactory.getLogger(name));
+        return wrapLogger(LoggerFactory.getLogger(name));
+    }
+
+    // package-private for testing.
+    static InternalLogger wrapLogger(Logger logger) {
+        return  logger instanceof LocationAwareLogger ?
```
carl-mastrangelo: `s/  / /`

normanmaurer:
@carl-mastrangelo good catch! Addressed all your comments. PTAL again

PR #8260, opened 2018-09-04
NioEventLoop should also use our special SelectionKeySet on Java9 and…

normanmaurer:
… later.

Motivation:

In Java8 and earlier we used reflection to replace the used key set if not otherwise told. This does not work on Java9 and later without special flags as its not possible to call setAccessible(true) on the Field anymore.

Modifications:

- Use Unsafe to instrument the Selector with out special set when sun.misc.Unsafe is present and we are using Java9+.

Result:

NIO transport produce less GC on Java9 and later as well.


normanmaurer:
@ejona86 lol yah... We already have a job on the CI that runs with `-Dio.netty.noKeySetOptimization=true` so we should be covered here.

AlanBateman:
Hacking the private fields in the Selector implementation is really horrible and can break at any time. If there are really issues with object allocation adding to the selected key set then use the select methods that invoke an action instead.

normanmaurer:
We will do this once we move to java8 only... that’s also why we have a fallback

PR #8266, opened 2018-09-05
Use a non-volatile read for ensureAccessible() whenever possible to r…

normanmaurer:
…educe overhead and allow better inlining.

Motiviation:

At the moment whenever ensureAccessible() is called in our ByteBuf implementations (which is basically on each operation) we will do a volatile read. That per-se is not such a bad thing but the problem here is that it will also reduce the the optimizations that the compiler / jit can do. For example as these are volatile it can not eliminate multiple loads of it when inline the methods of ByteBuf which happens quite frequently because most of them a quite small and very hot. That is especially true for all the methods that act on primitives.

It gets even worse as people often call a lot of these after each other in the same method or even use method chaining here.

The idea of the change is basically just ue a non-volatile read for the ensureAccessible() check as its a best-effort implementation to detect acting on already released buffers anyway as even with a volatile read it could happen that the user will release it in another thread before we actual access the buffer after the reference check.

Modifications:

- Try to do a non-volatile read using sun.misc.Unsafe if we can use it.
- Add a benchmark

Result:

Big performance win when multiple ByteBuf methods are called from a method.

With the change:
```
UnsafeByteBufBenchmark.setGetLongUnsafeByteBuf  thrpt   20  281395842,128 ± 5050792,296  ops/s
```

Before the change:
```
UnsafeByteBufBenchmark.setGetLongUnsafeByteBuf  thrpt   20  217419832,801 ± 5080579,030  ops/s
```

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
@@ -1442,9 +1442,17 @@ public abstract class AbstractByteBuf extends ByteBuf {
      * if the buffer was released before.
      */
     protected final void ensureAccessible() {
-        if (checkAccessible && refCnt() == 0) {
+        if (checkAccessible && internalRefCnt() == 0) {
             throw new IllegalReferenceCountException(0);
         }
+    }
+
+    /**
+     * Returns the reference count that is used internally by {@link #ensureAccessible()} to try to guard
+     * against using the buffer after it was released (best-effort).
+     */
+    int internalRefCnt() {
+        return refCnt();
     }
 
     final void setIndex0(int readerIndex, int writerIndex) {
--- a/buffer/src/main/java/io/netty/buffer/AbstractReferenceCountedByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/AbstractReferenceCountedByteBuf.java
@@ -17,6 +17,7 @@
 package io.netty.buffer;
 
 import io.netty.util.IllegalReferenceCountException;
+import io.netty.util.internal.PlatformDependent;
 
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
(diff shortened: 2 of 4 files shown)


normanmaurer:
I shamefully admit I kind of reverted the same code at some point (not exactly the same but kind of):

https://github.com/netty/netty/commit/f23d68b42fd71674fee6623f6779d70b6849da21

normanmaurer:
Also please note that the benchmarks reproduce the numbers all the time. Sometimes the win is even bigger.

Also big thanks to @nitsanw who did look into the perfasm output with me and trying to make sense out of my benchmark numbers before the change 👍

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -1442,11 +1442,19 @@ private void checkReadableBytes0(int minimumReadableBytes) {
      * if the buffer was released before.
      */
     protected final void ensureAccessible() {
-        if (checkAccessible && refCnt() == 0) {
+        if (checkAccessible && internalRefCnt() == 0) {
```
carl-mastrangelo: If this is best effort, why not just sample?  Maybe 1/16 times it checks.

normanmaurer: I think what we do now is better then sampling imho. so I would prefer to keep it as it is in the PR.

normanmaurer:
@ejona86 I think in the real world at the end it will be "propagated at some point" and so will be detected even on multi-threaded usage. Also as you said I think most of the times it will be no a big deal. If we are really super conservative about it we could also have different levels of "access checking":
 - volatile
 - non-volatile
 - none

This actually would kind of make sense as we already allow to disable checking in general. Let me push another commit here to show how this would look like and then we can discuss if we want to do it or if this would be overkill.

normanmaurer:
@ejona86 actually nevermind... I think this is kind of overkill and we can do this in a followup if we really feel the need later on.

Issue #327, opened 2012-05-16
EPollArrayWrapper.epollWait 100% CPU Usage


blucas:
Hi,

I believe I have an issue similar to #302 but on Linux (Ubuntu 10.04) with JDK (1.6.0u30) and JDK(1.7.0u4) using Netty-4.0.0 (Revision: 52a7d28cb59e3806fda322aecf7a85a6adaeb305)

The app is proxying connections to backend systems.  The proxy has a pool of channels that it can use to send requests to the backend systems.  If the pool is low on channels, new channels are spawned and put into the pool so that requests sent to the proxy can be serviced.  The pools get populated on app startup, so that is why it doesn't take long at all for the CPU to spike through the roof (22 seconds into the app lifecycle).

The test box has two CPUs, the output from 'top' is below:

```
PID  USER   PR  NI  VIRT  RES  SHR S %CPU %MEM    TIME+  COMMAND
8220 root   20   0 2281m 741m  10m R 50.2 18.7  0:22.57 java                                                                             
8218 root   20   0 2281m 741m  10m R 49.9 18.7  0:22.65 java                                                                             
8219 root   20   0 2281m 741m  10m R 49.2 18.7  0:22.86 java                                                                             
8221 root   20   0 2281m 741m  10m R 49.2 18.7  0:22.20 java 
```

Thread Dump for the four NioClient based Worker Threads that are chewing up all the CPU.

```
"backend-worker-pool-7-thread-1" prio=10 tid=0x00007f5918015800 nid=0x201a runnable [0x00007f5924ba3000]
   java.lang.Thread.State: RUNNABLE
    at sun.nio.ch.EPollArrayWrapper.epollWait(Native Method)
    at sun.nio.ch.EPollArrayWrapper.poll(EPollArrayWrapper.java:210)
    at sun.nio.ch.EPollSelectorImpl.doSelect(EPollSelectorImpl.java:65)
    at sun.nio.ch.SelectorImpl.lockAndDoSelect(SelectorImpl.java:69)
    - locked <0x000000008be93580> (a sun.nio.ch.Util$2)
    - locked <0x000000008be93570> (a java.util.Collections$UnmodifiableSet)
    - locked <0x000000008be92548> (a sun.nio.ch.EPollSelectorImpl)
    at sun.nio.ch.SelectorImpl.select(SelectorImpl.java:80)
    at io.netty.channel.socket.nio.SelectorUtil.select(SelectorUtil.java:55)
[...]

PR #8267, opened 2018-09-06
Added option to do busy-wait on epoll

merlimat:
Motivation:

Add an option (through a `SelectStrategy` return code) to have the Netty event loop thread to do busy-wait on the epoll.

The reason for this change is to avoid the context switch cost that comes when the event loop thread is blocked on the `epoll_wait()` call. 

On average, the context switch has a penalty of ~13usec. 

This benefits both: 
 * The latency when reading from a socket
 * Scheduling tasks to be executed on the event loop thread.

The tradeoff, when enabling this feature, is that the event loop thread will be using 100% cpu, even when inactive.

Modification:

 * Added `SelectStrategy` option to return `BUSY_WAIT` 
 * Epoll loop will do a `epoll_wait()` with no timeout
 * Use `pause` instruction to hint to processor that we're in a busy loop

Result:

 * When enabled, minimizes impact of context switch in the critical path

### DIFF
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/Native.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/Native.java
@@ -132,6 +132,7 @@ public final class Native {
         }
         return ready;
     }
+
     private static native int epollBusyWait0(int efd, long address, int len);
 
     public static void epollCtlAdd(int efd, final int fd, final int flags) throws IOException {
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/Native.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/Native.java
@@ -120,6 +120,20 @@ public final class Native {
     }
     private static native int epollWait0(int efd, long address, int len, int timerFd, int timeoutSec, int timeoutNs);
 
+    /**
+     * Non-blocking variant of
+     * {@link #epollWait(FileDescriptor, EpollEventArray, FileDescriptor, int, int)}
+     * that will also hint to processor we are in a busy-wait loop.
+     */
+    public static int epollBusyWait(FileDescriptor epollFd, EpollEventArray events) throws IOException {
+        int ready = epollBusyWait0(epollFd.intValue(), events.memoryAddress(), events.length());
+        if (ready < 0) {
+            throw newIOException("epoll_wait", ready);
+        }
+        return ready;
+    }
+    private static native int epollBusyWait0(int efd, long address, int len);
+
     public static void epollCtlAdd(int efd, final int fd, final int flags) throws IOException {
         int res = epollCtlAdd0(efd, fd, flags);
         if (res < 0) {
--- a/transport-native-epoll/src/test/java/io/netty/channel/epoll/EpollSocketStringEchoBusyWaitTest.java
+++ b/transport-native-epoll/src/test/java/io/netty/channel/epoll/EpollSocketStringEchoBusyWaitTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 The Netty Project
+ * Copyright 2018 The Netty Project
  *
  * The Netty Project licenses this file to you under the Apache License,
  * version 2.0 (the "License"); you may not use this file except in compliance
--- a/transport-native-epoll/src/test/java/io/netty/channel/epoll/EpollSocketStringEchoBusyWaitTest.java
+++ b/transport-native-epoll/src/test/java/io/netty/channel/epoll/EpollSocketStringEchoBusyWaitTest.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2014 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.channel.epoll;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+
+import io.netty.bootstrap.Bootstrap;
+import io.netty.bootstrap.ServerBootstrap;
+import io.netty.channel.EventLoopGroup;
+import io.netty.channel.SelectStrategy;
+import io.netty.channel.SelectStrategyFactory;
+import io.netty.testsuite.transport.TestsuitePermutation;
+import io.netty.testsuite.transport.TestsuitePermutation.BootstrapComboFactory;
+import io.netty.testsuite.transport.TestsuitePermutation.BootstrapFactory;
+import io.netty.testsuite.transport.socket.SocketStringEchoTest;
+import io.netty.util.IntSupplier;
+import io.netty.util.concurrent.DefaultThreadFactory;
+
+public class EpollSocketStringEchoBusyWaitTest extends SocketStringEchoTest {
+
+    private static EventLoopGroup EPOLL_LOOP;
+
+    @BeforeClass
+    public static void setup() throws Exception {
+        EPOLL_LOOP = new EpollEventLoopGroup(2, new DefaultThreadFactory("testsuite-epoll-busy-wait", true),
+                new SelectStrategyFactory() {
+                    @Override
+                    public SelectStrategy newSelectStrategy() {
+                        return new SelectStrategy() {
+                            @Override
+                            public int calculateStrategy(IntSupplier selectSupplier, boolean hasTasks)
+                                    throws Exception {
+                                return SelectStrategy.BUSY_WAIT;
+                            }
+                        };
+                    }
+                });
+    }
+
+    @AfterClass
+    public static void teardown() throws Exception {
+        if (EPOLL_LOOP != null) {
+            EPOLL_LOOP.shutdownGracefully();
+        }
+    }
+
+    @Override
+    protected List<TestsuitePermutation.BootstrapComboFactory<ServerBootstrap, Bootstrap>> newFactories() {
+        List<BootstrapComboFactory<ServerBootstrap, Bootstrap>> list =
+                new ArrayList<BootstrapComboFactory<ServerBootstrap, Bootstrap>>();
+        final BootstrapFactory<ServerBootstrap> sbf = serverSocket();
+        final BootstrapFactory<Bootstrap> cbf = clientSocket();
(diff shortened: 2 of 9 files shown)


transport-native-epoll/src/test/java/io/netty/channel/epoll/EpollSocketStringEchoBusyWaitTest.java
```diff
@@ -0,0 +1,102 @@
+/*
+ * Copyright 2014 The Netty Project
```
normanmaurer: 2018

normanmaurer:
@merlimat just curious what you use this for... Do you try to minimise the latency of some app that is really network heavy ?

merlimat:
@normanmaurer main goal is to reduce context switches in the critical path. Currently there can be multiple tasks executed in the Netty event-loop. For example, when writing to a channel from a different thread we have to "jump" to the event loop (to retain ordering and avoid mutex contention). 

By having the eventLoop thread spinning, when you submit the task it will be picked up immediately (generally in ~100ns) compared to waking up the thread from `epoll_wait` through the eventfd which will take ~10usec. The same latency reduction applies when reading something from a channel, if the thread is spinning it will "see" the message earlier.

While a single context switch is not the end of the world, in a complex application the number of these "thread jumps" and socket reads chains can be significant and become a dominating factor in the overall request/response latency.

Of course this is only really useful when trying to squeeze latency at very minimum, at the expense of CPU usage (and max throughput, since it would be keeping the cpu busy for event loop while it could be used for other threads).

Also, this would be typically useful paired with CPU affinity as well. The effects of the 2 are: 
  * Busy-wait brings down avg/med latency offset 
  * CPU affinity will lower 99pct (and above) long tails by removing intermittent jitter

normanmaurer:
@merlimat makes sense... do you have any benchmarks to share (Just curious) ?

transport-native-epoll/src/main/java/io/netty/channel/epoll/Native.java
```diff
@@ -120,6 +120,20 @@ public static int epollWait(FileDescriptor epollFd, EpollEventArray events, File
...
+     * that will also hint to processor we are in a busy-wait loop.
+     */
+    public static int epollBusyWait(FileDescriptor epollFd, EpollEventArray events) throws IOException {
+        int ready = epollBusyWait0(epollFd.intValue(), events.memoryAddress(), events.length());
+        if (ready < 0) {
+            throw newIOException("epoll_wait", ready);
+        }
+        return ready;
+    }
+    private static native int epollBusyWait0(int efd, long address, int len);
```
normanmaurer: nit: add empty line above.

transport-native-epoll/src/test/java/io/netty/channel/epoll/EpollSocketStringEchoBusyWaitTest.java
```diff
@@ -0,0 +1,102 @@
...
+    private BootstrapFactory<ServerBootstrap> serverSocket() {
+        return new BootstrapFactory<ServerBootstrap>() {
+            @Override
+            public ServerBootstrap newInstance() {
+                return new ServerBootstrap().group(EPOLL_LOOP, EPOLL_LOOP).channel(EpollServerSocketChannel.class);
+            }
+        };
+    }
+
+    private BootstrapFactory<Bootstrap> clientSocket() {
```
normanmaurer: nit: static ?

transport-native-epoll/src/test/java/io/netty/channel/epoll/EpollSocketStringEchoBusyWaitTest.java
```diff
@@ -0,0 +1,102 @@
...
+            @Override
+            public Bootstrap newClientInstance() {
+                return cbf.newInstance();
+            }
+        });
+
+        return list;
+    }
+
+    private BootstrapFactory<ServerBootstrap> serverSocket() {
```
normanmaurer: nit: static ?

transport-native-epoll/src/test/java/io/netty/channel/epoll/EpollSocketStringEchoBusyWaitTest.java
```diff
@@ -0,0 +1,102 @@
...
+    @BeforeClass
+    public static void setup() throws Exception {
+        EPOLL_LOOP = new EpollEventLoopGroup(2, new DefaultThreadFactory("testsuite-epoll-busy-wait", true),
+                new SelectStrategyFactory() {
+                    @Override
+                    public SelectStrategy newSelectStrategy() {
+                        return new SelectStrategy() {
+                            @Override
+                            public int calculateStrategy(IntSupplier selectSupplier, boolean hasTasks)
+                                    throws Exception {
```
normanmaurer: nit: remove throws ....

transport-native-epoll/src/test/java/io/netty/channel/epoll/EpollSocketStringEchoBusyWaitTest.java
```diff
@@ -0,0 +1,102 @@
...
+import io.netty.channel.SelectStrategy;
+import io.netty.channel.SelectStrategyFactory;
+import io.netty.testsuite.transport.TestsuitePermutation;
+import io.netty.testsuite.transport.TestsuitePermutation.BootstrapComboFactory;
+import io.netty.testsuite.transport.TestsuitePermutation.BootstrapFactory;
+import io.netty.testsuite.transport.socket.SocketStringEchoTest;
+import io.netty.util.IntSupplier;
+import io.netty.util.concurrent.DefaultThreadFactory;
+
+public class EpollSocketStringEchoBusyWaitTest extends SocketStringEchoTest {
```
normanmaurer: can you add one for ET and one for LT mode ? (just as we do for other Epoll tests.

merlimat:
@normanmaurer  Yes, I did some brief measurements.  This is the latency for an rpc request/response application. There are additional thread switches (though they are already using a local queue and busy-waiting on that as well). 

Conditions:
 * Client/server on same machine with TCP connection
 * Tested on a bare-metal node
 * Client sends 1K rps (size 1KB) and waits for response
 * Multiple request outstanding, responses in order
 * Latency measured in millis

|                      | 50pct | 95pct | 99pct | 99.9pct | 99.99pct | 99.999pct | max   |
|----------|-----|-----|-----|-----|-----|-----|-----|
| Baseline             | 0.049 | 0.059 | 0.066 | 0.095   | 0.449    | 0.465     | 0.466 |
| **Busy-Wait IO Threads** | 0.036 | 0.047 | 0.052 | 0.075   | 0.436    | 0.445     | 0.452 |
| Disable HyperThread  | 0.029 | 0.038 | 0.041 | 0.054   | 0.204    | 0.388     | 0.396 |
| CPU-Isolation	       | 0.029 | 0.037 | 0.040 | 0.050   | 0.066    | 0.090     | 0.104 |

I have included the disable hyper-threading and cpu-isolation as they are the next steps (and easily configured in linux settings). 
The remaining bulk of the median latency I think is related to SO_BUSY_POLL (#8268) being ineffective on loopback interface.

This is a very simple example, but in a more complex app, there would be more thread jumps and possibly more RPCs involved.

transport-native-epoll/src/test/java/io/netty/channel/epoll/EpollSocketStringEchoBusyWaitTest.java
merlimat: Sure, will address the comments shortly.

carl-mastrangelo:
@merlimat Have you considered busy polling on reads or writes, rather than on polls?   Even if poll does return something, you need to follow it up with a subsequent read, which is going to add a syscall of additional latency before you get your data.  How many active sockets do you have, and are you trying to busy poll on writes or reads more?

merlimat:
@carl-mastrangelo It's not just the read or write. The Netty event loop is also used to execute task in the same connection thread. 

For example, to ensure ordering of operations and to avoid mutex contention, we typically do : 

```java
ctx.channel().eventLoop().execute(() -> {
    ctx.writeAndFlush(buffer); 
});
```

So the event loop blocks on the `epoll_wait` but also keeps an eye on the tasks that are posted, using the eventfd mechanism to wake up `epoll_wait` immediately when a task is posted.

>  Even if poll does return something, you need to follow it up with a subsequent read, which is going to add a syscall of additional latency before you get your data. 

Sure, the read will involve a syscall, but the data will be available and it won't be blocking. The main goal here is to make sure the IO thread is never (or to the least possible extent) descheduled. The cost in terms of latency of the context switch (~10 usec) is much higher than the cost of the syscall.

Also, we always have multiple connections to manage, with the direct `read()` approach (which I don't think is currently possible to do through Netty), we'd need to spin on 1 thread/connection, while with epoll we can just spin on 1 (or few) IO thread and leave the other CPUs for the rest of application.

merlimat:
@normanmaurer addressed comments

carl-mastrangelo:
@merlimat  two things:  

1.  I asked how many sockets you have because if you do have multiple, it might make more sense to dedicate a single thread to each one, one per core.  Then you could do reads directly.
2.  On a recent PR I made (#7816) it was measured that the changing the timerfd took about 4us, so I'll pick that as a baseline for syscall time.   Why do you not care about the followup read, if you do care about the context switch?  They are in the same order of magnitude.

merlimat:
> 1. I asked how many sockets you have because if you do have multiple,  it might make more sense to dedicate a single thread to each one, one  per core.  Then you could do reads directly. 

There is always more than 1 connection, from 10s to 100s or 1000s, so dedicating one core per connection is not really feasible. 

> 2. On a recent PR I made (#7816)  it was measured that the changing the timerfd took about 4us, so I'll  pick that as a baseline for syscall time.   Why do you not care about  the followup read, if you do care about the context switch?  They are in  the same order of magnitude.

How did you measure that time? 4 usec for a syscall seems very high to me. 

Just did some experiments and was seeing timings of < 0.8usec for epoll syscalls with no timeout (including JNI overhead).

carl-mastrangelo:
> How did you measure that time? 4 usec for a syscall seems very high to me.

The latency win on the benchmark.  Also, I frequently run ptrace with timing output, so I think 4 us is not uncommon for syscalls.   

I guess I don't have anything really against this, I just don't believe it will the solution you actually need.

normanmaurer:
@carl-mastrangelo @merlimat so we want to pull this in or not ? I am happy either way :)

carl-mastrangelo:
@normanmaurer I'll defer to you, but I can confidently say Google cannot use this (busy looping is extremely expensive for power and opportunity cost to run something else).  

But, @merlimat may be willing to make the trade off.

merlimat:
Sure, this setting is definitely not for general use, or meant to be a default.

We plan to give the option to enable busy wait in BookKeeper and Pulsar, for cases where one wants to maximize latency at expense of throughput and CPU consumption.

normanmaurer:
@merlimat thanks a lot!

merlimat:
Thanks for getting this in!

johnou:
@merlimat you mean minimise latency? ;)

merlimat:
Ouch.  Or maybe “maximize the responsiveness” :)

VaughnVernon:
Hi All, We are looking for some help on applying the resolution you reached here to another project. Please see:

https://github.com/netty/netty/issues/327#issuecomment-580362336

PR #8268, opened 2018-09-06
Allow to configure socket option SO_BUSY_POLL

merlimat:
Motivation:

When using Epoll based transport, allow applications to configure `SO_BUSY_POLL` socket option:

```

       SO_BUSY_POLL (since Linux 3.11)
              Sets the approximate time in microseconds to busy poll on a
              blocking receive when there is no data.  Increasing this value
              requires CAP_NET_ADMIN.  The default for this option is con‐
              trolled by the /proc/sys/net/core/busy_read file.

              The value in the /proc/sys/net/core/busy_poll file determines
              how long select(2) and poll(2) will busy poll when they oper‐
              ate on sockets with SO_BUSY_POLL set and no events to report
              are found.

              In both cases, busy polling will only be done when the socket
              last received data from a network device that supports this
              option.

              While busy polling may improve latency of some applications,
              care must be taken when using it since this will increase both
              CPU utilization and power usage.
```

Modification:

 * Added `SO_BUSY_POLL` socket option

Result:

 * Able to configure SO_BUSY_POLL from Netty

### DIFF
--- a/transport-native-epoll/src/main/c/netty_epoll_linuxsocket.c
+++ b/transport-native-epoll/src/main/c/netty_epoll_linuxsocket.c
@@ -50,6 +50,11 @@
 #define TCP_NOTSENT_LOWAT 25
 #endif
 
+// SO_BUSY_POLL is defined in linux 3.11. We define this here so older kernels can compile.
+#ifndef SO_BUSY_POLL
+#define SO_BUSY_POLL 46
+#endif
+
 static jclass peerCredentialsClass = NULL;
 static jmethodID peerCredentialsMethodId = NULL;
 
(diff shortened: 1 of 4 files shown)


normanmaurer:
@merlimat this looks good... After reading https://blog.cloudflare.com/how-to-achieve-low-latency/ I wonder why SO_BUSY_POLL would be preferred from just busy poll via user-space. Any pointers ?

merlimat:
I think the difference is that `SO_BUSY_POLL` uses support from the NIC itself. It should be supported in most interfaces For example: 

```
ethtool -k eth0 | grep poll
busy-poll: on [fixed]
```

while on loopback, for example it's not supported:
```
ethtool -k lo | grep poll
busy-poll: off [fixed]
```

My understanding is that it should move the packets from NIC to TCP stack with less delay. 

According to https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/performance_tuning_guide/sect-red_hat_enterprise_linux-performance_tuning_guide-networking-configuration_tools  : 

> Busy polling helps reduce latency in the network receive path by allowing socket layer code to poll the receive queue of a network device, and disabling network interrupts. This removes delays caused by the interrupt and the resultant context switch. However, it also increases CPU utilization. Busy polling also prevents the CPU from sleeping, which can incur additional power consumption.

normanmaurer:
@merlimat thanks... merged it.

PR #8272, opened 2018-09-07
Pack reader and writerIndex markers into one int

normanmaurer:
Motivation:

We can save some memory (4 bytes) per AbstractByteBuf by pack reader and writerIndex markers into one int (via bitshifting).

Modifications:

Use bitshifting to store markers into one int.

Result:

Less memory overhead per ByteBuf.


normanmaurer:
This may not sound like a big win but I think its worthwhile as often ByteBuf are used for small amount of memory so we should try to minimise the overhead.

ejona86:
Why not use two `short`s?

I'm a bit surprised that the indexes can't be more than 64K. Is that really true? I don't see any verification to detect overflow.

normanmaurer:
@ejona86 nevermind... I am an idiot... closing this as it not works 👎

PR #8280, opened 2018-09-11
Don't cause ClassCastException if registration fails during constructing…

normanmaurer:
… DnsNameResolver.

Motivation:

We should not try to cast the Channel to a DatagramChannel as this will cause a ClassCastException.

Modifications:

- Do not cast
- rethrow from constructor if we detect the registration failed.
- Add unit test.

Result:

Propagate correct exception.

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
@@ -152,7 +152,7 @@ public class DnsNameResolver extends InetNameResolver {
     private static final DatagramDnsQueryEncoder ENCODER = new DatagramDnsQueryEncoder();
 
     final Future<Channel> channelFuture;
-    final DatagramChannel ch;
+    final Channel ch;
 
     // Comparator that ensures we will try first to use the nameservers that use our preferred address type.
     private final Comparator<InetSocketAddress> nameServerComparator;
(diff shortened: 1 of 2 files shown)


normanmaurer:
Found this one while trying to reproduce another problem

normanmaurer:
@bryce-anderson nope... I just did something stupid and noticed I did not receive the correct error but a CCE

bryce-anderson:
@normanmaurer, cool. I appreciate the attention to detail even if it's not actively causing an issue. 👍

Btw, I think the title of this PR is off, should it be s/Do/Don't ?

normanmaurer:
@bryce-anderson fixed title... thanks

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -356,7 +356,18 @@ protected void initChannel(DatagramChannel ch) throws Exception {
...
+        ChannelFuture future = b.register();
+        Throwable cause = future.cause();
+        if (cause != null) {
+            if (cause instanceof RuntimeException) {
+                throw (RuntimeException) cause;
+            }
+            if (cause instanceof Error) {
+                throw (Error) cause;
+            }
+            throw new IllegalStateException("Unable to create / register Channel", cause);
```
carl-mastrangelo: ISE is for calling methods out of order.  I believe this should be a plain RuntimeException.

normanmaurer: @carl-mastrangelo hmm not sure about it... But not strong either. If you think this would be more correct please provide a pr :)

PR #8283, opened 2018-09-12
    Ensure X509KeyManager methods are called on the correct time when using server-side and support more methods of ExtendedSSLSession.

normanmaurer:
Motivation:

    Before when on server-side we just called the X509KeyManager methods when handshake() was called the first time which is not quite correct as we may not have received the full SSL hello / handshake and so could not extra for example the SNI hostname that was requested.
    OpenSSL exposes the SSL_CTX_set_cert_cb function which allows to set a callback which is executed at the correct moment, so we should use it. This also allows us to support more methods of ExtendedSSLSession easily.

    Modifications:

    - Make use of new methods exposed by netty-tcnative since https://github.com/netty/netty-tcnative/pull/388 to ensure we select the key material at the correct time.
    - Implement more methods of ExtendedOpenSslSession
    - Add unit tests to ensure we are able to retrieve various things on server-side in the X509KeyManager and so verify it is called at the correct time.
    - Simplify code by using new netty-tcnative methods.

    Result:

    More correct implementation for server-side usage and more complete implemented of ExtendedSSLSession.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/OpenSslKeyMaterialManager.java
+++ b/handler/src/main/java/io/netty/handler/ssl/OpenSslKeyMaterialManager.java
@@ -74,36 +74,25 @@ final class OpenSslKeyMaterialManager {
             if (type != null) {
                 String alias = chooseServerAlias(engine, type);
                 if (alias != null && aliases.add(alias)) {
-                    OpenSslKeyMaterial keyMaterial = null;
-                    try {
-                        keyMaterial = provider.chooseKeyMaterial(engine.alloc, alias);
-                        if (keyMaterial != null) {
-                            SSL.setKeyMaterialServerSide(
-                                    ssl, keyMaterial.certificateChainAddress(), keyMaterial.privateKeyAddress());
-                        }
-                    } catch (SSLException e) {
-                        throw e;
-                    } catch (Exception e) {
-                        throw new SSLException(e);
-                    } finally {
-                        if (keyMaterial != null) {
-                            keyMaterial.release();
-                        }
-                    }
+                    setKeyMaterial(engine, alias);
                 }
             }
         }
     }
 
-    void setKeyMaterialClientSide(ReferenceCountedOpenSslEngine engine, long certOut, long keyOut, String[] keyTypes,
+    void setKeyMaterialClientSide(ReferenceCountedOpenSslEngine engine, String[] keyTypes,
                                   X500Principal[] issuer) throws SSLException {
         String alias = chooseClientAlias(engine, keyTypes, issuer);
+        setKeyMaterial(engine, alias);
+    }
+
+    private void setKeyMaterial(ReferenceCountedOpenSslEngine engine, String alias) throws SSLException {
         OpenSslKeyMaterial keyMaterial = null;
         try {
             keyMaterial = provider.chooseKeyMaterial(engine.alloc, alias);
             if (keyMaterial != null) {
-                SSL.setKeyMaterialClientSide(engine.sslPointer(), certOut, keyOut,
-                        keyMaterial.certificateChainAddress(), keyMaterial.privateKeyAddress());
+                SSL.setKeyMaterial(engine.sslPointer(),
+                                   keyMaterial.certificateChainAddress(), keyMaterial.privateKeyAddress());
             }
         } catch (SSLException e) {
             throw e;
@@ -115,7 +104,6 @@ final class OpenSslKeyMaterialManager {
             }
         }
     }
-
     private String chooseClientAlias(ReferenceCountedOpenSslEngine engine,
                                        String[] keyTypes, X500Principal[] issuer) {
         X509KeyManager manager = provider.keyManager();
--- a/handler/src/main/java/io/netty/handler/ssl/OpenSslServerContext.java
+++ b/handler/src/main/java/io/netty/handler/ssl/OpenSslServerContext.java
@@ -15,7 +15,6 @@
  */
 package io.netty.handler.ssl;
 
-import io.netty.handler.ssl.ReferenceCountedOpenSslServerContext.ServerContext;
 import io.netty.internal.tcnative.SSL;
 
 import java.io.File;
@@ -37,7 +36,6 @@ import static io.netty.handler.ssl.ReferenceCountedOpenSslServerContext.newSessi
  */
 public final class OpenSslServerContext extends OpenSslContext {
     private final OpenSslServerSessionContext sessionContext;
-    private final OpenSslKeyMaterialManager keyMaterialManager;
 
     /**
      * Creates a new instance.
@@ -349,10 +347,8 @@ public final class OpenSslServerContext extends OpenSslContext {
         // Create a new SSL_CTX and configure it.
         boolean success = false;
         try {
-            ServerContext context = newSessionContext(this, ctx, engineMap, trustCertCollection, trustManagerFactory,
-                                                      keyCertChain, key, keyPassword, keyManagerFactory);
-            sessionContext = context.sessionContext;
-            keyMaterialManager = context.keyMaterialManager;
+            sessionContext = newSessionContext(this, ctx, engineMap, trustCertCollection, trustManagerFactory,
+                                               keyCertChain, key, keyPassword, keyManagerFactory);
             success = true;
         } finally {
             if (!success) {
@@ -364,10 +360,5 @@ public final class OpenSslServerContext extends OpenSslContext {
     @Override
     public OpenSslServerSessionContext sessionContext() {
         return sessionContext;
-    }
-
-    @Override
-    OpenSslKeyMaterialManager keyMaterialManager() {
-        return keyMaterialManager;
     }
 }
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -20,6 +20,7 @@ import io.netty.buffer.ByteBufAllocator;
 import io.netty.internal.tcnative.Buffer;
 import io.netty.internal.tcnative.SSL;
 import io.netty.util.AbstractReferenceCounted;
+import io.netty.util.CharsetUtil;
 import io.netty.util.ReferenceCounted;
 import io.netty.util.ResourceLeakDetector;
 import io.netty.util.ResourceLeakDetectorFactory;
@@ -142,7 +143,6 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
     // OpenSSL state
     private long ssl;
     private long networkBIO;
-    private boolean certificateSet;
 
     private enum HandshakeState {
         /**
@@ -217,7 +217,6 @@ public class ReferenceCountedOpenSslEngine extends SSLEngine implements Referenc
     private final Certificate[] localCerts;
     private final ByteBuffer[] singleSrcBuffer = new ByteBuffer[1];
     private final ByteBuffer[] singleDstBuffer = new ByteBuffer[1];
-    private final OpenSslKeyMaterialManager keyMaterialManager;
     private final boolean enableOcsp;
     private int maxWrapOverhead;
     private int maxWrapBufferSize;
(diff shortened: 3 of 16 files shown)


normanmaurer:
This depends on https://github.com/netty/netty-tcnative/pull/388 and https://github.com/netty/netty/pull/8281

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -238,26 +237,76 @@ protected void deallocate() {
...
+                                        String converted = SignatureAlgorithmConverter.toJavaName(alg);
+                                        if (converted != null) {
+                                            algorithmList.add(converted);
+                                        }
+                                    }
+                                    peerSupportedSignatureAlgorithms = algorithmList.toArray(new String[0]);
+                                }
+                            }
+                        }
+                        return peerSupportedSignatureAlgorithms.clone();
```
carl-mastrangelo: why clone?

handler/src/main/java/io/netty/handler/ssl/SignatureAlgorithmConverter.java
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Converts OpenSSL signature Algorithm names to
+ * <a href="https://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html#Signature">
+ *     Java signature Algorithm names</a>.
+ */
+final class SignatureAlgorithmConverter {
+
+    private SignatureAlgorithmConverter() { }
+
+    // OpenSSL has 3 different formats it uses at the moment we will match against all of these.
```
carl-mastrangelo: Why not just enumerate them in the regex then?

handler/src/main/java/io/netty/handler/ssl/SignatureAlgorithmConverter.java
```diff
@@ -0,0 +1,55 @@
...
+     */
+    static String toJavaName(String opensslName) {
+        if (opensslName == null) {
+            return null;
+        }
+        Matcher matcher = PATTERN.matcher(opensslName);
+        if (matcher.matches()) {
+            String group2 = matcher.group(2);
+            if (group2 != null) {
+                return group2.toUpperCase(Locale.US) + "with" + matcher.group(3).toUpperCase(Locale.US);
```
carl-mastrangelo: You probably want Locale.ROOT I think

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
normanmaurer: because otherwise it will be possible to modify the array content by the caller ?

handler/src/main/java/io/netty/handler/ssl/SignatureAlgorithmConverter.java
normanmaurer: @carl-mastrangelo not sure I understand... Can you show me some code ?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
carl-mastrangelo: Right, but you only return an empty array (immutable), or a new copy (okay to modify).

normanmaurer: True

handler/src/main/java/io/netty/handler/ssl/SignatureAlgorithmConverter.java
carl-mastrangelo: Make the regex `"^(SHA256withECDSA|SHA512withRSA|SHA256withDSA)$"`

normanmaurer: Ah ok... maybe the comment is misleading. It supports more then 3 algorithms. It has just 3 different way to format the supported algorithms

handler/src/main/java/io/netty/handler/ssl/SignatureAlgorithmConverter.java
normanmaurer: addressed and rebased.

normanmaurer:
@carl-mastrangelo PTAL again

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
carl-mastrangelo: up to you, but this looks the same as before.  I won't belabor it.

normanmaurer: I forgot to change this... will do now.

normanmaurer: @carl-mastrangelo ah sorry no I did not forget... I need the close as this field is cached and so we may return the same again and so we need a clone() . I may be misleading because I did not add a `this.` here

handler/src/main/java/io/netty/handler/ssl/OpenSslKeyMaterialManager.java
```diff
@@ -115,7 +104,6 @@ void setKeyMaterialClientSide(ReferenceCountedOpenSslEngine engine, long certOut
             }
         }
     }
-
```
trustin: nit: Better leaving this as is?

handler/src/main/java/io/netty/handler/ssl/OpenSslServerContext.java
```diff
@@ -349,10 +347,8 @@ private OpenSslServerContext(
         // Create a new SSL_CTX and configure it.
         boolean success = false;
         try {
-            ServerContext context = newSessionContext(this, ctx, engineMap, trustCertCollection, trustManagerFactory,
+            sessionContext = newSessionContext(this, ctx, engineMap, trustCertCollection, trustManagerFactory,
                                                       keyCertChain, key, keyPassword, keyManagerFactory);
```
trustin: Indentation

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -238,26 +237,76 @@ protected void deallocate() {
...
+                            }
+                            return requestedServerNames;
+                        }
+                    }
+                }
+
+                @Override
+                public String[] getPeerSupportedSignatureAlgorithms() {
+                    synchronized (ReferenceCountedOpenSslEngine.this) {
+                        if (peerSupportedSignatureAlgorithms == null) {
```
trustin: More readable by returning early when `peerSupportedSignatureAlgorithms != null`?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -238,26 +237,76 @@ protected void deallocate() {
...
+                            if (isDestroyed()) {
+                                peerSupportedSignatureAlgorithms = EmptyArrays.EMPTY_STRINGS;
```
trustin: No need to clone after this

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -238,26 +237,76 @@ protected void deallocate() {
...
+                            } else {
+                                String[] algs = SSL.getSigAlgs(ssl);
+                                if (algs == null) {
+                                    peerSupportedSignatureAlgorithms = EmptyArrays.EMPTY_STRINGS;
```
trustin: No need to clone since the array is empty

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -238,26 +237,76 @@ protected void deallocate() {
...
+                                } else {
+                                    List<String> algorithmList = new ArrayList<String>(algs.length);
+                                    for (String alg: algs) {
+                                        String converted = SignatureAlgorithmConverter.toJavaName(alg);
+                                        if (converted != null) {
+                                            algorithmList.add(converted);
+                                        }
+                                    }
+                                    peerSupportedSignatureAlgorithms = algorithmList.toArray(new String[0]);
```
trustin: - No need to clone either because `toArray()` will return a new copy.
- `EMPTY_STRINGS` instead of `new String[0]`? (Ignore if this is necessary for performance)

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslServerContext.java
```diff
@@ -191,19 +181,44 @@ static ServerContext newSessionContext(ReferenceCountedOpenSslContext thiz, long
...
+
+        @Override
+        public void handle(long ssl, byte[] keyTypeBytes, byte[][] asn1DerEncodedPrincipals) throws Exception {
+            final ReferenceCountedOpenSslEngine engine = engineMap.get(ssl);
+            try {
+                // For now we just ignore the asn1DerEncodedPrincipals as this is kind of inline with what the
+                // OpenJDK SSLEngineImpl does.
+                keyManagerHolder.setKeyMaterialServerSide(engine);
+            } catch (Throwable cause) {
+                logger.debug("request of key failed", cause);
```
trustin: How about: `"Failed to set the server-side key material:", cause

handler/src/main/java/io/netty/handler/ssl/SignatureAlgorithmConverter.java
trustin: I guess the comment is correct - it says 'formats' which is not 'algorithms', but it's probably better giving some examples in the comment.

handler/src/main/java/io/netty/handler/ssl/SignatureAlgorithmConverter.java
```diff
@@ -0,0 +1,55 @@
...
+     */
+    static String toJavaName(String opensslName) {
+        if (opensslName == null) {
+            return null;
+        }
+        Matcher matcher = PATTERN.matcher(opensslName);
+        if (matcher.matches()) {
+            String group2 = matcher.group(2);
+            if (group2 != null) {
+                return group2.toUpperCase(Locale.ROOT) + "with" + matcher.group(3).toUpperCase(Locale.ROOT);
```
trustin: Learned `Locale.ROOT` just today. 👍

handler/src/test/java/io/netty/handler/ssl/OpenSslTestUtils.java
```diff
@@ -24,4 +24,8 @@ private OpenSslTestUtils() {
     static void checkShouldUseKeyManagerFactory() {
         assumeTrue(OpenSsl.supportsKeyManagerFactory() && OpenSsl.useKeyManagerFactory());
     }
+
+    static boolean isBoringSSL() {
+        return "BoringSSL".equals(OpenSsl.versionString());
```
trustin: Curious if we should be case-insensitive here.

handler/src/test/java/io/netty/handler/ssl/SignatureAlgorithmConverterTest.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Test
+    public void testWithUnderscore() {
+        assertEquals("SHA256withDSA", SignatureAlgorithmConverter.toJavaName("dsa_with_SHA256"));
+    }
+
+    @Test
+    public void testInvalid() {
+        assertNull(SignatureAlgorithmConverter.toJavaName("ThisIsSomeThingInvalid"));
```
trustin: nit: SomeThing -> Something

handler/src/main/java/io/netty/handler/ssl/SignatureAlgorithmConverter.java
normanmaurer: done

handler/src/test/java/io/netty/handler/ssl/OpenSslTestUtils.java
normanmaurer: I think its fine ...

normanmaurer:
I will merge this once the CI is done

PR #8286, opened 2018-09-13
Fix get charset from content-type header with multiple parameters, cl…

### LINKED ISSUE #8273: Resolve charset from ContentType header for SOAP 1.2 requests
Hello, there is a problem with charset resolution from `ContentType` header for SOAP 1.2 using `io.netty.handler.codec.http.HttpUtils` class. SOAP 1.2 could have action parameter in `ContentType` header and method `getCharsetAsSequence(CharSequence contentTypeValue)` from `io.netty.handler.codec.http.HttpUtils` class just finds charset word and take whole string from this word end to the end.

### Expected behavior
Resolve charset from `ContentType` header with value `text/xml; charset=utf-8; action="someaction"` as UTF-8

### Actual behavior
Resolve charset substring from `ContentType` header with value `text/xml; charset=utf-8; action="someaction"` as `utf-8; action="someaction"`. And then you will get an exception `java.nio.charset.IllegalCharsetNameException`

### Steps to reproduce

Resolve charset from any request which contains action parameter in `ContentType` header value

### Netty version

4.1.19

java version "1.8.0_181"
Java(TM) SE Runtime Environment (build 1.8.0_181-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.181-b13, mixed mode)

Linux 4.15.0-33-generic Ubuntu 18

amizurov:
Close Resolve charset from ContentType header for SOAP 1.2 requests.

Motivation:

Get charset from Content-Type header even it contains multiple parameters.

Modification:

Extract charset value from the charset parameter if it is not last.

Result:

Fixes #8273

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
@@ -15,17 +15,17 @@
  */
 package io.netty.handler.codec.http;
 
+import java.net.InetSocketAddress;
+import java.net.URI;
+import java.nio.charset.Charset;
+import java.nio.charset.UnsupportedCharsetException;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
 import io.netty.util.AsciiString;
 import io.netty.util.CharsetUtil;
 import io.netty.util.NetUtil;
-
-import java.net.InetSocketAddress;
-import java.net.URI;
-import java.util.ArrayList;
-import java.nio.charset.Charset;
-import java.nio.charset.UnsupportedCharsetException;
-import java.util.Iterator;
-import java.util.List;
 
 /**
  * Utility methods useful in the HTTP context.
@@ -60,12 +60,13 @@ public final class HttpUtil {
     /**
      * Returns {@code true} if and only if the connection can remain open and
      * thus 'kept alive'.  This methods respects the value of the.
+     *
      * {@code "Connection"} header first and then the return value of
      * {@link HttpVersion#isKeepAliveDefault()}.
      */
     public static boolean isKeepAlive(HttpMessage message) {
         CharSequence connection = message.headers().get(HttpHeaderNames.CONNECTION);
-        if (connection != null && HttpHeaderValues.CLOSE.contentEqualsIgnoreCase(connection)) {
+        if (HttpHeaderValues.CLOSE.contentEqualsIgnoreCase(connection)) {
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
```diff
@@ -459,13 +462,23 @@ public static CharSequence getCharsetAsSequence(CharSequence contentTypeValue) {
...
+            return null;
+        }
+
+        int indexOfEncoding = indexOfCharset + CHARSET_EQUALS.length();
+        if (indexOfEncoding < contentTypeValue.length()) {
+            CharSequence charsetCandidate = contentTypeValue.subSequence(indexOfEncoding, contentTypeValue.length());
+            int indexOfSemicolon = AsciiString.indexOfIgnoreCaseAscii(charsetCandidate, SEMICOLON, 0);
+            if (indexOfSemicolon == AsciiString.INDEX_NOT_FOUND) {
+                return charsetCandidate;
+            } else {
```
normanmaurer: nit: You can remove the `else` here as you return in the if block.

codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
```diff
@@ -60,12 +60,13 @@ public static boolean isAsteriskForm(URI uri) {
...
      * Returns {@code true} if and only if the connection can remain open and
      * thus 'kept alive'.  This methods respects the value of the.
+     *
      * {@code "Connection"} header first and then the return value of
      * {@link HttpVersion#isKeepAliveDefault()}.
      */
     public static boolean isKeepAlive(HttpMessage message) {
         CharSequence connection = message.headers().get(HttpHeaderNames.CONNECTION);
-        if (connection != null && HttpHeaderValues.CLOSE.contentEqualsIgnoreCase(connection)) {
+        if (HttpHeaderValues.CLOSE.contentEqualsIgnoreCase(connection)) {
```
normanmaurer: @amizurov can you add comment that passing null into `contentEqualsIngoreCase` is fine ?

codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
amizurov: Thanks, done

codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
amizurov: If we passed null to ```contentEqulasIgnoreCase(...)``` this always return false, null safe method.

normanmaurer:
@amizurov thanks a lot!

PR #8289, opened 2018-09-14
Only use KeyManagerFactory in SniClientTest when supported by OpenSSL…

normanmaurer:
… version.

Motivation:

6ed7c6c75d458047adc37470697f215e9d7436ea added a test which blindly assumed we can use a KeyManagerFactory all the time. This is only true if have OpenSSL 1.0.2 or later, which may not be the case.

Modifications:

Only use KeyManagerFactory in test if the OpenSSL version does support it.

Result:

More robust tests.


normanmaurer:
I will just pull this in once the CI pass to make it happy again and its a test-case change only as well.

PR #8290, opened 2018-09-14
Implemented ExtendedOpenSslSession.getStatusResponses() so it not thr…

normanmaurer:
…ows an UnsupportedOperationException.

Motivation:

6ed7c6c75d458047adc37470697f215e9d7436ea added support for ExtendedOpenSslSession but we did not override getStatusResponses(). This lead to test failures on java9.

Modifications:

Implement ExtendedOpenSslSession.getStatusResponses() so it just returns an empty list.

Result:

Test pass again on Java9.


normanmaurer:
I did not notice this as it only happens on Java9 and I would even argue that this may be a bug.

PR #8297, opened 2018-09-19
Correctly implement ExtendedSSLSession.getStatusResponses() for Refer…

normanmaurer:
…enceCountedOpenSslEngine

Motivation:

Java9 added getStatusResponses() to ExtendedSSLSession which we should correctly support when possible.

Modifications:

Implement the method correctly.

Result:

More complete and correct implementation.


normanmaurer:
Paging @rkapsi on this one

PR #8305, opened 2018-09-21
Ensure we always encode all data in JdkZlibEncoder.

normanmaurer:
Motivation:

In theory our estimation of the needed buffer could be off and so we need to ensure we grow it if there is no space left.

Modifications:

Ensure we grow the buffer if there is no space left in there but we still have data to deflate.

Result:

Correctly deflate data in all cases.


normanmaurer:
@netty-bot test this please

normanmaurer:
@netty-bot test this please

Issue #6664, opened 2017-04-24
Using an X509TrustManager with OPEN_SSL engine bypass hostname verification


vietj:
### Expected behavior

Using an `X509TrustManager` with `OPEN_SSL` engine should perform hostname verification. 

### Actual behavior

With JDK engine, when an X509TrustManager is passed in the `SslContextBuilder`, the returned `X509TrustManager` is wrapped by an `X509ExtendedTrustManager` that performs hostname verification, for instance here is a verification failure trace:

```
Caused by: java.security.cert.CertificateException: No name matching localhost_ found
	at sun.security.util.HostnameChecker.matchDNS(HostnameChecker.java:221)
	at sun.security.util.HostnameChecker.match(HostnameChecker.java:95)
	at sun.security.ssl.X509TrustManagerImpl.checkIdentity(X509TrustManagerImpl.java:455)
	at sun.security.ssl.AbstractTrustManagerWrapper.checkAdditionalTrust(SSLContextImpl.java:998)
	at sun.security.ssl.AbstractTrustManagerWrapper.checkServerTrusted(SSLContextImpl.java:937)
	at sun.security.ssl.ClientHandshaker.serverCertificate(ClientHandshaker.java:1496)
```

When using the `OPEN_SSL` engine, the `checkServerTrusted(X509Certificate[] x509Certificates, String s)` of `X509TrustManager ` is called for verification purpose but it cannot perform the hostname validation as it does not have access to the engine that contains the parameters.

### Steps to reproduce

https://github.com/vietj/netty-openssl-custom-trustmanager-verification-error

- change JDK engine to OPEN_SSL in the `SslContextBuilder`
- run the test : the hostname verification should fail

### Netty version

4.1.8.Final is used in the reproducer but it fails too with 4.1.9.Final

### JVM version (e.g. `java -version`)

java version "1.8.0_92"

### OS version (e.g. `uname -a`)

Darwin julien 15.6.0 Darwin Kernel Version 15.6.0: Fri Feb 17 10:21:18 PST 2017; root:xnu-3248.60.11.4.1~1/RELEASE_X86_64 x86_64
o

normanmaurer:
@vietj would have been nice if it would be a netty only unit test :)

vietj:
ah, I can convert it to pure Netty, I used Vert.x for setting up the server faster, but the important part is in Netty. I'll update it of course Norman, I just focused on providing something quickly :-)

vietj:
[...]

PR #8307, opened 2018-09-22
Do the same extended checks as the JDK when a X509TrustManager is use…

normanmaurer:
…d with the OpenSSL provider.

Motivation:

When a X509TrustManager is used while configure the SslContext the JDK automatically does some extra checks during validation of provided certs by the remote peer. We should do the same when our native implementation is used.

Modification:

- Automatically wrap a X509TrustManager and so do the same validations as the JDK does.
- Add unit tests.

Result:

More consistent behaviour. Fixes https://github.com/netty/netty/issues/6664.


normanmaurer:
@vietj this fixed https://github.com/netty/netty/issues/6664 that you reported a long time ago... I would love an cleaner way to do it but this is the best we can do before they add the extra method (which they may do at some point).

PR #8308, opened 2018-09-23
Bugfix #8295: WebSocket is closed without an error on protocol violations

### LINKED ISSUE #8295: WebSocket is closed without an error on protocol violations
When client sends a message exceeding server restrictions, web socket should be closed with correct error code and message.

### Expected behavior
response frame is sent: CloseWebSocketFrame(1002, "Max frame length of ... has been exceeded")

### Actual behavior
response frame is sent: CloseWebSocketFrame(1000, null)

### Steps to reproduce
Limit single message size accepted by server.
```
static int MAX_CLIENT_MESSAGE_SIZE = 128 * 1024;

ctx.pipeline().addAfter(ctx.name(), "ws-protocol-handler", new WebSocketServerProtocolHandler(
    webSocketPath, null, true, MAX_CLIENT_MESSAGE_SIZE, false, false));`
```

Send message from client with a greater size:
```
ByteBuf buf = Unpooled.wrappedBuffer(new byte[MAX_CLIENT_MESSAGE_SIZE + 1]);
clientChannel.writeAndFlush(new BinaryWebSocketFrame(buf));
```

n/a

### Netty version
4.1.29.Final

java version "1.8.0_181"
Java(TM) SE Runtime Environment (build 1.8.0_181-b25)
Java HotSpot(TM) 64-Bit Server VM (build 25.181-b25, mixed mode)

Linux gmafxu42 3.0.101-108.35-default #1 SMP Mon Feb 19 21:13:08 UTC 2018 (251ec24) x86_64 x86_64 x86_64 GNU/Linux

### Root cause

io.netty.handler.codec.http.websocketx.WebSocket08FrameDecoder#protocolViolation(io.netty.channel.ChannelHandlerContext, io.netty.handler.codec.CorruptedFrameException)

```
private void protocolViolation(ChannelHandlerContext ctx, CorruptedFrameException ex) {
    state = State.CORRUPT;
    --- if (ctx.channel().isActive()) {
    +++ if (ctx.channel().isActive() && closeOnProtocolViolation) {
        Object closeMessage;
        if (receivedClosingHandshake) {
            closeMessage = Unpooled.EMPTY_BUFFER;
        } else {
            --- closeMessage = new CloseWebSocketFrame(1002, null);
            +++ closeMessage = new CloseWebSocketFrame(1002, ex.getMessage());
        }
        --- ctx.writeAndFlush(closeMessage).addListener(ChannelFutureListener.CLOSE);
        +++ ctx.channel().writeAndFlush(closeMessage).addListener(ChannelFutureListener.CLOSE);
    }
    throw ex;
}
```

ursaj:
Motivation:

- Incorrect WebSockets closure affects our production system. 
- Enforced 'close socket on any protocol violation' prevents our custom termination sequence from execution.
- Huge number of parameters is a nightmare both in usage and in support (decoders configuration).

Modification:

- Fix violations handling - send proper response codes.
- Fix violations handling - send response into channel (originally it was sent into context).
- Encapsulate decoder's configuration into a separate class.
- Add new decoder's option - send close frame on protocol violation (enabled by default).

Result:

Fixes #8295.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/CloseWebSocketFrame.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/CloseWebSocketFrame.java
@@ -33,7 +33,31 @@ public class CloseWebSocketFrame extends WebSocketFrame {
     }
 
     /**
-     * Creates a new empty close frame with closing getStatus code and reason text
+     * Creates a new empty close frame with closing status code and reason text
+     *
+     * @param status
+     *            Status code as per <a href="http://tools.ietf.org/html/rfc6455#section-7.4">RFC 6455</a>. For
+     *            example, <tt>1000</tt> indicates normal closure.
+     */
+    public CloseWebSocketFrame(WebSocketCloseStatus status) {
+        this(status.code(), status.reasonText());
+    }
+
+    /**
+     * Creates a new empty close frame with closing status code and reason text
+     *
+     * @param status
+     *            Status code as per <a href="http://tools.ietf.org/html/rfc6455#section-7.4">RFC 6455</a>. For
+     *            example, <tt>1000</tt> indicates normal closure.
+     * @param reasonText
+     *            Reason text. Set to null if no text.
+     */
+    public CloseWebSocketFrame(WebSocketCloseStatus status, String reasonText) {
+        this(status.code(), reasonText);
+    }
+
+    /**
+     * Creates a new empty close frame with closing status code and reason text
      *
      * @param statusCode
      *            Integer status code as per <a href="http://tools.ietf.org/html/rfc6455#section-7.4">RFC 6455</a>. For
@@ -46,7 +70,7 @@ public class CloseWebSocketFrame extends WebSocketFrame {
     }
 
     /**
-     * Creates a new close frame with no losing getStatus code and no reason text
+     * Creates a new close frame with no losing status code and no reason text
      *
      * @param finalFragment
      *            flag indicating if this frame is the final fragment
@@ -105,7 +129,7 @@ public class CloseWebSocketFrame extends WebSocketFrame {
 
     /**
      * Returns the closing status code as per <a href="http://tools.ietf.org/html/rfc6455#section-7.4">RFC 6455</a>. If
-     * a getStatus code is set, -1 is returned.
+     * a status code is set, -1 is returned.
      */
     public int statusCode() {
         ByteBuf binaryData = content();
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/CorruptedWebSocketFrameException.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/CorruptedWebSocketFrameException.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.codec.http.websocketx;
+
+import io.netty.handler.codec.CorruptedFrameException;
+import io.netty.handler.codec.DecoderException;
+
+/**
+ * An {@link DecoderException} which is thrown when the received {@link WebSocketFrame} data could not be decoded by
+ * an inbound handler.
+ */
+public final class CorruptedWebSocketFrameException extends CorruptedFrameException {
+
+    private static final long serialVersionUID = 3918055132492988338L;
+
+    private final WebSocketCloseStatus closeStatus;
+
+    /**
+     * Creates a new instance.
+     */
+    public CorruptedWebSocketFrameException() {
+        this(WebSocketCloseStatus.PROTOCOL_ERROR, null, null);
+    }
+
+    /**
+     * Creates a new instance.
+     */
+    public CorruptedWebSocketFrameException(WebSocketCloseStatus status, String message, Throwable cause) {
+        super(message == null ? status.reasonText() : message, cause);
+        closeStatus = status;
+    }
+
+    /**
+     * Creates a new instance.
+     */
+    public CorruptedWebSocketFrameException(WebSocketCloseStatus status, String message) {
+        this(status, message, null);
+    }
+
+    /**
+     * Creates a new instance.
+     */
+    public CorruptedWebSocketFrameException(WebSocketCloseStatus status, Throwable cause) {
+        this(status, null, cause);
+    }
+
+    public WebSocketCloseStatus getCloseStatus() {
+        return closeStatus;
+    }
+
+}
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocket08FrameDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocket08FrameDecoder.java
@@ -58,8 +58,8 @@ import io.netty.buffer.Unpooled;
 import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.codec.ByteToMessageDecoder;
-import io.netty.handler.codec.CorruptedFrameException;
 import io.netty.handler.codec.TooLongFrameException;
+import io.netty.util.internal.ObjectUtil;
 import io.netty.util.internal.logging.InternalLogger;
 import io.netty.util.internal.logging.InternalLoggerFactory;
 
@@ -93,10 +93,7 @@ public class WebSocket08FrameDecoder extends ByteToMessageDecoder
     private static final byte OPCODE_PING = 0x9;
     private static final byte OPCODE_PONG = 0xA;
 
-    private final long maxFramePayloadLength;
-    private final boolean allowExtensions;
-    private final boolean expectMaskedFrames;
-    private final boolean allowMaskMismatch;
+    private final WebSocketDecoderConfig config;
 
     private int fragmentedFramesCount;
     private boolean frameFinalFlag;
@@ -142,10 +139,22 @@ public class WebSocket08FrameDecoder extends ByteToMessageDecoder
      */
     public WebSocket08FrameDecoder(boolean expectMaskedFrames, boolean allowExtensions, int maxFramePayloadLength,
                                    boolean allowMaskMismatch) {
-        this.expectMaskedFrames = expectMaskedFrames;
-        this.allowMaskMismatch = allowMaskMismatch;
-        this.allowExtensions = allowExtensions;
-        this.maxFramePayloadLength = maxFramePayloadLength;
+        this(WebSocketDecoderConfig.newBuilder()
+            .expectMaskedFrames(expectMaskedFrames)
+            .allowExtensions(allowExtensions)
+            .maxFramePayloadLength(maxFramePayloadLength)
+            .allowMaskMismatch(allowMaskMismatch)
+            .build());
+    }
+
+    /**
+     * Constructor
+     *
+     * @param decoderConfig
+     *            Frames decoder configuration.
+     */
+    public WebSocket08FrameDecoder(WebSocketDecoderConfig decoderConfig) {
+        this.config = ObjectUtil.checkNotNull(decoderConfig, "decoderConfig");
     }
 
     @Override
@@ -184,13 +193,13 @@ public class WebSocket08FrameDecoder extends ByteToMessageDecoder
                     frameMasked = (b & 0x80) != 0;
                     framePayloadLen1 = b & 0x7F;
 
-                    if (frameRsv != 0 && !allowExtensions) {
-                        protocolViolation(ctx, "RSV != 0 and no extension negotiated, RSV:" + frameRsv);
+                    if (frameRsv != 0 && !config.allowExtensions()) {
+                        protocolViolation(ctx, in, "RSV != 0 and no extension negotiated, RSV:" + frameRsv);
                         return;
                     }
 
-                    if (!allowMaskMismatch && expectMaskedFrames != frameMasked) {
-                        protocolViolation(ctx, "received a frame that is not masked as expected");
+                    if (!config.allowMaskMismatch() && config.expectMaskedFrames() != frameMasked) {
+                        protocolViolation(ctx, in, "received a frame that is not masked as expected");
                         return;
                     }
 
@@ -198,20 +207,20 @@ public class WebSocket08FrameDecoder extends ByteToMessageDecoder
 
                         // control frames MUST NOT be fragmented
                         if (!frameFinalFlag) {
-                            protocolViolation(ctx, "fragmented control frame");
+                            protocolViolation(ctx, in, "fragmented control frame");
                             return;
                         }
 
                         // control frames MUST have payload 125 octets or less
                         if (framePayloadLen1 > 125) {
-                            protocolViolation(ctx, "control frame with payload length > 125 octets");
+                            protocolViolation(ctx, in, "control frame with payload length > 125 octets");
                             return;
                         }
 
                         // check for reserved control frame opcodes
                         if (!(frameOpcode == OPCODE_CLOSE || frameOpcode == OPCODE_PING
                                 || frameOpcode == OPCODE_PONG)) {
-                            protocolViolation(ctx, "control frame using reserved opcode " + frameOpcode);
+                            protocolViolation(ctx, in, "control frame using reserved opcode " + frameOpcode);
                             return;
                         }
 
@@ -219,26 +228,26 @@ public class WebSocket08FrameDecoder extends ByteToMessageDecoder
                         // body MUST be a 2-byte unsigned integer (in network byte
                         // order) representing a getStatus code
                         if (frameOpcode == 8 && framePayloadLen1 == 1) {
-                            protocolViolation(ctx, "received close control frame with payload len 1");
+                            protocolViolation(ctx, in, "received close control frame with payload len 1");
                             return;
                         }
                     } else { // data frame
                         // check for reserved data frame opcodes
                         if (!(frameOpcode == OPCODE_CONT || frameOpcode == OPCODE_TEXT
                                 || frameOpcode == OPCODE_BINARY)) {
-                            protocolViolation(ctx, "data frame using reserved opcode " + frameOpcode);
+                            protocolViolation(ctx, in, "data frame using reserved opcode " + frameOpcode);
                             return;
                         }
 
                         // check opcode vs message fragmentation state 1/2
                         if (fragmentedFramesCount == 0 && frameOpcode == OPCODE_CONT) {
-                            protocolViolation(ctx, "received continuation data frame outside fragmented message");
+                            protocolViolation(ctx, in, "received continuation data frame outside fragmented message");
                             return;
                         }
 
                         // check opcode vs message fragmentation state 2/2
                         if (fragmentedFramesCount != 0 && frameOpcode != OPCODE_CONT && frameOpcode != OPCODE_PING) {
-                            protocolViolation(ctx,
+                            protocolViolation(ctx, in,
                                     "received non-continuation data frame while inside fragmented message");
                             return;
                         }
@@ -254,7 +263,7 @@ public class WebSocket08FrameDecoder extends ByteToMessageDecoder
                         }
                         framePayloadLength = in.readUnsignedShort();
                         if (framePayloadLength < 126) {
-                            protocolViolation(ctx, "invalid data frame length (not using minimal length encoding)");
+                            protocolViolation(ctx, in, "invalid data frame length (not using minimal length encoding)");
                             return;
                         }
                     } else if (framePayloadLen1 == 127) {
@@ -266,15 +275,16 @@ public class WebSocket08FrameDecoder extends ByteToMessageDecoder
                         // just check if it's negative?
 
                         if (framePayloadLength < 65536) {
-                            protocolViolation(ctx, "invalid data frame length (not using minimal length encoding)");
+                            protocolViolation(ctx, in, "invalid data frame length (not using minimal length encoding)");
                             return;
                         }
                     } else {
                         framePayloadLength = framePayloadLen1;
                     }
 
-                    if (framePayloadLength > maxFramePayloadLength) {
-                        protocolViolation(ctx, "Max frame length of " + maxFramePayloadLength + " has been exceeded.");
+                    if (framePayloadLength > config.maxFramePayloadLength()) {
+                        protocolViolation(ctx, in, WebSocketCloseStatus.MESSAGE_TOO_BIG,
+                            "Max frame length of " + config.maxFramePayloadLength() + " has been exceeded.");
                         return;
                     }
 
@@ -408,18 +418,31 @@ public class WebSocket08FrameDecoder extends ByteToMessageDecoder
         }
     }
 
-    private void protocolViolation(ChannelHandlerContext ctx, String reason) {
-        protocolViolation(ctx, new CorruptedFrameException(reason));
+    private void protocolViolation(ChannelHandlerContext ctx, ByteBuf in, String reason) {
+        protocolViolation(ctx, in, WebSocketCloseStatus.PROTOCOL_ERROR, reason);
     }
 
-    private void protocolViolation(ChannelHandlerContext ctx, CorruptedFrameException ex) {
+    private void protocolViolation(ChannelHandlerContext ctx, ByteBuf in, WebSocketCloseStatus status, String reason) {
+        protocolViolation(ctx, in, new CorruptedWebSocketFrameException(status, reason));
+    }
+
+    private void protocolViolation(ChannelHandlerContext ctx, ByteBuf in, CorruptedWebSocketFrameException ex) {
         state = State.CORRUPT;
-        if (ctx.channel().isActive()) {
+        int readableBytes = in.readableBytes();
+        if (readableBytes > 0) {
+            in.skipBytes(readableBytes);
+        }
+        if (ctx.channel().isActive() && config.closeOnProtocolViolation()) {
             Object closeMessage;
             if (receivedClosingHandshake) {
                 closeMessage = Unpooled.EMPTY_BUFFER;
             } else {
-                closeMessage = new CloseWebSocketFrame(1002, null);
+                WebSocketCloseStatus closeStatus = ex.getCloseStatus();
+                String reasonText = ex.getMessage();
+                if (reasonText == null) {
+                    reasonText = closeStatus.reasonText();
+                }
+                closeMessage = new CloseWebSocketFrame(closeStatus, reasonText);
             }
             ctx.writeAndFlush(closeMessage).addListener(ChannelFutureListener.CLOSE);
         }
@@ -441,7 +464,7 @@ public class WebSocket08FrameDecoder extends ByteToMessageDecoder
             return;
         }
         if (buffer.readableBytes() == 1) {
-            protocolViolation(ctx, "Invalid close frame body");
+            protocolViolation(ctx, buffer, WebSocketCloseStatus.INVALID_PAYLOAD_DATA, "Invalid close frame body");
         }
 
         // Save reader index
@@ -450,17 +473,16 @@ public class WebSocket08FrameDecoder extends ByteToMessageDecoder
 
         // Must have 2 byte integer within the valid range
         int statusCode = buffer.readShort();
-        if (statusCode >= 0 && statusCode <= 999 || statusCode >= 1004 && statusCode <= 1006
-                || statusCode >= 1015 && statusCode <= 2999) {
-            protocolViolation(ctx, "Invalid close frame getStatus code: " + statusCode);
+        if (!WebSocketCloseStatus.isValidStatusCode(statusCode)) {
+            protocolViolation(ctx, buffer, "Invalid close frame getStatus code: " + statusCode);
         }
 
         // May have UTF-8 message
         if (buffer.isReadable()) {
             try {
                 new Utf8Validator().check(buffer);
-            } catch (CorruptedFrameException ex) {
-                protocolViolation(ctx, ex);
+            } catch (CorruptedWebSocketFrameException ex) {
+                protocolViolation(ctx, buffer, ex);
             }
         }
 
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
@@ -20,10 +20,12 @@ package io.netty.handler.codec.http.websocketx;
  * <pre>
  *
  * RFC-6455 The WebSocket Protocol, December 2011:
- * <a href="https://tools.ietf.org/html/rfc6455#section-7.4.1">https://tools.ietf.org/html/rfc6455#section-7.4.1</a>
+ * <a href="https://tools.ietf.org/html/rfc6455#section-7.4.1"
+ *         >https://tools.ietf.org/html/rfc6455#section-7.4.1</a>
  *
  * WebSocket Protocol Registries, April 2019:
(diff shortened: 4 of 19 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/websocketx/CloseWebSocketFrame.java
```diff
@@ -32,6 +32,30 @@ public CloseWebSocketFrame() {
...
+     * @param status
+     *            Status code as per <a href="http://tools.ietf.org/html/rfc6455#section-7.4">RFC 6455</a>. For
+     *            example, <tt>1000</tt> indicates normal closure.
+     */
+    public CloseWebSocketFrame(WebSocketCloseStatus status) {
+        this(status.code(), status.reasonText());
+    }
+
+    /**
+     * Creates a new empty close frame with closing getStatus code and reason text
```
normanmaurer: s/getStatus/status/

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
```diff
@@ -165,23 +165,24 @@
...
 
-    public int getCode() {
-        return code;
+    public int code() {
+        return statusCode;
     }
 
-    public String getMessage() {
-        return message == null ? "Close status #" + code : message;
+    public String reasonText() {
```
normanmaurer: Please revert this one as it would be a breaking change.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
```diff
@@ -165,23 +165,24 @@
...
         }
-        this.code = code;
-        this.message = message;
+        this.statusCode = statusCode;
+        this.reasonText = reasonText;
     }
 
-    public int getCode() {
-        return code;
+    public int code() {
```
normanmaurer: please revert this one as it would be a breaking change

normanmaurer:
@netty-bot test this please

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocket08FrameDecoder.java
```diff
@@ -409,19 +418,28 @@ private void unmask(ByteBuf frame) {
...
-                closeMessage = new CloseWebSocketFrame(1002, null);
+                WebSocketCloseStatus closeStatus = ex.getCloseStatus();
+                String reasonText = ex.getMessage();
+                if (reasonText == null || reasonText.isEmpty()) {
+                    reasonText = closeStatus.reasonText();
+                }
+                closeMessage = new CloseWebSocketFrame(closeStatus, reasonText);
             }
-            ctx.writeAndFlush(closeMessage).addListener(ChannelFutureListener.CLOSE);
+            ctx.channel().writeAndFlush(closeMessage).addListener(ChannelFutureListener.CLOSE);
```
normanmaurer: revert this ... `ctx.writeAndFlush(...)` is correct.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
```diff
@@ -0,0 +1,236 @@
...
+ *
+ * 4000-4999
+ *
+ *    Status codes in the range 4000-4999 are reserved for private use
+ *    and thus can't be registered.  Such codes can be used by prior
+ *    agreements between WebSocket applications.  The interpretation of
+ *    these codes is undefined by this protocol.
+ * </pre>
+ */
+public final class WebSocketCloseStatus {
```
normanmaurer: Why not make this an enum ?

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
```diff
@@ -0,0 +1,236 @@
...
+ *
+ *    Status codes in the range 3000-3999 are reserved for use by
+ *    libraries, frameworks, and applications.  These status codes are
+ *    registered directly with IANA.  The interpretation of these codes
+ *    is undefined by this protocol.
+ *
+ * 4000-4999
+ *
+ *    Status codes in the range 4000-4999 are reserved for private use
+ *    and thus can't be registered.  Such codes can be used by prior
```
normanmaurer: nit: double space

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
```diff
@@ -0,0 +1,236 @@
...
+ *    agreements between WebSocket applications.  The interpretation of
```
normanmaurer: nit: double-space

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
```diff
@@ -0,0 +1,236 @@
...
+
+    public static WebSocketCloseStatus valueOf(int code) {
+        return valueOf(code, null);
+    }
+
+    public static WebSocketCloseStatus valueOf(int code, String message) {
+        WebSocketCloseStatus status = VALUES.get(code);
+        if (status == null) {
+            VALUES.putIfAbsent(code, new WebSocketCloseStatus(code, message));
+            status = VALUES.get(code);
```
normanmaurer: You should use the return value of `putIfAbsent`. There is no need to call `get`.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
```diff
@@ -0,0 +1,236 @@
...
+        }
+        if (message != null) {
+            status.reasonText = message;
```
normanmaurer: This is racy and may even override the reasonText if it was already in the map.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketDecoderConfig.java
```diff
@@ -0,0 +1,144 @@
+/*
+ * Copyright 2012 The Netty Project
```
normanmaurer: 2018

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketServerHandshaker.java
```diff
@@ -75,12 +75,11 @@
...
      *            CSV of supported protocols. Null if sub protocols not supported.
-     * @param maxFramePayloadLength
-     *            Maximum length of a frame's payload
+     * @param decoderConfig
+     *            Frames decoder configuration.
      */
     protected WebSocketServerHandshaker(
-            WebSocketVersion version, String uri, String subprotocols,
-            int maxFramePayloadLength) {
+            WebSocketVersion version, String uri, String subprotocols, WebSocketDecoderConfig decoderConfig) {
```
normanmaurer: we need to preserve the old constructor as otherwise it would be a breaking change.

codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatusTest.java
```diff
@@ -0,0 +1,129 @@
...
+            PROTOCOL_ERROR.code(),
+            INVALID_MESSAGE_TYPE.code(),
+            INVALID_PAYLOAD_DATA.code(),
+            POLICY_VIOLATION.code(),
+            MESSAGE_TOO_BIG.code(),
+            MANDATORY_EXTENSION.code(),
+            INTERNAL_SERVER_ERROR.code()
+        );
+
+        SortedSet<Integer> invalidCodes = new TreeSet<Integer>();
```
normanmaurer: `Set`

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
ursaj: This is a new class, introduced in this PR. It could NOT cause breaking change.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
ursaj: The same - this is a new class, introduced in this PR. It could NOT cause breaking change.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
ursaj: To unify both predefined and application-specific (code >= 3000) close statuses in the same way. E.g. our application terminates web socket connection with "Feed timed out" or "Out of liquidity" status codes. I want to use identical approach for these tasks.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
ursaj: exactly what I need. I want to override reasonText applications-wide.
in 99.9% cases nobody will use this feature and close statuses will be registered once per app start. however in the last 0.1% it would be impossible to override default 'reasonText' without such approach. See 'testUserLaterDefinedStatuses' test as an example.

codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatusTest.java
ursaj: 'Set' does NOT have '.first()' and '.last()' methods used by this test. SortedSet is used intentionally.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocket08FrameDecoder.java
ursaj: As I see in 'WebSocketServerHandshaker#handshake' codecs are added in the following order:

    p.addBefore(ctx.name(), "wsdecoder", newWebsocketDecoder());
    p.addBefore(ctx.name(), "wsencoder", newWebSocketEncoder());

this means, when decoder calls 'ctx.writeAndFlush' it will never pass through encoder and will be terminated in 'AbstractNioByteChannel.filterOutboundMessage'.

ursaj: anyway, in our applications we will use 'closeOnProtocolViolation=false', so It would not be a problem. 
reverted.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/CloseWebSocketFrame.java
ursaj: fixed

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
ursaj: fixed

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
ursaj: re-worked and simplified

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
ursaj: re-worked and simplified

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketDecoderConfig.java
ursaj: fixed for all new classes

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketServerHandshaker.java
ursaj: reverted both constructor and 'maxFramePayloadLength ' method

normanmaurer:
@ursaj I will try to have another look tomorrow. I will cut a release this week but I am not sure we will get this in in time so it may only make it in the next one.

codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatusTest.java
normanmaurer: Ah sorry mist this.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/CorruptedWebSocketFrameException.java
```diff
@@ -0,0 +1,64 @@
...
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.codec.http.websocketx;
+
+import io.netty.handler.codec.CorruptedFrameException;
+import io.netty.handler.codec.DecoderException;
+
+/**
+ * An {@link DecoderException} which is thrown when the received WebSocket frame data could not be decoded by
```
normanmaurer: {@link WebSocketFrame}

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/CorruptedWebSocketFrameException.java
```diff
@@ -0,0 +1,64 @@
...
+ * an inbound handler.
+ */
+public class CorruptedWebSocketFrameException extends CorruptedFrameException {
```
normanmaurer: mark as final

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocket08FrameDecoder.java
```diff
@@ -142,10 +138,22 @@ public WebSocket08FrameDecoder(boolean expectMaskedFrames, boolean allowExtensio
...
+    }
+
+    /**
+     * Constructor
+     *
+     * @param decoderConfig
+     *            Frames decoder configuration.
+     */
+    public WebSocket08FrameDecoder(WebSocketDecoderConfig decoderConfig) {
+        this.config = decoderConfig;
```
normanmaurer: `ObjectUtil.checkNotNull(...)`

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocket08FrameDecoder.java
```diff
@@ -409,17 +418,26 @@ private void unmask(ByteBuf frame) {
...
+        if (ctx.channel().isActive() && config.closeOnProtocolViolation()) {
             Object closeMessage;
             if (receivedClosingHandshake) {
                 closeMessage = Unpooled.EMPTY_BUFFER;
             } else {
-                closeMessage = new CloseWebSocketFrame(1002, null);
+                WebSocketCloseStatus closeStatus = ex.getCloseStatus();
+                String reasonText = ex.getMessage();
+                if (reasonText == null || reasonText.isEmpty()) {
+                    reasonText = closeStatus.reasonText();
```
normanmaurer: I think we should only "override" if null

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
```diff
@@ -0,0 +1,238 @@
...
+
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ConcurrentSkipListMap;
+
+/**
+ * WebSocket status codes specified in RFC-6455.
+ * <pre>
+ *
+ * RFC-6455 The WebSocket Protocol, December 2011:
+ * https://tools.ietf.org/html/rfc6455#section-7.4.1
```
normanmaurer: use `<a href=....`

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
```diff
@@ -0,0 +1,238 @@
...
+        if (status == null) {
+            status = new WebSocketCloseStatus(code, "Close status #" + code);
+            VALUES.putIfAbsent(code, status);
+        }
+        return status;
+    }
+
+    public static WebSocketCloseStatus register(int code, String reasonText) {
+        WebSocketCloseStatus status = new WebSocketCloseStatus(code, reasonText);
+        VALUES.put(code, status);
```
normanmaurer: I am not sure how I feel about overriding stuff in here ... @trustin @carl-mastrangelo @ejona86 thoughts ?

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
```diff
@@ -0,0 +1,238 @@
...
+            1000 <= code && code <= 1003 ||
+            1007 <= code && code <= 1011 ||
+            3000 <= code;
+    }
+
+    public static WebSocketCloseStatus valueOf(int code) {
+        WebSocketCloseStatus status = VALUES.get(code);
+        if (status == null) {
+            status = new WebSocketCloseStatus(code, "Close status #" + code);
+            VALUES.putIfAbsent(code, status);
```
normanmaurer: I think you should respect what the method returns and so always return the same instance when possible

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
```diff
@@ -0,0 +1,238 @@
...
+
+    @Override
+    public int hashCode() {
+        return statusCode;
+    }
+
+    @Override
+    public String toString() {
+        // E.g.: "1000 Bye", "1009 Message too big"
+        return code() + " " + reasonText();
```
normanmaurer: s/" "/' '/

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketServerHandshakerFactory.java
```diff
@@ -98,11 +94,29 @@ public WebSocketServerHandshakerFactory(
...
+     *            Frames decoder options.
+     */
+    public WebSocketServerHandshakerFactory(
+            String webSocketURL, String subprotocols, WebSocketDecoderConfig decoderConfig) {
         this.webSocketURL = webSocketURL;
         this.subprotocols = subprotocols;
-        this.allowExtensions = allowExtensions;
-        this.maxFramePayloadLength = maxFramePayloadLength;
-        this.allowMaskMismatch = allowMaskMismatch;
+        this.decoderConfig = decoderConfig;
```
normanmaurer: `ObjectUtil.checkNotNull(...)`

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketServerProtocolHandler.java
```diff
@@ -138,13 +136,20 @@ public WebSocketServerProtocolHandler(String websocketPath, String subprotocols,
...
+    public WebSocketServerProtocolHandler(String websocketPath, String subprotocols, boolean checkStartsWith,
+                                          boolean dropPongFrames, WebSocketDecoderConfig decoderConfig) {
         super(dropPongFrames);
         this.websocketPath = websocketPath;
         this.subprotocols = subprotocols;
-        this.allowExtensions = allowExtensions;
-        maxFramePayloadLength = maxFrameSize;
-        this.allowMaskMismatch = allowMaskMismatch;
         this.checkStartsWith = checkStartsWith;
+        this.decoderConfig = decoderConfig;
```
normanmaurer: `ObjectUtil.checkNotNull(...)`

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
ursaj: Assert.assertEquals("1000 Bue", NORMAL_CLOSURE.toString());
    
    org.junit.ComparisonFailure: 
        Expected :1000 Bue
        Actual   :1032Bye

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocket08FrameDecoder.java
ursaj: fixed

ursaj:
all `this.decoderConfig = ...` guarded with checkNotNull

ursaj:
New test on protocol violation discovered resource leak in original code. Fixed.

ursaj:
hi @normanmaurer . is there anything I can do (help with) to move forward?

normanmaurer:
@ursaj just returned from vacation. Will check soon

trustin:
This pull request mixes two changes - one for fixing the sudden disconnection and the other for introducing config object. Would it ever be possible to split this pull request into two for focused reviews?

ursaj:
> This pull request mixes two changes - one for fixing the sudden disconnection and the other for introducing config object. Would it be possible to split this pull request into two for focused reviews?

It looks meaningless for me. Could we please proceed with a single PR?
If you have any questions to clarify changes (for focused review) - u'r welcome, I'll do my best to answer.

By the way, it contains not two but five changes:
- fix for sudden disconnection (message max size close code)
- introducing config object for decoders 
- introducing WebSocketCloseStatus
- introducing an option to disable default behavior for protocol violations
- fix for messages leak

trustin:
That's fine if you are fine. I was suggesting so since I wanted to suggest some refactoring for the config object while other changes looked good to me.

Unless there's a good reason to let a user create a config object by him/herself, how about using a builder pattern instead? e.g.

```java
public class WebSocketFrameDecoderBuilder {
    public WebSocketFrameDecoderBuilder(/* some mandatory configuration parameters, e.g. protocol version */) { ... }

    // Setters
    // ...

    public WebSocketFrameDecoder build() { ... }
}

// User code, for example:
WebSocketFrameDecoder wsdecoder = new WebSocketFrameDecoderBuilder(WebSocketVersion.V13)
        .expectMaskFrames(false)
        .allowExtensions(true)
        .maxFramePayloadLength(...)
        .build();
```

trustin:
We may want to decide between two builder API styles though:

```java
// 1)
public class MyObject {
    public static MyObject.Builder builder() { ... }
}
// 2)
public class MyObject { ... }
public class MyObjectBuilder { ... }
```

I prefer the second because the LOC of a single Java file can be bigger for when there are many things to configure. Opinions?

ursaj:
> Unless there's a good reason to let a user create a config object by him/herself, how about using a builder pattern instead?

The key point is exactly opposite - there is NO good reason to NOT use config object. I want to describe decoder's behavior only once. I don't want to care about actual protocol or implementation underneath, making current api more developer-friendly:

```java
WebSocketDecoderConfig decoderConfig = WebSocketDecoderConfig.newBuilder()
    .allowMaskMismatch(true)
    .maxFramePayloadLength(16384)
    .closeOnProtocolViolation(false)
    .build();

pipeline.addLast(new WebSocketServerProtocolHandler("/test", null, false, false, decoderConfig));
```

Also I don't like the current approach - huge parameter chains copy-pasting (across handshaker and decoder implementations). Encapsulation of these options into single 'config object' looks much more convenient, reliable and simpler for further evolution.

ursaj:
> We may want to decide between two builder API styles though

I've used the same approach, as already exist in the code (to keep consistency). See ProtobufDecoderNano or WorldClockClientHandler. Actually it is a Protocol Buffer API, but it is already used.

> I prefer the second because the LOC of a single Java file can be bigger when there are many things to configure. Opinions?
Even, when there are 50+ options to configure - it will be simpler to keep consistency within a single file, then with separate entities. It's just my personal experience, I don't insist on this approach.

We need the third opinion - to keep current variant or to change.

trustin:
I now understand that you want to build a WebSocket*ProtocolHandler rather than a decoder. Then, why don't we just add:

- WebSocketProtocolHandlerBuilder (abstract, provides common setters)
- WebSocketServerProtocolHandlerBuilder (extends WebSocketProtocolHandlerBuilder, provides server-specific setters)
- WebSocketClientProtocolHandlerBuilder (extends WebSocketProtocolHandlerBuilder, provides client-specific setters)

I do not prefer providing a builder API for an intermediary object (config in this case) that's actually meant to be fed to another class because:

- it increases the level of indirection unnecessarily. Why build A when what you really wanna build is B?
- more public class means less freedom at the implementation level & more API to maintain as we evolve in general

If you still think configuration object is useful, please give me a specific counterexample that shows it yields better user experience.

ursaj:
> I do not prefer providing a builder API for an intermediary object (config in this case) that's actually meant to be fed to another class

Let's consider a simple scenario - I want a new option to be passed into decoders. To be precise - 'closeOnProtocolViolation'.

To make this change, I need to modify 8 classes (constructors, fields, configuration passing, etc):
- WebSocketServerHandshaker.java
- WebSocketServerHandshaker00.java
- WebSocketServerHandshaker07.java
- WebSocketServerHandshaker08.java
- WebSocketServerHandshaker13.java
- WebSocketServerHandshakerFactory.java
- WebSocketServerProtocolHandler.java
- WebSocketServerProtocolHandshakeHandler.java

The same for affected decoders:
- WebSocket07FrameDecoder.java
- WebSocket08FrameDecoder.java
- WebSocket13FrameDecoder.java

In total - 13 files changed + a lot of copy-pasting + very long constructor argument lists.

On the other hand, if we have config object, then the same change would be trivial:
- WebSocketConfigObject.java - add new option into config and builder
- WebSocket08FrameDecoder.java - apply new option where it is really needed

Imho, this simplifies code changes and support, when new option does NOT affect any intermediate code.

> Why build A when what you really wanna build is B?

a) Because A is re-used across multiple B's. 

Indeed, you always can keep such configuration on app-level and copy-paste it on pipeline build. But it's inconvenient and rather should be provided by framework. 

b) Because you can override only those options, you really care about. Leaving the rest "by default". 
E.g. I want to allow mask mismatch:

```java
public WebSocketServerProtocolHandler(String websocketPath, String subprotocols,
        boolean allowExtensions, int maxFrameSize, boolean allowMaskMismatch) {
```

But I'm enforced to explicitly set all intermediate parameters..
Or even worse - I don't won't to instantly close channel on protocol violation:

```java
public WebSocketServerProtocolHandler(String websocketPath, String subprotocols,
                                      boolean allowExtensions, int maxFrameSize, boolean allowMaskMismatch,
                                      boolean checkStartsWith, boolean dropPongFrames, 
                                      boolean closeOnProtocolViolation) {
```

Eight arguments in constructor, where only the first one and last one are meaningful? It's ridiculous...

c) Because you can override any intermediate class without worry about constructors hell.

d) Because 'config object' is much more readable:

```java
pipeline.addLast(new WebSocketServerProtocolHandler("/test", null, false, 65536, false, false, true))

vs

private final WebSocketDecoderConfig decoderConfig = WebSocketDecoderConfig.newBuilder()
                .maxFramePayloadLength(65536)
                .allowMaskMismatch(false)
                .allowExtensions(false)
                .closeOnProtocolViolation(false)
                .build();
...
pipeline.addLast(new WebSocketServerProtocolHandler("/test", null, false, true, decoderConfig))
```

BTW, I like your idea about WebSocket*ProtocolHandlerBuilder:

```java
pipeline.addLast(WebSocketServerProtocolHandler.newBuilder()
    .path("/test").decoderConfig(decoderConfig).build())
```

But even so, I think handler builder and decoder config should be independent.

trustin:
How about making the config class package-private so it's hidden from the public API? You can still reuse most of the code while keeping the public API minimal, right?

Re: reusability of configuration - a user can write a function that returns a builder.

ursaj:
> How about making the config class package-private so it's hidden from the public API? You can still reuse most of the code while keeping the public API minimal, right?
It's constructor is private. You cannot create or even inherit it directly. The same for builder. This approach is extremely safe for public api evolution, so you don't need to worry so much.

> Re: reusability of configuration - a user can write a function that returns a builder.

As I already said - I can keep such configuration on app-level and apply it on pipeline build (with function, with builder, with copy-paste, with any other way). But it is inconvenient.

trustin:
All right. I'm convinced :smiley::+1: Thanks for your patience. Let me do my final review soon.

trustin:
.. and did I say I'm sorry if I was too annoying? I wish I did not discourage your next PR. ;-)

ursaj:
Don't worry, this is not my first PR and definitely will not be the last =)
Do we need anything else to get it merged?

normanmaurer:
@trustin can you do another review please so we can get this one merged ?

normanmaurer:
@netty-bot test this please

trustin:
Yes. Will review soon.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
ursaj: ok, simplified:
```java
 public static WebSocketCloseStatus valueOf(int code) {
     WebSocketCloseStatus status = VALUES.get(code);
     if (status == null) {
         status = new WebSocketCloseStatus(code, "Close status #" + code);
     }
     return status;
 }
```

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/CorruptedWebSocketFrameException.java
ursaj: Fixed

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/CorruptedWebSocketFrameException.java
ursaj: fixed

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocket08FrameDecoder.java
ursaj: fixed

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
ursaj: fixed as suggested by @trustin

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketServerHandshakerFactory.java
ursaj: fixed

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketServerProtocolHandler.java
ursaj: fixed

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
ursaj: fixed

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
ursaj: fixed

ursaj:
@trustin, @normanmaurer, @ejona86 hi, guys. could we merge this PR? or are some changes still needed?

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketCloseStatus.java
ejona86: I was expecting that you _wouldn't_ compare them by reference. So the documentation would just mention that they shouldn't be compared by reference. Instead, either equals() should compare their `code` or the user compares the `code` directly.

ursaj: Note was added into class docs.

ursaj:
@trustin, @normanmaurer, @ejona86 hi, guys. Any updates on this PR?

ursaj:
@netty-bot test this please

ursaj:
@trustin, @normanmaurer, @ejona86 
hi, guys. I've updated this PR once again (resolved conflicts). could we merge it?
If no, what should I do to get it merged into `origin/4.1` branch?

ursaj:
Re-opened this patch as a new PR #9116

PR #8314, opened 2018-09-25
Add cache for CNAME mappings resolved during lookup of DNS entries.

normanmaurer:
Motivation:

If the CNAMEd hostname is backed by load balancing component, typically the final A or AAAA DNS records have small TTL. However, the CNAME record itself is setup with longer TTL.

For example:
* x.netty.io could be CNAMEd to y.netty.io with TTL of 5 min
* A / AAAA records for y.netty.io has a TTL of 0.5 min

In current Netty implementation, original hostname is saved in resolved cached with the TTL of final A / AAAA records. When that cache entry expires, Netty recursive resolver sends at least two queries — 1st one to be resolved as CNAME record and the 2nd one to resolve the hostname in CNAME record.
If CNAME record was cached, only the 2nd query would be needed most of the time. 1st query would be needed less frequently.

Modifications:

Add a new CnameCache that will be used to cache CNAMEs and so may reduce queries.

Result:

Less queries needed when CNAME is used.

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/CnameCache.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/CnameCache.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.resolver.dns;
+
+import io.netty.channel.EventLoop;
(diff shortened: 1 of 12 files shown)


resolver-dns/src/main/java/io/netty/resolver/dns/CnameCache.java
```diff
@@ -0,0 +1,59 @@
...
+package io.netty.resolver.dns;
+
+import io.netty.channel.EventLoop;
+import io.netty.util.internal.UnstableApi;
+
+/**
+ * A cache for {@code CNAME}s.
+ */
+@UnstableApi
+public interface CnameCache {
```
trustin: `DnsCnameCache`? Not strong on this though.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -425,6 +453,13 @@ public DnsCache resolveCache() {
         return resolveCache;
     }
 
+    /**
+     * Returns the cname cache.
```
trustin: - cname cache -> `{@link CnameCache}`?

resolver-dns/src/main/java/io/netty/resolver/dns/CnameCache.java
normanmaurer: sure why not...

normanmaurer:
@trustin addressed.

PR #8315, opened 2018-09-25
Allow to specify a RedirectDnsServerAddressStreamProvider when buildi…

normanmaurer:
…ng a DnsNameResolver.

Motivation:

Sometimes users need to adjust the logic on how we should follow dns redirects (for example adjust the order of the tried nameservers). This is currently possible by extending DnsNameResolver and overriding a method. This is really a bit "odd" as constructing this class directly involves a constructor with a lot of different arguments. We should allow the user to do this via the builder.

Modifications:

- Add a RedirectDnsServerAddressStreamProvider interface and a default implementation
- Add the extra methods to DnsNameResolverBuilder to configure such a provider
- Deprecate the public constructors of DnsNameResolver, as users should just us the builder.

Result:

Easier and cleaner way of adjusting the DnsNameResolver


normanmaurer:
Let me close this in favour of https://github.com/netty/netty/pull/8316

PR #8320, opened 2018-09-26
Update to final Java11 release

normanmaurer:
Motivation:

We should use final Java11 release during builds.

Modifications:

Update to final Java11 release

Result:

Use latest release.


normanmaurer:
Let me just pull this in as it is only a docker change for the java version

PR #8322, opened 2018-09-26
Make it clear that HashedWheelTimer only support millis.

normanmaurer:
Motivation:

    HWT does not support anything smaller then 1ms so we should make it clear that this is the case.

    Modifications:

    Log a warning if < 1ms is used.

    Result:

    Less suprising behaviour.

### DIFF
--- a/common/src/main/java/io/netty/util/HashedWheelTimer.java
+++ b/common/src/main/java/io/netty/util/HashedWheelTimer.java
@@ -84,6 +84,7 @@ public class HashedWheelTimer implements Timer {
     private static final AtomicInteger INSTANCE_COUNTER = new AtomicInteger();
     private static final AtomicBoolean WARNED_TOO_MANY_INSTANCES = new AtomicBoolean();
     private static final int INSTANCE_COUNT_LIMIT = 64;
+    private static final long MILLISECOND_NANOS = TimeUnit.MILLISECONDS.toNanos(1);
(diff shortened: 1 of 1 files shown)


common/src/main/java/io/netty/util/HashedWheelTimer.java
```diff
@@ -279,6 +290,112 @@ public HashedWheelTimer(
         }
     }
 
+    /**
+     * Creates a new timer with the default thread factory
+     * ({@link Executors#defaultThreadFactory()}) and default number of ticks
+     * per wheel.
+     *
+     * @param tickDurationMillis the duration between tick
```
vkostyukov: Maybe say explicitly it's in milliseconds? s/the duration between tick/the duration between ticks in milliseconds/

trustin:
I'm not sure if it's a good idea to remove `TimeUnit` from the constructors. A user may want to use higher tick duration such as 10 seconds or 1 minute. It also seems better to me have `TimeUnit` just like other JDK concurrency API.

How about:

- using 1 ms when a user specifies sub-millisecond tick duration and log a warning about it, or
- rejecting sub-millisecond tick duration by raising `IllegalArgumentException`?

smlance:
@trustin I'm a fan of the first approach since it's less disruptive.

mosesn:
Yeah, let's take the first approach for now, and maybe switch over to the second approach in netty 5?

normanmaurer:
@trustin @stefanlance @mosesn @johnou changed to log a warning only... PTAL again

johnou:
lgtm

PR #8323, opened 2018-09-26
Allow to run tests with leak detection enabled.

normanmaurer:
Motivation:

We should add some command to be able to run all tests with leak detection enabled. This will then be used on the CI during PR builds.

Modifications:

Add new docker-compose config to run with leak-detection enabled.

Result:

Easy way to enable leak detection while running tests via docker.


normanmaurer:
Lets merge so I can use it on the CI.

PR #8325, opened 2018-09-26
Ensure ByteToMessageDecoder.Cumulator implementations always release …

normanmaurer:
…in buffer.

Motivation:

We need to ensure the Cumulator always releases the input buffer if it can not take over the ownership of it as otherwise it may leak.

Modifications:

- Correctly ensure the buffer is always released.
- Add unit tests.

Result:

Ensure buffer is always released.

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
@@ -75,23 +75,28 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
     public static final Cumulator MERGE_CUMULATOR = new Cumulator() {
         @Override
         public ByteBuf cumulate(ByteBufAllocator alloc, ByteBuf cumulation, ByteBuf in) {
-            final ByteBuf buffer;
-            if (cumulation.writerIndex() > cumulation.maxCapacity() - in.readableBytes()
+            try {
+                final ByteBuf buffer;
+                if (cumulation.writerIndex() > cumulation.maxCapacity() - in.readableBytes()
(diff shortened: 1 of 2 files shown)


codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -104,28 +109,37 @@ public ByteBuf cumulate(ByteBufAllocator alloc, ByteBuf cumulation, ByteBuf in)
...
-                    composite.addComponent(true, cumulation);
+                    CompositeByteBuf composite;
+                    if (cumulation instanceof CompositeByteBuf) {
+                        composite = (CompositeByteBuf) cumulation;
+                    } else {
+                        composite = alloc.compositeBuffer(Integer.MAX_VALUE);
+                        composite.addComponent(true, cumulation);
+                    }
+                    composite.addComponent(true, in);
+                    releaseIn = false;
```
njhill: why not just set `in` to null?

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -75,23 +75,28 @@
...
+                    // - https://github.com/netty/netty/issues/2327
+                    // - https://github.com/netty/netty/issues/1764
+                    buffer = expandCumulation(alloc, cumulation, in.readableBytes());
+                } else {
+                    buffer = cumulation;
+                }
+                buffer.writeBytes(in);
+                return buffer;
+            } finally {
+                // We most release in in all cases as otherwise it may produce a leak if writeBytes(...) throw
```
njhill: typo `most` -> `must`

normanmaurer:
@njhill thanks for the review.. done :)

PR #8335, opened 2018-09-28
Avoid repeating the same field and hiding it

### LINKED ISSUE #8331: EpollSocketChannelConfig is repeating the channel field 3x
The`EpollSocketChannelConfig` class is repeating the `channel` field 3x through inheritance. The field has also protected and package-private visibility in the super classes which is potentially bug-prone.

screenshot from 2018-09-27 13-46-22

rkapsi:
Motivation

The EpollChannelConfig (same for KQueues) and its subclasses repeatetly declare their own channel field which leads to a 3x repetition for each config instance. Given the fields are protected or package-private it's exposing the code code to "field hiding" bugs.

Modifications

Use the the existing protected channel field from the DefaultChannelConfig class and simply cast it when needed.

Result

Fixes #8331


normanmaurer:
@netty-bot test this please

PR #8338, opened 2018-10-05
Cleanup: remove unnecessary vararg argument in PooledByteBufAllocator

doom369:
Motivation:

No need in varargs, the method always accepts array.

Modification:

`...` replaced with `[]`


normanmaurer:
@netty-bot test this please

PR #8339, opened 2018-10-05
Cleanup: deprecate IntegerHolder for removal

doom369:
Motivation:

Seems like `IntegerHolder counterHashCode` field is the very old legacy field that is no longer used. Should be marked as deprecated and removed in the future versions.

Modification:

`IntegerHolder` class, `InternalThreadLocalMap.counterHashCode()` and `InternalThreadLocalMap.setCounterHashCode(IntegerHolder counterHashCode)`  are now deprecated.


normanmaurer:
@netty-bot test this please

normanmaurer:
@doom369 thanks!

PR #8344, opened 2018-10-07
Fix SelectableChannel support in NioEventLoop

jvirtanen:
Motivation:

Unless the `io.netty.noKeySetOptimization` system property is set, registering a SelectableChannel instance to a NioEventLoop results in a ClassCastException:

    io.netty.channel.nio.SelectedSelectionKeySetSelector cannot be cast to java.nio.channels.spi.AbstractSelector

Modifications:

Instead of `selector`, pass `unwrappedSelector` to SelectableChannel.

Result:

It is possible to register a SelectableChannel instance without setting the `io.netty.noKeySetOptimization` system property.

### DIFF
--- a/transport/src/test/java/io/netty/channel/nio/NioEventLoopTest.java
+++ b/transport/src/test/java/io/netty/channel/nio/NioEventLoopTest.java
@@ -24,7 +24,10 @@ import io.netty.channel.socket.nio.NioServerSocketChannel;
 import io.netty.util.concurrent.Future;
 import org.junit.Test;
 
+import java.net.InetSocketAddress;
+import java.nio.channels.SelectionKey;
 import java.nio.channels.Selector;
+import java.nio.channels.SocketChannel;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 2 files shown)


transport/src/test/java/io/netty/channel/nio/NioEventLoopTest.java
```diff
@@ -132,4 +135,37 @@ public void run() {
...
+                public void channelReady(SocketChannel ch, SelectionKey key) {
+                    latch.countDown();
+                }
+
+                @Override
+                public void channelUnregistered(SocketChannel ch, Throwable cause) {
+                }
+            });
+
+            latch.await();
```
johnou: Await with timeout?

jvirtanen: That could be done, but please note that I set a timeout on the `@Test` annotation so that the whole test case will fail in case of a timeout.

johnou: Works for me.

transport/src/test/java/io/netty/channel/nio/NioEventLoopTest.java
```diff
@@ -132,4 +135,37 @@ public void run() {
...
+        } finally {
+            group.shutdownGracefully();
```
normanmaurer: Also close the channels ?

jvirtanen: Good catch! Fixed.

normanmaurer:
@jvirtanen merged... thx

PR #8347, opened 2018-10-10
ByteBufInputStream is always allocating a StringBuilder instance

franz1981:
Motivation:

Avoid creating any StringBuilder instance if
ByteBufInputStream::readLine isn't used

Modifications:

The StringBuilder instance is lazy allocated on demand and
are added new test case branches to address the increased
complexity of ByteBufInputStream::readLine

Result:

Reduced GC activity if ByteBufInputStream::readLine isn't used

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/ByteBufInputStream.java
+++ b/buffer/src/main/java/io/netty/buffer/ByteBufInputStream.java
@@ -16,6 +16,7 @@
 package io.netty.buffer;
 
 import io.netty.util.ReferenceCounted;
+import io.netty.util.internal.StringUtil;
 
 import java.io.DataInput;
 import java.io.DataInputStream;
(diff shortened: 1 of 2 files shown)


buffer/src/main/java/io/netty/buffer/ByteBufInputStream.java
```diff
@@ -263,11 +268,22 @@ public String readLine() throws IOException {
...
             }
         }
 
-        return lineBuf.toString();
+        return lineBuf != null && lineBuf.length() > 0 ? lineBuf.toString() : "";
+    }
+
+    private static String toStringIfAnyChar(StringBuilder lineBuf, boolean anyChars) {
+        if (anyChars) {
+            return lineBuf != null && lineBuf.length() > 0 ? lineBuf.toString() : "";
```
normanmaurer: Use StringUtil.EMPTY_STRING

buffer/src/main/java/io/netty/buffer/ByteBufInputStream.java
```diff
@@ -263,11 +268,22 @@ public String readLine() throws IOException {
...
                 default:
+                    if (lineBuf == null) {
+                        lineBuf = new StringBuilder();
+                    }
                     lineBuf.append((char) c);
             }
         }
 
-        return lineBuf.toString();
+        return lineBuf != null && lineBuf.length() > 0 ? lineBuf.toString() : "";
```
normanmaurer: Same as below

normanmaurer:
@netty-bot test this please

franz1981:
@normanmaurer Done, thanks!

normanmaurer:
@franz1981 thanks!

Issue #8348, opened 2018-10-11
epoll_wait produces an EINVAL error since 4.1.30


simonbasle:
### Expected behavior
epoll_wait should work in 4.1.30 like it did in 4.1.29

### Actual behavior
Since switching to 4.1.30, `EpollEventLoop`'s `handleLoopException` is triggered with 
 `io.netty.channel.ChannelException: timerfd_settime() failed: Invalid argument`, which points to `timerfd_settime`. This causes an epoll thread to be "blocked" sleeping.

The issue is visible in a Spring Boot test dealing with bad SSL certificates, which uses reactor/reactor-netty.

While investigating this remotely with limited resources (partial access to the logs and reproduction case, no local linux machine to test on), I found that the 4.1.30 suspiciously contained an issue related to epoll_wait.

Looking at the PR I think I might have found the regression:

https://github.com/netty/netty/pull/7816/files#diff-db3e069239a403b954e3ebc024ba9507R251

`Integer.MAX_VALUE` should be `MAX_SCHEDULED_TIMERFD_NS` (`999,999,999`) like it was before the PR, else `timerfd_settime` might return `EINVAL` if it is too large.

### Steps to reproduce
The issue is triggered during tests of Spring Boot, but this is a smaller reproduction snippet that is using Spring Framework 5:

```java
@Test
    public void strippedDown() {
        assertThatExceptionOfType(RuntimeException.class)
                .isThrownBy(() -> WebClient.create().get()
                        .uri("https://" + "self-signed.badssl.com/").exchange()
                        .block(Duration.ofSeconds(10)))
                .withCauseInstanceOf(SSLException.class);
    }
```

I can try to spin up a repository with a maven project that reproduces the issue and can be run without set up if you need.

### Netty version
4.1.30

### JVM version (e.g. `java -version`)
??

### OS version (e.g. `uname -a`)
??

wilkinsona:
I can fill in some of the blanks about the environment where we've seen the issue:

### JVM version (e.g. java -version)

```
openjdk version "1.8.0_181"
OpenJDK Runtime Environment (build 1.8.0_181-8u181-b13-1~deb9u1-b13)
OpenJDK 64-Bit Server VM (build 25.181-b13, mixed mode)
```

### OS version (e.g. name -a)

```
[...]

PR #8350, opened 2018-10-11
epoll_wait produces an EINVAL error since 4.1.30

johnou:
Fixes regression with 6d8745179844caba0f3a0f3d7f3a1fbf392c925f

Closes https://github.com/netty/netty/issues/8348

### DIFF
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
@@ -35,7 +35,6 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Queue;
-import java.util.concurrent.Callable;
 import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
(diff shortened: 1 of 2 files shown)


johnou:
@normanmaurer I verified using debugger it works as expected but had trouble adding a unit test, I tried a bunch of different ways eg.

```java
        Future<Integer> pendingTasks = el.submit(new Callable<Integer>() {
            @Override
            public Integer call() throws Exception {
                return ((EpollEventLoop) el).pendingTasks();
            }
        });
        assertEquals(Integer.valueOf(1), pendingTasks.get(1000, TimeUnit.SECONDS)); // accepted scheduled task
```

but no nice.

normanmaurer:
@netty-bot test this please

johnou:
@normanmaurer yep

screenshot from 2018-10-11 12-21-20

normanmaurer:
@johnou why not just make `EpollEventLoop` (but still keep it package-private) not final and change `handleLoopException` to be not static and package-private (+ add a comment this is because of testing). This way you could extend the class and override and just assert that it is not called. 

WDYT ?

johnou:
@normanmaurer i'll give that a go after lunch :+1:

johnou:
@normanmaurer I could only get it working by removing the final from EpollEventLoopGroup too..

normanmaurer:
@johnou it should not be needed to do this... what was the problem ?

johnou:
@normanmaurer PTAL I swapped out the EpollEventLoopGroup with EpollEventLoop directly in the test to avoid removing final from EpollEventLoopGroup.

normanmaurer:
@johnou this looks great... can you adjust the commit message to match our template tho ?

johnou:
@normanmaurer yep sure, done.

normanmaurer:
@johnou thanks a lot!

carl-mastrangelo:
LGTM

PR #8351, opened 2018-10-11
Correctly parse /etc/resolv.conf when contain multiple entries for se…

normanmaurer:
…archdomain.

Motivation:

ba594bcf4a62c47810f85c6d28e87367c6903ed4 added a utility to parse searchdomains defined in /etc/resolv.conf but did not correctly handle the case when multiple are defined that are seperated by either whitespace or tab.

Modifications:

- Correctly parse multiple entries
- Add unit test.

Result:

Correctly parse multiple searchdomain entries.

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/UnixResolverDnsServerAddressStreamProvider.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/UnixResolverDnsServerAddressStreamProvider.java
@@ -27,12 +27,12 @@ import java.io.FileReader;
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Pattern;
 
 import static io.netty.resolver.dns.DefaultDnsServerAddressStreamProvider.DNS_PORT;
 import static io.netty.util.internal.ObjectUtil.checkNotNull;
@@ -59,6 +59,7 @@ public final class UnixResolverDnsServerAddressStreamProvider implements DnsServ
     static final int DEFAULT_NDOTS = 1;
     private final DnsServerAddresses defaultNameServerAddresses;
     private final Map<String, DnsServerAddresses> domainToNameServerStreamMap;
+    private static final Pattern SEARCH_DOMAIN_PATTERN = Pattern.compile("\\s+");
 
     /**
      * Attempt to parse {@code /etc/resolv.conf} and files in the {@code /etc/resolver} directory by default.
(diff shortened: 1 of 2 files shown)


normanmaurer:
@mattayres  PTAL ...

resolver-dns/src/main/java/io/netty/resolver/dns/UnixResolverDnsServerAddressStreamProvider.java
```diff
@@ -323,7 +325,12 @@ static int parseEtcResolverFirstNdots(File etcResolvConf) throws IOException {
                 } else if (line.startsWith(SEARCH_ROW_LABEL)) {
                     int i = indexOfNonWhiteSpace(line, SEARCH_ROW_LABEL.length());
                     if (i >= 0) {
-                        searchDomains.add(line.substring(i));
+                        // May contain more then one entry, either seperated by whitespace or tab.
+                        // See https://linux.die.net/man/5/resolver
+                        String[] domains = SEARCH_DOMAIN_PATTERN.split(line.substring(i));
+                        if (domains != null) {
+                            searchDomains.addAll(Arrays.asList(domains));
```
trustin: Better using `Collections.addAll(searchDomains, domains)`?

resolver-dns/src/test/java/io/netty/resolver/dns/UnixResolverDnsServerAddressStreamProviderTest.java
```diff
@@ -139,6 +139,24 @@ public void searchDomainsWithMultipleSearch() throws IOException {
         assertEquals(Arrays.asList("linecorp.local", "squarecorp.local"), domains);
     }
 
+    @Test
+    public void searchDomainsWithMultipleSearchSeperatedByWhitespace() throws IOException {
+        File f = buildFile(
+                           "search linecorp.local squarecorp.local\n" +
```
trustin: - Could merge the two lines or format like in `searchDomainsWithMultipleSearchSeparatedByTab`.
- How about merging the two tests into one by adding a `search` directive that contains both spaces and tabs (single space/tab, double space/tab, mixture of spaces and tabs, etc)?

resolver-dns/src/main/java/io/netty/resolver/dns/UnixResolverDnsServerAddressStreamProvider.java
normanmaurer: TIL...

mattayres:
@normanmaurer LGTM.

resolver-dns/src/main/java/io/netty/resolver/dns/UnixResolverDnsServerAddressStreamProvider.java
```diff
@@ -323,7 +324,12 @@ static int parseEtcResolverFirstNdots(File etcResolvConf) throws IOException {
                 } else if (line.startsWith(SEARCH_ROW_LABEL)) {
                     int i = indexOfNonWhiteSpace(line, SEARCH_ROW_LABEL.length());
                     if (i >= 0) {
-                        searchDomains.add(line.substring(i));
+                        // May contain more then one entry, either seperated by whitespace or tab.
+                        // See https://linux.die.net/man/5/resolver
+                        String[] domains = SEARCH_DOMAIN_PATTERN.split(line.substring(i));
+                        if (domains != null) {
```
ejona86: split() always returns a non-null array.

normanmaurer: ah even better :)

PR #8378, opened 2018-10-12
Prevent NPE when attempting to set client key material with no alias

normanmaurer:
Motivation:

It is possible that a client is unable to locate a certificate alias given the list of issuers and key types. In this case the X509KeyManager
will return a null which when past to the OpenSslKeyMaterialProvider implementation may produce a NPE. If no matching alias could be found we should not
call OpenSslKeyMaterialProvider at all which is also consistent what OpenJDK does.

Modifications:

- Add null check before calling OpenSslKeyMaterialProvider
- Add unit test.

Result:

No more NPE caused by passing null as client alias.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/OpenSslKeyMaterialManager.java
+++ b/handler/src/main/java/io/netty/handler/ssl/OpenSslKeyMaterialManager.java
@@ -83,7 +83,12 @@ final class OpenSslKeyMaterialManager {
     void setKeyMaterialClientSide(ReferenceCountedOpenSslEngine engine, String[] keyTypes,
                                   X500Principal[] issuer) throws SSLException {
         String alias = chooseClientAlias(engine, keyTypes, issuer);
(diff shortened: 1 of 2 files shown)


handler/src/main/java/io/netty/handler/ssl/OpenSslKeyMaterialManager.java
```diff
@@ -83,7 +83,12 @@ void setKeyMaterialServerSide(ReferenceCountedOpenSslEngine engine) throws SSLEx
     void setKeyMaterialClientSide(ReferenceCountedOpenSslEngine engine, String[] keyTypes,
                                   X500Principal[] issuer) throws SSLException {
         String alias = chooseClientAlias(engine, keyTypes, issuer);
-        setKeyMaterial(engine, alias);
+        // Only try to set the keymaterial if we have an match. This is also consistent with what OpenJDK does:
```
trustin: an match -> a match

PR #8380, opened 2018-10-12
Simplify ByteBufInputStream.readLine() logic

njhill:
Motivation:

While looking at the nice optimization done in https://github.com/netty/netty/pull/8347 I couldn't help noticing that the logic could be simplified further. Apologies if this is just my OCD and inappropriate for a PR!

Modifications:

Reduce amount of code used for `ByteBufInputStream.readLine()`

Result:

Slightly smaller and simpler code


normanmaurer:
@franz1981 can you check

franz1981:
It looks fine to me!
Having a smaller and simpler code is always wellcome :)

normanmaurer:
@netty-bot test this please

PR #8383, opened 2018-10-15
Include the whole initial line in the exception message when failing …

normanmaurer:
…to create a message from it.

Motivation:

Sometimes we may fail to parse the initial line because some garbage was received. To make it easier to debug we should include the whole initial line in this case in the exception message.

Modifications:

Wrap thrown IllegalArgumentException in another IllegalArgumentException and include the full intial line in the message.

Result:

Easier to debug HTTP decoders.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectDecoder.java
@@ -26,6 +26,7 @@ import io.netty.handler.codec.TooLongFrameException;
 import io.netty.util.ByteProcessor;
 import io.netty.util.internal.AppendableCharSequence;
 
+import java.util.Arrays;
 import java.util.List;
 
 /**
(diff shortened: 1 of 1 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectDecoder.java
```diff
@@ -216,7 +217,13 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf buffer, List<Object> ou
                 return;
             }
 
-            message = createMessage(initialLine);
+            try {
+                message = createMessage(initialLine);
+            } catch (IllegalArgumentException e) {
+                // Include the whole initial line to easier debug.
```
Scottmitch: `to easier` -> `for easier`

vietj:
I think it would be good to use a specific exception for this and not have the initialLine value part of the message that will be printed by the exception when toString() is called. If the Netty user does not take proper care, this can result in being a log injection vector.

normanmaurer:
That may be a good idea… that said it should be a sub-class of IllegalArgumentException just in case people depend on the type of it.

WDYT ?

vietj:
no problem with me, the main issue is to have the initial line in the toString() message

vietj:
Whatever the decision is, I think it remains important to signal this exception so the user can handle it specially and not print the message.

normanmaurer:
@netty-bot test this please

normanmaurer:
Let me just close this for now...

PR #8389, opened 2018-10-16
Add full JdkSslContext public constructor, close #8384

### LINKED ISSUE #8384: Feature request: let one pass a SecureRandom instance when building SslContext
### Expected behavior

When spawning multiple `JdkSslContext`, it should be possible to reuse a single SecureRandom instance and save allocations.

### Actual behavior

`SslContextBuilder` won't let one pass a SecureRandom instance.
Both `JdkSslServerContext#newSSLContext` and `JdkSslClientContext#newSSLContext` pass a null random value to `SSLContext#init` that allocates a new `SecureRandom` instance.

NA

### Netty version

4.1.30

NA

NA

slandelle:
Motivation:

JdkSslContext provides public constructors to wrap an existing `javax.net.ssl.SSLContext`.

Sadly, some options combinations are not possible with the existing constructors, eg:
*  protocols is not exposed and always forced to null, so default protocols are always enforced
* startTls is not exposed and always forced to false

Modification:

Add full constructor that take protocols and startTls parameters.

Result:

It's possible to create a JdkSslContext from an existing SSLContext and still have control over protocols and startTls

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/JdkSslContext.java
+++ b/handler/src/main/java/io/netty/handler/ssl/JdkSslContext.java
@@ -140,7 +140,10 @@ public class JdkSslContext extends SslContext {
      * @param sslContext the {@link SSLContext} to use.
      * @param isClient {@code true} if this context should create {@link SSLEngine}s for client-side usage.
      * @param clientAuth the {@link ClientAuth} to use. This will only be used when {@param isClient} is {@code false}.
+     * @deprecated Use {@link #JdkSslContext(SSLContext, boolean, Iterable, CipherSuiteFilter,
+     * ApplicationProtocolConfig, ClientAuth, String[], boolean)}
      */
+    @Deprecated
     public JdkSslContext(SSLContext sslContext, boolean isClient,
                          ClientAuth clientAuth) {
         this(sslContext, isClient, null, IdentityCipherSuiteFilter.INSTANCE,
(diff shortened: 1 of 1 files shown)


handler/src/main/java/io/netty/handler/ssl/JdkSslContext.java
```diff
@@ -160,7 +160,37 @@ public JdkSslContext(SSLContext sslContext, boolean isClient,
...
+                         ClientAuth clientAuth,
+                         String[] protocols,
+                         boolean startTls) {
+        this(sslContext,
+                isClient,
+                ciphers,
+                cipherFilter,
+                toNegotiator(apn, !isClient),
+                clientAuth,
+                protocols,
```
normanmaurer: @slandelle do we need to call `protocols.clone()` if these are not null ?

handler/src/main/java/io/netty/handler/ssl/JdkSslContext.java
```diff
@@ -160,7 +160,37 @@ public JdkSslContext(SSLContext sslContext, boolean isClient,
     public JdkSslContext(SSLContext sslContext, boolean isClient, Iterable<String> ciphers,
```
normanmaurer: @slandelle maybe mark this constructor as `@deprecated` ?

slandelle: sure

handler/src/main/java/io/netty/handler/ssl/JdkSslContext.java
slandelle: right, that would be consistent with SslContextBuilder

slandelle:
@normanmaurer PTAL

handler/src/main/java/io/netty/handler/ssl/JdkSslContext.java
```diff
@@ -140,7 +140,9 @@
      * @param sslContext the {@link SSLContext} to use.
      * @param isClient {@code true} if this context should create {@link SSLEngine}s for client-side usage.
      * @param clientAuth the {@link ClientAuth} to use. This will only be used when {@param isClient} is {@code false}.
+     * @deprecated Use full constructor
```
normanmaurer: @slandelle please use java docs style via `{@link ... }`

handler/src/main/java/io/netty/handler/ssl/JdkSslContext.java
```diff
@@ -140,7 +140,9 @@
      * @param sslContext the {@link SSLContext} to use.
      * @param isClient {@code true} if this context should create {@link SSLEngine}s for client-side usage.
      * @param clientAuth the {@link ClientAuth} to use. This will only be used when {@param isClient} is {@code false}.
+     * @deprecated Use {@link #JdkSslContext(SSLContext, boolean, Iterable, CipherSuiteFilter, ApplicationProtocolConfig, ClientAuth, String[], boolean)}
```
slandelle: crossing fingers checkstyle won't complain on line length...

normanmaurer:
@netty-bot test this please

normanmaurer:
@slandelle yep lines too long... need to add some link-breaks:

```
16:31:14 [INFO] --- maven-checkstyle-plugin:2.12.1:check (check-style) @ netty-handler ---
16:31:16 [INFO] Starting audit...
16:31:16 /code/handler/src/main/java/io/netty/handler/ssl/JdkSslContext.java:143: Line is longer than 120 characters (found 153).
16:31:16 /code/handler/src/main/java/io/netty/handler/ssl/JdkSslContext.java:161: Line is longer than 120 characters (found 153).
16:31:16 Audit done.
16:31:16 
```

slandelle:
@normanmaurer fixed

normanmaurer:
@netty-bot test this please

normanmaurer:
@slandelle thanks ... merged

slandelle:
Thanks for your support !

PR #8388, opened 2018-10-16
Reduce overhead by ByteBufUtil.decodeString(...) which is used by `Ab…

normanmaurer:
…stractByteBuf.toString(...)` and `AbstractByteBuf.getCharSequence(...)`

Motivation:

Our current implementation that is used for toString(Charset) operations on AbstractByteBuf implementation is quite slow as it does a lot of uncessary memory copies. We should just use new String(...) as it has a lot of optimizations to handle these cases.

Modifications:

Rewrite ByteBufUtil.decodeString(...) to use new String(...)

Result:

Less overhead for toString(Charset) operations.

Benchmark                                         (charsetName)  (direct)  (size)   Mode  Cnt         Score         Error  Units
ByteBufUtilDecodeStringBenchmark.decodeString          US-ASCII     false       8  thrpt   20  22401645.093 ? 4671452.479  ops/s
ByteBufUtilDecodeStringBenchmark.decodeString          US-ASCII     false      64  thrpt   20  23678483.384 ? 3749164.446  ops/s
ByteBufUtilDecodeStringBenchmark.decodeString          US-ASCII      true       8  thrpt   20  15731142.651 ? 3782931.591  ops/s
ByteBufUtilDecodeStringBenchmark.decodeString          US-ASCII      true      64  thrpt   20  16244232.229 ? 1886259.658  ops/s
ByteBufUtilDecodeStringBenchmark.decodeString             UTF-8     false       8  thrpt   20  25983680.959 ? 5045782.289  ops/s
ByteBufUtilDecodeStringBenchmark.decodeString             UTF-8     false      64  thrpt   20  26235589.339 ? 2867004.950  ops/s
ByteBufUtilDecodeStringBenchmark.decodeString             UTF-8      true       8  thrpt   20  18499027.808 ? 4784684.268  ops/s
ByteBufUtilDecodeStringBenchmark.decodeString             UTF-8      true      64  thrpt   20  16825286.141 ? 1008712.342  ops/s
ByteBufUtilDecodeStringBenchmark.decodeString            UTF-16     false       8  thrpt   20   5789879.092 ? 1201786.359  ops/s
ByteBufUtilDecodeStringBenchmark.decodeString            UTF-16     false      64  thrpt   20   2173243.225 ?  417809.341  ops/s
ByteBufUtilDecodeStringBenchmark.decodeString            UTF-16      true       8  thrpt   20   5035583.011 ? 1001978.854  ops/s
[...]

PR #8393, opened 2018-10-17
Exploit PlatformDependent.allocateUninitializedArray() in more places

njhill:
Motivation:

There are currently many more places where this could be used which were possibly not considered when the method was added.

If https://github.com/netty/netty/pull/8388 is included in its current form, a number of these places could additionally make use of the same `BYTE_ARRAYS` threadlocal.

There's also a couple of adjacent places where an optimistically-pooled heap buffer is used for temp byte storage which could use the threadlocal too in preference to allocating a temp heap bytebuf wrapper. For example https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java#L1417.

Modifications:

Replace `new byte[]` with `PlatformDependent.allocateUninitializedArray()` in appropriate places; make use of `ByteBufUtil.getBytes()` in some places which currently perform the equivalent logic, including avoiding copy of backing array if possible (although would be rare).

Result:

Further potential speed-up with java9+ and appropriate compile flags (many of these places could be on high-frequency code paths).

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
+++ b/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
@@ -858,7 +858,7 @@ public final class ByteBufUtil {
             }
         }
 
-        byte[] v = new byte[length];
+        byte[] v = PlatformDependent.allocateUninitializedArray(length);
         buf.getBytes(start, v);
         return v;
     }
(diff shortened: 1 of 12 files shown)


normanmaurer:
@netty-bot test this please

normanmaurer:
@njhill thanks again!

PR #8394, opened 2018-10-17
Allow to use TLSv1.3 with netty-tcnative withe java versions prior to…

normanmaurer:
… 11.

Motivation:

Modification:

Add a workaround to be able to use TLSv1.3 also when using Java version prior to Java 11 and the default X509ExtendedTrustManager is used.

Result:

Be able to use TLSv1.3 also with past versions of Java.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
+++ b/handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
@@ -31,7 +31,6 @@ import io.netty.util.internal.SystemPropertyUtil;
 import io.netty.util.internal.logging.InternalLogger;
 import io.netty.util.internal.logging.InternalLoggerFactory;
 
-import javax.net.ssl.SSLException;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.ArrayList;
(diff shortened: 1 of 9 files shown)


handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
```diff
@@ -231,6 +220,8 @@ public Boolean run() {
             }
 
             addIfSupported(availableJavaCipherSuites, defaultCiphers, DEFAULT_CIPHER_SUITES);
+            addIfSupported(availableJavaCipherSuites, defaultCiphers, TLSV13_CIPHER_SUITES);
```
trustin: Can we merge `DEFAULT_CIPHER_SUITES` and `TLSV13_CIPHER_SUITES`? They are added only when supported anyway.

handler/src/main/java/io/netty/handler/ssl/OpenSslTlsv13X509ExtendedTrustManager.java
```diff
@@ -0,0 +1,499 @@
...
+import java.net.Socket;
+import java.nio.ByteBuffer;
+import java.security.Principal;
+import java.security.cert.Certificate;
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
+import java.util.List;
+
+/**
+ * Provide a way to use {@code TLSv1.3} with Java versions prior to Java11 by adding a
```
trustin: Java11 -> 11?

handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
normanmaurer: nope as we use it also in the JDK impl etc. So lets keep it as it is

PR #8404, opened 2018-10-18
Fix broken testsuite-shading when using with netty-tcnative-boringssl…

normanmaurer:
…-static

Motivation:

2109f14c24f90df3f43aee7f3248ac59e6088735 corrected how we run the testsuite with boringssl-static but missed to also adjust the testsuite-shading configuration which lead to test failures.

Modifications:

Correctly compose the native lib name when no classifier is used.

Result:

Testsuite passes again.


normanmaurer:
Everyone loves xml ;)

PR #8409, opened 2018-10-19
Remove workaround in tests for TLSv1.3 bug in Java11 as it was fixed …

normanmaurer:
…in 11.0.1

Motivation:

We had put some workaround in our tests due a bug in the Java11 implementation of TLSv1.3. This was now fixes as part of 11.0.1.

See https://bugs.openjdk.java.net/browse/JDK-8211067.

Modifications:

Remove workaround in SSL tests.

Result:

Run all tests with supported TLS version.


normanmaurer:
Less code == good code :)

See https://bugs.openjdk.java.net/browse/JDK-8211067

PR #8410, opened 2018-10-19
(fix) incorrect reachability assumption in ResourceLeakDetector

almson:
As mentioned [here](https://github.com/netty/netty/commit/c2f4daa7398a9363fde8e51bf52c0d0323f870a5#commitcomment-30975329), `trackedObject != null` gives no guarantee that trackedObject remains reachable. It is an expression that can obviously be moved around by the compiler. The recommended solution is to use synchronized blocks to ensure reachability, because they're never optimized out and releasing the monitor requires the reference to remain reachable. This is discussed in https://docs.oracle.com/javase/9/docs/api/java/lang/ref/Reference.html#reachabilityFence-java.lang.Object- However, I am not an expert with intimate knowledge of the JDK. (If you have one, ask them!)

### DIFF
--- a/common/src/main/java/io/netty/util/ResourceLeakDetector.java
+++ b/common/src/main/java/io/netty/util/ResourceLeakDetector.java
@@ -461,11 +461,41 @@ public class ResourceLeakDetector<T> {
             // Ensure that the object that was tracked is the same as the one that was passed to close(...).
             assert trackedHash == System.identityHashCode(trackedObject);
 
-            // We need to actually do the null check of the trackedObject after we close the leak because otherwise
-            // we may get false-positives reported by the ResourceLeakDetector. This can happen as the JIT / GC may
-            // be able to figure out that we do not need the trackedObject anymore and so already enqueue it for
-            // collection before we actually get a chance to close the enclosing ResourceLeak.
-            return close() && trackedObject != null;
+            try {
+                return close();
+            } finally {
+                // This method will do `synchronized(trackedObject)` and we should be sure this will not cause deadlock.
+                // It should not, because somewhere up the callstack should be a (successful) `trackedObject.release`,
+                // therefore it is unreasonable that anyone else, anywhere, is holding a lock on the trackedObject.
+                // (Unreasonable but possible, unfortunately.)
+                reachabilityFence0(trackedObject);
+            }
+        }
+
+         /**
+         * Ensures that the object referenced by the given reference remains
+         * <a href="package-summary.html#reachability"><em>strongly reachable</em></a>,
+         * regardless of any prior actions of the program that might otherwise cause
+         * the object to become unreachable; thus, the referenced object is not
+         * reclaimable by garbage collection at least until after the invocation of
+         * this method.
+         *
+         * <p> Recent versions of the JDK have a nasty habit of prematurely deciding objects are unreachable.
+         * see: https://stackoverflow.com/questions/26642153/finalize-called-on-strongly-reachable-object-in-java-8
+         * The Java 9 method Reference.reachabilityFence offers a solution to this problem.
+         *
+         * <p> This method is always implemented as a synchronization on {@code ref}, not as
+         * {@code Reference.reachabilityFence} for consistency across platforms and to allow building on JDK 6-8.
+         * <b>It is the caller's responsibility to ensure that this synchronization will not cause deadlock.</b>
+         *
+         * @param ref the reference. If {@code null}, this method has no effect.
+         * @see java.lang.ref.Reference#reachabilityFence
+         */
+        private static void reachabilityFence0(Object ref) {
+            if (ref != null) {
+                // Empty synchronized is ok: https://stackoverflow.com/a/31933260/1151521
+                synchronized (ref) { }
+            }
         }
 
         @Override


common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -461,11 +461,19 @@ public boolean close(T trackedObject) {
...
+            // see: https://stackoverflow.com/questions/26642153/finalize-called-on-strongly-reachable-object-in-java-8
+            // The method Reference.reachabilityFence offers a solution to this problem.
+            // However, besides only being available in Java 9+,
+            // it "is designed for use in uncommon situations of premature finalization where using
+            // synchronized blocks or methods [is] not possible or do not provide the desired control."
+            // Because somewhere up the callstack should be a (successful) `trackedObject.release`,
+            // it is unreasonable that anyone else, anywhere, is hold a lock on the trackedObject.
+            // Therefore, it seems using a synchronization block is possible here, so we will use one!
+            synchronized (trackedObject)
+            {
```
normanmaurer: can you move the `{` up one line to match our coding style ?

ejona86: Would it be more clear if written this way?
```
boolean b = close();
synchronized (trackedObject) {}
return b;
```

I know that looks weird, but it makes it clear we aren't relying on the synchronization when running the close() method.

normanmaurer: @ejona86 I wonder if it could just optimise it away when it detects that the block is empty ?

normanmaurer:
@carl-mastrangelo @ejona86 @almson this may be an alternative fix for it: https://github.com/netty/netty/pull/8422

common/src/main/java/io/netty/util/ResourceLeakDetector.java
ejona86: Nope, because it still has side-effects. See my comment in https://github.com/grpc/grpc-java/pull/1526/files#r55582622

almson:
@ejona86 I agree that `try { close() } finally { synchronized (trackedObject) { } }` *should* work. Empty synchronized can't be optimized away because it sets up a happens-before relationship for the surrounding code (people use them: https://stackoverflow.com/a/31933260/1151521). I suppose we can even stick `synchronized (trackedObject) { }` into `PlatformDependent`.

I got as far as I could with the new commit. However, I do not know what I should do in PlatformDependent to reference the Java 9 method.

normanmaurer:
Use reflection to call it 👍

almson:
@carl-mastrangelo I'm not sure what you're asking. The bug is that the expression `trackedObject != null` is not a substitute for `Reference.reachabilityFence`. (it can be moved all the way to the top of the method.) The documentation for `reachabilityFence` says that `synchronized` is a substitute. A `synchronized` block cannot be moved because it creates happens-before semantics. A `synchronized` block keeps the object reachable because it uses its reference to call `releaseMonitor`. Optimizing out a synchronize is nearly impossible because the compiler would have to exclude any possibility that anyone else might synchronize on the object. (Actually, the GC can tell that nobody holds a reference to the object and therefore can't synchronize on it, but not the compiler.) In any case, this seems to be the recommended solution even in Java 9.

See: https://docs.oracle.com/javase/9/docs/api/java/lang/ref/Reference.html#reachabilityFence-java.lang.Object-

normanmaurer:
@almson good point... then lets just put the synchronised stuff in in the leak detector. no need to add something to `PlatformDependent`.

almson:
Well, it might be useful in other parts. Does any other code anywhere use `finalizer` or a weak/phantom reference? Does netty only do leak detection, or is there code somewhere that tries to clean up leaks?

normanmaurer:
@almson I think we only need it here.

almson:
I see WeakOrderQueue, HashedWheelTimer, PoolArena, and PoolThreadCache have finalizers. I also see other places where WeakReference and SoftReference are used. All of those places might require reachabilityFence. I haven't actually looked at them yet.

normanmaurer:
@almson from my understanding they need not... Let us just focus on the leakdetector in this pr and we can do another one if we really need more changes.

almson:
Well, the question was should we create PlatformDependent.reachabilityFence.

almson:
Ok, I'm going to save `reachabilityFence` together with its javadoc, but put it in `ResourceLeakDetector`.

almson:
From what I see so far:
- `PoolArena.finalize` is useless. Its only purpose is to hurry up the GC. But because it's a finalizer, all it does is delay it! For correctness, PoolArena would need a reachabilityFence in every method. I'm guessing it hasn't yet caused problems because it's so long lived it only becomes unreachable at shutdown.
- `Recycler.WeakOrderQueue.finalize` also seems useless. It's supposed to be a GC optimization--but, again, a finalizer delays GC. If clearing the references really is a good optimization, then it will be even better if invoked directly not through finalization. It's not clear how dangerous this finalizer is.
- `HashedWheelTimer.finalize` seems to only be part of a logging feature, so is harmless.
- `PoolThreadCache.finalize` ok if `free` is called (but then, it doesn't do anything). Will suffer from premature finalization if `free` is not called. It should be hooked up to the leak detector instead.
- There is an ObjectCleaner, but I don't see it being used anywhere. If it is, it's going to cause problems.

The other uses of weak references don't use ReferenceQueues, other than the leak detector.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -461,11 +461,42 @@ public boolean close(T trackedObject) {
...
-            return close() && trackedObject != null;
+            try {
+                return close();
+            }
+            finally {
+                // This method will do `synchronized (trackedObject)` and we should be sure this will not cause deadlock.
+                // It should not, because somewhere up the callstack should be a (successful) `trackedObject.release`,
+                // therefore it is unreasonable that anyone else, anywhere, is holding a lock on the trackedObject.
+                // (Unreasonable but possible, unfortunately.)
+                reachabilityFence (trackedObject);
```
normanmaurer: nit: remove space before (

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -461,11 +461,42 @@ public boolean close(T trackedObject) {
...
+         * 
+         * <p> This method is always implemented as a synchronization on {@code ref}, not as {@code Reference.reachabilityFence}
+         * for consistency across platform version and to allow building on JDK 6-8.
+         * <b>It is the caller's responsibility to ensure that this synchronization will not cause deadlock.</b>
+         * 
+         * @param ref the reference. If {@code null}, this method has no effect.
+         * @see java.lang.ref.Reference#reachabilityFence
+         */
+        private static void reachabilityFence (Object ref) {
+            if (ref != null)
```
normanmaurer: Add { } to match our coding style

normanmaurer:
@almson let’s focus on the one case here and then we can open a new issue and discuss about the others there. I want to keep the focus on this change

normanmaurer:
@almson for now not... if we need it at the end we can easily do later on.

almson:
@normanmaurer I don't understand. Now not what?

Should I squash the commits?

normanmaurer:
@almson ignore my comment... I may have read an old comment and tried to answer 🤣 

Just squash and thanks again for your help!

normanmaurer:
@netty-bot test this please

almson:
Np! This was my first github pull request, btw.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -461,11 +461,43 @@ public boolean close(T trackedObject) {
...
 
-            // We need to actually do the null check of the trackedObject after we close the leak because otherwise
-            // we may get false-positives reported by the ResourceLeakDetector. This can happen as the JIT / GC may
-            // be able to figure out that we do not need the trackedObject anymore and so already enqueue it for
-            // collection before we actually get a chance to close the enclosing ResourceLeak.
-            return close() && trackedObject != null;
+            try {
+                return close();
+            }
+            finally {
```
normanmaurer: sorry missed this... please merge with previous line..

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -461,11 +461,43 @@ public boolean close(T trackedObject) {
...
+                // This method will do `synchronized (trackedObject)` and we should be sure this will not cause deadlock.
+                // It should not, because somewhere up the callstack should be a (successful) `trackedObject.release`,
+                // therefore it is unreasonable that anyone else, anywhere, is holding a lock on the trackedObject.
+                // (Unreasonable but possible, unfortunately.)
+                reachabilityFence(trackedObject);
+            }
+        }
+        
+
```
normanmaurer: please remove one empty line

normanmaurer:
@almson please fix all the check style errors:

```
11:57:35 [INFO] Starting audit...
11:57:35 /code/common/src/main/java/io/netty/util/ResourceLeakDetector.java:466:13: '}' should be on the same line.
11:57:35 /code/common/src/main/java/io/netty/util/ResourceLeakDetector.java:468: Line is longer than 120 characters (found 121).
11:57:35 /code/common/src/main/java/io/netty/util/ResourceLeakDetector.java:475: trailing whitespace
11:57:35 /code/common/src/main/java/io/netty/util/ResourceLeakDetector.java:476: trailing whitespace
11:57:35 /code/common/src/main/java/io/netty/util/ResourceLeakDetector.java:484: trailing whitespace
11:57:35 /code/common/src/main/java/io/netty/util/ResourceLeakDetector.java:488: trailing whitespace
11:57:35 /code/common/src/main/java/io/netty/util/ResourceLeakDetector.java:489: Line is longer than 120 characters (found 128).
11:57:35 /code/common/src/main/java/io/netty/util/ResourceLeakDetector.java:492: trailing whitespace
11:57:35 /code/common/src/main/java/io/netty/util/ResourceLeakDetector.java:496:47: '(' is preceded with whitespace.
11:57:35 /code/common/src/main/java/io/netty/util/ResourceLeakDetector.java:499:37: '{' is not followed by whitespace.
11:57:35 /code/common/src/main/java/io/netty/util/ResourceLeakDetector.java:499:37: '}' is not preceded with whitespace.
11:57:35 Audit done.
```

Just run `./mvnw clean package` to see if there are any left.

almson:
Updated. Thanks for your patience.

normanmaurer:
@netty-bot test this please

normanmaurer:
@almson you will need to change the name of the method to be able to compile on java9 and later. See the error message on the ci 

https://ci.netty.io/job/netty-centos6-java11-prb/152/console

almson:
Wow, ironic. Should I call it `reachabilityFence0`? That seems to be popular.

normanmaurer:
Yeah why not 😂

normanmaurer:
@netty-bot test this please

normanmaurer:
@almson this looks good now... One last thing can you please amend your commit message to match our commit template and force push ?
After this is done I will pull in:

https://netty.io/wiki/writing-a-commit-message.html

almson:
done

normanmaurer:
@almson thanks a lot for all your work here!

PR #8413, opened 2018-10-20
Fix minor bug in DefaultResourceLeak.toString()

almson:
As you can see, we were printing `dropped` instead of `duped`. Looks like nobody ever noticed.


normanmaurer:
Can you please sign the icla and let me know once done ?

https://Netty.io/s/icla

normanmaurer:
@netty-bot test this please

almson:
done

normanmaurer:
@almson thanks!

PR #8414, opened 2018-10-20
Inside ResourceLeakDetector, use linked list instead of ConcurrentHas…

almson:
This pattern is taken from java.lang.ref.Cleaner (or rather, jdk.internal.ref.CleanerImpl). Should be a lot more efficient than a HashSet.

Inspiration:
[jdk.internal.ref.CleanerImpl](https://hg.openjdk.java.net/jdk/jdk11/file/1ddf9a99e4ad/src/java.base/share/classes/jdk/internal/ref/CleanerImpl.java)
[jdk.internal.ref.PhantomCleanable](https://hg.openjdk.java.net/jdk/jdk11/file/1ddf9a99e4ad/src/java.base/share/classes/jdk/internal/ref/PhantomCleanable.java)

### DIFF
--- a/common/src/main/java/io/netty/util/ResourceLeakDetector.java
+++ b/common/src/main/java/io/netty/util/ResourceLeakDetector.java
@@ -159,8 +159,8 @@ public class ResourceLeakDetector<T> {
     }
 
     /** the collection of active resources */
-    private final ConcurrentMap<DefaultResourceLeak<?>, LeakEntry> allLeaks = PlatformDependent.newConcurrentHashMap();
-
+    private final DefaultResourceLeak<Object> refListHead = new DefaultResourceLeak<Object>();
+    
     private final ReferenceQueue<Object> refQueue = new ReferenceQueue<Object>();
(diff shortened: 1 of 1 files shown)


almson:
FYI, I've pulled out and refactored the reference counting bits into a stand-alone library: https://github.com/almson/almson-refcount The main feature is there's an actual base class that all reference-counted objects can use. It handles everything. A lot of chairs were rearranged, this linked list implementation was added, and leak detector presets were changed. I deleted a lot of the flexibility around leak detector factories and interfaces and stuff, because I don't see the value. I'd love to see the same refactoring being applied to Netty, but I realize it's a tall ask.

normanmaurer:
@almson why would it be a lot more efficient then a HashSet ?

johnou:
@normanmaurer I would close in favour of https://github.com/netty/netty/pull/8467 .. https://github.com/netty/netty/pull/8414/files#diff-e7f59893bb7f88824527d638b6cbba72R495 anyway stinks to me, looks like it should be iterating over the double linked list but it's just popping?

normanmaurer:
Yep closing

PR #8415, opened 2018-10-20
Correctly detect if KeyManagerFactory is supported by OpenSSL even wh…

normanmaurer:
…en sun.security.x509.* can not be accessed and bouncycastle is not on the classpath.

Motivation:

OpenSsl used SelfSignedCertificate in its static init block to detect if KeyManagerFactory is supported. Unfortunally this only works when either sun.security.x509.* can be accessed or bouncycastle is on the classpath.
We should not depend on either of it.

This came up in https://github.com/netty/netty-tcnative/issues/404#issuecomment-431551890.

Modifications:

Just directly use the bytes to generate the X509Certificate and so not depend on sun.security.x509.* / bouncycastle.

Result:

Correctly be able to detect if KeyManagerFactory can be supported in all cases.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
+++ b/handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
@@ -18,7 +18,6 @@ package io.netty.handler.ssl;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
-import io.netty.handler.ssl.util.SelfSignedCertificate;
 import io.netty.internal.tcnative.Buffer;
 import io.netty.internal.tcnative.Library;
 import io.netty.internal.tcnative.SSL;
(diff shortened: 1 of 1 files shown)


normanmaurer:
@ejona86 @carl-mastrangelo @trustin let me know if I should use a different hostname for the self-signed cert bytes. That said it really not matters at all.

handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
```diff
@@ -172,8 +173,40 @@
                             logger.debug("Hostname Verification not supported.");
                         }
                         try {
-                            cert = new SelfSignedCertificate();
-                            certBio = ReferenceCountedOpenSslContext.toBIO(ByteBufAllocator.DEFAULT, cert.cert());
+                            // Bytes of Self-Signed certificate for netty.io
+                            byte[] certBytes =
+                                    { 48, -126, 1, -92, 48, -126, 1, 13, -96, 3, 2, 1, 2, 2, 9, 0, -9, 61,
```
trustin: nit: Curly brace should be at the end of the previous line.

handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
```diff
@@ -172,8 +173,40 @@
                             logger.debug("Hostname Verification not supported.");
                         }
                         try {
-                            cert = new SelfSignedCertificate();
-                            certBio = ReferenceCountedOpenSslContext.toBIO(ByteBufAllocator.DEFAULT, cert.cert());
+                            // Bytes of Self-Signed certificate for netty.io
```
trustin: nit: Self-Signed -> self-signed

normanmaurer: done

PR #8420, opened 2018-10-22
Avoid allocations when wrapping byte[] and ByteBuffer arrays as ByteBuf

njhill:
Motivation:

`Unpooled.wrap(byte[]...)` and `Unpooled.wrap(ByteBuffer...)` currently allocate/copy an intermediate `ByteBuf` `ArrayList` and array, which can be avoided.

Modifications:

- Define new internal `ByteWrapper` interface and add a `CompositeByteBuf` constructor which takes a `ByteWrapper` with an array of the type that it wraps; modify the appropriate `Unpooled.wrap(...)` methods to take advantage of it
- Tidy up other constructors in `CompositeByteBuf` to remove duplication and misleading `len` arg (which is really an end offset into provided array)

Result:

Less allocation/copying when wrapping `byte[]` and `ByteBuffer` arrays, tidier code.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
@@ -53,61 +53,81 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
 
     private boolean freed;
 
-    public CompositeByteBuf(ByteBufAllocator alloc, boolean direct, int maxNumComponents) {
+    private CompositeByteBuf(ByteBufAllocator alloc, boolean direct, int maxNumComponents, int initSize) {
         super(AbstractByteBufAllocator.DEFAULT_MAX_CAPACITY);
         if (alloc == null) {
             throw new NullPointerException("alloc");
         }
+        if (maxNumComponents < 2) {
+            throw new IllegalArgumentException(
+                    "maxNumComponents: " + maxNumComponents + " (expected: >= 2)");
+        }
         this.alloc = alloc;
         this.direct = direct;
         this.maxNumComponents = maxNumComponents;
-        components = newList(0, maxNumComponents);
+        components = newList(initSize, maxNumComponents);
+    }
+
+    public CompositeByteBuf(ByteBufAllocator alloc, boolean direct, int maxNumComponents) {
+        this(alloc, direct, maxNumComponents, 0);
     }
 
     public CompositeByteBuf(ByteBufAllocator alloc, boolean direct, int maxNumComponents, ByteBuf... buffers) {
         this(alloc, direct, maxNumComponents, buffers, 0, buffers.length);
     }
 
-    CompositeByteBuf(
-            ByteBufAllocator alloc, boolean direct, int maxNumComponents, ByteBuf[] buffers, int offset, int len) {
-        super(AbstractByteBufAllocator.DEFAULT_MAX_CAPACITY);
-        if (alloc == null) {
-            throw new NullPointerException("alloc");
-        }
-        if (maxNumComponents < 2) {
-            throw new IllegalArgumentException(
-                    "maxNumComponents: " + maxNumComponents + " (expected: >= 2)");
-        }
+    CompositeByteBuf(ByteBufAllocator alloc, boolean direct, int maxNumComponents,
+            ByteBuf[] buffers, int offset, int endOffset) {
+        this(alloc, direct, maxNumComponents, endOffset - offset);
 
-        this.alloc = alloc;
-        this.direct = direct;
-        this.maxNumComponents = maxNumComponents;
-        components = newList(len, maxNumComponents);
-
-        addComponents0(false, 0, buffers, offset, len);
+        addComponents0(false, 0, buffers, offset, endOffset);
         consolidateIfNeeded();
         setIndex(0, capacity());
     }
 
     public CompositeByteBuf(
             ByteBufAllocator alloc, boolean direct, int maxNumComponents, Iterable<ByteBuf> buffers) {
-        super(AbstractByteBufAllocator.DEFAULT_MAX_CAPACITY);
-        if (alloc == null) {
-            throw new NullPointerException("alloc");
-        }
-        if (maxNumComponents < 2) {
-            throw new IllegalArgumentException(
-                    "maxNumComponents: " + maxNumComponents + " (expected: >= 2)");
-        }
-
-        int len = buffers instanceof Collection ? ((Collection<ByteBuf>) buffers).size() : 0;
-
-        this.alloc = alloc;
-        this.direct = direct;
-        this.maxNumComponents = maxNumComponents;
-        components = newList(len, maxNumComponents);
+        this(alloc, direct, maxNumComponents,
+                buffers instanceof Collection ? ((Collection<ByteBuf>) buffers).size() : 0);
 
         addComponents0(false, 0, buffers);
+        consolidateIfNeeded();
+        setIndex(0, capacity());
+    }
+
+    // support passing arrays of other types instead of having to copy to a ByteBuf[] first
+    interface ByteWrapper<T> {
+        ByteBuf wrap(T bytes);
+        boolean isEmpty(T bytes);
+    }
+
+    static final ByteWrapper<byte[]> BYTE_ARRAY_WRAPPER = new ByteWrapper<byte[]>() {
+        @Override
+        public ByteBuf wrap(byte[] bytes) {
+            return Unpooled.wrappedBuffer(bytes);
+        }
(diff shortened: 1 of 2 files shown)


normanmaurer:
@netty-bot test this please

buffer/src/main/java/io/netty/buffer/Unpooled.java
```diff
@@ -262,38 +262,86 @@ public static ByteBuf wrappedBuffer(ByteBuffer... buffers) {
...
+        }
+
+        int maxSize() {
+            return array.length - start;
+        }
+
+        @Override
+        public Iterator<ByteBuf> iterator() {
+            return new Iterator<ByteBuf>() {
+                int i = start;
```
normanmaurer: nit: private

buffer/src/main/java/io/netty/buffer/Unpooled.java
```diff
@@ -262,38 +262,86 @@ public static ByteBuf wrappedBuffer(ByteBuffer... buffers) {
...
+                    return;
+                }
+                if (!isEmpty(array[i])) {
+                    start = i;
+                    return;
+                }
+            }
+        }
+
+        int maxSize() {
```
normanmaurer: final

buffer/src/main/java/io/netty/buffer/Unpooled.java
```diff
@@ -262,38 +262,86 @@ public static ByteBuf wrappedBuffer(ByteBuffer... buffers) {
         return wrappedBuffer(buffers.length, buffers);
     }
 
+    // used to avoid temporary array allocation/copy when wrapping byte[][] and ByteBuffer[]
+    abstract static class WrappingIterable<T> implements Iterable<ByteBuf> {
+        private final T[] array;
+        private final int start;
+
+        protected WrappingIterable(T[] array) {
+            this.array = array;
```
normanmaurer: Is it really safe to not call `clone()` ?

buffer/src/main/java/io/netty/buffer/Unpooled.java
normanmaurer: nit: private

buffer/src/main/java/io/netty/buffer/Unpooled.java
```diff
@@ -262,38 +262,86 @@ public static ByteBuf wrappedBuffer(ByteBuffer... buffers) {
...
+                        }
+                    }
+                    return false;
+                }
+
+                @Override
+                public ByteBuf next() {
+                    if (hasNext()) {
+                        return wrap(array[i++]);
+                    }
```
normanmaurer: You could make this a bit more efficient by merge some code of `hasNext` in here you will need to do two array access here.

buffer/src/main/java/io/netty/buffer/Unpooled.java
```diff
@@ -262,38 +262,86 @@ public static ByteBuf wrappedBuffer(ByteBuffer... buffers) {
...
+        }
+
+        @Override
+        public Iterator<ByteBuf> iterator() {
+            return new Iterator<ByteBuf>() {
+                int i = start;
+
+                @Override
+                public boolean hasNext() {
+                    for (; i < array.length && array[i] != null; i++) {
```
normanmaurer: You should assign `array[I]` to something and re-use it, as at the moment you access each element two times.

buffer/src/main/java/io/netty/buffer/Unpooled.java
```diff
@@ -337,31 +385,27 @@ public static ByteBuf wrappedBuffer(int maxNumComponents, ByteBuf... buffers) {
...
+                ByteBuf wrap(ByteBuffer b) {
+                    return wrappedBuffer(b.order(BIG_ENDIAN));
                 }
-            }
-
-            if (!components.isEmpty()) {
-                return new CompositeByteBuf(ALLOC, false, maxNumComponents, components);
-            }
+            };
+            return iterable.maxSize() <= 0 ? EMPTY_BUFFER
```
normanmaurer: == 0 ?

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
```diff
@@ -367,31 +371,35 @@ private int addComponents0(boolean increaseIndex, int cIndex, Iterable<ByteBuf>
...
-                        }
+            }
+            return cIndex;
+        } finally {
+            while (it.hasNext()) {
+                ByteBuf b = it.next();
+                if (b != null) {
+                    try {
+                        b.release();
+                    } catch (Throwable ignored) {
```
normanmaurer: just use `ReferenceCountedUtil.safeRelease`

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
```diff
@@ -101,6 +102,9 @@ public CompositeByteBuf(
         }
 
         int len = buffers instanceof Collection ? ((Collection<ByteBuf>) buffers).size() : 0;
+        if (buffers instanceof Unpooled.WrappingIterable) {
+            len = ((Unpooled.WrappingIterable<?>) buffers).maxSize();
+        }
```
normanmaurer: Should we merge the two instanceof things ?

```
if (buffers instanceof Unpooled....) {
   len = ...
} else if (buffers instanceof Collection) {
    len = ...;
} else {
    len = 0;
}
```

buffer/src/main/java/io/netty/buffer/Unpooled.java
njhill: @normanmaurer I think so - `CompositeByteBuf` doesn't retain a reference to any `ByteBuf` array or `Iterable` passed to its constructors, it just extracts the contained `ByteBuf`s within the context of the ctor. Unless I am missing something (very possible!)

I've now addressed all the other comments.

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
normanmaurer: @njhill did you miss this one ?

buffer/src/main/java/io/netty/buffer/Unpooled.java
normanmaurer: @njhill ...

buffer/src/main/java/io/netty/buffer/Unpooled.java
normanmaurer: I think if this is the assumption (did not check this yet) we should add a comment. Also I think we should maybe better the Iterable into the CompositeByteBuf then as it has a clear contract with it.

buffer/src/main/java/io/netty/buffer/Unpooled.java
njhill: @normanmaurer i think this may be a github quirk - it didn't update the PR front page with the change, but I did make it (actually refactored the iterator a bit more to address this) - check the changes tab/commits

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
njhill: @normanmaurer no I didn't miss it, check the changes tab/commits ... I think it's a github quirk making it show the old version here still

buffer/src/main/java/io/netty/buffer/Unpooled.java
njhill: Agree with both... and actually I think maybe simpler would just be a generic package-private constructor for CBB which takes a `T[]` where `T extends ByteHolder` where the byte holder is internal interface with the wrap methods, will update this PR when I get a chance.

njhill:
@normanmaurer I've made some simplifications mostly based on the prior discussion:
- Replace `WrappingIterable` with a simpler approach which uses a new `ByteWrapper` interface defined in `CompositeByteBuf` rather than `Unpooled`
- Simplify `CompositeByteBuf` constructor duplication, rename misleadingly-named `len` constructor args to `endOffset`
- Revert some of the other changes to `CompositeByteBuf` which are now included in #8437

Should I just also merge the rest into #8437 and close this one?

normanmaurer:
@njhill I think its useful to keep this separated... The other PR is already huge. Let me review this one first.

normanmaurer:
@netty-bot test this please

normanmaurer:
@njhill can you just squash and adjust commit message. After this I will merge...

thanks a lot!

njhill:
@normanmaurer done, thanks! and then I will rebase the other one

normanmaurer:
@njhill merged 👍

carl-mastrangelo:
LGTM

PR #8421, opened 2018-10-22
Update to JDK 12 EA16 when running CI jobs against JDK 12.

normanmaurer:
Motivation:

A new EA release was done, we should always run against the latest.

Modifications:

Update to EA 16.

Result:

CI runs with latest EA release for JDK12.


normanmaurer:
@netty-bot test this please

PR #8423, opened 2018-10-23
Add testing ByteBufAllocator for leak unit testing

dain:
Motivation:

Currently the only way to detect leaks during testing is log scanning which is
difficult to automate.  As an alternative it would be nice to fail unit tests
that leak buffers.

Modifications:

Added try with resources friendly TestingPooledByteBufAllocator which throws
an AssertionError in close if all buffers have not been freed.  To implement
this, I modified modified ResourceLeakDetector to allow sub clases to force
tracking and error reporting regardless of the global static tracking level,
or the configuration of the logging system.  I also added a method to allow
sub classes to dispose of all tracked references, which triggers reporting
if the referant was not freed. Then I modified, AbstractByteBufAllocator and
PooledByteBufAllocator to allow package local sub classes to hook the call to
the global static toLeakAwareBuffer methods.  Finally, I added
TestingPooledByteBufAllocator which is a subclass of PooledByteBufAllocator
containing a private ResourceLeakDetector that always tracks and warns on
leaks.

Result:

Developers using Netty can now easily verify that all buffers have been
released in unit tests.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/TestingPooledByteBufAllocator.java
+++ b/buffer/src/main/java/io/netty/buffer/TestingPooledByteBufAllocator.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.netty.buffer;
+
+import io.netty.util.ResourceLeakDetector;
+import io.netty.util.ResourceLeakTracker;
+
+import java.io.Closeable;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import static io.netty.util.internal.StringUtil.NEWLINE;
+
+/**
+ * A {@link ByteBufAllocator} for unit testing.  This allocator performs leak tracking
+ * on all allocated buffers.  At the end the test, call {@link #close} to assert
+ * that all buffers have been freed.
+ */
+public class TestingPooledByteBufAllocator extends PooledByteBufAllocator implements Closeable {
+
+    private final TestingResourceLeakDetector leakDetector = new TestingResourceLeakDetector();
+
+    public TestingPooledByteBufAllocator() {
+    }
+
+    public TestingPooledByteBufAllocator(boolean preferDirect) {
+        super(preferDirect);
+    }
+
+    @Override
+    ByteBuf toLeakAwareBufferInternal(ByteBuf buf) {
+        ResourceLeakTracker<ByteBuf> leak = leakDetector.track(buf);
+        if (leak != null) {
+            buf = new AdvancedLeakAwareByteBuf(buf, leak);
+        }
+        return buf;
+    }
+
+    @Override
+    CompositeByteBuf toLeakAwareBufferInternal(CompositeByteBuf buf) {
+        ResourceLeakTracker<ByteBuf> leak = leakDetector.track(buf);
+        if (leak != null) {
+            buf = new AdvancedLeakAwareCompositeByteBuf(buf, leak);
+        }
+        return buf;
+    }
+
+    /**
+     * Throws an exception if all allocated buffers have been not freed.
+     * @throws AssertionError if buffers have been leaked.
+     */
+    @Override
+    public void close() {
+        List<String> allLeaks = leakDetector.getAllLeaks();
+        if (!allLeaks.isEmpty()) {
+            throwLeakDetected(allLeaks);
+        }
+    }
+
+    private static void throwLeakDetected(List<String> allLeaks) {
+        int expectedBufferSize = 1024;
+        for (String leak : allLeaks) {
+            expectedBufferSize += leak.length();
+        }
+
+        StringBuilder buf = new StringBuilder(expectedBufferSize)
+                .append("LEAK: ")
+                .append(allLeaks.size())
+                .append(" leak(s) detected. ")
+                .append("ByteBuf.release() was not called before it's garbage-collected. ")
+                .append("See http://netty.io/wiki/reference-counted-objects.html for more information.");
+        for (int i = 0; i < allLeaks.size(); i++) {
+            String leak = allLeaks.get(i);
+            buf.append(NEWLINE)
+                    .append("== ByteBuf ")
+                    .append(i + 1)
+                    .append(": ")
+                    .append(leak);
+        }
+        buf.append(NEWLINE)
+                .append("== Leak detection:");
+        throw new AssertionError(buf.toString());
+    }
+
+    private static class TestingResourceLeakDetector extends ResourceLeakDetector<ByteBuf> {
+
+        private final List<String> leaks = new ArrayList<String>(0);
+
+        public TestingResourceLeakDetector() {
+            super(ByteBuf.class);
+        }
+
+        @Override
+        protected boolean shouldTrack(ByteBuf obj) {
+            return true;
+        }
+
+        @Override
+        protected boolean isLeakReportingEnabled() {
+            return true;
+        }
+
+        @Override
+        protected synchronized void reportTracedLeak(String resourceType, String records) {
+            leaks.add(records);
+        }
+
+        @Override
+        protected synchronized void reportUntracedLeak(String resourceType) {
+            leaks.add("Untracked ByteBuf without created at information");
+        }
+
+        synchronized List<String> getAllLeaks() {
+            disposeAllReferences();
+            return Collections.unmodifiableList(new ArrayList<String>(leaks));
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


dain:
BTW, I attempted to make the smallest addition to public APIs in this change.  They are limited to the new `protected` methods on `ResourceLeakDetector`, which are required because it is in a different package from the buffer allocators, and the new TestingPooledByteBufAllocator.  As an alternative to this implementation, you might may want to add a constructor to the allocators that allow the `ResourceLeakDetector` to be set to an instance level instead of requiring, and the "hook" methods I added.  Additionally, you may want to add a new constructor to `ResourceLeakDetector` that force enables resource tracking.

buffer/src/main/java/io/netty/buffer/TestingPooledByteBufAllocator.java
```diff
@@ -0,0 +1,105 @@
...
+
+import java.io.Closeable;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * A {@link ByteBufAllocator} for unit testing.  This allocator performs leak tracking
+ * on all allocated buffers.  At the end the test, call {@link #close} to assert
+ * that all buffers have been freed.
+ */
+public class TestingPooledByteBufAllocator extends PooledByteBufAllocator implements Closeable {
```
normanmaurer: I wonder if this should better go into src/main/test and we should publish a test-jar to depend on ? 

@trustin @ejona86 @carl-mastrangelo  thoughts ?

buffer/src/main/java/io/netty/buffer/TestingPooledByteBufAllocator.java
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        @Override
+        protected boolean isLeakReportingEnabled() {
+            return true;
+        }
+
+        @Override
+        protected void reportTracedLeak(String resourceType, String records) {
+            leaked.set(true);
```
normanmaurer: I wonder if we should better store the resourceType and records here as well to make it easier to debug ?

We could include these in the AssertationError.

buffer/src/main/java/io/netty/buffer/TestingPooledByteBufAllocator.java
```diff
@@ -0,0 +1,105 @@
...
+            super.reportTracedLeak(resourceType, records);
+        }
+
+        @Override
+        protected void reportUntracedLeak(String resourceType) {
+            leaked.set(true);
+            super.reportUntracedLeak(resourceType);
```
normanmaurer: same as above

normanmaurer:
@netty-bot test this please

buffer/src/main/java/io/netty/buffer/TestingPooledByteBufAllocator.java
ejona86: gRPC uses a separate artifact (grpc-testing), which has been working well. But we also are able to make interesting dependency ordering, like grpc-testing depends on grpc-core, and grpc-core's tests depend on grpc-testing. That doesn't seem like a problem for this one class though.

Using a separate classifier in netty-buffer seems fine in many ways, although it probably won't show up in javadocs which would be unfortunate. With a separate testing artifact, you have to chose how many dependencies you're willing to have, so maybe the classifier is lighter-weight since it's not a big deal to have a separate jar for each existing artifact.

dain: I don't think putting this in test-jar would work since you end up pulling in testing deps like Junit.  We could move this to a new module, but we would need to switch the package due to the Java 9 split-module rule, and this would require making the new methods on `AbstractByteBufAllocator` protected (instead of package protected).

buffer/src/main/java/io/netty/buffer/TestingPooledByteBufAllocator.java
dain: I can do that.  Do you want a new public exception class so you can pull out the records programmatically, or private one so you need a debugger?  I use a debugger when working on stuff like this, so I don't have a strong opinion.

dain:
I updated the code to not call the super `reportTracedLeak` method as this logs a `LEAK` message which causes the build to fail.  Instead the new version stores the leak `records` and then throws an `AssertionError` containing all of the leaks in a single message, and the tests simply verify the message without printing.  Here is an example of the output you will get if the `AssertionError` is not caught https://gist.github.com/dain/5fece0946d7a1439b535a7af775f5a57

almson:
Retrieving log messages in a test is a pain, and this is a great idea, but

- I don't like how it's specific to ByteBufAllocator. Leak detector is used in other places. Seems like it should be a feature of the leak detector.
- There is a `ResourceLeakDetector.setLevel` which should obviate the changes to AbstractByteBufAllocator.
- The feature of the leak detector can be boiled down to having it write to a stringbuilder instead of a log, and offer a public method to reportLeaks as an assertion error. Asserting the sb is empty is the same as asserting on a list of strings.

dain:
@almson the approach you are suggestion has a few problems for my testing environment:
* My tests are executed concurrently to reduce build latencies, and the existing resource detection system is a global static, so leaks from one test can pollute results from another.
* The existing leak detector requires that an buffer be GCed, before a leak can be detected.  Since there can be a leak in a hard reference to a buffer, the object may never be GCed.  This change checks that all tracked objects are collected at the end of the test.

almson:
@dain 
- How do you execute your tests concurrently? Might not `forkMode` resolve the issue of statics? There's plenty of other statics all over the place.
- You're right that reportLeaks looks in refQueue while your disposeAllReferences looks in allLeaks. That's fine. My point is that such a method can be added directly to ResourceLeakDetector (named, "assertAllResourcesReleased" or somesuch) and properly documented without making the class any worse for it. This way, someone else stands a chance of discovering this functionality. It might even be useful in production (asserts often are.)

Btw, I'm not a core contributor, so don't take my criticism too seriously, but I did muck about in ResourceLeakDetector quite a bit just recently.

dain:
> * How do you execute your tests concurrently? Might not `forkMode` resolve the issue of statics? There's plenty of other statics all over the place.

We use multithreaded Maven tests and use the TestNG framework.  In my experience, `forkMode` option isn't great for fine grained parallelism (each test method could be creating an allocator), and not something the project uses.

> * You're right that reportLeaks looks in refQueue while your disposeAllReferences looks in allLeaks. That's fine. My point is that such a method can be added directly to ResourceLeakDetector (named, "assertAllResourcesReleased" or somesuch) and properly documented without making the class any worse for it. This way, someone else stands a chance of discovering this functionality.

I agree.  In a comment above, I mention that I tried to limit the additions to the public interfaces of this project, but in the long run, it might be better to explicitly support this without needing to subclass the allocator and resource tracker.

dain:
@normanmaurer updated.  Can you take another look?

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -275,7 +281,7 @@ private void clearRefQueue() {
     }
 
     private void reportLeak() {
-        if (!logger.isErrorEnabled()) {
+        if (isLeakReportingEnabled()) {
```
almson: This should be if (! isLeakReportingEnabled()).

almson:
My alternate proposal: https://github.com/netty/netty/pull/8431

buffer/src/main/java/io/netty/buffer/TestingPooledByteBufAllocator.java
carl-mastrangelo: +1 to a test jar.  There are some  testing utility functions / class I would like to add to Netty, but don't have anywhere to put them.

buffer/src/main/java/io/netty/buffer/TestingPooledByteBufAllocator.java
```diff
@@ -0,0 +1,133 @@
...
+    ByteBuf toLeakAwareBufferInternal(ByteBuf buf) {
+        ResourceLeakTracker<ByteBuf> leak = leakDetector.track(buf);
+        if (leak != null) {
+            buf = new AdvancedLeakAwareByteBuf(buf, leak);
+        }
+        return buf;
+    }
+
+    @Override
+    CompositeByteBuf toLeakAwareBufferInternal(CompositeByteBuf buf) {
```
carl-mastrangelo: Overloading with sub/super is probably not going to work out well.   Consider:

```java
ByteBuf composite = new CompositeByteBuffer(...);
toLeakAwareBufferInternal(composite);
```

That will call the wrong tracker.

buffer/src/main/java/io/netty/buffer/TestingPooledByteBufAllocator.java
```diff
@@ -0,0 +1,133 @@
...
+            return true;
+        }
+
+        @Override
+        protected boolean isLeakReportingEnabled() {
+            return true;
+        }
+
+        @Override
+        protected synchronized void reportTracedLeak(String resourceType, String records) {
```
carl-mastrangelo: Just use ConcurrentLinkedQueue?

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -245,18 +245,24 @@ public final ResourceLeak open(T obj) {
...
-    private DefaultResourceLeak track0(T obj) {
+    protected boolean shouldTrack(T obj) {
         Level level = ResourceLeakDetector.level;
         if (level == Level.DISABLED) {
-            return null;
+            return false;
         }
 
         if (level.ordinal() < Level.PARANOID.ordinal()) {
             if ((PlatformDependent.threadLocalRandom().nextInt(samplingInterval)) == 0) {
```
carl-mastrangelo: This should just be a return right?

`return (PlatformDependent.threadLocalRandom().nextInt(samplingInterval)) == 0);`

normanmaurer:
Let me close this ... I think a better approach would be to do something with a `RunListener`: https://junit.org/junit4/javadoc/4.12/org/junit/runner/notification/RunListener.html

PR #8428, opened 2018-10-24
Introduce SingleThreadEventLoop.registeredChannels

### LINKED ISSUE #8276: Feature request: Exposing Number of Channels per Eventloop
Netty doesn't expose the number of channels per event loop for Kqueue|Epoll|Nio. It might be helpful for monitoring performances to expose this. 
Could also be useful for a better way of assigning channels to event loops as well. (My results on this were inconclusive, but the simulations I wrote might not be good).

vkostyukov:
Motivation:

Systems depending on Netty may benefit (telemetry, alternative even loop scheduling algorithms) from knowing the number of channels assigned to each `EventLoop`.

Modification:

Expose the number of channels registered in the `EventLoop` via `SingleThreadEventLoop.registeredChannels`.

Result:

Fixes #8276.

### DIFF
--- a/testsuite/src/main/java/io/netty/testsuite/transport/AbstractSingleThreadEventLoopTest.java
+++ b/testsuite/src/main/java/io/netty/testsuite/transport/AbstractSingleThreadEventLoopTest.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2019 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.testsuite.transport;
+
+import org.junit.Test;
+
+import io.netty.channel.Channel;
+import io.netty.channel.EventLoopGroup;
+import io.netty.channel.SingleThreadEventLoop;
+import io.netty.channel.socket.ServerSocketChannel;
+
+import static org.junit.Assert.*;
+
+public abstract class AbstractSingleThreadEventLoopTest {
+
+    @Test
+    public void testChannelsRegistered()  {
+        EventLoopGroup group = newEventLoopGroup();
+        final SingleThreadEventLoop loop = (SingleThreadEventLoop) group.next();
+
+        try {
+            final Channel ch1 = newChannel();
+            final Channel ch2 = newChannel();
+
+            assertEquals(0, loop.registeredChannels());
+
+            assertTrue(loop.register(ch1).syncUninterruptibly().isSuccess());
+            assertTrue(loop.register(ch2).syncUninterruptibly().isSuccess());
+            assertEquals(2, loop.registeredChannels());
+
+            assertTrue(ch1.deregister().syncUninterruptibly().isSuccess());
+            assertEquals(1, loop.registeredChannels());
+        } finally {
+            group.shutdownGracefully();
+        }
+    }
+
+    protected abstract EventLoopGroup newEventLoopGroup();
+    protected abstract ServerSocketChannel newChannel();
+}
--- a/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
+++ b/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
@@ -227,6 +227,11 @@ class EpollEventLoop extends SingleThreadEventLoop {
         this.ioRatio = ioRatio;
     }
 
+    @Override
+    public int registeredChannels() {
+        return channels.size();
+    }
+
     private int epollWait(boolean oldWakeup) throws IOException {
         // If a task was submitted when wakenUp value was 1, the task didn't get a chance to produce wakeup event.
         // So we need to check task queue again before calling epoll_wait. If we don't, the task might be pended
--- a/transport-native-kqueue/src/main/java/io/netty/channel/kqueue/KQueueEventLoop.java
+++ b/transport-native-kqueue/src/main/java/io/netty/channel/kqueue/KQueueEventLoop.java
@@ -310,6 +310,11 @@ final class KQueueEventLoop extends SingleThreadEventLoop {
     }
 
     @Override
+    public int registeredChannels() {
+        return channels.size();
+    }
+
+    @Override
     protected void cleanup() {
         try {
             try {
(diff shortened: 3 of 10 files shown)


normanmaurer:
@netty-bot test this please

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
```diff
@@ -229,6 +229,11 @@ public void setIoRatio(int ioRatio) {
         this.ioRatio = ioRatio;
     }
 
+    @Override
+    public int registeredChannels() {
+        return channels.size();
```
normanmaurer: I think this does not really work as `channels` is not thread-safe.

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/KQueueEventLoop.java
```diff
@@ -313,6 +313,11 @@ public void setIoRatio(int ioRatio) {
         this.ioRatio = ioRatio;
     }
 
+    @Override
+    public int registeredChannels() {
+        return jniChannelPointers.size();
```
normanmaurer: I think this does not really work as `jniChannelPointers` is not thread-safe.

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
vkostyukov: I'd imagine maintaining a separate (volatile) counter for the number of channels involves some overhead that we could possible avoid accepting that the numbers may not be super precise. This is, perhaps, not a big deal for observability but can surely hurt if we decide to load-balance channels based on it.

What do you think about hoisting such counter under `SingleThreadEvelLoop`; increment it on `register` and decrement it on `channel.closePromise` resolution?

normanmaurer: @vkostyukov the problem here is that it is possible that a `Channel` will be deregistered but not closed. Also a `Channel` may be re-registered to an other EventLoop.

normanmaurer: @vkostyukov did you see this ?

medusar: @normanmaurer @vkostyukov `channels` are stored in `IntObjectHashMap`, as for method `size()`, i think it will be ok even if the `IntObjectHashMap` is not thread safe, because sometimes we don't need the `size()` to be very precise. Maybe we can change the `registeredChannels ` method name or add some information in the doc ?

vkostyukov: Right. This operation is pretty racy by nature and maybe it's fine if we're not super precise (as long as we document it). I'd imagine these pointers will be flushed pretty quickly on modern hardware anyway.

Happy to add some JavaDoc here explaining the trade off.

normanmaurer: @vkostyukov that's "ok" with me... That said I think we should "fix" this for netty 5 fundamentally. See https://github.com/netty/netty/pull/8680

vkostyukov:
@normanmaurer What do you think is missing here? 

We just talked about this again and would love to have that in 4.x.

transport/src/main/java/io/netty/channel/SingleThreadEventLoop.java
```diff
@@ -148,6 +148,14 @@ public int pendingTasks() {
         return super.pendingTasks() + tailTasks.size();
     }
 
+    /**
+     * Returns the number of {@link Channel}s registered with this {@link EventLoop} or {@code -1}
+     * if operation is not supported.
```
normanmaurer: @vkostyukov please just add some sentence that the number may be not 100 % correct.

normanmaurer: Also please tag the method with `@UnstableAPI`

vkostyukov: Done. Let me know what do you think.

normanmaurer:
@vkostyukov looks good ... can you just add a test case and then we can merge

vkostyukov:
> @vkostyukov looks good ... can you just add a test case and then we can merge

Added tests. Decided to put them into an abstract test class `AbstractSingleThreadEventLoopTest` that's part of the `testsuite` target so packages can depend on it. Let me know what do you think.

testsuite/src/main/java/io/netty/testsuite/transport/AbstractSingleThreadEventLoopTest.java
```diff
@@ -0,0 +1,38 @@
+package io.netty.testsuite.transport;
```
normanmaurer: @vkostyukov please add license header.

testsuite/src/main/java/io/netty/testsuite/transport/AbstractSingleThreadEventLoopTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+public abstract class AbstractSingleThreadEventLoopTest {
+
+  @Test
+  public void testChannelsRegistered()  {
+    EventLoopGroup group = newEventLoopGroup();
+    final SingleThreadEventLoop loop = (SingleThreadEventLoop) group.next();
+
+    try {
+      final Channel ch1 = newChannel();
```
normanmaurer: @vkostyukov please use 4 spaces everywhere to match our coding style.

vkostyukov: Oh sorry about this. I'm used to formatters doing this automagically. The change is coming.

testsuite/src/main/java/io/netty/testsuite/transport/AbstractSingleThreadEventLoopTest.java
vkostyukov: Done.

testsuite/src/main/java/io/netty/testsuite/transport/AbstractSingleThreadEventLoopTest.java
vkostyukov: Done.

normanmaurer:
@netty-bot test this please

transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
```diff
@@ -356,6 +356,11 @@ public void run() {
         rebuildSelector0();
     }
 
+    @Override
+    public int registeredChannels() {
+        return selector.keys().size() - cancelledKeys;
```
vkostyukov: @normanmaurer I looked at the [failures in CI](https://ci.netty.io/job/netty-centos6-java11-prb/780/testReport/junit/io.netty.channel.nio/NioEventLoopTest/testChannelsRegistered/) and I think this should fix them. What do you think?

vkostyukov:
@normanmaurer Any chance you can trigger a CI again?

normanmaurer:
@netty-bot test this please

normanmaurer:
@vkostyukov done

normanmaurer:
@vkostyukov thanks

PR #8431, opened 2018-10-26
Add testing ByteBufAllocator for leak unit testing (almson version)

almson:
This is an alternate proposal for PR #8423 https://github.com/netty/netty/pull/8423 It makes fewer changes and is more general, in my opinion. Does not require subclassing the leak detector.

Motivation:

Currently the only way to detect leaks during testing is log scanning which is
difficult to automate. As an alternative it would be nice to fail unit tests
that leak buffers.

Modifications:

Added ResourceLeakDetector.assertAllResourcesDisposed which can be used in unit tests to ensure that all resources have been disposed. However, invoking assertAllResourcesDisposed directly can be problematic due to the fact that the leak detector is usually a static singleton and/or private in the classes that use it.

To that end, added TestingPooledByteBufAllocator which maintains a non-static leak detector and is useful for concurrent tests. It implements AutoCloseable (for use with try-with-resources) and invokes assertAllResourcesDisposed in the close method. To implement this, AbstractByteBufAllocator.toLeakAwareBuffer was made non-static to allow sub classes to use their own leak detector.

ResourceLeakDetector.clearRefQueue was removed because it created a danger that assertAllResourcesDisposed would not detect all leaks. This method is a useless optimization which saves a few cycles in the unlikely event that error logging is off, leaks are happening, and the leak detector is enabled.

Result:

Developers using Netty can now easily verify that all buffers have been released in unit tests.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/AbstractByteBufAllocator.java
+++ b/buffer/src/main/java/io/netty/buffer/AbstractByteBufAllocator.java
@@ -34,7 +34,7 @@ public abstract class AbstractByteBufAllocator implements ByteBufAllocator {
         ResourceLeakDetector.addExclusions(AbstractByteBufAllocator.class, "toLeakAwareBuffer");
     }
 
-    protected static ByteBuf toLeakAwareBuffer(ByteBuf buf) {
+    protected ByteBuf toLeakAwareBuffer(ByteBuf buf) {
         ResourceLeakTracker<ByteBuf> leak;
         switch (ResourceLeakDetector.getLevel()) {
             case SIMPLE:
@@ -56,7 +56,7 @@ public abstract class AbstractByteBufAllocator implements ByteBufAllocator {
         return buf;
     }
 
-    protected static CompositeByteBuf toLeakAwareBuffer(CompositeByteBuf buf) {
+    protected CompositeByteBuf toLeakAwareBuffer(CompositeByteBuf buf) {
         ResourceLeakTracker<ByteBuf> leak;
         switch (ResourceLeakDetector.getLevel()) {
             case SIMPLE:
--- a/buffer/src/main/java/io/netty/buffer/TestingPooledByteBufAllocator.java
+++ b/buffer/src/main/java/io/netty/buffer/TestingPooledByteBufAllocator.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.netty.buffer;
+
+import io.netty.util.ResourceLeakDetector;
+import io.netty.util.ResourceLeakTracker;
+
+import java.io.Closeable;
+
+/**
+ * A {@link ByteBufAllocator} for unit testing. This allocator maintains a private, non-static leak detector
+ * and performs leak tracking on all allocated buffers. At the end the test, call {@link #close} to assert
+ * that all buffers have been freed.
+ */
+public class TestingPooledByteBufAllocator extends PooledByteBufAllocator implements Closeable {
+
+    private final ResourceLeakDetector leakDetector = new ResourceLeakDetector(AbstractByteBuf.class, 1);
+
+    {
+        // While we might try to set the level manually, or try other hacks,
+        // this check maintains generality should ResourceLeakDetector change in the future.
+        // Don't run your tests with leak detection disabled.
+        if (!ResourceLeakDetector.isEnabled()) {
+            throw new AssertionError("ResourceLeakDetector.isEnabled() returns false. Cannot continue.");
+        }
+    }
+
+    public TestingPooledByteBufAllocator() {
+    }
+
+    public TestingPooledByteBufAllocator(boolean preferDirect) {
+        super(preferDirect);
+    }
+
+    @Override
+    public ByteBuf toLeakAwareBuffer(ByteBuf buf) {
+        ResourceLeakTracker<ByteBuf> leak = leakDetector.track(buf);
+        if (leak != null) {
+            buf = new AdvancedLeakAwareByteBuf(buf, leak);
+        }
+        return buf;
+    }
+
+    @Override
+    public CompositeByteBuf toLeakAwareBuffer(CompositeByteBuf buf) {
+        ResourceLeakTracker<ByteBuf> leak = leakDetector.track(buf);
+        if (leak != null) {
+            buf = new AdvancedLeakAwareCompositeByteBuf(buf, leak);
+        }
+        return buf;
+    }
+
+    /**
+     * Throws an exception if all allocated buffers have been not freed.
+     * @throws AssertionError if buffers have been leaked.
+     */
+    @Override
+    public void close() {
+        leakDetector.assertAllResourcesDisposed();
+    }
+}
--- a/buffer/src/test/java/io/netty/buffer/TestingPooledByteBufAllocatorTest.java
+++ b/buffer/src/test/java/io/netty/buffer/TestingPooledByteBufAllocatorTest.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.netty.buffer;
+
+import org.junit.Test;
+
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.junit.Assert.assertThat;
+
+public class TestingPooledByteBufAllocatorTest {
+
+    @Test
+    public void testNoLeakByteBuffer() {
+        TestingPooledByteBufAllocator allocator = new TestingPooledByteBufAllocator();
+        try {
+            allocator.buffer(100).release();
+        } finally {
+            allocator.close();
+        }
+    }
+
+    @Test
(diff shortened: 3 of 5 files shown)


almson:
@dain It will assert them. All of those errors are stored in `Set<String> reportedLeaks`, even if logging is disabled.

almson:
@dain I'm new to GitHub. I see it says "dain suggested changes View Changes" but I can't seem to see what's different. Did you suggest a code change?

[](https://postimg.cc/jDVQp0kZ)

dain:
@almson, I see it now.  All reported leaks are already stored in `ConcurrentMap<String, Boolean> reportedLeaks`.  I'll read this again tomorrow.

BTW, I didn't make any suggested changes, and I'm not sure why Github said that.  That is a brand new feature, so maybe it was a bug.

buffer/src/main/java/io/netty/buffer/AbstractByteBufAllocator.java
```diff
@@ -34,7 +34,7 @@
         ResourceLeakDetector.addExclusions(AbstractByteBufAllocator.class, "toLeakAwareBuffer");
     }
 
-    protected static ByteBuf toLeakAwareBuffer(ByteBuf buf) {
+    protected ByteBuf toLeakAwareBuffer(ByteBuf buf) {
```
normanmaurer: This is a breaking change so you will need to preserve the old method signature as well as people may depend on it to be static atm when sub-classing `AbstractByteBufAllocator`

buffer/src/main/java/io/netty/buffer/AbstractByteBufAllocator.java
```diff
@@ -56,7 +56,7 @@ protected static ByteBuf toLeakAwareBuffer(ByteBuf buf) {
         return buf;
     }
 
-    protected static CompositeByteBuf toLeakAwareBuffer(CompositeByteBuf buf) {
+    protected CompositeByteBuf toLeakAwareBuffer(CompositeByteBuf buf) {
```
normanmaurer: same comment as above

buffer/src/main/java/io/netty/buffer/TestingPooledByteBufAllocator.java
```diff
@@ -0,0 +1,75 @@
...
+import io.netty.util.ResourceLeakTracker;
+
+import java.io.Closeable;
+
+/**
+ * A {@link ByteBufAllocator} for unit testing. This allocator maintains a private, non-static leak detector
+ * and performs leak tracking on all allocated buffers. At the end the test, call {@link #close} to assert
+ * that all buffers have been freed.
+ */
+public class TestingPooledByteBufAllocator extends PooledByteBufAllocator implements Closeable {
```
normanmaurer: As this is mainly used for testing I think we should better extend `UnpooledByteBufAllocator`.

normanmaurer: Also maybe we should call it `AssertingByteBufAllocator`.

buffer/src/test/java/io/netty/buffer/TestingPooledByteBufAllocatorTest.java
```diff
@@ -0,0 +1,99 @@
...
+ */
+
+package io.netty.buffer;
+
+import org.junit.Test;
+
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.junit.Assert.assertThat;
+
+public class TestingPooledByteBufAllocatorTest {
```
normanmaurer: when rename the class also rename the test please.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -247,39 +247,29 @@ public final ResourceLeak open(T obj) {
...
             }
-            ref.dispose();
+            return false;
         }
+        return true;
     }
 
     private void reportLeak() {
-        if (!logger.isErrorEnabled()) {
-            clearRefQueue();
```
normanmaurer: @almson shouldn't we presume this ?

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -303,6 +293,42 @@ private void reportLeak() {
         }
     }
 
+    /**
+     * Throws AssertionError if there's been any leaks.
```
normanmaurer: `{@link AssertionError}`.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -303,6 +293,42 @@ private void reportLeak() {
...
+            String records = leak.toString();
+            if (reportedLeaks.putIfAbsent(records, Boolean.TRUE) == null) {
+                if (records.isEmpty()) {
+                    reportUntracedLeak(resourceType);
+                } else {
+                    reportTracedLeak(resourceType, records);
+                }
+            }
+        }
+        if (!reportedLeaks.keySet().isEmpty()) {
```
normanmaurer: Store the `Set` before calling `isEmpty()` a you also access it in the for loop.

buffer/src/main/java/io/netty/buffer/TestingPooledByteBufAllocator.java
```diff
@@ -0,0 +1,75 @@
...
+
+package io.netty.buffer;
+
+import io.netty.util.ResourceLeakDetector;
+import io.netty.util.ResourceLeakTracker;
+
+import java.io.Closeable;
+
+/**
+ * A {@link ByteBufAllocator} for unit testing. This allocator maintains a private, non-static leak detector
```
normanmaurer: maybe reword to:

```
A  {@link ByteBufAllocator} that can enforce an `AssertionError` when any leaks are detected.
```

buffer/src/main/java/io/netty/buffer/TestingPooledByteBufAllocator.java
```diff
@@ -0,0 +1,75 @@
...
+        }
+        return buf;
+    }
+
+    /**
+     * Throws an exception if all allocated buffers have been not freed.
+     * @throws AssertionError if buffers have been leaked.
+     */
+    @Override
+    public void close() {
```
normanmaurer: I wonder if we should better *not* implement `Closeable` but just expose a method like `assertNoLeaks()` as technical speaking even after `close` is called the allocator is still usable and so not really "closed".

buffer/src/main/java/io/netty/buffer/TestingPooledByteBufAllocator.java
```diff
@@ -0,0 +1,75 @@
...
+    public CompositeByteBuf toLeakAwareBuffer(CompositeByteBuf buf) {
+        ResourceLeakTracker<ByteBuf> leak = leakDetector.track(buf);
+        if (leak != null) {
+            buf = new AdvancedLeakAwareCompositeByteBuf(buf, leak);
+        }
+        return buf;
+    }
+
+    /**
+     * Throws an exception if all allocated buffers have been not freed.
```
normanmaurer: `Throws an {@link AssertionError} if all allocated {@link ByteBuf}s ...`

buffer/src/main/java/io/netty/buffer/TestingPooledByteBufAllocator.java
```diff
@@ -0,0 +1,75 @@
...
+import java.io.Closeable;
+
+/**
+ * A {@link ByteBufAllocator} for unit testing. This allocator maintains a private, non-static leak detector
+ * and performs leak tracking on all allocated buffers. At the end the test, call {@link #close} to assert
+ * that all buffers have been freed.
+ */
+public class TestingPooledByteBufAllocator extends PooledByteBufAllocator implements Closeable {
+
+    private final ResourceLeakDetector leakDetector = new ResourceLeakDetector(AbstractByteBuf.class, 1);
```
normanmaurer: Do not hardcode `1` but add a constructor to the `allocator` that allows to specify the sampling rate.

normanmaurer:
@netty-bot test this please

normanmaurer:
@almson also there is a checkstyle error.

buffer/src/main/java/io/netty/buffer/TestingPooledByteBufAllocator.java
dain: @normanmaurer, what are the implications of extending `UnpooledByteBufAllocator`?  I ask because I always use pooled in my tests because that is what I use in production.  IIRC, the unpooled uses different methods for the leak callbacks.

buffer/src/main/java/io/netty/buffer/TestingPooledByteBufAllocator.java
dain: In my original proposal, I went with `Closeable` because it makes using this in tests very easy because you can use try with resources syntax.  So it is a tradeoff of convenient test code verses explicit naming.  I went with the former because the code gets pretty deeply nested due to the other IO related bits that need try with resources.

buffer/src/main/java/io/netty/buffer/TestingPooledByteBufAllocator.java
```diff
@@ -0,0 +1,75 @@
...
+
+    {
+        // While we might try to set the level manually, or try other hacks,
+        // this check maintains generality should ResourceLeakDetector change in the future.
+        // Don't run your tests with leak detection disabled.
+        if (!ResourceLeakDetector.isEnabled()) {
```
dain: I believe you need to override the `shouldTrack` method instead of checking the level here.  For this class to work perfectly, the level would be required to be `PARANOID`, and since level is a global static, that requirement would effect every test in the system.

trustin:
What do you think about building a plugin for testing framework or a JVM agent to make this functionality available without making any user code change? A user has to add some code to take advantage of this, but it's not always doable especially for integration tests that involve a lot of components. A user would find it is tedious to add try-with-resources or assertion for all test cases.

trustin:
For example, we could provide a new artifact 'netty-testing-junit4' and provide a RunListener implementation and document how to plug it into a build tool.

normanmaurer:
@trustin this sounds like an awesome idea actually

trustin:
I agree though that this still may be useful if a user needs to find a leak in a very specific scope, but I guess the problem we are trying to solve here in general is to fail the build reliably when there is a leak regardless of where it really is? Given the nature of how leak detector works, the point of leak detection may be much later then when a buffer actually leaked.

dain:
I agree that that might be interesting to those trying to catch general leaks.  In my case, I am interested in failing the specific test that created the leak, and I run tests in parallel, so a global tracker would not work for me.

@normanmaurer  or @trustin, can we get a call on which direction you want to go for this feature?  As I mentioned above, I believe the main difference is the addition of the new method `ResourceLeakDectector.assertAllResourcesDisposed()` to dispose all references and assert no leaks, whereas the other PR #8423 adds a method to "disposes" all references, and assert leaks in a separate "subclass" method.  If you want to go with my PR (#8423), I can may any changes you need quickly, but I'm more interested in getting this feature added, than my specific proposal.

normanmaurer:
Let me close this ... I think a better approach would be to do something with a `RunListener`: https://junit.org/junit4/javadoc/4.12/org/junit/runner/notification/RunListener.html

almson:
Do you have such an approach? Or are you speculating there might be an approach?

PR #8406, opened 2018-10-18
Improve CompositeByteBuf sequential access performance

njhill:
Motivation:

While running some other benchmarks I noticed that sequential access of `CompositeByteBuf`s is very slow, particularly if they are large.
 
Modifications:

- Override `forEachByte(...)` methods
- Remove redundant `Component.length` field and standardize component offset modifications via `setOffset()`
- Simplify the general pattern used where a sub-range is iterated over
- Add racy cache of last-accessed Component to `findComponent(int)` method
- Add benchmark to demonstrate improvement

The cache is based on the fact that a call to `findComponent` precededs every individual read/write access, and the assumption that very often the same component will be needed between successive accesses.
 
Result:

I will paste benchmark results below. Also some anecdotal observations:
- There is a "cache hit" rate of 75% when running the existing buffer package unit tests
- The complete buffer package unit tests appear to take 30% less time to run if `AbstractByteBufTest.CAPACITY` is increased by a factor of 10

In addition to the raw speed there should also be a minor mem/garbage benefit due to removal of `Component.length`.

PR #8437, opened 2018-10-27
Streamline CompositeByteBuf internals

njhill:
Motivation:

`CompositeByteBuf` is a powerful and versatile abstraction, allowing for manipulation of large data without copying bytes. There is still a non-negligible cost to reading/writing however relative to "singular" `ByteBufs`, and this can be mostly eliminated with some rework of the internals.

My use case is message modification/transformation while zero-copy proxying. For example replacing a string within a large message with one of a different length

Modifications:
- No longer slice added buffers and unwrap added slices
   - Components store target buf offset relative to position in composite buf
   - Less allocations, object footprint, pointer indirection, offset arithmetic
- Use `Component[]` rather than `ArrayList<Component>`
   - Avoid pointer indirection and duplicate bounds check, more efficient backing array growth
   - Facilitates optimization when doing bulk-inserts - inserting _n_ `ByteBuf`s behind _m_ is now _O(m + n)_ instead of _O(mn)_
- Avoid unnecessary casting and method call indirection via superclass
- Eliminate some duplicate range/ref checks via non-checking versions of `toComponentIndex` and `findComponent`
- Add simple fast-path for `toComponentIndex(0)`; add racy cache of last-accessed Component to `findComponent(int)`
- Override `forEachByte0(...)` and `forEachByteDesc0(...)` methods
- Make use of `RecyclableArrayList` in `nioBuffers(int, int)` (in line with `FasterCompositeByteBuf` impl)
- Modify `addComponents0(boolean,int,Iterable)` to use the `Iterable` directly rather than copy to an array first (and possibly to an `ArrayList` before that)
- Optimize `addComponents0(boolean,int,ByteBuf[],int)` to not perform repeated array insertions and avoid second loop for offset updates
- Simplify other logic in various places, in particular the general pattern used where a sub-range is iterated over
- Add benchmarks to demonstrate some improvements

While refactoring I also came across a couple of clear bugs. They are fixed in these changes but I will open another PR with unit tests and fixes to the current version.

Result:

Much faster creation, manipulation, and access; many fewer allocations and smaller footprint. Benchmark results to follow.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
@@ -46,7 +46,7 @@ public abstract class AbstractByteBuf extends ByteBuf {
     private static final InternalLogger logger = InternalLoggerFactory.getInstance(AbstractByteBuf.class);
     private static final String LEGACY_PROP_CHECK_ACCESSIBLE = "io.netty.buffer.bytebuf.checkAccessible";
     private static final String PROP_CHECK_ACCESSIBLE = "io.netty.buffer.checkAccessible";
-    private static final boolean checkAccessible;
+    static final boolean checkAccessible; // accessed from CompositeByteBuf
     private static final String PROP_CHECK_BOUNDS = "io.netty.buffer.checkBounds";
     private static final boolean checkBounds;
 
@@ -332,11 +332,11 @@ public abstract class AbstractByteBuf extends ByteBuf {
 
     @Override
     public ByteBuf order(ByteOrder endianness) {
-        if (endianness == null) {
-            throw new NullPointerException("endianness");
-        }
         if (endianness == order()) {
             return this;
+        }
+        if (endianness == null) {
+            throw new NullPointerException("endianness");
         }
         return newSwappedByteBuf();
     }
@@ -1293,7 +1293,7 @@ public abstract class AbstractByteBuf extends ByteBuf {
         }
     }
 
-    private int forEachByteAsc0(int start, int end, ByteProcessor processor) throws Exception {
+    int forEachByteAsc0(int start, int end, ByteProcessor processor) throws Exception {
         for (; start < end; ++start) {
             if (!processor.process(_getByte(start))) {
                 return start;
@@ -1325,7 +1325,7 @@ public abstract class AbstractByteBuf extends ByteBuf {
         }
     }
 
-    private int forEachByteDesc0(int rStart, final int rEnd, ByteProcessor processor) throws Exception {
+    int forEachByteDesc0(int rStart, final int rEnd, ByteProcessor processor) throws Exception {
         for (; rStart >= rEnd; --rStart) {
             if (!processor.process(_getByte(rStart))) {
                 return rStart;
--- a/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
@@ -15,7 +15,11 @@
  */
 package io.netty.buffer;
 
+import io.netty.util.ByteProcessor;
+import io.netty.util.IllegalReferenceCountException;
+import io.netty.util.ReferenceCountUtil;
 import io.netty.util.internal.EmptyArrays;
+import io.netty.util.internal.RecyclableArrayList;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -26,12 +30,12 @@ import java.nio.channels.FileChannel;
 import java.nio.channels.GatheringByteChannel;
 import java.nio.channels.ScatteringByteChannel;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.ConcurrentModificationException;
 import java.util.Iterator;
 import java.util.List;
-import java.util.ListIterator;
 import java.util.NoSuchElementException;
 
 import static io.netty.util.internal.ObjectUtil.checkNotNull;
@@ -48,8 +52,10 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
 
     private final ByteBufAllocator alloc;
     private final boolean direct;
-    private final ComponentList components;
     private final int maxNumComponents;
+
+    private int componentCount;
+    private Component[] components; // resized when needed
 
     private boolean freed;
 
@@ -65,7 +71,7 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
         this.alloc = alloc;
         this.direct = direct;
         this.maxNumComponents = maxNumComponents;
-        components = newList(initSize, maxNumComponents);
+        components = newCompArray(initSize, maxNumComponents);
     }
 
     public CompositeByteBuf(ByteBufAllocator alloc, boolean direct, int maxNumComponents) {
@@ -73,16 +79,16 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
     }
 
     public CompositeByteBuf(ByteBufAllocator alloc, boolean direct, int maxNumComponents, ByteBuf... buffers) {
-        this(alloc, direct, maxNumComponents, buffers, 0, buffers.length);
+        this(alloc, direct, maxNumComponents, buffers, 0);
     }
 
     CompositeByteBuf(ByteBufAllocator alloc, boolean direct, int maxNumComponents,
-            ByteBuf[] buffers, int offset, int endOffset) {
-        this(alloc, direct, maxNumComponents, endOffset - offset);
+            ByteBuf[] buffers, int offset) {
+        this(alloc, direct, maxNumComponents, buffers.length - offset);
 
-        addComponents0(false, 0, buffers, offset, endOffset);
+        addComponents0(false, 0, buffers, offset);
         consolidateIfNeeded();
-        setIndex(0, capacity());
+        setIndex0(0, capacity());
     }
 
     public CompositeByteBuf(
@@ -132,9 +138,9 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
         setIndex(0, capacity());
     }
 
-    private static ComponentList newList(int initComponents, int maxNumComponents) {
+    private static Component[] newCompArray(int initComponents, int maxNumComponents) {
         int capacityGuess = Math.min(AbstractByteBufAllocator.DEFAULT_MAX_COMPONENTS, maxNumComponents);
-        return new ComponentList(Math.max(initComponents, capacityGuess));
+        return new Component[Math.max(initComponents, capacityGuess)];
     }
 
     // Special constructor used by WrappedCompositeByteBuf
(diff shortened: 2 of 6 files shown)


normanmaurer:
@netty-bot test this please

njhill:
@normanmaurer I have re-based and fixed the failing unit tests (sorry, I just ran the buffer ones and forgot to run the whole suite). The failures were due to ref-count of components no longer being checked at addition time - it was done implicitly before when they were sliced.

normanmaurer:
@netty-bot test this please

njhill:
@normanmaurer I've now squashed and rebased this. Take your time! :)

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -46,7 +46,7 @@
     private static final InternalLogger logger = InternalLoggerFactory.getInstance(AbstractByteBuf.class);
     private static final String LEGACY_PROP_CHECK_ACCESSIBLE = "io.netty.buffer.bytebuf.checkAccessible";
     private static final String PROP_CHECK_ACCESSIBLE = "io.netty.buffer.checkAccessible";
-    private static final boolean checkAccessible;
+    static final boolean checkAccessible; // accessed from CompositeByteBuf
```
mosesn:
```suggestion
    protected static final boolean checkAccessible; // accessed from CompositeByteBuf
```

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -332,12 +332,12 @@ public int ensureWritable(int minWritableBytes, boolean force) {
 
     @Override
     public ByteBuf order(ByteOrder endianness) {
-        if (endianness == null) {
-            throw new NullPointerException("endianness");
-        }
         if (endianness == order()) {
             return this;
         }
+        if (endianness == null) {
```
mosesn: this change looks unnecessary, can you revert it?

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -1293,7 +1293,7 @@ public int forEachByte(int index, int length, ByteProcessor processor) {
         }
     }
 
-    private int forEachByteAsc0(int start, int end, ByteProcessor processor) throws Exception {
+    int forEachByteAsc0(int start, int end, ByteProcessor processor) throws Exception {
```
mosesn:
```suggestion
    protected int forEachByteAsc0(int start, int end, ByteProcessor processor) throws Exception {
```

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -1325,7 +1325,7 @@ public int forEachByteDesc(int index, int length, ByteProcessor processor) {
         }
     }
 
-    private int forEachByteDesc0(int rStart, final int rEnd, ByteProcessor processor) throws Exception {
+    int forEachByteDesc0(int rStart, final int rEnd, ByteProcessor processor) throws Exception {
```
mosesn:
```suggestion
    protected int forEachByteDesc0(int rStart, final int rEnd, ByteProcessor processor) throws Exception {
```

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
```diff
@@ -1431,21 +1477,40 @@ public ByteBuf internalComponent(int cIndex) {
...
+    private Component findComponent0(int offset) {
+        Component la = lastAccessed;
+        if (la != null && offset >= la.offset && offset < la.endOffset) {
+           return la;
+        }
+        return findIt(offset);
+    }
 
-        for (int low = 0, high = components.size(); low <= high;) {
+    private Component findIt(int offset) {
```
mosesn: this always searches one direction.  can we do better in large componentbuffers by checking which direction would be faster to search in?  also, if lastAccessed is close to the target offset, should we start from there?

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
```diff
@@ -1519,54 +1585,53 @@ public ByteBuffer nioBuffer(int index, int length) {
...
+
+                index += localLength;
+                length -= localLength;
+                i ++;
             }
 
-            index += localLength;
-            length -= localLength;
-            i ++;
+            return buffers.toArray(new ByteBuffer[0]);
```
mosesn: should we cache an empty array?

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
njhill: This change eliminates a null check in the most common case (BE) and doesn't change the number of checks in the other cases, is there a reason the other way round is preferable?

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
njhill: There was some discussion about this general question [here](https://github.com/netty/netty/pull/8075#issuecomment-401143387), I think the conclusion (and current precedent elsewhere in the netty code) is to _not_ cache such empty arrays.

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
njhill: I'm not sure what you mean by one direction here since it's a binary search. I did experiment a little with choosing the next component based on the target byte location within the remaining range rather than just choosing the middle component each time, but the cost of the additional ops appeared to outweigh any benefits (I expect with more attention _some_ improvement could be made though).

> if lastAccessed is close to the target offset, should we start from there?

@normanmaurer [suggested the same thing](https://github.com/netty/netty/pull/8406#discussion_r226628136). It would complicate the logic since the index would need to be cached separately (which would not be updated atomically with the component ref), or an index field could be added to the `Component` class which seemed overkill just for this purpose. But I'd be happy to experiment with it and see if there's measurable diff in the numbers.

njhill:
Thanks for the review @mosesn 
> where do most of the benefits come from? are they from the iterator and byte processing?

The biggest all-round benefits appear to be from the removal of the slices and component ArrayList. The ByteProcessor methods obviously benefit significantly from being overridden. Sequential reading/writing benefits significantly from the "lastAccessed" cache. Most of the other changes benefit the speed of construction and manipulation of the components.

normanmaurer:
@netty-bot test this please

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
normanmaurer: I thought it may be better to start with more restricted level. We can open it up later if we really want to.

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
normanmaurer: I don't have a preference here... that said null checks are really cheap. So whatever is more readable is the preferred way imho :)

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
normanmaurer: I think we can do this as a followup. This PR is already big enough .. @mosesn ok with it ?

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
normanmaurer: @mosesn I think we dont need to and we also dont do in many other places.

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
mosesn: yep, that's fine.

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
mosesn: I don't love that it changes the git blame, but that seems reasonable.  shipit.

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
```diff
@@ -2015,16 +2161,64 @@ public void remove() {
...
+    }
+
+    private void addComp(int i, Component c) {
+        // assert i >= 0 && i <= size
+        shiftComps(i, 1);
+        components[i] = c;
+    }
+
+    private void shiftComps(int i, int count) {
+        // assert i >= 0 && i <= size && count > 0
```
normanmaurer: @njhill either remove or remove `//`.

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
```diff
@@ -2015,16 +2161,64 @@ public void remove() {
...
+        int newSize = size - to + from;
+        for (int i = newSize; i < size; i++) {
+            components[i].slice = null;
+            components[i] = null;
+        }
+        componentCount = newSize;
+    }
+
+    private void addComp(int i, Component c) {
+        // assert i >= 0 && i <= size
```
normanmaurer: @njhill either remove or remove `//`.

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
```diff
@@ -2015,16 +2161,64 @@ public void remove() {
...
+    private void removeComp(int i) {
+        removeCompRange(i, i + 1);
+    }
+
+    private void removeCompRange(int from, int to) {
+        if (from >= to) {
+            return;
+        }
+        final int size = componentCount;
+        // assert from >= 0 && from < to && to <= size
```
normanmaurer: @njhill either remove or remove `//`.

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
```diff
@@ -398,37 +418,34 @@ public CompositeByteBuf addComponents(int cIndex, Iterable<ByteBuf> buffers) {
         return this;
     }
 
+    //TODO optimize further, similar to ByteBuf[] version
```
normanmaurer: nit: add white space after //

njhill:
@normanmaurer @mosesn I've added https://github.com/njhill/netty/commit/b8ed3ba83d23fec5d36e1f0492de93707c0f9eeb to address the comments; it also includes a couple of minor tweaks to things I noticed while looking over it again.

Strangely despite showing in the branch, the new commit & changes don't seem to show in this PR... not sure what github is up to, maybe they will appear at some point.

normanmaurer:
@netty-bot test this please

normanmaurer:
@njhill thanks a lot!

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
```diff
@@ -316,43 +339,40 @@ private int addComponent0(boolean increaseWriterIndex, int cIndex, ByteBuf buffe
...
-                }
-                cIndex = addComponent0(increaseWriterIndex, cIndex, b) + 1;
-                int size = components.size();
-                if (cIndex > size) {
-                    cIndex = size;
-                }
+            shiftComps(cIndex, count); // will increase componentCount
+            ci = cIndex; // only set this after we've shifted so that finally block logic is always correct
+            int nextOffset = cIndex > 0 ? components[cIndex - 1].endOffset : 0;
+            for (ByteBuf b; arrOffset < len && (b = buffers[arrOffset]) != null; arrOffset++, ci++) {
```
carl-mastrangelo: This is kind of difficult to read.   You should consider making you code slightly more verbose, but faster to consume.

njhill: Thanks @carl-mastrangelo, I do have a habit of over-compactness sometimes. I have some smaller follow-on updates to the class in mind anyhow so will try and modify this to be a bit more verbose/readable.

Issue #8442, opened 2018-10-30
Custom cipher list with BoringSSL breaks in 4.1.31.Final 


njhill:
### Expected behavior

TLS continues to work in 4.1.31.Final when providing an explicit list of ciphers via `SslContextBuilder.ciphers(...)`.

### Actual behavior

After upgrading, my `SslContextBuilder.build()` throws:

```
Caused by: java.lang.IllegalArgumentException: unsupported cipher suite: TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA384(ECDHE-ECDSA-AES256-SHA384)
       at io.netty.handler.ssl.CipherSuiteConverter.convertToCipherStrings(CipherSuiteConverter.java:418)
       at io.netty.handler.ssl.ReferenceCountedOpenSslContext.<init>(ReferenceCountedOpenSslContext.java:252)
       ... 34 more
```

It looks like this is due to changes in #8293, in particular [this](https://github.com/netty/netty/blob/netty-4.1.31.Final/handler/src/main/java/io/netty/handler/ssl/CipherSuiteConverter.java#L418) check in `CipherSuiteConverter.convertToCipherStrings()` which throws if false. Should it just `continue` instead of throwing so that the current cipherSuite is excluded from the lists being built (since a later cipher in the list may be available)? I verified that it works as before with this modification.

### Minimal yet complete reproducer code (or URL to code)

```java
KeyManagerFactory kmf = KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm());
kmf.init(null,null);
String[] ciphers = ((SSLSocketFactory)SSLSocketFactory.getDefault()).getDefaultCipherSuites();
SslContextBuilder.forServer(kmf).sslProvider(SslProvider.OPENSSL) 
    .ciphers(Arrays.asList(ciphers)).build();
```

### Netty version

4.1.31.Final

### JVM version (e.g. `java -version`)

```
java version "1.8.0_181"
Java(TM) SE Runtime Environment (build 1.8.0_181-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.181-b13, mixed mode)
```

### OS version (e.g. `uname -a`)

`Linux 3.10.0-862.14.4.el7.x86_64`

normanmaurer:
And this did not throw before ? I wonder why as I think the check was always there or I am missing something ?

normanmaurer:
[...]

PR #8443, opened 2018-10-30
Add test to verify that invalid ciphers are handled in all SSLEngine …

normanmaurer:
…implementations correctly.

Motivation:

https://github.com/netty/netty/issues/8442 reported that we fail to build a SslContext when an invalid cipher is used with netty-tcnative-boringssl-static, while it worked before. This test verifies that this is now consistent with all other SSLEngine implementations.

Modifications:

Add test-case to verify consistent behaviour

Result:

More tests to assert consistent behaviour across SSLEngine implementations


normanmaurer:
@njhill PTAL

njhill:
LGTM, thanks!

normanmaurer:
@njhill thanks for the review

PR #8445, opened 2018-10-30
Fix NPE when trying to build a DnsNameResolver with a null resolvedAddressTypes

slandelle:
Motivation:

It should be possible to build a DnsNameResolver with a null resolvedAddressTypes, defaulting then to DEFAULT_RESOLVE_ADDRESS_TYPES (see line 309).

Sadly, `preferredAddressType` is then called on line 377 with the original parameter instead of the instance attribute, causing an NPE when it's null.

Modification:

Call preferredAddressType with instance attribuet instead of constructor parameter.

Result:

No more NPE

### DIFF
--- a/resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
+++ b/resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
@@ -2403,4 +2403,27 @@ public class DnsNameResolverTest {
             }
         }
     }
+
+    @Test
+    public void testInstanceWithNullPreferredAddressType() {
+        new DnsNameResolver(
+                group.next(), // eventLoop
+                new ReflectiveChannelFactory<DatagramChannel>(NioDatagramChannel.class), // channelFactory
+                NoopDnsCache.INSTANCE, // resolveCache
+                NoopAuthoritativeDnsServerCache.INSTANCE, // authoritativeDnsServerCache
+                NoopDnsQueryLifecycleObserverFactory.INSTANCE, // dnsQueryLifecycleObserverFactory
+                100, // queryTimeoutMillis
+                null, // resolvedAddressTypes, see https://github.com/netty/netty/pull/8445
+                true, // recursionDesired
+                1, // maxQueriesPerResolve
+                false, // traceEnabled
+                4096, // maxPayloadSize
+                true, // optResourceEnabled
+                HostsFileEntriesResolver.DEFAULT, // hostsFileEntriesResolver
+                DnsServerAddressStreamProviders.platformDefault(), // dnsServerAddressStreamProvider
+                null, // searchDomains
+                1, // ndots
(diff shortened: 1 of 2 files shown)


normanmaurer:
@slandelle this looks good... Can you just also add a test ?

slandelle:
@normanmaurer done
I'm pretty swamped atm, but I'll try in the next days to stop using the deprecated constructor.

resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
```diff
@@ -17,24 +17,12 @@
...
 import io.netty.channel.socket.nio.NioDatagramChannel;
-import io.netty.handler.codec.dns.DefaultDnsQuestion;
-import io.netty.handler.codec.dns.DnsQuestion;
-import io.netty.handler.codec.dns.DnsRawRecord;
-import io.netty.handler.codec.dns.DnsRecord;
-import io.netty.handler.codec.dns.DnsRecordType;
-import io.netty.handler.codec.dns.DnsResponse;
-import io.netty.handler.codec.dns.DnsResponseCode;
-import io.netty.handler.codec.dns.DnsSection;
+import io.netty.handler.codec.dns.*;
```
normanmaurer: @slandelle can you please revert all the `wildcard import` changes ?

resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
```diff
@@ -2403,4 +2364,27 @@ public boolean clear(String hostname) {
             }
         }
     }
+
+    @Test
+    public void testInstanceWithNullPreferredAddressType() {
+        new DnsNameResolver(
```
normanmaurer: please also close it.

normanmaurer:
@slandelle thanks... two things to change tho. Also sorry that I was not more responsible for the dns stuff in the last weeks. Also super busy here :(

normanmaurer:
@netty-bot test this please

resolver-dns/src/test/java/io/netty/resolver/dns/DnsNameResolverTest.java
```diff
@@ -67,39 +62,17 @@
...
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
+import java.util.*;
```
normanmaurer: @slandelle please revert wildcard changes in this file.

slandelle:
done

normanmaurer:
@slandelle thanks a lot!

slandelle:
Thanks for the review and merging!

Issue #8444, opened 2018-10-30
epoll_wait fails with "timerfd_settime() failed: Invalid argument" due to tv_sec and tv_nsec being -1


wilkinsona:
### Expected behavior

epoll_wait works as it did in 4.1.29

### Actual behavior

epoll_wait fails when the optimisation added in #7816 is used. `timerfd_settime()` is called with the following `itimerspec`:

```
Thread 26 "java" hit Catchpoint 1 (returned from syscall timerfd_settime), 0x00007faff30e15da in timerfd_settime () at ../sysdeps/unix/syscall-template.S:78
78	in ../sysdeps/unix/syscall-template.S
(gdb) print (struct itimerspec) *$rdx
$9 = {it_interval = {tv_sec = 0, tv_nsec = 0}, it_value = {tv_sec = -1, tv_nsec = -1}}
```

This is a continuation of the problem reported in #8348.

By my reading of the code, `timerfd_settime` shouldn't be called when both `tvSec` and `tvNsec` are -1. That makes me suspect that this `if` is not evaluating as expected: https://github.com/netty/netty/blob/44c3b824ecabe5d4bf52d3a9a7a30c456ee391b8/transport-native-epoll/src/main/c/netty_epoll_native.c#L187-L196

### Steps to reproduce

TBD

### Minimal yet complete reproducer code (or URL to code)

Sorry, I don't have one yet. I will update the issue if/when I do.

### Netty version

4.1.31.Final

### JVM version (e.g. `java -version`)

```
openjdk version "1.8.0_181"
OpenJDK Runtime Environment (build 1.8.0_181-8u181-b13-0ubuntu0.18.04.1-b13)
OpenJDK 64-Bit Server VM (build 25.181-b13, mixed mode)
```

### OS version (e.g. `uname -a`)

```
Linux ubuntu 4.15.0-29-generic #31-Ubuntu SMP Tue Jul 17 15:39:52 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

normanmaurer:
@wilkinsona I have no idea what is going on... I was not able to reproduce with an isolated test-case that will pass in -1:

https://github.com/netty/netty/pull/8447

Are you sure that there is not somehow a different native lib on the class path ?

wilkinsona:
[...]

PR #8447, opened 2018-10-30
Add testcase for epollWait(...) with negative timerfd values.

normanmaurer:
Motivation:

https://github.com/netty/netty/issues/8444 reports that there is some issue with negative values passed to timerfd_settime. This test verifies that everything is working as expected.

Modifications:

Add testcase.

Result:

Test to verify expected behaviour.

### DIFF
--- a/transport-native-epoll/src/test/java/io/netty/channel/epoll/EpollTest.java
+++ b/transport-native-epoll/src/test/java/io/netty/channel/epoll/EpollTest.java
@@ -15,13 +15,58 @@
  */
 package io.netty.channel.epoll;
 
-import org.junit.Assert;
+import io.netty.channel.unix.FileDescriptor;
 import org.junit.Test;
+
+import java.util.concurrent.atomic.AtomicReference;
+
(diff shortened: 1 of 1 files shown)


transport-native-epoll/src/test/java/io/netty/channel/epoll/EpollTest.java
```diff
@@ -15,13 +15,58 @@
...
+    }
+
+    // Testcase for https://github.com/netty/netty/issues/8444
+    @Test(timeout = 5000)
+    public void testEpollWaitWithTimeOutMinusOne() throws Exception  {
+        final EpollEventArray eventArray = new EpollEventArray(8);
+        try {
+            final FileDescriptor epoll = Native.newEpollCreate();
+            final FileDescriptor timerFd = Native.newTimerFd();
+            final FileDescriptor eventfd = Native.newEventFd();
```
carl-mastrangelo: fd Fd inconsistency, fyi

Issue #8429, opened 2018-10-24
ChannelOption for FIRE_EXCEPTION_ON_FAILURE?


Bennett-Lynch:
I want to be able to react to *any* outbound channel failure.

I can add a handler to the tail of my pipeline that will call `promise.addListener(ChannelFutureListener.FIRE_EXCEPTION_ON_FAILURE)` for every outbound event, but this is not sufficient.

Specifically, in the case of a `MessageToMessageEncoder` that outputs more than one object, new promises are created for all but the last object:
https://github.com/netty/netty/blob/00afb19d7a37de21b35ce4f6cb3fa7f74809f2ab/codec/src/main/java/io/netty/handler/codec/MessageToMessageEncoder.java#L112-L127

If one of those objects then triggers a failure on a subsequent outbound handler, the exception is suppressed.  There might be other scenarios where this can happen as well.

How can I react to these failures when I don't have the opportunity to inject a `ChannelFutureListener.FIRE_EXCEPTION_ON_FAILURE`?

Would it be reasonable to have a `ChannelOption` for `FIRE_EXCEPTION_ON_FAILURE`?

### Netty version

4.1.30.Final

normanmaurer:
@Bennett-Lynch that's a good question.. I like the idea of the `ChannelOption` but I would need to investigate if this is easily doable without breaking any API (I suspect it should be doable).

normanmaurer:
@Bennett-Lynch https://github.com/netty/netty/pull/8448 PTAL

Bennett-Lynch:
@normanmaurer I would like to re-open this issue, as I believe that the current implementation of making users manually inject `ChannelFutureListener.FIRE_EXCEPTION_ON_FAILURE` to writes is still cumbersome, and problematic in more complex pipelines that might have writes triggered by multiple sources.

The `MessageToMessageEncoder` fix was helpful, but I am still encountering use cases where I would like to be able to default this behavior to ON for an entire pipeline.  For example, if I have a handler that detects some special case and writes a response object to a pipeline, I have to ensure that it writes to the `ChannelHandlerContext`, and not the `Channel`, in order to pass through my `FIRE_EXCEPTION_ON_FAILURE` injecting-handler.  This introduces some unnecessary coupling and is just generally awkward.

I want to be able to know about all outbound failures (even if they're my own programming fault), in a simple and transparent manner.  I'm okay paying a small performance cost for it, if it's opt-in.

[...]

PR #8448, opened 2018-10-31
Introduce ChannelOption for FIRE_EXCEPTION_ON_FAILURE

normanmaurer:
Motivation:

Sometimes users just want to have an easy way to handle errors without the need to manually add a ChannelFutureListener to each ChannelFuture / ChannelPromise (just like in netty 3.x). This is not easy to do at the moment in netty as handlers in between may create new ChannelPromises etc.

Modifications:

Introdue ChannelOption.FIRE_EXCEPTION_ON_FAILURE which will (when true) have the effect of automatically calling ChannelPipeline.fireExceptionCaught(...) when a ChannelPromise is failed.

Result:

Fixes https://github.com/netty/netty/issues/8429.

### DIFF
--- a/transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
+++ b/transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
@@ -878,12 +878,13 @@ abstract class AbstractChannelHandlerContext extends DefaultAttributeMap
 
     @Override
     public ChannelPromise newPromise() {
-        return new DefaultChannelPromise(channel(), executor());
+        return DefaultChannelPipeline.fireExceptionOndFailure(new DefaultChannelPromise(channel(), executor()));
     }
 
     @Override
     public ChannelProgressivePromise newProgressivePromise() {
-        return new DefaultChannelProgressivePromise(channel(), executor());
+        return DefaultChannelPipeline.fireExceptionOndFailure(
+                new DefaultChannelProgressivePromise(channel(), executor()));
     }
 
     @Override
@@ -897,7 +898,7 @@ abstract class AbstractChannelHandlerContext extends DefaultAttributeMap
 
     @Override
     public ChannelFuture newFailedFuture(Throwable cause) {
-        return new FailedChannelFuture(channel(), executor(), cause);
+        return DefaultChannelPipeline.fireExceptionOndFailure(new FailedChannelFuture(channel(), executor(), cause));
     }
 
     private boolean isNotValidPromise(ChannelPromise promise, boolean allowVoidPromise) {
--- a/transport/src/main/java/io/netty/channel/ChannelOption.java
+++ b/transport/src/main/java/io/netty/channel/ChannelOption.java
@@ -130,6 +130,12 @@ public class ChannelOption<T> extends AbstractConstant<ChannelOption<T>> {
             valueOf("SINGLE_EVENTEXECUTOR_PER_GROUP");
 
     /**
+     * If {@code true} {@link ChannelPipeline#fireExceptionCaught(Throwable)} is called whenever a {@link ChannelFuture}
+     * is failed.
+     */
+    public static final ChannelOption<Boolean> FIRE_EXCEPTION_ON_FAILURE = valueOf("FIRE_EXCEPTION_ON_FAILURE");
+
+    /**
      * Creates a new {@link ChannelOption} with the specified unique {@code name}.
      */
     private ChannelOption(int id, String name) {
--- a/transport/src/main/java/io/netty/channel/DefaultChannelConfig.java
+++ b/transport/src/main/java/io/netty/channel/DefaultChannelConfig.java
@@ -23,18 +23,7 @@ import java.util.Map.Entry;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 
-import static io.netty.channel.ChannelOption.ALLOCATOR;
-import static io.netty.channel.ChannelOption.AUTO_CLOSE;
-import static io.netty.channel.ChannelOption.AUTO_READ;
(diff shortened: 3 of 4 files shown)


normanmaurer:
@Bennett-Lynch PTAL if this is what you wanted.

normanmaurer:
Still need to add a unit test but wanted to see how others feel about it first.

trustin:
Sounds OK-ish but I guess it may break the behavior of out-of-the-box handler implementations. If we are going to add this feature, 1) we should not modify our handlers' exception handling code (e.g exceptionCaught) to support this option and 2) we should document clearly that this behavior is unsupported by our handler implementations.

normanmaurer:
@trustin why would it break out of the box handlers ?

trustin:
Because of double notifications? The handlers we provide currently may be OK but we can't guarantee that it will be true in the future. For example a handler may want to behave differently for outbound exceptions and inbound exceptions. Also, double notification (future failures and exceptionCaughts) can introduce additional checks to remove a duplicate notification, e.g. logging twice.

normanmaurer:
@trustin got it... Yeah that sounds fair to me. That said I am not strong on adding this. Just wanted to show it first to discuss.

bryce-anderson:
I'm a little skeptical of this change. The issue in #8429 where the encoder swallows exceptions is serious enough, though I think the right thing there is to use a `PromiseCombiner` like strategy (probably a singled hand-rolled ChannelFutureListener that forwards to the parent promise would be sufficient).

ejona86:
I'm skeptical of this change as well. The current state of the world already has issues with double notifications of most events, so that's not a big of a concern for me. I just feel this is a more fundamental change and it would increase the complexity of Handlers since they have to support with and without the option.

If this was part of a migration strategy so that in Netty 5 we would always fire exceptionCaught on error, I would look at it more favorably. But it's not clear that this is the model we would choose for Netty 5. I will note that one interesting consequence of this change is if a handler in the middle of the pipeline fails a promise all handlers would see the failure, even though only some of the handlers saw the write. (That isn't necessarily horrible, especially if the behavior is the connection is closed at this point, but it is a change that should not be made lightly.)

I do feel like the current error handling is haphazard, error prone, and lacking. In 99% of the cases if there is an error 1) the connection is gone or 2) the only valid response is to close the connection since the operation could have partially occurred. There are cases today where an operation fails because the message is invalid, or similar, but most of those cases are programmer error and aren't worth having special-cased clean-up logic (i.e., you should still close the connection). The most common case of I/O error causes _tons_ of duplicate notifications.

In my own application I want any exception anywhere to close the connection immediately (it's unclear whether it's worth the complexity of sending a GOAWAY), notify the handlers of the exception, and any further writes to include the original cause for the channel termination (so like ClosedChannelException with the cause set). That also means only the first exception needs to fire exceptionCaught. HTTP/2 makes this a bit more complicated because some failures only impact one stream, but I'd argue that most of these could be handled in the HTTP/2 protocol instead of the Netty Futures (e.g., fabricate response HEADERS or RST_STREAM; behaving like a proxy).

normanmaurer:
@ejona86 @bryce-anderson fair enough... Thats why I wanted to show it first :) @bryce-anderson would you be up to open a PR with the change you suggested ?

bryce-anderson:
@normanmaurer, I presume you're talking about submitting one? I suppose the answer is 'yes' either way.

As @ejona86 noted, for the vast majority of cases I doubt it makes a difference since if an intermediate write fails it's likely that all subsequent writes will fail, so perhaps I overstated in saying that the issue was 'serious enough' and maybe it's more 'can be serious under uncommon circumstances'. But that's not to say that it can't be better regardless.

normanmaurer:
Yes

Bennett-Lynch:
I'll be in the minority and say that I'm in favor of this change, though I do agree with the previously mentioned sentiments about Netty's current error handling leaving something to be desired.

I view the Netty pipeline as a more composition-based approach to typical call stacks.  In your classical servlet-type implementation, if anything goes wrong -- reading or writing -- an exception is propagated up until it is caught.  Unexpected exceptions typically mean you want to send an error message and/or close the connection.  There's nothing special about inbound vs. outbound in that world.

I come to Netty with the same expectation, that any unexpected failure will be exposed in a way so that I can react accordingly.  I find it very unintuitive that outbound failures are treated differently than inbound failures in this case.

I'm primarily in favor because:
1) It's an optional feature, with a default of no behavioral change
2) Manually adding `ChannelFutureListener.FIRE_EXCEPTION_ON_FAILURE` is awkward and cumbersome

Maybe I haven't been burned by double notifications as often as others here have.  Is there a concrete example of an out-of-the-box handler implementation that this would break today?  Maybe that would help sway my opinion.  Regardless, this option is largely a quality-of-life improvement for people interested in using `ChannelFutureListener.FIRE_EXCEPTION_ON_FAILURE` everywhere today.

ejona86:
@Bennett-Lynch, I think we're on the same side, but I'm disagreeing with this approach because

1. It doesn't follow the "call stack." When a failure would happen it would start propagating at the head of the pipeline
2. "Optional" isn't truly optional. Optional really means "everyone has to deal with it." The Netty project would need to "support" the option (both on and off) for all built-in Handlers. For your own handlers it is "invisible" which way they are expecting this option to be set, so you can't change this flag willy-nilly; you have to audit all your code. I'd only want this flag if there is a plan for removal (e.g., it always being enabled in Netty 5, because "everyone should use it").
3. Http2ConnectionHandler looks like it would be confused by this change, as it uses exceptions to route logic decisions on how to handle inbound failures. For example, during parsing if a stream for a frame doesn't exist, it will throw and then the exceptionCaught logic will issue a RST_STREAM. However, that's _not_ the correct behavior for such _outbound_ events. Basically, the same exceptions require different behavior depending on if they were inbound (the remote side probably made an error) vs outbound (the local side probably made an error).

So while I am an advocate for change, I'm not sure this is the right change.

normanmaurer:
Lets close this

PR #8454, opened 2018-11-01
Don't swallow intermediate write failures in MessageToMessageEncoder

### LINKED ISSUE #8429: ChannelOption for FIRE_EXCEPTION_ON_FAILURE?

bryce-anderson:
Motivation:

If the encoder needs to flush more than one outbound message it will
create a new ChannelPromise for all but the last write which will
swallow failures.

Modification:

Use a PromiseCombiner in the case of multiple messages and the parent
promise isn't the `VoidPromise`.

Result:

Intermediate failures are propagated to the original ChannelPromise.

Fixes #8429.

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/MessageToMessageEncoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/MessageToMessageEncoder.java
@@ -124,14 +124,14 @@ public abstract class MessageToMessageEncoder<I> extends ChannelOutboundHandlerA
         }
     }
 
-    private void writeVoidPromise(ChannelHandlerContext ctx, CodecOutputList out) {
+    private static void writeVoidPromise(ChannelHandlerContext ctx, CodecOutputList out) {
         final ChannelPromise voidPromise = ctx.voidPromise();
         for (int i = 0; i < out.size(); i++) {
             ctx.write(out.getUnsafe(i), voidPromise);
         }
     }
 
-    private void writePromiseCombiner(ChannelHandlerContext ctx, CodecOutputList out, ChannelPromise promise) {
+    private static void writePromiseCombiner(ChannelHandlerContext ctx, CodecOutputList out, ChannelPromise promise) {
         final PromiseCombiner combiner = new PromiseCombiner();
         for (int i = 0; i < out.size(); i++) {
             combiner.add(ctx.write(out.getUnsafe(i)));
--- a/codec/src/main/java/io/netty/handler/codec/MessageToMessageEncoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/MessageToMessageEncoder.java
@@ -22,6 +22,7 @@ import io.netty.channel.ChannelPipeline;
 import io.netty.channel.ChannelPromise;
 import io.netty.util.ReferenceCountUtil;
 import io.netty.util.ReferenceCounted;
+import io.netty.util.concurrent.PromiseCombiner;
 import io.netty.util.internal.StringUtil;
 import io.netty.util.internal.TypeParameterMatcher;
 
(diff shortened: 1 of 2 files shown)


bryce-anderson:
A much simpler alternative is to use the `VoidPromise` instead of a new, discarded, `ChannelPromise` for the intermediate writes. The behavior there would be to propagate the exceptions via the `exceptionCaught` pathway which is what the original issue actually wanted.

codec/src/test/java/io/netty/handler/codec/MessageToMessageEncoderTest.java
```diff
@@ -37,4 +42,36 @@ protected void encode(ChannelHandlerContext ctx, Object msg, List<Object> out) t
...
+                    firstWritten = true;
+                    promise.setFailure(firstWriteException);
+                }
+            }
+        };
+
+        EmbeddedChannel channel = new EmbeddedChannel(writeThrower, encoder);
+        Object msg = new Object();
+        ChannelFuture write = channel.writeAndFlush(msg);
+        assertEquals(firstWriteException, write.cause());
```
normanmaurer: Call channel.finish() and asset the return value

codec/src/main/java/io/netty/handler/codec/MessageToMessageEncoder.java
```diff
@@ -108,28 +109,36 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
+                        writePromiseCombiner(ctx, out, promise);
                     }
-                    ctx.write(out.getUnsafe(sizeMinusOne), promise);
                 }
                 out.recycle();
             }
         }
     }
 
+    private void writeVoidPromise(ChannelHandlerContext ctx, CodecOutputList out) {
```
normanmaurer: static

codec/src/main/java/io/netty/handler/codec/MessageToMessageEncoder.java
```diff
@@ -108,28 +109,36 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
+        final ChannelPromise voidPromise = ctx.voidPromise();
+        for (int i = 0; i < out.size(); i++) {
+            ctx.write(out.getUnsafe(i), voidPromise);
+        }
+    }
+
+    private void writePromiseCombiner(ChannelHandlerContext ctx, CodecOutputList out, ChannelPromise promise) {
```
normanmaurer: Static

normanmaurer:
@bryce-anderson this will not work in some cases as the VoidPromise will not allow to add listeners in any other handler

normanmaurer:
@netty-bot test this please

codec/src/test/java/io/netty/handler/codec/MessageToMessageEncoderTest.java
```diff
@@ -37,4 +42,37 @@ protected void encode(ChannelHandlerContext ctx, Object msg, List<Object> out) t
...
+        assertEquals(msg, channel.readOutbound());
```
normanmaurer: nit: you may want to use `assertSame`

normanmaurer:
@netty-bot test this please

codec/src/main/java/io/netty/handler/codec/MessageToMessageEncoder.java
```diff
@@ -108,28 +109,36 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
-                    ctx.write(out.getUnsafe(sizeMinusOne), promise);
                 }
                 out.recycle();
             }
         }
     }
 
+    private static void writeVoidPromise(ChannelHandlerContext ctx, CodecOutputList out) {
+        final ChannelPromise voidPromise = ctx.voidPromise();
+        for (int i = 0; i < out.size(); i++) {
```
johnou: why are we using out.size() here when the original code used sizeMinusOne?

johnou: oh right it loops the first bunch then writes the last..

normanmaurer: good catch! @bryce-anderson please fix and adjust test to catch this

normanmaurer:
@netty-bot test this please

normanmaurer:
@netty-bot test this please

normanmaurer:
@bryce-anderson thanks a lot!

PR #8467, opened 2018-11-06
 Replace ConcurrentHashMap at allLeaks with a thread-safe set

carryxyh:
Motivation:
allLeaks is to store the DefaultResourceLeak. When we actually use it, the key is DefaultResourceLeak, and the value is actually a meaningless value.
We only care about the keys of allLeaks and don't care about the values. So Set is more in line with this scenario.
Using Set as a container is more consistent with the definition of a container than Map.

Modification:

Replace allLeaks with set. Create a thread-safe set using 'Collections.newSetFromMap(new ConcurrentHashMap<DefaultResourceLeak<?>, Boolean>()).'

### DIFF
--- a/common/src/main/java/io/netty/util/ResourceLeakDetector.java
+++ b/common/src/main/java/io/netty/util/ResourceLeakDetector.java
@@ -367,9 +367,9 @@ public class ResourceLeakDetector<T> {
             assert referent != null;
 
             // Store the hash of the tracked object to later assert it in the close(...) method.
-            trackedHash = System.identityHashCode(referent);
             // It's important that we not store a reference to the referent as this would disallow it from
             // be collected via the WeakReference.
+            trackedHash = System.identityHashCode(referent);
             allLeaks.add(this);
             // Create a new Record so we always have the creation stacktrace included.
             headUpdater.set(this, new Record(Record.BOTTOM));
--- a/common/src/main/java/io/netty/util/ResourceLeakDetector.java
+++ b/common/src/main/java/io/netty/util/ResourceLeakDetector.java
@@ -367,9 +367,9 @@ public class ResourceLeakDetector<T> {
             assert referent != null;
 
             // Store the hash of the tracked object to later assert it in the close(...) method.
+            trackedHash = System.identityHashCode(referent);
             // It's important that we not store a reference to the referent as this would disallow it from
             // be collected via the WeakReference.
-            trackedHash = System.identityHashCode(referent);
             allLeaks.add(this);
             // Create a new Record so we always have the creation stacktrace included.
             headUpdater.set(this, new Record(Record.BOTTOM));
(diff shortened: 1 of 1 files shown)


common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -441,13 +443,13 @@ private void record0(Object hint) {
 
         boolean dispose() {
             clear();
-            return allLeaks.remove(this, LeakEntry.INSTANCE);
+            return allLeaks.remove(this);
         }
 
         @Override
         public boolean close() {
             // Use the ConcurrentMap remove method, which avoids allocating an iterator.
```
normanmaurer: @carryxyh please remove this comment.

common/src/main/java/io/netty/util/ResourceLeakDetector.java
```diff
@@ -353,22 +355,22 @@ protected void reportInstancesLeak(String resourceType) {
...
                 Object referent,
                 ReferenceQueue<Object> refQueue,
-                ConcurrentMap<DefaultResourceLeak<?>, LeakEntry> allLeaks) {
+                Set<DefaultResourceLeak<?>> allLeaks) {
             super(referent, refQueue);
 
             assert referent != null;
 
             // Store the hash of the tracked object to later assert it in the close(...) method.
+            trackedHash = System.identityHashCode(referent);
```
normanmaurer: please move this where it was before as I think the rest of the comments also belongs to it. It explains why we can not just store referent itself.

normanmaurer:
@carryxyh can you please sign our ICLA:

https://netty.io/s/icla

carryxyh:
@normanmaurer 
I have fix the doc and sign the ICLA.
Thx for your review.

carryxyh:
U mean that 
```
// Store the hash of the tracked object to later assert it in the close(...) method.
trackedHash = System.identityHashCode(referent);
// It's important that we not store a reference to the referent as this would disallow it from
// be collected via the WeakReference.
allLeaks.add(this);
```
is better?

normanmaurer:
@carryxyh yes I mean that... Please just put the `trackedHash ...` under all comments as it was before.

carryxyh:
OK,
I misunderstood what you mean just now. :)

normanmaurer:
@netty-bot test this please

carryxyh:
/code/common/src/main/java/io/netty/util/ResourceLeakDetector.java:164: Line is longer than 120 characters (found 141).
I split this sentence into two lines to fix the above problem:
```
private final Set<DefaultResourceLeak<?>> allLeaks =
            Collections.newSetFromMap(new ConcurrentHashMap<DefaultResourceLeak<?>, Boolean>());
```

normanmaurer:
@carryxyh thanks!

PR #8468, opened 2018-11-06
Refresh DNS configuration each 5 minutes.

normanmaurer:
Motivation:

We should refresh the DNS configuration each 5 minutes to be able to detect changes done by the user. This is inline with what OpenJDK is doing

Modifications:

Refresh config every 5 minutes.

Result:

Be able to consume changes made by the user.

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProviders.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProviders.java
@@ -33,7 +33,7 @@ public final class DnsServerAddressStreamProviders {
     // https://msdn.microsoft.com/en-us/library/aa365968(VS.85).aspx.
     private static final DnsServerAddressStreamProvider DEFAULT_DNS_SERVER_ADDRESS_STREAM_PROVIDER =
             new DnsServerAddressStreamProvider() {
-        private volatile DnsServerAddressStreamProvider currentProvider;
+        private volatile DnsServerAddressStreamProvider currentProvider = provider();
         private final AtomicLong lastRefresh = new AtomicLong(System.nanoTime());
 
         @Override
@@ -41,6 +41,8 @@ public final class DnsServerAddressStreamProviders {
             long last = lastRefresh.get();
             DnsServerAddressStreamProvider current = currentProvider;
             if (System.nanoTime() - last > REFRESH_INTERVAL) {
+                // This is slightly racy which means it will be possible still use the old configuration for a small
+                // amount of time, but that's ok.
                 if (lastRefresh.compareAndSet(last, System.nanoTime())) {
                     current = currentProvider = provider();
                 }
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProviders.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProviders.java
@@ -18,18 +18,43 @@ package io.netty.resolver.dns;
 import io.netty.util.internal.PlatformDependent;
 import io.netty.util.internal.UnstableApi;
 
+import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProviders.java
```diff
@@ -18,18 +18,43 @@
...
     // https://msdn.microsoft.com/en-us/library/aa365968(VS.85).aspx.
     private static final DnsServerAddressStreamProvider DEFAULT_DNS_SERVER_ADDRESS_STREAM_PROVIDER =
+            new DnsServerAddressStreamProvider() {
+        private volatile DnsServerAddressStreamProvider currentProvider;
+        private final AtomicLong lastRefresh = new AtomicLong(System.nanoTime());
+
+        @Override
+        public DnsServerAddressStream nameServerAddressStream(String hostname) {
+            long last = lastRefresh.get();
+            DnsServerAddressStreamProvider current = currentProvider;
```
ejona86: This will initially be null, and a refresh won't occur.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProviders.java
```diff
@@ -18,18 +18,43 @@
...
+            if (System.nanoTime() - last > REFRESH_INTERVAL) {
+                if (lastRefresh.compareAndSet(last, System.nanoTime())) {
```
ejona86: This is racy. That's probably fine in the normal case, but it would at least deserve a comment that we may end up using the old configuration.

Note that the race means it can't be relied on for initializing `currentProvider` to be non-null.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProviders.java
normanmaurer: ups... yeah we should ensure this is never null 👍

resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProviders.java
normanmaurer: yeah I know its a racy but this is fine as it should only be for a small amount of time.

normanmaurer:
@ejona86 addressed

resolver-dns/src/main/java/io/netty/resolver/dns/DnsServerAddressStreamProviders.java
```diff
@@ -18,18 +18,45 @@
...
     private static final DnsServerAddressStreamProvider DEFAULT_DNS_SERVER_ADDRESS_STREAM_PROVIDER =
+            new DnsServerAddressStreamProvider() {
+        private volatile DnsServerAddressStreamProvider currentProvider = provider();
+        private final AtomicLong lastRefresh = new AtomicLong(System.nanoTime());
+
+        @Override
+        public DnsServerAddressStream nameServerAddressStream(String hostname) {
+            long last = lastRefresh.get();
+            DnsServerAddressStreamProvider current = currentProvider;
+            if (System.nanoTime() - last > REFRESH_INTERVAL) {
```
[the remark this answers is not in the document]
normanmaurer: @carl-mastrangelo not sure if I get it... How would this help ?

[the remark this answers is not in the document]
carl-mastrangelo: Hmm, I guess you're right.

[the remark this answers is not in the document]
normanmaurer: @carl-mastrangelo so are we ok with this change as it is ?

PR #8470, opened 2018-11-06
Combine flushes in DnsNameResolver to allow usage of sendmmsg to redu…

normanmaurer:
…ce syscall costs.

Motivation:

Some of transports support gathering writes when using datagrams. For example this is the case for EpollDatagramChannel. We should minimize the calls to flush() to allow making efficient usage of sendmmsg in this case.

Modifications:

- minimize flush() operations by supress these when when we are currently reading from the Channel
- doResolve(...) will typically do multiple queries so supress flush() until we wrote all queries to the Channel

Result:

Be able to use sendmmsg internally in the DnsNameResolver.

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
@@ -47,6 +47,7 @@ import io.netty.resolver.InetNameResolver;
 import io.netty.resolver.ResolvedAddressTypes;
 import io.netty.util.NetUtil;
 import io.netty.util.ReferenceCountUtil;
+import io.netty.util.concurrent.EventExecutor;
 import io.netty.util.concurrent.FastThreadLocal;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.FutureListener;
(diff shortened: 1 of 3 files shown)


resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -962,14 +966,45 @@ static boolean doResolveAllCached(String hostname,
...
                                       DnsRecord[] additionals,
                                       Promise<List<InetAddress>> promise,
                                       DnsCache resolveCache) {
-        final DnsServerAddressStream nameServerAddrs =
-                dnsServerAddressStreamProvider.nameServerAddressStream(hostname);
-        new DnsAddressResolveContext(this, hostname, additionals, nameServerAddrs,
-                                     resolveCache, authoritativeDnsServerCache).resolve(promise);
+
+        try {
+            // As the resolve(...) method may trigger multiple flush() operations lets just suppress these until we
```
ejona86: Where are these flushes being suppressed coming from? Why can't the be removed?

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -962,14 +966,45 @@ static boolean doResolveAllCached(String hostname,
...
+            // are done with the resolve(...) operation and then flush all together to reduce syscalls.
+            handler.supressFlushes();
```
ejona86: Is reentrancy a problem? It looks like it would be possible.

resolver-dns/src/main/java/io/netty/resolver/dns/DnsNameResolver.java
```diff
@@ -1100,16 +1135,22 @@ final DnsServerAddressStream newNameServerAddressStream(String hostname) {
         return dnsServerAddressStreamProvider.nameServerAddressStream(hostname);
     }
 
-    private final class DnsResponseHandler extends ChannelInboundHandlerAdapter {
+    private final class DnsResolverHandler extends ChannelDuplexHandler {
 
         private final Promise<Channel> channelActivePromise;
+        private ChannelHandlerContext ctx;
+        private boolean supressFlushes;
```
johnou: supress => suppress

normanmaurer:
@ejona86 @carl-mastrangelo @johnou PTAL again... I reworked this one to be more clear...

normanmaurer:
@netty-bot test this please.

carl-mastrangelo:
@normanmaurer FYI, it would be helpful for code review if you squashed just before merging the PR.  GitHub's UI can show changes since last review, but not if the commits are squashed.

normanmaurer:
Ah yeah sorry I completely rewrote it so I thought it makes more sense to force push here

normanmaurer:
@ejona86 does this look better now ?

normanmaurer:
@ejona86 yep.

PR #8471, opened 2018-11-06
Defer HTTP/2 stream transition state on initial write until headers are written

### LINKED ISSUE #8434: HTTP/2 MultiplexCodec sends a RST_STREAM if the HEADERS fail to write due to HPACK issue
### Expected behavior

If the stream fails to write the headers due to a HPACK issue such as being larger than the MAX_HEADER_LIST_SIZE it should not send a RST_STREAM frame.

### Actual behavior

Sends a RST_STREAM frame.

There is a test for this sort of behavior [here](https://github.com/netty/netty/blob/1cc692dd7de54bfffa05813a90b14670b4eb50d1/codec-http2/src/test/java/io/netty/handler/codec/http2/Http2MultiplexCodecTest.java#L312) but it doesn't consider that the `Http2FrameCodec` will create a new stream [here](https://github.com/netty/netty/blob/1cc692dd7de54bfffa05813a90b14670b4eb50d1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java#L352) which is circumvented in the tests via the `Writer` interface.

The reset is sent [here](https://github.com/netty/netty/blob/1cc692dd7de54bfffa05813a90b14670b4eb50d1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java#L892).

Without having thought about it a ton, I _think_ it should be generally safe to not send a RST_STREAM if the first write fails. The first message must always be HEADERS of some form (either regular or a PUSH_PROMISE) and if it gets past validation to where it's being written to the socket at all it should write in its entirety unless there is a socket problem in which case sending a RST_STREAM is no longer meaningful.

### Netty version

4.1 (SHA 1cc692dd7de, Oct 24, 2018)

bryce-anderson:
Motivation:
When the `DefaultHttp2ConnectionEncoder` writes the initial headers for a new
locally created stream we create the stream in the half-closed state if the
end-stream flag is set which signals to the life cycle manager that the headers
have been sent. However, if we synchronously fail to write the headers the
life cycle manager then sends a RST_STREAM on our behalf which is a connection
level PROTOCOL_ERROR because the peer sees the stream in an IDLE state.

Modification:
Don't open the stream in the half-closed state if the end-stream flag is
set and let the life cycle manager take care of it.

Result:
Cleaner state management in the `DefaultHttp2ConnectionEncoder`.

Fixes #8434.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
@@ -164,9 +164,10 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
             if (stream == null) {
                 try {
                     // We don't create the stream in a `halfClosed` state because if this is an initial
-                    // HEADERS frame we don't want to connection state to transition until after the
-                    // headers have been encoded and placed in the outbound buffer. The `LifeCycleManager`
-                    // will take care of transitioning the state as appropriate.
+                    // HEADERS frame we don't want the connection state to signify that the HEADERS have
+                    // been sent until after they have been encoded and placed in the outbound buffer.
+                    // Therefore, we let the `LifeCycleManager` will take care of transitioning the state
+                    // as appropriate.
                     stream = connection.local().createStream(streamId, /*endOfStream*/ false);
                 } catch (Http2Exception cause) {
                     if (connection.remote().mayHaveCreatedStream(streamId)) {
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
@@ -163,7 +163,11 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
             Http2Stream stream = connection.stream(streamId);
             if (stream == null) {
                 try {
-                    stream = connection.local().createStream(streamId, endOfStream);
+                    // We don't create the stream in a `halfClosed` state because if this is an initial
+                    // HEADERS frame we don't want to connection state to transition until after the
+                    // headers have been encoded and placed in the outbound buffer. The `LifeCycleManager`
+                    // will take care of transitioning the state as appropriate.
+                    stream = connection.local().createStream(streamId, /*endOfStream*/ false);
                 } catch (Http2Exception cause) {
                     if (connection.remote().mayHaveCreatedStream(streamId)) {
                         promise.tryFailure(new IllegalStateException("Stream no longer exists: " + streamId, cause));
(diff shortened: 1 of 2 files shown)


bryce-anderson:
This should be considered a WIP as it needs a test or two, but wanted feedback before I write them.

codec-http2/src/test/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoderTest.java
```diff
@@ -725,11 +725,38 @@ public ChannelFuture answer(InvocationOnMock invocation) {
...
+                    return promise;
+                }
+            });
+
+        writeAllFlowControlledFrames();
         encoder.writeHeaders(ctx, STREAM_ID, EmptyHttp2Headers.INSTANCE, 0, true, promise);
 
         assertTrue(promise.isDone());
         assertFalse(promise.isSuccess());
+        assertFalse(stream(STREAM_ID).isHeadersSent());
```
bryce-anderson: This is the real nugget of the test: we want to make sure that if we fail to write the HEADERS our stream state is sane. If so the `Http2ConnectionHandler` (which implements the `Http2LifecycleManager` interface) can handle the derelict RST_STREAM frames correctly [here](https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java#L770).

normanmaurer:
@netty-bot test this please

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -163,7 +163,11 @@ public ChannelFuture writeHeaders(final ChannelHandlerContext ctx, final int str
             Http2Stream stream = connection.stream(streamId);
             if (stream == null) {
                 try {
-                    stream = connection.local().createStream(streamId, endOfStream);
+                    // We don't create the stream in a `halfClosed` state because if this is an initial
+                    // HEADERS frame we don't want to connection state to transition until after the
```
trustin: `we don't want to connnection state to transition` - Perhaps you meant `we don't want connection state to enter 'halfClosed' state`?

bryce-anderson: The comment was pretty poor in general. I attempted to clarify it, both in terms of wording and the intent, specifically that we don't want the `stream.isHeadersSent()` to be `true` until we've actually sent the headers.

normanmaurer:
@ejona86 @carl-mastrangelo any concerns ?

normanmaurer:
@netty-bot test this please

bryce-anderson:
Any update on this?

normanmaurer:
@ejona86 @carl-mastrangelo ping... any concerns ?

normanmaurer:
Let me pull this in ... @carl-mastrangelo @ejona86 please let me know if you had any concern once you had a chance to check

PR #8476, opened 2018-11-07
Nio|Epoll|KqueueEventLoop task execution might throw UnsupportedOpera…

### LINKED ISSUE #8475: NioEventLoop task execution might throw UnsupportedOperationException on shutdown
### Expected behavior
Execution instead throws `RejectedExecutionException`.

### Actual behavior
There is a racy `UnsupportedOperationException` instead because the task removal is delegated to `MpscChunkedArrayQueue` that does not support removal. This happens with `SingleThreadEventExecutor` that overrides the `newTaskQueue` to return an MPSC queue instead of the `LinkedBlockingQueue` returned by the base class such as `NioEventLoop`, `EpollEventLoop` and `KQueueEventLoop`.

### Steps to reproduce
Schedule task to an event loop concurrently to an event loop shutdown.

```
    @Test
    public void testTaskRemovalOnShutdownThrowsUnsupportedOperationException() throws Exception {
        NioEventLoopGroup group = new NioEventLoopGroup(1);
        final NioEventLoop loop = (NioEventLoop) group.next();
        final Runnable task = new Runnable() {
            @Override
            public void run() {
                try {
                    Thread.sleep(1);
                } catch (InterruptedException e) {
                    e.printStackTrace();
                }
            }
        };
        new Thread() {
            @Override
            public void run() {
                while (true) {
                    loop.execute(task);
                }
            }
        }.start();
        Thread.sleep(10);
        loop.shutdownNow();
        Thread.sleep(100000);
    }
```

this is a racy test, the event loop shutdown race against the the task producing thread, if you run it enough it will fail:

```
Exception in thread "Thread-0" java.lang.UnsupportedOperationException
	at org.jctools.queues.BaseMpscLinkedArrayQueue.iterator(BaseMpscLinkedArrayQueue.java:201)
	at java.util.AbstractCollection.remove(AbstractCollection.java:282)
	at io.netty.util.concurrent.SingleThreadEventExecutor.removeTask(SingleThreadEventExecutor.java:340)
	at io.netty.util.concurrent.SingleThreadEventExecutor.execute(SingleThreadEventExecutor.java:781)
	at io.netty.channel.nio.NioEventLoopTest$8.run(NioEventLoopTest.java:195)
```

### Netty version
At least 4.1.x

N/A

N/A

normanmaurer:
…tionException on shutdown

Motivation:

There is a racy UnsupportedOperationException instead because the task removal is delegated to MpscChunkedArrayQueue that does not support removal. This happens with SingleThreadEventExecutor that overrides the newTaskQueue to return an MPSC queue instead of the LinkedBlockingQueue returned by the base class such as NioEventLoop, EpollEventLoop and KQueueEventLoop.

Modifications:

- Catch the UnsupportedOperationException
- Add unit test.

Result:

Fix #8475

### DIFF
--- a/common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
+++ b/common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
@@ -778,8 +778,20 @@ public abstract class SingleThreadEventExecutor extends AbstractScheduledEventEx
         addTask(task);
         if (!inEventLoop) {
             startThread();
-            if (isShutdown() && removeTask(task)) {
-                reject();
+            if (isShutdown()) {
+                boolean reject = false;
+                try {
+                    if (removeTask(task)) {
+                        reject = true;
+                    }
+                } catch (UnsupportedOperationException e) {
+                    // The task queue does not support removal so the best thing we can do is to just move on and
+                    // hope we will be able to pick-up the task before its completely terminated.
+                    // In worst case we will log on termination.
+                }
+                if (reject) {
+                    reject();
+                }
             }
         }
 
--- a/transport/src/test/java/io/netty/channel/nio/NioEventLoopTest.java
+++ b/transport/src/test/java/io/netty/channel/nio/NioEventLoopTest.java
@@ -22,6 +22,7 @@ import io.netty.channel.EventLoopGroup;
 import io.netty.channel.socket.ServerSocketChannel;
 import io.netty.channel.socket.nio.NioServerSocketChannel;
 import io.netty.util.concurrent.Future;
+import org.hamcrest.core.IsInstanceOf;
 import org.junit.Test;
 
 import java.net.InetSocketAddress;
@@ -29,7 +30,9 @@ import java.nio.channels.SelectionKey;
 import java.nio.channels.Selector;
 import java.nio.channels.SocketChannel;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
 import static org.junit.Assert.*;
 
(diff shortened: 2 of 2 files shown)


normanmaurer:
@vietj PTAL

vietj:
looks good

johnou:
cc @davidmoten

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
```diff
@@ -170,6 +170,12 @@ protected void wakeup(boolean inEventLoop) {
         }
     }
 
+    @Override
```
davidmoten: Add a bit of javadoc about  the meaning of the returned boolean? (especially as we no longer delegate a call to the queue which could have effectively provided that information via its javadoc). Ditto the other `removeTask` methods.

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
```diff
@@ -775,12 +775,18 @@ public void execute(Runnable task) {
         }
 
         boolean inEventLoop = inEventLoop();
-        addTask(task);
         if (!inEventLoop) {
+            if (isShutdown()) {
```
davidmoten: Yes this makes things less racy. Why not use a CAS loop here to make even less racy (especially as we have a  volatile field updater available)?

davidmoten: I see the `shutdown` method is `@Deprecated`. Is that to remove it from the public API but it will still be called by something? A bit of javadoc about the deprecation would be great.

davidmoten:
Having looked at the `shutdown()` method I see it's a tad wasteful with its volatile reads (calling `isShutdown` and `isShuttingDown` and not reusing the read value of `state`). I could clean up if desired.

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
normanmaurer: @davidmoten sure..

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
normanmaurer: oh duh! you are right...

normanmaurer:
@davidmoten adjusted based on your feedback. 

@davidmoten if you like you can submit a pr for `shutdown` improvements, that said I am not overly concerned about performance for this method.

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollEventLoop.java
davidmoten: ah I meant what does true mean and what does false mean. Does true mean it was found on the queue and removed? Does false mean it is either not supported or not found on the queue?

davidmoten:
>@davidmoten if you like you can submit a pr for shutdown improvements, that said I am not overly concerned about performance for this method.

Yeah I can't think of a scenario where `shutdown` is called in a fast loop or experiences high contention so I'll leave it alone.

common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
```diff
@@ -778,8 +778,20 @@ public void execute(Runnable task) {
         addTask(task);
         if (!inEventLoop) {
             startThread();
-            if (isShutdown() && removeTask(task)) {
-                reject();
+            if (isShutdown()) {
```
ejona86: Why is this code even caring about `isShutdown()`? It looks like this block should just be deleted. `isShutdown()` is checked by `addTask()`/`offerTask()`. If we added the entry to the queue, then it seems we should commit ourselves to it; attempting to remove it (which is not guaranteed to be successful, since the task may have already run) seems weird.

If this `isShutdown()` check was atomic with `startThread()`, I could sort of understand the removal since we know we didn't start the thread, but as it is it just seems like useless effort.

PR #8481, opened 2018-11-08
Correctly convert supported signature algorithms when using BoringSSL

normanmaurer:
Motivation:

BoringSSL uses different naming schemes for the signature algorithms so we need to adjust the regex to also handle these.

Modifications:

- Adjust SignatureAlgorithmConverter to handle BoringSSL naming scheme
- Ensure we do not include duplicates
- Add unit tests.

Result:

Correctly convert boringssl signature algorithm names.


normanmaurer:
This depends somewhat on 
https://github.com/netty/netty-tcnative/pull/412

Issue #8477, opened 2018-11-07
SSLSession#getCipherSuite() returns SSL_NULL_WITH_NULL_NULL with TLS 1.3


rkapsi:
I have a `FutureListener` which I'm adding to `SslHandler#handshakeFuture()` with the purpose gathering statistics on TLS protocols and ciphers. Something like this:

```java
void log(SslHandler handler) {
  SSLEngine sslEngine = handler.engine();
    
  Future<Channel> handshakeFuture = handler.handshakeFuture();
  handshakeFuture.addListener(new MySslLogger(sslEngine));
}

class MySslLogger implements FutureListener<Channel> {
  private final SSLEngine sslEngine;

  public HandshakeListener(SSLEngine sslEngine) {
    this.sslEngine = sslEngine;
  }

  @Override
  public void operationComplete(Future<Channel> future) throws Exception {
    SSLSession sslSession = sslEngine.getSession();
    String protocol = sslSession.getProtocol();
    String cipher = sslSession.getCipherSuite();

    System.out.println("protocol=" + protocol + ", cipher=" + cipher);
  }
}
```

It appears the reported cipher suite as returned by `SSLSession#getCipherSuite()` remains `SSL_NULL_WITH_NULL_NULL` in conjunction with TLS 1.3. It works fine with TLS 1.2 and the handshake itself is fine otherwise.

### Netty version

4.1.31

### JVM version (e.g. `java -version`)

Tried OpenJDK 1.8.0_192 and 11.0.1

normanmaurer:
@rkapsi which netty-tcnative artifact ?

rkapsi:
@normanmaurer ah yes, a custom static build (Linux) from 2.0.19 (this [commit](https://github.com/netty/netty-tcnative/commit/9406285cbd942938a91239293dfb87fe0def4146) to be exact)

normanmaurer:
@rkapsi against openssl or BoringSSL ?

rkapsi:
openssl

rkapsi:
... just gave a spin with BoringSSL (built from same commit), appears to work as expected.

normanmaurer:
hmm interesting

normanmaurer:
@rkapsi what exact openssl version this is ?

rkapsi:
`OpenSSL 1.1.1  11 Sep 2018`

https://github.com/netty/netty-tcnative/blob/master/pom.xml#L72-L75

rkapsi:
[...]

PR #8485, opened 2018-11-09
Correctly convert between openssl / boringssl and java cipher names w…

normanmaurer:
…hen using TLSv1.3

Motivation:

We did not correctly convert between openssl / boringssl and java ciphers when using TLV1.3 which had different effects when either using openssl or boringssl.
 - When using openssl and TLSv1.3 we always returned SSL_NULL_WITH_NULL_NULL as cipher name
 - When using boringssl with TLSv1.3 we always returned an incorrect constructed cipher name which does not match what is defined by Java.

Modifications:

 - Add correct mappings in CipherSuiteConverter for both openssl and boringssl
 - Add unit tests for CipherSuiteConvert
 - Add unit in SSLEngine which checks that we do not return SSL_NULL_WITH_NULL_NULL ever and that server and client returns the same cipher name.

Result:

Fixes https://github.com/netty/netty/issues/8477.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
+++ b/handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
@@ -61,11 +61,6 @@ public final class OpenSsl {
     private static final boolean SUPPORTS_HOSTNAME_VALIDATION;
     private static final boolean USE_KEYMANAGER_FACTORY;
     private static final boolean SUPPORTS_OCSP;
-    private static final String TLSV13_CIPHERS = "TLS_AES_256_GCM_SHA384" + ':' +
-                                                 "TLS_CHACHA20_POLY1305_SHA256" + ':' +
-                                                 "TLS_AES_128_GCM_SHA256" + ':' +
-                                                 "TLS_AES_128_CCM_8_SHA256" + ':' +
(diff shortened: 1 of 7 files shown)


normanmaurer:
did adjust the tests a bit to even capture more stuff and verify.

normanmaurer:
Merging as @rkapsi verified the fix

handler/src/main/java/io/netty/handler/ssl/OpenSsl.java
```diff
@@ -149,8 +144,22 @@
...
+                        }
+                        if (tlsv13Ciphers.length() == 0) {
+                            tlsv13Supported = false;
+                        } else {
+                            tlsv13Ciphers.setLength(tlsv13Ciphers.length() - 1);
+                            SSLContext.setCipherSuite(sslCtx, tlsv13Ciphers.toString() , true);
+                            tlsv13Supported = true;
+                        }
+
                     } catch (Exception ignore) {
```
carl-mastrangelo: Considering there is much more logic in the Try block, I don't know it's still okay to ignore this.  Maybe log?

normanmaurer: Let me just move most of it out of the block.

PR #8487, opened 2018-11-09
Update to openjdk 12ea19

normanmaurer:
Motivation:

We should test against latest EA releases.

Modifications:

Update to openkdk 12ea19

Result:

Use latest openjdk 12 EA build on the CI.


normanmaurer:
Let me merge this now... Its just a java version bump.

PR #8492, opened 2018-11-10
Enable netty-tcnative shading test again

normanmaurer:
Motivation:

We disabled the test at some point but it should work now without any problems.

Modifications:

Remove @Ignore from test.

Result:

Verify shading of netty-tcnative on CI.


normanmaurer:
Let me merge this as it just removes an `@Ignore`

Issue #8493, opened 2018-11-11
Handshake timeout may never be scheduled if handshake starts via a flush  


normanmaurer:
Reported by @tbrooks8:

A client (like my case) initiates the handshake in a flush call. The client hello message is generated and flushed. The SSLEngine now returns a handshake status of NEED_UNWRAP.

The channelActive call happens, since the handshake status is not HandshakeStatus.NOT_HANDSHAKING (line 1740), the method returns immediately without scheduling a timeout. There is now no timeout for this handshake.

The flush route for starting a handshake does not called startHandshakeProcessing. That is only called when the handler is added and the channel is already active or in the channelActive call. The flush route allows for the SSLEngine handshake to start prior to these things happening. And then the status of the engine, causes the handshake(final Promise<Channel> newHandshakePromise) method to exit early.

This was reported as part of https://github.com/netty/netty/pull/8486

normanmaurer:
The same is also true when `starttls` is used.

PR #8494, opened 2018-11-11
Handshake timeout may never be scheduled if handshake starts via a fl…

normanmaurer:
…ush or starttls is used.

Motivation:

We did not correctly schedule the handshake timeout if the handshake was either started by a flush(...) or if starttls was used.

Modifications:

- Correctly setup timeout in all cases
- Add unit tests.

Result:

Fixes https://github.com/netty/netty/issues/8493.


normanmaurer:
@tbrooks8 FYI

normanmaurer:
@tbrooks8 yes please 👍

PR #8500, opened 2018-11-13
Unlogged throwable (fix of the issue #8499)

### LINKED ISSUE #8499: AbstractChannel swallows potential errors by ignoring throwable
### Expected behavior

AbstractChannel::localAddress() & AbstractChannel::remoteAddress() should not ignore and silently skip errors happened during their invocation.

### Actual behavior

AbstractChannel::localAddress() & AbstractChannel::remoteAddress() both contain such a block of code:

`
            } catch (Throwable t) {
                // Sometimes fails on a closed socket in Windows.
                return null;
            }
`
I.e. they simply ignore all possible errors, including `VirtualMachineError`s.

Andremoniy:
Unlogged throwable.

Motivation:

Besides an error caused by closing socket in Windows a bunch of other errors may happen at this place which won't be somehow logged. For instance any `VirtualMachineError` as `OutOfMemoryError` will be simply ignored. The library should at least log the problem.

Modification:

Added logging of the throwable object.

Result:

Fixes #8499.

### DIFF
--- a/transport/src/main/java/io/netty/channel/AbstractChannel.java
+++ b/transport/src/main/java/io/netty/channel/AbstractChannel.java
@@ -172,6 +172,7 @@ public abstract class AbstractChannel extends DefaultAttributeMap implements Cha
                 this.localAddress = localAddress = unsafe().localAddress();
             } catch (Throwable t) {
                 // Sometimes fails on a closed socket in Windows.
+                logger.error(t.getMessage(), t);
                 return null;
             }
         }
@@ -194,6 +195,7 @@ public abstract class AbstractChannel extends DefaultAttributeMap implements Cha
                 this.remoteAddress = remoteAddress = unsafe().remoteAddress();
             } catch (Throwable t) {
                 // Sometimes fails on a closed socket in Windows.
+                logger.error(t.getMessage(), t);
                 return null;
             }
         }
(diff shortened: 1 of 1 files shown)


normanmaurer:
@Andremoniy should we maybe just special handle `Error` ?

Andremoniy:
IMHO that no general exceptions should be ignored, but only very specific. I don't know which error happens for the Windows case. That's why I would propose to leave logging for all errors.

However I am not pretending to know this case better then you; taking a choice between two alternatives: not having errors logged at all, and having logged only the `Error` subsclass - I would certainly take the second opportunity :)

So could you tell me please, should I amend my PR and add a catch clause for the `Error` class?

Sincerely,
Andrey

normanmaurer:
@trustin thoughts ?

normanmaurer:
@Andremoniy yes please change to log `Error` and use a log message like "Error during ...."

Andremoniy:
@normanmaurer okay, and re-throw it or `return null` as well?

normanmaurer:
I think rethrow is the right thing to do... So basically just:

```
             try {
                 this.localAddress = localAddress = unsafe().localAddress();
             } catch (Error e) {
                  throw (Error) e;
             } catch (Throwable t) {
                 // Sometimes fails on a closed socket in Windows.
                 return null;
```

Andremoniy:
@normanmaurer please check my recent commit

normanmaurer:
@netty-bot test this please

normanmaurer:
@Andremoniy this looks good.. Can you please sign our ICLA so I can pull this in ?

https://netty.io/s/icla 

Let me know once done please.

Andremoniy:
@normanmaurer Thank you!I've sent you an email. Perhaps I will sign the ICLA only after the next Monday, I am sorry for the delay.

Andremoniy:
@normanmaurer I've just signed the ICLA, confirmed.

normanmaurer:
@Andremoniy merged.. thanks!

PR #8502, opened 2018-11-13
Fix of the issue #8501: In case of connection issue channel.remoteAdd…

### LINKED ISSUE #8501: In case of connection issue channel.remoteAddress() returns null
### Expected behavior

It should return the request address, which is correctly passed into the `AnnotatedConnectionException`:

`
        protected final Throwable annotateConnectException(Throwable cause, SocketAddress remoteAddress) {
...
}
`

### Actual behavior

It returns `null` when trying to get it from the channel on receiving an exception.

### Netty version

4.x.x.

Andremoniy:
In case of connection issue channel.remoteAddress() returns null

Motivation:

In case of connection issue channel.remoteAddress() returns null.  Would be nice to have it at least for AbstractNioChannel, which stores this remoteAddress in the `requestedRemoteAddress` field before trying to connect.

Modification:

If the `remoteAddress` calculated by the super method is null, then return the `requestedRemoteAddress`.

Result:

Fixes #8501 .

### DIFF
--- a/transport/src/main/java/io/netty/channel/nio/AbstractNioChannel.java
+++ b/transport/src/main/java/io/netty/channel/nio/AbstractNioChannel.java
@@ -515,4 +515,13 @@ public abstract class AbstractNioChannel extends AbstractChannel {
             connectTimeoutFuture = null;
         }
     }
+
+    @Override
+    public SocketAddress remoteAddress() {
+        final SocketAddress remoteAddress = super.remoteAddress();
+        if (remoteAddress == null) {
+            return requestedRemoteAddress;
+        }
+        return remoteAddress;
+    }
 }


normanmaurer:
@Andremoniy hmm I wonder if this is correct as `remoteAddress()` usually should return `null` until it is connected to be consistent.

Andremoniy:
> @Andremoniy hmm I wonder if this is correct as `remoteAddress()` usually should return `null` until it is connected to be consistent.

Indeed, this is what is stated in the method spec:

>  {@code null} if this channel is not connected.

Can we consider it as an improvement? (So, perhaps, the spec should be changed as well).

normanmaurer:
@Andremoniy I dont think so and again I think this would be inconsistent with other transports so I think we should not do this. Also changing the api doc would mean all other transports would need to be fixed as well. Beside this some people may depend on `null` here to detect if a Channel is connected etc. So with this in mind I think we should not do it.

Andremoniy:
Okay. Anyway, thanks for the fast reply!

PR #8505, opened 2018-11-13
Proof of concept for #8504 (Expose requestedRemoteAddress in the Channel interface)

### LINKED ISSUE #8504: Expose requestedRemoteAddress in the Channel interface (improvement)
### Expected behavior

Most part of Channel abstract implementations contain `requestedRemoteAddress` object variable, which might be useful for debug purposes. When channel fails to connect in certain circumstances we lose information about what has been requested connect to.

### Actual behavior

`Channel` interface should expose `requestedRemoteAddress()` getter method

### Steps to reproduce

n/a

Andremoniy:
Expose requestedRemoteAddress in the Channel interface (improvement) #8504

Motivation:

Expose requestedRemoteAddress in the Channel interface (improvement) #8504

Modification:

Fixes #8504 .

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
@@ -460,6 +460,8 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
         // We start with the writability of the channel when creating the StreamChannel.
         private volatile boolean writable;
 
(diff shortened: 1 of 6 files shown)


normanmaurer:
@Andremoniy first of thanks for your effort. That said we can not do it (also I think it is not a good idea as it is very specific to client side) as we can not just add methods to the Channel interface without breaking people.

What we may could do (not sure if it is a good idea yet) is to include the remoteAddress in the exception so you can get it when you cast it.

PR #8555, opened 2018-11-13
Cleanup SslHandler handshake/renegotiation

tbrooks8:
Motivation:

The code for initiating a TLS handshake or renegotiation process is
currently difficult to reason about.

Modifications:

This commit introduces to clear paths for starting a handshake. The
first path is a normal handshake. The handshake is started and a timeout
is scheduled.

The second path is renegotiation. If the first handshake is incomplete,
the renegotiation promise is added as a listener to the handshake
promise. Otherwise, the renegotiation promise replaces the original
promsie. At that point the handshake is started again and a timeout is
scheduled.

Result:

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -1670,10 +1670,9 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
                 // Begin the initial handshake.
                 // channelActive() event has been fired already, which means this.channelActive() will
                 // not be invoked. We have to initialize here instead.
-                handshake(null, true);
-            } else {
-                applyHandshakeTimeout(null);
+                handshake();
             }
+            applyHandshakeTimeout();
         }
     }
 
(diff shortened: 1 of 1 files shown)


normanmaurer:
@netty-bot test this please

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1707,50 +1706,43 @@ private void startHandshakeProcessing() {
...
+        renegotiateOnEventLoop(promise);
         return promise;
     }
 
+    private void renegotiateOnEventLoop(final Promise<Channel> newHandshakePromise) {
+        final Promise<Channel> oldHandshakePromise = handshakePromise;
+        if (!oldHandshakePromise.isDone()) {
+            // There's no need to handshake because handshake is in progress already.
+            // Merge the new promise into the old one.
+            oldHandshakePromise.addListener(new FutureListener<Channel>() {
```
normanmaurer: I know this existed before but maybe just use `PromiseNotifier` ? Less code == better code :)

normanmaurer:
@netty-bot test this please

normanmaurer:
@tbrooks8 thanks a lot!

normanmaurer:
@tbrooks8 merged.. thanks again

PR #8556, opened 2018-11-14
repair Http header decode bug

fengcone:
if the header like this:  GET /ws?name=Bruce Lee HTTP/1.1
the decode result must be:
    initialLine[0]=GET
    initialLine[1]=/ws?name=Bruce
    initialLine[2]=Lee HTTP/1.1

Motivation:

Modification:

Result:

Fixes #<GitHub issue number>. 


normanmaurer:
@fengcone actually this fix is incorrect... Have a whitespace there is not allowed per RFC. This will need to be decoded as `%20`.

PR #8557, opened 2018-11-14
Add DatagramObjectEncoder and Decoder

Moehritz:
**Motivation**:

These two classes enable usage of ObjectEncoder and ObjectDecoder with Datagrams, while keeping sender/receiver information available.
Before a User had to do something similar to this:
```
@ChannelHandler.Sharable
public class DatagramObjectDecoder extends MessageToMessageDecoder<DatagramPacket> {

    private final ObjectDecoder decoder;
    private final Method decodeMethod;

    public DatagramObjectDecoder(ObjectDecoder decoder) {
        this.decoder = decoder;
        // Use reflection as the decode() methods are protected
        try {
            this.decodeMethod = decoder.getClass().getDeclaredMethod("decode", ChannelHandlerContext.class, ByteBuf.class);
            decodeMethod.setAccessible(true);
        } catch (NoSuchMethodException e) {
            throw new RuntimeException(e);
        }
    }

    @Override
    protected void decode(ChannelHandlerContext ctx, DatagramPacket packet, List<Object> out) throws InvocationTargetException, IllegalAccessException {
        out.add(new DefaultAddressedEnvelope<>(decodeMethod.invoke(decoder, ctx, packet.content()), packet.recipient(), packet.sender()));
    }
}
```
This solution is more elegant and user-friendly.

**Modifications**:

Added two classes DatagramObjectDecoder and DatagramObjectEncoder based on ObjectEncoder and ObjectDecoder, aswell as a test class

**Result**:
This commit does not change any existing file, only adds two files. It therefore changes nothing for people who do not use these new classes.

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/serialization/DatagramObjectDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/serialization/DatagramObjectDecoder.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.codec.serialization;
+
+import io.netty.channel.AddressedEnvelope;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelPipeline;
+import io.netty.channel.DefaultAddressedEnvelope;
+import io.netty.channel.socket.DatagramPacket;
+import io.netty.handler.codec.MessageToMessageDecoder;
+
+import java.net.InetSocketAddress;
+import java.util.List;
+
+/**
+ * A decoder that decodes the content of the received {@link DatagramPacket} using the specified {@link ObjectDecoder}
+ * decoder. It wraps the result in a new {@link AddressedEnvelope}<{@link DatagramPacket}, {@link InetSocketAddress}>.
+ * <p>
+ * This allows usage of the {@link ObjectDecoder} while still being able to access sender and receiver information of
+ * the packet. Used in conjuntion with {@link DatagramObjectEncoder}.
+ *
+ * <pre><code>
+ * {@link ChannelPipeline} pipeline = ...;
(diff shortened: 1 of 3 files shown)


normanmaurer:
First of thanks for you contribution. That said I am not convinced we should include this in netty itself. @carl-mastrangelo @ejona86 @trustin WDYT ?

codec/src/main/java/io/netty/handler/codec/serialization/DatagramObjectDecoder.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private final ObjectDecoder objectDecoder;
+
+    /**
+     * Creates a new datagram decoder using the specified objectDecoder
+     *
+     * @param objectDecoder the {@link ObjectDecoder} to be used
+     */
+    public DatagramObjectDecoder(ObjectDecoder objectDecoder) {
+        this.objectDecoder = objectDecoder;
```
normanmaurer: this is problematic as you not forward all methods to the decoder.

codec/src/main/java/io/netty/handler/codec/serialization/DatagramObjectEncoder.java
```diff
@@ -0,0 +1,66 @@
...
+        this(new ObjectEncoder());
+    }
+
+    /**
+     * Creates a new datagram encoder with a custom implementation of {@link ObjectEncoder}
+     *
+     * @param objectEncoder the {@link ObjectEncoder} to be used
+     */
+    public DatagramObjectEncoder(ObjectEncoder objectEncoder) {
+        this.objectEncoder = objectEncoder;
```
normanmaurer: same comment as above... You need to forward all methods to the encoder as well.

carl-mastrangelo:
@Moehritz Do I understand correctly that the main purpose of this class is to expose `decode()` ?

Moehritz:
@carl-mastrangelo sort of, yes. But more user-oriented speaking, the purpose is to make ObjectDecoder/Encoder useful for UDP aswell.

@normanmaurer thanks for having a look. I am not sure which methods to forward, as neither ObjectEncoder nor ObjectDecoder do overwrite anything other than encode()/decode(). Only methods I would forward are `handlerRemoved(ChannelHandlerContext ctx)`, `channelReadComplete(ChannelHandlerContext ctx)`, `channelInactive(ChannelHandlerContext ctx)` and `userEventTriggered(ChannelHandlerContext ctx, Object evt)`, as they have a special implementation in ByteToMessageDecoder (superclass of ObjectDecoder) compared to MessageToMessageDecoder (superclass of DatagramObjectDecoder).
Did I get this right? Or do you mean to do it like in DatagramPacketDecoder, for example?

normanmaurer:
@Moehritz yes you get this right... that said generally speaking all of this sounds very error-prone as the implementation of `ByteToMessageDecoder` etc could change . 

Like I said at the moment I am not convinced we should pull this in.

normanmaurer:
Let me close this as I think the approach has issues (as stated before) and I am also not convinced this is something we want to maintain .

PR #8558, opened 2018-11-15
Fix "Upgrade" header value case

slandelle:
Motivation:

According to [RFC6455](https://tools.ietf.org/html/rfc6455#section-4.1):

> The request MUST contain a |Connection| header field whose value MUST include the "Upgrade" token.

Sadly, `HttpHeaderValues#UPGRADE` is in minor case, which causes some WebSocket servers implementation to reject the upgrade request generated by `WebSocketClientHandshaker`.

Modification:

Fix `HttpHeaderValues#UPGRADE` case from `upgrade` to `Upgrade`.

Result:

`WebSocketClientHandshaker` sends upgrade requests that are compliant with RFC6455

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderValues.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpHeaderValues.java
@@ -203,7 +203,7 @@ public final class HttpHeaderValues {
     /**
      * {@code "upgrade"}
      */
-    public static final AsciiString UPGRADE = AsciiString.cached("upgrade");
+    public static final AsciiString UPGRADE = AsciiString.cached("Upgrade");
     /**
      * {@code "websocket"}
      */


normanmaurer:
@slandelle I think it is lower-case because of http2... @ejona86 @carl-mastrangelo @bryce-anderson I am right ?

normanmaurer:
@netty-bot test this please

slandelle:
@normanmaurer Some context: regression was reported against latest Gatling generation as we switched from AHC's own WebSocket upgrade mechanism to Netty's `WebSocketClientHandshaker`.

> I think it is lower-case because of http2...

We should have 2 different values then. I can do that, please just let me know.

slandelle:
Tests fail because they expect the minor case form. Please advice.

BTW, I'm also a bit scared by the constants in `HttpHeaderNames` that are now in minor case. I know that HTTP header names are case insensitive, but I wouldn't be surprised that some server implementations would actually require the came case that was a kind of standard prior to HTTP/2.

normanmaurer:
@slandelle that said I think the server must also accept "upgraded" (lower-case).

See https://tools.ietf.org/html/rfc6455#section-4.2.1.

```
4.   A |Connection| header field that includes the token "Upgrade",
        treated as an ASCII case-insensitive value.
```

normanmaurer:
@slandelle all the names are lower-case for ages and honestly so I am not so worried honestly.

slandelle:
> 4.   A |Connection| header field that includes the token "Upgrade",
        treated as an ASCII case-insensitive value.

@normanmaurer No, that's the "Upgrade" header name, not the "Upgrade" header value :)

```
        GET /chat HTTP/1.1
        Host: server.example.com
        Upgrade: websocket <== here, "Upgrade" is a header name, hence case insensitive
        Connection: Upgrade <== here "Upgrade" is a header value, hence case sensitive
        Sec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ==
        Origin: http://example.com
        Sec-WebSocket-Protocol: chat, superchat
        Sec-WebSocket-Version: 13
```

normanmaurer:
@slandelle I am confused... Does it not directly state that its the "token" of the Connection header field ( where token == value) ?

slandelle:
Ahh, you're absolutely right.

Closing, sorry for noise. I'll warn the user that he has to fix his server.

Thanks!

normanmaurer:
@slandelle no worries :)

Issue #8566, opened 2018-11-16
NioEventLoop selector crash and can't recover


qinliujie:
### Expected behavior

```
WARN [HSF-Worker-2-thread-5:t.h.remoting] [] [] [] Unexpected exception in the selector loop.
java.io.IOException: Broken pipe
	at sun.nio.ch.EPollArrayWrapper.interrupt(Native Method)
	at sun.nio.ch.EPollArrayWrapper.interrupt(EPollArrayWrapper.java:323)
	at sun.nio.ch.EPollSelectorImpl.wakeup(EPollSelectorImpl.java:207)
	at io.netty.channel.nio.SelectedSelectionKeySetSelector.wakeup(SelectedSelectionKeySetSelector.java:73)
	at io.netty.channel.nio.NioEventLoop.selectNow(NioEventLoop.java:722)
	at io.netty.channel.nio.NioEventLoop$1.get(NioEventLoop.java:71)
	at io.netty.channel.DefaultSelectStrategy.calculateStrategy(DefaultSelectStrategy.java:30)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:405)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:858)
	at com.taobao.hsf.io.netty.util.PooledThreadFactory$PooledByteBufRunnable.run(PooledThreadFactory.java:37)
	at java.lang.Thread.run(Thread.java:882)

```

NioEventLoop can recover from exception or rebuild  a new selector

### Actual behavior

NioEventLoop dead totally

### Steps to reproduce

Can't reproduce this case,but we can mock;

see below code

```
protected void run() {
        for (;;) {
            try {
                switch (selectStrategy.calculateStrategy(selectNowSupplier, hasTasks())) {
                    case SelectStrategy.CONTINUE:
                        continue;
                    case SelectStrategy.SELECT:
                    .....
            } catch (Throwable t) {
                handleLoopException(t);
            }
```

notice that,`(selectStrategy.calculateStrategy(selectNowSupplier, hasTasks())`  can throw an exception every loop;

now we look at `selectNowSupplier`,

```
    private final IntSupplier selectNowSupplier = new IntSupplier() {
        @Override
        public int get() throws Exception {
            return selectNow();
        }
    };
```

and I find `selector` api document

```
    /**
     * Selects a set of keys whose corresponding channels are ready for I/O
     * operations.
     *
[...]

PR #8569, opened 2018-11-16
Recover from Selector IOException

normanmaurer:
Motivation:

When the Selector throws an IOException during our EventLoop processing we should rebuild it and transfer the registered Channels. At the moment we will continue trying to use it which will never work.

Modifications:

- Rebuild Selector when an IOException is thrown during any select*(...) methods.
- Add unit test.

Result:

Fixes https://github.com/netty/netty/issues/8566.

### DIFF
--- a/transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
+++ b/transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
@@ -401,7 +401,8 @@ public final class NioEventLoop extends SingleThreadEventLoop {
     protected void run() {
         for (;;) {
             try {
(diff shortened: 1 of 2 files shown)


transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
```diff
@@ -401,7 +401,8 @@ private void rebuildSelector0() {
     protected void run() {
         for (;;) {
             try {
-                switch (selectStrategy.calculateStrategy(selectNowSupplier, hasTasks())) {
+                try {
+                    switch (selectStrategy.calculateStrategy(selectNowSupplier, hasTasks())) {
                     case SelectStrategy.CONTINUE:
```
johnou: fix indenting?

PR #8573, opened 2018-11-19
Fix test that assumed detection of peer supported algs is not support…

normanmaurer:
…ed in BoringSSL.

Motivation:

0d2e38d5d6db6ed1abdf901230fa524c0f8db14c added supported for detection of peer supported algorithms but we missed to fix the testcase.

Modifications:

Fix test-case.

Result:

No more failing tests with BoringSSL.


normanmaurer:
Let me just pull this in so the CI is happy again

PR #8577, opened 2018-11-20
Fix javadoc to correctly explain how ChannelDuplexHandler.deregister(…

normanmaurer:
…...) works.

Motivation:

We had an error in the javadoc which was most likely caused by copy and paste.

Modifications:

Fix javadoc.

Result:

Correct javadoc.


normanmaurer:
Let me just pull in as its only a typo fix.

PR #8578, opened 2018-11-20
Remove UDT transport for Netty 5.

normanmaurer:
Motivation:

The UDT transport was marked as @Deprecated a long time ago as the underlying native library is not really maintained anymore. We should remove it as part of Netty 5.

Modifications:

Remove UDT transport

Result:

Dont try to maintain a transport which uses an unmaintained native lib internally.


normanmaurer:
@netty/contributors please note that this will be done as part of netty 5 only

normanmaurer:
CI failure is related to not correctly setup CI (yet). Will look into it.

Issue #8510, opened 2018-11-13
[Netty 5] OIO Transport removal


normanmaurer:
This transport is unique because it uses Java's blocking IO (`java.io` / `java.net`) under the hood. However it is not clear if this transport is actually useful so it should be removed.

The same would be true for all transports that sub-class OIO.

oliviercoma:
OIO is actually useful for serial communication (various Netty channel implementations and they are really easy to implement using OIO).

There is currently no viable and production ready alternative using NIO due to lack of support of NIO from unrelying libraries (JSSC, jSerialCom, RXTX, ...). The only NIO implementation I found is the one from plc4x. However, this implementation is hacky as hell as it does not use system NIO (epoll, ...), but simulating it by implementing kind of fake Selector.

In the end, for serial communication, we would be better with simple OIO implementation than faking NIO.

normanmaurer:
This is out of scope for netty.

PR #8579, opened 2018-11-20
Mark OIO based transports as deprecated as preparation for removal in…

normanmaurer:
… Netty 5.

Motivation:

We plan to remove the OIO based transports in Netty 5 so we should mark these as deprecated already.

Modifications:

Mark all OIO based transports as deprecated.

Result:

Give the user a heads-up for removal.


normanmaurer:
@rkapsi FYI

normanmaurer:
This is the preparation for https://github.com/netty/netty/issues/8510.

PR #8580, opened 2018-11-20
Remove OIO transport (and transports that depend on it).

normanmaurer:
Motivation:

This transport is unique because it uses Java's blocking IO (java.io / java.net) under the hood. However it is not clear if this transport is actually useful so it should be removed.

Modifications:

- Remove OIO transport and RXTX transport which depend on it.
- Remove Oio*Sctp* implementations
- Remove PerThreadEventLoop* which was only used by OIO transport.

Result:

Fixes https://github.com/netty/netty/issues/8510.


normanmaurer:
This is the netty 5 counter-part of https://github.com/netty/netty/pull/8579

example/src/main/java/io/netty/example/http/websocketx/server/WebSocketFrameHandler.java
```diff
@@ -21,15 +21,15 @@
 import io.netty.channel.SimpleChannelInboundHandler;
 import io.netty.handler.codec.http.websocketx.TextWebSocketFrame;
 import io.netty.handler.codec.http.websocketx.WebSocketFrame;
-import org.slf4j.Logger;
```
ejona86: Why is this change mixed in?

normanmaurer: Because this dependency was pulled in transitively which then failed after removing the dependencies for the example module ;). 

That said let me maybe just do a pr against 4.1 that removes the usage, pull it in and then rebase to make it more clear.

normanmaurer: @ejona86 https://github.com/netty/netty/pull/8582

normanmaurer:
rebased...

PR #8582, opened 2018-11-20
Remove transitive dependency on slf4j in example

normanmaurer:
Motivation:

We currently depend on slf4j in an transitive way in one of our classes in the examples. We should not do this.

Modifications:

Replace slf4j logger usage with netty logger.

Result:

Remove not needed dependency.

### DIFF
--- a/example/src/main/java/io/netty/example/http/websocketx/server/WebSocketFrameHandler.java
+++ b/example/src/main/java/io/netty/example/http/websocketx/server/WebSocketFrameHandler.java
@@ -21,15 +21,11 @@ import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.SimpleChannelInboundHandler;
 import io.netty.handler.codec.http.websocketx.TextWebSocketFrame;
 import io.netty.handler.codec.http.websocketx.WebSocketFrame;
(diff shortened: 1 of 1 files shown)


normanmaurer:
This came up in https://github.com/netty/netty/pull/8580#discussion_r235148203

example/src/main/java/io/netty/example/http/websocketx/server/WebSocketFrameHandler.java
```diff
@@ -21,15 +21,15 @@
 import io.netty.channel.SimpleChannelInboundHandler;
 import io.netty.handler.codec.http.websocketx.TextWebSocketFrame;
 import io.netty.handler.codec.http.websocketx.WebSocketFrame;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import io.netty.util.internal.logging.InternalLogger;
```
ejona86: Do we really want an example to use an internal class? Should we just add the slf4j dependency to the example instead?

normanmaurer: @ejona86 good question... maybe we should just not log at all ?

Issue #8513, opened 2018-11-13
[Netty 5] EventLoop and Channel contract


normanmaurer:
At the moment it’s possible to have a `Channel` in Netty that is not registered / assigned to an `EventLoop` until `register(...)` is called. This is suboptimal as if the `Channel` is not registered it is also not possible to do anything useful with a `ChannelFuture` that belongs to the `Channel`. We should think about if we should have the `EventLoop` as a constructor argument of a `Channel` and have the `register` / `deregister` method only have the effect of add a `Channel` to KQueue/Epoll/... It is also currently possible to deregister a `Channel` from one `EventLoop` and register it with another `EventLoop`. This operation defeats the threading model assumptions that are wide spread in Netty, and requires careful user level coordination to pull off without any concurrency issues. It is not a commonly used feature in practice, may be better handled by other means (e.g. client side load balancing), and therefore we propose removing this feature. 

To fix this we think an `EventLoop` must be provided when a `Channel` is constructed. The `Channel` would be “tied” to the `EventLoop` `Thread` from its creation. The channel may have a method to control when the registration actually occurs, but the `EventLoop` will not have a `register(Channel)` method in the public API. 
Related to this we also want to remove support for re-registration of a `Channel`. The current APIs can’t guarantee correctness and orchestrating the re-registration requires detailed knowledge of Netty’s threading model to use this feature. Due to these complexities we decided to just remove the `re-registration`, but keep `deregister()`. 

This is also in-line with what was done in [SwiftNIO](https://swiftnio.io).

normanmaurer:
Starting to work on this as other tasks depend on this as.well.

PR #4589, opened 2015-12-17
Added when it's safe to call channel read

jroper:
### Motivation:

I was reading the javadoc for Channel.read() just the other day to work out when the best time to first invoke it was, and it didn't say.  Coincidentally, in #4571, I found out in another project I was invoking it at a time that was illegal.
### Modifications:

This adds javadocs to Channel.read() and ChannelHandlerContext.read() to explain when it's legaal to invoke it.
### Result:

Hopefully other people won't make the same mistake I did in future.

Issue #8514, opened 2018-11-13
[Netty 5] EventLoop(Group) / EventExecutor(Group) extensibility 


normanmaurer:
The `EventLoop` and `Channel` implementations are very tightly coupled, and the `EventLoop` also has some factory methods internally which don't facilitate wrapping. We should make sure the ?`EventLoop` and `EventExecutor` APIs support wrapping so users can maintain metrics and otherwise extend the capability of existing implementations. 
Extending event loops such as `EpollEventLoop` would allow applications to implement asynchronous architectures such as thread per core, by handling all events in the loop (file, network and application). In order to support latency SLAs, some applications may not be able to use the loops offered by Netty as they are. For example, they may need to handle events according to different priorities, or support back pressure, or have more timely wake up mechanisms and so forth. These requirements are very application specific, so they should not be concerns of Netty. However, allowing extending event loops such as `EpollEventLoop` would remove from applications low level concerns that are platform specific. 

We want to support this, but more investigation is needed as to “how”. Extensibility in general is a priority and the `EventLoop` in particular is currently lacking.

johnou:
This is particularly interesting to me, I have gone to a great deal of effort to lower overhead (writing only from the correct eventloop) with our server application, I wrote up a small explanation here https://github.com/netty/netty/issues/7020#issuecomment-388007409

normanmaurer:
This should be complete

PR #8587, opened 2018-11-21
Tighten contract between Channel and EventLoop by require the EventLo…

normanmaurer:
…op on Channel construction.

Motivation:

At the moment it’s possible to have a Channel in Netty that is not registered / assigned to an EventLoop until register(...) is called. This is suboptimal as if the Channel is not registered it is also not possible to do anything useful with a ChannelFuture that belongs to the Channel. We should think about if we should have the EventLoop as a constructor argument of a Channel and have the register / deregister method only have the effect of add a Channel to KQueue/Epoll/... It is also currently possible to deregister a Channel from one EventLoop and register it with another EventLoop. This operation defeats the threading model assumptions that are wide spread in Netty, and requires careful user level coordination to pull off without any concurrency issues. It is not a commonly used feature in practice, may be better handled by other means (e.g. client side load balancing), and therefore we propose removing this feature.

Modifications:

- Change all Channel implementations to require an EventLoop for construction ( + an EventLoopGroup for all ServerChannel implementations)
- Remove all register(...) methods from EventLoopGroup
- Add ChannelOutboundInvoker.register(...) which now basically means we want to register on the EventLoop for IO.
- Change ChannelUnsafe.register(...) to not take an EventLoop as parameter (as the EventLoop is supplied on custruction).
- Change ChannelFactory to take an EventLoop to create new Channels and introduce ServerChannelFactory which takes an EventLoop and one EventLoopGroup to create new ServerChannel instances.
- Add ServerChannel.childEventLoopGroup()
- Ensure all operations on the accepted Channel is done in the EventLoop of the Channel in ServerBootstrap
- Change unit tests for new behaviour

Result:

A Channel always has an EventLoop assigned which will never change during its life-time. This ensures we are always be able to call any operation on the Channel once constructed (unit the EventLoop is shutdown). This also simplifies the logic in DefaultChannelPipeline a lot as we can always call handlerAdded / handlerRemoved directly without the need to wait for register() to happen.

Also note that its still possible to deregister a Channel and register it again. It's just not possible anymore to move from one EventLoop to another (which was not really safe anyway).

Fixes https://github.com/netty/netty/issues/8513.

### DIFF
--- a/transport/src/main/java/io/netty/bootstrap/AbstractBootstrap.java
+++ b/transport/src/main/java/io/netty/bootstrap/AbstractBootstrap.java
@@ -22,14 +22,10 @@ import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelOption;
 import io.netty.channel.ChannelPromise;
-import io.netty.channel.DefaultChannelPromise;
 import io.netty.channel.EventLoop;
 import io.netty.channel.EventLoopGroup;
-import io.netty.channel.ReflectiveChannelFactory;
 import io.netty.util.internal.SocketUtils;
 import io.netty.util.AttributeKey;
-import io.netty.util.concurrent.EventExecutor;
-import io.netty.util.concurrent.GlobalEventExecutor;
 import io.netty.util.internal.StringUtil;
 import io.netty.util.internal.logging.InternalLogger;
 
@@ -47,11 +43,10 @@ import java.util.Map;
  * <p>When not used in a {@link ServerBootstrap} context, the {@link #bind()} methods are useful for connectionless
  * transports such as datagram (UDP).</p>
  */
-public abstract class AbstractBootstrap<B extends AbstractBootstrap<B, C>, C extends Channel> implements Cloneable {
+public abstract class AbstractBootstrap<B extends AbstractBootstrap<B, C, F>, C extends Channel, F>
+        implements Cloneable {
 
     volatile EventLoopGroup group;
-    @SuppressWarnings("deprecation")
-    private volatile ChannelFactory<? extends C> channelFactory;
     private volatile SocketAddress localAddress;
     private final Map<ChannelOption<?>, Object> options = new LinkedHashMap<ChannelOption<?>, Object>();
     private final Map<AttributeKey<?>, Object> attrs = new LinkedHashMap<AttributeKey<?>, Object>();
@@ -61,9 +56,8 @@ public abstract class AbstractBootstrap<B extends AbstractBootstrap<B, C>, C ext
         // Disallow extending from a different package.
     }
 
-    AbstractBootstrap(AbstractBootstrap<B, C> bootstrap) {
+    AbstractBootstrap(AbstractBootstrap<B, C, F> bootstrap) {
         group = bootstrap.group;
-        channelFactory = bootstrap.channelFactory;
         handler = bootstrap.handler;
         localAddress = bootstrap.localAddress;
         synchronized (bootstrap.options) {
@@ -92,46 +86,6 @@ public abstract class AbstractBootstrap<B extends AbstractBootstrap<B, C>, C ext
     @SuppressWarnings("unchecked")
     private B self() {
         return (B) this;
-    }
-
-    /**
-     * The {@link Class} which is used to create {@link Channel} instances from.
-     * You either use this or {@link #channelFactory(io.netty.channel.ChannelFactory)} if your
-     * {@link Channel} implementation has no no-args constructor.
-     */
-    public B channel(Class<? extends C> channelClass) {
-        if (channelClass == null) {
-            throw new NullPointerException("channelClass");
-        }
-        return channelFactory(new ReflectiveChannelFactory<C>(channelClass));
-    }
-
-    /**
-     * @deprecated Use {@link #channelFactory(io.netty.channel.ChannelFactory)} instead.
-     */
-    @Deprecated
-    public B channelFactory(ChannelFactory<? extends C> channelFactory) {
-        if (channelFactory == null) {
-            throw new NullPointerException("channelFactory");
-        }
-        if (this.channelFactory != null) {
-            throw new IllegalStateException("channelFactory set already");
-        }
-
-        this.channelFactory = channelFactory;
-        return self();
-    }
-
-    /**
-     * {@link io.netty.channel.ChannelFactory} which is used to create {@link Channel} instances from
-     * when calling {@link #bind()}. This method is usually only used if {@link #channel(Class)}
-     * is not working for you because of some more complex needs. If your {@link Channel} implementation
-     * has a no-args constructor, its highly recommend to just use {@link #channel(Class)} to
-     * simplify your code.
-     */
-    @SuppressWarnings({ "unchecked", "deprecation" })
-    public B channelFactory(io.netty.channel.ChannelFactory<? extends C> channelFactory) {
-        return channelFactory((ChannelFactory<C>) channelFactory);
     }
 
     /**
@@ -211,9 +165,6 @@ public abstract class AbstractBootstrap<B extends AbstractBootstrap<B, C>, C ext
         if (group == null) {
             throw new IllegalStateException("group not set");
         }
-        if (channelFactory == null) {
-            throw new IllegalStateException("channel or channelFactory not set");
-        }
         return self();
     }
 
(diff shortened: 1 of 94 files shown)


normanmaurer:
I think based on this I have a working idea of how to allow to wrap / decorate `EventLoop` and so inject metrics etc and so address https://github.com/netty/netty/issues/8514

normanmaurer:
Ups missed some stuff on Linux.... fixing now

normanmaurer:
Fixed

arukshani:
This is just for clarification. In our implementation, we use netty underneath for both the client and server. Inside the server, when we have a client endpoint that passes the incoming request straightaway to client without first reading it, we first deregister the client  channel's eventloop and register it with the server channel's eventloop so that everything happens in the same thread. 

I guess with this change, we will no longer be able to achieve that.

normanmaurer:
@arukshani yes that’s correct.... the problem is that what you are doing is quite dangerous in terms of races as it is really trick to ensure there is no pending task in the old EventLoop for a Channel when you register it to a new EventLoop. This can lead to all kind of problems , like for example internal darastructure corruptions in Netty etc.

A better solution to your problem is usually to bootstrap the client with the correct EventLoop from the beginning like we do in our HexDump example https://github.com/netty/netty/blob/4.1/example/src/main/java/io/netty/example/proxy/HexDumpProxyFrontendHandler.java#L43

arukshani:
We do use the server's eventloop in client when bootstrapping it and after the channel is being used we return it to a pool that we maintain. So for the subsequent requests we do not actually know that the returning channel will be in the same eventloop as the server's. That's the reason for our channel reregistering thing. 

But I get your point and thanks for the explanation @normanmaurer .

johnou:
@normanmaurer this would be a real performance killer for us, we have high speed messaging for groups of clients, clients can change groups at any time.

By allowing the client to move between event groups we were able to de-register (if necessary) and re-register to a specific event group pinned to that group.. finally when broadcasting a message we can write from the known event loop to all clients in a group on the same event loop avoiding unnecessary wake-ups.

normanmaurer:
@johnou the problem really is that moving from one EventLoop to another is generally speaking not really safe in terms of thread-safety as there is really no way to ensure that there is no pending tasks in the original EventLoop scheduled easily before moving over the other EventLoop. This can be the reason of many problems then, like for example race-conditions, data-structure corruptions etc.

Honestly I am not really sure how this would be fixable without a lot of overhead but I am happy to be wrong

normanmaurer:
@arukshani why not have a pool that uses the `EventLoop` as the key ?

johnou:
@normanmaurer in our case the client essentially enters a match making mode when changing groups so there shouldn't be any dropped pending tasks, can you think of alternative implementations to achieve the same performance?

normanmaurer:
@johnou well its kind of hard to control what exactly happens... for example the channel have been idle and so a task may have been schedule for the channel to be executed in a few seconds using the current EventLoop. Then you re-register to a new EventLoop and start using it, later then the scheduled task is executed which will have two different threads act on the Channel. This may have all sort of strange side-effects.

I am not sure about your use-case yet. But maybe the right way to do this would to "transfer" the underlying Socket to a new Channel instance.

normanmaurer:
@johnou once this has happened the old Channel would not be usable anymore of course and each operation would fail with some sort of `Exception`.  I am just thinking out loud here though and so have not spent a lot of thoughts on this solution. The only thing I know for sure is that the current approach is not safe and also the possibility to have a Channel not assigned to an EventLoop at some point without have it "closed" is not a good semantic as various operations may fail in this case.

arukshani:
Currently, we use the route(host and port of target channel) as the key for our channel pool. But yes we can rethink our design based on your suggestions. Thank you.

trustin:
Sounds like going into the right direction. I just need to find time for reviewing this. :-)

normanmaurer:
@trustin thanks for taking a peak :)

transport/src/main/java/io/netty/channel/AbstractChannel.java
```diff
@@ -58,13 +59,13 @@
     private final Channel parent;
     private final ChannelId id;
     private final Unsafe unsafe;
-    private final DefaultChannelPipeline pipeline;
+    private final ChannelPipeline pipeline;
```
rkapsi: <3

johnou:
cc @doom369

normanmaurer:
@bryce-anderson @ejona86 @carl-mastrangelo @trustin @rkapsi any chance you will have some time to review this ? I would love to get this merged and so continue some changes on top if it.

normanmaurer:
@bryce-anderson @mosesn @vkostyukov  @ejona86 @carl-mastrangelo @trustin  ping :) I would love to get this in to continue ...

normanmaurer:
rebased.

normanmaurer:
This would also fix https://github.com/netty/netty/pull/4589

transport/src/main/java/io/netty/bootstrap/ServerBootstrap.java
```diff
@@ -239,20 +286,40 @@ public void run() {
...
+        @SuppressWarnings("unchecked")
+        private void initChild(final Channel child) {
+            assert child.eventLoop().inEventLoop();
             try {
-                childGroup.register(child).addListener(new ChannelFutureListener() {
+                child.pipeline().addLast(childHandler);
+
+                setChannelOptions(child, childOptions, logger);
+
+                for (Entry<AttributeKey<?>, Object> e: childAttrs) {
```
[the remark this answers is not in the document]
normanmaurer: 🤣

normanmaurer:
@johnou and don't worry I will come up with a way to handle your use-case. Just need to think a bit more about it.

transport/src/main/java/io/netty/bootstrap/AbstractBootstrap.java
```diff
@@ -47,11 +43,11 @@
  * <p>When not used in a {@link ServerBootstrap} context, the {@link #bind()} methods are useful for connectionless
  * transports such as datagram (UDP).</p>
  */
-public abstract class AbstractBootstrap<B extends AbstractBootstrap<B, C>, C extends Channel> implements Cloneable {
+public abstract class AbstractBootstrap<B extends AbstractBootstrap<B, C, F>, C extends Channel, F>
+        implements Cloneable {
 
     volatile EventLoopGroup group;
     @SuppressWarnings("deprecation")
```
johnou: should remove the suppress as well right?

transport/src/main/java/io/netty/bootstrap/Bootstrap.java
```diff
@@ -58,13 +60,15 @@
     private volatile AddressResolverGroup<SocketAddress> resolver =
             (AddressResolverGroup<SocketAddress>) DEFAULT_RESOLVER;
     private volatile SocketAddress remoteAddress;
+    volatile ChannelFactory<? extends Channel> channelFactory;
```
johnou: wouldn't a bootstrapbuilder make more sense here to avoid the volatile fields (=> final)?

transport/src/main/java/io/netty/bootstrap/FailedChannel.java
```diff
@@ -40,7 +40,7 @@ protected AbstractUnsafe newUnsafe() {
 
     @Override
     protected boolean isCompatible(EventLoop loop) {
-        return false;
+        return true;
```
johnou: on purpose?

transport/src/main/java/io/netty/bootstrap/ServerBootstrap.java
```diff
@@ -49,13 +54,15 @@
     private final ServerBootstrapConfig config = new ServerBootstrapConfig(this);
     private volatile EventLoopGroup childGroup;
     private volatile ChannelHandler childHandler;
+    volatile ServerChannelFactory<? extends ServerChannel> channelFactory;
```
johnou: same here wrt builder.

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -71,23 +71,6 @@
 
     private Map<EventExecutorGroup, EventExecutor> childExecutors;
     private volatile MessageSizeEstimator.Handle estimatorHandle;
-    private boolean firstRegistration = true;
```
johnou: oh the fond memories.

transport/src/main/java/io/netty/bootstrap/Bootstrap.java
normanmaurer: @johnou we can consider this as a followup. This is not related to this PR.

transport/src/main/java/io/netty/bootstrap/ServerBootstrap.java
normanmaurer: https://github.com/netty/netty/pull/8587#discussion_r247474424

transport/src/main/java/io/netty/bootstrap/FailedChannel.java
normanmaurer: Yes as its called from within the constructor. That said the code is removed in later prs that depend on this one.

carl-mastrangelo:
Delete this branch?

normanmaurer:
✅

PR #8590, opened 2018-11-23
Factor out less common code-path into own method to allow inlining.

normanmaurer:
Motivation:

During benchmarks two methods showed up as "hot method too big". We can easily make these smaller by factor out some less common code-path to an extra method and so allow inlining.

Modifications:

Factor out less common code path to an extra method.

Result:

Hot methods can be inlined.

### DIFF
--- a/transport/src/main/java/io/netty/channel/ChannelOutboundBuffer.java
+++ b/transport/src/main/java/io/netty/channel/ChannelOutboundBuffer.java
@@ -451,7 +451,7 @@ public final class ChannelOutboundBuffer {
         return nioBuffers;
     }
 
-    private int nioBuffers(Entry entry, ByteBuf buf, ByteBuffer[] nioBuffers, int nioBufferCount, int maxCount) {
+    private static int nioBuffers(Entry entry, ByteBuf buf, ByteBuffer[] nioBuffers, int nioBufferCount, int maxCount) {
         ByteBuffer[] nioBufs = entry.bufs;
         if (nioBufs == null) {
             // cached ByteBuffers as they may be expensive to create in terms
--- a/transport/src/main/java/io/netty/channel/ChannelOutboundBuffer.java
+++ b/transport/src/main/java/io/netty/channel/ChannelOutboundBuffer.java
@@ -434,21 +434,9 @@ public final class ChannelOutboundBuffer {
                         }
                         nioBuffers[nioBufferCount++] = nioBuf;
                     } else {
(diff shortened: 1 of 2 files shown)


transport/src/main/java/io/netty/channel/ChannelOutboundBuffer.java
```diff
@@ -434,21 +434,9 @@ private void clearNioBuffers() {
...
-                        for (int i = 0; i < nioBufs.length && nioBufferCount < maxCount; ++i) {
-                            ByteBuffer nioBuf = nioBufs[i];
-                            if (nioBuf == null) {
-                                break;
-                            } else if (!nioBuf.hasRemaining()) {
-                                continue;
-                            }
-                            nioBuffers[nioBufferCount++] = nioBuf;
-                        }
+                        // The code resist in an extra method to ensure the method is not too big to inline as this
```
jvirtanen: "The code *exists* in..."?

transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
```diff
@@ -793,17 +793,9 @@ private void select(boolean oldWakenUp) throws IOException {
...
-                    logger.warn(
-                            "Selector.select() returned prematurely {} times in a row; rebuilding Selector {}.",
-                            selectCnt, selector);
-
-                    rebuildSelector();
-                    selector = this.selector;
-
-                    // Select again to populate selectedKeys.
-                    selector.selectNow();
+                    // The code resist in an extra method to ensure the method is not too big to inline as this
```
jvirtanen: "The code _exists_ in ..."?

transport/src/main/java/io/netty/channel/ChannelOutboundBuffer.java
normanmaurer: fat fingers :D thanks

normanmaurer:
@jvirtanen fixed

transport/src/main/java/io/netty/channel/ChannelOutboundBuffer.java
```diff
@@ -463,6 +451,25 @@ private void clearNioBuffers() {
         return nioBuffers;
     }
 
+    private int nioBuffers(Entry entry, ByteBuf buf, ByteBuffer[] nioBuffers, int nioBufferCount, int maxCount) {
```
njhill: static?

normanmaurer:
Alright merging this as it just moves code around :)

PR #8593, opened 2018-11-23
Provide a way to cache the internal nioBuffer of the PooledByteBuffer…

normanmaurer:
… to reduce GC.

Motivation:

Often a temporary ByteBuffer is used which can be cached to reduce the GC pressure.

Modifications:

Add a Queue per PoolChunk which will be used for caching.

Result:

Less GC.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/PoolChunk.java
+++ b/buffer/src/main/java/io/netty/buffer/PoolChunk.java
@@ -16,6 +16,10 @@
 
 package io.netty.buffer;
 
+import java.nio.ByteBuffer;
+import java.util.ArrayDeque;
+import java.util.Deque;
+
 /**
  * Description of algorithm for PageRun/PoolSubpage allocation from PoolChunk
  *
@@ -107,7 +111,6 @@ final class PoolChunk<T> implements PoolChunkMetric {
     final T memory;
     final boolean unpooled;
     final int offset;
-
     private final byte[] memoryMap;
     private final byte[] depthMap;
     private final PoolSubpage<T>[] subpages;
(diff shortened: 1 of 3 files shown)


normanmaurer:
After:
screenshot 2018-11-23 at 15 30 46

Before:
screenshot 2018-11-23 at 13 49 54

normanmaurer:
also @franz1981 and @njhill

buffer/src/main/java/io/netty/buffer/PoolChunk.java
```diff
@@ -122,6 +125,11 @@
     /** Used to mark memory as unusable */
     private final byte unusable;
 
+    // Use as cache for ByteBuffer created from the memory. These are just duplicates and so are only a container
+    // around the memory itself. These are often needed for operations within the Pooled*DirectByteBuf and so
+    // may produce extra GC, which can be greatly reduced by caching the duplicates.
+    private final Deque<ByteBuffer> cachedNioBuffers = new ArrayDeque<ByteBuffer>();
```
franz1981: From the readability point of view I agree about using `ArrayDeque`, from the memory footprint it depends:
array q double the size of the backing `Object[]` *before* approaching the current capacity, while `ArrayList` would do it only while exceeding it, saving 1/2 of the array size if compared to `ArrayDeque`.
To not mention that `ArrayQueue` allow only pow of 2 capacities, while `ArrayList` is more strict with your requests.
For readability, is fine to use `ArrayDeque`, for memory footprint you can use `ArrayList` and `ArrayList::remove(size()-1))`  with a check about being empty first to avoid IOBE.

franz1981: My fault

> ArrayQueue allow only pow of 2 capacities

is instead:

> ArrayQueue adjust required capacity to be only pow of 2

So if DEFAULT_MAX_CACHED_BYTEBUFFER_PER_CHUNK would be a power of 2, by filling it until the limit it will make ArrayDeque allocate a 2*DEFAULT_MAX_CACHED_BYTEBUFFER_PER_CHUNK Objetc[].
There is no need to constraint the DEFAULT_MAX_CACHED_BYTEBUFFER_PER_CHUNK to be a power of 2, but the point is that we would loose some space anyway:

- if is a power of 2, because it would allocate twice the required capacity
- if not, because it use the next power of 2 of what is your limit

TBH it is just a small difference if we are not talking about big numbers if the limit is reasonable.

normanmaurer: @franz1981 yah... I think its fine as it is.

normanmaurer:
So fine with the change @njhill / @franz1981  ?

franz1981:
LGTM too and I approve the comment of @njhill ;)

PR #8594, opened 2018-11-25
Adding an execute burst cost benchmark for Netty executors

franz1981:
Motivation:

Netty executors doesn't have yet any means to compare with each others
nor to compare with the j.u.c. executors

Modifications:

A new benchmark measuring execute burst cost is being added

Result:

It's now possible to compare some of Netty executors with each others
and with the j.u.c. executors

### DIFF
--- a/microbench/src/main/java/io/netty/microbench/concurrent/BurstCostExecutorsBenchmark.java
+++ b/microbench/src/main/java/io/netty/microbench/concurrent/BurstCostExecutorsBenchmark.java
@@ -1,0 +1,331 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License, version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at:
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.netty.microbench.concurrent;
+
+import io.netty.channel.epoll.Epoll;
+import io.netty.channel.epoll.EpollEventLoopGroup;
+import io.netty.channel.kqueue.KQueue;
+import io.netty.channel.kqueue.KQueueEventLoopGroup;
+import io.netty.channel.nio.NioEventLoopGroup;
+import io.netty.microbench.util.AbstractMicrobenchmark;
+import io.netty.util.concurrent.DefaultEventExecutor;
+import io.netty.util.internal.PlatformDependent;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.infra.Blackhole;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Queue;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
+
+@State(Scope.Benchmark)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+public class BurstCostExecutorsBenchmark extends AbstractMicrobenchmark {
+
+    /**
+     * This executor is useful as the best burst latency performer because it won't go to sleep and won't be hit by the
+     * cost of being awaken on both offer/consumer side.
+     */
+    private static final class SpinExecutorService implements ExecutorService {
+
+        private static final Runnable POISON_PILL = new Runnable() {
+            @Override
+            public void run() {
+            }
+        };
+        private final Queue<Runnable> tasks;
+        private final AtomicBoolean poisoned = new AtomicBoolean();
+        private final Thread executorThread;
+
+        public SpinExecutorService(int maxTasks) {
+            tasks = PlatformDependent.newFixedMpscQueue(maxTasks);
+            executorThread = new Thread(new Runnable() {
+                @Override
+                public void run() {
+                    final Queue<Runnable> tasks = SpinExecutorService.this.tasks;
+                    Runnable task;
+                    while ((task = tasks.poll()) != POISON_PILL) {
+                        if (task != null) {
+                            task.run();
+                        }
+                    }
+                }
+            });
+            executorThread.start();
+        }
+
+        @Override
+        public void shutdown() {
+            if (poisoned.compareAndSet(false, true)) {
+                while (!tasks.offer(POISON_PILL)) {
+                    // Just try again
+                }
+                try {
+                    executorThread.join();
+                } catch (InterruptedException e) {
+                    //We're quite trusty :)
+                }
+            }
+        }
+
+        @Override
+        public List<Runnable> shutdownNow() {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public boolean isShutdown() {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public boolean isTerminated() {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public <T> Future<T> submit(Callable<T> task) {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public <T> Future<T> submit(Runnable task, T result) {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public Future<?> submit(Runnable task) {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public <T> List<Future<T>> invokeAll(Collection<? extends Callable<T>> tasks) throws InterruptedException {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public <T> List<Future<T>> invokeAll(Collection<? extends Callable<T>> tasks, long timeout, TimeUnit unit)
+                throws InterruptedException {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public <T> T invokeAny(Collection<? extends Callable<T>> tasks)
+                throws InterruptedException, ExecutionException {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public <T> T invokeAny(Collection<? extends Callable<T>> tasks, long timeout, TimeUnit unit)
+                throws InterruptedException, ExecutionException, TimeoutException {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public void execute(Runnable command) {
+            if (!tasks.offer(command)) {
+                throw new RejectedExecutionException(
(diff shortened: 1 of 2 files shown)


franz1981:
It is useful to compare improvements on both offer and execute side, in particular the cost of signaling the consumer thread to wakeup (given that it always perform the first execute with an empty list of tasks).

image

microbench/src/main/java/io/netty/microbench/concurrent/BurstCostExecutorsBenchmark.java
```diff
@@ -0,0 +1,140 @@
...
+            executor = new DefaultEventExecutor();
+            executorToShutdown = executor;
+            break;
+        case juc:
+            executor = Executors.newSingleThreadScheduledExecutor();
+            executorToShutdown = executor;
+            break;
+        case nioEventLoop:
+            NioEventLoopGroup group = new NioEventLoopGroup(1);
+            group.setIoRatio(1);
```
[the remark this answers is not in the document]
franz1981: I have configured the lowest value for it to allow being mostly executing CPU bounded tasks (that's what the bench does).
Fell free to propose other settings that seems useful to trace perf improvements/regressions ;)

franz1981:
@johnou  re epoll I would like to do it (same for the others), but probably need help to understand how to do it. It would be a conditional bench execution, because there are machines where it isn't available I suppose.

normanmaurer:
@franz1981 couldn't you just use `Epoll. available()` ?

microbench/src/main/java/io/netty/microbench/concurrent/BurstCostExecutorsBenchmark.java
```diff
@@ -0,0 +1,140 @@
...
+    @Param({ "1", "10" })
+    int burstLength;
+    @Param({ "defaultEventExecutor", "juc", "nioEventLoop" })
+    String executorType;
+
+    ExecutorService executor;
+    ExecutorService executorToShutdown;
+
+    @Setup
+    public synchronized void setup() {
```
normanmaurer: why synchronized ?

microbench/src/main/java/io/netty/microbench/concurrent/BurstCostExecutorsBenchmark.java
```diff
@@ -0,0 +1,140 @@
...
+            group.setIoRatio(1);
+            executor = group.next();
+            executorToShutdown = group;
+            break;
+        }
+
+    }
+
+    @TearDown
+    public synchronized void tearDown() {
```
normanmaurer: why synchronized ?

microbench/src/main/java/io/netty/microbench/concurrent/BurstCostExecutorsBenchmark.java
```diff
@@ -0,0 +1,140 @@
...
+        };
+
+        public void resetCompleted() {
+            DONE_UPDATER.lazySet(this, 0);
+        }
+
+        public int spinWaitCompletionOf(int value) {
+            int lastRead;
+            while ((lastRead = this.completed) < value) {
+
```
normanmaurer: add some "comment" here.

microbench/src/main/java/io/netty/microbench/concurrent/BurstCostExecutorsBenchmark.java
```diff
@@ -0,0 +1,140 @@
...
+    @TearDown
+    public synchronized void tearDown() {
+        executorToShutdown.shutdown();
+    }
+
+    @State(Scope.Thread)
+    public static class PerThreadState {
+        private static final AtomicIntegerFieldUpdater<PerThreadState> DONE_UPDATER =
+                AtomicIntegerFieldUpdater.newUpdater(PerThreadState.class, "completed");
+        //thanks to JMH it would be padded avoiding false-sharing for free
```
normanmaurer: Is this also the case why we not use a `AtomicInteger` ?

microbench/src/main/java/io/netty/microbench/concurrent/BurstCostExecutorsBenchmark.java
```diff
@@ -0,0 +1,140 @@
...
+        private volatile int completed = 0;
+
+        private final Runnable completeTask = new Runnable() {
+            @Override
+            public void run() {
+                DONE_UPDATER.lazySet(PerThreadState.this, completed + 1);
```
normanmaurer: can you add a comment why lazySet is good enough ? Like explain we dont need to have it act as a full memory barrier etc... Also for the other case below.

microbench/src/main/java/io/netty/microbench/concurrent/BurstCostExecutorsBenchmark.java
```diff
@@ -0,0 +1,140 @@
...
+            executorToShutdown = executor;
+            break;
+        case nioEventLoop:
+            NioEventLoopGroup group = new NioEventLoopGroup(1);
+            group.setIoRatio(1);
+            executor = group.next();
+            executorToShutdown = group;
+            break;
+        }
+
```
normanmaurer: nit: kill line

microbench/src/main/java/io/netty/microbench/concurrent/BurstCostExecutorsBenchmark.java
```diff
@@ -0,0 +1,140 @@
...
+        nioEventLoop
+    }
+
+    @Param({ "1", "10" })
+    int burstLength;
+    @Param({ "defaultEventExecutor", "juc", "nioEventLoop" })
+    String executorType;
+
+    ExecutorService executor;
+    ExecutorService executorToShutdown;
```
normanmaurer: private ?

microbench/src/main/java/io/netty/microbench/concurrent/BurstCostExecutorsBenchmark.java
franz1981: Added: I haven't noticed that it wasn't that easy to understand it...good point!

normanmaurer:
@netty-bot test this please

microbench/src/main/java/io/netty/microbench/concurrent/BurstCostExecutorsBenchmark.java
```diff
@@ -0,0 +1,166 @@
...
+            executor = nioEventLoopGroup.next();
+            executorToShutdown = nioEventLoopGroup;
+            break;
+        case epollEventLoop:
+            Epoll.ensureAvailability();
+            EpollEventLoopGroup epollEventLoopGroup = new EpollEventLoopGroup(1);
+            epollEventLoopGroup.setIoRatio(1);
+            executor = epollEventLoopGroup.next();
+            executorToShutdown = epollEventLoopGroup;
+            break;
```
normanmaurer: @franz1981 do you also want to include Kqueue ?

franz1981: I have tried to not get any `IllegalStateException` by using 2 separate profile (linux/mac) that would use 2 different bench classes using the right epoll/kqueue even loop, but I'm having some conflicts on the pom. Do you know any other way to address it?
I can live with IllegalStateException while running the bench on my linux machine, but I would prefer to make it clean :P

normanmaurer: hmm as long as you not include these in the `@Params` it should be fine, no ?

franz1981: Yep, but that means that you have to always configure the bench params while running it and you can't just throw a bench as it is: I will just add kqueue without worrying about it so much.
If anyone will complain I can always change it in another form :+1: 
I will do a run for it too, because I'm curious: in the next weeks I was planning to take a look on how to improve some of them. There is anything different on Netty 5 re event loop impls?

johnou: Yes, much to my dismay ;) https://github.com/netty/netty/pull/8587

franz1981: @johnou it seems a big change, but I do understand the concerns of @normanmaurer as well :O

franz1981: @normanmaurer @johnou 
I have notice that I was missing another thing in the bench: a baseline.
Given that I want to use this bench to check improvements/regressions I don't want to spend time (and I think is the same of others) trying to improve things without knowing what is the best that could be achieved. That's why I have put a naive implementation of single consumer executor that won't sleep and that would likely to have the best ever burst cost.
Re the conflicts on the pom, what can I do? I wanted to do with kqueue the same of epoll, but seems not possible...

normanmaurer: @franz1981 @johnou while #8587 does touch things related to EventLoops etc it not really change anything that should affect the benchmark here. Its more about correctness / contract / and safety.

normanmaurer: @franz1981 I think for your merge conflict you should be good after rebase on origin/4.1

normanmaurer:
@netty-bot test this please

normanmaurer:
@franz1981 let me know once I should re-review again.

franz1981:
@normanmaurer Done :) I'm waiting my machine to finish burning soI can put a result here too :)

normanmaurer:
@franz1981 <3

franz1981:
image

normanmaurer:
@franz1981 yep its good to have a benchmark now... That said usually the "effect" is not so bad imho as the EventLoop should be busy a lot of times anyway processing I/O so there is not the wakeup cost all the time.

microbench/src/main/java/io/netty/microbench/concurrent/BurstCostExecutorsBenchmark.java
```diff
@@ -0,0 +1,337 @@
...
+                }
+            });
+            executorThread.start();
+        }
+
+        @Override
+        public void shutdown() {
+            if (poisoned.compareAndSet(false, true)) {
+                while (true) {
+                    if (tasks.offer(POISON_PILL)) {
```
normanmaurer: nit: But I think this could be re-factored to:

```java
while (!task.offer(POISION_POLL) {
     // Just try again
}
```

microbench/src/main/java/io/netty/microbench/concurrent/BurstCostExecutorsBenchmark.java
```diff
@@ -0,0 +1,337 @@
...
+
+    private ExecutorService executor;
+    private ExecutorService executorToShutdown;
+
+    @Setup
+    public void setup() {
+        ExecutorType type = ExecutorType.valueOf(executorType);
+        switch (type) {
+        case spinning:
+            //the case with 3 producers can have a peak of 3*burstLength offers:
```
normanmaurer: nit: The

microbench/src/main/java/io/netty/microbench/concurrent/BurstCostExecutorsBenchmark.java
```diff
@@ -0,0 +1,337 @@
...
+                    while (true) {
+                        final Runnable task = tasks.poll();
+                        if (task != null) {
+                            if (task == POISON_PILL) {
+                                return;
+                            } else {
+                                task.run();
+                            }
+                        }
+                    }
```
normanmaurer: nit this could be simplified:

```java
Runnable task;
while ((task = tasks.poll()) != POISON_PILL) {
    if (task != null) {
        task.run();
    }
}
```

microbench/src/main/java/io/netty/microbench/concurrent/BurstCostExecutorsBenchmark.java
```diff
@@ -0,0 +1,337 @@
...
+            if (poisoned.compareAndSet(false, true)) {
+                while (true) {
+                    if (tasks.offer(POISON_PILL)) {
+                        break;
+                    }
+                }
+                try {
+                    executorThread.join();
+                } catch (InterruptedException e) {
+                    //we're quite trusty :)
```
normanmaurer: nit: we -> We

microbench/src/main/java/io/netty/microbench/concurrent/BurstCostExecutorsBenchmark.java
```diff
@@ -0,0 +1,337 @@
...
+         */
+        public void resetCompleted() {
+            //We can avoid the full barrier cost of a volatile set given that
+            //the counter can be reset from a single thread and it should be reset
+            //only after any submitted tasks are completed
+            DONE_UPDATER.lazySet(this, 0);
+        }
+
+        /**
+         * it would spin-wait until at least {@code value} tasks are being completed.
```
normanmaurer: nit: it -> It

franz1981:
Done :+1:

normanmaurer:
@netty-bot test this please

normanmaurer:
@franz1981 thanks a lot!

carl-mastrangelo:
LGTM.  Thanks @franz1981 !

Issue #8483, opened 2018-11-09
some problems with netty debug log


strongmanwj:
### Expected behavior
netty can print rigth debug log with its parameter.
### Actual behavior
can not print its parameter.
[DEBUG][2018-11-08 16:26:58,872][io.netty.util.internal.logging.InternalLoggerFactory]Using SLF4J as the default logging framework
[DEBUG][2018-11-08 16:26:58,882][io.netty.channel.MultithreadEventLoopGroup]-Dio.netty.eventLoopThreads: {}
[DEBUG][2018-11-08 16:26:58,913][io.netty.channel.nio.NioEventLoop]-Dio.netty.noKeySetOptimization: {}
[DEBUG][2018-11-08 16:26:58,913][io.netty.channel.nio.NioEventLoop]-Dio.netty.selectorAutoRebuildThreshold: {}
### Steps to reproduce
problem happpened when using slf4j-log4j12 for logger.  when i just  use log4j it is ok.  How to solve? 
<dependency>
			<groupId>org.slf4j</groupId>
			<artifactId>slf4j-log4j12</artifactId>
			<version>1.7.25</version>
		</dependency>
### Netty version
4.1.31
### JVM version (e.g. `java -version`)
1.8

normanmaurer:
Should be fixed by #8595

PR #8595, opened 2018-11-26
LocationAwareSlf4jLogger does not correctly format log message.

normanmaurer:
Motivation:

We did miss to use MessageFormatter inside LocationAwareSlf4jLogger and so {} was not correctly replaced in log messages when using slf4j.
This regression was introduced by afe0767e9c150369cd6ea150b5cc1a9021fd5c3c.

Modifications:

- Make use of MessageFormatter
- Add unit test.

Result:

Fixes https://github.com/netty/netty/issues/8483.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/logging/LocationAwareSlf4JLogger.java
+++ b/common/src/main/java/io/netty/util/internal/logging/LocationAwareSlf4JLogger.java
@@ -28,7 +28,7 @@ final class LocationAwareSlf4JLogger extends AbstractInternalLogger {
     // IMPORTANT: All our log methods first check if the log level is enabled before call the wrapped
     // LocationAwareLogger.log(...) method. This is done to reduce GC creation that is caused by varargs.
 
-    private static final String FQCN = LocationAwareSlf4JLogger.class.getName();
+    static final String FQCN = LocationAwareSlf4JLogger.class.getName();
     private static final long serialVersionUID = -8292030083201538180L;
 
     private final transient LocationAwareLogger logger;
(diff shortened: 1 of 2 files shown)


normanmaurer:
@franz1981 @njhill PTAL as well if you have time

common/src/main/java/io/netty/util/internal/logging/LocationAwareSlf4JLogger.java
```diff
@@ -54,28 +54,28 @@ public boolean isTraceEnabled() {
...
-            log(TRACE_INT, msg, null);
+            log(TRACE_INT, msg);
         }
     }
 
     @Override
     public void trace(String format, Object arg) {
         if (isTraceEnabled()) {
-            log(TRACE_INT, format, arg);
+            log(TRACE_INT, org.slf4j.helpers.MessageFormatter.format(format, arg).getMessage(), arg);
```
njhill: Presumably there's no need to also pass the args now? If so it would be better to pass null for the last arg to avoid varargs array allocation.

normanmaurer: from other implementations I think you still should pass the args.

njhill: I still don't follow (sorry I'm probably missing something) ... the additional args are only passed for the purpose of later substitution into the message, but now the substitution is done upfront. So the message passed won't have `{}` corresponding to the args..?

normanmaurer: @njhill when looking at other implementation they all forwarded the "args", so I did the same as I thought maybe they are used somehow later on. I think at worse you have one more alloc but not passing these down may produce other side-effects depending on the underlying implementations.

That said maybe you are right and we should just not do it.

normanmaurer: @njhill PTAL again... I think this addresses your concern and is also cleaner :)

njhill: @normanmaurer LGTM, I was about to suggest something similar :)

common/src/main/java/io/netty/util/internal/logging/LocationAwareSlf4JLogger.java
```diff
@@ -54,28 +58,28 @@ public boolean isTraceEnabled() {
     @Override
     public void trace(String msg) {
         if (isTraceEnabled()) {
```
njhill: Not part of this PR but I think methods like this one don't need the `if (isXxxEnabled())` right?

91he:
There is still a bug. 

When you use "Object... argArray" as an arg, you shall use "MessageFormatter.arrayFormat".

njhill:
@91he I think you are right ... want to open a PR?

njhill:
@91he I've opened a PR #8834 to fix, thanks for reporting this!

PR #8596, opened 2018-11-26
Use addAndGet(...) as a replacement for compareAndSet(...) when track…

normanmaurer:
…ing the direct memory usage.

Motivation:

We can change from using compareAndSet to addAndGet, which emits a different CPU instruction on x86 (CMPXCHG to XADD) when count direct memory usage. This instruction is cheaper in general and so produce less overhead on the "happy path". If we detect too much memory usage we just rollback the change before throwing the Error.

Modifications:

Replace compareAndSet(...) with addAndGet(...)

Result:

Less overhead when tracking direct memory.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent.java
@@ -649,16 +649,12 @@ public final class PlatformDependent {
 
     private static void incrementMemoryCounter(int capacity) {
         if (DIRECT_MEMORY_COUNTER != null) {
-            for (;;) {
-                long usedMemory = DIRECT_MEMORY_COUNTER.get();
-                long newUsedMemory = usedMemory + capacity;
-                if (newUsedMemory > DIRECT_MEMORY_LIMIT) {
-                    throw new OutOfDirectMemoryError("failed to allocate " + capacity
-                            + " byte(s) of direct memory (used: " + usedMemory + ", max: " + DIRECT_MEMORY_LIMIT + ')');
-                }
-                if (DIRECT_MEMORY_COUNTER.compareAndSet(usedMemory, newUsedMemory)) {
-                    break;
-                }
+            long newUsedMemory = DIRECT_MEMORY_COUNTER.addAndGet(capacity);
+            if (newUsedMemory > DIRECT_MEMORY_LIMIT) {
+                DIRECT_MEMORY_COUNTER.addAndGet(-capacity);
+                throw new OutOfDirectMemoryError("failed to allocate " + capacity
+                        + " byte(s) of direct memory (used: " + (newUsedMemory - capacity)
+                        + ", max: " + DIRECT_MEMORY_LIMIT + ')');
             }
         }
     }


normanmaurer:
Also @njhill ....

normanmaurer:
And @franz1981

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -649,16 +649,12 @@ public static void freeDirectNoCleaner(ByteBuffer buffer) {
...
-                long usedMemory = DIRECT_MEMORY_COUNTER.get();
-                long newUsedMemory = usedMemory + capacity;
-                if (newUsedMemory > DIRECT_MEMORY_LIMIT) {
-                    throw new OutOfDirectMemoryError("failed to allocate " + capacity
-                            + " byte(s) of direct memory (used: " + usedMemory + ", max: " + DIRECT_MEMORY_LIMIT + ')');
-                }
-                if (DIRECT_MEMORY_COUNTER.compareAndSet(usedMemory, newUsedMemory)) {
-                    break;
-                }
+            long newUsedMemory = DIRECT_MEMORY_COUNTER.addAndGet(capacity);
```
franz1981: The change is fine: if hammered by different threads it is correct that XADD scales much better, but if we do care that much on the hot path perf why not address false sharing too by padding it? 
Beside that I can't say any draw back beside a slight less precise counting near limits (that's fine, considering that we are trading precision for perfs).

franz1981: Yep: i see 2 options for this:

1. we want a shiny new class to be used elsewhere: hence, create an `PaddedAtomicXXX`
2. we just need a padded field once: on [JCtools](https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/queues/SpscArrayQueue.java#L35) it is addressed through abstract classes, like JMH does but with `long` padding instead (`byte` works better with oops)

I would suggest the latter option, mainly because it is more performant and we are seeking perf.
In additiona, it is less impactfull: no need to maintain/test any new concurrent classes.
Re performance the reason is that it does less pointer chasing (just `PlatformDependent::primitiveCounter` instead of `PlatformDependent::paddedAtomicCounter::primitiveValue`).

normanmaurer: Will do the padding change as followup

normanmaurer:
@njhill @franz1981 thanks for the review.

common/src/main/java/io/netty/util/internal/PlatformDependent.java
trustin: I found padding sometimes make performance worse when a padded object is allocated fairly often. Perhaps it's just OK because reference count update is not very contentious?

normanmaurer: @trustin this is not a problem here as this one will only be allocated once.

trustin: Ah, yes. I somehow thought this is for `ByteBuf`. :sweat_smile:

PR #8597, opened 2018-11-26
Change offset calculating formula to

uxgnoz:
Motivation:

When we create new chunk with memory aligned, the offset of direct memory should be 
'alignment - address & (alignment - 1)',
not just 'address & (alignment - 1)'.

Modification:

Change offset calculating formula to 
offset = alignment - address & (alignment - 1)
in PoolArena.DirectArena#offsetCacheLine
and add a unit test to assert that.

Result:

Change offset calculating formula to 
offset = alignment - address & (alignment - 1)
in PoolArena.DirectArena#offsetCacheLine
and add a unit test to assert that.

### DIFF
--- a/buffer/src/test/java/io/netty/buffer/PoolArenaTest.java
+++ b/buffer/src/test/java/io/netty/buffer/PoolArenaTest.java
@@ -16,9 +16,11 @@
 
 package io.netty.buffer;
 
+import io.netty.util.internal.PlatformDependent;
 import org.junit.Assert;
 import org.junit.Test;
 
+import java.lang.reflect.Method;
 import java.nio.ByteBuffer;
 
 public class PoolArenaTest {
@@ -40,6 +42,30 @@ public class PoolArenaTest {
         int[] expectedResult = {0, 64, 512, 1024, 1024, 2048};
         for (int i = 0; i < reqCapacities.length; i ++) {
             Assert.assertEquals(expectedResult[i], arena.normalizeCapacity(reqCapacities[i]));
+        }
+    }
+
+    @Test
+    public void testDirectArenaOffsetCacheLine() throws Exception {
+        int capacity = 5;
+        int alignment = 128;
+
+        for (int i = 0; i < 1000; i++) {
+            ByteBuffer bb = PlatformDependent.useDirectBufferNoCleaner()
+                    ? PlatformDependent.allocateDirectNoCleaner(capacity + alignment)
+                    : ByteBuffer.allocateDirect(capacity + alignment);
+
+            PoolArena.DirectArena arena = new PoolArena.DirectArena(null, 0, 0, 9, 9, alignment);
+            Method offsetCacheLineMethod = arena.getClass().getDeclaredMethod("offsetCacheLine", ByteBuffer.class);
+            offsetCacheLineMethod.setAccessible(true);
+
+            int offset = (Integer) offsetCacheLineMethod.invoke(arena, bb);
+
+            long address = PlatformDependent.directBufferAddress(bb);
+            int expected = alignment - (int) (address & (alignment - 1));
+
+            Assert.assertEquals(expected, offset);
+            PlatformDependent.freeDirectBuffer(bb);
         }
     }
 
--- a/buffer/src/test/java/io/netty/buffer/PoolArenaTest.java
+++ b/buffer/src/test/java/io/netty/buffer/PoolArenaTest.java
@@ -20,7 +20,6 @@ import io.netty.util.internal.PlatformDependent;
 import org.junit.Assert;
 import org.junit.Test;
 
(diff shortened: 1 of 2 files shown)


normanmaurer:
also @njhill @franz1981

buffer/src/test/java/io/netty/buffer/PoolArenaTest.java
```diff
@@ -43,6 +45,30 @@ public void testNormalizeAlignedCapacity() throws Exception {
...
+        int capacity = 5;
+        int alignment = 128;
+
+        for (int i = 0; i < 1000; i++) {
+            ByteBuffer bb = PlatformDependent.useDirectBufferNoCleaner()
+                    ? PlatformDependent.allocateDirectNoCleaner(capacity + alignment)
+                    : ByteBuffer.allocateDirect(capacity + alignment);
+
+            PoolArena.DirectArena arena = new PoolArena.DirectArena(null, 0, 0, 9, 9, alignment);
+            Method offsetCacheLineMethod = arena.getClass().getDeclaredMethod("offsetCacheLine", ByteBuffer.class);
```
normanmaurer: @uxgnoz can we just mark this method as package-private and not use reflection ?

uxgnoz:
Ok, will do that tmr.


| |
zongxu
邮箱：zongxu@yeah.net
|

Signature is customized by Netease Mail Master

On 11/27/2018 00:31, Norman Maurer wrote:

@normanmaurer requested changes on this pull request.

In buffer/src/test/java/io/netty/buffer/PoolArenaTest.java:

> @@ -43,6 +45,30 @@ public void testNormalizeAlignedCapacity() throws Exception {
         }
     }
 
+    @Test
+    public void testDirectArenaOffsetCacheLine() throws Exception {
+        int capacity = 5;
+        int alignment = 128;
+
+        for (int i = 0; i < 1000; i++) {
+            ByteBuffer bb = PlatformDependent.useDirectBufferNoCleaner()
+                    ? PlatformDependent.allocateDirectNoCleaner(capacity + alignment)
+                    : ByteBuffer.allocateDirect(capacity + alignment);
+
+            PoolArena.DirectArena arena = new PoolArena.DirectArena(null, 0, 0, 9, 9, alignment);
+            Method offsetCacheLineMethod = arena.getClass().getDeclaredMethod("offsetCacheLine", ByteBuffer.class);


@uxgnoz can we just mark this method as package-private and not use reflection ?

uxgnoz:
> @uxgnoz thanks a lot! Can you sign our ICLA and let me know once done ?
> 
> https://netty.io/s/icla

@normanmaurer, it's done.

buffer/src/test/java/io/netty/buffer/PoolArenaTest.java
```diff
@@ -43,6 +44,27 @@ public void testNormalizeAlignedCapacity() throws Exception {
...
+        for (int i = 0; i < 100000; i++) {
+            ByteBuffer bb = PlatformDependent.useDirectBufferNoCleaner()
+                    ? PlatformDependent.allocateDirectNoCleaner(capacity + alignment)
+                    : ByteBuffer.allocateDirect(capacity + alignment);
+
+            PoolArena.DirectArena arena = new PoolArena.DirectArena(null, 0, 0, 9, 9, alignment);
+            int offset = arena.offsetCacheLine(bb);
+
+            long address = PlatformDependent.directBufferAddress(bb);
+            int expected = alignment - (int) (address & (alignment - 1));
```
normanmaurer: Can we somehow change the test that we not duplicate just the logic here but ensure the correct one is set in the chunk. Or at least hardcore the number here that we expect?

uxgnoz: Since address is determined by the runtime dynamically, it seems unlikely to hardcode the offset number we expect here. We can't make sure the correct one is set if we  don't re-calculate based on the address.

Any ideas? :)

normanmaurer: True... Maybe we could just use a modulo operation here tho to proof that what we do with our bitwise tricks is correct ?

uxgnoz: What about this Assert.assertEquals(0, (offset + address) & (alignment - 1))?

normanmaurer: yeah I think thats better es well.

normanmaurer:
@njhill PTAL as well

normanmaurer:
@netty-bot test this please

normanmaurer:
@uxgnoz thanks a lot!

uxgnoz:
@normanmaurer :)

PR #8600, opened 2018-11-27
Move less common code-path to extra method to allow inlining of write…

normanmaurer:
…Utf8.

Motivation:

ByteBuf is used everywhere so we should try hard to be able to make things inlinable. During benchmarks it showed that writeCharSequence(...) fails to inline writeUtf8 because it is too big even if its hots.

Modifications:

Move less common code-path to extra method to allow inlining.

Result:

Be able to inline writeUtf8 in most cases.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
+++ b/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java
@@ -557,17 +557,8 @@ public final class ByteBufUtil {
                     buffer._setByte(writerIndex++, WRITE_UTF_UNKNOWN);
                     break;
                 }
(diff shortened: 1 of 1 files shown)


normanmaurer:
@njhill @franz1981 PTAL

normanmaurer:
@netty-bot test this please

normanmaurer:
@netty-bot test this please

normanmaurer:
@njhill I did this because the or failed with some strange errors and I suspect it may be caused na the reverted commit. So this is for testing this

normanmaurer:
Yep seems like this is the case... lets try again :)

PR #8603, opened 2018-11-27
Provide a way to cache the internal nioBuffer of the PooledByteBuffer…

normanmaurer:
… to reduce GC. (#8593)

Motivation:

Often a temporary ByteBuffer is used which can be cached to reduce the GC pressure.

Modifications:

Add a Deque per PoolChunk which will be used for caching.

Result:

Less GC.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/PoolChunk.java
+++ b/buffer/src/main/java/io/netty/buffer/PoolChunk.java
@@ -16,6 +16,10 @@
 
 package io.netty.buffer;
 
+import java.nio.ByteBuffer;
+import java.util.ArrayDeque;
+import java.util.Deque;
+
 /**
  * Description of algorithm for PageRun/PoolSubpage allocation from PoolChunk
  *
@@ -107,7 +111,6 @@ final class PoolChunk<T> implements PoolChunkMetric {
     final T memory;
     final boolean unpooled;
     final int offset;
-
     private final byte[] memoryMap;
     private final byte[] depthMap;
     private final PoolSubpage<T>[] subpages;
@@ -121,6 +124,13 @@ final class PoolChunk<T> implements PoolChunkMetric {
     private final int maxSubpageAllocs;
     /** Used to mark memory as unusable */
     private final byte unusable;
+
+    // Use as cache for ByteBuffer created from the memory. These are just duplicates and so are only a container
+    // around the memory itself. These are often needed for operations within the Pooled*DirectByteBuf and so
+    // may produce extra GC, which can be greatly reduced by caching the duplicates.
+    //
+    // This may be null if the PoolChunk is unpooled as pooling the ByteBuffer instances does not make any sense here.
+    private final Deque<ByteBuffer> cachedNioBuffers;
 
     private int freeBytes;
 
@@ -163,6 +173,7 @@ final class PoolChunk<T> implements PoolChunkMetric {
         }
 
         subpages = newSubpageArray(maxSubpageAllocs);
+        cachedNioBuffers = new ArrayDeque<ByteBuffer>();
     }
 
     /** Creates a special chunk that is not pooled. */
@@ -182,6 +193,21 @@ final class PoolChunk<T> implements PoolChunkMetric {
         chunkSize = size;
         log2ChunkSize = log2(chunkSize);
         maxSubpageAllocs = 0;
+        cachedNioBuffers = null;
+    }
+
+    ByteBuffer pollCachedNioBuffer() {
+        // We use LIFO to increase the chance that its still "hot" and so in the CPU / L* cache.
+        return cachedNioBuffers != null ? cachedNioBuffers.pollLast() : null;
+    }
+
+    void offerCachedNioBuffer(ByteBuffer nioBuffer) {
+        // Only cache if we did not reach the limit yet and if its not unpooled.
+        // If we do just drop it on the floor and let the GC collect it.
+        if (cachedNioBuffers != null &&
+                cachedNioBuffers.size() <= PooledByteBufAllocator.DEFAULT_MAX_CACHED_BYTEBUFFERS_PER_CHUNK) {
+            cachedNioBuffers.offer(nioBuffer);
+        }
     }
 
     @SuppressWarnings("unchecked")
--- a/buffer/src/main/java/io/netty/buffer/PooledByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/PooledByteBuf.java
@@ -56,13 +56,13 @@ abstract class PooledByteBuf<T> extends AbstractReferenceCountedByteBuf {
 
         this.chunk = chunk;
         memory = chunk.memory;
+        tmpNioBuf = chunk.pollCachedNioBuffer();
         allocator = chunk.arena.parent;
         this.cache = cache;
         this.handle = handle;
         this.offset = offset;
         this.length = length;
         this.maxLength = maxLength;
-        tmpNioBuf = null;
     }
 
     /**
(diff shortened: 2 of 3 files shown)


njhill:
@normanmaurer apologies I didn't review carefully enough the first time and had made some assumptions which I don't think are valid after looking closer.

I don't this is threadsafe in its current form - `ArrayDeque` isn't and the additions/removals could be from arbitrary threads without synchronization.. e.g. `deallocate()` could be in any user thread releasing the buffer. It doesn't look like the `init0` method is synchronized either in general.

I think it may work to add an nio buffer field to `PoolThreadCache.Entry` instead, but not sure if there are are other implications of doing that.

normanmaurer:
No worries.... looking today again

njhill:
I do think stashing it in `PoolThreadCache.Entry` instead of having a separate per-chunk cache would work, looks like it would be easy to plumb through the alloc and dealloc paths. WDYT?

normanmaurer:
@njhill yep agree that should work. I will take care.

normanmaurer:
@njhill done. PTAL again

normanmaurer:
Also @franz1981

buffer/src/main/java/io/netty/buffer/PoolArena.java
```diff
@@ -387,6 +387,7 @@ void reallocate(PooledByteBuf<T> buf, int newCapacity, boolean freeOldMemory) {
         }
 
         PoolChunk<T> oldChunk = buf.chunk;
+        ByteBuffer olNioBuffer = buf.tmpNioBuf;
```
njhill: `olNioBuffer` -> `oldNioBuffer`?

franz1981:
@njhill 
Not sure if it is what you mean, but pre PR:
```
io.netty.buffer.PoolThreadCache$MemoryRegionCache$Entry object internals:
 OFFSET  SIZE                            TYPE DESCRIPTION                               VALUE
      0    12                                 (object header)                           N/A
     12     4   io.netty.util.Recycler.Handle Entry.recyclerHandle                      N/A
     16     8                            long Entry.handle                              N/A
     24     4       io.netty.buffer.PoolChunk Entry.chunk                               N/A
     28     4                                 (loss due to the next object alignment)
Instance size: 32 bytes
Space losses: 0 bytes internal + 4 bytes external = 4 bytes total
```
With this PR:
```
io.netty.buffer.PoolThreadCache$MemoryRegionCache$Entry object internals:
 OFFSET  SIZE                            TYPE DESCRIPTION                               VALUE
      0    12                                 (object header)                           N/A
     12     4   io.netty.util.Recycler.Handle Entry.recyclerHandle                      N/A
     16     8                            long Entry.handle                              N/A
     24     4       io.netty.buffer.PoolChunk Entry.chunk                               N/A
     28     4             java.nio.ByteBuffer Entry.nioBuffer                           N/A
Instance size: 32 bytes
Space losses: 0 bytes internal + 0 bytes external = 0 bytes total
```
We're lucky that at least it won't make any difference on memory footprint (with oops).
My concern instead is related to the effect on GCs that use card-marking (G1 does SATB AFAIK): probably doing a bench aging on purpose the NIO ByteBuf to became old gen would help to check what's the effect on the allocation rate about copying the nioBuffer from Entry<T> to the pooled ByteBuf.

normanmaurer:
@njhill the downside of cache the `ByteBuffer` in the `PoolThreadCache` is that by default we only cache buffers <= 32k there, which means for other buffers there will be no caching happening of the ByteBuffer. That is a bit unfortunate as if pushed hard enough we may use bigger buffers to read from the socket.

njhill:
@normanmaurer I had a go at adding more plumbing onto this change to handle those cases, let me know what you think c29fa22e48f77ef71f03596ec4236a032fc4b651 (minor cleanup would still be needed)

normanmaurer:
@njhill yep I think your commit looks mostly fine. Do you mind if I bring it over into this pr with some cleanup ?

njhill:
The deque bits I copied from your original changes anyhow :)

trustin:
LGTM. Will review again once updated with @njhill's patch.

normanmaurer:
@njhill @trustin @franz1981 PTAL again... I incorporated the changes of @njhill with the exception that I did not cache the `ByteBuffer` for `PoolSubPage` allocations. Thats mainly because how it was done was not safe as we may allocate multiple times per sub-page. We could fix this be adding another `Deque` to the head of the `PoolSubPage` linked list and cache there.  But I think we just can ignore this as these allocations should be small enough to cache in `PoolThreadCache` and so have the `ByteBuffer` cached there anyway.

normanmaurer:
screenshot 2018-11-29 at 09 13 41

New allocation profile with the PR.

njhill:
Thanks @normanmaurer, looks great, I agree this seems good enough to not warrant the additional complexity/overhead of subpage-scoped caches, although without closer analysis/profiling can't judge that for sure.

How about this small improvement/simplification to the latest version though:
- revert the changes to `PoolChunk.allocateSubpage` and `PoolChunk.allocateRun` (so they take only `int` and return `long` again)
- move the `initBuf(...)` call into the common parent `PoolChunk.allocate` method, so that it polls the cache for both cases (as is currently done only in `allocateRun`)

It's "free" and may result in a few more cache hits, WDYT?

normanmaurer:
@njhill good feedback... Just pushed a new commit which should address it. PTAL again

buffer/src/main/java/io/netty/buffer/PoolChunk.java
```diff
@@ -223,11 +223,19 @@ private int usage(int freeBytes) {
...
+        final long handle;
         if ((normCapacity & subpageOverflowMask) != 0) { // >= pageSize
-            return allocateRun(buf, reqCapacity, normCapacity);
+            handle =  allocateRun(normCapacity);
         } else {
-            return allocateSubpage(buf, reqCapacity, normCapacity);
+            handle = allocateSubpage(normCapacity);
         }
+
+        if (handle > 0) {
```
njhill: This was effectively `>= 0` before, not sure if it makes a difference. Might also read slightly simpler if inverted i.e. initBuf outside the if block

normanmaurer: fixed

buffer/src/main/java/io/netty/buffer/PoolChunk.java
```diff
@@ -365,33 +385,41 @@ void free(long handle) {
             PoolSubpage<T> head = arena.findSubpagePoolHead(subpage.elemSize);
             synchronized (head) {
                 if (subpage.free(head, bitmapIdx & 0x3FFFFFFF)) {
+                    // this may be overwriting with same thing, in theory this should never
+                    // overwrite non-null with null.
```
njhill: this comment is N/A now, should be removed

normanmaurer:
@njhill @franz1981 thanks... fixed :)

franz1981:
@normanmaurer Looks fine to me :+1:

johnou:
@normanmaurer I believe this changeset substantially increased our applications memory footprint which resulted in low memory on some boxes and the kernel OOM killer axing java processes :boom: I will see how deploying a canary build when the holiday period is over but wanted to give you a heads up in the meantime.

normanmaurer:
@johnou I would be surprised if this would be the case as it only caches the "container" around the memory. 

You are running the snapshot ?

johnou:
@normanmaurer oh crap, "removed this from the 4.1.32.Final milestone" then it must be something else.. In the meantime downgrading to 4.1.30.Final / 2.0.17.Final ( should hold for Christmas :crossed_fingers: ) I will create a new issue once I get more information.

screenshot from 2018-12-19 19-29-38

normanmaurer:
@johnou let me know once you have more infos (like heap dump etc)

johnou:
@normanmaurer if it's the PooledByteBufAllocator that has simply cached too much that won't show up in a heap dump because it's off-heap right? I just added a new mbean to export ByteBufAllocatorMetric values.

normanmaurer:
@johnou while this is true it may still give some hints as you will see the "container" in the heap dump.

PR #8611, opened 2018-11-30
Set-Cookie headers should not be combined

### LINKED ISSUE #6473: CombinedHttpHeaders indiscriminately combines all headers
### Expected behavior

`CombinedHttpHeaders` combines all headers per name, even if they shouldn't be combined, like for example the `Set-Cookie` header (per [RFC-7230, section 3.2.2](https://tools.ietf.org/html/rfc7230#section-3.2.2)). In fact, there's only a limited set of headers that can be combined (see this [SO thread](http://stackoverflow.com/a/29550711/387927)).

### Actual behavior

`CombinedHttpHeaders` should support a configurable set of header names that can safely be combined (or a set of headers that should never be combined, if we want to combine be default). A default set should be available to in order to provide a sensible behaviour out-of-the-box.

### Netty version

`4.1.x`

madgnome:
Motivation:

According to the HTTP spec set-cookie headers should not be combined
because they are not using the list syntax.

Modifications:

Added a blacklist to CombinedHttpHeaders, headers in the blacklist are
not combined. The default blacklist contains Set-Cookie.

Result:

Set-Cookie headers won't be combined anymore
Fixes #6473

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/CombinedHttpHeaders.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/CombinedHttpHeaders.java
@@ -26,6 +26,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
+import static io.netty.handler.codec.http.HttpHeaderNames.SET_COOKIE;
 import static io.netty.util.AsciiString.CASE_INSENSITIVE_HASHER;
 import static io.netty.util.internal.StringUtil.COMMA;
 import static io.netty.util.internal.StringUtil.unescapeCsvFields;
@@ -87,7 +88,7 @@ public class CombinedHttpHeaders extends DefaultHttpHeaders {
         @Override
         public Iterator<CharSequence> valueIterator(CharSequence name) {
             Iterator<CharSequence> itr = super.valueIterator(name);
-            if (!itr.hasNext()) {
+            if (!itr.hasNext() || cannotBeCombined(name)) {
                 return itr;
             }
             Iterator<CharSequence> unescapedItr = unescapeCsvFields(itr.next()).iterator();
@@ -100,7 +101,7 @@ public class CombinedHttpHeaders extends DefaultHttpHeaders {
         @Override
         public List<CharSequence> getAll(CharSequence name) {
             List<CharSequence> values = super.getAll(name);
-            if (values.isEmpty()) {
+            if (values.isEmpty() || cannotBeCombined(name)) {
                 return values;
             }
             if (values.size() != 1) {
@@ -213,9 +214,13 @@ public class CombinedHttpHeaders extends DefaultHttpHeaders {
             return this;
         }
 
+        private static boolean cannotBeCombined(CharSequence name) {
+            return SET_COOKIE.contentEqualsIgnoreCase(name);
+        }
+
         private CombinedHttpHeadersImpl addEscapedValue(CharSequence name, CharSequence escapedValue) {
             CharSequence currentValue = super.get(name);
-            if (currentValue == null) {
+            if (currentValue == null || cannotBeCombined(name)) {
                 super.add(name, escapedValue);
             } else {
                 super.set(name, commaSeparateEscapedValues(currentValue, escapedValue));
--- a/codec-http/src/test/java/io/netty/handler/codec/http/CombinedHttpHeadersTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/CombinedHttpHeadersTest.java
@@ -22,9 +22,12 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
 
+import static io.netty.handler.codec.http.HttpHeaderNames.SET_COOKIE;
 import static io.netty.util.AsciiString.contentEquals;
+import static org.hamcrest.Matchers.hasSize;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 
 public class CombinedHttpHeadersTest {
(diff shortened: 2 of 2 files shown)


normanmaurer:
@netty-bot test this please

codec-http/src/main/java/io/netty/handler/codec/http/CombinedHttpHeaders.java
```diff
@@ -80,14 +97,16 @@ public CharSequence escape(CharSequence value) {
 
         public CombinedHttpHeadersImpl(HashingStrategy<CharSequence> nameHashingStrategy,
                 ValueConverter<CharSequence> valueConverter,
-                io.netty.handler.codec.DefaultHeaders.NameValidator<CharSequence> nameValidator) {
+                io.netty.handler.codec.DefaultHeaders.NameValidator<CharSequence> nameValidator,
```
normanmaurer: @madgnome can we please keep the original constructor as well ?

codec-http/src/main/java/io/netty/handler/codec/http/CombinedHttpHeaders.java
```diff
@@ -36,8 +39,21 @@
  * Please refer to section <a href="https://tools.ietf.org/html/rfc7230#section-3.2.2">RFC 7230, 3.2.2</a>.
  */
 public class CombinedHttpHeaders extends DefaultHttpHeaders {
+    private static final Set<CharSequence> NON_COMBINABLE_HEADERS;
+
+    static {
+        NON_COMBINABLE_HEADERS = new HashSet<CharSequence>();
+
+        // According to the spec Set-cookie header does not use the list syntax and cannot be combined
+        NON_COMBINABLE_HEADERS.add(SET_COOKIE);
```
normanmaurer: please store this as non-mutable set

normanmaurer:
@slandelle PTAL as well

codec-http/src/main/java/io/netty/handler/codec/http/CombinedHttpHeaders.java
```diff
@@ -213,9 +232,13 @@ public CombinedHttpHeadersImpl setObject(CharSequence name, Iterable<?> values)
             return this;
         }
 
+        private boolean canBeCombined(CharSequence name) {
```
slandelle: You always use the opposite `!canBeCombined` so IMHO it would be better to have a `cantBeCombined` instead (I would expect IntelliJ to complain as well).

slandelle:
LGTM except for a small nit.

codec-http/src/main/java/io/netty/handler/codec/http/CombinedHttpHeaders.java
```diff
@@ -213,9 +232,13 @@ public CombinedHttpHeadersImpl setObject(CharSequence name, Iterable<?> values)
...
+            return !nonCombinableHeaders.contains(name);
```
slandelle: If I'm not missing something, this should be case insensitive. You should probably be using the hasher instead.

normanmaurer: Thats a very good point...  @madgnome please also add a unit test for this.

codec-http/src/main/java/io/netty/handler/codec/http/CombinedHttpHeaders.java
ejona86: Could we just instead special-case this one key (with a plain condition)?

codec-http/src/main/java/io/netty/handler/codec/http/CombinedHttpHeaders.java
```diff
@@ -36,8 +39,21 @@
...
+        // According to the spec Set-cookie header does not use the list syntax and cannot be combined
+        NON_COMBINABLE_HEADERS.add(SET_COOKIE);
+    }
+
     public CombinedHttpHeaders(boolean validate) {
-        super(new CombinedHttpHeadersImpl(CASE_INSENSITIVE_HASHER, valueConverter(validate), nameValidator(validate)));
+        super(new CombinedHttpHeadersImpl(CASE_INSENSITIVE_HASHER, valueConverter(validate), nameValidator(validate), NON_COMBINABLE_HEADERS));
+    }
+
+    public CombinedHttpHeaders(boolean validate, Set<CharSequence> nonCombinableHeaders) {
```
ejona86: Adding this constructor seems like an overkill, to let users provide their own new headers-that-are-broken since [HTTP itself only specifies Set-Cookie as an exception](https://tools.ietf.org/html/rfc7230#section-3.2.2).

codec-http/src/main/java/io/netty/handler/codec/http/CombinedHttpHeaders.java
madgnome: Oh yeah good point, at some point I was using `canBeCombined` without the negation and wanted to avoid the double negative but it's gone now, so will change it.

madgnome:
I've actioned the feedback:

- Removed Set<> and just special cased `Set-Cookie`
- renamed to `cannotCombined` since we were alway using the negative form
- Using `CASE_INSENSITIVE_HASHER` to do the comparison

codec-http/src/main/java/io/netty/handler/codec/http/CombinedHttpHeaders.java
```diff
@@ -213,9 +215,13 @@ public CombinedHttpHeadersImpl setObject(CharSequence name, Iterable<?> values)
             return this;
         }
 
+        private static boolean cannotBeCombined(CharSequence name) {
+            return CASE_SENSITIVE_HASHER.equals(SET_COOKIE, name);
```
normanmaurer: I think you could just use `SET_COOKIE.contentEqualsIgnoreCase(...)`.

madgnome: Fixed!

normanmaurer:
@netty-bot test this please

PR #8612, opened 2018-11-30
NioEventLoop.register(...) should offload to the EventLoop if not alr…

normanmaurer:
…eady on it.

Motivation:

java.nio.channels.spi.AbstractSelectableChannel.register(...) need to obtain multiple locks during execution which may produce a long wait time if we currently select. This lead to multiple CI failures in the past.

Modifications:

Ensure the register call takes place on the EventLoop.

Result:

No more flacky CI test timeouts.

### DIFF
--- a/transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
+++ b/transport/src/main/java/io/netty/channel/nio/NioEventLoop.java
@@ -297,14 +297,19 @@ public final class NioEventLoop extends SingleThreadEventLoop {
         if (inEventLoop()) {
             register0(ch, interestOps, task);
         } else {
-            // Offload to the EventLoop as otherwise java.nio.channels.spi.AbstractSelectableChannel.register
-            // may block for a long time while trying to obtain an internal lock that may be hold while selecting.
-            submit(new Runnable() {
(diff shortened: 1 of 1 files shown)


normanmaurer:
For example as seen here. https://ci.netty.io/job/netty-centos6-java9-prb/317/testReport/io.netty.channel.nio/NioEventLoopTest/testSelectableChannel/

ejona86:
In practice, when do callers need to register when not on the event loop? I sort of question if the bug is with the tests for registering from the wrong thread.

The syncUninterruptably is a bad idea, as I've mentioned elsewhere, since this method could still be called from _an_ event loop. I think that's evidence that this method simply can't be called reliably from another thread.

I think I'd prefer to just put in an assertion check that it is being called on the event loop.

normanmaurer:
@ejona86 we allowed it for ever so I think we can not change this now and it should work. So let me maybe just adjust the code to not call `syncUninterruptably()` but `sync()`. WDYT ?

normanmaurer:
@ejona86 @trustin ping

normanmaurer:
Let me merge this one so the CI is happy.

Issue #8209, opened 2018-08-21
Follow-up question for #8147


rschmitt:
I had a question regarding #8147. On previous versions of Netty, I noticed a problem with the `FlowControlHandler` on HTTP/2 streams with `Http2MultiplexCodec` being used and auto read set to off. The sequence of events I observed was:

1. A handler downstream of the `FlowControlHandler` would call `read`
2. `FlowControlHandler` would look in its queue, find it empty, and call `read` upstream
3. The `Http2MultiplexCodec` would fire `channelReadComplete` without first sending any data downstream
4. The `FlowControlHandler` would [discard this event](https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/flow/FlowControlHandler.java#L155-L160), and the channel would stall

Is this one of the issues fixed by the changes in #8147?

normanmaurer:
@rschmitt sorry I don't get it... can you give some more details ?

rschmitt:
I've put together a repro case. First, the server code:

```java
package io.netty.testsuite.http2;

import io.netty.bootstrap.ServerBootstrap;
import io.netty.buffer.Unpooled;
import io.netty.channel.Channel;
import io.netty.channel.ChannelDuplexHandler;
import io.netty.channel.ChannelFactory;
import io.netty.channel.ChannelHandler;
import io.netty.channel.ChannelHandlerContext;
import io.netty.channel.ChannelInboundHandlerAdapter;
import io.netty.channel.ChannelInitializer;
import io.netty.channel.ChannelPipeline;
import io.netty.channel.EventLoopGroup;
import io.netty.channel.ServerChannel;
import io.netty.channel.nio.NioEventLoopGroup;
import io.netty.channel.socket.nio.NioServerSocketChannel;
import io.netty.handler.codec.http.DefaultFullHttpResponse;
import io.netty.handler.codec.http.HttpRequest;
import io.netty.handler.codec.http.HttpServerCodec;
import io.netty.handler.codec.http.HttpServerUpgradeHandler;
import io.netty.handler.codec.http.HttpServerUpgradeHandler.UpgradeCodec;
import io.netty.handler.codec.http.HttpServerUpgradeHandler.UpgradeCodecFactory;
import io.netty.handler.codec.http.HttpServerUpgradeHandler.UpgradeEvent;
import io.netty.handler.codec.http.HttpUtil;
import io.netty.handler.codec.http.LastHttpContent;
[...]

PR #8617, opened 2018-12-03
Respect ctx.read() calls while processing reads for the child channel…

normanmaurer:
…s when using the Http2MultiplexCodec.

Motivation:

We did not correct respect ctx.read() calls while processing a read for a child Channel. This could lead to read stales when auto read is disabled and no other read was requested.

Modifications:

- Keep track of extra read() calls while processing reads
- Add unit tests that verify that read() is respected when triggered either in channelRead(...) or channelReadComplete(...)

Result:

Fixes https://github.com/netty/netty/issues/8209.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
@@ -446,6 +446,26 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
         return !isStreamIdValid(stream.id()) || isWritable(stream);
     }
 
+    /**
+     * The current status of the read-processing for a {@link Http2StreamChannel}.
+     */
+    private enum ReadStatus {
+        /**
+         * No read in progress and no read was requested (yet)
+         */
+        IDLE,
+
+        /**
+         * Reading in progress
+         */
+        IN_PROGRESS,
+
+        /**
+         * A read operation was requested.
+         */
+        REQUESTED
+    }
+
     // TODO: Handle writability changes due writing from outside the eventloop.
     private final class DefaultHttp2StreamChannel extends DefaultAttributeMap implements Http2StreamChannel {
         private final Http2StreamChannelConfig config = new Http2StreamChannelConfig(this);
@@ -461,13 +481,15 @@ public class Http2MultiplexCodec extends Http2FrameCodec {
         private volatile boolean writable;
 
         private boolean outboundClosed;
+
         /**
-         * This variable represents if a read is in progress for the current channel. Note that depending upon the
-         * {@link RecvByteBufAllocator} behavior a read may extend beyond the {@link Http2ChannelUnsafe#beginRead()}
-         * method scope. The {@link Http2ChannelUnsafe#beginRead()} loop may drain all pending data, and then if the
-         * parent channel is reading this channel may still accept frames.
+         * This variable represents if a read is in progress for the current channe or was requested.
+         * Note that depending upon the {@link RecvByteBufAllocator} behavior a read may extend beyond the
+         * {@link Http2ChannelUnsafe#beginRead()} method scope. The {@link Http2ChannelUnsafe#beginRead()} loop may
+         * drain all pending data, and then if the parent channel is reading this channel may still accept frames.
          */
-        private boolean readInProgress;
+        private ReadStatus readStatus = ReadStatus.IDLE;
+
         private Queue<Object> inboundBuffer;
 
         /** {@code true} after the first HEADERS frame has been written **/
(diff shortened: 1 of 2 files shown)


normanmaurer:
@rschmitt FYI... this should fix your problem. Thanks again for the reproducer.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -1026,7 +1057,11 @@ void readEOS() {
 
             void notifyReadComplete(Handle allocHandle) {
                 assert next == null && previous == null;
-                readInProgress = false;
+                if (readStatus == ReadStatus.REQUESTED) {
+                    readStatus = ReadStatus.IN_PROGRESS;
```
rschmitt: Are you sure this is all you have to do? If auto-read is disabled on the parent channel, might you not have to request a read to ensure progress? (This may happen somewhere else, but I can't see where.)

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -461,13 +481,15 @@ private boolean initialWritability(DefaultHttp2FrameStream stream) {
         private volatile boolean writable;
 
         private boolean outboundClosed;
+
         /**
-         * This variable represents if a read is in progress for the current channel. Note that depending upon the
-         * {@link RecvByteBufAllocator} behavior a read may extend beyond the {@link Http2ChannelUnsafe#beginRead()}
-         * method scope. The {@link Http2ChannelUnsafe#beginRead()} loop may drain all pending data, and then if the
-         * parent channel is reading this channel may still accept frames.
+         * This variable represents if a read is in progress for the current channe or was requested.
```
rschmitt: `channe` => `channel`

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -757,7 +779,7 @@ void fireChildRead(Http2Frame frame) {
             assert eventLoop().inEventLoop();
             if (!isActive()) {
                 ReferenceCountUtil.release(frame);
-            } else if (readInProgress) {
+            } else if (readStatus != ReadStatus.IDLE) {
                 // If readInProgress there cannot be anything in the queue, otherwise we would have drained it from the
```
rschmitt: "If a read is in progress or has been requested, there cannot be anything in the queue" (etc)

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
normanmaurer: @rschmitt that is a different "issue" and I am not even sure if we should do anything here at all by default. You could easily add a `ChannelOutboundHandler` to the `ChannelPipeline` of the Channel which will trigger `channel.parent().read()` when a read is received.

normanmaurer: We could also make this a `ChannelOption` on the "child channel".

normanmaurer: Just to give some more background why I think we should not handle this by default... I think the user may want to handle the read pattern / back pressure independently in some cases.

rschmitt: If this is the philosophy behind this class, then I have no idea why my current h2 server works, because it only reads from the parent channel one time (right at the end of `initChannel`), and autoread is constantly disabled on both parent and child streams.

normanmaurer: I will investigate once this one is pulled in... Like I said I think these are different "concerns".

normanmaurer: @rschmitt now that this one is merged I will check your other concern during this week.

carl-mastrangelo:
Code LGTM

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2MultiplexCodec.java
```diff
@@ -783,7 +805,7 @@ void fireChildRead(Http2Frame frame) {
 
         void fireChildReadComplete() {
             assert eventLoop().inEventLoop();
-            assert readInProgress;
+            assert readStatus == ReadStatus.IN_PROGRESS;
```
bryce-anderson: Is this still accurate? If the user had requested a read while in the `IN_PROGRESS` state I think we transition to the `REQUESTED` state. In fact, the `unsafe.notifyReadComplete` call right triggered next will check if we're in the `REQUESTED` state and transition back to `IN_PROGRESS`.

normanmaurer: @bryce-anderson I think you are right... it should be either IN_PROGRESS or REQUESTED. Let me fix this.

normanmaurer: @bryce-anderson PTAL https://github.com/netty/netty/pull/8639 and thanks

PR #8618, opened 2018-12-03
Update to OpenJDK 12 ea22

normanmaurer:
Motivation:

We should use the latest OpenJDK 12 release when running tests against Java12.

Modifications:

Update to OpenJDK 12 ea22.

Result:

Use latest OpenJDK 12 version when running on the CI.


normanmaurer:
@netty-bot test this please

PR #8619, opened 2018-12-04
Validate pseudo and conditional HTTP/2 headers

madgnome:
**Motivation**:

Netty HTTP/2 implementation is not 100% compliant to the spec ( https://github.com/netty/netty/issues/5761 ). This
commit improves the compliance regarding headers validation,
in particular pseudo-headers and connection ones.

According to the spec:

>    All HTTP/2 requests MUST include exactly one valid value for the
>    ":method", ":scheme", and ":path" pseudo-header fields, unless it is
>    a CONNECT request (Section 8.3).  An HTTP request that omits
>    mandatory pseudo-header fields is malformed (Section 8.1.2.6).

**Modifications**:

- Introduce Http2HeadersValidator class capable of validating HTTP/2
headers
- Invoke validation from DefaultHttp2ConnectionDecoder#onHeadersRead
- Modify tests to use valid headers when required
- Modify HttpConversionUtil#toHttp2Headers to not add :scheme and
:path header on CONNECT method in order to conform to the spec

**Result**:

- Initial requests without :method, :path, :scheme will fail
- Initial requests with multiple values for :method, :path, :scheme
will fail
- Initial requests with an empty :path fail
- Requests with connection-specific header field will fail
- Requests with TE header different than "trailers" will fail
- Fixes 8.1.2.2 tests from h2spec #5761
- Fixes 8.1.2.3 tests from h2spec #5761

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
@@ -31,6 +31,9 @@ import static io.netty.handler.codec.http2.Http2Error.PROTOCOL_ERROR;
 import static io.netty.handler.codec.http2.Http2Error.STREAM_CLOSED;
 import static io.netty.handler.codec.http2.Http2Exception.connectionError;
 import static io.netty.handler.codec.http2.Http2Exception.streamError;
+import static io.netty.handler.codec.http2.Http2HeadersValidator.validateConnectionSpecificHeaders;
+import static io.netty.handler.codec.http2.Http2HeadersValidator.validateRequestPseudoHeaders;
+import static io.netty.handler.codec.http2.Http2HeadersValidator.validateResponsePseudoHeaders;
 import static io.netty.handler.codec.http2.Http2PromisedRequestVerifier.ALWAYS_VERIFY;
 import static io.netty.handler.codec.http2.Http2Stream.State.CLOSED;
 import static io.netty.handler.codec.http2.Http2Stream.State.HALF_CLOSED_REMOTE;
@@ -59,6 +62,7 @@ public class DefaultHttp2ConnectionDecoder implements Http2ConnectionDecoder {
     private final Http2PromisedRequestVerifier requestVerifier;
     private final Http2SettingsReceivedConsumer settingsReceivedConsumer;
     private final boolean autoAckPing;
+    private final boolean validateHeaders;
 
     public DefaultHttp2ConnectionDecoder(Http2Connection connection,
                                          Http2ConnectionEncoder encoder,
@@ -93,6 +97,15 @@ public class DefaultHttp2ConnectionDecoder implements Http2ConnectionDecoder {
         this(connection, encoder, frameReader, requestVerifier, autoAckSettings, true);
     }
 
+    public DefaultHttp2ConnectionDecoder(Http2Connection connection,
+                                         Http2ConnectionEncoder encoder,
+                                         Http2FrameReader frameReader,
+                                         Http2PromisedRequestVerifier requestVerifier,
+                                         boolean autoAckSettings,
+                                         boolean autoAckPing) {
+        this(connection, encoder, frameReader, requestVerifier, autoAckSettings, autoAckPing, false);
+    }
+
     /**
      * Create a new instance.
      * @param connection The {@link Http2Connection} associated with this decoder.
@@ -113,7 +126,8 @@ public class DefaultHttp2ConnectionDecoder implements Http2ConnectionDecoder {
                                          Http2FrameReader frameReader,
                                          Http2PromisedRequestVerifier requestVerifier,
                                          boolean autoAckSettings,
-                                         boolean autoAckPing) {
+                                         boolean autoAckPing,
+                                         boolean validateHeaders) {
         this.autoAckPing = autoAckPing;
         if (autoAckSettings) {
             settingsReceivedConsumer = null;
@@ -128,6 +142,7 @@ public class DefaultHttp2ConnectionDecoder implements Http2ConnectionDecoder {
         this.frameReader = checkNotNull(frameReader, "frameReader");
         this.encoder = checkNotNull(encoder, "encoder");
         this.requestVerifier = checkNotNull(requestVerifier, "requestVerifier");
+        this.validateHeaders = validateHeaders;
         if (connection.local().flowController() == null) {
             connection.local().flowController(new DefaultHttp2LocalFlowController(connection));
         }
@@ -342,6 +357,10 @@ public class DefaultHttp2ConnectionDecoder implements Http2ConnectionDecoder {
                 throw streamError(streamId, PROTOCOL_ERROR,
                                   "Stream %d received too many headers EOS: %s state: %s",
                                   streamId, endOfStream, stream.state());
+            }
+
+            if (validateHeaders) {
+                validateHeaders(streamId, headers, stream);
             }
 
             switch (stream.state()) {
@@ -630,6 +649,18 @@ public class DefaultHttp2ConnectionDecoder implements Http2ConnectionDecoder {
             if (!connection.streamMayHaveExisted(streamId)) {
                 throw connectionError(PROTOCOL_ERROR, "Stream %d does not exist", streamId);
             }
+        }
+
+        private void validateHeaders(int streamId, Http2Headers headers, Http2Stream stream) throws Http2Exception {
+            if (connection.isServer()) {
+                if (!stream.isHeadersReceived() || stream.state() == HALF_CLOSED_REMOTE) {
+                    validateRequestPseudoHeaders(headers, streamId);
+                }
+            } else {
+                validateResponsePseudoHeaders(headers, streamId);
+            }
+
+            validateConnectionSpecificHeaders(headers, streamId);
         }
     }
 
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
@@ -393,13 +393,8 @@ final class HpackDecoder {
                 throw streamError(streamId, PROTOCOL_ERROR, "Invalid HTTP/2 pseudo-header '%s' encountered.", name);
             }
 
-            final HeaderType currentHeaderType = pseudoHeader.isRequestOnly() ?
+            return pseudoHeader.isRequestOnly() ?
                     HeaderType.REQUEST_PSEUDO_HEADER : HeaderType.RESPONSE_PSEUDO_HEADER;
-            if (previousHeaderType != null && currentHeaderType != previousHeaderType) {
-                throw streamError(streamId, PROTOCOL_ERROR, "Mix of request and response pseudo-headers.");
-            }
-
-            return currentHeaderType;
         }
 
         return HeaderType.REGULAR_HEADER;
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2HeadersValidator.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2HeadersValidator.java
@@ -1,0 +1,159 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License, version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
+ * copy of the License at:
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.netty.handler.codec.http2;
+
+import io.netty.handler.codec.http.HttpMethod;
+import io.netty.handler.codec.http2.Http2Headers.PseudoHeaderName;
+import io.netty.util.AsciiString;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map.Entry;
+
+import static io.netty.handler.codec.http.HttpHeaderNames.CONNECTION;
+import static io.netty.handler.codec.http.HttpHeaderNames.KEEP_ALIVE;
+import static io.netty.handler.codec.http.HttpHeaderNames.TE;
+import static io.netty.handler.codec.http.HttpHeaderNames.TRANSFER_ENCODING;
+import static io.netty.handler.codec.http.HttpHeaderNames.UPGRADE;
+import static io.netty.handler.codec.http.HttpHeaderValues.TRAILERS;
+import static io.netty.handler.codec.http2.Http2Error.PROTOCOL_ERROR;
+import static io.netty.handler.codec.http2.Http2Exception.streamError;
+import static io.netty.handler.codec.http2.Http2Headers.PseudoHeaderName.METHOD;
+import static io.netty.handler.codec.http2.Http2Headers.PseudoHeaderName.PATH;
+import static io.netty.handler.codec.http2.Http2Headers.PseudoHeaderName.SCHEME;
+import static io.netty.handler.codec.http2.Http2Headers.PseudoHeaderName.hasPseudoHeaderFormat;
+
+final class Http2HeadersValidator {
+
+    private static final List<AsciiString> connectionSpecificHeaders = Collections.unmodifiableList(
+            Arrays.asList(CONNECTION, TRANSFER_ENCODING, KEEP_ALIVE, UPGRADE));
+
+    private Http2HeadersValidator() {
+    }
+
+    /**
+     * Validates connection-specific headers according to
+     * <a href="https://tools.ietf.org/html/rfc7540#section-8.1.2.2">RFC7540, section-8.1.2.2</a>
+     */
+    static void validateConnectionSpecificHeaders(Http2Headers headers, int streamId) throws Http2Exception {
+        for (int i = 0; i < connectionSpecificHeaders.size(); i++) {
+            final AsciiString header = connectionSpecificHeaders.get(i);
+            if (headers.contains(header)) {
+                throw streamError(streamId, PROTOCOL_ERROR,
+                                  "Connection-speficic headers like [%s] must not be used with HTTP/2.", header);
+            }
+        }
+
+        final CharSequence teHeader = headers.get(TE);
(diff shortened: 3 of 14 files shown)


normanmaurer:
@netty-bot test this please

madgnome:
@netty-bot test this please

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Headers.java
```diff
@@ -42,7 +42,7 @@ public boolean process(byte value) {
         public void validateName(CharSequence name) {
             if (name == null || name.length() == 0) {
                 PlatformDependent.throwException(connectionError(PROTOCOL_ERROR,
-                        "empty headers are not allowed [%s]", name));
+                        "empty headers are not allowed"));
```
normanmaurer: why this change ?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2HeadersValidator.java
```diff
@@ -0,0 +1,108 @@
...
+                throw streamError(streamId, PROTOCOL_ERROR,
+                                  "Header [:scheme] must be omitted when using CONNECT method.");
+            }
+
+            if (headers.contains(PATH.value())) {
+                throw streamError(streamId, PROTOCOL_ERROR,
+                                  "Header [:path] must be omitted when using CONNECT method.");
+            }
+        } else {
+            for (AsciiString header : mandatoryHeaders) {
```
normanmaurer: This will allocate a new Iterator on each loop. Consider using an old style for loop.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2HeadersValidator.java
```diff
@@ -0,0 +1,108 @@
...
+                    throw streamError(streamId, PROTOCOL_ERROR,
+                                      "Mandatory header [%s] is missing.", header);
+                } else if (valueSize > 1) {
+                    throw streamError(streamId, PROTOCOL_ERROR,
+                                      "Header [%s] should have a unique value.", header);
+                }
+            }
+
+            final CharSequence path = headers.get(PATH.value());
+            if (path != null && path.length() == 0) {
```
normanmaurer: should we move this check above the for loop as the check is less expensive.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2HeadersValidator.java
```diff
@@ -0,0 +1,108 @@
...
+     * <a href="https://tools.ietf.org/html/rfc7540#section-8.1.2.3">RFC7540, section-8.1.2.3</a>
+     */
+    static void validatePseudoHeaders(Http2Headers headers, int streamId) throws Http2Exception {
+        final CharSequence method = headers.get(METHOD.value());
+        if (method == null) {
+            throw streamError(streamId, PROTOCOL_ERROR,
+                              "Mandatory header [:method] is missing.");
+        }
+
+        if (HttpMethod.CONNECT.asciiName().contentEqualsIgnoreCase(method.toString())) {
```
normanmaurer: toString() is not needed here.

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2TestUtil.java
```diff
@@ -80,6 +80,15 @@ public void run() {
         return data;
     }
 
+    public static Http2Headers defaultWithPseudoHeaders() {
+        Http2Headers headers = new DefaultHttp2Headers(true);
+        headers.method("GET");
+        headers.path("/");
+        headers.scheme("https");
```
normanmaurer: use method chaining ?

codec-http2/src/test/java/io/netty/handler/codec/http2/Http2TestUtil.java
```diff
@@ -80,6 +80,15 @@ public void run() {
         return data;
     }
 
+    public static Http2Headers defaultWithPseudoHeaders() {
```
normanmaurer: should we maybe rename tis this `newHttp2HeadersWithPseudoHeaders()`

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2Headers.java
madgnome: Oops will change that back

normanmaurer:
@netty-bot test this please

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2HeadersValidator.java
```diff
@@ -0,0 +1,108 @@
...
+     * <a href="https://tools.ietf.org/html/rfc7540#section-8.1.2.3">RFC7540, section-8.1.2.3</a>
+     */
+    static void validatePseudoHeaders(Http2Headers headers, int streamId) throws Http2Exception {
+        final CharSequence method = headers.get(METHOD.value());
+        if (method == null) {
+            throw streamError(streamId, PROTOCOL_ERROR,
+                              "Mandatory header [:method] is missing.");
+        }
+
+        if (HttpMethod.CONNECT.asciiName().contentEqualsIgnoreCase(method)) {
```
bryce-anderson: Do we need to check for multiple `:method` headers?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2HeadersValidator.java
```diff
@@ -0,0 +1,108 @@
...
+            }
+        } else {
+            final CharSequence path = headers.get(PATH.value());
+            if (path != null && path.length() == 0) {
+                throw streamError(streamId, PROTOCOL_ERROR, "[:path] header cannot be empty.");
+            }
+
+            for (int i = 0; i < mandatoryHeaders.size(); i++) {
+                AsciiString header = mandatoryHeaders.get(i);
+                final int valueSize = headers.getAll(header).size();
```
bryce-anderson: This is going to be relatively poor performing considering we're going to make a list for each mandatory header. Since the spec clearly defines which ones are mandatory could we just iterate through the header list in order and count them up, eg have `boolean methodObserved, pathObserved, ...` and make sure we don't do more than one and that we got all three?

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -571,6 +570,22 @@ private void verifyStreamMayHaveExisted(int streamId) throws Http2Exception {
                 throw connectionError(PROTOCOL_ERROR, "Stream %d does not exist", streamId);
             }
         }
+
+        private void validateHeaders(int streamId, Http2Headers headers, Http2Stream stream, boolean endOfStream,
+                                     boolean isInformational) throws Http2Exception {
+            if (connection.isServer() &&
```
bryce-anderson: Do we worry about the client side validation somewhere else?

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2HeadersValidator.java
madgnome: Ah yeah good point! I should add that to h2spec as well

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2HeadersValidator.java
madgnome: Iterating through the whole list would scale linearly with the number of headers whereas this version has a constant run time. If we assume the number of headers is most of the time small iterating would be more performant.

Ideally we would have a `count(name)` method on headers that would just count the headers with a given name, not convinced this usecase justify adding that to the API though.

bryce-anderson: If we use my strategy of having an `observed` variable for each critical header we should count no more than 4 as at 4 we'd hit an exceptional condition. This actually gets to another conformance requirement of the RFC ([section 8.1.2.1](https://http2.github.io/http2-spec/#rfc.section.8.1.2.1)) which is that we should make sure we don't have bonus pseudo headers: `Pseudo-header fields defined for requests MUST NOT appear in responses; pseudo-header fields defined for responses MUST NOT appear in requests.`

madgnome: Nice, forgot about this requirement, will change it. 
(BTW 8.1.2.1 requirement is already checked in the HPackDecoder)

madgnome: They validate at different levels: in `HPackdecoder` we check if the headerBlock is valid, here we check if all the headers together make sense, that being said it would be nice if it was done at a single place.

The problem is that `HPackDecoder` level validation needs to know the order of the headers, we lose the order once we return Http2Headers and  `Http2HeadersValidation` needs to know more about the context: is this an initial header or trailing one? It didn't seem right to push that down to hpackdecoder.

bryce-anderson: That's true regarding the order validation in hpack. Maybe we could shift request/response validation to where you're adding more validation logic, but probably not all of it and at a minimum this isn't obviously right.

madgnome: > If we use my strategy of having an observed variable for each critical header we should count no more than 4 as at 4 we'd hit an exceptional condition.

Thinking about it more, we don't have this guarantee since the headers order is not guaranteed in `Http2Headers`.
The implementation seems to keep the insertion order but this is not specified so potentially subject to change later.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
madgnome: Baby steps, h2spec only validate server specification for now.

madgnome:
I've actioned PR feedbacks:
- provided a way to disable validation
- change the way we validate pseudo headers to be more efficient
- move request/response validation from HPackDecoder to Http2HeadersValidator

codec-http2/src/main/java/io/netty/handler/codec/http2/HpackDecoder.java
```diff
@@ -388,13 +388,8 @@ private static HeaderType validate(int streamId, CharSequence name,
...
             }
 
-            final HeaderType currentHeaderType = pseudoHeader.isRequestOnly() ?
+            return pseudoHeader.isRequestOnly() ?
                     HeaderType.REQUEST_PSEUDO_HEADER : HeaderType.RESPONSE_PSEUDO_HEADER;
-            if (previousHeaderType != null && currentHeaderType != previousHeaderType) {
-                throw streamError(streamId, PROTOCOL_ERROR, "Mix of request and response pseudo-headers.");
-            }
-
-            return currentHeaderType;
```
madgnome: This check was moved to `Http2HeadersValidator` as it doesn't rely on the ordering

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2HeadersValidator.java
```diff
@@ -0,0 +1,143 @@
...
+            for (Entry<CharSequence, CharSequence> entry : headers) {
+                final CharSequence key = entry.getKey();
+                if (!hasPseudoHeaderFormat(key)) {
+                    // We know that pseudo header appears first so we can stop
+                    // looking once we get to the first non pseudo headers.
+                    break;
+                }
+
+                final PseudoHeaderName pseudoHeader = PseudoHeaderName.getPseudoHeader(key);
+                if (previousPseudoHeader != null &&
```
bryce-anderson: Since this was moved out of HpackDecoder I think we should now validate the request pseudo headers as it leaves a hole otherwise.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2HeadersValidator.java
```diff
@@ -0,0 +1,143 @@
...
+                    pseudoHeader.isRequestOnly() != previousPseudoHeader.isRequestOnly()) {
+                    throw streamError(streamId, PROTOCOL_ERROR, "Mix of request and response pseudo-headers.");
```
bryce-anderson: Since we're only validating the request specific headers I think we can make this another `else` clause in the branches below.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2HeadersValidator.java
```diff
@@ -0,0 +1,143 @@
...
+                }
+                previousPseudoHeader = pseudoHeader;
+
+                if (METHOD.value().contentEqualsIgnoreCase(key)) {
```
bryce-anderson: H2 headers must always have lower case keys and pseudo-headers should be no exception, so we should be able to omit the `IngoreCase`.

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2HeadersValidator.java
```diff
@@ -0,0 +1,143 @@
...
+            throw streamError(streamId, PROTOCOL_ERROR,
+                              "TE header must not contain any value other than \"%s\"", TRAILERS);
+        }
+    }
+
+    /**
+     * Validates Pseudo-Header fields according to
+     * <a href="https://tools.ietf.org/html/rfc7540#section-8.1.2.3">RFC7540, section-8.1.2.3</a>
+     */
+    static void validatePseudoHeaders(Http2Headers headers, int streamId) throws Http2Exception {
```
bryce-anderson: Even if we don't plan on doing the client side now we should probably prepare for that by making the name reflect that it is request specific.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
```diff
@@ -571,6 +589,22 @@ private void verifyStreamMayHaveExisted(int streamId) throws Http2Exception {
...
+                (!stream.isHeadersReceived() || stream.state() == HALF_CLOSED_REMOTE)) {
+                validatePseudoHeaders(headers, streamId);
+            }
+
+            validateConnectionSpecificHeaders(headers, streamId);
+
+            if ((isInformational || !endOfStream) && stream.isHeadersReceived() || stream.isTrailersReceived()) {
```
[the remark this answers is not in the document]
madgnome: Oh yeah good point, will wait on other opinion for now

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2HeadersValidator.java
madgnome: This comment and the next one make me think we should probably leave this check in HPackDecoder.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionDecoder.java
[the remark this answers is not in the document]
madgnome: I moved that out of the condition in the end.

madgnome:
Any other opinion? @ejona86, @Scottmitch ?

normanmaurer:
@ejona86 @carl-mastrangelo @bryce-anderson @Scottmitch @mosesn @rkapsi  so what we want to do with this ? Sorry I totally lost track of this one :/

carl-mastrangelo:
@normanmaurer I like the idea of validation, but I don't know if I want to take on the performance penalty (Ctrl F for "bench" doesn't bring up anything).    In the case where most clients and servers are well behaved (i.e. both owned by the same party) it doesn't make sense to add so much validation.  Not sure this would catch any violations.

That said, if @madgnome can show that the default on case is not so expensive, following the spec more strictly can only be good for Netty users.

normanmaurer:
@madgnome @carl-mastrangelo @ejona86 @bryce-anderson what do we want to do with this one ? I think I am fine with adding this as long as we use a default of false for now. WDYT ?

bryce-anderson:
I'm fine with introducing it with a default of disabled.

normanmaurer:
@carl-mastrangelo @ejona86 @trustin WDYT?

trustin:
+1. Maybe better defaulting to 'enabled' and allowing to disable?

normanmaurer:
@trustin I am a bit afraid that this will break apps when upgrading ... you are not ?

trustin:
Shouldn't we consider this as a bug fix? What requests that violate the validation rules in this PR could be sent in reality?

normanmaurer:
@trustin I think it is more about what @bryce-anderson mentioned at some point... It may hurt proxies.

trustin:
What if we validate only pseudo headers and TE strictly and remove the connection headers from HTTP/1 then?

normanmaurer:
@bryce-anderson @ejona86 @carl-mastrangelo WDYT ?

bryce-anderson:
Will there be an option to disable all this new validation? I'm not strongly opinionated about it unless we can't disable it.

normanmaurer:
Wouldn’t it be good enough to allow to disable this with all the other header validation?

normanmaurer:
rebased this one

normanmaurer:
@bryce-anderson sorry missed to tag you on my response...

bryce-anderson:
@normanmaurer, yes, that should be fine if it can be disabled with the rest of the validation.

normanmaurer:
Let me merge this ... As stated it can be disabled with the rest of the validation and follows more closely the RFC. We will add a note in the release note.

normanmaurer:
@netty-bot test this please

normanmaurer:
@netty-bot test this please

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2HeadersValidator.java
```diff
@@ -0,0 +1,159 @@
...
+    /**
+     * Validates connection-specific headers according to
+     * <a href="https://tools.ietf.org/html/rfc7540#section-8.1.2.2">RFC7540, section-8.1.2.2</a>
+     */
+    static void validateConnectionSpecificHeaders(Http2Headers headers, int streamId) throws Http2Exception {
+        for (int i = 0; i < connectionSpecificHeaders.size(); i++) {
+            final AsciiString header = connectionSpecificHeaders.get(i);
+            if (headers.contains(header)) {
+                throw streamError(streamId, PROTOCOL_ERROR,
+                                  "Connection-speficic headers like [%s] must not be used with HTTP/2.", header);
```
jplock: spelling: specific

normanmaurer:
@madgnome thanks a lot! Sorry for the long turn-around

Issue #8616, opened 2018-12-01
ChannelInitializer.initChannel() executed more than once when used with context executor


atcurtis:
### Expected behavior

Expected ChannelInitializer.initChannel() to be executed one. 

### Actual behavior

A subsequent execution causes exception to be thrown because it can't find itself or remove itself again.

```
2018-12-01 01:22:02,527 [WARN] [pool-2-thread-1] ChannelInitializer ? Failed to initialize a channel. Closing: [id: 0xce54469f, L:local:test - R:local:E:d49bc512]
java.lang.AssertionError: expected object to not be null
	at org.testng.Assert.fail(Assert.java:93) ~[testng-6.11.jar:?]
	at org.testng.Assert.assertNotNull(Assert.java:422) ~[testng-6.11.jar:?]
	at org.testng.Assert.assertNotNull(Assert.java:407) ~[testng-6.11.jar:?]
	at test.TestChannelInitializer$2.initChannel(TestChannelInitializer.java:93) ~[espresso-router-impl/:?]
	at io.netty.channel.ChannelInitializer.initChannel(ChannelInitializer.java:113) [netty-all-4.1.22.Final.jar:4.1.22.Final]
	at io.netty.channel.ChannelInitializer.channelRegistered(ChannelInitializer.java:76) [netty-all-4.1.22.Final.jar:4.1.22.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRegistered(AbstractChannelHandlerContext.java:149) [netty-all-4.1.22.Final.jar:4.1.22.Final]
	at io.netty.channel.AbstractChannelHandlerContext.access$000(AbstractChannelHandlerContext.java:38) [netty-all-4.1.22.Final.jar:4.1.22.Final]
	at io.netty.channel.AbstractChannelHandlerContext$1.run(AbstractChannelHandlerContext.java:140) [netty-all-4.1.22.Final.jar:4.1.22.Final]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [?:1.8.0_121]
	at java.util.concurrent.FutureTask.run(FutureTask.java:266) [?:1.8.0_121]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180) [?:1.8.0_121]
[...]

PR #8620, opened 2018-12-04
ChannelInitializer may be invoked multiple times when used with custo…

normanmaurer:
…m EventExecutor.

Motivation:

The ChannelInitializer may be invoked multipled times when used with a custom EventExecutor as removal operation may be done asynchronously. We need to guard against this.

Modifications:

- Change Map to Set which is more correct in terms of how we use it.
- Ensure we only modify the internal Set when the handler was removed yet
- Add unit test.

Result:

Fixes https://github.com/netty/netty/issues/8616.

### DIFF
--- a/transport/src/main/java/io/netty/channel/ChannelInitializer.java
+++ b/transport/src/main/java/io/netty/channel/ChannelInitializer.java
@@ -18,11 +18,12 @@ package io.netty.channel;
 import io.netty.bootstrap.Bootstrap;
 import io.netty.bootstrap.ServerBootstrap;
 import io.netty.channel.ChannelHandler.Sharable;
-import io.netty.util.internal.PlatformDependent;
 import io.netty.util.internal.logging.InternalLogger;
 import io.netty.util.internal.logging.InternalLoggerFactory;
 
-import java.util.concurrent.ConcurrentMap;
+import java.util.Collections;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * A special {@link ChannelInboundHandler} which offers an easy way to initialize a {@link Channel} once it was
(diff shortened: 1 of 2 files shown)


normanmaurer:
@atcurtis This should fix your problem

transport/src/main/java/io/netty/channel/ChannelInitializer.java
```diff
@@ -125,14 +132,25 @@ private boolean initChannel(ChannelHandlerContext ctx) throws Exception {
...
         try {
             ChannelPipeline pipeline = ctx.pipeline();
             if (pipeline.context(this) != null) {
                 pipeline.remove(this);
             }
         } finally {
-            initMap.remove(ctx);
+            // The removal may happen in an async fashion if the EventExecutor we use does something funky.
+            if (ctx.isRemoved()) {
+                initMap.remove(ctx);
```
ejona86: What's the point of this line? Wouldn't we only get here if `handlerRemoved()` was called? Is the concern that someone could override `handlerRemoved()`?

transport/src/main/java/io/netty/channel/ChannelInitializer.java
```diff
@@ -125,14 +132,25 @@ private boolean initChannel(ChannelHandlerContext ctx) throws Exception {
...
+            } else {
+                // Ensure we always remove from the Map in all cases to not produce a memory leak.
+                ctx.channel().closeFuture().addListener(new ChannelFutureListener() {
```
ejona86: This will retain `ctx` until the channel is closed, even if `handlerRemoved()` is called. If this is a one-off initializer (not reused) then that could greatly increase the lifetime of some objects used during handshakes and the like.

Is this seriously the only way we can clean up properly? Is the problem that the methods can be overridden? If so, then maybe we should fix that in Netty 5? Or maybe we could avoid the check-for-double-register problem by having an alternative approach to 26aa34853a8974d212e12b98e708790606bea5fa in Netty 5?

It is possible to cancel the future to clean up, but the amount of effort necessary to do something simple makes me question if this is more of systemic problem that should be addressed in Netty 5.

transport/src/main/java/io/netty/channel/ChannelInitializer.java
normanmaurer: yes :/

transport/src/main/java/io/netty/channel/ChannelInitializer.java
normanmaurer: @ejona86 yeah I think its the best we can do for netty 4 to ensure we not leak if the user overrides `handlerRemoved(...)`.Thats also why I check `isRemoved()` first.

trustin: Perhaps we could check handlerRemoved() was overridden at instantiation time using reflection and take the fast path?

normanmaurer: @trustin we could ... that said I think it does not really worth it as `ctx.isRemoved()` should be true if the `EventExecutor` is not some funky custom implemention that is is used with `ChannelInitializer`. So I would prefer to keep things simple for now.

trustin: Agreed.

trustin:
You may want to document about this behavior in Javadoc though, e.g. risks of overriding `handlerRemoved()` and using a certain `EventExecutor` implementation.

normanmaurer:
Good idea

normanmaurer:
@trustin actually I think it is fine to not add docs as the same is true for other methods as well.

normanmaurer:
@ejona86 any other comment ?

PR #8621, opened 2018-12-04
Loosen bounds check on CompositeByteBuf's maxNumComponents

### LINKED ISSUE #8613: Allocating an unpooled CompositeBuffer of size 1 throws an IllegalArgumentException
### Expected behavior

Allocating a new composite buffer with max components equal to 1 via `Unpooled.compositeBuffer` allocates a new `CompositeBuffer` instance.

### Actual behavior

As of `4.1.32.Final`, this method will now throw an `IllegalStateException`, indicating that the argument must be greater than or equal to two:

```
java.lang.IllegalArgumentException: maxNumComponents: 1 (expected: >= 2)
        at io.netty.buffer.CompositeByteBuf.<init>(CompositeByteBuf.java:62)
        at io.netty.buffer.CompositeByteBuf.<init>(CompositeByteBuf.java:72)
        at io.netty.buffer.Unpooled.compositeBuffer(Unpooled.java:351)
```

This is firing in one of our tests. We should fix our code to not allocate composite buffers of this size, but this was a breaking change that we weren't expecting. This is more of a heads up about the breaking change.

This change was introduced in #8420.

### Steps to reproduce

Apply the following patch:

```diff
diff --git a/buffer/src/test/java/io/netty/buffer/UnpooledTest.java b/buffer/src/test/java/io/netty/buffer/UnpooledTest.java
index f7fd30ba9..c33570f0f 100644
--- a/buffer/src/test/java/io/netty/buffer/UnpooledTest.java
+++ b/buffer/src/test/java/io/netty/buffer/UnpooledTest.java
@@ -39,6 +39,15 @@ public class UnpooledTest {
     private static final ByteBuf[] EMPTY_BYTE_BUFS = new ByteBuf[0];
     private static final byte[][] EMPTY_BYTES_2D = new byte[0][];

+    @Test
+    public void testCompositeBufferSizeOne() {
+        CompositeByteBuf byteBuf = compositeBuffer(1);
+
+        assertEquals(1, byteBuf.maxNumComponents());
+
+        byteBuf.release();
+    }
+
     @Test
     public void testCompositeWrappedBuffer() {
         ByteBuf header = buffer(12);
```

and run

```
$ mvn -pl buffer -Dtest=UnpooledTest#testCompositeBufferSizeOne test
```

### Netty version

`4.1.32.Final`

```
$ java -version
java version "1.8.0_162"
Java(TM) SE Runtime Environment (build 1.8.0_162-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.162-b12, mixed mode)
```

```
$ uname -a
Darwin nickt.local 18.2.0 Darwin Kernel Version 18.2.0: Fri Oct  5 19:41:49 PDT 2018; root:xnu-4903.221.2~2/RELEASE_X86_64 x86_64
```

nicktrav:
Motivation:

In versions of Netty prior to 4.1.31.Final, a CompositeByteBuf could be
created with any size (including potentially nonsensical negative
values). This behavior changed in e7737b993, which introduced a bounds
check to only allow for a component size greater than one. This broke
some existing use cases that attempted to create a byte buf with a
single component.

Modifications:

Lower the bounds check on numComponents to include the single component
case, but still throw an exception for anything less than one.

Add unit tests for the case of numComponents being less than, equal to,
and greater than this lower bound.

Result:

Return to the behavior of 4.1.30.Final, allowing one component, but
still include an explicit check against a lower bound.

Note that while creating a CompositeByteBuf with a single component is
in some ways a contradiction of the term "composite", this patch caters
for existing uses while excluding the clearly nonsensical case of asking
for a CompositeByteBuf with zero or fewer components.

Fixes #8613.


nicktrav:
cc: @njhill

normanmaurer:
@netty-bot test this please

buffer/src/test/java/io/netty/buffer/CompositeByteBufTest.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * Assert that a new {@linkplain CompositeByteBuf} was created successfully with the desired number of max
+     * components.
+     */
+    private static void assertCompositeBufCreated(int expectedMaxComponents) {
+        CompositeByteBuf buf = new CompositeByteBuf(ALLOC, true, expectedMaxComponents);
+
+        assertNotNull(buf);
```
njhill: no need for this check

normanmaurer:
@nicktrav @njhill has a good point here... maybe do this :)

buffer/src/test/java/io/netty/buffer/CompositeByteBufTest.java
nicktrav: derp. of course. fixed.

nicktrav:
Move the tests 👍 .

normanmaurer:
@netty-bot test this please

normanmaurer:
@nicktrav thanks a lot buddy!

PR #8622, opened 2018-12-04
Correcting Maven Dependencies

### LINKED ISSUE #8496: Why dependencies to common and buffer are transitive?
Many of the modules use common and buffer directly (for example by defining a variable of a type declared in buffer module). However, they do not have explicit dependency to these modules (in maven) and the jars are added because of transitivity of dependencies. What is the reason? Isn't it a good practice to explicitly define these kind of dependencies?

Feri73:
Motivation:

Most of the maven modules do not explicitly declare their
dependencies and rely on transitivity, which is not always correct.

Modifications:

For all maven modules, add all of their dependencies to pom.xml

Result:

Fixes #8496

### DIFF
--- a/codec-xml/pom.xml
+++ b/codec-xml/pom.xml
@@ -35,12 +35,17 @@
   <dependencies>
     <dependency>
       <groupId>${project.groupId}</groupId>
-      <artifactId>netty-codec</artifactId>
+      <artifactId>netty-buffer</artifactId>
       <version>${project.version}</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
-      <artifactId>netty-handler</artifactId>
+      <artifactId>netty-transport</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>netty-codec</artifactId>
       <version>${project.version}</version>
     </dependency>
     <dependency>
--- a/transport-udt/pom.xml
+++ b/transport-udt/pom.xml
@@ -39,10 +39,14 @@
 
     <dependency>
       <groupId>${project.groupId}</groupId>
+      <artifactId>netty-common</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
       <artifactId>netty-buffer</artifactId>
       <version>${project.version}</version>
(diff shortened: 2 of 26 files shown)


slandelle:
Personnally, I’m not convinced this is an improvement. Looks like more boilerplate to me.

IMHO, the problem with transitive dependencies is not with intra-project modules deps, but with external ones, eg pulling libA and expect it to always pull libB (which might actually change from one version to another).

Could you please elaborate on how the current configuration is a potential issue?

Feri73:
Currently there is no pattern for specifying the dependencies (at least I cannot detect a pattern). Randomly some dependencies are not explicitly specified and are left to be resolved by transitivity. Having explicit dependencies and a neat standard for writing them makes it easy for programmers to read and reason about the program. e.g. they don't need to perform a dfs to see if a module is dependent on another module. Also analysis of the architecture will be more realistic.
About that external vs internal dependencies, the problem is still there in the internal ones, but you can solve it faster.

normanmaurer:
@netty-bot test this please

transport-udt/pom.xml
```diff
@@ -78,7 +82,7 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
-          <forkMode>always</forkMode>
+          <reuseForks>false</reuseForks>
```
normanmaurer: Is this just to replace "deprecated" naming ?

normanmaurer:
@slandelle I think its a good idea as it also makes it really clear what stuff is used where.

transport-udt/pom.xml
Feri73: yes

normanmaurer:
@Feri73 thanks!

codec-xml/pom.xml
```diff
@@ -35,12 +35,17 @@
...
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>netty-transport</artifactId>
       <version>${project.version}</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
-      <artifactId>netty-handler</artifactId>
```
johnou: was handler dropped on purpose?

Feri73: Yes it's not used.

Issue #8525, opened 2018-11-13
[Netty 5] Decoder/Encoder API 


normanmaurer:
`MessageToMessage` and `ByteToMessage` `encoder`/`decoder` require using an intermediate `List` to put results into. This intermediate list adds overhead (memory/CPU) which grows as the number of objects increases. This overhead can be avoided by directly propagating events through the `ChannelPipeline`. 

We will remove the intermediate `List`, and rely upon lower level queues that are already in place.

normanmaurer:
Let me do this tomorrow.

hepin1989:
@normanmaurer That's fine, this new design reduces the GC overhead.

normanmaurer:
This was done for the decoder side now.

PR #8626, opened 2018-12-05
Remove the intermediate List from ByteToMessageDecoder (and sub-class…

normanmaurer:
…es) to reduce overhead.

Motivation:

ByteToMessageDecoder requires using an intermediate List to put results into. This intermediate list adds overhead (memory/CPU) which grows as the number of objects increases. This overhead can be avoided by directly propagating events through the ChannelPipeline via ctx.fireChannelRead(...). This also makes the semantics more clear and allows us to keep track if we need to call ctx.read() in all cases.

Modifications:

- Remove List from the method signature of ByteToMessageDecoder.decode(...) and decodeLast(...)
- Adjust all sub-classes
- Adjust unit tests
- Fix javadocs.

Result:

Adjust ByteToMessageDecoder as noted in https://github.com/netty/netty/issues/8525.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpServerCodec.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpServerCodec.java
@@ -20,7 +20,6 @@ import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.CombinedChannelDuplexHandler;
 
 import java.util.ArrayDeque;
-import java.util.List;
 import java.util.Queue;
 
 /**
@@ -81,6 +80,9 @@ public final class HttpServerCodec extends CombinedChannelDuplexHandler<HttpRequ
     }
 
     private final class HttpServerRequestDecoder extends HttpRequestDecoder {
+
+        private ChannelHandlerContext context;
+
         HttpServerRequestDecoder(int maxInitialLineLength, int maxHeaderSize) {
             super(maxInitialLineLength, maxHeaderSize);
         }
@@ -96,16 +98,23 @@ public final class HttpServerCodec extends CombinedChannelDuplexHandler<HttpRequ
         }
 
         @Override
-        protected void decode(ChannelHandlerContext ctx, ByteBuf buffer, List<Object> out) throws Exception {
-            int oldSize = out.size();
-            super.decode(ctx, buffer, out);
-            int size = out.size();
-            for (int i = oldSize; i < size; i++) {
-                Object obj = out.get(i);
-                if (obj instanceof HttpRequest) {
-                    queue.add(((HttpRequest) obj).method());
+        protected void decode(final ChannelHandlerContext ctx, ByteBuf buffer) throws Exception {
+            super.decode(context, buffer);
+        }
+
+        @Override
+        protected void handlerAdded0(final ChannelHandlerContext ctx) {
+            context = new DelegatingChannelHandlerContext(ctx) {
+
+                @Override
+                public ChannelHandlerContext fireChannelRead(Object msg) {
+                    if (msg instanceof HttpRequest) {
+                        queue.add(((HttpRequest) msg).method());
+                    }
+                    super.fireChannelRead(msg);
+                    return this;
                 }
-            }
+            };
         }
     }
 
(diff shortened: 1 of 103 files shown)


normanmaurer:
This is the first part of https://github.com/netty/netty/issues/8525 ... PTAL. I think it simplifies things in general and makes things more clear.

normanmaurer:
Also @johnou @doom369 @njhill @franz1981

normanmaurer:
Another possible API would be to change from:

```java
void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out)
```

to:

```java
Object decode(ChannelHandlerContext ctx, ByteBuf in)
```

I am not sure what I like more yet ... any comments ?

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerCodec.java
```diff
@@ -96,16 +97,23 @@ public HttpServerRequestDecoder(int maxInitialLineLength, int maxHeaderSize, int
...
+        }
+
+        @Override
+        protected void handlerAdded0(final ChannelHandlerContext ctx) {
+            context = new DelegatingChannelHandlerContext(ctx) {
+
+                @Override
+                public ChannelHandlerContext fireChannelRead(Object msg) {
+                    if (msg instanceof HttpRequest) {
+                        queue.add(((HttpRequest) msg).method());
```
johnou: I haven't played with HttpServerCodec.java but this doesn't feel the same as the previous implementation where it delegated so the next handler, then pushed the read objects into the queue..?

normanmaurer: @johnou its actually doing exactly the same as the object that were in the queue were fired though the next handler.

franz1981:
@normanmaurer I vote fore `void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out)` :)

normanmaurer:
@franz1981 welcome to the code then ;) Any kind of feedback is welcome as always.

johnou:
> > Another possible API would be to change from:
> > ```java
> > void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out)
> > ```
> > 
> > 
> > to:
> > ```java
> > Object decode(ChannelHandlerContext ctx, ByteBuf in)
> > ```
> > 
> > 
> > I am not sure what I like more yet ... any comments ?
> 
> Returning an object seems reasonable yet may not be as flexible as calling into `ctx.fireRead`. The latter allows us to decode multiple object at once w/o bouncing through decode loop.

can we use generics?

normanmaurer:
@johnou sure but I am not sure what this would buy us.

doom369:
As far as I understood you want to make it as part of netty 5?

normanmaurer:
@doom369 yes

okou19900722:
> Another possible API would be to change from:
> 
> ```java
> void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out)
> ```
> 
> to:
> 
> ```java
> Object decode(ChannelHandlerContext ctx, ByteBuf in)
> ```
> 
> I am not sure what I like more yet ... any comments ?

The second might be better, it lets us know if the programmer is using the index correctly. Direct fireChannelRead cannot be guaranteed

trustin:
```java
Object decode(ChannelHandlerContext ctx, ByteBuf in)
```

This prohibits a decoder from producing more than one message in a single decode call. It's possible that a single protocol-level frame produces more than one application-level message in reality.

carl-mastrangelo:
@trustin What would you think about the decode() method just letting subclasses call ctx.fireChannelRead directly?

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -400,45 +394,29 @@ private void channelInputClosed(ChannelHandlerContext ctx, boolean callChannelIn
      * Called when the input of the channel was closed which may be because it changed to inactive or because of
      * {@link ChannelInputShutdownEvent}.
      */
-    void channelInputClosed(ChannelHandlerContext ctx, List<Object> out) throws Exception {
+    void channelInputClosed(ByteToMessageDecoderContext ctx) throws Exception {
```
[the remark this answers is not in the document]
normanmaurer: will do a followup

normanmaurer:
Once this pass the CI I will merge it to make some progress.

normanmaurer:
This is finally in...

doom369:
@normanmaurer by the way, I already adapted that code and it is used already in the production. No issues so far :). Great job, Norman!

Issue #3998, opened 2015-07-17
HTTP/2 non flow controlled frames ingest limits


Scottmitch:
Flow control allows us to limit the amount of data we ingest for flow controlled frames.  However for frames that are not subject to flow control we do not have any such limits for the amount of data we ingest. Should we have some controls or limits for the amount of data we ingest for non flow controlled frames?  For example a simplistic peer (or many peers) that has a large amount of send bandwidth could (D)DoS us with control frames (i.e. `PING`) and never reading any responses.

Scottmitch:
@ejona86 - You previously pointed out (https://github.com/netty/netty/issues/3783#issuecomment-105704927) that `auto_read` must be on, but I'm wondering what other options we have to protect against traffic from control frames.  I guess limiting what we read, or potentially closing the connection are the 2 options that come to mind. WDYT?

ejona86:
@Scottmitch, hmm... Okay, let me think through this aloud.

If we have a pretty full write queue (of control messages), let's say we stop reading. This isn't discernible from the HTTP/2 semantic level because anything we sent would be queued anyway. So it would only impact the peer's ability to send on the TCP connection.

However, there could still be sent frames en route to the peer, including something like PING. When the peer then tries to ack the PING it could determine its queue is too full and stop reading. Deadlock.

What would fill up the write queue though, in "normal" operation? HEADERS could easily do it if issued many requests at once, especially since a naïve implementation would queue such a frame immediately, independent of the outbound buffer. That is only in the client->server direction though.

How about server->client? I can't come up with something offhand. A bunch of WINDOW_UPDATES for different streams all at once seems highly unlikely. Maybe the server is overloaded and times out a bunch of streams with RST_STREAM? Doesn't seem like enough. Maybe just stopping reading does work... although deadlocks are hard to predict many times.

The HTTP/2 spec does permit detecting "suspicious" behavior and aborting the connection:
> An endpoint MAY treat activity that is suspicious as a connection error (Section 5.4.1) of type ENHANCE_YOUR_CALM.

Maybe we limit the number of pending PING + SETTING acks to something "insane" like "100" or "1000"? (When choosing the number, we do have to consider internal fragmentation of ByteBuffers causing more memory usage than the raw written byte count.) If it is exceeded, we fail with ENHANCE_YOUR_CALM.

[...]

PR #8632, opened 2018-12-06
Correct respect auto read configuration when using HTTP/2

normanmaurer:
Motivation:

We did not correctly respect auto read when HTTP/2 is used. This was due the fact that we extended ByteToMessageDecoder but not use the out List to put decoded messages in. This lead to the situation that we will trigger a ctx.read() in channelReadComplete() as we assume we was not able to decode any data.

Modifications:

- Correctly keep track of if we were able to decode a frame and only if not trigger a read if auto read is off.
- Fix handling in Http2FrameCodec as well
- Add unit test.

Result:

Correctly respect auto read also when using HTTP2

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
@@ -79,6 +79,7 @@ public class Http2ConnectionHandler extends ByteToMessageDecoder implements Http
     private ChannelFutureListener closeListener;
     private BaseDecoder byteDecoder;
     private long gracefulShutdownTimeoutMillis;
+    private AutoReadFrameListener frameListener;
 
     protected Http2ConnectionHandler(Http2ConnectionDecoder decoder, Http2ConnectionEncoder encoder,
                                      Http2Settings initialSettings) {
@@ -401,6 +402,10 @@ public class Http2ConnectionHandler extends ByteToMessageDecoder implements Http
         decoder.lifecycleManager(this);
         encoder.flowController().channelHandlerContext(ctx);
         decoder.flowController().channelHandlerContext(ctx);
+
+        // Ensure we can keep track of if we need to trigger read or not.
+        frameListener = new AutoReadFrameListener(decoder.frameListener());
+        decoder.frameListener(frameListener);
         byteDecoder = new PrefaceDecoder(ctx);
     }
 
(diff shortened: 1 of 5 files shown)


normanmaurer:
This was reported by @rschmitt here https://github.com/netty/netty/pull/8617#discussion_r238801970.

This is the first part of the fix to correctly handles it for `Http2ConnectHandler` and `Http2FrameCodec`. Once this is merged I will look into `Http2MultiplexCodec`

normanmaurer:
@netty-bot test this please

normanmaurer:
Actually @Scottmitch pointed out that it may be the correct to just keep on reading:

https://tools.ietf.org/html/rfc7540#section-5.2.2
> Even with full awareness of the current bandwidth-delay product,
implementation of flow control can be difficult.  When using flow
control, the receiver MUST read from the TCP receive buffer in a
timely fashion. Failure to do so could lead to a deadlock when
critical frames, such as WINDOW_UPDATE, are not read and acted upon.

normanmaurer:
Also related to https://github.com/netty/netty/issues/3998

normanmaurer:
That said I think we may still want to correctly support AUTO_READ to give the user the most flexibility. There is nothing wrong per se supporting it and give the user the flexibility. Its up to the user of-course to ensure reads are done.

normanmaurer:
@netty-bot test this please

normanmaurer:
@carl-mastrangelo  PTAL again.. I also fixed auto read for the multiplex codec. 

@rschmitt this should fix your problem completely. 

I am now working on some more unit tests.

rschmitt:
@bryce-anderson @normanmaurer Is the idea to shut off auto-read for child streams only? That's what would make the most sense to me, given HTTP/2's design.

bryce-anderson:
@rschmitt, my understanding of this commit was that it was for the socket pipeline. I completely get the need for supporting `autoRead(false)` in the stream channels (it's a beautiful way to manage flow control) but don't know of a reason that would motivate the added complexity necessary to support manually driving the HTTP/2 session itself.

rschmitt:
Neither do I; I just want to make sure my understanding is correct.

normanmaurer:
@rschmitt @bryce-anderson so you are saying its fine to NOT support `AUTO_READ` false for the "parent channel (socket)" in this case ? I am not strong on this PR, just noticed it while looking at some  PR. 

If we do not want to support AUTO_READ on the parent we should at least make it explicit. At the moment it just works like this by chance (a side product of implementation details in ByteToMessageDecoder and Http2ConnectionHandler). 

WDYT ?

normanmaurer:
Also /cc @trustin @carl-mastrangelo @ejona86

bryce-anderson:
@normanmaurer, explicitly not supporting `AUTO_READ=false` would be my preference, but I'm open to being educated about a legitimate reason to stall the socket of an HTTP/2 session.

normanmaurer:
this would work for me... @rschmitt WDYT ?

rschmitt:
@normanmaurer It works for me too, I also can't think of a reason to stall the parent channel (as long as it will behave correctly when the stream channels are not auto-reading).

normanmaurer:
@rschmitt if the child channels will not read we will not "give back" to the flow-controller and so the HTTP2 flow-controlling should do the right thing here.

normanmaurer:
Closing this in favour of https://github.com/netty/netty/pull/8647

PR #8633, opened 2018-12-06
More correct fix for using ChannelInitializer with custom EventExecutor.

normanmaurer:
Motivation:

8331248671b9c0ea07cf8dbdfa5d8d2f89fdf459 did make some changes to fix a race in ChannelInitializer when using with a custom EventExecutor. Unfortunally these where a bit racy and so the testcase failed sometimes.

Modifications:

- More correct fix when using a custom EventExecutor
- Adjust the testcase to be more correct.

Result:

Proper fix for https://github.com/netty/netty/issues/8616.

### DIFF
--- a/transport/src/main/java/io/netty/channel/ChannelInitializer.java
+++ b/transport/src/main/java/io/netty/channel/ChannelInitializer.java
@@ -79,6 +79,9 @@ public abstract class ChannelInitializer<C extends Channel> extends ChannelInbou
             // we called initChannel(...) so we need to call now pipeline.fireChannelRegistered() to ensure we not
             // miss an event.
             ctx.pipeline().fireChannelRegistered();
+
+            // We are done with init the Channel, removing the initializer now.
+            remove(ctx);
         } else {
             // Called initChannel(...) before which is the expected behavior, so just forward the event.
             ctx.fireChannelRegistered();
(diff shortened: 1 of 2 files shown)


transport/src/main/java/io/netty/channel/ChannelInitializer.java
```diff
@@ -125,32 +132,29 @@ private boolean initChannel(ChannelHandlerContext ctx) throws Exception {
...
+                if (pipeline.context(this) != null) {
+                    pipeline.remove(this);
+                }
             }
             return true;
         }
         return false;
     }
 
     private void remove(final ChannelHandlerContext ctx) {
```
ejona86: This is a confusing name now. Call it `removeState` or `clearState`?

transport/src/main/java/io/netty/channel/ChannelInitializer.java
```diff
@@ -79,6 +79,9 @@ public final void channelRegistered(ChannelHandlerContext ctx) throws Exception
             // we called initChannel(...) so we need to call now pipeline.fireChannelRegistered() to ensure we not
             // miss an event.
             ctx.pipeline().fireChannelRegistered();
+
+            // We are done with init the Channel, removing the initializer now.
```
ejona86: This comment seems to be wrong. The initializer was already removed in initChannel.

normanmaurer:
@ejona86 addressed

Issue #8535, opened 2018-11-13
[Netty 5] Remove mark indices support from ByteBuf 


normanmaurer:
`ByteBuf` supports “marker indexes”. The intended use case for these is if a speculative operation (e.g. decode) is in process the user can “mark” and interface and refer to it later if the operation isn’t successful (e.g. not enough data). However this is rarely used in practice, 
requires extra memory to maintain, and introduces complexity in the state management for derived/pooled buffer initialization, resizing, and other operations which may modify reader/writer indexes. 

We should drop the support of marking indices.

johnou:
@normanmaurer so if you had some code like this that reads ahead when decoding

```java
      case READ_SENDER_IP:
        if (buffer.readableBytes() < BYTE_SIZE) {
          return;
        }
        buffer.markReaderIndex();
        int octetsLength = buffer.readByte();
        if (buffer.readableBytes() < octetsLength) {
          buffer.resetReaderIndex();
          return;
        }

        byte[] octets = new byte[octetsLength];
        buffer.readBytes(octets);
        senderIp = InetAddress.getByAddress(octets);
```

I guess one could do something like this..

```java
      case READ_SENDER_IP:
        if (buffer.readableBytes() < BYTE_SIZE) {
          return;
        }
        //buffer.markReaderIndex();
        int octetsLength = buffer.getByte(buffer.readerIndex()); // get instead of read to avoid moving the reader index
        if (buffer.readableBytes() < octetsLength) {
          //buffer.resetReaderIndex();
          return;
        }

        byte[] octets = new byte[octetsLength];
        buffer.readBytes(octets);
        senderIp = InetAddress.getByAddress(octets);
        currentState = DecoderState.READ_SENDER_PORT;
```

normanmaurer:
@johnou yes or just store the original readerIndex and set it again if needed.

ejona86:
Worst-case here would could have an API that returns a "mark object" (with the saved index) that could then be "restored". But doing it manually is likely okay.

PR #8636, opened 2018-12-07
Remove support for marking reader and writerIndex in ByteBuf to reduc…

normanmaurer:
…e overhead and complexity.

Motivation:

ByteBuf supports “marker indexes”. The intended use case for these is if a speculative operation (e.g. decode) is in process the user can “mark” and interface and refer to it later if the operation isn’t successful (e.g. not enough data). However this is rarely used in practice,
requires extra memory to maintain, and introduces complexity in the state management for derived/pooled buffer initialization, resizing, and other operations which may modify reader/writer indexes.

Modifications:

Remove support for marking and adjust testcases / code.

Result:

Fixes https://github.com/netty/netty/issues/8535.


normanmaurer:
Merged... thanks

PR #8638, opened 2018-12-07
Fix typo in MessageToMessageDecoder api docs.

normanmaurer:
Motivation:

We had some typo (most likely caused by copy-and-paste) in the api docs which should be fixed.

Modifications:

Replace encoder by decoder word.

Result:

Correct apidocs.


normanmaurer:
noticed while working on some netty 5 stuff

PR #8639, opened 2018-12-07
Fix incorrect assert in Http2MultiplexCodec caused by 9f9aa1a.

normanmaurer:
Motivation:

9f9aa1a did some changes related to fixing how we handle ctx.read() in child channel but did incorrectly change some assert.

Modifications:

Fix assert to be correct.

Result:

Code does not throw an AssertionError due incorrect assert check.


normanmaurer:
@bryce-anderson great catch!

normanmaurer:
@bryce-anderson I am trying to come-up with a test.

normanmaurer:
@bryce-anderson how we test the `Http2MultiplexCodec` and `FrameCodec` is a bit unfortunate :( I think we should rewrite the tests to be less "hacky".

Let me pull this in first and then work on some rewriting of the tests.

bryce-anderson:
@normanmaurer, I'm dealing with some of my own difficult to test stuff so I feel your pain. 😄 Your plan sounds good to me.

Issue #2235, opened 2014-02-15
Static file server example: Forget to close the RandomAccessFile?


ngocdaothanh:
https://github.com/netty/netty/blob/master/example/src/main/java/io/netty/example/http/file/HttpStaticFileServerHandler.java

Should the `raf` be closed after the file has been sent?

trustin:
@lukehutch It should be closed automatically when the write attempt is done (either success or failure). If not, please let us know.

PR #8643, opened 2018-12-09
Fix Example HttpStaticFileServerHandler: raf is not closed closes #8604

### LINKED ISSUE #8604: Example HttpStaticFileServerHandler: raf is not closed.
Recently I use this HttpStaticFileServerHandler as file server, but I can't download mp4 file correctly. Then I found raf is not closed.

There is a issue mention about it: https://github.com/netty/netty/issues/2235, but it seems don't fix it.

I asked about it at stackoverflow: https://stackoverflow.com/questions/53500128/netty-httpstaticfileserverhandler-problem/53511572#53511572

Here is my change,and it's working.

    RandomAccessFile raf;
        try {
            raf = new RandomAccessFile(file, "r");
        } catch (FileNotFoundException ignore) {
            sendError(ctx, NOT_FOUND);
            return;
        }

    ...

    sendFileFuture.addListener(new ChannelProgressiveFutureListener() {
            @Override
            public void operationProgressed(ChannelProgressiveFuture future, long progress, long total) 
    {
                if (total < 0) { // total unknown
                    System.err.println(future.channel() + " Transfer progress: " + progress);
                } else {
                    System.err.println(future.channel() + " Transfer progress: " + progress + " / " + total);
                }
            }

            @Override
            public void operationComplete(ChannelProgressiveFuture future) {
                System.err.println(future.channel() + " Transfer complete.");
                raf.close();// close raf when transfer completed
            }
        });

samaitra:
Motivation:

Fixes the issue as mentioned below

Recently I use this HttpStaticFileServerHandler as file server, but I can't download mp4 file correctly. Then I found raf is not closed.

There is a issue mention about it: #2235, but it seems don't fix it.

I asked about it at stackoverflow: https://stackoverflow.com/questions/53500128/netty-httpstaticfileserverhandler-problem/53511572#53511572

Modifications:

made raf as final
added raf.close() call inside try/catch block

Result:

After this change the RandomAccessFile handler will be closed

Motivation:

Modification:

Result:

Fixes #<GitHub issue number>. 

### DIFF
--- a/example/src/main/java/io/netty/example/http/file/HttpStaticFileServerHandler.java
+++ b/example/src/main/java/io/netty/example/http/file/HttpStaticFileServerHandler.java
@@ -40,6 +40,7 @@ import io.netty.handler.stream.ChunkedFile;
 import io.netty.util.CharsetUtil;
 import io.netty.util.internal.SystemPropertyUtil;
 
+import java.io.IOException;
 import javax.activation.MimetypesFileTypeMap;
 import java.io.File;
 import java.io.FileNotFoundException;
@@ -165,7 +166,7 @@ public class HttpStaticFileServerHandler extends SimpleChannelInboundHandler<Ful
             }
         }
 
-        RandomAccessFile raf;
+        final RandomAccessFile raf;
         try {
             raf = new RandomAccessFile(file, "r");
         } catch (FileNotFoundException ignore) {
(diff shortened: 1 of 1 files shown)


normanmaurer:
@samaitra is this over SSL or not ?

samaitra:
@normanmaurer This is not over SSL. The issue is fixed in `HttpStaticFileServerHandler`

normanmaurer:
@samaitra I really dont understand why this should fix anything... Do you have a reproducer that I can run that fails before the fix and works after it ?

normanmaurer:
Without a reproducer I am not feeling comfortable pulling this in :/

samaitra:
@normanmaurer Thank you for your feedback. yes, it is not consistently reproducible. I will await further instructions from @yao-qingdong for steps to reproduce.

If you want me to close the PR I can close.

PR #8647, opened 2018-12-11
Explict always call ctx.read() when AUTO_READ is false and HTTP/2 is …

normanmaurer:
…used.

Motivation:

We should always call ctx.read() even when AUTO_READ is false as flow-control is enforced by the HTTP/2 protocol.

See also https://tools.ietf.org/html/rfc7540#section-5.2.2.

We already did this before but not explicit and only did so because of some implementation details of ByteToMessageDecoder. It's better to be explicit here to not risk of breakage later on.

Modifications:

- Ensure we always call ctx.read() when AUTO_READ is false
- Add unit test.

Result:

No risk of staling the connection when HTTP/2 is used.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
@@ -527,8 +527,18 @@ public class Http2ConnectionHandler extends ByteToMessageDecoder implements Http
         }
     }
 
-    void channelReadComplete0(ChannelHandlerContext ctx) throws Exception {
-        super.channelReadComplete(ctx);
+    final void channelReadComplete0(ChannelHandlerContext ctx) {
+        // Discard bytes of the cumulation buffer if needed.
+        discardSomeReadBytes();
+
+        // Ensure we never stale the HTTP/2 Channel. Flow-control is enforced by HTTP/2.
+        //
+        // See https://tools.ietf.org/html/rfc7540#section-5.2.2
+        if (!ctx.channel().config().isAutoRead()) {
+            ctx.read();
+        }
+
+        ctx.fireChannelReadComplete();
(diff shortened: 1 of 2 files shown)


normanmaurer:
Also @rschmitt

codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionHandler.java
```diff
@@ -527,8 +527,18 @@ public void channelReadComplete(ChannelHandlerContext ctx) throws Exception {
         }
     }
 
-    void channelReadComplete0(ChannelHandlerContext ctx) throws Exception {
-        super.channelReadComplete(ctx);
+    final void channelReadComplete0(ChannelHandlerContext ctx) {
+        // Discard bytes of the cumulation buffer if needed.
+        discardSomeReadBytes();
+
+        // Ensure we never stale the HTTP/2 Channel. Flow-control is enforced by HTTP/2.
```
ejona86: s/stale/stall/

ejona86:
I'm not quite sure I agree with this change. At the very least I don't like the idea that we can never undo this change, as it would break users whose channel has AUTO_READ==false. I would rather we "complain loudly" (exception caught?) if AUTO_READ is off instead of faking the setting.

normanmaurer:
@ejona86 hmm... it’s exactly the same what we do atm l. We just do it exploit now and not “depend” on the base class to do so

ejona86:
Oh, I see. Because we weren't using the list. Right, right, right.

normanmaurer:
@ejona86 so all good except the typo in the comment ?

ejona86:
Yes.

normanmaurer:
@ejona86 fixed by a followup commit (f8ec368eab6f7d0f028f21f29f9b5521c699f4a7)

PR #8649, opened 2018-12-12
Add test for correctly handling SSLSessionBindingEvent when acting on th…

normanmaurer:
…e SSLSession.

Motivation:

During some other work I noticed we do not have any tests to ensure we correctly use SSLSessionBindingEvent. We should add some testing.

Modifications:

- Added unit test to verify we correctly implement it.
- Ignore the test when using Conscrypt as it not correctly implements it.

Result:

More tests for custom SSL impl.

### DIFF
--- a/handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
@@ -85,6 +85,8 @@ import javax.net.ssl.SSLHandshakeException;
 import javax.net.ssl.SSLParameters;
 import javax.net.ssl.SSLPeerUnverifiedException;
 import javax.net.ssl.SSLSession;
+import javax.net.ssl.SSLSessionBindingEvent;
+import javax.net.ssl.SSLSessionBindingListener;
 import javax.net.ssl.SSLSocketFactory;
 import javax.net.ssl.TrustManager;
 import javax.net.ssl.TrustManagerFactory;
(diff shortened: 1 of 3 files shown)


normanmaurer:
@ejona86 @flooey there seems to be a bug in conscrypt as it uses the incorrect SSLSession object when creating the events. You can see this when enable the test for conscrypt.

flooey:
@normanmaurer Yep, looks like we need to handle wrapped sessions properly there.  I've filed google/conscrypt#593 to track it.

normanmaurer:
@netty-bot test this please

normanmaurer:
Let me merge this as its just a unit test ;)

handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
```diff
@@ -2729,6 +2731,86 @@ public void testGetCiphersuite() throws Exception {
...
+            cleanupClientSslEngine(clientEngine);
+            cleanupServerSslEngine(serverEngine);
+            ssc.delete();
+        }
+    }
+
+    private static void assertSSLSessionBindingEventValue(
+            String name, SSLSession session, SSLSessionBindingEvent event) {
+        assertEquals(name, event.getName());
+        assertEquals(session, event.getSession());
```
carl-mastrangelo: session and source are the same?

normanmaurer: @carl-mastrangelo yes... are you asking if we should better use `assertSame(...)` ?

carl-mastrangelo: Oh, no, I just thought this was a copy paste error, and perhaps it should have been event.getSource().    If its correct, then let's leave it alone.

PR #8651, opened 2018-12-12
Added "UTF-8 only" comments to three decoders.

alex-vas:
Added comments to LineBasedFrameDecoder, JsonObjectDecoder and XmlFrameDecoder that they are only compatible with UTF-8 encoded streams.

No code changes, therefore no tests.

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/LineBasedFrameDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/LineBasedFrameDecoder.java
@@ -26,7 +26,7 @@ import java.util.List;
  * <p>
  * Both {@code "\n"} and {@code "\r\n"} are handled.
  * <p>
- * The byte stream is expected to be in UTF-8 character encoding or ASCII, as current implementation 
+ * The byte stream is expected to be in UTF-8 character encoding or ASCII, as current implementation
  * uses direct {@code byte} to {@code char} cast.
  * <p>
  * For a more general delimiter-based decoder, see {@link DelimiterBasedFrameDecoder}.
--- a/codec/src/main/java/io/netty/handler/codec/LineBasedFrameDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/LineBasedFrameDecoder.java
@@ -25,6 +25,10 @@ import java.util.List;
  * A decoder that splits the received {@link ByteBuf}s on line endings.
  * <p>
  * Both {@code "\n"} and {@code "\r\n"} are handled.
+ * <p>
+ * The byte stream is expected to be in UTF-8 character encoding or ASCII, as current implementation 
+ * uses direct {@code byte} to {@code char} cast.
+ * <p>
  * For a more general delimiter-based decoder, see {@link DelimiterBasedFrameDecoder}.
  */
 public class LineBasedFrameDecoder extends ByteToMessageDecoder {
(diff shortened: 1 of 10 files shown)


normanmaurer:
@netty-bot test this please

normanmaurer:
@alex-vas can please fix the check style errors:

```
07:00:33 [INFO] Starting audit...
07:00:33 /code/codec/src/main/java/io/netty/handler/codec/json/JsonObjectDecoder.java:34: trailing whitespace
07:00:33 /code/codec/src/main/java/io/netty/handler/codec/LineBasedFrameDecoder.java:29: trailing whitespace
07:00:33 /code/codec/src/main/java/io/netty/handler/codec/xml/XmlFrameDecoder.java:63: trailing whitespace
07:00:33 Audit done.
```

codec/src/main/java/io/netty/handler/codec/LineBasedFrameDecoder.java
```diff
@@ -25,6 +25,10 @@
  * A decoder that splits the received {@link ByteBuf}s on line endings.
  * <p>
  * Both {@code "\n"} and {@code "\r\n"} are handled.
+ * <p>
+ * The byte stream is expected to be in UTF-8 character encoding or ASCII, as current implementation
+ * uses direct {@code byte} to {@code char} cast.
+ * <p>
```
jvirtanen: `byte` to `char` cast sounds a bit misleading in the context of UTF-8, which is a [variable-length encoding](https://tools.ietf.org/html/rfc3629#section-4). Though on a quick glance it looks like all of the decoders only operate on characters which do have the same one-byte representation in ASCII and UTF-8.

alex-vas: That's right. All decoders are using low range (<=0x7F) codepoints (e.g., "<>/="\r\n{}[]:") in their implementations, and in UTF-8 the same bytes are not appear in second, third or fourth byte of a multibyte codepoint representation. See [URT-8](https://en.wikipedia.org/wiki/UTF-8) description. So all decoders are going to handle UTF-8 and ASCII properly. 

Perhaps the comment could be further clarified by going "as current implementation uses direct `code` to `byte` cast and then compares it with low range ASCII characters like `<` `>`, `\r` `\n`, etc...

Or we can remove the reference to cast at all... Though I see some value in having this particular implementation disclosure.

Up to you.

jvirtanen: I think your clarification makes sense. :+1: It now also accounts for other ASCII compatible character sets than just UTF-8 (e.g. [ISO 8859](https://en.wikipedia.org/wiki/ISO_8859)).

normanmaurer:
@netty-bot test this please

normanmaurer:
@alex-vas can you squash and update the commit message to match our template please:

https://netty.io/wiki/writing-a-commit-message.html

alex-vas:
Not sure I understand the "squash", but I have updated the last comment.

normanmaurer:
@alex-vas looks good! can you just also sign our ICLA: https://netty.io/s/icla ?

alex-vas:
> @alex-vas looks good! can you just also sign our ICLA: https://netty.io/s/icla ?

Done.

normanmaurer:
@alex-vas thanks a lot... merged.

PR #8653, opened 2018-12-13
Introduce TlsMasterKeyHandler

### LINKED ISSUE #8663: Allow easier TLS debugging by Wireshark
Debugging SSL/TLS connections through Wireshark is a pain -- if the cipher used involves Diffie-Hellman then it is essentially impossible unless you can have the client dump out the master key [1]

It would be a great addition to allow Netty to dump the _master key_ for the TLS session, especially in a format that Wireshark can read such as NSS key log format [2]

The scenario in which this would be helpful:
I am terminating TLS via netty on my server. I would like to take a tcpdump and analyze it with Wireshark. If the Cipher exchange is Diffie-Hellman, this is impossible, even with the private key.

[1] https://jimshaver.net/2015/02/11/decrypting-tls-browser-traffic-with-wireshark-the-easy-way/
[2] https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/Key_Log_Format

fzakaria:
### Motivation

Debugging SSL/TLS connections through wireshark is a pain -- if the cipher used involves Diffie-Hellman then it is essentially impossible unless you can have the client dump out the master key [1]

This is a work-in-progress change (tests & comments to come!) that introduces a new handler you can set on the SslContext to receive the master key & session id. I'm hoping to get feedback if a change in this vein would be welcomed.

An implementation that conforms to Wireshark's NSS key log[2] file is also included.

Depending on feedback on the PR going forward I am planning to "clean it up" by adding documentation, example server & tests. Implementation will need to be finished as well for retrieving the master key from the OpenSSL context.

[1] https://jimshaver.net/2015/02/11/decrypting-tls-browser-traffic-with-wireshark-the-easy-way/
[2] https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/Key_Log_Format

### Modification

I've wired up a new functional handler that gets passed eventually to SSLHandler to be called with the SSL session id & master key.

An implementation of the handler that conforms to Wireshark's key log format is included.

### Result

I was able to set the handler to log to a file, which when provided to Wireshark, decrypted the TLS flow.

I did a quick test by modifying _HttpHelloWorldServer_

```java
            sslCtx = SslContextBuilder
                    .forServer(ssc.certificate(), ssc.privateKey())
                    .sslProvider(SslProvider.JDK)
                    .sslMasterKeyHandler(TlsMasterKeyHandler.WiresharkTlsMasterKeyHandler.INSTANCE)
                    .build();
```

```bash_
cat ~/key.log
RSA Session-ID:5c11f6c920488f4dc9891771ae001bd645a16791ab47b337f89136e0deba3cdb Master-Key:6a3aed11200f26d115de4fb7c6aaa01222532520592e93b45ae1e243d4ec06099f4057f3f75616ea6e741c9318f36a30
```

Fixes #8663 

screenshot 2018-12-12 22 36 55
screenshot 2018-12-12 22 37 06

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SslMasterKeyHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslMasterKeyHandler.java
@@ -1,0 +1,190 @@
+/*
+ * Copyright 2019 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.ssl;
+
+import io.netty.buffer.ByteBufUtil;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.internal.tcnative.SSL;
+import io.netty.util.internal.ReflectionUtil;
+import io.netty.util.internal.SystemPropertyUtil;
+import io.netty.util.internal.logging.InternalLogger;
+import io.netty.util.internal.logging.InternalLoggerFactory;
+
+import javax.crypto.SecretKey;
+import javax.crypto.spec.SecretKeySpec;
+import javax.net.ssl.SSLEngine;
+import javax.net.ssl.SSLSession;
+import java.lang.reflect.Field;
+
+/**
+ * The {@link SslMasterKeyHandler} is a channel-handler you can include in your pipeline to consume the master key
+ * & session identifier for a TLS session.
+ * This can be very useful, for instance the {@link WiresharkSslMasterKeyHandler} implementation will
+ * log the secret & identifier in a format that is consumable by Wireshark -- allowing easy decryption of pcap/tcpdumps.
+ */
+public abstract class SslMasterKeyHandler extends ChannelInboundHandlerAdapter {
+
+    private static final InternalLogger logger = InternalLoggerFactory.getInstance(SslMasterKeyHandler.class);
+
+    /**
+     * The JRE SSLSessionImpl cannot be imported
+     */
+    private static final Class<?> SSL_SESSIONIMPL_CLASS;
+
+    /**
+     * The master key field in the SSLSessionImpl
+     */
+    private static final Field SSL_SESSIONIMPL_MASTER_SECRET_FIELD;
+
+    /**
+     * A system property that can be used to turn on/off the {@link SslMasterKeyHandler} dynamically without having
+     * to edit your pipeline.
+     * <code>-Dio.netty.ssl.masterKeyHandler=true</code>
+     */
+    public static final String SYSTEM_PROP_KEY = "io.netty.ssl.masterKeyHandler";
+
+    /**
+     * The unavailability cause of whether the private Sun implementation of SSLSessionImpl is available.
+     */
+    private static final Throwable UNAVAILABILITY_CAUSE;
+
+    static {
+        Throwable cause = null;
+        Class<?> clazz = null;
+        Field field = null;
+        try {
+            clazz = Class.forName("sun.security.ssl.SSLSessionImpl");
+            field = clazz.getDeclaredField("masterSecret");
+            cause = ReflectionUtil.trySetAccessible(field, true);
+        } catch (Throwable e) {
+            cause = e;
+            logger.debug("sun.security.ssl.SSLSessionImpl is unavailable.", e);
+        }
+        UNAVAILABILITY_CAUSE = cause;
+        SSL_SESSIONIMPL_CLASS = clazz;
+        SSL_SESSIONIMPL_MASTER_SECRET_FIELD = field;
+    }
+
+    /**
+     * Constructor.
+    */
+    protected SslMasterKeyHandler() {
+    }
+
+    /**
+     * Ensure that SSLSessionImpl is available.
+     * @throws UnsatisfiedLinkError if unavailable
+     */
+    public static void ensureSunSslEngineAvailability() {
+        if (UNAVAILABILITY_CAUSE != null) {
+            throw new IllegalStateException(
+                    "Failed to find SSLSessionImpl on classpath", UNAVAILABILITY_CAUSE);
+        }
+    }
+
+    /**
+     * Returns the cause of unavailability.
+     *
+     * @return the cause if unavailable. {@code null} if available.
+     */
+    public static Throwable sunSslEngineUnavailabilityCause() {
+        return UNAVAILABILITY_CAUSE;
+    }
+
+    /* Returns {@code true} if and only if sun.security.ssl.SSLSessionImpl exists in the runtime.
+     */
+    public static boolean isSunSslEngineAvailable() {
+        return UNAVAILABILITY_CAUSE == null;
+    }
+
+    /**
+     * Consume the master key for the session and the sessionId
+     * @param masterKey A 48-byte secret shared between the client and server.
+     * @param session The current TLS session
+     */
+    protected abstract void accept(SecretKey masterKey, SSLSession session);
+
+    @Override
+    public final void userEventTriggered(ChannelHandlerContext ctx, Object evt) {
+        //only try to log the session info if the ssl handshake has successfully completed.
+        if (evt == SslHandshakeCompletionEvent.SUCCESS) {
+            boolean shouldHandle = SystemPropertyUtil.getBoolean(SYSTEM_PROP_KEY, false);
+
+            if (shouldHandle) {
+                final SslHandler handler = ctx.pipeline().get(SslHandler.class);
+                final SSLEngine engine = handler.engine();
+                final SSLSession sslSession = engine.getSession();
+
+                //the OpenJDK does not expose a way to get the master secret, so try to use reflection to get it.
+                if (isSunSslEngineAvailable() && sslSession.getClass().equals(SSL_SESSIONIMPL_CLASS)) {
+                    final SecretKey secretKey;
+                    try {
+                        secretKey = (SecretKey) SSL_SESSIONIMPL_MASTER_SECRET_FIELD.get(sslSession);
+                    } catch (IllegalAccessException e) {
+                        throw new IllegalArgumentException("Failed to access the field 'masterSecret' " +
+                                "via reflection.", e);
+                    }
+                    accept(secretKey, sslSession);
+                } else if (OpenSsl.isAvailable() && engine instanceof ReferenceCountedOpenSslEngine) {
+                    SecretKeySpec secretKey = new SecretKeySpec(
+                            SSL.getMasterKey(((ReferenceCountedOpenSslEngine) engine).sslPointer()), "AES");
+                    accept(secretKey, sslSession);
+                }
+            }
+        }
+
+        ctx.fireUserEventTriggered(evt);
+    }
+
+    /**
+     * Create a {@link WiresharkSslMasterKeyHandler} instance.
+     * This TLS master key handler logs the master key and session-id in a format
+     * understood by Wireshark -- this can be especially useful if you need to ever
+     * decrypt a TLS session and are using perfect forward secrecy (i.e. Diffie-Hellman)
+     * The key and session identifier are forwarded to the log named 'io.netty.wireshark'.
+     */
+    public static SslMasterKeyHandler newWireSharkSslMasterKeyHandler() {
+        return new WiresharkSslMasterKeyHandler();
+    }
+
+    /**
+     * Record the session identifier and master key to the {@link InternalLogger} named <code>io.netty.wireshark</code>.
+     * ex. <code>RSA Session-ID:XXX Master-Key:YYY</code>
+     * This format is understood by Wireshark 1.6.0.
+     * https://code.wireshark.org/review/gitweb?p=wireshark.git;a=commit;h=686d4cabb41185591c361f9ec6b709034317144b
+     * The key and session identifier are forwarded to the log named 'io.netty.wireshark'.
+     */
+    private static final class WiresharkSslMasterKeyHandler extends SslMasterKeyHandler {
+
+        private static final InternalLogger wireshark_logger =
+                InternalLoggerFactory.getInstance("io.netty.wireshark");
+
+        private static final char[] hexCode = "0123456789ABCDEF".toCharArray();
+
+        @Override
+        protected void accept(SecretKey masterKey, SSLSession session) {
+            if (masterKey.getEncoded().length != 48) {
+                throw new IllegalArgumentException("An invalid length master key was provided.");
+            }
+            final byte[] sessionId = session.getId();
+            wireshark_logger.warn("RSA Session-ID:{} Master-Key:{}",
+                    ByteBufUtil.hexDump(sessionId).toLowerCase(),
+                    ByteBufUtil.hexDump(masterKey.getEncoded()).toLowerCase());
+        }
+    }
+
+}
--- a/handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
@@ -46,6 +46,8 @@ import io.netty.util.concurrent.Promise;
 import io.netty.util.internal.EmptyArrays;
 import io.netty.util.internal.PlatformDependent;
 import io.netty.util.internal.StringUtil;
+import io.netty.util.internal.SystemPropertyUtil;
+import org.conscrypt.OpenSSLProvider;
 import org.junit.After;
 import org.junit.Assume;
 import org.junit.Before;
@@ -55,10 +57,12 @@ import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
 
 import java.io.ByteArrayInputStream;
+import java.io.Closeable;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
 import java.net.InetSocketAddress;
 import java.net.Socket;
 import java.nio.ByteBuffer;
@@ -84,12 +88,14 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
+import javax.crypto.SecretKey;
 import javax.net.ssl.ExtendedSSLSession;
 import javax.net.ssl.KeyManager;
 import javax.net.ssl.KeyManagerFactory;
 import javax.net.ssl.KeyManagerFactorySpi;
 import javax.net.ssl.ManagerFactoryParameters;
 import javax.net.ssl.SNIHostName;
+import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLEngineResult;
 import javax.net.ssl.SSLEngineResult.Status;
@@ -3242,6 +3248,89 @@ public abstract class SSLEngineTest {
             cleanupClientSslEngine(clientEngine);
             cleanupServerSslEngine(serverEngine);
             ssc.delete();
+        }
+    }
+
+    @Test
+    public void testMasterKeyLogging() throws Exception {
+
+        /*
+         * At the moment master key logging is not supported for conscrypt
+         */
+        Assume.assumeFalse(serverSslContextProvider() instanceof OpenSSLProvider);
+
+        /*
+         * The JDK SSL engine master key retrieval relies on being able to set field access to true.
+         * That is not available in JDK9+
+         */
+        Assume.assumeFalse(sslServerProvider() == SslProvider.JDK && PlatformDependent.javaVersion() > 8);
+
+        String originalSystemPropertyValue = SystemPropertyUtil.get(SslMasterKeyHandler.SYSTEM_PROP_KEY);
+        System.setProperty(SslMasterKeyHandler.SYSTEM_PROP_KEY, Boolean.TRUE.toString());
+
+        SelfSignedCertificate ssc = new SelfSignedCertificate();
+        serverSslCtx = SslContextBuilder.forServer(ssc.certificate(), ssc.privateKey())
+                .sslProvider(sslServerProvider())
+                .sslContextProvider(serverSslContextProvider())
+                .build();
+        Socket socket = null;
+
+        try {
+            sb = new ServerBootstrap();
+            sb.group(new NioEventLoopGroup(), new NioEventLoopGroup());
+            sb.channel(NioServerSocketChannel.class);
+
+            final Promise<SecretKey> promise = sb.config().group().next().newPromise();
+            serverChannel = sb.childHandler(new ChannelInitializer<Channel>() {
+                @Override
+                protected void initChannel(Channel ch) throws Exception {
+                    ch.config().setAllocator(new TestByteBufAllocator(ch.config().getAllocator(), type));
+
+                    SslHandler sslHandler = delegatingExecutor == null ?
+                            serverSslCtx.newHandler(ch.alloc()) :
+                            serverSslCtx.newHandler(ch.alloc(), delegatingExecutor);
+
+                    ch.pipeline().addLast(sslHandler);
+                    ch.pipeline().addLast(new SslMasterKeyHandler() {
+                        @Override
+                        protected void accept(SecretKey masterKey, SSLSession session) {
+                            promise.setSuccess(masterKey);
+                        }
+                    });
+                    serverConnectedChannel = ch;
+                }
+            }).bind(new InetSocketAddress(0)).sync().channel();
+
+            int port = ((InetSocketAddress) serverChannel.localAddress()).getPort();
+
+            SSLContext sslContext = SSLContext.getInstance("TLS");
+            sslContext.init(null, InsecureTrustManagerFactory.INSTANCE.getTrustManagers(), null);
+            socket = sslContext.getSocketFactory().createSocket(NetUtil.LOCALHOST, port);
+            OutputStream out = socket.getOutputStream();
+            out.write(1);
+            out.flush();
+
+            assertTrue(promise.await(10, TimeUnit.SECONDS));
+            SecretKey key = promise.get();
+            assertEquals("AES secret key must be 48 bytes", 48, key.getEncoded().length);
+        } finally {
+            closeQuietly(socket);
+            if (originalSystemPropertyValue != null) {
+                System.setProperty(SslMasterKeyHandler.SYSTEM_PROP_KEY, originalSystemPropertyValue);
+            } else {
+                System.clearProperty(SslMasterKeyHandler.SYSTEM_PROP_KEY);
+            }
+            ssc.delete();
+        }
+    }
+
+    private static void closeQuietly(Closeable c) {
+        if (c != null) {
+            try {
+                c.close();
+            } catch (IOException ignore) {
+                // ignore
+            }
         }
     }
 


normanmaurer:
@fzakaria will check soon. Just a bit busy atm

fzakaria:
Sounds good @normanmaurer

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1536,12 +1570,30 @@ private void setHandshakeSuccess() {
         }
         ctx.fireUserEventTriggered(SslHandshakeCompletionEvent.SUCCESS);
 
+        if (TlsMasterKeyHandler.DoNothingTlsMasterKeyHandler.INSTANCE != masterKeyHandler) {
```
normanmaurer: @fzakaria after looking at the patch I think we should change it to just do it in a different `ChannelInboundHandler` and not adjust SslHandler at all.

Why not just do something like this:

```java
class TlsMasterKeyHandler extends ChannelInboundHandlerAdapter {

        @Override
        public void userEventTriggered(ChannelHandlerContext ctx, Object evt) {
            if (evt instanceof SslHandshakeCompletionEvent) {
                SslHandler handler = ctx.pipeline().get(SslHandler.class);
                SSLEngine engine = handler.engine();
                .....
               // Do all the logging here.
            }
            ctx.fireUserEventTriggered(evt);
        }
}
```

This way the user can just add the extra handler to the pipeline if interested in this... WDYT ?

fzakaria: I think that is a very good suggestion. I will update the pull-request.

fzakaria:
@normanmaurer thanks for the feedback.

What do you think overall about the motivation of the PR with respect to having it accepted? I'd like to propose having it submitting in two parts: support for JDK engine (the majority of this pull-request) and a latter commit that introduces it for OpenSSL
(I'm not sure yet on the work needed for netty-tcnative)

If that sounds good, I'd like to address your feedback + clean it up (add comments, tests etc..) to get it ready for approval.

normanmaurer:
@fzakaria this sounds good yes :)

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyBiConsumer.java
```diff
@@ -0,0 +1,70 @@
...
+import io.netty.util.internal.logging.InternalLoggerFactory;
+
+import javax.crypto.SecretKey;
+
+/**
+ * An interface to implement that will be called via {@link TlsMasterKeyHandler} after
+ * {@link SslHandshakeCompletionEvent} with the contents of the master key for the session.
+ * This can be extremely useful in <b>debug-only</b> scenarios if you ever want to decrypt the TLS traffic.
+ */
+public interface TlsMasterKeyBiConsumer {
```
normanmaurer: imho we could just keep things simple and not adding this interface and just provide protected methods in `TlsMasterKeyHandler` that people override and make the class abstract.

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,105 @@
+/*
+ * Copyright 2014 The Netty Project
```
normanmaurer: 2018

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     * A system property that can be used to turn on/off the {@link TlsMasterKeyHandler} dynamically without having
+     * to edit your pipeline.
+     * <code>-Dio.netty.tls.masterKeyHandler=true</code>
+     */
+    private static final String SYSTEM_PROP_KEY = "io.netty.tls.masterKeyHandler";
+
+    static {
+        try {
+            SSL_SESSIONIMPL_CLASS = Class.forName("sun.security.ssl.SSLSessionImpl");
```
normanmaurer: imho we should do something similar to what we do in `OpenSsl.java` in terms of init.

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,105 @@
...
+     */
+    public TlsMasterKeyHandler() {
+        this(TlsMasterKeyBiConsumer.DoNothingTlsMasterKeyBiConsumer.INSTANCE);
+    }
+
+    @Override
+    public void userEventTriggered(ChannelHandlerContext ctx, Object evt) {
+
+        //only try to log the session info if the ssl handshake has successfully completed.
+        if (evt instanceof SslHandshakeCompletionEvent && evt == SslHandshakeCompletionEvent.SUCCESS) {
```
normanmaurer: you can remove the `instanceof` as its already covered by `==`

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,105 @@
...
+
+            boolean shouldHandle = Boolean.getBoolean(SYSTEM_PROP_KEY);
```
normanmaurer: use `SystemPropertyUtil.`... also should we only do it once ?

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,105 @@
...
+                final SslHandler handler = ctx.pipeline().get(SslHandler.class);
+                final SSLEngine engine = handler.engine();
+                final SSLSession sslSession = engine.getSession();
+
+                //the OracleJVM does not expose a way to get the master secret, so try to use reflection to get it.
+                try {
+                    if (sslSession.getClass().equals(SSL_SESSIONIMPL_CLASS)) {
+                        Field masterSecretField = SSL_SESSIONIMPL_CLASS.getDeclaredField("masterSecret");
+                        masterSecretField.setAccessible(true);
+                        SecretKey secretKey = (SecretKey) masterSecretField.get(sslSession);
```
normanmaurer: Should we use `PlatformDependent` (Which will use usafe) to access it ? This way it will also work with jigsaw.

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,105 @@
...
+            if (shouldHandle && TlsMasterKeyBiConsumer.DoNothingTlsMasterKeyBiConsumer.INSTANCE != consumer) {
+                final SslHandler handler = ctx.pipeline().get(SslHandler.class);
+                final SSLEngine engine = handler.engine();
+                final SSLSession sslSession = engine.getSession();
+
+                //the OracleJVM does not expose a way to get the master secret, so try to use reflection to get it.
+                try {
+                    if (sslSession.getClass().equals(SSL_SESSIONIMPL_CLASS)) {
+                        Field masterSecretField = SSL_SESSIONIMPL_CLASS.getDeclaredField("masterSecret");
+                        masterSecretField.setAccessible(true);
```
normanmaurer: this only needs to be done once in a static block.

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
fzakaria: I believe I've done something similar now according to your feedback.

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,166 @@
...
+     *                  active or resumable session state.
+     */
+    protected abstract void accept(SecretKey masterKey, byte[] sessionId);
+
+    @Override
+    public void userEventTriggered(ChannelHandlerContext ctx, Object evt) {
+        //only try to log the session info if the ssl handshake has successfully completed.
+        if (evt == SslHandshakeCompletionEvent.SUCCESS) {
+
+            boolean shouldHandle = SystemPropertyUtil.getBoolean(SYSTEM_PROP_KEY, false);
```
normanmaurer: should re really lookup the property every time ?

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,166 @@
...
+
+            //if the handler is not an instance of DoNothingTlsMasterKeyBiConsumer try to handle it
+            if (shouldHandle) {
+                final SslHandler handler = ctx.pipeline().get(SslHandler.class);
+                final SSLEngine engine = handler.engine();
```
normanmaurer: check for `handler != null` just in case it was removed in the meantime

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,166 @@
...
+    /**
+     * Record the session identifier and master key to the {@link InternalLogger} named <code>io.netty.wireshark</code>.
+     * ex. <code>RSA Session-ID:XXX Master-Key:YYY</code>
+     * This format is understood by Wireshark 1.6.0.
+     * https://code.wireshark.org/review/gitweb?p=wireshark.git;a=commit;h=686d4cabb41185591c361f9ec6b709034317144b
+     * The key and session identifier are forwarded to the log named 'io.netty.wireshark'.
+     */
+    public static class WiresharkTlsMasterKeyHandler extends TlsMasterKeyHandler {
+
+        private static final InternalLogger wireshark_logger =
```
normanmaurer: nit: wireshark_logger -> wiresharkLogger

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,166 @@
...
+                InternalLoggerFactory.getInstance("io.netty.wireshark");
+
+        @Override
+        protected void accept(SecretKey masterKey, byte[] sessionId) {
+            if (masterKey.getEncoded().length != 48) {
+                throw new IllegalArgumentException("An invalid length master key was provided.");
+            }
+            wireshark_logger.warn("RSA Session-ID:{} Master-Key:{}",
+                    javax.xml.bind.DatatypeConverter.printHexBinary(sessionId).toLowerCase(),
+                    javax.xml.bind.DatatypeConverter.printHexBinary(masterKey.getEncoded()).toLowerCase());
```
normanmaurer: can we get rid of `javax.xml.bind.DatatypeConverter` ? This is not accessible in Java9+ easily

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,166 @@
...
+     * The unavailability clause of whether the private Sun implementation of SSLSessionImpl is available.
+     */
+    private static final Throwable UNAVAILABILITY_CAUSE;
+
+    static {
+        Throwable cause = null;
+        try {
+            SSL_SESSIONIMPL_CLASS = Class.forName("sun.security.ssl.SSLSessionImpl");
+            SSL_SESSIONIMPL_MASTER_SECRET_FIELD = SSL_SESSIONIMPL_CLASS.getDeclaredField("masterSecret");
+            SSL_SESSIONIMPL_MASTER_SECRET_FIELD.setAccessible(true);
```
normanmaurer: You want to use `ReflectionUtil` to ensure we not produce any warnings on Java9+

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,166 @@
...
+ * log the secret & identifier in a format that is consumable by Wireshark -- allowing easy decryption of pcap/tcpdumps.
+ */
+public abstract class TlsMasterKeyHandler extends ChannelInboundHandlerAdapter {
+
+    private static final InternalLogger logger = InternalLoggerFactory.getInstance(TlsMasterKeyHandler.class);
+
+    /**
+     * The JRE SSLSessionImpl cannot be imported
+     */
+    private static Class<?> SSL_SESSIONIMPL_CLASS;
```
normanmaurer: final

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,166 @@
...
+
+    /**
+     * The master key field in the SSLSessionImpl
+     */
+    private static Field SSL_SESSIONIMPL_MASTER_SECRET_FIELD;
```
normanmaurer: final

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /**
+     * Record the session identifier and master key to the {@link InternalLogger} named <code>io.netty.wireshark</code>.
+     * ex. <code>RSA Session-ID:XXX Master-Key:YYY</code>
+     * This format is understood by Wireshark 1.6.0.
+     * https://code.wireshark.org/review/gitweb?p=wireshark.git;a=commit;h=686d4cabb41185591c361f9ec6b709034317144b
+     * The key and session identifier are forwarded to the log named 'io.netty.wireshark'.
+     */
+    public static class WiresharkTlsMasterKeyHandler extends TlsMasterKeyHandler {
```
normanmaurer: maybe not expose the class directly but just add a static method that returns this instance ?

fzakaria:
@normanmaurer do you have a good idea for a test or other tests I can draw inspiration from ?
I would need some sort of "TeeClient" that stores the encrypted TLS message (without TLS record header) so that I could then go and decrypt them with the logged master key ?
or
Are we just looking for a more surface-level test for _TlsMasterKeyHandler_ that verifies it gets a non-null key material ?
(leaving the more functional test to the one I introduced in #8699)

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
fzakaria: I'm not familiar with this class: do I just need to access the fieldOffset then get the object ?
(no need to set Accessible on the field?)

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
fzakaria: I've left this in for now -- I'm open to making it once lookup.
I imagined this is largely a test driven handler anyways not to be included in typical pipelines?

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
fzakaria: I couldn't figure out the Java voo-doo to get it to be set in the catch block without it complaining "It might have already been set"

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
fzakaria: (see note above)

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
normanmaurer: Sorry I meant to say use `ReflectionUtil` 🤦‍♂️

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
normanmaurer: Just store it in a local variable first and then at the end of the static block assign it. 

Like for example done here:
https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/internal/CleanerJava9.java#L75

normanmaurer:
@fzakaria no idea for a test yet :(

fzakaria:
@normanmaurer thanks for looking at the other pull-requests and getting them merged.
I've now finally revised this one with the new master key ability in OpenSSL netty-tcnative.

I've added a test and curious on your feedback.

normanmaurer:
@netty-bot test this please

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,186 @@
...
+    /**
+     * Consume the master key for the session and the sessionId
+     * @param masterKey A 48-byte secret shared between the client and server.
+     * @param sessionId The sessionId An arbitrary byte sequence chosen by the server to identify an
+     *                  active or resumable session state.
+     */
+    protected abstract void accept(SecretKey masterKey, byte[] sessionId);
+
+    @Override
+    public void userEventTriggered(ChannelHandlerContext ctx, Object evt) {
```
normanmaurer: final

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    /**
+     * Record the session identifier and master key to the {@link InternalLogger} named <code>io.netty.wireshark</code>.
+     * ex. <code>RSA Session-ID:XXX Master-Key:YYY</code>
+     * This format is understood by Wireshark 1.6.0.
+     * https://code.wireshark.org/review/gitweb?p=wireshark.git;a=commit;h=686d4cabb41185591c361f9ec6b709034317144b
+     * The key and session identifier are forwarded to the log named 'io.netty.wireshark'.
+     */
+    public static final class WiresharkTlsMasterKeyHandler extends TlsMasterKeyHandler {
```
normanmaurer: I think I would mark this private and just expose a static method in the abstract class to obtain this handler.

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,186 @@
...
+        }
+        UNAVAILABILITY_CAUSE = cause;
+        SSL_SESSIONIMPL_CLASS = clazz;
+        SSL_SESSIONIMPL_MASTER_SECRET_FIELD = field;
+    }
+
+    /**
+     * Constructor.
+    */
+    public TlsMasterKeyHandler() {
```
normanmaurer: as this class is abstract this should be protected

fzakaria: done.

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
fzakaria: Was this comment on the wrong line?

normanmaurer: basically I was suggesting marking the method as final :)

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,197 @@
...
+     * @param sessionId The sessionId An arbitrary byte sequence chosen by the server to identify an
+     *                  active or resumable session state.
+     */
+    protected abstract void accept(SecretKey masterKey, byte[] sessionId);
+
+    @Override
+    public void userEventTriggered(ChannelHandlerContext ctx, Object evt) {
+        //only try to log the session info if the ssl handshake has successfully completed.
+        if (evt == SslHandshakeCompletionEvent.SUCCESS) {
+
```
normanmaurer: nit: remove whitespace

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,197 @@
...
+                    //the OracleJVM does not expose a way to get the master secret, so try to use reflection to get it.
+                    if (isSunSslEngineAvailable() && sslSession.getClass().equals(SSL_SESSIONIMPL_CLASS)) {
+                        SecretKey secretKey = (SecretKey) SSL_SESSIONIMPL_MASTER_SECRET_FIELD.get(sslSession);
+                        accept(secretKey, sslSession.getId());
+                    } else if (OpenSsl.isAvailable() && engine instanceof OpenSslEngine) {
+                        SecretKeySpec secretKey = new SecretKeySpec(
+                                SSL.getMasterKey(((OpenSslEngine) engine).sslPointer()), "AES");
+                        accept(secretKey, sslSession.getId());
+                    }
+                } catch (IllegalAccessException e) {
```
normanmaurer: this `try / catch` block should only close the reflection stuff itself. The openssl related code should not be in it.

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,197 @@
...
+                final SslHandler handler = ctx.pipeline().get(SslHandler.class);
+                final SSLEngine engine = handler.engine();
+                final SSLSession sslSession = engine.getSession();
+
+                try {
+                    //the OracleJVM does not expose a way to get the master secret, so try to use reflection to get it.
+                    if (isSunSslEngineAvailable() && sslSession.getClass().equals(SSL_SESSIONIMPL_CLASS)) {
+                        SecretKey secretKey = (SecretKey) SSL_SESSIONIMPL_MASTER_SECRET_FIELD.get(sslSession);
+                        accept(secretKey, sslSession.getId());
+                    } else if (OpenSsl.isAvailable() && engine instanceof OpenSslEngine) {
```
normanmaurer: `engine instanceof ReferenceCountedOpenSslEngine` (which is a the parent of `OpenSslEngine`)

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,197 @@
...
+                        SecretKeySpec secretKey = new SecretKeySpec(
+                                SSL.getMasterKey(((OpenSslEngine) engine).sslPointer()), "AES");
```
normanmaurer: cast to `ReferenceCountedOpenSslEngine`.

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,197 @@
...
+        return UNAVAILABILITY_CAUSE == null;
+    }
+
+    /**
+     * Consume the master key for the session and the sessionId
+     * @param masterKey A 48-byte secret shared between the client and server.
+     * @param sessionId The sessionId An arbitrary byte sequence chosen by the server to identify an
+     *                  active or resumable session state.
+     */
+    protected abstract void accept(SecretKey masterKey, byte[] sessionId);
```
normanmaurer: I wonder if it would be better to pass in the `SslSession`. This would make it more flexible IMHO.

normanmaurer:
@netty-bot test this please

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
fzakaria: marked it as final -- i couldn't see any other examples of marking it as final.
I guess just to make sure it's not overwritten.

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
fzakaria: the implementation of how to get the master key is different per SslSession implementation though -- the idea was to keep that nitty gritty away from the handler.

normanmaurer: Sure I was talking about passing the je and the session in. The is can then be obtained from the session of needed

fzakaria:
@normanmaurer is the build expected to pass in JDK11 ?
I'm not familiar with that release and at the moment it's failing to the use of Unsafe.

```
10:37:46.956 [main] DEBUG i.n.util.internal.PlatformDependent0 - jdk.internal.misc.Unsafe.allocateUninitializedArray(int): unavailable
java.lang.IllegalAccessException: class io.netty.util.internal.PlatformDependent0$6 cannot access class jdk.internal.misc.Unsafe (in module java.base) because module java.base does not export jdk.internal.misc to unnamed module @67f639d3
````

File: handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java [hunk 25]
```diff
@@ -3245,6 +3250,71 @@ protected void engineInit(ManagerFactoryParameters managerFactoryParameters) {
...
+    @Test
+    public void testMasterKeyLogging() throws Exception {
+
+        /*
+         * At the moment master key logging is not supported for conscrypt
+         */
+        if (serverSslContextProvider() instanceof OpenSSLProvider) {
+            return;
+        }
+
```
normanmaurer: Also add the following code to ensure the test is skipped when JDK9+ is used and its the JDK implementation (as reflection is not allowed there by default):

```
if (sslServerProvider() == SslProvider.JDK && PlatformDependent.javaVersion() > 8) {
    return;
}
```

normanmaurer:
@fzakaria see https://github.com/netty/netty/pull/8653#discussion_r299154840

handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
fzakaria: Yea I'll do that -- I figured it out at the same time.. 
Had to install java and run `mvn.surefire.debug` and attach to the process.
IntelliJ is shitting the bed when I changed the SDK to 11....

I was going do `Assume.assumeTrue(TlsMasterKeyHandler.isSunSslEngineAvailable());` but I like the explicit platform version check.

normanmaurer: Actually I think your idea is even better...

So just make it :

```java
if (sslServerProvider()) {
    Assume.assumeTrue(TlsMasterKeyHandler.isSunSslEngineAvailable());
}
```

fzakaria: I opted for something in the middle.
I actually like encoding that it's specific to the JDK version --
`Assume.assumeTrue(TlsMasterKeyHandler.isSunSslEngineAvailable())` might hide cases where it should be available on JDK8.

fzakaria:
Gave it another go. I appreciate the quick turn around on the comments & feedback btw.

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
normanmaurer: @fzakaria did you miss this one ?

Basically I think I would make it:
```
protected abstract void accept(SecretKey masterKey, SSLSession session);
```

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,197 @@
...
+        if (evt == SslHandshakeCompletionEvent.SUCCESS) {
+            boolean shouldHandle = SystemPropertyUtil.getBoolean(SYSTEM_PROP_KEY, false);
+
+            //if the handler is not an instance of DoNothingTlsMasterKeyBiConsumer try to handle it
+            if (shouldHandle) {
+                final SslHandler handler = ctx.pipeline().get(SslHandler.class);
+                final SSLEngine engine = handler.engine();
+                final SSLSession sslSession = engine.getSession();
+
+                //the OracleJVM does not expose a way to get the master secret, so try to use reflection to get it.
```
normanmaurer: OracleJVM -> OpenJDK

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    /**
+     * Create a {@link WiresharkTlsMasterKeyHandler} instance.
+     * This TLS master key handler logs the master key and session-id in a format
+     * understood by Wireshark -- this can be especially useful if you need to ever
+     * decrypt a TLS session and are using perfect forward secrecy (i.e. Diffie-Hellman)
+     * The key and session identifier are forwarded to the log named 'io.netty.wireshark'.
+     */
+    public static TlsMasterKeyHandler wireSharkTlsMasterKeyHandler() {
```
normanmaurer: nit:

newWireSharkTlsMasterKeyHandler()

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
fzakaria: changed.

handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
```diff
@@ -3245,6 +3250,75 @@ protected void engineInit(ManagerFactoryParameters managerFactoryParameters) {
...
+                    });
+                    serverConnectedChannel = ch;
+                }
+            }).bind(new InetSocketAddress(0)).syncUninterruptibly().channel();
+
+            int port = ((InetSocketAddress) serverChannel.localAddress()).getPort();
+
+            SSLContext sslContext = SSLContext.getInstance("TLS");
+            sslContext.init(null, InsecureTrustManagerFactory.INSTANCE.getTrustManagers(), null);
+            Socket socket = sslContext.getSocketFactory().createSocket("localhost", port);
```
normanmaurer: you need to close the socket in the finally block.

handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
```diff
@@ -3245,6 +3250,75 @@ protected void engineInit(ManagerFactoryParameters managerFactoryParameters) {
...
+            OutputStream out = socket.getOutputStream();
+            out.write(1);
+
```
normanmaurer: call.flush() just in case...

handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
normanmaurer: use `NetUtil.LOCALHOST` as address

handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
```diff
@@ -3245,6 +3250,75 @@ protected void engineInit(ManagerFactoryParameters managerFactoryParameters) {
...
+        System.setProperty(TlsMasterKeyHandler.SYSTEM_PROP_KEY, Boolean.TRUE.toString());
+
+        SelfSignedCertificate ssc = new SelfSignedCertificate();
+        serverSslCtx = SslContextBuilder.forServer(ssc.certificate(), ssc.privateKey())
+                .sslProvider(sslServerProvider())
+                .sslContextProvider(serverSslContextProvider())
+                .build();
+        try {
+            sb = new ServerBootstrap();
+            sb.group(new NioEventLoopGroup(), new NioEventLoopGroup());
```
normanmaurer: these EventLoopGroups must be shutdown in a finally block

handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
```diff
@@ -3245,6 +3250,75 @@ protected void engineInit(ManagerFactoryParameters managerFactoryParameters) {
...
+            sb.channel(NioServerSocketChannel.class);
+
+            final Promise<SecretKey> promise = sb.config().group().next().newPromise();
+            serverChannel = sb.childHandler(new ChannelInitializer<Channel>() {
```
normanmaurer: this channel must be closed in a finally block

handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
```diff
@@ -3245,6 +3250,75 @@ protected void engineInit(ManagerFactoryParameters managerFactoryParameters) {
...
+                @Override
+                protected void initChannel(Channel ch) throws Exception {
+                    ch.config().setAllocator(new TestByteBufAllocator(ch.config().getAllocator(), type));
+
+                    SslHandler sslHandler = delegatingExecutor == null ?
+                            serverSslCtx.newHandler(ch.alloc()) :
+                            serverSslCtx.newHandler(ch.alloc(), delegatingExecutor);
+
+                    ch.pipeline().addFirst(sslHandler);
```
normanmaurer: just use `addLast(...)` to be consistent

handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
fzakaria: too bad no JRE 7 with try-with-resources ;)

handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
fzakaria: I think these are closed in the `tearDown` of this class
```java
        if (sb != null) {
            serverGroupShutdownFuture = sb.config().group().shutdownGracefully(0, 0, TimeUnit.MILLISECONDS);
            serverChildGroupShutdownFuture = sb.config().childGroup().shutdownGracefully(0, 0, TimeUnit.MILLISECONDS);
        }
```

normanmaurer: ah yeah.. sorry missed this when doing the review with safari :(

handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
fzakaria: Also this is closed in the `tearDown`
```java
        if (serverChannel != null) {
            serverCloseFuture = serverChannel.close();
            serverChannel = null;
        }
```

normanmaurer: missed this... thanks

normanmaurer:
@netty-bot test this please

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,197 @@
...
+import javax.net.ssl.SSLSession;
+import java.lang.reflect.Field;
+
+/**
+ * The {@link TlsMasterKeyHandler} is a channel-handler you can include in your pipeline to consume the master key
+ * & session identifier for a TLS session.
+ * This can be very useful, for instance the {@link WiresharkTlsMasterKeyHandler} implementation will
+ * log the secret & identifier in a format that is consumable by Wireshark -- allowing easy decryption of pcap/tcpdumps.
+ */
+public abstract class TlsMasterKeyHandler extends ChannelInboundHandlerAdapter {
```
normanmaurer: One thing I wondered... should we call it `SslMasterKeyHandler` just to make it more consistent with other TLS / SSL handlers which are all prefixed with `Ssl` ?

fzakaria:
That's a good point. Let me do that.

handler/src/main/java/io/netty/handler/ssl/TlsMasterKeyHandler.java
```diff
@@ -0,0 +1,197 @@
...
+            if (shouldHandle) {
+                final SslHandler handler = ctx.pipeline().get(SslHandler.class);
+                final SSLEngine engine = handler.engine();
+                final SSLSession sslSession = engine.getSession();
+
+                //the OpenJDK does not expose a way to get the master secret, so try to use reflection to get it.
+                if (isSunSslEngineAvailable() && sslSession.getClass().equals(SSL_SESSIONIMPL_CLASS)) {
+                    try {
+                        SecretKey secretKey = (SecretKey) SSL_SESSIONIMPL_MASTER_SECRET_FIELD.get(sslSession);
+                        accept(secretKey, sslSession);
```
normanmaurer: nit: I think this should go out of the try block

fzakaria:
For those following this pull-request a demo _logback.xml_ would be (just the raw message and new line):
```xml
  <appender name="FILE" class="ch.qos.logback.core.FileAppender">
    <file>key.log</file>
    <encoder>
      <pattern>%m%n</pattern>
    </encoder>
  </appender>

   <logger name="io.netty.wireshark" addivity="false">
    <appender-ref ref="FILE" />
  </logger>
```
Don't forget to also `-Dio.netty.tls.masterKeyHandler=true`

normanmaurer:
@netty-bot test this please

fzakaria:
Yes it passed :)
6 months in the making this PR 👍

handler/src/main/java/io/netty/handler/ssl/SslMasterKeyHandler.java
```diff
@@ -0,0 +1,197 @@
...
+
+    /**
+     * A system property that can be used to turn on/off the {@link SslMasterKeyHandler} dynamically without having
+     * to edit your pipeline.
+     * <code>-Dio.netty.ssl.masterKeyHandler=true</code>
+     */
+    public static final String SYSTEM_PROP_KEY = "io.netty.ssl.masterKeyHandler";
+
+    /**
+     * The unavailability clause of whether the private Sun implementation of SSLSessionImpl is available.
```
carl-mastrangelo: s/clause/cause/

handler/src/main/java/io/netty/handler/ssl/SslMasterKeyHandler.java
```diff
@@ -0,0 +1,197 @@
...
+        protected void accept(SecretKey masterKey, SSLSession session) {
+            if (masterKey.getEncoded().length != 48) {
+                throw new IllegalArgumentException("An invalid length master key was provided.");
+            }
+            final byte[] sessionId = session.getId();
+            wireshark_logger.warn("RSA Session-ID:{} Master-Key:{}",
+                    printHexBinary(sessionId).toLowerCase(), printHexBinary(masterKey.getEncoded()).toLowerCase());
+        }
+
+        private static String printHexBinary(byte[] data) {
```
carl-mastrangelo: Maybe use ByteBufUtil's hexdump method?

handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
```diff
@@ -3245,6 +3251,89 @@ protected void engineInit(ManagerFactoryParameters managerFactoryParameters) {
...
+                    ch.pipeline().addLast(sslHandler);
+                    ch.pipeline().addLast(new SslMasterKeyHandler() {
+                        @Override
+                        protected void accept(SecretKey masterKey, SSLSession session) {
+                            promise.setSuccess(masterKey);
+                        }
+                    });
+                    serverConnectedChannel = ch;
+                }
+            }).bind(new InetSocketAddress(0)).syncUninterruptibly().channel();
```
carl-mastrangelo: optional: use regular `sync()` so if the test times out, it can be interrupted and exit early.

handler/src/main/java/io/netty/handler/ssl/SslMasterKeyHandler.java
fzakaria: done.

fzakaria:
@carl-mastrangelo responded to your feedback.

handler/src/main/java/io/netty/handler/ssl/SslMasterKeyHandler.java
```diff
@@ -0,0 +1,190 @@
...
+     * This format is understood by Wireshark 1.6.0.
+     * https://code.wireshark.org/review/gitweb?p=wireshark.git;a=commit;h=686d4cabb41185591c361f9ec6b709034317144b
+     * The key and session identifier are forwarded to the log named 'io.netty.wireshark'.
+     */
+    private static final class WiresharkSslMasterKeyHandler extends SslMasterKeyHandler {
+
+        private static final InternalLogger wireshark_logger =
+                InternalLoggerFactory.getInstance("io.netty.wireshark");
+
+        private static final char[] hexCode = "0123456789ABCDEF".toCharArray();
```
carl-mastrangelo: unused?

fzakaria: removed.

fzakaria:
@normanmaurer good to go ? :P

normanmaurer:
@netty-bot test this please

normanmaurer:
@fzakaria thanks a lot!

Issue #8654, opened 2018-12-13
NoClassDefFoundError on Android platform


mrniko:
### Expected behavior

No exception

### Actual behavior

Exception arise during invocation DnsServerAddressStreamProviders.platformDefault method in Android application:
```java
java.lang.NoClassDefFoundError: Failed resolution of: Ljavax/naming/directory/InitialDirContext;
        at io.netty.resolver.dns.DefaultDnsServerAddressStreamProvider.<clinit>(DefaultDnsServerAddressStreamProvider.java:68)
        at io.netty.resolver.dns.UnixResolverDnsServerAddressStreamProvider.parseSilently(UnixResolverDnsServerAddressStreamProvider.java:76)
        at io.netty.resolver.dns.DnsServerAddressStreamProviders.<clinit>(DnsServerAddressStreamProviders.java:32)
        at io.netty.resolver.dns.DnsServerAddressStreamProviders.platformDefault(DnsServerAddressStreamProviders.java:45)
```
### Steps to reproduce

execute `DnsServerAddressStreamProviders.platformDefault` method

### Netty version

4.1.31.Final

normanmaurer:
@mrniko thanks for reporting... should be fixed by https://github.com/netty/netty/pull/8656

PR #8656, opened 2018-12-13
NoClassDefFoundError on Android platform when try to use DefaultDnsSe…

normanmaurer:
…rverAddressStreamProvider.

Motivation:

Andoid does not contain javax.naming.* so we should not try to use it to prevent a NoClassDefFoundError on init.

Modifications:

Only try to use javax.naming.* to retrieve nameservers when not using Android.

Result:

Fixes https://github.com/netty/netty/issues/8654.

### DIFF
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DirContextUtils.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DirContextUtils.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2018 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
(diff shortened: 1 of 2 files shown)


resolver-dns/src/main/java/io/netty/resolver/dns/DirContextUtils.java
```diff
@@ -0,0 +1,77 @@
...
+                        }
+                        int port  = uri.getPort();
+                        defaultNameServers.add(SocketUtils.socketAddress(uri.getHost(), port == -1 ?
+                                defaultPort : port));
+                    } catch (URISyntaxException e) {
+                        logger.debug("Skipping a malformed nameserver URI: {}", server, e);
+                    }
+                }
+            }
+        } catch (NamingException ignore) {
```
carl-mastrangelo: Optional: logging fine the the NamingExceptions was useful when developing gRPC's DNS support.

Issue #8575, opened 2018-11-20
Unexpected IllegalReferenceCountException on decode multipart request


ssserj:
### Actual behavior
```
io.netty.util.IllegalReferenceCountException: refCnt: 0, decrement: 1
	at io.netty.util.AbstractReferenceCounted.release0(AbstractReferenceCounted.java:87)
	at io.netty.util.AbstractReferenceCounted.release(AbstractReferenceCounted.java:71)
	at io.netty.handler.codec.http.multipart.MixedAttribute.release(MixedAttribute.java:320)
	at io.netty.handler.codec.http.multipart.HttpPostMultipartRequestDecoder.destroy(HttpPostMultipartRequestDecoder.java:947)
	at io.netty.handler.codec.http.multipart.HttpPostRequestDecoder.destroy(HttpPostRequestDecoder.java:247)
```

### Steps to reproduce
```java
import io.netty.buffer.ByteBuf;
import io.netty.buffer.Unpooled;
import io.netty.handler.codec.http.DefaultFullHttpRequest;
import io.netty.handler.codec.http.HttpMethod;
import io.netty.handler.codec.http.HttpVersion;
import io.netty.handler.codec.http.multipart.Attribute;
import io.netty.handler.codec.http.multipart.DefaultHttpDataFactory;
import io.netty.handler.codec.http.multipart.HttpPostRequestDecoder;
import io.netty.handler.codec.http.multipart.InterfaceHttpData;
import org.junit.Test;

import java.io.IOException;

import static org.junit.Assert.assertEquals;
import static org.junit.Assert.assertTrue;

public class HttpMultipartRequestTest {
    
    String getPostParam(HttpPostRequestDecoder requestDecoder, String name) {
        final InterfaceHttpData data = requestDecoder.getBodyHttpData(name);

        if (data != null && InterfaceHttpData.HttpDataType.Attribute.equals(data.getHttpDataType())) {
            final Attribute attr = (Attribute)data;

            try {
                return attr.getValue();
            } catch (IOException e) {
                return null;
            }
        }

        return null;
    }

    @Test
    public void testMultipartRequest() {
        ByteBuf byteBuf = Unpooled.wrappedBuffer(("------------------------------01f136d9282f\n" +
                "Content-Disposition: form-data; name=\"msg_id\"\n" +
                "\n" +
                "15200\n" +
                "------------------------------01f136d9282f\n" +
                "Content-Disposition: form-data; name=\"msg\"\n" +
                "\n" +
[...]

PR #8660, opened 2018-12-14
Fix unexpected IllegalReferenceCountException on decode multipart request

ssserj:
Fix unexpected IllegalReferenceCountException on decode multipart request. (#8575)

Motivation:

Http post request may be encoded as 'multipart/form-data' without any files and consist mixed attributes only.

Modifications:

Check reference count > 0 added on every InterfaceHttpData before release() call.

Result:

Code does not throw an IllegalReferenceCountException.


normanmaurer:
@ssserj imho that is not a correct fix. You should never need to check the refCnt before you release. This will just mask the bug. There must be something else wrong.

kachayev:
@normanmaurer Unfortunately, decoder already contains code like this [here](https://github.com/netty/netty/blob/948d4a9ec58aef092c84eb76e672fd06e81cc13c/codec-http/src/main/java/io/netty/handler/codec/http/multipart/HttpPostMultipartRequestDecoder.java#L940). I don't know why do we need to check `refCnt` there too.

ssserj:
@normanmaurer The problem might be introduced by the commit 2b4f667. More detailed in issue #8575 and PR #8964

normanmaurer:
@ssserj sorry for the long turn-around. Totally missed this one. Can you please sign our ICLA and let me know one done ?

https://netty.io/s/icla ?

codec-http/src/test/java/io/netty/handler/codec/http/multipart/HttpPostRequestDecoderTest.java
```diff
@@ -656,4 +656,36 @@ public void testDecodeMalformedEmptyContentTypeFieldParameters() throws Exceptio
...
+                        java.nio.charset.Charset.forName("utf-8"));
+
+        assertTrue(decoder.isMultipart());
+        assertFalse(decoder.getBodyHttpDatas().isEmpty());
+        assertEquals(2, decoder.getBodyHttpDatas().size());
+        assertEquals("test message", ((Attribute) decoder.getBodyHttpData("msg")).getValue());
+        assertEquals("15200", ((Attribute) decoder.getBodyHttpData("msg_id")).getValue());
+
+        decoder.destroy();
+    }
```
normanmaurer: can we also add a `assertEquals(0, req.refCnt());` after the `destroy()` call ?

codec-http/src/test/java/io/netty/handler/codec/http/multipart/HttpPostRequestDecoderTest.java
```diff
@@ -656,4 +656,36 @@ public void testDecodeMalformedEmptyContentTypeFieldParameters() throws Exceptio
...
+                "test message\n" +
+                "--" + BOUNDARY + "--").getBytes());
+
+        DefaultFullHttpRequest req = new DefaultFullHttpRequest(HttpVersion.HTTP_1_0, HttpMethod.POST, "/up", byteBuf);
+        req.headers().add(HttpHeaderNames.CONTENT_TYPE, "multipart/form-data; boundary=" + BOUNDARY);
+
+        HttpPostRequestDecoder decoder =
+                new HttpPostRequestDecoder(new DefaultHttpDataFactory(DefaultHttpDataFactory.MINSIZE),
+                        req,
+                        java.nio.charset.Charset.forName("utf-8"));
```
normanmaurer: use `CharsetUtil.UTF_8`

codec-http/src/test/java/io/netty/handler/codec/http/multipart/HttpPostRequestDecoderTest.java
```diff
@@ -656,4 +656,36 @@ public void testDecodeMalformedEmptyContentTypeFieldParameters() throws Exceptio
...
+                "Content-Disposition: form-data; name=\"msg_id\"\n" +
+                "\n" +
+                "15200\n" +
+                "--" + BOUNDARY + "\n" +
+                "Content-Disposition: form-data; name=\"msg\"\n" +
+                "\n" +
+                "test message\n" +
+                "--" + BOUNDARY + "--").getBytes());
+
+        DefaultFullHttpRequest req = new DefaultFullHttpRequest(HttpVersion.HTTP_1_0, HttpMethod.POST, "/up", byteBuf);
```
normanmaurer: nit: `FullHttpRequest req = ...`

ssserj:
@normanmaurer ICLA was signed. Changes was made. But `req.refCnt()` is not 0. I will be research why?((

codec-http/src/test/java/io/netty/handler/codec/http/multipart/HttpPostRequestDecoderTest.java
ssserj: Most probably, should be assertEquals(1, req.refCnt()) ?

normanmaurer:
@ssserj hmm... let me investigate a bit. I thought it should be 0 as we basically destroyed the decoder + everything that is still stored in there.

normanmaurer:
@ssserj actually yes your are right 1 is correct as this class does not take over ownership as of today.

normanmaurer:
@netty-bot test this please

codec-http/src/test/java/io/netty/handler/codec/http/multipart/HttpPostRequestDecoderTest.java
```diff
@@ -20,16 +20,7 @@
...
-import io.netty.handler.codec.http.DefaultHttpContent;
-import io.netty.handler.codec.http.DefaultHttpRequest;
-import io.netty.handler.codec.http.DefaultLastHttpContent;
-import io.netty.handler.codec.http.HttpContent;
-import io.netty.handler.codec.http.HttpHeaderNames;
-import io.netty.handler.codec.http.HttpHeaderValues;
-import io.netty.handler.codec.http.HttpMethod;
-import io.netty.handler.codec.http.HttpVersion;
-import io.netty.handler.codec.http.LastHttpContent;
+import io.netty.handler.codec.http.*;
```
normanmaurer: nit: please revert... we usually not do wildcard imports for this kind of things

normanmaurer:
Squashed, cleaned up and cherry-picked as 9e2922b04d2901647b717232ed8fa1745bdbe0df into 4.1

PR #8664, opened 2018-12-14
Support 1012, 1013 and 1014 WebSocket close status code

slandelle:
Motivation:

RFC 6455 doesn't define close status codes 1012, 1013 and 1014.
Yet, since then, IANA has defined them and web browsers support them.

From https://www.iana.org/assignments/websocket/websocket.xhtml:

* 1012: Service Restart
* 1013: Try Again Later
* 1014: The server was acting as a gateway or proxy and received an invalid response from the upstream server. This is similar to 502 HTTP Status Code.

Modification:

Make status codes 1012, 1013 and 1014 legit.

Result:

WebSocket status codes as defined by IANA are supported.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocket08FrameDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocket08FrameDecoder.java
@@ -451,7 +451,7 @@ public class WebSocket08FrameDecoder extends ByteToMessageDecoder
         // Must have 2 byte integer within the valid range
         int statusCode = buffer.readShort();
         if (statusCode >= 0 && statusCode <= 999 || statusCode >= 1004 && statusCode <= 1006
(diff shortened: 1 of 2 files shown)


slandelle:
Investigating if autobahn test suite can support those new codes.

slandelle:
Great!
Please note that I did’t add 1015 as a valid code.
One the one side, I probably should to stick to the spec.
On the other side, I don’t get how one could get a websocket frame in case of a TLS handshake error. Sadly, the links on the IANA page to the related mailing list are dead.
WDYT?

normanmaurer:
@slandelle yeah... I think what you did is good enough for now...

slandelle:
@normanmaurer What's the command for triggering CI, please?

normanmaurer:
@netty-bot test this please 

@slandelle

slandelle:
Thanks!

normanmaurer:
@slandelle thanks... merged!

PR #8665, opened 2018-12-16
Fix ClassCastException and native crash when using kqueue transport.

normanmaurer:
Motivation:

How we did the mapping from native code to AbstractKQueueChannel was not safe and could lead to heap corruption. This then sometimes produced ClassCastExceptions or could also lead to crashes. This happened sometimes when running the testsuite.

Modifications:

Use a Map for the mapping (just as we do in the native epoll transport).

Result:

No more heap corruption / crashes.


normanmaurer:
Let me also merge this so I can move forward. If we need we can do a followup.

carl-mastrangelo:
LGTM

Issue #6596, opened 2017-04-01
Netty producing exception when server is killed


sshntt:
### Expected behavior
I have used to following code to handle server shutdown:
```
public void stop() {
        try {
            this.channelFuture.channel().close();
            this.channelFuture.channel().closeFuture().sync();
        } catch (InterruptedException e) {
            Log.exceptionStack(e);
        } finally {
            this.bootstrap.config().childGroup().shutdownGracefully(10, 30, TimeUnit.SECONDS);
            this.bootstrap.config().group().shutdownGracefully(5, 10, TimeUnit.SECONDS);
        }
    }
```

It should shutdown without any exceptions

### Actual behavior
Sometimes the server shutsdown with the following exception:

```
Exception in thread "nioEventLoopGroup-3-1" java.lang.NoClassDefFoundError: io/netty/buffer/PoolArena$1
        at io.netty.buffer.PoolArena.freeChunk(PoolArena.java:293)
        at io.netty.buffer.PoolThreadCache$MemoryRegionCache.freeEntry(PoolThreadCache.java:471)
        at io.netty.buffer.PoolThreadCache$MemoryRegionCache.free(PoolThreadCache.java:441)
        at io.netty.buffer.PoolThreadCache$MemoryRegionCache.free(PoolThreadCache.java:433)
        at io.netty.buffer.PoolThreadCache.free(PoolThreadCache.java:290)
        at io.netty.buffer.PoolThreadCache.free(PoolThreadCache.java:281)
        at io.netty.buffer.PoolThreadCache.free0(PoolThreadCache.java:254)
        at io.netty.buffer.PoolThreadCache.free(PoolThreadCache.java:250)
        at io.netty.buffer.PooledByteBufAllocator$PoolThreadLocalCache.onRemoval(PooledByteBufAllocator.java:431)
        at io.netty.buffer.PooledByteBufAllocator$PoolThreadLocalCache.onRemoval(PooledByteBufAllocator.java:408)
        at io.netty.util.concurrent.FastThreadLocal.remove(FastThreadLocal.java:226)
        at io.netty.util.concurrent.FastThreadLocal.removeAll(FastThreadLocal.java:67)
        at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:146)
        at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.ClassNotFoundException: io.netty.buffer.PoolArena$1
[...]

PR #8666, opened 2018-12-16
Call FastThreadLocal.removeAll() before notify termination future of …

normanmaurer:
…SingleThreadEventExecutor.

Motivation:

We should try removing all FastThreadLocals for the Thread before we notify the termination. future. The user may block on the future and once it unblocks the JVM may terminate and start unloading classes.

Modifications:

Remove all FastThreadLocals for the Thread before notify termination future.

Result:

Fixes https://github.com/netty/netty/issues/6596.

### DIFF
--- a/common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
+++ b/common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
@@ -948,6 +948,12 @@ public abstract class SingleThreadEventExecutor extends AbstractScheduledEventEx
                                 }
                             }
 
+                            // Lets remove all FastThreadLocals for the Thread as we are about to terminate and notify
+                            // the future. The user may block on the future and once it unblocks the JVM may terminate
+                            // and start unloading classes.
+                            // See https://github.com/netty/netty/issues/6596.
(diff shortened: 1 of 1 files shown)


common/src/main/java/io/netty/util/concurrent/SingleThreadEventExecutor.java
```diff
@@ -948,6 +948,12 @@ public void run() {
                                 }
                             }
 
+                            // Lets remove all FastThreadLocals for the Thread as we are about to terminate and notify
+                            // the future. The user may block on the future and once it unblocks the JVM may terminate
+                            // and start unloading classes.
+                            // See https://github.com/netty/netty/issues/6596.
+                            FastThreadLocal.removeAll();
```
ejona86: Should this happen after `threadLock.release()`/before `STATE_UPDATER.set()`? It seems awaitTermination returns when threadLock is released.

normanmaurer: good point... let me fix this

normanmaurer:
@ejona86 @carl-mastrangelo PTAL again

Issue #8549, opened 2018-11-13
[Netty 5] ChannelPool removal


normanmaurer:
We will consider removing the ChannelPool and related implementations. It is often the case that having protocol knowledge can result in more effective pooling and `ChannelPool` currently doesn’t have this knowledge. This responsibility is assumed to be implemented at layers higher in the stack than Netty.

wuqinduo:
@normanmaurer Hello, How to implement channelpool in netty5 . Are there any examples ? Thanks

He-Pin:
@normanmaurer I think we should better extract it to a dedicated module, for a simpler migration path. and for most cases, the current pool is good enough.

He-Pin:
If you like I think I can send a PR to extract this module to something like `transport-pooling` module.

PR #8681, opened 2018-12-21
One line description of your change

normanmaurer:
Motivation:

We should remove the ChannelPool and related implementations. It is often the case that having protocol knowledge can result in more effective pooling and ChannelPool currently doesn’t have this knowledge. This responsibility is assumed to be implemented at layers higher in the stack than Netty.

Modifications:

Remove io.netty.channel.pool.*

Result:

Less code to maintain, fixes https://github.com/netty/netty/issues/8549.


normanmaurer:
@netty-bot test this please

Issue #8676, opened 2018-12-19
Race condition between adding and removing SslHandler causing NPE


mingyu89:
We are seeing the same issue as in ticket https://github.com/netty/netty/issues/6536 on version 4.1.28.
The variable `pendingUnencryptedWrites` in SslHandler is initialized in method "handlerAdded" if the following sequence of event happens, we will hit this NPE : 

1. connect channel
2. add handler to channel
3. disconnect channel
4. call `SslHandler.handlerRemoved0` from disconnecting channel. This will access `pendingUnencryptedWrites` and throw NPE
5. call `SslHandler.handlerAdded` and initialize pendingUnencryptedWrites. But it's too late.

```
2018-12-06 15:50:20.473 WARN [nioEventLoopGroup-3-15] io.netty.channel.DefaultChannelPipeline - An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.
io.netty.channel.ChannelPipelineException: io.netty.handler.ssl.SslHandler.handlerRemoved() has thrown an exception.
 at io.netty.channel.DefaultChannelPipeline.callHandlerRemoved0(DefaultChannelPipeline.java:676) [netty-all-4.1.28.Final.jar:4.1.28.Final]
 at io.netty.channel.DefaultChannelPipeline.destroyDown(DefaultChannelPipeline.java:922) [netty-all-4.1.28.Final.jar:4.1.28.Final]
 at io.netty.channel.DefaultChannelPipeline.destroyUp(DefaultChannelPipeline.java:888) [netty-all-4.1.28.Final.jar:4.1.28.Final]
 at io.netty.channel.DefaultChannelPipeline.destroy(DefaultChannelPipeline.java:880) [netty-all-4.1.28.Final.jar:4.1.28.Final]
 at io.netty.channel.DefaultChannelPipeline.access$700(DefaultChannelPipeline.java:46) [netty-all-4.1.28.Final.jar:4.1.28.Final]
 at io.netty.channel.DefaultChannelPipeline$HeadContext.channelUnregistered(DefaultChannelPipeline.java:1416) [netty-all-4.1.28.Final.jar:4.1.28.Final]
 at io.netty.channel.AbstractChannelHandlerContext.invokeChannelUnregistered(AbstractChannelHandlerContext.java:181) [netty-all-4.1.28.Final.jar:4.1.28.Final]
[...]

Issue #6536, opened 2017-03-13
SSLHandler handleRemoved called before handlerAdded


yogeshvedpathak:
I am getting following stack trace when a connection terminates. I think the null pointer exception is originated because `pendingUnencryptedWrites` is null at link `415`. This could happen when `handleRemoved` is called before the `handleAdded`

2017-03-10 00:04:12.063 WARN  [nioEventLoopGroup-3-5] io.netty.channel.DefaultChannelPipeline - An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.
io.netty.channel.ChannelPipelineException: io.netty.handler.ssl.SslHandler.handlerRemoved() has thrown an exception.
        at io.netty.channel.DefaultChannelPipeline.callHandlerRemoved0(DefaultChannelPipeline.java:621) [netty-all-4.0.41.Final.jar:4.0.41.Final]
        at io.netty.channel.DefaultChannelPipeline.destroyDown(DefaultChannelPipeline.java:868) [netty-all-4.0.41.Final.jar:4.0.41.Final]
        at io.netty.channel.DefaultChannelPipeline.destroyUp(DefaultChannelPipeline.java:834) [netty-all-4.0.41.Final.jar:4.0.41.Final]
        at io.netty.channel.DefaultChannelPipeline.destroy(DefaultChannelPipeline.java:826) [netty-all-4.0.41.Final.jar:4.0.41.Final]
        at io.netty.channel.DefaultChannelPipeline.access$700(DefaultChannelPipeline.java:43) [netty-all-4.0.41.Final.jar:4.0.41.Final]
        at io.netty.channel.DefaultChannelPipeline$HeadContext.channelUnregistered(DefaultChannelPipeline.java:1276) [netty-all-4.0.41.Final.jar:4.0.41.Final]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelUnregistered(AbstractChannelHandlerContext.java:185) [netty-all-4.0.41.Final.jar:4.0.41.Final]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelUnregistered(AbstractChannelHandlerContext.java:171) [netty-all-4.0.41.Final.jar:4.0.41.Final]
        at io.netty.channel.DefaultChannelPipeline.fireChannelUnregistered(DefaultChannelPipeline.java:811) [netty-all-4.0.41.Final.jar:4.0.41.Final]
[...]

PR #8684, opened 2018-12-22
Only call handlerRemoved(...) if handlerAdded(...) was called during …

normanmaurer:
…adding the handler to the pipeline.

Motivation:

Due a race in DefaultChannelPipeline / AbstractChannelHandlerContext it was possible to have only handlerRemoved(...) called during tearing down the pipeline, even when handlerAdded(...) was never called. We need to ensure we either call both of none to guarantee a proper lifecycle of the handler.

Modifications:

- Enforce handlerAdded(...) / handlerRemoved(...) semantics / ordering
- Add unit test.

Result:

Fixes https://github.com/netty/netty/issues/8676 / https://github.com/netty/netty/issues/6536 .

### DIFF
--- a/transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
+++ b/transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
@@ -962,14 +962,17 @@ abstract class AbstractChannelHandlerContext extends DefaultAttributeMap
         handlerState = REMOVE_COMPLETE;
     }
 
-    final void setAddComplete() {
+    final boolean setAddComplete() {
         for (;;) {
             int oldState = handlerState;
+            if (oldState == REMOVE_COMPLETE) {
+                return false;
+            }
             // Ensure we never update when the handlerState is REMOVE_COMPLETE already.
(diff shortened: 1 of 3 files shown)


transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
```diff
@@ -979,6 +982,26 @@ final void setAddPending() {
...
+
+    final void callHandlerRemoved() throws Exception {
+        try {
+            // Only call handlerRemoved(...) if we called handlerAdded(...) before.
+            if (handlerState == ADD_COMPLETE) {
+                handler().handlerRemoved(this);
+            }
+        } finally {
+            // Mark the handler as removed in any case.
+            setRemoved();
```
bryce-anderson: Not a part of this change, but it's surprising to me that we call `setRemoved()` _after_ we call `handlerRemoved`. For example, if the handler decides to call `ctx.isRemoved()` in it's `handlerRemoved` method it will report `false` which seems inconsistent.

normanmaurer: @bryce-anderson let me open another issue to investigate this.

Issue #8687, opened 2018-12-26
Dead Loop in DNS resolver


slggamerTrue:
### Problem
There is a dead loop in DnsResolveContext.onExpectedResponse when there is a loop in the CNAMES
![image](https://user-images.githubusercontent.com/20608396/50428555-8ac8c780-08f3-11e9-908c-5ea0c8cf1d6a.png)

We found this problem because one DNS meet the problem
![image](https://user-images.githubusercontent.com/20608396/50428579-ca8faf00-08f3-11e9-829c-8927ebf72baf.png)

### Steps to reproduce
There is a small chance that the "www.altura.org" would response the loop CNAME. It's easy to reproduce the problem by your own DNS server.
The same case can be handled by the JDK DNS Resolver.

### Netty version
4.1.27

### JVM version (e.g. `java -version`)
JDK 1.8.181

### OS version (e.g. `uname -a`)
Centos 7.0

normanmaurer:
@slggamerTrue thanks for reporting. Let me come up with a unit test and fix.

normanmaurer:
@slggamerTrue can you test https://github.com/netty/netty/pull/8691 ?

slggamerTrue:
Thanks @normanmaurer. The fix is perfect. Looking forward to the new release.

normanmaurer:
@slggamerTrue thanks for checking and reporting

PR #8691, opened 2018-12-27
Correctly detect and handle CNAME loops.

normanmaurer:
Motivation:

We do not correctly detect loops when follow CNAMEs and so may try to follow it without any success.

Modifications:

- Correctly detect CNAME loops
- Do not cache CNAME entries which point to itself
- Add unit test.

Result:

Fixes https://github.com/netty/netty/issues/8687.


normanmaurer:
Fix was verified by @slggamerTrue.

Issue #8692, opened 2018-12-28
HTTP/2 stream initialization fails on incoming preface of max concurrent stream = 1


angn:
### Expected behavior
No error.

### Actual behavior
IllegalStateException is fired.

### Steps to reproduce
This error happens with hosts that sends a preface with max concurrent stream = 1, which was `api.development.push.apple.com` in my case.
Digging the code, I found `frameStreamToInitialize` is not null opposed to the assertion at https://github.com/netty/netty/blob/netty-4.1.29.Final/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java#L363

### Minimal yet complete reproducer code (or URL to code)
```
import io.netty.bootstrap.Bootstrap;
import io.netty.buffer.Unpooled;
import io.netty.channel.Channel;
import io.netty.channel.ChannelFutureListener;
import io.netty.channel.ChannelInitializer;
import io.netty.channel.nio.NioEventLoopGroup;
import io.netty.channel.socket.nio.NioSocketChannel;
import io.netty.handler.codec.http.HttpMethod;
import io.netty.handler.codec.http2.*;
import io.netty.handler.logging.LogLevel;
import io.netty.handler.ssl.*;
import io.netty.util.concurrent.FutureListener;

import javax.net.ssl.SSLException;
import java.util.concurrent.CountDownLatch;

final class TestCase {
    public static void main(String[] args) throws InterruptedException, SSLException {
        ApplicationProtocolConfig apConfig = new ApplicationProtocolConfig(
                ApplicationProtocolConfig.Protocol.ALPN,
                ApplicationProtocolConfig.SelectorFailureBehavior.NO_ADVERTISE,
                ApplicationProtocolConfig.SelectedListenerFailureBehavior.ACCEPT,
                ApplicationProtocolNames.HTTP_2);

        SslContext sslContext = SslContextBuilder.forClient()
                .ciphers(Http2SecurityUtil.CIPHERS, SupportedCipherSuiteFilter.INSTANCE)
                .applicationProtocolConfig(apConfig)
                .build();

        NioEventLoopGroup group = new NioEventLoopGroup(1);

        try {
            Channel channel = new Bootstrap()
                    .group(group)
                    .channel(NioSocketChannel.class)
                    .handler(new ChannelInitializer<>() {
                        @Override
                        protected void initChannel(Channel ch) {
[...]

PR #8694, opened 2018-12-28
Correctly buffer multiple outbound streams if needed.

normanmaurer:
Motivation:

In Http2FrameCodec we made the incorrect assumption that we can only have 1 buffered outboundstream as maximum. This is not correct and we need to account for multiple buffered streams.

Modifications:

- Use a map to allow buffer multiple streams
- Add unit test.

Result:

Fixes https://github.com/netty/netty/issues/8692.


normanmaurer:
Reported by @angn

normanmaurer:
@netty-bot test this please

PR #8705, opened 2019-01-07
Close connection for CorruptedFrameException

riyafa:
Motivation:

The CorruptedFrameException from the finish() method of the Utf8Validator gets propagated to other handlers while the connection is still open.

Modification:

Override exceptionCaught method of the Utf8FrameValidator and close the connection if it is a CorruptedFrameException.

Result:

The CorruptedFrameException gets propagated to other handlers only after properly closing the connection.

Fixes #<GitHub issue number>. 

Properly handle the CorruptedFrameException in Utf8FrameValidator.

### DIFF
--- a/codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketUtf8FrameValidatorTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketUtf8FrameValidatorTest.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2019 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.handler.codec.http.websocketx;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import io.netty.channel.embedded.EmbeddedChannel;
+import io.netty.handler.codec.CorruptedFrameException;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class WebSocketUtf8FrameValidatorTest {
+
+    @Test
+    public void testCorruptedFrameExceptionInFinish() {
+        assertCorruptedFrameExceptionHandling(new byte[]{-50});
+    }
+
+    @Test
+    public void testCorruptedFrameExceptionInCheck() {
+        assertCorruptedFrameExceptionHandling(new byte[]{-8, -120, -128, -128, -128});
+    }
+
+    private void assertCorruptedFrameExceptionHandling(byte[] data) {
+        EmbeddedChannel channel = new EmbeddedChannel(new Utf8FrameValidator());
+        try {
+            channel.writeInbound(new TextWebSocketFrame(Unpooled.copiedBuffer(data)));
+            Assert.fail();
+        } catch (CorruptedFrameException e) {
+            // expected exception
+        }
+        Assert.assertTrue(channel.finish());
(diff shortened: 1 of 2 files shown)


normanmaurer:
@riyafa can you please add a unit test and explain why this is more correct then the old implementation ?

riyafa:
@normanmaurer  I have added unit test cases. 
This PR is only a small change and the reason that it is more correct than the previous one is because in the WebSocket spec a CorruptedFrame needs to be followed by the closing of the connection. In the case of invalid utf-8 frame the connection should be closed with a close code of 1007 or with a none. In the `Utf8Validator` class the `CorruptedFrameException` is thrown in two places. Once in the `process` method and again in the `finish` method. The exception from the `process` method was handled in the `checkUTF8String` method of the `Utf8FrameValidator` and was not propagated to the other handlers. But the exception from the `finish` method propagated to the other handlers without being handled. 

This PR propagates exceptions from both locations to other handlers while properly closing the connection when the exception occurs.

codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketUtf8FrameValidatorTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Test
+    public void testCorruptedFrameExceptionInFinish() {
+        EmbeddedChannel channel = new EmbeddedChannel(new Utf8FrameValidator());
+        try {
+            channel.writeInbound(new TextWebSocketFrame(Unpooled.copiedBuffer(new byte[]{-50})));
+            Assert.fail();
+        } catch (CorruptedFrameException e) {
+            // expected exception
+        }
```
normanmaurer: @riyafa can you also call `channel.finish()` and assert the return value ?

codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketUtf8FrameValidatorTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Test
+    public void testCorruptedFrameExceptionInCheck() {
+        EmbeddedChannel channel = new EmbeddedChannel(new Utf8FrameValidator());
+        try {
+            channel.writeInbound(new TextWebSocketFrame(Unpooled.copiedBuffer(new byte[]{-8, -120, -128, -128, -128})));
+            Assert.fail();
+        } catch (CorruptedFrameException e) {
+            // expected exception
+        }
```
normanmaurer: @riyafa can you also call `channel.finish()` and assert the return value ?

normanmaurer:
@netty-bot test this please

normanmaurer:
@netty-bot test this please

codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketUtf8FrameValidatorTest.java
```diff
@@ -0,0 +1,50 @@
...
+    private void assertCorruptedFrameExceptionHandling(byte[] data) {
+        EmbeddedChannel channel = new EmbeddedChannel(new Utf8FrameValidator());
+        try {
+            channel.writeInbound(new TextWebSocketFrame(Unpooled.copiedBuffer(data)));
+            Assert.fail();
+        } catch (CorruptedFrameException e) {
+            // expected exception
+        }
+        Assert.assertTrue(channel.finish());
+        EmptyByteBuf buf = channel.readOutbound();
```
normanmaurer: I think we should depend on the fact that this is an instance of EmptyByteBuf... just check it’s not readable and release it

riyafa: Thank you. I have fixed this.

codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketUtf8FrameValidatorTest.java
```diff
@@ -0,0 +1,55 @@
...
```
normanmaurer: Change `EmptyByteBuf` to `ByteBuf`.

riyafa: Thanks. Fixed.

normanmaurer:
@netty-bot test this please

normanmaurer:
@riyafa thanks a lot!

PR #8711, opened 2019-01-09
Remove unnecessary loop variable from `AsciiString`.

lewisd32:
Motivation:

Incrementing two variables in sync is not necessary when only one will do.

Modifications:

- Remove `j` from `for` loop and replace with `i`.
- Add more unit testing scenarios to cover changed code.

Results:

Unnecessary variable removed.


normanmaurer:
@netty-bot test this please

Issue #8700, opened 2019-01-02
ChunkedWriteHandler may report a failed write as a successful one


kachayev:
### Expected behavior

If the writing of the last chunk of the chunked input message failed, the corresponding promise should be failed with an appropriate exception. 

### Actual behavior

`operationComplete` callback [reports success](https://github.com/netty/netty/blob/fa84e2b3af45ec7fd47909eff0aa7d2be5a54972/handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java#L268) with no check of the `future.isSuccesful()`. If I'm not missing anything and this was not done on purpose, I'll submit a PR with a fix.

### Steps to reproduce

n/a

### Minimal yet complete reproducer code (or URL to code)

n/a

### Netty version

4.1.32

### JVM version (e.g. `java -version`)

n/a

### OS version (e.g. `uname -a`)

n/a

kachayev:
Another related question. When using `HttpChunkedInput` with `ChunkedWriteHandler`, `lastHttpContent` would be written/flushed even in case write promise is already reported as a failed one. We rely on `endOfInput` being `true` after `closeInput` was called, which doesn't work for  `HttpChunkedInput`.

normanmaurer:
@kachayev yes this is a bug... would you mind open a PR with a fix ?

kachayev:
@normanmaurer Sure! Will do shortly.

PR #8716, opened 2019-01-14
Correctly propagate write failures from ChunkedWriteHandler

kachayev:
Motivation:

`ChunkedWriteHandler` should report write operation as failed in case *any* chunked was not written. Right now this is not true for the last chunk.

Modifications:

* Check if the appropriate write operation was successful when reporting the last chunk
* Skip writing chunks if the write operation was already marked as "done"

Result:

Fixes https://github.com/netty/netty/issues/8700.

### DIFF
--- a/handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java
+++ b/handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java
@@ -209,6 +209,21 @@ public class ChunkedWriteHandler extends ChannelDuplexHandler {
             if (currentWrite == null) {
                 break;
             }
+
+            if (currentWrite.promise.isDone()) {
+                // This might happen e.g. in the case when a write operation
+                // failed, but there're still unconsumed chunks left.
+                // Most chunked input sources would stop generating chunks
+                // and report end of input, but this doesn't work with any
+                // source wrapped in HttpChunkedInput.
+                // Note, that we're not trying to release the message/chunks
+                // as this had to be done already by someone who resolved the
+                // promise (using ChunkedInput.close method).
+                // See https://github.com/netty/netty/issues/8700.
+                this.currentWrite = null;
+                continue;
+            }
+
             final PendingWrite currentWrite = this.currentWrite;
             final Object pendingMessage = currentWrite.msg;
 
@@ -264,9 +279,13 @@ public class ChunkedWriteHandler extends ChannelDuplexHandler {
                     f.addListener(new ChannelFutureListener() {
                         @Override
                         public void operationComplete(ChannelFuture future) throws Exception {
-                            currentWrite.progress(chunks.progress(), chunks.length());
-                            currentWrite.success(chunks.length());
-                            closeInput(chunks);
+                            if (!future.isSuccess()) {
+                                closeInput(chunks);
+                                currentWrite.fail(future.cause());
+                            } else {
+                                currentWrite.progress(chunks.progress(), chunks.length());
+                                currentWrite.success(chunks.length());
+                            }
                         }
                     });
                 } else if (channel.isWritable()) {
--- a/handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
+++ b/handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
@@ -21,8 +21,11 @@ import io.netty.buffer.Unpooled;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelPromise;
+import io.netty.channel.ChannelOutboundHandlerAdapter;
 import io.netty.channel.embedded.EmbeddedChannel;
 import io.netty.util.CharsetUtil;
+import io.netty.util.ReferenceCountUtil;
 import org.junit.Test;
 
 import java.io.ByteArrayInputStream;
@@ -31,10 +34,9 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.channels.Channels;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
 
 public class ChunkedWriteHandlerTest {
     private static final byte[] BYTES = new byte[1024 * 64];
@@ -162,8 +164,7 @@ public class ChunkedWriteHandlerTest {
 
         EmbeddedChannel ch = new EmbeddedChannel(new ChunkedWriteHandler());
         ch.writeAndFlush(input).addListener(listener).syncUninterruptibly();
-        ch.checkException();
-        ch.finish();
+        assertTrue(ch.finish());
 
         // the listener should have been notified
         assertTrue(listenerNotified.get());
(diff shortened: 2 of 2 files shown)


kachayev:
Sure!

kachayev:
@normanmaurer Actually... From what I can tell, there are no test cases to cover input close and failure propagations at all. I think it makes more sense to cover all of them (including `discard` of pending writes in the queue). Should I add them to this PR or maybe open a separate one?

normanmaurer:
Just add them here.... thanks

kachayev:
Deal!

handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java
```diff
@@ -209,6 +209,18 @@ private void doFlush(final ChannelHandlerContext ctx) {
...
             }
+
+            if (currentWrite.promise.isDone()) {
+                // This might happen e.g. in the case when a write operation
+                // failed, but there're still unconsumed chunks left.
+                // Most chunked input sources would stop generating chunks
+                // and report end of input, but this doesn't work with any
+                // source wrapped in HttpChunkedInput.
+                // See https://github.com/netty/netty/issues/8700.
+                this.currentWrite = null;
```
normanmaurer: do we also need to release the message ?

kachayev: Not here... I assume someone who resolved the promise (either with success or with failure) have already done this by calling `closeInput`. As we're not allocating any new chunks (calling `readChunk`), everything else should be the responsibility of a specific input implementation.

normanmaurer: good point. Maybe add a comment ?

kachayev: Will do!

normanmaurer:
@kachayev please ping me once I should review again

kachayev:
@normanmaurer Sure, I will

kachayev:
@normanmaurer Added 12 new test cases.

normanmaurer:
@kachayev <3

normanmaurer:
@netty-bot test this please

handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
```diff
@@ -227,6 +228,207 @@ public long progress() {
...
+    public void testFailureWhenLastChunkFailed() throws IOException {
+        ChannelOutboundHandlerAdapter failLast = new ChannelOutboundHandlerAdapter() {
+            private int passedWrites;
+
+            @Override
+            public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) {
+                if (++this.passedWrites < 4) {
+                    ctx.write(msg, promise);
+                } else {
+                    promise.tryFailure(new RuntimeException());
```
normanmaurer: You also need to call `ReferenceCountUtil.release(msg)` before failing the promise.

handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
```diff
@@ -227,6 +228,207 @@ public long progress() {
...
+                }
+            }
+        };
+
+        EmbeddedChannel ch = new EmbeddedChannel(failLast, new ChunkedWriteHandler());
+        ChannelFuture r = ch.writeAndFlush(new ChunkedFile(TMP, 1024 * 16)); // 4 chunks
+        ch.close().awaitUninterruptibly();
```
normanmaurer: just use `ch.finish()` and assert the return value.

handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
```diff
@@ -227,6 +228,207 @@ public long progress() {
...
+            }
+        };
+
+        EmbeddedChannel ch = new EmbeddedChannel(new ChunkedWriteHandler());
+
+        // write 3 messages and close channel before flushing
+        ChannelFuture r1 = ch.write(new ChunkedFile(TMP));
+        ChannelFuture r2 = ch.write(new ChunkedNioFile(TMP));
+        ch.write(notifiableInput);
+        ch.close().awaitUninterruptibly();
```
normanmaurer: just use `ch.finish()` and assert the return value.

handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
```diff
@@ -227,6 +228,207 @@ public long progress() {
...
+            @Override
+            public long progress() {
+                return 1;
+            }
+        };
+
+        ChannelOutboundHandlerAdapter noOpWrites = new ChannelOutboundHandlerAdapter() {
+            @Override
+            public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) {
+                promise.tryFailure(new RuntimeException());
```
normanmaurer: You also need to call `ReferenceCountUtil.release(msg)` before failing the promise.

handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
```diff
@@ -227,6 +228,207 @@ public long progress() {
...
+            }
+        };
+
+        EmbeddedChannel ch = new EmbeddedChannel(noOpWrites, new ChunkedWriteHandler());
+        ch.writeAndFlush(nonClosableInput).awaitUninterruptibly();
+        ch.close();
```
normanmaurer: just use `ch.finish()` and assert the return value.

handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
```diff
@@ -255,4 +457,64 @@ private static void check(Object... inputs) {
 
         assertEquals(BYTES.length * inputs.length, read);
     }
+
+    private static void checkFirstFailed(Object input) {
+        ChannelOutboundHandlerAdapter noOpWrites = new ChannelOutboundHandlerAdapter() {
+            @Override
+            public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) {
+                promise.tryFailure(new RuntimeException());
```
normanmaurer: Call `ReferenceCountUtil.release(msg)` before failing the promise.

handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
```diff
@@ -255,4 +457,64 @@ private static void check(Object... inputs) {
...
+            }
+        };
+
+        EmbeddedChannel ch = new EmbeddedChannel(noOpWrites, new ChunkedWriteHandler());
+        ChannelFuture r = ch.writeAndFlush(input).awaitUninterruptibly();
+        ch.close();
```
normanmaurer: just use `ch.finish()` and assert the return value.

handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
```diff
@@ -255,4 +457,64 @@ private static void check(Object... inputs) {
...
+    private static void checkSkipFailed(Object input1, Object input2) {
+        ChannelOutboundHandlerAdapter failFirst = new ChannelOutboundHandlerAdapter() {
+            private boolean alreadyFailed;
+
+            @Override
+            public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) {
+                if (alreadyFailed) {
+                    ctx.write(msg, promise);
+                } else {
+                    promise.tryFailure(new RuntimeException());
```
normanmaurer: call `ReferenceCountUtil.release(msg)` before failing the promise.

handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
```diff
@@ -255,4 +457,64 @@ private static void check(Object... inputs) {
...
+                    this.alreadyFailed = true;
+                }
+            }
+        };
+
+        EmbeddedChannel ch = new EmbeddedChannel(failFirst, new ChunkedWriteHandler());
+        ChannelFuture r1 = ch.write(input1);
+        ChannelFuture r2 = ch.writeAndFlush(input2).awaitUninterruptibly();
+        ch.close();
```
normanmaurer: just use `ch.finish()` and assert the return value.

normanmaurer:
@kachayev after my comments are addressed this should pass to build on the CI as it currently fails because of some buffer leaks.

handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
kachayev: @normanmaurer `ch.finish` would return `false` because we effectively don't do any writes here. All messages are still in `ChunkedWriteHandler` queue, until `flush`. Meaning both `inboundMessages` and `outboundMessages` queues are empty.

kachayev:
@normanmaurer I've added `release` calls. Sorry, missed those in the first place. I also switched to `ch.finish()` where reasonable. There are still a few tests where we don't do outbound writes at all because of `ChunkedWriteHandler` internal machinery (commented them out in the code).

handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
normanmaurer: @kachayev sure which is fine just use `assertFalse(...)` as this is what you expect, or what I am missing ?

kachayev: @normanmaurer Oh, yeah. You're right! 🤦‍♂️It's just mentally hard to read `assertFalse(ch.finish())` as "all good, just no messages" :) I'll leave a comment why this is expected.

kachayev: @normanmaurer Update once again. I also "fixed" a couple of older tests to be consistent in terms of `ch.finish()` usage.

normanmaurer:
@netty-bot test this please

normanmaurer:
@kachayev thanks a lot... merged

PR #8720, opened 2019-01-15
cleanup: extract duplicate code into method

kezhenxu94:
Motivation:

Clean up code to increase readability.

Modification:

Extract duplicate code blocks into method.


normanmaurer:
@kezhenxu94 thanks <3

Issue #8511, opened 2018-11-13
[Netty5] ChannelPipeline mutating methods signature


normanmaurer:
All the methods that mutate the `ChannelPipeline` are currently “synchronous”. This means all manipulation of the pipeline must complete before modifier methods return, and the`ChannelPipeline` implementation must therefore account for concurrency. All of the interaction with the `ChannelPipeline` is inside the `EventLoop` and we can therefore delay modification until we are in the `EventLoop`. This will reduce complexity of the `ChannelPipeline` implementation, and also remove the current “best effort” concurrency code in `AbstractChannelHandlerContext`. 

There are currently 2 different ideas here: 
1) Have all methods just return `ChannelFuture` which completes when the pipeline is modified (e.g. on the `EventLoop`). If we do this we also most likely should allow to pass in a `ChannelPromise` that is used for notifications. 
2) Only allow pipeline modifications on the `EventLoop`. 

I am currently in favour of 1) as this is also what we did in [SwiftNIO](https://swiftnio.io) with great success and still will allow people to just use `ChannelPipeline` from “any thread”.

ejona86:
I think this interacts with #8156.

njhill:
@normanmaurer is this obsolete in light of #8778?

Related to this and #8156, I had another idea w.r.t. the pipeline manipulation API design, i.e. how best to reference handlers after they are added. Why not have all the `addX`/`replace` methods return the new handler context, and move all of the `addBefore`, `addAfter`, `replace`, `remove` methods from the pipeline to the context?

I think this would address most/all of the frustrations raised in #8156, including most of the cartesian product explosion. You could do things such as:
```java
channel.pipeline().addFirst(handler1).addAfter(handler2, "myname").addAfter(handler3);

channel.pipeline().context("myname").replace(handler4);

MyHandler removed = channel.pipeline().context(MyHandler.class).remove();
```
etc.

If we still want to avoid the synchronization here, all these methods could instead return `Future<ChannelHandlerContext>` along the lines of your suggestion 1 above.

cc @ejona86

RoganDawes:
I quite like this approach to the problem of the Cartesian product of methods on ChannelPipeline.

One existing method that this approach does not seem to support is ```ch.pipeline().addLast(handler1, handler2, handler3);```
Which context would you return from such a method?

I think it would be reasonable, though, to require migration to ```ch.pipeline().addLast(handler1).addAfter(handler2).addAfter(handler3);```

[...]

PR #8723, opened 2019-01-16
    Skip execution of Channel*Handler method if annotated with @Skip and just use the next handler in the pipeline.

normanmaurer:
Motivation:

Invoking ChannelHandlers is not free and can result in some overhead when the ChannelPipeline becomes very long. This is especially true if most handlers will just forward the call to the next handler in the pipeline. When the user extends Channel*HandlerAdapter we can easily detect if can just skip the handler and invoke the next handler in the pipeline directly. This reduce the overhead of dispatch but also reduce the call-stack in many cases.

Modifications:

Detect if we can skip the handler when walking the pipeline.

Result:

Reduce overhead for long pipelines.

```
Benchmark                                       (extraHandlers)   Mode  Cnt       Score      Error  Units
DefaultChannelPipelineBenchmark.propagateEventOld             4  thrpt   10  267313.031 ± 9131.140  ops/s
DefaultChannelPipelineBenchmark.propagateEvent                4  thrpt   10  824825.673 ± 12727.594  ops/s
```

### DIFF
--- a/microbench/src/main/java/io/netty/microbench/channel/DefaultChannelPipelineBenchmark.java
+++ b/microbench/src/main/java/io/netty/microbench/channel/DefaultChannelPipelineBenchmark.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2019 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.microbench.channel;
+
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandler;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.channel.ChannelPipeline;
+import io.netty.channel.embedded.EmbeddedChannel;
+import io.netty.microbench.util.AbstractMicrobenchmark;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.infra.Blackhole;
+
+@Warmup(iterations = 5)
+@Measurement(iterations = 5)
+@State(Scope.Benchmark)
+public class DefaultChannelPipelineBenchmark extends AbstractMicrobenchmark {
+
+    private static final ChannelInboundHandler NOOP_HANDLER = new ChannelInboundHandlerAdapter() {
+        @Override
+        public boolean isSharable() {
+            return true;
+        }
+    };
+
+    private static final ChannelInboundHandler CONSUMING_HANDLER = new ChannelInboundHandlerAdapter() {
+        @Override
+        public void channelReadComplete(ChannelHandlerContext ctx) {
+            // NOOP
+        }
+
+        @Override
+        public boolean isSharable() {
+            return true;
+        }
+    };
+
+    @Param({ "4" })
+    public int extraHandlers;
+
+    private ChannelPipeline pipeline;
+
+    @Setup(Level.Iteration)
+    public void setup() {
+        pipeline = new EmbeddedChannel().pipeline();
+        for (int i = 0; i < extraHandlers; i++) {
+            pipeline.addLast(NOOP_HANDLER);
+        }
+        pipeline.addLast(CONSUMING_HANDLER);
+    }
+
+    @TearDown
+    public void tearDown() {
+        pipeline.channel().close();
+    }
+
+    @Benchmark
+    public void propagateEvent(Blackhole hole) {
+        for (int i = 0; i < 100; i++) {
+            hole.consume(pipeline.fireChannelReadComplete());
+        }
+    }
+}
--- a/transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
+++ b/transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
@@ -23,7 +23,9 @@ import io.netty.util.Recycler;
 import io.netty.util.ReferenceCountUtil;
 import io.netty.util.ResourceLeakHint;
 import io.netty.util.concurrent.EventExecutor;
+import io.netty.util.concurrent.FastThreadLocal;
 import io.netty.util.concurrent.OrderedEventExecutor;
+import io.netty.util.internal.PlatformDependent;
 import io.netty.util.internal.PromiseNotificationUtil;
 import io.netty.util.internal.ThrowableUtil;
 import io.netty.util.internal.ObjectUtil;
@@ -33,6 +35,10 @@ import io.netty.util.internal.logging.InternalLogger;
 import io.netty.util.internal.logging.InternalLoggerFactory;
 
 import java.net.SocketAddress;
+import java.security.AccessController;
+import java.security.PrivilegedExceptionAction;
+import java.util.Map;
+import java.util.WeakHashMap;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
 abstract class AbstractChannelHandlerContext extends DefaultAttributeMap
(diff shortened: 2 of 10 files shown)


normanmaurer:
While this change is against the master branch it can easily be applied to 4.1 as well.

@njhill @franz1981 @vietj FYI as well.

normanmaurer:
Also please note that once we require java8 (or later) for master we will most likely just implement all the "default" stuff in the interfaces directly which makes the code even nicer.

normanmaurer:
Some of the code was re-surrected from https://github.com/netty/netty/tree/master_deprecated 👍

normanmaurer:
@tjake @stef1927 @nitsanw @smaldini @rkapsi you may be interested as well :)

transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
```diff
@@ -846,7 +992,16 @@ private void invokeWriteAndFlush(Object msg, ChannelPromise promise) {
     }
 
     private void write(Object msg, boolean flush, ChannelPromise promise) {
-        AbstractChannelHandlerContext next = findContextOutbound();
+        final AbstractChannelHandlerContext next;
+        if (flush) {
+            AbstractChannelHandlerContext ctx = this;
+            do {
+                ctx = ctx.prev;
+            } while ((ctx.executionMask & MASK_WRITE) == 0 && (ctx.executionMask & MASK_FLUSH) == 0);
```
nitsanw: Isn't `(ctx.executionMask & MASK_WRITE) == 0 && (ctx.executionMask & MASK_FLUSH) == 0` the same as:
`(ctx.executionMask & (MASK_WRITE|MASK_FLUSH)) == 0`

transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
```diff
@@ -846,7 +992,16 @@ private void invokeWriteAndFlush(Object msg, ChannelPromise promise) {
     }
 
     private void write(Object msg, boolean flush, ChannelPromise promise) {
-        AbstractChannelHandlerContext next = findContextOutbound();
+        final AbstractChannelHandlerContext next;
```
nitsanw: Can we somehow cache the result here?

transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
normanmaurer: @nitsanw yes! you are right ....

transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
normanmaurer: Sorry but what you want to cache ? We already cache the bitmask.

nitsanw: I mean cache the result of this scan. The masks do not change, and the chain can change in theory, but usually doesn't. So the result of the loop is mostly the same, no?

normanmaurer: @nitsanw yeah it can change so I think we should not for now. That said we can easily do these kind of things once I will do https://github.com/netty/netty/issues/8511 as then we can remove all the volatile and can be sure all the changes are done in the EventLoop. In this case we can then just recalculate these things whenever the pipeline is mutated.

transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
```diff
@@ -977,19 +1123,19 @@ private boolean isNotValidPromise(ChannelPromise promise, boolean allowVoidPromi
         return false;
     }
 
-    private AbstractChannelHandlerContext findContextInbound() {
+    private AbstractChannelHandlerContext findContextInbound(int mask) {
         AbstractChannelHandlerContext ctx = this;
         do {
             ctx = ctx.next;
-        } while (!ctx.inbound);
+        } while ((ctx.executionMask & mask) == 0);
```
doom369: `(ctx.executionMask & mask) == 0` move to own method?

microbench/src/main/java/io/netty/microbench/channel/DefaultChannelPipelineBenchmark.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @TearDown
+    public void tearDown() {
+        pipeline.channel().close();
+    }
+
+    @Benchmark
+    public void propagateEvent(Blackhole hole) {
+        for (int i = 0; i < 100; i++) {
+            pipeline.fireChannelReadComplete();
```
doom369: Maybe add some counter inside the handler just to be 100% sure the compiler doesn't throw this away.

transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
```diff
@@ -977,19 +1123,19 @@ private boolean isNotValidPromise(ChannelPromise promise, boolean allowVoidPromi
...
         return ctx;
     }
 
-    private AbstractChannelHandlerContext findContextOutbound() {
+    private AbstractChannelHandlerContext findContextOutbound(int mask) {
         AbstractChannelHandlerContext ctx = this;
         do {
             ctx = ctx.prev;
-        } while (!ctx.outbound);
+        } while ((ctx.executionMask & mask) == 0);
```
doom369: Test that checks that handlers are actually skipped with `@Skip` and not skipped without it?

transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
```diff
@@ -63,8 +69,43 @@
      */
     private static final int INIT = 0;
 
-    private final boolean inbound;
-    private final boolean outbound;
+    // Using to mask which methods must be called for a ChannelHandler.
+    private static int NONE;
```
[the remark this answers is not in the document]
normanmaurer: check style fails when I do this

transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
```diff
@@ -63,8 +69,43 @@
...
+    private static int MASK_CHANNEL_WRITABILITY_CHANGED = 1 << 8;
+    private static int MASK_BIND = 1 << 9;
+    private static int MASK_CONNECT = 1 << 10;
+    private static int MASK_DISCONNECT = 1 << 11;
+    private static int MASK_CLOSE = 1 << 12;
+    private static int MASK_REGISTER = 1 << 13;
+    private static int MASK_DEREGISTER = 1 << 14;
+    private static int MASK_READ = 1 << 15;
+    private static int MASK_WRITE = 1 << 16;
+    private static int MASK_FLUSH = 1 << 17;
```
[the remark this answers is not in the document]
normanmaurer: doh... great catch!

transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
normanmaurer: good idea..

transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
normanmaurer: I actually like to keep it explicit here

transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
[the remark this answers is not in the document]
normanmaurer: but combined with final it works :)

microbench/src/main/java/io/netty/microbench/channel/DefaultChannelPipelineBenchmark.java
normanmaurer: I will just consume the return value of `fireChannelReadComplete()` in the blackhole to be sure.

transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
nitsanw: A volatile load will be cheaper than recalc. And since adding and removing handlers already has some exclusion/order guarantees(I would think) we just need to ride on that.

normanmaurer: I am not sure about this and would like to consider thousand a follow up (as we did not cache before as well). Another downside is object size. As there are a lot of context instances we would like to keep the memory overhead as low as possible.

transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
normanmaurer: @doom369 done

normanmaurer:
The test failure is not related and is caused by a flaky test. Will fix this in another pr.

normanmaurer:
@netty-bot test this please

normanmaurer:
@netty-bot test this please

normanmaurer:
Once this one is in I will do a followup PR to use the `@Skip` annotation where it makes sense :)

doom369:
Had the same thought, but didn't find good candidates after the quick overview, maybe I was looking in the wrong place.

normanmaurer:
@doom369 most of our handlers just extend the `*Adapters` which means they get it for free. That said I think there are a few which directly implement the interfaces (partly), like for example `SslHandler`.

transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
```diff
@@ -84,16 +124,122 @@
...
+            cache.put(clazz, mask);
+        }
+        return mask;
+    }
+
+    /**
+     * Calculate the {@code executionMask}.
+     */
+    private static int mask0(Class<? extends ChannelHandler> handlerType) {
+        int mask = MASK_EXCEPTION_CAUGHT;
```
johnou: shouldn't this init as 0?

normanmaurer: @doom369 nope as `exceptionCaught` is part of `ChannelHandler`.

transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
```diff
@@ -63,8 +69,42 @@
...
+    private static final int MASK_WRITE = 1 << 16;
+    private static final int MASK_FLUSH = 1 << 17;
+
+    private static final int MASK_ALL_INBOUND = MASK_CHANNEL_REGISTERED |
+            MASK_CHANNEL_UNREGISTERED | MASK_CHANNEL_ACTIVE | MASK_CHANNEL_INACTIVE | MASK_CHANNEL_READ |
+            MASK_CHANNEL_READ_COMPLETE | MASK_USER_EVENT_TRIGGERED | MASK_CHANNEL_WRITABILITY_CHANGED;
+    private static final int MASK_ALL_OUTBOUND = MASK_BIND | MASK_CONNECT | MASK_DISCONNECT |
+            MASK_CLOSE | MASK_REGISTER | MASK_DEREGISTER | MASK_READ | MASK_WRITE | MASK_FLUSH;
+
+    private static final FastThreadLocal<Map<Class<? extends ChannelHandler>, Integer>> MASKS =
```
njhill: Might [`java.lang.ClassValue<Integer>`](https://docs.oracle.com/javase/8/docs/api/java/lang/ClassValue.html) work here? Would be interesting to benchmark it.

normanmaurer: Interesting, never heard of this one. Will check out

normanmaurer:
@netty-bot Test this please

transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
doom369: It is only in Java 7+

normanmaurer: Which would be fine for netty 5

normanmaurer:
@netty-bot test this please

normanmaurer:
@netty-bot test this please

normanmaurer:
@netty-bot Test this please

transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
normanmaurer: @doom369 master is now java8+

normanmaurer: @njhill I merged it as it is for now... we can do a followup pr if needed :)

normanmaurer:
@netty/contributors do we want to back port this to 4.1 as well ?

njhill:
> do we want to back port this to 4.1 as well ?

@normanmaurer what would be the reason to not do so? More generally is it the case that 4.1 should now be considered in "maintenance mode" i.e. only bugfixes, or will new features still be added to both branches (apart from those which are only applicable to 5.x of course)?

normanmaurer:
@njhill thats a good question... I think in general we should be more conservative in adding new features / perf improvements to 4.1

Issue #8707, opened 2019-01-07
OOME in SslHandlerCoalescingBufferQueue w/ H2


rkapsi:
I have opened similar/same tickets (see #7659 and #8265) before but ended up closing them as they appeared to be user errors. It happens very rarely and trying to add instrumentation could be tricky. We see maybe 1 occurrence once or maybe even only every other week across a "large" (relatively speaking) fleet of Netty servers.

The theory was that flush is being called on a closed connection for example but I've added various guards against it and you can see on the screenshot below that the channel is active.

Basically, when it happen then something is filling up the SslHandlerCoalescingBufferQueue with millions of `Unpooled#EMPTY_BUFFER` very fast until an OOME occurs. It appears to be tied to H2.

Below is a stack trace which I find interesting and a view of the heap dump. From the state of the pipeline and access logs (we log certain TCP and TLS events) I can see that the client/server triggered this condition after the handshake but before any requests were received from the client.

A few places that I find interesting are:

DefaultHttp2RemoteFlowController line 626
FlowControlledData lines 411/412

 In particular the latter which may be the source for the EMPTY_BUFFER but I don't see (yet) the condition why it'd loop (although DefaultHttp2RemoteFlowController line 626 is technically inside an infinite loop).

```java
Stack traces at the moment of snapshot capture

HttpsWorkerThread-11 State: BLOCKED tid: 148
io.netty.channel.AbstractCoalescingBufferQueue.toChannelFutureListener(ChannelPromise) AbstractCoalescingBufferQueue.java:375
io.netty.channel.AbstractCoalescingBufferQueue.add(ByteBuf, ChannelPromise) AbstractCoalescingBufferQueue.java:84
io.netty.handler.ssl.SslHandler.write(ChannelHandlerContext, Object, ChannelPromise) SslHandler.java:759
io.netty.channel.AbstractChannelHandlerContext.invokeWrite0(Object, ChannelPromise) AbstractChannelHandlerContext.java:738
io.netty.channel.AbstractChannelHandlerContext.invokeWrite(Object, ChannelPromise) AbstractChannelHandlerContext.java:730
io.netty.channel.AbstractChannelHandlerContext.write(Object, boolean, ChannelPromise) AbstractChannelHandlerContext.java:816
io.netty.channel.AbstractChannelHandlerContext.write(Object, ChannelPromise) AbstractChannelHandlerContext.java:723
[...]

PR #8726, opened 2019-01-17
Ensure FlowControlled data frames will be correctly removed from the …

normanmaurer:
…flow-controller when a write error happens.

Motivation:

When a write error happens during writing of flowcontrolled data frames we miss to correctly detect this in the write loop which may result in an infinite loop as we will never detect that the frame should be removed from the queue.

Modifications:

- When we fail a flowcontrolled data frame we ensure that the next frame.write(...) call will signal back that the whole frame was handled and so can be removed.
- Add unit test.

Result:

Fixes https://github.com/netty/netty/issues/8707.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
@@ -416,13 +416,13 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
                         //
                         // See https://github.com/netty/netty/issues/8707.
                         padding = dataSize = 0;
-                        return;
+                    } else {
+                        // There's no need to write any data frames because there are only empty data frames in the
+                        // queue and it is not end of stream yet. Just complete their promises by getting the buffer
+                        // corresponding to 0 bytes and writing it to the channel (to preserve notification order).
+                        ChannelPromise writePromise = ctx.newPromise().addListener(this);
+                        ctx.write(queue.remove(0, writePromise), writePromise);
                     }
-                    // There's no need to write any data frames because there are only empty data frames in the queue
-                    // and it is not end of stream yet. Just complete their promises by getting the buffer corresponding
-                    // to 0 bytes and writing it to the channel (to preserve notification order).
-                    ChannelPromise writePromise = ctx.newPromise().addListener(this);
-                    ctx.write(queue.remove(0, writePromise), writePromise);
                     return;
                 }
 
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
@@ -397,6 +397,9 @@ public class DefaultHttp2ConnectionEncoder implements Http2ConnectionEncoder {
             queue.releaseAndFailAll(cause);
             // Don't update dataSize because we need to ensure the size() method returns a consistent size even after
             // error so we don't invalidate flow control when returning bytes to flow control.
+            //
+            // That said we will set dataSize and padding to 0 in the write(...) method if we cleared the queue
+            // because of an error.
             lifecycleManager.onError(ctx, true, cause);
         }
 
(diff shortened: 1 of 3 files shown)


rkapsi:
@normanmaurer LGTM. I went through a few memory dumps I had saved (September 5, 2018 being oldest) and the condition `queue.isEmpty()` is true in all cases. That should do the trick. Best Netty release ever!

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -405,6 +408,16 @@ public void write(ChannelHandlerContext ctx, int allowedBytes) {
...
                 if (queuedData == 0) {
+                    if (queue.isEmpty()) {
+                        // When the queue is empty it means we did clear it because of an error(...) call
+                        // (as otherwise we will have at least 1 entry in there), which will happen either when called
+                        // explicit or when the write itself fails. In this case just set dataSize and padding to 0
+                        // which will signal back that the whole frame was consumed.
+                        //
+                        // See https://github.com/netty/netty/issues/8707.
+                        padding = dataSize = 0;
+                        return;
```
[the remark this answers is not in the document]
normanmaurer: sure why not...

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -405,6 +408,16 @@ public void write(ChannelHandlerContext ctx, int allowedBytes) {
...
             if (!endOfStream) {
                 if (queuedData == 0) {
+                    if (queue.isEmpty()) {
+                        // When the queue is empty it means we did clear it because of an error(...) call
+                        // (as otherwise we will have at least 1 entry in there), which will happen either when called
+                        // explicit or when the write itself fails. In this case just set dataSize and padding to 0
+                        // which will signal back that the whole frame was consumed.
+                        //
+                        // See https://github.com/netty/netty/issues/8707.
+                        padding = dataSize = 0;
```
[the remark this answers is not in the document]
normanmaurer: @Scottmitch will check and see if I can add more tests. That said I think it should work as expected as we only reset after `write` is called and we failed it before.

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
```diff
@@ -416,13 +416,13 @@ public void write(ChannelHandlerContext ctx, int allowedBytes) {
                         //
                         // See https://github.com/netty/netty/issues/8707.
                         padding = dataSize = 0;
-                        return;
+                    } else {
+                        // There's no need to write any data frames because there are only empty data frames in the
+                        // queue and it is not end of stream yet. Just complete their promises by getting the buffer
+                        // corresponding to 0 bytes and writing it to the channel (to preserve notification order).
+                        ChannelPromise writePromise = ctx.newPromise().addListener(this);
```
rkapsi: Wonder if we want to move the `addListener()` after the `write(...)` as well. Right now it's skewing the stack should the `remove(...)` complete the promise.

normanmaurer: Not sure I understand... can you show me some code ?

rkapsi: Something like that...

```java
ChannelPromise writePromise = ctx.newPromise();
ctx.write(queue.remove(0, writePromise), writePromise)
    .addListener(this);
```

Otherwise if `queue.remove(0, writePromise)` was to complete the promise it'd potentially notify the listener first, followed by calling `ctx.write(...)`. Stack (traces) will look strange. Not sure there would be other side effects due to the order in which things execute but I think you want the listener to get notified after the write.

normanmaurer: Got it... I would like to investigate as a followup here. Trying to keep changes as minimal as possible atm

codec-http2/src/main/java/io/netty/handler/codec/http2/DefaultHttp2ConnectionEncoder.java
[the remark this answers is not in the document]
normanmaurer: @rkapsi @Scottmitch I was able to add a unit test that would result in the endless loop before the fix:

https://github.com/netty/netty/pull/8726/commits/3d2cff4cd9d8167bc3dd8f88956155b41a67deca

normanmaurer:
@netty-bot test this please

normanmaurer:
@netty-bot test this please

PR #8735, opened 2019-01-19
Fix racy ChannelOutboundBuffer.testWriteTaskRejected test.

normanmaurer:
Motivation:

testWriteTaskRejected was racy as we did not ensure we dispatched all events to the executor before shutting it down.

Modifications:

Add a latch to ensure we dispatched everything.

Result:

Fix racy test that failed sometimes before.


normanmaurer:
Let me merge this so the CI is happy again

PR #8738, opened 2019-01-21
Fix flaky ChannelInitializerTest.testChannelInitializerEventExecutor()

normanmaurer:
Motivation:

testChannelInitializerEventExecutor() did sometimes fail as we sometimes miss to count down the latch. This can happen when we remove the handler from the pipeline before channelUnregistered(...) was called for it.

Modifications:

Countdown the latch in handlerRemoved(...).

Result:

Fix flaky test.


normanmaurer:
Merging as it is only a simple fix in the test

Issue #8540, opened 2018-11-13
[Netty 5] Use Java11 as minimum Java version


normanmaurer:
Netty 5 will move to Java 11 as minimum. 

This is driven by the fact that we assume that the development of Netty 5 will take some time and for people that can not switch to Java11 there will still be support for netty 4.1.

smaldini:
Bold but interesting choice ! Java 11 is far from mainstream but it is all relative to 5.0 timeline as well so I am not too worried. We can still evaluate by then if we really needed any Java 9-11 API to achieve what we want. It seems that Finalizer/Cleaner/RefCounted might be one.

I believe the non Java 8 API tho will be pretty limited but if that happens, I wonder if we should have a 4.2 both introducing new features/changes, along with deprecations and requiring Java 8 only.

normanmaurer:
@smaldini sounds fair enough... That said I can at least think of two things that are interesting for us which are both not in Java8.

* Compact string to maybe replace `AsciiString` : 
https://github.com/netty/netty/issues/8548
* Cleaner to replace `finalize()` as these are for example not supported on substrate VM : 
https://github.com/netty/netty/issues/8542

smaldini:
Yeah so we need to look at AsciiString vs String but that seems like a small price to pay to keep compat if it was the only thing. That's up for debate tho.
About `finalize()` I would hope we could get away without using them. Implicit behavior depending on non deterministic GC behavior seems dangerous enough that maybe the best way would be to require explicit releasing. However doable it seems to be with 4 use cases (HashedWheelTimer, OpenSslContext, OpenSslEngine and PoolArena) there would be 1 case that might be tricky because leaky : Recycler. We already educate user to pay attention to `refcountable#release` calls and management of event loop / pool resources so it shouldn't be too far a gap to maybe transition to explicit behavior here too.

ejona86:
Java 11 may be a bit too aggressive for gRPC to use it soon, as we may need to support "the last two LTS releases." That would mean it'd be 3 years from now before gRPC could use Netty 5. Our support timelines are still a bit up-in-the-air, though.

[...]

PR #8739, opened 2019-01-21
Require Java8 as minimum

normanmaurer:
Motivation:

While we are not yet quite sure if we want to require Java11 as minimum we are at least sure we want to use java8 as minimum.

Modifications:

Change minimum version to java8 and update some tests which failed compilation after this change.

Result:

Use Java8 as minimum and be able to use Java8 features.

### DIFF
--- a/codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketServerHandshaker00Test.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketServerHandshaker00Test.java
@@ -15,6 +15,7 @@
  */
 package io.netty.handler.codec.http.websocketx;
 
+import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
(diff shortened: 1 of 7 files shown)


normanmaurer:
I did not replace anything with java8 APIs here yet as I think we want to do this step by step.

normanmaurer:
Related to https://github.com/netty/netty/issues/8540

doom369:
Omg! This is so cool. Thanks! So many code could be just removed now :).

kezhenxu94:
> I did not replace anything with java8 APIs here yet as I think we want to do this step by step.

I'm glad to help if needed

normanmaurer:
@kezhenxu94 please.... This is for the master branch only :)

kezhenxu94:
> @kezhenxu94 please.... This is for the master branch only :)

@normanmaurer Now I begin wondering when should I open a PR based on branch master and when on branch 4.1?

normanmaurer:
@kezhenxu94 master is usually the way to go

PR #8741, opened 2019-01-21
fix typo

kezhenxu94:
Motivation:

Correct typo 

Modification:

Correct typo

Result:

JavaDoc and method name are more readable


normanmaurer:
@kezhenxu94 thanks!

PR #8745, opened 2019-01-22
Java 8 migration. Replace Netty ConcurrentSet with Java KeySet.

doom369:
Motivation:

The concurrent set is present in Java 8 and above so we can use it instead of own implementation.

Modification:

io.netty.utik.internal.ConcurrentSet replaced with `ConcurrentHashMap.newKeySet()`.


normanmaurer:
@netty-bot test this please

normanmaurer:
@doom369 thanks a lot!

PR #8746, opened 2019-01-22
Leave responsibility to choose EventExecutor to the user

normanmaurer:
Motivation:

We should leave the responsibility to choose the EventExecutor for a ChannelHandler to the user for more flexibility and to keep things simple.

Modification:

- Change method signatures to take an EventExecutor and not an EventExecutorGroup
- Remove special ChannelOption that allowed to enable / disable EventExecutor pinning

Result:

Simpler and more flexible code.


normanmaurer:
At the end we may remove the ability to specify the `EventExecutor` at all but that is not for sure yet. This is a first step to simplify things at the end.

normanmaurer:
@netty-bot test this please

normanmaurer:
@ejona86 general speaking I think if the user wants to use the same EventExecutor for multiple handlers he / she should specify it. No need to complicate our own implementation.

That said we may even remove the overload all together, but I need to exerpiment a bit more first.

ejona86:
> if the user wants to use the same EventExecutor for multiple handlers he / she should specify it

The API is for using a different EventExecutor compared to the one for the Channel. It doesn't imply it will be the same for multiple handlers. It may just be for a single handler, like moving TLS processing to a different thread.

I can agree with removing `SINGLE_EVENTEXECUTOR_PER_GROUP`; that's just ugliness and I agree that if the user wanted to use a single event loop they can pass it. But that doesn't require changing the API. (I honestly didn't realize that that optimization existed; I already thought it was my responsibility to pass a loop instead of a group when applicable.)

I guess changing the API makes it less error-prone, where a user accidentally passed a Group and then gets poor performance without them realizing it. Sure. SGTM

normanmaurer:
@ejona86 yep that was the idea... make it more "clear" ;) Thanks for the input!

carl-mastrangelo:
LGTM

PR #8747, opened 2019-01-22
Not used IntegerHolder class removed

doom369:
Motivation:

`IntergerHolder` was left from some old days and not used anymore.

Modification:

Remove `IntergerHolder` from the codebase.


normanmaurer:
@netty-bot test this please

PR #8748, opened 2019-01-22
Java 8 migration. Removed custom MathUtil.compare methods

doom369:
Motivation:

Netty uses own `Integer.compare` and `Long.compare` methods. Since Java 7 we can use Java implementation instead.


normanmaurer:
@netty-bot test this please

normanmaurer:
@doom369 thanks

PR #8750, opened 2019-01-22
Java 8 migration. Removed custom LongCounter and LongAdderCounter cla…

doom369:
…sses and replaced with the direct reference to LongAdder.

Motivation:

After switching to Java 8 we no longer need `LongCounter` and `LongAdderCounter`. We can directly replace them with `LongAdder`.

Modification:

Removed `LongCounter` and `LongAdderCounter` classes.


normanmaurer:
@netty-bot test this please

PR #8757, opened 2019-01-22
Java 8 migration. Use reference to StandardCharsets in CharsetUtil.

doom369:
Motivation:

No need to initialize charsets from the string. We can take already allocated charset from `StandardCharsets` class.

Modification:

Replace `Charset.forName("US-ASCII")` with `StandardCharsets.US_ASCII`.
Removed `Charset[] values()` method and internal static variable as it was used only in tests.


normanmaurer:
@netty-bot test this please

PR #8762, opened 2019-01-22
Java 8 migration. Remove ThreadLocalProvider and inline java.util.con…

doom369:
…current.ThreadLocalRandom.current() where necessary.

Motivation:

Custom Netty `ThreadLocalRandom` and `ThreadLocalRandomProvider` classes are no longer needed and can be removed.


normanmaurer:
@netty-bot test this please

PR #8763, opened 2019-01-22
Remove ability to specify name when add handler to the pipeline.

normanmaurer:
Motivation:

We previous allowed to add a handler with a given name to the pipeline and then later use this name to lookup the handler. This is really error-prone as names can clash.

Modifications:

- Remove methods that take name as an argument
- Let methods return ChannelHandlerContext after modification of the pipeline to allow lookup later on.
- Remove var-args method version as adding these handlers is not an atomic operation.

Result:

Less error-prone usage of the ChannelPipeline.

### DIFF
--- a/codec-haproxy/src/test/java/io/netty/handler/codec/haproxy/HAProxyMessageDecoderTest.java
+++ b/codec-haproxy/src/test/java/io/netty/handler/codec/haproxy/HAProxyMessageDecoderTest.java
@@ -42,11 +42,9 @@ public class HAProxyMessageDecoderTest {
 
     @Test
     public void testIPV4Decode() {
-        int startChannels = ch.pipeline().names().size();
         String header = "PROXY TCP4 192.168.0.1 192.168.0.11 56324 443\r\n";
         ch.writeInbound(copiedBuffer(header, CharsetUtil.US_ASCII));
         Object msgObj = ch.readInbound();
-        assertEquals(startChannels - 1, ch.pipeline().names().size());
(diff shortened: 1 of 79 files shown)


normanmaurer:
I would love to get some feedback on the general idea here from @netty/netty-5 ...

normanmaurer:
Also @johnou @doom369

normanmaurer:
And @franz1981 @njhill

normanmaurer:
So to make it short if people hate it I am more then happy to just move all of this into the trash ;)

doom369:
I'm using this approach mostly to avoid varargs and invocation of `DefaultChannelPipeline.generateName` for `addLast` method.

So adding `addLast(ChannelHandler handler)` solves first part, but not the second one.
I think leaving `addLast(ChannelHandler... handlers)` is still a good idea.
Also, maybe it would be good idea to add `ChannelHandler.getName()` with some default implementation?

normanmaurer:
@doom369 the problem with the name really is that it is kind of fragile and easy to mess up as names may clash (like for example you use some other projects code which will use the same naming) or names may change and you forget to update some code. I really think the only "useful" feature of the name may be for debugging / logging purposes.

doom369:
> I really think the only "useful" feature of the name may be for debugging / logging purposes.

That's correct. Pipeline with handler names is priceless for debugging and saves a lot for the time, especially for our use cases where we have a dynamic pipelines and few different types of pipeline.

normanmaurer:
After more thinking I agree this is not a good PR... closing

PR #8764, opened 2019-01-23
clean: remove never thrown

thinkerou:
Motivation:

`throws` are declared in a method's signature but never thrown.


kezhenxu94:
@thinkerou I don't think so, throwing exceptions in most of the methods' signature is intentionaly designed, meaning that it requires the invoker to deal possible exception. So removing them is not desired

normanmaurer:
@thinkerou what @kezhenxu94 said.

thinkerou:
@kezhenxu94 @normanmaurer thank your explain!

PR #8758, opened 2019-01-22
Probable bug. Endless recursion in CompositeByteBuf

doom369:
Motivation:

During review found strange code:

```
    public int toComponentIndex(int offset) {
        checkIndex(offset);
        return toComponentIndex(offset);
    }
```

Seems like it was a typo and code should be 

```
    public int toComponentIndex(int offset) {
        checkIndex(offset);
        return toComponentIndex0(offset);
    }
```

Introduced with https://github.com/netty/netty/pull/8437. FYI @njhill

PR #8773, opened 2019-01-23
Fix three bugs in CompositeByteBuf

njhill:
Motivation

In #8758, @doom369 reported an infinite loop bug in `CompositeByteBuf` which was introduced in #8437.

This is the same small fix for that, along with fixes for two other bugs (also introduced in #8437 :disappointed:) found while re-inspecting the changes and adding unit tests.

Modifications

- Replace recursive call to `toComponentIndex` with `toComponentIndex0` as intended
- Add missed `lastAccessed` racy cache invalidation in `capacity(int)` method
- Fix incorrect determination of initial offset in non-zero cIndex case of `updateComponentOffsets` method
- New unit tests for previously uncovered methods

Results

Fewer bugs.


normanmaurer:
@netty-bot test this please

PR #8751, opened 2019-01-22
Java 8 migration: replace anonymous class with lambda

kezhenxu94:
Motivation:

Using lambda instead of anonymous inner class to imporve readablity

Modification:

Replace anonymous inner class with lambda

PR #8776, opened 2019-01-24
Migrate Java8: replace with Map#computeIfAbsent where possible

kezhenxu94:
Motivation:

There is a convenient m ethod to check the absence of a key and put if absent in Java8, this patch replaces the boilerplate code with the new method where possible.

Modification:

replace with Map#computeIfAbsent where possible

Result:

Codebase becomes more clean and readable

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
@@ -30,6 +30,7 @@ import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
 
 import static io.netty.util.internal.ObjectUtil.*;
 import static io.netty.util.internal.StringUtil.*;
@@ -256,11 +257,13 @@ public class QueryStringDecoder {
         }
         String name = decodeComponent(s, nameStart, valueStart - 1, charset, false);
         String value = decodeComponent(s, valueStart, valueEnd, charset, false);
-        List<String> values = params.get(name);
-        if (values == null) {
-            values = new ArrayList<>(1);  // Often there's only 1 value.
-            params.put(name, values);
-        }
+        List<String> values = params.computeIfAbsent(name, new Function<String, List<String>>() {
+            @Override
+            public List<String> apply(final String s) {
+                return new ArrayList<>(1);
+            }
+        });
+        // Often there's only 1 value.
         values.add(value);
         return true;
     }
--- a/codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
@@ -30,7 +30,6 @@ import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.function.Function;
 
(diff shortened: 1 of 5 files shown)


kezhenxu94:
The second parameter of the `computeIfAbsent` could have been a lambda but because of what #8751 said, it will be replaced with lambda after that

normanmaurer:
@netty-bot test this please

kezhenxu94:
> I think this will have a performance impact. @kezhenxu94 could you please make small benchmark?
> I believe `Function` allocation is not free and not well optimized by JVM.

The annonymous inner classes will be replaced with lambda expressions (after solving #8751 ), which is optimized in Java 8+

kezhenxu94:
Are there any documentation on how to create benchmark?

doom369:
@kezhenxu94 https://openjdk.java.net/projects/code-tools/jmh/

I was curious, so I made a test:

```
@Fork(1)
@State(Scope.Thread)
@Warmup(iterations = 5, time = 1, batchSize = 1000)
@Measurement(iterations = 10, time = 1, batchSize = 1000)
public class ComputeIfAbsent {

    Map<String, List<Integer>> map;
    String param;

    @Setup
    public void init() {
        map = new HashMap<>();
        param = "123";
    }

    @Benchmark
    @CompilerControl(CompilerControl.Mode.DONT_INLINE)
    public List<Integer> regularNoInline() {
        List<Integer> list = map.get(param);
        if (list == null) {
            list = new ArrayList<>(1);
            map.put(param, list);
        }
        return list;
    }

    @Benchmark
    @CompilerControl(CompilerControl.Mode.DONT_INLINE)
    public List<Integer> computeIfAbsentNoInline() {
        return map.computeIfAbsent(param, k -> new ArrayList<>(1));
    }

    @Benchmark
    public List<Integer> regular() {
        List<Integer> list = map.get(param);
        if (list == null) {
            list = new ArrayList<>(1);
            map.put(param, list);
        }
        return list;
    }

    @Benchmark
    public List<Integer> computeIfAbsent() {
        return map.computeIfAbsent(param, k -> new ArrayList<>(1));
    }
}
```

Result:

```
Benchmark                                 Mode  Cnt       Score      Error  Units
ComputeIfAbsent.computeIfAbsent          thrpt   10  150639.561 ± 2102.531  ops/s
ComputeIfAbsent.computeIfAbsentNoInline  thrpt   10  129804.936 ± 5166.769  ops/s
ComputeIfAbsent.regular                  thrpt   10  154046.253 ± 3720.224  ops/s
ComputeIfAbsent.regularNoInline          thrpt   10  139151.529 ±  819.271  ops/s
```

`VM version: JDK 1.8.0_171, Java HotSpot(TM) 64-Bit Server VM, 25.171-b11`

Seems like in the latest JVM this code is optimized already.

kezhenxu94:
@doom369 Thanks a lot

kezhenxu94:
@normanmaurer sure, will do it later

kezhenxu94:
> @kezhenxu94 please update this one to use lambdas as well

@normanmaurer Done

johnou:
@normanmaurer see interesting discussion here https://github.com/orbit/orbit/pull/253

JDK-8184907 : ConcurrentHashMap.computeIfAbsent acquires lock even when key present http://bugs.java.com/bugdatabase/view_bug.do?bug_id=8184907 (fixed in JDK9)

normanmaurer:
@johnou interesting... we may not want to pull this in then...

kezhenxu94:
thanks

Issue #8512, opened 2018-11-13
[Netty 5] ChannelHandler execution


normanmaurer:
ChannelPipeline `add*(...)` / `replace(...)` support using a different `EventExecutor` for a `ChannelHandler`. This complicates the propagation of events through the `ChannelPipeline` and adds some additional overhead. We would like to propose removing this from the core and instead introduce a `ChannelHandler` which can offload to another EventLoop if necessary. 
We want to remove the “feature” of specifying a different `EventExecutorGroup` here and just always execute on the EventLoop. This will simplify things a lot, and is not known to be widely used. To offload handler execution to a different Thread we should implement a `ChannelHandler` that will take care of this (by wrapping others). This is more or less the same as it was done in Netty 3 with the [ExecutionHandler](https://github.com/netty/netty/blob/3.10/src/main/java/org/jboss/netty/handler/execution/ExecutionHandler.java).

PR #8778, opened 2019-01-24
Remove ability to specify a custom EventExecutor when adding handlers…

normanmaurer:
… to the ChannelPipeline

Motiviation:

In the past we allowed to use different EventExecutors for different ChannelHandlers in the ChannelPipeline. This introduced a lot of complexity while not providing much gain. Also it made the pipeline racy in terms of adding / remove handlers in some situations. This feature is not really used in the wild and can be easily archived by offloading heavy logic to an Executor by the user itself.

Modifications:

- Remove the ability to provide custom EventExecutor when adding handlers to the pipeline.
- Remove testcode that is not needed any more
- Ensure a handler is correctly visible in the pipeline when asked for it by the user while not be used until the EventLoop runs. This ensures correct ordering and visibility.
- Correctly remove ChannelHandlers from pipeline when scheduling of handlerAdded(...) callbacks fail.

Result:

Remove races in DefaultChannelPipeline and simplify implementation of AbstractChannelHandlerContext.


transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -60,17 +64,24 @@
                     DefaultChannelPipeline.class, MessageSizeEstimator.Handle.class, "estimatorHandle");
     final AbstractChannelHandlerContext head;
     final AbstractChannelHandlerContext tail;
+    final OrderedEventExecutor executor;
 
     private final Channel channel;
     private final ChannelFuture succeededFuture;
     private final VoidChannelPromise voidPromise;
     private final boolean touch = ResourceLeakDetector.isEnabled();
+    private final List<AbstractChannelHandlerContext> handlers = new ArrayList<>(4);
```
normanmaurer: I decided to just use an `ArrayList` here and synchronize on it. The reason for it is that it greatly simplifies things. I don't expect a ChannelPipeline to be too large so scanning the while List at worst shouldn't be a big deal. We did the same before when walking the while linked-list.

normanmaurer:
If we decide this is the way to go... (and I think it is) I will do a followup PR which will allow to write a `ChannelInboundHandler` which automatically offload all its events to an `EventExecutor` to have kind of a replacement of the old feature. That said this will only handle inbound and not outbound, which I think is good enough as usually people only want to offload inbound processing.

normanmaurer:
Related to https://github.com/netty/netty/issues/8512.

normanmaurer:
@stef1927 @tjake I know you use a custom EventExecutor in Cassandra for a handler but after looking at the code this could easily just be a ChannelHandler that does hand-over to an `Executor`. So I think this change is not a problem for you

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -664,201 +681,192 @@ public final ChannelHandlerContext context(Class<? extends ChannelHandler> handl
...
+    @Override
+    public ChannelHandler first() {
+        ChannelHandlerContext ctx = firstContext();
+        return ctx == null ? null : ctx.handler();
+    }
+
+    @Override
+    public ChannelHandlerContext firstContext() {
+        synchronized (handlers) {
+            if (handlers.isEmpty()) {
```
njhill: simplify with conditional operator?
```java
return handlers.isEmpty() ? null : handlers.get(0);
```

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -664,201 +681,192 @@ public final ChannelHandlerContext context(Class<? extends ChannelHandler> handl
...
+        return ctx == null ? null : ctx.handler();
+    }
 
-            ctx = ctx.next;
-            if (ctx == tail) {
-                break;
+    @Override
+    public ChannelHandlerContext lastContext() {
+        synchronized (handlers) {
+            if (handlers.isEmpty()) {
```
njhill: simplify with conditional operator?

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -633,8 +651,13 @@ public final ChannelHandlerContext context(String name) {
...
             throw new NullPointerException("name");
         }
-
-        return context0(name);
+        synchronized (handlers) {
+            int idx = findCtxIdx((ctx) -> ctx.name().equals(name));
+            if (idx == -1) {
+                return null;
+            }
+            return handlers.get(idx);
```
njhill: simplify to `return idx == -1 ? null  : handlers.get(idx)`? (same for other similar places)

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -343,168 +356,198 @@ private String generateName(ChannelHandler handler) {
...
-        }
-        return (T) remove((AbstractChannelHandlerContext) ctx).handler();
-    }
-
-    private AbstractChannelHandlerContext remove(final AbstractChannelHandlerContext ctx) {
-        assert ctx != head && ctx != tail;
-
-        synchronized (this) {
-            remove0(ctx);
+    private <T extends ChannelHandler> T removeIfExists(Supplier<Integer> idxSupplier) {
```
njhill: use `IntSupplier`?

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -181,99 +200,103 @@ private void addLast0(AbstractChannelHandlerContext newCtx) {
...
+            ctx = handlers.get(i);
+            handlers.add(i, newCtx);
+        }
+
+        EventExecutor executor = executor();
+        if (executor.inEventLoop()) {
+            addBefore0(ctx, newCtx);
+        } else {
+            try {
+                executor.execute(() -> addBefore0(ctx, newCtx));
```
njhill: same comment as above

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -181,99 +200,103 @@ private void addLast0(AbstractChannelHandlerContext newCtx) {
...
-            newCtx = newContext(executor, name, handler);
+        checkMultiplicity(handler);
+        if (name == null) {
+            name = generateName(handler);
+        }
 
-            addBefore0(ctx, newCtx);
+        final AbstractChannelHandlerContext newCtx = newContext(name, handler);
+        synchronized (handlers) {
+            int i = findCtxIdx((context) -> context.name().equals(baseName));
```
njhill: single-arg lambda doesn't need parentheses, i.e. can be `just context -> ...` (same in other places)

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -343,168 +356,198 @@ private String generateName(ChannelHandler handler) {
...
+        assert ctx != null;
+        EventExecutor executor = executor();
+        if (executor.inEventLoop()) {
+            remove0(ctx);
+        } else {
+            executor.execute(() -> remove0(ctx));
+        }
+    }
+
+    private int findCtxIdx(Predicate<AbstractChannelHandlerContext> predicate) {
```
njhill: maybe have an equivalent `private AbstractChannelHandlerContext findCtx(...)` method? Quite a few of the usages only use the index to re-lookup the context

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -181,99 +200,103 @@ private void addLast0(AbstractChannelHandlerContext newCtx) {
...
+            ctx = handlers.get(i);
+            handlers.add(i + 1, newCtx);
+        }
+
+        EventExecutor executor = executor();
+        if (executor.inEventLoop()) {
+            addAfter0(ctx, newCtx);
+        } else {
+            try {
+                executor.execute(() -> addAfter0(ctx, newCtx));
```
njhill: same comment as above

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -343,168 +356,198 @@ private String generateName(ChannelHandler handler) {
...
-                        callHandlerRemoved0(ctx);
-                    }
-                });
-                return ctx.handler();
+        EventExecutor executor = executor();
+        if (executor.inEventLoop()) {
+            replace0(oldCtx, newCtx);
+        } else {
+            try {
+                executor.execute(() -> replace0(oldCtx, newCtx));
```
njhill: similar race condition here I think

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -664,201 +681,192 @@ public final ChannelHandlerContext context(Class<? extends ChannelHandler> handl
...
-            } else {
-                final AbstractChannelHandlerContext finalCtx = ctx;
-                executor.execute(new Runnable() {
-                    @Override
-                    public void run() {
-                        destroyDown(Thread.currentThread(), finalCtx, true);
-                    }
-                });
-                break;
+    private void destroy() {
```
njhill: is this now guaranteed to be called from the event loop? I think so but just to double-check...

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -105,34 +116,40 @@ public final Channel channel() {
...
+        return executor;
     }
 
     @Override
-    public final ChannelPipeline addFirst(EventExecutor executor, String name, ChannelHandler handler) {
+    public final ChannelPipeline addFirst(String name, ChannelHandler handler) {
         final AbstractChannelHandlerContext newCtx;
-        synchronized (this) {
-            checkMultiplicity(handler);
-            name = filterName(name, handler);
```
stef1927: Why do we no longer need to check for duplicate names?

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -105,34 +116,40 @@ public final Channel channel() {
...
-                        callHandlerAdded0(newCtx);
-                    }
-                });
-                return this;
+        EventExecutor executor = executor();
+        if (executor.inEventLoop()) {
+            addFirst0(newCtx);
+        } else {
+            try {
+                executor.execute(() -> addFirst0(newCtx));
```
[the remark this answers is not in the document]
stef1927: I agree. If the linked list and array list need to be consistent, then we need to perform both operations in the same synchronized block every time a handler is added or removed.

transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
```diff
@@ -107,30 +103,18 @@
 
     private final DefaultChannelPipeline pipeline;
     private final String name;
-    private final boolean ordered;
 
-    // Will be set to null if no child executor should be used, otherwise it will be set to the
-    // child executor.
     final EventExecutor executor;
```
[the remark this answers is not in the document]
normanmaurer: @njhill for the head context we want to use `EventLoop`. That said we can just override the method.. let me do it.

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -1186,7 +1150,7 @@ public void channelReadComplete(ChannelHandlerContext ctx) throws Exception {
         private final Unsafe unsafe;
 
         HeadContext(DefaultChannelPipeline pipeline) {
-            super(pipeline, null, HEAD_NAME, HeadContext.class);
+            super(pipeline, pipeline.channel.eventLoop(), HEAD_NAME, HeadContext.class);
```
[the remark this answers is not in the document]
normanmaurer: no the `HeadContext` needs to be handled by the `EventLoop` as all operations in `unsafe` must be called from within the `EventLoop`.

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
normanmaurer: We do... it was just moved a bit

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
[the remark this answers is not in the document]
njhill: Consider concurrent `addFirst`s of handlers A and B, the following order is currently possible:
1. A is added to the array list
2. B added to array list
3. `addFirst0(B)` added to executor queue
4. `addFirst0(A)` added to executor queue

so B precedes A in the array list but follows A in the linked list

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
normanmaurer: @njhill also thanks for the question as I had a small bug there 😢 ... Fixing

normanmaurer:
Also maybe @nitsanw

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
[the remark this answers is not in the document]
normanmaurer: got it... yeah makes sense. Let me fix it.

normanmaurer:
@njhill `@GuardedBy` sounds like a good idea but I think that deserves a separated PR

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
normanmaurer: done

normanmaurer:
@njhill @stef1927 PTAL again... addressed all comments

normanmaurer:
@stef1927 let me just write a patch for Cassandra and submit it ;) Honestly I think this change even makes sense for how 4.x is used at the moment and may remove some overhead

normanmaurer:
@netty-bot test this please

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -122,34 +122,31 @@ public final EventExecutor executor() {
...
-        final AbstractChannelHandlerContext newCtx;
-
         checkMultiplicity(handler);
         if (name == null) {
             name = generateName(handler);
         }
 
+        AbstractChannelHandlerContext newCtx = newContext(name, handler);
+        EventExecutor executor = executor();
+        boolean inEventLoop = executor().inEventLoop();
```
njhill: `executor()` -> `executor`

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -169,28 +166,26 @@ public final ChannelPipeline addLast(String name, ChannelHandler handler) {
             name = generateName(handler);
         }
 
-        final AbstractChannelHandlerContext newCtx = newContext(name, handler);
-
+        AbstractChannelHandlerContext newCtx = newContext(name, handler);
+        EventExecutor executor = executor();
+        boolean inEventLoop = executor().inEventLoop();
```
njhill: `executor()` -> `executor`

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -212,9 +207,11 @@ public final ChannelPipeline addBefore(String baseName, String name, ChannelHand
             name = generateName(handler);
         }
 
-        final AbstractChannelHandlerContext newCtx = newContext(name, handler);
+        AbstractChannelHandlerContext newCtx = newContext(name, handler);
+        EventExecutor executor = executor();
+        boolean inEventLoop = executor().inEventLoop();
```
njhill: `executor()` -> `executor`

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -260,9 +254,11 @@ public final ChannelPipeline addAfter(String baseName, String name, ChannelHandl
...
```
njhill: `executor()` -> `executor`

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -400,74 +383,125 @@ private int findCtxIdx(Predicate<AbstractChannelHandlerContext> predicate) {
...
         synchronized (handlers) {
-            int idx = findCtxIdx((context) -> context.handler() == handler);
+            int idx = findCtxIdx(context -> context.handler() == handler);
             if (idx == -1) {
                 throw new NoSuchElementException();
             }
             ctx = handlers.remove(idx);
+            assert ctx != null;
+
+            if (!inEventLoop) {
```
njhill: Technically I don't think this issue applies to the remove cases - the same kind of interleaving won't matter since the end result will be the same. So the execute could remain outside the sync block, but maybe it's better like this anyhow for consistency

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -628,22 +662,19 @@ private void callHandlerRemoved0(final AbstractChannelHandlerContext ctx) {
...
         ChannelHandlerContext ctx = context(handlerType);
-        if (ctx == null) {
-            return null;
-        } else {
-            return (T) ctx.handler();
-        }
+        return ctx == null ? null : (T) ctx.handler();
+    }
+
+    private AbstractChannelHandlerContext findCtx(Predicate<AbstractChannelHandlerContext> predicate) {
```
njhill: I had in mind more like:
```java
    private AbstractChannelHandlerContext findCtx(Predicate<AbstractChannelHandlerContext> predicate) {
        for (int i = 0; i < handlers.size(); i++) {
            AbstractChannelHandlerContext ctx = handlers.get(i);
            if (predicate.test(ctx)) {
                return ctx;
            }
        }
        return null;
    }
```

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -323,160 +349,239 @@ private String generateName(ChannelHandler handler) {
...
-                if (!sameName) {
-                    checkDuplicateName(newName);
+            if (!oldCtx.name().equals(newName)) {
+                if (context(newName) != null) {
+                    throw new IllegalArgumentException("Duplicate handler name: " + newName);
                 }
             }
+            AbstractChannelHandlerContext removed = handlers.remove(idx);
+            assert removed != null;
+            handlers.add(idx, newCtx);
```
njhill: how about just `removed = handlers.set(idx, newCtx)` here?

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -323,160 +349,239 @@ private String generateName(ChannelHandler handler) {
...
+        EventExecutor executor = executor();
+        if (executor.inEventLoop()) {
+            replace0(oldCtx, newCtx);
+        } else {
+            try {
+                executor.execute(() -> replace0(oldCtx, newCtx));
+            } catch (Throwable cause) {
+                synchronized (handlers) {
+                    handlers.remove(idx);
+                    handlers.add(idx, oldCtx);
```
njhill: just `handlers.set(idx, oldCtx)`?

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -323,160 +349,239 @@ private String generateName(ChannelHandler handler) {
...
-                    // those event handlers must be called after handlerAdded().
-                    callHandlerAdded0(newCtx);
-                    callHandlerRemoved0(ctx);
-                });
-                return ctx.handler();
+        EventExecutor executor = executor();
+        if (executor.inEventLoop()) {
+            replace0(oldCtx, newCtx);
+        } else {
+            try {
```
njhill: I think this still needs to be moved inside the sync block? The `final int idx;` declaration can then also be too

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
```diff
@@ -1148,11 +1178,16 @@ public void channelReadComplete(ChannelHandlerContext ctx) throws Exception {
...
         HeadContext(DefaultChannelPipeline pipeline) {
-            super(pipeline, null, HEAD_NAME, HeadContext.class);
+            super(pipeline, HEAD_NAME, HeadContext.class);
             unsafe = pipeline.channel().unsafe();
             setAddComplete();
         }
 
+        @Override
+        public EventExecutor executor() {
+            return channel.eventLoop();
```
njhill: @normanmaurer follow on from prior comment here ... I wonder if it's a problem though that the executor is different for the head context, given we now rely on the fact that all the changes to next/prev pointers in `AbstractChannelHandlerContext` happen in the same (ordered/serialized) executor?

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
njhill: Actually yes I guess better to keep like this so the rollback case is handled properly

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
normanmaurer: hmm yeah.... Let me just only support the Eventloop for the whole pipeline. If we really need the feature later on we can do a followup.

normanmaurer:
@njhill addressed

transport/src/main/java/io/netty/channel/DefaultChannelPipeline.java
stef1927: I saw it, thanks.

sbtourist:
@normanmaurer I agree with the general intent of this patch, as per-handler executors have definitely been a source of bugs and confusion, but regarding:

> [the same functionality] can be easily archived by offloading heavy logic to an Executor by the user itself.

That is true, but not enough to efficiently implement all use cases, in particular when the same executor is configured for different handlers. As an example, let's say we have a pipeline configured with `(executor1, handler1)` and `(executor1, handler2)`, on Netty4 a task travelling from the first to the second handler would not jump event loops because the `inEventLoop()` would return true, while on Netty5 `handler1` would have to manually submit the task on `executor1`, which then upon invoking any context method to "propagate" the task to `handler2` would go back on the pipeline event loop and manually enqueued again by `handler2` on `executor1`. Does this make sense? Or did I misunderstand anything?

Provided it makes sense, I _think_ a way to overcome that could be to provide an handler method or annotation to signal for "inline" execution (in lack of better terms). Thoughts?

normanmaurer:
@sbtourist you are 100 % right with all you said. I am not yet convinced we need to provide support for this tho. Bu there are definitely ways how we can without complicate the rest. IMHO we should handle this as a followup . WDYT?

sbtourist:
> IMHO we should handle this as a followup . WDYT?

Thanks @normanmaurer, I'm absolutely fine with that, just I'd really like to see that addressed as we're working on some new features using handler-specific executors and we need that kind of feature to eventually switch to Netty5.

normanmaurer:
@sbtourist btw can you also describe a bit more what you want to do ... do you want to build something like a SEDA like architecture ?

normanmaurer:
@netty-bot test this please

sbtourist:
> btw can you also describe a bit more what you want to do ... do you want to build something like a SEDA like architecture ?

Nope, it's a kind of load balancing. We need to (sometimes, on a given handler) submit a task to a different executor, and in such cases we want all subsequent handlers to stick on the same thread.

normanmaurer:
@sbtourist got it... my current idea at the moment would be to allow "child" pipelines and allow to have one executor per pipeline. That should be easy to do and would give you the guarantees you are talking about.

normanmaurer:
@sbtourist btw also a review on this PR is welcome :)

sbtourist:
> my current idea at the moment would be to allow "child" pipelines and allow to have one executor per pipeline.

I guess it depends. It might be easy to implement, but sounds semantically complex, as it's one more abstraction to grok: if you know/think it would be used frequently enough, then it's justified; otherwise, a method/annotation to signal for inlined execution would add almost no cognitive overload (as the default implementation/usage would just "hide" it), and still be available for advanced users.

> btw also a review on this PR is welcome

I had a cursory look, also my colleague @stef1927 already had a detailed review, so I think you're good to go :)

normanmaurer:
@netty-bot test this please

transport/src/main/java/io/netty/channel/ChannelPipeline.java
```diff
@@ -192,20 +189,16 @@
...
  *
- * // Tell the pipeline to run MyBusinessLogicHandler's event handler methods
- * // in a different thread than an I/O thread so that the I/O thread is not blocked by
- * // a time-consuming task.
- * // If your business logic is fully asynchronous or finished very quickly, you don't
- * // need to specify a group.
- * pipeline.addLast(group.next(), "handler", new MyBusinessLogicHandler());
+ * // If your business logic does block or take a lot of time you should offload the work to an extra
+ * // {@link java.util.concurrent.Executor} to ensure you don't block the {@link EventLoop}.
+ * pipeline.addLast(group, "handler",  new MyBusinessLogicHandler());
```
ejona86: Drive-by comment: this looks out-of-date. This method doesn't seem to exist any more.

normanmaurer:
Let me merge this to make progress... We can do a followup if needed.

normanmaurer:
@sbtourist I created https://github.com/netty/netty/issues/8810 as a followup

sbtourist:
Thank you @normanmaurer, I will have a look.

PR #8779, opened 2019-01-25
Migrate java8: remove omitted param

thinkerou:
Motivation:

calls to parametrized methods where explicit argument types can be omitted, so remove it.


normanmaurer:
@thinkerou thanks a lot

Issue #8641, opened 2018-12-07
CompositeByteBuf.Component freeIfNecessary always frees


jameskleeh:
### Expected behavior
The `freeIfNecessary` method should only release if the ref count > 0

### Actual behavior
The `freeIfNecessary` method always releases

I would like to be able to free any given component of a composite byte buffer myself, with release being called on the whole buffer only releasing the components that were not previously released. The method name indicates to me that is what should happen

### Netty version
4.1.30

njhill:
@jameskleeh the `Component` class is private/internal to `CompositeByteBuf` and I think this method behaves as intended. I acknowledge the name is misleading and I think we should change it to just `free()`, but this is only to make the internal logic clearer, not because it's an API method. I suggest `free()` rather than `release()` to avoid any confusion with the public `release()` methods.

`CompositeByteBuf` takes ownership of any buffers added to it. If you subsequently only access them via the composite buffer, you don't need to worry about releasing them individually. You can remove components via the `removeComponent(...)` methods which will automatically release the corresponding buffers, and when `release` is ultimately called on the composite buffer it will release only the contained buffers it still owns, not any which it previously released due to explicit or indirect removal.

jameskleeh:
@njhill Thanks for the explanation! I think it would be good to enforce the ownership of those components in code by returning a bytebuf that prevents the user from releasing it directly

PR #8784, opened 2019-01-25
CompositeByteBuf tidy-up

njhill:
Motivation

There's some further miscellaneous cleanup/simplification of `CompositeByteBuf` which would help make the code a bit clearer.

Modifications

- Simplify web of constructors and `addComponents` methods, reducing duplication of logic
- Rename `Component.freeIfNecessary()` method to just `free()`, which is less confusing (see #8641)
- Make loop in `addComponents0(...)` method more verbose/readable (see https://github.com/netty/netty/pull/8437#discussion_r232124414)
- Simplify addition/subtraction in `setBytes(...)` methods

Result

Smaller/clearer code

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
@@ -428,7 +428,11 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
             checkComponentIndex(cIndex);
 
             // No need for consolidation
-            for (ByteBuf b; it.hasNext() && (b = it.next()) != null;) {
+            while (it.hasNext()) {
+                ByteBuf b = it.next();
+                if (b == null) {
+                    break;
+                }
                 cIndex = addComponent0(increaseIndex, cIndex, b) + 1;
                 cIndex = Math.min(cIndex, componentCount);
             }
--- a/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
@@ -87,6 +87,7 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
         this(alloc, direct, maxNumComponents, buffers.length - offset);
 
         addComponents0(false, 0, buffers, offset);
+        consolidateIfNeeded();
         setIndex0(0, capacity());
     }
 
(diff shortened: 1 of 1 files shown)


normanmaurer:
@netty-bot test this please

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
```diff
@@ -413,39 +408,34 @@ private int addComponents0(boolean increaseWriterIndex, final int cIndex, ByteBu
...
             checkComponentIndex(cIndex);
 
             // No need for consolidation
-            while (it.hasNext()) {
-                ByteBuf b = it.next();
-                if (b == null) {
-                    break;
-                }
+            for (ByteBuf b; it.hasNext() && (b = it.next()) != null;) {
                 cIndex = addComponent0(increaseIndex, cIndex, b) + 1;
```
ejona86: Seems this should be `addComponent()` (not 0).

ejona86:
I gave up trying to figure out if various addComponents() vs addComponents0() were the correct version.

njhill:
Thanks for trying to navigate this @ejona86! I was focused primarily on the de-duplication and you're right that I should have paid more attention to the consistency of the behaviour of the 0-suffix versions. In my defence I think most of the inconsistency was preexisting!

I've added another small commit to make this consistent. Now none of the `addComponent0` and `addComponents0` methods perform both null checks or consolidation, and all of the non-suffixed `addComponent` and `addComponents` ones _do_.

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
njhill: @ejona86 this wasn't changed in this PR, but I think it was intended to be the 0 version as written. The buffer `b` is certain to be non-null, and we're avoiding consolidation after each addition so that it can be done once at the end of the enclosing method.

ejona86: Oh, I missed the `(b = it.next()) != null)`. I thought it was the normal `b = it.next()`. It doesn't seem the loop is any better with this change and I might argue it was more clear before. (It's not clear why it would choose to stop early if `b == null`, but it was obvious it was explicitly handling that case.)

normanmaurer:
@netty-bot test this please

normanmaurer:
@ejona86 want to take another look now that @njhill did a final commit or happy as it is ?

ejona86:
@normanmaurer, I was happy with it already. Reverting the loop change was just icing on the cake.

PR #8785, opened 2019-01-25
Remove deprecated SslContext constructors

slandelle:
Motivation:

SslContext implementations have tons of contructors, most of them deprecated as we want to enforce builder usage in Netty 5.

Cleaning them up is a requirement prior to introducing new parameters such as hostname verification.

Modifications:

* Make SslContext implementations classes and constructors package private, users are supposed to use the SslContextBuilder.
* Drop all but one constructor. The exception for now is with Jdk(Client|Server)Context that still has an additional constructor that takes an ApplicationProtocolNegotiator parameter. ApplicationProtocolNegotiator usage is supposed to be dropped in favor of ApplicationProtocolConfig and this constructor is only used in tests, so I guess it will be dropped to in a follow up.

Result:

Deprecated code dropped. Path cleaned up for introducing new features with having to introduce yet another constructor.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/JdkSslClientContext.java
+++ b/handler/src/main/java/io/netty/handler/ssl/JdkSslClientContext.java
@@ -16,260 +16,55 @@
 
 package io.netty.handler.ssl;
 
+import java.io.File;
+import java.security.PrivateKey;
 import java.security.Provider;
-import javax.net.ssl.KeyManager;
+import java.security.cert.X509Certificate;
+
 import javax.net.ssl.KeyManagerFactory;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLException;
 import javax.net.ssl.SSLSessionContext;
-import javax.net.ssl.TrustManager;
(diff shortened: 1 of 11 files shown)


slandelle:
@netty-bot test this please

normanmaurer:
@netty-bot test this please

slandelle:
@normanmaurer What should we do about the constructors that take an `ApplicationProtocolNegotiator` instead of a `ApplicationProtocolConfig`?
The former is supposed to be made an internal while the public API will be the latter.
They solely exist for the sake of unit tests. Should be still keep them around or should we modify the tests?

normanmaurer:
If possible I would prefer to adjust the unit tests

slandelle:
Will try to.

First, can I send a PR for dropping NPN support?

normanmaurer:
👍

slandelle:
Answer 5 minutes after landing, impressive :P

slandelle:
I realize NON support can't be dropped without dropping SPDY support first. Are you fine with it too?

PR #8786, opened 2019-01-26
Minimize memory footprint for AbstractChannelHandlerContext for handl…

normanmaurer:
…ers that execute in the EventExecutor.

Motivation:

We cache the Runnable for some tasks to reduce GC pressure in 4 different fields. This gives overhead in terms of memory usage in all cases, even if we always execute in the EventExecutor (which is the case most of the times).

Modifications:

Move the 4 fields to another class and only have one reference to this in AbstractChannelHandlerContext. This gives a small overhead in the case of execution that is done outside of the EventExecutor but reduce memory footprint in the more likily execution case.

Result:

Less memory used per AbstractChannelHandlerContext in most cases.

### DIFF
--- a/transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
+++ b/transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
@@ -76,10 +76,7 @@ abstract class AbstractChannelHandlerContext extends DefaultAttributeMap
 
     // Lazily instantiated tasks used to trigger events to a handler with different executor.
     // There is no need to make this volatile as at worse it will just create a few more instances then needed.
-    private Runnable invokeChannelReadCompleteTask;
-    private Runnable invokeReadTask;
-    private Runnable invokeChannelWritableStateChangedTask;
-    private Runnable invokeFlushTask;
+    private Tasks invokeTasks;
 
     private volatile int handlerState = INIT;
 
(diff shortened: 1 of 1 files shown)


normanmaurer:
@njhill @franz1981

normanmaurer:
@netty-bot Test this please

transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
```diff
@@ -1162,4 +1139,36 @@ public void write(AbstractChannelHandlerContext ctx, Object msg, ChannelPromise
             ctx.invokeFlush();
         }
     }
+
+    private static final class Tasks {
+        private final AbstractChannelHandlerContext next;
+        private final Runnable invokeChannelReadCompleteTask = new Runnable() {
```
franz1981: I was expecting this one to be initailized to `null`!

franz1981: Now I've understood the logic :) it's fine!

transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
```diff
@@ -1162,4 +1139,36 @@ public void write(AbstractChannelHandlerContext ctx, Object msg, ChannelPromise
...
+            @Override
+            public void run() {
+                next.invokeChannelReadComplete();
+            }
+        };
+        private Runnable invokeReadTask = new Runnable() {
```
normanmaurer: Doh... all of these should be final. Will fix later today

normanmaurer:
@netty-bot test this please

PR #8794, opened 2019-01-29
Drop NPN and SPDY support

slandelle:
Motivation:

* NPN has been superseeded by ALPN
* SPDY has been superseeded by HTTP/2 and Chrome has dropped support in 2016

Those protocols are deprecated and most likely no longer used at large.

Modifications:

Remove NPN and SPDY code.

`ApplicationProtocolConfig.Protocol` enum is left with 2 values: NONE and ALPN, as this might be more explicit than a boolean and would be more versatile should a new negotiation protocol appear.

Result:

Removed dead code


slandelle:
Note: NPN code will also have to be dropped from tcnative.

normanmaurer:
I am super happy to drop SPDY but would like to give others a chance to raise any concerns.

normanmaurer:
and also @smaldini

slandelle:
@vietj @smaldini Hey Frenchies! Any comment?

normanmaurer:
@ejona86 any idea about NPN and GRPC yet ?

carl-mastrangelo:
@slandelle I can't comment on SPDY at the moment, but I see a fairly large amount of NPN negotiations still happening here.  I can't give exact numbers, but it's a minority, and frequently with HTTP/1.1.  I don't see any gRPC traffic with its use, though.

ejona86:
Killing SPDY is easy. Killing NPN is a bit harder as there may be lingering clients still on old versions of OpenSSL (because most of the world is on those older versions). I talked with some others and believe that all our clients are shipping BoringSSL (or similar), so they should all support ALPN. But it's also one of those things that you only learn about users of old features when things go wrong.

Since this is slated for Netty 5 and can't really be backported to Netty 4, this is probably fine. If it was for Netty 4 (API stability aside), I'd have more pause.

slandelle:
@carl-mastrangelo People with very old Chrome with SPDY? How large is it? I guess Google has browsers versions usage stats, right?

carl-mastrangelo:
@slandelle I can't comment on the origin, the metrics I queried weren't so fine grained.

ejona86:
Firefox and IE implemented SPDY as well. GFE doesn't seem to negotiate SPDY any more (via ssllabs). Over a year ago, [Cloudflare said they saw < 4%](https://blog.cloudflare.com/introducing-http2/).

thinkerou:
`io.netty.handler.codec.spdy` package also need to remove?

rkapsi:
No dependence on SPDY or NPN on our end.

normanmaurer:
@slandelle @carl-mastrangelo @ejona86 maybe we should play safe here and only remove SPDY?

slandelle:
@normanmaurer That's for you big internet corps people to decide: how much of NPN traffic will your companies be seeing one year from now when Netty 5 will be out (or will go on production, which might be even later)?

Also, what would the consequence be exactly? Clients using NPN for SPDY or HTTP/2 being downgraded to HTTP/1.1? Do you know who those clients are and why they would not update OpenSSL (all the more as OpenSSL vulnerabilities get disclosed several times a year)? Would those clients upgrade if they face a performance regression?

Personally, I'm not fond of keeping on supporting some legacy protocol for the next major version. Would the pieces it depends on be maintained (OpenSSL's NPN, Jetty's NPN)?
If it can't be dropped now, it won't be dropped until Netty 6.

carl-mastrangelo:
@normanmaurer From my reading, SPDY traffic is zero, so I would be comfortable on our side axing it. NPN is not so clear cut.  My reading of the graph is that NPN counts for a little less than 4% of negotiations/second.

@slandelle The NPN I see on my end is for HTTP/1.1.   There is a minuscule amount for h2.   Lastly, looking back over the past 90 days it doesn't appear the NPN traffic is going down.  It's either steady or very slightly trending up.   I can't speak on what would happen if NPN were cut.

ejona86:
> Do you know who those clients are and why they would not update OpenSSL (all the more as OpenSSL vulnerabilities get disclosed several times a year)?

OpenSSL is upgraded when users upgrade to the next major version of their distribution. So it's not something that happens overnight.

> maybe we should play safe here and only remove SPDY?

That would be fair. I'd feel comfortable removing NPN when BoringSSL removes NPN.

> Would the pieces it depends on be maintained (OpenSSL's NPN, Jetty's NPN)?

Jetty NPN is already dead. It was for Java 7-only, which means it doesn't matter for Netty 5.

Between OpenSSL and BoringSSL, BoringSSL would most likely remove it first and could even do it within the next year. OpenSSL has a history of keeping old stuff around forever (while still managing to break API...). Since it'll be a while until Netty 5 is released, we may want to revisit this question later.

I don't know about LibreSSL.

ejona86:
Note: for gRPC the problem is we won't downgrade to HTTP/1; we would just break.

normanmaurer:
@slandelle after thinking more about all of this and taking the comments of @ejona86 and @carl-mastrangelo into account I think I would like to only remove SPDY for now and then revisit NPN later once we are closer to a release (as @ejona86 suggested).

slandelle:
@normanmaurer OK. I'm going to open a new PR for removing SPDY and open a dedicated issue for considering removing NPN later. Fine with you?

normanmaurer:
@slandelle yes

chrisvest:
Circling back here after more than 5 years to record the current state for historical purpose.

We've removed NPN in the Netty 4.2 and 5 (main) branches. Netty 4.1 will be the end of the line for NPN.

> I'd feel comfortable removing NPN when BoringSSL removes NPN.

BoringSSL has put out feelers to see where people are on deprecation: https://github.com/netty/netty-tcnative/issues/856
But as I understand the conversation, they'd prefer it if people stop using it before they remove it. So that's what 4.2 and 5 does.

Issue #8736, opened 2019-01-19
HttpObjectDecoder ignores HTTP trailer header.


KowalczykBartek:
### Behavior
When I was writing test for https://github.com/netty/netty/pull/8721 in **io.netty.handler.codec.http.HttpContentDecoderTest** I encountered a problem with parsing trailer header using HttpObjectDecoder. 
HttpObjectDecoder is not able to handle following byte sequence 
```java
assertTrue(channel.writeInbound(Unpooled.copiedBuffer("My-Trailer: 42", CharsetUtil.US_ASCII)));
assertTrue(channel.writeInbound(Unpooled.copiedBuffer("\r\n", CharsetUtil.US_ASCII)));
assertTrue(channel.writeInbound(Unpooled.copiedBuffer("\r\n\r\n", CharsetUtil.US_ASCII)));
```
but is able to deal with that 
```java
assertTrue(channel.writeInbound(Unpooled.copiedBuffer("My-Trailer: 42\r\n\r\n\r\n", CharsetUtil.US_ASCII)));
```
After small debugging session, I noticed that HttpObjectDecoder is actually parsing trailer but that line of  code 
```java
line = headerParser.parse(buffer);
if (line == null) {
    return null;
}
```
cause to ignore my trailer, because **headerParser.parse(buffer)** on empty **buffer** will return null instead of parsed trailer.

### Minimal yet complete reproducer code (or URL to code)
Didn't want to make pull-request, but following test (class **io.netty.handler.codec.http.HttpContentDecoderTest**) will reproduce the problem
```java
@Test
public void testChunkedRequestDecompression() {
    HttpResponseDecoder decoder = new HttpResponseDecoder();
    HttpContentDecoder decompressor = new HttpContentDecompressor();

    EmbeddedChannel channel = new EmbeddedChannel(decoder, decompressor, null);

    String headers = "HTTP/1.1 200 OK\r\n"
            + "Transfer-Encoding: chunked\r\n"
            + "Trailer: My-Trailer\r\n"
            + "Content-Encoding: gzip\r\n\r\n";

    channel.writeInbound(Unpooled.copiedBuffer((headers.getBytes(CharsetUtil.US_ASCII))));

    assertTrue(channel.writeInbound(Unpooled.copiedBuffer(Integer.toHexString(GZ_HELLO_WORLD.length) + "\r\n", CharsetUtil.US_ASCII)));
[...]

PR #8799, opened 2019-01-29
HttpObjectDecoder ignores HTTP trailer header when empty line is rece…

normanmaurer:
…ived in seperate ByteBuf

Motivation:

When the empty line that termines the trailers was sent in a seperate ByteBuf we did ignore the previous parsed trailers and just returned none.

Modifications:

- Correct respect previous parsed trailers.
- Add unit test.

Result:

Fixes https://github.com/netty/netty/issues/8736

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectDecoder.java
@@ -641,15 +641,9 @@ public abstract class HttpObjectDecoder extends ByteToMessageDecoder {
             return null;
         }
         LastHttpContent trailer = this.trailer;
-        if (line.length() == 0) {
-            // We have received the empty line which signals the trailer is complete. In this case we need to check if
-            // we have parsed the trailer before and if so return it, otherwise we just return an empty last content.
-            //
-            // See https://github.com/netty/netty/issues/8736
-            if (trailer != null) {
-                this.trailer = null;
-                return trailer;
-            }
+        if (line.length() == 0 && trailer == null) {
+            // We have received the empty line which signals the trailer is complete and did not parse any trailers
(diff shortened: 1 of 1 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectDecoder.java
```diff
@@ -681,6 +680,15 @@ private LastHttpContent readTrailingHeaders(ByteBuf buffer) {
             this.trailer = null;
             return trailer;
         }
+        // We have received the empty line which signals the trailer is complete. In this case we need to check if
```
ejona86: This flow is a bit convoluted. How about we remove the surrounding if and make it clear why the condition is there:

```
if (line.length() == 0 && this.trailer == null) {
  return LastHttpContent.EMPTY_LAST_CONTENT; // optimization
}
// everything from the if today
LastHttpContent trailer = this.trailer;
if (trailer == null) {
...
return trailer;
```

Alternatively the if could be changed to `line.length() > 0 || this.trailer != null`, but it's non-obvious why it would be that way. (It's still more obvious than this code, in my mind, though.)

normanmaurer:
@ejona86 PTAL again... I think it's better now (slightly different than what you suggested tho).

ejona86:
My suggestion removed a special-case: it used the normal do-while to handle the "previously saw a trailer". Instead of having 2 special cases like it is now, there was only one: returning EMPTY_LAST_CONTENT. It is makes sense that we should only return EMPTY_LAST_CONTENT if `trailer` is `null`, so it's easy to reason about.

With this code as-is, you really have to read more of the do-while loop to verify that the if block is correct. You have to look for things like making sure `trailer` is set to `null`.

normanmaurer:
@ejona86 I am still not sure if I did what you suggested but please have a look again 😆

PR #8800, opened 2019-01-29
#7756 Add @Sharable TYPE_USE support for inner class annotations

### LINKED ISSUE #7756: @Sharable should also support @Target(ElementType.TYPE_USE)
### Expected behavior
`@Sharable` should also be usable with anonymous inner types, so that it is no longer necessary to make named classes (top-level or nested) for sharable handlers. Since Java 8, this is possible with the new `ElementType` `TYPE_USE`.
The practical problem here is that either (a) Netty would have to, at some point, *require* Java 8 (don't think that's the case yet), or (b) an additional Java-8-specific `@Sharable` annotation would have to be introduced.

### Actual behavior
`@Sharable` only targets `ElementType` `TYPE` and requires a proper (i.e., named) class.

### Steps to reproduce
Add `@Sharable` annotation to inner class, observe compilation error.

```
ChannelInboundHandler echoHandler = new @Sharable ChannelInboundHandlerAdapter() {
      @Override
      public void channelRead(ChannelHandlerContext context, Object message) {
           context.write(message);
      }
}
```
### Netty version
4.1.22.Final

1.8.0_152

Darwin Kernel Version 16.7.0: Wed Oct  4 00:17:00 PDT 2017; root:xnu-3789.71.6~1/RELEASE_X86_64 x86_64 (MacOS 10.12.6)

kiril-me:
Motivation:

Make @Sharable annotation works with anonymous inner types. Add Java 8 ElementType.TYPE_USE feature that makes easy to use @Sharable annotation.

Modification:
* transport/src/main/java/io/netty/channel/ChannelHandler.java - Target ElementType.TYPE_USE added.
* transport/src/main/java/io/netty/channel/ChannelHandlerAdapter.java - isSharable method improved to verify AnnotatedSuperclass for annotation.
* transport/src/test/java/io/netty/channel/ChannelHandlerAdapterTest.java - Tests added.

Result:
```
ChannelInboundHandler handler = new @Sharable ChannelInboundHandlerAdapter() {
      @Override
      public void channelRead(ChannelHandlerContext context, Object message) {
           context.write(message);
      }
};
```

Note:
The following changes don't support local variable annotation:
 ChannelInboundHandler handler1 = new @Sharable ChannelInboundHandlerAdapter();
 @Sharable ChannelInboundHandler handler2 = new  ChannelInboundHandlerAdapter();

Fixes #7756

### DIFF
--- a/transport/src/main/java/io/netty/channel/ChannelHandlerAdapter.java
+++ b/transport/src/main/java/io/netty/channel/ChannelHandlerAdapter.java
@@ -18,6 +18,7 @@ package io.netty.channel;
 
 import io.netty.util.internal.InternalThreadLocalMap;
 
+import java.lang.reflect.AnnotatedType;
 import java.util.Map;
 
 /**
@@ -55,6 +56,10 @@ public abstract class ChannelHandlerAdapter implements ChannelHandler {
         Boolean sharable = cache.get(clazz);
         if (sharable == null) {
             sharable = clazz.isAnnotationPresent(Sharable.class);
+            if (!sharable) {
+                AnnotatedType annotatedType = clazz.getAnnotatedSuperclass();
+                sharable = annotatedType.isAnnotationPresent(Sharable.class);
+            }
             cache.put(clazz, sharable);
         }
         return sharable;
--- a/transport/src/test/java/io/netty/channel/ChannelHandlerAdapterTest.java
+++ b/transport/src/test/java/io/netty/channel/ChannelHandlerAdapterTest.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2019 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.channel;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
+import io.netty.channel.ChannelHandler.Sharable;
+
+public class ChannelHandlerAdapterTest {
+
+    @Sharable
+    private static final class SharableChannelHandlerAdapter extends ChannelHandlerAdapter {
+    }
+
+    @Test
+    public void testSharable() {
+        ChannelHandlerAdapter handler = new SharableChannelHandlerAdapter();
+        assertEquals(true, handler.isSharable());
+    }
+
+    @Test
+    public void testInnerClassSharable() {
+        ChannelHandlerAdapter handler = new @Sharable ChannelHandlerAdapter() { };
+        assertEquals(true, handler.isSharable());
+    }
+
+    @Test
+    public void testWithoutSharable() {
+        ChannelHandlerAdapter handler = new ChannelHandlerAdapter() { };
+        assertEquals(false, handler.isSharable());
+    }
+}
--- a/transport/src/main/java/io/netty/channel/ChannelHandler.java
+++ b/transport/src/main/java/io/netty/channel/ChannelHandler.java
@@ -210,7 +210,7 @@ public interface ChannelHandler {
      */
     @Inherited
     @Documented
-    @Target(ElementType.TYPE)
+    @Target({ElementType.TYPE, ElementType.TYPE_USE})
     @Retention(RetentionPolicy.RUNTIME)
     @interface Sharable {
         // no value


transport/src/test/java/io/netty/channel/ChannelHandlerAdapterTest.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright 2012 The Netty Project
```
normanmaurer: 2019

transport/src/test/java/io/netty/channel/ChannelHandlerAdapterTest.java
```diff
@@ -0,0 +1,49 @@
...
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
+import io.netty.channel.ChannelHandler.Sharable;
+
+public class ChannelHandlerAdapterTest {
+
+    @Sharable
+    private static class SharableChannelHandlerAdapter extends ChannelHandlerAdapter {
```
normanmaurer: nit: final

transport/src/test/java/io/netty/channel/ChannelHandlerAdapterTest.java
```diff
@@ -0,0 +1,49 @@
...
+    @Test
+    public void testSharable() {
+        ChannelHandlerAdapter handler = new SharableChannelHandlerAdapter();
+        assertEquals(true, handler.isSharable());
+    }
+
+    @Test
+    public void testInnerClassSharable() {
+        ChannelHandlerAdapter handler = new @Sharable ChannelHandlerAdapter() {
+        };
```
normanmaurer: nit: `new @Sharable ChannelHandlerAdapter() { };`

transport/src/test/java/io/netty/channel/ChannelHandlerAdapterTest.java
```diff
@@ -0,0 +1,49 @@
...
+        assertEquals(true, handler.isSharable());
+    }
+
+    @Test
+    public void testWithoutSharable() {
+        ChannelHandlerAdapter handler = new ChannelHandlerAdapter() {
+        };
```
normanmaurer: nit: `new ChannelHandlerAdapter() {};`

normanmaurer:
@netty-bot test this please

transport/src/main/java/io/netty/channel/ChannelHandlerAdapter.java
```diff
@@ -55,6 +56,10 @@ public boolean isSharable() {
         Boolean sharable = cache.get(clazz);
         if (sharable == null) {
             sharable = clazz.isAnnotationPresent(Sharable.class);
+            if (!sharable) {
```
doom369: Is this correct? Child class could be nonsharable, while it extends `@Sharable`.

normanmaurer: @doom369 not sure I understand you correctly.. but `@Sharable` uses `@Inherited` so if the super-class is sharable the child class should be as well.

doom369: @normanmaurer missed that. Anyway, in that case:

```
 * Indicates that an annotation type is automatically inherited.  If
 * an Inherited meta-annotation is present on an annotation type
 * declaration, and the user queries the annotation type on a class
 * declaration, and the class declaration has no annotation for this type,
 * then the class's superclass will automatically be queried for the
 * annotation type.  This process will be repeated until an annotation for this
 * type is found, or the top of the class hierarchy (Object)
 * is reached.  If no superclass has an annotation for this type, then
 * the query will indicate that the class in question has no such annotation.
```

```
    @ChannelHandler.Sharable
    private static class A {
    }
    private static class B extends A {
    }
    @Test
    public void aaa() {
        System.out.println(B.class.isAnnotationPresent(ChannelHandler.Sharable.class)); //prints true
    }
```

normanmaurer: @doom369 agree... so you are saying we do not need the extra "query" ? If so I think you are right... @kiril-me did you check if you really need it ?

doom369: @normanmaurer this check is useful only for the case when the class implements interface marked with `@Sharable`. I don't think this is real use case, however, for at least 1 more additional test case should be added.

normanmaurer: @kiril-me can you do this ?

kiril-me: @normanmaurer There are differences between annotation of the class and inner anonymous class annotation. 

> @Inherited JavaDoc
> Note that this meta-annotation type has no effect if the annotated
> type is used to annotate anything other than a class.  Note also
> that this meta-annotation only causes annotations to be inherited
> from superclasses; annotations on implemented interfaces have no
> effect.

As @doom369 mentioned, isAnnotationPresent(Sharable.class) looking for @Sharable annotation starting from current class till Object.

But this scenario doesn't work for isAnnotationPresent, because @Inherit does not apply directly to the class:
```
ChannelInboundHandler handler = new @Sharable ChannelInboundHandlerAdapter() {
      @Override
      public void channelRead(ChannelHandlerContext context, Object message) {
           context.write(message);
      }
};
```
Tests checking this. If you comment out following code, test will fails:
```
if (!sharable) {       
                AnnotatedType annotatedType = clazz.getAnnotatedSuperclass();
                sharable = annotatedType.isAnnotationPresent(Sharable.class);
 }
```

doom369: Thanks for the explanation.

normanmaurer:
@kiril-me thanks a lot!

PR #8804, opened 2019-01-30
remove explicit assigning default value

thinkerou:
Motivation:

remove explicit assigning default value.


normanmaurer:
As @njhill said I think we should just keep things as they are.

thinkerou:
OK thanks！

PR #8805, opened 2019-01-30
Compare HttpMethod by reference

doom369:
Motivation:

In most cases, `HttpMethod` instance is built from the factory method and the same instance is taken for known Http Methods. So we can implement fast path for `equals()`.

Modification:

- Replace `==` checks with `HttpMethod.equals`.
- `HttpMethod` constructor made private so we can guard users against cases like `new HttpMethod("GET")` in own decoders/encoders and force usres to use `HttpMethod.valueOf` factory method for performance benefit;
- Replaced known `new HttpMethod` with `HttpMethod.valueOf`;

Result:

Comparisons should be a bit faster in some cases.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentEncoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentEncoder.java
@@ -78,10 +78,10 @@ public abstract class HttpContentEncoder extends MessageToMessageCodec<HttpReque
             acceptedEncoding = HttpContentDecoder.IDENTITY;
         }
 
-        HttpMethod meth = msg.method();
-        if (meth == HttpMethod.HEAD) {
+        HttpMethod method = msg.method();
+        if (HttpMethod.HEAD.equals(method)) {
             acceptedEncoding = ZERO_LENGTH_HEAD;
-        } else if (meth == HttpMethod.CONNECT) {
+        } else if (HttpMethod.CONNECT.equals(method)) {
             acceptedEncoding = ZERO_LENGTH_CONNECT;
         }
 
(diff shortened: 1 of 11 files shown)


normanmaurer:
@doom369 I think I would prefer to still use `equals` but change the `equals` implementation and `compareTo` itself to add a fast path:

```java

    @Override
    public boolean equals(Object o) {
        if (o == this) {
            return true;
        }

        if (!(o instanceof HttpMethod)) {
            return false;
        }

        HttpMethod that = (HttpMethod) o;
        return name().equals(that.name());
    }

    @Override
    public int compareTo(HttpMethod o) {
        if (o == this) {
            return 0;
        }
        return name().compareTo(o.name());
    }
```

This would even allow us to backport it to 4.1 with almost no change (the only adjustment) we would need to do is the constructor change).

doom369:
Yeah, was thinking about this approach too, but direct `==` doesn't require method invocation.

I pushed this to master as this change requires `private` constructor and this may affect some users I believe.

normanmaurer:
@doom369 sure but I still would prefer `equals(...)` as the contract is cleaner imho.

While it is true that there could be no two instances of `HEAD` this may not be true anymore at some point. Depending on `equals` is a lot cleaner imho and I would not really worry to much about the extra method call which will most likely be inlined if hot enough anyway.

doom369:
@normanmaurer should I change existing `==` to `equals()` then?

normanmaurer:
@doom369 yes that makes sense imho :) Thanks a lot for all the PRs btw...

doom369:
@normanmaurer happy to help. Changed applied.

normanmaurer:
@netty-bot test this please

doom369:
Made `HttpMethod` `final` to pass checkstyle checks.

normanmaurer:
@netty-bot test this please

normanmaurer:
@ejona86 i think we should for master and backpprt to 4.1 without the breaking changes

normanmaurer:
@doom369 that said maybe @ejona86 is right and we should keep it compatible for 4.1 in this pr. Backport the commit to 4.1 and then do a followup for master where we change it to private + add final. WDYT ?

doom369:
Ok.

PR #8808, opened 2019-01-30
enhancement: use more semantic methods

kezhenxu94:
Motivation:

Improve code readability

Modification:

Replace some bitwise operations with more semantic methods

### DIFF
--- a/transport/src/main/java/io/netty/channel/nio/NioHandler.java
+++ b/transport/src/main/java/io/netty/channel/nio/NioHandler.java
@@ -606,7 +606,7 @@ public final class NioHandler implements IoHandler {
             int readyOps = k.readyOps();
             // We first need to call finishConnect() before try to trigger a read(...) or write(...) as otherwise
             // the NIO JDK channel implementation may throw a NotYetConnectedException.
-            if ((readyOps & SelectionKey.OP_CONNECT) != 0) {
+            if (k.isConnectable()) {
                 // remove OP_CONNECT as otherwise Selector.select(..) will always return without blocking
                 // See https://github.com/netty/netty/issues/924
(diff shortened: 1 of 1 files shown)


normanmaurer:
I am not sure I agree with this... Let me think a bit about it

transport/src/main/java/io/netty/channel/nio/NioHandler.java
```diff
@@ -617,14 +617,14 @@ private void processSelectedKey(SelectionKey k, AbstractNioChannel ch) {
...
-            if ((readyOps & SelectionKey.OP_WRITE) != 0) {
+            if (k.isWritable()) {
                 // Call forceFlush which will also take care of clear the OP_WRITE once there is nothing left to write
                 ch.unsafe().forceFlush();
             }
 
             // Also check for readOps of 0 to workaround possible JDK bug which may otherwise lead
             // to a spin loop
-            if ((readyOps & (SelectionKey.OP_READ | SelectionKey.OP_ACCEPT)) != 0 || readyOps == 0) {
+            if (readyOps == 0 || k.isReadable() || k.isAcceptable()) {
```
normanmaurer: this will be a bit slower as you do one extra check compared to bitwise stuff

normanmaurer: @kezhenxu94 basically what I am saying that before you needed 2 checks while with your its 3 now. That said I think it really not matters too much.

normanmaurer:
@kezhenxu94 I think I actually would prefer to just keep it as it is as we then use bitwise consistent in the implementation and not mixing stuff

kezhenxu94:
Thanks for reviewing

PR #8809, opened 2019-01-30
Update JDK12 and 13 to latest EA releases.

normanmaurer:
Motivation:

We use outdated EA releases when building and testing with JDK 12 and 13.

Modifications:

Update versions.

Result:

Use latest releases


normanmaurer:
interesting... seems like java12 did change something in their SSL impl. need to investigate once I have time

PR #8811, opened 2019-01-30
Cleanup DefaultChannelPipeline implementation

normanmaurer:
Motivation:

The DefaultChannelPipeline implementation can be cleaned up a bit and so we can remove the need for AbstractChannelHandlerContext all together.

Modifications:

- Merge DefautChannelHandlerContext and AbstractChannelHandlerContext
- Remove some unnecessary fields
- Some other minor cleanup

Result:

Cleaner code.


transport/src/main/java/io/netty/channel/AbstractChannelHandlerContext.java
```diff
@@ -1,1184 +0,0 @@
-/*
```
normanmaurer: All the code here was just moved to `DefaultChannelHandlerContext`.

normanmaurer:
@ejona86 yes your assumption is correct

PR #8816, opened 2019-01-30
Cache next context for channelRead(...) and prev context for write(..…

normanmaurer:
….) calls.

Motivation:

To determine the next ChannelHandlerContext that we need to invoke when walking the pipeline we do loop through the whole pipeline (starting at the current context) and do a bitwise operation until we find the one to use. While this is generally not super slow it still can produce some overhead, especially when you have a long pipeline. As channelRead(...) and write(...) are the operations that are called the most usually we can cache the next and prev context for these and so skip the searching for the next / prev context on each call.

Modifications:

- Cache the next and prev context for channelRead(...) / write(...) to reduce overhead
- Adjust benchmark to compare cached vs non cached invokations.

Result:

Less overhead for channelRead(...) / write(...) calls.

```
Benchmark                                                     (extraHandlers)   Mode  Cnt        Score       Error  Units
DefaultChannelPipelineBenchmark.propagateChannelRead                        4  thrpt   10  1030518.619 ± 29880.328  ops/s
DefaultChannelPipelineBenchmark.propagateChannelReadComplete                4  thrpt   10   827194.400 ± 54177.613  ops/s
```
The score for the first is cached the second is not cached.

### DIFF
--- a/microbench/src/main/java/io/netty/microbench/channel/DefaultChannelPipelineBenchmark.java
+++ b/microbench/src/main/java/io/netty/microbench/channel/DefaultChannelPipelineBenchmark.java
@@ -32,11 +32,11 @@ import org.openjdk.jmh.annotations.TearDown;
 import org.openjdk.jmh.annotations.Warmup;
 import org.openjdk.jmh.infra.Blackhole;
 
-@Warmup(iterations = 5)
-@Measurement(iterations = 5)
+@Warmup(iterations = 5,  time = 5)
+@Measurement(iterations = 5, time = 5)
 @State(Scope.Benchmark)
 public class DefaultChannelPipelineBenchmark extends AbstractMicrobenchmark {
-
+    private static final Object MSG = new Object();
     private static final ChannelInboundHandler NOOP_HANDLER = new ChannelInboundHandlerAdapter() {
         @Override
         public boolean isSharable() {
@@ -47,6 +47,11 @@ public class DefaultChannelPipelineBenchmark extends AbstractMicrobenchmark {
     private static final ChannelInboundHandler CONSUMING_HANDLER = new ChannelInboundHandlerAdapter() {
         @Override
         public void channelReadComplete(ChannelHandlerContext ctx) {
+            // NOOP
+        }
+
+        @Override
+        public void channelRead(ChannelHandlerContext ctx, Object msg) {
             // NOOP
         }
 
@@ -76,9 +81,16 @@ public class DefaultChannelPipelineBenchmark extends AbstractMicrobenchmark {
     }
 
     @Benchmark
-    public void propagateEvent(Blackhole hole) {
+    public void propagateChannelReadComplete(Blackhole hole) {
         for (int i = 0; i < 100; i++) {
             hole.consume(pipeline.fireChannelReadComplete());
+        }
+    }
+
+    @Benchmark
+    public void propagateChannelRead(Blackhole hole) {
+        for (int i = 0; i < 100; i++) {
+            hole.consume(pipeline.fireChannelRead(MSG));
         }
     }
 }
--- a/transport/src/main/java/io/netty/channel/DefaultChannelHandlerContext.java
+++ b/transport/src/main/java/io/netty/channel/DefaultChannelHandlerContext.java
@@ -105,8 +105,12 @@ final class DefaultChannelHandlerContext extends DefaultAttributeMap
     // There is no need to make this volatile as at worse it will just create a few more instances then needed.
     private Tasks invokeTasks;
 
+    private DefaultChannelHandlerContext nextChannelRead;
+    private DefaultChannelHandlerContext prevWrite;
+
     DefaultChannelHandlerContext next;
     DefaultChannelHandlerContext prev;
+
     private volatile int handlerState = INIT;
 
     DefaultChannelHandlerContext(DefaultChannelPipeline pipeline, String name,
@@ -133,6 +137,15 @@ final class DefaultChannelHandlerContext extends DefaultAttributeMap
     @Override
     public ChannelHandler handler() {
         return handler;
+    }
+
+    void calculateCachedContexts() {
+        if (next != null) {
+            nextChannelRead = findContextInbound(MASK_CHANNEL_READ);
+        }
+        if (prev != null) {
+            prevWrite = findContextOutbound(MASK_WRITE);
+        }
     }
 
     /**
@@ -434,7 +447,7 @@ final class DefaultChannelHandlerContext extends DefaultAttributeMap
     }
 
     private void findAndInvokeChannelRead(Object msg) {
-        findContextInbound(MASK_CHANNEL_READ).invokeChannelRead(msg);
+        nextChannelRead.invokeChannelRead(msg);
     }
 
     void invokeChannelRead(Object msg) {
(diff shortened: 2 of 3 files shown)


normanmaurer:
This is mainly for discussion as I am not sure yet if this really worth the effort ... Anyway just wanted to show and discuss the idea after talking about this with @nitsanw at some point.

Also @franz1981 and @njhill

normanmaurer:
This also means we would add some memory overhead per context (2 fields). How much this really matters depends a bit on padding etc tho.

njhill:
@normanmaurer i.m.o this is worthwhile, espec. as the memory overhead seems like it would be fairly insignificant in relative terms.

But I'd coincidentally just been thinking of making a similar change which would still probably be worth combining with this: the _existing_ `prev` and `next` fields could be made to always point to the next outgoing and incoming contexts respectively, since the others will be skipped over in all cases (so will benefit all methods not just read/write) - with the exception of `exceptionCaught` in the `next` case, but only while it remains present in `ChannelHandler` (presumably netty 5 is good opp. to remove this deprecated method? :) ). Also, implementing this change I think becomes easier now that you introduced the parallel array list. WDYT?

normanmaurer:
@njhill yep that's a good idea... And for teardown of the pipeline we could just get the "complete ordered" pipeline from the ArrayList via a quick synchronized block :)

normanmaurer:
@netty-bot test this please

ejona86:
This doesn't seem worth it. The benchmark is the absolute best case and any time the pipeline changed we'd need to loop through the entire pipeline to clear the caches.

The benchmark has 4 noops and 1 actual handler, which means only 1/5 of the handlers need to be called. In practice I'd expect the actual ratio to be 1/2 (half inbound, half outbound) to 2/3 (because of codecs).

njhill:
@ejona86 wouldn't it be true in virtually all cases though that marginal increase in pipeline modification cost would be insignificant relative to savings on the IO path?

ejona86:
@njhill, I'm a bit confused by your statement, since it seems to argue against this PR yet you felt the PR was worth it.

I do agree that even if the current benchmark numbers were the true numbers for a non-biased scenario, it still may not be worth the effort. It seemed easier to argue that there were virtually no gains.

I am also bothered by the benchmark measuring ops/s. When it is clear the benchmark is taking 12ns per iteration in the slowest case, that changes the perspective of how important this is or how good the benchmark is.

ejona86:
> All I meant is that I don't think "any time the pipeline changed we'd need to loop through the entire pipeline to clear the caches" should be a concern if doing so allows us to reduce the cost of traversing the pipeline at runtime, even if that cost reduction is much smaller, given the expected orders of magnitude higher frequency.

"at runtime"? Both are at runtime. And what gains?

Consider an HTTP/1 exchange with the HttpStaticFileServer example.

1. added handler HttpStaticFileServerInitializer
2. channelRegistered() fires.
  a. added handler HttpServerCodec, via initializer
  b. added handler HttpObjectAggregator, via initializer
  c. added handler ChunkedWriteHandler, via initializer
  d. added handler HttpStaticFileServerHandler, via initializer
3. channelActive()
4. channelRead(), for GET request
5. channelReadComplete()
6. write(), for small response
7. flush()
8. channelInactive(), client closed
9. channelUnregistered() fires

HttpServerCodec, HttpStaticFileServerHandler, and ChunkedWriteHandler are duplex. HttpObjectAggregator is inbound-only.

So you will pay the clearing of the cache cost 4 times (the one in step (1) could be avoided). There are 5 inbound events that can benefit from this cache, but one will initialize the cache, so we see an improvement for 4 calls × one handler. The benchmark has improvement for 1 call × 4 handlers. The benchmark per-iteration cost fell from 12ns to 9ns.

So for our effort we saved 3ns... except we still need to pay for 4 cache clears. And we have no idea how much that costs.

----

Ah! But we don't care about this case! We care about high throughput cases! Very well.

Let's say the response was large in (6). Then we will see 2x channelWritabilityChanged for every 32 KB (assuming 64 KB/32 KB high/low watermark). Extrapolating, we'd be saving 3ns per 32 KB...

So we have no idea how much we are paying to clear the cache and we won't save more than 3ns per 32 KB.

njhill:
@ejona86 this is not to assert that my reasoning is more valid than yours (it likely isn't! :)), but just to explain my thought process: I assumed that in any context where tput/latency really matters, channels would be mostly persistent since the overall setup cost is relatively high (e.g. protocol negotiation / TLS handshake). Given that, I would think the "cache updating cost" aspect of this can basically be ignored, especially as there's probably a way to do it with more local link rewiring rather than a full refresh each time. It's also why I casually (though incorrectly I agree!) characterized this as "non-runtime".

So it then comes down to whether any speedup on the IO path can justify whatever overhead the change has in terms of maintainability/complexity... and my intuition was that the complexity increase to do the pointer chasing up-front rather than on every read/write event shouldn't be significant. But maybe you're right that the absolute gain is so small that it's not even worth doing that.

It does still seem to me "cleaner"/simplifying somehow to at least keep the inbound/outbound chains comprising just the inbound/outbound handlers respectively, since they are only used for these kinds of IO traversals (and now that there's a separate ArrayList keeping track of the pipeline ordering as a whole).

I agree there won't be benefit to the specific examples you gave, but just because such example exist it doesn't mean there aren't other other realistic cases which could have a small benefit, again assuming negligible cost in terms of code complexity. In particular cases with very high volume of small messages (possibly produced by a multiplying handler) and/or long pipelines which have many uni-directional handlers. A "closer" example from the existing netty codebase would be the pipelines in `SctpEchoTest`.

Finally, apologies for turning this into a much larger debate than is probably warranted!

normanmaurer:
@ejona86 @njhill don't worry... this was just to show the idea and like I said I am happy to just delete the branch and move on :) That said I think @njhill has a point here and we could do a better job in terms of "separate" the inbound and outbound chains. I will provide a pr for doing this and then we can  take a look if it is worth it or not.

ejona86:
> I assumed that in any context where tput/latency really matters, channels would be mostly persistent since the overall setup cost is relatively high (e.g. protocol negotiation / TLS handshake)

That's fair to a point. But I do think there are latency/CPU-sensitive setups that change the pipeline frequently, like HTTP (which in my mind should change the pipeline every request except for trivial cases). The point I made is that the gain was small. It needs to be weighed against hurting other use-cases, and we have no idea the cost to other usecases.

Given the _extremely_ small gains that the benchmark showed (which I honestly don't completely believe; I think the gains are higher), the cost would need to be small as well (including review/discussion).

> But maybe you're right that the absolute gain is so small that it's not even worth doing that.

Yeah, to me there are likely much better gains to be had. For Netty 5, for example, we could remove channelWritabilityChanged when the channel becomes unwritable. It seems like that notification will always be "too late". Although the benchmark shows that would only gain 12ns, which isn't enough gain to even discuss it in my mind; applications that care are better served by increasing the watermarks.

I see a lot of reasons to not trust the benchmark though; it's far too much of a toy and the JIT would be able to make a lot of optimizations.

> It does still seem to me "cleaner"/simplifying somehow to at least keep the inbound/outbound chains comprising just the inbound/outbound handlers respectively

If we want to do that for other reasons, fine. But right now the performance benefit seems limited. I'll note that even when you are careful, with these sorts of nano-optimizations you can trivially make the overall performance worse. Nanobenchmarks do not measure cache pressure, cost of code size increases, and branch mispredictions.

> In particular cases with very high volume of small messages (possibly produced by a multiplying handler) and/or long pipelines which have many uni-directional handlers.

That's fine. But then we should have a benchmark that demonstrates that gain. We should also have a benchmark for a more pathological case and see what the cost is.

My # 1 complaint about this right now is the benchmark.

PR #8819, opened 2019-01-31
Rename SimpleChannelInboundHandler.channelRead0() to messageReceived()

njhill:
Motivation

Per javadoc in 4.1.x `SimpleChannelInboundHandler`:

> Please keep in mind that `channelRead0(ChannelHandlerContext, I)` will be renamed to `messageReceived(ChannelHandlerContext, I)` in 5.0.

Modifications

Rename aforementioned method and all references/overrides.

Result

Method is renamed.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
@@ -427,7 +427,7 @@ public abstract class WebSocketClientHandshaker {
             p.addAfter(ctx.name(), aggregatorName, new HttpObjectAggregator(8192));
             p.addAfter(aggregatorName, "handshaker", new SimpleChannelInboundHandler<FullHttpResponse>() {
                 @Override
-                protected void channelRead0(ChannelHandlerContext ctx, FullHttpResponse msg) throws Exception {
+                protected void messageReceived(ChannelHandlerContext ctx, FullHttpResponse msg) throws Exception {
                     // Remove ourself and do the actual handshake
                     ctx.pipeline().remove(this);
                     try {
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketServerHandshaker.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketServerHandshaker.java
@@ -291,7 +291,7 @@ public abstract class WebSocketServerHandshaker {
         p.addAfter(ctx.name(), aggregatorName, new HttpObjectAggregator(8192));
         p.addAfter(aggregatorName, "handshaker", new SimpleChannelInboundHandler<FullHttpRequest>() {
             @Override
-            protected void channelRead0(ChannelHandlerContext ctx, FullHttpRequest msg) throws Exception {
+            protected void messageReceived(ChannelHandlerContext ctx, FullHttpRequest msg) throws Exception {
                 // Remove ourself and do the actual handshake
                 ctx.pipeline().remove(this);
                 handshake(channel, msg, responseHeaders, promise);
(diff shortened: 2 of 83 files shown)


normanmaurer:
@netty-bot test this please

normanmaurer:
@netty-bot test this please

ejona86:
It would be possible to allow users to use the new name in 4.1, since `SimpleChannelInboundHandler` is an abstract class. Although channelRead0 removal would have to be in 5.0.

normanmaurer:
@ejona86 yeah...  not sure it worse it tho... wdyt ?

ejona86:
I'm a strong fan of being able to pre-migrate code. Having the new name accessible in 4.1 would allow us to mark the old name `@Deprecated` and then have the removal of the old name be a non-event.

njhill:
+1 to @ejona86's suggestion, maybe a shame that wasn't already done before! The only downside I can think of is application code switching to it immediately would break backwards compatibility with any older netty versions. Maybe a more conservative approach could be to first add the new method, but not deprecate `channelRead0` until a later release?

ejona86:
In my mind, an app worried about backward compatibility can just do `@SuppressWarnings("deprecation")` or honestly should just be compiling with the oldest version it supports. But if we really want to split it up into multiple releases that's fine other than if we forget to mark the old one deprecated.

njhill:
@ejona86 it's more that libraries depending on netty may not know what other netty-dependant libraries they will coexist with. In theory they should always be able to use the latest in that case, but I have encountered times when upgrading netty breaks some random library.

ejona86:
@njhill, I understand that problem. But the library could use a new API at any time or could be depending on a bug fix. If a library cares about what the oldest version of netty it is supporting, it can continue to support that older version in the same way it could maintain any workarounds for bugs and similar. If a user is downgrading the netty version to something that isn't supported by the library, then... it isn't supported by the library.

normanmaurer:
@njhill can you do a followup PR for netty 4.1 based on what @ejona86 suggested ?

normanmaurer:
@njhill can you rebase ? I will merge it after it then

normanmaurer:
@netty-bot test this please

normanmaurer:
@njhill please fix the compile errors:

```
13:27:37 [INFO] Netty/Testsuite/Shading ............................ SUCCESS [  9.555 s]
13:27:37 [INFO] Netty/Testsuite/NativeImage ........................ FAILURE [  1.170 s]
13:27:37 [INFO] Netty/Transport/BlockHound/Tests ................... SKIPPED
13:27:37 [INFO] Netty/Microbench ................................... SKIPPED
13:27:37 [INFO] Netty/BOM .......................................... SKIPPED
13:27:37 [INFO] ------------------------------------------------------------------------
13:27:37 [INFO] BUILD FAILURE
13:27:37 [INFO] ------------------------------------------------------------------------
13:27:37 [INFO] Total time: 21:38 min
13:27:37 [INFO] Finished at: 2019-10-31T12:27:37Z
13:27:38 [INFO] Final Memory: 186M/2882M
13:27:38 [INFO] ------------------------------------------------------------------------
13:27:38 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project netty-testsuite-native-image: Compilation failure: Compilation failure: 
13:27:38 [ERROR] /code/testsuite-native-image/src/main/java/io/netty/testsuite/svm/HttpNativeServerHandler.java:[33,7] error: HttpNativeServerHandler is not abstract and does not override abstract method messageReceived(ChannelHandlerContext,HttpObject) in SimpleChannelInboundHandler
13:27:38 [ERROR] /code/testsuite-native-image/src/main/java/io/netty/testsuite/svm/HttpNativeServerHandler.java:[43,4] error: method does not override or implement a method from a supertype
13:27:38 [ERROR] -> [Help 1]
13:27:38 [ERROR] 
13:27:38 [ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
13:27:38 [ERROR] Re-run Maven using the -X switch to enable full debug logging.
```

njhill:
@normanmaurer oops my bad, sorry :( Now fixed

normanmaurer:
@netty-bot test this please

normanmaurer:
@njhill thanks a lot!

PR #8820, opened 2019-01-31
Remove deprecated classes and methods in buffer package

njhill:
Motivation

Hopefully it's ok to remove all deprecated logic in the next major version. Since the changes are fairly extensive, I thought best to do one module at a time.

Modifications

Remove or hide deprecated classes and methods, including `ByteBuf` byte order property.

Result

Cleaner APIs, less code.


njhill:
Some things I'm unsure of:
1. `Unpooled.unmodifiableBuffer(ByteBuf)` was marked deprected in favour of `ByteBuf.asReadOnly()`. The impl of both wraps the buffer in a `ReadOnlyByteBuf`, a public class which was also marked deprecated. So I have made that class package-private but it turns out it's also needed in `ReplayingDecoderByteBuf` which doesn't extend `AbstractByteBuf` and is in the codec package. For this reason I've left the deprecated `Unpooled.unmodifiableBuffer(ByteBuf)` in place for now, not sure of how best to resolve.
2. The `Base64` class implicitly performed LE-based encoding/decoding based on the provided `ByteBuf`'s order. I've removed this logic, but should we be adding it back controlled by some new explicit parameter in the static encode/decode methods? There's already a large number of those methods with different parameter permutations.
3. Wrapping an nio `ByteBuffer` would previously result in a `ByteBuf` with the same byte order. Now the order of the underlying buffer is ignored, which could result in apparently inconsistent behaviour if it's LE. For example `bb.putInt(idx, val)` would have different result to `wrappedBuffer(bb).setInt(idx,val)`. Is this ok? or should `wrappedBuffer` throw if the nio buffer isn't BE?
4. Related: when wrapping a list of `ByteBuffer`s, for now I've left in the check which fails if they are not all of the same byte order, but not sure if that's the best thing to do.

thinkerou:
👍 5.0 version need to remove deprecated class/package? @normanmaurer thanks!

normanmaurer:
Let me close this for now... Sorry we didn't handle this fast enough

PR #8821, opened 2019-01-31
Improve DateFormatter parsing performance

doom369:
Motivation:

Just was looking through code and found 1 interesting place `DateFormatter.tryParseMonth` that was not very effective, so I decided to optimize it a bit.

Modification:

Changed `DateFormatter.tryParseMonth` method. Instead of invocation `regionMatch()` for every month - compare chars one by one.

Result:

`DateFormatter.parseHttpDate` method performance improved from ~3% to ~15%.
```
Benchmark                                                                (DATE_STRING)   Mode  Cnt        Score       Error  Units
DateFormatter2Benchmark.parseHttpHeaderDateFormatter     Sun, 27 Jan 2016 19:18:46 GMT  thrpt    6  4142781.221 ± 82155.002  ops/s
DateFormatter2Benchmark.parseHttpHeaderDateFormatter     Sun, 27 Dec 2016 19:18:46 GMT  thrpt    6  3781810.558 ± 38679.061  ops/s
DateFormatter2Benchmark.parseHttpHeaderDateFormatterNew  Sun, 27 Jan 2016 19:18:46 GMT  thrpt    6  4372569.705 ± 30257.537  ops/s
DateFormatter2Benchmark.parseHttpHeaderDateFormatterNew  Sun, 27 Dec 2016 19:18:46 GMT  thrpt    6  4339785.100 ± 57542.660  ops/s
```

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/DateFormatter.java
+++ b/codec/src/main/java/io/netty/handler/codec/DateFormatter.java
@@ -261,10 +261,6 @@ public final class DateFormatter {
         return false;
     }
 
-    private static boolean matchMonth(String month, CharSequence txt, int tokenStart) {
-        return AsciiString.regionMatchesAscii(month, true, 0, txt, tokenStart, 3);
-    }
-
     private boolean tryParseMonth(CharSequence txt, int tokenStart, int tokenEnd) {
         int len = tokenEnd - tokenStart;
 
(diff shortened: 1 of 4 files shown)


slandelle:
Nice! Which JDK did you use for the benchmark?

normanmaurer:
@doom369 would be interesting what it would look like on latest 1.8 and 11 releases

doom369:
@normanmaurer why? Do you think `regionMatch` was improved in some of JVMs? It actually doesn't matter, as this change mostly removes `regionMatch`es and anyway Netty uses own `regionMatches`.

normanmaurer:
@doom369 was just curious if they did :) Not saying this change is bad, was just wondering in general

normanmaurer:
@doom369 also didn't spot that we use Netty's region match here...

microbench/src/main/java/io/netty/handler/codec/DateFormatter2Benchmark.java
```diff
@@ -0,0 +1,95 @@
+/*
```
normanmaurer: nit: rename to `DateFormatterBenchmark`

normanmaurer:
@netty-bot test this please

microbench/src/main/java/io/netty/handler/codec/DateFormatter2Benchmark.java
doom369: There is already one with that name.

codec/src/main/java/io/netty/handler/codec/DateFormatter.java
```diff
@@ -261,40 +261,40 @@ private boolean tryParseDayOfMonth(CharSequence txt, int tokenStart, int tokenEn
...
         int len = tokenEnd - tokenStart;
 
         if (len != 3) {
             return false;
         }
 
-        if (matchMonth("Jan", txt, tokenStart)) {
+        char monthChar1 = AsciiString.toLowerCase(txt.charAt(tokenStart));
+        char monthChar2 = AsciiString.toLowerCase(txt.charAt(tokenStart + 1));
+        char monthChar3 = AsciiString.toLowerCase(txt.charAt(tokenStart + 2));
```
slandelle: Could you please try with `String#getChars(int srcBegin, int srcEnd, char dst[], int dstBegin)` with a cached `dst` please? This might help as it wouldn't count from scratch for `tokenStart + 1` and `tokenStart + 2`.
And could you please provide benchmark results with JDK8 and JDK11 too, as String implementation has change?.

doom369: `txt` is `CharSequence`. Also, `getChars` does a copy of the array. Not sure I got what do you mean.

doom369:
@slandelle @normanmaurer funny, latest 1.8 Java has even worse performance, but maybe I have to run more cycles.

```
Benchmark                                                                (DATE_STRING)   Mode  Cnt        Score        Error  Units
DateFormatter2Benchmark.parseHttpHeaderDateFormatter     Sun, 27 Jan 2016 19:18:46 GMT  thrpt    6  4046118.084 ± 294274.692  ops/s
DateFormatter2Benchmark.parseHttpHeaderDateFormatter     Sun, 27 Dec 2016 19:18:46 GMT  thrpt    6  3713073.230 ±  26300.420  ops/s
DateFormatter2Benchmark.parseHttpHeaderDateFormatterNew  Sun, 27 Jan 2016 19:18:46 GMT  thrpt    6  4370636.175 ±  56749.370  ops/s
DateFormatter2Benchmark.parseHttpHeaderDateFormatterNew  Sun, 27 Dec 2016 19:18:46 GMT  thrpt    6  4221012.498 ±  27361.655  ops/s
```

slandelle:
What's weird if that "Dec" is consistently slower than `Jan` with original impl. There might be something fishy with `regionMatchesAscii`.

doom369:
@slandelle this is because we have `if else if` condition, so for `Dec` `matchMonth` executed 12 times.

slandelle:
You're right.
Then, have you tried grouping? eg:

```
if (monthChar1 == 'j') {
  if (monthChar2 == 'a' && monthChar3 == 'n') {
    month = Calendar.JANUARY;
  } else if (monthChar2 == 'u') {
    if (monthChar3 == 'n') {
        month = Calendar.JUNE;
    } else if (monthChar3 == 'l') {
        month = Calendar.JULY;

    }
  }
}
```

jan, jun, jul
mar, may
apr, aug

doom369:
Yeah, that was I talking about. It would eliminate a few unnecessary checks, however, I don't think it worth it, as this is the comparison with char, so it is very cheap anyway.

slandelle:
Still, Dec is behind Jan. Maybe grouping the tests would make it closer.

doom369:
`If`s could be replaced with the `switch` (could be faster a bit), or

codec/src/main/java/io/netty/handler/codec/DateFormatter.java
doom369: @slandelle I can't =(. Can't run it due to strange errors.

> Generally speaking, I observe perf regression with JDK11 on most parsers, eg Jackson.
Same here. There were few regressions issues reported to the JDK 11. Also, the main difference - G1. G1 performs 40% worse than ParallelGC on our productions servers, so switching GC could give the large boost.

slandelle:
Before merging, please let me run benchmarks on my side.

normanmaurer:
@slandelle sure... still waiting for at least one more person to approve anyway ;)

slandelle:
I gave it a try on my side and didn't get and significant improvement.
LGTM then!

doom369:
@slandelle did you get a similar improvement? Have you a chance to check with Java 11?

common/src/main/java/io/netty/util/AsciiString.java
```diff
@@ -1827,7 +1827,7 @@ private static byte toLowerCase(byte b) {
         return isUpperCase(b) ? (byte) (b + 32) : b;
     }
 
-    private static char toLowerCase(char c) {
+    public static char toLowerCase(char c) {
```
ejona86: Add documentation.

normanmaurer: @doom369 please address this one... after this is done I will merge :)

doom369: done

Issue #8527, opened 2018-11-13
[Netty 5] ChannelOutboundHandler and exceptionCaught 


normanmaurer:
Outbound errors are typically propagated through the `Future` corresponding to the outbound error (with the exception of `flush()`/`read()`). However the `ChannelOutboundHandler` has an `exceptionCaught` method because it extends `ChannelHandler` which also has `exceptionCaught`. 

`exceptionCaught` will be removed from `ChannelOutboundHandler`, which means we will move the method from `ChannelHandler` to `ChannelInboundHandler`.

PR #8822, opened 2019-01-31
Remove ChannelHandler.exceptionCaught(...) as it should only exist in…

normanmaurer:
… ChannelOutboundHandler

Motivation:

ChannelHandler.exceptionCaught(...) was marked as @deprecated as it should only exist in inbound handlers.

Modifications:

Remove ChannelHandler.exceptionCaught(...) and adjust code / tests.

Result:

Fixes https://github.com/netty/netty/issues/8527

### DIFF
--- a/transport/src/main/java/io/netty/channel/DefaultChannelHandlerContext.java
+++ b/transport/src/main/java/io/netty/channel/DefaultChannelHandlerContext.java
@@ -82,7 +82,7 @@ final class DefaultChannelHandlerContext extends DefaultAttributeMap
     private static final int MASK_WRITE = 1 << 16;
     private static final int MASK_FLUSH = 1 << 17;
 
-    private static final int MASK_ALL_INBOUND = MASK_CHANNEL_REGISTERED |
+    private static final int MASK_ALL_INBOUND = MASK_EXCEPTION_CAUGHT | MASK_CHANNEL_REGISTERED |
             MASK_CHANNEL_UNREGISTERED | MASK_CHANNEL_ACTIVE | MASK_CHANNEL_INACTIVE | MASK_CHANNEL_READ |
             MASK_CHANNEL_READ_COMPLETE | MASK_USER_EVENT_TRIGGERED | MASK_CHANNEL_WRITABILITY_CHANGED;
     private static final int MASK_ALL_OUTBOUND = MASK_BIND | MASK_CONNECT | MASK_DISCONNECT |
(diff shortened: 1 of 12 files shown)


transport/src/main/java/io/netty/channel/DefaultChannelHandlerContext.java
```diff
@@ -744,7 +744,15 @@ private void invokeRead() {
         try {
             ((ChannelOutboundHandler) handler()).read(this);
         } catch (Throwable t) {
-            notifyHandlerException(t);
+            invokeExceptionCaughtFromOutbound(t);
```
ejona86: Aside: Hmmm... Why is `read()` in `ChannelOutboundHandler`? That seems just as wrong as exceptionCaught being on `ChannelInboundHandler`.

transport/src/main/java/io/netty/channel/DefaultChannelHandlerContext.java
```diff
@@ -744,7 +744,15 @@ private void invokeRead() {
...
+        }
+    }
+
+    private void invokeExceptionCaughtFromOutbound(Throwable t) {
+        if ((executionMask & MASK_EXCEPTION_CAUGHT) != 0) {
+            invokeExceptionCaught(t);
```
ejona86: Not `notifyHandlerException()`? It has the `inExceptionCaught()` checks, that seem like they apply here just like anywhere else.

transport/src/main/java/io/netty/channel/DefaultChannelHandlerContext.java
normanmaurer: because `read()` is really an outbound operation.... It tells the `Channel` we want to read something once it is ready... Just like calling `flush()` means we want to write everything that was written before as soon as the `Channel` is writable.

transport/src/main/java/io/netty/channel/DefaultChannelHandlerContext.java
normanmaurer: @ejona86 true... let me fix this

transport/src/main/java/io/netty/channel/DefaultChannelHandlerContext.java
ejona86: read() is outbound from the pipeline's perspective, but not the Handler. Just like exceptionCaught is inbound from the pipeline's perspective, but not the Handler. `read()` interacts with channelRead and channelReadComplete. A ChannelInboundHandler shouldn't need to become a ChannelOutboundHandler because it wants to interact with outbound flow control.

Ah, shoot. It seems `channelWritabilityChanged` is wrong too. A ChannelOutboundHandler observing flow control has to be a ChannelInboundHandler as well...

PR #8826, opened 2019-01-31
Mark ChannelHandlerAdapter.exceptionCaught(...) as @deprecated.

normanmaurer:
Motivation:

41e03adf249ee9d23938fecf9be980a096710b36 marked ChannelHandler.exceptionCaught(...) as @deprecated but missed to also mark ChannelHandlerAdapter.exceptionCaught(...) as @deprecated. We should do so as most people extend the base classes and not implement the interfaces directly.

Modifications:

Mark ChannelHandlerAdapter.exceptionCaught(...) as @deprecated as well.

Result:

Mark method as @deprecated to warn users about its removal.


normanmaurer:
Just noticed this was missing while working on something else.

PR #8832, opened 2019-02-01
Deprecate SICH.channelRead0(...) in favour of messageReceived(...)

njhill:
Motivation

`SimpleInboundChannelHandler.channelRead0(...)` will be renamed in netty 5 to `messageReceived(...)`, but as @ejona86 suggested in #8819, we can still introduce the new name in 4.1 and mark the old one deprecated. This will allow folks to switch over in advance and so reduce the impact when later upgrading to 5.x.

Modifications

Add new `messageReceived` method to `SimpleInboundChannelHandler` which delegates to `channelRead0` by default, and update callers and doc references to use it.

Mark `channelRead0` as deprecated, and add a default implementation which throws an `IllegalStateException`, thereby removing the requirement on subclasses to implement it while still enforcing that at least one of the two methods is overridden.

Result

Method rename pre-migration is possible on 4.1.x.

### DIFF
--- a/transport/src/main/java/io/netty/channel/ChannelHandler.java
+++ b/transport/src/main/java/io/netty/channel/ChannelHandler.java
@@ -72,7 +72,7 @@ import java.lang.annotation.Target;
  *     <b>private boolean loggedIn;</b>
  *
  *     {@code @Override}
- *     public void channelRead0({@link ChannelHandlerContext} ctx, Message message) {
+ *     public void messageReceived({@link ChannelHandlerContext} ctx, Message message) {
  *         if (message instanceof LoginMessage) {
  *             authenticate((LoginMessage) message);
  *             <b>loggedIn = true;</b>
--- a/transport/src/main/java/io/netty/channel/SimpleChannelInboundHandler.java
+++ b/transport/src/main/java/io/netty/channel/SimpleChannelInboundHandler.java
@@ -28,7 +28,7 @@ import io.netty.util.internal.TypeParameterMatcher;
  *             {@link SimpleChannelInboundHandler}&lt;{@link String}&gt; {
  *
  *         {@code @Override}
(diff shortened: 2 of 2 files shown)


ejona86:
Overall, this looks fine to me, although the `@Deprecated` will cause a lot of warnings in Netty's code. @normanmaurer, how do you feel about those warnings? Is it good as-is, or does it also need the majority of the renames from #8819 as well?

njhill:
@ejona86 when you say "majority" presumably you mean only those in final contexts (including anon subclasses)? Otherwise there would be potential for breakage.

Another possibility which should allow everywhere to be renamed would be to change both methods in this PR to have empty default impls and call both consecutively in `channelRead`. Downside of that would be no longer enforcing that at least one is overridden, but I'm actually not sure that's a big deal anyhow?

ejona86:
@njhill, yes, we wouldn't want to break anyone. I was thinking most cases would be safe because it was protected, but yes, non-final classes are common in Netty and I didn't consider that.

But it also appears that every rename in #8819 is safe, as it is in documentation, a test, or example.

njhill:
> But it also appears that every rename in #8819 is safe, as it is in documentation, a test, or example.

@ejona86 good observation! Then my concern is misplaced and maybe it would make sense to change all those other places too. Will wait for advice on how to proceed...

njhill:
@normanmaurer thoughts on @ejona86's suggestion? It will make the number of lines changed much bigger of course...

normanmaurer:
Honestly I think we should just not care and not do anything for now IMHO

njhill:
@normanmaurer this PR was actually originally [requested by you](https://github.com/netty/netty/pull/8819#issuecomment-459476247) :) based on a suggestion from @ejona86's, but closing it is fine by me!

normanmaurer:
Closing then :)

PR #8833, opened 2019-02-01
Rename MultithreadEventExecutorGroup to MultiThreadEventExecutorGroup…

normanmaurer:
… and MultithreadEventLoopGroup to MultiThreadEventLoopGroup

Motivation:

We should rename MultithreadEventExecutorGroup and MultithreadEventLoopGroup to MultiThread... to be more inline with SingleThreadEventExecutor / SingleThreadEventLoop.

Modifications:

Rename classes.

Result:

More consistent naming.

### DIFF
--- a/codec-http/src/test/java/io/netty/handler/codec/http/HttpClientCodecTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/HttpClientCodecTest.java
@@ -25,7 +25,7 @@ import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInitializer;
 import io.netty.channel.ChannelOption;
-import io.netty.channel.MultithreadEventLoopGroup;
+import io.netty.channel.MultiThreadEventLoopGroup;
 import io.netty.channel.SimpleChannelInboundHandler;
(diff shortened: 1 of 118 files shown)


ejona86:
Even if it is wrong, this doesn't seem worth the churn. But it's also not clear it is wrong, since one would be written "single thread" but the other is virtually always "multi-thread" or "multithread".

normanmaurer:
@ejona86 well it really is not so much churn in reality... It only is so much churn in this PR as we changed on how we construct `EventLoopGroup`s. In 4.1 really almost no-one did use `MultithreadEvent*Group` directly.

normanmaurer:
@ejona86 WDYT ?

ejona86:
As I mentioned, it's not clear that the new name is superior. To mirror `SingleThread` it seems it would need to be `MultipleThread`. `Multithread` seems an appropriate casing today.

trustin:
https://en.wikipedia.org/wiki/Thread_(computing)#Single_threading
https://en.wikipedia.org/wiki/Thread_(computing)#Multithreading

I remember you asked me to rename it to `MultiThread*` long time ago, but I told you `Multithread` is actually correct. :-)

normanmaurer:
Alright lets close this one then

PR #8834, opened 2019-02-01
Fix varargs parameter logging in LocationAwareSlf4JLogger

njhill:
Motivation

As pointed out by @91he in https://github.com/netty/netty/pull/8595#issuecomment-459181794, there is a remaining bug in `LocationAwareSlf4JLogger` following the updates done in #8595. The logging methods which take a varargs message parameter array should format using `MessageFormatter.arrayFormat` rather than `MessageFormatter.format`.

Modifications

Change varargs param methods in `LocationAwareSlf4JLogger` to use `MessageFormatter.arrayFormat` and extend unit test to cover these cases.

Results

Correct log output when logging messages with > 2 parameters when using `LocationAwareSlf4JLogger`.


normanmaurer:
@netty-bot test this please

normanmaurer:
@njhill thanks a lot! I did also cherry-pick to master.

PR #8836, opened 2019-02-02
Simplify WebSocketVersion code

doom369:
Motivation:

`WebSocketVersion.UNKNOWN` is not actually used, so could be removed and `WebSocketVersion.toHttpHeaderValue()` simplified.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketVersion.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketVersion.java
@@ -25,49 +25,43 @@ package io.netty.handler.codec.http.websocketx;
  * </p>
  */
 public enum WebSocketVersion {
-    UNKNOWN,
+    UNKNOWN("unknown"),
 
     /**
      * <a href= "http://tools.ietf.org/html/draft-ietf-hybi-thewebsocketprotocol-00"
      * >draft-ietf-hybi-thewebsocketprotocol- 00</a>.
      */
-    V00,
+    V00("0"),
 
     /**
      * <a href= "http://tools.ietf.org/html/draft-ietf-hybi-thewebsocketprotocol-07"
      * >draft-ietf-hybi-thewebsocketprotocol- 07</a>
      */
-    V07,
+    V07("7"),
 
     /**
      * <a href= "http://tools.ietf.org/html/draft-ietf-hybi-thewebsocketprotocol-10"
      * >draft-ietf-hybi-thewebsocketprotocol- 10</a>
      */
-    V08,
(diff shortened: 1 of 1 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketVersion.java
```diff
@@ -25,49 +25,42 @@
...
      * <a href="http://tools.ietf.org/html/rfc6455 ">RFC 6455</a>. This was originally <a href=
      * "http://tools.ietf.org/html/draft-ietf-hybi-thewebsocketprotocol-17" >draft-ietf-hybi-thewebsocketprotocol-
      * 17</a>
      */
-    V13;
+    V13("13");
+
+    private final String version;
+
+    WebSocketVersion(String version) {
```
normanmaurer: private ?

normanmaurer:
@netty-bot test this please

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketVersion.java
doom369: @normanmaurer not necessary:

https://stackoverflow.com/questions/7747948/why-can-a-enum-have-a-package-private-constructor

doom369:
> It appears WebSocketVersion is not returned anywhere today, but if it were then UNKNOWN may be used.

I'm not sure about this. This argument could be applied to any removed code. `UNKNOWN` is not used in current flows, so should be removed. When it will - it could be added.

normanmaurer:
@ejona86 any more comments ?

ejona86:
Enums having an "UNKNOWN" value is a normal style. It's generally used an an alternative to using null Enum values. Looking back at history, it seems that's exactly what the UNKNOWN value is/was. Just because it isn't used at the moment really has no bearing on whether it should be removed, in my mind.

I agree that using UNKNOWN or not is a stylistic thing, but style things are generally useless to change later since two people could change them every day (back and forth). It's doing its intended job, so just leave it.

normanmaurer:
@doom369

doom369:
@normanmaurer sorry for delay. `UNKNOWN` returned.

normanmaurer:
@doom369 seems like we could even back port this to 4.1, correct ?

normanmaurer:
@ejona86 PTAL again once you have time

doom369:
@normanmaurer nope, as exception is no longer thrown.

normanmaurer:
ah right... @doom369

normanmaurer:
@doom369 thanks!

PR #8840, opened 2019-02-03
migrate java8: use requireNonNull

thinkerou:
Motivation:

use `requireNonNull`


normanmaurer:
@netty-bot test this please

PR #8843, opened 2019-02-04
Don't update state of PromiseCombiner when finish(null) is called

normanmaurer:
Motivation:

When we fail a call to PromiseCombiner.finish(...) because of a null argument we must not update the internal state before throwing.

Modifications:

- First do the null check and only after we validated that the argument is not null update the internal state
- Add test case.

Modifications:

Do not mess up internal state of PromiseCombiner when finish(...) is called with a null argument.


normanmaurer:
noticed this when working on something else.

PR #8844, opened 2019-02-05
Add checkClosedInterval method

thinkerou:
Motivation:

add `checkClosedInterval(a, i, j, “a”)` instead of `if (a < i || a > j) {}`

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java
+++ b/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java
@@ -16,6 +16,7 @@
 
 package io.netty.buffer;
 
+import static io.netty.util.internal.ObjectUtil.checkClosedInterval;
 import static io.netty.util.internal.ObjectUtil.checkPositiveOrZero;
 
 import io.netty.util.NettyRuntime;
@@ -284,9 +285,7 @@ public class PooledByteBufAllocator extends AbstractByteBufAllocator implements 
     }
 
     private static int validateAndCalculateChunkSize(int pageSize, int maxOrder) {
-        if (maxOrder > 14) {
-            throw new IllegalArgumentException("maxOrder: " + maxOrder + " (expected: 0-14)");
-        }
+        checkClosedInterval(maxOrder, 0, 14, "maxOrder");
 
         // Ensure the resulting chunkSize does not overflow.
         int chunkSize = pageSize;
--- a/resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
+++ b/resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
@@ -26,8 +26,8 @@ import java.util.List;
 import java.util.concurrent.ConcurrentMap;
 
 import static io.netty.util.internal.ObjectUtil.checkClosedInterval;
-import static io.netty.util.internal.ObjectUtil.checkNotNull;
 import static io.netty.util.internal.ObjectUtil.checkPositiveOrZero;
+import static java.util.Objects.requireNonNull;
 
 /**
  * Default implementation of {@link DnsCache}, backed by a {@link ConcurrentMap}.
@@ -113,7 +113,7 @@ public class DefaultDnsCache implements DnsCache {
 
     @Override
     public boolean clear(String hostname) {
-        checkNotNull(hostname, "hostname");
+        requireNonNull(hostname, "hostname");
         return resolveCache.clear(appendDot(hostname));
     }
 
@@ -123,7 +123,7 @@ public class DefaultDnsCache implements DnsCache {
 
     @Override
     public List<? extends DnsCacheEntry> get(String hostname, DnsRecord[] additionals) {
-        checkNotNull(hostname, "hostname");
+        requireNonNull(hostname, "hostname");
         if (!emptyAdditionals(additionals)) {
             return Collections.emptyList();
         }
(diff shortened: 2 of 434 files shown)


buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java
```diff
@@ -284,9 +285,7 @@ private static int validateAndCalculatePageShifts(int pageSize) {
     }
 
     private static int validateAndCalculateChunkSize(int pageSize, int maxOrder) {
-        if (maxOrder > 14) {
-            throw new IllegalArgumentException("maxOrder: " + maxOrder + " (expected: 0-14)");
-        }
+        checkClosedInterval(maxOrder, 0, 14, "maxOrder");
```
thinkerou: because `expected: 0-14)`

carl-mastrangelo:
@normanmaurer quick thoughts:

* I like the idea for better error messages, and have done something similar on my own projects.
* `checkClosedInterval` does not give better messages, instead mixing the min and max case together.
* I prefer min/max as the bounds, rather than left/right.  

Not sure about benefit vs. the churn; I could go either way.

thinkerou:
@carl-mastrangelo thanks!
```
* I prefer min/max as the bounds, rather than left/right.
```
done!

resolver-dns/src/main/java/io/netty/resolver/dns/DefaultDnsCache.java
```diff
@@ -75,10 +76,7 @@ public DefaultDnsCache() {
     public DefaultDnsCache(int minTtl, int maxTtl, int negativeTtl) {
         this.minTtl = Math.min(Cache.MAX_SUPPORTED_TTL_SECS, checkPositiveOrZero(minTtl, "minTtl"));
         this.maxTtl = Math.min(Cache.MAX_SUPPORTED_TTL_SECS, checkPositiveOrZero(maxTtl, "maxTtl"));
-        if (minTtl > maxTtl) {
-            throw new IllegalArgumentException(
-                    "minTtl: " + minTtl + ", maxTtl: " + maxTtl + " (expected: 0 <= minTtl <= maxTtl)");
-        }
+        checkClosedInterval(minTtl, 0, maxTtl, "minTtl");
```
ejona86: This is going to give a misleading error message.

Ditto for DefaultDnsCnameCache.java

resolver-dns/src/main/java/io/netty/resolver/dns/UnixResolverDnsServerAddressStreamProvider.java
```diff
@@ -173,10 +174,8 @@ private boolean mayOverrideNameServers() {
                     }
                     if (line.startsWith(NAMESERVER_ROW_LABEL)) {
                         int i = indexOfNonWhiteSpace(line, NAMESERVER_ROW_LABEL.length());
-                        if (i < 0) {
-                            throw new IllegalArgumentException("error parsing label " + NAMESERVER_ROW_LABEL +
-                                    " in file " + etcResolverFile + ". value: " + line);
-                        }
+                        checkPositiveOrZero(i, "error parsing label " + NAMESERVER_ROW_LABEL +
```
ejona86: We should not do any string formatting unless we are throwing the exception. Otherwise we'll do string formatting on the fast path, when there are no errors, just to throw it away.

transport-native-epoll/src/main/java/io/netty/channel/epoll/TcpMd5Util.java
```diff
@@ -40,14 +41,7 @@
...
             requireNonNull(key, "newKeys[" + e.getKey() + "]");
-            if (key.length == 0) {
-                throw new IllegalArgumentException("newKeys[" + e.getKey() + "] has an empty key.");
-            }
-            if (key.length > Native.TCP_MD5SIG_MAXKEYLEN) {
-                throw new IllegalArgumentException("newKeys[" + e.getKey() +
-                    "] has a key with invalid length; should not exceed the maximum length (" +
-                        Native.TCP_MD5SIG_MAXKEYLEN + ')');
-            }
+            checkClosedInterval(key.length, 1, Native.TCP_MD5SIG_MAXKEYLEN, "newKeys[" + e.getKey() + "]");
```
ejona86: This is string formatting as well. Although apparently it was already computed once above? Still not a great idea.

transport/src/main/java/io/netty/channel/AdaptiveRecvByteBufAllocator.java
```diff
@@ -164,13 +165,9 @@ public AdaptiveRecvByteBufAllocator() {
...
     public AdaptiveRecvByteBufAllocator(int minimum, int initial, int maximum) {
-        checkPositive(minimum, "minimum");
-        if (initial < minimum) {
-            throw new IllegalArgumentException("initial: " + initial);
-        }
-        if (maximum < initial) {
-            throw new IllegalArgumentException("maximum: " + maximum);
-        }
+        // 1 <= minimum <= initial <= maximum.
+        checkClosedInterval(minimum, 1, initial, "minimum");
```
ejona86: I'd rather not use the methods in this case, since it isn't 'minimum' that may be wrong, it could be 'initial'. Yes, the pre-existing code behave similarly, but it would have been easy to fix.

transport/src/main/java/io/netty/channel/DefaultFileRegion.java
```diff
@@ -118,12 +119,8 @@ public long transferred() {
 
     @Override
     public long transferTo(WritableByteChannel target, long position) throws IOException {
+        checkClosedInterval(position, 0, this.count - 1, "position");
```
ejona86: Can't position be the same as `this.count`? Why the -1?

transport/src/main/java/io/netty/channel/WriteBufferWaterMark.java
```diff
@@ -56,13 +57,7 @@ public WriteBufferWaterMark(int low, int high) {
...
     WriteBufferWaterMark(int low, int high, boolean validate) {
         if (validate) {
-            checkPositiveOrZero(low, "low");
-            if (high < low) {
-                throw new IllegalArgumentException(
-                        "write buffer's high water mark cannot be less than " +
-                                " low water mark (" + low + "): " +
-                                high);
-            }
+            checkClosedInterval(low, 0, high, "write buffer's low water mark");
```
ejona86: This produces a worse error message.

normanmaurer:
After thinking more about this and also taking @ejona86 comments into account I think we should just not do this. Sorry for wasting your time @thinkerou  :/

PR #8845, opened 2019-02-05
Drop SPDY support

slandelle:
Motivation:

SPDY has been superseded by HTTP/2. Chrome has dropped support in 2016 and GFE no longer negociate it.

Modifications:

* drop codec
* drop examples
* drop constants from `ApplicationProtocolNames`

Result:

SPDY support dropped from Netty 5


normanmaurer:
@slandelle thanks a lot

Issue #7862, opened 2018-04-11
[Question] Custom TrustManager and performance impact ?


sbernard31:
If I want to define a custom behavior to validate certificate, I should implement a custom X509TrustManager, right ?

Now imagine, that my TrustManager needs to do some "slow" operation like calling a micro service or maybe doing some OCSP check. As TrustManager is a sync API, I ask my self how would this impact netty "performance".

I know that Netty is design for Better throughput, lower latency so how could be the right way to manage this use case ?

Thx :)

normanmaurer:
That would be very bad as you would block the EventLoop. So to make it short don’t do it

sbernard31:
Ok ok, but checking certificate using OCSP for example is not so eccentric and this could take time, So where should I do this kind of check if this is not in TrustManager ?

Thx again :)

johnou:
@sbernard31 try extending Netty's OCSP.

https://github.com/netty/netty/blob/4.1/example/src/main/java/io/netty/example/ocsp/OcspServerExample.java

sbernard31:
@johnou, maybe I missed something, but this is an example about how doing OCSP request, right ? 
But my question is more about where to plug the OCSP validation in a server. The natural way seems to use a custom TrustManager, but as I suspected this could be an issue if OCSP request take too many time as TrustManager is a sync API.

Does it could make sense to add a new async API like TrustManager ?

sbernard31:
@normanmaurer thx for the answer. Should I close the issue ?

Is there plan about introduce a way to do that later ? (I can open an issue/improvement If you want ?)

rkapsi:
To chime in here in regards OCSP. Normally it's the client that does the validation. The server fetches it out-of-band from the CA's OCSP responder server and caches it for up to the TTL that is configured in CA's response (I believe 10 days is the max). That is something you wouldn't want to do inline during the handshake.

As for the client side. If you were to write a Netty based SSL client and wanted to do OCSP validation that involves needing to reach out to the CA then it's indeed a problem. The underlying OpenSSL API is synchronous.

I believe a solution for dealing with this could be a dedicated SSL handhsake Thread pool so that the normal I/O event loop doesn't get blocked for other connections that are past the handshake (I think there is an open ticket for that).

ZhenLian:
@normanmaurer 

[...]

Issue #7020, opened 2017-07-25
Concurrent SSL Handshakes cause IO threads to stay busy for long time


anshuman-osc:
### Expected behavior
Existing connections should not be impacted and should continue to exchange messages if new connections are added 

### Actual behavior
If there are a few thousand connections to the server, and then we get a burst of say another few thousand connections, these new SSL handshakes keep the worker threads busy for a long time, which causes the existing connections to start timing out.

### Steps to reproduce
1. Compile and build the reproducer code using
 mvn clean package assembly:single
2. Start the tcp server using this command:
java -jar tcp-server-1.0.0-SNAPSHOT-jar-with-dependencies.jar
This starts a TCP server with 8 IO threads and 1 acceptor thread on port 8000. For customizing, look at the class TcpServer

3. Open mission control and monitor the Connection Count using the TcpServer.getConnectionCount managed attribute. This should be 0 to start with

4. On a second machine, trigger the load using the tcp client jar.
java -jar tcp-client-1.0.0-SNAPSHOT-jar-with-dependencies.jar <server-host>
This will trigger 5000 concurrent handshakes to the server. These connections are kept open and the client keeps sending requests to the server (1 at a time). These values can also be customized using the LoadRunner class

5. Monitor for a couple of minutes, the connection count reported would be 5000

6. On a third machine, repeat step 4 and monitor the connection count. You will see the connection count to rise to 10000, and after a couple of minutes the clients start timing out (graph attached)

### Minimal yet complete reproducer code (or URL to code)
https://github.com/anshuman-osc/netty.git

### Netty version
4.1.6.Final

### JVM version (e.g. `java -version`)
java version "1.8.0_65"
Java(TM) SE Runtime Environment (build 1.8.0_65-b17)
Java HotSpot(TM) 64-Bit Server VM (build 25.65-b01, mixed mode)

### OS version (e.g. `uname -a`)
Tested on 2 Linux distributions: 
Linux 4.4.0-83-generic #106-Ubuntu SMP Mon Jun 26 17:54:43 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

Linux 3.0.101-63-default #1 SMP Tue Jun 23 16:02:31 UTC 2015 (4b89d0c) x86_64 x86_64 x86_64 GNU/Linux

Scottmitch:
ChannelInboundHandlers will execute on the EventLoop thread by default. Here are general recommendations to work through this performance issue:

[...]

PR #8847, opened 2019-02-06
Support using an Executor to offload blocking / long-running tasks wh…

normanmaurer:
…en processing TLS / SSL via the SslHandler.

Motivation:

The SSLEngine does provide a way to signal to the caller that it may need to execute a blocking / long-running task which then can be offloaded to an Executor to ensure the I/O thread is not blocked. Currently how we handle this in SslHandler is not really optimal as while we offload to the Executor we still block the I/O Thread.

Modifications:

- Correctly support offloading the task to the Executor while suspending processing of SSL in the I/O Thread
- Add new methods to SslContext to specify the Executor when creating a SslHandler
- Remove @deprecated annotations from SslHandler constructor that takes an Executor
- Adjust tests to also run with the Executor to ensure all works as expected.

Result:

Be able to offload long running tasks to an Executor when using SslHandler. Partly fixes https://github.com/netty/netty/issues/7862 and https://github.com/netty/netty/issues/7020.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SniHandler.java
@@ -15,6 +15,7 @@
  */
 package io.netty.handler.ssl;
 
+import io.netty.buffer.ByteBufAllocator;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.codec.DecoderException;
 import io.netty.util.AsyncMapping;
@@ -129,7 +130,7 @@ public class SniHandler extends AbstractSniHandler<SslContext> {
     protected void replaceHandler(ChannelHandlerContext ctx, String hostname, SslContext sslContext) throws Exception {
         SslHandler sslHandler = null;
         try {
-            sslHandler = sslContext.newHandler(ctx.alloc());
+            sslHandler = newSslHandler(sslContext, ctx.alloc());
             ctx.pipeline().replace(this, SslHandler.class.getName(), sslHandler);
             sslHandler = null;
         } finally {
@@ -140,6 +141,14 @@ public class SniHandler extends AbstractSniHandler<SslContext> {
                 ReferenceCountUtil.safeRelease(sslHandler.engine());
             }
         }
+    }
+
+    /**
+     * Returns a new {@link SslHandler} using the given {@link SslContext} and {@link ByteBufAllocator}.
+     * Users may override this method to implement custom behavior.
+     */
+    protected SslHandler newSslHandler(SslContext context, ByteBufAllocator allocator) {
+        return context.newHandler(allocator);
     }
 
     private static final class AsyncMappingAdapter implements AsyncMapping<String, SslContext> {
--- a/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
+++ b/handler/src/main/java/io/netty/handler/ssl/SslHandler.java
@@ -33,6 +33,7 @@ import io.netty.channel.ChannelPipeline;
 import io.netty.channel.ChannelPromise;
 import io.netty.channel.ChannelPromiseNotifier;
 import io.netty.handler.codec.ByteToMessageDecoder;
+import io.netty.handler.codec.DecoderException;
 import io.netty.handler.codec.UnsupportedMessageTypeException;
 import io.netty.util.ReferenceCountUtil;
 import io.netty.util.ReferenceCounted;
@@ -55,10 +56,9 @@ import java.nio.ByteBuffer;
 import java.nio.channels.ClosedChannelException;
 import java.nio.channels.DatagramChannel;
 import java.nio.channels.SocketChannel;
-import java.util.ArrayList;
 import java.util.List;
-import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
@@ -390,6 +390,7 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
     private boolean flushedBeforeHandshake;
     private boolean readDuringHandshake;
     private boolean handshakeStarted;
+
     private SslHandlerCoalescingBufferQueue pendingUnencryptedWrites;
     private Promise<Channel> handshakePromise = new LazyChannelPromise();
     private final LazyChannelPromise sslClosePromise = new LazyChannelPromise();
@@ -402,6 +403,7 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
 
     private boolean outboundClosed;
     private boolean closeNotify;
+    private boolean processTask;
 
     private int packetLength;
 
@@ -417,7 +419,7 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
     volatile int wrapDataSize = MAX_PLAINTEXT_LENGTH;
 
     /**
-     * Creates a new instance.
+     * Creates a new instance which runs all delegating tasks directly on the {@link EventExecutor}.
      *
      * @param engine  the {@link SSLEngine} this handler will use
      */
@@ -426,29 +428,36 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
     }
 
     /**
-     * Creates a new instance.
+     * Creates a new instance which runs all delegating tasks directly on the {@link EventExecutor}.
      *
      * @param engine    the {@link SSLEngine} this handler will use
      * @param startTls  {@code true} if the first write request shouldn't be
      *                  encrypted by the {@link SSLEngine}
      */
-    @SuppressWarnings("deprecation")
     public SslHandler(SSLEngine engine, boolean startTls) {
         this(engine, startTls, ImmediateExecutor.INSTANCE);
     }
 
     /**
-     * @deprecated Use {@link #SslHandler(SSLEngine)} instead.
+     * Creates a new instance.
+     *
+     * @param engine  the {@link SSLEngine} this handler will use
+     * @param delegatedTaskExecutor the {@link Executor} that will be used to execute tasks that are returned by
+     *                              {@link SSLEngine#getDelegatedTask()}.
      */
-    @Deprecated
     public SslHandler(SSLEngine engine, Executor delegatedTaskExecutor) {
         this(engine, false, delegatedTaskExecutor);
     }
 
     /**
-     * @deprecated Use {@link #SslHandler(SSLEngine, boolean)} instead.
+     * Creates a new instance.
+     *
+     * @param engine  the {@link SSLEngine} this handler will use
+     * @param startTls  {@code true} if the first write request shouldn't be
+     *                  encrypted by the {@link SSLEngine}
+     * @param delegatedTaskExecutor the {@link Executor} that will be used to execute tasks that are returned by
+     *                              {@link SSLEngine#getDelegatedTask()}.
      */
-    @Deprecated
     public SslHandler(SSLEngine engine, boolean startTls, Executor delegatedTaskExecutor) {
         if (engine == null) {
             throw new NullPointerException("engine");
@@ -774,6 +783,10 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
             return;
         }
 
+        if (processTask) {
+            return;
+        }
+
         try {
             wrapAndFlush(ctx);
         } catch (Throwable cause) {
@@ -813,7 +826,7 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
             final int wrapDataSize = this.wrapDataSize;
             // Only continue to loop if the handler was not removed in the meantime.
             // See https://github.com/netty/netty/issues/5860
-            while (!ctx.isRemoved()) {
+            outer: while (!ctx.isRemoved()) {
                 promise = ctx.newPromise();
                 buf = wrapDataSize > 0 ?
                         pendingUnencryptedWrites.remove(alloc, wrapDataSize, promise) :
@@ -850,7 +863,11 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
 
                     switch (result.getHandshakeStatus()) {
                         case NEED_TASK:
-                            runDelegatedTasks();
+                            if (!runDelegatedTasks(inUnwrap)) {
+                                // We scheduled a task on the delegatingTaskExecutor, so stop processing as we will
+                                // resume once the task completes.
+                                break outer;
+                            }
                             break;
                         case FINISHED:
                             setHandshakeSuccess();
@@ -919,7 +936,7 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
         try {
             // Only continue to loop if the handler was not removed in the meantime.
             // See https://github.com/netty/netty/issues/5860
-            while (!ctx.isRemoved()) {
+            outer: while (!ctx.isRemoved()) {
                 if (out == null) {
                     // As this is called for the handshake we have no real idea how big the buffer needs to be.
                     // That said 2048 should give us enough room to include everything like ALPN / NPN data.
@@ -941,7 +958,11 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
                         setHandshakeSuccess();
                         return false;
                     case NEED_TASK:
-                        runDelegatedTasks();
+                        if (!runDelegatedTasks(inUnwrap)) {
+                            // We scheduled a task on the delegatingTaskExecutor, so stop processing as we will
+                            // resume once the task completes.
+                            break outer;
+                        }
                         break;
                     case NEED_UNWRAP:
                         if (inUnwrap) {
@@ -1243,6 +1264,9 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
 
     @Override
     protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) throws SSLException {
+        if (processTask) {
+            return;
+        }
         if (jdkCompatibilityMode) {
             decodeJdkCompatible(ctx, in);
         } else {
@@ -1252,6 +1276,10 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
 
     @Override
     public void channelReadComplete(ChannelHandlerContext ctx) throws Exception {
+        channelReadComplete0(ctx);
+    }
+
+    private void channelReadComplete0(ChannelHandlerContext ctx) {
         // Discard bytes of the cumulation buffer if needed.
         discardSomeReadBytes();
 
@@ -1366,7 +1394,11 @@ public class SslHandler extends ByteToMessageDecoder implements ChannelOutboundH
                         }
                         break;
                     case NEED_TASK:
-                        runDelegatedTasks();
+                        if (!runDelegatedTasks(true)) {
+                            // We scheduled a task on the delegatingTaskExecutor, so stop processing as we will
+                            // resume once the task completes.
+                            break unwrapLoop;
+                        }
                         break;
                     case FINISHED:
                         setHandshakeSuccess();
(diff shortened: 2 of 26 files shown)


normanmaurer:
Next step is to also allow our native SSLEngine to run verification in an extra thread to allow doing blocking operations in the TrustManager / KeyManager without blocking the I/O thread.

handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
```diff
@@ -540,7 +552,7 @@ public void testMutualAuthDiffCertsServerFailure() throws Exception {
         // Client trusts server but server only trusts itself
         mySetupMutualAuth(serverCrtFile, serverKeyFile, serverCrtFile, serverKeyPassword,
                           serverCrtFile, clientKeyFile, clientCrtFile, clientKeyPassword);
-        assertTrue(serverLatch.await(2, TimeUnit.SECONDS));
+        assertTrue(serverLatch.await(200, TimeUnit.SECONDS));
```
carl-mastrangelo: woah, intentional?

testsuite/src/main/java/io/netty/testsuite/transport/socket/SocketSslGreetingTest.java
```diff
@@ -119,46 +134,53 @@ public void testSslGreeting(ServerBootstrap sb, Bootstrap cb) throws Throwable {
...
+                    ChannelPipeline p = sch.pipeline();
+                    p.addLast(newSslHandler(clientCtx, sch.alloc(), executorService));
+                    p.addLast(new LoggingHandler(LOG_LEVEL));
+                    p.addLast(ch);
+                }
+            });
 
-        Channel sc = sb.bind().sync().channel();
-        Channel cc = cb.connect(sc.localAddress()).sync().channel();
+            Channel sc = sb.bind().sync().channel();
```
rgolden: Nit: tabs instead of spaces.

handler/src/test/java/io/netty/handler/ssl/SSLEngineTest.java
normanmaurer: nope... left-over from debugging... Let me fix.

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1366,7 +1394,11 @@ private int unwrap(
                         }
                         break;
                     case NEED_TASK:
-                        runDelegatedTasks();
+                        if (!runDelegatedTasks(true)) {
+                            // We scheduled a task on the delegatingTaskExecutor, so stop processing as we will
+                            // resume once the task completes.
```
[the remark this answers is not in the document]
normanmaurer: we still may need to notify about the closure of the SSLEngine. I will add some comments.

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1448,64 +1480,220 @@ private static ByteBuffer toByteBuffer(ByteBuf out, int index, int len) {
...
+     * Will either run the delegating task directly calling {@link Runnable#run()} and return {@code true} or will
+     * offload the delegating task using {@link Executor#execute(Runnable)} and return {@code false}.
+     *
+     * If the task is offloaded it will take care to resume its work on the {@link EventExecutor} once there is
+     * no more task to process.
      */
-    private void runDelegatedTasks() {
-        if (delegatedTaskExecutor == ImmediateExecutor.INSTANCE) {
+    private boolean runDelegatedTasks(boolean inUnwrap) {
+        if (delegatedTaskExecutor == ImmediateExecutor.INSTANCE || (delegatedTaskExecutor instanceof EventExecutor
```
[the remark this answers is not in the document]
normanmaurer: unfortunately this still not fits in the 120 char limit :(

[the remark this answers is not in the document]
normanmaurer: let me just extract the eventLoop check in an extra method to make it fit.

normanmaurer:
@ejona86 addressed your comments

normanmaurer:
ups @carl-mastrangelo I mean 🤣

ejona86:
CC @sanjaypujare

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1447,65 +1484,224 @@ private static ByteBuffer toByteBuffer(ByteBuf out, int index, int len) {
...
+            return false;
+        }
+    }
+
+    private void executeDelegatingTask(Runnable task, boolean inUnwrap) {
+        processTask = true;
+        try {
+            delegatedTaskExecutor.execute(new SslTask(task, inUnwrap));
+        } catch (RejectedExecutionException e) {
+            processTask = false;
```
rkapsi: Any risks for some other exception occurring and processTask falsely staying true?

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1447,65 +1484,224 @@ private static ByteBuffer toByteBuffer(ByteBuf out, int index, int len) {
...
+     * Will either run the delegating task directly calling {@link Runnable#run()} and return {@code true} or will
+     * offload the delegating task using {@link Executor#execute(Runnable)} and return {@code false}.
+     *
+     * If the task is offloaded it will take care to resume its work on the {@link EventExecutor} once there are no
+     * more tasks to process.
      */
-    private void runDelegatedTasks() {
-        if (delegatedTaskExecutor == ImmediateExecutor.INSTANCE) {
+    private boolean runDelegatedTasks(boolean inUnwrap) {
+        if (delegatedTaskExecutor == ImmediateExecutor.INSTANCE || inEventLoop(delegatedTaskExecutor)) {
```
rkapsi: For simplicity/no special case/unify maybe omit the ImmediateExecutor.INSTANCE check and let inEventLoop() take care of it.

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
normanmaurer: I dont think so...

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
normanmaurer: You mean basically move all the code to the new method ?

rkapsi: Oh, I think it's nvm. ImmediateExecutor is not an EventExecutor (mixed it up with ImmediateEventExecutor).

normanmaurer:
@netty-bot test this please

normanmaurer:
test failure is not related .... lets trigger the CI again...

@netty-bot test this please

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1447,65 +1484,224 @@ private static ByteBuffer toByteBuffer(ByteBuf out, int index, int len) {
...
+                if (engine.getHandshakeStatus() == HandshakeStatus.NEED_TASK) {
+                    final Runnable task = engine.getDelegatedTask();
+                    // We only call the method when NEED_TASK is returned so we must have a task!
+                    assert task != null;
+                    delegatedTaskExecutor.execute(new SslTask(task, inUnwrap));
+                } else {
+                    if (ctx.executor().inEventLoop()) {
+                        resumeOnEventExecutor();
+                    } else {
+                        ctx.executor().execute(new Runnable() {
```
njhill: Could make this `Runnable` a constant to avoid allocation?

It would be kind of nice for there to be an `EventExecutor.runInEventLoop(Runnable)` for cases like this :)

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1447,65 +1484,224 @@ private static ByteBuffer toByteBuffer(ByteBuf out, int index, int len) {
...
+        private void resumeOnEventExecutor() {
+            assert ctx.executor().inEventLoop();
+
+            processTask = false;
+
+            try {
+                HandshakeStatus status = engine.getHandshakeStatus();
+                switch (status) {
+                    // There is another task that needs to be executed and offloaded to the delegatingTaskExecutor.
+                    case NEED_TASK:
```
njhill: This should never happen, right? Maybe better to treat as some kind of state error?

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1447,65 +1484,224 @@ private static ByteBuffer toByteBuffer(ByteBuf out, int index, int len) {
...
         } else {
-            final List<Runnable> tasks = new ArrayList<Runnable>(2);
-            for (;;) {
-                final Runnable task = engine.getDelegatedTask();
-                if (task == null) {
-                    break;
+            final Runnable task = engine.getDelegatedTask();
+            // We only call the method when NEED_TASK is returned so we must have a task!
+            assert task != null;
+            executeDelegatingTask(task, inUnwrap);
```
njhill: Could you avoid passing an explicit task to the executor here? I think a constant `Runnable` would suffice, i.e. no need to allocate `SslTask`s - call `engine.getDelegatedTask()` first thing in the `run()` method. `inUnwrap` could be a plain instance field.

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
```diff
@@ -129,7 +130,7 @@ protected final void onLookupComplete(ChannelHandlerContext ctx,
     protected void replaceHandler(ChannelHandlerContext ctx, String hostname, SslContext sslContext) throws Exception {
         SslHandler sslHandler = null;
         try {
-            sslHandler = sslContext.newHandler(ctx.alloc());
+            sslHandler = newSslHandler(sslContext, ctx.alloc());
```
johnou: shouldn't we be passing hostname through to the ssl handler?

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1447,65 +1484,224 @@ private static ByteBuffer toByteBuffer(ByteBuf out, int index, int len) {
...
+                        ctx.executor().execute(new Runnable() {
+                            @Override
+                            public void run() {
+                                resumeOnEventExecutor();
+                            }
+                        });
+                    }
                 }
+            } catch (final Throwable cause) {
+                handleException(cause);
```
johnou: to play it safe shouldn't we be catching RejectedExecutionException here too and setting processTask to false, or even inside handleException?

handler/src/main/java/io/netty/handler/ssl/SniHandler.java
normanmaurer: @johnou I dont think so as the hostname overload is only used for client-side. (also we did not before as well).

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1447,65 +1484,224 @@ private static ByteBuffer toByteBuffer(ByteBuf out, int index, int len) {
...
-                @Override
-                public void run() {
-                    try {
-                        for (Runnable task: tasks) {
-                            task.run();
+        /**
+         * Executed after the wrapped {@code task} was executed via {@code delegatedTaskExecutor} to resume work
+         * on the {@link EventExecutor}.
+         */
+        private void resumeOnEventExecutor() {
```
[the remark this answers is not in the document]
normanmaurer: @njhill we need to have it `SslTask` as we need to take `inUnwrap` into account.

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
normanmaurer: @njhill as this is only done as part of the handshake and also it depends on `inUnwrap` I see no point of doing so.

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1447,65 +1484,224 @@ private static ByteBuffer toByteBuffer(ByteBuf out, int index, int len) {
...
+     * more tasks to process.
      */
-    private void runDelegatedTasks() {
-        if (delegatedTaskExecutor == ImmediateExecutor.INSTANCE) {
+    private boolean runDelegatedTasks(boolean inUnwrap) {
+        if (delegatedTaskExecutor == ImmediateExecutor.INSTANCE || inEventLoop(delegatedTaskExecutor)) {
+            // We should run the task directly in the EventExecutor thread and not offload at all.
             for (;;) {
                 Runnable task = engine.getDelegatedTask();
                 if (task == null) {
```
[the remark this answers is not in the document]
normanmaurer: @njhill sure we can do this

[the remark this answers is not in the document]
normanmaurer: that said I think the code is cleaner atm and the extra null check is super cheap anyway. Let me keep it as it is.

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
normanmaurer: @njhill it could happen... there is nothing wrong with have another task to process after you run the current task.

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
normanmaurer: yeah let me just set it to false here.

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
normanmaurer: I think allocate `SslTask`s is cleaner and as this is only during handshake its fine.

normanmaurer:
@johnou fixed

normanmaurer:
@njhill I think I would keep the parallelization as it is for now .

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
njhill: @normanmaurer good point, probably better to avoid the extra fields on the handler since they only apply during handshake. Kind of tangential but do you think it could be worth grouping all the handshake-related fields (including the executor) into a separate object which can be discarded once the handshake is complete? Just an idea, maybe limited benefit :)

It would still be a simple change I think to remove the `task` field from `SslTask` and have it grab the task itself from the engine. This eleminates a synchronized method call from the event loop and the next task dispatch in the `run()` method could be changed to `delegatedTaskExecutor.execute(this)`. wdyt?

Also given the current logic I don't think this can happen, but in general `NEED_TASK` doesn't mean that `getDelegatedTask()` won't return null. It remains in that state until previously returned task(s) have finished running. Might it be possible to end up in this situation if the delegated executor rejects the task?

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
njhill: @normanmaurer I still don't quite see that.. won't it only reach here in the case all prior tasks have run followed by `getDelegatedTask` returning null and no other wrap/unwrap interactions having happened with the engine (since they happen on the event loop and guarded by `processTask` flag)?

normanmaurer: @njhill actually you are right I think it cant happen (at least with the current guard). That said I think there is nothing wrong with keep the code as it is as it not harms and if it ever changes all will "just work".

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
normanmaurer: thats a good point... maybe I should better remove the assert and just return if task == null to be safe in the future.

njhill:
> I think I would keep the parallelization as it is for now .

@normanmaurer makes sense... the suggestion was motivated by the `SSLEngine` docs which suggest consecutive tasks returned from `getDelegatedTask`can be run in parallel (potentially using multiple threads from provided executor), which would be neat if it could improve handshake latency... but as far as I can see the current JDK impl will never supply a new task until the prior one has finished running anyhow!

njhill:
@normanmaurer one more question... how about the case where the task execution is delayed indefinitely (or possibly cancelled) by the provided executor. Would things timeout/fail cleanly? `processTask` would never change from true for example.

normanmaurer:
@njhill yes it should as we have a handshake timeout. I will add a test tho

normanmaurer:
@njhill fixed a small bug and also added more tests to ensure we timeout a handshake even if a task never ran.

njhill:
@normanmaurer made a change here to illustrate what I was saying re having the tasks always pulled by the executed runnable https://github.com/netty/netty/commit/9c34affa470f73375d75d4b72d84f06eb06a3527... it's a bit simpler imo but please feel free to ignore! Also included a couple of other small things that I noticed while doing it, for your consideration:
- when there's more than one task to run just do it inline instead of resubmitting to the executor
- `safeExceptionCaught` used instead of `handleException` in `resumeOnEventExecutor` since it's only run on the event loop
- renamed "delegating" to "delegated" in a couple of places which I think is more correct
- in the failure case try to revert `processTask` to false on the event loop instead of the delegate executor thread

normanmaurer:
@njhill thanks will check. I thought about eunning multiple tasks in place but decided against it to allow other SslHandler to make progress (and so be more fair) as the Executor is shared. That said it may not be a big problem, so let me think about it.

normanmaurer:
@njhill PTAL again... I did apply most of your suggestions :) Thanks a lot

handler/src/main/java/io/netty/handler/ssl/SslHandler.java
```diff
@@ -1488,9 +1488,19 @@ private static boolean inEventLoop(Executor executor) {
         return executor instanceof EventExecutor && ((EventExecutor) executor).inEventLoop();
     }
 
+    private static void runAllDelegatingTasks(SSLEngine engine) {
```
njhill: nit: runAllDelegatingTasks -> runAllDelegatedTasks?

johnou:
@normanmaurer right now the executor is only used for SSL tasks with no ByteBuf operations right? so if io.netty.allocator.useCacheForAllThreads is set to false there should be no overhead using a custom executor which doesn't employ FastThreadLocalThread

normanmaurer:
@johnou right...

johnou:
@normanmaurer cool, I finally got some time this sprint to integrate the offload changes into our application - i'll let you know how it goes.

normanmaurer:
@johnou ok cool...

johnou:
@njhill @normanmaurer I've configured a custom FJP and pass it into the pipeline but it seems the executor isn't used at all..? (using tcnative boringssl 2.0.25.Final)

```java
        pipeline.addLast(new SslHandler(sslContext.newEngine(channel.alloc()), false, sslForkJoinPool));
```

johnou:
@rkapsi have you had any luck with this? ^

normanmaurer:
@johnou did you also specify the needed system property ?

johnou:
@normanmaurer shit, thanks :)


static final boolean USE_TASKS =
--
94 | SystemPropertyUtil.getBoolean("io.netty.handler.ssl.openssl.useTasks", false);

normanmaurer:
@johnou no problem

rkapsi:
@normanmaurer @johnou finally had some time to give this a try but I'm unable to get anything to execute on the delegatedTaskExecutor? Is this only supported on BoringSSL (which I haven't tried)?

johnou:
@rkapsi I suppose you set `-Dio.netty.handler.ssl.openssl.useTasks=true`?

rkapsi:
@johnou apologies, I did everything and forgot to pass the Executor into SslHandler. :man_facepalming:

johnou:
@rkapsi out of interest what executor did you opt for?

rkapsi:
Currently using `DefaultEventExecutor`. Is there a preference?

johnou:
That's what I was wondering :) we currently pass in an instrumented ForkJoinPool..

```java
@Bean
public Executor sslForkJoinPool() {
    final ForkJoinPool.ForkJoinWorkerThreadFactory factory = pool ->
    {
        final ForkJoinWorkerThread worker = ForkJoinPool.defaultForkJoinWorkerThreadFactory.newThread(pool);
        worker.setName("SSLThread-" + worker.getPoolIndex());
        worker.setDaemon(true);
        return worker;
    };
    final ForkJoinPool forkJoinPool = new ForkJoinPool(Math.max(MINIMUM_PARALLELISM, Runtime.getRuntime().availableProcessors() * 2),
            factory, (t, e) -> logger.fatal("Uncaught exception", e), false);
    mBeanExporter = MBeanExporterUtil.exportMBeans(Collections.singletonMap("...:type=ForkJoinPool,name=SslExecutionPool",
                                                                            new ForkJoinPoolMBeanAdapter(forkJoinPool)));
    return forkJoinPool;
}
```

PR #8848, opened 2019-02-06
Fix h2c upgrade failure when multiple connection headers are present in upgrade request

### LINKED ISSUE #8846: When more than one connection header is present in h2c upgrade request, upgrade fails
### Expected behavior
$subject occurs, with this change https://github.com/netty/netty/pull/7824. (@bryce-anderson and @normanmaurer )

Now that more than one connection header is allowed with this https://github.com/netty/netty/pull/7824,
 line https://github.com/netty/netty/blob/737519314153f9146eaf532cd4e945a621cf3fa5/codec-http/src/main/java/io/netty/handler/codec/http/HttpServerUpgradeHandler.java#L295 returns false and the upgrade fails since the connection header value returned from this is not the upgrade but some other connection header value. 

If we are allowing multiple connection headers, then in HttpServerUpgradeHandler I think we need to check  whether any of the connection header value is upgrade and then remove other connection headers. https://tools.ietf.org/html/rfc7540#section-8.1.2.2

### Steps to reproduce
Send an upgrade request with multiple conection headers to HTTP/2 server and the request hangs.

Example request:
GET / HTTP/1.1
Host: localhost:9000
connection: keep-alive
forwarded: by=127.0.0.1
content-length: 0
upgrade: h2c
HTTP2-Settings: AAEAABAAAAIAAAABAAN_____AAQAAP__AAUAAEAAAAYAACAA
connection: HTTP2-Settings,upgrade

### Netty version
From 4.1.33.Final (This issue  started to occur from From 4.1.23.Final onwards)

arukshani:
Motivation:

When more than one connection header is present in h2c upgrade request, upgrade fails. This is to fix that. 

Modification:
In HttpServerUpgradeHandler's upgrade() method, check whether any of the connection header value is upgrade, not just the first header value which might return a different value other than upgrade. 

Result:
Fixes #8846. 

With this PR, now when multiple connection headers are sent with the upgrade request, upgrade will not fail.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpServerUpgradeHandler.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpServerUpgradeHandler.java
@@ -294,6 +294,7 @@ public class HttpServerUpgradeHandler extends HttpObjectAggregator {
         for (CharSequence connectionHeaderValue : connectionHeaderValues) {
             concatenatedConnectionValue.append(connectionHeaderValue).append(COMMA);
         }
+        concatenatedConnectionValue.setLength(concatenatedConnectionValue.length() - 1);
 
         // Make sure the CONNECTION header contains UPGRADE as well as all protocol-specific headers.
         Collection<CharSequence> requiredHeaders = upgradeCodec.requiredUpgradeHeaders();
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpServerUpgradeHandler.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpServerUpgradeHandler.java
@@ -29,10 +29,8 @@ import static io.netty.handler.codec.http.HttpResponseStatus.SWITCHING_PROTOCOLS
 import static io.netty.handler.codec.http.HttpVersion.HTTP_1_1;
 import static io.netty.util.AsciiString.containsAllContentEqualsIgnoreCase;
 import static io.netty.util.AsciiString.containsContentEqualsIgnoreCase;
-import static io.netty.util.AsciiString.containsIgnoreCase;
 import static io.netty.util.internal.ObjectUtil.checkNotNull;
 import static io.netty.util.internal.StringUtil.COMMA;
-import static io.netty.util.internal.StringUtil.DOUBLE_QUOTE;
 
 /**
  * A server-side handler that receives HTTP requests and optionally performs a protocol switch if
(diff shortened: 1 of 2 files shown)


normanmaurer:
@netty-bot test this please

normanmaurer:
@arukshani can you please sign our ICLA and let me know once done:

https://netty.io/s/icla

arukshani:
@normanmaurer I have signed the license agreement.

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerUpgradeHandler.java
```diff
@@ -284,7 +284,21 @@ private boolean upgrade(final ChannelHandlerContext ctx, final FullHttpRequest r
...
+
+        if (connectionHeaderValues == null) {
+            return false;
+        }
+
+        CharSequence connectionHeader = null;
+
+        for (CharSequence connectionHeaderValue : connectionHeaderValues) {
+            if (containsIgnoreCase(connectionHeaderValue, HttpHeaderNames.UPGRADE)) {
+                connectionHeader = connectionHeaderValue;
```
ejona86: This requires that Upgrade exists in the same Connection header value as requiredHeaders. But each value could be put in its own Connection header. It would be better to join all the values together with a `,` and then splitHeader(). @normanmaurer, is there not a utility to do that?

normanmaurer: @ejona86 thats a good point... I think there is no such utility yet (or at least I could not remember and also could not find any). /cc @arukshani can you do the adjustment and also add a unit test for this ?

arukshani: @ejona86 and @normanmaurer I have done the necessary changes and added a unit test as well for that. Please review.

codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
```diff
@@ -155,6 +155,99 @@ public void upgrade() throws Exception {
...
+    public void upgradeWithMultipleConnectionHeaders() {
+        setUpServerChannel();
+
+        String upgradeString = "GET / HTTP/1.1\r\n" +
+                "Host: example.com\r\n" +
+                "Connection: keep-alive\r\n" +
+                "Connection: Upgrade, HTTP2-Settings\r\n" +
+                "Upgrade: h2c\r\n" +
+                "HTTP2-Settings: AAMAAABkAAQAAP__\r\n\r\n";
+        ByteBuf upgrade = Unpooled.copiedBuffer(upgradeString, CharsetUtil.US_ASCII);
```
normanmaurer: @arukshani seems like the test code is completely identical except the different `upgradeString`. Consider moving all the code into an extra method and just pass in the `upgradeString` from both of the test methods to remove all the code-duplication

arukshani: Done

codec-http/src/main/java/io/netty/handler/codec/http/HttpServerUpgradeHandler.java
```diff
@@ -284,16 +284,22 @@ private boolean upgrade(final ChannelHandlerContext ctx, final FullHttpRequest r
...
+
+        if (connectionHeaderValues == null) {
             return false;
         }
 
+        final StringBuilder concatenatedConnectionValue = new StringBuilder(connectionHeaderValues.size() * 10);
+        for (CharSequence connectionHeaderValue : connectionHeaderValues) {
+            concatenatedConnectionValue.append(connectionHeaderValue).append(COMMA);
+        }
+
```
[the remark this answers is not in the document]
arukshani: @normanmaurer Done

codec-http2/src/test/java/io/netty/handler/codec/http2/CleartextHttp2ServerUpgradeHandlerTest.java
```diff
@@ -115,44 +122,38 @@ public void upgrade() throws Exception {
...
+        validateClearTextUpgrade(upgradeString);
+    }
 
-        // Check that the preface was send (a.k.a the settings frame)
-        ByteBuf settingsBuffer = channel.readOutbound();
-        assertNotNull(settingsBuffer);
-        settingsBuffer.release();
+    @Test
+    public void requiredHeadersInSeparateConnectionHeaders() {
+        setUpServerChannel();
```
normanmaurer: nit: you could also move all the `setUpServerChannel()` calls

arukshani: Done

normanmaurer:
@netty-bot test this please

PR #8850, opened 2019-02-07
Also use java.util.Base64 in handler-proxy module

njhill:
Motivation

In netty 5 we can use `java.util.Base64` which is simpler/faster for some purposes than netty's own `ByteBuf`-based Base64 utilities. This covers a couple of additional places to the one changed in #8837.

Modifications

Modify `HttpProxyHandler` and `HttpProxyServer` classes to use `java.util.Base64` instead of
`io.netty.handler.codec.base64.Base64` (latter is a test one).

Result

More efficient base64 processing in proxy impl.

cc @doom369


normanmaurer:
@netty-bot test this please

normanmaurer:
@njhill thanks a lot!

PR #8852, opened 2019-02-07
DefaultChannelHandlerContext only need to point to the next inbound /…

normanmaurer:
… next outbound handler and not walk the chain all the time.

Motivation:

At the moment we use a doubly-linked-list in our DefaultChannelPipeline where each DefaultChannelHandlerContext has a prev and next pointer. This means whenever we want to handle an inbound event we will need walk the next pointers until we find an context thand handles inbound events and the same also for outbound events where we need to walk the prev pointers. This introduces some unnessary work and can be made faster by replacing our "doubly-linked-list" by having two linked-list like datastructures. One will be used for inbound and the other for outbound events. This reduce the work needed to find the next handler for either and inbound or outbound event.

This idea was provided by @njhill in https://github.com/netty/netty/pull/8816#issuecomment-459085723.

Modifications:

Change DefaultChannelPipeline to basically use two linked-list like datastructures (one for inbound and one for outbound) to reduce the cost of finding the next context that handles the event.

Result:

Less overhead when looking for the next context that can process the event.

### DIFF
--- a/transport/src/main/java/io/netty/channel/DefaultChannelHandlerContext.java
+++ b/transport/src/main/java/io/netty/channel/DefaultChannelHandlerContext.java
@@ -106,8 +106,10 @@ final class DefaultChannelHandlerContext extends DefaultAttributeMap
     // There is no need to make this volatile as at worse it will just create a few more instances then needed.
     private Tasks invokeTasks;
 
-    DefaultChannelHandlerContext next;
-    DefaultChannelHandlerContext prev;
+    // Points the next contexts that will handle inbound / outbound events.
+    private DefaultChannelHandlerContext nextInbound;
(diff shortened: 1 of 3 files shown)


normanmaurer:
@njhill this implements your suggestion...

normanmaurer:
@netty-bot test this please

normanmaurer:
And again this also goes into the category of "I am not sure the gain justifies the complexity".

normanmaurer:
must likely it does not justify it...

normanmaurer:
alright let me close this... it was a nice experiment but the gain is just not big enough to justify this 🚀

carl-mastrangelo:
I did read this, but it was so complex I could not comment on it.  Closing is probably best for now.

PR #8744, opened 2019-01-22
Add logs for allocating large bytebuf in PooledByteBufAllocator

liupc:
Motivation:

Print logs for allocating large bytebuf in PooledByteBufAllocator, because many applications may care about the memory consumption, and want's to know more info about the memory allocation in case of OOM(for instance, Direct Memory OOM).

Modification:

This PR added logs when allocating large bytebuf

Result:

Fixes #8743 . 

If there is no issue then describe the changes introduced by this PR.

PR #8858, opened 2019-02-11
Add logs for allocating large bytebuf in PooledByteBufAllocator

### LINKED ISSUE #8743: Add logs for allocating large bytebuf in PooledByteBufAllocator
### Expected behavior
Print logs for allocating large bytebuf in PooledByteBufAllocator, because many applications may care about the memory consumption, and want's to know more info about the memory allocation in case of OOM(for instance, Direct Memory OOM). 
### Actual behavior
No logs
### Steps to reproduce
NA
```
  protected ByteBuf newDirectBuffer(int initialCapacity, int maxCapacity) {
        PoolThreadCache cache = threadCache.get();
        PoolArena<ByteBuffer> directArena = cache.directArena;

        final ByteBuf buf;
        if (directArena != null) {
            buf = directArena.allocate(cache, initialCapacity, maxCapacity);
        } else {
            buf = PlatformDependent.hasUnsafe() ?
                    UnsafeByteBufUtil.newUnsafeDirectByteBuf(this, initialCapacity, maxCapacity) :
                    new UnpooledDirectByteBuf(this, initialCapacity, maxCapacity);
        }

        return toLeakAwareBuffer(buf);
    }
```
### Netty version
4.1.30

NA
NA

liupc:
Motivation:

Print logs for allocating large bytebuf in PooledByteBufAllocator, because many applications may care about the memory consumption, and want's to know more info about the memory allocation in case of OOM(for instance, Direct Memory OOM).

Modification:

This PR added logs when allocating large bytebuf

Result:

Fixes #8743 .

NOTE: This is a reopen PR for https://github.com/netty/netty/pull/8744, when I put this PR to spark community as @normanmaurer 's suggestion, I got the following feedback:

> Good log is essential for all analysis. However, this is not a Spark specific issue. This is a little overkill if all `Netty`-based projects do this. Could you contribute to `Netty` project itself? Inside `Netty` project, it is one liner for each functions, isn't it?

> * https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java#L321

Ref link: https://github.com/apache/spark/pull/23610

I do agree with the point of spark community, this is a general need for analyzing memory issue, so we can not just let all up-level application/frameworks do this repeated work.

Can more reviewer to look at this PR? cc @normanmaurer

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java
+++ b/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java
@@ -307,6 +307,11 @@ public class PooledByteBufAllocator extends AbstractByteBufAllocator implements 
 
     @Override
     protected ByteBuf newHeapBuffer(int initialCapacity, int maxCapacity) {
+        if (initialCapacity > 1024 * 1024) {
+            logger.info("Try to allocate heap buffer with initialCapacity: " +
+                initialCapacity + " maxCapacity: " + maxCapacity + ", current usedHeapMemory: " +
+                metric.usedHeapMemory());
+        }
         PoolThreadCache cache = threadCache.get();
         PoolArena<byte[]> heapArena = cache.heapArena;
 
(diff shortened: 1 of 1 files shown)


liupc:
@normanmaurer  I think maybe we should have a discuss of whether this PR is useful rather than put it down to up-level applications/frameworks. If it does help, I think put it into `Netty` seems to be a good choice. Otherwise, we can just close it.

PR #8859, opened 2019-02-11
Support delegating task when using ReferenceCountedOpenSslEngine.

normanmaurer:
Motivation:

SSLEngine API has a notion of tasks that may be expensive and offload these to another thread. We did not support this when using our native implementation but can now for various operations during the handshake.

Modifications:

- Support offloading tasks during the handshake when using our native SSLEngine implementation
- Correctly handle the case when NEED_TASK is returned and nothing was consumed / produced yet

Result:

Be able to offload long running tasks from the EventLoop when using SslHandler with our native SSLEngine.

### DIFF
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
@@ -86,7 +86,8 @@ public abstract class ReferenceCountedOpenSslContext extends SslContext implemen
                                                       2048));
                 }
             });
-
+    private static final boolean USE_TASKS =
+            SystemPropertyUtil.getBoolean("io.netty.handler.ssl.openssl.useTasks", false);
     private static final Integer DH_KEY_LENGTH;
     private static final ResourceLeakDetector<ReferenceCountedOpenSslContext> leakDetector =
             ResourceLeakDetectorFactory.instance().newResourceLeakDetector(ReferenceCountedOpenSslContext.class);
@@ -342,6 +343,8 @@ public abstract class ReferenceCountedOpenSslContext extends SslContext implemen
             if (enableOcsp) {
                 SSLContext.enableOcsp(ctx, isClient());
             }
+
+            SSLContext.setUseTasks(ctx, USE_TASKS);
             success = true;
         } finally {
             if (!success) {
--- a/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
+++ b/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
@@ -76,6 +76,7 @@ import static io.netty.util.internal.ObjectUtil.checkNotNull;
 import static java.lang.Integer.MAX_VALUE;
 import static java.lang.Math.min;
 import static javax.net.ssl.SSLEngineResult.HandshakeStatus.FINISHED;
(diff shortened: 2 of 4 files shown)


normanmaurer:
This depends on https://github.com/netty/netty-tcnative/pull/430

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
```diff
@@ -342,6 +342,9 @@ public String run() {
             if (enableOcsp) {
                 SSLContext.enableOcsp(ctx, isClient());
             }
+
+            // TODO: Make the configurable.
```
johnou: default on / off?

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java
```diff
@@ -1293,10 +1313,29 @@ public final synchronized SSLEngineResult unwrap(ByteBuffer src, ByteBuffer[] ds
...
-        // Currently, we do not delegate SSL computation tasks
-        // TODO: in the future, possibly create tasks to do encrypt / decrypt async
-        return null;
+    public final synchronized Runnable getDelegatedTask() {
+        if (isDestroyed()) {
+            return null;
+        }
+        final Runnable task = SSL.getTask(ssl);
+        if (task == null) {
+            return task;
```
johnou: why not just return null?

johnou: (by returning task it looks like a typo with the equals check)

ejona86: +1 to returning null. I noticed it, but let it slide. But it slowed me down as well.

handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslContext.java
normanmaurer: I think I will make if off for now and allow to configure it with a System property... WDYT @ejona86 @johnou . Alternative I could allow to configure it via the `OpenSslContext`

ejona86: Is it just that we want to avoid unstable code, or is there a reason to believe it would have a performance cost?

normanmaurer: @ejona86 its more to allow to test it a bit more before make it default. I also still need to do some performance tests but I suspect the impact should be almost zero.

ejona86: Then system property sounds good. When we are comfortable we change the default. When it is "proven" we remove the option.

normanmaurer: Alright I introduced the system property "io.netty.handler.ssl.openssl.useTasks"

normanmaurer:
@netty-bot test this please

normanmaurer:
@netty-bot test this please

normanmaurer:
Let me merge this once we release netty-tcnative 2.0.21.Final

normanmaurer:
@netty-bot test this please

tomerd:
@netty-bot test this please

Issue #8862, opened 2019-02-12
Provide more information with the ClosedChannelException when fail operation with it


normanmaurer:
If we close a Channel because of a failed write we should store the "original" exception in the `ClosedChannelException` that will be used to fail further operations to make it easier to understand why the Channel was closed before.

This came up yesterday in a discussion with @Scottmitch @ejona86 @carl-mastrangelo .

ejona86:
+1

PR #8863, opened 2019-02-12
Include the original Exception that caused the Channel to be closed i…

normanmaurer:
…n the ClosedChannelException

Motivation:

To make it easier to understand why a Channel was closed previously and so why the operation failed with a ClosedChannelException we should include the original Exception.

Modifications:

- Store the original exception that lead to the closed Channel and include it in the ClosedChannelException that is used to fail the operation.
- Add unit test

Result:

Fixes https://github.com/netty/netty/issues/8862.

### DIFF
--- a/transport/src/test/java/io/netty/channel/AbstractChannelTest.java
+++ b/transport/src/test/java/io/netty/channel/AbstractChannelTest.java
@@ -15,8 +15,12 @@
  */
 package io.netty.channel;
 
+import java.io.IOException;
+import java.net.InetSocketAddress;
 import java.net.SocketAddress;
+import java.nio.channels.ClosedChannelException;
 
+import io.netty.util.NetUtil;
 import org.junit.Test;
 import org.mockito.invocation.InvocationOnMock;
 import org.mockito.stubbing.Answer;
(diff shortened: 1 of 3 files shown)


transport/src/test/java/io/netty/channel/AbstractChannelTest.java
```diff
@@ -82,6 +86,69 @@ public void ensureDefaultChannelId() {
...
+            @Override
+            public boolean isActive() {
+                return active;
+            }
+        };
+
+        EventLoop loop = new DefaultEventLoop();
+        try {
+            registerChannel(loop, channel);
+            channel.connect(new InetSocketAddress(NetUtil.LOCALHOST, 8888)).sync();
```
carl-mastrangelo: this will spuriously succeed on some machines won't it?

normanmaurer: It does not matter what we pass here as I implement `AbstractUnsafe.connect(...)` which will just notify the promise. Remember this is no socket etc.

normanmaurer:
@netty-bot test this please

PR #8864, opened 2019-02-13
Handle null "origin" header in "Old Hixie 75 handshake" as proper bad request.

artem-m:
Handle null origin header in Old Hixie 75 handshake as proper bad request.

Motivation:

Gracefully respond on bad client request.
We have a set of errors produced by Android 7.1.1/7.1.2 clients where both headers `HttpHeaderNames.SEC_WEBSOCKET_VERSION` and `HttpHeaderNames.ORIGIN` are not present. Absence of the first headers leads to WebSocketServerHandshaker00 be applied as a handshaker. However, null 2nd header causes
```
java.lang.NullPointerException: value
 io.netty.util.internal.ObjectUtil.checkNotNull(ObjectUtil.java:33)
 io.netty.handler.codec.DefaultHeaders.addObject(DefaultHeaders.java:327)
 io.netty.handler.codec.http.DefaultHttpHeaders.add(DefaultHttpHeaders.java:123)
 io.netty.handler.codec.http.websocketx.WebSocketServerHandshaker00.newHandshakeResponse(WebSocketServerHandshaker00.java:162)
```
Which causes connection close with unclear reason.

Modification:

Added null-check, and in case of null an appropriate WebSocketHandshakeException is thrown.

Result:

In case of null `HttpHeaderNames.ORIGIN` header a WebSocketHandshakeException is caught by  WebSocketServerProtocolHandler which sends a graceful `BAD_REQUEST`.

### DIFF
--- a/codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketServerHandshaker00Test.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketServerHandshaker00Test.java
@@ -32,7 +32,9 @@ import io.netty.util.CharsetUtil;
 import org.junit.Assert;
 import org.junit.Test;
 
-import static io.netty.handler.codec.http.HttpVersion.*;
+import static io.netty.handler.codec.http.HttpVersion.HTTP_1_1;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
 
 public class WebSocketServerHandshaker00Test {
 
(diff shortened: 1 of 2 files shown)


codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketServerHandshaker00Test.java
```diff
@@ -46,6 +48,32 @@ public void testPerformOpeningHandshakeSubProtocolNotSupported() {
...
+        WebSocketServerHandshaker00 handshaker00 = new WebSocketServerHandshaker00(
+            "ws://example.com/chat", "chat", Integer.MAX_VALUE);
+        try {
+            handshaker00.handshake(ch, req);
+            fail("Expecting WebSocketHandshakeException");
+        } catch (WebSocketHandshakeException e) {
+            assertEquals("Missing origin header, got only "
+                    + "[host, upgrade, connection, sec-websocket-key1, sec-websocket-protocol]",
+                e.getMessage());
+        }
```
normanmaurer: you need to also release the `req`.

Add something like:

```java
} finally {
    req.release();
}
```

normanmaurer:
@netty-bot test this please

normanmaurer:
@artem-m thanks a lot....

PR #8866, opened 2019-02-14
Support removal using values iterator.

michaelandrepearce:
Motivation:

As ActiveMQ project using netty, we want to make use of this class, unfortunately the iterator on values(), seems to not support remove method, even so the delegated iterator does. Currently we have to clone and modify this class locally albeit a one line change is needed, it would be ideal if netty could allow remove, then removing the need to maintain a clone.  

Modifications:

* remove throws UnsupportedOperationException, and instead call remove method on delegated iterator

Result

* allows users of the collection when iterating over values to call remove.

### DIFF
--- a/common/src/main/templates/io/netty/util/collection/KObjectHashMap.template
+++ b/common/src/main/templates/io/netty/util/collection/KObjectHashMap.template
@@ -236,7 +236,7 @@ public class @K@ObjectHashMap<V> implements @K@ObjectMap<V> {
 
                     @Override
                     public void remove() {
-                        throw new UnsupportedOperationException();
+                        iter.remove();
                     }
                 };
             }


normanmaurer:
@netty-bot Test this please

normanmaurer:
@michaelandrepearce there is a `KObjectHashMapTest.template` that you can modify.

normanmaurer:
@michaelandrepearce also can you please sign our icla: https://netty.io/s/icla

michaelandrepearce:
@normanmaurer i signed this back in 2017.
see.
https://github.com/netty/netty/pull/7077

normanmaurer:
@michaelandrepearce ah right... sorry did miss it. Then please just add the unit test and I will merge it :)

normanmaurer:
@michaelandrepearce ping

normanmaurer:
@michaelandrepearce I just merged this one and opened https://github.com/netty/netty/pull/8891 for a testcase.

Issue #8875, opened 2019-02-19
ChunkedWriteHandler does not close successful ChunkedInputs


lutovich:
### Expected behavior
`ChunkedWriteHandler` invokes `ChunkedInput#close()` after the input is fully consumed and `ChunkedInput#isEndOfInput()` returns `true.`

### Actual behavior
`ChunkedInput#close()` is not invoked when the input completes successfully. This can result in resource leaks when the input is based on an external resource, like a file handle.

### Steps to reproduce

Run the attached test.

### Minimal yet complete reproducer code (or URL to code)

```java
@Test
public void testCloseSuccessfulChunkedInput() throws Exception {
    final int totalChunks = 10;
    final AtomicBoolean closeInvoked = new AtomicBoolean();

    ChunkedInput<ByteBuf> input = new ChunkedInput<ByteBuf>() {
        int chunksProduced = 0;

        @Override
        public boolean isEndOfInput() throws Exception {
            return chunksProduced >= totalChunks;
        }

        @Override
        public void close() throws Exception {
            closeInvoked.set(true);
        }

        @Override
        public ByteBuf readChunk(ChannelHandlerContext ctx) throws Exception {
            return readChunk(ctx.alloc());
        }

        @Override
        public ByteBuf readChunk(ByteBufAllocator allocator) throws Exception {
            ByteBuf buf = allocator.buffer(4);
            buf.writeInt(chunksProduced);
            chunksProduced++;
            return buf;
        }

        @Override
        public long length() {
            return totalChunks;
        }

        @Override
        public long progress() {
            return chunksProduced;
        }
    };

    EmbeddedChannel ch = new EmbeddedChannel(new ChunkedWriteHandler());

    ch.writeAndFlush(input).sync();

    for (int i = 0; i < totalChunks; i++) {
        ByteBuf buf = ch.readOutbound();
        assertEquals(i, buf.readInt());
    }

    assertTrue(closeInvoked.get()); // this assertion fails
}
```

### Netty version

Problem exists only in `4.1.33.Final`.

### JVM version (e.g. `java -version`)

```
java version "1.8.0_201"
Java(TM) SE Runtime Environment (build 1.8.0_201-b09)
Java HotSpot(TM) 64-Bit Server VM (build 25.201-b09, mixed mode)
```

### OS version (e.g. `uname -a`)

```
[...]

PR #8876, opened 2019-02-19
Close consumed inputs in ChunkedWriteHandler

lutovich:
Motivation:

`ChunkedWriteHandler` needs to close both successful and failed `ChunkInput`s. It used to never close successful ones.

Modifications:

`ChunkedWriteHandler` always closes `ChunkInput`s after they are consumed.

Result:

Fixes https://github.com/netty/netty/issues/8875.

### DIFF
--- a/handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java
+++ b/handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java
@@ -166,25 +166,28 @@ public class ChunkedWriteHandler extends ChannelDuplexHandler {
             Object message = currentWrite.msg;
             if (message instanceof ChunkedInput) {
                 ChunkedInput<?> in = (ChunkedInput<?>) message;
+                boolean endOfInput;
+                long inputLength;
                 try {
-                    // read state of the input in local variables before closing it
-                    boolean endOfInput = in.isEndOfInput();
-                    long inputLength = in.length();
+                    endOfInput = in.isEndOfInput();
+                    inputLength = in.length();
                     closeInput(in);
-                    if (!endOfInput) {
-                        if (cause == null) {
-                            cause = new ClosedChannelException();
-                        }
-                        currentWrite.fail(cause);
-                    } else {
-                        currentWrite.success(inputLength);
-                    }
                 } catch (Exception e) {
                     closeInput(in);
                     currentWrite.fail(e);
                     if (logger.isWarnEnabled()) {
                         logger.warn(ChunkedInput.class.getSimpleName() + " failed", e);
                     }
+                    continue;
+                }
+
+                if (!endOfInput) {
+                    if (cause == null) {
+                        cause = new ClosedChannelException();
+                    }
+                    currentWrite.fail(cause);
+                } else {
+                    currentWrite.success(inputLength);
                 }
             } else {
                 if (cause == null) {
--- a/handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java
+++ b/handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java
@@ -280,12 +280,12 @@ public class ChunkedWriteHandler extends ChannelDuplexHandler {
                         @Override
                         public void operationComplete(ChannelFuture future) throws Exception {
                             if (!future.isSuccess()) {
-                                closeInput(chunks);
                                 currentWrite.fail(future.cause());
                             } else {
                                 currentWrite.progress(chunks.progress(), chunks.length());
                                 currentWrite.success(chunks.length());
                             }
+                            closeInput(chunks);
                         }
                     });
                 } else if (channel.isWritable()) {
--- a/handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java
+++ b/handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java
@@ -167,21 +167,24 @@ public class ChunkedWriteHandler extends ChannelDuplexHandler {
             if (message instanceof ChunkedInput) {
                 ChunkedInput<?> in = (ChunkedInput<?>) message;
                 try {
-                    if (!in.isEndOfInput()) {
+                    // read state of the input in local variables before closing it
+                    boolean endOfInput = in.isEndOfInput();
+                    long inputLength = in.length();
+                    closeInput(in);
+                    if (!endOfInput) {
                         if (cause == null) {
                             cause = new ClosedChannelException();
                         }
                         currentWrite.fail(cause);
                     } else {
-                        currentWrite.success(in.length());
+                        currentWrite.success(inputLength);
                     }
-                    closeInput(in);
                 } catch (Exception e) {
+                    closeInput(in);
                     currentWrite.fail(e);
                     if (logger.isWarnEnabled()) {
-                        logger.warn(ChunkedInput.class.getSimpleName() + ".isEndOfInput() failed", e);
+                        logger.warn(ChunkedInput.class.getSimpleName() + " failed", e);
                     }
-                    closeInput(in);
                 }
             } else {
                 if (cause == null) {
@@ -249,8 +252,8 @@ public class ChunkedWriteHandler extends ChannelDuplexHandler {
                         ReferenceCountUtil.release(message);
                     }
 
-                    currentWrite.fail(t);
                     closeInput(chunks);
+                    currentWrite.fail(t);
                     break;
                 }
 
@@ -280,12 +283,16 @@ public class ChunkedWriteHandler extends ChannelDuplexHandler {
                         @Override
                         public void operationComplete(ChannelFuture future) throws Exception {
                             if (!future.isSuccess()) {
+                                closeInput(chunks);
                                 currentWrite.fail(future.cause());
                             } else {
-                                currentWrite.progress(chunks.progress(), chunks.length());
-                                currentWrite.success(chunks.length());
+                                // read state of the input in local variables before closing it
+                                long inputProgress = chunks.progress();
+                                long inputLength = chunks.length();
+                                closeInput(chunks);
+                                currentWrite.progress(inputProgress, inputLength);
+                                currentWrite.success(inputLength);
                             }
-                            closeInput(chunks);
                         }
                     });
                 } else if (channel.isWritable()) {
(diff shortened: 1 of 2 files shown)


handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
```diff
@@ -433,6 +433,40 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
+
+        ch.writeAndFlush(input).sync();
+
+        for (int i = 0; i < chunks; i++) {
+            ByteBuf buf = ch.readOutbound();
+            assertEquals(i, buf.readInt());
+            buf.release();
+        }
+
+        assertTrue(input.closed);
```
normanmaurer: please also call `channel.finish()` and assert return value.

handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
```diff
@@ -433,6 +433,40 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
         assertEquals(1, chunks.get());
     }
 
+    @Test
+    public void testCloseSuccessfulChunkedInput() throws Exception {
+        int chunks = 10;
+        TestChunkedInput input = new TestChunkedInput(chunks);
+        EmbeddedChannel ch = new EmbeddedChannel(new ChunkedWriteHandler());
+
+        ch.writeAndFlush(input).sync();
```
normanmaurer: replace by `writeOutbound(...)` and assert return value

handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
```diff
@@ -433,6 +433,40 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
+
+    @Test
+    public void testCloseFailedChunkedInput() {
+        Exception error = new Exception("Unable to produce a chunk");
+        ThrowingChunkedInput input = new ThrowingChunkedInput(error);
+
+        EmbeddedChannel ch = new EmbeddedChannel(new ChunkedWriteHandler());
+
+        try {
+            ch.writeAndFlush(input).sync();
```
normanmaurer: replace by `writeOutbound(...)`

handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
```diff
@@ -433,6 +433,40 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
+            fail("Exception expected");
+        } catch (Exception e) {
+            assertEquals(error, e);
+        }
+
+        assertTrue(input.closed);
+    }
```
normanmaurer: please also call `channel.finish()` and assert return value.

handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
```diff
@@ -524,4 +558,88 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
 
         assertEquals(BYTES.length, read);
     }
+
+    private static class TestChunkedInput implements ChunkedInput<ByteBuf> {
```
normanmaurer: nit: final

handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
```diff
@@ -524,4 +558,88 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
+        final int chunksToProduce;
+
+        int chunksProduced;
```
normanmaurer: private

handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
```diff
@@ -524,4 +558,88 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
 
         assertEquals(BYTES.length, read);
     }
+
+    private static class TestChunkedInput implements ChunkedInput<ByteBuf> {
+        final int chunksToProduce;
```
normanmaurer: private

handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
```diff
@@ -524,4 +558,88 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
+        }
+
+        @Override
+        public long progress() {
+            return chunksProduced;
+        }
+    }
+
+    private static class ThrowingChunkedInput implements ChunkedInput<ByteBuf> {
+        final Exception error;
```
normanmaurer: private

handler/src/test/java/io/netty/handler/stream/ChunkedWriteHandlerTest.java
```diff
@@ -524,4 +558,88 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
+            return chunksToProduce;
+        }
+
+        @Override
+        public long progress() {
+            return chunksProduced;
+        }
+    }
+
+    private static class ThrowingChunkedInput implements ChunkedInput<ByteBuf> {
```
normanmaurer: final

handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java
```diff
@@ -285,6 +285,7 @@ public void operationComplete(ChannelFuture future) throws Exception {
                             } else {
                                 currentWrite.progress(chunks.progress(), chunks.length());
                                 currentWrite.success(chunks.length());
+                                closeInput(chunks);
```
kachayev: @lutovich Can we move this call out of the `if...else` block?

lutovich:
@normanmaurer, @kachayev thanks for reviews. Comments should now be addressed.

There is one small inconsistency in `ChunkedWriteHandler` somewhat related to this PR. `ChunkInput` is sometimes closed before and sometimes after resolving the `currentWrite` promise. I'm not sure if it's a real problem or not. Could be nice to always close the input first and then invoke all the listeners. What do you think?

normanmaurer:
Closing first sounds right

lutovich:
@normanmaurer PR updated to ensure input is always closed before completing the write promise

normanmaurer:
@netty-bot test this please

normanmaurer:
@kachayev PTAL again

handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java
normanmaurer: @lutovich seems like this was missed ?

lutovich: I changed it back on purpose in the second commit to make sure listener only invokes `ChunkedInput#close()` on failure. On success it also needs to read progress and length from the input. Does this make sense?

normanmaurer: ah ok yeah.

handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java
```diff
@@ -167,21 +167,24 @@ private void discard(Throwable cause) {
...
                             cause = new ClosedChannelException();
                         }
                         currentWrite.fail(cause);
                     } else {
-                        currentWrite.success(in.length());
+                        currentWrite.success(inputLength);
                     }
-                    closeInput(in);
                 } catch (Exception e) {
+                    closeInput(in);
```
kachayev: @lutovich We assume here that only `in.isEndOfInput` or `in.length` were able to fail, right? I would probably move this try/catch upper to be more "targeted" and specific. Otherwise, it's not obvious from the first glance why do we try to close the input once again here.

handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java
```diff
@@ -167,21 +167,24 @@ private void discard(Throwable cause) {
...
+                    long inputLength = in.length();
+                    closeInput(in);
+                    if (!endOfInput) {
                         if (cause == null) {
                             cause = new ClosedChannelException();
                         }
                         currentWrite.fail(cause);
                     } else {
-                        currentWrite.success(in.length());
+                        currentWrite.success(inputLength);
```
kachayev: @lutovich Not related to this specific change, but... I wonder if there's a chance that the write was already set to success/failure 🤔 Here's [a comment](https://github.com/netty/netty/blob/33c18469349d45b139506446f356e2b4f3af7086/handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java#L217-L225) from `doFlush`, not sure if the same is applicable here.

handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java
lutovich: @kachayev yeah, I think this try-catch is to handle failures in `#isEndOfInput()` and `#length()`. Added a commit with a tighter try-catch as you suggest. It makes code a bit longer and adds a `continue`. Not sure it's an improvement but please take a look :)

handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java
lutovich: @kachayev I'm not sure, to be honest. `currentWrite` is a `PendingWrite` and uses `trySuccess/tryFailure` to complete the underlying promise. It shouldn't be a problem to invoke `success/failure` more than one, right?

handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java
kachayev: We need Python's try/catch/**else** :) We can mark that first block failed in a separate variable, and test its value instead of continue.

handler/src/main/java/io/netty/handler/stream/ChunkedWriteHandler.java
kachayev: Right, should not be a problem. I'm just trying to understand if closing the input (potentially?) twice will cause any troubles.

normanmaurer: I think it should be ok... if not we can do another PR.

normanmaurer:
@netty-bot test this please

normanmaurer:
@kachayev ok to merge ?

kachayev:
@normanmaurer LGTM

normanmaurer:
@lutovich thanks a lot!

PR #8877, opened 2019-02-20
Fixed `invalid version format` caused by space in URI's parameter

jianwang-qau:
Motivation:

To fix IllegalArgumentException: `invalid version format` caused by space in URI's parameter.

Modification:

- Parse HttpVersion before URI.
- Add findNonWhitespaceRight method.
- Add findWhitespaceRight method.
- Add unit test.

Result:

Fixes IllegalArgumentException: `invalid version format` caused by space in URI's parameter.


normanmaurer:
@jianwang-qau please provide a unit test for this change.

normanmaurer:
Also general speaking spaces are not allowed in an URI and must be encoded as `%20`.

normanmaurer:
@jianwang-qau as said before this is not a valid URI... you need to replace ` ` by `%20`. Why you think this is valid ?

normanmaurer:
Let me close this as I think its not valid. Please re-open if you think I am wrong

Issue #8849, opened 2019-02-06
WARN  KQueueEventLoop - events[[1, 88, -1]]=[[1, 88, -1], {}] had no channel!


amizurov:
### Expected behavior
No warning ``2019-02-06 23:05:16.068 [KQueueEventLoopGroup-2-2] WARN  i.n.channel.kqueue.KQueueEventLoop - events[[1, 88, -1]]=[[1, 88, -1], {}] had no channel!``
### Actual behavior
Unexpected warning when client application kill -9 or reset peer
``2019-02-06 23:05:16.068 [KQueueEventLoopGroup-2-2] WARN  i.n.channel.kqueue.KQueueEventLoop - events[[1, 88, -1]]=[[1, 88, -1], {}] had no channel!``

Not received reset peer exception in ``exceptionCaught(ChannelHandlerContext ctx, Throwable cause)``
when interrupt client socket
### Steps to reproduce
Run snippet code below and kill client

### Minimal yet complete reproducer code (or URL to code)

public class ClientKQueue {

    public static void main(String[] args) throws Exception {
        Bootstrap clientBootstrap = new Bootstrap()
            .channel(KQueueSocketChannel.class).group(new KQueueEventLoopGroup())
            .handler(new ChannelInboundHandlerAdapter());
        Channel channel = clientBootstrap.connect("localhost", 8080).awaitUninterruptibly().channel();

        channel.closeFuture().awaitUninterruptibly();
    }
}

public class ServerKQueue {

    public static void main(String[] args) {
        ServerBootstrap serverBootstrap = new ServerBootstrap()
            .channel(KQueueServerSocketChannel.class).group(new KQueueEventLoopGroup())
            .childHandler(new ChannelInitializer() {

                @Override
                protected void initChannel(Channel ch) throws Exception {
                    ch.pipeline()
                        .addLast(new ChannelInboundHandlerAdapter() {

                            @Override
                            public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
                                System.out.println(cause.getMessage());
                            }
                        });
                }
            });

        Channel channel = serverBootstrap.bind(8080).awaitUninterruptibly().channel();
        channel.closeFuture().awaitUninterruptibly();
    }
}

### Netty version
4.1.33

### JVM version (e.g. `java -version`)
java version "1.8.0_181"
Java(TM) SE Runtime Environment (build 1.8.0_181-b13)
[...]

PR #8881, opened 2019-02-22
Don't deregister Channel as part of closing it when using native kque…

normanmaurer:
…ue transport

Motivation:

In https://github.com/netty/netty/pull/8665 we changed how we handle the registration of Channels to KQueue but missed to removed some code which would deregister the Channel before it actual closed the underlying socket. This could lead to have events triggered still while not have a mapping to the Channel anymore.

Modifications:

Remove deregister call during socket closure.

Result:

Fixes https://github.com/netty/netty/issues/8849.


normanmaurer:
This was reported to fix the issue of @kachayev and @amizurov so let me merge this as it is just code removal

PR #8882, opened 2019-02-22
Further reduce ByteBuf checkAccessible() overhead

njhill:
Motivation:

Some profiles of grpc-java indicate that overhead from `AbstractByteBuf.checkAccessible()` is quite significant due the relatively fine-grained way that `ByteBuf`s are accessed by the http2 hpack logic (5-10% of an entire end-to-end roundtrip benchmark!)

This PR takes the great improvement done in #8266 a bit further, simpliyfing the non-volatile/best-effort check. I did not dig deep enough to see why the existing `internalRefCnt()` logic was relatively slower
given that it does appear to get inlined. One observation though is that the prior optimization wasn't applied to derived buffers (slices etc.) which presumably was an oversight - this would be fairly simple to fix without the rest of this change (by overriding/delegating `internalRefCnt`) but that wouldn't help with the non-derived case.

Modifications:

- Replace `AbstractByteBuf.internalRefCnt()` with an internal (package-private) `ByteBuf.isAccessible()` method; the default impl still checks for nonzero `refCnt()`
- Set `maxCapacity` to -1 upon deallocation in `AbstractRefCountedByteBuf`, and add a final override of `isAccessible()` which checks this
- Appropriately forward `isAccessible()` invocations in `AbstractDerivedByteBuf`/`WrappedByteBuf`/`SwappedByteBuf` (also final overrides)
- Have `CompositeByteBuf` set `numComponents` to -1 upon deallocation so that the existing component index checks done on most methods also serve to check accessibility
- Add `ByteBufAccessBenchmark` which is an extension of `UnsafeByteBufBenchmark` (maybe latter could now be removed)

This change is similar to @ejona86's suggestion in #8266 (making use of `maxCapacity`), and I'd also thought of "invalidating" the reader/writer indices upon deallocation and having existing bounds checks do double-duty where possible. That could be a worthwhile addition irrespective of whether `checkAccessible` is otherwise changed to be false by default, but decided to keep this iteration minimally invasive. A complication is that bounds checking logic is in `AbstractByteBuf` but the index-invalidation-upon-deallocate logic must live in `AbstractReferenceCountedByteBuf` meaning that some further rework would be needed to avoid breaking the checks for any third-party impls that extend `AbstractByteBuf` directly. It's also the case that not all of the bounds checks involve these fields (absolute ones currently only use `capacity()` which is impl-specific),

Results:

Non-negligible speedups, especially for `CompositeByteBuf` which is twice as fast in some cases.

Before:
```
Benchmark                              (bufferType)   Mode  Cnt
Score         Error  Units
ByteBufAccessBenchmark.setGetLong            UNSAFE  thrpt   30 242844273.620 ± 1778416.963  ops/s
ByteBufAccessBenchmark.setGetLong      UNSAFE_SLICE  thrpt   30 184970819.249 ± 1990964.278  ops/s
ByteBufAccessBenchmark.setGetLong              HEAP  thrpt   30 242099517.501 ± 2661438.140  ops/s
ByteBufAccessBenchmark.setGetLong         COMPOSITE  thrpt   30  76961466.906 ± 1318414.178  ops/s
ByteBufAccessBenchmark.setGetLong               NIO  thrpt   30 279181674.131 ± 2162611.704  ops/s
```

After:
```
Benchmark                              (bufferType)   Mode  Cnt
Score          Error  Units
ByteBufAccessBenchmark.setGetLong            UNSAFE  thrpt   30 285754526.496 ±  3292540.505  ops/s
ByteBufAccessBenchmark.setGetLong      UNSAFE_SLICE  thrpt   30 209794441.861 ±  1646039.479  ops/s
ByteBufAccessBenchmark.setGetLong              HEAP  thrpt   30 283203453.736 ±  1987854.916  ops/s
ByteBufAccessBenchmark.setGetLong         COMPOSITE  thrpt   30 155437066.991 ±  1492219.965  ops/s
ByteBufAccessBenchmark.setGetLong               NIO  thrpt   30 277362327.090 ±  3070532.642  ops/s
```

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
@@ -94,11 +94,16 @@ public abstract class AbstractByteBuf extends ByteBuf {
 
     @Override
     public int maxCapacity() {
-        return maxCapacity;
+        return Math.abs(maxCapacity);
     }
 
     protected final void maxCapacity(int maxCapacity) {
         this.maxCapacity = maxCapacity;
+    }
+
+    // used only by AbstractRefCountedByteBuf
+    final int rawMaxCapacity() {
+        return maxCapacity;
     }
 
     @Override
@@ -280,16 +285,17 @@ public abstract class AbstractByteBuf extends ByteBuf {
         if (minWritableBytes <= writableBytes()) {
             return;
         }
+        int maxCap = maxCapacity();
         if (checkBounds) {
-            if (minWritableBytes > maxCapacity - writerIndex) {
+            if (minWritableBytes > maxCap - writerIndex) {
                 throw new IndexOutOfBoundsException(String.format(
                         "writerIndex(%d) + minWritableBytes(%d) exceeds maxCapacity(%d): %s",
-                        writerIndex, minWritableBytes, maxCapacity, this));
+                        writerIndex, minWritableBytes, maxCap, this));
             }
         }
 
         // Normalize the current capacity to the power of 2.
-        int newCapacity = alloc().calculateNewCapacity(writerIndex + minWritableBytes, maxCapacity);
+        int newCapacity = alloc().calculateNewCapacity(writerIndex + minWritableBytes, maxCap);
 
         // Adjust to the new capacity.
         capacity(newCapacity);
@@ -861,7 +867,7 @@ public abstract class AbstractByteBuf extends ByteBuf {
             return Unpooled.EMPTY_BUFFER;
         }
 
-        ByteBuf buf = alloc().buffer(length, maxCapacity);
+        ByteBuf buf = alloc().buffer(length, maxCapacity());
         buf.writeBytes(this, readerIndex, length);
         readerIndex += length;
         return buf;
@@ -1353,8 +1359,9 @@ public abstract class AbstractByteBuf extends ByteBuf {
             .append("(ridx: ").append(readerIndex)
             .append(", widx: ").append(writerIndex)
             .append(", cap: ").append(capacity());
-        if (maxCapacity != Integer.MAX_VALUE) {
-            buf.append('/').append(maxCapacity);
+        int maxCap = maxCapacity();
+        if (maxCap != Integer.MAX_VALUE) {
+            buf.append('/').append(maxCap);
         }
 
         ByteBuf unwrapped = unwrap();
@@ -1376,10 +1383,6 @@ public abstract class AbstractByteBuf extends ByteBuf {
 
     private static void checkRangeBounds(final int index, final int fieldLength, final int capacity) {
         if (isOutOfBounds(index, fieldLength, capacity)) {
-            if (capacity == -1) {
-                // special case: negative capacity implies buffer is deallocated
-                throw new IllegalReferenceCountException(0);
-            }
             throw new IndexOutOfBoundsException(String.format(
                     "index: %d, length: %d (expected: range(0, %d))", index, fieldLength, capacity));
         }
--- a/buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
@@ -1376,6 +1376,10 @@ public abstract class AbstractByteBuf extends ByteBuf {
 
     private static void checkRangeBounds(final int index, final int fieldLength, final int capacity) {
         if (isOutOfBounds(index, fieldLength, capacity)) {
+            if (capacity == -1) {
+                // special case: negative capacity implies buffer is deallocated
+                throw new IllegalReferenceCountException(0);
+            }
             throw new IndexOutOfBoundsException(String.format(
                     "index: %d, length: %d (expected: range(0, %d))", index, fieldLength, capacity));
         }
@@ -1437,17 +1441,9 @@ public abstract class AbstractByteBuf extends ByteBuf {
      * if the buffer was released before.
      */
     protected final void ensureAccessible() {
-        if (checkAccessible && internalRefCnt() == 0) {
+        if (checkAccessible && !isAccessible()) {
             throw new IllegalReferenceCountException(0);
         }
-    }
-
-    /**
-     * Returns the reference count that is used internally by {@link #ensureAccessible()} to try to guard
-     * against using the buffer after it was released (best-effort).
-     */
-    int internalRefCnt() {
-        return refCnt();
     }
 
     final void setIndex0(int readerIndex, int writerIndex) {
--- a/buffer/src/main/java/io/netty/buffer/AbstractReferenceCountedByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/AbstractReferenceCountedByteBuf.java
@@ -64,10 +64,8 @@ public abstract class AbstractReferenceCountedByteBuf extends AbstractByteBuf {
     }
 
     @Override
-    int internalRefCnt() {
-        // Try to do non-volatile read for performance as the ensureAccessible() is racy anyway and only provide
-        // a best-effort guard.
-        return realRefCnt(nonVolatileRawCnt());
+    final boolean isAccessible() {
+        return maxCapacity() != -1;
     }
 
     @Override
@@ -133,6 +131,7 @@ public abstract class AbstractReferenceCountedByteBuf extends AbstractByteBuf {
         int rawCnt = nonVolatileRawCnt(), realCnt = toLiveRealCnt(rawCnt, decrement);
         if (decrement == realCnt) {
             if (refCntUpdater.compareAndSet(this, rawCnt, 1)) {
+                maxCapacity(-1);
                 deallocate();
                 return true;
             }
@@ -155,6 +154,7 @@ public abstract class AbstractReferenceCountedByteBuf extends AbstractByteBuf {
             int rawCnt = refCntUpdater.get(this), realCnt = toLiveRealCnt(rawCnt, decrement);
             if (decrement == realCnt) {
                 if (refCntUpdater.compareAndSet(this, rawCnt, 1)) {
+                    maxCapacity(-1);
                     deallocate();
                     return true;
                 }
--- a/buffer/src/main/java/io/netty/buffer/AbstractReferenceCountedByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/AbstractReferenceCountedByteBuf.java
@@ -65,7 +65,7 @@ public abstract class AbstractReferenceCountedByteBuf extends AbstractByteBuf {
 
     @Override
     final boolean isAccessible() {
-        return maxCapacity() != -1;
+        return rawMaxCapacity() >= 0;
     }
 
     @Override
@@ -131,7 +131,7 @@ public abstract class AbstractReferenceCountedByteBuf extends AbstractByteBuf {
         int rawCnt = nonVolatileRawCnt(), realCnt = toLiveRealCnt(rawCnt, decrement);
         if (decrement == realCnt) {
             if (refCntUpdater.compareAndSet(this, rawCnt, 1)) {
-                maxCapacity(-1);
+                maxCapacity(-rawMaxCapacity());
                 deallocate();
                 return true;
             }
@@ -154,7 +154,7 @@ public abstract class AbstractReferenceCountedByteBuf extends AbstractByteBuf {
             int rawCnt = refCntUpdater.get(this), realCnt = toLiveRealCnt(rawCnt, decrement);
             if (decrement == realCnt) {
                 if (refCntUpdater.compareAndSet(this, rawCnt, 1)) {
-                    maxCapacity(-1);
+                    maxCapacity(-rawMaxCapacity());
                     deallocate();
                     return true;
                 }
--- a/buffer/src/main/java/io/netty/buffer/AbstractReferenceCountedByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/AbstractReferenceCountedByteBuf.java
@@ -64,8 +64,8 @@ public abstract class AbstractReferenceCountedByteBuf extends AbstractByteBuf {
     }
 
     @Override
-    final boolean isAccessible() {
-        return rawMaxCapacity() >= 0;
+    boolean isAccessible() {
+        return rawMaxCapacity() > 0 || (rawMaxCapacity() == 0 && (nonVolatileRawCnt() & 1) == 0);
     }
 
     @Override
@@ -131,7 +131,7 @@ public abstract class AbstractReferenceCountedByteBuf extends AbstractByteBuf {
         int rawCnt = nonVolatileRawCnt(), realCnt = toLiveRealCnt(rawCnt, decrement);
         if (decrement == realCnt) {
             if (refCntUpdater.compareAndSet(this, rawCnt, 1)) {
-                maxCapacity(-rawMaxCapacity());
+                maxCapacity(-maxCapacity());
                 deallocate();
                 return true;
             }
@@ -154,7 +154,7 @@ public abstract class AbstractReferenceCountedByteBuf extends AbstractByteBuf {
             int rawCnt = refCntUpdater.get(this), realCnt = toLiveRealCnt(rawCnt, decrement);
             if (decrement == realCnt) {
                 if (refCntUpdater.compareAndSet(this, rawCnt, 1)) {
-                    maxCapacity(-rawMaxCapacity());
+                    maxCapacity(-maxCapacity());
                     deallocate();
                     return true;
                 }
--- a/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
@@ -2123,6 +2123,11 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
     }
 
     @Override
+    boolean isAccessible() {
+        return componentCount != -1;
+    }
+
+    @Override
     public ByteBuf unwrap() {
         return null;
     }
--- a/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
@@ -470,8 +470,8 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
     }
 
     private void checkComponentIndex(int cIndex) {
+        ensureAccessible();
         if (cIndex < 0 || cIndex > componentCount) {
-            ensureAccessible();
             throw new IndexOutOfBoundsException(String.format(
                     "cIndex: %d (expected: >= 0 && <= numComponents(%d))",
                     cIndex, componentCount));
@@ -479,8 +479,8 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
     }
 
     private void checkComponentIndex(int cIndex, int numComponents) {
+        ensureAccessible();
         if (cIndex < 0 || cIndex + numComponents > componentCount) {
-            ensureAccessible();
             throw new IndexOutOfBoundsException(String.format(
                     "cIndex: %d, numComponents: %d " +
                     "(expected: cIndex >= 0 && cIndex + numComponents <= totalNumComponents(%d))",
@@ -1603,9 +1603,9 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
      * Consolidate the composed {@link ByteBuf}s
      */
     public CompositeByteBuf consolidate() {
+        ensureAccessible();
         final int numComponents = componentCount;
         if (numComponents <= 1) {
-            ensureAccessible();
             return this;
         }
 
@@ -2118,13 +2118,11 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
         for (int i = 0, size = componentCount; i < size; i++) {
             components[i].free();
         }
-        // allows us to omit explicit accessibility check in most cases
-        componentCount = -1;
     }
 
     @Override
     boolean isAccessible() {
-        return componentCount != -1;
+        return !freed;
     }
 
     @Override
--- a/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
@@ -300,7 +300,7 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
 
     @SuppressWarnings("deprecation")
     private Component newComponent(ByteBuf buf, int offset) {
-        if (checkAccessible && buf.refCnt() == 0) {
+        if (checkAccessible && !buf.isAccessible()) {
             throw new IllegalReferenceCountException(0);
         }
         int srcIndex = buf.readerIndex(), len = buf.readableBytes();
@@ -470,8 +470,8 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
     }
 
     private void checkComponentIndex(int cIndex) {
-        ensureAccessible();
         if (cIndex < 0 || cIndex > componentCount) {
+            ensureAccessible();
             throw new IndexOutOfBoundsException(String.format(
                     "cIndex: %d (expected: >= 0 && <= numComponents(%d))",
                     cIndex, componentCount));
@@ -479,8 +479,8 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
     }
 
     private void checkComponentIndex(int cIndex, int numComponents) {
-        ensureAccessible();
         if (cIndex < 0 || cIndex + numComponents > componentCount) {
+            ensureAccessible();
             throw new IndexOutOfBoundsException(String.format(
                     "cIndex: %d, numComponents: %d " +
                     "(expected: cIndex >= 0 && cIndex + numComponents <= totalNumComponents(%d))",
@@ -1603,9 +1603,9 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
      * Consolidate the composed {@link ByteBuf}s
      */
     public CompositeByteBuf consolidate() {
-        ensureAccessible();
         final int numComponents = componentCount;
         if (numComponents <= 1) {
+            ensureAccessible();
             return this;
         }
 
@@ -2118,6 +2118,8 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
         for (int i = 0, size = componentCount; i < size; i++) {
             components[i].free();
         }
+        // allows us to omit explicit accessibility check in most cases
+        componentCount = -1;
     }
 
     @Override
--- a/buffer/src/main/java/io/netty/buffer/WrappedCompositeByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/WrappedCompositeByteBuf.java
@@ -424,11 +424,6 @@ class WrappedCompositeByteBuf extends CompositeByteBuf {
     }
 
     @Override
-    int internalRefCnt() {
-        return wrapped.internalRefCnt();
-    }
-
-    @Override
     public ByteBuf duplicate() {
         return wrapped.duplicate();
     }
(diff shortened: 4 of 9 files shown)


njhill:
The knock-on improvement to HPACK benchmarks is also evident:

Before:
```
Benchmark                     (limitToAscii)  (sensitive)  (size)   Mode  Cnt        Score       Error  Units
HpackDecoderBenchmark.decode            true         true   SMALL  thrpt   30   370410.796 ±  4110.021  ops/s
HpackDecoderBenchmark.decode            true         true  MEDIUM  thrpt   30    49068.535 ±   406.485  ops/s
HpackDecoderBenchmark.decode            true        false   SMALL  thrpt   30   350153.682 ±  5251.114  ops/s
HpackDecoderBenchmark.decode            true        false  MEDIUM  thrpt   30    48853.463 ±   496.576  ops/s
HpackDecoderBenchmark.decode           false         true   SMALL  thrpt   30  1731930.970 ± 13894.478  ops/s
HpackDecoderBenchmark.decode           false         true  MEDIUM  thrpt   30   493797.242 ±  6731.768  ops/s
HpackDecoderBenchmark.decode           false        false   SMALL  thrpt   30  1573771.081 ± 13874.584  ops/s
HpackDecoderBenchmark.decode           false        false  MEDIUM  thrpt   30   428455.157 ±  6313.333  ops/s
```

After:
```
Benchmark                     (limitToAscii)  (sensitive)  (size)   Mode  Cnt        Score       Error  Units
HpackDecoderBenchmark.decode            true         true   SMALL  thrpt   30   378633.144 ± 10188.415  ops/s
HpackDecoderBenchmark.decode            true         true  MEDIUM  thrpt   30    58525.564 ±   329.438  ops/s
HpackDecoderBenchmark.decode            true        false   SMALL  thrpt   30   371989.033 ±  3195.292  ops/s
HpackDecoderBenchmark.decode            true        false  MEDIUM  thrpt   30    59141.990 ±   438.444  ops/s
HpackDecoderBenchmark.decode           false         true   SMALL  thrpt   30  1963143.732 ± 16181.923  ops/s
HpackDecoderBenchmark.decode           false         true  MEDIUM  thrpt   30   571016.787 ±  5614.790  ops/s
HpackDecoderBenchmark.decode           false        false   SMALL  thrpt   30  1891541.810 ± 19540.314  ops/s
HpackDecoderBenchmark.decode           false        false  MEDIUM  thrpt   30   533891.353 ±  4803.908  ops/s
```

buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java
```diff
@@ -1376,6 +1376,10 @@ protected final void checkIndex(int index, int fieldLength) {
 
     private static void checkRangeBounds(final int index, final int fieldLength, final int capacity) {
         if (isOutOfBounds(index, fieldLength, capacity)) {
+            if (capacity == -1) {
```
njhill: This check is required only for `PooledSlicedByteBuf` which performs a bounds check involving `maxCapacity` prior to accessibility check (and so would throw `IndexOutOfBoundsException` rather than `IllegalReferenceCountException` in the inaccessible case otherwise - failing some unit tests).

normanmaurer:
@njhill imho this change can not go into 4.1 as it would change semantics of `maxCapacity()` to not return the correct value anymore when deallocated.

normanmaurer:
@njhill as stated by you we should try to understand why this change is faster then what we had before as it should be the same speed imho... That said please provide a PR to fix the derived buffers :)

njhill:
@normanmaurer I thought that `maxCapacity()` would not really be meaningful for a deallocated buffer (in fact I thought all bets are off generally in this situation). For example, `PooledByteBuf`s reset this and the indicies when they are reused.

However I also can think of a trick to work around this, will push an update to see what you think. Worst case like you say I think there's some (possibly smaller) improvement here possible without relying on that field.

normanmaurer:
@njhill that's a good point that said I am still a bit afraid of the change here. I guess what you could do here is to make the value negative on release and then use `abs` when the user queries it. That said I still would love to understand why this is even needed as it theory the old behaviour should be good enough.

njhill:
@normanmaurer yep I already made the `abs` change, just want to benchmark before pushing it here. But I will open another PR regardless with related improvements that don't involve use of a different field. When I briefly tried this out before it appeared to give roughly half the gains.

normanmaurer:
@njhill any changes in the benchmark results after the abs change ? Also before pulling this in I would like to understand why this is needed and the current approach is slower. In theory I think it should not

buffer/src/main/java/io/netty/buffer/AbstractReferenceCountedByteBuf.java
```diff
@@ -133,6 +131,7 @@ private boolean release0(int decrement) {
         int rawCnt = nonVolatileRawCnt(), realCnt = toLiveRealCnt(rawCnt, decrement);
         if (decrement == realCnt) {
             if (refCntUpdater.compareAndSet(this, rawCnt, 1)) {
+                maxCapacity(-rawMaxCapacity());
```
normanmaurer: should we assert that this value will never be positive ?

buffer/src/main/java/io/netty/buffer/AbstractReferenceCountedByteBuf.java
```diff
@@ -155,6 +154,7 @@ private boolean retryRelease0(int decrement) {
             int rawCnt = refCntUpdater.get(this), realCnt = toLiveRealCnt(rawCnt, decrement);
             if (decrement == realCnt) {
                 if (refCntUpdater.compareAndSet(this, rawCnt, 1)) {
+                    maxCapacity(-rawMaxCapacity());
```
normanmaurer: should we assert that this value will never be positive ?

buffer/src/main/java/io/netty/buffer/WrappedCompositeByteBuf.java
```diff
@@ -1253,6 +1248,9 @@ public CompositeByteBuf discardSomeReadBytes() {
 
     @Override
     public final void deallocate() {
+        // this is to avoid overriding isAccessible(), so that it can be
+        // final in AbstractReferenceCountedByteBuf
+        maxCapacity(-rawMaxCapacity());
```
normanmaurer: should we assert that this value will never be positive ?

microbench/src/main/java/io/netty/buffer/ByteBufAccessBenchmark.java
```diff
@@ -0,0 +1,150 @@
...
+        bh.consume(buffer.readByte());
+        bh.consume(buffer.readByte());
+        bh.consume(buffer.readByte());
+        bh.consume(buffer.readByte());
+        bh.consume(buffer.readByte());
+    }
+
+    // doesn't seem to be possible to parameterize JVM args with JMH right now
+//  @Test
+//  public void runWithNoCheck() throws Exception {
```
normanmaurer: You can only be override:

```java
protected String[] jvmArgs() {
    return jvmArgs;
}
```

njhill: This applies to the entire run though, can't differ based on parameters.

njhill:
@normanmaurer benchmark with the `Math.abs` commit shows very little difference (possibly slightly slower). I agree this PR should still be considered tentative pending investigation/understanding of why it's necessary.

I've created a separate branch with some similar improvements to this one but avoiding use of `maxCapacity` - see cd26d4f394978da0ba58e04e3932ebb9e693cfc9. First benchmarks are similar to what I observed before: the gains are much smaller than those shown above for this PR. Some deeper analysis is needed I guess...

microbench/src/main/java/io/netty/buffer/ByteBufAccessBenchmark.java
```diff
@@ -0,0 +1,150 @@
...
+            @Override
+            ByteBuf newBuffer() {
+                return new NioFacade(ByteBuffer.allocateDirect(64));
+            }
+        };
+        abstract ByteBuf newBuffer();
+    }
+
+    @Param
+    public ByteBufType bufferType;
```
franz1981: I would add something to enable/disable check bounds checking too: https://github.com/netty/netty/pull/7521/files#diff-ed4e5dd4dd0a1138d1e5189d70a683c2R36

microbench/src/main/java/io/netty/buffer/ByteBufAccessBenchmark.java
franz1981: `System.setProperty("io.netty.buffer.checkAccessible", "false")` could be placed on `@Setup` before accessing any buffer class and it should allow setting it per fork.

microbench/src/main/java/io/netty/buffer/ByteBufAccessBenchmark.java
```diff
@@ -0,0 +1,150 @@
...
+
+    @Benchmark
+    public ByteBuf setLong() {
+        return buffer.setLong(0, 1);
+    }
+
+    @Benchmark
+    public void readEightBytes(final Blackhole bh) {
+        buffer.readerIndex(0).touch();
+        bh.consume(buffer.readByte());
```
franz1981: It is what you want to measure? 
`ByteBuf::getByte(index)` is another interesting case: it won't have a data dependency that will cause bounds check, because the positions accessed are fixed (I need to verify it on the assembly).
Using `BlackHole::consume` between each `readByte` call will force (at least) a bounds check each time, because `readerIndex` updates cannot be moved; each consme include a load acquire barrier AFAIK.

franz1981:
I've some numbers to share, 4.1:
```
Benchmark                              (bufferType)  (checkAccessible)  (checkBounds)   Mode  Cnt          Score         Error  Units
ByteBufAccessBenchmark.readEightBytes     COMPOSITE               true           true  thrpt   20   14655020.539 ±  425320.817  ops/s
ByteBufAccessBenchmark.readEightBytes     COMPOSITE               true          false  thrpt   20   17807254.275 ±  193070.149  ops/s
ByteBufAccessBenchmark.readEightBytes     COMPOSITE              false           true  thrpt   20   24776669.955 ±  269819.385  ops/s
ByteBufAccessBenchmark.readEightBytes     COMPOSITE              false          false  thrpt   20   28118611.519 ±  938274.799  ops/s
ByteBufAccessBenchmark.setGetLong         COMPOSITE               true           true  thrpt   20   72695175.133 ± 1029659.304  ops/s
ByteBufAccessBenchmark.setGetLong         COMPOSITE               true          false  thrpt   20   85866095.841 ±  955278.291  ops/s
ByteBufAccessBenchmark.setGetLong         COMPOSITE              false           true  thrpt   20   95962385.554 ±  356967.006  ops/s
ByteBufAccessBenchmark.setGetLong         COMPOSITE              false          false  thrpt   20  118367539.734 ± 7501704.810  ops/s
ByteBufAccessBenchmark.setLong            COMPOSITE               true           true  thrpt   20   74363920.616 ± 6862737.577  ops/s
ByteBufAccessBenchmark.setLong            COMPOSITE               true          false  thrpt   20   76531968.295 ± 1184693.315  ops/s
ByteBufAccessBenchmark.setLong            COMPOSITE              false           true  thrpt   20   80340118.090 ±  321379.275  ops/s
ByteBufAccessBenchmark.setLong            COMPOSITE              false          false  thrpt   20   90684143.719 ±  935820.121  ops/s
```

microbench/src/main/java/io/netty/buffer/ByteBufAccessBenchmark.java
franz1981: The numbers using the `readEightBytes` from above are actually very different (on the code of this PR).
Original `readEightBytes`:
```
Benchmark                              (bufferType)  (checkAccessible)  (checkBounds)   Mode  Cnt          Score          Error  Units
ByteBufAccessBenchmark.readEightBytes     COMPOSITE               true           true  thrpt   20   21234140.156 ±   540771.628  ops/s
ByteBufAccessBenchmark.readEightBytes     COMPOSITE               true          false  thrpt   20   25871522.124 ±   211593.779  ops/s
ByteBufAccessBenchmark.readEightBytes     COMPOSITE              false           true  thrpt   20   23694928.290 ±   288264.713  ops/s
ByteBufAccessBenchmark.readEightBytes     COMPOSITE              false          false  thrpt   20   27505818.147 ±   385209.348  ops/s
```
Modified `readEightBytes`:
```
Benchmark                              (bufferType)  (checkAccessible)  (checkBounds)   Mode  Cnt         Score         Error  Units
ByteBufAccessBenchmark.readEightBytes     COMPOSITE               true           true  thrpt   20  31660029.955 ± 1879564.146  ops/s
ByteBufAccessBenchmark.readEightBytes     COMPOSITE               true          false  thrpt   20  40709077.502 ± 2271395.159  ops/s
ByteBufAccessBenchmark.readEightBytes     COMPOSITE              false           true  thrpt   20  32216194.291 ± 1039418.921  ops/s
ByteBufAccessBenchmark.readEightBytes     COMPOSITE              false          false  thrpt   20  42907269.743 ±  819745.458  ops/s
```

microbench/src/main/java/io/netty/buffer/ByteBufAccessBenchmark.java
njhill: Good idea!

microbench/src/main/java/io/netty/buffer/ByteBufAccessBenchmark.java
njhill: Thanks alot @franz1981, you're right that benchmark was flawed, I had actually been wondering why it didn't show better improvement in the context of the other results.

I guess an alternative is to "combine" the byte values somehow prior to consuming/returning although it may have some cost.

franz1981: Exactly, that's why I have proposed the version with multiple consume after a batch of read byte: ideally it would mimic what could happen while reading bytes from the buffer to accumulate them in another (single-threaded) data-structure

njhill:
@franz1981 I like the of including the checks-disabled cases in the benchmark as a point of comparison, but for now maybe it simplifies things to exclude them.

What's of most interest imho is improvement we can get in the default setting case, and in particular understanding why checking a separate plain field is faster than the existing approach... i.e. some perfasm comparison between this PR and the other branch referenced above (which narrows the deltas).

I will have a go at this when I get a chance (have some other weekend commitments), but feel free to do so too in the meantime! Thanks again!

franz1981:
@njhill 
I've a strong feeling it is dependent by something related to CHA of the JIT instead: if you change `AbstractByteBuf` like this:
```
    protected int nonVolatileRawCnt() {
        throw new UnsupportedOperationException();
    }

    /**
     * Should be called by every method that tries to access the buffers content to check
     * if the buffer was released before.
     */
    protected final void ensureAccessible() {
        if (checkAccessible && AbstractReferenceCountedByteBuf.realRefCnt(nonVolatileRawCnt()) == 0) {
            throw new IllegalReferenceCountException(0);
        }
    }
```
making `AbstractReferenceCountedByteBuf.realRefCnt` public, you will see that the performance are similar to this PR. I hope soon to get some ASM to check :+1:

normanmaurer:
@franz1981 CHA ?

buffer/src/main/java/io/netty/buffer/AbstractReferenceCountedByteBuf.java
njhill: @normanmaurer I pushed a new commit which always updates with negative abs value here, hopefully that's just as good? Also I realized the `maxCapacity == 0` case wasn't handled, added that too. Still all tentative though pending the other investigation :)

njhill:
@normanmaurer class hierarchy analysis ... I had to look it up too :)

franz1981:
@normanmaurer Nick has preceded me re CHA :P
I've tried to get an ASM log to be used with JitWatch with no success...seems that I cannot print assembly on the JMH bench...

njhill:
@normanmaurer @franz1981 after some further experimentation I think I have got to the bottom of most (possibly all) of the discrepancy.

Focusing on the `UNSAFE` case first: the first thing which closed the gap quite a bit was avoiding "full" decoding of the non-volatile loaded ref count - this was unnecessary (my oversight from before) since we only care about the parity for this purpose. So the accessibility check becomes
```java
return (nonVolatileRawCount() & 1) == 0;
```
However, it seems the combination of a bitwise `&` operation and an int comparison is still noticeably more expensive than an isolated int comparison (`maxCapacity > 0`). This is unfortunate but I've mitigated for now by changing it to
```java
int rawCnt = nonVolatileRawCnt();
return rawCnt == 2 || rawCnt == 4 || rawCnt == 6 || rawCnt == 8 ||  (rawCnt & 1) == 0;
```
which gives a faster path to the assumed most common cases (refcounts 1-4) at expense of greater cost to larger refcount cases. This gets very close to the perf of this PR.

But while the `CompositeByteBuf` numbers improved slightly, there was still a stark difference to this PR (almost twice as slow). This turns out to be due to the stack growing too deep between adjacent external method invocations. It's more than 9 methods deep in the `setLong(...)` case so the crucial refCnt load at the bottom isn't getting inlined:
```
@ 10   io.netty.buffer.CompositeByteBuf::_setLong (90 bytes)   inline (hot)
  @ 2   io.netty.buffer.CompositeByteBuf::findComponent0 (33 bytes)   inline (hot)
  @ 27   io.netty.buffer.CompositeByteBuf$Component::idx (7 bytes)   inline (hot)
  @ 31   io.netty.buffer.AbstractByteBuf::setLong (15 bytes)   inline (hot)
   \-> TypeProfile (13959/13959 counts) = io/netty/buffer/UnpooledUnsafeDirectByteBuf
    @ 4   io.netty.buffer.AbstractByteBuf::checkIndex (11 bytes)   inline (hot)
      @ 1   io.netty.buffer.AbstractByteBuf::ensureAccessible (23 bytes)   inline (hot)
        @ 7   io.netty.buffer.AbstractReferenceCountedByteBuf::isAccessible (37 bytes)   inline (hot)
          @ 1   io.netty.buffer.AbstractReferenceCountedByteBuf::nonVolatileRawCnt (28 bytes)   inline (hot)
            @ 14   io.netty.util.internal.PlatformDependent::getInt (6 bytes)   inline (hot)
              @ 2   io.netty.util.internal.PlatformDependent0::getInt (9 bytes)   inlining too deep
```
I verified this by increasing `-XX:MaxInlineLevel`. So I guess we need to try and compress it a bit. There's a couple of obvious candidates, one being to manually inline the current `AbstractRefCountedByteBuf.nonVolatileRawCnt()` contents (it's only called from one place).

I've left this PR branch as-is for now and pushed the other changes described above to the [separate branch](https://github.com/netty/netty/compare/4.1...njhill:fast-ensure2) mentioned before, but haven't included any to reduce the stack depth for the composite buf case yet. I'm re-running the benchmarks and will post the results but probably won't be until tomorrow west coast time.

Also WDYT about the use of negative `componentCount` in `CompositeByteBuf` (similar to the `maxCapacity` idea)? I felt this should be even less of an issue since once a comp buf is deallocated it effectively gives up ownership of all its components so the count would kind of be zero at that point anyhow.

microbench/src/main/java/io/netty/buffer/ByteBufAccessBenchmark.java
nitsanw: If I may suggest replacing the 'read8' benchmark with a 'readBatch' which reads in a loop and "sums" the result somehow (e.g. `result += buf.readByte()`). The result can be sunk into a blackhole. The JVM has to read all the bytes to make the result and we want to evaluate the implications we have for batch reads IIUC. The loop will get unrolled which should end up being similar to the current case, and it would give us flexibility to test with different batch sizes.

normanmaurer:
@njhill I like the idea of using:

```
int rawCnt = nonVolatileRawCnt();
return rawCnt == 2 || rawCnt == 4 || rawCnt == 6 || rawCnt == 8 ||  (rawCnt & 1) == 0;
```
Just ensure to add a comment to explain why the hell we are doing this 🤣 

That said I am quite surprised that it makes such a big difference.

franz1981:
@njhill So It was due a mix of inlining + pure more ops? 
+100 for @normanmaurer advice re comments: next time someone will tell me that "the most performant code is just more readable" I will probably smile thinking about this eheh
There is an hidden (just received on the email) comment of @nitsanw about making a flexible readBatch method that deserve attention IMO

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
```diff
@@ -470,17 +470,17 @@ private void consolidateIfNeeded() {
     }
 
     private void checkComponentIndex(int cIndex) {
-        ensureAccessible();
         if (cIndex < 0 || cIndex > componentCount) {
+            ensureAccessible();
```
johnou: by moving it into the if aren't we changing the behaviour? and why check it only to throw an IOOBE?

normanmaurer: Yeah I think we should change this back @njhill

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
```diff
@@ -470,17 +470,17 @@ private void consolidateIfNeeded() {
...
             throw new IndexOutOfBoundsException(String.format(
                     "cIndex: %d (expected: >= 0 && <= numComponents(%d))",
                     cIndex, componentCount));
         }
     }
 
     private void checkComponentIndex(int cIndex, int numComponents) {
-        ensureAccessible();
         if (cIndex < 0 || cIndex + numComponents > componentCount) {
+            ensureAccessible();
```
johnou: same as above here.

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
njhill: @johnou @normanmaurer this was done in conjunction with setting `componentCount` to -1 upon deallocation - meaning the existing component index check catches both failure cases and the `ensureAccessible()` inside will ensure IRCE is thrown instead of IIOBE when appropriate.

However, benchmarking of this now that other changes have been made shows it provides negligible if any benefit, so I have reverted it in the latest commit (and just make use of the existing boolean `freed` field).

njhill:
> Just ensure to add a comment to explain why the hell we are doing this

@normanmaurer  Done! I wouldn't have left it without comments, it was just for feedback initially while iterating with the experiments. And I'm also surprised it makes such a big difference. I plan to do some other benchmarks around this, there's still more to understand here (see latest results i'll paste below).

njhill:
> So It was due a mix of inlining depth + too many ops?

@franz1981 the inlining depth yes for sure in composite buffer case, "too many ops" yes at least partially, but I'm not sure if it's the ops themselves and/or jit simplification that they may be interfering with... and the mitigation done may not be sufficient for some cases, or possibly there's still something else going on in addition (see below).

njhill:
I have pushed commits to both this PR branch and the other branch (which avoids using `maxCapacity`) based on the latest suggestions here. The changes in the other branch also include some minor manual inlining to flatten things sufficiently that the default max inline depth is no longer hit.

The results are quite interesting - on two of the benchmarks the non-maxCapacity branch is roughly the same speed, but this PR is still ~15-20% faster on the `readBatch` one. My suspicion is that the compiler is able to do some more aggressive simplification in the faster case that the more complex refCnt checks preclude in the other one. Will continue to analyse as I have time, and will re-run the HPACK bench too on these latest versions - would be interesting to see if the difference shows up in that closer-to-real-life case, which the "batch" benchmark was actually intended to mimic.

4.1 branch:
```
Benchmark   (bufferType)  (checkAccessible)  (checkBounds)   Mode  Cnt          Score          Error  Units
readBatch         UNSAFE               true           true  thrpt   30   84524972.863 ±   518338.811  ops/s
readBatch   UNSAFE_SLICE               true           true  thrpt   30   38608795.037 ±   298176.974  ops/s
readBatch           HEAP               true           true  thrpt   30   80003697.649 ±   974674.119  ops/s
readBatch      COMPOSITE               true           true  thrpt   30   18495554.788 ±   108075.023  ops/s
setGetLong        UNSAFE               true           true  thrpt   30  247069881.578 ± 10839162.593  ops/s
setGetLong  UNSAFE_SLICE               true           true  thrpt   30  196355905.206 ±  1802420.990  ops/s
setGetLong          HEAP               true           true  thrpt   30  245686644.713 ± 11769311.527  ops/s
setGetLong     COMPOSITE               true           true  thrpt   30   83170940.687 ±   657524.123  ops/s
setLong           UNSAFE               true           true  thrpt   30  278940253.918 ±  1807265.259  ops/s
setLong     UNSAFE_SLICE               true           true  thrpt   30  202556738.764 ± 11887973.563  ops/s
setLong             HEAP               true           true  thrpt   30  280045958.053 ±  2719583.400  ops/s
setLong        COMPOSITE               true           true  thrpt   30  121299806.002 ±  2155084.707  ops/s
```

This PR branch (using -ve `maxCapacity`):
```
Benchmark   (bufferType)  (checkAccessible)  (checkBounds)   Mode  Cnt          Score         Error  Units
readBatch         UNSAFE               true           true  thrpt   30  119541259.314 ± 1315960.350  ops/s
readBatch   UNSAFE_SLICE               true           true  thrpt   30   97742742.887 ±  593711.092  ops/s
readBatch           HEAP               true           true  thrpt   30  116364379.144 ± 2930052.587  ops/s
readBatch      COMPOSITE               true           true  thrpt   30   45086815.645 ±  928196.550  ops/s
setGetLong        UNSAFE               true           true  thrpt   30  297774971.597 ± 6596038.854  ops/s
setGetLong  UNSAFE_SLICE               true           true  thrpt   30  223155604.005 ± 2386659.044  ops/s
setGetLong          HEAP               true           true  thrpt   30  307324023.764 ± 1454568.509  ops/s
setGetLong     COMPOSITE               true           true  thrpt   30  162657739.658 ± 4006358.738  ops/s
setLong           UNSAFE               true           true  thrpt   30  298573314.566 ± 5078438.792  ops/s
setLong     UNSAFE_SLICE               true           true  thrpt   30  214502404.440 ± 4713830.912  ops/s
setLong             HEAP               true           true  thrpt   30  300545528.351 ± 2509813.277  ops/s
setLong        COMPOSITE               true           true  thrpt   30  157792835.151 ± 1398822.596  ops/s
```

[`fast-ensure2`](https://github.com/netty/netty/compare/4.1...njhill:fast-ensure2) branch (similar changes but avoiding use of `maxCapacity`):
```
Benchmark   (bufferType)  (checkAccessible)  (checkBounds)   Mode  Cnt          Score          Error  Units
readBatch         UNSAFE               true           true  thrpt   30  101641801.035 ±  3950050.059  ops/s
readBatch   UNSAFE_SLICE               true           true  thrpt   30   84395902.846 ±  4339579.057  ops/s
readBatch           HEAP               true           true  thrpt   30  100179060.207 ±  3222487.287  ops/s
readBatch      COMPOSITE               true           true  thrpt   30   42288494.472 ±   294919.633  ops/s
setGetLong        UNSAFE               true           true  thrpt   30  304530755.027 ±  6574163.899  ops/s
setGetLong  UNSAFE_SLICE               true           true  thrpt   30  212028547.645 ± 14277828.768  ops/s
setGetLong          HEAP               true           true  thrpt   30  309335422.609 ±  2272150.415  ops/s
setGetLong     COMPOSITE               true           true  thrpt   30  160383609.236 ±   966484.033  ops/s
setLong           UNSAFE               true           true  thrpt   30  298055969.747 ±  7437449.627  ops/s
setLong     UNSAFE_SLICE               true           true  thrpt   30  223784178.650 ±  9869750.095  ops/s
setLong             HEAP               true           true  thrpt   30  302543263.328 ±  8140104.706  ops/s
setLong        COMPOSITE               true           true  thrpt   30  157083673.285 ±  3528779.522  ops/s
```

franz1981:
@njhill interesting, I hope to have back my box able to profile with perf to go deeper into this :)

Just a suggestion, with so many results sometime is useful to add a simplified picture with the gains: I often use https://jmh.morethan.io with the JSON output from the netty benchs, that allow comparisons very easily!

normanmaurer:
@njhill thanks a lot... I think I would prefer to first apply the change which does NOT use the `maxCapacity` change and then revisit if we still need it or if we can do another change which makes it obsolete. WDYT ?

njhill:
@normanmaurer SGTM, I will probably do that in a new PR and close this one.

normanmaurer:
Sounds good

njhill:
Closing in favour of #8895.

By the way, the HPACK benchmarks for the new PR also look good (same faster speed as this one).

Issue #8868, opened 2019-02-15
DefaultFileRegion.transferTo with invalid count may cause hang 


georgeOsdDev:
### Expected behavior

`DefaultFileRegion.transferTo` will end successfully or throws Exception.

### Actual behavior

Thread will be hang and CPU usage increase to 100%

### Steps to reproduce

Initialize `DefaultFileRegion` with `count` larger than the actual `FileChannel ` size.

```
new DefaultFileRegion(raf.getChannel(), 0, fileLength)
```

### Minimal yet complete reproducer code (or URL to code)

https://github.com/netty/netty/blob/4.1/example/src/main/java/io/netty/example/http/file/HttpStaticFileServerHandler.java#L175-L193

Truncate files from another process between L175 and L192

Related issue.
https://github.com/xitrum-framework/xitrum/issues/675

### Netty version
4.1

### JVM version (e.g. `java -version`)

java version "1.8.0_202"
Java(TM) SE Runtime Environment (build 1.8.0_202-b08)
Java HotSpot(TM) 64-Bit Server VM (build 25.202-b08, mixed mode)

### OS version (e.g. `uname -a`)

Darwin OshidaIMac.local 17.7.0 Darwin Kernel Version 17.7.0: Wed Oct 10 23:06:14 PDT 2018; root:xnu-4570.71.13~1/RELEASE_X86_64 x86_64

normanmaurer:
@georgeOsdDev what you suggest we should do here ? You are saying we should check in the constructor that you do not construct something "invalid" ?

georgeOsdDev:
I'm not sure to check `count` in the constructor.
`sun.nio.ch.FileChannelImpl` does not throws Exception even if `count ` is larger than real file size.
https://docs.oracle.com/javase/8/docs/api/java/nio/channels/FileChannel.html#transferTo-long-long-java.nio.channels.WritableByteChannel-

---
When CPU usage increase, It seems `NioSocketChannel.doWrite` is called repeatedly. So it seems to be a matter of eventLoop handling
```
io.netty.channel.socket.nio.NioSocketChannel.doWrite(NioSocketChannel.java:375)
io.netty.channel.AbstractChannel$AbstractUnsafe.flush0(AbstractChannel.java:934)
io.netty.channel.nio.AbstractNioChannel$AbstractNioUnsafe.forceFlush(AbstractNioChannel.java:367)
io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:639)
[...]

PR #8885, opened 2019-02-25
DefaultFileRegion.transferTo with invalid count may cause busy-spin

normanmaurer:
Motivation:

`DefaultFileRegion.transferTo` will return 0 all the time when we request more data then the actual file size. This may result in a busy spin while processing the fileregion during writes.

Modifications:

- If we wrote 0 bytes check if the underlying file size is smaller then the requested count and if so throw an IOException
- Add DefaultFileRegionTest
- Add a test to the testsuite

Result:

Fixes https://github.com/netty/netty/issues/8868.


normanmaurer:
@netty-bot test this please

PR #8888, opened 2019-02-25
Add a way to control the treatment of the Http2 system frames in the …

sskrobotov:
…http2-to-http1 flow

**Motivation:**

`Http2StreamFrameToHttpObjectCodec` is used to convert objects in the pipeline from HTTP/2 classes to their HTTP1 counterparts so that the existing HTTP1 pipeline could be reused. There are some frames however that are not being converted. E.g. `Http2ResetFrame` or `Http2WindowUpdateFrame`. The existing behavior is for those frames to be sent through the `channelRead()` pipeline, which may be breaking the logic of pipelines designed to only support HTTP1 requests. This goal of this change is to allow clients to configure the treatment of the HTTP/2 "leftover" frames.

**Modification:**
A new `processSystemFrame()` method is added to the `Http2StreamFrameToHttpObjectCodec` class. The method is called for all `Http2Frame` objects that are not processed by the Codec itself.
The default implementation is to send the frame to the `channelRead()` pipeline.

**Result:**
A client now has a way to implement a custom way of handling HTTP/2 system frames.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
@@ -77,6 +77,15 @@ public class Http2StreamFrameToHttpObjectCodec extends MessageToMessageCodec<Htt
     }
 
     @Override
+    public void channelRead(final ChannelHandlerContext ctx, final Object msg) throws Exception {
+        if (!acceptInboundMessage(msg) && msg instanceof Http2Frame) {
+            processSystemFrame(ctx, (Http2Frame) msg);
+            return;
+        }
+        super.channelRead(ctx, msg);
+    }
+
+    @Override
     public boolean acceptInboundMessage(Object msg) throws Exception {
         return (msg instanceof Http2HeadersFrame) || (msg instanceof Http2DataFrame);
     }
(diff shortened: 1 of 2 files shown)


sskrobotov:
Another variant of this change is submitted as #8889.

normanmaurer:
@mosesn @bryce-anderson @ejona86 @carl-mastrangelo @trustin PTAL and let me know which one of the PRs you prefer

normanmaurer:
@netty-bot test this please

ejona86:
I don't understand the point of this. If the default is to pass the frames further through the pipeline, then couldn't you use another handler to throw them away?

I thought `Http2ResetFrame` or `Http2WindowUpdateFrame` aren't read on child channels... but it looks like that changed since my initial PR. `Http2WindowUpdateFrame` is supposed to get converted into Netty's flow control; that's right, that's still a TODO for the API. Even if Http2WindowUpdateFrames remain on the child channel for advanced use-cases, they need to be user events because you need to receive them even with AUTO_READ==false.

`Http2ResetFrame`, was just going to be a user event. You can see it still says that in the documentation of Http2MultiplexCodec:

> The child channel will be notified of user events that impact the stream, such as {`@link Http2GoAwayFrame`} and {`@link Http2ResetFrame`}, as soon as they occur.

normanmaurer:
@ejona86 I think this patch may still be useful as the Http2StreamFrameToHttpObjectCodec may also be used with the `Http2FrameCodec` only (so no multiplex codec at all). IMHO it's a bit "overkill" to add just another handler to the pipeline to filter out everything else. 

How we we handle the other frames when the `Http2MultiplexCodec` is in place is a different discussion imho. 

Let's close this anyway and continue in https://github.com/netty/netty/pull/8889 as this is the preferred implementation it seems

PR #8889, opened 2019-02-25
Add a way to control the treatment of the Http2 system frames in the …

sskrobotov:
…http2-to-http1 flow

**Motivation:**

`Http2StreamFrameToHttpObjectCodec` is used to convert objects in the pipeline from HTTP/2 classes to their HTTP1 counterparts so that the existing HTTP1 pipeline could be reused. There are some frames however that are not being converted. E.g. `Http2ResetFrame` or `Http2WindowUpdateFrame`. The existing behavior is for those frames to be sent through the `channelRead()` pipeline, which may be breaking the logic of pipelines designed to only support HTTP1 requests. This goal of this change is to allow clients to configure the treatment of the HTTP/2 "leftover" frames.

**Modification:**
A new `Http2SystemFramesTreatment` field is added to the `Http2StreamFrameToHttpObjectCodec`. `Http2SystemFramesTreatment` interface incapsulates the logic to be applied to all HTTP/2 frames that are not converted to the HTTP1 counterparts. A few pre-defined implementations are provided.

**Result:**
A client now has a way to chose one of the pre-defined handling strategies for HTTP/2 inbound system frames, or provide their own strategy implementation.

### DIFF
--- a/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
+++ b/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2StreamFrameToHttpObjectCodec.java
@@ -43,9 +43,12 @@ import io.netty.handler.codec.http.LastHttpContent;
 import io.netty.handler.ssl.SslHandler;
 import io.netty.util.Attribute;
 import io.netty.util.AttributeKey;
+import io.netty.util.ReferenceCountUtil;
 import io.netty.util.internal.UnstableApi;
 
 import java.util.List;
+
+import static io.netty.util.internal.ObjectUtil.checkNotNull;
 
 /**
  * This handler converts from {@link Http2StreamFrame} to {@link HttpObject},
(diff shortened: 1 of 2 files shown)


sskrobotov:
This PR is another variant of the changes suggested in https://github.com/netty/netty/pull/8888. In fact, I'd prefer this version over the #8888 one.

normanmaurer:
@mosesn @bryce-anderson @ejona86 @carl-mastrangelo @trustin PTAL and let me know which one of the PRs you prefer

mosesn:
I like this one.  Thanks for the PR!

normanmaurer:
@netty-bot test this please

trustin:
I prefer this one, but it'd be nice if 1) we have top-level classes instead 2) we do not use `enum` but usual final class with static finals and 2) rename treatment to handler (or something better).

normanmaurer:
@trustin frankly speaking I think I like to have it an inner interface as it really is only related to this codec. Also I like the enum for this stuff as it ensures that there is only a singleton of it.

trustin:
- I think using `enum` as a way to provide singleton functions is a kind of abusing what `enum` is intended for.
- An inner class is not a deal breaker and may be useful when used in a limited situation, but I prefer not using it when designing public API.
  - Why does a user have to do `import com.example.MyClass.MyOtherClass;` when he or she can do `import com.example.MyOtherClass`? Inner classes have their advantages, like making it easier to keep the related source code in a single file, but I see no gain (and even a little bit of loss) from API users' standpoint.
  - An example of a poor use of inner classes in my opinion is Guava's `Builder` inner classes. I find importing `com.google.common.collect.ImmutableList.Builder` results in fairly confusing code:

    ```java
    import com.google.common.collect.ImmutableList;
    import com.google.common.collect.ImmutableList.Builder;

    Builder<String> builder = ImmutableList.builder();
    builder.add(...);
    ImmutableList<String> list = builder.build();
    ```

    In the above example, the inner class `Builder` neither reduced the number of `import`s nor improved the readability of the code. As the code gains its complexity, like accepting `Builder` as a parameter, it becomes no longer clear to a reader whether `Builder` is from `ImmutableList` or `ImmutableSet`, etc. Also, when you end up having to use two different `Builder` classes in a single file, you have to un-import all `ImmutableList.Builder`, which is not a good experience as well.

  - To summarize:
    - using a verbose inner class name like it is in this pull request 1) increases the length of import statement and 2) incurs duplicate information in FQCN, e.g. 2 `Http`s in FQCN.
    - using a short inner class name like it is in Guava introduces ambiguity in user code as the code grows, i.e. `ImmutableListBuilder` was a better choice.

normanmaurer:
@trustin I get what you are saying but I still prefer the internal interface as it really is only useful for this use-case. 

I dont agree with that its an abuse of enums, it's even listed as good practise in "Effective Java". See also https://medium.com/@biratkirat/learning-effective-java-item-3-6552c87d8e54

trustin:
enum is for enumeration, so.. I'd think it's weird to use enums for other purposes. What's defined in this PR is obviously not even close to what a usual enum element does in enumeration. Why does a singleton function have a method like ordinal() and implement Comparable? It may be fine for private APIs but it never is for public ones.

normanmaurer:
@trustin I get your point... I am just saying it is considered as best-practise. That said I am not strong on it .

sskrobotov:
@trustin @normanmaurer Okay, how about:
- Move constants from enum to `public static final` fields of the `Http2StreamFrameToHttpObjectCodec` class.
- Keep inner interface.
- Rename it to `Http2SystemFramesHandler` or `Http2SystemFramesProcessor` (not to confuse with a `ChannelHandler`).
 
This way:
- When a predefined option is used, clients will only have to import `Http2StreamFrameToHttpObjectCodec`, or static import its fields.
- When a custom behavior is required:
  - Java8+ projects will be able to use lambda syntax, so the name of the interface will not make any difference.
  - The rest of the use cases will have to import the interface.

Moving this interface to `io.netty.handler.codec.http2` doesn't seem right in this specific case because all other interfaces in the package are directly related to the support of HTTP2 protocol, while this one is a very specific use case (i.e. if someone is implementing an HTTP/2-only pipeline, this interface will never be needed). A better way would probably be to introduce a new sub-package for all classes that support http2-http1 conversion, but doing this as a part of this change doesn't seem reasonable, as it will require a code change for all existing projects that use codec.

normanmaurer:
Works for me

ejona86:
The SEND_AS_MESSAGE vs SEND_AS_USER_EVENT is weird. We shouldn't be willy-nilly (in this case, blindly, without knowing what they are) be moving messages from one type to the other. That way leads madness as you can never predict what is supposed to happen. IGNORE and a PROPAGATE (if a user event keep as user event, if read keep as read) would be fine though.

As I mentioned in #8888, I think the two frames mentioned as being unhandled are simply mishandled by the HTTP/2 code today.

sskrobotov:
@ejona86 I agree, but my thinking here was that there are existing clients that currently have `Http2StreamFrameToHttpObjectCodec` in their pipelines, so switching to user events right away may break their logic. So the goal was to keep the existing by-default behavior (i.e. if clients just update the library version), but to provide an easy way to chose another option.
My understanding is that the whole HTTP/2 support is still "work in progress", so at some point there will be bigger changes to the API and that would be a good point for clients to revisit their pipelines. But it seems better to avoid changes that may affect existing logic between minor library releases.

ejona86:
> but my thinking here was that there are existing clients that currently have `Http2StreamFrameToHttpObjectCodec` in their pipelines, so switching to user events right away may break their logic

I was talking about changing the HTTP/2 code directly. Http2StreamFrameToHttpObjectCodec wouldn't be involved.

trustin:
@sskrobotov Thanks a lot!

normanmaurer:
@ejona86 let me duplicate my comment that I did in https://github.com/netty/netty/pull/8888 here:

>  I think this patch may still be useful as the Http2StreamFrameToHttpObjectCodec may also be used with the Http2FrameCodec only (so no multiplex codec at all). IMHO it's a bit "overkill" to add just another handler to the pipeline to filter out everything else.

How we we handle the other frames when the Http2MultiplexCodec is in place is a different discussion imho.

normanmaurer:
@sskrobotov closing this as https://github.com/netty/netty/pull/9290 is the proper fix ... Thanks a lot for reporting tho!

PR #8890, opened 2019-02-26
Message aware cumulators

belliottsmith:
Motivation:

Reduce the number of bytes that may need to be copied when parsing multiple messages in a class extending ByteToMessageDecoder, and permit bounding the size of the ByteBuf needed to stash unread bytes for future invocations.

Modification:

Modified the semantics and usage of Cumulator so that it may be "message aware" - knowing how much data is needed to be able to parse another message.  

Using this information, it may choose to only partially fill a cumulation with its input bytes, so long as an entire message may be read.  Ideally, only exactly the number of bytes necessary to read a message will be copied, after which the remaining bytes will be processed from the input ByteBuf.  

If any bytes remain unread at the end of channelRead, the Cumulator is invoked to decide how to persist them across invocations, so that a precisely sized buffer for the message may be allocated.

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
@@ -307,9 +307,9 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
                     cumulation = cumulator.cumulate(ctx.alloc(), cumulation, inProgress);
                     if (cumulation.readableBytes() == oldReadableBytes + newReadableBytes) {
                         inProgress = null;
-                        if (first) {
-                            break; // don't bother calling decode again if we've already passed it everything
-                        }
+                    }
+                    if (first) {
+                        break; // don't bother calling decode again if we've already passed it everything
                     }
 
                     boolean hadInProgress = inProgress != null;
--- a/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
@@ -155,8 +155,11 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
     };
 
     private static final byte STATE_INIT = 0;
-    private static final byte STATE_CALLING_CHILD_DECODE = 1;
-    private static final byte STATE_HANDLER_REMOVED_PENDING = 2;
+    private static final byte STATE_DEFER_HANDLER_REMOVED = 1;
+    private static final byte STATE_DEFER_INPUT_CLOSED = 2;
+    private static final byte STATE_HANDLER_REMOVED_PENDING = 4;
+    private static final byte STATE_INPUT_CLOSED_PENDING = 8;
+    private static final byte STATE_INPUT_CLOSED_PENDING_CALL_INACTIVE = 24;
 
     ByteBuf cumulation;
     private Cumulator cumulator = MERGE_CUMULATOR;
@@ -167,8 +170,11 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
      * A bitmask where the bits are defined as
      * <ul>
      *     <li>{@link #STATE_INIT}</li>
-     *     <li>{@link #STATE_CALLING_CHILD_DECODE}</li>
+     *     <li>{@link #STATE_DEFER_HANDLER_REMOVED}</li>
+     *     <li>{@link #STATE_DEFER_INPUT_CLOSED}</li>
      *     <li>{@link #STATE_HANDLER_REMOVED_PENDING}</li>
+     *     <li>{@link #STATE_INPUT_CLOSED_PENDING}</li>
+     *     <li>{@link #STATE_INPUT_CLOSED_PENDING_CALL_INACTIVE}</li>
      * </ul>
      */
     private byte decodeState = STATE_INIT;
@@ -197,6 +203,14 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
      */
     public boolean isSingleDecode() {
         return singleDecode;
+    }
+
+    private boolean hasState(byte checkIfHasState) {
+        return hasState(decodeState, checkIfHasState);
+    }
+
+    private static boolean hasState(byte decodeState, byte checkIfHasState) {
+        return checkIfHasState == (decodeState & checkIfHasState);
     }
 
     /**
@@ -243,8 +257,8 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
 
     @Override
     public final void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
-        if (decodeState == STATE_CALLING_CHILD_DECODE) {
-            decodeState = STATE_HANDLER_REMOVED_PENDING;
+        if (hasState(STATE_DEFER_HANDLER_REMOVED)) {
+            decodeState |= STATE_HANDLER_REMOVED_PENDING;
             return;
         }
         ByteBuf buf = cumulation;
@@ -278,14 +292,16 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
         if (msg instanceof ByteBuf) {
             CodecOutputList out = CodecOutputList.newInstance();
             ByteBuf data = (ByteBuf) msg;
+
+            byte state = STATE_DEFER_HANDLER_REMOVED | STATE_DEFER_INPUT_CLOSED;
             try {
 
-                decodeState = STATE_CALLING_CHILD_DECODE;
+                decodeState = state;
                 do {
                     first = cumulation == null;
                     if (first) {
                         callDecode(ctx, data, out);
-                        if (decodeState != STATE_CALLING_CHILD_DECODE || !data.isReadable()) {
+                        if (decodeState != state || !data.isReadable()) {
                             break;
                         }
                         cumulation = Unpooled.EMPTY_BUFFER;
@@ -306,7 +322,7 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
                     }
 
                     callDecode(ctx, cumulation, out);
-                    if (decodeState != STATE_CALLING_CHILD_DECODE) {
+                    if (decodeState != state) {
                         break;
                     }
 
@@ -322,11 +338,16 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
                 }
                 throw new DecoderException(e);
             } finally {
+                state = decodeState;
+                decodeState = STATE_INIT;
+
                 saveToCumulation(data);
-                if (decodeState == STATE_HANDLER_REMOVED_PENDING) {
+                if (hasState(state, STATE_HANDLER_REMOVED_PENDING)) {
                     handlerRemoved(ctx);
                 }
-                decodeState = STATE_INIT;
+                if (hasState(state, STATE_INPUT_CLOSED_PENDING)) {
+                    channelInputClosed(ctx, hasState(state, STATE_INPUT_CLOSED_PENDING_CALL_INACTIVE));
+                }
                 if (cumulation == null) {
                     numReads = 0;
                 } else if (!cumulation.isReadable()) {
@@ -433,6 +454,10 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
     }
 
     private void channelInputClosed(ChannelHandlerContext ctx, boolean callChannelInactive) throws Exception {
+        if (hasState(STATE_DEFER_INPUT_CLOSED)) {
+            decodeState |= callChannelInactive ? STATE_INPUT_CLOSED_PENDING_CALL_INACTIVE : STATE_INPUT_CLOSED_PENDING;
+            return;
+        }
         CodecOutputList out = CodecOutputList.newInstance();
         try {
             channelInputClosed(ctx, out);
@@ -563,17 +588,17 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
      */
     final void decodeRemovalReentryProtection(ChannelHandlerContext ctx, ByteBuf in, List<Object> out)
             throws Exception {
-        boolean handleRemove;
-        if (handleRemove = decodeState == STATE_INIT) {
-            decodeState = STATE_CALLING_CHILD_DECODE;
+        boolean handlePending;
+        if (handlePending = decodeState == STATE_INIT) {
+            decodeState = STATE_DEFER_HANDLER_REMOVED;
         }
         try {
             decode(ctx, in, out);
         } finally {
-            if (handleRemove) {
-                boolean removePending = decodeState == STATE_HANDLER_REMOVED_PENDING;
+            if (handlePending) {
+                boolean hasPendingRemove = hasState(STATE_HANDLER_REMOVED_PENDING);
                 decodeState = STATE_INIT;
-                if (removePending) {
+                if (hasPendingRemove) {
                     handlerRemoved(ctx);
                 }
             }
--- a/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
@@ -297,9 +297,10 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
                         cumulation = Unpooled.EMPTY_BUFFER;
                     }
 
-                    // we permit our cumulators to be smart, and guarantee exactly one message's worth of data is buffered
-                    // before we switch back to using the input buffer, so that if we are waiting for a message of known size
-                    // we can allocate an accurately sized buffer, and read any remaining messages without copying or composing
+                    // we permit our cumulators to be smart, and guarantee exactly one message's worth of data is
+                    // buffered before we switch back to using the input buffer, so that if we are waiting for a message
+                    // of known size we can allocate an accurately sized buffer, and read any remaining messages without
+                    // copying or composing
                     int oldReadableBytes = cumulation.readableBytes();
                     int newReadableBytes = data.readableBytes();
                     cumulation = cumulator.cumulate(ctx.alloc(), cumulation, data);
@@ -311,7 +312,15 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
                     }
 
                     callDecode(ctx, cumulation, out);
-                    if (!cumulation.isReadable()) {
+                    if (cumulation == null) {
+                        // this can only happen if we have been removed, so propagate any remaining bytes we may own
+                        if (data != null && data.isReadable()) {
+                            ctx.fireChannelRead(data);
+                            data = null;
+                            ctx.fireChannelReadComplete();
+                        }
+                        break;
+                    } else if (!cumulation.isReadable()) {
                         cumulation.release();
                         cumulation = null;
                     }
@@ -331,8 +340,9 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
                         data = null;
                     }
                 }
-                if (e instanceof DecoderException)
+                if (e instanceof DecoderException) {
                     throw e;
+                }
                 throw new DecoderException(e);
             } finally {
                 if (data != null) {
@@ -598,15 +608,22 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
          * The implementation is responsible to correctly handle the life-cycle of the given {@link ByteBuf}s and so
          * call {@link ByteBuf#release()} if a {@link ByteBuf} is fully consumed.
          *
-         * It is not required that all bytes from in make it to the result, so long as at least one message can be
-         * successfully decoded from the resulting buffer.  If this happens, cumulate will be called again after
-         * decode has been called on the result, successively, until we have attempted to decode the entirety of {@code in}.
-         * Once all of the bytes have been propagated, the lifetime of {@code in} is in the hands of the {@link Cumulator}.
+         * It is not required that all bytes from {@code in} make it to the result, so long as at least one message
+         * can be successfully decoded from the resulting buffer.  If this happens, cumulate will be called again after
+         * the resulting cumulation has been passed to {@link #callDecode}; this will happen repeatedly,
+         * until we have attempted to decode the entirety of {@code in} (or we can switch to {@code in} exclusively).
+         *
+         * Once all of the bytes have been propagated, the lifetime of {@code in} is in the hands of the
+         * {@link Cumulator}.
+         *
+         * The lifetime of {@code cumulation} is always handed to the control of this method; only the buffer that
+         * is returned, which may be {@code cumulation} itself, and {@code in} if its contents are propagated only
+         * partially, are returned to the ownership of the caller.
          *
          * @param cumulation if no readable bytes, we must return a buffer fully containing {@code in}
          * @param in the bytes from this buffer must be at least partially propagated to the output {@link ByteBuf};
-         *           it must be guaranteed that at least one message can be decoded from the result, or that
-         *           all of the input data is propagated.  If {@code cumulation} is empty, all of {@code in} should be propagated.
+         *           it must be guaranteed that at least one message can be decoded from the result, or that all of the
+         *           input data is propagated.  If {@code cumulation} is empty, all of {@code in} should be propagated.
          */
         ByteBuf cumulate(ByteBufAllocator alloc, ByteBuf cumulation, ByteBuf in);
     }
--- a/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
@@ -315,8 +315,9 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
                     if (cumulation == null) {
                         // this can only happen if we have been removed, so propagate any remaining bytes we may own
                         if (data != null && data.isReadable()) {
-                            ctx.fireChannelRead(data);
+                            ByteBuf propagate = data;
                             data = null;
+                            ctx.fireChannelRead(propagate);
                             ctx.fireChannelReadComplete();
                         }
                         break;
--- a/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
@@ -77,6 +77,10 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
     public static final Cumulator MERGE_CUMULATOR = new Cumulator() {
         @Override
         public ByteBuf cumulate(ByteBufAllocator alloc, ByteBuf cumulation, ByteBuf in) {
+            if (!cumulation.isReadable()) {
+                return in;
+            }
+
             try {
                 final ByteBuf buffer;
                 if (cumulation.writerIndex() > cumulation.maxCapacity() - in.readableBytes()
@@ -110,6 +114,10 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
     public static final Cumulator COMPOSITE_CUMULATOR = new Cumulator() {
         @Override
         public ByteBuf cumulate(ByteBufAllocator alloc, ByteBuf cumulation, ByteBuf in) {
+            if (!cumulation.isReadable()) {
+                return in;
+            }
+
             ByteBuf buffer;
             try {
                 if (cumulation.refCnt() > 1) {
@@ -267,20 +275,45 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
     public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
         if (msg instanceof ByteBuf) {
             CodecOutputList out = CodecOutputList.newInstance();
+            ByteBuf data = (ByteBuf) msg;
             try {
-                ByteBuf data = (ByteBuf) msg;
-                first = cumulation == null;
-                if (first) {
-                    cumulation = data;
-                } else {
+                do {
+                    first = cumulation == null;
+                    if (first) {
+                        callDecode(ctx, data, out);
+                        if (!data.isReadable())
+                            break;
+
+                        cumulation = Unpooled.EMPTY_BUFFER;
+                    }
+
+                    // we permit our cumulators to be smart, and guarantee exactly one message's worth of data is buffered
+                    // before we switch back to using the input buffer, so that if we are waiting for a message of known size
+                    // we can allocate an accurately sized buffer, and read any remaining messages without copying or composing
+                    int oldReadableBytes = cumulation.readableBytes();
+                    int newReadableBytes = data.readableBytes();
                     cumulation = cumulator.cumulate(ctx.alloc(), cumulation, data);
-                }
-                callDecode(ctx, cumulation, out);
-            } catch (DecoderException e) {
-                throw e;
+                    if (cumulation.readableBytes() == oldReadableBytes + newReadableBytes)
+                    {
+                        data = null;
+                        if (first)
+                            break; // don't bother calling decode again if we've already passed it everything
+                    }
+
+                    callDecode(ctx, data, out);
+                    if (!cumulation.isReadable()) {
+                        cumulation.release();
+                        cumulation = null;
+                    }
+                } while (data != null);
             } catch (Exception e) {
+                if (e instanceof DecoderException)
+                    throw e;
                 throw new DecoderException(e);
             } finally {
+                if (data != null) {
+                    data.release();
+                }
                 if (cumulation != null && !cumulation.isReadable()) {
                     numReads = 0;
                     cumulation.release();
@@ -540,6 +573,16 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
          * Cumulate the given {@link ByteBuf}s and return the {@link ByteBuf} that holds the cumulated bytes.
          * The implementation is responsible to correctly handle the life-cycle of the given {@link ByteBuf}s and so
          * call {@link ByteBuf#release()} if a {@link ByteBuf} is fully consumed.
+         *
+         * It is not required that all bytes from in make it to the result, so long as at least one message can be
+         * successfully decoded from the resulting buffer.  If this happens, cumulate will be called again after
+         * decode has been called on the result, successively, until we have attempted to decode the entirety of {@code in}.
+         * Once all of the bytes have been propagated, the lifetime of {@code in} is in the hands of the {@link Cumulator}.
+         *
+         * @param cumulation if no readable bytes, we must return a buffer fully containing {@code in}
+         * @param in the bytes from this buffer must be at least partially propagated to the output {@link ByteBuf};
+         *           it must be guaranteed that at least one message can be decoded from the result, or that
+         *           all of the input data is propagated.  If {@code cumulation} is empty, all of {@code in} should be propagated.
          */
         ByteBuf cumulate(ByteBufAllocator alloc, ByteBuf cumulation, ByteBuf in);
     }
--- a/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
@@ -281,8 +281,9 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
                     first = cumulation == null;
                     if (first) {
                         callDecode(ctx, data, out);
-                        if (!data.isReadable())
+                        if (!data.isReadable()) {
                             break;
+                        }
 
                         cumulation = Unpooled.EMPTY_BUFFER;
                     }
@@ -293,11 +294,11 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
                     int oldReadableBytes = cumulation.readableBytes();
                     int newReadableBytes = data.readableBytes();
                     cumulation = cumulator.cumulate(ctx.alloc(), cumulation, data);
-                    if (cumulation.readableBytes() == oldReadableBytes + newReadableBytes)
-                    {
+                    if (cumulation.readableBytes() == oldReadableBytes + newReadableBytes) {
                         data = null;
-                        if (first)
+                        if (first) {
                             break; // don't bother calling decode again if we've already passed it everything
+                        }
                     }
 
                     callDecode(ctx, data, out);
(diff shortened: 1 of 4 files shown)


codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
```diff
@@ -345,4 +344,177 @@ public CompositeByteBuf addComponent(boolean increaseWriterIndex, ByteBuf buffer
...
+                assertEquals(failed, throwAfterCount < firstMessageCount);
+                for (int i = 0 ; i < Math.min(throwAfterCount, firstMessageCount) ; ++i) {
+                    assertSame(decoded, channel.readInbound());
+                }
+                assertNull(channel.readInbound());
+
+                if (!failed) {
+                    try {
+                        channel.writeInbound(in2);
+                        assertTrue(false); // should never reach here
```
normanmaurer: replace with `fail();`

codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
```diff
@@ -345,4 +344,177 @@ public CompositeByteBuf addComponent(boolean increaseWriterIndex, ByteBuf buffer
...
+                        assertSame(in2, decoder.cumulation);
+                        assertEquals(3, decoder.cumulation.readableBytes());
+                    } else {
+                        assertNotSame(in2, decoder.cumulation);
+                        assertEquals(5, decoder.cumulation.readableBytes());
+                    }
+                }
+                decoder.cumulation.release();
+            }
+        }
```
normanmaurer: add `channel.finish();` and assert return value.

normanmaurer:
@netty-bot test this please

codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
```diff
@@ -316,6 +313,7 @@ public ByteBuf writeBytes(ByteBuf src) {
                 throw error;
             }
         };
+        cumulation.writeByte(1);
```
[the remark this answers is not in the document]
belliottsmith: The test assumes that MERGE_CUMULATOR will write some bytes and throw an exception, but the new behaviour (which is completely permitted by the API) is to return the right hand buffer if the left is empty.  To exercise the intended behaviour of the test, we just make sure the buffer is non-empty.

We could if you prefer change the behaviour of the default cumulators to only return the right hand side if the left hand buffer is both empty and non-writable, but this would still seem to make little sense for the COMPOSITE_CUMULATOR, which seems to never ideally want to aggregate an empty buffer.  The MERGE_CUMULATOR also probably benefits from defaulting to returning the only non-empty buffer to avoid a copy, although the case is muddier if the caller is hoping a writable but empty left hand side would be written to.

codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
```diff
@@ -336,6 +334,7 @@ public CompositeByteBuf addComponent(boolean increaseWriterIndex, ByteBuf buffer
...
```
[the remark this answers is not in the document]
belliottsmith: See MERGE_CUMULATOR explanation

codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
```diff
@@ -345,4 +344,177 @@ public CompositeByteBuf addComponent(boolean increaseWriterIndex, ByteBuf buffer
             assertEquals(0, in.refCnt());
         }
     }
+
+    private static abstract class PartialCumulationDecoder extends ByteToMessageDecoder
+    {
+        final int messageSize;
+        int cumulations;
+
+        private PartialCumulationDecoder(int messageSize) {
```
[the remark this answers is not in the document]
belliottsmith: sure

codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
```diff
@@ -345,4 +344,177 @@ public CompositeByteBuf addComponent(boolean increaseWriterIndex, ByteBuf buffer
             assertEquals(0, in.refCnt());
         }
     }
+
+    private static abstract class PartialCumulationDecoder extends ByteToMessageDecoder
+    {
```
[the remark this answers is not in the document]
belliottsmith: eep.  sorry for making you human lint :/ 

It's always harder than you think to switch code style.

codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
```diff
@@ -345,4 +344,177 @@ public CompositeByteBuf addComponent(boolean increaseWriterIndex, ByteBuf buffer
...
+            this.messageSize = messageSize;
+            setCumulator(new PartialCumulator());
+        }
+
+        private class PartialCumulator implements Cumulator {
+            @Override
+            public ByteBuf cumulate(ByteBufAllocator alloc, ByteBuf cumulation, ByteBuf in) {
+                ++cumulations;
+                if (!cumulation.isReadable()) {
+                    return in;
```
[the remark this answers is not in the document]
belliottsmith: It's actually unnecessary for our usage to call release in either implementation, but it may be that external users of the API pass pooled empty buffers as the cumulation parameter, so we want to be safe to that.

I'm absolutely happy to introduce a release call here for symmetry, though.

codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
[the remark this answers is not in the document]
normanmaurer: ok make sense... I just wanted to understand why this change was done.

codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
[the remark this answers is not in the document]
normanmaurer: I know the feeling... happens to me as well when I write patches to Cassandra ;)

codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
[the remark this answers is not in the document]
normanmaurer: I see.... I just wanted to understand if there is any "change" in behaviour related to who is responsible to release the buffer. From what I understand there is no change here really, correct ?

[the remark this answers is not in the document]
belliottsmith: Correct.  At least, not with normal usage, and in this case.  However, to be absolutely certain I've not been unclear elsewhere, IF the cumulator chooses not to fully copy the bytes (if it is message aware), then the buffer remains the responsibility of the caller (which is necessary, since it must consume the bytes).  This is only ever the case if you choose to use such a cumulator, though.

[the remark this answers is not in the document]
normanmaurer: got it.... Please add the release call here then as well. Just to make it consistent

normanmaurer:
@netty-bot test this please

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -267,20 +277,67 @@ protected void handlerRemoved0(ChannelHandlerContext ctx) throws Exception { }
...
-                } else {
+                do {
+                    first = cumulation == null;
+                    if (first) {
+                        // transfer data ownership to cumulation, so other methods have access to it
+                        cumulation = data;
+                        data = null;
+
+                        callDecode(ctx, cumulation, out);
+                        if (cumulation == null || !cumulation.isReadable()) {
```
normanmaurer: @belliottsmith I think you can remove the null check as it can never be null.

codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
```diff
@@ -15,14 +15,11 @@
  */
 package io.netty.handler.codec;
 
-import io.netty.buffer.ByteBuf;
-import io.netty.buffer.CompositeByteBuf;
-import io.netty.buffer.Unpooled;
-import io.netty.buffer.UnpooledByteBufAllocator;
-import io.netty.buffer.UnpooledHeapByteBuf;
+import io.netty.buffer.*;
```
normanmaurer: @belliottsmith can you revert this ? We usually do not wildcard imports.

codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
```diff
@@ -345,4 +344,176 @@ public CompositeByteBuf addComponent(boolean increaseWriterIndex, ByteBuf buffer
...
+                assertEquals(0, in1.refCnt());
+                assertEquals(2 + unreadFinalBytes, decoder.cumulations);
+                if (unreadFinalBytes > 0) {
+                    assertSame(in2, decoder.cumulation);
+                    assertEquals(1, decoder.cumulation.readableBytes());
+                    assertFalse(channel.finish());
+                } else {
+                    assertEquals(0, in2.refCnt());
+                    assertNull(decoder.cumulation);
+                    assertFalse(channel.finish());
```
normanmaurer: nit: you can remove this out of the else block and remove it from the if block.

codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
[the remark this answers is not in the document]
normanmaurer: can you maybe add a comment ?

normanmaurer:
@belliottsmith also please fix the check style errors:

```
12:47:55 [INFO] Starting audit...
12:47:55 [ERROR] /code/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java:300: Line is longer than 120 characters (found 122). [LineLength]
12:47:55 [ERROR] /code/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java:301: Line is longer than 120 characters (found 125). [LineLength]
12:47:55 [ERROR] /code/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java:302: Line is longer than 120 characters (found 127). [LineLength]
12:47:55 [ERROR] /code/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java:334: 'if' construct must use '{}'s. [NeedBraces]
12:47:55 [ERROR] /code/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java:603: Line is longer than 120 characters (found 124). [LineLength]
12:47:55 [ERROR] /code/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java:604: Line is longer than 120 characters (found 123). [LineLength]
12:47:55 [ERROR] /code/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java:609: Line is longer than 120 characters (found 130). [LineLength]
12:47:55 [ERROR] /code/codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java:18: Using the '.*' form of import should be avoided - io.netty.buffer.*. [AvoidStarImport]
12:47:55 [ERROR] /code/codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java:348:12: 'abstract' modifier out of order with the JLS suggestions. [ModifierOrder]
12:47:55 [ERROR] /code/codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java:435:25: Variable 'count' explicitly initialized to '0' (default value for its type). [ExplicitInitialization]
12:47:55 [ERROR] /code/codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java:507: Line is longer than 120 characters (found 136). [LineLength]
12:47:55 Audit done.
```

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
belliottsmith: I think it can - if the handler is removed before the callDecode completes, this will release the cumulation

codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
belliottsmith: Yep, of course

codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java
belliottsmith: Agh, IDEs...

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
normanmaurer: Thats a good point. Please add a comment and also add a null check here: https://github.com/netty/netty/pull/8890/files#diff-150d8a877292b8812bb4e3d27fb00a7fR314

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -540,6 +597,16 @@ static ByteBuf expandCumulation(ByteBufAllocator alloc, ByteBuf cumulation, int
          * Cumulate the given {@link ByteBuf}s and return the {@link ByteBuf} that holds the cumulated bytes.
          * The implementation is responsible to correctly handle the life-cycle of the given {@link ByteBuf}s and so
          * call {@link ByteBuf#release()} if a {@link ByteBuf} is fully consumed.
+         *
+         * It is not required that all bytes from in make it to the result, so long as at least one message can be
+         * successfully decoded from the resulting buffer.  If this happens, cumulate will be called again after
+         * decode has been called on the result, successively, until we have attempted to decode the entirety of {@code in}.
+         * Once all of the bytes have been propagated, the lifetime of {@code in} is in the hands of the {@link Cumulator}.
```
normanmaurer: can we also add some sentence about the life-time of cumulation.

normanmaurer:
@carl-mastrangelo @ejona86 PTAL ...

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -302,6 +371,24 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
         }
     }
 
+    private void saveToCumulation(ByteBuf data)
+    {
```
normanmaurer: nit: move to the line above

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -267,21 +291,66 @@ protected void handlerRemoved0(ChannelHandlerContext ctx) throws Exception { }
...
+                    if (!cumulation.isReadable()) {
+                        cumulation.release();
+                        cumulation = null;
+                    }
+                } while (data != null);
+
             } catch (Exception e) {
+                if (e instanceof DecoderException) {
+                    throw e;
+                }
```
normanmaurer: nit: just make this a `catch (DecoderException e) { throw e }`

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -267,21 +291,66 @@ protected void handlerRemoved0(ChannelHandlerContext ctx) throws Exception { }
...
             } finally {
-                if (cumulation != null && !cumulation.isReadable()) {
+                state = decodeState;
+                decodeState = STATE_INIT;
+
+                saveToCumulation(data);
+                if (hasState(state, STATE_HANDLER_REMOVED_PENDING)) {
+                    handlerRemoved(ctx);
+                }
+                if (hasState(state, STATE_INPUT_CLOSED_PENDING)) {
```
normanmaurer: shouldn't we handle this state first before `handlerRemoved` ?

belliottsmith: Perhaps, thanks for the insight - I wasn't too sure about this.  I'll make that change.  I also have an alternative idea for on approach that might on aggregate be sinmpler that I'll try out tomorrow, and we can choose between the two approaches.

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -77,6 +77,11 @@
     public static final Cumulator MERGE_CUMULATOR = new Cumulator() {
         @Override
         public ByteBuf cumulate(ByteBufAllocator alloc, ByteBuf cumulation, ByteBuf in) {
+            if (!cumulation.isReadable()) {
```
njhill: might make sense to move this into `channelRead`?

belliottsmith:
I've pushed another commit, that reverts the last two commits and takes an alternative approach for simplification, that leaves all of the other class' method semantics unchanged.  I think this is clearer, as a result.  Let me know what you think; I'm comfortable with either approach.

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
belliottsmith: This patch specifically wants to permit the cumulator greater control over what buffer is saved to cumulation, in all circumstances.  e.g. in the case that a message aware cumulator knows that we are waiting for a small message, but the input buffer is very large, it would be wasteful to save the entire buffer.  Or, for instance, you may be able to have a single dedicated buffer for all instances of a handler on an eventLoop, avoiding any allocation/recycling in typical operation, only copying bytes when an incomplete message arrives - which if your messages are typically small means minimal wasted memory and copying.

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -149,6 +159,7 @@ public ByteBuf cumulate(ByteBufAllocator alloc, ByteBuf cumulation, ByteBuf in)
     private static final byte STATE_HANDLER_REMOVED_PENDING = 2;
 
     ByteBuf cumulation;
+    ByteBuf inProgress;
```
normanmaurer: private ?

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
normanmaurer: @belliottsmith ^^

normanmaurer:
@ejona86 @njhill @carl-mastrangelo @trustin can you have a look here as well ?

normanmaurer:
@netty-bot test this please

normanmaurer:
@belliottsmith Please fix check style:

```
15:17:49 [INFO] Starting audit...
15:17:49 [ERROR] /code/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java:283: an empty line after a short line that ends with { [RegexpMultiline]
15:17:49 [ERROR] /code/codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java:321: Line is longer than 120 characters (found 121). [LineLength]
15:17:49 [ERROR] /code/codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java:342: Line is longer than 120 characters (found 121). [LineLength]
15:17:49 [ERROR] /code/codec/src/test/java/io/netty/handler/codec/ByteToMessageDecoderTest.java:511: Line is longer than 120 characters (found 136). [LineLength]
```

belliottsmith:
I've addressed those nits, thanks

normanmaurer:
@netty-bot test this please

normanmaurer:
@belliottsmith seems like your change introduced a unit test failure :/ 

```
java.lang.AssertionError: Channel inbound write failed.
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at io.netty.handler.codec.http.HttpClientCodecTest.sendRequestAndReadResponse(HttpClientCodecTest.java:243)
	at io.netty.handler.codec.http.HttpClientCodecTest.testAfterConnect(HttpClientCodecTest.java:230)
	at io.netty.handler.codec.http.HttpClientCodecTest.testPassThroughAfterConnect(HttpClientCodecTest.java:211)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:365)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeWithRerun(JUnit4Provider.java:273)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:238)
	at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:159)
	at org.apache.maven.surefire.booter.ForkedBooter.invokeProviderInSameClassLoader(ForkedBooter.java:384)
	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:345)
	at org.apache.maven.surefire.booter.ForkedBooter.execute(ForkedBooter.java:126)
	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:418)
```

belliottsmith:
Well that answers that question then.  I had been mulling whether or not we had to worry about invoking `internalBuffer` during a call to `callDecode`, and this answers it.  Fortunately that made it easy to spot and fix.

normanmaurer:
@netty-bot test this please

njhill:
@belliottsmith I like the change in principle but maybe it would be good to provide a benchmark demonstrating the overhead of the current logic and the improvement from this change? That should also help to verify there isn't any negative perf impact to existing decoders using the existing cumulators.

Also I'm curious whether you tried out `COMPOSITE_CUMULATOR` with your decoder if byte copying / buffer alloc overhead is significant. I know there's a warning that it could make your decoder slower if it performs granular accesses of the buffer, but there have been significant improvements in that area recently, including one just merged and possibly others coming.

belliottsmith:
cumulator|fBHCFOM|messageSize|totalInputSize|Score|Error|Units
:-----:|:-----:|:-----:|:-----:|:-----:|:-----:|:-----:
MERGE|true|16|1024|1,730,366|106,155|ops/s
COMPOSITE|true  |16|1024|1,898,419|196,428|ops/s
MESSAGE_AWARE_MERGE|true|16|1024|1,814,073|178,437|ops/s
MERGE|false|16|1024|1,808,904|75,931|ops/s
COMPOSITE|false|16|1024|1,837,803|60,314|ops/s
MESSAGE_AWARE_MERGE|false|16|1024|1,765,103|34,605|ops/s
MERGE|true|512|1024|2,548,085|185,921|ops/s
COMPOSITE|true|512|1024|3,196,712|153,890|ops/s
MESSAGE_AWARE_MERGE|true|512|1024|2,774,270|75,361|ops/s
MERGE|false|512|1024|2,473,081|506,062|ops/s
COMPOSITE|false|512|1024|3,249,218|327,554|ops/s
MESSAGE_AWARE_MERGE|false|512|1024|2,778,329|149,588|ops/s
MERGE|true|16|65536|56,421|7,706|ops/s
COMPOSITE|true|16|65536|71,128|13,878|ops/s
MESSAGE_AWARE_MERGE|true|16|65536|88,008|4,281|ops/s
MERGE|false|16|65536|58,354|33,086|ops/s
COMPOSITE|false|16|65536|71,206|15,796|ops/s
MESSAGE_AWARE_MERGE|false|16|65536|88,027|1,868|ops/s
MERGE|true|512|65536|405,108|119,209|ops/s
COMPOSITE|true|512|65536|1,303,665|216,483|ops/s
MESSAGE_AWARE_MERGE|true|512|65536|1,308,209|152,046|ops/s
MERGE|false|512|65536|411,177|246,373|ops/s
COMPOSITE|false|512|65536|1,285,318|766,012|ops/s
MESSAGE_AWARE_MERGE|false|512|65536|1,307,545|292,429|ops/s

belliottsmith:
@njhill thanks for the suggestion; I've posted above a table of some short runs of the benchmark I have pushed to the PR on my laptop; if anybody has dedicated hardware they can throw at more authoritative runs that would be great.

The benchmark writes a single byte to the channel, then a large buffer of varying sizes, parsing messages of varying sizes.  The first input buffer is optionally (fBHCFOM) large enough to fit a single message.

As you can see, the MESSAGE_AWARE_MERGE manages to get the best of both worlds, never really being appreciably slower than either the COMPOSITE or MERGE best cases.  

MERGE is never faster (modulo fractions of error bars).

COMPOSITE wins only the 512 messageSize, 1024 totalInputSize (i.e. an input of 1 byte, input of 1023 bytes, then skipping 512 bytes twice to parse).  This is weighted in favour of COMPOSITE, because of the efficiency of parsing.  The test has the option of making parsing happen byte-by-byte to simulate a heavier parsing burden, and with this enabled it loses handily to MESSAGE_AWARE_MERGE.  I have not included the results of this for all options, as I don't have time to dedicate my laptop to the task.

FWIW, I have also written a message aware Cumulator for lz4 that we can include in this patch if we want, or I can open a separate PR for that at a later time.

microbench/src/main/java/io/netty/microbench/codec/ByteToMessageDecoderBenchmark.java
```diff
@@ -0,0 +1,158 @@
+/*
+ * Copyright 2014 The Netty Project
```
normanmaurer: nit: 2019

microbench/src/main/java/io/netty/microbench/codec/ByteToMessageDecoderBenchmark.java
```diff
@@ -0,0 +1,158 @@
...
+@State(Scope.Benchmark)
+@Warmup(iterations = 10)
+@Measurement(iterations = 20)
+public class ByteToMessageDecoderBenchmark extends AbstractMicrobenchmark {
+
+    public enum SelectCumulator {
+        MERGE, COMPOSITE, MESSAGE_AWARE_MERGE
+    }
+
+    public static class MessageDecoder extends ByteToMessageDecoder {
```
normanmaurer: nit: private

microbench/src/main/java/io/netty/microbench/codec/ByteToMessageDecoderBenchmark.java
```diff
@@ -0,0 +1,158 @@
...
+        final int messageSize;
+        final int skipSize;
+        public MessageDecoder(int messageSize, int skipSize, SelectCumulator cumulator) {
```
normanmaurer: nit: remove public

microbench/src/main/java/io/netty/microbench/codec/ByteToMessageDecoderBenchmark.java
```diff
@@ -0,0 +1,158 @@
...
+                case MERGE:
+                    setCumulator(MERGE_CUMULATOR);
+                    break;
+                case COMPOSITE:
+                    setCumulator(COMPOSITE_CUMULATOR);
+                    break;
+                case MESSAGE_AWARE_MERGE:
+                    setCumulator(new MessageAwareCumulator(messageSize));
+                    break;
+            }
```
normanmaurer: nit: add a default case

microbench/src/main/java/io/netty/microbench/codec/ByteToMessageDecoderBenchmark.java
```diff
@@ -0,0 +1,158 @@
...
+        protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) {
+            while (in.readableBytes() >= messageSize) {
+                for (int skip = messageSize; skip > 0 ; skip -= skipSize) {
+                    in.skipBytes(skipSize);
+                }
+            }
+        }
+    }
+
+    static class MessageAwareCumulator implements ByteToMessageDecoder.Cumulator {
```
normanmaurer: nit: private final

normanmaurer:
@netty-bot test this please

njhill:
Thanks @belliottsmith this is great! I will probably try out some other variations too... in particular pre-existing decoder and cumulator perf before and after the other changes to `ByteToMessageDecoder`.

njhill:
@belliottsmith @normanmaurer I've given this some more thought and am not sure whether opening up the `Cumulator` as an extension point is the best way to "solve" this. The `ByteToMessageDecoder` is a simple abstraction that provides a continuous "stream" of input data, readable at your own pace to produce messages. The cumulation aspect imho should be considered an internal impl detail of that abstraction. I see the fact that it's currently configurable as more of a toggle between the merge/composite options, which is a break in the abstraction to allow a "hint" about the kind of access pattern that will be used and have things optimized accordingly. I expect it's rare that anyone is actually using a custom impl and I wonder if we should even deprecate that capability in the next version.

Could we achieve the same thing by just adding a protected `setBytesRequired(int)` method, which can optionally be called by `decode(...)` impls to signal how many bytes they require for their next invocation? Then we can similarly minimize how much and how frequently cumulation needs to be done, and avoid calling `decode` at all when there isn't sufficient data. This should make it easier to retrofit existing decoders and for new ones to exploit - for example doing so for the fixed-message size case would be trivial.

Implementation-wise there would be some similarity to what you've done - i.e. in the merge case it would allocate and fill a buffer (reused if possible) based on the current required bytes, only merging as much of the input buffer as required, and always offering the input buffer as-is if it is large enough when there's no cumulated data. It would also benefit the composite case in that we could trim the writer index of composite cumulations down to the required number of bytes, which makes it more likely that we can subsequently pass a single buffer rather than them continuing to operate "through" the composite on its final component.

This might be taking it too far but we could even have a version of the `setBytesRequired` that also takes a (boolean?) flag indicating the kind of access that next batch of bytes should be optimized for, which can be taken into account _if_ a cumulation is needed. This could be useful for cases where there's header-processing stages followed by larger chunks of content data.

Finally I noticed some related/tangential things:
- The logic to discard read bytes imho should be owned by the cumulator / cumulation logic - it is already naturally discarding data and knows how much is otherwise (easily) used/discardable so could trigger it based on some absolute threshold. Passing smaller buffers to `decode` will also reduce the amount of data that retained slices hold on to.
- Some other minor improvements can be made to the existing logic of both types of cumulators to reduce the amount of work done in various situations. I will withhold that for now pending decisions on the wider change
- There are more enhancements that can be made to `CompositeByteBuf` to reduce the cost of using it as a general alternative in places like this to avoid copying. I may open some PRs for that and it would be interesting to see how things compare with different benchmark workloads when all these things are combined 

I also have some comments on the benchmark but will save those for the time being!

belliottsmith:
@njhill this tactic would prohibit modifying the behaviour of pre-existing handlers, and also prevent more sophisticated use cases.  

A simple example: let's say you have a single buffer for the entire event loop in the upstream handler, that is sized to guarantee it is always sufficient; we want to ensure the downstream handler never saves the buffer it is passed, and currently this is impossible.  

Or another example: the upstream buffer is very large, and we want to ensure for this handler that we only save small buffers if possible.  How do we configure this behaviour in the API you propose?  This is orthogonal to the number of bytes, it's a decision about how we want to store those bytes, and this seems natural to offload to the Cumulator.

IMO, a networking library shouldn't try too hard to stay the hand of the sophisticated application developer, though I guess it depends what you consider your target demographic.

njhill:
> a networking library shouldn't try too hard to stay the hand of the sophisticated application developer

We are in total agreement there! I'm not suggesting to restrict what can be done, just debating whether there may be a more appropriate way to achieve it that preserves the current separation of concerns.

I _think_ I understand what you're getting at with these examples (apologies if not) - by "save" a buffer I presume you mean retain a reference to it via a retained slice. And in both cases the goal is to "protect" what could be a large input buffer from being retained, so that it can be reused.

Relying on custom cumulators for this doesn't seem like the right approach since it's kind of an independent and external concern to the cumulation strategy, which is more tied to the requirements of the decoder impl that it's associated with (merging vs composite, etc.). It presumably would also apply to any handler operating on said buffers - what if they don't happen to be decoders that extend `ByteToMessageDecoder`?

If this approach was taken though, would it suffice to define a third cumulator type for that purpose, like `MERGE_ALWAYS_COPY` or something? This could be used of course in combination with `setRequiredBytes` and would then end up behaving I think as you had planned. Is there any variation in anticipated usage beyond that which I haven't realized?

Had you considered just wrapping the input buffer (e.g. using `WrappedByteBuf` or a equiv since it's package-private) and overriding `retainedSlice(...)` to return a copy?

belliottsmith:
> Relying on custom cumulators for this doesn't seem like the right approach since it's kind of an independent and external concern to the cumulation strategy

The benefit with the cumulator approach is that we can also modify the existing decoders supplied with Netty, e.g. the SslHandler and the Lz4FrameDecoder, and we can do it in a uniform way.  

Of course, we could instead modify all of these handlers ourselves as well, providing some external method we can invoke to request this behaviour.  That would work for me, as I'm only interested in the capability, not the approach.

It still feels like a weird coupling of behaviour, because the other Cumulators cannot use this information.  If we were to modify the bounds of the buffer we supplied them from `ByteToMessageDecoder`, the existing Cumulators would behave incorrectly.  So I'm also not entirely sure how we would implement this, without completely deprecating Cumulator (which seems non-viable for a patch version)?

njhill:
I think it would help to be precise about what exactly the goal is.
1. I can absolutely see the current inefficiency w.r.t. the frequency and size of buffer combination operations being greater than necessary (whether by realloc + copying or composition), which could be substantially mitigated by the cumulation logic (wherever it may live) having some knowledge of the data size required by the decoder. This would appear to have very general applicability.
2. It sounds like the other more specialized goal (though partially related) is a general mechanism to prevent handlers in the inbound pipeline from holding references to the original upstream IO buffer (at least for longer than the scope of a single channel read operation), so that the buffer can be fully reused immediately.

Have I understood these two correctly? and is this exhaustive or are there other (even hypothetical) use cases which this is intended to address?

For (1), I agree that on the surface the existing public `Cumulator` interface isn't sufficient. But I'm not sure that matters given an assumption that existing custom cumulator impls are likely very rare. We just need to ensure that those still "work"; they don't need to be able to exploit new information provided for example by a new `setBytesRequired` method. The two existing cumulator _constants_ would remain, and in the (near universal) case that one of them is used, we would just have special handling inside `ByteToMessageDecoder`, possibly not even using their `cumulate` method. I expect the effective logic though would be quite similar to your changes here w.r.t. how the buffers are juggled.

For (2) assuming I've understood your aim correctly, I still think a wrapper which overrides `retainedSlice()` sounds like the best option. I really don't think the additional single allocation involved would be significant (less than slicing a buffer), and in fact may only need to be done once up front - i.e. you always send the same wrapper instance through the pipeline since it would be equivalent to the buffer it wraps.
> we don't retain a slice in ByteToMessageDecoder, we take the whole thing, since we already own it? 

But for that reason you also don't care about `BTMD`'s own use, right? It will always release the input buffer before `channelRead` returns (and can continue to do so), even if it holds onto it between successive calls to `decode`.

Again really sorry if I've missed, misunderstood or misrepresented anything. I do still think this is a great change at its core, just feel we should think carefully about how best to expose it. And of course this is just one view, thoughts from others would be helpful too!

njhill:
@belliottsmith apologies, I think for (2) I did misunderstand what you meant by "saving" the input buffer. Presumably you're referring to the case after the input buffer has been used directly and isn't fully consumed - i.e. where it's currently saved as the cumulation itself, to be appended to on a subsequent read call if possible?

I wonder if that could also be handled in a generic/default way using the information at hand. Specifically if we know the exact byte count required for the subsequent decode, it probably makes sense to _always_ perform the preemptive copy prior to returning from `channelRead` if the input buffer either doesn't have sufficient writable capacity, is read-only, or has a ref count > 1. Would this satisfy your use case - i.e. where the message size would be specified and you could ensure the refcount is > 1 on your single shared buffer?

I do still really like the idea of providing a `setRequiredBytes` method that can be called with potentially different values each time, and to make it even more broadly applicable it could take separate "necessary" and "sufficient" values. Then for example it could be called with `(1, 10)` if we are expecting a var int with max length of 10.

trustin:
I like the general idea of giving more information to `ByteToMessageDecoder`. Thanks for your inspiration.

I'm curious to know the benchmark numbers for the `MERGE` and `COMPOSITE` cases without this pull request applied, so we know the overhead incurred by it.

Is this change considered backward compatible? It seems to me that `Cumulator` is now given with less restriction than before. Should we schedule this for Netty 5?

trustin:
Another food for thought is how we could extend this idea of 'giving more information about how many bytes are expected to decode a message' to `RecvByteBufAllocator`.

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -302,6 +357,25 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
         }
     }
 
+    void saveInProgressToCumulation() {
```
franz1981: Can be made private final?

normanmaurer:
Sorry I was (an I am still) kind of lost in other stuff but will come back to this once I was able to release 4.1.37.Final

belliottsmith:
@normanmaurer that's ok, I _think_ we're at a point where other community members have indicated they would prefer an alternative approach.  Unfortunately I don't really have time to explore these other approaches, so I was letting that dialogue happen to see what people decided.  It looks to have stalled, so I'll await some input on whether or not I should pick up this approach again, or we should wait until another community member has a chance to trial a different one (I'm absolutely comfortable with whatever is decided).

njhill:
@belliottsmith @normanmaurer please see #9922! Sorry for the delay, took longer than anticipated to get into reasonable shape and I uncovered quite a few smaller things along the way while working on it that made sense to address first.

normanmaurer:
Let me close this for now... We can pick it up again if there is a need

PR #8891, opened 2019-02-26
Add test for Iterator.remove() on KObjectHashMap.values().iterator()

normanmaurer:
Motivation:

https://github.com/netty/netty/pull/8866 added support for calling Iterator.remove() but did not add a testcase.

Modifications:

Add testcase to ensure removal works.

Result:

Better test-coverage.


normanmaurer:
Let me merge this as its just a testcase for a PR we merged

PR #8895, opened 2019-02-27
Further reduce ensureAccessible() overhead

njhill:
Motivation:

This PR fixes some non-negligible overhead discovered in the `ByteBuf` accessibility (non-zero refcount) checking. The cause turned out to be mostly twofold:
- Unnecessary operations used to calculate the refcount from the "raw" encoded `refCnt` field value
- Call stack depths exceeding the default limit for inlining, in some places (`CompositeByteBuf` in particular)

It's a follow-on from #8882 which uses the `maxCapacity` field for a simpler non-negative check. The performance gap between these two variants appears to be _mostly_ closed, but there's one exception which may warrant further analysis (`readBatch` still appears to be 15-20% slower).

Modifications:

- Replace `AbstractByteBuf.internalRefCount()` with (package-private) `ByteBuf.isAccessible()`, the default still checks for non-zero `refCnt()`
- Just test for parity of raw `refCnt` instead of converting to "real", with fast-path for specific small values
- Make sure `isAccessible()` is delegated by derived/wrapper `ByteBuf`s
- Use existing freed flag in `CompositeByteBuf` for faster `isAccessible()`
- Manually inline some calls in methods like `CompositeByteBuf.setLong(...)` and `AbstractReferenceCountedByteBuf.isAccessible()` to reduce call stack depths
- Add `ByteBufAccessBenchmark` which is an extension of `UnsafeByteBufBenchmark` (maybe latter could now be removed)

Results:

Before:
```
Benchmark   (bufferType)  (checkAccessible)  (checkBounds)   Mode  Cnt          Score          Error  Units
readBatch         UNSAFE               true           true  thrpt   30   84524972.863 ±   518338.811  ops/s
readBatch   UNSAFE_SLICE               true           true  thrpt   30   38608795.037 ±   298176.974  ops/s
readBatch           HEAP               true           true  thrpt   30   80003697.649 ±   974674.119  ops/s
readBatch      COMPOSITE               true           true  thrpt   30   18495554.788 ±   108075.023  ops/s
setGetLong        UNSAFE               true           true  thrpt   30  247069881.578 ± 10839162.593  ops/s
setGetLong  UNSAFE_SLICE               true           true  thrpt   30  196355905.206 ±  1802420.990  ops/s
setGetLong          HEAP               true           true  thrpt   30  245686644.713 ± 11769311.527  ops/s
setGetLong     COMPOSITE               true           true  thrpt   30   83170940.687 ±   657524.123  ops/s
setLong           UNSAFE               true           true  thrpt   30  278940253.918 ±  1807265.259  ops/s
setLong     UNSAFE_SLICE               true           true  thrpt   30  202556738.764 ± 11887973.563  ops/s
setLong             HEAP               true           true  thrpt   30  280045958.053 ±  2719583.400  ops/s
setLong        COMPOSITE               true           true  thrpt   30  121299806.002 ±  2155084.707  ops/s
```

After:
```
Benchmark   (bufferType)  (checkAccessible)  (checkBounds)   Mode  Cnt          Score          Error  Units
readBatch         UNSAFE               true           true  thrpt   30  101641801.035 ±  3950050.059  ops/s
readBatch   UNSAFE_SLICE               true           true  thrpt   30   84395902.846 ±  4339579.057  ops/s
readBatch           HEAP               true           true  thrpt   30  100179060.207 ±  3222487.287  ops/s
readBatch      COMPOSITE               true           true  thrpt   30   42288494.472 ±   294919.633  ops/s
setGetLong        UNSAFE               true           true  thrpt   30  304530755.027 ±  6574163.899  ops/s
setGetLong  UNSAFE_SLICE               true           true  thrpt   30  212028547.645 ± 14277828.768  ops/s
setGetLong          HEAP               true           true  thrpt   30  309335422.609 ±  2272150.415  ops/s
setGetLong     COMPOSITE               true           true  thrpt   30  160383609.236 ±   966484.033  ops/s
setLong           UNSAFE               true           true  thrpt   30  298055969.747 ±  7437449.627  ops/s
setLong     UNSAFE_SLICE               true           true  thrpt   30  223784178.650 ±  9869750.095  ops/s
setLong             HEAP               true           true  thrpt   30  302543263.328 ±  8140104.706  ops/s
setLong        COMPOSITE               true           true  thrpt   30  157083673.285 ±  3528779.522  ops/s
```
There's also a similar knock-on improvement to other benchmarks (e.g. HPACK encoding/decoding) as shown in #8882.

For sanity I did a final comparison of the included "fast path" tweak using one of the HPACK benchmarks:

`(rawCnt & 1) == 0`:
```
Benchmark  (limitToAscii)  (sensitive)  (size)   Mode  Cnt      Score     Error  Units
decode               true         true  MEDIUM  thrpt   30  50914.479 ± 940.114  ops/s
```
`rawCnt == 2 || rawCnt == 4 || rawCnt == 6 || rawCnt == 8 ||  (rawCnt & 1) == 0`:
```
Benchmark  (limitToAscii)  (sensitive)  (size)   Mode  Cnt      Score      Error  Units
decode               true         true  MEDIUM  thrpt   30  60036.425 ± 1478.196  ops/s
```

### DIFF
--- a/microbench/src/main/java/io/netty/buffer/ByteBufAccessBenchmark.java
+++ b/microbench/src/main/java/io/netty/buffer/ByteBufAccessBenchmark.java
@@ -19,8 +19,11 @@ import java.nio.ByteBuffer;
 import java.util.concurrent.TimeUnit;
 
 import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
 import org.openjdk.jmh.annotations.Fork;
 import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
 import org.openjdk.jmh.annotations.Param;
 import org.openjdk.jmh.annotations.Setup;
 import org.openjdk.jmh.annotations.TearDown;
@@ -32,6 +35,8 @@ import io.netty.util.internal.PlatformDependent;
 @Warmup(iterations = 5, time = 1500, timeUnit = TimeUnit.MILLISECONDS)
 @Measurement(iterations = 10, time = 1500, timeUnit = TimeUnit.MILLISECONDS)
 @Fork(3)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
 public class ByteBufAccessBenchmark extends AbstractMicrobenchmark {
 
     static final class NioFacade extends WrappedByteBuf {
@@ -111,6 +116,9 @@ public class ByteBufAccessBenchmark extends AbstractMicrobenchmark {
     @Param({ "true", "false" })
     public String checkBounds;
 
+    @Param({ "8" })
+    public int batchSize; // applies only to readBatch benchmark
+
     @Setup
     public void setup() {
         System.setProperty("io.netty.buffer.checkAccessible", checkAccessible);
@@ -123,6 +131,8 @@ public class ByteBufAccessBenchmark extends AbstractMicrobenchmark {
     @TearDown
     public void tearDown() {
         buffer.release();
+        System.clearProperty("io.netty.buffer.checkAccessible");
+        System.clearProperty("io.netty.buffer.checkBounds");
     }
 
     @Benchmark
@@ -139,7 +149,7 @@ public class ByteBufAccessBenchmark extends AbstractMicrobenchmark {
     public int readBatch() {
         buffer.readerIndex(0).touch();
         int result = 0;
-        for (int i = 0; i < 8; i++) {
+        for (int i = 0, size = batchSize; i < size; i++) {
             result += buffer.readByte();
         }
         return result;
--- a/microbench/src/main/java/io/netty/buffer/ByteBufAccessBenchmark.java
+++ b/microbench/src/main/java/io/netty/buffer/ByteBufAccessBenchmark.java
@@ -1,0 +1,147 @@
+/*
+* Copyright 2019 The Netty Project
+*
+* The Netty Project licenses this file to you under the Apache License,
+* version 2.0 (the "License"); you may not use this file except in compliance
+* with the License. You may obtain a copy of the License at:
+*
+*   http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+* WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+* License for the specific language governing permissions and limitations
+* under the License.
+*/
+package io.netty.buffer;
+
+import java.nio.ByteBuffer;
+import java.util.concurrent.TimeUnit;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.annotations.Warmup;
+
+import io.netty.microbench.util.AbstractMicrobenchmark;
+import io.netty.util.internal.PlatformDependent;
+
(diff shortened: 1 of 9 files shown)


normanmaurer:
@netty-bot test this please

microbench/src/main/java/io/netty/buffer/ByteBufAccessBenchmark.java
```diff
@@ -0,0 +1,147 @@
...
+    @Param({ "true", "false" })
+    public String checkAccessible;
+
+    @Param({ "true", "false" })
+    public String checkBounds;
+
+    @Setup
+    public void setup() {
+        System.setProperty("io.netty.buffer.checkAccessible", checkAccessible);
+        System.setProperty("io.netty.buffer.checkBounds", checkBounds);
```
normanmaurer: we should reset these to the "default values" in `tearDown()` to ensure we not affect others benchmarks.

microbench/src/main/java/io/netty/buffer/ByteBufAccessBenchmark.java
```diff
@@ -0,0 +1,147 @@
...
+    @Benchmark
+    public ByteBuf setLong() {
+        return buffer.setLong(0, 1);
+    }
+
+    @Benchmark
+    public int readBatch() {
+        buffer.readerIndex(0).touch();
+        int result = 0;
+        for (int i = 0; i < 8; i++) {
```
franz1981: I would add a batchSize parameter defaulted to 8 :)

microbench/src/main/java/io/netty/buffer/ByteBufAccessBenchmark.java
```diff
@@ -0,0 +1,147 @@
...
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.annotations.Warmup;
+
+import io.netty.microbench.util.AbstractMicrobenchmark;
+import io.netty.util.internal.PlatformDependent;
+
+@Warmup(iterations = 5, time = 1500, timeUnit = TimeUnit.MILLISECONDS)
+@Measurement(iterations = 10, time = 1500, timeUnit = TimeUnit.MILLISECONDS)
+@Fork(3)
+public class ByteBufAccessBenchmark extends AbstractMicrobenchmark {
```
ejona86: Using Sampling or Average instead of Ops/s produces _much_ more clear results and allows you to do things like add two benchmarks together to get a rough idea of how long a composite operation may take (and then you can also compare that against a benchmark).

microbench/src/main/java/io/netty/buffer/ByteBufAccessBenchmark.java
njhill: @normanmaurer I've made this change but it could only apply if someone was running with forks set to 0, right?

njhill:
Thanks @normanmaurer @franz1981 @ejona86, have added a commit addressing the comments.

microbench/src/main/java/io/netty/buffer/ByteBufAccessBenchmark.java
normanmaurer: right

normanmaurer:
@netty-bot test this please

normanmaurer:
@njhill thanks a lot again for all the work. Nice one!

carl-mastrangelo:
LGTM

microbench/src/main/java/io/netty/buffer/ByteBufAccessBenchmark.java
franz1981: @njhill @ejona86 
Beware the precision & granularity of `System::nanoTime` used to collect the averageTime as shown on https://shipilev.net/blog/2014/nanotrusting-nanotime/.
The risk is that operations with < 15 ns cost won't be correctly measured in many *nix envs!

ejona86: @franz1981, I don't know what you're talking about. Average is just "total amount of time / number of runs". Aleksey even shows the code explicitly in latency_nanotime_AverageTime_measurementLoop and notes it is okay. I would agree sampling could have trouble, although the aliasing should also be obvious within the results.

franz1981: @ejona86 I'm saying that If you have a precision on nanoTime of 15 ns you wouldn't be able to get meaningful averages for operations that take less then this time to be executed...

PR #8896, opened 2019-02-27
WebSocket client handshaker to support "force close" after timeout

### LINKED ISSUE #8883: Websocket client closing handshake to support "force close" after given timetout
RFC455 defines that, generally, a WebSocket client should not close a TCP connection as far as a server is the one who's responsible for doing that. In practice, when I do not control the server, this might lead to connections leak if the server does not comply with the RFC.

[RFC455#7.1.1](https://tools.ietf.org/html/rfc6455#section-7.1.1) says

> In abnormal cases (such as not having received a TCP Close from the server after a reasonable amount of time) a client MAY initiate the TCP Close.

I suggest to add `forceCloseAfterMillis` option to `WebSocketClientHandshaker` (and to `WebSocketClientProtocolHandler` appropriately) and schedule a channel close when necessary after `CloseWebSocketFrame` being flushed [here](https://github.com/netty/netty/blob/f176384a729c1d9352c9ed878b9b967ca2f31bf8/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java#L430-L435). If there's no objections, I'm happy to provide a PR.

kachayev:
Motivation:

RFC 6455 defines that, generally, a WebSocket client should not close a TCP
connection as far as a server is the one who's responsible for doing that.
In practice tho', it's not always possible to control the server. Server's
misbehavior may lead to connections being leaked (if the server does not
comply with the RFC).

RFC 6455 #7.1.1 says

> In abnormal cases (such as not having received a TCP Close from the server
after a reasonable amount of time) a client MAY initiate the TCP Close.

Modifications:

* WebSocket client handshaker additional param `forceCloseAfterMillis`

* Switched off by default to keep backward compatibility

Result:

WebSocket client handshaker to comply with RFC. Fixes #8883.

---

Open questions/comments:

* I've implemented this as a `WebSocketClientHandshaker` functionality, not in handlers, as far as it's responsible for `close`.

* Nothing prevents you to call `handshaker.close` twice, in such a case only first timeout would be applied.

* Do we need to keep this feature switched OFF by default to ensure backward compatibility? I mean... it's hard to justify current implementation and I think 10 seconds default timeout should be enough to cover potential drawbacks. @normanmaurer WDYT?

* Testing is kind of clumsy... It's impossible to detect if the channel was closed because of timeout or because of the server or... whatever. Maybe I'm missing any better approach to test this?

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
@@ -40,6 +40,9 @@ import io.netty.util.internal.ThrowableUtil;
 import java.net.URI;
 import java.nio.channels.ClosedChannelException;
 import java.util.Locale;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
 /**
  * Base class for web socket client handshake implementations
@@ -50,12 +53,22 @@ public abstract class WebSocketClientHandshaker {
 
     private static final String HTTP_SCHEME_PREFIX = HttpScheme.HTTP + "://";
     private static final String HTTPS_SCHEME_PREFIX = HttpScheme.HTTPS + "://";
+    protected static final int DEFAULT_FORCE_CLOSE_TIMEOUT_MILLIS = 10000;
 
     private final URI uri;
 
     private final WebSocketVersion version;
 
     private volatile boolean handshakeComplete;
+
+    private volatile long forceCloseTimeoutMillis = DEFAULT_FORCE_CLOSE_TIMEOUT_MILLIS;
+
+    private volatile int forceCloseInit;
+
+    private static final AtomicIntegerFieldUpdater<WebSocketClientHandshaker> FORCE_CLOSE_INIT_UPDATER =
+            AtomicIntegerFieldUpdater.newUpdater(WebSocketClientHandshaker.class, "forceCloseInit");
+
+    private volatile boolean forceCloseComplete;
 
     private final String expectedSubprotocol;
 
@@ -82,11 +95,35 @@ public abstract class WebSocketClientHandshaker {
      */
     protected WebSocketClientHandshaker(URI uri, WebSocketVersion version, String subprotocol,
                                         HttpHeaders customHeaders, int maxFramePayloadLength) {
+        this(uri, version, subprotocol, customHeaders, maxFramePayloadLength, DEFAULT_FORCE_CLOSE_TIMEOUT_MILLIS);
+    }
+
+    /**
+     * Base constructor
+     *
+     * @param uri
+     *            URL for web socket communications. e.g "ws://myhost.com/mypath". Subsequent web socket frames will be
+     *            sent to this URL.
+     * @param version
+     *            Version of web socket specification to use to connect to the server
+     * @param subprotocol
+     *            Sub protocol request sent to the server.
+     * @param customHeaders
+     *            Map of custom headers to add to the client request
+     * @param maxFramePayloadLength
+     *            Maximum length of a frame's payload
+     * @param forceCloseTimeoutMillis
+     *            Close the connection if it was not closed by the server after timeout specified
+     */
+    protected WebSocketClientHandshaker(URI uri, WebSocketVersion version, String subprotocol,
+                                        HttpHeaders customHeaders, int maxFramePayloadLength,
+                                        long forceCloseTimeoutMillis) {
         this.uri = uri;
         this.version = version;
         expectedSubprotocol = subprotocol;
         this.customHeaders = customHeaders;
         this.maxFramePayloadLength = maxFramePayloadLength;
+        this.forceCloseTimeoutMillis = forceCloseTimeoutMillis;
     }
 
     /**
@@ -138,6 +175,29 @@ public abstract class WebSocketClientHandshaker {
 
     private void setActualSubprotocol(String actualSubprotocol) {
         this.actualSubprotocol = actualSubprotocol;
+    }
+
+    public long forceCloseTimeoutMillis() {
+        return forceCloseTimeoutMillis;
+    }
+
+    /**
+     * Flag to indicate if the closing handshake was initiated because of timeout.
+     * For testing only.
+     */
+    protected boolean isForceCloseComplete() {
+        return forceCloseComplete;
+    }
+
+    /**
+     * Sets timeout to close the connection if it was not closed by the server.
+     *
+     * @param forceCloseTimeoutMillis
+     *            Close the connection if it was not closed by the server after timeout specified
+     */
+    public WebSocketClientHandshaker setForceCloseTimeoutMillis(long forceCloseTimeoutMillis) {
+        this.forceCloseTimeoutMillis = forceCloseTimeoutMillis;
+        return this;
     }
 
     /**
@@ -431,7 +491,46 @@ public abstract class WebSocketClientHandshaker {
         if (channel == null) {
             throw new NullPointerException("channel");
         }
-        return channel.writeAndFlush(frame, promise);
+        channel.writeAndFlush(frame, promise);
+        applyForceCloseTimeout(channel, promise);
+        return promise;
+    }
+
+    private void applyForceCloseTimeout(final Channel channel, ChannelFuture flushFuture) {
+        final long forceCloseTimeoutMillis = this.forceCloseTimeoutMillis;
+        final WebSocketClientHandshaker handshaker = this;
+        if (forceCloseTimeoutMillis <= 0 || !channel.isActive() || forceCloseInit != 0) {
+            return;
+        }
+
+        flushFuture.addListener(new ChannelFutureListener() {
+            @Override
+            public void operationComplete(ChannelFuture future) throws Exception {
+                // If flush operation failed, there is no reason to expect
+                // a server to receive CloseFrame. Thus this should be handled
+                // by the application separately.
+                // Also, close might be called twice from different threads.
+                if (future.isSuccess() && channel.isActive() &&
+                        FORCE_CLOSE_INIT_UPDATER.compareAndSet(handshaker, 0, 1)) {
+                    final Future<?> forceCloseFuture = channel.eventLoop().schedule(new Runnable() {
+                        @Override
+                        public void run() {
+                            if (channel.isActive()) {
+                                channel.close();
+                                forceCloseComplete = true;
+                            }
+                        }
+                    }, forceCloseTimeoutMillis, TimeUnit.MILLISECONDS);
+
+                    channel.closeFuture().addListener(new ChannelFutureListener() {
+                        @Override
+                        public void operationComplete(ChannelFuture future) throws Exception {
+                            forceCloseFuture.cancel(false);
+                        }
+                    });
+                }
+            }
+        });
     }
 
     /**
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker00.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker00.java
@@ -48,7 +48,7 @@ public class WebSocketClientHandshaker00 extends WebSocketClientHandshaker {
     private ByteBuf expectedChallengeResponseBytes;
 
     /**
-     * Constructor specifying the destination web socket location and version to initiate
+     * Creates a new instance with the specified destination WebSocket location and version to initiate.
      *
      * @param webSocketURL
      *            URL for web socket communications. e.g "ws://myhost.com/mypath". Subsequent web socket frames will be
@@ -64,7 +64,31 @@ public class WebSocketClientHandshaker00 extends WebSocketClientHandshaker {
      */
     public WebSocketClientHandshaker00(URI webSocketURL, WebSocketVersion version, String subprotocol,
             HttpHeaders customHeaders, int maxFramePayloadLength) {
-        super(webSocketURL, version, subprotocol, customHeaders, maxFramePayloadLength);
+        this(webSocketURL, version, subprotocol, customHeaders, maxFramePayloadLength,
+                DEFAULT_FORCE_CLOSE_TIMEOUT_MILLIS);
+    }
+
+    /**
+     * Creates a new instance with the specified destination WebSocket location and version to initiate.
+     *
+     * @param webSocketURL
+     *            URL for web socket communications. e.g "ws://myhost.com/mypath". Subsequent web socket frames will be
+     *            sent to this URL.
+     * @param version
+     *            Version of web socket specification to use to connect to the server
+     * @param subprotocol
+     *            Sub protocol request sent to the server.
+     * @param customHeaders
+     *            Map of custom headers to add to the client request
+     * @param maxFramePayloadLength
+     *            Maximum length of a frame's payload
+     * @param forceCloseTimeoutMillis
+     *            Close the connection if it was not closed by the server after timeout specified
+     */
+    public WebSocketClientHandshaker00(URI webSocketURL, WebSocketVersion version, String subprotocol,
+                                       HttpHeaders customHeaders, int maxFramePayloadLength,
+                                       long forceCloseTimeoutMillis) {
+        super(webSocketURL, version, subprotocol, customHeaders, maxFramePayloadLength, forceCloseTimeoutMillis);
     }
 
     /**
@@ -243,4 +267,11 @@ public class WebSocketClientHandshaker00 extends WebSocketClientHandshaker {
     protected WebSocketFrameEncoder newWebSocketEncoder() {
         return new WebSocket00FrameEncoder();
     }
+
+    @Override
+    public WebSocketClientHandshaker00 setForceCloseTimeoutMillis(long forceCloseTimeoutMillis) {
+        super.setForceCloseTimeoutMillis(forceCloseTimeoutMillis);
+        return this;
+    }
+
 }
--- a/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker07.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker07.java
@@ -95,9 +95,42 @@ public class WebSocketClientHandshaker07 extends WebSocketClientHandshaker {
      *            accepted.
      */
     public WebSocketClientHandshaker07(URI webSocketURL, WebSocketVersion version, String subprotocol,
+                                       boolean allowExtensions, HttpHeaders customHeaders, int maxFramePayloadLength,
+                                       boolean performMasking, boolean allowMaskMismatch) {
+        this(webSocketURL, version, subprotocol, allowExtensions, customHeaders, maxFramePayloadLength, performMasking,
+                allowMaskMismatch, DEFAULT_FORCE_CLOSE_TIMEOUT_MILLIS);
+    }
+
+    /**
+     * Creates a new instance.
+     *
+     * @param webSocketURL
+     *            URL for web socket communications. e.g "ws://myhost.com/mypath". Subsequent web socket frames will be
+     *            sent to this URL.
+     * @param version
+     *            Version of web socket specification to use to connect to the server
+     * @param subprotocol
+     *            Sub protocol request sent to the server.
+     * @param allowExtensions
+     *            Allow extensions to be used in the reserved bits of the web socket frame
+     * @param customHeaders
+     *            Map of custom headers to add to the client request
+     * @param maxFramePayloadLength
+     *            Maximum length of a frame's payload
+     * @param performMasking
+     *            Whether to mask all written websocket frames. This must be set to true in order to be fully compatible
+     *            with the websocket specifications. Client applications that communicate with a non-standard server
+     *            which doesn't require masking might set this to false to achieve a higher performance.
+     * @param allowMaskMismatch
+     *            When set to true, frames which are not masked properly according to the standard will still be
+     *            accepted
+     * @param forceCloseTimeoutMillis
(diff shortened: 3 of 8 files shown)


normanmaurer:
@kachayev its RFC6455 ;)

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
```diff
@@ -57,6 +60,10 @@
 
     private volatile boolean handshakeComplete;
 
+    private volatile long forceCloseTimeoutMillis = -1;
```
normanmaurer: why is this volatile ?

codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketHandshakeHandOverTest.java
```diff
@@ -39,13 +43,33 @@
     private WebSocketServerProtocolHandler.HandshakeComplete serverHandshakeComplete;
     private boolean clientReceivedHandshake;
     private boolean clientReceivedMessage;
+    private boolean serverReceivedCloseHandshake;
+    private boolean clientForceClosed;
+
+    public class CloseNoOpServerProtocolHandler extends WebSocketServerProtocolHandler {
```
normanmaurer: private final ?

codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketHandshakeHandOverTest.java
```diff
@@ -39,13 +43,33 @@
...
+        public CloseNoOpServerProtocolHandler(String websocketPath) {
```
normanmaurer: remove public

codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketHandshakeHandOverTest.java
```diff
@@ -39,13 +43,33 @@
...
+            super(websocketPath, null, false);
+        }
+
+        @Override
+        protected void decode(ChannelHandlerContext ctx, WebSocketFrame frame, List<Object> out) throws Exception {
+            if (frame instanceof CloseWebSocketFrame) {
+                serverReceivedCloseHandshake = true;
+                frame.retain();
```
normanmaurer: why do you retain here ? This looks like a possible memory leak.

codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketHandshakeHandOverTest.java
```diff
@@ -95,6 +119,63 @@ protected void channelRead0(ChannelHandlerContext ctx, Object msg) throws Except
...
+            Thread.sleep(10);
+            // We need to run all pending tasks as the force close timeout is scheduled on the EventLoop.
+            clientChannel.runPendingTasks();
+        }
+
+        // It should not take too long before the connection would be closed
+        assertTrue(attemptsLeft > 0);
+
+        serverChannel.finishAndReleaseAll();
+        clientChannel.finishAndReleaseAll();
```
normanmaurer: assert both return values

kachayev:
> @kachayev its RFC6455 ;)

@normanmaurer Copy-paste error ;)

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
kachayev: Previously, I had an API to change this value after a handshaker is already created. As far as I removed that, I can remove `volatile` either.

codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketHandshakeHandOverTest.java
kachayev: Oops, that one should release the frame.

kachayev: Yeah, that's a bug

kachayev:
@normanmaurer Updated. I still think if need to cover this use case:

```
handshaker.setForceCloseTimeoutMillis(1_000).close(ch, new CloseWebsocketFrame());
```

It might be useful if the timeout depends on what happened during the communication (then I have no chance to setup it properly before the upgrade)... I just don't know how often you might need this in practice.

normanmaurer:
@kachayev another alternative would be to allow to overload the method with a timeout.

kachayev:
@normanmaurer Added setter to cover more use cases. Changed test to use an internal flag that indicated that the timeout was actually fired.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
```diff
@@ -431,7 +464,50 @@ public ChannelFuture close(Channel channel, CloseWebSocketFrame frame, ChannelPr
...
         }
-        return channel.writeAndFlush(frame, promise);
+        channel.writeAndFlush(frame, promise);
+        applyForceCloseTimeout(channel, promise);
+        return promise;
+    }
+
+    private void applyForceCloseTimeout(final Channel channel, ChannelFuture flushFuture) {
+        final long forceCloseTimeoutMillis = this.forceCloseTimeoutMillis;
+        if (forceCloseTimeoutMillis <= 0 || !channel.isActive()) {
```
normanmaurer: nit: you may also want to check `!forceCloseInit.get()`

codec-http/src/test/java/io/netty/handler/codec/http/websocketx/WebSocketHandshakeHandOverTest.java
```diff
@@ -95,6 +118,64 @@ protected void channelRead0(ChannelHandlerContext ctx, Object msg) throws Except
         assertTrue(clientReceivedMessage);
     }
 
+    @Test
```
normanmaurer: Use a timeout as we may never complete the loop here when there is a bug:

https://github.com/netty/netty/pull/8896/files#diff-96a945171dadb844c094502b4d37149bR164

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
kachayev: nice!

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
```diff
@@ -140,6 +153,25 @@ private void setActualSubprotocol(String actualSubprotocol) {
...
 
+    public long forceCloseTimeoutMillis() { return forceCloseTimeoutMillis; }
+
+    /**
+     * Flag to indicate if the closing handshake was initiated because of timeout.
+     * For testing only.
+     */
+    protected boolean isForceCloseComplete() { return forceCloseComplete; }
+
+    private void setForceCloseComplete() { forceCloseComplete = true; }
```
normanmaurer: either remove the method or fix the styling to use:

```java
private void setForceCloseComplete() {
    ....
}
```

I would personally just remove it.

kachayev: Removed.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
```diff
@@ -140,6 +153,27 @@ private void setActualSubprotocol(String actualSubprotocol) {
...
+     */
+    protected boolean isForceCloseComplete() {
+        return forceCloseComplete;
+    }
+
+    /**
+     * @param forceCloseTimeoutMillis
+     *            Close the connection if it was not closed by the server after timeout specified
+     */
+    public WebSocketClientHandshaker setForceCloseTimeoutMillis(long forceCloseTimeoutMillis) {
```
normanmaurer: I think you should override this one in each implementation to return the correct type.

kachayev: Done!

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker00.java
```diff
@@ -243,4 +266,15 @@ protected WebSocketFrameDecoder newWebsocketDecoder() {
...
         return new WebSocket00FrameEncoder();
     }
+
+    /**
+     * @param forceCloseTimeoutMillis
+     *            Close the connection if it was not closed by the server after timeout specified
+     */
+    @Override
+    public WebSocketClientHandshaker00 setForceCloseTimeoutMillis(long forceCloseTimeoutMillis) {
+        this.forceCloseTimeoutMillis = forceCloseTimeoutMillis;
```
normanmaurer: nit: jut call `super.setForceCloseTimeoutMillis(...)` this will also ensure we do the correct thing if we ever change the super method content and will allow to make `forceCloseTimeMillis` private

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker07.java
```diff
@@ -216,4 +249,15 @@ protected WebSocketFrameDecoder newWebsocketDecoder() {
...
         return new WebSocket07FrameEncoder(performMasking);
     }
+
+    /**
+     * @param forceCloseTimeoutMillis
+     *            Close the connection if it was not closed by the server after timeout specified
+     */
+    @Override
+    public WebSocketClientHandshaker07 setForceCloseTimeoutMillis(long forceCloseTimeoutMillis) {
+        this.forceCloseTimeoutMillis = forceCloseTimeoutMillis;
```
normanmaurer: nit: jut call `super.setForceCloseTimeoutMillis(...)` this will also ensure we do the correct thing if we ever change the super method content and will allow to make `forceCloseTimeMillis` private

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker08.java
```diff
@@ -217,4 +251,15 @@ protected WebSocketFrameDecoder newWebsocketDecoder() {
...
         return new WebSocket08FrameEncoder(performMasking);
     }
+
+    /**
+     * @param forceCloseTimeoutMillis
+     *            Close the connection if it was not closed by the server after timeout specified
+     */
+    @Override
+    public WebSocketClientHandshaker08 setForceCloseTimeoutMillis(long forceCloseTimeoutMillis) {
+        this.forceCloseTimeoutMillis = forceCloseTimeoutMillis;
```
normanmaurer: nit: jut call `super.setForceCloseTimeoutMillis(...)` this will also ensure we do the correct thing if we ever change the super method content and will allow to make `forceCloseTimeMillis` private

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker13.java
```diff
@@ -217,4 +252,15 @@ protected WebSocketFrameDecoder newWebsocketDecoder() {
...
         return new WebSocket13FrameEncoder(performMasking);
     }
+
+    /**
+     * @param forceCloseTimeoutMillis
+     *            Close the connection if it was not closed by the server after timeout specified
+     */
+    @Override
+    public WebSocketClientHandshaker13 setForceCloseTimeoutMillis(long forceCloseTimeoutMillis) {
+        this.forceCloseTimeoutMillis = forceCloseTimeoutMillis;
```
normanmaurer: nit: jut call `super.setForceCloseTimeoutMillis(...)` this will also ensure we do the correct thing if we ever change the super method content and will allow to make `forceCloseTimeMillis` private

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker13.java
```diff
@@ -217,4 +252,15 @@ protected WebSocketFrameDecoder newWebsocketDecoder() {
     protected WebSocketFrameEncoder newWebSocketEncoder() {
         return new WebSocket13FrameEncoder(performMasking);
     }
+
+    /**
+     * @param forceCloseTimeoutMillis
+     *            Close the connection if it was not closed by the server after timeout specified
+     */
```
normanmaurer: nit: you can remove the java docs as it just overrides a method.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker08.java
```diff
@@ -217,4 +251,15 @@ protected WebSocketFrameDecoder newWebsocketDecoder() {
     protected WebSocketFrameEncoder newWebSocketEncoder() {
         return new WebSocket08FrameEncoder(performMasking);
     }
+
+    /**
+     * @param forceCloseTimeoutMillis
+     *            Close the connection if it was not closed by the server after timeout specified
+     */
```
normanmaurer: nit: you can remove the java docs as it just overrides a method.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker07.java
```diff
@@ -216,4 +249,15 @@ protected WebSocketFrameDecoder newWebsocketDecoder() {
     protected WebSocketFrameEncoder newWebSocketEncoder() {
         return new WebSocket07FrameEncoder(performMasking);
     }
+
+    /**
+     * @param forceCloseTimeoutMillis
+     *            Close the connection if it was not closed by the server after timeout specified
+     */
```
normanmaurer: nit: you can remove the java docs as it just overrides a method.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker00.java
```diff
@@ -243,4 +266,15 @@ protected WebSocketFrameDecoder newWebsocketDecoder() {
     protected WebSocketFrameEncoder newWebSocketEncoder() {
         return new WebSocket00FrameEncoder();
     }
+
+    /**
+     * @param forceCloseTimeoutMillis
+     *            Close the connection if it was not closed by the server after timeout specified
+     */
```
normanmaurer: nit: you can remove the java docs as it just overrides a method.

normanmaurer:
@netty-bot test this please

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
```diff
@@ -79,14 +88,18 @@
      *            Map of custom headers to add to the client request
      * @param maxFramePayloadLength
      *            Maximum length of a frame's payload
+     * @param forceCloseTimeoutMillis
+     *            Close the connection if it was not closed by the server after timeout specified
      */
     protected WebSocketClientHandshaker(URI uri, WebSocketVersion version, String subprotocol,
-                                        HttpHeaders customHeaders, int maxFramePayloadLength) {
+                                        HttpHeaders customHeaders, int maxFramePayloadLength,
+                                        long forceCloseTimeoutMillis) {
```
normanmaurer: as this is protected we will need to also retain the old constructor as otherwise its API breakage:

```
20:53:51 [ERROR] Failed to execute goal com.github.siom79.japicmp:japicmp-maven-plugin:0.13.1:cmp (default) on project netty-codec-http: There is at least one incompatibility: io.netty.handler.codec.http.websocketx.WebSocketClientHandshaker.WebSocketClientHandshaker(java.net.URI,io.netty.handler.codec.http.websocketx.WebSocketVersion,java.lang.String,io.netty.handler.codec.http.HttpHeaders,int):CONSTRUCTOR_REMOVED -> [Help 1]
```

kachayev: Done!

normanmaurer:
@netty-bot test this please

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
```diff
@@ -82,11 +91,35 @@
      */
     protected WebSocketClientHandshaker(URI uri, WebSocketVersion version, String subprotocol,
                                         HttpHeaders customHeaders, int maxFramePayloadLength) {
+        this(uri, version, subprotocol, customHeaders, maxFramePayloadLength,-1);
```
normanmaurer: nit: add a whitespace after `,`

kachayev: Done!

normanmaurer:
@netty-bot test this please

normanmaurer:
@trustin can you have a look as well please ?

normanmaurer:
@netty-bot test this please

trustin:
> Switched off by default to keep backward compatibility

Probably better enabling by default with reasonable default? Leaky connection for backward compatibility doesn't sound right to me.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
```diff
@@ -57,6 +60,12 @@
 
     private volatile boolean handshakeComplete;
 
+    private volatile long forceCloseTimeoutMillis = -1;
+
+    private final AtomicBoolean forceCloseInit = new AtomicBoolean(false);
```
trustin: Could you use `Atomic*FieldUpdater` instead?

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
```diff
@@ -82,11 +91,35 @@
      */
     protected WebSocketClientHandshaker(URI uri, WebSocketVersion version, String subprotocol,
                                         HttpHeaders customHeaders, int maxFramePayloadLength) {
+        this(uri, version, subprotocol, customHeaders, maxFramePayloadLength, -1);
```
trustin: Could extract the default value into a static final field.

normanmaurer:
@trustin sounds fine to me... maybe 3 seconds (which is also the default we use for SSL close notify) is ok ? Or do we want to use 10 seconds as we use for handshake timeout in SSL by default ?

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
```diff
@@ -140,6 +173,27 @@ private void setActualSubprotocol(String actualSubprotocol) {
...
+    /**
+     * Flag to indicate if the closing handshake was initiated because of timeout.
+     * For testing only.
+     */
+    protected boolean isForceCloseComplete() {
+        return forceCloseComplete;
+    }
+
+    /**
+     * @param forceCloseTimeoutMillis
```
trustin: Could you add the Javadoc summary, because otherwise the generated Javadoc will have empty summary.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker00.java
```diff
@@ -64,7 +64,30 @@
      */
     public WebSocketClientHandshaker00(URI webSocketURL, WebSocketVersion version, String subprotocol,
             HttpHeaders customHeaders, int maxFramePayloadLength) {
-        super(webSocketURL, version, subprotocol, customHeaders, maxFramePayloadLength);
+        this(webSocketURL, version, subprotocol, customHeaders, maxFramePayloadLength, -1);
+    }
+
+    /**
+     * Constructor specifying the destination web socket location and version to initiate
```
trustin: How about:

```
Creates a new instance with the specified destination WebSocket location and version to initiate.
```

trustin:
Not a WebSocket expert. 😅 Leaving it up to @kachayev 😆

normanmaurer:
@slandelle can you have a look as well ?

kachayev:
@trustin 

> enabling by default

Absolutely, 100% for it. I will put 10 seconds to stay consistent with SSL handshake timeouts.

codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketClientHandshaker.java
```diff
@@ -431,7 +485,50 @@ public ChannelFuture close(Channel channel, CloseWebSocketFrame frame, ChannelPr
...
+                    // If flush operation failed, there is no reason to expect
+                    // a server to receive CloseFrame. Thus this should be handled
+                    // by the application separately.
+                    return;
+                }
+
+                if (!forceCloseInit.compareAndSet(false, true)) {
+                    // Close might be called twice from different threads.
+                    return;
+                }
```
slandelle: nit: more of a coding style preference, but I'd rather have one single boolean operation wrapping the method's body: 
`if (future.isSuccess() && channel.isActive() && forceCloseInit.compareAndSet(false, true))`

kachayev:
@trustin @slandelle Pushed changes, thanks for the review!

normanmaurer:
@trustin PTAL again

normanmaurer:
@netty-bot test this please

normanmaurer:
@kachayev thanks a lot!

PR #8899, opened 2019-02-28
Correctly resume wrap / unwrap when SslTask execution completes

normanmaurer:
Motivation:

fa6a8cb09c9679468a6c2d912ddfbbe885ee0c08 introduced correct dispatching of delegated tasks for SSLEngine but did not correctly handle some cases for resuming wrap / unwrap after the task was executed. This could lead to stales, which showed up during tests when running with Java11 and BoringSSL.

Modifications:

- Correctly resume wrap / unwrap in all cases.
- Fix timeout value which was changed in previous commit by mistake.

Result:

No more stales after task execution.


normanmaurer:
test failure was a network problem on the CI... let me trigger a test run again

normanmaurer:
@netty-bot test this please

PR #8900, opened 2019-02-28
Adjust tests to be able to build / test when using IBM J9 / OpenJ9

normanmaurer:
Motivation:

We should run a CI job using J9 to ensure netty also works when using different JVMs.

Modifications:

- Adjust PooledByteBufAllocatorTest to be able to complete faster when using a JVM which takes longer when joining Threads (this seems to be the case with J9).
- Skip UDT tests on J9 as UDT is not supported there.

Result:

Be able to run CI against J9.

### DIFF
--- a/buffer/src/test/java/io/netty/buffer/PooledByteBufAllocatorTest.java
+++ b/buffer/src/test/java/io/netty/buffer/PooledByteBufAllocatorTest.java
@@ -430,8 +430,14 @@ public class PooledByteBufAllocatorTest extends AbstractByteBufAllocatorTest<Poo
                 Thread.sleep(100);
             }
         } finally {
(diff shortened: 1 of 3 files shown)


buffer/src/test/java/io/netty/buffer/PooledByteBufAllocatorTest.java
```diff
@@ -430,8 +430,14 @@ public void testConcurrentUsage() throws Throwable {
...
             }
         } finally {
+            // First mark all AllocationThreads to complete their work and then wait until these are complete
+            // and rethrow if there was any error.
             for (AllocationThread t : threads) {
-                t.finish();
+                t.markAsFinished();
+            }
+
+            for (AllocationThread t: threads) {
```
carl-mastrangelo: nit: space after `:`

PR #6873, opened 2017-06-20
Add a default mapping from EventLoopGroup and ChannelFactory #6839

eolivelli:
Introduce a EventLoopGroup#defaultChannelFactory as an hint to build a ChannelFactory if not configured.
This way the Channel type will be automatically discovered for simple use-cases

Motivation:

see issue #6839 

Modification:

I have added a EventLoopGroup#defaultChannelFactory(boolean serverSide) method to let the EventLoopGroup describe the usual type of Channel it handles.

This way code will be simpler:

before the change:
```
 ServerBootstrap b = new ServerBootstrap();
            b.group(bossGroup, workerGroup)
                .channel(NetworkUtils.isEnableEpoolNative() ? EpollServerSocketChannel.class : NioServerSocketChannel.class)
                .childHandler(channelInitialized)
                .option(ChannelOption.SO_BACKLOG, 128);

```
After the change:
```
 ServerBootstrap b = new ServerBootstrap();
            b.group(bossGroup, workerGroup)
                .childHandler(channelInitialized)
                .option(ChannelOption.SO_BACKLOG, 128);

```

Result:

Fixes #6839 

I have some questions:
- I needed to change the 'validate' method in order to create the channelFactory if not given, this is not very good for me but I did not find another way
- this PR does not include new test cases, I will add them as soon as the general approach to the problem wil be approved (at the moment I have tested all netty based projects and all is running as expected)

PR #8901, opened 2019-02-28
Added ChannelSystem to easier initialize Bootstrap

SplotyCode:
Motivation:
Creating Bootstraps is harder than it must. Many people don't use EPOLL because they dont't know about it or don't want to check it EPOLL is available or if KQUEUE is available. But it also makes it easier to change the Bootstrap to a specific channel type.

Modification:
Added the ChannelSystem enum and Bootstrap#setChannelSystem(ChannelSystem).

Result:
This should make working with netty mush easier for simple projects and big projects. It should also improve performance because you don't need to change LoopGroup and ChannelType based on environment. 

This PR should not break anything and as 100% backward compatibility.
I didn't add KQUEUE but I will add it if you like the basic idea.

The ChannelSystem enum would be allot easier if we could use transport-native.

### DIFF
--- a/transport/src/main/java/io/netty/bootstrap/ChannelSystem.java
+++ b/transport/src/main/java/io/netty/bootstrap/ChannelSystem.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2012 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.bootstrap;
+
+import io.netty.channel.Channel;
+import io.netty.channel.EventLoopGroup;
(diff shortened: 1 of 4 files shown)


SplotyCode:
This would look a lot cleaner if transport-native could be imported. I could make an Interface that has the same methods as the enum and then implement the interface from the enum then we could move the enum to an other Module and use the interface in Bootstrap.

Bennett-Lynch:
I think that you're right to say that the native transport support is too difficult for users to discover.  It should be more obvious and intuitive.  Many people don't use something like EPOLL simply because they don't know that they should.  I, as a user, should not have to write my own `if EPOLL avail use EPOLL, elsif KQUEUE avail use KQUEUE, els NIO`, etc. logic.  If there's not much progress on this PR, I would at least recommend raising an issue so that people can discuss about a better approach.

eolivelli:
This patch is related 

https://github.com/netty/netty/pull/6873

transport/src/main/java/io/netty/bootstrap/ChannelSystem.java
```diff
@@ -0,0 +1,102 @@
...
+
+    private Class<? extends Channel> channelClass;
+    private Class<? extends ServerChannel> serverChannelClass;
+
+    ChannelSystem(Class<? extends Channel> channelClass, Class<? extends ServerChannel> serverChannelClass) {
+        this.channelClass = channelClass;
+        this.serverChannelClass = serverChannelClass;
+    }
+
+    public abstract EventLoopGroup newLoopGroup();
```
blueswhisper: No entry for specified thread count of `EventLoopGroup` ?

normanmaurer:
We will need to tackle this somehow for netty 5 but I think for netty 4.x we will just leave it as it is for now... Closing.

PR #8902, opened 2019-03-01
Optimize Hpack and AsciiString hashcode and equals

njhill:
Motivation:

While looking at hpack header-processing hotspots I noticed some low level too-big-to-inline methods which can be shrunk.

Modifications:

Reduce bytecode size and/or runtime operations used for the following methods:

- `PlatformDependent0.equals(byte[], ...)`
- `PlatformDependent0.equalsConstantTime(byte[], ...)`
- `PlatformDependent0.hashCodeAscii(byte[],int,int)`
- `PlatformDependent.hashCodeAscii(CharSequence)`

Result:

Existing benchmarks show decent improvement

Before
```
Benchmark                     (size)   Mode  Cnt         Score         Error  Units
HpackUtilBenchmark.newEquals   SMALL  thrpt    5  17200229.374 ± 1701239.198  ops/s
HpackUtilBenchmark.newEquals  MEDIUM  thrpt    5   3386061.629 ±   72264.685  ops/s
HpackUtilBenchmark.newEquals   LARGE  thrpt    5    507579.209 ±   65883.951  ops/s
```

After
```
Benchmark                     (size)   Mode  Cnt         Score         Error  Units
HpackUtilBenchmark.newEquals   SMALL  thrpt    5  29221527.058 ± 4805825.836  ops/s
HpackUtilBenchmark.newEquals  MEDIUM  thrpt    5   6556251.645 ±  466115.199  ops/s
HpackUtilBenchmark.newEquals   LARGE  thrpt    5    879828.889 ±  148136.641  ops/s
```

Before
```
Benchmark                          (size)  Mode  Cnt     Score     Error  Units
PlatformDepBench.unsafeBytesEqual       4  avgt   10     4.263 ±   0.110  ns/op
PlatformDepBench.unsafeBytesEqual      10  avgt   10     5.206 ±   0.133  ns/op
PlatformDepBench.unsafeBytesEqual      50  avgt   10     8.160 ±   0.320  ns/op
PlatformDepBench.unsafeBytesEqual     100  avgt   10    13.810 ±   0.751  ns/op
PlatformDepBench.unsafeBytesEqual    1000  avgt   10    89.077 ±   7.275  ns/op
PlatformDepBench.unsafeBytesEqual   10000  avgt   10   773.940 ±  24.579  ns/op
PlatformDepBench.unsafeBytesEqual  100000  avgt   10  7546.807 ± 110.395  ns/op
```

After
```
Benchmark                          (size)  Mode  Cnt     Score     Error  Units
PlatformDepBench.unsafeBytesEqual       4  avgt   10     3.337 ±   0.087  ns/op
PlatformDepBench.unsafeBytesEqual      10  avgt   10     4.286 ±   0.194  ns/op
PlatformDepBench.unsafeBytesEqual      50  avgt   10     7.817 ±   0.123  ns/op
PlatformDepBench.unsafeBytesEqual     100  avgt   10    11.260 ±   0.412  ns/op
PlatformDepBench.unsafeBytesEqual    1000  avgt   10    84.255 ±   2.596  ns/op
PlatformDepBench.unsafeBytesEqual   10000  avgt   10   591.892 ±   5.136  ns/op
PlatformDepBench.unsafeBytesEqual  100000  avgt   10  6978.859 ± 285.043  ns/op
```

I left out the `AsciiStringBenchmark` results for brevity but they show ~5-10% improvement, depending on length.

### DIFF
--- a/common/src/main/java/io/netty/util/internal/PlatformDependent.java
+++ b/common/src/main/java/io/netty/util/internal/PlatformDependent.java
@@ -754,82 +754,41 @@ public final class PlatformDependent {
      */
     public static int hashCodeAscii(CharSequence bytes) {
         int hash = HASH_CODE_ASCII_SEED;
-        final int remainingBytes = bytes.length() & 7;
+        final int length = bytes.length(), remainingBytes = length & 7;
         // Benchmarking shows that by just naively looping for inputs 8~31 bytes long we incur a relatively large
         // performance penalty (only achieve about 60% performance of loop which iterates over each char). So because
         // of this we take special provisions to unroll the looping for these conditions.
-        switch (bytes.length()) {
-            case 31:
-            case 30:
-            case 29:
-            case 28:
-            case 27:
-            case 26:
-            case 25:
-            case 24:
-                hash = hashCodeAsciiCompute(bytes, bytes.length() - 24,
-                        hashCodeAsciiCompute(bytes, bytes.length() - 16,
-                          hashCodeAsciiCompute(bytes, bytes.length() - 8, hash)));
-                break;
-            case 23:
-            case 22:
-            case 21:
-            case 20:
-            case 19:
-            case 18:
-            case 17:
-            case 16:
-                hash = hashCodeAsciiCompute(bytes, bytes.length() - 16,
-                         hashCodeAsciiCompute(bytes, bytes.length() - 8, hash));
-                break;
-            case 15:
-            case 14:
-            case 13:
-            case 12:
-            case 11:
-            case 10:
-            case 9:
-            case 8:
-                hash = hashCodeAsciiCompute(bytes, bytes.length() - 8, hash);
-                break;
-            case 7:
-            case 6:
-            case 5:
-            case 4:
-            case 3:
-            case 2:
-            case 1:
-            case 0:
-                break;
-            default:
-                for (int i = bytes.length() - 8; i >= remainingBytes; i -= 8) {
-                    hash = hashCodeAsciiCompute(bytes, i, hash);
+        if (length >= 32) {
+            for (int i = length - 8; i >= remainingBytes; i -= 8) {
+                hash = hashCodeAsciiCompute(bytes, i, hash);
+            }
+        } else if (length >= 8) {
+            hash = hashCodeAsciiCompute(bytes, length - 8, hash);
+            if (length >= 16) {
+                hash = hashCodeAsciiCompute(bytes, length - 16, hash);
+                if (length >= 24) {
+                    hash = hashCodeAsciiCompute(bytes, length - 24, hash);
                 }
-                break;
+            }
         }
-        switch(remainingBytes) {
-            case 7:
-                return ((hash * HASH_CODE_C1 + hashCodeAsciiSanitizeByte(bytes.charAt(0)))
-                              * HASH_CODE_C2 + hashCodeAsciiSanitizeShort(bytes, 1))
-                              * HASH_CODE_C1 + hashCodeAsciiSanitizeInt(bytes, 3);
-            case 6:
-                return (hash * HASH_CODE_C1 + hashCodeAsciiSanitizeShort(bytes, 0))
-                             * HASH_CODE_C2 + hashCodeAsciiSanitizeInt(bytes, 2);
-            case 5:
-                return (hash * HASH_CODE_C1 + hashCodeAsciiSanitizeByte(bytes.charAt(0)))
-                             * HASH_CODE_C2 + hashCodeAsciiSanitizeInt(bytes, 1);
-            case 4:
-                return hash * HASH_CODE_C1 + hashCodeAsciiSanitizeInt(bytes, 0);
-            case 3:
-                return (hash * HASH_CODE_C1 + hashCodeAsciiSanitizeByte(bytes.charAt(0)))
-                             * HASH_CODE_C2 + hashCodeAsciiSanitizeShort(bytes, 1);
-            case 2:
-                return hash * HASH_CODE_C1 + hashCodeAsciiSanitizeShort(bytes, 0);
-            case 1:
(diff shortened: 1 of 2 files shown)


normanmaurer:
@netty-bot test this please

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -754,82 +754,41 @@ public static int hashCodeAscii(byte[] bytes, int startPos, int length) {
      */
     public static int hashCodeAscii(CharSequence bytes) {
         int hash = HASH_CODE_ASCII_SEED;
-        final int remainingBytes = bytes.length() & 7;
+        final int length = bytes.length(), remainingBytes = length & 7;
         // Benchmarking shows that by just naively looping for inputs 8~31 bytes long we incur a relatively large
         // performance penalty (only achieve about 60% performance of loop which iterates over each char). So because
         // of this we take special provisions to unroll the looping for these conditions.
```
normanmaurer: @njhill is this comment still true ?

njhill: Not sure, I can test it...

normanmaurer: would be nice 👍

njhill: @normanmaurer interesting, it does appear to still be the case.

This PR as-is:
```
AsciiStringBenchmark.hashCodeBenchCharSequenceNew       8  thrpt    5   84938768.167 ±  3807502.233  ops/s
AsciiStringBenchmark.hashCodeBenchCharSequenceNew      10  thrpt    5   75864099.463 ±  3345939.549  ops/s
```

Without 8-31 bytes special case:
```
AsciiStringBenchmark.hashCodeBenchCharSequenceNew       8  thrpt    5   55068716.482 ±  8293068.816  ops/s
AsciiStringBenchmark.hashCodeBenchCharSequenceNew      10  thrpt    5   52594830.642 ±  4131819.596  ops/s
```

normanmaurer: @njhill thanks for testing :)

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
```diff
@@ -622,72 +622,57 @@ static void setMemory(Object o, long offset, long bytes, byte value) {
...
         int remainingBytes = length & 7;
-        final long end = baseOffset1 + remainingBytes;
-        for (long i = baseOffset1 - 8 + length, j = baseOffset2 - 8 + length; i >= end; i -= 8, j -= 8) {
-            if (UNSAFE.getLong(bytes1, i) != UNSAFE.getLong(bytes2, j)) {
-                return false;
+        final long baseOffset1 = BYTE_ARRAY_BASE_OFFSET + startPos1;
+        final long diff = startPos2 - startPos1;
+        if (length > 7) {
+            final long end = baseOffset1 + remainingBytes;
+            for (long i = baseOffset1 - 8 + length; i >= end; i -= 8) {
```
carl-mastrangelo: @normanmaurer IIRC Java aligns objects on word boundaries, which would imply arrays also start on word boundaries.  Doesn't this imply that iterating from a likely odd number offset backwards would result in torn reads?  Im surprised this isn't starting at the beginning.

common/src/main/java/io/netty/util/internal/PlatformDependent.java
```diff
@@ -754,82 +754,41 @@ public static int hashCodeAscii(byte[] bytes, int startPos, int length) {
      */
     public static int hashCodeAscii(CharSequence bytes) {
         int hash = HASH_CODE_ASCII_SEED;
-        final int remainingBytes = bytes.length() & 7;
+        final int length = bytes.length(), remainingBytes = length & 7;
```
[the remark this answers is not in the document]
normanmaurer: I agree with @carl-mastrangelo

common/src/main/java/io/netty/util/internal/PlatformDependent0.java
normanmaurer: good point. @njhill mind checking ?

njhill: @carl-mastrangelo I just left that aspect as it was before, and assumed the reason in this particular method at least was that statistically strings are more likely to differ near the end than the start (common prefixes). That consideration wouldn't apply to the other methods though which always cover all the bytes.

Also note that this is comparing arbitrary regions so the start might not align with a word boundary either, and even if it did it might not for the destination array. And presumably `BYTE_ARRAY_BASE_OFFSET` would need to be taken into account too.

So I guess it would need some experiments and if there was a big win for aligning the long reads in this way we could consider special-casing very large inputs for the other methods. But I'm not sure how common large inputs are in general since I think this is mainly used for headers?

carl-mastrangelo: > I think this is mainly used for headers?

Well, at least for gRPC, I think we use the constant time one usually, because we can't tell if a header is sensitive (like an auth header).  

> Also note that this is comparing arbitrary regions so the start might not align with a word boundary either.

Yeah, I considered the same thing, but my gut says pretty much everyone starts at 0.   Since this function is public and not specific to Hpack, we don't know the scope of usage.   You are probably right that it doesn't matter, I was jut expressing surprise for the original structure.

normanmaurer:
@netty-bot test this please

normanmaurer:
@carl-mastrangelo any more comments or you are happy now ?

normanmaurer:
@netty-bot test this please

normanmaurer:
Also @ejona86

normanmaurer:
@njhill thanks a lot!

PR #8903, opened 2019-03-01
Add docker-compose config to run build with OpenJ9 JVM

normanmaurer:
Motivation:

To ensure Netty works on different JVMs we should also run tests on the CI with these.

Modifications:

Add docker-compose config to run build with OpenJ9 JVM

Result:

Ensure Netty works with different JVMs


normanmaurer:
Let me merge this as its just a docker-compose file :)

PR #8904, opened 2019-03-01
Use maven plugin to prevent API/ABI breakage as part of build process

normanmaurer:
Motivation:

Netty is very widely used which can lead to a lot of pain when we break API / ABI. We should make use japicmp-maven-plugin during the build to verify we do not introduce breakage by mistake.

Modifications:

- Add japicmp-maven-plugin to the build process
- Fix a method signature change in HttpProxyHandler that was flagged as a possible problem.

Result:

Ensure no API/ABI breakage accour between releases.

### DIFF
--- a/codec-http/pom.xml
+++ b/codec-http/pom.xml
@@ -57,7 +57,6 @@
       <groupId>${project.groupId}</groupId>
       <artifactId>netty-handler</artifactId>
       <version>${project.version}</version>
-      <optional>true</optional>
     </dependency>
     <dependency>
       <groupId>com.jcraft</groupId>
--- a/handler-proxy/src/main/java/io/netty/handler/proxy/HttpProxyHandler.java
+++ b/handler-proxy/src/main/java/io/netty/handler/proxy/HttpProxyHandler.java
@@ -172,7 +172,7 @@ public final class HttpProxyHandler extends ProxyHandler {
     }
 
     @Override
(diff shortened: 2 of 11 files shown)


codec-http/pom.xml
```diff
@@ -57,7 +57,6 @@
       <groupId>${project.groupId}</groupId>
       <artifactId>netty-handler</artifactId>
       <version>${project.version}</version>
-      <optional>true</optional>
```
normanmaurer: this was incorrect as it is not really an optional dependency.

handler-proxy/src/main/java/io/netty/handler/proxy/HttpProxyHandler.java
```diff
@@ -172,7 +172,7 @@ protected Object newInitialMessage(ChannelHandlerContext ctx) throws Exception {
     }
 
     @Override
-    protected boolean handleResponse(ChannelHandlerContext ctx, Object response) throws HttpProxyConnectException {
+    protected boolean handleResponse(ChannelHandlerContext ctx, Object response) throws Exception {
```
normanmaurer: This was flagged as a breakage so I fixed it. That said I think this is a false positive as `HttpProxyHandler` is final so there should be no problem here in reality.

normanmaurer:
@ejona86 I will open an issue for it and hopefully they can fix it

PR #8906, opened 2019-03-01
Contribution

### LINKED ISSUE #8855: Support semicolon as a normal character in URI (no longer a parameter separator)
### Expected result
According to the 2014 W3C Recommendation, semicolon is now illegal as a parameter separator in a http URI.
`?foo=bar;baz` means the parameter `foo` will have the value `bar;baz`

see https://www.w3.org/TR/2014/REC-html5-20141028/forms.html#url-encoded-form-data
(thanks to https://stackoverflow.com/questions/3481664/semicolon-as-url-query-separator)

### Current result
The URI decoder now supports the obsolete 1999 W3C recommendation
 `?foo=bar;baz` is decoded as `?foo=bar & baz=`

(see QueryStringDecoder.java, line 230 :  `case '&': case ';':` both characters are handled equally)

### Netty version
tested (through vert.x) on 1.4.19

1.8.0_60

Windows i686-pc Intel

_Of course this issue is linked to  #3044 and #2896, but as the norm changed, I suppose it's better to open a new issue_

marcban:
Motivation:

correction for #8855, ";" should be treated as a normal character in query parameters

Modification:

deleted specific for ";" in QueryStringDecoder, updated tests

Result:

Fixes #8855.


normanmaurer:
there is something wrong... ensure you are on the latest commit of 4.1 branch

PR #8907, opened 2019-03-01
Support semicolons in query parameters (correction for #8855)

### LINKED ISSUE #8855: Support semicolon as a normal character in URI (no longer a parameter separator)

marcban:
Motivation:

Support semicolons in query parameters

Modification:

deleted specific case for ";" in QueryStringDecoder, corrected test

Result:

Fixes #8855

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java
@@ -227,7 +227,6 @@ public class QueryStringDecoder {
                 }
                 break;
             case '&':
-            case ';':
                 if (addParam(s, nameStart, valueStart, i, params, charset)) {
                     paramsLimit--;
                     if (paramsLimit == 0) {
--- a/codec-http/src/test/java/io/netty/handler/codec/http/QueryStringDecoderTest.java
+++ b/codec-http/src/test/java/io/netty/handler/codec/http/QueryStringDecoderTest.java
@@ -127,7 +127,8 @@ public class QueryStringDecoderTest {
         assertQueryString("/foo?a=", "/foo?&a=");
         assertQueryString("/foo?a=b&c=d", "/foo?a=b&c=d");
         assertQueryString("/foo?a=1&a=&a=", "/foo?a=1&a&a=");
-    }
+        // ";" should be treated as a normal character, see #8855
+        assertQueryString("/foo?a=1;2", "/foo?a=1%3B2");    }
 
     @Test
     public void testPathSpecific() {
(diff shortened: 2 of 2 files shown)


normanmaurer:
@netty-bot test this please

ejona86:
Looking at the issue, this seems like a change in algorithm. I don't understand how this can be considered safe, since it could be rightly depended on if someone is using the previous encoder.

Note that #8855 only linked to the _encoding_ algorithm. That is not the same as a _decoding_ algorithm, which has to take into account previous encoding algorithms for compatibility.

marcban:
- #8855 refers to a change in the W3C recommendation from 1999 to 2014 
- it concerns the decoding of query strings by the server, hence the change in QueryStringDecoder
- other servers (IIS, Apache, etc.) already implement this

ejona86:
I'm sorry. I now see that there is a decoding algorithm later in the section. I skimmed and searched for "decoding" (to mirror the "encoding algorithm" used earlier), but they just said "to decode...".

For a new implementation, I do think I would follow their recommendation. It's not clear how it impacts existing implementations though. Netty will be decoding parameters generated from an assortment of sources, and it would seem that changing the behavior of `;` could break some of them. Most of the time I would be happy to change the behavior to align to the standard, but this is a case where the standard changed in an incompatible way.

I'd be fine with an option, or a new method, or changing the behavior in Netty 5. But as has been made clearly evident with things like URLs, just because the standard says something doesn't mean it is actually compatible.

normanmaurer:
@ejona86 good point... maybe we should make it a constructor arg ?

normanmaurer:
@marcban as mentioned can you please add a new constructor arg to the `QueryStringDecoder` which allows to change how decoding is done (using the old way as default).

normanmaurer:
@ejona86 would this make you more comfortable ^^ ?

ejona86:
I'd have no reservation with a way to opt-in to the new behavior.

normanmaurer:
@marcb can you please update your PR to use the constructor flag ?

normanmaurer:
Closing this as the author did not respond anymore and the ICLA was not signed so there is no way I can finish it

marcban:
I'm very sorry, I realize I lack some elementary skills to work on open source projects, like if I should open another branch, how to choose parameter names, how many constructors to implement for this class, etc. I don't even know what an ICLA is and how to sign it... At the end I duplicated the modified method in my project, ignoring Vertx/Netty standard method for reading URI parameters

PR #8914, opened 2019-03-04
Carefully manage Keep-Alive connections in HttpStaticFileServer

kachayev:
Motivation:

Simple rules:

* close the connection when sending any error
* specify "Connection: close" header when closing the connection
* successful responses should keep the connection intact when otherwise is not requested by the client

Modifications:

* "send a response and clean up the connection" logic moved to a helper
* for all successful responses set correct "Content-Lenght" header
* do not specify "Connection: Keep-Alive" header as far it's a default for HTTP/1.1
* set "Connection: close" header when necessary

Result:

Keep-Alive connections management conforms with RFCs.

### DIFF
--- a/example/src/main/java/io/netty/example/http/file/HttpStaticFileServerHandler.java
+++ b/example/src/main/java/io/netty/example/http/file/HttpStaticFileServerHandler.java
@@ -122,6 +122,7 @@ public class HttpStaticFileServerHandler extends SimpleChannelInboundHandler<Ful
             return;
         }
 
+        final boolean keepAlive = HttpUtil.isKeepAlive(request);
         final String uri = request.uri();
         final String path = sanitizeUri(uri);
         if (path == null) {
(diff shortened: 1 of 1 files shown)


example/src/main/java/io/netty/example/http/file/HttpStaticFileServerHandler.java
```diff
@@ -331,12 +330,33 @@ private static void sendError(ChannelHandlerContext ctx, HttpResponseStatus stat
...
+                                                 boolean keepAlive) {
+        HttpUtil.setContentLength(response, response.content().readableBytes());
+        if (!keepAlive) {
+            // We're going to close the connection as soon as the response is sent,
+            // so we should also make it clear for the client.
+            response.headers().set(HttpHeaderNames.CONNECTION, HttpHeaderValues.CLOSE);
+        }
+
+        ctx.write(response);
+        ChannelFuture lastWrite = ctx.writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT);
```
normanmaurer: @kachayev this is not correct as you already write a `FullHttpResonse` and so MUST NOT write an extra `LastHttpContent`

PR #8918, opened 2019-03-05
ReferenceCountedOpenSslEngines SSLSession must provide local certific…

normanmaurer:
…ates during certificate verification

Motivation:

The SSLSession that is returned by SSLEngine.getHandshakeSession() must be able to provide the local certificates when the TrustManager is invoked on the server-side.

Modifications:

- Correctly return the local certificates
- Add unit test

Result:

Be able to obtain local certificates from handshake SSLSession during verification on the server side.

### DIFF
--- a/handler/src/test/java/io/netty/handler/ssl/JdkConscryptSslEngineInteropTest.java
+++ b/handler/src/test/java/io/netty/handler/ssl/JdkConscryptSslEngineInteropTest.java
@@ -90,5 +90,6 @@ public class JdkConscryptSslEngineInteropTest extends SSLEngineTest {
     @Override
     public void testHandshakeSession() throws Exception {
         // Ignore as Conscrypt does not correctly return the local certificates while the TrustManager is invoked.
+        // See https://github.com/google/conscrypt/issues/634
     }
 }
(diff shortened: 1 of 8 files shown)


handler/src/test/java/io/netty/handler/ssl/JdkConscryptSslEngineInteropTest.java
```diff
@@ -85,4 +85,10 @@ protected boolean mySetupMutualAuthServerIsValidClientException(Throwable cause)
         // TODO(scott): work around for a JDK issue. The exception should be SSLHandshakeException.
         return super.mySetupMutualAuthServerIsValidClientException(cause) || causedBySSLException(cause);
     }
+
+    @Ignore("Ignore due bug in Conscrypt")
+    @Override
+    public void testHandshakeSession() throws Exception {
```
normanmaurer: @ejona86 @flooey This also needs to be fixed in conscrypt.

flooey: @normanmaurer Can you be more specific about what behavior you see in Conscrypt and what you think it should be doing?

normanmaurer: Basically what this PR fixes. 

session.getLocalCertificates() and getLocalPrincipal() should not return null in the Trustmanager serverside. Check the included unit test

flooey: Oh, on the server.  Sure, makes sense.  Weird that we don't do that, I've filed https://github.com/google/conscrypt/issues/634 to track fixing it.

PR #8919, opened 2019-03-05
Add interopt tests between Conscrypt and OpenSSL SSLEngine implementa…

normanmaurer:
…tions.

Motivation:

In the past we found a lot of SSL related bugs because of the interopt tests we have in place between different SSLEngine implementations. We should have as many of these interopt tests as possible for this reason.

Modifications:

- Add interopt tests between Conscrypt and OpenSSL SSLEngine implementations

Result:

More tests for SSL.


normanmaurer:
Let me merge this as its just extra tests (which are basically a copy of existing classes).

PR #8921, opened 2019-03-06
Do not use GetPrimitiveArrayCritical(...) due multiple not-fixed bugs…

normanmaurer:
… related to GCLocker

Motivation:

GetPrimitiveArrayCritical(...) may cause multiple not-fixed bugs related to the GCLocker while there is little gain for our use-case. We should just use GetByteArrayRegion(...) and copy into a small on-stack buffer.

See also:

- https://shipilev.net/jvm/anatomy-quarks/9-jni-critical-gclocker/#_g1
- https://bugs.openjdk.java.net/browse/JDK-8048556
- https://bugs.openjdk.java.net/browse/JDK-8057573
- https://bugs.openjdk.java.net/browse/JDK-8057586

Special thanks to @jayv @shipilev @apangin for the pointers.

Modifications:

Replace GetPrimitiveArrayCritical(...) with GetByteArrayRegion(...)

### DIFF
--- a/transport-native-unix-common/src/main/c/netty_unix_socket.c
+++ b/transport-native-unix-common/src/main/c/netty_unix_socket.c
@@ -232,16 +232,20 @@ static jint _socket(JNIEnv* env, jclass clazz, int type) {
 int netty_unix_socket_initSockaddr(JNIEnv* env, jbyteArray address, jint scopeId, jint jport,
                                    const struct sockaddr_storage* addr, socklen_t* addrSize) {
     uint16_t port = htons((uint16_t) jport);
+    // We use 16 bytes as this allows us to fit ipv6, ipv4 and ipv4 mapped ipv6 addresses in the array.
+    jbyte addressBytes[16];
 
-    // Use GetPrimitiveArrayCritical and ReleasePrimitiveArrayCritical to signal the VM that we really would like
-    // to not do a memory copy here. This is ok as we not do any blocking action here anyway.
-    // This is important as the VM may suspend GC for the time!
-    jbyte* addressBytes = (*env)->GetPrimitiveArrayCritical(env, address, 0);
-    if (addressBytes == NULL) {
-        // No memory left ?!?!?
-        netty_unix_errors_throwOutOfMemoryError(env);
-        return -1;
-    }
+    int len = (*env)->GetArrayLength(env, address);
+
+    // We use GetByteArrayRegion(...) and copy into a small stack allocated buffer and NOT GetPrimitiveArrayCritical(...)
+    // as there are still multiple GCLocker related bugs which are not fixed yet.
+    //
+    // For example:
+    //     https://bugs.openjdk.java.net/browse/JDK-8048556
+    //     https://bugs.openjdk.java.net/browse/JDK-8057573
+    //     https://bugs.openjdk.java.net/browse/JDK-8057586
+    (*env)->GetByteArrayRegion(env, address, 0, len, addressBytes);
+
     if (socketType == AF_INET6) {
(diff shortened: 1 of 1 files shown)


transport-native-unix-common/src/main/c/netty_unix_socket.c
```diff
@@ -232,16 +232,20 @@ static jint _socket(JNIEnv* env, jclass clazz, int type) {
 int netty_unix_socket_initSockaddr(JNIEnv* env, jbyteArray address, jint scopeId, jint jport,
                                    const struct sockaddr_storage* addr, socklen_t* addrSize) {
     uint16_t port = htons((uint16_t) jport);
+    // 24 is the maximum possible
```
johnou: 128 bit (16 byte) for IPv6 address plus long port 64 bit (8 byte)?

normanmaurer: actually we only need 16 here, so I changed it and added a comment. Good catch!

normanmaurer:
@netty-bot test this please

transport-native-unix-common/src/main/c/netty_unix_socket.c
```diff
@@ -232,16 +232,20 @@ static jint _socket(JNIEnv* env, jclass clazz, int type) {
...
+    int len = (*env)->GetArrayLength(env, address);
+
+    // We use GetByteArrayRegion(...) and copy into a small stack allocated buffer and NOT GetPrimitiveArrayCritical(...)
+    // as there are still multiple GCLocker related bugs which are not fixed yet.
+    //
+    // For example:
+    //     https://bugs.openjdk.java.net/browse/JDK-8048556
+    //     https://bugs.openjdk.java.net/browse/JDK-8057573
+    //     https://bugs.openjdk.java.net/browse/JDK-8057586
+    (*env)->GetByteArrayRegion(env, address, 0, len, addressBytes);
```
ejona86: Seems we should do min(len, 16) or have an assert. Yes, we know it should be 16 or less, but I don't like playing with fire.

johnou: Or just 16? The code which reads the array below assumes that's always the case.

normanmaurer: let me just add a guard for now... can't hurt.

transport-native-unix-common/src/main/c/netty_unix_socket.c
```diff
@@ -232,16 +232,25 @@ static jint _socket(JNIEnv* env, jclass clazz, int type) {
...
-    // Use GetPrimitiveArrayCritical and ReleasePrimitiveArrayCritical to signal the VM that we really would like
-    // to not do a memory copy here. This is ok as we not do any blocking action here anyway.
-    // This is important as the VM may suspend GC for the time!
-    jbyte* addressBytes = (*env)->GetPrimitiveArrayCritical(env, address, 0);
-    if (addressBytes == NULL) {
-        // No memory left ?!?!?
-        netty_unix_errors_throwOutOfMemoryError(env);
+    int len = (*env)->GetArrayLength(env, address);
+
+    if (len > 16) {
```
johnou: I think it should be != 16 or it's going to segfault..

johnou: ah not if it's AF_INET4.

Hamlin-Li:
I hope [Region Pinning in JDK](https://bugs.openjdk.java.net/browse/JDK-8276094) may help eliminate or alleviate this problem in the future.

PR #8922, opened 2019-03-06
HttpContentDecoder must continue read when it did not produce any mes…

normanmaurer:
…sage and auto read is false

Motivation:

When HttpContentDecoder (and so HttpContentDecompressor) does not produce any message we need to make sure it calls ctx.read() if auto read is false to not stale.

Modifications:

- Keep track if we need to call ctx.read() or not
- Add unit test

Result:

Fixes https://github.com/netty/netty/issues/8915.

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentDecoder.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentDecoder.java
@@ -51,6 +51,7 @@ public abstract class HttpContentDecoder extends MessageToMessageDecoder<HttpObj
     protected ChannelHandlerContext ctx;
     private EmbeddedChannel decoder;
     private boolean continueResponse;
+    private boolean needRead;
 
     @Override
     protected void decode(ChannelHandlerContext ctx, HttpObject msg, List<Object> out) throws Exception {
@@ -145,7 +146,13 @@ public abstract class HttpContentDecoder extends MessageToMessageDecoder<HttpObj
             if (decoder == null) {
                 out.add(c.retain());
             } else {
-                decodeContent(c, out);
+                try {
+                    decodeContent(c, out);
+                } finally {
+                    if (!needRead) {
+                        needRead = out.isEmpty();
+                    }
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


codec-http/src/main/java/io/netty/handler/codec/http/HttpContentDecoder.java
```diff
@@ -145,7 +146,13 @@ protected void decode(ChannelHandlerContext ctx, HttpObject msg, List<Object> ou
             if (decoder == null) {
                 out.add(c.retain());
             } else {
-                decodeContent(c, out);
+                try {
+                    decodeContent(c, out);
+                } finally {
+                    if (!needRead) {
```
ejona86: I think this variable handling is inverted. `needRead` should start as `true`, and if `!out.isEmpty()` it should be set to `false`. Then in channelReadComplete() it should be set to `true` again.

If a read() causes three ByteBufs to be read (because using DefaultMaxBytesRecvByteBufAllocator), which calls decode() three times, the first two times may not put anything in `out` but the last decode does place something in `out`. It seems in that case we shouldn't call read.

But there's also the case where the _second_ bytebuf placed something in `out`. That seems like it shouldn't call read either.

normanmaurer:
@ejona86 its not mark of `MessageToMessageDecoder` because we messed in the past and allowed it to be `@Sharable`. In the next major release we should fix this but there is nothing we can do there for now :(

codec-http/src/main/java/io/netty/handler/codec/http/HttpContentDecoder.java
normanmaurer: @ejona86 sure why not.

andreisilviudragnea:
I can confirm too that #8915 is fixed by these changes. I have also understood now how the `Inflater` used behind the scenes by `JdkZlibDecoder` decompresses as much as it can from the input buffer, passing the output on. Initially, I thought `JdkZlibDecoder` also uses the accumulator buffer approach like the `ByteToMessageDecoder` Thank you for the quick fix! 😄

PR #8924, opened 2019-03-08
Fix HttpUtil.isKeepAlive to behave correctly when Connection is a comma separated list

violetagg:
Motivation:

According to the specification, the `Connection` header's syntax is:

```
The Connection header field's value has the following grammar:

     Connection        = 1#connection-option
     connection-option = token

Connection options are case-insensitive.
```
https://tools.ietf.org/html/rfc7230#section-6.1

This means that Connection's value can have at least one element or
a comma separated list with elements
When calculating whether the connection can remain open,
`HttpUtil.isKeepAlive(HttpMessage)` should take this into account.

Modifications:

- Check for `close` and `keep-alive` in a comma separated list
- Add unit test

Result:

`HttpUtil.isKeepAlive(HttpMessage)` works correctly when `Connection: Upgrade, close`

Related to https://github.com/reactor/reactor-netty/issues/632

### DIFF
--- a/codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
+++ b/codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
@@ -65,16 +65,9 @@ public final class HttpUtil {
      * {@link HttpVersion#isKeepAliveDefault()}.
      */
     public static boolean isKeepAlive(HttpMessage message) {
-        CharSequence connection = message.headers().get(HttpHeaderNames.CONNECTION);
-        if (HttpHeaderValues.CLOSE.contentEqualsIgnoreCase(connection)) {
-            return false;
-        }
-
-        if (message.protocolVersion().isKeepAliveDefault()) {
-            return !HttpHeaderValues.CLOSE.contentEqualsIgnoreCase(connection);
-        } else {
-            return HttpHeaderValues.KEEP_ALIVE.contentEqualsIgnoreCase(connection);
-        }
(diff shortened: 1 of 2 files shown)


normanmaurer:
@netty-bot test this please

codec-http/src/main/java/io/netty/handler/codec/http/HttpUtil.java
```diff
@@ -65,15 +65,17 @@ public static boolean isAsteriskForm(URI uri) {
...
-        if (HttpHeaderValues.CLOSE.contentEqualsIgnoreCase(connection)) {
+        boolean connectionClose = message.headers().containsValue(HttpHeaderNames.CONNECTION,
+                HttpHeaderValues.CLOSE, true);
+        if (connectionClose) {
             return false;
         }
 
         if (message.protocolVersion().isKeepAliveDefault()) {
-            return !HttpHeaderValues.CLOSE.contentEqualsIgnoreCase(connection);
+            return !connectionClose;
```
slandelle: connectionClose can only be false here

normanmaurer: @violetagg ^^ can you adjust this  ... @slandelle good catch!

violetagg: done

violetagg:
@slandelle Modified the code to become single line

slandelle:
LGTM

normanmaurer:
@violetagg thanks a lot!

PR #8929, opened 2019-03-11
Deprecate ChannelInboundHandlerAdapter and ChannelOutboundHandlerAdapter

normanmaurer:
Motivation:

As we now us java8 as minimum java version we can deprecate ChannelInboundHandlerAdapter / ChannelOutboundHandlerAdapter and just move the default implementations into the interfaces. This makes things a bit more flexible for the end-user and also simplifies the class-hierarchy.

Modifications:

- Mark ChannelInboundHandlerAdapter and ChannelOutboundHandlerAdapter as deprecated
- Add default implementations to ChannelInboundHandler / ChannelOutboundHandler
- Refactor our code to not use ChannelInboundHandlerAdapter / ChannelOutboundHandlerAdapter anymore

Result:

Cleanup class-hierarchy and make things a bit more flexible.


normanmaurer:
@franz1981 @vietj @njhill PTAL as well. I am not concerned about performance for default methods as we have the `@Skip` optimisation in place anyway.

vietj:
looks good :-)

acdebaca:
@normanmaurer 
Shouldn't [HttpContentDecoderTest.java line 549](https://github.com/netty/netty/blob/33072228efb110e2969701048e94a6832a3a4f25/codec-http/src/test/java/io/netty/handler/codec/http/HttpContentDecoderTest.java#L549) be:

```java
        EmbeddedChannel channel = new EmbeddedChannel(decoder, new ChannelInboundHandler() {
            @Override
            public void channelInactive(ChannelHandlerContext ctx) throws Exception {
                assertTrue(channelInactiveCalled.compareAndSet(false, true));
                ctx.fireChannelInactive();
            }
        });
```
?

normanmaurer:
@acdebaca good catch! Thanks

carl-mastrangelo:
@normanmaurer did the plan to move everything to ChannelHandler ever pan out?

normanmaurer:
@carl-mastrangelo I am still investigate... I thought it make sense to do it step by step in any case :)

PR #8931, opened 2019-03-11
Other ByteToMessageDecoder streamlining

njhill:
Motivation

While inspecting `ByteToMessageDecoder` during review of #8890 I noticed some other possibilities for optimization. These are independent of and should be complimentary to the changes under discussion in that PR.

Modifications

- Introduce local `STATE_REMOVED` state and use this in preference to `ctx.isRemoved()` to check for handler removal on the event loop (avoiding some volatile reads)
- Avoid allocating new buffer in `MERGE_CUMULATOR` if there is room after shifting existing live bytes
- Avoid copying bytes in `COMPOSITE_CUMULATOR` in all cases, performing a shallow copy instead when necessary; also guard against (unusual) case where input buffer is composite with writer index != capacity
- Optimize some related parts of `CompositeByteBuf` - in particular the discardReadBytes / discardReadComponents logic
- Integrate read-bytes discarding decision into cumulation logic and have it based on size of accumulated bytes - this is more of an RFC but to me makes more logical sense since in many cases the discard happens naturally (e.g. when a new buffer is used) and we are trying to guard against growth of unused bytes. It also means some fields and the counting logic can be removed. I have set the threshold to 1024 for now, maybe it could be made overridable similar to `setDiscardAfterReads` (assuming we go with this approach at all).

Result

Less copying/allocation in some existing cases, fewer volatile reads, simplified read-bytes discard logic.

### DIFF
--- a/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
@@ -449,7 +449,7 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
                 // If it was removed, it is not safe to continue to operate on the buffer.
                 //
                 // See https://github.com/netty/netty/issues/1664
-                if (ctx.isRemoved()) {
+                if (isRemoved()) {
                     break;
                 }
 
--- a/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
@@ -81,7 +81,8 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
         @Override
         public ByteBuf cumulate(ByteBufAllocator alloc, ByteBuf cumulation, ByteBuf in) {
             int discardable = cumulation.readerIndex();
-            if (cumulation.writerIndex() == discardable && !(in instanceof CompositeByteBuf)) {
+            if (cumulation.writerIndex() == discardable && in.isContiguous()) {
+                // If cumulation is empty and input buffer is contiguous, use it directly
                 cumulation.release();
                 return in;
             }
--- a/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
@@ -83,7 +83,9 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
             try {
                 int required = in.readableBytes();
                 int discardable = cumulation.readerIndex();
-                int remaining = cumulation.maxCapacity() - cumulation.writerIndex();
+                // If discardable == 0 then we allow the buffer resize itself internally (equivalent to explicit
+                // allocation + copy but cheaper)
+                int remaining = discardable > 0 ? cumulation.maxFastWritableBytes() : cumulation.maxWritableBytes();
                 if ((remaining + discardable) < required || cumulation.refCnt() > 1 || cumulation.isReadOnly()) {
                     // Expand cumulation (by replace it) when either there is not more room in the buffer
                     // or if the refCnt is greater then 1 which may happen when the user use slice().retain() or
--- a/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
@@ -255,20 +255,23 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
             decodeState = STATE_HANDLER_REMOVED_PENDING;
             return;
         }
-        ByteBuf buf = cumulation;
-        if (buf != null) {
-            // Directly set this to null so we are sure we not access it in any other method here anymore.
-            cumulation = null;
-            int readable = buf.readableBytes();
-            if (readable > 0) {
-                ctx.fireChannelRead(buf);
-                ctx.fireChannelReadComplete();
-            } else {
-                buf.release();
+        try {
+            ByteBuf buf = cumulation;
+            if (buf != null) {
+                // Directly set this to null so we are sure we not access it in any other method here anymore.
+                cumulation = null;
+                int readable = buf.readableBytes();
+                if (readable > 0) {
+                    ctx.fireChannelRead(buf);
+                    ctx.fireChannelReadComplete();
+                } else {
+                    buf.release();
+                }
             }
+            handlerRemoved0(ctx);
+        } finally {
+            decodeState = STATE_REMOVED;
         }
-        handlerRemoved0(ctx);
-        decodeState = STATE_REMOVED;
     }
 
     /**
--- a/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
+++ b/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
@@ -80,9 +80,13 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
     public static final Cumulator MERGE_CUMULATOR = new Cumulator() {
         @Override
         public ByteBuf cumulate(ByteBufAllocator alloc, ByteBuf cumulation, ByteBuf in) {
+            int discardable = cumulation.readerIndex();
+            if (cumulation.writerIndex() == discardable && !(in instanceof CompositeByteBuf)) {
+                cumulation.release();
+                return in;
+            }
             try {
                 int required = in.readableBytes();
-                int discardable = cumulation.readerIndex();
                 // If discardable == 0 then we allow the buffer resize itself internally (equivalent to explicit
                 // allocation + copy but cheaper)
                 int remaining = discardable > 0 ? cumulation.maxFastWritableBytes() : cumulation.maxWritableBytes();
@@ -118,6 +122,10 @@ public abstract class ByteToMessageDecoder extends ChannelInboundHandlerAdapter 
     public static final Cumulator COMPOSITE_CUMULATOR = new Cumulator() {
         @Override
         public ByteBuf cumulate(ByteBufAllocator alloc, ByteBuf cumulation, ByteBuf in) {
+            if (!cumulation.isReadable()) {
+                cumulation.release();
+                return in;
+            }
             CompositeByteBuf composite = null;
             try {
                 if (cumulation instanceof CompositeByteBuf && cumulation.refCnt() == 1) {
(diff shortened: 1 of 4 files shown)


buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
```diff
@@ -1675,16 +1675,28 @@ public CompositeByteBuf discardReadComponents() {
         }
 
         // Remove read components.
-        int firstComponentId = toComponentIndex0(readerIndex);
-        for (int i = 0; i < firstComponentId; i ++) {
-            components[i].free();
+        int firstComponentId = 0;
+        Component c = null;
+        for (int size = componentCount;
+                firstComponentId < size; firstComponentId ++) {
```
trustin: nit: Could fit into a single line.

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -110,49 +115,82 @@ public ByteBuf cumulate(ByteBufAllocator alloc, ByteBuf cumulation, ByteBuf in)
...
+        final int ridx = from.readerIndex();
+        final int widx = from.writerIndex();
+        final int compCount = from.numComponents();
+        int cidx = from.toComponentIndex(ridx);
+        int newReaderIdx = ridx - from.toByteIndex(cidx);
+        while (true) {
+            ByteBuf newComponent = from.internalComponent(cidx++).retain();
+            if (newReaderIdx > 0) {
+                newComponent = newComponent.slice(newReaderIdx,
+                        newComponent.readableBytes() - newReaderIdx);
```
trustin: nit: This one could fit into a single line as well. If you are concerned about readability, please consider aligning for aesthetics' sake:

```java
newComponent = newComponent.slice(newReaderIdx,
                                  newComponent.readableBytes() - newReaderIdx);
```

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -200,12 +236,12 @@ public void setCumulator(Cumulator cumulator) {
     }
 
     /**
-     * Set the number of reads after which {@link ByteBuf#discardSomeReadBytes()} are called and so free up memory.
-     * The default is {@code 16}.
+     * @deprecated discards are now performed based on accumulated size of read bytes
```
trustin: - discards -> Discards
- missing period at the end of the sentence.

How about adding a one-line description so that the generated Javadoc does not show an empty description cell in the summary table? e.g.

```java
/**
 * This method does nothing. Discards are now performed based on accumulated size of read bytes.
 * @deprecated This method does nothing.
 */
```

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -263,14 +299,20 @@ public final void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
      */
     protected void handlerRemoved0(ChannelHandlerContext ctx) throws Exception { }
 
+    /**
+     * Cheaper than calling {@code ctx.isRemoved()} from event loop
```
trustin: We should tell what this method does first, from the standpoint of an API user:

```java
/**
 * Tells whether this handle has been removed from the {@link ChannelPipeline}.
 * This method is useful because 1) a user can check if this handler is removed even if
 * he or she does not have a reference to {@link ChannelHandlerContext} and 2) it is
 * cheaper than {@link ChannelHandlerContext#isRemoved()}.
 */
```

njhill:
Thanks @trustin, have addressed your comments. I take it then that you're in favour of the proposed change of when read bytes are discarded? and if so do you think the "min bytes before discard" threshold should be configurable, e.g. via a `setMinimumReadBytesBeforeDiscard(int)` method, or better to keep it simple for now?

trustin:
I'm fine of either options but I think we can make it configurable when people asks for it, as long as we have the sensible default.

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
```diff
@@ -1712,30 +1723,26 @@ public CompositeByteBuf discardReadBytes() {
...
+        c.endOffset -= readerIndex;
+        c.adjustment += readerIndex;
+        c.slice = null;
+
+        Component la = lastAccessed;
+        if (la != null && la.endOffset < readerIndex) {
+            lastAccessed = null;
         }
 
         removeCompRange(0, firstComponentId);
```
normanmaurer: @njhill could we make a separate PR for the `CompositeByteBuf` changes and so keep things focused ?

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -110,49 +115,81 @@ public ByteBuf cumulate(ByteBufAllocator alloc, ByteBuf cumulation, ByteBuf in)
...
+                addFlattenedComponents(composite, in);
+                in = null;
+                return composite;
             } finally {
                 if (in != null) {
-                    // We must release if the ownership was not transferred as otherwise it may produce a leak if
-                    // writeBytes(...) throw for whatever release (for example because of OutOfMemoryError).
+                    // We must release if the ownership was not transferred as otherwise it may produce a leak
                     in.release();
+                    if (composite != null && composite != cumulation) {
```
normanmaurer: should this be moved out of the `in != null` block ?

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -110,49 +115,81 @@ public ByteBuf cumulate(ByteBufAllocator alloc, ByteBuf cumulation, ByteBuf in)
...
+        if (!(toAdd instanceof CompositeByteBuf)) {
+            return composite.addComponent(true, toAdd);
+        }
+        CompositeByteBuf from = (CompositeByteBuf) toAdd;
+        final int ridx = from.readerIndex();
+        final int widx = from.writerIndex();
+        final int compCount = from.numComponents();
+        int cidx = from.toComponentIndex(ridx);
+        int newReaderIdx = ridx - from.toByteIndex(cidx);
+        while (true) {
```
normanmaurer: nit: can you use `for(;;)` to match what we usually use in netty ?

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
njhill: @normanmaurer this was actually intentional since I think we only want to release it in the case we allocated it within the method _and_ aren't returning successfully.  Maybe it would be more readable to move it out and add `composite = null;` next to the existing `in = null;` ... I'll change that.

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
njhill: Sure, the reason I included them here is that the other changes kind of rely on discarding read components to be cheap, especially if there are none to discard (as would be common). I'll split into another PR though which can precede this one.

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
njhill: Oops no that won't work of course since we need to return `composite` :) I will just add a comment, hopefully that will be sufficient.

njhill:
@normanmaurer I've moved the `CompositeByteBuf` changes to #8939, PTAL. I also moved the shallow-copy "helper" logic to a new method there, since it could be more generally useful and can be more efficient since internals can be modified directly. If you think that one looks ok I will rebase this on it and add some benchmarks.

normanmaurer:
@netty-bot test this please

normanmaurer:
@njhill please fix the test errors...

njhill:
@normanmaurer sure, will look when i can, on vac this week so may be a bit slow

normanmaurer:
@njhill no worries... thanks a lot!

njhill:
Thanks @trustin!

@normanmaurer I've pushed 066afd4f5f576284fa3e4c3e4a3a70d30813e17e which fixes the test failures. It also includes "fixes" to other places which do the same thing, namely passing a wrapped buffer with refcount of 1 into `Channel.writeInbound()` assuming that its contents won't be changed (it should be completely giving up ownership in this case).

njhill:
@normanmaurer I have moved the commit mentioned in the prior comment to a new PR #9058 since it's kind of an independent fix. I will rebase this after that's merged, tests will fail again on this branch until then!

normanmaurer:
@netty-bot test this please

njhill:
@normanmaurer test failure is just related to the rebase, I can fix. But maybe hold off merging this a little longer, I was planning to open one more `ByteBuf` PR first to also be used in this one.

normanmaurer:
@njhill can you rebase ?

normanmaurer:
@njhill is this ready to re-review ?

normanmaurer:
@netty-bot test this please

njhill:
@normanmaurer yes I think so, thanks (and I did rebase as I'm sure you noticed). There's no need to wait for the other change mentioned in prior comment, can handle that in another update.

njhill:
@normanmaurer I pushed an additional commit after looking at this more myself. This will ensure that composite buffers will always be joined into a contiguous buffer when using `MERGE_CUMULATOR`, which I think is most consistent/expected/correct.

And one more thing I'd like to verify is that these changes haven't pushed either of the `cumulate(...)` methods over the default inlining size limit. I'll aim to do that soon and update here.

normanmaurer:
@netty-bot test this please

codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java
```diff
@@ -258,6 +268,7 @@ public final void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
             }
         }
         handlerRemoved0(ctx);
+        decodeState = STATE_REMOVED;
```
normanmaurer: I think this should be set in a `finally` block

normanmaurer:
@netty-bot test this please

njhill:
@normanmaurer I verified that everything is still small enough for inlining, so hopefully this is good to go. If #9735 is accepted then we may want to merge that first and update the composite check in this one to use it.

njhill:
@normanmaurer I've re-based and incorporated `isContiguous()` method. Hopefully it's now ready!

normanmaurer:
@netty-bot test this please

normanmaurer:
@njhill LGTM.... do we have any benchmarks ?

njhill:
@normanmaurer I will aim to run some benchmarks soon. Also have some related follow-on changes coming - hopefully will show incremental improvements from each.

normanmaurer:
@njhill ok thanks

normanmaurer:
@njhill please rebase and let me know once I should have a look again. Would be nice to process some of the PR backlog :)

njhill:
@normanmaurer sorry for the hold up... There's at least one edge case I came across while benchmarking (in terms of data arrival / message consumption patterns) which resulted in lower performance.

This case involves the cumulation never being retained or fully consumed, so that it's continually reused. By continually discarding, the new logic keeps it smaller than it would have grown to otherwise (using less mem) but means discards are more frequent and thus the rate of bytes copied is slightly higher. I don't know if it's a realistic case but anyhow would be eliminated by some follow-on changes I have so I'm thinking to reduce the scope of this PR a bit to avoid this behaviour and open some other ones.

normanmaurer:
Sounds good

njhill:
@normanmaurer I have now opened #9891 to replace this one - leaving current discard-related logic intact for now.

PR #8932, opened 2019-03-12
Support ALLOW_HALF_CLOSURE channel option on Unix domain socket.

zhongl:
Motivation:

Since DomainSocketChannel is a DuplexChannel,  which be able to shutdown input or output individually on demands, but ALLOW_HALF_CLOSURE channel option has not been supported yet.

I thought this could be a missing feature of Unix domain socket, so here the PR for it.

Modifications:

1. Added allHalfClosure property both in  EpollDomainSocketChannelConfig and KQueueDomainSocketChannelConfig,
2. Enabled isAllowHalfClosure method of native channel to support domain channel config,
3. Created EpollDomainSocketShutdownOutputByPeerTest and KQueueDomainSocketShutdownOutputByPeerTest to verify the change.

Result:

ALLOW_HALF_CLOSURE channel option can be set with DomainSocketChannel, and no more warning of Unknown channel option 'ALLOW_HALF_CLOSURE'.

### DIFF
--- a/testsuite/src/main/java/io/netty/testsuite/transport/socket/AbstractSocketShutdownOutputByPeerTest.java
+++ b/testsuite/src/main/java/io/netty/testsuite/transport/socket/AbstractSocketShutdownOutputByPeerTest.java
@@ -43,7 +43,7 @@ public abstract class AbstractSocketShutdownOutputByPeerTest<Socket> extends Abs
 
     public void testShutdownOutput(ServerBootstrap sb) throws Throwable {
         TestHandler h = new TestHandler();
-        Socket s = getSocket();
+        Socket s = newSocket();
         Channel sc = null;
         try {
             sc = sb.childHandler(h).childOption(ChannelOption.ALLOW_HALF_CLOSURE, true).bind().sync().channel();
@@ -84,7 +84,7 @@ public abstract class AbstractSocketShutdownOutputByPeerTest<Socket> extends Abs
 
     public void testShutdownOutputWithoutOption(ServerBootstrap sb) throws Throwable {
         TestHandler h = new TestHandler();
-        Socket s = getSocket();
+        Socket s = newSocket();
         Channel sc = null;
         try {
             sc = sb.childHandler(h).bind().sync().channel();
@@ -125,9 +125,9 @@ public abstract class AbstractSocketShutdownOutputByPeerTest<Socket> extends Abs
 
     protected abstract void close(Socket s) throws IOException;
 
-    protected abstract void write(Socket s, int i) throws IOException;
+    protected abstract void write(Socket s, int data) throws IOException;
 
-    protected abstract Socket getSocket();
+    protected abstract Socket newSocket();
 
     private static class TestHandler extends SimpleChannelInboundHandler<ByteBuf> {
         volatile DuplexChannel ch;
--- a/testsuite/src/main/java/io/netty/testsuite/transport/socket/AbstractSocketShutdownOutputByPeerTest.java
+++ b/testsuite/src/main/java/io/netty/testsuite/transport/socket/AbstractSocketShutdownOutputByPeerTest.java
@@ -66,9 +66,10 @@ public abstract class AbstractSocketShutdownOutputByPeerTest<Socket> extends Abs
             assertTrue(h.ch.isActive());
             assertTrue(h.ch.isInputShutdown());
             assertFalse(h.ch.isOutputShutdown());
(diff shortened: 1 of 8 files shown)


transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollDomainSocketChannelConfig.java
```diff
@@ -44,6 +46,9 @@
         if (option == EpollChannelOption.DOMAIN_SOCKET_READ_MODE) {
             return (T) getReadMode();
         }
+        if (option == EpollChannelOption.ALLOW_HALF_CLOSURE) {
```
normanmaurer: I think you also need to add it to the `getOptions` method. See : https://github.com/netty/netty/blob/1c66815c8aadb53f373e8070279a510a2eda89a7/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollSocketChannelConfig.java#L60

transport-native-kqueue/src/main/java/io/netty/channel/kqueue/KQueueDomainSocketChannelConfig.java
```diff
@@ -47,6 +49,9 @@
         if (option == DOMAIN_SOCKET_READ_MODE) {
             return (T) getReadMode();
         }
+        if (option == ALLOW_HALF_CLOSURE) {
```
normanmaurer: I think you also need to add it to the `getOptions` method. See : https://github.com/netty/netty/blob/1c66815c8aadb53f373e8070279a510a2eda89a7/transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollSocketChannelConfig.java#L60

zhongl:
@normanmaurer - fixed.

transport-native-epoll/src/main/java/io/netty/channel/epoll/EpollDomainSocketChannelConfig.java
```diff
@@ -28,14 +29,15 @@
...
+    private volatile boolean allowHalfClosure;
 
     EpollDomainSocketChannelConfig(AbstractEpollChannel channel) {
         super(channel);
     }
 
     @Override
     public Map<ChannelOption<?>, Object> getOptions() {
-        return getOptions(super.getOptions(), EpollChannelOption.DOMAIN_SOCKET_READ_MODE);
+        return getOptions(super.getOptions(), EpollChannelOption.DOMAIN_SOCKET_READ_MODE, ChannelOption.ALLOW_HALF_CLOSURE);
```
normanmaurer: use `EpollChannelOption.ALLOW_HALF_CLOSURE` here or use `ChannelOption.ALLOW_HALF_CLOSURE` bellow to make it consistent

testsuite/src/main/java/io/netty/testsuite/transport/socket/AbstractSocketShutdownOutputByPeerTest.java
```diff
@@ -0,0 +1,162 @@
...
+            h.closure.await();
+
+            assertFalse(h.ch.isOpen());
+            assertFalse(h.ch.isActive());
+            assertTrue(h.ch.isInputShutdown());
+            assertTrue(h.ch.isOutputShutdown());
+
+            assertEquals(1, h.halfClosure.getCount());
+            Thread.sleep(100);
+            assertEquals(0, h.halfClosureCount.intValue());
```
normanmaurer: @zhongl I wonder if it may be better to do something like this to reduce the risk of false-positives:

```
while (h.halfClosure.getCount() != 1 && h.halfClosureCount.intValue() != 0) {
    Thread.sleep(100);
}
```

Just ensure you use a timeout in the `@Test` declaration so we not end up in an endless loop if something went wrong

zhongl: ok, how about using this while pattern to improve another thread sleeping at line 70?

normanmaurer: sounds good

zhongl: done!

normanmaurer:
@netty-bot test this please

normanmaurer:
@netty-bot test this please

PR #8933, opened 2019-03-12
Fail build when Illegal reflective access is detected

normanmaurer:
Motivation:

We want to make the experience as smooth as possible for our users when using Java9+ and so should ensure we do not produce any 'Illegal reflective access' errors when using netty.

Modifications:

Add jvmArgs when running our tests that will deny reflective access and so will fail the build at the end due not be able to load some classes.

Result:

Ensure we do not produce any illegal refelctive access errors when using java9+

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -97,6 +97,9 @@
         <jdk>12</jdk>
       </activation>
       <properties>
+        <argLine.java9.extras />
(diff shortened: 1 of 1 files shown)


normanmaurer:
@netty-bot test this please

normanmaurer:
Let me merge this as its just a pom.xml change.

PR #8934, opened 2019-03-12
Remove --add-opens=java.base/java.nio=ALL-UNNAMED when running tests …

normanmaurer:
…as it is not needed anymore since a long time

Motivation:

At some point we needed --add-opens=java.base/java.nio=ALL-UNNAMED to run our native tests but this is not true anymore.

Modifications:

Remove --add-opens=java.base/java.nio=ALL-UNNAMED when running native tests.

Result:

Remove obsolate jvm arg.


normanmaurer:
Let me merge this as its just a removal of a flag during test runs

PR #8939, opened 2019-03-15
CompositeByteBuf optimizations and new addFlattenedComponents method

njhill:
Motivation:

The `CompositeByteBuf` `discardReadBytes` and `discardReadComponents` methods are currently quite inefficient, including when there are no read components to discard. We would like to call the latter more frequently in `ByteToMessageDecoder#COMPOSITE_CUMULATOR`.

In the same context it would be beneficial to perform a "shallow copy" of a composite buffer (for example when it has a refcount > 1) to avoid having to allocate and copy the contained bytes just to obtain an "independent" cumulation.

Modifications:

- Optimize `discardReadBytes()` and `discardReadComponents()` implementations (start at first comp rather than performing a binary search for the readerIndex).
- New `addFlattenedComponents(boolean,ByteBuf)` method which performs a shallow copy if the provided buffer is also composite and avoids adding any empty buffers, plus unit test.
- Other minor optimizations to avoid unnecessary checks.

Results:

`discardReadXX` methods are faster, composite buffers can be easily appended without deepening the buffer "tree" or retaining unused components.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
@@ -288,7 +288,7 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
                 c.reposition(components[cIndex - 1].endOffset);
             }
             if (increaseWriterIndex) {
-                writerIndex(writerIndex() + readableBytes);
+                writerIndex += readableBytes;
             }
             return cIndex;
         } finally {
@@ -371,7 +371,7 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
                 updateComponentOffsets(ci); // only need to do this here for components after the added ones
             }
             if (increaseWriterIndex && ci > cIndex && ci <= componentCount) {
-                writerIndex(writerIndex() + components[ci - 1].endOffset - components[cIndex].offset);
+                writerIndex += components[ci - 1].endOffset - components[cIndex].offset;
             }
         }
     }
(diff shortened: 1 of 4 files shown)


buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
```diff
@@ -414,6 +414,77 @@ public CompositeByteBuf addComponents(int cIndex, Iterable<ByteBuf> buffers) {
...
+        final int writerIndexBefore = writerIndex;
+        try {
+            for (int cidx = from.toComponentIndex0(ridx), newOffset = capacity();; cidx++) {
+                final Component component = fromComponents[cidx];
+                final int compOffset = component.offset;
+                final int fromIdx = Math.max(ridx, compOffset);
+                final int toIdx = Math.min(widx, component.endOffset);
+                final int len = toIdx - fromIdx;
+                if (len > 0) { // skip empty components
+                    final ByteBuf buf = component.buf, slice = component.slice, newSlice;
```
normanmaurer: can you split this in three lines (separate assignments) ?

njhill:
@normanmaurer I've addressed your comment and split the fix into separate PR #8946 (and amended this PR name and commit message to reflect that). I can rebase this once the other one is in.

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
```diff
@@ -414,6 +414,77 @@ public CompositeByteBuf addComponents(int cIndex, Iterable<ByteBuf> buffers) {
...
+     * Add the given {@link ByteBuf} and increase the {@code writerIndex} if {@code increaseWriterIndex} is
+     * {@code true}. If the provided buffer is a {@link CompositeByteBuf} itself, a "shallow copy" of its
+     * readable components will be performed. Thus the actual number of new components added may vary
+     * and in particular will be zero if the provided buffer is not readable.
+     * <p>
+     * {@link ByteBuf#release()} ownership of {@code buffer} is transferred to this {@link CompositeByteBuf}.
+     * @param buffer the {@link ByteBuf} to add. {@link ByteBuf#release()} ownership is transferred to this
+     * {@link CompositeByteBuf}.
+     */
+    public CompositeByteBuf addFlattenedComponents(boolean increaseWriterIndex, ByteBuf buffer) {
```
trustin: I'm curious if this can be named just as `addComponents()`.

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
```diff
@@ -414,6 +414,77 @@ public CompositeByteBuf addComponents(int cIndex, Iterable<ByteBuf> buffers) {
...
+        checkNotNull(buffer, "buffers");
```
trustin: `"buffer"`

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
```diff
@@ -414,6 +414,77 @@ public CompositeByteBuf addComponents(int cIndex, Iterable<ByteBuf> buffers) {
...
+            if (increaseWriterIndex) {
+                writerIndex = writerIndexBefore + (widx - ridx);
+            }
+            consolidateIfNeeded();
+            buffer.release();
+            buffer = null;
+            return this;
+        } finally {
+            if (buffer != null) {
+                // if we did not succeed, attempt to rollback any added
```
trustin: added components?

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
njhill: @trustin agree that would be much nicer but unfortunately there are already `addComponents(...)` methods with different semantics - in those cases the components added are 1-1 with the buffer(s) provided, so for example any `CompositeByteBuf` added is just "nested" as-is rather than flattened.

Would welcome other suggestions for a better name though, I don't love `addFlattenedComponents`!

njhill:
Thanks @trustin, have now rebased and addressed those comments.

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
normanmaurer: unfortunately I dont have any better name atm... @trustin any idea ?

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
```diff
@@ -414,6 +414,77 @@ public CompositeByteBuf addComponents(int cIndex, Iterable<ByteBuf> buffers) {
...
+                final int fromIdx = Math.max(ridx, compOffset);
+                final int toIdx = Math.min(widx, component.endOffset);
+                final int len = toIdx - fromIdx;
+                if (len > 0) { // skip empty components
+                    final ByteBuf buf = component.buf;
+                    ByteBuf slice = component.slice;
+                    if (slice == null) {
+                        buf.retain();
+                    } else {
+                        slice = slice.retainedSlice(fromIdx - compOffset, len);
```
normanmaurer: please add a comment why we need to take a slice

njhill: @normanmaurer thanks for highlighting this, after thinking/looking more I think a slice actually isn't required here! In particular there should be no harm retaining the `rootParent` of `PooledSlicedByteBuf`s, it's the release that we needed to be careful about in other places. I'll update.

njhill:
@normanmaurer now updated, PTAL

normanmaurer:
@netty-bot test this please

normanmaurer:
@netty-bot test this please

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
trustin: Hmm.. not really. Just name it as `addComponents()` and document better? 😅

njhill:
@normanmaurer I think the java12 test failures are unrelated to the changes?

normanmaurer:
@njhill yep... I think this test code is a bit racy... will have a look.

normanmaurer:
@netty-bot test this please

normanmaurer:
@vkostyukov it will be used here: https://github.com/netty/netty/pull/8931/files

buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
normanmaurer: @njhill ^^

njhill: @normanmaurer sorry, I'd kind of assumed @trustin was joking :). It would be a breaking chance since it would collide with the existing `addComponents(boolean, ByteBuf ...)` method (that performs a "non-flattened" addition). And even with clear doc imo it would still be confusing to have a variety of `addComponents()` methods with different args that all do the same thing and then one which happens to behave differently.

normanmaurer: true you are right.... lets keep it then.

normanmaurer:
@njhill thanks a lot!

PR #8941, opened 2019-03-15
Allow to automatically trim the PoolThreadCache in a timely interval

normanmaurer:
Motivation:

PooledByteBufAllocator uses a PoolThreadCache per Thread that allocates / deallocates to minimize the performance overhead. This PoolThreadCache is trimmed after X allocations to free up buffers that are not allocated for a long time. This works out quite well when the app continues to allocate but fails if the app stops to allocate frequently (for whatever reason) and so a lot of memory is wasted and not given back to the arena / freed.

Modifications:

- Add a ThreadExecutorMap that offers multiple methods that wrap Runnable / ThreadFactory / Executor and allow to call ThreadExecutorMap.currentEventExecutor() to get the current executing EventExecutor for the calling Thread.
- Use these methods in the constructors of our EventExecutor implementations (which also covers the EventLoop implementations)
- Add io.netty.allocator.cacheTrimIntervalMillis system property which can be used to specify a fixed rate / interval on which we should try to trim the PoolThreadCache for a EventExecutor that allocates.
- Add PooledByteBufAllocator.trimCurrentThreadCache() to allow the user to trim the cache of the calling thread manually.
- Add testcases
- Introduce FastThreadLocal.getIfExists()

Result:

Allow to better / more frequently trim PoolThreadCache and so give back memory to the area / system.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java
+++ b/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java
@@ -19,11 +19,13 @@ package io.netty.buffer;
 import static io.netty.util.internal.ObjectUtil.checkPositiveOrZero;
 
 import io.netty.util.NettyRuntime;
+import io.netty.util.concurrent.EventExecutor;
 import io.netty.util.concurrent.FastThreadLocal;
 import io.netty.util.concurrent.FastThreadLocalThread;
 import io.netty.util.internal.PlatformDependent;
 import io.netty.util.internal.StringUtil;
 import io.netty.util.internal.SystemPropertyUtil;
+import io.netty.util.internal.ThreadExecutorMap;
 import io.netty.util.internal.logging.InternalLogger;
 import io.netty.util.internal.logging.InternalLoggerFactory;
 
@@ -31,6 +33,7 @@ import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 
 public class PooledByteBufAllocator extends AbstractByteBufAllocator implements ByteBufAllocatorMetricProvider {
 
@@ -45,12 +48,20 @@ public class PooledByteBufAllocator extends AbstractByteBufAllocator implements 
     private static final int DEFAULT_NORMAL_CACHE_SIZE;
     private static final int DEFAULT_MAX_CACHED_BUFFER_CAPACITY;
     private static final int DEFAULT_CACHE_TRIM_INTERVAL;
+    private static final long DEFAULT_CACHE_TRIM_INTERVAL_MILLIS;
     private static final boolean DEFAULT_USE_CACHE_FOR_ALL_THREADS;
     private static final int DEFAULT_DIRECT_MEMORY_CACHE_ALIGNMENT;
     static final int DEFAULT_MAX_CACHED_BYTEBUFFERS_PER_CHUNK;
 
     private static final int MIN_PAGE_SIZE = 4096;
     private static final int MAX_CHUNK_SIZE = (int) (((long) Integer.MAX_VALUE + 1) / 2);
+
+    private final Runnable trimTask = new Runnable() {
+        @Override
+        public void run() {
+            PooledByteBufAllocator.this.trimCurrentThreadCache();
+        }
+    };
 
     static {
         int defaultPageSize = SystemPropertyUtil.getInt("io.netty.allocator.pageSize", 8192);
@@ -113,6 +124,9 @@ public class PooledByteBufAllocator extends AbstractByteBufAllocator implements 
         DEFAULT_CACHE_TRIM_INTERVAL = SystemPropertyUtil.getInt(
                 "io.netty.allocator.cacheTrimInterval", 8192);
 
+        DEFAULT_CACHE_TRIM_INTERVAL_MILLIS = SystemPropertyUtil.getLong(
+                "io.netty.allocation.cacheTrimIntervalMillis", 0);
+
         DEFAULT_USE_CACHE_FOR_ALL_THREADS = SystemPropertyUtil.getBoolean(
                 "io.netty.allocator.useCacheForAllThreads", true);
 
@@ -143,6 +157,7 @@ public class PooledByteBufAllocator extends AbstractByteBufAllocator implements 
             logger.debug("-Dio.netty.allocator.normalCacheSize: {}", DEFAULT_NORMAL_CACHE_SIZE);
             logger.debug("-Dio.netty.allocator.maxCachedBufferCapacity: {}", DEFAULT_MAX_CACHED_BUFFER_CAPACITY);
             logger.debug("-Dio.netty.allocator.cacheTrimInterval: {}", DEFAULT_CACHE_TRIM_INTERVAL);
+            logger.debug("-Dio.netty.allocator.cacheTrimIntervalMillis: {}", DEFAULT_CACHE_TRIM_INTERVAL_MILLIS);
             logger.debug("-Dio.netty.allocator.useCacheForAllThreads: {}", DEFAULT_USE_CACHE_FOR_ALL_THREADS);
             logger.debug("-Dio.netty.allocator.maxCachedByteBuffersPerChunk: {}",
                     DEFAULT_MAX_CACHED_BYTEBUFFERS_PER_CHUNK);
@@ -438,11 +453,20 @@ public class PooledByteBufAllocator extends AbstractByteBufAllocator implements 
             final PoolArena<byte[]> heapArena = leastUsedArena(heapArenas);
             final PoolArena<ByteBuffer> directArena = leastUsedArena(directArenas);
 
-            Thread current = Thread.currentThread();
+            final Thread current = Thread.currentThread();
             if (useCacheForAllThreads || current instanceof FastThreadLocalThread) {
-                return new PoolThreadCache(
+                final PoolThreadCache cache = new PoolThreadCache(
                         heapArena, directArena, tinyCacheSize, smallCacheSize, normalCacheSize,
                         DEFAULT_MAX_CACHED_BUFFER_CAPACITY, DEFAULT_CACHE_TRIM_INTERVAL);
+
+                if (DEFAULT_CACHE_TRIM_INTERVAL_MILLIS > 0) {
+                    final EventExecutor executor = ThreadExecutorMap.currentExecutor();
+                    if (executor != null) {
+                        executor.scheduleAtFixedRate(trimTask, DEFAULT_CACHE_TRIM_INTERVAL_MILLIS,
+                                DEFAULT_CACHE_TRIM_INTERVAL_MILLIS, TimeUnit.MILLISECONDS);
+                    }
+                }
+                return cache;
             }
             // No caching so just use 0 as sizes.
             return new PoolThreadCache(heapArena, directArena, 0, 0, 0, 0, 0);
@@ -601,6 +625,21 @@ public class PooledByteBufAllocator extends AbstractByteBufAllocator implements 
         PoolThreadCache cache =  threadCache.get();
         assert cache != null;
         return cache;
+    }
+
+    /**
+     * Trim thread local cache for the current {@link Thread}, which will give back any cached memory that was not
+     * allocated frequently since the last trim operation.
+     *
+     * Returns {@code true} if a cache for the current {@link Thread} exists and so was trimmed, false otherwise.
+     */
+    public boolean trimCurrentThreadCache() {
+        PoolThreadCache cache = threadCache.getIfExists();
+        if (cache != null) {
+            cache.trim();
+            return true;
+        }
+        return false;
     }
 
     /**
--- a/common/src/main/java/io/netty/util/concurrent/GlobalEventExecutor.java
+++ b/common/src/main/java/io/netty/util/concurrent/GlobalEventExecutor.java
@@ -15,6 +15,7 @@
  */
 package io.netty.util.concurrent;
 
+import io.netty.util.internal.ThreadExecutorMap;
 import io.netty.util.internal.logging.InternalLogger;
 import io.netty.util.internal.logging.InternalLoggerFactory;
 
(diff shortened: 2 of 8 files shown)


buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java
```diff
@@ -440,9 +448,22 @@ protected synchronized PoolThreadCache initialValue() {
...
                         heapArena, directArena, tinyCacheSize, smallCacheSize, normalCacheSize,
                         DEFAULT_MAX_CACHED_BUFFER_CAPACITY, DEFAULT_CACHE_TRIM_INTERVAL);
+
+                if (DEFAULT_CACHE_TRIM_RATE > 0) {
+                    EventExecutor executor = ThreadExecutorMap.currentExecutor();
+                    if (executor != null) {
+                        executor.scheduleAtFixedRate(new Runnable() {
+                            @Override
+                            public void run() {
+                                cache.trim();
```
johnou: try catch Exception to avoid suppressing subsequent executions?

normanmaurer: @johnou this should never throw, if it does there is something seriously wrong

common/src/main/java/io/netty/util/concurrent/GlobalEventExecutor.java
```diff
@@ -65,6 +65,8 @@ public void run() {
 
     private GlobalEventExecutor() {
```
johnou: why does shutdown throw UnsupportedOperationException? can we tweak the lifecycle to invoke clear on the thread local in ThreadExecutorMap?

normanmaurer: @johnou because you never shutdown by yourself. The used Thread will die after some time by itself which also means it will be removed from the map

buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java
johnou: @normanmaurer but unless the app developer has hooked into java.lang.Thread#setDefaultUncaughtExceptionHandler they are not going to know and the exception will be swallowed..

common/src/main/java/io/netty/util/concurrent/GlobalEventExecutor.java
johnou: @normanmaurer just wanted to raise a discussion, I was thinking of the use case where people redeploy webapps and it complains about thread locals / leaking memory - not a problem for us :wink:

eolivelli:
awesome @normanmaurer 
I will take a deeper look as soon as possible

buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java
```diff
@@ -440,9 +448,22 @@ protected synchronized PoolThreadCache initialValue() {
...
             Thread current = Thread.currentThread();
             if (useCacheForAllThreads || current instanceof FastThreadLocalThread) {
-                return new PoolThreadCache(
+                final PoolThreadCache cache = new PoolThreadCache(
                         heapArena, directArena, tinyCacheSize, smallCacheSize, normalCacheSize,
                         DEFAULT_MAX_CACHED_BUFFER_CAPACITY, DEFAULT_CACHE_TRIM_INTERVAL);
+
+                if (DEFAULT_CACHE_TRIM_RATE > 0) {
+                    EventExecutor executor = ThreadExecutorMap.currentExecutor();
+                    if (executor != null) {
```
eolivelli: Is it possible that we have a null ?
in that case DEFAULT_CACHE_TRIM_RATE won't have effect, should we log a warning ?

common/src/test/java/io/netty/util/internal/ThreadExecutorMapTest.java
```diff
@@ -0,0 +1,63 @@
...
+        ThreadExecutorMap.apply(new Runnable() {
+            @Override
+            public void run() {
+                Assert.assertSame(ImmediateEventExecutor.INSTANCE, ThreadExecutorMap.currentExecutor());
+            }
+        }, ImmediateEventExecutor.INSTANCE).run();
+    }
+
+    @Test
+    public void testDecorateTheadFactory() throws InterruptedException {
```
eolivelli: nit: typo, missing 'r'

buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java
normanmaurer: yes its possible if people call from there own thread which not power a `EventExecutor`. This is also why I think we should not log. If we log then it should be in trace I guess.

eolivelli: In one of my use cases I am pretty sure that I am not inside an event executor.
For instance, I am inside the request handler of a Servlet on Tomcat, it is not a Netty managed thread, in this case when I call allocator#directBuffer(), this may trigger this method and I will end up in a PoolThreadCache which does not observe this limit.

So it is better not to log, you are right.

But is there a way to leverage this feature in this case? 

I guess I should wrap my allocator#directBuffer() call with a ThreadExecutorMap.apply(Runnable, EventExecutor)
But I still don't have an EventExecutor.

So my self answer is that I should manage a special global EventExecutor and wrap all of the calls

can you confirm @normanmaurer  ?

normanmaurer: @eolivelli in this case you will need to call `PooledByteBufAllocator.trimCurrentThreadCache()` from the thread from time to time to have the same effect. That said we already trim by ourself from time to time, so you only need to explicit do it by yourself if you not allocate frequently enough.

See https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/PoolThreadCache.java#L192.

Alternative you can also disable the cache for threads that are not owned by netty:

https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java#L442

eolivelli: @normanmaurer  thank you.
It is very clear to me now.

common/src/main/java/io/netty/util/concurrent/UnorderedThreadPoolEventExecutor.java
```diff
@@ -59,6 +60,7 @@ public UnorderedThreadPoolEventExecutor(int corePoolSize) {
      */
     public UnorderedThreadPoolEventExecutor(int corePoolSize, ThreadFactory threadFactory) {
         super(corePoolSize, threadFactory);
+        this.setThreadFactory(ThreadExecutorMap.apply(getThreadFactory(), this));
```
njhill: Is this is safe? I think the scheduled repeating trimming task could be run on an arbitrary thread, i.e. not necessarily the one from which it was scheduled that owns the referenced `PoolThreadCache` instance.

Maybe for this one the scheduled task could just call `PoolThreadCache.trimCurrentThreadCache()`?

normanmaurer: you are right... let me fix it

normanmaurer: @njhill also with this in mind I wonder if we should maybe restrict all of this to `OrderedEventExecutor`

normanmaurer: @njhill done that ....

njhill: Thanks @normanmaurer generally LGTM, but I'm not sure about the `OrderedEventExecutor` check since my understanding is that although `SingleThreadEventExecutor` and `GlobalEventExecutor` are ok, that interface doesn't necessarily imply same-thread semantics, only serialized tasks. So having the check there _might_ be misleading... it's necessary but not sufficient. Maybe we could have a static `scheduledTasksRunOnSameThread(EventExecutor)` helper method which for now returns true only in the case of STEE and GEE?

(but I still like the change to make GEE implement OEE!)

buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java
```diff
@@ -113,6 +118,9 @@
         DEFAULT_CACHE_TRIM_INTERVAL = SystemPropertyUtil.getInt(
                 "io.netty.allocator.cacheTrimInterval", 8192);
 
+        DEFAULT_CACHE_TRIM_RATE = SystemPropertyUtil.getLong(
+                "io.netty.allocator.cacheTrimRateMs", 0);
```
trustin: How about `io.netty.allocation.cacheTrimIntervalMillis`? To avoid confusion, we could deprecate `cacheTrimInterval` in favor of `cacheTrimIntervalCount`. The current name `rate` is misleading because it's not really a rate.

common/src/main/java/io/netty/util/concurrent/GlobalEventExecutor.java
```diff
@@ -34,7 +35,7 @@
  * task pending in the task queue for 1 second.  Please note it is not scalable to schedule large number of tasks to
  * this executor; use a dedicated executor.
  */
-public final class GlobalEventExecutor extends AbstractScheduledEventExecutor {
+public final class GlobalEventExecutor extends AbstractScheduledEventExecutor implements OrderedEventExecutor {
```
trustin: Nice idea, @njhill

common/src/main/java/io/netty/util/internal/ThreadExecutorMap.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.concurrent.Executor;
+import java.util.concurrent.ThreadFactory;
+
+/**
+ * Allow to retrieve the {@link EventExecutor} for the calling {@link Thread}.
+ */
+public final class ThreadExecutorMap {
+
+    private static final FastThreadLocal<EventExecutor> mappings =
+            new FastThreadLocal<EventExecutor>();
```
trustin: nit: could be merged into a single line.

common/src/main/java/io/netty/util/concurrent/GlobalEventExecutor.java
```diff
@@ -65,6 +65,8 @@ public void run() {
 
     private GlobalEventExecutor() {
         scheduledTaskQueue().add(quietPeriodTask);
+        threadFactory = ThreadExecutorMap.apply(new DefaultThreadFactory(
+                DefaultThreadFactory.toPoolName(getClass()), false, Thread.NORM_PRIORITY, null), this);
```
trustin: How about moving the decoration into `FastThreadLocalThread`? Then we don't need to wrap here and potentially in many other places.

normanmaurer: I was thinking about this but I think some people may end up using their own `Thread` and not `FastThreadLocalThread`

trustin: I see. Then how about keeping it as it is here and modifying `FastThreadLocalThread` as well, with some duplication suppression logic if possible?

common/src/main/java/io/netty/util/concurrent/GlobalEventExecutor.java
johnou: What if a blocking IO task is submitted, won't it jam up the executor now?

normanmaurer: It should not be any different then at the moment... why you think there is an issue now ?

johnou: I think it depends on the underlying executor, but the fact it now implements OrderedEventExecutor doesn't it mean it cannot run submitted tasks OOO which may result in tasks queuing up by implementing this interface?

common/src/main/java/io/netty/util/concurrent/UnorderedThreadPoolEventExecutor.java
njhill: @normanmaurer that's probably fine but technically there's still a chance that someone has subclassed it with different `scheduleAtFixedRate` behaviour that breaks the assertion, since the method is non-final (though this wouldn't really make much sense to do).

common/src/main/java/io/netty/util/concurrent/GlobalEventExecutor.java
njhill: @johnou it's just a marker interface which doesn't change any behaviour itself. `GlobalEventExecutor` _already_ behaves in the manner it specifies (serial, in-order execution) and hence might as well implement it.

common/src/main/java/io/netty/util/concurrent/UnorderedThreadPoolEventExecutor.java
normanmaurer: @njhill true.. I guess we could also just return directly if the `Thread` is not the same we expect. WDYT ?

normanmaurer: @njhill alternatively I just noticed we could just call:

```java
PooledByteBufAllocator.this.trimCurrentThreadCache();
```

This should be safe in any case.

njhill: @normanmaurer yes, that would be safe for any of the cases (worst case it would be a repeating no-op)... that's actually what I was meaning to suggest [above](https://github.com/netty/netty/pull/8941#discussion_r266107986) but obviously got the detail wrong since that method is on the BBAllocator not PoolThreadCache!

common/src/main/java/io/netty/util/concurrent/GlobalEventExecutor.java
normanmaurer: let me actually do this in another PR as it is not really related.

normanmaurer: https://github.com/netty/netty/pull/8952

buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java
normanmaurer: @trustin I think I would like to still preserve the old one as well. Let me rename it tho.

common/src/main/java/io/netty/util/concurrent/GlobalEventExecutor.java
normanmaurer: Actually I think I would like to keep it as it is as otherwise we will need to also pass in the `EventExecutor` to the `FastThreadLocalThread` constructor.

normanmaurer:
Just rebased it.

PR #8944, opened 2019-03-15
Adjust testsuite-osgi to resolve bundles from local build

normanmaurer:
Motivation:

testsuite-osgi currently resolve its bundles from the local / remote maven repository, which means you will need to do `mvn install` before it can pick up the bundles. Beside this this also means that you may pick up old versions if you forgot to call `install` before running it.

Modifications:

Use alta-maven-plugin to be able to resolve bundles from the local build directory during the build.

Result:

No need to install jars before running the OSGI testsuite and ensure we always test with the latest jars.

### DIFF
--- a/testsuite-osgi/pom.xml
+++ b/testsuite-osgi/pom.xml
@@ -46,6 +46,40 @@
         <skipOsgiTestsuite>true</skipOsgiTestsuite>
       </properties>
     </profile>
+
+    <profile>
+      <id>linux</id>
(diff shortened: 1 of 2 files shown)


normanmaurer:
@netty-bot test this please

normanmaurer:
@netty-bot test

normanmaurer:
Let me merge this so the OSGI test failures will not happen anymore

PR #8497, opened 2018-11-12
Fix ref-counting when CompositeByteBuf is used with retainedSlice()

njhill:
Motivation:

`ByteBuf.retainedSlice()` and similar methods produce sliced buffers with an independent refcount to the buffer that they wrap.

One of the optimizations in 10539f4dc738c48e8d63c46b72ca32906d7f40ec was to use a ref to the unwrapped buffer object for added slices, but this did not take into account the above special case when later releasing.

Thanks to @rkapsi for discovering this via #8495.

Modifications:

Since a reference to the slice is still kept in the `Component` class, just changed `Component.freeIfNecessary()` to release the slice in preference to the unwrapped buf.

Also added a unit test which reproduces the bug.

Result:

Fixes #8495

PR #8946, opened 2019-03-15
Fix possible ByteBuf leak when CompositeByteBuf is resized

njhill:
Motivation:

The special case fixed in #8497 also requires that we keep a derived slice when trimming components in place, as done by the `capacity(int)` and `discardReadBytes()` methods.

Modifications:

Ensure that we keep a ref to trimmed components' original retained slice in `capacity(int)` and `discardReadBytes()` methods, so that it is released properly when the they are later freed. Add unit test which fails prior to the fix.

Result:

Edge case leak is eliminated.

### DIFF
--- a/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
+++ b/buffer/src/main/java/io/netty/buffer/CompositeByteBuf.java
@@ -753,7 +753,12 @@ public class CompositeByteBuf extends AbstractReferenceCountedByteBuf implements
                 if (bytesToTrim < cLength) {
                     // Trim the last component
                     c.endOffset -= bytesToTrim;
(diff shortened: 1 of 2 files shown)


normanmaurer:
@netty-bot test this please

normanmaurer:
@netty-bot test this please

njhill:
FYI I just updated the commit message to fix a typo `discardReadComponents` -> `discardReadBytes`.

normanmaurer:
@netty-bot test this please

normanmaurer:
@netty-bot test this please

normanmaurer:
@netty-bot test this please

normanmaurer:
@njhill thanks!

PR #8948, opened 2019-03-18
Add docker-sync config to step up docker-usage on macOS.

normanmaurer:
Motivation:

docker-sync.io helps to speed up docker FS access on macOS and so make builds there a lot faster. We should add some config to help users use it.

Modifications:

Add docker-sync configs for centos-6.18 which is what we use for releases.

Result:

Faster builds via docker and when using macOS possible.


normanmaurer:
As this is just a docker-sync config let me cherry-pick

Issue #8942, opened 2019-03-15
With OPENSSL as SSL provider, no alert message is sent after TLS handshake failure


lh44:
### Expected behavior
We have a secured netty server and client implementations and we currently use OPENSSL as the SSL provider. During the TLS handshake, when the server produces an expired certificate, the client should send a TCP alert message(21) to notify about the certificate expiry to the server before sending a TCP FIN/RST.
This behavior exists when the SSL provider is JDK.

Wireshark traces in case of SSL provider as JDK(works as expected)
![JDK](https://user-images.githubusercontent.com/44922213/54415174-79889d80-4721-11e9-9960-9f243f046a26.PNG)

### Actual behavior
Client sends a TCP RST directly without an alert. This is causing issues at the server, as we have a reconnection strategy with fall back certificates once the alert is detected. Currently this is not being triggered.

Current behavior
![OPENSSL](https://user-images.githubusercontent.com/44922213/54415556-95d90a00-4722-11e9-83fd-185df2a536e0.PNG)

### Steps to reproduce
Please run the reproducer code given below, with the server configured with an expired certificate. 
Also start a wireshark capture on the tcp.port 7766

### Minimal yet complete reproducer code (or URL to code)

Secured Server-Client
```java
import java.io.File;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import java.util.concurrent.TimeUnit;
 
import javax.net.ssl.SSLEngine;
 
import io.netty.bootstrap.Bootstrap;
import io.netty.bootstrap.ServerBootstrap;
import io.netty.channel.ChannelFuture;
import io.netty.channel.ChannelInitializer;
import io.netty.channel.ChannelPipeline;
import io.netty.channel.EventLoopGroup;
import io.netty.channel.nio.NioEventLoopGroup;
import io.netty.channel.socket.SocketChannel;
import io.netty.channel.socket.nio.NioServerSocketChannel;
import io.netty.channel.socket.nio.NioSocketChannel;
import io.netty.handler.ssl.SslContext;
import io.netty.handler.ssl.SslContextBuilder;
import io.netty.handler.ssl.SslHandler;
import io.netty.handler.ssl.SslProvider;
import io.netty.handler.ssl.SupportedCipherSuiteFilter;
 
public class NettyTlsExample {
 
    static int m_port = 7766;
    static final SslProvider m_sslProvider = SslProvider.OPENSSL;
    static SslContext m_serverContext;
    static SslContext m_clientContext;
[...]

PR #8949, opened 2019-03-18
Correctly produce ssl alert when certificate validation fails on the …

normanmaurer:
…client-side when using native SSL implementation.

Motivation:

When the verification of the server cert fails because of the used TrustManager on the client-side we need to ensure we produce the correct alert and send it to the remote peer before closing the connection.

Modifications:

- Use the correct verification mode on the client-side by default.
- Update tests

Result:

Fixes https://github.com/netty/netty/issues/8942.


normanmaurer:
@lh44 PTAL

Issue #8508, opened 2018-11-13
[Netty 5] AttributeMap removal 


normanmaurer:
We want to remove the `AttributeMap` and also the `attr(...)` method from the `Channel` / `ChannelHandlerContext`. 

There are two possibilities for this:
 * provide another way to map things to a `Channel`. How exactly this is done is not clear yet, but perhaps the extensibility improvements will allow for wrapping the `Channel`, or state can be kept in a `ChannelHandler`. 
 * Just don't offer a replacement and have the user use stateful `ChannelHandler`.

slandelle:
> We want to remove the AttributeMap and also the attr(...) method from the Channel / ChannelHandlerContext.

@normanmaurer Why? What's the issue?

normanmaurer:
@slandelle the API is sometimes hard to use and not really gives a lot of advantages over just store the state in the ChannelHandler itself or use a `Map<ChannelId, Whatever>`.

slandelle:
@normanmaurer I agree that `AttributeMap` is not useful for `ChannelHandler`.

But I tend to disagree for `Channel`.

I use `AttributeMap` on `Channel` to store concurrent HTTP/2 streams count. Using a global map would actually make things more complicated: I would have to deal with concurrency and clean up (when channel gets closed so I don't leak memory).

normanmaurer:
@slandelle thats a good feedback :) What about providing something like `ChannelLocal` that we had in Netty 3 ?https://github.com/netty/netty/blob/3.10/src/main/java/org/jboss/netty/channel/ChannelLocal.java

slandelle:
@normanmaurer Isn't ChannelLocal what was replaced by AttributeMap in Netty 4? I guess it was dropped/replaced for a reason. Maybe listener overhead?

The API could be simplified for sure, but I like the feature :)

normanmaurer:
@slandelle yep it was... The reason was basically be cause @trustin and myself felt that it should be part of the `Channel` itself. That said I am not convinced of it anymore and it comes with a cost.

rkapsi:
Alternatively use composition over inheritance and don't extend Channel from AttributeMap.

johnou:
@normanmaurer you don't foresee any performance hit with `Map<ChannelId, Whatever>` vs a session object stored directly in the channel which can easily be referenced every read?

normanmaurer:
[...]

PR #8958, opened 2019-03-20
 #8508 simplify channel hierarchy by inlining AttributeMap interface …

doom369:
…into Channel

Motivation:

Current `AttributeMap` usage seems confusing. For example, `Channel` implements `AttributeMap` and `DefaultAttributeMap` implements `AttributeMap` and `ChannelHandlerContext` doesn't implement `AttributeMap` while has embedded `AttributeMap` methods within own interface.

Modification:

- Removed `AttributeMap` interface and it methods embedded directly into `Channel`;
- Removed deprecated `AttributeMap` methods from `ChannelHandlerContext` and all childs;
- Removed inheritance of `AttributeMap` from `DefaultAttributeMap`;
- Removed `DefaultAttributeMap` and embedded it code into `AbstractChannel`;

Result:

Hierarchy is simpler and more clear while `Channel` still have `attr`, `hasAttr` methods.

Partially connected with https://github.com/netty/netty/issues/8508.


common/src/main/java/io/netty/util/DefaultAttributeMap.java
```diff
@@ -22,10 +22,11 @@
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 
 /**
- * Default {@link AttributeMap} implementation which use simple synchronization per bucket to keep the memory overhead
+ * Default {@link DefaultAttributeMap} implementation which
+ * use simple synchronization per bucket to keep the memory overhead
  * as low as possible.
  */
-public class DefaultAttributeMap implements AttributeMap {
+public class DefaultAttributeMap {
```
normanmaurer: @doom369 should we merge this class with `AbstractChannel` ?

doom369: @normanmaurer hm... Actually, very good idea. Didn't think about it. Let me check.

doom369:
@normanmaurer indeed, looks much better now. And what is more important - now every channel object should be a bit smaller in the size.

The only problem - `DefaultHttp2StreamChannel` is doesn't extend `AbstractChannel` for some reason. So for now I just skipped implementation of `attr` for it.

njhill:
I don't understand the reason behind this. Having the nontrivial and logically independent attribute map implementation details in a separate class seems preferable to me over polluting the other `AbstractChannel` logic. I also don't see what's wrong with existence of the generic `AttributeMap` interface - in theory could be implemented by any entity that allows attributes to be attached to it.

> `Channel` implements `AttributeMap` and `DefaultAttributeMap` implements `AttributeMap`

I don't follow why this is undesirable or confusing. Consumers of `Channel` as an interface only care that it extends `AttributeMap`, `DefaultAttributeMap` is the implementation of `AttributeMap` so of course must implement it.

> `ChannelHandlerContext` doesn't implement `AttributeMap` while has embedded `AttributeMap` methods within own interface

This isn't actually true, the CHC interface _does_ extend `AttributeMap`, it just overrides its methods so that they can be annotated as `@Deprecated`, i.e. indicating that it won't continue to extend it in the future. FWIW I actually also don't see the harm in this staying, it's a nice convenience that you can do `cxt.attr(key)` instead of `cxt.channel().attr(key)` but I guess others felt differently at some point!

I did notice that `AbstractChannelHandlerContext` unnecessarily extends `DefaultAttributeMap`, and opened a separate PR #8960 for that since it could be fixed in 4.1 too.

Do we also want to remove the deprecated `Attribute.remove()` and `Attribute.getAndRemove()` methods?

njhill:
Maybe the reason for deprecating/removing `ChannelHandlerContext`'s own `AttributeMap` methods is that it kind of implies the attributes are attached to the handler context, which could be misleading. On reflection that does seem like a good reason to do it, `cxt.channel().attr(key)` is much clearer in that sense.

transport/src/main/java/io/netty/channel/AbstractChannel.java
```diff
@@ -446,6 +459,167 @@ protected final void readIfIsAutoRead() {
         }
     }
 
+    @SuppressWarnings("unchecked")
+    public <T> Attribute<T> attr(AttributeKey<T> key) {
```
rkapsi: I'd keep `AttributeMap`, add `Channel#attributeMap()` or something like that and deprecate `Channel#attr(...)` et al. That's just me; composition over inheritance and `has-a` instead of `is-a` relationships.

njhill: @rkapsi that sounds like a reasonable option to me, but wouldn't it be more direct to think of `Channel` as having attributes rather than having a map which contains attributes? Maybe if `AttributeMap` was renamed to `AttributeHolder` or `AttributeAttachable` or something? This wouldn't prevent the implementation using composition if we didn't want to use inheritance.

rkapsi: Oh yea, it (Channel) not being an extension of AttributeMap is already huge. Love it. In terms of test-ability and future use-cases it's still a bit PITA though. In order to test this you've to create a instance of Channel (my eye twitches) and if you wanted to implement some new type of Channel you've to actually extend AbstractChannel because it's the only thing that provides an implementation thereof (tears of sadness).

doom369:
> I don't understand the reason behind this.

The reason is to simplify the Channel inheritance.

> I also don't see what's wrong with existence of the generic AttributeMap interface

Interfaces are extracted so they can be used as the generic abstraction over different implementations. Nettry extracts `AttributeMap`, but doesn't use it in any way, so there is no reason to extract the additional interface and hold it.

> Having the nontrivial and logically independent attribute map implementation details in a separate class seems preferable to me over polluting the other AbstractChannel logic

In general, composition is recommended over inheritance as it makes things simpler. Also, netty already has `AbstractChannel` that already holds all common logic for all channels. Channel state is part of channel (if we don't bother we 2 different implementations Stateful and Stateless channel). So nothing wrong here. It is mostly about style and individual preferences. However, composition approach in that particular case may have nice side effects due to java specific - composition could make allocated objects smaller. And for `Channel` objects this is important.

njhill:
@doom369 fair enough, I guess the main question is whether this is seen as channel-specific functionality or something which could apply more generally. I kind of assumed the latter but if not then agree that there’s no need for `AttributeMap` interface.

normanmaurer:
Honestly I am not sure if this buys us much. I mean what we could do in `AbstractChannel` is to not extend `DefaultAttributeMap` but just delegate to an instance. I think it may still be useful to re-use implementations.

trustin:
Armeria reuses the `AttributeMap` and `AttributeKey` API, but I'm fine with removing `AttributeMap` or not. I guess other people might be reusing it in their applications as well, though.

doom369:
No problem. I just wanted to show that code becomes simpler and easier with this PR. Up to you to decide.

normanmaurer:
I think I would like to keep it as it is. Let me close this

PR #8960, opened 2019-03-20
AbstractChannelHandlerContext doesn't need to extend DefaultAttributeMap

njhill:
Motivation:

It appears this was an oversight, maybe was valid at some point in the past. Noticed while reviewing #8958.

Modifications:

Change `AbstractChannelHandlerContext` to not extend `DefaultAttributeMap`.

Result:

Simpler hierarchy, eliminate unused `attributes` field from each context instance.


normanmaurer:
@netty-bot test this please

